PR #1631, opened 2020-04-05
LibLine: Add styling; LibJS/JS repl: add syntax highlighting

alimpfard:
Perhaps best served as separate PRs, but they were on the same branch...

These patches mainly add live styling to LibLine, and are my second take on the syntax-highlighted js repl.
If this goes well, we might even be able to provide optional highlighting for the shell itself~

PR #1639, opened 2020-04-05
LibJS: Plumb line and column information through Lexer / Parser

bgianfo:
While debugging LibJS test failures, it's pretty frustrating having to 
resort to printf debugging to figure out what test is failing right now.
While watching your JS Raytracer video it seemed like it was hindering
your progress and causing some frustration as well.

I wanted to start working on improving the diagnostics here.
A good first step seemed to be to start surfacing the information
in the obvious places first like the Lexer / Parser error messages.

In the future I hope we can eventually plumb the info down to the
JS Error classes so any thrown exceptions will contain enough
metadata to know where they came from.

### DIFF
--- a/Libraries/LibJS/Lexer.cpp
+++ b/Libraries/LibJS/Lexer.cpp
@@ -39,7 +39,7 @@ HashMap<char, TokenType> Lexer::s_single_char_tokens;
 
 Lexer::Lexer(StringView source)
     : m_source(source)
-    , m_current_token(TokenType::Eof, StringView(nullptr), StringView(nullptr))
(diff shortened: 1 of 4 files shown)


bgianfo:
Sigh.. looks like this overlaps somewhat with #1631

awesomekling:
All right :)

alimpfard:
@bgianfo you're passing the end position of a token to it -- intentional?

bgianfo:
@alimpfard I was trying to go for naive implementation, end position seemed to accurately describe the trivial test cases I played with on my machine. Start position would be nicer though. Would you mind fixing in your PR?

alimpfard:
np~

PR #1641, opened 2020-04-05
Lagom: Add clang address/memory/undefined-behavior analyzer support

bgianfo:
Adding the ability to turn on Clang analyzer support in the Lagom build.
Right now the following are working warning free on the LibJS test suite:

 -DENABLE_MEMORY_SANITIZER:BOOL=ON

 -DENABLE_ADDRESS_SANITIZER:BOOL=ON

The following analyzer produces errors when running the LibJS test suite:

 -DENABLE_UNDEFINED_SANITIZER:BOOL=ON


awesomekling:
Great idea! :)

PR #1642, opened 2020-04-05
QuickShow: Checkerboard background for transparency

xTibor:
How this PR looks with light and dark themes:

Screenshot_20200405_125415


awesomekling:
Oh man, I love that you made it based on the palette instead of just a constant color :)

PR #1646, opened 2020-04-05
LibJS: Add numeric literal parsing for different bases and exponents

sunverwerth:
WIP because I want to double check everything.
Also this depends on #1645 being merged

### DIFF
--- a/Libraries/LibJS/Tests/numeric-literals-basic.js
+++ b/Libraries/LibJS/Tests/numeric-literals-basic.js
@@ -1,0 +1,20 @@
+try {
+    assert(0xff === 255);
+    assert(0XFF === 255);
+    assert(0o10 === 8);
+    assert(0O10 === 8);
+    assert(0b10 === 2);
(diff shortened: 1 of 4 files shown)


sunverwerth:
I think this is working as it should.
However, I get js test errors for `Math.ceil` and `Math.trunc`.
It looks like `double ceil(double)` is not yet correctly implemented in `math.cpp`...

Libraries/LibJS/Tests/numeric-literals-basic.js
```diff
@@ -0,0 +1,22 @@
+function assert(x) { if (!x) throw 1; }
+
+try {
```
linusg, line 3:
```suggestion
try {
```

(https://github.com/SerenityOS/serenity/commit/bc40908d32934908ca76686c4288faec0a50c240)

sunverwerth: done!

PR #1650, opened 2020-04-05
LibJS: Update tests

linusg:
- The value of `Math.SQRT1_2` was zero as we were dividing two integers.
- Uncomment lines that are now parsing correctly :^)
- By borrowing an "expect close" function from `LibM/TestMath.cpp`, we can make this a lot simpler. Also the parser now understands decimals!


awesomekling:
Looks good but there's a little merge conflict

linusg:
fixed :+1:

PR #1654, opened 2020-04-05
[WIP] LibC+regex: Add first parts of posix compliant regex functionality 

lnzero1dev:
I tried my first way with a bytecode VM. Got lexer and parser code parts from LibJS and CppLexer and combined it with a new VM that is able to match regular expressions. Still very early, but a few tests are already in place and working :-)


awesomekling:
Is this based on an existing VM concept (looks like Trex opcodes)?
How will it handle multiple matches?

lnzero1dev:
Exactly, its Trex, I watched this [video](https://www.youtube.com/watch?v=8ate4QRzl0c) from Alexander Yakushev to get along.

The multiple matching is a good question. I guess you mean the /g or global operator to not stop matching after the first match or to be more precise, find all matches, no matter where they are in the provided haystack?
As from what I can see on the [posix specification](https://pubs.opengroup.org/onlinepubs/7908799/xsh/regex.h.html), this API doesn't allow global matching.
But I guess it would be nice to have it in place for other APIs :-D

A basic Idea that come to my mind would be to iterate over the string, and start matching at each character. If there is a match, continue after the length of the match to match again. :-) This could be done outside the recursive function, I think of something like this 

`
bool VM::match_global(StringView view, size_t& ops_count)
{
    MatchState state;
    state.m_view = view;
    size_t match_length;
    for(size_t i =0; i < view.length(); ++i) {
         i += match_recurse(state);
    }
    ops_count = state.m_ops;
    return ...;
}
`

So it would be outside of the recurse function and the actual VM code.
I guess there are better solutions out there so if anybody know, please help me learning this :-D

lnzero1dev:
Seems to be working like I described above with the match_all :-D But still, if there are more elegent solutions I would be more than happy to learn more about :-)
More test cases to be added...

lnzero1dev:
This was really a good learning experience but there are too many things imperfect. I guess I'll start from the beginning... =)
Some thougths on the learned things / process:

- [x] One observation was that implementing it by trying to fulfill a C API obviously led to a bad C++ API
- [x] There are many flavors of regex, as LibJS is progressing, the engine should at least support POSIX and JavaScript flavor. This can be done by having different Parsers for them.
- [x] Putting such thing in LibC seems to be not the right place. I think AK should fit better.
- [x] I learned a lot about regex and might be able to improve here and there a bit.

I tried to implement regexp search in TextEditor that worked basically, but as TE has a `Vector` of `StringView`s (lines), the search would not find things that go beyond a line break. My current idea is a `MultiStringView` class that has virtual APIs like `VectorIterator item_iterator() const;` and `StringView item_string_view(VectorIterator& it);`. Deriving classes should return the iterator of the vector containing the stringviews in `item_iterator` and the actual `StringView` of the item with `item_string_view`. With this, it could be possible to iterate of the whole document when passing 
a `MultiStringView` to e.g. the match function: `bool match(MultiStringView msv, Pattern p);`

@awesomekling  what do you think about a MultiStringView or something like that, or do you have other idea(s)? :-)

awesomekling:
> This was really a good learning experience but there are too many things imperfect. I guess I'll start over... =)

Sure, that might be a good idea :)

> * [x]  One observation was that implementing it by trying to fulfill a C API obviously led to a bad C++ API

Yeah, this tends to happens. In my experience it's most effective to build out a full C++ library and then wrap it in a C API at the end as needed.

> * [x]  There are many flavors of regex, as LibJS is progressing, the engine should at least support POSIX and JavaScript flavor. This can be done by having different Parsers for them.

Indeed, ideally we'd support both.

> * [x]  Putting such thing in LibC seems to be not the right place. I think AK should fit better.

Sure.

> * [x]  I learned a lot about regex and might be able to improve here and there a bit.

Great :)

> I tried to implement regexp search in TextEditor that worked basically, but as TE has a `Vector` of `StringView`s (lines), the search would not find things that go beyond a line break. My current idea is a `MultiStringView` class that has virtual APIs like `VectorIterator item_iterator() const;` and `StringView item_string_view(VectorIterator& it);`. Deriving classes should return the iterator of the vector containing the stringviews in `item_iterator` and the actual `StringView` of the item with `item_string_view`. With this, it could be possible to iterate of the whole document when passing
> a `MultiStringView` to e.g. the match function: `bool match(MultiStringView msv, Pattern p);`
> 
> @awesomekling what do you think about a MultiStringView or something like that, or do you have other idea(s)? :-)

How about a `TextIterator` interface (implemented as an abstract class) that can be used to traverse arbitrary underlying text? That would allow us to adapt this to searching things like text on web pages as well in the future :)

lnzero1dev:
Thanks for  the input :) Will start a new pull request soon for the regex. About the `TextIterator` I'll think a bit more. Could be a good solution.

PR #1655, opened 2020-04-05
QuickShow: Miscellaneous improvements

xTibor:
Major changes are:

* The layout and mouse handling has been rewritten to always center images in the window.
* QuickShow now accepts multiple images on drag and drop. The first image gets opened in the current window, further images are opened in new processes.
* QSWidget now loads images on its own with QSWidget::load_from_file().
* An on_drop callback has been introduced for QSWidget.
* Added an open menu.
* Added an about box and placeholder icons to the application.
* QuickShow now starts without loading the sunset-retro wallpaper.

### DIFF
--- a/Applications/QuickShow/QSWidget.cpp
+++ b/Applications/QuickShow/QSWidget.cpp
@@ -25,8 +25,6 @@
  */
 
 #include "QSWidget.h"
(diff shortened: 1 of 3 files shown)


awesomekling:
Solid improvements :)

PR #1656, opened 2020-04-05
Resources: Added more file type icon.

NicolasVanBossuyt:
Here is a set of icons for java, javascript, music, pdf, et text.

I didn't find how to make the file manager use these icons :/


linusg:
> I didn't find how to make the file manager use these icons :/

You need to update the code in these three places:

https://github.com/SerenityOS/serenity/blob/6a3b12664a7ce85bef93497df70c0b96cfc509d0/Libraries/LibGUI/FileSystemModel.cpp#L202-L209

https://github.com/SerenityOS/serenity/blob/6a3b12664a7ce85bef93497df70c0b96cfc509d0/Libraries/LibGUI/FileSystemModel.cpp#L412-L429

https://github.com/SerenityOS/serenity/blob/6a3b12664a7ce85bef93497df70c0b96cfc509d0/Libraries/LibGUI/FileSystemModel.h#L150-L157

NicolasVanBossuyt:
Thank you I was searching in the file browser

Libraries/LibGUI/FileSystemModel.cpp
```diff
@@ -425,6 +430,17 @@ Icon FileSystemModel::icon_for_file(const mode_t mode, const String& name) const
         return m_filetype_html_icon;
     if (name.to_lowercase().ends_with(".png"))
         return m_filetype_image_icon;
+    if (name.to_lowercase().ends_with(".cpp"))
```
oriko1010, line 433:
```suggestion
    if (name.to_lowercase().ends_with(".cpp") || name.to_lowercase().ends_with(".h"))
```
(Probably ought to extract all the `to_lowercase()` calls to a single one too if this keeps growing.)

NicolasVanBossuyt: I planned to have a different icon  for .h

awesomekling:
Coolio :)

PR #1658, opened 2020-04-05
Refresh OpenSSL port and make it build the command line utilities

xiaonuofu:
This patch refreshes the openssl port and makes it build the utilities in `apps/`, e.g. the `openssl` utility.

Now you can do this from Serenity:

    $ openssl s_client -connect example.org:443
    ...
    GET / HTTP/1.1
    Host: example.org

    <HTTP response here>

The download URL was bit-rotten and needed a fix.

### DIFF
--- a/Ports/openssl/package.sh
+++ b/Ports/openssl/package.sh
@@ -1,19 +1,20 @@
 #!/bin/bash ../.port_include.sh
 port=openssl
-version=1.0.2t
+branch='1.0.2'
+version="${branch}t"
 useconfigure=true
 configscript=Configure
-files="https://www.openssl.org/source/openssl-${version}.tar.gz openssl-${version}.tar.gz
-https://www.openssl.org/source/openssl-${version}.tar.gz.asc openssl-${version}.tar.gz.asc"
+files="https://ftp.openssl.org/source/old/${branch}/openssl-${version}.tar.gz openssl-${version}.tar.gz
+https://ftp.openssl.org/source/old/${branch}/openssl-${version}.tar.gz.asc openssl-${version}.tar.gz.asc"
 auth_type="sig"
 auth_import_key="8657ABB260F056B1E5190839D9C4D26D0E604491"
 auth_opts="openssl-${version}.tar.gz.asc openssl-${version}.tar.gz"
 
 depends="zlib"
 usr_local=$SERENITY_ROOT/Root/usr/local/
-configopts="--prefix=$usr_local --openssldir=$usr_local/ssl BSD-x86 zlib no-tests no-threads no-asm no-sock"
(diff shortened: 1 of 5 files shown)


Ports/openssl/package.sh
```diff
@@ -3,17 +3,17 @@ port=openssl
 version=1.0.2t
 useconfigure=true
 configscript=Configure
-files="https://www.openssl.org/source/openssl-${version}.tar.gz openssl-${version}.tar.gz
-https://www.openssl.org/source/openssl-${version}.tar.gz.asc openssl-${version}.tar.gz.asc"
+files="https://ftp.openssl.org/source/old/1.0.2/openssl-${version}.tar.gz openssl-${version}.tar.gz
```
deoxxa, line 6: Should the `1.0.2` part here be `${version}` or something instead? Or is this the right value for all versions?

Ports/openssl/patches/ocsp_fd_set.patch
```diff
@@ -0,0 +1,13 @@
+--- a/apps/ocsp.c	2020-04-05 16:49:54.499380971 +0100
++++ b/apps/ocsp.c	2020-04-05 16:50:13.175380426 +0100
+@@ -94,6 +94,10 @@
+ #  endif
+ # endif
+ 
++# if defined(OPENSSL_SYS_SERENITY)
++#  include <fd_set.h>
```
awesomekling, line 8: Please include `<select.h>` instead. `<fd_set.h>` is a non-standard header that only exists so the kernel can include it and get the same exact `fd_set`-related definitions as userspace code. :)

Ports/openssl/patches/s_apps.patch
```diff
@@ -0,0 +1,13 @@
+--- a/apps/s_apps.h	2020-04-05 13:52:44.967691150 +0100
++++ b/apps/s_apps.h	2020-04-05 13:52:50.147690999 +0100
+@@ -122,6 +122,10 @@
+ # define _kbhit kbhit
+ #endif
+ 
++#if defined(OPENSSL_SYS_SERENITY)
++# include <fd_set.h>
```
awesomekling, line 8: Same here.

Ports/openssl/package.sh
xiaonuofu: Thanks, I did miss this.

(I tried to compile `1.1.0l` quickly using the same patches, but they need rework for the newer branch).

Ports/openssl/patches/ocsp_fd_set.patch
xiaonuofu: Thanks, fixed this.

awesomekling:
Okay! :)

PR #1661, opened 2020-04-06
Libraries: LibCrypto + LibTLS

alimpfard:
It is time.
We must march forward towards TLS 1.2

 This PR adds:
- Encryption / Decryption
    - AES_CBC
- Cryptographic Hashing
    - SHA1
    - SHA256 / SHA512
    - MD5
- Authentication / Message Digest
    - HMAC
- Public-Key Cryptographic Systems
    - RSA

- BigInteger support (courtesy of @itamar8910)
    - addition
    - subtraction
    - multiplication
    - divmod

- Number theory "primitives" [`Crypto::NumberTheory`]
    - modular inverse
    - modular exponentiation
    - GCD / LCM
    - probable-prime-test (not rigorously tested yet)

- LibTLS
    - Basic TLS stack
    - code that is a little obnoxious 

now can load https://jcs.org (thanks @jcs) in a record-breaking 40 seconds! (ouch)

shot-2020-04-27_03-57-48


alimpfard:
I just realised I didn't make this a draft
@awesomekling please do not merge this for the time being, at least not until the base classes end up stabilising a bit

alimpfard:
TODO (mostly for me):

- [ ] Basic Cipher Suite
    - [ ] Block Ciphers
        - [X] AES
            - [X] CBC
        - [ ] DES
            - [ ] 3DES - EDE
                - [ ] CBC
    - [ ] Hashing
        - [ ] HMAC
    - [ ] Signatures / checking | SKE
        - [ ] RSA
        - [ ] DSA
        - [ ] ECDSA
        - [ ] DH
- [ ] Playing with TLS 1.2 to figure out what else we need

Libraries/LibCrypto/PK/RSA.h
```diff
@@ -0,0 +1,315 @@
...
+#include <LibCrypto/PK/Code/EMSA_PSS.h>
+#include <LibCrypto/PK/PK.h>
+
+// should also be removed once we have the functions in bigint
+#define HELPER_DEBUG
+
+namespace Crypto {
+namespace PK {
+
+    namespace RSAHelpers {
```
alimpfard, line 39: @itamar8910, this namespace outlines the functions and capabilities that BigInt should have.
note that most of the implementations are fairly inefficient.

Libraries/LibCrypto/BigInt/UnsignedBigInteger.cpp
```diff
@@ -64,18 +64,71 @@ UnsignedBigInteger UnsignedBigInteger::add(const UnsignedBigInteger& other)
...
+            result.m_words.append(static_cast<u32>(after_borrow - static_cast<u64>(other.m_words[i])));
+            borrow = 1;
+        } else {
+            result.m_words.append(m_words[i] - borrow - other.m_words[i]);
+            borrow = 0;
+        }
+    }
+
+    for (size_t i = other.length(); i < length(); ++i) {
+        ASSERT(!(borrow == 1 && m_words[i] == 0));
```
alimpfard, line 97: @itamar8910, This assert trips when
```
UnsignedBigInteger { 1 }.shift_left(256).sub(1)
```

itamar8910: Are you sure you're using the latest commit? I fixed a bug in operator< that also tripped this assert in 9b90722

alimpfard: Does iterating `other.length()` times make sense here? it might have leading zeros.

`a.divide(b)` trips a vector OOB assert, it could be because of this
```
b: 0|51657| a: 65535|
```

musa11971:
@alimpfard You can turn this into a draft if you’d like  
  

https://github.blog/changelog/2020-04-08-convert-pull-request-to-draft/

alimpfard:
Coming soon (:tm:) to a Serenity PR near you!

sunverwerth:
Have you done this before? I wouldn't even know where to start.
Is there a public test suite for this stuff? Anyway, impressive work 👏

alimpfard:
Not in one chunk, no.
There isn't any decent test suite for TLS itself, but you can fight against openssl/mbedtls/whatever for that.

as for where to start...well...you start by reading the RFC and then implementing its absolute minimum requirements in a depth-first fashion, lol

alimpfard:
@awesomekling your HTTP protocol code is too dependent on TCPSocket...
maybe that `class TLSSocket : public Core::TCPSocket` thing will actually happen...

Preferably not as part of this PR though :^)

alimpfard:
It seems LibHTTP is...inevitable...

danboid:
Exciting stuff!

It'd be great to see a serenity native SSH client and server based on this.

alimpfard:
@danboid It would certainly be neat.
Off the top of my head, SSH uses some crypto infrastructure that LibCrypto currently lacks (ECDH or DH), and to be honest...I am quite afraid of the former.

Perhaps once my head is clear I could try working on it.
Or someone else could take a stab at it, I'd certainly love that.

alimpfard:
I \*this\* I'm done adding things to this, GCM or ECDHE can be worked on in a later commit/PR

For now...I shall open the review floodgates and see where we go!

Dexesttp:
The Travis build error doesn't seem accurate (everything built but it just took too long and the job failed I think ?)

I've been skimming through the code without really diving hard in the algorithms, but something jumped out at me : the ASN1, DER, HMAC and a few other files are header-files with a lot of code in them. Is there a reason why they are not split into .cpp/.h files like AES, SHA, etc... are ?

Libraries/LibCrypto/ASN1/PEM.h
```diff
@@ -0,0 +1,74 @@
...
+
+        if (data_in[i] != '-') {
+            // read entire line
+            while ((i < input_length) && (data_in[i] != '\n'))
+                i++;
+            continue;
+        }
+
+        if (data_in[i] == '-') {
+            unsigned int end_idx = i;
```
awesomekling, line 55: Let's use the same type as `i` here (`size_t`)

Libraries/LibCrypto/PK/PK.h
```diff
@@ -0,0 +1,68 @@
...
+
+#include <AK/ByteBuffer.h>
+#include <AK/String.h>
+
+namespace Crypto {
+namespace PK {
+
+// FIXME: Fixing name up for grabs
+template<typename PrivKeyT, typename PubKeyT>
+class PKSystem {
```
Dexesttp, line 37: If you're looking for suggestions, I think C# uses `CryptoServiceProvider`.

Libraries/LibCrypto/PK/RSA.h
```diff
@@ -0,0 +1,234 @@
...
+    Integer m_private_exponent;
+    Integer m_public_exponent;
+    size_t m_length { 0 };
+};
+
+template<typename PubKey, typename PrivKey>
+struct RSAKeyPair {
+    PubKey public_key;
+    PrivKey private_key;
+};
```
Dexesttp, line 112: if I'm not mistaken, the RSA private key contains all the data of the public key. Does it make sense to store both in independent objects ?

Libraries/LibHTTP/Forward.h
```diff
@@ -0,0 +1,10 @@
+
```
Dexesttp, line 1: Missing license header

Libraries/LibTLS/ClientHandshake.cpp
```diff
@@ -0,0 +1,614 @@
...
+        size_t payload_size = buffer[1] * 0x10000 + buffer[2] * 0x100 + buffer[3] + 3;
+#ifdef TLS_DEBUG
+        dbg() << "payload size: " << payload_size << " buffer length: " << buffer_length;
+#endif
+        if (payload_size + 1 > buffer_length)
+            return (i8)Error::NeedMoreData;
+
+        switch (type) {
+        // hello request
+        case 0x00:
```
Dexesttp, line 304: Styling, feel free to ignore : Rather than having comments + magic values in the code, maybe add some `#DEFINE` or constexprs ?

Libraries/LibTLS/ClientHandshake.cpp
```diff
@@ -0,0 +1,614 @@
...
+                write_packet(packet);
+                break;
+            }
+            case Error::DecryptionFailed: {
+                auto packet = build_alert(true, (u8)AlertDescription::DecryptionFailed);
+                write_packet(packet);
+                break;
+            }
+            default:
+                break;
```
Dexesttp, line 545: We catched an unknown error, maybe log something or ASSERT ?

Libraries/LibTLS/ClientHandshake.cpp
```diff
@@ -0,0 +1,614 @@
...
+            }
+            if (payload_res < 0)
+                return payload_res;
+        }
+        switch (write_packets) {
+        case 1:
```
Dexesttp, line 551: Styling, feel free to ignore : Same as above with the constexpr or `#DEFINE`

Libraries/LibTLS/Exchange.cpp
```diff
@@ -0,0 +1,216 @@
...
+{
+    if (!secret.size()) {
+        dbg() << "null secret";
+        return;
+    }
+
+    // RFC 5246: "In this section, we define one PRF, based on HMAC.  This PRF with the
+    // 			  SHA-256 hash function is used for all cipher suites defined in this
+    //            document and in TLS documents published prior to this document when
+    //            TLS 1.2 is negotiated."
```
Dexesttp, line 106: Styling : Formatting issue in this comment ?

Libraries/LibTLS/Handshake.cpp
```diff
@@ -0,0 +1,159 @@
...
+#include <LibCore/Timer.h>
+#include <LibCrypto/ASN1/DER.h>
+#include <LibCrypto/PK/Code/EMSA_PSS.h>
+#include <LibTLS/TLSv12.h>
+
+namespace TLS {
+
+ByteBuffer TLSv12::build_hello()
+{
+    // arc4random_buf(&m_context.local_random, 32);
```
Dexesttp, line 36: In which case should we enable or disable creating a new random number each time ?

Libraries/LibTLS/Record.cpp
```diff
@@ -0,0 +1,316 @@
...
+
+            return (i8)Error::IntegrityCheckFailed;
+        }
+        plain = decrypted.slice(0, length - mac_size);
+    }
+    m_context.remote_sequence_number++;
+
+    switch (type) {
+    case MessageType::ApplicationData:
+        if (m_context.connection_status != 0xff) {
```
Dexesttp, line 246: Styliing, feel free to ignore : Here and below in ChangeCipher, we're checking `connection_status` against magic values. Maybe add constants and/or `#DEFINE`s about it ?

Libraries/LibTLS/Socket.cpp
```diff
@@ -0,0 +1,212 @@
...
+
+ByteBuffer TLSv12::read(size_t max_size)
+{
+    if (m_context.application_buffer.size()) {
+        auto length = min(m_context.application_buffer.size(), max_size);
+        auto buf = m_context.application_buffer.slice(0, length);
+        m_context.application_buffer = m_context.application_buffer.slice(length, m_context.application_buffer.size() - length);
+        return buf;
+    } else
+        return {};
```
Dexesttp, line 52: I don't know what the coding style is, but `if { /* curlies */ } else /* Single statement */` feels weird.

Dexesttp: Thinking about it, you can just skip the else entirely and return.

Libraries/LibCrypto/PK/Code/EMSA_PSS.h
```diff
@@ -0,0 +1,179 @@
...
+
+            for (size_t i = 0; i < mask_length; ++i)
+                DB[i] = masked_DB[i] ^ DB_mask[i];
+
+            DB[0] &= 0xff >> (8 - length_to_check);
+
+            auto check_octets = emsg.size() - HashFunction::DigestSize - SaltLength - 2;
+            for (size_t i = 0; i < check_octets; ++i)
+                if (DB[i])
+                    return VerificationConsistency::Inconsistent;
```
awesomekling: Please wrap multi-line statements in `{ }`

Libraries/LibCrypto/BigInt/UnsignedBigInteger.cpp
```diff
@@ -312,4 +312,32 @@ UnsignedBigInteger UnsignedBigInteger::create_invalid()
...
+
+    return integer;
+}
+
+size_t UnsignedBigInteger::export_data(AK::ByteBuffer& data)
+{
+    UnsignedBigInteger copy { *this };
+
+    size_t size = trimmed_length() * sizeof(u32);
+    size_t i { 0 };
```
awesomekling, line 334: Let's be consistent wrt `{}` vs `=` initialization.

Libraries/LibTLS/TLSPacketBuilder.h
```diff
@@ -0,0 +1,118 @@
...
+
+    PacketBuilder(MessageType type, Version version, size_t size_hint = 0xfdf)
+    {
+        m_packet_data = ByteBuffer::create_uninitialized(size_hint + 16);
+        m_current_length = 5;
+        m_packet_data[0] = (u8)type;
+        *(u16*)m_packet_data.offset_pointer(1) = convert_between_host_and_network((u16)version);
+    }
+
+    inline void append(u16 value)
```
awesomekling, line 63: `inline` is meaningless in this context; the function is implicitly inline by having its body in the declaration.

Libraries/LibTLS/TLSPacketBuilder.h
```diff
@@ -0,0 +1,118 @@
...
+    inline void set(size_t offset, u8 value)
+    {
+        ASSERT(offset < m_current_length);
+        m_packet_data[offset] = value;
+    }
+    size_t length() const { return m_current_length; }
+
+private:
+    ByteBuffer m_packet_data;
+    size_t m_current_length;
```
awesomekling, line 116: Please explicitly initialize primitive values.

Libraries/LibTLS/TLSv12.cpp
```diff
@@ -0,0 +1,1800 @@
...
+    dbgprintf("\n");
+}
+}
+
+using namespace Crypto;
+
+namespace {
+struct OIDChain {
+    void* root;
+    u8* oid;
```
awesomekling, line 37: Please explicitly initialize raw pointers.

Libraries/LibTLS/TLSv12.cpp
```diff
@@ -0,0 +1,1800 @@
...
+    _parse_asn1(m_context, cert, buffer.data(), buffer.size(), 1, fields, nullptr, 0, nullptr, nullptr);
+
+    dbg() << "Certificate issued for " << cert.subject << " by " << cert.issuer_subject;
+
+    return cert;
+}
+
+ByteBuffer TLSv12::build_hello()
+{
+    // arc4random_buf(&m_context.local_random, 32);
```
awesomekling: Should this really be commented out? ;)

Libraries/LibTLS/TLSv12.cpp
```diff
@@ -0,0 +1,516 @@
...
+                if (cert.der.size() < temp) {
+                    cert.der.grow(temp);
+                } else {
+                    cert.der.trim(temp);
+                }
+                cert.der.overwrite(0, buffer + start_position, temp);
+            }
+        } else {
+            switch (type) {
+            case 0x00:
```
Dexesttp, line 187: Styling, feel free to ignore : Same remark about magic values here, what are they used for / can they be replaced with `#DEFINE`, constexpr or enums ?

Libraries/LibTLS/TLSv12.h
```diff
@@ -0,0 +1,400 @@
...
+
+    bool expand_key();
+
+    bool compute_master_secret(size_t length);
+
+    Context m_context;
+    RefPtr<Core::Notifier> m_notifier;
+
+    OwnPtr<Crypto::Cipher::AESCipher::CBCMode> m_aes_local { nullptr };
+    OwnPtr<Crypto::Cipher::AESCipher::CBCMode> m_aes_remote { nullptr };
```
awesomekling, line 407: No need to explicitly initialize smart pointers; they are null by default.

Libraries/LibTLS/TLSv12.h
```diff
@@ -0,0 +1,400 @@
...
+            set_error(errno);
+        } else {
+            set_fd(fd);
+            set_mode(IODevice::ReadWrite);
+            set_error(0);
+        }
+    }
+
+    ByteBuffer& write_buffer() { return m_context.tls_buffer; }
+    bool established() const { return m_context.connection_status == 0xff; }
```
awesomekling, line 248: `is_established`

Libraries/LibTLS/TLSv12.h
```diff
@@ -0,0 +1,400 @@
...
+    bool load_certificates(const ByteBuffer& pem_buffer);
+    bool load_private_key(const ByteBuffer& pem_buffer);
+
+    ByteBuffer finish_build();
+
+    const StringView& alpn() const { return m_context.negotiated_alpn; }
+    void add_alpn(const StringView& alpn);
+    bool has_alpn(const StringView& alpn) const;
+
+    bool cipher_supported(CipherSuite suite) const
```
awesomekling, line 270: `supports_cipher`

Libraries/LibTLS/TLSv12.h
```diff
@@ -0,0 +1,400 @@
...
+    {
+        return suite == CipherSuite::RSA_WITH_AES_128_CBC_SHA256 || suite == CipherSuite::RSA_WITH_AES_256_CBC_SHA256;
+    }
+
+    bool version_supported(Version v) const
```
awesomekling, line 275: `supports_version`

Libraries/LibTLS/TLSv12.h
```diff
@@ -0,0 +1,400 @@
...
+    void set_sni(const StringView& sni)
+    {
+        if (m_context.is_server || m_context.critical_error || m_context.connection_status != 0) {
+            dbg() << "invalid state for set_sni";
+            return;
+        }
+        m_context.SNI = sni;
+    }
+
+    Optional<Certificate> parse_asn1(const ByteBuffer& buffer, bool client_cert = false);
```
awesomekling, line 260: Can this be `const`?

Libraries/LibTLS/TLSv12.cpp
```diff
@@ -0,0 +1,516 @@
...
+        hash_kind = Crypto::Hash::HashKind::SHA1;
+        break;
+    case Crypto::Hash::SHA256::DigestSize:
+        hash_kind = Crypto::Hash::HashKind::SHA256;
+        break;
+    case Crypto::Hash::SHA512::DigestSize:
+        hash_kind = Crypto::Hash::HashKind::SHA512;
+        break;
+    default:
+        break;
```
Dexesttp, line 506: If we didn't detect the digest size, should we do something ? Or is the passthrough with `hash_kind = None` enough ?

Servers/ProtocolServer/HttpsProtocol.cpp
```diff
@@ -0,0 +1,49 @@
+/*
+ * Copyright (c) 2018-2020, Andreas Kling <kling@serenityos.org>
```
Dexesttp, line 2: That's probably not right :^)

Servers/ProtocolServer/HttpsProtocol.h
```diff
@@ -0,0 +1,37 @@
+/*
+ * Copyright (c) 2018-2020, Andreas Kling <kling@serenityos.org>
```
Dexesttp, line 2: Same as above

Userland/Makefile
```diff
@@ -4,7 +4,7 @@ APPS = ${SRCS:.cpp=}
 
 EXTRA_CLEAN = $(APPS)
 
-LIB_DEPS = Web GUI Gfx Audio Protocol IPC Thread Pthread PCIDB Markdown JS Core Line X86 Debug
+LIB_DEPS = Crypto TLS Web GUI Gfx Audio Protocol IPC Thread Pthread PCIDB Markdown JS Core Line X86 Debug
```
Dexesttp, line 7: Possibility : Should we add HTTP as a precaution, or just wait until it's required ?

awesomekling:
Also, please explicitly initialize primitive variables. :)

Libraries/LibTLS/TLSv12.h
alimpfard: You will keep telling me this, and I will keep doing it because my brain goes "these start null"

Libraries/LibCrypto/Cipher/Cipher.h
```diff
@@ -0,0 +1,139 @@
...
+class Cipher;
+
+struct CipherBlock {
+public:
+    explicit CipherBlock(PaddingMode mode)
+        : m_padding_mode(mode)
+    {
+    }
+
+    static size_t block_size() { ASSERT_NOT_REACHED(); }
```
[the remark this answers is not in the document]
alimpfard: That is for children to implement, but `virtual static` is nonsensical.
It's mostly needed as a constexpr nicety, I guess

Libraries/LibCrypto/Hash/MD5.h
```diff
@@ -0,0 +1,131 @@
...
+
+struct MD5Digest {
+    constexpr static size_t Size = 16;
+    u8 data[Size];
+
+    const u8* immutable_data() const { return data; }
+    size_t data_length() { return Size; }
+};
+
+namespace MD5Constants {
```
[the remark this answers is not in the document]
alimpfard: Ah, the first half is used in the header itself, the second half is for the cpp file.

I guess in this one header, it's fine to move these to the top, they're not _that_ long

Libraries/LibCrypto/NumberTheory/ModularFunctions.h
```diff
@@ -0,0 +1,202 @@
...
+
+    Vector<UnsignedBigInteger, 256> tests;
+    UnsignedBigInteger seven { 7 };
+    for (size_t i = 0; i < tests.size(); ++i)
+        tests.append(random_number(seven, p.sub(2)));
+
+    return MR_primality_test(p, tests);
+}
+
+static UnsignedBigInteger random_big_prime(size_t bits)
```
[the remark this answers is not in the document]
alimpfard: well...the function returns a fairly-likely-to-be-prime number, and everyone seems to accept that if it passes some 200 primality tests, it's usable *as* a prime number anyway

Libraries/LibCrypto/Hash/MD5.h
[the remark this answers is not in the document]
Dexesttp: Oh, I didn't see that the first part was used. But it would be easier to read that way, thanks : )

Libraries/LibCrypto/PK/RSA.cpp
```diff
@@ -0,0 +1,309 @@
...
+                ASN1::Kind::Integer, 1, &e,
+                ASN1::Kind::Integer, 1, &d)) {
+            dbg() << "bad privkey " << n << " " << e << " " << d;
+            return keypair;
+        }
+        keypair.private_key.set(n, d, e);
+        return keypair;
+    }
+    if (n == 1) {
+        // multiprime key, we don't know how to deal with this
```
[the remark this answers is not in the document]
alimpfard: everything else returns an empty keypair, I don't think asserting on such remote-controlled data is a good idea.

I am...unsure about this.

Libraries/LibCrypto/PK/RSA.h
alimpfard: Yes, we might have a public key but no private key.

The data stored here is actually a subset of all the data, since we don't make much use of the private key, I opted to have it barebone.

Libraries/LibTLS/Handshake.cpp
alimpfard: we should always create a new random number, that was commented because I wanted the smallest number of invariants while testing

Servers/ProtocolServer/HttpsProtocol.cpp
alimpfard: I WILL BLAME ANDREAS FOR THIS

lol

Userland/Makefile
alimpfard: There's no particular need for HTTP, at least, not until someone (re)develops `curl` or whatever

Libraries/LibCrypto/PK/RSA.cpp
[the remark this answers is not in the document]
Dexesttp: Maybe `ASSERT`ing is too strong, yes.

Overall, the `parse_rsa_key` feels like it should return an `Optional<RSA::KeyPairType>` given that it fails so often, but the public and private keys are also "optional" (The `import_{*}_key` methods are even checking for existence of the individual keys with the `public_key.length()` and `private_key.length()` methods like they would with `.has_value()`).

I think maybe replacing the `KeyPairType` members with `Optional<Public/Private>` keys instead of direct values seems like it would be more explicit, but it would also be more unwieldy / harder to use so maybe the current code is fine too.

alimpfard:
squashing these fixes brings tears to my eyes

;;

alimpfard:
@Dexesttp some of them are because of C++'s nonsense about "templated stuff must be visible at callsite".
the ASN1/DER parser...well, I actually have no idea.

Libraries/LibTLS/TLSv12.cpp
alimpfard: I guess we'll log something at least

Libraries/LibCrypto/PK/PK.h
alimpfard: Isn't that a little _too_ general?
This is specifically a public-key cryptography system...

Libraries/LibTLS/Exchange.cpp
alimpfard: oh...I blame RFC texts using hard tabs

Libraries/LibTLS/Record.cpp
alimpfard: Absolutely.
I just spent non-zero time thinking what each value meant, putting them in an enum is very much the way to go

Libraries/LibTLS/TLSv12.h
alimpfard: it sure can.

Libraries/LibCrypto/NumberTheory/ModularFunctions.h
```diff
@@ -0,0 +1,202 @@
...
+        if (return_)
+            return false;
+    }
+
+    return true;
+}
+
+static UnsignedBigInteger random_number(const UnsignedBigInteger& min, const UnsignedBigInteger& max)
+{
+    auto range = max.minus(min);
```
Dexesttp, line 161: Maybe add an `ASSERT(min < max)` ? Otherwise, this would result in `range` being an invalid `UnsignedBigInteger` but its trimmed length doesn't take that into account.

alimpfard: :thinking: what would it do in that case, I wonder?
I'll add the assert~

Libraries/LibCrypto/NumberTheory/ModularFunctions.h
```diff
@@ -0,0 +1,202 @@
...
+    UnsignedBigInteger seven { 7 };
+    for (size_t i = 0; i < tests.size(); ++i)
+        tests.append(random_number(seven, p.minus(2)));
+
+    return MR_primality_test(p, tests);
+}
+
+static UnsignedBigInteger random_big_prime(size_t bits)
+{
+    UnsignedBigInteger min = UnsignedBigInteger::from_base10("6074001000").shift_left(bits - 33);
```
Dexesttp, line 192: Same here, let's add an `ASSERT(bits >= 33)` right before ?
If `bits` is less than 33, I think this would wrap around and we would end up with a lot of used memory, as well as `random_number(<big number>, <small number>)` which doesn't seem handled either.

Right now, it's only used when creating new RSA pairs and we never use them with less than 128 bits, so we won't hit that case.

alimpfard:
I kinda hate how github shows these commits based on their timestamp, as opposed to where they are in the commit history

Libraries/LibTLS/ClientHandshake.cpp
```diff
@@ -0,0 +1,613 @@
...
+
+        dbg() << "extension " << extension_type << " with length " << extension_length;
+        if (extension_length) {
+            if (buffer.size() - res < extension_length) {
+                dbg() << "not enough data for extension";
+                return (i8)Error::NeedMoreData;
+            }
+
+            // SNI
+            if (extension_type == 0x00) {
```
Dexesttp, line 151: I'm guessing these are some of the the TLS hello extensions that we support (from [the TLS extensiontype values website](http://www.iana.org/assignments/tls-extensiontype-values/tls-extensiontype-values.xhtml), I guess this one would be `server_name` and the ones below would be `application_layer_protocol_negotiation` and `signature_algorithms`). Let's put that in an enum for clarity ?

Libraries/LibTLS/ClientHandshake.cpp
```diff
@@ -0,0 +1,613 @@
...
+            }
+            break;
+        case ClientHello:
+            // FIXME: We only support client mode right now
+            if (m_context.is_server) {
+                ASSERT_NOT_REACHED();
+            }
+            payload_res = (i8)Error::UnexpectedMessage;
+            break;
+        case 0x02:
```
Dexesttp, line 329: This should be the `ServerHello` value of the `HandshakeType` enum, I guess ?

Libraries/LibTLS/ClientHandshake.cpp
```diff
@@ -0,0 +1,613 @@
...
+            if (payload_res > 0) {
+                memset(m_context.handshake_messages, 0, sizeof(m_context.handshake_messages));
+            }
+            break;
+        default:
+            dbg() << "message type not understood: " << type;
+            return (i8)Error::NotUnderstood;
+        }
+
+        if (type != 0x00) {
```
Dexesttp, line 479: Same here with `HelloRequest`

Libraries/LibTLS/Handshake.cpp
```diff
@@ -0,0 +1,167 @@
...
+
+ByteBuffer TLSv12::build_hello()
+{
+    arc4random_buf(&m_context.local_random, 32);
+
+    auto packet_version = (u16)m_context.version;
+    auto version = (u16)m_context.version;
+    PacketBuilder builder { MessageType::Handshake, packet_version };
+    // client hello
+    builder.append((u8)0x1);
```
Dexesttp, line 42: `ClientHello` from `HandshakeType` ?

Libraries/LibTLS/Handshake.cpp
```diff
@@ -0,0 +1,167 @@
...
+        sni_length = m_context.SNI.length();
+
+    if (sni_length)
+        extension_length += sni_length + 9;
+
+    builder.append((u16)extension_length);
+
+    if (sni_length) {
+        // SNI extension
+        builder.append((u16)0x00);
```
Dexesttp, line 98: If there is an HelloExtensionsEnum created, these values could be replaced as well.

alimpfard: Good idea, I'll do that

Libraries/LibTLS/ClientHandshake.cpp
alimpfard: enums for days :^)

Libraries/LibCrypto/Hash/SHA1.cpp
```diff
@@ -0,0 +1,183 @@
...
+
+namespace Crypto {
+namespace Hash {
+
+inline static constexpr auto ROTATE_LEFT(u32 value, size_t bits)
+{
+    return (value << bits) | (value >> (32 - bits));
+}
+
+inline static constexpr void Round0(u32 v, u32& w, u32 x, u32 y, u32& z, u32 block_i)
```
Dexesttp, line 38: That might be just me missing something, but are these constexpr unused ? I can't see them mentionned anywhere else.

Libraries/LibTLS/ClientHandshake.cpp
```diff
@@ -0,0 +1,613 @@
...
+            }
+            if (payload_res < 0)
+                return payload_res;
+        }
+        switch (write_packets) {
+        case WritePacketStage::Initial:
+            // nothing to write
+            break;
+        case WritePacketStage::ClientHandshake:
+            if (m_context.client_verified == 2) {
```
Dexesttp, line 552: When would this get to 2 ? `client_verified` never seems to be modified.

Libraries/LibTLS/Exchange.cpp
```diff
@@ -0,0 +1,216 @@
...
+    }
+
+    auto key_size = key_length();
+    auto mac_size = mac_length();
+    auto iv_size = iv_length();
+
+    pseudorandom_function(
+        key_buffer,
+        m_context.master_key,
+        (const u8*)"key expansion", 13,
```
Dexesttp, line 51: I'm guessing the secret for this PRNM function is "key expansion" for now, and that works. Is there any plans / need to change that in the future ?

Libraries/LibTLS/Exchange.cpp
```diff
@@ -0,0 +1,216 @@
...
+ByteBuffer TLSv12::build_certificate()
+{
+    dbg() << "FIXME: build_certificate";
+    return {};
+}
+
+ByteBuffer TLSv12::build_change_cipher_spec()
+{
+    PacketBuilder builder { MessageType::ChangeCipher, m_context.version, 64 };
+    builder.append((u8)1);
```
Dexesttp, line 176: That is a magic value, but the ChangeCipher spec only allows a value of 1 so I guess it's OK to not create a constant.

Libraries/LibTLS/Exchange.cpp
```diff
@@ -0,0 +1,216 @@
...
+ByteBuffer TLSv12::build_server_key_exchange()
+{
+    dbg() << "FIXME: build_server_key_exchange";
+    return {};
+}
+
+ByteBuffer TLSv12::build_client_key_exchange()
+{
+    PacketBuilder builder { MessageType::Handshake, m_context.version };
+    builder.append((u8)0x10);
```
Dexesttp, line 192: `ClientKeyExchange` ?

Libraries/LibTLS/Handshake.cpp
```diff
@@ -0,0 +1,167 @@
...
+    auto packet = builder.build();
+    update_packet(packet);
+
+    return packet;
+}
+
+ByteBuffer TLSv12::build_finished()
+{
+    PacketBuilder builder { MessageType::Handshake, m_context.version, 12 + 64 };
+    builder.append((u8)0x14);
```
Dexesttp, line 148: `Finsihed` from `HandshakeType` ?

Libraries/LibTLS/Record.cpp
```diff
@@ -0,0 +1,316 @@
...
+
+void TLSv12::update_packet(ByteBuffer& packet)
+{
+    u32 header_size = 5;
+    *(u16*)packet.offset_pointer(3) = convert_between_host_and_network((u16)(packet.size() - header_size));
+
+    if (packet[0] != (u8)MessageType::ChangeCipher) {
+        if (packet[0] == (u8)MessageType::Handshake && packet.size() > header_size) {
+            u8 handshake_type = packet[header_size];
+            if (handshake_type != 0x00 && handshake_type != 0x03) {
```
Dexesttp, line 47: `HelloRequest` and `HelloVerifyRequest` ?

Libraries/LibTLS/Exchange.cpp
alimpfard: That's not the secret, that's a "label", it's a constant defined in the RFC
I was quite perplexed about it at first too

Libraries/LibCrypto/Hash/SHA1.cpp
alimpfard: I think that's a leftover from copying the source off SHA2
it's probably unused, I'll remove it and see

Libraries/LibTLS/ClientHandshake.cpp
alimpfard: that would be my brain going "what if the server asks for a certificate?", writing this, then deciding not to handle it at all

alimpfard: I'll add another enum for that, I guess

Libraries/LibTLS/Exchange.cpp
alimpfard: You can see similar things with "client finished" and "master secret"

Libraries/LibTLS/Exchange.cpp
alimpfard: I'll just leave it as-is then.

alimpfard:
> Sorry for requesting so many enums

can't be mad about something that clearly helps with readability.

The DER parser is actually a little...duplicated, since X509 proved a bit too complicated for the little parser; I hope someone finds the energy to deal with DER (I hated every damn bit about it).

Libraries/LibCrypto/Cipher/AES.cpp
```diff
@@ -0,0 +1,426 @@
...
+    m_data.overwrite(0, data, length);
+    if (length < m_data.size()) {
+        switch (padding_mode()) {
+        case PaddingMode::Null:
+            // fill with zeros
+            __builtin_memset(m_data.data() + length, 0, m_data.size() - length);
+            break;
+        case PaddingMode::CMS:
+            // fill with the length of the padding bytes
+            __builtin_memset(m_data.data() + length, m_data.size() - length - 1, m_data.size() - length);
```
Dexesttp, line 415:
```suggestion
            __builtin_memset(m_data.data() + length, m_data.size() - length, m_data.size() - length);
```
The AES encryption tests fail here.

This took me a while to find, but RFC 5652 section 6.3 describes the CMS padding value as "k-(lth mod k)", which should be "block_size - length" without any -1 required. This should solve the issue of non-specialized encrypted data for AES.

Note that for the PaddingMode "Bit", the value at the end should indeed be "block_size - length - 1", but that's not needed / implemented yet.

Libraries/LibCrypto/Cipher/Mode/Mode.h
```diff
@@ -0,0 +1,105 @@
...
+        auto size = data.size();
+        switch (m_cipher.padding_mode()) {
+        case PaddingMode::CMS: {
+            auto maybe_padding_length = data[size - 1];
+            if (maybe_padding_length >= T::block_size() - 1) {
+                // cannot be padding (the entire block cannot be padding)
+                return;
+            }
+            for (auto i = maybe_padding_length; i > 0; --i) {
+                if (data[size - i - 1] != maybe_padding_length) {
```
Dexesttp, line 72:
```suggestion
            for (auto i = size - maybe_padding_length; i < size; ++i) {
                if (data[i] != maybe_padding_length) {
```

The current code has an off-by-one error due to the same issue as with the encryption (CMS padding is actually "data_size - length" without any -1).

The simpler fix is I think to stop the iteration to `i > 1`.

Styling-wise, I propose reworking the loop like this. It's functionally equivalent to the existing one, but I find it more intuitive to think about iterating from "somewhere in the middle of the block" to "the end of the block", rather than how it's written right now (backwards over the length of the padding but starting from the end of the block).

Dexesttp: I forgot to mention this in the review, but the `data.trim()` line right below this should also be changed to match the value : `data.trim(size - maybe_padding_length - 1);` => `data.trim(size - maybe_padding_length);`

Libraries/LibCrypto/Cipher/AES.cpp
alimpfard: I could swear there was a reason I did that.
perhaps I had mixed up the padding types? I'll investigate that and come back~

alimpfard: It's a weirdness that RFC 5246 created, apparently TLS v1.2 uses its own padding scheme (https://tools.ietf.org/html/rfc5246#section-6.2) down in the "padding_length" section, but since it looks like CMS, I thought my CMS implementation needed tweaking.
I'll add a new "RFC5246" padding mode, I suppose.

Dexesttp: It's already nuts how many documents are needed to implement TLS, and then there's this kind of things... Kudos to you to make this PR happen ^^

alimpfard:
> the RSA PKCS #1 1.5 and RSA | Encrypt tests ran for about 10 minutes on my machine, each.

I'm surprised you waited for that long lol

> Most of the time is spent allocating a lot of vectors for BigInt operations (this is expected, but still seems a bit excessive).

That's mainly because the bigint code copies...a lot of vectors.
I'm not _entirely_ sure what could be done about that, aside from providing mutating versions of the operations?

> I didn't spend much time analyzing why RSA decryption is lagging so much, but it is very computationally-heavy and the BigInt implementation is notvery optimized, so I don't think there's anything fundamentally wrong with it.

It's by far the slowest part of the whole thing (even encryption takes a good amount of time, even with the pubkey being so small)

alimpfard:
I am not sure what "Specialised Encrypt" is testing...teaches me to name my tests properly...

Libraries/LibCrypto/Cipher/AES.cpp
alimpfard: Yeah...the RFC-hopping was probably my least favourite part of doing this.
Issues like this one, luckily, can be found with a new perspective (I'm surprised I never ran the AES tests after starting TLS...).

It's been quite the learning experience, I'm happy I did this!

Dexesttp:
I did some more analysis of why RSA decryption is lagging, and it's pretty much as expected : when decrypting, we need to run the Modular Power operator for a big number with a big exponent, it's an iteration doing three to five multiplications / modulos for the entire size of the exponent (in bits), and it's way too much for the allocator which has to allocate a new bigint with each operation.

Even when setting the default BigInt size to 256 _words_ (8000 bits long !), to prevent vector growth, the vector creations themselves just take too much time (and there's still some vector growth, which shows how big some of the numbers are).

I have a few ideas about how to optimize this :
- Add some "{plus/minus/multiply_by/remainder_by/shift_by/divider_by}_into" operations which would only use already created BigInt passed by reference. That should drastically help reduce the number of allocations, and be easier to implement than copyless in-place operations, but it would be a pretty unnatural APIs since we'd have to pass in temporary BigInt variables for e.g. mupltiply/divides.
- As you proposed, add an in-place modulo multiplication to BigInts to replace all of the `x = x.multiply().divide().remainder` operations which are the bulk of the work, but I have no idea if it's even a possibility / if there's an algorithm.
- Change the algorithm to use the Montgomery multipication algorithm, which at least bounds the values in each step to the modulo's size (so fewer resizes and fewer malloc calls), but it wouldn't be that much of an improvement since we would still execute a lot of BigInt multiplications.
- Do some combination of the three ideas above where we have some in-place operations, some "non-allocating" operations, and an algorithm which limits the size of the numbers.

I'll try to iterate on these ideas during the next few days, but I don't think this RSA lag isssue should be blocking for the PR to go through !

alimpfard:
Neat~

I'll try to change as many `Vector::append()`'s as I can to `::unchecked_append()`'s, that should probably make this quite a bit faster, in lieu of inplace operations and such.

> I'll try to iterate on these ideas during the next few days

Please do!

AK/Vector.h
```diff
@@ -250,7 +250,7 @@ class Vector {
     // NOTE: Vector::is_null() exists for the benefit of String::copy().
     bool is_null() const { return false; }
     bool is_empty() const { return size() == 0; }
-    size_t size() const { return m_size; }
+    [[gnu::always_inline]] inline size_t size() const { return m_size; }
```
Dexesttp, line 253: With the 888e35f0fed393e8496414e5c285cf4ca4788f67 commit, this is now available as the `ALWAYS_INLINE` macro from `AK/Platform.h`

alimpfard: I'd smite that macro to depths of git history if this codebase was mine. grrrrr.

I'll try to remember and change them, along with the stuff in bigint

Zenyiar:
I downloaded your fork with the changes to allow https but it didnt work, did I do something wrong?

alimpfard:
"it didn't work" is a pretty broad term, if the browser says HTTP Load failed, that's likely because the handshake failed (us and the server do not implement/accept an agreeable ciphersuite)
if it's something else, I'd gladly take your debug log

Zenyiar:
Welp it was like any HTTPS website i went to, even my website that has like 0 things on it wouldnt load

alimpfard:
Well the currently implemented ciphersuites are not very widely accepted, _some_ servers accept them. most do not.

Zenyiar:
k

awesomekling:
I think we're good to merge here. There's a bunch of work to be done to improve performance and security of this code, but for a first import I think it's okay :)

alimpfard:
@awesomekling indeed. Security was not much of a focus here, most of the checks are still just marked as FIXME afterall :^)

And performance...well, I hope we manage to avoid all the copying in bigint somehow.

alimpfard:
:tada:

PR #1663, opened 2020-04-06
Kernel: Ensure non-pseudo file systems are backed by a file - part 2

supercomputer7:


awesomekling:
No need to mention that these are reverts of reverts.

Also, if you are reverting something but also making changes to it in the same commit, it's not really a revert :)

Base/usr/share/man/man2/mount.md
```diff
@@ -37,14 +39,22 @@ mounted file system.
...
+* `EFAULT`: The `fs_type` or `target` are invalid strings.
 * `EPERM`: The current process does not have superuser privileges.
-* `ENODEV`: The `fs_type` is unrecognized, or the device is not found, or the device doesn't contain a valid filesystem image.
+* `ENODEV`: The `fs_type` is unrecognized, or the file descriptor to source is
+  not found, or the source doesn't contain a valid filesystem image. Also, this
+  error occurs if `fs_type` is valid, but the file descriptor from `source_fd`
+  is not seekable.
+* `EBADF`: If the `source_fd` is not valid, and either `fs_type` specifies a
+  file-backed filesystem (and not a pseudo filesystem), or `MS_BIND` is
+  specified in flags.
```
awesomekling, line 57: This doesn't render correctly in the `man` or `Help` programs. Seems like we don't support newlines in the middle of bullet point lists.

PR #1665, opened 2020-04-06
LibJS: rename variable "max" to "min" in MathObject::min()

emanuele6:
Just giving this variable a more appropriate name.


awesomekling:
:+1:

PR #1666, opened 2020-04-06
LibJS: Fix some tests for Math.min()

emanuele6:
In some of the tests in Math.min.js, we were testing Math.max() instead
of Math.min()


awesomekling:
Oops, good catch!

PR #1671, opened 2020-04-06
LibM: Improve pow() and powf()

linusg:
Improving math functions part 2.

- If `y` in `x^y` is `0`, the result of `x^y` is `1`
- If `y` in `x^y` is `1`, the result of `x^y` is `x`
- If `y` in `x^y` is any positive integer, the result of `x^y` is `x` multiplied by itself `y - 1` times
- If `y` in `x^y` is any negative integer, the result of `x^y` is `1` divided by (`x` multiplied by itself `y - 1` times)

For decimal `y` we still use the current `exp(y * log(x))` with inaccuracies to some degree.

JS tests are all green again :^)

### DIFF
--- a/Libraries/LibM/math.cpp
+++ b/Libraries/LibM/math.cpp
@@ -101,12 +101,39 @@ float sinf(float angle)
 
 double pow(double x, double y)
 {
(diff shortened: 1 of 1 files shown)


awesomekling:
Great!

PR #1672, opened 2020-04-06
LibJS: Fix impossible member access for negative integers

Dexesttp:
I saw that in the last video, and thought that it would make a good first contribution :^)

The PropertyName class able to match a number or an array can only accept positive numerical values. However, the computed_property_name method sometimes returned negative values.

This commit also adds a basic object access test case for this and other standard access.


awesomekling:
Great first contribution! :)

PR #1673, opened 2020-04-06
LibJS: Allow parsing numeric and string literals in object expressions

Dexesttp:
This adds parsing for objects matching the patterns :
`o = { 1: "test", 3.14: "pi", "foo": "bar" }`

Also updated the object-basic.js test to include this change !


sunverwerth:
Negative numbers are not allowed, I think. At least Esprima and the Firefox console show this as a syntax error.

Dexesttp:
You're right, they aren't. This makes it a lot easier then ! Fixed :)
Thanks a lot for the info !

awesomekling:
:+1:

PR #1682, opened 2020-04-07
LibJS: Add Boolean constructor & fix conversion of NaN to bool

jack-karamanian:
This PR:
- Adds the [`Boolean`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Boolean) constructor
- Fixes the conversion of NaN to bool (should be false)

### DIFF
--- a/Libraries/LibJS/Interpreter.cpp
+++ b/Libraries/LibJS/Interpreter.cpp
@@ -28,6 +28,7 @@
 #include <LibJS/AST.h>
 #include <LibJS/Interpreter.h>
 #include <LibJS/Runtime/ArrayPrototype.h>
+#include <LibJS/Runtime/BooleanPrototype.h>
 #include <LibJS/Runtime/DatePrototype.h>
 #include <LibJS/Runtime/Error.h>
 #include <LibJS/Runtime/ErrorPrototype.h>
@@ -55,6 +56,7 @@ Interpreter::Interpreter()
     m_error_prototype = heap().allocate<ErrorPrototype>();
     m_date_prototype = heap().allocate<DatePrototype>();
     m_number_prototype = heap().allocate<NumberPrototype>();
+    m_boolean_prototype = heap().allocate<BooleanPrototype>();
 }
 
 Interpreter::~Interpreter()
@@ -180,6 +182,7 @@ void Interpreter::gather_roots(Badge<Heap>, HashTable<Cell*>& roots)
     roots.set(m_date_prototype);
     roots.set(m_function_prototype);
     roots.set(m_number_prototype);
+    roots.set(m_boolean_prototype);
 
     roots.set(m_exception);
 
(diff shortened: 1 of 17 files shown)


Libraries/LibJS/Interpreter.cpp
```diff
@@ -55,6 +56,7 @@ Interpreter::Interpreter()
     m_error_prototype = heap().allocate<ErrorPrototype>();
     m_date_prototype = heap().allocate<DatePrototype>();
     m_number_prototype = heap().allocate<NumberPrototype>();
+    m_boolean_prototype = heap().allocate<BooleanPrototype>();
```
bgianfo, line 59: You need to track this in: Interpreter::gather_roots

Libraries/LibJS/Runtime/BooleanPrototype.cpp
```diff
@@ -0,0 +1,70 @@
...
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include <AK/Function.h>
```
bgianfo, line 27: Did I miss where this is used?

Libraries/LibJS/Runtime/BooleanObject.h
```diff
@@ -0,0 +1,47 @@
...
+    explicit BooleanObject(bool);
+    virtual ~BooleanObject() override;
+
+    virtual Value value_of() const override
+    {
+        return Value(m_value);
+    }
+
+private:
+    virtual const char* class_name() const override { return "Boolean"; }
```
bgianfo, line 43: Since the c++ type name is BooleanObject, should this return "BooleanObject" like GlobalObject, ConsoleObject, StringObject, etc do?

Libraries/LibJS/Runtime/BooleanPrototype.cpp
jack-karamanian: It's forward declared in the other headers and needs to be included here.

Libraries/LibJS/Interpreter.cpp
jack-karamanian: Thanks!

awesomekling:
Neat :)

PR #1684, opened 2020-04-07
LibWeb: Add canvas.strokeRect(), and fix scale() & translate()

bgianfo:
Add an implementation of CanvasRenderingContext2DWrapper.strokeRect().
While implementing this I fixed fillRect() and the new strokeRect() to
honour the .scale() and .translate() values that had previously been plumbed.

Also enhance the canvas.html demo to utilize strokeRect(), scale(), and translate().

screenshot

### DIFF
--- a/Libraries/LibWeb/DOM/CanvasRenderingContext2D.cpp
+++ b/Libraries/LibWeb/DOM/CanvasRenderingContext2D.cpp
@@ -56,8 +56,29 @@ void CanvasRenderingContext2D::fill_rect(int x, int y, int width, int height)
     if (!painter)
         return;
 
-    Gfx::Rect rect(x, y, width, height);
+    Gfx::Rect rect = compute_rect(x, y, width, height);
     painter->fill_rect(rect, m_fill_style);
+    did_draw(rect);
+}
+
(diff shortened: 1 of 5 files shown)


Libraries/LibWeb/DOM/CanvasRenderingContext2D.cpp
```diff
@@ -77,6 +98,14 @@ void CanvasRenderingContext2D::translate(double x, double y)
     m_translate_y = y;
 }
 
+Gfx::Rect CanvasRenderingContext2D::compute_rect(int x, int y, int width, int height)
+{
+    return Gfx::Rect((x + m_translate_x) * m_scale_x,
+                     (y + m_translate_y) * m_scale_y,
+                     width * m_scale_x,
+                     height * m_scale_y);
```
awesomekling, line 106: Transforms are supposed to accumulate, so I we're gonna need to keep an affine transformation matrix and update it on calls to `scale()` and `translate()`. It will also allow us to implement `transform()` correctly. :)

awesomekling: Actually let's just merge this as-is and we can add transforms separately.

PR #1694, opened 2020-04-08
IRCClient: Add application and context menu items to hop/dehop users

bcoles:
Add application and context menu items to hop (+h) / dehop (-h) users.

Also renames `/HOP` command to `/CYCLE`. While some clients use HOP to join/part a channel, this is kind of confusing, and the general convention is to use `CYCLE`.

### DIFF
--- a/Applications/IRCClient/IRCWindow.cpp
+++ b/Applications/IRCClient/IRCWindow.cpp
@@ -111,6 +111,26 @@ IRCWindow::IRCWindow(IRCClient& client, void* owner, Type type, const String& na
                 m_client.handle_devoice_user_action(m_name.characters(), nick.characters());
             }));
 
+            m_context_menu->add_action(GUI::Action::create("Hop", [&](const GUI::Action&) {
+                GUI::ModelIndex new_index = member_view.selection().first();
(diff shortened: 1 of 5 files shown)


Applications/IRCClient/IRCWindow.cpp
```diff
@@ -111,6 +111,26 @@ IRCWindow::IRCWindow(IRCClient& client, void* owner, Type type, const String& na
                 m_client.handle_devoice_user_action(m_name.characters(), nick.characters());
             }));
 
+            m_context_menu->add_action(GUI::Action::create("Hop", [&](const GUI::Action&) {
+                GUI::ModelIndex new_index = member_view.selection().first();
+                auto nick = member_view.model()->data(new_index, IRCChannelMemberListModel::Role::Display).to_string();
```
awesomekling, line 116: This logic could be moved to a helper function since it's repeated a bunch of times :)

bcoles: Yes, yes it could. Seems like that should be a separate PR...

awesomekling: Or just a separate commit before this one

Issue #1681, opened 2020-04-07
LibJS: Variables aren't hoisted


ojhunt:
Variables are not being hoisted 

```
function f() {
   console.log(a); // should log undefined
   var a;
}
```

awesomekling:
Fixed with ac7459cb400125c952255da46ec5b2bb98d5d711

PR #1700, opened 2020-04-08
LibJS: Implement variable declaration hoisting

jgbyrne:
A naive attempt at implementing var declaration hoisting.
Iterates over the children of `ScopeNode`s when entered to look
for var declarations and declares them in the same manner as args.

### DIFF
--- a/Libraries/LibJS/AST.h
+++ b/Libraries/LibJS/AST.h
@@ -632,6 +632,8 @@ public:
     virtual Value execute(Interpreter&) const override;
     virtual void dump(int indent) const override;
 
+    const NonnullRefPtrVector<VariableDeclarator>& declarations() const { return m_declarations; }
+
 private:
     virtual const char* class_name() const override { return "VariableDeclaration"; }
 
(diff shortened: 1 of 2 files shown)


jgbyrne:
N.B. this is my first time writing C++, so apologies in advance if the code is unidiomatic or otherwise sub-par :-)

awesomekling:
Hello @jgbyrne and thanks for working on the project :)

Unfortunately this approach will only work for immediate children of the `ScopeNode`, it would need to recursively traverse the entire subtree to catch all the declarations.

However, we should really be doing this in the parser instead. I think a good approach would be to keep track of the nearest `var` scope that we're currently parsing, and append all `var` declarations to a list in the containing scope. That way, we don't have to worry about this at runtime at all, all the information would already be there :)

The issue number for this is #1681 btw
cc @sunverwerth @linusg @ojhunt

jgbyrne:
I had a feeling I was being too naive :-)
I will have a poke around in the parser code and see if it's something I can manage...

PR #1701, opened 2020-04-08
IRCClient: UI improvements

bcoles:
* Add nick_without_prefix and nick_at helpers
* Open query on double click of nick in channel member list


awesomekling:
V nice! Man, the IRC client is changing fast :)

PR #1702, opened 2020-04-08
LibJS: Handle empty values in Value::to_string()

linusg:
The REPL calls `Value::to_string()` on ~~everything that's~~ most things that are being `print()`ed, and that fails with `ASSERT_NOT_REACHED()` for anything that contains an empty value.

Before:

```
$ ./js
> a = Array(5)
js: serenity/Libraries/LibJS/Runtime/Value.cpp:78: AK::String JS::Value::to_string() const: Assertion `false' failed.
Aborted (core dumped)
$
```

After:

```
$ ./js
> new Array(5)
[ <empty>, <empty>, <empty>, <empty>, <empty> ]
> 
```

image

image

### DIFF
--- a/Userland/js.cpp
+++ b/Userland/js.cpp
@@ -161,6 +161,11 @@ static void print_error(const JS::Object& object, HashTable<JS::Object*>&)
 
 void print_value(JS::Value value, HashTable<JS::Object*>& seen_objects)
 {
+    if (value.is_empty()) {
+        printf("\033[34;1m<empty>\033[0m");
+        return;
+    }
+        
     if (value.is_object()) {
         if (seen_objects.contains(&value.as_object())) {
             // FIXME: Maybe we should only do this for circular references,


awesomekling:
We should never call `to_string()` on an empty `Value` in the first place.
Otherwise we risk not catching ourselves exposing the state to JavaScript which would be invalid.
Instead, let's just handle it in the REPL only.

linusg:
Fair point!

awesomekling:
:+1:

linusg:
@awesomekling I'm afraid this issue still exists to some extend:

```
$ ./js
> a = new Array(5)
[ <empty>, <empty>, <empty>, <empty>, <empty> ]
> console.log(a)
js: serenity/Libraries/LibJS/Runtime/Value.cpp:78: AK::String JS::Value::to_string() const: Assertion `false' failed.
Aborted (core dumped)
$
```

https://github.com/SerenityOS/serenity/blob/cd1d369cdd86b4870d65bc9ad856e8a0850cb4fb/Libraries/LibJS/Runtime/ConsoleObject.cpp#L44-L53

How should we deal with this one?

awesomekling:
That one needs to be fixed in `ArrayPrototype::to_string()` :)

PR #1706, opened 2020-04-08
LibJS: Add initial support for Promises

linusg:
I figured I'd start working on Promises - still WIP. Not aiming for the full Promise API in this PR, but at least `then()`, `catch()`, `finally()` should be working.

The code, including some function names is closely modelled after the spec: https://tc39.es/ecma262/#sec-promise-objects

~~I'll probably put a checklist here and start writing tests tomorrow to see what the actual limitations currently are.~~ Here it is:

- [x] `Promise()`
- [x] `new Promise()`
- [x] `Promise.prototype.then()`
  - [x] `onResolve`
  - [x] `onReject`
- [x] `Promise.prototype.catch()` ~~(there's a catch though, pun intended - the parser doesn't understand `new Promise(...).catch(...)`, so we have to use `new Promise(...)["catch"](...)` for now)~~
- [x] `Promise.prototype.finally()`

Not sure about the non-prototype functions yet, might be out of scope for this PR:

- [ ] `Promise.all()`
- [ ] `Promise.allSettled()`
- [ ] `Promise.race()`
- [x] `Promise.reject()` (should be relatively easy)
- [x] `Promise.resolve()` (should be relatively easy)

---

_The following is outdated, see further down in this PR for updates._

---

It can run this for example:

```js
new Promise((resolve, reject) => resolve("foo"))
    .then(value => {
        console.log("then() received " + value);
        return value + "bar";
    })
    .then(value => {
        console.log("then() received " + value);
    });
```

image

Browser console in Firefox for comparison:

image

Or this, with delayed resolving (which is usually the case):

image

Again, Firefox for comparison:

image


Libraries/LibJS/Runtime/PromisePrototype.h
```diff
@@ -0,0 +1,46 @@
...
+class PromisePrototype final : public Object {
+public:
+    PromisePrototype();
+    virtual ~PromisePrototype() override;
+
+private:
+    virtual const char* class_name() const override { return "PromisePrototype"; }
+
+    static Value then(Interpreter&);
+    static Value catch_(Interpreter&);
```
linusg, line 42: In Python it is common to append an underscore to identifiers if they are reserved keywords. Is there a convention for this in the C++ world? Just renaming is of course an option, but I think in this case these should match the method names of `Promise.prototype`...

awesomekling: I don't think there's one true convention. this_ works for me :)

Libraries/LibJS/Runtime/Promise.h
```diff
@@ -0,0 +1,117 @@
...
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#pragma once
+
+#include <LibJS/Runtime/Object.h>
+
+namespace JS {
+
```
emanuele6, line 32: I think you added an extra blank line by mistake here.

linusg: indeed - fixed :)

linusg:
This doesn't deserve special treatment, so "Closing this for lack of activity. Feel free to reopen if/when progress is made :)"

linusg:
Alright... almost exactly one year later, I'm happy to say this is finally done - see checklist in PR description / commit messages for details. Rebased to use the latest and greatest LibJS APIs, refactored/rewrote about half of it, added a bunch of tests, made it work in LibWeb, fixed all the bugs I know of.

Enjoy this (Lagom `js`) screenshot showcasing some basic functionality:

image

---

<sup>I *promise* this is not an April Fools' joke.</sup>

Userland/Libraries/LibJS/Runtime/VM.cpp
```diff
@@ -371,4 +381,44 @@ bool VM::in_strict_mode() const
...
+        auto* job = m_promise_jobs.take_first();
+        dbgln_if(PROMISE_DEBUG, "Calling promise job function @ {}", job);
+        [[maybe_unused]] auto result = call(*job, js_undefined());
+    }
+    // Ensure no job has created a new exception, they must clean up after themselves.
+    VERIFY(!m_exception);
+}
+
+// 9.4.4 HostEnqueuePromiseJob, https://tc39.es/ecma262/#sec-hostenqueuepromisejob
+void VM::enqueue_promise_job(NativeFunction& job)
```
Lubrsi, line 400: I don't know if this should be done now, but this needs to somehow be overridable in the future for integration with HTML event loops:
https://html.spec.whatwg.org/#hostenqueuepromisejob

linusg: Definitely one for later - from what I've seen LibWeb isn't driven by an event loop at all (yet), e.g. timers just run their callbacks right away instead of queuing a task, so I just embraced that and did this: 192f3a344477d6f09e42c533bbd7a92d50f30fa8

Since nothing would want to subclass `JS::VM` (probably), we might just add a `on_something` function and call that if it exists, leaving queueing and therefore running the job to whatever owns the VM.

PR #1707, opened 2020-04-08
Close combo box on mouseclick outside the window frame

### LINKED ISSUE #1493: LibGUI: ComboBox in GUI::Dialog rendering over windows in foreground
Screenshot_2020-03-20_12-37-18

rhin123:
I'm not sure if this is the proper way to implement this, but I think it makes some sense.
Fixes #1493 from happening


rhin123:
Yeah this isnt right, I'll try this again 😦

PR #1708, opened 2020-04-09
LibLine: Fix multiline editing

alimpfard:
This patchset fixes multiline editing via LibLine.

As a little bonus, it also allows changing the prompt in `on_display_refresh`.
and as a result of this bonus, the js repl can indent properly :tada:

Here's a demo (Lagom version), since building these PRs is annoying:
shot-2020-04-09_08-16-27

what it breaks (because NoFreeLunch):
Now the cursor would sometimes appear to be jumping to the beginning of the line.

I believe somehow implementing hiding of the cursor (`CSI?25l`) would also hide this issue.


alimpfard:
seems like this breaks a bit of functionality, `await fix()`

PR #1711, opened 2020-04-09
LibJS: Add globalThis

linusg:
We already have "`global`" as a (non-standard) way to access the global object in `js(1)` (both REPL and script mode). This replaces it with "`globalThis`", which is available in all environments, not just `js`.

`globalThis` is relatively new, but supported by all modern browsers/engines.

- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/globalThis
- https://tc39.es/ecma262/#sec-globalthis


awesomekling:
Great idea :)

PR #1712, opened 2020-04-09
IRCClient: Set nick and userinfo to OS username when not set in config

bcoles:
As a multi-user OS, it makes sense to use the OS username for default configuration options when there's no existing configuration file (or the configuration file is incomplete).

This also allows for auto-populating a new config file with reasonable default values when a new user runs the IRCClient for the first time and does not have an existing config file.

Serenity currently builds with a `IRCClient.ini` configuration file for the `anon` user which contains default options.

An alternative approach may be to set the `nick` to the OS username, then let the other fields such as `userinfo` and `finger` CTCP replies, default to whatever is set as the `nick`.

### DIFF
--- a/Applications/IRCClient/IRCClient.cpp
+++ b/Applications/IRCClient/IRCClient.cpp
@@ -39,6 +39,7 @@
 #include <netinet/in.h>
 #include <stdio.h>
 #include <sys/socket.h>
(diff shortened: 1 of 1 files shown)


awesomekling:
This seems reasonable :)

PR #1714, opened 2020-04-09
IRCClient: Autojoin channels after client registration

### LINKED ISSUE #1713: IRCClient: Channel autojoin should not immediately follow client registration
The IRC protocol requires the client to provide a valid `USER` and `NICK` to complete client registration.

IRC servers do not accept commands prior to completion of registration.

Sometimes the negotiation can be delayed. ie, when the server tries and fails to a hostname lookup.

This causes channel autojoin to fail, as the client rudely sends `NICK`, `USER`, and `JOIN #...` all at once.

451 JOIN

bcoles:
Wait for the server to advise negotiation was successful (RPL_WELCOME)
before autojoining channels.

Fixes #1713

RPL_WELCOME

### DIFF
--- a/Applications/IRCClient/IRCClient.cpp
+++ b/Applications/IRCClient/IRCClient.cpp
@@ -46,6 +46,7 @@
 #define IRC_DEBUG
 
 enum IRCNumeric {
+    RPL_WELCOME = 1,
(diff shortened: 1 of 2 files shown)


awesomekling:
Very polite!

PR #1722, opened 2020-04-09
LibJS: Throw exception if LHS of assignment is of unexpected type

DontBelieveMe:
If the left hand side of an assignment is not a valid type then throw an error instead of just crashing.
Throwing a ReferenceError is consistent with the behavior of other browsers (see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors/Invalid_assignment_left-hand_side)


awesomekling:
Could you also add a test in LibJS/Tests/ that hits this code path? :)

DontBelieveMe:
Dang , I knew I was forgetting something! :smile: I'll do that now

awesomekling:
Great! Thank you :)

PR #1724, opened 2020-04-10
LibGUI: Add ColorInput component and small fixes at DisplayProperties

asliturk:
`ColorInput` simplify usage of `ColorPicker` dialog and more elegant view of current selected color in input field.

ColorInput


awesomekling:
Great idea :)

PR #1725, opened 2020-04-10
Kernel: Add explicit offset parameter to File::read etc

deoxxa:


awesomekling:
I love how this makes `DiskPartition::read` and `DiskPartition::write` not look like a total shitshow :)

deoxxa:
Magic!

PR #1727, opened 2020-04-10
IRCClient: Add ShowNickChangeMessages/ShowJoinPartMessages conf options

bcoles:
Allow IRCClient to hide nick change spam and join/part spam.

`ConfigFile::read_bool_entry` uses `0` and `1` for Boolean values.

```cpp
bool ConfigFile::read_bool_entry(const String& group, const String& key, bool default_value) const
{
    return read_entry(group, key, default_value ? "1" : "0") == "1";
}
```

As such, the `IRCClient.ini` configuration file also uses `0` and `1`.


awesomekling:
It would seem reasonable for `Core::ConfigFile` to accept `true`/`false` as well :)

PR #1731, opened 2020-04-10
LibJS: Add String.prototype.pad{Start,End}()

linusg:


Libraries/LibJS/Runtime/StringPrototype.cpp
```diff
@@ -182,4 +184,55 @@ Value StringPrototype::to_string(Interpreter& interpreter)
...
+    if (!interpreter.argument(1).is_undefined())
+        fill_string = interpreter.argument(1).to_string();
+    if (fill_string.is_empty())
+        return js_string(interpreter, string);
+
+    auto fill_length = max_length - string.length();
+
+    String filler = "";
+    while (filler.length() < fill_length)
+        filler = String::format("%s%s", filler.characters(), fill_string.characters());
```
awesomekling, line 214: Let's use a `StringBuilder` here instead of repeatedly creating new `String` objects.

linusg: That's much better... Still making myself familiar with all the cool things in `AK` :blush:

PR #1734, opened 2020-04-10
Misc Kernel fixes

supercomputer7:


Kernel/PCI/Definitions.h
```diff
@@ -172,8 +172,24 @@ struct ChangeableAddress : public Address {
     }
 };
 
+class PhysicalID : public RefCounted<PhysicalID> {
```
awesomekling, line 175: This can just be a value type, no need to use reference counting :)

PR #1736, opened 2020-04-10
Fix non-configurable js attribute test

4ourbit:
I think this needs an extra assert.

### DIFF
--- a/Libraries/LibJS/Tests/Object.defineProperty.js
+++ b/Libraries/LibJS/Tests/Object.defineProperty.js
@@ -24,11 +24,14 @@ try {
     assert(d.writable === true);
     assert(d.value === "ho");
 
+    var exceptionThrown = false;
     try {
         Object.defineProperty(o, "bar", { value: "xx", enumerable: false });
     } catch (e) {
         assert(e.name === "TypeError");
+        exceptionThrown = true;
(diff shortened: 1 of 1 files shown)


linusg:
This feels a bit awkward imo - why not something like this instead?

```js
try {
    Object.defineProperty(o, "bar", { value: "xx", enumerable: false });
    assert(false);
} catch (e) {
    assert(e.name === "TypeError");
}
```

We could also add an alias for `assert(false)` like we have it in C++:

```js
const assertNotReached = () => assert(false);
```

and then do:

```js
try {
    Object.defineProperty(o, "bar", { value: "xx", enumerable: false });
    assertNotReached();
} catch (e) {
    assert(e.name === "TypeError");
}
```

supercomputer7:
I suggest to change the commit title to be more like "Foo: Bar". Also, please avoid using the word "fix" in the commit message, as it is ambiguous what is the change that you applied in the commit :)

4ourbit:
Could you squash before merge?

supercomputer7:
> Could you squash before merge?

I'd suggest to squash-rebase your commits by yourself :)

awesomekling:
LibJS: Improve Object.defineProperty test

PR #1737, opened 2020-04-10
MenuApplets: Rename CPUGraph to ResourceGraph and support multiple instances

linusg:
First, a screenshot:

Screenshot from 2020-04-10 23-50-01

TL;DR: CPUGraph is now ResourceGraph and supports multiple instances, a memory graph and custom colors. The next useful addition would probably be a network traffic graph, but something like one graph per CPU core is also a possibility.

I'm lazy so I'll just dump my commit messages here:

---

The plan is to extend what currently is known as "CPUGraph" and let the SystemServer spawn multiple instances of it - which then can show memory or network usages as well :^)

Simply renaming the applet is the first step.

---

The ResourceGraph menu applet now supports a few command line options:

- `--cpu` / `-C` to display a CPU usage graph
- `--memory` / `-M` to display a memory usage graph
- `--name` / `-n` to set a name which is used to order applets
- `--color` / `-c` to set the graph color (supports anything `Gfx::Color::from_string()` understands)

The `SystemServer.ini` and `WindowServer.ini` config files have been updated to spawn and show two ResourceGraph menu applets, one for CPU usage (green) and one for memory usage (cyan) - this matches the colors in the SystemMonitor graphs.

### DIFF
--- a/Base/etc/SystemServer.ini
+++ b/Base/etc/SystemServer.ini
@@ -48,7 +48,15 @@ User=anon
 KeepAlive=1
 User=anon
 
-[ResourceGraph.MenuApplet]
+[CPUGraph.MenuApplet]
+Executable=/bin/ResourceGraph.MenuApplet
+Arguments=--cpu --name=CPUGraph --color=limegreen
+KeepAlive=1
+User=anon
+
+[MemoryGraph.MenuApplet]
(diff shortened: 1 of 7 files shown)


awesomekling:
Cool idea :D

PR #1740, opened 2020-04-11
IRCClient: Add channel member context menus for common CTCP requests

bcoles:
Add menu items for CTCP: USERINFO, FINGER, TIME, VERSION, CLIENTINFO

Eventually these actions should be placed under a `CTCP` sub menu. Likewise, the `Control` actions (op, deop, kick, etc) should be placed under a `Control` sub menu.


awesomekling:
:+1:

PR #1741, opened 2020-04-11
LibLine+JS repl: completion fixup, primitive completion

alimpfard:
name says all.

I think LibLine is quickly becoming my "escape from implementing LibCrypto/TLS" place...


awesomekling:
Oh that's really neat!

alimpfard:
You typed `Object.defineProperty` so much in your last browser hacking video, I thought to myself "man, it would be pretty cool to use this as an excuse to stop reading this RFC for a few minutes".

PR #1742, opened 2020-04-11
LibJS: Make Function and CallFrame aware of their function name / Add console.trace()

linusg:
The first commit prepares `Function` (and thus `ScriptFunction` and `NativeFunction`), as well as `CallFrame` to store their own name / name of the currently executed function.

This also improves `Function.prototype.toString()` which now includes the function name for both `ScriptFunction`s and `NativeFunction`s:

image

The second commit adds `console.trace()` (https://developer.mozilla.org/en-US/docs/Web/API/Console/trace):

image


awesomekling:
Very tasty :)

PR #1743, opened 2020-04-11
 js: Stylize TokenType::{Throw,Switch,Case}

linusg:


linusg:
I did so much LibJS stuff that my brain now starts all commit messages with `LibJS: ` automatically :man_facepalming:

Libraries/LibLine/Editor.cpp
```diff
@@ -43,7 +43,6 @@ Editor::Editor()
     } else {
         m_num_columns = ws.ws_col;
         m_num_lines = ws.ws_row;
-        dbg() << m_num_lines;
```
alimpfard, line 46: ouch, I removed this in my PR too...

linusg: I'll rebase, no need to do it in a separate commit then :+1:

PR #1745, opened 2020-04-11
Debugger Application

itamar8910:
This PR adds a debugger application, called `sdb`.

A screenshot: 
sdb

When starting a program with `sdb`, we automatically insert a breakpoint
at the entry point.
Once the debugged program is paused, the user can issue commands.
The current commands are:
- Continue the debugged program.
- Disassemble a section of the program, starting from the current instruction pointer. (Using the fresh addition of `LibX86`).
- Print the register state of the stopped program.
- Insert a new breakpoint.

The breakpoints are "Software Breakpoints":
To break at an instruction, we patch the first byte to the `int3` instruction, which generates a breakpoint trap and allows the debugger to take over.
When the breakpoint is tripped, the debugger restores the instruction it patched.

I also implemented some additional ptrace() functionality required for the debugger: PT_SETREGS, PT_PEEK and PT_POKE.


awesomekling:
This looks very kickass!

Two main issues:
- Please wrap your commit messages at 72 characters per line instead of ~50 :)
- Please make sure that `ptrace` can't access the registers of a process while it's running inside the kernel.

zecke:
I just watched https://systemswe.love/videos/rogers-proc. Do you want to consider something more modern than ptrace?

Kernel/Ptrace.cpp
```diff
@@ -0,0 +1,165 @@
...
+
+    Thread* peer = nullptr;
+    {
+        InterruptDisabler disabler;
+        peer = Thread::from_tid(params.pid);
+    }
+    if (!peer)
+        return KResult(-ESRCH);
+
+    if (peer->process().uid() != caller.euid())
```
itamar8910, line 31: Thinking about it, we should probably also disallow tracing `setuid` processes.
PT_POKE & PT_SETREGS could be used in a crafty way for privilege escalation!

Applications/Debugger/main.cpp
```diff
@@ -0,0 +1,235 @@
...
+    auto path = String::format("/proc/%d/exe", pid);
+    dbg() << "path: " << path;
+    auto file = Core::File::construct(path);
+    if (!file->open(Core::File::ReadOnly)) {
+        fprintf(stderr, "Failed to open Debugged executable");
+        exit(1);
+    }
+    auto data = file->read_all();
+    dbg() << "data size:" << data.size();
+    ELFImage elf(data.data(), data.size());
```
ADKaster, line 71: I broke this in #1750 but it should be easy enough to change to `#include <LibELF/Image>` and construct an `ELF:: Image` instead.

Side note for this part: might be faster to just mmap the file instead of opening it and reading it all into a heap buffer? I absolutely don't have the numbers to back that up though. Seems a waste to read the entire executable here and then throw it away...

ADKaster: Alternatively, the start address should be available in the first page for sure. First 64 bytes even, as the elf header structure at the beginning of the file looks like this:
```c++
/* ELF Header */
typedef struct elfhdr {
    unsigned char e_ident[EI_NIDENT]; /* ELF Identification */
    Elf32_Half e_type;                /* object file type */
    Elf32_Half e_machine;             /* machine */
    Elf32_Word e_version;             /* object file version */
    Elf32_Addr e_entry; /** <--- here **/  /* virtual entry point */ 
    Elf32_Off e_phoff;                /* program header table offset */
...
} Elf32_Ehdr;
```
The debugger will need to be more clever about treating the virtual address in the elf header of the entry point as gospel in the future, because any ASLR or relocation that gets applied to the main executable will need to be added/accounted for.

The load address should be possible to calculate from /proc/[pid]/maps though?

itamar8910: > Side note for this part: might be faster to just mmap the file instead of opening it and reading it all into a heap buffer? I absolutely don't have the numbers to back that up though. Seems a waste to read the entire executable here and then throw it away...

Absolutely! Looks like `MappedFile` should be great for this. 

> The load address should be possible to calculate from /proc/[pid]/maps though?

Yeah, should be possible. IIRC, gdb disables ASLR by default, so maybe we could just do that as well.

itamar8910:
> I just watched https://systemswe.love/videos/rogers-proc. Do you want to consider something more modern than ptrace?

Exposing the address space in `/proc` and manipulating it with regular file operations really is neat! 

I think we can start with a hybrid between `ptrace` and `/proc` - I find `ptrace` to be pretty convenient for _some_ of the tracing operations.
Might be out of the scope of this pr, though? cc @awesomekling

awesomekling:
It's an interesting concept but not a must-have to get a debugger going. :)

itamar8910:
>    Please make sure that `ptrace` can't access the registers of a process while it's running inside the kernel.

I added two diffs to fix this:

- The kernel hangs if there's a breakpoint trap in ring0 (we dont have kernel debugging.. yet!)
- When copying the userspace registers from the stack, we copy `userspace_esp` into `esp`, instead of `esp` which points at the kernel's stack.

awesomekling:
What about `PT_SETREGS` and `PT_GETREGS` when the tracee is running code inside the kernel (i.e `Thread::in_kernel() == true`)?
We don't want a userspace process calling `ptrace()` to be able to see (or modify!) tracee register state while it's executing in ring 0.

Kernel/Ptrace.cpp
```diff
@@ -0,0 +1,153 @@
...
+        if (!tracer->has_regs())
+            return KResult(-EINVAL);
+
+        if (!caller.validate_read(params.addr, sizeof(PtraceRegisters)))
+            return KResult(-EFAULT);
+
+        {
+            SmapDisabler disabler;
+            PtraceRegisters* regs = reinterpret_cast<PtraceRegisters*>(params.addr);
+            RegisterState* peer_saved_registers = reinterpret_cast<RegisterState*>(peer->kernel_stack_top() - sizeof(RegisterState));
```
awesomekling, line 97: The somewhat awkwardly named `RegisterState& Thread::get_register_dump_from_stack()` can get you a reference to this place :)

awesomekling:
> What about `PT_SETREGS` and `PT_GETREGS` when the tracee is running code inside the kernel (i.e `Thread::in_kernel() == true`)?
> We don't want a userspace process calling `ptrace()` to be able to see (or modify!) tracee register state while it's executing in ring 0.

Oh never mind, we're doing this through the register state dump from the kernel entry. That's good :)

Kernel/Ptrace.cpp
```diff
@@ -0,0 +1,171 @@
...
+
+    case PT_SETREGS: {
+        if (!tracer->has_regs())
+            return KResult(-EINVAL);
+
+        if (!caller.validate_read(params.addr, sizeof(PtraceRegisters)))
+            return KResult(-EFAULT);
+
+        auto& peer_saved_registers = peer->get_register_dump_from_stack();
+        // Verify that the saved registers are in usermode context
```
itamar8910, line 94: I added this check, just to be on the safe side.

Kernel/Ptrace.cpp
```diff
@@ -0,0 +1,171 @@
...
+        if ((peer_saved_registers.cs & 0x03) != 3) {
```
itamar8910, line 95: I added this check, just to be on the safe side.

awesomekling:
Ok! Looks like a great start :)

PR #1746, opened 2020-04-11
LibLine: fix DSR spam; suggestion fix and beautification

alimpfard:
Fixes an issue @awesomekling noticed (with pasting into the buffer)
Also fixes an issue @linusg mentioned and builds up on it.

As screenshots are quite trendy today,
shot-2020-04-11_19-34-10

### DIFF
--- a/Libraries/LibLine/Editor.cpp
+++ b/Libraries/LibLine/Editor.cpp
@@ -43,7 +43,6 @@ Editor::Editor()
     } else {
         m_num_columns = ws.ws_col;
         m_num_lines = ws.ws_row;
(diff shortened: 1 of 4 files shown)


alimpfard:
A little problem with the current implementation is that it will not show __all__ of the returned suggestions, just the ones that fit on the screen.
However it will still cycle through them, so this causes a visual-behavioural desync...not sure if I love that.

awesomekling:
It's a decent compromise I guess :)

alimpfard:
I dislike avoidable compromises.
the question is, how much complexity are we willing to introduce into this line editor before it's too complex?

PR #1748, opened 2020-04-11
LibELF: Return false instead of assert on unrecognized program header

### LINKED ISSUE #1519: ELFImage::validate_program_headers: Found program header (6) of unrecognized type 1685382480!
Attempting to execute a malformed executable freezes the system.

For example, executing a program incorrectly compiled as a ~64-bit shared object~ ELF 32-bit LSB shared object:

```
Found program header (6) of unrecognized type 1685382480!
[Shell(20:20)]: ASSERTION FAILED: false
../Libraries/LibELF/ELFImage.cpp:411 in static bool ELFImage::validate_program_headers(const Elf32_Ehdr&, size_t, u8*, size_t, AK::String&)
[Shell(20:20)]: 0xc01184cb  __assertion_failed(char const*, char const*, unsigned int, char const*) +180
[Shell(20:20)]: 0xc0112e25  ELFImage::validate_program_headers(elfhdr const&, unsigned long, unsigned char*, unsigned long, AK::String&) +365
[Shell(20:20)]: 0xc0153f52  Kernel::Process::find_elf_interpreter_for_executable(AK::String const&, char (&) [4096], int, unsigned long) +256
[Shell(20:20)]: 0xc0155872  Kernel::Process::exec(AK::String, AK::Vector<AK::String, 0ul>, AK::Vector<AK::String, 0ul>, int) +1408
[Shell(20:20)]: 0xc0157d8e  Kernel::Process::sys$execve(Kernel::Syscall::SC_execve_params const*) +480
[Shell(20:20)]: 0xc01662c8  syscall_handler +1130
[Shell(20:20)]: 0xc0165d99  syscall_asm_entry +29
[Shell(20:20)]: 0x08070137  execvpe +287
[Shell(20:20)]: 0x08070348  execvp +28
[Shell(20:20)]: 0x0804e07b  run_command(AK::String const&) [clone .part.0] +6898
[Shell(20:20)]: 0x0804e524  run_command(AK::String const&) +72
[Shell(20:20)]: 0x080489e4  main +1430
[Shell(20:20)]: 0x08048bc3  _start +94
```

ADKaster:
No reason to assert here when the only callers (currently Process::find_elf_interpreter_for_executable) check the return value. Also, change the output to hex, since that will make it easier to google and figure out which program header the compiler generated that we don't know how to handle.

Fixes #1519

### DIFF
--- a/Libraries/LibELF/ELFImage.cpp
+++ b/Libraries/LibELF/ELFImage.cpp
@@ -407,9 +407,8 @@ bool ELFImage::validate_program_headers(const Elf32_Ehdr& elf_header, size_t fil
             break;
         default:
             // Not handling other program header types in other code so... let's not surprise them
-            dbgprintf("Found program header (%d) of unrecognized type %d!\n", header_index, program_header.p_type);
-            ASSERT_NOT_REACHED();
-            break;
+            dbgprintf("Found program header (%d) of unrecognized type %x!\n", header_index, program_header.p_type);
+            return false;
         }
     }
     return true;


awesomekling:
fwiw googling "1685382480" works perfect ;)

PR #1749, opened 2020-04-11
LibLine: Cycle backward through suggestions using Shift+Tab

linusg:


Libraries/LibLine/Editor.cpp
```diff
@@ -406,7 +437,7 @@ String Editor::get_line(const String& prompt)
                 m_refresh_needed = true;
                 continue;
             }
-            if (ch == 0xc) { // ^L
+            if (ch == 0xc) {                    // ^L
```
linusg, line 440: clang-format did this.

awesomekling:
Ok :)

PR #1750, opened 2020-04-11
LibELF: Move classes into their own namespace

ADKaster:
Move ELF classes into ELF namespace, to align better with other libraries. Also, move the ELF::Image::validate_* static methods into their own file to be free functions in the ELF namespace, and actually call them from ELF::DynamicLoader on the mapped elf file.


awesomekling:
Looks good to me! I accidentally a conflict, sorry about that :(

PR #1753, opened 2020-04-11
Kernel: store previous thread state upon all transitions to Stopped

### LINKED ISSUE #1752: Kernel: continuing a process stopped with SIGTSTP hangs system
## Steps to reproduce

1. Run command `sleep 20` in Shell
2. Hit Ctrl-Z to send SIGTSTP to `sleep` process
3. Open System Monitor, right click on `sleep` and click "Continue process"

### Expected result
`sleep` process transitions back from "Stopped" to "Sleeping"

### Actual result
System hangs, and an assertion failure is printed in debug console:

```
[SystemMonitor(23:23)]: ASSERTION FAILED: m_stop_state != State::Invalid
Thread.cpp:509 in Kernel::ShouldUnblockThread Kernel::Thread::dispatch_signal(u8)
```

peterdn:
We now store the previous thread state in m_stop_state for all
transitions to the Stopped state via Thread::set_state.

Fixes #1752 whereupon resuming a thread that was stopped with SIGTSTP,
the previous state of the thread is not remembered correctly, resulting
in m_stop_state == State::Invalid and the associated assertion fails.

### DIFF
--- a/Kernel/Thread.cpp
+++ b/Kernel/Thread.cpp
@@ -499,7 +499,6 @@ ShouldUnblockThread Thread::dispatch_signal(u8 signal)
     if (signal == SIGSTOP) {
         if (!is_stopped()) {
             m_stop_signal = SIGSTOP;
-            m_stop_state = m_state;
             set_state(State::Stopped);
         }
         return ShouldUnblockThread::No;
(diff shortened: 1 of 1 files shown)


awesomekling:
Oh, that's a nice catch! Thank you :)

PR #1754, opened 2020-04-11
 AK: Support fragment in URL / Parse query and fragment in URL::parse()

linusg:
This is going to be useful for the Browser :^)


awesomekling:
:+1:

PR #1758, opened 2020-04-12
Userspace: TaskBar with scroll

nimelehin:
TaskBar

### DIFF
--- a/Libraries/LibGUI/AbstractStackView.cpp
+++ b/Libraries/LibGUI/AbstractStackView.cpp
@@ -1,0 +1,84 @@
+/*
+ * Copyright (c) 2018-2020, Andreas Kling <kling@serenityos.org>
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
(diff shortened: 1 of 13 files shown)


awesomekling:
I don't think this is a good UI. I would much prefer the classical thing where the buttons get smaller so that more of them can fit.

zlotny:
Agree with being a little off. If for some reason you have 100 things opened the stack would be so large that finding a window scrolling would be really annoying. 

However the StackView implementation is neat :) How about keeping that? Maybe put an example of the StackView implementation on WidgetGallery?

Libraries/LibGUI/AbstractStackView.cpp
```diff
@@ -0,0 +1,84 @@
+/*
+ * Copyright (c) 2018-2020, Andreas Kling <kling@serenityos.org>
```
zlotny, line 2: Also, those copyright headers show wrong information.

Pulgovisk:
I think the concept of rows would be better
image

awesomekling:
Closing this for lack of activity. Feel free to reopen if/when progress is made :)

nimelehin:
I'm sorry for the delays - exams(. @awesomekling what do you think about StackView. Is it a thing to keep?

PR #1761, opened 2020-04-12
Build: fix kernel compilation

nimelehin:
Before I have experienced `fatal error: cxxabi.h: No such file or directory `and `fatal error: initializer_list: No such file or directory` compiling kernel and all the whole project with ./makeall.sh. This fixes that.


awesomekling:
Those defines are specifically intended for use only when building the LibC and toolchain respectively.

I think your issue is that you're running an older toolchain with GCC 9.2.0 instead of the 9.3.0 that we started using on March 22nd (c47ef61ed81b878e6631105e32a3e10466a4b713) :)

PR #1763, opened 2020-04-12
LibJS: Add console.{debug,info,warn,error}()

linusg:
Moved argument stringification & printing logic into a separate function. `info`'s functionality is equal to `log` for now, but I added a bit of color to `debug`, `warn` & `error` :^)

image


awesomekling:
Very nice!

PR #1770, opened 2020-04-12
FileManager: Update directories_model after retrieving the current path

### LINKED ISSUE #1765: FileManager: In refresh_tree_view determine the current_path before updating the model
If you update the tree view directories model, it changes the root path of the directory view. 
During the model update it triggers an on_selection_change, which causes the directory view to open "/" . I'd guess this is not intended behavior? 
Nevertheless, this causes the FileManager to crash every time refresh_tree_view is called, because it uses the directory view to get the current folder (which is now always "/") to reselect the folder in the tree view.

I'm a newcomer and I guess this is more of a hotfix.

rizzrackguy:
This avoids the crash after every refresh_tree_view call (mkdir, paste,
delete folder). During the model update it clears selection, which
triggers an on_selection_change with "/", which causes the directory
view to open it. After that the directory view is used to determine
the current path, which is now incorrect, to reselect the current
folder in the tree view.

Fixes #1765

### DIFF
--- a/Applications/FileManager/main.cpp
+++ b/Applications/FileManager/main.cpp
@@ -138,8 +138,6 @@ int main(int argc, char** argv)
     };
 
     auto refresh_tree_view = [&] {
-        directories_model->update();
-
         auto current_path = directory_view.path();
 
         struct stat st;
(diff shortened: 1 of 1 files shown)


awesomekling:
This doesn't appear to be an issue anymore.

PR #1772, opened 2020-04-12
LibJS: Parse "this" as ThisExpression

sunverwerth:
Saw in today's video that "this" was parsed as an identifier which Esprima doesn't seem to agree with.


awesomekling:
Great fix :D

PR #1773, opened 2020-04-12
LibJS: Throw on assignment of an const variable

bgianfo:
Was stubbed out as an assert, should be handled with a runtime exception.

### DIFF
--- a/Libraries/LibJS/Interpreter.cpp
+++ b/Libraries/LibJS/Interpreter.cpp
@@ -145,8 +145,10 @@ void Interpreter::set_variable(const FlyString& name, Value value, bool first_as
 
         auto possible_match = scope.variables.get(name);
         if (possible_match.has_value()) {
-            if (!first_assignment && possible_match.value().declaration_kind == DeclarationKind::Const)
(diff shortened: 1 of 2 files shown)


Libraries/LibJS/Interpreter.cpp
```diff
@@ -146,7 +146,7 @@ void Interpreter::set_variable(const FlyString& name, Value value, bool first_as
         auto possible_match = scope.variables.get(name);
         if (possible_match.has_value()) {
             if (!first_assignment && possible_match.value().declaration_kind == DeclarationKind::Const)
-                ASSERT_NOT_REACHED();
+                throw_exception<TypeError>("Assignment to constant variable");
```
awesomekling, line 149: We need to return immediately after throwing the exception, or we'll still actually set the variable!
It would be good if the test would also cover that the variable didn't change after an exception was thrown :)

bgianfo: Yea, when ever I read the code I expect throw_exception to also throw a C++ exception. Trips me up every time. 

Fixed, thanks!

PR #1774, opened 2020-04-12
LibJS: Tweak FunctionPrototype::to_string and constructors

sunverwerth:
This fixes the libjs tests.

The output of FunctionPrototype::to_string is now more in line
with the output in Firefox. The builtin constructors have been
extended to include their function name in the output.


awesomekling:
Yup this makes sense :)

PR #1776, opened 2020-04-13
LibJS: Implement Error.prototype.name setter

bgianfo:
The MDN example for creating a custom error type in javascript uses:

    function CustomError(foo, message, fileName, lineNumber) {
        var instance = new Error(message, fileName, lineNumber);
        instance.name = 'CustomError';
        instance.foo = foo;
        Object.setPrototypeOf(instance, Object.getPrototypeOf(this));
        return instance;
    }

The name property on the Error prototype needs to be settable for
this to work properly.

### DIFF
--- a/Libraries/LibJS/Runtime/Error.h
+++ b/Libraries/LibJS/Runtime/Error.h
@@ -39,6 +39,8 @@ public:
     const FlyString& name() const { return m_name; }
     const String& message() const { return m_message; }
 
+    void set_name(FlyString& name) { m_name = name; }
+
 private:
     virtual bool is_error() const final { return true; }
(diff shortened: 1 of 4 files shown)


Libraries/LibJS/Runtime/Error.h
```diff
@@ -39,6 +39,8 @@ class Error : public Object {
     const FlyString& name() const { return m_name; }
     const String& message() const { return m_message; }
 
+    void set_name(FlyString& name) { m_name = name; }
```
awesomekling, line 42: Since we don't intend to mutate the `FlyString` itself here:

```suggestion
    void set_name(const FlyString& name) { m_name = name; }
```

PR #1782, opened 2020-04-13
js/LibJS: Add & use assertNotReached() in tests

linusg:
Three commits:

- Add `assertNotReached()` function in test mode
- Fix test files indentation (4 spaces)
- Use `assertNotReached()` in tests

[1] should be self-explanatory, [2] because it was a bit of a mess, sometimes 2 spaces, mostly 4 spaces, sometimes 8 spaces and sometimes tabs and spaces mixed :scream:
[3] to ensure the tests actually fail when we test some code that's expected to fail but only assert that in the `catch` block - what if the interpreter (wrongly) never throws an exception?!


awesomekling:
Great idea!

PR #1785, opened 2020-04-13
Meta: Add Arch Linux (and derivatives) build instructions

zlotny:
This patch adds a one liner to install all the dependency packages
needed to compile and run Serenity on Arch Linux, Manjaro, etc


awesomekling:
:+1:

PR #1786, opened 2020-04-13
Debugger: Additional functionality

itamar8910:
I Added the following:
- Insert a breakpoint at a function name
- The debugger prints the address & function it is stopped at
- Breakpoints now persist after being tripped.
  So if you put a breakpoint in a loop / in a function that's called multiple times,
  you do not have to re-insert it after each time it's tripped
- Single-stepping command
- Repeat the previous command when an empty command is entered

A fun thing you can do now is do a 'single-step' once, and then hold enter and see the program go :smile:

### DIFF
--- a/Applications/Debugger/main.cpp
+++ b/Applications/Debugger/main.cpp
@@ -27,6 +27,7 @@
 #include "DebugSession.h"
 #include <AK/Assertions.h>
 #include <AK/ByteBuffer.h>
+#include <AK/Demangle.h>
 #include <AK/LogStream.h>
 #include <AK/StringBuilder.h>
 #include <AK/kmalloc.h>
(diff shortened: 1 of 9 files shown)


Applications/Debugger/main.cpp
```diff
@@ -222,3 +226,19 @@ int main(int argc, char** argv)
...
+    virtual int foo() = 0;
+};
+class B : public A {
+
+public:
+    virtual ~B() override;
+    int f() { return 2; }
+};
+
+B::~B() {}
```
awesomekling: What the...

Applications/Debugger/main.cpp
```diff
@@ -238,19 +239,3 @@ int main(int argc, char** argv)
...
-    virtual int foo() = 0;
-};
-class B : public A {
-
-public:
-    virtual ~B() override;
-    int f() { return 2; }
-};
-
-B::~B() {}
```
awesomekling: Please squash this change into the previous commit instead of adding it separately.
There's no need to maintain history of temporary hacks during development :)

Kernel/Arch/i386/CPU.cpp
```diff
@@ -297,6 +297,26 @@ void page_fault_handler(RegisterState regs)
     }
 }
 
+EH_ENTRY_NO_CODE(1, debug);
+void debug_handler(RegisterState regs)
+{
+    clac();
+    if (!Process::current || Process::current->is_ring0()) {
```
awesomekling, line 304: `Process::is_ring0` here is a bit deceptive.
It only tells us whether the current process is a "ring 0 process" AKA a kernel process.
It doesn't look at the current value of the CS register.
As such, this is not enough protection against trapping in ring 0. We need to check `regs.cs` :)

Applications/Debugger/main.cpp
awesomekling: Sorry, I meant the previous commit *adding it*, not the previous commit in sequence :D
I should have been more clear, that's my bad!

itamar8910:
@awesomekling I messed up the commits, please don't review yet :)

PR #1787, opened 2020-04-13
js/LibJS: Move test assert functions to pure javascript

bgianfo:
The addition of assert functions to Userland/js
was done before we had load(..) implemented. Now
that it exists, it seems like the right move the
test helper functions to pure javascript instead
of poluting js with random global functions.

### DIFF
--- a/Libraries/LibJS/Tests/Array.js
+++ b/Libraries/LibJS/Tests/Array.js
@@ -1,3 +1,5 @@
+load("test-common.js");
+
 try {
     assert(Array.length === 1);
     assert(Array.prototype.length === 0);
--- a/Libraries/LibJS/Tests/Array.prototype.pop.js
+++ b/Libraries/LibJS/Tests/Array.prototype.pop.js
@@ -1,3 +1,5 @@
+load("test-common.js");
+
 try {
     var a = [1, 2, 3];
     var value = a.pop();
--- a/Libraries/LibJS/Tests/Array.prototype.shift.js
+++ b/Libraries/LibJS/Tests/Array.prototype.shift.js
@@ -1,3 +1,5 @@
+load("test-common.js");
+
 try {
     var a = [1, 2, 3];
     var value = a.shift();
--- a/Libraries/LibJS/Tests/Array.prototype.toString.js
+++ b/Libraries/LibJS/Tests/Array.prototype.toString.js
@@ -1,3 +1,5 @@
+load("test-common.js");
+
 try {
     var a = [1, 2, 3];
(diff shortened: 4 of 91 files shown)


linusg:
Sure, `load()` as a "simple `require()` / `import`" is neat, but I don't like this change - it (kind of) re-introduces the problem we wanted to solve by adding a global `assert` function when in test mode: repetition in every test file. Now it's a function call instead of a function definition, how is that any better?

This will also complicate testing with the Lagom build of `js` - I often do `./js -t ../../../Libraries/LibJS/Tests/test-for-feature-im-working-on.js` which now breaks.

bgianfo:
Every test harness I've ever contributed grows it's own set of assertion helpers / test methods, etc. It's not really feasible to keep adding this stuff into Userland/js. It introduces more and more coupling between Userland/js and the test LibJS test suite, where there doesn't need to be any.

test-common.js will grow as more asserts / common test methods are added (assertNotReached() for example), so we need a way of importing all of those assets. I don't see a load() as repetition, is  #include repetition in every file? 

Another goal I had with this change was to address the issue where embedding assertions in Userland/js; we are no longer able to run tests via another JS implementation's repl anymore. d8 (developer v8) supports load(...) for example, so you could run LibJS tests without any other change. I'm assuming eventually we'll get require(..) and then nodejs will just work again as well. 

Your complaint about running tests from Lagom could easily be fixed by making run-tests have a mode where it runs an individual test specified test, and takes care of putting  you in the correct $pwd.

predmond:
> This will also complicate testing with the Lagom build of `js` - I often do `./js -t ../../../Libraries/LibJS/Tests/test-for-feature-im-working-on.js` which now breaks.

Or you could just do `../../../Meta/Lagom/build/js -t test-for-feature-im-working-on.js` ?

linusg:
> Another goal I had with this change was to address the issue where embedding assertions in Userland/js; we are no longer able to run tests via another JS implementation's repl anymore. d8 (developer v8) supports load(...) for example, so you could run LibJS tests without any other change. I'm assuming eventually we'll get require(..) and then nodejs will just work again as well.

That's much more convincing to me than calling two functions pollution :smile:

> Or you could just do `../../../Meta/Lagom/build/js -t test-for-feature-im-working-on.js` ?

Sure, but then I'm outside of the `build` directory and have to add paths to `cmake` and `make` to get there... Whatever, I'll just have to write myself an alias :)

awesomekling:
Yeah being able to run the tests in other engines is extremely valuable :)

PR #1790, opened 2020-04-13
Ports: work if bash is in a non-standard directory

Mindavi:
This change improves support for platforms that don't have bash
installed in the default /bin/bash directory.

As an example, for my platform (NixOS) bash is stored in:

/nix/store/ar3aqra4hzi56f7ivphrjy3n2cm9y4jr-bash-interactive-4.4-p23/bin/bash

I think this should work on (most) platforms and is more portable
than the hardcoded bash path.

### DIFF
--- a/Ports/.port_include.sh
+++ b/Ports/.port_include.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 set -eu
 
 SCRIPT=`dirname $0`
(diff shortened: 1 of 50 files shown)


bcoles:
```
/usr/bin/env: invalid option -- 'S'
```

Mindavi:
Mm, I'm afraid that option is also not very portable then :(

Mindavi:
I'm not sure how to fix this in a graceful way, so closing for now.

PR #1792, opened 2020-04-13
LibJS: Add String.prototype.{trim, trimStart, trimEnd}

kessejones:


linusg:
Please re-format the code with `clang-format`.

Functionality-wise this is incomplete, the trim functions are supposed to remove ALL characters ECMAScript considers whitespace, not just spaces: that includes newline, tab and a number of others. See https://www.ecma-international.org/ecma-262/5.1/#sec-7.2

Maybe we can deal with unicode later, but I think this should at least cover the ASCII characters :)

Libraries/LibJS/Tests/String.prototype.trim.js
```diff
@@ -0,0 +1,15 @@
+function assert(x) { if(!x) throw x; }
+
+try {
+    assert(String.prototype.trim.length === 0)
+    assert("   hello friends  ".trim() === "hello friends");
+    assert("hello friends   ".trim() === "hello friends");
+    assert("   hello friends".trim() === "hello friends");
+    assert("   hello friends".trimStart() === "hello friends");
+    assert("hello friends   ".trimEnd() === "hello friends");
```
linusg, line 9: this should probably also verify that `trimStart` does _not_ remove whitespace from the end and `timeEnd` from the start.

Libraries/LibJS/Tests/String.prototype.trim.js
```diff
@@ -0,0 +1,15 @@
+function assert(x) { if(!x) throw x; }
```
linusg, line 1: We don't need this anymore, `-t`/`--test-mode` takes care of this :)

kessejones:
Ok, I'll make the adjustments :)

kessejones:
Hey @linusg you can check this out because I believe it’s ready. :)

linusg:
I'm just a contributor, but not responsible merging PRs - @awesomekling will have a look :)

Libraries/LibJS/Tests/String.prototype.trim.js
```diff
@@ -0,0 +1,54 @@
+try {
```
linusg, line 1: This needs

```suggestion
load("test-common.js");

try {
```

since https://github.com/SerenityOS/serenity/pull/1787 was merged

awesomekling:
Looks all right, thank you :)

Issue #1142, opened 2020-01-26
LibC: Implement getservent, getservbyname, getservbyport, setservent, endservent functions


MWGuy:
**https://linux.die.net/man/3/getservbyname**

awesomekling:
https://pubs.opengroup.org/onlinepubs/9699919799/functions/getservbyname.html

rthugh02:
The services [man pages](https://linux.die.net/man/3/getservbyname) show that more than just getservbyname should be implemented to conform to the standard of retrieving services information i.e functions like endservent are expected to close the services file after it has been opened. Implementing just getservbyname leaves more to be desired.

I would like to propose this issue be expanded to include implementing all of the functions associated with accessing the services file.

rthugh02:
Would someone with write access mind closing/linking this to the related [pull request](https://github.com/SerenityOS/serenity/pull/1800)

awesomekling:
Fixed by #1800 :)

PR #1800, opened 2020-04-14
LibC: Implement getservent() function family

rthugh02:
Implemented service entry functionality in LibC as  defined in the [standard](https://linux.die.net/man/3/getservbyname)

Userland program used for testing:

```
#include <netdb.h>
#include <AK/LogStream.h>
int main() 
{
    struct servent* test;
    test = getservent();
    
    test = getservbyname("zserv", "udp");
    dbg() <<  test->s_name << " " << test->s_port << " " << test->s_proto << "\n";
    
    //should be fatserv 347/tcp
    test = getservent();
    dbg() <<  test->s_name << " " << test->s_port << " " << test->s_proto << "\n";

    //doing getservbyport, earlier than 347 to test reset, should also have aliases
    test = getservbyport(88, "tcp");
    dbg() <<  test->s_name << " " << test->s_port << " " << test->s_proto << " "
    << test->s_aliases[0] << " " << test->s_aliases[1] << " " << test->s_aliases[2] << "\n";

    //testing that sctp is found:
    test = getservbyname("amqp", "sctp");
    dbg() <<  test->s_name << " " << test->s_port << " " << test->s_proto << "\n";

    //testing that weird spaced services are found ok
    test = getservbyname("afs3-fileserver", "udp");
    dbg() <<  test->s_name << " " << test->s_port << " " << test->s_proto << " " << test->s_aliases[0] << "\n";
    //testing if no entry found:
    test = getservbyname("fake", "fake");
    dbg() << "test is null: " << (test == nullptr);
}
```

[related issue](https://github.com/SerenityOS/serenity/issues/1142)

### DIFF
--- a/Libraries/LibC/netdb.cpp
+++ b/Libraries/LibC/netdb.cpp
@@ -25,6 +25,7 @@
  */
 
 #include <AK/Assertions.h>
+#include <AK/ByteBuffer.h>
 #include <AK/ScopeGuard.h>
 #include <AK/String.h>
 #include <Kernel/Net/IPv4.h>
@@ -56,9 +57,9 @@ static servent __getserv_buffer;
 static char __getserv_name_buffer[512];
 static char __getserv_protocol_buffer[10];
 static int __getserv_port_buffer;
-static Vector<String> __getserv_alias_list_buffer = Vector<String>();
-static Vector<char*> __getserv_alias_list = Vector<char*>();
-static int keep_service_file_open = 0;
+static Vector<ByteBuffer> __getserv_alias_list_buffer;
+static Vector<char*> __getserv_alias_list;
+static bool keep_service_file_open = false;
 static ssize_t service_file_offset = 0;
 
 static int connect_to_lookup_server()
@@ -226,6 +227,7 @@ struct servent* getservent()
 
     if (fseek(services_file, service_file_offset, SEEK_SET) != 0) {
         perror("error seeking file");
+        fclose(services_file);
         return nullptr;
     }
     char* line = nullptr;
@@ -254,33 +256,18 @@ struct servent* getservent()
     }
 
     servent* service_entry = nullptr;
-    if (fill_getserv_buffers(line, read)) {
-        __getserv_buffer.s_name = __getserv_name_buffer;
-        __getserv_buffer.s_port = __getserv_port_buffer;
-        __getserv_buffer.s_proto = __getserv_protocol_buffer;
+    if (!fill_getserv_buffers(line, read))
+        return nullptr;
 
-        //freeing any previous alias data
-        for (size_t i = 0; i < __getserv_alias_list.size(); i++) {
-            free(__getserv_alias_list[i]);
-        }
-        __getserv_alias_list.clear();
-        size_t alias_list_size = __getserv_alias_list_buffer.size();
+    __getserv_buffer.s_name = __getserv_name_buffer;
+    __getserv_buffer.s_port = __getserv_port_buffer;
+    __getserv_buffer.s_proto = __getserv_protocol_buffer;
 
-        if (alias_list_size > 0) {
-            for (size_t i = 0; i < alias_list_size; i++) {
-
-                //Can't get non-const access to underlying String data, must duplicate to conform with servent member requirements
-                char* alias = strdup(__getserv_alias_list_buffer[i].characters());
-                if (alias == nullptr) {
-                    perror("error allocating space for alias");
-                    return nullptr;
-                }
-                __getserv_alias_list.append(alias);
-            }
-        }
-        __getserv_buffer.s_aliases = __getserv_alias_list.data();
-        service_entry = &__getserv_buffer;
+    for (auto& alias : __getserv_alias_list_buffer) {
+        __getserv_alias_list.append((char*)alias.data());
     }
+    __getserv_buffer.s_aliases = __getserv_alias_list.data();
+    service_entry = &__getserv_buffer;
 
     if (!keep_service_file_open) {
         endservent();
@@ -289,7 +276,7 @@ struct servent* getservent()
 }
 struct servent* getservbyname(const char* name, const char* protocol)
 {
-    int previous_file_open_setting = keep_service_file_open;
+    bool previous_file_open_setting = keep_service_file_open;
     setservent(1);
     struct servent* current_service = nullptr;
     auto service_file_handler = ScopeGuard([previous_file_open_setting] {
@@ -298,27 +285,21 @@ struct servent* getservbyname(const char* name, const char* protocol)
         }
     });
 
-    if (!protocol) {
-        do {
-            current_service = getservent();
-            if (!current_service) {
-                return nullptr;
-            }
-        } while (strcmp(current_service->s_name, name) != 0);
-    } else {
-        do {
-            current_service = getservent();
-            if (!current_service) {
-                return nullptr;
-            }
-        } while (!(strcmp(current_service->s_name, name) == 0 && strcmp(current_service->s_proto, protocol) == 0));
+    while (true) {
+        current_service = getservent();
+        if (current_service == nullptr)
+            break;
+        else if (!protocol && strcmp(current_service->s_name, name) == 0)
+            break;
+        else if (strcmp(current_service->s_name, name) == 0 && strcmp(current_service->s_proto, protocol) == 0)
+            break;
     }
 
     return current_service;
 }
 struct servent* getservbyport(int port, const char* protocol)
 {
-    int previous_file_open_setting = keep_service_file_open;
+    bool previous_file_open_setting = keep_service_file_open;
     setservent(1);
     struct servent* current_service = nullptr;
     auto service_file_handler = ScopeGuard([previous_file_open_setting] {
@@ -326,20 +307,14 @@ struct servent* getservbyport(int port, const char* protocol)
             endservent();
         }
     });
-    if (!protocol) {
-        do {
-            current_service = getservent();
-            if (!current_service) {
-                return nullptr;
-            }
-        } while (current_service->s_port != port);
-    } else {
-        do {
-            current_service = getservent();
-            if (!current_service) {
-                return nullptr;
-            }
-        } while (!(current_service->s_port == port && (strcmp(current_service->s_proto, protocol) == 0)));
+    while (true) {
+        current_service = getservent();
+        if (current_service == nullptr)
+            break;
+        else if (!protocol && current_service->s_port == port)
+            break;
+        else if (current_service->s_port == port && (strcmp(current_service->s_proto, protocol) == 0))
+            break;
     }
 
     return current_service;
@@ -420,7 +395,7 @@ static bool fill_getserv_buffers(char* line, ssize_t read)
             if (split_line[i].starts_with('#')) {
                 break;
             }
-            __getserv_alias_list_buffer.append(split_line[i]);
+            __getserv_alias_list_buffer.append(split_line[i].to_byte_buffer());
         }
     }
 
--- a/Libraries/LibC/netdb.cpp
+++ b/Libraries/LibC/netdb.cpp
@@ -48,6 +48,19 @@ static hostent __gethostbyaddr_buffer;
 static char __gethostbyaddr_name_buffer[512];
 static in_addr_t* __gethostbyaddr_address_list_buffer[2];
 
+static FILE* services_file = nullptr;
+static const char* services_path = "/etc/services";
+
+static bool fill_getserv_buffers(char* line, ssize_t read);
+static servent __getserv_buffer;
+static char __getserv_name_buffer[512];
+static char __getserv_protocol_buffer[10];
+static int __getserv_port_buffer;
+static Vector<String> __getserv_alias_list_buffer = Vector<String>();
+static Vector<char*> __getserv_alias_list = Vector<char*>();
+static int keep_service_file_open = 0;
+static ssize_t service_file_offset = 0;
+
 static int connect_to_lookup_server()
 {
     int fd = socket(AF_LOCAL, SOCK_STREAM | SOCK_CLOEXEC, 0);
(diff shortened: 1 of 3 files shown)


MWGuy:
Nice work!

awesomekling:
Hello, and thank you for working on this!

Per our [contribution guidelines](https://github.com/SerenityOS/serenity/blob/master/CONTRIBUTING.md), please make adjustments for the following points:

* *Conform to the [project coding style](https://github.com/SerenityOS/serenity/blob/master/Documentation/CodingStyle.md). Please use clang-format (version 8 or newer) to automatically format C++ files.*
* *The first line of the commit message should have the format "Category: Brief description of what's being changed". The "category" can be a subdirectory, but also something like "POSIX compliance" or "ClassName". Whatever seems logical.*
* *Write your commit messages in proper English, with care and punctuation.*
* *[Don't] iterate excessively on your design across multiple commits.*
* *[Don't] write in C.*

There's no need to preserve the history of whitespace changes you made while working on something. All it does is make reviewing the code harder. Instead, please squash this down to a handful of commits that don't iterate on the same thing over and over. :)

rthugh02:
Thank you for the feedback @awesomekling , I have just locally rebased and squashed the commits down and edited the messages. 

My only concerns are with your first and last points:

- I have already applied clang formatting (version 9.0.0-2 to be exact)
- Discussion in the IRC chat led me to believe that LibC is a special case in this project in regards to writing C, so I used C when seemed most appropriate (reading//seeking files, comparing char pointers, copying bytes to required structure members in the return buffer, etc.) and the AK libraries as well. Is there any places in particular you believe should be changed?

I will go ahead and push the squashed commit history so that you may review it, thanks!

Libraries/LibC/netdb.cpp
```diff
@@ -48,6 +49,8 @@ static hostent __gethostbyaddr_buffer;
 static char __gethostbyaddr_name_buffer[512];
 static in_addr_t* __gethostbyaddr_address_list_buffer[2];
 
+static auto services_file = Core::File::construct("/etc/services");
```
awesomekling: Please don't add something just to remove it in a subsequent commit unless it's necessary to support an ongoing change.

predmond:
> * Discussion in the IRC chat led me to believe that LibC is a special case in this project in regards to writing C, so I used C when seemed most appropriate (reading//seeking files, comparing char pointers, copying bytes to required structure members in the return buffer, etc.) and the AK libraries as well. Is there any places in particular you believe should be changed?

Sorry if I mislead you. What I meant by special case is you can't rely on other libraries in the system and must use the file io functions provided by LibC itself.

You can still use c++ language features.

Libraries/LibC/netdb.cpp
```diff
@@ -49,7 +48,18 @@ static hostent __gethostbyaddr_buffer;
 static char __gethostbyaddr_name_buffer[512];
 static in_addr_t* __gethostbyaddr_address_list_buffer[2];
 
-static auto services_file = Core::File::construct("/etc/services");
+static FILE* services_file = nullptr;
+const char* serv_file_loc = "/etc/services";
```
awesomekling: Does this variable need to be accessed from outside of this file? If not, please make it `static`.
Also, it's very abbreviation heavy which we try to avoid. How about `services_path`?

Libraries/LibC/netdb.cpp
```diff
@@ -202,10 +212,57 @@ hostent* gethostbyaddr(const void* addr, socklen_t addr_size, int type)
...
+//References for *servent implementations can be found here: https://pubs.opengroup.org/onlinepubs/007908799/xns/getservent.html
+//FIXME: description of getservbyname states that it gets the next entry of the database, opening a connection if necessary.
+//This to me says that the line to begin reading on should be saved even if connection is closed.
+//Also need to factor in the persistent connection provided byt setservent
 struct servent* getservent()
 {
-    dbg() << "FIXME: getservbyname()";
-    ASSERT_NOT_REACHED();
+    //If the services file is not open, attempt to open it and return null if it fails.
+    if(!services_file) {
```
awesomekling: Please use `clang-format` as you go instead of doing it in a separate commit at the end.

Libraries/LibC/netdb.cpp
```diff
@@ -202,10 +212,57 @@ hostent* gethostbyaddr(const void* addr, socklen_t addr_size, int type)
...
+        services_file = fopen(serv_file_loc, "r");
+
+        if(!services_file) {
+            perror("error opening services file");
+            return nullptr;
+        }
+    }
+
+    fseek(services_file, service_file_offset, SEEK_SET);
```
awesomekling, line 226: Missing an error check here.

Libraries/LibC/netdb.cpp
```diff
@@ -202,10 +212,57 @@ hostent* gethostbyaddr(const void* addr, socklen_t addr_size, int type)
...
+        if(!line) {
+            free(line);
+        }
+    });
+
+    //Read lines from services file until an actual service name is found.  
+    do {
+        read = getline(&line, &len, services_file);
+        service_file_offset += read;
+    } while(read != -1 && (line[0] < 65 || line[0] > 122));
```
awesomekling: If `read()` returns `0`, index `0` will not be valid.

Libraries/LibC/netdb.cpp
```diff
@@ -219,13 +276,77 @@ struct servent* getservbyport(int port, const char* protocol)
...
+    services_file = nullptr;
+    keep_service_file_open = 0;
+}
+
+//Fill the service entry buffer with the information contained in the currently read line, returns true if successfull, false if failure occurs.
+static bool fill_getserv_buffer(char* line, ssize_t read)
+{
+    //Splitting the line by tab delimiter and filling the servent buffers name, port, and protocol members.
+    auto split_line = String(line, read).split('\t');
+    strncpy(__getserv_name_buffer, split_line[0].characters(), split_line[0].length());
```
awesomekling: `strncpy` with a `String::length()` will not include a null terminator.
Also this way of using `strncpy()` seems kinda pointless -- we're not verifying that `split_line[0].length()` fits in `sizeof(__getserv_name_buffer)`.

Also, if `String::split()` returns an empty vector, index `0` will not be valid.

Libraries/LibC/netdb.cpp
```diff
@@ -219,13 +276,77 @@ struct servent* getservbyport(int port, const char* protocol)
...
+    __getserv_buffer.s_name = __getserv_name_buffer;
+    
+    auto port_protocol_split = String(split_line[1]).split('/');
```
awesomekling, line 342: `split_line[1]` may be an invalid index here.

Libraries/LibC/netdb.cpp
```diff
@@ -219,13 +276,77 @@ struct servent* getservbyport(int port, const char* protocol)
...
+    bool conversion_checker;
+    __getserv_buffer.s_port = port_protocol_split[0].to_int(conversion_checker);
```
awesomekling, line 344: `port_protocol_split[0]` may be an invalid index here.

Libraries/LibC/netdb.cpp
```diff
@@ -219,13 +276,77 @@ struct servent* getservbyport(int port, const char* protocol)
...
+    
+    if(!conversion_checker) {
+        return false;
+    }
+    strncpy(__getserv_protocol_buffer, port_protocol_split[1].characters(), port_protocol_split[1].length());
```
awesomekling: Same `strncpy()` comment as above.

Userland/testing_libc.cpp
```diff
@@ -0,0 +1,8 @@
+#include <netdb.h>
+int main() 
+{
+    getservent();
+    getservent();
+    getservent();
+    getservent();
+}
```
awesomekling: This seems more like a test utility to have on your local machine than something we should include in the userland utilities :)

Libraries/LibC/netdb.cpp
```diff
@@ -219,13 +276,77 @@ struct servent* getservbyport(int port, const char* protocol)
...
+            
+            char** allocated_location = __getserv_alias_list_buffer;
+            while(*allocated_location) {
+                free(*allocated_location);
+                *allocated_location = 0;
+                allocated_location++;
+            }
+        }
+        for(size_t i = 0; i < split_aliases.size(); i++) {
+                __getserv_alias_list_buffer[i] = strdup(split_aliases[i].characters());
```
awesomekling: Can we just use a C++ data structure with `String` and `Vector` as backing for `servent` (and synthesize a `servent` on demand when returning one, with pointers into our C++ structures)?
This seems overly complicated :)

Libraries/LibC/netdb.cpp
```diff
@@ -202,10 +212,57 @@ hostent* gethostbyaddr(const void* addr, socklen_t addr_size, int type)
...
+
+    fseek(services_file, service_file_offset, SEEK_SET);
+    char* line = nullptr;
+    size_t len = 0;
+    ssize_t read;
+    
+    auto free_line_on_exit = ScopeGuard([line] {
+        if(!line) {
+            free(line);
+        }
```
awesomekling: This doesn't look right.

Libraries/LibC/netdb.cpp
```diff
@@ -234,7 +234,7 @@ struct servent* getservent()
     ssize_t read;
     
     auto free_line_on_exit = ScopeGuard([line] {
-        if(!line) {
+        if(line) {
```
awesomekling: Please do this correctly in the first commit instead of iterating :)

Libraries/LibC/netdb.cpp
```diff
@@ -257,22 +257,64 @@ struct servent* getservent()
     }
     
     if(!keep_service_file_open) {
-        fclose(services_file);
-        services_file = nullptr;
+        endservent();
```
awesomekling: Please do this in the first commit instead of iterating.

Libraries/LibC/netdb.cpp
```diff
@@ -257,22 +257,64 @@ struct servent* getservent()
...
-    dbg() << "FIXME: getservbyname(\"" << name << "\", \"" << protocol << "\")";
-    ASSERT_NOT_REACHED();
+    int previous_file_open_setting = keep_service_file_open;
+    setservent(1);
+    struct servent* current_service = nullptr;
+    if(!protocol) {
+        do {
+            current_service = getservent();
+            if(!current_service) {
+                return nullptr;
```
awesomekling: Should we close the services file here, if it wasn't in "stay open" mode?
Can we use a `ScopeGuard` for this instead?

Libraries/LibC/netdb.cpp
```diff
@@ -257,22 +257,64 @@ struct servent* getservent()
...
+            }
+        } while(strcmp(current_service->s_name, name) != 0);
+    }
+    else {
+        do {
+            current_service = getservent();
+            if(!current_service) {
+                return nullptr;
```
awesomekling: Should we close the services file here, if it wasn't in "stay open" mode?
Can we use a `ScopeGuard` for this instead?

Libraries/LibC/netdb.cpp
```diff
@@ -257,22 +257,64 @@ struct servent* getservent()
...
-    dbg() << "FIXME: getservbyname(\"" << port << "\", \"" << protocol << "\")";
-    ASSERT_NOT_REACHED();
+    int previous_file_open_setting = keep_service_file_open;
+    setservent(1);
+    struct servent* current_service = nullptr;
+    if(!protocol) {
+        do {
+            current_service = getservent();
+            if(!current_service) {
+                return nullptr;
```
awesomekling: Should we close the services file here, if it wasn't in "stay open" mode?
Can we use a `ScopeGuard` for this instead?

Libraries/LibC/netdb.cpp
```diff
@@ -257,22 +257,64 @@ struct servent* getservent()
...
+            }
+        } while(current_service->s_port != port);
+    }
+    else {
+        do {
+            current_service = getservent();
+            if(!current_service) {
+                return nullptr;
```
awesomekling: Should we close the services file here, if it wasn't in "stay open" mode?
Can we use a `ScopeGuard` for this instead?

Libraries/LibC/netdb.cpp
```diff
@@ -294,15 +337,14 @@ void endservent()
...
-    keep_service_file_open = 0;
 }
 
 //Fill the service entry buffer with the information contained in the currently read line, returns true if successfull, false if failure occurs.
 static bool fill_getserv_buffer(char* line, ssize_t read)
 {
     //Splitting the line by tab delimiter and filling the servent buffers name, port, and protocol members.
     auto split_line = String(line, read).split('\t');
-    strncpy(__getserv_name_buffer, split_line[0].characters(), split_line[0].length());
+    strncpy(__getserv_name_buffer, split_line[0].characters(), split_line[0].length() + 1);
```
awesomekling: Please do this correctly in the first commit instead of iterating.
Also, we're still not checking if it fits into `sizeof(__getserv_name_buffer)`

Libraries/LibC/netdb.cpp
```diff
@@ -312,14 +354,14 @@ static bool fill_getserv_buffer(char* line, ssize_t read)
     if(!conversion_checker) {
         return false;
     }
-    strncpy(__getserv_protocol_buffer, port_protocol_split[1].characters(), port_protocol_split[1].length());
+    strncpy(__getserv_protocol_buffer, port_protocol_split[1].characters(), 3);
```
awesomekling: What if the file is malformed and the protocol is not 3 characters?

rthugh02:
Hi @awesomekling, I believe I have addressed all of the requested changes and properly squashed the commits to be atomic (sorry about that, I'm used to working on projects that want many small commits!)

Libraries/LibC/netdb.cpp
```diff
@@ -199,10 +212,218 @@ hostent* gethostbyaddr(const void* addr, socklen_t addr_size, int type)
...
+        for (size_t i = 0; i < __getserv_alias_list.size(); i++) {
+            free(__getserv_alias_list[i]);
+        }
+        __getserv_alias_list.clear();
+        size_t alias_list_size = __getserv_alias_list_buffer.size();
+
+        if (alias_list_size > 0) {
+            for (size_t i = 0; i < alias_list_size; i++) {
+
+                //Can't get non-const access to underlying String data, must duplicate to conform with servent member requirements
```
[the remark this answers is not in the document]
rthugh02: I initially attempted to use the ByteBuffer, in a line like this:

`__getserv_alias_list.append((char*)alias.to_byte_buffer().data());`

but accessing the aliases in my userland testing program causes a page fault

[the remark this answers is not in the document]
rthugh02: I discovered the problem, about to squash and push through :D

rthugh02:
@awesomekling Think this one should have everything covered

awesomekling:
All right, I think that's okay. Let's merge it :)

PR #1802, opened 2020-04-14
Ports: sl

Ivoah:
Port of https://github.com/mtoyoda/sl for SerenityOS


awesomekling:
Hey neat :)

bugaevc:
I'm getting
> cp: cannot stat '/home/sergey/dev/serenity/Root/usr/local/share/terminfo/78/xterm': No such file or directory

trying to build this. Does it depend on some other port, like ncurses or something?

bugaevc:
Also, it installs `sl` into `/usr/bin`, and not `/usr/local/bin` as it should. And we already have `/bin/sl` which launches Solitare.

Ivoah:
> I'm getting
> 
> > cp: cannot stat '/home/sergey/dev/serenity/Root/usr/local/share/terminfo/78/xterm': No such file or directory
> 
> trying to build this. Does it depend on some other port, like ncurses or something?

Yes it depends on ncurses. I have that listed in the depends section though. Is there something more I need to do for that?

Issue #1744, opened 2020-04-11
Kernel: ATA PIO mode doesn't work anymore


supercomputer7:
The bug was mentioned in the IRC channel by @deoxxa.

How to reproduce:

1. Run the kernel with the boot argument `force_pio`. Do that with `./run qcmd force_pio` or `./run q35_cmd force_pio`.

It's probably related to how we handle interrupts. One solution is to use the command `READ MULTIPLE (0xC4)` for reading multiple sectors and getting only one IRQ when the transfer is finished.

supercomputer7:
ATA PIO mode is working, but now we can't use PIO mode with the IOAPIC.

How to reproduce:

1. Run the kernel with the boot argument force_pio & smp=on. Do that with `./run qcmd force_pio smp=on` or `./run q35_cmd force_pio smp=on`.

supercomputer7:
The IOAPIC manual states that:
>Interrupt Mask—R/W. When this bit is 1, the interrupt signal is masked. Edge-sensitive 
>interrupts signaled on a masked interrupt pin are ignored (i.e., not delivered or held pending).

Code in QEMU that matches this requirement:
https://github.com/qemu/qemu/blob/2f7cc1fbd6f6655d900ca7f45973b9bd5330c6dd/hw/intc/ioapic.c#L181-L187

PR #1803, opened 2020-04-14
Kernel: Restore ATA PIO functionality

supercomputer7:


supercomputer7:
fixes #1744.

deoxxa:
Tested this on hardware and it works great :+1:

Libraries/LibBareMetal/IO.h
```diff
@@ -64,7 +64,7 @@ inline u32 in32(u16 port)
     return value;
 }
 
-inline void repeated_in16(u16 port, u8* buffer, int buffer_size)
+inline void repeated_in16(u16 port, u16* buffer, int buffer_size)
```
awesomekling, line 67: This seems like it would break existing code. If that's the case, it belongs in the following commit since it won't even build without this one.

awesomekling:
Great, thank you!

PR #1805, opened 2020-04-15
Browser: Default page title to URL if no title is provided

nicktiberi:
If a `<title>` tag is not provided on a web page, the browser will use the URL of the page as the default window title. This prevents **(null)** from appearing in the browser title bar for pages that don't have an explicit title set.


awesomekling:
Oh nice idea :)

PR #1806, opened 2020-04-15
LibJS: Check length property of Array.prototype.join in its test

linusg:
I assume this was copied from the `Array.prototype.push()` test :^)


awesomekling:
Whoops lol!

PR #1811, opened 2020-04-15
LibJS: Implement void operator

linusg:
- https://www.ecma-international.org/ecma-262/5.1/#sec-11.4.2
- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/void


awesomekling:
Cool!

PR #1813, opened 2020-04-15
Userland: Add "functrace" utility

itamar8910:
`functrace` is like `strace`, but in addition to syscalls it also logs regular userspace function calls.

It allows us to see what our programs do when they run :)

Screenshot:
functrace3

An improvement suggested by @kling is to parse debug symbols and print the arguments and return value of functions.


awesomekling:
This is super neat :)

PR #1815, opened 2020-04-15
LibJS: String.prototype.substring

kessejones:
Added substring function in String.prototype.

### DIFF
--- a/Libraries/LibJS/Runtime/StringPrototype.cpp
+++ b/Libraries/LibJS/Runtime/StringPrototype.cpp
@@ -56,6 +56,7 @@ StringPrototype::StringPrototype()
     put_native_function("trimStart", trim_start, 0);
     put_native_function("trimEnd", trim_end, 0);
     put_native_function("concat", concat, 1);
+    put_native_function("substring", substring, 2);
 }
 
 StringPrototype::~StringPrototype()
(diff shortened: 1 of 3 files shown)


awesomekling:
Please use `clang-format` so that your code [fits with the rest of the codebase](https://github.com/SerenityOS/serenity/blob/master/CONTRIBUTING.md).
Also please squash your commits instead of adding bugs and then fixing them. :)

Libraries/LibJS/Runtime/StringPrototype.cpp
```diff
@@ -307,4 +308,61 @@ Value StringPrototype::trim_end(Interpreter& interpreter)
     return trim_string(interpreter, *this_object, TrimMode::Right);
 }
 
+Value StringPrototype::substring(Interpreter& interpreter)
+{
+    auto* this_object = interpreter.this_value().to_object(interpreter.heap());
```
awesomekling, line 313: Missing a null check here. `to_object()` may fail due to an exception being thrown.
```suggestion
    auto* this_object = interpreter.this_value().to_object(interpreter.heap());
    if (!this_object)
        return {};
```

kessejones:
I don't know if there is a problem but I squash all the commits. Now I believe it is ready :)

awesomekling:
Okydoky!

PR #1817, opened 2020-04-16
LibJS: String.prototype.concat

kessejones:
Added concat function in String.prototype. :)

### DIFF
--- a/Libraries/LibJS/Runtime/StringPrototype.cpp
+++ b/Libraries/LibJS/Runtime/StringPrototype.cpp
@@ -55,6 +55,7 @@ StringPrototype::StringPrototype()
     put_native_function("trim", trim, 0);
     put_native_function("trimStart", trim_start, 0);
     put_native_function("trimEnd", trim_end, 0);
+    put_native_function("concat", concat, 1);
 }
 
 StringPrototype::~StringPrototype()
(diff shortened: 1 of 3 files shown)


Libraries/LibJS/Runtime/StringPrototype.cpp
```diff
@@ -305,4 +306,20 @@ Value StringPrototype::trimEnd(Interpreter& interpreter)
     return trim_string(interpreter, this_object, TrimMode::Right);
 }
 
+Value StringPrototype::concat(Interpreter& interpreter)
+{
+    auto* this_object = interpreter.this_value().to_object(interpreter.heap());
```
awesomekling, line 311: Missing a null check here in case the `to_object` operation throws.
```suggestion
    auto* this_object = interpreter.this_value().to_object(interpreter.heap());
    if (!this_object)
        return {};
```

Libraries/LibJS/Runtime/StringPrototype.cpp
```diff
@@ -307,4 +308,22 @@ Value StringPrototype::trim_end(Interpreter& interpreter)
...
+    if (!this_object)
+        return {};
+    auto& string = this_object->to_string().as_string()->string();
+
+    StringBuilder builder;
+    builder.append(string);
+
+    for(size_t i = 0; i < interpreter.argument_count(); ++i) {
```
awesomekling, line 321: Please use `clang-format`. Also please squash commits instead of stacking them, per the contribution guidelines.

awesomekling:
Great, thank you!

PR #1819, opened 2020-04-16
LibWeb: Scroll back to the top when a new page is loaded

jgbyrne:
Noticed while playing with the browser that this behaviour that I expected was missing and couldn't think of any reason why it shouldn't be implemented.


awesomekling:
Well that straight up makes sense!

PR #1824, opened 2020-04-17
LibWeb: Implement JS confirm()

nicktiberi:
Based on the following API docs: https://developer.mozilla.org/en-US/docs/Web/API/Window/confirm


linusg:
Too late now - but I believe `confirm()` with no argument should still show the `MessageBox`, with an empty string as `message`. The spec is not exactly clear about this, but it's the behaviour in at least Firefox and Chromium.

PR #1825, opened 2020-04-17
LibJS: Array.prototype.concat

kessejones:
Added concat function in Array.prototype. :)

### DIFF
--- a/Libraries/LibJS/Runtime/ArrayPrototype.cpp
+++ b/Libraries/LibJS/Runtime/ArrayPrototype.cpp
@@ -48,6 +48,7 @@ ArrayPrototype::ArrayPrototype()
     put_native_function("toString", to_string, 0);
     put_native_function("unshift", unshift, 1);
     put_native_function("join", join, 1);
+    put_native_function("concat", concat, 1);
     put("length", Value(0));
 }
 
@@ -227,6 +228,28 @@ Value ArrayPrototype::join(Interpreter& interpreter)
         separator = interpreter.argument(0).to_string();
 
     return join_array_with_separator(interpreter, *array, separator);
+}
+
+Value ArrayPrototype::concat(Interpreter& interpreter)
+{
+    auto* array = array_from(interpreter);
+    if (!array)
+        return {};
+
+    auto* new_array = interpreter.heap().allocate<Array>();
+    new_array->elements().append(array->elements());
+
+    for (size_t i = 0; i < interpreter.argument_count(); ++i) {
+        auto argument = interpreter.argument(i);
+        if (argument.is_array()) {
+            auto& argument_object = argument.as_object();
+            new_array->elements().append(argument_object.elements());
+        } else {
+            new_array->elements().append(argument);
+        }
+    }
+
+    return Value(new_array);
 }
 
 }
--- a/Libraries/LibJS/Runtime/ArrayPrototype.h
+++ b/Libraries/LibJS/Runtime/ArrayPrototype.h
@@ -48,6 +48,7 @@ private:
     static Value to_string(Interpreter&);
     static Value unshift(Interpreter&);
     static Value join(Interpreter&);
+    static Value concat(Interpreter&);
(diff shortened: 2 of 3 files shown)


Libraries/LibJS/Runtime/ArrayPrototype.cpp
```diff
@@ -229,4 +230,33 @@ Value ArrayPrototype::join(Interpreter& interpreter)
...
+    auto* new_array = interpreter.heap().allocate<Array>();
+    auto& start_array = *array->elements().begin();
+    auto* start_array_values = const_cast<const Value*>(&start_array);
+    new_array->elements().append(start_array_values, array->elements().size());
+
+    for (size_t i = 0; i < interpreter.argument_count(); ++i) {
+        auto argument = interpreter.argument(i);
+        if (argument.is_array()) {
+            auto* argument_object = argument.to_object(interpreter.heap());
+            auto* argument_array = static_cast<Array*>(argument_object);
```
awesomekling, line 248: Since `argument` is known to be an array, you can call `as_object()` and cast directly without calling `to_object()`.
```cpp
if (argument.is_array()) {
    auto& argument_array = static_cast<Array&>(argument.as_object());
```

Libraries/LibJS/Runtime/ArrayPrototype.cpp
```diff
@@ -229,4 +230,33 @@ Value ArrayPrototype::join(Interpreter& interpreter)
...
+
+            auto& arg_start_array = *argument_array->elements().begin();
+            auto* arg_start_array_values = const_cast<const Value*>(&arg_start_array);
+
+            new_array->elements().append(arg_start_array_values, argument_array->elements().size());
```
awesomekling, line 253: Can we just do this with `Vector::append(const Vector&)` instead?:
```suggestion
            new_array->elements().append(argument_array.elements());
```

Libraries/LibJS/Runtime/ArrayPrototype.cpp
```diff
@@ -229,4 +230,33 @@ Value ArrayPrototype::join(Interpreter& interpreter)
...
+Value ArrayPrototype::concat(Interpreter& interpreter)
+{
+    auto* array = array_from(interpreter);
+    if (!array)
+        return {};
+
+    auto* new_array = interpreter.heap().allocate<Array>();
+    auto& start_array = *array->elements().begin();
+    auto* start_array_values = const_cast<const Value*>(&start_array);
+    new_array->elements().append(start_array_values, array->elements().size());
```
awesomekling, line 242: Can we just do this with `Vector::append(const Vector&)`?:

```suggestion
    new_array->elements().append(array->elements());
```

Libraries/LibJS/Runtime/ArrayPrototype.cpp
kessejones: Ok

Libraries/LibJS/Runtime/ArrayPrototype.cpp
kessejones: Alright then, I just did it because I thought about reducing the grow_capacity calls

Libraries/LibJS/Runtime/ArrayPrototype.cpp
kessejones: same case as the previous

Libraries/LibJS/Runtime/ArrayPrototype.cpp
awesomekling: `Vector::append(Vector)` only does a single grow_capacity :)

kessejones: Yes, I got confused :)

kessejones:
I think it got better. :)

awesomekling:
Oh heck yeah dude, very nice and readable :)

PR #1826, opened 2020-04-17
LibJS: Implement automatic semicolon insertion

sunverwerth:
This implements one of the weirder parts of the Javascript spec.

### DIFF
--- a/AK/StringView.cpp
+++ b/AK/StringView.cpp
@@ -25,10 +25,10 @@
  */
 
 #include <AK/ByteBuffer.h>
+#include <AK/FlyString.h>
 #include <AK/Memory.h>
 #include <AK/String.h>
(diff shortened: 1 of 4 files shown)


awesomekling:
This might be even nicer if `StringView` had a `bool contains(char) const`, no? :)

sunverwerth:
True. Let me go ahead and add it

sunverwerth:
@awesomekling Done. JS tests are all green

awesomekling:
Cool! Now we just need to remember to consume or insert semicolon when parsing statements I guess :)

sunverwerth:
Ah crap! I spotted a mistake.
I'll open another PR 🙃

awesomekling:
Sounds like an opportunity to add a test too ;)

PR #1830, opened 2020-04-17
Shell: Explicitly check if command is a directory

linusg:
This is a bit nicer than getting "Exec format error" after trying to `execvp()` a directory.

[Return code in that case is 126.](https://pubs.opengroup.org/onlinepubs/9699919799//utilities/V3_chap02.html#tag_18_08_02)

### DIFF
--- a/Shell/main.cpp
+++ b/Shell/main.cpp
@@ -885,6 +885,12 @@ static int run_command(const String& cmd)
             if (handle_builtin(argv.size() - 1, argv.data(), retval))
                 return retval;
 
+            struct stat st;
+            if (stat(argv[0], &st) == 0 && S_ISDIR(st.st_mode)) {
+                fprintf(stderr, "Shell: %s: Is a directory\n", argv[0]);
+                return 126;
+            }
+
             pid_t child = fork();
             if (!child) {
                 setpgid(0, 0);


alimpfard:
hmmm...
my current shell (zsh) switches to the mentioned directory if there are no arguments present.
(Otherwise it says `command not found: <name of directory>`)

I'm wondering if it would be nice to have the "cd there if it's obviously a directory" behaviour?

linusg:
>  my current shell (zsh) switches to the mentioned directory if there are no arguments present.

So does mine (zsh :wink:) but personally I don't like it. IIRC it'll ask during initial setup is this behaviour should be enabled.
Bash for example doesn't do it.

Interested to see what others think? Either way, anything is nicer than an exec format error :D

alimpfard:
lol I don't remember the initial setup, I think I did that like...6 years ago?

Indeed, anything is better than an exec format error.

rthugh02:
I prefer how bash handles it personally. An action should only be taken with an explicit command, forgetting to type cd is annoying but what’s more annoying is when applications make bold assumptions about what they’re users’ intent is as opposed to working with the information given.

bugaevc:
> Interested to see what others think? Either way, anything is nicer than an exec format error :D

GNU Bash user here :wink:. I would never type the directory to `cd` into it, exactly because I can `cd`, so I don't find that intuitive, and that "selling point" of zsh always sounded moot to me. How about we detect trying to exec a directory client-side (in the Shell), and print a nice message like:

```
$ /etc
Shell: /etc is a directory, use `cd /etc` to go there
```

awesomekling:
Yeah I'm also with the "do what I tell you, not what you think I'm trying to tell you" crowd when it comes to file management. :)

alimpfard:
well, never really used that feature anyway.
@bugaevc "is a directory" sounds better to me tbqh.

bugaevc:
You mean it sounds better without the suggestion about using `cd`?

alimpfard:
Not gonna lie, I've always hated it when a shell tried to be helpful with
tips.
In the brief time I tried fish, the "tips" were extremely annoying.

--

bugaevc:
Well, I do like suggestions. Guess @awesomekling should decide then 😀

awesomekling:
I don't mind the suggestions. I'll let the person writing the code make the decision :)

linusg:
Lots of feedback, thanks!

> I'll let the person writing the code make the decision :)

I like @bugaevc's message suggestion, but I also don't feel like I _need_ that - the shell telling me "hey, this is a directory" should be enough to figure out "let's `cd` into it"...

PR #1831, opened 2020-04-17
3D Cube demo

sunverwerth:
Renders a spinning 3D cube. Demonstrates the Gfx::Painter::draw_triangle() method that is also included in this PR.

cube

The vector/matrix headers could be extracted in the future and shared with other parts of Serenity.

### DIFF
--- a/Demos/Cube/Cube.cpp
+++ b/Demos/Cube/Cube.cpp
@@ -1,0 +1,213 @@
+/*
+ * Copyright (c) 2020, Stephan Unverwerth <s.unverwerth@gmx.de>
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include <LibCore/ElapsedTimer.h>
+#include <LibGUI/Application.h>
+#include <LibGUI/Label.h>
+#include <LibGUI/Painter.h>
+#include <LibGUI/Widget.h>
+#include <LibGUI/Window.h>
+#include <LibGfx/Bitmap.h>
+#include <LibGfx/Matrix4x4.h>
+#include <LibGfx/Vector3.h>
+#include <math.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <time.h>
+
+const int WIDTH = 200;
+const int HEIGHT = 200;
+
+class Cube final : public GUI::Widget {
+    C_OBJECT(Cube)
+public:
+    virtual ~Cube() override;
+    void set_stat_label(RefPtr<GUI::Label> l) { m_stats = l; };
+
+private:
+    Cube();
+    RefPtr<Gfx::Bitmap> m_bitmap;
+    RefPtr<GUI::Label> m_stats;
+
+    virtual void paint_event(GUI::PaintEvent&) override;
+    virtual void timer_event(Core::TimerEvent&) override;
+
+    int m_accumulated_time;
+    int m_cycles;
+    int m_phase;
+};
+
+Cube::Cube()
+{
+    m_bitmap = Gfx::Bitmap::create(Gfx::BitmapFormat::RGBA32, { WIDTH, HEIGHT });
+
+    m_accumulated_time = 0;
+    m_cycles = 0;
+    m_phase = 0;
+
+    stop_timer();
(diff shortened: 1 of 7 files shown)


Quaker762:
> The vector/matrix headers could be extracted in the future and shared with other parts of Serenity.

Definitely. I'm working on an OpenGL implementation of Uni that I can't release yet, but at the end of the year I'll definitely use them (for things like `glFrustum()` etc. Does the compiler emit SIMD code for them?

Demos/Cube/Cube.cpp
```diff
@@ -0,0 +1,220 @@
...
+#include "Mat4x4f.h"
+#include "Vec3f.h"
+
+const int WIDTH = 200;
+const int HEIGHT = 200;
+
+/*
+ * Cube Widget
+*/
+class Cube : public GUI::Widget {
```
awesomekling, line 48: This class can be `final`

Demos/Cube/Cube.cpp
```diff
@@ -0,0 +1,220 @@
...
+    C_OBJECT(Cube)
+public:
+    virtual ~Cube() override;
+    void set_stat_label(RefPtr<GUI::Label> l) { stats = l; };
+
+private:
+    Cube();
+    RefPtr<Gfx::Bitmap> bitmap;
+    RefPtr<GUI::Label> stats;
```
awesomekling, line 57: Missing `m_` prefix on members.

Demos/Cube/Cube.cpp
```diff
@@ -0,0 +1,220 @@
...
+
+    virtual void paint_event(GUI::PaintEvent&) override;
+    virtual void timer_event(Core::TimerEvent&) override;
+
+    int timeAvg;
+    int cycles;
+    int phase;
```
awesomekling, line 64: Missing `m_` prefix on members.
Also `timeAvg` => `m_average_time` or similar.
Also please explicitly initialize primitive variables when declaring them.

Demos/Cube/Cube.cpp
```diff
@@ -0,0 +1,220 @@
...
+
+#include "Mat4x4f.h"
+#include "Vec3f.h"
+
+const int WIDTH = 200;
+const int HEIGHT = 200;
+
+/*
+ * Cube Widget
+*/
```
awesomekling, line 47: Not sure what this kind of comment adds but sure. :)

Demos/Cube/Cube.cpp
```diff
@@ -0,0 +1,220 @@
...
+        stats->set_text(String::format("%d ms", timeAvg / 50));
+        timeAvg = 0;
+    }
+
+    update();
+}
+
+/*
+ * Main
+*/
```
awesomekling, line 197: Not sure what this adds beyond +3 line count but ok.

Demos/Cube/Mat4x4f.h
```diff
@@ -0,0 +1,104 @@
...
+        float z = axis.z;
+
+        return Mat4x4f(
+            t * x * x + c, t * x * y - z * s, t * x * z + y * s, 0,
+            t * x * y + z * s, t * y * y + c, t * y * z - x * s, 0,
+            t * x * z - y * s, t * y * z + x * s, t * z * z + c, 0,
+            0, 0, 0, 1);
+    }
+
+    float m[4][4];
```
awesomekling, line 103: Missing `m_` prefix on member.

Demos/Cube/Mat4x4f.h
```diff
@@ -0,0 +1,104 @@
...
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#pragma once
+
+#include "Vec3f.h"
+#include <math.h>
+
+struct Mat4x4f {
```
awesomekling, line 32: This looks out of place. A more "Serenity" name might be `FloatMatrix4x4`, `Matrix4x4<float>` or just `Matrix4x4`.
Or maybe this is a term of art? I'm not well-versed in 3D things.

Demos/Cube/Vec3f.h
```diff
@@ -0,0 +1,93 @@
...
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#pragma once
+
+struct Vec3f {
```
awesomekling, line 29: This looks out of place. A more "Serenity" name might be `FloatVector3`, `Vector3<float>` or just `Vector3`.
Or maybe this is a term of art too?

Regardless, we tend to prefer `class` interfaces with getters and setters for complex types like this.

Demos/Cube/Vec3f.h
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    float length() const
+    {
+        return sqrt(x * x + y * y + z * z);
+    }
+
+    float x;
+    float y;
+    float z;
```
awesomekling, line 92: Missing `m_` prefix on members.

Demos/Cube/Vec3f.h
sunverwerth: I agree about the naming. I guess you can always `typedef` it if you like to type less.
I disagree about the getters/setters. These are plain value types with no invariants to uphold, adding setters/getters would just make the resulting code needlessly complex and verbose. Also since these are mathematical constructs they are unlikely to change in the future, so an access to x should always return the member x, nothing else.

awesomekling: Sure, but it's a stylistic choice for consistency over perceived simplicity. Look at all the other LibGfx primitives like `Point`, `Rect`, `Size`, etc. :)

Demos/Cube/Vec3f.h
sunverwerth: Can we make an exception for this struct? I would really prefer to write `foo.x` instead of `foo.m_x`. I'll add a comment there concerning the coding style, so nobody gets any silly ideas.

awesomekling: The style used throughout the project is something like:

* For types with many methods, prefer class over struct.
* For classes, make public getters and setters, keep members private with `m_` prefix.
* For structs, let everything be public and skip the `m_` prefix.

And we tend to use `foo.x()` rather than `foo.m_x`. :)

Demos/Cube/Vec3f.h
sunverwerth: Ok, this is a Serenity contribution after all, and not my personal project :^)

sunverwerth:
> 
> 
> > The vector/matrix headers could be extracted in the future and shared with other parts of Serenity.
> 
> Definitely. I'm working on an OpenGL implementation of Uni that I can't release yet, but at the end of the year I'll definitely use them (for things like `glFrustum()` etc. Does the compiler emit SIMD code for them?

@Quaker762 That's awesome. I thought about making a software OpenGL implementation for a second but then scrapped the thought immediately. :D Good luck with your project!

No SIMD code, just plain x87 ops. I think the cross compiler is set to produce i686 code. Maybe MMX would be possible, not sure how that works.

Demos/Cube/Cube.cpp
sunverwerth: Hey I stole this from the Fire Demo :D

awesomekling:
All right :D

PR #1832, opened 2020-04-17
LibJS: Implement nullish coalescing operator (??)

linusg:
I know, we're still in a phase where the parser can't even parse some of the basic stuff - but you gotta have [nullish coalescing](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator) :D

### DIFF
--- a/Libraries/LibJS/Tests/logical-expressions-basic.js
+++ b/Libraries/LibJS/Tests/logical-expressions-basic.js
@@ -17,7 +17,6 @@ try {
     assert(("foo" && true) === true);
     assert((false && "bar") === false);
     assert((true && "bar") === "bar");
-    assert((null && true) === null);
     assert((0 && false) === 0);
     assert((0 && true) === 0);
     assert((42 && false) === false);
@@ -55,7 +54,6 @@ try {
     assert(("foo" || true) === "foo");
     assert((false || "bar") === "bar");
     assert((true || "bar") === true);
-    assert((null || true) === true);
     assert((0 || false) === false);
     assert((0 || true) === true);
     assert((42 || false) === 42);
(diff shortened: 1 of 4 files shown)


awesomekling:
Very cool Linus :smile:

Libraries/LibJS/Tests/logical-expressions-basic.js
```diff
@@ -17,7 +17,6 @@ try {
     assert(("foo" && true) === true);
     assert((false && "bar") === false);
     assert((true && "bar") === "bar");
-    assert((null && true) === null);
```
awesomekling, line 20: This should still work, no? Why are we removing it?

Libraries/LibJS/Tests/logical-expressions-basic.js
```diff
@@ -55,7 +54,6 @@ try {
     assert(("foo" || true) === "foo");
     assert((false || "bar") === "bar");
     assert((true || "bar") === true);
-    assert((null || true) === true);
```
awesomekling, line 58: This should still work, no? Why are we removing it?

Libraries/LibJS/Tests/logical-expressions-basic.js
linusg: Oh no worries, it definitely still works. It's just a duplicated test :)

https://github.com/linusg/serenity/blob/d74ad81402c27113b2ed5586bc369703d6be61c4/Libraries/LibJS/Tests/logical-expressions-basic.js#L20

https://github.com/linusg/serenity/blob/d74ad81402c27113b2ed5586bc369703d6be61c4/Libraries/LibJS/Tests/logical-expressions-basic.js#L34

Libraries/LibJS/Tests/logical-expressions-basic.js
linusg: Same here :)

Libraries/LibJS/Tests/logical-expressions-basic.js
awesomekling: Aha! :)

PR #1835, opened 2020-04-18
Ports: Clean up lua port and enable dynamic library loading

Ivoah:


awesomekling:
We don't support dynamic libraries yet, so I don't see how this could work(?)

Ivoah:
This just adds the compile flag `-DLUA_USE_DLOPEN` for the future when serenity can load dynamic libraries. Actually trying to load an `.so` from lua trips an assert in `DynamicLoader`.

awesomekling:
Merged manually as f35b19e72dde3c11aaa1f90968037ebc7ad1c2d7 and f1a65d1d70da913af961761d3d2aa53455bd1cc0

PR #1836, opened 2020-04-18
Add Browser to QuickLaunch

bcoles:
With the increased focus on development of LibJS and LibWeb, it seems reasonable to add the `Browser` to the `QuickLaunch` tray.

Edit: Granted, using the QuickLaunch is slower than ctrl+shift+n, "br", enter, but that leaves a useless terminal window. As a desktop OS, Serenity should include the `Browser` in the QuickLaunch, in line with most other desktop OS.


awesomekling:
Heck yeah

PR #1837, opened 2020-04-18
Read-write lock, kind of

bugaevc:
A `Kernel::Lock` can now be locked in either _exclusive_ (the default) or _shared_ mode. Multiple threads can hold the same lock in shared mode at one time, but if any thread holds the lock in exclusive mode, no other thread can hold it at the same time in either mode.

Typically, the shared mode is used when reading a lockable resource, and exclusive mode when modifying it. This way, multiple threads can read the same resource without blocking each other, but while still blocking those threads that want to modify it.

The `Lock` remains _reentrant_: if a thread already holds a lock, it can claim it again, and the second time will succeed immediately (without blocking further), except when trying to "upgrade" a shared lock to a unique lock (which is not supported and will deadlock). If you hold a lock in exclusive mode, you can claim it again in either mode (the mode makes no difference in this case, you get the lock back immediately either way).

Note that `m_times_locked` is used for both "the reentrancy level" and "readers count", because they're essentially the same thing: how many times the lock has been locked; when it's unlocked back this many times, it actually becomes unlocked.

Caveat: no fairness is guaranteed, which means constantly arriving readers can starve writers.

The primary motivation for read-write locks was to make it possible for multiple threads to resolve paths (exercising Ext2FS as it's used for the root file system mount) without them blocking each other. This PR only prepares infrastructure for that, but does _not_ convert Ext2FS to use shared locking where appropriate, because I'm afraid to break it.

### DIFF
--- a/Kernel/FileSystem/FileDescription.cpp
+++ b/Kernel/FileSystem/FileDescription.cpp
@@ -168,7 +168,7 @@ ByteBuffer FileDescription::read_entire_file()
 
 ssize_t FileDescription::get_dir_entries(u8* buffer, ssize_t size)
 {
-    LOCKER(m_lock);
+    LOCKER(m_lock, Lock::Mode::Shared);
     if (!is_directory())
         return -ENOTDIR;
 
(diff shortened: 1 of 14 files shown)


awesomekling:
Ok!

PR #1838, opened 2020-04-18
Dual pivot quick sort implementation.

wilsonk:
Hopefully this is acceptable. Please let me know if something needs to be changed. This algorithm runs approximately 40% faster than the single pivot option on a large list of ints (may not be such a large difference for more complex types and, of course it isn't such a big difference on small lists of inputs).

### DIFF
--- a/AK/QuickSort.h
+++ b/AK/QuickSort.h
@@ -30,6 +30,61 @@
 
 namespace AK {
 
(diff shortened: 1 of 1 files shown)


awesomekling:
Sure thing :)

awesomekling:
Thanks!

PR #1844, opened 2020-04-18
Base: Add Redmond theme

bcoles:
Redmond


predmond:
Thanks :)

awesomekling:
I don't see why not! Seems a bit flat, but I'm sure we can improve it :)

bcoles:
> 
> 
> I don't see why not! Seems a bit flat, but I'm sure we can improve it :)

It's 16-bit style color scheme. Perhaps it would "pop" more with 32-bit style. Else I'm not really sure how to make it less flat with the available theme options. The gradient on the title bar could probably be improved.

bugaevc:
Inactive window titles look a bit wrong, don't you find?

The should be black on gray, and without a shadow, if possible.

bcoles:
> 
> 
> Inactive window titles look a bit wrong, don't you find?
> 
> The should be black on gray, and without a shadow, if possible.

Without the shadow, perhaps. Grey on grey is canon for 16-bit color.

Win2k 16-bit

bugaevc:
But that is NT, aka Redmond 2000 theme!

I'm talking about something more like this:
image

PR #1846, opened 2020-04-18
LibJS: Array.prototype.slice

kessejones:
Added slice function in Array.prototype.

### DIFF
--- a/Libraries/LibJS/Runtime/ArrayPrototype.cpp
+++ b/Libraries/LibJS/Runtime/ArrayPrototype.cpp
@@ -51,6 +51,7 @@ ArrayPrototype::ArrayPrototype()
     put_native_function("unshift", unshift, 1);
     put_native_function("join", join, 1);
     put_native_function("concat", concat, 1);
+    put_native_function("slice", slice, 2);
     put("length", Value(0));
 }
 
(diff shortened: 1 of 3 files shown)


Libraries/LibJS/Runtime/ArrayPrototype.cpp
```diff
@@ -254,4 +255,42 @@ Value ArrayPrototype::concat(Interpreter& interpreter)
...
+Value ArrayPrototype::slice(Interpreter& interpreter)
+{
+    auto* array = array_from(interpreter);
+    if (!array)
+        return {};
+
+    auto* new_array = Array::create(interpreter.global_object());
+    if (interpreter.argument_count() == 0) {
+        new_array->elements().append(array->elements());
+    } else {
```
awesomekling, line 267: Let's just return early here, then we don't need `else` and can avoid nesting.

Libraries/LibJS/Runtime/ArrayPrototype.cpp
```diff
@@ -254,4 +255,42 @@ Value ArrayPrototype::concat(Interpreter& interpreter)
...
+        ssize_t array_size = static_cast<ssize_t>(array->elements().size());
+        auto start_slice = interpreter.argument(0).to_i32();
+        auto end_slice = array_size;
+
+        if (start_slice > array_size)
+            return Value(new_array);
+        else if (start_slice < 0)
```
awesomekling, line 274: No need for `else` after `return`.

Libraries/LibJS/Runtime/ArrayPrototype.cpp
```diff
@@ -254,4 +255,42 @@ Value ArrayPrototype::concat(Interpreter& interpreter)
...
+    auto* new_array = Array::create(interpreter.global_object());
+    if (interpreter.argument_count() == 0) {
+        new_array->elements().append(array->elements());
+    } else {
+        ssize_t array_size = static_cast<ssize_t>(array->elements().size());
+        auto start_slice = interpreter.argument(0).to_i32();
+        auto end_slice = array_size;
+
+        if (start_slice > array_size)
+            return Value(new_array);
```
awesomekling, line 273: Object pointers like `Array*` can be returned directly:
```cpp
return new_array;
```

Libraries/LibJS/Tests/Array.prototype.slice.js
```diff
@@ -0,0 +1,53 @@
+load("test-common.js");
+
+try {
+    assert(Array.prototype.concat.length === 1);
```
linusg, line 4:
```suggestion
    assert(Array.prototype.slice.length === 1);
```

kessejones: thanks, it was because of ctrl+c and ctrl+v. :)

awesomekling:
👍

PR #1850, opened 2020-04-18
LibJS: Use AK::String::index_of() for StringPrototype::index_of()

linusg:
@itamar8910 added `AK::String::index_of()` in https://github.com/SerenityOS/serenity/commit/e207de84492a4a12bde4714f64263e2ed3a9db43 so that's one FIXME less :)


awesomekling:
Noice!

PR #1851, opened 2020-04-18
Demos: Add Screensaver demo

bcoles:
Rudimentary screensaver using color gradients. Computationally inexpensive.

Room for improvement:

* Pretty blending / dissolving.
* Hide the mouse cursor.
* The bitmap for first iteration (of `INTERVAL` length) is black.
* The screen resolution aspect ration could be used, rather than forcing a 4:3 aspect ratio. Although this shouldn't make much difference given that it's scaled.

Code mostly stolen from other demos.

Screensaver

### DIFF
--- a/Demos/Screensaver/Makefile
+++ b/Demos/Screensaver/Makefile
@@ -1,0 +1,8 @@
+OBJS = \
+    Screensaver.o
+
+PROGRAM = Screensaver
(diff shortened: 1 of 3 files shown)


sunverwerth:
I rate this 10/10 flying toasters

PR #1860, opened 2020-04-19
LibJS: Improve CallExpression::execute()'s error messages

linusg:
I've heard @awesomekling complaining about how "undefined is not a function" is more or less useless - hopefully for the last time :smile:

Let's improve this message.

```
> foo = {}
{  }
> foo.bar()
Uncaught exception: [TypeError]: undefined is not a function (evaluated from 'foo.bar')
> 
> foo[1 + 2]()
Uncaught exception: [TypeError]: undefined is not a function (evaluated from 'foo[<computed>]')
> 
> ({}).foo()
Uncaught exception: [TypeError]: undefined is not a function (evaluated from '<object>.foo')
> 
> ({})[1 + 2]()
Uncaught exception: [TypeError]: undefined is not a function (evaluated from '<object>[<computed>]')
>
> (1)()
Uncaught exception: [TypeError]: 1 is not a function
>
> "foo"()
Uncaught exception: [TypeError]: foo is not a function

```

Works with values other than `undefined` as well, of course, but that's going to be the most common one. E.g. via `ctx.someFunctionWeDontHaveYet()`.

We can't just (re-)evaluate the computed expression as that could have side effects, but maybe we can work towards every AST node knowing its source string in the future...

### DIFF
--- a/Libraries/LibJS/AST.cpp
+++ b/Libraries/LibJS/AST.cpp
@@ -92,16 +92,23 @@ Value CallExpression::execute(Interpreter& interpreter) const
 
     ASSERT(!callee.is_empty());
 
-    if (is_new_expression()) {
-        if (!callee.is_object()
-            || !callee.as_object().is_function()
-            || (callee.as_object().is_native_function()
-                && !static_cast<NativeFunction&>(callee.as_object()).has_constructor()))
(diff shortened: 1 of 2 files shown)


awesomekling:
Now that's a spicy improvement! Awesome :)

PR #1002, opened 2020-01-02
Build: add OpenBSD host support

jcs:
@Pagghiu: can you verify this still builds for you on macOS?

PR #841, opened 2019-12-02
Build serenity Toolchain, Kernel, Userland and Root FS natively on macOS

Pagghiu:
This pull request enables building serenity toolchain, kernel, userland and rootfs on macOS.
Some background discussion can be found visiting #808.

This involves using clang as bootstrap compiler to build GCC and the IPCCompiler tools, build everything and then mount and create the rootfs.

Notes:
- I have explicitly downloaded GCC dependencies to static link them, building a gcc cross-compiler that doesn't depend on any dylib installed. I've had problems in the past with brew installing different and incompatible versions of the support libraries (isl, mpfr etc.) so that's my suggested way of building GCC on macOS. Also the Linux build could statically link them to, to mitigate similar issues, if they exist.
- Patches in the build-root-filesystem.sh shell scripts are needed because macOS  mknod doesn't support -m option. 
- Some gnu tools and ext2 FUSE drivers need also to be installed, to allow building the root fs image, and that's what BuildFuseExt2.sh is for.

Everything should be compiling and working fine on macOS 10.14.6 and XCode 11.0 (my computer)
Some people have been able to build and run serenity from scratch on macOS using this pull request, so it's hopefully a little less "it works on my computer-esque" than usual ;)
I've been adding build instructions in the Readme file (not being sure if this is acceptable), but they can be moved anywhere it makes sense for them to be.

Finally, I've tried to be very conservative with all the changes, using a good amount of #ifdef and such.
I have been testing the build on a Linux Virtual Machine and it looks like I have not been braking anything. Nevertheless I suggest to run a full rebuild on a real Linux machine before merging it ;)
 
**Known Issues**
- AK Tests have been disabled because clang fails compiling them

qemu on macOS runs pretty slowly because:
- qemu kvm accelerator is not available on macOS
- qemu hypervisor framework accelerator doesn't work when using 32 bit virtualization 
- qemu intel haxm backend crashes serenity during kernel boot (to be investigated...)

PR #1861, opened 2020-04-19
Toolchain: Improve building toolchain on Darwin (macOS)

jarhill0:
At least on my system, `md5sum` exits with some kind of linking error
but `md5 -q` works, as on other BSD-based OSes. Likewise, `nproc` does
not exist but `sysctl -n hw.ncpu` works.

This commit uses those tools on Darwin (macOS) systems.

This change is unfortunately motivated solely by "it works on my machine." Before these changes, `BuildIt.sh` would stop at line 153 because `nproc` didn't exist. In addition, it would redownload tarballs over and over again because it couldn't successfully hash them. With these changes, `BuildIt.sh` runs successfully (exits with `0`), and doesn't redownload tarballs every run. I am then able to build and run SerenityOS.

Changes are inspired by #1002. However, in that PR, two people confirm that they can build on macOS at that point in time. This might indicate a configuration error on my end, which would make this PR unnecessary or ill-advised. I'm tagging @Pagghiu, author of #841, which initially added support for building on macOS, in hopes that he might have thoughts on this.

I'm on macOS 10.15.3 (19D76).

### DIFF
--- a/Toolchain/BuildIt.sh
+++ b/Toolchain/BuildIt.sh
@@ -30,6 +30,9 @@ elif [ "$(uname -s)" = "FreeBSD" ]; then
     MAKE=gmake
     MD5SUM="md5 -q"
     NPROC="sysctl -n hw.ncpu"
+elif [ "$(uname -s)" = "Darwin" ]; then
+    MD5SUM="md5 -q"
+    NPROC="sysctl -n hw.ncpu"
 fi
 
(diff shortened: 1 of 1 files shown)


jarhill0:
It seems my installation of `coreutils` was missing due to an installation error. Once I corrected that, the specified programs all exist on my `PATH` and run. 

Closing this as misconfiguration on my end.

PR #1863, opened 2020-04-19
LibJS: Implement Function.prototype.bind()

jack-karamanian:
This PR implements [`Function.prototype.bind`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/bind).
One design point I could go either way on: Storing the bound `this` and bound arguments on the `Function` class. 
An alternative could be to have virtual `bound_this()` and `bound_arguments()` which both return empty values for the `Function` class. A class such as `BoundFunction` could override those functions and provide the appropriate values. I went with the current approach because it removes the need for a hypothetical `ArrowFunction` class just to store a bound `this` value, as `ScriptFunction::create()` and its constructor could be modified to accept a bound `this` parameter.

### DIFF
--- a/Libraries/LibJS/Runtime/BoundFunction.cpp
+++ b/Libraries/LibJS/Runtime/BoundFunction.cpp
@@ -1,0 +1,71 @@
+/*
+ * Copyright (c) 2020, Jack Karamanian <karamanian.jack@gmail.com>
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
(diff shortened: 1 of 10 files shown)


Libraries/LibJS/Runtime/BoundFunction.cpp
```diff
@@ -0,0 +1,79 @@
...
+#include <LibJS/Interpreter.h>
+#include <LibJS/Runtime/BoundFunction.h>
+
+namespace JS {
+
+BoundFunction::BoundFunction(Function& target_function, Value new_bound_this, Vector<Value> arguments)
+    : Function::Function(*target_function.prototype())
+    , m_name(String::format("bound %s", target_function.name().characters()))
+{
+    if (auto length = target_function.get("length"); length.has_value() && length.value().is_number()) {
```
awesomekling, line 36: `Object::get()` can throw an exception, so I think we need to handle that scenario before we decide to construct a `BoundFunction`.

Libraries/LibJS/Runtime/BoundFunction.cpp
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    if (target_function.is_bound_function()) {
+        m_target_function = static_cast<BoundFunction&>(target_function).m_target_function;
+    } else {
+        m_target_function = &target_function;
+    }
+
+    m_constructor_prototype = [this]() -> Object* {
+        auto prototype = m_target_function->get("prototype");
```
awesomekling, line 49: Same issue here, `get()` can throw.

Libraries/LibJS/Runtime/Function.h
```diff
@@ -40,12 +40,28 @@ class Function : public Object {
...
 protected:
     explicit Function(Object& prototype);
     virtual const char* class_name() const override { return "Function"; }
 
+    void bind_this_and_arguments(Value bound_this, Vector<Value> bound_arguments);
+
 private:
     virtual bool is_function() const final { return true; }
+    Optional<Value> m_bound_this;
+    Vector<Value> m_bound_arguments;
```
awesomekling, line 64: To ensure that these don't get garbage collected, we have to mark them in `Function::visit_children()`.
See for example how `ScriptFunction` marks its parent lexical environment.

Libraries/LibJS/Tests/Function.prototype.bind.js
```diff
@@ -0,0 +1,114 @@
...
+
+    // Function.prototype.bind should not accept non-function values.
+    try {
+        Function.prototype.bind.call("foo");
+        assertNotReached();
+    } catch (e) {
+        assert(e instanceof Error);
+        assert(e.name === "TypeError");
+        assert(e.message === "Not a Function object");
+    }
```
linusg, line 90:
```suggestion
    assertThrowsError(() => {
        Function.prototype.bind.call("foo");
    }, {
        error: TypeError,
        message: "Not a Function object"
    });
```

jack-karamanian: :+1:

jack-karamanian:
@awesomekling Feedback has been addressed :)

awesomekling:
Looks legit. Nice work :)

PR #1865, opened 2020-04-19
Browser: Persistent bookmark removal.

theazgra:
Store the JSON model after the bookmark is removed.
Removed bookmark wasn't removed from the JSON model, so that on the next browser start it would show up again.


awesomekling:
Hello! Thank you for working on this :)

I don't understand the second commit, when would it fail to remove the bookmark, and why should we keep scanning if we find a match, but `remove()` returns `false` for whatever reason?

theazgra:
I fixed the code to exit the function once we find the corresponding item no matter what the `remove()` returns.
I still like to return the result of `remove()` even though it should never fail.

awesomekling:
All right :) Thanks for the fix!

PR #1868, opened 2020-04-19
LibLine: Complete only common prefixes, and tweak suggestion interface

alimpfard:
cc @bugaevc 

`/re<tab>` -> `/res/`
`/r<tab>` -> `/r` showing suggestions, without having anything selected
`ls www/ac<tab>` -> `ls www/acid`, and shows suggestions.
`<tab>` -> immediately show suggestions (nothing selected)

Changes the suggestion callbacks to return a `Vector<Line::CompletionSuggestion>`, which contains a `text` field, and a `trivia` field.
The `trivia` string is only appended when the suggestion is selected.

### DIFF
--- a/Libraries/LibLine/Editor.cpp
+++ b/Libraries/LibLine/Editor.cpp
@@ -290,12 +290,38 @@ String Editor::get_line(const String& prompt)
                 String token = is_empty_token ? String() : String(&m_buffer[token_start], m_cursor - token_start);
 
                 // ask for completions only on the first tab
+                // and scan for the largest common prefix to display
                 // further tabs simply show the cached completions
                 if (m_times_tab_pressed == 1) {
                     if (is_first_token)
                         m_suggestions = on_tab_complete_first_token(token);
                     else
                         m_suggestions = on_tab_complete_other_token(token);
+                    size_t common_suggestion_prefix { 0 };
+                    if (m_suggestions.size() == 1) {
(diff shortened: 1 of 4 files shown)


Libraries/LibLine/Editor.cpp
```diff
@@ -462,7 +508,7 @@ String Editor::get_line(const String& prompt)
                 m_refresh_needed = true;
                 continue;
             }
-            if (ch == 0xc) {                    // ^L
```
alimpfard, line 465: AAAAAAAAAAAAAAA
WHY
you know what, I'm gonna move this comment somewhere else dammit

alimpfard: Ha! take __that__ clang-format!

bugaevc:
> `/r<tab> -> /res` also showing suggestions

No no no, that is what I was complaining about in the first place! Do not append ambiguous completions, ever! It should only show suggestions in this case.

alimpfard:
showing suggestions but not completing...I see what you're mad about
let me see if I can deal with that in a more..."don't reinvent the entire suggestion subsystem" way

bugaevc:
Let me repeat it one more time here:
* Only ever append the unambiguous part, if any (for `/r`, there's no unambiguous part)
* If there is an ambiguity, show the list
* If there are no suggestions, beep

alimpfard:
There we go, @bugaevc here's your ordered slice of LibLine

PR #1871, opened 2020-04-19
FileManager: Ignore empty selections on tree view

anpep:
If the selection is empty, the model index will be invalid and the file
system model will return the root directory path by default. This causes
the file manager to jump to the root directory when the currently
selected item on the tree view is deselected.


awesomekling:
:+1:

PR #1874, opened 2020-04-19
LibJS: Do not assume that a call frame exists in {get,set}_variable

alimpfard:
Title says all~

### DIFF
--- a/Libraries/LibJS/Interpreter.cpp
+++ b/Libraries/LibJS/Interpreter.cpp
@@ -132,16 +132,18 @@ void Interpreter::exit_scope(const ScopeNode& scope_node)
 
 void Interpreter::set_variable(const FlyString& name, Value value, bool first_assignment)
 {
-    for (auto* environment = current_environment(); environment; environment = environment->parent()) {
-        auto possible_match = environment->get(name);
-        if (possible_match.has_value()) {
-            if (!first_assignment && possible_match.value().declaration_kind == DeclarationKind::Const) {
-                throw_exception<TypeError>("Assignment to constant variable");
+    if (m_call_stack.size()) {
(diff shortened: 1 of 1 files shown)


Libraries/LibJS/Interpreter.cpp
```diff
@@ -132,29 +132,31 @@ void Interpreter::exit_scope(const ScopeNode& scope_node)
 
 void Interpreter::set_variable(const FlyString& name, Value value, bool first_assignment)
 {
-    for (auto* environment = current_environment(); environment; environment = environment->parent()) {
-        auto possible_match = environment->get(name);
-        if (possible_match.has_value()) {
-            if (!first_assignment && possible_match.value().declaration_kind == DeclarationKind::Const) {
-                throw_exception<TypeError>("Assignment to constant variable");
+    if (m_call_stack.size())
```
awesomekling, line 135: Please wrap multi-line consequent blocks in `{` and `}` :)

Libraries/LibJS/Interpreter.cpp
```diff
@@ -132,29 +132,31 @@ void Interpreter::exit_scope(const ScopeNode& scope_node)
...
 }
 
 Optional<Value> Interpreter::get_variable(const FlyString& name)
 {
-    for (auto* environment = current_environment(); environment; environment = environment->parent()) {
-        auto possible_match = environment->get(name);
-        if (possible_match.has_value())
-            return possible_match.value().value;
-    }
+    if (m_call_stack.size())
```
awesomekling, line 154: Same here.

Libraries/LibJS/Interpreter.cpp
alimpfard: okies

PR #1881, opened 2020-04-19
PaintBrush: Fix crash with flood fill.

NicolasVanBossuyt:
Fixed a crash occurring when initiating a flood fill out of the bitmap rectangle.


awesomekling:
:+1:

PR #1884, opened 2020-04-19
LibLine: Implement some history searching stuff

alimpfard:
This commit adds searching in the editor history with ^R.
It does so by instantiating...another Line::Editor inside the current
Line::Editor :^)

and yes, @linusg, this works just fine with js's syntax highlighting :^)

- pressing Enter while the search is up will execute the selected history element
- pressing Tab will quit search, but keep the buffer and continue the prompt (better key for this?)
- ^R and backspace will cycle backwards and forwards, respectively.

### DIFF
--- a/Libraries/LibLine/Editor.cpp
+++ b/Libraries/LibLine/Editor.cpp
@@ -33,8 +33,9 @@
 
 namespace Line {
 
-Editor::Editor()
+Editor::Editor(bool always_refresh)
 {
+    m_always_refresh = always_refresh;
     m_pending_chars = ByteBuffer::create_uninitialized(0);
     struct winsize ws;
     if (ioctl(STDOUT_FILENO, TIOCGWINSZ, &ws) < 0) {
@@ -126,7 +127,17 @@ String Editor::get_line(const String& prompt)
 
     m_history_cursor = m_history.size();
     for (;;) {
+        if (m_always_refresh)
+            m_refresh_needed = true;
         refresh_display();
+        if (m_finish) {
+            m_finish = false;
+            printf("\n");
+            fflush(stdout);
+            auto string = String::copy(m_buffer);
+            m_buffer.clear();
+            return string;
+        }
         char keybuf[16];
         ssize_t nread = read(0, keybuf, sizeof(keybuf));
         // FIXME: exit()ing here is a bit off. Should communicate failure to caller somehow instead.
@@ -138,13 +149,17 @@ String Editor::get_line(const String& prompt)
                     m_was_interrupted = false;
                     if (!m_buffer.is_empty())
                         printf("^C");
+
+                    if (m_is_searching) {
+                        end_search();
+                        continue;
+                    }
                 }
                 if (m_was_resized)
                     continue;
 
-                m_buffer.clear();
-                putchar('\n');
-                return String::empty();
+                finish();
+                continue;
             }
             perror("read failed");
             // FIXME: exit()ing here is a bit off. Should communicate failure to caller somehow instead.
@@ -500,6 +515,9 @@ String Editor::get_line(const String& prompt)
             m_times_tab_pressed = 0; // Safe to say if we get here, the user didn't press TAB
 
             auto do_backspace = [&] {
+                if (m_is_searching) {
+                    return;
+                }
                 if (m_cursor == 0) {
                     fputc('\a', stdout);
                     fflush(stdout);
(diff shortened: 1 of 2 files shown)


linusg:
img

bugaevc:
Can we also support [this feature](https://codeinthehole.com/tips/the-most-important-command-line-tip-incremental-history-searching-with-inputrc/) as an alternative way of searching through history? Without any config file editing to enable it, of course.

alimpfard:
I had thought about that, but I skipped over it...
My thought process was "what if I *didn't* want to search for whatever that
was in the editor buffer?"

If you have any suggestions on how it should behave in that regard, let me
know and I'll add this search mode

bugaevc:
> My thought process was "what if I *didn't* want to search for whatever that was in the editor buffer?"

Sorry, I don't follow: if you didn't want to search for it, you wouldn't have inputted it and pressed Up?

alimpfard:
Imagine this scenario:

I am writing some random command when I remember that I have it already in
the history, but it won't match what I have written up to this point.
Either I erase to some point (ew)
Or I rewrite it up to that point (a waste)
Or...?

alimpfard:
While writing that I realised it could just search for the substring that
the cursor marks the end of...

I think that's a good enough semantic for it, I'll implement it.

bugaevc:
> I am writing some random command when I remember that I have it already in
the history, but it won't match what I have written up to this point.

You erase it, yes, using some clever key combo for "erase to beginning of the line". I personally just press Ctrl-C, which prints `^C` and puts me on a fresh line. And — is it any different than what you'd do with Ctrl-R?

alimpfard:
> You erase it
Gosh, that's so inefficient. Well, the scheme I came up with should cover this as well, so I guess that's fine?

bugaevc:
> Gosh, that's so inefficient

I still don't see how that's different from when you use Ctrl-R for searching?

> Well, the scheme I came up with should cover this as well, so I guess that's fine?

Maybe if you explain how your scheme is different from what libreadline does again :D

alimpfard:
I legit have no idea what libreadline does!
I don't think readline even supports searches like that

alimpfard:
> different from ^R

The comparison isn't with ^R, that's its own thing.
I'm just thinking about how we should handle normal "scrolling" in the history.

bugaevc:
> I legit have no idea what libreadline does! I don't think readline even supports searches like that

Well, please do try it out. See the post I linked about how to enable it.

> I'm just thinking about how we should handle normal "scrolling" in the history

Basically if you start pressing up/down without typing anything, it, naturally, matches *every* line in history, so you get all your history entries.

Libraries/LibLine/Editor.cpp
```diff
@@ -33,8 +33,9 @@
 
 namespace Line {
 
-Editor::Editor()
+Editor::Editor(bool lazy)
 {
+    m_is_lazy = lazy;
```
awesomekling, line 38: What does lazy mean in this context?

Libraries/LibLine/Editor.h
```diff
@@ -197,6 +216,15 @@ class Editor {
     void recalculate_origin();
     void reposition_cursor();
 
+    bool m_finish = false;
+
+    OwnPtr<Editor> m_search_editor = nullptr;
```
awesomekling, line 222: No need to explicitly initialize smart pointers.

Libraries/LibLine/Editor.h
```diff
@@ -197,6 +216,15 @@ class Editor {
     void recalculate_origin();
     void reposition_cursor();
 
+    bool m_finish = false;
```
awesomekling, line 220: Mixing `=` and `{}` initialization.

Libraries/LibLine/Editor.cpp
alimpfard: Whether it will draw lazily (only whatever is needed)
I guess I'll rename it to always_refresh

alimpfard:
hmmm...that went two commits back...

bugaevc:
* In Ctrl-R mode, pressing Enter should not result in the command from history getting executed, you should be able to edit it furter and then press enter later to execute it as you do normally.
* When normally editing a command, pressing Ctrl-C should not result in it getting executed; we should just drop the command, without executing it or adding it to the history.

alimpfard:
Ctrl-C executes? how did I mess that up?

But...what if I wanted it to?
I added pressing tab if further editing is desired, which makes a tiny bit of sense
if I go and swap those (enter to edit, tab to execute), it'll just be weird

bugaevc:
> In Ctrl-R mode, pressing Enter should not result in the command from history getting executed, you should be able to edit it furter and then press enter later to execute it as you do normally.

> I added pressing tab if further editing is desired, which makes a tiny bit of sense

I take that back, in Bash + libreadline it does get executed, and Tab lets you edit it further.

alimpfard:
arrrg @alimpfard

PR #1885, opened 2020-04-19
Debugger: Source-level Debugging

itamar8910:
This PR adds the following capabilities to `sdb`:
- Print the current source location, if available.
- Add a breakpoint at a source location (`bp <file>:<line>`)
- `sl`  - step to the next line in source.

To do source-level debugging, we parse the DWARF `.debug_line` data to get line number information.

I tried adding `-g` to `CXXFLAGS` for all the the executables in the build process,
but since we don't currently have shared libraries the debug info for libraries is duplicated many times over, and we run out of space in the disk image :disappointed:

So for now, we can test this out with small programs.
An example:

`test.c`
```c
1. #include <stdio.h>
2.
3. int foo();
4.
5. int main()
6. {
7.    int a = foo();
8.    printf("%d\n", a);
9. }
```
`foo.c`
```c
1. int foo(){
2.    int x = 1;
3.    int y = 3;
4.    return x + y;
5. }
```
sdb-source2

### DIFF
--- a/Libraries/LibDebug/DebugInfo.h
+++ b/Libraries/LibDebug/DebugInfo.h
@@ -1,0 +1,56 @@
+/*
+ * Copyright (c) 2020, Itamar S. <itamar8910@gmail.com>
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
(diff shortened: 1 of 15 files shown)


Libraries/LibDebug/DebugInfo.h
```diff
@@ -0,0 +1,56 @@
...
+#include <LibELF/Loader.h>
+#include <Libraries/LibDebug/Dwarf/LineProgram.h>
+
+class DebugInfo {
+public:
+    explicit DebugInfo(NonnullRefPtr<const ELF::Loader> elf);
+
+    struct SourcePosition {
+        String file_path;
+        size_t line_number;
```
awesomekling, line 41: Please explicitly initialize primitives like this.
```suggestion
        size_t line_number { 0 };
```

Libraries/LibDebug/Dwarf/LineProgram.h
```diff
@@ -0,0 +1,113 @@
...
+#pragma once
+#include <AK/BufferStream.h>
+#include <AK/Vector.h>
+
+class LineProgram {
+public:
+    explicit LineProgram(BufferStream& stream);
+
+    struct LineInfo {
+        u32 address;
```
awesomekling, line 36: Please explicitly initialize primitives.

Libraries/LibDebug/Dwarf/LineProgram.h
```diff
@@ -0,0 +1,113 @@
...
+        String file;
+        size_t line;
```
awesomekling, line 38: Please explicitly initialize primitives.

Libraries/LibDebug/Dwarf/LineProgram.h
```diff
@@ -0,0 +1,113 @@
...
+    enum ExtendedOpcodes {
+        EndSequence = 1,
+        SetAddress,
+        DefineFile,
+        SetDiscriminator,
+    };
+
+    struct FileEntry {
+        String name;
+        size_t directory_index;
```
awesomekling, line 93: Please explicitly initialize primitives.

Libraries/LibDebug/Dwarf/LineProgram.h
```diff
@@ -0,0 +1,113 @@
...
+    size_t m_unit_offset { 0 };
+    UnitHeader32 m_unit_header {};
+    Vector<String> m_source_directories;
+    Vector<FileEntry> m_source_files;
+
+    // The registers of the "line program" virtual machine
+    u32 m_address { 0 };
+    size_t m_line { 0 };
+    size_t m_file_index { 0 };
+    bool m_is_statement = { false };
```
awesomekling, line 110: This looks weird. Either `= false` or `{ false }` but let's not have both :D

Libraries/LibDebug/DebugSession.h
```diff
@@ -104,14 +108,14 @@ class DebugSession {
...
-    static constexpr u8 BREAKPOINT_INSTRUCTION
-        = 0xcc;
+    static constexpr u8 BREAKPOINT_INSTRUCTION = 0xcc;
 
     int m_debugee_pid { -1 };
     bool m_is_debugee_dead { false };
 
     MappedFile m_executable;
-    ELF::Loader m_elf;
+    NonnullRefPtr<const ELF::Loader> m_elf;
```
awesomekling, line 117: The fact that this didn't break immediately when you made `ELF::Loader` refcounted is why I like to add static construction helpers for ref-counted objects :)

sunverwerth:
@itamar8910 @awesomekling Now somebody needs to implement controlling `sdb` from within HackStudio. 😎

PR #1891, opened 2020-04-20
LibJS: Array.prototype.reverse

kessejones:
Added reverse function in Array.prototype. :)

### DIFF
--- a/Libraries/LibJS/Runtime/ArrayPrototype.cpp
+++ b/Libraries/LibJS/Runtime/ArrayPrototype.cpp
@@ -54,6 +54,7 @@ ArrayPrototype::ArrayPrototype()
     put_native_function("concat", concat, 1);
     put_native_function("slice", slice, 2);
     put_native_function("indexOf", index_of, 1);
+    put_native_function("reverse", reverse, 0);
     put("length", Value(0));
 }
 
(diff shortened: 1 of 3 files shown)


linusg:
We probably can skip this for now, so just FYI:

> The reverse function is intentionally generic; it does not require that its this value be an Array object. Therefore, it can be transferred to other kinds of objects for use as a method.

Libraries/LibJS/Runtime/ArrayPrototype.cpp
```diff
@@ -343,4 +344,26 @@ Value ArrayPrototype::index_of(Interpreter& interpreter)
...
+        return array;
+
+    auto* array_reverse = Array::create(interpreter.global_object());
+    array_reverse->elements().ensure_capacity(array->elements().size());
+
+    for (ssize_t i = array->elements().size() - 1; i >= 0; --i)
+        array_reverse->elements().append(array->elements().at(i));
+
+    array->elements().clear_with_capacity();
+    array->elements().append(array_reverse->elements());
```
awesomekling, line 364: No need to create a temporary throwaway `Array` here. Just make a `Vector<Value>` on the stack:

```cpp
Vector<Value> values;
values.ensure_capacity(array->elements.size());
for (ssize_t i = array->elements().size() - 1; i >= 0; --i)
    values.append(array->elements()[i]);
array->elements() = move(values);
```

And yeah, like Linus says, this should be made generic in the future, but an array-only version is okay for now. :)

PR #1894, opened 2020-04-20
LibGUI: Add Select all action to TextEditor

anpep:
Previously, TextEditor processed the Select all command directly on the
keydown event handler. For this reason, WindowManager would not process
it as an action for the focused control and an action with the same
keyboard shortcut from the parent could override the TextEditor's one
even when it is focused.

For instance, when pressing Ctrl+A on the FileManager's path bar, all
files were selected instead, which is not the expected behavior.

Now the Select all command is an actual action on TextEditor, so that
WindowManager can process it correctly before any other actions. I also
added an icon for it!

image


awesomekling:
Excellent change!

PR #1895, opened 2020-04-20
Meta: Build Lagom js and run LibJS tests on Travis

linusg:
Not sure if this works yet, let's see.


linusg:
it does now :)

awesomekling:
Now that's super neat :)

PR #1897, opened 2020-04-20
LibGui: Stop sending text selection updates on mousemove_event during automatic scrolling.

VKoskiv:
This little tweak fixes the issue where the scrolling speeds up
significantly if the user wiggles their cursor. Just something obvious I
spotted while watching the video :^)

### DIFF
--- a/Libraries/LibGUI/TextEditor.cpp
+++ b/Libraries/LibGUI/TextEditor.cpp
@@ -301,7 +301,7 @@ void TextEditor::mouseup_event(MouseEvent& event)
 void TextEditor::mousemove_event(MouseEvent& event)
 {
     m_last_mousemove_position = event.position();
-    if (m_in_drag_select) {
+    if (m_in_drag_select && !m_automatic_selection_scroll_timer->is_active()) {
         set_cursor(text_position_at(event.position()));
         m_selection.set_end(m_cursor);
         did_update_selection();


Libraries/LibGUI/TextEditor.cpp
```diff
@@ -301,7 +301,7 @@ void TextEditor::mouseup_event(MouseEvent& event)
 void TextEditor::mousemove_event(MouseEvent& event)
 {
     m_last_mousemove_position = event.position();
-    if (m_in_drag_select) {
+    if (m_in_drag_select && !m_in_drag_select) {
```
Dexesttp, line 304: That would just be `if (false) {`, and it would cause the scrolling to stop if the timer is disabled (e.g. when you leave then reenter the screen).
Maybe change it `m_in_drag_select && !m_automatic_selection_scroll_timer.is_active()` instead ?

VKoskiv: Wow. Evidently pushing a quick PR at 2AM on a weekday morning results in stuff like this.

That being said, stopping the scroll when re-entering the widget is expected behaviour, is it not?
I pushed a fix just now. Thanks for catching this!

VKoskiv:
Don't merge this in yet, I just caught another scrolling bug.

VKoskiv:
Well, phew! Lesson learned, double-check your commits before filing the PR. Funny enough, it worked fine when I played around with it with the 'if (false)' derpiness going on in there.

PR #1495, opened 2020-03-21
WindowServer: don't ignore mouse events after showing modal window

muscar:
This is slightly hacky, I guess.

Addresses https://github.com/SerenityOS/serenity/issues/1464.

PR #1900, opened 2020-04-21
WindowServer: Don't ignore modal events when showing menus

### LINKED ISSUE #1464: WindowServer: Mouse events ignored after activating modal window menu
When an alert is popped and you click on the left upper corner, it stops working mouse clicks and hover animations.
Screenshot_20200315_113755
Screenshot_20200315_113924
I could get out of it, pressing "Windows" key 2 times.

anpep:
PR #1495 fixes #1464 but only accounts for window menus. In File
Manager, for example, attempting to pop up the context menu on the file
name text box of the properties modal window, will result in the same
behavior.

Removing the code altogether solves the problem, altough I'm
not sure if it could have any bad implications.

I was thinking it might allow interaction with a parent window menu if
it remains open just before the popup window is shown, but I have not
seen a way to replicate this behavior.

image

### DIFF
--- a/Servers/WindowServer/MenuManager.cpp
+++ b/Servers/WindowServer/MenuManager.cpp
@@ -147,11 +147,12 @@ void MenuManager::event(Core::Event& event)
 
 void MenuManager::handle_mouse_event(MouseEvent& mouse_event)
 {
+    auto* active_window = WindowManager::the().active_window();
     bool handled_menubar_event = false;
     for_each_active_menubar_menu([&](Menu& menu) {
         if (menu.rect_in_menubar().contains(mouse_event.position())) {
             handle_menu_mouse_event(menu, mouse_event);
-            handled_menubar_event = true;
+            handled_menubar_event = !active_window || !active_window->is_modal();
             return IterationDecision::Break;
         }
         return IterationDecision::Continue;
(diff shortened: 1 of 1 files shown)


awesomekling:
This breaks the intentional "don't allow interacting with app menus while modal window is up" behavior.

If you are in e.g FileManager, and you do "create new directory", we don't want you to access the app menus until you have dismissed the popup dialog.

anpep:
> This breaks the intentional "don't allow interacting with app menus while modal window is up" behavior.
> 
> If you are in e.g FileManager, and you do "create new directory", we don't want you to access the app menus until you have dismissed the popup dialog.

Would it be a good UX to disable the app menu while a modal is being displayed?
Or maybe letting modal windows have their own menus? I see it could be useful in more complex modals such as FilePicker.

awesomekling:
> Would it be a good UX to disable the app menu while a modal is being displayed?

I think so. I borrowed it from macOS. :)

> Or maybe letting modal windows have their own menus? I see it could be useful in more complex modals such as FilePicker.

Serenity consciously chooses to be menubar-per-app rather than menubar-per-window.

anpep:
> > Would it be a good UX to disable the app menu while a modal is being displayed?
> 
> I think so. I borrowed it from macOS. :)
> 
> > Or maybe letting modal windows have their own menus? I see it could be useful in more complex modals such as FilePicker.
> 
> Serenity consciously chooses to be menubar-per-app rather than menubar-per-window.

I think I got it. I basically moved the modal check to a place I can actually verify if the mouse event target is the menubar and not just any menu ;)

awesomekling:
Yeah that seems all right :)

PR #1901, opened 2020-04-21
WindowServer: Unset previously active input window when active window changes

anpep:
This was not done previously and resulted in modal windows not being
able to accept input unless they were clicked one time if their parent
windows were the active input window.

(e.g. Properties window on FileManager would not accept input unless first clicked)


awesomekling:
:+1:

PR #1902, opened 2020-04-21
WindowServer: Allow the system menu to be open when a modal is shown

anpep:
A little heads up from my previous PR. Events going towards system menu would not be processed!
This fixes it. Also, I guess I should test a bit before making too many pull requests :upside_down_face:


awesomekling:
Oops, yeah that makes sense :D

PR #1905, opened 2020-04-21
LibJS: Add GeneratorFunction

alimpfard:
Note that this version of Generators cannot correctly evaluate `yield (yield)`, because our evaluation is granular only at the statement level.
We also do not (yet) support delegate yields (`yield* ...`)

### DIFF
--- a/Libraries/LibJS/AST.cpp
+++ b/Libraries/LibJS/AST.cpp
@@ -32,6 +32,7 @@
 #include <LibJS/Interpreter.h>
 #include <LibJS/Runtime/Array.h>
 #include <LibJS/Runtime/Error.h>
+#include <LibJS/Runtime/GeneratorFunction.h>
 #include <LibJS/Runtime/GlobalObject.h>
 #include <LibJS/Runtime/MarkedValueList.h>
 #include <LibJS/Runtime/NativeFunction.h>
@@ -49,14 +50,14 @@ Value ScopeNode::execute(Interpreter& interpreter) const
 
 Value FunctionDeclaration::execute(Interpreter& interpreter) const
 {
-    auto* function = ScriptFunction::create(interpreter.global_object(), name(), body(), parameters(), interpreter.current_environment());
+    auto* function = generator() ? GeneratorFunction::create(interpreter.global_object(), name(), body(), parameters(), interpreter.current_environment()) : ScriptFunction::create(interpreter.global_object(), name(), body(), parameters(), interpreter.current_environment());
     interpreter.set_variable(name(), function);
     return js_undefined();
 }
 
 Value FunctionExpression::execute(Interpreter& interpreter) const
 {
-    return ScriptFunction::create(interpreter.global_object(), name(), body(), parameters(), interpreter.current_environment());
+    return generator() ? GeneratorFunction::create(interpreter.global_object(), name(), body(), parameters(), interpreter.current_environment()) : ScriptFunction::create(interpreter.global_object(), name(), body(), parameters(), interpreter.current_environment());
 }
 
 Value ExpressionStatement::execute(Interpreter& interpreter) const
@@ -74,11 +75,13 @@ CallExpression::ThisAndCallee CallExpression::compute_this_and_callee(Interprete
     if (m_callee->is_member_expression()) {
         auto& member_expression = static_cast<const MemberExpression&>(*m_callee);
         auto object_value = member_expression.object().execute(interpreter);
-        if (interpreter.exception())
-            return {};
+
+        STOP_EXECUTION_IF_NEEDED(interpreter, (ThisAndCallee { js_undefined(), object_value }));
+
         auto* this_value = object_value.to_object(interpreter.heap());
-        if (interpreter.exception())
-            return {};
+
+        STOP_EXECUTION_IF_NEEDED(interpreter, (ThisAndCallee { js_undefined(), this_value }));
+
         auto callee = this_value->get(member_expression.computed_property_name(interpreter)).value_or(js_undefined());
         return { this_value, callee };
     }
@@ -88,8 +91,7 @@ CallExpression::ThisAndCallee CallExpression::compute_this_and_callee(Interprete
 Value CallExpression::execute(Interpreter& interpreter) const
 {
     auto [this_value, callee] = compute_this_and_callee(interpreter);
-    if (interpreter.exception())
-        return {};
+    STOP_EXECUTION_IF_NEEDED(interpreter, callee);
 
     ASSERT(!callee.is_empty());
 
@@ -118,11 +120,12 @@ Value CallExpression::execute(Interpreter& interpreter) const
 
     for (size_t i = 0; i < m_arguments.size(); ++i) {
         auto value = m_arguments[i].execute(interpreter);
-        if (interpreter.exception())
-            return {};
+
+        STOP_EXECUTION_IF_NEEDED(interpreter, value);
+
         arguments.append(value);
-        if (interpreter.exception())
-            return {};
+
+        STOP_EXECUTION_IF_NEEDED(interpreter, {});
     }
 
     auto& call_frame = interpreter.push_call_frame();
@@ -146,8 +149,7 @@ Value CallExpression::execute(Interpreter& interpreter) const
 
     interpreter.pop_call_frame();
 
-    if (interpreter.exception())
-        return {};
+    STOP_EXECUTION_IF_NEEDED(interpreter, {});
 
     if (is_new_expression()) {
         if (result.is_object())
@@ -157,20 +159,48 @@ Value CallExpression::execute(Interpreter& interpreter) const
     return result;
 }
 
+Value YieldExpression::execute(Interpreter& interpreter) const
+{
+    if (m_delegate) {
+        dbg() << "fixme: delegate";
+        ASSERT_NOT_REACHED();
+    }
+
+    // I'm SORRY
+    ++const_cast<YieldExpression*>(this)->m_executed_count;
+
+    if (m_executed_count % 2 == 0) {
+        // Have the expression evaluate to the argument
+        const auto& arguments = interpreter.call_frame().arguments;
+        if (arguments.size())
+            return arguments.first();
+
+        return js_undefined();
+    }
+
+    auto value = m_argument ? m_argument->execute(interpreter) : js_undefined();
+    interpreter.stop_execution(ExecutionStopReason::Yield);
+    // act like a "return"
+    interpreter.unwind(ScopeType::Function);
+    return value;
+}
+
 Value ReturnStatement::execute(Interpreter& interpreter) const
 {
     auto value = argument() ? argument()->execute(interpreter) : js_undefined();
-    if (interpreter.exception())
-        return {};
+
+    STOP_EXECUTION_IF_NEEDED(interpreter, value);
+
     interpreter.unwind(ScopeType::Function);
+    interpreter.did_return();
     return value;
 }
 
 Value IfStatement::execute(Interpreter& interpreter) const
 {
     auto predicate_result = m_predicate->execute(interpreter);
-    if (interpreter.exception())
-        return {};
+
+    STOP_EXECUTION_IF_NEEDED(interpreter, predicate_result);
 
     if (predicate_result.to_boolean())
         return interpreter.run(*m_consequent);
@@ -184,12 +214,13 @@ Value IfStatement::execute(Interpreter& interpreter) const
 Value WhileStatement::execute(Interpreter& interpreter) const
 {
     Value last_value = js_undefined();
-    while (m_test->execute(interpreter).to_boolean()) {
-        if (interpreter.exception())
-            return {};
+    while (auto test_value = m_test->execute(interpreter).to_boolean()) {
+
+        STOP_EXECUTION_IF_NEEDED(interpreter, Value(test_value));
+
         last_value = interpreter.run(*m_body);
-        if (interpreter.exception())
-            return {};
+
+        STOP_EXECUTION_IF_NEEDED(interpreter, last_value);
     }
 
     return last_value;
@@ -198,13 +229,15 @@ Value WhileStatement::execute(Interpreter& interpreter) const
 Value DoWhileStatement::execute(Interpreter& interpreter) const
 {
     Value last_value = js_undefined();
+    bool test_value { false };
     do {
-        if (interpreter.exception())
-            return {};
+        STOP_EXECUTION_IF_NEEDED(interpreter, Value(test_value));
+
         last_value = interpreter.run(*m_body);
-        if (interpreter.exception())
-            return {};
-    } while (m_test->execute(interpreter).to_boolean());
+
+        STOP_EXECUTION_IF_NEEDED(interpreter, last_value);
+
+    } while ((test_value = m_test->execute(interpreter).to_boolean()));
 
     return last_value;
 }
@@ -226,18 +259,18 @@ Value ForStatement::execute(Interpreter& interpreter) const
     Value last_value = js_undefined();
 
     if (m_init) {
-        m_init->execute(interpreter);
-        if (interpreter.exception())
-            return {};
+        auto value = m_init->execute(interpreter);
+        STOP_EXECUTION_IF_NEEDED(interpreter, value);
     }
 
     if (m_test) {
-        while (m_test->execute(interpreter).to_boolean()) {
-            if (interpreter.exception())
-                return {};
+        while (auto test_value = m_test->execute(interpreter).to_boolean()) {
+            STOP_EXECUTION_IF_NEEDED(interpreter, Value(test_value));
+
             last_value = interpreter.run(*m_body);
-            if (interpreter.exception())
-                return {};
+
+            STOP_EXECUTION_IF_NEEDED(interpreter, last_value);
+
             if (interpreter.should_unwind()) {
                 if (interpreter.should_unwind_until(ScopeType::Continuable)) {
                     interpreter.stop_unwind();
@@ -249,16 +282,17 @@ Value ForStatement::execute(Interpreter& interpreter) const
                 }
             }
             if (m_update) {
-                m_update->execute(interpreter);
-                if (interpreter.exception())
-                    return {};
+                auto value = m_update->execute(interpreter);
+
+                STOP_EXECUTION_IF_NEEDED(interpreter, value);
             }
         }
     } else {
         while (true) {
             last_value = interpreter.run(*m_body);
-            if (interpreter.exception())
-                return {};
+
+            STOP_EXECUTION_IF_NEEDED(interpreter, last_value);
+
             if (interpreter.should_unwind()) {
                 if (interpreter.should_unwind_until(ScopeType::Continuable)) {
                     interpreter.stop_unwind();
@@ -270,9 +304,9 @@ Value ForStatement::execute(Interpreter& interpreter) const
                 }
             }
             if (m_update) {
-                m_update->execute(interpreter);
-                if (interpreter.exception())
-                    return {};
+                auto value = m_update->execute(interpreter);
+
+                STOP_EXECUTION_IF_NEEDED(interpreter, value);
             }
         }
     }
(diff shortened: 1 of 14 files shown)


Libraries/LibJS/AST.h
```diff
@@ -204,6 +236,25 @@ class FunctionExpression final : public Expression
     virtual const char* class_name() const override { return "FunctionExpression"; }
 };
 
+class YieldExpression final : public Expression {
+public:
+    YieldExpression(bool delegate, RefPtr<Expression> argument)
+        : m_delegate(delegate)
+        , m_argument(move(argument))
+
```
awesomekling, line 244: Empty line here.

Libraries/LibJS/Interpreter.h
```diff
@@ -102,6 +124,8 @@ class Interpreter {
     void exit_scope(const ScopeNode&);
 
     Value call(Function*, Value this_value = {}, Optional<MarkedValueList> arguments = {});
+    bool has_returned() const { return m_returned; }
+    void returned() { m_returned = true; }
```
awesomekling, line 128: These names are too ambiguous. Suggestion: `has_returned()` and `did_return()`.

Libraries/LibJS/Makefile
```diff
@@ -26,6 +26,8 @@ OBJS = \
     Runtime/FunctionConstructor.o \
     Runtime/FunctionPrototype.o \
     Runtime/GlobalObject.o \
+    Runtime/GeneratorFunction.o \
+	Runtime/Iterator.o \
```
awesomekling, line 30: Inconsistent whitespace.

Libraries/LibJS/Runtime/GeneratorFunction.cpp
```diff
@@ -0,0 +1,129 @@
...
+
+GeneratorFunction::~GeneratorFunction()
+{
+}
+
+void GeneratorFunction::visit_children(Visitor& visitor)
+{
+    ScriptFunction::visit_children(visitor);
+    if (m_own_environment)
+        visitor.visit(m_own_environment);
```
awesomekling, line 57: No need for the null check here:
```suggestion
    visitor.visit(m_own_environment);
```

Libraries/LibJS/Runtime/GeneratorFunction.cpp
```diff
@@ -0,0 +1,129 @@
...
+    if (body().is_scope_node()) {
+        for (auto& declaration : static_cast<const ScopeNode&>(body()).variables()) {
+            for (auto& declarator : declaration.declarations()) {
+                variables.set(declarator.id().string(), { js_undefined(), DeclarationKind::Var });
+            }
+        }
+    }
+    if (variables.is_empty())
+        return m_own_environment = m_parent_environment;
+    return m_own_environment = heap().allocate<LexicalEnvironment>(move(variables), m_parent_environment);
```
awesomekling, line 79: This is duplicating way too much code from `ScriptFunction::create_environment()`.
Can we find a way to call that function here?

How about:
```cpp
m_own_environment = ScriptFunction::create_environment();
return m_own_environment;
```

Libraries/LibJS/Runtime/GeneratorFunction.cpp
```diff
@@ -0,0 +1,129 @@
...
+    auto new_body = adopt(*new BlockStatement);
+    auto* block = static_cast<ScopeNode*>(new_body);
+
+    auto& my_block = static_cast<const ScopeNode&>(body());
+
+    for (auto& node : my_block.children()) {
+        block->append(node);
+    }
+    auto iterator = heap().allocate<Iterator>(heap().allocate<GeneratorFunction>(name(), *new_body, parameters(), m_parent_environment, *prototype(), create_environment()), [&interpreter, argument_values, arguments](Object* generator_object, const Vector<Value>& next_arguments) -> Iterator::IteratorResult {
+        auto* gen = static_cast<GeneratorFunction*>(generator_object);
```
awesomekling, line 104: Please use full words like `generator`.
Also, how do we know it's safe to `static_cast` to a `GeneratorFunction*` here?

Libraries/LibJS/Interpreter.cpp
```diff
@@ -47,8 +47,47 @@ Interpreter::~Interpreter()
 {
 }
 
+Value Interpreter::destructive_run(Statement& statement, ArgumentVector arguments, ScopeType scope_type)
+{
+    m_returned = false;
```
awesomekling, line 52: What happens with this flag if we nest multiple calls to `destructive_run()`?

Libraries/LibJS/Interpreter.cpp
```diff
@@ -47,8 +47,47 @@ Interpreter::~Interpreter()
 {
 }
 
+Value Interpreter::destructive_run(Statement& statement, ArgumentVector arguments, ScopeType scope_type)
```
awesomekling, line 50: Too much copy-pasting here! Can we find a way to share this whole function with `run()` instead of making a copy?

Libraries/LibJS/Runtime/Iterator.h
```diff
@@ -0,0 +1,62 @@
...
+        bool finished { false };
+        Value value;
+    };
+
+    Iterator(Object* iterable, AK::Function<IteratorResult(Object*, const Vector<Value>&)>&& next_function);
+
+    virtual ~Iterator() override;
+
+    Object* iterable() { return m_iterable; }
+    AK::Function<IteratorResult(Object*, const Vector<Value>&)>& next_function() { return m_next_function; }
```
awesomekling, line 46: No need to qualify `AK::`

Libraries/LibJS/Runtime/Iterator.h
```diff
@@ -0,0 +1,62 @@
...
+
+private:
+    virtual void visit_children(Visitor&) override;
+    virtual const char* class_name() const override { return "Iterator"; }
+
+    static Value next(Interpreter&);
+
+    Object* m_iterable { nullptr };
+    bool m_done { false };
+    AK::Function<IteratorResult(Object*, const Vector<Value>&)> m_next_function;
```
awesomekling, line 59: No need to qualify `AK::`

Libraries/LibJS/Runtime/Iterator.h
```diff
@@ -0,0 +1,62 @@
...
+ */
+
+#pragma once
+
+#include <AK/Function.h>
+#include <LibJS/Runtime/Object.h>
+
+namespace JS {
+
+class Iterator : public Object {
```
awesomekling, line 34: Is this meant to be inherited? Otherwise, let's mark it `final`.

Libraries/LibJS/Runtime/Iterator.cpp
```diff
@@ -0,0 +1,76 @@
...
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include <LibJS/Interpreter.h>
+#include <LibJS/Runtime/Iterator.h>
+
+namespace JS {
+Iterator::Iterator(Object* iterable, AK::Function<IteratorResult(Object*, const Vector<Value>&)>&& next_function)
+
```
awesomekling, line 32: Empty line here.

Libraries/LibJS/Runtime/Iterator.cpp
```diff
@@ -0,0 +1,76 @@
...
+Value Iterator::next(Interpreter& interpreter)
+{
+    auto* this_object = interpreter.this_value().to_object(interpreter.heap());
+    if (!this_object)
+        return {};
+    if (!this_object->is_iterator())
+        return interpreter.throw_exception<TypeError>("Not an iterator");
+
+    auto* iterator = static_cast<Iterator*>(this_object);
+    auto return_value = interpreter.heap().allocate<Object>(nullptr);
```
awesomekling, line 53: I think what you want here is probably `Object::create_empty()`, otherwise it will not have the `ObjectPrototype` set.

Libraries/LibJS/Runtime/Iterator.cpp
```diff
@@ -0,0 +1,76 @@
...
+
+    if (iterator->done()) {
+        return_value->put("done", Value(true));
+        return_value->put("value", js_undefined());
+        return Value(return_value);
```
awesomekling, line 58:
```suggestion
        return return_value;
```

Libraries/LibJS/Runtime/Iterator.cpp
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    auto next_value = iterator->next_function()(iterator->iterable(), interpreter.call_frame().arguments);
+
+    return_value->put("done", Value(next_value.finished));
+    return_value->put("value", next_value.value);
+
+    return Value(return_value);
```
awesomekling, line 66:
```suggestion
    return return_value;
```

Libraries/LibJS/Runtime/Iterator.cpp
```diff
@@ -0,0 +1,76 @@
...
+}
+
+void Iterator::visit_children(Visitor& visitor)
+{
+    Object::visit_children(visitor);
+    if (m_iterable)
+        visitor.visit(m_iterable);
```
awesomekling, line 73:
```suggestion
    visitor.visit(m_iterable);
```

Libraries/LibJS/Runtime/Iterator.cpp
```diff
@@ -0,0 +1,76 @@
...
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include <LibJS/Interpreter.h>
+#include <LibJS/Runtime/Iterator.h>
+
+namespace JS {
+Iterator::Iterator(Object* iterable, AK::Function<IteratorResult(Object*, const Vector<Value>&)>&& next_function)
```
awesomekling, line 31: Please put an empty line between these two things as they are not part of a unit.
Also, no need to qualify `AK::`.

Libraries/LibJS/Runtime/GeneratorFunction.h
```diff
@@ -0,0 +1,56 @@
...
+    virtual bool is_generator_function() const final { return true; }
+    virtual const char* class_name() const override { return "GeneratorFunction"; }
+    virtual LexicalEnvironment* create_environment() override;
+    virtual void visit_children(Visitor&) override;
+
+    static Value iterator_method(Interpreter&);
+    static Value next(Interpreter&);
+
+    bool m_done { false };
+    LexicalEnvironment* m_own_environment;
```
awesomekling, line 53: Please explicitly initialize raw pointers `{ nullptr }`.

Libraries/LibJS/Runtime/GeneratorFunction.cpp
alimpfard: The `Iterator` gives you back whatever you pass as the first argument.

Libraries/LibJS/Interpreter.cpp
alimpfard: Nothing interesting, I don't think.
`destructive_run` is meant for generators, and we don't support `yield*`, so a generator cannot *both* return and yield in the same expression, in the same pass.

Libraries/LibJS/Interpreter.cpp
alimpfard: Not..really...short of making it a template anyway
The loop's range is different, so we can't just `?:` that

splitting it into parts and only keeping the loop is also rather silly.

Libraries/LibJS/Runtime/Iterator.h
alimpfard: Very need to qualify AK, otherwise it'd be a JS::Function

Libraries/LibJS/Runtime/Iterator.h
alimpfard: Same as above.

Libraries/LibJS/AST.cpp
```diff
@@ -157,20 +159,48 @@ Value CallExpression::execute(Interpreter& interpreter) const
...
 
+Value YieldExpression::execute(Interpreter& interpreter) const
+{
+    if (m_delegate) {
+        dbg() << "fixme: delegate";
+        ASSERT_NOT_REACHED();
+    }
+
+    // I'm SORRY
+    ++const_cast<YieldExpression*>(this)->m_executed_count;
```
awesomekling, line 170: It's a bit unfortunate that the AST now has state, but there's no obviously good other place to put it.

Libraries/LibJS/Runtime/Iterator.h
```diff
@@ -0,0 +1,62 @@
...
+namespace JS {
+
+class Iterator final : public Object {
+public:
+    struct IteratorResult {
+        bool finished { false };
+        Value value;
+    };
+
+    Iterator(Object* iterable, AK::Function<IteratorResult(Object*, const Vector<Value>&)>&& next_function);
```
awesomekling, line 41: Wait, is `iterable` ever `nullptr`? If not, let's use `Object&`.

Libraries/LibJS/Runtime/Iterator.h
awesomekling: Oh yeah lol :)

Libraries/LibJS/Runtime/Iterator.cpp
alimpfard: Again, `JS::Function` :^)

Libraries/LibJS/Runtime/GeneratorFunction.cpp
```diff
@@ -0,0 +1,114 @@
...
+
+    auto new_body = adopt(*new BlockStatement);
+    auto* block = static_cast<ScopeNode*>(new_body);
+
+    auto& my_block = static_cast<const ScopeNode&>(body());
+
+    for (auto& node : my_block.children()) {
+        block->append(node);
+    }
+    auto iterator = heap().allocate<Iterator>(heap().allocate<GeneratorFunction>(name(), *new_body, parameters(), m_parent_environment, *prototype(), create_environment()), [&interpreter, argument_values, arguments](Object* generator_object, const Vector<Value>& next_arguments) -> Iterator::IteratorResult {
```
awesomekling, line 88: Did you spill milk on your enter key? ;)

Libraries/LibJS/Runtime/GeneratorFunction.cpp
```diff
@@ -0,0 +1,114 @@
...
+        auto* generator = static_cast<GeneratorFunction*>(generator_object);
```
awesomekling, line 89: How do we know that `generator_object` is a `GeneratorFunction*`.
If it's always the case, can we communicate that through types instead of just blindly casting?

Libraries/LibJS/Parser.cpp
```diff
@@ -697,12 +729,19 @@ NonnullRefPtr<BlockStatement> Parser::parse_block_statement()
     return block;
 }
 
-template<typename FunctionNodeType>
+template <typename FunctionNodeType>
```
awesomekling, line 732: The project `.clang-format` file is not set up to make this change. Are you using a custom setting?

alimpfard: Is there one available? I'm using my own settings, tweaked to look like Serenity style.

Libraries/LibJS/Runtime/GeneratorFunction.cpp
alimpfard: In *this* specific callback, it's always `GeneratorFunction*`, but other users of Iterator may use any other Object.
I'm open to suggestions on how to do this more cleanly.

Libraries/LibJS/Parser.cpp
awesomekling: Certainly, https://github.com/SerenityOS/serenity/blob/master/.clang-format :)

Libraries/LibJS/Runtime/Iterator.h
alimpfard: > Okay I think you'll have to give a high-level overview of what's going on here.

Here goes.
The function is parsed into a GenratorFunction, which holds the original AST nodes.
At each call to the function, we create a new `GeneratorFunction` (this is due to GeneratorFunction being both an Iterable and an Iterator, I blame js), and then destructively go through its statements.
Once we hit a return, or run out of nodes, the generator is considered to have finished execution.

> Why do we need to duplicate the AST children into a temporary block statement and remove them as we go?

We would otherwise need to keep some sort of instruction pointer, which is something that our Interpreter does not have the facilities to handle. 

> What is the lexical environment allocation strategy?
Allocated once per call to a generator function, if needed.
reused for the entire lifetime of the generator instance

alimpfard: ...wrong place

alimpfard:
> Okay I think you'll have to give a high-level overview of what's going on here.

Here goes.
The function is parsed into a GenratorFunction, which holds the original AST nodes.
At each call to the function, we create a new `GeneratorFunction` (this is due to GeneratorFunction being both an Iterable and an Iterator, I blame js), and then destructively go through its statements.
Once we hit a return, or run out of nodes, the generator is considered to have finished execution.

> Why do we need to duplicate the AST children into a temporary block statement and remove them as we go?

We would otherwise need to keep some sort of instruction pointer, which is something that our Interpreter does not have the facilities to handle. 

> What is the lexical environment allocation strategy?

Allocated once per call to a generator function, if needed.
reused for the entire lifetime of the generator instance

Libraries/LibJS/Runtime/GeneratorFunction.cpp
alimpfard: I actually spilled milk on the entire right half of my keyboard lol

Libraries/LibJS/Parser.cpp
alimpfard: it's...right there in the root of the repository...
I feel stupid now.

alimpfard:
Thinking on it, we'd need to duplicate the LexicalEnvironment as well, I'll patch that up too :(

awesomekling:
Why do we need a separate stop-execution-reason mechanism instead of adapting the existing unwind mechanism?
If we need to swap out all the exception checks for stop-execution-reason, should this not also be done in LibWeb?

alimpfard:
> Why do we need a separate stop-execution-reason mechanism instead of adapting the existing unwind mechanism?

last I checked unwind couldn't carry a value along? 
also, we need to be able to detect both function boundaries and _generator_ function boundaries, and I did not see a way to have an unwind boundary without a separate scope (since the generator function instance needs two unwind tags, but has a single scope)

> If we need to swap out all the exception checks for stop-execution-reason, should this not also be done in LibWeb?

I don't think LibWeb checks for exceptions at all?
at least, it does not do so via `Interpreter::exception()`

awesomekling:
A lot of code implicitly checks for exceptions by checking if e.g `JS::Value::to_object()` returned an empty `JS::Value`. It usually looks like this:

```cpp
static CanvasRenderingContext2D* impl_from(JS::Interpreter& interpreter)
{
    if (!this_object)
        return nullptr;
    // FIXME: Verify that it's a CanvasRenderingContext2DWrapper somehow!
    return &static_cast<CanvasRenderingContext2DWrapper*>(this_object)->impl();
}

JS::Value CanvasRenderingContext2DWrapper::line_to(JS::Interpreter& interpreter)
{
    auto* impl = impl_from(interpreter);
    if (!impl)
        return {};
    double x = interpreter.argument(0).to_double();
    double y = interpreter.argument(1).to_double();
    impl->line_to(x, y);
    return JS::js_undefined();
}
```

alimpfard:
....wow, that's probably impossible to figure out now...at least for me.
I guess we're not getting these generators until I can find a smarter way to handle them.

This PR...your call, we could close it, or mark it as wip, or whatever else.

awesomekling:
> ....wow, that's probably impossible to figure out now...at least for me.
> I guess we're not getting these generators until I can find a smarter way to handle them.
> 
> This PR...your call, we could close it, or mark it as wip, or whatever else.

Yeah I think we need a smarter way to handle them so that it blends together with the exception checking. You can close the PR if you don't intend to work on this right now, and we can bring it back later :)

alimpfard:
PR close button goes _brrrrrrr_

PR #1906, opened 2020-04-21
LibJS: Parse while statements

linusg:


Libraries/LibJS/Tests/while-basic.js
```diff
@@ -0,0 +1,21 @@
...
+        number++;
+    }
+    assert(number === 9);
+
+    number = 0;
+    while(number < 3) number++;
+    assert(number === 3);
+
+    while (false) {
+        assertNotreached();
```
alimpfard, line 15: who's responsible for this sin against humanity?
`Notreached`

linusg: At least that shows the `while` statement working as expected :smile:

PR #1907, opened 2020-04-21
FileManager: Don't fail silently when attempting to open an inaccessible directory

### LINKED ISSUE #1890: FileManager: Handle directory access errors gracefully
I've been hacking around FileManager and noticed that when opening a non-existent directory (or an inaccessible directory) the underlying `GUI::FileSystemModel` just fails silently:

https://github.com/SerenityOS/serenity/blob/f85aa8462f2893cf6e719f7c17005ec44a1e9705/Libraries/LibGUI/FileSystemModel.cpp#L54-L65

Since this error is ignored in `FileSystemModel::Node::reify_if_needed`, there is no reliable way for the `DirectoryView` in FileManager to acknowledge and report the error.

I was thinking maybe including an `on_error` event on `FileSystemModel` could allow for this, but I'm not sure it is the right design decision. Feel free to discuss!

anpep:
This PR fixes #1890. I implemented `on_error` events on FileSystemModel and DirectoryView and replaced `on_path_change` with `on_complete`, which is only triggered if we could change the path successfully.

It would also make sense to trigger an `on_error` on DirectoryView when the currently open directory ceases to be accessible, but I could not yet figure a way to use the `watch_file` syscall to this end...

image
image

### DIFF
--- a/Applications/FileManager/main.cpp
+++ b/Applications/FileManager/main.cpp
@@ -588,6 +588,14 @@ int run_in_windowed_mode(RefPtr<Core::ConfigFile> config, String initial_locatio
         go_back_action->set_enabled(directory_view.path_history_position() > 0);
     };
 
+    directory_view.on_error = [&](int, const char* error_string, bool quit) {
+        auto error_message = String::format("Could not read directory: %s", error_string);
+        GUI::MessageBox::show(error_message, "File Manager", GUI::MessageBox::Type::Error);
+
+        if (quit)
+            exit(1);
+    };
+
     directory_view.on_status_message = [&](const StringView& message) {
         statusbar.set_text(message);
     };
--- a/Applications/FileManager/main.cpp
+++ b/Applications/FileManager/main.cpp
@@ -593,7 +593,7 @@ int run_in_windowed_mode(RefPtr<Core::ConfigFile> config, String initial_locatio
         GUI::MessageBox::show(error_message, "File Manager", GUI::MessageBox::Type::Error);
 
         if (quit)
-            exit(1);
+            app.quit(1);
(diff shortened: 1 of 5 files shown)


Applications/FileManager/main.cpp
```diff
@@ -588,6 +588,14 @@ int run_in_windowed_mode(RefPtr<Core::ConfigFile> config, String initial_locatio
         go_back_action->set_enabled(directory_view.path_history_position() > 0);
     };
 
+    directory_view.on_error = [&](int, const char* error_string, bool quit) {
+        auto error_message = String::format("Could not read directory: %s", error_string);
+        GUI::MessageBox::show(error_message, "File Manager", GUI::MessageBox::Type::Error);
+
+        if (quit)
+            exit(1);
```
awesomekling, line 596: To cleanly exit the application, let's use `GUI::Application::quit()`:

```suggestion
            app.quit(1);
```

anpep: That didn't work. Moreover, `GUI::Application::the().quit(1)` does not either.

alimpfard: when in doubt, global OwnPtr<GUI::Application> :^)

*braces for @awesomekling killing me*

awesomekling:
Oops, I merged this blindly while in the middle of something else. :(

anpep:
Sorry about that! Let's revert this PR for now, I'll look why GUI:: Application::quit() does not terminate the event loop when I have some time later

awesomekling:
Nah it's okay the way it is. I fixed the build. We can look into the quit() issue separately :)

PR #1909, opened 2020-04-21
LibJS: Implement basic for..in and for..of loops

linusg:
Draft for now, still needs a bit of work.

TODO:

- [x] Make object properties non-{writable,enumerable,configurable} unless specified otherwise (https://tc39.es/ecma262/#table-default-attribute-values)
- [x] Implement array-like indexing of strings - i.e. "foo" has the properties "0", "1" and "2"
- [x] go through the code with a fresh mind and clean that mess up (well, some of it at least)

I'd prefer to _not_ implement the iterable & iterator protocol in this PR and wing it instead, for now.

### DIFF
--- a/Libraries/LibJS/AST.cpp
+++ b/Libraries/LibJS/AST.cpp
@@ -322,6 +322,128 @@ Value ForStatement::execute(Interpreter& interpreter) const
     return last_value;
 }
 
+static FlyString variable_from_for_declaration(Interpreter& interpreter, NonnullRefPtr<ASTNode> node, RefPtr<BlockStatement> wrapper)
+{
+    FlyString variable_name;
+    if (node->is_variable_declaration()) {
+        auto* variable_declaration = static_cast<const VariableDeclaration*>(node.ptr());
+        ASSERT(!variable_declaration->declarations().is_empty());
+        if (variable_declaration->declaration_kind() != DeclarationKind::Var) {
+            wrapper = create_ast_node<BlockStatement>();
+            interpreter.enter_scope(*wrapper, {}, ScopeType::Block);
+        }
+        variable_declaration->execute(interpreter);
+        variable_name = variable_declaration->declarations().first().id().string();
+    } else if (node->is_identifier()) {
+        variable_name = static_cast<const Identifier&>(*node).string();
+    } else {
+        ASSERT_NOT_REACHED();
+    }
+    return variable_name;
+}
+
+Value ForInStatement::execute(Interpreter& interpreter) const
+{
+    if (!m_lhs->is_variable_declaration() && !m_lhs->is_identifier()) {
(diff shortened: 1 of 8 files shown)


Libraries/LibJS/AST.cpp
```diff
@@ -280,6 +282,139 @@ Value ForStatement::execute(Interpreter& interpreter) const
...
+    } else {
+        ASSERT_NOT_REACHED();
+    }
+    return variable_name;
+}
+
+Value ForInStatement::execute(Interpreter& interpreter) const
+{
+    if (!m_lhs->is_variable_declaration() && !m_lhs->is_identifier()) {
+        // Apparently this is legal in some cases, e.g. for (a.b in c) {}
```
alimpfard, line 308: Seems like it's taken as a assignment?
```js
for (a.b in c) stmt
```
->
```js
for (let _t in c) { a.b = _t; stmt; }
```

linusg: Mystery solved, thanks!

Confirmed using an object setter function in a mature:tm: engine:

```js
var o = {
    _foo: undefined,
    get foo() {
        console.log("foo was accessed, its value is " + this._foo);
    },
    set foo(value) {
        this._foo = value;
        console.log("foo was set to " + value);
    }
}
for (o.foo in "hello") {
    o.foo; 
}
```

```
foo was set to 0
foo was accessed, its value is 0
foo was set to 1
foo was accessed, its value is 1
foo was set to 2
foo was accessed, its value is 2
foo was set to 3
foo was accessed, its value is 3
foo was set to 4
foo was accessed, its value is 4
```

Libraries/LibJS/Parser.cpp
```diff
@@ -1121,6 +1139,9 @@ void Parser::consume_or_insert_semicolon()
     // ...token is a closing curly brace
     if (match(TokenType::CurlyClose))
         return;
+    // ...token is "in" or "of"
+    if (match(TokenType::In) || match(TokenType::Of))
+        return;
```
sunverwerth, line 1144: Is this mentioned somewhere in the ES spec? This function is supposed to adhere strictly to the semicolon insertion rules.

linusg: I doubt it, this was more of a workaround to re-use the existing variable declaration parsing (`parse_variable_declaration()`) which tries to `consume_or_insert_semicolon()`, and the is none... But you're right, I need to think of a better solution.

linusg:
@awesomekling let's say this is ready for review - it's far from perfect and can't parse & run _every_ possible for..in/for..of loop, but I really don't want to have the PR around for much longer.

awesomekling:
Great work! Like everything else, we'll improve it incrementally as we go :)

PR #1913, opened 2020-04-21
LibJS: Fix Array.prototype.indexOf fromIndex negative

kessejones:
If negative fromIndex considers displacement from the end of the array
without decreasing 1 of de size.


awesomekling:
Please add a test for this fix :)

kessejones:
Tests were added. :)

PR #1916, opened 2020-04-22
LibLine: Do not ignore ^C and do not finish() after interrupt

alimpfard:
I am not sure how this behaviour was left unnoticed for so long lol


awesomekling:
Indeed :D

PR #1917, opened 2020-04-22
Kernel: Basic core dump memory files

Quaker762:
The Kernel now has a way to generate core dump files when a process
terminates abnormally or outright crashes. It can currently dump
all virtual regions to a file, which can then be opened in GDB to
inspect what the procecss did before it crashed. Note that this
currently doesn't contain code to generate `PT_NOTE` headers,
so GDB functionality doesn't work (though memory regions can be
viewed for at least _some_ debugging functionality)

Here's an image of a generated file from `crash`

Image

This is going to be part of a bigger system that gives a popup window (similar to Ubuntu) for when a program crashes, so the user is actually informed that the program has terminated abnormally, similar to `Apport` or `Dr. Konqi`

### DIFF
--- a/Kernel/Process.cpp
+++ b/Kernel/Process.cpp
@@ -1540,15 +1540,115 @@ void Process::crash(int signal, u32 eip)
     } else {
         dbg() << "\033[31;1m" << String::format("%p", eip) << "  (?)\033[0m\n";
     }
+
     dump_backtrace();
     m_termination_signal = signal;
     dump_regions();
     ASSERT(is_ring3());
     die();
+
     // We can not return from here, as there is nowhere
     // to unwind to, so die right away.
     Thread::current->die_if_needed();
     ASSERT_NOT_REACHED();
+}
+
+void Process::create_core_dump()
+{
+    Elf32_Ehdr elf_file_header;
+    size_t offset = 0;
+
+    auto pname = String(m_name);
+    auto fd_or_error = VFS::the().open(String::format("%s_%u.core", pname.characters(), RTC::now()), O_CREAT | O_WRONLY | O_EXCL, 0400, current_directory(), UidAndGid { m_uid, m_gid });
+    auto& fd = fd_or_error.value();
+
+    elf_file_header.e_ident[EI_MAG0] = 0x7f;
+    elf_file_header.e_ident[EI_MAG1] = 'E';
+    elf_file_header.e_ident[EI_MAG2] = 'L';
+    elf_file_header.e_ident[EI_MAG3] = 'F';
+    elf_file_header.e_ident[EI_CLASS] = ELFCLASS32;
+    elf_file_header.e_ident[EI_DATA] = ELFDATA2LSB;
+    elf_file_header.e_ident[EI_VERSION] = EV_CURRENT;
+    elf_file_header.e_ident[EI_OSABI] = 0; // ELFOSABI_NONE
+    elf_file_header.e_ident[EI_ABIVERSION] = 0;
+    elf_file_header.e_ident[EI_PAD + 1] = 0;
+    elf_file_header.e_ident[EI_PAD + 2] = 0;
+    elf_file_header.e_ident[EI_PAD + 3] = 0;
+    elf_file_header.e_ident[EI_PAD + 4] = 0;
+    elf_file_header.e_ident[EI_PAD + 5] = 0;
+    elf_file_header.e_ident[EI_PAD + 6] = 0;
+    elf_file_header.e_ident[EI_NIDENT] = 16;
+    elf_file_header.e_type = ET_CORE;
+    elf_file_header.e_machine = EM_386;
+    elf_file_header.e_version = 1;
+    elf_file_header.e_entry = 0;
+    elf_file_header.e_phoff = sizeof(Elf32_Ehdr);
+    elf_file_header.e_shoff = 0;
+    elf_file_header.e_flags = 0;
+    elf_file_header.e_ehsize = sizeof(Elf32_Ehdr);
+    elf_file_header.e_shentsize = sizeof(Elf32_Shdr);
+    elf_file_header.e_phentsize = sizeof(Elf32_Phdr);
+    elf_file_header.e_phnum = m_regions.size() + 1; // +1 for NOTE section
+    elf_file_header.e_shnum = 0;
+    elf_file_header.e_shstrndx = SHN_UNDEF;
+
+    fd->write(reinterpret_cast<uint8_t*>(&elf_file_header), sizeof(Elf32_Ehdr));
+
+    // TODO: We need to generate appropriate PT_NOTE sections to store
+    // register information and process status stuff here. This is done
+    // by creating a PT_NOTE Phdr for every threa and filling
+    // in the appropriate fields.
+
+    // Dump all regions to file
+    for (auto& region : m_regions) {
+        Elf32_Phdr phdr;
+        auto& vmobj = region.vmobject();
+
+        phdr.p_type = PT_LOAD;
+        phdr.p_offset = offset;
+        phdr.p_vaddr = reinterpret_cast<uint32_t>(region.vaddr().as_ptr());
+        phdr.p_paddr = 0;
+
+        size_t mapped_page_count = 0;
+        for (size_t page_num = 0; page_num < region.page_count(); page_num++) {
+            PhysicalPage* page = vmobj.physical_pages()[page_num];
+            if (!page)
+                continue;
+
+            mapped_page_count++;
+        }
+
+        phdr.p_filesz = mapped_page_count * PAGE_SIZE;
+        phdr.p_memsz = region.page_count() * PAGE_SIZE;
+        phdr.p_align = 0;
+
+        phdr.p_flags = region.is_readable() ? PF_R : 0;
+        if (region.is_writable())
+            phdr.p_flags |= PF_W;
+        if (region.is_executable())
+            phdr.p_flags |= PF_X;
+
+        offset += phdr.p_filesz;
+
+        fd->write(reinterpret_cast<uint8_t*>(&phdr), sizeof(Elf32_Phdr));
+    }
+
+    // Let's actually dump all the memory into the core file
+    for (auto& region : m_regions) {
+        // Remap the region so that it's readable by us.
+        region.set_readable(true);
+        region.remap();
+
+        // Now let's walk the physical pages and see which ones are actually mapped.
+        auto& vmobj = region.vmobject();
+        for (size_t i = 0; i < region.page_count(); i++) {
+            PhysicalPage* page = vmobj.physical_pages()[region.first_page_index() + i];
+            if (!page)
+                continue;
+
+            fd->write(reinterpret_cast<uint8_t*>((region.vaddr().as_ptr() + (i * PAGE_SIZE))), PAGE_SIZE);
+        }
+    }
 }
 
 Process* Process::from_pid(pid_t pid)
@@ -3080,6 +3180,14 @@ void Process::finalize()
     dbg() << "Finalizing process " << *this;
 #endif
 
+    dbg() << m_should_dump_core;
+    if (m_should_dump_core) {
+        ProcessPagingScope scope(*this);
+        dbg() << "Creating dump of physical memory...";
+        create_core_dump();
+        dbg() << "Done!";
+    }
+
     if (m_perf_event_buffer) {
         auto description_or_error = VFS::the().open(String::format("perfcore.%d", m_pid), O_CREAT | O_EXCL, 0400, current_directory(), UidAndGid { m_uid, m_gid });
         if (!description_or_error.is_error()) {
@@ -3088,7 +3196,6 @@ void Process::finalize()
             description->write(json.data(), json.size());
         }
     }
-
     m_fds.clear();
     m_tty = nullptr;
     m_executable = nullptr;
--- a/Kernel/Process.h
+++ b/Kernel/Process.h
@@ -312,6 +312,7 @@ public:
 
     [[noreturn]] void crash(int signal, u32 eip);
     [[nodiscard]] static siginfo_t reap(Process&);
+    void create_core_dump();
 
     const TTY* tty() const { return m_tty; }
     void set_tty(TTY*);
@@ -382,6 +383,7 @@ public:
 
     void terminate_due_to_signal(u8 signal);
     void send_signal(u8, Process* sender);
+    void set_dump_core(bool enable) { m_should_dump_core = enable; }
 
     u16 thread_count() const { return m_thread_count; }
 
@@ -482,6 +484,7 @@ private:
     u8 m_termination_status { 0 };
     u8 m_termination_signal { 0 };
     u16 m_thread_count { 0 };
+    bool m_should_dump_core { false };
 
     bool m_dead { false };
     bool m_profiling { false };
--- a/Kernel/Thread.cpp
+++ b/Kernel/Thread.cpp
@@ -541,6 +541,8 @@ ShouldUnblockThread Thread::dispatch_signal(u8 signal)
             set_state(Stopped);
             return ShouldUnblockThread::No;
         case DefaultSignalAction::DumpCore:
+            m_process.set_dump_core(true);
+            dbg() << "Yikes!";
             process().for_each_thread([](auto& thread) {
                 thread.set_dump_backtrace_on_finalization();
                 return IterationDecision::Continue;
@@ -575,8 +577,7 @@ ShouldUnblockThread Thread::dispatch_signal(u8 signal)
 
     m_signal_mask |= new_signal_mask;
 
-    auto setup_stack = [&]<typename ThreadState>(ThreadState state, u32 * stack)
-    {
+    auto setup_stack = [&]<typename ThreadState>(ThreadState state, u32* stack) {
         u32 old_esp = *stack;
         u32 ret_eip = state.eip;
         u32 ret_eflags = state.eflags;


Kernel/Process.cpp
```diff
@@ -1551,6 +1556,127 @@ void Process::crash(int signal, u32 eip)
...
+    elf_file_header.e_ehsize = sizeof(Elf32_Ehdr);
+    elf_file_header.e_shentsize = sizeof(Elf32_Shdr);
+    elf_file_header.e_phentsize = sizeof(Elf32_Phdr);
+    elf_file_header.e_phnum = m_regions.size() + 1; // +1 for NOTE section
+    elf_file_header.e_shnum = 0;
+    elf_file_header.e_shstrndx = SHN_UNDEF;
+
+    fd->write(reinterpret_cast<uint8_t*>(&elf_file_header), sizeof(Elf32_Ehdr));
+
+    // TODO: We need to generate appropriate PT_NOTE sections to store
```
Quaker762, line 1601: @ADKaster Do you know how these are supposed to be generated?? The format is really really bizarre and seems to be text based????

Kernel/Process.cpp
```diff
@@ -1540,6 +1540,11 @@ void Process::crash(int signal, u32 eip)
     } else {
         dbg() << "\033[31;1m" << String::format("%p", eip) << "  (?)\033[0m\n";
     }
+
+    dbg() << "Creating dump of physical memory...";
+    create_core_dump();
```
Quaker762, line 1545: @awesomekling Would you prefer it if I made this some kind of switch from a configuration file (or a Kernel command line switch???)

Kernel/Process.cpp
ADKaster: Well, looking [here] (https://stackoverflow.com/a/38427225) it seems as though what you actually put in the PT_NOTE is not really part of the elf spec? But it looks like Linux at least puts some structures containing the register dump and signal info and generic process info (uptime, user info, crash time etc). I've not looked into that kind of header much yet though.

ADKaster: Maybe I should read the PR before commenting :D. You mentioned those things below. Looks like PT_NOTEs are just generic info? But specific ones are used for core dumps on Linux and the BSDs and Solaris. 

Seems there's a special entry in the section header table for SHT_NOTE that points to all the notes, and each note is a type, a name, and a pointer to some offset in the file that contains a note specific structure. https://www.gabriel.urdhr.fr/2015/05/29/core-file/
Looks like the actual structures and "note types" used are extremely OS specific.. I'd be surprised if it was trivial to get gdb to open the core files. Might be more interesting (and hair pulling exciting) to get LibDebug to know how to parse an ET_CORE file. @itamar8910

Quaker762: I see. I might have to have a peruse of the GDB source code to see where it tries to read it from then. I imagine there's some kind of tag for it (I think it's NT_PRSTATUS).

Kernel/Process.cpp
```diff
@@ -1551,6 +1556,127 @@ void Process::crash(int signal, u32 eip)
...
+        // because it will mess up the size of the file. GDB can detect, and
+        // WILL complain about this.
+        size_t mapped_page_count = 0;
+        for (size_t page_num = 0; page_num < vmobj.page_count(); page_num++) {
+            PhysicalPage* page = vmobj.physical_pages()[page_num];
+            if (!page)
+                continue;
+
+            mapped_page_count++; // Only consider mapped pages (so we don't go past the size of `region`)
+        }
```
awesomekling, line 1627: This code is a bit wrong. A `Region` is a wrapper around some or all of the pages in a `VMObject`.
If you `mmap()` some memory and then `unmap()` half of it, the `VMObject` may still be the original size, while you only have a half-sized `Region` now.

The correct number of pages to dump is the number of pages in the `Region`. So `Region::page_count()`.

Whether a page is backed by a physical page or not only matters when it comes to grabbing its data. (For unmapped pages, we can simply memset the output with zeroes.)

Kernel/Process.cpp
```diff
@@ -1551,6 +1556,127 @@ void Process::crash(int signal, u32 eip)
...
+
+        // FIXME: memsz should be `region.page_count() * PAGE_SIZE` according
+        // to the ELF spec, but that causes readelf to issue a warning about filesz
+        // being greater than memsz!!??? How is that even possible??
```
awesomekling, line 1631: I suspect this problem will go away if you fix the above. :)

Kernel/Process.cpp
```diff
@@ -1551,6 +1556,127 @@ void Process::crash(int signal, u32 eip)
...
+        fd->write(reinterpret_cast<uint8_t*>(&phdr), sizeof(Elf32_Phdr));
+    }
+
+    // Get all the threads so we can write stuff to disk (we can't in this lambda because
+    // interrupts are disabled)
+    Vector<Thread*> threads;
+    current->for_each_thread([&](Thread& thread) {
+        threads.append(&thread);
+        return IterationDecision::Continue;
+    });
```
awesomekling, line 1653: Won't be needed if done during finalization.

Kernel/Process.cpp
```diff
@@ -1551,6 +1556,127 @@ void Process::crash(int signal, u32 eip)
...
+
+    for (size_t i = 0; i < threads.size(); i++) {
+    }
```
awesomekling, line 1656: No-op loop here.

Kernel/Process.cpp
```diff
@@ -1551,6 +1556,127 @@ void Process::crash(int signal, u32 eip)
...
+        VMObject& vmobj = region.vmobject();
+        size_t mapped_page_count = 0;
+        for (size_t page_num = 0; page_num < vmobj.page_count(); page_num++) {
+            PhysicalPage* page = vmobj.physical_pages()[page_num];
+            if (!page)
+                continue;
+
+            fd->write(reinterpret_cast<uint8_t*>((region.vaddr().as_ptr() + (mapped_page_count * PAGE_SIZE))), PAGE_SIZE);
+            mapped_page_count++; // Only consider mapped pages (so we don't go past the size of `region`)
+        }
```
awesomekling, line 1674: This code is dumping whichever random pages in the `VMObject` underneath the `Region` happen to be mapped.
It should dump the exact set of pages in the `Region`. Remember that a `Region` is just a *slice* of a `VMObject`.

You can iterate from `0` to `region.page_count()`. Each page is the `vmobject.physical_pages()[region.first_page_index() + i]`

Kernel/Process.cpp
```diff
@@ -1551,6 +1556,127 @@ void Process::crash(int signal, u32 eip)
...
+    }
+    fd->close();
```
awesomekling, line 1676: This doesn't do anything. You can leave it out and `~FileDescription()` will take care of everything.

Kernel/Process.cpp
```diff
@@ -1551,6 +1556,127 @@ void Process::crash(int signal, u32 eip)
...
+    cli(); // Disable interrupts again
```
awesomekling, line 1677: Won't be needed when we move to the finalizer.

Kernel/Process.cpp
bugaevc: I believe it should be fine to dump setuid programs as long as we ensure the core file is owned by root (or whoever owns the process). Now, of course there's the issue that setuid is not only for root — in effect, a single process runs as multiple users, and it's not always that one of them is strictly more privileged than the other.

Kernel/Process.cpp
```diff
@@ -1551,6 +1556,127 @@ void Process::crash(int signal, u32 eip)
...
 }
 
+void Process::create_core_dump()
+{
+    Elf32_Ehdr elf_file_header;
+    size_t offset = 0;
+
+    sti(); // Re-enable interrupts so we don't shit the bed (we need I/O)!!!
+    auto pname = String(m_name);
+    auto fd_or_error = VFS::the().open(String(String::format("%s_%u.core", pname.characters(), RTC::now())), O_CREAT, 0400, current_directory(), UidAndGid { m_uid, m_gid });
```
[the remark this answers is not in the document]
bugaevc: Just `O_CREAT` is not enough, also pass `O_WRITE`. And probably `O_EXCL` — you wouldn't want to incidentally overwrite an existing file.

Kernel/Process.cpp
```diff
@@ -1551,6 +1556,127 @@ void Process::crash(int signal, u32 eip)
...
+
+    fd->write(reinterpret_cast<uint8_t*>(&elf_file_header), sizeof(Elf32_Ehdr));
+
+    // TODO: We need to generate appropriate PT_NOTE sections to store
+    // register information and process status stuff here. This is done
+    // by creating a PT_NOTE Phdr for every threa and filling
+    // in the appropriate fields.
+
+    // Dump all regions to file
+    for (Region& region : m_regions) {
```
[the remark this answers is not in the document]
bugaevc: Or just use `auto&`

Kernel/Process.cpp
bugaevc: It kind of does, it's supposed to flush enough caches to check for errors. It doesn't do it today, but checking for success at call site (so here) wouldn't hurt. If the call site was to check, that is.

Kernel/Process.cpp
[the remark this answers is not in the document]
awesomekling: `auto&` would be my personal choice too ;)

Kernel/Process.cpp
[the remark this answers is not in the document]
awesomekling: Good points @bugaevc!

Kernel/Process.cpp
awesomekling: This call site does not communicate anything to userspace anyway, so all we could do is log some whiny message?

bugaevc: Yes.

Kernel/Process.cpp
```diff
@@ -1551,6 +1556,127 @@ void Process::crash(int signal, u32 eip)
     ASSERT_NOT_REACHED();
 }
 
+void Process::create_core_dump()
+{
+    Elf32_Ehdr elf_file_header;
+    size_t offset = 0;
+
+    sti(); // Re-enable interrupts so we don't shit the bed (we need I/O)!!!
```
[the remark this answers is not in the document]
Quaker762: I see. So the call to `create_core_dump()` would be in `Process::finalize()` instead? I'll move it there.

Kernel/Process.cpp
Quaker762: So this could cause some kind of vulnerable or undefined state?

Kernel/Process.cpp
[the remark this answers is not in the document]
Quaker762: > Just `O_CREAT` is not enough, also pass `O_WRITE`. And probably `O_EXCL` — you wouldn't want to incidentally overwrite an existing file.

Ahh, I totally missed that one whoops.

> Also, you are wrapping String::format() in a String() for no reason :)

This is more egregious after the amount of times @bugaevc has smacked me over this particular function :laughing:

Kernel/Process.cpp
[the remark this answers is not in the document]
Quaker762: I'm a bit wary of using the `auto` keyword as it's bitten me in the arse in the past and the fact that I feel it generally abstracts things unnecessarily. Would you prefer it as a stylistic choice to be more consistent?

Kernel/Process.cpp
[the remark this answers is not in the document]
Quaker762: @awesomekling Interesting, this is occurring with the dumping code in `Process:finalize()`

```
[FinalizerTask(3:3)]: CRASH: Page Fault. Ring 0.
[FinalizerTask(3:3)]: exception code: 0000 (isr: 0000
[FinalizerTask(3:3)]:   pc=0008:c012d748 flags=0282
[FinalizerTask(3:3)]:  stk=0010:c084cc24
[FinalizerTask(3:3)]:   ds=0010 es=0010 fs=0010 gs=0000
[FinalizerTask(3:3)]: eax=c06fed98 ebx=c010d00c ecx=00000e6c edx=c084ccb4
[FinalizerTask(3:3)]: ebp=c084ccdc esp=c084cc24 esi=08048000 edi=c06fed98
[FinalizerTask(3:3)]: cr0=8001001b cr2=08048000 cr3=00841000 cr4=003006e4
[FinalizerTask(3:3)]: Oh shit, we've crashed in ring 0 :(
[FinalizerTask(3:3)]: 0xc0119d3f  Kernel::handle_crash(Kernel::RegisterState&, char const*, int) +2443
[FinalizerTask(3:3)]: 0xc011ad94  page_fault_handler +820
[FinalizerTask(3:3)]: 0xc01187d7  page_fault_asm_entry +22
[FinalizerTask(3:3)]: 0xc0131e0e  Kernel::InodeFile::write(Kernel::FileDescription&, unsigned long, unsigned char const*, int) +56
[FinalizerTask(3:3)]: 0xc012faa2  Kernel::FileDescription::write(unsigned char const*, int) +88
[FinalizerTask(3:3)]: 0xc015397d  Kernel::Process::create_core_dump() +1331
[FinalizerTask(3:3)]: 0xc015531d  Kernel::Process::finalize() +143
[FinalizerTask(3:3)]: 0xc016cc99  Kernel::Thread::finalize_dying_threads() +441
[FinalizerTask(3:3)]: 0xc0167594  Kernel::FinalizerTask::spawn() +14
```

Kernel/Process.cpp
Quaker762: Ah that makes sense, that'd explain what GDB was complaining about. So run the loop over `Region::page_count()` and then if the page isn't mapped (`!page`) just output a bunch of zeroes instead?

Kernel/Process.cpp
bugaevc: For example, `su` may have user's password in memory when it crashes and we wouldn't want `anon` to be able to inspect the core dump and get the password from there.

Kernel/Process.cpp
bugaevc: > if the page isn't mapped (`!page`) just output a bunch of zeroes instead?

`!page` is not about whether the (virtual) page is mapped, it's about whether it's backed by a physical page. It may not have a backing physical page in case it was purged, or never faulted in. In the former case it's fine to dump it as a bunch of zeros, but in the latter case we'd have to fault in the actual page and dump that.

Kernel/Process.cpp
Quaker762: Ahhh, yep, I thought as much.

Kernel/Process.cpp
Quaker762: These can surely just be ignored then? The whole point of the core dump is really to only dump whatever is actually backed by a physical page at the time of the crash if what I've read is correct. `FileSiz` is the size of the data that is immediately available when something goes wrong.

bugaevc: I'm not familiar with how it's stored in the ELF, but I would expect to see all the pages that were logically mapped (mmaped) for the process (even if they were paged out / never paged in).

Quaker762:
@awesomekling Okay this should be a bit better, still not sure if I'm being nughty w.r.t the memory so let me know if it's still wrong

Kernel/Process.cpp
```diff
@@ -1540,17 +1540,117 @@ void Process::crash(int signal, u32 eip)
...
+        region.remap();
+
+        // Now let's walk the physical pages and see which ones are actually mapped.
+        auto& vmobj = region.vmobject();
+        for (size_t i = 0; i < region.page_count(); i++) {
+            PhysicalPage* page = vmobj.physical_pages()[region.first_page_index() + i];
+            if (!page)
+                continue;
+
+            fd->write(reinterpret_cast<uint8_t*>((region.vaddr().as_ptr() + (i * PAGE_SIZE))), PAGE_SIZE);
```
ADKaster, line 1649: It kind of looks like this is assuming that a region will always be mapped from it's first page to some page towards the middle? I'm not sure this design will reconstruct the process's memory properly if you have
```
[ mapped | not mapped | mapped ]
```
?
The mapped physical pages would be copied into the file adjacent to each other. Do you have to create additional PT_LOAD segments for regions like this? Maybe I'm overthinking it.

ADKaster: Looking at a readelf -aW of [this core dump](https://pastebin.com/Xjn6YRgN), It kind of looks like some of the PT_LOAD segments overlap? Not sure what I'm looking at with the two R E load headers where the file size is much smaller than the memory size..

Quaker762: > It kind of looks like this is assuming that a region will always be mapped from it's first page to some page towards the middle? I'm not sure this design will reconstruct the process's memory properly if you have
> 
> ```
> [ mapped | not mapped | mapped ]
> ```
> 
> ?
> The mapped physical pages would be copied into the file adjacent to each other. Do you have to create additional PT_LOAD segments for regions like this? Maybe I'm overthinking it.

I think this should be taken into account with regards to how memory is supposed to be copied, but I'll have to reread the spec. Actually, I'm not 100% sure how the core files are supposed to handle pages that are unmapped, I was under the impression whatever _was_ mapped was just dumped into a `PT_LOAD` section, skipping over anything that's not currently mapped. Any ideas?

Also sorry for the horribly late reply on this, uni really kicked my ass the past month.

Quaker762:
@awesomekling Has something changed with regards to signaling/finalization? It seems this block now only works with `crash -a`

```
        case DefaultSignalAction::DumpCore:
            m_process.set_dump_core(true);
            dbg() << "Yikes!";
            process().for_each_thread([](auto& thread) {
                thread.set_dump_backtrace_on_finalization();
            });
            [[fallthrough]];
```

Quaker762:
Going to close this one. It's way too stale now, especially because signaling etc has changed so much. I'll have to start again :)

PR #1924, opened 2020-04-22
LibCore: read_bool_entry parse "true" / "false" strings in config files

bcoles:
`read_bool_entry()` can now interpret both integers (1 or 0) and
Boolean strings ("true" or "false") in configuration files.

All values other than "1" or "true" are considered false.


awesomekling:
Great!

PR #1925, opened 2020-04-22
LibWeb: Add XMLHttpRequest.readyState and constants

linusg:


Libraries/LibWeb/Bindings/XMLHttpRequestPrototype.cpp
```diff
@@ -40,7 +40,14 @@ XMLHttpRequestPrototype::XMLHttpRequestPrototype()
...
     put_native_function("open", open, 2);
     put_native_function("send", send, 0);
+    put_native_property("readyState", ready_state_getter, nullptr);
     put_native_property("responseText", response_text_getter, nullptr);
+
+    put("UNSENT", JS::Value((i32)XMLHttpRequest::ReadyState::Unsent));
+    put("OPENED", JS::Value((i32)XMLHttpRequest::ReadyState::Opened));
+    put("HEADERS_RECEIVED", JS::Value((i32)XMLHttpRequest::ReadyState::HeadersReceived));
+    put("LOADING", JS::Value((i32)XMLHttpRequest::ReadyState::Loading));
+    put("DONE", JS::Value((i32)XMLHttpRequest::ReadyState::Done));
```
linusg, line 50: Apparently these have to exist both on the constructor and prototype, presumably to make this possible:

```js
const xhr = new XMLHttpRequest();
if (xhr.readyState === xhr.DONE) ...
if (xhr.readyState === XMLHttpRequest.DONE) ...
```

Libraries/LibWeb/DOM/XMLHttpRequest.cpp
```diff
@@ -46,6 +46,12 @@ XMLHttpRequest::~XMLHttpRequest()
 {
 }
 
+void XMLHttpRequest::set_ready_state(ReadyState ready_state)
+{
+    // FIXME: call onreadystatechange once we have that
+    m_ready_state = ready_state;
+}
```
linusg, line 53: A setter function for this might seem overkill _right now_ (as we could just `m_ready_state = ...`) but considering we need to call a callback in the future this made sense to me.

awesomekling: That's fine :)

PR #1927, opened 2020-04-22
IRCClient: Connect to IRC server URL specified in command line argument

bcoles:
The IRCClient application can now connect to a specified IRC server using
a URL with `irc://` protocol as a command line argument.

If the URL is invalid, the application will report a relevant error message to `stderr` and exit.

If a URL is not supplied, the application will fall back to the server connection details specified in the user's `IRCClient.ini` configuration file. If the configuration file does not contain connection details, the user is prompted to enter a server address. (This is the current behavior, unchanged by this PR.)

IRCClient

Using a URL as a command line argument is inline with the current development direction, such as support for `file://` (`FileManager`, `SoundPlayer`, `QuickShow`) and `http://` (`Browser`) URLs elsewhere.

In my opinion, use of URI handlers fits with existing operating system design, and fits with "the essence of Serenity": putting a modern spin on a classic Unix thing.

As for "how is it better", I offer the following exaggeration for comparisson:

```
IRCClient ircs://irc.local.tld:6697
# versus
IRCClient --connect-to-this-server irc.local.tld --on-this-port 6697 --and-also-use-ssl
```

Note: Secure IRC `ircs` is not yet supported due to lack of SSL/TLS libraries.

Eventually this can be reworked to also allow `/#channel-name` in the URL path.

I toyed with handling the `server` and `port` in the `IRCAppWindow`, rather than `IRCClient`, however the latter makes mores sense. In the future, we may wish to support instantiating multiple `IRCClient` objects to support connecting to multiple servers, or handling changing connections from one server to another within one `IRCAppWindow`.

This is also a step towards decoupling the hard-coded connection settings in the configuration file, allowing for cleaner instantiation of new `IRCClient` objects.

### DIFF
--- a/Applications/IRCClient/main.cpp
+++ b/Applications/IRCClient/main.cpp
@@ -32,7 +32,7 @@
 
 int main(int argc, char** argv)
 {
-    if (pledge("stdio inet dns unix shared_buffer cpath rpath fattr", nullptr) < 0) {
+    if (pledge("stdio inet dns unix shared_buffer cpath rpath fattr wpath cpath", nullptr) < 0) {
         perror("pledge");
         return 1;
     }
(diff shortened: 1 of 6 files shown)


Applications/IRCClient/main.cpp
```diff
@@ -44,12 +44,35 @@ int main(int argc, char** argv)
...
+            fprintf(stderr, "Unsupported protocol\n");
+            return 1;
+        }
+
+        if (url.host().is_empty()) {
+            fprintf(stderr, "Invalid URL\n");
+            return 1;
+        }
+
+        if (!url.port() || url.port() == 80)
```
awesomekling, line 71: We should probably stop having 80 as the default port in `URL` :)

PR #1931, opened 2020-04-23
LibJS: Work towards getting tests262 harness running

linusg:
Implemented some features that are used in the tests262 harness, specifically making these files parse & run correctly:

- `test262/harness/nans.js`
- `test262/harness/decimalToHexString.js`
- `test262/harness/doneprintHandle.js`

The major milestones to reach here seem to be regular expressions and template strings. One step at a time :)


Libraries/LibJS/Runtime/Value.cpp
```diff
@@ -262,6 +262,10 @@ Value left_shift(Interpreter&, Value lhs, Value rhs)
 
 Value right_shift(Interpreter&, Value lhs, Value rhs)
 {
+    if (lhs.to_number().is_nan() || lhs.to_number().is_infinity())
+        return Value(0);
+    if (rhs.to_number().is_nan() || rhs.to_number().is_infinity())
+        return lhs.to_number();
     return Value((i32)lhs.to_number().as_double() >> (i32)rhs.to_number().as_double());
```
awesomekling, line 269: Perhaps we should have something like `Value::is_finite_number()`?
Also, do we really want to do number coercion multiple times like this?

linusg: Yeah, it's a bit repetitive - will add a helper function :+1:

awesomekling:
:+1:

PR #1932, opened 2020-04-23
LibJS: Object expressions computed properties

linusg:
Had to change the datastructure for `ObjectExpression`s from `HashMap` to `Vector` as we need to execute all keys and values, even for duplicates.

(see updated tests for examples)

### DIFF
--- a/Libraries/LibJS/AST.h
+++ b/Libraries/LibJS/AST.h
@@ -670,9 +670,30 @@ private:
     NonnullRefPtrVector<VariableDeclarator> m_declarations;
 };
 
+class ObjectProperty final : public ASTNode {
+public:
+    ObjectProperty(NonnullRefPtr<Expression> key, NonnullRefPtr<Expression> value)
+        : m_key(move(key))
+        , m_value(move(value))
+    {
+    }
+
+    const Expression& key() const { return m_key; }
+    const Expression& value() const { return m_value; }
+
+    virtual void dump(int indent) const override;
+    virtual Value execute(Interpreter&) const override;
+
+private:
+    virtual const char* class_name() const override { return "ObjectProperty"; }
+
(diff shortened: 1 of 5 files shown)


linusg:
build failing because of https://github.com/SerenityOS/serenity/pull/1933

Libraries/LibJS/AST.h
```diff
@@ -670,9 +670,30 @@ class VariableDeclaration : public Declaration {
...
+class ObjectProperty final : public ASTNode {
+public:
+    ObjectProperty(NonnullRefPtr<Expression> key, NonnullRefPtr<Expression> value)
+        : m_key(key)
+        , m_value(value)
+    {
+    }
+
+    const NonnullRefPtr<Expression> key() const { return m_key; }
+    const NonnullRefPtr<Expression> value() const { return m_value; }
```
awesomekling, line 682: No need to return a `NonnullRefPtr` copy here as we are not giving up ownership of these objects.
```suggestion
    const Expression& key() const { return m_key; }
    const Expression& value() const { return m_value; }
```

Libraries/LibJS/AST.h
```diff
@@ -670,9 +670,30 @@ class VariableDeclaration : public Declaration {
     NonnullRefPtrVector<VariableDeclarator> m_declarations;
 };
 
+class ObjectProperty final : public ASTNode {
+public:
+    ObjectProperty(NonnullRefPtr<Expression> key, NonnullRefPtr<Expression> value)
+        : m_key(key)
+        , m_value(value)
```
awesomekling, line 677: We can `move()` here to avoid an unnecessary temporary copy of the pointers.
```suggestion
        : m_key(move(key))
        , m_value(move(value))
```

PR #1934, opened 2020-04-23
LibJS: Fix ObjectExpression::execute()

linusg:
Change from code review changed key() and value() getters - forgot to
update this...


awesomekling:
lül

PR #1936, opened 2020-04-23
Browser: Add Browser.ini configuration file

bcoles:
The Browser can now load a home page URL from the user's Browser.ini
configuration file rather than using a hard-coded URL.

Implementing this in `main.cpp` may or may not be a good idea, given that the browser is shifting towards a tabular UI model.

### DIFF
--- a/Applications/Browser/main.cpp
+++ b/Applications/Browser/main.cpp
@@ -27,6 +27,7 @@
 #include "InspectorWidget.h"
 #include "Tab.h"
 #include "WindowActions.h"
+#include <LibCore/ConfigFile.h>
 #include <LibCore/File.h>
 #include <LibGUI/AboutDialog.h>
 #include <LibGUI/Application.h>
@@ -37,8 +38,6 @@
 #include <LibWeb/ResourceLoader.h>
 #include <stdio.h>
 #include <stdlib.h>
-
-static const char* home_url = "file:///home/anon/www/welcome.html";
 
 int main(int argc, char** argv)
 {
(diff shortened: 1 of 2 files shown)


Applications/Browser/main.cpp
```diff
@@ -74,6 +73,9 @@ int main(int argc, char** argv)
 
     unveil(nullptr, nullptr);
 
+    auto m_config = Core::ConfigFile::get_for_app("Browser");
+    auto home_url = m_config->read_entry("Preferences", "Home", "file:///home/anon/www/welcome.html");
```
bugaevc, line 77: The default (as in, when no config file is found) URL should be http://serenityos.org, don't you think?

bcoles: Probably, in the long run, yes.

For now, `welcome.html` acts as a useful hub of test pages. There's also the issue of home page verse new tab home page. The `welcome.html` page is still hard-coded in `Tab.cpp` for new tabs.

@awesomekling also once stated something to the effect of "Serenity will never call back to sernityos.org". While I understand the sentiment of the statement with respect to privacy, perhaps it also applies here.

bugaevc: > For now, welcome.html acts as a useful hub of test pages.

Sure, that's why it's configured as anon's homepage. But it's one thing to pre-configure anon's file as anon's homepage in anon's browser config, and another thing to hardcode it as the default web page for everyone.

bcoles: Personally, I prefer a blank page as a home page and a blank page as the default for new tabs, rather than a hardcoded callback to ~msn.com~ serenityos.org over an unencrypted channel.

The `Tab` functionality also needs to be addressed, as it still uses the hardcoded `welcome.html` path.

Perhaps the best solution is to add a `blank.html` file in `/res/html/` (along with other static pages for the Browser, such as `error.html`), and have this as the default hardcoded home page and default page for new tabs; then ship a `Browser.ini` for `anon` with a home page of `welcome.html`. (Optionally, a default bookmark could also be added for `welcome.html` to `bookmarks.json` for `anon`.)

I've created a pull request #1957 for this.

PR #1939, opened 2020-04-24
Browser/LibWeb: Open links with target="_blank" in a new tab

linusg:
Hopefully I got this mostly right. I couldn't get `create_new_tab`'s `bool activate = true` default parameter working though as it now needs to be declared before the lambda is defined :confused:


awesomekling:
:+1:

PR #1940, opened 2020-04-24
AK: StringView uses WeakPtr instead of raw pointer to StringImpl

lnzero1dev:
This allows more safety in terms of accessing string view characters
when the memory already freed. When using a WeakPtr to the StringImpl,
we can check if the impl already has gone and assure freed memory is
not being accessed anymore.
New constructors for the StringView allow to hand over the impl, and
Strings substring_view method and StringViews own substring_view*
methods hand over the impl to the StringView constructors.
StringViews created from plain C character are not affected.

Needed some const_casts to allow the creation of the WeakPtr and
we don't wont to loose the constness of the String or FlyString
handed over in the constructor.

### DIFF
--- a/AK/LogStream.h
+++ b/AK/LogStream.h
@@ -32,7 +32,6 @@
 
 #if !defined(KERNEL) && !defined(BOOTSTRAPPER)
 #    include <AK/ScopedValueRollback.h>
-#    include <AK/StringView.h>
 #    include <errno.h>
 #    include <unistd.h>
 #endif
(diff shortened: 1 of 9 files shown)


lnzero1dev:
there is a bug :^) good we have unit tests

awesomekling:
This seems like a large increase in complexity and CPU+memory cost for a small protection against some (not all) `StringView` bugs.

It also hides bugs, since it does not assert or crash when attempting to access a no-longer-valid string. Instead situations may previously have crashed will now silently "succeed" instead, except appearing as if the string is empty.

lnzero1dev:
Ok, I checked the StringView, seems to make no difference with `const StringImpl*` or `WeakPtr<const StrimgImpl>`. But `StringImpl` goes from 24 Bytes to 40 Bytes when derive from `Weakable`.
What complexity do you mean? For the additional checks in `is_null()` or for the construction/destruction of the weakable?
I see, String shall be as lightweight as possible. I thought it would be good to have such prevention, but the argument that is doesn't fail is right. Do you have any advice how to make it better? =)

PR #1945, opened 2020-04-24
Meta: Time for a new screenshot! :^)

linusg:
... maybe? A lot happened since the last one, especially with the browser!

I tried to put some interesting stuff in there, let me know what you think. Too much?

_(it's not showing up in the readme preview as it references the main serenity repo where it doesn't exist yet.)_


linusg:
After today's browser changes this already feels outdated again :smile:

PR #1946, opened 2020-04-24
HackStudio: Add cpp Debugger

itamar8910:
hackstudio-debugger

The user can set breakpoints at various positions in the source code,
and then run the program in debug mode.

When the program is stopped, the current execution position is displayed,
and the user can insert/remove breakpoints, continue execution, or single step the program.


Libraries/LibDebug/DebugSession.cpp
```diff
@@ -133,40 +133,59 @@ bool DebugSession::insert_breakpoint(void* address)
...
     bp.state = BreakPointState::Enabled;
     m_breakpoints.set(bp.address, bp);
     return true;
 }
 
+bool DebugSession::remove_breakpoint(void* address)
+{
+    bool rc = disable_breakpoint(address);
+    if (!rc)
+        return false;
```
awesomekling, line 178: This would read better without the temporary:
```suggestion
    if (!disable_breakpoint(address))
        return false;
```

Libraries/LibDebug/DebugSession.cpp
```diff
@@ -133,40 +133,59 @@ bool DebugSession::insert_breakpoint(void* address)
...
+
+    m_breakpoints.remove(address);
+    return true;
+}
+
+bool DebugSession::breakpoint_exists(void* address)
```
awesomekling, line 184: This function should be `const` qualified.

DevTools/HackStudio/BreakpointCallback.h
```diff
@@ -0,0 +1,37 @@
...
+#include <AK/Function.h>
+#include <AK/String.h>
+#include <AK/Types.h>
+
+enum class BreakpointChange {
+    Added,
+    Removed,
+};
+
+typedef Function<void(const String&, size_t, BreakpointChange)> BreakpointChangeCallback; // file, line, BreakPointChange
```
awesomekling, line 37:
```suggestion
typedef Function<void(const String& file, size_t line, BreakpointChange)> BreakpointChangeCallback;
```

DevTools/HackStudio/Editor.cpp
```diff
@@ -79,12 +80,20 @@ void Editor::focusout_event(Core::Event& event)
     GUI::TextEditor::focusout_event(event);
 }
 
+Gfx::Rect Editor::breakpoint_icon_rect(size_t line_number)
```
awesomekling, line 83: This function should be `const` qualified.

DevTools/HackStudio/Editor.cpp
```diff
@@ -338,3 +379,31 @@ void Editor::navigate_to_include_if_available(String path)
...
+{
+    if (!m_execution_position.has_value()) {
+        return;
+    }
+    size_t previous_position = m_execution_position.value();
+    m_execution_position = {};
+    update(breakpoint_icon_rect(previous_position));
+}
+
+RefPtr<Gfx::Bitmap> Editor::breakpoint_icon_bitmap()
```
awesomekling, line 399: This function can just return `const Gfx::Bitmap&` since it's not passing ownership of the bitmap to the caller.
It also doesn't touch any members, so it could be `static`.

DevTools/HackStudio/Editor.cpp
```diff
@@ -338,3 +379,31 @@ void Editor::navigate_to_include_if_available(String path)
...
+{
+    static auto bitmap = Gfx::Bitmap::load_from_file("/res/icons/breakpoint.png");
+    return bitmap;
+}
+
+RefPtr<Gfx::Bitmap> Editor::current_position_icon_bitmap()
```
awesomekling, line 405: This function can just return `const Gfx::Bitmap&` since it's not passing ownership of the bitmap to the caller.
It also doesn't touch any members, so it could be `static`.

DevTools/HackStudio/Editor.cpp
```diff
@@ -338,3 +379,31 @@ void Editor::navigate_to_include_if_available(String path)
 
     on_open(it->value);
 }
+
+void Editor::set_execution_position(size_t line_number)
+{
+    m_execution_position = Optional<size_t>(line_number);
```
awesomekling, line 385:
```suggestion
    m_execution_position = line_number;
```

DevTools/HackStudio/Editor.h
```diff
@@ -42,6 +44,12 @@ class Editor final : public GUI::TextEditor {
     EditorWrapper& wrapper();
     const EditorWrapper& wrapper() const;
 
+    Vector<size_t> breakpoint_lines() const { return m_breakpoint_lines; }
```
awesomekling, line 47: This creates a copy of the vector on every call. That's probably not what we want. Instead return a `const Vector&`
```suggestion
    const Vector<size_t>& breakpoint_lines() const { return m_breakpoint_lines; }
```

awesomekling:
Ok! :D

PR #1947, opened 2020-04-24
LibJS: String.prototype.includes

kessejones:
Added includes in String.prototype.

### DIFF
--- a/Libraries/LibJS/Runtime/StringPrototype.cpp
+++ b/Libraries/LibJS/Runtime/StringPrototype.cpp
@@ -57,6 +57,7 @@ StringPrototype::StringPrototype()
     put_native_function("trimEnd", trim_end, 0);
     put_native_function("concat", concat, 1);
     put_native_function("substring", substring, 2);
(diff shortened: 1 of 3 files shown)


Libraries/LibJS/Runtime/StringPrototype.cpp
```diff
@@ -368,4 +369,35 @@ Value StringPrototype::substring(Interpreter& interpreter)
...
 }
 
+Value StringPrototype::includes(Interpreter& interpreter)
+{
+    auto* this_object = interpreter.this_value().to_object(interpreter.heap());
+    if (!this_object)
+        return {};
+
+    if (interpreter.argument_count() == 0)
+        return Value(false);
```
linusg, line 379: This is not correct.

`"string contains undefined".includes()` is supposed to return true.

> https://tc39.es/ecma262/#sec-string.prototype.includes
>
> 21.1.3.7 String.prototype.includes ( searchString [ , position ] )
> ...
> Let searchStr be ? ToString(searchString).

PR #1949, opened 2020-04-25
LibPthread: implicitly call pthread_exit on return from start routine.

DrewStratford:
Previously, when returning from a pthread's start_routine, we would
segfault. Now we instead implicitly call pthread_exit as specified in
the standard.

pthread_create now creates a thread running the new
pthread_create_helper, which properly manages the calling and exiting
of the start_routine supplied to pthread_create. To accomplish this,
the thread's stack initialization has been moved out of
sys$create_thread and into the userspace function create_thread.


awesomekling:
Hey that's quite nice. Thanks for improving this :)

PR #1950, opened 2020-04-25
LibGfx + LibWeb: Initial static GIF decoding and rendering

peterdn:
This PR:
- Implements GIF LZW decoding for the first frame of a GIF image
- Defines GIFImageDecoderPlugin class methods
- Adds a `sniff` method to ImageDecoderPlugin interface that is used by ImageDecoder to determine which specific plugin to use 
- Handles .gifs as images in LibWeb, allowing them to be rendered in Browser

Screenshot:
image

Note: I haven't added any test gifs to the PR, unsure whether you want large binaries in there and also whether image copyright etc might be a problem.


awesomekling:
Strong PR :D

PR #1953, opened 2020-04-25
LibGfx: Fix crash on decoding small gifs

peterdn:
The LZW decode step will now copy and pad LZW data out to 4 bytes if there are
less than 4 bytes remaining in the buffer. This means it will now also work when
the total size of the LZW image data is less than 4 bytes.


awesomekling:
Great! Thanks for fixing this :)

PR #1956, opened 2020-04-25
Gif image support for QuickShow and File Manager

asliturk:


Applications/QuickShow/QSWidget.cpp
```diff
@@ -84,7 +84,7 @@ void QSWidget::navigate(Directions direction)
         Core::DirIterator iterator(current_dir, Core::DirIterator::Flags::SkipDots);
         while (iterator.has_next()) {
             String file = iterator.next_full_path();
-            if (!file.ends_with(".png")) // TODO: Find a batter way to filter supported images.
+            if (!file.ends_with(".png") && !file.ends_with(".gif")) // TODO: Find a batter way to filter supported images.
```
awesomekling, line 87: This change doesn't seem to match the commit message?

asliturk: This is about navigate `gif` and `png` files in same directory.

awesomekling: Yes, but the commit message says "QuickShow: Use Icon class instead of Bitmap"

asliturk: uhhh :(

PR #1957, opened 2020-04-25
Browser: Set hardcoded default home page to file:///res/html/blank.html

bcoles:


awesomekling:
This is no longer needed as we now use `about:blank` as the default home page :)

PR #1958, opened 2020-04-25
Add the LaunchServer, enabling user-configurable file opening

niax:
Adds the LaunchServer which moves the `DesktopServices::open` into a Server. This makes it easier to handle the loading of config without all clients dealing with `unveil`ing appropriate paths.

In order to integrate this into other applications, `DesktopServices` had to be moved out of LibCore in order to link with LibIPC.


bugaevc:
It also needs functionality to spawn a particular app (by path). Then, we can drop `fork`/`proc/`exec` pledges from other apps that open files (think FileManager) and apps (think Taskbar).

niax:
Thanks for the feedback, apologies for my C++ n00bishness. Force pushing seems to have made the original comments go away, so answering some of them below:

> What happens if there's no extension? Does this panic? Return an empty string?

It results in thinking the extension is the complete path. This then goes through the one of the the default paths if you haven't specified the path in the config.

> It should also `unveil()` the paths it needs, if any.

I tried `unveil("/", "r")` but the Kernel still rejected the stat call which is used for identifying directories and executables. I think that it _should_ let me do this, but fixing `unveil` is out of the scope of this PR.

> It also needs functionality to spawn a particular app (by path). Then, we can drop `fork`/`proc`/`exec` pledges from other apps that open files (think FileManager) and apps (think Taskbar).

It already has that, and is being already used by FileManager both to open files and to start apps (from the Desktop it launches DisplayProperties).

bugaevc:
> It results in thinking the extension is the complete path. This then goes through the one of the the default paths if you haven't specified the path in the config.

Well, I'd prefer if it would consider the extension to be a null string (use `{}` to construct a null string).

> I tried `unveil("/", "r")` but the Kernel still rejected the stat call which is used for identifying directories and executables. I think that it _should_ let me do this, but fixing `unveil` is out of the scope of this PR.

Right, I haven't thought of that (perhaps place a comment about this where the `unveil()` call would be?). And `unveil("/", "r")` is a pretty sad thing to do, so I'd rather not do it at all.

> It already has that, and is being already used by FileManager both to open files and to start apps (from the Desktop it launches DisplayProperties).

Hmm, right, you can just pass the path to the executable and it will spawn that. Well, great, let's get rid of all the app fork/exec-s everywhere else then :slightly_smiling_face:

niax:
> Well, I'd prefer if it would consider the extension to be a null string (use {} to construct a null string).

Fair enough; we might want some different options later for handling extensionless files (like `Makefile`, but I'll settle for no extension and always using the default for now.

>> I tried unveil("/", "r") but the Kernel still rejected the stat call which is used for identifying directories and executables. I think that it should let me do this, but fixing unveil is out of the scope of this PR.
> 
> Right, I haven't thought of that (perhaps place a comment about this where the unveil() call would be?). And unveil("/", "r") is a pretty sad thing to do, so I'd rather not do it at all.

I don't disagree, but at very least `unveil("/", "r")` prevents a compromised LaunchServer from being able to write your files.

>> It already has that, and is being already used by FileManager both to open files and to start apps (from the Desktop it launches DisplayProperties).
>
> Hmm, right, you can just pass the path to the executable and it will spawn that. Well, great, let's get rid of all the app fork/exec-s everywhere else then 🙂

This is a future PR, along with moving the notification handling into LibDesktop :)

bugaevc:
> The question about what to unveil is important here. We don't want to create an overly permissive service that allows any anon-owned process with the unix promise to execute any arbitrary command.

Right. I think running any apps with any args should be fine, and running any binary without arguments should also be fine.

But this policy shouldn't be implemented via `unveil()`, `unveil()` is for protecting the system from LaunchServer should it get hax'd, not for limiting the API LaunchServer provides to its clients.

awesomekling:
Merged manually in 2708cc0f72c545b9c079b74c9d546ef55e6325ea and b7810a31c3b50240a083a3e59ad8a38d2d12dddb. Thanks!

PR #1959, opened 2020-04-25
LibJS: Implement Object.is()

linusg:
Basically === with two particularities: comparing NaN to itself is is considered equal and comparing +0 and -0 is not.

### DIFF
--- a/Libraries/LibJS/Runtime/ObjectConstructor.cpp
+++ b/Libraries/LibJS/Runtime/ObjectConstructor.cpp
@@ -41,6 +41,7 @@ ObjectConstructor::ObjectConstructor()
     put("prototype", interpreter().global_object().object_prototype());
 
     put_native_function("defineProperty", define_property, 3);
+    put_native_function("is", is, 2);
     put_native_function("getOwnPropertyDescriptor", get_own_property_descriptor, 2);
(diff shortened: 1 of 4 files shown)


Libraries/LibJS/Runtime/ObjectConstructor.cpp
```diff
@@ -146,4 +147,19 @@ Value ObjectConstructor::define_property(Interpreter& interpreter)
...
+Value ObjectConstructor::is(Interpreter& interpreter)
+{
+    auto value1 = interpreter.argument(0);
+    auto value2 = interpreter.argument(1);
+    if (value1.is_nan() && value2.is_nan())
+        return Value(true);
+    if (value1.is_number() && value1.as_double() == 0 && value2.is_number() && value2.as_double() == 0) {
+        // check if both are +0 or -0
+        auto value1_is_positive_zero = 1.0 / value1.as_double() == __builtin_huge_val();
+        auto value2_is_positive_zero = 1.0 / value2.as_double() == __builtin_huge_val();
```
awesomekling, line 159: Maybe these would make sense as helpers on `Value`? I imagine we'll want to check if a `Value` is `+0` or `-0` somewhere else at some point..

linusg: Definitely! Done :)

PR #1961, opened 2020-04-26
Kernel: Support signalling all processes with pid == -1

bgianfo:
This is a special case that was previously not implemented.
The idea is that you can dispatch a signal to all other processes
the calling process has access to.
    
There was some minor refactoring to make the self signal logic
into a function so it could easily be easily re-used from do_killall.

This PR also includes a change to Userland/kill.cpp to allow it
to actually parse negative pids, as that we previously broken.
There's no easy way to test this otherwise.

### DIFF
--- a/Kernel/Process.cpp
+++ b/Kernel/Process.cpp
@@ -2219,6 +2219,45 @@ KResult Process::do_killpg(pid_t pgrp, int signal)
     return error;
 }
 
+KResult Process::do_killall(int signal)
+{
+    InterruptDisabler disabler;
+
+    bool any_succeeded = false;
+    KResult error = KSuccess;
+
+    // Send the signal to all processes we have access to for.
+    for (auto& process : *g_processes) {
+        KResult res = KSuccess;
+        if (process.pid() == m_pid)
+            res = do_killself(signal);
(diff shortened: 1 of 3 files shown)


Kernel/Process.cpp
```diff
@@ -2219,6 +2219,50 @@ KResult Process::do_killpg(pid_t pgrp, int signal)
...
+
+    pid_t init_pid = 1;
+    bool any_succeeded = false;
+    KResult error = KSuccess;
+
+    // Send the signal to all processes we have access to for.
+    for (auto& process : *g_processes) {
+        // POSIX specifies we skip init.
+        if (process.pid() == init_pid)
+            continue;
```
awesomekling, line 2234: We don't have an `init`. The closest thing we have is `SystemServer` but that's not guaranteed to have a specific PID. Also it's fine to send signals to `SystemServer` if you wanted to.

This line from POSIX makes it sound like this is an optional behavior regardless, so I think we should just skip it:
>The unspecified processes to which a signal cannot be sent may include the scheduler or init.

bgianfo: done

PR #1963, opened 2020-04-26
Added ability to modify hostname

lukepayne:
Kernel: Added the ability to set the hostname via new syscall
Userland/hostname: Now takes parameter to set the hostname
LibC/unistd: Added sethostname function

### DIFF
--- a/Kernel/Process.cpp
+++ b/Kernel/Process.cpp
@@ -712,6 +712,18 @@ int Process::sys$gethostname(char* buffer, ssize_t size)
     return 0;
 }
 
+int Process::sys$sethostname(const char* hostname, ssize_t length)
+{
+    REQUIRE_PROMISE(stdio);
+    if (length < 0)
+        return -EINVAL;
+    LOCKER(*s_hostname_lock, Lock::Mode::Shared);
+    if (length > 64)
+        return -ENAMETOOLONG;
+    *s_hostname = validate_and_copy_string_from_user(hostname, length);
+    return 0;
+}
+
 pid_t Process::sys$fork(RegisterState& regs)
 {
     REQUIRE_PROMISE(proc);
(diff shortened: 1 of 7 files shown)


awesomekling:
Coolio :)

Kernel/Process.cpp
```diff
@@ -712,6 +712,18 @@ int Process::sys$gethostname(char* buffer, ssize_t size)
     return 0;
 }
 
+int Process::sys$sethostname(const char* hostname, ssize_t length)
+{
+    REQUIRE_PROMISE(stdio);
+    if (length < 0)
+        return -EINVAL;
```
bugaevc, line 719: So what, any random process on the system can just go ahead and change hostname?

Kernel/Process.cpp
```diff
@@ -712,6 +712,18 @@ int Process::sys$gethostname(char* buffer, ssize_t size)
...
+    LOCKER(*s_hostname_lock, Lock::Mode::Shared);
```
bugaevc, line 720: Clearly, this cannot take a shred lock.

Userland/hostname.cpp
```diff
@@ -36,14 +36,24 @@ int main(int argc, char** argv)
...
     }
 
-    (void)argc;
-    (void)argv;
-    char buffer[HOST_NAME_MAX];
-    int rc = gethostname(buffer, sizeof(buffer));
-    if (rc < 0) {
-        printf("gethostname() error: %s\n", strerror(errno));
+    if (argc == 1) {
+      char buffer[HOST_NAME_MAX];
```
bugaevc, line 40: Don't do that, use `Core::ArgsParser`

Kernel/Process.cpp
awesomekling: My bad.

Kernel/Process.cpp
awesomekling: My bad.

PR #1964, opened 2020-04-26
Kernel/LibPthread: Plumb futex wait timeout support throughout.

bgianfo:
This change brings basic support for timeouts to the futex system call when invoked with the FUTEX_WAIT option. It's currently implemented by taking advantage of the timer queue to wake
the thread on timeout.

With that implemented LibPthread needed to be fixed to actually pass the option through.
A test case was added to very the feature works.

### DIFF
--- a/Kernel/Process.cpp
+++ b/Kernel/Process.cpp
@@ -2282,6 +2282,19 @@ timeval kgettimeofday()
     return const_cast<const timeval&>(((KernelInfoPage*)s_info_page_address_for_kernel.as_ptr())->now);
 }
 
+void compute_relative_timeout_from_absolute(const timeval& absolute_time, timeval& relative_time)
+{
+    // Convert absolute time to relative time of day.
+    timeval_sub(absolute_time, kgettimeofday(), relative_time);
+}
+
+void compute_relative_timeout_from_absolute(const timespec& absolute_time, timeval& relative_time)
+{
+    timeval tv_absolute_time;
+    timespec_to_timeval(absolute_time, tv_absolute_time);
+    compute_relative_timeout_from_absolute(tv_absolute_time, relative_time);
+}
+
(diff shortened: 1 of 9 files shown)


Kernel/Thread.cpp
```diff
@@ -884,14 +885,21 @@ const LogStream& operator<<(const LogStream& stream, const Thread& value)
...
     if (lock)
         *lock = false;
     set_state(State::Queued);
     queue.enqueue(*current);
+
+    u64 timer_id = 0;
+    if (timeout)
+        timer_id = TimerQueue::the().add_timer(*timeout, [&]() {
+            wake_from_queue();
+        });
```
awesomekling, line 901: Please wrap multi-line consequent statements in `{ }` even if the compiler allows you not to.

Kernel/Process.cpp
```diff
@@ -4615,21 +4628,34 @@ int Process::sys$futex(const Syscall::SC_futex_params* user_params)
...
         copy_from_user(&user_value, userspace_address);
         if (user_value != value)
             return -EAGAIN;
+
+        timespec ts_abstimeout { 0, 0 };
+        if (user_timeout && !validate_read_and_copy_typed(&ts_abstimeout, user_timeout))
+            return -EFAULT;
+
+        WaitQueue& wait_queue = futex_queue(userspace_address);
+        timeval* optional_timeout = nullptr;;
```
awesomekling, line 4644: Double semicolon here :)

Kernel/Thread.cpp
bgianfo: Ironically, I thought I was matching your style here, fixed. :+1:

Kernel/Process.cpp
bgianfo: whoops, nice catch!

Issue #1365, opened 2020-03-06
FileManager: Disable actions/controls when have no permission to use them


xTibor:
Instead of throwing "Permission denied" popups after the user tried to use those features.

Affected actions/controls:
- New directory action
- Paste action
- Delete actions
- File name textbox in properties
- Permission checkboxes in properties

awesomekling:
Fixed in #1967

PR #1967, opened 2020-04-26
FileManager: Disable actions based on user permissions

zlotny:
I changed the PR to a draft so I can add multiple changes related to this.

- [x] Change write permission logic to use access() syscall.
- [x] Disable Paste action if user doesn't have write permission.
- [x] Disable delete if user doesn't have write permissions in directory.
- [x] Disable name text box on PropertiesDialog if the user can't change the name.
- [x] Disable permission checkboxes on PropertiesDialog if the user can't change them.

**Edit:**
I'll ignore the delete action for now because of a bit of UI misleading behaviour. Some actions act not on the selected item but on the current folder and that probably needs fixup. So writing logic to disable/enable some items that are contextually wrong doesn't make a lot of sense. Also removed the checkbox from the list item, as it's not going to be changed and shows 2 of 3 constantly on the PR list :)

**Edit 2:**
Now that #1980 is merged I converted this to a draft again to keep adding polish/features to the actions on FileManager

### DIFF
--- a/Applications/FileManager/main.cpp
+++ b/Applications/FileManager/main.cpp
@@ -460,10 +460,10 @@ int run_in_windowed_mode(RefPtr<Core::ConfigFile> config, String initial_locatio
             }
         },
         window);
-    paste_action->set_enabled(GUI::Clipboard::the().type() == "file-list");
 
     GUI::Clipboard::the().on_content_change = [&](const String& data_type) {
-        paste_action->set_enabled(data_type == "file-list");
+        auto current_location = directory_view.path();
+        paste_action->set_enabled(data_type == "file-list" && access(current_location.characters(), W_OK) == 0);
     };
 
     auto properties_action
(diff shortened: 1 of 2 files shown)


Applications/FileManager/main.cpp
```diff
@@ -649,8 +649,23 @@ int run_in_windowed_mode(RefPtr<Core::ConfigFile> config, String initial_locatio
...
+        }
+
+        bool can_write_in_path = false;
+        if (st.st_uid == getuid() && st.st_mode & S_IWUSR) {
+            can_write_in_path = true;
+        } else if (st.st_uid != getuid() && st.st_gid == getgid() && st.st_mode & S_IWGRP) {
+            can_write_in_path = true;
+        } else if (st.st_uid != getuid() && st.st_gid != getgid() && st.st_mode & S_IWOTH) {
+            can_write_in_path = true;
+        }
```
sunverwerth, line 665: Something like this would make a great addition to the standard library. It would be useful in the Terminal for suggesting executables.

zlotny: Yes, I thought about that. I was thinking if I could move this to some place on the standard library and use to also check if we can paste or delete and enable/disable the button accordingly...

I'll go to IRC and ask, I suppose is better.

awesomekling: The kernel already provides this functionality via the `access` syscall :)

sunverwerth: Learned something new today

zlotny:
By the way i just found out about #1365, which refers to this.

Also, #1980 should be reviewed/merged before this PR as it may change some things and break easy rebasing. Also we would like to revisit this enable/disable decisions once actions really do what they should do.

zlotny:
Now it's really ready for review.

This fixes #1365

PR #1968, opened 2020-04-26
js: Follow Serenity C++ naming conventions

linusg:
Am I understanding https://github.com/SerenityOS/serenity/blob/master/Documentation/CodingStyle.md#names correctly? :^)

### DIFF
--- a/Userland/js.cpp
+++ b/Userland/js.cpp
@@ -62,10 +62,10 @@ private:
     static JS::Value save_to_file(JS::Interpreter&);
 };
 
-bool dump_ast = false;
-bool print_last_result = false;
-static OwnPtr<Line::Editor> editor;
-static int repl_line_level = 0;
+static bool s_dump_ast = false;
+static bool s_print_last_result = false;
+static OwnPtr<Line::Editor> s_editor;
+static int s_repl_line_level = 0;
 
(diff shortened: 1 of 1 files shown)


Userland/js.cpp
```diff
@@ -62,10 +62,10 @@ class ReplObject : public JS::GlobalObject {
     static JS::Value save_to_file(JS::Interpreter&);
 };
 
-bool dump_ast = false;
-bool print_last_result = false;
-static OwnPtr<Line::Editor> editor;
-static int repl_line_level = 0;
+static bool s_dump_ast = false;
```
sunverwerth, line 65: I think `g_...` would make more sense here.

linusg: And then change it to non-`static`? A search for `static .* g_` has only 6 results across the codebase...

sunverwerth: The coding style says static members should be prefixed with s_ but these are global variables and not members. Maybe @awesomekling  should have a look at it.

awesomekling: Yeah the `s_` prefix convention is really only for static data members, but I see we have a lot of static locals prefixed with `s_` as well. Accidental habit I guess :) This convention makes sense to me, so let's roll with it.

PR #1974, opened 2020-04-26
LibGfx: Add support for grayscale PNGs

LepkoQQ:
I also added a modified local copy of a test page that loads all basic PNG formats (http://schaik.com/pngsuite/pngsuite_bas_png.html) since the serenity html parser doesn't parse the malformed table present on that page.


awesomekling:
Stellar!

PR #1975, opened 2020-04-26
Base+Browser: Add an icon for the serenity Browser and make it use it

zlotny:
Browser was using the filetype-html icon instead of a dedicated one, so
we now have the globe from that icon reimagined and in good Buggie
company :^)

These are the new icons (32x32 and 16x16 respectively. Zoomed in to see details)

image

image


awesomekling:
Love it love it love it! Let's call it `app-browser.png` so it matches all the existing app icons :)

zlotny:
You're right, changed!

PR #1982, opened 2020-04-27
Toolchain: Don't create repository for patches

devsh0:
The toolchain builds just fine without the git repository (tested on
windows and linux). We can skip setting up the repo and apply the
patches directly. This improves build time.

### DIFF
--- a/Toolchain/BuildIt.sh
+++ b/Toolchain/BuildIt.sh
@@ -32,6 +32,15 @@ elif [ "$(uname -s)" = "FreeBSD" ]; then
     NPROC="sysctl -n hw.ncpu"
 fi
 
+git_patch=
+while [ "$1" != "" ]; do
+    case $1 in
+        --dev )           git_patch=1
+                          ;;
+    esac
+    shift
+done
+
 echo PREFIX is "$PREFIX"
 echo SYSROOT is "$SYSROOT"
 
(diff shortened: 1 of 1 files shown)


Toolchain/BuildIt.sh
```diff
@@ -111,10 +111,7 @@ pushd "$DIR/Tarballs"
         tar -xzf ${BINUTILS_PKG}
 
         pushd ${BINUTILS_NAME}
-            git init >/dev/null
-            git add . >/dev/null
-            git commit -am "BASE" >/dev/null
-            git apply "$DIR"/Patches/binutils.patch >/dev/null
+	    patch -p1 < "$DIR"/Patches/binutils.patch > /dev/null
```
awesomekling, line 114: Please use spaces instead of tabs just like the surrounding code.

Toolchain/BuildIt.sh
```diff
@@ -125,10 +122,7 @@ pushd "$DIR/Tarballs"
         tar -xzf $GCC_PKG
 
         pushd $GCC_NAME
-            git init >/dev/null
-            git add . >/dev/null
-            git commit -am "BASE" >/dev/null
-            git apply "$DIR"/Patches/gcc.patch >/dev/null
+	    patch -p1 < "$DIR"/Patches/gcc.patch > /dev/null
```
awesomekling, line 125: Please use spaces instead of tabs just like the surrounding code.

NicolasVanBossuyt:
The git repo is useful when working on the toolchain maybe a flag to toggle between the git patch and regular patch can be added `./BuildIt.sh --dev`

awesomekling:
Sure, an optional flag could work too.

Toolchain/BuildIt.sh
devsh0: Done! PR virgin mistakes :)

awesomekling:
Let's do the flag that @NicolasVanBossuyt asked for as well :)

devsh0:
Sure!

PR #1983, opened 2020-04-27
Screensaver: Draw screen before first timer iteration and seed srand()

bcoles:
Screensaver randomization is now more random. Also, the screen is no longer black for the first iteration of `INTERVAL` seconds.

### DIFF
--- a/Demos/Screensaver/Screensaver.cpp
+++ b/Demos/Screensaver/Screensaver.cpp
@@ -26,13 +26,13 @@
 
 #include <LibCore/ElapsedTimer.h>
 #include <LibGfx/Bitmap.h>
-#include <LibGUI/Application.h>
 #include <LibGUI/Action.h>
+#include <LibGUI/Application.h>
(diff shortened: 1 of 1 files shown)


Demos/Screensaver/Screensaver.cpp
```diff
@@ -26,19 +26,35 @@
...
+void my_srand(unsigned s)
+{
+    seed = s - 1;
+}
+
+static int my_rand(void)
+{
+    seed = 6364136223846793005ULL * seed + 1;
+    return seed >> 33;
+}
```
awesomekling, line 56: Why not use `srand()` and `rand()` from LibC?

bcoles: > 
> 
> Why not use `srand()` and `rand()` from LibC?

It seemed like a [good idea](https://github.com/SerenityOS/serenity/blob/master/Demos/Fire/Fire.cpp#L73) at the time.

PR #1986, opened 2020-04-27
LibJS: Support spreading of strings and string objects

linusg:


mattco98:
I originally put the tests in a file called `array-spread.js` because I planned on also adding object spreading and placing the object tests in a different file. It might be preferable to keep `array` somewhere in that file name so we can easily separate the object spread tests.

awesomekling:
In general, I tend to think it's better to just add more tests instead of editing/altering old ones.
As long as a test is correct, updating it really only creates confusion for our future selves when we bisect and have tests doing different things at different points in the repo history.

linusg:
Updated!

PR #1989, opened 2020-04-27
Ports: Add neofetch

linusg:
image

Adding colour to Buggie was a pain in the a\*\*, but hopefully worth it :^)


awesomekling:
My goodness!

PR #1990, opened 2020-04-27
LibCore: Trim decompressed Gzip output to size

alimpfard:
Prior to this commit, we would (re-)allocate the output buffer aligned to 1024 bytes, but never trim it down to size, which caused Gzip::decompress to return uninitialised data.

smh, @awesomekling


awesomekling:
:+1:

PR #1991, opened 2020-04-27
LibJS: Array.prototype.some

kessejones:
Added some in Array.prototype. :)

### DIFF
--- a/Libraries/LibJS/Runtime/ArrayPrototype.cpp
+++ b/Libraries/LibJS/Runtime/ArrayPrototype.cpp
@@ -61,6 +61,7 @@ ArrayPrototype::ArrayPrototype()
     put_native_function("includes", includes, 1, attr);
     put_native_function("find", find, 1, attr);
     put_native_function("findIndex", find_index, 1, attr);
+    put_native_function("some", some, 1, attr);
     put("length", Value(0), Attribute::Configurable);
 }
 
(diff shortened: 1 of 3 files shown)


Libraries/LibJS/Runtime/ArrayPrototype.cpp
```diff
@@ -493,4 +494,38 @@ Value ArrayPrototype::find_index(Interpreter& interpreter)
...
+    if (!callback)
+        return {};
+
+    auto this_value = interpreter.argument(1);
+    auto array_size = array->elements().size();
+
+    for (size_t i = 0; i < array_size; ++i) {
+        auto value = array->elements().at(i);
+        if (value.is_empty())
+            continue;
```
linusg, line 513: This can break if the array is modified and ends up being smaller than before. I'll have to check but it likely affects some other array prototype functions as well if you've copied this from `findIndex` :wink:

Example to reproduce:

```
> a = [1, 2, 3];
[ 1, 2, 3 ]
> a.find(() => { a.pop() })
js: AK/Vector.h:276: T& AK::Vector<T, inline_capacity>::at(size_t) [with T = JS::Value; long unsigned int inline_capacity = 0; size_t = long unsigned int]: Assertion `i >= 0 && i < m_size' failed.
[1]    9565 abort (core dumped) serenity/Meta/Lagom/build/js
```

Look how I did it in `forEach` for example:

https://github.com/SerenityOS/serenity/blob/c4d05049c478980e3087c4043ec36ec315ee0f71/Libraries/LibJS/Runtime/ArrayPrototype.cpp#L123-L129

awesomekling:
Needs a conflict fix after the put attribute changes

kessejones:
Conflict fixed.

PR #1993, opened 2020-04-27
LibJS: Add bounds check to Array.prototype.{find,findIndex}

### LINKED ISSUE #1992: LibJS: Array.prototype.{find,findIndex} don't do a bounds check before element access
The number of iterations is limited to the initial array size, but we still need to check if the array did shrink since then before accessing each element.

Example to reproduce:

```
> a = [1, 2, 3];
[ 1, 2, 3 ]
> a.find(() => { a.pop() })
js: AK/Vector.h:276: T& AK::Vector<T, inline_capacity>::at(size_t) [with T = JS::Value; long unsigned int inline_capacity = 0; size_t = long unsigned int]: Assertion `i >= 0 && i < m_size' failed.
[1]    9565 abort (core dumped) serenity/Meta/Lagom/build/js
```

https://github.com/SerenityOS/serenity/pull/1991#discussion_r416209317

linusg:
The number of iterations is limited to the initial array size, but we still need to check if the array did shrink since then before accessing each element.

The other "loop & callback" functions already do this.

Fixes #1992.

### DIFF
--- a/Libraries/LibJS/Runtime/ArrayPrototype.cpp
+++ b/Libraries/LibJS/Runtime/ArrayPrototype.cpp
@@ -439,6 +439,9 @@ Value ArrayPrototype::find(Interpreter& interpreter)
     auto array_size = array->elements().size();
 
     for (size_t i = 0; i < array_size; ++i) {
+        if (i >= array->elements().size())
+            break;
+
         auto value = array->elements().at(i);
         if (value.is_empty())
             continue;
@@ -473,6 +476,9 @@ Value ArrayPrototype::find_index(Interpreter& interpreter)
     auto array_size = array->elements().size();
 
     for (size_t i = 0; i < array_size; ++i) {
+        if (i >= array->elements().size())
+            break;
+
         auto value = array->elements().at(i);
         if (value.is_empty())
             continue;
--- a/Libraries/LibJS/Tests/array-shrink-during-find-crash.js
+++ b/Libraries/LibJS/Tests/array-shrink-during-find-crash.js
@@ -1,0 +1,25 @@
+load("test-common.js");
+
+try {
+    var a, callbackCalled;
+
+    callbackCalled = 0;
+    a = [1, 2, 3, 4, 5];
+    a.find(() => {
+        callbackCalled++;
+        a.pop();
+    });
+    assert(callbackCalled === 3);
+
+    callbackCalled = 0;
+    a = [1, 2, 3, 4, 5];
+    a.findIndex(() => {
+        callbackCalled++;
+        a.pop();
+    });
+    assert(callbackCalled === 3);
+
+    console.log("PASS");
+} catch (e) {
+    console.log("FAIL: " + e);
+}


awesomekling:
Can we add a test? :)

linusg:
Always!

alimpfard:
@linusg let's pretend that we have `Array.prototype.splice`, would this behave correctly if you splice out the element right before the last one?

linusg:
> @linusg let's pretend that we have Array.prototype.splice, would this behave correctly if you splice out the element right before the last one?

I investigated a bit and looks like the changes in this PR are not fully correct - damn.

Apparently `forEach` stops looping as soon as we run out of elements at the current index, `map` as well but will treat missing elements as **empty values** and still append them to the resulting array, `find` and `findIndex` will continue looping with the missing values being **undefined**.

Thanks JavaScript...

So no, I don't think it'll behave according to the spec. Will fix in a separate PR.

**forEach**

```js
a = [1, 2, 3, 4, 5];
result = a.forEach((value, index, array) => {
    console.log(`a = ${a} value = ${value} index = ${index} array = ${array}`);
    a.splice(-2, 1);
    console.log(`a = ${a} value = ${value} index = ${index} array = ${array}`);
    console.log("------------------------------------------------------------")
});
console.log(result);
```

```
a = 1,2,3,4,5 value = 1 index = 0 array = 1,2,3,4,5
a = 1,2,3,5 value = 1 index = 0 array = 1,2,3,5
------------------------------------------------------------
a = 1,2,3,5 value = 2 index = 1 array = 1,2,3,5
a = 1,2,5 value = 2 index = 1 array = 1,2,5
------------------------------------------------------------
a = 1,2,5 value = 5 index = 2 array = 1,2,5
a = 1,5 value = 5 index = 2 array = 1,5
------------------------------------------------------------
undefined
```

**map**

```js
a = [1, 2, 3, 4, 5];
result = a.map((value, index, array) => {
    console.log(`a = ${a} value = ${value} index = ${index} array = ${array}`);
    a.splice(-2, 1);
    console.log(`a = ${a} value = ${value} index = ${index} array = ${array}`);
    console.log("------------------------------------------------------------")
});
console.log(result);
```

```
a = 1,2,3,4,5 value = 1 index = 0 array = 1,2,3,4,5
a = 1,2,3,5 value = 1 index = 0 array = 1,2,3,5
------------------------------------------------------------
a = 1,2,3,5 value = 2 index = 1 array = 1,2,3,5
a = 1,2,5 value = 2 index = 1 array = 1,2,5
------------------------------------------------------------
a = 1,2,5 value = 5 index = 2 array = 1,2,5
a = 1,5 value = 5 index = 2 array = 1,5
------------------------------------------------------------
Array(5) [ undefined, undefined, undefined, <2 empty slots> ]
```

**find / findIndex**

```js
a = [1, 2, 3, 4, 5];
result = a.find((value, index, array) => {
    console.log(`a = ${a} value = ${value} index = ${index} array = ${array}`);
    a.splice(-2, 1);
    console.log(`a = ${a} value = ${value} index = ${index} array = ${array}`);
    console.log("------------------------------------------------------------")
});
console.log(result);
```

```
a = 1,2,3,4,5 value = 1 index = 0 array = 1,2,3,4,5
a = 1,2,3,5 value = 1 index = 0 array = 1,2,3,5
------------------------------------------------------------
a = 1,2,3,5 value = 2 index = 1 array = 1,2,3,5
a = 1,2,5 value = 2 index = 1 array = 1,2,5
------------------------------------------------------------
a = 1,2,5 value = 5 index = 2 array = 1,2,5
a = 1,5 value = 5 index = 2 array = 1,5
------------------------------------------------------------
a = 1,5 value = undefined index = 3 array = 1,5
a = 5 value = undefined index = 3 array = 5
------------------------------------------------------------
a = 5 value = undefined index = 4 array = 5
a =  value = undefined index = 4 array =
------------------------------------------------------------
undefined
```

PR #1995, opened 2020-04-28
LibJS: Implement correct attributes for (almost) all properties

mattco98:
Added the ability to include a `u8` `attributes` parameter with all of the various put methods in the Object class. They can be omitted, in which case it defaults to `Writable | Enumerable | Configurable`, just like before this commit.

All of the attribute values for each property were gathered from SpiderMonkey in the Firefox console. Some properties (e.g. all of the canvas element properties) have undefined property descriptors... not quite sure what that means. Those were left as the default specified above.

I considered the idea of having to always specify the value of `attributes` by remove the default value, as it may provide more clarity (it isn't obvious that omitting it results in the above value), but I wasn't sure. I'd love some opinions on that.


awesomekling:
Great idea :)

PR #1998, opened 2020-04-28
WindowManager+DisplayProperties: Allow us to roll back to a colour background from a bitmap one

zlotny:
This PR improves a bit how `DisplayProperties` work (and removes the empty item on the `ComboBox` for something more insightful) and allows us to go from a set bitmap to a background colour again.

### DIFF
--- a/Applications/DisplayProperties/DisplayProperties.cpp
+++ b/Applications/DisplayProperties/DisplayProperties.cpp
@@ -72,6 +72,8 @@ void DisplayPropertiesWidget::create_wallpaper_list()
 {
     Core::DirIterator iterator("/res/wallpapers/", Core::DirIterator::Flags::SkipDots);
 
+    m_wallpapers.append("Use background color");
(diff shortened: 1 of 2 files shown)


Applications/DisplayProperties/DisplayProperties.cpp
```diff
@@ -281,6 +287,8 @@ void DisplayPropertiesWidget::load_current_settings()
             m_wallpaper_combo->set_text(selected_wallpaper);
             m_wallpaper_combo->set_only_allow_values_from_model(true);
         }
+    }else{
```
awesomekling, line 290: pls `clang-format`

zlotny: Done! It slipped in the commit, sorry about that.

awesomekling: Np!

PR #2000, opened 2020-04-28
[joke] LibC: Reduce time_t to 32 bits so we can crash properly

alimpfard:
In memory of win2k :^)


alimpfard:
I stole your #2000 with a silly PR

alimpfard:
aww it doesn't build

silly printf size checks

awesomekling:
For maximum Y2K horror, we'll need to go down all the way to 8-bit BCD!

alimpfard:
That's more suited to an April 1st joke.

...I think this PR made enough people laugh, I'll close it now~

PR #2003, opened 2020-04-28
LibJS: Correctness fixes for Array.prototype.{map,find,findIndex}()

linusg:
I'll just leave this warning from MDN here.

image

This is peak JavaScript inconsistency. Don't make any assumptions, ever!


alimpfard:
These should've been what __wat__ covered...

I wonder why this sort of behaviour was even needed?

awesomekling:
Nice job tracking these down :)

Issue #1910, opened 2020-04-21
js: fuzzing results (that hit assertions)


Mindavi:
I thought I'd share some of the results I got using fuzzing.

# Layout

First (1) the offending code is shown, then (2) the output of `js -A <file>` and lastly (3) the output of spidermonkey is shown.

Only examples for unresolved cases are shown, if someone resolves one of these, it would be nice if a test case was added that tests the case :smile:.

# Fixed cases

- [x] Return for (2 examples above)
- [x] Addition and `++` operator (PR #1911)
- [x] Whole lot of `+` signs
- [x] ToString `hasOwnProperty` (PR #2004)
- [x] For loop with undefined variable (PR #2006)
- [x] FlyString value empty (PR #2012)
- [x] to_number assertion (PR #2011)

## Returning the `for` keyword

1

```js
return for;
```

2

```
Syntax Error: Unexpected token For. Expected Semicolon (line: 1, column: 9)
Syntax Error: Unexpected token Semicolon. Expected ParenOpen (line: 1, column: 12)
js: /home/rick/hobby/serenity/Libraries/LibJS/Parser.cpp:935: NonnullRefPtr<JS::ForStatement> JS::Parser::parse_for_statement(): Assertion `false' failed.
Aborted (core dumped)
```

3

```
crashing-inputs/for:1:0 SyntaxError: return not in function:
crashing-inputs/for:1:0 return for;
crashing-inputs/for:1:0 ^
```

## Returning the for keyword v2

1

```js
function test() {
  return for;
}
```

2

```
Syntax Error: Unexpected token For. Expected Semicolon (line: 2, column: 10)
Syntax Error: Unexpected token Semicolon. Expected ParenOpen (line: 2, column: 13)
js: /home/rick/hobby/serenity/Libraries/LibJS/Parser.cpp:1002: NonnullRefPtr<JS::ForStatement> JS::Parser::parse_for_statement(): Assertion `false' failed.
Aborted (core dumped)
```

3

```
for2:2:9 SyntaxError: expected expression, got keyword 'for':
for2:2:9   return for;
for2:2:9 .........^
```

linusg:
> I'm not sure if I should report the assertions that are hit, because they probably mean something is still TBD. However, I thought I'd share some of the results I got using fuzzing.

I still think reporting these is good, makes it easy to pick up something specific and fix it in the parser/AST :)

> They should all gracefully fail, I think.

[...]

PR #2011, opened 2020-04-28
LibJS: LibJS: Skip undefined and null in join_array_with_separator() / Don't handle arrays separately in Value::to_number()

linusg:
Small fix for `Array.prototype.join()`/`Array.prototype.toString()`: https://www.ecma-international.org/ecma-262/5.1/#sec-15.4.4.5

Subsequently we can remove our custom array code path in `Value::to_number()` and just use `to_primitive` as the spec dictates it.

See the commit descriptions for details.

This fixes a crash when doing `+[,]` reported by @Mindavi in https://github.com/SerenityOS/serenity/issues/1910 - keep 'em coming! :smile:


awesomekling:
Really nice!

PR #2013, opened 2020-04-28
LibJS: String.prototype.slice

kessejones:
Added slice in String.prototype. :)


Libraries/LibJS/Runtime/StringPrototype.cpp
```diff
@@ -399,4 +400,45 @@ Value StringPrototype::includes(Interpreter& interpreter)
     return Value(substring_search.contains(search_string));
 }
 
+Value StringPrototype::slice(Interpreter& interpreter)
+{
+    auto* this_object = interpreter.this_value().to_object(interpreter.heap());
+    if (!this_object)
+        return {};
+
+    auto& string = this_object->to_string().as_string()->string();
```
awesomekling, line 409: `to_string()` may call a `toString()` method on the object if present, and is not guaranteed to return a string value.
Therefore it's not safe to call `as_string()` on it without checking `is_string()` first.

PR #2014, opened 2020-04-29
LibJS: Object.{keys,values,entries} & correct object property ordering

mattco98:
Adds `Object.{keys,values,entries}()` which can be applied to all types except null and undefined.

This commit also introduces a way to get an object's own properties in the correct order. In this context, the order is the standard JS object property ordering: first, all array index-like properties (number keys), followed by all string properties. Each of the previous two groups is ordered by time of insertion, with the earliest inserted item coming first.

Also fixed the attributes of `Array.prototype.length` in order to make one of the tests work.

The implementation of `property_table_ordered` is not optimal, as it creates a new `Vector` every call. I was thinking about just constructing a vector alongside `m_property_table` and updating that whenever `m_property_table` is updated.

### DIFF
--- a/Libraries/LibJS/AST.cpp
+++ b/Libraries/LibJS/AST.cpp
@@ -1046,9 +1046,9 @@ Value ObjectExpression::execute(Interpreter& interpreter) const
             } else if (key_result.is_object()) {
                 auto& obj_to_spread = key_result.as_object();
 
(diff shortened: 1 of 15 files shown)


mattco98:
I'm gonna close this PR actually since it's kinda gross, I don't like how it's two slightly-but-not-really related features crammed into one commit. I'll reopen new separate PRs tonight or tomorrow :)

PR #2016, opened 2020-04-29
LibJS: Implement correct object property ordering

mattco98:
This commit introduces a way to get an object's own properties in the correct order. The "correct order" for JS object properties is first all array-like index properties (numeric keys) sorted by insertion order, followed by all string properties sorted by insertion order.

Objects also now print correctly in the repl! Before this commit:

```
courage ~/js-tests $ js
> ({ foo: 1, bar: 2, baz: 3 })
{ bar: 2, foo: 1, baz: 3 }
```

After:

```
courage ~/js-tests $ js
> ({ foo: 1, bar: 2, baz: 3 })
{ foo: 1, bar: 2, baz: 3 }
```

The implementation of `property_table_ordered` is not optimal though, as it creates a new Vector every call. I was thinking about just constructing a vector alongside `m_property_table` and updating that whenever `m_property_table` is updated.

### DIFF
--- a/Libraries/LibJS/Runtime/Object.cpp
+++ b/Libraries/LibJS/Runtime/Object.cpp
@@ -34,6 +34,7 @@
 #include <LibJS/Runtime/NativeProperty.h>
 #include <LibJS/Runtime/Object.h>
 #include <LibJS/Runtime/Shape.h>
+#include <LibJS/Runtime/StringObject.h>
 #include <LibJS/Runtime/Value.h>
 
 namespace JS {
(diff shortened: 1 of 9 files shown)


Libraries/LibJS/Runtime/Object.cpp
```diff
@@ -107,6 +108,69 @@ Value Object::get_own_property(const Object& this_object, const FlyString& prope
     return value_here;
 }
 
+Value Object::get_own_properties(const Object& this_object, GetOwnPropertiesMode kind, bool enumerable_only) const
```
awesomekling, line 111: Wait who calls this? I don't see a caller

mattco98: Oops, looks like that sneaked in from #2014. I decided to close that and open two separate PRs, and this code belongs in the other one

awesomekling:
Aha ok! :)

PR #2019, opened 2020-04-29
Implement a stream json parser

alimpfard:
This PR also patches ProfileViewer to use StreamJson{Parser,Builder} to parse the perfcores.

Not sure if that helps anything, I couldn't get `profile` to generate anything more than 4000 or so samples.


alimpfard:
will close until we end up needing it

PR #2024, opened 2020-04-29
Markdown images & don't assert on broken link markup

bugaevc:
Please check that the new assertion in `StringView` doesn't regress performance on all those benchmarks (JS, ProfileVIewer...). If it does, the first commit can be dropped.

We can now parse those images in ReadMe.md, but not display them because they're HTTPS. Sad face.

### DIFF
--- a/AK/StringView.h
+++ b/AK/StringView.h
@@ -38,14 +38,14 @@ class StringView {
 public:
     using ConstIterator = const char*;
 
-    StringView() { }
-    StringView(const char* characters, size_t length)
+    [[gnu::always_inline]] inline StringView() { }
+    [[gnu::always_inline]] inline StringView(const char* characters, size_t length)
         : m_characters(characters)
         , m_length(length)
     {
         ASSERT(!Checked<uintptr_t>::addition_would_overflow((uintptr_t)characters, length));
     }
-    StringView(const unsigned char* characters, size_t length)
+    [[gnu::always_inline]] inline StringView(const unsigned char* characters, size_t length)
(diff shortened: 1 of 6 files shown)


AK/StringView.h
```diff
@@ -36,16 +37,18 @@ class StringView {
 public:
     using ConstIterator = const char*;
 
-    StringView() {}
+    StringView() { }
     StringView(const char* characters, size_t length)
         : m_characters(characters)
         , m_length(length)
     {
+        ASSERT(!__builtin_add_overflow_p((uintptr_t)characters, length, (uintptr_t)0));
```
awesomekling, line 45: Can we add a helper in `Checked.h` for this?
Something like `bool Checked<T>::addition_would_overflow(U, V)`

bugaevc: Sure. But I've found that if I write it the the way `Checked::multiplication_would_overflow()` is currently written, GCC is unable to optimize out the intermediary variable (so it stores the flag in a register and then branches on that register, rather than branching on the flag directly). Would you be fine with using `__builtin_..._overflow_p` intrinsics in `Checked::..._would_overflow()` (instead of defining them in terms of other `Checked` operations)?

awesomekling: Of course :)

bugaevc:
OK, here's v2 with more assertion / overflow / string tweaks.

bugaevc:
Aaarrgh, ShitHub has messed up the order of commits as always.

awesomekling:
Very tasteful!

PR #2030, opened 2020-04-30
Shell: Provide the correct invariant length to the line editor

alimpfard:
LibLine: where every fix brings about another bug.


alimpfard:
Ideally we should not crash on invalid user-supplied values; yet I see no better outcome
would we rather do wonky stuff, or trip an assert?

awesomekling:
Trip an assert :)

alimpfard:
Then let's just change `Editor::get_line` to say `ASSERT_NOT_REACHED()`
No code is best code :^)

Issue #2001, opened 2020-04-28
PaintBrush: create from clipboard


engelju:
A huge use case of mine in MS Paint was to create a file from clipboard in paint, then edit it. I would use this mostly to annotate screenshots.

It would be nice for this to exist in Serenity.

linusg:
This was implemented a while ago, relevant commits: 6aa7f59608df5d29bb9f57ac221af62b97ae8785, f405cd3830fea7fa12e87aa3fecf3b7866ba016b

PR #2032, opened 2020-04-30
[Multiple] Add the posibility to copy/paste bitmaps to/from the Clipboard

zlotny:
This also adds the appropriate actions to PaintBrush and fixes the actions on TerminalWidget + TextEditor so they check that whatever they're trying to paste is indeed text.

It is a pretty barebones implementation but just enough to get started.

Kind of matches #2001 btw. Maybe once this is in the tree we could create a menu item that said "New from clipboard" and just create a new bitmap and paste.

EDIT: Title s/contents/bitmaps/

### DIFF
--- a/Applications/PaintBrush/main.cpp
+++ b/Applications/PaintBrush/main.cpp
@@ -31,6 +31,7 @@
 #include <LibGUI/Action.h>
 #include <LibGUI/Application.h>
 #include <LibGUI/BoxLayout.h>
+#include <LibGUI/Clipboard.h>
 #include <LibGUI/FilePicker.h>
 #include <LibGUI/Menu.h>
 #include <LibGUI/MenuBar.h>
@@ -96,7 +97,26 @@ int main(int argc, char** argv)
         return;
     }));
 
-    menubar->add_menu("Edit");
+    auto& edit_menu = menubar->add_menu("Edit");
+
+    auto paste_action = GUI::CommonActions::make_paste_action(
+        [&](const GUI::Action&) {
+            auto old_bitmap = RefPtr<Gfx::Bitmap>(paintable_widget.bitmap());
+            auto clipboard_bitmap = GUI::Clipboard::the().get_bitmap();
+            if (!clipboard_bitmap)
+                return;
+            paintable_widget.set_bitmap(*clipboard_bitmap->cloned());
+        },
+        window);
+
+    auto copy_action = GUI::CommonActions::make_copy_action(
+        [&](const GUI::Action&) {
+            GUI::Clipboard::the().set_data(paintable_widget.bitmap());
+        },
+        window);
+
+    edit_menu.add_action(copy_action);
+    edit_menu.add_action(paste_action);
 
     auto& help_menu = menubar->add_menu("Help");
     help_menu.add_action(GUI::Action::create("About", [&](auto&) {
--- a/Libraries/LibGUI/Clipboard.cpp
+++ b/Libraries/LibGUI/Clipboard.cpp
@@ -28,6 +28,8 @@
 #include <AK/SharedBuffer.h>
 #include <LibGUI/Clipboard.h>
 #include <LibGUI/WindowServerConnection.h>
+#include <LibGfx/Bitmap.h>
+#include <LibGfx/Color.h>
 
 namespace GUI {
 
@@ -62,6 +64,26 @@ Clipboard::DataAndType Clipboard::data_and_type() const
     return { data, type };
 }
 
+RefPtr<Gfx::Bitmap> Clipboard::get_bitmap()
+{
+    auto response = WindowServerConnection::the().send_sync<Messages::WindowServer::GetClipboardBitmap>();
+
+    if (response->shbuf_id() < 0)
+        return {};
+    auto shared_buffer = SharedBuffer::create_from_shbuf_id(response->shbuf_id());
+
+    if (!shared_buffer) {
+        dbgprintf("GUI::Clipboard::data() failed to attach to the shared buffer\n");
+        return {};
+    }
+    if (response->content_size() > shared_buffer->size()) {
+        dbgprintf("GUI::Clipboard::data() clipping contents size is greater than shared buffer size\n");
+        return {};
+    }
+
+    return Gfx::Bitmap::create_with_shared_buffer(response->bitmap_format(), *shared_buffer, response->bitmap_size());
+}
+
 void Clipboard::set_data(const StringView& data, const String& type)
 {
     auto shared_buffer = SharedBuffer::create_with_size(data.length() + 1);
@@ -79,10 +101,17 @@ void Clipboard::set_data(const StringView& data, const String& type)
     WindowServerConnection::the().send_sync<Messages::WindowServer::SetClipboardContents>(shared_buffer->shbuf_id(), data.length(), type);
 }
 
+void Clipboard::set_data(const Gfx::Bitmap& bitmap)
+{
+    auto bitmap_with_shared_buffer = bitmap.to_bitmap_backed_by_shared_buffer();
+    auto shared_buffer = bitmap_with_shared_buffer->shared_buffer();
+    shared_buffer->share_with(WindowServerConnection::the().server_pid());
+    WindowServerConnection::the().send_sync<Messages::WindowServer::SetClipboardBitmap>(shared_buffer->shbuf_id(), shared_buffer->size(), bitmap.format(), bitmap.size());
+}
+
 void Clipboard::did_receive_clipboard_contents_changed(Badge<WindowServerConnection>, const String& data_type)
 {
     if (on_content_change)
         on_content_change(data_type);
 }
-
 }
--- a/Libraries/LibGUI/Clipboard.h
+++ b/Libraries/LibGUI/Clipboard.h
@@ -27,8 +27,10 @@
 #pragma once
 
 #include <AK/Function.h>
+#include <AK/SharedBuffer.h>
 #include <AK/String.h>
 #include <LibGUI/Forward.h>
+#include <LibGfx/Bitmap.h>
 
 namespace GUI {
 
@@ -39,13 +41,14 @@ public:
     String data() const { return data_and_type().data; }
     String type() const { return data_and_type().type; }
     void set_data(const StringView&, const String& data_type = "text");
-
+    void set_data(const Gfx::Bitmap&);
     struct DataAndType {
         String data;
         String type;
     };
 
     DataAndType data_and_type() const;
+    RefPtr<Gfx::Bitmap> get_bitmap();
 
     void did_receive_clipboard_contents_changed(Badge<WindowServerConnection>, const String& data_type);
 
--- a/Libraries/LibGUI/TextEditor.cpp
+++ b/Libraries/LibGUI/TextEditor.cpp
@@ -1198,11 +1198,13 @@ void TextEditor::paste()
 {
     if (is_readonly())
         return;
-    auto paste_text = Clipboard::the().data();
-    printf("Paste: \"%s\"\n", paste_text.characters());
+    auto [data, type] = Clipboard::the().data_and_type();
+    if (type != "text")
+        return;
+    printf("Paste: \"%s\"\n", data.characters());
 
     TemporaryChange change(m_automatic_indentation_enabled, false);
-    insert_at_cursor_or_replace_selection(paste_text);
+    insert_at_cursor_or_replace_selection(data);
 }
 
 void TextEditor::enter_event(Core::Event&)
--- a/Libraries/LibGfx/Bitmap.cpp
+++ b/Libraries/LibGfx/Bitmap.cpp
@@ -29,8 +29,8 @@
 #include <AK/SharedBuffer.h>
 #include <AK/String.h>
 #include <LibGfx/Bitmap.h>
-#include <LibGfx/PNGLoader.h>
 #include <LibGfx/GIFLoader.h>
+#include <LibGfx/PNGLoader.h>
 #include <LibGfx/ShareableBitmap.h>
 #include <errno.h>
 #include <fcntl.h>
@@ -61,6 +61,16 @@ RefPtr<Bitmap> Bitmap::create_purgeable(BitmapFormat format, const Size& size)
     return adopt(*new Bitmap(format, size, Purgeable::Yes));
 }
 
+RefPtr<Bitmap> Bitmap::cloned()
+{
+    auto cloned_bitmap = Bitmap::create(format(), size());
+    if (!cloned_bitmap) {
+        return nullptr;
+    }
+    memcpy(cloned_bitmap->scanline(0), scanline(0), size_in_bytes());
+    return cloned_bitmap;
+}
+
 Bitmap::Bitmap(BitmapFormat format, const Size& size, Purgeable purgeable)
     : m_size(size)
     , m_pitch(round_up_to_power_of_two(size.width() * sizeof(RGBA32), 16))
@@ -86,9 +96,9 @@ RefPtr<Bitmap> Bitmap::create_wrapper(BitmapFormat format, const Size& size, siz
 
 RefPtr<Bitmap> Bitmap::load_from_file(const StringView& path)
 {
-    if(path.ends_with(".png"))
+    if (path.ends_with(".png"))
         return load_png(path);
-    if(path.ends_with(".gif"))
+    if (path.ends_with(".gif"))
         return load_gif(path);
 
     return nullptr;
--- a/Libraries/LibGfx/Bitmap.h
+++ b/Libraries/LibGfx/Bitmap.h
@@ -59,6 +59,7 @@ public:
     RefPtr<Gfx::Bitmap> flipped(Gfx::Orientation) const;
     RefPtr<Bitmap> to_bitmap_backed_by_shared_buffer() const;
 
+    RefPtr<Bitmap> cloned();
     ShareableBitmap to_shareable_bitmap(pid_t peer_pid = -1) const;
 
     ~Bitmap();
--- a/Libraries/LibVT/TerminalWidget.cpp
+++ b/Libraries/LibVT/TerminalWidget.cpp
@@ -568,10 +568,12 @@ void TerminalWidget::paste()
 {
     if (m_ptm_fd == -1)
         return;
-    auto text = GUI::Clipboard::the().data();
-    if (text.is_empty())
+    auto [data, type] = GUI::Clipboard::the().data_and_type();
+    if (type != "text")
         return;
-    int nwritten = write(m_ptm_fd, text.characters(), text.length());
+    if (data.is_empty())
+        return;
+    int nwritten = write(m_ptm_fd, data.characters(), data.length());
     if (nwritten < 0) {
         perror("write");
         ASSERT_NOT_REACHED();
--- a/Servers/WindowServer/Clipboard.h
+++ b/Servers/WindowServer/Clipboard.h
@@ -29,6 +29,7 @@
 #include <AK/Function.h>
 #include <AK/SharedBuffer.h>
 #include <AK/String.h>
+#include <LibGfx/Bitmap.h>
 
 namespace WindowServer {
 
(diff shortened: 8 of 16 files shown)


Libraries/LibGUI/TextEditor.cpp
```diff
@@ -1198,11 +1198,13 @@ void TextEditor::paste()
 {
     if (is_readonly())
         return;
-    auto paste_text = Clipboard::the().data();
-    printf("Paste: \"%s\"\n", paste_text.characters());
+    auto data_and_type = Clipboard::the().data_and_type();
+    if (data_and_type.type != "text")
+        return;
+    printf("Paste: \"%s\"\n", data_and_type.data.characters());
```
linusg, line 1204: Might be nice to use a [structured binding declaration](https://en.cppreference.com/w/cpp/language/structured_binding) here:

```cpp
auto [data, type] = clipboard.data_and_type();
if (type != "text")
    return;
printf("Paste: \"%s\"\n", data.characters());
...
```

Libraries/LibVT/TerminalWidget.cpp
```diff
@@ -568,10 +568,12 @@ void TerminalWidget::paste()
 {
     if (m_ptm_fd == -1)
         return;
-    auto text = GUI::Clipboard::the().data();
-    if (text.is_empty())
+    auto data_and_type = GUI::Clipboard::the().data_and_type();
+    if (data_and_type.type != "text")
```
linusg, line 572: Same here

Libraries/LibGUI/TextEditor.cpp
linusg: > feels like tuple unpacking in python

yes, I thought the same!

Libraries/LibGUI/Clipboard.cpp
```diff
@@ -62,6 +64,26 @@ Clipboard::DataAndType Clipboard::data_and_type() const
...
+
+    if (!shared_buffer) {
+        dbgprintf("GUI::Clipboard::data() failed to attach to the shared buffer\n");
+        return {};
+    }
+    if (response->content_size() > shared_buffer->size()) {
+        dbgprintf("GUI::Clipboard::data() clipping contents size is greater than shared buffer size\n");
+        return {};
+    }
+    shared_buffer->share_with(caller_pid);
```
bugaevc, line 82: No no no, if anyone, it's the WindowServer who should be calling `share_with`. And you're not going to be able to even `create_from_shbuf_id()` if the owner hasn't shared it with you.

bugaevc: > And you're not going to be able to even create_from_shbuf_id() if the owner hasn't shared it with you.

If that *does* work, it's a kernel bug that needs to be fixed.

Libraries/LibGUI/TextEditor.cpp
```diff
@@ -1198,11 +1198,13 @@ void TextEditor::paste()
 {
     if (is_readonly())
         return;
-    auto paste_text = Clipboard::the().data();
-    printf("Paste: \"%s\"\n", paste_text.characters());
+    auto [data, type] = Clipboard::the().data_and_type();
+    if (type != "text")
+        return;
+    printf("Paste: \"%s\"\n", data.characters());
```
bugaevc, line 1204:
```suggestion
    dbg() << "Paste: _" << data << "_";
```

Libraries/LibGfx/Bitmap.cpp
```diff
@@ -61,6 +61,13 @@ RefPtr<Bitmap> Bitmap::create_purgeable(BitmapFormat format, const Size& size)
     return adopt(*new Bitmap(format, size, Purgeable::Yes));
 }
 
+RefPtr<Bitmap> Bitmap::cloned_from(RefPtr<Bitmap> bitmap)
+{
+    auto cloned_bitmap = Bitmap::create(bitmap->format(), bitmap->size());
+    memcpy(cloned_bitmap->scanline(0), bitmap->scanline(0), bitmap->size_in_bytes());
+    return cloned_bitmap;
```
bugaevc, line 68: I wonder if there's a way to use virtual memory magic to avoid the actual copy.

Libraries/LibVT/TerminalWidget.cpp
```diff
@@ -568,10 +568,12 @@ void TerminalWidget::paste()
...
     if (m_ptm_fd == -1)
         return;
-    auto text = GUI::Clipboard::the().data();
-    if (text.is_empty())
+    auto [data, type] = GUI::Clipboard::the().data_and_type();
+    if (type != "text")
         return;
-    int nwritten = write(m_ptm_fd, text.characters(), text.length());
+    if (data.is_empty())
+        return;
```
bugaevc, line 575: What's wrong with taking the generic code path?

Libraries/LibGUI/Clipboard.cpp
zlotny: I'll have a look at this when I have time. I don't know why that's bad ; I assumed I had to share the buffer with PaintBrush before PB could do anything with it. Maybe I can pass the data without doing that. I can't check it right now. Thank you for the info :)

Libraries/LibGUI/TextEditor.cpp
zlotny: That line was already there wrote by @awesomekling so I didn't change it to don't over-do things for this PR. Shall I commit that change too? It seemed weird when I saw it, too.

Libraries/LibGfx/Bitmap.cpp
zlotny: This was the suggestion that @awesomekling gave to me over IRC as he was teaching me some things on how these things work. Maybe there's other ways of having that memory writable. Indeed it would be more efficient

Libraries/LibGUI/Clipboard.cpp
bugaevc: Somebody who already has access to the buffer has to share it with PB for PB to be able to access it, that much is true. But PB cannot share it with itself :slightly_smiling_face: Somebody else must share it with PB first.

Libraries/LibVT/TerminalWidget.cpp
zlotny: You mean without checking if it text first? It would try to paste as string the bitmap contents. I figured out that was not what we wanted while pasting

zlotny:
```
Terminal(8): First byte doesn't make sense, bytes: ��������������������������������������������������������������������
USERSPACE(8) ASSERTION FAILED: first_byte_makes_sense
../../AK/Utf8View.cpp:190
[colonel(0:0)]: Terminating due to signal 6

```

Libraries/LibGfx/Bitmap.cpp
bugaevc: I don't think there's any standard Unix/POSIX APIs that let you do what I want. On Linux, I bet you can `mmap("/proc/self/mem")` to achieve it. Maybe we should add a custom syscall, say `copy_mmap(void* old, size_t size)`?

Libraries/LibVT/TerminalWidget.cpp
bugaevc: No, checking that it's text is fine. But why check that it's not empty?

zlotny: Ohh I see. It was introduced in dbf8d6bc1eab19a57e9e6e7f81aa255505e971d4

I didn't want to touch things not strictly related to the PR

Libraries/LibGUI/Clipboard.cpp
zlotny: Okay I now see what I did. I'm not even using the shared buffer on PB, just interfacing with clipboard _Bitmap-wise_. So no need to share. I'll remove that. Thanks!

Libraries/LibGfx/Bitmap.cpp
zlotny: I don't quite understand this kind of things but shouldn't we just don't seal() SharedBuffers we could want to write later. From what I talked with Andreas it seemed to be the problem.

Like making a call specifically to allow writing on SharedBuffer's contents isn't it __hammering it__? Like, it's sealed for some reason.

Again, don't know what I'm saying, just trying to tie things together in my head. Still learning!.

PS: Andreas is tagged in here so he probably will answer you better :sweat_smile:

bugaevc: Don't worry, I don't understand what you're saying either 😀

All I'm saying is it would be nice to awoid the physical copy (while still making a virtual copy) using CoW semantics, with some help from the kernel.

Libraries/LibVT/TerminalWidget.cpp
bugaevc: I see.

Libraries/LibGfx/Bitmap.cpp
awesomekling: Let's leave that as a nice-to-have future improvement for now :)

Libraries/LibGUI/Clipboard.cpp
```diff
@@ -62,6 +64,25 @@ Clipboard::DataAndType Clipboard::data_and_type() const
     return { data, type };
 }
 
+AK::RefPtr<Gfx::Bitmap> Clipboard::get_bitmap(Gfx::BitmapFormat format, Gfx::Size size)
```
awesomekling, line 67: No need to qualify `AK::` basically ever, just say `RefPtr`.

More importantly, it seems unhealthy that the caller has to provide both `format` and `size`.
The caller should only have to say "gimme a bitmap" and the `Clipboard` class either says "OK here u go" or `nullptr`.

Libraries/LibGUI/Clipboard.cpp
```diff
@@ -62,6 +64,25 @@ Clipboard::DataAndType Clipboard::data_and_type() const
...
+    if (!shared_buffer) {
+        dbgprintf("GUI::Clipboard::data() failed to attach to the shared buffer\n");
+        return {};
+    }
+    if (response->content_size() > shared_buffer->size()) {
+        dbgprintf("GUI::Clipboard::data() clipping contents size is greater than shared buffer size\n");
+        return {};
+    }
+    auto bitmap = Gfx::Bitmap::create_with_shared_buffer(format, *shared_buffer, size);
+    return bitmap;
```
awesomekling, line 83: No need for a temporary here:
```suggestion
    return Gfx::Bitmap::create_with_shared_buffer(format, *shared_buffer, size);
```

Libraries/LibGUI/Clipboard.cpp
```diff
@@ -79,10 +100,18 @@ void Clipboard::set_data(const StringView& data, const String& type)
     WindowServerConnection::the().send_sync<Messages::WindowServer::SetClipboardContents>(shared_buffer->shbuf_id(), data.length(), type);
 }
 
+void Clipboard::set_data(const RefPtr<Gfx::Bitmap> bitmap, const String& type)
```
awesomekling, line 103: `RefPtr` here implies that passing null is a valid action, but it's not.
Instead, let's use `const Gfx::Bitmap&` to indicate that a non-null bitmap is required, and that this function does not take ownership. It's basically never required to pass a smart pointer as an argument. If the callee wants to take ownership of something, it can instantiate a `RefPtr` or `NonnullRefPtr` from a `T*` or `T&` at will :)

Also, why is the `type` parameter needed here? Do we need different types of bitmaps somehow?

Libraries/LibGUI/Clipboard.cpp
```diff
@@ -79,10 +100,18 @@ void Clipboard::set_data(const StringView& data, const String& type)
...
+
+    WindowServerConnection::the().send_sync<Messages::WindowServer::SetClipboardContents>(shared_buffer->shbuf_id(), shared_buffer->size(), type);
+}
+
 void Clipboard::did_receive_clipboard_contents_changed(Badge<WindowServerConnection>, const String& data_type)
 {
     if (on_content_change)
         on_content_change(data_type);
 }
-
```
awesomekling, line 87: This seems like an unnecessary whitespace change that `clang-format` should not do. Are you using the project's `.clang-format` configuration file (found in the project root)?

Applications/PaintBrush/main.cpp
```diff
@@ -96,7 +97,27 @@ int main(int argc, char** argv)
         return;
     }));
 
-    menubar->add_menu("Edit");
+    auto& edit_menu = menubar->add_menu("Edit");
+
+    auto paste_action = GUI::CommonActions::make_paste_action(
+        [&](const GUI::Action&) {
+            auto old_bitmap = RefPtr<Gfx::Bitmap>(paintable_widget.bitmap());
+            auto clipboard_bitmap = GUI::Clipboard::the().get_bitmap(old_bitmap->format(), old_bitmap->size());
```
awesomekling, line 105: I don't think we should assume the old bitmap and the clipping bitmap have identical format and size. :)

Applications/PaintBrush/main.cpp
```diff
@@ -96,7 +97,27 @@ int main(int argc, char** argv)
...
+            if (!clipboard_bitmap)
+                return;
+            auto new_bitmap = Gfx::Bitmap::cloned_from(clipboard_bitmap);
+            paintable_widget.set_bitmap(*new_bitmap);
```
awesomekling, line 109: Going forward, pasting shouldn't just replace what I'm working on.
Instead, it should become a new layer that can be moved around etc. :)

Libraries/LibGfx/Bitmap.h
```diff
@@ -54,6 +54,7 @@ class Bitmap : public RefCounted<Bitmap> {
     static RefPtr<Bitmap> create_wrapper(BitmapFormat, const Size&, size_t pitch, RGBA32*);
     static RefPtr<Bitmap> load_from_file(const StringView& path);
     static RefPtr<Bitmap> create_with_shared_buffer(BitmapFormat, NonnullRefPtr<SharedBuffer>&&, const Size&);
+    static RefPtr<Bitmap> cloned_from(RefPtr<Bitmap>);
```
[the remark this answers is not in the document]
zlotny: Done. It indeed does make more sense to request a clone from the bitmap itself.

Libraries/LibGfx/Bitmap.cpp
```diff
@@ -61,6 +61,13 @@ RefPtr<Bitmap> Bitmap::create_purgeable(BitmapFormat format, const Size& size)
     return adopt(*new Bitmap(format, size, Purgeable::Yes));
 }
 
+RefPtr<Bitmap> Bitmap::cloned_from(RefPtr<Bitmap> bitmap)
+{
+    auto cloned_bitmap = Bitmap::create(bitmap->format(), bitmap->size());
```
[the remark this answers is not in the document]
zlotny: Fixed

Libraries/LibGUI/Clipboard.cpp
zlotny: You're right. I'm just getting my hand on managing refs + ptrs and I was a bit lost when I made that. Now I understand clearly what you're saying (thanks, books).

Type is now fixed to bitmap so we match the IPC call at WindowServer.ipc:63

Libraries/LibGUI/Clipboard.cpp
zlotny: I'm using the one on the root (or that is what I configured) but that somewhat slipped in there. Formatting it again removed that blank line so I don't know when I introduced that there. Anyway now it isn't there

Applications/PaintBrush/main.cpp
zlotny: This is where it gets messy for me. I understand that the clipboard should hold info of format an size of the bitmap it has right now but I need a bit of time to think about how to do it properly.

Sorry about that. Noob things. Marking this as a draft to work on that!

Applications/PaintBrush/main.cpp
zlotny: Yeah, however that's a quite expensive feature on PaintBrush right now and I wanted to focus my less than optimal focus and knowledge on this particular task. However the next thing should be having at least support for 2 layers on PaintBrush, mspaint style :)

Applications/PaintBrush/main.cpp
zlotny: Okay so now clipboard knows the size and format of the last copied thing which makes applications stop assuming what size and format a bitmap from it is.

However I have to ask here because I couldn't find any solution for this. The contents of Clipboard::m_current_bitmap_format and Clipboard::m_current_bitmap_size are perfectly right if the same PaintBrush that copied into the clipboard tries to get the values.

However if some other PaintBrush (or application) tries to get_bitmap those 2 members always have default values. 

I'm assuming I'm copying values onto those members. Why are they dorking when other app calls the Clipboard singleton?

Sorry about the question ; I have the intuition that is just a trivial thing that I don't know.

zlotny: FML. They're freaking different applications. Ignore everything above.
image

I'll try later to add a call to the WindowServer via IPC and save the bitmap properties on the WindowServer to then get them.

zlotny: Okay now it's ready. It stores the contents, format, width and height of the bitmap on the window server clipboard. not the client's one... duh!

Servers/WindowServer/Clipboard.h
```diff
@@ -46,8 +46,13 @@ class Clipboard {
     const u8* data() const;
     int size() const;
 
+    const String& bitmap_format() const { return m_bitmap_format; }
```
awesomekling, line 49: This should return a `Gfx::BitmapFormat`

Servers/WindowServer/WindowServer.ipc
```diff
@@ -62,6 +62,9 @@ endpoint WindowServer = 2
     GetClipboardContents() => (i32 shbuf_id, i32 content_size, String content_type)
     SetClipboardContents(i32 shbuf_id, i32 content_size, String content_type) => ()
 
+    GetClipboardBitmap() => (i32 shbuf_id, i32 content_size, String format, i32 width, i32 height)
+    SetClipboardBitmap(i32 shbuf_id, i32 content_size, i32 width, i32 height, String bitmap_format) => ()
```
awesomekling, line 66: The width+height in both cases here should be represented by a `Gfx::Size`.

Libraries/LibGUI/Clipboard.h
```diff
@@ -39,13 +41,14 @@ class Clipboard {
...
     void set_data(const StringView&, const String& data_type = "text");
-
+    void set_data(const Gfx::Bitmap&);
     struct DataAndType {
         String data;
         String type;
     };
 
     DataAndType data_and_type() const;
+    RefPtr<Gfx::Bitmap> get_bitmap();
```
awesomekling, line 51: Can this be `const` qualified? Getting a bitmap from the clipboard seems like a non-mutating operation.

Applications/PaintBrush/main.cpp
```diff
@@ -96,7 +97,26 @@ int main(int argc, char** argv)
         return;
     }));
 
-    menubar->add_menu("Edit");
+    auto& edit_menu = menubar->add_menu("Edit");
+
+    auto paste_action = GUI::CommonActions::make_paste_action(
+        [&](const GUI::Action&) {
+            auto old_bitmap = RefPtr<Gfx::Bitmap>(paintable_widget.bitmap());
```
awesomekling, line 104: Remind me why we need to hold on to the old bitmap here

Applications/PaintBrush/main.cpp
```diff
@@ -96,7 +97,26 @@ int main(int argc, char** argv)
...
+            auto clipboard_bitmap = GUI::Clipboard::the().get_bitmap();
+            if (!clipboard_bitmap)
+                return;
+            paintable_widget.set_bitmap(*clipboard_bitmap->cloned());
```
awesomekling, line 108: This will crash if the bitmap cloning fails for whatever reason (e.g OOM)
Not that we handle this very gracefully in many places, but since bitmaps tend to be pretty chunky, maybe it's worth having a fallback path here.

zlotny:
I tried adding a way to encode/decode Gfx::Size and Gfx::BitmapFormat but I didn't manage to get it right and now all the WindowServer stuff is just broken:

image

I don't know what I've get wrong and I can't get it right, frankly. I'll force push with the clipboard working (it seems to copy and paste even with the super glitchy screen between 2 PaintBrush instances) ; I don't want to waste your time more.

PR #2033, opened 2020-04-30
LibJS: Implement (no-op) debugger statement

linusg:
https://tc39.es/ecma262/#sec-debugger-statement

> Evaluating a DebuggerStatement may allow an implementation to cause a breakpoint when run under a debugger. If a debugger is not present or active this statement has no observable effect.

Anyone wanna add a JS debugger to HackStudio?


alimpfard:
That's gonna be a very annoying thing to implement, what with the billions of implicit execution stuff that the Interpreter has...

sunverwerth:
This would be easier to implement once we have a VM for javascript.

linusg:
who am i kidding, the only debugger I use is `console.log()` anyway!

linusg:
@awesomekling may I ask for a review pls :)

awesomekling:
\o/

PR #2034, opened 2020-04-30
LibJS: Add Symbol objects (without Object integration)

mattco98:
These commits add support for the symbol primitive, as well as the Symbol object. It also modifies the Object class to use `PropertyName`s as `HashMap` keys instead of `FlyString`s to support inserting symbols into objects.

This is a step up from what I've been doing -- I thought it'd be fun to take a stab at something a bit more challenging (and it was!). I learned a lot, but I don't have a ton of experience with C++, so it's probably far from perfect. Everything works, but there were definitely certain decisions that I wasn't 100% sure how to handle. 

For example: I changed the `HashMap`s from `FlyString` keys to `PropertyName` keys. This certainly makes it nice to use from a programming perspective, but is it a large performance hit since we now have to construct a new struct for every property? Would it be better to have two maps, one for `FlyString` and one for `PrimitiveSymbol`? I'm not quite sure.

### DIFF
--- a/Libraries/LibJS/Runtime/SymbolObject.cpp
+++ b/Libraries/LibJS/Runtime/SymbolObject.cpp
@@ -1,0 +1,122 @@
+/*
+ * Copyright (c) 2020, Matthew Olsson <matthewcolsson@gmail.com>
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include <LibJS/Heap/Heap.h>
+#include <LibJS/Interpreter.h>
+#include <LibJS/Runtime/GlobalObject.h>
+#include <LibJS/Runtime/Symbol.h>
+#include <LibJS/Runtime/SymbolObject.h>
+#include <LibJS/Runtime/SymbolPrototype.h>
+#include <LibJS/Runtime/Value.h>
+
+namespace JS {
+
+HashMap<String, Value> SymbolObject::s_global_symbol_map;
+
+Value SymbolObject::s_well_known_iterator;
+Value SymbolObject::s_well_known_async_terator;
+Value SymbolObject::s_well_known_match;
+Value SymbolObject::s_well_known_match_all;
+Value SymbolObject::s_well_known_replace;
+Value SymbolObject::s_well_known_search;
+Value SymbolObject::s_well_known_split;
+Value SymbolObject::s_well_known_has_instance;
+Value SymbolObject::s_well_known_is_concat_spreadable;
+Value SymbolObject::s_well_known_unscopables;
+Value SymbolObject::s_well_known_species;
+Value SymbolObject::s_well_known_to_primtive;
+Value SymbolObject::s_well_known_to_string_tag;
+
+SymbolObject* SymbolObject::create(GlobalObject& global_object, Symbol& primitive_symbol)
+{
+    return global_object.heap().allocate<SymbolObject>(primitive_symbol, *global_object.symbol_prototype());
+}
+
+SymbolObject::SymbolObject(Symbol& symbol, Object& prototype)
+    : Object(&prototype)
+    , m_symbol(symbol)
+{
+}
+
+SymbolObject::~SymbolObject()
+{
+}
+
+Value SymbolObject::get_global(Interpreter& interpreter, String description)
+{
+    auto global_symbol = s_global_symbol_map.get(description);
+    if (global_symbol.has_value())
+        return global_symbol.value();
+
+    auto symbol = js_symbol(interpreter, description, true);
+    s_global_symbol_map.set(description, symbol);
+    return Value(symbol);
+}
+
+void SymbolObject::initialize_well_known_symbols(Interpreter& interpreter)
+{
+    SymbolObject::s_well_known_iterator = Value(js_symbol(interpreter, "Symbol.iterator", false));
+    SymbolObject::s_well_known_async_terator = Value(js_symbol(interpreter, "Symbol.asyncIterator", false));
+    SymbolObject::s_well_known_match = Value(js_symbol(interpreter, "Symbol.match", false));
+    SymbolObject::s_well_known_match_all = Value(js_symbol(interpreter, "Symbol.matchAll", false));
+    SymbolObject::s_well_known_replace = Value(js_symbol(interpreter, "Symbol.replace", false));
+    SymbolObject::s_well_known_search = Value(js_symbol(interpreter, "Symbol.search", false));
+    SymbolObject::s_well_known_split = Value(js_symbol(interpreter, "Symbol.split", false));
+    SymbolObject::s_well_known_has_instance = Value(js_symbol(interpreter, "Symbol.hasInstance", false));
+    SymbolObject::s_well_known_is_concat_spreadable = Value(js_symbol(interpreter, "Symbol.isConcatSpreadable", false));
+    SymbolObject::s_well_known_unscopables = Value(js_symbol(interpreter, "Symbol.unscopables", false));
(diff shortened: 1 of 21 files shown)


mattco98:
Whoops, build fails. Think I committed something I didn't mean to

Libraries/LibJS/Runtime/PrimitiveSymbol.cpp
```diff
@@ -0,0 +1,53 @@
+/*
+ * Copyright (c) 2020, Andreas Kling <kling@serenityos.org>
```
awesomekling, line 2: I didn't write this :)

Libraries/LibJS/Runtime/PrimitiveSymbol.h
```diff
@@ -0,0 +1,55 @@
+/*
+ * Copyright (c) 2020, Andreas Kling <kling@serenityos.org>
```
awesomekling, line 2: I didn't write this :)

Libraries/LibJS/Runtime/PrimitiveSymbol.h
```diff
@@ -0,0 +1,55 @@
...
+    virtual const char* class_name() const override { return "PrimitiveSymbol"; }
+
+    const String m_description;
+    const bool m_is_global;
+};
+
+PrimitiveSymbol* js_symbol(Heap&, String description, bool is_global);
+PrimitiveSymbol* js_symbol(Interpreter&, String description, bool is_global);
+
+}
```
awesomekling, line 55: Please configure your editor to add a newline at the end of files.

Libraries/LibJS/Runtime/PrimitiveSymbol.h
```diff
@@ -0,0 +1,55 @@
...
+
+namespace JS {
+
+class PrimitiveSymbol final : public Cell {
+public:
+    PrimitiveSymbol(String, bool);
+    virtual ~PrimitiveSymbol();
+
+    String description() { return m_description; }
+    const String description() const { return m_description; }
```
awesomekling, line 40: Since the symbol object is not transferring ownership of the string, we can just return it as a const reference.
Also we don't need a const and non-const version of this :)

```suggestion
const String& description() const { return m_description; }
```

Libraries/LibJS/Runtime/PrimitiveSymbol.h
```diff
@@ -0,0 +1,55 @@
...
+    bool is_global() { return m_is_global; }
+    String to_string() { return String::format("Symbol(%s)", description().characters()); }
+    const String to_string() const { return String::format("Symbol(%s)", description().characters()); }
```
awesomekling, line 43: No need for both const and non-const versions here; the const version is enough.

Libraries/LibJS/Runtime/PrimitiveSymbol.h
```diff
@@ -0,0 +1,55 @@
...
+namespace JS {
+
+class PrimitiveSymbol final : public Cell {
+public:
+    PrimitiveSymbol(String, bool);
+    virtual ~PrimitiveSymbol();
+
+    String description() { return m_description; }
+    const String description() const { return m_description; }
+    bool is_global() { return m_is_global; }
```
awesomekling, line 41: This should be const-qualified since it does not mutate the object.

Libraries/LibJS/Runtime/PrimitiveSymbol.h
```diff
@@ -0,0 +1,55 @@
...
+ */
+
+#pragma once
+
+#include <AK/String.h>
+#include <LibJS/Runtime/Cell.h>
+
+namespace JS {
+
+class PrimitiveSymbol final : public Cell {
```
awesomekling, line 34: The reason `PrimitiveString` is called `PrimitiveString` is because `String` is already a ubiquitous name in the codebase. We can just call this `Symbol` :)

Libraries/LibJS/Runtime/SymbolObject.cpp
```diff
@@ -0,0 +1,102 @@
...
+    auto global_symbol = s_global_symbol_map.get(description);
+    if (global_symbol.has_value())
+        return global_symbol.value();
+
+    auto symbol = js_symbol(interpreter, description, true);
+    s_global_symbol_map.set(description, symbol);
+    return Value(symbol);
+}
+
+void SymbolObject::initialize_wellknown_symbols(Interpreter& interpreter)
```
awesomekling, line 79: Let's write `well known` as two words :)

Libraries/LibJS/Runtime/SymbolObject.h
```diff
@@ -0,0 +1,95 @@
...
+
+class SymbolObject : public Object {
+public:
+    static SymbolObject* create(GlobalObject&, PrimitiveSymbol&);
+
+    SymbolObject(PrimitiveSymbol&, Object& prototype);
+    virtual ~SymbolObject() override;
+
+    PrimitiveSymbol* primitive_symbol() { return &m_symbol; }
+    const PrimitiveSymbol* primitive_symbol() const { return &m_symbol; }
```
awesomekling, line 42: Let's return a `PrimitiveSymbol&` instead of `PrimitiveSymbol*` here since we can never return `nullptr`.

Libraries/LibJS/Runtime/SymbolObject.h
```diff
@@ -0,0 +1,95 @@
...
+
+    String description() { return m_symbol.description(); }
+    const String description() const { return m_symbol.description(); }
```
awesomekling, line 45: No need for both const and non-const (const is enough!), also we can just return `const String&`.

Libraries/LibJS/Runtime/SymbolObject.h
```diff
@@ -0,0 +1,95 @@
...
+    bool is_global() { return m_symbol.is_global(); }
```
awesomekling, line 46: This should be const-qualified.

Libraries/LibJS/Runtime/PropertyName.h
```diff
@@ -36,41 +37,54 @@ class PropertyName {
...
+struct AK::Traits<JS::PropertyName> : public GenericTraits<JS::PropertyName> {
+    static unsigned hash(const JS::PropertyName& key)
+    {
+        ASSERT(key.is_valid());
+        if (key.is_string())
+            return key.as_string().hash();
+        if (key.is_number())
+            return key.as_number();
+        // FIXME: Better symbol hashing
+        return pair_int_hash(key.as_symbol().description().hash(), reinterpret_cast<intptr_t>(&key.as_symbol()));
```
mattco98, line 88: This is making the build fail right now. I don't know a good way to hash Symbols because they don't have any data members that make them unique.

Libraries/LibJS/Runtime/PropertyName.h
```diff
@@ -36,41 +37,54 @@ class PropertyName {
...
+
+template<>
+struct AK::Traits<JS::PropertyName> : public GenericTraits<JS::PropertyName> {
+    static unsigned hash(const JS::PropertyName& key)
+    {
+        ASSERT(key.is_valid());
+        if (key.is_string())
+            return key.as_string().hash();
+        if (key.is_number())
+            return key.as_number();
```
awesomekling, line 86:
```suggestion
            return int_hash(key.as_number());
```

Libraries/LibJS/Runtime/PropertyName.h
awesomekling: You can use `ptr_hash()` on pointers instead of casting to `intptr_t` which may confuse the host-side build.

Libraries/LibJS/Runtime/Shape.h
```diff
@@ -84,34 +84,34 @@ class Shape final : public Cell {
...
 
-    Optional<PropertyMetadata> lookup(const FlyString&) const;
-    const HashMap<FlyString, PropertyMetadata>& property_table() const;
+    Optional<PropertyMetadata> lookup(const PropertyName&) const;
+    const HashMap<PropertyName, PropertyMetadata>& property_table() const;
     size_t property_count() const;
 
     struct Property {
-        FlyString key;
+        PropertyName key;
```
awesomekling, line 92: `PropertyName` is a large object (12 bytes compared to `FlyString` being 4 bytes) that was never meant for persistent storage like this. In this context, the numeric index part of `PropertyName` is completely dead weight since shapes are unconcerned with indexes. The type field of `PropertyName` is also way too huge and really only needs a single bit.

Let's instead find a way to store either a symbol or a string here. Perhaps we could have something like a `SymbolOrString` class that is internally a pointer to a `StringImpl` (interned through `FlyString`) and using the lowest bit of the pointer to indicate whether it's a string or a symbol?

mattco98:
Alright, so I've been working on integrating symbols into the object system, and though I haven't managed to yet, _if_ I do it'll be quite a large PR, as there are (obviously) many areas it affects. I've decided it's best to just split those changes off from this PR, especially as I'm still not 100% confident that I'll be able to do it. I'll give it my best though :). My latest force push just removed what use to be the second commit in this PR.

Libraries/LibJS/Runtime/SymbolObject.cpp
```diff
@@ -0,0 +1,102 @@
...
+    SymbolObject::s_well_known_match_all = Value(js_symbol(interpreter, "Symbol.matchAll", false));
+    SymbolObject::s_well_known_replace = Value(js_symbol(interpreter, "Symbol.replace", false));
+    SymbolObject::s_well_known_search = Value(js_symbol(interpreter, "Symbol.search", false));
+    SymbolObject::s_well_known_split = Value(js_symbol(interpreter, "Symbol.split", false));
+    SymbolObject::s_well_known_has_instance = Value(js_symbol(interpreter, "Symbol.hasInstance", false));
+    SymbolObject::s_well_known_is_concat_spreadable = Value(js_symbol(interpreter, "Symbol.isConcatSpreadable", false));
+    SymbolObject::s_well_known_unscopables = Value(js_symbol(interpreter, "Symbol.unscopables", false));
+    SymbolObject::s_well_known_species = Value(js_symbol(interpreter, "Symbol.species", false));
+    SymbolObject::s_well_known_to_primtive = Value(js_symbol(interpreter, "Symbol.toPrimitive", false));
+    SymbolObject::s_well_known_to_string_tag = Value(js_symbol(interpreter, "Symbol.toStringTag", false));
```
awesomekling, line 93: Who protects these from garbage collection?

mattco98: Ahh, I didn't consider garbage collection. I assume that because they're static, I couldn't add them to a `visit_children` method. Would the best place to do that be `Interpreter.gather_roots`?

awesomekling: Sure, perhaps we could add a public static function to `SymbolObject`, and then call that from `Interpreter::gather_roots()`. That way the values can still live in `SymbolObject.cpp` :)

Libraries/LibJS/Runtime/SymbolObject.h
```diff
@@ -0,0 +1,95 @@
...
+    static Value well_known_to_string_tag() { return s_well_known_to_string_tag; };
+
+private:
+    virtual void visit_children(Visitor&) override;
+    virtual const char* class_name() const override { return "SymbolObject"; }
+    virtual bool is_symbol_object() const override { return true; }
+
+    Symbol& m_symbol;
+
+    static HashMap<String, Value> s_global_symbol_map;
```
awesomekling, line 78: Wait, how are these protected from GC?

PR #2035, opened 2020-04-30
Applications: TextEditor INI file syntax highlighter

asliturk:
INI file syntax highlighter for TextEditor and new 32px icons.


nvella:
Love the icons 😁 Clear inspiration from 9x/XP

PR #2036, opened 2020-05-01
LibGUI: File picker and manager

BenWiederhake:
Many small consistency changes, most notably:
- Fix GUI elements that didn't get initialized properly (several very small commits)
- Rename ItemView to IconView, because the hover text says "Icon View" (one very large commit)

### DIFF
--- a/Libraries/LibGUI/AbstractView.h
+++ b/Libraries/LibGUI/AbstractView.h
@@ -34,7 +34,6 @@ namespace GUI {
 
 class AbstractView : public ScrollableWidget {
     C_OBJECT_ABSTRACT(AbstractView)
-    friend class Model;
 
 public:
     void set_model(RefPtr<Model>);
--- a/Libraries/LibGUI/MultiView.cpp
+++ b/Libraries/LibGUI/MultiView.cpp
@@ -106,9 +106,9 @@ MultiView::MultiView()
             on_drop(index, event);
     };
 #endif
-    set_view_mode(ViewMode::Icon);
 
     build_actions();
+    set_view_mode(ViewMode::Icon);
 }
 
 MultiView::~MultiView()
@@ -123,16 +123,19 @@ void MultiView::set_view_mode(ViewMode mode)
     update();
     if (mode == ViewMode::Table) {
         set_active_widget(m_table_view);
+        m_view_as_table_action->set_checked(true);
         return;
     }
 #ifdef MULTIVIEW_WITH_COLUMNSVIEW
     if (mode == ViewMode::Columns) {
         set_active_widget(m_columns_view);
+        m_view_as_columns_action->set_checked(true);
         return;
     }
 #endif
     if (mode == ViewMode::Icon) {
         set_active_widget(m_icon_view);
+        m_view_as_icons_action->set_checked(true);
         return;
     }
     ASSERT_NOT_REACHED();
--- a/Libraries/LibGUI/MultiView.cpp
+++ b/Libraries/LibGUI/MultiView.cpp
@@ -40,14 +40,14 @@ MultiView::MultiView()
 {
     set_active_widget(nullptr);
     set_content_margins({ 2, 2, 2, 2 });
-    m_item_view = add<ItemView>();
+    m_icon_view = add<IconView>();
     m_table_view = add<TableView>();
 
 #ifdef MULTIVIEW_WITH_COLUMNSVIEW
     m_columns_view = add<ColumnsView>();
 #endif
 
-    m_item_view->on_activation = [&](auto& index) {
+    m_icon_view->on_activation = [&](auto& index) {
         if (on_activation)
             on_activation(index);
     };
@@ -66,7 +66,7 @@ MultiView::MultiView()
         if (on_selection_change)
             on_selection_change();
     };
-    m_item_view->on_selection_change = [this] {
+    m_icon_view->on_selection_change = [this] {
         if (on_selection_change)
             on_selection_change();
     };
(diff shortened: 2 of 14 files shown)


Libraries/LibGUI/AbstractView.h
```diff
@@ -34,7 +34,7 @@ namespace GUI {
 
 class AbstractView : public ScrollableWidget {
     C_OBJECT_ABSTRACT(AbstractView)
-    friend class Model;
+    //friend class Model;
```
zlotny, line 37: Just nitpicking but in CONTRIBUTING.md it's advised not to include commented out code.

```suggestion
```

Libraries/LibGUI/FilePicker.h
```diff
@@ -74,7 +74,7 @@ class FilePicker final : public Dialog {
     RefPtr<Label> m_preview_image_label;
     RefPtr<Label> m_preview_name_label;
     RefPtr<Label> m_preview_geometry_label;
-    Mode m_mode { Mode::Open };
+    Mode m_mode;
```
awesomekling, line 77: Let's always prefer initialization at the declaration instead of at construction.
Otherwise there is a time window where the value is uninitialized and it's nicer if we can just avoid that.

Libraries/LibGUI/AbstractView.h
awesomekling: Indeed, just remove it :)

Libraries/LibGUI/MultiView.cpp
```diff
@@ -106,9 +106,9 @@ MultiView::MultiView()
             on_drop(index, event);
     };
 #endif
-    set_view_mode(ViewMode::Icon);
 
     build_actions();
+    m_view_as_icons_action->set_checked(true);
```
awesomekling, line 111: Why was `set_view_mode(ViewMode::Icon)` not making the button checked?
I think this change worsens code readability, as we are replacing a high-level operation ("go into icon view mode") with a low-level operation ("change the checked state of this specific button") so it's less obvious that everything will be set up correctly afterwards.

Libraries/LibGUI/AbstractView.h
BenWiederhake: Oh! I thought I did that, but apparently forgot. Will do!

Libraries/LibGUI/MultiView.cpp
BenWiederhake: Good point.

`m_view_as_icons_action->set_checked` calls `set_view_mode`. Meanwhile, `set_view_mode` is a public method, and can be called from the outside.
So I guess the best way around this is calling `action->set_checked()` from `set_view_mode`, and stopping the recursion by this already existing snippet in `set_view_mode`:
```
    if (m_view_mode == mode)
        return;
    m_view_mode = mode;
```

What do you think about that?

BenWiederhake: Fun fact: FileManager also initializes its state by calling [`action->set_checked(true);`](https://github.com/SerenityOS/serenity/blob/46b79eaad9389e438d99bc06e5beca4408abb089/Applications/FileManager/main.cpp#L837)

BenWiederhake:
I implemented all suggestions, thanks! My code for [`set_checked`](https://github.com/SerenityOS/serenity/pull/2036#discussion_r418526095) works, but if you have a better idea let me hear it :).

I know that extending a PR after the first submission is a bit sneaky, but I believe the new commits are close enough to be in scope. If you wish I can also create a new PR for these:
- DisplaySettings: Don't crash after 'pape'
- LibGUI: Fix off-by-one in TextEditor
- LibGUI: The UI calls it 'Table view', fix the name in the code

Libraries/LibGUI/MultiView.cpp
```diff
@@ -117,22 +117,27 @@ MultiView::~MultiView()
 
 void MultiView::set_view_mode(ViewMode mode)
 {
+    // This method recursively calls itself through `action->set_checked`.
+    // The following check is not just for optimization; it prevents an infinite loop.
```
awesomekling, line 121: FWIW this pattern is prevalent throughout the codebase, so it's not unique here :)

BenWiederhake:
> The off-by-one fixed is in ScrollableWidget, not TextEditor like the commit message says.

This was intentional, but I see how the commit message is confusing. I'll reword it.

BenWiederhake:
Changes since your last comment:
- The `set_view_mode` thing didn't actually work, since the call from the constructor was ignored. I agree that calling `action->set_checked` directly is bad, because then `set_active_widget` is skipped. Currently this is not necessary due to unrelated reasons, but I don't like to rely on "it happens to do the right thing". So I introduced a new argument to "force" `set_active_widget` to do an update.
- Fix commits that accidentally got squashed (grmpf)
- Reworded the ScrollableWidget fix
- New feature: sort FilePicker by name instead of "unpredictable inode"

I think that's finally it ^^'

BenWiederhake:
Reworded and rebased.

Libraries/LibGUI/MultiView.cpp
```diff
@@ -106,33 +106,42 @@ MultiView::MultiView()
             on_drop(index, event);
     };
 #endif
-    set_view_mode(ViewMode::Icon);
 
     build_actions();
+    // `m_view_mode` already is ViewMode::Icon. See `set_view_mode`.
+    set_view_mode(ViewMode::Icon, Update::Force);
```
awesomekling, line 112: Let's just start `m_view_mode` with `ViewMode::Invalid`. Then you won't need this awkward forced update logic :)

BenWiederhake: I don't think it's awkward, but I changed it.

BenWiederhake:
Changed and rebased.

With the current flurry of activity you probably have to rebase it again anyway, so I hope it's not an issue that I rebased before the latest commit on master (3331098). I believe that commit is incomplete and needs a bit more love :)

PR #2038, opened 2020-05-01
LibJS: String.prototype.lastIndexOf

kessejones:
Added lastIndexOf in String.prototype. :)

### DIFF
--- a/Libraries/LibJS/Runtime/StringPrototype.cpp
+++ b/Libraries/LibJS/Runtime/StringPrototype.cpp
@@ -81,6 +81,7 @@ StringPrototype::StringPrototype()
     put_native_function("substring", substring, 2, attr);
     put_native_function("includes", includes, 1, attr);
     put_native_function("slice", slice, 2, attr);
+    put_native_function("lastIndexOf", last_index_of, 1, attr);
 }
 
 StringPrototype::~StringPrototype()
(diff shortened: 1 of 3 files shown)


Libraries/LibJS/Tests/String.prototype.lastIndexOf.js
```diff
@@ -0,0 +1,27 @@
...
+    assert("".lastIndexOf("", -1) === 0);
+    assert("hello friends serenity".lastIndexOf("h", 10) === 0);
+    assert("hello friends serenity".lastIndexOf("l", 4) === 3);
+    assert("hello friends serenity".lastIndexOf("s", 13) === 12);
+    assert("hello".lastIndexOf("serenity") === -1);
+
+    console.log("PASS");
+} catch (err) {
+    console.log("FAIL: " + err);
+}
```
awesomekling, line 27: Please configure your editor to add a newline at the end of files.

Libraries/LibJS/Runtime/StringPrototype.cpp
```diff
@@ -425,4 +426,37 @@ Value StringPrototype::slice(Interpreter& interpreter)
...
+            from_index = 0;
+        if (from_index > max_index)
+            from_index = max_index;
+    }
+
+    for (ssize_t i = from_index; i >= 0; --i) {
+        auto part_view = string.substring_view(i, search_string.length());
+        if (part_view == search_string)
+            return Value((i32)i);
+    }
```
awesomekling, line 457: Definitely not the fastest backward string search algorithm in the world, but I suppose it will be ok for now. It's always nice to leave performance opportunities for our future selves to find :)

kessejones: Not really, but that's what I was able to think for now. :)

PR #2042, opened 2020-05-01
LibJS: Implement console.count() and console.countReset()

emanuele6:
count: https://developer.mozilla.org/en-US/docs/Web/API/Console/count
countReset: https://developer.mozilla.org/en-US/docs/Web/API/Console/countReset


emanuele6:
`m_counters` should live in the interpreter (maybe with the name of `m_console_counters`), to avoid this problem:

```js
var lol = console
console = 2
lol.count()
```

I'll fix it now, please excuse the PR spam, I'll try to avoid it next time.

PR #2047, opened 2020-05-01
LibJS: Implement most of the Reflect object

linusg:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Reflect

A few FIXMEs here and there, but I think this is a good start :^)

### DIFF
--- a/Libraries/LibJS/AST.cpp
+++ b/Libraries/LibJS/AST.cpp
@@ -1087,11 +1087,17 @@ PropertyName MemberExpression::computed_property_name(Interpreter& interpreter) 
     auto index = m_property->execute(interpreter);
     if (interpreter.exception())
         return {};
+
     ASSERT(!index.is_empty());
-    // FIXME: What about non-integer numbers tho.
-    if (index.is_number() && index.to_i32() >= 0)
-        return PropertyName(index.to_i32());
-    return PropertyName(index.to_string());
+
+    if (!index.to_number().is_finite_number())
+        return PropertyName(index.to_string());
+
+    auto index_as_double = index.to_double();
+    if (index_as_double < 0 || (i32)index_as_double != index_as_double)
+        return PropertyName(index.to_string());
+
+    return PropertyName(index.to_i32());
 }
 
 String MemberExpression::to_string_approximation() const
--- a/Libraries/LibJS/Runtime/Object.cpp
+++ b/Libraries/LibJS/Runtime/Object.cpp
@@ -272,6 +272,12 @@ Value Object::get_by_index(i32 property_index) const
 
     const Object* object = this;
     while (object) {
+        if (is_string_object()) {
+            auto& string = static_cast<const StringObject*>(this)->primitive_string().string();
+            if (property_index < (i32)string.length())
+                return js_string(heap(), string.substring(property_index, 1));
+            return js_undefined();
+        }
         if (static_cast<size_t>(property_index) < object->m_elements.size()) {
             auto value = object->m_elements[property_index];
             if (value.is_empty())
@@ -397,6 +403,8 @@ bool Object::has_own_property(const FlyString& property_name) const
     bool ok;
     i32 property_index = property_name.to_int(ok);
     if (ok && property_index >= 0) {
+        if (is_string_object())
+            return property_index < (i32)static_cast<const StringObject*>(this)->primitive_string().string().length();
         if (static_cast<size_t>(property_index) >= m_elements.size())
             return false;
         return !m_elements[property_index].is_empty();
--- a/Libraries/LibJS/Runtime/Object.cpp
+++ b/Libraries/LibJS/Runtime/Object.cpp
@@ -177,7 +177,7 @@ void Object::set_shape(Shape& new_shape)
     m_shape = &new_shape;
 }
 
-void Object::put_own_property(Object& this_object, const FlyString& property_name, u8 attributes, Value value, PutOwnPropertyMode mode)
+bool Object::put_own_property(Object& this_object, const FlyString& property_name, u8 attributes, Value value, PutOwnPropertyMode mode)
 {
     auto metadata = shape().lookup(property_name);
     bool new_property = !metadata.has_value();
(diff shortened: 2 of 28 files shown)


Libraries/LibJS/Runtime/ReflectObject.cpp
```diff
@@ -0,0 +1,247 @@
...
+    if (!target)
+        return {};
+    return target->get_own_properties(*target, GetOwnPropertyMode::Key);
+}
+
+Value ReflectObject::prevent_extensions(Interpreter&)
+{
+    // FIXME: Implement me :^)
+    ASSERT_NOT_REACHED();
+}
```
linusg, line 216: This is probably as simple as adding a flag to the `JS::Object` class, but I'll leave that for another day.

Libraries/LibJS/Runtime/Value.cpp
```diff
@@ -183,6 +183,16 @@ double Value::to_double() const
...
 }
 
+size_t Value::to_size() const
+{
+    if (is_empty())
+        return 0;
+    auto number = to_number();
+    if (number.is_nan() || number.as_double() <= 0)
+        return 0;
+    return min(number.to_i32(), (i32)pow(2, 53) - 1);
```
awesomekling, line 193: Does the compiler recognize `pow(2, 53)` and constant-fold it? If not, we should use a constant here instead of making a call to `pow()` every time.

Also, `to_size()` feels a bit ambiguous since `Size` is commonly used in the codebase to refer to a width+height tuple. How about we call it `to_size_t()` instead?

Libraries/LibJS/AST.cpp
```diff
@@ -1088,8 +1088,7 @@ PropertyName MemberExpression::computed_property_name(Interpreter& interpreter)
     if (interpreter.exception())
         return {};
     ASSERT(!index.is_empty());
-    // FIXME: What about non-integer numbers tho.
-    if (index.is_number() && index.to_i32() >= 0)
+    if (index.to_number().is_finite_number() && (i32)index.to_double() == index.to_i32() && index.to_i32() >= 0)
```
awesomekling, line 1091: There's a lot of coercion hiding behind these calls. Perhaps we should coerce to number only once?

Libraries/LibJS/Runtime/Object.cpp
```diff
@@ -272,6 +302,12 @@ Value Object::get_by_index(i32 property_index) const
 
     const Object* object = this;
     while (object) {
+        if (is_string_object()) {
+            auto& string = static_cast<const StringObject*>(this)->primitive_string().string();
+            if (property_index < (i32)string.length())
+                return js_string(heap(), string.substring(property_index, 1));
+            return js_undefined();
+        }
```
awesomekling, line 310: Suddenly implementing indexed access to `StringObject` in the middle of an unrelated patch huh? ;)

Libraries/LibJS/Runtime/ObjectConstructor.cpp
```diff
@@ -108,47 +108,23 @@ Value ObjectConstructor::set_prototype_of(Interpreter& interpreter)
 
 Value ObjectConstructor::get_own_property_descriptor(Interpreter& interpreter)
 {
-    if (interpreter.argument_count() < 2)
```
awesomekling, line 111: What happens if we call `getOwnPropertyDescriptor()` with zero or one argument?

Libraries/LibJS/Runtime/ReflectObject.cpp
awesomekling: I think we can transition to a different `Shapé` and share the shape between all otherwise similar objects that have done a `preventExtensions()` :)

Libraries/LibJS/Runtime/Object.cpp
linusg: yeah, this belongs in a separate commit. very desperately wanted to get all the tests (I partially wrote beforehand) to pass :smile:

Libraries/LibJS/Runtime/Value.cpp
linusg: > Does the compiler recognize pow(2, 53) and constant-fold it? If not, we should use a constant here instead of making a call to pow() every time.

No idea, so I'll go with a constant number :)

> Also, to_size() feels a bit ambiguous since Size is commonly used in the codebase to refer to a width+height tuple. How about we call it to_size_t() instead?

agreed!

linusg:
Thanks for the review @awesomekling!

> I feel like we're being pretty sloppy about interpreter.arguments(i).something() where arguments(i) returns an empty value and something() will choke in one way or another. It's not exclusive to this patch, but it does add more of it. Would be good to find a pattern where we take care of this issue elegantly and concisely up-front in functions.

Huh - I was assuming `arguments(i)` would never return an empty value, unless we push an empty Value to the call frame's arguments internally - but why would we do that? Is there any situation where an empty value is a legitimate argument to a function - if now, how about this?

```
Value argument(size_t index) const
{
    if (m_call_stack.is_empty())
        return {};
    auto& arguments = m_call_stack.last().arguments;
	if (index < arguments.size())
		return js_undefined();
	auto value = arguments[index];
    if (value.is_empty())
		return js_undefined();
    return value;
}
```

Because unless there's a compelling reason to allow empty value arguments, manually handling empty values in each function using `arguments(i)` would be a massive pain in the ass.

Libraries/LibJS/Runtime/ObjectConstructor.cpp
linusg: - Zero: First argument (`target`) is `undefined` and the function throws `Uncaught exception: [TypeError]: First argument of Reflect.getOwnPropertyDescriptor() must be an object`
- One: Depends. Second argument (the property name) is `undefined` and will be stringified - so `"undefined"`
  So:
  - `Reflect.getOwnPropertyDescriptor({})` - `undefined`
  - `Reflect.getOwnPropertyDescriptor({ undefined: 123 })` - `{ "value": 123, "writable": true, "enumerable": true, "configurable": true }`
  - `Reflect.getOwnPropertyDescriptor("foo")` - `Uncaught exception: [TypeError]: First argument of Reflect.getOwnPropertyDescriptor() must be an object`

That's valid and even tested in the tests :)

awesomekling:
Oh wait lol, `argument(i)` returns undefined for out-of-range arguments. Never mind :)

PR #2058, opened 2020-05-02
Terminal: Add config for startup command

shannonbooth:
This is useful when working on a program as you can put the command into
the config file, instead of having to type it up each time on boot.


bugaevc:
I'd rather put it into my SystemServer config in that case. But I guess Terminal command customization can also be useful for... something, like maybe immediately telnet'ing to a different host.

PR #2062, opened 2020-05-02
16 connections should be enough for anybody

bugaevc:


awesomekling:
Countdown to "32 connections should be enough for everybody" starts now

bugaevc:
We have (counts) 10 autostarted WS clients now, so we should be safe for another half a year or so. Then we can add the option, bump WS to like 64, and leave the rest at 5.

PR #2063, opened 2020-05-02
js: Ignore property attributes for completion

linusg:
Only being able to complete enumerable properties is annoying, especially since we updated everything to use the correct attributes.

Most standard built-in objects are *not* enumerable.

- `A<tab>` - expecting `Array`, nothing happens.
- `B<tab>` - expecting `Boolean`, nothing happens.
- `c<tab>` - expecting `console`, nothing happens.
- ...
- `O<tab>` - expecting `Object`, nothing happens.


awesomekling:
Definitely better

PR #2066, opened 2020-05-02
Shell: Check for invalid #! in commands

### LINKED ISSUE #2046: Shell: invalid shebang should result in a `bad interpreter` error
Here is a screenshot.

bug_shell_serenity

ADKaster:
Closes #2046

Strategy is to open the file to look for '#!' and print out the first line as part of the error message if the kernel gives us ENOENT. This could happen if the argv[0] didn't exist, or its #! interpreter didn't exist.

This code looks rather C, but then so does the rest of Shell/main.cpp so, seems not that big a sin :)

### DIFF
--- a/Shell/main.cpp
+++ b/Shell/main.cpp
@@ -28,6 +28,7 @@
 #include "Parser.h"
 #include <AK/FileSystemPath.h>
 #include <AK/Function.h>
+#include <AK/ScopeGuard.h>
 #include <AK/StringBuilder.h>
 #include <LibCore/DirIterator.h>
 #include <LibCore/ElapsedTimer.h>
@@ -580,7 +581,7 @@ static bool handle_builtin(int argc, const char** argv, int& retval)
 
 class FileDescriptionCollector {
 public:
-    FileDescriptionCollector() { }
+    FileDescriptionCollector() {}
     ~FileDescriptionCollector() { collect(); }
 
     void collect()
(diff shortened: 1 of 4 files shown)


Shell/main.cpp
```diff
@@ -906,7 +907,21 @@ static int run_command(const String& cmd)
...
                 if (rc < 0) {
                     if (errno == ENOENT) {
-                        fprintf(stderr, "%s: Command not found.\n", argv[0]);
+                        int shebang_fd = open(argv[0], O_RDONLY);
+                        auto close_argv = ScopeGuard([shebang_fd]() { if (shebang_fd >= 0)  close(shebang_fd); });
+                        char shebang[256] {};
+                        ssize_t num_read = -1;
+                        if ((shebang_fd >= 0) && ((num_read = read(shebang_fd, shebang, sizeof(shebang))) >= 2) && (StringView(shebang).starts_with("#!"))) {
+                            ssize_t newline_pos = 2;
+                            for (; newline_pos < num_read; ++newline_pos) {
```
ADKaster, line 916: Should I just add `size_t AK::StringView::find_first_of(const StringView& chars)` instead? this works...

awesomekling: Yes, that would be a nice and expressive API. Let's return an `Optional<size_t>` though, in case we don't find anything :)

ADKaster: What, comparing against `AK::StringView::NO_MATCH` doesn't work for you? 😉. Sounds good, i'll push a slightly nicer looking fix.

PR #2069, opened 2020-05-02
LibGUI: Properly remove auto-destroyed child windows from reified_windows.

emilio:
I just spotted this while I saw you writing this code :-)


awesomekling:
Sharp eye, thank you! :)

PR #2070, opened 2020-05-02
LibJS: Set name of anonymous functions during assignment

linusg:
Apparently that's a thing :^)

### DIFF
--- a/Libraries/LibJS/AST.cpp
+++ b/Libraries/LibJS/AST.cpp
@@ -46,6 +46,23 @@
 
 namespace JS {
 
+static void update_function_name(Value& value, const FlyString& name)
+{
+    if (!value.is_object())
+        return;
+    auto& object = value.as_object();
+    if (object.is_function()) {
+        auto& function = static_cast<ScriptFunction&>(object);
+        if (function.name().is_empty())
+            function.set_name(name);
+    } else if (object.is_array()) {
+        auto& array = static_cast<Array&>(object);
+        for (size_t i = 0; i < array.elements().size(); ++i) {
+            update_function_name(array.elements()[i], name);
+        }
+    }
+}
+
 Value ScopeNode::execute(Interpreter& interpreter) const
 {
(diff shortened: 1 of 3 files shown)


alimpfard:
what if...you assign it to multiple things?
javascript is _weird_

alimpfard:
shot-2020-05-03_00-34-55
as if that explains it.
javascript is _weird_

linusg:
> what if...you assign it to multiple things?

Great question! I didn't even think about this (the implementation still gets it right!). The last variable name is supposed to be the final function name:

```js
a = b = function() {}
a.name; // "b"
b.name; // "b"
```

Will add a test case.

> javascript is weird

you keep saying that lol - I can't disagree...

alimpfard:
> The last variable name is supposed to be the final function name

It looks to more like `a = b = foo` is parsed as `a = (b = foo)`, and so the function gets `b` as its name...

> you keep saying that lol - I can't disagree...

The damn language has so many weird quirks;
e.g. why even give a name to an anonymous function? that makes no sense

linusg:
> shot-2020-05-03_00-34-55
> as if that explains it.

image

Damn, that's worse than I expected...

linusg:
```js
a = [function(){}, function(){}, function(){}]]
a[0].name; // "a"
a[1].name; // "a"
a[2].name; // "a"
```

ffs

alimpfard:
>a = [function(){}, function(){}, function(){}];
a[0].name; // "a"
a[1].name; // "a"
a[2].name; // "a"

what the...

awesomekling:
Based on countless experiences working on production browsers in the past, this sort of quirk usually exists because some website somewhere depended on it at some point, and browser vendors decided it was easier to maintain the behavior (and keep websites working) instead than switching to something sane. Then the Standards People came along and codified whatever the browsers were doing and boom, here we are. (Disclaimer: this particular thing may be an exception, I don't know) :)

PR #2071, opened 2020-05-02
LibJS: Add function default arguments

mattco98:
Adds the ability for function arguments to have default values. This works for standard functions as well as arrow functions. Default values are not printed in a `<function>.toString()` call, as nodes cannot print their source string representation.


awesomekling:
:+1:

PR #2074, opened 2020-05-02
WindowServer: Update desktop window rect on resolution change

### LINKED ISSUE #2073: WindowServer: Select Area is limited in unreasonable manner after resolution change
How to reproduce:
1. run Serenity with `./run`, a clean build from the master branch suffices.
2. open `DisplayProperties`, change the resolution something higher than what is configured in your setup.
3. Apply, close the `DisplayProperties` window. Try to select an area in the desktop. Stretch it as far as allowed by the new resolution, you should see you're actually limited to not the new resolution size.

A picture that demonstrates the problem: 
Screenshot_2020-05-02_23-25-31

linusg:
Fixes #2073.

cc @supercomputer7

### DIFF
--- a/Servers/WindowServer/WindowManager.cpp
+++ b/Servers/WindowServer/WindowManager.cpp
@@ -142,6 +142,8 @@ bool WindowManager::set_resolution(int width, int height)
     });
     if (success) {
         for_each_window([](Window& window) {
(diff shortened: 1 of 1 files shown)


awesomekling:
Strong patch

PR #2076, opened 2020-05-02
Kernel: fix pledge syscall applying new pledges when it fails

ToadKing:
If the exec promises fail to apply, then the normal promises should
not apply either. Add a test for this fixed functionality.


awesomekling:
Good fix!

PR #2079, opened 2020-05-03
LibCrypto: Correct RFC5246 un-padding behaviour

### LINKED ISSUE #2072: LibTLS: Concatenation of sucessive ApplicationData records produces invalid data
(just for documentation and because my memory is pure crap about these)

When loading https://templeos.org/Templeos-header.png we receive two app data records, both of which are verified to be consistent; however, the reassembled packet contains invalid data at the end of the first record (offset 0x3f15).

Things that could be wrong:
- bad AES padding decode (likely)
- reassembly issue (not very likely)
- ?

alimpfard:
The decrypted data is legally allowed to have any amount of padding, so long as it is block-aligned, we should not assume that padding bytes fall inside the same block, or that an entire block cannot be padding.

Fixes #2072

### DIFF
--- a/Libraries/LibCrypto/Cipher/Mode/Mode.h
+++ b/Libraries/LibCrypto/Cipher/Mode/Mode.h
@@ -79,11 +79,7 @@ protected:
         }
         case PaddingMode::RFC5246: {
             auto maybe_padding_length = data[size - 1];
-            if (maybe_padding_length >= T::block_size() - 1) {
-                // cannot be padding (the entire block cannot be padding)
-                return;
-            }
-            // FIXME: If we want to constant-time operations, this loop should not stop
+            // FIXME: If we want constant-time operations, this loop should not stop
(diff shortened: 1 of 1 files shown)


Libraries/LibCrypto/Cipher/Mode/Mode.h
```diff
@@ -79,11 +79,7 @@ class Mode {
         }
         case PaddingMode::RFC5246: {
             auto maybe_padding_length = data[size - 1];
-            if (maybe_padding_length >= T::block_size() - 1) {
-                // cannot be padding (the entire block cannot be padding)
-                return;
-            }
-            // FIXME: If we want to constant-time operations, this loop should not stop
+            // FIXME: If we want constant-time operations, this loop should not stop
```
awesomekling, line 82: (And yes, we definitely want constant-time operations :)

PR #2086, opened 2020-05-03
LibTLS: "Properly" handle the server dropping the connection

alimpfard:
Contrary to popular belief, not every implementation of TLS follows the specs.
Some of them just drop the connection without sending a proper close_notify, and we should handle that gracefully.

\*glares at google.com\*


awesomekling:
Excellent work!

PR #2088, opened 2020-05-03
Meta: Add configuration for setting up project in CLion

devsh0:
This commit adds a CMakeLists.txt file that will be used by CLion to
configure the project and documentation explaining the steps to follow.
Configuring CLion this way enables important features like code
completion and file search. The configuration isn't perfect. There are
source files for which CLion cannot pick up the headers and asks to
manually include them from certain directories. But for the most part,
it works all right.

### DIFF
--- a/Documentation/CLionConfiguration.md
+++ b/Documentation/CLionConfiguration.md
@@ -1,0 +1,19 @@
+## CLion Project Configuration
+
+Configuring CLion for development requires a CMakeLists.txt file. The one provided in Meta/CLion packs the bare minimum configuration so that you can take advantage of code completion and basic error checking. This is not necessarily the best setup, there could be better ways to configure the IDE. Below instructions are intended for people not familiar with CMake, or who don't want to get bothered setting up the IDE first.
+
+It's assumed that a directory named `serenity` contains all your source files. If you cloned the repository directly from GitHub, this is the directory that git creates to copy sources.
+
+- Create the project directory and cd to it: `mkdir serenity-project && cd serenity-project`
+- Move `serenity` to the project directory: `mv <path/to/serenity> .`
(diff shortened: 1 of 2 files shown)


MWGuy:
Love it ❤️

awesomekling:
Sure, if this is useful to the CLion users among us :)

linusg:
@devsh0 can we get rid of `Meta/CLion/CMakeLists.txt` now? Is `Documentation/CLionConfiguration.md` still up-to-date?

(see https://github.com/SerenityOS/serenity/commit/450a2a0f9ce2443bc06445ee018c9860d8cd5fc8)

devsh0:
I don't have the latest master yet. I think it should be okay to remove the CLion directory now. But let me double check by grabbing the latest changes.

devsh0:
Looks like the new CMakeLists file suffices to load the project in CLion, given that the toolchain has been built before and CLion is explicitly configured to use Cmake 3.17. Auto-completion works too. I should point out that there are some red markers indicating symbol not found errors (and a couple of other types of errors). It seems pretty okay. But do note that I develop in guest Linux. I'm not sure if that's relevant. I am not aware of the pitfalls, if there are any, to use WSL as the build environment.

devsh0:
@linusg Ignore my last comment, turns out that the new CMakeLists.txt doesn't help CLion at all. Previously it did maybe because the IDE had cached indexes. I think we should keep the CMakeLists.txt in Meta/CLion, as it continues to guide CLion to build the index properly and configure include directories. Adding a few entries to INCLUDE_DIRS in serenity/CMakeLists.txt should work too, but I think it's better to separate IDE specific patches from the main tree. Also, the CLion documentation needs minor changes. Will open a PR tomorrow.

PR #2092, opened 2020-05-03
LibJS: Add template literals

mattco98:
Adds fully functioning template literals. Because template literals contain expressions, most of the work has to be done in the Lexer rather than the Parser. And because of the complexity of template literals (expressions, nesting, escapes, etc), the Lexer needs to have some template-related state.

When entering a new template literal, a `TemplateLiteralStart` token is emitted. When inside a literal, all text will be parsed up until a `${` or `` ` `` (or `EOF`, but that's a syntax error) is seen, and then a `TemplateLiteralExprStart` token is emitted. At this point, the Lexer proceeds as normal, however it keeps track of the number of opening and closing curly braces it has seen in order to determine the close of the expression. Once it finds a matching curly brace for the `${`, a `TemplateLiteralExprEnd` token is emitted and the state is updated accordingly.

When the Lexer is inside of a template literal, but not an expression, and sees a `` ` ``, this must be the closing grave: a `TemplateLiteralEnd` token is emitted.

The state required to correctly parse template strings consists of a vector (for nesting) of two pieces of information: whether or not we are in a template expression (as opposed to a template string); and the count of the number of unmatched open curly braces we have seen (only applicable if the Lexer is currently in a template expression).

TODO: Add support for template literal newlines in the JS REPL (this will cause a syntax error currently):

```
> `foo
> bar`
'foo
bar'
```

### DIFF
--- a/Libraries/LibJS/AST.h
+++ b/Libraries/LibJS/AST.h
@@ -762,6 +762,24 @@ private:
     Vector<RefPtr<Expression>> m_elements;
 };
 
+class TemplateLiteral final : public Expression {
+public:
+    TemplateLiteral(NonnullRefPtrVector<Expression> expressions)
+        : m_expressions(expressions)
+    {
+    }
+
+    virtual Value execute(Interpreter&) const override;
+    virtual void dump(int indent) const override;
+
+    const NonnullRefPtrVector<Expression>& expressions() const { return m_expressions; }
+
+private:
+    virtual const char* class_name() const override { return "TemplateLiteral"; }
+
+    const NonnullRefPtrVector<Expression> m_expressions;
+};
+
 class MemberExpression final : public Expression {
 public:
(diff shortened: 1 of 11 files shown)


Libraries/LibJS/AST.h
```diff
@@ -762,6 +762,24 @@ class ArrayExpression : public Expression {
...
+public:
+    TemplateLiteral(NonnullRefPtrVector<Expression> expressions)
+        : m_expressions(expressions)
+    {
+    }
+
+    virtual Value execute(Interpreter&) const override;
+    virtual void dump(int indent) const override;
+
+    const NonnullRefPtrVector<Expression> expressions() const { return m_expressions; }
```
awesomekling, line 775: Let's not return a copy of the vector. :)
```suggestion
    const NonnullRefPtrVector<Expression>& expressions() const { return m_expressions; }
```

PR #2095, opened 2020-05-04
AK: rename variables with camelCase names in PrintfImplementation.h

emanuele6:
zeroPad => zero_pad
leftPad => left_pad
fieldWidth => field_width

These were the only variables with names in camelCase.
We were not consistent with the naming of these variables: some times we
called them zeroPad, leftPad, fieldWidth; other times we called them
zero_pad, left_pad, field_width.

These inconsistencies made the code hard to read, so I changed their
names to snake_case.

Also rename width => field_width in AK::print_hex()

### DIFF
--- a/AK/PrintfImplementation.h
+++ b/AK/PrintfImplementation.h
@@ -42,7 +42,7 @@ extern "C" size_t strlen(const char*);
 #endif
 
 template<typename PutChFunc, typename T>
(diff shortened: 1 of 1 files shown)


emanuele6:
oops, I forgot to `add` the changes before amending the commit.

PR #2097, opened 2020-05-04
HackStudio: Inspect local variables in the debugger

itamar8910:
This adds a Debug Information tab which displays local variables in a tree view when execution is paused.

A screenshot:
hs-debugger-vars
 
This feature is by no means complete :smile: 
For example, there is currently no support for nested scopes, and we only recognize `int` and `char` primitives.

I also experienced some infrequent cpu hangups that are probably caused by excessive copying in some parts of the DWARF parsing. I figure we can deal with that later, though.

### DIFF
--- a/DevTools/HackStudio/main.cpp
+++ b/DevTools/HackStudio/main.cpp
@@ -25,6 +25,7 @@
  */
 
 #include "CursorTool.h"
+#include "DebugInfoWidget.h"
 #include "Debugger.h"
 #include "Editor.h"
 #include "EditorWrapper.h"
@@ -508,6 +509,7 @@ int main(int argc, char** argv)
 
     auto& find_in_files_widget = s_action_tab_widget->add_tab<FindInFilesWidget>("Find in files");
     auto& terminal_wrapper = s_action_tab_widget->add_tab<TerminalWrapper>("Build", false);
+    auto& debug_info_widget = s_action_tab_widget->add_tab<DebugInfoWidget>("Debug");
 
     auto& locator = widget.add<Locator>();
 
@@ -602,12 +604,17 @@ int main(int argc, char** argv)
 
     RefPtr<EditorWrapper> current_editor_in_execution;
     Debugger::initialize(
-        [&](DebugInfo::SourcePosition source_position) {
+        [&](const PtraceRegisters& regs) {
             dbg() << "Program stopped";
-            current_editor_in_execution = get_editor_of_file(source_position.file_path);
-            current_editor_in_execution->editor().set_execution_position(source_position.line_number - 1);
+
+            auto source_position = Debugger::the().session()->debug_info().get_source_position(regs.eip);
+            ASSERT(source_position.has_value());
+            current_editor_in_execution = get_editor_of_file(source_position.value().file_path);
+            current_editor_in_execution->editor().set_execution_position(source_position.value().line_number - 1);
+            debug_info_widget.update_variables(regs);
             continue_action->set_enabled(true);
             single_step_action->set_enabled(true);
+            reveal_action_tab(debug_info_widget);
         },
         [&]() {
             dbg() << "Program continued";
--- a/Libraries/LibC/ulimit.cpp
+++ b/Libraries/LibC/ulimit.cpp
@@ -25,8 +25,8 @@
  */
 
 #include <assert.h>
-#include <ulimit.h>
 #include <sys/resource.h>
+#include <ulimit.h>
 
 extern "C" {
 
@@ -38,12 +38,10 @@ long ulimit(int cmd, long newlimit)
     return -1;
 }
 
-int getrusage(int who, struct rusage *usage)
+int getrusage(int who, struct rusage* usage)
 {
     (void)who;
     (void)usage;
-    ASSERT_NOT_REACHED();
     return -1;
 }
-
 }
--- a/Libraries/LibDebug/DebugInfo.cpp
+++ b/Libraries/LibDebug/DebugInfo.cpp
@@ -26,16 +26,66 @@
 
 #include "DebugInfo.h"
 #include <AK/QuickSort.h>
+#include <LibDebug/Dwarf/CompilationUnit.h>
+#include <LibDebug/Dwarf/DwarfInfo.h>
+#include <LibDebug/Dwarf/Expression.h>
 
 DebugInfo::DebugInfo(NonnullRefPtr<const ELF::Loader> elf)
     : m_elf(elf)
+    , m_dwarf_info(Dwarf::DwarfInfo::create(m_elf))
 {
+    prepare_variable_scopes();
     prepare_lines();
+}
+
+void DebugInfo::prepare_variable_scopes()
+{
+    m_dwarf_info->for_each_compilation_unit([&](const Dwarf::CompilationUnit& unit) {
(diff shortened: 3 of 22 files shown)


Libraries/LibC/ulimit.cpp
```diff
@@ -38,12 +38,10 @@ long ulimit(int cmd, long newlimit)
     return -1;
 }
 
-int getrusage(int who, struct rusage *usage)
+int getrusage(int who, struct rusage* usage)
 {
     (void)who;
     (void)usage;
-    ASSERT_NOT_REACHED();
```
awesomekling, line 45: Let's at least be nice and log something with `dbg()`

Libraries/LibDebug/DebugInfo.cpp
```diff
@@ -88,3 +137,94 @@ Optional<u32> DebugInfo::get_instruction_from_source(const String& file, size_t
...
+    auto type_die_offset = variable_die.get_attribute(Dwarf::Attribute::Type);
+    ASSERT(type_die_offset.has_value());
+    ASSERT(type_die_offset.value().type == Dwarf::DIE::AttributeValue::Type::DieReference);
+
+    auto type_die = variable_die.get_die_at_offset(type_die_offset.value().data.as_u32);
+    auto type_name = type_die.get_attribute(Dwarf::Attribute::Name);
+    if (type_name.has_value()) {
+        variable_info->type = type_name.value().data.as_string;
+    } else {
+        dbg() << "Unnamed DWRAF type at offset: " << type_die.offset();
```
awesomekling, line 189: DWRAF

Libraries/LibDebug/DebugInfo.h
```diff
@@ -45,6 +49,33 @@ class DebugInfo {
...
+            u32 address;
+        } location_data { 0 };
+
+        NonnullOwnPtrVector<VariableInfo> members;
+        VariableInfo* parent { nullptr };
+    };
+
+    struct VariablesScope {
+        bool is_function { false };
+        Optional<String> name;
```
awesomekling, line 71: `String` has a null state which is what you get with `String()`, we normally use that instead of wrapping it in `Optional`.

Libraries/LibDebug/Dwarf/DIE.h
```diff
@@ -0,0 +1,94 @@
...
+#include <AK/Function.h>
+#include <AK/NonnullOwnPtr.h>
+#include <AK/Optional.h>
+#include <AK/Types.h>
+
+namespace Dwarf {
+
+class CompilationUnit;
+
+// DIE = Dwarf Information Entry
```
awesomekling, line 40: I think technically this stands for Debugging Information Entry, but this works too :)

Libraries/LibDebug/Dwarf/DwarfInfo.cpp
```diff
@@ -0,0 +1,69 @@
...
+    m_debug_strings_data = section_data(".debug_str");
+
+    populate_compilation_units();
+}
+
+ByteBuffer DwarfInfo::section_data(const String& section_name)
+{
+    auto section = m_elf->image().lookup_section(section_name);
+    ASSERT(!section.is_undefined());
+    return ByteBuffer::wrap(reinterpret_cast<const u8*>(section.raw_data()), section.size());
```
awesomekling, line 45: It could be nice to have an API in `ELF::Image::Section` that gives you a `ByteBuffer` wrapping the whole section like you're doing here :)

Libraries/LibDebug/Dwarf/DwarfInfo.h
```diff
@@ -0,0 +1,72 @@
...
+#include <AK/NonnullRefPtr.h>
+#include <AK/RefCounted.h>
+#include <AK/String.h>
+#include <Libraries/LibELF/Loader.h>
+
+namespace Dwarf {
+
+class DwarfInfo : public RefCounted<DwarfInfo> {
+public:
+    static NonnullRefPtr<DwarfInfo> create(NonnullRefPtr<const ELF::Loader> elf) { return adopt(*new DwarfInfo(elf)); }
```
awesomekling, line 41: You can `move()` in situations like this to generate slightly tighter code:
```suggestion
    static NonnullRefPtr<DwarfInfo> create(NonnullRefPtr<const ELF::Loader> elf) { return adopt(*new DwarfInfo(move(elf))); }
```

DevTools/HackStudio/main.cpp
```diff
@@ -602,12 +604,17 @@ int main(int argc, char** argv)
...
     Debugger::initialize(
-        [&](DebugInfo::SourcePosition source_position) {
+        [&](const PtraceRegisters& regs) {
             dbg() << "Program stopped";
-            current_editor_in_execution = get_editor_of_file(source_position.file_path);
-            current_editor_in_execution->editor().set_execution_position(source_position.line_number - 1);
+
+            auto source_position = Debugger::the().session()->debug_info().get_source_position(regs.eip);
+            ASSERT(source_position.has_value());
+            current_editor_in_execution = get_editor_of_file(source_position.value().file_path);
```
awesomekling, line 612: Thought: We should probably be nicer if we can't find the file :)

Libraries/LibDebug/Dwarf/DwarfInfo.h
bugaevc: What you actually need is `&&` on the argument type:

```cpp
static NonnullRefPtr<DwarfInfo> create(NonnullRefPtr<const ELF::Loader>&& elf) { return adopt(*new DwarfInfo(elf)); }
```

which communicates to the caller that they're losing ownership of the loader if they pass it into the function. You can then use or not use `move()` in the body.

awesomekling: Yes, but in this case the caller doesn't want to give it up, so using `NonnullRefPtr&&` would force the caller into the awkward position of having to synthesize a temporary copy.

bugaevc: If it doesn't want to give it up, then how is it ok to take it form under them by casting the reference to `&&` (which is what `move()` does)?

If it *does* want to give it up, it should use `move()` itself.

awesomekling: The argument is a `NonnullRefPtr`, a copy made by the caller. `move()` here only yanks the argument value, not whatever the caller had on his side. :)

bugaevc: Ah, it's a `NonnullRefPtr`, I was thinking it's an `NonnullOwnPtr` for some reason :facepalm:

Libraries/LibDebug/Dwarf/DIE.h
itamar8910: :laughing:

itamar8910:
@awesomekling, I haven't pushed the changes you requested yet :sweat_smile:  
Does marking a conversation as resolved mean "I agree with the requested changes", or "I pushed the requested changes to the branch"?

awesomekling:
Oh whoops! I always interpret it as "I've pushed these changes to the branch"
So since they were trivial changes I didn't even look at them. That's my mistake :)

itamar8910:
I created #2118 2118

PR #2098, opened 2020-05-04
LibJS: Console Improvements

emanuele6:


Libraries/LibJS/Interpreter.h
```diff
@@ -168,6 +168,7 @@ class Interpreter {
     const Console& console() const { return m_console; }
 
     String join_arguments();
+    Vector<String> get_trace();
```
awesomekling: Can these be const qualified?

emanuele6: `join_arguments()` can be, but `get_trace()` can't because `call_stack()` is not const, should I add and overload?

emanuele6: sure

PR #2099, opened 2020-05-04
LibJS: Implement rest parameters

linusg:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/rest_parameters

### DIFF
--- a/Libraries/LibJS/Runtime/ScriptFunction.cpp
+++ b/Libraries/LibJS/Runtime/ScriptFunction.cpp
@@ -27,6 +27,7 @@
 #include <AK/Function.h>
 #include <LibJS/AST.h>
 #include <LibJS/Interpreter.h>
+#include <LibJS/Runtime/Array.h>
 #include <LibJS/Runtime/Error.h>
 #include <LibJS/Runtime/GlobalObject.h>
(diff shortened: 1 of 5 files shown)


Libraries/LibJS/Runtime/ScriptFunction.cpp
```diff
@@ -99,10 +100,17 @@ Value ScriptFunction::call(Interpreter& interpreter)
...
+        if (parameter.is_rest) {
+            auto* array = Array::create(interpreter.global_object());
+            for (size_t rest_index = i; rest_index < argument_values.size(); ++rest_index)
+                array->elements().append(argument_values[rest_index]);
+            value = Value(array);
+        } else {
+            if (i < argument_values.size() && !argument_values[i].is_undefined()) {
+                value = argument_values[i];
+            } else if (parameter.default_value) {
+                value = parameter.default_value->execute(interpreter);
```
awesomekling, line 112: I think we're missing an exception check here

linusg: We are - but in my defence I only moved this code :smile:

Fixed!

PR #2100, opened 2020-05-04
Userland: Adds unzip command and a file to test

zlotny:
This is a first implementation on unzip that can unzip zip files without compression (with "store" type of compression).

In the future we could work on that but here's the layout on how unzip could work.


bugaevc:
(standard complaint about commit messages having to be in imperative mood)

Userland/unzip.cpp
```diff
@@ -0,0 +1,132 @@
...
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include <LibCore/ArgsParser.h>
+#include <LibCore/File.h>
+#include <string.h>
+#include <sys/stat.h>
+
+static u8 central_directory_file_header_sig[] = "\x50\x4b\x01\x02";
```
bugaevc, line 32: `const`

Userland/unzip.cpp
```diff
@@ -0,0 +1,132 @@
...
+int main(int argc, char** argv)
+{
+    Vector<const char*> paths;
+
+    Core::ArgsParser args_parser;
+    args_parser.add_positional_argument(paths, "File to unzip", "path", Core::ArgsParser::Required::Yes);
+    args_parser.parse(argc, argv);
+
+    if (paths.size() > 1) {
+        printf("Too many arguments\n");
```
bugaevc, line 101: That doesn't make sense. Either you support multiple arguments (then don't abort when multiple are present), or you don't (then don't accept multiple arguments in the first place).

In the latter case, if you want `path` to be a single string rather than an array of strings, declare it as `const char* path`. `ArgsParser` will do the rest for you.

Userland/unzip.cpp
```diff
@@ -0,0 +1,132 @@
...
+        return 1;
+    }
+
+    String zip_file_path = { paths[0] };
+
+    auto file = Core::File::construct(zip_file_path);
+    if (!file->open(Core::IODevice::ReadOnly)) {
+        printf("Cant read file %s. %s\n", zip_file_path.characters(), strerror(errno));
```
bugaevc, line 109: `Can't` (or `Cannot`), and you should put a colon instead of a period there (that's what `perror` does).

Userland/unzip.cpp
```diff
@@ -0,0 +1,132 @@
...
+        return 2;
+    }
+
+    struct stat stat;
+    int rc = lstat(zip_file_path.characters(), &stat);
```
bugaevc, line 114: Why `lstat` rather than `stat`?

Userland/unzip.cpp
```diff
@@ -0,0 +1,132 @@
...
+    if (rc < 0) {
+        perror("lstat");
+        memset(&stat, 0, sizeof(stat));
+    }
+
+    printf("Archive: %s\n", zip_file_path.characters());
```
bugaevc, line 120: This (and the rest of logging) should go to `stderr`, shouldn't it?

Userland/unzip.cpp
zlotny: That's right. Thanks :)

Userland/unzip.cpp
zlotny: Didn't properly look to overloads. Sorry about that and thank you for the info :+1:

Userland/unzip.cpp
zlotny: Typo there. Added colon too. Thanks!

Userland/unzip.cpp
zlotny: Copy&Paste things. I started from an lstat and didn't notice it.

Userland/unzip.cpp
```diff
@@ -0,0 +1,127 @@
+/*
+ * Copyright (c) 2018-2020, Andrés Vieira <anvieiravazquez@gmail.com>
```
linusg, line 2:
```suggestion
 * Copyright (c) 2020, Andrés Vieira <anvieiravazquez@gmail.com>
```

bugaevc:
Then there's also a question of trusting the zip file to actually contain as many bytes as it claims. Each of those `read()` and `seek()` calls can fail (or — return less data than requested); and in that case you're using uninitialized memory, which is UB.

(Actually `IODevice::read(u8*, int)` itself contains UB in case of shorter reads, oh well.)

Userland/unzip.cpp
zlotny: Now only the error text goes on stderr. I checked on my linux unzip command and the rest goes on stdout. So mimicking that

Userland/unzip.cpp
zlotny: Right, damn copy paste.

zlotny:
@bugaevc Several things can fail right now on how the zip data is parsed, really, but the idea here was to have a first implementation to at least have something to work with.

Moreover as discussed on IRC in the future this kind of functionality could be on a server to interface with on IPC, so there's room for changes and improvements.

I wanted to get an MVP so we could iterate this in tree, improve data parsing and add other functionality.

Base/usr/share/man/man1/unzip.md
```diff
@@ -0,0 +1,25 @@
+## Name
+
+**unzip** - extract files in a ZIP archive
```
bugaevc, line 3: Shouldn't be bold.

And perhaps *from* a ZIP archive?

zlotny: You're right, I was looking at the wrong man when I took the format to write it. Also yeah, better with *from*. Thanks

Userland/unzip.cpp
```diff
@@ -0,0 +1,127 @@
...
+ */
+
+#include <LibCore/ArgsParser.h>
+#include <LibCore/File.h>
+#include <string.h>
+#include <sys/stat.h>
+
+static const u8 central_directory_file_header_sig[] = "\x50\x4b\x01\x02";
+
+bool find_next_central_directory(off_t file_size, Core::File* file, off_t current_index, off_t* return_index)
```
awesomekling, line 34: Since passing a null `file` is not valid, let's make it a `Core::File&`.
Same with `return_index`.

Userland/unzip.cpp
```diff
@@ -0,0 +1,127 @@
...
+{
+    u8 buffer[4];
+    off_t start_index = current_index == 0 ? current_index : current_index + 1;
+    for (off_t index = start_index; index < file_size - 4; index++) {
+        file->seek(index);
+        file->read(buffer, 4);
+        if (buffer[0] == central_directory_file_header_sig[0] && !memcmp(buffer, central_directory_file_header_sig, 4)) {
```
awesomekling, line 41: Why do we need to compare the first byte before comparing the first four bytes?

Userland/unzip.cpp
```diff
@@ -0,0 +1,127 @@
...
+
+static const u8 central_directory_file_header_sig[] = "\x50\x4b\x01\x02";
+
+bool find_next_central_directory(off_t file_size, Core::File* file, off_t current_index, off_t* return_index)
+{
+    u8 buffer[4];
+    off_t start_index = current_index == 0 ? current_index : current_index + 1;
+    for (off_t index = start_index; index < file_size - 4; index++) {
+        file->seek(index);
+        file->read(buffer, 4);
```
awesomekling, line 40: Missing error handling.

Userland/unzip.cpp
```diff
@@ -0,0 +1,127 @@
...
+        }
+    }
+    return false;
+}
+
+bool unpack_file_for_central_directory_index(off_t central_directory_index, Core::File* file)
+{
+    u8 buffer[4];
+
+    off_t local_file_header_index;
```
awesomekling, line 53: Please declare variables as late as possible instead. No need to declare this before assigning to it.

Userland/unzip.cpp
```diff
@@ -0,0 +1,127 @@
...
+    file->seek(central_directory_index + 42);
+    file->read(buffer, 4);
```
awesomekling, line 55: Missing error handling (same comment repeated many times.)
I bet we could make some kind of helper that does e.g "seek to P and read N bytes" :)

Userland/unzip.cpp
```diff
@@ -0,0 +1,127 @@
...
+    file->read(buffer, 2);
+    extra_field_length = buffer[1] << 8 | buffer[0];
+
+    char file_name[file_name_length + 1];
+    file->seek(local_file_header_index + 30);
+    file->read((u8*)file_name, file_name_length);
+    file_name[file_name_length] = '\0';
+
+    auto new_file = Core::File::construct(String { file_name });
+    new_file->open(Core::IODevice::WriteOnly);
```
awesomekling, line 79: Missing error check here.

Userland/unzip.cpp
```diff
@@ -0,0 +1,127 @@
...
+    String zip_file_path = { path };
+
+    auto file = Core::File::construct(zip_file_path);
+    if (!file->open(Core::IODevice::ReadOnly)) {
+        fprintf(stderr, "Can't read file %s: %s\n", zip_file_path.characters(), strerror(errno));
+        return 2;
+    }
+
+    struct stat st;
+    int rc = stat(zip_file_path.characters(), &st);
```
awesomekling, line 109: Since you've already opened the file, you can use `fstat()`:
```suggestion
    int rc = fstat(file->fd(), &st);
```

Userland/unzip.cpp
```diff
@@ -0,0 +1,127 @@
...
+    if (rc < 0) {
+        perror("st");
+        memset(&st, 0, sizeof(st));
```
awesomekling, line 112: Let's just exit the program if this fails instead of trying to recover.

Userland/unzip.cpp
```diff
@@ -0,0 +1,127 @@
...
+
+    Core::ArgsParser args_parser;
+    args_parser.add_positional_argument(path, "File to unzip", "path", Core::ArgsParser::Required::Yes);
+    args_parser.parse(argc, argv);
+
+    String zip_file_path = { path };
+
+    auto file = Core::File::construct(zip_file_path);
+    if (!file->open(Core::IODevice::ReadOnly)) {
+        fprintf(stderr, "Can't read file %s: %s\n", zip_file_path.characters(), strerror(errno));
```
awesomekling, line 104: `errno` may have been clobbered at this point, you can fetch the error string from `file->error_string()`.

Userland/unzip.cpp
```diff
@@ -0,0 +1,127 @@
...
+    new_file->open(Core::IODevice::WriteOnly);
+
+    printf(" extracting: %s\n", file_name);
+    u8 file_contents[compressed_file_size];
+    file->seek(local_file_header_index + 30 + file_name_length + extra_field_length);
+    file->read(file_contents, compressed_file_size);
+
+    // FIXME: Try to uncompress data here. We're just ignoring it as no decompression methods are implemented yet.
+    new_file->write(file_contents, compressed_file_size);
+    new_file->close();
```
awesomekling, line 88: Missing error checks.

Userland/unzip.cpp
```diff
@@ -0,0 +1,149 @@
...
+bool seek_and_read(u8* buffer, Core::File& file, off_t seek_to, int number_of_reads)
+{
+    if (!buffer)
+        return false;
+
+    if (!file.seek(seek_to))
+        return false;
+
+    if (file.read(buffer, number_of_reads) != number_of_reads)
+        return false;
```
bugaevc, line 43: You should instead loop while the total number of bytes read is less than we need (checking for errors).

zlotny: I don't understand that. Isn't Core::IODevice::read(...) supposed to return the number of bytes that it read? So if it reads something that is not exactly the size we need it's wrong data.

I may be wrong, also I don't get how the while loop could look like, sorry :(

bugaevc: > So if it reads something that is not exactly the size we need it's wrong data.

Not necessarily wrong data, but maybe it's not all of the data we want. A while loop could look like this:

```cpp
size_t need_bytes = ...;
size_t read_bytes = 0;

while (read_bytes < need_bytes) {
    auto nread = file.read(buffer + read_bytes, need_bytes - read_bytes);
    if (nread <= 0)
        return false;
    read_bytes += nread;
}
```

zlotny: I see. I assumed read() always read the number of bytes provided. Fixed it. Thank you so much :)

awesomekling:
All right, let's merge this. We can improve upon it in-tree :)

Issue #1676, opened 2020-04-06
SystemMonitor: processes of other users cannot be seen correctly


lnzero1dev:
As of the rights management, details of processes not belonging to the currently logged in user are not visible.
A solution would be a menu entry to restart the process manager in priviledged mode to gain more rights to view all the process details.
It might be also good, to hide non user related processes in the non priviledged mode to not confuse the user seeing e.g. empty file descriptor tables for those processes.

kleinesfilmroellchen:
All processes of all users currently show up.

PR #2106, opened 2020-05-04
SystemMonitor: Replace details with message when process isn't accesible

Dexesttp:
In addition to being less surprising, this sets up some groundwork for #1676

1604_Step1_Screen1
1604_Step1_Screen2
1604_Step1_Screen3

### DIFF
--- a/Applications/SystemMonitor/main.cpp
+++ b/Applications/SystemMonitor/main.cpp
@@ -75,6 +75,40 @@ static NonnullRefPtr<GUI::Widget> build_pci_devices_tab();
 static NonnullRefPtr<GUI::Widget> build_devices_tab();
 static NonnullRefPtr<GUI::Widget> build_graphs_tab();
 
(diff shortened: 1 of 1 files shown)


Applications/SystemMonitor/main.cpp
```diff
@@ -75,6 +76,43 @@ static NonnullRefPtr<GUI::Widget> build_pci_devices_tab();
...
+        painter.draw_text(frame_inner_rect(), text(), Gfx::TextAlignment::Center, palette().window_text(), Gfx::TextElision::Right);
+    }
+
+    String m_text;
+};
+
+static bool
+can_access_pid(pid_t pid)
+{
+    // FIXME : there's probably a cleaner way to check if we have access to the directory.
```
awesomekling, line 109:
```cpp
return access(pathname, X_OK) == 0;
```

Dexesttp: Oh, that's way cleaner indeed !

PR #2107, opened 2020-05-04
LibWeb/LibHTTP: Form submission improvements for google.com

shadowfacts:
### Improve <form> submit method
Defaults to GET if a method isn't provided or an invalid one is provided.
See: https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#attr-fs-method

### Actually include query parameters when serializing raw request
What it says on the tin. This is why before visiting `https://www.google.com/search?q=serenityos` would just redirect back to the google homepage, @linusg.

### When creating form action URL, only include value for the submit that was used to submit the form
Per the spec: https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#constructing-form-data-set, only the button that was used to submit the form has its value included.

### Add basic URL encoder for individual values and param lists
Otherwise clicking the google search button would produce a parameter "btnK=Google Search" containing a literal space which would result in a malformed raw HTTP request. This follows the spec: https://url.spec.whatwg.org/#urlencoded-serializing, but I'm sure there are more edge cases. Right now, it's in LibWeb since I'm not sure if it should be there or in AK.

With all of this, it's possible to submit a google search and receive results back, though Browser crashes when rendering them because the CSS is too complex.

### DIFF
--- a/Libraries/LibHTTP/HttpRequest.cpp
+++ b/Libraries/LibHTTP/HttpRequest.cpp
@@ -65,6 +65,10 @@ ByteBuffer HttpRequest::to_raw_request() const
     builder.append(method_name());
     builder.append(' ');
     builder.append(m_url.path());
+    if (!m_url.query().is_empty()) {
+        builder.append('?');
+        builder.append(m_url.query());
+    }
     builder.append(" HTTP/1.1\r\nHost: ");
     builder.append(m_url.host());
     builder.append("\r\nConnection: close\r\n\r\n");
--- a/Libraries/LibWeb/DOM/HTMLFormElement.cpp
+++ b/Libraries/LibWeb/DOM/HTMLFormElement.cpp
@@ -41,7 +41,7 @@ HTMLFormElement::~HTMLFormElement()
 {
 }
 
-void HTMLFormElement::submit()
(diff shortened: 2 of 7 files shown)


linusg:
> Although it redirects back to the home page for some reason...

That is because for something like `https://www.google.com/search?q=SerenityOS`	in the Serenity Browser Google returns a 302 status code and `Location: https://www.google.com/webhp` header (which we follow).

So that's the reason why it redirects, but I'm not sure why it would redirect us haha... maybe missing some request headers to convince it to serve us the search results?

alimpfard:
> maybe missing some request headers to convince it to serve us the search results?

Nope, we're just not sending the query at all, we're just sending the url path

shadowfacts:
@alimpfard Yep, I added a commit to this PR that fixed it: https://github.com/SerenityOS/serenity/pull/2107/commits/46ec879ee29e1cbf8828e4e6904357fa8c107f6c

alimpfard:
Alright, cool!

linusg:
That's the last thing I'd have expected (and didn't even check). lol
Good job tracking this down!

awesomekling:
Strong PR :)

PR #2109, opened 2020-05-04
LibJS: Limit scope of 'for' loop variables

### LINKED ISSUE #2103: LibJS: For-loop leaks let and const variables
```js
> for (var i = 40; false;);
undefined
> i
40 // expected
```

```js
> for (let j = 41; false;);
undefined
> j
41 // not expected, should throw `ReferenceError`
```

```js
> for (const k = 42; false;);
undefined
> k
42 // not expected, should throw `ReferenceError`
```

Jongy:
This required 2 changes:
1. In the parser, create a new variable scope, so the variable is declared in it
   instead of the scope in which the 'for' is found.
2. On execute, push the variable into the newly created block. Existing
   code created an empty block (no variables, no arguments) which allows
   `Interpreter::enter_scope()` to skip the creation of a new environment,
   therefore when the variable initializer is executed, it sets the variable
   to the outer scope. By attaching the variable to the new block, it receives
   a new environment.

This is only needed for 'let' / 'const' declarations, since 'var'
declarations are expected to leak.

I'm also a C guy so I'm fairly new to C++. Perhaps that `static_cast` trick is not legit, I'm not sure.

Also I'm not sure it's the right way to solve it: perhaps it's better to have the variable declared in the same scope, and make sure to delete it afterwards? I went with current solution because `ForStatement::execute` already had this logic of creating a new block for `let`/ `const`.

Fixes #2103 if we merge it :)

~~P.S `js-tests/run-tests` is broken when running on Serenity, so I didn't run them all. My test works. Tomorrow I'll try running all others, to make sure I didn't break anything.~~
~~I also hadn't run `clang-format` on it yet.~~
All tests (168 + 1 new) JS tests pass. Also ran `clang-format`.

### DIFF
--- a/Libraries/LibJS/AST.cpp
+++ b/Libraries/LibJS/AST.cpp
@@ -236,6 +236,9 @@ Value ForStatement::execute(Interpreter& interpreter) const
 
     if (m_init && m_init->is_variable_declaration() && static_cast<const VariableDeclaration*>(m_init.ptr())->declaration_kind() != DeclarationKind::Var) {
         wrapper = create_ast_node<BlockStatement>();
+        NonnullRefPtrVector<VariableDeclaration> decls;
+        decls.append(*static_cast<const VariableDeclaration*>(m_init.ptr()));
+        wrapper->add_variables(decls);
         interpreter.enter_scope(*wrapper, {}, ScopeType::Block);
     }
 
--- a/Libraries/LibJS/Parser.cpp
+++ b/Libraries/LibJS/Parser.cpp
@@ -1066,6 +1066,7 @@ NonnullRefPtr<ForStatement> Parser::parse_for_statement()
     consume(TokenType::ParenOpen);
 
     bool first_semicolon_consumed = false;
+    bool in_scope = false;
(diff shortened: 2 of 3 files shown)


linusg:
> P.S js-tests/run-tests is broken when running on Serenity, so I didn't run them all.

It's certainly *not* broken, I just tested on latest `master` - all 168 tests pass. Did you install the bash port?

Jongy:
> Did you install the bash port?

Argh. I certainly did not, lol. I'm also pretty sure I saw the same problem in #2046, guess I just forgot it. I'll try them all tomorrow then, thanks for the tip!

Jongy:
I compiled and installed bash, hadn't noticed the `Ports` directory earlier. Anyway all tests pass + clang-format is alright (no diff)

Jongy:
@awesomekling is this okay, or missing anything?

awesomekling:
Oh whoops, I missed that this was good to go! Sorry about that. It looks great. Nice and surgical :)

Jongy:
Great :)

PR #2110, opened 2020-05-05
LibGfx: Quadratic Bezier curve; LibWeb: canvas.quadraticCurveTo

alimpfard:
I ended up implementing quadratic bezier curve drawing, and figured I'd add it to LibWeb's canvas and serve it as a separate PR :)


awesomekling:
Very tasteful :)

PR #2113, opened 2020-05-05
LibTLS: Try to disambiguate errors in case of failure

alimpfard:
> can we get a bit more informative TLS handshake failure errors maybe?

Maybe :^)


awesomekling:
Thank you!

PR #2114, opened 2020-05-05
Lists and unveil pt 2

bugaevc:
This was meant to be in https://github.com/SerenityOS/serenity/pull/2101, but apparently a ~~rebase~~ push has gone wrong.


awesomekling:
:)

PR #2117, opened 2020-05-05
LibWeb: Recognise :focus pseudo-class

linusg:
It's still only a dummy as LibWeb doesn't have focused elements yet, but at least now we don't treat "selector:focus" as just "selector".

This fixes an issue on google.com which was mostly grey - coming from some menu item focus styles :^)

```css
.gbzt:hover,
.gbzt:focus,
.gbgt-hvr,
.gbgt:focus {
    background-color:#4c4c4c; /* <-- */
    background-image:none;
    _background-image:none;
    background-position:0 -102px;
    background-repeat:repeat-x;
    outline:none;
    text-decoration:none !important
}
```

(yes, the menu items are still all over the place)

image

### DIFF
--- a/Libraries/LibWeb/CSS/Selector.h
+++ b/Libraries/LibWeb/CSS/Selector.h
@@ -48,6 +48,7 @@ public:
             None,
             Link,
             Hover,
+            Focus,
             FirstChild,
             LastChild,
             OnlyChild,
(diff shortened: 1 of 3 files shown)


linusg:
image

awesomekling:
Great fix :)

Issue #1259, opened 2020-02-21
Help: Window title updated incorrectly


xTibor:
When opening a manual page the Help app sets the window title based on the currently selected tree view node, not the actual manual page path passed to the `open_page` function. This causes incorrect window titles when following hyperlinks in the document or navigating through the history.

shannonbooth:
I think it would make sense if following a hyperlink in About would update the tree view selection to that item.

shannonbooth:
Also, following some of the hyperlinks in About seems to lead to a blank page for me? e.g the mount(8) link in mount(2) - and vice versa

xTibor:
The blank pages are caused by the `../` unveil bug.

shannonbooth:
Made #1260 as I forgot about that bug. Kinda makes me want to do some Kernel work, but me reckons I already do enough low level kernel/bootloader work at my day job. IAlways nice to write some higher level code every so often.

PR #2119, opened 2020-05-05
Help: Clicking on link changes the treeview selection

Dexesttp:
When clicking a link in the help application, the treeview selection now changes to select the page on the left sidebar (when possible).

If the requested file isn't in the treeview, then the previous behaviour takes place (the page attempts to be loaded, but the treeview sidebar isn't updated).

This also solves issue #1259 since the page title is computed from the treeview selection.

## Limitation

When the target page's category is collapsed, it doesn't get auto-uncollapsed on selection and the selected item stays hidden.

There is a piece of code surrounded by `#if 0 (...) #endif` in TreeView.cpp l. 344-356 which enables the expected behaviour, but I assume it was disabled for a reason.

### DIFF
--- a/Applications/Help/ManualModel.cpp
+++ b/Applications/Help/ManualModel.cpp
@@ -47,6 +47,24 @@ ManualModel::ManualModel()
     m_page_icon.set_bitmap_for_size(16, Gfx::Bitmap::load_from_file("/res/icons/16x16/filetype-unknown.png"));
 }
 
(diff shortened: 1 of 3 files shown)


awesomekling:
I think the disabled code in `TreeView` was about `FileManager` opening literally every possible subtree at launch or something like that. Feel free to make it better so that problem goes away :)

awesomekling:
Great fix btw!

Issue #1878, opened 2020-04-19
Taskbar: Window blocked by modal window becomes incorrectly "active" in taskbar when clicked


awesomekling:
To reproduce:

1. Open TextEditor.
2. Open the app's about box.
3. Click on the TextEditor main window's entry in the taskbar.

Both the modal about box and the TextEditor main window now appear "active" in the taskbar with bold text.

Jongy:
Hi, I was looking for a "good first issue" and this seemed like one :) SerenityOS is a very cool project and I wanna try contributing.

Anyway, I found `TaskbarWindow::create_button` which appears to be where the taskbar buttons are created. If I change `button.set_checkable(true)` to  `set_checkable(false)` then the problem described here disappears. It's because `checkable` is processed in `Button:click` which just swaps the checkable status without respect to whether the click actually led to a window focus change. With this setting disabled - the checkable status is set only in `TaskbarWindow::wm_event` according to the active state.

But I'm sure sure if the behavior with modal windows is not the root cause itself: the About dialog completely blocks the application (that's maybe okay) but trying to focus on the application window won't do anything (I'd expect it to focus on the About window instead). Though, I just downloaded Serenity a hour ago, what do I know :D Will appreciate advice which direction to take.

Jongy:
>  If I change button.set_checkable(true) to set_checkable(false) then the problem described here disappears. If the button is checkable, the check status is swapped in Button:click without respect to whether the click actually led to a window focus change. With this checkable disabled - the checked status is set only in TaskbarWindow::wm_event according to the active state.

I thought about it more, and it makes sense to remove this `checkable` thing. So I'll submit a PR soon.

PR #2122, opened 2020-05-05
Taskbar: Remove checkable property from taskbar buttons

Jongy:
If a button is checkable, its "checked" state is inversed in
Button::click. It's not needed for taskbar buttons - their checked
state is updated in TaskbarWindow::wm_event, based on the
current state of their window.

Fixes: #1878.


awesomekling:
Great surgical fix!

PR #2125, opened 2020-05-06
Browser: Add "Copy link" context menu item

modmuss50:
Hello all! This is my first PR thought something simple would be a good way to get started.

This PR adds a "Copy Link" context menu item that will copy the right clicked href url to the clipboard.


awesomekling:
Hello! Solid first patch :)

PR #2128, opened 2020-05-06
LibJS: Add function call spreading

mattco98:
Adds support for the following syntax:

```js
Math.max(...x, ...[1, 2, 3], ...o.foo, ...'abcd')
```

### DIFF
--- a/Libraries/LibJS/AST.cpp
+++ b/Libraries/LibJS/AST.cpp
@@ -137,12 +137,28 @@ Value CallExpression::execute(Interpreter& interpreter) const
     arguments.values().append(function.bound_arguments());
 
     for (size_t i = 0; i < m_arguments.size(); ++i) {
-        auto value = m_arguments[i].execute(interpreter);
+        auto value = m_arguments[i].value->execute(interpreter);
         if (interpreter.exception())
             return {};
-        arguments.append(value);
-        if (interpreter.exception())
-            return {};
+        if (m_arguments[i].is_spread) {
+            // FIXME: Support generic iterables
+            Vector<Value> iterables;
+            if (value.is_string()) {
+                for (auto ch : value.as_string().string())
+                    iterables.append(Value(js_string(interpreter, String::format("%c", ch))));
+            } else if (value.is_object() && value.as_object().is_array()) {
+                iterables = static_cast<const Array&>(value.as_object()).elements();
+            } else if (value.is_object() && value.as_object().is_string_object()) {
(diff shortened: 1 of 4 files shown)


Libraries/LibJS/AST.cpp
```diff
@@ -138,12 +138,28 @@ Value CallExpression::execute(Interpreter& interpreter) const
...
+            if (value.is_string()) {
+                for (auto ch : value.as_string().string())
+                    iterables.append(Value(ch));
+            } else if (value.is_object() && value.as_object().is_array()) {
+                iterables = static_cast<const Array&>(value.as_object()).elements();
+            } else if (value.is_object() && value.as_object().is_string_object()) {
+                for (auto ch : static_cast<const StringObject&>(value.as_object()).primitive_string().string())
+                    iterables.append(Value(ch));
+            } else {
+                ASSERT_NOT_REACHED();
```
linusg, line 156: this should be a `TypeError` instead - `foo(...1)` should _not_ crash.

Libraries/LibJS/AST.cpp
```diff
@@ -138,12 +138,28 @@ Value CallExpression::execute(Interpreter& interpreter) const
...
-        arguments.append(value);
+        auto value = m_arguments[i].value->execute(interpreter);
         if (interpreter.exception())
             return {};
+        if (m_arguments[i].is_spread) {
+            // FIXME: Support generic iterables
+            Vector<Value> iterables;
+            if (value.is_string()) {
+                for (auto ch : value.as_string().string())
+                    iterables.append(Value(ch));
```
awesomekling, line 149: Wait, does this make sense? This means that spreading a string will iterate over the individual characters as integers, not as single-character strings.

Libraries/LibJS/AST.cpp
```diff
@@ -138,12 +138,28 @@ Value CallExpression::execute(Interpreter& interpreter) const
...
+            } else if (value.is_object() && value.as_object().is_array()) {
+                iterables = static_cast<const Array&>(value.as_object()).elements();
+            } else if (value.is_object() && value.as_object().is_string_object()) {
+                for (auto ch : static_cast<const StringObject&>(value.as_object()).primitive_string().string())
+                    iterables.append(Value(ch));
```
awesomekling, line 154: Same question here.

Libraries/LibJS/AST.cpp
mattco98: Nope, it doesn't... that's what I get for assuming `Math.max(...'abc')` would return the number, instead of the thing it actually returns -- `NaN`!

PR #2129, opened 2020-05-06
LibGfx: Filling paths; LibWeb: canvas.fill

alimpfard:
Next in the line of rasterising paths.

This makes the generated curves much more pretty, and adds filling paths.
There are some issues with for_each_line_segment_on_bezier_curve creating non-simple shapes, because of fitting points onto an int grid, but they are mostly okay...ish.

### DIFF
--- a/Libraries/LibGfx/Painter.cpp
+++ b/Libraries/LibGfx/Painter.cpp
@@ -31,6 +31,7 @@
 #include <AK/Assertions.h>
 #include <AK/Function.h>
 #include <AK/Memory.h>
+#include <AK/QuickSort.h>
 #include <AK/StdLibExtras.h>
 #include <AK/StringBuilder.h>
 #include <AK/Utf8View.h>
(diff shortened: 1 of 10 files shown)


Libraries/LibGfx/Painter.cpp
```diff
@@ -1137,4 +1149,157 @@ void Painter::stroke_path(const Path& path, Color color, int thickness)
...
+            return;
+        }
+
+        ASSERT_NOT_REACHED();
+    };
+
+    while (scanline >= last_y) {
+        if (active_list.size()) {
+            // sort the active list by 'x' from right to left
+            qsort(active_list.data(), active_list.size(), sizeof(Path::LineSegment), [](const void* d0, const void* d1) -> int {
```
awesomekling, line 1208: Can we use `AK::quick_sort` instead here?

alimpfard: Huh. I didn't know we _has_ an `AK::quick_sort`. we absolutely can.
(there's also one in Path.cpp)

alimpfard:
> This makes me think we should add an AffineTransform to Gfx::Painter and have all paint ops respect it. Instead of doing that only in CanvasRenderingContext2D.

Not sure what brought this up, but it _would_ simplify quite a lot of stuff.
I say that having to mirror the y coordinates of every glyph in a TTF font

PR #2130, opened 2020-05-06
LibJS: Implement tagged template literals (foo`bar`)

linusg:
To make processing tagged template literals easier, template literals will now add one empty `StringLiteral` before and after each template expression *if* there's no other string - e.g.:

```
`${foo}` -> "", foo, ""
`test${foo}${bar}test` -> "test", foo, "", bar, "test"
```

This also matches the behaviour of many other parsers.

### DIFF
--- a/Libraries/LibJS/AST.cpp
+++ b/Libraries/LibJS/AST.cpp
@@ -114,8 +114,7 @@ Value CallExpression::execute(Interpreter& interpreter) const
 
     ASSERT(!callee.is_empty());
 
-    if (!callee.is_object()
-        || !callee.as_object().is_function()
+    if (!callee.is_function()
         || (is_new_expression() && (callee.as_object().is_native_function() && !static_cast<NativeFunction&>(callee.as_object()).has_constructor()))) {
         String error_message;
         auto call_type = is_new_expression() ? "constructor" : "function";
@@ -132,7 +131,7 @@ Value CallExpression::execute(Interpreter& interpreter) const
         return interpreter.throw_exception<TypeError>(error_message);
     }
 
-    auto& function = static_cast<Function&>(callee.as_object());
+    auto& function = callee.as_function();
 
     MarkedValueList arguments(interpreter.heap());
     arguments.values().append(function.bound_arguments());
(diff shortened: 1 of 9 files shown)


Libraries/LibJS/AST.cpp
```diff
@@ -1297,6 +1296,46 @@ Value TemplateLiteral::execute(Interpreter& interpreter) const
...
+    m_template_literal->dump(indent + 2);
+}
+
+Value TaggedTemplateLiteral::execute(Interpreter& interpreter) const
+{
+    auto tag = m_tag->execute(interpreter);
+    if (interpreter.exception())
+        return {};
+    // FIXME: We really should have a helper for this.
+    if (!tag.is_object() || !tag.as_object().is_function()) {
```
awesomekling, line 1316: How about we just add `Value::is_function()`? :)

Libraries/LibJS/AST.cpp
```diff
@@ -1297,6 +1296,46 @@ Value TemplateLiteral::execute(Interpreter& interpreter) const
...
+        interpreter.throw_exception<TypeError>(String::format("%s is not a function", tag.to_string().characters()));
+        return {};
+    }
+    auto& tag_function = static_cast<Function&>(tag.as_object());
```
awesomekling, line 1320: And `Value::as_function()`!

Libraries/LibJS/AST.h
```diff
@@ -786,6 +786,24 @@ class TemplateLiteral final : public Expression {
     const NonnullRefPtrVector<Expression> m_expressions;
 };
 
+class TaggedTemplateLiteral final : public Expression {
+public:
+    TaggedTemplateLiteral(NonnullRefPtr<Expression> tag, NonnullRefPtr<TemplateLiteral> template_literal)
+        : m_tag(tag)
+        , m_template_literal(template_literal)
```
awesomekling, line 793:
```suggestion
        : m_tag(move(tag))
        , m_template_literal(move(template_literal))
```

Libraries/LibJS/AST.cpp
linusg: Yes! It was about time :)

Libraries/LibJS/Runtime/Object.cpp
```diff
@@ -477,10 +477,8 @@ Value Object::to_primitive(PreferredType preferred_type) const
 Value Object::to_string() const
 {
     auto to_string_property = get("toString");
-    if (!to_string_property.is_empty()
-        && to_string_property.is_object()
-        && to_string_property.as_object().is_function()) {
-        auto& to_string_function = static_cast<Function&>(to_string_property.as_object());
+    if (!to_string_property.is_empty() && to_string_property.is_function()) {
```
awesomekling, line 480: Even tighter:
```suggestion
    if (to_string_property.is_function()) {
```

PR #2132, opened 2020-05-06
Build: Switch to CMake :^)

bugaevc:
This is my work on converting Serenity to be built with CMake, loosely based on @awesomekling's prior work.

This is somewhat WIP, as I've broken some things that work with the old buildsystem.

Everybody, please try to fix building on systems you're interested in :smile: (tagging a few people: @jarhill0 @Pagghiu @jcs @brynet @zaklaus @BobbyRaduloff)

Oh, and the ports are most likely broken too. We're going to want to figure out how we would want them to work in the brave new out-of-tree builds world.


predmond:
You didn’t mean to include your Build/ dir did you?

bugaevc:
I didn't :laughing: fixed

JonArcherII:
`Kernel/makeall.sh` certainly isn't going to work anymore :wink: perhaps it's time to get rid of it?

bugaevc:
Yes, it was meant to be removed; I'll push that later with other fixes.

predmond:
When I tried using the ninja generator it failed to assemble .S files with an error like:
```
/home/predmond/src/bugaevc/Toolchain/Local/bin/i686-pc-serenity-as  -I../Libraries -I../. -I../Libraries/LibC -I../Servers -IServers -ILibraries  -MD -MT Libraries/LibC/CMakeFiles/LibC.dir/setjmp.S.o -MF Libraries/LibC/CMakeFiles/LibC.dir/setjmp.S.o.d -o Libraries/LibC/CMakeFiles/LibC.dir/setjmp.S.o -c ../Libraries/LibC/setjmp.S
/home/predmond/src/bugaevc/Toolchain/Local/bin/i686-pc-serenity-as: invalid option -- 'F'
```
when I switched to using the toolchain file `cmake -DCMAKE_TOOLCHAIN_FILE=../Toolchain/CMakeToolchain.txt ..` then it correctly built the .S file
```
/home/predmond/src/bugaevc/Toolchain/Local/bin/i686-pc-serenity-gcc --sysroot=/home/predmond/src/bugaevc//Root  -I../Libraries -I../. -I../Libraries/LibC -I../Servers -IServers -ILibraries  -MD -MT Libraries/LibC/CMakeFiles/LibC.dir/setjmp.S.obj -MF Libraries/LibC/CMakeFiles/LibC.dir/setjmp.S.obj.d -o Libraries/LibC/CMakeFiles/LibC.dir/setjmp.S.obj -c ../Libraries/LibC/setjmp.S
```

In the later case it uses gcc instead of as.

I believe this is because how cmake detects the assembler when you specify
```
project(SerenityOS C CXX ASM)
```
When the toolchain file is used is correctly sees the i686-pc-serenity-gcc is the assembler.

However, because you overwrite ASM_COMPILER with `i686-pc-serenity-as` it breaks because it initially detected gcc as the assembler (before the cross-compiler is set) and therefore it's passing invalid gcc args to as.

bugaevc:
@predmond changing the ASM compiler to gcc works, but then building with Ninja stumbles on this:
```
/home/sergey/dev/serenity/Toolchain/Local/lib/gcc/i686-pc-serenity/9.3.0/../../../../i686-pc-serenity/bin/ld: cannot find CMakeFiles/Kernel.dir/Arch/i386/Boot/boot.S.o
collect2: error: ld returned 1 exit status
```

The file is there (`Kernel/CMakeFiles/Kernel.dir/Arch/i386/Boot/boot.S.o`), but apparently the Ninja backend of CMake doesn't generate `cd`s into the individual subdirectories (which makes sense), so looking up the file as `CMakeFiles/Kernel.dir/Arch/i386/Boot/boot.S.o` fails. Any idea how to fix this nicely so it works both with Make and Ninja?

bgianfo:
@bugaevc  I played around with this a little bit, I was able to get consistent behavior across Ninja / Make generators with this:

```diff 

diff --git a/Kernel/CMakeLists.txt b/Kernel/CMakeLists.txt
index c601ef108..693ba6f8c 100644
--- a/Kernel/CMakeLists.txt
+++ b/Kernel/CMakeLists.txt
@@ -155,7 +155,10 @@ set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -mno-80387 -mno-mmx -mno-sse -mno-sse2")
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fno-asynchronous-unwind-tables")
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -nostdlib -nostdinc -nostdinc++")

-add_link_options(LINKER:-T ${CMAKE_CURRENT_SOURCE_DIR}/linker.ld -nostdlib)
+# Replace variables in linker.ld and generate output linker.ld
+configure_file(linker.ld linker.ld @ONLY)
+
+add_link_options(LINKER:-T ${CMAKE_CURRENT_BINARY_DIR}/linker.ld -nostdlib)

 if (${CMAKE_HOST_SYSTEM_NAME} MATCHES SerenityOS)
     include_directories(/usr/local/include/c++/9.3.0/)
diff --git a/Kernel/linker.ld b/Kernel/linker.ld
index b02ad6f4d..0906a9199 100644
--- a/Kernel/linker.ld
+++ b/Kernel/linker.ld
@@ -8,7 +8,7 @@ SECTIONS

     .text ALIGN(4K) : AT (ADDR(.text) - 0xc0000000)
     {
-        CMakeFiles/Kernel.dir/Arch/i386/Boot/boot.S.o
+        @CMAKE_CURRENT_BINARY_DIR@/CMakeFiles/Kernel.dir/Arch/i386/Boot/boot.S.o
         *(.multiboot)
         start_of_kernel_text = .;
         *(.text)

```

bgianfo:
Dunno if you want to take PR's to your fork for this PR or not.
Either way here's a small diff to get Base copied on make/ninja install:

```diff
diff --git a/Base/CMakeLists.txt b/Base/CMakeLists.txt
new file mode 100644
index 000000000..1f9f02de9
--- /dev/null
+++ b/Base/CMakeLists.txt
@@ -0,0 +1,5 @@
+INSTALL(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/etc DESTINATION .)
+INSTALL(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/home DESTINATION .)
+INSTALL(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/res DESTINATION .)
+INSTALL(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/usr DESTINATION .)
+INSTALL(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/www DESTINATION .)
diff --git a/CMakeLists.txt b/CMakeLists.txt
index e6694db44..53f81053d 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -122,3 +122,4 @@ add_subdirectory(MenuApplets)
 add_subdirectory(Shell)
 add_subdirectory(Demos)
 add_subdirectory(Userland)
+add_subdirectory(Base)
```

bgianfo:
Another diff to make the "run" target depend on the "image" target.

```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index e6694db44..2d9fc9851 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -12,6 +12,7 @@ add_custom_target(image
 add_custom_target(run
     COMMAND ${CMAKE_SOURCE_DIR}/Meta/run.sh
     USES_TERMINAL
+    DEPENDS ${CMAKE_BINARY_DIR}/_disk_image
 )

 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra -Werror -std=c++2a")
```

It's unfortunate that CMake doesn't have a way to make a target depend on "install".
So we'll always have to tell people to "make install" then "make run".
See: https://gitlab.kitware.com/cmake/cmake/-/issues/8438

bugaevc:
> I played around with this a little bit, I was able to get consistent behavior across Ninja / Make generators with this

Yep, that was my plan B if there's no way to get Make/Ninja to use the same directory.

> here's a small diff to get Base copied on make/ninja install

But why? This step is already handled by `build-root-filesystem.sh`, no point in duplicating all those files?

If, on the other hand, we were to get rid of `build-root-fileaystem.sh` completely and persuade CMake to build the complete root filesystem under `Build/Root/`... But that'd need to create device files and whatnot, so I decided we keep `build-root-filesystem.sh` separate for now.

> Another diff to make the "run" target depend on the "image" target.

Hmm, probably a good idea, thanks. Does it actually work (with both Make and Ninja) to add dependencies between custom targets like this?

bgianfo:
> > here's a small diff to get Base copied on make/ninja install
> 
> But why? This step is already handled by `build-root-filesystem.sh`, no point in duplicating all those files?
> 
> If, on the other hand, we were to get rid of `build-root-fileaystem.sh` completely and persuade CMake to build the complete root filesystem under `Build/Root/`... But that'd need to create device files and whatnot, so I decided we keep `build-root-filesystem.sh` separate for now.
> 
You want the file copy step to be in CMake so users get correct incremental build behaviour.
In general build logic belongs in the build, scripts should be used as utilities for the build system to utilize.

> > Another diff to make the "run" target depend on the "image" target.
> 
> Hmm, probably a good idea, thanks. Does it actually work (with both Make and Ninja) to add dependencies between custom targets like this?

Yes, and I tested it to double check. 
See: https://cmake.org/cmake/help/latest/command/add_custom_command.html

bugaevc:
> In general build logic belongs in the build, scripts should be used as utilities for the build system to utilize.

I understand that much 😉

> You want the file copy step to be in CMake so users get correct incremental build behaviour.

No, making an image always makes a fresh image, mounts it and copies files there. This process is always not incremental. I don't know if it would be possible to make it incremental.

> See: https://cmake.org/cmake/help/latest/command/add_custom_command.html

But both `run` and `image` are "custom targets", not "custom commands" :slightly_smiling_face:

bgianfo:
> > You want the file copy step to be in CMake so users get correct incremental build behaviour.
> 
> No, making an image always makes a fresh image, mounts it and copies files there. This process is always not incremental. I don't know if it would be possible to make it incremental.

I think it's a bit nuanced issue. My point of view is that, sure the image building step isn't currently incremental. However just because it's not incremental, should we should make design decisions that create other obstacles to it being make incremental in the future? In my mind no, we shouldn't. However I can understand a different point of view that leans more to the, if someone wants to fix it then fix all of it.

Straw-man idea:  we could add a wrapper around add_executable, etc that added all items we build to a list of dependencies, the image build target could then depend on that image list. If we could correctly mark dependencies on all of the inputs to the image (including the scripts used to create it). There would be nothing stopping us from making image build incremental. Which I would personally see as a very valuable feature.

In the end I'm not trying to push you to take the diff, it's not really worth going back and forth over any more? :man_shrugging: 

> > See: https://cmake.org/cmake/help/latest/command/add_custom_command.html
> 
> But both `run` and `image` are "custom targets", not "custom commands" slightly_smiling_face

I just posted the wrong link ... :disappointed:  https://cmake.org/cmake/help/latest/command/add_custom_target.html

bugaevc:
@bgianfo your `boot.S` change fails for me with:
```
/home/sergey/dev/serenity/Toolchain/Local/lib/gcc/i686-pc-serenity/9.3.0/../../../../i686-pc-serenity/bin/ld: CMakeFiles/Kernel.dir/Arch/i386/Boot/boot.S.o: in function `boot_pdpt':
(.page_tables+0x0): multiple definition of `boot_pdpt'; /home/sergey/dev/serenity/Build/Kernel/CMakeFiles/Kernel.dir/Arch/i386/Boot/boot.S.o:(.page_tables+0x0): first defined here
/home/sergey/dev/serenity/Toolchain/Local/lib/gcc/i686-pc-serenity/9.3.0/../../../../i686-pc-serenity/bin/ld: CMakeFiles/Kernel.dir/Arch/i386/Boot/boot.S.o: in function `boot_pd0':
(.page_tables+0x1000): multiple definition of `boot_pd0'; /home/sergey/dev/serenity/Build/Kernel/CMakeFiles/Kernel.dir/Arch/i386/Boot/boot.S.o:(.page_tables+0x1000): first defined here
/home/sergey/dev/serenity/Toolchain/Local/lib/gcc/i686-pc-serenity/9.3.0/../../../../i686-pc-serenity/bin/ld: CMakeFiles/Kernel.dir/Arch/i386/Boot/boot.S.o: in function `boot_pd3':
(.page_tables+0x2000): multiple definition of `boot_pd3'; /home/sergey/dev/serenity/Build/Kernel/CMakeFiles/Kernel.dir/Arch/i386/Boot/boot.S.o:(.page_tables+0x2000): first defined here
/home/sergey/dev/serenity/Toolchain/Local/lib/gcc/i686-pc-serenity/9.3.0/../../../../i686-pc-serenity/bin/ld: CMakeFiles/Kernel.dir/Arch/i386/Boot/boot.S.o: in function `boot_pd3_pde1023_pt':
(.page_tables+0x3000): multiple definition of `boot_pd3_pde1023_pt'; /home/sergey/dev/serenity/Build/Kernel/CMakeFiles/Kernel.dir/Arch/i386/Boot/boot.S.o:(.page_tables+0x3000): first defined here
/home/sergey/dev/serenity/Toolchain/Local/lib/gcc/i686-pc-serenity/9.3.0/../../../../i686-pc-serenity/bin/ld: CMakeFiles/Kernel.dir/Arch/i386/Boot/boot.S.o: in function `start':
(.text+0x0): multiple definition of `start'; /home/sergey/dev/serenity/Build/Kernel/CMakeFiles/Kernel.dir/Arch/i386/Boot/boot.S.o:(.text+0x0): first defined here
collect2: error: ld returned 1 exit status
```

bugaevc:
> But both `run` and `image` are "custom targets", not "custom commands" :slightly_smiling_face:

Tested it, and indeed that causes image generation to be run *every time* when you do `make run`, not just the first time. So I'm not pulling that patch either, sorry :(

awesomekling:
For ports, we need to pass `--sysroot` to everything, not just in `CFLAGS` but also in `CPPFLAGS`, `CXXFLAGS` and possibly `LDFLAGS`. This feels very brittle, and I wish we could just have the toolchain built with the correct sysroot like we did before. Why doesn't that work again? :/

bugaevc:
> Why doesn't that work again? :/

Because there's no one predefined build location anymore. If, on the other hand, we were to only support building in `Build`....

awesomekling:
> > Why doesn't that work again? :/
> 
> Because there's no one predefined build location anymore. If, on the other hand, we were to only support building in `Build`....

This seems like a reasonable compromise to me :)

predmond:
> @bgianfo your `boot.S` change fails for me with:
> 
> ```
> /home/sergey/dev/serenity/Toolchain/Local/lib/gcc/i686-pc-serenity/9.3.0/../../../../i686-pc-serenity/bin/ld: CMakeFiles/Kernel.dir/Arch/i386/Boot/boot.S.o: in function `boot_pdpt':
> (.page_tables+0x0): multiple definition of `boot_pdpt'; /home/sergey/dev/serenity/Build/Kernel/CMakeFiles/Kernel.dir/Arch/i386/Boot/boot.S.o:(.page_tables+0x0): first defined here
> /home/sergey/dev/serenity/Toolchain/Local/lib/gcc/i686-pc-serenity/9.3.0/../../../../i686-pc-serenity/bin/ld: CMakeFiles/Kernel.dir/Arch/i386/Boot/boot.S.o: in function `boot_pd0':
> (.page_tables+0x1000): multiple definition of `boot_pd0'; /home/sergey/dev/serenity/Build/Kernel/CMakeFiles/Kernel.dir/Arch/i386/Boot/boot.S.o:(.page_tables+0x1000): first defined here
> /home/sergey/dev/serenity/Toolchain/Local/lib/gcc/i686-pc-serenity/9.3.0/../../../../i686-pc-serenity/bin/ld: CMakeFiles/Kernel.dir/Arch/i386/Boot/boot.S.o: in function `boot_pd3':
> (.page_tables+0x2000): multiple definition of `boot_pd3'; /home/sergey/dev/serenity/Build/Kernel/CMakeFiles/Kernel.dir/Arch/i386/Boot/boot.S.o:(.page_tables+0x2000): first defined here
> /home/sergey/dev/serenity/Toolchain/Local/lib/gcc/i686-pc-serenity/9.3.0/../../../../i686-pc-serenity/bin/ld: CMakeFiles/Kernel.dir/Arch/i386/Boot/boot.S.o: in function `boot_pd3_pde1023_pt':
> (.page_tables+0x3000): multiple definition of `boot_pd3_pde1023_pt'; /home/sergey/dev/serenity/Build/Kernel/CMakeFiles/Kernel.dir/Arch/i386/Boot/boot.S.o:(.page_tables+0x3000): first defined here
> /home/sergey/dev/serenity/Toolchain/Local/lib/gcc/i686-pc-serenity/9.3.0/../../../../i686-pc-serenity/bin/ld: CMakeFiles/Kernel.dir/Arch/i386/Boot/boot.S.o: in function `start':
> (.text+0x0): multiple definition of `start'; /home/sergey/dev/serenity/Build/Kernel/CMakeFiles/Kernel.dir/Arch/i386/Boot/boot.S.o:(.text+0x0): first defined here
> collect2: error: ld returned 1 exit status
> ```

I got the same error when I tried this. Interestingly, if I replace the path in the configured file to relative (Kernel/CMakeFiles/Kernel.dir/Arch/i386/Boot/boot.S.o) then it works for ninja (but not for make)

predmond:
I was able to get the Kernel to compile by tweaking @bgianfo diff

```
diff --git a/Kernel/CMakeLists.txt b/Kernel/CMakeLists.txt
index c601ef108..e27143218 100644
--- a/Kernel/CMakeLists.txt
+++ b/Kernel/CMakeLists.txt
@@ -4,7 +4,6 @@ set(KERNEL_SOURCES
     ACPI/Initialize.cpp
     ACPI/MultiProcessorParser.cpp
     ACPI/Parser.cpp
-    Arch/i386/Boot/boot.S
     Arch/i386/CPU.cpp
     CMOS.cpp
     CommandLine.cpp
@@ -155,7 +154,10 @@ set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -mno-80387 -mno-mmx -mno-sse -mno-sse2")
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fno-asynchronous-unwind-tables")
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -nostdlib -nostdinc -nostdinc++")

-add_link_options(LINKER:-T ${CMAKE_CURRENT_SOURCE_DIR}/linker.ld -nostdlib)
+add_link_options(LINKER:-T ${CMAKE_CURRENT_BINARY_DIR}/linker.ld -nostdlib)
+
+add_library(boot OBJECT Arch/i386/Boot/boot.S)
+file(GENERATE OUTPUT linker.ld INPUT linker.ld)

 if (${CMAKE_HOST_SYSTEM_NAME} MATCHES SerenityOS)
     include_directories(/usr/local/include/c++/9.3.0/)
@@ -167,6 +169,7 @@ endif()

 add_executable(Kernel ${SOURCES})
 target_link_libraries(Kernel gcc stdc++)
+add_dependencies(Kernel boot)
 install(TARGETS Kernel RUNTIME DESTINATION boot)

 add_custom_command(
diff --git a/Kernel/linker.ld b/Kernel/linker.ld
index b02ad6f4d..ce3e125db 100644
--- a/Kernel/linker.ld
+++ b/Kernel/linker.ld
@@ -8,7 +8,7 @@ SECTIONS

     .text ALIGN(4K) : AT (ADDR(.text) - 0xc0000000)
     {
-        CMakeFiles/Kernel.dir/Arch/i386/Boot/boot.S.o
+        $<TARGET_OBJECTS:boot>
         *(.multiboot)
         start_of_kernel_text = .;
         *(.text)
```
Here I build boot.S as an OBJECT library then use `file(GENERATE ...)` which supports generator expressions. This will be more portable when you want to support different architectures.

@bugaevc can you try this change?

bugaevc:
@predmond thanks, that works! (I had to do another fix to get Ninja to work though; now it does, so I mentioned it in the BuildInstructions as well.)

@awesomekling the button should be green now :wink:

alimpfard:
yay fast builds
finally!
(and a pretty progress indicator)

CMakeLists.txt
```diff
@@ -0,0 +1,124 @@
+cmake_minimum_required (VERSION 3.17)
```
predmond, line 1: It works for me with 3.16.

awesomekling:
> @awesomekling the button should be green now

:green_apple: :green_book: :green_circle: :green_heart: :green_salad: :green_square:

awesomekling:
Thank you for the great work @bugaevc :)

jonnii:
One issue is that this requires globstar and/or bash 4+ which doesn't come by default on the mac. Sadness.

Jongy:
I'm probably missing something, but - I'm getting errors like `installing base system... chown: cannot access 'mnt/bin/su': No such file or directory` and I don't see anything in `build-root-filesystem.sh` that actually copies stuff from `Userland` to `bin/`. I had to add
```
cp -R Userland/* mnt/bin
cp Kernel/Kernel mnt/boot/Kernel
```
to get it running successfully.

(based on c38c2668da3fff187c7eeae3043322535aa60770)

P.S I'm building on Docker so `mnt` is not mounted. Perhaps this creates the variation.

I also might've copied the wrong kernel because Serenity bootloops afterwards :p

bugaevc:
It sounds like you were missing `make install`

Jongy:
> It sounds like you were missing make install

Uhh you're absolutely right @bugaevc. Thanks for the tip.
I'd expect the `install` target to be a dependency of `image`, wouldn't it make sense? I see you've discussed it here but not sure what was the resolution.

bugaevc:
It would make some sense, but I decided against it, because you might want to run `make image` without reinstalling the whole thing — such as after building a port, or adding a file to Base, or something. And CMake always considers custom targets out of date, so it would always get executed no matter if you've executed it previously.

PR #2135, opened 2020-05-06
LibTLS: Verify server certificate expiry date

alimpfard:
Let's be a _little_ bit more conscious about certificates by checking their expiry date.


Libraries/LibTLS/TLSv12.cpp
```diff
@@ -519,6 +597,28 @@ void TLSv12::ensure_hmac(size_t digest_size, bool local)
         m_hmac_remote = move(hmac);
 }
 
+bool Certificate::is_valid() const
+{
+    auto now = Core::DateTime::now();
+    auto now_string = String::format("%04d%02d%02d%02d%02d%02dZ", now.year(), now.month(), now.weekday(), now.hour(), now.minute(), now.second());
+
+    if (!not_before.is_empty()) {
+        if (strcasecmp(not_before.characters(), now_string.characters()) > 0) {
```
awesomekling, line 606: This is pretty awkward. Can we add something to `Core::DateTime` so we don't have to compare dates by converting them to strings first? :)

Libraries/LibTLS/ClientHandshake.cpp
```diff
@@ -363,6 +363,29 @@ ssize_t TLSv12::handle_payload(const ByteBuffer& vbuffer)
...
+                if (m_context.certificates.size()) {
+                    auto one_valid { false };
+                    size_t valid_index { 0 };
+                    for (auto& cert : m_context.certificates) {
+                        if (cert.is_valid()) {
+                            one_valid = true;
+                            break;
+                        }
+                        ++valid_index;
+                    }
```
awesomekling, line 375: How about using `Vector::find` here?

```cpp
auto it = m_context.certificates.find([&](auto& cert) { return cert.is_valid(); });
...
```

Maybe it's made more awkward by the iterator interface after that.. What do you think?

Libraries/LibTLS/TLSv12.cpp
alimpfard: an alternative would be something like strptime...
or we could just wrap this string compare in core::datetime, which is...quite meh

Libraries/LibTLS/ClientHandshake.cpp
alimpfard: I'll try.

Libraries/LibTLS/TLSv12.cpp
awesomekling: Hiding the ugliness inside `Core::DateTime` is better IMO because that at least makes this logic look better. Then we can go and fix up the datetime code later :)

Libraries/LibTLS/ClientHandshake.cpp
alimpfard: The iterator interface is just fine (`it.index()` exists!)

Libraries/LibCore/DateTime.h
```diff
@@ -54,6 +54,10 @@ class DateTime {
     static DateTime now();
     static DateTime from_timestamp(time_t);
 
+    // FIXME: This should be replaced with a proper comparison
+    //        operator when we get the equivalent of strptime
+    bool is_before(const String&);
```
awesomekling, line 59: pls const :)

PR #2137, opened 2020-05-06
LibWeb/Browser: Middle-clicking a link opens it in a new tab

FalseHonesty:
This is my first contribution, so I hope I've done everything correctly!

This patch allows links to be opened in a new tab when middle-clicked, which is a small, but handy quality of life feature from most browsers.


awesomekling:
Solid first PR :)

PR #2139, opened 2020-05-06
LibJS+Meta: fix warnings for run-tests and stop ignoring it in lint-shell-scripts.sh

emanuele6:
We can just fix the warnings: the script will still work fine in Serenity.

### DIFF
--- a/Libraries/LibJS/Tests/run-tests
+++ b/Libraries/LibJS/Tests/run-tests
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-if [ "`uname`" = "SerenityOS" ]; then
+if [ "$(uname)" = "SerenityOS" ]; then
     js_program=/bin/js
 else
     [ -z "$js_program" ] && js_program="$SERENITY_ROOT/Meta/Lagom/build/js"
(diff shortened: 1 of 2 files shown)


Libraries/LibJS/Tests/run-tests
```diff
@@ -9,24 +9,22 @@ else
...
 fi
 
-extra_args="$*"
-
-pass_count=0
-fail_count=0
-count=0
+(( pass_count = 0 ))
+(( fail_count = 0 ))
+(( count = 0 ))
```
awesomekling, line 14: Is shellcheck actually complaining about this? It looks really weird

Libraries/LibJS/Tests/run-tests
```diff
@@ -9,24 +9,22 @@ else
...
-    result=`$js_program $extra_args -t $f 2>/dev/null`
+    result="$("$js_program" "$@" -t "$f" 2>/dev/null)"
     if [ "$result" = "PASS" ]; then
-        let pass_count++
+        (( ++pass_count ))
         echo -ne "( \033[32;1mPass\033[0m ) "
     else
         echo -ne "( \033[31;1mFail\033[0m ) "
-        let fail_count++
+	(( ++fail_count ))
```
awesomekling, line 24: Please don't mix tabs and spaces

Libraries/LibJS/Tests/run-tests
emanuele6: Yes, it is.

Libraries/LibJS/Tests/run-tests
emanuele6: I fixed it.

Libraries/LibJS/Tests/run-tests
emanuele6: We could keep this..
```bash
pass_count=0
fail_count=0
count=0
```
..it's not complaining about that, declaring the variable between `((` and `))` declares the variable as a number instead of as a string though. The script would work either way.

Should I revert this change?

awesomekling: But these are not `let`?

awesomekling: Yes please revert it. I am already kinda skeptical of the value of `shellcheck` for us, and using it as a justification to make code less readable seems like a step in the wrong direction :)

emanuele6: Ok, I've done it.

PR #2141, opened 2020-05-07
LibGUI/Browser: Middle clicking a tab closes it

FalseHonesty:
In continuing with small quality of life changes in the browser, this patch makes it so that middle clicking a tab fires a hook on the `TabWidget`. In the Browser this is used to close the clicked tab.

### DIFF
--- a/Applications/Browser/main.cpp
+++ b/Applications/Browser/main.cpp
@@ -111,6 +111,11 @@ int main(int argc, char** argv)
         tab.did_become_active();
     };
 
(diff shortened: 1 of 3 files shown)


awesomekling:
Hmm, okay. At first I thought I didn't like this because I'm gonna do it by accident.
But then I tested it and apparently the browsers I use on my host machine already do this, and I've never done it by accident :D

FalseHonesty:
Yeah, I use it often in Firefox, and tested it in other browsers to make sure it was standard before opening this pull request. I should have mentioned that originally to save you some time!

PR #2142, opened 2020-05-07
Browser: assume http:// when no URL protocol is provided in the location bar

Sobak:
Hello!

After over half a year of watching all of the videos and following the commits here (even though I have no idea about C++) the time had come for the first, humble pull request.

Honestly, I have no idea whether it's correct when it comes to our let's say _business needs_ or to the code. I figured out that it shouldn't break anything (correct me if I'm wrong).

I purposely didn't touch CLI invocation of `Browser` because then, if I'm not mistaken, we'd have to actually check whether a file exists to assume the protocol. And that would require a total `unveil` so...

Please feel free to point any issues in the approach or the idea itself! Thanks

### DIFF
--- a/Applications/Browser/Tab.cpp
+++ b/Applications/Browser/Tab.cpp
@@ -30,6 +30,7 @@
 #include "History.h"
 #include "InspectorWidget.h"
 #include "WindowActions.h"
+#include <AK/StringBuilder.h>
 #include <LibGUI/Action.h>
 #include <LibGUI/Application.h>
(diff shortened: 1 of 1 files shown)


awesomekling:
Seems like a decent enough improvement :)

Sobak:
I know I get overly excited over ten lines of code that anyone could write given such simple abstraction layer but 

> Look mommy, I'm on GitHub!

image

PR #2144, opened 2020-05-07
Browser: Escape percent sign in download progress format string

linusg:
Otherwise the following is printed:

```
printf_internal: Unimplemented format specifier   (fmt: 0% of %s)
```


awesomekling:
D'oh!

PR #2146, opened 2020-05-07
Small changes in spinbox and system menu icon

asliturk:


bugaevc:
Should we may replace the system menu icon with a small buggie? :smile:

PR #2148, opened 2020-05-07
Browser: Add "Open in new window" to link context menu

linusg:


Applications/Browser/main.cpp
```diff
@@ -59,7 +59,7 @@ int main(int argc, char** argv)
     // Connect to the ProtocolServer immediately so we can drop the "unix" pledge.
     Web::ResourceLoader::the();
 
-    if (pledge("stdio shared_buffer accept cpath rpath wpath", nullptr) < 0) {
+    if (pledge("stdio shared_buffer accept cpath rpath wpath proc exec", nullptr) < 0) {
```
awesomekling, line 62: Let's not give the browser these powers. Instead let's just open another window in the same process.
We could add a `Browser::Window` class and move everything into there. That would make it easier to instantiate a new one :)

linusg: I tried moving to a multi-window architecture and everything fell apart :D

Have to practice using LibGUI first :^)

PR #2151, opened 2020-05-07
WindowServer: A bit less surprising

BenWiederhake:
ComboBoxes in Dialogs (and probably other WindowTypes) were broken, because the dropdown part was treated like a regular (= less privileged) window.

The "maximize" button used to be permanently lit under certain semi-regular conditions. Note that this patch introduces a different bug, but at least the bug happens under much rarer conditions (the maximize button must end up in basically the same spot), and the buggy behavior is less visible (the button fails to light up until the mouse is moved).


awesomekling:
Seems like good fixes :)

PR #2152, opened 2020-05-07
Browser: Show line numbers when viewing page source

nicktiberi:
When we view the source code of a web page, line numbers will now be shown in the text editor. Firefox and Chrome both do this, so I figured it might be nice if Browser did it too. 😄


awesomekling:
Great idea!

PR #2154, opened 2020-05-08
LibGfx: Handle filling complex shapes better

alimpfard:
This allows the painter to render filled complex shapes better, by constructing a path graph for (interesting) intersecting lines and omitting lines from the containing segments if they are detected
to take no part in defining the edges of a shape.

This approach would still fail if there are multiple logical shapes that are confined to the collection of lines.
For instance, two polygons intersecting each other in a way that one vertex of polygon A ends up inside polygon B.
we would detect that polygon A's edges are part of the shape (technically correct) even though they are not a part of polygon B at all.

If there _is_ a better way to handle filling complex shapes, I'm interested in knowing about it.
Google yielded nothing of interest, and I found no academic resources on the matter (not locked behind a paywall) that presented a solution.

### DIFF
--- a/Libraries/LibGfx/Painter.cpp
+++ b/Libraries/LibGfx/Painter.cpp
@@ -1154,7 +1154,7 @@ void Painter::stroke_path(const Path& path, Color color, int thickness)
 
 void Painter::fill_path(Path& path, Color color, WindingRule winding_rule)
 {
-    const auto& segments = path.split_lines();
+    const auto& segments = path.split_lines(Path::Simple);
 
     if (segments.size() == 0)
         return;
--- a/Libraries/LibGfx/Path.cpp
+++ b/Libraries/LibGfx/Path.cpp
@@ -25,10 +25,13 @@
  */
 
 #include <AK/Function.h>
+#include <AK/HashFunctions.h>
+#include <AK/HashTable.h>
 #include <AK/QuickSort.h>
 #include <AK/StringBuilder.h>
 #include <LibGfx/Painter.h>
 #include <LibGfx/Path.h>
+#include <math.h>
 
 namespace Gfx {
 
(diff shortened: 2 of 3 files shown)


Libraries/LibGfx/Painter.cpp
```diff
@@ -1300,6 +1300,8 @@ void Painter::fill_path(Path& path, Color color, WindingRule winding_rule)
     for (auto& segment : segments)
         draw_line(segment.from, segment.to, Color::from_hsv(++i / segments.size() * 255, 255, 255), 1);
 #endif
+
+    return;
```
awesomekling, line 1304: This seems useless

Libraries/LibGfx/Path.h
```diff
@@ -80,22 +97,35 @@ class Path {
...
+    bool is_part_of_closed_polygon(const Point& p0, const Point& p1);
+    static unsigned hash_line(const Point& from, const Point& to);
 
     Vector<Segment> m_segments;
 
     Optional<Vector<LineSegment>> m_split_lines {};
+
+    struct PathGraphNode {
+        PathGraphNode(u32 hash, const LineSegment& line)
+            : children({})
```
awesomekling, line 117: No need to explicitly initialize with `{}`

Libraries/LibGfx/Path.h
```diff
@@ -80,22 +97,35 @@ class Path {
...
+            , line(line)
+        {
+        }
+
+        Vector<const PathGraphNode*> children;
+        u32 hash;
+        LineSegment line;
+    };
+
+    Optional<HashMap<u32, OwnPtr<PathGraphNode>>> m_graph_node_map {};
```
awesomekling, line 128: No need to explicitly initialize with `{}`

Libraries/LibGfx/Path.cpp
```diff
@@ -141,4 +142,162 @@ void Path::segmentize_path()
...
+
+    if (!maybe_entry.has_value())
+        return true;
+
+    const auto& entry = maybe_entry.value();
+
+    // check if the entry is part of a loop
+    auto is_part_of_loop = false;
+    HashTable<u32> visited {};
+    Vector<const PathGraphNode*> queue {};
```
awesomekling, line 275: No need to explicitly initialize with `{}`

Libraries/LibGfx/Painter.cpp
alimpfard: lol there was a stroke there, I forgot to remove the return along with it :)

PR #2157, opened 2020-05-08
HackStudio: Add a backtrace to the debugger 

itamar8910:
We now show a backtrace in the Debug tab.
We can also go back to previous frames and inspect their variables.

hs-backtrace-vars


DevTools/HackStudio/Debugger/BacktraceModel.cpp
```diff
@@ -0,0 +1,62 @@
...
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include "BacktraceModel.h"
+#include "Debugger.h"
+
+RefPtr<BacktraceModel> BacktraceModel::create(const PtraceRegisters& regs)
+{
+    return adopt(*new BacktraceModel(create_backtrace(regs)));
+}
```
awesomekling, line 33: Since this function never fails, we don't need the ability of `RefPtr` to contain null.
Instead, let's return `NonnullRefPtr<BacktraceModel>`.

DevTools/HackStudio/Debugger/BacktraceModel.cpp
```diff
@@ -0,0 +1,62 @@
...
+    return {};
+}
+
+Vector<BacktraceModel::FrameInfo> BacktraceModel::create_backtrace(const PtraceRegisters& regs)
+{
+    u32 current_ebp = regs.ebp;
+    u32 current_instruction = regs.eip;
+    Vector<BacktraceModel::FrameInfo> frames;
+    do {
+        const auto& debug_info = Debugger::the().session()->debug_info();
```
awesomekling, line 50: Perhaps it would be nicer to decouple this code from the globally available `DebugSession`.
We could pass the active `DebugSession&` as an argument to this function instead, and then this code doesn't need to make the assumption that `Debugger::the().session()` is non-null.

DevTools/HackStudio/Debugger/BacktraceModel.h
```diff
@@ -47,12 +47,15 @@ class BacktraceModel final : public GUI::Model {
...
 
-private:
     struct FrameInfo {
         String function_name;
-        u32 address_in_frame;
+        u32 instruction_address;
+        u32 frame_base;
     };
 
+    const Vector<FrameInfo> frames() const { return m_frames; }
```
awesomekling, line 56: This will create a copy of the vector every times it's called. Let's use a const reference:
```suggestion
    const Vector<FrameInfo>& frames() const { return m_frames; }
```

PR #2158, opened 2020-05-08
LibGfx + LibWeb: Implement GIF animation

peterdn:
This PR:
- LibGfx: Add methods applicable to animated images to ImageDecoder and ImageDecoderPlugin interfaces.
- LibGfx: Implement animated image methods for GIFImageDecoderPlugin and optimise LZWDecoder a little.
- LibWeb: Add support for animated images to HTMLImageElement.

Upshot: animated GIFs are now displayed in the browser. Includes support for variable delays between frames, and single-loop GIFs.

SerenityGIFs

Things to pay particular attention to:
- Animation methods added to ImageDecoder{Plugin} interfaces: felt like the right place to put them but happy to consider alternative approaches.
- Decoding is done lazily but is quite slow and memory-intensive. I intend to optimise in a follow-up PR but can do it here if requested.

### DIFF
--- a/Libraries/LibGfx/ImageDecoder.h
+++ b/Libraries/LibGfx/ImageDecoder.h
@@ -26,14 +26,19 @@
 
 #pragma once
 
-#include <AK/NonnullRefPtr.h>
 #include <AK/OwnPtr.h>
 #include <AK/RefCounted.h>
+#include <AK/RefPtr.h>
 #include <LibGfx/Size.h>
 
 namespace Gfx {
 
 class Bitmap;
+
+struct ImageFrameDescriptor {
+    RefPtr<Bitmap> image;
+    int duration { 0 };
+};
 
 class ImageDecoderPlugin {
 public:
@@ -46,6 +51,11 @@ public:
     [[nodiscard]] virtual bool set_nonvolatile() = 0;
 
     virtual bool sniff() = 0;
+
+    virtual bool is_animated() = 0;
+    virtual size_t loop_count() = 0;
+    virtual size_t frame_count() = 0;
+    virtual ImageFrameDescriptor frame(size_t i) = 0;
 
 protected:
     ImageDecoderPlugin() {}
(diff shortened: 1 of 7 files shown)


Libraries/LibGfx/ImageDecoder.h
```diff
@@ -26,15 +26,20 @@
...
+#include <AK/RefPtr.h>
 #include <LibGfx/Size.h>
 
 namespace Gfx {
 
 class Bitmap;
 
+struct ImageFrameDescriptor {
+    RefPtr<Bitmap> image;
+    int duration;
```
awesomekling, line 40: Please initialize primitive variables when declaring them.

Libraries/LibWeb/DOM/HTMLImageElement.cpp
```diff
@@ -37,6 +38,9 @@ namespace Web {
 
 HTMLImageElement::HTMLImageElement(Document& document, const FlyString& tag_name)
     : HTMLElement(document, tag_name)
+    , m_current_frame_index(0)
+    , m_loops_completed(0)
```
awesomekling, line 42: Please initialize these at the declaration instead :)

peterdn:
Thanks @awesomekling!

> Feature request after this: interlacing
Could be fun :-)

awesomekling:
Why is `loop_count()` an `int`? Can it be negative?

peterdn:
> Why is `loop_count()` an `int`? Can it be negative?

Hmm, probably not. Think I was leaving my options open in case -1 was needed to mean infinite, but 0 does instead. I can make it a `u32` instead.

awesomekling:
Or `size_t` for symmetry with `frame_count()` :)

linusg:
Commits probably need some squashing now :)

https://github.com/SerenityOS/serenity/blob/master/CONTRIBUTING.md#code-submission-policy

> - Squash your commits when making revisions after a patch review.

awesomekling:
Grand!

PR #2159, opened 2020-05-08
LibJS: Handle multiple args in Array constructor, add Array.{isArray, of}()

linusg:


Libraries/LibJS/Runtime/ArrayConstructor.cpp
```diff
@@ -41,6 +41,9 @@ ArrayConstructor::ArrayConstructor()
 {
     put("prototype", interpreter().global_object().array_prototype(), 0);
     put("length", Value(1), Attribute::Configurable);
+
+    u8 attr = Attribute::Writable | Attribute::Configurable;
```
alimpfard, line 45: Just curious, why is this writable?
and for that matter, why is `length` configurable?

MDN doesn't specify, but v8 says what you're saying here.
I don't get why this alone would be writable though

linusg: I tried to think of a good answer but I couldn't, and the spec doesn't tell either. So probably: because that's what other engines have been doing before us? :/

alimpfard: Fantastic :)

PR #2164, opened 2020-05-08
Applications: FontEditor, relocate form items by fons size

asliturk:
FontEditor


awesomekling:
Oh that's really cool! But we need to fix it so we can see the whole "demo string"
In the screenshot, the large demo text should end in "WIZARD!", not "W..." :)

asliturk:
> Oh that's really cool! But we need to fix it so we can see the whole "demo string"
> In the screenshot, the large demo text should end in "WIZARD!", not "W..." :)

Fixed.

linusg:
`s/fons size/font size` in the last commit's message

awesomekling:
:+1:

bugaevc:
But... why rewrite the UI back in C++?

PR #2165, opened 2020-05-08
LibWeb: Implement "text-transform: {upper,lower}case"

linusg:
Grabbing the low-hanging fruits while they're still available :^)


awesomekling:
Cute :)

PR #2169, opened 2020-05-09
Misc: Add missing copyright headers

linusg:
Please don't merge right away, let's give these people some time to confirm they want to be copyright holders of the files they created, otherwise I'll change it to the generic `Copyright (c) YYYY, the SerenityOS developers.`.

---

- @awesomekling 
  - `AK/Memory.h`
  - `Applications/Browser/WindowActions.cpp`
  - `Applications/Browser/WindowActions.h`
  - `Demos/Mouse/main.cpp`
  - `Userland/disasm.cpp`
- @itamar8910 
  - `Kernel/Ptrace.cpp`
- Written by @predmond but no header with their info already exists
  - `Meta/Lagom/Fuzzers/FuzzJs.cpp`
- ~~@alimpfard~~
  - ~~`Userland/test-crypto.cpp`~~
- @howar6hill 
  - `AK/Tests/TestStringUtils.cpp`

---

I'd also like to propose a script that runs on CI and fails the build if `.{cpp,h}` files in certain directories don't contain a copyright header. (yes, I like automation)

### DIFF
--- a/Kernel/Ptrace.cpp
+++ b/Kernel/Ptrace.cpp
@@ -1,3 +1,29 @@
+/*
+ * Copyright (c) 2020, Itamar S. <itamar8910@gmail.com>
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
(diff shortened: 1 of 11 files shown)


alimpfard:
That error-if-copyright-missing script sounds nice.

I'm also just fine with "The SerenityOS developers." on test-crypto (is it even sensible to hold rights over a test suite?)

linusg:
> is it even sensible to hold rights over a test suite?

maybe :)

most of the files in `Tests/Kernel` don't have a header, but on the other hand the ones in `AK/Tests` do.

Kernel/Ptrace.cpp
```diff
@@ -1,3 +1,29 @@
+/*
```
itamar8910, line 1: Thanks @linusg :)

awesomekling:
A script sounds good :)

awesomekling:
Okay let's merge this. If someone wants their copyright removed in the future, we can fix that then. :)

PR #2171, opened 2020-05-09
Meta: Add script to enforce license headers & run it on Travis

linusg:
Depends on https://github.com/SerenityOS/serenity/pull/2169 being merged first!

### DIFF
--- a/Meta/check-license-headers.sh
+++ b/Meta/check-license-headers.sh
@@ -1,0 +1,30 @@
+#!/bin/bash
+
+script_path=$(cd -P -- "$(dirname -- "$0")" && pwd -P)
+cd "$script_path/.." || exit 1
+
+# We simply check if the file starts with:
+# /*
+#  * Copyright
+PATTERN=$'^/\*\n \* Copyright'
+ERRORS=()
+
+while IFS= read -r f; do
+    if [[ ! $(cat "$f") =~ $PATTERN ]]; then
+        ERRORS+=("$f")
+    fi
+done < <(git ls-files -- \
+'*.cpp' \
+'*.h' \
+':!:Tests' \
+':!:Base' \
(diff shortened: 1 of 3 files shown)


Meta/check-license-headers.sh
```diff
@@ -0,0 +1,33 @@
...
+    fi
+done < <(git ls-files -- \
+'*.cpp' \
+'*.h' \
+':!:Tests' \
+':!:Base' \
+':!:Kernel/FileSystem/ext2_fs.h' \
+':!:Libraries/LibC/getopt.cpp' \
+':!:Libraries/LibCore/puff.h' \
+':!:Libraries/LibELF/exec_elf.h' \
```
linusg, line 27: These are borrowed from somewhere and while they do contain a license header it doesn't match ours.

If someone adds a borrowed file where this is the case as well, it has to be added here manually.

Meta/check-license-headers.sh
```diff
@@ -0,0 +1,33 @@
...
+# We simply check if the file starts with:
+# /*
+#  * Copyright
+PATTERN=$'^/\*\n \* Copyright'
+ERRORS=()
+
+while IFS= read -r f; do
+    if [[ ! -s $f ]]; then
+        continue
+    fi
```
linusg, line 15: Apparently we have empty `.h` files - skip those.

awesomekling: Uhh, we can also remove them :D

Meta/check-license-headers.sh
awesomekling: Yes.. although we should strive to get rid of these and try to avoid adding new ones :)

linusg:
image

:+1:

linusg:
let me rebase this real quick and it should be good to go!

PR #2174, opened 2020-05-09
LibWeb: Implicitly close all subpaths when trying to do canvas.fill

### LINKED ISSUE #2172: LibWeb/LibGfx: Path filling behaves incorrectly
Just filing an issue for the sake of having one - I haven't actually investigated what's causing this. Might or might not be easy to fix.

`canvas-path-quadratic-curve.html`:

image

alimpfard:
Well...that was rather painless...

closes #2172


awesomekling:
Fantastic!

PR #2175, opened 2020-05-09
LibDesktop + open: Tiny fixes

alimpfard:
Can we make OwnPtr -> bool explicit?
This sort of thing shouldn't really happen...


awesomekling:
Having to `static_cast<bool>(ptr)` everywhere would be obscenely unpleasant

alimpfard:
Fair. that would be pretty darn irritating.
I dunno, this sort of mistake is something that the compiler should be able to prevent...yet here we are.

PR #2177, opened 2020-05-09
LibC: Fix get{sock,peer}name to match their kernel-side prototypes

Jongy:
In f4302b58fb0, the kernel-side syscalls (e.g `Process::sys$getsockname`)
were updated to use SC_get{sock,peer}name_params, but the libc functions were not updated.

Found this while working on a dropbear port (which will be ready in following
days, hopefully): this call was failing with `EFAULT` (was trying to read `params` from the "address" encoded in `sockfd`. I guess the NULL page is not mapped in Serenity ^^. Or maybe it did read it and failed with `params->addr`).


awesomekling:
Whoops! Nice catch. Guess we're not using those functions a lot :D

Issue #2176, opened 2020-05-09
LibWeb/LibGfx: Dotted borders with width > 1 look funny


linusg:
The way we currently draw these (stacking lots of 1px dotted lines) won't work in the long run.

![image](https://user-images.githubusercontent.com/19366641/81486601-b7ae1e00-924d-11ea-8f91-4793d88a74ff.png)

![image](https://user-images.githubusercontent.com/19366641/81486730-b8937f80-924e-11ea-8827-f144f1cff3bf.png)

PR #2178, opened 2020-05-10
LibGfx: Fix dotted lines and add a LibGfx demo

linusg:
image

(not related to https://github.com/SerenityOS/serenity/issues/2176 but fixing dotted lines is one step into that direction)


awesomekling:
Oh my god

sunverwerth:
This is art

linusg:
> This is art

Abitur in Bildender Kunst was worth it after all

Issue #807, opened 2019-11-20
ip.h - (improved) IP support needed for dropbear SSH


danboid:
Tonight I tried building openssh, libssh and dropbear. dropbear came the nearest to building but we need better IP/ethernet (header/kernel) support before we can build dropbear and get an ssh client/server. I expect git, apache and nginx will have similar requirements.

https://matt.ucc.asn.au/dropbear/dropbear.html

[dropbear.tar.gz](https://github.com/SerenityOS/serenity/files/3873141/dropbear.tar.gz)

I'm sure @awesomekling is well aware of this, I just thought it might be useful to others to document my findings.

Here are the current build errors:

```
Building dropbear!
+ make -j8
Creating default_options_guard.h
make -C libtomcrypt
make -C libtommath
i686-pc-serenity-gcc -c -I./libtomcrypt/src/headers/ -I. -I.  -Os -W -Wall -Wno-pointer-sign -fno-strict-overflow -fPIE -fstack-protector-strong -D_FORTIFY_SOURCE=2 -mfunction-return=thunk -mindirect-branch=thunk -DDROPBEAR_SERVER -DDROPBEAR_CLIENT  dbutil.c -o dbutil.o
i686-pc-serenity-gcc -c -I./libtomcrypt/src/headers/ -I. -I.  -Os -W -Wall -Wno-pointer-sign -fno-strict-overflow -fPIE -fstack-protector-strong -D_FORTIFY_SOURCE=2 -mfunction-return=thunk -mindirect-branch=thunk -DDROPBEAR_SERVER -DDROPBEAR_CLIENT  buffer.c -o buffer.o
i686-pc-serenity-gcc -c -I./libtomcrypt/src/headers/ -I. -I.  -Os -W -Wall -Wno-pointer-sign -fno-strict-overflow -fPIE -fstack-protector-strong -D_FORTIFY_SOURCE=2 -mfunction-return=thunk -mindirect-branch=thunk -DDROPBEAR_SERVER -DDROPBEAR_CLIENT  dbhelpers.c -o dbhelpers.o
i686-pc-serenity-gcc -c -I./libtomcrypt/src/headers/ -I. -I.  -Os -W -Wall -Wno-pointer-sign -fno-strict-overflow -fPIE -fstack-protector-strong -D_FORTIFY_SOURCE=2 -mfunction-return=thunk -mindirect-branch=thunk -DDROPBEAR_SERVER -DDROPBEAR_CLIENT  dss.c -o dss.o
[...]

PR #2184, opened 2020-05-10
Ports: Add dropbear

Jongy:
This is very basic and doesn't support many features. Instead
of describing what it *doesn't* support, I'll describe what I have tested:

1. Public key authentication (password is not supported)
2. Single command execution
3. PTY-less interactive bash shell (/bin/sh doesn't work)
4. Multi-user (i.e you can ssh as 'anon' as well as root)

I added these commands to my `sync-local.sh`:
```
mkdir -p mnt/root/.ssh
cat ../id_rsa.pub > mnt/root/.ssh/authorized_keys
mkdir -p mnt/etc/dropbear
```

After boot, run `dropbear` (can run with `-E` to see output in stdout) and connect with:
`ssh -o StrictHostKeyChecking=no -p 2222 -i id_rsa -T root@127.0.0.1 ps -ef` (this command is for when running with usermode networking, if you use tap / something else, fix IP:port accordingly).

This:
* disables strict host key checking because each run of Serenity regenerates a new key, if you replace the disk.
* uses `-T` because PTY allocation fails at the moment.

You can also run it with `/bin/bash` (after you compile the `bash` port) to get an interactive shell:
`ssh ... root@127.0.0.1 /bin/bash`. Though, PTY-less.

Serenity lacked many things dropper needed and with most I decided to resort to patching, so if you'll say this is too patchy and we better fix some of them instead, I'll totally accept that :D

Nonetheless I still think it's quite useful now, so I decided to push it as is. Originally I wanted this for `scp` (as an easy mean to copy executable files into Serenity, so I can compile usermode programs and test them without recreating the image). `scp` requires some more work which I believe I'll do next.

### DIFF
--- a/Kernel/UnixTypes.h
+++ b/Kernel/UnixTypes.h
@@ -379,6 +379,7 @@ struct pollfd {
 #define AF_UNSPEC 0
 #define AF_LOCAL 1
 #define AF_INET 2
+#define AF_MAX 3
 #define PF_LOCAL AF_LOCAL
 #define PF_INET AF_INET
 
--- a/Kernel/build-root-filesystem.sh
+++ b/Kernel/build-root-filesystem.sh
@@ -86,6 +86,7 @@ chmod 400 mnt/res/kernel.map
 
 chmod 660 mnt/etc/WindowServer/WindowServer.ini
 chown $window_uid:$window_gid mnt/etc/WindowServer/WindowServer.ini
+echo "/bin/sh" > mnt/etc/shells
 
 echo "done"
 
--- a/Kernel/run
+++ b/Kernel/run
@@ -116,7 +116,7 @@ else
         $SERENITY_COMMON_QEMU_ARGS \
         $SERENITY_KVM_ARG \
         $SERENITY_PACKET_LOGGING_ARG \
-        -netdev user,id=breh,hostfwd=tcp:127.0.0.1:8888-10.0.2.15:8888,hostfwd=tcp:127.0.0.1:8823-10.0.2.15:23,hostfwd=tcp:127.0.0.1:8000-10.0.2.15:8000 \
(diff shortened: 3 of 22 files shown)


awesomekling:
Let's merge this as-is and then we can fix the issues one by one and remove patches as we go :)

awesomekling:
Great work btw!

danboid:
Great work Jongy!

It's a shame pty isn't working yet as thats how I usually use ssh but its still useful if I can execute single commands remotely and I don't expect it will be too long before Serenity gains pty support.

Why is password auth not supported?

Thanks!

Jongy:
> Great work btw!

Thanks :)

> Great work Jongy!

Thanks, glad it's helpful for you.

About the next steps:

> It's a shame pty isn't working yet as thats how I usually use ssh

Remember you *can* get interactive shell, just w/o PTY. It's definitely usable, though less convenient.

I think Serenity does support pty (because I see it does have a `/dev/ptmx`) but dropbear failed to operate it. We'll need to investigate why.

> Why is password auth not supported?

Password auth requires the `crypt(3)` function (as you mentioned in your #807) which is something we better implement correctly, then use - instead of adding a patch here.

danboid:
> Thanks, glad it's helpful for you.

I've not tried it yet. I've not played with Serenity in a couple of months now really. I've got some catching up to do now!


> 
> About the next steps:
> 
> > It's a shame pty isn't working yet as thats how I usually use ssh
> 
> Remember you _can_ get interactive shell, just w/o PTY. It's definitely usable, though less convenient.

Will have to play with it.Serenity's shell still needs some loving last time I looked at it anyway - you couldn't resize it non-destructively being the main failing for me last time I checked.

> 
> I think Serenity does support pty (because I see it does have a `/dev/ptmx`) but dropbear failed to operate it. We'll need to investigate why.

Ah OK!

> 
> > Why is password auth not supported?
> 
> Password auth requires the `crypt(3)` function (as you mentioned in your #807) which is something we better implement correctly, then use - instead of adding a patch here.

Haha! So I did! I opened so many bug reports I lost track. I think most of them have been closed since.

Thanks

PR #2186, opened 2020-05-10
Fix parsing of numbers

### LINKED ISSUE #1979: LibC/LibJS: strtod() weirdness
This test is working fine on my host machine but fails on Serenity:

https://github.com/SerenityOS/serenity/blob/14c7988eeafd67722bbc246812d536345924c078/Libraries/LibJS/Tests/Number.isInteger.js#L11

I did some digging, the issue seems to be with Serenity's `strtod()` implementation:

https://github.com/SerenityOS/serenity/blob/95b51e857d8f77f8bf9c4e01a4388d42a4f49887/Libraries/LibJS/Token.cpp#L72

Simple program to reproduce:

```cpp
#include <AK/LogStream.h>
#include <stdlib.h>

int main(int, char**)
{
    double d1 = strtod("5.000000000000001", nullptr);
    double d2 = strtod("5.0000000000000001", nullptr);
    dbg() << (((i32)d1 == d1) ? "(i32)d1 == d1" : "(i32)d1 != d1");
    dbg() << (((i32)d2 == d2) ? "(i32)d2 == d2" : "(i32)d2 != d2");
}
```

Serenity:

```
(i32)d1 != d1
(i32)d2 != d2
```

Linux (via Lagom):

```
(i32)d1 != d1
(i32)d2 == d2
```

Some more weird stuff happening:

```
> 1.0000000000000000000000000000001
1.0000
> 1.00000000000000000000000000000001
Infinity
> 1.12345678912345678912345678912345
Infinity
> 1.123456789123456789123456789123456789123456789
Infinity
```

BenWiederhake:
This strtod implementation is not perfectly accurate, as evidenced by the test
(accuracy_strtod.cpp), but it is sufficiently close (up to 8 eps).

The main sources of inaccuracy are:
- Highly repeated division/multiplication by 'base'
  (Each operation brings a multiplicative error of 1+2^-53.)
- Loss during the initial conversion from long long to double (most prominently,
  69294956446009195 should first be rounded to 69294956446009200 and then
  converted to 69294956446009200.0 and then divided by ten, yielding
  6929495644600920.0. Currently, it converts first to double, can't represent
  69294956446009195.0, and instead represents 69294956446009190, which
  eventually yields 6929495644600919.0. Close, but technically wrong.)

I believe that these issues can be fixed by rewriting the part at and after
`    double value = digits.number();`
and I believe that the loss before that is acceptable.

Specifically, losing the exact exponent on overflow is obviously fine.
The only other loss occurs when the significant digits overflow a 'long long'.
But these store 64(-7ish) bits, and the mantissa of a double only stores 52 bits.
With a bit more thinking, one could probably get the error down to 1 or 2 eps.
(But not better.)

Furthermore, this PR fixes parsing of various ints, because I had to write new int parsing code anyway.

This fixes #1979. Yay! :D


linusg:
> This fixes #1979. Yay! :D

I am forever grateful :heart:

awesomekling:
Great improvements :D

PR #2188, opened 2020-05-10
LibGfx: Eliminate conditional branch in dither

BenWiederhake:
In theory, this should make dithering a teensy bit faster.


awesomekling:
Sure why not

PR #2189, opened 2020-05-11
LibWeb: Improve rendering of awesomekling.github.io :^)

linusg:
A bunch of small fixes specifically targeted at improving the rendering of https://awesomekling.github.io.

Even the code blocks and images show up - this is so cool :D

From this...

image

...to this:

image

image
image
image

### DIFF
--- a/AK/String.cpp
+++ b/AK/String.cpp
@@ -369,6 +369,20 @@ int String::replace(const String& needle, const String& replacement, bool all_oc
     return positions.size();
 }
 
+String String::trim_spaces() const
+{
+    size_t start = 0;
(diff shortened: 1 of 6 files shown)


shannonbooth:
wow

AK/String.cpp
```diff
@@ -369,6 +369,20 @@ int String::replace(const String& needle, const String& replacement, bool all_oc
     return positions.size();
 }
 
+String String::trim_spaces() const
```
awesomekling, line 372: Should this be `trim_whitespace()` and also remove tabs and newlines?

PR #2190, opened 2020-05-11
Travis: Fix Cache spam

BenWiederhake:
Back in 36ba0a35eecec73c3347c43aeb18f3ea4849117d I thought that Travis would
automagically delete the oldest files. Apparently it does not. This patch implements that.

Note that no dummy changes are needed, because BuildIt.sh lists itself
as a dependency for the Toolchain. Hooray for something that works!

Isn't it ironic how the approach that was meant to speed up builds slowed everything down? Anyway, this should make the build speedy again, and the overhead costs negligible.

### DIFF
--- a/Toolchain/BuildIt.sh
+++ b/Toolchain/BuildIt.sh
@@ -18,6 +18,10 @@ MAKE="make"
 MD5SUM="md5sum"
 NPROC="nproc"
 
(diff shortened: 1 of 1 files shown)


BenWiederhake:
Just for documentation, here's the really juicy bit: https://travis-ci.com/github/SerenityOS/serenity/builds/165089937#L729

PR #2192, opened 2020-05-11
Sleep: Allow fractional seconds, fix Kernel timing, fix overflow in Kernel

BenWiederhake:
My masterplan behind strtod: Supporting fractional seconds in `sleep`! :^)

This changes three interrelated things:
- Kernel: No longer overflow on large sleeps, which used to cause hard-to-predict actual delays
- Kernel: Fix "ticks_per_seconds" handling, because some code used to assume that there are always exactly 1000 ticks in a second.
- `/bin/sleep`: Support fractional seconds (even `inf`!)

Here's what it looked like intermediately:

Bildschirmfoto_2020-05-11_15-09-16

(I removed the debug output before pushing.)

### DIFF
--- a/Kernel/Process.cpp
+++ b/Kernel/Process.cpp
@@ -4610,13 +4610,14 @@ int Process::sys$clock_nanosleep(const Syscall::SC_clock_nanosleep_params* user_
             u64 time_to_wake = (requested_sleep.tv_sec * 1000 + requested_sleep.tv_nsec / 1000000);
             wakeup_time = Thread::current()->sleep_until(time_to_wake);
         } else {
-            u32 ticks_to_sleep = (requested_sleep.tv_sec * 1000 + requested_sleep.tv_nsec / 1000000);
+            u64 ticks_to_sleep = requested_sleep.tv_sec * TimeManagement::the().ticks_per_second();
+            ticks_to_sleep += requested_sleep.tv_nsec * TimeManagement::the().ticks_per_second() / 1000000000;
             if (!ticks_to_sleep)
                 return 0;
             wakeup_time = Thread::current()->sleep(ticks_to_sleep);
         }
         if (wakeup_time > g_uptime) {
-            u32 ticks_left = wakeup_time - g_uptime;
+            u64 ticks_left = wakeup_time - g_uptime;
             if (!is_absolute && params.remaining_sleep) {
                 if (!validate_write_typed(params.remaining_sleep)) {
                     // This can happen because the lock is dropped while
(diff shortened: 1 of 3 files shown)


Userland/sleep.cpp
```diff
@@ -24,16 +24,48 @@
...
+// The maximum integer we can definitely exactly represent as a double (2^52).
+// Note that clock_nanosleep can handle even larger numbers (up to around 2^63),
+// but I think it's fine to cap silently at a hundred million years.
+static const double MAX_SLEEPABLE = 4503599627370496;
+
 void handle_sigint(int)
 {
 }
 
+double parse_amount(char* arg, bool& ok) {
```
awesomekling, line 44: Please use `clang-format`.

Userland/sleep.cpp
```diff
@@ -24,16 +24,48 @@
...
+    // This can't go into AK/* because AK can't run anything from stdlib,
+    // so strtod is not available yet.
+    // TODO: Find a way to put this function into AK/
```
awesomekling, line 47: How about we move the bulk of `strtod()` into `AK` and make LibC's `strtod()` a thin wrapper around it? :)

Userland/sleep.cpp
```diff
@@ -42,22 +74,46 @@ int main(int argc, char** argv)
     }
 
     if (argc != 2) {
-        printf("usage: sleep <seconds>\n");
+        printf("Usage:\n        sleep <seconds>\n");
+        printf("\nArguments:\n        seconds Fractional amount of seconds, e.g. 3, 0.01, 1337.42, or inf. The number will be rounded down to the nearest possible amount (usually by 1 ms).\n");
```
awesomekling, line 78: \<sergey\>Please use `Core::ArgsParser`:smile:\</sergey\>

Userland/sleep.cpp
BenWiederhake: Right! I forgot it here.

Userland/sleep.cpp
BenWiederhake: Sure! I thought that would go beyond the scope of a single PR.

Userland/sleep.cpp
BenWiederhake: I wasn't sure whether it's overkill – looking at that string literal again, now I think `ArgsParser` is a better idea indeed!

Userland/sleep.cpp
awesomekling: Yeah you can do that in a separate PR ofc :)

elcuco: I would like to continue this issue. Code porting to master is quite trivial. Problem is the arg parsing, which is being worked on in https://github.com/SerenityOS/serenity/pull/2867. 

Technically - how to proceed ? Should I create a new PR for this issue, or my PR should be against this PR, and @BenWiederhake  should handle it? (I think closing this one and me handling it giving all credit to @BenWiederhake  is easier)

BenWiederhake:
I removed the commits that touch `sleep` from this commit. For history purposes, the worktree on my harddrive is now available as the tag [historic/dev-sleep-fractional-PR-2192](https://github.com/BenWiederhake/serenity/tree/historic/dev-sleep-fractional-PR-2192).

The remaining commits fix stuff in the Kernel, and seem to have been accepted without issue.

So I suggest that this PR (now without the commits that touch `sleep`, and freshly rebased for conflict-freeness) gets merged, and then @elcuco can proceed however desired. :)

elcuco:
https://github.com/SerenityOS/serenity/pull/2867 - as soon as this gets merged, I will make another PR for restoring support for `double` sleeping in the userspace utility (should be trivial, nah?)

Userland/sleep.cpp
elcuco: Has been done in 62283ade912bbe0c2ff92509a0ed2a787f330853 - no longer relevant.

PR #2195, opened 2020-05-11
LibMarkdown: Change MD Document parse API to return a RefPtr

FalseHonesty:
Now, markdown documents are now obtained via the static Document::parse
method, which returns a RefPtr<Document>, or nullptr on failure. This was mentioned in IRC, and will bring the Markdown parse API more in line with the rest of the Serenity's parse APIs.

### DIFF
--- a/Applications/Help/main.cpp
+++ b/Applications/Help/main.cpp
@@ -128,11 +128,10 @@ int main(int argc, char* argv[])
         auto buffer = file->read_all();
         StringView source { (const char*)buffer.data(), buffer.size() };
 
-        Markdown::Document md_document;
-        bool success = md_document.parse(source);
-        ASSERT(success);
+        auto md_document = Markdown::Document::parse(source);
+        ASSERT(md_document);
 
-        String html = md_document.render_to_html();
+        String html = md_document->render_to_html();
         auto html_document = Web::parse_html_document(html);
         page_view.set_document(html_document);
 
--- a/Applications/Help/main.cpp
+++ b/Applications/Help/main.cpp
@@ -128,10 +128,13 @@ int main(int argc, char* argv[])
         auto buffer = file->read_all();
         StringView source { (const char*)buffer.data(), buffer.size() };
 
-        auto md_document = Markdown::Document::parse(source);
-        ASSERT(md_document);
+        String html;
+        {
+            auto md_document = Markdown::Document::parse(source);
+            ASSERT(md_document);
+            html = md_document->render_to_html();
+        }
 
-        String html = md_document->render_to_html();
         auto html_document = Web::parse_html_document(html);
         page_view.set_document(html_document);
 
--- a/Libraries/LibMarkdown/CodeBlock.h
+++ b/Libraries/LibMarkdown/CodeBlock.h
@@ -26,6 +26,7 @@
 
 #pragma once
 
+#include <AK/OwnPtr.h>
 #include <LibMarkdown/Block.h>
 #include <LibMarkdown/Text.h>
 
@@ -33,11 +34,16 @@ namespace Markdown {
 
 class CodeBlock final : public Block {
 public:
-    virtual ~CodeBlock() override {}
+    CodeBlock(Text&& style_spec, const String& code)
+        : m_code(move(code))
+        , m_style_spec(move(style_spec))
+    {
+    }
+    virtual ~CodeBlock() override { }
 
     virtual String render_to_html() const override;
     virtual String render_for_terminal() const override;
-    virtual bool parse(Vector<StringView>::ConstIterator& lines) override;
+    static OwnPtr<CodeBlock> parse(Vector<StringView>::ConstIterator& lines);
 
 private:
     String style_language() const;
--- a/Libraries/LibMarkdown/Document.cpp
+++ b/Libraries/LibMarkdown/Document.cpp
@@ -75,25 +75,29 @@ static bool helper(Vector<StringView>::ConstIterator& lines, NonnullOwnPtrVector
     return true;
 }
 
-bool Document::parse(const StringView& str)
+RefPtr<Document> Document::parse(const StringView& str)
 {
     const Vector<StringView> lines_vec = str.lines();
     auto lines = lines_vec.begin();
+    auto document = adopt(*new Document);
+    auto& blocks = document->m_blocks;
 
     while (true) {
         if (lines.is_end())
-            return true;
+            break;
 
         if ((*lines).is_empty()) {
             ++lines;
             continue;
         }
 
-        bool any = helper<List>(lines, m_blocks) || helper<Paragraph>(lines, m_blocks) || helper<CodeBlock>(lines, m_blocks) || helper<Heading>(lines, m_blocks);
+        bool any = helper<List>(lines, blocks) || helper<Paragraph>(lines, blocks) || helper<CodeBlock>(lines, blocks) || helper<Heading>(lines, blocks);
 
         if (!any)
-            return false;
+            return nullptr;
     }
+
+    return document;
 }
 
 }
--- a/Libraries/LibMarkdown/Document.cpp
+++ b/Libraries/LibMarkdown/Document.cpp
@@ -67,19 +67,18 @@ String Document::render_for_terminal() const
 template<typename BlockType>
 static bool helper(Vector<StringView>::ConstIterator& lines, NonnullOwnPtrVector<Block>& blocks)
 {
-    NonnullOwnPtr<BlockType> block = make<BlockType>();
-    bool success = block->parse(lines);
-    if (!success)
+    OwnPtr<BlockType> block = BlockType::parse(lines);
+    if (!block)
         return false;
-    blocks.append(move(block));
+    blocks.append(block.release_nonnull());
     return true;
 }
 
-RefPtr<Document> Document::parse(const StringView& str)
+OwnPtr<Document> Document::parse(const StringView& str)
 {
     const Vector<StringView> lines_vec = str.lines();
     auto lines = lines_vec.begin();
-    auto document = adopt(*new Document);
+    auto document = make<Document>();
     auto& blocks = document->m_blocks;
 
     while (true) {
--- a/Libraries/LibMarkdown/Heading.cpp
+++ b/Libraries/LibMarkdown/Heading.cpp
@@ -59,26 +59,30 @@ String Heading::render_for_terminal() const
     return builder.build();
 }
 
-bool Heading::parse(Vector<StringView>::ConstIterator& lines)
+OwnPtr<Heading> Heading::parse(Vector<StringView>::ConstIterator& lines)
 {
     if (lines.is_end())
-        return false;
+        return nullptr;
 
     const StringView& line = *lines;
+    size_t level;
 
-    for (m_level = 0; m_level < (int)line.length(); m_level++)
-        if (line[(size_t)m_level] != '#')
+    for (level = 0; level < line.length(); level++)
+        if (line[level] != '#')
             break;
 
-    if (m_level >= (int)line.length() || line[(size_t)m_level] != ' ')
-        return false;
+    if (level >= line.length() || line[level] != ' ')
+        return nullptr;
 
-    StringView title_view = line.substring_view((size_t)m_level + 1, line.length() - (size_t)m_level - 1);
-    bool success = m_text.parse(title_view);
-    ASSERT(success);
+    StringView title_view = line.substring_view(level + 1, line.length() - level - 1);
+    auto text = Text::parse(title_view);
+    if (!text.has_value())
+        return nullptr;
+
+    auto heading = make<Heading>(move(text.value()), level);
 
     ++lines;
-    return true;
+    return heading;
 }
 
 }
--- a/Libraries/LibMarkdown/Heading.h
+++ b/Libraries/LibMarkdown/Heading.h
@@ -26,6 +26,7 @@
 
 #pragma once
 
+#include <AK/OwnPtr.h>
 #include <AK/StringView.h>
 #include <AK/Vector.h>
 #include <LibMarkdown/Block.h>
@@ -35,15 +36,20 @@ namespace Markdown {
 
 class Heading final : public Block {
 public:
-    virtual ~Heading() override {}
+    Heading(Text&& text, size_t level)
+        : m_text(move(text))
+        , m_level(level)
+    {
+    }
+    virtual ~Heading() override { }
 
     virtual String render_to_html() const override;
     virtual String render_for_terminal() const override;
-    virtual bool parse(Vector<StringView>::ConstIterator& lines) override;
+    static OwnPtr<Heading> parse(Vector<StringView>::ConstIterator& lines);
 
 private:
     Text m_text;
-    int m_level { -1 };
+    size_t m_level { 0 };
 };
 
 }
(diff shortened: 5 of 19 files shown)


bugaevc:
This just switches the external (`Document`) API though; what about all the internal `parse` methods?

FalseHonesty:
I was considering changing the internal methods as well, but I wasn't sure if that was desired behaviour. I can try and expand this to change the entire internal parse tree to follow the same API style if you wish

bugaevc:
Yes, would be great if the internals could also be switched over to this style.

Initially I didn't do this because I wanted the `parse()` method to be virtual; but then the `Document` class uses a templated function substituting concrete block types anyway, so that didn't pan out, so it'd be fine for `parse()` to be made a static non-virtual.

FalseHonesty:
I've never worked with smart pointers before, so I'm not entirely sure I'm using them correctly in these changes. Please let me know if anything here needs to be modified!

Libraries/LibMarkdown/Text.h
```diff
@@ -51,7 +52,7 @@ class Text final {
     String render_to_html() const;
     String render_for_terminal() const;
 
-    bool parse(const StringView&);
+    static OwnPtr<Text> parse(const StringView&);
```
bugaevc, line 55: Can't `Text` be simply returned by value? (as an `Optiomal<Text>` — and make sure to `move()` it where appropriate) Unlike, the other types, it's not a `Block` so we don't need it to be behind a pointer for that reason.

FalseHonesty: Sure, I can return it as an `Optional<Text>`, I just am not sure at all when to use smart pointers or not. My other question is about `move()`, when would it be appropriate to move? Apologies for the questions, but I've never worked with C++ before.

bugaevc: In case of LibMarkdown specifically, it makes sense to return subclasses of `Block` by pointer (so, by some smart pointer), because they have virtual methods, and you want to store them (multiple blocks of different concrete types) in a `Vector<SomePtr<Block>>`.

But `Text` is not a block, although it has a similar API to one. So no need for pointers. Though there's nothing preventing you from using a pointer with it, it's just unnecessary.

If you start passing the text by value, use `move()` basically anywhere you're moving it 🙂 to prevent implicit copying (yay C++). So, when you're constructing an `Optional` out of it, and when you're extracting it from the optional, and so on.

Or maybe scratch the optional, even — `Text` is designed in a way that parsing to it can never fail.

bugaevc: Oh, it will help if you add `AK_MAKE_NONCOPYABLE` to `Text`.

FalseHonesty: Ah thank you very much, that’s super helpful. Ill make these changes the next chance I get!

FalseHonesty: I believe that my latest changes are what you intended, though I'm still not super sure, so please let me know :)

Libraries/LibMarkdown/Heading.cpp
```diff
@@ -59,26 +59,30 @@ String Heading::render_for_terminal() const
...
 
-bool Heading::parse(Vector<StringView>::ConstIterator& lines)
+OwnPtr<Heading> Heading::parse(Vector<StringView>::ConstIterator& lines)
 {
     if (lines.is_end())
-        return false;
+        return nullptr;
 
     const StringView& line = *lines;
+    int level;
```
bugaevc, line 69: Why not make it a `size_t` and get rid of all those casts below?

Libraries/LibMarkdown/Heading.cpp
```diff
@@ -59,26 +59,30 @@ String Heading::render_for_terminal() const
...
-        return false;
+    if (level >= (int)line.length() || line[(size_t)level] != ' ')
+        return nullptr;
 
-    StringView title_view = line.substring_view((size_t)m_level + 1, line.length() - (size_t)m_level - 1);
-    bool success = m_text.parse(title_view);
-    ASSERT(success);
+    StringView title_view = line.substring_view((size_t)level + 1, line.length() - (size_t)level - 1);
+    auto text = Text::parse(title_view);
+    ASSERT(text.has_value());
```
bugaevc, line 79: Since we're here, let's return `nullptr` in this case instead of asserting — to better prepare for the possibility of text parsing suddenly becoming fallible some time.

Libraries/LibMarkdown/CodeBlock.h
```diff
@@ -26,25 +26,27 @@
...
-    virtual bool parse(Vector<StringView>::ConstIterator& lines) override;
+    static OwnPtr<CodeBlock> parse(Vector<StringView>::ConstIterator& lines);
 
 private:
     String style_language() const;
     Text::Style style() const;
 
     String m_code;
-    Text m_style_spec;
+    Text& m_style_spec;
```
bugaevc, line 49: No, this can't work. Block types need to own the text, not borrow it.

Libraries/LibMarkdown/List.cpp
```diff
@@ -66,19 +66,19 @@ String List::render_for_terminal() const
...
     auto flush_item_if_needed = [&] {
         if (first)
             return;
 
-        Text text;
-        bool success = text.parse(item_builder.string_view());
-        ASSERT(success);
-        m_items.append(move(text));
+        auto text = Text::parse(item_builder.string_view());
+        ASSERT(text.has_value());
```
bugaevc, line 80: Ditto

Libraries/LibMarkdown/List.h
```diff
@@ -26,6 +26,8 @@
 
 #pragma once
 
+#include <AK/NonnullOwnPtrVector.h>
```
bugaevc, line 29: Do we need it here anywhere?

Libraries/LibMarkdown/Text.h
```diff
@@ -26,12 +26,15 @@
 
 #pragma once
 
+#include <AK/Noncopyable.h>
+#include <AK/OwnPtr.h>
```
bugaevc, line 30: Do we need `OwnPtr` here?

bugaevc:
Looks great 😀! Left some comments inline.

FalseHonesty:
I think that this PR is finally in a good spot! Thank you very much for your help with this @bugaevc.

Libraries/LibMarkdown/Document.cpp
```diff
@@ -67,33 +67,36 @@ String Document::render_for_terminal() const
...
+    OwnPtr<BlockType> block = BlockType::parse(lines);
+    if (!block)
         return false;
-    blocks.append(move(block));
+    blocks.append(block.release_nonnull());
     return true;
 }
 
-bool Document::parse(const StringView& str)
+RefPtr<Document> Document::parse(const StringView& str)
```
bugaevc, line 77: Why is document returned in a `RefPtr` but everything else in a `OwnPtr`?

Not saying it's wrong, but if there's no particular reason for it let's not do it.

Libraries/LibMarkdown/Heading.cpp
```diff
@@ -59,26 +59,31 @@ String Heading::render_for_terminal() const
...
-    StringView title_view = line.substring_view((size_t)m_level + 1, line.length() - (size_t)m_level - 1);
-    bool success = m_text.parse(title_view);
-    ASSERT(success);
+    StringView title_view = line.substring_view(level + 1, line.length() - level - 1);
+    auto text = Text::parse(title_view);
+    if (!text.has_value())
+        return nullptr;
+
+    auto heading = make<Heading>(move(text.value()));
+    heading->m_level = level;
```
bugaevc, line 83: Sounds like the level should also be a constructor argument.

Libraries/LibMarkdown/Text.h
```diff
@@ -46,14 +48,18 @@ class Text final {
         Style style;
     };
 
+    explicit Text(Text&& text) : m_spans(move(text.m_spans)) {}
```
bugaevc, line 51: Is there any need for writing this constructor explicitly? It should get autogenerated.

Libraries/LibMarkdown/Text.h
```diff
@@ -46,14 +48,18 @@ class Text final {
...
     const Vector<Span>& spans() const { return m_spans; }
 
     String render_to_html() const;
     String render_for_terminal() const;
 
-    bool parse(const StringView&);
+    static Optional<Text> parse(const StringView&);
 
 private:
+    explicit Text() {}
```
bugaevc, line 61: No need for `explicit` on zero-arguments constructor.

Applications/Help/main.cpp
```diff
@@ -128,11 +128,10 @@ int main(int argc, char* argv[])
...
         StringView source { (const char*)buffer.data(), buffer.size() };
 
-        Markdown::Document md_document;
-        bool success = md_document.parse(source);
-        ASSERT(success);
+        auto md_document = Markdown::Document::parse(source);
+        ASSERT(md_document);
 
-        String html = md_document.render_to_html();
+        String html = md_document->render_to_html();
```
bugaevc, line 134: I'm thinking we should drop the `md_document` after this line. C++ has no explicit way to drop local variables, but you can use scopes:

```cpp
String html;

{
    auto md_document = Markdown::Document::parse(source);
    ASSERT(md_document);
    html = md_document->render_to_html();
}
```

Libraries/LibMarkdown/Text.h
FalseHonesty: I'm not sure, if I remove it, the `Paragraph` constructor complains about this:

bugaevc: So apparently it doesn't generate the move constructor if you delete the copy one?

Does writing `= default` work? And why make it `explicit`?

Libraries/LibMarkdown/Heading.h
```diff
@@ -35,15 +36,20 @@ namespace Markdown {
 
 class Heading final : public Block {
 public:
-    virtual ~Heading() override {}
+    Heading(Text&& m_text, int m_level)
+        : m_text(move(m_text))
```
bugaevc, line 40: I don't think you need to `move()` it here, it's already a `Text&&`

FalseHonesty: If I remove the `move`, it complains about calling a deleted constructor.

Libraries/LibMarkdown/Heading.h
```diff
@@ -35,15 +36,20 @@ namespace Markdown {
 
 class Heading final : public Block {
 public:
-    virtual ~Heading() override {}
+    Heading(Text&& m_text, int m_level)
```
bugaevc, line 39: Let's make `level`/`m_level` a `size_t` here too 🙂

Libraries/LibMarkdown/Text.h
FalseHonesty: Having the constructor be `Text(Text&& text) = default;` makes it work. Should I leave it like that?

Libraries/LibMarkdown/Heading.h
```diff
@@ -35,15 +36,20 @@ namespace Markdown {
...
 
     virtual String render_to_html() const override;
     virtual String render_for_terminal() const override;
-    virtual bool parse(Vector<StringView>::ConstIterator& lines) override;
+    static OwnPtr<Heading> parse(Vector<StringView>::ConstIterator& lines);
 
 private:
     Text m_text;
-    int m_level { -1 };
+    int m_level;
```
bugaevc, line 52: It still needs done initial value even if it's directly assigned in the constructor, because @awesomekling is paranoid for that sort of thing. When you make it a `size_t`, you wouldn't be able to use -1, so we'll have to do with 0.

Libraries/LibMarkdown/List.cpp
```diff
@@ -66,21 +66,24 @@ String List::render_for_terminal() const
...
 
-        Text text;
-        bool success = text.parse(item_builder.string_view());
-        ASSERT(success);
-        m_items.append(move(text));
+        auto text = Text::parse(item_builder.string_view());
+        if (!text.has_value())
+            return false;
+
+        list->m_items.append(move(text.value()));
```
bugaevc, line 83: Could we, perhaps, instead of appending items directly to `list->m_items`, keep our own vector of items, and pass it to the `List` constructor, along with whether it's ordered, at the very end?

Libraries/LibMarkdown/Text.cpp
```diff
@@ -181,12 +181,13 @@ String Text::render_for_terminal() const
...
 }
 
-bool Text::parse(const StringView& str)
+Optional<Text> Text::parse(const StringView& str)
 {
     Style current_style;
     size_t current_span_start = 0;
     int first_span_in_the_current_link = -1;
     bool current_link_is_actually_img = false;
+    Text text;
```
bugaevc, line 191: Same here, let's construct `Text` at the very end.

Libraries/LibMarkdown/Text.h
bugaevc: Yeah

Libraries/LibMarkdown/CodeBlock.h
```diff
@@ -26,18 +26,20 @@
...
 
+#include <AK/OwnPtr.h>
 #include <LibMarkdown/Block.h>
 #include <LibMarkdown/Text.h>
 
 namespace Markdown {
 
 class CodeBlock final : public Block {
 public:
+    CodeBlock(Text&& m_style_spec, const String& m_code) : m_code(move(m_code)), m_style_spec(move(m_style_spec)) {}
```
bugaevc, line 37:
```suggestion
    CodeBlock(Text&& style_spec, const String& code)
        : m_code(code)
        , m_style_spec(m_style_spec)
        { }
```

Libraries/LibMarkdown/List.h
```diff
@@ -34,11 +35,17 @@ namespace Markdown {
 
 class List final : public Block {
 public:
+    List(Vector<Text>& m_text, bool m_is_ordered)
```
bugaevc, line 38:
```suggestion
    List(Vector<Text>&& text, bool is_ordered)
```

Libraries/LibMarkdown/Text.h
```diff
@@ -46,14 +48,18 @@ class Text final {
...
     const Vector<Span>& spans() const { return m_spans; }
 
     String render_to_html() const;
     String render_for_terminal() const;
 
-    bool parse(const StringView&);
+    static Optional<Text> parse(const StringView&);
 
 private:
+    Text(Vector<Span>& m_spans) : m_spans(move(m_spans)) {}
```
bugaevc, line 61:
```suggestion
    Text(Vector<Span>&& spans)
         : m_spans(spans)
     { }
```

Libraries/LibMarkdown/Paragraph.h
```diff
@@ -26,18 +26,20 @@
...
 
+#include <AK/OwnPtr.h>
 #include <LibMarkdown/Block.h>
 #include <LibMarkdown/Text.h>
 
 namespace Markdown {
 
 class Paragraph final : public Block {
 public:
+    explicit Paragraph(Text&& m_text) : m_text(move(m_text)) {}
```
bugaevc, line 37: Here, too.

Libraries/LibMarkdown/Text.h
```diff
@@ -46,14 +48,21 @@ class Text final {
...
     const Vector<Span>& spans() const { return m_spans; }
 
     String render_to_html() const;
     String render_for_terminal() const;
 
-    bool parse(const StringView&);
+    static Optional<Text> parse(const StringView&);
 
 private:
+    Text(Vector<Span>&& m_spans)
```
bugaevc, line 61: And here

PR #2197, opened 2020-05-11
LibHTTP: Support Transfer-Encoding: chunked

alimpfard:
We advertise ourselves to servers as supporting HTTP/1.1; we should put our money where our mouth is, and start supporting some of its features.

p.s. I tried to open nyaa.si/?q=foo
> muh anime torrents

### DIFF
--- a/Libraries/LibC/stdlib.cpp
+++ b/Libraries/LibC/stdlib.cpp
@@ -929,6 +929,9 @@ long long strtoll(const char* str, char** endptr, int base)
         return 0;
     }
 
+    if (endptr)
+        *endptr = parse_ptr;
+
     if (overflow) {
         errno = ERANGE;
         if (sign != Sign::Negative) {
@@ -1002,6 +1005,9 @@ unsigned long long strtoull(const char* str, char** endptr, int base)
             *endptr = const_cast<char*>(str);
         return 0;
     }
+
+    if (endptr)
+        *endptr = parse_ptr;
 
     if (overflow) {
         errno = ERANGE;
--- a/Libraries/LibHTTP/Job.cpp
+++ b/Libraries/LibHTTP/Job.cpp
@@ -117,7 +117,7 @@ void Job::on_socket_connected()
             m_state = State::InHeaders;
             return;
         }
-        if (m_state == State::InHeaders) {
+        if (m_state == State::InHeaders || m_state == State::AfterChunkedEncodingTrailer) {
             if (!can_read_line())
                 return;
             auto line = read_line(PAGE_SIZE);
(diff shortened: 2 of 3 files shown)


Libraries/LibHTTP/Job.cpp
```diff
@@ -151,7 +151,50 @@ void Job::on_socket_connected()
...
+                if (remaining == -1) {
+                    // read size
+                    auto size_data = read_line(PAGE_SIZE);
+                    auto size_lines = StringView { size_data.data(), size_data.size() }.lines();
+                    if (size_lines.size() == 0) {
+                        dbg() << "Job: Reached end of stream";
+                        read_line(PAGE_SIZE);
+                    } else {
+                        String size_string = size_lines[0];
+                        auto size = strtoul(size_string.characters(), nullptr, 16);
```
awesomekling, line 167: What if it's not a valid number?

Libraries/LibHTTP/Job.cpp
```diff
@@ -163,9 +206,37 @@ void Job::on_socket_connected()
...
+                    size = -1;
+                    auto line = read_line(PAGE_SIZE);
+#ifdef JOB_DEBUG
+                    dbg() << "Line following (should be empty): _" << line << "_";
+#endif
+                    (void)line;
+
+                    if (m_current_chunk_total_size.value() == 0) {
+                        // this was the last chunk, read the rest and discard it.
+                        (void)receive(64 * KB);
```
awesomekling, line 232: Why is 64 KB the right max number of bytes to receive to "read the rest"?

Libraries/LibHTTP/Job.cpp
alimpfard: hmmm...I suppose we should fail then.

now how does that endptr work...?

Libraries/LibHTTP/Job.cpp
alimpfard: I figured it's big enough for a single chunk.
But this actually leads into another problem I just discovered, chunked encoding allows _more_ headers after the last chunk

Libraries/LibHTTP/Job.cpp
alimpfard: > how does that endptr work

It actually doesn't.
I'll add the fix to this PR, if that's alright?

awesomekling: Ofc!

Libraries/LibC/stdlib.cpp
```diff
@@ -1003,6 +1003,9 @@ unsigned long long strtoull(const char* str, char** endptr, int base)
         return 0;
     }
 
+    if (endptr)
+        *endptr = parse_ptr;
```
awesomekling, line 1007: How about we also fix `strtoll()` while we're here :]

PR #2199, opened 2020-05-11
Build: Improved speed of makeall.sh

Nicholas-Baron:
`Kernel/makeall.sh -f` now uses the `-j` option on make with the number of processors from `nproc`.
Additionally, if `nproc` returns greater than two processors, the same `-j` option is used.


bcoles:
Has this been tested?

I believe there was a reason that the `makeall` script does not build in parallel.

awesomekling:
`makeall.sh` is not safe for running in parallel at the moment, as the build system does not properly handle cross-directory dependencies. This will be fixed with #2132 :)

Issue #2201, opened 2020-05-11
LibWeb: Non-spec margin percentage processing. 


Vikingat-RAGE:
Percentages of margin-top and margin-bottom refer to the width of the container not the height.https://www.w3.org/TR/CSS22/box.html#propdef-margin-top.

Same is also true for padding.

Vikingat-RAGE:
Submitted PR. https://github.com/SerenityOS/serenity/pull/2202

PR #2202, opened 2020-05-12
LibWeb: Fixed non-spec processing of margin and padding

Vikingat-RAGE:
Closes https://github.com/SerenityOS/serenity/issues/2201


awesomekling:
Excellent! Thank you for paying attention to what the text actually says :smile:

PR #2203, opened 2020-05-12
LibJS: Add missing keywords/tokens

linusg:
Some of these are required for syntax we have not implemented yet, some are future reserved words in strict mode.


Zenyiar:
Why was Enum added? Afaik (as far as I know), Enums are only in Typescript

linusg:
Try `enum = "foo"` in any JavaScript engine. It doesn't work.

> https://www.ecma-international.org/ecma-262/5.1/#sec-7.6.1.2
>
> The following words are used as keywords in proposed extensions and are therefore reserved to allow for the possibility of future adoption of those extensions.
>
> Syntax 
> FutureReservedWord :: one of
> class enum extends super const export import

(that is for ECMAScript 5.1, later revisions already made all of these except `enum` part of the syntax of course.)

Zenyiar:
Ah

PR #2208, opened 2020-05-12
LibJS: Make string to number coercion work for doubles & trim whitespace beforehand // AK: Replace String::trim_spaces() with String::trim_whitespace()

linusg:
- LibJS: Make string to number coercion work for doubles

  Well... just that. `+"1.23"` works now.

- AK: Replace `String::trim_spaces()` with `String::trim_whitespace()`

  As suggested by @awesomekling in a code review and (initially) ignored by me :^)

  Implementation is roughly based on LibJS's `trim_string()`, but with a fix for trimming all-whitespace strings.

- LibJS: Use `String::trim_whitespace()` for `String.prototype.trim*()`

  No need to keep a second string trimming function with a bug around.

- LibJS: Trim whitespace from string before coercing to number

  That's how it's supposed to work. `+"<long string only whitespace>" === 0`.

### DIFF
--- a/AK/String.cpp
+++ b/AK/String.cpp
@@ -369,18 +369,42 @@ int String::replace(const String& needle, const String& replacement, bool all_oc
     return positions.size();
 }
 
-String String::trim_spaces() const
+String String::trim_whitespace(TrimMode mode) const
 {
-    size_t start = 0;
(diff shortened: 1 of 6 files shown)


awesomekling:
:star:

PR #2211, opened 2020-05-13
WindowServer: Invalidate the cursor when setting an override cursor

Dexesttp:
Fixes that one bug from the end of last video :^)


awesomekling:
Excellent! Thanks @Dexesttp :D

PR #2212, opened 2020-05-13
Shell: Break out of continuation when ^C is pressed

alimpfard:
This fixes the little issue with Shell not allowing cancellation of commands once they were in continuation mode

```sh
$ ls '
$ # No matter what we do here, we cannot escape 'ls'
```

### DIFF
--- a/Shell/main.cpp
+++ b/Shell/main.cpp
@@ -1540,8 +1540,25 @@ int main(int argc, char** argv)
 
     StringBuilder complete_line_builder;
 
+    bool should_break_current_command { false };
+
+    editor.on_interrupt_handled = [&] {
+        if (s_should_continue != ExitCodeOrContinuationRequest::ContinuationRequest::Nothing) {
+            should_break_current_command = true;
+            editor.finish();
+        }
+    };
+
     for (;;) {
         auto line = editor.get_line(prompt());
+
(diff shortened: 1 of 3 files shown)


Shell/main.cpp
```diff
@@ -655,7 +655,7 @@ static bool handle_builtin(int argc, const char** argv, int& retval)
 
 class FileDescriptionCollector {
 public:
-    FileDescriptionCollector() {}
+    FileDescriptionCollector() { }
```
alimpfard, line 658: I feel like I see this every time
has the project .clang-format changed without me noticing?

Dexesttp: On my end, formatting removes the extra space. Maybe it's a clang-format version issue ?

I'm using the `clang-format-9 --style=file -i Shell/main.cpp` command directly.

My clang version is: `9.0.0-2~ubuntu18.04.2 (tags/RELEASE_900/final)`

alimpfard: it's clang-format 10.0.0 here.

lemme see if I can grab a build of clang-format 9 to test with

alimpfard: it's `SpaceInEmptyBlocks` and the option is added in clang-format 11
:facepalm:

PR #2214, opened 2020-05-13
LibWeb: Make CSS properties and most values case insensitive / Support "transparent" CSS color value

linusg:


Libraries/LibWeb/CodeGenerators/Generate_CSS_PropertyID_cpp/Generate_CSS_PropertyID_cpp.cpp
```diff
@@ -59,15 +59,17 @@ int main(int argc, char** argv)
...
     out() << "namespace CSS {";
 
-    out() << "PropertyID property_id_from_string(const StringView& string) {";
+    out() << "PropertyID property_id_from_string(const StringView& view) {";
+    out() << "    auto string_lowercase = view.to_string().to_lowercase();";
 
     json.as_object().for_each_member([&](auto& name, auto& value) {
         ASSERT(value.is_object());
-        out() << "    if (string == \"" << name << "\")";
+        out() << "    if (string_lowercase == \"" << name << "\")";
```
awesomekling, line 72: Can we use `string.equals_ignoring_case()` here instead of converting the input to lowercase?

Libraries/LibWeb/Parser/CSSParser.cpp
```diff
@@ -136,20 +136,23 @@ static Optional<float> parse_number(const StringView& view)
 
 NonnullRefPtr<StyleValue> parse_css_value(const StringView& string)
 {
-    auto number = parse_number(string);
+    auto string_lowercase = string.to_string().to_lowercase();
```
awesomekling, line 142: Same here, we can use `equals_ignoring_case` and avoid creating a temporary string.

linusg:
Done. Still had to use `to_lower_case()` in two places (where we use `starts/ends_with()`) - maybe we should have `starts/ends_with_ignoring_case()`?

awesomekling:
Hmm, maybe! But this is okay for now :)

Issue #2216, opened 2020-05-14
Userspace: Exception Handling


ADKaster:
- LibUnwind: Create a replacement for the stack unwinding library facilities in libgcc.a
    - [ ] Analyze the Itanium exception handling ABI and the i386 System V processor specific ABI for the expected `_Unwind_*` definitions
   - [ ] Implement [Base ABI](https://itanium-cxx-abi.github.io/cxx-abi/abi-eh.html#base-abi)
      - [ ] Using SJLJ exceptions? (gross!)
      - [ ] Using DWARF .eh_frame Common Frame Information exceptions (modern!)
      - [ ] Create custom implementation of personality routine `__gxx_personality_v0` (if .eh_frame_hdr) or `__gxx_personality_sj0` (if using setjump/longjump)
   - [ ] Configure GCC to use `--with-system-libunwind` so that it doesn't build `_Unwind_*` symbols into libgcc.a
- LibC: Implement Itanium C++ abi for exception throwing and catching
   - [ ] Analyze Itanium exception handling ABI part II for what `__cxa_*` methods are required
   - [ ] Look into what Itanium ABI methods are required to support type info?
       - [ ] Implement whatever is necessary to make RTTI work. 
   - [ ] Make it possible to call __cxa_throw, __cxa_begin_catch, __cxa_unexpected, etc
- LibDebug: Parse CIEs and FDEs from .eh_frame and .debug_frame ELF sections
   - [ ] [Note](https://refspecs.linuxfoundation.org/LSB_5.0.0/LSB-Core-generic/LSB-Core-generic/ehframechpt.html) that .eh_frame uses version 1 and .debug_frame uses version pick-your-favorite-DWARF-revision (4?)
   - [ ] Parse EH Frame header or debug frame header using ByteBuffers into a list of Common Information Entries and Frame Description Entries
  - [ ] Identify all the common frame state machine instructions that the exception info and debug info sections are telling us to do in all these CIE objects
  - [ ] Make a parser/execution state machine for running the important frame instructions on a frame
  -  [ ] Fold into backtrace capability for LibDebug, and better exception handling

bugaevc:
But why?..

Let's maybe not have exceptions?

ADKaster:
Why not?

Need to parse the frame information to get reasonable debugging backtraces
and such, might as well go the extra mile to get exceptions in userspace.

awesomekling:
Seems fine to support it if it's relatively non-invasive.

[...]

PR #2217, opened 2020-05-14
LibUnwind: Create UnwindBase.h to capture the EH ABI in code

ADKaster:
The header is essentially just a commented + annotated version of both
the Itanium Exception Handling ABI Level I, and the 386-specific
SystemV ABI for the same. Wrapping one's head around the spec is half
the work :^)

Bullet 1 of #2216

I swear i won't drop this project on the floor after two weeks :)


awesomekling:
:^)

PR #2218, opened 2020-05-14
Implement basic Gemini protocol support for Browser

deoxxa:


Libraries/LibGemini/GeminiJob.cpp
```diff
@@ -0,0 +1,136 @@
...
+{
+    ASSERT(!m_socket);
+    m_socket = TLS::TLSv12::construct(this);
+    m_socket->on_tls_connected = [this] {
+#ifdef GEMINIJOB_DEBUG
+        dbg() << "GeminiJob: on_connected callback";
+#endif
+        on_socket_connected();
+    };
+    m_socket->on_tls_error = [&](TLS::AlertDescription error) {
```
bugaevc, line 48:
```suggestion
    m_socket->on_tls_error = [this](TLS::AlertDescription error) {
```

Libraries/LibGemini/GeminiJob.cpp
```diff
@@ -0,0 +1,136 @@
...
+            deferred_invoke([this](auto&) {
+                return did_fail(Core::NetworkJob::Error::ConnectionFailed);
+            });
+        } else {
+            deferred_invoke([this](auto&) {
+                return did_fail(Core::NetworkJob::Error::TransmissionFailed);
+            });
+        }
+    };
+    m_socket->on_tls_finished = [&] {
```
bugaevc, line 63:
```suggestion
    m_socket->on_tls_finished = [this] {
```

Libraries/LibGemini/GeminiJob.h
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    virtual ~GeminiJob() override
+    {
+    }
+
+    virtual void start() override;
+    virtual void shutdown() override;
+
+protected:
```
bugaevc, line 52: `protected` in a `final` class?

Libraries/LibGemini/Job.cpp
```diff
@@ -0,0 +1,161 @@
...
+{
+}
+
+Job::~Job()
+{
+}
+
+void Job::on_socket_connected()
+{
+    register_on_ready_to_write([&] {
```
bugaevc, line 47:
```suggestion
    register_on_ready_to_write([this] {
```

Libraries/LibGemini/Job.cpp
```diff
@@ -0,0 +1,161 @@
...
+        auto raw_request = m_request.to_raw_request();
+#ifdef JOB_DEBUG
+        dbg() << "Job: raw_request:";
+        dbg() << String::copy(raw_request).characters();
+#endif
+        bool success = write(raw_request);
+        if (!success)
+            deferred_invoke([this](auto&) { did_fail(Core::NetworkJob::Error::TransmissionFailed); });
+    });
+    register_on_ready_to_read([&] {
```
bugaevc, line 60:
```suggestion
    register_on_ready_to_read([this] {
```

Libraries/LibGemini/Job.cpp
```diff
@@ -0,0 +1,161 @@
...
+
+            auto line = read_line(PAGE_SIZE);
+            if (line.is_null()) {
+                fprintf(stderr, "Job: Expected status line\n");
+                return deferred_invoke([this](auto&) { did_fail(Core::NetworkJob::Error::TransmissionFailed); });
+            }
+
+            auto parts = String::copy(line, Chomp).split_limit(' ', 2);
+            if (parts.size() != 2) {
+                fprintf(stderr, "Job: Expected 2-part status line, got '%s'\n", line.data());
```
bugaevc, line 76: It's *probably* a bad idea to output non-guaranteed-correct UTF-8 as if it was a string; but OTOH I can't think of anything better at the moment. Maybe we should have an utility for these cases that would escape weird bytes or something.

Libraries/LibGemini/Job.cpp
```diff
@@ -0,0 +1,161 @@
...
+            } else {
+                fprintf(stderr, "Job: Expected status between 10 and 69; instead got %d\n", m_status);
+                return deferred_invoke([this](auto&) { did_fail(Core::NetworkJob::Error::ProtocolFailed); });
+            }
+
+            return;
+        }
+
+        ASSERT(m_state == State::InBody || m_state == State::Finished);
+        ASSERT(can_read());
```
bugaevc, line 110: What if the remote closes the connection at this point?

Moreover, does `can_read()` return `true` if you could read more, but that would block?

Libraries/LibGemini/GeminiJob.h
deoxxa: I actually have no idea how `protected` and `private` differ in this context - I shamelessly stole the structure for this from `HttpsJob` in LibHTTP. If we change this here, should we address the same issue in LibHTTP?

Libraries/LibGemini/Job.h
```diff
@@ -0,0 +1,80 @@
...
+
+    GeminiResponse* response() { return static_cast<GeminiResponse*>(Core::NetworkJob::response()); }
+    const GeminiResponse* response() const { return static_cast<const GeminiResponse*>(Core::NetworkJob::response()); }
+
+protected:
+    void finish_up();
+    void on_socket_connected();
+    virtual void register_on_ready_to_read(Function<void()>) = 0;
+    virtual void register_on_ready_to_write(Function<void()>) = 0;
+    // FIXME: I want const but Core::IODevice::can_read_line populates a cache with this
```
bugaevc, line 53: Then `Core::IODevice::can_read_line()` should be `const` and make its cache `mutable`

Services/ProtocolServer/GeminiDownload.cpp
```diff
@@ -0,0 +1,64 @@
...
+            set_payload(response->payload());
+            if (!response->meta().is_empty()) {
+                HashMap<String, String, CaseInsensitiveStringTraits> headers;
+                headers.set("meta", response->meta());
+                set_response_headers(headers);
+            }
+        }
+
+        // if we didn't know the total size, pretend that the download finished successfully
+        // and set the total size to the downloaded size
```
bugaevc, line 46: Does it actually pretend it finished successfully? To me this reads like it just reports a 100% loading progress, not any kind of success.

Libraries/LibGemini/Job.cpp
deoxxa: Actually this is totally broken if there's no body - we should only be checking for `can_read()` when we're expecting a response body. Good catch!

Services/ProtocolServer/GeminiDownload.cpp
```diff
@@ -0,0 +1,64 @@
...
+
+        did_finish(success);
+    };
+    m_job->on_progress = [this](Optional<u32> total, u32 current) {
+        did_progress(total, current);
+    };
+}
+
+GeminiDownload::~GeminiDownload()
+{
```
bugaevc, line 58: Should this destroy the job and/or unlink itself from its callbacks?

Services/ProtocolServer/GeminiDownload.h
```diff
@@ -0,0 +1,45 @@
...
+
+class GeminiProtocol;
+
+class GeminiDownload final : public Download {
+public:
+    virtual ~GeminiDownload() override;
+    static NonnullRefPtr<GeminiDownload> create_with_job(Badge<GeminiProtocol>, PSClientConnection&, NonnullRefPtr<Gemini::GeminiJob>&&);
+
+private:
+    explicit GeminiDownload(PSClientConnection&, NonnullRefPtr<Gemini::GeminiJob>&&);
```
bugaevc, line 42: How does `&&`ing a `NonnullRefPtr` work, even?

Libraries/LibGemini/Job.cpp
bugaevc: That's not what I meant actually (good catch yourself!)

I mean if the remote abruptly closes the connection when we expect the body, we should return a failure rather than aborting.

Libraries/LibGemini/Job.h
alimpfard: That's a thing?

C++: where every standard adds random things you never knew you needed, because you never wanted them.

Services/ProtocolServer/GeminiDownload.cpp
alimpfard: the proceeding `did_finish` implies success

Services/ProtocolServer/GeminiDownload.h
alimpfard: just forces it to be moved, I imagine?
IIRC `NonnullRefPtr` doesn't increment the refcount on move with `NonnullRefPtr&&`

Libraries/LibGemini/Job.h
bugaevc: It could totally be a type:

```c++
template<typename T>
class Mutable {
public:
    Mutable(T value): m_value(value) {}
    T& operator *() const { return const_cast<T&>(m_value); }
private:
    T m_value;
};


// Later:
Mutable<Cache> m_some_cache;
```

But they decided to go with a keyword instead.

alimpfard: Hmm...could be a nice addition; I personally didn't know the purpose of `mutable` until now
`Mutable<T>` would be quite clear about its purpose though.

alimpfard:
Further request, `URL` should probably default to port 1965 for `gemini://`

Services/ProtocolServer/GeminiDownload.h
bugaevc: Yes, but how does moving a <code><strong>Nonnull</strong>RefPtr</code> work? What does it leave in place when you move out of it?

alimpfard: The same content a before?
The impl seems to just leak the reference

bugaevc: So it does leave null in place (not so non-null after all!), but at least it asserts if you try to access it later:

https://github.com/SerenityOS/serenity/blob/3c3c1b40131b41454c59c4f5c8f6eaf9094556a1/AK/NonnullRefPtr.h#L80-L84
https://github.com/SerenityOS/serenity/blob/3c3c1b40131b41454c59c4f5c8f6eaf9094556a1/AK/NonnullRefPtr.h#L165-L171

alimpfard: Yeah, that's a bit surprising.
But I can see some sense in the behaviour, I guess

Libraries/LibGemini/Job.h
awesomekling: I don't think we should add a class to reimplement a language keyword :)

alimpfard: shame.
Maybe next time :)

Libraries/LibGemini/Job.cpp
deoxxa: Ah! Yes, I see what you mean now. Two bugs for the price of one!

Services/ProtocolServer/GeminiDownload.cpp
deoxxa: This comment brought to you by CTRL+C and CTRL+V, in association with: https://github.com/SerenityOS/serenity/blob/3c3c1b40131b41454c59c4f5c8f6eaf9094556a1/Services/ProtocolServer/HttpsDownload.cpp#L41-L44

Libraries/LibGemini/Job.cpp
deoxxa: I've decided the right thing to do here is just to delete the `can_read()` check completely - a zero-length body is totally valid, and results in EOF directly after sending the headers.

Libraries/LibGemini/Job.h
deoxxa: I've opened an issue for this here - https://github.com/SerenityOS/serenity/issues/2219 - I think this deserves its own change set.

alimpfard:
I likey this thing
I'm gonna try write a parser for `text/gemini` (-> html, because that seems to be our convention)
it seems to be simple enough

Services/ProtocolServer/GeminiDownload.cpp
deoxxa: If that's the case we should do the same thing for `HttpDownload` and `HttpsDownload` too. It certainly *looks* like we should do this, since I think what's happening right now is that the request keeps coming in, but we throw the bytes away by not reading them. It'd be better if we shut the request/response down upon stopping the `Download` itself.

deoxxa:
@alimpfard nice! Going via HTML is certainly going to be the quickest path.

alimpfard:
It's the web I've always wanted!
So damn cool.

shot-2020-05-14_20-18-09

bugaevc:
Something went wrong with htose bullet markers tho

alimpfard:
Yeah, I was putting a br after everything.
It's okay now~

alimpfard:
@deoxxa  Your current changes work just fine if you make PSClientConnection hold a RefPtr to the download instead of an OwnPtr
(Looks like it still drops the download when the client disconnects)

deoxxa:
@alimpfard I just had one remaining use-after-free bug to squash, which is now done. Hooray!

Libraries/LibHTTP/Job.cpp
```diff
@@ -268,7 +268,7 @@ void Job::on_socket_connected()
                     content_length = length;
             }
 
-            did_progress(content_length, m_received_size);
+            deferred_invoke([this, content_length](auto&) { did_progress(content_length, m_received_size); });
```
alimpfard, line 271: Have you tested https with this change?
I vaguely remember having problems with it

deoxxa: Only briefly - loading pages seems to work okay in the browser and some simple `pro` tests went through okay. Is there any specific circumstance I should be looking at?

alimpfard:
Nah, it didn't post the events when I tried that.
If pro works fine, then it's just fine

Issue #2207, opened 2020-05-12
LibJS: Abort parsing on first syntax error?


linusg:
I always found it odd that we continue parsing after encountering a syntax error, from which we cannot possibly recover anyway, and "fill" the missing parts of the AST with `ErrorExpression` and `ErrorStatement`.

I think we should just get rid of these two AST nodes, abort parsing, discard the partial AST and never have and show more than one syntax error.

cc @sunverwerth (maybe you can shed some light on this :) )

edit - example:

```
> foo)bar]baz
Syntax Error: Unexpected token ParenClose. Expected Semicolon (line: 1, column: 5)
Syntax Error: Unexpected token ParenClose. Expected statement (line: 1, column: 5)
Syntax Error: Unexpected token BracketClose. Expected Semicolon (line: 1, column: 9)
Syntax Error: Unexpected token BracketClose. Expected statement (line: 1, column: 9)
Parse error
```

awesomekling:
Totally agree. The "lol I'll just keep parsing" behavior is not really useful for anything :)

sunverwerth:
It *is* useful in one case, and that is in a code editor context. :) I adopted this style from contributing to the microsoft/tolerant-php-parser repo. Although I admit that "recovering" from syntax errors needs some more contextual info which we currently not provide. So errors after the first one are kinda useless at the moment.

It also simplifies the parser, as we can always expect a non null node as a return value. Aborting early means we either have to check every return value or use exceptions, which we don't have.

¯\\_(ツ)_/¯

What do?

awesomekling:
That makes some sense @sunverwerth. I like the parser simplification argument.
However, we should definitely make the parser not dump hundreds of errors like it does now :)

sunverwerth:
Since @ADKaster is working on exceptions, we could turn off error output after the first message for now and make it throw exceptions later. Maybe add a switch than turns errors-as-exceptions on/off?

awesomekling:
Exception support is strictly for third party software. The absence of exceptions is very much a feature :)

sunverwerth:
Dang! Ok. :smile:

linusg:
Let's keep it like this for now, I'm happy with the changes in #2221.

PR #2221, opened 2020-05-14
LibJS: Clean up how parser errors work

linusg:
Addresses https://github.com/SerenityOS/serenity/issues/2207, I guess.

- LibJS: Remove syntax errors from lexer 

  Giving the lexer the ability to generate errors adds unnecessary complexity - also it only calls its `syntax_error()` function in one place anyway ("unterminated string literal"). But since the lexer *also* emits tokens like `Eof` or `UnterminatedStringLiteral`, it should be up to the consumer of these tokens to decide what to do.

  Also remove the option to not print errors to stderr as that's not relevant anymore.

- LibJS: Let parser keep track of errors

  Rather than printing them to stderr directly the parser now keeps a `Vector<Error>`, which allows the "owner" of the parser to consume them individually after parsing.

  The `Error` struct has a message, line number, column number and a `to_string()` helper function to format this information into a meaningful error message.

  The `Function()` constructor will now include an error message when throwing a `SyntaxError`.

- js: Throw a regular SyntaxError for errors from the parser

  ```
  > foo]bar)baz
  Uncaught exception: [SyntaxError]: Unexpected token BracketClose. Expected Semicolon (line: 1, column: 5)
  >
  ```

### DIFF
--- a/Libraries/LibJS/Lexer.h
+++ b/Libraries/LibJS/Lexer.h
@@ -37,14 +37,8 @@ namespace JS {
 class Lexer {
 public:
     explicit Lexer(StringView source);
-    Lexer(StringView source, bool log_errors)
-        : Lexer(source)
-    {
-        m_log_errors = log_errors;
-    }
 
     Token next();
-    bool has_errors() const { return m_has_errors; }
 
 private:
     void consume();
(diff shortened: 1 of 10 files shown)


Libraries/LibJS/Lexer.h
```diff
@@ -37,14 +38,9 @@ namespace JS {
 class Lexer {
 public:
     explicit Lexer(StringView source);
-    Lexer(StringView source, bool log_errors)
-        : Lexer(source)
-    {
-        m_log_errors = log_errors;
-    }
+    // explicit Lexer(StringView source, AK::Function<void(String message, size_t line, size_t column)> on_error);
```
Dexesttp, line 41: With the new error-free lexer, this won't be needed anymore and should probably be removed. I guess that it was leftover from the lexer refactoring ?

linusg: Yep, that was me playing around with a lexer-just-reports-error-someone-else-takes-care-of-it approach. Thanks!

sunverwerth:
image

alimpfard:
@sunverwerth 
image

PR #2224, opened 2020-05-14
Meta: Update CodingStyle.md and BuildInstructions.md

### LINKED ISSUE #2213: clang-format gives different results in version 8-9 or version 10-11
Found during #2212

When using clang-format version 8 or 9, the formatter removes spaces inside of empty code blocks (usually inside destructors) :

```c++
class MyClass {
public:
    ~MyClass() {}
}
```

Starting from clang 10, the SpaceInEmptyBlock option was added enabled by default in Webkit (as per the Webkit style guide [here](https://webkit.org/code-style-guidelines/#braces-empty-block)). The addition to clang-format is described [here](https://reviews.llvm.org/D65925), and the new result is :

```c++
class MyClass {
public:
    ~MyClass() { }
}
```

Since [the coding style](https://github.com/SerenityOS/serenity/blob/master/Documentation/CodingStyle.md) just requires to use clang-format version 8 or greater without any additional restrictions, we get different results. It would be a good idea to standardize the usage.

The main issue is that the option is not available in older versions of clang, and if we add it to the .clang-format file, then the formatter version 8 and 9 errors out and refuses to run.

I think the options we have are to :

- Require everyone to only use clang-format 8 or 9. This requires an update of the documentation, and locks us in a version of clang which isn't the latest one.
- Require everyone to update to clang-format 10 or later and leave the option enabled (default value). This requires an update of the documentation and requires a codebase reformat to align everything, as well as requiring everyone to update their version.
- Require everyone to update to clang-format 10 or later and explicitely set the option to false in the .clang-format file. This requires an update of the documentation and the .clang-format file, as well as requiring everyone to update their version.
- Don't require a formatter, but change the .clang-format `BasedOnStyle` key to one that doesn't add a space by default (if one exists). This requires a lot of tweaking on the .clang-format file to find the same behaviour we had before.

Dexesttp:
List of changes:
- The required clang-format version has been updated to 10 as per #2213
- The cmake command has been moved away from the main build instructions since the BuildIt script executes it automatically
- The ninja install has been clarified to take the BuildIt script into account and explain that the folder needs to be cleaned before executing cmake
- `brew install bash` has been added to the macOS prerequisites (as per https://github.com/SerenityOS/serenity/pull/2132#issuecomment-628819467)
- The ports instructions have been updated to use the make commands
- The build instructions headers have been indented to group the prerequisites together
- The build instructions code snippets have been standardized into code blocks
- Fixed a typo

Closes #2213

### DIFF
--- a/Documentation/BuildInstructions.md
+++ b/Documentation/BuildInstructions.md
@@ -1,6 +1,8 @@
 ## SerenityOS build instructions
 
-### Linux prerequisites
+### Prerequisites
+
+#### Linux prerequisites
 Make sure you have all the dependencies installed:
 
 **Debian / Ubuntu**
@@ -30,7 +32,7 @@ sudo apt-get install gcc-9 g++-9
 sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-9 900 --slave /usr/bin/g++ g++ /usr/bin/g++-9
 ```
 
-### macOS prerequisites
+#### macOS prerequisites
 Make sure you have all the dependencies installed:
 ```bash
 brew install coreutils
@@ -43,6 +45,7 @@ brew install m4
 brew install autoconf
 brew install libtool
 brew install automake
+brew install bash
 brew cask install osxfuse
 Toolchain/BuildFuseExt2.sh
 ```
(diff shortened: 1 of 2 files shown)


Sobak:
Maybe it would be good to explain why do we install `bash` from brew? I know that from the linked comment but I somehow afraid that many people will think _"nah, I already have bash, I will simply skip this one"_.

PR #2227, opened 2020-05-15
LibCore: Make IODevice::can_read_line() const

### LINKED ISSUE #2219: Core::IODevice::can_read_line() should be const and make its cache mutable
Raised by @bugaevc in https://github.com/SerenityOS/serenity/pull/2218#discussion_r424984336

alimpfard:
such const-correctness, much wow

I'm going to leave the `{} -> { }` changes in (as per #2213), if we'd rather do that in a separate commit, I can drop those.

-> Fixes #2219


awesomekling:
Straight up pleasant change

PR #2232, opened 2020-05-15
Build: FreeBSD support

zaklaus:
This commit fixes **FreeBSD** support on the new CMake build system.

It changes the following:
* `CMakeLists.txt` was updated to take the GNU GCC branch so it can
  define `-Wno-expansion-to-defined` flag. It would originally
  take the **Clang** branch, which was incorrect.
* Updated tests to fix warnings: `-Wunused-private-field` and `-Wself-assign-overload` respectively.

Reason why build fails on FreeBSD is that when we generate Makefiles, CMake picks up our system's default compiler (clang) and applies rules designed for Clang. However, once we compile using our cross-compiler, these (clang) flags are then put in effect and break the build.

As a side note, I recommend installing `sysutils/coreutils` port as well as `bash` and `gmake` as these are required to build the project successfully.

You should also link `/usr/local/bin/ginstall` to `install` so that cmake can leverage this command properly.

### DIFF
--- a/AK/Tests/TestRefPtr.cpp
+++ b/AK/Tests/TestRefPtr.cpp
@@ -79,7 +79,16 @@ TEST_CASE(assign_copy_self)
 {
     RefPtr<Object> object = adopt(*new Object);
     EXPECT_EQ(object->ref_count(), 1);
-    object = object;
+
+    #ifdef __clang__
+    #pragma clang diagnostic push
+    #pragma clang diagnostic ignored "-Wself-assign-overloaded"
+    #endif
+        object = object;
+    #ifdef __clang__
+    #pragma clang diagnostic pop
+    #endif
+
     EXPECT_EQ(object->ref_count(), 1);
 }
 
--- a/AK/Tests/TestWeakPtr.cpp
+++ b/AK/Tests/TestWeakPtr.cpp
@@ -29,6 +29,11 @@
 #include <AK/Weakable.h>
 #include <AK/WeakPtr.h>
 
+#ifdef __clang__
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wunused-private-field"
+#endif
+
 class SimpleWeakable : public Weakable<SimpleWeakable> {
 public:
     SimpleWeakable() {}
@@ -36,6 +41,10 @@ public:
 private:
     int m_member { 123 };
 };
+
+#ifdef __clang__
+#pragma clang diagnostic pop
+#endif
 
 TEST_CASE(basic_weak)
 {
--- a/Documentation/BuildInstructions.md
+++ b/Documentation/BuildInstructions.md
@@ -65,6 +65,12 @@ Notes:
 pkg_add bash gmp gcc git flock gmake sudo
 ```
 
+#### FreeBSD prerequisites
+```
+$ pkg add coreutils gmake bash sudo git
+$ ln -s /usr/local/bin/ginstall /usr/local/bin/install
+```
+
 ### Build
 > Before starting, make sure that you have configured your global identity for git, or the first script will fail after running for a bit.
 


Services/WindowServer/ClientConnection.h
```diff
@@ -125,6 +125,8 @@ class ClientConnection final
     virtual OwnPtr<Messages::WindowServer::SetWindowBaseSizeAndSizeIncrementResponse> handle(const Messages::WindowServer::SetWindowBaseSizeAndSizeIncrement&) override;
     virtual void handle(const Messages::WindowServer::EnableDisplayLink&) override;
     virtual void handle(const Messages::WindowServer::DisableDisplayLink&) override;
+    virtual OwnPtr<Messages::WindowServer::GetClipboardContentsResponse> handle(const Messages::WindowServer::GetClipboardContents &) override;
+    virtual OwnPtr<Messages::WindowServer::SetClipboardContentsResponse> handle(const Messages::WindowServer::SetClipboardContents &) override;
```
awesomekling, line 129: These API's are no longer in `Messages::WindowServer`. You need to remove generated files from your source tree after the cmake changes. Something like `git clean -qdfx` should do the trick.

CMakeLists.txt
```diff
@@ -101,10 +101,10 @@ set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Os -Wno-sized-deallocation -fno-sized-d
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DDEBUG -DSANITIZE_PTRS")
 add_link_options(--sysroot ${CMAKE_BINARY_DIR}/Root)
 
-if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang")
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wconsumed")
-elseif ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
+if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU" OR "${CMAKE_SYSTEM_NAME}" STREQUAL "FreeBSD")
```
awesomekling, line 104: This seems weird. What is the `CMAKE_CXX_COMPILER_ID` on your system?

AK/Tests/TestRefPtr.cpp
```diff
@@ -79,7 +79,7 @@ TEST_CASE(assign_copy_self)
 {
     RefPtr<Object> object = adopt(*new Object);
     EXPECT_EQ(object->ref_count(), 1);
-    object = object;
+    object = static_cast<RefPtr<Object>>(object);
```
awesomekling, line 82: What was the warning here?

Services/WindowServer/ClientConnection.h
zaklaus: I see, got it, I will do so, thanks!

CMakeLists.txt
zaklaus: It would actually resolve to Clang on my system.

AK/Tests/TestRefPtr.cpp
zaklaus: I'd get `-Wself-assign-overloaded` here.

Dexesttp:
> As a side note, I recommend installing sysutils/coreutils port as well as bash and gmake as these are required to build the project successfully.
You should also link /usr/local/bin/ginstall to install so that cmake can leverage this command properly.

You might want to add a FreeBSD section in `Documentation/BuildInstructions.md` as part of this PR, and add these as prerequisites :)

zaklaus:
> > As a side note, I recommend installing sysutils/coreutils port as well as bash and gmake as these are required to build the project successfully.
> > You should also link /usr/local/bin/ginstall to install so that cmake can leverage this command properly.
> 
> You might want to add a FreeBSD section in `Documentation/BuildInstructions.md` as part of this PR, and add these as prerequisites :)

Will do, thanks!

linusg:
Build now failing on Travis. Also please squash your commits and write commit messages according to the [contributing guidelines](https://github.com/SerenityOS/serenity/blob/master/CONTRIBUTING.md#code-submission-policy) :)

zaklaus:
> Build now failing on Travis. Also please squash your commits and write commit messages according to the [contributing guidelines](https://github.com/SerenityOS/serenity/blob/master/CONTRIBUTING.md#code-submission-policy) :)

No worries, should've marked it as draft for now. I will definitely squash and improve the commit messages.

zaklaus:
It should work fine now, I'm a bit unsure on how to deal with `unused-private-field` here as Travis build would complain attribute `unused` does not exist. I'm sure there's a better way to deal with the WeakPtr test warning.

AK/Tests/TestWeakPtr.cpp
```diff
@@ -32,6 +32,7 @@
 class SimpleWeakable : public Weakable<SimpleWeakable> {
 public:
     SimpleWeakable() {}
+    inline void SetMember(int val) { m_member = val; }
```
awesomekling, line 35: Please adhere to the project coding style:
```suggestion
    void set_member(int value) { m_member = value; }
```

Also I think you can use `[[maybe_unused]]` on the member instead of this.

CMakeLists.txt
awesomekling: Something here doesn't add up. Your compiler is GCC but CMake identifies it as Clang?
Can anyone else with FreeBSD confirm this behavior?

AK/Tests/TestRefPtr.cpp
awesomekling: Can we temporarily disable the warning instead?
Something like:
```cpp
#ifdef __clang__
#    pragma clang diagnostic push
#    pragma clang diagnostic ignored "-Wself-assign-overloaded"
#endif
    object = object;
#ifdef __clang__
#    pragma clang diagnostic pop
#endif

File: AK/Tests/TestWeakPtr.cpp (hunk 4 above)
Comment (zaklaus) [reply AK/Tests/TestWeakPtr.cpp:35]: I will try! 👍

Comment (zaklaus) [reply AK/Tests/TestWeakPtr.cpp]: Shame, it didn't work, I will just drop set_member there for now, if that's ok.

Comment (awesomekling) [reply AK/Tests/TestWeakPtr.cpp]: Sure!

File: CMakeLists.txt (hunk 2 above)
Comment (zaklaus) [reply CMakeLists.txt:104]: Actually, checking again I do have clang, so the detection is correct. However, it seems to dislike `-Wconsumed` and requires `-Wno-expansion-to-defined` to compile, favoring the GNU path.

If anyone has an ability to test it out on FreeBSD, I'd be glad.

Comment (zaklaus) [reply CMakeLists.txt:104]: Okay, found out the reason. When CMake generates the project files, it detects my clang and correctly sets up the flags based on the rules.

However, since `-Wconsumed` is a flag served for clang and we compile the code using the `i686-pc-serenity-` cross compiler, it basically freaks out.

Since FreeBSD uses clang by default, it explains why I got into this situation at the first place. :)

**Comment by zaklaus** [comment]
Reason why build fails on FreeBSD is that when we generate Makefiles, CMake picks up our system's default compiler (clang) and applies rules designed for Clang. However, once we compile using our cross-compiler, these (clang) flags are then put in effect and break the build.

**Comment by awesomekling** [comment]
@bugaevc does this look reasonable to you?

**Comment by bugaevc** [comment]
Honestly, no. There should be no reason to hardcode FreeBSD in particular (any system can use Clang by default — think Darwin). And even then, these flags are only affecting what the cross compiler (which is always GCC!) sees, so that shouldn't be different between systems. In fact, this whole conditional statement should be [re]moved — we know for a fact that the compiler is GCC, why check it conditionally.

**Comment by zaklaus** [comment]
> Honestly, no. There should be no reason to hardcode FreeBSD in particular (any system can use Clang by default — think Darwin). And even then, these flags are only affecting what the cross compiler (which is always GCC!) sees, so that shouldn't be different between systems. In fact, this whole conditional statement should be [re]moved — we know for a fact that the compiler is GCC, why check it conditionally.

Actually I agree with this, since we only run it through our cross compiler, we don't need these conditional statements present here. I assume the reason we need them is because we still need to compile LibC using our host compiler, right? How would we approach this issue?

**Comment by ADKaster** [comment]
LibC and LibM are compiled with the cross-compiler. The only things that should be compiled with the host compiler should be the ones at the top of CMakeLists.txt, before the sysroot is unset

```
add_subdirectory(Meta/Lagom)
add_subdirectory(DevTools/IPCCompiler)
add_subdirectory(DevTools/FormCompiler)
add_subdirectory(Libraries/LibWeb/CodeGenerators)
add_subdirectory(AK/Tests)
```

If kept, the conditionals should probably be moved up before the host-built tools, libs, and tests are compiled?

bugaevc:
> LibC and LibM are compiled with the cross-compiler. The only things that should be compiled with the host compiler should be the ones at the top of CMakeLists.txt, before the sysroot is unset
> If kept, the conditionals should probably be moved up before the host-built tools, libs, and tests are compiled?

Yep

zaklaus:
Got it, I will try that out, hopefully that would conclude the changes.

awesomekling:
Okay with the -Wconsumed stuff removed in 76bcd284f9dddc93f975cccabae9dc4c040c5a53 this became a lot easier. Merging, thank you :)

PR #2239, opened 2020-05-15
Clipboard history

bugaevc:
image

In the future, we can also support deleting history items.


awesomekling:
Merged with copyright headers as acc107a44feadcaec5ecfc6fded97fb1a3421132 and f7ae66cee395db5d8fd16398071bc2917124633e :)

bugaevc:
Oof I was constantly reminding myself I need to add copyright headers and forgot it at the end 🙁

awesomekling:
Note that we fail the build on missing headers, which is why this PR was red :)

PR #2240, opened 2020-05-15
LibELF: Handle DT_SONAME dynamic entries

ADKaster:
Store the offset in the string table for the DT_SONAME entry. Now that
the build uses cmake, cmake is helpfully passing --Wl,-soname to the
linker for shared objects. This makes the LinkDemo run again.


awesomekling:
:+1:

PR #2241, opened 2020-05-15
Shell: Add job control; switch builtins arg parsing to ArgsParser, Use Core::EventLoop / Kernel: waitpid "improvements"

alimpfard:
This PR ultimately adds `jobs`, `fg` and `bg` to Shell, and patches Shell builtins to use ArgsParser.

A few additions and fixes are made to accommodate this:

- ArgsParser: Fixes to allow calling ::parse() multiple times in a process
- Kernel+LibC: Fixes to make `waitpid` usable for job control purposes (**this bit needs review**)
- Shell: All the refactorings to use `Core::EventLoop`


Kernel/Process.cpp
```diff
@@ -2486,8 +2486,9 @@ KResultOr<siginfo_t> Process::do_waitid(idtype_t idtype, int id, int options)
         return KResult(-EINVAL);
     }
 
-    if (Thread::current->block<Thread::WaitBlocker>(options, waitee_pid) != Thread::BlockResult::WokeNormally)
-        return KResult(-EINTR);
+    if (!(options & WNOHANG))
+        if (Thread::current->block<Thread::WaitBlocker>(options, waitee_pid) != Thread::BlockResult::WokeNormally)
+            return KResult(-EINTR);
```
awesomekling, line 2491: Seems like we should handle this in the `if (options & WNOHANG)` block above this code, no?
Also please wrap multi-line blocks in `{ }` :)

Libraries/LibC/sys/wait.cpp
```diff
@@ -72,6 +72,10 @@ pid_t waitpid(pid_t waitee, int* wstatus, int options)
         case CLD_STOPPED:
             *wstatus = siginfo.si_status << 8 | 0x7f;
             break;
+        case CLD_CONTINUED:
+            *wstatus = 0;
+            return 0; // return 0 if running
+            break;
```
awesomekling, line 78: No need for `break` after `return`.

Kernel/Process.cpp
```diff
@@ -2504,13 +2504,33 @@ KResultOr<siginfo_t> Process::do_waitid(idtype_t idtype, int id, int options)
...
-        ASSERT(waitee_thread->state() == Thread::State::Stopped);
+        ASSERT((options & WNOHANG) || waitee_thread->state() == Thread::State::Stopped);
         siginfo_t siginfo;
         memset(&siginfo, 0, sizeof(siginfo));
         siginfo.si_signo = SIGCHLD;
         siginfo.si_pid = waitee_process->pid();
         siginfo.si_uid = waitee_process->uid();
-        siginfo.si_code = CLD_STOPPED;
+
+        // FIXME: I _really_ don't know what this should actually do
```
awesomekling, line 2514: POSIX @ https://pubs.opengroup.org/onlinepubs/9699919799/functions/waitid.html
>The application shall ensure that the infop argument points to a siginfo_t structure. If waitid() returns because a child process was found that satisfied the conditions indicated by the arguments idtype and options, then the structure pointed to by infop shall be filled in by the system with the status of the process; the si_signo member shall be set equal to SIGCHLD. If waitid() returns because WNOHANG was specified and status is not available for any process specified by idtype and id, then the si_signo and si_pid members of the structure pointed to by infop shall be set to zero and the values of other members of the structure are unspecified.

Linux @ http://man7.org/linux/man-pages/man2/sigaction.2.html
>The si_code field inside the siginfo_t argument that is passed to a SA_SIGINFO signal handler is a value (not a bit mask) indicating why this signal was sent.  ...
>For a non-ptrace(2) event, the values that can appear in si_code are described in the remainder of this section.

So POSIX doesn't go into any detail about `si_code`, really. I think maybe we can skip this field until we know we actually need it for communicating something. No need to bend over backwards just to imitate some obscure part of Linux kernel behavior.

Kernel/Process.cpp
alimpfard: we should not-block the thread?

I could assign that to a variable, I guess?

Kernel/Process.cpp
alimpfard: Well that bit is trying to communicate whether the child is running, I guess it's fine to set it to whatever since POSIX doesn't really specify?

Kernel/Process.cpp
awesomekling: Hmm hold on.. This code is organized a bit weirdly.

The above block will return from the function if `WNOHANG` *except* if the waitee process is not dead. It looks like a happy accident that we fall through for stopped children and then go through the same checks again.

Can we organize this code so it reads a bit more clearly and you can easily tell what's going on? :)

alimpfard: Actually, come think of it, why is it even structured like this?
maybe we should just let `Thread::block()` happen, but exit immediately in WaitBlocker if WNOHANG?
(I'm trying to wrap my head around what would happen if we request P_ALL and WNOHANG)

awesomekling: That sounds like it could be a lot more elegant yeah. Great idea :)

Kernel/Process.cpp
alimpfard: I guess we can leave it like it until it breaks something or does something unexpected?
it seems to work fine for its current use-case

awesomekling: Either way is fine really. :)

alimpfard: Then let's just go with this
> too lazy to change it anway

Shell/main.cpp
```diff
@@ -191,33 +194,57 @@ static int sh_exit(int, const char**)
...
 }
 
 static int sh_unset(int argc, const char** argv)
 {
-    if (argc != 2) {
-        fprintf(stderr, "usage: unset variable\n");
+    Vector<const char*> vars;
+
+    Core::ArgsParser parser;
+    parser.add_positional_argument(vars, "list of variables", "variables", Core::ArgsParser::Required::Yes);
```
bugaevc, line 240: Here and below: the convention is to start the help text with capital letter.

Shell/main.cpp
```diff
@@ -379,23 +410,188 @@ static int sh_time(int argc, const char** argv)
...
+        case Basic:
+            printf("[%llu] %s %s\n", job.value.job_id(), status, job.value.cmd().characters());
+            break;
+        case OnlyPID:
+            printf("[%llu] %d %s %s\n", job.value.job_id(), pid, status, job.value.cmd().characters());
+            break;
+        case ListAll:
+            printf("[%llu] %d %d %s %s\n", job.value.job_id(), pid, job.value.pgid(), status, job.value.cmd().characters());
+            break;
+        }
```
bugaevc, line 467: Idea for the future: could we make the jobs in this list be links to open them in SystemManager, somehow?

Shell/main.cpp
```diff
@@ -847,6 +1060,35 @@ static ExitCodeOrContinuationRequest::ContinuationRequest is_complete(const Vect
...
+            return_value = WEXITSTATUS(wstatus);
+        g.jobs.remove(pid);
+    } else if (WIFSTOPPED(wstatus)) {
+        fprintf(stderr, "Shell: [%llu] %s(%d) %s\n", job_id, name.characters(), pid, strsignal(WSTOPSIG(wstatus)));
+    } else {
+        if (WIFSIGNALED(wstatus)) {
+            printf("Shell: [%llu] %s(%d) exited due to signal '%s'\n", job_id, name.characters(), pid, strsignal(WTERMSIG(wstatus)));
+        } else {
+            printf("Shell: [%llu] %s(%d) exited abnormally\n", job_id, name.characters(), pid);
+        }
```
bugaevc, line 1086: And here — again, for the future — instead of emitting boring text, emit some magic escape characters to let the terminal draw the card red and display the exit code / signal number elsewhere.

Shell/main.cpp
```diff
@@ -1487,6 +1714,13 @@ int main(int argc, char** argv)
         save_history();
     });
 
+    signal(SIGCHLD, [](int) {
+        auto child_pid = waitpid(-1, nullptr, WNOHANG);
+        if (child_pid != -1) {
+            g.jobs.remove(child_pid);
```
bugaevc, line 1720: Don't just remove it, we have to print the info ("Job foo terminated")

Shell/main.cpp
alimpfard: Could certainly be an interesting feature.
Perhaps a unique protocol for launchserver?

Shell/main.cpp
alimpfard: printing in signal handlers just...scares me

Shell/main.cpp
alimpfard: Noted.

Shell/main.cpp
bugaevc: Don't print from the signal handler, that's indeed scary. Just stash the info away for later. SystemServer uses a deferred invocation for this, dunno if the shell can do the same (does it even use `Core::EventLoop`?), but something like that:

https://github.com/SerenityOS/serenity/blob/cd298446328af0d71a74d31f750fb1326689a3fe/Services/SystemServer/main.cpp#L55-L62

bugaevc: (Although SystemServer does in fact `dbg()` from the handler, and it works just fine).

Shell/main.cpp
alimpfard: This needs a lot of thought and discussion, I am more in favour of doing this via RPC, but kling is also very right in that it should work for remote hosts as well
we could certainly add Serenity Terminal-specific escape sequences, but it just turns into another mess on top of more previous messes

Needless to say, I am very much willing to do this, once we settle on something.

Shell/main.cpp
alimpfard: I guess we change the shell loop to work with `Core::EventLoop`, and this stuff would be magically better

Kernel/Process.cpp
```diff
@@ -2483,13 +2458,32 @@ KResultOr<siginfo_t> Process::do_waitid(idtype_t idtype, int id, int options)
...
+        case Thread::State::Running:
+        case Thread::State::Runnable:
+        case Thread::State::Blocked:
+        case Thread::State::Dying:
+        case Thread::State::Queued:
+            siginfo.si_code = CLD_CONTINUED;
+            break;
+        case Thread::State::Dead:
+            siginfo.si_code = CLD_EXITED;
+            break;
```
bugaevc, line 2481: But we have just checked that it ain't dead above...

Kernel/Scheduler.cpp
```diff
@@ -274,7 +274,8 @@ bool Thread::WaitBlocker::should_unblock(Thread& thread, time_t, long)
         }
 
         bool wait_finished = ((m_wait_options & WEXITED) && child_exited)
-            || ((m_wait_options & WSTOPPED) && child_stopped);
+            || ((m_wait_options & WSTOPPED) && child_stopped)
+            || (m_wait_options & WNOHANG);
```
bugaevc, line 278: Why are we doing this again, instead of not blocking in the first place?

bugaevc:
Meta note: can we maybe get a saner abstraction for working with children & signals while using an event loop? Maybe `Core::EventLoop` itself should learn how to do this (perhap with some notifier-like helpers for children and signals that you would register). Doing this all by hand each time feels brittle and tedious.

Kernel/Scheduler.cpp
alimpfard: We need the child pid in case we pass it P_ALL

alimpfard:
> some notifier-like helpers for children and signals that you would register

I am not that familiar with the whole Notifier architecture; but the event loop itself handling children and signals look a lot like over-abstraction to me
I'm probably seeing it wrong.

bugaevc:
I'm actually thinking of the way kqueue supports `EVFILT_SIGNAL`, and Linux has `signalfd` — they're all supposed to be used for responding to signals when using an event loop the same way you respond to other events.

We don't have to add some special signal specifics to `select`/`poll` (or maybe it's a good idea), but it would be nice to be able to handle child/signal events the same way we do other events.

alimpfard:
Hmmm...I see.
as for switching to EventLoop...I can't get more than one event to be processed, so...perhaps more suited to another PR?

I think we should just fix glaring issues with this PR and mess with the architecture in another PR

Shell/main.cpp
alimpfard: This PR will do a very silly thing, and print from the signal handler...we should _really_ switch to Core::EventLoop asap

bugaevc:
Yea, I'm not blocking this PR on those ideas, just thinking out loud.

Shell/Execution.h
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2018-2020, Andreas Kling <kling@serenityos.org>
+ * Copyright (c) 2020, The SereniyOS developers.
```
alimpfard, line 2: Git is too darn smart for its own good...

Dexesttp: That typo tho ^^ You have the same one remaining in Shell.cpp / Shell.h fyi.

alimpfard: lol I made a script that auto-adds these...and typo'd the darn thing.
good job me.

alimpfard: Thanks for catching that.

alimpfard:
Oh ouch...I rebased b11c7ad away...

awesomekling:
I like the shell becoming event-driven. It would be nice if it could also idle in the event loop instead of idling by blocking on `read()` but that's more rearchitecting.

alimpfard:
Yeah, I was thinking if we could provide an opt-in event-driven API for LibLine and how that would work.
the shell _just_ might get even more event-driven, who knows

PR #2242, opened 2020-05-15
Meta: Remove libstdc++ as a dependency

ADKaster:
Changes included:
    
- Include <string.h> instead of <cstring> in one userspace program
- Add custom std::initializer_list implementation. The compiler will convert source code with initializer lists into this type, which must be defined externally. Why? I have no idea, but the committee probably had their reasons...
- Move operator new and delete from kmalloc.h to an implementation file. Files that included kmalloc.h were getting the inline definition, while files that did not were pulling their implementation from libstdc++. Odd. In the kernel, the only definition was coming from libstdc++.
- Add AK/AtomicRef.h for atomically modifying a remote variable (used in static var guards in a few lines :) )
- Disable AK/Demangler.h in serenity :'(. We'll need to implement __cxa_demangle manually.
- Implement __cxa_guard_acquire, __cxa_guard_release and __cxa_guard_abort that the compiler will insert calls for when you have a static block-scope variable with a non-trivial constructor. Phew this took a lot of staring at other implementations until I understood a non-broken algorithm.
- The new __cxa_guard_* methods are implemented in library "LibCxx" which currently masquerades as libstdc++. I couldn't find a way to get g++ to not link libstdc++ without also having to manually link libc.a, libgcc.a, and libm. If we prefer manually specifying that for all exeuctables and libs, then I can go back to including them in libc... If we used Clang, we could probably tell it that `-stdlib=c` and be done with it though.
- Ditch the multi-build of LibC and LibM. We don't need it anymore! Simplify the calculate hash script as well, it only needs to take the Patches directory into account for 'The Toolchain'.


awesomekling:
cc @bugaevc

awesomekling:
>Add custom std::initializer_list implementation. The compiler will convert source code with initializer lists into this type, which must be defined externally. Why? I have no idea, but the committee probably had their reasons...

Oh nice. I've been down this road before but never followed it to the end. Glad you did! The stuff where the compiler tentacles sneak into the `std` namespace are fricken weird. See also `std::nullptr_t`.

This is a cool PR :)

bugaevc:
Oh, this sounds amazing! We won't even have to do the ugly half-build during toolchain build, how cool is that!

The CMake changes look... understandable to me, although I thought I've previously achieved the same in some simpler way.

CMakeLists.txt
```diff
@@ -101,6 +101,11 @@ set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Os -Wno-sized-deallocation -fno-sized-d
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DDEBUG -DSANITIZE_PTRS")
 add_link_options(--sysroot ${CMAKE_BINARY_DIR}/Root)
 
+set(CMAKE_C_IMPLICIT_LINK_LIBRARIES "")
+set(CMAKE_CXX_IMPLICIT_LINK_LIBRARIES "")
+
+set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -nodefaultlibs -L${CMAKE_BINARY_DIR}/Root/usr/lib -lc -lm -lgcc")
```
ADKaster, line 107: I'd really like to not set nodefaultlibs, because 1) it's a hack and 2) I'll need to set the same for SHARED linker flags? It also makes it really hard to rebuild libc and libm... 

The google searches i found for clearing the implicit link libraries seem to suggest it should make the linker not look for -lstdc++, but from experimenting only setting nodefaultlibs seems to have caused that....

ADKaster:
cc @BenWiederhake on the cache changes

BenWiederhake:
> Looking at the changes... Once 2242 goes in then the only files that will need to be hashed outside of buidit.sh itself (and the md5 of gcc, binutils, options etc) is the Patches dir...

Sounds awesome! :D

(Btw, you don't need my permission to go ahead – but I do appreciate that you asked first! ^^)

awesomekling:
What's the state of this PR? Good to go from your end @ADKaster?
Any comments @bugaevc?

AK/InitializerList.h
```diff
@@ -0,0 +1,74 @@
...
+    typedef const E* const_iterator;
+
+    constexpr initializer_list() noexcept = default;
+    constexpr size_t size() const noexcept { return m_size; };
+    constexpr const E* begin() const noexcept { return m_begin; };
+    constexpr E* end() const noexcept { return m_begin + m_size; }
+
+private:
+    // Note: begin has to be first or the compiler gets very upset
+    E* m_begin = { nullptr };
```
bugaevc, line 51: `const`?

Libraries/LibCxx/CMakeLists.txt
```diff
@@ -0,0 +1,7 @@
+set(SOURCES
+    CxaGuard.cpp
+)
+
+# Shhh don't worry gcc, we are totally libstdc++
+serenity_libc(LibCxx stdc++)
+target_link_libraries(LibC)
```
bugaevc, line 7:
```suggestion
target_link_libraries(LibCxx LibC)
```
you mean?

CMakeLists.txt
```diff
@@ -101,7 +101,10 @@ set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Os -Wno-sized-deallocation -fno-sized-d
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DDEBUG -DSANITIZE_PTRS")
 add_link_options(--sysroot ${CMAKE_BINARY_DIR}/Root)
 
-include_directories(Libraries/LibC)
+include_directories(SYSTEM Libraries/LibC)
+include_directories(SYSTEM Libraries/LibM)
+include_directories(SYSTEM Libraries/LibPthread)
+
```
bugaevc, line 107: What do we need this for? `serenity_libc()` should already add the headers to the path of whoever depends on those libraries. Maybe change `serenity_libc()` to use `SYSTEM` instead?

ADKaster: Ah, i guess the idea was that when building an app cross-compiled or within serenity, the system headers in LibC, LibM, and LibPthread should be `-isystem`. But, it's true that if you're linking against these libs you'll get the headers, i overlooked that.

bugaevc: Yeah, including them with `-isystem` is probably a good idea (although it shouldn't matter for us). But maybe you can do that with the existing `target_include_directories()` command in `serenity_libc()`??

AK/InitializerList.h
ADKaster: Yep, here and a few other places in the file

Libraries/LibCxx/CMakeLists.txt
ADKaster: Oh yeah. I did mean that. Fixed

CMakeLists.txt
ADKaster: There's no `target_include_directories()` directive in the `serenity_libc()` function? Only found references to that in LibM/CMakeLists.txt and LibPthread/CmakeLists.txt, where it's marked PUBLIC.

I can change those to SYSTEM and add one to LibC/CMakeLists.txt? Haven't used cmake much before but that sounds like a reasonable change.

bugaevc: Right, I was thinking of `target_link_directories()`. Maybe add `target_include_directories()` there then and remove it from individual libcs?

ADKaster:
Assuming @bugaevc  is ok with my comment resolutions and cmake changes in https://github.com/SerenityOS/serenity/pull/2242/commits/fa7038d67b8c3da089dde9ed6a316d663346816e, this should be good to go!

CMakeLists.txt
ADKaster: See https://github.com/SerenityOS/serenity/pull/2242/commits/fa7038d67b8c3da089dde9ed6a316d663346816e. I may have gone a bit overboard...

bugaevc:
Yep, the CMake changes all look good to me. Now we only need to revert a part of https://github.com/SerenityOS/serenity/commit/5386508119e5f4d9837e1015a4a3f819da7082a7 that removed explicit CMake invocation from build instructions

awesomekling:
All right let's do this bois

awesomekling:
This broke the gcc port as its "libcpp" depends on the system having all the C++ headers available.

awesomekling:
I've reverted the changes to unbreak GCC. We'll need a clear path forward with that port before we can proceed with this :/

ADKaster:
Would it work to have libstdc++ as a port itself? Or to cross-compile GCC's libstdc++ explicitly in the gcc port script?

PR #2246, opened 2020-05-16
WindowServer: const correctness

shannonbooth:


Services/WindowServer/MenuItem.h
```diff
@@ -77,6 +77,7 @@ class MenuItem {
     void set_submenu_id(int submenu_id) { m_submenu_id = submenu_id; }
 
     Menu* submenu();
+    const Menu* submenu() const;
```
awesomekling, line 80: Where's the definition?

shannonbooth: brainfart, might go sleep now... suprised this builds and runs okay to be honest.

PR #2247, opened 2020-05-16
LaunchServer: Support multiple handlers for a single file type

niax:
This change makes the LaunchServer discover which apps can open which file types by reading the *.af files. Clients can get the list of appropriate handlers using the `GetHandlersForURL` API, and then pass back which handler they'd like to use to open the given URL in `OpenUrl`.

For now, the default behavior remains - not specifying a handler uses the existing user preferences. To avoid confusion between the two (preference maps were previously called handlers), I've changed the names of existing things.

I also updated LibVT to use the handler passing for OpenUrl - it was already set up to discover handlers.

### DIFF
--- a/Base/res/apps/TextEditor.af
+++ b/Base/res/apps/TextEditor.af
@@ -8,4 +8,4 @@ Category=Utilities
 32x32=/res/icons/32x32/app-texteditor.png
 
 [Launcher]
-FileTypes=txt,md,html
+FileTypes=txt,md,html,htm,js,json,ini
--- a/Base/res/apps/TextEditor.af
+++ b/Base/res/apps/TextEditor.af
@@ -6,3 +6,6 @@ Category=Utilities
 [Icons]
 16x16=/res/icons/TextEditor16.png
 32x32=/res/icons/32x32/app-texteditor.png
+
+[Launcher]
+FileTypes=txt,md,html
--- a/Libraries/LibDesktop/Launcher.h
+++ b/Libraries/LibDesktop/Launcher.h
@@ -32,7 +32,7 @@ namespace Desktop {
 
 class Launcher {
 public:
-    static bool open(const URL&);
+    static bool open(const URL&, const String& handler_name = {});
     static Vector<String> get_handlers_for_url(const URL&);
 };
 
--- a/Services/LaunchServer/Launcher.cpp
+++ b/Services/LaunchServer/Launcher.cpp
@@ -26,7 +26,9 @@
 
 #include "Launcher.h"
 #include <AK/FileSystemPath.h>
+#include <AK/Function.h>
 #include <LibCore/ConfigFile.h>
+#include <LibCore/DirIterator.h>
 #include <stdio.h>
 #include <sys/stat.h>
 
@@ -47,6 +49,33 @@ Launcher& Launcher::the()
     return *s_the;
 }
 
+void Launcher::load_handlers(const String& af_dir)
+{
+    auto load_hashtable = [](auto& af, auto& key) {
+        HashTable<String> table;
+
+        auto config_value = af->read_entry("Launcher", key, {});
+        for (auto& key : config_value.split(','))
+            table.set(key.to_lowercase());
+
+        return table;
+    };
+
+    Core::DirIterator dt(af_dir, Core::DirIterator::SkipDots);
+    while (dt.has_next()) {
+        auto af_name = dt.next_path();
+        auto af_path = String::format("%s/%s", af_dir.characters(), af_name.characters());
+        auto af = Core::ConfigFile::open(af_path);
+        if (!af->has_key("App", "Name") || !af->has_key("App", "Executable"))
+            continue;
+        auto app_name = af->read_entry("App", "Name");
+        auto app_executable = af->read_entry("App", "Executable");
+        auto file_types = load_hashtable(af, "FileTypes");
+        auto protocols = load_hashtable(af, "Protocols");
+        m_handlers.set(app_executable, { app_name, app_executable, file_types, protocols });
+    }
+}
+
 void Launcher::load_config(const Core::ConfigFile& cfg)
 {
     for (auto key : cfg.keys("FileType")) {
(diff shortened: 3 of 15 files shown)


Libraries/LibDesktop/Launcher.h
```diff
@@ -32,7 +32,7 @@ namespace Desktop {
 
 class Launcher {
 public:
-    static bool open(const URL&);
+    static bool open(const URL&, const Optional<String> handler_name={});
```
bugaevc, line 35: `String` is already implicitly optional, as it has the null value (the wonders of C++), so why not just use that?

Services/LaunchServer/Launcher.cpp
```diff
@@ -47,6 +49,33 @@ Launcher& Launcher::the()
     return *s_the;
 }
 
+void Launcher::load_handlers(const String af_dir)
```
bugaevc, line 52: `const String&` probably

Services/LaunchServer/Launcher.cpp
```diff
@@ -47,6 +49,33 @@ Launcher& Launcher::the()
...
+
+        return table;
+    };
+
+    Core::DirIterator dt(af_dir, Core::DirIterator::SkipDots);
+    while (dt.has_next()) {
+        auto af_name = dt.next_path();
+        auto af_path = String::format("%s/%s", af_dir.characters(), af_name.characters());
+        auto af = Core::ConfigFile::open(af_path);
+        if (!af->has_key("App", "Name") || !af->has_key("App", "Executable"))
```
bugaevc, line 69: What if opening the config file fails?

niax: Core::ConfigFile can't fail - https://github.com/SerenityOS/serenity/blob/cd298446328af0d71a74d31f750fb1326689a3fe/Libraries/LibCore/ConfigFile.h#L42

The condition for checking name & executable means we will just ignore the file.

Libraries/LibDesktop/Launcher.h
niax: Isn't C++ delightful! 🤮

Services/LaunchServer/Launcher.cpp
awesomekling: We should teach that thing how to fail btw :)

Base/res/apps/TextEditor.af
```diff
@@ -6,3 +6,6 @@ Category=Utilities
 [Icons]
 16x16=/res/icons/TextEditor16.png
 32x32=/res/icons/32x32/app-texteditor.png
+
+[Launcher]
+FileTypes=txt,md,html
```
awesomekling, line 11: Please include `js`, `json` and `ini` here as well :)

Libraries/LibDesktop/Launcher.h
```diff
@@ -32,7 +32,7 @@ namespace Desktop {
 
 class Launcher {
 public:
-    static bool open(const URL&);
+    static bool open(const URL&, const String handler_name={});
```
awesomekling, line 35: `const String&`
Also, please use `clang-format`.

Services/LaunchServer/Launcher.cpp
```diff
@@ -107,6 +156,29 @@ bool Launcher::open_with_handlers(const HashMap<String, String>& handlers, const
...
 
+Vector<String> Launcher::handlers_for(const String& key, HashMap<String, String> user_preference, Function<bool(Handler, const String&)> handler_matches)
+{
+    Vector<String> handlers;
+
+    auto user_preferred = user_preference.get(key);
+    if (user_preferred.has_value())
+        handlers.append(user_preferred.value());
+
+    for (auto handler : m_handlers) {
```
awesomekling, line 167: `auto&`

Services/LaunchServer/Launcher.cpp
```diff
@@ -47,6 +49,33 @@ Launcher& Launcher::the()
     return *s_the;
 }
 
+void Launcher::load_handlers(const String& af_dir)
+{
+    auto load_hashtable = [](auto af, auto key) {
+        HashTable<String> table;
+
+        auto config_value = af->read_entry("Launcher", key, {});
+        for (auto key : config_value.split(','))
```
awesomekling, line 58: `auto&`

Services/LaunchServer/Launcher.cpp
```diff
@@ -120,7 +192,10 @@ Vector<String> Launcher::handlers_for_path(const String& path)
         return { "/bin/FileManager" };
 
     auto extension = FileSystemPath(path).extension().to_lowercase();
-    return { m_file_handlers.get(extension).value_or(m_file_handlers.get("*").value_or({})) };
+
+    return handlers_for(extension, m_file_handlers, [](auto handler, auto key) {
```
awesomekling, line 196: `(auto& handler, auto& key)`

Services/LaunchServer/Launcher.h
```diff
@@ -27,26 +27,38 @@
...
-    bool open_url(const URL&);
+    bool open_url(const URL&, const String handler_name);
     Vector<String> handlers_for_url(const URL&);
 
 private:
+    HashMap<String, Handler> m_handlers;
     HashMap<String, String> m_protocol_handlers;
     HashMap<String, String> m_file_handlers;
 
+    Vector<String> handlers_for(const String& key, HashMap<String, String> user_preferences, Function<bool(Handler, const String&)> handler_matches);
```
awesomekling, line 58: Please pass parameters by reference to avoid deep copies:
```suggestion
    Vector<String> handlers_for(const String& key, const HashMap<String, String>& user_preferences, Function<bool(const Handler&, const String&)> handler_matches);
 ```

File: Services/LaunchServer/Launcher.h [hunk 10]
```diff
@@ -27,26 +27,38 @@
...
     Vector<String> handlers_for_path(const String&);
     bool open_file_url(const URL&);
-    bool open_with_handlers(const HashMap<String, String>& handlers, const String key, const String argument, const String default_program);
+    bool open_with_user_preferences(const HashMap<String, String>& user_preferences, const String key, const String argument, const String default_program);
+    bool open_with_handler_name(const URL&, const String handler_name);
```
awesomekling, line 62: const String&

Base/res/apps/TextEditor.af
Zenyiar: Wouldn't you also want cpp and h so itd open up in HackStudio? + HTM / the other html type file.

niax: I added these in, but for now it doesn't make any actual difference given the fallback handler defaults to TextEditor. Longer term, I think LaunchServer should become aware of the mapping between extension and MIME type, and then TextEditor can declare text/* so we don't have to keep adding extensions here.

HackStudio doesn't currently support launching with a given path through argv. When it does, it can declare which file types it wants to open.

Services/LaunchServer/Launcher.h
niax: I sprinkled some & around, I think in the right places. If you have a good reference on what's the standard such things that'd be great for a C++ noob like myself.

Issue #2250, opened 2020-05-16
Assertion Failed - Minecraft.net/en-us/


Zenyiar:
When vising https://www.minecraft.net/en-us/, a Assertion error caused browser to fail and close.

```
USERSPACE(31) ASSERTION FAILED: false
/home/zenyiarr/src/serenity/Libraries/LibJS/Parser.cpp:1165
[WindowServer(14:14)]: Terminating due to signal 6
[FinalizerTask(3:3)]: 0xdeadc0de  _ZN6Kernel9Scheduler10switch_nowEv +64
0xdeadc0de  _ZN6Kernel9Scheduler5yieldEv +85
0xdeadc0de  _ZN6Kernel6Thread30yield_without_holding_big_lockEv +36
0xdeadc0de  _ZN6Kernel7Process11do_killselfEi +191
0xdeadc0de  _ZN6Kernel7Process8sys$killEii +258
0xdeadc0de  syscall_handler +1149
0xdeadc0de  syscall_asm_entry +29
0x08112279  raise +22
0x080480bd  abort +13
0x080fd396  tolower +0
0x080be4d4  _ZN2JS6Parser19parse_for_statementEv +354
0x080be0a4  _ZN2JS6Parser15parse_statementEv +336
0x080be713  _ZN2JS6Parser13parse_programEv +155
0x08056ae1  _ZN3Web17HTMLScriptElement13inserted_intoERNS_4NodeE +709
0x08051cab  _ZN3Web8TreeNodeINS_4NodeEE12append_childEN2AK13NonnullRefPtrIS1_EE +185
0x08065927  _ZZN3WebL19parse_html_documentERKN2AK10StringViewERNS_8DocumentERNS_10ParentNodeEENKUlvE2_clEv +299
0x080661d8  _ZN3WebL19parse_html_documentERKN2AK10StringViewERNS_8DocumentERNS_10ParentNodeE +1785
0x0806673e  _ZN3Web19parse_html_documentERKN2AK10StringViewERKNS0_3URLERKNS0_6StringE +131
0x0805b518  _ZNK2AK8FunctionIFvRKNS_10ByteBufferERKNS_7HashMapINS_6StringES5_NS_27CaseInsensitiveStringTraitsEEEEE15CallableWrapperIZN3Web8HtmlView4loadERKNS_3URLEEUlT_RT0_E_E4callES3_S9_ +3424
[...]

PR #2253, opened 2020-05-16
AK: Fix URL's operator<<() and use it / Handle "protocol relative URLs" in URL::complete_url()

linusg:
The existing `operator<<()` for `AK::URL` wouldn't compile when using it - no idea why but moving it inside the namespace fixed it.

Also handle URLs starting with `//` properly, i.e. prepend the document URL's protocol.

Closes https://github.com/SerenityOS/serenity/issues/2250, where `//some.host/some/script.js` would be completed as `https://www.minecraft.net//some.host/some/script.js` and the resulting 404 HTML page fed to the parser making it blow up.

(now it chokes on the CSS parser but that's another story)

### DIFF
--- a/AK/URL.cpp
+++ b/AK/URL.cpp
@@ -293,6 +293,12 @@ URL URL::complete_url(const String& string) const
     if (url.is_valid())
         return url;
 
+    if (string.starts_with("//")) {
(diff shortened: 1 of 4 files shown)


Zenyiar:
:)

PR #2256, opened 2020-05-16
Base: Emoji stuff

linusg:
Add fire emoji and add a file containing all the emojis we support.

image


awesomekling:
I know I'm not the best artist but that was supposed to be the facepalm emoji 😢 🤦

linusg:
oh... now that's a facepalm on my end :facepalm:

but as you said it's hard to communicate on so few pixels. Will drop the commit :)

awesomekling:
That fire emoji is literally lit btw

awesomekling:
Feel free to improve the facepalm emoji if you have ideas :D

PR #2260, opened 2020-05-17
Shell: Treat builtin names as programs and suggest them

alimpfard:
cc @bugaevc 

I'm definitely glad I did that `ENUMERATE_SHELL_BUILTINS` thing :^)

### DIFF
--- a/Shell/Shell.cpp
+++ b/Shell/Shell.cpp
@@ -1373,6 +1373,10 @@ void Shell::cache_path()
         }
     }
 
+    // add shell builtins to the cache
+    for (const auto& builtin_name : builtin_names)
+        cached_path.append(escape_token(builtin_name));
+
     quick_sort(cached_path);
 }
 
--- a/Shell/Shell.h
+++ b/Shell/Shell.h
@@ -174,6 +174,14 @@ private:
 
 #undef __ENUMERATE_SHELL_BUILTIN
 
+    constexpr static const char* builtin_names[] = {
+#define __ENUMERATE_SHELL_BUILTIN(builtin) #builtin,
+
+        ENUMERATE_SHELL_BUILTINS()
+
+#undef __ENUMERATE_SHELL_BUILTIN
+    };
+
     ExitCodeOrContinuationRequest::ContinuationRequest m_should_continue { ExitCodeOrContinuationRequest::Nothing };
     StringBuilder m_complete_line_builder;
     bool m_should_break_current_command { false };


Shell/Shell.h
```diff
@@ -174,6 +174,26 @@ class Shell : public Core::Object {
 
 #undef __ENUMERATE_SHELL_BUILTIN
 
+    constexpr static const char* BuiltinNames[] = {
```
awesomekling, line 177: `builtin_names`

Shell/Shell.h
```diff
@@ -174,6 +174,26 @@ class Shell : public Core::Object {
...
+#define __ENUMERATE_SHELL_BUILTIN(builtin) #builtin,
+
+        ENUMERATE_SHELL_BUILTINS()
+
+#undef __ENUMERATE_SHELL_BUILTIN
+    };
+    constexpr static auto BuiltinCount = sizeof(BuiltinNames) / sizeof(char*);
```
bugaevc, line 184:
```suggestion
    constexpr static auto BuiltinCount = sizeof(BuiltinNames) / sizeof(const char*);
```

😀

Shell/Shell.cpp
```diff
@@ -1373,6 +1373,10 @@ void Shell::cache_path()
         }
     }
 
+    // add shell builtins to the cache
+    for (const auto& builtin_name : builtins())
+        cached_path.append(escape_token(builtin_name));
```
awesomekling, line 1378: I don't think you need the separate function for this, you can just iterate the array with range-for:

```cpp
for (auto* builtin_name : builtin_names)
    ...
```

Shell/Shell.h
```diff
@@ -174,6 +174,26 @@ class Shell : public Core::Object {
...
+
+    static Vector<StringView, BuiltinCount> builtins()
```
bugaevc, line 186: Why even use a vector (and not say an array) if the size is statically known?

Shell/Shell.h
alimpfard: but...why?
I guess...

Shell/Shell.h
alimpfard: I had taken to doing PascalCase for constexpr members...sure.

Shell/Shell.cpp
alimpfard: range-based for works with constexpr arrays? neat!

PR #2261, opened 2020-05-17
LibGemini: Implement rendering text/gemini documents to HTML

alimpfard:
This also sets Content-Type to whatever 'meta' contains on success, to allow the browser to pick up what the document contains.

### DIFF
--- a/Libraries/LibGemini/Document.cpp
+++ b/Libraries/LibGemini/Document.cpp
@@ -1,0 +1,112 @@
+/*
+ * Copyright (c) 2020, The SerenityOS developers.
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include <AK/NonnullRefPtr.h>
+#include <AK/String.h>
+#include <AK/StringBuilder.h>
+#include <AK/Vector.h>
+#include <LibGemini/Document.h>
+
+namespace Gemini {
+
+String Document::render_to_html() const
+{
+    StringBuilder html_builder;
+    html_builder.append("<!DOCTYPE html>\n<html>\n");
+    html_builder.append("<head>\n<title>");
+    html_builder.append(m_url.path());
+    html_builder.append("</title>\n</head>\n");
+    html_builder.append("<body>\n");
+    for (auto& line : m_lines) {
+        html_builder.append(line.render_to_html());
+    }
+    html_builder.append("</body>");
+    html_builder.append("</html>");
+    return html_builder.build();
+}
+
+NonnullRefPtr<Document> Document::parse(const StringView& lines, const URL& url)
+{
+    auto document = adopt(*new Document(url));
+    document->read_lines(lines);
+    return document;
+}
+
+void Document::read_lines(const StringView& source)
+{
+    auto close_list_if_needed = [&] {
+        if (m_inside_unordered_list) {
+            m_inside_unordered_list = false;
+            m_lines.append(make<Control>(Control::UnorderedListEnd));
+        }
+    };
+
+    for (auto& line : source.lines()) {
+        if (line.starts_with("```")) {
+            close_list_if_needed();
+
+            m_inside_preformatted_block = !m_inside_preformatted_block;
+            if (m_inside_preformatted_block) {
+                m_lines.append(make<Control>(Control::PreformattedStart));
+            } else {
+                m_lines.append(make<Control>(Control::PreformattedEnd));
+            }
+        }
+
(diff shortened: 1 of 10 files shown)


Libraries/LibGemini/Document.h
```diff
@@ -0,0 +1,144 @@
...
+    }
+
+    void read_lines(const StringView&);
+    NonnullOwnPtrVector<Line> m_lines;
+    URL m_url;
+    enum {
+        PreformattedOff,
+        PreformattedOn,
+    } m_state { PreformattedOff };
+    bool inside_unordered_list { false };
```
bugaevc, line 73:
```suggestion
    bool m_inside_unordered_list { false };
```

Libraries/LibGemini/Document.cpp
```diff
@@ -0,0 +1,108 @@
...
+NonnullRefPtr<Document> Document::parse(const StringView& lines, const URL& url)
+{
+    auto document = adopt(*new Document(url));
+    document->read_lines(lines);
+    return document;
+}
+
+void Document::read_lines(const StringView& lines)
+{
+    auto handle_list = [&] {
```
bugaevc, line 60: Should probably be called "close_list_if_needed()` or something like that?

Libraries/LibGemini/Document.cpp
```diff
@@ -0,0 +1,108 @@
...
+    for (auto& line : lines.lines()) {
+        if (line.starts_with("```")) {
+            m_state = m_state == PreformattedOn ? PreformattedOff : PreformattedOn;
+            if (m_state == PreformattedOn) {
+                m_lines.append(make<Control>(Control::PreformattedStart));
+            } else {
+                m_lines.append(make<Control>(Control::PreformattedEnd));
+            }
+
+            handle_list();
```
bugaevc, line 76: What do we want to do first, do we close the list or do we open/close the pre?

Libraries/LibGemini/Document.h
```diff
@@ -0,0 +1,147 @@
...
+
+protected:
+    String m_text;
+};
+
+class Document : public RefCounted<Document> {
+public:
+    String render_to_html() const;
+
+    static NonnullRefPtr<Document> parse(const StringView& lines, const URL&);
```
bugaevc, line 56: `StringView lines` sounds wrong to me, `lines` is what you get after calling `lines()` on it.

Libraries/LibGemini/Document.h
```diff
@@ -0,0 +1,147 @@
...
+    const URL& url() const { return m_url; };
+
+private:
+    explicit Document(const URL& url)
+        : m_url(url)
+    {
+    }
+
+    void read_lines(const StringView&);
+    NonnullOwnPtrVector<Line> m_lines;
```
bugaevc, line 67: Can I haz a blank line between these two

Libraries/LibGemini/Document.h
```diff
@@ -0,0 +1,147 @@
...
+    URL m_url;
+    enum {
+        PreformattedOff,
+        PreformattedOn,
+    } m_state { PreformattedOff };
```
bugaevc, line 72: I'd just keep it a boolean like below, `bool m_inside_preformatted_block { false };`

Libraries/LibGemini/Line.cpp
```diff
@@ -0,0 +1,150 @@
...
+
+#include <AK/StringBuilder.h>
+#include <LibGemini/Document.h>
+
+namespace Gemini {
+
+String Text::render_to_html() const
+{
+    StringBuilder builder;
+    builder.append(m_text);
```
bugaevc, line 35: Is the text getting HTML-escaped anywhere?

Libraries/LibGemini/Line.cpp
```diff
@@ -0,0 +1,150 @@
...
+Text::~Text()
+{
+}
+
+String Heading::render_to_html() const
+{
+    StringBuilder builder;
+    size_t level = 0;
+    while (m_text.length() > level && m_text[level] == '#')
+        ++level;
```
bugaevc, line 49: Surely we don't want to parse it at `render_to_html()` time?

Libraries/LibGemini/Line.cpp
```diff
@@ -0,0 +1,150 @@
...
+{
+}
+
+String UnorderedList::render_to_html() const
+{
+    // 1.3.5.4.2 "Advanced clients can take the space of the bullet symbol into account"
+    //           FIXME: _where_ is the space?
+    StringBuilder builder;
+    builder.append("<li>");
+    builder.append(m_text.substring_view(1, m_text.length() - 1));
```
bugaevc, line 66: Same here, what about escaping?

Libraries/LibGemini/Line.cpp
```diff
@@ -0,0 +1,150 @@
...
+        return "<pre>";
+    case Kind::UnorderedListStart:
+        return "<ul>";
+    case Kind::UnorderedListEnd:
+        return "</ul>";
+    }
+
+    dbg() << "Unknown control kind _" << m_kind << "_";
+    ASSERT_NOT_REACHED();
+    return "";
```
bugaevc, line 89: This goes into `default:` in the above switch.

Libraries/LibGemini/Line.cpp
```diff
@@ -0,0 +1,150 @@
...
+    if (space_offset.has_value()) {
+        url = url_string.substring_view(0, space_offset.value());
+        auto offset = space_offset.value();
+        while (offset < url_string.length() && (url_string[offset] == ' ' || url_string[offset] == '\t'))
+            ++offset;
+        m_name = url_string.substring_view(offset, url_string.length() - offset);
+    }
+    m_url = URL(url);
+    if (!m_url.is_valid()) {
+        // relative URL
```
bugaevc, line 113: I wouldn't rely on `URL::is_valid()` for telling relative vs absolute URLs...

Libraries/LibGemini/Document.cpp
alimpfard: sure.

Libraries/LibGemini/Document.cpp
alimpfard: hmmm...HTML logic says close list first
I wish there was a test-suite for this format that we could load

Libraries/LibGemini/Document.h
alimpfard: I'm gonna pull a @awesomekling and say,
does `data` sound ok?

Libraries/LibGemini/Line.cpp
alimpfard: I looked for that, do we have a html escape?

Libraries/LibGemini/Line.cpp
alimpfard: any alternatives?

Libraries/LibGemini/Document.h
bugaevc: Heh :smile: it does, though maybe `source` would be better.

Libraries/LibGemini/Line.cpp
bugaevc: I think LibMD implements its own. I'd say, move that out to ~LibHTML~ LibWeb and make it public API.

Libraries/LibGemini/Line.cpp
bugaevc: Maybe add a `URL::is_absolute()`, IDK :D

Libraries/LibGemini/Line.cpp
alimpfard: I guess `String::escape_html_entities` is what I'm looking for.

alimpfard: Minor issue with this, why on earth is it static and named like a member function?

bugaevc: Yeah, it sounds like it should be a member method.

Libraries/LibGemini/Line.cpp
alimpfard: That's quite an annoying thing to do, since it would effectively duplicate a lot of stuff that `URL::complete` is doing, e.g. `//blah.blah/blah` is absolute

alimpfard: It seems `complete_url` already handles this, so I'm just gonna drop this bit off

PR #2268, opened 2020-05-17
LibJS: Add hex and unicode escape sequences

mattco98:
Inspired by the recent terminal UTF-8 support, this PR adds support for the following:

```js
'\x55'      // => 'U'
'\u26a0'    // => '⚠'
'\u{1f41e}' // => '🐞'
```

All of the above also work within template literals. I was planning on also adding octal escape sequences, but it seems like they're deprecated in ES6 -- using them in strict mode generates a syntax error. Knowing that, I didn't think it was super important to add them in this PR (or, really at all).

### DIFF
--- a/Libraries/LibJS/Token.cpp
+++ b/Libraries/LibJS/Token.cpp
@@ -27,6 +27,7 @@
 #include "Token.h"
 #include <AK/Assertions.h>
 #include <AK/StringBuilder.h>
+#include <AK/Utf32View.h>
 #include <ctype.h>
 
 namespace JS {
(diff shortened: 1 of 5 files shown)


mattco98:
Oops, looks like there was a lot of works done with UTF32 while I was working on this. That'll make my life easier :)

Libraries/LibJS/Token.cpp
```diff
@@ -27,6 +27,7 @@
 #include "Token.h"
 #include <AK/Assertions.h>
 #include <AK/StringBuilder.h>
+#include <AK/Utf32View.h>
```
awesomekling, line 30: Hmm, how is this used here?

mattco98: It's used in `StringBuilder.append(Utf32View)`:

https://github.com/SerenityOS/serenity/blob/096fb691673d4e6c71173cb0950912b12f1f5b99/Libraries/LibJS/Token.cpp#L175-L177

awesomekling: Aha, got it! Harder to spot with just the `{ }` :)

PR #2269, opened 2020-05-17
Shell: Attach links to completed paths 

alimpfard:
This patch adds the concept of "anchored" styles, which are applied to a specific part of the line, and are tracked to always stay applied to that specific part.

Further, the main goal of the PR is to attach links to completed paths:
```
ls ./Ter<tab>
```
gets you a link to ./Terminal.ini, which you may edit without the link disappearing or changing.

As with all lunches, nothing is ever free, we lost some performance for relocating anchored styles, but we only pay for them when we use them :tm:

### DIFF
--- a/Libraries/LibLine/Editor.cpp
+++ b/Libraries/LibLine/Editor.cpp
@@ -89,6 +89,8 @@ void Editor::insert(const u32 cp)
     auto str = builder.build();
     m_pending_chars.append(str.characters(), str.length());
 
+    readjust_anchored_styles(m_cursor, ModificationKind::Insertion);
+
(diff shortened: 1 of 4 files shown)


awesomekling:
This is a little glitchy;
`ls m<tab>` causes the `s` in `ls` to become hyperlinked as well as the completed `myfile.txt` :]

awesomekling:
But it's also really neat :D

alimpfard:
> ls m<tab> causes the s in ls to become hyperlinked

:thinking: I'm pretty sure I made sure that doesn't happen...

PR #2270, opened 2020-05-17
LibJS: Make Value::to_number() failproof

### LINKED ISSUE #2267: LibJS: Value::to_number() on Type::Object can fail
`to_primitive()` can throw, returning an empty value, and `to_number()` on that empty value then asserts.

https://github.com/SerenityOS/serenity/blob/4ced12670453df331632c980e7479159ef5ddf83/Libraries/LibJS/Runtime/Value.cpp#L236-L237

https://github.com/SerenityOS/serenity/blob/4ced12670453df331632c980e7479159ef5ddf83/Libraries/LibJS/Runtime/Value.cpp#L208-L210

```
> +({ toString: () => { throw 1; } })
js: /home/linus/Dev/git/serenity/Libraries/LibJS/Runtime/Value.cpp:209: JS::Value JS::Value::to_number() const: Assertion `false' failed.
[1]    44378 abort (core dumped)  /home/linus/Dev/git/serenity/Build/Meta/Lagom/js
```

I'm not sure about the solution to this. The best would probably be to pass an `Interpreter&` to `to_number()` and then check `interpreter.exception()`, but that'll require a whole lot of changes - but might have to happen eventually:

https://github.com/SerenityOS/serenity/blob/4ced12670453df331632c980e7479159ef5ddf83/Libraries/LibJS/Runtime/Value.cpp#L233-L235

linusg:
This patch is unfortunately rather large and might make some things feel bloated, but it is necessary to fix a few flaws in LibJS, primarily blindly coercing values to numbers without exception checks - i.e.

```cpp
interpreter.argument(0).to_i32();  // can fail!!!
```

Some examples where the interpreter would actually crash:

```js
var o = { toString: () => { throw Error() } };
+o;
o - 1;
"foo".charAt(o);
"bar".repeat(o);
```

To fix this, we now have the following...

```cpp
to_double(Interpreter&)
to_i32()
to_i32(Interpreter&)
to_size_t()
to_size_t(Interpreter&)
```

...and a whole lot of exception checking.

There's intentionally no `to_double()`, use `as_double()` directly instead.

This way we still can use these convenient utility functions but don't need to check for exceptions if we are sure the value already is a number.

Fixes #2267.

---

**Should we maybe also rename the ones that aren't being passed the `Interpreter&` to `as_i32()` / `as_size_t()`?**

### DIFF
--- a/Libraries/LibJS/AST.cpp
+++ b/Libraries/LibJS/AST.cpp
@@ -941,11 +941,12 @@ Value UpdateExpression::execute(Interpreter& interpreter) const
     auto reference = m_argument->to_reference(interpreter);
     if (interpreter.exception())
         return {};
-
     auto old_value = reference.get(interpreter);
     if (interpreter.exception())
         return {};
-    old_value = old_value.to_number();
+    old_value = old_value.to_number(interpreter);
+    if (interpreter.exception())
+        return {};
 
     int op_result = 0;
     switch (m_op) {
--- a/Libraries/LibJS/AST.cpp
+++ b/Libraries/LibJS/AST.cpp
@@ -96,7 +96,7 @@ CallExpression::ThisAndCallee CallExpression::compute_this_and_callee(Interprete
         auto object_value = member_expression.object().execute(interpreter);
         if (interpreter.exception())
             return {};
-        auto* this_value = object_value.to_object(interpreter.heap());
+        auto* this_value = object_value.to_object(interpreter);
         if (interpreter.exception())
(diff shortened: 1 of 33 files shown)


awesomekling:
>Should we maybe also rename the ones that aren't being passed the Interpreter& to as_i32() / as_size_t()?

Yeah I think that's a good idea. The convention throughout the codebase is `to_*` for (potential) conversions and `as_*` for casting to something we already know the thing to be.

PR #2275, opened 2020-05-18
LibJS: Rename to_{i32,size_t}() to as_{i32,size_t}() for clarity

linusg:
As these parameter-less overloads don't change the value's type and just assume `Type::Number`, naming them `as_i32()` and `as_size_t()` is more appropriate.

---

Confirmed by @awesomekling in https://github.com/SerenityOS/serenity/pull/2270#issuecomment-630002105 and then he ninja-merged the PR before I could make amendments smh

### DIFF
--- a/Libraries/LibJS/Runtime/Value.cpp
+++ b/Libraries/LibJS/Runtime/Value.cpp
@@ -82,7 +82,7 @@ String Value::to_string_without_side_effects() const
             return as_double() < 0 ? "-Infinity" : "Infinity";
 
         if (is_integer())
-            return String::number(to_i32());
+            return String::number(as_i32());
         return String::format("%.4f", as_double());
     }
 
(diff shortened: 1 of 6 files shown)


Libraries/LibJS/Runtime/Value.cpp
```diff
@@ -242,33 +242,33 @@ Value Value::to_number(Interpreter& interpreter) const
...
 {
-    auto number = to_number(interpreter);
-    if (interpreter.exception())
-        return 0;
-    return number.as_double();
+    return static_cast<i32>(as_double());
 }
 
-i32 Value::to_i32() const
+size_t Value::as_size_t() const
```
awesomekling, line 250: Maybe this should assert that the value is non-negative instead of giving you 0?

linusg: Yep, that seems reasonable

awesomekling:
s m h

PR #2276, opened 2020-05-18
Ports: Make bash link again

linusg:
No idea why this was suddenly broken, but removing these duplicated declarations make it build to completion again.

Bash doesn't seem to like the recent shell changes though, it hangs a lot during execution now... e.g. the run-tests script won't actually run :(
Needs investigating.


awesomekling:
Weird. I wonder if this worked accidentally sometimes because it found ncurses headers somehow? Meh.

PR #2277, opened 2020-05-18
LibJS: Fix stupid mistakes

linusg:
That's gonna cost me a few GBP.


Libraries/LibJS/Runtime/Value.cpp
```diff
@@ -249,9 +249,9 @@ i32 Value::as_i32() const
 
 size_t Value::as_size_t() const
 {
-    ASSERT(as_double() >= 0);
     if (is_nan())
         return 0;
+    ASSERT(as_double() >= 0);
```
awesomekling, line 254: Well hold on now, maybe the `is_nan()` check should just go away instead?
This is `as_size_t()`, which implies we already know this is a valid `size_t`-like value (i.e a positive integer)

PR #2278, opened 2020-05-18
LibLine (also Shell,js): Handle unicode

alimpfard:
This allows inputting arbitrary codepoints to LibLine, at the expense of sprinkling Utf{8,32}View everywhere in the editor.

Also finally does _something_ with the inputs that were discarded in `vt_dsr`

pls merge before #2269


awesomekling:
Yep, pretty awkward indeed. But we can nicen this up as API's improve :)

PR #2279, opened 2020-05-18
Shell: Parse source as utf-8

alimpfard:
This fixes the issue with the parser reporting invalid token positioning information.


alimpfard:
Closing this as the issue it tried to solve was solved in a different (some might argue better) way in #2287.

PR #2281, opened 2020-05-18
LibLine: Handle <return>s in incomplete data correctly

alimpfard:
Previously, we would concatenate all the commands together:
```
> sleep 5
echo well
echo hello
echo friends

> echo wellecho helloecho friends
```

Thanks, @bugaevc

### DIFF
--- a/Libraries/LibLine/Editor.cpp
+++ b/Libraries/LibLine/Editor.cpp
@@ -222,14 +222,20 @@ String Editor::get_line(const String& prompt)
         }
 
         Utf8View input_view { StringView { m_incomplete_data.data(), valid_bytes } };
+        size_t consumed_codepoints = 0;
 
-        for (auto ch : input_view) {
-            if (ch == 0)
+        for (auto codepoint : input_view) {
+            if (m_finish)
+                break;
+
+            ++consumed_codepoints;
+
+            if (codepoint == 0)
                 continue;
 
             switch (m_state) {
             case InputState::ExpectBracket:
-                if (ch == '[') {
+                if (codepoint == '[') {
                     m_state = InputState::ExpectFinal;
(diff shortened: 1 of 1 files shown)


Libraries/LibLine/Editor.cpp
```diff
@@ -222,8 +222,14 @@ String Editor::get_line(const String& prompt)
         }
 
         Utf8View input_view { StringView { m_incomplete_data.data(), valid_bytes } };
+        size_t index_in_view = 0;
 
         for (auto ch : input_view) {
+            if (m_finish)
```
bugaevc, line 228: `ch` is actually `codepoint` isn't it?

Libraries/LibLine/Editor.cpp
```diff
@@ -222,8 +222,14 @@ String Editor::get_line(const String& prompt)
         }
 
         Utf8View input_view { StringView { m_incomplete_data.data(), valid_bytes } };
+        size_t index_in_view = 0;
```
bugaevc, line 225: And the index is — the index, counting in codepoints (not bytes)?

alimpfard: well yes, it's `index_in_view`, you're thinking of a better name?

bugaevc: `number_of_codepoints`? idk, I'm not familiar with LibLine internals, just bringing up that `Utf8View` doesn't actually support indexing, so "index in view" doesn't tell me much

alimpfard: does `consumed_codepoints` sound ok?

bugaevc: Yeah

PR #2287, opened 2020-05-18
LibLine: Default to resolving Spans as byte offsets

alimpfard:
This allows all the unicode processing to be internal to the line editor.

Supersedes #2279, and obsoletes it.


awesomekling:
@bugaevc

PR #2288, opened 2020-05-18
Less copying in Ext2FS

bugaevc:
Will probably eat your data though — yet appears to work in my brief testing.

cc @supercomputer7

### DIFF
--- a/Kernel/FileSystem/Ext2FileSystem.cpp
+++ b/Kernel/FileSystem/Ext2FileSystem.cpp
@@ -155,7 +155,7 @@ InodeIdentifier Ext2FS::root_inode() const
     return { fsid(), EXT2_ROOT_INO };
 }
 
-bool Ext2FS::read_block_containing_inode(unsigned inode, unsigned& block_index, unsigned& offset, u8* buffer) const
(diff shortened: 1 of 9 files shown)


Kernel/FileSystem/Ext2FileSystem.cpp
```diff
@@ -449,11 +451,12 @@ Vector<Ext2FS::BlockIndex> Ext2FS::block_list_for_inode_impl(const ext2_inode& e
     auto process_block_array = [&](unsigned array_block_index, auto&& callback) {
         if (include_block_list_blocks)
             callback(array_block_index);
-        auto array_block = ByteBuffer::create_uninitialized(block_size());
-        read_block(array_block_index, array_block.data());
+        unsigned count = min(blocks_remaining, entries_per_block);
+        size_t read_size = count * sizeof(const __u32*);
```
awesomekling, line 455: This calculation is wrong, it should use `sizeof(u32)` and not `sizeof(u32*)` :)

bugaevc: Good catch!

awesomekling:
Have tested and messed with this a bit and it seems to work! Great ideas, way less churn. :)

PR #2290, opened 2020-05-18
Browser: Support fullscreen view

linusg:
Fullscreen mode hides all widgets except the tab widget (but hides its tab bar), so effectively only the HTML view is shown. Switching/closing the active tab of course also works in fullscreen mode.

To make this possible, `set_bar_visible()`/`is_bar_visible()` was added to `GUI::TabWidget`, and it was pretty straightforward as all the event handling & rect calculation code already uses `bar_height()` - so we can just adjust that accordingly (as well as skipping painting the tab bar).

image

### DIFF
--- a/Applications/Browser/Tab.cpp
+++ b/Applications/Browser/Tab.cpp
@@ -39,6 +39,7 @@
 #include <LibGUI/Menu.h>
 #include <LibGUI/MenuBar.h>
 #include <LibGUI/StatusBar.h>
+#include <LibGUI/TabWidget.h>
 #include <LibGUI/TextBox.h>
 #include <LibGUI/ToolBar.h>
 #include <LibGUI/ToolBarContainer.h>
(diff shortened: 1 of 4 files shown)


bugaevc:
Oh, could you make it so that the tab bar hides automatically when there's only one tab?

linusg:
@bugaevc done :)

image

Applications/Browser/Tab.cpp
```diff
@@ -231,6 +232,19 @@ Tab::Tab()
         GUI::Application::the().quit();
     }));
 
+    auto& view_menu = m_menubar->add_menu("View");
+    view_menu.add_action(GUI::CommonActions::make_fullscreen_action(
+        [&](auto&) {
```
awesomekling, line 237: It's a dangerous habit to capture using `[&]` when creating a lambda that will persist (such as those we attach as callbacks to a `GUI::Action`), since it captures all stack variables in the surrounding context by reference, and these references will become invalid once we leave the `Tab` constructor!

Instead, just capture `[this]` which gives you access to everything you need anyway. :)

PR #2291, opened 2020-05-18
LibGUI/Browser: Right clicking a tab opens a context menu

FalseHonesty:
To continue making the Browser's tabs more interactive, this patch pops up a context menu when a tab is right clicked. Currently, the only options provided in this context menu are to reload and close the tab, but I'm sure more will come with time :)

### DIFF
--- a/Libraries/LibGUI/TabWidget.cpp
+++ b/Libraries/LibGUI/TabWidget.cpp
@@ -370,4 +370,19 @@ void TabWidget::keydown_event(KeyEvent& event)
     Widget::keydown_event(event);
 }
 
+void TabWidget::context_menu_event(ContextMenuEvent& context_menu_event)
+{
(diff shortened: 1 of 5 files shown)


Libraries/LibGUI/TabWidget.cpp
```diff
@@ -244,6 +244,13 @@ void TabWidget::mousedown_event(MouseEvent& event)
                 if (on_middle_click && widget)
                     on_middle_click(*widget);
             });
+        } else if (event.button() == MouseButton::Right) {
```
awesomekling, line 247: The system has a separate concept for context menu events (by overriding `context_menu_event()`.)
We should use that here too, instead of special-casing right-click in `mousedown_event()` :)

FalseHonesty: I didn't realize there was a specific event for that, handy!

awesomekling:
:star:

PR #2293, opened 2020-05-19
LibGUI/Browser: Right clicking a bookmark opens a context menu

FalseHonesty:
Another PR adding more interactivity to the browser, this time by giving the bookmark buttons a context menu that (currently) allows for opening in a new tab, and removing.

For future reference, should I have included this in #2291? They feel distinct enough to be separate, but I also do not wish to open an excessive amount of PRs. Thanks!


awesomekling:
Separate PR's is just fine!
Same comment as the other one though, let's express this in terms of context menus. :)

PR #2294, opened 2020-05-19
LibLine: Unify completion hooks and adapt its users

alimpfard:
LibLine should ultimately not care about what a "token" means in the
context of its user, so force the user to split the buffer itself.
This also allows the users to pick up contextual clues as well, since
they have to lex the line themselves.

This commit pacthes Shell and the JS repl to better handle completions,
so certain wrong behaviours are now corrected as well:
- JS repl can now complete "Object . getOw<tab>"
- Shell can now complete "echo | ca<tab>" and paths inside strings

@bugaevc, following our prior discussion, I realised how much I hated having two "on_complete" hooks, so here we are.

I am not sure if it is sensible to complete paths inside strings, ideas on that are welcome.

### DIFF
--- a/Libraries/LibLine/Editor.cpp
+++ b/Libraries/LibLine/Editor.cpp
@@ -38,7 +38,7 @@
 namespace Line {
 
 Editor::Editor(Configuration configuration)
-    : m_configuration(configuration)
+    : m_configuration(move(configuration))
 {
     m_always_refresh = configuration.refresh_behaviour == Configuration::RefreshBehaviour::Eager;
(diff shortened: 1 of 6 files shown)


Libraries/LibLine/Editor.cpp
```diff
@@ -1222,11 +1185,24 @@ Vector<size_t, 2> Editor::vt_dsr()
     return { x, y };
 }
 
-String Editor::line() const
+String Editor::line(size_t upto_index) const
```
bugaevc, line 1188: `up_to_index`

Issue #1576, opened 2020-04-02
Move data size pretty printing to libs


xTibor:
Pretty printing data sizes a common feature in apps, it's reimplemented over and over several times with minute differences in userspace (number of decimals, K/KB/KiB, etc.) This would benefit some deduplication and a consistent format throughout the system if possible. I don't know which lib would be the best place to introduce this feature, `AK`/`Core`?

https://github.com/SerenityOS/serenity/blob/cd1d369cdd86b4870d65bc9ad856e8a0850cb4fb/Applications/FileManager/DirectoryView.cpp#L41-L50
https://github.com/SerenityOS/serenity/blob/cd1d369cdd86b4870d65bc9ad856e8a0850cb4fb/Applications/SystemMonitor/main.cpp#L62-L71
https://github.com/SerenityOS/serenity/blob/cd1d369cdd86b4870d65bc9ad856e8a0850cb4fb/Userland/allocate.cpp#L103-L105
https://github.com/SerenityOS/serenity/blob/cd1d369cdd86b4870d65bc9ad856e8a0850cb4fb/Userland/df.cpp#L59-L68
https://github.com/SerenityOS/serenity/blob/cd1d369cdd86b4870d65bc9ad856e8a0850cb4fb/Userland/ifconfig.cpp#L41-L50
https://github.com/SerenityOS/serenity/blob/cd1d369cdd86b4870d65bc9ad856e8a0850cb4fb/Userland/ls.cpp#L203-L212

awesomekling:
I would say `AK` as this seems like something the kernel could use as well.

PR #2295, opened 2020-05-19
Userland/SystemMonitor: Use AK::human_readable_size()

shannonbooth:
Bitta issue tidy up

Fixes: #1576


shannonbooth:
Oops actually I missed one in allocate.cpp

AK/NumberFormat.h
```diff
@@ -34,6 +34,7 @@ namespace AK {
 static String number_string_with_one_decimal(float number, const char* suffix)
 {
     float decimals = number - (int)number;
+    // FIXME: Remove this hackery once printf() supports floats.
```
awesomekling, line 37: But `printf()` does support floats now, so let's just use that instead? :)

shannonbooth: woah this wasn't a thing last time I looked, very nice :)

awesomekling:
Have you tested this? It makes `ls -lh` look pretty silly

shannonbooth:
@awesomekling yeah I ran it on a few programs, and it looked okay to me. But yeah you're right it does look a bit silly saying "bytes" in `ls -lh`, hmm

PR #2297, opened 2020-05-19
LibTLS: Only try to flush data when needed

alimpfard:
This patchset drops the write notifier, and schedules writes only when necessary.
As a result, the CPU utilisation no longer spikes to the skies :^)

Though empirical tests :tm: (aka trying to load google) have shown that the loading process becomes a bit slower as we defer writes now.

### DIFF
--- a/Libraries/LibTLS/Socket.cpp
+++ b/Libraries/LibTLS/Socket.cpp
@@ -113,67 +113,15 @@ bool TLSv12::common_connect(const struct sockaddr* saddr, socklen_t length)
 
     Core::Socket::on_connected = [this] {
         Core::Socket::on_ready_to_read = [this] {
-            if (!Core::Socket::is_open() || !Core::Socket::is_connected() || Core::Socket::eof()) {
-                // an abrupt closure (the server is a jerk)
-                dbg() << "Socket not open, assuming abrupt closure";
-                m_context.connection_finished = true;
-            }
-            if (m_context.critical_error) {
-                dbg() << "READ CRITICAL ERROR " << m_context.critical_error << " :(";
-                if (on_tls_error)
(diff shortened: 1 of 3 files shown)


Libraries/LibTLS/Socket.cpp
```diff
@@ -184,6 +132,57 @@ bool TLSv12::common_connect(const struct sockaddr* saddr, socklen_t length)
...
+{
+#ifdef TLS_DEBUG
+    dbg() << "Flushing cached records: " << m_context.tls_buffer.size() << " established? " << is_established();
+#endif
+    m_has_scheduled_write_flush = false;
+    if (!check_connection_state(false))
+        return;
+    flush();
+
+    if (!is_established()) {
```
alimpfard, line 145: Unfortunately this little bit is needed to start the AppData transfer, I'll need to look a bit into what can be done about it.

awesomekling:
What's missing for us to avoid deferring writes unnecessarily?

alimpfard:
The only bit where we spam flushes is the period between sending ClientFinished and reading ServerFinished, which I pointed out above.

Otherwise, every single write is scheduled because it's needed.

alimpfard:
Oh, I was too sleepy to understand that apparently.
I don't think it's sensible to immediately attempt to send records, the overhead of the protocol is rather huge.
Personally, I'd rather the initiation be a bit slower, but get higher overall throughput.

Nothing is stopping us from immediately flushing packets, but the overhead is pretty large

awesomekling:
That makes sense :)

alimpfard:
...you beat my push darn it

awesomekling:
N I N J A

alimpfard:
...I'll make another PR
:facepalm:

PR #2299, opened 2020-05-19
In which S.B. wants to improve id(1) performance, but ends up rewriting most of stdio as well as awkwardly attempting to write a ringbuffer

bugaevc:


alimpfard:
did s.b. manage to improve id(1) performance?
or did you just rewrite stdio to look actually readable?

Libraries/LibC/stdio.cpp
```diff
@@ -285,87 +704,47 @@ int puts(const char* s)
...
-
-    ssize_t rc = read(stream->fd, ptr, nmemb * size);
-    if (rc < 0) {
-        stream->error = errno;
-        return 0;
-    }
-    if (rc == 0)
-        stream->eof = true;
-    nread += rc;
+    size_t nread = stream->read(reinterpret_cast<u8*>(ptr), size * nmemb);
```
awesomekling, line 719: This looks like a good place to check for multiplication overflow.

Libraries/LibC/stdio.cpp
```diff
@@ -285,87 +704,47 @@ int puts(const char* s)
...
     assert(stream);
-    auto* bytes = (const u8*)ptr;
-    ssize_t nwritten = 0;
-    for (size_t i = 0; i < (size * nmemb); ++i) {
-        int rc = fputc(bytes[i], stream);
-        if (rc == EOF)
-            break;
-        ++nwritten;
-    }
+    size_t nwritten = stream->write(reinterpret_cast<const u8*>(ptr), size * nmemb);
```
awesomekling, line 726: Multiplication overflow again :)

bugaevc:
> did s.b. manage to improve id(1) performance?

No idea :smile: but `strace id` does look a lot more reasonable than before.

alimpfard:
well, a win is a win.
at the very least, the code looks better :)

PR #2300, opened 2020-05-19
LibTLS: Flush some packets as soon as more packets are written

alimpfard:
As I was trying to push before ninja-merge,

This seems like a better compromise between throughput and latency, and it doesn't _really_ affect the performance, so let's just compromise.


alimpfard:
Kling-merge-prevention engaged lol

PR #2301, opened 2020-05-19
LibWeb: window.location spec compliance fixes

linusg:


awesomekling:
Thank you for fixing that up :D

PR #2302, opened 2020-05-20
LibGfx: Implement JPGLoader

devsh0:
- [x] Decode JFIFs
- [x] Add support for chroma-subsampling

Support for progressive JPEGs will be added in a future PR.


devsh0:
I am opening this PR for reviews. I'm still working on supporting progressive JPEGs and I intend to add the feature in another PR. It's taking a while coz I have been super caught up with a few things lately. Besides, I'm a little confused about the decoding procedure of progressive JPEGs (refining AC scans to be exact) at the moment. I will have to revisit parts of the spec. The decoder should be able to decode majority of JPEGs found on the web. Since we ignore interchange format metadata the decoder should be able to parse and decode EXIF JPEGs in addition to pure JFIFs.

Dexesttp:
Nice work, I'll try to do a full review !

For starters, and as more of a general remark, I don't think you ran "clang-format" on the files you changed :)

You can see [the Coding Style document](https://github.com/SerenityOS/serenity/blob/master/Documentation/CodingStyle.md) for more details, but basically you need to make sure to get clang-format v10+, otherwise there will be inconsistencies.

linusg:
Also don't include commented-out code (I found at least two lines) :)

devsh0:
@Dexesttp I did run clang-format. I guess the version I've installed is dated. Will fix that.
@linusg Thanks! I will remove the commented lines.

Dexesttp:
> @Dexesttp I did run clang-format. I guess the version I've installed is dated. Will fix that.

Hmm. That's unexpected, even an older version shouldn't have that result. Make sure you used the `-style=file` option while running it, and not just the default parameters ?

There's a lot of missing spaces between variable names and brackets (such as `size_t compressed_size{ 0 }` which should be `size_t compressed_size { 0 }`), and honestly I'm not sure where that could be coming from...

devsh0:
Ahh...I probably forgot about the `-style` option. Still, I think it's probably better to update it. Where to download it though? Do I have to get the all-in-one LLVM binary from here: https://releases.llvm.org/download.html#10.0.0

Dexesttp:
LLVM has an one-command install page and dedicated channels for their releases (https://apt.llvm.org/). I guess you can retrieve just clang-format-10 if you tune in to the xenial channel ? But the easiest way might still be to install LLVM as a whole, I haven't tried either.

Libraries/LibGfx/JPGLoader.cpp
```diff
@@ -0,0 +1,1057 @@
...
+#include <AK/Vector.h>
+#include <LibGfx/Bitmap.h>
+#include <LibGfx/JPGLoader.h>
+#include <Libraries/LibM/math.h>
+
+namespace Gfx {
+
+static bool load_jpg_impl(JPGLoadingContext&);
+RefPtr<Gfx::Bitmap> load_jpg(const StringView& path);
+RefPtr<Gfx::Bitmap> load_jpg_from_memory(const u8*, size_t);
```
Dexesttp, line 42: Are these methods leftover from a previous draft ?
I don't see `load_jpg_from_memory` used anywhere, and `load_jpg` seems defined but unused...

Libraries/LibGfx/JPGLoader.cpp
```diff
@@ -0,0 +1,1057 @@
...
+
+static const u8 zigzag_map[64] {
```
Dexesttp, line 44:
```suggestion
constexpr static u8 zigzag_map[64] {
```

Libraries/LibGfx/JPGLoader.cpp
```diff
@@ -0,0 +1,1057 @@
...
+     *
+     * In the function below, `hcursor` and `vcursor` denote the location of the block
+     * we're building in the macroblock matrix. `vfactor_i` and `hfactor_i` are cursors
+     * that iterate over the vertical and horizontal subsampling factors, respectively.
+     * When we finish one iteration of the innermost loop, we'll have the coefficients
+     * of one of the components of block at position `mb_index`. When the outermost loop
+     * finishes first iteration, we'll have all the luminance coefficients for all the
+     * macroblocks that share the chrominance data. Next two iterations (assuming that
+     * we are dealing with three components) will fill up the blocks with chroma data.
+     */
```
Dexesttp, line 125: I'm guessing `clang-format` didn't like that block comment, or just ignored it :)

There's very few multi-line block comments similar to this in the codebase, but back during [LibCrypto BigInts](https://github.com/SerenityOS/serenity/blob/master/Libraries/LibCrypto/BigInt/UnsignedBigInteger.cpp#L646), we ended up using the typedoc-like `/** <comment> */` syntax for function descriptions, and several `// <comment>`-style comments for big blocks in other places.

Libraries/LibGfx/JPGLoader.cpp
```diff
@@ -0,0 +1,1057 @@
...
+
+Optional<Vector<Macroblock>> decode_huffman_stream(JPGLoadingContext& context)
+{
+    Vector<Macroblock> macroblocks;
+    macroblocks.resize(context.mblock_meta.padded_total);
+
+    dbg() << "Image width: " << context.frame.width;
+    dbg() << "Image height: " << context.frame.height;
+    dbg() << "Macroblocks in a row: " << context.mblock_meta.hpadded_count;
+    dbg() << "Macroblocks in a column: " << context.mblock_meta.vpadded_count;
```
Dexesttp, line 224: This will be more useful when everything's ready for production, but placing your "normal behaviour" debugging in a `#ifdef JPG_DEBUG / #endif` block already helps separating it from error-path debugging.
You can just keep a `#define JPG_DEBUG` command at the top of the file while you need to see the output, it just makes it easier to disable down the line ^^

Libraries/LibGfx/JPGLoader.cpp
```diff
@@ -0,0 +1,1057 @@
...
+        return false;
+    }
+}
+
+static inline u16 read_endian_swapped_word(BufferStream& stream)
+{
+    u16 temp { 0 };
+    stream >> temp;
+    return (temp >> 8) | (temp << 8);
+}
```
Dexesttp, line 305: Since everything in the JPEG is big-endian (I think, at least according to T.81 Appendix B. section 1.1.1), renaming the method to `read_be_word` and enforcing reading a BE word via `stream >> temp0; stream >> temp1; return (((u16)temp0) << 8) | ((u16)temp1);` would be more appropriate.
I don't know if Serenity is supposed to be able to run on big-endian machines, but we might as well make sure that if it does then JPG still works ^^

Libraries/LibGfx/ImageDecoder.cpp
```diff
@@ -40,7 +41,9 @@ ImageDecoder::ImageDecoder(const u8* data, size_t size)
     if (m_plugin->sniff())
         return;
 
-    m_plugin = nullptr;
+    m_plugin = make<JPGImageDecoderPlugin>(data, size);
+    if (m_plugin->sniff())
+        return;
 }
```
Dexesttp, line 47: I think you should keep the "m_plugin = nullptr;" part, and make it so `sniff()` actually does stuff (see below).
```suggestion

        m_plugin = nullptr;
        return;
}
```

Libraries/LibGfx/JPGLoader.cpp
```diff
@@ -0,0 +1,1057 @@
...
+{
+    if (!m_context->bitmap)
+        return false;
+    return m_context->bitmap->set_nonvolatile();
+}
+
+bool JPGImageDecoderPlugin::sniff()
+{
+    return false;
+}
```
Dexesttp, line 1020: I think that code should be functional before going in branch.. Since JPEG files have a "header" of sorts, maybe reading the first marker and checking whether it's JPG_SOI should be enough ?
(apparently, checking a six-byte "FF D8 FF" sequence is also a possibility for JPG files. Since the third and fourth byte should also be a valid marker, then it should start with "FF".)
```suggestion
{
    return m_context->compressed_size > 3
        && m_context->compressed_data[0] == 0xFF
        && m_context->compressed_data[1] == 0xD8
        && m_context->compressed_data[2] == 0xFF;
}
```

Libraries/LibGfx/JPGLoader.cpp
```diff
@@ -0,0 +1,1057 @@
...
+    }
+    return true;
+}
+
+JPGImageDecoderPlugin::JPGImageDecoderPlugin(const u8* data, size_t size)
+{
+    m_context = make<JPGLoadingContext>();
+    m_context->compressed_data = data;
+    m_context->compressed_size = size;
+    m_context->huffman_stream.stream.ensure_capacity(500 * KB);
```
Dexesttp, line 972: 500 Kb of preallocated buffer seems a little... excessive. But I don't know enough about JPEG to know if it's a size that could be reached often.
Reading the code, though, it feels like it should never really get bigger than the file itself so, in that case, it could at least be sized with `min(500 * KB, size)` instead of always allocating a flat 500 Kb.
Is this something that could / should be more properly sized ?

Libraries/LibGfx/JPGLoader.cpp
devsh0: Yes. Removed.

Libraries/LibGfx/JPGLoader.cpp
devsh0: You're right. 500kb is too much. I'm not sure what I was thinking there.

Libraries/LibWeb/Loader/Resource.cpp
```diff
@@ -83,6 +83,8 @@ static String guess_mime_type_based_on_filename(const URL& url)
         return "image/png";
     if (url.path().ends_with(".gif"))
         return "image/gif";
+    if (url.path().ends_with(".jpg"))
```
LepkoQQ, line 86: should this also check for .jpeg ?

devsh0: That's probably a good idea. Personally I have never come across an image with .jpeg or .JPEG extension, but I believe there are servers and image manipulation programs allowing .jpeg in their workflow.

devsh0:
Hey there reviewers!

Is someone still reviewing this? I have a feeling that this PR has hit a race condition. Do you guys expect changes to be pushed? I believe I resolved all the issues mentioned (@Dexesttp @LepkoQQ ?). Support for progressive JPEGs will be added in the future. For PJPEGs, we need a way to allow `ImageDecoder` to show a low res copy of the original JPEG but also communicate somehow that it has to come back for refining scans. Alternatively, we can process all the refining scans at once and then tell the client that the image is ready to render. This does take away from the advantages of progressive JPEGs, but some older browsers have previously done that.

If you guys are not done reviewing, don't mind me. I'm just making sure nobody is confused. If you think it's ready to be merged, let me know so I can fix the conflicts :)

awesomekling:
"This branch has conflicts that must be resolved."
Also, your commits are full of back-and-forth whitespace changes, please clean that up :)

Dexesttp:
You should probably add `.jpg` as a supported file format for normal files (not just libweb). I think that's done in [`LibGfx/Bitmap.h`](https://github.com/asliturk/serenity/blob/master/Libraries/LibGfx/Bitmap.h#L36) (the enum was added recently but should handle everything).
Note that this requires the `RefPtr<Gfx::Bitmap> load_jpg(const StringView& path);` method in order to work, so you might want to implement them (I think I asked you to remove the declaration last time since you didn't define/use it, but turns out it's useful !). It should just be a simple wrapper around the context (see [the PBM loader](https://github.com/asliturk/serenity/blob/master/Libraries/LibGfx/PBMLoader.cpp#L313) for reference).

Apart from that, the code for the last state of the PR seemed good to me, so once the conflicts are solved that should be fine ! Sorry for not getting back to you about it, I wanted to deep-dive into the spec earlier this week but ended up not being able to do it.

devsh0:
@Dexesttp No need to apologize :) Too much disturbance in this part of the world too. I was just making sure I wasn't wrong about something.

Libraries/LibGfx/Bitmap.h
```diff
@@ -38,7 +38,8 @@
     __ENUMERATE_IMAGE_FORMAT(png, ".png") \
     __ENUMERATE_IMAGE_FORMAT(gif, ".gif") \
     __ENUMERATE_IMAGE_FORMAT(bmp, ".bmp") \
-    __ENUMERATE_IMAGE_FORMAT(ico, ".ico")
+    __ENUMERATE_IMAGE_FORMAT(ico, ".ico") \
+    __ENUMERATE_IMAGE_FORMAT(jpg, ".jpg")
```
Dexesttp, line 42: Since libweb matches both ".jpg" and ".jpeg", it would be a good idea to do it here too :
```suggestion
    __ENUMERATE_IMAGE_FORMAT(jpg, ".jpg") \
    __ENUMERATE_IMAGE_FORMAT(jpg, ".jpeg")
```

Libraries/LibGfx/JPGLoader.h
```diff
@@ -0,0 +1,206 @@
...
+#define JPG_EOI 0xFFD9
+#define JPG_RST 0XFFDD
+#define JPG_SOF0 0XFFC0
+#define JPG_SOI 0XFFD8
+#define JPG_SOS 0XFFDA
+#define JPG_COM 0xFFFE
+
+namespace Gfx {
+
+RefPtr<Gfx::Bitmap> load_jpg(const StringView& path);
```
Dexesttp, line 90: Both methods are defined in the .cpp file :)
```suggestion
RefPtr<Gfx::Bitmap> load_jpg(const StringView& path);
RefPtr<Gfx::Bitmap> load_jpg_from_memory(const u8* data, size_t length);
```

devsh0:
> If you are comfortable with git rebase and are okay with spending some time reorganizing the commits, they could be more cleanly split between `LibGfx: Add JPEG decoder` + `LibGfx: Support loading JPEG images` + `LibWeb: support JPEG images`, while focusing on which file goes where rather than incremental changes

I totally agree. I'm sorry for being ignorant about my commits. Over the years I've acquired some really bad habits when working with git. I'll try my best to improve and hopefully my next PR will not be a total mess :)

Dexesttp:
No worries, it doesn't really matter at all !

Also, in case you didn't get notified (I don't know if you do), #2611 created some more conflicts.

As soon as they're resolved, I think this PR can go ahead @awesomekling !

awesomekling:
Please move all the JPEG decoder internal structs etc into the cpp file, since they are not meant to be used outside of that .cpp file.
Other than that, I would just echo the same thing Dexesttp said, it's nice to have separate commits for things :)

devsh0:
@awesomekling I had to squash to get rid of the whitespacey commits. I will be more careful next time :)

awesomekling:
Let's do it :)

PR #2303, opened 2020-05-20
Applications/Calculator - It is now able to handle multiple operations

supex0fan:
The main goal of this rework was to allow the calculator to be able to handle more than one operation at a time. That goal was achieved at the cost of a few features that existed in the previous version of the calculator. These features were the toggle sign button (+/-), the percentage button, the inverse button and the square root button. The reason these features weren't implemented is because there aren't characters in the UTF8 character set that I could use to represent these operations in the lexer.


alimpfard:
> there aren't characters in the UTF8 character set that I could use to represent these operations in the lexer.

That sounds like you're trying to do it the wrong way around, the user already inputs tokens, why are you trying to lex it?
Perhaps store the user input as a token list and serialise when needed?

supex0fan:
Yes, that is another way to do it, and definitely worth considering as it allows for far more complex operations. It is also quite a simple change to make. 
You're also correct, I must have looked as a bogus UTF8 character set table, because I looked at another one and it is definitely much better than I thought. 
Thanks for the tips.

awesomekling:
Hello there and thank you for working on Serenity! :)

Unfortunately I don't think these changes are right for the Calculator app, as it's meant to model a basic physical desk calculator in the style of W2K's calc.exe.

If we wanted to build a more powerful expression/graphing/whatever calculator (like SpeedCrunch or "bc"), then I think that would do better as a separate application.

Also, please observe the contribution guidelines: https://github.com/SerenityOS/serenity/blob/master/CONTRIBUTING.md

Before doing an unexpected rework of existing functionality, please open an issue so we can discuss it first :)

supex0fan:
That reply confused me for a bit, until I picked up a non-graphing calculator and found out that they actually don't handle full expressions as I had always assumed they did. They handle them the way your calculator does.

Also, sorry for not opening an issue, this is my first time working on an open source project (no excuse for not reading the CONTRIBUTING info though).

And I just wanted to tell you that I love your videos.

Thanks.

PR #2305, opened 2020-05-20
 LibGUI: Add MessageBox question icon 

asliturk:
MessageBox


awesomekling:
Nice, this will be better than using Warning/Info when asking the user for something :)

PR #2312, opened 2020-05-20
LibHTTP+LibTLS: Handle silly chunked-encoding chunk sizes

alimpfard:
Apparently that's allowed and the RFC is just unclear about it.
Some servers seem to zero-pad the chunk size for whatever reason, and previously, we interpreted that as the last chunk.

This still does not fix loading github, as the CSS parser chokes on one of the stylesheets.
This sort of infinite loop seems to stem from us handing it something that has invalid data in it, so I am still quite suspicious of Transfer-Encoding's.

### DIFF
--- a/Libraries/LibHTTP/Job.cpp
+++ b/Libraries/LibHTTP/Job.cpp
@@ -163,13 +163,24 @@ void Job::on_socket_connected()
                     // read size
                     auto size_data = read_line(PAGE_SIZE);
                     auto size_lines = StringView { size_data.data(), size_data.size() }.lines();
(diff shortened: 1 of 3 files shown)


alimpfard:
Update: the CSS parser is choking on https://github.githubassets.com/assets/github-a8ed939bfc2953ff1e7c45e3f0ad39c4.css

awesomekling:
Leading zeroes huh? Thanks for tracking it down :)

alimpfard:
Leading zeros. of all things.
Once again, the web manages to confuse me, what with everything in it being so unintuitive and straight up random

PR #2313, opened 2020-05-20
Build: Add Dockerfile + instructions

Jongy:
Yesterday I set up the Serenity environment on another PC, so I decided it's better to write a Dockerfile once and make it available for everyone to use :) This really lowers the (already low) 
"entrance fee" for new curious developers.


bugaevc:
Docker should absolutely not be the primary or recommended way, or even the first way to be mentioned. That should always be the native build.

Jongy:
> That should always be the native build.

Why is that? As long as it works either way, I'd say it is a matter of taste/choice.

bugaevc:
No, because one is the right way to do it and another is a terrible hack :slightly_smiling_face:

I don't know how @awesomekling feels about Docker, but my vote is on mentioning it, but not recommending or promoting it in any way.

Jongy:
Well IMO it's just the easiest way to get a reproducible build environment across different distros. Not trying to promote it, just suggesting an option. If that's bothering then it can be pushed after the native approach.

bugaevc:
> the easiest way to get a reproducible build environment across different distros

That's one of the issues with the Docker mindset: it encourages users to sidestep the system / distro and the integration work they do, instead bringing your own copy of everything in the name of reproducibility and uniformity "across distros" (except you aren't actually getting any — you're just making everyone use the exact same distro set up in exactly one way, on top of what they're already using). At least in Serenity itself, we strongly favor system integration; and I feel the same way about Linux distros.

awesomekling:
> This really lowers the (already low) "entrance fee" for new curious developers.

The "entrance fee" is there on purpose. We're not looking for the kind of bug reports and attention that one-click users bring to a project.

We can add your docker script to the repo if it makes development work easier for you, but it's not appropriate for the build instructions. :)

Jongy:
NP, I'll do the change later today. Should I remove Docker completely from the build instructions or just place it after the native approaches?

awesomekling:
Please remove it entirely from instructions. :)

nooga:
Please excuse me if this is out of line but I'm curious about this. 

>  it encourages users to sidestep the system / distro and the integration work they do
Why would I want to integrate some custom toolchain and foreign OS (pardon the term) noise with the system I'm running? 

Background: I wanted to hack on `LibJS` so I'm trying to build Serenity since yesterday on MacOS and it fails every time even when following instructions to the letter (which is fine!).

I basically have three choices now:
1. Get a nice thinkpad with my favourite linux distro, then proceed to tinker with Serenity there.
2. Run my favourite linux distro in a relatively heavy VM under MacOS and then proceed to tinker with Serenity while running it in another VM :^)
3. Fix the toolchain build on MacOS

Docker runs in a linux VM (hyperkit) on MacOS and having the entire pre-built Serenity dev environment in a container would make no sense at all. 

What makes sense is this:
- keeping Serenity src on host, (macos)
- building the toolchain inside a container (linux), 
- building Serenity inside that same container (linux),
- running the resulting image on host's qemu (macos)

This keeps the toolchain + required deps off my MacOS environment and makes the toolchain build a breeze, and... well, it's a form of integration right?

I struggle to see why Docker so bad of an idea to be kept out of the docs.

Jongy:
Well @nooga your arguments are more concrete, I'm just lazy, trying to keep dev environments of different projects separate so there's less chances I run into dependency mismatches.

Your comment encourages me haha. So besides from being much convenient to *some* of the users, it's also an enabler for others...

And about the "entrance fee" - I know Serenity is developed by its developers, for its developers, and for no-one else. But still, having a low entrance fee as possible for *any and all* OSS project is appreciated IMO.

Anyway @awesomekling I'll be waiting for your final say on these latest comments, I'm fine with either choice.

awesomekling:
I understand that you have reasons for using this kind of software.

However, this project chooses not to cater to developers who aren't already highly autonomous and able to set up a build environment for themselves. Docker and other turnkey solutions invite unwanted attention and we're simply opting out of that. :)

Jongy:
Okay, I removed all Docker references from the documentation.

bugaevc:
@nooga

> I basically have three choices now:
> 1. Get a nice thinkpad with my favourite linux distro, then proceed to tinker with Serenity there.
> 2. Run my favourite linux distro in a relatively heavy VM under MacOS and then proceed to tinker with Serenity while running it in another VM :^)
> 3. Fix the toolchain build on MacOS

The right way, of course, is the third way: if you're macOS user by choice (i.e. if you prefer it over Linux, and don't want to switch to Linux), you should figure out building Serenity under Darwin.

PR #2314, opened 2020-05-20
LibWeb: Canvas things

linusg:
- LibWeb: Let `HTMLCanvasElement.getContext()` return null for unknown types

  > https://developer.mozilla.org/en-US/docs/Web/API/HTMLCanvasElement/getContext
  >
  > The `HTMLCanvasElement.getContext()` method returns a drawing context on the canvas, or null if the context identifier is not supported.

- LibWeb: Add `CanvasRenderingContext2D.canvas`

### DIFF
--- a/Libraries/LibWeb/Bindings/CanvasRenderingContext2DWrapper.cpp
+++ b/Libraries/LibWeb/Bindings/CanvasRenderingContext2DWrapper.cpp
@@ -35,6 +35,7 @@
 #include <LibWeb/Bindings/HTMLImageElementWrapper.h>
 #include <LibWeb/Bindings/ImageDataWrapper.h>
 #include <LibWeb/DOM/CanvasRenderingContext2D.h>
+#include <LibWeb/DOM/HTMLCanvasElement.h>
 #include <LibWeb/DOM/HTMLImageElement.h>
 #include <LibWeb/DOM/ImageData.h>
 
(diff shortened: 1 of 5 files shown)


Libraries/LibWeb/Bindings/CanvasRenderingContext2DWrapper.cpp
```diff
@@ -397,5 +397,16 @@ JS::Value CanvasRenderingContext2DWrapper::put_image_data(JS::Interpreter& inter
...
 
+JS::Value CanvasRenderingContext2DWrapper::canvas_getter(JS::Interpreter& interpreter)
+{
+    auto* impl = impl_from(interpreter);
+    if (!impl)
+        return {};
+    auto* element = impl->element();
+    if (!element)
+        return JS::js_null();
+    return wrap(interpreter.heap(), static_cast<Element&>(*element));
```
awesomekling, line 408: Why the cast to `Element` here? It should be fine to pass a `HTMLCanvasElement` to `wrap()`, no? :)

linusg: hah indeed - I took that from `getElementById`

https://github.com/SerenityOS/serenity/blob/1a1394f7a2f17bb0631cc4aaaa9c78f0cf700312/Libraries/LibWeb/Bindings/DocumentWrapper.cpp#L87

PR #2315, opened 2020-05-21
LibLine: Correctly track the completion start and end

alimpfard:
> This is a little glitchy;
> `ls m<tab>` causes the `s` in `ls` to become hyperlinked as well as the completed `myfile.txt` :]

We can't have that :^)

### DIFF
--- a/Libraries/LibLine/Editor.h
+++ b/Libraries/LibLine/Editor.h
@@ -169,9 +169,17 @@ public:
     void stylize(const Span&, const Style&);
     void strip_styles(bool strip_anchored = false);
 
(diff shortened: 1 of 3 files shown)


Libraries/LibLine/Editor.h
```diff
@@ -3,7 +3,7 @@
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
- * modification, are permitted provided that the following conditions are met:
+ *  = 0modification, are permitted provided that the following conditions are met:
```
alimpfard, line 6: lol oops

PR #2316, opened 2020-05-21
LibJS: Add getter/setter support

mattco98:
This patch adds a `GetterSetterPair` object. Values can now store pointers to objects of this type. These objects are created when using `Object.defineProperty` and providing an accessor descriptor.

This allows for property getter and setter support. For example:

```js
Object.defineProperty(o, 'foo', { get() { return 5; } });
console.log(o.foo);
```
Object getter and setter shorthand will be coming in a future PR :)

### DIFF
--- a/Libraries/LibJS/Runtime/Object.cpp
+++ b/Libraries/LibJS/Runtime/Object.cpp
@@ -27,6 +27,7 @@
 #include <AK/String.h>
 #include <LibJS/Heap/Heap.h>
 #include <LibJS/Interpreter.h>
+#include <LibJS/Runtime/Accessor.h>
 #include <LibJS/Runtime/Array.h>
 #include <LibJS/Runtime/Error.h>
 #include <LibJS/Runtime/GlobalObject.h>
@@ -96,6 +97,9 @@ Value Object::get_own_property(const Object& this_object, const FlyString& prope
 
     auto value_here = m_storage[metadata.value().offset];
     ASSERT(!value_here.is_empty());
+    if (value_here.is_accessor()) {
+        return value_here.as_accessor().call_getter(Value(const_cast<Object*>(this)));
+    }
     if (value_here.is_object() && value_here.as_object().is_native_property()) {
         auto& native_property = static_cast<const NativeProperty&>(value_here.as_object());
         auto& interpreter = const_cast<Object*>(this)->interpreter();
@@ -180,10 +184,16 @@ Value Object::get_own_property_descriptor(const FlyString& property_name) const
     if (interpreter().exception())
         return {};
     auto* descriptor = Object::create_empty(interpreter(), interpreter().global_object());
-    descriptor->put("value", value.value_or(js_undefined()));
-    descriptor->put("writable", Value(!!(metadata.value().attributes & Attribute::Writable)));
-    descriptor->put("enumerable", Value(!!(metadata.value().attributes & Attribute::Enumerable)));
-    descriptor->put("configurable", Value(!!(metadata.value().attributes & Attribute::Configurable)));
+    descriptor->put("enumerable", Value((metadata.value().attributes & Attribute::Enumerable) != 0));
+    descriptor->put("configurable", Value((metadata.value().attributes & Attribute::Configurable) != 0));
+    if (value.is_accessor()) {
+        auto& pair = value.as_accessor();
+        descriptor->put("get", pair.getter());
+        descriptor->put("set", pair.setter());
+    } else {
+        descriptor->put("value", value.value_or(js_undefined()));
+        descriptor->put("writable", Value((metadata.value().attributes & Attribute::Writable) != 0));
+    }
     return descriptor;
 }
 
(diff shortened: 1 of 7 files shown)


awesomekling:
This bloats up the size of every object property value 3x, going from 12 bytes per value to 36 bytes. That's way too much just to accomodate getters and setters.

How about this: Since getter/setter pairs are a rare case, let's instead store them in a subclass of `JS::Cell`, e.g `JS::GetterAndSetter`. That way we can store it in a plain `JS::Value` and we pay 1 cell per getter/setter pair instead of paying 24 bytes per property value. :)

mattco98:
Oh, after some reading, it seems that I didn't realize how C++ unions worked. I thought that the size of the class would either be 13 bytes (common case) or 25 bytes (getter/setter), but the size of the union is always at least the size of it's largest member, which I didn't know. Will change!

Libraries/LibJS/Runtime/GetterSetterPair.h
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    GetterSetterPair(Value getter, Value setter)
+        : m_getter(getter)
+        , m_setter(setter)
+    {
+    }
+
+    Value& getter() { return m_getter; }
+    Value& setter() { return m_setter; }
```
awesomekling, line 48: I think these should just return `Value`, or is there some reason for the outside to have mutable access to the members?

Libraries/LibJS/Runtime/GetterSetterPair.h
```diff
@@ -0,0 +1,78 @@
...
+    void call_setter(Value this_object, Value setter_value)
+    {
+        if (setter().is_undefined())
+            return;
+        MarkedValueList arguments(interpreter().heap());
+        arguments.values().append(setter_value);
+        interpreter().call(setter().as_function(), this_object, move(arguments));
+    }
+
+    void visit_children(Cell::Visitor& visitor) override {
```
awesomekling, line 66: Please use `clang-format`.

Libraries/LibJS/Runtime/Object.cpp
```diff
@@ -195,19 +205,60 @@ void Object::set_shape(Shape& new_shape)
...
+    u8 attributes = configurable | enumerable;
+
+    if (is_accessor_property) {
+        if (descriptor.has_property("value") || descriptor.has_property("writable")) {
+            if (throw_exceptions)
+                interpreter().throw_exception<TypeError>("Accessor property descriptors cannot specify a value or writable key");
+            return false;
+        }
+
+        auto getter = descriptor.get("get").value_or(js_undefined());
```
awesomekling, line 222: Missing an exception check after the first `get()` here.

Libraries/LibJS/Runtime/GetterSetterPair.h
```diff
@@ -0,0 +1,78 @@
...
+        , m_setter(setter)
+    {
+    }
+
+    Value& getter() { return m_getter; }
+    Value& setter() { return m_setter; }
+
+    Value call_getter(Value this_object)
+    {
+        if (getter().is_undefined())
```
awesomekling, line 52: Instead of storing `undefined` in `m_getter` and `m_setter`, let's just store empty values.
That allows us to distinguish between empty (engine-only) and `undefined` (producible by JS.)

Libraries/LibJS/Runtime/GetterSetterPair.h
```diff
@@ -0,0 +1,78 @@
...
+ */
+
+#pragma once
+
+#include <LibJS/Runtime/Cell.h>
+#include <LibJS/Runtime/Value.h>
+
+namespace JS {
+
+class GetterSetterPair : public Cell {
```
awesomekling, line 34: This class has no subclasses so let's make it `final`.
Also maybe `Accessor` would be a better name, what do you think? Seems closer to the spec.
I know I kept saying "getter/setter pair", sorry about that :D

Libraries/LibJS/Runtime/Object.cpp
```diff
@@ -228,33 +279,37 @@ bool Object::put_own_property(Object& this_object, const FlyString& property_nam
...
+        }
     }
 
+    // FIXME: Transitions from accessor to data, or vice versa
     if (mode == PutOwnPropertyMode::DefineProperty && attributes != metadata.value().attributes) {
         if (m_shape->is_unique()) {
             m_shape->reconfigure_property_in_unique_shape(property_name, attributes);
         } else {
             set_shape(*m_shape->create_configure_transition(property_name, attributes));
         }
```
mattco98, line 298: I wasn't quite sure how to enforce a shape transition here, where an object property is changing from a data descriptor to an accessor descriptor or vice versa (because the attributes don't necessary have to change). In the first iteration of the PR, I had Shapes and TransitionKeys take a `bool is_data_property`, which I could then change in the `create_configure_transition` call to make a transition. Or to save space, I could put it on the `u8 attributes`. I can re-implement that behavior if that's the best way.

awesomekling: Hmm. I think the most straightforward way to do this would be adding two (engine-internal) attributes: something like `HasGet` and `HasSet`. Then we can just piggyback on the existing "configure" transitions. Does that make sense to you?

Libraries/LibJS/Runtime/Value.cpp
```diff
@@ -63,6 +64,18 @@ Function& Value::as_function()
     return static_cast<Function&>(as_object());
 }
 
+GetterSetterPair& Value::as_getter_setter_pair()
+{
+        ASSERT(is_getter_setter_pair());
```
awesomekling, line 69: Strange amounts of whitespace here.

Libraries/LibJS/Runtime/GetterSetterPair.h
mattco98: My thought process here was that, since the only types of values that are valid for getters and setters are `undefined` and functions, that I should store them only as one of those two types. Otherwise, in the places that I currently check for them not being undefined, I'd have to check for both empty and undefined. 
Which of course isn't really a problem, it only happens in five or so places, but I'm not sure that we would really ever need to distinguish between empty and undefined values for Accessors.

awesomekling: Seems like the thing we should really be checking is `is_function()`, no? :)

mattco98: Ehh, yup... that makes a lot more sense :)

PR #2319, opened 2020-05-21
LibWeb: Fix set{Interval,Timout}() callback this value

linusg:
We were giving it an empty value, which shouldn't happen, ever - so now we

1. set the window object as this value and
2. enforce explicitly setting a this value when calling `Interpreter::call()` by making `this_value` a required argument.

Also add a simple demo for `setInterval()` as we had nothing to showcase it :^)


awesomekling:
Great fix!

PR #2320, opened 2020-05-21
Where cxbyte attempts a comprehensive fix of LibLine, and sprinkles more Utf8Views in it...and ends up refactoring the shit out of it

alimpfard:
- [x] successive tabs after completion should not mess with the anchored styles
- [x] emojis inside suggestions should not mess with the anchored styles
- [x] cycling emoji suggestions should not mess with the anchored styles
- [x] suggestion previews should be spaced properly, even if they have emojis
- [x] anchored styles should not disappear after
    - [x] failed completion attempt
    - [x] successful completion attempt for an unrelated token
- [x] anchored styles should be correctly applied after cycling suggestion pages
- [x] paging should work
- [x] going back a page. (how = by precalculating page boundaries)
- [x] cute little next-page indicator
- [x] redrawing completion previews should not be so dang slow
- [x] Userland/js should not do weird things when suggesting stuff

Please tell me of any edge cases you might think of~

This PR (so far) unbreaks paging because refactoring made that easy.

### DIFF
--- a/Libraries/LibLine/Editor.cpp
+++ b/Libraries/LibLine/Editor.cpp
@@ -78,6 +78,12 @@ void Editor::clear_line()
     m_inline_search_cursor = m_cursor;
 }
 
+void Editor::insert(const Utf32View& string)
+{
+    for (size_t i = 0; i < string.length(); ++i)
+        insert(string.codepoints()[i]);
+}
+
 void Editor::insert(const String& string)
 {
     for (auto ch : Utf8View { string })
@@ -494,17 +500,18 @@ String Editor::get_line(const String& prompt)
                     m_suggestions = on_tab_complete(*this);
                     size_t common_suggestion_prefix { 0 };
                     if (m_suggestions.size() == 1) {
-                        m_largest_common_suggestion_prefix_length = m_suggestions[0].text.length();
+                        m_largest_common_suggestion_prefix_length = m_suggestions[0].text_view.length();
                     } else if (m_suggestions.size()) {
-                        char last_valid_suggestion_char;
+                        u32 last_valid_suggestion_codepoint;
+
                         for (;; ++common_suggestion_prefix) {
-                            if (m_suggestions[0].text.length() <= common_suggestion_prefix)
+                            if (m_suggestions[0].text_view.length() <= common_suggestion_prefix)
                                 goto no_more_commons;
 
-                            last_valid_suggestion_char = m_suggestions[0].text[common_suggestion_prefix];
+                            last_valid_suggestion_codepoint = m_suggestions[0].text_view.codepoints()[common_suggestion_prefix];
 
-                            for (const auto& suggestion : m_suggestions) {
-                                if (suggestion.text.length() <= common_suggestion_prefix || suggestion.text[common_suggestion_prefix] != last_valid_suggestion_char) {
+                            for (auto& suggestion : m_suggestions) {
+                                if (suggestion.text_view.length() <= common_suggestion_prefix || suggestion.text_view.codepoints()[common_suggestion_prefix] != last_valid_suggestion_codepoint) {
                                     goto no_more_commons;
                                 }
                             }
@@ -546,7 +553,7 @@ String Editor::get_line(const String& prompt)
                         case 2:
                             actual_offset = m_cursor - m_largest_common_suggestion_prefix_length + m_next_suggestion_invariant_offset;
                             if (can_complete)
-                                shown_length = m_largest_common_suggestion_prefix_length + m_last_shown_suggestion.trailing_trivia.length();
+                                shown_length = m_largest_common_suggestion_prefix_length + m_last_shown_suggestion.trivia_view.length();
                             break;
                         default:
                             if (m_last_shown_suggestion_display_length == 0)
@@ -558,28 +565,31 @@ String Editor::get_line(const String& prompt)
 
                         for (size_t i = m_next_suggestion_invariant_offset; i < shown_length; ++i)
                             remove_at_index(actual_offset);
+
                         m_cursor = actual_offset;
                         m_inline_search_cursor = m_cursor;
                         m_refresh_needed = true;
                     }
                     m_last_shown_suggestion = m_suggestions[m_next_suggestion_index];
-                    m_last_shown_suggestion.token_start_index = token_start - m_next_suggestion_invariant_offset - m_next_suggestion_static_offset;
-#ifdef SUGGESTIONS_DEBUG
-                    dbg() << "Last shown suggestion token start index: " << m_last_shown_suggestion.token_start_index << " Token Start " << token_start << " invariant offset " << m_next_suggestion_invariant_offset;
-#endif
-                    m_last_shown_suggestion_display_length = m_last_shown_suggestion.text.length();
+
+                    if (m_last_shown_suggestion_display_length)
+                        m_last_shown_suggestion.token_start_index = token_start - m_next_suggestion_static_offset - m_last_shown_suggestion_display_length;
+                    else
+                        m_last_shown_suggestion.token_start_index = token_start - m_next_suggestion_static_offset - m_next_suggestion_invariant_offset;
+
+                    m_last_shown_suggestion_display_length = m_last_shown_suggestion.text_view.length();
                     m_last_shown_suggestion_was_complete = true;
                     if (m_times_tab_pressed == 1) {
                         // This is the first time, so only auto-complete *if possible*
                         if (can_complete) {
-                            insert(m_last_shown_suggestion.text.substring_view(m_next_suggestion_invariant_offset, m_largest_common_suggestion_prefix_length - m_next_suggestion_invariant_offset));
+                            insert(m_last_shown_suggestion.text_view.substring_view(m_next_suggestion_invariant_offset, m_largest_common_suggestion_prefix_length - m_next_suggestion_invariant_offset));
                             m_last_shown_suggestion_display_length = m_largest_common_suggestion_prefix_length;
                             // do not increment the suggestion index, as the first tab should only be a *peek*
                             if (m_suggestions.size() == 1) {
                                 // if there's one suggestion, commit and forget
                                 m_times_tab_pressed = 0;
                                 // add in the trivia of the last selected suggestion
-                                insert(m_last_shown_suggestion.trailing_trivia);
+                                insert(m_last_shown_suggestion.trivia_view);
                                 m_last_shown_suggestion_display_length = 0;
                                 readjust_anchored_styles(m_last_shown_suggestion.token_start_index, ModificationKind::ForcedOverlapRemoval);
                                 stylize({ m_last_shown_suggestion.token_start_index, m_cursor, Span::Mode::CodepointOriented }, m_last_shown_suggestion.style);
@@ -590,10 +600,10 @@ String Editor::get_line(const String& prompt)
                         ++m_times_tab_pressed;
                         m_last_shown_suggestion_was_complete = false;
                     } else {
-                        insert(m_last_shown_suggestion.text.substring_view(m_next_suggestion_invariant_offset, m_last_shown_suggestion.text.length() - m_next_suggestion_invariant_offset));
+                        insert(m_last_shown_suggestion.text_view.substring_view(m_next_suggestion_invariant_offset, m_last_shown_suggestion.text_view.length() - m_next_suggestion_invariant_offset));
                         // add in the trivia of the last selected suggestion
-                        insert(m_last_shown_suggestion.trailing_trivia);
-                        m_last_shown_suggestion_display_length += m_last_shown_suggestion.trailing_trivia.length();
+                        insert(m_last_shown_suggestion.trivia_view);
+                        m_last_shown_suggestion_display_length += m_last_shown_suggestion.trivia_view.length();
                         if (m_tab_direction == TabDirection::Forward)
                             increment_suggestion_index();
                         else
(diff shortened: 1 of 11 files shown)


awesomekling:
cc @bugaevc

alimpfard:
I always forget to cc him :facepalm:

Libraries/LibLine/Editor.cpp
```diff
@@ -126,6 +126,36 @@ static size_t codepoint_length_in_utf8(u32 codepoint)
     return 3;
 }
 
+Editor::CodepointRange Editor::scan_for_byte_offsets(size_t start_offset, size_t end_offset, size_t scan_offset, bool reverse) const
+{
+    size_t byte_offset = 0;
+    size_t codepoint_offset = scan_offset + reverse;
```
bugaevc, line 138: Adding a `bool` to a `size_t`? If that is what you intended to do, maybe write it explicitly?

Libraries/LibLine/Editor.h
```diff
@@ -293,6 +289,11 @@ class Editor {
     void recalculate_origin();
     void reposition_cursor();
 
+    struct CodepointRange {
+        size_t start { 0 }, end { 0 };
```
bugaevc, line 310:
```suggestion
        size_t start { 0 };
        size_t end { 0 };
```

Would be more Serenity C++ style

Libraries/LibLine/Editor.cpp
```diff
@@ -126,6 +126,36 @@ static size_t codepoint_length_in_utf8(u32 codepoint)
...
+    CodepointRange offsets;
```
bugaevc, line 139: `start_offset`, `end_offset`, `scan_offset`, `byte_offset`, `codepoint_offset`, `offsets`, oh my!

If you tell me what this method does, I'll think of better names for these all.

Libraries/LibLine/Editor.cpp
alimpfard: Fine, I'll cast it.
those C habits die hard

bugaevc: I mean maybe `reverse ? scan_offset + 1 : scan_offset`

Libraries/LibLine/Editor.cpp
alimpfard: This method converts a range described by relative byte offsets ({start,end}_offset) to a codepoint offset (`offsets`), `scan_offset` is where it should start from, and reverse is quite clear.

Libraries/LibLine/Editor.cpp
alimpfard: but...why introduce a conditional branch that the compiler might decide not to optimise away?

awesomekling: Let's always prioritize clarity until profiling proves that something is a bottleneck :)

alimpfard: Certainly, but `number + boolean` was very clear to me.
regardless, I changed it to be explicit `foo + (bar ? 1 : 0)`

Libraries/LibLine/Editor.cpp
bugaevc: OK, it's taken me some more reads of the function even after your explanation, but I think I get it now.

So the first reason why it's confusing is you call both byte offsets and codepoint indexes *offsets*. Don't, I guess?

What you do is you iterate, over a UTF-32 buffer (`m_buffer`), from `scan_offset` (which is a codepoint index, not a byte offset) either forwards or backwards, and you're counting how many bytes this would take if encoded in UTF-8, without actually encoding it in UTF-8. You're interested in the range of codepoint indexes for a range of byte offsets, `start_offset` and `end_offset`. Is that right?

I would call the method `offset_range_to_codepoint_range()`. I still don't understand why you want to scan in reverse; perhaps instead of scanning from `index_to_scan_from` down to 0 (you're always scanning to the end, don't you?), I'd scan (from 0) up until I find the codepoint index for `end_offset` and stop there. So I wouldn't need the other two arguments. Or I'm misunderstanding how this is supposed to work.

alimpfard: my brain just calls relative indices offsets, I can't help it (hence why your help on naming is greatly appreciated!)

on `reverse==true`, you're right, the scan is always from 0.
but otherwise, the scan is not from the beginning or the end of the buffer.
the buffer might be N codepoints long, but the reverse scan might be initiated at codepoint offset M (M < N).

alimpfard: Here's a nice diagram:
```
buffer [ 0 1 2 3 . . . A . . . B . . . M . . . N ]
                       ^       ^       ^       ^
                       |       |       |       +- end of buffer
                       |       |       +- scan offset
                       |       +- range end
                       +- range start 
```

bugaevc: Yes, but instead of M→0 you could be scanning 0→M, right? And then instead of passing M explicitly, you could make the method stop as soon as it hits `end_offset`.

bugaevc: Nice diagram, perhaps put that, and some of the prose, into the code as comments?

alimpfard: How would that work?
I need offsets from M, not from 0

alimpfard: Sure :)

bugaevc: Ah, you need offsets from M! That wasn't at all clear!

OK, at this point I'd add tons of comments about all of this :smile:

alimpfard: If the comments (now) included are not clear enough, please tell me so I can try clarify more~

alimpfard:
This fixes all the issues I could find.
If there is any other issues, _please_ tell me about them.

bugaevc:
Other than that, I still don't understand LibLine internals well enough to tell :|

alimpfard:
\*cough\* sometimes I don't get the thing either, it's become much more complex than I really wanted it to be.

The price to pay for having niceties, I guess.

bugaevc:
No, it's just a sign that you have to rethink the design :wink: Split it into several types, where each one has its own clearly defined scope of responsibility, and doesn't know about internals of others.

(That's a generic advice, I don't have anything specific in mind — in particular because, again, I don't understand it.)

alimpfard:
:thinking: 
my immediate thought is "but they're all interconnected!"
I guess I'll take a few steps back and try to split this thing up.

bugaevc:
Off the top of my head, *presenting* the completions to the user (whether in a terminal that speaks xterm escape codes and positions UTF-8 codepoints on a grid, a terminal that doesn't, or potentially a GUI autocompletion widget, — sounds like an abstract class with a few implementations to me), splitting it into tokens is another thing, and managing completions — finding a common prefix and so on — is yet another thing.

alimpfard:
> presenting the completions to the user [...] potentially a GUI autocompletion widget

I'm not sure I want this library to be used for a GUI widget, it's very much not compatible.

> splitting it into tokens is another thing

Ultimately, I want this line editor to just not care.
I want it to have no such notion of "token", and I am partly there, once I handle this pesky C-w, there will be no concept of "token" in this thing.

> managing completions — finding a common prefix and so on — is yet another thing.

This makes sense to me.

bugaevc:
> I'm not sure I want this library to be used for a GUI widget, it's very much not compatible.

I don't mean using LibLine to drive a `GUI::Model`. I mean having LibLine emit some magical escape sequences that the Terminal will then interpret and use to display suggestions in a GUI popup. (Actually a lot like the recent hyperlinks.)

The point is, you wouldn't have to align the suggestions into columns, you would present them some other way. That means the presenter should not know too much about how the suggestions get generated / decided upon, and the code that generates and filters them should know next to nothing about how they are presented.

alimpfard:
> I mean having LibLine emit some magical escape sequences that the Terminal will then interpret and use to display suggestions in a GUI popup

I love that idea.

alimpfard:
That seemed to be the only reason.
I can't think of any other reason why that specific change would make the whole thing look less jumpy.

Try it, make like 400 files and try to cycle between the suggestions by just holding tab down.
The cursor appears to jump to the middle before that patch, and not after.

alimpfard:
I _did_ confirm that it's not the multiple printfs inside the loop...
Maybe I should try profiling it somehow...

alimpfard:
oh...kay? now it's not jumping around, even with `VT::{save,restore}_cursor` :/
I'll just drop that commit

alimpfard:
The jumping around was because of a `dbg()` in that loop lol

bugaevc:
But why would `dbg()` affect stdout/stderr?

alimpfard:
it probably blocks long enough for the print pause to be visible between the two styling flushes

alimpfard:
@bugaevc @awesomekling, what do yall think about the "more>" at the bottom? should it be a thing or no?

shot-2020-05-22_19-07-23

awesomekling:
I'm kinda liking the `more` thing, although it doesn't make it clear how to actually "get more"

alimpfard:
hmmm...I'd like to keep its footprint to be small, otherwise I'd say "more pages available>"

awesomekling:
> hmmm...I'd like to keep its footprint to be small, otherwise I'd say "more pages available>"

Since you say "pages", perhaps it could say "page 1 of 3" or similar?

alimpfard:
that could work.
`< page 2 / 3 >`...with or without the arrows?

alimpfard:
I'm gonna go with the arrows, I dunno.

alimpfard:
I think that's it for this PR.
if @bugaevc approves, I have nothing to add or change here.

Further splitting of LibLine would be nice, but I don't think I'm in the right headspace for more refactoring work right now.

Libraries/LibLine/XtermSuggestionDisplay.cpp
```diff
@@ -62,8 +64,45 @@ void XtermSuggestionDisplay::display(const SuggestionManager& manager)
         lines_used += max_line_count;
         longest_suggestion_length = 0;
     }
+
     VT::move_absolute(max_line_count + m_origin_x, 1);
 
+    if (m_pages.size() == 0) {
```
bugaevc, line 70: [Please use](https://rust-lang.github.io/rust-clippy/master/#len_zero) `is_empty()` instead of checking for `size() == 0`.

Libraries/LibLine/XtermSuggestionDisplay.cpp
```diff
@@ -62,8 +64,45 @@ void XtermSuggestionDisplay::display(const SuggestionManager& manager)
...
+        size_t num_printed = 0;
+        size_t lines_used { 1 };
```
bugaevc, line 72:
```suggestion
        size_t lines_used = 1;
```

for trivially constructed locals is the Serenity style.

alimpfard: Noted.

Libraries/LibLine/XtermSuggestionDisplay.cpp
```diff
@@ -31,7 +31,7 @@
 
 namespace Line {
 
-void DumbTerminalSuggestionDisplay::display(const SuggestionManager& manager)
+void XtermSuggestionDisplay::display(const SuggestionManager& manager)
```
bugaevc, line 36: This doesn't seem to belong in this commit.

Libraries/LibLine/XtermSuggestionDisplay.cpp
alimpfard: Does this really make sense in this context? we know that `Vector::size()` is just a member access, inline at that...

Applied nonetheless.

Libraries/LibLine/XtermSuggestionDisplay.cpp
alimpfard: I thought I moved that change to its origin commit...ouch.

alimpfard: Fixing that is gonna _hurt_...T_T

alimpfard: Fixed :)

Libraries/LibLine/SuggestionManager.cpp
```diff
@@ -0,0 +1,190 @@
...
+
+            result.offset_region_to_remove = { m_next_suggestion_invariant_offset, shown_length };
+            result.new_cursor_offset = actual_offset;
+        }
+
+        auto& suggestion = suggest();
+        set_current_suggestion_initiation_index(initiation_start_index);
+
+        if (mode == CompletePrefix) {
+            // only auto-complete *if possible*
```
awesomekling, line 151: https://github.com/SerenityOS/serenity/blob/master/Documentation/CodingStyle.md#comments

alimpfard: uhhh...capitalisation?

alimpfard: Mind if I do this in a separate commit?
tracking the comments through the history is a bit...annoying.
I could do that if needed though.

awesomekling: That's fine! :)

alimpfard: Github seems very broken today...

anyway, proper comments go _brrrrrrrrrrrrrrrrrrrrrrrrrrrrrr_

awesomekling:
:star:

PR #2324, opened 2020-05-21
LibJS: Refactor Array.prototype callback functions and make them generic

linusg:
There are more to be made generic, but doing the ones taking a callback and optional this value was most straightforward as they have more or less the same logic every time.

### DIFF
--- a/Libraries/LibJS/Tests/Array.prototype-generic-functions.js
+++ b/Libraries/LibJS/Tests/Array.prototype-generic-functions.js
@@ -1,0 +1,47 @@
+load("test-common.js");
+
+try {
+    const o = { length: 5, 0: "foo", 1: "bar", 3: "baz" };
+
+    ["every"].forEach(name => {
+        const visited = [];
+        Array.prototype[name].call(o, function (value) {
+            visited.push(value);
+            return true;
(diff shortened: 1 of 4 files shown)


Libraries/LibJS/Tests/Array.prototype-generic-functions.js
```diff
@@ -0,0 +1,47 @@
+load("test-common.js");
+
+try {
+    const o = { length: 5, 0: "foo", 1: "bar", 3: "baz" };
+
+    ["every"].forEach(name => {
```
Dexesttp, line 6: I'is there a reason for "every" to be away from "filter", forEach", etc..., or is it leftover from debugging ?

linusg: It's on purpose, the latter have `return false` in the callback and `every` has `return true`, so it doesn't abort after the first item and visits all of them for the sake of testing.

Dexesttp: Oh, didn't see that !

awesomekling:
These are great!

Issue #1549, opened 2020-03-31
Ext2FS: Copying a large file then deleting the copy causes system freeze


bcoles:
Copying a large (11MB) file then deleting the copy causes the system to freeze.

Reproduce:

```
cp DOOM.WAD DOOM.WAD.3
sleep 10
rm DOOM.WAD.3
```

Log:

```
Shell(19): Command "cp DOOM.WAD DOOM.WAD.3" finished in 2543 ms
[syncd(2:2)]: Ext2FS: Flushed 3040 blocks to disk
[Shell(204:204)]: do_exec(/bin/rm)
Shell(19): Command "rm DOOM.WAD.3" finished in 56 ms
[syncd(2:2)]: Ext2FS: Flushed 3 blocks to disk
[syncd(2:2)]: ASSERTION FAILED: false
FileSystem/Ext2FileSystem.cpp:1302 in bool Kernel::Ext2FS::set_block_allocation_state(Kernel::Ext2FS::BlockIndex, bool)
[syncd(2:2)]: 0xc01186bb  __assertion_failed(char const*, char const*, unsigned int, char const*) +180
[syncd(2:2)]: 0xc012ada4  Kernel::Ext2FS::set_block_allocation_state(unsigned int, bool) +244
[syncd(2:2)]: 0xc012cdb9  Kernel::Ext2FS::free_inode(Kernel::Ext2FSInode&) +193
[syncd(2:2)]: 0xc012ce8c  Kernel::Ext2FSInode::~Ext2FSInode() +48
[syncd(2:2)]: 0xc012ced6  Kernel::Ext2FSInode::~Ext2FSInode() +28
[syncd(2:2)]: 0xc018705b  AK::HashMap<unsigned int, AK::RefPtr<Kernel::Ext2FSInode>, AK::Traits<unsigned int> >::remove(unsigned int const&) +207
[syncd(2:2)]: 0xc012b1de  Kernel::Ext2FS::uncache_inode(unsigned int) +50
[syncd(2:2)]: 0xc012e34e  Kernel::Ext2FS::flush_writes() +410
[syncd(2:2)]: 0xc012ff98  Kernel::FS::sync() +638
[syncd(2:2)]: 0xc013b817  Kernel::VFS::sync() +21
[syncd(2:2)]: 0xc0175756  init::{lambda()#1}::_FUN() +40
```

supercomputer7:
[...]

PR #2325, opened 2020-05-21
Ext2FS: Fix indirect-blocks iteration

Jongy:
For singly-indirect blocks, "callback" is just "add_block".
For doubly-indirect blocks, "callback" is the lambda function
iterating on singly-indirect blocks: so instead of adding itself to the
list, the doubly-indirect block will add all its childs, but they add
themselves again when they run the callback of singly-indirect blocks.
And nothing adds the doubly-indirect block itself :(

This leads to a double free of all child blocks of the doubly-indirect
block, which is the failed assert described in #1549.

Closes: #1549.


awesomekling:
Excellent! :)

PR #2327, opened 2020-05-22
LibJS: Add object literal getter/setter shorthand

mattco98:
Adds support for the following syntax:

```js
let foo = {
    get x() {
        // ...
    },
    set x(value) {
        // ...
    }
}
```

The tests file is a bit shorter than I'd like it to be, but that's due to some inconsistencies with getters and setters that aren't related to this PR. I intend to fix up the edge cases and add more tests in a different PR, which will include more tests for these shorthand properties.


awesomekling:
LGTM!

PR #2331, opened 2020-05-22
LibC: Sync file position when dropping read ahead buffer

bugaevc:
When we flush a `FILE`, we behave differently depending on whether we reading from the file or writing to it:

* If we're writing, we actually write out the buffered data.
* If we're reading, we just drop the buffered (read ahead) data.

After flushing, there should be no additional buffered state stdio keeps about a `FILE`, compared to what is true about the underlying file. This includes file position (offset). When flushing writes, this is taken care of automatically, but dropping the buffer is not enough to achieve that when reading. This commit fixes that by seeking back explicitly in that case.

One way the problem manifested itself was upon `fseek(SEEK_CUR)` calls, as the position of the underlying file was oftentimes different to the logical position of the `FILE`. Since `FILE::seek()` already calls `FILE::flush()` prior to actually modifying the position, fixing `FILE::flush()` to sync the positions is enough to fix that issue.

### DIFF
--- a/Libraries/LibC/stdio.cpp
+++ b/Libraries/LibC/stdio.cpp
@@ -92,6 +92,7 @@ private:
 
         bool may_use() const { return m_ungotten || m_mode != _IONBF; }
         bool is_not_empty() const { return m_ungotten || !m_empty; }
+        size_t buffered_size() const;
 
         const u8* begin_dequeue(size_t& available_size) const;
         void did_dequeue(size_t actual_size);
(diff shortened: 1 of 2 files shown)


Libraries/LibC/stdio.cpp
```diff
@@ -440,6 +454,19 @@ void FILE::Buffer::drop()
...
+size_t FILE::Buffer::buffered_size() const
+{
+    // Note: does not include the ungetc() buffer.
+
+    if (m_empty)
+        return 0;
+
+    if (m_begin < m_end)
+        return m_end - m_begin;
+    else
```
awesomekling, line 466: No need for `else` after `return`.

bugaevc: Yes, but it also doesn't hurt. This isn't an early return in some simple case (like say `m_empty` check above); the two case are equally important, so it makes sense to treat them both equally.

awesomekling: Meh, if you insist. It's just unnecessary cognitive burden to me :)

awesomekling: Thanks for the fix btw, looks great!

PR #2332, opened 2020-05-22
LibJS: Fix Array.prototype.{indexOf,lastIndexOf}()

linusg:
Both were returning `-1` for `[undefined].indexOf()`, which is incorrect - also `lastIndexOf` was *really* not working according to the spec.

Since our tests passed nonetheless we should learn from that and occasionally run them against a mature engine - there's no real value in adding tests if they test an incorrect implementation :clown_face:


awesomekling:
>there's no real value in adding tests if they test an incorrect implementation

But without tests how will we know if our wrong implementation regresses and accidentally becomes correct!?

linusg:
you have a point there. +1 for incorrect tests!

Issue #2330, opened 2020-05-22
LibJS: Allow index properties to have descriptors


mattco98:
Right now, any index property on an `Object` has no descriptors because they are stored in a separate `Vector<Value>`, and are not associated with any shapes. This leads to some odd behavior, such as the following:

```js
// Serenity
> let o = { 1: 20 }
undefined
> Object.getOwnPropertyDescriptor(o, 1)
undefined
```

```js
// Spidermonkey
> let o = { 1: 20 }
undefined
> Object.getOwnPropertyDescriptor(o, 1)
{ value: 20, writable: true, enumerable: true, configurable: true }
```

This problems also affects more than just that single `Object` method. Consider the following:

```js
let o = {
  get 1() { return 10; }
}
```

Because getters and setters are always inserted into an object using `put_own_property`, the getter above will be inserted into the object's `m_storage`, not it's indexed `m_elements`. When accessing `o[1]`, the object above will return `undefined` because `o[1]` will use `get_by_index`, which will not find the object in `m_elements`.

I know that the `m_elements` is an optimization thing, so I'm not quite sure what the best way to fix this problem is. At one point I thought just adding a `m_is_indexed` to `Shape` could fix it, but at that point we might as well just remove `m_elements` all together. 

Another idea I had was to make `m_elements` a `Vector<size_t>`, where each element simply points to the index of a value stored in `m_storage`. This way, the indexed property would still cause and be affected by Shape transitions while still allowing array iteration to be cheap.

It's also very possible that this is how every JS engine works and I have no idea what I'm talking about :smiley: 

Curious to get some feedback on this one.

awesomekling:
Indeed, indexed properties can have uncommon attributes. In the vast majority of cases they won't though, so I think the best solution here would be having two modes:

1. Common case fast-path. We store values in `m_elements` like we do now and they all have the same exact attributes.
2. Uncommon case slow-path. We store values in an alternate data structure (instead of `m_elements`) that also has an attribute field for each entry.

In fact, we will need an alternate array storage format regardless, to handle sparse arrays where some madman inserts something at index 99999999. Currently, that will cause us to expand `m_elements` to fit that index, which is obviously not ideal!

Note that this should not require any shape transitions since shapes are only concerned with named properties.

alimpfard:
> `get 1() { return 10; }`

Dear lord...why? what's the use-case for such a thing?

mattco98:
> Dear lord...why? what's the use-case for such a thing?

[...]

PR #2334, opened 2020-05-22
LibJS: Add Array.prototype.reduce

nooga:
Adds `Array.prototype.reduce` implementation based on https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/Reduce.

I chose not to use `for_each_item` in `ArrayPrototype::reduce` because `reduce` usually does not iterate over each array item and expects a slightly different callback than `map`, `each` etc.

### DIFF
--- a/Libraries/LibJS/Runtime/ArrayPrototype.cpp
+++ b/Libraries/LibJS/Runtime/ArrayPrototype.cpp
@@ -1,6 +1,7 @@
 /*
  * Copyright (c) 2020, Andreas Kling <kling@serenityos.org>
  * Copyright (c) 2020, Linus Groh <mail@linusgroh.de>
+ * Copyright (c) 2020, Marcin Gasperowicz <xnooga@gmail.com>
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
@@ -56,6 +57,7 @@ ArrayPrototype::ArrayPrototype()
     put_native_function("concat", concat, 1, attr);
     put_native_function("slice", slice, 2, attr);
     put_native_function("indexOf", index_of, 1, attr);
+    put_native_function("reduce", reduce, 1, attr);
     put_native_function("reverse", reverse, 0, attr);
     put_native_function("lastIndexOf", last_index_of, 1, attr);
     put_native_function("includes", includes, 1, attr);
@@ -345,6 +347,65 @@ Value ArrayPrototype::index_of(Interpreter& interpreter)
     }
 
     return Value(-1);
+}
+
+Value ArrayPrototype::reduce(Interpreter& interpreter)
+{
+    auto* this_object = interpreter.this_value().to_object(interpreter);
+    if (!this_object)
+        return {};
+
+    auto initial_length = get_length(interpreter, *this_object);
+    if (interpreter.exception())
+        return {};
+
+    size_t start = 0;
+
+    auto accumulator = js_undefined();
+    if (interpreter.argument_count() > 1) {
+        accumulator = interpreter.argument(1);
+    } else {
+        bool start_found = false;
+        while (!start_found && start < initial_length) {
+            auto value = this_object->get_by_index(start);
+            if (interpreter.exception())
+                return {};
+            start_found = !value.is_empty();
+            if (start_found)
+                accumulator = value;
+            start += 1;
+        }
+        if (!start_found) {
+            interpreter.throw_exception<TypeError>("Reduce of empty array with no initial value");
+            return {};
+        }
+    }
+
+    auto* callback_function = callback_from_args(interpreter, "reduce");
+    if (!callback_function)
+        return {};
+
+    auto this_value = js_undefined();
+
+    for (size_t i = start; i < initial_length; ++i) {
+        auto value = this_object->get_by_index(i);
+        if (interpreter.exception())
+            return {};
+        if (value.is_empty())
+            continue;
+
+        MarkedValueList arguments(interpreter.heap());
+        arguments.append(accumulator);
+        arguments.append(value);
+        arguments.append(Value((i32)i));
+        arguments.append(this_object);
+
+        accumulator = interpreter.call(*callback_function, this_value, move(arguments));
+        if (interpreter.exception())
+            return {};
+    }
+
+    return accumulator;
 }
 
 Value ArrayPrototype::reverse(Interpreter& interpreter)
--- a/Libraries/LibJS/Tests/Array.prototype.reduce.js
+++ b/Libraries/LibJS/Tests/Array.prototype.reduce.js
@@ -1,0 +1,105 @@
+load("test-common.js");
+
+try {
+    assert(Array.prototype.reduce.length === 1);
+
+    assertThrowsError(() => {
+        [1].reduce();
+    }, {
+        error: TypeError,
+        message: "Array.prototype.reduce() requires at least one argument"
+    });
+
+    assertThrowsError(() => {
+        [1].reduce(undefined);
+    }, {
+        error: TypeError,
+        message: "undefined is not a function"
+    });
+
+    assertThrowsError(() => {
+        [].reduce((a, x) => x);
+    }, {
+        error: TypeError,
+        message: "Reduce of empty array with no initial value"
+    });
+
+    assertThrowsError(() => {
+        [, ,].reduce((a, x) => x);
+    }, {
+        error: TypeError,
+        message: "Reduce of empty array with no initial value"
+    });
+
+    [1, 2].reduce(() => { assert(this === undefined) });
+
+    var callbackCalled = 0;
+    var callback = () => { callbackCalled++; return true };
+
+    assert([1].reduce(callback) === 1);
+    assert(callbackCalled === 0);
+
+    assert([, 1].reduce(callback) === 1);
+    assert(callbackCalled === 0);
+
+    callbackCalled = 0;
+    assert([1, 2, 3].reduce(callback) === true);
+    assert(callbackCalled === 2);
+
+    callbackCalled = 0;
+    assert([, , 1, 2, 3].reduce(callback) === true);
+    assert(callbackCalled === 2);
+
+    callbackCalled = 0;
+    assert([1, , , 10, , 100, , ,].reduce(callback) === true);
+    assert(callbackCalled === 2);
+
+    var constantlySad = () => ":^(";
+    var result = [].reduce(constantlySad, ":^)");
+    assert(result === ":^)");
+
+    result = [":^0"].reduce(constantlySad, ":^)");
+    assert(result === ":^(");
+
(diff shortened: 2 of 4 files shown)


Libraries/LibJS/Runtime/ArrayPrototype.cpp
```diff
@@ -347,6 +348,54 @@ Value ArrayPrototype::index_of(Interpreter& interpreter)
...
+        interpreter.throw_exception<TypeError>(String::format("Reduce of empty array with no initial value"));
+        return {};
+    }
+
+    auto* callback_function = callback_from_args(interpreter, "reduce");
+    if (!callback_function)
+        return {};
+
+    for (size_t i = start; i < initial_length; ++i) {
+        auto value = this_object->get_by_index(i);
```
linusg, line 380: Missing exception check:

```suggestion
        auto value = this_object->get_by_index(i);
        if (interpreter.exception())
            return {};
```

Libraries/LibJS/Runtime/ArrayPrototype.cpp
```diff
@@ -347,6 +348,54 @@ Value ArrayPrototype::index_of(Interpreter& interpreter)
...
+        return {};
+
+    size_t start = 1;
+
+    auto accumulator = js_undefined();
+    if (interpreter.argument_count() > 1) {
+        accumulator = interpreter.argument(1);
+        start = 0;
+    } else if (initial_length >= 1) {
+        accumulator = this_object->get_by_index(0);
```
linusg, line 369: Missing exception check:


```suggestion
        accumulator = this_object->get_by_index(0);
        if (interpreter.exception())
            return {};
```

Libraries/LibJS/Runtime/ArrayPrototype.cpp
```diff
@@ -347,6 +348,54 @@ Value ArrayPrototype::index_of(Interpreter& interpreter)
...
+            continue;
+        }
+
+        MarkedValueList arguments(interpreter.heap());
+        arguments.append(accumulator);
+        arguments.append(value);
+        arguments.append(Value((i32)i));
+        arguments.append(this_object);
+
+        accumulator = interpreter.call(*callback_function, this_value, move(arguments));
```
linusg, line 391: The *this* value here seems to be incorrect, the spec says:

> Set accumulator to ? Call(callbackfn, undefined, « accumulator, kValue, k, O »).

Libraries/LibJS/Tests/Array.prototype.reduce.js
```diff
@@ -0,0 +1,90 @@
...
+    result = [":^0"].reduce(constantlySad, ":^)");
+    assert(result === ":^(");
+
+    result = [":^0"].reduce(constantlySad);
+    assert(result === ":^0");
+
+    result = [5, 4, 3, 2, 1].reduce((accum, elem) => accum + elem);
+    assert(result === 15);
+
+    // likely parser bug: arrow func parser eats ", 100" as a part of the function, hence the parens
```
linusg, line 54: yep, `Parser::try_parse_arrow_function_expression()` needs to be able to take a min precedence value rather than just blindly eating up expressions (`parse_expression(0)`).

Libraries/LibJS/Runtime/ArrayPrototype.cpp
```diff
@@ -347,6 +348,54 @@ Value ArrayPrototype::index_of(Interpreter& interpreter)
...
+    auto this_value = interpreter.this_value();
+    auto* this_object = this_value.to_object(interpreter);
+    if (!this_object)
+        return {};
+
+    auto initial_length = get_length(interpreter, *this_object);
+    if (interpreter.exception())
+        return {};
+
+    size_t start = 1;
```
linusg, line 362: We can't just assume 1 as start if the initial value is missing, it needs to be a bit more complex than that - consider the following:

```js
> [,].reduce(() => {})
<empty>
```

i.e. an array with one element, an empty value - which we would expose.

Spec says:

```
If initialValue is present, then
    Set accumulator to initialValue.
Else,
    Let kPresent be false.
    Repeat, while kPresent is false and k < len,
        Let Pk be ! ToString(k).
        Set kPresent to ? HasProperty(O, Pk).
        If kPresent is true, then
            Set accumulator to ? Get(O, Pk).
        Set k to k + 1.
    If kPresent is false, throw a TypeError exception.
```

Libraries/LibJS/Runtime/ArrayPrototype.cpp
nooga: Yes, it looks like it should be just `js_undefined()` for `reduce` although V8 just leaves the function binding as it is. I will implement what the spec suggests though.

Libraries/LibJS/Tests/Array.prototype.reduce.js
nooga: Since this is not only about `reduce` I would leave this for another PR if that's ok.

linusg: yes, absolutely!

Libraries/LibJS/Runtime/ArrayPrototype.cpp
linusg: > although V8 just leaves the function binding as it is.

Looks like both SpiderMonkey and V8 do use `undefined` as this value:

```
> [1, 2, 3].reduce(function() { "use strict"; this.foo = "bar" })
Uncaught TypeError: Cannot set property 'foo' of undefined
    at repl:1:54
    at Array.reduce (<anonymous>)
```

Libraries/LibJS/Runtime/ArrayPrototype.cpp
nooga: Seems like those two places would be the only calls to `get_by_index` followed by an exception check. I'm not sure if this method can even throw a JS exception.

Libraries/LibJS/Runtime/ArrayPrototype.cpp
nooga: Ah! They do in strict mode! Right, so we're in strict by default?

Libraries/LibJS/Runtime/ArrayPrototype.cpp
linusg: It definitely can throw - `get_by_index()` has a code path for negative values that just calls regular `get()` which you can very easily blow up using a property getter function for example.

https://github.com/SerenityOS/serenity/blob/45dfa094e9ab88c4af833674e85a93b704c79988/Libraries/LibJS/Runtime/Object.cpp#L371-L372

Now since here the index is guaranteed to be >= 0 we could leave it out RIGHT NOW but that's actually due to a missing piece in the implementation - getters for positive indices will be possible too, see https://github.com/SerenityOS/serenity/issues/2330.

> Seems like those two places would be the only calls to get_by_index followed by an exception check.

There are others:

https://github.com/SerenityOS/serenity/blob/040c75a3cc8ae4ac87bcc03bf741b47bc688d2a0/Libraries/LibJS/Runtime/ArrayPrototype.cpp#L224-L226

https://github.com/SerenityOS/serenity/blob/040c75a3cc8ae4ac87bcc03bf741b47bc688d2a0/Libraries/LibJS/Runtime/ArrayPrototype.cpp#L275-L277

`for_each_item` is missing one.

Libraries/LibJS/Runtime/ArrayPrototype.cpp
linusg: No strict mode in LibJS yet - that was just for demonstration purposes, as in non-strict mode this is still `undefined` here but assigning a property to it just fails silently :)

In this PR's implementation it would assign the property to the array that `reduce` is being called on:

```
> a = [1, 2, 3]
[ 1, 2, 3 ]
> a.reduce(() => { this.foo = "bar" })
> a.foo
"bar"
```

Libraries/LibJS/Runtime/ArrayPrototype.cpp
```diff
@@ -347,6 +349,67 @@ Value ArrayPrototype::index_of(Interpreter& interpreter)
...
+    auto this_value = js_undefined();
+
+    for (size_t i = start; i < initial_length; ++i) {
+        auto value = this_object->get_by_index(i);
+        if (interpreter.exception())
+            return {};
+
+        if (value.is_empty()) {
+            continue;
+        }
```
linusg, line 397: It's not explicitly mentioned in https://github.com/SerenityOS/serenity/blob/master/Documentation/CodingStyle.md as far as I can see, but Serenity code usually omits curly braces if the body is a single line - i.e.

```cpp
if (value.is_empty())
    continue;
```

Libraries/LibJS/Tests/Array.prototype.reduce.js
```diff
@@ -0,0 +1,114 @@
...
+    a1.reduce((a, v, i, t) => { a2 = t });
+    assert(a1 === a2);
+
+
+
+
+
+
+
+
```
linusg, line 110: That's a lot of whitespace - let's please get rid of that :)

Also, if you don't mind - a short test in `Array.prototype-generic-functions.js` showcasing `reduce`'s ability to be generic and work with anything that has a `length` would be fantastic!

Libraries/LibJS/Runtime/ArrayPrototype.cpp
```diff
@@ -347,6 +349,65 @@ Value ArrayPrototype::index_of(Interpreter& interpreter)
...
+            auto value = this_object->get_by_index(start);
+            if (interpreter.exception())
+                return {};
+            start_found = !value.is_empty();
+            if (start_found)
+                accumulator = value;
+            start += 1;
+        }
+        if (!start_found) {
+            interpreter.throw_exception<TypeError>(String::format("Reduce of empty array with no initial value"));
```
linusg, line 379:
```suggestion
            interpreter.throw_exception<TypeError>("Reduce of empty array with no initial value");
```

nooga: whoops... fixed

awesomekling:
Patch looks good! 
When squashing commits, please tidy up the commit message so it reflects the new content of the patch instead of just being all the old commit messages mashed together :)

awesomekling:
:star:

Libraries/LibJS/Tests/Array.prototype.reduce.js
nooga: @linusg I might be wrong but it looks like it just needs to call `parse_expression(1)` instead of `0`. Is there a sense in passing this value to `Parser::try_parse_arrow_function_expression()`.

linusg: at the moment I don't see a need for it, so we can probably just change it inside the function itself

PR #2335, opened 2020-05-22
LibGUI/Browser: Browser location box improvements

FalseHonesty:
This patch works on improving the functionality and feel of the Browser's location box.

Firstly, this patch adds a "Paste & Go" action to the location box's context menu, which will take the clipboard data, put it into the location box, and navigate to that page. Note: the clipboard's text overwrites the location box, which seems curious, but that is how Firefox works, and actually seems convenient.

Secondly, this patch makes the location bar lose focus after being activated. It felt odd that the location bar was still taking keyboard input after activating it, so this just makes it feel more natural.

Last but not least, this patch improves the double-click selection on the location box (and `TextEditor`s in general). Previously, double clicking would select the range around your click up until it found a space, and in the browser's location bar this behaviour didn't suffice. Now, it will select the range around your click until there is a "word break". A word break is considered to be when your selection changes from being alphanumeric to being non alphanumeric, or vice versa.

### DIFF
--- a/Applications/Browser/Tab.cpp
+++ b/Applications/Browser/Tab.cpp
@@ -113,7 +113,13 @@ Tab::Tab()
         }
 
         m_html_widget->load(location);
+        m_location_box->set_focus(false);
     };
+
+    m_location_box->add_custom_context_menu_action(GUI::Action::create("Paste & Go", [this](auto&) {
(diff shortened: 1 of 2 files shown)


awesomekling:
Digging it :)

PR #2336, opened 2020-05-22
LibC: Add (empty) netinet/tcp.h back

linusg:
This file is required for building the git port.

It was already added before and then removed again when the CI script for license header checks was added as it seemed irrelevant.

cc @bugaevc


awesomekling:
Oops! Thanks for fixing this Linus!

PR #2338, opened 2020-05-23
AK/LibWeb/Browser: Fix fragment/anchor handling

FalseHonesty:
This patch focuses on the handling of fragment links in the system.

For starters, it fixes URL completion of fragment strings. Previously, passing a fragment string (`#section3`) to the URL::complete_url method would result in a URL that looked like `file:///home/anon/www/#section3` which was obviously incorrect. Now the result looks like `file:///home/anon/www/afrag.html#section3`.

Next, it fixes HtmlView not automatically scrolling to an anchor when a URL with a fragment is loaded.

Finally, it fixes the handling of fragment links not opening in a new tab. Previously, clicking on an anchor link (`href="#section1"`) would always scroll to it on the current page, even if control was held or the target="_blank" attribute was set. This fixes that behaviour, and the link will instead open in a new tab if required.

### DIFF
--- a/AK/URL.cpp
+++ b/AK/URL.cpp
@@ -307,6 +307,12 @@ URL URL::complete_url(const String& string) const
         return url;
     }
 
+    if (string.starts_with("#")) {
+        url = *this;
+        url.set_fragment(string.substring(1, string.length() - 1));
+        return url;
+    }
+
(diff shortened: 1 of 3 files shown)


awesomekling:
Excellent fixes :)

PR #2341, opened 2020-05-23
Build: Make Lagom build under MacOS

nooga:
Lagom now builds under MacOS. 

Only two minor adjustments were required:
* LibCore TCP/UDP code requires `SOCK_{NONBLOCK, CLOEXEC}` - ~defined them in terms of `O_{NONBLOCK, CLOEXEC}`  from `fcntl.h`~ set them using `ioctl(...FIONBIO...)` and `fnctl(...FD_CLOEXEC)`
* 
* LibJS `Heap` code pthread usage ported to MacOS

Antalet ändringar var Lagom :^)

### DIFF
--- a/Libraries/LibCore/TCPServer.cpp
+++ b/Libraries/LibCore/TCPServer.cpp
@@ -33,17 +33,21 @@
 #include <sys/socket.h>
 
 #ifndef SOCK_NONBLOCK
-#    include <fcntl.h>
-#    define SOCK_NONBLOCK O_NONBLOCK
-#    define SOCK_CLOEXEC O_CLOEXEC
+#    include <sys/ioctl.h>
 #endif
-
 namespace Core {
 
 TCPServer::TCPServer(Object* parent)
     : Object(parent)
 {
+#ifdef SOCK_NONBLOCK
     m_fd = socket(AF_INET, SOCK_STREAM | SOCK_NONBLOCK | SOCK_CLOEXEC, 0);
+#else
+    m_fd = socket(AF_INET, SOCK_STREAM, 0);
+    int option = 1;
+    ioctl(m_fd, FIONBIO, &option);
+    fcntl(m_fd, F_SETFD, FD_CLOEXEC);
+#endif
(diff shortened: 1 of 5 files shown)


Libraries/LibCore/TCPServer.cpp
```diff
@@ -32,6 +32,12 @@
 #include <stdio.h>
 #include <sys/socket.h>
 
+#ifndef SOCK_NONBLOCK
+#    include <fcntl.h>
+#    define SOCK_NONBLOCK O_NONBLOCK
+#    define SOCK_CLOEXEC O_CLOEXEC
+#endif
```
awesomekling, line 39: I don't think this will work. `O_NONBLOCK` and `O_CLOEXEC` are options to `open()`, not to `socket()`.
On macOS, you'll need to call `fcntl()` separately. See for example how it already works in `Core::LocalSocket` :)

awesomekling:
:+1:

Libraries/LibJS/Heap/Heap.cpp
```diff
@@ -157,6 +157,14 @@ void Heap::gather_conservative_roots(HashTable<Cell*>& roots)
         ASSERT_NOT_REACHED();
     }
     pthread_attr_destroy(&attr);
+#elif __MACH__
```
bugaevc, line 160: Does this have anything to do with Mach? `__APPLE__` would be more appropriate here if you want to check for Darwin.

nooga: Both `__APPLE__` and `__MACH__` are defined on my system. Tbh. I'm not sure why I chose the latter. I guess I'd need to submit another PR to fix this then? :D

bugaevc: Yes, because macOS is a Darwin distribution, and Darwin uses Mach. But not anything that uses Mach is a version of Darwin :wink:

Yes, please submit a separate PR, and don't forget to cc me.

PR #2344, opened 2020-05-23
Userland/js: Set default styles for unstyled elements

### LINKED ISSUE #2340: Userland/js: Syntax highlighting is broken
Nonsense code, but you get the point.

image

image

cc @alimpfard

alimpfard:
Fixes #2340

LibLine: where fixing one thing breaks another


alimpfard:
Let's make LibLine behave more correctly, this patch feels like a hack.

PR #2345, opened 2020-05-23
Unbork sdb & functrace

itamar8910:
Some LibDebug fixes that fix regressions we've had in `sdb` and `functrace` :smiley:

### DIFF
--- a/Libraries/LibDebug/DebugInfo.cpp
+++ b/Libraries/LibDebug/DebugInfo.cpp
@@ -69,6 +69,10 @@ void DebugInfo::parse_scopes_impl(const Dwarf::DIE& die)
         if (name.has_value())
             scope.name = name.value().data.as_string;
 
+        if (!child.get_attribute(Dwarf::Attribute::LowPc).has_value()) {
+            dbg() << "DWARF: Couldn't find attribtue LowPc for scope";
+            return;
+        }
         scope.address_low = child.get_attribute(Dwarf::Attribute::LowPc).value().data.as_u32;
         // The attribute name HighPc is confusing. In this context, it seems to actually be a positive offset from LowPc
         scope.address_high = scope.address_low + child.get_attribute(Dwarf::Attribute::HighPc).value().data.as_u32;
@@ -88,9 +92,10 @@ void DebugInfo::parse_scopes_impl(const Dwarf::DIE& die)
 void DebugInfo::prepare_lines()
 {
     auto section = m_elf->image().lookup_section(".debug_line");
-    ASSERT(!section.is_undefined());
+    if (section.is_undefined())
+        return;
 
-    auto buffer = ByteBuffer::wrap(reinterpret_cast<const u8*>(section.raw_data()), section.size());
+    auto buffer = section.wrapping_byte_buffer();
     BufferStream stream(buffer);
 
     Vector<LineProgram::LineInfo> all_lines;
--- a/Applications/Debugger/main.cpp
+++ b/Applications/Debugger/main.cpp
@@ -31,6 +31,7 @@
 #include <AK/StringBuilder.h>
 #include <AK/kmalloc.h>
 #include <LibC/sys/arch/i386/regs.h>
+#include <LibCore/ArgsParser.h>
 #include <LibCore/File.h>
 #include <LibDebug/DebugInfo.h>
 #include <LibDebug/DebugSession.h>
@@ -44,12 +45,6 @@
 #include <unistd.h>
 
 static Line::Editor editor {};
-
-static int usage()
-{
-    printf("usage: sdb [command...]\n");
-    return 1;
-}
 
 OwnPtr<DebugSession> g_debug_session;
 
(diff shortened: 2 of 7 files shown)


Libraries/LibDebug/DebugInfo.cpp
```diff
@@ -69,6 +69,10 @@ void DebugInfo::parse_scopes_impl(const Dwarf::DIE& die)
         if (name.has_value())
             scope.name = name.value().data.as_string;
 
+        if (!child.get_attribute(Dwarf::Attribute::LowPc).has_value()) {
+            dbg() << "DWARF: Couldn't find attribtue LowPc for scope";
```
awesomekling, line 73: Typo: attribtue

Libraries/LibDebug/Dwarf/DwarfInfo.h
```diff
@@ -51,12 +52,12 @@ class DwarfInfo : public RefCounted<DwarfInfo> {
...
-    ByteBuffer section_data(const String& section_name);
+    Optional<ByteBuffer> section_data(const String& section_name);
 
     NonnullRefPtr<const ELF::Loader> m_elf;
-    ByteBuffer m_debug_info_data;
-    ByteBuffer m_abbreviation_data;
-    ByteBuffer m_debug_strings_data;
+    Optional<ByteBuffer> m_debug_info_data;
+    Optional<ByteBuffer> m_abbreviation_data;
+    Optional<ByteBuffer> m_debug_strings_data;
```
awesomekling, line 60: `ByteBuffer` already has a null state (which is what you get with `ByteBuffer()`).
Let's use that instead of wrapping them in `Optional` :)

bugaevc:
Please also port them to use `Core::ArgsParser`. (I've noticed that Debugger implements its own args parsing, and refers to itself as "sdb" — not even `argv[0]`...)

alimpfard:
One day...@bugaevc will figure out a way to force everything to use ArgsParser...and I'll figure out a way to force ArgsParser to tell the Shell about its arguments.
Then Serenity Shell will be better than every other fucking shell in the universe.

bugaevc:
> Then Serenity Shell will be better than every other fucking shell in the universe.

Heh, we're already getting there thanks to your work 😀

> and I'll figure out a way to force ArgsParser to tell the Shell about its arguments.

That should be pretty easy actually. For example, we could say that `--dump-supported-options` is a special option that dumps info about supported options — a lot like the existing `--help`, but in JSON. Or we could just parse `--help` output.

alimpfard:
Hate to hijack this (nice) PR with this, but..
> That should be pretty easy actually

Oops, I started the browser while trying to complete its arguments :P

itamar8910:
An idea: We could store info about the required args format in some custom
elf section, a-la `.arguments`.
Both the Shell and `Core::ArgsParser` would read from it.
The question is how to populate this section. Maybe we could move the
arguments declaration to some preprocessor / template magic?

בתאריך יום א׳, 24 במאי 2020, 01:34, מאת Ali Mohammad Pur ‏<
notifications@github.com>:

bugaevc:
> Oops, I started the browser while trying to complete its arguments :P

I mean, `--dump-supported-options` would exit after dumping the options, much like `--help`. For extra safety, we could run it with nothing but stdio pledged, so it wouldn't be able to connect to WindowServer, or act on any files, and so on, even if it doesn't use `Core::ArgsParser` and/or mistakenly tries to perform some action before parsing its args.

alimpfard:
We (I and @Dexesttp) did discuss this a bit on irc,

@itamar8910 that wouldn't work for cases where the program has somewhat complex logic regarding its args - say custom handlers.

@bugaevc I don't think it's a sane idea by itself, a program with just stdio would at least be able to block forever, no, we need a definite indication that the program actually uses argsparser...

bugaevc:
> a program with just stdio would at least be able to block forever

That's not a problem either, just kill it after a timeout.

> no, we need a definite indication that the program actually uses argsparser...

I don't think we *need* it, but indeed it'd be nice to have.

alimpfard:
I dunno, maybe I'm blowing this way outta proportion, but it seems unwise to run an arbitrary program to maybe get a description of its arguments

awesomekling:
@bugaevc @alimpfard -> #2357 :nerd_face:

bugaevc:
No, I totally get your concern. Perhaps we should indeed have a definite upfront indication if the program uses `Core::ArgsParser` — perhaps with a special ELF section, like @itamar8910 proposes. (Which actually reminds me of how A/UX Commando gets info about options supported by an executable from a special resource fork...)

But on the other hand, consider this. All of our programs are (or should be) using `Core::ArgsParser`. So it's only external ports that could pose a problem. If we run anything completely sandboxed, with a timeout, where basically the only thing they can do is output some data to stdout, and attempting anything else, or not finishing in 0.5 sec, gets them terminated... it doesn't sound so problematic to me.

PR #2346, opened 2020-05-23
LibLine: Default to resetting styles when a cell has no styles set

### LINKED ISSUE #2340: Userland/js: Syntax highlighting is broken

alimpfard:
Fixes #2340


awesomekling:
That makes sense. Perhaps there are more situations where we should reset the style to default instead of leaving it unchanged?

alimpfard:
Everything else uses `find_applicable_style` :)

PR #2348, opened 2020-05-23
LibWeb: Improve support for white-space CSS property

jgbyrne:
Add reasonable support for all values of white-space CSS property.

Values of the property are translated into a 3-tuple of rules:
    do_collapse :       whether whitespace is to be collapsed
    do_wrap_lines:      whether to wrap on word boundaries when
                        lines get too long
    do_wrap_breaks:     whether to wrap on linebreaks

The previously seperate handling of per-line splitting and per-word
splitting have been unified. The Word structure is now a more general Chunk,
which represents different amounts of text depending on splitting rules.

### DIFF
--- a/Libraries/LibWeb/Layout/LayoutText.cpp
+++ b/Libraries/LibWeb/Layout/LayoutText.cpp
@@ -27,8 +27,8 @@
 #include <AK/StringBuilder.h>
 #include <AK/Utf8View.h>
 #include <LibCore/DirIterator.h>
-#include <LibGfx/Font.h>
 #include <LibGUI/Painter.h>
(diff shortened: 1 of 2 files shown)


awesomekling:
This breaks rendering of https://www.100mb.nl/ and turns it into a single line of text.

jgbyrne:
A classic error: when a newline chunk was pending to be created at the end of the chunking loop it was not being picked up. Fixed, in only two attempts at using git :p

jgbyrne:
Turns out --amend screws things up. Trying that again....

awesomekling:
Great! :star:

PR #2349, opened 2020-05-23
Browser: Add JS Console Window

FalseHonesty:
The JavaScript console can be opened with Control+I, or using the menu option. The console is currently a text box with JS syntax highlighting which will send commands to the document's interpreter. All output is printed to an HTML view in the console. The output is an HtmlView to easily allow complex output, such as expandable views for JS Objects in the long run.

The console as of this PR looks like this
It's currently not the prettiest, but output highlighting & formatting can probably be worked on in tree.

I'm not sure if this is what was envisioned for the Browser's JS Console, and if it's not, I'm happy to rework/abandon this PR.

### DIFF
--- a/Applications/Browser/BrowserConsoleClient.cpp
+++ b/Applications/Browser/BrowserConsoleClient.cpp
@@ -1,0 +1,131 @@
+/*
+ * Copyright (c) 2020, Hunter Salyer <thefalsehonesty@gmail.com>
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include "BrowserConsoleClient.h"
+#include "ConsoleWidget.h"
+#include <AK/StringBuilder.h>
+#include <LibGUI/BoxLayout.h>
+#include <LibGUI/JSSyntaxHighlighter.h>
+#include <LibGUI/TextBox.h>
+#include <LibJS/Interpreter.h>
+#include <LibWeb/DOM/DocumentType.h>
+#include <LibWeb/DOM/ElementFactory.h>
+#include <LibWeb/DOM/HTMLBodyElement.h>
+#include <LibWeb/DOM/Text.h>
+#include <LibWeb/DOMTreeModel.h>
+
+namespace Browser {
+
+JS::Value BrowserConsoleClient::log()
+{
+    m_console_widget.print_html(interpreter().join_arguments());
+    return JS::js_undefined();
+}
+
+JS::Value BrowserConsoleClient::info()
+{
+    StringBuilder html;
+    html.append("<span class=\"info\">");
+    html.append("(i) ");
+    html.append(interpreter().join_arguments());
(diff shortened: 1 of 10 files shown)


Applications/Browser/BrowserConsoleClient.cpp
```diff
@@ -0,0 +1,123 @@
+/*
+ * Copyright (c) 2018-2020, Hunter Salyer <thefalsehonesty@gmail.com>
```
Dexesttp, line 2: Just 2020 probably ? (it's the same for all the files)

Applications/Browser/BrowserConsoleClient.cpp
```diff
@@ -0,0 +1,123 @@
...
+
+JS::Value BrowserConsoleClient::log()
+{
+    m_console_widget.print_html(interpreter().join_arguments());
+    return JS::js_undefined();
+}
+
+JS::Value BrowserConsoleClient::info()
+{
+    m_console_widget.print_html(String::format("(i) %s", interpreter().join_arguments().characters()));
```
Dexesttp, line 50: Since you're adding an `(i)` for infos, how about adding a `(d)`, `(w)`, `(e)`, etc.. for the other calls ?

Applications/Browser/BrowserConsoleClient.cpp
linusg, line 2:
```suggestion
 * Copyright (c) 2020, Hunter Salyer <thefalsehonesty@gmail.com>
```

Applications/Browser/BrowserConsoleClient.cpp
```diff
@@ -0,0 +1,123 @@
...
+    return JS::js_undefined();
+}
+
+JS::Value BrowserConsoleClient::debug()
+{
+    StringBuilder html;
+    html.append("<i>");
+    html.append(interpreter().join_arguments());
+    html.append("</i>");
```
linusg, line 59: I don't think there's support for italics yet. can we use coloured text for debug/warn/error instead?

Applications/Browser/BrowserConsoleClient.cpp
```diff
@@ -0,0 +1,123 @@
...
+
+JS::Value BrowserConsoleClient::trace()
+{
+    StringBuilder html;
+    html.append(interpreter().join_arguments());
+    auto trace = interpreter().get_trace();
+    for (auto& function_name : trace) {
+        if (function_name.is_empty())
+            function_name = "<anonymous>";
+        html.appendf(" -> %s\n", function_name.characters());
```
linusg, line 98: should that be

```suggestion
        html.appendf(" -> %s<br>", function_name.characters());
```

?

Applications/Browser/BrowserConsoleClient.cpp
```diff
@@ -0,0 +1,123 @@
...
+}
+
+JS::Value BrowserConsoleClient::trace()
+{
+    StringBuilder html;
+    html.append(interpreter().join_arguments());
+    auto trace = interpreter().get_trace();
+    for (auto& function_name : trace) {
+        if (function_name.is_empty())
+            function_name = "<anonymous>";
```
linusg, line 97: is this being displayed correctly if inserted as HTML?

Applications/Browser/ConsoleWidget.cpp
```diff
@@ -0,0 +1,243 @@
...
+    m_console_input = add<GUI::TextBox>();
+    m_console_input->set_syntax_highlighter(make<GUI::JSSyntaxHighlighter>());
+    m_console_input->set_size_policy(GUI::SizePolicy::Fill, GUI::SizePolicy::Fixed);
+    m_console_input->set_preferred_size(0, 22);
+    // FIXME: Syntax Highlighting breaks the cursor's position on non fixed-width fonts.
+    m_console_input->set_font(Gfx::Font::default_fixed_width_font());
+
+    m_console_input->on_return_pressed = [this] {
+        auto js_source = m_console_input->text();
+        m_console_input->set_text("");
```
linusg, line 75: Probably better:

```suggestion
        m_console_input->clear();
```

Applications/Browser/BrowserConsoleClient.cpp
Dexesttp: Another possibility could be to wrap all of them in the right `<span class="error|warning|...">`, and have some custom css styling them. This way, once we have support for `::before` pseudoclasses we can easily add icons !

Applications/Browser/BrowserConsoleClient.cpp
FalseHonesty: Ah good catch, copy-paste strikes again!

Applications/Browser/BrowserConsoleClient.cpp
FalseHonesty: Sure, sounds reasonable. I just took the output message formatting from the `ReplConsoleClient`, so it might need to be updated there too.

Applications/Browser/BrowserConsoleClient.cpp
FalseHonesty: I think that I'm going to use @Dexesttp's suggestion and wrap them in spans with `info`, `error`, etc. CSS classes so formatting can be done in the base document.

Applications/Browser/BrowserConsoleClient.cpp
FalseHonesty: Yep absolutely, must have missed it in the conversion from `ReplConsoleClient`.

Applications/Browser/BrowserConsoleClient.cpp
FalseHonesty: Definitely not, I converted the `<` and `>` to `&lt;` and `&gt;`, I must have just missed this, nice spot.

Applications/Browser/ConsoleWidget.cpp
FalseHonesty: Ah didn't know about that one!

Applications/Browser/BrowserConsoleClient.cpp
FalseHonesty: The logging is all now wrapped in spans with `info`, `debug`, etc. classes :)

FalseHonesty:
> Vey cool stuff :)
> 
> I'm wondering whether you could have the console window created all the time, but just hidden. This way, you could store all the debug ouptut from the Browser's launch and it woudl appear when opening the console. Just a proposition though, and maybe it could be done later in another PR !

That's definitely something that will need to happen, we wouldn't want to lose all of the console output from the page before the browser's console actually opens.

awesomekling:
Very cool! The lifetime of the `Interpreter` will have to change a bit soon, but this is fine for now. :)

PR #2351, opened 2020-05-23
Base: Add a really bad, but yummy Slice of Pizza emoji 🍕

brynet:
I don't know how to draw, but noticed a very glaring omission.


awesomekling:
:pizza::heart:

PR #2354, opened 2020-05-23
LibGUI: Use word breaks to intelligently navigate a TextEditor

FalseHonesty:
More text editor improvements!

Previously, holding Control while using the left/right arrow keys
to navigate through a TextEditor would only be helpful if the document
had spans. Now, if there are no spans, it will navigate to the
next "word break", defined to be the threshold where text changes
from alphanumeric to non-alphanumeric, or vice versa.

This also refactors out the word break calculation from the TextEditor's selection code, and makes it usable everywhere else, so maybe we'll find more places it will be helpful!

### DIFF
--- a/Libraries/LibGUI/TextDocument.cpp
+++ b/Libraries/LibGUI/TextDocument.cpp
@@ -427,6 +427,54 @@ Optional<TextDocumentSpan> TextDocument::first_non_skippable_span_after(const Te
     return {};
 }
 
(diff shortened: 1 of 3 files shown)


alimpfard:
:thinking: does this also apply to the location bar in the browser?

FalseHonesty:
Yes absolutely! I fixed up selection in #2335, and now this helps navigation :)

awesomekling:
Fricken nice!

PR #2355, opened 2020-05-24
LibWeb: HtmlTokenizer.cpp: fix ON_WHITESPACE macro

emanuele6:
We were treating the "audible bell" character ('\a' U+0007) as whitespace instead of treating the "line feed" character ('\n' U+0000a) as whitespace.


awesomekling:
Hah, that's an interesting typo! I guess because `0x0a` -> `'\a'` :)

PR #2356, opened 2020-05-24
LibJS: add Array.prototype.reduceRight()

nooga:
This patch adds `Array.prototype.reduceRight()` method to LibJS Runtime. The implementation is (to my best knowledge) conformant to https://tc39.es/ecma262/#sec-array.prototype.reduceright.

Short test in `LibJS/Tests/Array.prototype-generic-functions.js` demonstrates that the function can be applied to other objects besides `Array`.

### DIFF
--- a/Libraries/LibJS/Runtime/ArrayPrototype.cpp
+++ b/Libraries/LibJS/Runtime/ArrayPrototype.cpp
@@ -59,6 +59,7 @@ ArrayPrototype::ArrayPrototype()
     put_native_function("slice", slice, 2, attr);
     put_native_function("indexOf", index_of, 1, attr);
     put_native_function("reduce", reduce, 1, attr);
+    put_native_function("reduceRight", reduce_right, 1, attr);
     put_native_function("reverse", reverse, 0, attr);
     put_native_function("lastIndexOf", last_index_of, 1, attr);
     put_native_function("includes", includes, 1, attr);
@@ -398,6 +399,10 @@ Value ArrayPrototype::reduce(Interpreter& interpreter)
     if (interpreter.exception())
         return {};
 
+    auto* callback_function = callback_from_args(interpreter, "reduce");
+    if (!callback_function)
+        return {};
+
     size_t start = 0;
 
     auto accumulator = js_undefined();
@@ -420,10 +425,6 @@ Value ArrayPrototype::reduce(Interpreter& interpreter)
         }
     }
 
-    auto* callback_function = callback_from_args(interpreter, "reduce");
-    if (!callback_function)
-        return {};
-
     auto this_value = js_undefined();
 
     for (size_t i = start; i < initial_length; ++i) {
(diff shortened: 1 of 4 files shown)


Libraries/LibJS/Runtime/ArrayPrototype.cpp
```diff
@@ -447,6 +448,65 @@ Value ArrayPrototype::reduce(Interpreter& interpreter)
     return accumulator;
 }
 
+Value ArrayPrototype::reduceRight(Interpreter& interpreter)
```
linusg, line 451: Please follow the [Serenity C++ coding style](https://github.com/SerenityOS/serenity/blob/master/Documentation/CodingStyle.md#names) and use snake_case for function names.

Libraries/LibJS/Runtime/ArrayPrototype.cpp
```diff
@@ -447,6 +448,65 @@ Value ArrayPrototype::reduce(Interpreter& interpreter)
...
+        }
+        if (!start_found) {
+            interpreter.throw_exception<TypeError>("Reduce of empty array with no initial value");
+            return {};
+        }
+    }
+
+    auto* callback_function = callback_from_args(interpreter, "reduceRight");
+    if (!callback_function)
+        return {};
```
linusg, line 485: Can we move this (and probably also the one in `reduce` after the `get_length()` call but before the if/else? Not that it *really* matters as both are type errors, but that way the "missing/invalid callback" error is thrown before the "Reduce of empty array with no initial value" one, which is closer to the spec and other implementations :)

Libraries/LibJS/Runtime/ArrayPrototype.cpp
```diff
@@ -447,6 +448,65 @@ Value ArrayPrototype::reduce(Interpreter& interpreter)
...
+        auto value = this_object->get_by_index(i);
+        if (interpreter.exception())
+            return {};
+        if (value.is_empty())
+            continue;
+
+        MarkedValueList arguments(interpreter.heap());
+        arguments.append(accumulator);
+        arguments.append(value);
+        arguments.append(Value((i32)i));
```
linusg, line 499: Redundant cast - I guess ideally we would make `i` a `size_t` but as `start` can be -1 that could be ugly...

Libraries/LibJS/Runtime/ArrayPrototype.cpp
nooga: sorry! force of habit :)

Libraries/LibJS/Runtime/ArrayPrototype.cpp
nooga: Indeed, I kept it as `size_t` initially and it immediately became very ugly ;)

linusg:
@awesomekling ninja-merge at your discretion!

PR #2358, opened 2020-05-24
LibWeb: Copy paste error in HTMLDockumentParser

Danten:
When watching the video of the new html parser I noticed a small copy and paste error. In one of the cases in `handle_after_head` the code was checking for end tags when it should check for start tags.

I haven't tested this change just looking at the spec: https://html.spec.whatwg.org/multipage/parsing.html#the-after-head-insertion-mode


awesomekling:
Good catch! Thank you :)

PR #2360, opened 2020-05-24
LibJS: Refactor Accessor

linusg:
This changes `Accessor`'s `m_{getter,setter}` from `Value` to `Function*` which seems like a better API to me - a getter/setter must either be a function or missing, and the creation of an accessor with other values must be prevented by the parser and `Object.defineProperty()` anyway.

Also add `Accessor::set_{getter,setter}()` so we can reuse an already created accessor when evaluating an ObjectExpression with getter/setter shorthand syntax.

---

I assume `Value` was initially chosen as the parser (incorrectly) allows `{ get foo: bar }`.


awesomekling:
Yup, tighter typing is always nicer :)

PR #2361, opened 2020-05-24
Shell: Add the `disown' shell builtin; Parse and correctly execute "command &" and "command &&" 

alimpfard:
Adding one more job control primitive to the mix, and fixing a little oversight in the shell.
Also adds some nice `&` and `&&` handling; `te& disown` ftw!

### DIFF
--- a/Shell/Shell.cpp
+++ b/Shell/Shell.cpp
@@ -1052,8 +1052,7 @@ IterationDecision Shell::wait_for_pid(const Shell::SpawnedProcess& process, bool
         if (WEXITSTATUS(wstatus) != 0)
             dbg() << "Shell: " << process.name << ":" << process.pid << " exited with status " << WEXITSTATUS(wstatus);
 
-        if (is_first_command_in_chain)
-            return_value = WEXITSTATUS(wstatus);
+        return_value = WEXITSTATUS(wstatus);
 
         if (job) {
             auto* mutable_job = const_cast<Job*>(job);
@@ -1128,6 +1127,7 @@ ExitCodeOrContinuationRequest Shell::run_command(const StringView& cmd)
                     break;
                 }
             }
+
             dbgprintf("\n");
             for (auto& redirecton : command.subcommands[i].redirections) {
                 for (size_t j = 0; j < i; ++j)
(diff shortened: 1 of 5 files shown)


Shell/Shell.cpp
```diff
@@ -442,6 +442,65 @@ int Shell::builtin_fg(int argc, const char** argv)
...
+        // An error message has already been printed about the nonexistence of each listed job.
+        return 1;
+    }
+    for (auto job_index : keys_of_jobs_to_disown) {
+        auto job = jobs.get(job_index).value();
+
+        job->deactivate();
+
+        if (!job->is_running_in_background())
+            printf("disown warning: job %llu is currently not running, 'kill -%d %d' to make it continue\n", job->job_id(), SIGCONT, job->pid());
```
alimpfard, line 497: I _really_ want this "`kill -%d %d`" to be a link, so let's talk specifics about how this would work (summoning @bugaevc)

- the command _must_ be run in the invoking shell
- it would be nice to get its output
- writing the command into the stdin of the shell is not particularly wise

I think that json-escape-sequence-xpc is gonna have to come sooner than later...

alimpfard:
I just realised that this would not work with `false && foo; bar`.
let's fix that first.

Shell/Shell.cpp
bugaevc: XPC?? nonononono

alimpfard: > writing the command into the stdin of the shell is not particularly wise

maybe it is...maybe clicking a command link should just write it to stdin, and the user should be expected to know this?

alimpfard: > nonononono

lol

bugaevc: You meant IPC, RPC, or what? :slightly_smiling_face:

alimpfard: I legit meant cross-process communication there.
it waz bait to get you here :)

alimpfard: apparently that's an apple-specific term?
I dunno, I've only had use of it with objc, so I called it xpc

alimpfard: aaaaaaaaaaaanyway, should we just make clicking a link to a command write it to the stdin of the shell (plus a newline)?

alimpfard: shot-2020-05-25_00-38-10

:thinking:

Shell/Shell.cpp
```diff
@@ -337,6 +337,7 @@ int Shell::builtin_dirs(int argc, const char** argv)
 
 int Shell::builtin_exit(int, const char**)
 {
+    stop_all_jobs();
```
awesomekling, line 340: Most shells I've used don't exit at first if you have active jobs. Instead they say something like "hey dude u have active jobs". To exit with active jobs, you have to type exit twice.

I think we should do something similar :)

alimpfard: Sure, I'll have it say "hey dude u have %d active jobs" :^)

PR #2362, opened 2020-05-24
Eyes

bugaevc:
It's not a Unix if it doesn't have xeyes

But that doesn't mean you have to run X11 😉

image

This is a neat clone of xeyes. It demoes the new ellipse filling functionality, as well as the existing global mouse cursor tracking functionality :^)

### DIFF
--- a/Demos/Eyes/EyesWidget.cpp
+++ b/Demos/Eyes/EyesWidget.cpp
@@ -1,0 +1,129 @@
+/*
+ * Copyright (c) 2020, Sergey Bugaev <bugaevc@serenityos.org>
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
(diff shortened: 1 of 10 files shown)


Libraries/LibM/math.h
```diff
@@ -37,6 +37,8 @@ __BEGIN_DECLS
 #define M_TAU (M_PI * 2)
 #define M_LN2 0.69314718055995
 #define M_LN10 2.30258509299405
+#define M_SQRT2 1.4142135623730951
+#define M_SQRT1_2 0.7071067811865475
```
linusg, line 41: Please update LibJS's `Math` object to use these:

https://github.com/SerenityOS/serenity/blob/9c8d390682fb7e1cf28aeb96975d86f45d61d441/Libraries/LibJS/Runtime/MathObject.cpp#L65-L66

bugaevc: Done, thanks :slightly_smiling_face:

alimpfard:
can we make the eyes look like buggie's eyes and call it `beyes` pls?

bugaevc:
Oh, and it can render an arbitrary number of eyes. Here's Mike Wazowski version:

image

bugaevc:
> can we make the eyes look like buggie's eyes and call it beyes pls?

That could be an option, or a separate program if you want :smile:

Demos/Eyes/EyesWidget.cpp
```diff
@@ -0,0 +1,104 @@
...
+    Gfx::Rect bounds { index * eye_width, 0, eye_width, height() };
+    auto thickness = max((width() + height()) / 15, 1);
+
+    bounds.shrink(width() / 25, 0);
+    painter.fill_ellipse(bounds, palette().base_text());
+    bounds.shrink(thickness, thickness);
+    painter.fill_ellipse(bounds, palette().base());
+
+    // OK, here comes a bunch of write-only math.
+    // You are not expected to understand this.
```
alimpfard, line 76: ...doesn't mean it has to look like a data scientist wrote it...can't we have more descriptive names than `ksq`? even `k_squared` would look better imho...

bugaevc: Honestly, `ksq` is the least of with the problems with it. I have no idea how to make math like this readable.

bugaevc: OK, let me try.

alimpfard: well the math part of it would have to stay, but at the very least, the names could say what you want, I assume "da" is "distance_to_a" or something?

awesomekling:
This is amazing and the name Eyes is great! :eyes:

Demos/Eyes/EyesWidget.cpp
bugaevc: Worse, _a_ was radius in one of the directions, and Dₐ was the corresponding diameter (twice the radius). Hold on, I'm trying to translate this to Serenity-style naming, and fix a division by zero.

alimpfard: :facepalm: 
only robotics people call radii _a_, that's so silly...

bugaevc: Well, the ellipse equation is



(as seen even on Wikipedia)

alimpfard: `git commit -m suicide`

bugaevc: Should be more readable now.

alimpfard: Much better!

PR #2367, opened 2020-05-25
Browser: Add output styles to values printed in the console

FalseHonesty:
This patch adds spans around most of the console's output, allowing for a global document stylesheet to customize the highlighting of the console's output. It also adds some basic styling for values like strings, numbers, and arrays using the system Palette.

Note: This patch simply adds support for highlighting output values, the lines of JS code printed to console are still unformatted.

This also includes some super basic QOL things to the console, primarily auto-scrolling to the bottom when printing.

Obligatory image :)


awesomekling:
Spicy :)

PR #2369, opened 2020-05-25
Misc fixes, pt. 4

bugaevc:
cc @alimpfard for the waiting fix

### DIFF
--- a/Services/WindowServer/WindowManager.cpp
+++ b/Services/WindowServer/WindowManager.cpp
@@ -761,10 +761,8 @@ void WindowManager::process_mouse_event(MouseEvent& event, Window*& hovered_wind
     HashTable<Window*> windows_who_received_mouse_event_due_to_cursor_tracking;
 
     for (auto* window = m_windows_in_order.tail(); window; window = window->prev()) {
-        if (!window->global_cursor_tracking())
+        if (!window->global_cursor_tracking() || !window->is_visible() || window->is_minimized())
             continue;
-        ASSERT(window->is_visible());    // Maybe this should be supported? Idk. Let's catch it and think about it later.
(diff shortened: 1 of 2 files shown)


Services/WindowServer/WindowManager.cpp
```diff
@@ -761,10 +761,8 @@ void WindowManager::process_mouse_event(MouseEvent& event, Window*& hovered_wind
     HashTable<Window*> windows_who_received_mouse_event_due_to_cursor_tracking;
 
     for (auto* window = m_windows_in_order.tail(); window; window = window->prev()) {
-        if (!window->global_cursor_tracking())
+        if (!window->global_cursor_tracking() || !window->is_visible() || window->is_minimized())
             continue;
-        ASSERT(window->is_visible());    // Maybe this should be supported? Idk. Let's catch it and think about it later.
-        ASSERT(!window->is_minimized()); // Maybe this should also be supported? Idk.
```
awesomekling, line 767: I love the confidence in these comments

alimpfard:
@bugaevc 
How would this work for `waitpid(-1, WNOHANG)`?

alimpfard:
Would we still get _a_ child pid?

bugaevc:
This does work for `waitpid(-1, WNOHANG)`, and we do get the correct `child_pid`, at least in case there was somebody to wait for (see SystemServer). If it returns because there's nobody to wait for, it will *probably* result in `-ECHILD` instead of the proper zero return code & zero siginfo. That's something further to investigate & fix, if you want to work on that.

alimpfard:
Alright.
I'll that to my TODO list, but I don't think it's a pressing issue.

PR #2372, opened 2020-05-25
LibLine: Change get_line to return a Result<String, Error>

### LINKED ISSUE #2370: Shell busyloops when the PTY goes away
To reproduce:
* `killall Terminal`
* Observe the CPU usage growing to 100%
* Perhaps use this opportunity to warm your hands a bit

@alimpfard

alimpfard:
This fixes a bunch of FIXME's in LibLine.
Also handles the case where read() would read zero bytes in vt_dsr() and effectively block forever by erroring out.

Also disallows `^Z` in shell, because that was annoying.

Fixes #2370

Morale of this story: I do not Unix enough.

### DIFF
--- a/Applications/Terminal/main.cpp
+++ b/Applications/Terminal/main.cpp
@@ -57,6 +57,12 @@
 static void run_command(int ptm_fd, String command)
 {
     pid_t pid = fork();
+    if (pid < 0) {
+        perror("fork");
+        dbg() << "run_command: could not fork to run '" << command << "'";
+        return;
+    }
+
     if (pid == 0) {
         const char* tty_name = ptsname(ptm_fd);
         if (!tty_name) {
--- a/Libraries/LibLine/Editor.cpp
+++ b/Libraries/LibLine/Editor.cpp
@@ -224,7 +224,7 @@ void Editor::suggest(size_t invariant_offset, size_t static_offset, Span::Mode o
     m_suggestion_manager.set_suggestion_variants(internal_static_offset, internal_invariant_offset, 0);
 }
 
-String Editor::get_line(const String& prompt)
+Result<String, Editor::Error> Editor::get_line(const String& prompt)
 {
     initialize();
     m_is_editing = true;
@@ -239,6 +239,9 @@ String Editor::get_line(const String& prompt)
         if (m_always_refresh)
             m_refresh_needed = true;
         refresh_display();
+        if (m_input_error.has_value()) {
+            return m_input_error.value();
+        }
         if (m_finish) {
             m_finish = false;
             printf("\n");
@@ -279,17 +282,18 @@ String Editor::get_line(const String& prompt)
                 m_refresh_needed = true;
                 continue;
             }
+
+            ScopedValueRollback errno_restorer(errno);
             perror("read failed");
-            // FIXME: exit()ing here is a bit off. Should communicate failure to caller somehow instead.
-            exit(2);
+
+            return Error::ReadFailure;
         }
 
         m_incomplete_data.append(keybuf, nread);
         nread = m_incomplete_data.size();
 
-        // FIXME: exit()ing here is a bit off. Should communicate failure to caller somehow instead.
         if (nread == 0)
-            exit(0);
+            return Error::Empty;
 
         auto reverse_tab = false;
         auto ctrl_held = false;
@@ -710,11 +714,18 @@ String Editor::get_line(const String& prompt)
                     fflush(stdout);
 
                     auto search_prompt = "\x1b[32msearch:\x1b[0m ";
-                    auto search_string = m_search_editor->get_line(search_prompt);
+                    auto search_string_result = m_search_editor->get_line(search_prompt);
 
                     m_search_editor = nullptr;
                     m_is_searching = false;
                     m_search_offset = 0;
+
+                    if (search_string_result.is_error()) {
+                        // Somethine broke, fail
+                        return search_string_result;
+                    }
+
+                    auto& search_string = search_string_result.value();
 
                     // Manually cleanup the search line.
                     reposition_cursor();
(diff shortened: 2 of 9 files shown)


bugaevc:
Somebody forgot to cc me :wink:

alimpfard:
Curses.
I _always_ forget dammit.

I also do not memory, apparently.

bugaevc:
> Curses.

ncurses, you surely mean

Applications/Terminal/main.cpp
```diff
@@ -57,6 +57,7 @@
 static void run_command(int ptm_fd, String command)
 {
     pid_t pid = fork();
+
```
bugaevc, line 60: I won't complain about this extra newline that has nothing to do with this commit if you add proper handling for the `pid < 0` case :smile:

alimpfard:
> ncurses, you surely mean

I'm all about the original version :^)

Libraries/LibLine/Editor.cpp
```diff
@@ -279,17 +282,25 @@ String Editor::get_line(const String& prompt)
                 m_refresh_needed = true;
                 continue;
             }
+
+            auto errno_save = errno;
```
bugaevc, line 286: Usually called `saved_erro`

Applications/Terminal/main.cpp
alimpfard: what if I just remove the extra newline :thinking:

Libraries/LibLine/Editor.cpp
```diff
@@ -279,17 +282,25 @@ String Editor::get_line(const String& prompt)
...
         }
 
         m_incomplete_data.append(keybuf, nread);
         nread = m_incomplete_data.size();
 
-        // FIXME: exit()ing here is a bit off. Should communicate failure to caller somehow instead.
-        if (nread == 0)
-            exit(0);
+        if (nread == 0) {
+            dbg() << "Read failed (no chars)";
```
bugaevc, line 301: EOF is not a failure, it's just an EOF

bugaevc:
So, have you figured out why it doesn't receive a SIGHUP, or?..

alimpfard:
> So, have you figured out why it doesn't receive a SIGHUP, or?..

Nope. I don't think SIGHUP is being sent at all here, killall sends a SIGTERM and...nothing else happens

Libraries/LibLine/Editor.cpp
alimpfard: There's no reason to log there anyway

Applications/Terminal/main.cpp
```diff
@@ -124,207 +130,206 @@ static void run_command(int ptm_fd, String command)
...
-    return window;
-}
-
-int main(int argc, char** argv)
-{
-    if (pledge("stdio tty rpath accept cpath wpath shared_buffer proc exec unix fattr", nullptr) < 0) {
-        perror("pledge");
-        return 1;
+    RefPtr<GUI::Window> create_settings_window(TerminalWidget & terminal)
+    {
```
bugaevc, line 135: Ooof

alimpfard: lol.
saw your other nitpick and forgot about that

Libraries/LibLine/Editor.cpp
```diff
@@ -279,17 +282,23 @@ String Editor::get_line(const String& prompt)
...
+
             perror("read failed");
-            // FIXME: exit()ing here is a bit off. Should communicate failure to caller somehow instead.
-            exit(2);
+            dbg() << "Read failed: " << strerror(saved_errno);
+
+            if (saved_errno == EINVAL || saved_errno == EBADF)
+                return Error::ReadFailure;
+
+            return Error::Unspecified;
```
bugaevc, line 294: Why single out (double out?) those two error codes?

alimpfard: because they're the only other error codes that `read()` can fail with in this scenario?

alimpfard: To my (admittedly shit memory), read can fail with BADF, INTR, INVAL, FAULT and IO.
we're not doing any disk IO, so EIO shouldn't come out at all.
our buffer on the stack so FAULT is out too
INTR is handled, leaving the other two

bugaevc: `read()` can fail for way more reasons than you can ever imagine. Any device driver or any magical fine can return whatever errors it can think of.

> we're not doing any disk IO

Aren't we?

alimpfard: we're reading from `stdin`...
if the drivers spit out something weird, that's just an unspecified error...
I guess there's no need for `Unspecified` then, I'll drop that.

bugaevc: > we're reading from `stdin`...

Aka whatever happens to be fd 0

alimpfard: > Aka whatever happens to be fd 0

We're working on a level of abstraction here, LibLine assumes some stuff about that...I don't think worrying about things we don't consider legitimate read errors is that important.
Nonetheless, every error except INTR is just a read error now.

Libraries/LibLine/Editor.cpp
```diff
@@ -279,17 +282,20 @@ String Editor::get_line(const String& prompt)
                 m_refresh_needed = true;
                 continue;
             }
+
+            auto saved_errno = errno;
+
             perror("read failed");
-            // FIXME: exit()ing here is a bit off. Should communicate failure to caller somehow instead.
-            exit(2);
+            dbg() << "Read failed: " << strerror(saved_errno);
```
bugaevc, line 289: `perror` should already print to `dbg()` log itself, so no need for `saved_errno` at all.

Unless you want to preserve `errno` for the caller; then use `ScopedValueRollback errno_restorer(errno);`

alimpfard: Very well.
please tell me I didn't forget to put errno back...I did...

Shell/main.cpp
```diff
@@ -104,6 +104,9 @@ int main(int argc, char** argv)
         }
     });
 
+    // Ignore SIGTSTP as the shell should not be suspended with ^Z.
+    signal(SIGTSTP, [](auto) {});
```
bugaevc, line 108: Why not `SIG_IGN` it?

PR #2375, opened 2020-05-25
Browser: Add output styles to JS source printed in the console

FalseHonesty:
This patch uses the JS lexer to add spans around all of the source code printed in the console's output panel. These spans all have corresponding class names, and thus are styled by the global document style sheet.

Can't forget the image!


linusg:
Let's not do this in the console widget but somewhere else - that way we'd get JS syntax highlighting to be embedded in HTML for free, which could be useful for TextEditor's Markdown preview functionality.

FalseHonesty:
> Let's not do this in the console widget but somewhere else - that way we'd get JS syntax highlighting to be embedded in HTML for free, which could be useful for TextEditor's Markdown preview functionality.

I was considering this, do you have a suggestion for where this should live?

alimpfard:
`<span style="color:red">` lol

FalseHonesty:
> `<span style="color:red">` lol

Yeah that's totally possible, but it should respect the system palette. I guess if there was some way to communicate directly to LibWeb that a color should be straight from the system palette that would work, but I'm not sure how that would work. Something like `color: palette(syntax-operator)`? I feel like that might be going off the rails a little.

alimpfard:
I'm not in the loop about libweb, but do we really want to add extensions to css already?

FalseHonesty:
> I'm not in the loop about libweb, but do we really want to add extensions to css already?

Yeah that's what I'm saying, that is getting pretty iffy. I'm just spitballing ideas on how to make it respect the system palette.

awesomekling:
Nothing wrong with CSS extensions for system integration IMO :)
Note that CSS has "system colors" already (although they are deprecated.)
The right way to do this IMO is basic prefixed values, i.e `color: -libweb-palette-syntax-operator`

alimpfard:
I'm not entirely sure what the general consensus around standard extensions is, but in the little bit of the system I have taken up, extensions are a total mess.
Pardon my ignorance about this matter in web standards

FalseHonesty:
> Nothing wrong with CSS extensions for system integration :)
> Note that CSS has "system colors" already (although they are deprecated.)
> The right way to do this IMO is basic prefixed values, i.e `color: -libweb-palette-syntax-operator`

Ah, interesting, I'll have to look into the CSS engine then I suppose! Do you have a suggestion for where would it make sense to place the JS source code -> syntax highlighted HTML conversion?

alimpfard:
Does `<pre>` have a standard language tag? If so, that could be a decent place, I think?

awesomekling:
> > Nothing wrong with CSS extensions for system integration :)
> > Note that CSS has "system colors" already (although they are deprecated.)
> > The right way to do this IMO is basic prefixed values, i.e `color: -libweb-palette-syntax-operator`
> 
> Ah, interesting, I'll have to look into the CSS engine then I suppose! Do you have a suggestion for where would it make sense to place the JS source code -> syntax highlighted HTML conversion?

Sure, you can implement it similarly to how `-libweb-link` is implemented (used for link colors.)
Basically, make `parse_css_value()` produce an `IdentifierStyleValue` with something like `CSS::ValueID::VendorSpecificPaletteSyntaxOperator` for `-libweb-palette-syntax-operator`.

When resolving an `IdentifierStyleValue` to a color for rendering, the engine will call `IdentifierStyleValue::to_color(Document&)`, so that's where you can fetch the appropriate palette color depending on the `CSS::ValueID`.

The shortest way to the palette from a `Document` is `document.frame()->html_view()->palette()`. :)

awesomekling:
> Do you have a suggestion for where would it make sense to place the JS source code -> syntax highlighted HTML conversion?

Hmm. I don't know. Maybe somewhere in LibJS? It's essentially about taking JS and turning into a string of HTML right?

FalseHonesty:
> Does `<pre>` have a standard language tag? If so, that could be a decent place, I think?

Well, it doesn't look like it does. [This](https://stackoverflow.com/questions/5134242/semantics-standards-and-using-the-lang-attribute-for-source-code-in-markup) StackOverflow article goes pretty in-depth on the matter.

> Sure, you can implement it similarly to how `-libweb-link` is implemented (used for link colors.)
> Basically, make `parse_css_value()` produce an `IdentifierStyleValue` with something like `CSS::ValueID::VendorSpecificPaletteSyntaxOperator` for `-libweb-palette-syntax-operator`.
> 
> When resolving an `IdentifierStyleValue` to a color for rendering, the engine will call `IdentifierStyleValue::to_color(Document&)`, so that's where you can fetch the appropriate palette color depending on the `CSS::ValueID`.
> 
> The shortest way to the palette from a `Document` is `document.frame()->html_view()->palette()`. :)

Ah, very helpful information, thank you for that!



> > Do you have a suggestion for where would it make sense to place the JS source code -> syntax highlighted HTML conversion?
> 
> Hmm. I don't know. Maybe somewhere in LibJS? It's essentially about taking JS and turning into a string of HTML right?

Yeah, I think that might be the way to go. I haven't really delved into LibJS either, do you have a preferred place for this to go? Should it be a member somewhere, or perhaps a top level method in the `Lexer.cpp` file?

awesomekling:
> > > Do you have a suggestion for where would it make sense to place the JS source code -> syntax highlighted HTML conversion?
> > 
> > 
> > Hmm. I don't know. Maybe somewhere in LibJS? It's essentially about taking JS and turning into a string of HTML right?
> 
> Yeah, I think that might be the way to go. I haven't really delved into LibJS either, do you have a preferred place for this to go? Should it be a member somewhere, or perhaps a top level method in the `Lexer.cpp` file?

A separate top level function in its own file would work. `MarkupGenerator.cpp` or something like that. :)

FalseHonesty:
> A separate top level function in its own file would work. `MarkupGenerator.cpp` or something like that. :)

Sounds great, I'll get to work!

FalseHonesty:
> Sure, you can implement it similarly to how -libweb-link is implemented (used for link colors.)
> Basically, make parse_css_value() produce an IdentifierStyleValue with something like CSS::ValueID::VendorSpecificPaletteSyntaxOperator for -libweb-palette-syntax-operator.

In reference to this @awesomekling, the `-syntax-operator` suffix would be specific to the Palette's `syntax_operator` method. It doesn't make sense to have a specific ValueID for every single Palette color right? Instead, would it make more sense to have a custom subclass of `StyleValue` called `PaletteStyleValue` that can store its own value of the `ColorRole` enum corresponding to the text following `-libweb-palette-`?

awesomekling:
> > Sure, you can implement it similarly to how -libweb-link is implemented (used for link colors.)
> > Basically, make parse_css_value() produce an IdentifierStyleValue with something like CSS::ValueID::VendorSpecificPaletteSyntaxOperator for -libweb-palette-syntax-operator.
> 
> In reference to this @awesomekling, the `-syntax-operator` suffix would be specific to the Palette's `syntax_operator` method. It doesn't make sense to have a specific ValueID for every single Palette color right?

It does make sense! A `ValueID` is a possible CSS identifier value, and all of the `-libweb-palette-syntax-foo` variants are possible CSS identifier values :)

FalseHonesty:
> It does make sense! A `ValueID` is a possible CSS identifier value, and all of the `-libweb-palette-syntax-foo` variants are possible CSS identifier values :)

Oh interesting, that's going to be a big enum then I suppose! It also means I'd have to re-do the `ValueID` -> `ColorRole` lookup every time, but that's just going to be a switch, probably not too big of a deal at all. Thanks :)

FalseHonesty:
With these newest commits, there is now a `MarkupGenerator` in LibJS that can be used to do the JS source to HTML conversion, and the JS Value to HTML conversion, as well as vendor specific styles in css for the system palette.

The output will still be wonky until #2377 is resolved.

Image taken from after the refactor, looking the exact same :)

awesomekling:
Cool!

bugaevc:
I would say,
* `-libweb-link` can now be replaced by `-libweb-palette-link`
* It should be a color function after all, `-libweb-palette(active-window-border2)` looks better to me than `-libweb-palette-active-window-border2`

awesomekling:
`-libweb-link` is different from `-libweb-palette-link`, `-libweb-link` is for implementing the `<body link>` attribute :)

PR #2378, opened 2020-05-25
Eyes: Correctly calculate the eye bounds for arbitrary number of eyes

alimpfard:
cc @bugaevc I WILL NOT FORGET TO CC

Old:
shot-2020-05-25_21-04-09

New (xeyes for comparison)
shot-2020-05-25_21-06-09

### DIFF
--- a/Demos/Eyes/EyesWidget.cpp
+++ b/Demos/Eyes/EyesWidget.cpp
@@ -65,11 +65,12 @@ void EyesWidget::render_eyeball(int index, GUI::Painter& painter) const
 {
     auto eye_width = width() / m_num_eyes;
     Gfx::Rect bounds { index * eye_width, 0, eye_width, height() };
-    auto thickness = max((width() + height()) / 15, 1);
+    auto width_thickness = max(int(eye_width / 5.5), 1);
+    auto height_thickness = max(int(height() / 5.5), 1);
 
-    bounds.shrink(width() / 25, 0);
+    bounds.shrink(int(eye_width / 12.5), 0);
     painter.fill_ellipse(bounds, palette().base_text());
-    bounds.shrink(thickness, thickness);
+    bounds.shrink(width_thickness, height_thickness);
     painter.fill_ellipse(bounds, palette().base());
 
     Gfx::Point pupil_center = this->pupil_center(bounds);
--- a/Demos/Eyes/main.cpp
+++ b/Demos/Eyes/main.cpp
@@ -51,7 +51,7 @@ int main(int argc, char* argv[])
 
     auto window = GUI::Window::construct();
     window->set_title("Eyes");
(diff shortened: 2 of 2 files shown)


Demos/Eyes/EyesWidget.cpp
```diff
@@ -105,17 +106,11 @@ Gfx::Point EyesWidget::pupil_center(Gfx::Rect& eyeball_bounds) const
...
-        );
+        max_distance_along_this_direction = 0.25 * sqrt((slope_squared + 1) / (1 / width_squared + slope_squared / height_squared));
     } else if (dy != 0 && abs(dy) >= abs(dx)) {
         double slope = dx / dy;
         double slope_squared = slope * slope;
-        max_distance_along_this_direction = 0.25 * sqrt(
-            (slope_squared + 1) /
-            (slope_squared / width_squared + 1 / height_squared)
-        );
+        max_distance_along_this_direction = 0.25 * sqrt((slope_squared + 1) / (slope_squared / width_squared + 1 / height_squared));
```
alimpfard, line 113: why.
@bugaevc why.

bugaevc: Because this is way longer than 80 characters, not quite readable, and clang-format apparently doesn't want to care. I'd leave it as is.

alimpfard: I'm not touching this.
Not my problem, run clang-format at your own risk.

Demos/Eyes/EyesWidget.cpp
```diff
@@ -65,11 +65,12 @@ void EyesWidget::render_eyeball(int index, GUI::Painter& painter) const
 {
     auto eye_width = width() / m_num_eyes;
     Gfx::Rect bounds { index * eye_width, 0, eye_width, height() };
-    auto thickness = max((width() + height()) / 15, 1);
+    auto height_thickness = max(int(height() / 5.5), 1);
+    auto width_thickness = max(int(eye_width / 5.5), 1);
```
bugaevc, line 69: Just one tiny nit: let's put width before height. (Swap the two above lines.)

alimpfard: Okies.

Demos/Eyes/EyesWidget.cpp
alimpfard: at least `// clang-format {off,on}` so people don't have to mess with `git add -p`...

bugaevc:
Looks & works great!

I have one more request (in addition to the nit): could we size the window on startup according to the number of eyes?

For two eyes, we have to stay at 150×100, but maybe we can do `75 • num_eyes` × 100 or something

alimpfard:
Hmmm...Sure, I guess, why not.

PR #2383, opened 2020-05-25
LibLine: Don't add empty lines to history

linusg:


Libraries/LibLine/Editor.cpp
```diff
@@ -63,6 +63,8 @@ Editor::~Editor()
 
 void Editor::add_to_history(const String& line)
 {
+    if (!line.length())
```
awesomekling, line 66: Let's say `if (line.is_empty())` for that nice read-out-loud feel :)

PR #2384, opened 2020-05-26
[WIP] LibEmail: Added classes for receiving/parsing IMAP email messages

csb6:
Right now, there is support for the IMAP commands FETCH, RENAME,
CREATE, DELETE, SELECT, and LOGIN, with the ability to fetch the
header and body of plaintext emails over an unencrypted TCP connection.

I am currently developing on macOS, so in its current state the code isn't compiling, since I was forced to develop it natively on macOS initially and have only partly ported it back to Serenity. Please forgive its current ugliness/incompleteness :)

The scope of this pull request is adding a new library, LibEmail, with a couple of classes adding support for the IMAP protocol and the representation of email messages.

I will continue to add commits as I fix up the code to run properly on Serenity; I just wanted to make a WIP pull request to show where I am at with the IMAP support. Thanks for your feedback!


Libraries/LibEmail/IMAPClient.cpp
```diff
@@ -0,0 +1,284 @@
...
+        dbg() << "Connected successfully";
+    else
+        dbg() << "Failed to connect";
+
+    // Setup thread for receiving/queuing server responses
+    m_receive_thread = LibThread::Thread::construct(
+        [this] {
+            while (true) {
+                const std::string response { m_socket->receive(1000) };
+                if (response.empty())
```
emanuele6, line 58: please, use curly braces here:
```cpp
if (response.empty()) {
    // For some reason servers sometimes send empty messages
    continue;
}
```

Libraries/LibEmail/Message.cpp
```diff
@@ -0,0 +1,398 @@
...
+                --i;
+            } else if (source[i] == '(') {
+                set_stack.append('(');
+                curr_state = State::InSet;
+            }
+            break;
+        case State::InCommand:
+            if (source[i] != ' ') {
+                curr_value.append(source[i]);
+            } else {
```
emanuele6, line 54: Instead of doing this:
```cpp
if (...) {
    ...
} else {
    ...
}
break;
```
You can do something like this:
```cpp
if (...) {
    ...
    break;
}
...
break;
```

Libraries/LibEmail/IMAPClient.h
```diff
@@ -0,0 +1,55 @@
+#pragma once
```
emanuele6, line 1: You need to add a licence header.

Libraries/LibEmail/IMAPClient.cpp
```diff
@@ -0,0 +1,284 @@
+#include <AK/StringBuilder.h>
```
emanuele6, line 1: You need to add a licence header.

Libraries/LibEmail/Message.cpp
```diff
@@ -0,0 +1,398 @@
+#include <AK/Optional.h>
```
emanuele6, line 1: You need to add a licence header.

Libraries/LibEmail/Message.h
```diff
@@ -0,0 +1,76 @@
+#pragma once
```
emanuele6, line 1: You need to add a licence header.

Libraries/LibEmail/IMAPClient.cpp
```diff
@@ -0,0 +1,284 @@
...
+    if (connected)
+        dbg() << "Connected successfully";
+    else
+        dbg() << "Failed to connect";
+
+    // Setup thread for receiving/queuing server responses
+    m_receive_thread = LibThread::Thread::construct(
+        [this] {
+            while (true) {
+                const std::string response { m_socket->receive(1000) };
```
emanuele6, line 57: How about implementing an `AK::String(AK::ByteBuffer&)` constructor instead of using `std::string`?

Libraries/LibEmail/IMAPClient.cpp
```diff
@@ -0,0 +1,284 @@
...
+                if (response.empty())
+                    // For some reason servers sometimes send empty messages
+                    continue;
+                m_queue_lock.lock();
+                m_message_queue.push_back(response);
+                m_queue_lock.unlock();
+            }
+            return 0;
+        });
```
emanuele6, line 66: This looks weird:
```cpp
        });
    m_receive_thread.start();
}
```
You could use this style instead:
```cpp
    m_receive_thread = LibThread::Thread::construct([this] {
        ...
    });
    m_receive_thread.start();
}
    
```

linusg: just do what clang-format is doing - and sometimes it is doing that.

emanuele6: Both styles are accepted by `clang-format`.

csb6:
Ok, I made the suggested changes and got the library compiling on Serenity. I will work on testing it out to make sure it is fully working!

Libraries/LibEmail/IMAPClient.cpp
emanuele6: .

csb6:
I am having some trouble getting the TCPSocket class working and connecting to localhost; does anyone have some experience with doing this successfully? I have a local IMAP server running on localhost, but I am not able to send/receive responses from within Serenity. Right now, I am using `0.0.0.0` as the socket address and port 143 as the port number, which works when I telnet into the IMAP server in my native OS. Is there any trick to bridging the network connection between Serenity and the native OS? Thanks for your help and feedback.

Libraries/LibEmail/IMAPClient.cpp
```diff
@@ -0,0 +1,320 @@
...
+{
+    if (m_state == ConnectionState::NotAuthenticated) {
+        dbg() << "Cannot create mailbox: improper current state";
+        return ResponseStatus::FailedToSend;
+    }
+
+    StringBuilder command;
+    command.append("create ");
+    command.append(mailbox);
+    bool status = send_command(command.string_view());
```
elcuco, line 164: I see this design you make.. and I was wondering if you can improve. 
You have a worker thread, to which you send a message, and then wait until you read a message (`receive_response()`). This will lock up the thread calling.

IMHO, you can design an async API - and to each of these methods, also pass a "callback" - which will be called from the thread you created in the constructor.

Libraries/LibEmail/IMAPClient.cpp
```diff
@@ -0,0 +1,320 @@
...
+                m_queue_lock.unlock();
+            }
+            return 0;
+        });
+    m_receive_thread->start();
+}
+
+IMAPClient::~IMAPClient()
+{
+    m_receive_thread->quit();
```
elcuco, line 99: How about changing the `while(true)` and doing a clean close? (asking also the `m_socket` to "abort" or something). Killing threads on non-managed languages is not a good idea.

Libraries/CMakeLists.txt
```diff
@@ -4,6 +4,7 @@ add_subdirectory(LibCore)
 add_subdirectory(LibCrypto)
 add_subdirectory(LibDebug)
 add_subdirectory(LibDesktop)
+add_subdirectory(LibEmail)
```
elcuco, line 7: `LibIMAP` ;-)

Libraries/LibEmail/IMAPClient.cpp
csb6: @elcuco I agree. Unfortunately it will require quite a bit of rearchitecting, so I haven't gotten around to it. The way it is architected now worked decently on macOS natively, but it seems like Serenity supports the asynchronous style a lot more naturally. Thanks for your feedback!

Libraries/CMakeLists.txt
csb6: @elcuco My intention was to reuse some of the classes I wrote for other email protocols (e.g. SMTP, POP3) in the future, so I'm not sure what to call it. I guess we can always rearrange the library namespaces if in the future more protocols are added. I'll change the name to `LibIMAP` for now to be safe.

csb6:
@awesomekling I don't think I am going to be able to finish this PR (not been able to build Serenity successfully on macOS and have less time to work on the library). If someone else wants to have a go, they should (Apologies if you're the wrong person to send this to). The code needs some refactoring to fit a more asynchronous model, where email messages can be sent/received using jobs across multiple threads. Right now, the code is mostly synchronous, and I couldn't get it working with Serenity's TCPSocket model.

The project is very cool and hope this code is useful to someone! Apologies for overextending myself here. Thanks.

awesomekling:
Closing as timed-out.

PR #2385, opened 2020-05-26
Kernel: Plumb KResult through FileDescription::read_entire_file() implementation

bgianfo:
Allow file system implementation to return meaningful error codes to
callers of the FileDescription::read_entire_file(). This allows both
Process::sys$readlink() and Process::sys$module_load() to return more
detailed errors to the user.


awesomekling:
:star:

PR #2387, opened 2020-05-26
AK: Add case insensitive String::ends_with support

bgianfo:
FileSystemPath::has_extension was jumping through hoops and allocating
memory to do a case insensitive comparison needlessly. Extend the
existing String::ends_with method to allow the caller to specify the
case sensitivity required.

### DIFF
--- a/AK/String.cpp
+++ b/AK/String.cpp
@@ -280,13 +280,7 @@ bool String::starts_with(char ch) const
 
 bool String::ends_with(const StringView& str) const
 {
-    if (str.is_empty())
-        return true;
-    if (is_empty())
-        return false;
-    if (str.length() > length())
-        return false;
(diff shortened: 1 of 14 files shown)


AK/String.cpp
```diff
@@ -278,8 +278,11 @@ bool String::starts_with(char ch) const
     return characters()[0] == ch;
 }
 
-bool String::ends_with(const StringView& str) const
+bool String::ends_with(const StringView& str, CaseSensitivity case_sensitivity) const
 {
+    if (case_sensitivity == CaseSensitivity::CaseInsensitive)
+        return StringUtils::ends_with_ignoring_case(*this, str);
```
awesomekling, line 284: Let's move the whole `ends_with()` to `StringUtils` so it can be used from `FlyString` and `StringView` as well :)

bgianfo: Sure, done.

awesomekling:
Awesome!

PR #2390, opened 2020-05-26
Kernel: fix assertion on readlink() syscall

anpep:
The is_error() check on the KResultOr returned when reading the link
target had a stray ! operator which causes link resolution to crash the
kernel with an assertion error.


bgianfo:
Thanks!

PR #2394, opened 2020-05-26
Almost reliable abort(), sigaction hardening, and LexicalPath

bugaevc:
With this, using `LexicalPath` to canonicalize (local) paths is considered deprecated: please let the kernel deal with it, as it can handle things like symlinks (and other filesystem magic) properly. `LexicalPath` is still useful for things like getting an extension of a file.

### DIFF
--- a/Libraries/LibC/unistd.h
+++ b/Libraries/LibC/unistd.h
@@ -72,7 +72,7 @@ int execlp(const char* filename, const char* arg, ...);
 int chroot(const char* path);
 int chroot_with_mount_flags(const char* path, int mount_flags);
 void sync();
-void _exit(int status);
+__attribute__((noreturn)) void _exit(int status);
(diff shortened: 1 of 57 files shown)


Libraries/LibC/unistd.h
```diff
@@ -72,7 +72,7 @@ int execlp(const char* filename, const char* arg, ...);
 int chroot(const char* path);
 int chroot_with_mount_flags(const char* path, int mount_flags);
 void sync();
-void _exit(int status);
+__attribute__((noreturn)) void _exit(int status);
```
awesomekling, line 75: Let's use the standard style `[[noreturn]]` :)

bugaevc: This is a C header.

awesomekling: Oops!

PR #2395, opened 2020-05-26
LibJS: Add Array.prototype.fill

anpep:
I open a new PR because LibJS has seen some changes since I last attempted to implement this, and I wanted a clean commit.

### DIFF
--- a/Libraries/LibJS/Runtime/ArrayPrototype.cpp
+++ b/Libraries/LibJS/Runtime/ArrayPrototype.cpp
@@ -68,6 +68,7 @@ ArrayPrototype::ArrayPrototype()
     put_native_function("some", some, 1, attr);
     put_native_function("every", every, 1, attr);
     put_native_function("splice", splice, 2, attr);
+    put_native_function("fill", fill, 1, attr);
     put("length", Value(0), Attribute::Configurable);
 }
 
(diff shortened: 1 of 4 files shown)


Libraries/LibJS/Runtime/ArrayPrototype.cpp
```diff
@@ -743,4 +744,48 @@ Value ArrayPrototype::splice(Interpreter& interpreter)
...
+    
+    ssize_t length = get_length(interpreter, *this_object);
+    if (interpreter.exception())
+        return {};
+    
+    Value value;
+    if (interpreter.argument_count() >= 1)
+        value = interpreter.argument(0);
+    else
+        value = js_undefined();
```
linusg, line 761: This is not necessary - if the argument at index `i` is missing, `interpreter.argument(i)` returns `undefined` anyway. So:

```suggestion
    auto value = interpreter.argument(0);
```

Libraries/LibJS/Runtime/ArrayPrototype.cpp
```diff
@@ -743,4 +744,48 @@ Value ArrayPrototype::splice(Interpreter& interpreter)
...
+    
+    ssize_t relative_start = 0;
+    ssize_t relative_end = length;
+    
+    if (interpreter.argument_count() >= 2)
+        relative_start = interpreter.argument(1).to_i32(interpreter);
```
linusg, line 767: Missing exception check

Libraries/LibJS/Runtime/ArrayPrototype.cpp
```diff
@@ -743,4 +744,48 @@ Value ArrayPrototype::splice(Interpreter& interpreter)
...
+    if (interpreter.argument_count() >= 3)
+        relative_end = interpreter.argument(2).to_i32(interpreter);
```
linusg, line 769: Missing exception check

Libraries/LibJS/Runtime/ArrayPrototype.cpp
```diff
@@ -743,4 +744,48 @@ Value ArrayPrototype::splice(Interpreter& interpreter)
...
+        
+    if (relative_end < 0)
+        to = max(length + relative_end, 0L);
+    else
+        to = min(relative_end, length);
+    
+    while (from < to) {
+        this_object->put_by_index(from, value);
+        from++;
+    }
```
linusg, line 786: Missing exception check - also this could be written as:

```cpp
for (size_t i = from; i < to; ++i) {
    this_object->put_by_index(i, value);
    if (interpreter.exception())
        return {};
}
```

Libraries/LibJS/Tests/Array.prototype.fill.js
```diff
@@ -0,0 +1,34 @@
+load("test-common.js");
+
+Array.prototype.equals = function(other) {
```
linusg, line 3: Please don't do this - write a helper function in `test-common.js` instead :)

Also strict equality is preferred whenever possible - this function would declare `[1, 2, 3]` and `["1", "2", "3"]` equal!

Libraries/LibJS/Runtime/ArrayPrototype.cpp
```diff
@@ -68,6 +68,7 @@ ArrayPrototype::ArrayPrototype()
     put_native_function("some", some, 1, attr);
     put_native_function("every", every, 1, attr);
     put_native_function("splice", splice, 2, attr);
+    put_native_function("fill", fill, 3, attr);
```
linusg, line 71: length is supposed to be 1 here

anpep:
I'm very grateful for your feedback, @linusg!
I hope this time everything checks out

linusg:
yes, looks good! One more small request, please squash your two commits as per the [contribution guidelines](https://github.com/SerenityOS/serenity/blob/master/CONTRIBUTING.md#code-submission-policy) :)

anpep:
There we go, that should do it

Issue #2273, opened 2020-05-18
LibGUI: A prettier About box


nvella:
Thought I'd get a conversation going on these ideas as I could see myself potentially taking on implementing them. Seeing as we have a system-wide About dialog in LibGUI, I think it could be a good idea if we gave it a good spit-and-polish to really present a kind of consistent branding between the built-in applications. 

Obligatory w2k reference;
![image](https://user-images.githubusercontent.com/8158648/82176068-05471c80-9919-11ea-9830-fa6253ef2f8d.png)

When viewing the About dialog for a system application, it includes the application's icon. (Apologies, couldn't find a period correct dialog for Notepad)
![image](https://user-images.githubusercontent.com/8158648/82177714-b51e8900-991d-11ea-81b0-cfaf53827fd0.png)

---

If we make the About dialog larger to potentially include a TabWidget, I'm thinking we could include information such as
- Version information, maybe by git hash
  - Also a 'git log' as a simple changelog perhaps?
- List of contributors generated at build-time from GitHub
- List of patreons/sponsors generated at build-time

It might also be a smart idea if we switch over the About application to use the LibGUI AboutDialog for the sake of consistency.

---

### TODO

- [x] Banner image designs
- [x] Alter `About` to use LibGUI AboutDialog
- [x] Redesign LibGUI AboutDialog
- [x] `/etc/version.ini` version information, generate on build filesystem step
- [x] AboutDialog - pull version information from /etc/version.ini

Dexesttp:
> If we make the About dialog larger to potentially include a TabWidget, I'm thinking we could include information such as
> 
> * Version information, maybe by git hash
>   * Also a 'git log' as a simple changelog perhaps?
> * List of contributors generated at build-time from GitHub
> * List of patreons/sponsors generated at build-time

I'm not a fan on making the build process linked to either the git folder or http(s) calls. For normal everyday builds (and especially for iterative development builds) that seems like an extra unwanted thing.

It would also force every build station to have both a working git folder (so no zip downloads) and a working internet connection (so no offline downloads), or at least require extra checks and fallbacks for these conditions.

However, that seems like an interesting feature to add if we get to the point where there's an automated "nightly" release :)

awesomekling:
Go for it @nvella!

> Obligatory w2k reference;

The w2k dialog is gorgeous IMO.

> If we make the About dialog larger to potentially include a TabWidget, I'm thinking we could include information such as
> 
> * Version information, maybe by git hash

[...]

PR #2397, opened 2020-05-26
LibGUI, About: A prettier About dialog

nvella:
Following on from the discussion in #2273, this PR contains the implementation for the prettier about dialog.

The version information (currently short Git hash) is written to `/res/version.ini` at build time (in `build-root-filesystem.sh`), which is then later displayed in the dialog.

Screenshots:
image

image


bugaevc:
> `/res/version.ini`

Should be `/ect/version.ini`

linusg:
> Should be /ect/version.ini

From #2273:

> Currently I'm writing the version information to /res/version.ini as part of the build-root-filesystem.sh script. I'm thinking /res is a good place for this file, as most graphical applications unveil it for their icons.

bugaevc:
Hmmm, but that's more of a workaround for not unveiling `/etc/` in every GUI app than an actual reason why it belongs in `/res`. ini config files is exactly the kind of thing that belongs into `/etc`, while resources (such as images) belong in `/res`.

If there's no good solution, we can indeed leave it in `/res`, but let's not pretend it's not a workaround and it belongs there :slightly_smiling_face:

nvella:
It definitely is a work-around 😅 I felt that `/res` was 'best-effort' for what will definitely be unveiled, but it's very much not preferable. It'd be cool if there was a 'guarantee' of sorts around certain 'safe' files which will always be unveiled - but these are definitely not decisions for me to make.

awesomekling:
Let's roll with it. We can tweak it as we go, if we think of improvements :)

Nice work @nvella :star:

PR #2398, opened 2020-05-26
LibLine: Core::EventLoop-ification; a few custom property additions to RPC inspections

alimpfard:
This PR switches LibLine to a Core::Object and its looping to Core::EventLoop.
Individual characters will still block, but that's much better than blocking for an entire line.

Also adds some inspection properties to LibLine, Shell, and ResourceLoader.

I tried making the Property display widget of Inspector a treeview, but due to the lack of a GUI _thing_ processor in my brain, I couldn't do so (In particular, I found `parent_index` annoying to implement, as we don't have a "JsonPath" or anything similar)

### DIFF
--- a/Shell/Shell.cpp
+++ b/Shell/Shell.cpp
@@ -1835,3 +1835,27 @@ void Shell::stop_all_jobs()
         }
     }
 }
+
+void Shell::save_to(JsonObject& object)
+{
+    Core::Object::save_to(object);
+    object.set("working_directory", cwd);
+    object.set("username", username);
+    object.set("user_home_path", home);
+    object.set("user_id", uid);
+    object.set("directory_stack_size", directory_stack.size());
+    object.set("cd_history_size", cd_history.size());
+
+    // Jobs.
+    JsonArray job_objects;
+    for (auto& job_entry : jobs) {
+        JsonObject job_object;
+        job_object.set("pid", job_entry.value->pid());
+        job_object.set("pgid", job_entry.value->pgid());
+        job_object.set("running_time", job_entry.value->timer().elapsed());
+        job_object.set("command", job_entry.value->cmd());
+        job_object.set("is_running_in_background", job_entry.value->is_running_in_background());
+        job_objects.append(move(job_object));
+    }
+    object.set("jobs", move(job_objects));
+}
--- a/Shell/Shell.cpp
+++ b/Shell/Shell.cpp
@@ -345,6 +345,7 @@ int Shell::builtin_exit(int, const char**)
         }
     }
     stop_all_jobs();
+    save_history();
     printf("Good-bye!\n");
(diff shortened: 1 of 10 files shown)


alimpfard:
> Please use variable-like names without spaces in save_to

There are so few of these I couldn't tell the convention. very well.

> why do we need to block on each character

well, if I understand `Core::Notifier` right, it already does a select.
making stdin nonblocking completely breaks everything (dsr in particular), so I opted to leave that for a future PR to deal with.

alimpfard:
...:facepalm:

@bugaevc come up with new pledges at a better time darn it

bugaevc:
> dsr

what's that? 🤔

alimpfard:
Device Status Report
tldr tells you the cursor position in the terminal.
we ask for it once per get_line()

alimpfard:
It always attempted to read 16 bytes at a time, but that `read` will succeed with just 1 byte if you're typing and the event loop will continue

bugaevc:
Well, that makes sense, doesn't it? We `select()` read 1 byte (or up to 16 bytes, if available), process it, `select()` again. `select()` is level-triggered, so it should all work?

alimpfard:
It _does_ work, we just end up waiting for that one byte.

bugaevc:
Waiting as in blocking, on first call? Or wdym?

alimpfard:
we block _once_ per get_line, since it tries to "prime" itself and refresh the display.
otherwise the prompt and stuff would only appear _after_ the user has input something

alimpfard:
Reading what I wrote before, I worded that rather poorly, my bad.

bugaevc:
> otherwise the prompt and stuff would only appear after the user has input something

why is that, why can't it display the prompt without any user input?

alimpfard:
...I'm not sure?
Let me try doing that, and seeing why it breaks.

alimpfard:
Well, that makes sense.
there is no reason to do that at all.

Shell/Shell.cpp
```diff
@@ -1835,3 +1835,26 @@ void Shell::stop_all_jobs()
...
+    Core::Object::save_to(object);
+    object.set("working_directory", cwd);
+    object.set("username", username);
+    object.set("user_home_path", home);
+    object.set("user_id", uid);
+    object.set("directory_stack_size", directory_stack.size());
+    object.set("cd_history_size", cd_history.size());
+
+    // Jobs.
+    object.set("number of jobs", jobs.size());
```
awesomekling, line 1860: How about we put the jobs in a `JsonArray` instead? Then we don't need a separate value for the count.

alimpfard:
:facepalm: I did it again...it's as if every PR is a fix-random-things-around-the-thing-that-the-pr-is-touching for me

PR #2400, opened 2020-05-26
LibLine: Use LibC's getline() when the terminal seems to not have escapes

alimpfard:
…apes

We just look at $TERM and refuse to emit any escape sequences if it
doesn't start with "xterm".
This could be made much better, at detecting, and at not caling
getline().

### DIFF
--- a/Libraries/LibLine/Editor.cpp
+++ b/Libraries/LibLine/Editor.cpp
@@ -226,10 +226,52 @@ void Editor::suggest(size_t invariant_offset, size_t static_offset, Span::Mode o
     m_suggestion_manager.set_suggestion_variants(internal_static_offset, internal_invariant_offset, 0);
 }
 
+void Editor::initialize()
+{
+    if (m_initialized)
+        return;
+
+    struct termios termios;
+    tcgetattr(0, &termios);
+    m_default_termios = termios; // grab a copy to restore
+
+    auto* term = getenv("TERM");
+    if (StringView { term }.starts_with("xterm"))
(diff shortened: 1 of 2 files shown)


awesomekling:
cc @bugaevc :thinking:

Libraries/LibLine/Editor.cpp
```diff
@@ -226,11 +226,55 @@ void Editor::suggest(size_t invariant_offset, size_t static_offset, Span::Mode o
...
+    // Because we use our own line discipline which includes echoing,
+    // we disable ICANON and ECHO.
+    if (auto* term = getenv("TERM")) {
+        if (StringView { term }.starts_with("xterm"))
+            m_configuration.set(Configuration::Full);
+        else
+            m_configuration.set(Configuration::NoEscapeSequences);
+    } else {
+        m_configuration.set(Configuration::NoEscapeSequences);
+    }
```
bugaevc, line 246: `StringView` can handle `nullptr` just fine, no need for the duplication here

alimpfard: That's news to me.
much better :)

bugaevc:
`#include <english-is-not-my-first-language.h>`

Commit message suggestion: ... when the terminal appears not to support escape sequences

alimpfard:
Hmmm...how about "when the terminal claims not to support escape sequences", setting TERM is a claim afterall :thinking:

alimpfard:
"claims no support for"?

commit messages are hard

PR #2401, opened 2020-05-26
LibJS: fix incorrect token column values

predmond:
- initializing m_line_column to 1 in the lexer results in incorrect
  column values in tokens on the first line of input.
- not incrementing m_line_column when EOF is reached results in
  an incorrect column value on the last token.


awesomekling:
Thanks folks!

PR #2405, opened 2020-05-26
LibWeb: Fix step within reconstruct the active elements

kevmeyer:
In step 4 of the "renstruct the active formatting elements" algorithm it
says:
  Rewind: If there are no entries before entry in the list of active
  formatting elements, then jump to the step labeled create.

Prior to this patch, the implementation accorded to the spec only for
the first loop iteration.

Disclaimer: Not tested yet in qemu, as I could not yet get the qemu booting the built image on macOS...


awesomekling:
Hey that's a nice catch! Thank you :)

PR #2407, opened 2020-05-26
LibWeb: Add more HTML entities

linusg:
®, ß and all the lowercase and uppercase umlaut characters.


awesomekling:
Sure go for it. This parser's days are numbered though :)

linusg:
I know, happy to add them to the new parser as well :)

(haven't actually looked at it though, other than watching the videos)

PR #2408, opened 2020-05-26
LibWeb: Handle additional parser inputs in "initial" and "before html".

TheDumpap:


TheDumpap:
I know it can be shortened a bit, but tried to follow the spec and write it very expressively :)
Please let me know if there is anything that should be changed, first commit here ^^

Libraries/LibWeb/Parser/HTMLDocumentParser.cpp
```diff
@@ -116,18 +116,39 @@ void HTMLDocumentParser::process_using_the_rules_for(InsertionMode mode, HTMLTok
...
         return;
     }
+
     ASSERT_NOT_REACHED();
 }
 
 void HTMLDocumentParser::handle_before_html(HTMLToken& token)
 {
+    if(token.is_doctype()) {
+        PARSE_ERROR();
```
awesomekling, line 143: Note that `PARSE_ERROR()` does not automatically return, since there are sometimes steps to take after a parse error.
Because of this, we have to `return;` here, otherwise we will fall down into the "anything else" section below :)

Libraries/LibWeb/Parser/HTMLDocumentParser.cpp
```diff
@@ -139,7 +160,22 @@ void HTMLDocumentParser::handle_before_html(HTMLToken& token)
...
         return;
     }
-    ASSERT_NOT_REACHED();
+
+    if(token.is_end_tag() && token.tag_name().is_one_of("head", "body", "html", "br")) {
+        goto AnythingElse;
+    }
+
+    if(token.is_end_tag()) {
+        return;
```
awesomekling, line 169: We should `PARSE_ERROR()` here.

TheDumpap:
Thanks for the swift feedback :) Fixed it up now.

Also, very sorry for this messy PR. I had some trouble squashing the commits, but I got it figured out now. Won't be a problem going forward :)

awesomekling:
:star:

PR #2409, opened 2020-05-26
Port of VirtualConsole to LibVT, text mode

bugaevc:
cc @deoxxa @alimpfard 

There are now two different ways to get a text mode shell:

* `boot_mode=text init=/bin/Shell` — this will make the kernel spawn `/bin/Shell` as PID 4 instead of SystemServer. There'll be *nothing* but the Shell (and the kernel tasks) running. No environment variables set, so the shell will not do any fancy effects unless you set `TERM=xterm` ~and restart it~. No filesystems mounted but the root, so `/tmp` is on the root Ext2FS, `/proc` is empty, and devices in `/dev` are unopenable. (Run `mount -a` manually to get to a more standard environment.) No sockets in `/tmp/portal` are present, so you won't be able to, say, do hostname lookup, unless you start LookupServer manually the way SystemServer does it. This is as spartan as it gets.

* just `boot_mode=text` — recognized by the SystemServer; it will disable all the graphical services (such as the WindowServer and its many clients), and instead spawn a pair of comfy shells for you on tty0 and tty1. In this configuration, you do not run as PID 4 (SystemServer does), you do get `TERM=xterm`, all the filesystems are mounted, and a few background services that make sense for a text mode are enabled. This is for your regular usage in text mode.

And as always, you can customize what gets spawned on startup (and how) by tweaking `/etc/SystemServer.ini`. For example, if you're hacking on ~PaintBrush~ PixelPaint, why not replace the default Terminal with ~PaintBrush~ PixelPaint to have SystemServer start it up for you? Or if you want to provide remote access to a Serenity box, why not start TelnetServer when booting in text mode.


alimpfard:
> unless you set TERM=xterm and restart it.

The line editor should pick it up on the following line after `export TERM=xterm`...

DevTools/HackStudio/CursorTool.cpp
```diff
@@ -147,6 +147,8 @@ void CursorTool::on_keydown(GUI::KeyEvent& event)
         case Key_Right:
             move_selected_widgets_by(m_editor.form_widget().grid_size(), 0);
             break;
+        default:
+            break;
```
alimpfard, line 151: I don't get your persistence on this, its existence makes no difference, and this exact behaviour is implied by its nonexistence.

Not nitpicking, just wondering out loud

Services/SystemServer/Service.cpp
```diff
@@ -343,6 +348,16 @@ void Service::save_to(JsonObject& json)
...
+
+    JsonArray boot_modes;
+    for (String& mode : m_boot_modes)
+        boot_modes.append(mode);
+    json.set("boot_modes", boot_modes);
+
+    JsonArray environment;
+    for (String& env : m_environment)
+        boot_modes.append(env);
+    json.set("environment", environment);
```
alimpfard, line 360: :thinking: why is this stuff commented out?

alimpfard: Does passing a JsonArray crash the inspector?

bugaevc:
> The line editor should pick it up on the following line after export TERM=xterm...

Oh, good. I think it didn't for me, but I only tested it once.

DevTools/HackStudio/CursorTool.cpp
bugaevc: 😀 it's not me, blame GCC

Services/SystemServer/Service.cpp
bugaevc: An array, I think. At least it did when I first wrote that code.

Kernel/init.cpp
```diff
@@ -302,28 +303,16 @@ void init_stage2()
...
-            hang();
-        }
-        thread->set_priority(THREAD_PRIORITY_HIGH);
+    // FIXME: It would be nicer to set the mode from userspace.
+    tty0->set_graphical(!text_mode);
+    Thread* thread = nullptr;
+    auto userspace_init = kernel_command_line().lookup("init").value_or("/bin/SystemServer");
+    Process::create_user_process(thread, userspace_init, (uid_t)0, (gid_t)0, (pid_t)0, error, {}, {}, tty0);
+    if (error != 0) {
+        klog() << "init_stage2: error spawning SystemServer: " << error;
```
bugaevc, line 312: Should update the error message here I guess.

Kernel/init.cpp
```diff
@@ -302,28 +303,16 @@ void init_stage2()
...
-        tty0->set_graphical(true);
-        Thread* thread = nullptr;
-        Process::create_user_process(thread, "/bin/SystemServer", (uid_t)0, (gid_t)0, (pid_t)0, error, {}, {}, tty0);
-        if (error != 0) {
-            klog() << "init_stage2: error spawning SystemServer: " << error;
-            hang();
-        }
-        thread->set_priority(THREAD_PRIORITY_HIGH);
+    // FIXME: It would be nicer to set the mode from userspace.
+    tty0->set_graphical(!text_mode);
```
bugaevc, line 307: By that, I mean some `ioctl` on the tty

Services/SystemServer/Service.cpp
alimpfard: I know that an empty array doesn't crash; let me check with an array that has something.

alimpfard: It seems to be okay with showing arrays (`[{"pid":25,"command":"te","running_time":5157,"pgid":25,"is_running_in_background":true}]` at least)

bugaevc: Why don't you uncomment the code here and try that.

alimpfard: Fiiine, I'll pull the PR

DevTools/HackStudio/CursorTool.cpp
alimpfard: Oh there are more enum values, I see.
\*blames GCC\*

Services/SystemServer/Service.cpp
alimpfard: Seems OK to me.

awesomekling:
This is super awesome :)

PR #2411, opened 2020-05-26
LibTLS: Add more TLS 1.2 error descriptions

Lubrsi:
Adds more TLS 1.2 error descriptions according to the specification:
https://tools.ietf.org/html/rfc5246#section-7.2.2

This changes the DecryptionFailed description, as the specification
says that this alert should NEVER be sent by a compliant server.

### DIFF
--- a/Libraries/LibTLS/TLSv12.cpp
+++ b/Libraries/LibTLS/TLSv12.cpp
@@ -636,7 +636,7 @@ bool Certificate::is_valid() const
 
 void TLSv12::try_disambiguate_error() const
 {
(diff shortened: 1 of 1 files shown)


alimpfard:
> This changes the DecryptionFailed description, as the specification says that this alert should NEVER be sent by a compliant server.

That was there because one-can-never-be-too-sure-about-their-own-infrastructure :tm:
(IIRC openssl's `s_server` _does_ send DecryptionFailed up until v1.1, but that's besides the point)

PR #2412, opened 2020-05-27
LibGUI/Browser: Add history to the JS Console

FalseHonesty:
With this patch, gone are the days of re-typing every command in the JS Console, just navigate with the up & down arrow keys!

### DIFF
--- a/Libraries/LibGUI/TextBox.cpp
+++ b/Libraries/LibGUI/TextBox.cpp
@@ -37,4 +37,55 @@ TextBox::~TextBox()
 {
 }
 
+void TextBox::keydown_event(GUI::KeyEvent& event)
+{
+    TextEditor::keydown_event(event);
+
+    if (event.key() == Key_Up) {
+        if (on_up_pressed)
+            on_up_pressed();
+
+        if (has_no_history() || !can_go_backwards_in_history())
+            return;
+
+        if (m_history_index >= static_cast<int>(m_history.size()))
+            m_saved_input = text();
+
+        m_history_index--;
+        set_text(m_history[m_history_index]);
+    } else if (event.key() == Key_Down) {
+        if (on_down_pressed)
+            on_down_pressed();
+
+        if (has_no_history())
+            return;
+
+        if (can_go_forwards_in_history()) {
+            m_history_index++;
+            set_text(m_history[m_history_index]);
+        } else if (m_history_index < static_cast<int>(m_history.size())) {
+            m_history_index++;
+            set_text(m_saved_input);
+        }
+    }
+}
+
+void TextBox::add_current_text_to_history()
+{
+    if (!m_history_enabled)
+        return;
+    
+    auto input = text();
+    if (m_history.is_empty() || m_history.last() != input)
+        add_input_to_history(input);
+    m_history_index = static_cast<int>(m_history.size());
+    m_saved_input = {};
+}
+
+void TextBox::add_input_to_history(String input)
+{
+    m_history.append(move(input));
+    m_history_index++;
+}
+
 }
--- a/Libraries/LibGUI/TextBox.h
+++ b/Libraries/LibGUI/TextBox.h
@@ -35,6 +35,25 @@ class TextBox : public TextEditor {
 public:
     TextBox();
     virtual ~TextBox() override;
+
+    Function<void()> on_up_pressed;
+    Function<void()> on_down_pressed;
+
+    void set_history_enabled(bool enabled) { m_history_enabled = enabled; }
+    void add_current_text_to_history();
+
+private:
+    virtual void keydown_event(GUI::KeyEvent&) override;
+
+    bool has_no_history() const { return !m_history_enabled || m_history.is_empty(); }
+    bool can_go_backwards_in_history() const { return m_history_index > 0; }
+    bool can_go_forwards_in_history() const { return m_history_index < static_cast<int>(m_history.size()) - 1; }
+    void add_input_to_history(String);
+
+    bool m_history_enabled { false };
+    Vector<String> m_history;
+    int m_history_index { -1 };
+    String m_saved_input;
 };
 
 }
--- a/Applications/Browser/ConsoleWidget.cpp
+++ b/Applications/Browser/ConsoleWidget.cpp
@@ -54,26 +54,29 @@ ConsoleWidget::ConsoleWidget()
     html_element->append_child(head_element);
     auto body_element = create_element(base_document, "body");
     html_element->append_child(body_element);
-    m_console_output_container = body_element;
+    m_output_container = body_element;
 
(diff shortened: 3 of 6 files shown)


Libraries/LibGUI/TextBox.cpp
```diff
@@ -37,4 +37,14 @@ TextBox::~TextBox()
 {
 }
 
+void TextBox::keydown_event(GUI::KeyEvent& event)
+{
+    if (event.key() == Key_Up)
+        on_up_pressed();
+    else if (event.key() == Key_Down)
+        on_down_pressed();
```
awesomekling, line 45: We have to null-check the `Function` objects before calling them:

```cpp
if (on_up_pressed)
    on_up_pressed();
```

awesomekling:
Having text box history seems like a generally useful thing! We also want this in the IRC client, for example :)

alimpfard:
@awesomekling but _please_, let's have that IRC history per-chat...and maybe cleared after a few minutes...

FalseHonesty:
> Having text box history seems like a generally useful thing! We also want this in the IRC client, for example :)

Oh, cool idea! I can put this behind a flag on TextBox, but having it disabled by default probably makes the most sense right?

awesomekling:
> > Having text box history seems like a generally useful thing! We also want this in the IRC client, for example :)
> 
> Oh, cool idea! I can put this behind a flag on TextBox, but having it disabled by default probably makes the most sense right?

Yeah that'd be hella nice

Libraries/LibGUI/TextBox.cpp
FalseHonesty: Ah of course, copy-paste strikes again :)

FalseHonesty:
I think I'll fix up IRC to start using this in a different PR if that's okay, it seems separate enough, and it might require a little extra work to make it work as @alimpfard suggested :)

awesomekling:
Sure thing

Libraries/LibGUI/TextBox.h
```diff
@@ -35,6 +35,27 @@ class TextBox : public TextEditor {
...
+
+    void set_history_enabled(bool enabled) { is_history_enabled = enabled; }
+    void add_current_text_to_history();
+
+private:
+    virtual void keydown_event(GUI::KeyEvent&) override;
+
+    bool has_any_history() const { return !m_input_history.is_empty(); }
+    bool has_previous_input() const { return m_current_historical_input > 1; }
+    bool has_next_input() const { return m_current_historical_input < m_input_history.size(); }
```
awesomekling, line 50: These look a bit strange; I would expect to "have previous input" if I'm at index 1.
Also, "previous input" is rather ambiguous, can we call these something like `can_go_backwards_in_history` / `can_go_forwards_in_history` to make it a bit clearer what's going on?

Libraries/LibGUI/TextBox.h
```diff
@@ -35,6 +35,27 @@ class TextBox : public TextEditor {
...
+    bool has_current_historical_input();
+    String current_historical_input();
+    void add_input_to_history(String);
+
+    bool is_history_enabled { false };
+    Vector<String> m_input_history;
+    size_t m_current_historical_input { 0 };
+    String m_possible_input;
```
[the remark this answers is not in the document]
FalseHonesty: Does `m_saved_input` make sense?

Libraries/LibGUI/TextBox.h
FalseHonesty: Yeah, renaming them sounds good. They're off set like that because I was making this a whole lot more complicated then it needed to be, let me make them 0-indexed again :^)

Libraries/LibGUI/TextBox.h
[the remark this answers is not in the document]
awesomekling: Yeah that works :)

Libraries/LibGUI/TextBox.cpp
```diff
@@ -37,4 +37,67 @@ TextBox::~TextBox()
...
+    } else if (event.key() == Key_Down) {
+        if (on_down_pressed)
+            on_down_pressed();
+
+        if (!m_history_enabled || !has_any_history())
+            return;
+
+        if (can_go_forwards_in_history()) {
+            m_history_index++;
+            set_text(current_historical_input());
```
awesomekling, line 65: This feels muddy. It's not obvious that incrementing `m_history_index` causes `current_historical_input` to return a different value.
How about:
```cpp
++m_history_index;
set_text(m_input_history[m_history_index]);
```

Libraries/LibGUI/TextBox.cpp
```diff
@@ -37,4 +37,67 @@ TextBox::~TextBox()
...
+    }
+}
+
+void TextBox::add_current_text_to_history()
+{
+    if (!m_history_enabled)
+        return;
+    
+    auto input = text();
+    if (!has_any_history() || m_input_history[m_input_history.size() - 1] != input)
```
awesomekling, line 79: This is also a bit muddy. "if not has any history" is a bit hard for my brain to parse. :)
Suggestion:
```cpp
if (m_input_history.is_empty() || m_input_history.last() != input)
```

Libraries/LibGUI/TextBox.cpp
```diff
@@ -37,4 +37,67 @@ TextBox::~TextBox()
...
+        add_input_to_history(input);
+    m_history_index = static_cast<int>(m_input_history.size());
+    m_saved_input = {};
+}
+
+bool TextBox::has_current_historical_input()
```
awesomekling, line 85: I think the code would be clearer without this helper.

Libraries/LibGUI/TextBox.cpp
```diff
@@ -37,4 +37,67 @@ TextBox::~TextBox()
...
+{
+    return m_history_index >= 0 && m_history_index < static_cast<int>(m_input_history.size());
+}
+
+String TextBox::current_historical_input()
```
awesomekling, line 90: I think the code would be clearer without this helper.

Libraries/LibGUI/TextBox.cpp
FalseHonesty: Well, it seems we're inverting the return value of `has_any_history` everywhere it's used, would `if (has_no_history() || m_input_history.last() != input)` be reasonable, or should I just use the `is_empty` method on the history list?

awesomekling: Let's just use the history vector directly. We could even rename it to `m_history` since there is only one kind of history here anyway :)

FalseHonesty:
> Okay, let's do one more round of nitpicks to improve clarity a bit :)

It's the nitpicking that keeps the code manageable, I'm all for it!

awesomekling:
Cool :)

PR #2413, opened 2020-05-27
LibWeb: s_initialized should be static in the AttributeNames initialiser

emanuele6:
`s_initialized`(in `Web::HTML::AttributeNames::initialize()`) was not `static`.

### DIFF
--- a/Libraries/LibWeb/DOM/AttributeNames.cpp
+++ b/Libraries/LibWeb/DOM/AttributeNames.cpp
@@ -35,7 +35,7 @@ FlyString class_;
 
 void initialize()
 {
-    bool s_initialized = false;
+    static bool s_initialized = false;
     if (s_initialized)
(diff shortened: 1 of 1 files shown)


emanuele6:
In `AK/Base.cpp`, we are using this style even though `s_initialiazed` is only used in `build_lookup_table_if_needed()`:
```cpp
static bool s_initialized = false;

static void build_lookup_table_if_needed()
{
    if (s_initialized)
        return;
    for (size_t i = 0; i < sizeof(s_alphabet) - 1; ++i)
        s_table[s_alphabet[i]] = i;
    s_initialized = true;
}
```

Should I declare `s_initialized` outside of `Web::HTML::AttributeNames::initialize()`?

awesomekling:
It's all good, thanks for looking out :)

PR #2414, opened 2020-05-27
Meta: Update C++ standard to C++20 in CONTRIBUTING.md

emanuele6:
We are compiling with `-std=c++2a` and we are using some C++20 features, e.g. in `Kernel/VM/Region.h`
```cpp
bool m_shared : 1 { false };
bool m_user_accessible : 1 { false };
bool m_cacheable : 1 { false };
bool m_stack : 1 { false };
bool m_mmap : 1 { false };
```

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -33,7 +33,7 @@ Nobody is perfect, and sometimes we mess things up. That said, here are some goo
 
 **Do:**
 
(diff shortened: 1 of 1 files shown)


bugaevc:
Ah, bitflags, my favorite C++20 feature. "What do you mean you've seen it before, it's brand new!"

PR #2415, opened 2020-05-27
LibJS: Simplify and normalize publicly-exposed Object functions

mattco98:
Previously, the `Object` class had many different types of functions for each action. For example: `get_by_index`, `get(PropertyName)`, and `get(FlyString)`. This is a bit verbose, so these methods have been shortened to simply use the `PropertyName` structure. The methods then internally call `_by_index` if necessary. Note that the `_by_index` methods have been made private to enforce this change.

Secondly, a clear distinction has been made between "putting" and "defining" an object property. "Putting" should mean modifying a (potentially) already existing property. This is akin to doing `a.b = 'foo'`.

This implies two things about put operations:

- They will search the prototype chain for setters and call them, if necessary.
- If no property exists with a particular key, the put operation should create a new property with the default attributes (configurable, writable, and enumerable).

In contrast, "defining" a property should completely overwrite any existing value without calling setters (if that property is configurable, of course).

Thus, all of the many JS objects have had any relevant `put` calls changed to `define_property` calls. Additionally, `put_native_function` and `put_native_property` have had their `put` replaced with `define`.

Finally, `put_own_property` has been made private, as all necessary functionality should be exposed with the `put` and `define_property` methods.

This is progress towards the implementation of #2330. Allowing all public methods to deal with `PropertyName` allows the easy integration of numeric properties in a future PR.

### DIFF
--- a/Libraries/LibJS/AST.cpp
+++ b/Libraries/LibJS/AST.cpp
@@ -1282,20 +1282,20 @@ Value ObjectExpression::execute(Interpreter& interpreter) const
                 for (size_t i = 0; i < elements.size(); ++i) {
                     auto element = elements.at(i);
                     if (!element.is_empty())
-                        object->put_by_index(i, element);
+                        object->define_property(i, element);
                 }
             } else if (key_result.is_object()) {
                 auto& obj_to_spread = key_result.as_object();
 
                 for (auto& it : obj_to_spread.shape().property_table_ordered()) {
                     if (it.value.attributes & Attribute::Enumerable)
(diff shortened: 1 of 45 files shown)


mattco98:
Also note that there are a few `ASSERT_NOT_REACHED` calls in some active code branches. While I _could_ have implemented that functionality in this PR, it's going to be very quickly replaced by the new indexed property code. I already have it all written, but it was very large and I figured it'd be good to split up the PR.

awesomekling:
This nicely disambiguates a bunch of things. Good :)

PR #2416, opened 2020-05-27
Userland: Add a basic xargs program

alimpfard:
This adds a basic implementation of xargs.
This xargs is not particularly Dr. POSIX-compliant.

cc @bugaevc pls nitpick the heck out of my badly written manpage

### DIFF
--- a/AK/StringView.cpp
+++ b/AK/StringView.cpp
@@ -75,6 +75,32 @@ Vector<StringView> StringView::split_view(const char separator, bool keep_empty)
     return v;
 }
 
+Vector<StringView> StringView::split_view(const StringView& separator, bool keep_empty) const
+{
+    ASSERT(!separator.is_empty());
+
+    if (is_empty())
+        return {};
+
+    StringView view { *this };
+
+    Vector<StringView> parts;
+
+    auto maybe_separator_index = find_first_of(separator);
+    while (maybe_separator_index.has_value()) {
+        auto separator_index = maybe_separator_index.value();
+        auto part_with_separator = view.substring_view(0, separator_index + separator.length());
+        if (keep_empty || separator_index > 0)
+            parts.append(part_with_separator.substring_view(0, separator_index));
+        view = view.substring_view_starting_after_substring(part_with_separator);
+        maybe_separator_index = view.find_first_of(separator);
+    }
+    if (keep_empty || !view.is_empty())
+        parts.append(view);
+
+    return parts;
+}
+
 Vector<StringView> StringView::lines(bool consider_cr) const
 {
     if (is_empty())
@@ -244,7 +270,7 @@ Optional<size_t> StringView::find_first_of(const StringView& view) const
 
 Optional<size_t> StringView::find_last_of(char c) const
 {
-    for (size_t pos = m_length; --pos >0;) {
+    for (size_t pos = m_length; --pos > 0;) {
         if (m_characters[pos] == c)
             return pos;
     }
--- a/Base/usr/share/man/man1/xargs.md
+++ b/Base/usr/share/man/man1/xargs.md
@@ -1,0 +1,47 @@
+## Name
+
+xargs - build and execute commandlines from input
+
+## Synopsis
+
+```**sh
+$ xargs [options...] [command [initial-arguments...]]
+```
+
+## Description
+
+`xargs` reads items from a stream, delimited by some blank character (`delimiter`), and executes the `command` as many times as there are items, with any processed `initial-arguments`, possibly followed by a number of items read from the input.
+
+If a `placeholder` is explicitly specified, the `max-lines` limit is set to 1, and each argument in `initial-arguments` is processed by replacing any occurence of the `placeholder` with the input item, and treating the entire resulting value as _one_ argument.
+
+It is to be noted that `command` is also subject to substitution in this mode.
+
+If no argument in `command` or `initial-arguments` contains the `placeholder`, an argument is added at the end of the list containing only the `placeholder`.
+
+
+If a `placeholder` is not explicitly specified, no substitution will be performed, rather, the item(s) will be appended to the end of the command line, until either of the following conditions are met:
+- Adding another argument would overflow the system maximum command length (or the provided `max-chars` limit)
+- The number of lines used per command (`max-lines`) would be exceeded
+
+
+`xargs` will read the items from standard input by default, and when data is read from standard input, the standard input of `command` is redirected to read from `/dev/null`.
+The standard input is left as-is if data is read from a file.
+
+## Options
+
+* `-I`, `--replace`: Set the `placeholder`, and force `max-lines` to 1
+* `-0`, `--null`: Split the items by zero bytes (null characters) instead of `delimiter`
+* `-d`, `--delimiter`: Set the `delimiter`, which is a newline (`\n`) by default
+* `-v`, `--verbose`: Display each expanded command on standard error before executing it
+* `-a`, `--arg-file`: Read the items from the speified file, `-` refers to standard input and is the default
+* `-L`, `--line-limit`: Set `max-lines`, `0` means unlimited (which is the default)
+* `-s`, `--char-limit`: Set `max-chars`, which is `ARG_MAX` (the maximum command size supported by the system) by default
+
+## Examples
+
+```sh
+$ pro http://list-of-example-urls.com/plain | xargs -I 'URL' pro URL > concatenated-outputs
+$ xargs -a list-of-files-to-delete --verbose rm
+$ xargs -a list-of-moves -L 2 mv
+$ xargs -a stuff --null -s 1024
+```
--- a/Libraries/LibC/stdio.cpp
+++ b/Libraries/LibC/stdio.cpp
@@ -77,7 +77,6 @@ public:
     void set_popen_child(pid_t child_pid) { m_popen_child = child_pid; }
 
 private:
-
     struct Buffer {
         // A ringbuffer that also transparently implements ungetc().
     public:
@@ -911,23 +910,36 @@ void perror(const char* s)
 static int parse_mode(const char* mode)
 {
     int flags = 0;
+
     // NOTE: rt is a non-standard mode which opens a file for read, explicitly
     // specifying that it's a text file
-    if (!strcmp(mode, "r") || !strcmp(mode, "rb") || !strcmp(mode, "rt"))
-        flags = O_RDONLY;
-    else if (!strcmp(mode, "r+") || !strcmp(mode, "rb+"))
-        flags = O_RDWR;
-    else if (!strcmp(mode, "w") || !strcmp(mode, "wb"))
-        flags = O_WRONLY | O_CREAT | O_TRUNC;
-    else if (!strcmp(mode, "w+") || !strcmp(mode, "wb+"))
-        flags = O_RDWR | O_CREAT | O_TRUNC;
-    else if (!strcmp(mode, "a") || !strcmp(mode, "ab"))
-        flags = O_WRONLY | O_APPEND | O_CREAT;
-    else if (!strcmp(mode, "a+") || !strcmp(mode, "ab+"))
-        flags = O_RDWR | O_APPEND | O_CREAT;
-    else {
-        dbg() << "Unexpected mode _" << mode << "_";
-        ASSERT_NOT_REACHED();
+    for (; *mode; ++mode) {
+        switch (*mode) {
+        case 'r':
+            flags |= O_RDONLY;
+            break;
+        case 'w':
+            flags |= O_WRONLY | O_CREAT | O_TRUNC;
+            break;
+        case 'a':
+            flags |= O_WRONLY | O_APPEND | O_CREAT;
+            break;
+        case '+':
+            flags |= O_RDWR;
+            break;
+        case 'e':
+            flags |= O_CLOEXEC;
+            break;
+        case 'b':
+            // Ok...
+            break;
+        case 't':
+            // Ok...
+            break;
+        default:
+            dbg() << "Unsupported mode _" << mode << "_ (because of '" << *mode << "')";
+            ASSERT_NOT_REACHED();
+        }
     }
 
     return flags;
--- a/Userland/xargs.cpp
+++ b/Userland/xargs.cpp
@@ -1,0 +1,298 @@
+/*
+ * Copyright (c) 2020, The SerenityOS developers.
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include <AK/Function.h>
+#include <AK/StdLibExtras.h>
+#include <AK/StringBuilder.h>
+#include <LibCore/ArgsParser.h>
+#include <errno.h>
+#include <fcntl.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <sys/wait.h>
+
+bool run_command(Vector<char*>&& child_argv, bool verbose, bool is_stdin, int devnull_fd);
+
+enum Decision {
+    Unget,
+    Continue,
+    Stop,
+};
+bool read_items(FILE* fp, char entry_separator, Function<Decision(StringView)>);
+
+class ParsedInitialArguments {
+public:
+    ParsedInitialArguments(Vector<const char*>&, const StringView& placeholder);
+
+    void for_each_joined_argument(const StringView&, Function<void(const String&)>) const;
+
+    size_t size() const { return m_all_parts.size(); }
+
+private:
+    Vector<Vector<StringView>> m_all_parts;
+};
+
+int main(int argc, char** argv)
+{
+    if (pledge("stdio rpath proc exec", nullptr) < 0) {
+        perror("pledge");
+        return 1;
+    }
+
+    const char* placeholder = nullptr;
+    bool split_with_nulls = false;
+    const char* specified_delimiter = "\n";
+    Vector<const char*> arguments;
+    bool verbose = false;
+    const char* file_to_read = "-";
+    int max_lines_for_one_command = 0;
+    int max_bytes_for_one_command = ARG_MAX;
+
+    Core::ArgsParser args_parser;
+    args_parser.add_option(placeholder, "Placeholder string to be replaced in arguments", "replace", 'I', "placeholder");
+    args_parser.add_option(split_with_nulls, "Split input items with the null character instead of newline", "null", '0');
+    args_parser.add_option(specified_delimiter, "Split the input items with the specified character", "delimiter", 'd', "delim");
+    args_parser.add_option(verbose, "Display each command before executing it", "verbose", 'v');
+    args_parser.add_option(file_to_read, "Read arguments from the specified file instead of stdin", "arg-file", 'a', "file");
+    args_parser.add_option(max_lines_for_one_command, "Use at most max-lines lines to create a command", "line-limit", 'L', "max-lines");
+    args_parser.add_option(max_bytes_for_one_command, "Use at most max-chars characters to create a command", "char-limit", 's', "max-chars");
+    args_parser.add_positional_argument(arguments, "Command and any initial arguments for it", "command", Core::ArgsParser::Required::No);
+    args_parser.parse(argc, argv);
+
+    size_t max_bytes = min(ARG_MAX, max_bytes_for_one_command);
+    size_t max_lines = max(max_lines_for_one_command, 0);
+
+    if (!split_with_nulls && strlen(specified_delimiter) > 1) {
+        fprintf(stderr, "xargs: the delimiter must be a single byte\n");
+        return 1;
+    }
+
+    char entry_separator = split_with_nulls ? '\0' : *specified_delimiter;
+
+    StringView placeholder_view { placeholder };
+
+    if (!placeholder_view.is_empty())
+        max_lines = 1;
+
+    if (arguments.is_empty())
+        arguments.append("echo");
+
+    ParsedInitialArguments initial_arguments(arguments, placeholder_view);
+
+    FILE* fp = stdin;
+    bool is_stdin = true;
+
+    if (StringView { "-" } != file_to_read) {
+        // A file was specified, try to open it.
+        fp = fopen(file_to_read, "re");
+        if (!fp) {
+            perror("fopen");
+            return 1;
+        }
+        is_stdin = false;
+    }
+
+    StringBuilder builder;
+    Vector<char*> child_argv;
+
+    int devnull_fd = 0;
+
+    if (is_stdin) {
+        devnull_fd = open("/dev/null", O_RDONLY | O_CLOEXEC);
+        if (devnull_fd < 0) {
+            perror("open");
+            return 1;
+        }
+    }
+
+    size_t total_command_length = 0;
+    size_t items_used_for_this_command = 0;
+
+    auto fail = read_items(fp, entry_separator, [&](StringView item) {
+        if (item.ends_with('\n'))
+            item = item.substring_view(0, item.length() - 1);
+
+        if (item.is_empty())
+            return Continue;
+
+        // The first item is processed differently, as all the initial-arguments are processed _with_ that item
+        // as their substitution target (assuming that substitution is enabled).
+        // Note that if substitution is not enabled, we manually insert a substitution target at the end of initial-arguments,
+        // so this item has a place to go.
+        if (items_used_for_this_command == 0) {
+            child_argv.ensure_capacity(initial_arguments.size());
+
+            initial_arguments.for_each_joined_argument(item, [&](const String& string) {
+                total_command_length += string.length();
+                child_argv.append(strdup(string.characters()));
+            });
+
+            ++items_used_for_this_command;
+        } else {
+            if ((max_lines > 0 && items_used_for_this_command + 1 > max_lines) || total_command_length + item.length() + 1 >= max_bytes) {
+                // Note: This `move' does not actually move-construct a new Vector at the callsite, it only allows perfect-forwarding
+                //       and does not invalidate `child_argv' in this scope.
+                //       The same applies for the one below.
+                if (!run_command(move(child_argv), verbose, is_stdin, devnull_fd))
+                    return Stop;
+                items_used_for_this_command = 0;
+                total_command_length = 0;
+                return Unget;
+            } else {
+                child_argv.append(strndup(item.characters_without_null_termination(), item.length()));
+                total_command_length += item.length();
+                ++items_used_for_this_command;
+            }
+        }
+
+        return Continue;
+    });
+
+    if (!fail && !child_argv.is_empty())
+        fail = !run_command(move(child_argv), verbose, is_stdin, devnull_fd);
+
+    if (!is_stdin)
+        fclose(fp);
+
+    return fail ? 1 : 0;
+}
+
+bool read_items(FILE* fp, char entry_separator, Function<Decision(StringView)> callback)
+{
+    bool fail = false;
+
+    for (;;) {
+        char* item = nullptr;
+        size_t buffer_size = 0;
+
+        auto item_size = getdelim(&item, &buffer_size, entry_separator, fp);
+
+        if (item_size < 0) {
+            // getdelim() will return -1 and set errno to 0 on EOF.
+            if (errno != 0) {
+                perror("getdelim");
+                fail = true;
+            }
+            break;
+        }
+
+        Decision decision;
+        do {
+            decision = callback(item);
+            if (decision == Stop) {
+                free(item);
+                return true;
+            }
+        } while (decision == Unget);
+
+        free(item);
+    }
+
+    return fail;
+}
+
+bool run_command(Vector<char*>&& child_argv, bool verbose, bool is_stdin, int devnull_fd)
+{
+    child_argv.append(nullptr);
+
+    if (verbose) {
+        StringBuilder builder;
+        builder.join(" ", child_argv);
+        fprintf(stderr, "xargs: %s\n", builder.to_string().characters());
+        fflush(stderr);
+    }
+
+    auto pid = fork();
+    if (pid < 0) {
+        perror("fork");
+        return false;
+    }
+
+    if (pid == 0) {
+        if (is_stdin)
+            dup2(devnull_fd, STDIN_FILENO);
+
+        execvp(child_argv[0], child_argv.data());
+        exit(1);
+    }
+
+    for (auto* ptr : child_argv)
+        free(ptr);
+
+    child_argv.clear_with_capacity();
+
+    int wstatus = 0;
+    if (waitpid(pid, &wstatus, 0) < 0) {
+        perror("waitpid");
+        return false;
+    }
+
+    if (WIFEXITED(wstatus)) {
+        if (WEXITSTATUS(wstatus) != 0)
+            return false;
+    } else {
+        return false;
+    }
+
+    return true;
+}
+
+ParsedInitialArguments::ParsedInitialArguments(Vector<const char*>& arguments, const StringView& placeholder)
+{
+    m_all_parts.ensure_capacity(arguments.size());
+    bool some_argument_has_placeholder = false;
+
+    for (auto argument : arguments) {
+        StringView arg { argument };
+
+        if (placeholder.is_empty()) {
+            m_all_parts.append({ arg });
+        } else {
+            auto parts = arg.split_view(placeholder, true);
+            some_argument_has_placeholder = some_argument_has_placeholder || parts.size() > 1;
+            m_all_parts.append(move(parts));
+        }
+    }
+
+    // Append an implicit placeholder at the end if no argument has any placeholders.
+    if (!some_argument_has_placeholder) {
+        Vector<StringView> parts;
+        parts.append("");
+        parts.append("");
+        m_all_parts.append(move(parts));
+    }
+}
+
+void ParsedInitialArguments::for_each_joined_argument(const StringView& separator, Function<void(const String&)> callback) const
+{
+    StringBuilder builder;
+    for (auto& parts : m_all_parts) {
+        builder.clear();
+        builder.join(separator, parts);
+        callback(builder.to_string());
+    }
+}
--- a/AK/StringView.h
+++ b/AK/StringView.h
@@ -88,6 +88,7 @@ public:
 
     StringView substring_view(size_t start, size_t length) const;
     Vector<StringView> split_view(char, bool keep_empty = false) const;
+    Vector<StringView> split_view(const StringView&, bool keep_empty = false) const;
 
     // Create a Vector of StringViews split by line endings. As of CommonMark
     // 0.29, the spec defines a line ending as "a newline (U+000A), a carriage


Userland/xargs.cpp
```diff
@@ -0,0 +1,186 @@
...
+    const char* replace_str = "{}";
+    bool split_with_nulls = false;
+    const char* specified_delimiter = "\n";
+    Vector<const char*> arguments;
+    bool verbose = false;
+    const char* file_to_read = "-";
+
+    Core::ArgsParser args_parser;
+    args_parser.add_positional_argument(command, "Command to run", "command");
+    args_parser.add_positional_argument(arguments, "Initial arguments to pass to command", "initial_args", Core::ArgsParser::Required::No);
```
bugaevc, line 54: Not important, but the convention is (aka I did this when porting stuff to `Core::ArgsParser`) to add options first, then positional args.

Userland/xargs.cpp
```diff
@@ -0,0 +1,186 @@
...
+    args_parser.add_option(replace_str, "Value to be replaced in arguments", "replace", 'I', "replace_str");
```
bugaevc, line 55: I think the word you're looking for is "placeholder"

Userland/xargs.cpp
alimpfard: Noted.
Thanks for letting me know.

Userland/xargs.cpp
alimpfard: Hmmm...sure, it's better than "replacement string", I guess

Userland/xargs.cpp
bugaevc: Is there a reason to differentiate the command from the arguments? Like, isn't this all just argv?

Userland/xargs.cpp
```diff
@@ -0,0 +1,186 @@
...
+    args_parser.add_option(split_with_nulls, "Split input items with the null character instead of newline", "null", '0');
+    args_parser.add_option(specified_delimiter, "Split the input items with the specified character", "delimiter", 'd', "delim");
+    args_parser.add_option(verbose, "Display each command before executing it", "verbose", 'v');
+    args_parser.add_option(file_to_read, "Read arguments from the specified file instead of stdin", "arg-file", 'a', "file");
+    args_parser.parse(argc, argv);
+
+    char entry_separator = split_with_nulls ? '\0' : *specified_delimiter;
```
bugaevc, line 62: So is it always a single byte?

alimpfard: Yes.
Do we have a add_option(char&,...)? I didn't see it in the suggestions so I assumed we didn't have it

Userland/xargs.cpp
```diff
@@ -0,0 +1,186 @@
...
+        Vector<StringView> parts;
+        size_t previous_index = 0;
+        if (auto first_index = arg.find_first_of(replace_view); first_index.has_value()) {
+            some_argument_has_replacement = true;
+            while (first_index.has_value()) {
+                auto current_index = first_index.value();
+                auto substring_until_replacement = arg.substring_view(previous_index, current_index);
+                parts.append(substring_until_replacement);
+                previous_index = current_index;
+                first_index = arg.substring_view(current_index + replace_view.length(), arg.length() - current_index - replace_view.length()).find_first_of(replace_view);
```
bugaevc, line 82: Try one of these:

https://github.com/SerenityOS/serenity/blob/dd08c992e8d884f646bdf9e9e75b2236ad5d38c8/AK/StringView.h#L101-L118

bugaevc: And generally, I have troubles understanding what this block of code is doing. Perhaps a comment with some examples would help?

alimpfard: That won't cut it, we have
```
foo{}bar
   ^ find_first_of
     ^ where we want to start at
   ^ where substr_after would give for substring_until_replacement
```

Userland/xargs.cpp
```diff
@@ -0,0 +1,186 @@
...
+            perror("open");
+            return 1;
+        }
+    }
+
+    for (;;) {
+        auto line_size = getdelim(&line, &buffer_size, entry_separator, fp);
+
+        if (line_size <= 0)
+            break;
```
bugaevc, line 137: This just swallows read errors doesn't it?

Userland/xargs.cpp
alimpfard: - We don't want to replace anything in the command
- the command isn't optional

Userland/xargs.cpp
alimpfard: Ah, we should show the error too, yeah.

Userland/xargs.cpp
```diff
@@ -0,0 +1,186 @@
...
+        }
+
+        if (verbose) {
+            fprintf(stderr, "xargs: ");
+            for (auto* ptr : child_argv) {
+                if (!ptr)
+                    break;
+                fprintf(stderr, "%s ", ptr);
+            }
+            fprintf(stderr, "\n");
```
bugaevc, line 159: Flush here.

(Though in case of stderr specifically, buffering is disabled by default on Serenity.)

alimpfard: stderr shouldn't be buffered anywhere...anyone that does that is insane

Userland/xargs.cpp
```diff
@@ -0,0 +1,186 @@
...
+
+        auto pid = fork();
+        if (pid < 0) {
+            perror("fork");
+            return 1;
+        }
+
+        if (pid == 0) {
+            if (is_stdin)
+                dup2(devnull_fd, STDIN_FILENO);
```
bugaevc, line 170: What about all the other fds that you've left open? You don't even close `devnull_fd`... Either use `O_CLOEXEC` (preferred), or make (very) sure to close them all here.

alimpfard: ouch.
I also left `line` un-freed

alimpfard: Do we support that gnu extension mode for fopen that does `CLOEXEC`?
I think it was `e`?

alimpfard: That's a no.

Userland/xargs.cpp
```diff
@@ -0,0 +1,186 @@
...
+    if (is_stdin) {
+        devnull_fd = open("/dev/null", O_RDONLY);
+        if (devnull_fd < 0) {
+            perror("open");
+            return 1;
+        }
+    }
+
+    for (;;) {
+        auto line_size = getdelim(&line, &buffer_size, entry_separator, fp);
```
bugaevc, line 134: As far as I can see, this runs the command with a *single* argument out of those specified in the input file. But that's not what `xargs` does, it passes them all to one invocation (unless they don't fits into argv). E.g.

```
$ xargs ps <<EOF
35
42
411
EOF
```

should invoke

```
$ ps 35 42 411
```

not

```
$ ps 35
$ ps 42
$ ps 411
```

alimpfard: I _did_ document that behaviour in the manpage, since I didn't add `-L`, I let the default behaviour be `-L 1`

I personally like this better, but I guess I can change it to work more like posix xargs

Base/usr/share/man/man1/xargs.md
```diff
@@ -0,0 +1,33 @@
...
+
+```**sh
+$ xargs [options...] command [initial-arguments...]
+```
+
+## Description
+
+`xargs` reads items from a stream, delimited by some blank character (`delimiter`), and executes the `command` as many times as there are items, with any processed `initial-arguments`, possibly followed by an item read from the input.
+Each argument in `initial-arguments` is processed by replacing any occurence of the `replacement string` with the input item, and treating the entire resulting value as _one_ argument.
+If no argument in `initial-arguments` contains the `replacement string`, an argument is added at the end of the list containing only the `replacement string`.
```
bugaevc, line 15: I'm sorry but "replacement string" is not a valid identifier/token (as it contains the space), so it feels wrong to put it in monospace. Either add a dash (`replacement-string`), or better, just say `placeholder`.

alimpfard: I swapped it for placeholder along with the first suggestion, no worries :)

Base/usr/share/man/man1/xargs.md
```diff
@@ -0,0 +1,33 @@
...
+`xargs` will read the items from stdin by default, and when data is read from stdin, the standard input of `command` is redirected to `/dev/null`.
+The standard input is left as-is if data is read from a file.
```
bugaevc, line 17: These two lines sound like they should be in a separate paragraph, don't you think?

alimpfard: 15,16?
I wouldn't know, I've never written a manpage before

Base/usr/share/man/man1/xargs.md
```diff
@@ -0,0 +1,33 @@
...
+
+## Options
+
+* `-I`, `--replace`: Sets the `replacement string`, this is `{}` by default
+* `-0`, `--null`: Splits the items by null characters instead of `delimiter`
```
bugaevc, line 22: s/null characters/zero bytes/

or both: _zero bytes (null characters)_

Base/usr/share/man/man1/xargs.md
```diff
@@ -0,0 +1,33 @@
...
+* `-d`, `--delimiter`: Sets the `delimiter`, this is a newline (`\n`) by default
+* `-v`, `--verbose`: Displays each expanded command on stderr before executing it
```
bugaevc, line 24: s/stderr/standard error/

same goes for the other mentions of stdin/stdout/stderr

Base/usr/share/man/man1/xargs.md
```diff
@@ -0,0 +1,33 @@
...
+* `-a`, `--arg-file`: Reads the items from the speified file, `-` refers to stdin and is the default
+
+## Examples
+
+```sh
+$ cat list-of-urls | xargs -I 'URL' pro URL > concatenated-outputs
```
bugaevc, line 30: Are you seriously showcasing a Useless Use Of :cat: as an example?

Base/usr/share/man/man1/xargs.md
```diff
@@ -0,0 +1,33 @@
...
+$ xargs -a list-of-files-to-delete -v rm {}
```
bugaevc, line 31: How about some `--long-options` in this example? E.g. s/`-v`/`--verbose`/

Base/usr/share/man/man1/xargs.md
alimpfard: I'll add another argument to cat then, tee-hee.

(I'll fix it to be more elaborate, then)

bugaevc:
> @bugaevc pls nitpick the heck out of my badly written manpage

You got it! But not before I nitpicked the heck out of your beautifully written code :wink: :smile:

alimpfard:
> But not before I nitpicked the heck out of your beautifully written code

I expect that when I cc you lol

Userland/xargs.cpp
bugaevc: > We don't want to replace anything in the command

Why not support it?

> the command isn't optional

Yes, so the argv as a whole isn't optional either. (Actually, Dr. POSIX says "If utility is omitted, the default shall be the echo utility.")

Userland/xargs.cpp
bugaevc: We don't, but we actually have a generic way to register any sort of options/arguments that you need, with any custom types / logic. The provided int / C string / vector of C strings / whatever are just common enough to be implemented in `Core::ArgsParser` itself, on top of the same generic API.

Userland/xargs.cpp
alimpfard: Might I ask what the point of that is?
I mean, it's pretty trivial to support, but I don't see any point in a glorified `cat my-file`

Userland/xargs.cpp
alimpfard: We don't have it.
I'll just add a check to make that more obvious.

Userland/xargs.cpp
bugaevc: > Do we support that gnu extension mode for fopen that does `CLOEXEC`?

If you need it, why not implemented it :wink:

Userland/xargs.cpp
bugaevc: > since I didn't add `-L`, I let the default behaviour be `-L 1`

`L` is (I bet) for "limit". It's for limiting how many items you process in one go; the default is no limit at all, not one.

I mean, it's fine to not support advanced options / modes & stuff, but `xargs`, even in its simplest form, does a different thing from what you've implemented.

Side note: for the longest time, I actually thought `xargs` does just that, invokes one command per item.

alimpfard: Oh, I _knowingly_ didn't do what xargs does, I prefer it with this behaviour.
but again, I can make it cram as many arguments as it can into a single command...

Base/usr/share/man/man1/xargs.md
```diff
@@ -0,0 +1,34 @@
...
+$ xargs [options...] command [initial-arguments...]
+```
+
+## Description
+
+`xargs` reads items from a stream, delimited by some blank character (`delimiter`), and executes the `command` as many times as there are items, with any processed `initial-arguments`, possibly followed by an item read from the input.
+Each argument in `initial-arguments` is processed by replacing any occurence of the `placeholder` with the input item, and treating the entire resulting value as _one_ argument.
+If no argument in `initial-arguments` contains the `placeholder`, an argument is added at the end of the list containing only the `placeholder`.
+
+`xargs` will read the items from standard input by default, and when data is read from standard input, the standard input of `command` is redirected to `/dev/null`.
```
bugaevc, line 17: `#include <english-is-not-my-first-language.h>`

stdin cannot be redirected _to_, only _from_

I think it sounds best to say "the standard input is redirected _to read_ from `/dev/null`

Base/usr/share/man/man1/xargs.md
```diff
@@ -0,0 +1,34 @@
...
+The standard input is left as-is if data is read from a file.
+
+## Options
+
+* `-I`, `--replace`: Sets the `placeholder`, this is `{}` by default
```
bugaevc, line 22:
```suggestion
* `-I`, `--replace`: Set the `placeholder`, which is `{}` by default
```

Base/usr/share/man/man1/xargs.md
```diff
@@ -0,0 +1,34 @@
...
+* `-0`, `--null`: Splits the items by zero bytes (null characters) instead of `delimiter`
```
bugaevc, line 23:
```suggestion
* `-0`, `--null`: Split the items by zero bytes (null characters) instead of `delimiter`
```

bugaevc: (you get the idea)

Userland/xargs.cpp
bugaevc: It replaces any whitespace with simple spaces for one thing. But I'm not saying this is useful, this is just what POSIX specifies.

Treating argv[0] same as the rest of argv, on the other hand, is a beautiful principle that we should follow whenever possible.

Userland/xargs.cpp
```diff
@@ -0,0 +1,201 @@
...
+            int wstatus = 0;
+            if (waitpid(pid, &wstatus, 0) < 0) {
+                perror("waitpid");
+                return 1;
+            }
+            // FIXME: Should we take a look at the exit code and fail if the command fails?
+            for (size_t i = 0; i < argument_parts.size(); ++i)
+                free(child_argv[i + 1]);
+
+            free(line);
```
bugaevc, line 193: Why wait for completion before freeing these all?

Userland/xargs.cpp
bugaevc: Unresolving, `fopen()` still doesn't cloexec the file, does it?

Userland/xargs.cpp
alimpfard: No particular reason, I guess it just felt better this way.
I'll move it up, I guess

Userland/xargs.cpp
alimpfard: So we should allow replacements in the command? I mean...why not.

Userland/xargs.cpp
alimpfard: _do_ we want to do it the posix way?
it's a bit of a rewrite if we do

bugaevc: I'm afraid we do

alimpfard: T-T
Now i have to support half a million nonsensical modes just to appease Dr. POSIX.

bugaevc: How about we drop `-I` support instead? You'd simply have to append items to the argv.

alimpfard: Let's just full-arse this thing lol.
I _want_ the one-line-per-command functionality.

alimpfard: `-L` support full-arsed :^)

Userland/xargs.cpp
```diff
@@ -0,0 +1,276 @@
...
+    size_t max_bytes = min(ARG_MAX, max_bytes_for_one_command);
+
+    if (!split_with_nulls && strlen(specified_delimiter) > 1) {
+        fprintf(stderr, "xargs: the delimiter must be a single byte\n");
+        return 1;
+    }
+
+    char entry_separator = split_with_nulls ? '\0' : *specified_delimiter;
+
+    StringView replace_view { placeholder };
```
bugaevc, line 75:
```suggestion
    StringView placeholder_view { placeholder };
```

Userland/xargs.cpp
```diff
@@ -0,0 +1,276 @@
...
+    if (is_stdin) {
+        devnull_fd = open("/dev/null", O_RDONLY | O_CLOEXEC);
+        if (devnull_fd < 0) {
+            perror("open");
+            return 1;
+        }
+    }
+
+    bool fail = false;
+    int lines_used_for_this_command = 0;
```
bugaevc, line 148:
```suggestion
    size_t lines_used_for_this_command = 0;
```

Userland/xargs.cpp
```diff
@@ -0,0 +1,276 @@
...
+    char* last_line = nullptr;
```
bugaevc, line 149:
```suggestion
    char* previous_line = nullptr;
```
perhaps?

Userland/xargs.cpp
```diff
@@ -0,0 +1,276 @@
...
+    for (auto argument : arguments) {
+        StringView arg { argument };
+        Vector<StringView> parts;
+        size_t previous_index = 0;
+        if (!replace_view.is_empty()) {
+            if (auto first_index = arg.find_first_of(replace_view); first_index.has_value()) {
+                some_argument_has_replacement = true;
+                while (first_index.has_value()) {
+                    auto current_index = first_index.value();
+                    auto substring_until_replacement = arg.substring_view(previous_index, current_index);
```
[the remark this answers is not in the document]
alimpfard: I mean...sure, but I don't think it's common enough to warrant that

Userland/xargs.cpp
alimpfard: Sure.
makes no difference to me.

Userland/xargs.cpp
```diff
@@ -0,0 +1,276 @@
...
+        } else {
+            line_size = getdelim(&line, &buffer_size, entry_separator, fp);
+        }
+
+        if (line_size < 0) {
+            if (errno != 0) {
+                perror("getdelim");
+                fail = true;
+            }
+            break;
```
bugaevc, line 172: Apparently `getdelim()` will return -1 (not 0) on EOF! This was not clear to me at all. Could use a comment in the code, e.g.

```c++
// getdelim() will return -1 and set errno to 0 on EOF.
```

Userland/xargs.cpp
```diff
@@ -0,0 +1,276 @@
...
+
+        if (line_size == 0) {
+            free(line);
+            last_line = nullptr;
+            continue;
+        }
+
+        ++lines_used_for_this_command;
+
+        if (line_size > 0) {
```
bugaevc, line 186: Haven't we established by this point that `line_size > 0`?

In general, I feel all this line handling logic would be also better off in a separate function/class.

Userland/xargs.cpp
alimpfard: I hit that with "Success (not an error)" and figured that out the hard way...but sure, helpful comments are good things.

Userland/xargs.cpp
```diff
@@ -0,0 +1,276 @@
...
+            while (line[line_size - 1] == '\n')
+                --line_size;
```
bugaevc, line 188: This will underflow if the whole line only consists of `\n`s

Userland/xargs.cpp
```diff
@@ -0,0 +1,276 @@
...
+
+            if (!line_size) {
+                last_line = nullptr;
```
bugaevc, line 191: Leaking `line` here.

Userland/xargs.cpp
alimpfard: Hmmm...that should actually check for only one newline.

Userland/xargs.cpp
alimpfard: I get the concern, but I don't see its point.
There's no need for reuse, so putting stuff in classes is a bit overdoing it...it's just a loop with a one-line history

Userland/xargs.cpp
```diff
@@ -0,0 +1,276 @@
...
+    char entry_separator = split_with_nulls ? '\0' : *specified_delimiter;
+
+    StringView replace_view { placeholder };
+
+    if (arguments.is_empty())
+        arguments.append("echo");
+
+    Vector<Vector<StringView>> argument_parts;
+    argument_parts.ensure_capacity(arguments.size());
+    size_t index = 0;
```
bugaevc, line 82: I don't see where this is used?

Userland/xargs.cpp
```diff
@@ -0,0 +1,276 @@
...
+            if (!line_size) {
+                last_line = nullptr;
+                continue;
+            }
+
+            StringView line_view { line, (size_t)line_size };
+
+            dbg() << "Getline '" << line_view << "'";
+
+            size_t index = 0;
```
bugaevc, line 199: ...nor where this is used...

Userland/xargs.cpp
alimpfard: This is int because casting `max_lines_for_one_command` was annoying.
I'll do that instead, sure.

Userland/xargs.cpp
```diff
@@ -0,0 +1,276 @@
...
+                    builder.clear();
+                    builder.join(line_view, parts);
+                    auto string = builder.build();
+                    total_command_length += string.length();
+                    child_argv.unchecked_append(strdup(string.characters()));
+                    ++index;
+                }
+            } else {
+                if ((max_lines_for_one_command > 0 && lines_used_for_this_command > max_lines_for_one_command) || total_command_length + line_size + 1 >= max_bytes) {
+                    run_command(child_argv, verbose, is_stdin, devnull_fd);
```
bugaevc, line 212: You're running the command as is after you've established that `lines_used_for_this_command > max_lines_for_one_command`. You're not actually using the line for this command, are you? Then don't increment `lines_used_for_this_command` until you do decide that you are definetely going to use it.

alimpfard: it's reset to zero below, then set to 1 on the next iteration for the next command.
it's easier to write out like this, without `lines_used_for_this_command + 1`.

alimpfard: I guess its name is a bit misleading, it would be `lines_that_would_be_used_for_this_command`, but that's so damn long

Userland/xargs.cpp
alimpfard: curse gcc for not complaining about `++whatever` being unused.
I don't notice this stuff darn it

Userland/xargs.cpp
bugaevc: And yes, it sounds like the code would benefit a lot from untangling "read more lines until <condition>" from "construct child argv from the lines being read". The former part doesn't have to know about `child_argv`, `lines_used_for_this_command`, etc — just read lines and invoke a callback (perhaps the callback would either accept or "unget" the line). The latter part doesn't have to know about `getdelim`, separators, `last_line`, `line_size`, all that crap — it just sees if the new line fits and if so, append it to the argv, otherwise reject it and `run_command`.

Userland/xargs.cpp
```diff
@@ -0,0 +1,276 @@
...
+
+    if (!is_stdin)
+        fclose(fp);
+
+    return fail ? 1 : 0;
+}
+
+int run_command(Vector<char*>& child_argv, bool verbose, bool is_stdin, bool devnull_fd)
+{
+    child_argv.append(0);
```
bugaevc, line 237:
```suggestion
    child_argv.append(nullptr);
```

Userland/xargs.cpp
```diff
@@ -0,0 +1,276 @@
...
+
+    if (verbose) {
+        fprintf(stderr, "xargs: ");
+        for (auto* ptr : child_argv) {
+            if (!ptr)
+                break;
+            fprintf(stderr, "%s ", ptr);
```
bugaevc, line 244: Super nitpick: not cool to output a space after the last arg; perhaps use `join()` or something.

alimpfard: \*cough\* it's not even visible

Userland/xargs.cpp
```diff
@@ -0,0 +1,276 @@
...
+    if (!child_argv.is_empty())
+        run_command(child_argv, verbose, is_stdin, devnull_fd);
+
+    if (!is_stdin)
+        fclose(fp);
+
+    return fail ? 1 : 0;
+}
+
+int run_command(Vector<char*>& child_argv, bool verbose, bool is_stdin, bool devnull_fd)
```
bugaevc, line 235:
```suggestion
bool run_command(Vector<char*>& child_argv, bool verbose, bool is_stdin, int devnull_fd)
```

Userland/xargs.cpp
```diff
@@ -0,0 +1,276 @@
...
+
+    if (pid == 0) {
+        if (is_stdin)
+            dup2(devnull_fd, STDIN_FILENO);
+
+        execvp(child_argv[0], child_argv.data());
+    }
+
+    for (size_t i = 0; i < child_argv.size() - 2; ++i)
+        free(child_argv[i + 1]);
```
bugaevc, line 264: Why treat `child_argv[0]` differently?

Userland/xargs.cpp
alimpfard: ...did I seriously...

bugaevc:
I'm really sorry for all these new nitpicks...

Userland/xargs.cpp
alimpfard: Oh, we shouldn't, not anymore.

alimpfard:
I'm really sorry for the dumb problems

Userland/xargs.cpp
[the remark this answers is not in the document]
bugaevc: It's not about code reuse, really, it's about separating things out so you can think about one thing at a time. Right now the logic here is too hard for me to comprehend (without straining my brain)

Base/usr/share/man/man1/xargs.md
```diff
@@ -0,0 +1,47 @@
...
+The standard input is left as-is if data is read from a file.
+
+## Options
+
+* `-I`, `--replace`: Set the `placeholder`, and forces `max-lines` to 1
+* `-0`, `--null`: Split the items by zero bytes (null characters) instead of `delimiter`
+* `-d`, `--delimiter`: Set the `delimiter`, which is a newline (`\n`) by default
+* `-v`, `--verbose`: Display each expanded command on standard error before executing it
+* `-a`, `--arg-file`: Read the items from the speified file, `-` refers to standard input and is the default
+* `-L`, `--line-limit`, Set `max-lines`, `0` means unlimited (which is the default)
```
bugaevc, line 37:
```suggestion
* `-L`, `--line-limit`: Set `max-lines`, `0` means unlimited (which is the default)
```

Base/usr/share/man/man1/xargs.md
```diff
@@ -0,0 +1,47 @@
...
+* `-s`, `--char-limit`, Set `max-chars`, which is `ARG_MAX` (the maximum command size supported by the system) by default
```
bugaevc, line 38:
```suggestion
* `-s`, `--char-limit`: Set `max-chars`, which is `ARG_MAX` (the maximum command size supported by the system) by default
```

Base/usr/share/man/man1/xargs.md
alimpfard: oops.

Userland/xargs.cpp
```diff
@@ -0,0 +1,285 @@
...
+                fail = true;
+            }
+            break;
+        }
+
+        previous_line = line;
+        previous_line_size = line_size;
+
+        auto decision = callback({ line, (size_t)line_size });
+        free(line);
```
bugaevc, line 232: You can't `free()` it yet, `previous_line` refers to it.

An idea how to simplify this further:

```c++
while (true) {
    char* line;
    getdelim(...);

    Decision decision;
    do {
        decision = callback(line);
        if (decision == Stop) {
            free(line);
            return;
        }
    } while (decision == Unget);

    free(line);
}
```

and no need for separate `previous_*` then.

Userland/xargs.cpp
```diff
@@ -0,0 +1,285 @@
...
+
+    if (pid == 0) {
+        if (is_stdin)
+            dup2(devnull_fd, STDIN_FILENO);
+
+        execvp(child_argv[0], child_argv.data());
+    }
+
+    for (size_t i = 0; i < child_argv.size() - 1; ++i)
+        free(child_argv[i]);
```
bugaevc, line 273: Now you can simply do `for (auto ptr : child_argv) free(ptr);`

Userland/xargs.cpp
alimpfard: I like that, sure, let's go with that.

Userland/xargs.cpp
```diff
@@ -0,0 +1,285 @@
...
+        StringBuilder builder;
+        builder.join(" ", child_argv);
+        fprintf(stderr, "xargs: %s\n", builder.to_string().characters());
+        fflush(stderr);
+    }
+
+    auto pid = fork();
+    if (pid < 0) {
+        perror("fork");
+        return 1;
```
bugaevc, line 262:
```suggestion
        return false;
```

Userland/xargs.cpp
```diff
@@ -0,0 +1,285 @@
...
+    size_t lines_used_for_this_command = 0;
+
+    auto fail = read_lines(fp, entry_separator, [&](StringView line) {
+        if (line.ends_with('\n'))
+            line = line.substring_view(0, line.length() - 1);
+
+        if (line.is_empty())
+            return Continue;
+
+        if (lines_used_for_this_command == 0) {
```
bugaevc, line 161: My understanding is that we process the first line differently for two reasons:
* It must fit into `max_lines` (makes no sense otherwise)
* We substitute the first line in place of the placeholder; whereas we simply append other lines as separate args

Might be a good idea to spell this out in a comment.

Also perhaps rename `line`s to `item`s everywhere?

Userland/xargs.cpp
```diff
@@ -0,0 +1,285 @@
...
+            for (auto& parts : argument_parts) {
+                builder.clear();
+                builder.join(line, parts);
+                auto string = builder.build();
+                total_command_length += string.length();
+                child_argv.unchecked_append(strdup(string.characters()));
+            }
+            ++lines_used_for_this_command;
+        } else {
+            if ((max_lines > 0 && lines_used_for_this_command > max_lines - 1) || total_command_length + line.length() + 1 >= max_bytes) {
```
bugaevc, line 172: I'd rejiggle it as
```suggestion
            if ((max_lines > 0 && lines_used_for_this_command + 1 > max_lines) || total_command_length + line.length() + 1 >= max_bytes) {
```
because this way, it reads as "if we add one more line, is it still gonna fit?", which is the point here. The byte check on this line already does that.

Userland/xargs.cpp
```diff
@@ -0,0 +1,307 @@
...
+    for (auto argument : arguments) {
+        StringView arg { argument };
+        Vector<StringView> parts;
+        size_t previous_index = 0;
+        if (!placeholder.is_empty()) {
+            // Split the argument into segments delimited by the placeholder.
+            if (auto first_index = arg.find_first_of(placeholder); first_index.has_value()) {
+                some_argument_has_replacement = true;
+                while (first_index.has_value()) {
+                    auto current_index = first_index.value();
```
bugaevc, line 274: So `current_index` is simply the non-optional `first_index`? Perhaps the naming could be improved, though I don't have immediate suggestions.

Userland/xargs.cpp
```diff
@@ -0,0 +1,307 @@
...
+                    auto substring_until_replacement = arg.substring_view(previous_index, current_index);
```
bugaevc, line 275:
```suggestion
                    auto substring_until_placeholder = arg.substring_view(previous_index, current_index);
```

or perhaps simply

```suggestion
                    auto part = arg.substring_view(previous_index, current_index);
```

Userland/xargs.cpp
```diff
@@ -0,0 +1,307 @@
...
+
+    for (auto argument : arguments) {
+        StringView arg { argument };
+        Vector<StringView> parts;
+        size_t previous_index = 0;
+        if (!placeholder.is_empty()) {
+            // Split the argument into segments delimited by the placeholder.
+            if (auto first_index = arg.find_first_of(placeholder); first_index.has_value()) {
+                some_argument_has_replacement = true;
+                while (first_index.has_value()) {
```
bugaevc, line 273: Why does this have to be nested (`while` inside a `for`)?

bugaevc: Wait, that wasn't a `for`, it was an `if`!

But still, I don't think it has to be nested.

Userland/xargs.cpp
```diff
@@ -0,0 +1,307 @@
...
+                    parts.append(substring_until_replacement);
+                    previous_index = current_index;
+                    first_index = arg.substring_view(current_index + placeholder.length(), arg.length() - current_index - placeholder.length()).find_first_of(placeholder);
+                }
+                parts.append(arg.substring_view(previous_index + placeholder.length(), arg.length() - previous_index - placeholder.length()));
+            } else {
+                parts.append(arg);
+            }
+        } else {
+            parts.append(arg);
```
bugaevc, line 285: Super nitpick: let's process this simple case first, otherwise it's not too clear what the `else` refers to.

Userland/xargs.cpp
```diff
@@ -0,0 +1,307 @@
...
+        }
+        m_all_parts.append(move(parts));
+    }
+
+    // Append a '{}' at the end if no argument has replacements.
```
bugaevc, line 290: You're not really appending `{}`, you're just saying that an implicit placeholder goes at the end in that case.

Userland/xargs.cpp
```diff
@@ -0,0 +1,307 @@
...
+
+    int wstatus = 0;
+    if (waitpid(pid, &wstatus, 0) < 0) {
+        perror("waitpid");
+        return false;
+    }
+
+    if (WIFEXITED(wstatus)) {
+        if (WEXITSTATUS(wstatus) != 0)
+            return false;
```
bugaevc, line 252: My reading of POSIX is that this should be only considered to fail if the command exits with code 255. Which is oddly specific, and I have no idea why they want that.

Userland/xargs.cpp
alimpfard: wdym?
"loop over the inner segments, then add the last bit" but that only applies if there's a placeholder in there somewhere

un-nesting this would just add another if at the end, not sure what we gain here.
the if with the initialiser might be a bit weird, I'll move that variable out.

but I see no real point to un-nesting that while, if anything, it hurts readability here

Userland/xargs.cpp
alimpfard: :shrug: honestly, I don't give two shits about the weird behaviours, and might I point out the note that says "does not follow posix closely"?
This is the sort of behaviour that I'd only add if I saw any particular need for.

Userland/xargs.cpp
bugaevc: * Add all but last (if any)
* Add the last

reads cleaner to me then

* If there are more than one:
    * Add all but last
    * Add the last
* Else:
    * Add the only one

Don't you agree? And I don't see why you'd need an `if` at the end?

alimpfard: we can't find the last in a clean way here.
"last" is offset `placeholder.length()` bytes if the placeholder exists, and not otherwise

Userland/xargs.cpp
alimpfard: Eh, `first_index` is named badly, it'd be `maybe_placeholder_start_index`
and `current_index`...`placeholder_start_index`, I guess?

Userland/xargs.cpp
bugaevc: No, just keep an offset where to start the next part from. Initially it's 0, if you find the placeholder it becomes the end of that placeholder. In other words, that "offset `placeholder.length()` bytes" should happen at the previous iteration of the loop.

Userland/xargs.cpp
bugaevc: Agreed on that, just wondering aloud what the heck they mean by that.

Userland/xargs.cpp
alimpfard: Alright, let's try that.

Userland/xargs.cpp
bugaevc: Yeah, that sounds better.

I also wish C++ would let you rebind names, like

```c++
auto foo_bar = baz();
if (foo_bar.is_error())
    return foo_bar.error();
auto foo_bar = foo_bar.value();
```

(Or even
```c++
auto foo_bar = baz()?;
```
but that's going too far with the idea.)

Userland/xargs.cpp
alimpfard: Fixing the logic takes a minute, renaming things right takes 19 minutes.
let's just revert back to `var0`, `var1`, etc

bugaevc: No worries, naming things is one of the hardest computer science problems :smile:

Userland/xargs.cpp
```diff
@@ -0,0 +1,304 @@
...
+            }
+        } while (decision == Unget);
+
+        free(item);
+    }
+
+    return fail;
+}
+
+bool run_command(Vector<char*>& child_argv, bool verbose, bool is_stdin, int devnull_fd)
```
bugaevc, line 214: Actually let's make this

```suggestion
bool run_command(Vector<char*>&& child_argv, bool verbose, bool is_stdin, int devnull_fd)
```

and change the call sites to do an explicit `move()`.

Userland/xargs.cpp
```diff
@@ -0,0 +1,304 @@
...
+    bool some_argument_has_replacement = false;
+
+    for (auto argument : arguments) {
+        StringView arg { argument };
+        Vector<StringView> parts;
+        if (placeholder.is_empty()) {
+            parts.append(arg);
+        } else {
+            auto maybe_placeholder_index = arg.find_first_of(placeholder);
+            size_t last_placeholder_end_offset = 0;
```
bugaevc, line 272: Now here's a trick: instead of storing this index, just modify the `arg` view. A whole bunch of index math will just vanish.

Userland/xargs.cpp
```diff
@@ -0,0 +1,304 @@
...
+            // Split the argument into segments delimited by the placeholder.
+            while (maybe_placeholder_index.has_value()) {
+                some_argument_has_replacement = true;
+                auto placeholder_end_offset = maybe_placeholder_index.value() + placeholder.length();
+                auto part = arg.substring_view(last_placeholder_end_offset, placeholder_end_offset - placeholder.length());
+                parts.append(part);
+                last_placeholder_end_offset += placeholder_end_offset;
+                maybe_placeholder_index = arg.substring_view(last_placeholder_end_offset, arg.length() - last_placeholder_end_offset).find_first_of(placeholder);
```
bugaevc, line 280: And if you do that, I think you'll be able to use `substring_view_starting_after_substring()` here nicely after all, and get rid of even more index math.

Userland/xargs.cpp
alimpfard: I'm gonna have to ask for a reason, as moving this is entirely too wrong to me, we're not giving the vector to `run_command`, we're giving the contents to it.

i.e. moving it out will force us to recreate a new Vector for each iteration, while we can reuse the same backing memory and not reallocate a billion vectors

Userland/xargs.cpp
```diff
@@ -0,0 +1,304 @@
...
+
+    for (auto argument : arguments) {
+        StringView arg { argument };
+        Vector<StringView> parts;
+        if (placeholder.is_empty()) {
+            parts.append(arg);
+        } else {
+            auto maybe_placeholder_index = arg.find_first_of(placeholder);
+            size_t last_placeholder_end_offset = 0;
+            // Split the argument into segments delimited by the placeholder.
```
bugaevc, line 273: Also hold ooon, why don't we just use `arg.split_view()`?

If that doesn't support strings, then well, it should; move this code to an `StringView::substring_view()` overload.

alimpfard: It doesn't support strings, no.
I'm a little uncomfortable with messing with the stuff in AK, mainly due to the _ugh, rebuilding everything again_ factor of working on them

Userland/xargs.cpp
bugaevc: I think you might misunderstand what `&&` and `move()` do in C++. They don't actually move anything; they just signify to the reader (and to the compiler, when resolving overloads, see below) that you do not care about this value staying the same (or valid) after the call.

Now, if the method you're invoking is overloaded and has separate versions that take a `const T&` and a `T&&`, the former would usually copy the `T` value, and the latter would "steal" the value from the caller, as the caller indicated it doesn't care about the value.

Here, `run_command` empties (and otherwise cleans up) the vector, which to me sounds exactly like what a function taking a `&&` can be expected to do. It should have *zero* impact on the generated code (feel free to verify), 'cause we're not overloading anything, but it serves as a sort of comment for the reader.

Userland/xargs.cpp
bugaevc: You don't have to rebuild everything, `ninja xargs` should only rebuild what's necessary for you. (And it only takes a minute to rebuild the whole thing anyway — it took me longer to write this comment, and I'm not done typing it yet.)

The logic for splitting a string by a separator clearly belongs in AK. That's actually the same thing, separation of concerns — the splitting code shouldn't care about multiple arguments, and about adding an implicit placeholder at the end, and so on; and the placeholder parsing code shouldn't care about how exactly string splitting is implemented in terms of indexes and the like. But this time there's actually a chance of code reuse, though clearly this is not a common use case.

bugaevc: By the way, you can clearly tell the simplification is working — I'm now able to understand what all this code is supposed to be doing, suggest improvements, and even recognize patterns like an ad-hoc `split_view()` implementation.

Userland/xargs.cpp
alimpfard: to my information, there is never any copy involved with passing a reference, rvalue or lvalue.
you might end up with a copy constructor, but that's completely irrelevant here.

so far as the compiler (and the reader) is concerned, the value left behind after an rvalue reference is passed to some function is completely crap data.
Even though I'm pretty sure that will not mess with the value of our Vector here, it sends a completely wrong message.

I _do_ care about the vector instance, if our Vector impl someday decided to actually move on a move constructor from an rvalue reference, this code will be completely screwed.

Userland/xargs.cpp
alimpfard: Oh, I'm putting it in AK anyway, I'm just bitching about the annoyances of messing with AK

Userland/xargs.cpp
alimpfard: Update: our Vector instance _does_ actually move on `Vector(Vector&&)`, so I'm gonna go ahead and say that this _would_ actually reallocate the vector every time

bugaevc: > if our Vector impl someday decided to actually move on a move constructor from an rvalue reference, this code will be completely screwed.

I bet it already does. But we're not move-constructing any vectors here, just passing one rvalue Vector reference to a function.

alimpfard: Why fight on theoretical stuff when we can fight over profiles?
let's see if there actually is a lot of reallocations :shrug:

alimpfard: Fantastic.
The generated code is literally the same, equal offsets, equal instructions.
Same number of everything.

bugaevc: What did I tell ya? :wink:

> It should have zero impact on the generated code (feel free to verify), 'cause we're not overloading anything, but it serves as a sort of comment for the reader.

alimpfard: At this point, even though I find the notion of telling the compiler to `move` the vector very weird for the use-case...I don't mind "move"'ing the vector.

I will say that doing this might confuse future readers, however.

alimpfard: Also, `clear_with_capacity` is so very misleading, it reads to me as "clear everything, including the capacity"

bugaevc: > Also, `clear_with_capacity` is so very misleading, it reads to me as "clear everything, including the capacity"

Agreed, perhaps rename it to `clear_keeping_capacity()` (or `clear_preserving_capacity()`) if you're messing with AK anyway (and if @awesomekling approves)?

alimpfard:
Also, I'm surprised `fork` is this slow, on a 5-second run of `xargs -a stuff -IX echo .X.X.. > /dev/null` (with stuff having 5000 lines), `fork` is at the top of the profile, taking 4 seconds...that's 800 microseconds for each fork...

Userland/xargs.cpp
alimpfard: I went ahead and added a comment about this move behaviour.
commenting a comment ftw

AK/StringView.cpp
```diff
@@ -75,6 +75,34 @@ Vector<StringView> StringView::split_view(const char separator, bool keep_empty)
...
 
+Vector<StringView> StringView::split_view(const StringView& separator, bool keep_empty) const
+{
+    if (is_empty())
+        return {};
+
+    StringView view { *this };
+
+    Vector<StringView> parts;
+    if (separator.is_empty()) {
```
bugaevc, line 86: Mmm, it's not clear to me how that should work. I'd rather have this method `ASSERT(!separator.is_empty())`, and have the callers deal with the empty case. Unless you can make an argument why it's a good idea to handle it this way.

AK/StringView.cpp
```diff
@@ -75,6 +75,34 @@ Vector<StringView> StringView::split_view(const char separator, bool keep_empty)
...
+        parts.append(view);
+        return parts;
+    }
+
+    auto maybe_separator_index = find_first_of(separator);
+    while (maybe_separator_index.has_value()) {
+        auto separator_offset = maybe_separator_index.value();
+        auto part_with_placeholder = view.substring_view(0, separator_offset + separator.length());
```
bugaevc, line 94:
```suggestion
        auto part_with_separator = view.substring_view(0, separator_offset + separator.length());
```

AK/StringView.cpp
alimpfard: Not really, no.
we could also go the javascript way and return the characters as strings?

AK/StringView.cpp
alimpfard: oops, forgot that one :)

AK/StringView.cpp
```diff
@@ -75,6 +75,34 @@ Vector<StringView> StringView::split_view(const char separator, bool keep_empty)
...
+    StringView view { *this };
+
+    Vector<StringView> parts;
+    if (separator.is_empty()) {
+        parts.append(view);
+        return parts;
+    }
+
+    auto maybe_separator_index = find_first_of(separator);
+    while (maybe_separator_index.has_value()) {
```
bugaevc, line 92: Now we can probably make this into a `for` loop, because it really is one:

```c++
for (
    auto maybe_separator_index = find_first_of(separator);
    maybe_separator_index.has_value();
    maybe_separator_index = find_first_of(separator)
)
```

alimpfard: I mean...that reads like 'please don't do this to me'

AK/StringView.cpp
bugaevc: Yeah, let's not do either of those. If the caller is passing an empty string it's probably a mistake on their side. Haven't validated input or something. So it's a good idea to assert to catch those.

AK/StringView.cpp
alimpfard: No, really, I like the `while` version better.

```c++
    for (
        auto maybe_separator_index = find_first_of(separator);
        maybe_separator_index.has_value();
        maybe_separator_index = view.find_first_of(separator)) {

        auto separator_offset = maybe_separator_index.value();
        auto part_with_separator = view.substring_view(0, separator_offset + separator.length());
        if (keep_empty || separator_offset > 0)
            parts.append(part_with_separator.substring_view(0, separator_offset));
        view = view.substring_view_starting_after_substring(part_with_separator);
    }
```
...meh

AK/StringView.cpp
alimpfard: Let's assert then.
fine by me.

AK/StringView.cpp
bugaevc: Feel free not to do this if you don't like it,

but my general logic is to use `while` loops for, you know, doing stuff while a condition is true, and `for` loops for iterating over things. Especially loops where you have actions that can be clearly classified as *moving to the next step* (rather than processing this current item) look like `for` loops to me. This is not a 100% strict rule, of course, and it's not me who writes this code, so I'm completely fine with not doing it.

alimpfard: The next-step thing would have us doing this-
```c++
 for (
        auto maybe_separator_index = find_first_of(separator);
        maybe_separator_index.has_value();
        view = view.substring_view_starting_after_substring(part_with_separator),
        maybe_separator_index = view.find_first_of(separator)) {

        auto separator_offset = maybe_separator_index.value();
        auto part_with_separator = view.substring_view(0, separator_offset + separator.length());
        if (keep_empty || separator_offset > 0)
            parts.append(part_with_separator.substring_view(0, separator_offset));
    }
```

yeah no.

Userland/xargs.cpp
```diff
@@ -0,0 +1,292 @@
...
+ParsedInitialArguments::ParsedInitialArguments(Vector<const char*>& arguments, const StringView& placeholder)
+{
+    m_all_parts.ensure_capacity(arguments.size());
+    bool some_argument_has_replacement = false;
+
+    for (auto argument : arguments) {
+        StringView arg { argument };
+        m_all_parts.append(arg.split_view(placeholder));
+        if (!some_argument_has_replacement)
+            some_argument_has_replacement = arg.find_first_of(placeholder).has_value();
```
bugaevc, line 272: I would write this as

```c++
auto parts = arg.split_view(placeholder);
some_argument_has_placeholder = some_argument_has_placeholder || parts.size() > 1;
m_all_parts.append(move(parts));
```

but it's also fine this way, so whichever you like best.

Also `some_argument_has_replacement` → `some_argument_has_placeholder`

Userland/xargs.cpp
```diff
@@ -0,0 +1,292 @@
...
+        arguments.append("echo");
+
+    ParsedInitialArguments initial_arguments(arguments, placeholder_view);
+
+    FILE* fp = stdin;
+    bool is_stdin = true;
+
+    if (StringView { "-" } != file_to_read) {
+        // A file was specified, try to open it.
+        fp = fopen(file_to_read, "r");
```
bugaevc, line 111: I believe you're still leaking it to the children. If you don't want to implement cloexec support for `fopen()`, just make sure to close it after forking.

Userland/xargs.cpp
alimpfard: Hmm...yeah, recomputing that isn't nice, let's go with this

Userland/xargs.cpp
alimpfard: oh, I'll mess with fopen then
should I just do the gnu extension `e`?

bugaevc: I would do that yeah.

Libraries/LibC/stdio.cpp
```diff
@@ -911,20 +910,26 @@ void perror(const char* s)
 static int parse_mode(const char* mode)
 {
     int flags = 0;
+
+    if (*mode == 'e') {
+        flags = O_CLOEXEC;
+        ++mode;
```
alimpfard, line 916: As with every other gnu extension, there's no particular definition of where this goes or how it goes where it goes, so I just assumed it goes at the beginning, since that was the most clean way :shrug:

ADKaster: I've personally written code that calls both `::fopen(filename, "re" )` and `::fopen(filename, "we")` and they both seem to work? Not sure that's a good assumption ;)

alimpfard: Well that's concerning...
how do these work anyway? does, for instance, "re+" mean the same thing as "r+e"?
"ab" ~~ "ba"?

If so, it's just a slightly weird thing that can be done relatively easily...if not, I don't like implementing it.

ADKaster: In the bugs section of this page
https://www.man7.org/linux/man-pages/man3/fopen.3.html#bugs

It says that it will parse "up to 7 characters". So I think you're allowed to put them in whatever order you want.

alimpfard: If this random repository is to be believed, the first character has to be one of rwa, the rest can be in any order
https://github.com/lattera/glibc/blob/895ef79e04a953cac1493863bcae29ad85657ee1/libio/fileops.c#L225

alimpfard: Let's hope this change doesn't break anything :P

bugaevc: > As with every other gnu extension, there's no particular definition of where this goes or how it goes where it goes, so I just assumed it goes at the beginning, since that was the most clean way shrug

The man page says:

> The argument mode points to a string beginning with one of the following sequences (possibly followed by additional characters, as described below).

> The mode string can also include the letter 'b' either as a last character or as a character between the characters in any of the two-character strings described above.

And lists "r", "r+", "w", w+", "a", "a+" as those sequences it has to begin with, and "c", "e", "m", "x" (and ",ccs=...") as additional characters.

So I don't think your criticism of GNU extensions is justified, at least not in this case.

bugaevc: Now that I think of it, and read your patch, — why don't we just support any letters in any order? Just loop over them and `|=` the required flags.

Libraries/LibC/stdio.cpp
```diff
@@ -911,25 +910,53 @@ void perror(const char* s)
 static int parse_mode(const char* mode)
 {
     int flags = 0;
-    // NOTE: rt is a non-standard mode which opens a file for read, explicitly
-    // specifying that it's a text file
-    if (!strcmp(mode, "r") || !strcmp(mode, "rb") || !strcmp(mode, "rt"))
+    switch (*mode) {
+    case 'r':
```
alimpfard, line 914: @bugaevc a little late for that lol

bugaevc: What do you mean?

alimpfard: Just that it already does the parsing thing you suggested above.

bugaevc: But it parses the first character different to others, while I'm suggesting it could not do that and treat all the characters the same way.

alimpfard: Ah, I see, I managed to misread that somehow.
I could do that, but I believe it's done this way to prevent bogus modes such as "rw" or "awwwwwww"

bugaevc: I don't insist on doing it that way, it just sounds simpler and more generic to me.

alimpfard: I'm not against it, I'm just saying we explicitly define its semantics first if we do that,
i.e. what should "rw" mean? same as "r+"? Repetition can be ignored, conflicting information...not so much

bugaevc: > we explicitly define its semantics first if we do that

We don't have to do that, either. Garbage in, garbage out — you'll probably get some nonsensical combination of flags and the kernel will reject it with `EINVAL`. That being said, "rw" will naturally resolve to `O_RDONLY | O_WRONLY`, which on Serenity is the same as `O_RDWR`, so it'll behave as expected.

alimpfard: Garbage in, garbage out, but at which level? Why should an abstraction leak failures it can detect?
Ideally, fopen wouldn't do a syscall if you feed it crap at all

bugaevc: To spell it out: I'm not suggesting we should officially support and define semantics for any combination of any mode characters in any order. I'm mainly suggesting we could simplify the implementation, and as a side effect enable specifying characters in any order. It still means that if you pass an invalid combination, you're going to get yourself in trouble, just perhaps into some different sort of trouble.

alimpfard: And I'm saying that we shouldn't allow things that we have no defined semantics for.
If we don't support crap like "awr", we shouldn't let it get down to the kernel before it gets possibly rejected (although I don't think there's any particular combination that would fail, since "awr" is just "a+" with our flags).

But why leak crap when we can simply not?

bugaevc: Dr. POSIX says, "The mode argument points to a string. If the string is one of the following, the file shall be opened in the indicated mode. Otherwise, the behavior is undefined."

Feeding a possibility nonsensical combination of flags to the kernel in those cases sounds alright to me.

> Why should an abstraction leak failures it can detect?

If you want to, you can detect those cases, e.g.
```c++
case 'w':
    ASSERT(!(mode & O_RDONLY));
    mode |= O_WRONLY;
```
I'm fine with both doing and not doing this.

alimpfard: Well...that's a whole lot of asserts for a mere "er" support, I'm not convinced that the trade-off is worth the extra weirdness.
But stdio is practically yours, so I'll go ahead and do it your way.

bugaevc: > If we don't support crap like "awr", we shouldn't let it get down to the kernel before it gets possibly rejected (although I don't think there's any particular combination that would fail, since "awr" is just "a+" with our flags).

> But why leak crap when we can simply not?

To me this sounds like asking for `O_WRONLY | O_APPEND | O_CREAT | O_WRONLY | O_TRUNC | O_CREAT | O_RDONLY`, just specified as a string. That combination doesn't make sense, but why not pass it to the kernel and let it run its logic to decide whether it makes sense or not.

On the other hand it's `fopen()` that does the translation, so it might as well run some validation. But it's not "leaking crap" if it doesn't — it would just pass to the kernel the exact flags it has been asked to.

alimpfard: `fopen` is already being biased with regards to its input by ignoring things it doesn't understand, but alright, let's leave the does-this-flag-make-sense to the kernel.

alimpfard: Speaking of modes, where does this `rt` mode come from?
text would be the default mode anyway, and we don't even do anything with it

AK/StringView.cpp
```diff
@@ -75,6 +75,32 @@ Vector<StringView> StringView::split_view(const char separator, bool keep_empty)
...
+
+    ASSERT(!separator.is_empty());
+
+    StringView view { *this };
+
+    Vector<StringView> parts;
+
+    auto maybe_separator_index = find_first_of(separator);
+    while (maybe_separator_index.has_value()) {
+        auto separator_offset = maybe_separator_index.value();
```
bugaevc, line 91: `offset` vs `index`

alimpfard: Ah, that made sense before changing the view itself, not so much now

Libraries/LibC/stdio.cpp
bugaevc: No idea. It's not in POSIX ('b' is).

It was added in https://github.com/SerenityOS/serenity/commit/c2a8c4cedd8c682fb861b446d4bf1f151e97ee2d

AK/StringView.cpp
```diff
@@ -75,6 +75,32 @@ Vector<StringView> StringView::split_view(const char separator, bool keep_empty)
     return v;
 }
 
+Vector<StringView> StringView::split_view(const StringView& separator, bool keep_empty) const
+{
+    if (is_empty())
+        return {};
+
+    ASSERT(!separator.is_empty());
```
bugaevc, line 83: I'd move the assert above

alimpfard: Ehhhhhhhhhhhhhhhhhhh, I can't think of a reason not to, sure.

Libraries/LibC/stdio.cpp
alimpfard: Sounds like some program trying to be fancy, but ok.

Libraries/LibC/stdio.cpp
```diff
@@ -911,25 +910,43 @@ void perror(const char* s)
...
-    else if (!strcmp(mode, "a+") || !strcmp(mode, "ab+"))
-        flags = O_RDWR | O_APPEND | O_CREAT;
-    else {
-        dbg() << "Unexpected mode _" << mode << "_";
-        ASSERT_NOT_REACHED();
+    auto length = strlen(mode);
+    for (size_t i = 0; i < length; ++i) {
+        switch (mode[i]) {
+        case '\0':
+            return flags;
```
bugaevc, line 920: (I think this is the last nitpick!)

Since we return when we see the `'\0'`, we don't have to pre-calculate length. I'd just do

```c++
for (; *mode; mode++)
    ....
return flags;
```
But feel free to make it a `while` loop if you would prefer that

alimpfard: This one I can happily use `for` for (for for for for...)
It's just the multiline `for`'s that make me want to switch to a `while` loop

alimpfard: > (I think this is the last nitpick!)

It's not, the comment at the end is wrong now lol

bugaevc: I was hoping you'd catch that one yourself :wink:

bugaevc: (and you don't need the `case '\0'` here anymore)

alimpfard: I'll go ahead and add it back so you can add one more nitpick teehee

bugaevc: > and you don't need the `case '\0'` here anymore

Bump

alimpfard: oh. 
sorry, didn't notice that.

PR #2417, opened 2020-05-27
LibJS: Implement standard semantics for relational operators

nooga:
Previously, the relational operators where casting any value to double and comparing the results according to C++ semantics.

This patch makes the relational operators in JS behave according to the standard specification.

Since we don't have BigInt yet, the implementation doesn't take it into account.


Libraries/LibJS/Runtime/Value.cpp
```diff
@@ -673,4 +662,77 @@ bool abstract_eq(Interpreter& interpreter, Value lhs, Value rhs)
     return false;
 }
 
+TriState abstract_relation(Interpreter& interpreter, bool left_first, Value lhs, Value rhs)
+{
+    Value x_primitive = {};
+    Value y_primitive = {};
+
+    if (left_first) {
+        x_primitive = lhs.to_primitive(interpreter);
```
linusg, line 671: I think we should update `Value::to_primitive()` to take a `PreferredType` and pass that on to `object::to_primitive()`and then be explicit and give it the number hint mentioned in the spec.

Libraries/LibJS/Runtime/Value.cpp
```diff
@@ -673,4 +662,77 @@ bool abstract_eq(Interpreter& interpreter, Value lhs, Value rhs)
...
+    } else {
+        y_primitive = lhs.to_primitive(interpreter);
+        if (interpreter.exception())
+            return {};
+        x_primitive = rhs.to_primitive(interpreter);
+        if (interpreter.exception())
+            return {};
+    }
+
+    if (x_primitive.type() == Value::Type::String && y_primitive.type() == Value::Type::String) {
```
linusg, line 686:
```suggestion
    if (x_primitive.is_string() && y_primitive.is_string()) {
```

Libraries/LibJS/Runtime/Value.cpp
```diff
@@ -673,4 +662,77 @@ bool abstract_eq(Interpreter& interpreter, Value lhs, Value rhs)
...
+            return TriState::False;
+        if (y_string.starts_with(x_string))
+            return TriState::True;
+
+        Utf8View x_codepoints { x_string };
+        Utf8View y_codepoints { y_string };
+        for (auto k = x_codepoints.begin(), l = y_codepoints.begin();
+             k != x_codepoints.end() && l != y_codepoints.end();
+             ++k, ++l) {
+            if (*k != *l) {
```
linusg, line 700: this looks weird - is clang-format doing this?

Libraries/LibJS/Runtime/Value.cpp
```diff
@@ -673,4 +662,77 @@ bool abstract_eq(Interpreter& interpreter, Value lhs, Value rhs)
...
+                    return TriState::False;
+                }
+            }
+        }
+        ASSERT_NOT_REACHED();
+    }
+
+    // FIXME add BigInt cases here once we have BigInt
+
+    Value x_numeric = x_primitive.to_number(interpreter);
```
linusg, line 713: Not that it matters that much, but I'd say Serenity code should try to leverage `auto` as much as possible - 

```suggestion
    auto x_numeric = x_primitive.to_number(interpreter);
```

same for `double x_value = x_numeric.as_double();` - looks like Java :smile:

Libraries/LibJS/Runtime/ArrayPrototype.cpp
```diff
@@ -706,8 +706,7 @@ Value ArrayPrototype::splice(Interpreter& interpreter)
                 this_object->put(to, from);
                 if (interpreter.exception())
                     return {};
-            }
-            else
+            } else
```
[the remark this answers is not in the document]
nooga: I rebased onto the master and ran clang-format so this just happened :D

Libraries/LibJS/Runtime/Value.cpp
nooga: Yeah, that would be useful. Would you like to do it as a part of this PR?

Libraries/LibJS/Runtime/Value.cpp
nooga: I was unsure if we want a super long line so I've broken it up, clang-format left it as it is.

linusg: either way is fine :+1:

Libraries/LibJS/Runtime/Value.cpp
linusg: sure, if you don't mind :)

nooga:
@linusg please have a look. I've moved some things around in order to handle the preferred type.

nooga:
Going to squash them anyway.

PR #2420, opened 2020-05-27
LibJS: Object index properties have descriptors; Handle sparse indices

### LINKED ISSUE #2330: LibJS: Allow index properties to have descriptors

mattco98:
This patch adds an `IndexedProperties` object for storing indexed properties within an `Object`. This accomplishes two goals: indexed properties now have an associated descriptor, and objects now gracefully handle sparse properties.

The `IndexedProperties` class is a wrapper around two other classes, one for simple indexed properties storage, and one for general indexed property storage. Simple indexed property storage is the common-case, and is simply a vector of properties which all have attributes of `default_attributes` (writable, enumerable, and configurable).

General indexed property storage is for a collection of indexed properties where EITHER one or more properties have attributes other than `default_attributes` OR there is a property with a large index (in particular, large is `200` or higher).

Indexed properties are now treated relatively the same as storage within the various `Object` methods. Additionally, there is a custom iterator class for `IndexedProperties` which makes iteration easy. The iterator skips empty values by default, but can be configured otherwise. Likewise, it evaluates getters by default, but can be set not to.

Closes #2330.

Things that need to be done after this PR is merged:

- Comprehensive exception checking. With this PR, all get and put operations on an object have side effects, and thus can throw exceptions. Everywhere a get or put is called, an exception check must follow.
- A review of `const` types. All `get`-type methods right now are const, but because they can have side effects, they really shouldn't be `const`. This would get rid of a few `const_cast`s scattered around in `Object.cpp`.
- Respect the `enumerable` attribute where necessary. This is something I didn't really try to do in the PR, since it is a problem throughout LibJS. This should be handled in a separate PR.

### DIFF
--- a/Libraries/LibJS/AST.cpp
+++ b/Libraries/LibJS/AST.cpp
@@ -57,9 +57,8 @@ static void update_function_name(Value& value, const FlyString& name)
             function.set_name(name);
     } else if (object.is_array()) {
         auto& array = static_cast<Array&>(object);
-        for (size_t i = 0; i < array.elements().size(); ++i) {
-            update_function_name(array.elements()[i], name);
-        }
+        for (auto& entry : array.indexed_properties().values_unordered())
+            update_function_name(entry.value, name);
     }
 }
 
@@ -142,20 +141,22 @@ Value CallExpression::execute(Interpreter& interpreter) const
             return {};
         if (m_arguments[i].is_spread) {
             // FIXME: Support generic iterables
-            Vector<Value> iterables;
             if (value.is_string()) {
                 for (auto ch : value.as_string().string())
-                    iterables.append(Value(js_string(interpreter, String::format("%c", ch))));
+                    arguments.append(Value(js_string(interpreter, String::format("%c", ch))));
             } else if (value.is_object() && value.as_object().is_array()) {
-                iterables = static_cast<const Array&>(value.as_object()).elements();
+                auto& array = static_cast<Array&>(value.as_object());
+                for (auto& entry : array.indexed_properties()) {
+                    arguments.append(entry.value_and_attributes(&array).value);
+                    if (interpreter.exception())
+                        return {};
+                }
             } else if (value.is_object() && value.as_object().is_string_object()) {
                 for (auto ch : static_cast<const StringObject&>(value.as_object()).primitive_string().string())
-                    iterables.append(Value(js_string(interpreter, String::format("%c", ch))));
+                    arguments.append(Value(js_string(interpreter, String::format("%c", ch))));
             } else {
                 interpreter.throw_exception<TypeError>(String::format("%s is not iterable", value.to_string_without_side_effects().characters()));
             }
-            for (auto& value : iterables)
-                arguments.append(value);
         } else {
             arguments.append(value);
         }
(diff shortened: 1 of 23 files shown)


linusg:
> Comprehensive exception checking. With this PR, all get and put operations on an object have side effects, and thus can throw exceptions. Everywhere a get or put is called, an exception check must follow.

Isn't that already the case?!

mattco98:
Some discussion points for this PR:

- Is 200 a good index limit to automatically switch from simple to generic index storage?
- I only ever increase the index packed storage array, as I figure I wouldn't want to shrink and then increase the array in loops. Is this a good design?
- I only increase the index packed storage by a certain fixed number (20 right now) if necessary, as I figured it'd be bad to resize the array repeatedly in loops. Is this a good design choice? Is 20 a good number?

mattco98:
> Isn't that already the case?!

You ninja'd me! Not quite sure what you're referring too -- if it's the fact that they already have side effects, then yes. Idk why I worded it that way. If it's that there are already exception checks everywhere, then I wouldn't be too sure... I feel like I've seen at least a few spots where they're missing.

awesomekling:
> * Is 200 a good index limit to automatically switch from simple to generic index storage?

Seems like a fine place to start. It's easy to tune this as we learn more about our performance on real world content. :)

> * I only ever increase the index packed storage array, as I figure I wouldn't want to shrink and then increase the array in loops. Is this a good design choice?

Same thing here. Fine place to start. Could be an interesting area for optimization later if it leads to pathological behavior somewhere.

> * I only increase the index packed storage by a certain fixed number (20 right now) if necessary, as I figured it'd be bad to resize the array repeatedly in loops. Is this a good design choice? Is 20 a good number?

This is definitely a good design choice. I might suggest increasing it by a percentage rather than a fixed number, but again, this is fine for a first cut. :)

awesomekling:
Great work on this patch. It fixes an architectural flaw with the object property system without excessively increasing the cognitive load. :)

PR #2425, opened 2020-05-28
LibJS: Add strict mode

mattco98:
Adds the ability for a scope (either a function or the entire program) to be in strict mode. Scopes default to non-strict mode.

There are two ways to determine the strictness of the JS engine:

1. In the parser, this can be accessed with the parser_state variable `m_is_strict_mode` boolean. If true, the Parser is currently parsing in strict mode. This is done so that the Parser can generate syntax errors at parse time, which is required in some cases. The lexer will also need this information to determine whether or not to tokenize octal escapes. 

2. With `Interpreter.is_strict_mode()`. This allows strict mode checking at runtime as opposed to compile time.

Additionally, in order to test this, a global `isStrictMode()` function has been added to the JS `ReplObject` under the test-mode flag.

This PR also includes a small strict-mode change, just to demonstrate the functionality: left-hand-side identifier references cannot be the string `eval` or `arguments` under strict mode.

### DIFF
--- a/Libraries/LibJS/AST.h
+++ b/Libraries/LibJS/AST.h
@@ -120,6 +120,9 @@ public:
     void add_variables(NonnullRefPtrVector<VariableDeclaration>);
     const NonnullRefPtrVector<VariableDeclaration>& variables() const { return m_variables; }
 
+    bool in_strict_mode() const { return m_strict_mode; }
+    void set_strict_mode() { m_strict_mode = true; }
+
 protected:
     ScopeNode() { }
 
@@ -127,6 +130,7 @@ private:
     virtual bool is_scope_node() const final { return true; }
     NonnullRefPtrVector<Statement> m_children;
     NonnullRefPtrVector<VariableDeclaration> m_variables;
+    bool m_strict_mode { false };
 };
 
 class Program : public ScopeNode {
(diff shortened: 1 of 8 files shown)


Libraries/LibJS/AST.h
```diff
@@ -120,13 +120,17 @@ class ScopeNode : public Statement {
     void add_variables(NonnullRefPtrVector<VariableDeclaration>);
     const NonnullRefPtrVector<VariableDeclaration>& variables() const { return m_variables; }
 
+    bool is_strict_mode() const { return m_is_strict_mode; }
```
awesomekling, line 123: Let's call this `in_strict_mode()` so it reads more naturally.
And `m_strict_mode` for the member.

Same comment repeated for all occurrences of the pattern :)

Libraries/LibJS/Tests/program-strict-mode.js
```diff
@@ -0,0 +1,30 @@
...
+try {
+    assert(isStrictMode());
+
+    (function() {
+       assert(isStrictMode());
+    })();
+
+    (function() {
+        "use strict";
+       assert(isStrictMode());
```
linusg, line 14:
```suggestion
        assert(isStrictMode());
```

PR #2430, opened 2020-05-28
Mounts, part 4

bugaevc:
* Read-only FS mounts
* Mount `/` read-only, bind-mount `/home` read-write
* Remounting
* Misc fixes

Caveats:
* It's currently possible to circumvent read-only mounts using `mprotect(PROT_WRITE)`, we have to think of a solution
* Remounting `/` as read-write doesn't do what you think it does

cc @alimpfard if you want some revenge for https://github.com/SerenityOS/serenity/pull/2416

### DIFF
--- a/Base/etc/fstab
+++ b/Base/etc/fstab
@@ -1,10 +1,12 @@
 # Root file system. This is a fake entry which gets ignored by `mount -a`;
 # the actual logic for mounting root is in the kernel.
-/dev/hda	/	ext2	nodev,nosuid
-# Remount /bin and /dev while adding the appropriate permissions.
-/dev	/dev	bind	bind,nosuid
-/bin	/bin	bind	bind,nodev
+/dev/hda	/	ext2	nodev,nosuid,ro
+# Remount /bin, /dev, /root, and /home while adding the appropriate permissions.
+/dev	/dev	bind	bind,nosuid,ro
+/bin	/bin	bind	bind,nodev,ro
+/home	/home	bind	bind,nodev,nosuid
+/root	/root	bind	bind,nodev,nosuid
 
 none	/proc	proc	nosuid
-none	/dev/pts	devpts	noexec,nosuid
+none	/dev/pts	devpts	noexec,nosuid,ro
 none	/tmp	tmp	nodev,nosuid
--- a/Kernel/UnixTypes.h
+++ b/Kernel/UnixTypes.h
@@ -52,6 +52,7 @@
 #define MS_NOEXEC (1 << 1)
 #define MS_NOSUID (1 << 2)
 #define MS_BIND (1 << 3)
+#define MS_RDONLY (1 << 4)
 
 #define PERF_EVENT_MALLOC 1
 #define PERF_EVENT_FREE 2
--- a/Kernel/UnixTypes.h
+++ b/Kernel/UnixTypes.h
@@ -28,6 +28,31 @@
 
 #include <AK/Types.h>
 
+#define O_RDONLY (1 << 0)
+#define O_WRONLY (1 << 1)
+#define O_RDWR (O_RDONLY | O_WRONLY)
+#define O_ACCMODE (O_RDONLY | O_WRONLY)
+#define O_EXEC (1 << 2)
+#define O_CREAT (1 << 3)
+#define O_EXCL (1 << 4)
+#define O_NOCTTY (1 << 5)
+#define O_TRUNC (1 << 6)
+#define O_APPEND (1 << 7)
+#define O_NONBLOCK (1 << 8)
+#define O_DIRECTORY (1 << 9)
+#define O_NOFOLLOW (1 << 10)
+#define O_CLOEXEC (1 << 11)
+#define O_DIRECT (1 << 12)
+
+// Kernel internal options.
+#define O_NOFOLLOW_NOERROR (1 << 29)
+#define O_UNLINK_INTERNAL (1 << 30)
+
+#define MS_NODEV (1 << 0)
+#define MS_NOEXEC (1 << 1)
+#define MS_NOSUID (1 << 2)
+#define MS_BIND (1 << 3)
+
 #define PERF_EVENT_MALLOC 1
 #define PERF_EVENT_FREE 2
 
--- a/Kernel/UnixTypes.h
+++ b/Kernel/UnixTypes.h
@@ -53,6 +53,7 @@
 #define MS_NOSUID (1 << 2)
 #define MS_BIND (1 << 3)
 #define MS_RDONLY (1 << 4)
+#define MS_REMOUNT (1 << 5)
 
 #define PERF_EVENT_MALLOC 1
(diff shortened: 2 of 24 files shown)


alimpfard:
Oh no code review.
That's more difficult than writing the code...let's seeeeeee.

Kernel/UnixTypes.h
```diff
@@ -28,6 +28,31 @@
...
+#define O_NOCTTY (1 << 5)
+#define O_TRUNC (1 << 6)
+#define O_APPEND (1 << 7)
+#define O_NONBLOCK (1 << 8)
+#define O_DIRECTORY (1 << 9)
+#define O_NOFOLLOW (1 << 10)
+#define O_CLOEXEC (1 << 11)
+#define O_DIRECT (1 << 12)
+
+// Kernel internal options
```
alimpfard, line 47: Can I have a dot?

bugaevc: Wasn't me, I just moved this from the other file. But sure, I can add a dot :smile:

Userland/chroot.cpp
```diff
@@ -57,6 +57,8 @@ int main(int argc, char** argv)
                     flags |= MS_NOEXEC;
                 else if (part == "nosuid")
                     flags |= MS_NOSUID;
+                else if (part == "ro")
+                    flags |= MS_RDONLY;
```
alimpfard, line 61: Just curious, of what use is chrooting read-only?

Kernel/UnixTypes.h
alimpfard: yay!

Userland/chroot.cpp
bugaevc: Umm, if you want to run some process(es) in with another directory as the root where they cannot write to that directory? I mean, for the exact same things read-only root and chroot are useful separately.

Base/etc/fstab
```diff
@@ -1,10 +1,12 @@
 # Root file system. This is a fake entry which gets ignored by `mount -a`;
 # the actual logic for mounting root is in the kernel.
-/dev/hda	/	ext2	nodev,nosuid
-# Remount /bin and /dev while adding the appropriate permissions.
-/dev	/dev	bind	bind,nosuid
-/bin	/bin	bind	bind,nodev
+/dev/hda	/	ext2	nodev,nosuid,ro
+# Remount /bin, /dev, /root, and /home while adding the appropriate permissions.
```
alimpfard, line 4: Shouldn't `/etc` be rw for root?
Similarly, I don't understand why we're limiting `/usr` to readonly

Or am I completely misunderstanding this?

Userland/chroot.cpp
bugaevc: That being said, it's probably not terribly useful for us in practice (in fact, we didn't end up using `chroot()` for isolation at all, because we have `unveil()`) — but it would be weird not to support it if we support other mount flags.

Rather, ask yourself: if you chroot someone to limit what they can access, do you want them to always be able to write to that directory? Sometimes, surely, yes, other times no. Principle of least privilege and all that.

alimpfard: I guess it's okay to support :shrug:

Base/etc/fstab
bugaevc: You are misunderstanding this.

Are you familiar with Android internals? Or SIP on Darwin/macOS? This is something similar.

If you do need to edit `/etc` or install new software into root, you can explicitly opt into modifying it:

```sh
$ chroot / -o rw
$ te /etc/foo
```

but by default, even processes running as root cannot modify those files.

Libraries/LibCore/EventLoop.cpp
```diff
@@ -233,20 +233,10 @@ EventLoop::EventLoop()
...
         s_event_loop_stack->append(this);
 
-        auto rpc_path = String::format("/tmp/rpc.%d", getpid());
-        rc = unlink(rpc_path.characters());
-        if (rc < 0 && errno != ENOENT) {
-            perror("unlink");
-            ASSERT_NOT_REACHED();
+        if (!s_rpc_server) {
+            if (!start_rpc_server())
+                dbg() << "Failed to start a RPC server";
```
alimpfard, line 239: Would be nice to know who failed to start _an_ RPC server.

Also, the commit message needs some love

Base/etc/fstab
bugaevc: Oh, alternatively you could bind-mount `/etc` as writable when you need it:

```sh
$ mount /etc /etc -o bind
```

Libraries/LibCore/EventLoop.cpp
bugaevc: `dbg()` outputs the name and PID of the process, so you get it automatically.

Base/etc/fstab
alimpfard: Oh I see.
would be a bit weird as I don't think we're interested in _not_ letting the user change their configs easily, but I guess a mount for editing is a fine step to go through.

Libraries/LibCore/EventLoop.cpp
bugaevc: > Also, the commit message needs some love

what do you mean, it looks fine to me?

alimpfard: I was thinking `"EventLoop: Failed to do whatever"`, but that was just a nitpick to get the "the commit message needs some love" nitpick attached to the commit :)

alimpfard: "LibCore: Do not assert we can start the RPC server" reads like a few things to me, I guess you can make it better by adding a "that" after assert

just pointing that out since the sentence gave me a pause.

bugaevc: > I was thinking "EventLoop: Failed to do whatever"

Hmm, I was assuming everyone knows what is the RPC server in context of Serenity, but sure, let's spell it out.

> "the commit message needs some love" nitpick attached to the commit

Some more details, please? :slightly_frowning_face:

bugaevc: Inded, thanks

alimpfard: > I was assuming everyone knows what is the RPC server in context of Serenity

Think of the lives of the poor tech support people you'll save doing this :P 

> Some more details, please?

You just omitted a "that" in the message and my brain refused to parse it for a few seconds.
"LibCore: Do not assert that we can start the RPC server"

that's all.

alimpfard:
This is it from me.
I'm not particularly familiar with the VFS, so I can't really validate anything specific to that, but the general check-for-error-checks test passed!

awesomekling:
Hey, very neat! For the `sys$mprotect` issue, maybe we can remember the underlying fs permissions when first creating a `Region`?

bugaevc:
> maybe we can remember the underlying fs permissions when first creating a Region?

Yeah that probably works

PR #2431, opened 2020-05-28
LibJS: Add {Object,Array}.prototype.toLocaleString()

linusg:


Libraries/LibJS/Runtime/Object.cpp
```diff
@@ -662,4 +662,17 @@ Value Object::to_string() const
     return js_string(heap(), String::format("[object %s]", class_name()));
 }
 
+Value Object::invoke(const FlyString& property_name, Optional<MarkedValueList> arguments) const
```
awesomekling, line 665: This function seems like it should not const qualified.

Libraries/LibJS/Runtime/Object.cpp
```diff
@@ -662,4 +662,17 @@ Value Object::to_string() const
...
+{
+    auto& interpreter = const_cast<Object*>(this)->interpreter();
```
awesomekling, line 667: Then we could just use `interpreter()` without `const_cast`.

Libraries/LibJS/Runtime/Object.cpp
```diff
@@ -662,4 +662,17 @@ Value Object::to_string() const
...
+    auto property = get(property_name).value_or(js_undefined());
+    if (interpreter.exception())
+        return {};
+    if (!property.is_function()) {
+        interpreter.throw_exception<TypeError>(String::format("%s is not a function", property.to_string_without_side_effects().characters()));
+        return {};
+    }
+    return interpreter.call(property.as_function(), const_cast<Object*>(this), move(arguments));
```
awesomekling, line 675: And no need for the `const_cast` here either :)

PR #2432, opened 2020-05-28
Meta: Make travis also check for the presense of "#pragma once" in header files.

emanuele6:
We can expand this script to also force one code styles, where `clang-format` allows more than one, once we agree on which one to use. e.g.: 
```cpp
class A {
    C_OBJECT(A)
public:
}
/* ---------------- */
class A {
    C_OBJECT(A)

public:
}
/* ---------------- */ // clang-format adds a blank line when a semicolon is used
class A {
    C_OBJECT(A);

public:
}
```
```cpp
class A
    : B
    , C {
}
/* ---------------- */
class A : B
    , C {
}
```
---------------------------
NOTE: I didn't add `#pragma once` to `Kernel/FileSystem/ext2_fs.h` and `Libraries/LibELF/exec_elf.h` (which are now the only header files without `#pragma once` in the whole repository, see command below), because they already have include guards and we are not checking them.
```sh
git ls-files -- '*.h' | xargs -I{} awk '/#pragma once/ { a = 1; exit } END { if (!a) { print "{}" } }' {}
```


Meta/check-style.sh
```diff
@@ -0,0 +1,56 @@
...
+#  * Copyright
+LICENSE_HEADER_PATTERN=$'^/\*\n \* Copyright'
+MISSING_LICENSE_HEADER_ERRORS=()
+
+# We check that "#pragma once" is present
+PRAGMA_ONCE_PATTERN='#pragma once'
+MISSING_PRAGMA_ONCE_ERRORS=()
+
+# We make sure that there's a blank line before and after pragma once
+GOOD_PRAGMA_ONCE_PATTERN=$'(^|\S\n\n)#pragma once(\n.|$)'
```
emanuele6, line 17: This should be `(^|\S\n\n)#pragma once(\n\n\S|$)`: it wasn't actually checking the presence of a blank line, it also allowed the abscence of `\n` at the EOF.  
Let's also replace `\S` with `\\S` just in case.
```suggestion
GOOD_PRAGMA_ONCE_PATTERN=$'(^|\S\n\n)#pragma once(|$)'
```

awesomekling:
LGTM

PR #2433, opened 2020-05-28
PixelPaint: Set active layer to nullptr after its removal

gabrielgggg:
PixelPaint would crash if "Remove Active Layer" would be used twice, without selecting a new layer in the meanwhile.


emanuele6:
Please, squash your commits.

awesomekling:
:star:

PR #2434, opened 2020-05-28
LibJS: Parse and handle labelled statements

mattco98:
This patch introduces labels to the JS engine. All statements can be labelled, and most statements (not functions) can be broken out of. 

Known label issues:

1. Attempting to type something like the following in the REPL does not compile on line 1 (I'm not sure how to fix this; I'll open an issue if this PR gets merged):

```js
label: 
{
}
```

2. Function labels should not be allowed in strict mode. To do this, I'm thinking of introducing a way to test syntax errors. My thought is that we can have a `LibJS/Tests/SyntaxErrors` folder, where every file has to produce a syntax error to pass.

I'm sure there are more, but this is just the initial feature introduction. Other issues can be fixed as they get discovered.

### DIFF
--- a/Libraries/LibJS/Interpreter.cpp
+++ b/Libraries/LibJS/Interpreter.cpp
@@ -72,8 +72,11 @@ Value Interpreter::run(const Statement& statement, ArgumentVector arguments, Sco
     m_last_value = js_undefined();
     for (auto& node : block.children()) {
         m_last_value = node.execute(*this);
-        if (m_unwind_until != ScopeType::None)
+        if (should_unwind()) {
+            if (should_unwind_until(ScopeType::Breakable, block.label()))
+                stop_unwind();
             break;
+        }
     }
 
     bool did_return = m_unwind_until == ScopeType::Function;
--- a/Libraries/LibJS/Parser.cpp
+++ b/Libraries/LibJS/Parser.cpp
@@ -269,6 +269,9 @@ NonnullRefPtr<Statement> Parser::parse_statement()
         consume();
         return create_ast_node<EmptyStatement>();
     default:
+        auto result = try_parse_labelled_statement();
+        if (!result.is_null())
+            return result.release_nonnull();
         if (match_expression()) {
             auto expr = parse_expression(0);
             consume_or_insert_semicolon();
@@ -381,6 +384,27 @@ RefPtr<FunctionExpression> Parser::try_parse_arrow_function_expression(bool expe
     }
 
     return nullptr;
+}
+
+RefPtr<Statement> Parser::try_parse_labelled_statement()
+{
+    save_state();
+    ArmedScopeGuard state_rollback_guard = [&] {
+        load_state();
+    };
+
+    auto identifier = consume(TokenType::Identifier).value();
+    if (!match(TokenType::Colon))
+        return {};
+    consume(TokenType::Colon);
+
+    if (!match_statement())
+        return {};
+    auto statement = parse_statement();
+
+    statement->set_label(identifier);
+    state_rollback_guard.disarm();
+    return statement;
 }
 
 NonnullRefPtr<Expression> Parser::parse_primary_expression()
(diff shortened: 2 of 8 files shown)


Dexesttp:
>Function labels should not be allowed in strict mode. To do this, I'm thinking of introducing a way to test syntax errors. My thought is that we can have a LibJS/Tests/SyntaxErrors folder, where every file has to produce a syntax error to pass.

If I'm not mistaken, calling `new Function("<js code>")` can raise `SyntaxError` exceptions ([like here, where it's testing for an error on `[`](https://github.com/SerenityOS/serenity/blob/master/Libraries/LibJS/Tests/Function.js#L29)). I don't know how it would interact with safe mode, but maybe it can work for your test cases ?

mattco98:
Ohh, nice tip! Function constructors play well with the `'use strict'` directive, so no worries there. Plus that allows me to test the message of the SyntaxError. Thanks!

sunverwerth:
Alright, it's review time

Libraries/LibJS/Interpreter.cpp
```diff
@@ -62,7 +62,7 @@ Value Interpreter::run(const Statement& statement, ArgumentVector arguments, Sco
             m_call_stack.append(move(global_call_frame));
         }
     }
-
+
```
sunverwerth, line 65: nit: Remove whitespace

Libraries/LibJS/Parser.cpp
```diff
@@ -269,6 +269,9 @@ NonnullRefPtr<Statement> Parser::parse_statement()
         consume();
         return create_ast_node<EmptyStatement>();
     default:
+        auto result = try_parse_labelled_statement();
```
sunverwerth, line 272: This should be wrapped in `if (match(TokenType::Identifier))` so we do not have to invoke the whole state pushing mechanism for every expression.

sunverwerth: Ideally we'd use token lookahead to match `Identifier` `Colon` directly, but oh well. Maybe in a next iteration of the parser.

Libraries/LibJS/Parser.cpp
```diff
@@ -1089,17 +1113,29 @@ NonnullRefPtr<ThrowStatement> Parser::parse_throw_statement()
 NonnullRefPtr<BreakStatement> Parser::parse_break_statement()
 {
     consume(TokenType::Break);
+    FlyString target_label;
+    if (match(TokenType::Semicolon)) {
+        consume();
+        return create_ast_node<BreakStatement>(target_label);
+    }
+    if (match_identifier_name() && !m_parser_state.m_current_token.trivia().contains('\n'))
```
sunverwerth, line 1121: `match(TokenType::Identifier)` instead of `match_identifier_name()`. Labels can not be reserved keywords.

Libraries/LibJS/Parser.cpp
```diff
@@ -1089,17 +1113,29 @@ NonnullRefPtr<ThrowStatement> Parser::parse_throw_statement()
...
 
 NonnullRefPtr<ContinueStatement> Parser::parse_continue_statement()
 {
     consume(TokenType::Continue);
+    FlyString target_label;
+    if (match(TokenType::Semicolon)) {
+        consume();
+        return create_ast_node<ContinueStatement>(target_label);
+    }
+    if (match_identifier_name() && !m_parser_state.m_current_token.trivia().contains('\n'))
```
sunverwerth, line 1135: See above. Use `match(TokenType::Identifier)`

sunverwerth:
A general comment: I feel that adding a potential label to every statement node is a bit overkill for such an obscure feature. It's just a FlyString so propably not that big a deal, but maybe we should think about another way to do this going forward.

Opinions?

mattco98:
Maybe some kind of `Labellable` class that can be inherited by nodes who actually care about labels? So labels like `label: 1 + 2` don't actually get stored anywhere because `Expression` wouldn't inherit from `Labellable`.

sunverwerth:
I was thinking more about storing a map of Label->Statement in the scope.
But that could be a separate change, if at all.
The way you've done it here is pretty idiomatic I think.

awesomekling:
Yeah this seems okay for now. Let's not sweat about the memory usage of our AST too much since we'll probably translate it to bytecode later on anyway, once the language support has matured. :)

PR #2436, opened 2020-05-28
LibJS: New expressions look for expressions with correct precedence

mattco98:
This seems too easy, but from all my testing it seems to work correctly with precedence. The reason there aren't any tests with this PR is because I can't really test it without either 1) being able to test files for syntax errors or 2) new expressions actually working with proper `this` values. So until one of those things happens, this will be test-less.

Also, shout-out to google for easy bug hunting :)


mattco98:
Ok yeah turns out it wasn't that easy :) Everything works but `new Foo()`, the call gets evaluated and _then_ `new`'d. I'll leave this as a draft for now and see if I can get calls working.

PR #2440, opened 2020-05-29
Core: Notifier::Exceptional; Shell: more completion stuff

alimpfard:
- `EINTR` treated as "exceptional event" for Notifiers, callback added
- LibLine: ^C works again
- Shell: redirection path completion
```
$ ls foob*
foobar
$ foo > "foob<tab>
$ foo > "foobar"
```

update: @bugaevc pls

### DIFF
--- a/Libraries/LibCore/Event.h
+++ b/Libraries/LibCore/Event.h
@@ -42,18 +42,19 @@ public:
         Timer,
         NotifierRead,
         NotifierWrite,
+        NotifierExceptional,
         DeferredInvoke,
         ChildAdded,
         ChildRemoved,
         Custom,
     };
 
-    Event() {}
+    Event() { }
     explicit Event(unsigned type)
         : m_type(type)
     {
     }
-    virtual ~Event() {}
+    virtual ~Event() { }
 
     unsigned type() const { return m_type; }
 
@@ -87,7 +88,7 @@ public:
         , m_timer_id(timer_id)
     {
     }
-    ~TimerEvent() {}
+    ~TimerEvent() { }
 
     int timer_id() const { return m_timer_id; }
 
@@ -102,7 +103,7 @@ public:
         , m_fd(fd)
     {
     }
-    ~NotifierReadEvent() {}
+    ~NotifierReadEvent() { }
 
     int fd() const { return m_fd; }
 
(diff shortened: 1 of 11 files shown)


bugaevc:
Isn't `Notifier::Exceptional` meant to represent read/write errors?

alimpfard:
_technically_, `EINTR` is an error...
I could add another mask, I just saw that it wasn't used and took it for myself :P

bugaevc:
I'd expect a signal notifier to not be tied to any fd at all — you'd create a separate notifier and say which signals you're interested in.

alimpfard:
This is not a signal notifier, this is just a "notify me when your select was interrupted".
I briefly thought about adding signal notifiers, but that would be pretty weird to have in an event loop, as...you know, event loops are not immediate and signals are.

bugaevc:
You don't have to actually handle the signal (notify the notifier) from the signal handler, you could just record somewhere that the signal was received, and then process it when you get `-EINTR` back from the syscall

alimpfard:
Furthermore, what _should_ `Exceptional` notify you of?
Notifier does not do any reads, nor any writes, it only tells you if you can do those, and the only failure point is `select`...I suppose I could flesh it out a little and have it include errno in the SyscallException object, and remove that assertion

alimpfard:
> and then process it when you get -EINTR back from the syscall

That completely disables the _urgent_ meaning of signals, the event queue could be pretty damn saturated and then the event would be dispatched waaaaaaaaaaaay too late

bugaevc:
> Furthermore, what should Exceptional notify you of?
> Notifier does not do any reads, nor any writes, it only tells you if you can do those, and the only failure point is select

True; perhaps Andreas was thinking of the `poll()` syscall which does report errors at `poll()` time (and even calls that "exceptional condition").

> That completely disables the urgent meaning of signals, the event queue could be pretty damn saturated and then the event would be dispatched waaaaaaaaaaaay too late

So when `select()` returns, check for pending signals first and dispatch I/O later.

alimpfard:
> So when select() returns, check for pending signals first and dispatch I/O later.

Sure, why not.

alimpfard:
Closing this as I don't feel like dealing with SignalNotifiers

PR #2441, opened 2020-05-29
LibCrypto: Fix some issues preventing compilation with clang 10

alimpfard:
Yuck.


alimpfard:
I'm on clang 10, I dunno about clang 11.

Issue #91, opened 2019-05-25
LibC: Rewrite getopt()


awesomekling:
We're currently using a getopt() borrowed from some BSD. It should be replaced with our own.

rburchell:
It would be handy to grow getopt_long, too.

PR #2445, opened 2020-05-29
NIH: rewrite getopt

bugaevc:
Fixes https://github.com/SerenityOS/serenity/issues/91

cc @alimpfard :wink:

### DIFF
--- a/Libraries/LibC/getopt.cpp
+++ b/Libraries/LibC/getopt.cpp
@@ -1,610 +1,371 @@
-/*	$OpenBSD: getopt_long.c,v 1.26 2013/06/08 22:47:56 millert Exp $	*/
-/*	$NetBSD: getopt_long.c,v 1.15 2002/01/31 22:43:40 tv Exp $	*/
-
 /*
- * Copyright (c) 2002 Todd C. Miller <Todd.Miller@courtesan.com>
- *
- * Permission to use, copy, modify, and distribute this software for any
- * purpose with or without fee is hereby granted, provided that the above
- * copyright notice and this permission notice appear in all copies.
- *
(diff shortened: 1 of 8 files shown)


alimpfard:
I swear, if I see the silly globals that the old getopt had, I will...do nothing...and just nag about it

Libraries/LibC/getopt.cpp
```diff
@@ -1,610 +1,355 @@
...
-#endif
-
-#define PRINT_ERROR	((opterr) && (*options != ':'))
-
-#define FLAG_PERMUTE	0x01	/* permute non-options to the end of argv */
-#define FLAG_ALLARGS	0x02	/* treat non-options as args to option "-1" */
-#define FLAG_LONGONLY	0x04	/* operate as getopt_long_only */
+int opterr = 1;
+int optopt = 0;
+int optind = 1;
```
alimpfard, line 35: OH FFS

alimpfard:
I like the idea, but can we have `OptionParser` not depend on the `optind` and co globals?
You could synchronise them at the end of `getopt` and `getopt_long`, and perhaps just keep a `g_option_parser` or something that tracks the global stupidity of getopt?

bugaevc:
We could do that, but I thought that would be more boilerplate for little gain

alimpfard:
can we at least document the behaviour then?
"to reset `getopt{_long,}`, please set `optind` to 0"
or maybe call `serenity_getopt_reset`?

bugaevc:
> can we at least document the behaviour then?

Yes we can, and that reminds me, I forgot to write the man page, shame on me!

bugaevc:
Now with `optreset` and `optind = 0` support, extensive man pages, colorful error messages, and misc stuff.

bugaevc:
image

alimpfard:
Very cool stuff!

PR #2448, opened 2020-05-29
LibJS: Parse arrow function expression with correct precedence

nooga:
The parser was chomping on commas present after the arrow function expression. eg. `[x=>x,2]` would parse as `[x=>(x,2)]` instead of `[(x=>x),2]`.

This is not the case anymore. I've added a small test to prove this.


nooga:
@linusg sorry about that, I wasn't thinking when I commited this 😅

I've added a prettier config file and reformatted the affected files in an attempt to fix my error but they are still affected heavily.

nooga:
@alimpfard I'll rebase this one to master. #2447 is not related to this one but I had to branch out of it in order to build and test.

alimpfard:
Fair enough.

Issue #2449, opened 2020-05-30
High CPU usage in ProtocolServer after failed TLS-related request in Browser


deoxxa:
Accessing `gemini://80h.dev/` in the browser will put `ProtocolServer` into a state where it seems to spin forever (indirectly?) on `TLSv12::write_into_socket`. Turning on debug output for `Core::EventLoop` and the `TLSv12` functions then loading `gemini://80h.dev/` in the browser is enough to reproduce it. Interestingly this doesn't seem to spin forever if you do something like `pro gemini://80h.dev/`.

PR #2457, opened 2020-05-30
LibTLS: Timeout after 10 seconds while waiting for handshake completion

alimpfard:
We have to busy-wait for ServerHello after the server accepts our connection; this causes us to wait forever if the server decides to ignore our handshake request.
Timeout after 10 seconds to avoid too much CPU usage.

This is related to #2449, while it does not close that issue, this mitigation is needed regardless, as we do not want a handshake to continue forever.

### DIFF
--- a/Libraries/LibTLS/TLSv12.h
+++ b/Libraries/LibTLS/TLSv12.h
@@ -142,6 +142,7 @@ enum class Error : i8 {
     FeatureNotSupported = -17,
     DecryptionFailed = -20,
     NeedMoreData = -21,
+    TimedOut = -22,
 };
 
 enum class AlertLevel : u8 {
@@ -293,6 +294,8 @@ struct Context {
     StringView negotiated_alpn;
 
     size_t send_retries { 0 };
+
+    time_t handshake_initiation_timestamp { 0 };
 };
 
 class TLSv12 : public Core::Socket {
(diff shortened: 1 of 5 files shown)


Libraries/LibTLS/TLSv12.h
```diff
@@ -468,6 +471,8 @@ class TLSv12 : public Core::Socket {
     OwnPtr<Crypto::Cipher::AESCipher::CBCMode> m_aes_remote;
 
     bool m_has_scheduled_write_flush = false;
+
+    i32 m_max_wait_time_for_handshake = 10; // seconds
```
awesomekling, line 475: Please use `{}` for member variable initialization.
Also, let's encode the "// seconds" in the variable name, something like `m_max_time_to_wait_for_handshake_in_seconds`.

alimpfard:
I really have no idea _why_ the busy wait is needed, I'm certain that it shouldn't be...but no matter what I try, it just...doesn't work.
Perhaps someone other than me should spend some time trying to figure this out, as I might be thinking within the current architecture.

I really, _really_ don't get why the wait seems to be needed, and it confuses the heck out of me.

Libraries/LibTLS/TLSv12.h
alimpfard: Of course.

alimpfard: It's just that the line above used `=`, so I stuck to the local theme.
Should I change that too?

awesomekling: Please :)

bugaevc:
So if we no longer busy-wait, should https://github.com/SerenityOS/serenity/issues/2449 be considered fixed?

alimpfard:
We still have issues connecting to some gemini servers, so I opted to leave that issue alone.
But what the issue title targets is fixed

PR #2458, opened 2020-05-30
Userland/js: Prompt for more input on labels and object keys

### LINKED ISSUE #2452: LibJS: REPL doesn't allow hanging labels
The REPL doesn't allow a label to be entered on it's own line, as seen in the following image:

If the label precedes a statement on the same line, such as `label: while (...) {`, it works fine.

alimpfard:
Closes #2452

Not sure who to tag for this, so I'll just...not.

### DIFF
--- a/Userland/js.cpp
+++ b/Userland/js.cpp
@@ -86,8 +86,12 @@ String read_next_piece()
 {
     StringBuilder piece;
 
+    auto line_level_delta_for_next_line { 0 };
+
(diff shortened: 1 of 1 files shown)


alimpfard:
Note that this will allow nonsensical code like below, but I think that's probably fine?
```js
> "test":
> foo
```

awesomekling:
Seems fine?

PR #2460, opened 2020-05-30
HackStudio: Allow changing variable values in debugger

FalseHonesty:
This patch adds a context menu to variables in the debugger variable tree view that has an option to set the value of a variable. An input box will pop up asking for the new value of the variable, which is then parsed and used to set the actual variable.

### DIFF
--- a/DevTools/HackStudio/Debugger/VariablesModel.cpp
+++ b/DevTools/HackStudio/Debugger/VariablesModel.cpp
@@ -25,6 +25,8 @@
  */
 
 #include "VariablesModel.h"
+#include <LibGUI/Application.h>
+#include <LibGUI/MessageBox.h>
 
 GUI::ModelIndex VariablesModel::index(int row, int column, const GUI::ModelIndex& parent_index) const
 {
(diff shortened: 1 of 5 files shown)


DevTools/HackStudio/Debugger/VariablesModel.cpp
```diff
@@ -85,9 +87,45 @@ String variable_value_as_string(const DebugInfo::VariableInfo& variable)
...
 }
 
+void VariablesModel::set_variable_value(const GUI::ModelIndex& index, const StringView& string_value, GUI::Window* parent_window)
+{
+    auto variable = static_cast<const DebugInfo::VariableInfo*>(index.internal_data());
+
+    u32 new_value = 0;
+    bool success = false;
+
+    if (variable->type == "int") {
```
itamar8910, line 106: I think we can factor out the logic here to a lambda or a static function, early return would simplify the `success` logic.
Maybe something like `evaluate_variable_value` that returns an `Optional<u32>`?

DevTools/HackStudio/Debugger/VariablesModel.cpp
```diff
@@ -85,9 +87,45 @@ String variable_value_as_string(const DebugInfo::VariableInfo& variable)
...
+        if (string_value == "true" || string_value == "false") {
+            new_value = string_value == "true";
+            success = true;
+        } else {
+            success = false;
+        }
+    }
+
+    if (success) {
+        Debugger::the().session()->poke((u32*)variable->location_data.address, new_value);
```
itamar8910, line 118: `poke` can fail. I think we should at least `ASSERT` here :)

itamar8910: Also, no need for the `else`, we can just `return` after this.

DevTools/HackStudio/Debugger/VariablesModel.cpp
FalseHonesty: Definitely a good idea, makes the code much more manageable.

FalseHonesty:
> Very cool smile

Thank you, its based off some really great work that is the debugger :)

awesomekling:
:+1:

Issue #1747, opened 2020-04-11
Userspace: Shared Library Support


ADKaster:
The big checklist for  #13 :

There's no way this can be done in one PR...

Build system
   - [ ] Modify gcc/libgcc configuration to set the interpreter to /lib/ld-elf.so for shared objects
   - [ ] Modify the gcc spec files to enable position independent executables
   - [ ] Build LibC and LibM as shared objects, so that we can build libstdc++ as a shared object.
        - Note -> this item gets really annoying if we want /lib/ld-elf.so written in C++ ...
   - [ ] To not break existing code, set the default build spec to static? In top level Makefiles probably. 
   - [ ] Create a standard way in the Makefile to create a shared library
   - [ ] Create a workaround way in the Makefile to optionally build a library as both shared and static for different applications. Atomic conversion sounds like a head-chasing problem

Dynamic Loader Program (/lib/ld-elf.so)
  - [ ] Create entry point (assembly)
  - [ ] Move LibC/dlfcn.cpp implementation to ld-elf.so, and revert dlfcn.cpp to have weak-referenced stubs.
  - [ ] Relocate main executable's absolute offsets, then the dynamic loader itself.
  - [ ] Create an ELFDynamicObject for the main executable, using the _DYNAMIC symbol to locate the program headers
  - [ ] Try not to panic when realizing that if LibC is a DSO, then we'll have to handle shared object + main executable in the first go-around. If /lib/ld-elf.so wants to use any methods from LibC, it will also have to have them re-located first... So it really can't do that.

Thread Local Storage (TLS)
  - [ ] Read and re-read https://akkadia.org/drepper/tls.pdf
  - [ ] Move TLS block allocation from Kernel to userspace. LibC? Might involve some fun asm.
  - [ ] Allocate a TLS block for each shared object (starting with main executable)
  - [ ] Handle TLS_OFF relocations properly, now that we are actually doing TLS in userspace

Exception handling (.eh_frame)
  - [ ] Build libgcc_s.so? To enable unwind of eh_frame
  - [ ] OR -- Implement unwind tables manually, doesn't that sound fun? :D

bugaevc:
> set the interpreter to /lib/ld-elf.so

Could it be named something else? Like maybe `/bin/loader`...

> Build LibC and LibM as shared objects, so that we can build libstdc++ as a shared object.
> Note -> this item gets really annoying if we want /lib/ld-elf.so written in C++ ...

Wait, libstdc++? Why would we need that?

[...]

PR #2461, opened 2020-05-30
Dynamic Loader

itamar8910:
This is a draft for a work-in-progress attempt at a dynamic loader.
Relevant issue: #1747 

Changes made:
- When `exec` is called with an elf program that has an interpreter, the kernel loads the interpreter (In our case, the dynamic loader) instead, and passes control to it. The interpreter then takes care of loading the main program and its dependencies.
- The loader program is in: `Demos/DynamicLoader/Loader/`
- A demo dynamically-linked executable is in: `Demos/DynamicLoader/DynExec`. It has two dependencies: `ExampleSO`, and `LibC`.
Run `DynExec` to see it in action.

This is definitely not ready for review yet :smile:
However, discussion and opinions would be much appreciated - cc @ADKaster who did much of the work on dynamic loading.

With the current state of the loader, we only call a simple test function in `LibC` which simply returns an int.

Some issues with the current loader (anyone, feel free to raise other issues):
- The Loader itself is statically linked with `libc`. It gets loaded by the kernel as if it were one of our regular programs. While not ideal, I think this is a reasonable compromise that allows us to get a first version going more quickly.
- The loader is installed as `/usr/lib/libc.so.1`. This is because I didn't yet succeed to convince gcc to change the loader's path :sweat_smile:
- Normally, we would pass arguments from kernel to the dynamic loader via an "Auxiliary Vector". In the current version, since the loader uses the regular entry point of `libc`, we instead just use special environment variables.
- We do not initialize `LibC`. This is required for calling functions like `printf`.
- There is no support for TLS allocation for the shared objects we load.
I think @ADKaster suggested that the loader maintain some TLS memory pool, and allocate chunks of it to the shared objects, which sounds like a good way to go about this.


Demos/DynamicLoader/DynExec/CMakeLists.txt
```diff
@@ -0,0 +1,12 @@
+set(SOURCES
+    main.cpp
+)
+
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -nostdlib -pie")
+#set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -nostdlib -fpic -pie")
+#set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} --dynamic-linker=/bin/Loader")
```
ADKaster, line 7: I think you want this (or similar) in Toolchain/Patches/gcc.patch
```
diff --git a/Toolchain/Patches/gcc.patch b/Toolchain/Patches/gcc.patch
index 8e2dd8342..d22589fd1 100644
--- a/Toolchain/Patches/gcc.patch
+++ b/Toolchain/Patches/gcc.patch
@@ -121,14 +121,14 @@ index 000000000..2692b8e57
 +/* Files that are linked before user code.
 +   The %s tells GCC to look for these files in the library directory. */
 +#undef STARTFILE_SPEC
-+#define STARTFILE_SPEC "%{!shared:crt0.o%s} crti.o%s %{shared:crtbeginS.o%s; :crtbegin.o%s}"
++#define STARTFILE_SPEC "%{!shared:crt0.o%s} crti.o%s %{shared|pie:crtbeginS.o%s; :crtbegin.o%s}"
 +
 +/* Files that are linked after user code. */
 +#undef ENDFILE_SPEC
-+#define ENDFILE_SPEC "%{shared:crtendS.o%s; :crtend.o%s} crtn.o%s"
++#define ENDFILE_SPEC "%{shared|pie:crtendS.o%s; :crtend.o%s} crtn.o%s"
 +
 +#undef LINK_SPEC
-+#define LINK_SPEC "%{shared:-shared} %{static:-static} %{!shared: %{!static: %{rdynamic:-export-dynamic}}}"
++#define LINK_SPEC "%{shared:-shared} %{static:-static} %{!shared: %{!static: %{rdynamic:-export-dynamic} -dynamic-linker /lib/loader.so}}"
 +
 +/* Additional predefined macros. */
 +#undef TARGET_OS_CPP_BUILTINS
```

Libraries/LibELF/DynamicLoader.cpp
```diff
@@ -273,6 +290,16 @@ void DynamicLoader::do_relocations()
             *patch_ptr = relocation.offset() - (u32)m_tls_segment_address.as_ptr() - *patch_ptr;
             break;
         }
+        case R_386_TLS_DTPMOD3: {
```
ADKaster, line 293: If we do
```
set(CMAKE_POSITION_INDEPENDENT_CODE ON)
```
and
```
-ftls-model=initial-exec
```
in the top level CMakeLists.txt, then  we get 1) PIC code for everyone (still works with statically linked :) and 2) Forced Initial Executable TLS model (which generates the R_386_TLS_TPOFF we were handling before with the non-PIE executable
https://docs.oracle.com/cd/E19120-01/open.solaris/819-0690/chapter8-20/index.html
https://docs.oracle.com/cd/E19120-01/open.solaris/819-0690/chapter8-1/index.html

Libraries/LibELF/DynamicLoader.cpp
```diff
@@ -323,6 +354,31 @@ void DynamicLoader::setup_plt_trampoline()
...
+{
+    //
+    // At this stage, the .got.plt contains addresses that point back to
+    // the plt. This is a part of the lazy linking mechanism.
+    // The plt was probably loaded somewhere else than the compiler expected,
+    // so we need to relocate the addresses in .got.plt to point at the actual
+    // indtended parts in the plt.
+
+    // TODO: is this actually the way it's supposed to be done?
+    // why doesn't the compiler explicitly generate relocations for this?
```
ADKaster, line 367: Not sure if this is your TODO or mine, but it seems to be the correct way of getting the offset of _DYNAMIC and  setting up the PLT for the later R_386_JUMP_SLOT relocations

Demos/DynamicLoader/Loader/main.cpp
```diff
@@ -0,0 +1,84 @@
...
+    ASSERT(main_program_path_ptr);
+    String main_program_path(main_program_path_ptr);
+    char* main_program_fd_str = getenv("_MAIN_PROGRAM_FD");
+    ASSERT(main_program_fd_str);
+    int main_program_fd = atoi(main_program_fd_str);
+    lseek(main_program_fd, 0, SEEK_SET);
+    dbg() << "main_program: " << main_program_path << ", fd: " << main_program_fd;
+
+    printf("Loading main program\n");
+    void* res = serenity_dlopen(main_program_fd, main_program_path.characters(), RTLD_LAZY | RTLD_GLOBAL);
```
ADKaster, line 69: I think the loader main should have some kind of 
```
static alignas(Elf::DynamicObject) u8 g_main_executable_storage[sizeof(Elf::DynamicObject)];
```
That gets the DynamicObject for the main executable placement new'd into it. It'd be nice to avoid the heap for the loader until we load libc.so, but that might not be possible.

Demos/DynamicLoader/Loader/main.cpp
```diff
@@ -0,0 +1,84 @@
...
+#include <sys/types.h>
+#include <unistd.h>
+
+#include <AK/LogStream.h>
+#include <AK/String.h>
+#include <LibELF/DynamicLoader.h>
+
+volatile int g_x = 0;
+
+int main(int argc, char** argv, char** envp)
```
ADKaster, line 52: I think we're still gonna want the fancy assembly entry point that hands off the location of _DYNAMIC to the loader entry point, along with original argv/argc etc.

Demos/DynamicLoader/Loader/CMakeLists.txt
```diff
@@ -0,0 +1,22 @@
...
+
+#set(CMAKE_CXX_FLAGS "-fPIE ${CMAKE_CXX_FLAGS}")
+#set_target_properties(Loader PROPERTIES OUTPUT_NAME Loader)
+
+#set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -nostdlib -fpic -pie")
+#set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS}")
+#set(LINKER_FLAGS "${LINKER_FLAGS} -")
+
+
+add_executable(libc.so.1 ${SOURCES})
```
ADKaster, line 13: Sergey had suggested "/bin/loader" as the name, but i think it should still go in /lib. "/lib/loader" ?

itamar8910: yep, /lib/loader would by my pick as well. I mean, we wouldn't want shell autocomplete to suggest running `loader` right? :laughing:

Demos/DynamicLoader/DynExec/main.cpp
```diff
@@ -0,0 +1,40 @@
...
+int libfunc();
+
+extern "C" void __libc_init();
+
+int main(int argc, char** argv)
+{
+    (void)argc;
+    (void)argv;
+    libc_testfunc();
+    // __libc_init();
```
ADKaster, line 37: Instead of trying to use libc right out of the gate, it might make sense to see if we can get just an PIE running via the  loader that doesn't have any library dependencies first. 

Maybe a hacky manual dbgputstr "Hello World"? Without worrying about DT_NEEDED until this part works.

Something like
```c++
#include <Kernel/syscall.h>
const char* g_string = "Hello, World!\n";

void local_dbgputstr(const char* str, int len)
{
    constexpr unsigned int function = Kernel::SC_dbgputc;
    for (int i = 0; i < len; ++i) {
        unsigned int result;
        asm volatile("int $0x82"
                 : "=a"(result)
                 : "a"(function), "d"((u32)str[i])
                 : "memory");
    }
}

int main (int argc, char** argv)
{
    local_debugputstr(g_string, 15);
    return 0;
}
```

Kernel/Process.cpp
```diff
@@ -1008,6 +925,18 @@ int Process::do_exec(NonnullRefPtr<FileDescription> main_program_description, Ve
         }
     }
 
+    if (interpreter_description) {
+        main_program_description->set_readable(true); // So that the dynamic loader can mmap the main program
+        auto main_program_fd = alloc_fd();
+        // FIXME: The proper way to do this is to pass a special auxilary vector to
+        // the dynamic loader. However, since the dynamic loader currently uses
+        // the normal libc _start, we can just pass data via the environ for now.
```
ADKaster, line 933: Can we load both of them into memory at once? Main exe and loader? I think the BSDs and Linux pass that kind of info via the Auxiliary vector, but there's no reason we couldn't get hacky and pass a pointer to where the kernel loaded the text segment by pushing it on the stack, for example

Demos/DynamicLoader/Loader/CMakeLists.txt
bugaevc: > Sergey had suggested "/bin/loader" as the name, but i think it should still go in /lib.

But why? It's a binary isn't it?

Demos/DynamicLoader/Loader/main.cpp
```diff
@@ -0,0 +1,84 @@
...
+
+    dbg() << "Loader main";
+    printf("Dynamic loader started\n");
+    char* main_program_path_ptr = getenv("_MAIN_PROGRAM_PATH");
+    ASSERT(main_program_path_ptr);
+    String main_program_path(main_program_path_ptr);
+    char* main_program_fd_str = getenv("_MAIN_PROGRAM_FD");
+    ASSERT(main_program_fd_str);
+    int main_program_fd = atoi(main_program_fd_str);
+    lseek(main_program_fd, 0, SEEK_SET);
```
bugaevc, line 65: Is this how it works, the loader mmaps the file in itself?

At least on Linux, I believe, both the loader and the executable itself are mmaped in by the kernel.

We do want the loader to be callable directly, as, say, `/path/to/loader /bin/foo bar baz`, so it should keep this manual mmaping functionality, and detect if it's being called by the kernel, as the interpreter, or as the executable.

ADKaster: I think the kernel should map both and pass an address to the loader... somehow of where it loaded the main exe (like i said in my comment above :) ). Either ELF Auxiliary vector, or just push it on the stack. Not sure what the kernel will currently try to do when you have a PT_INTERP, been a while since i wrote that part

Demos/DynamicLoader/Loader/CMakeLists.txt
ADKaster: It's not a standard program you'll run to do some work, it's the system program interpreter. /lib or /usr/lib is also the standard location for a program loader for every SYSV following OS. The spec actually says that if you want to conform to the i386 Processor supplement, it *has* to be named /usr/lib/libc.so.1. Which, that's silly.

bugaevc: Ah, that makes sense. But I don't think I agree. You're thinking of how in (at least) modern GNU/Linux, there's a distinction between how `/bin` is for tools you'd want to have in your `PATH`, and `/usr/lib` is for various system craft, including for binaries that you wouldn't run directly, but something else may need.

But we haven't been doing that in Serenity. Is, say, AudioServer a standard program you run to do some work? No, but we still put it in `/bin`.

ADKaster: systemd isn't something i'd want to run manually to do some work, but it still lives in /bin on my ubuntu VM. System Software (e.g. FooServer, System Monitor, etc) for sure go in bin. Seems to me the dynamic loader is a special piece of system software that isn't a normally executed binary.

bugaevc: > but it still lives in /bin on my ubuntu VM

Ubuntu must be patching it then; because [its proper location](https://github.com/systemd/systemd/blob/6bce17455e880404984761bfe064ed8ee1dce303/meson.build#L1726) is [`/usr/lib/systemd/`](https://github.com/systemd/systemd/blob/6bce17455e880404984761bfe064ed8ee1dce303/meson.build#L117)...

Dexesttp: > But we haven't been doing that in Serenity. Is, say, AudioServer a standard program you run to do some work? No, but we still put it in /bin.

Shouldn't it be moved, then ? It _does_ clutter the list of useful programs when autocompleting, even if it's not actually an issue yet.
Maybe there should be another folder for "service" applications which shouldn't appear in $PATH (either `/usr/lib` or a serenity-specific folder).

bugaevc: I would say either we do that (create a separate location for "service" binaries) — then it would make sense to also put the dynamic loader there.

Or, and I think I prefer that, keep putting all binaries in `/bin`, libraries in `/usr/lib`, and so on.

Speaking of autocompletion, it so happens that all of our non-command-line programs start with a capital letter (examples: AudioServer, Clipboard), so they shouldn't actually clutter your autocomplete unless you intentionally type a capital letter. So how about we name the loader `/bin/Loader`?

itamar8910:
So, the current state of the loader (Which I ended putting in `/usr/lib/Loader.so`, btw) is that most simple Userland apps run OK with the dynamic loader and a shared LibC. However, when I try to convert more complex libraries such as LibGUI to a shared version I'm getting weird TLS-related bugs.
This was not unexpected, since the current TLS in the dynamic loader is very hacky and ad-hoc. Turns out having the dynamic loader itself link to a static version of LibC complicates things when it comes to TLS, because we now have two separate TLS pools - one for the dynamic loader itself (which is statically loaded), and another one for the main program and its libraries.
So I think a redesign of the loader is required, and probably one where the dynamic loader is treated as just another shared object - It actually should simplify things.
I'm closing this draft for now, this attempt has gotten pretty far but I think it's a good place to take a step back and redo some things.

PR #2465, opened 2020-05-31
[WIP, RFC] unify all of our extensions to the escape sequences into one, structured sequence

alimpfard:
This commit also adds helper functions for constructing some terminal escape sequences, and replaces all previous instances of \x1b]9;value;max\x1b\\ with the new escape system.

@awesomekling, @bugaevc pls throw ideas at me, or tell me if this idea is completely insane.

The escape sequence is such: 
`OSC { S JSON-object OSC`

where the json object encodes commands as `command:{...parameters}`.
This is of course, a very rough draft, and any perf loss from this should be addressed if we decide to go with it.

as an aside, I also want some feedback on the helpers, if they look OK, I'll yank them out of this PR and finish their implementation.

### DIFF
--- a/Libraries/LibVT/Escape.h
+++ b/Libraries/LibVT/Escape.h
@@ -1,0 +1,134 @@
+/*
+ * Copyright (c) 2020, The SerenityOS developers.
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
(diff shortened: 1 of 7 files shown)


bugaevc:
No, hyperlink and title are intentionally compatible to what other terminals support

alimpfard:
The hyperlink and title helpers generate xterm sequences.
The only one implemented in that style is progress

Libraries/LibVT/Escape.h
```diff
@@ -0,0 +1,135 @@
...
+
+namespace VT {
+
+enum Property {
+    Progress,
+    Hyperlink,
+    Title,
+};
+
+struct escape_sequence_base {
```
awesomekling, line 40: `EscapeSequenceBase`

Libraries/LibVT/Escape.h
```diff
@@ -0,0 +1,135 @@
...
+    escape_sequence_base(String data)
+        : cached(data)
+    {
+    }
+
+    String cached;
+};
+
+template<int PropertyName>
+struct escape_sequence_for : public escape_sequence_base {
```
awesomekling, line 85: `EscapeSequenceFor`

Libraries/LibVT/Escape.h
```diff
@@ -0,0 +1,135 @@
...
+};
+
+struct UnsetTag {
+};
+
+constexpr static UnsetTag Unset;
```
awesomekling, line 91: `unset`

alimpfard:
As compared to hijacking the xterm-specific sequences, this new sequence has _some_ disadvantages, but it should be noted that any program that generates these specific sequences is only expected to run in serenity, or have that functionality inside serenity.

One immediate disadvantage I can think of is that other terminals will not hide the unknown data anymore.

awesomekling:
> One immediate disadvantage I can think of is that other terminals will not hide the unknown data anymore.

It would be nice if we didn't have that problem :(

Libraries/LibVT/Escape.h
alimpfard: nooooo my pretty constexpr tag names :(

VT::Unset looks so much better though...

Libraries/LibVT/Escape.h
alimpfard: Honestly, I was thinking of this as a transparent thing you treat like a function, but okay.

alimpfard:
Shit clicked the wrong button.

alimpfard:
> It would be nice if we didn't have that problem :(

I mean...we could maybe check if the terminal supports them and not emit them.
But I'm doing this in the name of not adding yet another nonstandard xterm sequence, and for future possibilities, so I'm very much against hiding it via a misbehaviour of another terminal (being that they ignore bad xterm sequences and omit them)

Libraries/LibVT/Escape.h
alimpfard: P.S. we already have Chomp and not chomp, what gives?

awesomekling: `Chomp` is an enum value :)

alimpfard: That's all?
I can just `enum UnsetTag { Unset }` and reap the benefits of nice constexpr tags yay

alimpfard:
How did I manage to do _that_?

alimpfard:
Update to this: We're not interested in not polluting the escape sequence namespace (me sad), so this PR goes away.
The helpers live though!

PR #2470, opened 2020-06-01
LibLine+Shell: Handle interrupts again

alimpfard:
This commit makes LibLine handle interrupts (as reported via
interrupted() and resized()) again.
There is a little catch with the shell:
```
$     ls |
pipe> <C-c> (prompt stays here until a key is pressed)
```

### DIFF
--- a/Shell/main.cpp
+++ b/Shell/main.cpp
@@ -133,7 +133,7 @@ int main(int argc, char** argv)
     }
 
     editor->on_interrupt_handled = [&] {
-        if (!shell->should_read_more()) {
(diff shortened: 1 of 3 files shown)


Shell/main.cpp
```diff
@@ -133,7 +133,7 @@ int main(int argc, char** argv)
     }
 
     editor->on_interrupt_handled = [&] {
-        if (!shell->should_read_more()) {
```
alimpfard, line 136: I can never decide which one is right lol.
This one is the correct one :tm:

awesomekling:
:smile:

PR #2474, opened 2020-06-02
LibJS: Add Object.{isExtensible,preventExtensions}()

mattco98:
Adds the concept of extensibility to Objects. Defining/putting new properties into a non-extensible object will fail, and will throw a TypeError in strict mode.

### DIFF
--- a/Libraries/LibJS/Runtime/ObjectConstructor.cpp
+++ b/Libraries/LibJS/Runtime/ObjectConstructor.cpp
@@ -48,6 +48,8 @@ ObjectConstructor::ObjectConstructor()
     define_native_function("getOwnPropertyNames", get_own_property_names, 1, attr);
     define_native_function("getPrototypeOf", get_prototype_of, 1, attr);
     define_native_function("setPrototypeOf", set_prototype_of, 2, attr);
+    define_native_function("isExtensible", is_extensible, 1, attr);
+    define_native_function("preventExtensions", prevent_extensions, 1, attr);
     define_native_function("keys", keys, 1, attr);
     define_native_function("values", values, 1, attr);
     define_native_function("entries", entries, 1, attr);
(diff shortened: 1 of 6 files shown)


Libraries/LibJS/Runtime/ObjectConstructor.cpp
```diff
@@ -104,6 +105,19 @@ Value ObjectConstructor::set_prototype_of(Interpreter& interpreter)
...
 }
 
+
+Value ObjectConstructor::prevent_extensions(Interpreter& interpreter)
+{
+    auto argument = interpreter.argument(0);
+    if (!argument.is_object())
+        return argument;
+    if (!argument.as_object().prevent_extensions()) {
+        interpreter.throw_exception<TypeError>("Proxy preventExtensions handler returned false");
```
mattco98, line 115: Just a note here:

This might look like a really strange error message, but from what I can tell by looking through the spec, the _only_ way a `prevent_extensions()` call can return false is if a user defines a Proxy trap that returns false. Plus I'm working on Proxies right now :) so it'll be less weird in a couple days.

mattco98, line 115: Just a note here:

This might look like a really strange error message, but from what I can tell by looking through the spec, the only way a prevent_extensions() call can return false is if a user defines a Proxy trap that returns false. Plus I'm working on Proxies right now :) so it'll be less weird in a couple days.

mattco98:
Just a note here:

https://github.com/SerenityOS/serenity/blob/65a3de5aa14aaa83c453d1d55d5cf00bd8321e4e/Libraries/LibJS/Runtime/ObjectConstructor.cpp#L115

This might look like a really strange error message, but from what I can tell by looking through the spec, the only way a prevent_extensions() call can return false is if a user defines a Proxy trap that returns false. Plus I'm working on Proxies right now :) so it'll be less weird in a couple days.

PR #2475, opened 2020-06-02
APIC: Detect processors and boot them into protected mode

tomuta:
This isn't fully working, though not stopping the OS from booting (the APs are in protected mode, but halted for now)

    For some reason we have what might be a paging issue. Accessing
    the stack or any memory after lgdt/lidt triggers a page fault, which
    it for some reason can't handle, so it ends up triple-faulting here.
    Interestingly, the CR2 register shows a fault address that is equal
    to &s_idt + 0x40, which happens to be IDT entry #8 (double fault):

    GDT=     c01c9240 000007ff
    IDT=     c01c9a40 000007ff
    CR0=e0000011 CR2=c01c9a80 CR3=00109000 CR4=00000060

    Same GDT/IDT and CR3 as the BSP. Also, both GDT and IDT look
    fine in both qemu and bochs debugger. qemu's `monitor info mem`
    also shows correct mappings (they also look fine in bochs):

    0000000000008000-0000000000009000 0000000000001000 -rw
    0000000000100000-0000000000200000 0000000000100000 -rw
    00000000c0000000-00000000c0118000 0000000000118000 -rw
    00000000c0118000-00000000c01c7000 00000000000af000 -r-
    00000000c01c7000-00000000c0800000 0000000000639000 -rw
    00000000c0801000-00000000c0802000 0000000000001000 -rw
    00000000c0803000-00000000c0804000 0000000000001000 -r-
    00000000c0805000-00000000c0815000 0000000000010000 -rw
    00000000ffe04000-00000000ffe05000 0000000000001000 -rw
    00000000ffe08000-00000000ffe09000 0000000000001000 -rw

    Qemu's `monitor info tlb` also has entries, but they look
    a little suspicious:

    00000000c07fe000: 00000000007fe000 --------W
    00000000c07ff000: 00000000007ff000 --------W
    00000000c0801000: 80000000fee00000 X--DA---W
    00000000c0803000: 800000000ffe1000 X---A----
    00000000c0805000: 8000000000802000 X-------W
    00000000c0806000: 8000000000803000 X-------W
    00000000c0807000: 8000000000804000 X-------W
    00000000c0808000: 8000000000805000 X-------W
    00000000c0809000: 8000000000806000 X-------W
    00000000c080a000: 8000000000807000 X-------W
    00000000c080b000: 8000000000808000 X-------W
    00000000c080c000: 8000000000809000 X-------W
    00000000c080d000: 800000000080a000 X-------W
    00000000c080e000: 800000000080b000 X-------W
    00000000c080f000: 800000000080c000 X-------W
    00000000c0810000: 800000000080d000 X-------W
    00000000c0811000: 800000000080e000 X-------W
    00000000c0812000: 800000000080f000 X-------W
    00000000c0813000: 8000000000810000 X-------W
    00000000c0814000: 8000000000811000 X-------W
    00000000ffe04000: 000000000010b000 ---DA---W
    00000000ffe08000: 0000000000801000 ---DA---W

    Notice the phyiscal address bit 63 being set. Bad tables?

https://github.com/SerenityOS/serenity/commit/9e5351ad3a410a55cb32ffb775911a01cb354fc2#diff-b10e477bf060d4b90924a2559b177384R320-R373

### DIFF
--- a/Kernel/Arch/i386/Boot/boot.S
+++ b/Kernel/Arch/i386/Boot/boot.S
@@ -218,7 +218,6 @@ start:
     pushl $exit_message
     call kprintf
     add $4, %esp
-
     cli
 
 loop:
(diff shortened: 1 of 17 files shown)


Kernel/Arch/i386/Boot/boot.S
```diff
@@ -227,3 +226,227 @@ loop:
...
+    00000000c080f000: 800000000080c000 X-------W
+    00000000c0810000: 800000000080d000 X-------W
+    00000000c0811000: 800000000080e000 X-------W
+    00000000c0812000: 800000000080f000 X-------W
+    00000000c0813000: 8000000000810000 X-------W
+    00000000c0814000: 8000000000811000 X-------W
+    00000000ffe04000: 000000000010b000 ---DA---W
+    00000000ffe08000: 0000000000801000 ---DA---W
+
+    Notice the phyiscal address bit 63 being set. Bad tables?
```
awesomekling, line 371: That's the NX bit :)

awesomekling:
Can we avoid identity-mapping the bottom 2MB? The first 2MB at the 3GB mark are already identity mapped to the bottom 2MB physical.

8infy:
Wait did you mean to say APIC instead of ACPI?

tomuta:
> Can we avoid identity-mapping the bottom 2MB? The first 2MB at the 3GB mark are already virtually mapped to the bottom 2MB physical.

I'm not sure this is possible. I tried all kinds of things, but I think this may be a chicken-egg problem. You can't use a temporary GDT in the 0x8000 range and then use the BSP's CR3 to enable paging, because as soon as you do, you lose the ability to execute the code in the 0x8000 range so you triple fault because you can't execute the jump instruction to jump above 3GB. But you also can't load the BSP's GDT (> 3gb) without enabling paging first.

While it's possible to switch into protected mode without identity mapping, it seems to be impossible to enable paging and switch to 3GB land without identity mapping for at least a brief period of time. Setting up separate paging on the APs to accomplish that, I'm not sure who this would work out when caches are shared...

It seems the only way is to briefly identity-map 0x8000, which allows the APs to switch to protected mode while still being able to address it, then jump above 3GB. Meanwhile the BSP is waiting until all APs are done with that, and then it can immediately undo the identity mapping and continue booting.

Maybe for the same reason you're doing the same thing here: https://github.com/SerenityOS/serenity/blob/1c6e4e04a8c02b27d24d65a99df3faa83762d5b9/Kernel/Arch/i386/Boot/boot.S#L198

Or am I missing something?

tomuta:
Basically it seems like a GDT pointer has to be accessible at any given time, so if either enabling paging renders it inaccessible, or if paging doesn't map a GDT pointer you're trying to set, then you end up with a triple fault. At least that seems to be my experience trying to make it work without identity mapping.

I still don't understand why I can't access the stack after having switched. All registers, TLB, memory maps, gdt, idt, they all look fine in both bochs and qemu.

awesomekling:
Hm, I guess I would have expected the CPU to keep fetching instructions at least until you cross a page boundary. But maybe it doesn't :)

If this identity mapping is only required during system boot, we should find a way to implement it that doesn't add permanent branches to the VM code. (Like the two separate range allocators for example.)

Oh, and what's the error code of the page fault you're getting when accessing the stack?

tomuta:
I think there are some major flaws, maybe segment related when trying to switch to protected mode. I just realized that while qemu boots ok, bochs actually doesn't like it at all. I guess more digging.

tomuta:
How can i verify if there is a Present bit not set? It seems like I can jump into the identity mapped code at 0xc0008000 but accessing the stack (or anything else) above 0xc0800000 always attempts to invoke the page fault handler, which itself is above 0xc0800000, so it ends up double faulting. With the changes I have locally I always have the pointer to the double fault idt entry in CR2, which is also above 0xc0800000. Accessing the same pointer on the BSP works just fine, but on the AP it always invokes the page fault handler. The identity-mapped area at 0xc0008000 works fine, though. So, I wonder if somewhere in the page hierarchy for >= 0xc0800000 maybe a present bit isn't set?

awesomekling:
`ERROR_RESERVED | ERROR_PROTECTION` means that one of Intel's reserved bits in a PDE/PTE is 1 instead of 0.

tomuta:
I think the problem is in boot.S. It looks like it just copies the physical address as-is into the PDE, but looking at the spec it should be broken up into bits (M–20):13 and 31:22, and there are reserved bits in between that must be zero.

tomuta:
Hm, is `MemoryManager::ensure_pte` missing a call to `pde.set_huge(true)` to set bit 7 of the PDE, which would cause it to point to a PTE rather than a 2MB page directly? Looks like the latter does have a reserved range that we could accidentally write into. Though it seems weird that it doesn't trigger a page fault on the BSP, either (unless it is silently handling it somehow).

awesomekling:
> Hm, is `MemoryManager::ensure_pte` missing a call to `pde.set_huge(true)` to set bit 7 of the PDE, which would cause it to point to a PTE rather than a 2MB page directly? Looks like the latter does have a reserved range that we could accidentally write into. Though it seems weird that it doesn't trigger a page fault on the BSP, either (unless it is silently handling it somehow).

We're not using huge pages for anything other than 3GB...(3GB+8MB), the rest is 4KB.

tomuta:
I think I just figured it out... It's the NX bit. The problem was NX wasn't enabled on the AP...

tomuta:
Alright, so I finally got it working. It's booting all APs into C++ land where they notify the BSP that they are done initializing. And because MM and scheduler can't handle it yet, I'm just putting them to sleep indefinitely.

I don't think it's possible to do this without identity mapping. Nothing I tried worked, and all other implementations appear to be relying on it as well.

Kernel/Interrupts/APIC.h
```diff
@@ -27,16 +27,72 @@
...
+            Fixed = 0x0,
+            LowPriority = 0x1,
+            SMI = 0x2,
+            NMI = 0x4,
+            INIT = 0x5,
+            StartUp = 0x6,
+        };
+        enum DestinationMode {
+            Physical = 0x0,
+            Logical = 0x0,
```
bugaevc, line 61: Typo?

Kernel/Interrupts/APIC.cpp
```diff
@@ -223,9 +243,58 @@ void enable(u32 cpu)
 
             IO::delay(200);
         }
+
+        // Now wait until the ap_cpu_init_pending variable dropped to 0, which means all APs are initialized and no longer need these special mappings
+        if (m_apic_ap_count.load(AK::MemoryOrder::memory_order_consume) != aps_to_enable) {
+            klog() << "APIC: Waiting for " << aps_to_enable << " AP(s) to finish initialization...";
+            do {
+                // Wait a little bit
+                IO::delay(200);
```
bugaevc, line 252: :(

Kernel/Interrupts/APIC.h
```diff
@@ -27,16 +27,72 @@
...
+            : m_reg(vector | (delivery_mode << 8) | (destination_mode << 11) | (level << 14) | (static_cast<u32>(trigger_mode) << 15) | (destination << 18))
+        {
+        }
+    
+        u32 low() const { return m_reg; }
+        u32 high() const { return 0; }
+    };
+
+    OwnPtr<Region> m_apic_base;
+    AK::Vector<OwnPtr<Region>> m_apic_ap_stacks;
```
bugaevc, line 88:
```suggestion
    NonnullOwnPtrVector<Region> m_apic_ap_stacks;
```

Kernel/Interrupts/APIC.h
tomuta: Good catch, indeed

Kernel/Interrupts/APIC.cpp
tomuta: ?

Kernel/Interrupts/APIC.h
tomuta: Hm I can't figure out how to get it to compile with `NonnullOwnPtrVector`. The append method doesn't like to take a `OwnPtr<Region>` and I haven't been able to figure out how to solve it. Any hints? I looked around and at the implementations but haven't found any clues yet.

awesomekling: If you have an `OwnPtr` that you trust to be non-null, you can convert it into a `NonnullOwnPtr` with `OwnPtr::release_nonnull()`. :)

Kernel/Interrupts/APIC.cpp
bugaevc: Could we maybe not delay boot-up while the application CPUs are starting?

Perhaps this could be done in a separate thread or something.

awesomekling: 200 microseconds is not too bad. Doesn't seem worth complicating this by bringing in the scheduler while booting AP's.

bugaevc: Notice this is in a loop :(

awesomekling:
Okay let's merge this and continue iterating on top.

I dislike that we're adding a separate identity range allocator, but we can work towards removing that again later as this code matures. :)

tomuta:
> Okay let's merge this and continue iterating on top.
> 
> I dislike that we're adding a separate identity range allocator, but we can work towards removing that again later as this code matures. :)

I agree, don't like that change either but I'm still learning the code ;)

PR #2477, opened 2020-06-02
Userland/js: Add `global` property to the global object + help fix

nooga:
This adds a `global` property to the global object so that `global == this` when typed into the REPL. 

This matches what Node.js does and helps with testing on existing JS code that uses `global` to detect its environment.

### DIFF
--- a/Userland/js.cpp
+++ b/Userland/js.cpp
@@ -359,6 +359,7 @@ ReplObject::ReplObject()
 void ReplObject::initialize()
 {
     GlobalObject::initialize();
+    define_property("global", this, JS::Attribute::Enumerable);
     define_native_function("exit", exit_interpreter);
     define_native_function("help", repl_help);
     define_native_function("load", load_file, 1);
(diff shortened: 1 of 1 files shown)


linusg:
But... we already have `globalThis`, which does exactly the same and was created to solve exactly that problem: different names for the global `this` in different environments...

nooga:
I'm aware of `globalThis` but since I started working with libraries like lodash and others it seems like they don't take it into account. 

In Node.js this holds: `global === globalThis && global === this`,
In browsers, this holds: `window === globalThis &&  window === this`.

I wanted to match Node behavior here.

linusg:
Right, we're still building upon a burning pile of legacy code. :shrug:

awesomekling:
Please wrap commit messages at 72 characters per the [contributing guidelines](https://github.com/SerenityOS/serenity/blob/master/CONTRIBUTING.md)

nooga:
@awesomekling 
```
$ echo "Userland/js: Add 'global' property to the global object + help fix" | wc -c
      67
```

awesomekling:
The other two lines are well over 100 characters long

nooga:
@awesomekling sorry 😅
It didn't occur to me that this also applies to the non-title lines.

awesomekling:
:+1:

PR #2481, opened 2020-06-02
LibJS+js: Exception traces

linusg:
image


Libraries/LibJS/Runtime/Exception.h
```diff
@@ -37,12 +37,14 @@ class Exception : public Cell {
     virtual ~Exception() override;
 
     Value value() const { return m_value; }
+    Vector<String> trace() const { return m_trace; }
```
awesomekling, line 40: Let's return this as a `const Vector<String>&` to avoid copying the `Vector` every time.

awesomekling:
Kickass, this is awesome :)

PR #2483, opened 2020-06-02
Browser: Add network history window

Lubrsi:
This tool provides a table of network requests and a few extra tools.
It can be opened via the inspect menu or Ctrl+E.

The information it shows is the base name, path, host and protocol
from the URL. It also shows how long each request took in ms.

There are currently three extra tools provided:
- History across multiple pages or a single page
- Cache bypass (while the window is open)
- Open requested resource in a new tab (in context menu)

Demo: 
History Demo

Note: Path can be very long, so it's hidden by default.

The main issue I had with this is passing data to the model. It can surely be cleaned up :^)

### DIFF
--- a/Applications/Browser/NetworkHistoryWidget.cpp
+++ b/Applications/Browser/NetworkHistoryWidget.cpp
@@ -1,0 +1,154 @@
+/*
+ * Copyright (c) 2020, Luke Wilde <luke.wilde@live.co.uk>
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include "NetworkHistoryWidget.h"
+#include <LibGUI/Action.h>
+#include <LibGUI/BoxLayout.h>
+#include <LibGUI/CheckBox.h>
+#include <LibGUI/Menu.h>
+#include <LibGUI/Splitter.h>
+#include <LibGUI/TableView.h>
+#include <LibGUI/ToolBar.h>
+#include <LibGUI/ToolBarContainer.h>
+#include <LibWeb/Loader/ResourceLoader.h>
+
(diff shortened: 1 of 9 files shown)


linusg:
Please configure your editor to add a newline at the end of files :)

Applications/Browser/NetworkHistoryWidget.cpp
```diff
@@ -0,0 +1,169 @@
...
+
+    m_context_menu = GUI::Menu::construct();
+
+    m_context_menu->add_action(GUI::Action::create("Open in new tab", [this](auto&) {
+        if (on_tab_open_request)
+            on_tab_open_request(m_context_menu_url);
+    }));
+
+    m_history_view->on_context_menu_request = [this](const auto& index, const auto& event) {
+        if (index.row() == -1 || index.column() == -1)
```
bugaevc, line 85: Please use `index.is_valid()`

Applications/Browser/NetworkHistoryWidget.cpp
```diff
@@ -0,0 +1,169 @@
...
+        Web::NetworkHistoryModel::Entry new_entry;
+        new_entry.url = url;
+
+        m_network_history.set(load_id, new_entry);
+
+        update_view();
+
+        Core::ElapsedTimer load_timer;
+        load_timer.start();
+        m_active_timers.set(load_id, load_timer);
```
bugaevc, line 120: Why isn't the timer a part of the entry?

Applications/Browser/NetworkHistoryWidget.cpp
```diff
@@ -0,0 +1,169 @@
...
+}
+
+NetworkHistoryWidget::~NetworkHistoryWidget()
+{
+    unregister_callbacks();
+}
+
+void NetworkHistoryWidget::update_view()
+{
+    m_history_view->set_model(Web::NetworkHistoryModel::create(m_network_history));
```
bugaevc, line 100: Don't do that, just tell the model to update

Applications/Browser/NetworkHistoryWidget.h
```diff
@@ -0,0 +1,67 @@
...
+    bool m_clear_on_navigation { true };
+    bool m_paused { false };
+    bool m_cache_disabled { false };
+
+    HashMap<u32, Web::NetworkHistoryModel::Entry> m_network_history;
+    HashMap<u32, Core::ElapsedTimer> m_active_timers;
+
+    RefPtr<GUI::TableView> m_history_view;
+    RefPtr<GUI::Menu> m_context_menu;
+    String m_context_menu_url;
```
bugaevc, line 64: Shouldn't it be, indeed, a URL?

Libraries/LibWeb/NetworkHistoryModel.cpp
```diff
@@ -0,0 +1,138 @@
...
+    case Column::Time:
+        return "Time";
+    default:
+        ASSERT_NOT_REACHED();
+    }
+}
+
+GUI::Variant NetworkHistoryModel::data(const GUI::ModelIndex& index, Role role) const
+{
+    auto& entry = m_entries.at(index.row());
```
bugaevc, line 80: I'd assert that the index is valid first

Libraries/LibWeb/NetworkHistoryModel.cpp
```diff
@@ -0,0 +1,138 @@
...
+
+    if (role == Role::Display) {
+        if (index.column() == Column::Name) {
```
bugaevc, line 83: This chain of `if`s should clearly be a `switch`

Libraries/LibWeb/NetworkHistoryModel.cpp
```diff
@@ -0,0 +1,138 @@
...
+        if (index.column() == Column::Path) {
+            if (entry.url.protocol() == "data")
+                return "N/A";
+
+            return entry.url.path();
+        }
+
+        if (index.column() == Column::Host) {
+            if (entry.url.protocol() == "data" || entry.url.protocol() == "file")
+                return "N/A";
```
bugaevc, line 99: No, we can have a host for file URLs!

Applications/Browser/NetworkHistoryWidget.cpp
Lubrsi: I originally thought that once I have the elapsed time, I no longer need the timer. But after thinking about it again, yeah, it's probably not worth a whole other hash map :^)

Lubrsi:
Comments addressed 👍

Applications/Browser/NetworkHistoryWidget.cpp
```diff
@@ -0,0 +1,154 @@
...
+            return;
+
+        Web::NetworkHistoryModel::Entry new_entry;
+        new_entry.url = url;
+        new_entry.load_timer.start();
+
+        m_network_history.set(load_id, new_entry);
+
+        // This will add some microseconds to the load timer, which isn't that big of a deal.
+        update_view();
```
bugaevc, line 118: Guess you could schedule the update for later, instead of doing it inside the callback, if that's a concern

Lubrsi: Eh, it's only really noticable with cached resources, where it switches between 0 and 1 ms. So currently it's not much of a concern.

awesomekling:
Hello Lubrsi and thank you for working on this!

Unfortunately I feel that it's much too soon to start adding visualizations of loader state. Work has only just begun on the loader so let's not multiply the maintenance burden by adding GUI just yet.

The general idea here is very nice and this is something we'll come back to a bit later for sure. :)

PR #2484, opened 2020-06-02
LibJS: Make typeof return undefined for undefined variables

nooga:
This makes `typeof i_dont_exist` return `undefined` instead of 
throwing an error.


Libraries/LibJS/AST.cpp
```diff
@@ -585,6 +585,26 @@ Value UnaryExpression::execute(Interpreter& interpreter) const
...
+            return {};
+        }
+        // FIXME: standard recommends checking with is_unresolvable but it ALWAYS return false here
+        if (reference.is_local_variable() || reference.is_global_variable()) {
+            auto name = reference.name();
+            Value value;
+            if (reference.is_local_variable())
+                value = interpreter.get_variable(name.to_string());
+            else
+                value = interpreter.global_object().get(name);
```
linusg, line 600: Don't think this is necessary as `get_variable()` also handles global variables.

```suggestion
            auto value = interpreter.get_variable(name.to_string());
```

nooga: > now typeof foo will output ...

You mean the double `get`? Other JS engines print just one `get` but the result is still `"undefined"`.

About the `get_variable`... I basically ripped this part from:
https://github.com/SerenityOS/serenity/blob/8ff4587f653bdebebacd1d311a3263875c660dc7/Libraries/LibJS/Runtime/Reference.cpp#L87-L90

I don't have any other idea on how to check if the variable is defined without throwing. 
My previous attempt was trying to use `Reference::is_unresolvable()`:
https://github.com/SerenityOS/serenity/blob/8ff4587f653bdebebacd1d311a3263875c660dc7/Libraries/LibJS/Runtime/Reference.h#L67

The problem is that for variables the `Reference::m_base` gets initialized with `js_null()` making the truly undefined vars slip through `Reference::is_unresolvable()`. Attempting to initialize it differently wreaks total havoc.

nooga: What we are trying to do here is basically:
> if the argument is a variable and it will throw ReferenceErrror on access return "undefined"

linusg: > You mean the double get?

Yes, we shouldn't call the getter twice :)

Should be as easy as putting `m_lhs->execute(interpreter);` in an `else` code branch, and instead of returning the `"undefined"` string in the `if` added in this PR, set `lhs_result` to `js_undefined()` and let the function figure it out itself later on.

> The problem is that for variables the Reference::m_base gets initialized with js_null() making the truly undefined vars slip through Reference::is_unresolvable(). Attempting to initialize it differently wreaks total havoc.

I'm not very familiar with all the `Reference` code, you seem to already have dived deeper than I have...

nooga: > Should be as easy as putting m_lhs->execute(interpreter);
I think this will throw a `ReferenceError` and unwind the stack.

linusg: What I mean is something like this:


```diff
diff --git a/Libraries/LibJS/AST.cpp b/Libraries/LibJS/AST.cpp
index 833536e46..d1b1dcd5e 100644
--- a/Libraries/LibJS/AST.cpp
+++ b/Libraries/LibJS/AST.cpp
@@ -585,9 +585,22 @@ Value UnaryExpression::execute(Interpreter& interpreter) const
         return base_object->delete_property(reference.name());
     }
 
-    auto lhs_result = m_lhs->execute(interpreter);
-    if (interpreter.exception())
-        return {};
+    Value lhs_result;
+    if (m_op == UnaryOp::Typeof && m_lhs->is_identifier()) {
+        auto reference = m_lhs->to_reference(interpreter);
+        if (interpreter.exception())
+            return {};
+        // FIXME: standard recommends checking with is_unresolvable but it ALWAYS return false here
+        if (reference.is_local_variable() || reference.is_global_variable()) {
+            auto name = reference.name();
+            lhs_result = interpreter.get_variable(name.to_string()).value_or(js_undefined());
+        }
+    } else {
+        lhs_result = m_lhs->execute(interpreter);
+        if (interpreter.exception())
+            return {};
+    }
+
     switch (m_op) {
     case UnaryOp::BitwiseNot:
         return bitwise_not(interpreter, lhs_result);
```

Libraries/LibJS/AST.cpp
```diff
@@ -585,9 +585,23 @@ Value UnaryExpression::execute(Interpreter& interpreter) const
...
+    Value lhs_result;
+    if (m_op == UnaryOp::Typeof && m_lhs->is_identifier()) {
+        auto reference = m_lhs->to_reference(interpreter);
+        if (interpreter.exception()) {
+            return {};
+        }
+        // FIXME: standard recommends checking with is_unresolvable but it ALWAYS return false here
+        if (reference.is_local_variable() || reference.is_global_variable()) {
+            auto name = reference.name();
+            lhs_result = interpreter.get_variable(name.to_string()).value_or(js_undefined());
```
awesomekling, line 597: `get_variable` can do a get on the global object and so it needs an exception check

nooga: added one :)

PR #2487, opened 2020-06-03
LibJS: Allow null or undefined as a bound |this| value in strict mode

jack-karamanian:
This PR allows `null` or `undefined` to be a bound `this` value in functions produced from `Function.prototype.bind` when in strict mode.


awesomekling:
Nice :)

PR #2489, opened 2020-06-03
LibJS: Add Proxy objects

mattco98:
This PR adds Proxy objects. There are multiple things of note:

Firstly, any method that accepted or returned `u8 attributes` has been changed to use the new `PropertyAttribute` class. In addition to have many nice utility methods for dealing with attributes (no more bitwise operators!), it also distinguishes between an attribute being defined and it being false. For example, the following two property descriptors are now different when passed to `Object.defineProperty`: `{}`, `{ configurable: false }`. This is important because some of the `Proxy` logic has to distinguish between those two states. And let me put your mind at ease in advance: no extra memory is being consumed by this new class! It is simply a wrapper around a `u8`, and since we now have 8 `Attribute` flags, this works perfectly.

And, of course, this adds `Proxy` objects. It doesn't add everything, so here's a TODO list for the proxies:

1. Add traps for `[[Call]]`, `[[Construct]]`, and `[[OwnPropertyKeys]]`. These were not added in this PR because they'd take some reworking (and this PR is already _way_ too big).
2. `Proxy.revocable(target, handler)`. I attempted to add this, but struggled to keep a reference to the created proxy in the created `revoke` function. 
3. Protect all `virtual` methods in the `Object` class with respect to exceptions. With this change, _all_ `virtual` methods in the `Object` class can throw exceptions, as the `Object` could potentially be a `ProxyObject`. These calls will therefore have to be followed by an exception check in a future PR.

~I'm marking this PR as a draft because currently the tests only test the invariants and the objects passed to the proxy traps; they don't test the actual functionality of well-behaving proxies. I will add them later, but they're not important for a code review.~ Added!

### DIFF
--- a/Libraries/LibJS/Runtime/Object.cpp
+++ b/Libraries/LibJS/Runtime/Object.cpp
@@ -123,7 +123,7 @@ Value Object::get_own_property(const Object& this_object, PropertyName property_
     return value_here;
 }
 
-Value Object::get_own_properties(const Object& this_object, GetOwnPropertyMode kind, u8 attributes) const
+Value Object::get_own_properties(const Object& this_object, GetOwnPropertyMode kind, PropertyAttributes attributes) const
 {
     auto* properties_array = Array::create(interpreter().global_object());
 
@@ -168,7 +168,7 @@ Value Object::get_own_properties(const Object& this_object, GetOwnPropertyMode k
     }
 
     for (auto& it : this_object.shape().property_table_ordered()) {
-        if (it.value.attributes & attributes) {
+        if (it.value.attributes.bits() & attributes.bits()) {
             size_t offset = it.value.offset + property_index;
 
             if (kind == GetOwnPropertyMode::Key) {
@@ -194,7 +194,7 @@ Value Object::get_own_properties(const Object& this_object, GetOwnPropertyMode k
 Value Object::get_own_property_descriptor(PropertyName property_name) const
 {
     Value value;
-    u8 attributes;
+    PropertyAttributes attributes;
 
     if (property_name.is_number()) {
         auto existing_value = m_indexed_properties.get(nullptr, property_name.as_number(), false);
@@ -214,12 +214,12 @@ Value Object::get_own_property_descriptor(PropertyName property_name) const
     }
 
     auto* descriptor = Object::create_empty(interpreter(), interpreter().global_object());
-    descriptor->define_property("enumerable", Value((attributes & Attribute::Enumerable) != 0));
-    descriptor->define_property("configurable", Value((attributes & Attribute::Configurable) != 0));
+    descriptor->define_property("enumerable", Value(attributes.is_enumerable()));
+    descriptor->define_property("configurable", Value(attributes.is_configurable()));
     if (value.is_object() && value.as_object().is_native_property()) {
         auto result = call_native_property_getter(const_cast<Object*>(this), value);
         descriptor->define_property("value", result);
-        descriptor->define_property("writable", Value((attributes & Attribute::Writable) != 0));
+        descriptor->define_property("writable", Value(attributes.is_writable()));
     } else if (value.is_accessor()) {
         auto& pair = value.as_accessor();
         if (pair.getter())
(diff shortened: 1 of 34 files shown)


Libraries/LibJS/Runtime/Object.cpp
```diff
@@ -123,7 +123,7 @@ Value Object::get_own_property(const Object& this_object, PropertyName property_
     return value_here;
 }
 
-Value Object::get_own_properties(const Object& this_object, GetOwnPropertyMode kind, u8 attributes) const
+Value Object::get_own_properties(const Object& this_object, GetOwnPropertyMode kind, PropertyAttributes attributes) const
```
awesomekling, line 166: Now that we're looking at this, this API seems a bit goofy.

What does it mean to `get_own_properties` with a set of `attributes` anyway?

This is really just a flag that filters based on the `enumerable` attribute AFAIK so maybe it can be a `bool` instead?

Libraries/LibJS/Runtime/Object.h
```diff
@@ -39,6 +39,19 @@
 
 namespace JS {
 
+struct PropertyDescriptor {
+    PropertyAttributes attributes;
+    Value value;
+    Function* getter;
+    Function* setter;
```
awesomekling, line 46: Let's be good bois and initialize these explicitly:
```suggestion
    Function* getter { nullptr };
    Function* setter { nullptr };
```

Libraries/LibJS/Runtime/Object.cpp
```diff
@@ -40,6 +40,46 @@
 
 namespace JS {
 
+PropertyDescriptor PropertyDescriptor::from_object(Interpreter& interpreter, const Object& object)
```
awesomekling, line 43: Perhaps we could call this `from_dictionary`? `from_object` sounds like it's doing something slightly different :)

Libraries/LibJS/Runtime/ProxyObject.cpp
```diff
@@ -0,0 +1,465 @@
...
+
+#include <LibJS/Interpreter.h>
+#include <LibJS/Runtime/Accessor.h>
+#include <LibJS/Runtime/Error.h>
+#include <LibJS/Runtime/GlobalObject.h>
+#include <LibJS/Runtime/ProxyObject.h>
+
+namespace JS {
+
+bool static is_compatible_property_descriptor(Interpreter& interpreter, bool is_extensible, PropertyDescriptor Desc, Optional<PropertyDescriptor> current_opt)
```
awesomekling, line 35: Please use full words for variable names. Something like:
* `Desc` -> `new_descriptor`
* `current_opt` -> `current_descriptor`

Libraries/LibJS/Runtime/ProxyObject.h
```diff
@@ -0,0 +1,67 @@
...
+
+#pragma once
+
+#include <LibJS/Runtime/Object.h>
+
+namespace JS {
+
+class ProxyObject : public Object {
+public:
+    static ProxyObject* create(GlobalObject&, Object&, Object&);
```
awesomekling, line 35: Please don't omit parameter names unless it's obvious from type or context what is what.
In this case, two `Object&` in a row is completely ambiguous :)

Libraries/LibJS/Runtime/ProxyObject.h
```diff
@@ -0,0 +1,67 @@
...
+
+    ProxyObject(Object& target, Object& handler, Object& prototype);
+    virtual ~ProxyObject() override;
+
+    Object& target() const { return m_target; }
+    Object& handler() const { return m_handler; }
```
awesomekling, line 41: Do we really need to vend mutable references from a const `ProxyObject`?
Or can these be split into two each (one const and one non-const)?

Libraries/LibJS/Runtime/ProxyObject.cpp
mattco98: Oops, I had it in spec-jargon to make it easy to implement and forgot to change it back

mattco98:
>Hmm, are you sure we need to distinguish present vs omitted property attributes at the object property storage level? 

Well, for every property (shape or indexed), this needs to be tracked. I'm not sure where else it'd even be an option to put something like this, because I really didn't think too much about it. I just noticed that it'd be exactly 8 attributes, which fits nicely, so I just did it :) I'm totally open for suggestions though. 

>Is this behavior exposed somehow?

Sorry, but I'm not sure what you mean here. Do you mean exposed to the JS runtime? If so, then no; there's no way to tell through `Object.getOwnPropertyDescriptor` whether a property was explicitly set to `false` or just omitted. But `Proxy` methods unfortunately do need this information.

awesomekling:
This seems okay to me then :)

PR #2490, opened 2020-06-03
LibJS: Fix Parser.parse_template_literal looping forever

mattco98:
`parse_template_literal` now breaks out of the loop if it sees something it doesn't expect. Additionally, it now checks for EOFs.

### DIFF
--- a/Libraries/LibJS/Parser.cpp
+++ b/Libraries/LibJS/Parser.cpp
@@ -719,7 +719,7 @@ NonnullRefPtr<TemplateLiteral> Parser::parse_template_literal(bool is_tagged)
     if (!match(TokenType::TemplateLiteralString))
         append_empty_string();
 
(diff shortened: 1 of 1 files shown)


Libraries/LibJS/Parser.cpp
```diff
@@ -741,6 +741,13 @@ NonnullRefPtr<TemplateLiteral> Parser::parse_template_literal(bool is_tagged)
 
             if (!match(TokenType::TemplateLiteralString))
                 append_empty_string();
+        } else {
+            syntax_error(
+                "Unexpected token",
+                m_parser_state.m_current_token.line_number(),
+                m_parser_state.m_current_token.line_column()
+            );
```
sunverwerth, line 749:
```suggestion
            expected("Template literal string");
```

PR #2495, opened 2020-06-04
LibCrypto: Correctly pad blocks with FinalBlockSize < size < BlockSize 

### LINKED ISSUE #2488: The mysterious TLS bug when loading twitter.com
After much trial and error, here's an MWE for triggering a BadRecordMac:
```html
<html>
<body>
<script src="https://abs.twimg.com/responsive-web/web/main.d7deeee4.js"></script>
</body>
</html>
```

It should be noted that loading that URL via `pro` does *not* trigger the bug, and the download is successful if attempted (via `pro`).

However, trying to load that file through the address bar (but _not_ downloading it) also triggers this bug.

alimpfard:
This fixes #2488

I am not entirely sure if we can load twitter, but at least we don't crash while doing so.
One out of bounds write is also addressed now (we were writing out of bounds in the hash function when `FinalBlockSize < size < BlockSize`).

Oh, and MORE ASSERTIONS :tada:

### DIFF
--- a/Libraries/LibCrypto/Hash/SHA1.cpp
+++ b/Libraries/LibCrypto/Hash/SHA1.cpp
@@ -115,17 +115,26 @@ SHA1::DigestType SHA1::peek()
     __builtin_memcpy(data, m_data_buffer, m_data_length);
     __builtin_memcpy(state, m_state, 20);
 
+    if (BlockSize == m_data_length) {
+        transform(m_data_buffer);
+        m_bit_length += BlockSize * 8;
+        m_data_length = 0;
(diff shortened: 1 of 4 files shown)


awesomekling:
That's some heavy code repetition in those SHAxxx peek functions..

awesomekling:
Thanks for tracking this down btw :D

alimpfard:
> That's some heavy code repetition in those SHAxxx peek functions

That's why I wanted the template, but the compiler no likey
maybe I should just use a base class for them...there are a lot of subtle differences though

PR #2497, opened 2020-06-04
LibJS: Hoist function declarations

nooga:
This patch adds function declaration hoisting. The mechanism
is similar to var hoisting. Hoisted function declarations are to be put
before the hoisted var declarations, hence they have to be treated
separately.

### DIFF
--- a/Libraries/LibJS/AST.cpp
+++ b/Libraries/LibJS/AST.cpp
@@ -67,10 +67,8 @@ Value ScopeNode::execute(Interpreter& interpreter) const
     return interpreter.run(*this);
 }
 
-Value FunctionDeclaration::execute(Interpreter& interpreter) const
+Value FunctionDeclaration::execute(Interpreter&) const
 {
-    auto* function = ScriptFunction::create(interpreter.global_object(), name(), body(), parameters(), function_length(), interpreter.current_environment());
-    interpreter.set_variable(name(), function);
     return js_undefined();
 }
 
@@ -1763,6 +1761,11 @@ Value DebuggerStatement::execute(Interpreter&) const
 void ScopeNode::add_variables(NonnullRefPtrVector<VariableDeclaration> variables)
 {
     m_variables.append(move(variables));
+}
+
+void ScopeNode::add_functions(NonnullRefPtrVector<FunctionDeclaration> functions)
+{
+    m_functions.append(move(functions));
 }
 
 }
--- a/Libraries/LibJS/AST.h
+++ b/Libraries/LibJS/AST.h
@@ -40,6 +40,7 @@
 namespace JS {
 
 class VariableDeclaration;
+class FunctionDeclaration;
 
 template<class T, class... Args>
 static inline NonnullRefPtr<T>
@@ -124,8 +125,9 @@ public:
     virtual void dump(int indent) const override;
 
     void add_variables(NonnullRefPtrVector<VariableDeclaration>);
+    void add_functions(NonnullRefPtrVector<FunctionDeclaration>);
     const NonnullRefPtrVector<VariableDeclaration>& variables() const { return m_variables; }
-
+    const NonnullRefPtrVector<FunctionDeclaration>& functions() const { return m_functions; }
     bool in_strict_mode() const { return m_strict_mode; }
     void set_strict_mode() { m_strict_mode = true; }
 
@@ -136,6 +138,7 @@ private:
     virtual bool is_scope_node() const final { return true; }
     NonnullRefPtrVector<Statement> m_children;
     NonnullRefPtrVector<VariableDeclaration> m_variables;
+    NonnullRefPtrVector<FunctionDeclaration> m_functions;
     bool m_strict_mode { false };
 };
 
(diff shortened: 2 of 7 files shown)


awesomekling:
What happens when there's a var and a function declaration with the same name?

nooga:
> What happens when there's a var and a function declaration with the same name?

Excellent question - The function wins. I'm not sure why though...

```
> var x = 1; function x() { return 1 }
undefined
> x
[ScriptFunction]
```

```
>  { var f = f(); function f() { return 1; } console.log(f) }
[object ScriptFunction]
undefined
```

Which is not what should happen apparently... in node:
```
> var x = 1; function x() { return 1 }
undefined
> x
1
```

```
> { var f = f(); function f() { return 1; } console.log(f) }
({ var f = f(); function f() { return 1; } console.log(f) })
       ^

Uncaught SyntaxError: Unexpected identifier
```

linusg:
> Which is not what should happen apparently. In Node:

That's because node interprets `{ ... }` on its own as an object literal, even though that's technically not valid. Stupid decision, IMO.

(also see the added parens in the error output)

nooga:
@linusg yes, but change the first `f` into something else so it doesn't collide and it's suddenly this:
```
> { var x = f(); function f() { return 1; } console.log(f) }
[Function: f]
undefined
```

nooga:
So I have investigated the behavior of major JS engines regarding cases when function declaration clashes with different kinds of variable declarations (var/let/const) within different scopes and... the results are wildly different from each other. 

Almost as if the standard didn't give any guidance here and each engine took its own route. 

We can either pick one of the implementations and copy its behavior or come up with our own. 

At the moment, my implementation leaves the value set to whatever declaration/assignment came last in given scope and the functions are scoped like vars, i.e. not taking { } scopes into account. 

Please let me know what do you think is best approach here.

linusg:
> So I have investigated the behavior of major JS engines regarding cases when function declaration clashes with different kinds of variable declarations (var/let/const) within different scopes and... the results are wildly different from each other.

Do you have any examples? I do get consistent behaviour, except for the error messages in some cases - but that shouldn't matter if both engines agree it's a syntax error.

(SpiderMonkey via `js`/`jsshell`, V8 via `node`)

---

```js
var foo;
function foo() {}
console.log(typeof foo);
```

SpiderMonkey: `function`
V8: `function`

---

```js
var foo = 1;
function foo() {}
console.log(typeof foo);
```

SpiderMonkey: `number`
V8: `number`

---

```js
let foo;
function foo() {}
console.log(typeof foo);
```

(same for `let foo = 1` and `const foo = 1`)

SpiderMonkey:

```
test.js:2:9 SyntaxError: redeclaration of let foo:
test.js:2:9 function foo() {}
test.js:2:9 .........^
test.js:1:4 note: Previously declared at line 1, column 4
```

V8:

```
/home/linus/test.js:2
function foo() {}
^

SyntaxError: Identifier 'foo' has already been declared
```

---

```js
function foo() {}
var foo;
console.log(typeof foo)
```

SpiderMonkey: `function`
V8: `function`

---

```js
function foo() {}
var foo = 1;
console.log(typeof foo)
```

SpiderMonkey: `number`
V8: `number

---

````js
function foo() {}
let foo;
console.log(typeof foo);
```

(same for `let foo = 1` and `const foo = 1`)

SpiderMonkey:

```
test.js:2:4 SyntaxError: redeclaration of function foo:
test.js:2:4 let foo;
test.js:2:4 ....^
test.js:1:9 note: Previously declared at line 1, column 9
```

V8:

```
/home/linus/test.js:2
let foo;
    ^

SyntaxError: Identifier 'foo' has already been declared
```

---

```js
if (true) {
    var foo;
    function foo() {}
    console.log(typeof foo)
}
```

SpiderMonkey:

```
test.js:3:13 SyntaxError: redeclaration of var foo:
test.js:3:13     function foo() {}
test.js:3:13 .............^
test.js:2:8 note: Previously declared at line 2, column 8
```

V8:

```
/home/linus/test.js:2
    var foo;
        ^

SyntaxError: Identifier 'foo' has already been declared
```

---

```js
if (true) {
    let foo;
    function foo() {}
    console.log(typeof foo)
}
```

SpiderMonkey:

```
test.js:3:13 SyntaxError: redeclaration of let foo:
test.js:3:13     function foo() {}
test.js:3:13 .............^
test.js:2:8 note: Previously declared at line 2, column 8
```

V8:

```
/home/linus/test.js:3
    function foo() {}
    ^

SyntaxError: Identifier 'foo' has already been declared
```

---

```js
if (true) {
    function foo() {}
    var foo;
    console.log(typeof foo)
}
```

SpiderMonkey:

```
test.js:3:8 SyntaxError: redeclaration of function foo
test.js:2:13 note: Previously declared at line 2, column 13
```

V8:

```
/home/linus/test.js:3
    var foo;
        ^

SyntaxError: Identifier 'foo' has already been declared
```

---

```js
if (true) {
    function foo() {}
    let foo;
    console.log(typeof foo)
}
```

(same for `let foo = 1` and `const foo = 1`)

SpiderMonkey:

```
test.js:3:8 SyntaxError: redeclaration of function foo:
test.js:3:8     let foo;
test.js:3:8 ........^
test.js:2:13 note: Previously declared at line 2, column 13
```

V8:

```
/home/linus/test.js:2
    var foo;
        ^

SyntaxError: Identifier 'foo' has already been declared
```

awesomekling:
Ok let's roll with it. If we find compat issues we can adjust accordingly. :)

PR #2503, opened 2020-06-04
find(1)

bugaevc:
This is a `find` :^)

It's not a complete implementation of what POSIX specifies, but more commands can always be implemented later on; all the infrastructure is there.

cc @alimpfard code review please :smile:
cc @danboid you requested this

### DIFF
--- a/Userland/find.cpp
+++ b/Userland/find.cpp
@@ -1,0 +1,466 @@
+/*
+ * Copyright (c) 2020, Sergey Bugaev <bugaevc@serenityos.org>
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include <AK/NonnullOwnPtr.h>
+#include <AK/OwnPtr.h>
+#include <AK/Vector.h>
+#include <LibCore/DirIterator.h>
+#include <getopt.h>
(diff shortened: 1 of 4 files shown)


Userland/find.cpp
```diff
@@ -0,0 +1,468 @@
...
+        } else {
+            int status;
+            int rc = waitpid(pid, &status, 0);
+            if (rc < 0) {
+                perror("waitpid");
+                g_there_was_an_error = true;
+                return false;
+            }
+            return WIFEXITED(status) && WEXITSTATUS(status) == 0;
+        }
```
alimpfard, line 270: What if it was signaled? Are we gonna leave it to run in the background?

Userland/find.cpp
```diff
@@ -0,0 +1,468 @@
...
+OwnPtr<Command> parse_complex_command(char* argv[])
+{
+    auto command = parse_simple_command(argv);
+
+    while (command && argv[optind] && argv[optind + 1]) {
+        StringView arg = argv[++optind];
+
+        enum { And,
+            Or } binary_operation
+            = And;
```
alimpfard, line 364: What's up with the odd formatting here?

bugaevc: clang-format ¯\_(ツ)_/¯

Manually changed to a single line for now.

Userland/find.cpp
bugaevc: wdym, if it was killed by a signal, we'll reap it (and evaluate to false), otherwise we'll keep waiting for it.

Userland/find.cpp
alimpfard: I think you can add a trailing comma to get clang-format to add an extra newline after the last element

bugaevc: I don't like what it does with the first line either. I like either

```cpp
enum { And, Or } binary_operation = And;
```

or

```cpp
enum {
    And,
    Or,
} binary_operation = And;
```

alimpfard: Adding a comma and bringing the And down manually will make clang-format produce the second one (I like that too)

Userland/find.cpp
alimpfard: Ah, nvm, I got it mixed up.

Userland/find.cpp
bugaevc: It produces

```cpp
enum {
    And,
    Or,
} binary_operation
    = And;
```

over here. (clang-format version 10.0.0)

Userland/find.cpp
```diff
@@ -0,0 +1,466 @@
...
+        if (binary_operation == And)
+            command = make<AndCommand>(command.release_nonnull(), rhs.release_nonnull());
+        else
+            command = make<OrCommand>(command.release_nonnull(), rhs.release_nonnull());
+    }
+
+    return command;
+}
+
+NonnullOwnPtr<Command> parse_all_commands(char* argv[])
```
alimpfard, line 392: `parse_command_list`?

alimpfard: Actually, that's not what it does either, there's no "commands" at all.
How about just `parse_commands`?

bugaevc: Kind of, but actually no, It's not really a list, it's more like an expression.

`-foo -bar -baz` is actually `-foo -a -bar -a -baz`, which parses to single top-level command, `AndCommand(AndCommand(FooCommand, BarCommand), BazCommand)`.

The point of this function is to call `parse_complex_command()` and then add `-print` if no action commands are found; perhaps the name could reflect that, if you have any ideas.

Userland/find.cpp
```diff
@@ -0,0 +1,466 @@
...
+    }
+}
+
+void walk_tree(const char* root_path, Command& command)
+{
+    command.evaluate(root_path);
+
+    Core::DirIterator dir_iterator(root_path, Core::DirIterator::SkipParentAndBaseDir);
+    if (dir_iterator.has_error() && dir_iterator.error() == ENOTDIR)
+        return;
```
alimpfard, line 449: Maybe say something?
I think this is about `find some-file -whatever`?

my host's `find` runs the commands on the single entry.

Userland/find.cpp
alimpfard: Ah, initialise with braces :)

Userland/find.cpp
bugaevc: It's not just about the top-level file; any time that we descend into the tree and find a file, we don't descend further (because you cannot descend into a file).

alimpfard: Ah, the command is evaluated above.
Gotcha.

Userland/find.cpp
bugaevc: But I don't *want* to initialize with braces here, it's just a simple assignment.

alimpfard: I guess you'll have to live with `// clang-format {off,on}` then

that idea of "braces for nontrivials" is a bit...backwards...
C++ would refuse to do some casts with braces that it would allow in `= ...`

awesomekling:
Hey this is a pretty nice implementation of `find`! Way to make it C++y :)
I don't understand why you love writing `else` so much, but hey everyone's different.

PR #2506, opened 2020-06-04
WindowServer: Make perror() strings slightly more detailed.

nico:
My system has an umask of 027. This means that every file that's created gets the w bit erased for group members, and all of rwx erased for others. It's likely some security measure.

This has the effect that the files in Base/ don't have rx set for others, but those bits are needed when serenity runs so that e.g. the window user can read /etc/WindowServer/WindowServer.ini.
Instead of just copying the on-disk permissions into the image, explicitly set w for group and others,
and set x for group and others for files that are x for user (ie executables and directories -- that's
what X does).

(I added this locally why debugging my umask issue and it seems like it's maybe generally
useful.)

### DIFF
--- a/Services/WindowServer/main.cpp
+++ b/Services/WindowServer/main.cpp
@@ -45,22 +45,22 @@ int main(int, char**)
     }
 
     if (unveil("/res", "r") < 0) {
-        perror("unveil");
+        perror("unveil /res");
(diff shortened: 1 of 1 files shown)


awesomekling:
Sure why not :)

PR #2509, opened 2020-06-05
LibJS: Add BigInt

linusg:
- [x] Tokenisation, syntax highlighting
- [x] Parsing
- [x] AST
- [x] bigint primitives
- [x] `BigInt()`
- [ ] ~`BigInt.asIntN()`~ (weird, not doing that now)
- [ ] ~`BigInt.asUintN()`~ (weird, not doing that now)
- [x] `BigInt.prototype`
- Operators:
  - [x] unary `typeof`
  - [x] unary `-`
  - [x] arithmetic
  - [x] bitwise (partially)
  - [x] comparison
  - [x] logical (got those for free)
- [x] Tests

cc @alimpfard


linusg:
> There already is a separate UnsignedBigInteger(u32) and SignedBigInteger(i32) constructor, but this is fine too.

Oops - not sure how I missed that. Much better :)

alimpfard:
what does this `asUintN` even do? sounds like a weird way of specifying the width of a bigint, which is dumb

alimpfard:
> not sure how I missed that.

blame it on the rest of LibCrypto, those hash functions always get in the way.

alimpfard:
[The spec](https://tc39.es/ecma262/#sec-bigint-objects) doesn't seem to say anything, but are these expected to be in 2's complement?
if so, bitwise ops with negatives will be wonky, since we're doing sign magnitude...

alimpfard:
Hmmm...my benchmark (:tm:) assumed that `bigint++` made sense (I see node is fine with that), but the spec strictly prohibits this (`++` is defined in terms of ToNumber, and we follow that part of the spec.

```
let a = BigInt(1);
a++;
```
every js impl I have access to seems to be okay with doing this.

what's firefox even doing?
shot-2020-06-06_06-13-57

alimpfard:
well...expected benchmark completion time is 5 hours and a half (C++: 28 seconds)
I _think_ there's the smallest room for improvement here lol

linusg:
> but the spec strictly prohibits this (++ is defined in terms of ToNumber, and we follow that part of the spec.)

I think it's perfectly fine: https://tc39.es/ecma262/#sec-postfix-increment-operator

It's not using `ToNumber` but `toNumeric`, which either returns a number or bigint (or throws). So... I'll implement it :)

linusg:
Ok, I think I'm mostly done here - still missing bitwise ops (need some additional APIs on `Crypto::SignedBigInteger` for that, but on the JS side of things everything is prepared).

cc @nooga, have a look at the `abstract_relation` changes if you want (list in `binary-relational.js` is now even larger than before :smile:)

Libraries/LibJS/Runtime/Value.cpp
```diff
@@ -28,10 +29,19 @@
 #include <AK/String.h>
 #include <AK/StringBuilder.h>
 #include <AK/Utf8View.h>
+#include <LibCrypto/BigInt/SignedBigInteger.h>
+// FIXME: This SHOULD NOT be required to build with Lagom
+// Compiler is complaining about: random_big_prime, LCM, GCD, ModularInverse
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wunused-function"
+#include <LibCrypto/NumberTheory/ModularFunctions.h>
+#pragma GCC diagnostic pop
```
linusg, line 38: If someone can help me solving this in a nicer way that'd be great :D

alimpfard:
> need some additional APIs on Crypto::SignedBigInteger for that

take a look at your fork's PRs :^)

Libraries/LibJS/Runtime/Value.cpp
alimpfard: hmmm
could you try making them non-static?
if that fails, try marking them `inline` explicitly?

alimpfard:
> It's not using ToNumber but toNumeric

Oh it is, I guess I was just blind or something lol

Libraries/LibJS/Runtime/Value.cpp
```diff
@@ -305,55 +395,71 @@ Value less_than_equals(Interpreter& interpreter, Value lhs, Value rhs)
...
+    auto lhs_numeric = lhs.to_numeric(interpreter);
     if (interpreter.exception())
         return {};
-    return Value(~(i32)lhs_number.as_double());
+    if (lhs_numeric.is_number())
+        return Value(~(i32)lhs_numeric.as_double());
+    TODO();
+    // FIXME: this gives weird results:
+    // ~1 === -2
+    // ~1n === -4294967294n // ???
```
alimpfard, line 461: This is a little mistake on my part, not'ing the bigint probably shouldn't not the most significant word, i.e.
(assuming 16 bit words) `0x0001 | 0x0001` should become `0x0000 | 0x1110` (the leading zeros shouldn't be negated)

but even then, we're doing sign-magnitude and even though the spec doesn't specify, everyone expects 2's complement.
while that's rather simple to do for simple arithmetic, I don't want to deal with its edge-cases in division.
This will fix it to say `~1n == -0n`, and `~~1n == 1n`, which is expected of SM.

```diff
diff --git a/Libraries/LibCrypto/BigInt/UnsignedBigInteger.cpp b/Libraries/LibCrypto/BigInt/UnsignedBigInteger.cpp
index 607efdd8a..bfc141f38 100644
--- a/Libraries/LibCrypto/BigInt/UnsignedBigInteger.cpp
+++ b/Libraries/LibCrypto/BigInt/UnsignedBigInteger.cpp
@@ -505,10 +505,22 @@ FLATTEN void UnsignedBigInteger::bitwise_not_without_allocation(
         output.invalidate();
         return;
     }
+    if (right.length() == 0) {
+        output.set_to_0();
+        return;
+    }
+
     output.m_words.resize_and_keep_capacity(right.length());

-    for (size_t i = 0; i < right.length(); ++i)
-        output.m_words[i] = ~right.words()[i];
+    if (right.length() > 1) {
+        for (size_t i = 0; i < right.length() - 1; ++i)
+            output.m_words[i] = ~right.words()[i];
+    }
+
+    auto last_word_index = right.length() - 1;
+    auto last_word = right.words()[last_word_index];
+
+    output.m_words[last_word_index] = ((u32)0xffffffffffffffff >> __builtin_clz(last_word)) & ~last_word;
 }

 /**
diff --git a/Libraries/LibJS/Runtime/Value.cpp b/Libraries/LibJS/Runtime/Value.cpp
index 6d1d520b9..8426eb6d7 100644
--- a/Libraries/LibJS/Runtime/Value.cpp
+++ b/Libraries/LibJS/Runtime/Value.cpp
@@ -455,11 +455,7 @@ Value bitwise_not(Interpreter& interpreter, Value lhs)
         return {};
     if (lhs_numeric.is_number())
         return Value(~(i32)lhs_numeric.as_double());
-    TODO();
-    // FIXME: this gives weird results:
-    // ~1 === -2
-    // ~1n === -4294967294n // ???
-    // return js_bigint(interpreter, lhs_numeric.as_bigint().big_integer().bitwise_not());
+    return js_bigint(interpreter, lhs_numeric.as_bigint().big_integer().bitwise_not());
 }

 Value unary_plus(Interpreter& interpreter, Value lhs)
```

linusg: but `~1n` should result in `-2n` it seems, i.e. same as `~1` results in `-2`?

alimpfard: well, assuming it works in 2's complement, yes
\<basic numeric system stuff\>
1 has the lowest bit set, negating that gives us a bunch of 1's and the lsb not set, which is -1 (all bit set) - 1 (sans the firs bit):
1: `1000`
-2: `0111`
\</basic numeric system stuff\>

That's actually a decently simple system, except that division would be rather difficult to implement in terms of unsigned numbers.
unless the spec says they _must_ be in this system, we can just ignore it until it breaks something...or just disallow bitwise operations on negatives.

linusg:
@awesomekling feel free to review before I continue making changes to this PR lol

awesomekling:
This looks pleasant!

Issue #2234, opened 2020-05-15
Build is broken on macOS


diversys:
macOS 10.14.6 

```
cd Toolchain
./BuildIt_x86_64.sh
cd ..
mkdir build
cmake ..
make -i
```
```
[  0%] Building CXX object Meta/Lagom/CMakeFiles/LagomCore.dir/__/__/AK/Base64.cpp.o
[  0%] Building CXX object Meta/Lagom/CMakeFiles/LagomCore.dir/__/__/AK/FileSystemPath.cpp.o
[  0%] Building CXX object Meta/Lagom/CMakeFiles/LagomCore.dir/__/__/AK/FlyString.cpp.o
^C
bash-5.0$ make -i -j8
[  0%] Building ASM object Kernel/CMakeFiles/boot.dir/Arch/i386/Boot/boot.S.o
/bin/sh: /Users/diver/serenity/Toolchain/Local/bin/i686-pc-serenity-gcc: No such file or directory
[  0%] Building CXX object Libraries/LibC/CMakeFiles/crt0.dir/crt0.cpp.o
[  0%] Building CXX object Kernel/Modules/CMakeFiles/TestModule.dir/TestModule.o
/bin/sh: /Users/diver/serenity/Toolchain/Local/bin/i686-pc-serenity-g++: No such file or directory
/bin/sh: /Users/diver/serenity/Toolchain/Local/bin/i686-pc-serenity-g++: No such file or directory
[  0%] Built target boot
[  0%] Building CXX object Meta/Lagom/CMakeFiles/LagomCore.dir/__/__/AK/FlyString.cpp.o
[  0%] Building CXX object Meta/Lagom/CMakeFiles/LagomCore.dir/__/__/AK/MappedFile.cpp.o
[  0%] Building CXX object Meta/Lagom/CMakeFiles/LagomCore.dir/__/__/AK/LogStream.cpp.o
[  0%] Building CXX object Meta/Lagom/CMakeFiles/LagomCore.dir/__/__/AK/JsonParser.cpp.o
[  0%] Linking CXX static library libcrt0.a
[  0%] Linking CXX static library libTestModule.a
[  0%] Building CXX object Meta/Lagom/CMakeFiles/LagomCore.dir/__/__/AK/JsonValue.cpp.o
Error running link command: No such file or directory
Error running link command: No such file or directory
install: illegal option -- D
usage: install [-bCcpSsv] [-B suffix] [-f flags] [-g group] [-m mode]
[...]

PR #2510, opened 2020-06-05
Docs: Update FreeBSD instructions

zaklaus:
ginstall should not be symlinked to install globally. The actual issue of having GNU coreutils prefixed with `g` and not being picked up by CMake can be fixed on CMake level as noted here: https://github.com/SerenityOS/serenity/issues/2234#issuecomment-629754194


awesomekling:
What's the fix at CMake level then? :)
cc @bugaevc

PR #2512, opened 2020-06-05
[WIP] LibGfx: Initial implementation of TrueType font rendering

### LINKED ISSUE #743: LibGUI: Implement TrueType and/or OpenType fonts
Currently Serenity only supports pixel based fonts, and as such are limited in size and cannot be scaled. (unless of course you did some type of pixel based scaling with anti-aliasing, but that would probably look terrible). On higher resolution screens with higher pixel densities the fonts also become unreadable, so implementing this would also allow for possible display scaling in the future.

The browser would also benefit greatly from this because it's currently limited to the sizes provided by the existing fonts.

SrimantaBarua:
Fix #743 .

This is currently incomplete. It does not support CFF fonts (that's probably best added in a later PR), and also does not support composite TTF outlines (will be adding that soon).

I posted a screenshot on the linked issue of the demo application and TrueType Label widget I added, but I haven't included them here yet, since they require a TTF font to be present on the system. Repeating my question from the issue - would there be any licensing issues with including a font file within the project?

**Note to reviewers**: I don't know if my use of ByteBuffer and different smart pointers (RefPtr, OwnPtr) is correct and/or idiomatic, so it'd be great if you could comment on that.

**Notes on the implementation**
I've tried to make this nearly zero-allocation and zero-copy beyond the buffer created to read the file, and the buffer allocated by `Rasterizer`. Structures from the font file aren't parsed on load. The `Font` object stores slices of the underlying buffer that correspond to the tables we are interested in. This means that startup should be fast, but rastering glyphs to bitmaps would be slightly slower. Which should be okay, since glyph bitmaps should be cached anyway.

A `Font` cannot be used directly, it must be scaled to get a `ScaledFont` wrapper. The API requires users to first get a `glyph_id` for a `codepoint`, and then get a bitmap for the `glyph_id`. This is one step more than the current `Font` API, but I designed this keeping in mind a future HarfBuzz-like shaper, which returns a sequence of glyph IDs after shaping a run of text.


alimpfard:
I don't normally do code reviews, but I've touched on this thing before (also, @Dexesttp not sure if you're interested in TTF, but hey).

I saw some extra copies of things, `min` and `max` are in AK, please use those.
we already have `quadratic_bezier_curve_to`/`draw_line` in LibGfx, please use that, or if you find it not good enough, modify that to fit your standards.

Also, (personal opinion, you don't _have_ to do this) please use enums instead of magic constants to make it easier to read for idiots like me :)

Dexesttp:
That's some really nice stuff :) I like how you kept it as low-allocation as possible, and still managed to make it fairly easy to follow. The point iterator, especially, is a neat idea !

I will try to do a full review of the TTF loading code to point out some specific things later (maybe tomorrow), but here are some first impressions :
- Like Ali said, there is a _lot_ of magic values and it makes the code hard to read in some cases for people who don't know how TTF works. Enums would solve that.
- The "everything in a single big file" strategy will be an issue as more parts of TTF are supported - especially when planning CCF support, which should be equivalent in size, but also composite glyphs, gsub / gpos tables, cvt programs and some parts of WOF2 decompressing would benefit from extra files.
I would recommend creating a TTF folder and putting code inside. Having the glyph & rasterizer-specific code on its own would be nice. Some tables with a lot of options that are partially implemented, such as cmap, could also already live in their own files.
- A lot of "ASSERT_NOT_REACHED()" should be "TODO()" macros instead. (granted, it's a recent addition to AK, so maybe you diddn't knwo about it ^^).
- Some parts seem duplicated specifically for TTF fonts - the parts Ali pointed out, but also the difference between AABitmap and Bitmap isn't very clear to me.
Is there a reason to not use bitmap ? AABitmap just looks like a lightweight bitmap (which 1) would limit the path towards supporting color glyphs and 2) looks like possibly premature optimization).

jsdevv:
Can't wait for TTF support on SerenityOS!

SrimantaBarua:
Thanks @alimpfard , @Dexesttp ! I haven't worked on this codebase before, just kludged things together from what I've seen on the YouTube videos. So thanks for pointing out the duplication! I'll address the other issues too.

Also, would you guys be able to comment on the licensing aspect of including a font file within the project source, in /res/fonts? If that's okay, I can add a TTF Label widget and demo to this PR too.

sunverwerth:
@SrimantaBarua There are open fonts that should be license compatible with SerenityOS. Question is if including some random font is compatible with the spirit of this project :D On the other hand, I don't think we have a font designer here who is willing to create a custom "Serenity Sans" for us

awesomekling:
Indeed @sunverwerth, we're not going to bundle 3rd party fonts with the project. It's something that will have to be included as an optional port. :)

Libraries/LibGfx/TTFont.cpp
```diff
@@ -105,6 +116,247 @@ u16 Font::Maxp::num_glyphs() const
     return be_u16(m_slice.offset_pointer(4));
 }
 
+struct Point {
```
awesomekling, line 141: Please use `Gfx::FloatPoint` and add functionality as needed :)

Libraries/LibGfx/TTFont.cpp
```diff
@@ -105,6 +116,247 @@ u16 Font::Maxp::num_glyphs() const
...
+    float m_x_scale;
+    float m_y_scale;
+};
+
+class Rasterizer {
+public:
+    Rasterizer(Size size)
+        : m_size(size)
+    {
+        m_data = OwnPtr(new float[m_size.width() * m_size.height()]);
```
awesomekling, line 257: Please use a `FixedArray`

Libraries/LibGfx/TTFont.h
```diff
@@ -27,17 +27,39 @@
...
 namespace Gfx {
 namespace TTF {
 
-class Font;
+class AABitmap {
+public:
+    AABitmap(Size size)
+        : m_size(size)
+    {
+        m_data = OwnPtr(new u8[size.width() * size.height()]);
```
awesomekling, line 44: Please use a `ByteBuffer` (or a `FixedArray<u8>`)

Libraries/LibGfx/Painter.cpp
```diff
@@ -960,6 +1016,132 @@ void Painter::draw_text_line(const Rect& a_rect, const Utf32View& text, const Fo
     }
 }
 
+void Painter::draw_text_line(const Rect& a_rect, const Utf8View& text, const TTF::ScaledFont& font, TextAlignment alignment, Color color, TextElision elision)
```
awesomekling, line 1019: The number of duplications of this function is getting silly now. We need to find a way to share all this code between the slightly different implementations :)

Libraries/LibGfx/Painter.cpp
```diff
@@ -1086,6 +1268,124 @@ void Painter::draw_text(const Rect& rect, const Utf32View& text, const Font& fon
     }
 }
 
+void Painter::draw_text(const Rect& rect, const StringView& raw_text, const TTF::ScaledFont& font, TextAlignment alignment, Color color, TextElision elision)
```
awesomekling, line 1271: Same duplication issue here. We need to find a way to share all of it between different font types.

Let's rearchitect the `Gfx::Font` class to be a wrapper around either a simple bitmap font (the system format) or a TrueType font. Then we can have a font-type-agnostic API in `Painter`.

Libraries/LibGfx/TTFont.h
```diff
@@ -232,6 +263,24 @@ class Font {
             static Glyph simple(const ByteBuffer& slice, u16 num_contours, i16 xmin, i16 ymin, i16 xmax, i16 ymax);
             static Glyph composite(const ByteBuffer& slice); // FIXME: This is currently just a dummy. Need to add support for composite glyphs.
             AABitmap raster(float x_scale, float y_scale) const;
+            int ascender() const
+            {
+                if (m_type == Type::Simple) {
+                    return m_meta.simple.ymax;
+                } else {
```
awesomekling, line 270: No need for `else` after a `return` statement. Same comment repeated throughout.

SrimantaBarua:
Rebased, and addressed some of the comments related to localized changes (e.g. API usage, coding style, magic constants). Will start working on the higher-level comments (reorganization into LibTTF, reducing duplication in Painter) now.

Regarding @Dexesttp 's comment on Bitmap vs AABitmap - I've switched to using Bitmap. But currently, this doesn't support color glyphs, so only the alpha values are relevant.

Dexesttp:
I'll wait so you can address the library issues before reviewing the code completely. But overall, apart from some styling things (such as member initializations not being done or using `=`), it looks fine and accurate so I'm not sure there's much to review :)

> Please put this in a LibTTF and let LibGfx depend on it.
@awesomekling This won't be possible for the TTF rasterizer code, so I wonder whether it's worth splitting TTF between a LibTTF and a LibGfx.

The rasterizer needs to create bitmap data, so it makes sense that it should be based on Gfx::Painter (for drawing bezier curves). This probably means that it should keep being in some kind of Gfx::TTFRasterizer class.
Right now, only  "simple" OpenType glyphs are implemented, so the rasterizer is short.
However, the complete TTF rasterizer is huge. It needs to handle different data inputs from all kind of TTF formats, and also support special behaviours for specific tables. As the support grows, the rasterizer will also support drawing CCF / Postscript glyphs, gsub/gpos glyph variations, and executing the CVT program for each glyph data. This won't be easily maintanable if it lives in a single file.

So, even by splitting the TTF stuff between Gfx (for the rasterizer) and LibTTF (for the ttf file parsing), there will still be a lot of TTF-specific code in LibGfx which depends on a specific behaviour of LibTTF parsing and data transfer... Which makes the split feel less useful and more of a burden as we try to find which parts of TTF font usage should go in which library.

awesomekling:
> > Please put this in a LibTTF and let LibGfx depend on it.
> 
> @awesomekling This won't be possible for the TTF rasterizer code, so I wonder whether it's worth splitting TTF between a LibTTF and a LibGfx.

Oops, sorry, I meant to say put this in a LibTTF and make it depend on LibGfx. LibGfx then gets some kind of abstract Font interface that can be implemented by LibTTF.

> The rasterizer needs to create bitmap data, so it makes sense that it should be based on Gfx::Painter (for drawing bezier curves). This probably means that it should keep being in some kind of Gfx::TTFRasterizer class.

I don't see why the rasterizer needs to be in LibGfx just because it uses Painter. LibGUI has `GUI::Painter` which is a subclass of `Gfx::Painter` for example.

> Right now, only "simple" OpenType glyphs are implemented, so the rasterizer is short.
> However, the complete TTF rasterizer is huge. It needs to handle different data inputs from all kind of TTF formats, and also support special behaviours for specific tables. As the support grows, the rasterizer will also support drawing CCF / Postscript glyphs, gsub/gpos glyph variations, and executing the CVT program for each glyph data. This won't be easily maintanable if it lives in a single file.

Why do we have to put the rasterizer in a single file?

> So, even by splitting the TTF stuff between Gfx (for the rasterizer) and LibTTF (for the ttf file parsing), there will still be a lot of TTF-specific code in LibGfx which depends on a specific behaviour of LibTTF parsing and data transfer... Which makes the split feel less useful and more of a burden as we try to find which parts of TTF font usage should go in which library.

Again, I'm not seeing why the rasterizer must be in LibGfx. It seems like poor layering with no clear benefit. What am I missing?

alimpfard:
From what I saw in my experimental run of doing TTF, the parser and all the data-handling stuff can be stashed away nicely.
We can _probably_ make Font abstract and use some font-agnostic API in LibGfx (not tested).

> Rasteriser iz huge

The rasteriser turns rather ugly with kerning (I suggest passing a Transform around), and gpos made me rage quit...I'd say that's fair
Factoring in all the vm stuff, I'd say that's a future issue though

alimpfard:
Oh and @Dexesttp, compound glyphs aren't that bad if you have the right representation of glyphs.
Actual complexity issues come up when you don't take them into account from the start

Dexesttp:
Well... Most of the argument isn't valid if you meant that LibTTF should depend on LibGfx instead of the other way around, so you can ignore it ^^

However, that means that the existing (and nice to use) Gfx::Painter::draw_text() API can't create TTF fonts anymore, and each program will need to manually implement drawing text using TTF fonts. That feels like a shame compared to just having TTF support "for free" for all programs.

Maybe we can move all the draw_text() stuff to GUI::Painter and only allow LibGfx to handle lines and bitmaps (i.e. move the font and emoji support from Gfx to GUI) ? This way, we can create a mechanism to use both TTF and standard fonts interchangeably without needing programs to support TTF text independently.

awesomekling:
> Well... Most of the argument isn't valid if you meant that LibTTF should depend on LibGfx instead of the other way around, so you can ignore it ^^

Okay, that's my bad then :)

> However, that means that the existing (and nice to use) Gfx::Painter::draw_text() API can't create TTF fonts anymore, and each program will need to manually implement drawing text using TTF fonts. That feels like a shame compared to just having TTF support "for free" for all programs.

That's what the abstract Font interface is for. `Painter::draw_text()` will take a `Font`, but `Font` would internally delegate to a LibTTF implementation or a simple bitmap font (current system font format) implementation.

> Maybe we can move all the draw_text() stuff to GUI::Painter and only allow LibGfx to handle lines and bitmaps (i.e. move the font and emoji support from Gfx to GUI) ? This way, we can create a mechanism to use both TTF and standard fonts interchangeably without needing programs to support TTF text independently.

Text rendering needs to live in LibGfx or we lose the ability to draw text in WindowServer (which does not use LibGUI since LibGUI is basically the WindowServer client API.)

SrimantaBarua:
One issue I see is that bitmap fonts return, well, bitmaps (by value), and the TTF rasterizer returns `Gfx::Bitmap` which is more like a u32map.

Dexesttp:
>That's what the abstract Font interface is for. Painter::draw_text() will take a Font, but Font would internally delegate to a LibTTF implementation or a simple bitmap font (current system font format) implementation.

OOh. Yeah, that's a good idea :)
That could be done in a separate PR as well. If, through this PR, we have a first version which allows drawing TTFFont text via a specific API (independently of Painter), we should then be able to refactor Gfx::Painter::draw_text() to use an abstract Font interface and switch between fonts quite easily.

SrimantaBarua:
I'm working on a PoC for an abstract `Font` class. I'll try pushing it with this PR if it doesn't touch too many things.

SrimantaBarua:
Nope. A lot of stuff would have to be changed. I think let's limit the scope of this PR to adding a LibTTF with a dependency on LibGfx, with functions to draw text independent of Painter, as @Dexesttp suggested. I'll switch on working on parsing composite glyphs.

danboid:
This will be a major boost for Serenity - great to see it being worked on!

Libraries/LibTTF/Font.cpp
```diff
@@ -0,0 +1,381 @@
...
+    }
+    if (tag == tag_from_str("OTTO")) {
+        dbg() << "CFF fonts not supported yet";
+        return nullptr;
+    }
+    if (tag != 0x00010000) {
+        dbg() << "Not a valid  font";
+        return nullptr;
+    }
+    return adopt(*new Font(move(buffer), 0));
```
Dexesttp, line 194: It's more general than just this line,  but...
```suggestion
    return Font::load_from_offset(move(buffer), 0);
```

The idea is to move the header parsing code out of the constructor. By using a `static RefPtr<Font> load_from_offset(buffer, offset)` method instead, you can parse the header and split the table.
Then, once everything is parsed properly, you can create a font via `adopt(*new Font(move(head_slice), move(hhea_slice), ...)`.

The main advantage is gracious failures. Right now, if you try loading an invalid font file, you would just `ASSERT()` that the required tables are missing, which is not ideal (especially since this could happen in some important programs such as WindowServer, which does font rendering...).

It would also help with future-proofing, especially the load mechanism for WOFF/WOF2 formats (which requires to decompress some tables before being able to read them).

Libraries/LibTTF/Font.h
```diff
@@ -0,0 +1,129 @@
...
+    ByteBuffer m_head_slice;
+    ByteBuffer m_hhea_slice;
+    ByteBuffer m_maxp_slice;
+    ByteBuffer m_hmtx_slice;
+    ByteBuffer m_loca_slice;
+    ByteBuffer m_glyf_slice;
+    // These are stateful wrappers around tables
+    Cmap m_cmap;
+
+    friend ScaledFont;
```
Dexesttp, line 100: Usually, this seems done at the top of the class (right around `AK_MAKE_NONCOPYABLE`).
See [here](https://github.com/SerenityOS/serenity/blob/master/Libraries/LibWeb/Parser/HTMLToken.h#L39) for an example.
(We should have guidelines for these)

Libraries/LibTTF/Font.h
```diff
@@ -0,0 +1,129 @@
...
+    RefPtr<Gfx::Bitmap> raster_glyph(u32 glyph_id) const { return m_font->raster_glyph(glyph_id, m_x_scale, m_y_scale); }
+    u32 glyph_count() const { return m_font->glyph_count(); }
+    int width(const StringView&) const;
+    int width(const Utf8View&) const;
+    int width(const Utf32View&) const;
+
+private:
+    RefPtr<Font> m_font;
+    float m_x_scale;
+    float m_y_scale;
```
Dexesttp, line 124:
```suggestion
    float m_x_scale { 0 };
    float m_y_scale { 0 };
```

Libraries/LibTTF/Glyf.h
```diff
@@ -0,0 +1,182 @@
...
+                glyph.raster_inner(rasterizer, affine_here);
+            }
+            return rasterizer.accumulate();
+        }
+
+        Type m_type { Type::Composite };
+        i16 m_xmin;
+        i16 m_ymin;
+        i16 m_xmax;
+        i16 m_ymax;
```
Dexesttp, line 157:
```suggestion
        i16 m_xmin { 0 };
        i16 m_ymin { 0 };
        i16 m_xmax { 0 };
        i16 m_ymax { 0 };
```

Libraries/LibTTF/Tables.h
```diff
@@ -0,0 +1,145 @@
...
+
+private:
+    enum class Sizes {
+        LongHorMetric = 4,
+        LeftSideBearing = 2
+    };
+
+    ByteBuffer m_slice;
+    u32 m_num_glyphs;
+    u32 m_number_of_h_metrics;
```
Dexesttp, line 142:
```suggestion
    u32 m_num_glyphs { 0 };
    u32 m_number_of_h_metrics { 0 };
```

Libraries/LibTTF/Glyf.h
```diff
@@ -0,0 +1,182 @@
...
+        case IndexToLocFormat::Offset32:
+            ASSERT(m_slice.size() >= m_num_glyphs * 4);
+            break;
+        }
+    }
+    u32 get_glyph_offset(u32 glyph_id) const;
+
+private:
+    ByteBuffer m_slice;
+    u32 m_num_glyphs;
```
Dexesttp, line 71:
```suggestion
    u32 m_num_glyphs { 0 };
```

SrimantaBarua:
@Dexesttp in `draw_line`, I'm calculating the "coverage" of pixels that a line passes through. e.g. assuming a line drawn from `(0.2, 0.3)` to `(3.4, 4.5)`. Since this line goes upwards, we're going to add a float value to each pixel this line passes through corresponding to the area of the pixel to the right of the line. If a pixel isn't fully covered (based on y, i.e. "full coverage" for `(0, 0)` in this example is 0.7 and `(3, 4)` is 0.5), we're going to add (full_coverage - area) to the pixel on its right. If a line goes down instead, we're going to do the same thing but subtracting. The area between the line going up and the line going down would then be the interior of the glyph.

The accumulator "accumulates" this float value. So it would be 1.0 in the interior region, 0.0 outside, and somewhere in between on the border pixels. Here I'm getting the `abs()`, in case in the glyph outline, the line on the left goes down and the line on the right goes up, in which case the accumulated value would be negative.

alimpfard:
So...calculating a winding number?
never seen that done with floats, I thought you were trying to do AA while I was reading that

SrimantaBarua:
Basically yes. The accumulator loop goes over each scanline, and adds/subtracts a value based on whether the pixel is inside or outside the glyph. But it does AA at the same time, since the value is a float between 0 and 1 corresponding to "how much" of a pixel is to the right of a upward-moving line, or the left of a downward moving line.

It's actually based on this article by Raph Levien: https://medium.com/@raphlinus/inside-the-fastest-font-renderer-in-the-world-75ae5270c445

alimpfard:
I see. sounds like a neat idea.

Would be cool to have this as part of `Gfx::Path` and `Gfx::Painter` interactions instead, it sounds very useful for other applications. but I'm not gonna push on that, so feel free to ignore this~

SrimantaBarua:
Sure, but maybe it should be done after people have tested this implementation enough.

I had a few questions about what sort of API you guys feel this should have, keeping in mind that it will be used later by `Painter` (so probably shouldn't be too high level). The current code provides -

- `glyph_id_for_codepoint()` since if/when a shaper is written, glyphs should be queried by glyph IDs.
- `metrics()` to get overall font metrics (ascender, descender, design line gap, max advance width)
- `glyph_metrics()` to get glyph-specific metrics (ascender, descender, advance width, left side bearing) that can be used to position the "pen" before/after drawing a glyph.
-  `raster_glyph()` to get a bitmap after rasterizing a glyph.
- Utility functions like `glyph_count()`, to get number of glyphs (I've seen that used somewhere in `Painter`), and `width()` to get the width that would be required to draw a `StringView`, `Utf8View`, or `Utf32View`

Anything else you think I should add?

Dexesttp:
> Anything else you think I should add?

Actually, I think having a simple `ScaledFont::paint_text(Gfx::Painter, const [StringView/UTF8View/UTF32View]&, const Gfx::IntPoint top_left)` method rather than only low-level things could be the way to go. The `ScaledFont::width(const [StringView/UTF8View/UTF32View]&)` method would be needed too, but the other methods are not really required.

Based on what was discussed earlier in the PR, the "end goal" would be to have an abstract `Font` interface which can be used to draw text regardless of what kind of font backs it up. A high-level method would work for that goal, since it would allow any "kind" of font to manually handle extra position adjustments, create ligatures, etc... And the painter would only need to handle centering the text, splitting lines, and similar things.

So my advice would be for `ScaledFont` to expose just enough data for positionning text, but no low-level things which wouldn't necessarily have equivalents or would be unwieldy for other kinds of fonts (emoji and system fonts):
- `int width(const SomethingView&) const`
- `int baseline_height(const SomethingView&) const`
- `int full_line_height(const SomethingView&) const` (optiolnal, might be useful)
- `Vector<SomethingView&> split_text(const SomethingView&, int width, bool can_break_words) const` (optional, would be used to help decide where to split the text. can_break_words should be an enum too, but it's for the idea)
- `paint_text(Gfx::Painter, const SomethingView&, const Gfx::Point top_left) const`

I'm not sure if it's the best solution, though. Your low-level one might actually work better.

Also, I think the API would be easier to reason about after creating a TTFFont test program that tries to use it ^^

SrimantaBarua:
Hey! To update, I've been juggling quite a few responsibilities IRL, haven't found time to devote to this. Will get back to it as soon as I can, around a week or so.

sunverwerth:
Hi @SrimantaBarua are you still going to work on this? I would like to continue work starting from your branch if you don't mind. Not right now but next week or so.

SrimantaBarua:
Hey guys, sorry about the complete lack of activity. I (quite unexpectedly) started with grad school this August! Haven't had time at all.
@sunverwerth sure, go ahead! It'd be great to see TTF in Serenity. :)

sunverwerth:
Ayyy :) good luck with your studies, buddy!

SrimantaBarua:
Thanks! :)

sunverwerth:
@awesomekling I've rebased this branch and pushed to sunverwerth:serenity/truetype I think you can edit this PR to point to the new repo

danboid:
Great!

Choice avatar you have there @sunverwerth ! The Amiga passed @awesomekling by, poor boy!

awesomekling:
I don’t see an option to switch the source branch of the PR.. maybe I’m missing something? Otherwise you can just open a new one :)

sunverwerth:
@awesomekling you're right. You can only change the target branch, I'll open a new one.
@danboid Amiga forever!

PR #2514, opened 2020-06-06
cmake: Make setting CMAKE_BUILD_TYPE an error.

nico:
Not sure if this is correct!

I tried setting it to Release, then noticed that it didn't build
due to gcc's optimizer-level dependent warnings and -Werror, then
started fixing the warnings for a bit (all false positives),
then looked at the global CMakeLists.txt and realized that the
default build is aleady using compiler optimizations. It looks like
people aren't supposed to change this, so make that explicit to
be friendly to people familiar with cmake but new to serenity.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,13 @@
 cmake_minimum_required(VERSION 3.16)
 project(SerenityOS C CXX ASM)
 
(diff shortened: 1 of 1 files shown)


awesomekling:
Seems fine to me! Indeed, we're going the "optimized with assertions" route. :)

PR #2515, opened 2020-06-06
Theme: Discord Theme Created

nanoproductions:
A New Theme for Serenity which brings colors from the Discord Theme colors. Some colors were modified to ensure great looks on Serenity.

image


jsdevv:
Looks good :D

awesomekling:
This looks quite ugly to me but it's not like we don't have ugly themes already. :)

PR #2516, opened 2020-06-06
LibWeb: Fully implement all script tokenizer states

Lubrsi:
Also fixes RAWTEXTLessThanSign having a separate emit and reconsume.


awesomekling:
Looks good :)

PR #2517, opened 2020-06-06
LibWeb: Fix location.reload.length

Lubrsi:
This was accidentally being set to JS::Attribute::Enumerable
instead of 0.


awesomekling:
Oopsie!

PR #2518, opened 2020-06-06
LibWeb: Add initial support for the history object.

Lubrsi:
This adds initial support for window.history. It currently only
includes the navigation functions and the length property.

~~Marked as draft as I forgot the events.~~ Nevermind, the events should also be fired when pressing the back/forward buttons. We simply don't support them quite yet. It should also be asynchronous but that can probably be done another time.

Not sure why the style check is failing.

### DIFF
--- a/Applications/Browser/Tab.cpp
+++ b/Applications/Browser/Tab.cpp
@@ -171,6 +171,31 @@ Tab::Tab()
         }
     };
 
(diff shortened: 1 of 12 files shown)


Lubrsi:
Okay, done 👍
I haven't really looked into the refactor, so I'm not sure if it's the right way to get to the tab's history now.

awesomekling:

PR #2519, opened 2020-06-06
Toolchain: Fix building the toolchain on macOS

Pagghiu:
This PR restores building the toolchain and serenityOS on macOS, fixing #2234.
The root problem is how the current CMake build system has been setup for cross compiling host tools and the actual OS.
This PR builds host tools (IPCCompiler etc.) in a separate CMake invocation and then passes the CMake host tools build output directory to the regular CMake invocation for OS build.
The problem exists because changing compiler in the middle of a single CMake invocation is not supported. In other words targets using different compilers cannot be defined and built in a single CMake invocation.
Unfortunately CMake makes it not so obvious, because it actually allows to change `CMAKE_C_COMPILER` and friends, like currently being done in Serenity root `CMakeLists.txt`. It's not possible, to my knowledge, modifying some other variables impacting the build like the autodetected `CXXFLAGS` that are automatically appended after user specified ones, filled before declaring the `project()` section.
Current single invocation CMake build works on recent Linux because such autodetected flags are compatible enough not to causing problem when passed serenityOS GCC.
On macOS instead (and probably on other hosts as well), if you don't flag the CMake Project as a "cross-compiling" build, by changing `CMAKE_SYSTEM_NAME` before opening the `project()` section, you will get the autodetected `CXXFLAGS` filled with things like `-mmacosx-version-min=10.14` and `-isysroot=...` even when cross compiling using serenity custom GCC (i686-pc-serenity).
In this case serenity GCC will complain about wrong/unknown options and fail to build soon after building host tools.
Of course if you set `CMAKE_SYSTEM_NAME` to something like `Generic` unconditionally, host tools will not compile because macOS system compiler (Apple Clang included inside XCode) will now loose the important autodetected flag (`-isysroot` and `-mmacosx-version-min`).

I am not a CMake expert (because I actually dislike it a lot) and there are obviously many ways and opinions on how to solve this problem, but that's a start ;)

I have tried to make this PR as transparent as possible to avoid breaking current linux build, that still works with a single CMake invocation. I've tested it on a Ubuntu 2020.04 VM and it looks fine.
I think, however, that it could be a good idea splitting the CMake Build in two invocations even when building under Linux, to future-proof the build system when porting it to other unix-es or beyond 😄

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,6 +23,7 @@
 *.autosave
 Meta/Lagom/build
 Build
+BuildHost
 build
 CMakeFiles
 Toolchain/Tarballs
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,17 @@
 cmake_minimum_required(VERSION 3.16)
+
+if (BUILD_CROSS_OS_PHASE2)
+set(CMAKE_SYSTEM_NAME Generic)
+endif()
+
+if (NOT DEFINED BUILD_HOST_TOOLS_PHASE1)
+set(BUILD_HOST_TOOLS_PHASE1 True)
+endif()
+
+if (NOT DEFINED BUILD_CROSS_OS_PHASE2)
+set(BUILD_CROSS_OS_PHASE2 True)
+endif()
+
 project(SerenityOS C CXX ASM)
 
 if(NOT "${CMAKE_BUILD_TYPE}" STREQUAL "")
@@ -45,11 +58,38 @@ set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra -Werror -std=c++2a -fdiagn
 include_directories(Libraries)
 include_directories(.)
 
+if(BUILD_HOST_TOOLS_PHASE1)
+
 add_subdirectory(Meta/Lagom)
 add_subdirectory(DevTools/IPCCompiler)
 add_subdirectory(DevTools/FormCompiler)
 add_subdirectory(Libraries/LibWeb/CodeGenerators)
 add_subdirectory(AK/Tests)
+
+else()
+
+# Fake empty Targets needed when building with a two phase build
+add_custom_target(IPCCompiler)
+add_custom_target(FormCompiler)
+add_custom_target(WrapperGenerator)
+add_custom_target(Generate_CSS_PropertyID_h)
+add_custom_target(Generate_CSS_PropertyID_cpp)
+
+endif()
+
+if(DEFINED BUILD_HOST_TOOLS_LOCATION)
+set(IPC_COMPILER_EXECUTABLE "${BUILD_HOST_TOOLS_LOCATION}/DevTools/IPCCompiler/IPCCompiler")
+set(LIBWEB_WRAPPER_GENERATOR_EXECUTABLE "${BUILD_HOST_TOOLS_LOCATION}/Libraries/LibWeb/CodeGenerators/WrapperGenerator")
+set(LIBWEB_GENERATE_CSS_PROPERTYID_H_EXECUTABLE "${BUILD_HOST_TOOLS_LOCATION}/Libraries/LibWeb/CodeGenerators/Generate_CSS_PropertyID_h")
+set(LIBWEB_GENERATE_CSS_PROPERTYID_CPP_EXECUTABLE "${BUILD_HOST_TOOLS_LOCATION}/Libraries/LibWeb/CodeGenerators/Generate_CSS_PropertyID_cpp")
+else()
+set(IPC_COMPILER_EXECUTABLE "IPCCompiler")
+set(LIBWEB_WRAPPER_GENERATOR_EXECUTABLE "WrapperGenerator")
+set(LIBWEB_GENERATE_CSS_PROPERTYID_H_EXECUTABLE "Generate_CSS_PropertyID_h")
+set(LIBWEB_GENERATE_CSS_PROPERTYID_CPP_EXECUTABLE "Generate_CSS_PropertyID_cpp")
+endif()
+
+if (BUILD_CROSS_OS_PHASE2)
 
 function(serenity_install_headers target_name)
     file(GLOB_RECURSE headers RELATIVE ${CMAKE_CURRENT_SOURCE_DIR} "*.h")
@@ -96,7 +136,7 @@ function(compile_ipc source output)
     set(source ${CMAKE_CURRENT_SOURCE_DIR}/${source})
     add_custom_command(
         OUTPUT ${output}
-        COMMAND IPCCompiler ${source} > ${output}
+        COMMAND ${IPC_COMPILER_EXECUTABLE} ${source} > ${output}
         VERBATIM
         DEPENDS IPCCompiler
         MAIN_DEPENDENCY ${source}
@@ -149,3 +189,5 @@ add_subdirectory(MenuApplets)
 add_subdirectory(Shell)
 add_subdirectory(Demos)
 add_subdirectory(Userland)
+
+endif()
--- a/Libraries/LibC/CMakeLists.txt
+++ b/Libraries/LibC/CMakeLists.txt
@@ -55,10 +55,17 @@ set(ELF_SOURCES ${ELF_SOURCES} ../LibELF/Arch/i386/plt_trampoline.S)
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DSERENITY_LIBC_BUILD")
 
 add_library(crt0 STATIC crt0.cpp)
-add_custom_command(
-    TARGET crt0
-    COMMAND install -D $<TARGET_OBJECTS:crt0> ${CMAKE_INSTALL_PREFIX}/usr/lib/crt0.o
-)
+if(CMAKE_HOST_APPLE)
+    add_custom_command(
+        TARGET crt0
+        COMMAND ginstall -D $<TARGET_OBJECTS:crt0> ${CMAKE_INSTALL_PREFIX}/usr/lib/crt0.o
+    )
+else() 
+    add_custom_command(
+        TARGET crt0
+        COMMAND install -D $<TARGET_OBJECTS:crt0> ${CMAKE_INSTALL_PREFIX}/usr/lib/crt0.o
+    )
+endif()
 
 set(SOURCES ${LIBC_SOURCES} ${AK_SOURCES} ${ELF_SOURCES})
 serenity_libc(LibC c)
--- a/Libraries/LibWeb/CMakeLists.txt
+++ b/Libraries/LibWeb/CMakeLists.txt
@@ -136,7 +136,7 @@ function(libweb_js_wrapper class)
     add_custom_command(
         OUTPUT Bindings/${class}Wrapper.h
         COMMAND /bin/mkdir -p Bindings
-        COMMAND WrapperGenerator --header ${CMAKE_CURRENT_SOURCE_DIR}/DOM/${class}.idl > Bindings/${class}Wrapper.h
+        COMMAND ${LIBWEB_WRAPPER_GENERATOR_EXECUTABLE} --header ${CMAKE_CURRENT_SOURCE_DIR}/DOM/${class}.idl > Bindings/${class}Wrapper.h
         VERBATIM
         DEPENDS WrapperGenerator
         MAIN_DEPENDENCY DOM/${class}.idl
@@ -144,7 +144,7 @@ function(libweb_js_wrapper class)
     add_custom_command(
         OUTPUT Bindings/${class}Wrapper.cpp
         COMMAND /bin/mkdir -p Bindings
-        COMMAND WrapperGenerator --implementation ${CMAKE_CURRENT_SOURCE_DIR}/DOM/${class}.idl > Bindings/${class}Wrapper.cpp
+        COMMAND ${LIBWEB_WRAPPER_GENERATOR_EXECUTABLE} --implementation ${CMAKE_CURRENT_SOURCE_DIR}/DOM/${class}.idl > Bindings/${class}Wrapper.cpp
         VERBATIM
         DEPENDS WrapperGenerator
         MAIN_DEPENDENCY DOM/${class}.idl
@@ -171,7 +171,7 @@ set(SOURCES ${SOURCES} ${WRAPPER_SOURCES})
 add_custom_command(
     OUTPUT CSS/PropertyID.h
     COMMAND /bin/mkdir -p CSS
-    COMMAND Generate_CSS_PropertyID_h ${CMAKE_CURRENT_SOURCE_DIR}/CSS/Properties.json > CSS/PropertyID.h
+    COMMAND ${LIBWEB_GENERATE_CSS_PROPERTYID_H_EXECUTABLE} ${CMAKE_CURRENT_SOURCE_DIR}/CSS/Properties.json > CSS/PropertyID.h
     VERBATIM
     DEPENDS Generate_CSS_PropertyID_h
     MAIN_DEPENDENCY CSS/Properties.json
@@ -181,7 +181,7 @@ add_custom_target(generate_PropertyID.h DEPENDS CSS/PropertyID.h)
 add_custom_command(
     OUTPUT CSS/PropertyID.cpp
     COMMAND /bin/mkdir -p CSS
-    COMMAND Generate_CSS_PropertyID_cpp ${CMAKE_CURRENT_SOURCE_DIR}/CSS/Properties.json > CSS/PropertyID.cpp
+    COMMAND ${LIBWEB_GENERATE_CSS_PROPERTYID_CPP_EXECUTABLE} ${CMAKE_CURRENT_SOURCE_DIR}/CSS/Properties.json > CSS/PropertyID.cpp
     VERBATIM
     DEPENDS Generate_CSS_PropertyID_cpp
     MAIN_DEPENDENCY CSS/Properties.json
--- a/Toolchain/BuildIt.sh
+++ b/Toolchain/BuildIt.sh
@@ -1,6 +1,11 @@
 #!/usr/bin/env bash
 set -e
+if [ "$(uname)" = "Darwin" ]; then
+INSTALL=ginstall
+else
 shopt -s globstar
+INSTALL=install
+fi
 # This file will need to be run in bash, for now.
 
 
@@ -14,6 +19,7 @@ ARCH=${ARCH:-"i686"}
 TARGET="$ARCH-pc-serenity"
 PREFIX="$DIR/Local"
 BUILD=$(realpath "$DIR/../Build")
+BUILDHOST=$(realpath "$DIR/../BuildHost")
 SYSROOT="$BUILD/Root"
 
 MAKE="make"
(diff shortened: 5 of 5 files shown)


bugaevc:
Let's try to *not* make anything in the build system conditional on "Apple". There are other systems that use non-gcc as their default compiler, if that's what matters.

Pagghiu:
The best and cleanest way to do it would be to split the CMake invocations for host tools and for everything else for all hosts (Linux, macOS or whatever). Should I go for it?

bugaevc:
I'd like to avoid that if possible too. As it is, CMake will pick up all the changes that influence code generators, rebuild those, regenerate the code, rebuild things that depend on them — all automatically, whenever you modify some code in, say, AK, and this is because it keeps one single task/target graph with all the dependencies, and properly tracks things like generators and generated code.

Pagghiu:
Ok, one way to do this would be swapping "Apple" checks with "NOT Linux" checks.

bugaevc:
"Not Linux" is wrong too; what does it matter what the kernel is?

It would make sense to check if the host compiler is GCC (and thus accepts the same flags as the cross-compiler, which is always GCC).

jonnii:
I can confirm that fixes the build on my machine and I can successfully boot serenity again! Thanks!

Pagghiu:
> "Not Linux" is wrong too; what does it matter what the kernel is?
> 
> It would make sense to check if the host compiler is GCC (and thus accepts the same flags as the cross-compiler, which is always GCC).

"check if the host compiler is GCC" is wrong too; what does it matter what the host compiler is?

CMake extrapolates options based on current architecture and sensible host OS defaults. Flags and defaults that it extrapolates for the host compiler, depend on the host OS. While these being valid GCC style options, they may be wrong if CMake is passing them anyway on the command line when cross compiling serenity, causing failed builds or even more obscure wrong behaviors.

If the problem is just making sure that a change in base library (like AK) is correctly causing a rebuild of host tools and then the OS with a single CMake invocation, probably custom commands or external projects could be investigated. In my opinion they smell as hacks as well and they just look like glorified shell scripts in disguise. Adding complex build logic should not really be handled by a language with such poor design like CMake if one is striving to keep things simple.

That being said I am not sure if such advanced CMake hacking, to go around one of its fundamental limitation, is more readable and easier to maintain than a two lines shell script issuing the two CMake invocations, considering that this project uses already a large number of shell scripts to setup many things.

One thing that can be done is moving the Apple (or NOT OS_That_Uses_GCC_With_Same_Options_As_Serenity_cross_compiler_so_that_we_can_build_it_in_a_single_cmake_invocation) check in  `BuildIt.sh` to allow passing a generic "two phase build" flag to cmake invocations. 
This is just moving such check earlier in the build pipeline though. 
It can be done however if this is considered to be more appropriate for how things are currently setup.

bugaevc:
> "check if the host compiler is GCC" is wrong too; what does it matter what the host compiler is?

Because, if I understood you and other people who try to build Serenity on systems where Clang is the host compiler, the issue that arises is CMake shares some options between the host and the cross compilers, and there's no guarantee both understand the same set of options. You're right that the deeper problem is that even if the two compilers support those options, it doesn't mean it makes sense to share those options.

> If the problem is just making sure that a change in base library (like AK) is correctly causing a rebuild of host tools and then the OS with a single CMake invocation, probably custom commands or external projects could be investigated. In my opinion they smell as hacks as well and they just look like glorified shell scripts in disguise. Adding complex build logic should not really be handled by a language with such poor design like CMake if one is striving to keep things simple.

Not *any* change in AK, only those changes that affect the relevant host tools; and then only files generated by those updated tools will get regenerated, and so on. And there's little to no hacks involved in this, at least so far: it's all handled by CMake's native code generation facilities and the target dependency graph. CMake knows that this generator depends on this and that (transitive dependencies, and implicit header dependencies, are tracked properly too), and this generated code depends on the generator, and voila, you have proper dependency chains and incremental builds. How would you achieve this with two-phase builds?

vkoskiv:
I hate to bump a PR that I have nothing to do with, but what's the status of this?

Anything I can do to help get this merged? I'm asking for completely selfish reasons - I'm one of the freaks trying to poke at SerenityOS on a macOS host!

Pagghiu:
> I hate to bump a PR that I have nothing to do with, but what's the status of this?
> 
> Anything I can do to help get this merged? I'm asking for completely selfish reasons - I'm one of the freaks trying to poke at SerenityOS on a macOS host!

Honestly I don't know how to modify this PR to get it merged, so I am looking for any input.

I have been updating the PR to latest master, and it works on my mac.

> How would you achieve this with two-phase builds?

I think that a 2 phase build is fine, you can build host tools just before building the actual os with two shell commands.
With this approach, modification in the actual host tools own source code will not be triggering a regeneration the target files that are being built/generated by them. Modifications in AK and such are probably triggering rebuild of almost everything so it  should not be a big deal.
To achieve something that is 1:1 with current approach, the only idea that I have is to create some "fake targets" for host tools, that will not be actually built/compiled during phase 2 (os build) but that can be stilly used as a Dependency to flag for rebuild all of the generated source code when changes in host tools happen.

vkoskiv:
I tested this on my system (10.15), and it's indeed working perfectly. As long as it doesn't affect building on Linux (one would assume, given the CI passes), I vote that this could be merged in.

@awesomekling @bugaevc Thoughts?

nooga:
Tested on 10.15 - verks. Ditto @vkoskiv

vkoskiv:
On my system, I didn't get any untracked files. At least `git status` didn't show any.

bugaevc:
> @awesomekling @bugaevc Thoughts?

I still think that:

* hardcoding "Apple" is wrong (the same issues are faced by people compiling Serenity on BSDs, I hear)
* two-phase builds is a non-starter, this is not how CMake is supposed to work

But on the other hand, if people need this badly, and it only affects Apple hosts and doesn't make situation worse for free systems, and someone is working on a better solution (anybody?...), then I wouldn't object to merging this.

Pagghiu:
> > @awesomekling @bugaevc Thoughts?
> 
> I still think that:
> 
> * hardcoding "Apple" is wrong (the same issues are faced by people compiling Serenity on BSDs, I hear)

I have modified the PR to introduce the concept of an optional two phase build. 
I am doing platform detection inside `BuildIt.sh` instead inside of the CMake File, hoping that this will be considered more acceptable, while still being an hardcoding.
Any other platform that needs to do a two phase build can add itself in that `BuildIt.sh` script, including BSDs.

> * two-phase builds is a non-starter, this is not how CMake is supposed to work

That's an opinion.
CMake is not designed to build software that requires multiple compilers in a single invocation.
Of course nobody is preventing anybody else from abusing external commands or other similar hacks to achieve what they think it's the right way of creating a build. 
I've seen people creating Raytracers in Excel and yeah, they can do it, but it doesn't mean that Excel is supposed to work that way or that's a good idea in general.

> But on the other hand, if people need this badly, and it only affects Apple hosts and doesn't make situation worse for free systems, and someone is working on a better solution (anybody?...), then I wouldn't object to merging this.

I have made it so that non-free Apple systems are using a correct two phase build, while Year of Desktop systems should be able to continue using the single command, single phase build.
I am open for any better solution.

bugaevc:
> Year of Desktop systems

Hey! :slightly_smiling_face: I appreciate your efforts for making the macOS build work, even if we disagree on technical grounds. Besides, this has nothing to do with desktop or GUI.

> I am open for any better solution.

Honest, and perhaps a bit uninformed question — can't you just reset `CMAKE_CXX_FLAGS` when switching to the cross-compiler? We currently have the cross-compiler inherit flags from the host compiler (while appending more flags):

https://github.com/SerenityOS/serenity/blob/6105f063cb936e027b5419ae9696ce00f97991e9/CMakeLists.txt#L131

Can't we just set `CMAKE_CXX_FLAGS` to something *not* based on its previous value? I just checked and that does drop flags inherited from `CXXFLAGS`. Does it not drop `-mmacosx-version-min=10.14` and `-isysroot=...` on macOS?

bugaevc:
> I just checked and that does drop flags inherited from CXXFLAGS. Does it not drop `-mmacosx-version-min=10.14` and `-isysroot=...` on macOS?

After doing some quick research, no it does not, but unsetting `CMAKE_CXX_OSX_DEPLOYMENT_TARGET_FLAG` and `CMAKE_CXX_SYSROOT_FLAG` does the trick.

bugaevc:
So, I can successfully build Serenity on Darwin with just the following patch:

```patch
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 2238ca6e4..a68848efd 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -126,6 +126,13 @@ set(CMAKE_RANLIB ${TOOLCHAIN_PREFIX}ranlib)
 set(CMAKE_STRIP ${TOOLCHAIN_PREFIX}strip)
 set(CMAKE_AR ${TOOLCHAIN_PREFIX}ar)
 
+unset(CMAKE_CXX_OSX_DEPLOYMENT_TARGET_FLAG)
+unset(CMAKE_CXX_SYSROOT_FLAG)
+unset(CMAKE_SHARED_LIBRARY_SONAME_CXX_FLAG)
+set(CMAKE_INSTALL_NAME_TOOL "true")
+set(CMAKE_SHARED_LIBRARY_SUFFIX ".so")
+set(CMAKE_SHARED_LIBRARY_CREATE_CXX_FLAGS "-shared")
+
 #FIXME: -fstack-protector
 
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Os -Wno-sized-deallocation -fno-sized-deallocation -fno-exceptions -fno-rtti -Wno-address-of-packed-member -Wundef -Wcast-qual -Wwrite-strings -Wimplicit-fallthrough -Wno-nonnull-compare -Wno-deprecated-copy -Wno-expansion-to-defined")
diff --git a/Libraries/LibC/CMakeLists.txt b/Libraries/LibC/CMakeLists.txt
index 2d37a9ce3..287597351 100644
--- a/Libraries/LibC/CMakeLists.txt
+++ b/Libraries/LibC/CMakeLists.txt
@@ -54,10 +54,15 @@ set(ELF_SOURCES ${ELF_SOURCES} ../LibELF/Arch/i386/plt_trampoline.S)
 
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DSERENITY_LIBC_BUILD")
 
+find_program(INSTALL_COMMAND ginstall)
+if(NOT INSTALL_COMMAND)
+    set(INSTALL_COMMAND install)
+endif()
+
 add_library(crt0 STATIC crt0.cpp)
 add_custom_command(
     TARGET crt0
-    COMMAND install -D $<TARGET_OBJECTS:crt0> ${CMAKE_INSTALL_PREFIX}/usr/lib/crt0.o
+    COMMAND ${INSTALL_COMMAND} -D $<TARGET_OBJECTS:crt0> ${CMAKE_INSTALL_PREFIX}/usr/lib/crt0.o
 )
 
 set(SOURCES ${LIBC_SOURCES} ${AK_SOURCES} ${ELF_SOURCES})
diff --git a/Toolchain/BuildIt.sh b/Toolchain/BuildIt.sh
index 57098f3a7..ebfced693 100755
--- a/Toolchain/BuildIt.sh
+++ b/Toolchain/BuildIt.sh
@@ -24,6 +24,12 @@ NPROC="nproc"
 # It seems that Travis starts having trouble at 35 entries, so I think this is a good amount.
 KEEP_CACHE_COUNT=8
 
+if command -v ginstall &>/dev/null; then
+    INSTALL=ginstall
+else
+    INSTALL=install
+fi
+
 if [ "$(uname -s)" = "OpenBSD" ]; then
     MAKE=gmake
     MD5SUM="md5 -q"
@@ -230,11 +236,11 @@ pushd "$DIR/Build/"
         pushd "$BUILD"
             CXXFLAGS="-DBUILDING_SERENITY_TOOLCHAIN" cmake ..
             cmake --build . --target LibC
-            install -D Libraries/LibC/libc.a Libraries/LibM/libm.a Root/usr/lib/
+            "$INSTALL" -D Libraries/LibC/libc.a Libraries/LibM/libm.a Root/usr/lib/
             SRC_ROOT=$(realpath "$DIR"/..)
             for header in "$SRC_ROOT"/Libraries/Lib{C,M}/**/*.h; do
                 target=$(echo "$header" | sed -e "s@$SRC_ROOT/Libraries/LibC@@" -e "s@$SRC_ROOT/Libraries/LibM@@")
-                install -D "$header" "Root/usr/include/$target"
+                "$INSTALL" -D "$header" "Root/usr/include/$target"
             done
             unset SRC_ROOT
         popd
```

@Pagghiu @vkoskiv @nooga please verify it works for you :smile:

bugaevc:
> and someone is working on a better solution (anybody?...)

I guess this is me now, heh :slightly_smiling_face:

Pagghiu:
> 
> @Pagghiu @vkoskiv @nooga please verify it works for you 😄

At first you get an issue with building assembly files for LibC build needed for libstdc++

```
[ 33%] Building CXX object Libraries/LibC/CMakeFiles/LibC.dir/scanf.cpp.o
[ 33%] Building CXX object Libraries/LibC/CMakeFiles/LibC.dir/sched.cpp.o
[ 33%] Building CXX object Libraries/LibC/CMakeFiles/LibC.dir/serenity.cpp.o
[ 50%] Building ASM object Libraries/LibC/CMakeFiles/LibC.dir/setjmp.S.o
i686-pc-serenity-gcc: error: unrecognized command-line option '-mmacosx-version-min=10.14'
make[3]: *** [Libraries/LibC/CMakeFiles/LibC.dir/setjmp.S.o] Error 1
make[2]: *** [Libraries/LibC/CMakeFiles/LibC.dir/all] Error 2
make[1]: *** [Libraries/LibC/CMakeFiles/LibC.dir/rule] Error 2
make: *** [LibC] Error 2
~ cmake --version
cmake version 3.16.5
```

that I think I got past by adding

```
unset(CMAKE_ASM_OSX_DEPLOYMENT_TARGET_FLAG)
unset(CMAKE_ASM_SYSROOT_FLAG)
```
(I don't even know if these variable exist but I've tried and it seems to fix it)

Even doing that you are able to build host tools but it stops when trying to do a cross-compile build of LibCore, because it looks like you still have the fancy ` '-mmacosx-version-min=10.14'`

```
[  3%] Built target LagomCore
[  3%] Built target IPCCompiler
[  3%] Built target FormCompiler
[  3%] Built target WrapperGenerator
[  3%] Built target Generate_CSS_PropertyID_cpp
[  3%] Built target Generate_CSS_PropertyID_h
[  3%] Built target TestOptional
[  3%] Built target TestWeakPtr
[  4%] Built target TestVector
[  4%] Built target TestURL
[  4%] Built target TestStringUtils
[  4%] Built target TestStringView
[  4%] Built target TestString
[  4%] Built target TestNonnullRefPtr
[  5%] Built target TestHashMap
[  5%] Built target TestFixedArray
[  5%] Built target TestLexicalPath
[  5%] Built target TestUtf8
[  5%] Built target TestCircularDeque
[  5%] Built target TestJSON
[  5%] Built target TestBinarySearch
[  6%] Built target TestRefPtr
[  6%] Built target TestBitmap
[  7%] Built target TestBase64
[  7%] Built target TestQueue
[  7%] Built target TestCircularQueue
[  7%] Built target TestAtomic
[  7%] Built target boot
[ 18%] Built target Kernel
[ 18%] Built target TestModule
[ 18%] Built target generate_AudioServerEndpoint.h
[ 19%] Built target crt0
[ 19%] Built target LibM
[ 24%] Built target LibC
[ 24%] Building C object Libraries/LibCore/CMakeFiles/LibCore.dir/puff.c.o
i686-pc-serenity-gcc: error: unrecognized command-line option '-mmacosx-version-min=10.14'
make[2]: *** [Libraries/LibCore/CMakeFiles/LibCore.dir/puff.c.o] Error 1
make[1]: *** [Libraries/LibCore/CMakeFiles/LibCore.dir/all] Error 2
make: *** [all] Error 2
```

bugaevc:
Try adding

```
unset(CMAKE_C_OSX_DEPLOYMENT_TARGET_FLAG)
unset(CMAKE_C_SYSROOT_FLAG)
```

Pagghiu:
> Try adding
> 
> ```
> unset(CMAKE_C_OSX_DEPLOYMENT_TARGET_FLAG)
> unset(CMAKE_C_SYSROOT_FLAG)
> ```

Oh, I could have been figuring this out myself as it's the same done for asm files :)

This gets to the point of building the entire OS but I get a Warning from ld that I've never seen when building every executable and the resulting image doesn't boot.

```
[100%] Built target IRCClient
[100%] Linking CXX executable Browser
/Users/stefano/Development/Opensource/OS/serenity/Toolchain/Local/lib/gcc/i686-pc-serenity/10.1.0/../../../../i686-pc-serenity/bin/ld: warning: cannot find entry symbol arch_paths_first; defaulting to 00000000080480b0
[100%] Built target Browser
[100%] Linking CXX executable HackStudio
/Users/stefano/Development/Opensource/OS/serenity/Toolchain/Local/lib/gcc/i686-pc-serenity/10.1.0/../../../../i686-pc-serenity/bin/ld: warning: cannot find entry symbol arch_paths_first; defaulting to 00000000080480b0
[100%] Built target HackStudio
```

bugaevc:
> `serenity/Toolchain/Local/lib/gcc/i686-pc-serenity/10.1.0/../../../../i686-pc-serenity/bin/ld: warning: cannot find entry symbol arch_paths_first; defaulting to 00000000080480b0`

Hm, so it sounds like it still tries to pass some weird flags to the linker. Could you figure out which ones? I'd run the build with `VERBOSE=1` and find the appropriate line(s).

Pagghiu:
> > `serenity/Toolchain/Local/lib/gcc/i686-pc-serenity/10.1.0/../../../../i686-pc-serenity/bin/ld: warning: cannot find entry symbol arch_paths_first; defaulting to 00000000080480b0`
> 
> Hm, so it sounds like it still tries to pass some weird flags to the linker. Could you figure out which ones? I'd run the build with `VERBOSE=1` and find the appropriate line(s).

Looks like the weird flags are `-Wl,-search_paths_first -Wl,-headerpad_max_install_names`
`
/serenity/Toolchain/Local/bin/i686-pc-serenity-g++  -DBUILDING_SERENITY_TOOLCHAIN -Wall -Wextra -Werror -std=c++2a -fdiagnostics-color=always -Os -Wno-sized-deallocation -fno-sized-deallocation -fno-exceptions -fno-rtti -Wno-address-of-packed-member -Wundef -Wcast-qual -Wwrite-strings -Wimplicit-fallthrough -Wno-nonnull-compare -Wno-deprecated-copy -Wno-expansion-to-defined -DDEBUG -DSANITIZE_PTRS -Wl,-search_paths_first -Wl,-headerpad_max_install_names
`

I got rid of all offending flags with the following loop
`
foreach(lang C CXX OBJC OBJCXX)
  unset(CMAKE_SHARED_LIBRARY_SONAME_${lang}_FLAG)
  unset(CMAKE_${lang}_OSX_DEPLOYMENT_TARGET_FLAG)
  unset(CMAKE_${lang}_SYSROOT_FLAG)
  unset(CMAKE_${lang}_OSX_COMPATIBILITY_VERSION_FLAG)
  unset(CMAKE_${lang}_OSX_CURRENT_VERSION_FLAG)
  unset(CMAKE_${lang}_LINK_FLAGS)
  unset(CMAKE_SHARED_LIBRARY_CREATE_${lang}_FLAGS)
  unset(CMAKE_SHARED_MODULE_CREATE_${lang}_FLAGS)
  unset(CMAKE_SHARED_MODULE_LOADER_${lang}_FLAG )
endforeach()
`

And now everything compiles and boots fine 👍 
I've added a few changes to `BuildIt.sh` to avoid needing bash 4.
Do you want me to push these changes based on your work on my branch here or will you be creating your own PR?

bugaevc:
> And now everything compiles and boots fine 👍

Yay! 😀

> Do you want me to push these changes based on your work on my branch here or will you be creating your own PR?

Feel free to push then yourself; but consider making it a new PR because it's a very different solution to the one originally proposed here. Pushing it to this PR is also fine, on the other hand, so up to you.

Pagghiu:
Ok, it makes sense to keep the two phases build approach as described here, also because it has lower "hack-factor" and it may become useful again in the future, maybe if we want to allow building everything on non unix platforms (Ex. Windows through MSVC).

Thanks for the input, I will be creating another PR, linking to this one as historical reference, and then closing this one.

bugaevc:
> maybe if we want to allow building everything on non unix platforms (Ex. Windows through MSVC).

Side note: I don't think that's ever going to happen. The code generators use LibCore, and LibCore is very Unix-specific.

PR #2521, opened 2020-06-07
LibJS: Lex and parse regex literals, add RegExp objects

mattco98:
This adds regex lexing/parsing, as well as a relatively empty `RegExpObject`. The purpose of this patch is to allow the engine to not get hung up on parsing regexes. This will aid in finding new syntax errors (say, from google or twitter) without having to replace all of their regexes first!

The objects added in this PR aren't perfect, but the primary point of this PR is to do the lexing and parsing. The objects can be ironed out in future PRs when LibRegex gets added.


awesomekling:
Great idea :)

nooga:
Awesome stuff @mattco98! I was working on the same idea today but haven't finished.

mattco98:
Thanks @nooga! Sorry about the duplicated work though 😅

PR #2522, opened 2020-06-07
LibJS: Add interpreter exception checks

mattco98:


mattco98:
Looks like this worked locally before a rebase, but not with the BigInt stuff. Don't have time to look at it right now, I'll check it out tonight :)

PR #2526, opened 2020-06-07
LibLine: Correctly handle line content overflow when on last line

### LINKED ISSUE #2525: LibLine: Strange behavior when reaching end of bottom row
To reproduce: 
1) Get to the bottom of the terminal (press enter in the shell a bunch of times)
2) Type something until you get to the end of the line and try crossing into the next line

cc @alimpfard

alimpfard:
Fixes #2525

### DIFF
--- a/Libraries/LibLine/Editor.cpp
+++ b/Libraries/LibLine/Editor.cpp
@@ -556,7 +556,7 @@ void Editor::handle_read_event()
                 ctrl_held = false;
                 continue;
             default:
-                dbgprintf("Shell: Unhandled final: %02x (%c)\r\n", codepoint, codepoint);
+                dbgprintf("LibLine: Unhandled final: %02x (%c)\r\n", codepoint, codepoint);
                 m_state = InputState::Free;
                 ctrl_held = false;
                 continue;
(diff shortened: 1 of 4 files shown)


Libraries/LibLine/Editor.cpp
```diff
@@ -992,6 +993,16 @@ void Editor::refresh_display()
         m_cached_buffer_size = m_buffer.size();
         return;
     }
+    // We might be at the last line, and have more than one line;
+    // Refreshing the display will cause the terminal to scroll,
+    // so note that fact and bring origin up.
+    auto current_num_lines = num_lines();
+    if (m_origin_x + current_num_lines > m_num_lines + 1) {
```
awesomekling, line 1000: Adding together x (horizontal) and lines (vertical) doesn't look right. What's happening here?

alimpfard: the x and y are inverted here, x is the vertical axis, y is the horizontal axis.
kinda odd, but :shrug:

alimpfard: I'm not even sure why I named them like this, I could go through and fix their names.
will take but a moment.

awesomekling: Yeah let's fix that. This is way confusing :)

alimpfard: There we go.
no more confusion :)

alimpfard: ugh there's more of them T-T

PR #2528, opened 2020-06-08
Misc fixes

alimpfard:
Some teeny tiny misc fixes and improvements

- emitting plaintext when `!isatty(stdout)` in a few places
- highlighting `foo >bar` and co.

:thinking: I guess @bugaevc might know of some other places to do this in...I summon thee!

### DIFF
--- a/Userland/ls.cpp
+++ b/Userland/ls.cpp
@@ -49,7 +49,7 @@
 static int do_file_system_object_long(const char* path);
 static int do_file_system_object_short(const char* path);
 
-static bool flag_colorize = true;
+static bool flag_colorize = false;
 static bool flag_long = false;
 static bool flag_show_dotfiles = false;
 static bool flag_show_inode = false;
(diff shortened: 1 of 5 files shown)


Userland/ls.cpp
```diff
@@ -300,7 +305,6 @@ bool print_filesystem_object(const String& path, const String& name, const struc
         printf("  %4u,%4u ", major(st.st_rdev), minor(st.st_rdev));
     } else {
         if (flag_human_readable) {
-            ASSERT(st.st_size > 0);
```
alimpfard, line 303: This makes no sense, why was it here at all?
```sh
ls > foo -alh
```
would trip this, why do we care?

alimpfard:
whelp
guess we'll talk in IRC lol

Userland/ls.cpp
awesomekling: We don't care. No idea why that was there :)

PR #2534, opened 2020-06-08
Multi-instance services & accept socket connections

bugaevc:
You can now ask SystemServer to not only listen for connections on the socket, but to actually accept them, and to spawn an instance of the service for each client connection. In this case, it's the accepted, not listening, socket that the service processes will receive using socket takeover.

I haven't added a helper like the existing `Core::LocalServer::take_over_from_system_server()` to `Core::LocalSocket`, because I don't want to duplicate the logic, and I have no idea where a shared version of this could live. Suggestions welcome.

### DIFF
--- a/Base/usr/share/man/man5/SystemServer.md
+++ b/Base/usr/share/man/man5/SystemServer.md
@@ -28,6 +28,14 @@ describing how to launch and manage this service.
 * `WorkingDirectory` - the working directory in which the service is spawned. By default, services are spawned in the root (`"/"`) directory.
 * `BootModes` - a comma-separated list of boot modes the service should be enabled in. By default, services are only enabled in the "graphical" mode. The current boot mode is read from the kernel command line, and is assumed to be "graphical" if not specified there.
 * `Environment` - a space-separated list of "variable=value" pairs to set in the environment for the service.
+* `MultiInstance` - whether multiple instances of the service can be running simultaneously.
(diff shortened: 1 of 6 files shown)


alimpfard:
I'm very confused about the utility of this...but it sounds like a neat capability to have.

bugaevc:
Note that:
* This (accepting connections) is how inetd has worked on other Unixes for ages
* This was requested by @awesomekling yesterday, for LibWeb/LibJS sandboxing
* SystemServer was always meant to implement this one day when the need arises; apparently that day was yesterday

As to utility, it's twofold:
* You don't have to run a server to just listen for connections — this is like exit on idle, on steroids. But this also means separate instance of the service are, well, separate, and cannot directly communicate, so it's not a good fit for *every* service. But I imagine some services, like perhaps ProtocolServer or NotificationServer, could use this.
* It should, generally, be easier to use an already accepted socket than to listen on a listening socket. But we have enough utilities in place (`Core::LocalServer`) that this doesn't matter.

alimpfard:
Hmmm...it seems I missed quite a bit of stuff yesterday.

That makes sense, I can see how it can be useful for on-demand isolated servers.
Although I fear for the day someone decides to initialise a billion things on start and then finds this, but that's probably an unfounded fear.

Thanks for explaining this~

awesomekling:
This is great, thank you @bugaevc! :)

PR #2535, opened 2020-06-09
Ports: Use keyserver.ubuntu.com as .sig keyserver.

nico:
Increases the number of successfully building ports from
27 to 36 (of 56) on my system.

Before: http://codepad.org/zsED2a52
Now:    http://codepad.org/ms2meH42

See also https://dev.gnupg.org/T4604#127788

(But, as usual, no idea if this is the officially sanctioned way of doing things. The change feels a bit weird, but so does the error message.)

### DIFF
--- a/Ports/.port_include.sh
+++ b/Ports/.port_include.sh
@@ -52,7 +52,9 @@ func_defined post_fetch || post_fetch() {
 fetch() {
     if [ "$auth_type" == "sig" ] && [ ! -z "${auth_import_key}" ]; then
         # import gpg key if not existing locally
-        gpg --list-keys $auth_import_key || gpg --recv-key $auth_import_key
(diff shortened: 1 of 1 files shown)


nico:
I'm using gpg (GnuPG) 2.2.20.

awesomekling:
I have no idea if this is a good or bad fix. I guess if it works better, that makes it a better fix than nothing.
We should probably have a flag to skip these checks.

nico:
Thanks!

Yes, since all the code is downloaded over https I'm not sure how much the sig checking really helps. I guess the scenario is if you locally install the mitmproxy CA, forget about it, and then get https proxied?

Maybe it could even be an opt-in flag instead of an opt-out flag.

PR #2536, opened 2020-06-09
Move character map from Kernel to LibKeyboard

asliturk:


Kernel/Process.cpp
```diff
@@ -4316,8 +4319,15 @@ int Process::sys$setkeymap(const Syscall::SC_setkeymap_params* user_params)
     if (!validate_read(altgr_map, 0x80))
         return -EFAULT;
 
+    for (int i = 0; i < 0x80; i++) {
+        character_map_data.map[i] = params.map[i];
+        character_map_data.shift_map[i] = params.shift_map[i];
+        character_map_data.alt_map[i] = params.alt_map[i];
+        character_map_data.altgr_map[i] = params.altgr_map[i];
```
awesomekling, line 4326: Since we already copy out `params.map` to `map` (and friends), it seems like we should use `map` here too.

Libraries/LibKeyboard/CharacterMap.cpp
```diff
@@ -26,22 +26,65 @@
...
 {
-    Syscall::SC_setkeymap_params params { m_character_map.map, m_character_map.shift_map, m_character_map.alt_map, m_character_map.altgr_map };
+    Syscall::SC_setkeymap_params params { m_character_map_data.map, m_character_map_data.shift_map, m_character_map_data.alt_map, m_character_map_data.altgr_map };
     return syscall(SC_setkeymap, &params);
 }
 
+char CharacterMap::get_char(KeyEvent event)
+{
+    auto modifiers = event.modifiers();
+    auto index = event.scancode & 0xFF; // Remove E0 flah from scancode
```
awesomekling, line 57: This comment is confusing, since `& 0xff` does not remove `0xe0` from something.
The bits being removed are `0xe000`. Also it's not really a flag, more like a scancode prefix.

asliturk: Index is last byte of scan code, e.g. for `0xE01D` index `0x1D`.

PR #2538, opened 2020-06-10
LibJS: Consolidate error messages into ErrorTypes.h

### LINKED ISSUE #1842: LibJS: Put error messages in a central place?
This is by no means urgent and more of a question for now, but I'm interested to see what others think - I believe having error messages for LibJS in a central place would help a lot with consistency (and avoiding duplicates).

E.g. this is how Mozilla's SpiderMonkey is doing it:

https://dxr.mozilla.org/mozilla-central/source/js/src/js.msg

And Google's V8:

https://chromium.googlesource.com/v8/v8.git/+/refs/heads/master/src/common/message-template.h

mattco98:
Now, exceptions can be thrown with `interpreter.throw_exception<T>(ErrorType:Type, "format", "args",
"here")`. Also makes a lot of error messages more consistent (such as capitalization, removing ending punctuation, fixing typos, and improving some of the error messages). 

I was also thinking of integrating the type of error to throw into the macro, so you wouldn't have to specify the generic, but I don't know if any error messages will need to be thrown from multiple error types (I wouldn't think so). Thoughts?

Closes #1842

### DIFF
--- a/Libraries/LibJS/AST.cpp
+++ b/Libraries/LibJS/AST.cpp
@@ -120,15 +120,15 @@ Value CallExpression::execute(Interpreter& interpreter, GlobalObject& global_obj
         auto call_type = is_new_expression() ? "constructor" : "function";
         if (m_callee->is_identifier() || m_callee->is_member_expression()) {
             String expression_string;
-            if (m_callee->is_identifier())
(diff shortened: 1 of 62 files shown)


awesomekling:
Very nice. :)

PR #2540, opened 2020-06-11
Added thin wrappers around pthread to use instead of std classes.

disks86:
Added join method to Thread.
Fixed bug in Thread where ASSERT in destructor would fire because the primary thread had not been notified of a write that occurred on the spawned thread.
Added a small Demo to test threading support.


awesomekling:
cc @bugaevc

Libraries/LibThread/CMakeLists.txt
```diff
@@ -1,7 +1,11 @@
...
-    BackgroundAction.cpp
-    Thread.cpp
-)
+		BackgroundAction.cpp
+		Thread.cpp
+		Mutex.cpp
+		UniqueLock.cpp
+		LockGuard.cpp
+		ConditionVariable.cpp
+		)
```
bugaevc, line 8: Please fix the indentation (here and elsewhere) :slightly_smiling_face:

bugaevc: Also, these file lists are supposed to be sorted

Demos/Threading/main.cpp
```diff
@@ -0,0 +1,91 @@
...
+    printf("Started Wait Thread\n");
+
+    signal_thread->start();
+    printf("Started Signal Thread\n");
+
+    wait_thread->join();
+    printf("Joined Wait Thread \n");
+
+    signal_thread->join();
+    printf("Joined Signal Thread \n");
```
bugaevc, line 88: Please remove the trailing spaces in output.

And shouldn't this go to stderr? And Not Be In This Each Word Is Capitalized For Some Reason Case?

Demos/Threading/main.cpp
```diff
@@ -0,0 +1,91 @@
...
+
+    return 0;
+}
```
bugaevc, line 91: Missing `\n` :slightly_smiling_face:

Demos/Threading/main.cpp
```diff
@@ -0,0 +1,91 @@
...
+
+int main()
+{
+    /*
+     * Currently there appears to be a bug in pthread that causes only one thread to be notified.
+     * The can be reproduced by adding another wait thread.
+     */
```
bugaevc, line 74: C++ style comments, please

Libraries/LibThread/ConditionVariable.cpp
```diff
@@ -0,0 +1,59 @@
...
+{
+    ASSERT(!pthread_cond_init(&m_condition_variable, NULL));
+}
+
+LibThread::ConditionVariable::~ConditionVariable()
+{
+    ASSERT(!pthread_cond_destroy(&m_condition_variable));
+}
+
+void LibThread::ConditionVariable::notify_one() noexcept
```
bugaevc, line 41: Why are these functions marked `noexcept`, but other aren't?

Libraries/LibThread/ConditionVariable.cpp
```diff
@@ -0,0 +1,59 @@
...
+}
+
+void LibThread::ConditionVariable::notify_all() noexcept
+{
+    ASSERT(!pthread_cond_broadcast(&m_condition_variable));
+}
+
+void LibThread::ConditionVariable::wait(UniqueLock& lock)
+{
+    ASSERT(!pthread_cond_wait(&m_condition_variable, lock.mutex()->native_handle()));
```
bugaevc, line 53: Never put any meaningful logic inside `ASSERT()`, as it can be compiled out in non-debug mode. Instead do this:
```suggestion
    int rc = pthread_cond_wait(&m_condition_variable, lock.mutex()->native_handle());
    ASSERT(rc == 0);
```
(Here and elsewhere.)

Libraries/LibThread/UniqueLock.cpp
```diff
@@ -0,0 +1,69 @@
...
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include "UniqueLock.h"
```
bugaevc, line 27:
```suggestion
#include <LibThread/UniqueLock.h>
```

Libraries/LibThread/UniqueLock.cpp
```diff
@@ -0,0 +1,69 @@
...
+}
+
+void LibThread::UniqueLock::unlock()
+{
+    m_mutex.unlock();
+}
+
+bool LibThread::UniqueLock::owns_lock() const noexcept
+{
+    return m_is_owner;
```
bugaevc, line 58: It's weird to name the field and the getter different...

Libraries/LibThread/UniqueLock.cpp
```diff
@@ -0,0 +1,69 @@
...
+}
+
+LibThread::Mutex* LibThread::UniqueLock::mutex() const noexcept
+{
+    return &m_mutex;
```
bugaevc, line 63: Why return a pointer?

Libraries/LibThread/UniqueLock.h
```diff
@@ -0,0 +1,54 @@
...
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#pragma once
+
+#include <pthread.h>
+
+#include "Mutex.h"
```
bugaevc, line 31: In library headers, it's even more important not to use relative includes:

```suggestion
#include <LibThread/Mutex.h>
```

Libraries/LibThread/UniqueLock.h
```diff
@@ -0,0 +1,54 @@
...
+namespace LibThread {
+
+class UniqueLock {
+
+public:
+    UniqueLock(Mutex& mutex); //TODO: add support for && for ownership transfer
+    ~UniqueLock();
+    void lock();
+    bool try_lock();
+    //TODO: add support for try_lock
```
bugaevc, line 42:
```suggestion
    // TODO: add support for try_lock()
```

Libraries/LibThread/UniqueLock.h
```diff
@@ -0,0 +1,54 @@
...
+    //TODO: add support for try_lock_until
+    void unlock();
+    bool owns_lock() const noexcept;
+    Mutex* mutex() const noexcept;
+    explicit operator bool() const noexcept;
+
+private:
+    LibThread::Mutex& m_mutex;
+    bool m_is_owner = false;
```
bugaevc, line 51:
```suggestion
    bool m_is_owner { false };
```

Libraries/LibThread/UniqueLock.cpp
```diff
@@ -0,0 +1,69 @@
...
+
+LibThread::UniqueLock::~UniqueLock()
+{
+    if (m_is_owner)
+        pthread_mutex_destroy(m_mutex.native_handle());
+}
+
+void LibThread::UniqueLock::lock()
+{
+    m_mutex.lock();
```
bugaevc, line 43: Most of these methods look short enough to be inline; why aren't they?

Libraries/LibThread/Thread.h
```diff
@@ -41,12 +42,14 @@ class Thread final : public Core::Object {
     virtual ~Thread();
 
     void start();
-    void quit(void *code = 0);
+    void quit(void* code = 0);
```
bugaevc, line 45: It's not cool to use 0 as a pointer, not even by C standards :slightly_smiling_face:

Libraries/LibThread/Thread.h
```diff
@@ -41,12 +42,14 @@ class Thread final : public Core::Object {
...
+    void join();
 
 private:
     Function<int()> m_action;
     pthread_t m_tid;
     String m_thread_name;
+    AK::Atomic<bool> m_is_running;
```
bugaevc, line 52:
```suggestion
    Atomic<bool> m_is_running;
```

One of the SerenityOS idiosyncrasies is that in AK, things are declared inside the `namespace AK`, but usually imported back into the public namespace, and then the rest of the code basically never uses the `AK::` prefix.

Libraries/LibThread/ConditionVariable.cpp
disks86: They are drop in replacements for the std functions so I defined them the same.

Libraries/LibThread/UniqueLock.cpp
disks86: Convenience honestly basically everything that uses the mutex wants a pointer. The spec just says it's a native handle so I went with what the pthread calls use.

Libraries/LibThread/Thread.h
disks86: Fair enough that is how it is defined but I can change it to nullptr it will resolve to the same thing so it shouldn't break compatibility.

Demos/Threading/main.cpp
disks86: It's not an error it's just informational to prove it works. I just looked at what some other code in the tree was doing but I don't have a strong preference either way.

Demos/Threading/main.cpp
disks86: What do you mean missing an \n are you asking for an additional blank line after the last message?

Demos/Threading/main.cpp
bugaevc: stderr is not just for errors, it's for any sort of logs and debug output that isn't the primary output you run the program to produce (that's what stdout is for). A more appropriate name for it would be stdlog

That being said, I'm not sure about whether this can be considered the primary output or logs in this case.

Demos/Threading/main.cpp
bugaevc: I mean a missing newline character at the end of the file. Git and GitHub should warn you about it if you look at the diff

Libraries/LibThread/ConditionVariable.cpp
bugaevc: But why do we need drop-in replacements for std functions?

Libraries/LibThread/UniqueLock.cpp
bugaevc: Wait, what spec are we talking about? LibThread is not adhering to any spec, it's our own library of threading utilities

Libraries/LibThread/Thread.h
bugaevc: Again, compatibility with what? This is new code

Demos/Threading/CMakeLists.txt
```diff
@@ -1,6 +1,6 @@
 set(SOURCES
-		main.cpp
-		)
+	main.cpp
```
bugaevc, line 2:
```suggestion
    main.cpp
```

Demos/Threading/main.cpp
```diff
@@ -68,24 +68,22 @@ int signals()
 
 int main()
 {
-    /*
-     * Currently there appears to be a bug in pthread that causes only one thread to be notified.
-     * The can be reproduced by adding another wait thread.
-     */
+    //Currently there appears to be a bug in pthread that causes only one thread to be notified.
```
bugaevc, line 71:
```suggestion
    // Currently there appears to be a bug in pthread that causes only one thread to be notified.
```

Libraries/LibThread/ConditionVariable.cpp
```diff
@@ -26,34 +26,39 @@
...
 #include "assert.h"
 
-#include "ConditionVariable.h"
+#include <LibThread/ConditionVariable.h>
 
 LibThread::ConditionVariable::ConditionVariable()
 {
-    ASSERT(!pthread_cond_init(&m_condition_variable, NULL));
+    auto rc = pthread_cond_init(&m_condition_variable, NULL);
+    ASSERT(!rc);
```
bugaevc, line 34:
```suggestion
    ASSERT(rc == 0);
```

Libraries/LibThread/Mutex.cpp
```diff
@@ -26,34 +26,35 @@
 
 #include "assert.h"
```
bugaevc, line 27:
```suggestion
#include <assert.h>
```

This is a LibC header...

awesomekling:

PR #2541, opened 2020-06-11
Switch from character to code point in KeyEvent(s) to support non-English characterset to type from keyboard

asliturk:
This PR replace character from key events with code point. And adds support to map keys to code point in keymap files ( in LibKeyboard).

In current state we are able to write custom characters in Turkish.

**Known bug:** If OS start in text mode, custom chars in keyboard produce '?' character.

Screenshot from 2020-06-11 11-48-43


bugaevc:
😍, but please fix the commit messages

asliturk:
Hi @bugaevc ,
I fixed commit messages. Could you give a recommendation to solve the problem in text mode?

bugaevc:
Ah, that is simple: the kernel text mode console doesn't know how to display those characters is all.

PR #2545, opened 2020-06-11
LibJS+AK: Add JSON object, AK JsonParser improvements

mattco98:
Adds `JSON.{stringify,parse}`, including all of their weird parameters.

AK `JsonParser` improvements:
- Parsing invalid JSON no longer asserts. Instead of asserting when coming across malformed JSON, `JsonParser::parse` now takes a `bool& ok` parameter that can be checked after parsing the JSON. If not passed, it defaults to the old behavior of asserting.
- Disallow trailing commas in JSON objects and arrays
- No longer parse `undefined`, as that is a purely JS thing
- No longer allow non-whitespace after anything consumed by the initial `parse()` call. Examples of things that were valid and no longer are:
    - `6dfz`
    - `{"foo": 1}abcd`
    - `[1,2,3]4`
- `JsonObject.for_each_member` now iterates in original insertion order

### DIFF
--- a/AK/JsonObject.h
+++ b/AK/JsonObject.h
@@ -41,26 +41,32 @@ public:
     ~JsonObject() {}
 
     JsonObject(const JsonObject& other)
-        : m_members(other.m_members)
+        : m_order(other.m_order)
+        , m_members(other.m_members)
     {
     }
 
     JsonObject(JsonObject&& other)
-        : m_members(move(other.m_members))
+        : m_order(move(other.m_order))
+        , m_members(move(other.m_members))
     {
     }
 
     JsonObject& operator=(const JsonObject& other)
     {
-        if (this != &other)
+        if (this != &other) {
             m_members = other.m_members;
+            m_order = other.m_order;
+        }
         return *this;
     }
 
     JsonObject& operator=(JsonObject&& other)
     {
-        if (this != &other)
+        if (this != &other) {
             m_members = move(other.m_members);
+            m_order = move(other.m_order);
+        }
         return *this;
     }
 
@@ -70,7 +76,7 @@ public:
     JsonValue get(const String& key) const
     {
         auto* value = get_ptr(key);
-        return value ? *value : JsonValue(JsonValue::Type::Undefined);
+        return value ? *value : JsonValue(JsonValue::Type::Null);
     }
 
     JsonValue get_or(const String& key, JsonValue alternative) const
@@ -94,14 +100,17 @@ public:
 
     void set(const String& key, JsonValue value)
     {
+        m_order.append(key);
         m_members.set(key, move(value));
     }
 
     template<typename Callback>
     void for_each_member(Callback callback) const
     {
-        for (auto& it : m_members)
-            callback(it.key, it.value);
+        for (size_t i = 0; i < m_order.size(); ++i) {
+            auto property = m_order[i];
+            callback(property, m_members.get(property).value());
+        }
     }
 
     template<typename Builder>
@@ -113,6 +122,7 @@ public:
     String to_string() const { return serialized<StringBuilder>(); }
 
 private:
+    Vector<String> m_order;
     HashMap<String, JsonValue> m_members;
 };
 
@@ -192,9 +202,6 @@ inline void JsonValue::serialize(Builder& builder) const
         break;
     case Type::UnsignedInt64:
         builder.appendf("%llu", as_u64());
-        break;
-    case Type::Undefined:
-        builder.append("undefined");
         break;
     case Type::Null:
         builder.append("null");
--- a/AK/JsonParser.cpp
+++ b/AK/JsonParser.cpp
@@ -62,15 +62,16 @@ void JsonParser::consume_whitespace()
     consume_while([](char ch) { return is_whitespace(ch); });
 }
 
-void JsonParser::consume_specific(char expected_ch)
+bool JsonParser::consume_specific(char expected_ch)
 {
     char consumed_ch = consume();
-    ASSERT(consumed_ch == expected_ch);
+    return consumed_ch == expected_ch;
 }
 
 String JsonParser::consume_quoted_string()
 {
-    consume_specific('"');
+    if (!consume_specific('"'))
+        return {};
     Vector<char, 1024> buffer;
 
     for (;;) {
@@ -136,7 +137,8 @@ String JsonParser::consume_quoted_string()
             break;
         }
     }
-    consume_specific('"');
+    if (!consume_specific('"'))
+        return {};
 
     if (buffer.is_empty())
         return String::empty();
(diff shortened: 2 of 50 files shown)


Libraries/LibJS/Runtime/JSONObject.cpp
```diff
@@ -0,0 +1,484 @@
...
+            return value.to_string(interpreter);
+        return "null";
+    } else if (value.is_object() && !value.is_function()) {
+        if (value.is_array())
+            return serialize_json_array(interpreter, state, static_cast<Array&>(value.as_object()));
+        return serialize_json_object(interpreter, state,  value.as_object());
+    } else if (value.is_bigint()) {
+        interpreter.throw_exception<TypeError>(ErrorType::JsonBigInt);
+    }
+    return {};
```
Dexesttp, line 179: While the "return {};" here works, I think a "return {};" after the BigInt and an ASSERT_NOT_REACHED(); at the end would be clearer (unless it passes through the if/else in other handled cases ?)

Libraries/LibJS/Runtime/Uint8ClampedArray.h
```diff
@@ -39,15 +39,14 @@ class Uint8ClampedArray final : public Object {
 
     i32 length() const { return m_length; }
 
-    virtual bool put_by_index(u32 property_index, Value value) override;
+    virtual bool put_by_index(u32 property_index, Value value, bool) override;
```
Dexesttp, line 42: It's a bit unclear what the "bool" part is here (even though it comes from the override).
```suggestion
    virtual bool put_by_index(u32 property_index, Value value, bool throw_exceptions) override;
```

Libraries/LibJS/Runtime/JSONObject.cpp
Dexesttp: Hmm. Rereading the code, it seems like there is values not explicitely handled which should be passing through and just be silently swallowed (namely, functions and symbols). So the `ASSERT_NOT_REACHED();` wouldn't make sense, but adding an extra `return {};` after `interpreter.throw_exception()` could still make it less surprising / more future-proof.

mattco98: Yes, this is intentional; Symbols and Functions reach the end of the function, as they are ignored

mattco98:
Yes, the `stringify` method takes two extra arguments (`space` and `replacer`) that control additional aspects of the output. With that in mind, it makes more sense to go directly from objects to strings, rather than having an intermediate `JsonValue` step.

Libraries/LibJS/Runtime/Uint8ClampedArray.h
mattco98: Well my thinking was that because it was ignored in the implementation of the function, it should be "ignored" in the declaration of the function too. But I can still add it if you think it makes more sense

Libraries/LibJS/Runtime/JSONObject.cpp
mattco98: Ehh, not sure what use an additional `return {}` would have when the very next statement after the if statement is a `return {}`.

Dexesttp: It's mostly for readability sake, at this point, so feel free to let everything as-is. I just had to think about it for a bit while reading this piece of code, and having an additional useless `return {};` would have made the overall behaviour more obvious ^^

AK/JsonParser.cpp
```diff
@@ -310,17 +350,33 @@ JsonValue JsonParser::parse()
...
+
+JsonValue JsonParser::parse()
+{
+    bool ok = true;
+    auto result = parse(ok);
+    ASSERT(ok);
+    return result;
+}
+
+JsonValue JsonParser::parse(bool& ok) {
```
awesomekling, line 374: Please use `clang-format`.
Also please make this kind of API return `Optional<JsonValue>` instead of using an out-parameter. :)

AK/JsonObject.h
```diff
@@ -113,6 +122,7 @@ class JsonObject {
     String to_string() const { return serialized<StringBuilder>(); }
 
 private:
+    HashMap<u32, String> m_order;
```
awesomekling, line 125: Hmm, in the absence of something like an `OrderedHashMap` wouldn't this make more sense as a `Vector`?

AK/JsonParser.cpp
```diff
@@ -62,15 +62,16 @@ void JsonParser::consume_whitespace()
...
-void JsonParser::consume_specific(char expected_ch)
+bool JsonParser::consume_specific(char expected_ch)
 {
     char consumed_ch = consume();
-    ASSERT(consumed_ch == expected_ch);
+    return consumed_ch == expected_ch;
 }
 
-String JsonParser::consume_quoted_string()
+Optional<String> JsonParser::consume_quoted_string()
```
awesomekling, line 71: Since `String` already has a null state, we tend not to wrap it in `Optional`.

Libraries/LibJS/Runtime/JSONObject.h
```diff
@@ -52,6 +50,12 @@ class JSONObject final : public Object {
     static String serialize_json_array(Interpreter&, StringifyState&, Object&);
     static String quote_json_string(String);
 
+    // Parse helpers
+    static Object* parse_json_object(Interpreter&, JsonObject);
+    static Array* parse_json_array(Interpreter&, JsonArray);
+    static Value parse_json_value(Interpreter&, JsonValue);
```
awesomekling, line 56: Please pass `JsonFoo` objects by (const) reference as they are deep-copied.

Libraries/LibJS/Runtime/Object.cpp
```diff
@@ -645,21 +645,21 @@ bool Object::put_by_index(u32 property_index, Value value)
         if (interpreter().exception())
             return {};
     }
-    return put_own_property_by_index(*this, property_index, value, default_attributes, PutOwnPropertyMode::Put);
+    return put_own_property_by_index(*this, property_index, value, default_attributes, PutOwnPropertyMode::Put, throw_exceptions);
 }
 
-bool Object::put(PropertyName property_name, Value value)
+bool Object::put(PropertyName property_name, Value value, bool throw_exceptions)
```
awesomekling, line 651: Why is this change needed? Is it effectful? I don't see us branching on this value anywhere.

mattco98: Ahh, looks like the spec actually does a `[[DefineProperty]]` in `InternalizeJSONProperty` instead of a `[[Set]]`, so yeah this isn't needed.

Libraries/LibJS/Runtime/BooleanObject.h
```diff
@@ -43,7 +43,7 @@ class BooleanObject : public Object {
 
 private:
     virtual const char* class_name() const override { return "BooleanObject"; }
-    virtual bool is_boolean() const override { return true; }
+    virtual bool is_boolean_object() const override { return true; }
```
awesomekling, line 46: Good change. That was very ambiguous and would have fooled me.

PR #2547, opened 2020-06-12
RefCounted fixes

bugaevc:
cc @rhin123 — you authored those Calendar changes in https://github.com/SerenityOS/serenity/pull/1721

### DIFF
--- a/AK/Atomic.h
+++ b/AK/Atomic.h
@@ -26,6 +26,7 @@
 
 #pragma once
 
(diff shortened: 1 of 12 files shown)


awesomekling:
That refcount copying bug is ridiculous :D

awesomekling:
Really great improvement. Suspect we'll eat a perf hit from the atomic refcount, but we'll figure something out.

PR #2549, opened 2020-06-12
LibIPC: Use create_single_shot to construct timer

kevmeyer:


awesomekling:
Oh yeah that's way nicer :)

PR #2550, opened 2020-06-12
 LibGUI: Add new LinkLabel component 

asliturk:
Screenshot from 2020-06-12 19-49-00

### DIFF
--- a/Libraries/LibGUI/AboutDialog.cpp
+++ b/Libraries/LibGUI/AboutDialog.cpp
@@ -30,6 +30,7 @@
 #include <LibGUI/BoxLayout.h>
 #include <LibGUI/Button.h>
 #include <LibGUI/Label.h>
+#include <LibGUI/LinkLabel.h>
 #include <LibGUI/Widget.h>
 #include <LibGfx/Font.h>
 
@@ -40,7 +41,7 @@ AboutDialog::AboutDialog(const StringView& name, const Gfx::Bitmap* icon, Window
     , m_name(name)
     , m_icon(icon)
 {
-    resize(413, 205);
+    resize(413, 225);
     set_title(String::format("About %s", m_name.characters()));
     set_resizable(false);
 
@@ -97,6 +98,13 @@ AboutDialog::AboutDialog(const StringView& name, const Gfx::Bitmap* icon, Window
         make_label("SerenityOS");
     make_label(version_string());
     make_label("Copyright \xC2\xA9 the SerenityOS developers, 2018-2020");
+    make_label("Licensed under the BSD 2-Clause License.");
+
+    auto& link_label = right_container.add<LinkLabel>("http://serenityos.org");
+    link_label.set_text_alignment(Gfx::TextAlignment::CenterLeft);
+    link_label.set_size_policy(SizePolicy::Fill, SizePolicy::Fixed);
+    link_label.set_preferred_size(0, 14);
+
 
     right_container.layout()->add_spacer();
 
--- a/Libraries/LibGUI/LinkLabel.cpp
+++ b/Libraries/LibGUI/LinkLabel.cpp
@@ -1,0 +1,89 @@
+/*
+ * Copyright (c) 2020, Hüseyin ASLITÜRK <asliturk@hotmail.com>
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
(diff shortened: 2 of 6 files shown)


bugaevc:
Look, I've been meaning to tell you:

`Component: Subcomponent, change`

is not a proper commit message format, at least we don't do that. Use

`Component: Change something in subcomponent` or perhaps `Component: Make subcomponent to foo` or something like that. In this case,

* Just use `About:` for the prefix in the first commit, cause apps are considered components by themselves
* For the second commit, `LibGUI: Add license info and website link to AboutDialog`

See?

And when in doubt, look at previous commits that modify the same code. (I usually obsess over this sort of thing, so my commits are always an indication of the proper prefix :wink:; @awesomekling was just recently saying how much he cares about commit messages, but he still does it wrong every now and then).

Libraries/LibGUI/LinkLabel.cpp
```diff
@@ -0,0 +1,89 @@
...
+LinkLabel::~LinkLabel()
+{
+}
+
+void LinkLabel::mousedown_event(MouseEvent&)
+{
+    URL url = URL::create_with_url_or_path(m_href);
+    bool result = Desktop::Launcher::open(url);
+    if (!result) {
+        dbg() << String::format("Failed to open '%s'\n", url.path().characters());
```
bugaevc, line 58: No need to use `String::format` (or output the `\n`) when using `dbg()` 🙂

Libraries/LibGUI/LinkLabel.cpp
```diff
@@ -0,0 +1,89 @@
...
+    }
+}
+
+void LinkLabel::enter_event(Core::Event& event)
```
bugaevc, line 62: That... sounds like a bad name for a mouse enter event. @awesomekling, should we change it to `mouseenter_event()`, or do we want to preserve the Qt name?

Libraries/LibGUI/LinkLabel.h
```diff
@@ -0,0 +1,52 @@
...
+protected:
+    explicit LinkLabel(const StringView& text = {});
+
+    virtual void enter_event(Core::Event&) override;
+    virtual void leave_event(Core::Event&) override;
+    virtual void mousedown_event(MouseEvent&) override;
+    virtual void paint_event(PaintEvent&) override;
+
+private:
+    String m_href;
```
bugaevc, line 49: What's the difference between this and the "text"? Aren't they always set to the same value?

Libraries/LibGUI/AboutDialog.cpp
```diff
@@ -97,6 +98,13 @@ AboutDialog::AboutDialog(const StringView& name, const Gfx::Bitmap* icon, Window
         make_label("SerenityOS");
     make_label(version_string());
     make_label("Copyright \xC2\xA9 the SerenityOS developers, 2018-2020");
+    make_label("Licensed under the BSD 2-Clause License.");
+
+    auto& link_label = right_container.add<LinkLabel>("http://serenityos.org");
```
bugaevc, line 103: Here's a little problem: this dialog is not only shown in the About app, it's shown in virtually every app.

Libraries/LibGUI/LinkLabel.h
asliturk: Hi,
In general use, the text is the same as href, so the constructor sets the text to href.
If you want to set a different href than the text, there is the `set_href `method.
```
auto& link_label = right_container.add<LinkLabel>("Web Site");
link_label.set_href("http://serenityos.org");
```

bugaevc: Ah, that makes sense. Perhaps let's add a comment about it?

Libraries/LibGUI/AboutDialog.cpp
asliturk: Hi,
Uppss this is a big problem :( @awesomekling what do you prefer for this.
I prefer drop commits for about dialog but i cant find other use case for this component for now.

awesomekling: Yeah this is a big problem that will eventually require some kind of architectural decision to be made. @bugaevc what do you think we should do here?

bugaevc: (I assume you're talking about the same thing I was — that now basically every app requires access to the launch server.)

That's actually a deeper problem with `pledge()` and `unveil()`: they affect the process as a whole, but you don't control the whole process if you're using some library (and you always do), and that library expects to make use of some standard functionality. The file dialog is another example: every app that wants to display the file dialog now has to pledge thread simply because the dialog uses threading. Also everyone needs Unix and accept and whatnot just because of the RPC server in LibCore.

I honestly don't know how we should solve the problem on the fundamental level. The file picker should be moved out of process; and for the link, perhaps we can only show it in the system about box and not in every app.

asliturk: Can we create WindowServer API for common dialog boxes?
In this approach, all permissions go to WindowServer and we don't need to add any lib to the application.

awesomekling: WindowServer cannot use LibGUI, so it wouldn't be a good place to implement these. But perhaps a DialogServer service that provides AboutDialog, FilePicker, etc..

asliturk: I am trying to implement the DialogServer approach. With this solution, we should add 'pledge("unix")' and 'unveil("/tmp/portal/dialog")' to the application that wants to use a DialogServer.

At this point, we are at the same point as we are trying to avoid.

BenWiederhake:
Parts of this PR are now obsolete, due to 9014d300ed470c3500c26ca3e62173ea816ca110

Issue #2498, opened 2020-06-04
LibGUI: Lots of crashes when triple-clicking and typing in GUI::Text{Box,Editor}


linusg:
E.g. in the browser address bar or TextEditor's main widget or search:

- type some text if empty
- select it all by triple-clicking the text input, Ctrl+A or mouse selection won't reproduce
- type any character

Result: lots of stuff crashing.

```
[Browser(23:23)]: NP(error) fault at invalid address V0x00000000
[Browser(23:23)]: Unrecoverable page fault, read from address V0x00000000
[Browser(23:23)]: Note: Address V0x00000000 looks like a possible nullptr dereference
[Browser(23:23)]: CRASH: Page Fault. Ring 3.
[Browser(23:23)]: exception code: 0004 (isr: 0000
[Browser(23:23)]:   pc=001b:081091d6 flags=0206
[Browser(23:23)]:  stk=0023:01902308
[Browser(23:23)]:   ds=0023 es=0023 fs=0023 gs=07d3
[Browser(23:23)]: eax=00000000 ebx=00000021 ecx=00000000 edx=01970216
[Browser(23:23)]: ebp=01902328 esp=c4392fe8 esi=00000000 edi=01902370
[Browser(23:23)]: cr0=8001001b cr2=00000000 cr3=02664000 cr4=00300ee4
[Browser(23:23)]: code: ff 34 b0 46 ff 75 08 e8
[Browser(23:23)]: 0x081091d6  Gfx::Font::width(AK::Utf32View const&) const +38

[Browser(23:23)]: 0xc015a824  Kernel::Process::crash(int, unsigned int, bool) +906
[Browser(23:23)]: 0xc011facf  illegal_instruction_handler +0
[Browser(23:23)]: 0xc0120dcd  page_fault_handler +836
[Browser(23:23)]: 0xc011e80d  page_fault_asm_entry +22
[Browser(23:23)]: 0x080b0e84  GUI::TextEditor::content_x_for_position(GUI::TextPosition const&) const +234
[Browser(23:23)]: 0x080b1324  GUI::TextEditor::line_content_rect(unsigned long) const +62
[Browser(23:23)]: 0x080b1513  GUI::TextEditor::line_widget_rect(unsigned long) const +25
[Browser(23:23)]: 0x080b19e4  GUI::TextEditor::set_cursor(GUI::TextPosition const&) +140
[...]

PR #2551, opened 2020-06-12
LibGUI: Don't iterate visual lines, if visual data out-of-date

kevmeyer:


Libraries/LibGUI/TextEditor.cpp
```diff
@@ -1460,6 +1462,9 @@ void TextEditor::for_each_visual_line(size_t line_index, Callback callback) cons
     size_t start_of_line = 0;
     size_t visual_line_index = 0;
 
+    if (!is_visual_data_up_to_date())
+        return;
```
awesomekling, line 1466: This doesn't feel like a good fix.
Instead of making `for_each_visual_line()` a no-op while visual line data is invalid, we should be checking for this at the call site instead.
Who is trying to iterate over the visual lines while they are invalid and why? And how can we fix it? :)

FalseHonesty: I was looking into this, and it is what is causing #2498. When typing a letter to replace a selection, the TextDocument first performs a RemoveTextCommand, which causes the TextDocumentLine's codepoints vector to get nulled out (note: only when the entire line is being replaced). This causes the call to `content_x_for_position` when trying to set the cursor to its new position to fail because it directly accesses the Vector's backing data, which is still null.

kevmeyer: @awesomekling as @FalseHonesty this happens through `content_x_for_position`. I added the check there, which is still not ideal, but it seems to do the job.

Libraries/LibGUI/TextEditor.cpp
```diff
@@ -947,6 +947,9 @@ int TextEditor::content_x_for_position(const TextPosition& position) const
     int x_offset = 0;
     switch (m_text_alignment) {
     case Gfx::TextAlignment::CenterLeft:
+        if (!is_visual_data_up_to_date()) {
+            return m_horizontal_content_padding + x_offset;
+        }
```
awesomekling, line 952: This is a hack fix that doesn't get at the root of the problem and instead introduces different behavior in a certain state.

Based on what @FalseHonesty is saying, it sounds like the problem is that we're trying to set the cursor position while the visual line data is invalid. I think a proper fix would be to prevent `set_cursor()` from trying to use the visual lines for anything while they are invalid.

I think we can just skip the whole `if (m_cursor != position) {` block in `set_cursor()` if there's a pending reflow. Does that make sense?

PR #2552, opened 2020-06-13
[WIP] Add base64 encoder

tomleb:
Adds a simple base64 encoder.
This is still a work-in-progress because I want to add support the modified base64 encoding for IMAP protocol.
Maybe the name of the PR is not the best because I've also added test cases for the existing base64 decoder..

### DIFF
--- a/Userland/base64.cpp
+++ b/Userland/base64.cpp
@@ -1,0 +1,80 @@
+/*
+ * Copyright (c) 2020, Tom Lebreux <tomlebreux@hotmail.com>
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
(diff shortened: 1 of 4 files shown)


awesomekling:
Looks fine to me. :)

Userland/base64.cpp
```diff
@@ -0,0 +1,81 @@
...
+    if (pledge("stdio", nullptr) < 0) {
+        perror("pledge");
+        return 1;
+    }
+
+    if (decode) {
+        auto decoded = decode_base64(StringView(buffer));
+        printf("%s\n", String::copy(decoded).characters());
+        return 0;
+    } else {
```
awesomekling, line 76: No need for `else` after `return`.

awesomekling:
I'm gonna merge this despite WIP since it seems good enough on its own and adds a nice utility. :)

PR #2553, opened 2020-06-13
GUI and SystemMonitor improvements

sppmacd:
- [x] Indeterminate progress bars
- [x] SystemMonitor: "Open in FileManager" context menu option in ProcessTableView
- [x] LibCore: ProcessStatisticsReader functions to get CWD and executable path for specific PIDs (from /proc/$pid/cwd and /proc/$pid/exe)
- [x] Userland: readlink command

### DIFF
--- a/Libraries/LibGfx/JPGLoader.cpp
+++ b/Libraries/LibGfx/JPGLoader.cpp
@@ -90,6 +90,7 @@
 #define JPG_EOI 0xFFD9
 #define JPG_RST 0XFFDD
 #define JPG_SOF0 0XFFC0
+#define JPG_SOF2 0xFFC2
 #define JPG_SOI 0XFFD8
 #define JPG_SOS 0XFFDA
 #define JPG_COM 0xFFFE
@@ -438,6 +439,9 @@ static inline bool is_valid_marker(const Marker marker)
     case JPG_SOI:
     case JPG_SOS:
         return true;
+    case JPG_SOF2:
+        dbg() << "Progressive JPEGs are not supported yet. Decoder will fail!";
+        return false;
     default:
         return false;
     }


Userland/readlink.cpp
```diff
@@ -0,0 +1,53 @@
+/*
+ * Copyright (c) 2018-2020, sppmacd <sppmacd@pm.me>
```
bugaevc, line 2: You haven't been working on this in 2018 🙂

Also, use the real name here, perhaps?

Userland/readlink.cpp
```diff
@@ -0,0 +1,53 @@
...
+        return 1;
+    }
+
+    Core::ArgsParser parser;
+    Vector<const char*> paths;
+    parser.add_positional_argument(paths, "Link path", "path");
+    parser.parse(argc, argv);
+
+    for (auto path : paths) {
+        char buffer[PATH_MAX];
```
bugaevc, line 42: :extremely_sad_face:

We should really have a better interface for this (`readlink()`), if we don't have one yet.

Userland/readlink.cpp
```diff
@@ -0,0 +1,53 @@
...
+        int length = readlink(path, buffer, sizeof(buffer));
+        if (length < 0) {
+            perror("readlink");
```
bugaevc, line 45: There can be multiple input paths, let's specify which one the error is about

Applications/SystemMonitor/main.cpp
```diff
@@ -78,7 +83,7 @@ static NonnullRefPtr<GUI::Widget> build_graphs_tab();
 class UnavailableProcessWidget final : public GUI::Frame {
     C_OBJECT(UnavailableProcessWidget)
 public:
-    virtual ~UnavailableProcessWidget() override { }
+    virtual ~UnavailableProcessWidget() override {}
```
bugaevc, line 86: This is a regression

Applications/SystemMonitor/main.cpp
```diff
@@ -220,7 +221,7 @@ int main(int argc, char** argv)
             if (path.has_value())
                 Desktop::Launcher::open(URL::create_with_file_protocol(LexicalPath(path.value()).dirname()));
             else
-                GUI::MessageBox::show_error("Couldn't open path!");
+                GUI::MessageBox::show_error(String::format("Failed to open path: %s", strerror(errno)));
```
bugaevc, line 224: Let's also specify which path was it that we tried and failed to open.

But the way, this clearly doesn't belong in the readlink commit...

Userland/readlink.cpp
```diff
@@ -0,0 +1,54 @@
...
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include <LibCore/ArgsParser.h>
+
+int main(int argc, char* argv[])
+{
+    if (pledge("stdio rpath unix", nullptr) < 0) {
```
bugaevc, line 31: What do we need Unix for?

Libraries/LibCore/ProcessStatisticsReader.cpp
```diff
@@ -108,6 +109,43 @@ HashMap<pid_t, Core::ProcessStatistics> ProcessStatisticsReader::get_all()
...
+    builder.append("/cwd");
+
+    char buffer[PATH_MAX];
+    int length = readlink(builder.to_string().characters(), buffer, sizeof(buffer));
+    if (length < 0) {
+        perror("readlink");
+    } else {
+        ASSERT(length > 0);
+        return String(buffer, length);
+    }
```
awesomekling, line 126: Please add a convenience API like `static String Core::File::readlink(const String& path)` for this so we don't have to see this awkward C business more than necessary. :)

Libraries/LibCore/ProcessStatisticsReader.cpp
```diff
@@ -108,6 +109,43 @@ HashMap<pid_t, Core::ProcessStatistics> ProcessStatisticsReader::get_all()
     return map;
 }
 
+Optional<String> ProcessStatisticsReader::get_current_working_directory(pid_t pid)
```
awesomekling, line 112: `String` already has a null state and we use that instead of wrapping them in `Optional`.

Libraries/LibCore/ProcessStatisticsReader.h
```diff
@@ -85,10 +85,11 @@ struct ProcessStatistics {
 class ProcessStatisticsReader {
 public:
     static HashMap<pid_t, Core::ProcessStatistics> get_all();
+    static Optional<String> get_current_working_directory(pid_t pid);
+    static Optional<String> get_executable_path(pid_t pid);
```
awesomekling, line 89: `String` already has a null state and we use that instead of wrapping them in `Optional`.

Applications/SystemMonitor/main.cpp
```diff
@@ -83,7 +84,7 @@ static NonnullRefPtr<GUI::Widget> build_graphs_tab();
 class UnavailableProcessWidget final : public GUI::Frame {
     C_OBJECT(UnavailableProcessWidget)
 public:
-    virtual ~UnavailableProcessWidget() override {}
+    virtual ~UnavailableProcessWidget() override { }
```
awesomekling: These changes are unrelated to the new `readlink` program and should be squashed into the previous commit.

Userland/readlink.cpp
```diff
@@ -0,0 +1,54 @@
...
+    parser.parse(argc, argv);
+
+    char buffer[PATH_MAX];
+    int length;
+    for (auto path : paths) {
+        length = readlink(path, buffer, sizeof(buffer));
+        if (length < 0) {
+            perror(String::format("readlink: %s", path).characters());
+            return 1;
+        } else {
```
awesomekling, line 48: No need for `else` after `return`.

Userland/readlink.cpp
```diff
@@ -0,0 +1,54 @@
...
+        return 1;
+    }
+
+    Core::ArgsParser parser;
+    Vector<const char*> paths;
+    parser.add_positional_argument(paths, "Link path", "path");
+    parser.parse(argc, argv);
+
+    char buffer[PATH_MAX];
+    int length;
```
awesomekling, line 42: Please declare variables when first using them (or as late as possible.)

Libraries/LibCore/File.cpp
```diff
@@ -129,5 +130,16 @@ String File::real_path_for(const String& filename)
     return real_path;
 }
 
+String File::read_link(const String& filename)
+{
+    char buffer[PATH_MAX];
+    int length = readlink(filename.characters(), buffer, sizeof(buffer));
+    if (length <= 0) {
+        return {};
```
awesomekling, line 138: Is a zero-length `readlink()` result actually a failure? A null `String` result should indicate failure.
Note that `String()` and `String("")` are not the same thing. The former is null, the latter is not.

Libraries/LibCore/File.cpp
```diff
@@ -129,5 +130,16 @@ String File::real_path_for(const String& filename)
...
+    } else {
```
awesomekling, line 139: No need for `else` after `return`.

Libraries/LibGUI/FileSystemModel.cpp
```diff
@@ -73,13 +74,11 @@ bool FileSystemModel::Node::fetch_data(const String& full_path, bool is_root)
     mtime = st.st_mtime;
 
     if (S_ISLNK(mode)) {
-        char buffer[PATH_MAX];
-        int length = readlink(full_path.characters(), buffer, sizeof(buffer));
-        if (length < 0) {
+        String link = Core::File::read_link(full_path.characters());
+        if (link.is_null()) {
             perror("readlink");
```
awesomekling, line 79: In general, we can't rely on `errno` still having a relevant value like this, since any number of things may overwrite it (`malloc` is a common one.)
Perhaps it's still fine in this case, but I would prefer that we not use things like `errno` or `perror()` except directly after calls to LibC that communicate errors via `errno`.
If you want to log the error from `readlink()`, I think it would be better to do that inside `Core::File::read_link()`.

Userland/readlink.cpp
```diff
@@ -38,17 +39,13 @@ int main(int argc, char* argv[])
...
 
-    char buffer[PATH_MAX];
-    int length;
     for (auto path : paths) {
-        length = readlink(path, buffer, sizeof(buffer));
-        if (length < 0) {
-            perror(String::format("readlink: %s", path).characters());
+        String link = Core::File::read_link(path);
+        if (link.is_null()) {
+            perror(path);
```
awesomekling, line 45: Same comment here about `errno`.

Libraries/LibCore/ProcessStatisticsReader.cpp
```diff
@@ -103,47 +103,31 @@ HashMap<pid_t, Core::ProcessStatistics> ProcessStatisticsReader::get_all()
 
         // and synthetic data last
         process.username = username_from_uid(process.uid);
+        process.working_directory = get_current_working_directory(process.pid);
+        process.executable_path = get_executable_path(process.pid);
```
awesomekling, line 107: This seems like an aggressive amount of extra work to do for clients that don't need the information.
The change in resource graph applet pledges below are necessary because you now visit every accessible `/proc/PID` directory and read some links.

The resource graphs are updated every second, so this is a whole bunch of extra I/O for no reason in those processes.

To avoid this, how about we have `working_directory()` and `executable_path()` functions instead, and we can fetch the information for clients who actually need it?

Applications/SystemMonitor/ProcessModel.cpp
```diff
@@ -123,6 +123,10 @@ String ProcessModel::column_name(int column) const
         return "Pledge";
     case Column::Veil:
         return "Veil";
+    case Column::WorkingDirectory:
+        return "PWD";
+    case Column::ExecutablePath:
+        return "EXE";
```
awesomekling, line 129: `EXE` is not an acronym. Let's call this `Executable`.

Applications/SystemMonitor/ProcessModel.cpp
```diff
@@ -123,6 +123,10 @@ String ProcessModel::column_name(int column) const
         return "Pledge";
     case Column::Veil:
         return "Veil";
+    case Column::WorkingDirectory:
+        return "PWD";
```
awesomekling, line 127: The commit message calls this `CWD`, and here you call it `WorkingDirectory` but also `PWD` (?)
Let's settle on a single name. Some options:
* Directory
* WorkingDirectory
* CWD

Libraries/LibCore/ProcessStatisticsReader.cpp
sppmacd: The resource graphs don't need a majority of ```ProcessStatisticsReader::get_all()``` data, only memory and CPU stats.
Why load and parse a lot of data only to result in 2 numbers? Maybe ```ProcessStatisticsReader::get_performance_stats()``` or something like this?

Libraries/LibCore/File.cpp
sppmacd: True, ```length``` is not possible to be 0 (can symlink point to zero-length file path?)

bugaevc: Yes it can

Applications/SystemMonitor/ProcessModel.cpp
sppmacd: It should be CWD.

Libraries/LibCore/File.cpp
sppmacd: ok

PR #2554, opened 2020-06-13
LibGfx: PNG improvements

MegabytePhreak:
I was inspired after watching the ACID2 video on Youtube!
Implement interlaced PNG support. 
Fix an bug in 16-bit RGB PNG decoding I found along the way
Add some interlaced PNG examples for the browser


awesomekling:
Looks nice and clean! Thank you for implementing this :)

PR #2555, opened 2020-06-14
LibWeb: Implement HTML caption, column group and table insertion modes.

Lubrsi:
Implements the "in caption", "in column group" and "in table" insertion modes in the HTML parser.

Also fixes some little mistakes in the "in body" insertion mode that I found whilst cross-referencing.

### DIFF
--- a/Libraries/LibWeb/Parser/HTMLDocumentParser.cpp
+++ b/Libraries/LibWeb/Parser/HTMLDocumentParser.cpp
@@ -968,7 +968,7 @@ void HTMLDocumentParser::handle_in_body(HTMLToken& token)
     }
 
     if (token.is_start_tag() && token.tag_name() == HTML::TagNames::form) {
-        if (m_form_element && m_stack_of_open_elements.contains(HTML::TagNames::template_)) {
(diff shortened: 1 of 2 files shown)


Libraries/LibWeb/Parser/HTMLDocumentParser.cpp
```diff
@@ -1798,7 +1805,53 @@ void HTMLDocumentParser::handle_in_table(HTMLToken& token)
...
+        if (type_attribute.is_null() || !type_attribute.equals_ignoring_case("hidden")) {
+            goto AnythingElse;
+        }
+
+        PARSE_ERROR();
+        insert_html_element(token);
+
+        // FIXME: Is this the correct interpretation of "Pop that input element off the stack of open elements."?
+        //        Because this wording is the first time it's seen in the spec.
+        //        Other times it's worded as: "Immediately pop the current node off the stack of open elements."
```
awesomekling, line 1828: That's how I interpret it FWIW. But yeah, it's annoying the way the spec sometimes subtly changes the wording like this. Makes you feel like they're implying something different.

PR #2556, opened 2020-06-14
LibWeb: Implement all comment, DOCTYPE and CDATA tokenizer states

Lubrsi:
Also fixes TagOpen having a seperate emit and reconsume in ANYTHING_ELSE.

Even though we haven't implemented any switches to the CDATA states yet,
we may as well have them ready for when we do implement the switches.


awesomekling:
:+1:

PR #2557, opened 2020-06-15
  Move the image file detection code to the File class

asliturk:


Dexesttp:
It's way better than the way it was done before ! (side note, you probably want to pass a `const StringView&` instead of a `String&` as argument. That would simplify things ^^)

However, the list of allowed extensions / formats is still is duplicated between your new `LibCore/File.cpp#is_image_file(String& path)` method and the [`LibGfx/Bitmap.cpp#load_from_file(const StringView& path)`](https://github.com/SerenityOS/serenity/blob/master/Libraries/LibGfx/Bitmap.cpp#L87) method. And they have to be kept in sync for most of the logic to work properly.

While it seems kind of weird to put another file-extension-detecting method in Bitmap, maybe moving the method to `LibGfx::Bitmap::is_path_a_supported_image_format(const StringView&)` (or something similar) would mitigate that ?
This would also allow us to use a `#define ENUMERATE_IMAGE_FORMATS __ENUMERATE_IMAGE_FORMAT(load_gif, .gif)`-like macro for matching the file formats when parsing and reading (which would make keeping both of them in sync very easy).

alimpfard:
`LibMagic` time?

bugaevc:
> LibCore: File, Add new is_image_file method

Please see https://github.com/SerenityOS/serenity/pull/2550#issuecomment-643572519 :slightly_smiling_face: — the commit message should be "LibCore: Add a new `File::is_image_file()` method"

Libraries/LibCore/File.h
```diff
@@ -38,6 +38,15 @@ class File final : public IODevice {
     virtual ~File() override;
 
     static Result<NonnullRefPtr<File>, String> open(const String& filename, IODevice::OpenMode, mode_t = 0644);
+    static bool is_image_file(const String& path)
```
bugaevc, line 41:
```suggestion
    static bool is_image_file(const StringView& path)
```

asliturk:
Hi @Dexesttp,

It is very good recommendation. I moved the new method to the `Gfx:.Bitmap` class as you suggested.

PR #2558, opened 2020-06-15
QuickShow: Small improvements

asliturk:
- Switch to full screen mode on double click
- Add reset zoom action to toolbar with new 1:1 icon


Applications/QuickShow/QSWidget.cpp
```diff
@@ -247,6 +253,21 @@ void QSWidget::load_from_file(const String& path)
...
+
+    win->set_fullscreen(!win->is_fullscreen());
+
+    for (auto& child : win->main_widget()->children()) {
+        if (child.name() == "toolbar_container") {
+            auto& toolbar = static_cast<GUI::ToolBarContainer&>(child);
+            toolbar.set_visible(!win->is_fullscreen());
+            break;
+        }
+    }
```
awesomekling, line 271: Instead of doing this in `QSWidget`, let's move it to `main()` which is really the "owner" of the window.
I would suggest something like this:

```cpp
class QSWidget {
    Function<void()> on_doubleclick;
};

void QSWidget::doubleclick_event(GUI::MouseEvent&)
{
    on_doubleclick();
}

int main()
{
    ...
    qswidget->on_doubleclick = [&] {
        window->set_fullscreen(!window->is_fullscreen());
    };
}

```

awesomekling:
Merged manually as 016fc9c13327af4b0010cf65daf343235e48d8bc, bf3d98012ab9641c5935fda6aca0e176ac0e92c6, ea1ebe866299ef75376dc79c9406faa35c431d24

PR #2561, opened 2020-06-16
LibC: Add posix_spawn().

nico:
All the file actions stuff is still missing for now,
as is POSIX_SPAWN_SETSCHEDULER (not sure what that's
supposed to do) and POSIX_SPAWN_RESETIDS.

Implemented in userspace for now. Once there are users,
it'll likely make sense to make this a syscall for
performance reasons.

A simple test program of the form

    extern char **environ;
    int main(int argc, char* argv[])
    {
        pid_t pid;
        char* args[] = { "ls", NULL };
        posix_spawnp(&pid, "ls", nullptr, nullptr, args, environ);
    }

works fine.

### DIFF
--- a/Libraries/LibC/spawn.cpp
+++ b/Libraries/LibC/spawn.cpp
@@ -1,0 +1,239 @@
+/*
+ * Copyright (c) 2020, Nico Weber <thakis@chromium.org>
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+/* posix_spawn and friends
+ *
+ * values from POSIX standard unix specification
+ *
+ * https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/spawn.h.html
+ */
+
+#include <spawn.h>
+
+#include <assert.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <unistd.h>
+
+extern "C" {
+
+[[noreturn]] static void posix_spawn_child(const char* path, const posix_spawn_file_actions_t* file_actions, const posix_spawnattr_t* attr, char* const argv[], char* const envp[], int (*exec)(const char*, char* const[], char* const[]))
+{
+    if (file_actions) {
+        // FIXME
+    }
+    if (attr) {
+        short flags = attr->flags;
+        if (flags & POSIX_SPAWN_SETPGROUP) {
+            if (setpgid(0, attr->pgroup) < 0) {
+                perror("posix_spawn setpgid");
+                exit(127);
+            }
+        }
+        if (flags & POSIX_SPAWN_SETSCHEDPARAM) {
+            if (sched_setparam(0, &attr->schedparam) < 0) {
+                perror("posix_spawn sched_setparam");
+                exit(127);
+            }
+        }
+        if (flags & POSIX_SPAWN_SETSIGDEF) {
+            struct sigaction default_action;
+            default_action.sa_flags = 0;
+            sigemptyset(&default_action.sa_mask);
+            default_action.sa_handler = SIG_DFL;
+
+            sigset_t sigdefault = attr->sigdefault;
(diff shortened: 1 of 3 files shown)


Libraries/LibC/spawn.h
```diff
@@ -0,0 +1,118 @@
...
+struct posix_spawnattr_t {
+    short flags;
+    pid_t pgroup;
+    sched_param schedparam;
+    int schedpolicy;
+    sigset_t sigdefault;
+    sigset_t sigmask;
+};
+
+#ifndef RESTRICT
```
awesomekling, line 65: Maybe this should go in sys/cdefs.h?

Libraries/LibC/spawn.cpp
```diff
@@ -0,0 +1,250 @@
...
+}
+
+int posix_spawnattr_setpgroup(posix_spawnattr_t* attr, pid_t pgroup)
+{
+    attr->pgroup = pgroup;
+    return 0;
+}
+
+int posix_spawnattr_setschedparam(posix_spawnattr_t* RESTRICT attr,
+    const struct sched_param* RESTRICT schedparam)
```
awesomekling, line 224: Pls turn off "wrap all lines at 80 columns" or whatever causes this :)

bugaevc:
Please don't end the commit message with a period

Libraries/LibC/spawn.h
bugaevc: Do we even need this `restrict`? We don't use `restrict` elsewhere where POSIX specifies it, e.g. [`stat`(2)](https://pubs.opengroup.org/onlinepubs/009695399/functions/stat.html)

Libraries/LibC/spawn.cpp
```diff
@@ -0,0 +1,250 @@
...
+        if (flags & POSIX_SPAWN_SETSIGMASK)
+            if (sigprocmask(SIG_SETMASK, &attr->sigmask, nullptr) != 0)
+                exit(127);
+        // FIXME: POSIX_SPAWN_RESETIDS, POSIX_SPAWN_SETSCHEDULER
+    }
+
+    if (search_path)
+        execvpe(path, argv, envp);
+    execve(path, argv, envp);
+    __assertion_failed("returned from exec");
```
bugaevc, line 85: * We don't call `__assertion_failed()` directly, we use `ASSERT_NOT_REACHED()` or `ASSERT(something)`
* In this case, exec clearly can return due to an error, so just asserting like that is not appropriate. Perhaps something like this:
   ```cpp
   perror("exec");
   exit(1);
   ```
    would be more appropriate. It'd be even better to include the path we attempted to exec in the error message.

bugaevc:
General note: I'd rather see (parts of) spawn functionality implemented in kernelspace, to be able to return meaningful errors to the caller. (There's a well-known pipe trick for achieving this in userspace though.)

Libraries/LibC/spawn.cpp
nico: I copied it over from  https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/spawn.h.html and ran clang-format. Am I supposed to join the whole file into a single line before running clang-format?

(Will do a best guess at what you want, but not 100% sure.)

nico:
> General note: I'd rather see (parts of) spawn functionality implemented in kernelspace, to be able to return meaningful errors to the caller. (There's a well-known pipe trick for achieving this in userspace though.)

Yes, I'd like to implement it in the kernel. The idea is:

1. Implement in user space.
2. Port a program that uses it.
3. Move to kernel.

That way, each step is a bit smaller.

Libraries/LibC/spawn.h
nico: `restrict` is never necessary as far as I know (prototypes that differ only in `restrict` or no get merged without warning). It's only there to help the compiler generate code. I can remove it.

bugaevc:
I don't think kernel-side implementation would be much more complex, so I'd go for it directly.

> Port a program that uses it

Do you have some specific port that needs `posix_spawn()` in mind?

nico:
> Please don't end the commit message with a period

Ok!

nico:
> I don't think kernel-side implementation would be much more complex, so I'd go for it directly.
> 
> > Port a program that uses it
> 
> Do you have some specific port that needs `posix_spawn()` in mind?

Yes, I'd like to port ninja. When I moved that to posix_spawn, that was a big runtime improvement on some OSs in some cases.

Libraries/LibC/spawn.cpp
nico: Good point, thanks. Added a few perror()s for now.

I also looked over the style guide again and saw the (good) suggestion to prefer enums over boolean parameters so I changed that, but then I figured I can just pass in the right exec directly. Let me know if you think that's overly cutesy, then I'll make it an enum insteaad.

Libraries/LibC/spawn.cpp
```diff
@@ -0,0 +1,233 @@
...
+
+[[noreturn]] static void posix_spawn_child(const char* path, const posix_spawn_file_actions_t* file_actions, const posix_spawnattr_t* attr, char* const argv[], char* const envp[], int (*exec)(const char*, char* const[], char* const[]))
+{
+    if (file_actions) {
+        // FIXME
+    }
+    if (attr) {
+        short flags = attr->flags;
+        if (flags & POSIX_SPAWN_SETPGROUP)
+            if (setpgid(0, attr->pgroup) < 0) {
```
bugaevc, line 52: We always put braces around the outer `if` body in cases like this :slightly_smiling_face:

Libraries/LibC/spawn.h
```diff
@@ -0,0 +1,91 @@
...
+struct posix_spawnattr_t {
+    short flags;
+    pid_t pgroup;
+    sched_param schedparam;
+    int schedpolicy;
+    sigset_t sigdefault;
+    sigset_t sigmask;
+};
+
+int posix_spawn(pid_t*, const char*, const posix_spawn_file_actions_t*, const posix_spawnattr_t*, char* const[], char* const[]);
```
bugaevc, line 65: I don't think it's going to work to use bare `posix_spawnattr_t` (not `struct posix_spawnattr_t`) in pure C.

You should do
```c
typedef struct {
    ...
} posix_spawnattr_t;
```

instead. The same goes for `posix_spawn_file_actions_t`.

Libraries/LibC/spawn.h
```diff
@@ -0,0 +1,94 @@
...
+
+#include <sched.h>
+#include <signal.h>
+#include <sys/cdefs.h>
+#include <sys/types.h>
+
+__BEGIN_DECLS
+
+enum {
+    POSIX_SPAWN_RESETIDS = 1 << 0,
```
bugaevc, line 45: As much as I like `enum`s, I believe by "symbolic constants", POSIX means `#define`s — those are `#ifdef`able, unlike `enum`s.

nico: What most of the headers do is enums, and then a few headers has `#define FOO FOO` for the enum entries so that they're `#ifdef`able. From what I can tell, only enums that user code tests for has the defines.

(Happy to change it, but it's this way for consistency with existing code.)

bugaevc:
```
$ find Libraries/LibC -name \*.h | xargs grep enum | wc -l
5
$ find Libraries/LibC -name \*.h | xargs grep define | wc -l
821
```

I'd say those few `enum`s in LibC headers should be replaced with `#define`s, especially those that are then `#define`d back to their values.

awesomekling: Yeah, we should just switch everything to `#define` instead of using enums.
The enums have bit me in the ass more than once due to ports checking `#ifdef something_we_had_as_an_enum`

PR #2562, opened 2020-06-16
Docs: Add comment about required cmake version

projectgoav:
Toolchain/BuildIt.sh failed to complete on fresh clone expecting a cmake version of 3.16. My machine had 3.13 installed which was the latest that `apt` would provide. 

Added note in build instructions for future reference.


awesomekling:
Thanks!

bugaevc:
Nitpick: it's not apt that provides CMake, it's your distro's repositories. Also, CMake (the project) is spelled "CMake", `cmake` lowercase is just the command. Please fix it :slightly_smiling_face:

PR #2564, opened 2020-06-16
readlink, goodboys and meow

bugaevc:
Yay for some less `PATH_MAX`s!

cc @balatt for wc(1) changes, cc @alimpfard for meowing and whatnot :smile:

### DIFF
--- a/Userland/cat.cpp
+++ b/Userland/cat.cpp
@@ -25,13 +25,30 @@
  */
 
 #include <AK/Vector.h>
+#include <LibCore/ArgsParser.h>
 #include <assert.h>
 #include <errno.h>
 #include <fcntl.h>
 #include <stdio.h>
 #include <stdlib.h>
+#include <sys/stat.h>
 #include <string.h>
 #include <unistd.h>
+
+bool detect_useless_use_of_cat(int argc)
+{
+    // I can think of somewhat valid uses of zero-argument cat
+    // from a tty to a pipe, so let's allow those.
+    if (argc != 2)
+        return false;
+
+    struct stat statbuf;
+    int rc = fstat(1, &statbuf);
+    if (rc < 0)
+        return false;
+
+    return S_ISFIFO(statbuf.st_mode);
+}
 
 int main(int argc, char** argv)
 {
@@ -40,12 +57,23 @@ int main(int argc, char** argv)
         return 1;
     }
 
+    Vector<const char*> paths;
+
+    Core::ArgsParser args_parser;
+    args_parser.add_positional_argument(paths, "File path", "path", Core::ArgsParser::Required::No);
+    args_parser.parse(argc, argv);
+
+    if (detect_useless_use_of_cat(argc))
+        fprintf(stderr, "\033[34;1mMeow! Useless use of cat detected\033[0m\n");
+
     Vector<int> fds;
-    if (argc > 1) {
-        for (int i = 1; i < argc; i++) {
+    if (!paths.is_empty()) {
+        for (const char* path : paths) {
             int fd;
-            if ((fd = open(argv[i], O_RDONLY)) == -1) {
-                fprintf(stderr, "Failed to open %s: %s\n", argv[i], strerror(errno));
+            if (StringView { path } == "-") {
+                fd = 0;
+            } else if ((fd = open(path, O_RDONLY)) == -1) {
+                fprintf(stderr, "Failed to open %s: %s\n", path, strerror(errno));
                 continue;
             }
             fds.append(fd);
@@ -69,14 +97,18 @@ int main(int argc, char** argv)
                 perror("read");
                 return 2;
             }
-            ssize_t nwritten = write(1, buf, nread);
-            if (nwritten < 0) {
-                perror("write");
-                return 3;
+            size_t already_written = 0;
+            while (already_written < (size_t)nread) {
+                ssize_t nwritten = write(1, buf + already_written, nread - already_written);
+                if (nwritten < 0) {
+                    perror("write");
+                    return 3;
+                }
+                already_written += nwritten;
             }
-            ASSERT(nwritten == nread);
         }
         close(fd);
     }
+
     return 0;
 }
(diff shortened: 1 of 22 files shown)


Userland/cat.cpp
```diff
@@ -25,27 +25,55 @@
...
 #include <stdio.h>
 #include <stdlib.h>
+#include <sys/stat.h>
 #include <string.h>
 #include <unistd.h>
 
+bool detect_useless_use_of_cat(int argc)
+{
+    // I can think of somewhat valid uses of zero-argument cat
+    // from a tty to a pipe, so let's allow those.
```
alimpfard, line 41: Poifect:
```c++
 if (command.argv.is_empty() && !(command.rewirings.is_empty() && command.redirections.is_empty())) {
        // Set the argv to 'cat' when some redirections exist
        // This allows `<foo' to work.
        command.argv.append("cat");
    }
```

Userland/cat.cpp
```diff
@@ -25,27 +25,55 @@
...
 {
     if (pledge("stdio rpath", nullptr) < 0) {
         perror("pledge");
         return 1;
     }
 
+    Vector<const char*> paths;
+
+    Core::ArgsParser args_parser;
+    args_parser.add_positional_argument(paths, "File path", "path", Core::ArgsParser::Required::No);
```
alimpfard, line 63: File path _s_?

Userland/cat.cpp
bugaevc: I would argue `<foo` shouldn't cat `foo` to stdout, it should `close(open("foo", O_RDONLY))`

alimpfard: Uhhh...what's the point of that?
`<foo >bar` (as done by bash and ash) takes the stuff from foo and dumps it into bar.
I kinda like that tbh

Userland/cat.cpp
bugaevc: That's questionable — it looks like this, for some context:
image
and I would say `path` is *one* path which can be repeated (but most of the time it's used with a single path). This is also what other users of `Core::ArgsParser` seem to do in similar cases.

alimpfard: Oh, it puts dots after it.
My bad. 'path' makes more sense.

Userland/cat.cpp
bugaevc: Pretty sure Bash doesn't do that and behaves as I described. It's not very *useful* (although it is nice to be able to create files by just saying `> foo`), but it's consistent:
* An empty command (empty argv) does nothing
* Redirections open (and close, and dup) files
* Redirections on an empty command just open files

Bash (and [POSIX shell](https://pubs.opengroup.org/onlinepubs/009695399/utilities/exec.html)) also has "exec", which is basically a way to tell it not to fork before executing the command. If you leave out the command, it doesn't actually do the `exec*()`, but it still performs redirections, which is a (somewhat) convenient way to open and close files in the shell process itself:

```
$ exec 3>/tmp/foo
$ echo hi >&3
$ exec 3>&-
```

alimpfard: Hmmm...yeah, only zsh does that...weird.

what the hell is `>&-`?

bugaevc: "close this fd"

alimpfard: wait it keeps the filedesricptors open?
that's _weird_

alimpfard: me no want that weirdness.
I can make it do nothing except resolve redirections if no argv, but you will not convince me to do _that_. redirection targets live until the command ends, otherwise it's just...hard to reason about...

alimpfard: weird tho, does it leak all those fds when it forks?
how does it manage those, I wonder

bugaevc: No, you don't understand. `exec` is what makes it *not fork*. Normally, you'd use it like that:

`exec some-command args 2> /dev/null`

This would make the shell execute the command as usual, except it won't fork, so basically the shell itself applies redirections and exec's in the shell process; after the command exits the process is gone.

But if you specify an empty command, it doesn't actually exec, so you keep the shell — but it performs redirections as usual, but since they happen in the Shell process, not in the fork, they persist.

bugaevc: What I'm trying to say, if we take a step back, is that these parts — fork, exec/argv, file redirections/operations — are orthogonal in Bash, are controlled independently (can do file operations without exec or without fork or without either, can do exec without fork or without redirections...), and behave consistently no matter which combination you use.

alimpfard: Yeah, allowing the user to mess with the descriptors of the shell itself is pretty silly.
we use 5 fds already, `3` would just overwrite some event loop thing

bugaevc: No, we must absolutely support that — maybe not with that exact syntax if that's up for debate, but the general idea is that the shell is just a convenient way to *script Unix stuff* and that chiefly means forking, execing, and messing with file descriptors, which the shell is just a convenient (ugh that word again) scripting syntax for.

As for the event loop, yeah, we're going to have to make it use some high fd numbers (bash uses fds above 1024 I believe)

alimpfard: I mean...if you're this persistent about it...sure, I'll let the user shoot themselves in the foot.
it feels pretty gross to me, but...whatever, sure.

_can_ we change the event loop fds?

alimpfard: worst of both worlds: `$ 3>foo` applies the redirection to the shell (it's otherwise a meaningless operation anyway)

Userland/cat.cpp
```diff
@@ -25,27 +25,55 @@
...
+    args_parser.parse(argc, argv);
+
+    if (detect_useless_use_of_cat(argc))
+        fprintf(stderr, "\033[34;1mMeow! Useless use of cat detected\033[0m\n");
```
awesomekling, line 67: This seems like an arbitrary thing to condescend to the user about?

alimpfard: but...isn't it an adorable thing to do?
and it's harmless anyway?

bugaevc: It's not arbitrary, it's such [a part of the Unix culture](http://porkmail.org/era/unix/award.html)...

bugaevc: Perhaps it's worth tweaking the message to make it sound as a friendly suggestion rather than condescending?

First attempt:

"Meow! Hey, you can simply use file redirection (< `argv[1]`) instead of spawning a cat process"

awesomekling: Oh, I was not aware of this old Unix meme :)
With that in mind, I think it makes perfect sense to word it like you did.

PR #2565, opened 2020-06-17
LibGfx: Add support for .ico format images

MegabytePhreak:
This adds support for images in .ico format to LibGfx. The main use case is for favicons.
Two types are supported: uncompressed 32-bit  and PNG encoded. There are potentially some other uncompressed formats out there, but this is enough to load the favicons for Hacker News, Google, GitHub and Twitter. 

Since .ico files can contain multiple images and LibGfx/ImageDecoder doesn't have a way to expose this, for now just find the largest image in the file and load that one. This give the best chance for good results since it limits the need for upscaling.

### DIFF
--- a/Libraries/LibGfx/ICOLoader.cpp
+++ b/Libraries/LibGfx/ICOLoader.cpp
@@ -113,6 +113,7 @@ struct ICOLoadingContext {
     const u8* data { nullptr };
     size_t data_size { 0 };
     Vector<ImageDescriptor> images;
+    size_t largest_index;
 };
 
 RefPtr<Gfx::Bitmap> load_ico(const StringView& path)
@@ -164,6 +165,21 @@ static Optional<ImageDescriptor> decode_ico_direntry(BufferStream& stream)
     return { desc };
 }
 
+static size_t find_largest_image(const ICOLoadingContext& context)
+{
+    size_t max_area = 0;
+    size_t index = 0;
+    size_t largest_index = 0;
+    for(const auto& desc : context.images) {
+        if (desc.width * desc.height > max_area) {
+            max_area = desc.width * desc.height;
+            largest_index = index;
+        }
+        ++index;
+    }
+    return largest_index;
+}
+
 static bool load_ico_directory(ICOLoadingContext& context)
 {
     auto buffer = ByteBuffer::wrap(context.data, context.data_size);
@@ -197,6 +213,7 @@ static bool load_ico_directory(ICOLoadingContext& context)
 #endif
         context.images.append(desc);
     }
+    context.largest_index = find_largest_image(context);
     context.state = ICOLoadingContext::State::DirectoryDecoded;
     return true;
 }
@@ -292,7 +309,7 @@ static bool load_ico_bmp(ICOLoadingContext& context, ImageDescriptor& desc)
     return true;
 }
 
-static bool load_ico_bitmap(ICOLoadingContext& context, size_t index)
+static bool load_ico_bitmap(ICOLoadingContext& context, Optional<size_t> index)
 {
     if (context.state < ICOLoadingContext::State::DirectoryDecoded) {
         if (!load_ico_directory(context)) {
@@ -301,18 +318,21 @@ static bool load_ico_bitmap(ICOLoadingContext& context, size_t index)
         }
         context.state = ICOLoadingContext::State::DirectoryDecoded;
     }
-    if (index >= context.images.size()) {
+    size_t real_index = context.largest_index;
+    if (index.has_value())
+        real_index = index.value();
+    if (real_index >= context.images.size()) {
         return false;
     }
 
-    ImageDescriptor& desc = context.images[index];
+    ImageDescriptor& desc = context.images[real_index];
 
     PNGImageDecoderPlugin png_decoder(context.data + desc.offset, desc.size);
     if (png_decoder.sniff()) {
         desc.bitmap = png_decoder.bitmap();
         if (!desc.bitmap) {
 #ifdef ICO_DEBUG
-            printf("load_ico_bitmap: failed to load PNG encoded image index: %lu\n", index);
+            printf("load_ico_bitmap: failed to load PNG encoded image index: %lu\n", real_index);
 #endif
             return false;
         }
@@ -320,7 +340,7 @@ static bool load_ico_bitmap(ICOLoadingContext& context, size_t index)
     } else {
         if (!load_ico_bmp(context, desc)) {
 #ifdef ICO_DEBUG
-            printf("load_ico_bitmap: failed to load BMP encoded image index: %lu\n", index);
+            printf("load_ico_bitmap: failed to load BMP encoded image index: %lu\n", real_index);
 #endif
             return false;
         }
@@ -351,7 +371,7 @@ IntSize ICOImageDecoderPlugin::size()
         m_context->state = ICOLoadingContext::State::DirectoryDecoded;
     }
 
-    return { m_context->images[0].width, m_context->images[0].height };
+    return { m_context->images[m_context->largest_index].width, m_context->images[m_context->largest_index].height };
 }
 
 RefPtr<Gfx::Bitmap> ICOImageDecoderPlugin::bitmap()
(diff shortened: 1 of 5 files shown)


Libraries/LibGfx/BMPStructures.h
```diff
@@ -0,0 +1,67 @@
...
+
+#pragma once
+
+#include <AK/Types.h>
+
+
+namespace Gfx {
+
+// FIXME: This is in little-endian order. Maybe need a NetworkOrdered<T> equivalent
+struct [[gnu::packed]]  BMPFILEHEADER {
```
nico, line 35: clang-format this

Libraries/LibGfx/BMPStructures.h
```diff
@@ -0,0 +1,67 @@
...
+    u32 vres;
+    u32 hres;
+    u32 palette_size;
+    u32 important_colors;
+};
+static_assert(sizeof(BITMAPINFOHEADER) == 40);
+
+}
+
+struct [[gnu::packed]]  XRGB {
```
nico, line 61: same

Libraries/LibGfx/ICOLoader.cpp
```diff
@@ -0,0 +1,387 @@
...
+
+static Optional<ImageDescriptor> decode_ico_direntry(BufferStream& stream)
+{
+    ICONDIRENTRY entry;
+    stream.read_raw((uint8_t*)&entry, sizeof(entry));
+    if (stream.handle_read_failure())
+        return {};
+
+    ImageDescriptor desc = { entry.width, entry.height, entry.offset, entry.size, nullptr };
+    if (desc.width == 0)
```
nico, line 126: Out of interest, which web sites have favicons with invalid dimensions?

Libraries/LibGfx/ICOLoader.cpp
```diff
@@ -0,0 +1,387 @@
...
+            printf("load_ico_directory: offset: %lu size: %lu doesn't fit in ICO size: %lu\n",
+                desc.offset, desc.size, context.data_size);
+#endif
+            return false;
+        }
+#ifdef ICO_DEBUG
+        printf("load_ico_directory: index %lu width: %u height: %u offset: %lu size: %lu\n",
+            i, desc.width, desc.height, desc.offset, desc.size);
+#endif
+        if (desc.width * desc.height > max_area) {
```
nico, line 167: Personally, I find it a bit surprising that finding the largest-area image is mixed with decoding. I'd probably do this elsewhere.

Libraries/LibGfx/ICOLoader.cpp
```diff
@@ -0,0 +1,387 @@
...
+
+    if (info.planes != 1) {
+#ifdef ICO_DEBUG
+        printf("load_ico_bmp: planes: %d != 1", info.planes);
+#endif
+        return false;
+    }
+
+    if (info.bpp != 32) {
+        dbg() << "load_ico_bmp: unsupported bpp " << info.bpp;
```
nico, line 211: missing `return false;`?

Libraries/LibGfx/ICOLoader.cpp
```diff
@@ -0,0 +1,387 @@
...
+    if (info.width != desc.width || info.height != 2 * desc.height) {
+#ifdef ICO_DEBUG
+        printf("load_ico_bmp: size mismatch: ico %dx%d, bmp %dx%d\n",
+            desc.width, desc.height, info.width, info.height);
+#endif
+        return false;
+    }
+
+    // Mask is 1bpp, and each row must be 4-byte aligned
+    size_t mask_row_len = ((((desc.width + 7) / 8) + 3) / 4) * 4;
```
nico, line 236: I'd put a `size_t alignTo(size_t i, size_t a) { return ((i + a - 1) / a) * a; }` somewhere (AK?) and write this as `alignTo((desc.width + 7) / 8, 4)`. Something like `alignTo` can be used in many places and makes things slightly easier to read.

Libraries/LibGfx/ICOLoader.cpp
MegabytePhreak: This is actually just how the ICO format specifies a dimension of 256, since the width/height fields are only 8 bits wide. This is see on Hacker News for example.

MegabytePhreak:
> Cool stuff! I've actually been working on a full BMPLoader, which currently has support for most of the images [on this BMP test suite](https://entropymine.com/jason/bmpsuite/bmpsuite/html/bmpsuite.html). I need to do a lot of cleanup, but I hope to get it PR'd by today or tomorrow. I'd love to help integrate it into this loader after I open the initial PR :)

I actually looked at trying to write a BMP parser that would be usable to help with ICO decoding, but a lot of the implementation details would need to be exposed between them, since the ICO format uses the BMP structures oddly. In particular the BMP is declared as twice the height of the ICO puts a 1bpp transparency mask after the main image. Still, we might be able to at least share the data structures.

Libraries/LibGfx/ICOLoader.cpp
MegabytePhreak: Fixed

mattco98:
Interesting. I didn't see that part, only the part where the BMP header (or header + DIB, can't remember) isn't included in the ICO file. But I didn't dig into ICO too much and just jumped straight into BMP, so I'd have to do some more reading to see if it'd be possible to share implementations.

Libraries/LibGfx/ICOLoader.cpp
```diff
@@ -0,0 +1,402 @@
...
+
+    return { desc };
+}
+
+static size_t find_largest_image(ICOLoadingContext& context)
+{
+    size_t max_area = 0;
+    size_t index = 0;
+    size_t largest_index = 0;
+    for(auto desc : context.images) {
```
nico, line 140: `const auto&`?

MegabytePhreak: Yep. I guess my c++ is a bit rusty.

Libraries/LibGfx/ICOLoader.cpp
```diff
@@ -0,0 +1,286 @@
...
+        auto maybe_desc = decode_ico_direntry(stream);
+        if (!maybe_desc.has_value()) {
+#ifdef ICO_DEBUG
+            printf("load_ico_directory: error loading entry: %lu\n", i);
+#endif
+            return false;
+        }
+
+        auto& desc = maybe_desc.value();
+        // FIXME: Overflow
```
awesomekling, line 169: Please fix :)

MegabytePhreak: Should be fixed now.

Libraries/LibGfx/ICOLoader.cpp
```diff
@@ -328,16 +347,16 @@ RefPtr<Gfx::Bitmap> ICOImageDecoderPlugin::bitmap()
 
     if (m_context->state < ICOLoadingContext::State::BitmapDecoded) {
         // NOTE: This forces the chunk decoding to happen.
-        bool success = load_ico_bitmap(*m_context, 0);
+        bool success = load_ico_bitmap(*m_context, -1);
```
awesomekling: Please use `Optional` instead of magic numbers like `-1`

Libraries/LibGfx/ICOLoader.cpp
```diff
@@ -279,15 +298,15 @@ static bool load_ico_bitmap(ICOLoadingContext& context, size_t index)
         desc.bitmap = png_decoder.bitmap();
         if (!desc.bitmap) {
 #ifdef ICO_DEBUG
-            printf("load_ico_bitmap: failed to load PNG endcoded image index: %lu\n", index);
+            printf("load_ico_bitmap: failed to load PNG encoded image index: %d\n", index);
```
awesomekling: Please squash fixes like this into the previous commit instead of having them here. :)

Libraries/LibGfx/ICOLoader.cpp
```diff
@@ -131,6 +132,21 @@ static Optional<ImageDescriptor> decode_ico_direntry(BufferStream& stream)
     return { desc };
 }
 
+static size_t find_largest_image(ICOLoadingContext& context)
```
awesomekling, line 135: Seems like we can take a `const ICOLoadingContext&` here.

PR #2566, opened 2020-06-17
Docs: Spell CMake correctly in BuildInstructions

projectgoav:
cc @bugaevc (#2562)

Re: apt providing CMake. I'm not quite sure a better wording for this part. From my limited understanding, each package manage has its own set of repositories that supplement those from your distro? So while `apt` might provide something a version 1, `snap` might provide the same package at version 1.5.

### DIFF
--- a/Documentation/BuildInstructions.md
+++ b/Documentation/BuildInstructions.md
@@ -27,7 +27,7 @@ sudo apt-get install gcc-9 g++-9
 sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-9 900 --slave /usr/bin/g++ g++ /usr/bin/g++-9
 ```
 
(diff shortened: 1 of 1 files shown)


bugaevc:
> I'm not quite sure a better wording for this part.

"If your system doesn't provide a recent enough version of CMake, you can download a binary release from ..."

> each package manage has its own set of repositories that supplement those from your distro?

No, package managers and repos are (to some extent) orthogonal. You can use the same package manager (say `yum`) with very different repos depending on your distro and which repos you have enabled.

In case of Ubuntu, both the dpkg and snap repos are provided by Canonical, but they're only compatible with their respective package managers (apt, snap), and may indeed provide different versions of software.

projectgoav:
Thanks very much for your explaination :) I've updated the wording.

PR #2568, opened 2020-06-17
Kernel: clang-format ENUMERATE_SYSCALLS

nico:
Alternatively, we could surround this with `// clang-format off` and `// clang-format on` if you don't want all these lines to change every time the length of the longest syscall changes.

If you like being able clang-format these but don't like what it does to `git blame`, we could add a `.git-blame-ignore-revs ` file containing revs with changes like this and use git 2.23's `git config blame.ignoreRevsFile .git-blame-ignore-revs`.


awesomekling:
That's fine :)

PR #2570, opened 2020-06-17
LibC / Kernel: Implement seteuid() and friends!

nico:
Add seteuid()/setegid() under _POSIX_SAVED_IDS semantics,
which also requires adding suid and sgid to Process, and
changing setuid()/setgid() to honor these semantics.

The exact semantics aren't specified by POSIX and differ
between different Unix implementations. This patch makes
serenity follow FreeBSD. The 2002 USENIX paper
"Setuid Demystified" explains the differences well.

In addition to seteuid() and setegid() this also adds
setresuid()/setresgid(), and the accessors getresuid()/getresgid().

Also reorder uid/euid functions so that they are the
same order everywhere (namely, the order that
geteuid()/getuid() already have).

### DIFF
--- a/Kernel/Process.cpp
+++ b/Kernel/Process.cpp
@@ -986,9 +986,9 @@ int Process::do_exec(NonnullRefPtr<FileDescription> main_program_description, Ve
 
     if (!(main_program_description->custody()->mount_flags() & MS_NOSUID)) {
         if (main_program_metadata.is_setuid())
-            m_euid = main_program_metadata.uid;
+            m_euid = m_suid = main_program_metadata.uid;
         if (main_program_metadata.is_setgid())
-            m_egid = main_program_metadata.gid;
+            m_egid = m_sgid = main_program_metadata.gid;
     }
 
     Thread::current->set_default_signal_dispositions();
@@ -1386,10 +1386,12 @@ Process* Process::create_kernel_process(Thread*& first_thread, String&& name, vo
 Process::Process(Thread*& first_thread, const String& name, uid_t uid, gid_t gid, pid_t ppid, RingLevel ring, RefPtr<Custody> cwd, RefPtr<Custody> executable, TTY* tty, Process* fork_parent)
     : m_name(move(name))
     , m_pid(allocate_pid())
-    , m_uid(uid)
-    , m_gid(gid)
     , m_euid(uid)
     , m_egid(gid)
+    , m_uid(uid)
+    , m_gid(gid)
+    , m_suid(uid)
+    , m_sgid(gid)
     , m_ring(ring)
     , m_executable(move(executable))
     , m_cwd(move(cwd))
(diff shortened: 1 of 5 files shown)


awesomekling:
cc @bugaevc

bugaevc:
Commit messages: the way we do it when mentioning multiple subsystems is with a plus & no spaces around it, e.g. `Kernel+LibC: Implement ...` (oh and yeah, the kernel usually comes first if that's where the main change is).

I would prefer if FreeBSD, Linux, and Solaris weren't mentioned by name in the comments (so many times, too!), it just feels wrong. "Some other systems also allow that and that, but we don't because <reason>" sounds fine on the other hand.

`sys$setregid()` is missing a `REQUIRE_PROMISE`

Kernel/Process.cpp
```diff
@@ -2125,23 +2127,105 @@ int Process::sys$killpg(int pgrp, int signum)
...
+    REQUIRE_PROMISE(id);
+
+    if (ruid == (uid_t)-1)
+        ruid = m_uid;
+    if (euid == (uid_t)-1)
+        euid = m_euid;
+    if (suid == (uid_t)-1)
+        suid = m_suid;
+
+    auto ok = [&](uid_t id) { return id == m_uid || id == m_euid || id == m_suid; };
```
bugaevc, line 2201:
```suggestion
    auto ok = [this](uid_t id) { return id == m_uid || id == m_euid || id == m_suid; };
```
(same in other lambdas)

bugaevc:
A man page for this would be nice, but that might be too much to ask :slightly_smiling_face:

Other than that, "this looks perfect as it is" :smile:

nico:
Thanks for taking a look, and for catching the missing required promise.

I'll write man pages in a follow-up (for posix_spawn too), and I'll remove all the references to Linux/FreeBSD then. Until then, the comments hold state I'd like to keep somewhere until I've written it into the man pages if that's ok :)

Kernel/Process.cpp
```diff
@@ -2384,6 +2510,24 @@ pid_t Process::sys$getppid()
     return m_ppid;
 }
 
+int Process::sys$getresuid(uid_t* ruid, uid_t* euid, uid_t* suid)
+{
+    REQUIRE_PROMISE(stdio);
+    *ruid = m_uid;
+    *euid = m_euid;
```
bugaevc, line 2475: Here and below: you can't just dereference user pointers like that 😀

You have to `validate_write_typed()`, `return EFAULT`, and `copy_to_user`. Otherwise this is an arbitrary kernel write primitive 🙂

bugaevc:
One more thing I haven't noticed yesterday...

And yeah, it's fine to mention other systems by name in man page — in the history section and maybe in other places. See [`pledge`(2)](https://github.com/SerenityOS/serenity/blob/master/Base/usr/share/man/man2/pledge.md) for an example

nico:
Thanks for catching that, makes sense :)

nico:
On that note, do we put code we use for testing new functions somewhere? Is there a place for LibC demo programs?

bugaevc:
There are some `test...` programs in userland (e.g. https://github.com/SerenityOS/serenity/blob/master/Userland/test_io.cpp), and there are AK tests for AK functionality.

awesomekling:
> There are some `test...` programs in userland (e.g. https://github.com/SerenityOS/serenity/blob/master/Userland/test_io.cpp), and there are AK tests for AK functionality.

This is an area where we could improve dramatically :)

bugaevc:
> I'll write man pages in a follow-up (for posix_spawn too), and I'll remove all the references to Linux/FreeBSD then.

A gentle reminder that this still needs to be done :slightly_smiling_face:

PR #2572, opened 2020-06-17
AK+Optional: Add `is_empty()`

devsh0:
It's frustrating to type ahead and later realize that you forgot the '!' in the front :^)


awesomekling:
Let's only have one way of doing this. The ambiguity cost doesn't seem worth it. :)

PR #2574, opened 2020-06-18
Use curl instead of wget for Toolchain/BuildIt*.

ericonr:
- For Linux: curl is already listed as a dependency;
- For macOS: curl is pre-installed;
- For OpenBSD and FreeBSD: curl is a dependecy of git.


awesomekling:
:+1:

0xtrzy:
what using curl, wget on serenity?
how I can put my program when i cant use ssh or network transport?

awesomekling:
@0xtrzy Please stop spamming the bug tracker. This is not your personal support chat.

PR #2579, opened 2020-06-18
Piano: Multi-track functionality

willmcpherson2:


awesomekling:
All right, this is a nice direction to move in :)

Some notes for future improvements:

Thinking about how a typical DAW works, we should try to separate out the "GUI piano" from the track/audio classes. The "current octave" is really just a GUI concern, and the GUI should turn key events into MIDI-like note inputs for the currently selected track.

The "current track" should be a strictly GUI concept as well.

We want to move towards a graph-based architecture where we can chain sound generators, FX, etc, in a node graph. The final output would be generated by a mixer node that takes N inputs and mixes them together with per-input gain etc etc, the usual stuff.

Disclaimer: I have never done any of these things before so I'm probably not saying it right :)

willmcpherson2:
@awesomekling 

>The "current octave" is really just a GUI concern

>The "current track" should be a strictly GUI concept

>graph-based architecture where we can chain sound generators, FX, etc, in a node graph

>The final output would be generated by a mixer node that takes N inputs and mixes them together

Definitely all of these. At the moment I'm just moving forward and not thinking about design. The last two points are pretty essential when we start doing more serious stuff. The way `TrackManager` "mixes" tracks at the moment is pretty dumb.

Also those first two points are pretty obvious. Myopia on my part

PR #2581, opened 2020-06-18
QuickShow: Small Improvements (vol:2)

asliturk:
- if image is larger then window size resize it to fit the image 
- "Set as desktop wallpaper" menu item

### DIFF
--- a/Applications/QuickShow/QSWidget.cpp
+++ b/Applications/QuickShow/QSWidget.cpp
@@ -48,7 +48,7 @@ void QSWidget::clear()
     m_bitmap = nullptr;
     m_path = {};
 
-    on_scale_change(100);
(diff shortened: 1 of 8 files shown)


Sobak:
@asliturk absolute nitpicking but your last commit message says:

> QuickShow: Resize app window if image is larger then

You probably meant something along the lines of "(...) than a screen size".

asliturk:
> @asliturk absolute nitpicking but your last commit message says:
> 
> > QuickShow: Resize app window if image is larger then
> 
> You probably meant something along the lines of "(...) than a screen size".

I have to change my writing commit message in the last second behavior. :)

Sobak:
I fully get what are you talking about :) BTW: "than", not "then"

PR #2584, opened 2020-06-18
WavLoader: Search for DATA marker by reading single bytes

tlmrgvf:
When loading some files, the loader would just keep searching for the DATA marker, never finding it.
This seemed to happen especially when the files were big or had comments

### DIFF
--- a/Libraries/LibAudio/WavLoader.cpp
+++ b/Libraries/LibAudio/WavLoader.cpp
@@ -155,16 +155,27 @@ bool WavLoader::parse_header()
     // Read chunks until we find DATA
     bool found_data = false;
     u32 data_sz = 0;
+    u8 search_byte = 0;
     while (true) {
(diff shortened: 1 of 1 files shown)


Libraries/LibAudio/WavLoader.cpp
```diff
@@ -155,16 +155,27 @@ bool WavLoader::parse_header()
...
     bool found_data = false;
     u32 data_sz = 0;
+    u8 search_byte = 0;
     while (true) {
-        u32 chunk_id;
-        stream >> chunk_id;
-        CHECK_OK("Reading chunk ID searching for data");
+        stream >> search_byte;
+        CHECK_OK("Reading byte searching for data");
+        if(search_byte != 0x64) //D
```
awesomekling, line 162: Please use `clang-format`.

PR #2585, opened 2020-06-18
LibGfx: Add BMP loader

mattco98:
Adds a BMP loader to process .bmp files.

Features:
- All header formats are supported
- Full RLE4/8/24 support
- Color scaling (e.g. distributing a 5-bit color throughout the 8-bit range, so 5-bit white is `0xffffff` instead of `0xf8f8f8`)
- Full BITMASK/ALPHABITMASK support

This loader was tested with the images at https://entropymine.com/jason/bmpsuite/bmpsuite/html/bmpsuite.html. This loader correctly renders 81 out of the 90 total images (for reference, Firefox displays 64 correctly).

I'm quite happy with the result of this loader, but I'm not sure how good the code itself is. This is _really_ out of my comfort zone, but I decided to go for it anyways, and I learned a lot while doing it. I expect this PR to be WIP for a while, so please let me know all of the ways I can improve my code! :)

Also, while this PR is WIP, I've included a commit with the entire bmp testsuite included in `/res/html/misc`, but this is super overkill, so I'll remove it before the PR gets merged.


Libraries/LibGfx/BMPLoader.h
```diff
@@ -0,0 +1,195 @@
...
+#include <LibGfx/Bitmap.h>
+#include <LibGfx/ImageDecoder.h>
+
+#define BMP_DEBUG 0
+
+namespace Gfx {
+
+RefPtr<Gfx::Bitmap> load_bmp(const StringView& path);
+
+const u8 bmp_header_size = 14;
```
MegabytePhreak, line 39: All these internal details should probably go in BMPLoader.cpp instead of the header. 
I think the only things that should be in the header are the load_bmp function and the BMPImageDecoderPlugin class.

Libraries/LibGfx/BMPLoader.h
```diff
@@ -0,0 +1,195 @@
...
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#pragma once
+
+#include <AK/Vector.h>
+#include <LibGfx/Bitmap.h>
+#include <LibGfx/ImageDecoder.h>
+
+#define BMP_DEBUG 0
```
MegabytePhreak, line 33: This should probably go in the cpp instead of the header.

Libraries/LibGfx/BMPLoader.cpp
```diff
@@ -0,0 +1,1178 @@
...
+{
+    return out << "(" << ep.x << ", " << ep.y << ", " << ep.z << ")";
+}
+
+const LogStream& operator<<(const LogStream& out, Endpoint<u32> ep)
+{
+    return out << "(" << ep.x << ", " << ep.y << ", " << ep.z << ")";
+}
+
+class Streamer {
```
MegabytePhreak, line 61: This is very similar to the class of the same name in PNGLoader.cpp, can it be factored out and shared?

mattco98: Possibly, yeah. There were two reasons I didn't though. First of all, PNG is big-endian, whereas BMP is little-endian. Secondly, I really don't like the whole return-by-reference-parameter pattern, which is why I wanted to make my own Streamer class with a simpler API. I'll play around with it and see what I come up with.

mattco98: I don't really think it's worth it. They're different (and small) enough where there isn't really much gain. None of the functions can really be taken out, since both loaders have different requirements.

PR #2587, opened 2020-06-19
Themes: Add Windows95 Theme

jsdevv:
Adds a classical theme from the 90's.

image


sunverwerth:
Isn't that the same as the Redmond theme? Also I don't think we should have trademarked names like Windows95 in the UI.

jsdevv:
So, should I remove this or rename it or?

sunverwerth:
I gues we can keep it.
We already have Redmond(Windows 3.0) Redmond2000 (Windows 2000) so you should rename it to Redmond95

awesomekling:
The existing "Redmond" theme is already using the colors from Windows 95.

PR #2591, opened 2020-06-19
LibC: Implement file actions for posix_spawn

nico:


Libraries/LibC/spawn.cpp
```diff
@@ -117,33 +131,42 @@ int posix_spawnp(pid_t* out_pid, const char* path, const posix_spawn_file_action
...
+int posix_spawn_file_actions_destroy(posix_spawn_file_actions_t* actions)
 {
+    delete actions->state;
+    return 0;
 }
 
-int posix_spawn_file_actions_init(posix_spawn_file_actions_t*)
+int posix_spawn_file_actions_init(posix_spawn_file_actions_t* actions)
 {
+    actions->state = new posix_spawn_file_actions_state;
```
nico, line 167: Hm, I suppose this could throw. What's serenity's policy on OOM from alloc? Should I use `new (std::nothrow)` here? Does Vector have a nothrow append() version? Should wrap this in a try/catch(bad_alloc)? Or does oom just always kill the process?

alimpfard: I'm pretty sure we don't have any exceptions (yet?), so oom would just crash the process.

awesomekling: We don't use exceptions (by choice) :)
malloc() currently does not fail (it crashes instead.) If you want failable allocations, there's mmap().

nico: Ok, then I think this is should be fine as-is.

PR #2592, opened 2020-06-19
CMake: Make the install step more zen

nico:
The "Installing $foo..." messages are just noise, so turn them off.

Not sure how folks feel about this; personally I haven't found the 'installing' messages useful so far.

If others do like them, there's also LAZY which at least turns off the "$foo is up-to-date" messages which maybe more people can agree on are useless.


awesomekling:
Sure why not!

PR #2598, opened 2020-06-20
LibWeb: Add atob/btoa to WindowObject

XXLuigiMario:
Adds functions [atob](https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/atob) and [btoa](https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/btoa) to the `window` object.
Note that `atob` is supposed to fail if the passed string is not a valid Base64-encoded string.
As it stands, our decoder will gladly decode any character, even if it's not part of the Base64 alphabet.


awesomekling:
Cool :)

PR #2604, opened 2020-06-21
LibWeb: Update parser to support more insertion modes

stelar7:
Implements InHeadNoScript, InSelectInTable, InTemplate, InFrameset, AfterFrameset, and AfterAfterFrameset.

Adds a missing case in InHead.

Fills out some `TODO()`s that I found in InHead, AfterHead, InBody

The handling of attributes in MathML and SVG elements feels not correct, but for now I've just expanded the Attribute object...


awesomekling:
Great stuff! :star:

Issue #2493, opened 2020-06-04
LibM: Add moar math for Janet Lang


The-King-of-Toasters:
I'm trying to port [The Janet Language](https://github.com/janet-lang/janet), and it creates wrappers for mathematical functions we don't have. These include:

- [x] `acosh`
- [x] `asinh`
- [x] `atanh`
- [x] `exp2`
- [x] `expm1`
- [x] `cbrt`
- [ ] `trunc`, implemented but not defined in `math.h`.
- [x] `gamma`
- [x] `log1p`
- [x] `erf`
- [x] `erfc`
- [x] `hypot`
- [ ] `nextafter`

We're also missing some macros:
- [x] `INFINITY`
- [ ] `isnan` and `isinf`. The rest of the `fpclassify` functions would be nice too.

bugaevc:
OMG, a language named after Janet! Yes please, and we have to get the port working  :)

alimpfard:
And here we see @bugaevc fanboying over something (admittedly nice).
meanwhile here I am liking its lispiness more than the character it references...

IdanHo:
all of these functions are implemented at this point, so this can probably be closed.

The-King-of-Toasters:
Hurray! Unfortunately janet makes use of `getaddrinfo` so I'll have to patch that out for a port :(.

PR #2605, opened 2020-06-21
LibM: Add some more math functions

stelar7:
Progress on #2493, still needs `nextafter`, and possibly `float` and `long double` versions of the functions.. :^)


Libraries/LibM/math.cpp
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2018-2020, Andreas Kling <kling@serenityos.org>
+ * Copyright (c) 2018-2020, SerenityOS developers
```
awesomekling, line 2: Instead of removing someone else from a copyright header, please add yourself to it :)

Libraries/LibM/math.cpp
```diff
@@ -474,4 +474,66 @@ double modf(double x, double* intpart)
...
+
+double expm1(double x)
+{
+    return pow(M_E, x) - 1;
+}
+
+double cbrt(double x)
+{
+    if (x > 0) {
+        return pow(x, 1 / 3);
```
awesomekling, line 492: This can't be right. `1` and `3` are integers here, so you'll be calling `pow(x, 0)`

You can expose these functions on the `Math` object in `LibJS`. That will allow you to add tests for them so we can see that they work :)

stelar7: Good catch!

I've tried adding the tests in LibJS, but they all fail, even though the output is correct in the console.. 🤔 

image

awesomekling: Yeah I think this happens because string/double conversion is not super great in Serenity (yet).
You can work around the issue by using the `isClose()` helper, see for example how it's done in the `Math.exp.js` test :)

Libraries/LibJS/Tests/Math.acosh.js
```diff
@@ -0,0 +1,13 @@
+load("test-common.js");
+
+try {
+    assert(Math.acosh(-1) === NaN);
+    assert(Math.acosh(0) === NaN);
+    assert(Math.acosh(0.5) === NaN);
```
linusg, line 6: That's not how NaN checks work.

Libraries/LibJS/Tests/Math.atanh.js
```diff
@@ -0,0 +1,14 @@
+load("test-common.js");
+
+try {
+    assert(Math.atanh(-2) === NaN);
```
linusg, line 4: That's not how NaN checks work.

Libraries/LibJS/Tests/Math.atanh.js
```diff
@@ -0,0 +1,14 @@
...
+    assert(Math.atanh(-1) === -Infinity);
+    assert(Math.atanh(0) === 0);
+    assert(isClose(Math.atanh(0.5), 0.549306));
+    assert(Math.atanh(1) === Infinity);
+    assert(Math.atanh(2) === NaN);
```
linusg, line 9: That's not how NaN checks work.

Libraries/LibJS/Tests/Math.cbrt.js
```diff
@@ -0,0 +1,16 @@
+load("test-common.js");
+
+try {
+    assert(Math.cbrt(NaN) === NaN);
```
linusg, line 4: That's not how NaN checks work.

Libraries/LibJS/Tests/Math.log1p.js
```diff
@@ -0,0 +1,13 @@
+load("test-common.js");
+
+try {
+    assert(Math.log1p(-2) === NaN);
```
linusg, line 4: That's not how NaN checks work.

Libraries/LibM/math.cpp
stelar7: Unfortunately `isClose` isnt close enough in this case; 
Chrome gives the value
`Math.acosh(2) == 1.3169578969248166`
while serenity gives 
`Math.acosh(2) == 1.3169589711848481`

And that falls just outside the isClose range :(
Its the same, if not worse for `asinh` and `atanh`. So either `exp` or `log` (or both) needs an update aswell

Libraries/LibJS/Runtime/MathObject.cpp
```diff
@@ -262,4 +267,48 @@ JS_DEFINE_NATIVE_FUNCTION(MathObject::clz32)
     return Value(__builtin_clz((unsigned)number.as_double()));
 }
 
+JS_DEFINE_NATIVE_FUNCTION(MathObject::acosh) {
```
awesomekling, line 270: Please use `clang-format`.

Libraries/LibJS/Runtime/MathObject.cpp
```diff
@@ -262,4 +267,48 @@ JS_DEFINE_NATIVE_FUNCTION(MathObject::clz32)
...
+    auto number = interpreter.argument(0).to_number(interpreter);
+    if (interpreter.exception())
+        return {};
+    if (number.as_double() < 1)
+        return JS::js_nan();
+    return Value(::acosh(number.as_double()));
+}
+
+JS_DEFINE_NATIVE_FUNCTION(MathObject::asinh) {
+    {
```
awesomekling, line 280: No need for these extra braces.

Libraries/LibJS/Runtime/MathObject.cpp
```diff
@@ -262,4 +267,48 @@ JS_DEFINE_NATIVE_FUNCTION(MathObject::clz32)
...
+        auto number = interpreter.argument(0).to_number(interpreter);
+        if (interpreter.exception())
+            return {};
+        return Value(::asinh(number.as_double()));
+    }
+}
+JS_DEFINE_NATIVE_FUNCTION(MathObject::atanh)
```
awesomekling, line 287: Please leave an empty line between functions.

Libraries/LibJS/Runtime/MathObject.cpp
```diff
@@ -262,4 +267,48 @@ JS_DEFINE_NATIVE_FUNCTION(MathObject::clz32)
...
+JS_DEFINE_NATIVE_FUNCTION(MathObject::log1p)
+{
+    auto number = interpreter.argument(0).to_number(interpreter);
+    if (interpreter.exception())
+        return {};
+    if (number.as_double() < -1)
+        return JS::js_nan();
+    return Value(::log1p(number.as_double()));
+}
+JS_DEFINE_NATIVE_FUNCTION(MathObject::cbrt)
```
awesomekling, line 306: Please leave an empty line between functions.

awesomekling:
:star:

PR #2607, opened 2020-06-21
Add CLion directories to .gitignore

mattco98:
I saw some discussion on IRC about these two directories and realized that I've just been manually ignoring them for some strange reason instead of just submitting a PR :)

P.S.: Should my commit have a prefix? `Meta: `?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,3 +35,5 @@ compile_commands.json
 .clang_complete
 *Endpoint.h
 Ports/packages.db
+.idea/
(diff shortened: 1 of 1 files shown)


awesomekling:
Yeah this is normally the kind of stuff that goes under `Meta:` :)

Sobak:
@mattco98 thank you :) It is something that I should do, in fact, but I had to leave for some time. Glad that you took the initiative.

PR #2608, opened 2020-06-21
Build: Support GENERATED_SOURCES in serenity_{bin,libc}() as well

linusg:
Not sure if that's the CMake way (cc @bugaevc) but it works. I think it can't figure out by itself that the `WebView` demo depends on generated `WebContent` files?

This fixes parallel building with `make` (`-jN`), which would currently fail with:

```
[ 87%] Building CXX object Demos/WebView/CMakeFiles/WebView.dir/WebContentView.cpp.o
In file included from /home/linus/Dev/serenity/Demos/WebView/WebContentView.cpp:28:
/home/linus/Dev/serenity/Demos/WebView/WebContentClient.h:31:10: fatal error: WebContent/WebContentClientEndpoint.h: No such file or directory
   31 | #include <WebContent/WebContentClientEndpoint.h>
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
make[2]: *** [Demos/WebView/CMakeFiles/WebView.dir/build.make:96: Demos/WebView/CMakeFiles/WebView.dir/WebContentView.cpp.o] Error 1
make[2]: *** Waiting for unfinished jobs....
```


bugaevc:
> Not sure if that's the CMake way

There's no proper support in CMake for dependencies on generated sources, unfortunately.
The recommended hack is to make a dummy target that depends on the generated files (you have to make the target while you're in the directory where you're generating the files, because there depending on them kinda does work):

https://github.com/SerenityOS/serenity/blob/dd7cd92de4f5dbe94d1aaca36ea3cc0a3befd580/CMakeLists.txt#L98-L99

And then depend on that other target where you need those generated files:

https://github.com/SerenityOS/serenity/blob/dd7cd92de4f5dbe94d1aaca36ea3cc0a3befd580/CMakeLists.txt#L67-L73

The CMake buildscript for WebView already uses `GENERATED_SOURCES`, so all should be well... except I only added support for `GENERATED_SOURCES` to `serenity_lib()`, not `serenity_bin()`, that's why it doesn't work. Add that (ideally, without increasing code duplication) ~is left as an exercise for the reader~ would be the right way to go about this.

P.S. What you've done is making WebView depend on WebContent, not the other way around.

linusg:
> P.S. What you've done is making WebView depend on WebContent, not the other way around.

Yes, that was my intention. Title and commit messages are wrong, lol.

bugaevc:
If I'm not being clear enough, please don't actually do that (making WebView depend on WebContent), just add `GENERATED_SOURCES` support to `serenity_bin`

linusg:
@bugaevc oh it certainly was clear enough, just actually doing it took a while :)

Please have a look if my solution looks sane...

bugaevc:
It does look sane :smiley: I'd name it simply `serenity_generated_sources()`, but up to you. And add it to `serenity_libc()` as well, perhaps?

PR #2609, opened 2020-06-21
LibC: Implement pselect.

nico:
@bugaevc As far as I can tell you didn't touch any of these files, but it looks like the kind of thing you might find interesting, so be cc'd.

### DIFF
--- a/AK/Time.h
+++ b/AK/Time.h
@@ -35,7 +35,7 @@ inline void timeval_sub(const TimevalType& a, const TimevalType& b, TimevalType&
     result.tv_usec = a.tv_usec - b.tv_usec;
     if (result.tv_usec < 0) {
         --result.tv_sec;
-        result.tv_usec += 1000000;
+        result.tv_usec += 1'000'000;
     }
 }
 
@@ -44,9 +44,31 @@ inline void timeval_add(const TimevalType& a, const TimevalType& b, TimevalType&
 {
     result.tv_sec = a.tv_sec + b.tv_sec;
     result.tv_usec = a.tv_usec + b.tv_usec;
-    if (result.tv_usec > 1000000) {
+    if (result.tv_usec > 1'000'000) {
         ++result.tv_sec;
-        result.tv_usec -= 1000000;
+        result.tv_usec -= 1'000'000;
+    }
+}
+
+template<typename TimespecType>
+inline void timespec_sub(const TimespecType& a, const TimespecType& b, TimespecType& result)
+{
+    result.tv_sec = a.tv_sec - b.tv_sec;
+    result.tv_nsec = a.tv_nsec - b.tv_nsec;
+    if (result.tv_nsec < 0) {
+        --result.tv_sec;
+        result.tv_nsec += 1'000'000'000;
+    }
+}
+
+template<typename TimespecType>
+inline void timespec_add(const TimespecType& a, const TimespecType& b, TimespecType& result)
+{
+    result.tv_sec = a.tv_sec + b.tv_sec;
+    result.tv_nsec = a.tv_nsec + b.tv_nsec;
+    if (result.tv_nsec > 1000'000'000) {
+        ++result.tv_sec;
+        result.tv_nsec -= 1000'000'000;
     }
 }
 
@@ -66,7 +88,9 @@ inline void timespec_to_timeval(const TimespecType& ts, TimevalType& tv)
 
 }
 
+using AK::timespec_add;
+using AK::timespec_sub;
+using AK::timespec_to_timeval;
 using AK::timeval_add;
 using AK::timeval_sub;
 using AK::timeval_to_timespec;
-using AK::timespec_to_timeval;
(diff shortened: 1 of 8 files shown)


bugaevc:
> As far as I can tell you didn't touch any of these files, but it looks like the kind of thing you might find interested, so be cc'd.

Yep, thank you very much! Nitpicks incoming :smile:

AK/Time.h
```diff
@@ -44,9 +44,31 @@ inline void timeval_add(const TimevalType& a, const TimevalType& b, TimevalType&
...
+}
+
+template<typename TimespecType>
+inline void timespec_sub(const TimespecType& a, const TimespecType& b, TimespecType& result)
+{
+    result.tv_sec = a.tv_sec - b.tv_sec;
+    result.tv_nsec = a.tv_nsec - b.tv_nsec;
+    if (result.tv_nsec < 0) {
+        --result.tv_sec;
+        result.tv_nsec += 1000 * 1000 * 1000;
```
bugaevc, line 60: *Perhaps* it'd be better to write it using C++14 ticks as `1'000'000'000`? Not sure it'd be better though.

Libraries/LibC/sys/time.h
```diff
@@ -64,6 +42,95 @@ struct timezone {
     int tz_dsttime;
 };
 
+
 int gettimeofday(struct timeval* __restrict__, void* __restrict__) __attribute__((nonnull(1)));
 
+
+static inline void timeradd(struct timeval* a, struct timeval* b, struct timeval* out)
```
bugaevc, line 49:
```suggestion
static inline void timeradd(const struct timeval* a, const struct timeval* b, struct timeval* out)
```

AK/Time.h
nico: Ah yeah, I did that in sys/time.h but then figured it doesn't work in C. It would work in this file, but then it'd be different from sys/time.h, which seems fine actually. Changing, thanks.

Libraries/LibC/sys/time.h
nico: Err, good point. https://man7.org/linux/man-pages/man3/timercmp.3.html didn't have them and I wondered about it, but that page says "macros" so anything goes. Changing, thanks!

bugaevc:
> Change the syscall from `SC_select` to `SC_pselect` and implement `select()` in terms of `pselect()`.

* The usual way we refer to syscalls (or at least I do, heh) is by their implementation method name, so `sys$pselect` in the commit message instead of `SC_pselect`
* We usually name a syscall that implements a more generic operation after its simple form, so e.g. `sys$open` actually corresponds to `openat_with_path_length()`. So in this case, the syscall should keep its name

nico:
That's a bit surprising to me, but done.

Kernel/Process.cpp
```diff
@@ -2958,7 +2959,8 @@ int Process::sys$select(const Syscall::SC_select_params* params)
     fd_set* readfds = params->readfds;
     fd_set* writefds = params->writefds;
     fd_set* exceptfds = params->exceptfds;
-    timeval* timeout = params->timeout;
+    const timespec* timeout = params->timeout;
+    const sigset_t *sigmask = params->sigmask;
```
bugaevc, line 2963:
```suggestion
    const sigset_t* sigmask = params->sigmask;
```

nico: I ran `git rebase -i HEAD~3`, picked 'edit' for each commit, and ran `git clang-format --diff HEAD^; git commit -a --amend --no-edit; git rebase --continue` for each of the 3 commits. Is there some nicer way to do this?

bugaevc: I would try `git rebase --exec` or even `git filter-branch`. Or simpler, just run clang-format once, then commit the changes, splitting them into different commits for different parts of code, then squash those commits into the originals (may even use autosquash functionality for this).

awesomekling: I have `clang-format` bound to a keyboard shortcut in my editor and press it compulsively. It works pretty well :)

Libraries/LibC/sys/select.cpp
```diff
@@ -28,13 +28,26 @@
...
-int select(int nfds, fd_set* readfds, fd_set* writefds, fd_set* exceptfds, struct timeval* timeout)
+int select(int nfds, fd_set* readfds, fd_set* writefds, fd_set* exceptfds, timeval* timeout_tv)
 {
-    Syscall::SC_select_params params { nfds, readfds, writefds, exceptfds, timeout };
+    timespec* timeout_ts = nullptr;
+    timespec timeout;
+    if (timeout_tv) {
+      timeout_ts = &timeout;
+      TIMEVAL_TO_TIMESPEC(timeout_tv, timeout_ts);
+    }
```
awesomekling, line 42: Please use `clang-format`.

nico: I tried! I think the problem was that I passed `--diff` to `git clang-format` which made it print a diff instead of apply it.

This seems to work well to clang-format all commits on a branch, without needing any user input:

    git rebase -x 'git clang-format HEAD^ && git commit -a --amend --no-edit' master

PR #2610, opened 2020-06-21
Ports: Add ninja

nico:
- 1.8.2 for now, newer versions need high-res timestamp file APIs
  which serenity doesn't have yet
- pselect() instead of ppoll() for now, same reason (depends on #2609)
- no good default for -j yet (see nproc.patch)
- `-l` probably doesn't work yet (see loadavg.patch), but I've never
  used that anyways
- some minor include patches that I've also sent upstream

Other than that, this seems to work reasonably well. It currently
produces some spam on stdout from probably the shell.

### DIFF
--- a/Ports/ninja/package.sh
+++ b/Ports/ninja/package.sh
@@ -1,0 +1,17 @@
+#!/bin/bash ../.port_include.sh
+port=ninja
+version=1.8.2
+workdir=ninja-1.8.2
+files="https://github.com/ninja-build/ninja/archive/v1.8.2.tar.gz ninja-v1.8.2.tar.gz"
+
+build() {
+    CXX=i686-pc-serenity-g++ CXXFLAGS="--sysroot=${SERENITY_ROOT}/Build/Root" \
+    LDFLAGS="--sysroot=${SERENITY_ROOT}/Build/Root" \
+    # platform=linux is close enough.
+    run ./configure.py --bootstrap --platform=linux --force-pselect
+    strip "${workdir}/ninja"
+}
+
+install() {
+    cp "${workdir}/ninja" "${SERENITY_ROOT}/Build/Root/usr/local/bin/ninja"
+}
--- a/Ports/ninja/patches/loadavg.patch
+++ b/Ports/ninja/patches/loadavg.patch
@@ -1,0 +1,19 @@
+diff --git a/src/util.cc b/src/util.cc
+index ae94d34..72cf501 100644
+--- a/src/util.cc
++++ b/src/util.cc
+@@ -566,11 +566,14 @@ double GetLoadAverage() {
+ #else
+ double GetLoadAverage() {
+   double loadavg[3] = { 0.0f, 0.0f, 0.0f };
++  /*
++   * FIXME: serenity way of getting loadavg
+   if (getloadavg(loadavg, 3) < 0) {
+     // Maybe we should return an error here or the availability of
+     // getloadavg(3) should be checked when ninja is configured.
+     return -0.0f;
+   }
++  */
+   return loadavg[0];
+ }
+ #endif // _WIN32
--- a/Ports/ninja/patches/nproc.patch
+++ b/Ports/ninja/patches/nproc.patch
@@ -1,0 +1,13 @@
+diff --git a/src/util.cc b/src/util.cc
+index ae94d34..72cf501 100644
+--- a/src/util.cc
++++ b/src/util.cc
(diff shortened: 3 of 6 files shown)


Ports/ninja/patches/loadavg.patch
```diff
@@ -0,0 +1,19 @@
+diff --git a/src/util.cc b/src/util.cc
+index ae94d34..72cf501 100644
+--- a/src/util.cc
++++ b/src/util.cc
+@@ -566,11 +566,14 @@ double GetLoadAverage() {
+ #else
+ double GetLoadAverage() {
+   double loadavg[3] = { 0.0f, 0.0f, 0.0f };
++  /*
++   * FIXME: synergos way of getting loadavg
```
awesomekling, line 10: synergos? :thinking:

Ports/ninja/patches/nproc.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/src/util.cc b/src/util.cc
+index ae94d34..72cf501 100644
+--- a/src/util.cc
++++ b/src/util.cc
+@@ -476,7 +476,7 @@ int GetProcessorCount() {
+   GetNativeSystemInfo(&info);
+   return info.dwNumberOfProcessors;
+ #else
+-  return sysconf(_SC_NPROCESSORS_ONLN);
++  return 1;  // FIXME: synergyos way of saying sysconf(_SC_NPROCESSORS_ONLN);
```
awesomekling, line 10: :thinking:

Ports/ninja/package.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash ../.port_include.sh
+port=ninja
+version=1.8.2
+workdir=ninja-1.8.2
+files="https://github.com/ninja-build/ninja/archive/v1.8.2.tar.gz ninja-v1.8.2.tar.gz"
+
+build() {
+    CXX=i686-pc-serenity-g++ CXXFLAGS=--sysroot=${SERENITY_ROOT}/Build/Root \
+    LDFLAGS=--sysroot=${SERENITY_ROOT}/Build/Root \
```
bugaevc, line 9: Always quote `SERENITY_ROOT` — I happen to have it in `/home/sergey/dev/serenity`, which doesn't contain any special characters, but others may not. (By the way, this wouldn't be a problem in our Shell :slightly_smiling_face:)

Ports/ninja/package.sh
```diff
@@ -0,0 +1,16 @@
...
+    run ./configure.py --bootstrap --force-pselect
+    strip ${workdir}/ninja
```
bugaevc, line 11: Same here

Ports/ninja/package.sh
nico:
```
$ rg SERENITY_ROOT .. | grep -v '"' | grep -v ninja | wc -l
14
```
:thinking:

Ports/ninja/patches/loadavg.patch
nico: :disappointed:

Ports/ninja/patches/nproc.patch
nico: :thought_balloon:

Ports/ninja/package.sh
bugaevc: Clearly, some code does it wrong, doesn't mean we should add more code that does it wrong :slightly_smiling_face:

Ports/ninja/package.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash ../.port_include.sh
+port=ninja
+version=1.8.2
+workdir=ninja-1.8.2
+files="https://github.com/ninja-build/ninja/archive/v1.8.2.tar.gz ninja-v1.8.2.tar.gz"
+
+build() {
+    CXX=i686-pc-serenity-g++ CXXFLAGS="--sysroot=${SERENITY_ROOT}/Build/Root" \
+    LDFLAGS="--sysroot=${SERENITY_ROOT}/Build/Root" \
+    # platform=linux is close enough.
```
bugaevc, line 10: Question: why not add proper Serenity support to `configure.py`? As far as I can see `'linux'` is only mentioned in a few places, and the only thing it actually matters for is whether the system supports `ppoll`...

nico: Because it's as you say not needed. Fewer patches seems good.

PR #2611, opened 2020-06-22
LibGfx: Add PPM image file type support

asliturk:
Hi @awesomekling,
Is it a good way to put test images in change set?


awesomekling:
> Hi @awesomekling,
> Is it a good way to put test images in change set?

Sure that seems fine :)

PR #2612, opened 2020-06-22
Implement ppoll and use it in ninja

nico:


Kernel/Process.cpp
```diff
@@ -3053,18 +3053,33 @@ int Process::sys$select(const Syscall::SC_select_params* params)
     return marked_fd_count;
 }
 
-int Process::sys$poll(pollfd* fds, int nfds, int timeout)
+int Process::sys$poll(const Syscall::SC_poll_params* params)
 {
     REQUIRE_PROMISE(stdio);
-    if (!validate_read_typed(fds))
```
nico, line 3059: I think this should've passed nfds as 2nd parameter.

nico:
@bugaevc

Kernel/Process.cpp
```diff
@@ -3053,18 +3053,33 @@ int Process::sys$select(const Syscall::SC_select_params* params)
...
+    if (sigmask && !validate_read_typed(sigmask))
+        return -EFAULT;
+    if (!validate_read_typed(fds))
+        return -EFAULT;
+
     Thread::SelectBlocker::FDVector rfds;
     Thread::SelectBlocker::FDVector wfds;
 
-    for (int i = 0; i < nfds; ++i) {
+    for (unsigned long i = 0; i < nfds; ++i) {
```
bugaevc, line 3082: ~We usually~ Unixes of all kinds use `int` for fds, so why use `unsigned long` here?

bugaevc: Hmm, probably because of `nfds_t`. Then why not use that here? TBH I'd prefer an `int` nevertheless

nico: https://pubs.opengroup.org/onlinepubs/009695399/basedefs/poll.h.html "An unsigned integer type used for the number of file descriptors." I guess there's no reason to make it long though.

nico: `int` makes gcc complain about different signage. I think Kernel can't linclude LibC (?); at least several other structs don't use the LibC typedefs. Moved it from unsigned long to unsigned.

bugaevc: The kernel can't include LibC, but we have `UnixTypes.h` for that

awesomekling: Indeed, although let's be honest, it's pretty smelly to have so many things duplicated in `UnixTypes.h`.
Maybe one day we will have some kind of common type header shareable between the Kernel and LibC..

nico: With UnixTypes it can be an nfds_t in Process.cpp but not in Syscall.h as far as I understand. Is that what you want?

Kernel/Process.cpp
awesomekling: Yes, you are right!

PR #2613, opened 2020-06-23
Kernel: Implement FreeBSD's Fortuna algorithm for /dev/random

Petelliott:
Serenity's current random implementation directly calls the `rdrand` instruction. This pull request will hopefully provide secure rng for platforms without `rdrand` and to provide additional randomness.

Fortuna does not block, and does not attempt to estimate entropy, making it suitable for both `/dev/random` and `/dev/urandom`.

My implementation is based on [this](https://www.schneier.com/academic/paperfiles/fortuna.pdf#page=8) textbook chapter from  *Cryptography Engineering: Design Principles and Practical Applications*.

Disclaimer: I am not a cryptographer by any means.

- [x] Implement the Fortuna algorithm
- [x] Replace the current random implementation (without changing the API)
- [x] Gather entropy from various drivers
- [x] ~Save some randomness to a seed file so we can generate random numbers right after boot.~

(EDIT: FreeBSD does not implement a seed file, and neither does any other unix I can find. FreeBSD blocks until enough random data is available, however I think it is fine for us to just seed the generator with `rdseed`, and the generator will be reseeded several times during boot anyways.)

### DIFF
--- a/Kernel/Random.h
+++ b/Kernel/Random.h
@@ -1,5 +1,6 @@
 /*
  * Copyright (c) 2018-2020, Andreas Kling <kling@serenityos.org>
+ * Copyright (c) 2020, Peter Elliott <pelliott@ualberta.ca
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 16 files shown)


Kernel/Random.h
```diff
@@ -26,10 +27,98 @@
...
+        m_pools[pool].update(reinterpret_cast<u8*>(&event_data), sizeof(T));
+    }
+
+private:
+    void generate_block(CipherType cipher, u8* buffer, size_t size)
+    {
+        BlockType input((u8*)m_counter, sizeof(m_counter));
+        BlockType output;
+        cipher.encrypt_block(input, output);
+        m_counter++;
```
alimpfard, line 95: :thinking: 

This looks like CTR mode, if it has all the implications of CTR mode, you could just move these stuff into `Mode/CTR`?
Feel free to stub out the decrypt operation if you do.

Petelliott: Done! and as a side effect, `get_random_bytes` is way more elegant. However, in order to use the CTR mode, we now need to create an `AK::ByteBuffer` that is the same length as requested by the user.

I can think of a few solutions to this:

1) Add an API specific to CTR mode that does not encrypt anything.
2) Make the input argument to `Mode::encrypt` optional.
3) Treat empty input to `CTR::encrypt` specially.
4) Just ignore it until it becomes a problem.

alimpfard: hmmm...I see how that might be a little verbose to use, but honestly...isn't all of crypto too damn verbose?
mode-specific APIs are fine, if you want to do that, go ahead.
but for the time being, I don't think it's strictly necessary.

Petelliott: I am more concerned by the extra allocation (which could be quite large), rather than the verbosity. I'll probably add a specific API.

Petelliott: I have added the `key_stream` method to `CTR` which takes no input buffer, and amended my commit.

Libraries/LibCrypto/Cipher/Mode/CTR.h
```diff
@@ -0,0 +1,138 @@
...
+                num_view[i] = 0;
+            } else {
+                num_view[i]++;
+                break;
+            }
+        }
+        return new_buffer;
+    }
+
+    Optional<ByteBuffer> encrypt_or_stream(Optional<const ByteBuffer*> in, ByteBuffer& out, Optional<ByteBuffer> ivec)
```
alimpfard, line 95: what's with the `Optional<const ByteBuffer*>`? the pointer itself can just be null and you'll have your optional none value...

alimpfard: (sorry, I should probably keep these nitpicks to when you mark it as not-a-draft, but I'm impatient)

Petelliott: nitpick away. I think that's leftover from when i was trying to make an `Optional<const ByteBuffer&>`, which is impossible. I'll remove it.

Kernel/Random.h
```diff
@@ -120,6 +122,36 @@ class KernelRng : public FortunaPRNG<Crypto::Cipher::AESCipher, Crypto::Hash::SH
...
+public:
+    EntropySource()
+        : m_source(next_source++)
+    {
+    }
+
+    template<typename T>
+    void add_random_event(const T& event_data)
+    {
+        // we don't lock this because on the off chance a pool is corrupted, entropy isn't lost
```
alimpfard, line 142: The general format of the comments here is similar to a sentence, e.g. `// We don't [...] entropy isn't lost.`

Kernel/Random.h
```diff
@@ -111,12 +99,11 @@ class FortunaPRNG {
         m_key = AK::ByteBuffer::copy(digest.immutable_data(),
             digest.data_length());
 
-        m_counter++;
         m_reseed_number++;
         m_p0_len = 0;
     }
 
-    size_t m_counter { 0 };
+    AK::ByteBuffer m_counter;
```
alimpfard, line 108: There's no need to qualify `AK` here.

Kernel/Random.h
```diff
@@ -67,9 +69,10 @@ class FortunaPRNG {
 
         // generate a new key
 
-        for (size_t i = 0; i < KeySize; i += block_size) {
+        for (size_t i = 0; i < KeySize/8; i += block_size) {
```
alimpfard: I'm getting a feeling that you're not running clang-format?
Please run clang-format on these (and if possible, squash the changes to their respective commits).

Petelliott: All files are clang formatted (at lease at HEAD) Except for this file in this commit. i'm not sure that I can format and squash these changes, because they no longer exist.

alimpfard: That's fine then.
I was just making sure :)

alimpfard:
All that remains is to fix the commit messages, I have nothing else to add.
@awesomekling I'm not particularly familiar with the lock stuff in the kernel...all yours...

Petelliott:
> All that remains is to fix the commit messages, I have nothing else to add.

Done

awesomekling:
Cool! :)

PR #2616, opened 2020-06-23
Taskbar: Fix divide-by-zero

pimvanriezen:
FIx problem where presenting the Taskbar process with a progress message with a min/max range of size 0 would probably crash it. As far as I can tell this could also be triggered by getting the proper escape sequence sent to a terminal window.

### DIFF
--- a/Services/Taskbar/TaskbarButton.cpp
+++ b/Services/Taskbar/TaskbarButton.cpp
@@ -60,7 +60,7 @@ void TaskbarButton::resize_event(GUI::ResizeEvent& event)
 static void paint_custom_progress_bar(GUI::Painter& painter, const Gfx::IntRect& rect, const Gfx::IntRect& text_rect, const Palette& palette, int min, int max, int value, const StringView& text, const Gfx::Font& font, Gfx::TextAlignment text_alignment)
 {
     float range_size = max - min;
-    float progress = (value - min) / range_size;
+    float progress = (value - min) / (range_size > 0.0f ? range_size : 1.0f);
     float progress_width = progress * rect.width();
 
     Gfx::IntRect progress_rect { rect.x(), rect.y(), (int)progress_width, rect.height() };


awesomekling:
Hello, and thank you for working on SerenityOS!

Please observe the contribution guidelines here: https://github.com/SerenityOS/serenity/blob/master/CONTRIBUTING.md

Specifically:
* Please use `clang-format` (version 10+) 
* Please wrap your commit message at 72 characters per line

Also, your commit message speculates about how this would probably cause a crash. Generally speaking, please test your changes to make sure they actually work. It also refers to the same bug possibly existing in another component. Perhaps you could fix that as well? :)

pimvanriezen:
Will amend the patch for formatting.

The crash will only occur in programs that consume the custom progress bar messags and interpret its meaning without sanity checks, which, at the moment, is only Taskbar.

The point I was trying to make that one of the programs that can -send- this bogus data is the Terminal, which allows a special escape sequence to trigger sending these events. Traditionally this is a potential security problem if someone can manage to trigger a Network-facing service to put that escape sequence into a logfile that the user is likely to watch inside the terminal.

Since it's only capable of crashing Taskbar, and it's with a divide-by-zero exception, I don't think there's a path to privileged code execution, so at worst it's an annoying DoS-vulnerability.

pimvanriezen:
As an aside, from a wider perspective, you may want to wonder about how to make it likely that messages received from programs running in a different security context are actually up-to-sniff in any service set up to receive messages. Since they are already beholden to some kind of schema, getting those sanity-checks in without burdening the developer of the message consumer seems within the realm of possibilities. A bit too ambitious for me to take it on at that level, right now.

Services/Taskbar/TaskbarButton.cpp
```diff
@@ -60,7 +60,7 @@ void TaskbarButton::resize_event(GUI::ResizeEvent& event)
 static void paint_custom_progress_bar(GUI::Painter& painter, const Gfx::IntRect& rect, const Gfx::IntRect& text_rect, const Palette& palette, int min, int max, int value, const StringView& text, const Gfx::Font& font, Gfx::TextAlignment text_alignment)
 {
     float range_size = max - min;
-    float progress = (value - min) / range_size;
+    float progress = (value - min) / (range_size > 0 ? range_size : 1);
```
emilengler, line 63: Please use `0.0` and `1.0`, otherwise it can be interpreted as an integer. Also adding a `f` as a suffix doesn't hurt either

pimvanriezen: on it, thanks for catching that.

pimvanriezen:
Giving this some more thought, perhaps it's better to just out-right return out if the parameters are senseless, instead of normalizing them to something also nonsensical that just happens not to crash the program.

pimvanriezen:
Giving the surrounding code another read, it looks like the progress message has already been changed back to being clamped from 0 to 100 and I'm a massive idiot for chasing this.

Closing this merge request. Sorry for the time waste :-)

PR #2617, opened 2020-06-24
Meta: allow running run.sh from anywhere

emanuele6:
Also move `Kernel/.bochsrc` in `Meta/` and tweak `Meta/CLion/run.sh` to use `Meta/run.sh`.

~This pull request should also fix Bochs support.~


emanuele6:
We are documenting how to build serenity in two places:
1. [Toolchain/README.md](https://github.com/SerenityOS/serenity/tree/26eef6501765e6801a86497b5ab98e059c05ea61/Toolchain/README.md)
2. [Documentation/INSTALL.md](https://github.com/SerenityOS/serenity/tree/26eef6501765e6801a86497b5ab98e059c05ea61/Documentation/INSTALL.md)

Should we remove the `Serenity (Full build)` section from [Toolchain/README.md](https://github.com/SerenityOS/serenity/tree/26eef6501765e6801a86497b5ab98e059c05ea61/Toolchain/README.md)?

Also the `Running SerenityOS in an emulator` section needs to be updated to use `Meta/run.sh` instead of `./run` and to use `Meta/run.sh qcmd boot_mode=text` instead of `./run qtext`, should we move those instructions into another file (e.g. `Documentation/RunInstructions.md`)?

This could be another pull request.

emanuele6:
> This pull request should also fix Bochs support.

It fixes `bochs` support at least in the sense that now we can use `Meta/run.sh b` to launch `bochs` so that it uses `Meta/bochsrc`.

I tried usign `bochs`, but I get this error message and I don't know how to install the `e1000` plugin:
```
========================================================================
                       Bochs x86 Emulator 2.6.11
              Built from SVN snapshot on January 5, 2020
                Timestamp: Sun Jan  5 08:36:00 CET 2020
========================================================================
00000000000i[      ] BXSHARE not set. using compile time default '/usr/share/bochs'
00000000000i[      ] reading configuration from /home/emanuele6/pro/serenityOS/serenity/Meta/bochsrc
00000000000e[      ] /home/emanuele6/pro/serenityOS/serenity/Meta/bochsrc:12: ataX-master/slave CHS set to 0/0/0 - autodetection enabled
00000000000p[SIM   ] >>PANIC<< optional plugin 'e1000' not found
00000000000e[SIM   ] notify called, but no bxevent_callback function is registered
00000000000e[SIM   ] notify called, but no bxevent_callback function is registered
========================================================================
Bochs is exiting with the following message:
[SIM   ] optional plugin 'e1000' not found
========================================================================
00000000000i[SIM   ] quit_sim called with exit code 1
```

I also tried removing that plugin from `bochsrc`: `bochs` wasn't able to boot from `_disk_image` (it's generated by `Meta/build-image-qemu.sh` after all), so we probably need to edit `bochsrc` to use `grub_disk_image` instead of `_disk_image`.

`bochs` was able to boot from Hard Disk with the `grub_disk_image`, but after loading `serenity` from `grub`, I only got a black screen; maybe that's just because I don't have the `e1000` plugin enabled.

@deoxxa, can you please help me with this since you were the one who added `bochs` support?

Making `serenity` work again in `bochs` could also be another pull request.

emanuele6:
http://bochs.sourceforge.net/cgi-bin/topper.pl?name=New+Bochs+Documentation&url=http://bochs.sourceforge.net/doc/docbook/user/index.html

> --enable-e1000 | no | Enable Intel(R) 82540EM Gigabit Ethernet adapter support.
`e1000` is an ethernet adapter: I don't think not having it enabled is what's causing `serenity` to stop at black screen when running in `bochs` then.

Maybe we dropped support for `bochs` at some point and we didn't notice.

emanuele6:
Screenshot of the black screen I am getting (note: `e1000` disabled):

black_screen_bochs

awesomekling:
cc @bugaevc

Meta/CLion/run.sh
```diff
@@ -1,22 +1,23 @@
...
 
 set -e
 
-#SERENITY_PACKET_LOGGING_ARG="-object filter-dump,id=hue,netdev=breh,file=e1000.pcap"
+: "${SERENITY_RAM_SIZE:='256M'}"
+export SERENITY_RAM_SIZE
 
-[ -e /dev/kvm ] && [ -r /dev/kvm ] && [ -w /dev/kvm ] && SERENITY_KVM_ARG="-enable-kvm"
+# set SERENITY_ROOT to the path to the serenity directory
+: "${SERENITY_ROOT:='/mnt/c/Users/Ragnarok/serenity-project/serenity'}"
```
bugaevc, line 9: 😐 I don't think this is the way to allow running run.sh from anywhere

By the way, you can already do exactly that, you only need to set `SERENITY_ROOT` properly, so...

bugaevc: Or — perhaps I'm mistaken and you cannot run `run.sh` from anywhere, if it expects to find `_disk_image` and `Kernel/Kernel` in the current directory.

The question then is why would you want that?

emanuele6: That's the CLion version of `run.sh`, it used to be a copy of the standard `run.sh` with the default values of the variables changed.
That didn't make much sense: you can just override the value of the environment variables and then run the standard `run.sh` from that file. 

`run.sh` expected to find `./Kernel/Kernel` and `./_disk_image`; when running the script with the `b` (`bochs`) argument, it expects to find a bochs configuration file in `./.bochsrc`.

When using `run.sh` with the `qgrub` or the `b` option, we should use `grub_disk_image` as the disk image instead of using `_disk_image`.

> The question then is why would you want that?

Why not? The `bochs` configuration file is not in the build directory and we are not copying it there; since we have to use an absolute path to that file, let's also allow running `run.sh` anywhere.

If you want to run `run.sh` on the current directory, you can still do it: `SERENITY_BUILD= /path/to/run.sh`.

Meta/run.sh
```diff
@@ -3,6 +3,14 @@
 
 set -e
```
emanuele6, line 4: This one is the file I mainly changed.

Meta/CLion/run.sh
bugaevc: You still have to pass `run.sh` the directory where to look for `_disk_image` and the kernel for, *somehow*. I feel using the current directory for this is a nice, Unix-y way — better than using an environment variable. But perhaps you can convince me otherwise. Are there any cases where it's convenient to pass an environment variable, but inconvenient to set the current directory, for example?

bugaevc: Oh, and hardcoding a path like this is certainly a wrong thing to do 😀

emanuele6: > Oh, and hardcoding a path like this is certainly a wrong thing to do

What path are you refering to? `/mnt/c/Users/Ragnarok/serenity-project/serenity` ? That was already there (in `Meta/CLion/run.sh`).

I totally agree with you!

I just rewrote that script to use `Meta/run.sh`.

emanuele6: > You still have to pass run.sh the directory where to look for _disk_image and the kernel for, somehow.

Not really unless we use `-` options. `run.sh` already has arguments: `/path/run.sh b` emulates serenity using `bochs` for example, or `/path/to/run.sh qcmd boot_mode=text another_argument hello`, emulates serenity in qemu passing to the kernel `boot_mode=text`, `another_argument` and `hello` as command line argument.

I can just default `run.sh` to look in `.` for `_disk_image`, `grub_disk_image` and `Kernel/Kernel`, but to look on "$SERENITY_ROOT/Meta` for the `bochs` configuration file (`bochsrc`), unless SERENITY_BUILD is specified if you want.

Meta/run.sh
```diff
@@ -47,19 +61,30 @@ $SERENITY_EXTRA_QEMU_ARGS
 
 export SDL_VIDEO_X11_DGAMOUSE=0
 
+[ ! "${SERENITY_BUILD+is_set}" ] && {
+    if [ -d "$SERENITY_ROOT/Build" ]; then
+        SERENITY_BUILD="$SERENITY_ROOT/Build"
+    elif [ -d "$SERENITY_ROOT/build" ]; then
+        SERENITY_BUILD="$SERENITY_ROOT/build"
+    fi
+}
```
emanuele6, line 70: All it takes is removing this.

Meta/run.sh
```diff
@@ -47,19 +61,30 @@ $SERENITY_EXTRA_QEMU_ARGS
...
+: "${SERENITY_BUILD:='.'}"
```
ericonr, line 71: Can't you put this inside the previous block? Something like a final `else` if neither directory exists.

Meta/run.sh
```diff
@@ -47,19 +61,30 @@ $SERENITY_EXTRA_QEMU_ARGS
 
 export SDL_VIDEO_X11_DGAMOUSE=0
 
+[ ! "${SERENITY_BUILD+is_set}" ] && {
```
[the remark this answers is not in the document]
emanuele6: > I believe the correct way to to specify an alternative value is with :+ instead of just +. That said, this bit would be clearer with just if [ -z "$SERENITY_BUILD" ]; then

Why are you saying this? that is just wrong:
It is true that `[ ! "${SERENITY_BUILD:+is_set}" ]` would be the same as `[ -z  "$SERENITY_BUILD" ]`, but I'm using `+`, not `:+`...

The difference is what happens when "the variable is set, but empty", example:
```sh
#!/bin/sh
unset VARIABLE; echo unset VARIABLE
[ ! "${VARIABLE:+is_set}" ] && echo '[ ! "${VARIABLE:+is_set}" ]'
[ !  "${VARIABLE+is_set}" ] && echo '[ !  "${VARIABLE+is_set}" ]'
[ -z          "$VARIABLE" ] && echo '[ -z          "$VARIABLE" ]'
VARIABLE=; echo VARIABLE=
[ ! "${VARIABLE:+is_set}" ] && echo '[ ! "${VARIABLE:+is_set}" ]'
[ !  "${VARIABLE+is_set}" ] && echo '[ !  "${VARIABLE+is_set}" ]'
[ -z          "$VARIABLE" ] && echo '[ -z          "$VARIABLE" ]'
VARIABLE=hello; echo VARIABLE=hello
[ ! "${VARIABLE:+is_set}" ] && echo '[ ! "${VARIABLE:+is_set}" ]'
[ !  "${VARIABLE+is_set}" ] && echo '[ !  "${VARIABLE+is_set}" ]'
[ -z          "$VARIABLE" ] && echo '[ -z          "$VARIABLE" ]'
```
 The output of this script will be:
```
unset VARIABLE
[ ! "${VARIABLE:+is_set}" ]
[ !  "${VARIABLE+is_set}" ]
[ -z          "$VARIABLE" ]
VARIABLE=
[ ! "${VARIABLE:+is_set}" ]
[ -z          "$VARIABLE" ]
VARIABLE=hello
```
As you can see, `[ !  "${VARIABLE+is_set}" ]` is only true when the variable `VARIABLE` is not set while `[ -z          "$VARIABLE" ]` and `[ ! "${VARIABLE:+is_set}" ]` are also true when VARIABLE is set, but empty.

If you run the `run.sh` without setting SERENITY_BUILD, this block of code will be executed and SERENITY_BUILD will be set to either `SERENITY_ROOT/Build` or `SERENITY_ROOT/build`.
If you run the script, setting SERENITY_BUILD to an empty value (i.e. `SERENITY_BUILD= Meta/run.sh`), this code will not be executed and SERENITY_BUILD will be set to `.` at line 71.

Meta/run.sh
emanuele6: Why? If `SERENITY_BUILD` is not set, `SERENITY_ROOT/Build` does not exist/is not a directory and `SERENITY_ROOT/build` does not exist/is not a directory, then `: "${SERENITY_BUILD:='.'}" will set it to `.` (intended behaviour: read commit message).

We could fail in that case, I guess. Is that what you mean?

emanuele6: > Can't you put this inside the previous block?

Oh, I missed this part: no, I can't: if the variable is set, but empty, I still want to set `SERENITY_BUILD` to `.`

Meta/CLion/run.sh
emanuele6: ping @bugaevc

Meta/run.sh
emanuele6: ping @bugaevc

Meta/run.sh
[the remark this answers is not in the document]
emanuele6: > source: https://www.gnu.org/software/bash/manual/html_node/Shell-Parameter-Expansion.html

Your own source explains the difference between `:+` and `+`. :^)

> When not performing substring expansion, using the form described below (e.g., ‘:-’), Bash tests for a parameter that is unset or null. Omitting the colon results in a test only for a parameter that is unset.

Meta/CLion/run.sh
bugaevc: > I can just default `run.sh` to look in `.` for `_disk_image`, `grub_disk_image` and `Kernel/Kernel` unless `SERENITY_BUILD` is set as an environment variable, but to look in `"$SERENITY_ROOT/Meta"` for the `bochs` configuration file (`bochsrc`) if you want.

Yep, that sounds good to me

Meta/run.sh
bugaevc: Are you, uh, sure the `{}` doesn't make it a subshell (and so the variables set won't persist to the outer shell)?

emanuele6: Yeah, yeah, I'm sure. `()` creates a subshell; `{}` only groups commands.

bugaevc: Right; my bad.

Meta/run.sh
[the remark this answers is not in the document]
ericonr: Thanks for the example script, I didn't know that expansion was possible. I can't find documentation for it, though.

> Your own source explains the difference between :+ and +. :^)

Searching for `+` in it gives me a single result, which is in `${name:+word}`.

> If you run the script, setting SERENITY_BUILD to an empty value (i.e. SERENITY_BUILD= Meta/run.sh),

This looks like a weird use case, but I don't know the full requirements for it. I see your point, though.

> Note that the documentation you posted is for the bash shell; this is not a bash script, but these parameter expansions also work in POSIX sh.

I am aware. I couldn't find a POSIX shell specification, and between posting the bash manual and a link to dash's man page, I thought the Bash link had a better look.

> > I believe the correct way to to specify an alternative value is with :+ instead of just +.

> Also, are you confusing the + operator with the - operator?

The dash manual calls `${parameter:+word}` an alternative value.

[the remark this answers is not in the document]
emanuele6: > Thanks for the example script, I didn't know that expansion was possible. I can't find documentation for it, though.

I was refering to this paragraph.
> When not performing substring expansion, using the form described below (e.g., ‘:-’), Bash tests for a parameter that is unset or null. Omitting the colon results in a test only for a parameter that is unset. Put another way, if the colon is included, the operator tests for both parameter’s existence and that its value is not null; if the colon is omitted, the operator tests only for existence.

It basically says "when omitting the colon, the operator will only test if the variable is unset, instead of testing if the variable is unset or null".

i.e.
```man
${parameter:-word}
If parameter is unset or null, the expansion of word is substituted. Otherwise, the value of parameter is substituted.
```
without the colon becomes:
```man
${parameter-word}
If parameter is unset, the expansion of word is substituted. Otherwise, the value of parameter is substituted.
```

NOTE: "null" is how you call a variable which is set to an empty string.

> I couldn't find a POSIX shell specification

I usually use this website for reading POSIX specs (https://pubs.opengroup.org/onlinepubs/9699919799/).
Here is the spec for POSIX shell parameter expansions: https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_06_02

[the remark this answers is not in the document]
ericonr: I see, thank you very much for the explanations.

PR #2618, opened 2020-06-24
Userland: Add a 'test' utility

alimpfard:
@bugaevc 

This adds an incomplete implementation of the test util, missing some user/group checks, and `-l STRING`.
created because me want shell tests, and we had no way to compare things.

"minor" issue, if I name it "test", cmake complains that there's already a test target.
if I try to manually rename the target (I tried `set_target_properties(Test PROPERTIES RUNTIME_OUTPUT_NAME test)`), ninja complains that there are multiple rules that generate `Userland/test`, which I assume is the test target binary.
not sure how to fix that.


Base/usr/share/man/man1/test.md
```diff
@@ -0,0 +1,86 @@
...
+
+## Synopsis
+
+```**sh
+$ test EXPRESSION
+```
+
+## Description
+
+`test` takes a given expression and returns 0 if the expression evaluates to true, or 1 if not.
```
bugaevc, line 13: Rather, its exit code is 0 or 1. Saying "returns" about processes feels wrong

Base/usr/share/man/man1/test.md
```diff
@@ -0,0 +1,86 @@
...
+
+An omitted expression defaults to false, and an unexpected error causes a return value of 126.
+
+The expression can take any of the following forms:
```
bugaevc, line 17: I'd make this a separate section, like `## Expressions`

Base/usr/share/man/man1/test.md
```diff
@@ -0,0 +1,86 @@
...
+
+* `( EXPRESSION )` (grouping)
+
+Boolean operations
```
bugaevc, line 21: And these would be h3s

Base/usr/share/man/man1/test.md
```diff
@@ -0,0 +1,86 @@
...
+* `EXPRESSION -a EXPRESSION` (and)
+* `EXPRESSION -o EXPRESSION` (or)
+
+String comparison
+
+* `-n STRING` (non-empty)
+* `-z STRING` (empty)
+* `STRING = STRING` (equality)
+* `STRING != STRING` (unequality)
+* `STRING` (non-empty, if the only expression available)
```
bugaevc, line 33: By*available*, you probably mean *specified* or *present*?

Base/usr/share/man/man1/test.md
```diff
@@ -0,0 +1,86 @@
...
+File type checks
+
+* `-b FILE` (is block device?)
+* `-c FILE` (is character device?)
+* `-f FILE` (is regular file?)
+* `-d FILE` (is directory?)
+* `-p FILE` (is pipe?)
+* `-S FILE` (is socket?)
+* `-h FILE` (is symbolic link?)
+* `-L FILE` (is symbolic link?)
```
bugaevc, line 59: What's the difference?

Base/usr/share/man/man1/test.md
```diff
@@ -0,0 +1,86 @@
...
+None.
+
+## Examples
+
+```sh
+# Conditionally do something based on the value of a variable
+$ test "$foo" = bar && echo foo is bar
+# Check some numbers
+$ test \( 10 -gt 20 \) -o \( ! 10 -ne 10 \) && echo "magic numbers!"
+```
```
bugaevc, line 86: See also: `find`(1) and `stat`(2) (if we have the it)

Base/usr/share/man/man1/test.md
```diff
@@ -0,0 +1,86 @@
...
+
+None.
+
+## Examples
+
+```sh
+# Conditionally do something based on the value of a variable
+$ test "$foo" = bar && echo foo is bar
+# Check some numbers
+$ test \( 10 -gt 20 \) -o \( ! 10 -ne 10 \) && echo "magic numbers!"
```
bugaevc, line 85: Do the `\`s get srendered properly by LibMarkdown, by the way?

Userland/Test.cpp
```diff
@@ -0,0 +1,540 @@
...
+        : m_path(path)
+        , m_kind(kind)
+    {
+    }
+
+private:
+    virtual bool check() const override
+    {
+        if (!Core::File::exists(m_path))
+            return false;
```
bugaevc, line 150: Why check twice (and introduce more potential races), let's just handle `ENOENT` below.

Userland/Test.cpp
```diff
@@ -0,0 +1,540 @@
...
```
bugaevc, line 206: Same here

Userland/Test.cpp
```diff
@@ -0,0 +1,540 @@
...
+
+        struct stat statbuf;
+        int rc = lstat(m_path.characters(), &statbuf);
```
bugaevc, line 209: And, uh, why do we need to stat it here?

Userland/Test.cpp
```diff
@@ -0,0 +1,540 @@
...
+    {
+    }
+
+private:
+    virtual bool check() const override
+    {
+        if (m_mode == Equal)
+            return m_lhs == m_rhs;
+
+        return m_lhs != m_rhs;
```
bugaevc, line 253: Super nitpick: let's get rid of that blank line between the two branches

Base/usr/share/man/man1/test.md
emanuele6: there is no difference: https://www.man7.org/linux/man-pages/man1/test.1p.html

Userland/Test.cpp
```diff
@@ -0,0 +1,540 @@
...
+    } else if (arg == "(") {
+        optind++;
+        auto command = parse_complex_expression(argv);
+        if (command && argv[optind] && StringView(argv[++optind]) == ")")
+            return command;
+        fatal_error("Unmatched \033[1m(");
+    } else if (arg == "!") {
+        optind++;
+        auto command = parse_complex_expression(argv);
+        if (command)
```
bugaevc, line 376: I'd flip the condition, make the happy path straight

Userland/Test.cpp
```diff
@@ -0,0 +1,540 @@
...
+            return make<Not>(command.release_nonnull());
+        fatal_error("Expected an expression after \033[1m!");
+    } else if (arg == "-e") {
+        StringView name = argv[++optind];
+        if (name.is_null())
+            fatal_error("-e expects a file name");
```
bugaevc, line 382:
```suggestion
            fatal_error("-e requires a file name");
```
Perhaps?

Base/usr/share/man/man1/test.md
```diff
@@ -0,0 +1,86 @@
...
+* `-p FILE` (is pipe?)
+* `-S FILE` (is socket?)
+* `-h FILE` (is symbolic link?)
+* `-L FILE` (is symbolic link?)
+
+File permission checks
+
+* `-r FILE` (is readable?)
+* `-w FILE` (is writable?)
+* `-x FILE` (is executable?)
```
bugaevc, line 65: I feel like this is not communicated clear enough, neither in the man page nor in the code: this is not a check whether the file is readable/writable/executable, it's whether the *current user* has permissions to read/write/execute it.

Base/usr/share/man/man1/test.md
emanuele6: Since the `\`s are in a code block (\`\`\`), yes

Userland/Test.cpp
```diff
@@ -0,0 +1,540 @@
...
+private:
+    virtual bool check() const override
+    {
+        return !m_cond->check();
+    }
+
+    NonnullOwnPtr<Condition> m_cond;
+};
+
+class Exists : public Condition {
```
bugaevc, line 113: I'd actually expect this to be a variant of the permission checking condition, with `F_OK`

Userland/Test.cpp
```diff
@@ -0,0 +1,540 @@
...
+        if (name.is_null())
+            fatal_error("-w expects a file name");
+        return make<HasPermission>(name, HasPermission::Write);
+    } else if (arg == "-x") {
+        StringView name = argv[++optind];
+        if (name.is_null())
+            fatal_error("-x expects a file name");
+        return make<HasPermission>(name, HasPermission::Execute);
+    } else if (arg == "-a") {
+        return nullptr;
```
bugaevc, line 435: This here requires a comment for why we return `nullptr` 🙂


I suppose it's that "and" is not a simple command, so we're returning back to our caller to handle it.

Userland/Test.cpp
```diff
@@ -0,0 +1,540 @@
...
+        return make<StringCompare>("", value, StringCompare::NotEqual);
+    } else if (arg == "-z") {
+        StringView value = argv[++optind];
+        return make<StringCompare>("", value, StringCompare::Equal);
+    } else if (arg == "-g" || arg == "-G" || arg == "-k" || arg == "-N" || arg == "-O" || arg == "-s") {
+        fatal_error("Unsupported operator \033[1m%s", argv[optind]);
+    } else {
+        // Try to read a binary op.
+        auto lhs = arg;
+        arg = argv[++optind];
```
bugaevc, line 449: But what if the RHS of the binary operation is not a single argument, but a complex expression?

Base/usr/share/man/man1/test.md
bugaevc: Then let's make it clear in the man page that this is not a typo. For example, we could put them on one row like this:

* `-h FILE`, `-L FILE` (is a symbolic link?)

Base/usr/share/man/man1/test.md
bugaevc: Heh, I know what a code block is; I'm wondering if I messed up while implementing them in LibMarkdown.

emanuele6: yes, it's rendered properly.

bugaevc: Awesome 🙂

Base/usr/share/man/man1/test.md
```diff
@@ -0,0 +1,86 @@
...
+
+File existence check
+
+* `-e FILE` (does file exist?)
```
emanuele6, line 69:
```suggestion
* `-e FILE` (does FILE exist?)
```

bugaevc: Or perhaps let's drop all the SCREAMING and say `-e <file>`?

bugaevc:
> if I try to manually rename the target (I tried `set_target_properties(Test PROPERTIES RUNTIME_OUTPUT_NAME test)`), ninja complains that there are multiple rules that generate `Userland/test`, which I assume is the test target binary.

Hmm; we use `PROPERTIES OUTPUT_NAME` [for Lagom](https://github.com/SerenityOS/serenity/blob/master/Meta/Lagom/CMakeLists.txt#L71), and it worked... Can you try with `OUTPUT_NAME` just to be sure; and what are the two conflicting rules?

Userland/Test.cpp
```diff
@@ -0,0 +1,540 @@
...
+    }
+
+    return command;
+}
+
+int main(int, char* argv[])
+{
+    if (pledge("stdio rpath", nullptr) < 0) {
+        perror("pledge");
+        return 1;
```
emanuele6, line 532: Don't return 1 here, return something else!

Userland/Test.cpp
alimpfard: how would that work?
AFAIK these binary ops take single args around them, e.g. `file -ef file`, you can't put anything except a single argument in its place...

bugaevc: Ah, these aren't binary ops like "and" and "or", they're binary ops on two files, got it. Add a comment perhaps then?

alimpfard:
> Can you try with OUTPUT_NAME just to be sure

 Sure, same issue:
```
CMake Error:
  Running
   '/usr/bin/ninja' '-C' '/home/Test/Documents/buildground/serenity/Build' '-t' 'cleandead'

  failed with:
   ninja: error: build.ninja:22957: multiple rules generate Userland/test [-w dupbuild=err]

```

the two rules conflicting are (afaik from reading build.ninja) `CMakeFiles/test.util` and the one I added.
as I mentioned, `test.util` seems to be the binary for CTest.

bugaevc:
Ah, almost forgot: let's also support the `[ ]` syntax for it!

Userland/Test.cpp
alimpfard: Sure.

alimpfard:
> let's also support the [ ] syntax for it!

Alright, I'll make it require a trailing `]` if invoked as `[` then.

Base/usr/share/man/man1/test.md
```diff
@@ -0,0 +1,89 @@
+## Name
+
+test - checks files and compare values
+
+## Synopsis
+
+```**sh
+$ test EXPRESSION
```
bugaevc, line 8: Here, too, `[expression]`

Base/usr/share/man/man1/test.md
```diff
@@ -0,0 +1,89 @@
...
+
+### File permission checks
+
+* `-r <file>` (is readable?)
+* `-w <file>` (is writable?)
+* `-x <file>` (is executable?)
+* `-e <file>` (does file exist?)
+
+
+Except for `-h/-L`, all FILE checks dereference symbolic links.
```
bugaevc, line 71: And here, *file checks*

Base/usr/share/man/man1/test.md
```diff
@@ -0,0 +1,89 @@
...
+* `-c <file>` (is character device?)
+* `-f <file>` (is regular file?)
+* `-d <file>` (is directory?)
+* `-p <file>` (is pipe?)
+* `-S <file>` (is socket?)
+* `-h <file>`, `-L <file>` (is symbolic link?)
+
+### File permission checks
+
+* `-r <file>` (is readable?)
```
bugaevc, line 65: Perhaps these can be rewritten without the parens and question marks?

E.g. `-r <file>` — whether the file is readable

Userland/Test.cpp
```diff
@@ -0,0 +1,517 @@
...
+        , m_rhs(move(rhs))
+        , m_mode(mode)
+    {
+    }
+
+private:
+    virtual bool check() const override
+    {
+        struct stat statbuf_l;
+        int rc = lstat(m_lhs.characters(), &statbuf_l);
```
bugaevc, line 298: Are these checks supposed to dereference symlinks? Either way, it needs to be documented in the man page.

alimpfard: it _is_ documented "all FILE checks dereference symbolic links".

bugaevc: Yes, this is why I'm surprised to see this check *not* dereferencing symlinks, so I thought perhaps that only applies to that section on individual file checks, not on binary checks.

Then we need to fix it here in the code and make it more clear in the man page that this applies to both individual file checks and binary checks.

alimpfard: ...I think I was too sleepy when I wrote this...
I somehow managed to mix up `lstat` and `stat`...

bugaevc: No worries, that's what code review is for 😀

alimpfard:
Help a cmake idiot out, how do I make this thing copy the binary?
I doubt
```cmake
add_custom_command(
    TARGET Test POST_BUILD
    COMMAND ${CMAKE_COMMAND} -E copy
            ${CMAKE_CURRENT_BINARY_DIR}/Test
            ../Root/bin/[)
```
is what we want to do?

Base/usr/share/man/man1/test.md
alimpfard: I'm not sure how much more clear the code can get, it's literally called `HasPermission::Read`

bugaevc:
* Can you still name the source file `test.cpp` (and only the CMake target for it `Test`, or better `bin-test` or something)?
* I've previously worked around a similar problem [this way](https://github.com/darlinghq/darling-file_cmds/commit/f8e320088d9388d78017bf510ef68e19684fddf8)

bugaevc:
Oh, oh, you're asking about the `/bin/[` thing, I see. Don't copy it, just make a symlink.

alimpfard:
> Can you still name the source file test.cpp

Nah, the foreach...add_executable gets mad that "test is a special target for CTest, you can't do this"

> I've previously worked around a similar problem this way

I'll try moving the file then, not sure if that will work though.

> just make a symlink

Well, sure, but my question was more like "is there a better way to specify the target than ../Root/bin/["

Base/usr/share/man/man1/test.md
bugaevc: "Has permission" still sounds like we're checking if the file has this permission (or at least it doesn't sound sufficiently unlike that). `UserHasPermission` or something.

bugaevc:
No, put the symlink [here](https://github.com/SerenityOS/serenity/blob/master/Meta/build-root-filesystem.sh#L165)

alimpfard:
Ah, sure.

bugaevc:
> Nah, the foreach...add_executable gets mad that "test is a special target for CTest, you can't do this"

I'd still name the file `test.cpp`, and special-cased it in the build script — for each source file... if the name is "test" or "install", do this, otherwise do (generic thing).

alimpfard:
I guess...kinda dislike adding branches like that, but you're the cmake guy here :laughing:

Base/usr/share/man/man1/test.md
```diff
@@ -0,0 +1,92 @@
+## Name
+
+test - checks files and compare values
+
+## Synopsis
+
+```**sh
+$ test [expression]
+$ [ expression ]
```
emanuele6, line 9:
```suggestion
$ [ [expression] ]
```

alimpfard: that's _so_ confusing though...
I'll add `[ ]` instead.

bugaevc:
I can't say I like those hacks either, but there point is to keep hacks that workaround buildsystem shortcomings in the buildsystem 🙂

Base/usr/share/man/man1/test.md
bugaevc: Yeah that is becoming confusing, maybe let's just drop the inner brackets altogether for this synopsis?

Userland/Test.cpp
```diff
@@ -0,0 +1,534 @@
...
+}
+
+int main(int, char* argv[])
+{
+    if (pledge("stdio rpath", nullptr) < 0) {
+        perror("pledge");
+        return 126;
+    }
+
+    if (StringView { argv[0] } == "[")
```
emanuele6, line 525: We should use this instead check instead.

This way `/bin/[ 1 ]` works as expected

```suggestion
    StringView arg0(argv[0]);
    if (arg0 == "[" || arg0.ends_with("/["))
```

Base/usr/share/man/man1/test.md
alimpfard: Then I'll make it also say `$ test`, for completeness.

Userland/Test.cpp
emanuele6: Is there a function to get the filename out of a path?

If there is, we could also write this as:
```cpp
if (function_that_gets_filename_from_path(StringView(argv[0])) == "[")
```

bugaevc: https://github.com/SerenityOS/serenity/blob/master/Base/usr/share/man/man3/basename.md, or a similar thing in `AK::LexicalPath`

Base/usr/share/man/man1/test.md
emanuele6: > Then I'll make it also say `$ test`, for completeness.

You should also add `$ [ ]` then.

Userland/Test.cpp
alimpfard: I guess we can use `LexicalPath{argv[0]}.basename()`?

bugaevc: Yep

Base/usr/share/man/man1/test.md
```diff
@@ -0,0 +1,94 @@
...
+$ [ ]
+```
+
+## Description
+
+`test` takes a given expression and sets the exit code according to its truthiness, 0 if true, 1 if false.
+An omitted expression defaults to false, and an unexpected error causes an exit code of 126.
+
+If `test` is invoked as `[`, a trailing `]` is _required_ after the expression, and everything after the `]` is ignored.
```
emanuele6, line 19: I haven't read the implementation yet, but that's not how it's supposed to work:

```sh
[ -n ] ]
```
This should return 0 because the length of the `]` string is non-zero.

alimpfard: Hmmm...so parse away until the end and then check?
it seems `[` is fine with `[ -n ]`, so I'll just make it do that instead.

Userland/test.cpp
```diff
@@ -0,0 +1,524 @@
...
+
+int main(int, char* argv[])
+{
+    if (pledge("stdio rpath", nullptr) < 0) {
+        perror("pledge");
+        return 126;
+    }
+
+    if (LexicalPath { argv[0] }.basename() == "[")
+        g_was_invoked_as_bracket = true;
```
emanuele6, line 516: Instead of using this flag, how about we do this:
```suggestion
    if (LexicalPath { argv[0] }.basename() == "[") {
        if (StringView { argv[argc - 1] } != "]")
            fatal_error("test invoked as '[' requires a closing bracket ']'");
        argv[argc - 1][0] = '\0';
    }
```
and let the program parse the arguments as if it was invoked as `test`?

Userland/test.cpp
```diff
@@ -0,0 +1,524 @@
...
+{
+    fputs("\033[31m", stderr);
+
+    va_list ap;
+    va_start(ap, format);
+    vfprintf(stderr, format, ap);
+    va_end(ap);
+
+    fputs("\033[0m\n", stderr);
+    exit(1);
```
emanuele6, line 53: Don't exit 1 here

Userland/test.cpp
emanuele6: I was thinking of this solution for the `[ -n ] ]` thing @alimpfard .

emanuele6: With this method we also don't have to care about `g_was_invoked_as_bracket`

alimpfard: That's...actually pretty nice, yeah.

Base/usr/share/man/man1/test.md
```diff
@@ -0,0 +1,94 @@
...
+None.
+
+## Examples
+
+```sh
+# Conditionally do something based on the value of a variable
+$ test "$foo" = bar && echo foo is bar
+# Check some numbers
+$ test \( 10 -gt 20 \) -o \( ! 10 -ne 10 \) && echo "magic numbers!"
+```
```
emanuele6, line 90: These examples are not great: our shell, for POSIX compliance, will need to have `[` and `test` as shell builtin.

If anuser were to try to run these examples in a POSIX compliant shell, these examples would not test the `/bin/test` utility, but rather the `test` shell builtin.

We should either not put `Examples` in this `man` page or replace `test` with `/bin/test` or `\test` (if we support it).

Maybe we don't need to care about this now, though.

alimpfard: Or we could just slap a note on it that says "[ and test might be shell builtins"?
we don't support the `\foo` thing (yet?).

emanuele6: > we don't support the \foo thing (yet?).

I don't know? Does the shell even have any bultins?

alimpfard: Yeah, it has...
```
    __ENUMERATE_SHELL_BUILTIN(cd)      \
    __ENUMERATE_SHELL_BUILTIN(cdh)     \
    __ENUMERATE_SHELL_BUILTIN(pwd)     \
    __ENUMERATE_SHELL_BUILTIN(exit)    \
    __ENUMERATE_SHELL_BUILTIN(export)  \
    __ENUMERATE_SHELL_BUILTIN(unset)   \
    __ENUMERATE_SHELL_BUILTIN(history) \
    __ENUMERATE_SHELL_BUILTIN(umask)   \
    __ENUMERATE_SHELL_BUILTIN(dirs)    \
    __ENUMERATE_SHELL_BUILTIN(pushd)   \
    __ENUMERATE_SHELL_BUILTIN(popd)    \
    __ENUMERATE_SHELL_BUILTIN(time)    \
    __ENUMERATE_SHELL_BUILTIN(jobs)    \
    __ENUMERATE_SHELL_BUILTIN(disown)  \
    __ENUMERATE_SHELL_BUILTIN(fg)      \
    __ENUMERATE_SHELL_BUILTIN(bg)
```

and soon `alias`, once my shell PR gets merged.

bugaevc: > These examples are not great: our shell, for POSIX compliance, will need to have [ and test as shell builtins.

It has been decided that our Shell has nothing to do with POSIX, and we come up with our own semantics that we believe to make sense. If the user needs a Bourne-style shell, they can use the GNU Bash port.

emanuele6:
`test 1 -o 2 -gt 3` is not getting parsed correctly:

```sh
$ test 1 -o 2 -gt 3
Unsupported option -o
```

alimpfard:
but...a single integer is not a valid expression?

emanuele6:
It's not "a single integer": it's a string of non-null length and should evaluate to `true`.

It should be interpreted as if it was: `test -n 1 -o 2 -gt 3`

>-n STRING:    the length of STRING is nonzero
>STRING:    equivalent to -n STRING

Base/usr/share/man/man1/test.md
```diff
@@ -0,0 +1,96 @@
...
+* `<expression> -a <expression>` boolean conjunction of the values
+* `<expression> -o <expression>` boolean disjunction of the values
+
+### String comparison
+
+* `-n <string>` whether the string is non-empty
+* `-z <string>` whether the string is empty
+* `<string> = <string>` whether the two strings are equal
+* `<string> != <string>` whether the two strings not equal
+* `<string>` whether the string is non-empty, only allowed when it is the only expression specified
```
emanuele6, line 41: This is wrong, `<string>` is always a valid expression.

alimpfard: Fixed. it was just misunderstanding on my part.

Base/usr/share/man/man1/test.md
```diff
@@ -0,0 +1,96 @@
...
+* `<expression> -a <expression>` boolean conjunction of the values
+* `<expression> -o <expression>` boolean disjunction of the values
+
+### String comparison
+
+* `-n <string>` whether the string is non-empty
+* `-z <string>` whether the string is empty
+* `<string> = <string>` whether the two strings are equal
+* `<string> != <string>` whether the two strings not equal
+* `<string>` whether the string is non-empty
```
emanuele6, line 41: put it before `-n <string>` since they do the same thing they should be near each other.

emanuele6:
```sh
$ test -o 4 -gt 3
$ echo $?
1
$ # expected 126 (missing expression before `-o`)
$ test -o
$ echo $?
1
$ # expected 0 (non-null string)
$ test '' -o 3 -gt 4
$ echo $?
1
$ # expected 1 (correct)
$ test '' -o 6 -gt 4
$ echo $?
1
$ # expected 0
```

alimpfard:
Does the current shell parser handle `''` correctly? 
I'll take a look at the `-o ..` case

emanuele6:
> Does the current shell parser handle '' correctly?
Yes, it does: I wrote a simple program to test it:
```cpp
// test_shell.cpp
int main(int argc, char** argv)
{
    for (int i = 0; i < argc; ++i)
        printf("%d: %s\n", i, argv[i]);
    return 0;
}
```

```
$ test_shell '' -o 4 -gt 3
0: test_shell
1: 
2: -o
3: 4
4: -gt
5: 3
```

alimpfard:
Fixed.
I ran your tests, it seems to behave ok now.

the parser was eating the `-o` whenever it was encountered.

Userland/test.cpp
```diff
@@ -0,0 +1,525 @@
...
+        return 126;
+    }
+
+    if (LexicalPath { argv[0] }.basename() == "[") {
+        --argc;
+        if (StringView { argv[argc] } != "]")
+            fatal_error("test invoked as '[' requires a closing bracket ']'");
+        argv[argc] = nullptr;
+    }
+
```
emanuele6, line 518: Oh, yeah, my bad... xd.

alimpfard: lol nothing to worry about~

emanuele6:
```sh
$ test = 2
$ echo $?
1
$ # expected 126 (missing string before `=`)
```

Meta/build-root-filesystem.sh
```diff
@@ -163,6 +163,7 @@ ln -s ProfileViewer mnt/bin/pv
 ln -s WebServer mnt/bin/ws
 ln -s Solitaire mnt/bin/sl
 ln -s WebView mnt/bin/wv
+ln -s test mnt/bin/\[
```
emanuele6, line 166:
```suggestion
ln -s test mnt/bin/[
```
We don't need a backslash here.

emanuele6, line 166: This isn't really a shortcut.

emanuele6:
I think that `ln -s test /mnt/bin[` fits better here:
```diff
diff --git a/Meta/build-root-filesystem.sh b/Meta/build-root-filesystem.sh
index 51928b1a0..b88b164e7 100755
--- a/Meta/build-root-filesystem.sh
+++ b/Meta/build-root-filesystem.sh
@@ -50,6 +50,8 @@ chmod 4750 mnt/bin/reboot
 chmod 4750 mnt/bin/shutdown
 chmod 4750 mnt/bin/keymap
 
+ln -s test mnt/bin/[
+
 echo "done"
 
 printf "creating initial filesystem structure... "
@@ -163,7 +165,6 @@ ln -s ProfileViewer mnt/bin/pv
 ln -s WebServer mnt/bin/ws
 ln -s Solitaire mnt/bin/sl
 ln -s WebView mnt/bin/wv
-ln -s test mnt/bin/\[
 echo "done"
 
 # Run local sync script, if it exists
```

Meta/build-root-filesystem.sh
alimpfard: It was suggested by @bugaevc, so...what say you, @bugaevc?

alimpfard:
> test = 2 shouldn't exit with 0

Makes sense. it wasn't making sure that all input was exhausted.

Meta/build-root-filesystem.sh
emanuele6: `[` is a shell builtin, not a special character like `(`, `)`, `;`, `&`, `\`, `|`, etc.

You don't need to escape builtins: `echo alias` works fine, you don't need to use `echo \alias`

bugaevc: But `[` is not a shell builtin in our case. And I do think `[` can be considered a shortcut for `test` (although an unusual one, because it requires the matching `]`).

emanuele6: > [ is a shell builtin, not a special character like (, ), ;, &, |, etc.
> You don't need to escape builtins: echo alias works fine, you don't need to use echo \alias

I was talking about this change @bugaevc .

alimpfard: So what's the verdict? should that `ln` stay where it is?
I have no particular preference, so whatever y'all decide on is fine.

emanuele6: I think that `[` shouldn't be considered a shortcut for `test` since it has a different behaviour.

bugaevc: Yeah, here you don't need a backslash of course, `[` is just a character like any other; it's just that some shells have a builtin named `[`

alimpfard: Although even Dr. POSIX says it's fine, and that `[` is a "second form" of `test`.
> https://pubs.opengroup.org/onlinepubs/9699919799/utilities/test.html
> In the second form of the utility, where the utility name used is [ rather than test, [...]. The test and [ utilities may be implemented as a single linked utility

bugaevc: Of course, that's how it works everywhere. It wasn't me who came up with symlinking it this way :smile:

emanuele6: I'm not against symlinking!
I just think that `ln -s test mnt/bin/[` fits better in the `installing base system... ` section rather than in the `installing shortcuts... ` section.

https://github.com/SerenityOS/serenity/pull/2618#issuecomment-649044754

emanuele6:
`test -n` shouldn't fail and should exit with 0.

```sh
$ test -n
Too many arguments
$ echo $?
126
$ # expected 0 (the `-n` string has non-zero length)
```

alimpfard:
Uhh...why?
Being passed an empty string is very different from being passed no string

emanuele6:
`-n` is a STRING, a STRING in this context.

alimpfard:
that's...very confusing.

emanuele6:
`test -n` is the same as `test -n -n`, in the same way that `test "$anything"` is the same as `test -n "$anything".

emanuele6:
`test -n -a -n` should also be valid:
$1 = `-n`
$2 = `-a`
$3 = `-n`

$1 is a string and evaluates to `true` since `-n` has non-zero length. -> true
$3 is a string and evaluates to `true` since `-n` has non-zero length. -> true

(true) -a (true) -> true

`test` should return 0.

alimpfard:
Hmmm...
Honestly, I'd say those are super niche uses, and frankly, very confusing.
perhaps we should just disallow plain strings as expressions?

emanuele6:
> Hmmm...
Honestly, I'd say those are super niche uses, and frankly, very confusing.
perhaps we should just disallow plain strings as expressions?

not really...

take this shell script for example: (`echo hello friend | /path/to/script` -> `h e l l o  f r i e n d `)
```sh
#!/bin/sh
read a
test "$a" || exit 1
echo "$a" | sed 's,\w,& ,g'
```

emanuele6:
> perhaps we should just disallow plain strings as expressions?

we could do that if we don't care POSIX-compliance, but why? we could it doesn't seem that confusing or difficult to implement to me.

alimpfard:
> it doesn't seem that confusing or difficult to implement to me

Not really about the difficulty, it just sounds like a mess of arg processing.

alimpfard:
Well whatever, there's no particular reason to be stingy about it, I'm sure someone would be surprised that `test -n` is bitching at them :shrug:

bugaevc:
> we could do that if we don't care about POSIX-compliance, but why? it doesn't seem that confusing or difficult to implement to me.

(In my opinion,) we don't care about *POSIX compliance* specifically — we're never gonna be fully POSIX compliant, and it would bring us no benefits to be so. That means we are under no obligation to implement things exactly the way POSIX specifies them.

But we are a Unix-like system, and in many cases it makes sense to implement the same APIs (and in this case, command line tools) that other Unix-like systems already have. I see three reasons for doing that:
1. Why the heck not, if the APIs make sense
2. Other Unix-like systems, and POSIX the spec, already have many kinks and fine details figured out
3. If many other systems implement the same APIs, chances are programs support and use those APIs. If we implement them too, it'd be easier to port those programs to Serenity.

Clearly not all APIs quality. A good example is the recent `sendfd()`/`recvfd()` APIs: the alternative APIs that are implemented by Linux and others, and specified by POSIX, are ugly, and it should be easy to port any programs that use them to use ours instead.

An example of where we do follow POSIX: we've changed how `mkdir(2)` treats trailing slashes, for reasons number 3 (ported software — namely git — wanted it that way), and number 2 (POSIX people clearly have done some research figuring out the best way to handle this exact case, so let's leverage their insight).

We also usually try to read the POSIX man pages rather than util-linux man pages when working on a feature — not of a particular respect of POSIX, and not because of feeling obligated to implement everything exactly as POSIX specifies, but because of a desire to be a Unix-like system, not a Linux clone.

So to recap: POSIX is a good source of ready-made solutions to tricky cases, and it frequently makes sense for us to follow POSIX, but we are under no obligation to do everything exactly as POSIX specifies, nor be "compliant".

(Paging in @awesomekling, perhaps he can add something or disagree with me 😀)

awesomekling:
> (Paging in @awesomekling, perhaps he can add something or disagree with me )

I agree with everything you said. The examples are good, too. `SCM_RIGHTS` is a hilariously unpleasant API compared to `sendfd()`/`recvfd()` :)

awesomekling:
That's a really nice implementation of `test`. Just the right mix of C++ and Unix that we should aim for in our Userland code IMO :)

PR #2619, opened 2020-06-24
Meta: Remove default user name from Meta/CLion/run.sh

devsh0:
`Meta/CLion/run.sh` assumes the user name 'Ragnarok' and sets build directory accordingly. It is a left over from when I was setting up the project for my use. This patch also updates WSL build docs that seem to suggest that `run.sh` requires no modification if serenity is being built on a 64 bit machine :^)

### DIFF
--- a/Documentation/NotesOnWSL.md
+++ b/Documentation/NotesOnWSL.md
@@ -39,6 +39,6 @@ In a 64-bit machine, it's located at `/mnt/c/Program Files/qemu/qemu-system-i386
 
 - Edit `serenity/Meta/CLion/run.sh`. Set **SERENITY_QEMU_BIN**  to point to the windows installation of `qemu-system-i386.exe`.
 Also verify that the value of **SERENITY_BUILD** is valid.
-In a 64-bit machine, if qemu was installed in the default location you shouldn't need to alter anything.
+In a 64-bit machine, if qemu was installed in the default location you should only need to alter user name in the script.
 
 - Execute `serenity/Meta/CLion/run.sh`.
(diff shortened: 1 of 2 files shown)


emanuele6:
That's weird: `LibCrypto`'s unit test failed.

Anyhow, can you please make this pull request [here](https://github.com/emanuele6/serenity/tree/meta_work) instead of `SerenityOS:master`: I modified `Meta/CLion/run.sh` to use `Meta/run.sh` so that we don't have to maintain two `run.sh`.

Also, can you check if using the `$USER` environment variable works in windows? i.e.

Edit this line in `Meta/CLion/run.sh` (NB: using `"` instead of `'`):

```bash
: "${SERENITY_ROOT:="/mnt/c/Users/$USER/serenity-project/serenity"}"
```

alimpfard:
> LibCrypto's unit test failed

It does that sometimes
I haven't found out why yet.
if only cmake showed the output of failing tests...

devsh0:
Closing this PR to make the changes to meta_work.

> Also, can you check if using the $USER environment variable works in windows?

I don't use Windows anymore. I noticed the user name while going through the changes in your PR. But I will substitute `$USER` anyways to see if that satisfies travis.

devsh0:
@emanuele6 For some reason I cannot select your fork to open pull requests. I even tried by creating a new repository with your fork as the base. Can you just make the change locally and update the PR? Note that if substituting `$USER` is supposed to work, you wouldn't need to change the WSL documentation. So basically, just replace `Ragnarok` with `$USER` and push.

emanuele6:
> Note that if substituting $USER is supposed to work, you wouldn't need to change the WSL documentation. So basically, just replace Ragnarok with $USER and push.

If you aren't sure about USER working, let's not change it.

> I don't use Windows anymore. I noticed the user name while going through the changes in your PR.

If you don't use CLion on Windows anymore and nobody else uses it, there is no point in updating the documentation imo (especially since we can't test what we write in the documentation).

devsh0:
Fair enough :)

PR #2620, opened 2020-06-24
Lagom: enable building and running AK tests

predmond:
Add enable_testing() to the Lagom CMakeLists.txt so you can do
`ninja test` to run tests. Also add the AK/Tests directory to
build the AK testsuite.

### DIFF
--- a/Meta/Lagom/CMakeLists.txt
+++ b/Meta/Lagom/CMakeLists.txt
@@ -1,6 +1,8 @@
 cmake_minimum_required (VERSION 3.0)
 project (Lagom)
 
+enable_testing()
+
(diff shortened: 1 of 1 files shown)


awesomekling:
cc @bugaevc

bugaevc:
We already do it in the root `CMakeLists.txt`, so why do it again here?

Or are you trying to build Lagom without going through the rest of buildsystem?

predmond:
I didn't realize Lagom was included by the top-level file. I was still building it by running cmake from Meta/Lagom/build 😬 

My primary goal was to run the AK tests on my host. I didn't realize this already worked. Sorry!

PR #2623, opened 2020-06-25
Help: Add a button to open the man page on github for editing

bgianfo:
Inspired by Sphinx, Cargo, MSDN, etc, add the ability to jump
to the github file edit UI in the browser. The idea being that
users can quickly and easily contribute fixes.

### DIFF
--- a/Applications/Help/main.cpp
+++ b/Applications/Help/main.cpp
@@ -42,6 +42,7 @@
 #include <LibGUI/ToolBarContainer.h>
 #include <LibGUI/TreeView.h>
 #include <LibGUI/Window.h>
+#include <LibDesktop/Launcher.h>
 #include <LibMarkdown/Document.h>
 #include <LibWeb/Layout/LayoutNode.h>
 #include <LibWeb/PageView.h>
@@ -102,6 +103,9 @@ int main(int argc, char* argv[])
 
     RefPtr<GUI::Action> go_back_action;
     RefPtr<GUI::Action> go_forward_action;
+    RefPtr<GUI::Action> edit_on_github;
+
+    String last_path;
 
     auto update_actions = [&]() {
         go_back_action->set_enabled(history.can_go_back());
@@ -124,7 +128,9 @@ int main(int argc, char* argv[])
             GUI::MessageBox::show(strerror(saved_errno), "Failed to open man page", GUI::MessageBox::Type::Error, GUI::MessageBox::InputType::OK, window);
             return;
         }
+        last_path = path;
         auto buffer = file->read_all();
+
         StringView source { (const char*)buffer.data(), buffer.size() };
 
         String html;
@@ -185,6 +191,25 @@ int main(int argc, char* argv[])
 
     toolbar.add_action(*go_back_action);
     toolbar.add_action(*go_forward_action);
+
+    edit_on_github = GUI::Action::create("Edit On GitHub", Gfx::Bitmap::load_from_file("/res/icons/16x16/filetype-markdown.png"), [&](const GUI::Action&) {
+
+        if (last_path.is_null()) {
+            return;
+        }
+
+        // Open a browser to the man page on github.
+        StringBuilder builder;
+        builder.append("https://github.com/SerenityOS/serenity/blob/master/Base");
+        builder.append(last_path);
+
+        auto edit_on_gh = builder.build();
+        Desktop::Launcher::open(edit_on_gh);
+    });
+
+    edit_on_github->set_enabled(true);
+    toolbar.add_separator();
+    toolbar.add_action(*edit_on_github);
 
     auto menubar = GUI::MenuBar::construct();
 
--- a/Applications/Help/CMakeLists.txt
+++ b/Applications/Help/CMakeLists.txt
@@ -7,4 +7,4 @@ set(SOURCES
 )
 
 serenity_bin(Help)
-target_link_libraries(Help LibWeb LibMarkdown LibGUI)
+target_link_libraries(Help LibWeb LibMarkdown LibGUI LibDesktop)


bgianfo:
This being actually useful kind of depends on the browser being able to render github, but I figured it's good to have a reason to get there :).

Applications/Help/main.cpp
```diff
@@ -51,14 +51,14 @@
...
-    if (pledge("stdio shared_buffer accept rpath unix cpath fattr", nullptr) < 0) {
+    if (pledge("stdio shared_buffer accept exec rpath unix cpath fattr", nullptr) < 0) {
         perror("pledge");
         return 1;
     }
 
     GUI::Application app(argc, argv);
 
-    if (pledge("stdio shared_buffer accept rpath", nullptr) < 0) {
+    if (pledge("stdio shared_buffer accept exec rpath", nullptr) < 0) {
```
alimpfard, line 61: could we use LibDesktop's `open(URL)` instead of running the browser directly?

bgianfo: Good idea, done!

bugaevc:
Perhaps it would have been a good idea to request a review from the guy who actually wrote the app, or at least cc him? 😉

Now to be honest, I'm not sure this is a good thing to have. In particular I wouldn't encourage anyone to edit code in the browser.

alimpfard:
@bugaevc we should just make all PRs cc you by default :^)

bugaevc:
That's the running joke, yeah. But in reality, I tried subscribing to all notifications for the repo last year, and only lasted a few days — there's just too much noise, because there are all these things I don't know enough about / I'm not interested enough in (for example, anything related to hardware, and LibJS).

So all I ask is to be cc'ed in *relevant* PRs. Example: I would cc *you* in PRs having to do with LibLine, the shell, crypto, HTTPS, and all things ProtocolServer, and probably in new userland tools. And eyes. And anything else I can remember (or see in git blame) you've been hacking on.

bgianfo:
> Perhaps it would have been a good idea to request a review from the guy who actually wrote the app, or at least cc him? 😉
Sure my apologies, I normally let kling assign reviewers as he see's fit.
Maybe we should consider setting up CODEOWNERS if you want to be notified by the system.
See: https://github.blog/2017-07-06-introducing-code-owners/ 

> Now to be honest, I'm not sure this is a good thing to have. In particular I wouldn't encourage anyone to edit code in the browser.
Why is editing markdown in the browser to be discouraged? I've used it successfully for small drive by PR's across different repo's in the past. 
IMHO documentation with a clear path towards how to fix it when you find something wrong, is better maintained.

alimpfard:
> I tried subscribing to all notifications for the repo last year, and only lasted a few days

I imagine it would be a pain, yeah.
Although recently github has stopped sending me notifications, I just come here and manually look through the PRs and issues...

linusg:
>  Maybe we should consider setting up CODEOWNERS if you want to be notified by the system.

this came up before, the issue with CODEOWNERS is, emphasis mine:

> **The people you choose as code owners must have write permissions for the repository.** When the code owner is a team, that team must have write permissions, even if all the individual members of the team already have write permissions directly, through organization membership, or through another team membership.

awesomekling:
> Now to be honest, I'm not sure this is a good thing to have. In particular I wouldn't encourage anyone to edit code in the browser.

I agree. The implementation is fine, but I think the feature itself is a bit un-Serenity-like. We want the system to encourage use of its own tools for authoring whenever possible (the built-in TextEditor app even has live Markdown preview!) :)

bgianfo:
> 
> 
> > Maybe we should consider setting up CODEOWNERS if you want to be notified by the system.
> 
> this came up before, the issue with CODEOWNERS is, emphasis mine:
> 
> > **The people you choose as code owners must have write permissions for the repository.** When the code owner is a team, that team must have write permissions, even if all the individual members of the team already have write permissions directly, through organization membership, or through another team membership.

Ah, sure, I think with github actions people have been able to get around this.
For example: https://github.com/marketplace/actions/auto-assign-reviewer-by-files

bgianfo:
> 
> 
> > Now to be honest, I'm not sure this is a good thing to have. In particular I wouldn't encourage anyone to edit code in the browser.
> 
> I agree. The implementation is fine, but I think the feature itself is a bit un-Serenity-like. We want the system to encourage use of its own tools for authoring whenever possible (the built-in TextEditor app even has live Markdown preview!) :)

👍

bugaevc:
Disclaimer, sorta: this is my personal opinion; let's see if others feel the same way. And it's certainly not meant to offend anyone, or discourage anyone from contributing. Really 🙂

It just may be that I dislike and oppose all things "modern web" (aka let's send the user a shitton of minified JavaScript to imitate a native app, badly). The web was meant to be, and should be, a network of *rich documents*, with hyperlinks between them — not another platform for apps.

So when it comes to editing source code, why not use an actual text editor? Like ~vim~ glorious GNU Emacs, or our own TextEditor? We even have live Markdown preview there 😀! Once you're done editing the code, you should commit it, with Git — not whatever UI GitHub surfaces on the web. Oh and yeah, you should check whether it's being rendered by LibMarkdown correctly (or, the way you expect it to be) — if you're using the live preview in TextEditor, that basically happens automatically; I don't think there's a way to do this from GitHub UI.

Using a proper text editor + git workflow is just sort of like... hygiene. Of course I'm not gonna judge anyone if they find using GitHub's web UI more convenient in certain cases, but it seems wrong to suggest people to do that.

Lastly, I think it's important for software projects to not be tied to GitHub (or any other code hosting service), or in other words, to remain agnostic to some Salas being used in its development. We host the main repo on GitHub, that's true (although it can be argued that the main repo is in fact kling's local tree), and we track issues/PRs here — but this is not reflected in the features of the project anywhere. If GitHub were to disappear, we could simply use another platform, the system itself remains agnostic to that. (Though it can be argued that it wouldn't be too hard to just change this URL in this unlikely case).

Yeah, so these are my reasons for not wanting to link people to GitHub's web editor. How about we add an option to open the man page in TextEditor instead? 😀

bgianfo:
> Disclaimer, sorta: this is my personal opinion; let's see if others feel the same way.

I see where you are coming from. 

> How about we add an option to open the man page in TextEditor instead? 😀

Once you've opened the man page in TextEditor and made your changes you now have an edited file on disk.
The file is inside your VM image, which the user now has to figure out how to get upstream, so they have to mount
their disk image and then copy it out into their source tree? I don't really see how that's useful, or something you would
recommend to users. Such a feature only makes sense to me in a future where the serenity VM image comes with it's use-able own clone of the source, or the OS is self hosting and the Help app could detect where your source is located.

I'm perfectly happy just abandoning the idea. 🥰

bugaevc:
> The file is inside your VM image, which the user now has to figure out how to get upstream, so they have to mount their disk image and then copy it out into their source tree? I don't really see how that's useful, or something you would recommend to users.

Well, that is useful and is something I'd recommend to users (Serenity developers, that is). It's true that it's not super straightforward.

But if we're talking about the bright future where GitHub is usable in the Browser, why can't the user commit the modified file to the repo (git works already) and open a PR, all from inside Serenity? 😀

bgianfo:
> But if we're talking about the bright future where GitHub is usable in the Browser, why can't the user commit the modified file to the repo (git works already) and open a PR, all from inside Serenity? 😀

They can, and they should, I look forward that reality.

PR #2625, opened 2020-06-25
Kernel: Deemphasize inode identifiers

bugaevc:
These APIs were clearly modeled after Ext2FS internals, and make perfect sense in Ext2FS context. The new APIs are more generic, and map better to the semantics exported to the userspace, where inode identifiers only appear in `stat()` and `readdir()` output, but never in any input.

This will also hopefully reduce the potential for races (see commit https://github.com/SerenityOS/serenity/commit/c44b4d61f350703fcf1bbd8f6e353b9c6c4210c2).

Lastly, this makes it way more viable to implement a filesystem that only synthesizes its inodes lazily when queried, and destroys them when they are no longer in use. With inode identifiers being used to reference inodes, the only choice for such a filesystem is to persist any inode it has given out the identifier for, because it might be queried at any later time. With direct references to inodes, the filesystem will know when the last reference is dropped and the inode can be safely destroyed.

### DIFF
--- a/Kernel/FileSystem/Ext2FileSystem.cpp
+++ b/Kernel/FileSystem/Ext2FileSystem.cpp
@@ -150,9 +150,9 @@ const char* Ext2FS::class_name() const
     return "Ext2FS";
 }
 
-InodeIdentifier Ext2FS::root_inode() const
+NonnullRefPtr<Inode> Ext2FS::root_inode() const
 {
-    return { fsid(), EXT2_ROOT_INO };
+    return *get_inode({ fsid(), EXT2_ROOT_INO });
 }
 
 bool Ext2FS::find_block_containing_inode(unsigned inode, unsigned& block_index, unsigned& offset) const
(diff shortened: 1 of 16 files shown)


Kernel/FileSystem/Ext2FileSystem.cpp
```diff
@@ -1405,11 +1409,6 @@ KResultOr<NonnullRefPtr<Inode>> Ext2FS::create_inode(InodeIdentifier parent_id,
         return KResult(-ENOSPC);
     }
 
-    // Try adding it to the directory first, in case the name is already in use.
```
bugaevc, line 1408: Note: this is already taken care of by VFS.

bugaevc:
(Wait, don't merge this yet)

bugaevc:
Fixed, can be merged now.

awesomekling:
Very sensible changes! Thank you for working on this :)

PR #2627, opened 2020-06-25
Travis: Limit notifications to SerenityOS/serenity repo

linusg:
I hope this works.

Docs:

- https://docs.travis-ci.com/user/notifications/#conditional-notifications
- https://docs.travis-ci.com/user/conditions-v1#integration

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,6 +13,7 @@ cache:
 
 notifications:
   irc:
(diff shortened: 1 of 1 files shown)


linusg:
That should solve the problem for the future. Can't do anything about existing repos I'm afraid, unless they update ofc.

awesomekling:
Let's try it :)

PR #2629, opened 2020-06-25
LibJS: run-test.sh emits test output if it is not "PASS"

mattco98:
Previously, debugging a test with `console.log` statements was impossible, because it would just cause the test to fail with no additional output. Now, if the output of a test is not "PASS", the output will be printed under the line where the test failed.

Empty output will have a special message attached to it -- useful when a test author has forgotten to include `console.log("PASS")` at the end of a test.

Preview:

preview

### DIFF
--- a/Libraries/LibJS/Tests/run-tests.sh
+++ b/Libraries/LibJS/Tests/run-tests.sh
@@ -31,6 +31,22 @@ for f in *.js; do
     fi
     echo -ne "\033]9;${count};${test_count}\033\\"
     echo "$f"
+
+    if [ "$result" != "PASS" ]; then
+        if [ -z "$result" ]; then
+            echo -e "    \033[31;1mNo output. Did you forget 'console.log(\"PASS\");'?\033[0m"
+        else
+            readarray -t split_result <<< "$result";
+
+            echo -ne "    \033[31;1mOutput:\033[0m: "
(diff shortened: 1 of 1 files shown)


Libraries/LibJS/Tests/run-tests.sh
```diff
@@ -31,6 +31,22 @@ for f in *.js; do
...
     echo -ne "\033]9;${count};${test_count}\033\\"
     echo "$f"
+
+    if [ "$result" != "PASS" ]; then
+        if [ -z "$result" ]; then
+            echo -e "    \033[31;1mNo output. Did you forget 'console.log(\"PASS\");'?\033[0m"
+        else
+            readarray -t split_result <<< "$result";
+
+            echo -ne "    \033[31;1mOutput:\033[0m: "
```
linusg, line 41: Is the double colon on purpose?

linusg:
Thank you for this! Makes finding the one failing test in between the twohundredsomething we have now a lot easier, especially for colourblind people :)

mattco98:
> Thank you for this! Makes finding the one failing test in between the twohundredsomething we have now a lot easier, especially for colourblind people :)

Np! I just decided today was the last time I am going to spend 10 minutes debugging a test only to find I forgot that last console statement.

awesomekling:
Great idea!

PR #2630, opened 2020-06-25
LibJS: Implement Proxy [[Call]] and [[Construct]] traps

### LINKED ISSUE #2624: LibJS: Proxy objects with function targets behave incorrectly
Expected behaviour:

```js
> function f() { return 42; }
undefined
> const p = new Proxy(f, {});
undefined
> typeof p
"function"
> p()
42
```

Actual behaviour:

```js
> function f() { return 42; }
undefined
> const p = new Proxy(f, {})
undefined
> typeof p
"object"
> p()
Uncaught exception: [TypeError]: [object ProxyObject] is not a function (evaluated from 'p')
```

cc @mattco98

mattco98:
In order to do this, Proxy now extends `Function` rather than `Object,` and whether or not it returns true for `is_function()` depends on it's `m_target`. 

A couple of other changes introduced in this PR:
- `Function::construct` now takes a `Function& new_target`, in order to facilitate the Proxy passing into it's handler's`[[Construct]]` trap.
- `Function.{call,construct]()` now take `GlobalObject&`. I was already refactoring these functions, so I figured I might as well just slip that change in here too.

Fixes #2624

### DIFF
--- a/Libraries/LibJS/Interpreter.h
+++ b/Libraries/LibJS/Interpreter.h
@@ -136,6 +136,15 @@ public:
 
     bool in_strict_mode() const { return m_scope_stack.last().scope_node->in_strict_mode(); }
 
+    template<typename Callback>
+    void for_each_argument(Callback callback)
+    {
+        if (m_call_stack.is_empty())
+            return;
+        for (auto& value : m_call_stack.last().arguments)
+            callback(value);
+    }
+
     size_t argument_count() const
     {
         if (m_call_stack.is_empty())
--- a/Libraries/LibJS/Runtime/ProxyObject.cpp
+++ b/Libraries/LibJS/Runtime/ProxyObject.cpp
@@ -501,7 +501,7 @@ Value ProxyObject::call(Interpreter& interpreter) {
     return interpreter.call(trap.as_function(), Value(&m_handler), move(arguments));
 }
 
-Value ProxyObject::construct(Interpreter& interpreter) {
+Value ProxyObject::construct(Interpreter& interpreter, Function& new_target) {
     if (!is_function())
         return interpreter.throw_exception<TypeError>(ErrorType::NotAConstructor, Value(this).to_string_without_side_effects().characters());
 
@@ -513,7 +513,7 @@ Value ProxyObject::construct(Interpreter& interpreter) {
     if (interpreter.exception())
         return {};
     if (trap.is_empty() || trap.is_undefined() || trap.is_null())
-        return static_cast<Function&>(m_target).construct(interpreter);
+        return static_cast<Function&>(m_target).construct(interpreter, new_target);
     if (!trap.is_function())
         return interpreter.throw_exception<TypeError>(ErrorType::ProxyInvalidTrap, "construct");
 
@@ -524,9 +524,7 @@ Value ProxyObject::construct(Interpreter& interpreter) {
         arguments_array->indexed_properties().append(argument);
     });
     arguments.values().append(arguments_array);
-    // FIXME: We need access to the actual newTarget property here. This is just
-    // a quick fix
-    arguments.values().append(Value(this));
+    arguments.values().append(Value(&new_target));
     auto result = interpreter.call(trap.as_function(), Value(&m_handler), move(arguments));
     if (!result.is_object())
         return interpreter.throw_exception<TypeError>(ErrorType::ProxyConstructBadReturnType);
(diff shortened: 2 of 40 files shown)


Libraries/LibJS/Runtime/ProxyObject.cpp
```diff
@@ -466,9 +467,79 @@ Value ProxyObject::delete_property(PropertyName name)
 
 void ProxyObject::visit_children(Cell::Visitor& visitor)
 {
-    Object::visit_children(visitor);
+    Function::visit_children(visitor);
     visitor.visit(&m_target);
     visitor.visit(&m_handler);
 }
 
+Value ProxyObject::call(Interpreter& interpreter) {
```
awesomekling: Please use `clang-format`.

Libraries/LibJS/Interpreter.h
```diff
@@ -136,6 +136,13 @@ class Interpreter : public Weakable<Interpreter> {
 
     bool in_strict_mode() const { return m_scope_stack.last().scope_node->in_strict_mode(); }
 
+    Vector<Value> arguments() const
```
awesomekling, line 139: It's sad to return a `Vector` by value here since that means we have to copy it (meaning heap allocation.)
Let's use an iteration helper to avoid it:

```cpp
template<typename Callback>
void for_each_argument(Callback callback)
{
    if (m_call_stack.is_empty())
        return;
    for (auto& value : m_call_stack.last().arguments)
        callback(value);
}
```

Then you can use that wherever you want to iterate over the current arguments:

```cpp
interpreter.for_each_argument([&](auto& argument) {
    do_stuff_with(argument);
});
```

Libraries/LibJS/Runtime/ProxyObject.cpp
```diff
@@ -466,9 +467,79 @@ Value ProxyObject::delete_property(PropertyName name)
...
+    if (interpreter.exception())
+        return {};
+    if (trap.is_empty() || trap.is_undefined() || trap.is_null())
+        return static_cast<Function&>(m_target).call(interpreter);
+    if (!trap.is_function())
+        return interpreter.throw_exception<TypeError>(ErrorType::ProxyInvalidTrap, "apply");
+
+    MarkedValueList arguments(interpreter.heap());
+    arguments.values().append(Value(&m_target));
+    arguments.values().append(Value(&m_handler));
```
awesomekling, line 493: Can we just use `MarkedValueList::append()`?

```suggestion
    arguments.append(&m_target);
    arguments.append(&m_handler);
```

(+Same suggestion where it applies elsewhere)

Libraries/LibJS/Runtime/ProxyObject.h
```diff
@@ -59,6 +64,8 @@ class ProxyObject : public Object {
 private:
     virtual void visit_children(Visitor&) override;
     virtual bool is_proxy_object() const override { return true; }
+    bool is_function() const override { return m_target.is_function(); }
```
awesomekling, line 67: Please mark this `virtual`

Libraries/LibJS/Runtime/ProxyObject.h
```diff
@@ -26,19 +26,24 @@
...
 public:
     static ProxyObject* create(GlobalObject&, Object& target, Object& handler);
 
     ProxyObject(Object& target, Object& handler, Object& prototype);
     virtual ~ProxyObject() override;
 
+    Value call(Interpreter& interpreter) override;
+    Value construct(Interpreter& interpreter) override;
+    const FlyString& name() const override;
+    LexicalEnvironment* create_environment() override;
```
awesomekling: Please mark these `virtual`
Also please omit the parameter names for `call` and `construct` since they are identical to the type name.

Libraries/LibJS/Runtime/ProxyObject.cpp
mattco98: Huh, for some reason I've done this in all of my LibJS code. I think I copied it from somewhere I saw (that no longer does it) and just never bothered to actually look at that class to see that method. I think I'll leave it here, and fix all places where I've done it in a new QOL PR that will also fix some other issues.

awesomekling:
Hmm, why do `call()` and `construct()` need the global object as a parameter?
They can already access the global object they belong to by calling `Object::global_object()` on themselves.

mattco98:
Alright, I removed that commit.

awesomekling:
Conflicts :'(

mattco98:
Oops! Good to go now :)

PR #2631, opened 2020-06-26
LibWeb: Remove duplicate `if` branch in `parse_html_fragment`.

theazgra:
In the video I noticed that one `if` branch is duplicate of its predecessor.
This PR just removes the duplicated if check.


awesomekling:
Oops! Thanks for fixing :)

PR #2632, opened 2020-06-26
LibLine: LibMultipleLines! (Support multiline editing)

alimpfard:
This commit also updates Shell, which uses actual_rendered_length.

Now we can edit all the lines!


alimpfard:
Hmm...I think @bugaevc might be interested, so be cc'd!

PR #2635, opened 2020-06-26
LaunchServer: Associate JPEG files with QuickShow

### LINKED ISSUE #2636: LibGfx: JPGLoader failing to decode pictures in context of QuickShow
The decoder happily decodes JPEGs when used in browser or markdown viewer but fails when asked to decode in context of QuickShow or File Manger (when rendering thumbnails).

Strangely enough, I have been able to fix this by adding a debug log in `load_jpg` implementation of `JPGLoader.cpp` before making the call to `load_jpg_impl`. Something like this:

    MappedFile mapped_file(path);
    if (!mapped_file.is_valid()) {
        return nullptr;
    }
    
    dbg() << String::format("anything here");
    auto bitmap = load_jpg_impl((const u8*)mapped_file.data(), mapped_file.size());

This succeeded every time I tried but @mattco98 confirmed that he couldn't get the same behaviour by adding the line locally. In my case, adding the aforementioned line makes JPGLoader decode and render properly in QuickShow and File Manager.

devsh0:
This seemed missing from the JPEG PR.
Fixes #2636.


Libraries/LibGfx/JPGLoader.cpp
```diff
@@ -1112,14 +1112,17 @@ static bool decode_jpg(JPGLoadingContext& context)
 
 static RefPtr<Gfx::Bitmap> load_jpg_impl(const u8* data, size_t data_size)
 {
-    JPGLoadingContext context;
-    context.data = data;
-    context.data_size = data_size;
+    OwnPtr<JPGLoadingContext> context = make<JPGLoadingContext>();
```
awesomekling, line 1115: This is just a hack fix, we need to figure out what's going wrong :)

devsh0: Today's lesson: Always initialize your variables :)

awesomekling: Aha there we go! Please initialize all other (primitive) variables as well :)

PR #2640, opened 2020-06-26
LibWeb: Implement almost all missing tokenizer cases

Lubrsi:


Libraries/LibWeb/Parser/HTMLTokenizer.cpp
```diff
@@ -1038,6 +1042,20 @@ Optional<HTMLToken> HTMLTokenizer::next_token()
...
+                }
+                ON(0)
+                {
+                    PARSE_ERROR();
+                    m_current_token.m_tag.attributes.last().local_name_builder.append_codepoint(0xFFFD);
+                    continue;
+                }
+                ON('"') { PARSE_ERROR(); }
+                ON('\'') { PARSE_ERROR(); }
+                ON('<') { PARSE_ERROR(); }
```
awesomekling, line 1058: These look diffferent from the rest of the file.
Also `PARSE_ERROR` does not implicitly return. If we want to skip ahead to the "anything else" case, let's make that explicit with a `goto` :)

Libraries/LibWeb/Parser/HTMLTokenizer.cpp
```diff
@@ -1176,8 +1198,15 @@ Optional<HTMLToken> HTMLTokenizer::next_token()
...
-                    TODO();
+                    PARSE_ERROR();
+                    m_current_token.m_tag.attributes.last().value_builder.append_codepoint(0xFFFD);
+                    continue;
                 }
+                ON('"') { PARSE_ERROR(); }
+                ON('\'') { PARSE_ERROR(); }
+                ON('<') { PARSE_ERROR(); }
+                ON('=') { PARSE_ERROR(); }
+                ON('`') { PARSE_ERROR(); }
```
awesomekling, line 1209: Same issue here.

Lubrsi:
Done 👍

Issue #2634, opened 2020-06-26
First boot, Fortuna RNG assertion failed m_reseed_number > 0


RobertKosten:
It looks to me like an initial `this->reseed();` is missing.

RobertKosten:
[88cf780e65e8ccc70d62ce8950411578bd88b41b](https://github.com/3541/serenity/commit/88cf780e65e8ccc70d62ce8950411578bd88b41b) does not seem to resolve the issue yet :-(

I'm following the [SerenityOS build instructions](https://github.com/SerenityOS/serenity/blob/master/Documentation/BuildInstructions.md) for `Ubuntu 18.04 LTS Bionic Beaver`. My `gcc --version` is `gcc (Ubuntu 9.3.0-11ubuntu0~18.04.1) 9.3.0`, my `cmake --version` is `cmake version 3.17.3`.

The output from `make run`:

```
CPU Reset (CPU 0)
EAX=00000000 EBX=00000000 ECX=00000000 EDX=00000000
ESI=00000000 EDI=00000000 EBP=00000000 ESP=00000000
EIP=00000000 EFL=00000000 [-------] CPL=0 II=0 A20=0 SMM=0 HLT=0
ES =0000 00000000 00000000 00000000
CS =0000 00000000 00000000 00000000
SS =0000 00000000 00000000 00000000
DS =0000 00000000 00000000 00000000
FS =0000 00000000 00000000 00000000
GS =0000 00000000 00000000 00000000
LDT=0000 00000000 00000000 00000000
TR =0000 00000000 00000000 00000000
GDT=     00000000 00000000
IDT=     00000000 00000000
CR0=00000000 CR2=00000000 CR3=00000000 CR4=00000000
DR0=00000000 DR1=00000000 DR2=00000000 DR3=00000000 
DR6=00000000 DR7=00000000
CCS=00000000 CCD=00000000 CCO=DYNAMIC 
EFER=0000000000000000
FCW=0000 FSW=0000 [ST=0] FTW=ff MXCSR=00000000
FPR0=0000000000000000 0000 FPR1=0000000000000000 0000
FPR2=0000000000000000 0000 FPR3=0000000000000000 0000
FPR4=0000000000000000 0000 FPR5=0000000000000000 0000
FPR6=0000000000000000 0000 FPR7=0000000000000000 0000
XMM00=00000000000000000000000000000000 XMM01=00000000000000000000000000000000
[...]

PR #2641, opened 2020-06-27
Kernel: Add g_cpu_supports_rdseed

3541:
CPUs which support RDRAND do not necessarily support RDSEED. This
introduces a flag g_cpu_supports_rdseed which is set appropriately
by CPUID. This causes Haswell CPUs in particular (and probably a lot
of AMD chips) to now fail to boot with #2634, rather than an illegal
instruction.

It seems like the KernelRng needs either an initial reseed call or
more random events added before the first call to get_good_random,
in order to prevent that issue, but I don't feel qualified to make that kind
of change to the crypto code.

Apologies in advance if I've screwed up some kind of contribution etiquette, 
as this is my first interaction with the project.

### DIFF
--- a/Kernel/Arch/i386/CPU.cpp
+++ b/Kernel/Arch/i386/CPU.cpp
@@ -714,6 +714,7 @@ bool g_cpu_supports_nx;
 bool g_cpu_supports_pae;
 bool g_cpu_supports_pge;
 bool g_cpu_supports_rdrand;
+bool g_cpu_supports_rdseed;
 bool g_cpu_supports_smap;
 bool g_cpu_supports_smep;
 bool g_cpu_supports_sse;
@@ -736,6 +737,7 @@ void cpu_detect()
     g_cpu_supports_smap = (extended_features.ebx() & (1 << 20));
     g_cpu_supports_smep = (extended_features.ebx() & (1 << 7));
     g_cpu_supports_umip = (extended_features.ecx() & (1 << 2));
+    g_cpu_supports_rdseed = (extended_features.ebx() & (1 << 18));
 }
 
 void cpu_setup()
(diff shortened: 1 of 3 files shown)


awesomekling:
cc @alimpfard @Petelliott 

Hi there! I'm a bit confused by the commit message here. Are you saying that this *fixes* #2634 or that this is just a workaround that allows those systems (with `rdrand` but not `rdseed`) to boot albeit with an unseeded RNG?

Seems like we should have some kind of fallback mechanism in that case.

3541:
It fixes a separate bug to #2634 (that is, how Serenity assumes rdrand support implies rdseed support), but because of #2634  issue, systems affected by this bug are still unable to boot, because in the absence of rdseed it doesn't run the loop at Random.cpp:48.

I wasn't sure what the correct way to go about fixing that issue was, so I left it alone for someone better-versed in the crypto code. It seems that, while there are other randomness sources in other devices, it's still not enough to trigger the reseed at Random.h:60.

alimpfard:
I don't have access to a real cpu that doesn't have rdseed, so I can't really test this, but I did try telling qemu to emulate a haswell processor without rdrand and rdseed, it did not hit the aforementioned assertion...

But we could just fallback to the old shitty rng if there's no entropy available, or seed with it on the spot and block a little?

awesomekling:
All right, thanks for the explanation. Let's merge this and we can fix the other issue separately. :)

3541:
I'm happy to do whatever testing might help. My system has rdrand but not rdseed, and I have some others which have neither instruction. Either way, it seems like asserting on insufficient entropy is probably not the correct behavior. I'm not sufficiently familiar with their comparative benefits, but perhaps it would be possible to do a multi-level fallback sort of thing where, say:
* Has rdseed => use rdseed
* No rdseed but has rdrand => do the same thing as is currently done for rdseed, but use rdrand instead.
* Has neither => fall back to old behavior

PR #2642, opened 2020-06-27
DisplaySettings: Allow setting wallpapers from outside /res/wallpapers

devsh0:


Applications/DisplaySettings/DisplaySettings.cpp
```diff
@@ -125,6 +125,11 @@ void DisplaySettingsWidget::create_frame()
     m_wallpaper_combo->set_model(*ItemListModel<AK::String>::create(m_wallpapers));
     m_wallpaper_combo->on_change = [this](auto& text, const GUI::ModelIndex& index) {
         String path = text;
+        if (this->m_monitor_widget->set_wallpaper(path)) {
+            this->m_monitor_widget->update();
```
awesomekling, line 129: No need for `this` when accessing member variables.

devsh0: I didn't remove `this` everywhere in the file, just in the function I modified. Should I do?

PR #2646, opened 2020-06-27
Kernel: Make Random work on CPUs without rdrand

### LINKED ISSUE #2634: First boot, Fortuna RNG assertion failed m_reseed_number > 0

Petelliott:
- If rdseed is not available, fallback to rdrand.
- If rdrand is not available, block when user needs good random, and use insecure PRNG when user needs fast random.

Fixes #2634

### DIFF
--- a/Kernel/Random.cpp
+++ b/Kernel/Random.cpp
@@ -44,17 +44,39 @@ KernelRng& KernelRng::the()
 
 KernelRng::KernelRng()
 {
-    if (g_cpu_supports_rdseed) {
+    if (g_cpu_supports_rdseed || g_cpu_supports_rdrand) {
         for (size_t i = 0; i < resource().pool_count * resource().reseed_threshold; ++i) {
             u32 value = 0;
-            asm volatile(
-                "1:\n"
-                "rdseed %0\n"
-                "jnc 1b\n"
-                : "=r"(value));
+            if (g_cpu_supports_rdseed) {
+                asm volatile(
+                    "1:\n"
+                    "rdseed %0\n"
+                    "jnc 1b\n"
+                    : "=r"(value));
+            } else {
+                asm volatile(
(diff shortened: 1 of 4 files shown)


Kernel/Random.h
```diff
@@ -86,6 +86,16 @@ class FortunaPRNG {
         m_pools[pool].update(reinterpret_cast<const u8*>(&event_data), sizeof(T));
     }
 
+    bool seeded()
```
awesomekling, line 89: This should be `const`. Also let's call it `is_seeded()` per the project coding style.

Kernel/Random.h
```diff
@@ -86,6 +86,16 @@ class FortunaPRNG {
...
+    {
+        return m_reseed_number > 0;
+    }
+
+    bool ready()
```
awesomekling, line 94: This should be `const`. Also let's call it `is_ready()` per the project coding style.

Kernel/Random.h
```diff
@@ -86,6 +86,16 @@ class FortunaPRNG {
...
+    {
+        return this->seeded() || m_p0_len >= reseed_threshold;
```
awesomekling, line 96: No need for `this->` here.

Kernel/Random.cpp
```diff
@@ -73,7 +97,25 @@ void get_good_random_bytes(u8* buffer, size_t buffer_size)
 
 void get_fast_random_bytes(u8* buffer, size_t buffer_size)
 {
-    return get_good_random_bytes(buffer, buffer_size);
+    if (KernelRng::the().resource().ready()) {
+        return get_good_random_bytes(buffer, buffer_size);
+    } else {
```
awesomekling, line 102: No need for `else` after `return`.

PR #2647, opened 2020-06-27
Documentation: Add filesystem note to WSL2 notes

ADKaster:
WSL2 is now available in non-insider builds of Windows, starting
with version 2004.

Add a filesystem note regarding use of /mnt/c et al from WSL2 for
compiling serenity. Namely, recommend against it as the performance
across the 9p file system protocol is terrible for IO heavy jobs.

For the last note there, when pointing SERENITY_QEMU_BIN at `/mnt/c/Program Files/qemu/qemu-system-i386.exe` and running from my WSL2 local filesystem, every version of QEMU I tried gave me an error such as

```
ERROR:/home/stefan/src/qemu/repo.or.cz/qemu/ar7/block.c:1498:bdrv_open_driver: assertion failed: (is_power_of_2(bs->bl.r
equest_alignment))
```
Which clearly seems bogus. However, moving the disk image and kernel to `/mnt/c` or `/mnt/e` fixed the issue, and ran the native QEMU just fine from inside WSL. 🤷‍♂️ Fileystems.

CC: @edrochenski @devsh0

### DIFF
--- a/Documentation/NotesOnWSL.md
+++ b/Documentation/NotesOnWSL.md
@@ -3,10 +3,7 @@
 SerenityOS can also be built and run under WSL Version 2.
 WSL Version 1 is not supported since Version 1 does not support ext2, which is needed for the setup.
 
-WSL Version 2 is currently only available for Insider Builds of Windows which are more unstable and prone to crashes.
-Therefore, running SerenityOS on WSL Version 2 and running Insider Builds, in general, is not recommended on production systems.
-
-Nevertheless, here is a guide on how to [get an Insider Build](https://insider.windows.com/en-us/how-to-pc/) and how to [get WSL2](https://docs.microsoft.com/en-us/windows/wsl/wsl2-install).
+WSL Version 2 requires Windows version 2004 or higher, with OS Build 19041 or greater. Here is a guide on how to [get WSL2](https://docs.microsoft.com/en-us/windows/wsl/install-win10).
 Once installed, you will need to make sure the distribution you want to use (and the new default) is using Version 2:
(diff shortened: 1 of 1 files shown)


devsh0:
I moved away from Windows a while ago, being super annoyed by the build speed and runtime performance in general :(

ADKaster:
> I moved away from Windows a while ago, being super annoyed by the build speed and runtime performance in general :(

Well, when I was testing yesterday it was compiling ~4x faster than inside my VirtualBox VM, when I had the project living in the WSL filesystem. Cloning into, for example `C:\Users\Andrew\serenity` and building there to test out why QEMU was not co-operating really sucked though, it look like an hour to finally unarchive binutils and gcc, and build the dang things along with the OS. Never doing that again.

devsh0:
Yep! extracting the binaries after downloading them from ridiculously slow mirrors. I remember that. Furthermore, windows has this weird power management policy that it won't by default allow VMs to utilize CPU beyond ~20%. Imagine that. And ofc, if you manage to get through all of that there's unstable windows ready to update on your next reboot, which you had to do because system processes won't stop choking the disk.

jaychandra6:
Hi @ADKaster I just built the serenity on WSL2 following the build instructions by cloning it into WSL fs, everything went fine but I'm getting this error while running with native QEMU installation

```
ERROR:/home/stefan/src/qemu/repo.or.cz/qemu/ar7/block.c:1498:bdrv_open_driver: assertion failed: (is_power_of_2(bs->bl.r
equest_alignment))
```
Is it possible to fix this and run with the native installation of QEMU?

PR #2654, opened 2020-06-28
Everywhere: Replace some uses of fork/exec with posix_spawn

nico:
It's less code, and it's potentially more efficient once
posix_spawn is a real syscall.

### DIFF
--- a/Applications/FileManager/main.cpp
+++ b/Applications/FileManager/main.cpp
@@ -58,6 +58,7 @@
 #include <LibGUI/Window.h>
 #include <LibGfx/Palette.h>
 #include <signal.h>
+#include <spawn.h>
 #include <stdio.h>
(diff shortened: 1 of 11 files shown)


nico:
I tried to match the previous error checking style at each place. (posix_spawn internally does some checking / perror)

nico:
Looks like the commit got merged (it's on https://github.com/SerenityOS/serenity/commits/master) but it doesn't show up on this PR. Not sure if GitHub is just slow in updating the PR or you patched this in to try something and then accidentally pushed it, or what :)

awesomekling:
Oh weird. GitHub was [having trouble this morning it seems](https://www.githubstatus.com/incidents/d60fmldlvw48) but it should be good now :)

PR #2656, opened 2020-06-29
LibKeyboard: Don't crash when file is empty

sppmacd:
Just return empty map, like when file error occurs.


Libraries/LibKeyboard/CharacterMapFile.cpp
```diff
@@ -50,7 +50,8 @@ Optional<CharacterMapData> CharacterMapFile::load_from_file(const String& file_n
 
     auto file_contents = file->read_all();
     auto json_result = JsonValue::from_string(file_contents);
-    ASSERT(json_result.has_value());
+    if(!json_result.has_value())
```
awesomekling, line 53: Please use `clang-format`.

PR #2657, opened 2020-06-29
LibGUI: SpinBox: update the displayed value when set_range() clamps it

benit8:
Consider the following: upon instanciation of a GUI::SpinBox, its internal value and displayed value are both 0. When calling `set_min(1)` on it (same as `set_range(1, max())`), the internal value gets clamped to 1 correctly, but "0" is still displayed by the widget.

The displayed value is now updated accordingly to the internal value when it gets clamped.


awesomekling:
Yup, great fix :)

PR #2658, opened 2020-06-29
Userland: Add watch tool

ccapitalK:
This adds a simple watch utility that runs a child program in a loop, similar to the watch utility in procps-ng. It runs the passed command through exec() instead of 'sh -c', unlike the actual watch utility, so shell variables in the passed command line are not evaluated.

### DIFF
--- a/Base/usr/share/man/man1/watch.md
+++ b/Base/usr/share/man/man1/watch.md
@@ -1,0 +1,30 @@
+## Name
+
+watch - execute a program periodically
+
+## Synopsis
+
+`watch [-n interval] [-t] [-b] command`
+
+## Description
+
+Run a command full-screen periodically until interrupted, then return the
+aggregated error code.
+
+## Options
+
+* '-n': Interval between executions, in seconds. By default, the program is run every 2 seconds.
+* '-t': Don't print the title bar.
+* '-b': Beep each time the command exits with a non-zero status
+
+## Exit Values
+
+* 0 - Success
+* 1 - At least one invocation of the command failed or exited with a non-zero status
+
+## Examples
+
+```sh
+$ watch -n1 ls
+$ watch -t -- uname -a
+```
--- a/Userland/watch.cpp
+++ b/Userland/watch.cpp
@@ -1,0 +1,178 @@
+/*
+ * Copyright (c) 2020, Sahan Fernando <sahan.h.fernando@gmail.com>
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include <AK/String.h>
+#include <AK/StringBuilder.h>
+#include <AK/Time.h>
+#include <AK/Vector.h>
+#include <LibCore/ArgsParser.h>
+#include <stdio.h>
+#include <sys/time.h>
+#include <sys/wait.h>
+#include <time.h>
+#include <unistd.h>
+
+static int opt_interval = 2;
+static bool flag_noheader = false;
+static bool flag_beep_on_fail = false;
+static volatile int exit_code = 0;
+static volatile pid_t child_pid = -1;
+
+String build_header_string(const Vector<const char*>& command, const struct timeval& interval)
+{
+    StringBuilder builder;
+    builder.appendf("Every %d", interval.tv_sec);
+    builder.appendf(".%ds: \x1b[1m", interval.tv_usec / 100000);
+    builder.join(' ', command);
+    builder.append("\x1b[0m");
+    return builder.build();
+}
+
+struct timeval get_current_time()
+{
+    struct timespec ts;
+    struct timeval tv;
+    clock_gettime(CLOCK_MONOTONIC, &ts);
+    timespec_to_timeval(ts, tv);
+    return tv;
+}
+
+int64_t usecs_from(const struct timeval& start, const struct timeval& end)
+{
+    struct timeval diff;
+    timeval_sub(end, start, diff);
+    return 1000000 * diff.tv_sec + diff.tv_usec;
+}
+
+void handle_signal(int signal)
+{
+    if (child_pid > 0) {
+        if (kill(child_pid, signal) < 0) {
+            perror("kill");
+        }
+        int status;
+        if (waitpid(child_pid, &status, 0) < 0) {
+            perror("waitpid");
+        } else if (!WIFEXITED(status) || WEXITSTATUS(status) != 0) {
+            exit_code = 1;
+        }
+    }
+    exit(exit_code);
+}
+
+int run_command(const Vector<const char*>& command)
+{
+    child_pid = fork();
+    if (child_pid < 0) {
+        // We failed to fork, so we shall print an error message and update the exit code.
+        exit_code = 1;
+        perror("fork");
+        return errno;
+    } else if (child_pid == 0) {
+        // We are in the child process, so we should run the command.
+        execvp(command[0], const_cast<char* const*>(command.data()));
+        perror("exec");
+        exit(1);
+    } else {
+        // We are still in the parent process, so we shall wait for the child to terminate,
+        // then return its exit code.
+        int status;
+        pid_t exited_pid;
+        do {
+            exited_pid = waitpid(child_pid, &status, 0);
+        } while (exited_pid < 0 && errno == EINTR);
+        ASSERT(exited_pid == child_pid);
+        child_pid = -1;
+        if (exited_pid < 0) {
+            perror("waitpid");
+            return 1;
+        }
+        if (WIFEXITED(status)) {
+            return WEXITSTATUS(status);
+        } else {
+            return 1;
+        }
+    }
+}
+
+int main(int argc, char** argv)
+{
+    signal(SIGINT, handle_signal);
+    if (pledge("stdio proc exec", nullptr) < 0) {
+        perror("pledge");
+        return 1;
+    }
+
+    Vector<const char*> command;
+    Core::ArgsParser args_parser;
+    args_parser.add_option(opt_interval, "Amount of time between updates", "interval", 'n', "seconds");
+    args_parser.add_option(flag_noheader, "Turn off the header describing the command and interval", "no-title", 't');
+    args_parser.add_option(flag_beep_on_fail, "Beep if the command has a non-zero exit code", "beep", 'b');
+    args_parser.add_positional_argument(command, "Command to run", "command");
+    args_parser.parse(argc, argv);
+
+    struct timeval interval;
+    if (opt_interval <= 0) {
+        interval = { 0, 100000 };
+    } else {
+        interval = { opt_interval, 0 };
+    }
+
+    auto header = build_header_string(command, interval);
+    command.append(nullptr);
+
+    auto now = get_current_time();
+    auto next_run_time = now;
+    while (true) {
+        int usecs_to_sleep = usecs_from(now, next_run_time);
+        while (usecs_to_sleep > 0) {
+            usleep(usecs_to_sleep);
+            now = get_current_time();
+            usecs_to_sleep = usecs_from(now, next_run_time);
+        }
+        // Clear the screen, then reset the cursor position to the top left.
+        fprintf(stderr, "\033[H\033[2J");
+        // Print the header.
+        if (!flag_noheader) {
+            fprintf(stderr, "%s\n\n", header.characters());
+        } else {
+            fflush(stderr);
+        }
+        if (run_command(command) != 0) {
+            exit_code = 1;
+            if (flag_beep_on_fail) {
+                fprintf(stderr, "\a");
+                fflush(stderr);
+            }
+        }
+        now = get_current_time();
+        timeval_add(next_run_time, interval, next_run_time);
+        if (usecs_from(now, next_run_time) < 0) {
+            // The next execution is overdue, so we set next_run_time to now to prevent drift.
+            next_run_time = now;
+        }
+    }
+}


Userland/watch.cpp
```diff
@@ -0,0 +1,171 @@
...
+    command.append(nullptr);
+
+    signal(SIGINT, handle_signal);
+
+    struct timeval interval {
+        opt_interval, 0
+    };
+    struct timeval now = get_current_time();
+    struct timeval next_run_time = now;
+    bool ran_immediately = false;
```
alimpfard, line 69: These guys could benefit from some `auto`.

Userland/watch.cpp
```diff
@@ -0,0 +1,171 @@
...
+    while (true) {
+        int usecs_to_sleep = usecs_till(now, next_run_time);
+        ran_immediately = usecs_to_sleep < 0;
+        while (usecs_to_sleep > 0) {
+            usleep(usecs_to_sleep);
+            now = get_current_time();
+            usecs_to_sleep = usecs_till(now, next_run_time);
+        }
+        // clear screen, reset cursor
```
alimpfard, line 78: This an other comments: Please format them like a proper english sentence.
e.g. `// Clear screen and reset cursor.`

Userland/watch.cpp
```diff
@@ -0,0 +1,171 @@
...
+        if (run_command(command) != 0) {
+            exit_code = 1;
+            if (flag_beep_on_fail) {
+                printf("\a");
+                fflush(stdout);
+            }
+        }
+        if (ran_immediately) {
+            // calculate next run time based on current time
+            AK::timeval_add(now, interval, next_run_time);
```
alimpfard, line 94: Here and other places; There's no need to qualify `AK` (unless someone forgot to bring `timeval_*` into the global namespace?).

Userland/watch.cpp
```diff
@@ -0,0 +1,171 @@
...
+String build_header_string(const Vector<const char*>& command)
+{
+    StringBuilder builder;
+    builder.appendf("Every %d.0s: \'", opt_interval);
+    for (auto i = 0u; i < command.size(); ++i) {
+        if (i > 0) {
+            builder.append(' ');
+        }
+        builder.append(command[i]);
+    }
```
alimpfard, line 112:
```c++
builder.join(' ', command);
```

Userland/watch.cpp
```diff
@@ -0,0 +1,171 @@
...
+    builder.append("\'");
+    return builder.build();
+}
+
+struct timeval get_current_time()
+{
+    struct timespec ts {
+    };
+    struct timeval tv {
+    };
```
alimpfard, line 122: Here and in `usecs_till`: Just drop the `{}`? imho it would look better.

Userland/watch.cpp
```diff
@@ -0,0 +1,171 @@
...
+static pid_t child_pid = -1;
+
+void handle_signal(int signal);
+int run_command(const Vector<const char*>& command);
+int64_t usecs_till(const struct timeval& now, const struct timeval& targ);
+struct timeval get_current_time();
+String build_header_string(const Vector<const char*>&);
+
+int main(int argc, char** argv)
+{
```
alimpfard, line 51: Oh, and also, let's do some `pledge`s?
I think `stdio proc exec` should be enough after `signal()`?

Userland/watch.cpp
ccapitalK: Fixed

Userland/watch.cpp
ccapitalK: Fixed

Userland/watch.cpp
ccapitalK: Fixed

Userland/watch.cpp
ccapitalK: Fixed

Userland/watch.cpp
ccapitalK: Fixed

Userland/watch.cpp
ccapitalK: I believe that I fixed it, I moved the `signal()` call up so that it predates the argument parsing.

Userland/watch.cpp
```diff
@@ -0,0 +1,165 @@
...
+    args_parser.add_option(opt_interval, "Amount of time between updates", "interval", 'n', "seconds");
+    args_parser.add_option(flag_noheader, "Turn off the header describing the command and interval", "no-title", 't');
+    args_parser.add_option(flag_beep_on_fail, "Beep if the command has a non-zero exit code", "beep", 'b');
+    args_parser.add_positional_argument(command, "Command to run", "command");
+    args_parser.parse(argc, argv);
+    auto header = build_header_string(command);
+    command.append(nullptr);
+
+    struct timeval interval {
+        opt_interval, 0
```
predmond, line 66: the behaviour of watch seems to set interval to 0.1s if the interval is <= 0. Maybe you should also validate opt_interval?

ccapitalK: Good catch, fixing now.

Userland/watch.cpp
alimpfard: looks alright to me.

Userland/watch.cpp
alimpfard: You can close these now lol

Userland/watch.cpp
ccapitalK: Fixed.

Userland/watch.cpp
```diff
@@ -0,0 +1,171 @@
...
+void handle_signal(int signal);
+int run_command(const Vector<const char*>& command);
+int64_t usecs_till(const struct timeval& now, const struct timeval& targ);
+struct timeval get_current_time();
+String build_header_string(const Vector<const char*>&, const struct timeval&);
+
+int main(int argc, char** argv)
+{
+    signal(SIGINT, handle_signal);
+    pledge("stdio proc exec", nullptr);
```
alimpfard, line 53: Some error checking would be nice.
```c++
if (pledge("stdio proc exec", nullptr) < 0) {
    perror("pledge");
    return 1;
}
```

alimpfard:
Also, I'm not entirely sure, but I think we want to have manpages for userland utilities.
I'll defer to @awesomekling about whether adding a manpage is a requirement for this PR though.

awesomekling:
> Also, I'm not entirely sure, but I think we want to have manpages for userland utilities.
> I'll defer to @awesomekling about whether adding a manpage is a requirement for this PR though.

Ahem.. \**channels inner @bugaevc*\*..
A manpage is always appropriate! :slightly_smiling_face:

Userland/watch.cpp
```diff
@@ -0,0 +1,171 @@
...
+    exit(exit_code);
+}
+
+int run_command(const Vector<const char*>& command)
+{
+    child_pid = fork();
+    if (child_pid < 0) {
+        // We failed to fork, so print an error message and update exit code
+        exit_code = 1;
+        perror("Fork() failed");
```
bugaevc, line 154: Simply

```suggestion
        perror("fork");
```

Userland/watch.cpp
```diff
@@ -0,0 +1,171 @@
...
+static bool flag_noheader = false;
+static bool flag_beep_on_fail = false;
+static int exit_code = 0;
+static pid_t child_pid = -1;
+
+void handle_signal(int signal);
+int run_command(const Vector<const char*>& command);
+int64_t usecs_till(const struct timeval& now, const struct timeval& targ);
+struct timeval get_current_time();
+String build_header_string(const Vector<const char*>&, const struct timeval&);
```
bugaevc, line 48: It would make sense to just move these utilities up and not have to forward-declare them 🙂

Userland/watch.cpp
```diff
@@ -0,0 +1,171 @@
...
+#include <sys/time.h>
+#include <sys/wait.h>
+#include <time.h>
+#include <unistd.h>
+
+static int opt_interval = 2;
+static bool flag_noheader = false;
+static bool flag_beep_on_fail = false;
+static int exit_code = 0;
+static pid_t child_pid = -1;
```
bugaevc, line 42: Accessed from signal handler => `volatile`

Userland/watch.cpp
```diff
@@ -0,0 +1,171 @@
...
+    }
+}
+
+String build_header_string(const Vector<const char*>& command, const struct timeval& interval)
+{
+    StringBuilder builder;
+    builder.appendf("Every %d", interval.tv_sec);
+    builder.appendf(".%ds: \'", interval.tv_usec / 100000);
+    builder.join(' ', command);
+    builder.append("\'");
```
bugaevc, line 115: Would perhaps be useful to add some color and/or bold

Userland/watch.cpp
```diff
@@ -0,0 +1,171 @@
...
+                printf("\a");
+                fflush(stdout);
+            }
+        }
+        if (ran_immediately) {
+            // Calculate the next run time based on the current time
+            timeval_add(now, interval, next_run_time);
+        } else {
+            // Calculate the next run time based on the previous run time
+            timeval_add(next_run_time, interval, next_run_time);
```
bugaevc, line 103: Does `timeval_add()` handle the first and the last arguments being the same correctly?

Userland/watch.cpp
```diff
@@ -0,0 +1,171 @@
...
+        }
+        fflush(stdout);
+        if (run_command(command) != 0) {
+            exit_code = 1;
+            if (flag_beep_on_fail) {
+                printf("\a");
+                fflush(stdout);
+            }
+        }
+        if (ran_immediately) {
```
bugaevc, line 98: This logic is not too clear to me; perhaps you could expand why it works like this; and perhaps it's an indication that is needs more commenting (or changing code to be even more clear).

Userland/watch.cpp
bugaevc: Note the period at the end 🙂.

Userland/watch.cpp
```diff
@@ -0,0 +1,171 @@
...
+        while (usecs_to_sleep > 0) {
+            usleep(usecs_to_sleep);
+            now = get_current_time();
+            usecs_to_sleep = usecs_till(now, next_run_time);
+        }
+        // Clear the screen, then reset the cursor position to the top left
+        printf("\033[H\033[2J");
+        // Print the header
+        if (!flag_noheader) {
+            printf("%s\n\n", header.characters());
```
bugaevc, line 88: Should this perhaps go to stderr?
Not that it matters that much since in this case both are very likely the same file (the terminal)

Userland/watch.cpp
```diff
@@ -0,0 +1,171 @@
...
+        // We are in the child process, so close stdin and run the command
+        close(STDIN_FILENO);
+        execvp(command[0], const_cast<char* const*>(command.data()));
+        perror("Exec() failed");
+        exit(1);
+    } else {
+        // We are still in the parent process, so wait for the child to terminate,
+        // then return its code
+        int status;
+        waitpid(child_pid, &status, 0);
```
bugaevc, line 166: Missing error check

Userland/watch.cpp
ccapitalK: Yes, I believe it does, since in its implementation no part of either of the first two parameters are read from after the corresponding part in the third parameter is written. I'm not sure if this is an intended part of its API though.

Userland/watch.cpp
ccapitalK: Fixed.

Userland/watch.cpp
ccapitalK: Fixed.

Userland/watch.cpp
ccapitalK: Fixed.

Userland/watch.cpp
ccapitalK: I've added bolding around the command, not sure what color would look good here. Maybe blue for the rest?

Userland/watch.cpp
ccapitalK: Yeah, I re-wrote this chunk this morning when my brain was working better, hopefully it's more clear now.

Userland/watch.cpp
ccapitalK: Fixed, I agree that it makes sense to do that so that piping works better.

Userland/watch.cpp
ccapitalK: Fixed

Userland/watch.cpp
ccapitalK: Fixed

Userland/watch.cpp
ccapitalK: Hmm, on second thought, this might be UB since the first two parameters are const references.

Base/usr/share/man/man1/watch.md
```diff
@@ -0,0 +1,30 @@
...
+## Exit Values
+
+* 0 - Success
+* 1 - At least one invocation of command failed or returned non-zero
+
+## Examples
+
+```sh
+$ watch -n1 ls
+$ watch -t -- uname -a
```
bugaevc, line 29: Maybe it's a task for another time, but it'd be cool if we could tell `ArgsParser` to tell `getopt` to treat all options after the first argument as arguments, not options. Then we'd be able to use that setting here in `watch`, and would not need the `--` in this example.

Base/usr/share/man/man1/watch.md
```diff
@@ -0,0 +1,30 @@
...
+`watch [-n interval] [-t] [-b] command`
+
+## Description
+
+Run a command full-screen periodically until interrupted, then return the
+aggregated error code.
+
+## Options
+
+* '-n': Interval between executions, in seconds
```
bugaevc, line 16: "By default, the program is run each 2 seconds."

Base/usr/share/man/man1/watch.md
```diff
@@ -0,0 +1,30 @@
...
+* '-t': Don't print the title bar
+* '-b': Beep if the command has a non-zero exit status
```
bugaevc, line 18: Each time the program exits with non-zero status, right?

Base/usr/share/man/man1/watch.md
```diff
@@ -0,0 +1,30 @@
...
+
+## Exit Values
+
+* 0 - Success
+* 1 - At least one invocation of command failed or returned non-zero
```
bugaevc, line 23:
```suggestion
* 1 - At least one invocation of the command failed or exited with a non-zero status
```

Userland/watch.cpp
```diff
@@ -0,0 +1,169 @@
...
+    timeval_sub(targ, now, diff);
+    return 1000000 * diff.tv_sec + diff.tv_usec;
+}
+
+void handle_signal(int signal)
+{
+    if (child_pid > 0) {
+        int status;
+        kill(child_pid, signal);
+        waitpid(child_pid, &status, 0);
```
bugaevc, line 75: Please move the `status` declaration to below `kill()` (right above where it's first needed, `waitpid()`), and do some error checks — both `kill()` and `waitpid()` could fail, for example if the child dies, or if the child is setuid.

Userland/watch.cpp
```diff
@@ -0,0 +1,169 @@
...
+        if (WEXITSTATUS(status) != 0) {
```
bugaevc, line 76: Can only use `WEXITSTATUS()` after `WIFEXITED()`

Userland/watch.cpp
```diff
@@ -0,0 +1,169 @@
...
+{
+    child_pid = fork();
+    if (child_pid < 0) {
+        // We failed to fork, so print an error message and update exit code
+        exit_code = 1;
+        perror("fork");
+        return errno;
+    } else if (child_pid == 0) {
+        // We are in the child process, so close stdin and run the command
+        close(STDIN_FILENO);
```
bugaevc, line 93: Should it really close stdin?

It's a *very* unexpected thing for a process (even more so, for a command line utility) not to have FD 0 open.

To me, it would make sense to either just keep stdin as is, or set it to `/dev/null`.

Userland/watch.cpp
```diff
@@ -0,0 +1,169 @@
...
+        execvp(command[0], const_cast<char* const*>(command.data()));
+        perror("exec");
+        exit(1);
+    } else {
+        // We are still in the parent process, so wait for the child to terminate,
+        // then return its exit code
+        int status;
+        pid_t exited_pid;
+        while ((exited_pid = waitpid(child_pid, &status, 0)) < 0 && errno == EINTR);
```
bugaevc, line 102: Uh, clever, but let's please do a good old `do {} while ();` without any assignments in the condition

Userland/watch.cpp
```diff
@@ -0,0 +1,169 @@
...
+        child_pid = -1;
```
bugaevc, line 103: At this point, we could `ASSERT(exited_pid == child_pid);`, right?

Userland/watch.cpp
```diff
@@ -0,0 +1,169 @@
...
+struct timeval get_current_time()
+{
+    struct timespec ts;
+    struct timeval tv;
+    clock_gettime(CLOCK_MONOTONIC, &ts);
+    timespec_to_timeval(ts, tv);
+    return tv;
+}
+
+int64_t usecs_till(const struct timeval& now, const struct timeval& targ)
```
bugaevc, line 63: *perhaps* it would make sense to name this `usecs_between()`?

Userland/watch.cpp
ccapitalK: Hmm, I've tested this with the procps-ng implementation, and it appears that one doesn't modify stdin. I think I'll keep it open then.

Userland/watch.cpp
bugaevc: Yeah, much clearer now 🙂

Userland/watch.cpp
ccapitalK: I considered that, but I was worried that the name `usecs_between()` would imply an unsigned return value. What do you think about `usecs_from(const struct timeval& now, const struct timeval& then)`?

Userland/watch.cpp
ccapitalK: Done.

Userland/watch.cpp
bugaevc: Yeah, "from" also sounds OK. My problem with "till" is to suggests the function only accepts one argument, and automatically fetches the value of "now"; when in reality it takes two arguments.

bugaevc:
Looks good to me now ...I think 🙂

At least nothing nitpickable stands out (other than missing periods at the end of comments, but I'll let that slide 😀)

ccapitalK:
I fixed the periods on comments thing (since might as well).

Userland/watch.cpp
ccapitalK: I think I'm just going to stick with bolding it.

ccapitalK:
Can it be merged, then? How many approvals are needed?

bugaevc:
> How many approvals are needed?

It's always @awesomekling who merges PRs; I can only raise concerns or "approve" PRs 🙂

PR #2661, opened 2020-06-29
LibVT/Terminal: add a scroll length to the TerminalWidget

benit8:
The scroll length is the number of lines by which the terminal will go up/down when scrolling the mouse wheel once.

### Why
Open a new Terminal and try the following: `ls -l js-tests/`. Now scroll back to the top (only using your mouse).
As you can see, the Terminal scrolls lines one by one, which I think is rather slow.

So I added the possibility the tell the TerminalWidget by how much you want to go up/down with one scroll: the scroll length.
It remains 1 by default in case most people actually like scrolling slowly; else I would recommend setting it between 5 or 7 :)

I also added a `GUI::SpinBox` in the Terminal's settings to adjust the scroll length gracefully. The minimum is clamped to 1 and I left the maximum untouched at 100 (though I can't imagine why anyone would want to scroll so much at once).

### DIFF
--- a/Applications/Terminal/main.cpp
+++ b/Applications/Terminal/main.cpp
@@ -37,6 +37,7 @@
 #include <LibGUI/MenuBar.h>
 #include <LibGUI/RadioButton.h>
 #include <LibGUI/Slider.h>
+#include <LibGUI/SpinBox.h>
 #include <LibGUI/Widget.h>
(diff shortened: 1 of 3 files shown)


awesomekling:
Cool, now I can't unsee the overly slow scrolling :D

PR #2665, opened 2020-06-30
LibGUI: Only show FilePicker preview pane on demand

linusg:
`FilePicker::set_preview()` and `FilePicker::clear_preview()` now show and hide the preview pane respectively.

image
image


awesomekling:
Omg yes!

PR #2666, opened 2020-06-30
WindowServer: Resolution fix

sppmacd:
The screen resolution was not saved due to /etc "Read-only file system" error. This PR changes this.
Also removed potentially harmful assert in WindowServer.

### DIFF
--- a/Services/WindowServer/Compositor.cpp
+++ b/Services/WindowServer/Compositor.cpp
@@ -421,7 +421,10 @@ bool Compositor::set_resolution(int desired_width, int desired_height)
         return true;
 
     // Make sure it's impossible to set an invalid resolution
-    ASSERT(desired_width >= 640 && desired_height >= 480);
+    if (!(desired_width >= 640 && desired_height >= 480)) {
+        dbg() << "Compositor: Tried to set invalid resolution: " << desired_width << "x" << desired_height;
+        return false;
+    }
     bool success = Screen::the().set_resolution(desired_width, desired_height);
     init_bitmaps();
     compose();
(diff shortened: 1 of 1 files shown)


alimpfard:
Hmmm...I think @bugaevc did that on purpose?
I don't quite remember _why_ though

bugaevc:
Uhhh, we definitely don't want WindowServer to save its settings to `/etc`, or be able to do so. Since it runs "for anon", it would make sense to me that anon's graphics settings are stored in an ini file in his home directory. I think it has worked this way before, or am I misremembering?

sppmacd:
@bugaevc Some settings (wallpaper) are stored in ```/home/anon/WindowManager.ini``` and some - resolution, wallpaper mode (scaled etc.) - are stored in ```/etc/WindowServer/WindowServer.ini```.

bugaevc:
Well, that's a really weird split — storing wallpaper and wallpaper mode in different places (user vs system-wide). Any idea why it's done this way? And if there's no good reason, let's consolidate everything in user's settings.

sppmacd:
> let's consolidate everything in user's settings.
WindowServer doesn't have access to user profile.

bugaevc:
I'd be much more comfortable giving WindowServer access to its own config file inside user's home directory (that one file, not the whole profile) than making `/etc` writable and letting WindowServer write there.

sppmacd:
We don't know for which user WindowServer is running.
Possible solution: pass user name to WindowServer (as environment variable or argument).
But it's better than letting entire /etc only for WindowServer.

bugaevc:
For now, everything in Serenity kinda assumes that GUI programs run for anon. It's possible to launch GUI programs with different UID/GID, but they will still use the hardcoded paths inside anon's home directory.

I don't know what should be done about it (should we fix everything to support multiple users? should we declare SerenityOS a single-user system where Unix UIDs are used for privilege separation but not for multiple physical users?), but now's not the time to fix that situation in any case.

Issue #92, opened 2019-05-25
LibC: Rewrite qsort()


awesomekling:
We're currently using a borrowed qsort() from some BSD. We should have our own.

ccapitalK:
I believe this is fixed now, right?

PR #2667, opened 2020-06-30
LibC: Replace Berkley's qsort() with AK::dual_pivot_quick_sort() wrapper

ccapitalK:
This uses a hack where we call AK::quick_sort using a custom iterator class which dereferences to a char, but iterates by a step size that is known at run time.

This should solve #92. I believe it is safe to remove the UC Berkley copyright header, since we now are only using our own code.

### DIFF
--- a/Libraries/LibC/qsort.cpp
+++ b/Libraries/LibC/qsort.cpp
@@ -24,103 +24,81 @@
  * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
-/*-
- * Copyright (c) 1980, 1983, 1990 The Regents of the University of California.
- * All rights reserved.
- *
- * Redistribution and use in source and binary forms, with or without
- * modification, are permitted provided that the following conditions
- * are met:
- * 1. Redistributions of source code must retain the above copyright
- *    notice, this list of conditions and the following disclaimer.
- * 2. Redistributions in binary form must reproduce the above copyright
- *    notice, this list of conditions and the following disclaimer in the
- *    documentation and/or other materials provided with the distribution.
- * 3. All advertising materials mentioning features or use of this software
- *    must display the following acknowledgement:
- *	This product includes software developed by the University of
- *	California, Berkeley and its contributors.
- * 4. Neither the name of the University nor the names of its contributors
- *    may be used to endorse or promote products derived from this software
- *    without specific prior written permission.
- *
- * THIS SOFTWARE IS PROVIDED BY THE REGENTS AND CONTRIBUTORS ``AS IS'' AND
- * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
- * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
- * ARE DISCLAIMED.  IN NO EVENT SHALL THE REGENTS OR CONTRIBUTORS BE LIABLE
- * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
- * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS
- * OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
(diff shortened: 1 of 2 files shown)


bugaevc:
The problem I've ran into when trying to do this is that `AK::quick_sort()` used `AK::swap()` and/or direct assignment to swap items; and that just doesn't do the right thing for things that are actually dynamically sized.

Have you been able to solve this somehow?

Libraries/LibC/qsort.cpp
```diff
@@ -24,103 +24,65 @@
...
+class SizedObjectIterator {
+public:
+    SizedObjectIterator() = delete;
+    SizedObjectIterator(void *pos, size_t element_size) : m_pos(static_cast<char*>(pos)), m_element_size(element_size) {}
+    bool operator!=(const SizedObjectIterator& other) { return m_pos != other.m_pos; }
+    bool operator==(const SizedObjectIterator& other) { return m_pos == other.m_pos; }
+    bool operator<(const SizedObjectIterator& other) { return m_pos < other.m_pos; }
+    bool operator>(const SizedObjectIterator& other) { return m_pos > other.m_pos; }
+    bool operator<=(const SizedObjectIterator& other) { return m_pos <= other.m_pos; }
+    bool operator>=(const SizedObjectIterator& other) { return m_pos >= other.m_pos; }
```
predmond, line 44: Out of curiousity, can you use the <=> operator here? I think gcc 10 supports it.

ccapitalK:
@bugavec Ah, I hadn't considered that. I believe this can be fixed by having the iterator dereference to a type that overloads assignment to a memcpy, though this may require a bit of trickery to get the size of the object inside the assignment, so it might not be worth it.

ccapitalK:
@bugaevc Ok, so this should fix the concerns you had (and as a plus, switch over to using dual_pivot_quick_sort). I still need to write actual tests though, and I'm worried I might be violating const correctness though.

ccapitalK:
I think we need to create a proper test runner for the LibC tests. The test program I provided can only really be run by copy pasting it into Userland/ before building, which isn't great.

ccapitalK:
I believe that it works now, the qsort test utility I wrote passes with my implementation and the CI task succeeds

ccapitalK:
@bugaevc On solving the issue of AK::swap not working, what I ended up doing was instead using a small custom collection type, that has an index operation that returns an instance of a pointer-size pair helper class instead of an object reference. I then specialized AK::swap for this helper class so that it swaps the pointed-to values instead of the helper objects, using the size that is stored. It seems a bit hacky I guess, but it compiles fine, doesn't invoke UB (at least I don't believe it does) and runs correctly based on the test program I wrote.

bugaevc:
Honesty I think it'd be easier to just duplicate the implementation than to do all these hacks.

But please attach your test program :)

ccapitalK:
> Honesty I think it'd be easier to just duplicate the implementation than to do all these hacks.
> 
> But please attach your test program :)

I've attached it as Tests/LibC/qsort-sorts-and-copies.cpp, I'm not sure whether there's supposed to be an official way of running the tests in that directory, I get the feeling that it used to be included in the image but then they were removed and then left unmaintained. 

Yeah, I considered duplicating it, but the duplicated code was about the same length, and the hack isn't that dirty, so I figured it was worth it to not have two independent quicksort implementations in the repo. I'm fine with the duplication if you think it to be a better strategy going forward, though.

Tests/LibC/qsort-sorts-and-copies.cpp
```diff
@@ -0,0 +1,107 @@
...
+#include <AK/QuickSort.h>
+#include <AK/String.h>
+#include <AK/Vector.h>
+#include <stdio.h>
+#include <stdlib.h>
+
+const size_t NUM_RUNS = 100;
+
+struct SortableObject {
+    String m_key;
```
bugaevc, line 36: No, you can't sort C++ objects with LibC `qsort()`, as it doesn't invoke move constructors. It probably happens to work in this case because `AK::String` is simple enough, but don't do that.

Do: C structures that contain multiple bytes. For a fun number, three bytes:

```cpp
struct foo {
    char a, b, c;
};
```

And don't initialize them all to zero, that misses the whole point of sorting :slightly_smiling_face:

ccapitalK: Good point, I forgot about that.

I've rewritten the test struct to be a POD type (and the test to randomly generate a test vector as well).

awesomekling:
This seems okay. It's a bit of an unusual approach, but if it's performance-neutral I don't mind it :)

PR #2669, opened 2020-06-30
Browser: Unset location bar icon on load start

linusg:
This avoids having a stale icon from a previous page load in the location box if a subsequently visited page doesn't trigger an icon change.

---

Fixes... the bug mentioned in the update video :^)


awesomekling:
Thank you :smile:

PR #2670, opened 2020-06-30
Add man pages for seteuid() and friends.

nico:
Also add an overview page that explains the general concepts.


nico:
@bugaevc

Base/usr/share/man/man2/geteuid.md
```diff
@@ -0,0 +1,25 @@
...
+gid_t getegid(void);
+```
+
+## Description
+
+Returns the effective user or group id.
+
+## See also
+
+* [`setuid_overview`(7)](../man7/setuid_overview.md)
```
bugaevc, line 20: No need for "overview", it should be just `setuid(7)`

Base/usr/share/man/man2/geteuid.md
```diff
@@ -0,0 +1,25 @@
...
+* [`getuid`(2) / `getgid`(2)](getuid.md)
+* [`getresuid`(2) / `getresgid`(2)](getresuid.md)
+* [`seteuid`(2) / `setegid`(2)](seteuid.md)
+* [`setuid`(2) / `setgid`(2)](setuid.md)
+* [`setresuid`(2) / `setresgid`(2)](setresuid.md)
```
bugaevc, line 25: Can we perhaps not have this many separate pages, and put everything into one or two pages instead?

Base/usr/share/man/man2/getresuid.md
```diff
@@ -0,0 +1,34 @@
...
+
+getresuid, getresgid - get real, effective, and saved user / group id
+
+## Synopsis
+
+```**c++
+#include <unistd.h>
+
+int getresuid(uid_t *, uid_t *, uid_t *);
+int getresgid(gid_t *, gid_t *, gid_t *);
```
bugaevc, line 11: The Serenity style is to put the pointer star (and reference ampersand) next to the basic type, and man pages should follow the same style.

Base/usr/share/man/man2/seteuid.md
```diff
@@ -0,0 +1,40 @@
...
+```**c++
+#include <unistd.h>
+
+int seteuid(uid_t);
+int setegid(gid_t);
+```
+
+## Description
+
+Sets the effective user or group id.
```
bugaevc, line 16: In text, ID is always uppercase.

Base/usr/share/man/man2/seteuid.md
```diff
@@ -0,0 +1,40 @@
...
+
+The effective user or group id is the id used to decide the permissions of the current process. An effective user id of 0 determines if the user is superuser.
```
bugaevc, line 18: `#include <english-ain't-my-first-language.h>`

"Decide permissions" sounds weird to me; perhaps "decide on permissions" or "determine permissions"?

Besides, such an explanation about how these IDs work should be in `setuid`(7), not in (2).

Base/usr/share/man/man2/setresuid.md
```diff
@@ -0,0 +1,38 @@
...
+
+int setresuid(uid_t, uid_t, uid_t);
+int setresgid(gid_t, gid_t, gid_t);
+```
+
+## Description
+
+Sets all of real, effective, and saved user or group id to the given values.
+
+`-1` means "no change".
```
bugaevc, line 18: "If one or more of the values is set to -1, the corresponding ID is not changed by this syscall.", or something like that.

Base/usr/share/man/man7/setuid_overview.md
```diff
@@ -0,0 +1,71 @@
+## Name
+
+Overview of real, effective, and saved user and group ids
+
+## Description
+
+Each process runs as user and a group. By default, a process started by a user has its user id set to that user and its group id set to that user's primary group.
```
bugaevc, line 7: That's, uh, kinda true, but not really. Users don't start processes; processes spawn other processes — perhaps, on behalf of a user. As you're surely know, the IDs a process runs at are determined/affected by ~three~ four simple rules:
* The kernel always starts processes as UID/GID 0. In particular, PID 4, the first userspace process, is spawned as root.
* A process can change its own IDs by one of these `set*()` calls.
* IDs are inherited on fork.
* IDs can be changed at exec time if the filesystem feels like it.

So these rules have to be clearly written out on this page. Now if, say, anon runs a shell with UID 100 and it spawns some new process, it inherits UID 100 by default — so in that sense, yes, the programs started by a user run as their UID/GID. But it's just that, a common case, not the underlying mechanism.

Base/usr/share/man/man7/setuid_overview.md
```diff
@@ -0,0 +1,71 @@
...
+
+Each file has permissions that control if it can be read, written, and executed by its owner, other users in the owner's group, and by others. A process has the same access permissions as its user and its group.
```
bugaevc, line 9: This, too, works the other way around. It's not that the process has the permissions of its user; those user permissions determine exactly that, what processes running as that UID/GID can do. The user cannot do anything by themselves other than generate some interrupts by moving mouse & typing on the keyboard; in particular they cannot access the files directly. It's always processes that act in behalf of the user.

Base/usr/share/man/man7/setuid_overview.md
```diff
@@ -0,0 +1,71 @@
...
+
+The motivation behind SUID binaries is that they allow users to do tasks that would normally require elevated permissions, without having to give users these permissions fully.
+
+For example, `/usr/ping` has the Set User ID bit set and is owned by user root in group root. So if any user executes `/usr/ping`, it will run as root, which means it will be able to send network packets, even if the current user doesn't normally have network access.
+
+For another example, `passwd` changes the password of the current user, which means it has to write to `/etc/shadow` -- but the current user is not supposed to have write access to `/etc/shadow` so that they cannot change passwords of other users. The solution is to make `passwd` a SUID binary. (SerenityOS currently doesn't have support for user passwords; this examples is more convincing on other UNIXes.)
```
bugaevc, line 17: Do we have actually `passwd`(1)?

Ah, I see you even mention that we don't; then you should start this with "many other Unix systems also include a `passwd` utility..." or something like that, instead of simply mentioning it at the end.

Base/usr/share/man/man7/setuid_overview.md
```diff
@@ -0,0 +1,71 @@
...
+
+Overview of real, effective, and saved user and group ids
+
+## Description
+
+Each process runs as user and a group. By default, a process started by a user has its user id set to that user and its group id set to that user's primary group.
+
+Each file has permissions that control if it can be read, written, and executed by its owner, other users in the owner's group, and by others. A process has the same access permissions as its user and its group.
+
+In addition to the access permissions bits, executables may have the "Set User ID"  and the "Set Group ID" bit set. Executables with these bits set run with the user or group id of the executable, instead of with the user and group id of the current user. These binaries are also called "SUID binaries" or "setuid binaries" (or "SGID binaries" or "setgid binaries" for the "Set Group ID" bit).
```
bugaevc, line 11: Looks like GitHub just lost one of my comments; I'm not gonna retype the whole thing but the point was that "Executables with these bits set run with the user or group id of the executable" is not quite true, it just affects what happens at exec time.

Base/usr/share/man/man2/geteuid.md
nico: How'd you organize it? Getters on one page, setters on another?

bugaevc: Honestly I'd just stuff everything here into one page 🙂

bugaevc:
To put it another way, I guess: your explanation of how UIDs work is IMO overly user-centric. It describes things as if users are actual people that run programs.

The way I think about it, there are processes that have certain tags, misleadingly named user and group IDs, slapped onto them, and those tags determine what they're allowed to access and do. There are no actual users (people) in the system. Some programs believe they're performing some logic at the request of some people (say, the shell reads keypresses and interprets that as the commands the person typing wants to launch), but that's entirely their own business. One program can even act at the request of *multiple* people at the same time, e.g. the WebServer does both what the system owner told it to do (serve those pages!), and what somebody else tells it to do over the internet (GET me those pages!). None of this matters to the kernel and to the UID/GID/permissions system.

Now, there are multiple ways to make use of the UID/GID mechanism, and understanding those use cases is important for understanding why the mechanism is designed this way, but they are just that, use cases for the mechanism, not the mechanism itself.

* You can make UIDs correspond to actual people, and GIDs to actual groups of people (say, "professors")
* You can make UIDs correspond to different applications, and GIDs to resources apps may want to access, then add UIDs of the apps that *should* be able to access a resource to that resource's group
* You can ignore the whole thing and run everything as root (say, on an embedded system)
* On a single-user system, you can have one "user account" (besides root) for everything, and temporary switch to the root account when performing administrative tasks, to make sure apps can't perform those tasks behind your back

and so on.

Now, Serenity is, at least for now, a single-user desktop system, and we *mostly* use UIDs/GIDs to limiting what different *services* have access to. But we also have "anon", who is assumed to be the person using SerenityOS, and "nona" who is an imaginary second person.

My point is, the man page should first describe the actual rules, in terms of processes, IDs and `execve`(2). Then it should state that "traditionally on Unix systems, each UID used to correspond to an actual person using the system..." and give an example of how access control ends up working in practice for that case. Then describe how this is used in SerenityOS.

Does this make sense?

Base/usr/share/man/man2/geteuid.md
nico: I claim that most people don't have the category numbers for man pages memorized :)

Base/usr/share/man/man2/getresuid.md
nico: Done. It was wrong i the code too, fixed there as well.

Base/usr/share/man/man2/seteuid.md
nico: Done.

Base/usr/share/man/man2/seteuid.md
nico: Ok, removed this line.

Base/usr/share/man/man2/setresuid.md
nico: Tweaked.

nico:
Thanks for the thorough review!

Makes sense. I fixed the minor nits, and will rewrite the text to address the bigger points in a bit :)

Base/usr/share/man/man2/geteuid.md
nico: I think I'd like to start with a bunch of pages. I kind of like it as-is at the moment. It's easy to consolidate later if we want to.

Base/usr/share/man/man7/setuid_overview.md
nico: Done.

nico:
I rewrote the overview to be more process-centric.

nico:
If this is close enough, I'd say let's merge this and you can rewrite more it in a follow-up if you want.

awesomekling:
Actually yeah that's fair. LGTM; @bugaevc feel free to follow up :)

bugaevc:
Yeah, I was about to suggest the same. It looks good enough, but we might want to do some additional tweaks later.

nico:
Thanks, all!

Issue #2671, opened 2020-06-30
FTP protocol


machinecp:
Hello guys in the functionalities of the operating system already have an FTP system?
I would like to work on it, although my strength is not the c ++ language, if they do not have I will try to do it, because I think they already have network support.

linusg:
No FTP support yet (I haven't looked if something in `Ports/` might qualify, but that doesn't count anyway :grin:) . There was some work towards it, but the PR has been closed by its author (@sppmacd) recently: https://github.com/SerenityOS/serenity/pull/2284

Also see https://github.com/SerenityOS/serenity/issues/2223 for some related discussion.

machinecp:
oh thank you very much :) I will review the issues

linusg:
Closing as dupe of #2223.

PR #2675, opened 2020-07-02
Plan9FS

bugaevc:
It's incomplete and somewhat crashy, but it works! :smile:

image

Here's how I'm testing it. First, it needs (remote) user name, and there's currently no way to pass that from user space:

```diff
diff --git a/Kernel/FileSystem/Plan9FileSystem.cpp b/Kernel/FileSystem/Plan9FileSystem.cpp
index 2a82e6505..81efa35ca 100644
--- a/Kernel/FileSystem/Plan9FileSystem.cpp
+++ b/Kernel/FileSystem/Plan9FileSystem.cpp
@@ -70,7 +70,7 @@ bool Plan9FS::initialize()
     Message attach_message { *this, Message::Type::Tattach };
     // FIXME: This needs a user name and an "export" name; but how do we get them?
     // Perhaps initialize() should accept a string of FS-specific options...
-    attach_message << root_fid << (u32)-1 << "" << "/";
+    attach_message << root_fid << (u32)-1 << "sergey" << "/";
     if (m_remote_protocol_version >= ProtocolVersion::v9P2000u)
         attach_message << (u32)-1;
```

Next, `mount`(8) currently has no support for this, so I use this utility for connecting & mounting it:

```cpp
#include <arpa/inet.h>
#include <errno.h>
#include <fcntl.h>
#include <netinet/in.h>
#include <stdio.h>
#include <sys/socket.h>
#include <unistd.h>

int main()
{
    int fd = socket(AF_INET, SOCK_STREAM, 0);
    if (fd < 0) {
        perror("socket");
        return 1;
    }

    struct sockaddr_in addr;
    addr.sin_family = AF_INET;
    int rc = inet_pton(AF_INET, "10.0.2.2", &addr.sin_addr);
    if (rc < 0) {
        perror("inet_pton");
        return 1;
    }
    addr.sin_port = htons(3542);
    rc = connect(fd, (const struct sockaddr*)&addr, sizeof(addr));
    if (rc < 0) {
        perror("connect");
        return 1;
    }

    rc = mount(fd, "/mnt", "9p", 0);
    if (rc < 0) {
        perror("mount");
        return 1;
    }

    fprintf(stderr, "Done!\n");
    return 0;
}
```

(adjust the host & port to your liking).

Then on the host, I run one of the 9P servers for Unix; for example Diod:

```
$ diod/diod --foreground --no-auth --export / --listen 0.0.0.0:3542
```

cc @alimpfard, @supercomputer7


bugaevc:
Caveat: errno numbers and stat modes are passed as-is; in 9P2000.L they're supposed to be Linux values, we seem to be using the same values so it seems to work.

Another caveat: it currently doesn't do any caching (and it's not clear how client-side caching should work), so it's kinda slow. Path resolution is especially slow, because for each path segment VFS asks the FS:

* Is it a symlink? (Fetch metadata)
* Is it a directory? (Fetch metadata again!)
* Resolve this child name

9P (9P2000) has protocol-level support for server-side path resolution (so you can pass it multiple path components, and the server would figure it out). If we want to make use of this, we'll need some API to let VFS know the FS would like to handle (some subset of) path resolution by itself.

alimpfard:
Very neat! 
looks like you're mounting a socket...which makes me curious about the general security implications of this
but that's a thing for later.

> it's not clear how client-side caching should work

:thinking: Is this just the client?
I remember you saying that 9p doesn't have any change notifications, nor the concept of watching inodes, so...caching is probably impossible in a general case?

bugaevc:
> looks like you're mounting a socket...which makes me curious about the general security implications of this

Well, it is a remote FS... don't mount file systems from servers you don't trust, or perhaps mount them `noexec,nosuid,nodev`. Also this implementation will actually panic if the messages coming from the server are malformed (so yeah, don't use it with untrusted servers yet); this is also something to improve upon.

> thinking Is this just the client?

Yes, `Kernel::Plan9FS` is a 9P client; it talks to a server that actually host the FS (or synthesizes it on the fly, or anything it wants). Yes, no change notifications. There's no 100% correct way to do caching; but perhaps something like caching data for a few seconds would be *correct enough* and speed up lookups significantly.

alimpfard:
> don't mount file systems from servers you don't trust

I was more concerned about MITM attacks, I don't see any way to detect messages that are tampered with, and it's sending the inode data over an untrusted channel.

> perhaps something like caching data for a few seconds would be correct enough and speed up lookups significantly

Yeah, that could work, I guess.

as an aside, a server component for this would be nice to have, we could use it as FUSE (and #2671 and co could benefit from it as well).

alimpfard:
> I use this utility for connecting & mounting it

why not rip `fusermount` off and just make a `9pfsmount` or something? I don't think `mount` should worry about _making_ the sockets.

bugaevc:
> I was more concerned about MITM attacks, I don't see any way to detect messages that are tampered with, and it's sending the inode data over an untrusted channel.

9P itself doesn't include any encryption or integrity protection. But you can forward it over TLS or something if both the client and the server agree. But for the cases we *actually* want to use it — the kernel talking to userspace on the same system (FUSE) and the kernel talking to QEMU (shared folders) we can be pretty sure no one's messing with the channel.

> why not rip fusermount off and just make a 9pfsmount or something?

I did name the utility `mount9p` (as visible in the screenshot). I *think* I'd rather have `mount`(8) handle it eventually rather than making a separate utility, but I'm not proposing to settle it one way or another now :slightly_smiling_face:

alimpfard:
> But for the cases we actually want to use it (...) we can be pretty sure no one's messing with the channel

I suppose so.

> I'm not proposing to settle it one way or another now

Yeah, let's keep _that_ debate for when we actually need it

about the user thing, does 9p _require_ that you give it a remote user? 
is that for authentication or...?

overall, it looks great!
with some caching, it could be pretty nice.

bugaevc:
> about the user thing, does 9p require that you give it a remote user? is that for authentication or...?

Yes... No, authentication is separate (and unimplemented for now, so tell your 9P server to not require it). I believe it's for access control, that is for the remote server to decide for what files/actions it should return `EPERM`. So for example if you have a bunch of users on the server, then the whole remote hierarchy is mounted on a machine that has (roughly) the same set of users, the server can do permission checking for you. We also have permission checking at VFS level; I haven't quite thought of how to make the two play nice (need to see what Linux does, and ideas welcome).

alimpfard:
Oh I see.
Then we _should_ just pass some args to `initialize()`.
what immediately comes to my mind is that if someone has access to the underlying socket, they should have any access that the remote gives.
in other words, whatever uid/gid owns the socket, owns everything in the remote fs (in our VFS's eyes)?

bugaevc:
It's not that simple... Say you're root@courage, you have the socket, and you even have the auth credentials (when we implement support for those). You mount the remote FS, but you want other UIDs on the same machine to be able to access it, but only according to the permissions tracked by the server. Other UIDs, say anon, don't have the socket and know the credentials, but they can access the FS because root has mounted it for them. For this to work, the client would have to *attach* multiple times, at least one per requesting UID, and one inode would correspond to multiple fids.

Or perhaps we don't want to support that; who knows. To make it clear, this implementation of Plan9FS is in no way complete, I avoided thinking about, and deciding on, many things.

alimpfard:
uhhh...I feel like I don't understand this enough, but it seems to me that just copying the socket's perms would be fine.
say root wants to allow anon to access the stuff, she would just make the socket as anon, or set it readable for anon, and that'd be the end of that.
but I'm clearly missing something here, since it looks like you're trying to map uids on one system to another.

bugaevc:
Two things:
* Plan9FS just needs a socket-like fd (readable and writable stream), it doesn't know or care if it's a TCP socket, a local socket, a virtio device, etc. — it doesn't control who has access to the socket (if it is even a socket), and it doesn't know, if it's a Unix socket accessible in the file system, what its UID/GID are.
* The idea (not mine, the idea of Plan 9 developers) with *attach* and local/remote users is that someone (say, anon) can, under controlled circumstances, access *some* of the files on a 9p-mounted remote filesystem — namely those that they're permitted to access by file permissions — even without having direct access to the 9P connection/socket/credentials. In your example, root wouldn't want to let anon access the socket directly, because that'd allow anon to read *everything*. No, root would *mount* the filesystem, and that would let anon access *some* of the files, subject to permissions checking. Just like with an on-disk filesystem (ext2), you don't give anon direct access to the disk device, because that would allow anon to read (& write) everything on the disk, no, you *mount* it and then anon can access *some* of the files via the VFS.

Linux lets you pick one of the few modes for this (<code>curl https://www.kernel.org/doc/Documentation/filesystems/9p.txt | grep access</code>). Currently Plan9FS does something like Linux's "client" or "any". But as I've said, I have not given this enough thought yet.

alimpfard:
I see, thanks for explaining that~

well, we can always make things better when the need arises~

Kernel/Process.cpp
```diff
@@ -232,12 +232,14 @@ Region* Process::allocate_region_with_vmobject(VirtualAddress vaddr, size_t size
 
 bool Process::deallocate_region(Region& region)
 {
+    OwnPtr<Region> region_protector;
     ScopedSpinLock lock(m_lock);
```
awesomekling, line 237: The commit message here seems inaccurate. This makes sure we drop the region with `Process::m_lock` unlocked, but has nothing to do with CPU interrupts. Or am I missing some bigger picture detail here?

Kernel/FileSystem/Plan9FileSystem.cpp
```diff
@@ -0,0 +1,752 @@
...
+            m_tags_to_ignore.remove(header.tag);
+        } else {
+            dbg() << "Received a 9p message of type " << header.type << " with an unexpected tag " << header.tag << ", dropping";
+        }
+        return KSuccess;
+    }
+    ReceiveCompletion& completion = *optional_completion.value();
+    completion.result = KSuccess;
+    completion.message.~Message();
+    new (&completion.message) Message(move(buffer));
```
awesomekling, line 305: Can we write this as a regular assignment, or does the placement new sidestep something here?

Kernel/FileSystem/Plan9FileSystem.h
```diff
@@ -0,0 +1,352 @@
...
+            Rclunk = 121,
+            Tremove = 122,
+            Rremove = 123,
+            Tstat = 124,
+            Rstat = 125,
+            Twstat = 126,
+            Rwstat = 127
+        };
+
+        class Decoder {
```
awesomekling, line 149: Does this really need to be in the header file? This header seems pretty bloated with implementation details..

Kernel/Process.cpp
bugaevc: Ah; this was an `InterruptDisabler` and not a spinlock when I wrote the code (before the SMP PR got merged). But it looks like a spinlock still disables interrupts:

https://github.com/SerenityOS/serenity/blob/e7393bfb7b4cc8a256f94bf7f8333bb2d17ca4e0/Kernel/SpinLock.h#L122

So I kept this. I haven't tested if it now works without it, perhaps I'm wrong about spinlock+interrupts and it does.

Kernel/FileSystem/Plan9FileSystem.h
bugaevc: No, not really. And the header being way too bloated is one of the problems I see with this PR. I'll see if it's possible to move some of this into .cpp then.

Kernel/FileSystem/Plan9FileSystem.cpp
bugaevc: No regular assignment unfortunately ☹️. The reason is, because of the `union` (and `KBuffer` which doesn't have a null state — or perhaps it does, hmm) you cannot assign or move `Message`s. With explicit destructor + placement new, we're not assigning or moving anything, so it works.

That being said, I need to actually experiment with overriding `operator=`, perhaps that would be not super ugly and clean up some code like this.

Kernel/Process.cpp
awesomekling: Ah, no you are right. I'm silly :)

Kernel/FileSystem/Plan9FileSystem.h
bugaevc: Moved all of `Message` (including `Message::Decoder`) into .cpp

Kernel/FileSystem/Plan9FileSystem.cpp
bugaevc: `operator=()` actually seems to work

awesomekling:
All right, let's merge this and keep iterating. In particular, having a way to pass arbitrary file system options to the kernel seems like a valuable next step. :)

PR #2676, opened 2020-07-02
SystemMonitor: Correctly check error of posix_spawn()

nico:
Hooray for using posix_spawn()!

I learned about this API peculiarity long after I started using it
(but before I implemented it in Serenity :) ).

posix_spawn() has a bit of a whacky API in that it doens't return
0 on success and -1 on failure while writing the error code to
errno -- it instead returns 0 on success and the error code on
error.

So assign the return value to errno so that perror() does the
right thing.


awesomekling:
Oopsie! Good catch. I should have used better ink for my mental notes about `posix_spawn()` return values :D

PR #2677, opened 2020-07-02
LibC: Add isblank()

linusg:
This is required to build the "editline" port.

cc @The-King-of-Toasters - how did this ever work for you in #2532?

### DIFF
--- a/Libraries/LibC/ctype.h
+++ b/Libraries/LibC/ctype.h
@@ -51,6 +51,7 @@ int iscntrl(int);
 int isdigit(int);
 int isxdigit(int);
 int isspace(int);
+int isblank(int);
 int ispunct(int);
 int isprint(int);
 int isgraph(int);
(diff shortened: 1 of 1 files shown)


awesomekling:
This will break the GCC port. We had `isblank()` in the past but it had to be reverted in 14bd978928496baf00483db4602cc0e976ba3a70

There's something in the details about `isblank()` being a C99 addition that we need to take into account.

linusg:
Summary from IRC: This was implemented before (https://github.com/SerenityOS/serenity/commit/4e79a60b78d46869e3edb52d8eb4fc0e557e80d3) and then reverted (https://github.com/SerenityOS/serenity/commit/14bd978928496baf00483db4602cc0e976ba3a70) so this isn't going to work. Question remains how to build `editline` w/o `isblank()`.

The-King-of-Toasters:
Oh yeah, I forgot I implemented a version of isblank that just checked for ' ' and '\t', oops.

PR #2681, opened 2020-07-03
LibJS: Format tests and output + QOL changes

mattco98:
QOL changes:

- Hide exception debug output behind a flag. This makes running the test suite significantly faster from inside Serenity
- Remove `MarkedArgumentList.values().append()`, which I for some reason used all over the place in my code

Test changes:

- Reorganize the tests into subfolders. Not every file is in a subfolder. It's probably not perfect, but any file can be moved to a different location and the test suite will still run fine, so I'm not too worried about it.
- Use emojis instead of the pass/fail text. I'm willing to revert this, but I think it looks good :)
  - Created two emojis for this in `/res/emojis`
- Removed extra colon in the test output

To support the changes, `cut` can now read from stdin!

New test output preview:

preview

And the same from inside Serenity:

 preview


Libraries/LibJS/Interpreter.cpp
```diff
@@ -299,7 +301,7 @@ Value Interpreter::construct(Function& function, Function& new_target, Optional<
 
 Value Interpreter::throw_exception(Exception* exception)
 {
-#ifdef __serenity__
+#if defined(__serenity__) and defined(INTERPRETER_DEBUG)
```
linusg, line 304: Let's remove the `__serenity__` definition check, I only added that as it's writing stdout and dbg output to the same terminal on the host, but having this behind a single define is enough IMO.

awesomekling:
I'd say that's a tasteful use of emojis.

PR #2682, opened 2020-07-03
new tokens in Cpplexer

DenisCampredon:
While looking out at serenityos I noticed that some tokens were not lexed in HackStudio. I tried to fix that with all the tokens I remembered.

Also I borrowed the new tokens names from the js lexer.

### DIFF
--- a/Libraries/LibGUI/CppLexer.cpp
+++ b/Libraries/LibGUI/CppLexer.cpp
@@ -40,9 +40,8 @@ CppLexer::CppLexer(const StringView& input)
 char CppLexer::peek(size_t offset) const
 {
     while ((m_index + offset) < m_input.length() && m_input[m_index + offset] == '\\'
-        && (m_index + offset + 1) < m_input.length() && m_input[m_index + offset + 1] == '\n') {
+           && (m_index + offset + 1) < m_input.length() && m_input[m_index + offset + 1] == '\n')
         offset += 2;
-    }
     if ((m_index + offset) >= m_input.length())
         return 0;
     return m_input[m_index + offset];
@@ -67,9 +66,8 @@ char CppLexer::consume()
     if (ch == '\n') {
         m_position.line++;
         m_position.column = 0;
-    } else {
+    } else
         m_position.column++;
-    }
     return ch;
 }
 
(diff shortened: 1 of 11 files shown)


Libraries/LibGUI/CppSyntaxHighlighter.cpp
```diff
@@ -56,9 +56,60 @@ static TextStyle style_for_token_type(Gfx::Palette palette, CppToken::Type type)
...
+    case CppToken::Type::LessThanEquals:
+    case CppToken::Type::GreaterThan:
+    case CppToken::Type::GreaterThanEquals:
+    case CppToken::Type::ShiftLeft:
+    case CppToken::Type::ShiftLeftEquals:
+    case CppToken::Type::ShiftRight:
+    case CppToken::Type::ShiftRightEquals:
+    case CppToken::Type::QuestionMark:
+    case CppToken::Type::Colon:
+        return {palette.syntax_operator()};
```
awesomekling, line 94: Please use `clang-format`.

Libraries/LibGUI/CppLexer.cpp
```diff
@@ -38,6 +38,9 @@ CppLexer::CppLexer(const StringView& input)
 
 char CppLexer::peek(size_t offset) const
 {
+    while ((m_index + offset) < m_input.length() && m_input[m_index + offset] == '\\'
+           && (m_index + offset + 1) < m_input.length() && m_input[m_index + offset + 1] == '\n')
+        offset += 2;
```
awesomekling, line 44: Please use `{` and `}` here.

Libraries/LibGUI/CppLexer.cpp
```diff
@@ -46,14 +49,25 @@ char CppLexer::peek(size_t offset) const
...
+        return 0;
+    }
+    char ch = m_input[m_index++];
     if (ch == '\n') {
         m_position.line++;
         m_position.column = 0;
-    } else {
+    } else
         m_position.column++;
-    }
```
awesomekling, line 56: Please use `{` and `}` for this `else` to match the `if`.

PR #2513, opened 2020-06-05
Kernel: SMP: Initial changes to make kernel more processor-aware

tomuta:
This is still work in progress:

- [x] Add SpinLock class
- [x] Serialize debug output with new SpinLock class
- [x] Add `Processor` structure and assign it to each CPU using the `fs` segment register
- [x] Get `Scheduler` to work on the BSP using software context switching
- [x] Add some locks to MM, Process, and other places
- [x] Boot each AP all the way into the `Scheduler::idle_loop`
- [x] Expose all processors in `/proc/cpuinfo`
- [x] Add CPU graphs and current processor information to SystemMonitor

PR #2683, opened 2020-07-03
Kernel: Fix signal delivery

tomuta:
This fixes a regression with signal delivery (e.g. for strace) introduced by #2513

### DIFF
--- a/AK/Atomic.h
+++ b/AK/Atomic.h
@@ -46,6 +46,18 @@ static inline T atomic_exchange(volatile T* var, T desired, MemoryOrder order = 
     return __atomic_exchange_n(var, desired, order);
 }
 
+template<typename T, typename V = typename RemoveVolatile<T>::Type>
+static inline V* atomic_exchange(volatile T** var, V* desired, MemoryOrder order = memory_order_seq_cst) noexcept
+{
+    return __atomic_exchange_n(var, desired, order);
+}
+
+template<typename T, typename V = typename RemoveVolatile<T>::Type>
+static inline V* atomic_exchange(volatile T** var, std::nullptr_t, MemoryOrder order = memory_order_seq_cst) noexcept
+{
+    return __atomic_exchange_n(const_cast<V**>(var), nullptr, order);
+}
+
 template<typename T>
 static inline bool atomic_compare_exchange_strong(volatile T* var, T& expected, T desired, MemoryOrder order = memory_order_seq_cst) noexcept
 {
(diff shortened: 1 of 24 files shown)


AK/Atomic.h
```diff
@@ -31,6 +31,9 @@
 
 namespace AK {
 
+template <typename T> struct remove_volatile { typedef T type; };
+template <typename T> struct remove_volatile<volatile T> { typedef T type; };
```
awesomekling, line 35: Please put these together with the other `RemoveFoo` helpers in `AK/StdLibExtras.h` :)

AK/LogStream.h
```diff
@@ -57,17 +58,66 @@ class LogStream {
 #endif
 };
 
-class DebugLogStream final : public LogStream {
+class BufferedLogStream : public LogStream {
+    mutable int m_size{0};
```
awesomekling, line 62: Please use `clang-format`.

AK/LogStream.h
```diff
@@ -57,17 +58,66 @@ class LogStream {
...
+    mutable int m_capacity{128};
+    union
+    {
+        mutable u8* m_buffer{nullptr};
+        mutable u8 m_local_buffer[128];
+    } u;
+
+    void grow(int byte_needed) const
```
awesomekling, line 70: s/byte_needed/bytes_needed/ (or perhaps "capacity")
Also, this seems like it should be a `size_t` instead of an `int`.

AK/LogStream.h
```diff
@@ -57,17 +58,66 @@ class LogStream {
...
+        }
+        u.m_buffer = new_data;
+        m_capacity = new_capacity;
+    }
+protected:
+    u8* data() const
+    {
+        if (m_capacity <= (int)sizeof(u.m_local_buffer))
+            return u.m_local_buffer;
+        else
```
awesomekling, line 88: No need for `else` after `return`.

AK/LogStream.h
```diff
@@ -57,17 +58,66 @@ class LogStream {
 #endif
 };
 
-class DebugLogStream final : public LogStream {
+class BufferedLogStream : public LogStream {
+    mutable int m_size{0};
+    mutable int m_capacity{128};
```
awesomekling, line 63: Let's make this a `size_t`.

AK/LogStream.h
tomuta: I initially had it as `size_t` but ended up changing it to `int` because the `write` function takes an `int`, so I ended up with lots of casting.

AK/Atomic.h
tomuta: Ah there they are. They're already there without me realizing it...

AK/LogStream.h
tomuta: Yeah you're right, we can change it. I think it had a slightly different implementation before where it being `int` was easier. I'll change it.

tomuta:
I hope this looks better

Kernel/Arch/i386/CPU.cpp
```diff
@@ -645,181 +646,215 @@ void sse_init()
...
+    ASSERT(procs.size() > cpu);
+    return *procs[cpu];
+}
+
+
+void Processor::cpu_detect()
+{
+    // NOTE: This is called during Processor::early_initialize, we cannot
+    //       safely log at this point because we don't have kmalloc
+    //       initialized yet!
```
awesomekling, line 715: `kprintf()` should still work I would think :)

PR #2687, opened 2020-07-03
Kernel: Fix /proc/PID/stack

### LINKED ISSUE #2678: Kernel: /proc/PID/stack, /proc/PID/regs broken
cc @tomuta 

These broke with the SMP changes. They were relying on up-to-date register information in Thread::tss().

Note that the current process can still get its own stack trace by opening /proc/PID/stack, but other processes cannot be inspected. Easily noticed in SystemMonitor, by trying to look at stacks for processes other than SystemMonitor itself.

tomuta:
Fixes #2678


awesomekling:
Excellent! Thank you :)

PR #2689, opened 2020-07-03
LibJS: Completely refactor the test system

mattco98:
This is a complete overhaul of the LibJS test system. 

First and foremost, ~the bulk of the~ **all** test execution has now been moved to the new `test-js` userspace program. ~A user still has to run `run-tests.sh` in order to run the tests, however this is just because it's a bit easier to get the list of files. This can be moved to the `test-js` program fully in the future.~

Secondly, the tests themselves have been completely rewritten to follow the jest style of testing (if you are not familiar with jest, I have converted ~10 tests already, so have a look at those). Because we are now doing testing through via a userspace program, we can ditch the purely stdin/stdout based communication in favor of communicating via a JS object. Communication is done via a special JS variable called `__TestResults__`, which is inspected after every test file is ran.

Finally, the test output has been completely overhauled (again...). This followed a similar format to the jest command line utility. 

This PR is far from complete, as it will take quite a while to convert all of the tests. ~Right now, I'm only looking for feedback on the new style of testing, as well as the test output. I'm not looking for feedback on `test-js.cpp` or `test-common.js`, as they are far from being finalized.~

TODO:

- [X] Get rid of `run-tests.sh`
- [X] Implement `test.skip` to allow people to write tests for functionality that should work, but doesn't. This will remove a lot of `// FIXME: ` notes within the tests.
- [X] Show total files at the end of the test output
- [X] Handle console output better
- [X] Only show file output for files which have failed or skipped tests (not applicable with `-t`)

<details>
<summary>Sample output: </summary>
<br />

</details>

<details>
<summary>And from inside Serenity: </summary>
<br />

</details>

<details>
<summary>Also, individual test durations with <code>-t</code> for profiling!</summary>
<br />

</details>

cc @linusg @nooga


Libraries/LibJS/Tests/automatic-semicolon-insertion.js
```diff
@@ -1,30 +1,26 @@
...
+if (1)
+    return 1
+else
+    return 0;`;
 
-    if (1)
-        return 1
-    else
-        return 0
+    expect(source).toCompile();
```
linusg, line 18: This is a great API, but I think the naming can be improved, as we're not really "compiling" anything - maybe `toParseWithoutErrors()`?

Libraries/LibJS/Tests/automatic-semicolon-insertion.js
```diff
@@ -43,30 +39,32 @@ function foo() {
...
-    assert(foo() === undefined);
-    assert(baz() === 5);
+return counter;`;
 
-    console.log("PASS");
-} catch (e) {
-    console.log("FAIL: " + e);
-}
+    expect(source).toCompile();
+    expect(new Function(source)()).toBe(5);
```
linusg, line 65: I'd suggest we add a shortcut for this, similar to `eval` - maybe `run()` or `evaluate()`? Could be native C++ or just `new Function(source)()`, but makes the test a bit easier to read and understand IMO.

E.g.

```suggestion
    expect(evaluate(source)).toBe(5);
```

Libraries/LibJS/Tests/debugger-statement.js
```diff
@@ -1,9 +1,3 @@
...
-
-try {
-    debugger;
-
-    console.log("PASS");
-} catch (e) {
-    console.log("FAIL: " + e);
-}
+test("debugger keyword", () => {
+    expect("debugger").toCompile();
```
linusg, line 2: I feel like we should distinguish between "parses without errors" and "runs without errors" - previously it would actually execute the statement. Anything against keeping that behaviour for this test?

OTOH `debugger` is somewhat special and being able to execute it without any effects might be counter-intuitive, so I understand if this was a concious decision :)

Libraries/LibJS/Tests/empty-statements.js
```diff
@@ -1,11 +1,19 @@
...
+    expect("if (false); else if (false); else;").toCompile();
+});
+
+test("while with no body", () => {
+    expect("while (false);").toCompile();
+});
+
+test("do while with no body", () => {
+    expect("do; while (false);").toCompile();
+});
```
linusg, line 19: Same here, behaviour changes from actually running the stuff to "just" parsing it.

Libraries/LibJS/Tests/exponentiation-basic.js
```diff
@@ -1,30 +1,37 @@
...
+    
+    expect("-2 ** 3").not.toCompile();
+});
 
-    console.log("PASS");
-} catch (e) {
-    console.log("FAIL: " + e);
-}
+test("exponentation with non-numeric primitives", () => {
+
```
linusg, line 21: Extra empty line

Libraries/LibJS/Tests/exponentiation-basic.js
```diff
@@ -1,30 +1,37 @@
...
+    expect("2" ** "3").toBe(8);
+    expect("" ** []).toBe(1);
+    expect([] ** null).toBe(1);
+    expect(null ** null).toBe(1);
+    expect(undefined ** null).toBe(1);
+})
+
+test("exponentation that produces NaN", () => {
```
linusg, line 29: being able to form groups now is super cool!

Libraries/LibJS/Tests/exception-ReferenceError.js
```diff
@@ -1,9 +1,3 @@
...
-
-try {
-    i < 3;
-} catch (e) {
-    assert(e.name === "ReferenceError");
-}
-
-console.log("PASS");
+test("unknown variable produces ReferenceError", () => {
+    expect(new Function("i < 3")).toThrow(ReferenceError);
```
linusg, line 2: Umm, if that's throwing that's a bug - just constructing the function object should not throw yet, only calling it should.

I.e.

```suggestion
    expect(new Function("i < 3")()).toThrow(ReferenceError);
```

(or rather the shortcut proposed above :) )

Libraries/LibJS/Tests/labels.js
```diff
@@ -1,20 +1,22 @@
-load("test-common.js");
-
-try {
+test("labeled plain scope", () => {
     test: {
         let o = 1;
-        assert(o === 1);
+        expect(o).toBe(1);
```
linusg, line 4: In other files you're using `expect.fail()`, let's stick with one :)

Libraries/LibJS/Tests/tagged-template-literals.js
```diff
@@ -1,100 +1,106 @@
...
+    test("template object structure", () => {
+        const getTemplateObject = (...rest) => rest;
+        const getRawTemplateStrings = arr => arr.raw;
+        
+        let o = getTemplateObject`foo\nbar`;
+        expect(Object.getOwnPropertyNames(o[0])).toContain("raw");
+        
+        let raw = getRawTemplateStrings`foo${1 + 3}\nbar`;
+        expect(Object.getOwnPropertyNames(raw)).not.toContain("raw");
+        expect(raw.length).toBe(2);
```
linusg, line 101:
```suggestion
        expect(raw).toHaveLength(2);
```

Libraries/LibJS/Tests/labels.js
linusg: Whoops, wrong line. A bit below (it's `expect().fail()`)

Libraries/LibJS/Tests/exception-ReferenceError.js
mattco98: `toThrow` operates on a function target, and does the calling itself :) however, I wrote this really weird, so I understand the confusion. It should be `expect(() => { i < 3; }).toThrow(ReferenceError)`.

Libraries/LibJS/Tests/labels.js
mattco98: Oops, `expect.fail()` is wrong one. Nice catch

Libraries/LibJS/Tests/exception-ReferenceError.js
linusg: oh I see, what I proposed doesn't really make sense lol - my bad :D

Libraries/LibJS/Tests/empty-statements.js
linusg: ...this might or might not be what we want, open to suggestions :)

mattco98:
Updated the screenshots to be the current look of the tests, both on the host and from inside Serenity. Does anyone know why the warning emoji (⚠️) has a question mark next to it? I know that means unknown character, but it's strange that it would print both the emoji _and_ the question mark. The emoji is printed in `Userland/test-js.cpp` line 485, if anyone wants to take a look.

linusg:
@mattco98 that's "emoji variation selector" character - I can highly recommend this site: https://www.babelstone.co.uk/Unicode/whatisit.html

Copy and paste the line into it and you'll see:

image

That's because some emojis have a non-emoji variant (usually just an outline and same colour as the surrounding text), and combining that with this character makes it a colourful "proper" emoji. Serenity doesn't support that, goes full colour straight away and ignores the unknown next character.

linusg:
here's just "WARNING SIGN": `⚠`
with emoji variant: `⚠️`

mattco98:
Oh interesting, I didn't know that! I'll definitely be bookmarking that site. Thanks :)

linusg:
Weird bug: when a test case is just called "toString", "toLocaleString" or "valueOf" (possibly more), `test-js` crashes:

```js
test("toString", () => {});
```

```
test-js: /home/linus/Dev/git/serenity/Libraries/LibJS/Runtime/Value.cpp:86: JS::Array& JS::Value::as_array(): Assertion `is_array()' failed.
```

We can sell it as a feature for now ("just name your tests properly!"), but it would be nice to not have this restriction eventually :D

mattco98:
That's a pretty funny issue actually, it was colliding with `Object.prototype.toString`. I just set the prototype of the test result object to `null`, so that's fixed.

Also, I think I've finalized `test-js.cpp` and `test-common.js`, as well as all of the test output formatting. All of my TODOs are done, so I'll take any feedback at all :) now all that's really left are the rest of the tests.

linusg:
> That's a pretty funny issue actually, it was colliding with Object.prototype.toString. I just set the prototype of the test result object to null, so that's fixed.

A test named any of the above mentioned keywords still fails for me, although it's not a hard crash anymore.

mattco98:
Well, that's a bit odd, I'm fine with not a hard crash, since it's a really obscure problem. And as you said, the test author can just pick a better name :)

linusg:
@mattco98 I saw you added some prettier formatting, which is nice... but can we please not have 2-spaces-indented JS? I'm not against that in general, but I hate mixing indentations in the same codebase, and given Serenity C++ is 4 spaces I'd like to have the same for JS :)

mattco98:
Oh, sure. I don't share that opinion, but I also don't really care. It's just that two spaces is more common for JS code.

PR #2694, opened 2020-07-04
Help: Search-as-you-type and full text search

### LINKED ISSUE #2685: Help: Search functionality
GNOME Help for example has search functionality and it's incredibly helpful:

image

Would be nice to have this in Serenity's Help application as well. Maybe not a fancy dropdown with updates as you type, but a toolbar button, dialog with text input and then list of search results would be a nice start.

cc @bugaevc :slightly_smiling_face:

alimpfard:
This closes #2685.

~~There is a minor issue where the scrollbar on the ListView is updated _after_ the contents change, but I couldn't figure out how to fix that.~~

cc for people that I know are interested: @bugaevc @linusg @awesomekling

### DIFF
--- a/AK/MappedFile.h
+++ b/AK/MappedFile.h
@@ -33,8 +33,9 @@ namespace AK {
 
 class MappedFile {
     AK_MAKE_NONCOPYABLE(MappedFile);
+
 public:
-    MappedFile() {}
+    MappedFile() { }
     explicit MappedFile(const StringView& file_name);
     MappedFile(MappedFile&&);
     ~MappedFile();
@@ -42,6 +43,11 @@ public:
     MappedFile& operator=(MappedFile&&);
 
     bool is_valid() const { return m_map != (void*)-1; }
+    int errno_if_invalid() const
+    {
+        ASSERT(!is_valid());
+        return m_errno;
+    }
     void unmap();
 
     void* data() { return m_map; }
@@ -51,6 +57,7 @@ public:
 private:
     size_t m_size { 0 };
     void* m_map { (void*)-1 };
+    int m_errno { 0 };
 };
 
 }
--- a/AK/StringView.cpp
+++ b/AK/StringView.cpp
@@ -178,6 +178,11 @@ bool StringView::contains(char needle) const
     return false;
 }
 
+bool StringView::contains(const StringView& needle) const
+{
+    return memmem(m_characters, m_length, needle.m_characters, needle.m_length) != nullptr;
+}
+
 bool StringView::equals_ignoring_case(const StringView& other) const
 {
     return StringUtils::equals_ignoring_case(*this, other);
--- a/Applications/Help/ManualModel.cpp
+++ b/Applications/Help/ManualModel.cpp
@@ -28,6 +28,9 @@
 #include "ManualNode.h"
 #include "ManualPageNode.h"
 #include "ManualSectionNode.h"
+#include <AK/ByteBuffer.h>
+#include <LibCore/File.h>
+#include <LibGUI/FilteringProxyModel.h>
 
 static ManualSectionNode s_sections[] = {
     { "1", "Command-line programs" },
@@ -74,6 +77,25 @@ String ManualModel::page_path(const GUI::ModelIndex& index) const
         return {};
     auto* page = static_cast<const ManualPageNode*>(node);
     return page->path();
+}
+
+Result<StringView, int> ManualModel::page_view(const String& path) const
+{
+    if (path.is_empty())
+        return StringView {};
+
+    auto mapped_file = m_mapped_files.get(path);
+    if (mapped_file.has_value())
+        return StringView { (const char*)mapped_file.value()->data(), mapped_file.value()->size() };
+
+    auto map = make<MappedFile>(path);
+    if (!map->is_valid())
+        return map->errno_if_invalid();
+
+    StringView view { (const char*)map->data(), map->size() };
+    m_mapped_files.set(path, move(map));
+
+    return view;
 }
 
 String ManualModel::page_and_section(const GUI::ModelIndex& index) const
@@ -137,6 +159,10 @@ GUI::Variant ManualModel::data(const GUI::ModelIndex& index, Role role) const
 {
     auto* node = static_cast<const ManualNode*>(index.internal_data());
     switch (role) {
+    case Role::Search:
+        if (!node->is_page())
+            return {};
+        return String(page_view(page_path(index)).value());
     case Role::Display:
         return node->name();
     case Role::Icon:
@@ -154,6 +180,15 @@ void ManualModel::update_section_node_on_toggle(const GUI::ModelIndex& index, co
 {
     auto* node = static_cast<ManualSectionNode*>(index.internal_data());
     node->set_open(open);
+}
+
+TriState ManualModel::data_matches(const GUI::ModelIndex& index, GUI::Variant term) const
+{
+    auto view_result = page_view(page_path(index));
+    if (view_result.is_error() || view_result.value().is_empty())
+        return TriState::False;
+
+    return view_result.value().contains(term.as_string()) ? TriState::True : TriState::False;
 }
 
 void ManualModel::update()
--- a/Applications/Help/main.cpp
+++ b/Applications/Help/main.cpp
@@ -26,7 +26,6 @@
 
 #include "History.h"
 #include "ManualModel.h"
-#include <AK/ByteBuffer.h>
 #include <AK/URL.h>
 #include <LibCore/File.h>
 #include <LibDesktop/Launcher.h>
@@ -34,11 +33,14 @@
 #include <LibGUI/Action.h>
 #include <LibGUI/Application.h>
 #include <LibGUI/BoxLayout.h>
+#include <LibGUI/FilteringProxyModel.h>
+#include <LibGUI/ListView.h>
 #include <LibGUI/Menu.h>
 #include <LibGUI/MenuBar.h>
 #include <LibGUI/MessageBox.h>
 #include <LibGUI/Splitter.h>
-#include <LibGUI/TextEditor.h>
+#include <LibGUI/TabWidget.h>
+#include <LibGUI/TextBox.h>
 #include <LibGUI/ToolBar.h>
 #include <LibGUI/ToolBarContainer.h>
 #include <LibGUI/TreeView.h>
@@ -100,10 +102,41 @@ int main(int argc, char* argv[])
 
     auto model = ManualModel::create();
 
-    auto& tree_view = splitter.add<GUI::TreeView>();
+    auto& left_tab_bar = splitter.add<GUI::TabWidget>();
+    auto& tree_view = left_tab_bar.add_tab<GUI::TreeView>("Tree");
+    auto& search_view = left_tab_bar.add_tab<GUI::Widget>("Search");
+    search_view.set_layout<GUI::VerticalBoxLayout>();
+    auto& search_box = search_view.add<GUI::TextBox>();
+    auto& search_list_view = search_view.add<GUI::ListView>();
+    search_box.set_preferred_size(0, 20);
+    search_box.set_size_policy(GUI::SizePolicy::Fill, GUI::SizePolicy::Fixed);
+    search_box.set_text("Search...");
+    search_box.on_focusin = [&] {
+        if (search_box.text() == "Search...")
+            search_box.set_text("");
+    };
+    search_box.on_change = [&] {
+        if (auto model = search_list_view.model()) {
+            auto& search_model = *static_cast<GUI::FilteringProxyModel*>(model);
+            search_model.set_filter_term(search_box.text());
+            search_model.update();
+        }
+    };
+    search_box.on_focusout = [&] {
+        if (search_box.text().is_empty()) {
+            if (auto model = search_list_view.model()) {
+                auto& search_model = *static_cast<GUI::FilteringProxyModel*>(model);
+                search_model.set_filter_term("");
+            }
+            search_box.set_text("Search...");
+        }
+    };
+    search_list_view.set_model(GUI::FilteringProxyModel::construct(model));
+    search_list_view.model()->update();
+
     tree_view.set_model(model);
-    tree_view.set_size_policy(GUI::SizePolicy::Fixed, GUI::SizePolicy::Fill);
-    tree_view.set_preferred_size(200, 500);
+    left_tab_bar.set_size_policy(GUI::SizePolicy::Fixed, GUI::SizePolicy::Fill);
+    left_tab_bar.set_preferred_size(200, 500);
 
     auto& page_view = splitter.add<Web::PageView>();
 
@@ -123,19 +156,13 @@ int main(int argc, char* argv[])
             return;
         }
 
-        dbg() << "Opening page at " << path;
-
-        auto file = Core::File::construct();
-        file->set_filename(path);
-
-        if (!file->open(Core::IODevice::OpenMode::ReadOnly)) {
-            int saved_errno = errno;
-            GUI::MessageBox::show(window, strerror(saved_errno), "Failed to open man page", GUI::MessageBox::Type::Error);
+        auto source_result = model->page_view(path);
+        if (source_result.is_error()) {
+            GUI::MessageBox::show(window, strerror(source_result.error()), "Failed to open man page", GUI::MessageBox::Type::Error);
             return;
         }
-        auto buffer = file->read_all();
-        StringView source { (const char*)buffer.data(), buffer.size() };
 
+        auto source = source_result.value();
         String html;
         {
             auto md_document = Markdown::Document::parse(source);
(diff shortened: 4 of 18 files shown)


Libraries/LibGUI/ModelIndex.h
```diff
@@ -68,6 +71,7 @@ class ModelIndex {
     int m_row { -1 };
     int m_column { -1 };
     void* m_internal_data { nullptr };
+    void* m_role_specific_data { nullptr };
```
bugaevc, line 74: So what is this role-specific data, exactly? It's specific to *what* role, seems like it's only one member of data? And how's it different from the already existing internal data?

Libraries/LibGUI/FilteringProxyModel.cpp
```diff
@@ -0,0 +1,126 @@
...
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include <LibGUI/FilteringProxyModel.h>
+#include <LibGUI/TextBox.h>
+
+namespace GUI {
+
+FilteringProxyModel::FilteringProxyModel(NonnullRefPtr<TextBox> textbox, NonnullRefPtr<Model> model)
```
bugaevc, line 32: Don't pass ref ptrs up, just pass `&`s

Libraries/LibGUI/FilteringProxyModel.cpp
```diff
@@ -0,0 +1,126 @@
...
+    : m_model(move(model))
+    , m_textbox(move(textbox))
+{
+    m_textbox->on_change = [&] {
```
bugaevc, line 36: Hold on, why does the model hook into the view??... That looks very wrong.

Libraries/LibGUI/ModelIndex.h
alimpfard: it's specific to whatever role that was passed to `Model::data()`.
perhaps changing the signature of that function is the more sensible thing to do?

the internal data is...well, internal to the model that owns the index, this thing is a part of is public interface.

Libraries/LibGUI/FilteringProxyModel.cpp
```diff
@@ -0,0 +1,126 @@
...
+        if (m_is_searching) {
+            m_search_term = m_textbox->text();
+            search();
+        }
+    };
+}
+
+ModelIndex FilteringProxyModel::index(int row, int column, const ModelIndex& parent_index) const
+{
+    return m_model->index(row, column, parent_index);
```
bugaevc, line 46: So, does it pass everything indexes through unmodified? Shouldn't it, like, map them according to "matching indices" or something?

Libraries/LibGUI/FilteringProxyModel.cpp
alimpfard: to update-as-you-type?
I guess a `update_filter` would be smarter, I'll do it that way.

Libraries/LibGUI/FilteringProxyModel.cpp
```diff
@@ -0,0 +1,126 @@
...
+
+    return m_model->column_count(m_matching_indices[index.row()]);
+}
+
+Variant FilteringProxyModel::data(const ModelIndex& index, Role role) const
+{
+    if (!index.is_valid())
+        return {};
+
+    if ((size_t)index.row() > m_matching_indices.size() || index.row() < 0)
```
bugaevc, line 70: I believe `index.row() < 0` is already caught by `index.is_valid()` check above; but it doesn't hurt to check twice either.

Libraries/LibGUI/FilteringProxyModel.cpp
```diff
@@ -0,0 +1,126 @@
...
+        return 0;
+
+    return m_model->data(m_matching_indices[index.row()], role);
+}
+
+void FilteringProxyModel::update()
+{
+    m_model->update();
+    did_update();
```
bugaevc, line 79: Don't we want to re-filter here?

Libraries/LibGUI/FilteringProxyModel.cpp
alimpfard: :thinking: 
that's...actually an interesting point, now I'm wondering how it works at all...

Libraries/LibGUI/FilteringProxyModel.cpp
```diff
@@ -0,0 +1,126 @@
...
+    m_matching_indices.clear();
+    SearchData search_data { m_search_term, m_full_search };
+
+    Function<void(ModelIndex&)> add_matching = [&](ModelIndex& parent_index) {
+        for (auto i = 0; i < m_model->row_count(parent_index); ++i) {
+            auto index = m_model->index(i, 0, parent_index);
+            if (!index.is_valid())
+                continue;
+
+            index.set_role_specific_data(&search_data);
```
bugaevc, line 94: I, uh, don't think I understand or like this. Mutating indexes after they've been created, from the outside?..

Libraries/LibGUI/FilteringProxyModel.cpp
```diff
@@ -0,0 +1,126 @@
...
+                m_matching_indices.append(index);
+
+            add_matching(index);
+        }
+    };
+
+    ModelIndex parent_index;
+    add_matching(parent_index);
+}
+ModelIndex FilteringProxyModel::translate(const ModelIndex& index) const
```
bugaevc, line 106: Missing blank line

Libraries/LibGUI/ModelIndex.h
alimpfard: Oh right, changing the signature of `Model::data` was too much of a chore, now I remember why I put that into the index lol

Libraries/LibGUI/FilteringProxyModel.cpp
alimpfard: Yeah, hold on to that thought, messing with `Model::data` is gonna cost us a few hundred years of copy-pastes

Applications/Help/ManualModel.cpp
```diff
@@ -133,10 +136,42 @@ int ManualModel::column_count(const GUI::ModelIndex&) const
...
+    if (file_or_error.is_error())
+        return false;
+
+    auto file = file_or_error.value();
+
+    for (;;) {
+        auto buffer = file->read_line(4096);
+        if (buffer.is_null())
+            return false;
+        if (String { (const char*)buffer.data() }.contains(term))
```
bugaevc, line 151: This is wrong for two reasons:
* You can't assume that the file doesn't have 0 bytes, nor that `read_line()` will return a 0-terminated buffer (in fact one of the paths there does 0-terminate the buffer, and the other doesn't, and that is wrong too...)
* If the line is long and doesn't fit into the provided buffer size (interesting to see how robust `Core::IODevice::read_line()` is for that case...), you'd miss the occurrence at the edge of the reads. Example: searching for "needle", first read returns "... ne", second read returns "edle ..."

Libraries/LibGUI/ModelIndex.h
bugaevc: Changing the signature of `data()` sounds way more sensible than putting this into the index.

Libraries/LibGUI/FilteringProxyModel.cpp
bugaevc: No, it doesn't matter what feature you're implementing, the Model shouldn't know about the View (the widgetry). The View and the Controller, on the other hand, know about the Model, and can tell it things (call methods on it), like "the filter term has been updated, update yourself".

Libraries/LibGUI/FilteringProxyModel.cpp
bugaevc: So this role specific data is basically an additional argument to `data()`, but done this way to avoid changing the million declarations?

Libraries/LibGUI/ModelIndex.h
awesomekling: I don't understand why an additional argument is necessary in the first place.
Why does the underlying target model of a filter model need to know anything about the filtering state?

Applications/Help/ManualModel.cpp
alimpfard: > You can't assume that the file doesn't have 0 bytes in it, nor that read_line() will return a 0-terminated buffer

Perfect, an excuse to add `memmem` to `AK::ByteBuffer`.
that class is my favourite buffer type :P 

> If the line is long and doesn't fit into the provided buffer size

That's fair. I'd really rather just map the file and do a single pass over it, but I'm not very familiar with `MappedFile` and whatever interfaces it provides for this use, if any.

Libraries/LibGUI/FilteringProxyModel.cpp
alimpfard: Yep.
Now changing the million declarations.

Libraries/LibGUI/ModelIndex.h
bugaevc: Yeah, that, too. The model being filtered shouldn't know it's being filtered, it just provides the data.

alimpfard: We have two options here
- ask the model "does the stuff at this index match this search term?"
- ask the model for the data and do a specific kind of test (e.g. `String.contains`)

I personally think the first makes much more sense.

Applications/Help/ManualModel.cpp
bugaevc: Yeah, using mmap for this sounds much better. `MappedFile` gives you data + size, from which you can construct a string view, and go from there. Bonus point: no copying, other than the initial paging in from the inode.

Libraries/LibGUI/ModelIndex.h
bugaevc: But doesn't that mean each and every model will have to implement searching?

Perhaps we need to scale the ambition back, and only implement this for the man pages model. It lets you get the file path, and then the outer thing could search in that. We could think of a generic search+filter functionality for models later.

awesomekling: > But doesn't that mean each and every model will have to implement searching?

Yes, and that's clearly wrong. A filtering-proxy-model should be just that: a proxy. Something that sits in between a model and a view, filtering the target model, and requiring no additional capabilities from the target model.

alimpfard: I really don't see the problem here, we can just fall back to checking the `Role::Display` string if the model doesn't implement searching (it's not really searching, it's a "does this _index_ match this term").
The only real cost is adding an argument to `data()` that will mostly do nothing at the moment.

awesomekling: If you want an escape hatch to do optimized searches without going through generic data lookup, then we can add *that* instead of messing with the simplicity of `data()` :)

A model could have something like a `virtual TriState data_matches(ModelIndex, Variant) const;` that returns True/False if it is/isn't a match, and Unknown if it's not supported.

alimpfard: _that_ works for me.

alimpfard:
ping -afq @bugaevc

bugaevc:
This looks much better! The filtering model of course shouldn't flatten, but we can improve on that later (preferably, when we fix up the model API).

alimpfard:
Heh. flooding people really works :P 

> The filtering model of course shouldn't flatten, but we can improve on that later (preferably, when we fix up the model API).
Yeah, agreed on that.

Libraries/LibGUI/Variant.h
```diff
@@ -281,6 +297,7 @@ class Variant {
         RawPoint as_point;
         RawSize as_size;
         RawRect as_rect;
+        RawStringView as_string_view;
```
awesomekling, line 300: Let's not do this. All other forms of `GUI::Variant` own their contained value. This suddenly introduces a subtype of `Variant` whose internal value has completely external ownership. That's a huge degradation in API quality.

AK/StringView.cpp
```diff
@@ -178,6 +178,26 @@ bool StringView::contains(char needle) const
...
+        return true;
+
+    if (needle.length() == 1)
+        return contains(needle[0]);
+
+    auto length_diff = length() - needle.length();
+    for (size_t i = 0; i < length_diff; ++i) {
+        if (memcmp(m_characters + i, needle.m_characters, needle.length()) == 0)
+            return true;
+    }
```
awesomekling, line 196: Perhaps we could use `memmem()` here instead. Not that we have `memmem()` yet but it would match `String::contains()` a bit better :)

Libraries/LibGUI/FilteringProxyModel.h
```diff
@@ -0,0 +1,76 @@
...
+    virtual ~FilteringProxyModel() override {};
+
+    virtual int row_count(const ModelIndex& = ModelIndex()) const override;
+    virtual int column_count(const ModelIndex& = ModelIndex()) const override;
+    virtual Variant data(const ModelIndex&, Role = Role::Display) const override;
+    virtual void update() override;
+    virtual ModelIndex index(int row, int column = 0, const ModelIndex& parent = ModelIndex()) const override;
+
+    void start_filtering() { m_is_filtering = true; }
+    void stop_filtering();
```
awesomekling, line 53: These API's should not be necessary. You set a filter string, it filters. You clear the filter term, it no longer filters.

Applications/Help/main.cpp
```diff
@@ -283,6 +291,18 @@ int main(int argc, char* argv[])
...
 
+    if (term_to_search_for_at_launch) {
+        left_tab_bar.set_active_widget(&search_view);
+        search_box.set_text(term_to_search_for_at_launch);
+        if (auto model = search_list_view.model()) {
+            auto& search_model = *static_cast<GUI::FilteringProxyModel*>(model);
+            search_model.start_filtering();
+            search_model.set_filter_term(search_box.text());
+            search_model.update();
+            search_model.stop_filtering();
```
awesomekling, line 302: This should just be a single call to `set_filter_term(search_box.text())`

Libraries/LibGUI/FilteringProxyModel.h
```diff
@@ -0,0 +1,76 @@
...
+    void set_filter_term(const String& term) { m_filter_term = term; }
+
+    bool is_filtering() const { return m_is_filtering; }
+
+    ModelIndex map(const ModelIndex&) const;
+
+private:
+    void filter();
+    FilteringProxyModel(Model& model)
```
awesomekling, line 62: Please mark single-argument constructors `explicit`

Libraries/LibGUI/FilteringProxyModel.h
```diff
@@ -0,0 +1,76 @@
...
+        : m_model(model)
+    {
+    }
+
+    Model& m_model;
+
+    // Maps row to actual model index.
+    Vector<ModelIndex> m_matching_indices;
+
+    bool m_is_filtering { false };
```
awesomekling, line 72: There should be no need for this member; we can just use `m_filter_term`'s null state.

AK/MappedFile.h
```diff
@@ -33,15 +33,17 @@ namespace AK {
...
-    MappedFile() {}
+    MappedFile() { }
     explicit MappedFile(const StringView& file_name);
     MappedFile(MappedFile&&);
     ~MappedFile();
 
     MappedFile& operator=(MappedFile&&);
 
     bool is_valid() const { return m_map != (void*)-1; }
+    int errno_if_invalid() const { return m_errno; }
```
awesomekling, line 46: This could also `ASSERT(!is_valid());`

Libraries/LibGUI/Variant.h
alimpfard: Hmmm...that would mean copying the string if some model does not implement the `data_matches()` API.
I guess if you're fine with it, I'll let it be.

Libraries/LibGUI/FilteringProxyModel.h
alimpfard: This is more about keeping a filtered view as it is, but I can't quite remember why that's needed. I'll remove these.

AK/StringView.cpp
alimpfard: so..."implement memmem" is what you're saying?

awesomekling: Eh, I was just thinking out loud. But sure, it would be nice to have `memmem`!

Kernel/StdLib.cpp
```diff
@@ -91,6 +119,31 @@ void* memmove(void* dest, const void* src, size_t n)
...
+const void* memmem(const void* haystack, size_t haystack_length, const void* needle, size_t needle_length)
+{
+    if (needle_length == 0)
+        return haystack;
+
+    if (haystack_length < needle_length)
+        return nullptr;
+
+    if (haystack_length == needle_length)
+        return memcmp(haystack, needle, haystack_length) ? haystack : nullptr;
```
alimpfard, line 131: I hate that `*cmp` return 0 on equal
that's annoying.

PR #2697, opened 2020-07-04
Kernel/LibCore: Time without syscalls

tomuta:
This provides a way for the Kernel to share data with usermode that requires no system calls. Some OSs like Windows use this technique for certain data. Time is a good example.

### DIFF
--- a/Kernel/Arch/i386/CPU.cpp
+++ b/Kernel/Arch/i386/CPU.cpp
@@ -31,6 +31,7 @@
 #include <Kernel/Arch/i386/CPU.h>
 #include <Kernel/Arch/i386/ProcessorInfo.h>
 #include <Kernel/Arch/i386/ISRStubs.h>
+#include <Kernel/GlobalPage.h>
 #include <Kernel/Interrupts/APIC.h>
 #include <Kernel/Interrupts/GenericInterruptHandler.h>
 #include <Kernel/Interrupts/IRQHandler.h>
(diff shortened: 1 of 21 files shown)


bugaevc:
This was last removed in https://github.com/SerenityOS/serenity/commit/3a92d0828de6ffc040c6006a01a8710dd8a28bf7; you'd have to explain why it makes sense to add it back now :)

tomuta:
Hm I had no idea we something like this before. Well, this also does what `clock_gettime` does without any system calls. Anyway, what was the reason for removing it? It just seems like a better way to measure time. Some benchmark I wrote shows this mechanism to be able to do calls in the tens of millions per second vs in the tens of thousands per second with `clock_gettime` using a system call, so about at least a 1000x improvement.

Anyway, it was just fun to work on, not hard set on absolutely needing it.

awesomekling:
> Hm I had no idea we something like this before. Well, this also does what `clock_gettime` does without any system calls. Anyway, what was the reason for removing it? It just seems like a better way to measure time. Some benchmark I wrote shows this mechanism to be able to do calls in the tens of millions per second vs in the tens of thousands per second with `clock_gettime` using a system call, so about at least a 1000x improvement.
> 
> Anyway, it was just fun to work on, not hard set on absolutely needing it.

Sure, I think we should have the optimization. But we should make all the LibC time functions use it as well :)

tomuta:
I actually initially had LibC do that but then I thought that maybe for regular LibC stuff it should continue to use the syscall as there is some behavior that's not 100% duplicated, e.g. if you pass an invalid pointer it would probably crash rather than setting errno and returning with -1, etc. So, I think it would be better to mainly use the fast implementation on Serenity specific code. Also, it would no longer show up in strace. That's how I initially benchmarked it, but then I pulled the LibC changes for those reasons.

awesomekling:
So I think we should do this, but let's use the old approach of just making a regular mapping inside the process.
We can pass the userspace address of the time page via an ELF auxiliary vector at startup.

tomuta:
I'm going to have to take a peek at what the ELF auxiliary vector is, unless you want to take the bits and pieces from this and do it ;-)

awesomekling:
Closing this PR as "timed out" due to abandonment. Feel free to reopen if work continues. :)

PR #2699, opened 2020-07-04
Kernel: Give Userspace the ELF Auxiliary Vector

ADKaster:
... that we know it's always wanted.

The auxiliary vector entries are placed just after the envp terminating nullptr, to make it easier to find (by iterating one past the end of the envp terminator).

This helps with #1747 by making a bunch of junk that the Loader will need available on the stack.

Also opens the door for putting extra random junk that userspace might think is cool onto its stack in a standard way (by adding custom AT_* values).

Capture

### DIFF
--- a/Kernel/Process.cpp
+++ b/Kernel/Process.cpp
@@ -33,6 +33,7 @@
 #include <AK/Time.h>
 #include <AK/Types.h>
 #include <Kernel/ACPI/Parser.h>
+#include <Kernel/API/Syscall.h>
 #include <Kernel/Arch/i386/CPU.h>
 #include <Kernel/Console.h>
 #include <Kernel/Devices/BlockDevice.h>
@@ -67,7 +68,6 @@
 #include <Kernel/Scheduler.h>
 #include <Kernel/SharedBuffer.h>
 #include <Kernel/StdLib.h>
-#include <Kernel/API/Syscall.h>
 #include <Kernel/TTY/MasterPTY.h>
 #include <Kernel/TTY/TTY.h>
 #include <Kernel/Thread.h>
@@ -888,7 +888,7 @@ int Process::do_exec(NonnullRefPtr<FileDescription> main_program_description, Ve
     //      It also happens to be the static Virtual Addresss offset every static exectuable gets :)
     //      Without this, some assumptions by the ELF loading hooks below are severely broken.
     //      0x08000000 is a verified random number chosen by random dice roll https://xkcd.com/221/
-    u32 totally_random_offset = interpreter_description ? 0x08000000 : 0;
+    m_load_offset = interpreter_description ? 0x08000000 : 0;
 
     // FIXME: We should be able to load both the PT_INTERP interpreter and the main program... once the RTLD is smart enough
     if (interpreter_description) {
@@ -906,7 +906,7 @@ int Process::do_exec(NonnullRefPtr<FileDescription> main_program_description, Ve
     Region* master_tls_region { nullptr };
     size_t master_tls_size = 0;
     size_t master_tls_alignment = 0;
-    u32 entry_eip = 0;
+    m_entry_eip = 0;
 
     MM.enter_process_paging_scope(*this);
     RefPtr<ELF::Loader> loader;
@@ -935,7 +935,7 @@ int Process::do_exec(NonnullRefPtr<FileDescription> main_program_description, Ve
                 prot |= PROT_WRITE;
             if (is_executable)
                 prot |= PROT_EXEC;
-            if (auto* region = allocate_region_with_vmobject(vaddr.offset(totally_random_offset), size, *vmobject, offset_in_image, String(name), prot)) {
+            if (auto* region = allocate_region_with_vmobject(vaddr.offset(m_load_offset), size, *vmobject, offset_in_image, String(name), prot)) {
                 region->set_shared(true);
                 return region->vaddr().as_ptr();
             }
@@ -949,7 +949,7 @@ int Process::do_exec(NonnullRefPtr<FileDescription> main_program_description, Ve
                 prot |= PROT_READ;
             if (is_writable)
                 prot |= PROT_WRITE;
-            if (auto* region = allocate_region(vaddr.offset(totally_random_offset), size, String(name), prot))
+            if (auto* region = allocate_region(vaddr.offset(m_load_offset), size, String(name), prot))
                 return region->vaddr().as_ptr();
             return nullptr;
         };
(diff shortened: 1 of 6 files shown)


Kernel/Process.h
```diff
@@ -661,8 +665,7 @@ inline void Process::for_each_thread(Callback callback) const
     if (my_pid == 0) {
         // NOTE: Special case the colonel process, since its main thread is not in the global thread table.
         Processor::for_each(
-            [&](Processor& proc) -> IterationDecision
-            {
+            [&](Processor& proc) -> IterationDecision {
```
ADKaster, line 668: Should I revert these odd clang-format changes? :/ I'm using clang-format-9...

Kernel/Process.cpp
```diff
@@ -1086,6 +1089,44 @@ int Process::do_exec(NonnullRefPtr<FileDescription> main_program_description, Ve
...
+    auxv.append({ AuxiliaryValue::Gid, (long)m_gid });
+    auxv.append({ AuxiliaryValue::EGid, (long)m_egid });
+
+    // FIXME: Don't hard code this? We might support other platforms later.. (e.g. x86_64)
+    auxv.append({ AuxiliaryValue::Platform, "i386" });
+    // FIXME: This is platform specific
+    auxv.append({ AuxiliaryValue::HwCap, (long)CPUID(1).edx() });
+
+    // FIXME: Is this 100HZ or 1000HZ?
+    auxv.append({ AuxiliaryValue::ClockTick, 100 });
```
awesomekling, line 1113: The correct value here is `TimeManagement::the().ticks_per_second()` :)

Kernel/Process.cpp
```diff
@@ -1086,6 +1089,44 @@ int Process::do_exec(NonnullRefPtr<FileDescription> main_program_description, Ve
...
+
+    // FIXME: Also take into account things like extended filesystem permissions? That's what linux does...
+    auxv.append({ AuxiliaryValue::Secure, ((m_uid != m_euid) || (m_gid != m_egid)) ? 1 : 0 });
+
+    // +1 for null term, we're going to be lazy and store this as a string
+    char random_bytes[17] {};
+    get_good_random_bytes((u8*)random_bytes, 16); // FIXME: Are these "secure random bytes?"
```
awesomekling, line 1120: The `get_good_random_bytes` API is indeed intended to be our most secure random bytes.
So while we might want to improve our randomness, this line of code will not need to change. Hence we don't need this FIXME :)

Kernel/Process.cpp
```diff
@@ -1086,6 +1089,44 @@ int Process::do_exec(NonnullRefPtr<FileDescription> main_program_description, Ve
...
+
+    auxv.append({ AuxiliaryValue::Random, String(random_bytes) });
```
awesomekling, line 1122: The `String(const char*)` constructor will stop fetching bytes if it encounters a null terminator.
So if you get a zero early in the string of random bytes, we'll have a much shorter sequence as a result.

Kernel/Process.h
```diff
@@ -510,6 +512,8 @@ class Process : public InlineLinkedListNode<Process> {
     gid_t m_sgid { 0 };
 
     pid_t m_exec_tid { 0 };
+    FlatPtr m_load_offset { 0U };
+    FlatPtr m_entry_eip = { 0U };
```
awesomekling, line 516: Let's not have both `=` and `{ }` :)

Kernel/Process.h
awesomekling: Yes, please use clang-format 10 :)

Kernel/Process.h
```diff
@@ -475,6 +476,7 @@ class Process : public InlineLinkedListNode<Process> {
     ssize_t do_write(FileDescription&, const u8*, int data_size);
 
     KResultOr<NonnullRefPtr<FileDescription>> find_elf_interpreter_for_executable(const String& path, char (&first_page)[PAGE_SIZE], int nread, size_t file_size);
+    Vector<AuxiliaryValue> generate_auxiliary_vector();
```
awesomekling, line 479: Can this be `const`?

Kernel/Thread.h
```diff
@@ -381,7 +382,7 @@ class Thread {
     void set_default_signal_dispositions();
     void push_value_on_stack(FlatPtr);
 
-    u32 make_userspace_stack_for_main_thread(Vector<String> arguments, Vector<String> environment);
+    u32 make_userspace_stack_for_main_thread(Vector<String> arguments, Vector<String> environment, Vector<AuxiliaryValue> auxv);
```
awesomekling, line 385: I would omit the last parameter name here since it's obvious from the type.

Libraries/LibELF/AuxiliaryVector.h
```diff
@@ -0,0 +1,116 @@
+/*
+ * Copyright (c) 2020, The SerenityOS developers.
```
awesomekling, line 2: I notice you are using the generic copyright here while you have a personal one in other parts of the project.
Either way is fine of course, just pointing out the deviation :)

Kernel/Process.h
ADKaster: Welp, everything left in the clang-format department is left from using clang-format-10

Kernel/Process.cpp
ADKaster: Aha, was looking for that but didn't know where. Changed

Kernel/Process.cpp
ADKaster: Fair enough, fixme removed.

Kernel/Process.cpp
ADKaster: Oh whoops. Was expecting it to select a `template <size_t> String(const char (&array)[N]) ` sort of constructor instead. Changed to char* + length constructor

Kernel/Process.h
ADKaster: True. no =

Kernel/Process.h
ADKaster: Sure can! constified.

Kernel/Thread.h
ADKaster: True, originally I had a `Vector<auxv_t>` which wasn't quite as obvious.

Libraries/LibELF/AuxiliaryVector.h
ADKaster: Yeah, was intentional here.

awesomekling:
Whoops, this has some conflicts. Sorry about that :| Patches look good otherwise!

PR #2700, opened 2020-07-05
Add manpages for posix_spawn

nico:


nico:
@bugaevc

nico:
(Putting them in man2 instead of man3 is admittedly a bit aspirational.)

Base/usr/share/man/man2/posix_spawn.md
```diff
@@ -0,0 +1,47 @@
...
+
+int posix_spawn(pid_t* pid, const char* file, const posix_spawn_file_actions_t*, const posix_spawnattr_t*, char* const argv[], char* const envp[]);
+int posix_spawnp(pid_t* pid, const char* file, const posix_spawn_file_actions_t*, const posix_spawnattr_t*, char* const argv[], char* const envp[]);
+```
+
+## Description
+
+Spawns a new process reading the binary `file`, passing `argv` as arguments to `main()` and setting `envp` as argument.
+
+Places the pid of the new process in `pid`.
```
bugaevc, line 18: "PID", or "process ID".

Base/usr/share/man/man2/posix_spawn.md
```diff
@@ -0,0 +1,47 @@
...
+
+If `file` passed to `posix_spawn` is a relative path, it is resolved relative to the current working directory.
+
+If `file` passed to `posix_spawnp` is a relative path, it is resolved by searching `PATH`.
```
bugaevc, line 22: Uh, perhaps we should name the argument something other than `file` then? `executable_path`, for example.

"By searching through directories specified in the `PATH` environment variable."

Base/usr/share/man/man2/posix_spawn.md
```diff
@@ -0,0 +1,47 @@
...
+
+The `posix_spawn_file_actions_t` and `posix_spawnattr_t` arguments may be `nullptr`. If they aren't, see [`posix_spawn_file_actions`(2)](posix_spawn_file_actions_init.md) and [`posix_spawnattr`(2)](posix_spawnattr_init.md) for what they do.
+
+The last entry in `argv` and `envp` has to be `nullptr`.
+
+To make the subprocess use the parent's environment, declare `extern char* environ;` in your program and pass that.
```
bugaevc, line 28: It's not "subprocess", it's "child process".

This remark belongs in "notes" or something, not in "description".

You don't have to declare `environ`, it's in unistd:

https://github.com/SerenityOS/serenity/blob/6d5bd8c76b1eec679ec6ca8201eef6ec5ff21a9e/Libraries/LibC/unistd.h#L55

And it's a `char**`.

Base/usr/share/man/man2/posix_spawn.md
```diff
@@ -0,0 +1,47 @@
...
+
+The new process is started as if the following steps are executed in this order:
+
+1. A new process is started as if `fork()` was called.
+2. If the `posix_spawnattr_t` parameter is non-NULL, its [effects](posix_spawnattr_init.md) are processed.
```
bugaevc, line 33: `NULL` -> `nullptr`

bugaevc: `#include <english-ain't-my-first-language.h>`

its effects are... applied? executed? performed?

Base/usr/share/man/man2/posix_spawn.md
```diff
@@ -0,0 +1,47 @@
...
+3. If the `posix_spawn_file_actions_t` parameter is non-NULL, its [effects](posix_spawn_file_actions_init.md) are processed.
+4. `file` is started as if `execve` or `execvpe` was called.
```
bugaevc, line 35: We've already "started" the process.

We could say that "the new process begins executing the `file`, as if after an `execve()` call", or something like that.

Base/usr/share/man/man2/posix_spawn.md
```diff
@@ -0,0 +1,47 @@
...
+
+## Return value
+
+If the process is successfully forked, returns 0.
+Else, returns an error number. This function does *not* return -1 on error and sets `errno`, it instead returns what other functions set `errno` to as result.
```
bugaevc, line 40: Super super nitpick: this sentence should use the word "otherwise" instead of "else". Just for consistency with how this section is written in other pages.

Also: "if the process is successfully spawned, *this function*" (or, `posix_spawn()`) "returns 0."

Also: "does not return -1 on error and **set** `errno`"

Base/usr/share/man/man2/posix_spawn_fle_actions_init.md
```diff
@@ -0,0 +1,50 @@
+## Name
+
+`posix_spawn` - launch a new process
```
bugaevc, line 3: But this is a man page for `posix_spawn_file_actions_t`, not for `posix_spawn`.

Which, while we're at it, should all be on one page, ideally.

Base/usr/share/man/man2/posix_spawnattr_init.md
```diff
@@ -0,0 +1,79 @@
...
+int posix_spawnattr_getschedpolicy(const posix_spawnattr_t*, int*);
+int posix_spawnattr_getsigdefault(const posix_spawnattr_t*, sigset_t*);
+int posix_spawnattr_getsigmask(const posix_spawnattr_t*, sigset_t*);
+int posix_spawnattr_setflags(posix_spawnattr_t*, short);
+int posix_spawnattr_setpgroup(posix_spawnattr_t*, pid_t);
+int posix_spawnattr_setschedparam(posix_spawnattr_t*, const struct sched_param*);
+int posix_spawnattr_setschedpolicy(posix_spawnattr_t*, int);
+int posix_spawnattr_setsigdefault(posix_spawnattr_t*, const sigset_t*);
+int posix_spawnattr_setsigmask(posix_spawnattr_t*, const sigset_t*);
+
```
bugaevc, line 35: Extra blank line.

Base/usr/share/man/man2/posix_spawn.md
alimpfard: Nitpick alongside your nitpick:
"does not <do X> and <do Y>" is super confusing when someone skims through, perhaps repeat the 'does not', or just omit that entirely and say "this function returns the would-be value of `errno`, and does Z to `errno`"
(what does it do to `errno`?)

Base/usr/share/man/man2/posix_spawn.md
```diff
@@ -0,0 +1,47 @@
+## Name
+
+`posix_spawn` - launch a new process
```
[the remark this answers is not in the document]
nico: It's markdown, so functions that contain a `_` must have backticks, since markdown treats `_` even in the middle of a word as start of italics. Functions that don't contain underscores don't need this.

Base/usr/share/man/man2/posix_spawn.md
nico: I was led astray by https://man7.org/linux/man-pages/man7/environ.7.html which says " (This variable must be declared in the user program, but is declared in the header file <unistd.h> if the _GNU_SOURCE feature test macro is defined.)".

https://pubs.opengroup.org/onlinepubs/9699919799/functions/environ.html doesn't have any of that language, so thanks :)

Base/usr/share/man/man2/posix_spawn.md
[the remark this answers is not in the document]
bugaevc: No; <code>``</code> in Markdown emits the <code>&lt;code&gt;</code> HTML tags, and disables underscore (and star, and other special characters) handling as a side effect. The proper way to deal with underscores inside words is to escape them, like this: `posix\_spawn`

Base/usr/share/man/man2/posix_spawn_fle_actions_init.md
nico: No, most calls to `posix_spawn` pass nullptr for both file actions and attribs. It makes sense to keep the main manpage focused for the common case.

Base/usr/share/man/man2/posix_spawn.md
bugaevc: Heh, a bit of a friendly advice: do read Linux and POSIX man pages — they can be helpful at times, — but do not mistake then for the "real thing" (SerenityOS). SerenityOS ≠ Linux && SerenityOS ≠ POSIX.

Base/usr/share/man/man2/posix_spawn_fle_actions_init.md
bugaevc: I don't think there's any problem with adding more information to the `posix_spawn` page — perhaps as separate sections of you don't want to clutter "description". (Also adding a bunch of examples, both simple and complex, wouldn't hurt 😉)

But I do see an issue with creating a bazillion pages (especially if you count all those symlinks) to describe various helpers of `posix_spawn`.

Base/usr/share/man/man2/posix_spawn.md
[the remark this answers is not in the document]
nico: Done.

Base/usr/share/man/man2/posix_spawn_fle_actions_init.md
nico: I added a short example inline. I want to add a longer example in the directory with the example programs (and the mention it here once it exists).

nico:
Forgot to reply to the Help app question. Currently it doesn't seem to do anything with symlinks, which isn't super surprising given the setuid man pages added the first symlink :)

I'm not sure just omitting symlinks is the best thing to do (...but once Help has search it probably doesn't matter too much either way). Maybe symlinked pages could instead be in a subcategory below their link target?

alimpfard:
why care at all about whether a page is a symlink to another in Help?
aside from an optimisation on mapping (we could maybe just refer to the mapped object instead of remapping it), it doesn't really matter if two pages are really just one page...right?

bugaevc:
I'm not concerned about memory usage; but imagine opening Help and seeing a bajillion pages about `posix_spawn` when in reality there are just three and all the rest are symlinks to them.

Symlinks make sense for enabling you to type `man thing` and have it open the appropriate manpage, even if it's named differently. But that's not how you use Help, you browse, or (soon) search through, the available help pages.

alimpfard:
Well, the way I see it is that symlinked pages really refer to different parts of the same page, so...perhaps we should have some form of anchoring instead? that's probably too ambitious though.

bugaevc:
I don't think it's about anchoring, it's just about several things being described on the same page. The sections are "name, synopsis, description, return value", etc — and every section applies to all the functions/programs described in the page.

nico:
@bugaevc Happy to make one page per function if that's what you want, but the other day you seemed to ask for the opposite of that.

I think it makes sense to have every function with documentation available for both `man myfun` and as an entry in Help, since you want to see it and click on it. (As said upreview, less important once the PR that adds search to Help.)

bugaevc:
> Happy to make one page per function if that's what you want, but the other day you seemed to ask for the opposite of that.

No no no, I'm not suggesting one page per function 😮😀

We could nest symlinks under their parent page, I guess 🤔

nico:
Ok, sounds like this is fine as-is for now and we can change the Help UI to do…something?…with symlinks in some way in a follow-up (?).

awesomekling:
> Ok, sounds like this is fine as-is for now and we can change the Help UI to do…something?…with symlinks in some way in a follow-up (?).

I agree (?)

PR #2701, opened 2020-07-05
WebView: Use available size for bitmap

kevmeyer:
Before:
Screenshot 2020-07-05 at 12 32 29

After:
Screenshot 2020-07-05 at 12 33 32


awesomekling:
Excellent patch :D

Issue #2693, opened 2020-07-04
Kernel: Mysterious process hangs after SMP changes


awesomekling:
cc @tomuta 

I keep seeing random process hangs after the SMP changes.
One slightly unreliable way to reproduce:

1. (Re)boot the system.
2. Run `ls | ls` in the terminal.
3. If control returns to the shell, goto 1. Otherwise you have reproduced the bug.

tomuta:
Do you have any stack traces? I can't seem to be able to reproduce it :(

tomuta:
Oh just as I wrote it I managed to reproduce one using Browser. Let me see what I can figure out.

awesomekling:
Hunch: Perhaps we are doing a context switch at a bad time after `exec()` and we're missing some logic to suppress context switches there.

tomuta:
Hm, not sure if this is really what's going on. While I can see a context switch happen during a fork sometimes, it doesn't look like it clobbers registers as they match when it gets scheduled again, which makes sense because technically it may clobber them but it left off at the same spot anyway when switching out. It seems like what sometimes happens is that threads get stuck in `Queued` state somehow and never seem to get out of it.

tomuta:
I enhanced the `Thread::wait_on` to store the wait reason, and changed `Scheduler`'s `SCHEDULER_RUNNABLE_DEBUG` logging a bit, when this happens it seems like the processes get stuck in `Queued` state due to `Inode`. Not sure how or why, but this is what's getting logged while in this condition:

```
[colonel(0:0)]: Scheduler[0]: Switch to colonel(0:0) @ 0008:c011eff0
[colonel(0:0)]: Non-runnables:
[colonel(0:0)]:   Queued       (futex) AudioServer(14:21) @ 0008:c011eff0
[colonel(0:0)]:   Selecting    ClipboardHistory.MenuApplet(10:10) @ 0008:c011eff0
[colonel(0:0)]:   Selecting    Audio.MenuApplet(11:11) @ 0008:c011eff0
[colonel(0:0)]:   Selecting    UserName.MenuApplet(18:18) @ 0008:c011eff0
[colonel(0:0)]:   Selecting    DHCPClient(17:17) @ 0008:c011eff0
[colonel(0:0)]:   Selecting    AudioServer(14:14) @ 0008:c011eff0
[colonel(0:0)]:   Selecting    SystemMenu(16:16) @ 0008:c011eff0
[...]

PR #2706, opened 2020-07-05
Kernel: Fix a number of problems with the software context switching

tomuta:
I believe this should fix the hanging processes issue as reported in #2693

### DIFF
--- a/Kernel/Heap/SlabAllocator.cpp
+++ b/Kernel/Heap/SlabAllocator.cpp
@@ -114,6 +114,7 @@ private:
 static SlabAllocator<16> s_slab_allocator_16;
 static SlabAllocator<32> s_slab_allocator_32;
 static SlabAllocator<64> s_slab_allocator_64;
+static SlabAllocator<128> s_slab_allocator_128;
 
 static_assert(sizeof(Region) <= s_slab_allocator_64.slab_size());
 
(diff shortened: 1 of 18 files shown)


Kernel/Heap/SlabAllocator.cpp
```diff
@@ -130,6 +131,7 @@ void slab_alloc_init()
     s_slab_allocator_16.init(128 * KB);
     s_slab_allocator_32.init(128 * KB);
     s_slab_allocator_64.init(512 * KB);
+    s_slab_allocator_64.init(512 * KB);
```
awesomekling, line 134: Not initializing the right bucket here ;)

tomuta: Whoops, not sure what bad things this caused...

tomuta:
Ok cleaned up a few more cosmetic things and log statements, I think this is ready to be merged

nico:
I patched this in and things seem to work again on my system with this applied.

tomuta:
It still may hang early in the boot process every now and then with -smp 2 or more, I have another branch that I started working on before this issue that I think will help address this (basically inter-cpu messaging/interrupts for MM etc). It'll be a challenge rebasing that on top of these changes, but maybe for now we should change run.sh to use -smp 1 by default

awesomekling:
Stellar, thank you. :)

PR #2707, opened 2020-07-06
Kernel: Add SMP ICI support

tomuta:
We can now properly initialize all processors without
crashing by sending SMP IPI messages to synchronize memory
between processors.

We now initialize the APs once we have the scheduler running.
This is so that we can process IPI messages from the other
cores.

Also rework interrupt handling a bit so that it's more of a
1:1 mapping. We need to allocate non-sharable interrupts for
IPIs.

This also fixes the occasional hang/crash because all
CPUs now synchronize memory with each other.


tomuta:
What's interesting is that with these changes the `ls | ls` test causes Shell to hang more likely the more cores there are. But the same can be observed with `-smp 1` when the system is busy (e.g. Browser in the background hogging the cpu). It seems like all other applications, including the task bar, work fine, so I think this is some odd issue in Shell.

Also, bootup with more cores is noticably slower. That's because `Region::map` is spamming the system with a ton of "flush tlb" messages right now, it sends one for each page mapped. That code should be changed to map and then call `Processor::flush_tlb` with the appropriate number of pages to flush, that way only one ICI is sent for the entire region. There are probably more optimizations needed elsewhere but that one likely has the most impact right now.

awesomekling:
Isn't this called IPI? I've never heard the term ICI.

Kernel/Arch/i386/CPU.h
```diff
@@ -624,10 +618,53 @@ struct TrapFrame;
...
+    ProcessorMessage* msg;
+};
 
 class Processor {
     friend class ProcessorInfo;
 
+    Processor(const Processor&) = delete;
+    Processor(Processor&&) = delete;
+    Processor& operator=(const Processor&) = delete;
+    Processor& operator=(Processor&&) = delete;
```
awesomekling, line 666:
```suggestion
    AK_MAKE_NONCOPYABLE(Processor);
    AK_MAKE_NONMOVABLE(Processor);
```

Kernel/Arch/i386/CPU.h
```diff
@@ -641,29 +678,67 @@ class Processor {
...
+    static void smp_broadcast_message(ProcessorMessage& msg, bool async);
+    static void smp_broadcast_halt();
 
     void cpu_detect();
     void cpu_setup();
 
     String features_string() const;
 
 public:
+    Processor() noexcept
```
awesomekling, line 712: We compile without exception support so `noexcept` adds nothing.

Kernel/Arch/i386/CPU.h
```diff
@@ -674,8 +749,9 @@ class Processor {
     static inline IterationDecision for_each(Callback callback)
     {
         auto& procs = processors();
-        for (auto it = procs.begin(); it != procs.end(); ++it) {
-            if (callback(**it) == IterationDecision::Break)
+        size_t cnt = procs.size();
```
awesomekling, line 752: Please write whole words; `cnt` -> `count`.

Kernel/Arch/i386/CPU.h
```diff
@@ -784,6 +871,29 @@ class Processor {
...
+    static void smp_enable();
+    bool smp_process_pending_messages();
+
+    template <typename Callback>
+    static void smp_broadcast(Callback callback, bool async)
+    {
+        auto* data = new Callback(move(callback));
+        smp_broadcast(
+            [](void* data)
+            {
```
awesomekling, line 883: Please use `clang-format` version 10 or later.

Kernel/Interrupts/APIC.cpp
```diff
@@ -62,6 +70,67 @@ namespace Kernel {
 
 static APIC *s_apic;
 
+class APICICIInterruptHandler final : public GenericInterruptHandler {
+public:
+    explicit APICICIInterruptHandler(u8 interrupt_vector):
+        GenericInterruptHandler(interrupt_vector, true)
```
awesomekling, line 76: Please use `clang-format`

tomuta:
You're right, I guess I thought of "Inter CPU Interrupt", but yes Intel calls it Interprocessor Interrupts (IPI). Let me rename it.

awesomekling:
Very interesting changes :)
The TLB flushing issue will be good to fix, indeed!

alimpfard:
> I think this is some odd issue in Shell

Is this on top of the recent shell changes?

Kernel/Interrupts/APIC.cpp
tomuta: We need to add a pre-commit hook with clang-format, I keep forgetting...

tomuta:
> > I think this is some odd issue in Shell
> 
> Is this on top of the recent shell changes?

Yes. It can be reproduced without these changes too, but it's more difficult. Shell seems to have some odd bugs while the CPU is under heavy load, e.g. Browser being busy in the background. If you do `ls | ls`, even without these changes, if the system is busy, odds are you're going to run into hangs or even crashes after a couple of attempts.

I can't say for sure it's in fact a bug in Shell, but since most other things appear to be working without issues it makes me suspect it is.

tomuta:
Also just a note, the `Scheduler` won't actually use the other cores yet, as we don't implement APIC timers that fire on them yet. They're just sitting in their idle loop waiting, but they can process IPIs.

alimpfard:
> It can be reproduced without these changes too, but it's more difficult

I expected it to be harder now...oh dear.
I'll investigate that.

PR #2708, opened 2020-07-06
TextEditor: Use ArgsParser and allow the user to specify preview mode

alimpfard:
This makes looking at HTML files that somehow fail a little less painful by allowing the user to actually _see_ the html.

TL;DR, I tried to look at what twitter.com serves and had to wait forever for the preview to load.
sidenote: when can we have DoH? I'm tired of mounting /etc rw and changing the nameserver to my local doh proxy...

### DIFF
--- a/Applications/TextEditor/TextEditorWidget.h
+++ b/Applications/TextEditor/TextEditorWidget.h
@@ -51,6 +51,7 @@ public:
     };
 
     void set_preview_mode(PreviewMode);
+    void set_auto_detect_preview_mode(bool value) { m_auto_detect_preview_mode = value; }
 
 private:
     TextEditorWidget();
@@ -107,6 +108,7 @@ private:
 
     bool m_document_dirty { false };
     bool m_document_opening { false };
+    bool m_auto_detect_preview_mode { false };
 
     PreviewMode m_preview_mode { PreviewMode::None };
 };
--- a/Applications/TextEditor/main.cpp
+++ b/Applications/TextEditor/main.cpp
@@ -25,6 +25,7 @@
  */
 
 #include "TextEditorWidget.h"
+#include <LibCore/ArgsParser.h>
 #include <LibGfx/Bitmap.h>
 #include <stdio.h>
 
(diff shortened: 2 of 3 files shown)


Applications/TextEditor/TextEditorWidget.h
```diff
@@ -46,6 +46,7 @@ class TextEditorWidget final : public GUI::Widget {
 
     enum class PreviewMode {
         None,
+        Default,
```
awesomekling, line 49: `Default` is not a preview mode in itself, and so doesn't belong in this enum :)
Please reorganize the code so this state is not required.

alimpfard: how about "auto"?
that's really what it is.

Applications/TextEditor/main.cpp
```diff
@@ -42,6 +43,16 @@ int main(int argc, char** argv)
         return 1;
     }
 
+    const char* preview_mode = "default";
+    const char* file_to_edit = nullptr;
+    Core::ArgsParser parser;
+    parser.add_option(preview_mode, "Preview mode, one of 'none', 'html', 'markdown', 'default'", "mode", 'm', "mode");
```
bugaevc, line 49: I wouldn't call it `--mode`/`-m`. `--preview-mode` (with no short variant) sound better.

Applications/TextEditor/TextEditorWidget.h
awesomekling: What I mean is that this shouldn't be a new `PreviewMode` value

Applications/TextEditor/main.cpp
alimpfard: why make it so darn hard to set a preview mode?
(no qualms about `--preview-mode` itself)

bugaevc: Why hard? It's just that, in the context of a text editor, "mode" doesn't immediately mean "whether we're previewing Markdown or HTML" to me. "Preview mode" at least says a mode of *what* it is.

alimpfard: nonono, I mean, why disable the short form?
`--preview-mode` makes more sense, that's fine by me.

alimpfard:
Oh ffs, I pushed the dns thing too

Applications/TextEditor/main.cpp
bugaevc: Because, IMO, neither `-p` nor `-m` are good shortenings of `--preview-mode`. `-m` would have been a good shortening of `--mode`. Something weird would work, like `-M`, for a system that doesn't support long options, but we don't have to do that.

That being said, I'm not inherently against it, just saying it doesn't make much sense to me. If you want it, I won't object.

alimpfard: how about an (admittedly unorthodox) alias like "--pm"?
nevermind, terrible idea.

alimpfard: I'll just remove the short version, some day we'll have tab-completions for these anyway.

bugaevc:
> I'm tired of mounting /etc rw and changing the nameserver to my local doh proxy...

You only need to remount `/etc` if you want to change DNS settings *at runtime*. Which likely wouldn't work, since LookupServer may already be running, so you'd have to kill it. Why not just edit the file on the host?

alimpfard:
because I tend to suddenly remember that "oops, I need to do that" while I have a bunch of other things done on the system

PR #2711, opened 2020-07-06
Shell: Handle the case where the child we're waiting for doesn't exist

alimpfard:
Normally, this should not happen, as the child should stay around until we do a waitid on it.
So for the sake of keeping it working, slap a fixme on it and adapt to the problem until it goes away :P

### DIFF
--- a/Shell/main.cpp
+++ b/Shell/main.cpp
@@ -78,6 +78,16 @@ int main(int argc, char** argv)
         for (auto& job : jobs) {
             int wstatus = 0;
             auto child_pid = waitpid(job.value->pid(), &wstatus, WNOHANG);
+            if (child_pid < 0) {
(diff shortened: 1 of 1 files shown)


Shell/main.cpp
```diff
@@ -78,6 +78,16 @@ int main(int argc, char** argv)
         for (auto& job : jobs) {
             int wstatus = 0;
             auto child_pid = waitpid(job.value->pid(), &wstatus, WNOHANG);
+            if (child_pid < 0) {
+                if (errno == ECHILD) {
+                    // The child process went away before we could process its death, just assume it exited all ok.
+                    // FIXME: This should never happen, the child should stay around until we do the waitpid above.
```
awesomekling, line 84: Uhh okay, `FIXME` indeed :)

PR #2712, opened 2020-07-06
Kernel: Minor MM optimization for SMP

tomuta:
MemoryManager::quickmap_pd and MemoryManager::quickmap_pt can only
be called by one processor at the time anyway, since anything using
these must have the MM lock held. So, no need to inform the other
CPUs to flush their TLBs, we can just flush our own.


tomuta:
This helps a little, but the big one is still `Region::map`...

awesomekling:
Yup that makes sense.

PR #2713, opened 2020-07-06
LibLine+Shell: SIGWINCH and stuff

alimpfard:
```sh
> cat
# resize window
^D
> # Shell should not have the wrong idea about the terminal size
```

@awesomekling  :^)

### DIFF
--- a/Libraries/LibLine/Editor.cpp
+++ b/Libraries/LibLine/Editor.cpp
@@ -48,6 +48,18 @@ Editor::Editor(Configuration configuration)
 {
     m_always_refresh = configuration.refresh_behaviour == Configuration::RefreshBehaviour::Eager;
     m_pending_chars = ByteBuffer::create_uninitialized(0);
+    get_terminal_size();
+    m_suggestion_display = make<XtermSuggestionDisplay>(m_num_lines, m_num_columns, m_cached_prompt_metrics);
+}
+
+Editor::~Editor()
+{
+    if (m_initialized)
+        restore();
+}
+
+void Editor::get_terminal_size()
+{
     struct winsize ws;
     if (ioctl(STDOUT_FILENO, TIOCGWINSZ, &ws) < 0) {
         m_num_columns = 80;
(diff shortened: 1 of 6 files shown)


alimpfard:
Oops, wrong commit

Libraries/LibLine/Editor.cpp
```diff
@@ -976,14 +982,7 @@ void Editor::refresh_display()
...
 
-        struct winsize ws;
-        if (ioctl(STDOUT_FILENO, TIOCGWINSZ, &ws) < 0) {
-            m_num_columns = 80;
-            m_num_lines = 25;
-        } else {
-            m_num_columns = ws.ws_col;
-            m_num_lines = ws.ws_row;
-        }
+        get_terminal_size();
```
awesomekling, line 985: Wait, why isn't it enough to do this on `SIGWINCH`?

alimpfard: because we don't get that signal when we're not the owning process.
when running `cat`, the signals are delivered to `cat`, not the shell.
(we already do this on `SIGWINCH`)

awesomekling: Hmm, that doesn't seem right! The kernel should `SIGWINCH` all processes on that TTY, I would think, no?

alimpfard: That sounds logical, yeah.
I'll take a look.

alimpfard: Nope, looks like it sends the stuff to the owning process group
why are we setting that to the child at all?

alimpfard: so here's what I see:
- shell `tcsetpgrp(0, child_pgrp)`'s to avoid getting the interrupts and all that jazz
- that includes `SIGWINCH`
- shell doesn't get notified about the tty size change

if we _do_ include the shell, we have to deal with a bunch of unrelated signals, which I guess we can do...?

alimpfard:
Note that we will still update the display size in `Editor::initialize` if we were signaled when out of get_line, and we didn't update the display (since there's nothing to update).

PR #2714, opened 2020-07-06
Kernel: Aggregate TLB flush requests for Regions for SMP

tomuta:
Rather than sending one TLB flush request for each page,
aggregate them so that we're not spamming the other
processors with FlushTLB IPIs.


awesomekling:
For non-mergeable WIP's, you can use the PR "draft" state :)

tomuta:
Hm, the new code is simply too fast, triggering some other race condition :(

tomuta:
No it's not.... `Processor::flush_tlb_local` has a fatal flaw... ;-) I'll give it another try after lunch :)

tomuta:
That made it work, but it's not as much of a speed improvement as I hoped.

PR #2715, opened 2020-07-06
LibLine: Support Ctrl-b/f and Ctrl-d

nico:
b/f go backward/forward by one character, d deletes one character.

Also do some minor preparatory changes to make it easier to support these shortcuts.

### DIFF
--- a/Libraries/LibLine/Editor.cpp
+++ b/Libraries/LibLine/Editor.cpp
@@ -742,17 +742,22 @@ void Editor::handle_read_event()
             m_refresh_needed = true;
             continue;
         }
+        // ^A
+        if (codepoint == ctrl('A')) {
+            m_cursor = 0;
+            continue;
+        }
+        // ^E
+        if (codepoint == ctrl('E')) {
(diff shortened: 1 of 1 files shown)


nico:
I'm never quite sure if these things should be handled in the terminal emulator or the shell btw. Maybe the Right Fix is to instead have the shell send the corresponding ansi escape code for these keys?

Libraries/LibLine/Editor.cpp
```diff
@@ -422,6 +422,69 @@ void Editor::handle_read_event()
     Utf8View input_view { StringView { m_incomplete_data.data(), valid_bytes } };
     size_t consumed_codepoints = 0;
 
+    enum Amount { kCharacter, kWord };
```
alimpfard, line 425: uhhh...why not just `Character` and `Word`?

nico: Err, sorry. Fixed.

awesomekling:
Quite tasteful!

Issue #2717, opened 2020-07-06
Shell: Frequent crashes


tomuta:
Tying to do `ls | ls` often results in crashes, in particular when the system is busy (e.g. Browser in the background).

This change https://github.com/SerenityOS/serenity/commit/dec27e5e6fad25fedd9203087b8107242369dd2b and a stack trace in this region prompted me to look at the code, and I think the problem is fairly straight forward. We can't do these types of things within a signal handler. There cannot be any assumptions as to the state of the application when it gets called, so it's not safe to e.g. call `editor->resized()`, `editor->save_history()`, or in this case iterate `s_shell->jobs`.

I think what we need to solve this is an enhancement for `Core::EventLoop` where you can subscribe to a signal, in essence an approach similar to `boost::signal_set`, which under the hood creates a pipe where the signal handler merely pushes the signal code into, and then the event loop would get notified and pulls it out, asynchronously executing the signal.

alimpfard:
That has been discussed, I made an attempt in #2440, but it was not a particularly good attempt.
Really, putting anything in the signal handler was not at all something desired.
+1 for the signal subscription thing, so long as I don't have to do it :P

tomuta:
Well, writing to a pipe in a signal handler is safe, much more than that, especially anything involving mutexes, such as `malloc` is not... That's why boost uses a pipe internally.

PR #2718, opened 2020-07-06
Shell: Fix hanging (signal handling fixes)

tomuta:
See #2717

### DIFF
--- a/AK/HashMap.h
+++ b/AK/HashMap.h
@@ -54,13 +54,16 @@ public:
     size_t capacity() const { return m_table.capacity(); }
     void clear() { m_table.clear(); }
 
-    void set(const K& key, const V& value) { m_table.set({ key, value }); }
-    void set(const K& key, V&& value) { m_table.set({ key, move(value) }); }
-    void remove(const K& key)
+    HashSetResult set(const K& key, const V& value) { return m_table.set({ key, value }); }
+    HashSetResult set(const K& key, V&& value) { return m_table.set({ key, move(value) }); }
+    bool remove(const K& key)
     {
         auto it = find(key);
-        if (it != end())
+        if (it != end()) {
             m_table.remove(it);
+            return true;
+        }
+        return false;
     }
     void remove_one_randomly() { m_table.remove(m_table.begin()); }
 
--- a/Libraries/LibCore/EventLoop.cpp
+++ b/Libraries/LibCore/EventLoop.cpp
@@ -29,6 +29,7 @@
 #include <AK/JsonObject.h>
 #include <AK/JsonValue.h>
 #include <AK/NeverDestroyed.h>
+#include <AK/TemporaryChange.h>
 #include <AK/Time.h>
 #include <LibCore/Event.h>
 #include <LibCore/EventLoop.h>
@@ -40,6 +41,7 @@
 #include <LibThread/Lock.h>
 #include <errno.h>
 #include <fcntl.h>
+#include <signal.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
@@ -79,6 +81,10 @@ static NeverDestroyed<IDAllocator> s_id_allocator;
 static HashMap<int, NonnullOwnPtr<EventLoopTimer>>* s_timers;
 static HashTable<Notifier*>* s_notifiers;
 int EventLoop::s_wake_pipe_fds[2];
+HashMap<int, EventLoop::SignalHandlers> EventLoop::s_signal_handlers;
+int EventLoop::s_handling_signal = 0;
+int EventLoop::s_next_signal_id = 0;
+pid_t EventLoop::s_pid;
 static RefPtr<LocalServer> s_rpc_server;
 HashMap<int, RefPtr<RPCClient>> s_rpc_clients;
 
@@ -225,6 +231,7 @@ EventLoop::EventLoop()
 
     if (!s_main_event_loop) {
         s_main_event_loop = this;
+        s_pid = getpid();
 #if defined(SOCK_NONBLOCK)
         int rc = pipe2(s_wake_pipe_fds, O_CLOEXEC);
 #else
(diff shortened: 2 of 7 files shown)


tomuta:
I think this may be because of the introduction of `BlockResult::NotBlocked`

tomuta:
Hm even fixing this still doesn't make it work right

alimpfard:
Did you mean to revert dec27e5?
I don't think 5cdb0ef messes with anything interesting, it just skips setting pgrp to get the tty-related signals while the child is running

Libraries/LibCore/EventLoop.cpp
```diff
@@ -409,6 +412,91 @@ void EventLoop::post_event(Object& receiver, NonnullOwnPtr<Event>&& event)
...
+void EventLoop::handle_signal(int signo)
+{
+    // We need to protect the handler from being removed while handling it
+    ProtectSignalHandler protect(m_handling_signal, signo,
+        [&] {
+           ASSERT(m_handling_signal == signo); // should have been restored
+           unregister_signal(signo);
+        });
+    auto handler = m_signal_handlers.find(signo);
+    if (handler != m_signal_handlers.end()) {
```
alimpfard, line 454: This does not do what you think it does.
`HashMap::find` looks for hashes, not for keys.

alimpfard: Nevermind, that's the two-argument form.
the one-argument form looks for keys, my bad.

alimpfard:
Ah, I see what's wrong now.
You're setting the signal handlers in one event loop, so any loop on top of that won't know about the handlers.
You could move it to the global scope and just give it a unique key per loop (id+signal perhaps?).

Libraries/LibCore/EventLoop.cpp
```diff
@@ -409,6 +412,91 @@ void EventLoop::post_event(Object& receiver, NonnullOwnPtr<Event>&& event)
...
+
+void EventLoop::unregister_signal(int signo)
+{
+    ASSERT(signo != 0);
+    if (m_handling_signal == signo) {
+        // We're removing the signal while handling it, flag removal
+        m_handling_signal = 0;
+    } else {
+        if (m_signal_handlers.remove(signo)) {
+            signal(signo, SIG_DFL);
```
ADKaster, line 495: It might be slightly "correcter" to actually save the return value from `signal` and restore that? Is there a use case in the Shell for actually setting the signal mask for the signals it's trying to register handlers for? If so, exposing an API that looks more like `sigaction` would be the way to go. It is kind of sad that Serenity doesn't support the SA_SIGINFO flag yet, but that's just a wishlist item :)

Libraries/LibCore/EventLoop.cpp
tomuta: > Did you mean to revert [dec27e5](https://github.com/SerenityOS/serenity/commit/dec27e5e6fad25fedd9203087b8107242369dd2b)?
> I don't think [5cdb0ef](https://github.com/SerenityOS/serenity/commit/5cdb0ef2e5bd741e82073069fe8a5ae1b29ec0ff) messes with anything interesting, it just skips setting pgrp to get the tty-related signals while the child is running
Hm well it seemed that this new `m_is_accepting_signals` flag was trying to work mask the issue, but maybe that commit is really two separate things?

tomuta:
> Ah, I see what's wrong now.
> You're setting the signal handlers in one event loop, so any loop on top of that won't know about the handlers.
Hm I didn't realize there could be more than one event loop...
> You should move `m_signal_handlers` to the global scope (like the rest of the handlers :P)
Hm, you mean making it static/global? Not sure what other handlers you're referring to, I'm not very familiar with LibCore.

tomuta:
Thanks for the suggestions, I made those changes and now it seems like it doesn't hang anymore. It still crashes every once in a while but that seems to be another bug:

```
0xdeadc0de  _ZN6Kernel9Scheduler14context_switchEPNS_6ThreadE +183
0xdeadc0de  _ZN6Kernel9Scheduler9pick_nextEv +1549
0xdeadc0de  _ZN6Kernel9Scheduler5yieldEv +109
0xdeadc0de  _ZN6Kernel6Thread30yield_without_holding_big_lockEv +36
0xdeadc0de  _ZN6Kernel7Process11do_killselfEi +188
0xdeadc0de  _ZN6Kernel7Process8sys$killEii +318
0xdeadc0de  syscall_handler +1271
0xdeadc0de  syscall_asm_entry +49
0x08078cc3  raise +22
0x080480bd  abort +13
0x08075692  tolower +0
0x080507ed  _ZN3AST4JoinC1ENS_8PositionEN2AK6RefPtrINS_4NodeEEES5_ +89
0x0805bf39  _ZN6Parser6createIN3AST4JoinEJN2AK6RefPtrINS1_4NodeEEENS4_INS1_11SyntaxErrorEEEEEENS4_IT_EEDpT0_ +125
0x0805bb8a  _ZN6Parser5parseEv +110
0x0805f91a  _ZNK5Shell9highlightERN4Line6EditorE +100
0x0806f0b3  _ZN4Line6Editor15refresh_displayEv +313
0x08071f53  _ZNK2AK8FunctionIFvvEE15CallableWrapperIZN4Line6Editor8get_lineERKNS_6StringEEUlvE_E4callEv +77
0x0806b9ac  _ZN4Core6Object14dispatch_eventERNS_5EventEPS0_ +74
0x080652f9  _ZN4Core9EventLoop4pumpENS0_8WaitModeE +307
0x08065477  _ZN4Core9EventLoop4execEv +43
0x08070b9f  _ZN4Line6Editor8get_lineERKN2AK6StringE +877
0x08060181  _ZN5Shell16read_single_lineEv +69
0x080602f3  _ZN5Shell12custom_eventERN4Core11CustomEventE +29
0x0806b9ac  _ZN4Core6Object14dispatch_eventERNS_5EventEPS0_ +74
0x080652f9  _ZN4Core9EventLoop4pumpENS0_8WaitModeE +307
0x08065477  _ZN4Core9EventLoop4execEv +43
0x08048a10  main +1576
0x08048d4e  _start +94
```
Usually preceded by a log message like this:
```
Shell(23): Terminal DSR issue; received garbage col
```
But the commands seem to complete reliably now, regardless of how many cpu cores or whether Browser is hogging the cpu.

Libraries/LibCore/EventLoop.cpp
alimpfard: Not quite, `m_is_accepting_signals` is trying to mask away signals that the shell doesn't care about (e.g. SIGINT and SIGHUP) while a child process is running.
That's really to fix this interaction:
```sh
> cd
# change terminal size
^D
> # Shell now has the wrong idea about the terminal size
```

alimpfard:
> it still crashes every once in a while but that seems to be another bug

That looks really suspicious to me, a nullptr dereference in `Join::Join(Position, RefPtr<Node>, RefPtr<SyntaxError>>)`...?
what causes that? `ls|ls`?

The DSR issue is definitely odd, that's requested only at the start of a line, the only way to mess it up is to write something in the short window between LibLine requesting it, and the terminal responding to it...

alimpfard:
The Join crash is my issue, it can be reliably reproduced by trying to type `|ls`.
This PR is in the clear in regards to that~

AK/HashMap.h
```diff
@@ -54,13 +54,16 @@ class HashMap {
     size_t capacity() const { return m_table.capacity(); }
     void clear() { m_table.clear(); }
 
-    void set(const K& key, const V& value) { m_table.set({ key, value }); }
-    void set(const K& key, V&& value) { m_table.set({ key, move(value) }); }
-    void remove(const K& key)
+    bool set(const K& key, const V& value) { return m_table.set({ key, value }); }
```
bugaevc, line 57: I would never for the life of mine guess what it means if `HashMap::set()` returns `false` — it failed to set?

If we want to know whether the key was already present, we could make a special enum, perhaps like this:

```cpp
enum SetResult {
    InsertedNewEntry,
    ReplacedExistingEntry
};
```

tomuta: Good idea, I like it

tomuta:
Made a few changes, how is this looking?

Libraries/LibCore/EventLoop.h
```diff
@@ -92,7 +100,57 @@ class EventLoop {
         NonnullOwnPtr<Event> event;
     };
 
+    struct SignalHandlers {
+        AK_MAKE_NONCOPYABLE(SignalHandlers);
+        friend class EventLoop;
+        friend class HashMap<int, SignalHandlers>;
```
awesomekling, line 106: Uhh, why is this friendship necessary?

tomuta: The friendship for HashMap is needed because because `HashMap::remove` requires access to the `SignalHandlers` destructor, which is `private` because the entire class is considered `private`:
```
././AK/HashMap.h:58:68: error: 'Core::EventLoop::SignalHandlers::SignalHandlers(Core::EventLoop::SignalHandlers&&)' is private within this context
   58 |     HashSetResult set(const K& key, V&& value) { return m_table.set({ key, move(value) }); }
```

And the friendship for `EventLoop` is needed for this:
```
../Libraries/LibCore/EventLoop.cpp:466:34: error: 'void Core::EventLoop::SignalHandlers::dispatch()' is private within this context
  466 |         handlers->value.dispatch();
```

alimpfard: why is a struct private?
does `AK_MAKE_NONCOPYABLE` define a private ctor?

alimpfard: oh it leaks the `private`-ness too...
that needs fixing, @awesomekling:
```c++
#define AK_MAKE_NONCOPYABLE(c) \
private:                       \
    c(const c&) = delete;      \
    c& operator=(const c&) = delete;
```

should probably be (for structs):
```c++
#define AK_MAKE_NONCOPYABLE(c) \
private:                       \
    c(const c&) = delete;      \
    c& operator=(const c&) = delete; \
public:
```

is there a way to detect if some type is defined as a struct or a class...?

tomuta: Ah that's a good catch, I didn't realize these macros mess with visibility ;-)

tomuta: So how should we resolve this? I'm thinking that it shouldn't be necessary to mock with the visibility in these macros, because either way these functions are being deleted so there's no way code that needs them could compile, regardless of visibility. Or am I missing something?

awesomekling: What we normally do is just put a `public:` after them, e.g:

```cpp
class Foo {
    AK_MAKE_NONCOPYABLE(Foo);
public:
    ...
};
```

tomuta:
I think I addressed all suggestions. Any additional concerns?

Libraries/LibCore/EventLoop.cpp
```diff
@@ -409,10 +416,114 @@ void EventLoop::post_event(Object& receiver, NonnullOwnPtr<Event>&& event)
...
+        handlers->value.dispatch();
+    }
+}
+
+void EventLoop::handle_signal(int signo)
+{
+    ASSERT(signo != 0);
+    // We MUST check if the current pid still matches, because there
+    // is a window between fork() and exec() where a signal delivered
+    // to our fork could be inadvertedly routed to the parent process!
```
awesomekling, line 475: What does this mean? Signals are sent to a PID, and the fork will have a different PID?
So how would a signal delivered to the child end up in the parent?

tomuta: Because the pipe doesn't get closed until `exec`. `fork` inherits file descriptors. So, if you fork, if anything sends a signal to that new PID and that signal handler gets executed before `exec`, the signal handler will write to the pipe, and the parent process reads from the other end thinking it needs to asynchronously handle that signal.

bugaevc: By the way, it's not like `fork()` is always soon followed by `exec()`. It's perfectly valid to `fork()` without `exec`ing, and any design for handling signals in the `EventLoop` should take that into account.

tomuta: This is true on Linux, too. A few months ago I spent quite some time at work tracking down a bug where this is what ended up happening. Because the `exec` (on Linux) took a long time because it was a large program and the system was relatively busy just at the same time, the parent process sent a `SIGTERM` thinking it must be not responding, but the signal handler in the fork sent it into the cloned file descriptor, so the parent ended up receiving it in its own pipe shutting itself down by accident.

awesomekling: Ah yes, that's awkward. Perhaps we should solve this with a custom fd flag a la `O_CLOEXEC`.
`fcntl(fd, FD_SETFL, ... | FD_NOINHERIT);` or something like that.

tomuta: Yeah that would do it, too. This is the "portable" solution ;-) Not that we really aim to be...

tomuta: Actually, the problem I see with something like `FD_NOINHERIT` is that there is no way to detect that the process had been forked, unless you compare the PID. The only thing you would get is an error trying to write to the now closed file descriptor, but what if that file descriptor was re-allocated to something completely different already, like a socket? Seems like this is the only reliable way to solve this problem.

bugaevc: We've been thinking about introducing signalfd, and perhaps it would let each process read its own signals.

tomuta: We can always change it once that's in place. The small overhead really isn't a big deal.

awesomekling:
Okay let's do it. Longer term I think signalfds would be a nicer underlying implementation for this.

ADKaster:
https://ldpreload.com/blog/signalfd-is-useless

Would it? Seems it's very tricky to get correct :)

awesomekling:
Nobody says we have to copy the exact semantics of Linux signalfds. :)

PR #2720, opened 2020-07-06
Base: Add man page for test-js(1)

linusg:
cc @bugaevc who is not much into LibJS stuff (I think) but likes man pages and might want to have a look :slightly_smiling_face:

### DIFF
--- a/Base/usr/share/man/man1/test-js.md
+++ b/Base/usr/share/man/man1/test-js.md
@@ -1,0 +1,37 @@
+## Name
+
+test-js - run the LibJS test suite
+
+## Synopsis
+
+```**sh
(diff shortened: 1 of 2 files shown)


Base/usr/share/man/man1/test-js.md
```diff
@@ -0,0 +1,37 @@
+## Name
+
+test-js - run the LibJS test suite
+
+## Synopsis
+
+```**sh
+$ test-js [options...]
```
bugaevc, line 8: Since there's only one option, I'd write it out here instead:
```suggestion
$ test-js [--show-time]
```
But it's also fine this way

PR #2721, opened 2020-07-06
Kernel: Fix checking BlockResult

tomuta:
We now have BlockResult::WokeNormally and BlockResult::NotBlocked,
both of which indicate no error. We can no longer just check for
BlockResult::WokeNormally and assume anything else must be an
interruption.

This is related to #2718, but unfortunately not the solution. Still should be fixed, though.

### DIFF
--- a/Kernel/Thread.h
+++ b/Kernel/Thread.h
@@ -300,12 +300,42 @@ public:
     u64 sleep(u32 ticks);
     u64 sleep_until(u64 wakeup_time);
 
-    enum class BlockResult {
-        WokeNormally,
-        NotBlocked,
-        InterruptedBySignal,
-        InterruptedByDeath,
-        InterruptedByTimeout,
+    class BlockResult {
+    public:
+        enum Type {
+            WokeNormally,
+            NotBlocked,
+            InterruptedBySignal,
+            InterruptedByDeath,
+            InterruptedByTimeout,
+        };
+
+        BlockResult() = delete;
+
+        BlockResult(Type type)
+            : m_type(type)
+        {
+        }
+
+        bool operator==(Type type) const
+        {
+            return m_type == type;
+        }
+
+        bool was_interrupted() const
+        {
+            switch (m_type) {
+            case InterruptedBySignal:
+            case InterruptedByDeath:
+            case InterruptedByTimeout:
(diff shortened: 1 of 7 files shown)


Kernel/Thread.h
```diff
@@ -299,12 +299,42 @@ class Thread {
...
+
+        bool operator==(Type type) const
+        {
+            return m_type == type;
+        }
+
+        bool was_interrupted() const
+        {
+            switch (m_type) { 
+                case InterruptedBySignal:
```
awesomekling, line 327: Please use `clang-format`.

bugaevc:
> `BlockResult::NotBlocked`

If I understand correctly, this is what gets returned if a thread requests to wait on a queue, and the queue has been previously signalled to wake up someone, so the thread is not even put to sleep. Can we find a better name for it, perhaps `WokeUpImmediately`?

tomuta:
> > `BlockResult::NotBlocked`
> 
> If I understand correctly, this is what gets returned if a thread requests to wait on a queue, and the queue has been previously signalled to wake up someone, so the thread is not even put to sleep. Can we find a better name for it, perhaps `WokeUpImmediately`?

That's correct. But in my opinion `WokeUpImmediately` is a misnomer, because it never went into blocked state. It really was `NotBlocked`.

bugaevc:
Agreed; but on the other hand `BlockResult::NotBlocked` sounds as if blocking _failed_, whereas what happened is blocking succeeded but did not actually block because reasons.

tomuta:
> Agreed; but on the other hand `BlockResult::NotBlocked` sounds as if blocking _failed_, whereas what happened is blocking succeeded but did not actually block because reasons.

That's why I went the route of adding `BlockResult::was_interrupted` which makes it clear of what happened. We could add a method `BlockResult::was_blocked` and maybe even `BlockResult::was_blocked_or_interrupted` or something like that.

As I see it, there is no failure path, any failure path would be an assertion because there is no good way to handle it from a caller perspective.

tomuta:
Also, the caller typically doesn't care whether it actually blocked or not. They generally only care about whether they were interrupted.

awesomekling:
Seems fine with me.

PR #2723, opened 2020-07-07
LibJS: Symbol table fix

### LINKED ISSUE #2709: LibJS: Global symbol table persists across interpreter invocations
If a program creates two _separate_ JS interpreters, and runs programs that involve global symbols (i.e. symbols created from `Symbol.for(<description>)`), both interpreters will share the same global symbol table, since it's just a plain static property on `SymbolObject`.

This is why `Tests/builtins/Symbol/Symbol.for.js` and `Tests/builtins/Symbol/Symbol.keyFor.js` are commented out. It's a relatively simple fix, we just have to move the symbol table to the interpreter, but I don't want to mix that in to my test PR. I'll submit another PR after that one gets merged.

mattco98:
This PR moves the global symbol table from a static property on `SymbolObject` to a non-static property on `Interpreter`. Also makes `Symbol` non-copyable and non-movable. Fixes #2709.

### DIFF
--- a/Libraries/LibJS/Interpreter.cpp
+++ b/Libraries/LibJS/Interpreter.cpp
@@ -48,6 +48,19 @@ Interpreter::Interpreter()
     : m_heap(*this)
     , m_console(*this)
 {
+    m_well_known_symbol_map.set("iterator", js_symbol(*this, "Symbol.iterator", false));
+    m_well_known_symbol_map.set("asyncIterator", js_symbol(*this, "Symbol.asyncIterator", false));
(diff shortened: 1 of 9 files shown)


Libraries/LibJS/Interpreter.cpp
```diff
@@ -201,6 +214,21 @@ Reference Interpreter::get_reference(const FlyString& name)
     return { Reference::GlobalVariable, name };
 }
 
+Value Interpreter::get_global_symbol(const String& description) {
```
awesomekling, line 217: Please use `clang-format`

PR #2725, opened 2020-07-07
LibWeb: Fix PageView::url() null-check

kevmeyer:


awesomekling:
Yeah that's a big "duh" for me, thanks!

PR #2726, opened 2020-07-07
Help: Open the books!

thankyouverycool:
Adds new icons for the Help app and gives visual feedback when section nodes are toggled:
help


awesomekling:
Kick ass!

PR #2730, opened 2020-07-07
man pages: Consistently use "Otherwise," in the few places that said "Else,"

nico:
@bugaevc says this is house style (and `rg -l 'Otherwise' | wc -l` backs that up).


bugaevc:
Yep :)

PR #2731, opened 2020-07-07
Base: Update /etc/shellrc aliases

linusg:
- Remove "`alias sd=SystemDialog`" from `/etc/shellrc`
- Add "`alias ll='ls -l'`" to `/etc/shellrc`

### DIFF
--- a/Base/etc/shellrc
+++ b/Base/etc/shellrc
@@ -28,3 +28,4 @@ alias ws=WebServer
 alias sl=Solitaire
 alias wv=WebView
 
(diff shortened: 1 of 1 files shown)


Base/etc/shellrc
```diff
@@ -28,3 +28,4 @@ alias ws=WebServer
 alias sl=Solitaire
 alias wv=WebView
 
+alias ll='ls -l'
```
awesomekling, line 31: Literal madman

Issue #2649, opened 2020-06-27
Browser: Navigating backwards from iframe test crashes in layout


ADKaster:
Big ol stack trace below.

Essentially, just navigate to `file:///res/html/misc/iframe-subframe.html` then hit the "Go Back" button and watch the fireworks.

```
Browser(27): Loading iframe document from iframe-subframe.html
Browser(27): FrameLoader::load: file:///res/html/misc/iframe-subframe.html
Browser(27): I believe this content has MIME type 'text/html', encoding 'utf-8'
Browser(27): Parse error! void Web::HTMLDocumentParser::handle_initial(Web::HTMLToken&) @ 203
Browser(27): Stop parsing! :^)
[Browser(27:27)]: NP(error) fault at invalid address V0x00000190
[Browser(27:27)]: Unrecoverable page fault, read from address V0x00000190
[Browser(27:27)]: Note: Address V0x00000190 looks like a possible nullptr dereference
[Browser(27:27)]: CRASH: Page Fault. Ring 3.
[Browser(27:27)]: exception code: 0004 (isr: 0000
[Browser(27:27)]:   pc=001b:0807250c flags=0202
[Browser(27:27)]:  stk=0023:012020b0
[Browser(27:27)]:   ds=0023 es=0023 fs=0023 gs=07d3
[Browser(27:27)]: eax=00000000 ebx=012f10d2 ecx=01202080 edx=081a8294
[Browser(27:27)]: ebp=012020e8 esp=c4417fe8 esi=012f87ee edi=01202158
[Browser(27:27)]: cr0=80010013 cr2=00000190 cr3=026cd000 cr4=003006e4
[Browser(27:27)]: code: d9 80 90 01 00 00 d9 80
[Browser(27:27)]: 0x0807250c  Web::PageView::page_did_layout() +16

[Browser(27:27)]: 0xc015b462  Kernel::Process::crash(int, unsigned int, bool) +906
[Browser(27:27)]: 0xc011f0fd  illegal_instruction_handler +0
[Browser(27:27)]: 0xc01203fb  page_fault_handler +836
[Browser(27:27)]: 0xc011de03  page_fault_asm_entry +22
[Browser(27:27)]: 0x0806184e  Web::Frame::set_document(Web::Document*) +112
[Browser(27:27)]: 0x0806d7dd  Web::FrameLoader::resource_did_load() +359
[...]

PR #2733, opened 2020-07-07
LibWeb+WebContent: IFrame related fixes

kevmeyer:
Debugging https://github.com/SerenityOS/serenity/issues/2649 uncovered:

- that the history is adding all iframe urls
- going back/forward can cause the iframe's document to be laid out prior to the main frame's document (leading to a crash).

### DIFF
--- a/Libraries/LibWeb/DOM/Document.cpp
+++ b/Libraries/LibWeb/DOM/Document.cpp
@@ -246,7 +246,8 @@ void Document::layout()
     m_layout_root->layout();
     m_layout_root->set_needs_display();
 
-    frame()->page().client().page_did_layout();
+    if (frame()->is_main_frame())
+        frame()->page().client().page_did_layout();
 }
 
 void Document::update_style()
--- a/Libraries/LibWeb/PageView.cpp
+++ b/Libraries/LibWeb/PageView.cpp
@@ -160,6 +160,7 @@ String PageView::selected_text() const
 
 void PageView::page_did_layout()
 {
+    ASSERT(layout_root());
     set_content_size(layout_root()->size().to_int_size());
 }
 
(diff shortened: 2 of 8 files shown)


awesomekling:
Sure, that's reasonable :)

Services/WebContent/PageHost.cpp
```diff
@@ -116,9 +116,15 @@ void PageHost::page_did_change_selection()
     m_client.post_message(Messages::WebContentClient::DidChangeSelection());
 }
 
-void PageHost::page_did_layout()
+void PageHost::page_did_layout(Web::Document* document)
 {
     auto* layout_root = this->layout_root();
+    if (!layout_root && document != page().main_frame().document())
+    {
+        // If document of main frame was not yet layed out
```
linusg, line 124:
```suggestion
        // If document of main frame was not yet laid out
```

Libraries/LibWeb/PageView.cpp
```diff
@@ -158,8 +158,15 @@ String PageView::selected_text() const
     return builder.to_string();
 }
 
-void PageView::page_did_layout()
+void PageView::page_did_layout(Document* document)
 {
+    if (!layout_root() && document != this->document())
+    {
+        // If document of main frame was not yet layed out
```
linusg, line 165:
```suggestion
        // If document of main frame was not yet laid out
```

Libraries/LibWeb/DOM/Document.cpp
```diff
@@ -246,7 +246,7 @@ void Document::layout()
     m_layout_root->layout();
     m_layout_root->set_needs_display();
 
-    frame()->page().client().page_did_layout();
+    frame()->page().client().page_did_layout(this);
```
awesomekling, line 249: Instead of this, let's just only send this notification to the `PageClient` when the *main frame* is laid out.
It doesn't really need to know about subframe layouts anyway since that does not affect widget-level layout.

```suggestion
    if (frame()->is_main_frame())
        frame()->page().client().page_did_layout();
```

PR #2735, opened 2020-07-07
LibLine: Add ctrl-t shortcut for transposing characters

nico:


Libraries/LibLine/Editor.cpp
```diff
@@ -921,6 +921,18 @@ void Editor::handle_read_event()
...
             continue;
         }
+        // ^T
+        if (codepoint == ctrl('T')) {
+            if (m_cursor > 0 && m_buffer.size() >= 2) {
+                if (m_cursor < m_buffer.size())
+                    ++m_cursor;
+                u32 tmp = m_buffer[m_cursor - 1];
+                m_buffer[m_cursor - 1] = m_buffer[m_cursor - 2];
+                m_buffer[m_cursor - 2] = tmp;
```
alimpfard, line 931: don't we have a `swap`?

nico: Oh, we do! I tried looking for it, but only in Libraries, because there's some gigantic json file in AK that matches most words that makes searching in AK difficult. (How do others deal with that?) Fixing...

nico: Done, better, thanks :)

alimpfard:
This will mess the anchored styles, but no one uses those right now.
```
> foo b|ar
        -- <- anchored style
> foo a|br
        -- <- style now here
what should happen
> foo a|br
      -  - <- style like this
```
could you put a little FIXME there about that?

nico:
Done. Isn't that a problem for ctrl-u, delete, etc too?

alimpfard:
Nah, the remove function they call handles that.

Libraries/LibLine/Editor.cpp
```diff
@@ -921,6 +921,17 @@ void Editor::handle_read_event()
             }
             continue;
         }
+        // ^T
```
awesomekling, line 924: ^Today I learned about this

alimpfard: why are these puns all over the place ^today?

Issue #1289, opened 2020-02-24
Close window on double clicking window icon


diversys:
It would be handy if double clicking a window icon would close current window/app.

danboid:
Is it just me that doesn't understand this or why you'd want this feature?

This kinda reminds of FF/Chromes closing tabs with a middle click which annoys the hell out of me, if you mean what I think you do.

Pulgovisk:
@danboid Windows has this feature. It's "useless", but sometimes save some time.

bugaevc:
Let's not have this, it's not intuitive at all. We have a close button for a reason.

I wouldn't be against making middle-click closing windows though, exactly because this is already familiar from how I use the browser, and it brings additional value — the whole titlebar is large, so you don't have to hunt for that close button.

diversys:
I see it as a convenience shortcut for the cases when cursor happens to be near the window icon. This is also helpful when a window is partially off-screen, so you don't have to drag it back first to be able to close it with normal close button.

willmcpherson2:
I vote for middle-click on the taskbar and titlebar

danboid:
The window icon menu already has a close option. That seems sufficient to me.

It shouldn't be too easy to close windows by accident IMO. It really frustrates me when I accidentally close browser tabs via middle click even though I can pretty easily re-open the the tab. I've always wanted to disable this feature in browsers and I don't want Serenity to replicate it.

If someone does add this, please add the option to disable it.

pjanx:
If Serenity UI is supposed to mimic old Windows, this should be the default.
It gets irritating when things don't work as they're supposed to.

Anyway, for this to be implemented, the location of the popup will need to be changed to show underneath the icon. That way the decoration can receive the double click event, otherwise the popup window stands in the way.

xTibor:
> If Serenity UI is supposed to mimic old Windows

If you check the project manifesto, it's not supposed to mimic Windows. It's supposed to lift interface ideas from all over the place. Just to set the record straight.

pjanx:
It turns out a patch is not straight-forward because double click handling is quirky. I'll just leave this here as it happens to work. Don't beat me.
```patch
diff --git a/Servers/WindowServer/WindowFrame.cpp b/Servers/WindowServer/WindowFrame.cpp
index fac04e933..7f6737d61 100644
--- a/Servers/WindowServer/WindowFrame.cpp
+++ b/Servers/WindowServer/WindowFrame.cpp
[...]

PR #2736, opened 2020-07-07
WindowServer: Add support for default MenuItem

tomuta:
This allows marking a MenuItem as a default action, e.g. in a
context menu for an action that reflects what e.g. a double click
would perform.

Also enhance the window menu to mark the close action as the
default, and when double clicked perform that action.

I keep finding myself trying to close windows by double clicking on the window menu (I know, bad behavior)... This makes it more "natural" to my habits. No, I haven't used Windows much for a long, long time, but it's hard to get rid of habits...

Screenshot from 2020-07-07 11-25-47

### DIFF
--- a/Libraries/LibGUI/Menu.h
+++ b/Libraries/LibGUI/Menu.h
@@ -27,7 +27,9 @@
 #pragma once
 
 #include <AK/NonnullOwnPtrVector.h>
+#include <AK/WeakPtr.h>
 #include <LibCore/Object.h>
+#include <LibGUI/Action.h>
 #include <LibGUI/Forward.h>
 #include <LibGfx/Forward.h>
 
@@ -39,11 +41,7 @@ public:
     explicit Menu(const StringView& name = "");
     virtual ~Menu() override;
 
-    void realize_menu_if_needed()
-    {
-        if (menu_id() == -1)
-            realize_menu();
-    }
+    void realize_menu_if_needed();
 
     static Menu* from_menu_id(int);
     int menu_id() const { return m_menu_id; }
@@ -56,19 +54,20 @@ public:
     void add_separator();
     Menu& add_submenu(const String& name);
 
-    void popup(const Gfx::IntPoint& screen_position);
+    void popup(const Gfx::IntPoint& screen_position, const RefPtr<Action>& default_action = nullptr);
     void dismiss();
 
 private:
     friend class MenuBar;
 
-    int realize_menu();
+    int realize_menu(RefPtr<Action> default_action = nullptr);
     void unrealize_menu();
-    void realize_if_needed();
+    void realize_if_needed(const RefPtr<Action>& default_action);
 
     int m_menu_id { -1 };
     String m_name;
     NonnullOwnPtrVector<MenuItem> m_items;
+    WeakPtr<Action> m_last_default_action;
 };
 
 }
--- a/Applications/FileManager/main.cpp
+++ b/Applications/FileManager/main.cpp
@@ -764,7 +764,7 @@ int run_in_windowed_mode(RefPtr<Core::ConfigFile> config, String initial_locatio
                 folder_specific_paste_action->set_enabled(should_get_enabled);
                 directory_context_menu->popup(event.screen_position());
             } else {
-                file_context_menu->popup(event.screen_position());
+                file_context_menu->popup(event.screen_position(), open_in_text_editor_action);
             }
         } else {
             directory_view_context_menu->popup(event.screen_position());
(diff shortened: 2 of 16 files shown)


awesomekling:
The concept of a default menu action doesn't seem generally useful outside of this one specific situation.
Where else would it be used?

tomuta:
> The concept of a default menu action doesn't seem generally useful outside of this one specific situation.
> Where else would it be used?

E.g. right clicking something in File Manager, the action that would be performed by a double click (open the folder, open the file, execute a program, etc)?

tomuta:
To me it's a usability feature I miss from the good old days ;-)

awesomekling:
Ohhhh the text is in *bold*! I didn't realize, duh. Now I get it! :smile:

awesomekling:
And by that I mean **bold**, obviously.

tomuta:
Yeah, it highlights what it would have done if you had clicked or double-clicked something (if there is a context menu).

tomuta:
Screenshot from 2020-07-07 16-45-16

diversys:
This seems to fix https://github.com/SerenityOS/serenity/issues/1289.

bugaevc:
> E.g. right clicking something in File Manager, the action that would be performed by a double click (open the folder, open the file, execute a program, etc)?

So, an action that would be performed on activation?

Does anything ensure that this default action actually does the same thing as activation, or is it up to individual apps/widgets to ensure?

awesomekling:
> > E.g. right clicking something in File Manager, the action that would be performed by a double click (open the folder, open the file, execute a program, etc)?
> 
> So, an action that would be performed on activation?
> 
> Does anything ensure that this default action actually does the same thing as activation, or is it up to individual apps/widgets to ensure?

Great question. Can we find a way to make this programmatic?

Also, I don't think it makes sense for an action to be "default", as an action can occur in multiple context.
You can add the same action to a toolbar, menu, etc.

It seems that `is_default` should be a parameter to the `add_action()` API.

tomuta:
> > > E.g. right clicking something in File Manager, the action that would be performed by a double click (open the folder, open the file, execute a program, etc)?
> > 
> > 
> > So, an action that would be performed on activation?
> > Does anything ensure that this default action actually does the same thing as activation, or is it up to individual apps/widgets to ensure?
> 
> Great question. Can we find a way to make this programmatic?
> 
> Also, I don't think it makes sense for an action to be "default", as an action can occur in multiple context.
> You can add the same action to a toolbar, menu, etc.

It's only a visual indicator (bold text) applied to menu items. Just like it is only a visual indicator on menu items. Similar to it e.g. being checked or not.

> It seems that `is_default` should be a parameter to the `add_action()` API.
I thought about that, but since this is a rather rarely used feature I thought it would be better to just add a separate `set_default` call if one really wants an action marked as "default".

Maybe we just need to come up with a better name? If I remember correctly from my old Windows days, that's what they call it there, and similarly it's just a visual indicator.

awesomekling:
My point is that the action may be the default behavior in one context, but not in another.

tomuta:
> My point is that the action may be the default behavior in one context, but not in another.

Hm. What if we when realizing a context menu provide an optional default action, and if then compares the actions in the menu to that, and if it matches then it calls `MenuItem::set_default`? This way, actions wouldn't have a "default" flag, but when you display a context menu then you could also indicate what you consider the "default", and the menu would get populated as expected?

tomuta:
How does this look? I reworked it, so that now you can specify an optional default action when calling `Menu::popup`, since these things are clearly only applicable to context menu items anyway. I removed all the default stuff from `Action`. I quite like this ;)

Libraries/LibGUI/Menu.h
```diff
@@ -56,19 +56,20 @@ class Menu final : public Core::Object {
...
-    int realize_menu();
+    int realize_menu(Action* default_action = nullptr);
     void unrealize_menu();
-    void realize_if_needed();
+    void realize_if_needed(Action* default_action);
 
     int m_menu_id { -1 };
     String m_name;
     NonnullOwnPtrVector<MenuItem> m_items;
+    Action* m_last_default_action { nullptr };
```
awesomekling, line 72: Using a raw pointer like this is pretty sloppy. Let's use a `WeakPtr<Action>` instead :)

tomuta: Well, the pointer is only ever compared to determine a match, the object is never accessed using it. Still want me to change it to a WeakPtr instead?

tomuta: Ok, changed it to a WeakPtr

nico:
If I right-click a `.font` file in File Manager, the context menu shows in bold "Open in Text Editor...", but if I double-click the file it opens in Font Editor. is this expected?

tomuta:
> If I right-click a `.font` file in File Manager, the context menu shows in bold "Open in Text Editor...", but if I double-click the file it opens in Font Editor. Is this expected?

That was just a missing feature. Added a pull request: #2795

PR #2739, opened 2020-07-07
WindowServer+LibGUI+Base: Moar cursors

linusg:
Added new cursors:

- resize row
- resize column
- help
- wait

image image image image

Applied new cursors:

- resize row/column is now used for splitter and table view
- wait cursor is applied to unresponsive windows

image

Added new tab to WidgetGallery to demo all the cursors:

image

---

These are also going to be useful for LibWeb as there's a vast amount of cursors to be supported: https://developer.mozilla.org/en-US/docs/Web/CSS/cursor


alimpfard:
> https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Basic_User_Interface/Using_URL_values_for_the_cursor_property

_/me tries that_
_/me screams_

This is...just no...why?

PR #2743, opened 2020-07-08
Kernel: Detect syscall/sysenter support

tomuta:


tomuta:
Anyone know how to enable sysenter (sep) in qemu? I tried `-cpu max,+sep` with and without kvm, no dice :( Tried all sorts of models for `-cpu` as well...

Kernel/Arch/i386/CPU.cpp
```diff
@@ -788,10 +788,20 @@ void Processor::cpu_detect()
         set_feature(CPUFeature::TSC);
     if (processor_info.ecx() & (1 << 30))
         set_feature(CPUFeature::RDRAND);
-
+    if (processor_info.edx() & (1 << 11)) {
+        u32 stepping = processor_info.eax() & 0xf;
+        u32 model = (processor_info.eax() >> 4) & 0xf;
+        u32 family = (processor_info.eax() >> 8) & 0xf;
```
ADKaster, line 794: The more of these flags we check, the more it feels like a bunch of defines or enums for the feature flags might be useful...

PR #2744, opened 2020-07-08
LibJS: Integrate symbols into objects as valid keys

mattco98:
This PR finally allows Symbols to be used as object properties alongside strings. This is round two of attempting to implement this feature. The `StringOrSymbol` class is designed to be as memory efficient as possible, being a wrapper around only a single tagged pointer ([see comment here](https://github.com/SerenityOS/serenity/pull/2034#discussion_r418582570)).

I have a few questions though. First of all, for my tagged pointer, I assumed that the lowest bit of a pointer is normally always zero, since that's what I found to be true while testing both inside serenity and on my host machine. Is this a safe assumption to make?

Secondly, I was unable to figure out how to use FlyString'd `StringImpl`s as the comment above suggests. Whenever I tried, I ended up getting `RefCounted` errors. Some help with this would be much appreciated :)

Might have merge conflicts with #2723, not sure, so merge that one first.

cc @linusg

### DIFF
--- a/Libraries/LibJS/AST.cpp
+++ b/Libraries/LibJS/AST.cpp
@@ -65,6 +65,15 @@ static void update_function_name(Value& value, const FlyString& name)
     }
 }
 
+static String get_function_name(Interpreter& interpreter, Value value)
+{
+    if (value.is_symbol())
+        return String::format("[%s]", value.as_symbol().description().characters());
+    if (value.is_string())
+        return value.as_string().string();
+    return value.to_string(interpreter);
+}
+
 Value ScopeNode::execute(Interpreter& interpreter, GlobalObject& global_object) const
 {
     return interpreter.run(global_object, *this);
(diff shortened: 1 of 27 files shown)


Dexesttp:
Since symbols live on the heap, would they get GC'd if the symbol got out of scope but an object with the symbol as key still existed ?
Something like :
```js
let o = (() => {
    let s = Symbol("test");
    return { [s]: 123 };
})();
// Trigger GC here, cleaning up the "s" variable
var symbol = Object.getOwnPropertySymbols(o)[0];
// Is "symbol" valid or was the cell dropped ?
```
I don't yet understand this PR well enough to be sure of whether the above case was handled or not, but I couldn't find any changes to the shape or the object's `visit_children()` (where I'd assume it would be).

mattco98:
Ahh, yeah you're right. I always seem to forget about the GC when I do this kind of stuff.

awesomekling:
This looks great. :)

PR #2746, opened 2020-07-08
LibGUI: Implement IconView auto-scrolling when trying to select items…

tomuta:
If selecting in a large icon view, and dragging the mouse outside
of the widget, or close to the border of it, start scrolling
automatically. This allows for selecting large amount of items
that exceed the amount that can be displayed.

iconview_scrolling


awesomekling:
Just delightful :)

PR #2749, opened 2020-07-09
LibGUI: Make scrollbar thumb size relative to content size

tomuta:
In order to calculate a thumb size that is a representation
of the visible portion (page) of the content, that information
needs to be taken into account.

scrollbar_size


awesomekling:
Loving these patches :)

PR #2750, opened 2020-07-09
Toolchain: Allow building using CMake on macOS

Pagghiu:
This is a single phase build patch to allow compiling on macOS.
This is based on @bugaevc base patch as discussed in #2519 and effectively replaces that PR.
I have been fixing a few other compile / linking issues and I've been changing `BuildIt.sh` to make it work even without bash 4 installed.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -126,6 +126,22 @@ set(CMAKE_RANLIB ${TOOLCHAIN_PREFIX}ranlib)
 set(CMAKE_STRIP ${TOOLCHAIN_PREFIX}strip)
 set(CMAKE_AR ${TOOLCHAIN_PREFIX}ar)
 
+foreach(lang ASM C CXX OBJC OBJCXX)
+    unset(CMAKE_${lang}_OSX_COMPATIBILITY_VERSION_FLAG)
+    unset(CMAKE_${lang}_OSX_CURRENT_VERSION_FLAG)
+    unset(CMAKE_${lang}_LINK_FLAGS)
+    unset(CMAKE_SHARED_LIBRARY_CREATE_${lang}_FLAGS)
+    unset(CMAKE_SHARED_MODULE_CREATE_${lang}_FLAGS)
+    unset(CMAKE_SHARED_MODULE_LOADER_${lang}_FLAG )
+    unset(CMAKE_${lang}_OSX_DEPLOYMENT_TARGET_FLAG)
+    unset(CMAKE_${lang}_SYSROOT_FLAG)
+    unset(CMAKE_SHARED_LIBRARY_SONAME_${lang}_FLAG)
+endforeach()
+
+set(CMAKE_INSTALL_NAME_TOOL "true")
+set(CMAKE_SHARED_LIBRARY_SUFFIX ".so")
+set(CMAKE_SHARED_LIBRARY_CREATE_CXX_FLAGS "-shared")
+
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -126,6 +126,22 @@ set(CMAKE_RANLIB ${TOOLCHAIN_PREFIX}ranlib)
 set(CMAKE_STRIP ${TOOLCHAIN_PREFIX}strip)
 set(CMAKE_AR ${TOOLCHAIN_PREFIX}ar)
 
+foreach(lang ASM C CXX OBJC OBJCXX)
+  unset(CMAKE_${lang}_OSX_COMPATIBILITY_VERSION_FLAG)
```
bugaevc, line 130: Nitpick: the indentation here should be four spaces.

Pagghiu: sure

Toolchain/BuildIt.sh
```diff
@@ -230,12 +238,20 @@ pushd "$DIR/Build/"
...
             cmake --build . --target LibC
-            install -D Libraries/LibC/libc.a Libraries/LibM/libm.a Root/usr/lib/
+            "$INSTALL" -D Libraries/LibC/libc.a Libraries/LibM/libm.a Root/usr/lib/
             SRC_ROOT=$(realpath "$DIR"/..)
-            for header in "$SRC_ROOT"/Libraries/Lib{C,M}/**/*.h; do
-                target=$(echo "$header" | sed -e "s@$SRC_ROOT/Libraries/LibC@@" -e "s@$SRC_ROOT/Libraries/LibM@@")
-                install -D "$header" "Root/usr/include/$target"
-            done
+            if [ "$(uname)" = "Darwin" ]; then
+                FILES=`find "$SRC_ROOT"/Libraries/LibC "$SRC_ROOT"/Libraries/LibM -name '*.h' -print`
```
bugaevc, line 244: Uhh, I've been trying to avoid conditioning anything on Apple or Darwin (and telling you about it the whole time). Let's support older Bash versions unconditionally, so that we have less code paths to maintain, and so that other people/systems that use older Bash can benefit from it.

Using `find` for this is a good idea; but please use `$()` instead of backticks, and name the variable something more descriptive (`HEADERS`, perhaps). Also I'm concerned about file names that contain spaces and other fun bytes, but guess that was already broken here.

ADKaster: Would it be better to tell people on MacOS to install a more recent version of bash than 3.x via homebrew?

bugaevc: We are already telling them to install Bash via homebrew, by the way, but we don't mention a version explicitly.

Pagghiu: I've been removing the "Darwin" branching. 
My understanding was that the conditioning on Apple or Darwin was to be avoided mainly on the build files and not on shell, also because there are already Apple/Darwin conditioning in files.
I agree anyway that it's much better to reduce them when technically possible like in this case!

Regarding @ADKaster consideration, I think it's easier to reduce the number of things that people need to install, in particular if it requires a few line changes. This script works also on zsh, that is default macOS shell from Catalina going forward.

Pagghiu:
Is there anything else I need to be doing to get this merged? 😃

bugaevc:
It's all fine by me, up to @awesomekling now :)

Pagghiu:
Thanks! :^)

PR #2751, opened 2020-07-09
LibJS: Start implementing iterable framework, add ArrayIterator

mattco98:
With the addition of symbol keys, work can now be done on starting to implement the well-known symbol functionality. The most important of these well-known symbols is by far `Symbol.iterator`.

This PR adds `IteratorPrototype`, as well as `ArrayIterator` and `ArrayIteratorPrototype`. In the future, sometime after `StringIterator` has also been added, this will allow us to use `Symbol.iterator` directly in `for..of` loops, enabling the use of custom iterator objects. Also makes adding iterator support to native objects much easier (as will have to be done for `Map` and `Set`, when they get added).

### DIFF
--- a/Libraries/LibJS/Runtime/ArrayIterator.cpp
+++ b/Libraries/LibJS/Runtime/ArrayIterator.cpp
@@ -1,0 +1,49 @@
+/*
+ * Copyright (c) 2020, Matthew Olsson <matthewcolsson@gmail.com>
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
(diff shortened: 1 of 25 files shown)


Libraries/LibJS/Runtime/IteratorPrototype.cpp
```diff
@@ -0,0 +1,55 @@
...
+}
+
+IteratorPrototype::~IteratorPrototype()
+{
+}
+
+JS_DEFINE_NATIVE_FUNCTION(IteratorPrototype::symbol_iterator)
+{
+    auto* this_object = interpreter.this_value(global_object).to_object(interpreter, global_object);
+    if (interpreter.exception() || !this_object)
```
linusg, line 50: Usually we just check the returned pointer when using `to_object()` which is probably enough here as well

Libraries/LibJS/Runtime/IteratorOperations.h
```diff
@@ -0,0 +1,49 @@
...
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#pragma once
+
+#include <LibJS/Runtime/Object.h>
+
+namespace JS {
+
+// Common iterator operations defined in ECMA262 7.4
```
linusg, line 33: Not sure how others feel about this, but personally I like to link to specs if I mention them in the code anyway :)

Also maybe we don't need a newline after each declaration?

Libraries/LibJS/Runtime/ArrayIterator.cpp
```diff
@@ -0,0 +1,49 @@
...
+
+#include <LibJS/Interpreter.h>
+#include <LibJS/Runtime/ArrayIterator.h>
+#include <LibJS/Runtime/GlobalObject.h>
+
+namespace JS {
+
+ArrayIterator* ArrayIterator::create(GlobalObject& global_object, Value array, Object::PropertyKind iteration_kind)
+{
+    return global_object.interpreter().heap().allocate<ArrayIterator>(global_object, *global_object.array_iterator_prototype(), array, iteration_kind);
```
linusg, line 35:
```suggestion
    return global_object.heap().allocate<ArrayIterator>(global_object, *global_object.array_iterator_prototype(), array, iteration_kind);
```

Libraries/LibJS/Runtime/ArrayIteratorPrototype.cpp
```diff
@@ -0,0 +1,92 @@
...
+    if (iteration_kind == Object::PropertyKind::Key)
+        return create_iterator_result_object(interpreter, global_object, Value(static_cast<i32>(index)), false);
+
+    auto value = array.get(index);
+    if (interpreter.exception())
+        return {};
+    if (iteration_kind == Object::PropertyKind::Value)
+        return create_iterator_result_object(interpreter, global_object, value, false);
+
+    Array* entry_array = Array::create(global_object);
```
linusg, line 86:
```suggestion
    auto* entry_array = Array::create(global_object);
```

Libraries/LibJS/Runtime/ArrayIteratorPrototype.cpp
```diff
@@ -0,0 +1,92 @@
...
+    entry_array->put(0, Value(static_cast<i32>(index)));
+    entry_array->put(1, value);
```
linusg, line 88: We should probably be more consistent with how we add values to an array - `indexed_properties().append()` vs `define_property()` vs `put()`. Feel free to decide which one is actually the best though! :smile:

https://github.com/SerenityOS/serenity/blob/ba641e97d9326be94b0b281fe140b5849c626b50/Libraries/LibWeb/Bindings/NavigatorObject.cpp#L45

https://github.com/SerenityOS/serenity/blob/7a1d485b19a43b3538b3ac69040ebc2b3bb14d85/Libraries/LibJS/Runtime/Object.cpp#L196-L203

mattco98: Great point. We should be using `define_property` when we've just created the object, since it avoids the prototype chain lookups and inserts the property directly into the object. When we're putting into an object we've received, we should use `put`. `indexed_properties.append` avoids a couple lines of code, but probably isn't worth it :)

awesomekling:
Oops, this has a conflict, sorry about that!

mattco98:
All good :) should be good to go now

awesomekling:
👍

PR #2753, opened 2020-07-10
LibGUI: Multiple on_update handlers for Model

tomuta:
This solves a problem where the SortingProxyModel doesn't
receive the on_update call because other code overwrote
the handler later on.

Also, SortingProxyModel needs to lazily re-populate its
cache. Doing so in the on_update handler may not have any
items available yet, so merely flag it to require another
resort/cache update when accessed again.

### DIFF
--- a/Libraries/LibGUI/Model.h
+++ b/Libraries/LibGUI/Model.h
@@ -95,7 +95,8 @@ public:
     void register_view(Badge<AbstractView>, AbstractView&);
     void unregister_view(Badge<AbstractView>, AbstractView&);
 
-    Function<void()> on_update;
+    int register_update(Function<void(unsigned)>);
+    void unregister_update(int);
 
 protected:
     Model();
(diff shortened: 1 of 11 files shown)


tomuta:
This solves the infamous SortingProxyModel bug log message. I have another change in the works that depends on this being fixed...

tomuta:
Hm System Monitor has some odd issue with int on the File Systems page

tomuta:
Ok, got that resolved, too

Libraries/LibGUI/Model.h
```diff
@@ -107,6 +108,7 @@ class Model : public RefCounted<Model> {
 
 private:
     HashTable<AbstractView*> m_views;
+    Vector<Function<void(unsigned)>> m_on_update;
```
awesomekling, line 111: Instead of having both of these, how about this:

An abstract `ModelClient` class that you can inherit from, and each `Model` has a set of N clients.
When something interesting happens, we iterate through the set of clients and call a virtual on each one.
Both `AbstractView` and `SortingProxyModel` then inherit from `ModelClient` and we funnel everything through a uniform interface. :)
That also makes it a bit more generic and opens up for easy extensibility with more hooks.

tomuta: I made these changes in another merge request: #2761
I'm closing this one.

PR #2754, opened 2020-07-10
AK: Add template argument to all uses of Json{Array,Object}Serializer

linusg:
This makes AK/Json* build properly from "outside" Serenity, i.e. by using Serenity's CMake toolchain file. Without these changes it would fail with "missing template arguments" errors.

---

Disclaimer: I have absolutely no idea what I'm doing, this was all trial and error. Serenity was building fine and still is, but any attempt to `#include <AK/JsonObject.h>` would result in a flood of errors when building something for Serenity but from the outside:

```
[ 50%] Building CXX object CMakeFiles/FooProject.dir/src/main.cpp.obj
In file included from /home/linus/Dev/git/serenity/Build/Root/usr/include/AK/JsonObject.h:30,
                 from /home/linus/Dev/git/FooProject/src/main.cpp:1:
/home/linus/Dev/git/serenity/Build/Root/usr/include/AK/JsonArray.h: In member function 'void AK::JsonArray::serialize(Builder&) const':
/home/linus/Dev/git/serenity/Build/Root/usr/include/AK/JsonArray.h:99:25: error: missing template arguments before 'serializer'
   99 |     JsonArraySerializer serializer { builder };
      |                         ^~~~~~~~~~
```

### DIFF
--- a/AK/JsonArray.h
+++ b/AK/JsonArray.h
@@ -96,7 +96,7 @@ private:
 template<typename Builder>
 inline void JsonArray::serialize(Builder& builder) const
 {
-    JsonArraySerializer serializer { builder };
+    JsonArraySerializer<Builder> serializer { builder };
     for_each([&](auto& value) { serializer.add(value); });
 }
 
--- a/AK/JsonObject.h
+++ b/AK/JsonObject.h
@@ -129,7 +129,7 @@ private:
 template<typename Builder>
 inline void JsonObject::serialize(Builder& builder) const
 {
-    JsonObjectSerializer serializer { builder };
+    JsonObjectSerializer<Builder> serializer { builder };
     for_each_member([&](auto& key, auto& value) {
         serializer.add(key, value);
     });
--- a/AK/JsonObjectSerializer.h
+++ b/AK/JsonObjectSerializer.h
@@ -130,13 +130,13 @@ public:
     JsonArraySerializer<Builder> add_array(const StringView& key)
     {
         begin_item(key);
-        return JsonArraySerializer(m_builder);
+        return JsonArraySerializer<Builder>(m_builder);
     }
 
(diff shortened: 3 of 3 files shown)


predmond:
Serenity builds with `-std=c++2a`. What version were you compiling with?

alimpfard:
That's...kinda odd, is lagom building with c++17?

alimpfard:
Lagom _is_ being built with c++17
https://github.com/SerenityOS/serenity/blob/f4ddca0a73fb2d911a7c068cc38524806ff8d748/Meta/Lagom/CMakeLists.txt#L4

predmond:
> Lagom _is_ being built with c++17
> 
> https://github.com/SerenityOS/serenity/blob/f4ddca0a73fb2d911a7c068cc38524806ff8d748/Meta/Lagom/CMakeLists.txt#L4

he's building outside the serenity source tree (`/home/linus/Dev/git/FooProject/src/main.cpp`)

alimpfard:
Oh, I'm blind.
well, @linusg just build with c++20...and someone should probably bump that lagom version anyway

linusg:
I added that `CMAKE_CXX_FLAGS` line to my `CMakeLists.txt` and now it indeed builds without this PR's changes. So I'll assume it's a new-ish feature?

Either way, doesn't seem like this is needed :smile:

predmond:
> I added that `CMAKE_CXX_FLAGS` line to my `CMakeLists.txt` and now it indeed builds without this PR's changes. So I'll assume it's a new-ish feature?
> 
> Either way, doesn't seem like this is needed 😄

This feature is class template argument deduction which was added in c++17 (see https://devblogs.microsoft.com/cppblog/how-to-use-class-template-argument-deduction/)

alimpfard:
oh no a microsoft link as reference
__/me runs far away__

PR #2755, opened 2020-07-10
LibWeb: Verify that layout_node still exists after dispatch_event

kevmeyer:


Libraries/LibWeb/Frame/EventHandler.cpp
```diff
@@ -124,6 +124,10 @@ bool EventHandler::handle_mousedown(const Gfx::IntPoint& position, unsigned butt
 
     auto offset = compute_mouse_event_offset(position, *result.layout_node);
     node->dispatch_event(MouseEvent::create("mousedown", offset.x(), offset.y()));
+
+    if (!this->layout_root())
+        return false;
```
awesomekling, line 129: This is not completely correct as there may be a new layout tree in place, and so this check would fail.

Instead, what we should do is just not cache a pointer to the layout root across a call to event dispatch, since as you noted, event dispatch runs arbitrary code and the layout tree may be gone or mutated after returning :)

awesomekling:
Great :)

PR #2759, opened 2020-07-11
WebServer: Allow the user to specify the base directory

alimpfard:
This commit makes the WebServer accept a base path to serve.
This makes taking files out of the system significantly more simple (depending on whom you ask).

cc @bugaevc for nitpicks and possibly killing me for passing a dynamic path to `unveil`.

### DIFF
--- a/Services/WebServer/Client.cpp
+++ b/Services/WebServer/Client.cpp
@@ -38,9 +38,10 @@
 
 namespace WebServer {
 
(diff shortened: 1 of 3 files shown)


bugaevc:
LGTM, and unveiling a dynamic path is not actually that bad 😀

awesomekling:
Neat! :)

Issue #2365, opened 2020-05-24
Meta: Some way to opt in to checking copyright  headers before a push


alimpfard:
It would be nice to avoid further "Add copyright header" commits, and automate the check that we all manually perform...sometimes.

My proposed idea was to add a pre-push hook to git, making git refuse to push until all the copyright headers are there.

Perhaps enabled by something like `Meta/opt-in-to-pre-push-copyright-header-checks.sh` (likely dumb example)
```sh
#!/bin/bash

script_path=$(cd -P -- "$(dirname -- "$0")" && pwd -P)
cd "$script_path/.." || exit 1

git config core.hooksPath ./git/hooks
mkdir -p git/hooks
echo "#!/bin/bash ${script_path}/check-license-headers.sh" > git/hooks/pre-push
```

linusg:
I had a good experience with https://pre-commit.com in the past - that'd mean putting a `.pre-commit-config.yaml` file in the repo root, and people can opt into using it by installing `pre-commit` via pip/homebrew and running `pre-commit install` - it then installs stuff :tm: in `./git/hooks`.

That way various other hooks could be added easily, like running clang-format before committing - but I also understand if people think that's overkill.

alimpfard:
I'm open to anything that makes travis not blame my commits :^)
while pre-commit.com feels rather overkill, it might come in handy for clang-format and all, I see that being useful.

awesomekling:
Seems useful. If someone knows how to set this up, please do :)

PR #2761, opened 2020-07-11
LibGUI: Improve Iconview selection performance

tomuta:
This implements the following optimizations:

* Rather than clearing a HashTable of selected items and re-populating
  it every time the selection rectangle changes, determine the delta
  by only examining the items that might be in the area where the
  selection may have changed compared to the previous area. Then
  only add/remove selection items as needed.

* When painting, only query and paint the items actually visible.
  Also, keep a local cache of item information such as calculated
  rectangles and selection state, so it doesn't have to be calculated
  over and over again.

Watch the CPU graph on the top right. The CPU usage on the previous implementation spiked very quickly and the system becomes laggy once the mouse is moved around rather quickly.

Before:
iconview_selection_before

After:
iconview_selection_after

### DIFF
--- a/Applications/Browser/BookmarksBarWidget.h
+++ b/Applications/Browser/BookmarksBarWidget.h
@@ -27,11 +27,13 @@
 #pragma once
 
 #include <LibGUI/Forward.h>
+#include <LibGUI/Model.h>
 #include <LibGUI/Widget.h>
 
 namespace Browser {
 
-class BookmarksBarWidget final : public GUI::Widget {
+class BookmarksBarWidget final : public GUI::Widget
+    , private GUI::ModelClient {
     C_OBJECT(BookmarksBarWidget)
 public:
     static BookmarksBarWidget& the();
@@ -52,7 +54,7 @@ public:
 private:
     BookmarksBarWidget(const String&, bool enabled);
 
-    virtual void did_update_model();
+    virtual void on_model_update(unsigned) override;
     virtual void resize_event(GUI::ResizeEvent&) override;
 
     void update_content_size();
--- a/Libraries/LibGUI/FilePicker.h
+++ b/Libraries/LibGUI/FilePicker.h
@@ -31,10 +31,11 @@
 #include <LibCore/StandardPaths.h>
 #include <LibGUI/Dialog.h>
 #include <LibGUI/Image.h>
+#include <LibGUI/Model.h>
 
 namespace GUI {
 
-class FilePicker final : public Dialog {
+class FilePicker final : public Dialog, private ModelClient {
     C_OBJECT(FilePicker)
 public:
     enum class Mode {
@@ -55,6 +56,8 @@ private:
     void clear_preview();
     void on_file_return();
 
+    virtual void on_model_update(unsigned) override;
+
     FilePicker(Mode type = Mode::Open, const StringView& file_name = "Untitled", const StringView& path = Core::StandardPaths::home_directory(), Window* parent_window = nullptr);
 
     static String ok_button_name(Mode mode)
@@ -74,6 +77,7 @@ private:
     LexicalPath m_selected_file;
 
     RefPtr<TextBox> m_filename_textbox;
+    RefPtr<TextBox> m_location_textbox;
     RefPtr<Frame> m_preview_container;
     RefPtr<Image> m_preview_image;
     RefPtr<Label> m_preview_name_label;
--- a/Libraries/LibGUI/FilePicker.h
+++ b/Libraries/LibGUI/FilePicker.h
@@ -40,11 +40,21 @@ class FilePicker final : public Dialog, private ModelClient {
 public:
     enum class Mode {
         Open,
+        OpenMultiple,
         Save
     };
 
-    static Optional<String> get_open_filepath(const String& window_title = {});
-    static Optional<String> get_save_filepath(const String& title, const String& extension);
+    enum class Options : unsigned {
+        None = 0,
+        DisablePreview = (1 << 0)
+    };
+
+    static Optional<String> get_open_filepath(Options options)
+    {
+        return get_open_filepath({}, options);
+    }
+    static Optional<String> get_open_filepath(const String& window_title = {}, Options options = Options::None);
+    static Optional<String> get_save_filepath(const String& title, const String& extension, Options options = Options::None);
     static bool file_exists(const StringView& path);
 
     virtual ~FilePicker() override;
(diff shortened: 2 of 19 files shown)


Libraries/LibGUI/FilePicker.h
```diff
@@ -31,10 +31,11 @@
 #include <LibCore/StandardPaths.h>
 #include <LibGUI/Dialog.h>
 #include <LibGUI/Image.h>
+#include <LibGUI/Model.h>
 
 namespace GUI {
 
-class FilePicker final : public Dialog {
+class FilePicker final : public Dialog, private ModelClient {
```
Dexesttp, line 38:
```suggestion
class FilePicker final
    : public Dialog
    , private ModelClient {
```
I think this is usually formatted by `clang-format`, though I'm not sure. It's the Serenity style tho ^^

Applications/Browser/BookmarksBarWidget.h
```diff
@@ -27,11 +27,12 @@
 #pragma once
 
 #include <LibGUI/Forward.h>
+#include <LibGUI/Model.h>
 #include <LibGUI/Widget.h>
 
 namespace Browser {
 
-class BookmarksBarWidget final : public GUI::Widget {
+class BookmarksBarWidget final : public GUI::Widget, private GUI::ModelClient {
```
Dexesttp, line 35: Same here,
```suggestion
class BookmarksBarWidget final
    : public GUI::Widget
    , private GUI::ModelClient {
```

Libraries/LibGUI/ModelSelection.cpp
```diff
@@ -38,8 +38,14 @@ void ModelSelection::remove_matching(Function<bool(const ModelIndex&)> filter)
...
     }
-    for (auto& index : to_remove)
-        m_indexes.remove(index);
+    if (!to_remove.is_empty()) {
+        for (auto& index : to_remove)
+            m_indexes.remove(index);
+        if (!m_disable_notify)
+            notify_selection_changed();
+        else
+            m_notify_pending = true;
```
Dexesttp, line 47: This exact code is repeated six times. I think it would make more sense to move that logic inside the `notify_selection_changed` function (and expect that function to just set the pending flag if notifications are disabled).

Libraries/LibGUI/SortingProxyModel.h
```diff
@@ -30,7 +30,7 @@
 
 namespace GUI {
 
-class SortingProxyModel final : public Model {
+class SortingProxyModel final : public Model, private ModelClient {
```
Dexesttp, line 33: Same again ^^
```suggestion
class SortingProxyModel final
    : public Model
    , private ModelClient {
```

Libraries/LibGUI/SortingProxyModel.h
```diff
@@ -55,9 +55,12 @@ class SortingProxyModel final : public Model {
 private:
     explicit SortingProxyModel(NonnullRefPtr<Model>&&);
 
+    virtual void on_model_update(unsigned) override;
```
Dexesttp, line 58: It took a moment for me to understand that this takes flags, so it should probably be explicit
```suggestion
    virtual void on_model_update(unsigned flags) override;
```

Libraries/LibGUI/Model.h
```diff
@@ -44,6 +44,13 @@ enum class SortOrder {
     Descending
 };
 
+class ModelClient {
+public:
+    virtual ~ModelClient() { }
+
+    virtual void on_model_update(unsigned) = 0;
```
Dexesttp, line 51: Same here, possibly even more important since there's no reference implementation to directly check.
```suggestion
    virtual void on_model_update(unsigned flags) = 0;
```

Libraries/LibGUI/IconView.cpp
```diff
@@ -69,9 +73,75 @@ void IconView::resize_event(ResizeEvent& event)
...
+        }
+    }
+    if ((size_t)m_first_selected_hint >= new_item_count)
+        m_first_selected_hint = 0;
+    m_item_data_cache.resize(new_item_count);
+    for (size_t i = 0; i < items_to_invalidate; i++) {
+        auto& item_data = m_item_data_cache[i];
+        // TODO: It's unfortunate that we have no way to know whether any
+        // data actually changed, so we have to invalidate *everyone*
+        if (item_data.is_valid()/* && !model()->is_valid(item_data.index)*/)
```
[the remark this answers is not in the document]
tomuta: The TODO is related to the commented out code. Basically, it's just saying that if `Model::is_valid` could tell us whether the item itself needs to be refreshed then we didn't have to invalidate that cache item.

Libraries/LibGUI/IconView.cpp
```diff
@@ -210,17 +260,46 @@ bool IconView::update_rubber_banding(const Gfx::IntPoint& position)
...
+                toggle_selection(item_data);
             }
-        }
+            return IterationDecision::Continue;
+        });
+        // Now toggle all items that are in the new selected area, once only
+        for_each_item_intersecting_rects(select_area, [&](ItemData& item_data) -> IterationDecision {
+            if (!item_data.selection_toggled && !item_data.is_intersecting(prev_rect) && item_data.is_intersecting(rubber_band_rect)) {
+                item_data.selection_toggled = true;
+                toggle_selection(item_data);
```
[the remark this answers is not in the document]
tomuta: No, you want toggle_selection every time because when you add to the selection holding the Ctrl key, you want to be able to invert selected items as a way to deselect some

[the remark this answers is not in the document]
tomuta: Also, you need the `selection_toggled` flag because you only want to toggle each item once, because it's possible that you match the same item more than once during that loop. That's because the `Rect::hammer` may give you more than one rectangle that intersects with that item.

tomuta:
Good suggestions. Got them incorporated into the changes.

We really need a git pre-commit hook for `clang-format`... ;-)

alimpfard:
I guess pre-commit is no longer overkill, someone else wants it too #2365

tomuta:
I think the `SortingProxyModel` changes are in part wrong. Seems like there are a few cases where these changes cause problems. It looks like `FileSystemModel` is misbehaving, so I think I need to chase that bug down first and then undo some of the changes that are trying to compensate the bad behavior.

tomuta:
Ok so the `SortingProxyModel` fix ended up being more straight forward after understanding what exactly was going on with `FileSystemModel`. I pulled the lazy cache loading which caused other problems sometimes.

PR #2763, opened 2020-07-11
LibGUI: Fix keyboard shortcut conflicts in TreeView

thankyouverycool:
Changes the shortcut to expand and collapse subtrees from alt to ctrl+right/left arrows. Current shortcut conflicts with navigation controls in file manager and causes erratic behavior.

### DIFF
--- a/Libraries/LibGUI/TreeView.cpp
+++ b/Libraries/LibGUI/TreeView.cpp
@@ -110,13 +110,16 @@ void TreeView::doubleclick_event(MouseEvent& event)
 
 void TreeView::set_open_state_of_all_in_subtree(const ModelIndex& root, bool open)
 {
-    if (root.is_valid())
+    if (root.is_valid()) {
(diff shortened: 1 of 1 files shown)


Libraries/LibGUI/TreeView.cpp
```diff
@@ -110,13 +110,16 @@ void TreeView::doubleclick_event(MouseEvent& event)
...
 void TreeView::set_open_state_of_all_in_subtree(const ModelIndex& root, bool open)
 {
-    if (root.is_valid())
+    if (root.is_valid()) {
         ensure_metadata_for_index(root).open = open;
+        if (model()->row_count(root)) {
+            if (on_toggle)
+                on_toggle(root, open);
+        }
+    }
```
awesomekling, line 119: This is entirely unrelated to the other change, so please put it in a separate commit :)

thankyouverycool: Righto!

PR #2766, opened 2020-07-11
ProcFS: JSONify /proc/cpuinfo

linusg:
To be more in line with other parts of Serenity's procfs, the "key: value" format of `/proc/cpuinfo` was replaced with JSON, namely an array of objects (one for each core).

The available keys remain the same, though "features" has been changed from a space-separated string to an array of strings.

SystemMonitor has been updated to parse `/proc/cpuinfo` as JSON. CPU detection in neofetch is disabled for now until we can come up with a good JSON parsing tool for the shell.

image

### DIFF
--- a/Applications/SystemMonitor/ProcessModel.h
+++ b/Applications/SystemMonitor/ProcessModel.h
@@ -89,14 +89,12 @@ public:
     virtual GUI::Variant data(const GUI::ModelIndex&, Role = Role::Display) const override;
     virtual void update() override;
 
-    struct CpuInfo
-    {
+    struct CpuInfo {
         u32 id;
-        float total_cpu_percent{0.0};
-        HashMap<String, String> values;
-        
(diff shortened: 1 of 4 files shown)


Applications/SystemMonitor/ProcessModel.h
```diff
@@ -89,14 +89,12 @@ class ProcessModel final : public GUI::Model {
     virtual GUI::Variant data(const GUI::ModelIndex&, Role = Role::Display) const override;
     virtual void update() override;
 
-    struct CpuInfo
-    {
+    struct CpuInfo {
         u32 id;
-        float total_cpu_percent{0.0};
-        HashMap<String, String> values;
```
linusg, line 96: this wasn't being used anywhere, so I removed it for now - but since all the values are in a JSON object anyway, accessing them is now a lot easier once we need to do that. :^)

PR #2767, opened 2020-07-11
 Shell: Add a 'for' loop 

### LINKED ISSUE #2760: Shell: Needs more loops
After having used the shell on my host system for...basically forever, I have determined that loops are the most important next feature I want.
So let's discuss design!

Here are some initial ideas:
- Keep the bash-y syntax of `for <varname> in <expr>; block`
    - keep the block syntax: `do <stuff>... done`
    - use braces for block: `{ <stuff>... }`
- Borrow syntax from some existing language
    - js: `for (<var> in <expr>) block`
    - c++: `for (<var> : <expr>) block`
- Be more functional about it with `map`/`fold`
    - Syntax for blocks? semantics for passing blocks around?
- Invent something
    - ???

alimpfard:
Closes #2760.

The syntax I went with is:
- `for <expr> { sequence }`
- `for <ident> in <expr> { sequence }`

the braces are mandatory, and in the first one, the value is implicitly called `it`.
This PR also adds variable frames, because a proper language is a good language :^)

### DIFF
--- a/Shell/AST.cpp
+++ b/Shell/AST.cpp
@@ -732,6 +732,88 @@ Fd2FdRedirection::~Fd2FdRedirection()
 {
 }
 
+void ForLoop::dump(int level) const
+{
+    Node::dump(level);
+    print_indented(String::format("%s in\n", m_variable_name.characters()), level + 1);
+    m_iterated_expression->dump(level + 2);
+    print_indented("Running", level + 1);
+    if (m_block)
+        m_block->dump(level + 2);
+    else
+        print_indented("(null)", level + 2);
+}
+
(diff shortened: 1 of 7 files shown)


alimpfard:
And I forgot again...cc @bugaevc (and @awesomekling, you seemed to also have some input on this).

I would also like to incorporate the `map` idea, by allowing them to be specific 'forms', in that the parser treats them differently (but that's for another PR).

awesomekling:
@bugaevc any nitpicks here? ;)

bugaevc:
¯\\\_(ツ)\_/¯

I haven't looked too closely, but nothing looks immediately wrong

awesomekling:
Let's do it :)

PR #2773, opened 2020-07-12
Shell: Recursively resolve aliases

alimpfard:
Fun story: aliases were initially recursive, but I got rid of that to disallow an alias resolving to itself.
That's no longer an issue though :shrug:


awesomekling:
Thanks!

PR #2774, opened 2020-07-12
Userland: Add a basic version the tree command

S-D-R:
After watching @awesomekling's youtube videos about serenity for a while I got inspired to start contributing myself.

One of the first things I tried after building the system was the "tree" command, but sadly it was missing. So I thought why not add it myself. Obviously it's not compliant to the original, and it could probably be written better, but it's a start! :)

image

### DIFF
--- a/Userland/tree.cpp
+++ b/Userland/tree.cpp
@@ -1,0 +1,153 @@
+/*
+ * Copyright (c) 2020, Stijn De Ridder <stijn.deridder@hotmail.com>
+ * All rights reserved.
+ *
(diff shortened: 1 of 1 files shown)


awesomekling:
Hello, and thank you for working on this, looks like a nice addition! :)
Please observe the [contribution guidelines](https://github.com/SerenityOS/serenity/blob/master/CONTRIBUTING.md). Specifically, please squash your commits.

awesomekling:
v. nice :)

PR #2777, opened 2020-07-12
FontEditor: Don't add duplicate extension in Save As dialog

nico:


nico:
Actually, let's do this differently.

PR #2778, opened 2020-07-12
Make LexicalPath give "foo.bar" as title() for "foo.bar.txt", and use LexicalPath in FontEditor

nico:


nico:
A different, arguably nicer fix than #2777.

alimpfard:
:thinkst:
why not just `name()`?

PR #2781, opened 2020-07-13
Default Fonts: Tweak 'F' and 'W'

nico:
In all default fonts, make the lower bar of the F one pixel shorter to
match the middle bare of the E.

Make the W in CsillaThin a bit shorter on the sides and make it
go less high in the middle. This makes it look more like the W in
CsillaBold, makes the middle high spot in W match the height of
the same spot in X Y E F H. Making it shorter on the side makes
the letter look better when its next ot other full-width letters,
e.g. in "WWW".

Make the w in Katica10 match new new W in CsillaThin. The bold
letters already match, and in general it looks like Csilla is
a monospace version of Katica.


nico:
Before: https://imgur.com/a/HMyZNsm
After: https://imgur.com/a/wxWZcK7

Can't argue about taste, but to me it's a slight improvement.

It's a bit strange that the bold N has its blank line to its left, while the non-bold one has it to the right. That's for another day.

awesomekling:
Great tweaks :)

PR #2782, opened 2020-07-13
Shell: Resolve aliases in builtin_time

alimpfard:
This moves out a bunch of execution stuff to Shell and makes `builtin_time` resolve aliases.

It should be noted that `time` will respect the `Background` nodes and not wait for them, this is not necessarily correct according to other shells.
According to bash and zsh:
```sh
alias foo='ls&'
time foo .
# Tries to execute two commands, 'time ls&' and '.'
```

since we treat aliases as "splice into command sequence" rather than "text replace", we would see this behaviour:
```sh
alias foo='ls&'
time foo .
# times the execution of 'ls .&'
```

I personally do not see this as being wrong, but other opinions would be helpful.

### DIFF
--- a/Shell/AST.cpp
+++ b/Shell/AST.cpp
@@ -742,79 +742,7 @@ RefPtr<Value> Execute::run(RefPtr<Shell> shell)
     if (m_command->would_execute())
         return m_command->run(shell);
 
-    auto initial_commands = m_command->run(shell)->resolve_as_commands(shell);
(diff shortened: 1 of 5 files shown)


alimpfard:
dammit, mess up the commits again

alimpfard:
I somehow managed to confuse github into ignoring anything pushed to the branch...closing this and opening another pr.

PR #2783, opened 2020-07-13
Shell: Resolve aliases in builtin_time

alimpfard:
This moves out a bunch of execution stuff to Shell and makes builtin_time resolve aliases.

It should be noted that time will respect the Background nodes and not wait for them, this is not necessarily correct according to other shells.
According to bash and zsh:

```sh
alias foo='ls&'
time foo .
# Tries to execute two commands, 'time ls&' and '.'
```
since we treat aliases as "splice into command sequence" rather than "text replace", we would see this behaviour:

```sh
alias foo='ls&'
time foo .
# times the execution of 'ls .&'
```

I personally do not see this as being wrong, but other opinions would be helpful.

### DIFF
--- a/Shell/AST.cpp
+++ b/Shell/AST.cpp
@@ -742,79 +742,7 @@ RefPtr<Value> Execute::run(RefPtr<Shell> shell)
     if (m_command->would_execute())
         return m_command->run(shell);
 
-    auto initial_commands = m_command->run(shell)->resolve_as_commands(shell);
-    decltype(initial_commands) commands;
-
-    Function<void(Command&)> resolve_aliases_and_append = [&](auto& command) {
-        if (!command.argv.is_empty()) {
-            auto alias = shell->resolve_alias(command.argv[0]);
-            if (!alias.is_null()) {
-                auto argv0 = command.argv.take_first();
-                auto subcommand_ast = Parser { alias }.parse();
-                if (subcommand_ast) {
-                    while (subcommand_ast->is_execute()) {
-                        auto* ast = static_cast<Execute*>(subcommand_ast.ptr());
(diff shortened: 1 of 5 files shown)


awesomekling:
This seems so awkward. Shouldn't we be resolving aliases at another stage?
It doesn't seem right that builtins have to worry about this.

alimpfard:
There's not much else we can do, either we try to resolve every single argument as an alias (sounds silly) or we only resolve the first one.
If we do the first, bad behaviour
If we do the second, if something wants to treat an argument as a command, it has to worry about the aliases.

Any other option I'm just not seeing?

awesomekling:
> Any other option I'm just not seeing?

Yes, `time` could pop the `time` token from the command string and then pass the rest to some kind of "run command" machine (while timing how long that machine takes, ofc.)

Said machine would then resolve the first token (in each statement in case there are multiple) against the alias list.

Does that not make sense? Maybe I'm missing something.

alimpfard:
builtins are not recognised until a command is being executed, so by the time `time` gets the arguments, they're in a fully processed form (i.e. just a bunch of arguments in `argv`)

Doing what you're suggesting would require the parser to recognise the builtins, but only as the first element in a command list, and then completely bypass the normal evaluation/execution stuff and call directly into the builtin.
currently the parser is not equipped to do this sort of thing, but I am planning to have 'special forms' (as I've mentioned in the `for` PR) which require an AST transformation. builtins could also be such forms where they take the AST instead of the arguments.

though I personally prefer keeping the builtins as close to a normal utility program as possible.

awesomekling:
Yeah yeah okay fine. :)

awesomekling:
Thanks for fixing this btw!

PR #2784, opened 2020-07-13
unzip: load entire file into memory (or fail if too big) / rm magic numbers

alimpfard:
This PR makes unzip map the entire file instead of repeatedly seek()/read()'ing, which makes it actually possible to unpack relatively large zip files.
A limit for file size is added and set to 32MB for s a f e t y.

Also fixes directory creation.
Also replaces some magic offsets with enum values.

(cc @zlotny)


awesomekling:
:ninja:

PR #2786, opened 2020-07-13
LibLine: Add Ctrl-N/P as history next/previous shortcuts

nico:


Libraries/LibLine/Editor.cpp
```diff
@@ -559,14 +562,11 @@ void Editor::handle_read_event()
             }
             case 'D': // left
 		do_cursor_left(ctrl_held ? Word : Character);
-                m_inline_search_cursor = m_cursor;
```
alimpfard, line 562: So this would always search for the entire phrase now?
I guess we can just delete from the end to get the inline search phrase to be right, but...couldn't we have made it so C-{b,f} update this too?
```sh
> f|oo
   ^ cursor here, search phrase is "f"
> fo|o
    ^ cursor here, search phrase is "fo"
```

alimpfard:
Aside from that, very much in favour of this change.

Libraries/LibLine/Editor.cpp
nico: No, I just moved this into do_cursor_left(), so that it also happens for ctrl-b.

Libraries/LibLine/Editor.cpp
```diff
@@ -431,39 +431,78 @@ void Editor::handle_read_event()
...
+                    if (skipped_at_least_one_character && isspace(m_buffer[m_cursor - 1])) // stop *after* a space, but only if it changes the position
+                        break;
+                    skipped_at_least_one_character = true;
+                    --m_cursor;
+                }
+            } else {
+                --m_cursor;
+            }
+        }
+        m_inline_search_cursor = m_cursor;
```
nico, line 449: See here.

Libraries/LibLine/Editor.cpp
alimpfard: Ah. understood.
the commit message could perhaps be a bit less..."remove"y and more "move"y?

nico: I was certain I had written "Move...". Not sure how I didn't. Fixed now, thanks for catching :)

PR #2789, opened 2020-07-13
Csilla: Make 'W' and 'w' even thinner

nico:
'W' doesn't have to go up to the edges which makes 'WWW'
look better, and it imho looks fine in other contexts too.
Update 'w' to match.

Don't change Katica since it has enough room for the current W.

Before: https://imgur.com/a/wxWZcK7
After: https://imgur.com/a/T2lVL67


awesomekling:
Looks p good :)

PR #2792, opened 2020-07-13
Add a Markdown fuzzer and fix a bug found by it.

nico:
Also add some docs on how to build and run fuzzers.


Meta/Lagom/ReadMe.md
```diff
@@ -7,3 +7,14 @@ The Serenity C++ library, for other Operating Systems.
 If you want to bring the comfortable Serenity classes with you to another system, look no further. This is basically a "port" of the `AK` and `LibCore` libraries to generic \*nix systems.
 
 *Lagom* is a Swedish word that means "just the right amount." ([Wikipedia](https://en.wikipedia.org/wiki/Lagom))
+
+## Fuzzing
+
+Lagom can be used to fuzz parts of SerenityOS's code base on Linux. This requires buildling with clang, so it's convenient to use a different build directory for that. Run CMake like so:
```
bugaevc, line 13: Why on Linux?

On any Unix-like OS that supports LLVM and sanitizers.

Also Clang is spelled with a capital C :)

nico: I've only tried it on Linux. Besides, serenity currently doesn't build on macOS and likely will never build on Windows, and everybody knows that Linux, macOS, and Windows are the only operating systems that exist.

I work on clang and I call it "clang". I can wrap it in backticks if it makes you sleep better.

Pagghiu: Just as FYI, serenity just recently gained back the ability to build on macOS (#2750)
I think it's possible to build it on windows if somebody wants to do it :)

awesomekling: I see you also work on serenity and call it "serenity" :smile:

awesomekling:
Cool!

Meta/Lagom/ReadMe.md
nico: I used to call it "synergy" no too long ago!

Issue #2787, opened 2020-07-13
FileManager crashes when triggering context menu in list view


nico:
1. Switch to list view, e.g. by hitting Ctrl-L
2. Right-click anything

Expected: Context menu
Actual: Null-ptr deref:

```
[FileManager(39:39)]: CPU[0] NP(error) fault at invalid address V0x0000000c
[FileManager(39:39)]: Unrecoverable page fault, read from address V0x0000000c
[FileManager(39:39)]: Note: Address V0x0000000c looks like a possible nullptr dereference
[FileManager(39:39)]: CRASH: CPU #0 Page Fault. Ring 3.
[FileManager(39:39)]: exception code: 0004 (isr: 0000
[FileManager(39:39)]:   pc=001b:0804d1eb flags=0206
[FileManager(39:39)]:  stk=0023:019024f0
[FileManager(39:39)]:   ds=0023 es=0023 fs=0023 gs=002b
[FileManager(39:39)]: eax=00000000 ebx=019312b6 ecx=00000000 edx=01902580
[FileManager(39:39)]: ebp=01902528 esp=c4493fe8 esi=0190250c edi=019025f8
[FileManager(39:39)]: cr0=80010013 cr2=0000000c cr3=0277e000 cr4=00100ee4
[FileManager(39:39)]: code: 8b 40 0c 66 25 00 f0 66
[FileManager(39:39)]: 0x0804d1eb  AK::Function<void (GUI::AbstractView const&, GUI::ModelIndex const&, GUI::ContextMenuEvent const&)>::CallableWrapper<run_in_windowed_mode(AK::RefPtr<Core::ConfigFile>, AK::String)::{lambda(GUI::AbstractView const&, GUI::ModelIndex const&, GUI::ContextMenuEvent const&)#32}>::call(GUI::AbstractView const&, GUI::ModelIndex const&, GUI::ContextMenuEvent const&) const +67

[FileManager(39:39)]: 0xc01656a6  Kernel::Process::crash(int, unsigned int, bool) +1010
[FileManager(39:39)]: 0xc0120937  illegal_instruction_handler +0
[FileManager(39:39)]: 0xc0121f42  page_fault_handler +945
[FileManager(39:39)]: 0xc011e1cf  page_fault_asm_entry +38
[...]

PR #2793, opened 2020-07-13
LibGUI: Verify internal_data exists prior return

kevmeyer:
Fixes https://github.com/SerenityOS/serenity/issues/2787

### DIFF
--- a/Libraries/LibGUI/FileSystemModel.cpp
+++ b/Libraries/LibGUI/FileSystemModel.cpp
@@ -332,6 +332,8 @@ const FileSystemModel::Node& FileSystemModel::node(const ModelIndex& index) cons
 {
     if (!index.is_valid())
         return *m_root;
+
+    ASSERT(index.internal_data());
     return *(Node*)index.internal_data();
 }
 
--- a/Libraries/LibGUI/SortingProxyModel.cpp
+++ b/Libraries/LibGUI/SortingProxyModel.cpp
@@ -79,6 +79,16 @@ String SortingProxyModel::column_name(int index) const
     return target().column_name(index);
 }
 
+ModelIndex SortingProxyModel::index(int row, int column, const ModelIndex& parent) const
+{
+    if (static_cast<size_t>(row) >= m_row_mappings.size() || column >= column_count())
+        return {};
+    auto mapped_index = target().index(m_row_mappings[row], column, parent);
+    if (!mapped_index.is_valid())
+        return {};
+    return create_index(row, column, mapped_index.internal_data());
+}
+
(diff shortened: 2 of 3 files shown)


bugaevc:
No; a valid index with a null internal data is always (in case of the file system model) a mistake of whoever's supplying the index — clearly they've corrupted the index somehow, because the only two ways to create a valid model index are:

* Default-constructing an (invalid) index with `{}`
* Asking the model itself to create a new index into it (with `index()` or `parent_index()`), and the file system model never gives out valid indexes with null data

bugaevc:
Also I believe the table/list view in FileManager (and in the file dialog) are the only times the sorting proxy model is used on a tree-like model (correct me if I'm wrong! 🙂) — so it's important that we find & fix the root cause here, and not just paper over it like this.

kevmeyer:
> No; a valid index with a null internal data is always (in case of the file system model) a mistake of whoever's supplying the index — clearly they've corrupted the index somehow, because the only two ways to create a valid model index are:
> 
> * Default-constructing an (invalid) index with `{}`
> * Asking the model itself to create a new index into it (with `index()` or `parent_index()`), and the file system model never gives out valid indexes with null data

Initially I also had this intention, but only until I saw this default `nullptr` argument here, so I assumed that this might still be a valid case:
https://github.com/SerenityOS/serenity/blob/master/Libraries/LibGUI/Model.h#L114

I found something in the Proxy, where the internal_data got lost.

bugaevc:
> Initially I also had this intention, but only until I saw this default `nullptr` argument here, so I assumed that this might still be a valid case

`create_index()` is a protected method (not a public API for the model users!) to aid the model itself in creating indexes; it's often reasonable to create indexes with nullptr internal data, so that's why it's the default. But it's up to the model to do decide whether it uses internal data or not; and the file system model in particular always creates indexes with non-null internal data.

> I found something in the Proxy, where the internal_data got lost.

It doesn't look like there's a problem there either. Here, we're creating an index into the *proxy* model, so it's up to the proxy model to decide how it uses internal data — and it decides that it doesn't, and that it always sets internal data to null. So it's fine not to override that method.

The important part is for the proxy model to create indexes into the *target model* properly — that is, by calling target's `index()`, and not just creating a row/column index itself. And the proxy model seems to do that correctly:

https://github.com/SerenityOS/serenity/blob/7739497e34f7617cb2f6d838cc3710596b91d13e/Libraries/LibGUI/SortingProxyModel.cpp#L74

bugaevc:
Submitted a proper fix in https://github.com/SerenityOS/serenity/pull/2809

awesomekling:
Superseded by #2809

PR #2794, opened 2020-07-14
LibGUI: ComboBox improvements

thankyouverycool:
A few new visual effects and keyboard controls for combo boxes, making them look and operate more like classic drop-down lists when read-only. The selected/hovered row is now highlighted and the list closes when the item is activated or the widget loses focus.
combobox

### DIFF
--- a/Libraries/LibGUI/TextEditor.h
+++ b/Libraries/LibGUI/TextEditor.h
@@ -127,8 +127,13 @@ public:
     void redo() { document().redo(); }
 
     Function<void()> on_change;
+    Function<void()> on_mousedown;
     Function<void()> on_return_pressed;
     Function<void()> on_escape_pressed;
+    Function<void()> on_up_pressed;
+    Function<void()> on_down_pressed;
+    Function<void()> on_pageup_pressed;
+    Function<void()> on_pagedown_pressed;
 
     Action& undo_action() { return *m_undo_action; }
     Action& redo_action() { return *m_redo_action; }
--- a/Libraries/LibGUI/TextEditor.h
+++ b/Libraries/LibGUI/TextEditor.h
@@ -46,6 +46,13 @@ public:
         MultiLine,
         SingleLine
     };
+
+    enum Mode {
+        Editable,
+        ReadOnly,
+        DisplayOnly
+    };
+
     virtual ~TextEditor() override;
 
     const TextDocument& document() const { return *m_document; }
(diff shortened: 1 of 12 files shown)


Libraries/LibGUI/TextEditor.h
```diff
@@ -248,6 +254,9 @@ class TextEditor
     bool m_automatic_indentation_enabled { false };
     bool m_line_wrapping_enabled { false };
     bool m_readonly { false };
+    bool m_displayonly { false };
```
awesomekling, line 263: Since these two can be combined in invalid ways, it would be nicer to use an enum for the mode. :)

tomuta:
Is there a way to change it so that the application window doesn't appear to lose focus while the drop down is active? Similar to menus?

awesomekling:
> Is there a way to change it so that the application window doesn't appear to lose focus while the drop down is active? Similar to menus?

There's currently no way to do that, buuut...

Windows can be tied to a "parent window" via the `parent_window_id` argument to the `WindowServer::CreateWindow` IPC message.

We could add a new window mode that means it shares focus with its parent. The main awkwardness here is that message boxes tend to be parented to the window that spawn them, so this would make message boxes behave differently.

Someone could definitely figure out a nice solution here, I'm sure :)

tomuta:
> > Is there a way to change it so that the application window doesn't appear to lose focus while the drop down is active? Similar to menus?
> 
> There's currently no way to do that, buuut...
> 
> Windows can be tied to a "parent window" via the `parent_window_id` argument to the `WindowServer::CreateWindow` IPC message.
> 
> We could add a new window mode that means it shares focus with its parent. The main awkwardness here is that message boxes tend to be parented to the window that spawn them, so this would make message boxes behave differently.
> 
> Someone could definitely figure out a nice solution here, I'm sure :)

I came up with a nice solution that works really well even in combination with this pull request. I'll create a merge request for that shortly.

nico:
Great change, thanks!

I noticed that the system menu supports mac-style press-button-mouse-over-item-release-button to pick items with a single click. Combo boxes don't: Doing this over the text part opens the popup on press but release doesn't select, and doing this over the button part only opens the menu on release, not on press. In case you're looking for other tweaks to hack on :)

tomuta:
> Great change, thanks!
> 
> I noticed that the system menu supports mac-style press-button-mouse-over-item-release-button to pick items with a single click. Combo boxes don't: Doing this over the text part opens the popup on press but release doesn't select, and doing this over the button part only opens the menu on release, not on press. In case you're looking for other tweaks to hack on :)

Doesn't Windows also do that? Either way, I have some ideas on how to make that work, I'll take a look when I have some more time again.

PR #2795, opened 2020-07-14
FileManager: Add "Open with" menu

tomuta:
Screenshot from 2020-07-13 19-00-40


tomuta:
Fixed double-clicking binaries to match the default action rather than opening them with the default handler:
Screenshot from 2020-07-14 09-35-50

PR #2796, opened 2020-07-14
unzip: Correct compression method offset value

alimpfard:
`+10` is the modification time offset, and through sheer chance, minizip sets that to zero...
and I need to look at diagrams more carefully next time.


awesomekling:
Oopsie!

PR #2798, opened 2020-07-14
LibGfx: Minor IntRect::shatter optimization

tomuta:
All the shards always intersect unless they're empty.
So rather than checking for intersection, just check
whether they're empty or not.


awesomekling:
Does this not apply to `FloatRect::shatter()`?

PR #2799, opened 2020-07-14
Shell: Explicitly declare 'environ' to make macOS Lagom build happy

alimpfard:
~It is a mystery how this ever compiled successfully...~
(`environ` is not declared in Builtin.cpp when compiling ~with clang~ Lagom on macOS)

### DIFF
--- a/Shell/Builtin.cpp
+++ b/Shell/Builtin.cpp
@@ -31,7 +31,9 @@
 #include <inttypes.h>
 #include <signal.h>
 #include <sys/wait.h>
+#include <unistd.h>
 
+extern char** environ;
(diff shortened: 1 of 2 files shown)


tomuta:
```
../Libraries/LibC/unistd.h:55:15: note: previous declaration as 'char** environ'
   55 | extern char** environ;
```

alimpfard:
Thanks for clearing that up~

nico:
I believe this is the correct fix.

awesomekling:
Indeed, seems like Dr. POSIX only specifies that you declare environ as `extern char** environ;`

PR #2800, opened 2020-07-14
LibC+Kernel: Start implementing sysconf

nico:
For now, only the non-standard _SC_NPROCESSORS_CONF and
_SC_NPROCESSORS_ONLN are implemented.

Use them to make ninja pick a better default -j value.
While here, make the ninja package script not fail if
no other port has been built yet.


awesomekling:
So cool when we can remove port patches :)

PR #2802, opened 2020-07-14
LibJS: Add Number.prototype.toString

mattco98:
Unfortunately, this isn't necessarily spec-compliant. This is because the [spec for this method](https://www.ecma-international.org/ecma-262/11.0/index.html#sec-numeric-types-number-tostring) isn't exactly... straightforward. I seemed to have figured out what it was saying after a few minutes with a paper and pencil, however when I compared my results to V8, they didn't quite match up.

For integer numbers, it's easy. Floating point numbers are where it gets a bit hairy. The only difficulty is determining how many digits to print. My interpretation of step 5 of the spec is basically "print as many characters to the right of the decimal point as possible given the floating point precision constraints". So, my naive way to find how many characters of precision a given radix can have is `floor(52 * ln(2) / ln(radix))`, which I placed in a lookup table. 

Here's a comparison of my `toString` vs V8's `toString`:

preview

My implementation tends to print more characters, however I have found that in certain cases, V8 prints _more_ characters than my implementation (in other words, more characters than are actually accurate given a standard floating point number).

If anyone has a better interpretation of the spec, or a way to improve this algorithm, let me know :) otherwise, it's pretty good for now. Note that I don't have tests for decimal numbers since they're not 100% correct.

### DIFF
--- a/Libraries/LibJS/Runtime/ErrorTypes.h
+++ b/Libraries/LibJS/Runtime/ErrorTypes.h
@@ -46,6 +46,7 @@
     M(InstanceOfOperatorBadPrototype, "Prototype property of %s is not an object")                     \
     M(InvalidAssignToConst, "Invalid assignment to const variable")                                    \
     M(InvalidLeftHandAssignment, "Invalid left-hand side in assignment")                               \
+    M(InvalidRadix, "Radix must be an integer no less than 2, and no greater than 36")                 \
     M(IsNotA, "%s is not a %s")                                                                        \
     M(IsNotAEvaluatedFrom, "%s is not a %s (evaluated from '%s')")                                     \
     M(JsonBigInt, "Cannot serialize BigInt value to JSON")                                             \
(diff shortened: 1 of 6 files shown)


Libraries/LibJS/Runtime/ErrorTypes.h
```diff
@@ -60,6 +61,7 @@
     M(NotASymbol, "%s is not a symbol")                                                                \
     M(NotIterable, "%s is not iterable")                                                               \
     M(NonExtensibleDefine, "Cannot define property %s on non-extensible object")                       \
+    M(NumberIncomptableThis, "Number.prototype.%s method called with incompatible this target")        \
```
awesomekling, line 64: Typo here, `NumberIncomptable` => `NumberIncompatible`

alimpfard:
_/me cries in excessive `Vector` copy_

awesomekling:
Yeh it could do with less `Vector` copying. At the same time, it's not the end of the world.
If it turns out to be a bottleneck one day, it's an easy fix :)

mattco98:
Yeah I wasn't really sure how else to do it, since we don't really know the number of characters it'll produce in advance. Like I said though, I'm all ears for algorithm improvements!

PR #2805, opened 2020-07-15
WindowServer: Add accessory windows

tomuta:
Accessory windows are windows that, when activated, will activate
their parent and bring all other accessory windows of that parent
to the front of the window stack. Accessory windows can only be
active input windows. The accessory window's parent is always the
active window regardless of whether it is also the active input
window.

In order to route input correctly, input is now sent to the active
input window, which can be any accessory window or their parent,
or any regular window.

Screenshot from 2020-07-14 20-33-49

This allows special accessory windows, such as a ComboBox dropdown, to receive input without changing the active window. Other types of accessory windows could be toolboxes, activating those would activate their parent window, bring all the other accessory windows of that window to the front as well, and set the active input window to that accessory window.

This works nicely in combination with #2794


awesomekling:
Great solution! Unfortunately merging combobox PR has introduced some conflicts here, sorry about that.

tomuta:
> Great solution! Unfortunately merging combobox PR has introduced some conflicts here, sorry about that.

I knew about it and expected having to rebase it ;-) Should be good to merge now.

PR #2807, opened 2020-07-15
BuildInstructions: correct build dir

diversys:


diversys:
Actually, reading Toolchain/BuildIt.sh I see that `BUILD=$(realpath "$DIR/../Build")`. 
I'm not sure why I end up with `build` dir on macOS.

PR #2809, opened 2020-07-15
Fix sorting proxy model

### LINKED ISSUE #1440: FileManager: Crash when right clicking in list view
When right clicking on any item in list view, the FileManager crashes due to a pagefault.
I've tried to fix it myself, but after tracing it down to a specific line of code, I'm at a loss.

On line 571 in FileManager's main.ccp, in the on_context_menu_request lambda

```
    directory_view.on_context_menu_request = [&](const GUI::AbstractView&, const GUI::ModelIndex& index, const GUI::ContextMenuEvent& event) {
        if (index.is_valid()) {
            auto& node = directory_view.model().node(index);

            if (node.is_directory())
                directory_context_menu->popup(event.screen_position());
            else
                file_context_menu->popup(event.screen_position());
        } else {
            directory_view_context_menu->popup(event.screen_position());
        }
    };
```
It seems that while using TableView, the index's internal_data will always be nullptr, which later propagates through directory_view.model().node() causing node to be null as well and the line if (node.is_directory()) dereferences it causing the segfault.

That's as far as i got though, i can't really tell where this nullptr originally comes from.

bugaevc:
Fix forgetting to map sorting proxy model indexes

Fixes #1440 and #2787 (which are duplicates) :smile:

(Also throwing in `virt$pipe()` for UE)

### DIFF
--- a/Applications/FileManager/DirectoryView.cpp
+++ b/Applications/FileManager/DirectoryView.cpp
@@ -166,8 +166,7 @@ DirectoryView::DirectoryView()
         handle_activation(index);
     };
     m_table_view->on_activation = [&](auto& index) {
-        auto& filter_model = (GUI::SortingProxyModel&)*m_table_view->model();
-        handle_activation(filter_model.map_to_target(index));
+        handle_activation(map_table_view_index(index));
     };
 
     m_table_view->on_selection_change = [this] {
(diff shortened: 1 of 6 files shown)


awesomekling:
Dude. Nice!

PR #2810, opened 2020-07-15
FileManager: Make Ctrl-L focus location bar, and change shortcuts for…

nico:
… changing view

Ctrl-L focuses the location bar in all file managers I know of,
and does so in SerenityOS's browser too. It should work in
SerenityOS's file manager as well.

Unfortunately, Ctrl-L was already "View List", so change the
shortcuts of all the view modes to Ctrl-1/2/3 which is what
several other file managers use. (I tried Ctrl-Shift-1/2/3
which is what Windows Explorer uses after 8.1, but it didn't
Just Work, and Ctrl-1/2/3 are currently free anyways. If
we ever want to use them for tabs or whatever, we can
change the view shortcuts then.)


awesomekling:
A most reasonable change!

Issue #2691, opened 2020-07-04
LibCrypto: Sporadic test failure


alimpfard:
`test-crypto` randomly fails sometimes.
One incident of `RSA PKCS#1 1.5 EME` failing has been seen, but I've been unable to reproduce it after that.

alimpfard:
Failure source confirmed as `RSA PKCS#1 1.5 EME` (in https://travis-ci.com/github/SerenityOS/serenity/builds/174626346#L13591).

_investigations shall now begin_

alimpfard:
https://travis-ci.com/github/SerenityOS/serenity/builds/175737954 has failed in the Signed BigInt set
I am quite confused as that bit is practically the same as the unsigned bigint...

alimpfard:
Whatever it is, is probably due to intolerance of some specific pattern in the input.
I suspect zeros at the beginning (for the bigint)

alimpfard:
This is very likely due to this FIXME:
https://github.com/SerenityOS/serenity/blob/ed327e7feb36d415281ad022d6f99a9d15539e6a/Libraries/LibCrypto/PK/RSA.cpp#L129-L131

as soon as 10,000 tests pass (if they do), I'll submit a patch.

PR #2815, opened 2020-07-15
WindowServer: Don't make window active by hovering over the icon

tomuta:


tomuta:
CI failed due to #2691

awesomekling:
Whoa, that's goofy :)

PR #2816, opened 2020-07-15
WindowServer: Fixes for modal windows

tomuta:
This fixes a few problems with modal windows:

* If any child window, or any child window further down the
  tree is considered modal, then all windows in that chain
  are modal.
* When trying to activate a window blocked by a modal child
  bring the entire stack of modal windows to the front and
  activate the modal window.
* A window is modal if it has a parent and it's flagged as
  modal, regardless of whether the ClientConnection has
  created modal windows.

This technically supports diverging modal window trees as well,
where two modal windows share the same parent, allowing both to
be activated (including for input) but not the parent. And it
should also support modal window stacks of arbitrary depth.


awesomekling:
More solid improvements. :)

PR #2821, opened 2020-07-16
Userland: Add a "seq" utility

nico:
It's useful with the newly added for loops in Shell.

SerenityOS's printf() doesn't seem to print decimals for
doubles correctly, so this doesn't print the right output for
fractional numbers yet.

### DIFF
--- a/Userland/seq.cpp
+++ b/Userland/seq.cpp
@@ -1,0 +1,108 @@
+/*
+ * Copyright (c) 2020, Nico Weber <thakis@chromium.org>
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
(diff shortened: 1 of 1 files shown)


Userland/seq.cpp
```diff
@@ -0,0 +1,108 @@
...
+        return 1;
+    }
+    if (unveil(nullptr, nullptr) < 0) {
+        perror("unveil");
+        return 1;
+    }
+
+    double start = 1, step = 1, end = 1;
+    int number_of_start_decimals = 0, number_of_step_decimals = 0, number_of_end_decimals = 0;
+    switch (argc) {
```
nico, line 61: (ArgsParser doens't do doubles, and it can't handle positional arguments that switch role depending on the number of arguments, as far as I can tell.)

Userland/seq.cpp
```diff
@@ -0,0 +1,108 @@
...
+        exit(1);
+    }
+    if (char* dot = strchr(d_string, '.'))
+        *number_of_decimals = strlen(dot + 1);
+    else
+        *number_of_decimals = 0;
+    return d;
+}
+
+int main(int argc, const char* argv[]) {
```
awesomekling, line 49: Hello! Please use `Clang-format` :^)

nico: I try, I try...tried harder.

Userland/seq.cpp
bugaevc: `ArgParser` surely does do doubles, and any type you like — you just have to use the longer / more verbose API. The APIs we normally use (for integers, strings, booleans and lists of strings), are just *shortcuts* for the generic API, exactly because they're used a lot.

https://github.com/SerenityOS/serenity/blob/9545ed1c216314c0bc1cb99fade62522cf0381f1/Libraries/LibCore/ArgsParser.cpp#L311-L325

(And since then, the shortcut API for `double` has been implemented, too: https://github.com/SerenityOS/serenity/pull/2867)

It can, to a certain extent, handle positional arguments that may be placed in different places. Namely, it supports optional arguments (of course), and has some logic for distributing available argv among the registered arguments:

https://github.com/SerenityOS/serenity/blob/9545ed1c216314c0bc1cb99fade62522cf0381f1/Libraries/LibCore/ArgsParser.cpp#L129-L138

That's not super featureful, and it lacks support for the case where a command has several different invocation syntaxes — but for `seq` it should work fine to just represent `first` and `increment` as optional arguments.

P.S. I'm only finding out about this comment now, several months later, as apparently there's now [a PR to add explicit usage message](https://github.com/SerenityOS/serenity/pull/3749). If only you would cc me on this PR or this comment, I could have helped you earlier, back in July.

PR #2823, opened 2020-07-16
WindowServer: Fix picking new active window after destroy

tomuta:
We need to mark windows as destroyed and not consider them
when picking a new active window. Fixes lost focus after
closing some windows.


tomuta:
Since this has a dependency on #2822, combined them into one merge request.

PR #2824, opened 2020-07-16
Shell: Start writing a man page 

alimpfard:
This adds two man pages describing the parts of the shell that have been so far finalised.

@awesomekling you asked for this.
@nico you also indirectly asked for this.
also cc @bugaevc, please implement nested lists in markdown :^)

(draft until we find a way to make the manpages look good on serenity)

### DIFF
--- a/Base/usr/share/man/man1/Shell.md
+++ b/Base/usr/share/man/man1/Shell.md
@@ -1,0 +1,40 @@
+## Name
+
+`Shell` - Shell, the command language interpreter
+
+## Synopsis
+
+```**sh
+$ Shell [--skip-shellrc]
+$ Shell [--skip-shellrc] command_file [arguments...]
+$ Shell [--skip-shellrc] -c command_string [arguments...]
+```
+
+## Description
+
+The `Shell` utility is a command language interpreter, which reads commands from either a command string, a specified file, or the standard input.
+The command language shall be described in [`Shell`(5)](../man5/Shell.md), _The Shell Command Language_.
+
+Any extra arguments passed into `arguments` are placed in the local variable `$ARGV` and can also be accessed through the special variable `$*`.
+
+**NOTE**:
+
+The `Shell` utility does not promise POSIX `sh` interoperability.
+
+## Options
+
+* `-c`, `--command-string`: Executes the given string as a command and exits
+* `--skip-shellrc`: Skips running the initialisation file (at `~/.shellrc`)
+
+## Examples
+
+```sh
+# Start an interactive REPL, ignoring the shellrc
+Shell --skip-shellrc
+
+# Execute a given string
+Shell -c 'rm foo*'
+
+# Execute the contents of a file with some arguments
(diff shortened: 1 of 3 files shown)


Base/usr/share/man/man1/sh-language.md
```diff
@@ -0,0 +1,133 @@
+## Name
+
+The Shell Command Language
+
+## Introduction
+
+The shell operates according to the following general steps:
```
bugaevc, line 7: How is this a (1) page? Sounds like `sh`(5) to me...

alimpfard: A...file format?
I put it in (1) because I had zero ideas where it would go, so I decided to plop it next to the shell manpage itself.

I guess I can see that...sure, why not.

alimpfard:
I'm gonna split the '&&' fix up, this PR is probably not the right place for it anyway (am I learning how to PR yet?)

Base/usr/share/man/man1/sh.md
```diff
@@ -0,0 +1,34 @@
+## Name
+
+`sh` - shell, the command language interpreter
```
awesomekling, line 3: The real name of the shell is `Shell`.
`sh` is a POSIX program that you've elected not to pursue compatibility with.
That's fine, but let's not pretend to be an `sh` since that will break the vast majority of scripts that may try to run with `#!/bin/sh`

alimpfard: So what would you suggest we do (setting the name in the manpage aside)?
make Shell run `/bin/bash` if invoked as `sh`?

awesomekling: I think we should probably stop calling it `/bin/sh` altogether and just be `/bin/Shell`.

alimpfard: That has quite a few implications, we'd need to make `system()` and co use `/bin/Shell` for one (since now there's no `/bin/sh`).
Perhaps we could add a minimal sh interpreter and use that if the shell is invoked as `sh`?

awesomekling: Yes, that would also be fine. My point is just that a non-`sh` shell should not pretend to be a `sh` :)

nico: Re `sh` discussions: I'll point out that ninja uses `/bin/sh -c ...` to spawn subprocesses, and it assumes that that does something sensible (which it currently mostly does). So there's some value in something sh-like existing at /bin/sh.

alimpfard: I'm gonna go ahead and say that this 'sh' issue is outside the scope of _this_ PR, so I'll refer to the Shell as `Shell` in the manpages, and we'll see where we go with `/bin/sh` (perhaps as a separate issue).

awesomekling: That's totally fine. And I agree with @nico there.

PR #2826, opened 2020-07-16
Userland: Add implementation of sponge

matthewgraham1:
Sponge is a command from the "moreutils" collection that soaks up all
standard input and outputs it to the desired file. This is a rewrite for
SerenityOS since it's such a simple and useful program. I named it 
"spng" because it is a lot more pleasant to type.

### DIFF
--- a/Userland/spng.cpp
+++ b/Userland/spng.cpp
@@ -1,0 +1,63 @@
+/* 
+ * Copyright (C) 2020 Matthew Graham
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include <AK/ByteBuffer.h>
+#include <LibCore/ArgsParser.h>
+#include <LibCore/File.h>
+#include <stdio.h>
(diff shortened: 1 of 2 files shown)


matthewgraham1:
It looks like the auto builder doesn't like the license header I use. Is the 2-clause BSD license a requirement or is there some specific way that I should format the header?

awesomekling:
It is absolutely a requirement yes :)

Userland/spng.cpp
```diff
@@ -0,0 +1,72 @@
...
+
+int main(int argc, char** argv)
+{
+    if (pledge("stdio wpath cpath rpath", nullptr) < 0) {
+        perror("pledge");
+        return 1;
+    }
+
+    auto in_file = Core::File::construct("/dev/stdin");
+    if (!in_file->open(Core::IODevice::ReadOnly)) {
```
bugaevc, line 41: Please don't do it like that. We already have stdin opened (fd 0), just use that.

Userland/spng.cpp
```diff
@@ -0,0 +1,72 @@
...
+    }
+    auto output_buffer = in_file->read_all();
+
+    if (pledge("stdio wpath cpath", nullptr) < 0) {
+        perror("pledge");
+        return 1;
+    }
+
+    bool append = false;
+    const char* output_file_name = "/dev/stdout";
```
bugaevc, line 53: Don't do that either :)

Userland/spng.cpp
```diff
@@ -0,0 +1,72 @@
...
+        perror("pledge");
+        return 1;
+    }
+
+    auto in_file = Core::File::construct("/dev/stdin");
+    if (!in_file->open(Core::IODevice::ReadOnly)) {
+        perror("open");
+        return 1;
+    }
+    auto output_buffer = in_file->read_all();
```
bugaevc, line 45: It's not super awesome to read up all the data into RAM; the original sponge cat's the data to a temporary file, and later replaces the destination file with that temp file.

But I guess we can just go with reading everything for now

matthewgraham1:
Sorry for the commit mess, I am new to git

awesomekling:
This is not mergeable as-is, due to the commit mess you mention.

awesomekling:

PR #2827, opened 2020-07-16
LibGfx: Split JPG frametype into component parts, and adjust errors

stelar7:
This updates some of the error checks to be more inline with the spec parameter value bounds.

### DIFF
--- a/Libraries/LibGfx/JPGLoader.cpp
+++ b/Libraries/LibGfx/JPGLoader.cpp
@@ -155,24 +155,26 @@ struct ComponentSpec {
 
 struct StartOfFrame {
 
-    // Of these, only the first 3 are in mainstream use, and refers to SOF0-2.
-    enum class FrameType {
-        Baseline_DCT = 0,
-        Extended_Sequential_DCT = 1,
-        Progressive_DCT = 2,
-        Sequential_Lossless = 3,
-        Differential_Sequential_DCT = 5,
-        Differential_Progressive_DCT = 6,
-        Differential_Sequential_Lossless = 7,
-        Extended_Sequential_DCT_Arithmetic = 9,
-        Progressive_DCT_Arithmetic = 10,
-        Sequential_Lossless_Arithmetic = 11,
-        Differential_Sequential_DCT_Arithmetic = 13,
-        Differential_Progressive_DCT_Arithmetic = 14,
-        Differential_Sequential_Lossless_Arithmetic = 15,
+    // Only the combinations (Huffman, Baseline_DCT, false), (Huffman, Sequential_DCT, false), (Huffman, Progressive_DCT, false) are in mainstream use
+    struct FrameType {
+        enum class Encoding {
+            Huffman = 0,
+            Arithmetic = 1,
+        };
+
+        enum class Type {
+            Baseline_DCT = 0,
+            Sequential_DCT = 1,
+            Progressive_DCT = 2,
+            Lossless = 3,
+        };
+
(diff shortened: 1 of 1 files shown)


Libraries/LibGfx/JPGLoader.cpp
```diff
@@ -155,24 +155,26 @@ struct ComponentSpec {
...
+        enum class Type {
+            Baseline_DCT = 0,
+            Sequential_DCT = 1,
+            Progressive_DCT = 2,
+            Lossless = 3,
+        };
+
+        Encoding encoding { Encoding::Huffman };
+        Type type { Type::Baseline_DCT };
+        bool differential { false };
```
devsh0, line 174: The name `FrameType` doesn't suit the struct as it clearly includes more information about the frame than just the type. Also, `frame.type.type` looks weird. Maybe rename the struct and the variable `type` to something else? Or alternatively, leave `FrameType` an enum and move non-type related data outside?

Libraries/LibGfx/JPGLoader.cpp
```diff
@@ -432,13 +434,13 @@ static inline bool bounds_okay(const size_t cursor, const size_t delta, const si
 
 static inline bool is_valid_marker(const Marker marker)
 {
-    if (marker >= JPG_APPN0 && marker <= JPG_APPNF) {
-        if (marker != JPG_APPN0)
-            dbg() << String::format("%04x not supported yet. The decoder may fail!", marker);
+    // APPn markers are application dependant. Since interpretation is left to the application, these can be safely ignored.
+    if (marker >= JPG_APPN0 && marker <= JPG_APPNF)
         return true;
```
devsh0, line 439: The decoder was implemented with JFIF in mind. Non-JFIF images (APP1 or above) may pack information in different ways. For example, the assumption that component IDs won't exceed 3 comes from [JFIF](https://www.w3.org/Graphics/JPEG/jfif3.pdf). The only promise that JPEG makes about component ids is that they will be consistent across multiple scan headers and frame specs. The debug info is printed so that we know, in case of decoding failures, that the image not being a JFIF may be the reason why it failed. We don't give up decoding when we come across non-JFIF images because some interchange formats are perfectly compatible with JFIF and would not cause any trouble while decoding.

Libraries/LibGfx/JPGLoader.cpp
```diff
@@ -540,13 +571,66 @@ static bool read_start_of_scan(BufferStream& stream, JPGLoadingContext& context)
...
+
+    if (successive_approximation_low_bits > successive_approximation_low_max) {
+        dbg() << stream.offset() << String::format(": ERROR! Successive Approximation low bits must be in range 0-%i was (%i)", successive_approximation_low_max, successive_approximation_low_bits);
         return false;
     }
+
+    if (successive_approximation_high_bits > successive_approximation_high_max) {
+        dbg() << stream.offset() << String::format(": ERROR! Successive Approximation high bits must be in range 0-%i was (%i)", successive_approximation_high_max, successive_approximation_high_bits);
+        return false;
+    }
```
devsh0, line 632: These validation checks better go in a different function. Just curious...are you planning to add support for progressive JPEG decoding?

devsh0:
Unless you're currently adding support for non-baseline JPEGs, most of the changes in this pull request look like an attempt to extensively future-proof the decoder. We shouldn't worry about the details that are defined outside the scope of what we currently support. Please let us know where you're going with these changes :)

awesomekling:

PR #2829, opened 2020-07-16
Kernel: Implement FIFOs/named pipes

### LINKED ISSUE #203: Kernel: Implement mkfifo()
We should support creating a FIFO file (aka "named pipe") via the `mkfifo()` syscall.

Ideally this would share code with the current implementation of `pipe()` which currently creates a perhaps-not-super-awesomely-named `FIFO` object..

See: https://pubs.opengroup.org/onlinepubs/9699919799/functions/mkfifo.html

Petelliott:
This pull request implements FIFOs/named pipes.

- [x] Opening a named pipe opens a pipe instead of a regular file.
- [x] Opening a named pipe blocks until the other end is open.
- [x] Allow `mknod(1)` with no major/minor numbers when opening a named pipe.
- [x] ~~`mkfifo(2)`~~ `mkfifo(3)`
- [x] `mkfifo(1)`

closes #203

### DIFF
--- a/Kernel/FileSystem/FIFO.cpp
+++ b/Kernel/FileSystem/FIFO.cpp
@@ -60,6 +60,35 @@ NonnullRefPtr<FileDescription> FIFO::open_direction(FIFO::Direction direction)
     return description;
 }
 
+NonnullRefPtr<FileDescription> FIFO::open_direction_blocking(FIFO::Direction direction)
+{
+    Locker locker(m_open_lock);
+
+    auto description = open_direction(direction);
+
+    if (direction == Direction::Reader) {
+        m_read_open_queue.wake_all();
+
+        if (m_writers == 0) {
+            locker.unlock();
+            Thread::current()->wait_on(m_write_open_queue, "FIFO");
+            locker.lock();
+        }
+    }
+
+    if (direction == Direction::Writer) {
+        m_write_open_queue.wake_all();
+
+        if (m_readers == 0) {
+            locker.unlock();
+            Thread::current()->wait_on(m_read_open_queue, "FIFO");
+            locker.lock();
+        }
+    }
+
+    return description;
+}
+
 FIFO::FIFO(uid_t uid)
     : m_uid(uid)
 {
--- a/Kernel/FileSystem/Inode.h
+++ b/Kernel/FileSystem/Inode.h
@@ -32,6 +32,7 @@
 #include <AK/RefCounted.h>
 #include <AK/String.h>
 #include <AK/WeakPtr.h>
+#include <Kernel/FileSystem/FIFO.h>
 #include <Kernel/FileSystem/FileSystem.h>
 #include <Kernel/FileSystem/InodeIdentifier.h>
 #include <Kernel/FileSystem/InodeMetadata.h>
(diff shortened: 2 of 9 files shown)


Kernel/FileSystem/FIFO.cpp
```diff
@@ -60,6 +60,27 @@ NonnullRefPtr<FileDescription> FIFO::open_direction(FIFO::Direction direction)
...
+        m_read_open_queue.wake_all();
+
+        if (m_writers == 0)
+            Thread::current()->wait_on(m_write_open_queue, "FIFO");
+    }
+
+    if (direction == Direction::Writer) {
+        m_write_open_queue.wake_all();
+
+        if (m_readers == 0)
```
bugaevc, line 77: Is there any locking here?

Petelliott: Good catch, I now lock the fifo unless we are waiting for the other end to be opened.

Petelliott: If you're referring to deadlock, then that won't happen because when one end is blocking it's `m_readers`/`m_writers` is already incremented, so the other side would never block.

Kernel/FileSystem/Inode.h
```diff
@@ -134,6 +137,7 @@ class Inode : public RefCounted<Inode>
     RefPtr<LocalSocket> m_socket;
     HashTable<InodeWatcher*> m_watchers;
     bool m_metadata_dirty { false };
+    RefPtr<FIFO> m_fifo;
```
bugaevc, line 140: So, what keeps the inode alive?

Say I `mkfifo("/foo")`, and that creates an inode and sets its `m_fifo` to a new FIFO object. Then the inode gets dropped and destructed, and so does the FIFO object since nothing else references it.

Now I `open("/foo")`, and that should open that same FIFO, but instead it finds me another inode (which is supposed to be "the same" FS inode, but it's a different object in memory), which gives me a fresh FIFO.

Petelliott: It is expected behavior that the FIFO would disappear when no one has it open. To keep it alive when it is open, I set the original inode of the `FileDescription` to the inode (after this review).

bugaevc: > It is expected behavior that the FIFO would disappear when no one has it open.

I still don't see how that could possibly work. Someone calls `mkfifo()`, a new FIFO is created, and (since no one has opened it yet) immediately destroyed?

Petelliott: When someone calls mkfifo, a fifo special file is created. `m_fifo` is actually a pipe, due to a weird naming choice in serenity. the code to actually create a fifo file already existed, it just wouldn't be opened properly. The pipe `m_fifo` is created when the inode is opened.

from the Linux Man pages, `FIFO(7)`
> The kernel maintains exactly one pipe object for each FIFO special file that is opened by at least one process.

bugaevc: Ah, I see, so `mkfifo()` doesn't actually create the FIFO, only a record in the FS about one.

Yeah, `set_original_inode()` would probably do the trick. I wonder how this works for local sockets, and whether it does actually work. (TmpFS, of course, never drops otherwise unused inodes, so that might be alleviating the problem.)

awesomekling:
LGTM. The question about what happens to unreffed inodes in the kernel is interesting, but it's a separate issue from this feature imo. :)

PR #2831, opened 2020-07-18
Taskbar: Use Name from .af file as tooltip

nico:
With this, System Monitor has "System Monitor" instead of
"SystemMonitor" as tooltip, matching the app's title bar
and menu bar title.

Same for File Manager and Text Editor.


awesomekling:
Sure that makes sense!

PR #2834, opened 2020-07-18
LibPThread: Make pthread_exit a noreturn function 

tryfinally:
LibPThread: mark pthread_exit a noreturn function using compiler attributes
LibThread: remove a call to pthread_exit from Thread::start lambda expression
                  as it make the return from the lambda unreachable.
                 Thread function should either call pthread_exit or return a status at end of function.

### DIFF
--- a/Libraries/LibPthread/pthread.h
+++ b/Libraries/LibPthread/pthread.h
@@ -35,7 +35,7 @@
 __BEGIN_DECLS
 
 int pthread_create(pthread_t*, pthread_attr_t*, void* (*)(void*), void*);
-void pthread_exit(void*);
+void pthread_exit(void*) __attribute__ ((noreturn));
 int pthread_kill(pthread_t, int);
 void pthread_cleanup_push(void (*)(void*), void*);
 void pthread_cleanup_pop(int);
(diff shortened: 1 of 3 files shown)


Libraries/LibPthread/pthread.h
```diff
@@ -35,7 +35,7 @@
 __BEGIN_DECLS
 
 int pthread_create(pthread_t*, pthread_attr_t*, void* (*)(void*), void*);
-void pthread_exit(void*);
+[[noreturn]] void pthread_exit(void*);
```
bugaevc, line 38: Nope, this is a C header.

tryfinally: My bad. Should revert pthread.h

bugaevc: Nah, it's fine, just use the C syntax for attributes

awesomekling:
Please observe the [contribution guidelines](https://github.com/SerenityOS/serenity/blob/master/CONTRIBUTING.md)

Specifically, squash your commits and prefix the commit message with a category.

Libraries/LibPthread/pthread.h
tryfinally: actually, its only extern "C" declarations. the actuall code is in pthread.cpp and is compiled with c++ compiler.
Why we should prefer __attribute__ ((noreturn)) ?

bugaevc: Because it's a C header: pure C code (mainly from ports of third-party software) can `#include` it.

tryfinally: Of course! thanks.

PR #2841, opened 2020-07-20
TextEditor: Preserve preview scroll position across page refreshes

ccapitalK:
This is a tiny fix to make the text editor preview pane not scroll to the top every time it refreshes. This makes editing long documents with the TextEditor much more pleasant.


ccapitalK:
Small change to make the code a bit simpler.

PR #2842, opened 2020-07-20
LibWeb: Implement quirks mode detection

Lubrsi:
This allows us to determine which mode to render the page in.

Exposes "doctype" and "compatMode" on Document.
Exposes "name", "publicId" and "systemId" on DocumentType.


awesomekling:
Excellent work :)

PR #2844, opened 2020-07-21
LibTLS: Reschedule the timeout if we're too slow

### LINKED ISSUE #2843: LibTLS: Handshake timeout doesn't take into account that we may timeout.
During handshake, there is a 10 second timeout in place. The comment on line 125 of Socket.cpp assumes that the server took too long to respond, but we may actually be the one who took too long to respond.

From Wireshark:
Wireshark_smIrslYYHK

Once the server sends us "Change cipher spec" and we acknowledge it, we then take about 10 seconds to respond with the client key exchange, and then the timeout kills the connection.

This happens when connecting to [Linus's ":^)" tracker.](https://happy-serenityos.linus.dev/)
Removing the timeout allows us to connect successfully and load the page, albeit pretty slowly.

~~A seperate issue is that the timeout may fire multiple times.~~ Nevermind, this is because the favicon fails too.

alimpfard:
Previously, we would not care if the handshake timer timed out because
the server was too slow, or because we were too slow, this caused
connections to fail when the system was under heavy load.
This patch fixes this behaviour (and closes #2843) by checking if the
timeout delay was within margin of error of the max timeout.

### DIFF
--- a/Libraries/LibTLS/Socket.cpp
+++ b/Libraries/LibTLS/Socket.cpp
@@ -122,18 +122,28 @@ bool TLSv12::common_connect(const struct sockaddr* saddr, socklen_t length)
         deferred_invoke([&](auto&) {
             m_handshake_timeout_timer = Core::Timer::create_single_shot(
                 m_max_wait_time_for_handshake_in_seconds * 1000, [&] {
-                    // The server did not respond fast enough,
-                    // time the connection out.
-                    alert(AlertLevel::Critical, AlertDescription::UserCanceled);
-                    m_context.connection_finished = true;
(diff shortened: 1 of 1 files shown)


awesomekling:
Seems fair. Why are we so slow? :|

alimpfard:
I didn't see any particular slowness personally, the only way I could reproduce the behaviour was to keep the cpu busy by running a `for(;;);` in js.
Otherwise the site in question loaded just fine...

Lubrsi:
I found it because I’m currently using WSL2 :P
I tried doing nested virtualisation to get KVM but it seemingly didn’t improve performance...

PR #2846, opened 2020-07-21
BitTorrent client

deoxxa:
BitTorrent is the way of the future! Or so I'm told. I hope to build a functional BitTorrent client and utilities comprised of several libraries (Bencode, Kademlia DHT, BitTorrent HTTP/UDP trackers, BitTorrent protocol), a daemon, a GUI program, and some CLI programs.


Libraries/LibBencode/Dictionary.h
```diff
@@ -0,0 +1,107 @@
...
+#pragma once
+
+#include <AK/HashMap.h>
+#include <AK/StringBuilder.h>
+#include <AK/Vector.h>
+#include <LibBencode/Value.h>
+
+namespace Bencode {
+
+class Dictionary {
```
elcuco, line 36: Why not deriving `HashMap<String, Value>`  instead of this re-implementation/encapsulation?

Libraries/LibBencode/Parser.h
```diff
@@ -0,0 +1,63 @@
...
+
+#pragma once
+
+#include <AK/Function.h>
+#include <AK/Vector.h>
+#include <LibBencode/Value.h>
+
+namespace Bencode {
+
+class Parser {
```
elcuco, line 35: So basically a variant? Should this be in `AK`?

Libraries/LibBencode/Value.h
```diff
@@ -0,0 +1,118 @@
...
+#include <AK/String.h>
+#include <AK/StringBuilder.h>
+#include <AK/Vector.h>
+
+namespace Bencode {
+
+class Dictionary;
+class List;
+
+class Value {
```
elcuco, line 39: ... or is this the variant...?

elcuco (commented):
All I see is a `BtInfo` program. No download?

deoxxa:
> All I see is a `BtInfo` program. No download?

Not yet - but one day! Unless someone beats me to it (which I would be totally fine with).

Libraries/LibBencode/Dictionary.h
deoxxa: I stole most of the ideas here from the JSON types in AK. I'll welcome any cleanup that might help to remove duplicate functionality.

PR #2847, opened 2020-07-21
LibWeb: Assert we're parsing a fragment on fragment cases

Lubrsi:
The specification says that parts labelled as a "fragment case" will
only occur when parsing a fragment. It says that if it occurs when
not parsing a fragment, then it is a specification error.

We should probably assume at this point that it's an implementation
error. This fixes a few little mistakes that were caught out by this.

Also moves the context element outside insertion mode reset,
as other (unimplemented) parts refer to it, such as
"adjusted current node".

Also cleans up insertion mode reset.


awesomekling:
Great idea :)

PR #2849, opened 2020-07-22
TextEditor: Jump to word break when deleting and holding Ctrl modifier

gootik:
I often use this shortcut for quicker text editing. Thought it'd be useful for Serenity as well.


awesomekling:
Since we already use Ctrl+Left/Right for jumping word breaks left/right, let's use Ctrl as the modifier here as well. :)

gootik:
@awesomekling makes sense. fixed it :D

PR #2851, opened 2020-07-22
LibJS: Implement String.prototype.charCodeAt

nico:
It's broken for strings with characters outside 7-bit ASCII, but
it's broken in the same way as several existing functions (e.g.
charAt()), so that's probably ok for now.


awesomekling:
Hmm I guess this is okay. The lack of proper UTF16 support in LibJS is generally pretty sad though.

PR #2852, opened 2020-07-22
LibJS: Add FIXMEs to a few functions that need UTF-16 handling

nico:


Libraries/LibJS/Runtime/StringPrototype.cpp
```diff
@@ -319,6 +321,7 @@ JS_DEFINE_NATIVE_FUNCTION(StringPrototype::substring)
     if (interpreter.argument_count() == 0)
         return js_string(interpreter, string);
 
+    // FIXME: index_start and index_end should index a UFF-16 codepoint view of the string.
```
awesomekling, line 324: Typo, UFF-16

Libraries/LibJS/Runtime/StringPrototype.cpp
```diff
@@ -381,6 +385,7 @@ JS_DEFINE_NATIVE_FUNCTION(StringPrototype::slice)
...
```
awesomekling, line 388: Typo, UFF-16

Libraries/LibJS/Runtime/StringPrototype.cpp
```diff
@@ -429,10 +434,10 @@ JS_DEFINE_NATIVE_FUNCTION(StringPrototype::last_index_of)
         return {};
     if (search_string.length() > string.length())
         return Value(-1);
-
     auto max_index = string.length() - search_string.length();
     auto from_index = max_index;
     if (interpreter.argument_count() >= 2) {
+        // FIXME: from_index should index a UFF-16 codepoint view of the string.
```
awesomekling, line 440: Also here

Libraries/LibJS/Runtime/StringPrototype.cpp
nico: It's German: https://en.wiktionary.org/wiki/uff ...and apparently Norwegian: https://en.wikipedia.org/wiki/Uff_da

PR #2853, opened 2020-07-22
LibGUI: Remove unnecessary LibHTTP #include

### LINKED ISSUE #2848: Unnecessary LibHTTP #include (I think?)
https://github.com/SerenityOS/serenity/blob/223b96c820c516c96c0d7dc9b860369b13633643/Libraries/LibGUI/ToolBarContainer.h#L31

Found this while looking for where LibHTTP is used. It looks like it might not be required.

alimpfard:
This closes #2848.

Looks like everything is very happy to compile without it :shrug:


awesomekling:
Ding dong

PR #2854, opened 2020-07-22
Make btoa() and atob() correctly handle values between 128 and 255

nico:


nico:
Looks like there's no test-js for browser-only js.

awesomekling:
> Looks like there's no test-js for browser-only js.

Yes maybe it's time to add something like that..

PR #2857, opened 2020-07-22
LibCrypto: Optimize UnsignedBigInteger import_data/export_data

tomuta:
No need to do complicated math to import or export numbers,
just convert the byte stream to words and vice versa.

### DIFF
--- a/Libraries/LibCrypto/BigInt/UnsignedBigInteger.cpp
+++ b/Libraries/LibCrypto/BigInt/UnsignedBigInteger.cpp
@@ -29,6 +29,25 @@
 
 namespace Crypto {
 
+UnsignedBigInteger::UnsignedBigInteger(const u8* ptr, size_t length)
+{
+    m_words.resize_and_keep_capacity((length + sizeof(u32) - 1) / sizeof(u32));
+    size_t in = length, out = 0;
+    while (in >= sizeof(u32)) {
+        in -= sizeof(u32);
+        u32 word = ((u32)ptr[in] << 24) | ((u32)ptr[in + 1] << 16) | ((u32)ptr[in + 2] << 8) | (u32)ptr[in + 3];
+        m_words[out++] = word;
+    }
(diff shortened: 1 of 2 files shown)


tomuta:
Hm, I didn't realize there were unit tests for this. How do I run these? Seems like `ninja test` only runs the stuff in `Tests/`

alimpfard:
> How do I run these? 

`test-crypto test -c` inside serenity, or `Meta/Lagom/test-crypto test -c` outside.

`ninja test` _should_ run them, as the first test even.

tomuta:
> > How do I run these?
> 
> `test-crypto test -c` inside serenity, or `Meta/Lagom/test-crypto test -c` outside.
> 
> `ninja test` _should_ run them, as the first test even.

Hm not for me. That command only runs 21 tests, looks like only the tests for AK. Seems like there's only Build/Userland/test-crypto, I don't see one in Meta/Lagom.

alimpfard:
Have you built with `BUILD_LAGOM=1`?
that might be why.

tomuta:
> Have you built with `BUILD_LAGOM=1`?
> that might be why.

Well I did not because I had no idea about that variable ;-)

alimpfard:
lol
we should probably mention that somewhere outside of CMakeLists.txt, could probably be useful.

tomuta:
Well I can't get `BUILD_LAGOM=1` to work. I tried `BUILD_LAGOM=ON`, too. Not even blowing away all of `Build/` and rebuilding the toolchain seems to make it work. Whatever, I think this will make it happy...

alimpfard:
Ah, you need to pass it to cmake, sorry if I wasn't clear on that, I use `cmake .. -GNinja -DBUILD_LAGOM=1`.

awesomekling:
Not sure if you got tests working locally, but I ran them here and they seem fine :)
(Also the CI runs them on PR's)

PR #2858, opened 2020-07-23
LibJS: Add tests for bitwise & and ^

nico:
And fix some edge case conversion bugs found by the tests.


nico:
Not sure what the story is with tests and license headers, but the bitwise | test doesn't have any either, so I made these consistent with that.

linusg:
> Not sure what the story is with tests and license headers

(JS) Test files don't need license headers and won't be checked by the script that checks for license headers :)

PR #2859, opened 2020-07-23
Browser: Focus input text field in JS console by default

nico:


nico:
Not sure if this is the nicest way to do this implementation wise, but it's nice to be able to type immediately after hitting ctrl-i.

awesomekling:
Good enough for me :)

PR #2860, opened 2020-07-23
Browser: Escape JS source operators for HTML display in console

nico:
Console inputs to try before and after this patch:
- `0xffff & 0xff`
- `"a & b"`
- `"<div>"`
- `a &` (to see the escaping in the error hint)


awesomekling:
Please use `escape_html_entities()` from `AK/String.h` :)

nico:
Done!

PR #2861, opened 2020-07-23
LibWeb: Add basic SVG support

mattco98:
This PR adds initial SVG support. The SVG element has an entire spec to itself, and it is quite extensive, so this PR merely scratches the surface, but it is a solid start. 

Currently, the only supported svg child-tag is `path`. I started with `path` because all other graphical elements can be described in terms of a path, so once path is supported, the other graphical elements are trivial. Currently, path supports all instructions except `CurveTo`, `SmoothCurveTo`, and `SmoothBezierCurveTo`. 

To support the `EllipticalArcTo` command, elliptical arc support was added to `Gfx::Path`. I pretty much mirrored the way Bezier curves were implemented (the divide and conquer method). cc @alimpfard for that. Here's an ss of the same demo in Firefox and Serenity:

<details>
<summary>svg.html from Firefox</summary>
<br />

</details>

<details>
<summary>svg.html from Serenity</summary>
<br />

</details>

As far as future PRs go, the short-term roadmap is to implement the rest of the graphical elements, as well as supported user-to-screen coordinate transformations (i.e. the `viewBox` property). After that, we can look at some more complex features, such as grouping and painting order. There's no shortage of work to be done here :)


alimpfard:
:star_struck: whoa svg

alimpfard:
I think this (SVG parse/render that is) would be more useful as a separate library, who knows, someone might come along and make the second coming of Inkscape :P 

Although keeping it in libweb until we end up needing it outside is also ok by me, I think :shrug:

mattco98:
To be honest, I wasn't really aware that SVG was useful outside of a browser/HTML context. It would not be super hard to refactor this PR into a LibSVG, as LibGfx does most of the heavy lifting. It would sure be a hell of a lot easier to do it now than it would be 10 SVG-related PRs down the road!

awesomekling:
> To be honest, I wasn't really aware that SVG was useful outside of a browser/HTML context. It would not be super hard to refactor this PR into a LibSVG, as LibGfx does most of the heavy lifting. It would sure be a hell of a lot easier to do it now than it would be 10 SVG-related PRs down the road!

IMO that would be more work than value. If you want to render an SVG into a bitmap, you can just go via LibWeb.
But perhaps you have some specific standalone feature in mind?

alimpfard:
I've personally always wanted svg's in the terminal, perhaps this is really outside what serenity should feel like...

Either way, making things depend on libweb (and its interface) just because they want to do things with vector graphics is a bit...silly? although it's definitely possible refactor out the svg stuff whenever they're needed outside libweb.

I don't have any concrete plans, so...no problems here with keeping svg's inside libweb.
if @mattco98 hasn't already moved the stuff out, we can just continue with the PR as is.

awesomekling:
> Either way, making things depend on libweb (and its interface) just because they want to do things with vector graphics is a bit...silly? although it's definitely possible to refactor out the svg stuff whenever they're needed outside libweb.

Forcing a higher level of abstraction onto something as complex as SVG because you "might wanna do something later" is a bit hard to justify :)

Either way, it wouldn't be terribly difficult to link with LibWeb and then use the SVG DOM and layout functionality without creating a web view widget. :) The LibWeb painting code is widget-agnostic, you just give it a `PaintContext` and it'll draw into it. Betcha we could build on that.

alimpfard:
Certainly willing to wait for actual tangible use cases before making any sort of API/abstraction.

> The LibWeb painting code is widget-agnostic

LibWeb is rather complex, reaching in and grabbing its SVG impl would be pretty difficult, but I can see a how library for doing just that could help.
Fair enough.

tomuta:
High DPI screens make svg an interesting choice for icons. Same goes for scaling features.

awesomekling:
Exciting stuff @mattco98!

Please use the same interface names as the spec, also let's put this in a separate directory, `LibWeb/SVG/`
So `HTMLPathElement` => `SVGPathElement`, `HTMLSvgElement` => `SVGSVGElement` etc.
Also please add a new `Web::SVG::TagNames` instead of putting them into the `Web::HTML::TagNames` since we'll eventually need to namespace (in the DOM sense) these tags differently from each other :+1:

awesomekling:
Also, let's follow the SVG object model precisely. For example, here's a nice diagram on MDN for `SVGPathElement`: https://developer.mozilla.org/en-US/docs/Web/API/SVGPathElement

So, `SVGPathElement` inherits `SVGGeometryElement` inherits `SVGGraphicsElement` inherits `SVGElement` inherits `Element`. :)

mattco98:
I honestly have no clue what to do to fix this build. It works fine on my machine, and I don't understand the error in the Travis logs. FYI, this is all due to 3f45e9ab1ef560eb606ae03a18d6435b718d9982. If I rebase on the commit before, it works fine.

alimpfard:
Looks like one gcc bootstrap phase does not like `noexcept`, which fails the toolchain compile just when it's building its libstdc++, which _then_ causes there to be no `initializer_list`.
as for how 3f45e9a ties into all this...probably just forcing you to add `noexcept` after all math.h functions, which is also a C header.

what comes to mind is adding `noexcept` conditionally (only if in C++)...

mattco98:
Finally builds! 

> what comes to mind is adding `noexcept` conditionally (only if in C++)...

This was the fix, but only if we're in C++11 or greater. Thanks!

alimpfard:
wow.
just...wow.

awesomekling:
Going forward, we should move towards creating LayoutNodes for SVG elements and making them part of the regular layout tree.

PR #2862, opened 2020-07-23
LibWeb+test-web: Create test-web program, add doctype test

Lubrsi:
LibWeb currently has no test suite or program. Let's change that :^)

test-web is mostly a copy of test-js, but modified for LibWeb.
test-web imports both LibJS/Tests/test-common.js and
LibWeb/Test/test-common.js

LibWeb's suite provides the ability to specify the page to load,
what to do before the page is loaded, and what to do after it's
loaded.

This also provides a test of document.doctype and its close sibling
document.compatMode.

Currently, this isn't added to Lagom because of CodeGenerators.


awesomekling:
Awesome effort :D

PR #2863, opened 2020-07-23
LibC: Remove duplicate gs touch during gettid()/getpid() fast path

bgianfo:
While profiling I noticed that gettid() was hitting gs register
twice, once for the initial fetch of s_cache_tid out of TLS for
the initialization check, and then again when we return the actual
value.

Optimize the implementation to cache the value so we avoid the
double fetch during the 99% case where it's already set. With
this change gettid() goes from being the 3rd most sampled function
in test-js, to pretty much disappearing into  ~20th place.

Additionally implement the same optimization for getpid().

### DIFF
--- a/Libraries/LibC/unistd.cpp
+++ b/Libraries/LibC/unistd.cpp
@@ -202,9 +202,12 @@ gid_t getgid()
 
 pid_t getpid()
 {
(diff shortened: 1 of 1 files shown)


awesomekling:
Huh, okay!

PR #2864, opened 2020-07-23
LibWeb: Add tests for atob() and btoa()

nico:
Thanks to #2862, it's now easy to write a test for #2854 :)


awesomekling:
V. nice!

PR #2865, opened 2020-07-23
Lagom: Add LibGemini, LibGfx

nico:
They are dependencies of LibWeb and might be useful for
running test-web on GitHub actions one day.

I happened to have this in a local branch for...something, so I figured
with test-web landing there might be some interest in this upstream.


linusg:
> running test-web on GitHub actions one day.

Considering we're already running `test-js` on CI (using Travis though) this should totally be a thing, yes.

nico:
Lagom being built with `-O2` instead of `-Os` made gcc's -Wsometimes-uninitialized find a bunch of legit issues in the jpeg and gif decoders. Optimization-level-dependent diagnostics are silly, but they did find a bunch of bugs here.

awesomekling:
Fantastic! :D

PR #2867, opened 2020-07-24
Add support for double on ArgParser

elcuco:
Code is pretty trivial - it just does not compile.

Build system is confused between math.h from rootfs, and toolchain. Hoping for some
one to assist here :)

### DIFF
--- a/Libraries/LibCore/ArgsParser.cpp
+++ b/Libraries/LibCore/ArgsParser.cpp
@@ -324,6 +324,28 @@ void ArgsParser::add_positional_argument(int& value, const char* help_string, co
     add_positional_argument(move(arg));
 }
 
+static constexpr bool isnan(double __x) { return __builtin_isnan(__x); }
+
+void ArgsParser::add_positional_argument(double& value, const char* help_string, const char* name, Required required)
+{
+    Arg arg {
+        help_string,
+        name,
+        required == Required::Yes ? 1 : 0,
+        1,
+        [&value](const char* s) {
+            char *p;
+            double v = strtod(s, &p);
+            bool valid_value = !isnan(v) && p != s;
+            if (valid_value) {
+                value = v;
+            }
+            return valid_value;
+        }
+    };
+    add_positional_argument(move(arg));
+}
+
 void ArgsParser::add_positional_argument(Vector<const char*>& values, const char* help_string, const char* name, Required required)
 {
     Arg arg {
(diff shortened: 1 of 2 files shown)


alimpfard:
I think you can just use `strtod`'s `endptr` argument to check if all the input was consumed, to sidestep the math.h issue for the time being.

elcuco:
@alimpfard yes, this will fix somethings. However, "123.123 FAIL" is not a valid number, but can be detected as such by this implementation. An alternative can be to "implement" `isnan()` locally - 

``` C
static constexpr bool isnan(double __x)
{ return __builtin_isnan(__x); }
```

This does compile (and I don't need `#include <math.h>` but - its ugly as a wet dog.

awesomekling:
Please observe the [contribution guidelines](https://github.com/SerenityOS/serenity/blob/master/CONTRIBUTING.md). Specifically, please squash your commits, and follow the project's commit message style.

elcuco:
I am looking for a better alternative than just re-defining `isnan()` statically on this file. Including `math.h` fails, and I do need `isnan()` to verify that the number.

I think @awesomekling will not be happy about using low level libC here, and he wishes `StringUtils` would have been used. This is a better approach on the long run. Since the OS has several integer parsing functions for example.

Libraries/LibCore/ArgsParser.cpp
```diff
@@ -324,6 +324,29 @@ void ArgsParser::add_positional_argument(int& value, const char* help_string, co
     add_positional_argument(move(arg));
 }
 
+// Theoritically this is bad. In practice - this is not that harmful :)
+static constexpr bool isnan(double __x) { return __builtin_isnan(__x); }
```
awesomekling, line 328: I don't understand this comment. Why is this theoretically bad?
We use `__builtin_isnan()` elsewhere already.

elcuco: Sorry for late reply.

I only see it here: https://github.com/SerenityOS/serenity/blob/6075defd55e66a3c627cb6912a1431971eaa1ed6/Libraries/LibJS/Runtime/Value.h#L78 (and in the toolchain itself).

I am worried by the fact that `#include <math.h>`  (or `cmath`) fails. I am also worried that this will lock serenityOS to gcc, and moving to clang or a self made compiler (wtf not, this project already did everything else).

PR #2868, opened 2020-07-24
test-web: Add ability to change page mid-test

Lubrsi:
This allows you to not have to write a separate test file
for the same thing but in a different situation.

This doesn't handle when you change the page with location.href
however.

Changes the name of the page load handlers to prevent confusion
with this.

### DIFF
--- a/Userland/test-web.cpp
+++ b/Userland/test-web.cpp
@@ -25,6 +25,7 @@
  */
 
 #include <AK/URL.h>
+#include <AK/Function.h>
 #include <AK/JsonValue.h>
 #include <AK/JsonObject.h>
 #include <AK/QuickSort.h>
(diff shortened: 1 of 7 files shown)


Userland/test-web.cpp
```diff
@@ -163,6 +205,34 @@ Vector<String> get_test_paths(const String& test_root)
 void TestRunner::run() {
     size_t progress_counter = 0;
     auto test_paths = get_test_paths(m_web_test_root);
+
+    g_on_page_change = [this](auto& page_to_load) {
+        if (!page_to_load.is_valid())
+        {
```
awesomekling, line 211: Please use `clang-format`.

Lubrsi:
Done 👍

PR #2869, opened 2020-07-24
PixelPaint: Add a bunch of spatial filters

alimpfard:
_/me blurs the entire layer_
whoo fun

It should be noted that the single dialog added (a matrix input dialog) is _bad_.

### DIFF
--- a/Applications/PixelPaint/Filters/Filter.h
+++ b/Applications/PixelPaint/Filters/Filter.h
@@ -1,0 +1,64 @@
+/*
+ * Copyright (c) 2020, the SerenityOS developers.
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ *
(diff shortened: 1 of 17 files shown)


awesomekling:
Cool feature :)

Let's organize the code so that filters don't ask the ImageEditor for the active layer, but rather filter just get handed a layer (or even better, perhaps just a bitmap+rect) and told to filter it. :)

alimpfard:
Bitmap+rect is best, but I didn't see any way to overwrite the contents of a bitmap, am I missing something?

alimpfard:
Hey that made the code smaller!
Nice!

Applications/PixelPaint/Filters/Filter.h
```diff
@@ -0,0 +1,63 @@
...
+            , target_rect(rect)
+
+        {
+        }
+
+        Gfx::Bitmap& bitmap() const { return target_bitmap; }
+        const Gfx::IntRect& rect() const { return target_rect; }
+
+        Gfx::Bitmap& target_bitmap;
+        Gfx::IntRect target_rect;
```
awesomekling, line 49: No need for getters since the members are public anyway. Or perhaps you intended to make this a full class but forgot about it halfway through? :D

alimpfard: ...memory issues man...

PR #2872, opened 2020-07-25
Kernel: Allow process creation during low-entropy condition

### LINKED ISSUE #2871: Lack of entropy during boot hangs system
tl;dr: Lack of entropy means that processes cannot be created.

Process creation should instead only require weak entropy.

## Log lines

``` 
[Kernel]: CPU[0]: Supported features: nx pae pge
[Kernel]: CPU[0]: No RDRAND support detected, randomness will be poor
[Kernel]: Installing Unhandled Handlers
[Kernel]: MM: boot_pdpt @ P0x00109000
[Kernel]: MM: boot_pd0 @ P0x0010a000
[Kernel]: MM: boot_pd3 @ P0x0010b000
<SNIP FOR BREVITY>
[colonel(0:0)]: Scheduler[0]: idle loop running
[init_stage2(1:1)]: BW: about to mount VFS
[init_stage2(1:1)]: VFS: mounted root from Ext2FS (device:3,0 (PATADiskDevice))
[init_stage2(1:1)]: Loading kernel symbol table...
[init_stage2(1:1)]: BW: done loading
[init_stage2(1:1)]: BW: done set graphical
[init_stage2(1:1)]: BW: done parsing commandline
```

I inserted those "BW" lines to narrow it down, and it consistently hangs on this line:

```C++
// init.cpp:349
Process::create_user_process(thread, userspace_init, (uid_t)0, (gid_t)0, (pid_t)0, error, {}, {}, tty0);
```

## I built it correctly

When I follow the build instructions in [`BuildInstructions.md`](https://github.com/SerenityOS/serenity/blob/master/Documentation/BuildInstructions.md) quite verbatim, serenity hangs.

If you'd like to know [yes](https://github.com/BenWiederhake/serenity-hang/blob/master/serenity-from-scratch.sh) I have followed the [entire build process from scratch](https://github.com/BenWiederhake/serenity-hang/blob/master/full-build-from-scratch-1595639004.log).

## I waited long enough

After two minutes, still nothing happens.

## Impatience gets rewarded!?

The following things seem to get the Kernel "unstuck" though:
- Clicking about 20 times
- Typing about 100 keys
- Moving the mouse around for about 5 seconds
(These numbers are just guesstimates to give you a feel. Please don't actually count the clicks/keys/seconds.)

## Randomness?

```
(04:51:42) Ben: When scrolling through the commit logs it looked like Andreas is using an RDRAND processor, and other processors are not tested as thoroughly.
(04:51:59) Ben: My processor doesn't have RDRAND, so maybe that has something to do with it.
(04:52:35) Ben: That would leave the question why SystemServer hangs, because why should it need randomness?
(04:53:28) CxByte: I guess we don't use a shitty algorithm when there's no entropy then...get_good_random_bytes() does a wait_for_entropy()
(04:54:02) CxByte: need randomness: Don't we give some sort of random value to every process? lemme check
(04:56:00) CxByte: yep, do_exec() -> generate_auxiliary_vector() -> get_good_random_bytes()
(04:56:34) Ben: That would explain why SystemServer "needs" entropy.
```
(Thanks, CxByte!)

## Potential fixes

I see two ways:
1. Try even harder to get good entropy on a "shut-in" machine without RDRAND and with too few naturally-occurring interrupts during boot. (But then we'll end up with the same problem sooner or later. Also, the Kernel should *show* that this is what's going on.)
1. Remove the requirement of giving good entropy to SystemServer. (But then it will hang on process 2, whatever that is.)
1. When waiting for new entropy to trickle in, use the time to cause new entropy, e.g. by harddrive reads. (But that's a lot of work to implement, and is frankly hard to argue as correct.)
1. Remove the requirement of giving good entropy to new processes.

## The fix

So … `generate_auxiliary_vector()` should probably call `get_fast_random_bytes()`.

However, I don't know for sure where this entropy is used, or why it was originally decided that strong entropy is needed here, so I don't want to change it yet.

## Workaround

Wiggle your mouse.

Stop laughing! I'm a keyboard guy, so this *is* an entirely separate step for me! :)

## Why so elaborate

This was an unintentional rubberduck. I started writing this thinking that Serenity/SystemServer/qemu are behaving completely unpredictably. Then I thought it's something to do with Kernel cmdline parsing. Then I thought it's something to do with interrupts. Then I thought it's maybe a qemu bug. Then I thought maybe it's a "promptless" prompt. Then I thought it's a button that I accidentally press when clicking blindly. Then I thought it's something that increases with each interrupt … Oh god it's the entropy bug all over again.

I hope you enjoyed the ride. I certainly did! :^)

BenWiederhake:
Fixes #2871.

Ignoring the 'securely generated bytes' constraint seems to
be fine for Linux, so it's probably fine for Serenity.

Note that there *might* be more bottlenecks down the road
if Serenity is started in a non-GUI way. Currently though,
loading the GUI seems to generate enough interrupts to
seed the entropy pool, even on my non-RDRAND setup. Yay! :^)

### DIFF
--- a/Kernel/Process.cpp
+++ b/Kernel/Process.cpp
@@ -1118,7 +1118,7 @@ Vector<AuxiliaryValue> Process::generate_auxiliary_vector() const
     auxv.append({ AuxiliaryValue::Secure, ((m_uid != m_euid) || (m_gid != m_egid)) ? 1 : 0 });
 
     char random_bytes[16] {};
-    get_good_random_bytes((u8*)random_bytes, sizeof(random_bytes));
+    get_fast_random_bytes((u8*)random_bytes, sizeof(random_bytes));
 
     auxv.append({ AuxiliaryValue::Random, String(random_bytes, sizeof(random_bytes)) });
 


awesomekling:
Seems reasonable.

PR #2874, opened 2020-07-25
Shell: Don't crash when autocompleting a syntax error

BenWiederhake:
For example, type `Hello?` including the question mark, and then press TAB.

Previously, this would crash the Shell. Now, it merely refuses to make any suggestions.

Maybe it really is just as straight-forward, or maybe that assert there was supposed to be "impossible"? Maybe @alimpfard can help out? :)


alimpfard:
Really, the assert is there as a "don't worry about it, will probably deal with it later".
You might as well apply this to `PathRedirectionNode` and `Execute`.

alimpfard:
This isn't about a syntax error though, this is about completing a glob in program name position

awesomekling:
@alimpfard How would you like to proceed here?

alimpfard:
I suppose not completing non-barewords is fine.
@BenWiederhake should just change the commit message to be more accurate.

PR #2875, opened 2020-07-25
AudioApplet: AudioWindowlet

thankyouverycool:
Comfy volume slider for audio applet. Percentages can now be toggled on/off.
volume-slider


MenuApplets/Audio/main.cpp
```diff
@@ -52,21 +58,88 @@ class AudioWidget final : public GUI::Widget {
...
+        if (event.button() == GUI::MouseButton::Left) {
+            if (!m_slider_window->is_visible())
+                open();
+            else
+                close();
             return;
-        m_audio_client->set_muted(!m_audio_muted);
-        update();
+        }
+        if (event.button() == GUI::MouseButton::Right) {
```
benit8, line 139: You could remove the above early `return` if you added an `else` to this `if`, or perhaps make the whole thing into a `switch` statement

Services/WindowServer/WindowManager.cpp
```diff
@@ -379,6 +379,9 @@ void WindowManager::notify_rect_changed(Window& window, const Gfx::IntRect& old_
 
     tell_wm_listeners_window_rect_changed(window);
 
+    if (window.type() == WindowType::MenuApplet)
+        MenuManager::the().set_needs_window_resize();
```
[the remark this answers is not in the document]
thankyouverycool: We can and originally did! Thought I was missing some logic because of strange text artifacting, but it seems a separate issue, and I forgot to change it back. Dunno if KVM=disabled exaggerates the problem, but the applets in menubar have never drawn correctly for me on boot. Requires an additional action like focus change to update.

MenuApplets/Audio/main.cpp
thankyouverycool: True, this is just the style I see most often in mouse/key_events

PR #2876, opened 2020-07-25
AK: demonstarte that binary_search will fail on large array. [Test Fails]

tryfinally:
Binary search will fail due to signed/unsigned conversions and overflow.
**Not to be merged**

### DIFF
--- a/AK/Tests/TestBinarySearch.cpp
+++ b/AK/Tests/TestBinarySearch.cpp
@@ -28,6 +28,7 @@
 
 #include <AK/BinarySearch.h>
 #include <cstring>
+#include <new>
 
 TEST_CASE(vector_ints)
 {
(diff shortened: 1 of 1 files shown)


BenWiederhake:
I've been prodding this bug, and is seems that the magic threshold is:
```
const size_t N = 0x40000001; // segfault
const size_t N = 0x40000000; // no segfault
```

Turns out, it's a simple singedness issue in `binary_search`. It ends up trying to compute the "middle" between `0x40000000` and `0x40000000` by computing `(0x40000000 + 0x40000000) / 2 = 0x80000000 / 2 = 0xc0000000`, because `0x80000000` is *negative*.

I'll make a PR soon.

tryfinally:
updated test included in PR #2961

PR #2879, opened 2020-07-25
Userland: Add passwd utility

### LINKED ISSUE #1168: Userland: passwd utility
Now that we have a `useradd` utility, it would be nice to have a matching `passwd`!

Petelliott:
- [x] `crypt` and `crypt_r`
- [x] `passwd`
- [x] passwords in `su`

note: `passwd` will not work with a read only filesystem, which serenity currently defaults to.

I might add `/etc/shadow` or `/etc/master.passwd` In a future PR, so let me know if you have a preference.

closes #1168


Petelliott:
closing this because travis can't figure out how to build the right commit.

Petelliott:
I can't figure out why the build is failing.
```
[ 19%] Building CXX object Kernel/CMakeFiles/Kernel.dir/ACPI/DynamicParser.cpp.o
In file included from /home/travis/build/SerenityOS/serenity/./Kernel/ACPI/Definitions.h:31,
                 from /home/travis/build/SerenityOS/serenity/./Kernel/ACPI/Parser.h:30,
                 from /home/travis/build/SerenityOS/serenity/./Kernel/ACPI/DynamicParser.h:30,
                 from /home/travis/build/SerenityOS/serenity/Kernel/ACPI/DynamicParser.cpp:27:
/home/travis/build/SerenityOS/serenity/./AK/Vector.h:40:14: fatal error: initializer_list: No such file or directory
   40 | #    include <initializer_list>
      |              ^~~~~~~~~~~~~~~~~~
```
I didn't change anything related to this, and it builds on my machine.

If anyone has seen something before, your help is greatly appreciated.

AK/String.cpp
```diff
@@ -125,7 +125,7 @@ String String::isolated_copy() const
 String String::substring(size_t start, size_t length) const
```
awesomekling, line 125: The commit message refers to `String::split()` but the function changed is `String::substring()`.
Also it would be nice to add a test to `AK/Tests/TestString.cpp` to document this change.

Libraries/LibCore/CMakeLists.txt
```diff
@@ -7,6 +7,7 @@ set(SOURCES
     Event.cpp
     EventLoop.cpp
     File.cpp
+    get_password.cpp
```
awesomekling, line 10: Please call this `GetPassword.cpp` so it fits with the group.

Libraries/LibCore/get_password.cpp
```diff
@@ -0,0 +1,63 @@
...
+#include <LibCore/get_password.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <termios.h>
+
+namespace Core {
+
+Result<String, int> get_password(const StringView& prompt)
+{
+    fputs(prompt.characters_without_null_termination(), stdout);
```
awesomekling, line 36: The name `characters_without_null_termination()` is meant to be a clue in situations like this.
If you're not passing the length along with it, it's basically unsafe usage, since it's indeed not guaranteed to be null-terminated :)

Libraries/LibCore/get_password.cpp
```diff
@@ -0,0 +1,63 @@
...
+    struct termios original;
+    tcgetattr(STDIN_FILENO, &original);
+
+    struct termios no_echo = original;
+    no_echo.c_lflag &= ~ECHO;
+    if (tcsetattr(STDIN_FILENO, TCSAFLUSH, &no_echo) < 0) {
+        return errno;
+    }
+
+    char* password = NULL;
```
awesomekling, line 48: `NULL` -> `nullptr`

Userland/passwd.cpp
```diff
@@ -0,0 +1,144 @@
...
+#include <LibCore/ArgsParser.h>
+#include <LibCore/File.h>
+#include <LibCore/get_password.h>
+#include <pwd.h>
+#include <string.h>
+#include <unistd.h>
+
+static String get_salt()
+{
+    auto random_data = Core::File::open("/dev/random", Core::IODevice::OpenMode::ReadOnly).value()->read(12);
```
awesomekling, line 39: Please use `arc4random_buf(buffer, size)` instead of reading from `/dev/random` :)

mattco98:
I had the same build error along the way in #2861, which I managed to fix. The PR just got merged, so try to rebase onto master and see if that fixes the build.

AK/String.cpp
Petelliott: I've updated the commit message, and added additional tests for `String::split()` and `String::Substring()`

Petelliott:
> I had the same build error along the way in #2861, which I managed to fix. The PR just got merged, so try to rebase onto master and see if that fixes the build.

unfortunately that didn't work.

Libraries/CMakeLists.txt
```diff
@@ -2,6 +2,7 @@ add_subdirectory(LibAudio)
 add_subdirectory(LibC)
 add_subdirectory(LibCore)
 add_subdirectory(LibCrypto)
+add_subdirectory(LibCrypt)
```
BenWiederhake, line 5: "LibC" comes before "LibCore". This is also the usual way of how dictionaries are sorted: prefix first, then the longer words. If we want to follow this principle, "LibCrypt" should come before "LibCrypto".

Petelliott:
Turns out I used `bool` in an `extern "C"` header, which worked on my machine but not travis. I've changed it to an int.

Libraries/LibCrypt/crypt.cpp
```diff
@@ -0,0 +1,78 @@
...
+}
+
+static constexpr size_t crypt_salt_max = 16;
+static constexpr size_t sha_string_length = 44;
+
+char* crypt_r(const char* key, const char* salt, struct crypt_data* data)
+{
+    if (!data->initialized) {
+        errno = EINVAL;
+        return NULL;
```
awesomekling, line 49: `nullptr`

Libraries/LibCrypt/crypt.cpp
```diff
@@ -0,0 +1,78 @@
...
+
+            strncpy(data->result + header_len + 1, string.characters(), sha_string_length);
+
+            return data->result;
+        }
+    }
+
+    // DES crypt is not available.
+    errno = EINVAL;
+    return NULL;
```
awesomekling, line 76: `nullptr`

Libraries/LibCrypt/crypt.cpp
Petelliott: sorry I missed this so many times.

Petelliott:
> but isn't libcrypt some GNU thing?

- `crypt(3)` is a part of posix, but is optional due to export regulations.
- The `$id$salt` is an extension, but is used by Glibc, FreeBSD, and OpenBSD.
- `crypt_r(3)` is a GNU extension, but is included in FreeBSD.
- `libcrypt` is used by GNU, and FreeBSD, whereas OpenBSD has `libcrypto`

I think providing `cypt` is important because unix passwords are expected to be hashed with `crypt`.
I don't think we should put `crypt` in `LibC`, because it would make `LibC` depend on `LibCrypto`.
We could merge `LibCrypt` and `LibCrypto`, and give `LibCrypto` the aditional name of crypt, but i'm not quite sure if we would need to `serenity_libc(LibCrypto)`, and I don't know how that would work.

awesomekling:
Okydoky. Great work :)

PR #2882, opened 2020-07-26
LibCrypto: CTR mode, test vectors, and marginals

BenWiederhake:
This PR handles the following issues, in descending priority:
- "Fix" CTR mode (see below)
- Implement test vectors for CTR mode
- Remove a "FIXME" that is already fixed
- Touch LibJS to allow building Lagom with gcc 9.3.0

Technically, CTR already "worked": It was only used in the Kernel as PRNG. In this function, it was already successful. However, because of the unusual counter it used, no tests could be written. As I went down the rabbit hole of counter variation and the surprising lack of consensus on how counting works (!!!), I also documented it in `CTR.h` and explained why this new counter should work in most cases.

> while you're at it, mind making the increment function a template parameter? you could default it to the existing increment()

I decided against this, because it seems too much of a hassle, and currently we don't need it.


BenWiederhake:
For documentation: This was the error message I used to get with LibJS:

```
/home/user/workspace/serenity/Libraries/LibJS/AST.cpp: In member function ‘virtual void JS::ClassExpression::dump(int) const’:
/home/userworkspace/serenity/Libraries/LibJS/AST.cpp:933:11: error: ‘%s’ directive argument is null [-Werror=format-overflow=]
  933 |     printf("ClassExpression: \"%s\"\n", m_name.characters());
      |     ~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/user/workspace/serenity/Libraries/LibJS/AST.cpp:933:11: error: ‘%s’ directive argument is null [-Werror=format-overflow=]
cc1plus: all warnings being treated as errors
```

awesomekling:
Please wrap commit message text at 72 characters so they fit in with all the other commit messages :)

BenWiederhake:
Sure, I would do that.

I'm not sure which commit you mean though: The longest line in this PR is commit 2b0e7e2aaa51bbc66f3ff4bda1c1a0b2964e33db, `LibCrypto: Fix broken CTR mode, implement RFC 3686 Test Vectors`, with 63 characters.

Or do you mean the code itself? I loosely aimed at 80 characters per line, and I believe I never went beyond 100 characters per line.

awesomekling:
This one:

```
LibCrypto: Already using strong crypto

ModularFunctions::random_number calls into
AK::fill_with_random calls (on Serenity) into
arc4random_buf calls into
Process::sys calls into
get_good_random_bytes,
which is cryptographically secure.
```

BenWiederhake:
Oh! I thought this makes it easier to see that it's supposed to be a call stack. Sure, I'll just re-wrap it.

Libraries/LibCrypto/Cipher/Mode/CTR.h
```diff
@@ -92,7 +93,7 @@ class CTR : public Mode<T> {
 public:
     constexpr static size_t IVSizeInBits = 128;
 
-    virtual ~CTR() { }
+    virtual ~CTR() {}
```
alimpfard, line 96: This looks wrong, are you using clang-format 10?

Libraries/LibCrypto/Cipher/Mode/CTR.h
```diff
@@ -162,7 +162,7 @@ class CTR : public Mode<T> {
         // FIXME: We should have two of these encrypt/decrypt functions that
         //        we SFINAE out based on whether the Cipher mode needs an ivec
         ASSERT(ivec.has_value());
-        auto iv = ivec.value();
+        auto iv = ByteBuffer(ivec.value());
```
alimpfard, line 167: stylistic decision, your choice:
`ByteBuffer iv = ivec.value()` would do the same

Libraries/LibCrypto/Cipher/Mode/CTR.h
```diff
@@ -95,10 +95,14 @@ class CTR : public Mode<T> {
...
+    // Must intercept `Intent`, because AES must always be set to
+    // Encryption, even when decrypting AES-CTR.
+    // TODO: How to deal with other ciphers?
     template<typename... Args>
-    explicit constexpr CTR<T>(Args... args)
-        : Mode<T>(args...)
+    explicit constexpr CTR<T>(const ByteBuffer& user_key, size_t key_bits, Intent intent = Intent::Encryption, Args... args)
+        : Mode<T>(user_key, key_bits, args...)
     {
+        (void)intent;
```
alimpfard, line 105: You can just say
```c++
explicit constexpr CTR<T>(const ByteBuffer& user_key, size_t key_bits, Intent = Intent::Encryption, Args... args)
```
and get rid of the `(void)intent;`.

Libraries/LibCrypto/Cipher/Mode/CTR.h
BenWiederhake: I did indeed! Thanks, this also removes all other annoying false-positives.

Libraries/LibCrypto/Cipher/Mode/CTR.h
BenWiederhake: I was defensive here, because we change the ivec later and I wasn't sure whether we're allowed to modify it.

I overlooked that we actually get a *copy* of the `Optional`, not a reference, so we're free to modify the ivec as we please. I'll change the code to save those precious nanoseconds :)

Libraries/LibCrypto/Cipher/Mode/CTR.h
BenWiederhake: Lovely! I wasn't aware that C++ can do that.

PR #2884, opened 2020-07-26
Implement AK::Span which is like std::span and represents a contiguous sequence of objects.

asynts:
There are functions in the code base that have signatures like `void foo(int* values, size_t count)`. With `AK::Span` these parameter pairs could be replaced with a single parameter, for example `void foo(Span<int> values)`.

I added a commit that changes the signature of `AK::binary_search` to use this new class. I wasn't sure if it would be a good idea to dig though the code base and start changing a ton of function signatures to use this new class, but this should at least demonstrate the usage of this class.

I can edit this pull request or create follow up pull requests that change some code to use this class, especially `AK::ByteBuffer` could benefit from this.

---

There is one drawback with this class, the following won't compile:

~~~c++
Span<int> foo(void* values, size_t count) {
    // error: invalid conversion from 'void*' to 'int*'
    return { values, count };
}
~~~

There are a few places where this could be problematic, but these could be fixed using `reinterpret_cast` or by propagating `Span` further.

### DIFF
--- a/Userland/allocate.cpp
+++ b/Userland/allocate.cpp
@@ -37,14 +37,14 @@ void usage(void)
     exit(1);
 }
 
-enum Unit { Bytes,
+enum class Unit { Bytes,
     KiloBytes,
     MegaBytes };
 
 int main(int argc, char** argv)
 {
     int count = 50;
-    Unit unit = MegaBytes;
+    auto unit = Unit::MegaBytes;
 
     if (argc >= 2) {
         auto number = String(argv[1]).to_uint();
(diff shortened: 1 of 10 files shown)


Userland/allocate.cpp
```diff
@@ -37,14 +37,14 @@ void usage(void)
     exit(1);
 }
 
-enum Unit { Bytes,
+enum class Unit { Bytes,
```
awesomekling, line 40: These seem like completely arbitrary unrelated changes :)

asynts: I changed this because of a naming collision with `using AK::Bytes = Span<u8>`.

https://github.com/SerenityOS/serenity/pull/2884/files#diff-d0f68934dcacd7f05ac0792071e08ad6R123-R124

awesomekling: A-ha! Roger that. This looks very nice :)

asynts:
@awesomekling Is it good if I create another pull request that changes other code to use `AK::Span`? I don't want to be the guy that shows up and then starts refactoring everything if that isn't welcome.

awesomekling:
Yes, that sounds great, as long as it's tasteful. :)

PR #2888, opened 2020-07-26
LibGfx: Templatize Point, Size, and Rect

mattco98:
These classes were just begging to be templatized :)


Applications/PixelPaint/EllipseTool.cpp
```diff
@@ -94,8 +94,8 @@ void EllipseTool::on_second_paint(const Layer& layer, GUI::PaintEvent& event)
 
     GUI::Painter painter(*m_editor);
     painter.add_clip_rect(event.rect());
-    auto preview_start = m_editor->layer_position_to_editor_position(layer, m_ellipse_start_position).to_int_point();
-    auto preview_end = m_editor->layer_position_to_editor_position(layer, m_ellipse_end_position).to_int_point();
+    auto preview_start = m_editor->layer_position_to_editor_position(layer, m_ellipse_start_position);
+    auto preview_end = m_editor->layer_position_to_editor_position(layer, m_ellipse_end_position);
```
awesomekling, line 98: This change makes conversion implicit rather than explicit. That feels like a separate thing not really related to templatizing the types(?)

mattco98: I just thought it was convenient, but I can definitely make the conversion constructors explicit

mattco98: Now all three of the classes have an explicit conversion constructor, as well as a nice `to_type<T>()` conversion helper.

awesomekling: Thanks! The explicit conversion is important IMO, at least in the float->int direction since it's lossy.

PR #2898, opened 2020-07-27
SystemMenu: Add icons to top-level

BenWiederhake:
In this PR:
- Make LibGUI able to display icons for submenus
- Use that feature in `SystemMenu` with some already-existing icons
- Refactor/cleanup along the way (but still in separate commits)
- Add new icons for "Themes" (uses the Default, Redmond 2000, and Xmas theme) and "Eyes" (which is basically just a screenshot)

I'll probably make another PR soon that moves files like `/res/icons/gear16.png` to more fitting places, in this case `/home/eispin/workspace/serenity/Base/res/icons/16x16/gear.png`, I just didn't want to force it into this PR.


BenWiederhake:
And because I can: Here's a sneak peek into how it looks with this PR:

icons-screenshot

BenWiederhake:
(Removed the "Eyes" icon commit. It should go into the next PR.)

awesomekling:
Please avoid using generic words like "refactor" without further explanation in commit messages.
(From the contribution guidelines):
> **Don't**
> * Use weasel-words like "refactor" or "fix" to avoid explaining what's being changed.

BenWiederhake:
You're right, I tried to pin down the metaphorical weasel:

- `LibGUI: Refactor icon realization in Menu`: I still use the word "Refactor", because even in arithmetic this operation would be called refactoring: I take a factor (piece of code), and pull it out (make it a separate function). I hope the commit description is sufficient.
- `SystemMenu: Untangle app/category discovery and GUI building` works indeed better without the word "refactor".

awesomekling:
> You're right, I tried to pin down the metaphorical weasel:
> 
> * `LibGUI: Refactor icon realization in Menu`: I still use the word "Refactor", because even in arithmetic this operation would be called refactoring: I take a factor (piece of code), and pull it out (make it a separate function). I hope the commit description is sufficient.
> * `SystemMenu: Untangle app/category discovery and GUI building` works indeed better without the word "refactor".

Excellent! And yeah, the point is not to avoid using those words, but to avoid the common antipattern of using *only* those words to describe a change. :)

PR #2899, opened 2020-07-27
Shell: Ignore leading semicolons

alimpfard:
This makes things like `foo&; bar` behave as expected.
Such behaviour is actually closer to the grammar defined in `Parser.h` anyway :P


awesomekling:
Great!

PR #2903, opened 2020-07-27
Propagate the use of AK::Span though the codebase.

asynts:
There is this "anti-pattern" in the code base where functions take a `const ByteBuffer&` as input. Even if the input is immutable, it has to be put into a mutable buffer. This is usually done using `ByteBuffer::wrap(const void*, size_t)` which does a `const_cast` internally. The caller has to keep track that such buffers must not be written to (no copy on write!).

This pull request first fixes a few issues with the implementation of `Span` and then changes a few functions to take `ReadonlyBytes` instead of `const ByteBuffer&` or `Bytes` instead of `ByteBuffer&`.

I've mostly focused on eliminating calls to that `wrap` overload, but it can't quite be removed yet, because `BufferStream` has no immutable version.

### DIFF
--- a/AK/Base64.cpp
+++ b/AK/Base64.cpp
@@ -86,7 +86,7 @@ ByteBuffer decode_base64(const StringView& input)
     return ByteBuffer::copy(output.data(), output.size());
 }
 
-String encode_base64(const ByteBuffer& input)
+String encode_base64(ReadonlyBytes input)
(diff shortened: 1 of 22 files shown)


awesomekling:
These are great changes and I'm glad to see that crappy pattern finally getting fixed. :)

PR #2907, opened 2020-07-28
Shell: Fix parse mistake in '&&' not being recursive

alimpfard:
This fix brought to you by ~~Raid: Shadow Legends~~ the sudden split of #2824

Try:
```sh
$ foo && bar && baz
```

### DIFF
--- a/AK/Tests/TestString.cpp
+++ b/AK/Tests/TestString.cpp
@@ -87,8 +87,8 @@ TEST_CASE(starts_with)
     EXPECT(!test_string.starts_with('B'));
     EXPECT(test_string.starts_with("ABCDEF"));
     EXPECT(!test_string.starts_with("DEF"));
-    EXPECT(test_string.starts_with("abc",  CaseSensitivity::CaseInsensitive));
-    EXPECT(!test_string.starts_with("abc",  CaseSensitivity::CaseSensitive));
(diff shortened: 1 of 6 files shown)


alimpfard:
And yet...this one turned into a mess as well...
:facepalm:

AK/StringView.cpp
```diff
@@ -95,7 +95,7 @@ Vector<StringView> StringView::split_view(const StringView& separator, bool keep
         view = view.substring_view_starting_after_substring(part_with_separator);
         maybe_separator_index = view.find_first_of(separator);
     }
-    if (keep_empty || !view.is_empty())
+    if (!view.is_empty())
```
awesomekling, line 98: Does any of the other split functions have this same problem?
* `String::split()`
* `String::split_view()`
Also please add a test :)

alimpfard: Thinking a little about this, I am a little lost on whether this is actually a correct behaviour, should `"axbx".split_view('x', true)` be `"a", "b"` or `"a", "b", ""`?

I can think of reasons of both being correct, so I instinctively reach for keeping it as-is...

awesomekling: The idea with `keep_empty` is that no empty strings will appear in the output.

alimpfard: Then I'll revert my change and just add some tests.

PR #2909, opened 2020-07-28
Userland: Fix nc by not memset()'ing the input address char*

### LINKED ISSUE #2908: nc: Does not work anymore
To reproduce, start a WebServer e.g `ws .` and then try connecting to it with `nc localhost 8000`

cc @deoxxa

linusg:
We were accidentally calling `memset()` on "`addr`" (the input `char*`), not "`dst_addr`" (the target `struct sockaddr_in`), which was causing a simple "`nc localhost 8000`" to crash.

Fixes #2908.

### DIFF
--- a/Userland/nc.cpp
+++ b/Userland/nc.cpp
@@ -133,7 +133,7 @@ int main(int argc, char** argv)
         char addr_str[100];
 
         struct sockaddr_in dst_addr;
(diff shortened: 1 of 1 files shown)


awesomekling:
Nice. Straight from the "how did this ever work" department :)

deoxxa:
Ohhhh whoops!

PR #2910, opened 2020-07-28
ProcessChooser: give Inspector a choice

thankyouverycool:
Lets Inspector use the now general purpose ProcessChooser and adds new icons for Profiler and Inspector to pass along to it.


awesomekling:
:star:

PR #2911, opened 2020-07-28
LibX86: Disassemble most FPU instructions starting with D9

nico:
Some of these don't just use the REG bits of the mod/rm byte
as slashes, but also the R/M bits to have up to 9 different
instructions per opcode/slash combination (1 opcode requires
that MOD is != 11, the other 8 have MODE == 11).

This is done by making the slashes table two levels deep for
these cases.

Some of this is cosmetic (e.g "FST st0" has no effect already,
but its bit pattern gets disassembled as "FNOP"), but for
most uses it isn't.

FSTENV and FSTCW have an extraordinary 0x9b prefix. This is
not yet handled in this patch.


Libraries/LibX86/Instruction.cpp
```diff
@@ -199,6 +200,28 @@ static void build_slash(InstructionDescriptor* table, u8 op, u8 slash, const cha
     build(d.slashes, slash, mnemonic, format, handler, lock_prefix_allowed);
 }
 
+static void build_slash_rm(u8 op, u8 slash, u8 rm, const char* mnemonic, InstructionFormat format, InstructionHandler handler, IsLockPrefixAllowed lock_prefix_allowed = LockPrefixNotAllowed)
+{
+    ASSERT((rm & 0xc0) == 0xc0);
+    ASSERT(((rm >> 3) & 7) == slash);
+    InstructionDescriptor* table = s_table32;  // FIXME: Is there a 16-bit chip with FPU instrs?
```
awesomekling, line 207: `s_table16` vs `s_table32` is not about the chip type, but rather which operand size is currently in effect.
`mov ax, bx` and `mov eax, ebx` are the same opcode, just with different operand sizes.
UserspaceEmulator currently runs with 32-bit operands by default, but an operand size override prefix (`0x66`) turns it into a 16-bit op and then we look in `s_table16` instead. :)

Libraries/LibX86/Instruction.cpp
```diff
@@ -199,6 +200,28 @@ static void build_slash(InstructionDescriptor* table, u8 op, u8 slash, const cha
...
+
+    InstructionDescriptor& d0 = table[op];
+    ASSERT(d0.format == MultibyteWithSlash);
+    InstructionDescriptor& d = d0.slashes[slash];
+
+    if (!d.slashes) {
+        // Slash/RM instructions are not always dense, so make them all default to the slash instruction.
+        d.slashes = new InstructionDescriptor[8];
+        for (int i = 0; i < 8; ++i) {
```
awesomekling, line 216: c l a n g f o r m a t

Libraries/LibX86/Instruction.cpp
nico: Is there a way I can test this (other than writing a machine code file by hand, I suppose)? I tried `BITS 16` in nasm, but that prefixes x87 instructions that point to memory with 0x67, not 0x66, and it confuses objdump, so maybe nasm doesn't do it right.

Libraries/LibX86/Instruction.cpp
nico: done.

Libraries/LibX86/Instruction.cpp
awesomekling: You can always force out a prefix by just doing `db 0x66`

nico: Ah, right, thanks. Well, it seems to work :)

PR #2912, opened 2020-07-28
Meta: Calculate image size based on size of Build/Root and Base

Petelliott:
This reduces the size of the default build, while allowing people to install as many ports as they want, without having to manually specify disk size.


awesomekling:
Great idea :)

PR #2914, opened 2020-07-28
Icons: Improve contrast, file permissions, Eyes

BenWiederhake:
In this PR:
- Improve contrast of several icons that I believe now look better (see below)
- Fix file permissions. Somehow many icons end up with executable permissions set. This PR also includes a linter that checks for "executable" icons, so this won't be an issue in the future anymore :)
- Add icons and an About dialog for "Eyes". (Doesn't quite fit in this PR, I know, but at least it's "all about icons".)
- Add a lint that tries to find missing resources. It has a few false positives, so I didn't include it in `.travis.yml`. It also found some actual issues (true positives) which I fixed.

EDIT: Forgot the second lint script, sorry.

Here's a demonstration of the icons that changed. (The screenshots are from a few days ago, that's why thankyouverycool's beautiful icons for Profiler and Inspector are missing.)

comparison


awesomekling:
These are just great. :)

PR #2916, opened 2020-07-28
LibGui+WindowServer: Offset applications that are opened directly on eachother

Petelliott:
If you opened many terminals before they would be right on top of eachother, and it didn't look like anything happened.
This change should make it more obvious when you application opens, and make it easier to grab the old windows.

example:
2020-07-28-155650_3200x1800_scrot

### DIFF
--- a/Applications/Browser/main.cpp
+++ b/Applications/Browser/main.cpp
@@ -121,7 +121,8 @@ int main(int argc, char** argv)
     auto bookmarks_bar = Browser::BookmarksBarWidget::construct(Browser::bookmarks_filename, bookmarksbar_enabled);
 
     auto window = GUI::Window::construct();
-    window->set_rect(100, 100, 640, 480);
+    window->move_to_recommended_position(100, 100);
+    window->resize(640, 400);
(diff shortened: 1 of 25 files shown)


awesomekling:
It would be nice if this could just happen automagically instead of apps having to use an API for it.
How about this:
- All new windows start out in a special "WindowServer will choose the position" mode.
- WindowServer clients can override this mode by explicitly setting a position before showing the window.
- Most apps only specify initial window size instead of initial window rect.

Petelliott:
How do you suggest WindowServer chooses the initial position?

Some ideas:
- Randomly. This will basically eliminate the purpose of this PR, but have the same desired effect.
- Based on a default for all apps, with an api to request a recommended position
- Based on a hash of the application, so the same app always opens at the same Location.

I personally like the hash one.

awesomekling:
Another idea:
- At an offset from the last window that was opened with unspecified position and has not been moved by the user.

Petelliott:
Sounds good. I'm going to make a new PR, because not much functionality is shared with this one

PR #2917, opened 2020-07-28
Shave off a few minutes of every single build (starting next week)

BenWiederhake:
6.5 minutes of each build is just network traffic due to the large cache. This cuts down the size of each future item in the Toolchain cache. As the Toolchain cache rotates, those 6.5 minutes should gradually go down to about 2 minutes.

This is not a lot, but it adds up. It should become especially noticable when the Toolchain doesn't need to be rebuilt.

Avenues I didn't explore yet but plan to do so in future PRs:
- Trying to cut down the list of "dependencies", so we can re-use a Toolchain more often.
- Figuring out why building the Toolchain takes double the time with gcc10 than with gcc9.

### DIFF
--- a/Toolchain/BuildIt.sh
+++ b/Toolchain/BuildIt.sh
@@ -19,9 +19,10 @@ MAKE="make"
 MD5SUM="md5sum"
 NPROC="nproc"
 
-# Each cache entry is 260 MB. 8 entries are 4 GiB.
-# It seems that Travis starts having trouble at 35 entries, so I think this is a good amount.
-KEEP_CACHE_COUNT=8
+# Each cache entry is 70 MB. 10 entries are 700 MiB.
+# It seems that Travis starts having trouble around a total
+# cache size of 9 GiB, so I think this is a good amount.
+KEEP_CACHE_COUNT=10
 
 if command -v ginstall &>/dev/null; then
     INSTALL=ginstall
@@ -110,7 +111,6 @@ pushd "$DIR"
                 ls -l
             popd
         fi
-
     fi
 popd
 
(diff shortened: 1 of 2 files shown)


awesomekling:
It's hilarious that we've been using a 280MB compiler this whole time.

Toolchain/BuildIt.sh
```diff
@@ -275,6 +277,23 @@ pushd "$DIR"
...
             mkdir -p Cache/
+            # We *most definitely* don't need debug symbols in the linker/compiler.
+            # This cuts the uncompressed size from 1.2 GiB per Toolchain down to about 250 MiB.
+            pushd "Local/libexec/gcc/i686-pc-serenity/${GCC_VERSION}"
+                for binary in cc1plus lto1; do
+                    echo "Before: $(du -h "${binary}")"
+                    strip "${binary}"
+                    echo "After: $(du -h "${binary}")"
+                done
+                # C is forbidden anyway by the Contribution Guidelines, so we can delete the C compiler:
```
nico, line 288: What about Libraries/LibCore/puff.c?

BenWiederhake: Argh! Sorry, we should have waited on Travis to complete.

I see there's already efforts going on to convert puff to C++. If that works: Great!

In case that doesn't work or is too much effort, `BuildIt.sh` doesn't need to delete `cc1`, it's also fine if we just trip it instead. In order to do so delete the handful of lines that `rm cc1`, and add it to the loop around `strip`.

I can't make the commits myself right now because I'm on my work laptop, sorry :/

nico: Now that puff.c is renamed, it complains about cc1's lack when building boot.S.

PR #2918, opened 2020-07-29
LibX86: Disassemble most remaining FPU instructions

nico:
A few of the remaining instructions have different behavior for
register and non-register ops.  Since we already have the
two-level flags tables, model this by setting all handlers in
the two-level table to the register op handler, while the
first-level flags table stores the action for the non-reg handler.


nico:
Looks like the red CI is due to @BenWiederhake deleting cc1.

Meta/Lagom/CMakeLists.txt
```diff
@@ -1,7 +1,7 @@
 cmake_minimum_required (VERSION 3.0)
 project (Lagom)
 
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -O2 -Wall -Wextra -Werror -std=c++2a -fPIC -g -Wno-deprecated-copy")
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -O2 -Wall -Wextra -Werror -std=c++2a -fPIC -g -Wno-deprecated-copy -Wno-format-overflow")
```
awesomekling, line 4: ?

nico:
Ok, now the failing tests are just the crypto tests as far as I can tell.

alimpfard:
> crypto tests

_/me cries in randomness_

Issue #362, opened 2019-07-23
Implement zlib compression/uncompression


awesomekling:
The code in `Libraries/LibDraw/puff.c` is a single-file implementation of **zlib**
As part of our NIH efforts, we should consider writing our own version of this. :)

This is currently only used by `PNGLoader` but there are more places where we'll want this.

IdanHo:
this can probably be closed now

PR #2922, opened 2020-07-29
LibCrypto: Add some checksum algorithms

stelar7:
Namely CRC32 and Adler32

Does some work towards #362 by implementing some of the sub-algorithms used.

### DIFF
--- a/Libraries/LibCrypto/Checksum/Adler32.h
+++ b/Libraries/LibCrypto/Checksum/Adler32.h
@@ -1,0 +1,58 @@
+/*
+ * Copyright (c) 2020, the SerenityOS developers
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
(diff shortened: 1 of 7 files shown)


Libraries/LibCrypto/Checksum/CRC32.h
```diff
@@ -0,0 +1,68 @@
...
+
+        table[i] = value;
+    }
+
+    return table;
+}
+
+static auto crc_table = generate_crc_table();
+
+uint32_t crc32_update(uint32_t initial_value, ReadonlyBytes data)
```
alimpfard, line 54: Same old thing about making this a class and having it keep the value as a member.
Doing so would allow successive updates without knowing the implementation details of this specific checksum function.

alimpfard:
__and also__, there's no particular reason to keep these as header-only, that's really only used when the class itself is templated.
so please move the implementations to cpp files.

Libraries/LibCrypto/Checksum/CRC32.h
```diff
@@ -0,0 +1,68 @@
...
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#pragma once
+
+#include <AK/Span.h>
+
+static auto generate_crc_table()
+{
+    auto table = new uint32_t[256];
```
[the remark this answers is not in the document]
stelar7: I tried doing this, but i couldn't find a way to return an array from the generate function.. 
Maybe some smarter people than me can show a fix :)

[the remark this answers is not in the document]
stelar7: I could'nt make it accept an array as the type, so its a struct instead now.

Libraries/LibCrypto/Checksum/Adler32.h
```diff
@@ -0,0 +1,58 @@
...
+        , state_b(initial_b)
+    {
+        update(data);
+    }
+
+    void update(ReadonlyBytes data);
+    u32 digest();
+
+private:
+    u32 state_a { 1 };
```
alimpfard, line 54: `u32 m_state_a`

Userland/test-crypto.cpp
```diff
@@ -1558,6 +1596,50 @@ void tls_test_client_hello()
     loop.exec();
 }
 
+int adler32_tests()
+{
+    auto do_test = [](ReadonlyBytes input, u32 expected_result) {
+        I_TEST();
```
alimpfard, line 1602: Test what?
Please at the very least say that it's testing Adler32: `I_TEST((Adler32));`

Userland/test-crypto.cpp
```diff
@@ -1558,6 +1596,50 @@ void tls_test_client_hello()
...
+    do_test(String("message digest").bytes(), 0x29750586);
+    do_test(String("abcdefghijklmnopqrstuvwxyz").bytes(), 0x90860b20);
+
+    return g_some_test_failed ? 1 : 0;
+}
+
+int crc32_tests()
+{
+    auto do_test = [](ReadonlyBytes input, u32 expected_result) {
+        I_TEST();
```
alimpfard, line 1625: Same here.

Libraries/LibCrypto/Checksum/CRC32.h
```diff
@@ -0,0 +1,85 @@
...
+        : state(initial_state)
+    {
+        update(data);
+    }
+
+    void update(ReadonlyBytes data);
+    u32 digest();
+
+private:
+    u32 state { ~0u };
```
alimpfard, line 82: `u32 m_state`

Libraries/LibCrypto/Checksum/CRC32.h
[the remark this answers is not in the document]
alimpfard: That's fine, I think.

Though the compiler should've been ok with just an array...:thinkst:

alimpfard:
Took me a while to figure out why travis is sad...CRLF.

Should we make `Meta/check-style.sh` accept CRLF too? (or warn about it if not?) @linusg [that was your handiwork] + @awesomekling [style referece :P]

linusg:
> Should we make `Meta/check-style.sh` accept CRLF too?

nooooo :(

Can't Windows users just set `git config core.autocrlf true`? (https://docs.github.com/en/github/using-git/configuring-git-to-handle-line-endings#global-settings-for-line-endings)

> Configure Git to ensure line endings in files you checkout are correct for Windows.
> For compatibility, line endings are converted to Unix style when you commit files.

alimpfard:
> nooooo :(

Yeah, I'm not too enthusiastic about that either.
Still though, the script should probably be like "hey pal, this is using CRLF, use the sane line endings please" instead of "this is missing a license header" and the license header is visibly there.

> autocrlf

TIL that's an option. very neat.

PR #2923, opened 2020-07-29
WindowFrame: Convert button bitmaps to themable PNGs

thankyouverycool:
One step closer to Microsoft Corp. v. Kling 👍
changetitlebuttons


awesomekling:
Let's just use PNG images. I used CharacterBitmap early on before there was any PNG support, not because it's somehow better :D

PR #2924, opened 2020-07-30
LibWeb: Complete the redirect URL before loading it

alimpfard:
the "Location" header is allowed to be a relative URL (as is the case in our very own WebServer!)


awesomekling:
Great catch!

PR #2928, opened 2020-07-30
Userland: Add basic `lsof' command

sppmacd:


Userland/lsof.cpp
```diff
@@ -0,0 +1,152 @@
...
+    }
+
+    if (unveil("/proc", "r") < 0) {
+        perror("unveil /proc");
+        return 1;
+    }
+
+    // needed by ProcessStatisticsReader::get_all()
+    if (unveil("/etc/passwd", "r") < 0) {
+        perror("unveil /proc");
```
nico, line 91: typo

Userland/lsof.cpp
```diff
@@ -0,0 +1,152 @@
...
+        return 1;
+    }
+
```
nico, line 94: You want to call `unveil(nullptr, nullptr);` after you're done setting up veiled paths, to lock the veil.

PR #2929, opened 2020-07-31
LibCrypto: Do not trim leading zeros in export_data by default

### LINKED ISSUE #2691: LibCrypto: Sporadic test failure

alimpfard:
This fixes the issue with the exported data having a leading zero, causing RSA::encrypt to trim the block down, and ruining the encryption.

Fix #2691 :^)

---
I am not entirely certain if this definitely fixes the issue, as random things are...random.
However, a local loop on the breaking test (`RSA PKCS#1 v1.5`) successfully passed about 1000 times before I got bored.
For reference, that takes about two hours.

### DIFF
--- a/Libraries/LibCrypto/BigInt/SignedBigInteger.cpp
+++ b/Libraries/LibCrypto/BigInt/SignedBigInteger.cpp
@@ -36,11 +36,15 @@ SignedBigInteger SignedBigInteger::import_data(const u8* ptr, size_t length)
     return { move(unsigned_data), sign };
 }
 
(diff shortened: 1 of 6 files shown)


awesomekling:
>The following tests FAILED:
>	  1 - Crypto (Failed)
:crying_cat_face:

alimpfard:
This PR fixes the sporadic failure by making it a consistent failure.
definitely worth it.

PR #2930, opened 2020-07-31
Ports: Fix binutils and gcc ports

itamar8910:
I don't actually know what lead to their sudden breakage, but this fixes them.

### DIFF
--- a/Kernel/Syscalls/sysconf.cpp
+++ b/Kernel/Syscalls/sysconf.cpp
@@ -24,6 +24,7 @@
  * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
+#include <Kernel/Arch/i386/CPU.h>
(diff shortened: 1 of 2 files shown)


awesomekling:
Hm, that's weird, I just built these ports last night and they worked fine.
Regardless, my instinct says this may have something to do with adding `sysconf()` recently.
Perhaps gcc/binutils is trying to get the page size from `sysconf(_SC_PAGESIZE)` but that doesn't work so it falls back to its own `getpagesize()` for whatever reason.

itamar8910:
@awesomekling, yep you had it right. They checked whether `_SC_PAGESIZE` was defined, and since it wasn't they brought their own `getpagesize`.

awesomekling:
Nice!

nico:
After the enum addition, _SC_PAGESIZE still isn't defined. If the port is still broken, we also need a `#define _SC_PAGESIZE _SC_PAGESIZE` I suppose.

PR #2931, opened 2020-07-31
AK: On failed mmap, don't claim that there's a readable page

BenWiederhake:
Behavior before:

```
 ./disasm --help
open: No such file or directory
Data is at 0xffffffffffffffff, there's 4096 bytes.
Segmentation fault
[$? = 139]
```

That's because the argument `--help` was interpreted as a filename, and the constructor of `MappedFile` sets `m_size` to a page size, even though it already is correctly initialized to 0.

Even more generally, `disasm` forgot to check whether the mmap'd file `.is_valid()`, and now exits with an error before any access is attempted.

Behavior now:
```
$ ./disasm --help
open: No such file or directory
[$? = 1]
```

(In case you're wondering about unzip: That only crashes when there's a race between unzip's `stat()`, the file deletion, and unzip's `MappedFile()`, so I can't easily reproduce that bug.)

### DIFF
--- a/AK/MappedFile.cpp
+++ b/AK/MappedFile.cpp
@@ -24,8 +24,8 @@
  * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
-#include <AK/String.h>
 #include <AK/MappedFile.h>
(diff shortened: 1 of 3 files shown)


awesomekling:
What a silly bug :)

PR #2937, opened 2020-08-01
AK: Use C++20 concepts to only allow Userspace wrappers of pointers

emanuele6:
It was a bit odd that you could create a `Userspace<int>` and that `Userspace<int>::ptr()` returned an `int` instead of an `int*`.
    
Let's use C++20 concepts to only allow creating `Userspace` objects with pointer types. :^)

We could also rewrite `Userspace<T>` as something like `UserspacePtr<T>` and make it store `m_ptr` (etc.) as a `T*`, but since we are using `Userspace` to also wrap const-pointers and void-pointers we would need to make separate classes for those (something like `ConstUserspacePtr<T>` and `VoidUserspacePtr`).

The C++20 concepts solution is definitely cleaner.

-----------------------------
I also replaced `#if defined(KERNEL)` with `#ifdef KERNEL` in `Userspace.h`.

### DIFF
--- a/AK/Userspace.h
+++ b/AK/Userspace.h
@@ -42,7 +42,7 @@ public:
     operator bool() const { return m_ptr; }
     operator FlatPtr() const { return (FlatPtr)m_ptr; }
 
-#if defined(KERNEL)
+#ifdef KERNEL
     Userspace(FlatPtr ptr)
         : m_ptr(ptr)
     {
@@ -60,7 +60,7 @@ public:
 #endif
 
 private:
-#if defined(KERNEL)
+#ifdef KERNEL
     FlatPtr m_ptr { 0 };
 #else
     T m_ptr { nullptr };
--- a/AK/Userspace.h
+++ b/AK/Userspace.h
@@ -26,11 +26,15 @@
 
 #pragma once
 
+#include <AK/StdLibExtras.h>
 #include <AK/Types.h>
 
 namespace AK {
 
 template<typename T>
+concept PointerTypeName = IsPointer<T>::value;
+
+template<PointerTypeName T>
 class Userspace {
 public:
     Userspace() { }
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -40,7 +40,7 @@ add_custom_target(check-style
     USES_TERMINAL
 )
 
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra -Werror -std=c++2a -fdiagnostics-color=always")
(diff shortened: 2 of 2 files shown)


AK/Userspace.h
```diff
@@ -27,10 +27,14 @@
 #pragma once
 
 #include <AK/Types.h>
+#include <type_traits>
```
awesomekling, line 30: Please don't use STL headers. Instead, use `IsPointer` from AK/StdLibExtras.h :)

emanuele6: Oh, ok. I didn't know that existed. I will do it now.

emanuele6:
```log
In file included from /home/travis/build/SerenityOS/serenity/./Kernel/API/Syscall.h:30,
                 from /home/travis/build/SerenityOS/serenity/AK/SharedBuffer.cpp:31:
/home/travis/build/SerenityOS/serenity/./AK/Userspace.h:35:1: error: ‘concept’ does not name a type
 concept PointerTypeName = IsPointer<T>::value;
 ^~~~~~~
/home/travis/build/SerenityOS/serenity/./AK/Userspace.h:35:1: note: ‘concept’ only available with -fconcepts
/home/travis/build/SerenityOS/serenity/./AK/Userspace.h:37:10: error: ‘PointerTypeName’ has not been declared
 template<PointerTypeName T>
          ^~~~~~~~~~~~~~~
/home/travis/build/SerenityOS/serenity/./AK/Userspace.h:54:16: error: expected ‘)’ before ‘ptr’
     Userspace(T ptr)
              ~ ^~~~
                )
/home/travis/build/SerenityOS/serenity/./AK/Userspace.h:59:5: error: ‘T’ does not name a type
     T ptr() const { return m_ptr; }
     ^
/home/travis/build/SerenityOS/serenity/./AK/Userspace.h:66:5: error: ‘T’ does not name a type
     T m_ptr { nullptr };
     ^
```
I can be lagom fine with `gcc`, travis is probably using `clang` to build `lagom` and needs `-fconcepts`.

CMakeLists.txt
```diff
@@ -40,7 +40,7 @@ add_custom_target(check-style
     USES_TERMINAL
 )
 
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra -Werror -std=c++2a -fdiagnostics-color=always")
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra -Werror -std=c++2a -fdiagnostics-color=always -fconcepts")
```
emanuele6, line 43: Let's see if this works.

emanuele6: It worked.

PR #2938, opened 2020-08-01
PixelPaint: Make `GenericConvolutionFilterInputDialog` cancellable

alimpfard:
Also tweaks the GUI to look a bit less...bad.

those types tho...they look horrible...

### DIFF
--- a/Applications/PixelPaint/Filters/GenericConvolutionFilter.cpp
+++ b/Applications/PixelPaint/Filters/GenericConvolutionFilter.cpp
@@ -106,12 +106,15 @@ void GenericConvolutionFilter<N>::apply(const Filter::Parameters& parameters)
 }
 
 template<size_t N>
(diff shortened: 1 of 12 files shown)


Applications/PixelPaint/Filters/GenericConvolutionFilter.cpp
```diff
@@ -49,8 +49,12 @@ GenericConvolutionFilter<N>::~GenericConvolutionFilter()
 }
 
 template<size_t N>
-void GenericConvolutionFilter<N>::apply(const Filter::Parameters& parameters)
+void GenericConvolutionFilter<N>::apply(const OwnPtr<Filter::Parameters>& maybe_parameters)
 {
+    if (!maybe_parameters)
+        return;
```
awesomekling, line 55: This API is weird. How about we just don't call `apply` when there are no parameters instead? :)

alimpfard: and add more `if`s?
sure, why not.

PR #2940, opened 2020-08-01
Setenv: Permit setting a new envvar, add tests

BenWiederhake:
This bug was brought to you by: "Boolean logic is !!hard" :)

I tried putting it in Lagom, but apparently then it links against the system version of `setenv` etc. I'm not even sure whether that's possible. So I put it in the same place with many other existing tests so far. Maybe would should have a dedicated "test executables" directory instead?

### DIFF
--- a/Libraries/LibC/stdlib.cpp
+++ b/Libraries/LibC/stdlib.cpp
@@ -290,7 +290,7 @@ int unsetenv(const char* name)
 
 int setenv(const char* name, const char* value, int overwrite)
 {
-    if (!overwrite && !getenv(name))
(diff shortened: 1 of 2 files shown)


alimpfard:
what I like to do about tests that are not useful in userland is put them in `$SERENITY_ROOT/Tests` and link them locally to Userland, so it gets built along with the system (and I can run them inside Serenity)

BenWiederhake:
Hmm, interesting idea. The tests should always be built though, at least on CI. In the long run, there should be a way to say "run all tests" inside Serenity.

However, that's a topic for another PR ^^

PR #2811, opened 2020-07-15
WindowServer: New title bar colors and metrics for themes

Nullspeak:
This exposes the window title bar height, title text shadow color, title stripe color and title button size to the theming system. This could be extended to generic widgets for theme-defined padding or border widths.

PR #2941, opened 2020-08-01
Themes: Fix default theme window stripes and shadow after #2811

nico:
Change #2811 made window title stripes and window title shadow themable,
but it used the same stripe and shadow color for all window modes.
This is fine for the new 'basalt' theme which uses the same color
in all four window modes, but it changed the default theme so that
background windows had brown stripes and a brown shadow.

Instead, make the title stripe and title shadow themable per window mode,
and change the default theme to restore the colors it had before
change #2811: The title stripe color is the same as Border1 for all
window modes, and the title shadow is the same as the title stripe
darkened by 0.6.


awesomekling:
Excellent! Thank you for fixing this :)

PR #2942, opened 2020-08-01
Travis: Marginally faster builds, maybe

BenWiederhake:
At least in theory, this should make it a bit faster:
- Build *Serenity* in parallel
- Build *libstdc++* in parallel
- Less chatty `./configure` output for binutils/gcc

The bulk of the compilation time is gcc/libgcc itself, so this has only little impact, if at all. I tried looking into making gcc build faster, but … no idea. Building gcc is slow. Deal with it. Or fix it. Yes, fix it, please, actually! :P

And this time I let the test complete before making the PR. :)

> Trying to cut down the list of "dependencies", so we can re-use a Toolchain more often. #2917 

That'll come next, eventually, maybe, soon, probably.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,7 +33,7 @@ before_install:
 - cmake --version
 
 script:
-- export SERENITY_ROOT=$(pwd)
+- export SERENITY_ROOT="$(pwd)"
 - Meta/lint-shell-scripts.sh
 - Meta/lint-executable-resources.sh
 - Meta/check-style.sh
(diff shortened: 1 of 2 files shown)


alimpfard:
Does travis provide multiple processors?
last I checked there was only one, but you could do some stuff with many jobs sharing memory space or whatnot?

BenWiederhake:
`nproc` on Travis [returns `2`](https://travis-ci.com/github/BenWiederhake/serenity/builds/178062139#L784), so we already build binutils and gcc with `-j 2`.

The official [documentation suggests](https://docs.travis-ci.com/user/speeding-up-the-build/#makefile-optimization) `-j 2`, and the [table of VMs](https://docs.travis-ci.com/user/reference/overview/#virtualisation-environment-vs-operating-system) confirms that our environment (and currently all others, too) have two cores.

awesomekling:
Great stuff :)

PR #2950, opened 2020-08-01
LibDebug: Disable and cleanup DebugSession breakpoints on destruction

bgianfo:
Breakpoints need to be disabled before we detach from the debugee.

I noticed this while looking into the fact that if you continue
executing a program in sdb (/bin/ls) where you had previously
set a breakpoint, it would crash on sdb exit once the debugee died
with an assert on HashMap destruction where we were iterating
while clearing is set. This change also happens to fix this assert.

FYI  @itamar8910

### DIFF
--- a/Libraries/LibDebug/DebugSession.cpp
+++ b/Libraries/LibDebug/DebugSession.cpp
@@ -38,6 +38,11 @@ DebugSession::DebugSession(int pid)
 
 DebugSession::~DebugSession()
 {
+    for (const auto& bp : m_breakpoints) {
+        disable_breakpoint(bp.key);
+    }
+    m_breakpoints.clear();
+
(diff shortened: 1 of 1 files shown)


Libraries/LibDebug/DebugSession.cpp
```diff
@@ -38,6 +38,12 @@ DebugSession::DebugSession(int pid)
 
 DebugSession::~DebugSession()
 {
+    for (const auto& bp : m_breakpoints)
+    {
```
awesomekling, line 42: Please use `clang-format` :)

bgianfo: whoops, sure

awesomekling:
👍

PR #2956, opened 2020-08-02
ProtocolServer+LibTLS: (very basic) Client Certificates

alimpfard:
This PR adds support for client certificates (but does not yet send any) and pipes certificate requests all the way from LibTLS to ProtocolServer clients.

This makes gemini work again (with a few more sites to boot!)


alimpfard:
LibGemini: cc @deoxxa

PR #2957, opened 2020-08-02
Kernel: Use Userspace<T> in sched_setparam syscall

bgianfo:
Note: I switched from copying the single element out of the sched_param
struct, to copy struct it self as it is identical in functionality.
This way the types match up nicer with the Userpace<T> api's and it
conforms to the conventions used in other syscalls.


awesomekling:
That's nicer indeed :)

Issue #2877, opened 2020-07-25
Libraries are completely rebuilt when a .cpp file in a dependency changes


Petelliott:
For instance, if I change `Libraries/LibCore/File.cpp` all l.cpp files in all libraries/applications that depend on it are rebuilt.

alimpfard:
rebuilt, or just re-linked?
the later is to be expected, we statically link everything as shared object loading is still a WIP.

Petelliott:
each cpp file in the library Is rebuilt, not just relinked.

alimpfard:
I guess I never noticed because ccache :shrug:

bugaevc:
> For instance, if I change `Libraries/LibCore/File.cpp` all l.cpp files in all libraries/applications that depend on it are rebuilt.

That's not true. To see it for yourself, try touching a file in LibGUI instead, you'll see that GUI apps get relinked, but no other .cpp files are recompiled. Now, LibCore is a bit special 'cause it's used by code generators, so changing LibCore will regenerate those files/headers, and cause everything that includes those headers (but not *all cpp files in all libraries/applications*) to be recompiled.

So this kind of dependency tracking between code generators & generated code is not only not a bug, but a very cool & desirable feature of CMake. See also: https://github.com/SerenityOS/serenity/pull/2519

nico:
Ninja (and I think Make) have a 'restat' feature where they stat the output of a build command after it ran, and if the build command didn't touch the output file, then it cancels the build actions down that path. That means if the code generators read the output file (if it exists) and exited early if what they were to write is identical to what's already there, then that optimization could fire and most of these compiles would get canceled if the output of the code generators didn't change since the last build.

Issue #2901, opened 2020-07-27
Meta: "make clean" leaves some build artifacts


BenWiederhake:
I was jumping back and fourth between commits, and ended up with some bad `.h` files in `Build/Root/`, and `make clean` doesn't clean them up:

```
user@machine:~/workspace/serenity/Build$ make clean && file Root/usr/include/LibWeb/Dump.h
Root/usr/include/LibWeb/Dump.h: C++ source, ASCII text
```

Sure, it's easy to remove it myself, just `rm -rf Root/usr/include/`. However, this is not how the problem presents itself. Instead, the console is spammed with weird error messages in modules I didn't even touch:
```
[ 74%] Building CXX object Libraries/LibWeb/CMakeFiles/LibWeb.dir/Bindings/NodeWrapperFactory.cpp.o
In file included from /home/user/workspace/serenity/Build/Root/usr/include/LibWeb/DOM/HTMLCanvasElement.h:31,
                 from /home/user/workspace/serenity/Build/Libraries/LibWeb/Bindings/HTMLCanvasElementWrapper.h:4,
                 from /home/user/workspace/serenity/Libraries/LibWeb/Bindings/NodeWrapperFactory.cpp:29:
/home/user/workspace/serenity/Build/Root/usr/include/LibWeb/DOM/HTMLElement.h:33:36: error: expected class-name before '{' token
   33 | class HTMLElement : public Element {
      |                                    ^
/home/user/workspace/serenity/Build/Root/usr/include/LibWeb/DOM/HTMLElement.h:37:25: error: expected ')' before '&' token
   37 |     HTMLElement(Document&, const FlyString& local_name);
      |                ~        ^
      |                         )
/home/user/workspace/serenity/Build/Root/usr/include/LibWeb/DOM/HTMLElement.h:38:13: error: 'virtual Web::HTMLElement::~HTMLElement()' marked 'override', but does not override
   38 |     virtual ~HTMLElement() override;
      |             ^
/home/user/workspace/serenity/Build/Root/usr/include/LibWeb/DOM/HTMLElement.h: In member function 'AK::String Web::HTMLElement::title() const':
/home/user/workspace/serenity/Build/Root/usr/include/LibWeb/DOM/HTMLElement.h:40:35: error: 'attribute' was not declared in this scope; did you mean 'Attribute'?
   40 |     String title() const { return attribute(HTML::AttributeNames::title); }
      |                                   ^~~~~~~~~
      |                                   Attribute
```

[...]

PR #2960, opened 2020-08-02
Build: Support make's and ninja's restat optimization

nico:
After running a build command, make by default stat()s the command's
output, and if it wasn't touched, then it cancels all build steps
that were scheduled only because this command was expected to change
the output.

Ninja has the same feature, but it's opt-in behind the per-command
"restat = 1" setting. However, CMake enables it by default for all
custom commands.

Use Meta/write-only-on-difference.sh to write the output to a temporary
file, and then copy the temporary file only to the final location if the
contents of the output have changed since last time.
write-only-on-difference.sh automatically creates the output's parent
directory, so stop doing that in CMake.

Reduces the number of build steps that run after touching a file
in LibCore from 522 to 312.

Since we now no longer trigger the CMake special case "If COMMAND
specifies an executable target name (created by the add_executable()
command), it will automatically be replaced by the location of the
executable created at build time", we now need to use qualified paths to
the generators.

Somewhat related to #2877.

---

Not sure if this is worth it since it's kind of complicated and Serenity builds fairly quickly anyhow.

### DIFF
--- a/Libraries/LibWeb/CMakeLists.txt
+++ b/Libraries/LibWeb/CMakeLists.txt
@@ -146,16 +146,14 @@ function(libweb_js_wrapper class)
     add_wrapper_sources(Bindings/${basename}Wrapper.cpp Bindings/${basename}Wrapper.h)
     add_custom_command(
         OUTPUT Bindings/${basename}Wrapper.h
-        COMMAND /bin/mkdir -p Bindings
-        COMMAND WrapperGenerator --header ${CMAKE_CURRENT_SOURCE_DIR}/${class}.idl > Bindings/${basename}Wrapper.h
+        COMMAND ${write_if_different} Bindings/${basename}Wrapper.h CodeGenerators/WrapperGenerator --header ${CMAKE_CURRENT_SOURCE_DIR}/${class}.idl
         VERBATIM
         DEPENDS WrapperGenerator
         MAIN_DEPENDENCY ${class}.idl
     )
     add_custom_command(
         OUTPUT Bindings/${basename}Wrapper.cpp
-        COMMAND /bin/mkdir -p Bindings
-        COMMAND WrapperGenerator --implementation ${CMAKE_CURRENT_SOURCE_DIR}/${class}.idl > Bindings/${basename}Wrapper.cpp
+        COMMAND ${write_if_different} Bindings/${basename}Wrapper.cpp CodeGenerators/WrapperGenerator --implementation ${CMAKE_CURRENT_SOURCE_DIR}/${class}.idl
         VERBATIM
         DEPENDS WrapperGenerator
         MAIN_DEPENDENCY ${class}.idl
@@ -203,8 +201,7 @@ set(SOURCES ${SOURCES} ${WRAPPER_SOURCES})
 
 add_custom_command(
     OUTPUT CSS/PropertyID.h
-    COMMAND /bin/mkdir -p CSS
-    COMMAND Generate_CSS_PropertyID_h ${CMAKE_CURRENT_SOURCE_DIR}/CSS/Properties.json > CSS/PropertyID.h
+    COMMAND ${write_if_different} CSS/PropertyID.h CodeGenerators/Generate_CSS_PropertyID_h ${CMAKE_CURRENT_SOURCE_DIR}/CSS/Properties.json
     VERBATIM
     DEPENDS Generate_CSS_PropertyID_h
     MAIN_DEPENDENCY CSS/Properties.json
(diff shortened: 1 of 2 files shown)


nico:
@BenWiederhake This seems like the kind of thing you might find interesting.

awesomekling:
I almost wonder if this could be condensed into a wrapper utility of some kind..
A command-line utility that runs a command, sending all the output to a file, but keeps the old timestamp if the file doesn't change.
Maybe it's too specific to be generally useful.

BenWiederhake:
@nico It does! :)

@awesomekling: I'll go with that, because among the generators is `GenerateStyleSheetSource.sh`, which obviously can't be affected by purely by changing the LogStream-templates.

@nico: While you are/were at it, can you take a look at #2901? Sometimes cmake seems to assume that a `.h` file doesn't need to be regenerated, and it also can't be `make clean`'d.

Libraries/LibWeb/CMakeLists.txt
```diff
@@ -142,21 +142,31 @@ function(libweb_js_wrapper class)
...
+        COMMAND WrapperGenerator --implementation ${CMAKE_CURRENT_SOURCE_DIR}/${class}.idl > Bindings/${basename}Wrapper.cpp.tmp
         VERBATIM
         DEPENDS WrapperGenerator
         MAIN_DEPENDENCY ${class}.idl
     )
+    add_custom_command(
+        OUTPUT Bindings/${basename}Wrapper.cpp
+        COMMAND ${CMAKE_COMMAND} -E copy_if_different Bindings/${basename}Wrapper.cpp.tmp Bindings/${basename}Wrapper.cpp
+        DEPENDS Bindings/${basename}Wrapper.cpp.tmp
+    )
```
bgianfo, line 169: It seems like this logic could/should be encapsulated into a single cmake function, like serenity_bin/serenity_lib

nico: I don't know, the plan is to not have this in cmake code eventually anyways. And it's not immediately clear to me how to do custom commands with optionally more than one COMMAND from a function. So since it's (for me) not easy to do and temporary, I deemed it not worth it.

bgianfo: > the plan is to not have this in cmake code eventually anyways.

Care to elaborate?

nico:
See "It'd be nicer if the code generators did this "write
output only if it's different from what's on disk" optimization
themselves, generating the potentially-written output only in
memory. " above.

Libraries/LibWeb/CMakeLists.txt
nico: See "It'd be nicer if the code generators did this "write
output only if it's different from what's on disk" optimization
themselves, generating the potentially-written output only in
memory. " above.

BenWiederhake: *poke* See #2970. Does that solve the problem as intended? :)

nico: I think it's nicer if the generator generates stuff in memory and only writes to disk if it would write something new, but I also think it's even more nicer if someone else does it, so feel free to take over ;)

BenWiederhake: > in memory

Yes, but then the comparison would also need to happen in-memory, and that makes things unnecessarily complicated. At least I don't know how to do that in `sh` :P

So all you've gotta do now is wait until #2970 lands.

nico: Well, it needs cooperation from each generator (http://reviews.llvm.org/rL373664 is a real-world example). #2970 means the approach in this change here doesn't permanently leave the .tmp file on disk which is nice, but you still need an extra build edge, you still have the disk traffic, and you still need the cmake goop. I'm always happy when I can move logc out of the build description into an actual programming language :) (But #2970 still definitely makes this change here nicer.)

BenWiederhake: Hmm? #2970 does not need further cooperation from each generator. (They all already write to stdout.) I agree that not touching the disk would be even nicer, I'll spend some more minutes trying to get that to work (but not hours, because this affects only a few files).

@cmake goop: Hmm? With the script, you only need a few more characters: `path/to/Meta/write-only-if-changed.sh `.

awesomekling: Having to author generators to not save output if it's unchanged seems like increasing the complexity of writing a generator for quite little benefit, no?

nico: Ah, I misread how #2970 is used. That makes everything much nicer indeed.

awesomekling:
Cool!

Issue #2958, opened 2020-08-02
Browser unkillable


tomuta:
Some recent change appears to have introduced some odd problem: Browser can't be killed.

Steps to reproduce:
1. Run Browser
2. Click the "SerenityOS @ GitHub" bookmark, notice that Browser starts hogging the CPU and appears to be hung
3. Try killing Browser with `killall -9 Browser` or System Monitor, it is unkillable

What's interesting is that if you watch Browser's stack trace in SystemMonitor it appears to be running in ultra-slow motion. You can watch it walk through the functions.

But it's really odd that it can't be killed. I'm not sure when it was introduced, but I wonder if the recent IP fragmentation changes may have something to do with it. After a few seconds, these log statements show up:
```
[NetworkTask(6:6)]: IPv4Socket(0xc04d3884): did_receive refusing packet since buffer is full.
[NetworkTask(6:6)]: IPv4Socket(0xc04d3884): did_receive refusing packet since buffer is full.
[NetworkTask(6:6)]: IPv4Socket(0xc04d3884): did_receive refusing packet since buffer is full.
[NetworkTask(6:6)]: IPv4Socket(0xc04d3a04): did_receive refusing packet since buffer is full.
[NetworkTask(6:6)]: IPv4Socket(0xc04d3b84): did_receive refusing packet since buffer is full.
ProtocolServer(25): Alert: 0
ProtocolServer(25): perror(): IODevice::write: write: Try again
ProtocolServer(25): perror(): IODevice::write: write: Try again
ProtocolServer(25): Alert: 0
ProtocolServer(25): perror(): IODevice::write: write: Try again
ProtocolServer(25): perror(): IODevice::write: write: Try again
ProtocolServer(25): Alert: 0
ProtocolServer(25): perror(): IODevice::write: write: Try again
[NetworkTask(6:6)]: handle_tcp: unexpected flags in CloseWait state
```

alimpfard:
The "recent" IP fragmentation stuff only applies to packets sent by us, we don't yet reassemble incoming packets. (SLiRP does that for us if run in QEMU)
(the `write: try again` stuff is very likely because LibTLS sends a close alert, but the socket is already closed, it shouldn't be an issue if that's the cause)

I wonder if the "ultra-slow motion" thing is just because it's repeating the same thing, and so the updates happen while the same-ish functions are running?

it being un-killable is rather odd though, no idea about that :shrug:

tomuta:
[...]

PR #2962, opened 2020-08-02
LibWeb: Fix some SVG crashes/hangs

### LINKED ISSUE #2945: LibWeb: SVG Parser hits infinite loop while parsing SVGs from https://github.com/SerenityOS/serenity/
Using latest master if you launch the browser, and try to navigate to the SerenityOS github the page never loads.
I look a profile, and it looks like the SVG parser is stuck in an infinite loop?

SVG-profile

mattco98:
- `parse_flag` now only parses one digit instead of consuming an entirely valid number
- `match_number` => `match_coordinate`
- `match_coordinate` now returns true if `ch()` is `.`
- `parse_number` no longer matches a `+`/`-`
- Don't crash when encountering one of the three unsupported path commands. Instead, just skip them. No reason to crash the browser over a silly SVG element

Bonus: check out these SVGs! They're a little rough around the edges, but it's better than nothing :)

a

Fixes #2945 and the hang observed in #2958

### DIFF
--- a/Libraries/LibWeb/SVG/SVGPathElement.cpp
+++ b/Libraries/LibWeb/SVG/SVGPathElement.cpp
@@ -36,6 +36,67 @@
 
 namespace Web::SVG {
 
+#ifdef PATH_DEBUG
+static void print_instruction(const PathInstruction& instruction)
+{
+    auto& data = instruction.data;
+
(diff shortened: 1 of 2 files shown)


awesomekling:
Very nice! :)

Issue #2388, opened 2020-05-26
AK TestFileSystemPath - relative_paths tests are broken


bgianfo:
FYI @bugaevc 
a3e4dfdf9859a9b955bf4728328f740a47de5851 appears to have broken the TestFileSystemPath - relative_paths tests.

bugaevc:
Indeed, thanks you!

```
FAIL: ../AK/Tests/TestFileSystemPath.cpp:60: assertEqual(path.string(), "./simple") failed
FAIL: ../AK/Tests/TestFileSystemPath.cpp:61: assertEqual(path.parts().size(), 2u) failed
FAIL: ../AK/Tests/TestFileSystemPath.cpp:67: assertEqual(path.string(), "./a/relative/path") failed
FAIL: ../AK/Tests/TestFileSystemPath.cpp:68: assertEqual(path.parts().size(), 4u) failed
FAIL: ../AK/Tests/TestFileSystemPath.cpp:74: assertEqual(path.string(), "./foo") failed
FAIL: ../AK/Tests/TestFileSystemPath.cpp:75: assertEqual(path.parts().size(), 2u) failed
FAIL: ../AK/Tests/TestFileSystemPath.cpp:82: assertEqual(path.string(), ".") failed
FAIL: ../AK/Tests/TestFileSystemPath.cpp:83: assertEqual(path.parts().size(), 1u) failed
FAIL: ../AK/Tests/TestFileSystemPath.cpp:84: assertEqual(path.basename(), ".") failed
```

Most of those are because of me removing the initial `.` in relative paths, 'cause I don't think that's what canonicalizing a path should result in (@awesomekling — was that a bug or a feature?)

A single dot should not become an empty path though, that one's definitely a bug.

awesomekling:
I think that was neither bug nor feature, just how I ended up implementing canonicalization for relative paths. Maybe `FileSystemPath` should be renamed to `LexicalPath` or similar, to make it clearer that it doesn't actually interact with the filesystem :|

bugaevc:
Yes, `LexicalPath` sounds good. PR incoming...

PR #2964, opened 2020-08-02
AK: Remove relative_paths test from TestLexicalPath

bgianfo:
This test appears to be testing functionality that doesn't exist.
Just remove it.

Closes old bug #2388

CC @bugaevc


bugaevc:
Yay, finally someone remembered to cc me 😀

PR #2965, opened 2020-08-02
Kernel: Fix signal delivery when no syscall is made

### LINKED ISSUE #2958: Browser unkillable

tomuta:
This fixes a regression since introducing software context
switching where the Kernel would not deliver a signal unless the
process is making system calls. This is because the TSS no longer
updates the CS value, so the scheduler never considered delivery
as the process always appeared to be in kernel mode. With software
context switching we can just set up the signal trampoline at
any time and when the processor returns back to user mode it'll
get executed. This should fix e.g. killing programs that are
stuck in some tight loop that doesn't make any system calls and
is only pre-empted by the timer interrupt.

Fixes #2958

### DIFF
--- a/Kernel/Process.cpp
+++ b/Kernel/Process.cpp
@@ -773,7 +773,7 @@ void Process::terminate_due_to_signal(u8 signal)
 {
     ASSERT_INTERRUPTS_DISABLED();
     ASSERT(signal < 32);
-    dbg() << "Terminating due to signal " << signal;
+    dbg() << "Terminating " << *this << " due to signal " << signal;
     m_termination_status = 0;
     m_termination_signal = signal;
     die();
--- a/Kernel/Scheduler.cpp
+++ b/Kernel/Scheduler.cpp
@@ -405,17 +405,6 @@ bool Scheduler::pick_next()
     Thread::for_each_living([&](Thread& thread) -> IterationDecision {
         if (!thread.has_unmasked_pending_signals())
             return IterationDecision::Continue;
-        // FIXME: It would be nice if the Scheduler didn't have to worry about who is "current"
-        //        For now, avoid dispatching signals to "current" and do it in a scheduling pass
-        //        while some other process is interrupted. Otherwise a mess will be made.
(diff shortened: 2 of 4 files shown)


awesomekling:
Oh hey, that's so much nicer this way :D

PR #2967, opened 2020-08-02
Kernel: Use for-each loops in unveil syscall

bgianfo:


awesomekling:
I think the proper name for these is range-for loops, but for-each works for me :)

PR #2969, opened 2020-08-02
Hymn tribute

devfoFikiCar:
11. August is coming up, 2 years since the CIA killed Terry, a small tribute.


awesomekling:
I appreciate the thought, but we are not putting 5 megabytes of this into the repository.

devfoFikiCar:
@awesomekling if I compress and make it smaller? I understand either way.

PR #2971, opened 2020-08-02
SystemMonitor: More beautiful

BenWiederhake:
This makes the SystemMonitor GUI a bit more beautiful (new actions, better usability, better readability).


awesomekling:
Beautiful! :lipstick:

PR #2972, opened 2020-08-02
shell: implement Shell::is_runnable and highlight commands

math2001:
Hey Serenity!

Inspired from the fish shell, the terminal now highlights the current command based on whether it exists or not (red if it is invalid, and bold white if it's valid).

serenity-terminal-command-highlight

I add the method `is_runnable` to the shell. I'm not sure if that's the best name/go to way about it. Feedback more than welcome :slightly_smiling_face: 

Also, the builtin `alias` wasn't refreshing the cache `cached_path` when an aliases was added. That meant that no aliases were cached because the cache is populated *before* the rc files are loaded. So this is fixed (each alias call insert the alias in the cache).

Shouldn't `cached_path` be called `m_cached_path`? Or am i missing something? 

Thanks this awesome project, it's really inspiring :heart:

### DIFF
--- a/Shell/AST.cpp
+++ b/Shell/AST.cpp
@@ -367,9 +367,9 @@ RefPtr<Value> BarewordLiteral::run(RefPtr<Shell>)
 void BarewordLiteral::highlight_in_editor(Line::Editor& editor, Shell& shell, HighlightMetadata metadata)
 {
     if (metadata.is_first_in_list) {
-        if (shell.is_runnable(m_text))
+        if (shell.is_runnable(m_text)) {
             editor.stylize({ m_position.start_offset, m_position.end_offset }, { Line::Style::Bold });
-        else {
+        } else {
             editor.stylize({ m_position.start_offset, m_position.end_offset }, { Line::Style::Foreground(Line::Style::XtermColor::Red) });
         }
 
--- a/Shell/AST.cpp
+++ b/Shell/AST.cpp
@@ -367,9 +367,15 @@ RefPtr<Value> BarewordLiteral::run(RefPtr<Shell>)
 void BarewordLiteral::highlight_in_editor(Line::Editor& editor, Shell& shell, HighlightMetadata metadata)
 {
     if (metadata.is_first_in_list) {
-        editor.stylize({ m_position.start_offset, m_position.end_offset }, { Line::Style::Bold });
+        if (shell.is_runnable(m_text))
+            editor.stylize({ m_position.start_offset, m_position.end_offset }, { Line::Style::Bold });
+        else {
+            editor.stylize({ m_position.start_offset, m_position.end_offset }, { Line::Style::Foreground(Line::Style::XtermColor::Red) });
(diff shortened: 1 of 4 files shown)


awesomekling:
cc @alimpfard

Shell/AST.cpp
```diff
@@ -367,9 +367,15 @@ RefPtr<Value> BarewordLiteral::run(RefPtr<Shell>)
 void BarewordLiteral::highlight_in_editor(Line::Editor& editor, Shell& shell, HighlightMetadata metadata)
 {
     if (metadata.is_first_in_list) {
-        editor.stylize({ m_position.start_offset, m_position.end_offset }, { Line::Style::Bold });
+        if (shell.is_runnable(m_text))
+            editor.stylize({ m_position.start_offset, m_position.end_offset }, { Line::Style::Bold });
+        else {
```
alimpfard, line 372: Please either omit the braces here, or add one to the `if` above.

Shell/Builtin.cpp
```diff
@@ -64,6 +64,20 @@ int Shell::builtin_alias(int argc, const char** argv)
...
+                },
+                &index);
+
+            if (match)
+                continue;
+
+            while (strcmp(cached_path[index].characters(), parts[0].characters()) < 0) {
+                index++;
+            }
+            cached_path.insert(index, parts[0]);
```
alimpfard, line 80: How about a separate function for this?

Shell/AST.cpp
```diff
@@ -367,9 +367,15 @@ RefPtr<Value> BarewordLiteral::run(RefPtr<Shell>)
...
+            editor.stylize({ m_position.start_offset, m_position.end_offset }, { Line::Style::Foreground(Line::Style::XtermColor::Red) });
```
alimpfard, line 373: Let's keep it bold either way, it's nice to have an indicator that doesn't depend on the text itself.

Also, I'm not too sure about the color here, red is also used to highlight syntax errors...but no other good alternative comes to mind, feel free to disregard this bit.

alimpfard:
> Shouldn't cached_path be called m_cached_path?

It should, but no one has gotten around to dealing with it yet (same applies for most of the state in the Shell).
it should be simple to fix now that Shell's behavior is part of the class itself.

Shell/Builtin.cpp
math2001: Oh yeah I'll do this

Shell/AST.cpp
math2001: Oops that wasn't intended

Shell/AST.cpp
math2001: Bold red seems a bit in your face IMO. Are you sure? I'll push it if you want though.

serenity-terminal

alimpfard: Well...the non-bold variant looks better, I guess.
let's leave it non-bold then.

PR #2973, opened 2020-08-03
Fix terminal hang

tomuta:
On my quest to finally solve the occasional Terminal hang, I found some possible races with `Thread::block`. But fixing all these still didn't solve the problem. It turned out to be a problem with LibIPC. This subtle log message ended up giving it away:
```
WindowServer(14): AddMenuItem: Bad menu ID: 1982
```

### DIFF
--- a/Kernel/Thread.h
+++ b/Kernel/Thread.h
@@ -126,7 +126,11 @@ public:
     class Blocker {
     public:
         virtual ~Blocker() { }
-        virtual bool should_unblock(Thread&, time_t now_s, long us) = 0;
+        virtual bool should_unblock(Thread& thread, time_t, long)
+        {
+            return should_unblock(thread);
+        }
+        virtual bool should_unblock(Thread&) = 0;
         virtual const char* state_string() const = 0;
         virtual bool is_reason_signal() const { return false; }
         void set_interrupted_by_death() { m_was_interrupted_by_death = true; }
@@ -143,7 +147,7 @@ public:
     class JoinBlocker final : public Blocker {
     public:
         explicit JoinBlocker(Thread& joinee, void*& joinee_exit_value);
-        virtual bool should_unblock(Thread&, time_t now_s, long us) override;
+        virtual bool should_unblock(Thread&) override;
         virtual const char* state_string() const override { return "Joining"; }
         void set_joinee_exit_value(void* value) { m_joinee_exit_value = value; }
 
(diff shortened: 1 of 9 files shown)


tomuta:
Hm so I think the LibIPC change is not quite right. While it solves the problem with the occasional Terminal hang, it causes DisplaySettings to hang hwen changing the wallpaper... Not sure what the right solution is, I guess I don't really understand why `LocalEndpoint` and `PeerEndpoint` are both read from the same socket. Why does `ServerConnection::drain_messages_from_server` decode `LocalEndpoint` messages at all?

tomuta:
Ok, not sure if this is the right solution but I changed `ServerConnection::wait_for_specific_message` to only look at `LocalEndpoint` messages and `ServerConnection::send_sync` to only look at `PeerEndpoint` message responses. Seems to work, but I don't know if that is an assumption that can be made. Thoughts?

Kernel/Thread.h
```diff
@@ -126,7 +126,11 @@ class Thread {
     class Blocker {
     public:
         virtual ~Blocker() { }
-        virtual bool should_unblock(Thread&, time_t now_s, long us) = 0;
+        virtual bool should_unblock(Thread& thread, time_t, long)
+        {
+            return should_unblock(thread);
+        }
+        virtual bool should_unblock(Thread&) = 0;
```
awesomekling, line 133: What's the point of breaking this into two separate virtuals?
A subclass would never want to override both.

awesomekling:
> Ok, not sure if this is the right solution but I changed `ServerConnection::wait_for_specific_message` to only look at `LocalEndpoint` messages and `ServerConnection::send_sync` to only look at `PeerEndpoint` message responses. Seems to work, but I don't know if that is an assumption that can be made. Thoughts?

Yes, that's a great fix. Thank you for digging into this :)

Kernel/Thread.h
tomuta: My vision for this is to remove the `virtual bool should_unblock(Thread& thread, time_t, long)` variant. Right now there are 3 or 4 that all implement the timeout logic (the same way), but the rest don't. For `Thread::block` I need to be able to check the block condition without taking a timeout into account, so I'm using `bool should_unblock(Thread&)` there. But I don't think the `Blocker` classes really should be in the business of determining whether a timeout happened, that in my mind should be universal and the same logic should apply to all.

So, for now, because of `Thread::block` I temporarily split them up, but my plan is to remove the first variant entirely.

awesomekling: Okay, that's a good reason :)

PR #2974, opened 2020-08-03
LibGUI: WidgetScrollable scrolls horizontally when shift is pressed

math2001:
If you want to scroll horizontally, just press `shift` and then the scroll wheel will move the horizontal scroll bar.

`shift` is the standard modifier I think, and that's how it works in firefox on ubuntu.

serenity-horizontal-scroll


awesomekling:
Sure, that works :) Great idea!

PR #2975, opened 2020-08-03
LibWeb: Move contentEditable from Element to HTMLElement

Lubrsi:
HTMLElement is the only interface that includes ElementContentEditable
in the HTML specification. This makes sense, as Element is also a base
class for elements in other specifications such as SVG,
which definitely shouldn't be editable.

Also adds a test for the attribute based on what Andreas did in the
video that added it.


awesomekling:
Ah fair enough. Thanks for having my back @Lubrsi :D

PR #2976, opened 2020-08-03
AK: Fix min(a,b)/max(a,b) in case a and b  are equivalent

tryfinally:
min(a, b) should return a if both are equivalent
max(a, b) should return a if both are equivalent
see: C++ standard [25.8 Sorting and related operations[alg.sorting]](http://eel.is/c++draft/alg.min.max#2)


awesomekling:
`AK` is not an implementation of the C++ standard template library, I don't understand the reasoning here.
Also you didn't fix spelling or add a test.

tryfinally:
@awesomekling While it is not an implementation of C++ standard library, current implementation :
 min(a, b)  will return b if they are equivalent
 max(a,b)  will return a if they are equivalent
IMHO this is confusing and inconsistent.

I see the following options:
1. No Surprise Rule:  provide min/max with the same post conditions as the C++ standard
2. We are not implementing the Standard: provide a better implementation that is generic and consistent.
    Yes, the C++ standard is wrong on this: it makes min stable and max unstable.
IMHO option 2 is the way to go: make both stable

imagine they are used to sort a pair:
p.first = min(p.first, p.second);
p.second = max(p.first, p.second);

PR #2980, opened 2020-08-03
Kernel: Consolidate timeout logic

tomuta:
Allow passing in an optional timeout to Thread::block and move
the timeout check out of Thread::Blocker. This way all Blockers
implicitly support timeouts and don't need to implement it
themselves. Do however allow them to override timeouts (e.g.
for sockets).

### DIFF
--- a/AK/Time.h
+++ b/AK/Time.h
@@ -61,11 +61,33 @@ inline void timespec_sub(const TimespecType& a, const TimespecType& b, TimespecT
     }
 }
 
+template<typename TimespecType, typename TimevalType>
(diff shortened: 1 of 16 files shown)


awesomekling:
V. nice

nico:
Since we're using c++20, the comparison operators could probably be simplified with the spaceship operator. (Haven't tried it though, maybe it doesn't work for some reason.)

tomuta:
> Since we're using c++20, the comparison operators could probably be simplified with the spaceship operator. (Haven't tried it though, maybe it doesn't work for some reason.)

I thought about it, but I didn't feel like experimenting ;-)

PR #2982, opened 2020-08-03
LibJS: The Math.ceil() of a number between -1 and 0 should be -0, according to the spec.

qeeg:


Libraries/LibJS/Runtime/MathObject.cpp
```diff
@@ -125,10 +125,13 @@ JS_DEFINE_NATIVE_FUNCTION(MathObject::floor)
 JS_DEFINE_NATIVE_FUNCTION(MathObject::ceil)
 {
     auto number = interpreter.argument(0).to_number(interpreter);
+    auto number_double = number.as_double();
     if (interpreter.exception())
         return {};
     if (number.is_nan())
         return js_nan();
+    if (number_double < 0 && number_double > -1)
+        return Value(-0.f);
```
linusg, line 134:
```suggestion
        return Value(-0);
```

qeeg: The test doesn't work if I change it to that.

linusg: huh - `-0.0` then? Either way, there's no float overload of `JS::Value`.

Libraries/LibJS/Runtime/MathObject.cpp
```diff
@@ -125,10 +125,13 @@ JS_DEFINE_NATIVE_FUNCTION(MathObject::floor)
 JS_DEFINE_NATIVE_FUNCTION(MathObject::ceil)
 {
     auto number = interpreter.argument(0).to_number(interpreter);
+    auto number_double = number.as_double();
```
linusg, line 128: This needs to be after the exception check below as `to_number()` can return an empty value (if it throws), and `as_double()` would assert in that case.

PR #2984, opened 2020-08-04
Shell: Add support for ARGV (and $*, $#)

### LINKED ISSUE #2948: Shell: `echo $*` is surprising
So I was playing around with the Shell in the Terminal, and:

```
courage ~ $ echo $*
Browser.ini: Command not found

courage ~ $
```

This also happens if the command is executed from a file:

```
courage ~ $ ./foo.sh
Browser.ini: Command not found

courage ~ $
```

I'm not sure what *precisely* `$*` is supposed to do, but ignoring echo and trying to run a file is probably a bug :)

CC @alimpfard

alimpfard:
This patchset also adds the 'shift' builtin, as well as the usual tests.
closes #2948.

It should be noted that this does not adopt the '`$*` is a string, and `$@` is a list` idea, `$*` is a list and `$@` is currently not defined.

more gritty details:
- `$*` is an alias for `$ARGV`, which can be redefined by the user.
- `$#` is an alias for the size of `$ARGV` (we do not yet support any way of checking for list length inside the shell)
- the `shift` builtin also works on `$ARGV`.
- a non-list `$ARGV` is treated as a list containing that only that element.

cc @bugaevc (if you have any input on this)

### DIFF
--- a/Shell/AST.cpp
+++ b/Shell/AST.cpp
@@ -1998,6 +1998,7 @@ RefPtr<Value> SimpleVariableValue::resolve_without_cast(RefPtr<Shell> shell)
 SpecialVariableValue::~SpecialVariableValue()
 {
 }
+
 Vector<String> SpecialVariableValue::resolve_as_list(RefPtr<Shell> shell)
 {
     switch (m_name) {
@@ -2005,6 +2006,19 @@ Vector<String> SpecialVariableValue::resolve_as_list(RefPtr<Shell> shell)
         return { String::number(shell->last_return_code) };
     case '$':
         return { String::number(getpid()) };
+    case '*':
+        if (auto argv = shell->lookup_local_variable("ARGV"))
+            return argv->resolve_as_list(shell);
+        return {};
+    case '#':
+        if (auto argv = shell->lookup_local_variable("ARGV")) {
+            if (argv->is_list()) {
+                auto list_argv = static_cast<AST::ListValue*>(argv.ptr());
+                return { String::number(list_argv->values().size()) };
+            }
+            return { "1" };
+        }
+        return { "0" };
     default:
         return { "" };
     }
--- a/Shell/AST.cpp
+++ b/Shell/AST.cpp
@@ -540,7 +540,7 @@ void CloseFdRedirection::dump(int level) const
 RefPtr<Value> CloseFdRedirection::run(RefPtr<Shell>)
 {
     Command command;
-    command.redirections.append(*new CloseRedirection(m_fd));
+    command.redirections.append(adopt(*new CloseRedirection(m_fd)));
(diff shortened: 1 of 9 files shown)


Shell/main.cpp
```diff
@@ -181,6 +183,13 @@ int main(int argc, char** argv)
         run_rc_file(Shell::local_init_file_path);
     }
 
+    {
+        Vector<String> args;
+        for (auto* arg : script_args)
+            args.empend(arg);
+        shell->set_local_variable("ARGV", *new AST::ListValue(move(args)));
```
awesomekling, line 190: This looks like a reference leak; the `*new AST::ListValue` is not adopted.

alimpfard: Oh oops!

alimpfard: (there are quite a few around, I think, I'll search for them and fix them along with this one)

awesomekling: A `create()` helper that always returns a `NonnullRefPtr` would be nice here :)

Shell/AST.cpp
```diff
@@ -1116,8 +1116,8 @@ RefPtr<Value> Pipe::run(RefPtr<Shell> shell)
 
     auto pipe_write_end = new FdRedirection(STDIN_FILENO, -1, Rewiring::Close::Destination);
     auto pipe_read_end = new FdRedirection(STDOUT_FILENO, -1, pipe_write_end, Rewiring::Close::RefreshDestination);
-    first_in_right.redirections.append(*pipe_write_end);
-    last_in_left.redirections.append(*pipe_read_end);
+    first_in_right.redirections.empend(adopt(static_cast<Redirection&>(*pipe_write_end)));
+    last_in_left.redirections.empend(adopt(static_cast<Redirection&>(*pipe_read_end)));
```
awesomekling, line 1120: Please wrap the `new` in `adopt` directly instead of doing it later.

Shell/AST.h
```diff
@@ -122,7 +122,7 @@ struct PathRedirection : public Redirection {
 };
 
 struct FdRedirection : public Redirection
-    , public Rewiring {
+    , virtual Rewiring {
```
awesomekling, line 125: Wait, what's even going on here..
`Redirection` and `Rewiring` are both `RefCounted`. That's no good, that gives us two reference counts in the same object. That needs to be fixed.

alimpfard: That's what I'm trying right now.
Not sure what I was doing when I came up with that idea.

Shell/AST.cpp
alimpfard: With the double-`RefCounted` weirdness now fixed, this `static_cast` isn't needed anymore.

PR #2987, opened 2020-08-04
Shell: Give the TTY to the foreground process

alimpfard:
This fixes the bug with the shell not waiting for any foreground process
that attempts to read from the terminal in the Lagom build.

@awesomekling 
> if I run vim in lagom-shell, I can't exit

All you have to do is `:wq`, trust me!

(this works fine with the `SIGTTIN` and `SIGTTOU` patch, but I didn't put it here, I can - if you want)


awesomekling:
Cool!

PR #2992, opened 2020-08-05
AK + Kernel: Switch IPv4Socket receive queue to new SinglyLinkedListWithCount<T>

bgianfo:
Noticed the FIXME while working on something else, death to all O(n) implementations :)

We can avoid walking the packet queue, instead use a linked list with a count.
To do this add a new list type SinglyLinkedListWithCount<T>, which just wraps
SinglyLinkedList<T> and intercepts calls which need to modify the count.


awesomekling:
Yup, excellent addition :)

PR #2994, opened 2020-08-05
Windowserver flicker fixes

tomuta:
- [x] Render multiple overlapping Fire and Cube instances without flickering
- [x] With few exceptions, render only what is actually visible on the screen (e.g. skip rendering areas covered by opaque windows)
- [x] Desktop repainting
- [x] Desktop selection rubberband repainting
- [x] Fix rendering of semi-transparent windows (Eyes)
- [x] Triple-buffering for transparency regions to prevent flickering with Eyes
- [x] Fix `Window::invalidate` translation so frame buttons are updated properly
- [x] Render frame only when needed (currently inefficient, should cache in a backing bitmap)
- [x] Fix ../Libraries/LibGUI/Window.cpp:289 assertion triggered by System Monitor Graph's tab
- [x] Fix minor screen corruption when cursor is close to left/top screen edge
- [x] Fix animations
- [x] Fix Drag&Drop artifacts
- [x] Fix transparency issues with logo + scroll wheel
- [x] Fix combobox not repainting background upon closing when something is selected. Same bug seems to affect tooltips, too.

### DIFF
--- a/Services/WindowServer/Compositor.cpp
+++ b/Services/WindowServer/Compositor.cpp
@@ -32,10 +32,14 @@
 #include "Window.h"
 #include "WindowManager.h"
 #include <AK/Memory.h>
+#include <AK/ScopeGuard.h>
 #include <LibCore/Timer.h>
 #include <LibGfx/Font.h>
 #include <LibGfx/Painter.h>
 #include <LibThread/BackgroundAction.h>
+
+//#define COMPOSE_DEBUG
+//#define OCCLUSIONS_DEBUG
 
 namespace WindowServer {
 
@@ -96,12 +100,15 @@ void Compositor::init_bitmaps()
     else
         m_back_bitmap = Gfx::Bitmap::create(Gfx::BitmapFormat::RGB32, size);
 
+    m_temp_bitmap = Gfx::Bitmap::create(Gfx::BitmapFormat::RGB32, size);
+
     m_front_painter = make<Gfx::Painter>(*m_front_bitmap);
     m_back_painter = make<Gfx::Painter>(*m_back_bitmap);
+    m_temp_painter = make<Gfx::Painter>(*m_temp_bitmap);
 
     m_buffers_are_flipped = false;
 
-    invalidate();
+    invalidate_screen();
 }
 
 void Compositor::compose()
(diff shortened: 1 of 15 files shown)


tomuta:
Oddly, now that I fixed many of the fundamental problems, the desktop wallpaper tends to flicker. That's because it's a transparent window which invalidates the entire area. This then causes the wallpaper to be redrawn, and then the transparent window over it. It doesn't flicker over other windows though (as expected).

I'm not sure what the best solution for this would be. Maybe the desktop window shouldn't be transparent and itself render the wallpaper? Triple buffering for anything that involves transparency?
Eyes has a similar "problem" in that it may flicker where the area underneath is repainted, because we don't triple buffer. I'm not sure if the overhead to triple buffer those things is worth it...

tomuta:
Hm my last few changes cause stuff to get rendered unnecessarily again...

Services/WindowServer/Compositor.cpp
```diff
@@ -111,21 +111,23 @@ void Compositor::compose()
...
 
-    dirty_rects.add(Gfx::IntRect::intersection(m_last_geometry_label_rect, Screen::the().rect()));
-    dirty_rects.add(Gfx::IntRect::intersection(m_last_cursor_rect, Screen::the().rect()));
-    dirty_rects.add(Gfx::IntRect::intersection(m_last_dnd_rect, Screen::the().rect()));
-    dirty_rects.add(Gfx::IntRect::intersection(current_cursor_rect(), Screen::the().rect()));
+    auto dirty_screen_rects = move(m_dirty_screen_rects);
+    dirty_screen_rects.add(Gfx::IntRect::intersection(m_last_geometry_label_rect, Screen::the().rect()));
+    dirty_screen_rects.add(Gfx::IntRect::intersection(m_last_dnd_rect, Screen::the().rect()));
+    if (m_invalidated_cursor) {
+        // TODO: what is this?
```
awesomekling, line 123: Drag n drop!

tomuta: Ah, thanks! I kinda wish all these special things (desktop, dnd, ...) were just windows themselves, then we wouldn't need these special cases and the overhead that comes with it. And looks like they're not quite working right, I'm guessing I just need to add the previous position to the `flush_rects`. I'll look at it later tonight.

tomuta:
Aside from the few remaining bugs, rendering works without flickering, unless a window is non-opaque (Eyes), in which case any portions that need updating may flicker because we don't triple-buffer.

Services/WindowServer/Compositor.cpp
awesomekling: Agreed that it would be sweet if they were just windows.

tomuta:
As you can see in these screen captures any flicker is no constrainted to non-opaque windows:

Frame before:
cap_flicker_00:00:03_01

Flicker:
cap_flicker_00:00:03_02

Frame after:
cap_flicker_00:00:03_03

tomuta:
Well, I'm adding triple-buffering because that flickering still bothers me. The overhead seems small enough to not really notice much of a difference (other than it doesn't flicker)...

bugaevc:
Heh, so the 👀 turned out to be actually useful for testing stuff 🙂

tomuta:
How do you type the "Windows" key in qemu (on Ubuntu 20.04)? Aside from doing it through the monitor sending the key through command line? I need to test the window switcher and some other shortcuts...

tomuta:
What applications other than FileManager support drag&drop? That one unfortunately keeps crashing (it does in `master` as well)...

bugaevc:
> What applications other than FileManager support drag&drop?

The terminal.

tomuta:
How do I type the Logo/Windows key in QEMU aside from the send key through the monitor cli? I'm trying to test some features, window switcher, and I guess you can change opacity somehow with it as well?

awesomekling:
I just press it on the keyboard. I guess your host window manager steals it. Try doing a full QEMU capture with Ctrl Alt G

tomuta:
I think I got the rendering in fairly good shape. While windows don't flicker anymore, the following few things can still flicker a little:
- mouse pointer when over transparent areas that are frequently updated (can be solved with a little more complexity)
- dnd may flicker slightly, could be solved by making it a transparent `Window`
- possibly the geometry label, which, similar to dnd could be solved by making it a regular `Window`

Other added benefits are that windows that are covered consume a whole lot less CPU, e.g. if the Fire application is covered, CPU usage drops significantly. But there is a small overhead due to added accounting, though it seems to be a wash overall.

awesomekling:
This looks real nice. Super great work!
Let's merge it and see if we run into any issues :)

PR #2995, opened 2020-08-05
Shell+LibLine: Do not assume that stdin/stdout is a TTY

### LINKED ISSUE #2989: Shell assumes that stdin should be a tty
Examples:

1. `Shell -c ls < /dev/null` This runs ls fine, but it first complains that "ttyname_r: Not a TTY". It probably just shouldn't complain.
2. `Shell -c 'echo ls | Shell'`dies with "Good-bye!: Not a TTY" the first time I run it, then the next I run it it prints the output in (1) instead and breaks the Terminal session.

(The background is that ninja runs subprocesses in `sh -c some_command` with stdin replaced with /dev/null.)

alimpfard:
Pipes and such.

This closes #2989.

### DIFF
--- a/Libraries/LibLine/Editor.cpp
+++ b/Libraries/LibLine/Editor.cpp
@@ -250,11 +250,18 @@ void Editor::initialize()
     if (m_was_resized)
         get_terminal_size();
 
-    auto* term = getenv("TERM");
-    if (StringView { term }.starts_with("xterm"))
-        m_configuration.set(Configuration::Full);
-    else
-        m_configuration.set(Configuration::NoEscapeSequences);
+    if (m_configuration.operation_mode == Configuration::Unset) {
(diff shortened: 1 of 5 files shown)


Libraries/LibLine/Editor.cpp
```diff
@@ -429,7 +447,8 @@ void Editor::handle_read_event()
     Utf8View input_view { StringView { m_incomplete_data.data(), valid_bytes } };
     size_t consumed_code_pointss = 0;
 
-    enum Amount { Character, Word };
+    enum Amount { Character,
+        Word };
```
awesomekling, line 451: `<bugaevc>`:cry:`</bugaevc>`

alimpfard: smh These always fly unnoticed...

PR #2997, opened 2020-08-05
Kernel: Partial usage of Userspace<T> for the poll syscall

bgianfo:
This change mostly converts poll to Userspace<T> with the caveat
of the fds member of SC_poll_params. It's current usage is a bit
too gnarly for me to take on right now, this appears to need a lot
more love.

In addition to enlightening the syscall to use Userspace<T>, I've
also re-worked most of the handling to use validate_read_and_copy
instead of just directly de-referencing the user pointer. We also
appeared to be missing a re-evaluation of the fds array after the
thread block is awoken.

I separated this from PR #2996 as I thought it deserved a closer review.

### DIFF
--- a/Kernel/Syscalls/select.cpp
+++ b/Kernel/Syscalls/select.cpp
@@ -139,27 +139,29 @@ int Process::sys$select(const Syscall::SC_select_params* params)
     return marked_fd_count;
 }
 
-int Process::sys$poll(const Syscall::SC_poll_params* params)
+int Process::sys$poll(Userspace<const Syscall::SC_poll_params*> user_params)
 {
     REQUIRE_PROMISE(stdio);
+
     // FIXME: Return -EINVAL if timeout is invalid.
-    if (!validate_read_typed(params))
+    Syscall::SC_poll_params params;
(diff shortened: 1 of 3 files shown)


Kernel/Syscalls/select.cpp
```diff
@@ -139,27 +139,29 @@ int Process::sys$select(const Syscall::SC_select_params* params)
...
     if (fds && !validate_read_typed(fds, nfds))
         return -EFAULT;
-    if (timeout && !validate_read_typed(timeout))
-        return -EFAULT;
-    if (sigmask && !validate_read_typed(sigmask))
+
+    timespec timeout = {};
+    if (params.timeout && !validate_read_and_copy_typed(&timeout, params.timeout))
         return -EFAULT;
-    if (!validate_read_typed(fds))
```
bgianfo, line 162: This appeared to be a duplicate check so I removed? Am I missing something?

bgianfo: Thoughts?

awesomekling: Sounds good. Also this check was too narrow as it would only validate the first entry in `fds`.

PR #2999, opened 2020-08-05
Kernel: Decorate KResultOr with [[nodiscard]] to catch misbehaving callers, and fix existing issues. 

bgianfo:
The [[nodiscard]] decorator enables us to have the compile emit a
warning anytime the decorated type or function's return value isn't
observed. It's very useful for catching error checking bugs in systems
which use C style error handling. This change enables it for KResultOr<T>

In order to get this to compile clean I've fixed all of the Inode::directory_entry_count()
implementations to return KResultOr<size_t> so that we can properly observe and handle
all of paths which were missing the observation of KResultOr<T> return values.

I was looking at decorating KResult with [[nodiscard]] as well, but that cleanup
is a much larger chore that will need to be tackled separately.


awesomekling:
Great idea

PR #3001, opened 2020-08-05
Kernel: Decorate KResult with [[nodiscard]] to catch misbehaving callers, and fix and suppress issues it found. 

bgianfo:
Continuing off of #2999, this enables [[nodiscard]] for KResult.

There are unfortunately a bunch of dead end paths that require
plumbing error paths throughout the codebase to fix. To get the
warning enabled and stop from new problems being added, I've
suppressed the existing difficult to deal with issues, and left FIXME
bread crumbs so we can hopefully go back and fix them over time.

### DIFF
--- a/Kernel/TTY/TTY.cpp
+++ b/Kernel/TTY/TTY.cpp
@@ -55,7 +55,9 @@ void TTY::set_default_termios()
 KResultOr<size_t> TTY::read(FileDescription&, size_t, u8* buffer, size_t size)
 {
     if (Process::current()->pgid() != pgid()) {
-        Process::current()->send_signal(SIGTTIN, nullptr);
+        KResult result = Process::current()->send_signal(SIGTTIN, nullptr);
+        if (result.is_error())
+            return result;
         return KResult(-EINTR);
     }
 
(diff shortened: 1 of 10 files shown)


Kernel/TTY/TTY.cpp
```diff
@@ -55,7 +55,9 @@ void TTY::set_default_termios()
 KResultOr<size_t> TTY::read(FileDescription&, size_t, u8* buffer, size_t size)
 {
     if (Process::current()->pgid() != pgid()) {
-        Process::current()->send_signal(SIGTTIN, nullptr);
+        KResult result = Process::current()->send_signal(SIGTTIN, nullptr);
+        if (result.is_error())
+            return result;
```
awesomekling, line 60: I'm not sure this is appropriate. A process should never fail to send a signal to itself (although I suppose our current slightly-dumb implementation might, since it only tries to signal the main thread of a process, instead of preferring it, and then trying any thread that doesn't block the signal...), but if for whatever reason it did, it doesn't seem right to fail `read()` with `ESRCH`.

Same comment for `TTY::write()`

bgianfo: Would you like me to suppress or assert then?

awesomekling: Yeah, please do :)

PR #3002, opened 2020-08-05
Build: Re-enable -Werror when building with GCC

bgianfo:
It looks like PR #2986 mistakenly removed this from both the
Clang and GCC CXX_FLAGS, when the intention seems to have been
to only disable it for Clang.

CC @nico


awesomekling:
*clickety click*

nico:
Whoops, sorry. The intention was to not remove it at all, that PR made things build with -Werror with clang.

PR #3003, opened 2020-08-05
AK: Fix StringUtils tests to actually observe return value.

bgianfo:
Noticed this while playing around with making Optional<T>
nodiscard.


awesomekling:
Oopsie!

PR #3005, opened 2020-08-05
AK: Decorate Optional<T> with [[nodisard]], suppress issues it found. 

bgianfo:
Continuing off of #2999 and #3001 this enables [[nodiscard]] for AK::Optional.

CC @alimpfard for the LibTLS / test-crypto suppressions
CC @BenWiederhake  for the LibCrypto CTR.h suppressions.

### DIFF
--- a/AK/Optional.h
+++ b/AK/Optional.h
@@ -34,7 +34,7 @@
 namespace AK {
 
 template<typename T>
-class alignas(T) Optional {
+class alignas(T) [[nodiscard]] Optional {
 public:
     Optional() {}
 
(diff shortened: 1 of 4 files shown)


BenWiederhake:
The discards in `crypto-test` are perfectly fine.

The discard in `CTR.h` is … weird. How is the caller supposed to know the new IV? I didn't think this through when implementing decryption or its test. This needs to be changed in the base class `Mode` and sibling class `CBC`.

The discard in LibTLS looks like a bug. How can LibTLS possibly know the new IV? Someone should probably have a look at that. @Petelliott maybe?

alimpfard:
> How can LibTLS possibly know the new IV

It doesn't need that.
New IVs are generated for each record (a little above, random IV)

The data returned was initially meant to allow for chaining things together, the uses for that are extremely rare though (AFAIK)

PR #3007, opened 2020-08-05
AK: Add InputMemoryStream to replace the usage of BufferStream on read-only buffers.

asynts:
This is a bit of a follow up to #2903. Essentially there are still a few calls to `ByteBuffer::wrap(const void*, size_t)` which can't be removed that easily. Many of these calls exist because there is no constant version of `AK::BufferStream`.

This pull requests adds `AK::InputStream` which is an abstract class for byte streams that is implemented by `AK::InputMemoryStream`. It also replaces all usages to `BufferStream` and `ByteBuffer::wrap(const void*, size_t)` in `Libraries/LibDebug`.

Additionally, all calls to `ByteBuffer::wrap(const void*, size_t)` are replaced by `ByteBuffer::wrap(const_cast<void*>(data), size)` to make the behavior more visible, the constant overload is removed entirely.

I've used C++20 concepts to define all `operator>>` overloads at once. [GCC seems to have pretty good support for concepts][1], so this seemed reasonable.

  [1]: https://gcc.gnu.org/projects/cxx-status.html


AK/String.h
```diff
@@ -83,6 +83,11 @@ class String {
     {
     }
 
+    inline explicit String(ReadonlyBytes bytes, ShouldChomp shouldChomp = NoChomp)
```
awesomekling, line 87: `inline` is implied here.

AK/String.h
```diff
@@ -196,6 +201,18 @@ class String {
         return *this;
     }
 
+    String& operator=(std::nullptr_t)
+    {
+        m_impl = nullptr;
+        return *this;
+    }
+
+    inline String& operator=(ReadonlyBytes bytes)
```
awesomekling, line 211: `inline` is implied here.

AK/String.h
```diff
@@ -196,6 +201,18 @@ class String {
         return *this;
     }
 
+    String& operator=(std::nullptr_t)
```
awesomekling, line 205: Is this necessary?

AK/Stream.h
```diff
@@ -0,0 +1,169 @@
...
+    {
+        ASSERT(!error() && !fatal());
+    }
+
+    inline bool error() const { return m_error; }
+    inline bool fatal() const { return m_fatal; }
+
+    inline bool handle_error() { return exchange(m_error, false); }
+
+    inline operator bool() const { return !m_error && !m_fatal; }
```
awesomekling, line 48: `inline` is implied here.

AK/Stream.h
```diff
@@ -0,0 +1,169 @@
...
+protected:
+    mutable bool m_error { false };
+    mutable bool m_fatal { false };
+};
+
+}
+
+namespace AK {
+
+class InputStream : public virtual AK::Detail::Stream {
```
awesomekling, line 59: Why `virtual` inheritance?

AK/Stream.h
```diff
@@ -0,0 +1,169 @@
...
+class InputMemoryStream final : public InputStream {
+    friend InputMemoryStream& operator>>(InputMemoryStream& stream, String& string);
+
+public:
+    inline InputMemoryStream(ReadonlyBytes bytes)
+        : m_bytes(bytes)
+    {
+    }
+
+    inline bool eof() const override { return m_offset >= m_bytes.size(); }
```
awesomekling, line 108: More comments about `inline` being implied.

awesomekling:
Yep, look pretty cool :)

AK/Stream.h
asynts: I was thinking a bit ahead here, like `OutputStream` and `DuplexStream` classes a bit like the `std::istream`, `std::ostream` and `std::iostream` thing. `DuplexSteam` would inherit from both `InputStream` and `OutputStream` thus inheriting `Detail::Stream` twice, then virtual inheritance would be needed. I can remove this for now.

I actually had something like this ready, but I put it aside because I was simply touching too many things at once.

AK/String.h
asynts: I did use that in the implementation of `InputMemoryStream& operator>>(InputMemoryStream&, String&)`. I feel like this is a useful overload, what should I do otherwise?

AK/String.h
asynts: > A function defined entirely inside a class/struct/union definition, whether it's a member function or a non-member friend function, is implicitly an inline function.
  > https://en.cppreference.com/w/cpp/language/inline

I was not aware of this, I'll remove that in all the places I added it in.

AK/Concepts.h
```diff
@@ -0,0 +1,43 @@
...
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#pragma once
+
+#include <AK/StdLibExtras.h>
+
+namespace AK::Concepts {
+
+// clang-format off
```
nico, line 33: Out of interest, why did you need this? clang-format seems to format both this and the place below fine for me. What does it do for you? Which clang-format version are you on?

asynts: I used `clang-format` 10.0.1 but it seems to work for me now too. I opened a pull request for this: #3038.

Originally, I used a the other syntax like this:

~~~c++
template<typename Integral>
requires Concepts::Integral<Integral>
InputStream& operator>>(InputStream& stream, Integral& value)
{
    stream.read_or_error({ &value, sizeof(value) });
    return stream;
}
~~~

instead of

~~~c++
template<Concepts::Integral Integral>
InputStream& operator>>(InputStream& stream, Integral& value)
{
    stream.read_or_error({ &value, sizeof(value) });
    return stream;
}
~~~

but both work for me now.

PR #3012, opened 2020-08-05
AK: Decorate AK::OwnPtr::leak_ptr() + AK::NonnullOwnPtr::leak_ptr() with [[nodiscard]]

bgianfo:
We should always leak to an observed variable, otherwise
it's an actual leak. This is similar to AK::RefPtr::leak_ref()
which is also marked as [[nodiscard]].


awesomekling:
Very sensible

PR #3014, opened 2020-08-05
HackStudio: Open file from argv

linusg:
When HackStudio is invoked with one or more arguments it will attempt to treat the first argument as a project or source file and open it accordingly.

When/if HS is updated to have tabs for multiple files opened in parallel it would make sense to take all files from `argv` into account.

Also LaunchServer now associates ~`.files`~ `.hackstudio` files with HackStudio.

---

If you try to open a file that's not a project file as project file it still fails spectacularly (aka assert), but that's a separate (and existing) issue.

### DIFF
--- a/Base/home/anon/LaunchServer.ini
+++ b/Base/home/anon/LaunchServer.ini
@@ -12,6 +12,7 @@ wav=/bin/SoundPlayer
 txt=/bin/TextEditor
 frm=/bin/VisualBuilder
 font=/bin/FontEditor
+hackstudio=/bin/HackStudio
 *=/bin/TextEditor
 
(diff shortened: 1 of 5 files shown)


awesomekling:
Let's come up with a less generic extension than `.files` :)

linusg:
That's the existing HackStudio project file format though... 

https://github.com/SerenityOS/serenity/blob/master/Base/home/anon/little/little.files

`.hs`? `.hsproject`? `.project`?

awesomekling:
Sure I'm aware of that. All of your suggestions are better than `.files` IMO. Maybe `.project` is also on the too generic side.

PR #3015, opened 2020-08-05
LibGUI: Use set_cursor() in TextEditor::set_document()

linusg:
Instead of setting `m_cursor` directly to reset the cursor position, `TextEditor::set_document()` now uses `set_cursor()` which will call cursor change callback functions, if any.

This fixes a bug in HackStudio where the cursor information text would not update immediately after changing the active `TextDocument`, even though the cursor is always visibly being reset to 0, 0.

---

*screenshot of that bug:*

image


awesomekling:
Ah yes, excellent!

awesomekling:
Unrelated: that screenshot really highlights our lack of support for tab characters in `GUI::TextEditor` :)

PR #3020, opened 2020-08-05
Kernel: Send a SIGCHLD to the parent upon suspension of a process that is controlling a TTY

alimpfard:
Fixes the ^Z issue...again.

@nico your idea - send `SIGCHLD` to parent upon suspension.

### DIFF
--- a/Kernel/TTY/TTY.cpp
+++ b/Kernel/TTY/TTY.cpp
@@ -155,6 +155,10 @@ void TTY::emit(u8 ch)
         if (ch == m_termios.c_cc[VSUSP]) {
             dbg() << tty_name() << ": VSUSP pressed!";
             generate_signal(SIGTSTP);
+            if (auto process = Process::from_pid(m_pgid)) {
+                if (auto parent = Process::from_pid(process->ppid()))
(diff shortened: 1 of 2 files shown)


awesomekling:
Hey that makes sense

Kernel/TTY/TTY.cpp
```diff
@@ -155,6 +155,9 @@ void TTY::emit(u8 ch)
         if (ch == m_termios.c_cc[VSUSP]) {
             dbg() << tty_name() << ": VSUSP pressed!";
             generate_signal(SIGTSTP);
+            auto process = Process::from_pid(m_pgid);
```
awesomekling, line 158: What if this returns `nullptr`

alimpfard: _can_ that return `nullptr`?

AFAIK the pgid has to be a valid pid?

awesomekling: Valid when setting yeah. But just looking at this I get suspicious that maybe we're not clearing `m_pgid` when a process dies.

alimpfard: Makes sense.

PR #3021, opened 2020-08-05
Kernel+LibKeyboard+ProcFS: Store and expose the current kernel keymap

### LINKED ISSUE #2684: LibKeyboard: API to get current active keymap
It seems like there is no syscall or otherwise implemented API to get the current active keymap - this would be useful for the "Keyboard settings" application, which only shows a blank selection for now:

image

vkoskiv:
This set of patches enables storing the loaded keymap name in the kernel, and then querying it back by reading from `/proc/keymap`

<p align="left">
  
</p>

Once this PR is merged, I can modify the KeyboardSettings program to show the currently selected keymap.
Resolves #2684

### DIFF
--- a/Kernel/Devices/KeyboardDevice.h
+++ b/Kernel/Devices/KeyboardDevice.h
@@ -51,7 +51,9 @@ public:
     KeyboardDevice();
 
     void set_client(KeyboardClient* client) { m_client = client; }
-    void set_maps(Keyboard::CharacterMapData character_map);
+    void set_maps(const Keyboard::CharacterMapData& character_map, const String& character_map_name);
+
+    const String keymap_name() { return m_character_map.character_map_name(); }
 
     // ^CharacterDevice
     virtual KResultOr<size_t> read(FileDescription&, size_t, u8*, size_t) override;
--- a/Kernel/FileSystem/ProcFS.cpp
+++ b/Kernel/FileSystem/ProcFS.cpp
@@ -33,6 +33,7 @@
 #include <Kernel/CommandLine.h>
 #include <Kernel/Console.h>
 #include <Kernel/Devices/BlockDevice.h>
+#include <Kernel/Devices/KeyboardDevice.h>
 #include <Kernel/FileSystem/Custody.h>
 #include <Kernel/FileSystem/FileBackedFileSystem.h>
 #include <Kernel/FileSystem/FileDescription.h>
@@ -85,6 +86,7 @@ enum ProcFileType {
     FI_Root_inodes,
     FI_Root_dmesg,
     FI_Root_interrupts,
+    FI_Root_keymap,
     FI_Root_pci,
     FI_Root_devices,
     FI_Root_uptime,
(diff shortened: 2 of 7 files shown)


linusg:
Closes #2684 :^)

Libraries/LibKeyboard/CharacterMap.h
```diff
@@ -43,9 +43,13 @@ class CharacterMap {
 
     u32 get_char(KeyEvent);
     void set_character_map_data(CharacterMapData character_map_data);
+    void set_character_map_name(String& character_map_name);
```
awesomekling, line 46: Please use a `const String&` here, as `set_character_map_name` does not intend to mutate the passed-in `String`.
Also, you can omit the parameter name sine it's already part of the function name :)

Libraries/LibKeyboard/CharacterMap.h
```diff
@@ -43,9 +43,13 @@ class CharacterMap {
...
+
+    String character_map_name();
```
awesomekling, line 48: This should be `const` qualified, as it does not mutate the state of the `CharacterMap`.

Kernel/Syscalls/setkeymap.cpp
```diff
@@ -27,6 +27,8 @@
 #include <Kernel/Devices/KeyboardDevice.h>
 #include <Kernel/Process.h>
 
+#define CHAR_MAP_NAME_SIZE 50
```
awesomekling, line 30: No need for this to be a macro, instead just use a function-local `constexpr size_t`

Kernel/Devices/KeyboardDevice.h
```diff
@@ -51,7 +51,9 @@ class KeyboardDevice final : public IRQHandler
     KeyboardDevice();
 
     void set_client(KeyboardClient* client) { m_client = client; }
-    void set_maps(Keyboard::CharacterMapData character_map);
+    void set_maps(Keyboard::CharacterMapData character_map, String& character_map_name);
```
awesomekling, line 54: Both of these arguments should be const references.

Kernel/Devices/KeyboardDevice.h
```diff
@@ -51,7 +51,9 @@ class KeyboardDevice final : public IRQHandler
...
+
+    String keymap_name() { return m_character_map.character_map_name(); }
```
awesomekling, line 56: Please make this `const` qualified.

Kernel/FileSystem/ProcFS.cpp
```diff
@@ -372,6 +374,16 @@ Optional<KBuffer> procfs$interrupts(InodeIdentifier)
     return builder.build();
 }
 
+Optional<KBuffer> procfs$keymap(InodeIdentifier)
+{
+    KBufferBuilder builder;
+    JsonObjectSerializer<KBufferBuilder> json { builder };
+    auto name = KeyboardDevice::the().keymap_name();
+    json.add("keymap", name);
```
awesomekling, line 382: I would lose the temporary here:
```suggestion
    json.add("keymap", KeyboardDevice::the().keymap_name());
```

awesomekling:
Please don't add fixups to the last commit of the PR, add them to the relevant commit.
You can use the fixup feature of `git rebase -i` for this. :)

vkoskiv:
Fixup applied with a bit of rebase magic!

awesomekling:
Ayy nice :) Today we learned about git!

PR #3022, opened 2020-08-06
Kernel: Dequeue dying threads from WaitQueue

tomuta:
If a thread is waiting but getting killed, we need to dequeue
the thread from the WaitQueue so that a potential wake before
finalization doesn't happen.

Fixes Piano crashing the system upon exit.


awesomekling:
Ah, good fix.

PR #3026, opened 2020-08-06
Shell: Add Alt-d binding to forward-delete a word

nico:
Also add comments with the usual (and, currently, always) bindings for VWERASE and VKILL.


alimpfard:
Uhhh...Shell?
isn't this a LibLine thing?

awesomekling:
> Uhhh...Shell?
> isn't this a LibLine thing?

Oh dang u rite. I didn't even notice on the previous one!

PR #3027, opened 2020-08-06
LibLine: Add bindings for Alt-u, Alt-l, Alt-c

nico:
@alimpfard questions the utility of these shortcuts on IRC, to which I reply: Nonsense, nonsense, NONSENSE.


alimpfard:
It's almost worth setting it to fire off a request to some server that records the use of these (especially alt+u) :P

awesomekling:
I am learning so many new tricks

PR #3028, opened 2020-08-06
LibLine: Add binding for Alt-.

nico:
This one's even useful, doubly so since Shell doesn't support `!$` yet.


alimpfard:
Can this at least respect `m_configuration.split_mechanism`?

nico:
Per IRC, added a FIXME.

PR #3031, opened 2020-08-06
LibLine: Fix "word" handling for alt-d/u/l/c/f/b and ctrl-left/right

nico:
All these shortcuts treat consecutive alnums as a word, not consecutive
non-spaces.

For example, `alias KILL='kill -9'` can now be written by typing it
out lowercase, then hitting ctrl-a alt-f alt-u.

Ctrl-W still treats a word as a sequence of non-spaces. Alt-backspace
in a future patch will add the ability to backward-delete a word
that's a sequence of alnums.


awesomekling:
>alias kill='kill -9'

madman

PR #3033, opened 2020-08-06
LibLine: Add Alt-t shortcut for transposing words

nico:
It feels like this should be possible to implement this much shorter, but I don't see it right now.

Probably should extract a few "move left/right by a word" functions and use them in a few places (including here), which would probably help too.

### DIFF
--- a/Libraries/LibLine/Editor.cpp
+++ b/Libraries/LibLine/Editor.cpp
@@ -591,7 +591,7 @@ void Editor::handle_read_event()
                 do_cursor_right(Word);
                 m_state = InputState::Free;
                 continue;
-            case ctrl('H'): // ^[^H: alt-backspace: backward delete word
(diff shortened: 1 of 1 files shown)


Libraries/LibLine/Editor.cpp
```diff
@@ -644,6 +644,57 @@ void Editor::handle_read_event()
...
+                while (start > 0 && !isalnum(m_buffer[start - 1]))
+                    --start;
+                size_t start_gap = start;
+
+                // Move left over first word.
+                while (start > 0 && isalnum(m_buffer[start - 1]))
+                    --start;
+
+                if (start != start_gap) {
+                    // To swap the two words, swap each word (and the gap) individually, and then swap the whole range.
```
awesomekling, line 682: Ooh a classic programming interview question, applied in real life!

PR #3034, opened 2020-08-06
KeyboardSettings: Preselect the current keymap in the dropdown menu.

vkoskiv:
And the cherry on top. KeyboardSettings now checks `/proc/keymap` to see which keymap is currently in use,
and shows it preselected on the list. Neat!


awesomekling:
:star:

PR #3037, opened 2020-08-07
AudioServer: Lock the process veil after startup

bgianfo:
We can't do any file syscalls anyway because of the reduction of pledges,
so we might as well lock the veil anyway.


awesomekling:
Sure why not. :)

PR #3040, opened 2020-08-07
LibThread: remove redundant --m_level in unlock in case of a release

tryfinally:


tomuta:
Isn't decrementing and testing for ZF cheaper than comparing to 1?

tryfinally:
the point is that we avoid an extra memory write on unlocking.

PR #3041, opened 2020-08-07
[WIP] AK: Vector - refactor non-const methods to call const methods + tests

tryfinally:


awesomekling:
What is the improvement here?

awesomekling:
Also, please always use `clang-format`.

tryfinally:
WIP

PR #3042, opened 2020-08-08
Shell: Stop a for loop upon receiving two consecutive interruptions

alimpfard:
This does not work perfectly (just like every other shell...), if the
running program handles the signal (SIGINT in this case) and quits
cleanly, the shell cannot detect the interruption.
This is the case with our `sleep(1)` (<- @awesomekling):
```sh
for $(seq 1 100) { sleep 1 } # This can only be terminated via ^\ (and not ^C)
```


awesomekling:
Yuhp, that makes sense!

PR #3044, opened 2020-08-08
DistinctNumeric: Intentionally-incompatible types to type-check IDs and other numerics

BenWiederhake:
This will be mostly useful for the PR I intend to write next: creating a new type `tid_t` and making it incompatible to `pid_t`. There are some weird quirks in the Kernel because these ID spaces aren't properly separated. The idea is to do something like this:

```
typedef /* … */ pid_t;
typedef /* … */ tid_t;

void some_function() {
    pid_t a;
    tid_t b;
    b = a; // Compilation error: Distinct types, even though both are "basically" int!
}
```

Future uses could include coordinate spaces, indexes, units, etc.

Because `DistinctNumeric` is not useful on its own, I intentionally make this a *draft*, because probably the interface needs some changes.

I put this PR up for discussion to get some early feedback: Right now it's easy to change the design :)

---

(In case you're wondering: The 'distinction' comes from using the macros `__FILE__` and `__LINE__` as template parameters. This way, there are no accidental distinction, and accidental identity is very unlikely and can be detected by a simple lint script.)

### DIFF
--- a/AK/DistinctNumeric.h
+++ b/AK/DistinctNumeric.h
@@ -1,0 +1,308 @@
+/*
+ * Copyright (c) 2020, Ben Wiederhake <BenWiederhake.GitHub@gmx.de>
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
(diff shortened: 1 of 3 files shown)


AK/DistinctNumeric.h
```diff
@@ -0,0 +1,314 @@
...
+ * Furthermore, some of these overloads make more sense with specific types, like `a&&b` which should be able to operate
+ *
+ * I intentionally decided against overloading `&a` because these shall remain
+ * numeric types.
+ *
+ * The C++20 `operator<=>` would require, among other things `std::weak_equality`.
+ * Since we do not have that, it cannot be implemented.
+ *
+ * The are many operators that do not work on `int`, so I left them out:
+ * `a[b]`, `*a`, `a->b`, `a.b`, `a->*b`, `a.*b`.
```
alimpfard, line 70: Not sure if any of these are even useful to have for _numeric_ types, but you can probably use a little template magic (or `requires`, we have concepts...kinda) like `auto helper(a, b) -> decltype(a.b)`...probably.

BenWiederhake: Interesting! But I intend to stick to "just" numeric types. However, I expect to run into something like `some_vector[my_distinct_numeric]`, so I'll probably have to think hard about that, eventually.

BenWiederhake:
Changes since the last push:
- Use struct tags instead of file-line tuple (no false negatives anymore)
- Nicer names
- Try and fail to incorporate a nicer scheme like https://gist.github.com/alimpfard/a3b750e8c3a2f44fb3a2d32038968ddf

PR #3045, opened 2020-08-08
LibCore: close socket on dtor in UDPServer/TCPServer/LocalServer

tryfinally:


Libraries/LibCore/TCPServer.cpp
```diff
@@ -53,6 +53,7 @@ TCPServer::TCPServer(Object* parent)
 
 TCPServer::~TCPServer()
 {
+    ::close(m_fd);
```
elcuco, line 56: Why no `if` here? Because `TCPServer` ctor `assert()`s. However, `LocalServer` does not. We should choose one behavior for consistency.

tryfinally: LocalServer doesn't open the socket in the constructor.

PR #3052, opened 2020-08-09
disasm: For ELF files, disassemble .text section

nico:


nico:
Hm, looks like this requires making LibELF build in lagom, which looks a bit involved.

PR #3053, opened 2020-08-09
LibELF+Lagom: Work towards getting LibELF in Lagom

nico:
Mostly -Wformat fixes, some of which pointed out real (if benign) bugs.


nico:
This is not everything, there are a few "conversion from pointer to u32 loses information" warnings left.

nico:
actually, let's try to do this in the disasm pull request

PR #3057, opened 2020-08-09
Kernel: Apply type-checking to PID/TID/PGID/SID, demonstrate and fix bugs

BenWiederhake:
The Kernel used to have many ID-confusion bugs, where it treats some ID as another kind of ID. For example, `Process::send_signal` (on the hotpath for `kill(2)`) would treat the process ID as a thread ID, and misbehave accordingly.

This PR:
- Introduces a new type `DistinctNumeric` that enables type-checking otherwise "identical" types (e.g. `typedef int pid_t` and `typedef int tid_t`), see #3044 for previous discussion.
- Applies this type-checking to the Kernel (and only the Kernel, for sanity). In the process of this, I found a few bugs and weird-looking things, and annotated them accordingly.
- Demonstrates 2 bugs, by creating thests. These tests cannot be automated yet, due to other, unrelated bugs :/
- Fixes 3 ID-confusion bugs (`setpgid`, `tcsetpgrp`, `send_signal`)
- Probably introduces a few new locking issues, although I tried my best to avoid this. Can someone review this? Maybe @bgianfo ?

I know that in theory I should merge the three "PID/___ typing" commits, but they're already hard to read. Also, they provide compiling and working checkpoints for bisecting, in case that's necessary. Therefore, I would like to *not* squash them.

### DIFF
--- a/AK/Tests/TestDistinctNumeric.cpp
+++ b/AK/Tests/TestDistinctNumeric.cpp
@@ -1,0 +1,315 @@
+/*
+ * Copyright (c) 2020, Ben Wiederhake <BenWiederhake.GitHub@gmx.de>
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include <AK/DistinctNumeric.h>
+#include <AK/TestSuite.h>
+
+template<typename T>
+class ForType {
+public:
+    static void check_size()
+    {
+        TYPEDEF_DISTINCT_NUMERIC_GENERAL(T, false, false, false, false, false, false, TheNumeric);
+        EXPECT_EQ(sizeof(T), sizeof(TheNumeric));
+    }
+};
+
+TEST_CASE(check_size)
+{
+#define CHECK_SIZE_FOR_SIGNABLE(T)         \
+    do {                                   \
+        ForType<signed T>::check_size();   \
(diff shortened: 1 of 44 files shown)


AK/Tests/TestDistinctNumeric.cpp
```diff
@@ -0,0 +1,315 @@
...
+    EXPECT_EQ(-b, GeneralNumeric(-1));
+    EXPECT_EQ(a + b, b);
+    EXPECT_EQ(b * GeneralNumeric(42), GeneralNumeric(42));
+}
+
+/*
+ * FIXME: These `negative_*` tests should cause precisely one compilation error
+ * each, and always for the specified reason. Currently we do not have a harness
+ * for that, so in order to run the test you need to set the #define to 1, compile
+ * it, and check the error messages manually.
```
alimpfard, line 253: If you forego the pretty `static_asserts` for SFINAE, you can do something like this: https://gist.github.com/alimpfard/5b1fd04c1b109db6d31fba59cedcf896

I personally like clear error messages though :shrug:

alimpfard: (the `Option` thing is not required, it's just what I had on hand)

BenWiederhake: I bow before the C++ template adept! :bowing_man: 

Also, yeah, let's stick with clear error messages. The tests can still be run by hand, so it's not a huge deal.

Kernel/UnixTypes.h
```diff
@@ -320,6 +321,12 @@ typedef u32 gid_t;
 typedef u32 clock_t;
 typedef u32 socklen_t;
 typedef int pid_t;
+// Avoid interference with AK/Types.h and LibC/sys/types.h by defining *separate* names:
+TYPEDEF_DISTINCT_NUMERIC_GENERAL(pid_t, false, false, true, false, false, false, ProcessID);
```
bgianfo, line 325: This macro is very hard to read, can we use enums instead of blind bools?

Userland/Tests/Kernel/kill-pidtid-confusion.cpp
```diff
@@ -0,0 +1,171 @@
...
+static void* run_pz_t2_wrap(void* fd_ptr);
+static void run_pz_t2();
+
+int main(int, char**)
+{
+    // This entire function is the entirety of process PX.
+
+    // Time 0: PX forks into PZ (mnemonic: Zombie)
+    // Serenity doesn't support O_NONBLOCK for pipes yet, so
+    // sadly the test will hang if something goes wrong.
```
awesomekling, line 112: Hmm, `O_NONBLOCK` is implemented at the syscall level (read/write/etc), not at the File subclass level. What's the issue here about?

Kernel/UnixTypes.h
BenWiederhake: Right, while writing this I thought "we can do (something like this) in the end, when we know what we actually need". And then forgot to do that. Sure, will do.

Userland/Tests/Kernel/kill-pidtid-confusion.cpp
BenWiederhake: That comment is outdated, from a time when I tried to communicate pid/tid of "T2" back to the main process through a pipe. I abandoned that approach, as it was too much work, and also currently there's some other bug that makes any call to `dbg()`, `printf()`, `write()`, and probably most other things crash the process (here: PZ) under those circumstances. And I wanted to make the test work *now*, so that it can show *now* what the difference is.

So, in short: outdated comment, will remove.

BenWiederhake:
All addressed and rebased. :)

awesomekling:
This is a great idea. :)

Kernel/Process.cpp
```diff
@@ -790,9 +790,20 @@ void Process::terminate_due_to_signal(u8 signal)
...
-    // FIXME: PID/TID BUG
-    if (auto* thread = Thread::from_tid(m_pid.value())) {
-        thread->send_signal(signal, sender);
+    Thread* receiver_thread;
+    // Try to send it to the "obvious" main thread:
+    receiver_thread = Thread::from_tid(m_pid.value());
+    // If the main thread has died, there may still be other threads:
+    if (!receiver_thread) {
+        // The first one should be good enough.
+        // Neither kill(2) nor kill(3) specify any selection precedure.
```
awesomekling, line 799: I think we should also check that the receiving thread doesn't block this signal.

BenWiederhake: I believe `Thread::send_signal` handles that? If not, the bug was already there before the commit.

Also, now that I think more deeply about that, this introduces non-determinism: Assume a process with three threads: Main, T1, and T2. Main does `pthread_exit`. T1 (but not T2) masks a signal, say SIGINT. Then the process receives a SIGINT. Now it's implementation-defined which thread receives the signal, and thus whether the process dies or not. (Currently it will be received in the chronologically-first thread, but only due to the iteration order of `Process::for_each_thread`.)

Maybe POSIX is right and we should do it differently:

> ["The signal disposition is a per-process attribute: in a multithreaded application, the disposition of a particular signal is the same for all threads."](https://linux.die.net/man/7/signal)

PR #3060, opened 2020-08-09
Base: Replace downscaled 2048 16x16 app icon with a hand-drawn one

linusg:
Downscaling icons to 16x16 pretty much never works... in this case it's blurry af :upside_down_face:

image :arrow_right:  image

cc @alimpfard


awesomekling:
Great!

PR #3064, opened 2020-08-09
Kernel: Fix SlabAllocator initialization

tomuta:
Since the constructors are invoked after the slab allocator was
initialized, we need to make sure we don't use objects that use
default constructors, as those would mess up the state. An exception
to this is the m_lock, which we manually initialize and it will
then get re-initialized. But since that lock should be not in use
at that time anyway, it should be safe.

Also, for_each_allocator did not include the 128 byte allocator.

### DIFF
--- a/Kernel/CMakeLists.txt
+++ b/Kernel/CMakeLists.txt
@@ -1,3 +1,8 @@
+set(KERNEL_HEAP_SOURCES
+    Heap/SlabAllocator.cpp
+    Heap/kmalloc.cpp
+)
+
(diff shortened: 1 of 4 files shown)


awesomekling:
Perhaps a better option here would be to placement-new the SlabAllocators into some reserved static storage.
Then we wouldn't have to worry about all this stuff. What do you think?

tomuta:
I like that idea, I created a separate constructor list for the kernel heap code.

awesomekling:
Oh that's even nicer yeah :D

PR #3066, opened 2020-08-10
disasm: Insert symbol names in disassembly stream

nico:
The symbol name insertion scheme is different from objdump -d's.
Compare the output on Build/Userland/id:

* disasm:

        ...
        _start (08048305-0804836b):
        08048305  push ebp
        ...
        08048366  call 0x0000df56

        0804836b  o16 nop
        0804836d  o16 nop
        0804836f  nop

        (deregister_tm_clones (08048370-08048370))

        08048370  mov eax, 0x080643e0
        ...
        _ZN2AK8Utf8ViewC1ERKNS_6StringE (0805d9b2-0805d9b7):
        _ZN2AK8Utf8ViewC2ERKNS_6StringE (0805d9b2-0805d9b7):
        0805d9b2  jmp 0x00014ff2

        0805d9b7  nop

* objdump -d:

        08048305 <_start>:
         8048305:       55                      push   %ebp
        ...
         8048366:       e8 9b dc 00 00          call   8056006 <exit>
         804836b:       66 90                   xchg   %ax,%ax
         804836d:       66 90                   xchg   %ax,%ax
         804836f:       90                      nop

        08048370 <deregister_tm_clones>:
         8048370:       b8 e0 43 06 08          mov    $0x80643e0,%eax
        ...
        0805d9b2 <_ZN2AK8Utf8ViewC1ERKNS_6StringE>:
         805d9b2:       e9 eb f6 ff ff          jmp    805d0a2 <_ZN2AK10StringViewC1ERKNS_6StringE>
         805d9b7:       90                      nop

Differences:

1. disasm can show multiple symbols that cover the same instructions.
   I've only seen this happen for C1/C2 (and D1/D2) ctor/dtor pairs,
   but it could conceivably happen with ICF as well.

2. disasm separates instructions that do not belong to a symbol with
   a newline, so that nop padding isn't shown as part of a function
   when it technically isn't.

3. disasm shows symbols that are skipped (due to having size 0)
   in parenthesis, separated from preceding and following instructions.

### DIFF
--- a/AK/StringView.h
+++ b/AK/StringView.h
@@ -158,6 +158,13 @@ public:
         return !(*this == other);
     }
 
+    bool operator<(const StringView& other) const
+    {
+        if (int c = __builtin_memcmp(m_characters, other.m_characters, min(m_length, other.m_length)))
+            return c < 0;
+        return m_length < other.m_length;
+    }
+
     const StringImpl* impl() const { return m_impl; }
 
     String to_string() const;
--- a/AK/Vector.h
+++ b/AK/Vector.h
@@ -73,6 +73,7 @@ public:
         return *this;
     }
     ALWAYS_INLINE ElementType& operator*() { return m_vector[m_index]; }
+    ALWAYS_INLINE ElementType* operator->() { return &m_vector[m_index]; }
     size_t operator-(const VectorIterator& other) { return m_index - other.m_index; }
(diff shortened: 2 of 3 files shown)


awesomekling:
Very nice. Would be nice to demangle also.

PR #3070, opened 2020-08-10
AK: Don't keep equal JsonObject keys in the order vector twice

### LINKED ISSUE #3069: AK: JsonObject with key overwritten once or more serializes same value multiple times
Repro:

```cpp
#include <AK/JsonObject.h>

int main(int, char**)
{
    JsonObject json;
    json.set("foo", "bar");
    json.set("foo", "baz");
    dbg() << json.to_string();
}
```

Output:

```json
{"foo":"baz","foo":"baz"}
```

This is caused by how values are stored internally - key/value pairs in a `HashMap<String, JsonValue>` (so no duplicates are allowed) and key order in a `Vector<String>`, which happily takes the same key multiple times (which later affects serialization):

```cpp
private:
    Vector<String> m_order;
    HashMap<String, JsonValue> m_members;
```

```cpp
    void set(const String& key, JsonValue value)
    {
        m_order.append(key);
        m_members.set(key, move(value));
    }
```

linusg:
Fixes #3069.

### DIFF
--- a/AK/JsonObject.h
+++ b/AK/JsonObject.h
@@ -100,8 +100,9 @@ public:
 
     void set(const String& key, JsonValue value)
     {
+        if (m_members.set(key, move(value)) == HashSetResult::ReplacedExistingEntry)
+            m_order.remove(m_order.find_first_index(key).value());
         m_order.append(key);
-        m_members.set(key, move(value));
     }
 
     template<typename Callback>
--- a/AK/Tests/TestJSON.cpp
+++ b/AK/Tests/TestJSON.cpp
@@ -106,7 +106,7 @@ TEST_CASE(json_string)
     EXPECT_EQ(json.type(), JsonValue::Type::String);
     EXPECT_EQ(json.as_string().is_null(), false);
     EXPECT_EQ(json.as_string().length(), size_t { 1 });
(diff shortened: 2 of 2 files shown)


AK/JsonObject.h
```diff
@@ -100,6 +100,9 @@ class JsonObject {
 
     void set(const String& key, JsonValue value)
     {
+        auto key_index = m_order.find_first_index(key);
+        if (key_index.has_value())
+            m_order.remove(key_index.value());
```
awesomekling, line 105: Is it correct behavior that the order changes when you replace an entry?
Also, you can use the return value from `m_members.set` to know if you've just replaced an existing entry :)

linusg: > Is it correct behavior that the order changes when you replace an entry?

Debatable, but that's what I'd expect.

```cpp
json.set("some", "foo");
json.set("key", "bar");
json.set("some", "baz");
```

should IMO serialize to

```json
{"key": "bar", "some": "baz"}
```

```json
{"some": "foo", "key": "bar", "some": "baz"}
```

OR

but not

```json
{"some": "baz", "key": "bar"}
```

>  Also, you can use the return value from m_members.set to know if you've just replaced an existing entry :)

Neat! will update :+1:

PR #3076, opened 2020-08-10
Shell: Allow redirection of (most) builtins to work

### LINKED ISSUE #3072: `time` doesn't play well with pipes
`disasm /bin/id | head` prints 10 lines of output, as it should.

`time disasm /bin/id | head` doesn't limit output. It acts as if the `| head` bit didn't exist.

@alimpfard

alimpfard:
Theoretically closes #3072 in a way I do not like.
Some builtins cannot safely accept redirections (e.g. `fg`), and some cannot be redirected in a meaningful way (e.g. `bg`), these are all disallowed here.

It should be noted that this is still WIP and very likely not the best way to solve this particular issue.

As an aside, I really like the idea of `fork_and_resolve()` here (do some setup, run given function, return the value)

---

Some examples to try:
```sh
$ time ls | head > stuff
...
$ dirs > dirs
```

### DIFF
--- a/Shell/Builtin.cpp
+++ b/Shell/Builtin.cpp
@@ -33,52 +33,119 @@
 #include <sys/wait.h>
 #include <unistd.h>
 
+#define DISALLOW_REWIRING(rewirings, name)                               \
+    if (!rewirings.is_empty()) {                                         \
+        fprintf(stderr, "Shell: Rewiring `" #name "' is not allowed\n"); \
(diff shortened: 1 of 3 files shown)


alimpfard:
calling for comments: @awesomekling @nico @bugaevc (and anyone else that cares about this)

alimpfard:
I absolutely detest this, I'll do something else.

Issue #3073, opened 2020-08-10
Writing to a closed pipe seems to not send SIGPIPE but hangs instead


nico:
Run `disasm /bin/id | head`.

This prints 10 lines as expected, but the `disasm` process never terminates. It hangs in Process::do_write() -> Scheduler::yield() -> Processor::switch_context() according to System Monitor.

BenWiederhake:
Here's a good guess: When `head` terminates, it *should* close the pipe (as part of process destruction in the Kernel, not as a responsibility of `head`), and thus abort the pending `write` from `disasm`. So probably either the pipe-closing goes wrong (e.g. `Shell` still has an open fd to the reading end), or aborting the write goes wrong.

tomuta:
I think `Process::do_write` would have to also check for a new `FileDescription::has_error()` and either send the signal and return `EPIPE`.  And then I think `FIFO::detach` would have to mark the descriptor having an error once `m_readers` or `m_writers` drops to 0, so that it can unblock and check for the error.

nico:
Process::do_write calls description.can_write which returns true for FIFO if there are no readers, and then FIFO::write returns EPIPE if there are no readers. SO I think that part probably works.

It feels more like a file handle leak somewhere -- when `head` close()s the input fd (stdin in this case), there are still more refs left to that description object for some reason, and so the pipe isn't closed, and so the disasm process doesn't get a SIGPIPE.

It'd be nice if `ue` could flag pipe leaks somehow...

tomuta:
Actually, I think one of the leaks may be that `Thread::wait_on` may never return if the thread got killed, which means references to the descriptors it has on the stack wouldn't get cleaned up. I think we need to do a two-staged return, if killing a thread we need to allow `Thread::wait_on` to always return and then perform the actual destruction upon exiting the trap.

tomuta:
Not sure if that's exactly the case here, but it's one way to leak a reference...

nico:
Looks like this is a fairly boring fd leak in the shell. There's a fancy rewiring data structure that does:

```
        } else if (rewiring->fd_action == AST::Rewiring::Close::RefreshDestination) {
            ASSERT(rewiring->other_pipe_end);

            int pipe_fd[2];
            int rc = pipe(pipe_fd);
            if (rc < 0) {
                perror("pipe(RedirRefresh)");
                return IterationDecision::Break;
            }
            rewiring->dest_fd = pipe_fd[1];
[...]

PR #3077, opened 2020-08-10
Shell: Fix fd leak with pipes

nico:
Fixes the problem reported in #3073. While trying to write a test
for this, I thought I'd use

    Shell -c 'for i in $(seq 100) { echo $i }' | head -n 1

but that makes the cpu spin at 100% and doesn't terminate even
with this fix here. But at least piping disasm into head now works.

### DIFF
--- a/Shell/Shell.cpp
+++ b/Shell/Shell.cpp
@@ -535,6 +535,10 @@ RefPtr<Job> Shell::run_command(const AST::Command& command)
                 perror("dup2(run)");
                 return nullptr;
             }
+            // dest_fd is closed via the `fds` collector, but rewiring.other_pipe_end->dest_fd
+            // isn't yet in that collector when the first child spawns.
+            if (rewiring.other_pipe_end && close(rewiring.other_pipe_end->dest_fd) < 0)
+                perror("close other pipe end");
         }
 
         fds.collect();


nico:
Without this change, `Shell -c 'for i in $(seq 100) { echo $i }' | head -n 1` does terminate, but it spawns 100 echo subprocesses, which seems incorrect as well. So I claim this change here is progress :)

nico:
This makes the Shell -c test case go:

```
$ git diff Shell/
diff --git a/Shell/AST.cpp b/Shell/AST.cpp
index 64a9ef86c..e8716ed2c 100644
--- a/Shell/AST.cpp
+++ b/Shell/AST.cpp
@@ -777,6 +777,8 @@ RefPtr<Value> ForLoop::run(RefPtr<Shell> shell)
             if (!job || job->is_running_in_background())
                 continue;
             shell->block_on_job(job);
+            if (job->signaled() && job->termination_signal() == SIGPIPE)
+                break;
             if (job->signaled()
                 && (job->termination_signal() == SIGINT
                     || job->termination_signal() == SIGKILL
```

So I'm fairly confident this change here is correct.

(Up to `seq 1000000`, after this it gets slow. I'm guessing the for loop implementation or the `$()` implementation runs the whole seq command before it starts reading from it.)

PR #3079, opened 2020-08-10
WindowServer: Limit maximizing windows to top window in modal stack

### LINKED ISSUE #3074: WindowServer: Parent window of file picker becomes maximized as well
To reproduce:
* Open TextEditor
* Press Ctrl+O to open a file
* Open the window menu of the file picker dialog by clicking the icon in the top left corner
* Choose "maximize"
Both the file picker, and the TextEditor window are now maximized(!) Only the file picker should be maximized.

tomuta:
When maximizing a window that is blocked by a modal window, only
maximize the top window in the stack. However, if the stack is
minimized, restore all of them in addition.

Fixes #3074


awesomekling:
Thank you @tomuta!

PR #3080, opened 2020-08-10
Shell: Add test for 37d5e3e0df6 / #3073

nico:


nico:
Let's see if this is the right path for Travis.

nico:
The test passes incorrectly due to (at least) #3082.

nico:
The test was correct as it was, but until just now I didn't understand why.

PR #3084, opened 2020-08-11
Shell: Use setsid() instead of setpgid()

### LINKED ISSUE #3083: ctrl-z and `bg` print in Shell both print "killpg: No such process"
1. Start any GUI app in terminal (`te`, `br`)
2. Hit Ctrl-Z
3. Type `bg`

Expected: Does its thing.
Actual: After 2, prints "killpg(job): No such process". After 3, prints "killpg: No such process".

@alimpfard Sorry about being such an enthusiastic Shell user.

alimpfard:
This fixed #3083.


alimpfard:
Let's fix this in a way that also fixes pipelines

PR #3085, opened 2020-08-11
Shell: Move pipelined processes into one process group (+tiny fixes)

alimpfard:
This also deals with #3083.

### DIFF
--- a/Kernel/Syscalls/setpgid.cpp
+++ b/Kernel/Syscalls/setpgid.cpp
@@ -124,8 +124,13 @@ int Process::sys$setpgid(pid_t specified_pid, pid_t specified_pgid)
     ProcessGroupID new_pgid = specified_pgid ? ProcessGroupID(specified_pgid) : process->m_pid.value();
     SessionID current_sid = sid();
     SessionID new_sid = get_sid_from_pgid(new_pgid);
-    if (current_sid != new_sid) {
+    if (new_sid != -1 && current_sid != new_sid) {
         // Can't move a process between sessions.
+        return -EPERM;
+    }
+    if (new_sid == -1 && new_pgid != process->m_pid.value()) {
+        // The value of the pgid argument is valid, but is not
+        // the calling pid, and is not an existing process group.
         return -EPERM;
     }
     // FIXME: There are more EPERM conditions to check for here..
--- a/Shell/Shell.cpp
+++ b/Shell/Shell.cpp
@@ -518,15 +518,10 @@ RefPtr<Job> Shell::run_command(const AST::Command& command)
         perror("fork");
         return nullptr;
     }
+
     if (child == 0) {
-        setpgid(0, 0);
         tcsetattr(0, TCSANOW, &default_termios);
-        if (command.should_wait) {
-            auto pid = getpid();
-            auto pgid = getpgid(pid);
-            tcsetpgrp(STDOUT_FILENO, pgid);
-            tcsetpgrp(STDIN_FILENO, pgid);
-        }
+
         for (auto& rewiring : rewirings) {
 #ifdef SH_DEBUG
             dbgprintf("in %s<%d>, dup2(%d, %d)\n", argv[0], getpid(), rewiring.dest_fd, rewiring.source_fd);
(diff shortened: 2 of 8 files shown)


Kernel/Syscalls/setpgid.cpp
```diff
@@ -124,7 +124,7 @@ int Process::sys$setpgid(pid_t specified_pid, pid_t specified_pgid)
     ProcessGroupID new_pgid = specified_pgid ? ProcessGroupID(specified_pgid) : process->m_pid.value();
     SessionID current_sid = sid();
     SessionID new_sid = get_sid_from_pgid(new_pgid);
-    if (current_sid != new_sid) {
+    if (new_sid != -1 && current_sid != new_sid) {
```
alimpfard, line 127: I remain unsure whether this is the right thing to do, if the `pgid` doesn't exist, there will be no "new session" -- is `setpgid()` only allowed for valid, existing `pgid`'s, or can it also create a new group?

Shell/Shell.cpp
```diff
@@ -571,10 +566,27 @@ RefPtr<Job> Shell::run_command(const AST::Command& command)
...
+    if (command.should_wait) {
+        tcsetpgrp(STDOUT_FILENO, pgid);
+        tcsetpgrp(STDIN_FILENO, pgid);
+    }
+
     StringBuilder cmd;
     cmd.join(" ", command.argv);
 
-    auto job = Job::create(child, (unsigned)child, cmd.build(), find_last_job_id() + 1);
+    auto job = Job::create(child, pgid, cmd.build(), find_last_job_id() + 1, command.pipeline);
```
alimpfard, line 589: _Something_ isn't quite right here, as I see some stack corruptions here and there, but other than that, I think this is a pretty ok way of dealing with pipelines

Kernel/Syscalls/setpgid.cpp
alimpfard: Dr. POSIX says this is the correct behaviour (https://pubs.opengroup.org/onlinepubs/9699919799/functions/setpgid.html), and I am apparently very blind.

BenWiederhake: What I wanted to achieve while changing `get_sid_from_pid` is that it never "accidentally" returns -1. Also, 

`current_sid` should *never* be -1, no matter what happens. So if `new_sid==-1` (which should only happen with "invalid-ish" pgids), then `current_sid != new_sid` already catches this.

Did you ever observe a `sid` of -1? If so, we should probably add an assert there, and fix that issue first.

BenWiederhake: Btw, I like using this website, as that's what kling also uses: https://linux.die.net/man/3/setpgid

Shell/Shell.cpp
```diff
@@ -572,10 +567,27 @@ RefPtr<Job> Shell::run_command(const AST::Command& command)
...
+            command.pipeline->pgid = child;
+        }
+    }
+
+    pid_t pgid = is_first ? child : (command.pipeline ? command.pipeline->pgid : child);
+    if (setpgid(child, pgid) < 0)
+        perror("setpgid");
+
+    if (command.should_wait) {
+        tcsetpgrp(STDOUT_FILENO, pgid);
```
BenWiederhake, line 583: You unconditionally assign a new process group to the new children. This is fine.

You conditionally allow those children to write to the terminal, only if they run in the background? That doesn't sound right to me. Did I misread the logic here?

Kernel/Syscalls/setpgid.cpp
BenWiederhake: *coin drops*

I've been misreading this diff the whole time. This diff is 100% correct. I introduced a bug by disallowing to change to a new session ID! :o

Shell/Shell.cpp
alimpfard: The shell immediately gets control (and the terminal) back if `!command.should_wait`.

alimpfard: (that is equivalent to `command&`)

BenWiederhake: Oh! Then I just got my logic mixed up. (I thought `command.should_wait` means that the *command* should "wait" by going to background.)

alimpfard: ouch.
funny story, it was originally called `should_be_run_in_background` or something like that.

Kernel/Syscalls/setpgid.cpp
BenWiederhake: Actually, now that I think about it, this is weird.
- Creating a new PGID should only be allowed if the new PGID=PID
- However, POSIX seems to be silent on this issue.

I'm strongly in favor of adding the requirement, because otherwise a malicious process (even a different user `nona`!) can guess PGIDs and "block" them, thus interfering with `setpgid` of `anon`'s processes.

I suggest implementing it like so:

```
    if (your previous expression)
        return -EPERM;
    if (new_sid == -1 && new_pgid != pid)
        return -EPERM;
```

alimpfard: That sounds reasonable to me.

BenWiederhake:
LGTM ^^

alimpfard:
There's still a little stack mishap around~

Shell/Shell.cpp
alimpfard: Now I'm truly perplexed, building with `-fno-omit-frame-pointer` makes the stack mishaps vanish into thin air
:whoosh:

I guess I'll leave it be for the time being until it resurfaces.

alimpfard:
:shrug:

PR #3088, opened 2020-08-11
TextEditor: Open links from Web::PageView using Desktop::Launcher

linusg:
So far, clicking on a link from the Markdown/HTML preview `Web::PageView` did nothing - now we pass that link to `Desktop::Launcher`, which will then open it in Browser, FileManager, another TextEditor instance etc.

### DIFF
--- a/Applications/TextEditor/TextEditorWidget.cpp
+++ b/Applications/TextEditor/TextEditorWidget.cpp
@@ -30,6 +30,7 @@
 #include <AK/URL.h>
 #include <LibCore/File.h>
 #include <LibCore/MimeData.h>
+#include <LibDesktop/Launcher.h>
 #include <LibGUI/AboutDialog.h>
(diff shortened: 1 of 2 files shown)


Applications/TextEditor/TextEditorWidget.cpp
```diff
@@ -483,8 +493,7 @@ void TextEditorWidget::set_path(const LexicalPath& lexical_path)
     m_name = lexical_path.title();
     m_extension = lexical_path.extension();
 
-    if (m_extension == "c" || m_extension == "cc" || m_extension == "cxx" ||
-        m_extension == "cpp" || m_extension == "h") {
+    if (m_extension == "c" || m_extension == "cc" || m_extension == "cxx" || m_extension == "cpp" || m_extension == "h") {
```
linusg, line 496: clang-format did this

PR #3089, opened 2020-08-11
Crypto: Make AES call malloc much less often

nico:
ByteBuffer::slice_view() allocates a new ByteBuffer object,
which as a RefPtr and everything.

Nowadays it should probably return a Bytes / Span<u8>, but AES was only
using slice_view() to extract the pointer again.  So just add ady a
range check at the top, and do pointer arithmetic to do the same thing
faster.

Reduces time to run `disasm /bin/id` by a bit under 40%,
from ~8.3s to ~5s (min-of-3 runs).

### DIFF
--- a/Libraries/LibCrypto/Cipher/AES.cpp
+++ b/Libraries/LibCrypto/Cipher/AES.cpp
@@ -67,6 +67,7 @@ void AESCipherKey::expand_encrypt_key(const ByteBuffer& user_key, size_t bits)
 
     ASSERT(!user_key.is_null());
     ASSERT(is_valid_key_size(bits));
+    ASSERT(user_key.size() >= bits / 8);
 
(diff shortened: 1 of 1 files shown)


alimpfard:
But then out goes the range access protection...

Perhaps a `ByteBuffer::offset_pointer(size_t offset)` would be better? or just using Span or whatever?

nico:
The function calls `    ASSERT(is_valid_key_size(bits));` up top, which does the range check.

alimpfard:
Nah, that just checks whether the key size is right, not that the buffer is big enough

alimpfard:
I guess all I'm saying is, please assert that there is enough data.

nico:
Ah I misunderstood that code, thanks for catching this. Added an assert.

Issue #3091, opened 2020-08-11
LibJS: Calling an undefined function causes everything afterwards to become undefined


Lubrsi:
For example, running this test:
```
<!DOCTYPE html>
<html>
	<head>
		<script>var a = {}; a.test()</script>
	</head>
	<body>
		<script>alert("It worked!")</script>
	</body>
</html>
```
This causes ```alert``` to become undefined:
```
Browser(26): Throwing JavaScript Error: TypeError, undefined is not a function (evaluated from 'a.test')
Browser(26):   (global execution context)
Browser(26): Throwing JavaScript Error: ReferenceError, 'alert' is not defined
Browser(26):   (global execution context)
```

linusg:
> Calling an undefined function

This is misleading, any exception will reproduce this. See #3092 for an explanation and fix :)

Lubrsi:
Ah, sorry about that. It's because I only remember seeing it happen after a script calls a missing function.

PR #3092, opened 2020-08-11
LibWeb+LibJS: Clear exceptions in each Document::run_javascript() call / Assert when exception is not cleared before Interpreter::run()

linusg:
**LibWeb: Clear exceptions in each Document::run_javascript() call**

We don't want to carry over exceptions across multiple `Document::run_javascript()` calls as `Interpreter::run()` and every of its exception checks will get confused - in this case there would be an exception, but not because a certain action failed.

Real-life example:

```html
<script>var a = {}; a.test()</script>
<script>alert("It worked!")</script>
```

The above HTML will invoke `Document::run_javascript()` twice, the first call will result in a `TypeError`, which is still stored during the second call. The interpreter will eventually call the following functions (in order) for the `alert()` invocation:

- `Identifier::execute()`
- `Interpreter::get_variable()`
- `Object::get()` (on the global object)

That last `Object::get()` call has an exception check which is triggered as we still carry around the exception from earlier - and eventually returns an empty value.

Long story short, the second script will wrongly fail with "ReferenceError, 'alert' is not defined".

**LibJS: Assert when exception is not cleared before Interpreter::run()**

This is to prevent bugs like #3091 (fixed in 04d47ea) in the future; we generally
don't want `Interpreter::run()` to be called if the interpreter still has an exception stored. Sure, it could clear those itself but letting users of the interpreter do it explicitly seems sensible.

### DIFF
--- a/Libraries/LibWeb/DOM/Document.cpp
+++ b/Libraries/LibWeb/DOM/Document.cpp
@@ -436,7 +436,11 @@ JS::Value Document::run_javascript(const StringView& source)
         parser.print_errors();
         return JS::js_undefined();
     }
-    return document().interpreter().run(document().interpreter().global_object(), *program);
+    auto& interpreter = document().interpreter();
+    auto result = interpreter.run(interpreter.global_object(), *program);
+    if (interpreter.exception())
+        interpreter.clear_exception();
+    return result;
 }
 
 NonnullRefPtr<Element> Document::create_element(const String& tag_name)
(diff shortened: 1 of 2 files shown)


Lubrsi:
Confirmed fixed 👍 . This also happened across external scripts so loading a site such as Twitter had a bunch of weird exceptions thrown such as "window" is not defined.

Libraries/LibWeb/DOM/Document.cpp
```diff
@@ -436,6 +436,7 @@ JS::Value Document::run_javascript(const StringView& source)
         parser.print_errors();
         return JS::js_undefined();
     }
+    document().interpreter().clear_exception();
     return document().interpreter().run(document().interpreter().global_object(), *program);
```
awesomekling, line 440: Wouldn't it be nicer to do an exception check *after* running code, and then clearing it?

linusg: yeah that makes sense :)

awesomekling:
Great fix. Good idea with the assertion too

PR #3094, opened 2020-08-11
AK+LibCrypto+LibTLS+Kernel: Switch the Cipher::Mode interface to use Span

alimpfard:
Following #3089, this gets rid of more ByteBuffer overhead, bringing the runtime of `disasm /bin/id` down to 2.5 seconds (best of 3)

and also (technically) doubles(-ish) the speed of all AES operations performed through the Mode API.

### DIFF
--- a/Kernel/Random.h
+++ b/Kernel/Random.h
@@ -68,12 +68,13 @@ public:
 
         typename CipherType::CTRMode cipher(m_key, KeySize);
 
-        auto wrapped_buffer = ByteBuffer::wrap(buffer, n);
-        m_counter = cipher.key_stream(wrapped_buffer, m_counter).value();
+        Bytes buffer_span { buffer, n };
+        auto counter_span = m_counter.span();
+        cipher.key_stream(buffer_span, counter_span, &counter_span);
 
         // Extract a new key from the prng stream.
 
-        m_counter = cipher.key_stream(m_key, m_counter).value();
+        cipher.key_stream(buffer_span, counter_span, &counter_span);
     }
 
     template<typename T>
--- a/Libraries/LibCrypto/Authentication/HMAC.h
+++ b/Libraries/LibCrypto/Authentication/HMAC.h
@@ -68,8 +68,11 @@ public:
         m_inner_hasher.update(message, length);
     }
 
+    TagType process(const ReadonlyBytes& span) { return process(span.data(), span.size()); }
     TagType process(const ByteBuffer& buffer) { return process(buffer.data(), buffer.size()); }
     TagType process(const StringView& string) { return process((const u8*)string.characters_without_null_termination(), string.length()); }
+
+    void update(const ReadonlyBytes& span) { return update(span.data(), span.size()); }
     void update(const ByteBuffer& buffer) { return update(buffer.data(), buffer.size()); }
     void update(const StringView& string) { return update((const u8*)string.characters_without_null_termination(), string.length()); }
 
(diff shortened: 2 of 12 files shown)


Kernel/Random.h
```diff
@@ -68,12 +68,13 @@ class FortunaPRNG {
 
         typename CipherType::CTRMode cipher(m_key, KeySize);
 
-        auto wrapped_buffer = ByteBuffer::wrap(buffer, n);
-        m_counter = cipher.key_stream(wrapped_buffer, m_counter).value();
+        Span<u8> buffer_span { buffer, n };
```
nico, line 71: We have the `Bytes` typedef for `Span<u8>`.

Libraries/LibCrypto/Authentication/HMAC.h
```diff
@@ -68,8 +68,11 @@ class HMAC {
         m_inner_hasher.update(message, length);
     }
 
+    TagType process(const Span<const u8>& span) { return process(span.data(), span.size()); }
```
nico, line 71: We have `ReadonlyBytes` for `Span<const u8>`.

Kernel/Random.h
alimpfard: Ah, I didn't see that.
will change.

Userland/test-crypto.cpp
```diff
@@ -99,7 +99,7 @@ int crc32_tests();
 
 // stop listing tests
 
-void print_buffer(const ByteBuffer& buffer, int split)
+void print_buffer(const Span<const u8>& buffer, int split)
```
awesomekling, line 102: Seems like this should be `ReadonlyBytes`

Libraries/LibTLS/Record.cpp
```diff
@@ -77,12 +77,12 @@ void TLSv12::update_packet(ByteBuffer& packet)
 
             if (m_context.crypto.created == 1) {
                 // `buffer' will continue to be encrypted
-                auto buffer = ByteBuffer::create_zeroed(length);
+                auto buffer = ByteBuffer::create_uninitialized(length);
```
awesomekling, line 80: Hmm?

Libraries/LibTLS/Record.cpp
```diff
@@ -77,12 +77,12 @@ void TLSv12::update_packet(ByteBuffer& packet)
...
                 size_t buffer_position = 0;
                 auto iv_size = iv_length();
 
                 // We need enough space for a header, iv_length bytes of IV and whatever the packet contains
-                auto ct = ByteBuffer::create_zeroed(length + header_size + iv_size);
+                auto ct = ByteBuffer::create_uninitialized(length + header_size + iv_size);
```
awesomekling, line 85: Hmm?

Libraries/LibTLS/Record.cpp
alimpfard: There's no need for that to be zeroed, it's initialised right there

`buffer`:
- L91: 0:(`packet.size() - header_size`)
- L98: (`packet.size() - header_size`):(`packet.size() - header_size + mac_size`)
- L102: (`packet.size() - header_size + mac_size`):`buffer.size()`
- (asserted at L103)

`ct`:
- L88: 0:(`header_size - 2`)
- L123: (`header_size - 2`):(`header_size`)
- L111: (`header_size`):(`header_size + iv_size`)
- L117: (`header_size + iv_size`):(`header_size + iv_size + length`)
- (asserted at L113)

awesomekling: I believe you, just seems like a sneaky unrelated change not mentioned in commit message or anywhere :)

Userland/test-crypto.cpp
alimpfard: _/me patches stuff and doesn't add it before amending_

Libraries/LibTLS/Record.cpp
alimpfard: I'll split it up

(that also served to prove it to myself...)

awesomekling:
:star:

PR #3095, opened 2020-08-11
ProcFS+SystemMonitor: Better support for multi-threaded applications

BenWiederhake:
This was my big secret plan behind #3057: Properly displaying stacks in SystemMonitor. :^)

Backstory: When trying to figure out what SystemMonitor was doing, I tried among other things using "kill" on a multi-threaded application, and ran into one of the bugs fixed by #3057 (PID/TID confusion in the `kill` syscall). Now that I know that SystemMonitor lists Threads, not processes, it is much clearer how to handle stacks!

Explanation: Until now, the "Stacks" tab displayed all stacks of a process, in plaintext format. This is a bit unwieldy. Because SystemMonitor lists the individual threads as distinct rows, the "Stack" (now singular) tab should instead show only a single stack: That of the selected thread.
The data comes from ProcFS, so now the individual stacks get listed as files in the new directory `/proc/<pid>/stacks/`. Thus, other tools automatically have access to this interface. (I found nothing that used `/proc/stacks`, so no changes needed.)

I considered and disregarded alternative designs:
- Putting it all into `/proc/all`: Total overkill, and would unnecessarily bloat that already-large file.
- Changing `/proc/<pid>/stack` to a JSON file: Kinda silly when the stacks themselves aren't JSON.
- Changing `/proc/<pid>/stack` to a JSON file, and changing all backtraces to JSON: Too much work, and the only real use currently is SystemMonitor, so the format of the individual stacks are fine already.

### DIFF
--- a/Applications/SystemMonitor/CMakeLists.txt
+++ b/Applications/SystemMonitor/CMakeLists.txt
@@ -7,7 +7,7 @@ set(SOURCES
     ProcessFileDescriptorMapWidget.cpp
     ProcessMemoryMapWidget.cpp
     ProcessModel.cpp
-    ProcessStacksWidget.cpp
+    ProcessStackWidget.cpp
     ProcessUnveiledPathsWidget.cpp
 )
 
--- a/Applications/SystemMonitor/ProcessModel.cpp
+++ b/Applications/SystemMonitor/ProcessModel.cpp
@@ -90,6 +90,12 @@ String ProcessModel::column_name(int column) const
         return "PID";
     case Column::TID:
         return "TID";
+    case Column::PPID:
+        return "PPID";
+    case Column::PGID:
+        return "PGID";
+    case Column::SID:
+        return "SID";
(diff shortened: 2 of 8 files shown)


Applications/SystemMonitor/CMakeLists.txt
```diff
@@ -7,7 +7,7 @@ set(SOURCES
     ProcessFileDescriptorMapWidget.cpp
     ProcessMemoryMapWidget.cpp
     ProcessModel.cpp
-    ProcessStacksWidget.cpp
+    ProcessStackWidget.cpp
```
awesomekling, line 10: Seems like this should be called `ThreadStackWidget` now, no? :)

BenWiederhake: [Indeed!](https://www.youtube.com/watch?v=bqxJdl5Wbwc)

PR #3096, opened 2020-08-11
Meta: Work towards enabling -Wmissing-declarations

BenWiederhake:
This PR is about symbol visibility (i.e., functions and global variables), dead code detection, and unavoidably some type checking due to "ninja-imports" that had to be removed.

In C/C++, symbols have different visibilities:
- `static`, which means: Other compilation units don't have access. This is sometimes useful when other compilation units use the same symbol name (but that's not what this PR is about). In other words, it tells the compiler that this symbol is only for this compilation unit. Therefore, if the compiler doesn't immediately see a use of a function/symbol, it can confidently report that it's dead code/memory.
- non-`static`, which means: Other compilation units *do* have access. They might "know about" the symbol through other means, for example by copy-pasting the declaration into their own code. This is what I call "ninja-importing". Hence the compiler can never be sure whether something is actually used or not. Neither does the linker (for some reason).
- And of course this is a simplification, but the other stuff doesn't matter in this PR.

This is what makes `-Wmissing-declarations` interesting: It flags any *definition* of a function/variable that isn't preceded by a *declaration*. This may sound silly, but it's the only consistent and predictable way to differentiate what we would simple regard as "header" and "implementation" file.

Ideally, we could switch `-Wmissing-declarations` on immediately. However, there are a few reasons against it:
- Code is constantly changing, and I'm sure some code will get merged between me making this PR and this PR getting accepted – if that code contains a `-Wmissing-declarations` violation, then the merge would fail miraculously. That's why I want to make it at least two different PRs.
- ~There are still some false-positives or places where I'm not sure. This particularly affects the Kernel and LibC. (Also, LibELF and Demos/DynamicLink.)~ See EDIT below.

<details>

OUTDATED, see "EDIT" below.

- No idea, not gonna touch it.
    - Kernel/Arch/i386/CPU.cpp:1081:17: void Kernel::enter_thread_context(Kernel::Thread*, Kernel::Thread*)
    - Kernel/Arch/i386/CPU.cpp:1170:17: void Kernel::context_first_init(Kernel::Thread*, Kernel::Thread*, Kernel::TrapFrame*)
    - Kernel/Arch/i386/CPU.cpp:1304:16: u32 Kernel::do_init_context(Kernel::Thread*, u32)
    - Kernel/Arch/i386/CPU.cpp:1347:17: void Kernel::pre_init_finished()
    - Kernel/Arch/i386/CPU.cpp:1360:17: void Kernel::post_init_finished()
- Called by boot.S. Maybe put it into a header?
    - Kernel/init.cpp:104:30: void Kernel::init()
    - Kernel/init.cpp:167:30: void Kernel::init_ap(u32, Kernel::Processor*)
    - Kernel/init.cpp:184:17: void Kernel::init_finished(u32)
- Not sure how to tell the compiler that the compiler is already using them. Also, maybe __cxa_atexit should go into StdLib.cpp?
    - Kernel/init.cpp:383:16: int Kernel::__cxa_atexit(void (*)(void*), void*, void*)
    - Kernel/StdLib.cpp:285:19: void __cxa_pure_virtual()
    - Kernel/StdLib.cpp:300:19: void __stack_chk_fail()
    - Kernel/StdLib.cpp:305:19: void __stack_chk_fail_local()
- Could maybe go into a new header. This would also provide type-checking for new modules.
    - Kernel/Modules/TestModule.cpp:31:17: void module_init()
    - Kernel/Modules/TestModule.cpp:40:17: void module_fini()
- Not sure how to handle this.
    - Libraries/LibC/crt0.cpp:41:5: int _start(int, char**, char**)
- Not sure how to tell the compiler that the compiler is already using them.
    - Libraries/LibC/cxxabi.cpp:48:5: int __cxa_atexit(AtExitFunction, void*, void*)
    - Libraries/LibC/cxxabi.cpp:58:6: void __cxa_finalize(void*)
- These are ninja-imported by other LibC/crt0 functions. Maybe we should have some kind of "internals.h" header.
    - Libraries/LibC/libcinit.cpp:36:6: void __libc_init()
    - Libraries/LibC/libcinit.cpp:55:19: void __stack_chk_fail()
    - Libraries/LibC/malloc.cpp:430:6: void __malloc_init()
    - Libraries/LibC/stdio.cpp:562:6: void __stdio_init()
- Not sure how to handle this.
    - Libraries/LibELF/DynamicLoader.cpp:334:23: Elf32_Addr ELF::_fixup_plt_entry(ELF::DynamicLoader*, u32)
- Should probably be marked `extern` to make sure it never gets "optimized away", but I didn't want to touch it.
    - Demos/DynamicLink/LinkLib/DynamicLib.cpp:60:6: void global_lib_function()
    - Demos/DynamicLink/LinkLib/DynamicLib.cpp:65:13: const char* other_lib_function(int)

</details>

EDIT: That is all. Now it build cleanly with `-Wmissing-declarations`. Please merge this!

I still want to wait a bit before I make the PR that sets the flag, because actually *setting* the flag will inevitably break some code that is currently in flight. I don't want to break the build due to that ^^'.


BenWiederhake:
> Code is constantly changing

QED! Rebased due to conflicting changes in `rm` ^^

BenWiederhake:
> Code is constantly changing

QED! Rebased due to conflicting changes in test-crypto

Libraries/LibCompress/Deflate.cpp
```diff
@@ -392,7 +390,8 @@ CanonicalCode::CanonicalCode(Vector<u8> codes)
     }
 }
 
-i32 binary_search(Vector<u32>& heystack, u32 needle)
+// FIXME: Use AK::Span::binary_search instead
```
stelar7, line 393: Span doesnt have a binary_search as far as i can see?
There is one in AK, but since it returns the element instead of the index, its not usable here

Libraries/LibC/crt0.cpp
```diff
@@ -38,6 +38,9 @@ extern void _init();
 extern char** environ;
 extern bool __environ_is_malloced;
 
+// Tell the compiler that this may be called from somewhere else.
+int _start(int argc, char** argv, char** env);
```
awesomekling, line 39: The commit message seems disconnected from the code change here.

Libraries/LibCompress/Deflate.cpp
BenWiederhake: Oh! I'll remove it then.

Libraries/LibC/crt0.cpp
BenWiederhake: I'm not sure what you mean? I agree that "LibC: Tell compiler about invisible call to _init" and "Tell the compiler that this may be called from somewhere else." isn't word-for-word identical, but both are about the same thing: Declaring a function, so that the compiler knows that calls into `_init` may happen from outside this compilation unit. Otherwise, if the function isn't declared (separately from its definition), the heuristic in `-Wmissing-declarations` thinks that no other place can call the function, and issues a warning that maybe the function should be `static`.

For special functions like `_start` or `__cxa_pure_virtual`, this is a bit silly, sure, but I can't change gcc to ignore those special functions .

awesomekling: The commit message talks about `_init` and you're adding a prototype for `_start`

BenWiederhake: Oh how silly XD

You're right, it's supposed to say `_start` everywhere.

BenWiederhake:
All done and rebased. I also fixed the commit description of the HackStudio commit (it used to talk about a file `common.h` that I instead called `HackStudio.h`)

awesomekling:
All right let's dew it!

PR #3097, opened 2020-08-11
UserspaceEmulator: Remove some silly semicolons

nico:
Wonder who put those here!


awesomekling:
Who was it!

nico:
I wonder!

Issue #3047, opened 2020-08-08
After successful build, 'make' remakes all bindings several times


BenWiederhake:
Some bindings get generated again and again and again. `Bindings/HTMLTitleElementWrapper.cpp
` alone gets built at least three times in the snippet below!

My best guess is that, for example, `Bindings/HTMLTitleElementWrapper.cpp` transitively depends on something else, let's say `Types.h`. Now `make` might see that `Bindings/HTMLTitleElementWrapper.cpp` is older than `Types.h` and therefore "rebuilds" it. Due to #2960, this does not change the mtime of `Bindings/HTMLTitleElementWrapper.cpp`. So the next time when `make` is run, the same thing happens! (And because cmake does complicate `make` invocations, `make` doesn't always remember that it was already invoked.)

- Either we keep it like this, where `Bindings/HTMLTitleElementWrapper.cpp` and it's dependencies get rebuilt again and again, which spams the console but doesn't really hurt because it's fast. (But it's annoying.)
- Or we roll back #2960, and everything that depends on `Bindings/HTMLTitleElementWrapper.cpp` gets rebuilt needlessly *once*.
- Or try to teach cmake that if the timestamp wasn't updated, it can safely `touch` that file and everything that depends on it. But I guess I'm trying to make cmake an [eierlegende Wollmilchsau](https://en.wiktionary.org/wiki/eierlegende_Wollmilchsau#Noun) :P

CC @nico 

Manual workaround: Make the final `mv -f …` call in `Meta/write-only-on-difference.sh` unconditional.

<details>

```
$ make -j1 # Parallelism isn't the cause of this
<… SNIP …>
[ 51%] Built target LibKeyboard
[ 51%] Built target LibLine
[ 51%] Built target LibMarkdown
[ 51%] Built target LibPCIDB
[ 51%] Generating ProtocolServerEndpoint.h
[ 51%] Built target generate_ProtocolServerEndpoint.h
[ 51%] Generating ProtocolClientEndpoint.h
[ 51%] Built target generate_ProtocolClientEndpoint.h
[ 52%] Built target LibProtocol
[ 52%] Built target LibTextCodec
[ 53%] Built target LibVT
[ 53%] Generating WebContentServerEndpoint.h
[ 53%] Built target generate_WebContentServerEndpoint.h
[ 53%] Generating WebContentClientEndpoint.h
[ 53%] Built target generate_WebContentClientEndpoint.h
[...]

PR #3098, opened 2020-08-11
Meta: Explain how to build with ninja

BenWiederhake:
Inspired by #3047, and my struggles to understand how cmake is supposed to work ^^

CC @Dexesttp, because he wrote the original description, and probably knows better commands for day-to-day use.

### DIFF
--- a/Documentation/BuildInstructions.md
+++ b/Documentation/BuildInstructions.md
@@ -107,7 +107,18 @@ Outside of QEMU, Serenity will run on VirtualBox. If you're curious, see how to 
 
 Later on, when you `git pull` to get the latest changes, there's no need to rebuild the toolchain. You can simply run `make install`, `make image`, `make run` again. CMake will only rebuild those parts that have been updated.
 
-You may also want to replace `make` with `ninja` in the above commands for some additional build speed benefits. To do this, go to an empty directory at the root and call `cmake .. -G Ninja` inside that directory. You might either create a new directory or reuse the existing `Build` directory after cleaning it.
+#### Faster than make: "Ninja"
+
+You may also want to replace `make` with `ninja` in the above commands for some additional build speed benefits, like reduced double-building of headers.
+Most of the process stays the same:
+- Go to an empty directory at the root (e.g. `Build/`) and call `cmake .. -G Ninja` inside that directory
+- You might either create a new directory or reuse the existing `Build` directory after cleaning it.
+- `make` becomes `ninja`
+- `make install` becomes `ninja install`
+- `make image` becomes `ninja image`
+- `make run` becomes `ninja run`
+
+Note that ninja automatically chooses a sane value for `-j` automatically, and if something goes wrong it will print the full compiler invocation. Otherwise, `ninja` behaves just like `make`. (And is a tad faster.)
 
 #### Ports
 To add a package from the ports collection to Serenity, for example curl, go into `Ports/curl/` and run **./package.sh**. The sourcecode for the package will be downloaded and the package will be built. After that, run **make image** from the `Build/` directory to update the disk image. The next time you start Serenity with **make run**, `curl` will be available.


Documentation/BuildInstructions.md
```diff
@@ -107,7 +107,16 @@ Outside of QEMU, Serenity will run on VirtualBox. If you're curious, see how to
...
-You may also want to replace `make` with `ninja` in the above commands for some additional build speed benefits. To do this, go to an empty directory at the root and call `cmake .. -G Ninja` inside that directory. You might either create a new directory or reuse the existing `Build` directory after cleaning it.
+#### Faster than make: "Ninja"
+
+You may also want to replace `make` with `ninja` in the above commands for some additional build speed benefits, like reduced double-building of headers.
+Most of the process stays the same:
+- Go to an empty directory at the root (e.g. `Build/`) and call `cmake .. -G Ninja` inside that directory
+- You might either create a new directory or reuse the existing `Build` directory after cleaning it.
+- `make` becomes `cmake --build .` (or possibly `cmake --build . --verbose -j 4`)
+- `make install` becomes `cmake --install .` (or possibly `cmake --install . --strip`, if you know what you're doing)
+- `make image` becomes `cmake --build . -- image`
```
bgianfo, line 118: Don't all the targets exist for ninja as well? 
- ninja 
- ninja install
- ninja image
- ninja run

BenWiederhake: Ah, for some reason I thought "ninja probably needs some extra weird arguments", looked at how `./BuildIt.sh` works, found `cmake --build . --target LibC`, and then it was all downhill from there ^^'

I'll simplify it then.

BenWiederhake: Done.

Dexesttp:
Hmm. Yeah, I think this wasn't clear enough in the original documentation, sorry ^^'

`cmake` is not a build tool. It is a tool that generates config files for build tools, but you cannot build anything with `cmake` directly (just like you cannot compile with `make` directly, the compiler will be `gcc` or `clang`).

By default, when you run `cmake ..`, it generates Makefile config files for your build system (equivalent to running `cmake .. -G Makefile`). Then, once your Makefile files are created properly, you just use the `make` command to use them, and `make` will call `gcc` and so on. It works just like if you wrote your Makefile yourself, except you didn't have to.

The alternative (and advantage of cmake) is that you can decide to generate Ninja config files instead of Makefile config files (by running the `cmake .. -G Ninja` command). Once you execute the command, the Ninja config files are read when you use the `ninja` command, and `ninja` will call `gcc` and so on. Same thing, you could have written your Ninja config files yourself, but you didn't have to (and Ninja files are harder to write by yourself than Makefile files).

The thing is that, since both the Makefile and Ninja config files are generated from the same "guidelines", it doesn't matter what tool you actually use, they will have the same generated subcommands. So, from the user's point of view, `ninja` becomes a drop-in replacement of  `make`.

This means that all the commands you are used to writing with `make` can be used identically with `ninja`, just with `s/make/ninja/` :
- `make` becomes `ninja`
- `make install` becomes `ninja install`
- `make image` becomes `ninja image`
- `make run` becomes `ninja run`
- `make my-custom-target --with-custom-options` becomes `ninja my-custom-target --with-custom-options`

And so on :)

BenWiederhake:
Thanks for all the feedback! I also included the thing about `-j`, that's really nice! :^)

selectiveduplicate:
Thanks for this; to me it's now become much easier to understand how to replace `make` with `ninja`.

BenWiederhake:
In case you actually mean it: Thanks, it was a journey for me, too ^^

In case this was sarcasm: Sure, if one already knows how to use `ninja`, then the explanation doesn't add any value. However, the `BuildInstructions.md` are, like all kinds of documentation, for the people who don't know yet how Stuff™ works. And as you can see from my initial attempts, my guesses about ninja were far from reality, and a short quip that `make == ninja` would have saved me a few minutes. So I *do* believe this change is an improvement.

selectiveduplicate:
> In case you actually mean it: Thanks, it was a journey for me, too ^^
> 
> In case this was sarcasm: Sure, if one already knows how to use `ninja`, then the explanation doesn't add any value. However, the `BuildInstructions.md` are, like all kinds of documentation, for the people who don't know yet how Stuff™ works. And as you can see from my initial attempts, my guesses about ninja were far from reality, and a short quip that `make == ninja` would have saved me a few minutes. So I _do_ believe this change is an improvement.

It wasn't sarcasm! I hardly know enough about build systems, let alone `ninja`; honestly this change has made the documentation better and clearer.

Dexesttp:
Well, it definitely feels like this documentation change is worth merging then !

@awesomekling any comments on the proposed updated documentation for ninja ?

PR #3102, opened 2020-08-12
ScrollBar: Let clicking the gutter scroll by one page

nico:
Shift-clicking has the old behavior of jumping to the click position.

This matches scrollbar behavior in macOS and Windows, and in many Linux apps.

Scrolling by page while the mouse is down isn't implemented yet. For now, this scrolls by one page for every click.


awesomekling:
Yah okay :)

PR #3103, opened 2020-08-12
LibGfx: Modularize StylePainter

Nullspeak:
This splits StylePainter into separate classes like WindowTheme. This allows for future themes to pick their own widget drawing engines, although at the moment only the original classic theme is implemented.


Libraries/LibGfx/ClassicStylePainter.cpp
```diff
@@ -0,0 +1,329 @@
+/*
+ * Copyright (c) 2018-2020, Andreas Kling <kling@serenityos.org>
```
awesomekling, line 2: I didn't add these classes, so feel free to take some credit and add yourself here as well (if you like.) :)

Nullspeak: Done, thanks :^)

PR #3104, opened 2020-08-12
Meta: Replace remaining LibM/math.h includes with math.h and ensure it's not being used going forward

linusg:


awesomekling:
Looks great, but let's allow Travis to chew on it :)

linusg:
build broken for seemingly unrelated reasons :grimacing:

awesomekling:
> build broken for seemingly unrelated reasons

Oops my bad :disappointed:

linusg:
not to worry, all green now!

BenWiederhake:
Fun fact: This broke the "heuristic" that tries to cache the toolchain. Finally I have very good reason to get rid of the heuristic, and write something that figures out the actual set of files used for libstdc++.a. ^^'

PR #3105, opened 2020-08-12
LibWeb: Fix #include <LibWeb/{DOM => HTML}/AttributeNames.h>

linusg:
This file has been moved from DOM/ to HTML/ in a784090b91139776b26fbac2a8426de3abdea308.

Unbreaks the build :)


alimpfard:
Yay! for once I'm not the one breaking the build

_libcrypto test flash backs_

PR #3106, opened 2020-08-12
AK: Complete the URL class

benit8:
This PR's goal is to complete the URL class.

First, I renamed "protocol" for "scheme" in most places (read below). Then I moved the `url_{en,de}code()` functions to the URL class, as static members. I added two similar functions: `URL::query_{en,de}code()`. The first ones comply to the RFC3689 and are used to encode path components in URLs, while the latter ones are used to encode query strings and form data (mime type application/x-www-form-urlencoded).

From there I started rewriting the parsing, using the `GenericLexer`. I shortly realized that data URLs are much different from regular ones, thus I created a helper class `URLHelper`. It can take over the parsing and serializing of URLs and provides a few rules to help with regular parsing. URL validation was also redone, but should definitely use regex if it ever becomes a part of serenity.

I created a sub-class for the data payloads, it contains a `ByteBuffer`, the mime-type and encoding (as an enum) of the payloads. The data is decoded upon parsing, and encoded when serialized, so no need to do it by hand.

Finally, I added query string handling. You can get the query fields as a `HashMap` and set them by providing the same structure.

And all of the above is tested with a bunch of new tests :^)

-- https://stackoverflow.com/a/48955169

In terms of programming, protocols would be like classes with their methods, and schemes would be strings.


awesomekling:
URL parsing should be implemented per the URL spec:
https://url.spec.whatwg.org/#concept-basic-url-parser

This looks over-engineered. Why do the helpers need reference counting?

benit8:
> Why do the helpers need reference counting?

I thought about using `RefPtr`s instead of raw pointers.
Maybe raw ones would have been enough?

PR #3108, opened 2020-08-12
LibC: fix realloc in case of allocation failure

tryfinally:
If the space cannot be allocated, the original memory block shall remain
unchanged and the function should return null

### DIFF
--- a/Libraries/LibC/stdlib.h
+++ b/Libraries/LibC/stdlib.h
@@ -42,7 +42,7 @@ __attribute__((malloc)) __attribute__((alloc_size(1))) void* malloc(size_t);
 __attribute__((malloc)) __attribute__((alloc_size(1, 2))) void* calloc(size_t nmemb, size_t);
 size_t malloc_size(void*);
 void free(void*);
-void* realloc(void* ptr, size_t);
+__attribute__((alloc_size(2))) void* realloc(void* ptr, size_t);
 char* getenv(const char* name);
 int putenv(char*);
 int unsetenv(const char*);
--- a/Libraries/LibC/malloc.cpp
+++ b/Libraries/LibC/malloc.cpp
@@ -405,7 +405,8 @@ void* calloc(size_t count, size_t size)
 {
     size_t new_size = count * size;
     auto* ptr = malloc(new_size);
-    memset(ptr, 0, new_size);
+    if (ptr)
+        memset(ptr, 0, new_size);
     return ptr;
 }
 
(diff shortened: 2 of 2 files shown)


awesomekling:
Note that malloc() does not fail; it will crash if unable to allocate.

tryfinally:
currently malloc_impl can return null , see commit 4291e96

Libraries/LibC/stdlib.h
```diff
@@ -42,7 +42,7 @@ __attribute__((malloc)) __attribute__((alloc_size(1))) void* malloc(size_t);
 __attribute__((malloc)) __attribute__((alloc_size(1, 2))) void* calloc(size_t nmemb, size_t);
 size_t malloc_size(void*);
 void free(void*);
-void* realloc(void* ptr, size_t);
+__attribute__((malloc)) __attribute__((alloc_size(2))) void* realloc(void* ptr, size_t);
```
awesomekling, line 45: This doesn't seem right. From [GCC documentation](https://gcc.gnu.org/onlinedocs/gcc/Common-Function-Attributes.html#Common-Function-Attributes):

>This tells the compiler that a function is malloc-like, i.e., that the pointer P returned by the function cannot alias any other pointer valid when the function returns, and moreover no pointers to valid objects occur in any storage addressed by P.

>Using this attribute can improve optimization. Compiler predicts that a function with the attribute returns non-null in most cases. Functions like malloc and calloc have this property because they return a pointer to uninitialized or zeroed-out storage. **However, functions like realloc do not have this property, as they can return a pointer to storage containing pointers.**

tryfinally: 👍 I missed that the definition of the malloc attributes was changed in recent gcc version
Up to gcc version 4.64 it was applicable.
https://gcc.gnu.org/onlinedocs/gcc-4.6.4/gcc/Function-Attributes.html

tryfinally: @awesomekling why is it so mem allocation will crash the app? isn't it more appropriate to let the application decide how to proceed?
if malloc/calloc/realloc will never fail, then may we should mark them as `returns_nonnull`

awesomekling: The majority of the system (both apps and libraries) is designed around many small heap-allocated objects. It would be massively obnoxious to have to null-check allocation everywhere. It's also far from obvious what to do in each case where an allocation might fail.

The thing that tends to matter in practice is large allocations and those generally go through `mmap` which can fail (if we run out of virtual address space, for example.)

PR #3109, opened 2020-08-12
LibGfx: fix transparency issues in GIFDecoder

### LINKED ISSUE #2990: LibGfx: The current Google Doodle appears corrupted
Here's how the [current Google Doodle](https://www.google.co.uk/logos/doodles/2020/wear-a-mask-save-lives-copy-6753651837108810-law.gif) appears in the browser:
AvARSxIuv2

peterdn:
This fixes #2990 caused by transparent pixels in GIF animation frames having their alpha values incorrectly set to zero, allowing the background behind the GIF to show through, instead of the previous animation frame.

Additionally, transparent pixels are now correctly identified based on their index matching the image transparency index, instead of their color values.

Lastly, the disposal method on a GIF animation frame now correctly applies to rendering of the next frame, which fixes issues with the RestoreBackground disposal method not working when applied to first frame.

Google Doodle from 05/08/2020 now renders correctly:

transparentgif


awesomekling:
Great!

PR #3110, opened 2020-08-12
AK: Add initial support for obscure IPv4 address notations

AlwaysLivid:
This change aims to add support for obscure IPv4 address notations, such as `1.1` (which equals to `1.0.0.1`), or the hypothetical address `1` (which is equal to `0.0.0.1`).

This is supported on other platforms as well, such as Linux, Windows, *BSD, and even Haiku. There isn't a specific RFC that specifies this, but here's a [man page](http://linux.die.net/man/3/inet_aton), if it helps.

### DIFF
--- a/AK/IPv4Address.h
+++ b/AK/IPv4Address.h
@@ -40,7 +40,7 @@ namespace AK {
 class [[gnu::packed]] IPv4Address
 {
 public:
-    IPv4Address() {}
+    IPv4Address() { }
     IPv4Address(const u8 data[4])
     {
         m_data[0] = data[0];
@@ -76,12 +76,37 @@ public:
         if (string.is_null())
             return {};
         auto parts = string.split_view('.');
-        if (parts.size() != 4)
+
+        u32 a;
+        u32 b;
+        u32 c;
+        u32 d;
+
+        if (parts.size() == 1) {
+            a = 0;
+            b = 0;
+            c = 0;
+            d = parts[1].to_uint().value_or(256);
+        } else if (parts.size() == 2) {
+            a = parts[1].to_uint().value_or(256);
+            b = 0;
+            c = 0;
+            d = parts[2].to_uint().value_or(256);
+        } else if (parts.size() == 3) {
+            a = parts[0].to_uint().value_or(256);
+            b = parts[1].to_uint().value_or(256);
+            c = 0;
+            d = parts[2].to_uint().value_or(256);
+        } else if (parts.size() == 4) {
+            a = parts[0].to_uint().value_or(256);
+            b = parts[1].to_uint().value_or(256);
+            c = parts[2].to_uint().value_or(256);
+            d = parts[3].to_uint().value_or(256);
+        } else {
             return {};
-        auto a = parts[0].to_uint().value_or(256);
-        auto b = parts[1].to_uint().value_or(256);
-        auto c = parts[2].to_uint().value_or(256);
-        auto d = parts[3].to_uint().value_or(256);
+        }
+
+
         if (a > 255 || b > 255 || c > 255 || d > 255)
             return {};
         return IPv4Address((u8)a, (u8)b, (u8)c, (u8)d);


AlwaysLivid:
The change hasn't been thoroughly tested and doesn't support some other obscure forms, which were brought to my attention on social media while debating about the change that was made in the IRC room and aren't consequently in the scope of this pull request. (e.g. decimal; `16843009` or `1.65793`). Nevertheless, my changes were written in a way that will leave the room to add support for these addresses later on, if needed. In order for that to happen, in combination with the necessity to declare the variables outside of the if... else clause while keeping the `return` part outside of it (in order to leave room for the if... else clause that checks whether any of the "parts" are higher than 255 for the time being).

An additional concern of mine that I'd like to discuss further is the way that there is memory that is being wasted in certain given cases.

AK/IPv4Address.h
```diff
@@ -76,12 +76,34 @@ class [[gnu::packed]] IPv4Address
...
+            a = parts[0].to_uint().value_or(256);
+            b = parts[1].to_uint().value_or(256);
+            c = 0;
+            d = parts[2].to_uint().value_or(256);
+        } else {
+            a = parts[0].to_uint().value_or(256);
+            b = parts[1].to_uint().value_or(256);
+            c = parts[2].to_uint().value_or(256);
+            d = parts[3].to_uint().value_or(256);
+        }
```
linusg, line 105: I think this should probably be another `else if (parts.size() == 4)` and then a final `else` returning `{}`? Otherwise this would silently accept and ignore more than 4 parts, no? (1.2.3.4.5.6)

AlwaysLivid: True, will correct.

AK/IPv4Address.h
```diff
@@ -76,12 +76,34 @@ class [[gnu::packed]] IPv4Address
...
+        if (parts.size() == 1) {
+            a = 0;
+            b = 0;
+            c = 0;
+            d = parts[1].to_uint().value_or(256);
+        } else if (parts.size() == 2) {
+            a = parts[1].to_uint().value_or(256);
+            b = 0;
+            c = 0;
+            d = parts[2].to_uint().value_or(256);
```
BenWiederhake, line 94: Have you actually tested it with your case of "1.65793"? I don't see how b and c can end up being set to 1 here.

BenWiederhake: Have a look at the `AK/Tests/` folder, the tests are really easy to write :)

AlwaysLivid: That case is a hypothetical extreme case scenario meant to illustrate what my change does **not** support.

AlwaysLivid: That case is a hypothetical extreme case scenario meant to illustrate what my change does **not** support.

AlwaysLivid: To clarify, my change is meant to deal with one particular scenario, and I tried my best to design it in a way that will let other people to be able to build on top of my change and support cases like `1.65793` later on.

AK/IPv4Address.h
AlwaysLivid: Corrected in 70b7687.

AK/IPv4Address.h
```diff
@@ -76,12 +76,37 @@ class [[gnu::packed]] IPv4Address
...
-        auto a = parts[0].to_uint().value_or(256);
-        auto b = parts[1].to_uint().value_or(256);
-        auto c = parts[2].to_uint().value_or(256);
-        auto d = parts[3].to_uint().value_or(256);
+        }
+
+
         if (a > 255 || b > 255 || c > 255 || d > 255)
             return {};
         return IPv4Address((u8)a, (u8)b, (u8)c, (u8)d);
```
Petelliott, line 112: I know you don't intend to support obscure forms like `1.65793`. But i would suggest that you should try to do it in this PR. It's actually quite a simple change to `IPv4Address::IPv4Address` with some bitwise arithmetic.

See the spoiler below if you want to know how I would do it. If not, thats fine too.

<details> 
  <summary> SPOILER</summary>
   <pre>
IPv4Address(u8 a, u8 b, u8 c, u32 d)
{
    m_data[0] = a | ((d >> 24) & 0xff);
    ...
}
   </pre>
</details>

AK/IPv4Address.h
```diff
@@ -76,12 +76,37 @@ class [[gnu::packed]] IPv4Address
...
             return {};
         auto parts = string.split_view('.');
-        if (parts.size() != 4)
+
+        u32 a;
+        u32 b;
+        u32 c;
+        u32 d;
+
+        if (parts.size() == 1) {
```
stelar7, line 85: This doesnt seem to be correct according to the man page you linked?
```
a
The value a is interpreted as a 32-bit value that is stored directly into the binary address without any byte rearrangement.
```

AK/IPv4Address.h
AlwaysLivid: I'm not familiar with this operating system and writing networking-related code yet. If I were comfortable with doing so, I'd have done it in the first place. I'd much rather do this at a later stage, which is what I'm currently preparing for, after having discussed the details with other people, rather than trying to act based on a code sample that I'm not sure I even understand.

AK/IPv4Address.h
AlwaysLivid: It's better than the previous implementation. As I tried to convey repeatedly earlier, I was thinking of adding another nested if case for the situations where some variables **exceed the decimal value of `255`** (which includes hex and decimal values). That way, I believe that the logic could be kept clean and understandable from a visual perspective, in accordance with the rest of the codebase. However, as described earlier, that's a step I am not sure I really want to take just yet within the scope of this pull request.

AlwaysLivid: It may seem a bit extreme to just do the same calculations for the same result, considering that one could achieve the same thing with a bitwise operator, but I do feel like this could be a way to catch more extreme cases, if necessary, as well as keeping things close to the original implementation and easily understood by everyone.

awesomekling:
Yeah this is obviously incomplete but since it covers your case of `1.1` it's still an improvement so let's roll with it. :)

AlwaysLivid:
As a note for the future; I tried re-implementing this entire thing in order to do it the [POSIX way](https://pubs.opengroup.org/onlinepubs/9699919799/functions/inet_addr.html), as instructed by Andreas.

Taking untrusted input and comparing every single part against the desired data type (depending on the amount of parts, which defines the expected size of the last part) will probably require a more mature status of the headers/parts that are made available in the operating system (e.g. the std namespace). The two other alternative approaches would be creating a custom implementation reminding of `std::numeric_limits`, or comparing against manually defined integers. I couldn't really manage to get past the validation stage because of that.

PR #3111, opened 2020-08-12
HackStudio: Add ArgsParser

matusf:
Hi guys, this PR adds ArgsParser to HackStudio. I also wanted to silence `make` when checking if make is installed since it prints the version information right above the help message when running `hs --help`. However, after adding `posix_spawn_file_actions_addopen` `make` fails with write error. Do you have any idea what I am getting wrong? Thanks!

Screenshot from 2020-08-13 00-36-02

### DIFF
--- a/DevTools/HackStudio/main.cpp
+++ b/DevTools/HackStudio/main.cpp
@@ -39,6 +39,7 @@
 #include "WidgetTool.h"
 #include "WidgetTreeModel.h"
 #include <AK/StringBuilder.h>
+#include <LibCore/ArgsParser.h>
 #include <LibCore/Event.h>
 #include <LibCore/EventLoop.h>
 #include <LibCore/File.h>
@@ -190,12 +191,14 @@ int main(int argc, char** argv)
     if (!make_is_available())
         GUI::MessageBox::show(g_window, "The 'make' command is not available. You probably want to install the binutils, gcc, and make ports from the root of the Serenity repository.", "Error", GUI::MessageBox::Type::Error);
 
-    String argument_absolute_path;
-    if (argc >= 2)
-        argument_absolute_path = Core::File::real_path_for(argv[1]);
+    const char* workspace_argument = nullptr;
+    Core::ArgsParser args_parser;
+    args_parser.add_positional_argument(workspace_argument, "Path to workspace", "workspace", Core::ArgsParser::Required::No);
+    args_parser.parse(argc, argv);
 
-    if (!argument_absolute_path.is_empty() && argument_absolute_path.ends_with(".hackstudio"))
-        open_project(argument_absolute_path);
+    auto workspace = Core::File::real_path_for(workspace_argument);
+    if (workspace.ends_with(".hackstudio"))
+        open_project(workspace);
     else
         open_project("/home/anon/Source/little/little.hackstudio");
 
@@ -692,8 +695,8 @@ int main(int argc, char** argv)
 
     g_open_file = open_file;
 
-    if (!argument_absolute_path.is_empty() && !argument_absolute_path.ends_with(".hackstudio"))
-        open_file(argument_absolute_path);
+    if (!workspace.is_empty() && !workspace.ends_with(".hackstudio"))
+        open_file(workspace);
     else
         open_file(g_project->default_file());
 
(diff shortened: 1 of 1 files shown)


DevTools/HackStudio/main.cpp
```diff
@@ -190,12 +193,14 @@ int main(int argc, char** argv)
...
-        argument_absolute_path = Core::File::real_path_for(argv[1]);
+    const char* workspace_argument = nullptr;
+    Core::ArgsParser args_parser;
+    args_parser.add_positional_argument(workspace_argument, "Path to workspace", "workspace", Core::ArgsParser::Required::No);
+    args_parser.parse(argc, argv);
 
-    if (!argument_absolute_path.is_empty() && argument_absolute_path.ends_with(".hackstudio"))
-        open_project(argument_absolute_path);
+    auto workspace = Core::File::real_path_for(workspace_argument);
+    if (workspace.ends_with(".hackstudio"))
```
Petelliott, line 202: why no `!workspace.is_empty()` check here?

DevTools/HackStudio/main.cpp
```diff
@@ -776,12 +781,16 @@ bool make_is_available()
 {
     pid_t pid;
     const char* argv[] = { "make", "--version", nullptr };
-    if ((errno = posix_spawnp(&pid, "make", nullptr, nullptr, const_cast<char**>(argv), environ))) {
+    posix_spawn_file_actions_t action;
+    posix_spawn_file_actions_init(&action);
+    posix_spawn_file_actions_addopen(&action, STDOUT_FILENO, "/dev/null", O_RDONLY, 0);
```
Petelliott, line 786: I think the reason make gives that error is you are opening `STDOUT_FILENO` for reading, rather than writing.

DevTools/HackStudio/main.cpp
matusf: Since [`ends_with`](https://github.com/SerenityOS/serenity/blob/master/AK/StringUtils.cpp#L186) already checks for `is_empty`.

DevTools/HackStudio/main.cpp
matusf: Thank you, change to `O_WRONLY` fixes the issue.

DevTools/HackStudio/main.cpp
Petelliott: makes sense.

DevTools/HackStudio/main.cpp
```diff
@@ -692,8 +695,8 @@ int main(int argc, char** argv)
 
     g_open_file = open_file;
 
-    if (!argument_absolute_path.is_empty() && !argument_absolute_path.ends_with(".hackstudio"))
-        open_file(argument_absolute_path);
+    if (!workspace.is_empty() && !workspace.ends_with(".hackstudio"))
+        open_file(workspace);
```
linusg, line 701: Not happy with the terminology here (and above) - I made it specifically so HackStudio will open either a project (probably what you mean by "workspace") **or** a regular file.

```cpp
open_file(workspace);
```

Doesn't really make sense.

matusf: Ah, I see. Changed it back to `argument_absolute_path`.

DevTools/HackStudio/main.cpp
```diff
@@ -190,11 +193,13 @@ int main(int argc, char** argv)
     if (!make_is_available())
         GUI::MessageBox::show(g_window, "The 'make' command is not available. You probably want to install the binutils, gcc, and make ports from the root of the Serenity repository.", "Error", GUI::MessageBox::Type::Error);
 
-    String argument_absolute_path;
-    if (argc >= 2)
-        argument_absolute_path = Core::File::real_path_for(argv[1]);
+    const char* workspace_argument = nullptr;
+    Core::ArgsParser args_parser;
+    args_parser.add_positional_argument(workspace_argument, "Path to workspace", "workspace", Core::ArgsParser::Required::No);
```
linusg, line 198: Please change the description to indicate this can also be a regular file :)

matusf: Changed

PR #3117, opened 2020-08-13
WindowServer: Handle global cursor tracking before ongoing drag/move/resize

### LINKED ISSUE #3116: WindowServer: Global mouse tracking events are not delivered while resizing windows
**Reproduction:**
The easiest way to reproduce this with the Eyes app. Resize the eyes' window (or any window on the workspace) and watch the pupils looking at the position where the cursor has started resizing that window.

https://imgur.com/YfaZ8L7

linusg:
In the case of an ongoing window drag/move/resize action `WindowManager::process_mouse_event()` would return early, even before delivering mouse events to windows with global cursor tracking enabled. They would only continue to receive new mouse events once those actions were completed.

Fixes #3116.

### DIFF
--- a/Services/WindowServer/WindowManager.cpp
+++ b/Services/WindowServer/WindowManager.cpp
@@ -871,6 +871,16 @@ void WindowManager::deliver_mouse_event(Window& window, MouseEvent& event)
 
 void WindowManager::process_mouse_event(MouseEvent& event, Window*& hovered_window)
 {
+    HashTable<Window*> windows_who_received_mouse_event_due_to_cursor_tracking;
+
+    for (auto* window = m_windows_in_order.tail(); window; window = window->prev()) {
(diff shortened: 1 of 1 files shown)


xTibor:
This introduced a regression. Windows now cannot be maximized by double clicking on their titlebars while an app is globally tracking the cursor.

linusg:
@xTibor thanks for catching this! Wasn't obvious to me when I did this little patch, now it is (`deliver_mouse_event()` does a little more than just "delivering" events)...

I tried to fix it in a sane way, but couldn't - a simple workaround would be replacing `deliver_mouse_event(*window, translated_event);` with `window->dispatch_event(translated_event);` - that'd make double-click window resizing work again, but windows with global cursor tracking would never receive double click events. Meh.

As it stands now it's either this bug or the one this patch fixed, so @awesomekling please decide which one is worse and possibly revert 23ea17d.

PR #3118, opened 2020-08-13
LibGUI: Fix ComboBox desktop intersection rect

### LINKED ISSUE #3115: LibGUI: Comboboxes act weird near the bottom of the screen
**Reproduction:**

Open an app that has comboboxes (Settings, Widget Gallery, etc.) and position its window such as that the combobox is near the bottom of the screen. Based on the distance from the bottom of the screen comboboxes could act in two weird way:

1. The dropdown is placed at the correct position but partially consumed by a black void.
2. The dropdown is displayed as a black rectangle at the top left corner of the screen.

https://imgur.com/ZUGyGlj

linusg:
Subtracting 128 from the desktop rect's height was far to much and and leading to weird rendering issues - now it's calculated exactly from taskbar and menubar heights as well as a little additional offset to make it fit perfectly.

Fixes #3115.

image

### DIFF
--- a/Libraries/LibGUI/ComboBox.cpp
+++ b/Libraries/LibGUI/ComboBox.cpp
@@ -186,8 +186,13 @@ void ComboBox::open()
         model()->row_count() * m_list_view->item_height() + m_list_view->frame_thickness() * 2
     };
 
+    auto taskbar_height = GUI::Desktop::the().taskbar_height();
+    auto menubar_height = GUI::Desktop::the().menubar_height();
+    // NOTE: This is so the combobox bottom edge exactly fits the taskbar's
+    //       top edge - the value was found through trial and error though.
+    auto offset = 8;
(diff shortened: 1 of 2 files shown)


Libraries/LibGUI/ComboBox.cpp
```diff
@@ -186,8 +186,13 @@ void ComboBox::open()
         model()->row_count() * m_list_view->item_height() + m_list_view->frame_thickness() * 2
     };
 
+    auto taskbar_height = 28;
+    auto menubar_height = 19;
```
awesomekling, line 190: Let's not hardcode this here, instead please add some getters to `GUI::Desktop`.
They can return hardcoded values, but at least we'll have them in a logical place if we want to make these dynamic in the future :)

awesomekling:
:star:

PR #3120, opened 2020-08-13
add tooltips to cpu and memory graphs

nico:


MenuApplets/ResourceGraph/main.cpp
```diff
@@ -118,6 +118,17 @@ class GraphWidget final : public GUI::Frame {
         }
     }
 
+    virtual void enter_event(Core::Event& event) override
+    {
+        // FIXME: Remember if tooltip is open and keep updating it while it's open.
```
awesomekling, line 123: How about we just `set_tooltip()` unconditionally after every update?

nico: I think that's better. But the main problem is that even static tooltips don't show up for MenuApplets :)

awesomekling: Oh I see! I guess because they look at `Window::rect()` which is empty. There is `Window::rect_in_menubar()` for applets. I'm not sure why those are separate APIs really. Maybe it should always be `Window::rect()` from the LibGUI side, makes things easier.

nico: Yep, see FIXME in https://github.com/SerenityOS/serenity/pull/3120/commits/f83f324b761f2a4b51aa523fe31e859c4d10ee6b :) @linusg said he's interested in looking at this on IRC :)

awesomekling: Oh duh, it says so right there! :)

nico:
Thanks to #3127, this works now :)

Once people have gigabytes of memory, maybe it's worth sometimes showing "GiB" instead of "MiB", but for now this is probably fine.

awesomekling:
:star:

PR #3121, opened 2020-08-13
Kernel: Don't request a random u32 when all but 5 bits are immediatel…

nico:
…y masked off


awesomekling:
Sure I guess! Gonna make us buffer up randomness? :)

PR #3123, opened 2020-08-13
Kernel: Fix rng regression from bc7a149039

nico:
@alimpfard @Petelliott Not sure what the implications of this were, but before bc7a149039 this refreshed m_key, so it's probably good if it keeps doing that.

### DIFF
--- a/Kernel/Random.h
+++ b/Kernel/Random.h
@@ -73,8 +73,8 @@ public:
         cipher.key_stream(buffer_span, counter_span, &counter_span);
 
         // Extract a new key from the prng stream.
-
-        cipher.key_stream(buffer_span, counter_span, &counter_span);
+        Bytes key_span = m_key.span();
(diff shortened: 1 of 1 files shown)


awesomekling:
oopsie

Kernel/Random.h
```diff
@@ -73,8 +73,8 @@ class FortunaPRNG {
         cipher.key_stream(buffer_span, counter_span, &counter_span);
 
         // Extract a new key from the prng stream.
-
-        cipher.key_stream(buffer_span, counter_span, &counter_span);
+        Bytes key_span = m_key.span();
+        cipher.key_stream(key_span, counter_span, &counter_span);
```
Petelliott, line 77: Setting the key is the correct behaviour. This looks like a classic copy-paste error when converting to `AK::Span`

alimpfard:
Oopsie indeed.

PR #3124, opened 2020-08-13
Kernel: mark strdup as __attribute__((malloc))

tryfinally:


awesomekling:
`strdup` was not used anywhere in the kernel so I've removed it in b777640fef90d1170fd5bd3ea43fe7a4df2f19d1

PR #3125, opened 2020-08-13
Kernel: Request random numberfs for syscall stack noise in larger chunks

nico:
Cuts time needed for `disasm /bin/id` from 2.5s to 1s -- identical
to the time it needs when not doing the random adjustment at all.

The downside is that it's now very easy to get the random offsets
with out-of-bounds reads, so it does make this mitigation less
effective.

--

Not sure if we want this, but it does make syscall-heavy programs much faster.

We can probably get the 2.5s down some more by optimizing the rng code more, but likely not this low.

### DIFF
--- a/Kernel/Syscall.cpp
+++ b/Kernel/Syscall.cpp
@@ -120,6 +120,10 @@ int handle(RegisterState& regs, u32 function, u32 arg1, u32 arg2, u32 arg3)
 
 }
 
+constexpr int RandomByteBufferSize = 256;
+u8 g_random_byte_buffer[RandomByteBufferSize];
+int g_random_byte_buffer_offset = RandomByteBufferSize;
+
 void syscall_handler(TrapFrame* trap)
 {
     auto& regs = *trap->regs;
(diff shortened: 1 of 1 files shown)


alimpfard:
Take the next byte and index into the array by that?
Not a perfect solution, but an exploit would now require two successful reads instead of one :shrug:

awesomekling:
Given the magnitude of the slowdown, I think the risk of buffering some randomness is acceptable.

Kernel/Syscall.cpp
```diff
@@ -135,7 +139,13 @@ void syscall_handler(TrapFrame* trap)
 
     // Apply a random offset in the range 0-255 to the stack pointer,
     // to make kernel stacks a bit less deterministic.
-    auto* ptr = (char*)__builtin_alloca(get_fast_random<u8>());
+    // Since this is very hot code, request random data in chunks instead of
+    // one byte at a time. This is a noticeable speedup.
+    if (g_random_byte_buffer_offset == RandomByteBufferSize) {
```
awesomekling, line 144: Hmm, doesn't this mean we'll not fill the buffer with randomness until the 256th syscall?

alimpfard: The main slowdown (I think) is `AESCipher` allocating a new `AESCipherBlock` when it's constructed, pulling data out is then allocation-free, so I'd say this is the right approach here

nico: Whoops!

awesomekling:
This basically made everything a lot faster.

PR #3126, opened 2020-08-13
disasm: Print correct offset-relative jumps in ELF file disassembly

nico:


Userland/disasm.cpp
```diff
@@ -141,7 +141,7 @@ int main(int argc, char** argv)
             is_first_symbol = false;
         }
 
-        out() << String::format("%08x", virtual_offset) << "  " << insn.value().to_string(offset);
+        out() << String::format("%08x", virtual_offset) << "  " << insn.value().to_string(virtual_offset);
```
awesomekling, line 144: Oh oops :)

PR #3127, opened 2020-08-13
LibGUI+WindowServer: Make tooltips work for menu applets

linusg:
cc @nico

Lots of debugging, pretty simple fixes. Never have I been so happy about a working tooltip :^)

Allow me to dump some commit messages.

---

**LibGUI: Add Window::window_type()**

We already have `Window::set_window_type()`, so a getter for `m_window_type` makes sense as well.

_Side note, in WindowServer-land this is just called `m_type` and `type()`, but as I'm saying - `set_window_type()` already is a thing, so `window_type()` made sense. (this is required for the next commit, btw)_

---

**LibGUI: Fix Widget::screen_relative_rect() for WindowType::MenuApplet**

This was using `window()->position()`, which is unset for windows with `WindowType::MenuApplet`. Now it checks the window type and then uses `rect_in_menubar()` for `MenuApplet` windows and `rect()` for everything else.

This makes tooltips show up for `MenuApplet` windows, previously they were positioned off-screen :^)

---

**WindowServer: Take MenuApplet windows into account for hovered_window**

This finally makes tooltips on menu applets the same as everywhere else!

Here's what went wrong:
`WindowManager::process_mouse_event()` receives a `Window*&`, determines the hovered window and sets it accordingly. However there's a branch that tests for `menubar_rect().contains(event.position())` and returns early - which resulted in `hovered_window` never being set to any `MenuApplet` window, even hovered ones.

The `hovered_window` result is being used in `WindowManager::event()` and passed to `WindowManager::set_hovered_window()`, which is responsible for creating `WindowLeft` and `WindowEntered` events when the hovered window changes, as a result of the mentioned chain of events this also never happens for `MenuApplet` windows. The `WindowLeft` event would the cause `Window::handle_left_event()` in LibGUI to be called, which unsets the window's hovered widget, which
is necessary for the widget to receive a subsequent `Enter` event - again, all of this never happened.

Now it's working as expected though, so we can start using tooltips on menu applets :^)

### DIFF
--- a/Libraries/LibGUI/Widget.cpp
+++ b/Libraries/LibGUI/Widget.cpp
@@ -464,7 +464,10 @@ Gfx::IntRect Widget::window_relative_rect() const
 
 Gfx::IntRect Widget::screen_relative_rect() const
 {
-    return window_relative_rect().translated(window()->position());
+    auto window_position = window()->window_type() == WindowType::MenuApplet
+        ? window()->rect_in_menubar().location()
+        : window()->rect().location();
+    return window_relative_rect().translated(window_position);
 }
 
 Widget* Widget::child_at(const Gfx::IntPoint& point) const
--- a/Libraries/LibGUI/Window.h
+++ b/Libraries/LibGUI/Window.h
@@ -82,6 +82,8 @@ public:
     void set_double_buffering_enabled(bool);
     void set_has_alpha_channel(bool);
     void set_opacity(float);
+
+    WindowType window_type() const { return m_window_type; }
     void set_window_type(WindowType);
 
     int window_id() const { return m_window_id; }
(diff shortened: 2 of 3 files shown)


nico:
Very cool!

Instead of https://github.com/SerenityOS/serenity/pull/3127/commits/7ca945bf84f5978bf0c5ff43963bc57762926a43 , what do you think about the WindowServer sending back a WindowResized() to the client from...somewhere (maybe `AppletManager::calculate_applet_rects()` so that MenuApplet windows have correct rects too? Then we could remove rect_in_menubar() from LibGUI and the whole MenuApplet stuff would mostly be WindowServer only. (I haven't looked at this in detail, maybe it doesn't work for some reason, but from a distance it seems like slightly nicer from a layering perspective to me.)

linusg:
Yep, that sounds good! Will try to do this later today, but if @awesomekling decides this is good enough for the moment feel free to merge it, don't want to block #3120 :)

awesomekling:
This is definitely good enough for a start, and @nico's suggestions are great too.
One more idea: it might be nice if we could get rid of the special branch for the menubar in WindowServer event handling and treat it like any other window (just one that happens to sit above other windows.)

Issue #2977, opened 2020-08-03
Inspector [Not responding] when try to inspect any process


umasankar-yedida:
When I'm playing around with Inspector, I observed weird behaviour. Follow below steps to reproduce.

1. Open Inspector
2. Open Calculator [same case with any application] from the System menu
3. Inspector won't show the calculator application. So, close it
4. Close the opened application as well
5. Reopen the Inspector
6. You can see that Calculator is still running you can also see another instance of Inspector too in the table. Weird
7. When try to inspect, that Inspector instance will show **Not responding** title

I'm not sure why the Calculator process is still running.

I'm new to Serenity and want to fix it myself. Please guide me to fix it

tomuta:
I can't seem to be able to reproduce this with most applications with the latest code, but I did find one that caused this:
```
[UserName.MenuApplet(19:19)]: Has not pledged accept
[UserName.MenuApplet(19:19)]: 0x00000000  EventLoop.cpp +0

[UserName.MenuApplet(19:19)]: 0xc0160070  _ZN6Kernel7Process5crashEijb +1044
[UserName.MenuApplet(19:19)]: 0xc0177674  _ZN6Kernel7Process10sys$acceptEiP8sockaddrPj +124
[UserName.MenuApplet(19:19)]: 0xc0164ed8  syscall_handler +1320
[UserName.MenuApplet(19:19)]: 0xc0164921  syscall_asm_entry +49
[UserName.MenuApplet(19:19)]: 0x080a25bb  _ZN4Core11LocalServer6acceptEv +67
[UserName.MenuApplet(19:19)]: 0x0809d535  _ZNK2AK8FunctionIFvvEE15CallableWrapperIZN4Core9EventLoop16start_rpc_serverEvEUlvE_E4callEv +41
[UserName.MenuApplet(19:19)]: 0x080a2f06  _ZN4Core6Object14dispatch_eventERNS_5EventEPS0_ +74
[UserName.MenuApplet(19:19)]: 0x0809cde5  _ZN4Core9EventLoop4pumpENS0_8WaitModeE +307
[UserName.MenuApplet(19:19)]: 0x0809cf63  _ZN4Core9EventLoop4execEv +43
[UserName.MenuApplet(19:19)]: 0x08048575  main +397
[UserName.MenuApplet(19:19)]: 0x08048dbe  _start +94
```
[...]

PR #3132, opened 2020-08-14
LibGUI: Add a timer in ProcessChooser to get processes for every 1000ms

umasankar-yedida:
Added a `timer` in `ProcessChooser` to get the process list for every 150ms.

This PR addresses https://github.com/SerenityOS/serenity/issues/2977#issuecomment-668109216.

### DIFF
--- a/Libraries/LibGUI/ProcessChooser.cpp
+++ b/Libraries/LibGUI/ProcessChooser.cpp
@@ -91,6 +91,34 @@ ProcessChooser::ProcessChooser(const StringView& window_title, const StringView&
     };
 
     table_view.model()->update();
+
+    m_refresh_timer = add<Core::Timer>();
+
+    m_refresh_timer->start(m_refresh_interval); // Start the timer to update the processes
+    m_refresh_timer->on_timeout = [&table_view] {
+        auto previous_selected_pid = -1; // Store the selection index to not to clear the selection upon update.
+        if (!table_view.selection().is_empty()) {
+            auto pid_as_variant = table_view.model()->data(table_view.selection().first(), GUI::Model::Role::Custom);
+            previous_selected_pid = pid_as_variant.as_i32();
+        }
+
+        table_view.model()->update();
+
+        if (previous_selected_pid == -1) {
+            return;
+        }
+
+        auto model = table_view.model();
+        auto row_count = model->row_count();
+        auto column_index = 1; // Corresponds to PID column in the table_view.
+        for (int row_index = 0; row_index < row_count; ++row_index) {
+            auto cell_index = model->index(row_index, column_index);
+            auto pid_as_variant = model->data(cell_index, GUI::Model::Role::Custom);
+            if (previous_selected_pid == pid_as_variant.as_i32()) {
+                table_view.selection().set(cell_index); // Set only if PIDs are matched.
+            }
+        }
+    };
 }
 
 ProcessChooser::~ProcessChooser()
--- a/Libraries/LibGUI/ProcessChooser.h
+++ b/Libraries/LibGUI/ProcessChooser.h
@@ -27,6 +27,7 @@
 #pragma once
 
 #include <LibGUI/Dialog.h>
+#include <LibCore/Timer.h>
 
 namespace GUI {
 
@@ -46,6 +47,10 @@ private:
     String m_window_title;
     String m_button_label;
     RefPtr<Gfx::Bitmap> m_window_icon;
+
+    bool m_refresh_enabled { true };
+    unsigned m_refresh_interval { 1000 };
+    RefPtr<Core::Timer> m_refresh_timer;
 };
 
 }


Libraries/LibGUI/ProcessChooser.cpp
```diff
@@ -91,10 +92,31 @@ ProcessChooser::ProcessChooser(const StringView& window_title, const StringView&
...
+        m_refresh_timer = add<Core::Timer>();
+
+        // Start the timer to update the processes
+        m_refresh_timer->start(m_refresh_interval);
+        m_refresh_timer->on_timeout = [&table_view] {
+            auto selection = table_view.selection();
+            table_view.model()->update();
+
+            // If there is already a selection, we select the same index
+            // FIXME We are selecting the same index instead of the PID
```
awesomekling, line 105: Indexes are logically invalid after a model update and cannot be safely reused.

Libraries/LibGUI/ProcessChooser.h
```diff
@@ -46,6 +53,10 @@ class ProcessChooser final : public GUI::Dialog {
     String m_window_title;
     String m_button_label;
     RefPtr<Gfx::Bitmap> m_window_icon;
+
+    bool m_refresh_enabled { true };
+    unsigned int m_refresh_interval { 150 };
```
awesomekling, line 58: This is very aggressive, let's go with 1000 ms instead.
Also, `unsigned int` -> `unsigned`.

Libraries/LibGUI/ProcessChooser.h
```diff
@@ -38,6 +39,12 @@ class ProcessChooser final : public GUI::Dialog {
 
     pid_t pid() const { return m_pid; }
 
+    bool is_refresh_enabled() { return m_refresh_enabled; }
+    void set_refresh_enabled(bool enabled) { m_refresh_enabled = enabled; }
+
+    unsigned int get_refresh_interval() { return m_refresh_interval; }
+    void set_refresh_interval(unsigned int interval) { m_refresh_interval = interval; }
```
awesomekling, line 46: Per [coding style](https://github.com/SerenityOS/serenity/blob/master/Documentation/CodingStyle.md), please omit `int` when using `unsigned` modifier.

Also, please don't use the `get_` prefix for getters, just use bare words (`refresh_interval()` in this case.)

Libraries/LibGUI/ProcessChooser.cpp
```diff
@@ -91,10 +92,31 @@ ProcessChooser::ProcessChooser(const StringView& window_title, const StringView&
...
+        };
+    }
 }
 
 ProcessChooser::~ProcessChooser()
 {
+    if (m_refresh_enabled) {
+        // Stop the timer
+        m_refresh_timer->stop();
+    }
```
awesomekling, line 118: This is not necessary, the timer stops by itself on destruction.

Libraries/LibGUI/ProcessChooser.cpp
```diff
@@ -91,10 +92,31 @@ ProcessChooser::ProcessChooser(const StringView& window_title, const StringView&
     };
 
     table_view.model()->update();
+
+    if (m_refresh_enabled) {
```
awesomekling, line 95: We are in the constructor here, so will I have no chance to change `m_refresh_enabled` before this happens.
So in practice, there's no way to turn off auto-refresh.

Libraries/LibGUI/ProcessChooser.h
alimpfard: hint hint `u32` looks better hint hint

Libraries/LibGUI/ProcessChooser.cpp
umasankar-yedida: I have thought of this but I'm not sure when to start a timer? Should I move this logic to another method named `populate_data` or `fill_table` would be better? So, whoever uses this must call that method to see the data in the table. Does that makes sense?

awesomekling: I would suggest just removing the setting and always auto-refreshing the table.

umasankar-yedida: Got it !!

Libraries/LibGUI/ProcessChooser.cpp
umasankar-yedida: Changed as suggested. Now the selection is based on PID

awesomekling:
Please squash the fixups into the first commit instead of stacking changes (see [contribution guidelines](https://github.com/SerenityOS/serenity/blob/master/CONTRIBUTING.md)) :)

umasankar-yedida:
> Please squash the fixups into the first commit instead of stacking changes (see [contribution guidelines](https://github.com/SerenityOS/serenity/blob/master/CONTRIBUTING.md)) :)

Done

awesomekling:
This does not build

umasankar-yedida:
> error: use of deleted function 'GUI::ModelSelection::ModelSelection(const GUI::ModelSelection&)'
   99 |         auto selection = table_view.selection();

Is it normal? But, when I run in my local, I didn't see it.

PR #3133, opened 2020-08-14
2048: Better scoring

alimpfard:
> The scoring system is a bit lame

No longer!
The added score is logarithmically proportional to the merged tiles, and reversely proportional to the game time (i.e. the more turns you take, the less score you get).

Also moves out the undo action to the app menu.

### DIFF
--- a/Games/2048/main.cpp
+++ b/Games/2048/main.cpp
@@ -75,6 +75,9 @@ int main(int argc, char** argv)
     app_menu.add_action(GUI::Action::create("New game", { Mod_None, Key_F2 }, [&](auto&) {
         game.reset();
     }));
+    app_menu.add_action(GUI::CommonActions::make_undo_action([&](auto&) {
+        game.undo();
+    }));
     app_menu.add_action(GUI::CommonActions::make_quit_action([](auto&) {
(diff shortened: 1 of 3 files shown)


awesomekling:
I just tried the original 2048 game and it gives you 8 points when combining 4+4 into an 8 tile, 16 points when combining 8+8 into a 16 tile, etc.

alimpfard:
But then that wouldn't incentivise finishing the game in less turns...
I could do that, but I just thought that this is more interesting.

Games/2048/main.cpp
```diff
@@ -75,6 +75,9 @@ int main(int argc, char** argv)
     app_menu.add_action(GUI::Action::create("New game", { Mod_None, Key_F2 }, [&](auto&) {
         game.reset();
     }));
+    app_menu.add_action(GUI::Action::create("Undo", { Mod_None, Key_U }, [&](auto&) {
+        game.undo();
+    }));
```
awesomekling, line 80: Let's use `GUI::CommonActions::make_undo_action` so we get the right icon and shortcut :)

alimpfard: Oh we have that? neat!

PR #3138, opened 2020-08-14
LibC: fix bsearch for large arrays

tryfinally:
fix  bsearch for large arrays due to index computation overflow and mixed signed/unsigned operations.

### DIFF
--- a/Libraries/LibC/stdlib.cpp
+++ b/Libraries/LibC/stdlib.cpp
@@ -745,18 +745,17 @@ char* mkdtemp(char* pattern)
 
 void* bsearch(const void* key, const void* base, size_t nmemb, size_t size, int (*compar)(const void*, const void*))
 {
-    int low = 0;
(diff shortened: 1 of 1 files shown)


awesomekling:
Please use commit messages to explain what you are changing and why.
From [contribution guidelines](https://github.com/SerenityOS/serenity/blob/master/CONTRIBUTING.md):

>**Don't:**
>Use weasel-words like "refactor" or "fix" to avoid explaining what's being changed.

BenWiederhake:
Awesome!

Is there a way to make `bsearch` a simple wrapper around `AK::binary_search`? This way, we would avoid having to fix the same bugs in several places, like it happened now with 615ba0f368d5012c7507a10086a54d1fcdb1fbc0 and your commit ^^

tryfinally:
I will into it. I think refactoring both methods to use same core.

PR #3144, opened 2020-08-14
Shell: Allow redirections and pipes on builtins

### LINKED ISSUE #3072: `time` doesn't play well with pipes

alimpfard:
Fixes #3072.

While this does indeed fix that, `time` will still fail to play well with pipes, as the command will run to completion before any other piped command after it is even invoked.
e.g.
```sh
time long-running-command foo bar | head
```
will wait for the entirety of `long-running-command`'s runtime before invoking `head`.

### DIFF
--- a/Shell/main.cpp
+++ b/Shell/main.cpp
@@ -56,6 +56,38 @@ void FileDescriptionCollector::add(int fd)
     m_fds.append(fd);
 }
 
+SavedFileDescriptors::SavedFileDescriptors(const NonnullRefPtrVector<AST::Rewiring>& intended_rewirings)
+{
+    for (auto& rewiring : intended_rewirings) {
+        int new_fd = dup(rewiring.source_fd);
+        if (new_fd < 0) {
+            if (errno != EBADF)
+                perror("dup");
+            // The fd that will be overwritten isn't open right now,
+            // it will be cleaned up by the exec()-side collector
+            // and we have nothing to do here, so just ignore this error.
+            continue;
+        }
+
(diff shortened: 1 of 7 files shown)


Shell/main.cpp
```diff
@@ -56,6 +56,31 @@ void FileDescriptionCollector::add(int fd)
     m_fds.append(fd);
 }
 
+SavedFileDescriptors::SavedFileDescriptors(const NonnullRefPtrVector<AST::Rewiring>& intended_rewirings)
+{
+    for (auto& rewiring : intended_rewirings) {
+        int new_fd = fcntl(0, F_DUPFD, rewiring.source_fd);
```
alimpfard, line 62: This is one weird interface...
`fcntl` already takes an fd as its first arg, _why_ would anyone think "hey, let's just ignore the perfectly fine fd argument and ask for a third argument for `F_DUPFD`"? hm Dr. POSIX? hm?

alimpfard: Hmmm...this needs `CLOEXEC`

awesomekling: Uh what's wrong with using `dup()` here?

alimpfard: Nothing wrong with `dup()`, I just wanted to whine about how weird `fcntl` is lol

PR #3146, opened 2020-08-14
FileManager: Apply wallpaper on startup

Lubrsi:
I noticed that nothing actually applies the wallpaper on startup.
I wasn't sure where to put the responsibility, so I gave it to
the desktop mode file manager.

Also adds a save_config option to set_wallpaper so it doesn't
needlessly save the config.

### DIFF
--- a/Applications/FileManager/main.cpp
+++ b/Applications/FileManager/main.cpp
@@ -40,6 +40,7 @@
 #include <LibGUI/Application.h>
 #include <LibGUI/BoxLayout.h>
 #include <LibGUI/Clipboard.h>
+#include <LibGUI/Desktop.h>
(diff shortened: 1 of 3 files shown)


awesomekling:
This is weird since `WindowServer` currently manages the wallpaper.
@tomuta talked about moving wallpaper rendering into the desktop FM instance at some point

tomuta:
> This is weird since `WindowServer` currently manages the wallpaper.
> @tomuta talked about moving wallpaper rendering into the desktop FM instance at some point

Well that was an idea, but this issue sounds different. What I was talking about is that it would simplify the rendering in `Compositor` if the bottom window of the screen was just a regular `Window` that always existed and was owned by WindowServer.

PR #3147, opened 2020-08-14
AK: HashTable add a constructor that allows preallocation of capacity + Use in CppLexer.

tryfinally:
1. Add general utility to get array number of elements.
2. Add Needed API to AK::HashTable
3. Refcator CppLexer to presize keywords and types hash tables top prevent extraneous rehashing.


awesomekling:
Where is this used?

tryfinally:
I want to initialize done hash tables to initial capacity to prevent rehashing. Specifically some global tables that gets populated and it make sense to allocate to sensible initial capacity.
Sensible as in measuring those with typical user.

Libraries/LibGUI/CppLexer.cpp
```diff
@@ -151,7 +151,7 @@ static bool is_keyword(const StringView& string)
 
 static bool is_known_type(const StringView& string)
 {
-    static HashTable<String> types;
+    static HashTable<String> types(101);
```
awesomekling, line 154: Please find a way to do this that doesn't require hard-coding the exact number of upcoming entries.
You could use a `constexpr` array of strings for example.

AK/StdLibExtras.h
```diff
@@ -514,4 +522,4 @@ using AK::max;
 using AK::min;
 using AK::move;
 using AK::RemoveConst;
-using AK::swap;
+using AK::swap;
```
awesomekling, line 525: Please configure your editor to leave a newline at the end of files.

AK/HashTable.h
```diff
@@ -164,6 +169,15 @@ class HashTable {
 
     HashSetResult set(const T&);
     HashSetResult set(T&&);
+
+    template<typename U, size_t N>
+    void setFrom(U (&from_array)[N])
```
awesomekling, line 174: `set_from`

PR #3149, opened 2020-08-14
AK: HashMap add a constructor that allows creation with initial capacity

tryfinally:


awesomekling:
Please don't add API's that are not used anywhere. Instead include them in the PR that actually uses them :)

bgianfo:
Hashtable commit is duped in #3147?

tryfinally:
Will restructure and submit new PR with API + usage

PR #3154, opened 2020-08-15
Travis: Toolchain only depends on headers, not impls

BenWiederhake:
When libstdc++ was added in 4977fd22b874fb9d6d089665e36badd03bcde827, just calling
'make install' was the easiest way to install the headers. And the headers are all
that is needed for libstdc++ to determine the ABI. Since then, BuildIt.sh was
rewritten again and again, and somehow everyone just silently assumed that
libstdc++ also depends on libc.a and libm.a, because surely it does?

Turns out, it doesn't! This massively reduces the dependencies of libstdc++,
hopefully meaning that the Toolchain doesn't need to be rebuilt so often on Travis.

Furthermore, the old method of trying to determine the dependency tree with
bash/grep/etc. has finally broken anyways:

> https://travis-ci.com/github/SerenityOS/serenity/builds/179805569#L567

In summary, this should eliminate most of the Toolchain rebuilds on Travis,
and therefore make Travis build blazingly fast! :^)

(In case you're wondering: I reduce the number of cached toolchains because the churn should go down anyway, and less caching means faster builds.)

### DIFF
--- a/Toolchain/ComputeDependenciesHash.sh
+++ b/Toolchain/ComputeDependenciesHash.sh
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
-set -eu
+set -euo pipefail
 # This file will need to be run in bash, for now.
 
 if [ $# -lt 1 ] ; then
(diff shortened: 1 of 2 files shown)


Toolchain/ComputeDependenciesHash.sh
```diff
@@ -23,89 +23,106 @@ function finish {
...
+#     strace -ff -e trace=file "make install-target-libstdc++-v3" 2>&1 >/dev/null | perl -ne 's/^[^"]+"(([^\\"]|\\[\\"nt])*)".*/$1/ && print' | sort -u | grep -P 'serenity/Build/Root/usr/include/.*\.h$'
+# This might change in the future, but I *extremely* doubt it.
+#
+# Note that this list contains files that don't exist in "our" LibC.
+# This is fine! We might need to recompile the toolchain if those headers define stuff that libstdc++ recognizes.
+# Note that this list don't contain all files of "our" LibC.
+# This is fine! libstdc++ most definitely doesn't depend on those symbols then.
+# Note that these files include files not on the list.
+# This is ... probably fine? The files are currently bits/endian.h, sched.h, stdarg.h, stdbool.h, stddef.h.
+# They are sufficiently unlikely to change breakingly. And if they do, then one of the files below will change anyway.
```
awesomekling, line 44: Wouldn't it be easier and more future-proof to just enumerate all `.h` files on our side?

BenWiederhake: - Advantage: "Future-proof", in case of changes that change the very small part that libstdc++ cares about. If this happens, then one of the other files will change, so I don't see the point.
- Disadvantage: False positives all over the place, like before. The point is to prevent Travis from constantly rebuilding the Toolchain, isn't it?

awesomekling: I mean `.h` files in our LibC. Wouldn't that be very narrow anyway, plus cover future GCC updates?
Or maybe I'm missing something obvious here.

BenWiederhake: Future gcc updates are covered anyway, as `BuildIt.sh` (which contains the gcc version string) and the gcc version string are (among other things) hashed into the "Toochain Hash", together with the files we're currently discussion. So gcc is already handled at this point.

Regarding `.h` files in our LibC: That's what the above code does: hashes most of our `.h` files, so that the Toolchain will be rebuilt whenever any of the files change, no matter how. It's only "most" files because libstdc++ doesn't look at the other `.h` files, and `strace`ing the build and installation of libstdc++ proves this.

So … I think you're suggesting to do what the code already does? :)

awesomekling: Well, I just see a bit static list of files and it makes me wonder if it could be dynamically generated, that's all :)

BenWiederhake: It can! You just need to `strace` the build process of `libstdc++`, which is a little bit slower than "just" building `libstdc++` directly, and also requires a freshly built gcc. That's why I decided against doing it dynamically. :P

awesomekling: What I don't understand is: why is the static list better than a simple `find Libraries/LibC -type f -name '*.h'`?
Also note that future changes to our LibC headers may cause the `libstdc++` configure script to reach different conclusions about feature availability and thus alter the dependency graph.

BenWiederhake: Because libstdc++ by far doesn't care about all of our headers, and I expect the probability of this happening *and* none of the `.h` changing to be negligible.

But okay, I'll change it: It seems we have very different estimates for "cost of unnecessary rebuild" and "probability of unrecognized ABI change" very differently.

BenWiederhake:
Done.

*Technically*, `sys/arch/i386/regs.h:29` depends on `AK/Types.h`, but this dependency is only for `u32`. I'll eat my hat if `u32` ever changes ABI-breakingly!

Also, the same typedef also appears in `stdint.h`, so maybe in the future someone wants to clean that up; but not me, not in this PR.
```
bits/stdint.h:34:typedef __UINT32_TYPE__ uint32_t;
../../AK/StdLibExtras.h:30:typedef __UINT32_TYPE__ u32;
../../AK/Types.h:34:typedef __UINT32_TYPE__ u32;
```

PR #3155, opened 2020-08-15
AK: Add BinaryStream

devsh0:


awesomekling:
Please don't add new APIs that are not used. Instead include them in the PR that uses them :)

devsh0:
Sure :)

That was the plan actually. But since it's unclear when I would be able to continue my last project (which used this API) I thought I should at least put this out to prevent further repetition of similar stream implementations in the Serenity codebase. But it makes sense to not add API unless they're actually used.

PR #3156, opened 2020-08-15
LibGUI: Update active tooltip when source widget changes the label

linusg:
`Application::show_tooltip()` now keeps track of the application's active tooltip source widget so it can be updated while being shown when the same widget updates its tooltip label.
`Application::hide_tooltip()` will unset the tooltip source widget, respectively.

This is pretty useful for the `ResourceGraph` applet's tooltips!

Also re-use the `Application::TooltipWindow`'s rect position in its `set_tooltip()` method to avoid flickering from the window temporarily being moved to 100, 100 and the position adjusted moments later.

---

Demo:

tooltip

### DIFF
--- a/Libraries/LibGUI/Application.h
+++ b/Libraries/LibGUI/Application.h
@@ -32,6 +32,7 @@
 #include <LibCore/Object.h>
 #include <LibGUI/Forward.h>
 #include <LibGUI/Shortcut.h>
+#include <LibGUI/Widget.h>
 #include <LibGfx/Forward.h>
 
 namespace GUI {
@@ -53,8 +54,9 @@ public:
     void register_global_shortcut_action(Badge<Action>, Action&);
     void unregister_global_shortcut_action(Badge<Action>, Action&);
 
-    void show_tooltip(const StringView&, const Gfx::IntPoint& screen_location);
+    void show_tooltip(const StringView&, const Gfx::IntPoint& screen_location, const Widget* tooltip_source_widget);
     void hide_tooltip();
+    Widget* tooltip_source_widget() { return m_tooltip_source_widget; };
 
     bool quit_when_last_window_deleted() const { return m_quit_when_last_window_deleted; }
     void set_quit_when_last_window_deleted(bool b) { m_quit_when_last_window_deleted = b; }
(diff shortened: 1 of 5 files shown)


Libraries/LibGUI/Application.h
```diff
@@ -53,8 +54,9 @@ class Application : public Core::Object {
     void register_global_shortcut_action(Badge<Action>, Action&);
     void unregister_global_shortcut_action(Badge<Action>, Action&);
 
-    void show_tooltip(const StringView&, const Gfx::IntPoint& screen_location);
+    void show_tooltip(const StringView&, const Gfx::IntPoint& screen_location, const RefPtr<Widget> tooltip_source_widget);
     void hide_tooltip();
+    RefPtr<Widget> tooltip_source_widget() const { return m_tooltip_source_widget; };
```
awesomekling, line 59: No need to return a `RefPtr` here let the caller decide if he wants to participate in ownership of the widget.
```suggestion
    Widget* tooltip_source_widget() { return m_tooltip_source_widget; }
```

Libraries/LibGUI/Widget.h
```diff
@@ -113,7 +114,7 @@ class Widget : public Core::Object {
 
     bool has_tooltip() const { return !m_tooltip.is_empty(); }
     String tooltip() const { return m_tooltip; }
-    void set_tooltip(const StringView& tooltip) { m_tooltip = tooltip; }
+    void set_tooltip(const StringView& tooltip);
```
awesomekling, line 117: I would drop the parameter name here since it's obvious from context.

Libraries/LibGUI/Application.h
```diff
@@ -53,8 +54,9 @@ class Application : public Core::Object {
     void register_global_shortcut_action(Badge<Action>, Action&);
     void unregister_global_shortcut_action(Badge<Action>, Action&);
 
-    void show_tooltip(const StringView&, const Gfx::IntPoint& screen_location);
+    void show_tooltip(const StringView&, const Gfx::IntPoint& screen_location, const RefPtr<Widget> tooltip_source_widget);
```
awesomekling, line 57: Oops, forgot to complain about this one. No need to take a `RefPtr<Widget>` here as that just incurs unnecessary refcount incrementing+decrementing. Just take a `Widget*` :)

PR #3157, opened 2020-08-15
Meta: Cleanup minor stuff

BenWiederhake:
Sorry for the mixed bag, but I think this is easier than making two trivial PRs.

- SystemMonitor: [You asked for it](https://github.com/SerenityOS/serenity/pull/3095#discussion_r469113354) :)
- `Meta/refresh-serenity-qtcreator.sh`: Can operate without `$SERENITY_ROOT`, and most other scripts get it from Travis or CMake. Building ports still requires it, so the hint should remain.


awesomekling:
Performance improvements? In the Qt Creator refresh script? **MERGE MERGE MERGE**

PR #3158, opened 2020-08-15
Travis: Don't use Python 2, clarify reason of cmake reinstall

BenWiederhake:
This removes the dependency on the dead language python2.

Also, when we upgrade the base linux image (in 2020 or maybe 2021?), this entire thing can hopefully be removed.

It feels very related to #3154, but I didn't want to muddle these two things.


alimpfard:
less python2 is a good thing! yay!

PR #3159, opened 2020-08-15
Userland: Provide a misbehaving application

BenWiederhake:
This will be a test case for #2977.

This shows that any "hanging" application currently takes down the inspector with it.

Maybe you don't want to merge it, but it's definitely going to be useful to demonstrate that #2977 can still be reproduced.


nico:
Feels like this fits in with the programs in Userland/Tests maybe?

BenWiederhake:
Not really, since the program has basically no way to determine whether Inspector is hanging.

nico:
They're (almost?) all manual tests I think, no?

BenWiederhake:
Those in `Userland/Tests/`? Not really, most of them seem to be intended to output either `PASS` or `FAIL`, and about half of them actually do.

PR #3164, opened 2020-08-15
Everywhere: Consistently use IEC prefixes for memory sizes

nico:
Windows uses "KB", "MB", "GB" as powers of two.
macOS uses "kB", "MB", "GB" as powers of ten.

"k", "M", "G" are SI standard prefixes that normally refer to powers of ten.

The IEC introduced "KiB", "MiB", "GiB" to unambiguously refer to
powers of two. It admittedly hasn't caught on that much, but it
does have the advantage that it's unabigious what it means.
So let's use it for user-visible sizes in SerenityOS.

(Linux does all of the above in different places, depending on app and
toolkit.)

--

This is admittedly fairly bike-sheddy, sorry. But SerenityOS was already self-inconsistent about this, and making some kind of decision on which style to use and using it consistently is probably a good idea. This was what sounded like the right decision per IRC :)

### DIFF
--- a/Applications/SystemMonitor/main.cpp
+++ b/Applications/SystemMonitor/main.cpp
@@ -559,7 +559,7 @@ NonnullRefPtr<GUI::Widget> build_graphs_tab()
         memory_graph.set_text_color(Color::Cyan);
         memory_graph.set_graph_color(Color::from_rgb(0x00bbbb));
         memory_graph.text_formatter = [](int value, int max) {
-            return String::format("%d / %d KB", value, max);
+            return String::format("%d / %d KiB", value, max);
         };
 
         self.add<MemoryStatsWidget>(memory_graph);
--- a/Applications/SystemMonitor/main.cpp
+++ b/Applications/SystemMonitor/main.cpp
@@ -31,8 +31,9 @@
 #include "ProcessFileDescriptorMapWidget.h"
 #include "ProcessMemoryMapWidget.h"
 #include "ProcessModel.h"
-#include "ThreadStackWidget.h"
 #include "ProcessUnveiledPathsWidget.h"
+#include "ThreadStackWidget.h"
+#include <AK/NumberFormat.h>
 #include <LibCore/Timer.h>
 #include <LibGUI/AboutDialog.h>
 #include <LibGUI/Action.h>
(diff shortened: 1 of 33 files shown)


Applications/SystemMonitor/main.cpp
```diff
@@ -559,7 +559,7 @@ NonnullRefPtr<GUI::Widget> build_graphs_tab()
         memory_graph.set_text_color(Color::Cyan);
         memory_graph.set_graph_color(Color::from_rgb(0x00bbbb));
         memory_graph.text_formatter = [](int value, int max) {
-            return String::format("%d / %d KB", value, max);
+            return String::format("%d / %d KiB", value, max);
```
alimpfard, line 562: wait what?

> This keeps just "K" instead of "KiB"

> -KB, +KiB

nico: "This just keeps K in the table columns". This one was iirc what's on the "filesystems" tab, not a table.

alimpfard: I'm not sure what's going on here, but the commit message seems a bit misleading/irrelevant to me. maybe that's just me though.

nico: I found it pretty clear, but I reworded it a bit, maybe it helps. Also fixed "captino" while here.

alimpfard: I guess I was just in need of some zzz's. sorry for the trouble~

Userland/allocate.cpp
```diff
@@ -33,18 +33,19 @@
 
 static void usage(void)
 {
-    printf("usage: allocate [number [unit (B/KB/MB)]]\n");
+    printf("usage: allocate [number [unit (B/kiB/MiB)]]\n");
```
Dexesttp, line 36: Casing (this actually doesn't match with the implementation too ^^')
```suggestion
    printf("usage: allocate [number [unit (B/KiB/MiB)]]\n");
```

Userland/allocate.cpp
```diff
@@ -33,18 +33,19 @@
...
     exit(1);
 }
 
 enum class Unit { Bytes,
```
Dexesttp, line 40: I dunno why the reformatting failed here, but...
```suggestion
enum class Unit {
Bytes,
```

Userland/allocate.cpp
nico: Looked through all of `git log -SkiB master..HEAD`, hopefully finally got them all.

Userland/allocate.cpp
nico: Done.

PR #3165, opened 2020-08-15
Shell+Kernel: TTY signals, ProcessGroup's and pipe stuff

### LINKED ISSUE #3160: Shell: Reading from a closed pipe is ... weird?
Create a simple file with two lines:

```
$ echo "hello\nfriends" > well.txt
```

Great. Let's look how `head` (which tries to read 10 lines) behaves:

- `$ head well.txt` → Works fine.
- `$ cat well.txt | head` → Works fine.
- `$ cat well.txt | head | head` → No output. Huh?
- `$ cat well.txt | head | head` → Hangs. What? But it's the same command! o.O

@alimpfard Please take it as a compliment, because I really like using a shell, and yours has all these nice features ^^

Edit: This *feels* related to #3073, dunno if it is. Just in case: CC @nico

Edit 2: Running this in Lagom works just fine. Huh.

alimpfard:
This fixes #3160 

I'm not sure if the "let everyone read from and write to a tty with no controlling process" thing is a very good idea though.

### DIFF
--- a/Kernel/Process.h
+++ b/Kernel/Process.h
@@ -40,6 +40,7 @@
 #include <Kernel/FileSystem/InodeMetadata.h>
 #include <Kernel/Forward.h>
 #include <Kernel/Lock.h>
+#include <Kernel/ProcessGroup.h>
 #include <Kernel/StdLib.h>
 #include <Kernel/Thread.h>
 #include <Kernel/UnixTypes.h>
@@ -159,8 +160,8 @@ public:
     ProcessID pid() const { return m_pid; }
     SessionID sid() const { return m_sid; }
     bool is_session_leader() const { return m_sid.value() == m_pid.value(); }
-    ProcessGroupID pgid() const { return m_pgid; }
-    bool is_group_leader() const { return m_pgid.value() == m_pid.value(); }
+    ProcessGroupID pgid() const { return m_pg ? m_pg->pgid() : 0; }
+    bool is_group_leader() const { return pgid().value() == m_pid.value(); }
     const FixedArray<gid_t>& extra_gids() const { return m_extra_gids; }
     uid_t euid() const { return m_euid; }
     gid_t egid() const { return m_egid; }
@@ -620,7 +621,7 @@ private:
 
     ProcessID m_pid { 0 };
     SessionID m_sid { 0 };
-    ProcessGroupID m_pgid { 0 };
+    RefPtr<ProcessGroup> m_pg;
 
     uid_t m_euid { 0 };
     gid_t m_egid { 0 };
--- a/Kernel/ProcessGroup.cpp
+++ b/Kernel/ProcessGroup.cpp
@@ -1,0 +1,70 @@
+/*
+ * Copyright (c) 2020, the SerenityOS developers.
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include "ProcessGroup.h"
+
+namespace Kernel {
+
+RecursiveSpinLock g_process_groups_lock;
+InlineLinkedList<ProcessGroup>* g_process_groups;
+
+ProcessGroup::~ProcessGroup()
+{
+    ScopedSpinLock lock(g_process_groups_lock);
+    g_process_groups->remove(this);
+}
+
+NonnullRefPtr<ProcessGroup> ProcessGroup::create(ProcessGroupID pgid)
+{
+    auto process_group = adopt(*new ProcessGroup(pgid));
+    {
+        ScopedSpinLock lock(g_process_groups_lock);
+        g_process_groups->prepend(process_group);
+    }
+
+    return process_group;
+}
+
+NonnullRefPtr<ProcessGroup> ProcessGroup::find_or_create(ProcessGroupID pgid)
+{
+    if (auto existing = from_pgid(pgid))
+        return *existing;
+
+    return create(pgid);
+}
+
+ProcessGroup* ProcessGroup::from_pgid(ProcessGroupID pgid)
+{
+    ScopedSpinLock lock(g_process_groups_lock);
+
+    for (auto& group : *g_process_groups) {
+        if (group.pgid() == pgid)
+            return &group;
+    }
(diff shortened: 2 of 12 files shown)


awesomekling:
The kernel change seems sketchy. PGID 0 is harmless (the kernel monopolizes that ID anyway.)
Shouldn't we figure out someone to inherit the TTY instead?

alimpfard:
I guess? I'll try giving it to another process in the group then.
inheriting it sounds better indeed.

Kernel/TTY/TTY.cpp
```diff
@@ -395,7 +399,26 @@ void TTY::hang_up()
...
+        return m_process->pgid();
+
+    // Try to find a new process in the same pgid.
+    if (m_last_assigned_pgid != 0) {
+        InterruptDisabler disabler;
+        Process::for_each_in_pgrp(m_last_assigned_pgid, [&](auto& process) -> IterationDecision {
+            m_process = process.make_weak_ptr();
+            return IterationDecision::Break;
+        });
+    }
```
awesomekling, line 412: What do other systems do?

alimpfard: Seems to me that at least linux doesn't care which process owns the tty, it just checks for the pgrp.
and `!pgrp` just makes it allow reads/writes:
```c
        pgrp = task_pgrp(current);

	tty_pgrp = tty->pgrp;
	
	if (tty_pgrp && pgrp != tty_pgrp) {
		if (is_ignored(sig)) {
			if (sig == SIGTTIN)
				ret = -EIO;
		} else if (is_current_pgrp_orphaned())
			ret = -EIO;
		else {
			kill_pgrp(pgrp, sig, 1);
			set_thread_flag(TIF_SIGPENDING);
			ret = -ERESTARTSYS;
		}
	}
```

BenWiederhake:
If you change the handling, you may also want to take a look at the `TIOCSPGRP` ioctl: Currently, it is permitted to `spgrp` to a PGID that doesn't exist, because the code doesn't distinguish between "group exists but has no leader" and "group doesn't exist".

Background: `m_process` is only used to send a SIGCHLD to the parent of `m_process` when the user presses the suspend sequence (`VSUSP`, so Ctrl-Z). In all other cases, it actually only needs a pgid.

alimpfard:
Is there anything I'm missing about `TIOCSPGRP`? 
or are you saying that we should just get rid of `m_process` and keep a link to the parent for `SIGCHLD`?

BenWiederhake:
I was just giving background information.

Getting rid of `m_process` and keeping a link to the parent for SIGCHLD sounds like a good plan.

Kernel/TTY/TTY.h
```diff
@@ -91,7 +91,8 @@ class TTY : public CharacterDevice {
     virtual bool is_tty() const final override { return true; }
 
     CircularDeque<u8, 1024> m_input_buffer;
-    WeakPtr<Process> m_process;
+    WeakPtr<Process> m_original_process_parent;
+    ProcessGroupID m_pgid { 0 };
```
awesomekling, line 95: But what happens when this PGID dies?

alimpfard: Nothing interesting.
we still act as if the pgid existed and `SIGTTIN`/`SIGTTOU` everyone that doesn't have the (now invalid) pgid.

BenWiederhake: So the shell has to "grab back" the tty? This makes a lot of sense, since something else had to "grab" the tty earlier.

Headsup: I don't know how it works on Linux, so maybe you need to detect whether the Shell is running in Lagom or in Serenity.

awesomekling: But then this is vulnerable to PID reuse attacks, no?

alimpfard: > PID reuse

well, it'd be a PGID reuse "attack", not sure what the use of such a thing would be.
In a sense, the TTY is owned by a PGID, not caring about who _is_ in that process group.

Not sure how we would detect that every process in a pgrp died (and even if that's a sensible thing to do!)

alimpfard: > I don't know how it works on Linux

Not sure about other places, but the TTY only tracks the PGID, and somehow detects "orphaned" pgrp's, disallowing writes.

awesomekling: > well, it'd be a PGID reuse "attack", not sure what the use of such a thing would be.

I'd rather we don't leave this door open. :)

alimpfard: Fixing _that_ specific problem would be a bit more involved - we'd have to figure out when all members of a pgrp die, but I don't think we have an abstraction for process groups, they're just a field on actual processes.

I'll mark this PR as a draft until we figure something out for this case then.

awesomekling: One solution would be to have a `ProcessGroup` object in the kernel.

BenWiederhake: One way would be to track *some* Process of the owning PGID, and whenever it dies, grab any other process with the right PGID. That's slow as hell, but process creation/destruction shouldn't happen that often, especially not without a leader.

This would get the PR unstuck, no?

alimpfard: Nope, that leaves a window of vulnerability where no one is in a process group:

pgid P: {A,B}
TTY proc: A
(A dies)
(B dies)
...
(C makes pgid P)
pgid P: {C}
(C writes to the TTY)
TTY looks for another member in the PGID and finds C
(TTY allows C to do whatever)

alimpfard:
Are there any tests for process groups? or will we have to wait and see what breaks?

alimpfard:
I did a bit of testing, nothing seems to be broken by this :shrug:

Kernel/ProcessGroup.cpp
```diff
@@ -0,0 +1,73 @@
...
+#include "ProcessGroup.h"
+
+namespace Kernel {
+
+RecursiveSpinLock g_process_groups_lock;
+InlineLinkedList<ProcessGroup>* g_process_groups;
+
+ProcessGroup::~ProcessGroup()
+{
+    {
```
awesomekling, line 36: These curlies, they do nothing

Kernel/ProcessGroup.cpp
```diff
@@ -0,0 +1,73 @@
...
+    }
+}
+
+NonnullRefPtr<ProcessGroup> ProcessGroup::create(ProcessGroupID pgid)
+{
+    auto process_group = adopt(*new ProcessGroup(pgid));
+    {
+        ScopedSpinLock lock(g_process_groups_lock);
+        g_process_groups->prepend(process_group);
+        process_group->ref();
```
awesomekling, line 48: Where is the matching `unref()`?

alimpfard: Hmmm...interesting question...

alimpfard: I don't think that `ref()` is even necessary, any process that has it will hold on to it.

alimpfard: _this `ref()`, it does nothing!_

awesomekling:
cc @bugaevc

Kernel/ProcessGroup.cpp
```diff
@@ -0,0 +1,70 @@
...
+
+ProcessGroup* ProcessGroup::from_pgid(ProcessGroupID pgid)
+{
+    ScopedSpinLock lock(g_process_groups_lock);
+
+    for (auto& group : *g_process_groups) {
+        if (group.pgid() == pgid)
+            return &group;
+    }
+    return {};
```
bugaevc, line 67: Default-constructing a pointer argument... are you 100% sure it'll give you a `nullptr`?

alimpfard: This is not a default initializer, this is a zero initializer.

Kernel/TTY/TTY.cpp
```diff
@@ -310,13 +308,27 @@ int TTY::ioctl(FileDescription&, unsigned request, FlatPtr arg)
...
         SessionID new_sid = process ? process->sid() : Process::get_sid_from_pgid(pgid);
         if (!new_sid || new_sid != current_process.sid())
             return -EPERM;
         if (process && pgid != process->pgid())
             return -EPERM;
-        m_process = process ? process->make_weak_ptr() : WeakPtr<Process>();
+        m_pg = process_group->make_weak_ptr();
+        if (process) {
+            auto parent = Process::from_pid(process->ppid());
+            if (parent)
```
bugaevc, line 325:
```suggestion
            if (auto parent = ...)
```

alimpfard: Never in my life would I have thought _you_ would suggest that syntax!

Kernel/ProcessGroup.cpp
alimpfard: (To be more specific, this is scalar value initializer, which is just a zero initializer,
see https://en.cppreference.com/w/cpp/language/value_initialization effects, point 3.

bugaevc: Can't you just write `nullptr` here? :cry:

alimpfard: Sure, `nullptr` is fine too

Kernel/Process.h
```diff
@@ -159,8 +160,8 @@ class Process
     ProcessID pid() const { return m_pid; }
     SessionID sid() const { return m_sid; }
     bool is_session_leader() const { return m_sid.value() == m_pid.value(); }
-    ProcessGroupID pgid() const { return m_pgid; }
-    bool is_group_leader() const { return m_pgid.value() == m_pid.value(); }
+    ProcessGroupID pgid() const { return m_pg ? m_pg->pgid() : 0; }
```
BenWiederhake, line 163: For a future issue: All processes should have a non-zero process group. But that's an existing bug, and this PR doesn't need to fix it.

PR #3166, opened 2020-08-15
AK: Make some tweaks to AK::Span.

asynts:
This pull request contains a few minor tweaks,

  - renames the `span()` methods on containers to `bytes()` if it is known at compile time that bytes are contained, and

  - adds some utility functions like `Span::copy_to()` or `Span::fill()` that are abstracting for `memcpy()`, `memmove()` and `memset()`.

### DIFF
--- a/AK/Span.h
+++ b/AK/Span.h
@@ -148,6 +148,12 @@ public:
         return { this->m_values + start, size };
     }
 
+    ALWAYS_INLINE Span slice(size_t start) const
+    {
+        ASSERT(start < this->m_size);
+        return { this->m_values + start, this->m_size - start };
+    }
+
     ALWAYS_INLINE T* offset(size_t start) const
     {
         ASSERT(start < this->m_size);
--- a/AK/Span.h
+++ b/AK/Span.h
@@ -105,8 +105,6 @@ public:
     using Iterator = T*;
     using ConstIterator = const T*;
 
-    static_assert(!IsPointer<T>::value);
-
     using Detail::Span<T>::Span;
 
     ALWAYS_INLINE Span(std::nullptr_t)
--- a/AK/Span.h
+++ b/AK/Span.h
@@ -176,6 +176,13 @@ public:
         __builtin_memmove(other.data(), data(), sizeof(T) * min(size(), other.size()));
     }
 
+    ALWAYS_INLINE void fill(const T& value)
+    {
+        for (size_t idx = 0; idx < size(); ++idx) {
+            data()[idx] = value;
+        }
+    }
+
     ALWAYS_INLINE const T& at(size_t index) const
     {
         ASSERT(index < this->m_size);
--- a/AK/Span.h
+++ b/AK/Span.h
@@ -154,6 +154,28 @@ public:
         return this->m_values + start;
     }
 
+    ALWAYS_INLINE void copy_to(Span other) const
+    {
+        ASSERT(other.size() >= size());
+        __builtin_memcpy(other.data(), data(), sizeof(T) * size());
+    }
+
+    ALWAYS_INLINE void copy_trimmed_to(Span other) const
+    {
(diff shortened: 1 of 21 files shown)


AK/Span.h
```diff
@@ -150,12 +148,47 @@ class Span : public Detail::Span<T> {
...
+        ASSERT(other.size() >= size());
+        __builtin_memcpy(other.data(), data(), sizeof(T) * size());
+    }
+
+    ALWAYS_INLINE void copy_trimmed_to(Span other) const
+    {
+        __builtin_memcpy(other.data(), data(), sizeof(T) * min(size(), other.size()));
+    }
+
+    ALWAYS_INLINE void move_to(Span other) const
```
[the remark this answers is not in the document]
asynts: That's a really good point, the default should be to assume overlapping spans. But `copy_to_nonoverlapping` is quite a mouthful and the common case is that the memory regions are not overlapping, I mean how many times do you use `memmove`?

Also `copy_trimmed_to` would become something like `copy_trimmed_nonoverlapping_to` that's really to much.

Maybe we can add an enum?

~~~c++
enum class Overlapping { Yes, No };

void copy_to(Span, Overlapping = Overlapping::Yes);
~~~

I can setup a pull request for that.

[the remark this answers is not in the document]
nico: That's a nice API, but I'd personally wait with adding it until there's a clear need. I'd just always call memmove() and only add the explicit nonoverlapping bit once there's a proven need. But if you feel having both is important, I think your proposal is better than what's in the tree :)

Libraries/LibCrypto/Cipher/Cipher.h
```diff
@@ -64,8 +64,8 @@ struct CipherBlock {
     virtual ByteBuffer get() const = 0;
     virtual const ByteBuffer& data() const = 0;
 
-    virtual void overwrite(const ReadonlyBytes&) = 0;
-    virtual void overwrite(const ByteBuffer& buffer) { overwrite(buffer.bytes()); }
```
nico: clang points out:

```
../Libraries/LibCrypto/Cipher/Cipher.h:68:54: error: all paths through this function will call itself [-Werror,-Winfinite-recursion]
    virtual void overwrite(const ByteBuffer& buffer) { overwrite(buffer); }
                                                     ^
```

Libraries/LibCrypto/Cipher/AES.h
```diff
@@ -54,8 +54,8 @@ struct AESCipherBlock : public CipherBlock {
     virtual ByteBuffer get() const override { return m_data; };
     virtual const ByteBuffer& data() const override { return m_data; };
 
-    virtual void overwrite(const ReadonlyBytes&) override;
-    virtual void overwrite(const ByteBuffer& buffer) override { overwrite(buffer.bytes()); }
+    virtual void overwrite(ReadonlyBytes) override;
+    virtual void overwrite(const ByteBuffer& buffer) override { overwrite(buffer); }
```
nico, line 58: clang points out:

```
../Libraries/LibCrypto/Cipher/AES.h:58:63: error: all paths through this function will call itself [-Werror,-Winfinite-recursion]
    virtual void overwrite(const ByteBuffer& buffer) override { overwrite(buffer); }
                                                              ^
```

AK/Span.h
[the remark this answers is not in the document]
asynts: > That's a nice API, but I'd personally wait with adding it until there's a clear need.

I agree. #3179

Libraries/LibCrypto/Cipher/Cipher.h
asynts: Oops. #3179.

Libraries/LibCrypto/Cipher/AES.h
asynts: Oops. #3179.

PR #3167, opened 2020-08-15
AK: Workaround for Apple Clang not supporting concepts

Lubrsi:
Apple Clang (not Clang specifically) does not currently support
concepts.

Reference: https://en.cppreference.com/w/cpp/compiler_support

This was tested by using #ifndef __GNUC__ as I don't have a Mac to
try this on.

### DIFF
--- a/AK/Concepts.h
+++ b/AK/Concepts.h
@@ -30,10 +30,14 @@
 
 namespace AK::Concepts {
 
+// HACK: Apple Clang does not currently support concepts.
+//       This is only an issue when building Lagom.
+#ifndef __APPLE__
 template<typename T>
 concept Integral = IsIntegral<T>::value;
 
 template<typename T>
 concept FloatingPoint = IsFloatingPoint<T>::value;
+#endif
 
 }
--- a/AK/Stream.h
+++ b/AK/Stream.h
@@ -62,6 +62,9 @@ public:
     virtual bool discard_or_error(size_t count) = 0;
 };
 
+// HACK: Apple Clang does not currently support concepts, so let's just use a template with no concepts.
+//       This is only an issue when building Lagom.
+#ifndef __APPLE__
 template<Concepts::Integral Integral>
 InputStream& operator>>(InputStream& stream, Integral& value)
 {
@@ -72,6 +75,14 @@ InputStream& operator>>(InputStream& stream, Integral& value)
 #ifndef KERNEL
 template<Concepts::FloatingPoint FloatingPoint>
 InputStream& operator>>(InputStream& stream, FloatingPoint& value)
+{
+    stream.read_or_error({ &value, sizeof(value) });
+    return stream;
+}
+#endif
+#else
+template<typename T>
+InputStream& operator>>(InputStream& stream, T& value)
 {
     stream.read_or_error({ &value, sizeof(value) });
     return stream;


nico:
Does it work with Xcode 12 beta? If so, maybe we can require that instead?

Lubrsi:
I'm not quite sure, someone with a Mac will have to test it as I don't have one.
This was in response to @nooga having issues with build it with Apple Clang 11:
```
<nooga> yeah and my clang seems not to like concepts, even with -std=c++20
<nooga> what a joke
<lukew> what clang version do you have? apparently clang 10 supports concepts https://en.cppreference.com/w/cpp/compiler_support
<nooga> Apple clang version 11.0.3 (clang-1103.0.32.62)
* nooga shrugs
<lukew> oh wait
<nooga> I'm getting
<lukew> clang and apple clang are separate
<nooga> /Users/nooga/lab/serenity/./AK/Concepts.h:34:1: error: expected unqualified-id
<nooga> concept Integral = IsIntegral<T>::value;
<lukew> clang supports concepts. apple clang doesn't
```
The release notes don't say anything about concept support either: https://developer.apple.com/documentation/xcode-release-notes/xcode-12-beta-release-notes
Also I'm not totally sure about requiring Xcode 12, since x86 is deprecated for it.

nico:
> Also I'm not totally sure about requiring Xcode 12, since x86 is deprecated for it.

Is that a problem? The host build is probably done 64-bit build, no? ...yes:

```
~/src/serenity$ file Build/Meta/Lagom/test-js 
Build/Meta/Lagom/test-js: ELF 64-bit LSB shared object, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, BuildID[sha1]=7b69f95ab18a3c0e67fadb6927ca14e328e33eb3, for GNU/Linux 3.2.0, with debug_info, not stripped
```

Only serenity binaries themselves are 32-bit, and those are built with the serenity gcc toolchain.

Lubrsi:
It's only an issue for building Lagom, but I suppose if x86 does get removed from Xcode, you could use GCC instead.

nico:
What I'm trying to say is that lagom is used for building binaries for the host, and those are built 64-bit. We don't need the 32-bit output of clang for building lagom.

Lubrsi:
Ohhh right. Sadly not, I meant both x86 and x86_64 are deprecated. From the release notes:
```The stand-alone system assemblers for i386, x86_64, and arm are deprecated and may be removed in a future Xcode release.```

**Comment by nooga** [comment]
I've been able to build Lagom (and run resulting binaries) successfully many times before on my system. The current problem is with Apple Clang not supporting C++ concepts which I think we started using recently. 

After `cmake ..` in clean `Build/`:
```
➜  Build git:(master) ✗ make install
Scanning dependencies of target LagomCore
[  0%] Building CXX object Meta/Lagom/CMakeFiles/LagomCore.dir/__/__/AK/Base64.cpp.o
In file included from /Users/nooga/lab/serenity/AK/Base64.cpp:29:
In file included from /Users/nooga/lab/serenity/./AK/String.h:31:
In file included from /Users/nooga/lab/serenity/./AK/Stream.h:29:
/Users/nooga/lab/serenity/./AK/Concepts.h:34:1: error: expected unqualified-id
concept Integral = IsIntegral<T>::value;
^
/Users/nooga/lab/serenity/./AK/Concepts.h:37:1: error: expected unqualified-id
concept FloatingPoint = IsFloatingPoint<T>::value;
^
In file included from /Users/nooga/lab/serenity/AK/Base64.cpp:29:
In file included from /Users/nooga/lab/serenity/./AK/String.h:31:
/Users/nooga/lab/serenity/./AK/Stream.h:65:20: error: no type named 'Integral' in namespace 'AK::Concepts'
template<Concepts::Integral Integral>
         ~~~~~~~~~~^
/Users/nooga/lab/serenity/./AK/Stream.h:66:46: error: unknown type name 'Integral'
InputStream& operator>>(InputStream& stream, Integral& value)
                                             ^
/Users/nooga/lab/serenity/./AK/Stream.h:73:20: error: no type named 'FloatingPoint' in namespace 'AK::Concepts'
template<Concepts::FloatingPoint FloatingPoint>
         ~~~~~~~~~~^
/Users/nooga/lab/serenity/./AK/Stream.h:74:46: error: unknown type name 'FloatingPoint'
InputStream& operator>>(InputStream& stream, FloatingPoint& value)
                                             ^
6 errors generated.
make[2]: *** [Meta/Lagom/CMakeFiles/LagomCore.dir/__/__/AK/Base64.cpp.o] Error 1
make[1]: *** [Meta/Lagom/CMakeFiles/LagomCore.dir/all] Error 2
make: *** [all] Error 2
```
I'm not sure if there are any other problems since the compiler does not get past that. Lagom requires some porting now and then.

My main problem is that I'm trying to build the system for qemu and not Lagom for mac but cmake still puts it in the build for some reason.

**Comment by Lubrsi** [comment]
> My main problem is that I'm trying to build the system for qemu and not Lagom for mac but cmake still puts it in the build for some reason.

It's required for all code generators, since they have to be run on the host system. From the root CMakeLists:
```
add_subdirectory(Meta/Lagom)
add_subdirectory(DevTools/IPCCompiler)
add_subdirectory(DevTools/FormCompiler)
add_subdirectory(Libraries/LibWeb/CodeGenerators)
add_subdirectory(AK/Tests)
```

nico:
> Ohhh right. Sadly not, I meant both x86 and x86_64 are deprecated. From the release notes:
> `The stand-alone system assemblers for i386, x86_64, and arm are deprecated and may be removed in a future Xcode release.`

That refers to clang calling an external assembler instead of using the (default) `-fintegrated-as`. The extern assemblers aren't used by default, and aren't needed. This doesn't affect serenity's build.

nico:
Oh, I think https://github.com/SerenityOS/serenity/blob/master/Documentation/BuildInstructions.md#macos-prerequisites mentions this btw -- see the last bullet.

Lubrsi:
> That refers to clang calling an external assembler instead of using the (default) -fintegrated-as. The extern assemblers aren't used by default, and aren't needed. This doesn't affect serenity's build.

Ah, ok. Yeah, there should be no problem requiring XCode 12+ then.

> Oh, I think https://github.com/SerenityOS/serenity/blob/master/Documentation/BuildInstructions.md#macos-prerequisites mentions this btw -- see the last bullet.

Hm, didn't realise that. Should we close this then?

Lubrsi:
Yeah, I'm gonna close this. Thanks for informing me more on the Mac side of things @nico :^)

PR #3168, opened 2020-08-15
LibCrypto: Fix random number generation and primality tests

BenWiederhake:
While reading LibCrypto, I noticed that the primality test actually executes only 0 rounds, which is a bit too little. Some other things weren't quite right, like the random values being outside the supplied `min` and `max`.

Goofy fun fact of the day: We used to accept `25!` (twenty-five factorial) as a *prime* number. :D

CC @alimpfard, as he wrote the original code.

Also, `Power` warns about being slow. It is currently being only used by `JS::exp`, and is probably being exposed to remote code. This sounds like a bad idea, maybe someone should fix that?

(In case anyone is wondering: "ModularPower" is correct, but I did have some suspicions while writing the primality test, hence the many new test cases.)


Userland/test-crypto.cpp
```diff
@@ -1480,17 +1480,114 @@ static void bigint_test_number_theory()
...
+            } else {
+                FAIL(Wrong result);
+                printf("b: %s\ne: %s\nm: %s\nexpect: %s\nactual: %s\n",
+                    test_case.base_b10, test_case.exp_b10, test_case.mod_b10, test_case.expected_b10, actual.to_base10().characters());
+            }
+        }
+    }
+    {
+        struct {
+            const char* base10;
```
alimpfard, line 1518: Just as a little usability thing, the `_bigint` UDL is already defined for these cases.
It's not constexpr or anything, but it's a bit nicer to read than `blahblah::from_base10(...)`

BenWiederhake: Awesome! I didn't notice that.

Libraries/LibCrypto/NumberTheory/ModularFunctions.h
```diff
@@ -329,15 +333,34 @@ static UnsignedBigInteger random_number(const UnsignedBigInteger& min, const Uns
...
+        // Is it the multiple of a very small prime?
+        if (p_value % 2 == 0 || p_value % 3 == 0 || p_value % 5 == 0 || p_value % 7 == 0)
+            return false;
+        // Then it must be a prime, but not a very small prime, like 37.
         return true;
+    }
 
     Vector<UnsignedBigInteger, 256> tests;
-    UnsignedBigInteger seven { 7 };
-    for (size_t i = 0; i < tests.size(); ++i)
```
alimpfard, line 334: Schrödinger's primes - prime until observed!

BenWiederhake: Yup, that's actually the thing that made me suspicious in the first place. I thought "Wait, either `Vector` has a horrible API or … oh no. Could it be that it doesn't actually run any tests?!" ^^

BenWiederhake:
This has revealed some RSA issues. Do not pull yet.

PR #3169, opened 2020-08-16
AK: Add initial stubs for C++ name demangling

bgianfo:
Many subsystems in Serenity could benefit from functional C++
name demangling. Debug logger stack traces, the profile viewer
the serenity debugger, etc.

This starts framing out that functionality, so it can be filled
in over time.


bgianfo:
@awesomekling told me on IRC I just have an outdated toolchain, so this isn't needed afterall :+1: (thank god)

PR #3170, opened 2020-08-16
Kernel: Emit ProcFS pledge/veil properties as null for kernel processes

bgianfo:
Pledges and Veil state don't really make sense for kernel mode
processes, as they can do what ever they want since they are in
kernel mode. Make this clear in the system monitor UI by keeping
them empty.

### DIFF
--- a/Kernel/FileSystem/ProcFS.cpp
+++ b/Kernel/FileSystem/ProcFS.cpp
@@ -824,26 +824,32 @@ static Optional<KBuffer> procfs$all(InodeIdentifier)
     auto build_process = [&](const Process& process) {
         auto process_object = array.add_object();
 
-        StringBuilder pledge_builder;
+        if (process.is_ring3()) {
+            StringBuilder pledge_builder;
+
 #define __ENUMERATE_PLEDGE_PROMISE(promise)      \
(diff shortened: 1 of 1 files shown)


bugaevc:
A string `"N/A"` is a weird value to use to describe the absence of a value when you have all of JSON types. How about `null`?

SystemMonitor could then display that null as N/A, if desired.

bgianfo:
> A string `"N/A"` is a weird value to use to describe the absence of a value when you have all of JSON types. How about `null`?
Sure, is String() the correct way to emit a null in the JSON subsytem? I couldn't find any explicit API... this looked liked the closest thing.

bugaevc:
> Sure, is String() the correct way to emit a null in the JSON subsytem?

That works too:

https://github.com/SerenityOS/serenity/blob/b0aa8115c2c1d240edc75bbd5263f7dc2ca3d28b/AK/JsonValue.cpp#L195-L199

but...

> I couldn't find any explicit API...

The default value of `JsonValue` is null:

https://github.com/SerenityOS/serenity/blob/b0aa8115c2c1d240edc75bbd5263f7dc2ca3d28b/AK/JsonValue.h#L56

So you should be able to simply use `{}` whenever you need a null `JsonValue`.

PR #3174, opened 2020-08-16
LibGUI: Don't center dialog within parent if window is not visible

### LINKED ISSUE #3172: WindowServer: Error message box not visible on TextEditor startup
To reproduce:
`te /tmp/portal/window`

TextEditor shows an error message box, but the window is nowhere to be found.

cc @tomuta

linusg:
Windows have an initial off-screen rect, so when a dialog is created and shown before its parent window, in which it is centered, it would be invisible to the user.

Fixes #3172.


awesomekling:
Thank you for fixing that :)

PR #3175, opened 2020-08-16
TextEditor: Ruler fixes

thankyouverycool:
Fixes ruler width and boldface glyph issues. Width is now based on the number of digits in the line count plus a 1 digit buffer, similar to gedit, and the ruler tries to match a bold family font when drawing the current line instead of defaulting to Katica.
ruler

### DIFF
--- a/Libraries/LibGUI/TextEditor.cpp
+++ b/Libraries/LibGUI/TextEditor.cpp
@@ -42,6 +42,7 @@
 #include <LibGfx/Palette.h>
 #include <ctype.h>
 #include <fcntl.h>
+#include <math.h>
 #include <stdio.h>
 #include <unistd.h>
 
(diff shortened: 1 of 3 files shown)


Libraries/LibGUI/TextEditor.cpp
```diff
@@ -40,6 +40,7 @@
 #include <LibGfx/Bitmap.h>
 #include <LibGfx/Font.h>
 #include <LibGfx/Palette.h>
+#include <LibM/math.h>
```
awesomekling, line 43: This should be `<math.h>`

Libraries/LibGfx/Font.h
```diff
@@ -130,6 +134,9 @@ class Font : public RefCounted<Font> {
     static RefPtr<Font> load_from_memory(const u8*);
     static size_t glyph_count_by_type(FontTypes type);
 
+    void set_family_fonts();
+    Font* m_bold_family_font { nullptr };
```
awesomekling, line 138: Let's not use a raw pointer here, and let's not leak this.

PR #3181, opened 2020-08-16
LibWeb: Remove unused private field m_queued_finish in HttpsJob

nico:
Doesn't look like this changed recently, so not sure why clang complains about this now when it didn't earlier.


awesomekling:
That's LibHTTP :)

PR #3183, opened 2020-08-17
AK: Implement StringBuilder as a buffered Roped string builder

alimpfard:
Ropes and stuff.
Also allows fairly big perf gains just by adding an inline buffer.

Still a bit WIP because `Rope::to_string()` is pretty allocation heavy - to be fixed a bit later.


bgianfo:
Should there be new tests for AK::Rope?

alimpfard:
> new tests

Absolutely!

alimpfard:
Hmmm
Nope, not a fan of shoving two data structures in one, will make a new PR with them split up.

PR #3184, opened 2020-08-17
AK + Libraries + Applications + Games + DevTools: Coverity static analysis fixes

bgianfo:
This is a collection of fixes to:
- Get Serenity Userspace building with the Coverity static analysis tools, as it doesn't support C++ concepts.
- Fix an initial set of issues found by the one off build + upload I did of the latest Serenity master branch.

There do seem to be some false positives, like double free's being detected around usage of NonnullRefPtr<T>, and other C++ template types. 

Reference: https://scan.coverity.com/projects/serenityos-serenity

### DIFF
--- a/Applications/FileManager/FileUtils.cpp
+++ b/Applications/FileManager/FileUtils.cpp
@@ -26,6 +26,7 @@
 
 #include "FileUtils.h"
 #include <AK/LexicalPath.h>
+#include <AK/ScopeGuard.h>
 #include <AK/StringBuilder.h>
 #include <LibCore/DirIterator.h>
 #include <stdio.h>
@@ -85,6 +86,8 @@ bool copy_file_or_directory(const String& src_path, const String& dst_path)
     if (src_fd < 0) {
         return false;
     }
+
+    ScopeGuard close_fd_guard([src_fd]() { close(src_fd); });
 
     struct stat src_stat;
     int rc = fstat(src_fd, &src_stat);
--- a/Applications/FileManager/FileUtils.cpp
+++ b/Applications/FileManager/FileUtils.cpp
@@ -144,6 +144,8 @@ bool copy_file(const String& src_path, const String& dst_path, const struct stat
         }
     }
 
+    ScopeGuard close_fd_guard([dst_fd]() { close(dst_fd); });
+
     if (src_stat.st_size > 0) {
         if (ftruncate(dst_fd, src_stat.st_size) < 0) {
             perror("cp: ftruncate");
@@ -180,7 +182,6 @@ bool copy_file(const String& src_path, const String& dst_path, const struct stat
     }
 
     close(src_fd);
-    close(dst_fd);
     return true;
 }
 
--- a/Applications/Piano/RollWidget.h
+++ b/Applications/Piano/RollWidget.h
@@ -46,8 +46,8 @@ private:
 
     TrackManager& m_track_manager;
 
-    int m_roll_width;
-    int m_num_notes;
-    double m_note_width;
+    int m_roll_width { 0 };
+    int m_num_notes { 0 };
+    double m_note_width { 0.0 };
     int m_zoom_level { 1 };
 };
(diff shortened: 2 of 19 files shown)


Applications/FileManager/FileUtils.cpp
```diff
@@ -86,6 +87,8 @@ bool copy_file_or_directory(const String& src_path, const String& dst_path)
         return false;
     }
 
+    ScopeGuard close_fd_guard([src_fd]() { close(src_fd); });
```
awesomekling, line 90: This is awkward because it will also be closed in `copy_file()` (but not in `copy_directory()`)
A double `close()` is not the end of the world here of course.

awesomekling:
Quite spicy :hot_pepper:

Applications/FileManager/FileUtils.cpp
bugaevc: A double close is the end of the world, especially in a multithreaded program like the FileManager.

Applications/Piano/RollWidget.h
```diff
@@ -46,8 +46,8 @@ class RollWidget final : public GUI::ScrollableWidget {
 
     TrackManager& m_track_manager;
 
-    int m_roll_width;
-    int m_num_notes;
-    double m_note_width;
+    int m_roll_width { 0 };
+    int m_num_notes { 0 };
+    double m_note_width { 0.0 };
```
bugaevc, line 51: This is embarrassing, how come we didn't catch it during review?

DevTools/UserspaceEmulator/Emulator.cpp
```diff
@@ -648,9 +648,9 @@ int Emulator::virt$select(FlatPtr params_addr)
     Syscall::SC_select_params params;
     mmu().copy_from_vm(&params, params_addr, sizeof(params));
 
-    fd_set readfds;
-    fd_set writefds;
-    fd_set exceptfds;
+    fd_set readfds {};
+    fd_set writefds {};
+    fd_set exceptfds {};
```
bugaevc, line 653: `fd_set` is a C type and should be initialized with `FD_ZERO()` or something.

And correct me if I'm wrong, but does adding an explicit `{}` even change anything? I believe the default constructor is invoked anyway.

Libraries/LibCore/SocketAddress.h
```diff
@@ -95,7 +95,7 @@ class SocketAddress {
     sockaddr_in to_sockaddr_in() const
     {
         ASSERT(type() == Type::IPv4);
-        sockaddr_in address;
+        sockaddr_in address {};
```
bugaevc, line 98: Same

Userland/more.cpp
```diff
@@ -36,7 +36,7 @@ static void wait_for_key()
     printf("\033[7m--[ more ]--\033[0m");
     fflush(stdout);
     char dummy;
-    read(key_fd, &dummy, 1);
+    (void)read(key_fd, &dummy, 1);
     printf("\n");
```
bugaevc, line 40: That is 😟 though

We shouldn't ignore errors, not EOF.

DevTools/UserspaceEmulator/MallocTracer.cpp
```diff
@@ -138,7 +138,7 @@ void MallocTracer::audit_read(FlatPtr address, size_t size)
         Emulator::the().dump_backtrace();
         report("==%d==  Address is %zu byte(s) into block of size %zu, allocated at:\n", getpid(), offset_into_mallocation, mallocation->size);
         Emulator::the().dump_backtrace(mallocation->malloc_backtrace);
-        report("==%d==  Later freed at:\n", getpid(), offset_into_mallocation, mallocation->size);
+        report("==%d==  Later freed at:\n", getpid());
```
bugaevc, line 141: Clearly, `report()` needs `attribute((format))`

DevTools/UserspaceEmulator/Emulator.cpp
awesomekling: `{}` is aggregate initialization, and with no default constructor present, it will zero-initialize primitive members.
(At least that's how I understand it. :)

Applications/Piano/RollWidget.h
awesomekling: These are basically harmless since they get assigned on first paint event anyway.
But still, should have been caught!

Applications/FileManager/FileUtils.cpp
awesomekling: I'm having some serious race condition blindness lately. Thanks for setting me straight :)

PR #3189, opened 2020-08-17
LibWeb: Change cursor to IBeam when hovering text

AnicJov:
This is what most browsers do, hopefully it isn't too silly :^)


awesomekling:
Not silly at all. Good idea :)

PR #3192, opened 2020-08-17
AK:IDAllocator seed random generator to prevent sequence regeneration

tryfinally:
1. Randomize ID between runs. Otherwise, generated IDs can be guessed.
2. Remove check for generated ID == 0 from inside the loop

### DIFF
--- a/AK/IDAllocator.h
+++ b/AK/IDAllocator.h
@@ -26,25 +26,24 @@
 
 #pragma once
 
-#include <stdlib.h>
 #include <AK/HashTable.h>
+#include <stdlib.h>
+#include <time.h>
 
 namespace AK {
 
 class IDAllocator {
 
 public:
-    IDAllocator() {}
-    ~IDAllocator() {}
+    IDAllocator() { srand(time(0)); }
+    ~IDAllocator() { }
 
     int allocate()
     {
         int r = rand();
-        for (int i = 0; i < 100000; ++i) {
+        // Make sure we never vend ID 0, as some code may interpret that as "no ID"
+        for (int i = 1; i < 100000; ++i) {
             int allocated_id = r + i;
-            // Make sure we never vend ID 0, as some code may interpret that as "no ID"
-            if (allocated_id == 0)
-                ++allocated_id;
             if (!m_allocated_ids.contains(allocated_id)) {
                 m_allocated_ids.set(allocated_id);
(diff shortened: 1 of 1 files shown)


awesomekling:
What bug is this fixing? Please explain why you are doing things, that's what commit messages (and PR descriptions) are for :)

bgianfo:
Is there a reason why the IdAllocator is a HashSet? It seems like a bitmap would be more fitting, faster, and smaller size.

AK/IDAllocator.h
```diff
@@ -26,25 +26,24 @@
...
     int allocate()
     {
         int r = rand();
-        for (int i = 0; i < 100000; ++i) {
+        // Make sure we never vend ID 0, as some code may interpret that as "no ID"
+        for (int i = 1; i < 100000; ++i) {
             int allocated_id = r + i;
-            // Make sure we never vend ID 0, as some code may interpret that as "no ID"
-            if (allocated_id == 0)
-                ++allocated_id;
```
BenWiederhake, line 47: This part doesn't make sense to me. What if `r` happens to be `4294967296-i` or slightly lower? Then `r+i` will be `0`. In fact, `i=0` is not a special case at all.

It should take at least 65536 allocated IDs until this loop actually running more than once or twice (see: birthday paradox, and `int` is 32 bit`). Let's not over-optimize this too early. Or do you profile something and IDAllocator was the culprit?

BenWiederhake (changes requested):

BenWiederhake:
> It seems like a bitmap would be more fitting, it's faster, and will consume less memory.

Are you sure about that? The IDs are effectively randomly picked `int`s, if we can trust `rand()`. So a bitset shouldn't be able to help.

AK/IDAllocator.h
```diff
@@ -26,25 +26,24 @@
...
+#include <time.h>
 
 namespace AK {
 
 class IDAllocator {
 
 public:
-    IDAllocator() {}
-    ~IDAllocator() {}
+    IDAllocator() { srand(time(0)); }
```
BenWiederhake, line 38: Calling `srand` means we change global state. What if a program wants to be reproducible and intentionally called `srand(MAGIC_BUT_CONSTANT_NUMBER)` initially? Do we care about such "weird" setups?

PR #3193, opened 2020-08-17
LibLine: Allow the user to override (or add) keybinds in the config file

alimpfard:
This PR adds configurable keybindings.
My current favourite keybinding is `e=internal:finish_edit`, it makes almost everything impossible to do lol

### DIFF
--- a/Libraries/LibCore/ConfigFile.cpp
+++ b/Libraries/LibCore/ConfigFile.cpp
@@ -34,6 +34,14 @@
 
 namespace Core {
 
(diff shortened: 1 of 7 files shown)


awesomekling:
Sure sure, as long as we have nice defaults :)

alimpfard:
They're the same defaults as before~
I can't stop people from `" "=internal:transpose_words` though

awesomekling:
I think this broke `^W` and `^U`

alimpfard:
> I think this broke ^W and ^U

Fixed by #3203~

PR #3194, opened 2020-08-17
Implement various JavaScript and DOM things.

nico:
Adds:

- A basic, incomplete implementation of `Array.from()`
- A basic, incomplete implementation of `Date.toLocaleString()`, `Date.toLocaleDateString(), `Date.toLocaleTimeString()`
- A mostly-working implementation of `Node.textContent`

Also slightly simplifies the Clock MenuApplet.


nico:
The motivation was to get the timezone localization on http://45.33.8.238/ working, but I ran out of time to implement `innerText` so it doesn't quite work yet. I settled for the simpler `textContent` instead. Maybe I'll change the web page to use that instead of `innerText`, or maybe I'll come back and implement `innerText` later.

awesomekling:
Ah yes, the famous JavaScript DOM thing: clock menu applet :D

nico:
Well I was looking at the clock menuapplet to find out what the `toLocale*String()` methods should do. It all makes sense in my head!

alimpfard:
> the famous JavaScript DOM thing: clock menu applet

It'll eventually happen when we get libjs bindings~

PR #3195, opened 2020-08-17
AK:HashMap+LibJs:Lexer add ctor with capacity + use it in Js Lexer

tryfinally:
1. Add ctor to Hashmap to allow ensuring capacity.
2. Refactor LibJS:Lexer to use it.

### DIFF
--- a/AK/HashMap.h
+++ b/AK/HashMap.h
@@ -48,6 +48,10 @@ private:
 
 public:
     HashMap() {}
+    HashMap(size_t capacity)
+        : m_table(capacity)
+    {
+    }
 
     bool is_empty() const { return m_table.is_empty(); }
     size_t size() const { return m_table.size(); }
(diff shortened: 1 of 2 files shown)


nico:
A question I forgot to ask on the previous similar change: Why is the number of elements passed to the hash map the best size for the hash map? It virtually guarantees that there are buckets with more than 1 element. Maybe there's some reason why this is the best size/collision tradeoff, but it's at least not obvious to me.

If the "hash of fixed list of strings" use case is important, maybe it would make sense to add a perfect hash function generator that takes a list of fixed strings and generates a perfect hash function, then we could do a n-elt hash table without collisions (and with a fast, special-purpose hash function).

tryfinally:
I think that a good idea. Especially on those lexer maps.
For now maybe just use a load factor?

alimpfard:
Or maybe just add a constexpr hashmap and throw all this out?

_/me pushes for more constexpr_

awesomekling:
Also please put the HashMap change in a separate commit

bgianfo:
> Or maybe just add a constexpr hashmap and throw all this out?

@alimpfard Are you talking  about this? https://www.youtube.com/watch?v=INn3xa4pMfg
The token mapping does seem like a pretty nice use case for for something like what's mentioned in the video.

PR #3200, opened 2020-08-18
LibWeb: Implement Element.innerText

nico:
Reading the property has a few warts (see FIXMEs in the included
tests), but with this the timestamps on http://45.33.8.238/
get localized :^)

Since the Date() constructor currently ignores all arguments,
they don't get localized correctly but are all set to the current
time, but hey, it's still progress from a certain point of view.

Also adds another test for textContent, which behaves correctly for the most part.

### DIFF
--- a/Libraries/LibWeb/DOM/Element.cpp
+++ b/Libraries/LibWeb/DOM/Element.cpp
@@ -41,6 +41,7 @@
 #include <LibWeb/Layout/LayoutTableCell.h>
 #include <LibWeb/Layout/LayoutTableRow.h>
 #include <LibWeb/Layout/LayoutTableRowGroup.h>
+#include <LibWeb/Layout/LayoutText.h>
 #include <LibWeb/Layout/LayoutTreeBuilder.h>
 
 namespace Web::DOM {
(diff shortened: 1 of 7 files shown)


Libraries/LibWeb/DOM/Element.cpp
```diff
@@ -295,6 +296,39 @@ String Element::inner_html() const
...
+
+    // innerText for element being rendered takes visibility into account, so force a layout and then walk the layout tree.
+    document().layout();
+    if (!layout_node())
+        return text_content();
+
+    Function<void(const LayoutNode&)> recurse = [&](auto& node) {
+        for (auto* child = node.first_child(); child; child = child->next_sibling()) {
+            if (child->is_text())
+                builder.append(downcast<LayoutText>(*child).text_for_rendering());
```
awesomekling, line 321: `text_for_rendering()` has collapsed whitespace characters, is that really what we want here?
If not, we should get `downcast<LayoutText>(*child).node().data()` instead (the DOM text.)

nico: Yes, this should do whitespace collapsing. I tweaked the test to make this explicit.

http://perfectionkills.com/the-poor-misunderstood-innerText/ (which I found just now) has some details if you're curious. Sounds like the property's been standardized since: https://html.spec.whatwg.org/multipage/dom.html#the-innertext-idl-attribute The algorithm there is a lot more detailed than this first draft.

The blog post makes the interesting point that Selection.toString() (which what select->copy implicitly uses, I guess?) should have the same behavior as reading innerText.

PR #3201, opened 2020-08-18
Kernel: Use Userspace<T> for the sendto + recvfrom syscalls

bgianfo:
More progress on the Userspace<T> all the things long term goal.

### DIFF
--- a/Kernel/Syscalls/socket.cpp
+++ b/Kernel/Syscalls/socket.cpp
@@ -188,7 +188,7 @@ int Process::sys$shutdown(int sockfd, int how)
     return socket.shutdown(how);
 }
 
-ssize_t Process::sys$sendto(const Syscall::SC_sendto_params* user_params)
+ssize_t Process::sys$sendto(Userspace<const Syscall::SC_sendto_params*> user_params)
 {
     REQUIRE_PROMISE(stdio);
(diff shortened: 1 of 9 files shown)


Kernel/Syscalls/socket.cpp
```diff
@@ -227,18 +227,21 @@ ssize_t Process::sys$recvfrom(const Syscall::SC_recvfrom_params* user_params)
         return -EFAULT;
 
     int flags = params.flags;
-    sockaddr* addr = params.addr;
-    socklen_t* addr_length = params.addr_length;
+    Userspace<sockaddr*> user_addr = params.addr;
+    Userspace<socklen_t*> user_addr_length = params.addr_length;
 
     SmapDisabler disabler;
```
awesomekling, line 233: Can we remove this now?

bgianfo: I didn't look if the write to the data buffer is protected or not, that was the only thing that stopped me from removing it.

bgianfo:
> What path do I have to take for the kernel write at arbitrary address? I don't see it.
> The `addr_length` out-param is validated in `sys$recvfrom`, no?

You are right, I must have confused my self while sitting in the socket implementations.

PR #3204, opened 2020-08-18
Awesome 2048

bugaevc:
* Visuals (metrics, colors) to match the original
* Major refactoring
* Various other fixes

Before:

After:

cc @alimpfard (read [this](https://github.com/SerenityOS/serenity/commit/663397c04e1e10d8f4eeb068a45d37a8fd2d83cb))


bugaevc:
A regression of sorts: since colors no longer depend on the theme, it now looks too bright on dark themes

image

alimpfard:
> since colors no longer depend on the theme, it now looks too bright on dark themes

Why not just use the trick I pulled and mix in 50% of the theme base colour?

bugaevc:
> Why not just use the trick I pulled and mix in 50% of the theme base colour?

But then the colors would be different (to *the* 2048 color scheme).

There exist tricks to make 2048 look good in dark environments while still having it "look like 2048" (I would know — I've previously spent some time trying to make my WearOS port of 2048 look good in the dark environment). Basically, you make the empty cells & the background darker, and switch all the text to dark. But I've left all that for another time here.

alimpfard:
Hm.

template<typename T>
I don't see the value in copying the original to the T, but whatever floats your boat.

PR #3207, opened 2020-08-18
LibJS: Parser refactored to use constexpr precedence table

tryfinally:
Replaced the HashMap based implementation of operator precedence with a constexpr OperatorPrecedenceTable class using 
constexpr array lookup instead.

### DIFF
--- a/Libraries/LibJS/Parser.cpp
+++ b/Libraries/LibJS/Parser.cpp
@@ -26,7 +26,6 @@
  */
 
 #include "Parser.h"
-#include <AK/HashMap.h>
 #include <AK/ScopeGuard.h>
 #include <AK/StdLibExtras.h>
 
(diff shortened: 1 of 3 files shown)


awesomekling:
Why is it relevant that this is `constexpr`? The lookups are made at runtime anyway.
This is just replacing a hashmap with an array, no?

cc @sunverwerth

tryfinally:
The table is initialized at compile time.

sunverwerth:
On one hand this could provide a small performance boost. On the other hand, turning this into a switch/case would make it a lot clearer and probably also optimized.

tryfinally:
How a switch case is better than a straight lookup table?

sunverwerth:
Compilers employ different optimization strategies for switch statements. Jump tables, binary search, etc..

sunverwerth:
I'm just saying, if this can provide a considerable performance boost, go for it. Otherwise I'd rather go with the 'as simple as possible' solution.

tryfinally:
The lookup table is one direct memory read without need to jump tables or binary search.
The operator token is used to index an array of int. It can't get simpler than that.

alimpfard:
@tryfinally I'd lose the storage thing, just do a linear search on the precedence table and the compiler _will_ pick up what you're doing.
Though make sure to keep them in the right order, gcc can't deal with it when out-of-order.

> https://gcc.godbolt.org/z/K3Thsx

tryfinally:
@alimpfard  see [https://godbolt.org/z/x4f6ha](https://godbolt.org/z/x4f6ha)
gcc/clang will optimize all of this code to a simple lookup using one index memory access.
*It can't get faster than that!*

alimpfard:
My comment wasn't really about one or the other being faster, it was just about less code

But, it works if it works :)

tryfinally:
The amazing C++, all that code collapse into a couple of machine instructions :-)

PR #3208, opened 2020-08-18
Userland: Add an implementation of 'printf'

alimpfard:
Since this is a pretty template-heavy change, I'll hold off on writing a manpage and stuff until the `PrintfImplementation` gets a pass~

The only other way I could think of was to essentially rewrite `printf(3)` but with `printf(1)`-specific semantics...I prefer the template magic, personally.

cc @bugaevc (you seem to be back, nitpick away), @awesomekling.

### DIFF
--- a/AK/PrintfImplementation.h
+++ b/AK/PrintfImplementation.h
@@ -32,6 +32,8 @@
 #include <AK/Types.h>
 #include <stdarg.h>
 
+namespace PrintfImplementation {
+
 static constexpr const char* printf_hex_digits_lower = "0123456789abcdef";
 static constexpr const char* printf_hex_digits_upper = "0123456789ABCDEF";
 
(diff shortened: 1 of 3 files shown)


AK/PrintfImplementation.h
```diff
@@ -283,166 +285,227 @@ ALWAYS_INLINE int print_signed_number(PutChFunc putch, char*& bufptr, int number
     return print_number(putch, bufptr, number, left_pad, zero_pad, field_width);
 }
 
-template<typename PutChFunc>
-ALWAYS_INLINE int printf_internal(PutChFunc putch, char* buffer, const char*& fmt, va_list ap)
+struct State {
```
bugaevc, line 288: It's not really State, it's more like "modifiers" or "display options". But of course those display options are accumulated as we scan through the format string, so there's a point in calling it State too.

Perhaps you can think of a name that represents both?

AK/PrintfImplementation.h
```diff
@@ -283,166 +285,227 @@ ALWAYS_INLINE int print_signed_number(PutChFunc putch, char*& bufptr, int number
...
+    bool left_pad = false;
```
bugaevc, line 289: This syntax:

```suggestion
    bool left_pad { false };
```

for member initialization

AK/PrintfImplementation.h
```diff
@@ -283,166 +285,227 @@ ALWAYS_INLINE int print_signed_number(PutChFunc putch, char*& bufptr, int number
...
+    unsigned fraction_length = 6;
+    unsigned long_qualifiers = 0;
+    bool size_qualifier = false;
+    bool alternate_form = 0;
+    bool always_sign = false;
+};
+
+template<typename PutChFunc, typename ArgumentListT, template<typename T, typename U = ArgumentListT> typename NextArgument>
+struct PrintfImpl {
+    ALWAYS_INLINE int format_s(PutChFunc putch, char*& bufptr, int, const State& state, ArgumentListT ap) const
```
bugaevc, line 303: Argument list is passed by value here and by reference below. Why?

AK/PrintfImplementation.h
```diff
@@ -283,166 +285,227 @@ ALWAYS_INLINE int print_signed_number(PutChFunc putch, char*& bufptr, int number
...
+    }
+    ALWAYS_INLINE int format_f(PutChFunc putch, char*& bufptr, int, const State& state, ArgumentListT& ap) const
+    {
+        return print_double(putch, bufptr, NextArgument<double>()(ap), state.left_pad, state.zero_pad, state.field_width, state.fraction_length);
+    }
+    ALWAYS_INLINE int format_o(PutChFunc putch, char*& bufptr, int, const State& state, ArgumentListT& ap) const
+    {
+        if (state.alternate_form) {
+            putch(bufptr, '0');
+            return 1;
```
bugaevc, line 347: ???

AK/PrintfImplementation.h
```diff
@@ -283,166 +285,227 @@ ALWAYS_INLINE int print_signed_number(PutChFunc putch, char*& bufptr, int number
...
+    bool has_fraction_length = false;
+    unsigned fraction_length = 6;
+    unsigned long_qualifiers = 0;
+    bool size_qualifier = false;
+    bool alternate_form = 0;
+    bool always_sign = false;
+};
+
+template<typename PutChFunc, typename ArgumentListT, template<typename T, typename U = ArgumentListT> typename NextArgument>
+struct PrintfImpl {
```
bugaevc, line 302: Perhaps all these arguments you pass to every function (put char callback, buffer pointer, number of written bytes, modifiers state, argument list) could be made into fields in this class?

AK/PrintfImplementation.h
```diff
@@ -283,166 +285,227 @@ ALWAYS_INLINE int print_signed_number(PutChFunc putch, char*& bufptr, int number
...
+    }
+};
+
+#define PRINTF_IMPL_DELEGATE_TO_IMPL(c)                        \
+    case* #c:                                                  \
+        ret += impl.format_##c(putch, bufptr, ret, state, ap); \
+        break;
+
+template<typename PutChFunc, template<typename T, typename U, template<typename X, typename Y> typename V> typename Impl = PrintfImpl, typename ArgumentListT = va_list, template<typename T, typename V = decltype(declval<ArgumentListT&>())> typename NextArgument = VaArgNextArgument>
+ALWAYS_INLINE int printf_internal(PutChFunc putch, char* buffer, const char*& fmt, ArgumentListT ap)
```
bugaevc, line 412: This could, too, be moved to the type as a method

Userland/printf.cpp
```diff
@@ -0,0 +1,297 @@
...
+#include <AK/PrintfImplementation.h>
+#include <AK/String.h>
+#include <AK/StringBuilder.h>
+#include <AK/Types.h>
+#include <stdio.h>
+#include <unistd.h>
+
+[[gnu::noreturn]] void fail(const char* message)
+{
+    puts(message);
```
bugaevc, line 36: Some nice red please

Userland/printf.cpp
```diff
@@ -0,0 +1,297 @@
...
+        return state_copy.field_width;
+    }
+};
+
+template<typename T, typename V>
+struct ArgvNextArgument {
+    ALWAYS_INLINE T operator()(V) const
+    {
+        static_assert(sizeof(V) != sizeof(V), "Base instantiated");
+        return declval<T>();
```
bugaevc, line 106: So *that's* what you need `declval` for?

Can't you just `ASSERT_NOT_REACHED()`?

AK/PrintfImplementation.h
alimpfard: `OptionState`?
`AccumulatedDisplayOptions`?

I dunno, I'm not great at naming things

AK/PrintfImplementation.h
alimpfard: Sure.

AK/PrintfImplementation.h
alimpfard: It's not a reference here, the name is misleading, I suppose.
I'll rename it to `ArgumentListRefT`

AK/PrintfImplementation.h
bugaevc: Actually `ModifierState` sounds pretty good to me.

AK/PrintfImplementation.h
alimpfard: Ask whomever wrote that (@awesomekling, probably), I haven't changed any behaviour.

AK/PrintfImplementation.h
```diff
@@ -283,166 +285,227 @@ ALWAYS_INLINE int print_signed_number(PutChFunc putch, char*& bufptr, int number
...
+    unsigned field_width = 0;
+    bool has_fraction_length = false;
+    unsigned fraction_length = 6;
+    unsigned long_qualifiers = 0;
+    bool size_qualifier = false;
+    bool alternate_form = 0;
+    bool always_sign = false;
+};
+
+template<typename PutChFunc, typename ArgumentListT, template<typename T, typename U = ArgumentListT> typename NextArgument>
```
bugaevc, line 301: I still don't see why that `U` is needed.

Userland/printf.cpp
```diff
@@ -0,0 +1,297 @@
...
+                }
+            } else {
+                builder.append(c);
+            }
+        } else {
+            builder.append(c);
+        }
+    }
+
+end:;
```
bugaevc, line 269: :thinking:

I'd just say `return builder.build()` in that one place above.

AK/PrintfImplementation.h
alimpfard: I'd like not to store the putch function anywhere, storing the argument list is a no-no.
My preference with this series of `Impl` structs is to keep them merely a collection of functions (they're not static because even naming the parent class in a child is a hard thing to do, since the type of `NextArgument` is...not that easy to name.

AK/PrintfImplementation.h
alimpfard: then the rest of the codebase would need to be changed too, `printf_internal` is _the_ interface for printf, no one should be able to override it, otherwise, `PrintfImpl` is useless.

Userland/printf.cpp
alimpfard: No, I need `declval` to be able to resolve `va_list&`.

AK/PrintfImplementation.h
alimpfard: Since `va_list&` cannot be named, it has to be passed as a type parameter, making the template type of `NextArgument` require two params; but the second parameter is just the input argument list type.

in short, this is here to make it possible to pass `va_list` as a reference.

AK/PrintfImplementation.h
bugaevc: My main concern is that all these methods accept lots of arguments, and all those arguments are the same, *and* some of them clearly come from state maintained somewhere else (such as buffer pointer and nwritten). Would be more clear to make fields of this type a canonical location for all that state.

You're not really storing the argument list, you're storing the getter function.

Userland/printf.cpp
alimpfard: Alright.
I honestly didn't expect `%c` to just terminate the format, so I threw in a `goto end` on instinct lol

AK/PrintfImplementation.h
alimpfard: `ModifierState` it is, then.

AK/PrintfImplementation.h
bugaevc: What do you mean it cannot be named?

AK/PrintfImplementation.h
alimpfard: > `buffer pointer`

That's _your own_ state, i.e. the caller's.

> `nwritten`

This can go into the state, that makes sense to me.

> You're not really storing the argument list, you're storing the getter function

You mean `NextArgument`? that needs to be instantiated where needed, I can't store any `NextArgument` anywhere

AK/PrintfImplementation.h
alimpfard: Try making a function take a `va_list&`, and pass a `va_list` into it.
you cannot name it in a normal way, array type and all that.

bugaevc: Oh, that is simple to workaround, just wrap it in a struct.

AK/PrintfImplementation.h
alimpfard: Actually, a `ModifierState` is only valid for the duration of _one_ format call, so `nwritten` cannot go there either.
hmmm

AK/PrintfImplementation.h
bugaevc: Like this:

```cpp
struct { va_list va_list; } args;
va_copy(args.va_list, original_args);
```

alimpfard: Then _everything_ will have to be wrapped in a struct.
or we'll have to do some SFINAE trickery to selectively wrap `va_list`'s
Also, can you even move a `va_list`? how would wrapping it even work?

```c++
struct Foo {
    va_list list;
};

va_list ap;
Foo x { move(ap) }; // This doesn't seem to make much sense to me
```

AK/PrintfImplementation.h
bugaevc: > That's your own state, i.e. the caller's.

I'm basically suggesting to unify the caller (`printf_internal`) and this callee type

AK/PrintfImplementation.h
bugaevc: > Then everything will have to be wrapped in a struct. or we'll have to do some SFINAE trickery to selectively wrap va_list's

No? Just the implementation of `NextArgument` on to of `va_list` needs to care. The rest, and wasn't this the whole point, is oblivious to how exactly you're getting the argument values.

alimpfard: > `va_copy(args.va_list, original_args);`

That will either require the ctor of that struct to take a `va_list` as a reference (defeating the point), or the instantiating site to specialise (probably via `if constexpr`) for `va_list`.
Not a fan of either, but if the template args method is really much worse than either of these, I'll swap it out

alimpfard: > No? Just the implementation of NextArgument on top of va_list needs to care.

I missed that entirely somehow. _that's_ reasonable.

AK/PrintfImplementation.h
alimpfard: but the entire point is to keep `printf_internal` completely separate from this type...?

bugaevc: Yeah...

It's just that whenever I see too many arguments being shoveled in and out of functions, my brain goes "perhaps these could be members in a class, and the functions would be the methods".

Can we get the best of both worlds somehow?

alimpfard: I guess the buffer pointer and the nwritten args can be members of the class, not sure if it's possible to keep a reference to `PutChFunc` (for _any_ PutChFunc), but other two - sure.
`ap` and `state` can't be members by necessity though

alimpfard: Now the `format_*` functions take two args, I doubt it can get any better without storing `ap` somewhere, which is...not something I'm comfortable with.

AK/PrintfImplementation.h
```diff
@@ -283,166 +285,238 @@ ALWAYS_INLINE int print_signed_number(PutChFunc putch, char*& bufptr, int number
...
+    }
+    ALWAYS_INLINE int format_unrecognized(char format_op, const char* fmt, const ModifierState&, ArgumentListRefT&) const
+    {
+        dbg() << "printf_internal: Unimplemented format specifier " << format_op << " (fmt: " << fmt << ")";
+        return 0;
+    }
+
+protected:
+    char*& m_bufptr;
+    const int& m_nwritten;
```
bugaevc, line 406: Now, instead of returning how many characters were written, why don't methods just increase this value directly?

alimpfard: > \*chaos\*

I don't want the methods messing around with `nwritten`, and passing it around as they please...
it feels safer for them to return how many they've written instead of changing that value.

AK/PrintfImplementation.h
awesomekling: That's `%#o`, which gives the "alternate" octal representation (zero-prefixed.)

bugaevc: Sure, but why does it return at this point, instead of falling through to printing the actual value? As written, it just writes out a `'0'` character.

alimpfard: oops.
Appears that is my mistake.

alimpfard: it's `++ret` on master, my bad.

AK/PrintfImplementation.h
alimpfard: I tried doing this (conditionally wrapping the `va_list`), but it just doesn't want to cooperate.
I don't think losing one template template parameter is worth-
- The wrapping cost (if any)
- The number of new structures added to make this conditional wrapping possible

AK/PrintfImplementation.h
alimpfard: I'm gonna go ahead and assume that this is ok now.

AK/PrintfImplementation.h
alimpfard: (repeated below in the 'typename U' bit)

awesomekling:
Seems good enough to iterate on :)

PR #3213, opened 2020-08-18
Kernel/CommandLine parsing with preallocation to prevent rehashing

tryfinally:
resize hashmap upfront
reduce string copies


Kernel/CommandLine.cpp
```diff
@@ -46,17 +46,19 @@ CommandLine::CommandLine(const String& string)
 {
     s_the = this;
 
-    for (auto str : m_string.split(' ')) {
+    const auto& args = m_string.split(' ');
+    m_params.ensure_capacity(args.size());
+    for (auto&& str : args) {
```
alimpfard, line 51: with `args` being `const`, does this forced const xvalue make any sense?
it's not particularly _wrong_, just weird to see it taken as a `const String&&` that isn't even modified as an expiring object...

tryfinally: @alimpfard  _What force const xvalue you see?_
args is a const&  so that it capture teh temporary returned by the m_string.split(...)
str is actually a const&, no problem here because we only call split on it and split is a const mem-function.
we then split str and capture the result by value. the move is on the elements of this vector.

alimpfard: `for (auto&& str : args)`
That reads like a const xvalue, it'd make sense if you hadn't extended the lifetime of `args` by naming it; though it works out in this case because it's expiring anyway (end of function)

tryfinally: actually the auto && inside the ranger for is a forwarding reference and resolved to const lvalue. The auto && is not rvalue ref.
I might be wrong, but I don't see a problem here

alimpfard: It is indeed not an rvalue (or, well, an xvalue here), though it really served to confuse me.
Not sure what you stand to gain from it, but sure, it's fine.

tryfinally: Do you think it should be rewritten as
`for(const auto& str : args)` ?

alimpfard: To be less odd-looking? yeah.
To be correct? No, it's already correct (thanks for making me go read up on "forwarding references")

awesomekling:
Please format your commit messages like every other commit message in the project. :)

PR #3214, opened 2020-08-18
ChessEngine: A chess engine for Serenity's Chess app

Petelliott:
Plus the creation of LibChess for making other chess engines and some miscellaneous chess fixes.

Also I created some helper functions in `Core::File` for getting stdin, stdout, and stderr.


Libraries/LibChess/Chess.h
```diff
@@ -200,16 +180,21 @@ void Chess::generate_moves(Callback callback, Colour colour) const
...
-                try_move({sq, {sq.rank-1, sq.file}}) &&
-                try_move({sq, {sq.rank-2, sq.file}}) &&
-                try_move({sq, {sq.rank+1, sq.file+1}}) &&
-                try_move({sq, {sq.rank+1, sq.file-1}}) &&
-                try_move({sq, {sq.rank-1, sq.file+1}}) &&
-                try_move({sq, {sq.rank-1, sq.file-1}});
+            for (auto& piece : Vector({ Type::None, Type::Knight, Type::Bishop, Type::Rook, Type::Queen })) {
+                // clang-format off
+                keep_going =
+                    try_move({sq, {sq.rank+1, sq.file}, piece}) &&
```
alimpfard, line 186: Unless you really want `A &&` here, you can start the lines with `&&` and clang-format will keep them on different lines IIRC

Petelliott: Thanks! that really annoyed me

Libraries/LibChess/Chess.h
```diff
@@ -32,80 +32,84 @@
...
+    }
+    bool operator==(const Square& other) const { return rank == other.rank && file == other.file; }
+
+    template<typename Callback>
+    static void for_each(Callback callback)
+    {
+        for (int rank = 0; rank < 8; ++rank) {
+            for (int file = 0; file < 8; ++file) {
+                if (callback(Square(rank, file)) == IterationDecision::Break) {
+                    goto exit;
```
awesomekling, line 82: How about just `return`? :)

Petelliott: not sure how that happened

Libraries/LibChess/Chess.cpp
```diff
@@ -24,13 +24,60 @@
...
+    default:
+        return "";
+    }
+}
+
+Chess::Type piece_for_char(const StringView& str)
+{
+    String string = String(str).to_lowercase();
+    if (string == "")
+        return Type::None;
```
Dexesttp, line 60: Possibly a typo ? :P
```suggestion
    if (string == "")
        return Type::Pawn;
```

Libraries/LibChess/Chess.cpp
```diff
@@ -452,12 +533,14 @@ Chess::Result Chess::game_result() const
...
+bool Board::is_promotion_move(const Move& move, Colour colour) const
 {
     if (colour == Colour::None)
         colour = turn();
 
-    if (!is_legal(move, colour))
+    Move queen_move = move;
+    queen_move.promote_to = Type::Queen;
+    if (!is_legal(queen_move, colour))
         return false;
```
Dexesttp, line 544: Not sure about this one, but it feels like a legal knight promotion would not be accepted by this algorithm.

Libraries/LibChess/UCICommand.cpp
```diff
@@ -0,0 +1,333 @@
...
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include "UCICommand.h"
+#include "AK/StringBuilder.h"
```
Dexesttp, line 28:
```suggestion
#include <AK/StringBuilder.h>
```

Libraries/LibChess/UCIEndpoint.cpp
```diff
@@ -0,0 +1,132 @@
...
+NonnullOwnPtr<Command> Endpoint::read_command()
+{
+    String line(ReadonlyBytes(m_in->read_line(4096).bytes()), Chomp);
+
+#ifdef UCI_DEBUG
+    dbg() << class_name() << " Recieved UCI Command: " << line;
+#endif
+
+    if (line == "uci") {
+        return make<UCICommand>(UCICommand::from_string(line));
```
Dexesttp, line 103: This (and below).

How about making the Command::from_string return a `NonNullOwnPtr` directly instead ? This way, you don't have to create a temporary object before wrapping it in its pointer.

Also, you could return an `OwnPtr` instead of a `NonNullOwnPtr` to handle invalid input cleanly, but maybe that's not needed (right now, all invalid inputs would assert)

Libraries/LibChess/Chess.cpp
Petelliott: This is not a typo, but I see how it could be confusing. `piece_for_char` is used only for promotions, where no provided piece give `promote_to = Type::None`

Libraries/LibChess/Chess.cpp
Petelliott: `is_promotion_move` checks if a pawn is being promoted regardless of what it's being promoted to (even if the promoted piece is invalid). The move `h8=N` or `h8` would become `h8=Q`, only for the purpose of checking that the pawn is in promotion position. This method is used by `ChessWidget` to spawn a `PromotionDialog` to select the type of piece to promote to.

Libraries/LibChess/UCIEndpoint.cpp
Petelliott: > How about making the Command::from_string return a `NonNullOwnPtr` directly instead ? This way, you don't have to create a temporary object before wrapping it in its pointer.

I would consider this a premature optimization, given that for each command you'll spend a second waiting for the engine. Also I think there is a chance that `from_string` could be paired with `send_command`.

> Also, you could return an `OwnPtr` instead of a `NonNullOwnPtr` to handle invalid input cleanly, but maybe that's not needed (right now, all invalid inputs would assert)

Since a chess engine is not user facing, any malformatted commands are a bug in something, and just ignoring them would likely lead to hangs, crashes, illegal moves and all other kinds of shenanigans. Also, because Chess is not particularly critical to system function, I think it is best to assert here.

Libraries/LibChess/Chess.cpp
Dexesttp: Oh, that makes sense ! And yeah, definitely, especially since its behaviour is not symmetric with `char_for_piece` (having `piece_for_char(char_for_piece(Type::Pawn)) == Type::Null` is very counter-intuitive)

Libraries/LibChess/UCIEndpoint.cpp
Dexesttp: Okay ! I think I misunderstood the purpose of UCI, for some reason I thought it could be used as an online protocol as well. It makes sense for it to ASSERT if it's only for local communications (like IPC currently does).

PR #3217, opened 2020-08-19
LibHTTP: Fix processing terminating chunk

### LINKED ISSUE #3197: ProtocolServer: Spins when downloading a resource from ajax.googleapis.com over HTTP
For example: [JQuery](http://ajax.googleapis.com/ajax/libs/jquery/1.12.4/jquery.min.js)

This is because for some reason, we do not send a fin ack to them:
Wireshark_iq7Z3oGUa4

The stack from system monitor says that protocol server is spinning waiting for an event:
vmplayer_fMGlpcndVh

However, getting a 404 works, so it's not just because it's ajax.googleapis.com.

It works over HTTPS, so no problem there.

tomuta:
After encountering the terminating chunk we need to read the
trailing headers line by line, until encountering the final
empty line.

Fixes #3197

### DIFF
--- a/Libraries/LibHTTP/Job.cpp
+++ b/Libraries/LibHTTP/Job.cpp
@@ -117,12 +117,12 @@ void Job::on_socket_connected()
             m_state = State::InHeaders;
             return;
         }
-        if (m_state == State::InHeaders || m_state == State::AfterChunkedEncodingTrailer) {
+        if (m_state == State::InHeaders || m_state == State::Trailers) {
             if (!can_read_line())
                 return;
             auto line = read_line(PAGE_SIZE);
             if (line.is_null()) {
-                if (m_state == State::AfterChunkedEncodingTrailer) {
+                if (m_state == State::Trailers) {
                     // Some servers like to send two ending chunks
                     // use this fact as an excuse to ignore anything after the last chunk
                     // that is not a valid trailing header.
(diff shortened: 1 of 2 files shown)


alimpfard:
Were we not already doing that?

Seems like the main thing that fixes #3197 is https://github.com/SerenityOS/serenity/pull/3217/commits/dca5681a4fa30302b4ae17fdaf067801e9faecfc#diff-65e04583e298e6f1470366065aca28ddR189

So...could you point that out in the commit message too?

tomuta:
> Were we not already doing that?

No, we were not. I added a more detailed description to why it's failing in #3197.

> Seems like the main thing that fixes #3197 is [dca5681#diff-65e04583e298e6f1470366065aca28ddR189](https://github.com/SerenityOS/serenity/commit/dca5681a4fa30302b4ae17fdaf067801e9faecfc#diff-65e04583e298e6f1470366065aca28ddR189)
> 
> So...could you point that out in the commit message too?

No, that block of code isn't hit at all in this case. if `size_lines .size() == 0` then the server would be sending an invalid response with a missing chunk size. Hence the call to `finish_up`. That would be a badly behaving server.

alimpfard:
> I added a more detailed description to why it's failing in #3197.

Ah, I missed that.

_Understandable, have a good day_ lol.

PR #3220, opened 2020-08-19
LibCompress: Turn DEFLATE implementation into a stream.

asynts:
Previously, the implementation would produce a `Vector<u8>` which would contain the entire decompressed data.

With these changes it is still necessary to store the whole input data in one piece (I am working on this next,) but the output can be read block by block. (That's not optimal either because blocks can be arbitrarily large, but it's good for now.)


Libraries/LibCompress/Deflate.h
```diff
@@ -65,34 +66,120 @@ class CanonicalCode {
...
-    CanonicalCode m_literal_length_codes;
-    CanonicalCode m_fixed_distance_codes;
+    void decompress_uncompressed_block() const;
+    void decompress_static_block() const;
+    void decompress_dynamic_block() const;
+    void decompress_huffman_block(CanonicalCode&, CanonicalCode*) const;
+
+    Vector<CanonicalCode> decode_huffman_codes() const;
+    u32 decode_run_length(u32) const;
+    u32 decode_distance(u32) const;
```
awesomekling, line 165: It doesn't seem healthy to have all of these `const` when they mutate the internal state of the stream

asynts: The problem is that there can be empty blocks, so `eof()` is not trivial, but that the signature is `bool eof() const`. I can't think of a way to avoid this.

PR #3221, opened 2020-08-19
2048: board size 'n' target score

### LINKED ISSUE #3219: 2048: Game should allow you to continue towards 4096 if you like
More challenges for the people!

alimpfard:
This makes "New game" prompt the user for a board size and a target
tile.
Closes #3219.

cc @bugaevc


alimpfard:
Don't mind me, just playing my 2048 game with a target score of 2^31 on a 32x32 board lol

shot-2020-08-19_20-51-54

bugaevc:
> The logic only works with nxn grids

How about we fix that instead? :smiley: I imagine there shouldn't be much that doesn't work...

alimpfard:
But the logic is based on the assumption that the grid is transposable to its own size
And why do that when a rectangular grid looks ugly anyway...

Games/2048/Game.h
```diff
@@ -30,7 +30,7 @@
 
 class Game final {
 public:
-    Game(size_t);
+    Game(size_t board_size, size_t target_score = 0);
```
bugaevc, line 33: This accepts target *score*, but the implementation accepts target *tile*.

alimpfard: Oh, I missed that
Thanks for catching it~

nico:
Can we make 'new game' create a default 4x4 grid and have a 2nd menu item "new game with custom size" or similar that invokes the dialog?

awesomekling:
Dios mio

alimpfard:
How about keeping the same size across "New game" and resets until changed by a second menu "New custom game"?

nico:
> How about keeping the same size across "New game" and resets until changed by a second menu "New custom game"?

Kind of sounds more like a settings dialog than a "new game" action then maybe?

alimpfard:
But moving it to a "settings" dialog would imply that it applies to the current game - which it doesn't.
hmm...

nico:
Maybe "New Game" could start a standard game always, and "new custom game" could remember what was filled in last time so that you only have to hit return to create a new game with a nondefault size?

alimpfard:
I think I like the settings approach better.

PR #3224, opened 2020-08-19
LibWeb: Implement <template> parsing

Lubrsi:
Note that there is currently no way to display them as we can't
currently clone nodes.

Adds special case for templates for dumping to console.
Doesn't add it to the DOM inspector as I'm not sure how to do it.


Libraries/LibWeb/HTML/HTMLTemplateElement.h
```diff
@@ -36,6 +37,14 @@ class HTMLTemplateElement final : public HTMLElement {
...
     HTMLTemplateElement(DOM::Document&, const FlyString& local_name);
     virtual ~HTMLTemplateElement() override;
+
+    NonnullRefPtr<DOM::DocumentFragment> content() { return m_content; }
+    const NonnullRefPtr<DOM::DocumentFragment> content() const { return m_content; }
+
+private:
+    DOM::Document& appropriate_template_contents_owner_document(DOM::Document&);
+
+    DOM::DocumentFragment m_content;
```
awesomekling, line 47: This is not safe. We must store the `DOM::DocumentFragment` in a `RefPtr` of some type, or it will be destroyed when the `HTMLTemplateElement` is destroyed, and then any outstanding `RefPtr`s to this fragment will be dangling.

awesomekling (changes requested):

PR #3228, opened 2020-08-20
LibJS: Basic implementation of most of Date's constructor arguments

nico:
The constructor with a string argument isn't implemented yet, but
this implements the other variants.

The timestamp constructor doens't handle negative timestamps correctly.

Out-of-bound and invalid arguments aren't handled correctly.

### DIFF
--- a/Kernel/RTC.cpp
+++ b/Kernel/RTC.cpp
@@ -155,7 +155,7 @@ time_t now()
     unsigned year, month, day, hour, minute, second;
     read_registers(year, month, day, hour, minute, second);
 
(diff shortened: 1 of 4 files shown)


nico:
The LibCore change is wrong, hold off for now.

nico:
Ok, better.

nico:
Maybe "New Game" could start a standard game always, and "new custom game" could remember what was filled in last time so that you only have to hit return to create a new game with a nondefault size?

awesomekling:
Ok!

PR #3229, opened 2020-08-20
Kernel: Add an atomic Singleton class

tomuta:
This allows easier one-time thread-safe atomic initialization
of the various singletons in the Kernel.


tryfinally:
Woa, you beat me to the PR :-)

Kernel/Net/IPv4Socket.cpp
```diff
@@ -45,11 +46,12 @@
 
 namespace Kernel {
 
+static Singleton<Lockable<HashTable<IPv4Socket*>>, []() {
+    return new Lockable<HashTable<IPv4Socket*>>;
+}> s_table;
+
```
awesomekling, line 52: This feels like macro territory.

tomuta: What if the code in the initialization function needs to be a bit more complex? Or are you purely referring to the definition?

tomuta:
Added a `MAKE_SINGLETON` macro.

tomuta:
Hm macro expansion isn't too happy with the commas in template arguments

tomuta:
Added a `make_singleton` instead

tomuta:
OK I converted most (if not all) singletons over to this.

PR #3231, opened 2020-08-20
AK: Remove exceptions from TestSuite.h.

asynts:
This pull request makes it possible to use `AK/TestSuite.h` in Serenity. Previously, this was not possible because Serenity is build with `-fno-exceptions` and exceptions were used in that header.

However, since nobody actually throws an exception like that, it can be removed without issues.

I've also changed `Userpace/test-compress.cpp` to use this header.

### DIFF
--- a/AK/TestSuite.h
+++ b/AK/TestSuite.h
@@ -63,32 +63,6 @@ private:
     std::chrono::time_point<clock> m_started;
 };
 
-class TestException {
-public:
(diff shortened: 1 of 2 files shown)


asynts:
I just noticed that this header also includes `<chrono>`, that won't work. I'll close this and reopen when I fixed that issue as well. (Why does this even build?)

PR #3233, opened 2020-08-20
Kernel: Fix race condition using thread safe singelton

### LINKED ISSUE #3226: Kernel/Net Race Condition?
https://github.com/SerenityOS/serenity/blob/413db2d6d5a49400d9587c9aa37bedd83968a6dd/Kernel/Net/IPv4Socket.cpp#L50-L53

There is a potential race condition on `s_table` .
Shouldn't checking if null, assignment to and dereferencing s_table be  **_atomic_** ?
I have prepared a _fix_, but am holding the PR.

tryfinally:
- [x] Modify Atomic to support Atomic enums
- [x] Implement thread safe singleton usable from both kernel and userland + Unit test
singleton uses atomic acquire/release memory order for performance
- [x] fix #3226 
- [x] fix other race condition in kernel
- [x] fix one race in AK FlyString

### DIFF
--- a/AK/Atomic.h
+++ b/AK/Atomic.h
@@ -85,7 +85,6 @@ template<typename T, typename V = typename RemoveVolatile<T>::Type>
         return __atomic_compare_exchange_n(const_cast<V**>(var), &expected, nullptr, false, order, order);
 }
 
-
 template<typename T>
 static inline T atomic_fetch_add(volatile T* var, T val, MemoryOrder order = memory_order_seq_cst) noexcept
 {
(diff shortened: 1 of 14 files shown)


tryfinally:
The race condition is present in other parts of the OS 
I will push more commits later to fix

awesomekling:
Please format your commit messages like all other commit messages in the project.
See the [contribution guidelines](https://github.com/SerenityOS/serenity/blob/master/CONTRIBUTING.md)

tomuta:
While very similar approach, I like my implementation in #3229 a little better. Rather than doing a compare_exchange every time, it's simply doing a load and in the unlikely event that it hasn't been initialized do the more expensive operations. My approach also only uses one pointer.

tryfinally:
@awesomekling I there is a typo in the commit message: Kernet should be Kernel.
Anything else I missed?

PR #3236, opened 2020-08-20
Kernel: Add support for VirtIO

tomuta:
Still very much work in progress...


ccapitalK:
If anyone dares, we could even do hardware accelerated graphics (if someone tries to implement Virgl). :D

awesomekling:
Superseded by #4748

PR #3237, opened 2020-08-20
AK+Kernel+LibC: Move out the implementation of memmem / Shell: Make 'for' loops read their input as an stream 

alimpfard:
This makes `for` loops process their inputs as an stream.
Though it should be noted that the shell will still wait for the jobs to finish - that shall be a FIXME for now.


alimpfard:
cc @nico - you asked for this a billion years ago.

PR #3259, opened 2020-08-22
Calendar: A E S T H E T I C S

thankyouverycool:
New and improved calendar widget and a comfy mini-cal for the clock
mini-cal


nico:
This looks super nice.

Do you think the date string in the menu bar should be drawn with a menu highlight while this is open, like the menu items in the upper left do when they're active?

thankyouverycool:
> menu highlight

Mate does something similar in its systray and it looks pretty slick, but all its applets open into actual menus. I think it would provide a nice looking contrast to the toolbar text, just not sure it sends the right visual clue considering the window doesn't have any menu options to highlight (for now).

PR #3272, opened 2020-08-23
LibJS: Implement Date.getUTC*

nico:
The motivation, obviously, is to use Date.getUTCDay() to test the tm_wday field after calling gmtime.


BenWiederhake:
> supposed to canonicalize out-of-range values (that is, "Jan 31" is turned into "Feb 1").

Is that just a harmless typo, or should I be worried that my calender is buggy? :D

PR #3275, opened 2020-08-23
Introduce strlcpy, clean up wrong str{,n}cpy uses

BenWiederhake:
This PR:
- Fixes five obscure bugs caused by wrong strcpy/strncpy usages (they are pretty hard to trigger)
- Fixes a few unstylish strcpy/strncpy uses that *technically* don't break, but don't look right either
- Introduces strlcpy, a "safer" variant of strncpy, popular with *BSD folks
- Uses it to fix the above stuff

EDIT:

If anyone cares, I left some unsafe-looking usages of strcpy/strncpy in, because it was too convoluted to fix in the same drive-by fashion that I did with the other bugs:
- LibC/scanf.cpp: Lots of manual copying. If there is *one* mistake in there, it all blows up.
- LibC/termcap.cpp: Those usages of `strncpy` look suspicious to me.
- LibCrypto/crypt.cpp: Looks like manual arrangement, but using static offsets. Either they are always correct, or always broken.
- Userland/Tests: Nah.

### DIFF
--- a/Libraries/LibCore/SocketAddress.h
+++ b/Libraries/LibCore/SocketAddress.h
@@ -43,7 +43,7 @@ public:
         Local
     };
 
-    SocketAddress() {}
+    SocketAddress() { }
     SocketAddress(const IPv4Address& address)
         : m_type(Type::IPv4)
         , m_ipv4_address(address)
@@ -82,12 +82,14 @@ public:
         }
     }
 
-    sockaddr_un to_sockaddr_un() const
+    Optional<sockaddr_un> to_sockaddr_un() const
     {
         ASSERT(type() == Type::Local);
         sockaddr_un address;
         address.sun_family = AF_LOCAL;
-        RELEASE_ASSERT(m_local_address.length() < (int)sizeof(address.sun_path));
+        if (m_local_address.length() >= sizeof(address.sun_path)) {
+            return {};
+        }
         strcpy(address.sun_path, m_local_address.characters());
         return address;
     }
--- a/Services/SystemServer/Service.cpp
+++ b/Services/SystemServer/Service.cpp
@@ -135,7 +135,12 @@ void Service::setup_socket()
     }
 
     auto socket_address = Core::SocketAddress::local(m_socket_path);
-    auto un = socket_address.to_sockaddr_un();
+    auto un_optional = socket_address.to_sockaddr_un();
+    if (!un_optional.has_value()) {
+        dbg() << "Socket name " << m_socket_path << " is too long. BUG! This should have failed earlier!";
+        ASSERT_NOT_REACHED();
+    }
+    auto un = un_optional.value();
     int rc = bind(m_socket_fd, (const sockaddr*)&un, sizeof(un));
     if (rc < 0) {
         perror("bind");
@@ -358,6 +363,8 @@ Service::Service(const Core::ConfigFile& config, const StringView& name)
     ASSERT(!m_accept_socket_connections || (!m_socket_path.is_null() && m_lazy && m_multi_instance));
     // MultiInstance doesn't work with KeepAlive.
     ASSERT(!m_multi_instance || !m_keep_alive);
+    // Socket path (plus NUL) must fit into the structs sent to the Kernel.
+    ASSERT(m_socket_path.length() < UNIX_PATH_MAX);
 
     if (!m_socket_path.is_null() && is_enabled()) {
         auto socket_permissions_string = config.read_entry(name, "SocketPermissions", "0600");
--- a/AK/String.cpp
+++ b/AK/String.cpp
@@ -36,10 +36,6 @@
 #    include <inttypes.h>
 #endif
 
-#ifdef KERNEL
-extern "C" char* strstr(const char* haystack, const char* needle);
-#endif
-
 namespace AK {
 
 String::String(const StringView& view)
--- a/AK/String.h
+++ b/AK/String.h
@@ -142,6 +142,7 @@ public:
     bool is_null() const { return !m_impl; }
     ALWAYS_INLINE bool is_empty() const { return length() == 0; }
     ALWAYS_INLINE size_t length() const { return m_impl ? m_impl->length() : 0; }
+    // Includes NUL-terminator, if non-nullptr.
     ALWAYS_INLINE const char* characters() const { return m_impl ? m_impl->characters() : nullptr; }
 
     ALWAYS_INLINE ReadonlyBytes bytes() const { return m_impl ? m_impl->bytes() : nullptr; }
--- a/AK/StringImpl.h
+++ b/AK/StringImpl.h
@@ -59,6 +59,7 @@ public:
     ~StringImpl();
 
     size_t length() const { return m_length; }
+    // Includes NUL-terminator.
     const char* characters() const { return &m_inline_buffer[0]; }
 
     ALWAYS_INLINE ReadonlyBytes bytes() const { return { characters(), length() }; }
--- a/Kernel/StdLib.cpp
+++ b/Kernel/StdLib.cpp
@@ -97,15 +97,6 @@ const void* memmem(const void* haystack, size_t haystack_length, const void* nee
     return AK::memmem(haystack, haystack_length, needle, needle_length);
 }
 
-char* strcpy(char* dest, const char* src)
-{
-    auto* dest_ptr = dest;
-    auto* src_ptr = src;
-    while ((*dest_ptr++ = *src_ptr++) != '\0')
-        ;
-    return dest;
-}
-
 void memset_user(void* dest_ptr, int c, size_t n)
 {
     ASSERT(Kernel::is_user_range(VirtualAddress(dest_ptr), n));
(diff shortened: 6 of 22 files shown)


Libraries/LibCore/SocketAddress.h
```diff
@@ -82,12 +82,15 @@ class SocketAddress {
...
 
-    sockaddr_un to_sockaddr_un() const
+    Optional<sockaddr_un> to_sockaddr_un() const
     {
         ASSERT(type() == Type::Local);
         sockaddr_un address;
         address.sun_family = AF_LOCAL;
-        RELEASE_ASSERT(m_local_address.length() < (int)sizeof(address.sun_path));
+        if (m_local_address.length() >= sizeof(address.sun_path)) {
+            errno = E2BIG;
```
awesomekling, line 91: LibCore does not communicate errors through `errno`

Services/SystemServer/Service.cpp
```diff
@@ -135,7 +135,12 @@ void Service::setup_socket()
     }
 
     auto socket_address = Core::SocketAddress::local(m_socket_path);
-    auto un = socket_address.to_sockaddr_un();
+    auto un_optional = socket_address.to_sockaddr_un();
+    if (!un_optional.has_value()) {
+        perror("bind");
```
awesomekling, line 140: This is a lie, we're not reporting an error from the `bind` syscall

Libraries/LibCore/SocketAddress.h
BenWiederhake: Then I'll remove it.

awesomekling:
Really nice stuff! :)

Services/SystemServer/Service.cpp
BenWiederhake: It absolutely is! I was having a hard time coming up with anything.

Now that I think about it more deeply, I notice that it will end up writing `bind: Success` which is arguably even worse than the lie itself.

So it's `dbg()` then?

BenWiederhake: @hard time: I mean how to deal with error propagation across the stack, not only in this particular function.

awesomekling: Yeah just `dbg()` :)

BenWiederhake: Switched to dbg().

bugaevc:
Whether using `strncpy()` or `strlcpy()`, we should never just carry on in case the string gets truncated. At the very least, let's mark their return values `[[nodiscard]]`, and always either assert the strings fit, or return an error up (or very rarely, explicitly ignore truncation if that's what we want in that case for some reason). Or perhaps let's introduce a variant of `str*cpy()` that automatically asserts the string fits.

BenWiederhake:
While that sounds like a good idea at first, there's to issues with marking them as `[[nodiscard]]`:
- It would break C/C++ code that assumes the result can be discarded, and that will heavily affect all Ports.
- Even Serenity usually disregards it.

I agree that it should be checked, somehow – but I don't know how. Introducing checks *everywhere* seems a bit intrusive. Also, in many cases (like reading `/etc/services` and `/etc/groups`, see the commits above) it seems like a be a bad idea to terminate the process.

bugaevc:
> It would break C/C++ code that assumes the result can be discarded, and that will heavily affect all Ports.

It's a warning, it won't break ports, those are built without `Werror`.

> Even Serenity usually disregards it.

That's my point, we should stop doing that.

> I agree that it should be checked, somehow – but I don't know how. Introducing checks _everywhere_ seems a bit intrusive.

It's not *checking*, it's more of *ensuring* there's always enough space or we're prepared to deal with truncation. There are basically two possibilities:

* You measure the string first, then allocate a large enough buffer, then use `strncpy()`/`strlcpy()`. In that case, you always expect the string to fit, so you should `ASSERT()` that it does. If that's too repetitive, we could introduce a variant that does it automatically, as I suggested above.

* You use a fixed size buffer, and ~hope for the better~ properly detect truncation and return an error to the caller / print an error message to the user. In that case, yes, you'll always have to check for it explicitly.

I don't think it's too intrusive, not compared to all the system call whose result you have to check, or all those `KResult`-returning functions in the kernel. The good thing is most of our code uses the AK string types, and so doesn't have any of these truncation/overflow issues.

[See also](https://lwn.net/Articles/612244/)

BenWiederhake:
I am aware of the religious war around strncpy/strlcpy. However, given the wrong usages of strncpy (see: off-by-ones fixed by the above commits), I think that statements like this (from the LWN article) don't help any discussion:

> This is horribly inefficient BSD crap. Using these function only leads to other errors.

Replacing C-strings by AK::Strings sounds like a good idea. If a side effect of that is that `strlcpy` gets removed, I don't mind.

In either case, my main motivation was to find and fix existing bugs, fix them, make future bugs less likely. I've satisfied that appetite, and I'd be happy to see even further improvements.

You're right, a better "fix" should include proper error propagation instead of chugging on; for the cases I looked at, "chugging on" felt acceptable enough and I didn't want to rewrite such a large chunk of LibC.

bugaevc:
> I think that statements like this (from the LWN article)

(To make it clear in case it's not, that quote is from Ulrich Drepper, not from Jonathan Corbet, the LWN article author/editor.)

> In either case, my main motivation was to find and fix existing bugs, fix them, make future bugs less likely

You have certainly made some nice improvements 😀

But as I continue to stress, any time you use either of `str*cpy()` variants with a fixed size buffer without checking the result, there's still a bug (that sometimes could be escalated to a vulnerability) of silently truncating the string. Imagine what happens if a setuid program is tricked to act on one group instead of another because one name happens to be another's prefix, or on one network interface instead of another.

It's not just a potential future improvement, or quote-unquote fix. Until we get this right everywhere, swapping `str*cpy()` variants and fixing off-by-ones will only prevent buffer overflows (which is, again, important!), but the underlying bugs of carelessly using fixed size buffers is still there.

> I've satisfied that appetite, and I'd be happy to see even further improvements.

Sounds like I better work on that myself then, heh 🙂

BenWiederhake:
> Imagine what happens if a setuid programt is tricked to act on one group instead of another because one name happens to be another's prefix, or on one network interface instead of another.

Absolutely! That's why I only called my improvements "acceptable enough". :)

bugaevc:
I propose the following API:

```cpp
namespace AK {

enum class TruncationMode {
    AssertFits,
    Detect,
};

class String {

    bool copy_characters_to_buffer(char* buffer, size_t buffer_size, TruncationMode) const;
};


bool String::copy_characters_to_buffer(char* buffer, size_t buffer_size, TruncationMode truncation_mode) const
{
    // We must fit at least the NUL-terminator.
    ASSERT(buffer_size > 0);

    size_t characters_to_copy = min(length(), buffer_size - 1);

    if (truncation_mode == TruncationMode::AssertFits)
        ASSERT(characters_to_copy == length());

    memcpy(buffer, characters(), characters_to_copy);
    buffer[characters_to_copy] = 0;

    return characters_to_copy == length();
}

}
```

This is how you would use it: say, in `grp.cpp`, instead of

https://github.com/SerenityOS/serenity/blob/c2b9f8857c7b9189dab5eca8be06ecc8811968eb/Libraries/LibC/grp.cpp#L142

you'd do

```cpp
if (!e_name.copy_characters_to_buffer(__grdb_entry->name_buffer, GRDB_STR_MAX_LEN, TruncationMode::Detect)) {
    errno = ENAMETOOLONG;
    return nullptr;
}
```

I've started implementing this; in the meantime, better ideas and suggestions are welcome :smile:

awesomekling:
My main suggestion would be to replace C string usage by C++ strings wherever possible.
That stuff in grp.cpp looks like it could be C++ strings instead, for example :)

BenWiederhake:
I think he suggests exactly that, and `copy_characters_to_buffer` is about how to get the string data our of a `AK::String` again.

Maybe one simplification:
- make the return value of `copy_characters_to_buffer` `[[nodiscard]]`
- remove the argument to `copy_characters_to_buffer`
- if the caller feels confident, they can do `ASSERT(mystring.copy_characters_to_buffer(…))`
- if the caller wants to actually check, they can do `if (!mystring.copy_characters_to_buffer(…))`

How about that? :)

Btw: `grp` is used at least in `SystemServer`, so it should be possible to continue *somehow* with the iteration, to make sure the system isn't always bricked. "Skipping" a group seems safe to me. Does your proposed diff take that into account, i.e., leave the global state behind such that the caller can continue?

(Also, shouldn't the standard say something about this? I'm sure we're not the first to ever wonder what to do in case of a broken `/etc/groups`.)

bugaevc:
> I think he suggests exactly that, and `copy_characters_to_buffer` is about how to get the string data our of a `AK::String` again.

No, I wasn't suggesting to port `grp.cpp` (and `pwd.cpp`) to store string in `AK::String` internally; but that is indeed a great idea, so I've implemented it now. We don't need to copy out the characters in that case though, because `struct group` only contains a pointer to the buffer.

> Maybe one simplification

Yeah, I actually ended up doing just that.

PR incoming :wink:

BenWiederhake:
As a bit of background: I suggested to do it this way because it's part of LibC, and I half-expect that eventually maybe some port will try to use it, with the "usual" definition – I haven't actually tried to figure out how reasonable that is.

bugaevc:
If you mean `grp.cpp`, then the proposed changes don't break any public APIs, it'll only ensure we don't choke on long group names. It should be more, not less, robust.

Or what do you mean?

PR #3279, opened 2020-08-24
Applications: Add a (very basic) Spreadsheet

alimpfard:
This PR adds a spreadsheet program, because why not.
(actually, because I watched this guy make a web-based one in F# in NDC, and I was like "me too")

It comes with a single sheet and a bunch of cells, a few tweaks to LibGUI, a javascript runtime (yes, the expressions are in js), it automatically updates cells that depend on each other, and can save (but not load) in a pretty lossy way to JSON.
It's not particularly capable yet, but we'll get there!

Obligatory image demonstrating just about everything it can do rn:
shot-2020-08-24_13-29-03


Base/res/js/SpreadSheet/runtime.js
```diff
@@ -0,0 +1,190 @@
+const sheet = this
```
linusg, line 1: Can we please put `Libraries/LibJS/Tests/.prettierrc` into the repository root and reformat this file accordingly for consistency?

linusg: also can we maybe stick to `camelCase` in JavaScript land, the mixture of `camelCase` built-ins and `snake_case` additional functions looks strange

alimpfard: Sure.
let's see if I can't install prettier.

the naming thing poses an interesting question, do we want to keep the javascript-esque style in the spreadsheet expressions as well? should, for instance, `sumif` be `sumIf` to comply with js naming standards - which I have no idea about?

Applications/SpreadSheet/SpreadSheet.cpp
```diff
@@ -0,0 +1,294 @@
...
+#include <AK/GenericLexer.h>
+#include <LibCore/File.h>
+#include <LibJS/Parser.h>
+#include <LibJS/Runtime/Error.h>
+#include <LibJS/Runtime/GlobalObject.h>
+#include <LibJS/Runtime/Object.h>
+
+namespace SpreadSheet {
+
+class SheetGlobalObject : public JS::GlobalObject {
```
awesomekling, line 41: Please use the `JS_OBJECT` macro to ensure that all the right helpers get inserted.

Applications/SpreadSheet/SpreadSheet.cpp
```diff
@@ -0,0 +1,294 @@
...
+    SheetGlobalObject(Sheet& sheet)
+        : m_sheet(sheet)
+    {
+    }
+
+    virtual ~SheetGlobalObject() override
+    {
+    }
+
+    virtual JS::Value get(const JS::PropertyName& name, JS::Value receiver = {}) const override
```
awesomekling, line 52: Hmm: I wonder if we should virtualize `get_own_property` as well as `get`.
In this case I don't think it matters, but for non-global objects we probably don't want to hijack property chain lookups.

Applications/SpreadSheet/SpreadSheet.cpp
```diff
@@ -0,0 +1,294 @@
...
+        return m_columns.last();
+    }
+
+    TODO();
+}
+
+void Sheet::update()
+{
+    m_visited_cells_in_update.clear();
+    for (auto& it : m_cells) {
```
awesomekling, line 166: Since you have to repeatedly access `it.value` here, it might be nice to alias it:
```cpp
auto& cell = *it.value;
```

Applications/SpreadSheet/SpreadSheet.cpp
```diff
@@ -0,0 +1,294 @@
...
+
+JS::Value Sheet::evaluate(const StringView& source, Cell* on_behalf_of)
+{
+    TemporaryChange cell_change { m_current_cell_being_evaluated, on_behalf_of };
+
+    auto parser = JS::Parser(JS::Lexer(source));
+    if (parser.has_errors())
+        return JS::js_undefined();
+
+    auto program = parser.parse_program();
```
awesomekling, line 196: Unrelated to your patch, but LibJS should have a convenience function for this :)

Applications/SpreadSheet/SpreadSheet.cpp
```diff
@@ -0,0 +1,294 @@
...
+// FIXME: Find a better way to figure out dependencies
+void Cell::reference_from(Cell* other)
+{
+    if (!other || other == this)
+        return;
+
+    if (!referencees.find([other](auto& ptr) { return ptr.ptr() == other; }).is_end())
+        return;
+
+    referencees.append(other->make_weak_ptr());
```
awesomekling, line 295: `referencee` would be someone being referred *to*.
In this case, they are `referencers`, but perhaps `referencing_cells` would be a bit easier to brain-parse? :)

Applications/SpreadSheet/SpreadSheet.h
```diff
@@ -0,0 +1,209 @@
...
+
+        evaluated_data = move(new_data);
+    }
+
+    String source() const;
+
+    JS::Value js_data();
+
+    void update(Badge<Sheet>) { update_data(); }
+    void update();
```
awesomekling, line 119: What is the difference between these two? From just looking at the API I would think "oh, well if I'm not in `Sheet`, I can still update, I just don't pass a badge"

alimpfard: The difference is that `update()` goes through `Sheet` as well.
This ensures that there is no equirecursion.

Applications/SpreadSheet/SpreadSheet.cpp
alimpfard: Not everyone would want to get an `undefined` for parse errors though...?

awesomekling: That's true!

Applications/SpreadSheet/SpreadSheet.cpp
alimpfard: Why isn't it the other way around anyway?
An object shouldn't have a particularly good reason to override chain lookup order

nico:
bikeshed: wdyt about "Spreadsheet" instead of "SpreadSheet"? It's not a "spread sheet", it's a "spreadsheet".

alimpfard:
> bikeshed: wdyt about "Spreadsheet" instead of "SpreadSheet"? It's not a "spread sheet", it's a "spreadsheet".

suuuuuuure...I'd personally go for the branch name, all lowercase lol

_/me starts replacing names_

alimpfard:
Things may or may not have been lost in the process, I accidentally squashed two commits (they were luckily related and would've been squashed anyway, but still).
Let's see if it builds :P

PR #3282, opened 2020-08-24
Documentation: Update BuildInstructions.md

sahumada:
Add missing `cmake ..' line before compiling with make

### DIFF
--- a/Documentation/BuildInstructions.md
+++ b/Documentation/BuildInstructions.md
@@ -84,6 +84,7 @@ Once the toolchain and cmake have been built, go into the `Build/` directory and
 ```bash
 $ cd ..
 $ cd Build
+$ cmake ..
 $ make
 $ make install
 ```


nico:
`./BuildIt.sh` in toolchain is supposed to do that already, so this shouldn't be needed.

tomuta:
> `./BuildIt.sh` in toolchain is supposed to do that already, so this shouldn't be needed.

I actually found it quite confusing initially that BuildIt.sh did the cmake step for me. I think it would be better to not have it do that as you can't specifiy ninja vs make, or other -D arguments. And then sometimes certain modifications require you to manually run the cmake step.

All other projects I've worked with always require you to manually run the cmake step. Which of course doesn't prevent you from scripting your workflow.

sahumada:
you are right .. BuildIt.sh will do it .. but I didn't use it so that's why I had to manually run cmake 

up to you if you reject this PR

awesomekling:
This makes sense to me. It's indeed confusing that the toolchain build creates a partial `Build` directory (that you also have to get rid of if you want to switch to Ninja.)

PR #3283, opened 2020-08-24
Spreadsheet: Hotfixes & expression highlight

alimpfard:
The highlighter is just a conditional javascript highlighter, so I had to mark that guy non-final

Fixes two crashes as well.


alimpfard:
oh, it seems the menubar was lost in the process of renaming

Issue #3257, opened 2020-08-22
Kernel memory corruption: Global variables crossing into kalloc pool


tomuta:
Trying to figure out some odd bugs related to #3229, I found that global variables are crossing into `ETERNAL_BASE_PHYSICAL`, which causes random memory corruption. One issue similar (though I don't believe the same) is related to the kernel command line getting corrupted, which is solved in a commit in #3248. However, I'm not sure how to solve this issue deterministically. It seems like `BASE_PHYSICAL` and `ETERNAL_BASE_PHYSICAL` need to account for global variables and other stuff. Maybe it should start no lower than `end_of_kernel_image`?

This is an example of what can happen:
```
Thread 1 hit Hardware watchpoint 2: *0xc0200000

Old value = 1
New value = 3
AK::Bitmap::set_range () at .././AK/Bitmap.h:100
100	        for (size_t index = start; index < start + len; ++index) {
(gdb) bt
#0  AK::Bitmap::set_range () at .././AK/Bitmap.h:100
#1  0xc01a1c2a in kmalloc_allocate () at ../Kernel/Heap/kmalloc.cpp:130
#2  kmalloc_impl () at ../Kernel/Heap/kmalloc.cpp:184
#3  0xc012f989 in Kernel::Ext2FS::get_inode () at ../Kernel/FileSystem/Ext2FileSystem.cpp:653
#4  0xc012fe49 in Kernel::Ext2FSInode::lookup ()
    at ../Kernel/FileSystem/Ext2FileSystem.cpp:1501
#5  0xc0144ea8 in Kernel::VFS::resolve_path_without_veil ()
    at ../Kernel/FileSystem/VirtualFileSystem.cpp:933
#6  0xc014646d in Kernel::VFS::resolve_path ()
    at ../Kernel/FileSystem/VirtualFileSystem.cpp:885
#7  0xc0147077 in Kernel::VFS::access () at ../Kernel/FileSystem/VirtualFileSystem.cpp:441
#8  0xc016780c in Kernel::Process::sys$access () at ../Kernel/Syscalls/access.cpp:39
#9  0xc0167537 in handle () at ../Kernel/Syscall.cpp:118
#10 syscall_handler () at ../Kernel/Syscall.cpp:177
#11 0xc0166f6d in syscall_asm_entry ()
#12 0xc3965fac in ?? ()
```

[...]

PR #3287, opened 2020-08-24
Kmalloc corruption

tomuta:
Since #3248 was reverted entirely, bring back two commits, one with further improvements. Which happens to make all the other changes work, without breaking Browser.

We now log the kernel and kmalloc range, and exclude that from any of the regions MM considers.

Also see #3257

### DIFF
--- a/AK/FlyString.cpp
+++ b/AK/FlyString.cpp
@@ -27,6 +27,7 @@
 #include <AK/FlyString.h>
 #include <AK/HashTable.h>
 #include <AK/Optional.h>
+#include <AK/Singleton.h>
 #include <AK/String.h>
 #include <AK/StringUtils.h>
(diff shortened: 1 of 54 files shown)


tomuta:
As a simple validation step you can e.g. add a global variable to the kernel like `char foo[32 * PAGE_SIZE];` and you'll see the image/kmalloc range move accordingly. There's still a limit as to how much you can add to the kernel image size before it won't boot at all, I'm guessing maybe that's a grub limitation maybe. But this should prevent global variables from getting corrupted by kmalloc.

tomuta:
Interestingly, this seems to make the Singleton changes work without breaking Browser :)

tomuta:
Pulled those changes back into this pull request.

awesomekling:
Excellent :)

PR #3289, opened 2020-08-25
Kernel: Scroll wheel amount on Intellimouse is 4 bits instead of 8

Lubrsi:
When trying to scroll up on virtualizers that don't use the VMware
backdoor and instead use PS2MouseDevice, it would actually scroll
down rapidly.

Looking into it, the mouse delta for scrolling down was 1 and 15
for scrolling up. 15 is suspiciously -1 for a nibble.

According to https://isdaman.com/alsos/hardware/mouse/ps2interface.htm
the Z is actually 4 bits for Intellimouse.

This fixes scrolling up on virtualizers such as VirtualBox.


awesomekling:
Great catch :)

PR #3290, opened 2020-08-25
HackStudio+LibDebug: Add a disassembly and registers view, namespace LibDebug

Lubrsi:
Disassembly view:
qemu-system-i386_PwL5kU3Wcr
Nothing too fancy right now (e.g. disassembly breakpoints or pointing to the current instruction)

Registers view:
qemu-system-i386_iD6cxRRflP


awesomekling:
Very cool! And the more advanced features you mention will be interesting to see too :)

PR #3294, opened 2020-08-25
AK: Add Endian.h header to replace NetworkOrdered.h.

asynts:
We have this nice `NetworkOrdered<T>` type which stores the underlying type in big endian format and provides an implicit conversion to the host endianess.

This pull requests adds `LittleEndian<T>`, `BigEndian<T>` and makes `NetworkOrdered<T>` an alias for `BigEndian<T>`. I've also used the opportunity to rename `convert_between_host_and_network` to `convert_between_host_and_network_endian`.

The implementation is moved to `Endian.h` which is more generic than `NetworkOrdered.h`.

I've also added overloads for the stream operators, making it possible to read/write them from/to streams.


awesomekling:
Nice idea :)

PR #3295, opened 2020-08-25
LibJS: Make Interpreter::throw_exception() a void function

linusg:
The motivation for this change is twofold:
    
- Returning a `JS::Value` is misleading as one would expect it to carry some meaningful information, like maybe the error object that's being created, but in fact it is always empty. Supposedly to serve as a shortcut for the common case of "throw and return empty value", but that's just leading us to my second point.
- Inconsistent usage / coding style: as of this commit there are 114 uses of `throw_exception()` discarding its return value and 55 uses directly returning the call result (in LibJS, not counting LibWeb); with the first style often having a more explicit empty value (or `nullptr` in some cases) return anyway. One more line to always make the return value obvious is should be worth it.
    
So now it's basically always these steps, which is already being used in the majority of cases (as outlined above):
    
- Throw an exception. This mutates interpreter state by updating `m_exception` and unwinding, but doesn't return anything.
- Let the caller explicitly return an empty value, `nullptr` or anything else itself.

### DIFF
--- a/Libraries/LibWeb/CodeGenerators/WrapperGenerator.cpp
+++ b/Libraries/LibWeb/CodeGenerators/WrapperGenerator.cpp
@@ -645,12 +645,13 @@ void generate_implementation(const IDL::Interface& interface)
         out() << "    if (!impl)";
         out() << "        return {};";
         if (function.length() > 0) {
-            out() << "    if (interpreter.argument_count() < " << function.length() << ")";
-
+            out() << "    if (interpreter.argument_count() < " << function.length() << ") {";
             if (function.length() == 1)
-                out() << "        return interpreter.throw_exception<JS::TypeError>(JS::ErrorType::BadArgCountOne, \"" << function.name << "\");";
+                out() << "        interpreter.throw_exception<JS::TypeError>(JS::ErrorType::BadArgCountOne, \"" << function.name << "\");";
             else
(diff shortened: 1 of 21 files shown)


Libraries/LibWeb/CodeGenerators/WrapperGenerator.cpp
```diff
@@ -645,12 +645,12 @@ void generate_implementation(const IDL::Interface& interface)
...
-            out() << "    if (interpreter.argument_count() < " << function.length() << ")";
-
+            out() << "    if (interpreter.argument_count() < " << function.length() << ") {";
             if (function.length() == 1)
-                out() << "        return interpreter.throw_exception<JS::TypeError>(JS::ErrorType::BadArgCountOne, \"" << function.name << "\");";
+                out() << "        interpreter.throw_exception<JS::TypeError>(JS::ErrorType::BadArgCountOne, \"" << function.name << "\");";
             else
-                out() << "        return interpreter.throw_exception<JS::TypeError>(JS::ErrorType::BadArgCountMany, \"" << function.name << "\", \"" << function.length() << "\");";
+                out() << "        interpreter.throw_exception<JS::TypeError>(JS::ErrorType::BadArgCountMany, \"" << function.name << "\", \"" << function.length() << "\");";
+            out() << "    }";
```
awesomekling, line 653: This seems wrong. Now we will no longer return immediately after throwing an exception.

linusg: Good catch, I was meaning to put a `return {};` before the closing `}`. Fixed :+1:

PR #3296, opened 2020-08-25
Strings

bugaevc:
This is the result of the discussion in https://github.com/SerenityOS/serenity/pull/3275

* Port a bunch of code from `static char buffer[SOME_STATIC_SIZE]` to `static String buffer`
* Add `String::copy_characters_to_buffer()`

    This is a `strcpy()`-like method with actually sane semantics:

    * It accepts a non-empty buffer along with its size in bytes.
    * It copies as much of the string as fits into the buffer.
    * It always null-terminates the result.
    * It returns, as a non-discardable boolean, whether the whole string has been copied.

    Intended usage looks like this:

    ```cpp
    bool fits = string.copy_characters_to_buffer(buffer, sizeof(buffer));
    ```

    and then either

    ```cpp
    if (!fits) {
        fprintf(stderr, "The name does not fit!!11");
        return nullptr;
    }
    ```

    or, if you're sure the buffer is large enough,

    ```cpp
    // I'm totally sure it fits because [reasons go here].
    ASSERT(fits);
    ```

    or if you're feeling extremely adventurous,

    ```cpp
    (void)fits;
    ```

    but don't do that, please.

* Port most of code that used `str*cpy()` to `String::copy_characters_to_buffer()`, handling truncation in one of the ways outlined above. (@BenWiederhake, I'm sorry, but this removes most of the usages of `strlcpy()` that you've added.)

* Deprecate `strcpy()`, `strncpy()`, `strcat()` and `strncat()`, and mark `strlcpy()` and `strlcat()` with `__attribute__((warn_unused_result))`. Since our code is warning-free, this ensures we never misuse those functions.

  Unfortunately, there is a couple of places where we still have to unsafely use those functions; so turn off the warnings for there specifically.

* Misc improvements.

cc @BenWiederhake because you're clearly interested in this, @alimpfard (please review the shit out of me!), and @rthugh02 for netdb changes.

Also, [`passwd` the Userland executable](https://github.com/SerenityOS/serenity/blob/master/Userland/passwd.cpp) doesn't work for me (with these changes, haven't tested without them) — it generates a corrupted `/etc/passwd`. It would seem that's because it keeps a shallow copy of `struct passwd` across `getpwent()` calls (and expects it to keep the same value), which obviously wrong. But maybe there's something else going on, because the resulting file looks even more broken than I would expect it to be from *that*. Are my `passwd` changes playing a role here? cc @Petelliott for that.

image

### DIFF
--- a/AK/String.cpp
+++ b/AK/String.cpp
@@ -106,6 +106,18 @@ String String::empty()
     return StringImpl::the_empty_stringimpl();
 }
 
+bool String::copy_characters_to_buffer(char* buffer, size_t buffer_size) const
+{
+    // We must fit at least the NUL-terminator.
+    ASSERT(buffer_size > 0);
+
+    size_t characters_to_copy = min(length(), buffer_size - 1);
+    __builtin_memcpy(buffer, characters(), characters_to_copy);
+    buffer[characters_to_copy] = 0;
+
+    return characters_to_copy == length();
+}
+
 String String::isolated_copy() const
 {
     if (!m_impl)
--- a/Libraries/LibC/grp.cpp
+++ b/Libraries/LibC/grp.cpp
@@ -26,54 +26,51 @@
 
 #include <AK/String.h>
 #include <AK/Vector.h>
+#include <errno_numbers.h>
 #include <grp.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
-#include <sys/mman.h>
 #include <unistd.h>
 
 extern "C" {
 
-#define GRDB_STR_MAX_LEN 256
+static FILE* s_stream = nullptr;
+static unsigned s_line_number = 0;
+static struct group s_group;
 
-struct group_with_strings : public group {
-    char name_buffer[GRDB_STR_MAX_LEN];
-    char passwd_buffer[GRDB_STR_MAX_LEN];
-    char* members[32];
-    char members_buffer[32][32];
-};
-
-static FILE* __grdb_stream = nullptr;
-static unsigned __grdb_line_number = 0;
-static struct group_with_strings* __grdb_entry = nullptr;
+static String s_name;
+static String s_passwd;
+static Vector<String> s_members;
+static Vector<const char*> s_members_ptrs;
 
 void setgrent()
 {
-    __grdb_line_number = 0;
-    if (__grdb_stream) {
-        rewind(__grdb_stream);
+    s_line_number = 0;
+    if (s_stream) {
+        rewind(s_stream);
     } else {
-        __grdb_stream = fopen("/etc/group", "r");
-        if (!__grdb_stream) {
+        s_stream = fopen("/etc/group", "r");
+        if (!s_stream) {
             perror("open /etc/group");
         }
-        assert(__grdb_stream);
-        __grdb_entry = (struct group_with_strings*)mmap_with_name(nullptr, getpagesize(), PROT_READ | PROT_WRITE, MAP_ANONYMOUS | MAP_PRIVATE, 0, 0, "setgrent");
     }
 }
 
 void endgrent()
 {
-    __grdb_line_number = 0;
-    if (__grdb_stream) {
-        fclose(__grdb_stream);
-        __grdb_stream = nullptr;
+    s_line_number = 0;
+    if (s_stream) {
+        fclose(s_stream);
+        s_stream = nullptr;
     }
-    if (__grdb_entry) {
-        munmap(__grdb_entry, getpagesize());
-        __grdb_entry = nullptr;
-    }
+
+    memset(&s_group, 0, sizeof(s_group));
+
+    s_name = {};
+    s_passwd = {};
+    s_members = {};
+    s_members_ptrs = {};
 }
 
 struct group* getgrgid(gid_t gid)
@@ -96,52 +93,69 @@ struct group* getgrnam(const char* name)
     return nullptr;
 }
 
+static bool parse_grpdb_entry(const String& line)
+{
+    auto parts = line.split_view(':', true);
+    if (parts.size() != 4) {
+        fprintf(stderr, "getgrent(): Malformed entry on line %u: '%s' has %zu parts\n", s_line_number, line.characters(), parts.size());
+        return false;
+    }
+
+    s_name = parts[0];
+    s_passwd = parts[1];
+
+    auto& gid_string = parts[2];
+    String members_string = parts[3];
+
+    auto gid = gid_string.to_uint();
+    if (!gid.has_value()) {
+        fprintf(stderr, "getgrent(): Malformed GID on line %u\n", s_line_number);
+        return false;
+    }
+
+    s_members = members_string.split(',');
+    s_members_ptrs.clear_with_capacity();
+    s_members_ptrs.ensure_capacity(s_members.size() + 1);
+    for (auto& member : s_members) {
+        s_members_ptrs.append(member.characters());
+    }
+    s_members_ptrs.append(nullptr);
+
+    s_group.gr_gid = gid.value();
+    s_group.gr_name = const_cast<char*>(s_name.characters());
+    s_group.gr_passwd = const_cast<char*>(s_passwd.characters());
+    s_group.gr_mem = const_cast<char**>(s_members_ptrs.data());
+
+    return true;
+}
+
 struct group* getgrent()
 {
-    if (!__grdb_stream)
+    if (!s_stream)
         setgrent();
 
-    assert(__grdb_stream);
-    if (feof(__grdb_stream))
-        return nullptr;
+    while (true) {
+        if (!s_stream || feof(s_stream))
+            return nullptr;
 
-next_entry:
-    char buffer[1024];
-    ++__grdb_line_number;
-    char* s = fgets(buffer, sizeof(buffer), __grdb_stream);
-    if (!s)
-        return nullptr;
-    assert(__grdb_stream);
-    if (feof(__grdb_stream))
-        return nullptr;
-    String line(s, Chomp);
-    auto parts = line.split(':', true);
-    if (parts.size() != 4) {
-        fprintf(stderr, "getgrent(): Malformed entry on line %u: '%s' has %zu parts\n", __grdb_line_number, line.characters(), parts.size());
-        goto next_entry;
+        if (ferror(s_stream)) {
+            fprintf(stderr, "getgrent(): Read error: %s\n", strerror(ferror(s_stream)));
+            return nullptr;
+        }
+
+        char buffer[1024];
+        ++s_line_number;
+        char* s = fgets(buffer, sizeof(buffer), s_stream);
+
+        // Silently tolerate an empty line at the end.
+        if ((!s || !s[0]) && feof(s_stream))
+            return nullptr;
+
+        String line(s, Chomp);
+        if (parse_grpdb_entry(line))
+            return &s_group;
+        // Otherwise, proceed to the next line.
     }
-    auto& e_name = parts[0];
-    auto& e_passwd = parts[1];
-    auto& e_gid_string = parts[2];
-    auto& e_members_string = parts[3];
-    auto e_gid = e_gid_string.to_uint();
-    if (!e_gid.has_value()) {
-        fprintf(stderr, "getgrent(): Malformed GID on line %u\n", __grdb_line_number);
-        goto next_entry;
-    }
-    auto members = e_members_string.split(',');
-    __grdb_entry->gr_gid = e_gid.value();
-    __grdb_entry->gr_name = __grdb_entry->name_buffer;
-    __grdb_entry->gr_passwd = __grdb_entry->passwd_buffer;
-    for (size_t i = 0; i < members.size(); ++i) {
-        __grdb_entry->members[i] = __grdb_entry->members_buffer[i];
-        strlcpy(__grdb_entry->members_buffer[i], members[i].characters(), sizeof(__grdb_entry->members_buffer[i]));
-    }
-    __grdb_entry->members[members.size()] = nullptr;
-    __grdb_entry->gr_mem = __grdb_entry->members;
-    strlcpy(__grdb_entry->name_buffer, e_name.characters(), GRDB_STR_MAX_LEN);
-    strlcpy(__grdb_entry->passwd_buffer, e_passwd.characters(), GRDB_STR_MAX_LEN);
-    return __grdb_entry;
 }
 
 int initgroups(const char* user, gid_t extra_gid)
--- a/Libraries/LibC/netdb.cpp
+++ b/Libraries/LibC/netdb.cpp
@@ -41,12 +41,10 @@ extern "C" {
 int h_errno;
 
 static hostent __gethostbyname_buffer;
-static char __gethostbyname_name_buffer[512];
 static in_addr_t __gethostbyname_address;
 static in_addr_t* __gethostbyname_address_list_buffer[2];
 
 static hostent __gethostbyaddr_buffer;
-static char __gethostbyaddr_name_buffer[512];
 static in_addr_t* __gethostbyaddr_address_list_buffer[2];
 
 //Get service entry buffers and file information for the getservent() family of functions
@@ -55,8 +53,8 @@ static const char* services_path = "/etc/services";
 
 static bool fill_getserv_buffers(char* line, ssize_t read);
 static servent __getserv_buffer;
-static char __getserv_name_buffer[512];
-static char __getserv_protocol_buffer[10];
+static String __getserv_name_buffer;
+static String __getserv_protocol_buffer;
 static int __getserv_port_buffer;
 static Vector<ByteBuffer> __getserv_alias_list_buffer;
 static Vector<char*> __getserv_alias_list;
@@ -69,7 +67,7 @@ static const char* protocols_path = "/etc/protocols";
 
 static bool fill_getproto_buffers(char* line, ssize_t read);
 static protoent __getproto_buffer;
-static char __getproto_name_buffer[512];
+static String __getproto_name_buffer;
 static Vector<ByteBuffer> __getproto_alias_list_buffer;
 static Vector<char*> __getproto_alias_list;
 static int __getproto_protocol_buffer;
@@ -98,12 +96,13 @@ static int connect_to_lookup_server()
 
 hostent* gethostbyname(const char* name)
 {
+    static String s_name_buffer;
+
     auto ipv4_address = IPv4Address::from_string(name);
+
     if (ipv4_address.has_value()) {
-        auto ip4_string = ipv4_address.value().to_string();
-        ASSERT(ip4_string.length() < sizeof(__gethostbyname_name_buffer));
-        strlcpy(__gethostbyname_name_buffer, ip4_string.characters(), sizeof(__gethostbyname_name_buffer));
-        __gethostbyname_buffer.h_name = __gethostbyname_name_buffer;
+        s_name_buffer = ipv4_address.value().to_string();
+        __gethostbyname_buffer.h_name = const_cast<char*>(s_name_buffer.characters());
         __gethostbyname_buffer.h_aliases = nullptr;
         __gethostbyname_buffer.h_addrtype = AF_INET;
         new (&__gethostbyname_address) IPv4Address(ipv4_address.value());
@@ -152,9 +151,8 @@ hostent* gethostbyname(const char* name)
     if (rc <= 0)
         return nullptr;
 
-    strlcpy(__gethostbyname_name_buffer, name, sizeof(__gethostbyaddr_name_buffer));
-
-    __gethostbyname_buffer.h_name = __gethostbyname_name_buffer;
+    s_name_buffer = name;
+    __gethostbyname_buffer.h_name = const_cast<char*>(s_name_buffer.characters());
     __gethostbyname_buffer.h_aliases = nullptr;
     __gethostbyname_buffer.h_addrtype = AF_INET;
     __gethostbyname_address_list_buffer[0] = &__gethostbyname_address;
@@ -167,6 +165,8 @@ hostent* gethostbyname(const char* name)
 
 hostent* gethostbyaddr(const void* addr, socklen_t addr_size, int type)
 {
+    static String s_name_buffer;
+
     if (type != AF_INET) {
         errno = EAFNOSUPPORT;
         return nullptr;
@@ -214,11 +214,8 @@ hostent* gethostbyaddr(const void* addr, socklen_t addr_size, int type)
     if (responses.is_empty())
         return nullptr;
 
-    auto& response = responses[0];
-
-    strlcpy(__gethostbyaddr_name_buffer, response.characters(), sizeof(__gethostbyaddr_name_buffer));
-
-    __gethostbyaddr_buffer.h_name = __gethostbyaddr_name_buffer;
+    s_name_buffer = responses[0];
+    __gethostbyaddr_buffer.h_name = const_cast<char*>(s_name_buffer.characters());
     __gethostbyaddr_buffer.h_aliases = nullptr;
     __gethostbyaddr_buffer.h_addrtype = AF_INET;
     // FIXME: Should we populate the hostent's address list here with a sockaddr_in for the provided host?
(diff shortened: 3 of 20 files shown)


Libraries/LibC/netdb.cpp
```diff
@@ -84,7 +84,9 @@ static int connect_to_lookup_server()
 
     sockaddr_un address;
     address.sun_family = AF_LOCAL;
-    strlcpy(address.sun_path, "/tmp/portal/lookup", sizeof(address.sun_path));
+    bool fits = String("/tmp/portal/lookup").copy_characters_to_buffer(address.sun_path, sizeof(address.sun_path));
+    // The (constant) address is short enough to always fit into the buffer.
+    ASSERT(fits);
```
alimpfard, line 89: This is kinda weird, why not a compiletime assert?
Probably something like
```c++
constexpr static auto lookup_path = "/tmp/portal/lookup";
static_assert(__builtin_strlen(lookup_path) < sizeof(declval<sockaddr_un>().sun_path),
    "sockaddr_un::sun_path not large enough, or lookup_path invalid?");
__builtin_memcpy(address.sun_path, lookup_path, __builtin_strlen(lookup_path));
```
, even as a template inline function somewhere so it could be reused?

Userland/ping.cpp
```diff
@@ -125,7 +125,10 @@ int main(int argc, char** argv)
         ping_packet.header.code = 0;
         ping_packet.header.un.echo.id = htons(pid);
         ping_packet.header.un.echo.sequence = htons(seq++);
-        strlcpy(ping_packet.msg, "Hello there!\n", sizeof(ping_packet.msg));
+
+        bool fits = String("Hello there!\n").copy_characters_to_buffer(ping_packet.msg, sizeof(ping_packet.msg));
+        // It's a constant string, we can be sure that it fits.
+        ASSERT(fits);
```
alimpfard, line 131: Same thing as above here?

Libraries/LibC/netdb.cpp
```diff
@@ -359,24 +364,28 @@ void endservent()
...
-//Fill the service entry buffer with the information contained in the currently read line, returns true if successfull, false if failure occurs.
-static bool fill_getserv_buffers(char* line, ssize_t read)
+// Fill the service entry buffer with the information contained
+// in the currently read line, returns true if successfull,
+// false if failure occurs.
+static bool fill_getserv_buffers(const char* line, ssize_t read)
 {
     //Splitting the line by tab delimiter and filling the servent buffers name, port, and protocol members.
     String string_line = String(line, read);
     string_line.replace(" ", "\t", true);
```
alimpfard, line 374: Is this inplace?

Libraries/LibC/netdb.cpp
```diff
@@ -385,15 +394,15 @@ static bool fill_getserv_buffers(char* line, ssize_t read)
 
     __getserv_port_buffer = number.value();
 
-    //Removing any annoying whitespace at the end of the protocol.
+    // Remove any annoying whitespace at the end of the protocol.
     port_protocol_split[1].replace(" ", "", true);
     port_protocol_split[1].replace("\t", "", true);
     port_protocol_split[1].replace("\n", "", true);
```
alimpfard, line 400: > Is this inplace?


This is weird, newlines can't be here, spaces are replaced with tabs, and the line is split by tabs to get this `port_protocol_split`.

Libraries/LibC/netdb.cpp
```diff
@@ -459,11 +468,11 @@ struct protoent* getprotoent()
...
 
-    __getproto_alias_list.clear();
-
-    for (auto& alias : __getproto_alias_list_buffer) {
-        __getproto_alias_list.append((char*)alias.data());
-    }
+    __getproto_alias_list.clear_with_capacity();
+    __getproto_alias_list.ensure_capacity(__getproto_alias_list_buffer.size() + 1);
+    for (auto& alias : __getproto_alias_list_buffer)
+        __getproto_alias_list.append(reinterpret_cast<char*>(alias.data()));
```
alimpfard, line 474: Same `unchecked_append()` suggestion?

alimpfard:
> because the resulting file looks even more broken than I would expect it to be from that

If you mean the big, weird-looking blob of b64 data, I think that's scheduled to be moved out to some other file, but it's not corrupted or anything, it's just a hash with some prefix

Libraries/LibC/grp.cpp
```diff
@@ -96,52 +93,69 @@ struct group* getgrnam(const char* name)
...
-    auto e_gid = e_gid_string.to_uint();
-    if (!e_gid.has_value()) {
-        fprintf(stderr, "getgrent(): Malformed GID on line %u\n", __grdb_line_number);
-        goto next_entry;
+
+    s_name = parts[0];
+    s_passwd = parts[1];
+
+    auto& gid_string = parts[2];
+    String members_string = parts[3];
```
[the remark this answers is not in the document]
bugaevc: The whole point of using `split_view()` instead of `split()` (which would have returned a `Vector<` of `String`s`>`) is to only copy those strings I need copied, while keeping others (namely, just `gid_string`, it seems?) non-copied (borrowed from `line`).

That being said, we shouldn't really need `members_string` itself to be a `String` here, if there was a `Vector<String> StringView::split()`. But IIRC there's no such method.

Libraries/LibC/unistd.cpp
```diff
@@ -552,13 +552,13 @@ int set_process_icon(int icon_id)
 
 char* getlogin()
 {
-    static char __getlogin_buffer[256];
-    if (auto* passwd = getpwuid(getuid())) {
-        strlcpy(__getlogin_buffer, passwd->pw_name, sizeof(__getlogin_buffer));
+    static String buffer;
+    if (buffer.is_null()) {
+        if (auto* passwd = getpwuid(getuid())) {
+            buffer = String(passwd->pw_name);
```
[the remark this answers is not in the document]
bugaevc: Yeah, the implicit conversion should work here too, I just wanted to make it explicit that I'm copying out the C string to the buffer as a `String`. Then you should of course ask why I haven't done this elsewhere — dunno, just felt like that.

AK/String.cpp
```diff
@@ -106,6 +106,18 @@ String String::empty()
     return StringImpl::the_empty_stringimpl();
 }
 
+bool String::copy_characters_to_buffer(char* buffer, size_t buffer_size) const
+{
+    // We must fit at least the NUL-terminator.
+    ASSERT(buffer_size > 0);
+
+    size_t characters_to_copy = min(length(), buffer_size - 1);
+    memcpy(buffer, characters(), characters_to_copy);
```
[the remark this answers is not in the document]
bugaevc: Great question!

This wouldn't be the first time, even in `String.cpp`:

```cpp
$ grep memcpy -r AK
AK/MACAddress.h:        __builtin_memcpy(m_data, data, 6);
AK/BufferStream.h:        __builtin_memcpy(raw_data, m_buffer.data() + m_offset, size);
AK/Bitmap.h:            __builtin_memcpy(m_data, previous_data, previous_size_bytes);
AK/StringImpl.cpp:    memcpy(buffer, cstring, length * sizeof(char));
AK/Platform.h:    memcpy(tmp, path, path_length);
AK/Stream.h:        __builtin_memcpy(bytes.data(), m_bytes.data() + m_offset, count);
AK/Stream.h:        __builtin_memcpy(bytes.data(), m_bytes.data() + m_offset, bytes.size());
AK/StringBuilder.cpp:    memcpy(m_buffer.data() + m_length, str.characters_without_null_termination(), str.length());
AK/LogStream.h:            __builtin_memcpy(new_data, u.m_local_buffer, m_size);
AK/LogStream.h:            __builtin_memcpy(new_data, u.m_buffer, m_size);
AK/LogStream.h:        __builtin_memcpy(data() + m_size, str, len);
AK/ByteBuffer.h:        __builtin_memcpy(this->data() + old_size, data, data_size);
AK/ByteBuffer.h:        __builtin_memcpy(this->data() + offset, data, data_size);
AK/ByteBuffer.h:    __builtin_memcpy(m_data, data, size);
AK/ByteBuffer.h:    __builtin_memcpy(new_data, m_data, m_size);
AK/String.cpp:    memcpy(buffer, characters(), characters_to_copy);
AK/String.cpp:    memcpy(buffer, m_impl->characters(), m_impl->length());
```

I could switch this to `__builtin_memcpy()` just to be sure though.

Libraries/LibC/grp.cpp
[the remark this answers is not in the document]
alimpfard: > The whole point [...] is to only copy those strings I need copied

Right, I agree with that, I'm just talking about this specific `members_string`.

>  if there was a `Vector<String> StringView::split()`

There's a  `Vector<String> StringView::split_view()`, which is why I was talking about `Vector<StringView>` -> `Vector<String>`

Libraries/LibC/netdb.cpp
bugaevc: :thinking: perhaps even

```cpp
address = {
    .sun_path = "/tmp/portal/lookup"
};
```

Lemme see if that works.

Libraries/LibC/unistd.cpp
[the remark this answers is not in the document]
alimpfard: Works for me lol

Libraries/LibC/netdb.cpp
bugaevc: :shrug: I haven't touched this logic

Libraries/LibC/netdb.cpp
bugaevc: Which other `unchecked_append()` suggestion you're referring to? :eyes:

But yeah, let me do that.

Libraries/LibC/netdb.cpp
alimpfard: I doubt it'd be happy with that...

Libraries/LibC/netdb.cpp
alimpfard: huh. the other one got lost on the way.

bugaevc:
> If you mean the big, weird-looking blob of b64 data, I think that's scheduled to be moved out to some other file, but it's not corrupted or anything, it's just a hash with some prefix

No, I mean that the line should go something like

`nona:$smth$:200:200:groups:/home/nona:/bin/sh`

and not

`/bin/sh:$smth$:200:200:nona:!:/home/nona`

Libraries/LibC/netdb.cpp
```diff
@@ -278,10 +278,11 @@ struct servent* getservent()
...
     __getserv_buffer.s_proto = const_cast<char*>(__getserv_protocol_buffer.characters());
 
-    __getserv_alias_list.clear();
-    for (auto& alias : __getserv_alias_list_buffer) {
-        __getserv_alias_list.append((char*)alias.data());
-    }
+    __getserv_alias_list.clear_with_capacity();
+    __getserv_alias_list.ensure_capacity(__getserv_alias_list_buffer.size() + 1);
+    for (auto& alias : __getserv_alias_list_buffer)
+        __getserv_alias_list.append(reinterpret_cast<char*>(alias.data()));
```
alimpfard, line 284: This other `unchecked_append()` one

Libraries/LibC/grp.cpp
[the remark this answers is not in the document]
bugaevc: Ah, I see

alimpfard:
> `/bin/sh:$smth$:200:200:nona:!:/home/nona`

Ah, I see.

Libraries/LibC/netdb.cpp
bugaevc: It works!

```cpp
../Libraries/LibC/netdb.cpp: In function 'int connect_to_lookup_server()':
../Libraries/LibC/netdb.cpp:87:9: error: initializer-string for 'char [108]' is too long [-fpermissive]
   87 |         "/tmp/portal/lookup/xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx"
      |         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
ninja: build stopped: subcommand failed.
```

alimpfard: Noice!

bugaevc: I don't seem to be able to convince the compiler about the ping buffer thing though...

`error: C99 designator 'msg' outside aggregate initializer`

Dang it, C++.

So I'll leave that one as an assertion.

alimpfard: Perhaps the ping message class has a ctor?
Ah well. 1/2 is still good; though I'll push a little more for a compiletime array copy thingy

BenWiederhake:
Don't be sorry, I'm happy that the "uhh unexpected truncation is better than crashing, right?" gives way to actual error handling! :)

AK/String.cpp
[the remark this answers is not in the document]
BenWiederhake: If you look at the assembly output by gcc, `__builtin_strlen` usually becomes just `strlen`. I expect the same to happen with `__builtin_memcpy`, so I personally don't quite see the difference. (Other than that `__builtin_memcpy` probably breaks with other compilers.)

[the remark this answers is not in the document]
alimpfard: In such cases, why don't we move the implementation of memcpy and co into AK? I've already moved memmem, so it wouldn't be a first~

Libraries/LibC/grp.cpp
```diff
@@ -26,54 +26,51 @@
...
+    s_line_number = 0;
+    if (s_stream) {
+        fclose(s_stream);
+        s_stream = nullptr;
     }
+
+    s_name = {};
+    s_passwd = {};
+    s_members = {};
+    s_members_ptrs = {};
```
BenWiederhake, line 71: Nit: First clear the pointers, then the data pointed-to. In this case it doesn't matter though, as nothing could possibly happen in the "meantime", except maybe a debugger getting confused.

bugaevc: Yeah I'm this case I'm clearing out the whole thing, so it doesn't matter.

Libraries/LibC/netdb.cpp
```diff
@@ -468,7 +454,7 @@ struct protoent* getprotoent()
     if (!fill_getproto_buffers(line, read))
         return nullptr;
 
-    __getproto_buffer.p_name = __getproto_name_buffer;
+    __getproto_buffer.p_name = const_cast<char*>(__getproto_name_buffer.characters());
```
BenWiederhake, line 467: These const casts everywhere are somewhat unfortunate.

For this commit, adding them is fine: the API requires returning/offering a `char*`.

But maybe in another commit/PR, we could "lock down" the API (here and everywhere else) and mark the fields in the structs as `const char*`s? I hope this only produces a warning in ports that copy the point into a `char* foo;`, and not an error. And within Serenity, this would remove a lot of `const-cast`s.

bugaevc: It is unfortunate, but that's Unix... The API isn't beautiful in other aspects either.

I... don't know, I think I'd rather `const_cast<>` it and keep the API as-is.

In C++, it's surely an error to implicitly cast a `const char` to `char*`. Do we have any C++ ports? But then again, they're ports, it's fine to patch them if that's required to make them build/work...

Libraries/LibC/unistd.cpp
```diff
@@ -552,13 +552,13 @@ int set_process_icon(int icon_id)
...
+        }
         endpwent();
-        return __getlogin_buffer;
```
BenWiederhake, line 559: Before, this function "usually" returned a string.
Now, this function *always* returns nullptr.

Could it be there's a missing `return buffer.characters();` or something?

BenWiederhake (changes requested):

Libraries/LibC/unistd.cpp
bugaevc: Yep, thanks for spotting it!

Libraries/LibC/grp.cpp
bugaevc: Switched the two around

Userland/passwd.cpp
```diff
@@ -132,7 +132,7 @@ int main(int argc, char** argv)
     } else {
         auto new_password = Core::get_password("New password: ");
         if (new_password.is_error()) {
-            fprintf(stderr, strerror(new_password.error()));
+            fprintf(stderr, "%s\n", strerror(new_password.error()));
```
BenWiederhake, line 135: Good catch!

alimpfard: mmm yes, I too make exploits around compromised `strerror` values (is that even possible?)

Petelliott:
> Are my `passwd` changes playing a role here?

Kind of. my implementation was incorrect. your changes to `pwd.cpp` just brings them to the forefront, because I was accessing freed memory.

I don't have time to look in to this today, but I might in a few days.

BenWiederhake (changes requested):

AK/String.cpp
```diff
@@ -106,6 +106,18 @@ String String::empty()
     return StringImpl::the_empty_stringimpl();
 }
 
+bool String::copy_characters_to_buffer(char* buffer, size_t buffer_size) const
+{
+    // We must fit at least the NUL-terminator.
+    ASSERT(buffer_size > 0);
```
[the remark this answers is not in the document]
bugaevc: Because I wanted to guarantee that even when it `return`s `false`, the buffer is left in a consistent state — namely, it contains the truncated but properly null-terminated string. So you can still pass it to somewhere where a C string is expected and have it "work" (as much as it can given the string got truncated). In this light, the "yet another case that one has to consider" is "what if the string ended up not being null-terminated because the buffer was empty in the first place", and I'd rather we catch that straight away.

That being said, your logic makes sense too. If you care enough, we can switch it around.

> Cases where this might remotely make sense: We're assembling a larger buffer out of "stuff", and somehow have exactly 0 bytes left, which so far is not yet a problem.

No, if you're assembling a buffer (so basically, catting together some strings), you're going to want to overwrite last part's 0-terminator each time, so that the strings are directly adjacent. So you always have at least one byte of space.

BenWiederhake:
Sorry for the conflict ^^'

nico:
17109a3a3167fb1b39a6878b23519331020ac1f8 regressed the time to run `disasm /bin/id` by over 50% (~500ms -> ~780ms on my system). It likely slowed down a bunch of other programs as well. More safety is nice, but 50% is quite a bit -- could that be optimized?

bugaevc:
That's unexpected, we should be actually doing less of copying/allocating now 🤔

I'll look into it, thanks

awesomekling:
This also broke ports:

```
Configuring binutils!
+ ./configure --host=i686-pc-serenity --target=i686-pc-serenity --with-sysroot=/ --with-build-sysroot=/home/kling/src/serenity/Build/Root --disable-werror --disable-gdb --disable-nls
checking build system type... x86_64-pc-linux-gnu
checking host system type... i686-pc-serenity
checking target system type... i686-pc-serenity
checking for a BSD-compatible install... /usr/bin/install -c
checking whether ln works... yes
checking whether ln -s works... yes
checking for a sed that does not truncate output... /usr/bin/sed
checking for gawk... gawk
checking for i686-pc-serenity-gcc... i686-pc-serenity-gcc
checking whether the C compiler works... no
configure: error: in `/home/kling/src/serenity/Ports/binutils/binutils-2.32':
configure: error: C compiler cannot create executables
See `config.log' for more details
```

From config.log:
```
configure:4326: checking whether the C compiler works
configure:4348: i686-pc-serenity-gcc    conftest.c  >&5 
/home/kling/src/serenity/Toolchain/Local/lib/gcc/i686-pc-serenity/10.1.0/../../../../i686-pc-serenity/bin/ld: /home/kling/src/serenity/Build/Root/usr/lib/libc.a(unistd.cpp.o): in function `getlogin':
/home/kling/src/serenity/Build/../Libraries/LibC/unistd.cpp:560: undefined reference to `__cxa_guard_acquire'
/home/kling/src/serenity/Toolchain/Local/lib/gcc/i686-pc-serenity/10.1.0/../../../../i686-pc-serenity/bin/ld: /home/kling/src/serenity/Build/../Libraries/LibC/unistd.cpp:560: undefined reference to `__cxa_guard_release'
collect2: error: ld returned 1 exit status
```

bugaevc:
@nico oh, you're talking about `disasm /bin/id`, not `/bin/id` itself!

Then I cannot imagine how it could possibly affect `disasm`, since I very much doubt `disasm` calls into `pwd.h`, `grp.h`, `netdb.h` or `getlogin()` (nor do they show up in a profile), and that's the only files/functions that commit changes.

It could be that https://github.com/SerenityOS/serenity/commit/17109a3a3167fb1b39a6878b23519331020ac1f8 has somehow resulted in `/bin/id` having more instructions (or instructions that are harder for `disasm` to parse?), but that sounds unlikely too.

bugaevc:
@awesomekling 

<blockquote><pre>/home/kling/src/serenity/Toolchain/Local/lib/gcc/i686-pc-serenity/10.1.0/../../../../i686-pc-serenity/bin/ld: /home/kling/src/serenity/Build/Root/usr/lib/libc.a(unistd.cpp.o): in function `getlogin':
/home/kling/src/serenity/Build/../Libraries/LibC/unistd.cpp:560: undefined reference to `__cxa_guard_acquire'
/home/kling/src/serenity/Toolchain/Local/lib/gcc/i686-pc-serenity/10.1.0/../../../../i686-pc-serenity/bin/ld: /home/kling/src/serenity/Build/../Libraries/LibC/unistd.cpp:560: undefined reference to `__cxa_guard_release'</pre></blockquote>

So apparently GCC decided to insert the complicated synchronized static initialization thing for a simple `static String buffer;` (where `String`'s constructor simply sets one pointer to null, so it could have just allocated a single pointer in `.bss` and be done with it). And then for whatever reason that works when building the system normally, but not for ports.

Moving the static out of the function fixes it:

```diff
diff --git a/Libraries/LibC/unistd.cpp b/Libraries/LibC/unistd.cpp
index 94ccc2697..22a885675 100644
--- a/Libraries/LibC/unistd.cpp
+++ b/Libraries/LibC/unistd.cpp
@@ -555,16 +555,17 @@ int set_process_icon(int icon_id)
     __RETURN_WITH_ERRNO(rc, rc, -1);
 }
 
+static String getlogin_buffer;
+
 char* getlogin()
 {
-    static String buffer;
-    if (buffer.is_null()) {
+    if (getlogin_buffer.is_null()) {
         if (auto* passwd = getpwuid(getuid())) {
-            buffer = String(passwd->pw_name);
+            getlogin_buffer = String(passwd->pw_name);
         }
         endpwent();
     }
-    return const_cast<char*>(buffer.characters());
+    return const_cast<char*>(getlogin_buffer.characters());
 }
 
 int ftruncate(int fd, off_t length)
```

Perhaps just as interestingly, other in-function `static String`s that commit adds don't seem to cause any problems...

nico:
> Then I cannot imagine how it could possibly affect `disasm`, since I very much doubt `disasm` calls into `pwd.h`, `grp.h`, `netdb.h` or `getlogin()` (nor do they show up in a profile), and that's the only files/functions that commit changes.

I'm surprised too, but it's fairly reproducible (I ran `disasm /bin/id` 5 times and looked at the dbg log output to get the runtime and took the min of the 5 times, but they were within 10% of each other each time, and a lot higher after this change) and `git bisect` says it's that commit.

bugaevc:
Could you test new `disas` on old `id` and vice versa? To confirm it didn't actually make `disas` slower on the same binary.

nico:
ProfileViewer shows pretty different profiles, so I doubt it's that.

nico:
Hey, 386c7201d8d8e6900bf256cf5df08c30b6d9b5bb fixed the slowdown too. Thanks, @awesomekling!

BenWiederhake:
We could write a linter script that looks for "forbidden symbols" (like `__cxa_guard_acquire` in LibC), and run the lint on Travis.

PR #3297, opened 2020-08-25
LibGUI: Make scrollbars keep scrolling by page while clicking the gutter

nico:
I'm happy about the "always keep the timer on" change, it makes things (imho) much easier to understand. My local branch where I didn't do this has much more subtle behavior.


awesomekling:
:star:

PR #3298, opened 2020-08-25
LibJS: Add a helper for calling JS::Function's with arguments

alimpfard:
The fact that a `MarkedValueList` had to be created and populated was just annoying, so here's an alternative.

cc @linusg

### DIFF
--- a/Libraries/LibJS/Interpreter.h
+++ b/Libraries/LibJS/Interpreter.h
@@ -83,6 +83,21 @@ public:
         return interpreter;
     }
 
+    template<typename... Args>
+    [[nodiscard]] ALWAYS_INLINE Value call(Function& function, Value this_value, Args... args)
+    {
+        // Are there any values in this argpack?
+        // args = [] -> if constexpr (false)
+        // args = [x, y, z] -> if constexpr ((void)x, true || ...)
+        if constexpr ((((void)args, true) || ...)) {
+            MarkedValueList arglist { heap() };
+            (..., arglist.append(move(args)));
+            return call(function, this_value, move(arglist));
+        }
+
+        return call(function, this_value);
+    }
+
     ~Interpreter();
 
     Value run(GlobalObject&, const Statement&, ArgumentVector = {}, ScopeType = ScopeType::Block);
@@ -118,7 +133,6 @@ public:
     void enter_scope(const ScopeNode&, ArgumentVector, ScopeType, GlobalObject&);
     void exit_scope(const ScopeNode&);
 
-    [[nodiscard]] Value call(Function&, Value this_value, Optional<MarkedValueList> arguments = {});
     Value construct(Function&, Function& new_target, Optional<MarkedValueList> arguments, GlobalObject&);
 
     CallFrame& push_call_frame()
@@ -215,6 +229,8 @@ public:
 private:
     Interpreter();
 
+    [[nodiscard]] Value call_internal(Function&, Value this_value, Optional<MarkedValueList>);
+
     Heap m_heap;
 
     Value m_last_value;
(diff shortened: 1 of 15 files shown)


Userland/test-web.cpp
```diff
@@ -360,7 +360,7 @@ JSFileResult TestRunner::run_file_test(const String& test_path)
             new_interpreter.run(new_interpreter.global_object(), *file_program.value());
 
             auto& before_initial_page_load = new_interpreter.get_variable("__BeforeInitialPageLoad__", new_interpreter.global_object()).as_function();
-            (void)new_interpreter.call(before_initial_page_load, JS::js_undefined());
```
alimpfard, line 363: tsk tsk tsk, why can `undefined` be cast to an `Optional<MarkedValueList>` to begin with?

alimpfard: nvm, am dumb.

Libraries/LibJS/Interpreter.h
```diff
@@ -83,6 +83,21 @@ class Interpreter : public Weakable<Interpreter> {
         return interpreter;
     }
 
+    template<typename... Args>
+    [[nodiscard]] Value invoke(Function& function, Value this_value, Args... args)
```
linusg, line 87: Can we maybe find a different name here or make this an overload of `call()` or something, I see confusion with `Object::invoke()` upon us - "invoke" is a specific thing in the spec: https://tc39.es/ecma262/#sec-invoke

https://github.com/SerenityOS/serenity/blob/bbe2d4a2d97ce7968b0985e4f99cdfc01fe5823e/Libraries/LibJS/Runtime/Object.cpp#L857-L868

Libraries/LibWeb/DOM/Node.cpp
```diff
@@ -132,10 +132,8 @@ void Node::dispatch_event(NonnullRefPtr<Event> event)
             dbg() << "calling event listener with this=" << this_value;
 #endif
             auto* event_wrapper = wrap(global_object, *event);
-            JS::MarkedValueList arguments(global_object.heap());
-            arguments.append(event_wrapper);
             auto& interpreter = document().interpreter();
-            (void)interpreter.call(function, this_value, move(arguments));
+            (void)interpreter.invoke(function, this_value, event_wrapper);
```
alimpfard, line 136: This might introduce a behaviour change, I do not know if `global_object.heap()` is the same as `document().interpreter().heap()`, but if it's not...what are you even doing here?

linusg (commented):
> The fact that a MarkedValueList had to be created and populated was just annoying, so here's an alternative.

Absolutely, this is a great improvement!

Libraries/LibJS/Interpreter.h
alimpfard: Hmmm...I got this name from `std::invoke`, but we can for sure name it `call()`

Libraries/LibWeb/DOM/Node.cpp
awesomekling: It's the same heap :)

awesomekling:
> > The fact that a MarkedValueList had to be created and populated was just annoying, so here's an alternative.
> 
> Absolutely, this is a great improvement!

Agreed! Super nice change :)

Libraries/LibJS/Runtime/ArrayPrototype.cpp
```diff
@@ -559,7 +548,7 @@ JS_DEFINE_NATIVE_FUNCTION(ArrayPrototype::reduce_right)
         arguments.append(Value(i));
         arguments.append(this_object);
```
alimpfard, line 549: Oops, didn't remove these

alimpfard: It was such a nice diff size too :(

PR #3299, opened 2020-08-25
LibGUI: Scrollbar painting tweaks

nico:
The middle commit is the one with the interesting behavior change.

I think it'd be nice if the gutter also had both a "hover" highlight and a "pressed" highlight, but I haven't yet found one I like.


awesomekling:
These are all great

PR #3301, opened 2020-08-25
Color picker improvements

LepkoQQ:
Here are some improvements to the color picker widget:
- current color is auto-selected and it's position marked in the custom color picker field
- the full color field was trying to represent a 3D HSV color vector in 2D space, so i replaced it with a 2D color field and a hue slider
- improved color selections in the color field, you can now drag out of the field and selection will stay constrained to the edge
- marked the selected color position with a black line with a white border for better visibility on all colors
- added a preview of the current and new colors for comparison
- added a spinbox for alpha channel when color picker has alpha enabled
- changed some sizes and margins

color-picker


awesomekling:
Awesome work :)

PR #3303, opened 2020-08-25
WindowServer+LibGfx: Move title bar button layout to WindowTheme

thankyouverycool:


Libraries/LibGfx/ClassicWindowTheme.h
```diff
@@ -26,8 +26,10 @@
 
 #pragma once
 
-#include <LibGfx/WindowTheme.h>
+#include <AK/NonnullOwnPtrVector.h>
 #include <LibGfx/Color.h>
+#include <LibGfx/WindowTheme.h>
+#include <Services/WindowServer/Button.h>
```
awesomekling, line 32: This isn't right. WindowServer is a separate program and LibGfx should not depend on it.
Let's find a way to represent the necessary data in a LibGfx data structure instead :)

thankyouverycool: Apologies for the brainletism. Second attempt :^)

awesomekling:
Perfect!

PR #3309, opened 2020-08-26
LibCompress: Make the DEFLATE implementation a true stream implemenation.

asynts:
In #3220 I've already tried to make this a stream implementation, but I think I took the wrong approach. It is possible to read a DEFLATE stream while only looking at 64KiB of data at a time, which this implementation now does. (Previously, I read an entire block at once which could be significantly more than 64KiB and required allocating buffers on the heap.)

There are still a few FIXMEs in this implementation but I want to do some other unrelated stuff first.

### DIFF
--- a/AK/Stream.h
+++ b/AK/Stream.h
@@ -31,6 +31,7 @@
 #include <AK/Endian.h>
 #include <AK/Forward.h>
 #include <AK/MemMem.h>
+#include <AK/Optional.h>
 #include <AK/Span.h>
 #include <AK/StdLibExtras.h>
 #include <AK/Vector.h>
@@ -101,6 +102,15 @@ template<typename T>
 InputStream& operator<<(InputStream& stream, BigEndian<T> value)
 {
     stream << static_cast<T>(value);
+    return stream;
+}
+
+template<typename T>
+InputStream& operator>>(InputStream& stream, Optional<T>& value)
+{
+    T temporary;
+    stream >> temporary;
+    value = temporary;
     return stream;
 }
 
--- a/Libraries/LibCompress/Deflate.h
+++ b/Libraries/LibCompress/Deflate.h
@@ -26,160 +26,90 @@
 
 #pragma once
 
-#include <AK/CircularQueue.h>
-#include <AK/Span.h>
-#include <AK/Stream.h>
-#include <AK/Types.h>
+#include <AK/BitStream.h>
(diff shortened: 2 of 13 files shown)


AK/Stream.h
```diff
@@ -104,6 +105,22 @@ InputStream& operator<<(InputStream& stream, BigEndian<T> value)
...
+    value = temporary;
+    return stream;
+}
+template<typename T>
+OutputStream& operator<<(OutputStream& stream, const Optional<T>& value)
+{
+    if (value.has_value())
+        return stream << value.value();
+    return stream;
+}
```
awesomekling, line 122: This feels like a very weird API. It may or may not write to the stream, depending of whether the optional is empty or not.

asynts: Initially, I had the same thought but then I dismissed it, but I have to agree it's pretty weird and I can't think of a use case. I've rebased and removed it.

Libraries/LibCompress/Deflate.h
```diff
@@ -26,160 +26,90 @@
...
+    State m_state { State::Idle };
+    union {
+        CompressedBlock m_compressed_block;
+        UncompressedBlock m_uncompressed_block;
+    };
 
-    mutable bool m_read_last_block { false };
-    mutable DuplexMemoryStream m_intermediate_stream;
+    InputBitStream m_input_stream;
+    CircularDuplexStream<64 * 1024> m_output_stream;
```
stelar7, line 112: I might be misunderstanding this piece of code, but the circular buffer used in DEFLATE should only be 32K

asynts: You are correct this should be 32KiB, no idea why I got 64KiB there.

Libraries/LibCompress/Deflate.h
```diff
@@ -26,160 +26,90 @@
...
+    State m_state { State::Idle };
+    union {
+        CompressedBlock m_compressed_block;
+        UncompressedBlock m_uncompressed_block;
+    };
 
-    mutable bool m_read_last_block { false };
-    mutable DuplexMemoryStream m_intermediate_stream;
+    InputBitStream m_input_stream;
+    CircularDuplexStream<32 * 1024> m_output_stream;
```
awesomekling, line 112: I guess the commit message should be updated too? :)

asynts: done.

PR #3312, opened 2020-08-26
Spreadsheet: *Spreadsheets + icons (and showing errors better)

alimpfard:
Now with as many sheets as you would like, deleting them from a live session isn't trivial though, so I left that out.

p.s. I like the new row buttons, much better!


Base/res/js/Spreadsheet/runtime.js
```diff
@@ -1,10 +1,8 @@
-const sheet = this;
-
 function range(start, end, columnStep, rowStep) {
     columnStep = integer(columnStep ?? 1);
     rowStep = integer(rowStep ?? 1);
-    start = sheet.parse_cell_name(start) ?? { column: "A", row: 0 };
-    end = sheet.parse_cell_name(end) ?? start;
+    start = global.parse_cell_name(start) ?? { column: "A", row: 0 };
+    end = global.parse_cell_name(end) ?? start;
```
alimpfard, line 5: Oh I forgot about these

alimpfard:
Figured I'd try making some icons:

16x16:
filetype-spreadsheet
app-spreadsheet

32x32:
app-spreadsheet
filetype-spreadsheet

I also swapped the extension to "sheets", because why not :shrug:

Applications/Spreadsheet/SpreadsheetView.cpp
```diff
@@ -94,6 +100,20 @@ SpreadsheetView::SpreadsheetView(Sheet& sheet)
     };
 }
 
+void SpreadsheetView::switched_out() {
```
awesomekling: c l a n g f o r m a t

Applications/Spreadsheet/SpreadsheetView.h
```diff
@@ -46,6 +46,9 @@ class SpreadsheetView final : public GUI::Widget {
     Function<void(const Position&, Cell&)> on_selection_changed;
     Function<void()> on_selection_dropped;
 
+    void switched_out();
+    void switched_in();
```
awesomekling, line 50: Instead of these, you can override `Widget::hide_event()` and `Widget::show_event()` which will get delivered when the `TabWidget` switches between tabs :)

Applications/Spreadsheet/SpreadsheetView.cpp
alimpfard: Huh.
That shouldn't have happened, I'm about 99% sure format-on-save is on

alimpfard: whatever it is, the latest of qtcreator borked the formatting, it either doesn't format on save, or it completely messes up the file.
turned that off again, I guess that's a no-go

Applications/Spreadsheet/SpreadsheetView.cpp
```diff
@@ -79,6 +79,12 @@ SpreadsheetView::SpreadsheetView(Sheet& sheet)
     };
 
     m_table_view->on_selection_change = [&] {
+        m_sheet->selected_cells().clear();
+        for (auto& index : m_table_view->selection().indexes()) {
+            Position position {m_sheet->column(index.column()), (size_t)index.row()};
```
awesomekling: c l a n g f o r m a t

alimpfard: w h y m e r g e i t t h e n aaaaaaaaaa

awesomekling: because it was fixed in the next commit aaaaaaaaaaaaaaa

alimpfard: nooooooo formatting segmentation bad

alimpfard: there should be something like `git clang-format-defrag` that takes each commit, formats it, then amends it with the formatted files

PR #3313, opened 2020-08-26
Nicer AK_MAKE_NONXXXABLE, remove dead data, document test

BenWiederhake:
Another mixed bag – I assume this is preferred these over three individual PRs.

### DIFF
--- a/AK/Noncopyable.h
+++ b/AK/Noncopyable.h
@@ -29,9 +29,9 @@
 #define AK_MAKE_NONCOPYABLE(c) \
 private:                       \
     c(const c&) = delete;      \
-    c& operator=(const c&) = delete;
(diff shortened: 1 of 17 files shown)


AK/Noncopyable.h
```diff
@@ -29,9 +29,9 @@
 #define AK_MAKE_NONCOPYABLE(c) \
 private:                       \
```
nico, line 30: Is the `private:` even needed now that these are `= delete`?

BenWiederhake: I didn't check whether any of the ~80 usages depends on that. My main motivation was to get rid of "double-semicolon" warnings I sometimes see with specific compilers/options. :)

PR #3314, opened 2020-08-26
Always build debug code on Travis

BenWiederhake:
Historically, many commits contained fixes to "rotten" debug code. This happens when, for example, the name of a function changes but noone updates the debug code.

This PR forces Travis to always build *ALL* the debug code.

The list of debug macros is manually generated like this:
```
grep -Prh '^#if.*DEBUG' AK/ Applications/ Demos/ DevTools/ Games/ Kernel/ \
     Libraries/ MenuApplets/ Services/ Shell/ Userland/ \
     | sort -u
```

I wasn't quite sure how to deal with already-enabled debug macros. I went for the `#ifndef #define` route, because leaving the definition out from the list in `CMakeLists.txt` feels error-prone.

EDIT: This intentionally skips over the `ACPI_DEBUG` debug macro, because I couldn't figure out how to fix that debug code. Maybe someone should have a look at that? :)

EDIT 2, because I can:

A picture of the "all the X" meme, but with the text "all the debug macros" written over it.


alimpfard:
:thinking: why not just add `test-js` to ctest as well? that would hide all the logs unless it fails _automagically_

BenWiederhake:
You mean `ctest` swallows output when `$? == 0`? Good idea, I'll look at `CMakeLists.txt` to figure out where to put it.

alimpfard:
Yes, that's how `test-crypto` isn't spamming builds with debug logs :P

BenWiederhake:
"Just" using ctest won't quite work: With all debug options on, `test-js` currently outputs 858081 lines on my machine. How does `ctest` store that amount of intermediate data, and can it cope with such insane amounts? Also, if `js-test` ever fails, would `ctest` then attempt to print all 858081 lines, thus overflowing Travis, erroring the build, and hiding the error message? That doesn't sound like a good idea to me.

How about `./test-js 2>&1 | tail -n 100`? This is plenty of space to list all possible failing tests, although the actual error information "remains" lost.

Alternatively, we could think about recompiling `test-js` without `-DALL_THE_DEBUG_MACROS`. However, I just tried this on my machine, and this seems to trigger some bugs in our CMakeLists, as not all libraries get recompiled(!). So taking this route would require some additional work.

Conclusion: I'm in favor of `./test-js 2>&1 | tail -n 100`. Do you have a better idea?

BenWiederhake:
Looks like it works as intended. I'm open for better ideas than just `| tail -n 100`.

alimpfard:
How about something radical like dumping the debug log into fd 3, and just `3>/dev/null` in the runner script?
we don't actually _care_ about the information itself, just that it builds, so get rid of the `dbg()` output?

BenWiederhake:
Oh well, in that case there's an even easier solution: `./test-js 2> /dev/null`

alimpfard:
> ./test-js 2> /dev/null

That would get rid of anything it prints to stderr though, does it do that?

BenWiederhake:
Thast's exactly what it does, because currently there is no way to distinguish `dbg()` output and manual `fprintf(stderr, "Hello!\n);`s.

alimpfard:
So what if we make `dbg()` just dump out to fd 3?

BenWiederhake:
We would need to distinguish the "usual" case where it still needs to go to `stderr`. (1? 2? I always forget.)

How to do that? Environment variable? Command-line argument? Neither of these seem right.

alimpfard:
Hmmm...we could check if fd 3 is open, and write into that if it is
otherwise write to stderr (that's fd 2, typically)

BenWiederhake:
That breaks if any Serenity app ever wants to use fd 3 for something else :P

How about this:
- `AK::DebugLogStream` gets a new static method `set_enabled(bool)` (to avoid the dependency on anything `environ`-related)
- `test-js` checks the environment first, and optionally disables all debug output.

Would that work?

alimpfard:
`dbgputstr` is an actual syscall inside serenity, this change only applies to lagom :)

alimpfard:
But a disabling mechanism is actually much better...

BenWiederhake:
There. It now does `DebugLogStream::set_enabled(false);` in `test-js.cpp` if the envvar `DISABLE_DBG_OUTPUT` is set. That sounds like a good solution to me, and preserves all test errors.

PR #3320, opened 2020-08-27
Add a bare-bones NTP client to Userland

nico:
Mind if I try to register serenityos.pool.ntp.org at https://manage.ntppool.org/manage/vendor ?

From what I understand, that's so that we can be blacklisted if we do something silly (but I might be wrong).

### DIFF
--- a/Userland/ntpquery.cpp
+++ b/Userland/ntpquery.cpp
@@ -1,0 +1,149 @@
+/*
+ * Copyright (c) 2020, Nico Weber <thakis@chromium.org>
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#define _GNU_SOURCE
+#include <LibCore/ArgsParser.h>
+#include <arpa/inet.h>
+#include <endian.h>
+#include <netdb.h>
+#include <netinet/in.h>
+#include <stdio.h>
+#include <string.h>
+#include <sys/socket.h>
+#include <sys/time.h>
+
+struct [[gnu::packed]] NtpPacket {
+    uint8_t li_vn_mode;
+    uint8_t stratum;
+    int8_t poll;
+    int8_t precision;
+
+    uint32_t root_delay;
+    uint32_t root_dispersion;
+    uint32_t reference_id;
+
+    uint64_t reference_timestamp;
+    uint64_t origin_timestamp;
+    uint64_t receive_timestamp;
+    uint64_t transmit_timestamp;
+};
+
+static_assert(sizeof(NtpPacket) == 48);
+
+int main(int argc, char** argv)
+{
+    if (pledge("stdio inet dns", nullptr) < 0) {
+        perror("pledge");
+        return 1;
+    }
+
+    // FIXME: Request serenityos.pool.ntp.org here https://manage.ntppool.org/manage/vendor
+    // and then use that as the default value.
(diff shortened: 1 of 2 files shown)


awesomekling:
Go for it :)

nico:
Cool, I submitted an application :) They say it might take a week to even start processing it, so I don't think we need to block merging this (and potential follow-ups) on the application getting processed.

Userland/ntpquery.cpp
```diff
@@ -0,0 +1,136 @@
...
+    Core::ArgsParser args_parser;
+    args_parser.add_positional_argument(host, "NTP server", "host", Core::ArgsParser::Required::No);
+    args_parser.parse(argc, argv);
+
+    if (!host)
+        return 1;
+
+    auto* hostent = gethostbyname(host);
+    if (!hostent) {
+        printf("Lookup failed for '%s'\n", host);
```
bugaevc, line 122: stderr?

Userland/ntpquery.cpp
```diff
@@ -0,0 +1,136 @@
...
+    // and then use that as the default value.
+    // Until then, explicitly pass this as `ntpquery pool.ntp.org`.
+    const char* host = nullptr;
+
+    Core::ArgsParser args_parser;
+    args_parser.add_positional_argument(host, "NTP server", "host", Core::ArgsParser::Required::No);
+    args_parser.parse(argc, argv);
+
+    if (!host)
+        return 1;
```
bugaevc, line 118: Not cool, make it required then

Userland/ntpquery.cpp
```diff
@@ -0,0 +1,136 @@
...
+        perror("sendto");
+        return 1;
+    }
+
+    socklen_t peer_address_size = sizeof(peer_address);
+    if (recvfrom(fd, &packet, sizeof(packet), 0, (struct sockaddr*)&peer_address, &peer_address_size) < 0) {
+        perror("recvfrom");
+        return 1;
+    }
+
```
bugaevc, line 171: Don't we want to check here that the packet we've just recv'ed indeed comes from the peer we've sent the request to?
A connectionless protocol and all that, we can't be sure we're actually receiving a reply to the request, only that we've received a datagram sent to the same port...

Also, this should do a length check: `recvfrom()` returns a size and you have to compare that against `sizeof(packet)`

alimpfard: Why not use a `Core::UDPServer`?
that'd have a nice secondary effect of making it easier to use SystemServer to spawn it...once we have IP socket takeover?

bugaevc: Can `Core::UDPServer` be used with a preexisting socket?

Besides, if I understand @nico's intent here, it's meant to be a one-shot Userland utility for querying the time once, not a persistent service for timesync. And even if it was, I don't think socket takeover would be appropriate here — we're not listening for packets sent to us by clients (on a well-known port), we're waiting for a reply to our own request (on a random port).

alimpfard: I doubt it can, but adding a ctor shouldn't be that difficult, I don't think?

> we're not listening for packets sent to us by clients (on a well-known port), we're waiting for a reply to our own request (on a random port).

Oh that's right; guess I wasn't thinking clearly there~

Userland/ntpquery.cpp
nico: See the FIXME 10 lines further up.

bugaevc: Yes, but while we have no default value, and have to explicitly pass a host, it's not OK to just silently `exit(1)` without outputting an error message to the user (and not even a single line to the debug log!).

Just mark it as `Core::ArgsParser::Required::Yes`, and `ArgsParser` will do everything (ensure it's not-null) for you. When `serenityos.pool.ntp.org` becomes available, we put it in as a default value and switch the argument back to being not required.

Userland/ntpquery.cpp
nico: The intent is to get something started. At the moment it's susceptible to MITM, but at the moment that's fine. If this one day becomes a system service, that needs fixing first then.

I looked at UDPSocket and it didn't do anything helpful. I expect this to grow a message loop and then that probably changes, but for now I don't see the point.

Userland/ntpquery.cpp
nico: Userland/ seems very inconsistent about using stdout or stderr for error messages. I'm in favor of making that consistent, but given the current state either's consistent with what's there :P

bugaevc: Let's do it the right way in new code then, and we can fix existing code later :slightly_smiling_face:. All the errors, info messages — or anything that is not the primary output of the program — should go to stderr.

The only tricky case is where usage info goes:
* if the usage output is printed because the program is invoked improperly, usage info counts as an error message and goes to stderr;
* but if a program is invoked with `--help`, usage info counts as primary output and goes to stdout.

Thankfully, `ArgsParser` handles that one for you.

nico:
Any hold up here?

awesomekling:
> Any hold up here?

Yes, there was some review feedback and all the conversations are "resolved" but none of the requested changes were made(?)

nico:
Yup, I explained why I disagreed with all the comments :) And there were no replies to that.

Userland/ntpquery.cpp
awesomekling: @nico I don't see a response here.

Userland/ntpquery.cpp
awesomekling: @nico Nor here

Userland/ntpquery.cpp
bugaevc: Guess I should add a reply here too, for completeness sake.

IMO, checking that a function has initialized the whole buffer is not advanced functionality, it's basic C hygiene, much like checking return values for errors. You do all that even in simplest programs and tests (perhaps not in examples that are meant to showcase something, though, in order not to clutter the example).

Checking the peer address, in the other hand... I mean, it won't prevent MITM attacks either since we're not encrypting or signing anything. It just seems like a correctness thing to me, but whatever, I don't insist.

nico:
Oh, there are replies! Apparently github doesn't unresolve discussions when a reply is added (?)

nico:
Well, apologies for missing the replies! Will reply to them once near a keyboard :(

awesomekling:
Great! Yeah, I've been bitten by the lack of auto-unresolve too :)

Userland/ntpquery.cpp
nico: I added a length check. I missed that part of your reply on first read, and this is a great catch indeed. Thanks for noticing this!

Userland/ntpquery.cpp
nico: Again: Done, but spending time on getting temporary code to be Just So doesn't seem useful (nor fun) to me.

Userland/ntpquery.cpp
nico: > Let's do it the right way in new code then

I've changed this to stderr here now.

Having done that, I don't think this is a super important thing to hold this up over. ping.cpp has the exact same code, and Userland/ is generally just fairly inconsistent about this. If you're excited about making it more consistent -- great! I low-key agree that it'd be good to be consistent about this, but it's not what I'd like to work on right now.

Issue #3254, opened 2020-08-21
Reproducible profiler crash with `BUG! Unexpected NP fault at V0x01b7cdd8`


nico:
Repro:

1. Open http://45.33.8.238/linux/ in browser
2. Open Profiler, double-click Browser process to start profiling
3. Hit "Reload" in browser, wait a second or so
4. In Profiler, hit Stop
5. Select the part of the profile where stuff happens (during the reload) in the strip at the top of the profiler
6. In the tree view at the bottom, keep hitting arrow-right/arrow-down to drill many stack frames into the profile -- basically try to get to the frame where most work is happening (~700 / 1100 samples in my case)

Actual: Crashed 3/3 times. Stack:

<details>

```
[Profiler(33:33)]: BUG! Unexpected NP fault at V0x01b7cdd8
[Profiler(33:33)]: Unrecoverable page fault, read from address V0x01b7cdd8
[Profiler(33:33)]: CRASH: CPU #0 Page Fault. Ring 3.
[Profiler(33:33)]: exception code: 0004 (isr: 0000
[Profiler(33:33)]:   pc=001b:080c8a39 flags=0202
[Profiler(33:33)]:  stk=0023:015024e0
[Profiler(33:33)]:   ds=0023 es=0023 fs=0023 gs=002b
[Profiler(33:33)]: eax=00000bcb ebx=01b70000 ecx=01502650 edx=01b7cdd8
[Profiler(33:33)]: ebp=01502558 esp=c588ffe8 esi=0811ba0c edi=01502690
[Profiler(33:33)]: cr0=80010013 cr2=01b7cdd8 cr3=03685000 cr4=00100ee4
[Profiler(33:33)]: code: 8b 0a 48 66 89 43 14 89
[Profiler(33:33)]: 0x080c8a39  malloc +491

[Profiler(33:33)]: Process regions:
[Profiler(33:33)]: BEGIN       END         SIZE        ACCESS  NAME
[Profiler(33:33)]: 08048000 -- 0810cfff    000c5000    R XS      elf-map-rx
[Profiler(33:33)]: 0810d000 -- 0811cfff    00010000    RW        elf-alloc-rw
[Profiler(33:33)]: 01101000 -- 01101fff    00001000    RW        (null)
[Profiler(33:33)]: 01103000 -- 01502fff    00400000    RW  T     Stack (Main thread)
[Profiler(33:33)]: 01504000 -- 01504fff    00001000    RW        Thread-specific
[...]

PR #3324, opened 2020-08-28
Kernel: Release page tables when no longer needed

tomuta:
When unmapping regions, check if page tables can be freed.

This is a follow-up change for #3254.

Turns out, it it should have remained a `HashMap`. This brings it back but uses a more appropriate key, which we can use to free page tables that are no longer needed. I believe turning it into a `HashTable` introduced a page table leak because we would never free page tables unless the entire `PageDirectory` was freed (process terminated).

### DIFF
--- a/Kernel/VM/MemoryManager.cpp
+++ b/Kernel/VM/MemoryManager.cpp
@@ -188,7 +188,7 @@ void MemoryManager::parse_memory_map()
     ASSERT(m_user_physical_pages > 0);
 }
 
-const PageTableEntry* MemoryManager::pte(const PageDirectory& page_directory, VirtualAddress vaddr)
+PageTableEntry* MemoryManager::pte(const PageDirectory& page_directory, VirtualAddress vaddr)
 {
     ASSERT_INTERRUPTS_DISABLED();
     ASSERT(s_mm_lock.own_lock());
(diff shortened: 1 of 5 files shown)


awesomekling:
> I believe turning it into a `HashTable` introduced a page table leak because we would never free page tables unless the entire `PageDirectory` was freed (process terminated).

You didn't cause that, we never had the feature of deallocating page tables before process death. So this is a very nice new feature :)

Kernel/VM/MemoryManager.cpp
```diff
@@ -227,13 +227,53 @@ PageTableEntry& MemoryManager::ensure_pte(PageDirectory& page_directory, Virtual
...
+    if (pde.is_present()) {
+        auto* page_table = quickmap_pt(PhysicalAddress((FlatPtr)pde.page_table_base()));
+        auto& pte = page_table[page_table_index];
+        pte.clear();
+
+        if (is_last_release || page_table_index == 0x1ff) {
+            // If this is the last PTE in a region or the last PTE in a page table then
+            // check if we can also release the page table
+            bool all_clear = true;
+            for (u32 i = 0; i <= 0x1ff; i++) {
```
awesomekling, line 258: A nice future improvement could be keeping a count somewhere, so we don't have to loop over the page table.

PR #3326, opened 2020-08-28
Enable "-Wmissing-declarations"

BenWiederhake:
All the heavy lifting was already done in #3096.

From here on, build system and CI will ensure that all new code defines compilation-unit-only code as 'static', and that dead code can be found more easily. Also, this style encourages type checking by suggesting that you put a proper declaration in a shared header.


awesomekling:
Noice!

PR #3332, opened 2020-08-28
Spreadsheet: Cell types (very basic) and implicit integers

alimpfard:
This adds a generic interface for cell types and hooks it up.
There is no way to set these from the UI, and so they're not saved anywhere yet.
Also implicitly converts numeric values (strictly integers) to numeric javascript values, as `numbery-looking + numbery-looking === string` is not very interesting.

This implicit thingy is probably temporary, until a proper UI is added for these types


alimpfard:
Hmmm...this probably does something wonky to cells with js values...

PR #3333, opened 2020-08-28
Fix git port

itamar8910:
Previously, we were unable to stage untracked files in the git port.
This fixes this behavior with a small change to Ext2FS code.
We can now create a new git repo and then stage and commit new files :smile:

### DIFF
--- a/Kernel/FileSystem/Ext2FileSystem.cpp
+++ b/Kernel/FileSystem/Ext2FileSystem.cpp
@@ -855,6 +855,28 @@ ssize_t Ext2FSInode::write_bytes(off_t offset, ssize_t count, const u8* data, Fi
     return nwritten;
 }
 
+u8 Ext2FSInode::file_type_for_directory_entry(const ext2_dir_entry_2& entry)
+{
+    switch (entry.file_type) {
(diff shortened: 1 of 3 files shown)


Kernel/FileSystem/Ext2FileSystem.cpp
```diff
@@ -855,6 +855,28 @@ ssize_t Ext2FSInode::write_bytes(off_t offset, ssize_t count, const u8* data, Fi
     return nwritten;
 }
 
+u8 Ext2FSInode::posix_file_type(const ext2_dir_entry_2& entry)
```
itamar8910, line 858: Anyone have a suggestion for a better name? :sweat_smile:

awesomekling: Well, we should probably not call it `posix_file_type` since this is not in POSIX :)
How about `file_type_for_directory_entry`?

awesomekling:
Great fix :)

awesomekling:
Hey @itamar8910 I tried adding a directory to my initial commit but that doesn't work.
Files work though!

itamar8910:
Oh, thanks. I'll look into why that is

PR #3334, opened 2020-08-28
LibJS: Demonstrate and fix weird behaviour with 'break'

alimpfard:
This demos and fixes a `break` issue that shows up as this:
```js
var j = 0;
for (var i = 0; i < 9; ++i) {
    if (j == 4) {
        break; // This doesn't break!
    }
}
```

cc @linusg, I have no idea if this is the right fix

(does `test-js` actually fail the CI?
    https://travis-ci.com/github/SerenityOS/serenity/builds/181823301#L3348
    Nope!)


linusg (approved):
Looks reasonable, thanks for the test cases!

PR #3335, opened 2020-08-28
Spreadsheet: Add a whole bunch of basic statistical functions

alimpfard:
you get stats and you get stats
have some more basic stats


alimpfard:
Mwahahaha

This should cover most entry-level stats, so now we can claim we have a spreadsheet that can do stuff!

PR #3337, opened 2020-08-28
Fracional sleep take 2 - now allow fractional sleeping in userspace utility

elcuco:
Following https://github.com/SerenityOS/serenity/pull/2192:

I added fractional arguments for sleeping in the userspace utility. I had to modify the kernel - now the `usleep` syscall tries to behave like `sleep` (returns remaining time).

Two things:

1. I need to change the syscall to return 64 bits, intead of 32. Unsure how to do this. Without it - we will get `-1` a lot.
2. I was unable to test the code path that displays how much time is left. Help would be great.

### DIFF
--- a/Kernel/Syscalls/sleep.cpp
+++ b/Kernel/Syscalls/sleep.cpp
@@ -26,6 +26,7 @@
 
 #include <Kernel/Process.h>
 #include <Kernel/Time/TimeManagement.h>
+#include <stdint.h>
 
 namespace Kernel {
 
@@ -35,8 +36,12 @@ int Process::sys$usleep(useconds_t usec)
     if (!usec)
         return 0;
     u64 wakeup_time = Thread::current()->sleep(usec * TimeManagement::the().ticks_per_second() / 1000000);
-    if (wakeup_time > g_uptime)
-        return -EINTR;
+    if (wakeup_time > g_uptime) {
+        u64 ticks_left_until_original_wakeup_time = wakeup_time - g_uptime;
+        u64 dt = ticks_left_until_original_wakeup_time / (TimeManagement::the().ticks_per_second() * 1000000);
+        // TODO - can we make this syscall return 64 bits instead of 32 bits?
+        return (dt < UINTMAX_MAX) ? dt : -1;
+    }
     return 0;
 }
 
--- a/Userland/sleep.cpp
+++ b/Userland/sleep.cpp
@@ -36,10 +36,10 @@ static void handle_sigint(int)
 
 int main(int argc, char** argv)
 {
-    int secs;
+    double fractional_seconds;
 
     Core::ArgsParser args_parser;
-    args_parser.add_positional_argument(secs, "Number of seconds to sleep for", "num-seconds");
+    args_parser.add_positional_argument(fractional_seconds, "Number of seconds to sleep for (accepts fractions)", "num-seconds");
     args_parser.parse(argc, argv);
 
     struct sigaction sa;
@@ -52,9 +52,21 @@ int main(int argc, char** argv)
         return 1;
     }
 
-    unsigned remaining = sleep(secs);
-    if (remaining) {
-        printf("Sleep interrupted with %u seconds remaining.\n", remaining);
+    useconds_t usecs = 0;
+
+    // We overflow microseconds... lets split the wait.
+    static const auto OVERFLOW_VALUE = UINTMAX_MAX / (1000 * 1000);
+    if (fractional_seconds > OVERFLOW_VALUE) {
+        perror("Number of seconds too large for sleep");
+        usecs = 0;
     }
-    return 0;
+
+    usecs = fractional_seconds * (1000 * 1000);
+
+    int remaining = usleep(usecs);
+    if (remaining) {
+        printf("Sleep interrupted with %d seconds remaining.\n", remaining / (1000 * 1000));
+        return EXIT_FAILURE;
+    }
+    return EXIT_SUCCESS;
 }


Kernel/Syscalls/sleep.cpp
```diff
@@ -29,14 +29,20 @@
 
 namespace Kernel {
 
+#define UINTMAX_MAX __UINTMAX_MAX__
```
BenWiederhake, line 32: Why not just use `__UINTMAX_MAX__`?

BenWiederhake:
If you want to know the time left after an interrupted sleep, you can use `clock_nanosleep`: It has an even higher precision, and this approach doesn't require changing syscalls that are already used elsewhere. :)

Also, `printf` in userland supports `%f`, maybe it's nicer to print fractional seconds instead of rounding the remaining time to the closest second? Just imagine getting silly output like `Sleep interrupted with 0 seconds remaining.` :D

Kernel/Syscalls/sleep.cpp
elcuco: I am unsure if SerenityOS will support another compiler, and `__UINTMAX_MAX__` is compiler specific. If and when, a simple `#ifdef` would help.

I also don't like seeing `__` in the main code, seems... ugly.

elcuco:
@BenWiederhake https://youtu.be/X7O08TrxG6I?t=220

So, basically you are telling me - that the current implementation of `sleep` was broken until now. ...

Kernel/Syscalls/sleep.cpp
BenWiederhake: Regarding compiler: That's not a question of compiler, it's a question of which libc the program uses. And since Serenity brings its own LibC, you only need compatibility with [Serenity's LibC](https://github.com/SerenityOS/serenity/blob/master/Libraries/LibC/bits/stdint.h#L75). :D

Regarding ugly: I agree that it's a bit ugly, and you don't need to use it at all. Our `stdint.h` does `#define UINTMAX_MAX __UINTMAX_MAX__`, so you can use `UINTMAX_MAX` directly! :)

BenWiederhake:
Hmm, broken? No, it "only" uses integer seconds, as you already noticed.

elcuco:
> Hmm, broken? No, it "only" uses integer seconds, as you already noticed.

I mean there is dead code which cannot be executed ("print how much time is left").  That is because, even when code is killed (sigterm) - the syscall returns 0. This should be addressed in another PR (unless you insist of fixing it here right now).

BTW - using `clock_nanosleep()` does not help me - since it does not return "how much is left". I assume I can `time(NULL)` before and after... but this seems like a workaround, and and will return wrong values when multitasking ("critical section and stuff).

BenWiederhake:
Hmm, I think there may be some misunderstandings:
- Are you sure the `sleep` syscall always returns 0? [The syscall code](https://github.com/SerenityOS/serenity/blob/master/Kernel/Syscalls/sleep.cpp#L51) looks like this shouldn't happen.
- `clock_nanosleep` absolutely tells you how much time remains, as part of its [specification (`struct timespec *remain`)](https://linux.die.net/man/2/clock_nanosleep). Or do you mean that this is broken somehow?

elcuco:
I type  "sleep 32131.8" and I press control-c - I added a `printf()` that displays how much time is left. Program stops immediately, and the syscall returns 0. Am I testing wrong?

Completely ignoring `clock_nanosleep` at the moment, since I would like to know if I am doing something wrong, or this specific syscall is broken.

Kernel/Syscalls/sleep.cpp
elcuco: Crap, using the same libc in userspace and kernel space is epic. I need to get used to it, thanks! (fixed, pushed waiting for CI to give me green light)

BenWiederhake:
Which syscall?

- `sys$sleep` returns the number of seconds remaining, used to be used in `Userland/sleep.cpp`, and seems it works just fine:
  sleep
- `sys$usleep` is supposed to return [only `0` or `-1`](https://linux.die.net/man/3/usleep). Looking at the source code of `Process::sys$usleep`, this should work just fine. (Note that LibC stores the `EINTR` into `errno` and returns `-1`.)

Regarding "ignoring nanosleep": Thatt's a good habit. (Stopping and checking whether is something is broken) Keep it up! :)

Kernel/Syscalls/sleep.cpp
```diff
@@ -35,8 +36,12 @@ int Process::sys$usleep(useconds_t usec)
     if (!usec)
         return 0;
     u64 wakeup_time = Thread::current()->sleep(usec * TimeManagement::the().ticks_per_second() / 1000000);
-    if (wakeup_time > g_uptime)
-        return -EINTR;
+    if (wakeup_time > g_uptime) {
+        u64 ticks_left_until_original_wakeup_time = wakeup_time - g_uptime;
+        u64 dt = ticks_left_until_original_wakeup_time / (TimeManagement::the().ticks_per_second() * 1000000);
+        // TODO - can we make this syscall return 64 bits instead of 32 bits?
```
BenWiederhake, line 42: There already is a 64 bit sleep-like syscall: `clock_nanosleep` :P

In theory, we could even remove `sys$sleep` and `sys$usleep` and implement the LibC functions just through `sys$clock_nanosleep`.

awesomekling: Indeed, we don't need 3 separate syscalls. I've removed `sys$sleep` and `sys$usleep` since everything can be implemented in terms of `sys$clock_nanosleep` with `CLOCK_MONOTONIC`

awesomekling:
Fixed in fd29bed656ffdad79edd8d8a614195d4ae53ae58, closing abandoned PR.

PR #3341, opened 2020-08-29
LibCompress: Add GzipDecompressor implemenation and gunzip fontend.

asynts:
I am totally aware that there is another Gzip implementation in `Libraries/LibCore/Gzip.h` but I can't remove that until dynamic blocks are supported by the deflate implementation. The days of `Libraries/LibCore/puff.h` are counted too.

---

Currently, the deflate implementation doesn't support dynamic blocks which are very commonly used. Typically, very short files use fixed blocks, I've setup one such file for testing here: [simple-gzip-file.gz][1]. The following should produce "abcabc":
~~~none
pro http://static.asynts.com/08/29/simple-gzip-file.gz > simple-gzip-file.gz
gunzip -c simple-gzip-file.gz
~~~

  [1]: http://static.asynts.com/08/29/simple-gzip-file.gz

### DIFF
--- a/Libraries/LibCompress/Gzip.cpp
+++ b/Libraries/LibCompress/Gzip.cpp
@@ -1,0 +1,186 @@
+/*
+ * Copyright (c) 2020, the SerenityOS developers.
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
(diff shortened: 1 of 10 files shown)


alimpfard:
> curl -O http://static.asynts.com/08/29/simple-gzip-file.gz

Friendly reminder that
`pro http://static.asynts.com/08/29/simple-gzip-file.gz > simple-gzip-file.gz`
is a thing :P

asynts:
@alimpfard Thanks! I didn't know that utility existed. I've updated the code snippet.

Libraries/LibCompress/Gzip.h
```diff
@@ -0,0 +1,84 @@
...
+        u8 extra_flags;
+        u8 operating_system;
+
+        bool valid_magic_number() const;
+        bool supported_by_implementation() const;
+    };
+
+    struct Flags {
+        static constexpr u8 FTEXT = 1 << 0;
+        static constexpr u8 FHCRC = 1 << 1;
```
stelar7, line 62: Something doesnt seem right about these shifts :P

Libraries/LibCompress/Gzip.cpp
```diff
@@ -0,0 +1,188 @@
...
+        size_t nread = m_block_stream.read(bytes);
+
+        if (nread < bytes.size()) {
+            m_block_stream.~DeflateDecompressor();
+            m_state = State::Idle;
+
+            LittleEndian<u32> crc32, input_size;
+            m_input_stream >> crc32 >> input_size;
+
+            // FIXME: Validate crc32 and input_size.
```
stelar7, line 83: There is a CRC impl in LibCrypto/Checksum that you can use :)

Libraries/LibCompress/Gzip.cpp
```diff
@@ -0,0 +1,188 @@
...
+namespace Compress {
+
+bool GzipDecompressor::BlockHeader::valid_magic_number() const
+{
+    return identification_1 == 0x1f && identification_2 == 0x8b;
+}
+
+bool GzipDecompressor::BlockHeader::supported_by_implementation() const
+{
+    if (compression_method != 0x08) {
```
stelar7, line 40: Maybe a const for 0x08? (DEFLATE)

Libraries/LibCompress/Gzip.h
asynts: Oops.

Libraries/LibCompress/Gzip.cpp
asynts: There are really no other compression methods that I am aware of. The standard only defines deflate. I think the comment is enough.

Libraries/LibCompress/Gzip.cpp
asynts: I've added it and it seems to work.

PR #3342, opened 2020-08-29
Userland: Make 'pro' capable of guessing a filename

alimpfard:
From #3341 I realised that `pro` needs some love too.
This PR adds support for:
- ~~`-o filename` (Save to given file)~~
- `-O` (Save to a file named as the server or the URL suggests)

and also partially implements `freopen()`.
No more `pro url > filename`!

### DIFF
--- a/Libraries/LibC/stdio.cpp
+++ b/Libraries/LibC/stdio.cpp
@@ -77,6 +77,8 @@ public:
     pid_t popen_child() { return m_popen_child; }
     void set_popen_child(pid_t child_pid) { m_popen_child = child_pid; }
 
+    void reopen(int fd, int mode);
+
 private:
     struct Buffer {
         // A ringbuffer that also transparently implements ungetc().
@@ -413,6 +415,23 @@ long FILE::tell()
         return -1;
 
     return lseek(m_fd, 0, SEEK_CUR);
+}
+
+void FILE::reopen(int fd, int mode)
+{
+    // Dr. POSIX says: "Failure to flush or close the file descriptor
+    //                  successfully shall be ignored"
+    // and so we ignore any failures these two might have.
+    flush();
+    close();
+
+    // Just in case flush() and close() didn't drop the buffer.
+    m_buffer.drop();
+
+    m_fd = fd;
+    m_mode = mode;
+    m_error = 0;
+    m_eof = false;
 }
 
 FILE::Buffer::~Buffer()
@@ -967,10 +986,19 @@ FILE* fopen(const char* pathname, const char* mode)
 
 FILE* freopen(const char* pathname, const char* mode, FILE* stream)
 {
-    (void)pathname;
-    (void)mode;
-    (void)stream;
-    ASSERT_NOT_REACHED();
+    ASSERT(stream);
+    if (!pathname) {
+        // FIXME: Someone should probably implement this path.
+        TODO();
+    }
+
+    int flags = parse_mode(mode);
+    int fd = open(pathname, flags, 0666);
+    if (fd < 0)
+        return nullptr;
+
+    stream->reopen(fd, flags);
+    return stream;
 }
 
 FILE* fdopen(int fd, const char* mode)
--- a/Userland/pro.cpp
+++ b/Userland/pro.cpp
@@ -24,20 +24,116 @@
  * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
+#include <AK/GenericLexer.h>
+#include <AK/LexicalPath.h>
 #include <AK/NumberFormat.h>
 #include <AK/SharedBuffer.h>
 #include <AK/URL.h>
 #include <LibCore/ArgsParser.h>
 #include <LibCore/EventLoop.h>
+#include <LibCore/File.h>
 #include <LibProtocol/Client.h>
 #include <LibProtocol/Download.h>
 #include <stdio.h>
 
+// FIXME: Move this somewhere else when it's needed (e.g. in the Browser)
+class ContentDispositionParser {
+public:
+    ContentDispositionParser(const StringView& value)
+    {
+        GenericLexer lexer(value);
+
+        lexer.consume_while([](auto c) { return is_whitespace(c); });
+
+        if (lexer.consume_specific("inline")) {
+            m_kind = Kind::Inline;
+            if (!lexer.is_eof())
+                m_might_be_wrong = true;
+            return;
+        }
+
+        if (lexer.consume_specific("attachment")) {
+            m_kind = Kind::Attachment;
+            if (lexer.consume_specific(";")) {
+                lexer.consume_while([](auto c) { return is_whitespace(c); });
+                if (lexer.consume_specific("filename=")) {
+                    m_filename = lexer.consume_quoted_string();
+                } else {
+                    m_might_be_wrong = true;
+                }
+            }
+            return;
+        }
+
+        if (lexer.consume_specific("form-data")) {
+            m_kind = Kind::FormData;
+            while (lexer.consume_specific(";")) {
+                lexer.consume_while([](auto c) { return is_whitespace(c); });
+                if (lexer.consume_specific("name=")) {
+                    m_name = lexer.consume_quoted_string();
+                } else if (lexer.consume_specific("filename=")) {
+                    m_filename = lexer.consume_quoted_string();
+                } else {
+                    m_might_be_wrong = true;
+                }
+            }
+
+            return;
+        }
+
+        // FIXME: Support 'filename*'
+        m_might_be_wrong = true;
+    }
+
+    enum class Kind {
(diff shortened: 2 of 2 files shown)


bugaevc:
> `-o filename` (Save to given file)

Let's *not* do that (add options to programs to redirect input/output). With my capabilities fan hat on, you don't make a service obtain the capabilities it needs to service a request on its own, you pass it the required capabilities. With my Unix user hat on, you just do `>` instead of `-o`. That's always more consistent, and one less thing to worry about for the program implementation — you just *always* write the (primary) output to stdout, because that's already where whoever's invoked you wants you to write it.

`-O` makes some more sense because the program may reason about a desirable file name in a way the invoking user likely cannot (such as by receiving a filename from the server it talks to).

Userland/pro.cpp
```diff
@@ -24,23 +24,107 @@
...
 {
     const char* url_str = nullptr;
+    const char* output_file = nullptr;
+    bool save_at_provided_name = false;
+    bool silent = false;
 
     Core::ArgsParser args_parser;
+    args_parser.add_option(output_file, "Output file path", "output", 'o', "file");
+    args_parser.add_option(save_at_provided_name, "Write to a file named as the remote file", nullptr, 'O');
+    args_parser.add_option(silent, "Hide all output", "silent", 's');
```
bugaevc, line 116: This is typically named `-q`/`--quiet`. And I'm not convinced having it is a good idea either, compared to `2>/dev/null`. And it doesn't seem to be used anywhere?

alimpfard:
Hmmm...if it needs to be able to write to a specific file (in the case of `-O`), it feels like a waste to not use that and instead force it to write to stdout instead :/

Userland/pro.cpp
alimpfard: Oh I added it, decided against it, then I guess forgot to remove the option!

bugaevc:
> if it needs to be able to write to a specific file (in the case of `-O`)

You just do it like that: `freopen(stdout, "whatever name", "w")`, and then continue using stdout. Or the usual open + dup2 + close dance.

alimpfard:
You make sense, and in a usual program, I'd agree with the 'dump to stdout' idea.
but in a utility geared towards downloading things, I feel like I never want it to dump stuff to stdout

alimpfard:
As a case-in-point, I've never once used `curl` without `-o` or `-O`
(I specially hate the "pipe curl to sh" thing too)

tomuta:
I agree with @alimpfard, I feel a `-o` or a way to specify a filename is better than redirecting stdout. I script quite a few things at work and always hate doing that.

bugaevc:
> but in a utility geared towards downloading things, I feel like I never want it to dump stuff to stdout

*but stdout is not the terminal!!*

stdout is *where whoever's invoking you wants you to output your primary, em, output*. stdin is wherever they want you to read input from. stderr is wherever they want you to output misc data / logs / errors / progress to.

By default, when the Terminal starts up, it spawns the shell with its stdin, stdout, and stderr connected to the tty/pty — because it wants the shell to read its input, write its output, and write its logs/errors/prompts from/to the tty. But that doesn't mean programs should assume that stdin/stdout/stderr is the terminal "by default, unless redirected" — no, that's not how this works, it's just whichever 0/1/2 fds have been passed to you.

(You might have noticed this is one of those things I feel very strongly about... 😀)

It might, indeed, be a not-so-great idea to dump downloaded stuff *onto the terminal*. So we might want to detect if we are to output binary to the tty (basically, `isatty(1)`, and if so, suggest the user to pick a different destination.

> As a case-in-point, I've never once used `curl` without `-o` or `-O`

Hmm, I do use curl myself without `-o` or `-O` all the time.

> I feel a `-o` or a way to specify a filename is better than redirecting stdout.

Please explain how is it better?

alimpfard:
> but stdout is not the terminal!!

To an absentminded user...it is.
Now as for detecting whether it's a tty, and refusing to run normally _is_ acceptable, but then others that _do_ make curl dump to stdout will complain about having to pass a flag for what should obviously be the default behaviour.

Also in general, redirection seems like the second thing someone reaches for when trying to do highly effectful things

alimpfard:
as a joking aside, you probably hate `tee` to the core; it should just tee `stdin` to `stdout` and `stderr` lol

bugaevc:
> but then others that do make curl dump to stdout will complain about having to pass a flag for what should obviously be the default behaviour.

You're confusing stdout and tty again 😉 it's perfectly fine to have `pro` dump to stdout (that's what I'm advocating for!), but probably a bad idea to dump binary data to a tty. Yes, it's unclear if refusing to run with stdout=tty unless a flag is passed is a good idea — perhaps it could, like, ask for confirmation, and only in case the data is indeed binary (non-textual), but I'm not completely sure that's a good idea either.

Don't get me wrong, a simple option won't hurt anyone; but I feel it encourages inconsistency (unless *everything* starts to add `-o` options, which is, ooof... a nightmare) and "doing it the wrong way".

> as a joking aside, you probably hate `tee` to the core

Actually `tee` (and its doppelganger, `cat`) just have to accept filenames in argv, because the Unix stdio conventions only provide one input and one output stream (not counting stderr), which is absolutely enough for 99% of programs, bit not for all of them. Another example is `diff`, which has to accept two files, so it does that via filenames.

On a somewhat related note, the `xclip` tool (one of the X11 clipboard utilities) has an `xclip -f` mode, where you can pipe a stream into it, and it will copy it to the clipboard and also output it to its stdout, so you can pipe it further, like this `foo | xclip -f | bar`. There's no such option in Serenity's `copy`, you know why? Because you can do the same with existing tools like this: `foo | tee >(copy) | bar`. Just like that, every program does one thing and does it (hopefully) well: `copy` copies to clipboard, `tee` duplicates a stream — and each program doesn't have to grow support for a `-f` argument to be usable this way.

alimpfard:
> it's perfectly fine to have `pro` dump to stdout

I guess I just can't think of a reason for that to be the behaviour of a program that is meant to download things.

I'll go with dump-to-stdout and reopen-stdout for now, but I wonder if doing so is just forcing things into leaky abstractions.

alimpfard:
@bugaevc - `freopen()` review pls

Libraries/LibC/stdio.cpp
```diff
@@ -77,6 +77,8 @@ struct FILE {
     pid_t popen_child() { return m_popen_child; }
     void set_popen_child(pid_t child_pid) { m_popen_child = child_pid; }
 
+    void reopen(int fd, int flags);
```
bugaevc, line 80: The same thing is called `mode` in the methods above, so let's be consistent.

Libraries/LibC/stdio.cpp
```diff
@@ -415,6 +417,18 @@ long FILE::tell()
     return lseek(m_fd, 0, SEEK_CUR);
 }
 
+void FILE::reopen(int fd, int flags)
+{
+    // Should these failing fail something?
```
bugaevc, line 422: Dr. POSIX says "Failure to flush or close the file descriptor successfully shall be ignored. The error and end-of-file indicators for the stream shall be cleared.", so no.

(Change the comment to state that we ignore errors intentionally.)

alimpfard: Dr. POSIX is just a suggestion at this point, I thought we might want something so I left that comment there _specifically_ for you :)

Libraries/LibC/stdio.cpp
```diff
@@ -415,6 +417,20 @@ long FILE::tell()
...
 }
 
+void FILE::reopen(int fd, int mode)
+{
+    // Dr. POSIX says: "Failure to flush or close the file descriptor
+    //                  successfully shall be ignored"
+    // and so we ignore any failures these two might have.
+    flush();
+    close();
+
```
bugaevc, line 427: I *think* we have to drop the buffer here (as in, `m_buffer.drop()`). `flush()` normally does that (or equivalent), but flushing may not have succeeded, and since we ignore the error, it's up to us to ensure we don't retain any buffered data from the old file.

Libraries/LibC/stdio.cpp
bugaevc: Sure, it is just a suggestion, but in this case it makes sense for us to take its suggestion :)

Libraries/LibC/stdio.cpp
alimpfard: I don't think it's dropped when something fails while writing, but I didn't look deep enough into it, so I'll call `drop()` there

alimpfard:
oddly interesting, why is test-js segfaulting?

awesomekling:
fwiw I like `-o` but I also recognize the beauty in using `stdout`
I don't have an argument against `stdout`

alimpfard:
Well, as of right now this patch adds exactly what I want, which is `-O`.
I have no issues with using `stdout` when `-O` isn't passed (except the abstraction thing, but that's really just a thought).

PR #3343, opened 2020-08-29
FontEditor: Don't corrupt nearby glyphs, don't crash on fixed-width

BenWiederhake:
Elevator pitch: "Two bugfixes for the price of one PR!" ;)

This PR:
- prevents `GlyphEditor` from drawing into adjacent glyphs when the user just continues dragging the mouse outside of the window.
- fixes FontEditor (which used to crash) when the checkbox for fixed-witdth is used.
- combs through the rest of Serenity for similar "reference to lambda on stack frame" bugs (found none, but maybe it inspires someone to avoid this bug in the future)
- removes a few unnecessary lambdas.

The "reference to lambda on stack frame"-bug is evil because it's a leaky abstraction. On the one hand, the abstraction is that the programmer doesn't need to worry about scopes and lifetimes, the compiler will just copy/reference everything as needed. On the other hand, references to local variables mean that when the lambda is run, it still holds a reference to somewhere inside a stack frame that may *or may not* still be valid! In the case of FontEditor, using the checkbox triggered the "may not" case.

This (potential) bug affects all lambdas that takes its arguments via `[&]` and doesn't pay close attention to what exactly this pulls in. This is one of those bugs that Rust's model of making lifetimes part of the type system would have prevented, btw.

### DIFF
--- a/Applications/FontEditor/FontEditor.cpp
+++ b/Applications/FontEditor/FontEditor.cpp
@@ -263,7 +263,7 @@ FontEditorWidget::FontEditorWidget(const String& path, RefPtr<Gfx::Font>&& edite
         info_label.set_text(builder.to_string());
     };
 
-    fixed_width_checkbox.on_checked = [&](bool checked) {
+    fixed_width_checkbox.on_checked = [&, update_demo](bool checked) {
         m_edited_font->set_fixed_width(checked);
         glyph_width_spinbox.set_enabled(!m_edited_font->is_fixed_width());
         glyph_width_spinbox.set_value(m_edited_font->glyph_width(m_glyph_map_widget->selected_glyph()));
--- a/Applications/TextEditor/TextEditorWidget.cpp
+++ b/Applications/TextEditor/TextEditorWidget.cpp
@@ -86,15 +86,9 @@ TextEditorWidget::TextEditorWidget()
             update_title();
     };
 
-    auto update_statusbar_cursor_position = [this] {
-        StringBuilder builder;
-        builder.appendf("Line: %d, Column: %d", m_editor->cursor().line() + 1, m_editor->cursor().column());
-        m_statusbar->set_text(builder.to_string());
-    };
-
     m_page_view = splitter.add<Web::InProcessWebView>();
     m_page_view->set_visible(false);
-    m_page_view->on_link_hover = [this, update_statusbar_cursor_position = move(update_statusbar_cursor_position)](auto& url) {
+    m_page_view->on_link_hover = [this](auto& url) {
         if (url.is_valid())
             m_statusbar->set_text(url.to_string());
         else
@@ -300,9 +294,7 @@ TextEditorWidget::TextEditorWidget()
 
     m_statusbar = add<GUI::StatusBar>();
 
-    m_editor->on_cursor_change = [update_statusbar_cursor_position = move(update_statusbar_cursor_position)] {
-        update_statusbar_cursor_position();
-    };
+    m_editor->on_cursor_change = [this] { update_statusbar_cursor_position(); };
 
     m_new_action = GUI::Action::create("New", { Mod_Ctrl, Key_N }, Gfx::Bitmap::load_from_file("/res/icons/16x16/new.png"), [this](const GUI::Action&) {
         if (m_document_dirty) {
@@ -642,4 +634,11 @@ void TextEditorWidget::update_html_preview()
     auto current_scroll_pos = m_page_view->visible_content_rect();
     m_page_view->load_html(m_editor->text(), URL::create_with_file_protocol(m_path));
     m_page_view->scroll_into_view(current_scroll_pos, true, true);
+}
+
+void TextEditorWidget::update_statusbar_cursor_position()
+{
+    StringBuilder builder;
+    builder.appendf("Line: %d, Column: %d", m_editor->cursor().line() + 1, m_editor->cursor().column());
+    m_statusbar->set_text(builder.to_string());
 }
--- a/Kernel/FileSystem/Ext2FileSystem.cpp
+++ b/Kernel/FileSystem/Ext2FileSystem.cpp
@@ -454,6 +454,8 @@ Vector<Ext2FS::BlockIndex> Ext2FS::block_list_for_inode_impl(const ext2_inode& e
     if (!blocks_remaining)
         return list;
 
+    // Don't need to make copy of add_block, since this capture will only
+    // be called before block_list_for_inode_impl finishes.
     auto process_block_array = [&](unsigned array_block_index, auto&& callback) {
         if (include_block_list_blocks)
             add_block(array_block_index);
--- a/Applications/Calendar/AddEventDialog.cpp
+++ b/Applications/Calendar/AddEventDialog.cpp
@@ -63,15 +63,11 @@ AddEventDialog::AddEventDialog(Core::DateTime date_time, Window* parent_window)
     top_container.set_preferred_size(0, 45);
     top_container.layout()->set_margins({ 4, 4, 4, 4 });
 
-    auto make_label = [&](const StringView& text, GUI::Widget& widget, bool bold = false) {
-        auto& label = widget.add<GUI::Label>(text);
-        label.set_text_alignment(Gfx::TextAlignment::CenterLeft);
-        label.set_size_policy(GUI::SizePolicy::Fill, GUI::SizePolicy::Fixed);
-        label.set_preferred_size(0, 14);
-        if (bold)
-            label.set_font(Gfx::Font::default_bold_font());
-    };
-    make_label("Add title & date:", top_container, true);
+    auto& add_label = top_container.add<GUI::Label>("Add title & date:");
+    add_label.set_text_alignment(Gfx::TextAlignment::CenterLeft);
+    add_label.set_size_policy(GUI::SizePolicy::Fill, GUI::SizePolicy::Fixed);
+    add_label.set_preferred_size(0, 14);
+    add_label.set_font(Gfx::Font::default_bold_font());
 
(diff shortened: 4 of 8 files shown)


Games/2048/main.cpp
```diff
@@ -131,7 +131,7 @@ int main(int argc, char** argv)
 
         GUI::MessageBox::show(window, "New settings have been set and will be applied on the next game", "Settings Changed Successfully", GUI::MessageBox::Type::Information);
     };
-    auto start_a_new_game = [&] {
+    auto start_a_new_game = [&, update] {
```
alimpfard, line 134: How is this thing captured here if not by ref?
lambdas with captures shouldn't be copyable, so that's either a const reference or just a plain reference?

alimpfard: Yeah no, it's grabbed by const reference, even when `start_a_new_game` is explicitly mutable:
```c++
auto start_a_new_game = [&, update]() mutable {
        update = 1;
```

->

```
../Games/2048/main.cpp:101:21: note:   no known conversion for argument 1 from
'int' to 'const main(int, char**)::<lambda()>&'
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

BenWiederhake: Interesting! In FontEditor it's clearly a `const` value, not ref:

```
../Applications/FontEditor/FontEditor.cpp:267:23: error: no match for 'operator=' (operand types are 'const FontEditorWidget::FontEditorWidget(const AK::String&, AK::RefPtr<Gfx::Font>&&)::<lambda()>' and 'int')
  267 |         update_demo = 1;
      |                       ^
```

Why can it be copied here, but not there?

BenWiederhake: Lambdas *are* copyable. That's what `TextEditor: Don't try to move(lambda)` relies on, and here it's in the specification, ctrl-f for "copy constructor": https://en.cppreference.com/w/cpp/language/lambda

alimpfard: Did you mark the containing lambda as `mutable`?
Come think of it I'm not exactly sure how it determines if a lambda is copyable.
(If it's just "if all captured refs are copyable", then no `GUI::Widget` is copyable, and so it shouldn't even compile...)

alimpfard: > Lambdas are copyable

IIRC only ones with no captures?

BenWiederhake: Well, it says "according to the usual rules for copy constructors", so:
- In FontEditor, `update_demo` stores a reference to some labels. References are copyable. Therefore, `update_demo` is copyable.
- In 2048, `update` takes everything by reference. Therefore, `update` is copyable.

So … no problem there, right?

alimpfard: I suppose that makes sense.
Though copying lambdas feels weird, but eh, better than it being wrong

tomuta: The same copy/move requirements should apply to lambdas and their captures. A lambda is copyable as long as all captures are, and movabale as long as all captures are. references/const references are always copyable/movable.

BenWiederhake: Yup, that's why I've been quoting the exact same thing from the [reference](https://en.cppreference.com/w/cpp/language/lambda) :)

> better than it being wrong

That's a critical point actually, because most of usages are *correct* – but error-prone, as we have now seen with FontEditor, and that's why I'm making all these changes.

So I'd say "better copy a few times too many (and waste a hundred bytes) than once too few (and try to read from an invalid stackframe)".

What do you think about that? If you don't like it, I can just chuck all the other changes, and only keep the rest (i.e., fixes, explanations, and lambda-removals).

tomuta: In this case, `update` should be not copied. An implicit reference to the value on the stack is perfectly ok. Besides, `update` itself references the stack. Copying it probably won't hurt anything, but it won't solve any problem. It's fine the way it is right now. None of the stack variables ever go out of scope until the main function exits, so it's perfectly fine to do it the way it's done currently.

BenWiederhake: That actually answers my questions. Alright, I'll chuck the "precautionary" commits :)

Applications/FontEditor/FontEditor.cpp
```diff
@@ -263,7 +263,7 @@ FontEditorWidget::FontEditorWidget(const String& path, RefPtr<Gfx::Font>&& edite
         info_label.set_text(builder.to_string());
     };
 
-    fixed_width_checkbox.on_checked = [&](bool checked) {
+    fixed_width_checkbox.on_checked = [&, update_demo](bool checked) {
```
tomuta, line 266: This is a legitimate case, however update_demo itself also references the stack. The stack variables referenced by update_demo probably should be either capture-copied, or a better solution would be to move them to members of the class.

tomuta: And along the lines of class members, maybe `update_demo` really should just be a member function.

BenWiederhake:
```
    auto update_demo = [&] {
        demo_label_1.update();
        demo_label_2.update();
    };
```
This is not reference to the stack, as `demo_label_{1,2}` are themselves references: `GUI::Label&` to be precise.

But you're right, this is not as obvious as it should be. How could this be made more "obviously correct"?

Games/2048/main.cpp
BenWiederhake: I removed the commits in question, which resolves this discussion as far as I understand. If you disagree, or if I misunderstood something, please respond :)

Applications/TextEditor/TextEditorWidget.cpp
```diff
@@ -300,7 +300,7 @@ TextEditorWidget::TextEditorWidget()
 
     m_statusbar = add<GUI::StatusBar>();
 
-    m_editor->on_cursor_change = [update_statusbar_cursor_position = move(update_statusbar_cursor_position)] {
```
tomuta, line 303: Well, a capture-move actually does move and not copy it. So the change for `on_link_hover` is correct, but I would keep the `on_cursor_change`. Or better yet, make `update_statusbar_cursor_position` a member function.

BenWiederhake: > make update_statusbar_cursor_position a member function.

Will do!

> Well, a capture-move actually does move and not copy it.

Can you elaborate this a bit more? I feel like I don't have a good enough grasp on the move-semantics of a closure. (Which, here, happens to be in the capture of another closure.)

Applications/FontEditor/FontEditor.cpp
tomuta: I'm not sure if a reference to a reference is clearly defined to do this. Are you certain that in this case it will capture those references by copying the values?

Applications/TextEditor/TextEditorWidget.cpp
tomuta: Well, that's just moving a capture into the closure rather than copying it. That was a big problem in C++11 where you couldn't move objects that were movable but not copyable into closure captures. Think of types that e.g. represent file descriptors, where such classes may be movable, but probably not copyable. This move assignment for captures allows you to take such objects and move them into a lambda.

Applications/FontEditor/FontEditor.cpp
tomuta: Did a simple test and at least GCC does copy references to references implicitly for lambdas, so you're correct.

Applications/TextEditor/TextEditorWidget.cpp
tomuta: And such lambdas then obviously would only be movable, not copyable.

BenWiederhake: Done ^^

Applications/FontEditor/FontEditor.cpp
BenWiederhake: I believe C++ does not allow references-to-references, which is one of the reasons the `T&&` syntax is even possible.

I'm not sure where in the reference to look for this, though. What do you suggest? Keep it like this? Add a `static_assert` for typechecking? Write something else entirely?

tomuta: Based on a simple test it looks like at least GCC copies the value (pointer) of the reference, which would make it work as expected. I don't know what the standard says, but it probably says the same thing then. I'd keep it like that, or make `update_demo` a member function altogether.

BenWiederhake: In the other thread, making it a member function was a good idea. Here it would require two new members plus the function itself – nah, I think I'll stick with the lambda.

I believe that resolves it. Let me know if I missed anything :)

Kernel/FileSystem/Ext2FileSystem.cpp
```diff
@@ -454,6 +454,8 @@ Vector<Ext2FS::BlockIndex> Ext2FS::block_list_for_inode_impl(const ext2_inode& e
     if (!blocks_remaining)
         return list;
 
+    // Don't need to make copy of add_block, since this capture will only
+    // be called before block_list_for_inode_impl finishes.
```
awesomekling, line 458: This comment seems kind of excessive to me. Are we supposed to put these whenever we use lambdas that don't persist past the end of a function?

BenWiederhake: My thought process was: "It took me quite a while to verify that it's correct, let me just put a comment so the next person can verify it more quickly."

If you'd like me to remove the comment, I can.

awesomekling: Nah, it's all good. Just commenting :)

PR #3344, opened 2020-08-29
Icons: Go hard in the paint

thankyouverycool:
A very Seredmondy icon refresh for SystemMenu


awesomekling:
Very rad! You have a great eye for these :)

PR #3350, opened 2020-08-29
Kernel: Make Heap implementation reusable, and make kmalloc expandable

tomuta:
Add an ExpandableHeap and switch kmalloc to use it, which allows
for the kmalloc heap to grow as needed.

In order to make heap expansion to work, we keep around a 1 MiB backup
memory region, because creating a region would require space in the
same heap. This means, the heap will grow as soon as the reported
utilization is less than 1 MiB. It will also return memory if an entire
subheap is no longer needed, although that is rarely possible.

As a fun fact, this boots with as little as a 32 KiB initial heap, which is enough to get `MemoryManager` initialized (after which the kmalloc heap will be expandable).
```c++
#define POOL_SIZE (32 * KiB)
```
Screenshot from 2020-08-29 16-53-54
Screenshot from 2020-08-29 16-55-20


awesomekling:
Oh this is very nice!

PR #3360, opened 2020-08-31
Add __progname, err(3) functions.

The-King-of-Toasters:
These commits add a new global string, `__progname`, which contains the basename of the current process.
It also adds the err(3) family error reporting functions.

Both of these are based on existing BSD systems. I ran into some troubles with the dynlib demo, so I can't confirm if it will work there. Seems to compile fine.

### DIFF
--- a/Libraries/LibC/CMakeLists.txt
+++ b/Libraries/LibC/CMakeLists.txt
@@ -6,6 +6,7 @@ set(LIBC_SOURCES
     cxxabi.cpp
     dirent.cpp
     dlfcn.cpp
+    err.cpp
     fcntl.cpp
     getopt.cpp
     grp.cpp
--- a/Libraries/LibC/crt0.cpp
+++ b/Libraries/LibC/crt0.cpp
@@ -54,6 +54,15 @@ int _start(int argc, char** argv, char** env)
     for (size_t i = 0; i < size; i++)
         (*__init_array_start[i])(argc, argv, env);
 
+    if (argc > 0 && argv[0] != nullptr) {
+        __progname = argv[0];
+        for (const char *s = __progname; *s != '\0'; s++) {
+            if (*s == '/') {
+                __progname = s + 1;
+            }
+        }
+    }
+
     int status = main(argc, argv, environ);
 
     exit(status);
--- a/Libraries/LibC/err.cpp
+++ b/Libraries/LibC/err.cpp
@@ -1,0 +1,140 @@
+/*
+ * Copyright (c) 2020, The SerenityOS developers.
+ * All rights reserved.
+ *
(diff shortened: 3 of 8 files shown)


awesomekling:
Hmm. What is the purpose of adding non-standard BSD extensions to the C library?

We're not going to use them ourselves, since we write in C++, and it seems unlikely that 3rd party ports will require them since they are indeed non-standard extensions.
Also, `warn()` collides with the existing stream-based `warn()` from `AK/LogStream.h`

The-King-of-Toasters:
I mentioned this in the commit, but they are used to simplify the following idioms:

```c
perror("func");
exit(1);
// Becomes
err(1, "func");
```

or

```c
fprintf(stderr, "program: something wrong: %d", important_value);
// Becomes
warn("something wrong: %d", important_value);
```

I see both of these a lot in the current codebase, so I thought it would be cool to hack crt0 and add this, then go through later and replace them all.

Also, some of the ports bundle their own versions of err() - mandoc and anything from https://github.com/ibara/libpuffy (I think? I don't actually know how he bulids them). I also like to try compiling my stuff on serenity as well, which uses https://github.com/kristapsdz/oconfigure.

awesomekling:
I don't think we'll be needing these since we're moving away from `printf`-style API's in general and to our own C++ string formatters.

The-King-of-Toasters:
That's understandable. Would you consider adding the first patch on it's own, for portability and convince reasons? A lot of other *NIXs provide similar functionality. BSD has `__progname` Linux has `program_invocation_(short_)name` and illumos has `getexecname()`, all of which are created similarly (with the exception being illumos which calls dlinfo on itself and gets `argv` from there).

bugaevc:
We have

https://github.com/SerenityOS/serenity/blob/0f5214447790afa11907f57a2dcac70cb64655a7/Libraries/LibC/unistd.h#L57

The-King-of-Toasters:
That works, although having a buffer size check makes it complicated compared to the others which are "free". I guess this is from having `m_name` be an `AK::String`.

bugaevc:
How would you do this without having to allocate a buffer of some size? `argv[0]` (as pushed by the kernel onto the stack, which is also if you think of it a buffer of sorts) does not cut it, because it doesn't have to match what Serenity considers "process name".

The-King-of-Toasters:
Hmmm, I was going to ask "what's wrong with passing a pointer to the `char *` underneath `m_name`" but I can imagine some stack antics taking place.

bugaevc:
> what's wrong with passing a pointer to the `char *` underneath `m_name`

It's, uh, allocated on the kernel heap, which is not accessible to the userspace

PR #3362, opened 2020-08-31
LibCompress: Add support for dynamic deflate blocks.

asynts:
Now it should theoretically be possible to decompress any Gzip file with `gunzip`. However, the `InputFileStream` thing doesn't do any buffering which makes it effectively impossible to read actual large files. However, files with strong compression can still be read and decompressed (because the input files are small). I've setup a file with 64KiB of zeroes which can be decompressed:

~~~none
pro -O http://static.asynts.com/08/31/foo.bin.gz
gunzip foo.bin.gz
~~~

---

I'll tackle the `<LibCore/FileStream.h>` header next.


Libraries/LibCompress/Deflate.cpp
```diff
@@ -345,12 +346,97 @@ u32 DeflateDecompressor::decode_distance(u32 symbol)
     ASSERT_NOT_REACHED();
 }
 
-void DeflateDecompressor::decode_codes(CanonicalCode&, CanonicalCode&)
+void DeflateDecompressor::decode_codes(CanonicalCode& literal_code, Optional<CanonicalCode>& distance_code)
```
alimpfard, line 354: This API feels weird to me, why is an `Optional<T>` passed by reference?
is this for optimisation purposes?
the function doesn't seem to be called in a tight loop, so returning an optional would probably work as well, if not better (RVO and all)
am I missing something?

asynts: Essentially, this function has two return values, the code used to encode literal values which is always present and the code that is used to encode distances which is only sometimes present.

alimpfard: Ah, both of those are out parameters.
I guess that's fine :shrug:

PR #3367, opened 2020-08-31
Kernel: Fix Processor::features_string() stopping too early and detect more features

Lubrsi:
The exit condition for the loop was sizeof(m_features) * 8,
which was 32. Presumably this was supposed to mean 32 bits, but it
actually made it stop as soon as it reached the 6th bit.

Also add detection for more SIMD CPU features.


tomuta:
Amazing! And for a while I was wondering why qemu cpu features seemingly never wanted to show up in the supported features log statement. Blamed it on myself thinking I must not be using the qemu command line option properly...

PR #3368, opened 2020-08-31
Userland: Allow unquoted 'filename' values in 'pro'

alimpfard:
- Applies a tip from https://github.com/SerenityOS/serenity/commit/cb7526fca0aa6cb07d279fec93288d0169904bfc#r41907965
- Tries to be a bit more lenient with `filename`

### DIFF
--- a/AK/GenericLexer.h
+++ b/AK/GenericLexer.h
@@ -75,6 +75,11 @@ protected:
     size_t m_index { 0 };
 };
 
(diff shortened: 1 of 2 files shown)


alimpfard:
Hmmm...might as well do it right...

alimpfard:
cc @benit8 for the `GenericLexer` change.

benit8:
Nice idea about the `is_any_of` :+1:

PR #3369, opened 2020-08-31
LibLine: Reset suggestion state on any non-tab key

alimpfard:
This fixes the following (and more!):
```sh
$ /bin/dis<tab><tab><backspace><backspace><backspace><backspace><tab>
$ /bink_benchmark
```

P.S. I hate the multi-key input changes, I should someday implement multi-key bindings to get rid of those too...
Not today though.

### DIFF
--- a/Libraries/LibLine/Editor.cpp
+++ b/Libraries/LibLine/Editor.cpp
@@ -27,6 +27,7 @@
 #include "Editor.h"
 #include <AK/GenericLexer.h>
 #include <AK/JsonObject.h>
+#include <AK/ScopeGuard.h>
 #include <AK/StringBuilder.h>
 #include <AK/Utf32View.h>
 #include <AK/Utf8View.h>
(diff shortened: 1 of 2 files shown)


awesomekling:
This looks very repetitive and hard to maintain. Can we reorganize things so that this can be RAII'd or something similar?

alimpfard:
Unless you have any ideas, I don't see a decent way to do RAII for this, just _one_ of the cases in one switch doesn't clear the state, so I can't just wrap it in a block...

alimpfard:
I guess I can pull that one offending case out, yeah, let's do that

Libraries/LibLine/Editor.cpp
```diff
@@ -721,11 +726,15 @@ void Editor::handle_read_event()
...
             if (code_point == 27) {
                 m_state = InputState::GotEscape;
+                cleanup_suggestions();
                 continue;
             }
             break;
         }
 
+        // There are no sequences past this point, so short of 'tab', we will want to cleanup the suggestions.
+        ArmedScopeGuard suggestion_cleanup { [this] { cleanup_suggestions(); } };
```
alimpfard, line 736: Top ten classes you didn't know existed:
one: This one.

Libraries/LibLine/Editor.cpp
```diff
@@ -653,14 +654,18 @@ void Editor::handle_read_event()
...
                 }
+                cleanup_suggestions();
                 continue;
             }
             }
         case InputState::GotEscapeFollowedByLeftBracket:
-            switch (code_point) {
-            case 'O': // mod_ctrl
+            if (code_point == 'O') {
+                // mod_ctrl
```
nico, line 663: why move this out of the switch?

nico: err nvm :)

PR #3372, opened 2020-09-01
Shell: Fix job control and backgrounding

alimpfard:
Let's fix some job control stuff

Things that this PR aims to fix:
- [x] Allowing more things in background
    - [x] And / Or
    - [x] Entire pipelines
    - [x] control structures
    - [x] subshells (they were free at this point)
- [x] Fixing job control stuff
    - [x] that weird duplication bug as seen in the latest update video
- [x] Resetting event loops across forks without upsetting any deities 
---

To make lazy execution possible (which is required for putting logic in the background), I'm gonna make nodes evaluate to thunks  in (what I know as) weak head-normal form, so the transition might be a bit ugly - if someone has better ideas, please do tell!

### DIFF
--- a/Libraries/LibCore/EventLoop.cpp
+++ b/Libraries/LibCore/EventLoop.cpp
@@ -519,6 +519,26 @@ void EventLoop::unregister_signal(int handler_id)
         s_signal_handlers.remove(remove_signo);
 }
 
+void EventLoop::notify_forked(ForkEvent event)
+{
+    switch (event) {
+    case ForkEvent::Child:
+        s_main_event_loop = nullptr;
+        s_event_loop_stack->clear();
+        s_timers->clear();
+        s_notifiers->clear();
+        s_signal_handlers.clear();
+        s_handling_signal = 0;
+        s_next_signal_id = 0;
+        s_pid = 0;
+        s_rpc_server = nullptr;
+        s_rpc_clients.clear();
+        return;
+    }
+
+    ASSERT_NOT_REACHED();
+}
+
 void EventLoop::wait_for_event(WaitMode mode)
 {
     fd_set rfds;
--- a/Base/usr/share/man/man5/Shell.md
+++ b/Base/usr/share/man/man5/Shell.md
@@ -122,7 +122,7 @@ Evaluate expressions take the general form of a dollar sign (`$`) followed by so
 
 ## Commands
 
-A `Command` is a single simple command, containing arguments and redirections for a single program. The shell can evaluate a sequence of commands, a conditional relation between commands, or various semantic elements composed of commands and intrinsics.
+A `Command` is a single simple command, containing arguments and redirections for a single program, or a compound command containing a shell control structure. The shell can evaluate a sequence of commands, a conditional relation between commands, or various semantic elements composed of commands and intrinsics.
 
 Commands can be either calls to Shell builtins, or external programs.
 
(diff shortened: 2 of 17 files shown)


awesomekling:
Please expand a bit on what's going on in the commit messages.
"Fix thing" followed by 100+ lines of diff is not very helpful :)

alimpfard:
Makes sense.
I shall try to remember what I did here

alimpfard:
a 100-line diff deserves an essay

Libraries/LibCore/EventLoop.cpp
```diff
@@ -519,6 +519,18 @@ void EventLoop::unregister_signal(int handler_id)
...
+    s_main_event_loop = nullptr;
+    s_event_loop_stack->clear();
+    s_timers->clear();
+    s_notifiers->clear();
+    s_signal_handlers.clear();
+    s_handling_signal = 0;
+    s_next_signal_id = 0;
+    s_pid = 0;
+}
+
```
alimpfard, line 533: I'm not so sure about this; it feels like a mess.
The main reason why I'm doing this is that signals are not delivered to forks, and I would like them to be delivered to forks (in specific conditions of course).
 
cc @tomuta (`s_pid` and signals and stuff), @awesomekling (this _is_ a `Core::EventLoop` change)

Any (better) alternative solutions?

tomuta: So the `s_pid` is basically only used to avoid accidently redirecting a signal sent to a fork to the parent. That's because the signal notification uses a pipe, which is inherited by the fork. Checking the current PID against `s_pid` when a signal is delivered allows it to detect whether it was forked and therefore can't send it through the existing pipe (the other end of that pipe is still read by the parent).

tomuta: If you expect the event loop to work in a fork you'd have to close the EventLoop's notification pipe and recreate it in the fork. Otherwise the event loop in the fork also going to notify the parent process, which is quite weird and probably going to result in odd behavior. For example, boost has a notify_fork method that is supposed to be called before and after fork, but due to the odd nature of fork it can deadlock (if multi-threaded). But its purpose is to recreate these types of pipes. But even boost isn't getting this right, because there is a window where a signal can be sent to the fork before notify_fork was called, and then the signal notification will unexpectedly leak to the parent as well. Hence the check against `s_pid` in the signal handler, which closes that loophole.

alimpfard: Well, I want to have a fresh event loop in the fork, no notifiers, no deferred invocations from the parent, etc.
I'm not against recreating the signal handlers.

So...so long as I recreate the signal pipe, everything signal-handler related is honky dory?

tomuta: Well, if you're not using threads then you might be able to re-create a new event loop. If you're using threads the nature of `fork` kind of dictates an `exec` (this is true even on Linux).
Last time I looked at the code (that was to add the `s_pid` fix) the `EventLoop` wasn't really designed with `fork` in mind. We'd have to add functionality similar to boost's notify_fork which, among other things, would recreate the notification pipe in a fork. But the `s_pid` check is still necessary due to the window between `fork` and `notify_fork`.

alimpfard: No threads here, so I guess we can call this okay and deal with `notify_fork()` and co when we need it...probably.

tomuta: If you only want to get the signal delivery working, you could create a new notification pipe when you detect the `s_pid` mismatch and use that from that point on. But anything else might be wonky, such as pending asynchronous stuff that might be carried into the fork. Which is why boost has the `notify_fork` API, which does the appropriate changes internally.

alimpfard: I decided to add a barebones `notify_forked()` that just tears down the event loop (renews pipes, clears timers, etc) in the child.
we can flesh it out or change it whenever we need the extra functionality.

tomuta:
@alimpfard You sure you want to push such a bad commit? https://github.com/SerenityOS/serenity/commit/bad81b843661f5df0d8a07afb0c750537d432433 ;-)

alimpfard:
lol its hash perfectly describes the contents!

awesomekling:
Looking at these changes, can you remind me what's the point of using a Core::EventLoop for the shell?
The way we're using it is so bizarre compared to all other users.

alimpfard:
> kling (or was it solid_black?) maybe we could have a fully event driven shell
> kling: what's the point of the event loop in the shell

I can think of three main uses:
- Line editing without blocking everything
- async job control: whenever a child dies/changes state, we get a notification, so we can do `block_on_job()` and treat it as async/await
- capturing $(...) output

awesomekling:
Just because I had a thought in the past doesn't mean I can remember why it seemed good! ;)
Thanks for reminding me.

alimpfard:
I personally think it's a nice abstraction, even if quite unusual

No problemo~

PR #3383, opened 2020-09-01
LibJS: IndexedProperties size stuff

linusg:
**LibJS: Let set_array_like_size() switch to generic storage if necessary**

This is already considered in `put()`/`insert()`/`append_all()` but not `set_array_like_size()`, which crashed the interpreter with an assertion when creating an array with more than `SPARSE_ARRAY_THRESHOLD` (200) initial elements as the simple storage was being resized beyond its limit.

**LibJS: Actually change size in generic storage's set_array_like_size()**

Looks like an oversight to me - we were not actually setting a new value for `m_array_size`, which would cause arrays created with generic storage to report a `.length` of 0.

### DIFF
--- a/Libraries/LibJS/Runtime/IndexedProperties.cpp
+++ b/Libraries/LibJS/Runtime/IndexedProperties.cpp
@@ -346,6 +346,13 @@ void IndexedProperties::append_all(Object* this_object, const IndexedProperties&
     }
 }
 
+void IndexedProperties::set_array_like_size(size_t new_size)
+{
+    if (m_storage->is_simple_storage() && new_size > SPARSE_ARRAY_THRESHOLD)
+        switch_to_generic_storage();
+    m_storage->set_array_like_size(new_size);
+}
+
 Vector<ValueAndAttributes> IndexedProperties::values_unordered() const
 {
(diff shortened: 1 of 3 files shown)


Libraries/LibJS/Runtime/IndexedProperties.cpp
```diff
@@ -346,6 +346,13 @@ void IndexedProperties::append_all(Object* this_object, const IndexedProperties&
     }
 }
 
+void IndexedProperties::set_array_like_size(size_t new_size)
+{
+    if (m_storage->is_simple_storage() && new_size > SPARSE_ARRAY_THRESHOLD)
```
awesomekling, line 352: When I search for `SPARSE_ARRAY_THRESHOLD` in the codebase, the mix of `<`, `<=` and `>=` makes it pretty rough to understand if the switch happens at 200 or 201 elements.

linusg: True :confused:

linusg: Mixing of zero-based indices and "one-based" sizes does not make it easier, too

PR #3396, opened 2020-09-03
Memory purging improvements

tomuta:
This branch aims to improve memory purging. Still very much work in progress

### DIFF
--- a/AK/Forward.h
+++ b/AK/Forward.h
@@ -112,6 +112,9 @@ template<typename T>
 class Optional;
 
 template<typename T>
+class RefPtrTraits;
+
+template<typename T, typename PtrTraits = RefPtrTraits<T>>
 class RefPtr;
 
 template<typename T>
--- a/AK/IntrusiveList.h
+++ b/AK/IntrusiveList.h
@@ -27,6 +27,7 @@
 #pragma once
 
 #include <AK/Assertions.h>
+#include <AK/Types.h>
 
 namespace AK {
 
@@ -56,6 +57,9 @@ public:
     T* last() const;
 
     T* take_first();
+    T* take_last();
+    
+    size_t size_slow() const;
 
     class Iterator {
     public:
@@ -231,6 +235,25 @@ inline T* IntrusiveList<T, member>::take_first()
         return ptr;
     }
     return nullptr;
+}
+
+template<class T, IntrusiveListNode T::*member>
+inline T* IntrusiveList<T, member>::take_last()
+{
+    if (auto* ptr = last()) {
+        remove(*ptr);
+        return ptr;
+    }
+    return nullptr;
+}
+
+template<class T, IntrusiveListNode T::*member>
+inline size_t IntrusiveList<T, member>::size_slow() const
+{
+    size_t count = 0;
+    for (auto* ptr = first(); ptr; ptr = next(ptr))
+        count++;
+    return count;
 }
 
 template<class T, IntrusiveListNode T::*member>
--- a/AK/NonnullOwnPtr.h
+++ b/AK/NonnullOwnPtr.h
@@ -35,7 +35,7 @@
 
 namespace AK {
 
-template<typename T>
+template<typename T, typename PtrTraits>
 class RefPtr;
 template<typename T>
 class NonnullRefPtr;
@@ -83,14 +83,14 @@ public:
     template<typename U>
     NonnullOwnPtr& operator=(const NonnullOwnPtr<U>&) = delete;
 
-    template<typename U>
-    NonnullOwnPtr(const RefPtr<U>&) = delete;
+    template<typename U, typename PtrTraits = RefPtrTraits<U>>
+    NonnullOwnPtr(const RefPtr<U, PtrTraits>&) = delete;
     template<typename U>
     NonnullOwnPtr(const NonnullRefPtr<U>&) = delete;
     template<typename U>
     NonnullOwnPtr(const WeakPtr<U>&) = delete;
-    template<typename U>
-    NonnullOwnPtr& operator=(const RefPtr<U>&) = delete;
+    template<typename U, typename PtrTraits = RefPtrTraits<U>>
+    NonnullOwnPtr& operator=(const RefPtr<U, PtrTraits>&) = delete;
     template<typename U>
     NonnullOwnPtr& operator=(const NonnullRefPtr<U>&) = delete;
     template<typename U>
--- a/AK/NonnullRefPtr.h
+++ b/AK/NonnullRefPtr.h
@@ -35,7 +35,7 @@ namespace AK {
 
 template<typename T>
 class OwnPtr;
-template<typename T>
+template<typename T, typename PtrTraits>
 class RefPtr;
 
 template<typename T>
--- a/AK/RefPtr.h
+++ b/AK/RefPtr.h
@@ -38,7 +38,31 @@ template<typename T>
 class OwnPtr;
 
 template<typename T>
+struct RefPtrTraits {
+    static T* as_ptr(FlatPtr bits)
+    {
+        return (T*)bits;
+    }
+
+    static FlatPtr as_bits(T* ptr)
+    {
+        return (FlatPtr)ptr;
+    }
+
+    static bool is_null(FlatPtr bits)
+    {
+        return !bits;
+    }
+
+    static constexpr FlatPtr default_null_value = 0;
+
+    typedef std::nullptr_t NullType;
+};
+
+template<typename T, typename PtrTraits>
 class RefPtr {
+    template<typename U, typename P>
+    friend class RefPtr;
 public:
     enum AdoptTag {
         Adopt
@@ -46,66 +70,71 @@ public:
 
     RefPtr() { }
     RefPtr(const T* ptr)
-        : m_ptr(const_cast<T*>(ptr))
+        : m_bits(PtrTraits::as_bits(const_cast<T*>(ptr)))
     {
-        ref_if_not_null(m_ptr);
+        ref_if_not_null(const_cast<T*>(ptr));
     }
     RefPtr(const T& object)
-        : m_ptr(const_cast<T*>(&object))
+        : m_bits(PtrTraits::as_bits(const_cast<T*>(&object)))
     {
-        m_ptr->ref();
+        T* ptr = const_cast<T*>(&object);
+        ASSERT(ptr);
+        ASSERT(!ptr == PtrTraits::is_null(m_bits));
+        ptr->ref();
     }
     RefPtr(AdoptTag, T& object)
-        : m_ptr(&object)
+        : m_bits(PtrTraits::as_bits(&object))
     {
+        ASSERT(&object);
+        ASSERT(!PtrTraits::is_null(m_bits));
     }
     RefPtr(RefPtr&& other)
-        : m_ptr(other.leak_ref())
+        : m_bits(other.leak_ref_raw())
     {
     }
     ALWAYS_INLINE RefPtr(const NonnullRefPtr<T>& other)
-        : m_ptr(const_cast<T*>(other.ptr()))
+        : m_bits(PtrTraits::as_bits(const_cast<T*>(other.ptr())))
     {
-        ASSERT(m_ptr);
-        m_ptr->ref();
+        ASSERT(!PtrTraits::is_null(m_bits));
+        PtrTraits::as_ptr(m_bits)->ref();
     }
     template<typename U>
     ALWAYS_INLINE RefPtr(const NonnullRefPtr<U>& other)
-        : m_ptr(const_cast<U*>(other.ptr()))
+        : m_bits(PtrTraits::as_bits(const_cast<U*>(other.ptr())))
     {
-        ASSERT(m_ptr);
-        m_ptr->ref();
+        ASSERT(!PtrTraits::is_null(m_bits));
+        PtrTraits::as_ptr(m_bits)->ref();
     }
     template<typename U>
     ALWAYS_INLINE RefPtr(NonnullRefPtr<U>&& other)
-        : m_ptr(&other.leak_ref())
+        : m_bits(PtrTraits::as_bits(&other.leak_ref()))
     {
-        ASSERT(m_ptr);
+        ASSERT(!PtrTraits::is_null(m_bits));
     }
-    template<typename U>
-    RefPtr(RefPtr<U>&& other)
-        : m_ptr(other.leak_ref())
+    template<typename U, typename P = RefPtrTraits<U>>
+    RefPtr(RefPtr<U, P>&& other)
+        : m_bits(other.leak_ref_raw())
     {
     }
     RefPtr(const RefPtr& other)
-        : m_ptr(const_cast<T*>(other.ptr()))
+        : m_bits(PtrTraits::as_bits(const_cast<T*>(other.ptr())))
     {
-        ref_if_not_null(m_ptr);
+        ref_if_not_null(const_cast<T*>(other.ptr()));
     }
-    template<typename U>
-    RefPtr(const RefPtr<U>& other)
-        : m_ptr(const_cast<U*>(other.ptr()))
+    template<typename U, typename P = RefPtrTraits<U>>
+    RefPtr(const RefPtr<U, P>& other)
+        : m_bits(PtrTraits::as_bits(const_cast<U*>(other.ptr())))
     {
-        ref_if_not_null(m_ptr);
+        ref_if_not_null(const_cast<U*>(other.ptr()));
     }
     ALWAYS_INLINE ~RefPtr()
     {
         clear();
 #ifdef SANITIZE_PTRS
         if constexpr (sizeof(T*) == 8)
-            m_ptr = (T*)(0xe0e0e0e0e0e0e0e0);
+            m_bits = 0xe0e0e0e0e0e0e0e0;
         else
-            m_ptr = (T*)(0xe0e0e0e0);
+            m_bits = 0xe0e0e0e0;
 #endif
     }
     RefPtr(std::nullptr_t) { }
@@ -116,9 +145,9 @@ public:
     RefPtr& operator=(const OwnPtr<U>&) = delete;
 
     template<typename U>
-    void swap(RefPtr<U>& other)
+    void swap(RefPtr<U, PtrTraits>& other)
     {
-        ::swap(m_ptr, other.m_ptr);
+        ::swap(m_bits, other.m_bits);
     }
 
     ALWAYS_INLINE RefPtr& operator=(RefPtr&& other)
@@ -129,7 +158,7 @@ public:
     }
 
     template<typename U>
-    ALWAYS_INLINE RefPtr& operator=(RefPtr<U>&& other)
+    ALWAYS_INLINE RefPtr& operator=(RefPtr<U, PtrTraits>&& other)
     {
         RefPtr tmp = move(other);
         swap(tmp);
@@ -141,7 +170,7 @@ public:
     {
         RefPtr tmp = move(other);
         swap(tmp);
-        ASSERT(m_ptr);
+        ASSERT(!PtrTraits::is_null(m_bits));
         return *this;
     }
 
@@ -149,7 +178,7 @@ public:
     {
         RefPtr tmp = other;
         swap(tmp);
-        ASSERT(m_ptr);
+        ASSERT(!PtrTraits::is_null(m_bits));
         return *this;
     }
 
@@ -158,7 +187,7 @@ public:
     {
         RefPtr tmp = other;
         swap(tmp);
-        ASSERT(m_ptr);
+        ASSERT(!PtrTraits::is_null(m_bits));
         return *this;
     }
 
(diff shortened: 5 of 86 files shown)


awesomekling:
This looks like it's trying to make partial regions volatile/non-volatile. When is that feature useful in practice?
Our purgeable memory is modeled after XNU, where volatility is also per-object, not per-page.

tomuta:
Yeah it does. I kind of went down this path trying to figure out why low memory conditions kept crashing random processes where pages were seemingly purged that shouldn't have been. It seemed odd to me that `madvise` would flip the entire vmobject volatile. Especially if shared between processes. What if one says make it non-volatile, assumes it can read/write to it, and then the other process turns it volatile, which then would trigger the other process to crash unexpectedly if those pages get purged.

This feature is useful for large mappings (think databases, large heaps). I believe Windows uses this for heaps where they reserve a relatively large virtual address space but only commit as needed. Linux supports this as well, IIRC. Our `malloc` just allocates more regions as needed, which works just fine as well.

These changes don't quite solve any problem yet. Again, I thought the issue was that by flipping entire vmobjects that are shared between processes to volatile by one side, causes the other side to unexpectedly crash if those pages get purged. That was the issue I was trying to solve, and in the process I also thought it was odd that one couldn't mark only portions as volatile/non-volatile.

awesomekling:
> Yeah it does. I kind of went down this path trying to figure out why low memory conditions kept crashing random processes where pages were seemingly purged that shouldn't have been. It seemed odd to me that `madvise` would flip the entire vmobject volatile. Especially if shared between processes. What if one says make it non-volatile, assumes it can read/write to it, and then the other process turns it volatile, which then would trigger the other process to crash unexpectedly if those pages get purged.

Yeah, it's up to the sharing processes to coordinate volatility among themselves. If you're seeing crashes like that, it's a bug in the coordination, not in the kernel :)

> This feature is useful for large mappings (think databases, large heaps). I believe Windows uses this for heaps where they reserve a relatively large virtual address space but only commit as needed. Linux supports this as well, IIRC. Our `malloc` just allocates more regions as needed, which works just fine as well.

Yes indeed, I've had something like XNU's `MADV_FREE_*` in the back of my mind for a while but never started on it. I agree that it would be nice to have something like this for better coordination between `malloc` and kernel. (It also annoys me that we do two syscalls (`madvise` + `mprotect`) in `malloc` right now, would be cool if a single `madvise` took care of it.)

> These changes don't quite solve any problem yet. Again, I thought the issue was that by flipping entire vmobjects that are shared between processes to volatile by one side, causes the other side to unexpectedly crash if those pages get purged. That was the issue I was trying to solve, and in the process I also thought it was odd that one couldn't mark only portions as volatile/non-volatile.

My longer-term idea with purgeable memory was to add priority values/queues and have the kernel be smarter about which object gets purged. I went down the `PurgeableVMObject` route as it seemed like the easiest path towards adding this feature later on.

I'm fine with merging the concepts if you think it still be easy and logical to add priorities. I'm not married to the XNU style full objects if there's a better way :)

tomuta:
Well, priorities would be nice as well. However, there shouldn't be any crashes even if all memory got purged instantly as soon as marking it volatile. Maybe that would be a nice `#ifdef` to catch bugs...

tomuta:
How would processes communicate volatility if the kernel doesn't implicitly account for it? How could it guarantee that the other process won't be using what it thinks should be available without a whole bunch of negotiation essentially, where both sides need to sync up and wait on each other before making changes to the shared vmobject?

tomuta:
Specific use case for this would be e.g. GUI::Window and how it changes backing bitmaps volatile/nonvolatile at will.

awesomekling:
> Specific use case for this would be e.g. GUI::Window and how it changes backing bitmaps volatile/nonvolatile at will.

The basic idea there is that we synchronize around the around`SetWindowBackingStore` IPC message.
WindowServer knows that after he responds to it, the window's *previous* backing store may become volatile at any time.

tomuta:
Hm. Well I added some better logging to figure out what exactly is being purged, here's e.g. a crash (there are many different variants) of WindowServer:
```
[allocate(25:25)]: kmalloc(): Adding memory to heap at V0xc0801000, bytes: 1048576
[allocate(25:25)]: Purged 16 pages from region malloc: ChunkedBlock(32) owned by SystemServer(5) at V0x02410000 - V0x02420000
[allocate(25:25)]: MM: Purge saved the day! Purged 16 pages from PurgeableVMObject{0xc044c598}
[allocate(25:25)]: Purged 15 pages from region malloc: ChunkedBlock(4090) owned by SystemServer(5) at V0x02430000 - V0x02440000
[allocate(25:25)]: MM: Purge saved the day! Purged 15 pages from PurgeableVMObject{0xc044c738}
[allocate(25:25)]: Purged 16 pages from region malloc: ChunkedBlock(8) owned by SystemServer(5) at V0x024b0000 - V0x024c0000
[allocate(25:25)]: MM: Purge saved the day! Purged 16 pages from PurgeableVMObject{0xc044c9d8}
[allocate(25:25)]: Purged 16 pages from region malloc: ChunkedBlock(2036) owned by SystemServer(5) at V0x024d0000 - V0x024e0000
[allocate(25:25)]: MM: Purge saved the day! Purged 16 pages from PurgeableVMObject{0xc044c538}
[allocate(25:25)]: Purged 16 pages from region malloc: ChunkedBlock(1016) owned by SystemServer(5) at V0x024f0000 - V0x02500000
[allocate(25:25)]: MM: Purge saved the day! Purged 16 pages from PurgeableVMObject{0xc0468298}
[allocate(25:25)]: Purged 16 pages from region malloc: ChunkedBlock(508) owned by SystemServer(5) at V0x02510000 - V0x02520000
[allocate(25:25)]: MM: Purge saved the day! Purged 16 pages from PurgeableVMObject{0xc0468398}
[allocate(25:25)]: Purged 15 pages from region malloc: ChunkedBlock(4090) owned by WindowServer(14) at V0x00c10000 - V0x00c20000
[allocate(25:25)]: MM: Purge saved the day! Purged 15 pages from PurgeableVMObject{0xc047f7f8}
[allocate(25:25)]: Purged 15 pages from region malloc: ChunkedBlock(4090) owned by AudioServer(15) at V0x01210000 - V0x01220000
[allocate(25:25)]: MM: Purge saved the day! Purged 15 pages from PurgeableVMObject{0xc04822f8}
[allocate(25:25)]: Purged 16 pages from region malloc: ChunkedBlock(8) owned by AudioServer(15) at V0x01230000 - V0x01240000
[allocate(25:25)]: MM: Purge saved the day! Purged 16 pages from PurgeableVMObject{0xc0482478}
[WindowServer(14:14)]: CPU[0] NP(error) fault at invalid address V0x000008d2
[WindowServer(14:14)]: Unrecoverable page fault, read from address V0x000008d2
[WindowServer(14:14)]: Note: Address V0x000008d2 looks like a possible nullptr dereference
[WindowServer(14:14)]: CRASH: CPU #0 Page Fault. Ring 3.
[WindowServer(14:14)]: exception code: 0004 (isr: 0000
[WindowServer(14:14)]:   pc=001b:0806502e flags=0206
[WindowServer(14:14)]:  stk=0023:00c01b30
[WindowServer(14:14)]:   ds=0023 es=0023 fs=0023 gs=002b
[WindowServer(14:14)]: eax=00c01ba0 ebx=00c1200c ecx=00000004 edx=00000000
[WindowServer(14:14)]: ebp=00c01b48 esp=c34a5fe8 esi=00000000 edi=00c01b30
[WindowServer(14:14)]: cr0=80010013 cr2=000008d2 cr3=00d47000 cr4=00300ee4
[WindowServer(14:14)]: code: 80 ba d2 08 00 00 00 8d

...

[WindowServer(14:14)]: 0x0806502e  WindowServer::WindowFrame::rect() const +24

[WindowServer(14:14)]: 0xc0162915  Kernel::Process::crash(int, unsigned int, bool) +1041
[WindowServer(14:14)]: 0xc01209d9  illegal_instruction_handler +0
[WindowServer(14:14)]: 0xc0122013  page_fault_handler +972
[WindowServer(14:14)]: 0xc011e2cb  page_fault_asm_entry +38
[WindowServer(14:14)]: 0x08056fd7  WindowServer::Compositor::run_animations(Gfx::DisjointRectSet&) +175
[WindowServer(14:14)]: 0x0805b583  WindowServer::Compositor::compose() +3439
[WindowServer(14:14)]: 0x08077b18  Core::Object::dispatch_event(Core::Event&, Core::Object*) +74
[WindowServer(14:14)]: 0x08070b46  Core::EventLoop::pump(Core::EventLoop::WaitMode) +308
[WindowServer(14:14)]: 0x08070cc5  Core::EventLoop::exec() +43
[WindowServer(14:14)]: 0x08048971  main +1321
[WindowServer(14:14)]: 0x0804904e  _start +94
```
Looks like it's purging the heap of several applications. Wonder why, unless it's purging heap areas that they marked as volatile because they're not in use, but then why would they use it...

tomuta:
Ah I think that was a stupid mistake by myself, I changed the behavior of `mmap` to make it volatile entirely if `MAP_PURGEABLE` was set...

tomuta:
Hm this looks promising. After fixing this one small mistake, things change to this behavior: If `allocate` can't allocate enough itself, it generally crashes itself. If it can and you try to e.g. start another application and that other application can't acquire enough memory, the other application will crash. I did see a crash in ProtocolServer when trying to start Browser without enough memory, but I'm not sure if that is started on demand or if that always hangs around. After allocate exited I as able to start Browser again and things loaded fine. But it looks like it no longer takes a whole bunch of random processes down. I added some log statements to `malloc` and it seems like various processes that had memory purged do correctly detect that those heap regions were purged and they continue on without crashing.

tomuta:
Hm well maybe I just got lucky. This time while Browser crashed as expected, I can no longer launch new applications. But the task bar is still responding. Odd...

tomuta:
Hm just ran into this:
```
[FileManager (Desktop)(9:9)]: Purged 16 pages from region malloc: ChunkedBlock(2036) owned by ResourceGraph.MenuApplet(27) at V0x027b0000 - V0x027c0000
[FileManager (Desktop)(9:9)]: MM: Purge saved the day! Purged 16 pages from PurgeableVMObject{0xc04dae58}
[FileManager (Desktop)(9:9)]: MM: no user physical pages available
[FileManager (Desktop)(9:9)]: MM: handle_zero_fault was unable to allocate a physical page
[FileManager (Desktop)(9:9)]: Unrecoverable page fault, write to address V0xc1258000

#0  Kernel::handle_crash () at ../Kernel/Arch/i386/CPU.cpp:160
#1  0xc0122013 in page_fault_handler () at ../Kernel/Arch/i386/CPU.cpp:284
#2  0xc011e2cb in page_fault_asm_entry ()
#3  0xc344fa6c in ?? ()
#4  0xc0129cec in Kernel::PATADiskDevice::read_sectors_with_dma ()
    at ../Kernel/Devices/PATADiskDevice.cpp:172
#5  0xc0129ec2 in Kernel::PATADiskDevice::read ()
    at ../Kernel/Devices/PATADiskDevice.cpp:103
#6  0xc01336fb in Kernel::FileDescription::read ()
    at ../Kernel/FileSystem/FileDescription.cpp:136
#7  0xc012c70e in Kernel::BlockBasedFS::read_block ()
    at ../Kernel/FileSystem/BlockBasedFileSystem.cpp:226
#8  0xc0130400 in Kernel::Ext2FSInode::read_bytes ()
    at ../Kernel/FileSystem/Ext2FileSystem.cpp:711
#9  0xc0190266 in Kernel::Region::handle_inode_fault ()
    at ../Kernel/VM/Region.cpp:590
#10 0xc018b3cf in Kernel::MemoryManager::handle_page_fault ()
    at ../Kernel/VM/MemoryManager.cpp:380
#11 0xc0121d59 in page_fault_handler () at ../Kernel/Arch/i386/CPU.cpp:247
#12 0xc011e2cb in page_fault_asm_entry ()
#13 0xc3450fac in ?? ()
```
I'm guessing we should commit (remap) before doing DMAs to make sure enough pages actually exist (and gracefully fail if we can't commit them all). Lazily allocating them seems a little dangerous there?

awesomekling:
The DMA page used by in PATAChannel is not purgeable nor lazily allocated. It's allocated up front in `PATAChannel::initialize()`

We could commit a physical page before doing the disk access, although I don't think it would achieve anything besides failing earlier (and failing without accessing the disk, which I suppose is nice.)

tomuta:
Oh I thought this was a hard kernel crash... ;-) Never mind

tomuta:
I think overall this makes the system behave better in low memory conditions. It doesn't solve all the issues, but I think this is a step forward at least.

tomuta:
I think I found one more improvement that might make things even better: when `madvise` requests memory to become nonvolatile, we right now call into `Region::remap_page_range` which basically ensures that all page tables at least exist, but I think it probably should also attempt to turn all zero pages into actual pages so that they're guaranteed to be available. I'm not sure what other systems do in that regard. It feels like by default memory should always be fully allocated unless specifically requesting otherwise?

awesomekling:
> I think I found one more improvement that might make things even better: when `madvise` requests memory to become nonvolatile, we right now call into `Region::remap_page_range` which basically ensures that all page tables at least exist, but I think it probably should also attempt to turn all zero pages into actual pages so that they're guaranteed to be available. I'm not sure what other systems do in that regard. It feels like by default memory should always be fully allocated unless specifically requesting otherwise?
> 
> I wonder if this is what Linux has `MAP_NORESERVE` for, meaning if `MAP_NORESERVE` is set then reading/writing could fail if no pages are available?

Sounds like the age-old overcommit debate. Serenity currently overcommits (i.e allows you to allocate as much VM as you like, and will lazily populate it with physical pages as needed. This is beneficial for workloads that overrequest and underutilize.) 

Reserving physical pages up front is generally more expensive since you always do the extra work of allocating pages, but the benefit is that you can fail early instead of failing lazily and making the OOM killer clean up the mess.

I'm not religiously against overcommit like some people, but I do see the niceness of not *relying* on the OOM killer.

Moving towards eager page reservation seems fine. But let's defer zeroing out memory until the first fault. :)

tomuta:
Well, yeah, the problem that we have right now is that random stuff crashes. I think the best of both worlds is a hybrid:
Change mmap to allocate lazily when `MAP_PURGEABLE` is requested, and require the caller (e.g. malloc) to call `madvise` and force allocation of how much it actually needs. Right now it commits everything (though also lazily, which is creating these random crashes). This way you could create huge heaps but then also request how much you actually need, and that memory would be guaranteed to be available.

Also, we could have a zero-pages task that zeros available physical pages in the background. Then we'd try to pull from a pool of already zero'ed pages, and only if that pool is empty fall back to zeroing empty pages immediately.

tomuta:
Actually, I think just adding `MAP_NORESERVE` in addition to `MAP_PURGEABLE` when calling `mmap` could solve this, then you could request a huge mapping but we wouldn't commit the entire block unless you subsequently call `madvise` where you then request memory to be committed for whatever portion should be guaranteed to be available.

tomuta:
With these changes we can allow over-committing (by specifying `MAP_NORESERVE` for `mmap`), but otherwise memory will be committed. Otherwise random crashes are unavoidable, because without these changes there is no way for an application to check or ensure that memory is actually available before using it, which is why random processes crashed when memory has to be purged. Not because their memory was purged, but because allocated (but not yet committed) memory could no longer be committed because there were still not enough pages available despite trying to purge more. Now e.g. the application heaps get fully committed while still leaving it up to the application to mark unused sub-heaps as volatile, allowing them to be purged. Then, when `malloc` wants to re-use those purged subheaps, it uses `madvise` and at that point it will attempt to commit that memory, and if that fails `malloc` will get an error back, allowing it to gracefully handle the situation.

tomuta:
So now that stuff no longer randomly crashes, processes seem to like to get stuck waiting on the FS lock, and then WindowManager shows that they're not responding:
```
Queued       SystemMonitor(24:24) @ 0008:c011f3c1 Reason: FS
Queued       Terminal(8:8) @ 0008:c011f3c1 Reason: FS
```
Not sure why, though

tomuta:
> So now that stuff no longer randomly crashes, processes seem to like to get stuck waiting on the FS lock, and then WindowManager shows that they're not responding:
> 
> ```
> Queued       SystemMonitor(24:24) @ 0008:c011f3c1 Reason: FS
> Queued       Terminal(8:8) @ 0008:c011f3c1 Reason: FS
> ```
> 
> Not sure why, though

I don't think that's really related to these changes. Merely adding [these log statements](https://github.com/tomuta/serenity/commit/68eb255e091b9ecf8e92451a2878661f152d2067) to `master` also causes the system to hang in weird ways...

awesomekling:
I think this is heading in a muddy direction due to partially overlapping use-cases.

How about this:
* We make a separate syscall API for purgeable objects (e.g `sys$purgeable_alloc`, `sys$purgeable_dealloc` and `sys$purgeable_ctl`). These operate on full regions backed by `PurgeableVMObject`, and they're not visible to `mmap` related APIs.
* We allow partial control of physical page commitments through APIs like the ones you're adding in this PR. This can be used by malloc to frontload and optimize physical memory utilization as needed.

I think trying to piggyback on `mmap` for purgeable allocations was a mistake. Using them for `malloc` was another. From an API perspective, purgeable objects are intended to be "a costly thing that you could reconstitute if necessary", not a mechanism for managing page-level physical memory. :)

tomuta:
I don't think this is the wrong direction. I am pretty sure this is standard behavior for most OS, including Linux. If you allocate a region of virtual memory, unless you specify otherwise (`MAP_NORESERVE` on Linux and with this) you should get a region of committed storage so that you can safely use it. If you merely want to allocate virtual address space you'd simply use this flag (`MAP_NORESERVE`) and then mark page ranges as nonvolatile as needed (at which point they will be committed), or drop then by marking them as volatile (which allows them to be purged, if necessary). This is also how Windows Virtual Address APIs work.

Unless we match this standard behavior for `mmap` this would mean ported apps that expect this standard behavior would have to be further modified.

To be honest, I don't see much wrong with this approach. If anything we could drop `MAP_PURGEABLE` because `MAP_NORESERVE` is kind of the same thing, but it's more explicit. Right now `MAP_PURGEABLE` actually doesn't create purgeable memory, the entire region is non-volatile, yet lazily allocated, which quite weird because on one hand memory may not be purged but on the other it may not be available as it wasn't committed. It's impossible to use it properly as-is. You can't write code that can't crash in an unexpected way.

I would advise against more system calls, I think `mmap` as it is today is broken. To me the best approach would be to actually drop the `MAP_PURGEABLE` flag and go with Linux's `MAP_NORESERVE`, and keep the behavior as in this pull request. I think the `malloc` implementation as-is is in no immediate need for changing (other than dropping `MAP_PURGEABLE`). There's nothing wrong with it. It already marks regions as volatile/nonvolatile as needed.

tomuta:
The `malloc` implementation is just fine, which is why applications no longer randomly crash accessing memory with these changes, despite me not even touching it. It would work just fine on Linux using `mmap` (obviously with the correct mapping of flags), or on Windows, even in low-memory situations.

tomuta:
The `Gfx::Bitmap` code looks even more standard after removing `MAP_PURGEABLE`. And by using `PurgeableVMObject` (which isn't actually purgeable by default, so same as `AnonymousVMObject`) in `Process::allocate_region` the `mmap` behavior makes a whole lot more sense. This also solves the problems where accessing TLS could crash processes when the underlying storage can't be accessed.

I think this is quite a clean approach that makes sense (which is very similar to Windows/Linux at least).

tomuta:
Also added a fix for a heap expansion bug I found by fat fingering `allocate 2109`...

tomuta:
More enhancements are in the works, including the requested lazy-allocating committed/reserved pages. which does make a noticeable difference in terms of performance.

awesomekling:
> I don't think this is the wrong direction. I am pretty sure this is standard behavior for most OS, including Linux. If you allocate a region of virtual memory, unless you specify otherwise (`MAP_NORESERVE` on Linux and with this) you should get a region of committed storage so that you can safely use it. If you merely want to allocate virtual address space you'd simply use this flag (`MAP_NORESERVE`) and then mark page ranges as nonvolatile as needed (at which point they will be committed), or drop then by marking them as volatile (which allows them to be purged, if necessary). This is also how Windows Virtual Address APIs work.

The Linux man pages say that `MAP_NORESERVE` reserve *swap* space, not physical RAM pages. That seems different from what you're describing here. But maybe I'm misunderstanding something, I haven't looked at the implementation.

> Unless we match this standard behavior for `mmap` this would mean ported apps that expect this standard behavior would have to be further modified.

My understanding is that you're trying to implement `MADV_DONTNEED` and you think the purgeable memory feature is a broken implementation of that. Again, purgeable memory is not the same as `MADV_DONTNEED`. This is why I say that it was a mistake to use the `mmap` / `madvise` syscalls for purgeable memory, and that it should be moved to its own special syscalls. The purpose of purgeable memory is to keep the kernel clued in about the current relative necessity of specific large allocations, not to micromanage physical pages of arbitrary `mmap` allocations.

I don't mind bringing `mmap` closer to the behavior of other systems, but let's not conflate unrelated concepts.

tomuta:
In regards to "reserving swap space), this implementation will effectively work the same as on a Linux system that doesn't have swap configured.

My main motivation here is to solve random crashes in low-memory situations. Based on my debugging of the many crashes I've run into it's always come down to the current implementations being a bit broken by design. It's not possible to write applications that won't randomly crash, because regions/vmobjects don't ensure available backing (physical) memory in any case.

The changes I have in works are effectively these:
- Remove `PurgeableVMObject` and merge it into `AnonymousVMObject`. That's because `PurgaebleVMObject` today already isn't actually purgeable by default anyway. By merging them, all `AnonymousVMObject` become purgeable (though also not by default), allowing any regions to be marked volatile with `madvise`.
- Ensure we can reserve enough physical pages unless `MAP_NORESERVE` is specified. Once we support swap, the available pages that can be reserved would grow by the available swap space (which is why it's advisable to always configure swap on Linux, though it may be rarely ever used.
- Before cloning a region/vmobject, we need to make sure we can reserve sufficient physical pages as well (depending on non-volatile areas) because all pages will be marked as COW. This is especially important for the process forking, otherwise a `fork` could trigger the parent to crash even if the parent had all pages marked nonvolatile. For this purpose, the number of nonvolatile pages would get reserved and that pool would be shared between the original and cloned region because at most if every page in both were to be COW'ed the physical memory requirements would be double (for the nonvolatile regions). A COW in a volatile region would not draw from that pool. If we can't reserve enough prior to cloning it, `fork` should fail with `ENOMEM`.

We can tweak the over-commitment rules, but in general I would only add risk for a fork but not the parent. Otherwise you can't guarantee that even important services like WindowServer randomly crash (yes I've had that happen).

tomuta:
If my theory is right, once I get all this working applications should no longer crash other applications or services when they exhaust available memory.

awesomekling:
> That's because `PurgeableVMObject` today already isn't actually purgeable by default anyway.

"Purgeable" is a class of memory, not a state it can be in. The state is volatile/nonvolatile. Whether it is backed by physical pages or not is unrelated.

> By merging them, all `AnonymousVMObject` become purgeable (though also not by default), allowing any regions to be marked volatile with `madvise`.

Yet again, we should not conflate the purgeable memory class with arbitrary anonymous memory regions.

> * Ensure we can reserve enough physical pages unless `MAP_NORESERVE` is specified.

That seems fine.

> Once we support swap, the available pages that can be reserved would grow by the available swap space (which is why it's advisable to always configure swap on Linux, though it may be rarely ever used.

I'm not sure swap (to disk) support is worthwhile given the complexity/benefit ratio. A zlib (or similar) compression pager for cold pages might be. Then again maybe disk paging is really simple and we should just do it. I don't know, I have never implemented it. :)

> * Before cloning a region/vmobject, we need to make sure we can reserve sufficient physical pages as well (depending on non-volatile areas) because all pages will be marked as COW. This is especially important for the process forking, otherwise a `fork` could trigger the parent to crash even if the parent had all pages marked nonvolatile. For this purpose, the number of nonvolatile pages would get reserved and that pool would be shared between the original and cloned region because at most if every page in both were to be COW'ed the physical memory requirements would be double (for the nonvolatile regions). A COW in a volatile region would not draw from that pool. If we can't reserve enough prior to cloning it, `fork` should fail with `ENOMEM`. A clone of a clone would do the same thing, but replace it's own shared pool with the new one and leave the original region's pool in place.

This seems fine too.

At the same time, I think we can also get rid of many/most COW mappings by implementing something like `vfork()` (or simply moving `posix_spawn()` into the kernel.)

> We can tweak the over-commitment rules, but in general I would only add risk for a fork but not the parent. Otherwise you can't guarantee that even important services like WindowServer won't randomly crash (yes I've had that happen).

Agreed, the child should absorb all the risk of failure.

tomuta:
I understand your point, but the "purgeable" class of memory *is* effectively just "anonymous" memory. Which is why you won't find that distinction e.g. in Linux or Windows, it's all just "anonymous" or "virtual" memory. Whether it's purgeable or not merely depends on whether it is marked as volatile or not. Ironically, `PurgeableVMObject` as it is today isn't even purgeable by default. These changes basically remove that distinction that doesn't really exist anyway.

To create actually purgeable memory:
1. Without this change: call `mmap(... MAP_PURGEABLE)` and `madvise(... MADV_SET_VOLATILE)`
2. With this change: call `mmap(... MAP_NORESERVE)`  and `madvise(... MADV_SET_VOLATILE)`

To create non-purgeable memory (e.g. for heaps, stacks, tls, ... or anything else that shouldn't be wiped out without knowing about it):
1. Without this change: impossible, it may not be purged but there may not be enough memory to allocate as pages are accessed
2. With this change: call `mmap` without `MAP_NORESERVE`. You can still make it purgeable later by calling `madvise(... MADV_SET_VOLATILE)`, or make regions guaranteed to be available again by calling `madvise(... MADV_SET_NONVOLATILE)`

And yes, a `vfork` would be useful, but in order to remain somewhat portable I don't think we should get rid of `fork` altogether. `vfork` would be very useful when trying to `fork`/`exec` from a large process (e.g. Browser), and with these changes it would actually be quite easy to implement. It would allow to potentially execute a small new process if we're tight on memory already.

awesomekling:
Yeah yeah all right. As long as `madvise(MADV_SET_NONVOLATILE)` returns informs the caller whether the *entire* range is still intact, I guess it doesn't matter.

tomuta:
> Yeah yeah all right. As long as `madvise(MADV_SET_NONVOLATILE)` returns informs the caller whether the _entire_ range is still intact, I guess it doesn't matter.

Yeah, this implementation not only informs the caller whether any memory from that range had been purged. It keeps track of that, so if something outside of the range gets purged but not in the range to be marked nonvolatile then it won't say that anything was purged. Also, requesting a nonvolatile range will fail if the system is not able to reserve the amount of pages needed for that range, so if the call succeeds the caller is guaranteed to have that memory available (either backed by physical memory, or down the road worst case by swapping).

tomuta:
I'm still working on refining some aspects of these changes, but as they are here in this pull request, they no longer cause odd crashes. Well, other bugs still do, e.g. procfs causes some menu applets to crash because KBuffer fails to create memory regions. I'm working on these parts right now, basically removing the need for KBuffer and being able to safely write directly to user buffers (essentially special page fault handling for `copy_to_user`/`copy_from_user`, the current approach is not really safe in multi-processor cases anyway)

tomuta:
One downside is that we probably need to up the default ram size for qemu to maybe 512 mb, because we don't support swap space and we don't support shared libraries (which wastes a lot of memory)...

tomuta:
I didn't expect so many other problems preventing the system from being somewhat stable once running out of resources. At least I solved the ResourceGraph crashes (which was a combination of ProcFS handling low memory poorly and ResourceGraph assuming it can always get the data it wants). Now it shows a nice red bar if the system is hopelessly low on memory and it is unable to read from /proc ;-)

Screenshot from 2020-09-18 19-35-36

I suppose one issue at a time...

PR #3398, opened 2020-09-04
First few fuzzing fixes

BenWiederhake:
This PR contains the first few results of my fuzzing attempts:
- Some more documentation and warning flag fixes
- LibJS: The parser used to sometimes report outrageous positions such as `line: 2, column: 4294967292`, which can lead to memory exhaustion when trying to generate a `source_location_hint`.
- LibGfx: We used to accept images out outrageous sizes – note that 16384×16384×32bit would mean a 1GiB allocation. If any program wants to handle even larger files, it'll have to use a different approach anyway, to avoid memory exhaustion.
- BMP loader: We used to blindly trust a file-internal offset, leading to segmentation fault.

I'm still fuzzing the BMP code, there still are crashes. However, it'll be a while until my next PR, that's why I'm trying to push this.

If you're interested in fuzzing Serenity, I would appreciate if you don't also focus on BMP, or else we just duplicate our efforts needlessly. :P

### DIFF
--- a/Applications/PixelPaint/BucketTool.cpp
+++ b/Applications/PixelPaint/BucketTool.cpp
@@ -58,10 +58,10 @@ static void flood_fill(Gfx::Bitmap& bitmap, const Gfx::IntPoint& start_position,
     while (!queue.is_empty()) {
         auto position = queue.dequeue();
 
-        if (bitmap.get_pixel<Gfx::BitmapFormat::RGBA32>(position.x(), position.y()) != target_color)
+        if (bitmap.get_pixel<Gfx::StorageFormat::RGBA32>(position.x(), position.y()) != target_color)
             continue;
 
-        bitmap.set_pixel<Gfx::BitmapFormat::RGBA32>(position.x(), position.y(), fill_color);
+        bitmap.set_pixel<Gfx::StorageFormat::RGBA32>(position.x(), position.y(), fill_color);
 
         if (position.x() != 0)
             queue.enqueue(position.translated(-1, 0));
--- a/Libraries/LibGfx/Bitmap.cpp
+++ b/Libraries/LibGfx/Bitmap.cpp
@@ -44,11 +44,16 @@
 
 namespace Gfx {
 
-static bool size_would_overflow(BitmapFormat format, const IntSize& size)
+static bool size_would_overflow(BitmapFormat, const IntSize& size)
 {
     if (size.width() < 0 || size.height() < 0)
         return true;
-    return Checked<size_t>::multiplication_would_overflow(size.width(), size.height(), Bitmap::bpp_for_format(format));
+    // This check is a bit arbitrary, but should protect us from most shenanigans:
+    if (size.width() >= 32768 || size.height() >= 32768)
+        return true;
+    // This check is absolutely necessary. Note that Bitmap::Bitmap always stores
+    // data as RGBA32 internally, so currently we ignore the indicated format.
+    return Checked<size_t>::multiplication_would_overflow(size.width(), size.height(), sizeof(RGBA32));
 }
 
 RefPtr<Bitmap> Bitmap::create(BitmapFormat format, const IntSize& size)
--- a/Libraries/LibGfx/Bitmap.cpp
+++ b/Libraries/LibGfx/Bitmap.cpp
@@ -145,15 +145,25 @@ Bitmap::Bitmap(BitmapFormat format, const IntSize& size, size_t pitch, void* dat
 
 RefPtr<Bitmap> Bitmap::create_with_shared_buffer(BitmapFormat format, NonnullRefPtr<SharedBuffer>&& shared_buffer, const IntSize& size)
 {
-    if (size_would_overflow(format, size))
-        return nullptr;
-    return adopt(*new Bitmap(format, move(shared_buffer), size, {}));
+    return create_with_shared_buffer(format, move(shared_buffer), size, {});
 }
 
 RefPtr<Bitmap> Bitmap::create_with_shared_buffer(BitmapFormat format, NonnullRefPtr<SharedBuffer>&& shared_buffer, const IntSize& size, const Vector<RGBA32>& palette)
 {
     if (size_would_overflow(format, size))
         return nullptr;
+
+    unsigned actual_size = shared_buffer->size();
+    // FIXME: Code duplication of size_in_bytes() and m_pitch
+    unsigned expected_size_min = minimum_pitch(size.width(), format) * size.height();
+    unsigned expected_size_max = round_up_to_power_of_two(expected_size_min, PAGE_SIZE);
+    if (expected_size_min > actual_size || actual_size > expected_size_max) {
+        // Getting here is most likely an error.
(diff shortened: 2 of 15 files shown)


predmond:
Glad to see this stuff being used :)

Did you mean to include your FuzzBMP in this PR?

BenWiederhake:
> Glad to see this stuff being used :)

Those commits were the inspiration for me to do this; it's been sitting on my todo-list for quite a long time now :)

> Did you mean to include your FuzzBMP in this PR?

I left that out intentionally, since I intend to push that later, together with some kind of adapter that lets us re-use the `LLVMFuzzOneInput` functions with AFL. It *nearly* works, but not quite – hence I didn't incluce it in the list.

Likewise, I will eventually push the modified JS test. That's the way  I found the LibJS issue.

If you're curious about the BMP test, I have to disappoint: It's just a single boring call to BMPLoader::parse or something. If you really insist, I can add it to the PR.

Meta/Lagom/CMakeLists.txt
```diff
@@ -1,7 +1,7 @@
 cmake_minimum_required (VERSION 3.0)
 project (Lagom)
 
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-unknown-warning-option -O2 -Wall -Wextra -Werror -std=c++2a -fPIC -g -Wno-deprecated-copy")
```
nico, line 4: this is needed if your clang is new enough (i think 10+), which it has to be for the concept stuff

BenWiederhake: Huh, interesting. I never ran into trouble with my clang version 9.0.1-13 o.O

Alright, I'll upgrade my system (somehow?) and remove the commit.

predmond:
> 
> If you're curious about the BMP test, I have to disappoint: It's just a single boring call to BMPLoader::parse or something. If you really insist, I can add it to the PR.

No, just curious since you mentioned that one in the doc changes.

BenWiederhake:
Do I? I do! Then I really should push the `FuzzBMP.cpp`, too. Thanks!

Meta/Lagom/CMakeLists.txt
BenWiederhake: Removed.

BenWiederhake:
@nico Wanna do a proper review?

I haven't included the things from your gist yet – that's for another PR ^^

Libraries/LibGfx/Bitmap.cpp
```diff
@@ -44,11 +44,16 @@
...
+static bool size_would_overflow(BitmapFormat, const IntSize& size)
 {
     if (size.width() < 0 || size.height() < 0)
         return true;
-    return Checked<size_t>::multiplication_would_overflow(size.width(), size.height(), Bitmap::bpp_for_format(format));
+    // This check is a bit arbitrary, but should protect us from most shenanigans:
+    if (size.width() >= 16384 || size.height() >= 16384)
+        return true;
+    // This check is absolutely necessary. Note that Bitmap::Bitmap always stores
+    // data as RGBA32 internally, so currently we ignore the indicated format.
```
awesomekling, line 55: This is not accurate. `Bitmap` respects the `BitmapFormat`. Don't be fooled by the `RGBA32*`, it's cast to `u8*` for indexed 8-bit bitmaps.

nico:
Looks good to me!

Libraries/LibGfx/Bitmap.cpp
BenWiederhake: Interesting! We do seem to currently allocate that amount of memory, *even if* we only use it for `u8`s. Is that a bug?

// In Bitmap.h:
```
    size_t size_in_bytes() const { return m_pitch * m_size.height(); }
```

// In Bitmap.cpp:
```
Bitmap::Bitmap(BitmapFormat format, const IntSize& size, Purgeable purgeable)
…
    , m_pitch(round_up_to_power_of_two(size.width() * sizeof(RGBA32), 16))
…
{
…
    m_data = (RGBA32*)mmap_with_name(nullptr, size_in_bytes(), …);
…
}
```

BenWiederhake:
(For documentation: GitHub doesn't show the discussion in https://github.com/SerenityOS/serenity/commit/1e813748a89921190ed9da07c2cf3d7910249cef in this thread anymore, so that's why I add this link.)

Libraries/LibGfx/Bitmap.cpp
awesomekling: That's a bug :)

BenWiederhake: Fixed it :)

BenWiederhake:
Changes since the last push:
- `size_would_overflow` now uses 32K as an upper bound instead of 16K
- Slightly more error checking regarding pitch and shared memory size – this will only fail for existing bugs
- Bitmap pitch: Drop the automatic rounding up to a multiple of 16 – why did we have this in the first place? If it already was a multiple, then 0 bytes were added, so there is no secret additional data. If this is something about the framebuffer, we already offer a separate function specifically about that.
- Fix the pitch selection: Indexed bitmaps used to require four times the necessary memory. See [this discussion](https://github.com/SerenityOS/serenity/pull/3398#discussion_r483870887)

The build will fail falsely because of [this](https://github.com/SerenityOS/serenity/pull/3414#discussion_r484116787).

Applications/PixelPaint/BucketTool.cpp
```diff
@@ -58,10 +58,10 @@ static void flood_fill(Gfx::Bitmap& bitmap, const Gfx::IntPoint& start_position,
     while (!queue.is_empty()) {
         auto position = queue.dequeue();
 
-        if (bitmap.get_pixel<Gfx::BitmapFormat::RGBA32>(position.x(), position.y()) != target_color)
+        if (bitmap.get_pixel(position.x(), position.y()) != target_color)
             continue;
 
-        bitmap.set_pixel<Gfx::BitmapFormat::RGBA32>(position.x(), position.y(), fill_color);
+        bitmap.set_pixel(position.x(), position.y(), fill_color);
```
awesomekling, line 64: This was specifying the template type to avoid branching on every pixel during a flood fill. The PixelPaint application always uses RGBA32 format, so it's okay to take this shortcut.

BenWiederhake: The PR currently gets rid of the specializations, thus simplifying the interface, and decreasing the chances that it gets misused. If you insist on this microoptimization (does it actually make a difference?), I can add the templates back in, this time depending on `StorageFormat`.

awesomekling: Yes, it makes a difference. There might be better ways to implement the flood fill, but with the current algorithm, the extra branching slowed things down noticeably :/

BenWiederhake:
Changes since the last push:
- Make error checking not introduce a crash, but rather a "refusal" (before this PR, it was an obscure crash whenever a shared malformed Bitmap was used; before the most recent push, it was a crash when a shared malformed Bitmap was received; now it just returns an empty `RefPtr<Bitmap>`, like for the other allocation errors.)
- Revert the template specialization thing; [see discussion for details](https://github.com/SerenityOS/serenity/pull/3398#discussion_r485514295).

What a long journey ^^

awesomekling:
Excellent work, Ben :)

PR #3400, opened 2020-09-04
LibCore+Userland: Add Core::Account for managin user accounts

Petelliott:
`Core::Account` is a convenient way of managing `/etc/passwd`, and will be a way to manage `/etc/group` and `/etc/shadow` in the future.

This PR also fixes issues with `passwd(1)` that were discovered in #3296

### DIFF
--- a/Libraries/LibCore/Account.cpp
+++ b/Libraries/LibCore/Account.cpp
@@ -1,0 +1,207 @@
+/*
+ * Copyright (c) 2020, Peter Elliott <pelliott@ualberta.ca>
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include <AK/Base64.h>
+#include <AK/Random.h>
+#include <LibCore/Account.h>
+#include <LibCore/File.h>
+#include <grp.h>
+#include <pwd.h>
+#include <stdio.h>
+#include <string.h>
+#include <unistd.h>
+
+namespace Core {
+
+static String get_salt()
+{
+    char random_data[12];
+    AK::fill_with_random(random_data, sizeof(random_data));
+
+    StringBuilder builder;
+    builder.append("$5$");
+    builder.append(encode_base64(ReadonlyBytes(random_data, sizeof(random_data))));
+
+    return builder.build();
+}
+
+static Vector<gid_t> get_gids(const StringView& username)
+{
+    Vector<gid_t> extra_gids;
+    for (auto* group = getgrent(); group; group = getgrent()) {
+        for (size_t i = 0; group->gr_mem[i]; ++i) {
+            if (username == group->gr_mem[i]) {
+                extra_gids.append(group->gr_gid);
+                break;
+            }
+        }
+    }
+    endgrent();
+    return extra_gids;
+}
+
+Result<Account, String> Account::from_name(const char* username)
+{
+    struct passwd* pwd = nullptr;
+    errno = 0;
+    pwd = getpwnam(username);
+    if (!pwd) {
+        if (errno == 0)
+            return String("No such user");
+
+        return String(strerror(errno));
+    }
+
+    Account account(pwd, get_gids(pwd->pw_name));
+    endpwent();
+    return account;
+}
+
+Result<Account, String> Account::from_uid(uid_t uid)
+{
+    struct passwd* pwd = nullptr;
+    errno = 0;
+    pwd = getpwuid(uid);
+    if (!pwd) {
+        if (errno == 0)
+            return String("No such user");
+
+        return String(strerror(errno));
+    }
+
+    Account account(pwd, get_gids(pwd->pw_name));
+    endpwent();
+    return account;
+}
+
+bool Account::authenticate(const char* password) const
+{
+    // An empty passwd field indicates that no password is required to log in.
+    if (m_password_hash.is_empty())
+        return true;
+
+    // FIXME: Use crypt_r if it can be built in lagom.
+    char* hash = crypt(password, m_password_hash.characters());
+    return hash != nullptr && strcmp(hash, m_password_hash.characters()) == 0;
+}
+
+bool Account::login() const
+{
+    if (setgroups(m_extra_gids.size(), m_extra_gids.data()) < 0)
+        return false;
+
+    if (setgid(m_gid) < 0)
+        return false;
+
+    if (setuid(m_uid) < 0)
+        return false;
+
+    return true;
+}
+
+void Account::set_password(const char* password)
+{
+    m_password_hash = crypt(password, get_salt().characters());
+}
+
+void Account::set_password_enabled(bool enabled)
+{
+    if (enabled && m_password_hash != "" && m_password_hash[0] == '!') {
+        m_password_hash = m_password_hash.substring(1, m_password_hash.length() - 1);
+    } else if (!enabled && (m_password_hash == "" || m_password_hash[0] != '!')) {
+        StringBuilder builder;
+        builder.append('!');
+        builder.append(m_password_hash);
+        m_password_hash = builder.build();
+    }
+}
+
+void Account::delete_password()
+{
+    m_password_hash = "";
+}
+
+Account::Account(struct passwd* pwd, Vector<gid_t> extra_gids)
+    : m_username(pwd->pw_name)
+    , m_password_hash(pwd->pw_passwd)
+    , m_uid(pwd->pw_uid)
+    , m_gid(pwd->pw_gid)
+    , m_gecos(pwd->pw_gecos)
+    , m_home_directory(pwd->pw_dir)
+    , m_shell(pwd->pw_shell)
+    , m_extra_gids(extra_gids)
+{
+}
+
+bool Account::sync()
+{
+    StringBuilder new_passwd_file;
+
+    setpwent();
+
+    struct passwd* p;
+    errno = 0;
+    while ((p = getpwent())) {
+        if (p->pw_uid == m_uid) {
+            new_passwd_file.appendf("%s:%s:%u:%u:%s:%s:%s\n",
+                m_username.characters(),
+                m_password_hash.characters(),
+                m_uid, m_gid,
+                m_gecos.characters(),
+                m_home_directory.characters(),
+                m_shell.characters());
+
+        } else {
+            new_passwd_file.appendf("%s:%s:%u:%u:%s:%s:%s\n",
+                p->pw_name, p->pw_passwd, p->pw_uid,
+                p->pw_gid, p->pw_gecos, p->pw_dir,
+                p->pw_shell);
+        }
+    }
+    endpwent();
+
+    if (errno)
+        return false;
+
+    String contents = new_passwd_file.build();
+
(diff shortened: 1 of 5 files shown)


awesomekling:
cc @bugaevc

Libraries/LibCore/Account.cpp
```diff
@@ -0,0 +1,181 @@
...
+    return builder.build();
+}
+
+static Vector<gid_t> get_gids(const StringView& username)
+{
+    Vector<gid_t> extra_gids;
+    for (auto* group = getgrent(); group; group = getgrent()) {
+        for (size_t i = 0; group->gr_mem[i]; ++i) {
+            if (username == group->gr_mem[i])
+                extra_gids.append(group->gr_gid);
```
bugaevc, line 56: At this point we could break out of the inner loop.

Libraries/LibCore/Account.h
```diff
@@ -0,0 +1,74 @@
...
+    static Result<Account, String> get(uid_t uid);
+
+    bool authenticate(const StringView& password) const;
+    Result<uid_t, int> login() const;
+
+    String username() const { return m_username; }
+    String passwd() const { return m_passwd; }
+    void set_passwd(const StringView& passwd);
+    void set_passwd_enabled(bool enabled);
+    void delete_passwd();
```
bugaevc, line 50: Can't these fail?

Libraries/LibCore/Account.h
```diff
@@ -0,0 +1,74 @@
...
+
+    uid_t uid() const { return m_uid; }
+    gid_t gid() const { return m_gid; }
+    String gecos() const { return m_gecos; }
+    String home_directory() const { return m_home_directory; }
+    String shell() const { return m_shell; }
+    Vector<gid_t> extra_gids() const { return m_extra_gids; }
```
bugaevc, line 57:
```suggestion
    const Vector<gid_t>& extra_gids() const { return m_extra_gids; }
```

And preferably the same for strings, although strings are cheap to copy.

Libraries/LibCore/Account.h
```diff
@@ -0,0 +1,74 @@
...
+
+    void sync();
+
+private:
+    explicit Account(struct passwd* pwd, Vector<gid_t> extra_gids);
```
bugaevc, line 62: Multiple arguments, so
```suggestion
    Account(struct passwd* pwd, Vector<gid_t> extra_gids);
```

Libraries/LibCore/Account.h
```diff
@@ -0,0 +1,74 @@
...
+
+    String m_username;
+    String m_passwd;
+    uid_t m_uid;
+    gid_t m_gid;
```
bugaevc, line 67: Need initial values for these two.

Libraries/LibCore/Account.cpp
```diff
@@ -0,0 +1,181 @@
...
+        }
+    }
+    endgrent();
+    return extra_gids;
+}
+
+Result<Account, String> Account::get(const StringView& username)
+{
+    String username_with_terminator(username);
```
bugaevc, line 65: Doesn't matter, but I'd just make this method accept `const char*`, and let the caller worry about providing a null-terminated string. If they already have it as a `String` (quite likely), or as a C string (also quite likely if it comes from argv), we can skip a copy/allocation.

Libraries/LibCore/Account.cpp
```diff
@@ -0,0 +1,181 @@
...
+
+    struct passwd* pwd = nullptr;
+    pwd = getpwnam(username_with_terminator.characters());
+    if (!pwd)
+        return String("No such user");
```
bugaevc, line 70: Not necessarily; it could be an error reading pwd database. Set `errno = 0` before the call and check it here.

Libraries/LibCore/Account.cpp
```diff
@@ -0,0 +1,181 @@
...
+    endpwent();
+    return account;
+}
+
+Result<Account, String> Account::get(uid_t uid)
+{
+    struct passwd* pwd = nullptr;
+    pwd = getpwuid(uid);
+    if (!pwd)
+        return String("No such user");
```
bugaevc, line 82: Ditto

Libraries/LibCore/Account.cpp
```diff
@@ -0,0 +1,181 @@
...
+
+    Account account(pwd, get_gids(pwd->pw_name));
+    endpwent();
+    return account;
+}
+
+bool Account::authenticate(const StringView& password) const
+{
+    if (m_passwd == "")
```
bugaevc, line 91:
```suggestion
    if (m_passwd.is_empty())
```

Also, does the empty password have a special meaning?

Libraries/LibCore/Account.h
```diff
@@ -0,0 +1,74 @@
...
+    String shell() const { return m_shell; }
+    Vector<gid_t> extra_gids() const { return m_extra_gids; }
+
+    void sync();
+
+private:
+    explicit Account(struct passwd* pwd, Vector<gid_t> extra_gids);
+
+    String m_username;
+    String m_passwd;
```
bugaevc, line 65: Is this actually a password, or just some hash of something?

Libraries/LibCore/Account.cpp
```diff
@@ -0,0 +1,181 @@
...
+
+    // FIXME: Use crypt_r if it can be built in lagom.
+    char* hash = crypt(password_with_terminator.characters(), m_passwd.characters());
+    return hash != nullptr && strcmp(hash, m_passwd.characters()) == 0;
+}
+
+Result<uid_t, int> Account::login() const
+{
+    if (setgroups(m_extra_gids.size(), m_extra_gids.data()) < 0)
+        return errno;
```
bugaevc, line 104: I don't trust type-based dispatch to work too well here (what if `uid_t` is an `int`? or if `errno` is not an `int`, but say an `unsigned int` or an `int32_t`?)

Is there an explicit way to initialize one of the `Result` variants?

Libraries/LibCore/Account.cpp
```diff
@@ -0,0 +1,181 @@
...
+            new_passwd_file.appendf("%s:%s:%u:%u:%s:%s:%s\n",
+                p->pw_name, p->pw_passwd, p->pw_uid,
+                p->pw_gid, p->pw_gecos, p->pw_dir,
+                p->pw_shell);
+        }
+    }
+    endpwent();
+
+    auto file = Core::File::open("/etc/passwd", Core::IODevice::OpenMode::WriteOnly);
+    file.value()->write(new_passwd_file.build());
```
bugaevc, line 177: Reminder that any of these (opening the files, reading entries, writing data) can fail, and you need to nicely propagate the failure to the caller rather than swallowing it and carrying on.

Also, `IODevice::write()` is clearly broken:

https://github.com/SerenityOS/serenity/blob/02b3cb812367266da673edda30c3393fa25e2e52/Libraries/LibCore/IODevice.cpp#L269-L278

Please either fix it, or use some other means of writing to a file.

Userland/passwd.cpp
```diff
@@ -105,41 +71,36 @@ int main(int argc, char** argv)
 
     uid_t current_user = getuid();
 
-    struct passwd pwd;
+    auto account = Core::Account::get(current_user).value();
```
bugaevc, line 74: You can't just get a `.value()` like that, you need to handle the failure case

Userland/passwd.cpp
```diff
@@ -105,41 +71,36 @@ int main(int argc, char** argv)
...
     if (username) {
-        pwd = *getpwnam(username);
-        if (current_user != 0 && current_user != pwd.pw_uid) {
+        auto result = Core::Account::get(username);
```
bugaevc, line 76: To be less confusing, I'd add comments (and more descriptive variable names) that the above is the account for the current user, and this is for the user we're modifying.

Userland/passwd.cpp
```diff
@@ -105,41 +71,36 @@ int main(int argc, char** argv)
...
+        if (result.is_error()) {
+            fprintf(stderr, "Core::Account::get: %s.\n", result.error().characters());
```
bugaevc, line 78:
```suggestion
            fprintf(stderr, "Core::Account::get: %s\n", result.error().characters());
            exit(1);
```

Userland/su.cpp
```diff
@@ -47,66 +45,34 @@ int main(int argc, char** argv)
...
-        gid = pwd->pw_gid;
-    }
-
-    if (!pwd)
-        pwd = getpwuid(0);
-
-    if (!pwd) {
+    auto account_or_error = (user) ? Core::Account::get(user) : Core::Account::get(0);
+    if (account_or_error.is_error()) {
         fprintf(stderr, "No passwd entry.\n");
```
bugaevc, line 50: Print the actual error (e.g. `Cannot open /etc/passwd: no such file or directory`), not the catch-all "no passed entry".

Userland/su.cpp
```diff
@@ -47,66 +45,34 @@ int main(int argc, char** argv)
...
         return 1;
     }
 
-    if (getuid() != 0 && pwd->pw_passwd[0] != '\0') {
+    Core::Account account = account_or_error.value();
+
+    if (getuid() != 0 && account.passwd() != "") {
         auto password = Core::get_password();
         if (password.is_error()) {
             fprintf(stderr, strerror(password.error()));
```
bugaevc, line 59: Another `strerror`-based injection vulnerability! 😆

Libraries/LibCore/Account.cpp
Petelliott: yes. an empty password means no password is required to authenticate as the user.

Libraries/LibCore/Account.h
Petelliott: it is equivalent  the `pw_passwd` field of `struct pwd`, so it contains a hash, nothing, `x`, `!`.

I tried to keep `Core::Account` conceptually similar to `struct pwd`.

Libraries/LibCore/Account.cpp
bugaevc: Can the user not have an actual empty password?

Or, perhaps this is an empty *encrypted* password, not the password itself? In that case, can we maybe use null instead of an empty string?

In any case, this needs a comment or two 🙂

Libraries/LibCore/Account.h
bugaevc: Sure, but the field names are not exactly identical already, so we can tweak it further. How about `m_password_hash`?

> so it contains a hash, nothing, `x`, `!`

That, too, needs to be a comment 😀

Libraries/LibCore/Account.h
Petelliott: No. These are just setters to in memory copy. `sync()` can fail though, so I'll propagate errors from it

bugaevc: Yeah, I see that now. But as it was not obvious for me while reading just the header, can we have a comment here too, please :smile:

Libraries/LibCore/Account.cpp
Petelliott: an actual empty password would still be hashed. The use of an empty password field to mean no password is a unix thing.

from linux's `passwd(5)`:
> The encrypted password field may be empty, in which case no password is required to authenticate as the specified login name.

If we were to use null, we would have to convert to and from null when reading and writing the field.

Libraries/LibCore/Account.cpp
Petelliott: Honestly, there is no reason to return the `uid_t` in the first place.

Is there an idiomatic way to return just an error in serenity?

bugaevc: In the kernel, that'd be `KResult`. In userspace, I don't think we have anything quite like that. We could just do `bool` + `errno` though.

Libraries/LibCore/Account.cpp
bugaevc: Still relevant: all of these can fail, and `IODevice::write()` doesn't actually write out the whole thing, only a part of it, and doesn't even tell you how much.

Petelliott: I haven't gotten to that part yet

bugaevc: @awesomekling this is still unsolved, @Petelliott hasn't gotten to that part yet :smile:

Userland/su.cpp
bugaevc: This, too, still unsolved, though these lines are not changed by this PR, would be nice to fix nevertheless.

Libraries/LibCore/Account.cpp
awesomekling: Ah okay! The GitHub UI is confusing :]

bugaevc: It is confusing, yeah!

Still beats sending and reviewing patches over email :slightly_smiling_face:

Petelliott: I've changed this to use `stdio` rather than `IODevice`. Sorry for the extremely long delay.

Libraries/LibCore/Account.cpp
```diff
@@ -0,0 +1,202 @@
...
+
+    String contents = new_passwd_file.build();
+
+    FILE* passwd_file = fopen("/etc/passwd", "w");
+    if (!passwd_file)
+        return false;
+
+    fwrite(contents.characters(), 1, contents.length(), passwd_file);
+    if (ferror(passwd_file)) {
+        // fclose sets errno if write fails.
```
bugaevc, line 193: But what if the `fwrite()` above failed (say, with `ENOSPC`), but `fclose()` succeeds?

I mean, it tires to flush, so it'll probably get the same error again, but I wouldn't rely on that.

Libraries/LibCore/Account.cpp
```diff
@@ -0,0 +1,202 @@
...
+}
+
+bool Account::sync()
+{
+    StringBuilder new_passwd_file;
+
+    setpwent();
+
+    struct passwd* p;
+    while ((p = getpwent())) {
```
bugaevc, line 166: This, too, can fail. If we don't distinguish errors from reaching the end of users, we'll just end up truncating the file, and so silently dropping some users.

Libraries/LibCore/Account.cpp
bugaevc: Great!

I've got some more suggestions, sorry :(

Libraries/LibCore/Account.cpp
bugaevc: > Would you prefered that I set `errno = ferror()`, or return an error number?

Just make sure to do the same thing no matter which exit path this function takes 🙂 I think you're already setting `errno` in other places (or calling functions that set it for you), so I'd set `errno = ferror()` here.

bugaevc:
Looks pretty good now 🙂

We should replace the file atomically, but guess we can implement that later.

Petelliott:
I think this is complete. Is there anything I am missing?

awesomekling:
🚀

PR #3401, opened 2020-09-04
LibGUI: fix GIF animation in ImageWidget

peterdn:
Recent changes to GIF decoder mean that the ImageDecoder::frame method
no longer returns a completely new Bitmap object for each frame, but
instead updates the contents of a single cached Bitmap. Therefore after
advancing the frame we now directly update() instead of calling
set_bitmap which is a null operation if the Bitmap reference has not
changed.

### DIFF
--- a/Libraries/LibGUI/ImageWidget.cpp
+++ b/Libraries/LibGUI/ImageWidget.cpp
@@ -72,7 +72,7 @@ void ImageWidget::animate()
     m_current_frame_index = (m_current_frame_index + 1) % m_image_decoder->frame_count();
 
     const auto& current_frame = m_image_decoder->frame(m_current_frame_index);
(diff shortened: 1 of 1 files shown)


awesomekling:
Hmm, thinking about the implications of this, I don't think the `ImageDecoder` API is sane this way.
Vending a `Gfx::Bitmap` that may change as a side effect of the animation progressing seems pretty rickety.

peterdn:
> Hmm, thinking about the implications of this, I don't think the `ImageDecoder` API is sane this way.
> Vending a `Gfx::Bitmap` that may change as a side effect of the animation progressing seems pretty rickety.

Yes, on second thoughts this might be a pretty surprising side effect. I'll have `frame()` return a clone of the GIF decoder's framebuffer instead.

PR #3402, opened 2020-09-05
AK: FixedArray::clear fixed to destroy in reverse order

tryfinally:
order of destructors should be reversed order of constructors

### DIFF
--- a/AK/FixedArray.h
+++ b/AK/FixedArray.h
@@ -73,7 +73,8 @@ public:
     {
         if (!m_elements)
             return;
-        for (size_t i = 0; i < m_size; ++i)
+        size_t i = m_size;
+        while (i-- > 0)
             m_elements[i].~T();
         kfree(m_elements);
         m_elements = nullptr;


awesomekling:
Why make this change?
What bug does it fix?
Please explain why you are making changes.

asynts:
C++ actually defines the order in which elements of a regular array are destroyed:

~~~c++
#include <iostream>

class A {
public:
    A(int value)
        : m_value(value)
    {
    }

    ~A() {
        std::cout << m_value << '\n';
    }

private:
    int m_value;
};

int main() {
    A a[] = { A(1), A(2), A(3), A(4) };

    // output: 4 3 2 1
}
~~~

There is something in the standard that says that this must destroy the elements of the array in the reverse order, so this change makes sense.

awesomekling:
Right, I know that is the case for "primitive" arrays, but `FixedArray` is not an implementation of a standard array.
None of the other `AK` containers destroy elements in reverse order, so why is `FixedArray` special?
I'm not necessarily against the change itself but I dislike the pattern of submitting code without proper justification or explanation.

tryfinally:
It's not only a standard library thing. All objects should be destroyed in reverse order of construction. It's correct that AK is not STL replacement but such behavior goes against expectations of C++ Programmers using it.

tryfinally:
BTW, AK CircularQueue does exactly that destroying objects in reverse order

awesomekling:
> It's not only a standard library thing.
Nobody said it was.

>All objects should be destroyed in reverse order of construction.
There exists no such requirement for manually managed objects.

>It's correct that AK is not STL replacement but such behavior goes against expectations of C++ Programmers using it.
It does not go against my expectations. The AK containers are heavily based on WebKit's WTF template library, which also destroys in forward (or random) order.

> It's try for local objects, global objects etc.
Yes, stack allocated objects are destroyed in order of allocation. That has nothing to do with the behavior of arbitrary code executing in a custom template class destructor.

>BTW, AK CircularQueue does exactly that destroying objects in reverse order
No it doesn't.

Look, again, I'm not against changing the destruction order. Byt then it should be done for *all* containers and we should add tests for it, not only change `FixedArray` with no explanation of why it's special and should behave differently from the other containers.

The main issue here is that you keep making pull requests that change core library code without explanation or justification other than "change it to do things my way". If you want to make changes like this, then:

1. Write proper commit messages that explain what you are doing and why it's better
2. Include unit tests that demonstrate the changes and protect against future regressions
3. For API changes to library code, apply the same pattern to the whole library, not just one individual class

tryfinally:
My line of thought was that sequence containers emulating array _should_ destroy objects in reverse.

> > All objects should be destroyed in reverse order of construction.
> There exists no such requirement for manually managed objects.
You are right. I have researched the **The C++ standard** it mandates that for native array elements but stays silent on std::array.
BTW, std::array has many shortcomings as defined by the standard this is not the only one.

> > BTW, AK CircularQueue does exactly that destroying objects in reverse order
> 
> No it doesn't.
> 
I meant should. My mistake.

> Look, again, I'm not against changing the destruction order. Byt then it should be done for _all_ containers and we should add tests for it.
Maybe that's a good idea, just make the order of destruction explicit. reverse only where appropriate.

> not only change `FixedArray` with no explanation of why it's special and should behave differently from the other containers.
Because it's intended as a replacement of native arrays and it doesn't support insert or delete in the middle.

> The main issue here is that you keep making pull requests that change core library code without explanation or justification other than "change it to do things my way". 
Not my intention. For this PR, I really thought its self explanatory.

If you want to make changes like this, then:
> 
>     1. Write proper commit messages that explain what you are doing and why it's better
> 
>     2. Include unit tests that demonstrate the changes and protect against future regresand forget itsions
> 
>     3. For API changes to library code, apply the same pattern to the whole library, not just one individual class
Ok, thanks again for you taking the time and patience. 

Should I go ahead and implement the change to CyclicQueue according to the above or just close the PR and forget it?

Issue #3399, opened 2020-09-04
Mark imports and license-borders properly


BenWiederhake:
For example, we contain a copy of the [bmpsuite](https://github.com/jsummers/bmpsuite) at `/Base/res/html/misc/bmpsuite.html`, which we can probably use under the [GLPv3](https://github.com/jsummers/bmpsuite/blob/master/COPYING.txt). However, the current serenity repository doesn't seem to indicate that this is an external resource.

I believe we "sufficiently" adhere to all the terms and conditions, for bmpsuite.
However, it's probably a good idea to mark those project/license borders somehow, in order to properly attribute copyright and also declare licenses wherever we have to – after all, bmpsuite isn't the only "imported" code, by far.

Debian solves this by having a central-ish `debian/copyright` file per source package, which essentially lists pairs of shell-globs and licenses. Maybe something similar might work for us, too? We could create a few `copyright` files of a similar format in few, strategic places like `Libraries/`, `Base/res/`, etc. That would reduce the chances of legal trouble from "extremely low but not zero" to "virtually zero".

awesomekling:
Hmm, that's a good point. I think the cleanest approach would be to (incrementally) move non-original resources to one or more separate repositories. It's not like we need to ship the bmpsuite anyway, it's just something used for testing.

BenWiederhake:
For things like `bmpsuite`, that's a good idea.

Some of the code is copied from *BSD, for example `/Libraries/LibELF/exec_elf.h`. In that case, this doesn't work as easily. (Actually, this might be the only such file, I don't know.)

nico:
For actual code, I think the plan is to have no external code. It's probably a good idea to make a list of files that need converting.

BenWiederhake:
Looks like `Meta/check_style.sh` already contains such a list:
- `Kernel/FileSystem/ext2_fs.h`
- `Libraries/LibC/getopt.cpp`
- `Libraries/LibCore/puff.h`
- `Libraries/LibCore/puff.cpp`
- `Libraries/LibELF/exec_elf.h`

Furthermore, these files contain "unusual" copyright notices:
- `Libraries/LibC/scanf.cpp:2: * Copyright (c) 2000-2002 Opsycon AB  (www.opsycon.se)` (dead website)
- `AK/Checked.h:2: * Copyright (C) 2011-2019 Apple Inc. All rights reserved.` (probably Andreas, but none of my business)
[...]

PR #3403, opened 2020-09-05
Meta: Fix wonky copyright headers in Tests

BenWiederhake:
I noticed these issues while researching for #3399. Thankfully, only my own headers are affected, so no "re-signing coordination" is needed.


awesomekling:
W O N K Y

PR #3410, opened 2020-09-06
Give ntpquery a `-s` flag to make it adjust the local clock

nico:
Also use time.google.com as default NTP server until the pool.ntp.org application is through.

This is a very simple implementation for now. Eventually, I want to add the smooth `adjtime` adjustment if local time isn't too far off, and implement the NTP rfc filtering algorithms, and probably switch to the timespec clock_get/settime functions. But for now, let's go with this.

### DIFF
--- a/Userland/ntpquery.cpp
+++ b/Userland/ntpquery.cpp
@@ -168,6 +168,7 @@ int main(int argc, char** argv)
 
     socklen_t peer_address_size = sizeof(peer_address);
     rc = recvfrom(fd, &packet, sizeof(packet), 0, (struct sockaddr*)&peer_address, &peer_address_size);
+    gettimeofday(&t, nullptr);
     if (rc < 0) {
         perror("recvfrom");
         return 1;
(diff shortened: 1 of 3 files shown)


Userland/ntpquery.cpp
```diff
@@ -100,27 +100,33 @@ static String format_ntp_timestamp(NtpTimestamp ntp_timestamp)
...
+    args_parser.add_option(set_time, "Adjust system time (requires root)", "set", 's');
+    args_parser.add_positional_argument(host, "NTP server", "host", Core::ArgsParser::Required::No);
     args_parser.parse(argc, argv);
 
+    if (!set_time) {
+        if (pledge("stdio inet dns settime", nullptr) < 0) {
+            perror("pledge");
+            return 1;
+        }
+    }
```
asynts, line 121: Shouldn't this be:

```suggestion
    if (pledge("stdio inet dns settime", nullptr) < 0) {
        perror("pledge");
        return 1;
    }
```

nico: Oh, good catch. The `pledge("stdio inet dns settime", nullptr) < 0) ` is at the start of main(); this was supposed to drop settime if -s isn't passed. Now it does.

asynts: Oh, I misunderstood what you were trying to do, but somehow still found a bug...

Kernel/Process.cpp
```diff
@@ -567,6 +567,11 @@ void kgettimeofday(timeval& tv)
     tv = kgettimeofday();
 }
 
+void ksettimeofday(timeval tv)
+{
+    g_timeofday = tv;
```
awesomekling, line 572: This doesn't seem like it would work. `g_timeofday` will be overwritten on next timer IRQ anyway.
Also, is this meaningfully different from `sys$clock_settime()` (which we already have)?

nico: Oh, that's how this works. Why does kgettimeofday() not return TimeManagement::the().epoch_time()? Concurrency?

Changed things to make this a LibC-only implementation.

awesomekling: Yeah basically. And the idea going forward is to move the timestamp to a single canonical location that is also accessible from userspace via a special mapping.

PR #3412, opened 2020-09-06
AK:Vector+SinglyLinkedList use Traits<T>::equals in search functions

tryfinally:
make search/find mem-functions use Traits<T>::equals instead of operator== to align these containers with implementation in rest of AK.

### DIFF
--- a/AK/SinglyLinkedList.h
+++ b/AK/SinglyLinkedList.h
@@ -28,6 +28,7 @@
 
 #include <AK/Assertions.h>
 #include <AK/StdLibExtras.h>
+#include <AK/Traits.h>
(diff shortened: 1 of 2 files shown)


emanuele6:
You are writing `Taris<T>::equals` in your commit messages instad of  `Traits<T>:equals`.
You are also saying that you are using it instead of `operator=`, but you are using it instead of `operator==`.

tryfinally:
> You are writing `Taris<T>::equals`, in your commit messages, instead of `Traits<T>:equals`.
> You are also saying that you are using `Traits<T>:equals` instead of `operator=`, but you are using it instead of `operator==`.

fixed the typos

PR #3414, opened 2020-09-06
Fix git port

itamar8910:
Fixed some things in LibC that changed recently and broke building the git port.
I also added a script to Travis that run post-build and fails if it finds "forbidden symbols" in LibC,
as @BenWiederhake suggested [here](https://github.com/SerenityOS/serenity/pull/3296#issuecomment-685055450).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -49,6 +49,7 @@ script:
 - cd Build
 - cmake .. -DBUILD_LAGOM=1 -DALL_THE_DEBUG_MACROS=1
 - make -j2
+- Meta/check-symbols.sh
 - CTEST_OUTPUT_ON_FAILURE=1 make test
 - cd Meta/Lagom
 - DISABLE_DBG_OUTPUT=1 ./test-js
--- a/Meta/check-symbols.sh
+++ b/Meta/check-symbols.sh
@@ -1,0 +1,21 @@
+#!/bin/sh
+
+script_path=$(cd -P -- "$(dirname -- "$0")" && pwd -P)
+cd "$script_path/.."
+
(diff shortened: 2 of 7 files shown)


Meta/check-symbols.sh
```diff
@@ -0,0 +1,21 @@
...
+
+script_path=$(cd -P -- "$(dirname -- "$0")" && pwd -P)
+cd "$script_path/.."
+
+# The __cxa_guard_* calls are generated for (non trivial) initialzation of local static objects.
+# These symbols are OK to use within serenity code, but they are problematic in LibC because their
+# existence breaks ports (the implementation of these symbols resides in libstdc++.a which we do not link against in ports).
+# To eliminate the need for these symbols, avoid doing non-trivial construction of local statics in LibC.
+
+FORBIDDEN_SYMBOLS="__cxa_guard_acquire __cxa_guard_release
```
alimpfard, line 11: You probably meant to end this string at EOL?

itamar8910: huh :sweat_smile:

awesomekling:
An even more powerful idea would be building a handful of ports on Travis. For example `git` :)

.travis.yml
```diff
@@ -49,6 +49,7 @@ script:
 - cd Build
 - cmake .. -DBUILD_LAGOM=1 -DALL_THE_DEBUG_MACROS=1
 - make -j2
+- Meta/check-symbols.sh
```
BenWiederhake, line 52: @itamar8910 @awesomekling Because of the `cd Build` a few lines earlier, Travis is in the `Build` folder at this point, so naturally it tries to search for the script at `Build/Meta/check-symbols.sh`. That's why all builds are currently failing. :P

PR #3420, opened 2020-09-07
Let `ntpquery -s` set time with precision of tens of milliseconds instead of tens of seconds.

nico:


awesomekling:
Very cool :)

PR #3422, opened 2020-09-07
LibLine: Treat ^D as EOF only when the buffer is empty

### LINKED ISSUE #3421: Regression: Ctrl-D treated as EOD instead of delete-char during editing
1. Enter "asf"
2. Hit ctrl-a to jump to start of line
3. Hit ctrl-d

expected: "a" is deleted
actual: shell closes

@alimpfard

alimpfard:
As opposed to when the cursor is at the start of the buffer.
Fixes #3421.


awesomekling:
Oh good fix!

PR #3426, opened 2020-09-07
Add Array<T, Size> template to replace heap allocated FixedArray.

asynts:
This pull request does the following:

  - Replace `VectorIterator` with `SimpleIterator`, the old class was already used for classes other than `Vector` and seemed more of a workaround. `String` and `StringView` now also use this iterator class instead of raw pointers.

  - Adds `Array<T, Size>` class; no more `ReadonlyBytes{ buffer, sizeof(buffer) }`! I thought about adding something like `std::experimental::make_array` but decided against it, because I don't like it and it doesn't make a big difference anyways.

  - Removes `FixedArray` class. This class was used in two ways:

      - When an array with compile-time known size was needed. `Array<T, Size>` is better for that because it doesn't put the array on the heap.

      - When the author didn't know about `Vector`? I mean how is an array "fixed" when you call something like `FixedArray::resize`?

---

I am not sure about the changes I made in `Kernel/Syscalls/getuid.cpp`, this loop seemed unnecessary to me, I removed it:

https://github.com/asynts/serenity/blob/48d9f3c2e6a77f9596a4dbc5d19eb7d5d1976b00/Kernel/Syscalls/getuid.cpp#L89-L91

However, I am not super familiar with kernel stuff, so this copy might be intentional?


awesomekling:
> * When the author didn't know about `Vector`? I mean how is an array "fixed" when you call something like `FixedArray::resize`?

It's used where the dynamic growth feature of `Vector` is not needed. `FixedArray` itself is smaller (no need to track capacity separate from size) and it allocates precisely-sized buffers instead of capacity-padded ones. That's what it was used for.

I agree that the name is bad. :)

awesomekling:
> I am not sure about the changes I made in `Kernel/Syscalls/getuid.cpp`, this loop seemed unnecessary to me, I removed it:
> 
> https://github.com/asynts/serenity/blob/48d9f3c2e6a77f9596a4dbc5d19eb7d5d1976b00/Kernel/Syscalls/getuid.cpp#L89-L91
> 
> However, I am not super familiar with kernel stuff, so this copy might be intentional?

That looks unnecessary indeed! I think it was originally a `HashTable` and so it had to be copied to flat storage before copying it to userspace. :)

asynts:
> It's used where the dynamic growth feature of Vector is not needed. `FixedArray` itself is smaller (no need to track capacity separate from size) and it allocates precisely-sized buffers instead of capacity-padded ones. That's what it was used for.

That should only provide a marginal benefit. All of these cases were arrays of relatively small types like `gid_t` or `RefPtr`. On the other hand, it is a bit more descriptive. Hmm.

Should I add it back? Maybe as `FixedVector`?

awesomekling:
Meh. I think we can just get rid of it. I just wanted to provide some background on where it came from.
You are right that the benefit is marginal.

PR #3430, opened 2020-09-07
WindowServer: Fix invalidating window frame

### LINKED ISSUE #3427: WindowServer: Stale window title in new Terminal windows
To reproduce:

1. Press Ctrl-Shift-N in Terminal to create a new terminal.
2. Notice that the window title is "Terminal" instead of "anon@courage:~"
3. Move the window slightly
4. Notice that the window title changes when you move the window

cc @tomuta

tomuta:
When invalidating the frame we need to properly flag that so that
we trigger rendering the frame, even if "all" was flagged as being
invalidated. Otherwise it will only get rendered if anything else
happens to trigger it (such as focus change).

Fixes #3427

### DIFF
--- a/Services/WindowServer/Window.cpp
+++ b/Services/WindowServer/Window.cpp
@@ -409,21 +409,25 @@ void Window::invalidate(bool invalidate_frame)
     Compositor::the().invalidate_window();
 }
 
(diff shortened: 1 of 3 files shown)


awesomekling:
👍

PR #3432, opened 2020-09-07
LibJS: Some copy elisions and MVL interface niceties

alimpfard:
This commit also eliminates quite a few unnecessary copies of Vectors.

Why redo Vector features, or forward calls when we can just inherit from Vector

cc @linusg - in case I messed some required copy up.

### DIFF
--- a/Libraries/LibJS/Runtime/IndexedProperties.cpp
+++ b/Libraries/LibJS/Runtime/IndexedProperties.cpp
@@ -318,7 +318,7 @@ void IndexedProperties::insert(u32 index, Value value, PropertyAttributes attrib
     m_storage->insert(index, value, attributes);
 }
 
-ValueAndAttributes IndexedProperties::take_first(Object *this_object)
+ValueAndAttributes IndexedProperties::take_first(Object* this_object)
 {
     auto first = m_storage->take_first();
     if (first.value.is_accessor())
@@ -326,7 +326,7 @@ ValueAndAttributes IndexedProperties::take_first(Object *this_object)
     return first;
 }
 
-ValueAndAttributes IndexedProperties::take_last(Object *this_object)
+ValueAndAttributes IndexedProperties::take_last(Object* this_object)
 {
     auto last = m_storage->take_last();
     if (last.value.is_accessor())
--- a/Libraries/LibJS/Runtime/IndexedProperties.cpp
+++ b/Libraries/LibJS/Runtime/IndexedProperties.cpp
@@ -273,7 +273,7 @@ Optional<ValueAndAttributes> IndexedProperties::get(Object* this_object, u32 ind
         return result;
     if (!result.has_value())
         return {};
-    auto value = result.value();
+    auto& value = result.value();
     if (value.value.is_accessor()) {
         ASSERT(this_object);
         auto& accessor = value.value.as_accessor();
@@ -315,7 +315,7 @@ void IndexedProperties::insert(u32 index, Value value, PropertyAttributes attrib
 {
     if (m_storage->is_simple_storage() && (index >= SPARSE_ARRAY_THRESHOLD || attributes != default_attributes || array_like_size() == SPARSE_ARRAY_THRESHOLD))
         switch_to_generic_storage();
-    m_storage->insert(index, value, attributes);
+    m_storage->insert(index, move(value), attributes);
 }
 
 ValueAndAttributes IndexedProperties::take_first(Object* this_object)
@@ -340,7 +340,7 @@ void IndexedProperties::append_all(Object* this_object, const IndexedProperties&
         switch_to_generic_storage();
 
     for (auto it = properties.begin(false); it != properties.end(); ++it) {
-        auto element = it.value_and_attributes(this_object, evaluate_accessors);
+        const auto& element = it.value_and_attributes(this_object, evaluate_accessors);
         if (this_object && this_object->interpreter().exception())
             return;
         m_storage->put(m_storage->array_like_size(), element.value, element.attributes);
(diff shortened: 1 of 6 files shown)


Libraries/LibJS/Runtime/IndexedProperties.cpp
```diff
@@ -374,7 +374,7 @@ Vector<ValueAndAttributes> IndexedProperties::values_unordered() const
 
 void IndexedProperties::switch_to_generic_storage()
 {
-    auto storage = static_cast<const SimpleIndexedPropertyStorage&>(*m_storage);
+    auto& storage = static_cast<SimpleIndexedPropertyStorage&>(*m_storage);
```
alimpfard, line 377: "necessary copy": In particular, this one.

Libraries/LibJS/Runtime/ArrayPrototype.cpp
```diff
@@ -558,7 +558,7 @@ JS_DEFINE_NATIVE_FUNCTION(ArrayPrototype::reverse)
     if (array->indexed_properties().is_empty())
         return array;
 
-    Vector<Value> array_reverse;
+    MarkedValueList array_reverse{array->heap()};
```
linusg, line 561:
```suggestion
    MarkedValueList array_reverse { array->heap() };
```

at least I think that's what clang-format should do?

alimpfard: it should lol

linusg (approved):
> This commit also eliminates quite a few unnecessary copies of Vectors.

I'm definitely responsible for a few of those!

LGTM - except for the changed constructors, I don't understand those well enough to judge :)

Libraries/LibJS/Runtime/Object.h
```diff
@@ -169,7 +169,7 @@ class Object : public Cell {
 
     bool m_is_extensible { true };
     Shape* m_shape { nullptr };
-    Vector<Value> m_storage;
+    MarkedValueList m_storage;
```
awesomekling, line 172: This change is not needed, as `JS::Object` will mark the values in its storage in `Object::visit_children()`

Libraries/LibJS/Runtime/IndexedProperties.h
```diff
@@ -83,13 +87,13 @@ class SimpleIndexedPropertyStorage final : public IndexedPropertyStorage {
...
     virtual bool is_simple_storage() const override { return true; }
-    Vector<Value> elements() const { return m_packed_elements; }
+    const MarkedValueList& elements() const { return m_packed_elements; }
 
 private:
     friend GenericIndexedPropertyStorage;
 
     size_t m_array_size { 0 };
-    Vector<Value> m_packed_elements;
+    MarkedValueList m_packed_elements;
```
awesomekling, line 96: This change is not needed, `Object` will mark these in `visit_children()`.

awesomekling (changes requested):

Libraries/LibJS/Runtime/Object.h
alimpfard: Should I make `Object` _not_ mark these, or just revert this (and the one below)?

awesomekling: Just revert it :)
We have to visit the objects anyway, so let's avoid tracking more MVL's

Libraries/LibJS/Runtime/Function.h
```diff
@@ -66,12 +66,12 @@ class Function : public Object {
...
 protected:
     explicit Function(Object& prototype);
-    Function(Object& prototype, Value bound_this, Vector<Value> bound_arguments);
+    Function(Object& prototype, Value bound_this, MarkedValueList bound_arguments);
 
 private:
     virtual bool is_function() const override { return true; }
     Value m_bound_this;
-    Vector<Value> m_bound_arguments;
+    MarkedValueList m_bound_arguments;
```
awesomekling, line 74: This is also already marked by `Function::visit_children()`.

Libraries/LibJS/Interpreter.h
```diff
@@ -59,9 +59,25 @@ struct ScopeFrame {
...
+        , this_value(move(this_value))
+        , arguments(move(arguments))
+        , environment(environment)
+    {
+    }
+
     FlyString function_name;
     Value this_value;
-    Vector<Value> arguments;
+    MarkedValueList arguments;
```
awesomekling, line 80: This is also already marked by `Interpreter::gather_roots()`.

awesomekling (changes requested):
Moar:

Libraries/LibJS/Runtime/Function.h
alimpfard: Hmmm...at this point, I question: what _isn't_ marked and is stored in a `Vector<Value>`?

The commit message certainly no longer applies.

awesomekling: Hmm. The stuff returned from `IndexedProperties::values_unordered()` isn't marked.

awesomekling: `SimpleIndexedPropertyStorage::elements()` returns a vector by copy (for no reason I think), which won't be marked

awesomekling: All the `ValueAndAttributes` usage looks really suspicious and unmarked actually

Libraries/LibJS/Interpreter.h
alimpfard: So is everything in a `CallStack`.

Libraries/LibJS/Runtime/Function.h
alimpfard: Aside from the odd copies, it looks like they _are_ marked, through `Object::visit_children()`, `values_unordered()`.

alimpfard: It's a bit of a maze to go through, but I don't think your forgotten objects are here...

PR #3433, opened 2020-09-07
LibWeb: Add support for viewport relative length units

networkException:
You can now use vh, vw, vmax and vmin as length units in css

This will allow sites to render a little better and more responsively :^)

### DIFF
--- a/Libraries/LibWeb/CSS/Length.cpp
+++ b/Libraries/LibWeb/CSS/Length.cpp
@@ -27,6 +27,7 @@
 #include <LibWeb/CSS/Length.h>
 #include <LibWeb/DOM/Document.h>
 #include <LibWeb/HTML/HTMLHtmlElement.h>
+#include <LibWeb/Page/Frame.h>
 
 namespace Web::CSS {
 
@@ -39,6 +40,28 @@ float Length::relative_length_to_px(const LayoutNode& layout_node) const
         return m_value * layout_node.font_size();
     case Type::Rem:
         return m_value * layout_node.document().document_element()->layout_node()->font_size();
+    case Type::Vw:
+        return layout_node.document().frame()->viewport_rect().width() * (m_value / 100);
+    case Type::Vh:
+        return layout_node.document().frame()->viewport_rect().height() * (m_value / 100);
+    case Type::Vmin: {
+        auto viewport = layout_node.document().frame()->viewport_rect();
+
+        if (viewport.width() > viewport.height()) {
+            return viewport.height() * (m_value / 100);
+        } else {
+            return viewport.width() * (m_value / 100);
+        }
+    }
+    case Type::Vmax: {
+        auto viewport = layout_node.document().frame()->viewport_rect();
+
+        if (viewport.width() < viewport.height()) {
+            return viewport.height() * (m_value / 100);
+        } else {
+            return viewport.width() * (m_value / 100);
+        }
+    }
     default:
         ASSERT_NOT_REACHED();
     }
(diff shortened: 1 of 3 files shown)


Libraries/LibWeb/CSS/Length.cpp
```diff
@@ -39,6 +40,28 @@ float Length::relative_length_to_px(const LayoutNode& layout_node) const
...
+    case Type::Vh:
+        return layout_node.document().frame()->viewport_rect().height() * (m_value / 100);
+    case Type::Vmin: {
+        auto viewport = layout_node.document().frame()->viewport_rect();
+
+        if (viewport.width() > viewport.height()) {
+            return viewport.height() * (m_value / 100);
+        } else {
+            return viewport.width() * (m_value / 100);
+        }
```
linusg, line 54: if you want to be less repetitive/verbose this could be

```suggestion
        return min(viewport.width(), viewport.height()) * (m_value / 100);
```

Libraries/LibWeb/CSS/Length.cpp
```diff
@@ -39,6 +40,28 @@ float Length::relative_length_to_px(const LayoutNode& layout_node) const
...
+    }
+    case Type::Vmax: {
+        auto viewport = layout_node.document().frame()->viewport_rect();
+
+        if (viewport.width() < viewport.height()) {
+            return viewport.height() * (m_value / 100);
+        } else {
+            return viewport.width() * (m_value / 100);
+        }
```
linusg, line 63: same here, with `max()`

Libraries/LibWeb/CSS/Length.cpp
```diff
@@ -39,6 +40,28 @@ float Length::relative_length_to_px(const LayoutNode& layout_node) const
         return m_value * layout_node.font_size();
     case Type::Rem:
         return m_value * layout_node.document().document_element()->layout_node()->font_size();
+    case Type::Vw:
+        return layout_node.document().frame()->viewport_rect().width() * (m_value / 100);
+    case Type::Vh:
+        return layout_node.document().frame()->viewport_rect().height() * (m_value / 100);
+    case Type::Vmin: {
+        auto viewport = layout_node.document().frame()->viewport_rect();
```
linusg, line 48:
```suggestion
        auto& viewport = layout_node.document().frame()->viewport_rect();
```

networkException: This change would result in an error and I sadly don't have the expertise to dig further into the code

linusg: Ah, right - try this then:

```suggestion
        const auto& viewport = layout_node.document().frame()->viewport_rect();
```

networkException: That compiles, however if it's not really necessary I would spare myself having to deal with squashing (I can also just make a new commit however that's probably a little too atomic)

awesomekling: It's not the end of the world to copy this rect. If it turns out to be an issue someday in the future, we can easily tweak it :)

Libraries/LibWeb/CSS/Length.cpp
awesomekling: Agreed

Libraries/LibWeb/CSS/Length.cpp
awesomekling: Yup

Libraries/LibWeb/CSS/Length.cpp
networkException: Alright :)

PR #3434, opened 2020-09-07
Svg fixes

danners:
Hi,

i wanted to fix loading heise.de in the browser, with these changes to the SVG parsing, the parsing no longer fails for any svgs loaded on that side.
I also implemented another SVG drawing command, which seems to work at least for a path like:

    <path d="M 10 80 Q 52.5 10, 95 80 T 180 80" stroke="black" fill="transparent"/>


awesomekling:
Very nice :)

PR #3435, opened 2020-09-08
WindowServer: Inverted minimize animation

tomuta:
This makes it easier to see the minimize animation, e.g. above a maximized TextEditor.


awesomekling:
Love it :)

PR #3436, opened 2020-09-08
Physical page allocation

tomuta:
This make single-page allocations a little more lightweight. We don't need to scan for 1 "contiguous" range of bits, we can use simpler logic to just find one single available bit. And since when freeing a page we have an idea of where it may be, we can use that as a hint for next time.

### DIFF
--- a/AK/Bitmap.h
+++ b/AK/Bitmap.h
@@ -96,12 +96,6 @@ public:
         else
             m_data[index / 8] &= static_cast<u8>(~(1u << (index % 8)));
     }
-    void set_range(size_t start, size_t len, bool value)
-    {
-        for (size_t index = start; index < start + len; ++index) {
-            set(index, value);
-        }
-    }
 
     size_t count_slow(bool value) const
     {
@@ -159,6 +153,7 @@ public:
 
     void grow(size_t size, bool default_value)
     {
+        ASSERT(m_owned);
         ASSERT(size > m_size);
 
         auto previous_size_bytes = size_in_bytes();
@@ -179,7 +174,7 @@ public:
     }
 
     template<bool VALUE>
-    void fill_range(size_t start, size_t len)
+    void set_range(size_t start, size_t len)
     {
         ASSERT(start < m_size);
         ASSERT(start + len <= m_size);
@@ -217,12 +212,12 @@ public:
         }
     }
 
-    void fill_range(size_t start, size_t len, bool value)
+    void set_range(size_t start, size_t len, bool value)
     {
         if (value)
-            fill_range<true>(start, len);
+            set_range<true>(start, len);
         else
-            fill_range<false>(start, len);
+            set_range<false>(start, len);
     }
 
     void fill(bool value)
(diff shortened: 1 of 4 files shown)


awesomekling:
It would be good to make physical page allocation less deterministic (by not reliably returning the most recently freed page) to make exploitation harder.

tomuta:
Yeah I actually thought about that (effectively doing `m_free_hint += 8` on every single-page allocation rather than setting it when freeing a page, but I think that may cause problems for contiguous allocations because it would spread out single page allocations all over the place. I think what might come in handy is a zero-page task that would take the pages, and then after zeroing them it would actually mark them as "free again". This would also make allocations faster as we could skip the page zeroing as long as we can draw from a pool of already zeroed pages.

awesomekling:
Pre-zeroing pages is not an optimization on modern systems (and sometimes a pessimization) according to DragonFlyBSD:
http://lists.dragonflybsd.org/pipermail/commits/2016-August/624202.html

I think something like your buddy allocator thingy could make sense here too. It would make returning contiguous memory pretty easy if all we have to do is split until we have a good size (instead of having to scan)

For the security aspect, what do you think about a "delayed release" mechanism where freed pages are put into a queue which is periodically fed back into the allocator (or drained by the allocator synchronously if needed)?

tomuta:
Hm interesting about the per-zeroing, I guess I'll just abandon that thought ;-) Makes sense why it's not really that useful anymore.

Are you saying we should split physical pages into halves? I don't really see that much useful, because at that level we always want at least a full page.

The buddy allocator still sits on top of the bitmap, so if there isn't any link to a freed buddy, it will still have to scan the bitmap for a free chunk. Applying that to the physical page allocator would be similar. The physical page allocator has another thing that makes it a little trickier: We can't access the pages unless we quickmap them, which is more expensive than just following already mapped pointers (because the kmalloc pool is already entirely accessible in the virtual address space).

How about some sort of a ringbuffer that we write freed pages into, and once full pick a random position and actually return the freed page (and move the ringbuffer pointer to point to the freed slot)? Not truly random.

Another option would be to sort of reserve a contiguous physical page range (not sure how we could determine a reasonable size upfront), and randomly distribute allocation of physical pages from the remaining physical address space, e.g. by simply doing `m_free_hint += 8` at allocation time and not touching it when freeing a page? And if allocating contiguous memory, only look at the reserved range in which we wouldn't do this "randomization"?

tomuta:
By essentially increasing the `m_free_hint` by 8 we would effectively allocate 4 pages in each 32 bit slot in the bitmap, then move onto the next. Or we just increase it by 32, that way we would distribute it even more evenly. Not really "random" but if you free a page and allocate another you'll end up in the next slot, therefore unlikely to get exactly the same page again.

awesomekling:
> Are you saying we should split physical pages into halves? I don't really see that much useful, because at that level we always want at least a full page.

Nah, I meant clustering them more like the Linux and various other allocators do (to my understanding). Dividing it into levels that can be split to accommodate allocation requests. Each level would be a power-of-two number of pages, with 1 page being the smallest unit.

The approach described here seems pretty neat: https://www.kernel.org/doc/gorman/html/understand/understand009.html

> How about some sort of a ringbuffer that we write freed pages into, and once full pick a random position and actually return the freed page (and move the ringbuffer pointer to point to the freed slot)? Not truly random.

That sounds kinda nifty, and easy to implement too. :)

tomuta:
That's indeed an interesting approach. But I think it would require a bit more rework than I'm willing to do right now. In the meanwhile I added that return queue that should hopefully make it just a bit more random ;-)

Issue #3419, opened 2020-09-06
Shell for loop can't be interrupted, and causes kernel crash


nico:
1. Run `for i in $(seq 1000) { sleep 1; ls }`
2. Hit ctrl-c a few times (or keep it depressed)

Expected: Quits loop
Actual: Ignored, then triggers assert in kernel.

@alimpfard for first bug, @tomuta for second :)

```
Shell(22): Command "sleep 1" finished in 31 ms
Shell(22): Command "ls" finished in 10 ms
[Terminal(7:7)]: /dev/pts/0: VINTR pressed!
[Terminal(7:7)]: /dev/pts/0: Send signal 2 to everyone in pgrp 85
[Terminal(7:7)]: /dev/pts/0: Send signal 2 to sleep(85)
Shell(22): Command "sleep 1" finished in 44 ms
Shell(22): Command "ls" finished in 13 ms
[Terminal(7:7)]: /dev/pts/0: VINTR pressed!
[Terminal(7:7)]: /dev/pts/0: Send signal 2 to everyone in pgrp 87
[Terminal(7:7)]: /dev/pts/0: Send signal 2 to Shell(87)
[Terminal(7:7)]: ASSERTION FAILED: Kernel::is_user_range(VirtualAddress(dest_ptr), n)
../Kernel/StdLib.cpp:47 in void copy_to_user(void*, const void*, size_t)
[Terminal(7:7)]: 0xc011eaed  __assertion_failed(char const*, char const*, unsigned int, char const*) +185
[Terminal(7:7)]: 0xc01677d0  copy_to_user +111
[Terminal(7:7)]: 0xc018192e  Kernel::push_value_on_user_stack(unsigned int*, unsigned int) +42
[Terminal(7:7)]: 0xc0182ce5  .L321 +239
[Terminal(7:7)]: 0xc0182f3d  Kernel::Thread::dispatch_one_pending_signal() +147
[Terminal(7:7)]: 0xc01659d1  Kernel::Thread::for_each_living<Kernel::Scheduler::pick_next()::{lambda(Kernel::Thread&)#4}>(Kernel::Scheduler::pick_next()::{lambda(Kernel::Thread&)#4})::{lambda(Kernel::Thread&)#1}::operator()(Kernel::Thread&) const [clone .isra.0] +93
[Terminal(7:7)]: 0xc01663b0  Kernel::Scheduler::pick_next() +996
[...]

PR #3438, opened 2020-09-08
sleep: On SIGINT, call default SIGINT handler after printing remaining time

nico:
With this, hitting ctrl-c twice in `for i in $(seq 10) { sleep 1 }`
terminates the loop as expected (...well, I'd expect it to quit after
just one ctrl-c, but serenity's shell makes a single ctrl-c only
quit the current loop iteration).

Part of #3419.

### DIFF
--- a/Userland/sleep.cpp
+++ b/Userland/sleep.cpp
@@ -30,8 +30,10 @@
 #include <string.h>
 #include <unistd.h>
 
+static bool g_interrupted;
 static void handle_sigint(int)
 {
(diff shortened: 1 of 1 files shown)


alimpfard:
> ...well, I'd expect it to quit after
just one ctrl-c, but serenity's shell makes a single ctrl-c only
quit the current loop iteration

Guilty as charged, I like having the ability to quit any one iteration, it comes in handy :P

Issue #1234, opened 2020-02-17
Add err/warn, switch from perror/ASSERT_NOT_REACHED


jcs:
A common pattern in use is:

````
perror("open");
ASSERT_NOT_REACHED();
````

The output of this looks like so:

````
open: No such file or directory
assertion "false" failed: file "Service.cpp", line 226, function "spawn"
````

The failure message of "open" is not very helpful without the filename it was trying to open, nor the program that output it when run from something like SystemServer.  I'd like to implement the [BSD family of error functions](http://man.openbsd.org/err) in Serenity's LibC (at least `err` and `warn` for now) which take variadic arguments and change uses of `perror` to be something like:

````
err(1, "open: %s", m_stdio_file_path.characters());
````

Which prints a more helpful message with the program name, message, and filename, and then `exit`s with a code of 1.  This is in contrast to `ASSERT_NOT_REACHED` which terminates the program abruptly without the normal exit path.

````
SystemServer: open: /dev/tty1: No such file or directory
````

Assertions can still be used as sanity checks, but for normal error handling, I think using `err` or `warn` would be better.

What say you?

awesomekling:
I like the idea and I think it would be even better if err() and warn() worked like dbg() so we don’t have to use format strings. :)

jcs:
I thought I could get away with `err.h` defining the BSD interfaces of `err(int, const char *, ...)` and `AK/LogStream.h` defining `FatalErrorLogStream err(int);` to have `err(1) << "blah"` but clang doesn't seem to like that.  I guess we have to call the C++ `err` interface something else?

bugaevc:
Why not have *just* the C++ `err` and `warn`, without the C functions?

I'm not sure it's that nice if they exit automatically; I can call `exit(1)` on the next line explicitly. OTOH `ASSERT_NOT_REACHED()` prints a nice stack trace to the console so you know immediately where the problem was [detected] *in the code*.

awesomekling:
I agree with @bugaevc, I think we should do just the C++ variants.

It does seem neat to have a variant that exits automatically, although perhaps we could call it something even more ominous like `fatal()` to really drive the point home.

awesomekling:
Oh, and getting stack traces is v nice. But we can just make these guys call `dump_backtrace()` before the auto exit.

jcs:
[...]

PR #3442, opened 2020-09-09
JPGLoader: Check existense of huffman tables in scan header segment

devsh0:


Mindavi:
This patch seems to work fine to resolve the issue. I can't comment on the JPG format itself, but it does stop the assertion from being hit :).

awesomekling:
If this fixes a bug, please put "Fixes #1234" in the commit message so it gets closed when merged :)

devsh0:
Wait...don't merge.

devsh0:
The solution would have been slightly more elegant if I could correctly use HashMap but things went wrong for me if when I replaced the vectors with hashmaps. Anyways, the bug is fixed and patch is ready to be merged.

Libraries/LibGfx/JPGLoader.cpp
```diff
@@ -536,8 +536,31 @@ static bool read_start_of_scan(BufferStream& stream, JPGLoadingContext& context)
         stream >> table_ids;
         if (stream.handle_read_failure())
             return false;
+
         component->dc_destination_id = table_ids >> 4;
         component->ac_destination_id = table_ids & 0x0F;
+
+        if (context.dc_tables.size() != context.ac_tables.size()) {
+            dbg() << stream.offset() << String::format(": DC & AC table count mismatch!");
```
awesomekling, line 544:
```suggestion
            dbg() << stream.offset() << ": DC & AC table count mismatch!";
```

Libraries/LibGfx/JPGLoader.cpp
```diff
@@ -536,8 +536,31 @@ static bool read_start_of_scan(BufferStream& stream, JPGLoadingContext& context)
...
+
+        bool dc_table_exists = false;
+        bool ac_table_exists = false;
+        for (unsigned t = 0; t < context.dc_tables.size(); t++) {
+            dc_table_exists = dc_table_exists || (context.dc_tables[t].destination_id == component->dc_destination_id);
+            ac_table_exists = ac_table_exists || (context.ac_tables[t].destination_id == component->ac_destination_id);
+        }
+
+        if (!dc_table_exists) {
+            dbg() << stream.offset() << String::format(": Invalid DC huffman table destination id: %i", component->dc_destination_id);
```
awesomekling, line 556:
```suggestion
            dbg() << stream.offset() << ": Invalid DC huffman table destination id: " << component->dc_destination_id;
```

Libraries/LibGfx/JPGLoader.cpp
```diff
@@ -536,8 +536,31 @@ static bool read_start_of_scan(BufferStream& stream, JPGLoadingContext& context)
...
+            return false;
+        }
+
+        if (!ac_table_exists) {
+            dbg() << stream.offset() << String::format(": Invalid AC huffman table destination id: %i", component->ac_destination_id);
```
awesomekling, line 561:
```suggestion
            dbg() << stream.offset() << ": Invalid AC huffman table destination id: " << component->ac_destination_id;
```

Libraries/LibGfx/JPGLoader.cpp
```diff
@@ -576,6 +599,31 @@ static bool read_reset_marker(BufferStream& stream, JPGLoadingContext& context)
...
 
+static bool huffman_table_reset_helper(HuffmanTableSpec& src, JPGLoadingContext& context)
+{
+    auto& table = src.type == 0 ? context.dc_tables : context.ac_tables;
+    if (src.destination_id == 0) {
+        if (table.is_empty())
+            table.append(move(src));
+        else
+            table[0] = move(src);
+    }
```
awesomekling, line 610: I would `return true;` inside this block. Since you are doing a `move()` on `src`, it is logically invalid after that point, and so the code is a lot less confusing if we don't stick around longer than necessary.

Libraries/LibGfx/JPGLoader.cpp
```diff
@@ -576,6 +599,31 @@ static bool read_reset_marker(BufferStream& stream, JPGLoadingContext& context)
...
+
+    if (src.destination_id == 1) {
+        if (table.size() < 1) {
+            String table_str = src.type == 0 ? "DC" : "AC";
+            dbg() << String::format("%s[1] showed up before %s[0]!", table_str, table_str);
```
awesomekling, line 615: You cannot pass a `String` directly to `String::format()` like this, you need to pass the `String::characters()` (same as `printf`).
Also, let's not create unnecessary temporary strings with `String::format()` when we can just stream to `dbg()`:

```suggestion
            dbg() << table_str << "[1] showed up before " << table_str << "[0]!";
```

Libraries/LibGfx/JPGLoader.cpp
```diff
@@ -576,6 +599,31 @@ static bool read_reset_marker(BufferStream& stream, JPGLoadingContext& context)
...
+            return false;
+        }
+        if (table.size() == 1)
+            table.append(move(src));
+        else
+            table[1] = move(src);
+    }
```
awesomekling, line 622: I would suggest returning here as well, and then having some kind of fail state if `destination_id` is neither `0` nor `1`.

awesomekling (changes requested):

awesomekling:
:star:

PR #3444, opened 2020-09-09
sleep: Make variable written in signal handler volatile

nico:
No difference in practice, but it's tidier and protects us
if we ever use g_interrupted earlier in main -- in that case,
the compiler might chose to keep the variable in a register without
the volatile.

Found by @bugaevc, thanks!

### DIFF
--- a/Userland/sleep.cpp
+++ b/Userland/sleep.cpp
@@ -30,7 +30,7 @@
 #include <string.h>
 #include <unistd.h>
 
-static bool g_interrupted;
+static volatile bool g_interrupted;
 static void handle_sigint(int)
 {
     g_interrupted = true;


bugaevc:
@nico it occurred to me that there's another thing that probably doesn't matter in practice, but could, theoretically.

If you're checking for `g_interrupted` first and resetting the signal disposition to default after that (and only if you were indeed interrupted), there's a window right after the check (and before exiting) where you could be interrupted, and you would even notice it (as in, save it into `g_interrupted`), but you still would exit successfully, because you're already past the check. But this order should work:

```cpp
signal(SIGINT, SIG_DFL);
if (g_interrupted)
    raise(SIGINT);
return 0;
```

bugaevc:
Submitted as https://github.com/SerenityOS/serenity/pull/3450

PR #3445, opened 2020-09-09
LibGUI: Do not paint scrollbar button highlight for scrollbars without scrubber

nico:
If a scrollbar doesn't have a scrubber (because the view it scrolls is
large enough to display all its contents without scrolling), then
it ignores all clicks. We shouldn't draw a hover highlight that suggests
clickability in that case.


awesomekling:
Oh hey good point!

PR #3446, opened 2020-09-09
Shell: Do not reset the terminal attributes when command is run in bg 

alimpfard:
cc @nico.

Fixes:
```sh
$ help&
%aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa *sad LibLine noises*
```


nico:
Nice! I patched this in and verified that it fixes my issue.

PR #3447, opened 2020-09-09
LibVT: Make Terminal::scroll_up() O(1) in size of history by using a circular buffer

nico:
Reduces time to run `disasm /bin/id` with the default terminal window size from 530ms to 409ms (min-of-5) on my system.

I likely want to do something very similar to m_lines, but I'm out of evening. Probably want to extract some kind of CircularVectorView or similar after that's done. (The existing Circular classes in AK don't quite do what I need, as far as I can tell.)


awesomekling:
Ahh excellent. I love that this optimization train is still rolling :)

PR #3449, opened 2020-09-10
AK: Replace LogStream operator for ReadonlyBytes with dump_bytes function.

asynts:
This pull request fixes a bug I introduced in #3409.

It wasn't actually possible to call

~~~c++
const LogStream& operator<<(const LogStream&, ReadonlyBytes);
~~~

because it was shadowed by

~~~c++
template<typename T>
const LogStream& operator<<(const LogStream& stream, Span<T> span);
~~~

not sure how I didn't notice when I added the overload.

It would be possible to use SFINAE to disable the other overload, however, I think it is better to use a different function entirely because the output can be very verbose:

~~~c++
void dump_bytes(ReadonlyBytes);
~~~

(I originally wanted to put the implementation into `Span.cpp` but that file didn't exist yet and it seemed unnecessary for one small function.)

### DIFF
--- a/AK/LogStream.cpp
+++ b/AK/LogStream.cpp
@@ -209,7 +209,7 @@ const LogStream& operator<<(const LogStream& stream, float value)
 
 #endif
 
-const LogStream& operator<<(const LogStream& stream, ReadonlyBytes bytes)
+void dump_bytes(ReadonlyBytes bytes)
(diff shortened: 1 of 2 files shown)


awesomekling:
Yeah this seems more sensible anyway :)

PR #3453, opened 2020-09-10
Kernel: Fix detecting in what ring a crash happened

tomuta:
The ring is determined based on the CS register. This fixes crashes
being handled as ring 3 crashes even though EIP/CS clearly showed
that the crash happened in the kernel.


awesomekling:
Oh nice! :)

PR #3454, opened 2020-09-10
Ports: Add OpenSSH port

Lubrsi:
Demo of connecting to a DO server:
w2a8Th4g4w

Known limitiations:
- No SK/FIDO support.
- No DNS support.
- No proxy support.
- Assumes SSH2.0 for now.
- Cannot determine compatibility flags. 
  This means there may be some weird bugs when connecting to certain SSH implementations.
- SSHD does not work as it requires socketpair. It will start, but will crash on connection.

Remaining concerns:
- For some reason, the limits.h header doesn't work in OpenSSH. It includes it correctly, but for some reason all the macros are undeclared?
Here's the comment I added in serenity_limits.h in the port:
```c
/*
 * This header exists because limits.h just straight up does not work.
 * OpenSSH requires PATH_MAX and SSIZE_MAX and while all the files that require it do include limits.h, it just does not work.
 * For example:
 * auth.c:430:30: error: 'PATH_MAX' undeclared (first use in this function)
 *   430 |  char *file, uidstr[32], ret[PATH_MAX];
 *       |
 * But, defining them here does work? No idea why any of this happens.
 */
```
- I'm not sure if the PT_GNU_RELRO validation is correct.
- I'm not sure if ss_family is in the right spot in sockaddr_storage.


Libraries/LibC/errno_numbers.h
```diff
@@ -99,3 +99,4 @@
 #define EPROTO 71
 #define ENOTSUP 72
 #define EMAXERRNO 73
+#define EPFNOSUPPORT 74
```
bugaevc, line 102: If you add a new error, it should go before `EMAXERRNO`, and be recognized by `strerror()`

Libraries/LibC/ctype.h
```diff
@@ -57,18 +57,6 @@ int isgraph(int);
...
-#define isalpha(c) (_ctype_[(unsigned char)(c)] & (_U | _L))
-#define iscntrl(c) (_ctype_[(unsigned char)(c)] & (_C))
-#define isdigit(c) (_ctype_[(unsigned char)(c)] & (_N))
-#define isxdigit(c) (_ctype_[(unsigned char)(c)] & (_N | _X))
-#define isspace(c) (_ctype_[(unsigned char)(c)] & (_S))
-#define ispunct(c) (_ctype_[(unsigned char)(c)] & (_P))
-#define isprint(c) (_ctype_[(unsigned char)(c)] & (_P | _U | _L | _N | _B))
-#define isgraph(c) (_ctype_[(unsigned char)(c)] & (_P | _U | _L | _N))
-#define islower(c) ((_ctype_[(unsigned char)(c)] & (_U | _L)) == _L)
-#define isupper(c) ((_ctype_[(unsigned char)(c)] & (_U | _L)) == _U)
```
bugaevc, line 70: Could we at least have them as `static inline` functions in the header? It'd be sad to lose inlining for them.

bugaevc:
> SSH keygen does not work as `/usr` is read-only.

The keys should go into `/etc`, not `/usr`. But if you want to, root can always enter rw mode like this:

```
# chroot / -o rw
# stuff > /usr/foo
```

FireFox317:
Can't you just have these commits that are in the fork, be as patch files, which is what is normally done? This way there is no for the separate fork :)

Lubrsi:
> > SSH keygen does not work as `/usr` is read-only.
> 
> The keys should go into `/etc`, not `/usr`. But if you want to, root can always enter rw mode like this:
> 
> ```
> # chroot / -o rw
> # stuff > /usr/foo
> ```

Ah, didn't know about that :^)
Plus, the reason it goes into usr is because it's by default for some reason. Probably because of the prefix.

Lubrsi:
> Can't you just have these commits that are in the fork, be as patch files, which is what is normally done? This way there is no for the separate fork :)

Sure, I'll have a look into it

bugaevc:
> Plus, the reason it goes into usr is because it's by default for some reason. Probably because of the prefix.

Well, please figure it out. You may have to patch OpenSSH, but a per-machine generated key is clearly `/etc` material, not `/usr`. (And it's telling that the path is `/usr/local/etc/`.)

Lubrsi:
Turns it they thankfully have a config option for it:
```
--sysconfdir=DIR        read-only single-machine data [PREFIX/etc]
```

Lubrsi:
Should I remove the pre generated keys or keep them?

Libraries/LibC/errno_numbers.h
Lubrsi: I tried doing this. While I thought it worked at first, it actually fails with undefined references when I rebuilt. Here's my ctype.h: https://gist.github.com/Lubrsi/17b4db531338f6a53692086a47cbbd59

Libraries/LibC/ctype.h
Lubrsi: I tried doing this. While I thought it worked at first, it actually fails with undefined references when I rebuilt. Here's my ctype.h: https://gist.github.com/Lubrsi/17b4db531338f6a53692086a47cbbd59
And the log: https://gist.github.com/Lubrsi/e6912f26216949a93a8f5f54f119885d
Put that on the completely wrong comment 😄

Lubrsi: Nevermind, I did a clean rebuild today and it worked.

Lubrsi:
I think this is ready. I've added patch files but I'm gonna keep the repo open as it makes it easier for me to generate patch files. I've updated the original PR message with remaining concerns.

Lubrsi:
```
/home/travis/.travis/functions: line 109: 22014 Segmentation fault      (core dumped) DISABLE_DBG_OUTPUT=1 ./test-js
The command "DISABLE_DBG_OUTPUT=1 ./test-js" exited with 139.
```
Weird build failure.

awesomekling:
Do the GCC and Binutils ports still build with this change? They have been weirdly sensitive to ctype.h changes in the past :|

Lubrsi:
Seems to, did a fresh build for them and tried them on HackStudio.

PR #3455, opened 2020-09-11
Refactor and fix bugs in LibCompress.

asynts:
There are still a few nasty bugs in the Deflate implementation. I have to be honest, I am having some trouble tracking them down. In the meantime, I found a really really nasty typo in `DuplexMemoryStream::read_without_consuming` and cleaned up some code in general.

I hope that I can remove the Deflate and Gzip implementation in LibCore soon, but that's not possible at the moment.

### DIFF
--- a/AK/MemoryStream.h
+++ b/AK/MemoryStream.h
@@ -232,7 +232,7 @@ public:
     {
         size_t nread = 0;
         while (bytes.size() - nread > 0 && m_write_offset - m_read_offset - nread > 0) {
-            const auto chunk_index = (m_read_offset - m_base_offset) / chunk_size;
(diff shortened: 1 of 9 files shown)


asynts:
Actually, that `DuplexMemoryStream` bug seemed to be responsible for all failures I was aware of. It seems to be possible to decompress arbitrary files. However, the performance is still really bad.

nico:
If you haven't seen it, https://www.hanshq.net/zip.html has a great explanation and very readable code describing a performant deflate codec (it doesn't do streaming though, mostly for clarity of exposition). Maybe it can serve as implementation inspiration -- binary searching during decompression shouldn't be needed for example :)

asynts:
@nico Thanks, that looks like a really good resource, I am definitely checking that out.

PR #3456, opened 2020-09-11
LibCore: Fixed DeferredInvoke debug message

pkotzbach:


awesomekling:
Great, thank you for patching that up :)

PR #3457, opened 2020-09-11
LibJS: Separate run() into "public" and "private" API, fix TryStatement finalizers, try/catch/except tests

### LINKED ISSUE #3437: LibJS: incorrect assertion when exception is thrown in catch block when finally block is present
Some more fuzzing results :smile: 

I think the assertion that's fired when an exception is thrown in a `catch` block is invalid.

Example input:

```js
try {
    foo();
} catch (e) {
    bar();
} finally {
  var a = "abc";
}
```

spidermonkey output:

```
assert_no_exception:4:4 ReferenceError: bar is not defined
```

serenity `js` output:

```
js: /serenity/Libraries/LibJS/Interpreter.cpp:63: JS::Value JS::Interpreter::run(JS::GlobalObject&, const JS::Statement&, JS::ArgumentVector, JS::ScopeType): Assertion `!exception()' failed.
Aborted (core dumped)
```

Maybe a more 'idiomatic' example might be this:

```js
try {
    foo();
} catch (e) {
    throw Error(42);
} finally {
  var a = "abc";
}
```

I'm not sure if the assertion should just be removed or if this should be handled in another way. You could nest these try catch blocks (I think), so that should be handled correctly as well.

linusg:
**LibJS: Extract most of `Interpreter`'s `run()` into `execute_statement()`**

`Interpreter::run()` was so far being used both as the "public API entry point" for running a `JS::Program` as well as internally to execute `JS::Statement`s of all kinds - this is now more distinctly separated.
A program as returned by the parser is still going through `run()`, which is responsible for creating the initial global call frame, but all other statements are executed via `execute_statement()` directly.

Fixes #3437, a regression introduced by adding `ASSERT(!exception())` to `run()` without considering the effects that would have on internal usage.

**LibJS: Stop unwinding and reset exception for `TryStatement` finalizer**

This fixes two issues with running a `TryStatement` finalizer:

- Temporarily store and clear the exception, if any, so we can run the finalizer block statement without it getting in our way, which could have unexpected side effects otherwise (and will likely return early somewhere).
- Stop unwinding so more than one child node of the finalizer `BlockStatement` is executed if an exception has been thrown previously (which would have called `unwind(ScopeType::Try)`). Re-throwing as described above ensures we still unwind after the finalizer, if necessary.

Also add some tests specifically for try/catch/finally blocks, we didn't have any!

### DIFF
--- a/Libraries/LibJS/AST.cpp
+++ b/Libraries/LibJS/AST.cpp
@@ -1786,8 +1786,20 @@ Value TryStatement::execute(Interpreter& interpreter, GlobalObject& global_objec
         }
     }
 
-    if (m_finalizer)
+    if (m_finalizer) {
+        // Keep, if any, and then clear the current exception so we can
+        // execute() the finalizer without an exception in our way.
+        auto* previous_exception = interpreter.exception();
+        interpreter.clear_exception();
+        interpreter.stop_unwind();
         m_finalizer->execute(interpreter, global_object);
+        // If we previously had an exception and the finalizer didn't
+        // throw a new one, restore the old one.
+        // FIXME: This will print debug output in throw_exception() for
+        // a seconds time with INTERPRETER_DEBUG enabled.
+        if (previous_exception && !interpreter.exception())
+            interpreter.throw_exception(previous_exception);
+    }
 
     return js_undefined();
 }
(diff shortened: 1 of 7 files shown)


awesomekling:
Very excellent :) Thank you for fixing this Linus!

PR #3465, opened 2020-09-12
LibGfx+LibWeb: clone GIF frames instead of returning underlying frame buffer

peterdn:
This PR:
1. LibGfx: add Bitmap::clone() method
2. LibWeb: use RefPtr instead of raw pointer for ImageResource bitmap refs
3. LibGfx: return clone from frame() instead of underlying framebuffer

Fixes GIF animation in ImageWidget

### DIFF
--- a/Libraries/LibGfx/Bitmap.cpp
+++ b/Libraries/LibGfx/Bitmap.cpp
@@ -182,6 +182,25 @@ Bitmap::Bitmap(BitmapFormat format, NonnullRefPtr<SharedBuffer>&& shared_buffer,
         allocate_palette_from_format(m_format, palette);
 }
 
+RefPtr<Gfx::Bitmap> Bitmap::clone() const
+{
+    RefPtr<Gfx::Bitmap> new_bitmap {};
+    if (m_purgeable) {
+        new_bitmap = Bitmap::create_purgeable(format(), size());
+    } else {
+        new_bitmap = Bitmap::create(format(), size());
+    }
+
+    if (!new_bitmap) {
+        return nullptr;
+    }
+
+    ASSERT(size_in_bytes() == new_bitmap->size_in_bytes());
+    memcpy(new_bitmap->scanline(0), scanline(0), size_in_bytes());
+
+    return new_bitmap;
+}
+
 RefPtr<Gfx::Bitmap> Bitmap::rotated(Gfx::RotationDirection rotation_direction) const
 {
     auto w = this->width();
--- a/Libraries/LibGfx/Bitmap.cpp
+++ b/Libraries/LibGfx/Bitmap.cpp
@@ -26,6 +26,7 @@
 
 #include <AK/Checked.h>
 #include <AK/Memory.h>
+#include <AK/Optional.h>
 #include <AK/SharedBuffer.h>
 #include <AK/String.h>
 #include <LibGfx/BMPLoader.h>
@@ -43,6 +44,12 @@
 #include <sys/mman.h>
 
 namespace Gfx {
+
+struct BackingStore {
+    void* data { nullptr };
+    size_t pitch { 0 };
+    size_t size_in_bytes { 0 };
+};
 
 size_t Bitmap::minimum_pitch(size_t width, BitmapFormat format)
 {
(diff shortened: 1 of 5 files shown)


Libraries/LibGfx/Bitmap.cpp
```diff
@@ -182,6 +182,23 @@ Bitmap::Bitmap(BitmapFormat format, NonnullRefPtr<SharedBuffer>&& shared_buffer,
...
+RefPtr<Gfx::Bitmap> Bitmap::clone() const
+{
+    const auto w = this->width();
+    const auto h = this->height();
+    const auto purgeable = this->m_purgeable ? Purgeable::Yes : Purgeable::No;
+
+    auto new_bitmap = new Bitmap(this->format(), { w, h }, purgeable);
+    if (!new_bitmap) {
+        return nullptr;
+    }
```
awesomekling, line 194: `new` itself will not fail here since it's only concerned with heap-allocating `sizeof(Bitmap)`
The real issue is the size of the backing store for the bitmap.
We should reorganize `Gfx::Bitmap` construction so that the backing store is allocated before we create and return a `Gfx::Bitmap`.

Libraries/LibGfx/Bitmap.cpp
```diff
@@ -182,6 +182,28 @@ Bitmap::Bitmap(BitmapFormat format, NonnullRefPtr<SharedBuffer>&& shared_buffer,
         allocate_palette_from_format(m_format, palette);
 }
 
+RefPtr<Gfx::Bitmap> Bitmap::clone() const
+{
+    const auto w = this->width();
+    const auto h = this->height();
```
awesomekling, line 182: No need to specify `this` all the time, it's implied from context :)
Also, you can use `Bitmap::size()` directly instead of making a temporary `IntSize`.

Libraries/LibWeb/Loader/ImageLoader.h
```diff
@@ -38,7 +38,7 @@ class ImageLoader : public ImageResourceClient {
 
     void load(const URL&);
 
-    const Gfx::Bitmap* bitmap() const;
+    const RefPtr<Gfx::Bitmap> bitmap() const;
```
awesomekling, line 41: (Since I can't comment on the commit message of this commit, I'll comment on one of the changes)

This doesn't seem accurate. The `ImageResource` object still has a `RefPtr` member which will keep the bitmap alive as long as it's pointed to. This patch is only changing the `ImageResource` API to return a `RefPtr` rather than a `const Gfx::Bitmap*`.

I don't think this patch fixes anything other than introducing some additional reference count churn whenever `bitmap()` is called.

Also, note that:
`const RefPtr<Gfx::Bitmap>` is a `RefPtr` to a **mutable** `Gfx::Bitmap` that cannot be retargeted.
`RefPtr<const Gfx::Bitmap>` would be a `RefPtr` to a **const** `Gfx::Bitmap`. :)

Libraries/LibGfx/Bitmap.cpp
peterdn: That makes sense. I've restructured the above to use the static `create()` methods, so as to as least limit where `new`s are happening.

Also had a bash at your reorganisation suggestion by moving backing store allocation to the appropriate static `create()` methods. I've introduced a `BackingStore` convenience struct so that we don't have to risk recalculating the pitch in several different places.

I'm still a little nervous that the size of the allocated memory chunk is not coupled to the backing store at all, as the latter is ultimately just a `void*`. That said, the constructor is private so maybe it's not a big deal.

Libraries/LibGfx/Bitmap.h
```diff
@@ -79,6 +79,13 @@ static StorageFormat determine_storage_format(BitmapFormat format)
     }
 }
 
+struct BackingStore {
+public:
```
awesomekling, line 83: You can omit `public:` here, `struct` members are public by default :)

Libraries/LibGfx/Bitmap.h
```diff
@@ -79,6 +79,13 @@ static StorageFormat determine_storage_format(BitmapFormat format)
...
+    void* data;
+    size_t pitch;
+    size_t size_in_bytes;
```
awesomekling, line 86: Please always initialize primitive variables:
```suggestion
    void* data { nullptr };
    size_t pitch { 0 };
    size_t size_in_bytes { 0 };
```

Libraries/LibGfx/Bitmap.h
```diff
@@ -79,6 +79,13 @@ static StorageFormat determine_storage_format(BitmapFormat format)
     }
 }
 
+struct BackingStore {
```
awesomekling, line 82: Since this is not part of the `Bitmap` public API, it would be better to move it into `Bitmap.cpp` to hide it from the world.

Libraries/LibWeb/Loader/ImageLoader.h
peterdn: Ah OK. This was fixing an apparent use-after-free crash I was seeing with gifs in Browser. But it looks like ImageResource is not caching the gif frame itself, so it's being freed after returning the underlying pointer: https://github.com/SerenityOS/serenity/blob/master/Libraries/LibWeb/Loader/ImageResource.cpp#L65

I'll change it so it does cache the gif frame in the appropriate `RefPtr` member.

awesomekling:
:star:

BenWiederhake:
I like the idea of a `BackingStore` :D

PR #3467, opened 2020-09-12
Kernel: Make copy_to/from_user safe and remove unnecessary checks

tomuta:
Since the CPU already does almost all necessary validation steps
for us, we don't really need to attempt to do this. Doing it
ourselves doesn't really work very reliably, because we'd have to
account for other processors modifying virtual memory, and we'd
have to account for e.g. pages not being able to be allocated
due to insufficient resources.

So change the copy_to/from_user (and associated helper functions)
to use the new safe_memcpy, which will return whether it succeeded
or not. The only manual validation step needed (which the CPU
can't perform for us) is making sure the pointers provided by user
mode aren't pointing to kernel mappings.

To make it easier to read/write from/to either kernel or user mode
data add the UserOrKernelBuffer helper class, which will internally
either use copy_from/to_user or directly memcpy, or pass the data
through directly using a temporary buffer on the stack.

This is needed to be able to continue the work started in #3396

### DIFF
--- a/Kernel/UserOrKernelBuffer.cpp
+++ b/Kernel/UserOrKernelBuffer.cpp
@@ -1,0 +1,88 @@
+/*
+ * Copyright (c) 2020, the SerenityOS developers.
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include <Kernel/UserOrKernelBuffer.h>
+#include <Kernel/VM/MemoryManager.h>
+
+namespace Kernel {
+
+bool UserOrKernelBuffer::is_kernel_buffer() const
+{
+    return !is_user_address(VirtualAddress(m_buffer));
+}
+
+String UserOrKernelBuffer::copy_into_string(size_t size) const
+{
+    if (!m_buffer)
+        return {};
+    if (is_user_address(VirtualAddress(m_buffer))) {
+        char* buffer;
+        auto data_copy = StringImpl::create_uninitialized(size, buffer);
+        if (!copy_from_user(buffer, m_buffer, size))
+            return {};
+        return data_copy;
+    }
+
+    return String({ m_buffer, size });
+}
+
+bool UserOrKernelBuffer::write(const void* src, size_t offset, size_t len)
+{
+    if (!m_buffer)
+        return false;
+
+    if (is_user_address(VirtualAddress(m_buffer)))
+        return copy_to_user(m_buffer + offset, src, len);
+
+    memcpy(m_buffer + offset, src, len);
+    return true;
+}
+
+bool UserOrKernelBuffer::read(void* dest, size_t offset, size_t len) const
+{
+    if (!m_buffer)
+        return false;
+
+    if (is_user_address(VirtualAddress(m_buffer)))
+        return copy_from_user(dest, m_buffer + offset, len);
+
+    memcpy(dest, m_buffer + offset, len);
+    return true;
+}
+
+bool UserOrKernelBuffer::memset(int value, size_t offset, size_t len)
+{
+    if (!m_buffer)
+        return false;
+
+    if (is_user_address(VirtualAddress(m_buffer)))
+        return memset_user(m_buffer + offset, value, len);
+    
+    ::memset(m_buffer + offset, value, len);
+    return true;
+}
+
+}
--- a/DevTools/UserspaceEmulator/Emulator.cpp
+++ b/DevTools/UserspaceEmulator/Emulator.cpp
@@ -930,7 +930,7 @@ int Emulator::virt$execve(FlatPtr params_addr)
     auto copy_string_list = [this](auto& output_vector, auto& string_list) {
         for (size_t i = 0; i < string_list.length; ++i) {
             Syscall::StringArgument string;
-            mmu().copy_from_vm(&string, (FlatPtr)&string_list.strings.ptr()[i], sizeof(string));
+            mmu().copy_from_vm(&string, (FlatPtr)&string_list.strings[i], sizeof(string));
             output_vector.append(String::copy(mmu().copy_buffer_from_vm((FlatPtr)string.characters, string.length)));
         }
     };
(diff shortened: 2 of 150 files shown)


awesomekling:
Oh, that's really clever! Using the MMU to implement `EFAULT` instead of validating ourselves.
This will definitely improve the performance of system calls :)

tomuta:
> Oh, that's really clever! Using the MMU to implement `EFAULT` instead of validating ourselves.
> This will definitely improve the performance of system calls :)

Well that plus doing it ourselves is extremely difficult and likely won't ever work right ;-) Especially once I get around to adding the APIC timer interrupts so we can actually schedule on multiple processors ;-)

awesomekling:
A fair point as well! The main downside to this approach is that it adds a little complexity for anyone wanting to do a port. But I think this is absolutely worth it.

tomuta:
> A fair point as well! The main downside to this approach is that it adds a little complexity for anyone wanting to do a port. But I think this is absolutely worth it.

I believe most systems, including Linux and Windows do something like this. So it should be portable in one way or another.

tomuta:
So I'm running into this weird paging issue that I also ran into when trying some test code in my purging improvements branch. I blamed it on my own changes, but this is happening here as well. For some bizarre reason after `fork` it fails to copy the page contents on a cow fault. I added some more logging and swapped it with the new `safe_memcpy` just for fun, but basically when getting the first page fault cow exception (flags 0x7 so protection violation + user + write) as expected, but then when copying the page it gets another page fault exception (flags 0x1 so protection violation + supervisor + read). But I looked at the PTE and it looks fine, the PDE also looks fine. And based on logs the TLB was also flushed. Any thoughts what else could trigger a protection violation?

```
[SystemServer(5:5)]: Region::clone(): CoWing Thread-specific (V0x01704000)
[SystemServer(5:5)]: MM: Region::map() will map VMO pages 0 - 0 (VMO page count: 1)
[SystemServer(5:5)]: quickmap_pd: Mapping P0x00a88000 at 0xffe04000 in pte @ 0xc0114020
[SystemServer(5:5)]: MM: Flush 1 pages at V0xffe04000 on CPU#0
[SystemServer(5:5)]: quickmap_pt: Mapping P0x00abc000 at 0xffe00000 in pte @ 0xc0114000
[SystemServer(5:5)]: MM: Flush 1 pages at V0xffe00000 on CPU#0
[SystemServer(5:5)]: MM: >> region map (PD=11038720, PTE=0x00abe065{0xffe00820}) Thread-specific V0x01704000 => P0x00abe000 (@0xc0229948)
[SystemServer(5:5)]: MM: Flush 1 pages at V0x01704000
[SystemServer(5:5)]: MM: Region::map() will map VMO pages 0 - 0 (VMO page count: 1)
[SystemServer(5:5)]: quickmap_pd: Mapping P0x00b2f000 at 0xffe04000 in pte @ 0xc0114020
[SystemServer(5:5)]: MM: Flush 1 pages at V0xffe04000 on CPU#0
[SystemServer(5:5)]: quickmap_pt: Mapping P0x00b45000 at 0xffe00000 in pte @ 0xc0114000
[SystemServer(5:5)]: MM: Flush 1 pages at V0xffe00000 on CPU#0
[SystemServer(5:5)]: MM: >> region map (PD=11722752, PTE=0x00abe005{0xffe00820}) Thread-specific V0x01704000 => P0x00abe000 (@0xc0229948)
[SystemServer(5:5)]: MM: Flush 1 pages at V0x01704000

...

[SystemServer(5:5)]: MM: CPU[0] handle_page_fault(0007) at V0x01704000
[SystemServer(5:5)]: PV(cow) fault in Region{0xc02caff8}[0]
[SystemServer(5:5)]:     >> It's a COW page and it's time to COW!
[SystemServer(5:5)]: MM: allocate_user_physical_page vending P0x00b46000
[SystemServer(5:5)]: quickmap_page: Mapping P0x00b46000 at 0xffe08000 in pte @ 0xc0114040
[SystemServer(5:5)]: MM: Flush 1 pages at V0xffe08000 on CPU#0
[SystemServer(5:5)]:       >> COW P0x00b46000 <- P0x00abe000
[SystemServer(5:5)]: MM: CPU[0] handle_page_fault(0001) at V0x01704000
[SystemServer(5:5)]: PV(error) fault in Region{0xc02caff8}[0] at V0x01704000
[SystemServer(5:5)]:       >> COW: error copying page P0x00abe000/V0x01704000 to P0x00b46000/V0xffe08000: failed to read from page at V0x01704000
```
Why would the kernel not be able to read that page based on that PTE?

tomuta:
Doh SMAP...

tomuta:
Ok now the problem is the `SC_` structures in `Syscall.h` are not trivial, so we can't simply `copy_from_user`, e.g. `Userspace<>` won't work. These really need to be trivial structures...

tomuta:
Hm I guess I could use something like `std::is_trivially_copyable` if we don't have that already in AK

tomuta:
Alright that solved the problems as well :)

bugaevc:
Would it be too much deviation from POSIX and other Unixes to just `SIGSEGV` the process instead of returning `EFAULT`?

awesomekling:
> Would it be too much deviation from POSIX and other Unixes to just `SIGSEGV` the process instead of returning `EFAULT`?

I don't know. None of our own software in-tree depends on `EFAULT` in any way. Perhaps signalling the process is enough. I'm not even sure if all Unixes have remotely the same semantics wrt this issue.

bugaevc:
> None of our own software in-tree depends on `EFAULT` in any way.

I'm 99% sure no sane software would ever intentionally pass possibly-invalid addresses to syscalls and depend on getting `EFAULT` to detect that the address has been, in fact, invalid. On the other hand, it is somewhat likely that some programs call a syscall with an invalid address, get `EFAULT`, forget to check the return code, and the bug goes unnoticed.

Besides,
> On GNU/Hurd systems, this error never happens; you get a signal instead.

awesomekling:
We still have to unwind the kernel stack though, so I'm not sure things would be much simpler without `EFAULT`. Or maybe I'm missing something?

bugaevc:
Right, I haven't thought of that.

On the other hand, if we validate addresses first, before taking any additional locks or references, we might be able to get away with just killing execution without unwinding. But probably not.

tomuta:
Yeah, we *always* have to unwind the kernel stack. Even if killing a sleeping/blocked process. So we'll need this error code regardless. To implement this behavior you'd have to probably check in `Syscall::handle` for that error code, and if that error code is returned then send an urgent signal to the calling process.

For portability reasons (mainly because I've seen this happen in real software that is able to continue just fine) my vote is to keep the behavior as-is (matching Linux). I however wouldn't oppose an opt-in mechanism allowing Serenity software to request a signal in this case.

tomuta:
> Right, I haven't thought of that.
> 
> On the other hand, if we validate addresses first, before taking any additional locks or references, we might be able to get away with just killing execution without unwinding. But probably not.

The problem with validating addresses is over-commitment. Sometimes valid memory regions may not have sufficient physical memory available to be able to do the COW upon access, which is why I need this for #3396. And, of course, short of locking the MM for the entire duration of a syscall (and re-validating upon each and every possible return from a block), it is a lot of unnecessary overhead that is prone to errors. The CPU validates all this for us anyway.

Kernel/UserOrKernelBuffer.cpp
```diff
@@ -0,0 +1,90 @@
...
+#include <Kernel/VM/MemoryManager.h>
+
+namespace Kernel {
+
+bool UserOrKernelBuffer::is_kernel_buffer() const
+{
+    return !is_user_address(VirtualAddress(m_buffer));
+}
+
+Optional<String> UserOrKernelBuffer::copy_into_string(size_t size) const
```
awesomekling, line 37: `String` already has a null state, so let's use that instead of wrapping it in `Optional`.

AK/String.h
```diff
@@ -88,6 +88,14 @@ class String {
...
     }
 
+    template<typename F>
+    String(size_t length, F f)
+    {
+        char* buffer;
+        m_impl = StringImpl::create_uninitialized(length, buffer);
+        f(buffer);
+    }
+
```
awesomekling, line 98: This does not feel like an idiomatic fit, why is this better than simply using `StringImpl::create_uninitialized()` at the call site?

awesomekling:
What is the purpose of shoveling things through an intermediate buffer when copying from kernel to userspace?

tomuta:
The purpose is to avoid having to do a bunch of e.g. 1-byte `copy_from/to_user` calls if the buffer is a user buffer (see console/serial stuff as an example). So it can a bulk copy to/from user memory using a buffer on the kernel stack. It's not needed in most cases, but mostly for those copy 1 byte or one structure at a time things. If the target is a kernel buffer, it'll call the closure with the kernel buffer directly, so it bypasses the intermediate buffer.

AK/String.h
tomuta: Ah I didn't realize that `StringImpl` was meant to be used outside of `String`... Thought it was internal to `String`.

Kernel/UserOrKernelBuffer.cpp
tomuta: So how would I create a non-null `String` that has a length of 0? I need to be able to distinguish between a fault (which can happen trying to read a 0-length string because the null-byte would need to be accessed) and a successful 0-length string. It looks like the `String()` would result in `is_null` returning `true`...

tomuta: Maybe `KResultOr<String>` instead?

awesomekling: `String::empty()` (or `String("")`) :)

tomuta: Ah :-) Can't believe I didn't think of `String("")`...

tomuta:
Added a comment explaining the purpose of the intermediate buffer

tomuta:
Hm I guess this needs a little more work, just noticed it breaks loading webpages...

tomuta:
Alright that was just a silly typo ;-)

PR #3468, opened 2020-09-12
Userland: Add {md5,sha1,sha256,sha512}sum

linusg:
~The program code is shared by using a huge macro that's just being passed the name of the hashing algorithm - hope that's a decent solution here?!~ (not anymore) I certainly don't want to maintain four separate copies of basically the same program.

Other than that I mimicked the behaviour of those programs on my Linux machine, for now they support multiple files, reading from stdin, file paths or both.

image

*Not sure what that setpgid message means...*

### DIFF
--- a/Userland/checksum.cpp
+++ b/Userland/checksum.cpp
@@ -1,0 +1,98 @@
+/*
+ * Copyright (c) 2020, Linus Groh <mail@linusgroh.de>
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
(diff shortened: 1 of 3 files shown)


alimpfard:
> Not sure what that setpgid message means...

Probably Shell bug lul

alimpfard:
You could also make them all the same binary that chooses between the calls by the name it was invoked as

linusg:
> You could also make them all the same binary that chooses between the calls by the name it was invoked as

yeah that was my first idea too, but then you have to have a common binary around that's all of these programs but also none of them at the same time and that's kind of awkward?!

asynts:
> You could also make them all the same binary that chooses between the calls by the name it was invoked as.

Serenity has `alias` right? Just set `alias sha256sum='checksum --sha256'`.

alimpfard:
> that's kind of awkward

busybox: \*cries in awkward\*

linusg:
aaaaand I forgot a `#pragma once`.

alimpfard:
> `alias sha256sum='checksum --sha256'`

That's a better solution, yeah!

awesomekling:
Aliases won't work nicely in some situations though, e.g if you're running an autoconf-generated configure script via bash inside Serenity.

alimpfard:
So...how about a `checksum` program, and individual `foosum` programs just `execv`'ing them with `--foo`?

nico:
I think "checksum" binary that changes behavior based on its name + symlinks is nice and reasonably idiomatic (clang/clang++/clang-cl do this too for example).

linusg:
Done.

Userland/checksum.cpp
```diff
@@ -0,0 +1,98 @@
...
+    Vector<const char*> paths;
+
+    Core::ArgsParser args_parser;
+    args_parser.add_positional_argument(paths, paths_help_string.characters(), "path", Core::ArgsParser::Required::No);
+    args_parser.parse(argc, argv);
+
+    if (paths.is_empty())
+        paths.append("-");
+
+    Crypto::Hash::Manager hash;
```
alimpfard, line 69: I see you found this (badly named...) guy.
I was gonna suggest it lol

awesomekling:
Very nice!

PR #3470, opened 2020-09-12
HackStudio: Add Git integration

itamar8910:
This adds the beginning of Git integration into HackStudio.

Current features are:
- Create a git repo for the project.
- View unstaged, modified and staged files.
- Stage/Unstage files.
- Make commits.
- View diffs in modified files before staging them.

The goal here is to make git integration feel intuitive and not abstract away the "flow" of actual git usage :smiley: 

The git integration features are optional and rely on the git port being installed, you can still use all other HackStudio features without having the git port.

Screenshot:
hs-git


DevTools/HackStudio/Git/GitFilesModel.cpp
```diff
@@ -0,0 +1,56 @@
...
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include "GitFilesModel.h"
+
+namespace HackStudio {
+
+RefPtr<GitFilesModel> GitFilesModel::create(Vector<LexicalPath>&& files)
```
awesomekling, line 31: Since this never returns null, it should return a `NonnullRefPtr<GitFilesModel>`

DevTools/HackStudio/Git/GitFilesView.cpp
```diff
@@ -0,0 +1,81 @@
...
+}
+
+void GitFilesView::mousedown_event(GUI::MouseEvent& event)
+{
+    if (event.button() != GUI::MouseButton::Left) {
+        ListView::mousedown_event(event);
+        return;
+    }
+
+    if (event.position().x() < action_icon_rect(0).x() || event.position().x() > action_icon_rect(0).top_right().x()) {
```
awesomekling, line 64: Note that `GUI::MouseEvent` has `x()` and `y()` getters :)

DevTools/HackStudio/Git/GitFilesView.h
```diff
@@ -0,0 +1,55 @@
...
+
+#pragma once
+
+#include <AK/LexicalPath.h>
+#include <LibGUI/ListView.h>
+#include <LibGfx/Bitmap.h>
+
+namespace HackStudio {
+
+typedef Function<void(const LexicalPath& file)> OnActionCallback;
```
awesomekling, line 35: This seems like an arbitrary thing to typedef.

DevTools/HackStudio/Git/GitFilesView.h
```diff
@@ -0,0 +1,55 @@
...
+
+class GitFilesView : public GUI::ListView {
+    C_OBJECT(GitFilesView)
+public:
+    virtual ~GitFilesView() override;
+
+protected:
+    virtual Gfx::Bitmap& action_bitmap() const = 0;
+    GitFilesView(OnActionCallback);
```
awesomekling, line 44: Please mark single-argument constructors `explicit` unless they are specifically intended for implicit conversion.

DevTools/HackStudio/Git/GitRepo.cpp
```diff
@@ -0,0 +1,196 @@
...
+
+    ASSERT(git_repo_exists(repository_root));
+    return adopt(*new GitRepo(repository_root));
+}
+
+Vector<LexicalPath> GitRepo::unstaged_files() const
+{
+    auto modified = modified_files();
+    auto untracked = untracked_files();
+    modified.append(untracked.data(), untracked.size());
```
awesomekling, line 64: I think you can do `modified.append(untracked);`
If not, we should make that work :)

DevTools/HackStudio/Git/GitRepo.cpp
```diff
@@ -0,0 +1,196 @@
...
+{
+    auto lines = raw_result.split('\n');
+    Vector<LexicalPath> files;
+    for (const auto& line : lines) {
+        files.empend(line);
+    }
+    return files;
+}
+
+Optional<String> GitRepo::command(const String& git_command) const
```
awesomekling: `String` already has a null state, so let's use that instead of wrapping it in `Optional` :)

DevTools/HackStudio/Git/GitRepo.cpp
```diff
@@ -0,0 +1,196 @@
...
+{
+    return command_wrapper(git_command, m_repository_root);
+}
+
+Optional<String> GitRepo::command_wrapper(const String& git_command, const LexicalPath& chdir)
```
awesomekling: Same comment about `String` null state.

Also, this function feels generally useful. Maybe something for LibCore? :)

DevTools/HackStudio/Git/GitRepo.cpp
```diff
@@ -0,0 +1,196 @@
...
+    };
+
+    if (WEXITSTATUS(wstatus) != 0) {
+        dbg() << "git command: " << git_command << " failed";
+        dbg() << "stderr: " << read_all_from_pipe(stderr_pipe);
+        return {};
+    }
+
+    auto result = read_all_from_pipe(stdout_pipe);
+    return result;
```
awesomekling, line 169: I don't think the temporary variable adds anything here.

DevTools/HackStudio/Git/GitRepo.cpp
```diff
@@ -0,0 +1,196 @@
...
+    close(stderr_pipe[1]);
+
+    auto read_all_from_pipe = [](int pipe[2]) {
+        auto result_file = Core::File::construct();
+        if (!result_file->open(pipe[0], Core::IODevice::ReadOnly, Core::File::ShouldCloseFileDescription::Yes)) {
+            perror("open");
+            ASSERT_NOT_REACHED();
+        }
+        auto data = result_file->read_all();
+        return String((const char*)data.data(), data.size());
```
awesomekling, line 160:
```suggestion
        return String::copy(result_file->read_all());
```

DevTools/HackStudio/main.cpp
```diff
@@ -711,7 +722,18 @@ static int main_impl(int argc, char** argv)
...
+        auto git_widget = s_action_tab_widget->active_widget();
+        if (!git_widget)
+            return;
+        if (strcmp(git_widget->class_name(), "GitWidget") != 0)
+            return;
+        reinterpret_cast<GitWidget*>(git_widget)->refresh();
+    };
+
+    g_open_file
+        = open_file;
```
awesomekling, line 753: Mysterious indentation here.

DevTools/HackStudio/main.cpp
```diff
@@ -711,7 +722,18 @@ static int main_impl(int argc, char** argv)
...
     };
 
-    g_open_file = open_file;
+    on_action_tab_change = [&]() {
+        update_actions();
+        auto git_widget = s_action_tab_widget->active_widget();
+        if (!git_widget)
+            return;
+        if (strcmp(git_widget->class_name(), "GitWidget") != 0)
+            return;
```
awesomekling, line 748: This looks a little awkward. How about `if (StringView(git_widget->class_name()) != "GitWidget")`
Hrm.. I guess that's also a little awkward :)

Libraries/LibGUI/ListView.h
```diff
@@ -62,9 +62,11 @@ class ListView : public AbstractView {
     virtual void move_cursor(CursorMovement, SelectionUpdate) override;
     void move_cursor_relative(int steps, SelectionUpdate);
 
-private:
+protected:
     ListView();
+    virtual void paint_list_item(Painter& painter, int row_index, int painted_item_index);
```
awesomekling, line 67: Per the coding style, please omit parameter names that are just the type name again (`Painter& painter` -> `Painter&` in this case)

awesomekling (changes requested):
Very interesting stuff! I feel like maybe we don't need separate classes for the staged/unstaged views, can they just share a class instead?

DevTools/HackStudio/main.cpp
alimpfard: I'd personally do it the other way around:
```c++
if (StringView { "GitWidget" } != git_widget->class_name())
```

alimpfard: Perhaps we should have a UDL for `StringView`'s?

DevTools/HackStudio/Git/DiffViewer.cpp
```diff
@@ -0,0 +1,253 @@
...
+
+    size_t num_lines = 0;
+    size_t current_original_line_index = 0;
+    for (const auto& hunk : m_hunks) {
+        num_lines += ((int)hunk.original_start_line - (int)current_original_line_index);
+
+        num_lines += hunk.removed_lines.size();
+        if (hunk.added_lines.size() - hunk.removed_lines.size()) {
+            num_lines += ((int)hunk.added_lines.size() - (int)hunk.removed_lines.size());
+        }
```
devsh0, line 238: Is this `if` necessary? `num_lines += (int)(hunk.added_lines.size() - hunk.removed_lines.size())`?

DevTools/HackStudio/main.cpp
itamar8910: > I'd personally do it the other way around:
> 
> ```c++
> if (StringView { "GitWidget" } != git_widget->class_name())
> ```

sure this reads nicely

DevTools/HackStudio/Git/DiffViewer.cpp
itamar8910: oops, means to do:
```c++
if (hunk.added_lines.size() > hunk.removed_lines.size())
```

DevTools/HackStudio/Git/GitRepo.cpp
itamar8910: Definitely. Added this to LibCore in 0fde8da

itamar8910:
@awesomekling @alimpfard @devsh0 thanks for the reviews!

PR #3473, opened 2020-09-13
Shell: Add functions (and $0/$1/...)

alimpfard:
An example overview:
```sh
fn(a b c) {
    echo $a $b $c \( $* \)
}
$ fn 1 2 3 4
# 1 2 3 ( 1 2 3 4 )
```

cc @bugaevc - feel free to nitpick, the syntax and semantics are relatively close to other shells :shrug:

### DIFF
--- a/Base/usr/share/man/man5/Shell.md
+++ b/Base/usr/share/man/man5/Shell.md
@@ -220,6 +220,37 @@ $ { for * { te $it } }&
 $ disown
 ```
 
+##### Functions
+A function is a user-defined entity that can be used as a simple command to execute a compound command, optionally with some parameters.
+Such a function is defined via the syntax below:
+
+```sh
+function_name(explicitly_named_arguments...) { compound_command }
+```
+
+The function is named `function_name`, and has some explicitly named arguments `explicitly_named_arguments...`, which *must* be supplied by the caller, failure to do so will cause the command to exit with status 1.
+
+The compound command shall be executed whenever the simple command `function_name` is executed.
+This execution shall be performed in a new local frame.
+
+Additionally, should the simple command containing the function name be in a pipeline, or requested to be run in the background, this execution shall be moved to a subshell; naturally, in such a case any changes to the shell state (such as variables, aliases, etc) shall not be leaked to the parent shell process.
+
+The passed arguments shall be stored in the special variables `*` and `ARGV`, and the explicitly named arguments shall be set, in order, from the first passed argument onwards.
+
+
+The exit status of a function simple command shall be the exit status of the last command executed within the command, or 0 if the function has no commands.
+The declaration is *not* a command, and will not alter the exit status.
+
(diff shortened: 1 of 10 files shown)


nico:
Thanks for fixing my FIXME!

alimpfard:
@nico Sorry I didn't cc you for that change, it totally slipped my mind...

Base/usr/share/man/man5/Shell.md
```diff
@@ -220,6 +220,35 @@ $ { for * { te $it } }&
...
+Such a function is defined via the syntax below:
+
+```sh
+function_name(explicitly_named_arguments...) { compound_command }
+```
+
+The function is named `function_name`, and has some explicitly named arguments `explicitly_named_arguments...`, which *must* be supplied by the caller, failure to do so will cause the command to exit with status 1.
+
+The compound command shall be executed whenever the simple command `function_name` is executed.
+This execution shall be performed in a subshell, in a new local frame.
```
bugaevc, line 234: Wait, why in a subshell?

alimpfard: is it supposed to be run in the current shell process itself?
how would job control work with that?

bugaevc: Uh, I've always thought shells run functions in-process, just like other control flow, or like other languages

Not sure about job control; do you allow job control on builtins (e.g. `cd foo/ &`?)

alimpfard: > Job control on builtins

Hmmm...I don't think so, no.

alimpfard: Is there any good reason to make functions run in-process? the only thing I can think of is to make modifying the shell state possible...
But always running it in a subshell has a nice uniformity; piping two functions together wouldn't be possible if we make them run in-process

bugaevc: If the user tries to use I/O redirections (pipes) or job control, *then* run it in a subshell. Or even require the user to do it explicitly: `{ fn 1 2 } | head`, but I'm not sure that's a good idea.

What do other shells do?

alimpfard: It seems zsh does as you're suggesting, if it's run as part of a pipeline or something that would require job control, it's run in a subshell.
otherwise it's run in the current process.

alimpfard: An alternative way would be to simply make a syntactical distinction between the two, one would be always run in the current process, and the other always in a subprocess

alimpfard: > If the user tries to use I/O redirections (pipes) or job control, then run it in a subshell.

Ah heck it, I'll just do this. seems to be the most sane option (according to other shells) of the bunch

PR #3474, opened 2020-09-13
AK: Lower the requirements for InputStream::eof and rename it.

asynts:
Consider the following snippet:

~~~c++
void foo(InputStream& stream) {
    if(!stream.eof()) {
        u8 byte;
        stream >> byte;
    }
}
~~~

There is a very subtle bug in this snippet, for some input streams `stream.eof()` might return false even if no more data can be read. In this case an error flag would be set on the stream.

Until now I've always ensured that this is not the case, but this made the implementation unnecessarily complicated. `InputFileStream::eof` had to keep a `ByteBuffer` around just to make this possible. That meant a ton of unnecessary copies just to get a reliable end-of-file check.

In most cases it isn't actually necessary to have a reliable check for end-of-file.

In most other cases a reliable end-of-file check is available anyways because in some cases like `InputMemoryStream` it is very easy to implement.

---

I did a little benchmark with the Gzip implementation and found that decompressing a ~47MiB file took 102753ms with these changes compared to 128411ms before. I only did a single benchmark so this is might not be very representative but that's about a 20% improvement.

### DIFF
--- a/AK/BitStream.h
+++ b/AK/BitStream.h
@@ -66,7 +66,7 @@ public:
         return true;
     }
 
-    bool eof() const override { return !m_next_byte.has_value() && m_stream.eof(); }
+    bool unreliable_eof() const override { return !m_next_byte.has_value() && m_stream.unreliable_eof(); }
 
     bool discard_or_error(size_t count) override
     {
@@ -86,6 +86,11 @@ public:
 
         size_t nread = 0;
         while (nread < count) {
+            if (m_stream.has_any_error()) {
+                set_fatal_error();
+                return 0;
+            }
+
             if (m_next_byte.has_value()) {
                 const auto bit = (m_next_byte.value() >> m_bit_offset) & 1;
                 result |= bit << nread;
@@ -93,9 +98,6 @@ public:
 
                 if (m_bit_offset++ == 7)
                     m_next_byte.clear();
-            } else if (m_stream.eof()) {
-                set_fatal_error();
-                return 0;
             } else {
                 m_stream >> m_next_byte;
                 m_bit_offset = 0;
--- a/AK/Buffered.h
+++ b/AK/Buffered.h
@@ -77,7 +77,7 @@ public:
         return nread;
     }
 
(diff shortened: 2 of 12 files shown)


awesomekling:
Surely we can find a better name than `unreliable_eof()`, that just makes it sound like you should never call it :)

asynts:
A couple of ideas, but I think they are all worse:

  - `lazy_eof()` because it doesn't read additional input.
  - `trivial_eof()` because the implementation is trivial.
  - `sufficient_eof_check()` because `sufficient_eof_check() = true` is sufficient to conclude that no more data can be read.

awesomekling:
Hmm, so the semantics here are basically "if this thing returns false, do another read. if that read fails, you're actually EOF"?

asynts:
Exactly, `read(Bytes)` reads at least one byte unless zero bytes are requested or none are available. `unreliable_eof()` returns true if end-of-file can be detected somehow or if the nothing was read in the last call.

asynts:
I've added comments to make the semantics clear:

~~~diff
--- a/AK/Stream.h
+++ b/AK/Stream.h
@@ -75,10 +75,15 @@
 
 class InputStream : public virtual AK::Detail::Stream {
 public:
-    // Does nothing and returns zero if there is already an error.
+    // Reads at least one byte unless none are requested or none are avaliable. Does nothing
+    // and returns zero if there is already an error.
     virtual size_t read(Bytes) = 0;
-    virtual bool read_or_error(Bytes) = 0;
+
+    // If this function returns true, then no more data can be read. If read(Bytes) previously
+    // returned zero even though bytes were requested, then the inverse is true as well.
     virtual bool unreliable_eof() const = 0;
+
+    virtual bool read_or_error(Bytes) = 0;
     virtual bool discard_or_error(size_t count) = 0;
 };
 
 ~~~

awesomekling:
Blergh all right. I dislike the name but okay for now :)

asynts:
If I come up with something better, I'll just refactor-rename and create a pull request.

PR #3476, opened 2020-09-13
Base: Make I-Beam cursor 4px smaller

### LINKED ISSUE #3475: Default I-Beam cursor a little big?
Is it just me or does the I-Beam cursor feel overly big compared to all the other cursors?

linusg:
That might seem like a lot for a cursor image, but I think it looks pretty decent.

Before / after / regular cursor for comparison:

image image image

Fixes #3475.


awesomekling:
Oh yeah, that's a lot nicer!

PR #3479, opened 2020-09-13
LibGUI: ComboBox fix

### LINKED ISSUE #3431: ComboBox: Value is changed when moving mouse over entries
A recent change changed `ComboBox` to highlight the entry the mouse is moving over. But I don't think that should actually change the selected value. At least that is not a behavior I'm accustomed to. E.g. in Display Properties that also causes quite some CPU cycles because it has to load the images. The bigger problem I see is that if you dismiss the ComboBox you're under the impression that the value that is now the selected value was the original value before dismissing it.

pkotzbach:
Fixes #3431

### DIFF
--- a/Libraries/LibGUI/ComboBox.cpp
+++ b/Libraries/LibGUI/ComboBox.cpp
@@ -60,10 +60,6 @@ ComboBox::ComboBox()
 {
     m_editor = add<ComboBoxEditor>();
     m_editor->set_has_open_button(true);
-    m_editor->on_change = [this] {
-        if (on_change)
-            on_change(m_editor->text(), m_list_view->cursor_index());
-    };
     m_editor->on_return_pressed = [this] {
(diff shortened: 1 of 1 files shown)


Lubrsi:
If you put "Fixes" followed by the issue number, e.g. "Fixes #3479", GitHub will automagically close the issue :^)

Lubrsi:
Sorry, I forgot to clarify that it has to be in the body, not the title.

pkotzbach:
Thank you, I'm starting to getting used to github :)

awesomekling:
Please observe the [contribution guidelines](https://github.com/SerenityOS/serenity/blob/master/CONTRIBUTING.md)

Specifically:
>**Don't**
>Use weasel-words like "refactor" or "fix" to avoid explaining what's being changed.

We need to have some kind of description of what's being fixed and why. :)

PR #3483, opened 2020-09-13
Switch to Ubuntu Focal Fossa

BenWiederhake:
Travis' support for Ubuntu Focal is getting along quite nicely, we should finally switch to it.

This PR also:
- Gets rid of the PPA and extra caching necessary for a less-than-ancient gcc.
- Gets rid of the manual download and installation of cmake.
- Makes the Toolchain just a little bit smaller, because we can.

Also, @alimpfard suggested that we could automatically run `clang-format`, which should become available more easily on Ubuntu Focal than on Bionic. :)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 os: linux
-dist: bionic
+dist: focal
 
 language: cpp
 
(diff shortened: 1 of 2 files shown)


alimpfard:
image

PR #3485, opened 2020-09-14
LibLine: Support alt-arrow left/right, and ctrl-delete

nico:
Requires some gymnastics to better send modifier keys with special keys. Spruces up LibLine's CSI parser to handle CSI parameters and intermediates, changes LibVT to send modifiers as CSI parameter like xterm does, and handles modifiers for some keys in LibLine.

(It's a bit unfortunate that LibLine and Terminal.cpp both contain lots of ANSI escape code machinery. Maybe we can pull that out to some shared place at some point in the future.)

### DIFF
--- a/Libraries/LibLine/Editor.cpp
+++ b/Libraries/LibLine/Editor.cpp
@@ -736,7 +736,10 @@ void Editor::handle_read_event()
                 continue;
             case '~':
                 if (param1 == 3) { // ^[[3~: delete
-                    erase_character_forwards();
+                    if (modifiers == CSIMod::Ctrl)
+                        erase_alnum_word_forwards();
+                    else
+                        erase_character_forwards();
                     m_search_offset = 0;
                     continue;
                 }
--- a/Libraries/LibLine/Editor.cpp
+++ b/Libraries/LibLine/Editor.cpp
@@ -717,13 +717,13 @@ void Editor::handle_read_event()
                 search_forwards();
                 continue;
             case 'D': // ^[[D: arrow left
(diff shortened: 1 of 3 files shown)


Libraries/LibLine/Editor.cpp
```diff
@@ -616,7 +616,6 @@ void Editor::handle_read_event()
     }
 
     auto reverse_tab = false;
-    auto ctrl_held = false;
```
alimpfard, line 619: I can immediately see that I like this change!

Libraries/LibLine/Editor.cpp
```diff
@@ -635,6 +634,11 @@ void Editor::handle_read_event()
     Vector<unsigned, 4> csi_parameters;
     Vector<u8> csi_intermediate_bytes;
     u8 csi_final;
+    enum CSIMod {
+        Shift = 1,
+        Alt = 2,
+        Ctrl = 4,
+    };
```
alimpfard, line 641: maybe make this an `enum class`? you seem to be using it qualified anyway.

Libraries/LibLine/Editor.cpp
nico: Then I need to explicitly cast it to int.

alimpfard: wrong comment to reply, but fair enough.

Libraries/LibLine/Editor.cpp
nico: Then I need to explicitly cast it to int.

PR #3493, opened 2020-09-14
Kernel: Handle safe_memcpy/safe_memset/safe_strnlen faults in irq han…

### LINKED ISSUE #3492: Kernel: Profiling is broken
To reproduce: try to profile something

```sh
profile -c /bin/id
```

Lots of this error:
```
[profile(28:28)]: CPU[0] BUG! Page fault while handling IRQ! code=0, vaddr=V0x2b536bce, irq level: 1
```

cc @tomuta

tomuta:
…dlers

Fix gracefully failing these calls if used within IRQ handlers. If we're
handling IRQs, we need to handle these failures first, because we can't
really resolve page faults in a meaningful way. But if we know that it
was one of these functions that failed, then we can gracefully handle
the situation.

This solves a crash where the Scheduler attempts to produce backtraces
in the timer irq, some of which cause faults.

Fixes #3492

### DIFF
--- a/Kernel/Arch/i386/CPU.cpp
+++ b/Kernel/Arch/i386/CPU.cpp
@@ -401,9 +401,23 @@ void page_fault_handler(TrapFrame* trap)
     dump(regs);
 #endif
 
-    bool faulted_in_userspace = (regs.cs & 3) == 3;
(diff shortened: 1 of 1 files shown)


awesomekling:
Cool :)

PR #3499, opened 2020-09-15
LibGUI: Increase slider acceleration with control

networkException:
When holding control and scrolling on a slider widget
the scrolling acceleration gets increased.

This can make it faster to get to the knob location
you want to get to :^)

---

I am not too sure about what modifier to use, control
felt natural though

### DIFF
--- a/Libraries/LibGUI/Slider.cpp
+++ b/Libraries/LibGUI/Slider.cpp
@@ -161,7 +161,7 @@ void Slider::mousewheel_event(MouseEvent& event)
 {
     auto acceleration_modifier = m_step;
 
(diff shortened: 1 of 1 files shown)


Libraries/LibGUI/Slider.cpp
```diff
@@ -159,10 +159,15 @@ void Slider::mouseup_event(MouseEvent& event)
 
 void Slider::mousewheel_event(MouseEvent& event)
 {
+    auto acceleration_modifier = m_step;
+
+    if(event.modifiers() == KeyModifier::Mod_Ctrl && knob_size_mode() == KnobSizeMode::Fixed)
```
awesomekling, line 164: Please use `clang-format`

networkException: Oops I should maybe set my IDE up to format on save

PR #3500, opened 2020-09-15
UserspaceEmulator: Intercept sendto(), fix recvfrom()

nico:
With this, `ue /bin/ntpquery` successfully does network i/o and prints the result (but later hits an unimplemented FILD_RM64).


awesomekling:
Awesome fixes :D

PR #3505, opened 2020-09-16
Shell: Formatting and misc fixes

alimpfard:
This PR adds...well, formatting.
The formatted code is not the nicest thing in the world, but hopefully we can build up on this and define a decent shell script format.

as an aside, it would be pretty neat to have formatting in `js` too \*hint hint\*

cc @nico - you hinted at this, so you get more than you asked for.

As mentioned above, the format isn't very nice, here's a sample:
<details>
<summary>Expand for sample</summary>

Input:

```sh
#!/bin/sh

thoughts=( "Formatting is odd" "This needs more work"\
    "Help! I'm drowning!")

IFS="\n"
for $(cat stuff) {
  clear && echo "AAAAAAAAAAAAAAAAAAAAAAAAAAAAAA" && echo $it


  # Comment here
  sleep 4 || exit 3
}

clear
echo "That's all!"
```

Formatted:
```sh
#!/bin/sh
thoughts=("Formatting is odd" "This needs more work" "Help! I'm drowning!")
IFS="\n"
for $(cat stuff) {
  clear \
    && echo "AAAAAAAAAAAAAAAAAAAAAAAAAAAAAA" \
    && echo $it
  # Comment here
  sleep 4 \
    || exit 3
}
clear
echo "That's all!"
```
</details>


Libraries/LibLine/Editor.h
```diff
@@ -204,6 +204,12 @@ class Editor : public Core::Object {
     }
 
     size_t cursor() const { return m_cursor; }
+    void set_cursor(size_t cursor)
+    {
+        if (cursor >= m_buffer.size())
+            cursor = m_buffer.size();
```
devsh0, line 210:
```suggestion
        if (cursor > m_buffer.size())
            cursor = m_buffer.size();
```

Shell/Formatter.cpp
```diff
@@ -0,0 +1,648 @@
...
+void Formatter::insert_separator()
+{
+    current_builder().append('\n');
+    insert_indent();
+}
+
+void Formatter::insert_indent()
+{
+    for (size_t i = 0; i < m_current_indent; ++i)
+        current_builder().append("  ");
```
devsh0, line 88: I believe there are two spaces there. Do you mean it?

Shell/Formatter.cpp
```diff
@@ -0,0 +1,648 @@
...
+
+    current_builder().append(") {");
+
+    {
+        TemporaryChange indent { m_current_indent, m_current_indent + 1 };
+        insert_separator();
+
+        if (node->block())
+            node->block()->visit(*this);
+    }
```
devsh0, line 244: Do we need this block?

Shell/Formatter.cpp
```diff
@@ -0,0 +1,648 @@
...
+        node->condition()->visit(*this);
+
+    current_builder().append(" {");
+    {
+        TemporaryChange indent { m_current_indent, m_current_indent + 1 };
+
+        insert_separator();
+        if (node->true_branch())
+            node->true_branch()->visit(*this);
+    }
```
devsh0, line 323: Same.

Shell/Formatter.cpp
```diff
@@ -0,0 +1,648 @@
...
+
+    if (node->false_branch()) {
+        if (node->false_branch()->class_name() != "IfCond") {
+            current_builder().append(" else {");
+            {
+                TemporaryChange indent { m_current_indent, m_current_indent + 1 };
+
+                insert_separator();
+                node->false_branch()->visit(*this);
+            }
```
devsh0, line 336: Since it's been repeated multiple times, I believe adding the block was intentional. I'd personally not introduce unnecessary blocks if I don't have to but feel free to ignore this comment if you like it that way.

Shell/Formatter.cpp
```diff
@@ -0,0 +1,648 @@
...
+
+            current_builder().append(" {");
+            {
+                TemporaryChange indent { m_current_indent, m_current_indent + 1 };
+
+                insert_separator();
+                if (entry.body)
+                    entry.body->visit(*this);
+            }
```
devsh0, line 408: In case you do consider removing the block, here's another.

Shell/Formatter.cpp
alimpfard: Yes, I decided to initially indent with two spaces, later we can make that an option.

Shell/Formatter.cpp
alimpfard: We do, `TemporaryChange` reverts the indentation when this block ends.

alimpfard: Thanks for raising this issue.
I abstracted this stuff away in `in_new_block()` and `with_added_indent()`.

Shell/AST.cpp
```diff
@@ -190,12 +190,17 @@ Node::~Node()
...
+    if (m_left)
+        m_left->dump(level + 1);
+    if (m_right)
+        m_right->dump(level + 1);
 }
 
 RefPtr<Value> And::run(RefPtr<Shell> shell)
 {
+    ASSERT(m_left);
+    ASSERT(m_right);
```
awesomekling, line 202: Instead of this, `m_left` and `m_right` should be `NonnullRefPtr`s :)

alimpfard: Previously, this made sense, as the parser wouldn't construct half-correct nodes
but now that the actual structure is needed (and examined) via the formatter, these may well be null.

I don't think `NonnullRefPtr` would work in general.

Though in the case of `And`, I believe the parser does substitute a `SyntaxError` node if the rhs is null, so it applies in this specific case :thinking:

alimpfard: I'll try making them all `NonnullRefPtr`'s and putting `SyntaxError` nodes inside them, let's be a bit more hygienic~

awesomekling: Grand!

alimpfard: That was a workout and a half :P

Shell/AST.cpp
```diff
@@ -924,13 +930,18 @@ HitTestResult ForLoop::hit_test_position(size_t offset)
...
 
-    if (auto result = m_iterated_expression->hit_test_position(offset); result.matching_node)
-        return result;
+    if (m_iterated_expression) {
+        if (auto result = m_iterated_expression->hit_test_position(offset); result.matching_node)
+            return result;
+    }
+
+    if (!m_block)
+        return {};
```
alimpfard, line 939: what the heck, this looks wrong

awesomekling:
>live formatting

Well that sounds terrible :D

This would be cool to have for all languages, for sure. Anyways, there's a conflict here, otherwise okay I think.

alimpfard:
> Well that sounds terrible :D

Not only does it _sound_ terrible, it _is_ terrible too!

Will fixup the conflict in a bit!

PR #3509, opened 2020-09-16
 LibIPC: Disable Notifier before closing socket 

### LINKED ISSUE #3508: LibIPC: Connection::did_misbehave closes socket, breaking EventLoop
It seems like it should somehow get rid of the `Notify` objects that are registered with the `EventLoop` before closing the socket, rather than triggering an `ASSERT` leading to a crash.

This stack trace from #3494 shows that WindowServer crashes entirely only because TaskBar triggered a call to `did_misbehave`, causing this bug in the `EventLoop`.

```
WindowServer(14): ClientConnection{0x028306fc} (id=2, pid=21) misbehaved (WM_SetWindowTaskbarRect: Bad window ID), disconnecting.
[WindowServer(14:14)]: sys$select: Bad fd number 11
USERSPACE(14) ASSERTION FAILED: false
../Libraries/LibCore/EventLoop.cpp:608
[WindowServer(14:14)]: Terminating WindowServer(14) due to signal 6
USERSPACE(21) ASSERTION FAILED: response
../Libraries/LibIPC/Connection.h:109
```

tomuta:
Because we're closing a file descriptor, we need to disable any
Notifier that is using it so that the EventLoop does not use invalid
file descriptors.

Fixes #3508

### DIFF
--- a/Libraries/LibCore/Notifier.cpp
+++ b/Libraries/LibCore/Notifier.cpp
@@ -46,10 +46,20 @@ Notifier::~Notifier()
 
 void Notifier::set_enabled(bool enabled)
 {
+    if (m_fd < 0)
+        return;
     if (enabled)
         Core::EventLoop::register_notifier({}, *this);
     else
         Core::EventLoop::unregister_notifier({}, *this);
+}
+
+void Notifier::close()
+{
+    if (m_fd < 0)
+        return;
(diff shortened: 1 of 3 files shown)


tomuta:
This solves the `ASSERT` being triggered in WindowServer when a client misbehaves.

PR #3510, opened 2020-09-16
Implement most of recvmsg(), sendmsg(), add SO_TIMESTAMP, use it in ntpquery

nico:
I went with SO_TIMESTAMP instead of SIOCGSTAMP since it seems more common (the BSDs have SO_TIMESTAMP; SIOCGSTAMP looks like it's maybe linux-only), and if a socket is used for more than one recv() it needs fewer syscalls too.

sendmsg() and recvmsg() aren't the most easy-on-the-eyes APIs, but they're POSIX, so it seems good to grow support for them. I didn't implement full iovec support since I don't need that part at the moment, and the existing iovec stuff in the kernel is a bit bolted on. Maybe I'll clean it up and add readv in the future. That's also why the Socket-level in-kernel api is still recvfrom() / sendto().

The in-kernel network adapter now records timestamps for all incoming packets. I believe that matches Linux, and it's arguably useful for adding in-kernel latency metrics later on.

--

I'm still somewhat new to all the copy_to_user() / copy_from_user() stuff both in kernel and in UserspaceEmulator, so it's probably worth having a closer look at those parts. Everything seems to work, but it's possible I'm doing overly roundabout things there.

This ended up being quite a lot of code for what feels like not much feature, but I suppose that's the nature of network code.

### DIFF
--- a/Libraries/LibC/sys/socket.cpp
+++ b/Libraries/LibC/sys/socket.cpp
@@ -28,7 +28,9 @@
 #include <Kernel/API/Syscall.h>
 #include <errno.h>
 #include <stdio.h>
+#include <string.h>
 #include <sys/socket.h>
+#include <sys/uio.h>
 
 extern "C" {
 
@@ -68,11 +70,17 @@ int shutdown(int sockfd, int how)
     __RETURN_WITH_ERRNO(rc, rc, -1);
 }
 
+ssize_t sendmsg(int sockfd, const struct msghdr* msg, int flags)
+{
+    int rc = syscall(SC_sendmsg, sockfd, msg, flags);
+    __RETURN_WITH_ERRNO(rc, rc, -1);
+}
+
 ssize_t sendto(int sockfd, const void* data, size_t data_length, int flags, const struct sockaddr* addr, socklen_t addr_length)
 {
-    Syscall::SC_sendto_params params { sockfd, { data, data_length }, flags, addr, addr_length };
-    int rc = syscall(SC_sendto, &params);
-    __RETURN_WITH_ERRNO(rc, rc, -1);
+    iovec iov = { const_cast<void*>(data), data_length };
+    msghdr msg = { const_cast<struct sockaddr*>(addr), addr_length, &iov, 1, nullptr, 0, 0 };
+    return sendmsg(sockfd, &msg, flags);
 }
 
 ssize_t send(int sockfd, const void* data, size_t data_length, int flags)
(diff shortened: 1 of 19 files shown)


awesomekling:
cc @bugaevc @bgianfo

Libraries/LibC/sys/socket.cpp
```diff
@@ -68,25 +70,48 @@ int shutdown(int sockfd, int how)
...
+{
+    if (!addr_length && addr) {
+        errno = EINVAL;
+        return -1;
+    }
+
+    sockaddr_storage internal_addr;
+    iovec iov = { buffer, buffer_length };
+    msghdr msg = { addr ? &internal_addr : nullptr, addr ? sizeof(internal_addr) : 0, &iov, 1, nullptr, 0, 0 };
+    int rc = recvmsg(sockfd, &msg, flags);
```
bugaevc, line 107:
```suggestion
    ssize_t rc = recvmsg(sockfd, &msg, flags);
```

Libraries/LibC/sys/socket.h
```diff
@@ -102,8 +113,10 @@ int accept(int sockfd, struct sockaddr*, socklen_t*);
 int connect(int sockfd, const struct sockaddr*, socklen_t);
 int shutdown(int sockfd, int how);
 ssize_t send(int sockfd, const void*, size_t, int flags);
+ssize_t sendmsg(int socket, const struct msghdr*, int flags);
```
bugaevc, line 140:
```suggestion
ssize_t sendmsg(int sockfd, const struct msghdr*, int flags);
```

And likewise below

Libraries/LibC/sys/socket.h
```diff
@@ -135,4 +149,30 @@ int getpeername(int sockfd, struct sockaddr*, socklen_t*);
...
 
+// These three are non-POSIX, but common:
+#define CMSG_ALIGN(x) (((x) + sizeof(void*) - 1) & ~(sizeof(void*) - 1))
+#define CMSG_SPACE(x) (CMSG_ALIGN(sizeof(struct cmsghdr)) + CMSG_ALIGN(x))
+#define CMSG_LEN(x) (CMSG_ALIGN(sizeof(struct cmsghdr)) + (x))
+
+static inline struct cmsghdr* CMSG_FIRSTHDR(const struct msghdr* msg)
+{
+    if (msg->msg_controllen < sizeof(struct cmsghdr))
+        return 0;
```
bugaevc, line 160:
```suggestion
        return NULL;
```
perhaps?

Also are you sure it should be casting a `const` pointer to non-const pointer?

nico: Re const cast: I suppose it depends on if you picture the msghdr as a container of cmsghdrs (in which case you shouldn't hand out a mutable pointer since you can mutate part of the msghdr object through it), or as an object that's independent of the cmsghdrs and just points to them.

Semantically, it definitely would make sense to me if CMSG_NXTHDR took a const pointer in the 2nd argument and returned a mutable pointer -- the next cmsg isn't related to the first. But making that work would require const_cast<>s which we can't use in LibC headers.

I guess I'll just make all the args non-const, that's self-consistent and consistent with what most other unixes seem to do.

PR #3511, opened 2020-09-16
Allow executing binaries from PATH with env.

asynts:
Adds two new features to `/bin/env`:

 1. `env FOO=bar` sets the environment variable `FOO` to `bar` before doing anything else.

 2. `env foo` looks up `foo` in `PATH` and then executes it.

This is useful for shebangs:

~~~none
#!/bin/env Shell
echo "Hello, World"
~~~

(The following would not work because the interpreter is not looked up on the path:)

~~~none
#!Shell
echo "Hello, World"
~~~

### DIFF
--- a/DevTools/UserspaceEmulator/main.cpp
+++ b/DevTools/UserspaceEmulator/main.cpp
@@ -30,7 +30,7 @@
 #include <AK/LogStream.h>
 #include <AK/MappedFile.h>
 #include <AK/StringBuilder.h>
-#include <LibCore/ArgsParser.h>
+#include <LibCore/DirIterator.h>
 #include <LibELF/Loader.h>
 #include <getopt.h>
 #include <pthread.h>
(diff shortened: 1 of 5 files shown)


DevTools/UserspaceEmulator/main.cpp
```diff
@@ -39,12 +39,11 @@
...
 {
     if (argc == 1) {
-        out() << "usage: UserspaceEmulator <command>";
+        out() << "Usage: UserspaceEmulator <command>";
         return 0;
     }
 
-    // FIXME: Allow specifying any command in $PATH instead of requiring a full executable path.
-    const char* executable_path = argv[1];
+    auto executable_path = Core::find_executable_in_path(argv[1]);
```
nico, line 46: This seems to not work:

```
$ ue ls
open: Bad address
Unable to map
```

(But `ue /bin/ls` works fine.)

asynts: This is weird, I could swear that I tried to run `UserspaceEmulator id` and it worked. But somehow it's broken.

asynts: #3514

PR #3514, opened 2020-09-16
LibCore: Add missing .characters() for String::format.

asynts:
I messed this up while rebasing #3511.

This caused `UserspaceEmulator id` to fail because the path was formatted as `(null)//bin:/bin/usr:/usr/local/bin`.

### DIFF
--- a/Libraries/LibCore/DirIterator.cpp
+++ b/Libraries/LibCore/DirIterator.cpp
@@ -24,8 +24,8 @@
  * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
-#include <LibCore/DirIterator.h>
 #include <AK/Vector.h>
+#include <LibCore/DirIterator.h>
 #include <errno.h>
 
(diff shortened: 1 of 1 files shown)


awesomekling:
Oh duh :]

Libraries/LibCore/DirIterator.cpp
```diff
@@ -108,8 +108,8 @@ String find_executable_in_path(String filename)
         return {};
     }
 
-    for (auto directory : StringView { getenv("PATH") }.split_view(':')) {
-        auto fullpath = String::format("%s/%s", directory, filename);
+    for (auto directory : String { getenv("PATH") }.split(':')) {
+        auto fullpath = String::format("%s/%s", directory.characters(), filename.characters());
```
nico, line 112: FYI you can also format() a StringView by doing `"%.*s", directory.length(), directory.characters_without_null_termination()`

asynts: I actually knew that, but it's soo cumbersome.

PR #3518, opened 2020-09-17
LibJS: Optional chaining

linusg:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

I hope I didn't miss any edge cases in the parser :)

### DIFF
--- a/Libraries/LibJS/Parser.cpp
+++ b/Libraries/LibJS/Parser.cpp
@@ -653,6 +653,8 @@ NonnullRefPtr<Expression> Parser::parse_unary_prefixed_expression()
         auto rhs_start_line = m_parser_state.m_current_token.line_number();
         auto rhs_start_column = m_parser_state.m_current_token.line_column();
         auto rhs = parse_expression(precedence, associativity);
+        // FIXME: Apparently for functions this should also not be enforced on a parser level,
+        // other engines throw ReferenceError for ++foo()
         if (!rhs->is_identifier() && !rhs->is_member_expression())
             syntax_error(String::format("Right-hand side of prefix increment operator must be identifier or member expression, got %s", rhs->class_name()), rhs_start_line, rhs_start_column);
         return create_ast_node<UpdateExpression>(UpdateOp::Increment, move(rhs), true);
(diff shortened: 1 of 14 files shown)


Libraries/LibJS/Parser.cpp
```diff
@@ -1084,23 +1093,53 @@ NonnullRefPtr<Expression> Parser::parse_secondary_expression(NonnullRefPtr<Expre
...
+    case TokenType::Period: {
         consume();
         if (!match_identifier_name())
             expected("IdentifierName");
-        return create_ast_node<MemberExpression>(move(lhs), create_ast_node<Identifier>(consume().value()));
+        auto rhs = create_ast_node<Identifier>(consume().value());
+        if (lhs->is_optional_member_expression())
+            return create_ast_node<OptionalMemberExpression>(move(lhs), move(rhs));
+        else
+            return create_ast_node<MemberExpression>(move(lhs), move(rhs));
```
linusg, line 1104: I know what you're going to say, no need for else here... but it looks a lot nicer this way, imo.

```cpp
if (lhs->is_optional_member_expression())
    return create_ast_node<OptionalMemberExpression>(move(lhs), move(rhs));
return create_ast_node<MemberExpression>(move(lhs), move(rhs));
```
:unamused:

awesomekling: I agree that sometimes it does look nicer your way when choosing between two paths of equal semantic importance. :)

linusg:
Actually I think I forgot something...

Libraries/LibJS/Parser.cpp
```diff
@@ -1067,7 +1076,7 @@ NonnullRefPtr<Expression> Parser::parse_secondary_expression(NonnullRefPtr<Expre
         consume();
         return create_ast_node<AssignmentExpression>(AssignmentOp::UnsignedRightShiftAssignment, move(lhs), parse_expression(min_precedence, associativity));
     case TokenType::ParenOpen:
-        return parse_call_expression(move(lhs));
+        return parse_call_expression(move(lhs), lhs->is_optional_member_expression());
```
devsh0, line 1079: Is it okay to use `lhs` after moving?

Libraries/LibJS/Tests/builtins/Array/Array.from.js
```diff
@@ -50,8 +50,9 @@ describe("normal behavior", () => {
                     let value = begin - 1;
                     return {
                         next() {
-                            if (value < end)
+                            if (value < end) {
                                 value += 1;
+                            }
```
devsh0, line 55: Or if you prefer a more linear-looking flow, `value += value < end` or `value += value < end ? 1 : 0`. Feel free to ignore this comment if you like it the way it is.

nico:
Out of interest, what was wrong with this? :)

linusg:
@nico I had everything working but this simple test:

```js
const a = { b: null };
expect(() => { a?.b.c; }).toThrow(TypeError);
expect(() => { a?.b(); }).toThrow(TypeError);
```

(https://github.com/SerenityOS/serenity/pull/3518/files#diff-54894fc2c222055318c1fbadd78ba33eR72-R76)

So I guess I implemented something fundamentally wrong (parser is working fine though), but I didn't have the mental energy or time to revise either.

PR #3523, opened 2020-09-17
Kernel: Simplify dropping privileges with pledge.

asynts:
Dropping privileges with pledge is painful, you always need to know exactly which privileges you have currently. This is especially bad if privileges are dropped conditionally.

This commit adds the possibility to drop privileges by adding a placeholder "all" promise that expands to all privileges that have not been dropped yet, an example:

~~~c++
pledge("stdio rpath");
int fd = open("example.txt", O_RDONLY);
pledge("all -rpath");

// Using an already opened file descriptor only needs "stdio".
~~~

(Obviously, we should check the return values.) "all" means in this context "stdio rpath" and "-rpath" then only leaves "stdio".

I've tried to make it biased towards dropping permissions, meaning that

 1. `pledge("-stdio stdio", nullptr)` drops stdio.

 2. `pledge("-stdio all", nullptr)` returns `-EPERM` if `stdio` was already dropped.

 3. `pledge("-dns", nullptr)` drops `unix` even though it's just an alias (I dislike this very much.)

Previously valid `pledge()` calls remain valid as well.

---

I've marked this as draft because I am not 100% sure that this is a good idea. I have only adjusted a few application to the new syntax as demonstration, I would go and push these changes into remaining applications in the following weeks.

#### Pro:

  - Explicitly listing all permissions is a annoying and makes it less obvious when permissions are dropped.

  - Conditionally dropping privileges is nasty at best (look at [Userland/ntpquery.cpp][1] for example) this could lead to a bias not to drop privileges for the sake of simplicity.

  - It could drive people to dynamically construct pledge strings using `StringBuilder`.

#### Contra:

  - Previously, it was very visible which permissions are available at any given point in time.

  - Previously, it was easier to spot dangerous permissions like "exec" (it can still be explicitly listed like "all exec -rpath" for example.)

  [1]: https://github.com/SerenityOS/serenity/blob/07d4b41bac1330804ad2bb4cd7656e5c8ca14717/Userland/ntpquery.cpp

### DIFF
--- a/Applications/About/main.cpp
+++ b/Applications/About/main.cpp
@@ -39,7 +39,7 @@ int main(int argc, char** argv)
 
     auto app = GUI::Application::construct(argc, argv);
 
-    if (pledge("stdio shared_buffer accept rpath", nullptr) < 0) {
+    if (pledge("all -unix -cpath -fattr", nullptr) < 0) {
         perror("pledge");
         return 1;
     }
--- a/Applications/Browser/main.cpp
+++ b/Applications/Browser/main.cpp
@@ -85,7 +85,7 @@ int main(int argc, char** argv)
     Web::ResourceLoader::the();
 
     // FIXME: Once there is a standalone Download Manager, we can drop the "unix" pledge.
-    if (pledge("stdio shared_buffer accept unix cpath rpath wpath", nullptr) < 0) {
+    if (pledge("all -fattr", nullptr) < 0) {
         perror("pledge");
         return 1;
     }
--- a/Base/usr/share/man/man2/pledge.md
+++ b/Base/usr/share/man/man2/pledge.md
@@ -30,6 +30,7 @@ If the process later attempts to use any system functionality it has previously 
 
 ## Promises
 
+* `all`: Placeholder for all promises that have not been dropped yet. (\*)
 * `stdio`: Basic I/O, memory allocation, information about self, various non-destructive syscalls
 * `thread`: The POSIX threading API (\*)
 * `id`: Ability to change UID/GID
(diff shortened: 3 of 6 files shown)


nico:
Maybe there could be a helper in LibCore for this? Then the syscall could stay mostly compatible with the bsd it's borrowed from.

awesomekling:
The pledge syntax is intentionally mimicking OpenBSD where it's considered a feature that you have to repeat the promises each time (since as you note, it makes it completely clear what the promises are at each point.) I tend to agree with them on that issue.

awesomekling:
I agree that the `dns` alias is dumb. Since we've ended up not copying the higher-level sandboxing semantics from OpenBSD, I think we can just drop that.

asynts:
Maybe this can be a feature that is just used when pledging dynamically? Hmm, I think I dislike this even more, having "forbidden" features.

How about a `drop_pledge` syscall? Then `pledge` is the default but if you have some complicated logic you can just use that syscall instead.

asynts:
I got mixed feelings about this, I think this pull request could be beneficial in some cases and detrimental in other cases. I am going to close this for now, maybe there is some variation of this that could work though.

asynts:
@awesomekling I had another idea, how about allowing listing privileges that have already been dropped? Here a mockup:

~~~c++
pledge("stdio inet dns settime", nullptr);

// Drop settime conditionally
if (!want_to_use_settime)
    pledge("stdio inet dns", nullptr);

use_dns();

// We add a question mark if we aren't sure if we still have the privilege, but if we do, we want to keep it.
pledge("stdio inet settime?", nullptr);

if (want_to_use_settime)
    use_settime();
~~~

This way we can drop privileges conditionally and also reason about which privileges we have at any given point in time.

PR #3527, opened 2020-09-17
Kernel: Clone FileDescriptor on fork

tomuta:
A fork inheriting file descriptors from the parent should be able
to use them independently. E.g. a seek in the fork should only affect
the fork but not the parent, and vice versa.

I came across this while working on some `ProcFS` changes for #3396.


tomuta:
Never mind, I'm not sure if this is really the right behavior after all...

PR #3532, opened 2020-09-18
Minesweeper: Fix inverted Single-Click Chording setting

Petelliott:
This was introduced by 705cee528a803b1671d16eeaf222d3318708500b, where the '!' was copied from the previous implementation, but the behavior was not.

I would like to apply for the record of smallest PR in serenity :^)


awesomekling:
Oopsie thank you! That may be the smallest PR indeed, next record will be in the 1-7 bit range :D

PR #3534, opened 2020-09-18
HackStudio: Only refresh Git widget on save if it was initialized

itamar8910:
This fixes an annoyance where the "Do you want to initialize a git repo?" popup would show when saving a file when the user hasn't event interacted with the git widget itself.


DevTools/HackStudio/Git/GitWidget.cpp
```diff
@@ -95,19 +95,36 @@ GitWidget::GitWidget(const LexicalPath& repo_root)
...
+}
+
+void GitWidget::refresh()
+{
+    if (!initialized()) {
+        bool rc = initialize();
+        if (!rc) {
+            dbg() << "GitWidget initialization failed";
+            return;
+        }
```
awesomekling, line 127: Maybe move this to an `initialize_if_needed()` helper to clarify intent here?

PR #3541, opened 2020-09-18
Automatically check code against clang-format

BenWiederhake:
Once upon a time:

> (21:35:09) CxByte: BenW: since you're the travis guy around here, how feasible is it to make it run clang-format and fail if anything changes?
> (21:45:04) BenW: CxByte: Precisely as difficult as getting clang-format available at all.
> (21:45:43) CxByte: Ah, basically impossible (until we [bump up the image distro version](https://github.com/SerenityOS/serenity/pull/3483)?)
> (21:47:01) BenW: We probably should. Ubuntu Focal has been around for long enough, and has been working nicely for other projects all the time.
> (21:47:28) kling: not a bad idea :)
> (21:47:58) CxByte: well, consider that a feature request then :P

Here is one way we could get `clang-format` to run on Travis.

Design decisions for this particular PR:
- The criterion is "All files are checked by clang-format, regardless of what the commit/PR touched".
- Whereever clang-format gets confused, `// clang-format off` is used.
- The two exceptional files that are assembly-heavy (clang-format can't handle assembly), so they get ignored completely.
- It is acceptable that there are 11 large-ish commits that fix whitespace and other trivial stuff. (Mostly from clang-format, with a little help from me.)
- The script is written to prefer speed (less than a second on travis; about 8 seconds on my PC.)
- It is acceptable that the diffs have to be inspected through `git diff`. (They are very [readable on Travis](https://travis-ci.com/github/BenWiederhake/serenity/builds/185257289#L261), though, and cause [no noise on success](https://travis-ci.com/github/BenWiederhake/serenity/builds/185256846#L261).)
- Commits with a normal name should be rather uncontroversial.
- The controversial stuff is separated into "DELAYED FIXUP" commits. I will later squash them later into the corresponding "normal" commit. I hope that this helps with discussion (and rebasing).

### False positives

> We *could* also skip the files with false positives, or with a lot more
> effort, test for such chunks (as you said, inline asm; nodiscard, etc.)
> before failing the build.
> @alimpfard

Turns out, there's not too many false positives after all. (I saw so many false positives because I tackled the Kernel first, and of course there are a few more corner cases due to asm and compiler-attributes.)

All false-positives are handled in "DELAYED FIXUP" commits. I think it's pretty non-invasive. The weird behavior of clang-format (is that a bug maybe?) around `AK::Result` and `Kernel::KResult` is unfortunate, but tolerable.

The alternative is trying to do magical white-listing.

Note that checking only the diffs is only a temporary solution: Sooner or later someone will add valid, sanely formatted code that disagrees with clang-format. "Only check the diffs" doesn't help with this. Allowing a handful of `// clang-format off` in the code base *does*.

### Messing up `git blame`

Now that I see the diffs, it doesn't seem too bad to me: It mostly affects the trailing `\` in macros, placement of opening braces, empty braces, and header inclusion order. These things change all the time. We have a history of noisy commits that constantly touch these things because the commit has to change the file anyway.

So these 11 commits shouldn't with git blame more than the "fix as we go" commits we had so far do. In particular, after this PR we're guaranteed to have a lot less of this noise, and if someone accidentally ends up here they can easily see that it's not the commit they were looking for.

Hence I argue these "whitespace changes" are okay. Especially because:

> There's a config option `blame.ignoreRevsFile`, which can be set to a
> file containing a list of revs to ignore when `blame`-ing:
> https://git-scm.com/docs/git-blame#Documentation/git-blame.txt---ignore-revs-fileltfilegt
> @alimpfard

I didn't do that for this PR, but if anyone insists I'll look into it.

### Why I don't want to check only the diffs

> We could also leave existing mis-formats and only run git-clang-format
> against PRs. This would solve the blame issue, and allow people to pass
> CI even when master is malformatted.
> @Petelliott

On it's face, this looks like a great suggestion!

However, the script will be run in three different setups:
- Pull requests: Here we can easily figure out the commit range through [environment variables](https://docs.travis-ci.com/user/environment-variables/).
- Pushes to private forks, e.g. the own repository: Here it's not so clear what to check. The Travis envvars could tell us the range that was pushed. However, this would hide errors in the first push on the second push, and those errors will magically resurface when making a PR.
  * One could try to fetch from kling's "original repository", then use `git merge-base` to figure out where we diverged, and then check that range. This sounds fragile, and we would need to handle "direct pushes by kling" as yet another thing.
- Local executions: All other lints can be executed locally, and should be! However, how would the script even begin to determine a reasonable range?

One might check the diffs only against a fixed known-good commit. This won't help with the topic of false-positives, but it would help against massive codebase-wide reformat commits.

- In the future: If the next update to (e.g.) `clang-format-11` would require huge "trivial" diffs, then going back to this diff-idea is a good idea.
- For now: I believe the 11 commits below are acceptable. So let's *not* do that.

## What this PR needs

- [ ] Is my assessment too far from reality? Did I overlook an approach that might end up being better?
- [x] Are those "DELAYED FIXUP" commits okay?
- [ ] After that, I need to squash them, and switch the PR from "draft" to non-draft. :)


Meta/lint-clang-format.sh
```diff
@@ -0,0 +1,37 @@
...
+
+set -e
+
+script_path=$(cd -P -- "$(dirname -- "$0")" && pwd -P)
+cd "${script_path}/.." || exit 1
+
+if ! command -v clang-format-10 >/dev/null 2>&1 ; then
+    echo "clang-format-10 is not available. Either skip this script, or install clang-format-10."
+    exit 1
+fi
```
linusg, line 11: Maybe we can extend the message to include a hint for people running this script on non-Ubuntu systems - like Fedora:

```
$ command -v clang-format-10; echo $?
1
$ command -v clang-format; echo $?
0
$ clang-format --version
clang-format version 10.0.0 (Fedora 10.0.0-2.fc32)
```

I think creating a symlink on those systems will make the script usable.

alimpfard: It's either `clang-format-10` or `clang-format`, why not just try both :shrug:

BenWiederhake: Now it tries both, and even warns when using a non-10 version of `clang-format` :)

BenWiederhake:
Changes since the last push:
- Squashed the "DELAYED FIXUP" commits since there were no comments anyway
- Rebased and resolved conflicts
- Fixed format-issues that arose in the meantime (all trivial)

Therefore, I mark this PR as ready :)

BenWiederhake:
Rebased for the win

awesomekling:
Ok let's do it!

BenWiederhake:
> The build failed, just like the previous build.

Aw man.

alimpfard:
it's just `test-js`

BenWiederhake:
Not my fault? Aww yiss!

PR #3543, opened 2020-09-18
ResourceGraph: Fix graph line position calculation

### LINKED ISSUE #3540: ResourceGraph: Graph does not match tooltip values
I haven't checked which one is actually wrong, but something is off here.

Three examples (the black area is 14px tall):

image

```
34.8÷247.8 = 0.140435835  -> 14% of available memory used
0.140435835×14 = 1.9661   -> Graph should fill 2px of 14px with color
                          -> Actually fills 3px
```

image

```
47.1÷247.8 = 0.190072639  -> 19% of available memory used
0.190072639×14 = 2.661    -> Graph should fill 3px of 14px with color
                          -> Actually fills 4px
```

image

```
124.7÷247.8 = 0.50322841  -> 50% of available memory used
0.50322841×14 = 7.0452    -> Graph should fill 7px of 14px with color
                          -> Actually fills 9px
```

linusg:
The calculation was only taking the rect's inner height, but not its top offset into account.

Fixes #3540.

### DIFF
--- a/MenuApplets/ResourceGraph/main.cpp
+++ b/MenuApplets/ResourceGraph/main.cpp
@@ -104,7 +104,7 @@ private:
         for (auto value : m_history) {
             painter.draw_line(
                 { rect.x() + i, rect.bottom() },
-                { rect.x() + i, (int)(rect.height() - (value * (float)rect.height())) },
+                { rect.x() + i, rect.top() + (int)(round(rect.height() - (value * rect.height()))) },
                 m_graph_color);
             ++i;
         }


MenuApplets/ResourceGraph/main.cpp
```diff
@@ -104,7 +104,7 @@ class GraphWidget final : public GUI::Frame {
         for (auto value : m_history) {
             painter.draw_line(
                 { rect.x() + i, rect.bottom() },
-                { rect.x() + i, (int)(rect.height() - (value * (float)rect.height())) },
+                { rect.x() + i, (int)(rect.top() + ((1 - value) * rect.height())) },
```
linusg, line 107: An alternative would be

```cpp
(int)(rect.top() + rect.height() - (value * (float)rect.height()))
```

nico: Should we `round()` before casting to int too?

linusg: Good point, will add that :)

tomuta: Yes rounding will make it work better:
```c++
rect.top() + (int)(round((float)rect.height() - (value * (float)rect.height())))
```
looks better to me.

tomuta (commented):
I don't think `rect.top()` should be included in the `round`.

PR #3550, opened 2020-09-19
Kernel: Remove Process::big_lock

tomuta:
We don't really need this lock anymore as we have locks around
everything that needs to be synchronized.


tomuta:
Hm might be a little premature...

PR #3552, opened 2020-09-19
Kernel: Fix assertion statement in GenericInterruptHandler

### LINKED ISSUE #3531: Kernel::IRQHandler: change_irq_number disables interrupts, but uses change_interrupt_number which expects interrupts to be enabled
IRQHandler::change_irq_number:
https://github.com/SerenityOS/serenity/blob/07d4b41bac1330804ad2bb4cd7656e5c8ca14717/Kernel/Interrupts/IRQHandler.cpp#L79-L84

GenericInterruptHandler::change_interrupt_number:
https://github.com/SerenityOS/serenity/blob/07d4b41bac1330804ad2bb4cd7656e5c8ca14717/Kernel/Interrupts/GenericInterruptHandler.cpp#L57-L64

IRQHandler::change_irq_number is only used in one place right now:
https://github.com/SerenityOS/serenity/blob/07d4b41bac1330804ad2bb4cd7656e5c8ca14717/Kernel/Devices/SB16.cpp#L166-L173

This issue never came up because SB16::set_irq_line is seemingly never called.
Also, since this uses change_irq_number, it has an extra interrupt disabler.
Plus, the only place where GenericInterruptHandler::change_interrupt_number is seemingly called right now is IRQHandler::change_irq_number.

supercomputer7:
We need to assert if interrupts are not disabled when changing the interrupt number of an interrupt handler.
Before this fix, any change like this would lead to a crash, because we are using InterruptDisabler in IRQHandler::change_irq_number.

Fixes #3531.

### DIFF
--- a/Kernel/Interrupts/GenericInterruptHandler.cpp
+++ b/Kernel/Interrupts/GenericInterruptHandler.cpp
@@ -56,7 +56,7 @@ GenericInterruptHandler::~GenericInterruptHandler()
 
 void GenericInterruptHandler::change_interrupt_number(u8 number)
 {
-    ASSERT_INTERRUPTS_ENABLED();
+    ASSERT_INTERRUPTS_DISABLED();
     ASSERT(!m_disable_remap);
     unregister_generic_interrupt_handler(InterruptManagement::acquire_mapped_interrupt_number(interrupt_number()), *this);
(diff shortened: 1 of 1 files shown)


awesomekling:
Oops, nice catch :)

PR #3561, opened 2020-09-20
AK: Eradicate BufferStream.

asynts:
This is a follow up to #3496 and #3007. Now all uses of `BufferStream` are gone and replaced by the much nicer `InputMemoryStream`, `OutputMemoryStream` and `DuplexMemoryStream` classes.

I added a long explanation in the commit message that remove `AK/BufferStream.h` in the hope that if someone is missing it, they can find the new memory stream classes by running `git log -- AK/BufferStream.h`.


awesomekling:
This does not build (see Travis log)

asynts:
It appears my build environment is broken, the problem was a missing `const` but it build fine on my computer (even though it should not.)

asynts:
I was not able to reproduce the build errors of the Travis system, on my system everything builds fine. I fixed the issues regardless (somehow the compiler was convinced that some values were potentially uninitialized.)

Now the Travis build fails because a JavaScript test fails, I have not touched any LibJS stuff and haven't really interacted with anything that is used by it. I can run `test-js` on my host system and in Serenity without any issues, yet in Travis it fails with a Segmentation Fault.

No idea what to do now.

awesomekling:
Very nice finally being rid of `BufferStream` :)

PR #3562, opened 2020-09-20
LibMarkdown: Add support for Tables

alimpfard:
I've wanted this for ages, I figured I might as well add it.
I'm not entirely sure what our markdown is based on, but some comments lead me to believe that it is loosely based on CommonMark, without the HTML support. regardless, I think tables are pretty useful to have.

cc @bugaevc - Also note the comment about `Text::parse()`.

### DIFF
--- a/Libraries/LibMarkdown/List.cpp
+++ b/Libraries/LibMarkdown/List.cpp
@@ -47,7 +47,7 @@ String List::render_to_html() const
     return builder.build();
 }
 
-String List::render_for_terminal() const
+String List::render_for_terminal(size_t) const
 {
     StringBuilder builder;
 
--- a/Libraries/LibMarkdown/Table.cpp
+++ b/Libraries/LibMarkdown/Table.cpp
@@ -1,0 +1,241 @@
+/*
+ * Copyright (c) 2020, the SerenityOS developers.
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include <AK/StringBuilder.h>
+#include <LibMarkdown/Table.h>
+
+namespace Markdown {
+
+String Table::render_for_terminal(size_t view_width) const
+{
+    auto unit_width_length = view_width == 0 ? 4 : ((float)(view_width - m_columns.size()) / (float)m_total_width);
+    StringBuilder format_builder;
+    StringBuilder builder;
+
+    auto write_aligned = [&](const auto& text, auto width, auto alignment) {
+        size_t original_length = 0;
+        for (auto& span : text.spans())
+            original_length += span.text.length();
+        auto string = text.render_for_terminal();
+        format_builder.clear();
+        if (alignment == Alignment::Center) {
+            auto padding_length = (width - original_length) / 2;
+            builder.appendf("%*s%s%*s", padding_length, "", string.characters(), padding_length, "");
+            if ((width - original_length) % 2)
+                builder.append(' ');
+        } else {
+            format_builder.appendf("%%%s%zus", alignment == Alignment::Left ? "-" : "", width + (string.length() - original_length));
+            builder.appendf(
+                format_builder.to_string().characters(),
+                string.characters());
+        }
+    };
+
+    bool first = true;
+    for (auto& col : m_columns) {
+        if (!first)
+            builder.append('|');
+        first = false;
+        size_t width = col.relative_width * unit_width_length;
+        write_aligned(col.header, width, col.alignment);
+    }
+
+    builder.append("\n");
+    for (size_t i = 0; i < view_width; ++i)
+        builder.append('-');
+    builder.append("\n");
+
+    for (size_t i = 0; i < m_row_count; ++i) {
+        bool first = true;
+        for (auto& col : m_columns) {
+            ASSERT(i < col.rows.size());
+            auto& cell = col.rows[i];
+
+            if (!first)
+                builder.append('|');
+            first = false;
+
+            size_t width = col.relative_width * unit_width_length;
+            write_aligned(cell, width, col.alignment);
+        }
+        builder.append("\n");
+    }
+
+    return builder.to_string();
+}
+
+String Table::render_to_html() const
+{
+    StringBuilder builder;
+
+    builder.append("<table>");
+    builder.append("<thead>");
+    builder.append("<tr>");
+    for (auto& column : m_columns) {
+        builder.append("<th>");
+        builder.append(column.header.render_to_html());
+        builder.append("</th>");
+    }
+    builder.append("</tr>");
+    builder.append("</thead>");
+    builder.append("<tbody>");
+    for (size_t i = 0; i < m_row_count; ++i) {
+        builder.append("<tr>");
+        for (auto& column : m_columns) {
+            ASSERT(i < column.rows.size());
+            builder.append("<td>");
+            builder.append(column.rows[i].render_to_html());
+            builder.append("</td>");
+        }
+        builder.append("</tr>");
+    }
+    builder.append("</tbody>");
+    builder.append("</table>");
+
+    return builder.to_string();
+}
+
+OwnPtr<Table> Table::parse(Vector<StringView>::ConstIterator& lines)
+{
+    auto peek_it = lines;
+    auto first_line = *peek_it;
+    if (!first_line.starts_with('|'))
+        return nullptr;
+
+    ++peek_it;
+
+    if (peek_it.is_end())
+        return nullptr;
+
+    auto header_segments = first_line.split_view('|', true);
+    auto header_delimiters = peek_it->split_view('|', true);
+
+    if (!header_segments.is_empty())
+        header_segments.take_first();
+    if (!header_segments.is_empty() && header_segments.last().is_empty())
+        header_segments.take_last();
+
+    if (!header_delimiters.is_empty())
+        header_delimiters.take_first();
+    if (!header_delimiters.is_empty() && header_delimiters.last().is_empty())
+        header_delimiters.take_last();
+
+    ++peek_it;
+
+    if (header_delimiters.size() != header_segments.size())
+        return nullptr;
+
+    if (header_delimiters.is_empty())
+        return nullptr;
+
+    size_t total_width = 0;
+
+    auto table = make<Table>();
+    table->m_columns.resize(header_delimiters.size());
+
+    for (size_t i = 0; i < header_segments.size(); ++i) {
+        auto text_option = Text::parse(header_segments[i]);
+        if (!text_option.has_value())
+            return nullptr; // An invalid 'text' in the header should just fail the table parse.
+
+        auto text = text_option.release_value();
+        auto& column = table->m_columns[i];
+
+        column.header = move(text);
+
+        auto delimiter = header_delimiters[i].trim_whitespace();
+
+        auto align_left = delimiter.starts_with(':');
+        auto align_right = delimiter != ":" && delimiter.ends_with(':');
+
+        if (align_left)
+            delimiter = delimiter.substring_view(1, delimiter.length() - 1);
+        if (align_right)
+            delimiter = delimiter.substring_view(0, delimiter.length() - 1);
+
+        if (align_left && align_right)
+            column.alignment = Alignment::Center;
+        else if (align_right)
+            column.alignment = Alignment::Right;
+        else
+            column.alignment = Alignment::Left;
+
+        size_t relative_width = delimiter.length();
+        for (auto ch : delimiter) {
+            if (ch != '-') {
+                dbg() << "Invalid character _" << ch << "_ in table heading delimiter (ignored)";
+                --relative_width;
+            }
+        }
+
+        column.relative_width = relative_width;
+        total_width += relative_width;
+    }
+
+    table->m_total_width = total_width;
+
+    for (off_t i = 0; i < peek_it - lines; ++i)
+        ++lines;
+
+    size_t row_count = 0;
+    ++lines;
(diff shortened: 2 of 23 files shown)


Libraries/LibMarkdown/List.cpp
```diff
@@ -58,7 +58,7 @@ String List::render_for_terminal() const
             builder.appendf("%d. ", ++i);
         else
             builder.append("* ");
-        builder.append(item.render_for_terminal());
+        builder.append(item.render_for_terminal(view_width));
```
bugaevc, line 61: This seems wrong — we've already taken up some of that horizontal space with the `* ` or the `1. `, yet you're passing the full width.

Libraries/LibMarkdown/Text.h
```diff
@@ -53,7 +54,7 @@ class Text final {
     const Vector<Span>& spans() const { return m_spans; }
 
     String render_to_html() const;
-    String render_for_terminal() const;
+    String render_for_terminal(size_t view_width = 0) const;
```
bugaevc, line 60: Does text need the width too?

(Note that Text is not a Block subclass)

Libraries/LibMarkdown/List.cpp
alimpfard: Ah, I forgot to fix those.
my bad!

Userland/md.cpp
```diff
@@ -32,22 +32,34 @@
...
     bool html = false;
+    int view_width = 0;
 
     Core::ArgsParser args_parser;
     args_parser.add_option(html, "Render to HTML rather than for the terminal", "html", 'H');
+    args_parser.add_option(view_width, "Viewport width for the terminal (defaults to current terminal width)", "view-width", 0, "width");
     args_parser.add_positional_argument(file_name, "Path to Markdown file", "path", Core::ArgsParser::Required::No);
     args_parser.parse(argc, argv);
 
+    if (!html && view_width == 0 && isatty(STDOUT_FILENO)) {
```
bugaevc, line 55: So if `!html && view_width == 0`, but `!isatty(STDOUT_FILENO)`, you're going to continue with 0 (and not say 80)?

Libraries/LibMarkdown/Text.h
alimpfard: I suppose not, but same-looking function with different arguments is kinda weird too?

Libraries/LibMarkdown/Table.h
```diff
@@ -0,0 +1,66 @@
...
+        Text header;
+        Vector<Text> rows;
+        Alignment alignment { Alignment::Left };
+        size_t relative_width { 0 };
+    };
+
+    Table()
+    {
+    }
+    virtual ~Table() override { }
```
bugaevc, line 54: This looks weird; use the same brace style for both constructor and destructor :slightly_smiling_face:

alimpfard: I don't even look at these, I just put my brain on boilerplate mode and have it churn out ctors/dtors. guess I should take a look at them once in a while '^^

Libraries/LibMarkdown/Table.cpp
```diff
@@ -0,0 +1,214 @@
...
+    if (!first_line.starts_with('|'))
+        return nullptr;
+
+    ++peek_it;
+
+    if (peek_it.is_end())
+        return nullptr;
+
+    auto header_segments = first_line.split_view('|');
+    auto header_delimiters = peek_it->split_view('|');
```
bugaevc, line 135: `keep_empty = true` on both, no?

alimpfard: Hmmm...I somehow had this odd idea that `keep_empty` applied to the last element
yeah, sowwy~

Libraries/LibMarkdown/Text.h
bugaevc: Yes; but at the same time it's unclear how text would use this info. Text, in CSS terms, is `inline`, unlike blocks which are `block`. It's clear what a viewport width means for blocks — it's the width they should take up (and adjust their height based on that). It's not clear what it means for text, because it can wrap between lines and whatnot.

I'm not saying you should remove it, but perhaps you should define a clearer semantics of what this argument means here.

alimpfard: well it really means nothing for Text, I guess its existence is more confusing than not.
removing it probably makes more sense.

Libraries/LibMarkdown/Table.cpp
```diff
@@ -0,0 +1,221 @@
...
+        return nullptr;
+
+    size_t total_width = 0;
+
+    auto table = make<Table>();
+    table->m_columns.resize(header_delimiters.size());
+
+    for (size_t i = 0; i < header_segments.size(); ++i) {
+        auto text_option = Text::parse(header_segments[i]);
+        ASSERT(text_option.has_value()); // Text::parse() cannot fail yet returns an Optional. why?
```
bugaevc, line 159: That's right, it cannot fail, yet we pretend it can. I thought that was best, in case we ever add availability to it.

You're basically expected to treat this as a parsing failure rather than asserting. E.g. `return nullptr;` yourself.

Libraries/LibMarkdown/Table.cpp
```diff
@@ -0,0 +1,221 @@
...
+        auto& column = table->m_columns[i];
+
+        column.header = move(text);
+
+        auto& delimiter = header_delimiters[i];
+
+        while (delimiter.starts_with(' '))
+            delimiter = delimiter.substring_view(1, delimiter.length() - 1);
+        while (delimiter.ends_with(' '))
+            delimiter = delimiter.substring_view(0, delimiter.length() - 1);
```
bugaevc, line 170: Don't we have a nicer method to trim spaces?

If we don't, let's add one and use it here.

Libraries/LibMarkdown/Table.cpp
```diff
@@ -0,0 +1,221 @@
...
+
+        auto left = delimiter.starts_with(':');
+        auto right = delimiter != ":" && delimiter.ends_with(':');
```
bugaevc, line 173: Rather, `align_left`, `align_right`

Libraries/LibMarkdown/Table.cpp
```diff
@@ -0,0 +1,221 @@
...
+
+        if (left)
+            delimiter = delimiter.substring_view(1, delimiter.length() - 1);
+        if (right)
+            delimiter = delimiter.substring_view(0, delimiter.length() - 1);
+
+        if (left && right)
+            column.alignment = Alignment::Center;
+        else if (right)
+            column.alignment = Alignment::Right;
```
bugaevc, line 183:
```cpp
else
    column.alignment = Alignment::Left;
```

even though it's the default value. The default may change, and this code shouldn't rely on it.

Libraries/LibMarkdown/Table.cpp
```diff
@@ -0,0 +1,221 @@
...
+
+        size_t relative_width = delimiter.length();
+        for (auto ch : delimiter) {
+            if (ch != '-') {
+                dbg() << "Invalid character '" << ch << "' in table heading delimiter";
```
bugaevc, line 188: I *think* the idiomatic *Serenity way* to quote weird characters is with a pair of \_'s instead of single quotes, but you might want to ask @awesomekling about that.

Libraries/LibMarkdown/Table.cpp
alimpfard: a parse failure in the header is fine, but what should a parse failure inside a structure that is _known_ to be a table do?
be ignored, treated as literal text, or just immediately fail the table parse?

|a|b|
|---|---|
|`|x|

github seems to take it as literal text

Libraries/LibMarkdown/Table.cpp
alimpfard: why would it change?

Libraries/LibMarkdown/Table.cpp
```diff
@@ -0,0 +1,221 @@
...
+    builder.append("</tbody>");
+    builder.append("</table>");
+
+    return builder.to_string();
+}
+
+OwnPtr<Table> Table::parse(Vector<StringView>::ConstIterator& lines)
+{
+    auto peek_it = lines;
+    auto first_line = *peek_it;
```
bugaevc, line 125: I don't really get the point of using two separate iterators here.

For the cases where the markdown is malformed, yet this was clearly meant to be a table, it's fine to `return nullptr;` while leaving `lines` in the middle of the supposed-table. It's only important that you don't move `lines` until you commit to this being a table.

Libraries/LibMarkdown/Table.cpp
```diff
@@ -0,0 +1,221 @@
...
+    }
+
+    table->m_total_width = total_width;
+
+    for (off_t i = 0; i < peek_it - lines; ++i)
+        ++lines;
+
+    size_t row_count = 0;
+    ++lines;
+    while (!lines.is_end()) {
```
bugaevc, line 204: No other way this loop can exit??

what | if
---|---
I | have
a | table

and then some more text?

Libraries/LibMarkdown/Table.cpp
```diff
@@ -0,0 +1,221 @@
...
+        auto segments = lines++->split_view('|');
```
bugaevc, line 205: `keep_empty`

Libraries/LibMarkdown/Table.cpp
bugaevc: Oh, did you mean it as an actual default and not as just *always give some reasonable default to all fields*?

Libraries/LibMarkdown/Table.cpp
alimpfard: `peek_it` is only used for the first two lines, which is how we decide if this is a table.
example:
|a|b|
|---|
|foo|

is not a table

but

|a|b|
|---|---|
|foo|

is a table.

Libraries/LibMarkdown/Table.cpp
alimpfard: that sounds like an error...oops

Libraries/LibMarkdown/Table.cpp
```diff
@@ -0,0 +1,221 @@
...
+        builder.append(column.header.render_to_html());
+        builder.append("</th>");
+    }
+    builder.append("</tr>");
+    builder.append("</thead>");
+    builder.append("<tbody>");
+    for (size_t i = 0; i < m_row_count; ++i) {
+        builder.append("<tr>");
+        for (auto& column : m_columns) {
+            if (column.rows.size() > i) {
```
bugaevc, line 108: I'd write this as
```suggestion
            if (i < column.rows.size()) {
```
(pronounced "if `i` is a valid index into `column.rows`)
and perhaps add a comment that columns can have different numbers of rows.

Libraries/LibMarkdown/Table.cpp
```diff
@@ -0,0 +1,221 @@
...
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include <LibMarkdown/Table.h>
+
+namespace Markdown {
+
+String Table::render_for_terminal(size_t view_width) const
+{
+    auto segment_width_proportion = view_width == 0 ? 4 : ((float)(view_width - m_columns.size()) / (float)m_total_width);
```
bugaevc, line 33: Not immediately clear what this is ⇒ needs a better/longer name and a comment.

Libraries/LibMarkdown/Table.cpp
alimpfard: Hmmm...actually, this might be a tad bit wrong; it'll shift all empty cells to the end of the column.

Libraries/LibMarkdown/Table.cpp
bugaevc: I'd treat

```
|a|b|
|---|
|foo|
```

as a (broken) table too. CommonMark doesn't specify a syntax for tables, so we're good.

Libraries/LibMarkdown/Table.cpp
alimpfard: I did mean it as an actual default, seems like github centers by default though :thinking:

Libraries/LibMarkdown/Table.cpp
```diff
@@ -0,0 +1,221 @@
...
+
+    auto write_aligned = [&](const auto& text, auto width, auto alignment) {
+        size_t original_length = 0;
+        for (auto& span : text.spans())
+            original_length += span.text.length();
+        auto string = text.render_for_terminal();
+        format_builder.clear();
+        if (alignment == Alignment::Center) {
+            auto padding_length = (width - original_length) / 2;
+            builder.appendf("%*s%s%*s", padding_length, "", string.characters(), padding_length, "");
```
bugaevc, line 45: Doesn't printf have a built-in way to pad/align things?

alimpfard: Only left xor right.

Libraries/LibMarkdown/Table.cpp
alimpfard: so what should it do? 
- just ignore the extra columns?
- fail right in the delimiters?
- ?

Libraries/LibMarkdown/Table.cpp
alimpfard: Gladly!

Libraries/LibMarkdown/Table.cpp
alimpfard: well, we can tweak it later if centering by default seems better.

Libraries/LibMarkdown/Table.cpp
alimpfard: it seems rather clear to me, but how about `unit_width_size`?

bugaevc: Yeah, or `size_unit_width`? (as in, width of a *unit of size*)

alimpfard: lol and I was thinking "size of a unit of width"

Libraries/LibMarkdown/Table.cpp
alimpfard: I've seen that too, sure, will use pair of _'s.

Libraries/LibMarkdown/Table.cpp
bugaevc: `return nullptr;` the moment something is too broken to handle, and from there on it's somebody else's responsibility; most likely it's going to be a Paragraph with some Text.

Which reminds me, Paragraph has special code to detect if it's actually should fail parsing and defer to some other kind of block:

https://github.com/SerenityOS/serenity/blob/e3e5e57fdefa90fd682d6a8a6f6b64ce311ed93e/Libraries/LibMarkdown/Paragraph.cpp#L64-L79

Perhaps you should add some support for tables there. Or, since it's not quite clear how that would work (detecting that the line contains `|`s? but what if they're escaped or inside <code>`</code>s?), make super sure Paragraph is always the last type of Block to be tried, and remove all that mess.

alimpfard: Hmmm...does it
(no whitespace)
|work|like|that|
|---|---|---|
|I|wonder|?|

(yes whitespace)
well, I can't really check for "is it a table" in two places, perhaps there should be a `Block::can_be_parsed()` which Paragraph tests with?

Either way, since the semantics of tables are defined by us, we can just make whitespace mandatory before them; it's visually a separate paragraph after all

bugaevc: I'm not talking about whitespaces, even.

I'm saying that as of now, there's no guarantee in which order we ask various block types to try and see if the current line can be parsed as that block. `Paragraph` can, in principle, eat anything, as it just wraps `Text`, so it needs to either:
* always be checked last,
* explicitly reject lines that look like they belong to other types of blocks.

Right now the latter option is implemented. I'm proposing we switch to the former (guarantee we always try Paragraph last) and remove all of this logic from `Paragraph::parse()`, and have it just call `Text::parse()`.

alimpfard: "always be checked last" doesn't really work out though, a paragraph eats all the lines it can (as of right now, until it reaches something that can be parsed as another block).
unless we make `Line`s instead, and join them later, perhaps; but that just reeks of doing things twice
---

foobar
#### This is a test heading which comes right after a "paragraph"

bugaevc: Right, so we can both require a newline before a table and always check paragraphs last.

That would still mean we lose non-newlined lists though, that'd be quite sad.

alimpfard: well, the only other solution that comes to me is the `Line` one, we'd have the cake (even if it's in pieces) and we can eat it too?

alimpfard: -> 886c6d9

awesomekling:
Did you have more comments here @bugaevc?

bugaevc:
Not really, no.

I don't quite like making `Paragraph` *this* special, but it also makes sense, so, fine 🤷

alimpfard:
I'll also add that I don't really like this form either, but I like it much more than doubling up on where "is this an X" checks are performed.

(although I don't see any reason not to specialise for the specific language we're trying to parse)

alimpfard:
That should deal with the conflict

PR #3564, opened 2020-09-20
LibGUI: Pad row/column headers of AbstractTableView

xTibor:
How the table headers look after this patch:
Screenshot_20200920_165812


awesomekling:
Oh nice, that looks way better :)

PR #3565, opened 2020-09-20
HackStudio: Basic autocomplete for C++

itamar8910:
This adds basic autocomplete for C++ in HackStudio, triggered with `Ctrl+Space`.

Nothing fancy here, using the `CppLexer` to find identifiers that match with what the user wants to complete.

In the future, it would be interesting to have some sort of Language Server which would have more sophisticated logic than what we currently do.

Screenshot:
hs-autocomplete

### DIFF
--- a/DevTools/HackStudio/AutoCompleteBox.cpp
+++ b/DevTools/HackStudio/AutoCompleteBox.cpp
@@ -1,0 +1,166 @@
+/*
+ * Copyright (c) 2020, Itamar S. <itamar8910@gmail.com>
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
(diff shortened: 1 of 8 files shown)


DevTools/HackStudio/AutoCompleteWidget.h
```diff
@@ -0,0 +1,58 @@
...
+
+#include <AK/WeakPtr.h>
+#include <LibGUI/Widget.h>
+
+namespace HackStudio {
+
+class Editor;
+class AutoCompleteWidget final : public GUI::Widget {
+    C_OBJECT(AutoCompleteWidget)
+public:
```
awesomekling, line 37: This doesn't seem like it should be a widget since it doesn't participate in widget layout, nor does it paint itself.

itamar8910: ofc! Now it's a non-widget, `AutoCompleteBox`

PR #3568, opened 2020-09-20
FontEditor: Draw the baseline in the glyph editor widget

xTibor:
To make it [easier to determine](https://youtu.be/YjNtCClX6fM?t=1033) where font baselines should go.

**Screenshot:**
Screenshot_20200920_213023


awesomekling:
So simple but perfect :)

PR #3569, opened 2020-09-20
LibGUI: Add optional placeholder to TextEditor

Petelliott:
I was using Help, and I noticed that it implemented an Ad-hoc placeholder for the search bar. I decided to implement it more formally, and add it to some other applications.


awesomekling:
Very tasteful :)

Issue #3329, opened 2020-08-28
UserspaceEmulator: Support floating point (x87) instructions


awesomekling:
UE should support x87 FPU instructions.

qeeg:
Using UE for `/bin/Fire` needs this.

awesomekling:
It's needed for all GUI applications :)

nico:
I poke at this every now and then at https://github.com/nico/serenity/tree/fpu

Not to call dibs, but if someone wants to try and beat me to it, maybe they want to look at that branch for inspiration :) (Or, eventually, I'll make a pull request myself hopefully.)

bcoles:
I believe all of the instructions implemented in nico's fork mentioned above have been merged.

About half of the [FPU instruction set](http://linasm.sourceforge.net/docs/instructions/fpu.php) has been implemented so far.

However, a significant portion of the implemented instructions are only partially implemented as the C0, C2 and C3 flags are not set appropriately and exceptions are not generated.

The majority of the remaining instructions are comparison instructions and control instructions.

Hendiadyoin1:
Fixed, but buggy

PR #3576, opened 2020-09-21
some fpu support for ue

nico:
Part of #3329.

This is nowhere near done. It doesn't implement fpu exceptions, quiet/signaling nans, exception on div by 0, correct rounding modes, or even most fpu instructions (no fsave/frestore, fnstsw doesn't return anything, ...). It generally assumes that things work fine and aren't weird or corner-case-y. It *can* run ntpclient and TextEditor. 

Things to do to get this mergeable (in addition to removing debug logging, and cleaning up the commits):

- [ ] Come up with a shadow handling strategy for fpu words. The FPU registers are 80-bit wide so per-bit shadow state gets smeared out immediately. Probably just keep a single "is valid" bit per register and set that if any input bit is invalid and use it to taint outputs?

- [ ] Probably replace all bit_cast<> calls with just reinterpret_cast pointer cast + deref? That seems what most other places do, but something like bit_cast<> is what the standard wants us to do.

- [ ] Many of the instructions I did implement explicitly check `modrm().is_register()`, which looks wrong -- Instruction.cpp should probably use different InstructionFormats and call different functions for register and memory operands (?). Or maybe this is fine as-is; not sure what the desired design here is. Most other existing instruction handlers don't have this check though.

@awesomekling Input on these 3 points would be appreciated :)

### DIFF
--- a/DevTools/UserspaceEmulator/SoftCPU.cpp
+++ b/DevTools/UserspaceEmulator/SoftCPU.cpp
@@ -1474,7 +1474,6 @@ void SoftCPU::FLDCW(const X86::Instruction& insn)
 
 void SoftCPU::FLD1(const X86::Instruction&)
 {
-dbg() << "fld1";
     fpu_push(1.0);
 }
 
@@ -1483,7 +1482,12 @@ void SoftCPU::FLDL2E(const X86::Instruction&) { TODO_INSN(); }
 void SoftCPU::FLDPI(const X86::Instruction&) { TODO_INSN(); }
 void SoftCPU::FLDLG2(const X86::Instruction&) { TODO_INSN(); }
 void SoftCPU::FLDLN2(const X86::Instruction&) { TODO_INSN(); }
-void SoftCPU::FLDZ(const X86::Instruction&) { TODO_INSN(); }
+
+void SoftCPU::FLDZ(const X86::Instruction&)
+{
+    fpu_push(0.0);
+}
+
 void SoftCPU::FNSTENV(const X86::Instruction&) { TODO_INSN(); }
 void SoftCPU::F2XM1(const X86::Instruction&) { TODO_INSN(); }
 void SoftCPU::FYL2X(const X86::Instruction&) { TODO_INSN(); }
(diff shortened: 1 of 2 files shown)


qeeg:
I think the "shadow handling" stuff that you're talking about related to the x87 Tag Word register, which you didn't seem to implement.

awesomekling:
Very cool progress!

> Things to do to get this mergeable (in addition to removing debug logging, and cleaning up the commits):
> 
> * [ ]  Come up with a shadow handling strategy for fpu words. The FPU registers are 80-bit wide so per-bit shadow state gets smeared out immediately. Probably just keep a single "is valid" bit per register and set that if any input bit is invalid and use it to taint outputs?

Yes, a single bit per register / stack slot seems fine. I don't think tracking the state of individual bytes in FPU registers is particularly useful anyway :)

> * [ ]  Probably replace all bit_cast<> calls with just reinterpret_cast pointer cast + deref? That seems what most other places do, but something like bit_cast<> is what the standard wants us to do.

Maybe bit_cast is best then? I'm not sure. Are we causing some kind of UB without it?

> * [ ]  Many of the instructions I did implement explicitly check `modrm().is_register()`, which looks wrong -- Instruction.cpp should probably use different InstructionFormats and call different functions for register and memory operands (?). Or maybe this is fine as-is; not sure what the desired design here is. Most other existing instruction handlers don't have this check though.

Yeah the modrm() abstraction in general doesn't seem to make sense here since FPU instructions don't take the usual kind of R/M address bytes. I think it would be nicer to add something to X86::Instruction tailored for these instructions and the way they are specified by the Intel manual instead of piggybacking on modrm() just because it's there :)

nico:
> Maybe bit_cast is best then? I'm not sure. Are we causing some kind of UB without it?

Using `reinterpret_cast<>` to cast a pointer type to `void*` and then from `void*` back to the original type is the only valid use of `reinterpret_cast<>`. Every other use is technically UB. But it's fairly common, and `-fno-strict-aliasing` makes it work in practice. ...ah, https://blog.regehr.org/archives/959 looks like a good write-up. A bit old, but I believe still true. (The `static_assert` in bit_cast is a nice safety addition too.)

I've since found several `*reinterpret_cast<T*>(...)` in serenity so I'll use that for now since it's independent of this particular issue. Let me know if I should send a separate pull request for adding `bit_cast<>` somewhere in AK and using it in a few places.

> > * [ ]   Many of the instructions I did implement explicitly check `modrm().is_register()`, which looks wrong -- Instruction.cpp should probably use different InstructionFormats and call different functions for register and memory operands (?). Or maybe this is fine as-is; not sure what the desired design here is. Most other existing instruction handlers don't have this check though.
> 
> Yeah the modrm() abstraction in general doesn't seem to make sense here since FPU instructions don't take the usual kind of R/M address bytes. I think it would be nicer to add something to X86::Instruction tailored for these instructions and the way they are specified by the Intel manual instead of piggybacking on modrm() just because it's there :)

From what I understand, FPU instructions generally do use the R/M scheme, it's just that the register number in there refers to an FPU register instead of a gpreg. But I think I still get what you're saying; I'll cook something up.

nico:
I tried removing bit_cast and ran into these 2 issues:

```
../DevTools/UserspaceEmulator/SoftCPU.cpp:1508:58: error: lvalue required as unary '&' operand
 1508 |         fpu_push(*reinterpret_cast<float*>(&new_f32.value()));
      |                                             ~~~~~~~~~~~~~^~
```

That means I have to write to a temporary and cast that. Somewhat annoying.

```
../DevTools/UserspaceEmulator/SoftCPU.cpp: In member function 'virtual void UserspaceEmulator::SoftCPU::FST_RM32(const X86::Instruction&)':
../DevTools/UserspaceEmulator/SoftCPU.cpp:1525:67: error: dereferencing type-punned pointer will break strict-aliasing rules [-Werror=strict-aliasing]
 1525 |     insn.modrm().write32(*this, insn, shadow_wrap_as_initialized(*reinterpret_cast<u32*>(&f32)));
      |                                                                   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

This can be hacked around by writting to a `char* foo` temporary, and maybe the optimizer can see through this even when the compiler can't and maybe it'd miscompile this somehow.

So maybe it's worth adding bit_cast first.

awesomekling:
bit_cast seems fine to me :)

awesomekling:
I merged this in 04d9af7 so we can iterate on it in tree.
Thank you @nico for getting this work started :)

asynts:
> This is all work done by @nico for #3576. I'm just merging it all in this patch since it's a decent foundation to continue working on. :^)

GitHub actually has support for `Co-Authored-By`. You can add that to the bottom of commit messages and it will attribute it to multiple people:

~~~none
Co-Authored-By: Alice <alice@example.com>
~~~

(Not that this is really relevant, but I thought this might be useful information.)

https://docs.github.com/en/free-pro-team@latest/github/committing-changes-to-your-project/creating-a-commit-with-multiple-authors#creating-co-authored-commits-on-the-command-line

PR #3580, opened 2020-09-22
AK: Make format avaliable in kernel.

asynts:
Previously, it was not possible to use `format` in the kernel, this was because of two oversights of mine:

 1. I forgot to add `Format.cpp` to the CMake configuration of the kernel.

 2. I used `strtoull` which isn't available in the kernel.

I also added an method to `StringBuilder` and changed `String::number` to use the new `format`.

### DIFF
--- a/AK/Format.cpp
+++ b/AK/Format.cpp
@@ -182,13 +182,18 @@ void Formatter<T, typename EnableIf<IsIntegral<T>::value>::Type>::format(StringB
 
 template struct Formatter<StringView>;
 template struct Formatter<String>;
-template struct Formatter<u8, void>;
-template struct Formatter<u16, void>;
-template struct Formatter<u32, void>;
-template struct Formatter<u64, void>;
-template struct Formatter<i8, void>;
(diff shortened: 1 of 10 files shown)


tomuta:
I guess I'm a little late to the party, but for APIs like `String::number` we should consider a dedicated version that converts the number directly rather than going through a string scan first. `libfmt` has separate APIs for that reason as well. Their implementation is also supposedly one of the fastest out there, though I haven't looked at the implementation myself.

awesomekling:
No such thing as late to the party; this party is permanent :)
I agree that sounds like a good idea @tomuta

asynts:
That does sound like a good feature. I'll add if when I come up with a good name.

  > [...] Their implementation is also supposedly one of the fastest out there, though I haven't looked at the implementation myself.

fmtlib is extremely optimized not just for speed, but also for binary size. One big trade-off is that it adds significant compile time. I don't think this implementation even with optimizations will ever get even close to that.

PR #3584, opened 2020-09-23
UserspaceEmultor+LibX86: Add support for 64-bit memory reads and writes

nico:
This is useful for reading and writing doubles for #3329.
It is also useful for emulating 64-bit binaries.

MemoryOrRegisterReference assumes that 64-bit values are always
memory references since that's enough for fpu support. If we
ever want to emulate 64-bit binaries, that part will need minor
updating.

### DIFF
--- a/DevTools/UserspaceEmulator/SoftCPU.h
+++ b/DevTools/UserspaceEmulator/SoftCPU.h
@@ -360,6 +360,7 @@ public:
     ValueWithShadow<u8> read_memory8(X86::LogicalAddress);
     ValueWithShadow<u16> read_memory16(X86::LogicalAddress);
     ValueWithShadow<u32> read_memory32(X86::LogicalAddress);
+    ValueWithShadow<u64> read_memory64(X86::LogicalAddress);
 
     template<typename T>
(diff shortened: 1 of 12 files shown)


DevTools/UserspaceEmulator/SoftCPU.h
```diff
@@ -1154,4 +1157,15 @@ ALWAYS_INLINE u32 SoftCPU::read32()
     return value;
 }
 
+ALWAYS_INLINE u64 SoftCPU::read64()
+{
+    if (!m_cached_code_ptr || (m_cached_code_ptr + 8) >= m_cached_code_end)
```
nico, line 1162: The `+ 8` here is consistent with `read16()` and `read32()`, but not with `read8()`. I believe `read8()` is right and `read16()`, `read32()` and `read64()` should add 1, 3, and 7 instead of 2, 4, and 8. But that should probably be in a separate commit and even a separate pull request, and it conflicts with the one here, so I'd prefer to do it after this one is merged. But if you prefer I can send that out first, wait for that to be merged, and then rebase this on the other one.

PR #3589, opened 2020-09-23
UserspaceEmulator: Fix off-by-one in code cache access

nico:
m_cached_code_end points at the first invalid byte, so we need to
update the cache if the last byte we want to read points at the
end or past it.  Previously we updated the cache 1 byte prematurely in
read16, read32, read64 (but not in read8).

Noticed by reading the code (the code looked different from read8() and
the other 3).  I didn't find anything that actually hit this case.

--

I hope I'm updating the right functions. I think read8() is right and the others are wrong, but maybe it's the other way round. It is probably unknowable!


awesomekling:
Oh good catch :)

PR #3597, opened 2020-09-25
AK: Borrow format specifier syntax from std::format.

asynts:
I explained my thoughts in the commit message of c74db5b0e45a93217ae315860b82db7c237acdd4.

Essentially, this commit just adds a parser for the [standard format specifier syntax][1] used by `std::format`.

I also fixed a few bugs (and surely introduced at least as many new ones.)

Finally, I moved the assertions closer to the format specifier parser. Previously, `Formatter<T>::parse` would return false if parsing fails, now it actually asserts, which is super useful for debugging. (Previously, we would assert that this function returns true.)

  [1]: https://en.cppreference.com/w/cpp/utility/format/formatter#Standard_format_specification

### DIFF
--- a/AK/Format.cpp
+++ b/AK/Format.cpp
@@ -73,33 +73,51 @@ static void write_escaped_literal(StringBuilder& builder, StringView literal)
     }
 }
 
-static size_t parse_number(StringView input)
+static bool parse_number(GenericLexer& lexer, size_t& value)
 {
-    size_t value = 0;
+    value = 0;
 
-    for (char ch : input) {
-        value *= 10;
-        value += ch - '0';
+    bool consumed_at_least_one = false;
+    while (!lexer.is_eof()) {
+        if (lexer.next_is(is_digit)) {
+            value *= 10;
+            value += lexer.consume() - '0';
+            consumed_at_least_one = true;
+        } else {
+            break;
+        }
     }
 
-    return value;
+    return consumed_at_least_one;
 }
+
+constexpr size_t use_next_index = NumericLimits<size_t>::max();
 
 static bool parse_format_specifier(StringView input, FormatSpecifier& specifier)
 {
-    specifier.index = NumericLimits<size_t>::max();
-
     GenericLexer lexer { input };
 
-    auto index = lexer.consume_while([](char ch) { return StringView { "0123456789" }.contains(ch); });
-
-    if (index.length() > 0)
-        specifier.index = parse_number(index);
+    if (!parse_number(lexer, specifier.index))
+        specifier.index = use_next_index;
 
     if (!lexer.consume_specific(':'))
         return lexer.is_eof();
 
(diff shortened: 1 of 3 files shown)


AK/Format.cpp
```diff
@@ -73,28 +73,32 @@ static void write_escaped_literal(StringBuilder& builder, StringView literal)
...
-    size_t value = 0;
-
-    for (char ch : input) {
-        value *= 10;
-        value += ch - '0';
+    value = 0;
+
+    bool consumed_at_least_one = false;
+    while (!lexer.is_eof()) {
+        if (StringView { "0123456789" }.contains(lexer.peek())) {
```
benit8, line 82:
```suggestion
        if (lexer.next_is(is_digit)) {
```
You can find ctype-like functions compatible with the GenericLexer under its definition in https://github.com/SerenityOS/serenity/blob/e8ef2b18580af33a5eb453a354310692da5371ff/AK/GenericLexer.h#L85

AK/Format.cpp
```diff
@@ -149,43 +166,133 @@ void vformat(const LogStream& stream, StringView fmtstr, Span<const TypeErasedPa
     stream << builder.to_string();
 }
 
-bool Formatter<StringView>::parse(StringView flags)
+void StandardFormatter::parse(StringView specifier)
 {
-    return flags.is_empty();
+    GenericLexer lexer { specifier };
+
+    if (StringView { "<^>" }.contains(lexer.peek(1))) {
```
benit8, line 173:
```suggestion
    if (lexer.next_is(is_any_of("<^>"))) {
```
You can find ctype-like functions compatible with the GenericLexer under its definition in https://github.com/SerenityOS/serenity/blob/e8ef2b18580af33a5eb453a354310692da5371ff/AK/GenericLexer.h#L85

benit8 (commented):
Is there any specific reason why you're not using the `next_is()` method from the `GenericLexer`? Especially the `next_is(Condition)` form; I wrote it because (to me) it felt more readable in the code, but I don't see it used when it could be. Perhaps I didn't add enough explanations in the GenericLexer definition?

AK/Format.cpp
benit8: Woops, just noticed you wanted to peek the 2nd character here

asynts:
> I wrote it because (to me) it felt more readable in the code, but I don't see it used when it could be.

@benit8 It's much nicer than what I did. I simply skimmed over the file to quickly and missed that there was some functional stuff going on.

I left the second suggestion unchanged because as you said, I look ahead one character. (Even though it would theoretically be possible to do the following, but I don't like the syntax:)

~~~diff
-if (StringView { "<^>" }.contains(lexer.peek(1))) {
+if (is_any_of("<^>")(lexer.peek(1)) {
~~~

PR #3601, opened 2020-09-25
LibJS+LibGUI+js: Handle UnterminatedRegexLiteral in syntax highlighters

linusg:
This token was added 2 months later than most others, explaining its absence in all the JS syntax highlighters (they should probably share the list of tokens for each style btw...)


linusg:
Hmm, probably not worth separate commits - let's squash those.

PR #3606, opened 2020-09-26
HackStudio: Detach from debugged process before terminating

### LINKED ISSUE #3308: HackStudio: Unkillable after debugging sample
Steps to reproduce:
1. Open HackStudio
2. Build the sample app: Build->Build
3. Set break point in the main function
4. Debug the sample app: Build->Debug
5. Continue the sample: Build->Run
6. Sample should have exited successfully
7. Close HackStudio: HackStudio->Quit
8. HackStudio won't terminate and can't be killed through System Monitor, either

Apparently it's stuck in `Queued` state somehow.

Screenshot from 2020-08-25 21-15-30

itamar8910:
Fixes #3308 and some issues with icon loading in the FormEditor.

~~Currently a draft, waiting on #3603 to get merged as it's required for this fix to work.~~

### DIFF
--- a/DevTools/HackStudio/Debugger/Debugger.cpp
+++ b/DevTools/HackStudio/Debugger/Debugger.cpp
@@ -58,8 +58,8 @@ Debugger::Debugger(
     , m_on_continue_callback(move(on_continue_callback))
     , m_on_exit_callback(move(on_exit_callback))
 {
-    pthread_mutex_init(&m_continue_mutex, nullptr);
-    pthread_cond_init(&m_continue_cond, nullptr);
+    pthread_mutex_init(&m_ui_action_mutex, nullptr);
+    pthread_cond_init(&m_ui_action_cond, nullptr);
 }
 
 void Debugger::on_breakpoint_change(const String& file, size_t line, BreakpointChange change_type)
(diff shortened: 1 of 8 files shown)


tomuta:
Hm somehow it feels like #3603 also should have addressed this. Which it didn't...

DevTools/HackStudio/Debugger/Debugger.cpp
```diff
@@ -259,4 +267,12 @@ void Debugger::insert_temporary_breakpoint(FlatPtr address)
     m_state.add_temporary_breakpoint(address);
 }
 
+void Debugger::set_requested_debugger_action(DebuggerAction action)
+{
+    pthread_mutex_lock(Debugger::the().continue_mutex());
+    Debugger::the().m_requested_debugger_action = action;
+    pthread_cond_signal(Debugger::the().continue_cond());
+    pthread_mutex_unlock(Debugger::the().continue_mutex());
```
awesomekling, line 275: Since we are in `Debugger` here already, no need to keep calling `Debugger::the()` :)

awesomekling: Also maybe `set_requested_action` is enough for the name, then we don't have to say "debugger" twice when calling it.

itamar8910:
> Hm somehow it feels like #3603 also should have addressed this. Which it didn't...

Yeah. Probably something to do with exiting the main thread while another thread is `wait`ing on something. I'll see if I can reproduce it in a small example.

tomuta:
Ok I think I know why #3603 isn't fixing it. We need to bring back a change that I removed at some point where we need to de-queue threads in certain situations (after calling `Thread::should_die`). This then would allow the threads to run again to clean up the kernel stack and then should be finalized.

tomuta:
I'm not saying these changes are bad, but SystemManager should always be able to kill HackStudio, whether it has a bug or not.

PR #3612, opened 2020-09-26
AK: Add a new backend for format, adding support for a ton more options.

asynts:
I admit, this makes the `<AK/PrintfImplementation.h>` header a mess. I'll have to clean this up later, by changing the printf implementation to use the new functions. (From what I can tell this should be possible, but needs some adjustments because the format implementation uses `StringBuilder` while the printf stuff caries a function pointer around.)

With this pull request, our format implementation supports all valid options for integers except the `c` type (cast integer to character.)


awesomekling:
Very nice progress :)

PR #3615, opened 2020-09-26
Kernel: Allow killing queued threads

tomuta:
We need to dequeue and wake threads that are waiting if the process
terminates.

Fixes #3603 without the HackStudio fixes in #3606.

There's still a separate remaining issue where the debugee (e.g. "little") doesn't get killed. I believe that's because if a tracer terminates, we don't resume/kill its traced processes.


awesomekling:
👍🏻

PR #3619, opened 2020-09-26
Kernel: Make Thread refcounted

tomuta:
Similar to Process, we need to make Thread refcounted. This will solve
problems that will appear once we schedule threads on more than one
processor. This allows us to hold onto threads without necessarily
holding the scheduler lock for the entire duration.

### DIFF
--- a/Kernel/Process.cpp
+++ b/Kernel/Process.cpp
@@ -265,7 +265,7 @@ void Process::kill_all_threads()
     });
 }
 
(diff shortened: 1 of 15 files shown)


Kernel/Process.cpp
```diff
@@ -556,6 +557,7 @@ siginfo_t Process::reap(Process& process)
 #endif
     ASSERT(process.is_dead());
     g_processes->remove(&process);
+    process.m_next = process.m_prev = nullptr; // mark it as reaped
```
awesomekling, line 560: Wait, `InlineLinkedList` doesn't clear out the next/prev pointers when removing something?
Let's fix that in `InlineLinkedList` instead of patching the pointers here :)

PR #3620, opened 2020-09-27
ref changed to Toolchain/BuildIt.sh

pfeilbr:
Toolchain/UseIt.sh ref changed to Toolchain/BuildIt.sh

### DIFF
--- a/Toolchain/CMakeToolchain.txt
+++ b/Toolchain/CMakeToolchain.txt
@@ -1,7 +1,7 @@
 set(CMAKE_SYSTEM_NAME Generic)
 
 if (NOT DEFINED ENV{SERENITY_ROOT})
-    message(FATAL_ERROR "SERENITY_ROOT not set. Please source Toolchain/UseIt.sh.")
+    message(FATAL_ERROR "SERENITY_ROOT not set. Please source Toolchain/BuildIt.sh.")
 endif()
 
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fno-rtti -fno-exceptions")


pfeilbr:
Toolchain/UseIt.sh ref changed to Toolchain/BuildIt.sh

Toolchain/CMakeToolchain.txt
```diff
@@ -1,7 +1,7 @@
 set(CMAKE_SYSTEM_NAME Generic)
 
 if (NOT DEFINED ENV{SERENITY_ROOT})
-    message(FATAL_ERROR "SERENITY_ROOT not set. Please source Toolchain/UseIt.sh.")
+    message(FATAL_ERROR "SERENITY_ROOT not set. Please source Toolchain/BuildIt.sh.")
```
awesomekling, line 4: Sourcing `BuildIt.sh` will not export `SERENITY_ROOT`, it will try to build a toolchain :)

This error message is outdated, but it should tell the user to export `SERENITY_ROOT` in his environment.
Or.. is this supposed to be done by the build system? cc @bugaevc

bugaevc: The Serenity build system itself doesn't need `SERENITY_ROOT` at all, we just rely on `CMAKE_SOURCE_DIR`. But some scripts need it, and the build system provides it automatically:

https://github.com/SerenityOS/serenity/blob/0f5214447790afa11907f57a2dcac70cb64655a7/CMakeLists.txt#L23-L32

And the ports build system needs it, but it tries to infer it from the location of the script itself:

https://github.com/SerenityOS/serenity/blob/0f5214447790afa11907f57a2dcac70cb64655a7/Ports/.port_include.sh#L5

So there should never be a need for the user to set `SERENITY_ROOT` explicitly.

pfeilbr: sorry for original mistake. ... and thanks for the gained learnings 😀

triggered good/discussion.  all tasks being self contained with relative deps without the need for env variable hint/bootstraper is one signal/validation of good project structure 😀

PR #3621, opened 2020-09-27
Spreadsheet: A few more functions

alimpfard:
This PR adds `lookup` (which, according to [Excel's docs](https://support.microsoft.com/en-us/office/excel-functions-by-category-5f91f4e9-7b42-46d2-9bd1-63f26a86c0eb), is one of the top ten most used functions in excel...huh), and its indirect friend `reflookup` without its weird parts.

TIL that excel (and whatever copies it) treats `A1` as a relative address, I had a blast finding ways to make a better interface than "imagine the topleft cell is here".

Achievement unlocked: Worst DBMS on the planet!
Be amazed (the conditional formatting is there to make it look better than it actually is):
shot-2020-09-27_05-28-21

### DIFF
--- a/Base/res/js/Spreadsheet/runtime.js
+++ b/Base/res/js/Spreadsheet/runtime.js
@@ -267,6 +267,72 @@ function here() {
     return new Position(position.column, position.row, thisSheet);
 }
 
+function internal_lookup(
+    req_lookup_value,
+    lookup_inputs,
+    lookup_outputs,
+    if_missing,
+    mode,
+    reference
+) {
+    lookup_outputs = lookup_outputs ?? lookup_inputs;
+
+    if (lookup_inputs.length > lookup_outputs.length)
+        throw new Error(
+            `Uneven lengths in outputs and inputs: ${lookup_inputs.length} > ${lookup_outputs.length}`
+        );
+
+    let references = lookup_outputs;
+    lookup_inputs = resolve(lookup_inputs);
+    lookup_outputs = resolve(lookup_outputs);
+    if_missing = if_missing ?? undefined;
+    mode = mode ?? "exact";
+    const lookup_value = req_lookup_value;
+    let matches = null;
+
+    if (mode === "exact") {
+        matches = value => value === lookup_value;
+    } else if (mode === "nextlargest") {
+        matches = value => value >= lookup_value;
+    } else if (mode === "nextsmallest") {
+        matches = value => value <= lookup_value;
+    } else {
+        throw new Error(`Match mode '${mode}' not supported`);
+    }
+
+    let retval = if_missing;
+    for (let i = 0; i < lookup_inputs.length; ++i) {
+        if (matches(lookup_inputs[i])) {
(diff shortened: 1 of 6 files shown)


Base/res/js/Spreadsheet/runtime.js
```diff
@@ -1,8 +1,79 @@
...
+        this.row = row;
+        this.sheet = sheet ?? thisSheet;
+        this.name = column + "" + row;
+    }
+
+    static from_name(name) {
+        let sheet = thisSheet;
+        let obj = sheet.parse_cell_name(name);
+        return new Position(obj.column, obj.row, sheet);
+    }
```
linusg, line 14: Honestly blows my mind that LibJS supports all of this natively!

Base/res/js/Spreadsheet/runtime.js
```diff
@@ -1,8 +1,79 @@
+// FIXME: Figure out a way to document non-function entities too.
+class Position {
+    constructor(column, row, sheet) {
+        this.column = column;
+        this.row = row;
+        this.sheet = sheet ?? thisSheet;
+        this.name = column + "" + row;
```
linusg, line 7: I prefer this, but it's up to you :)

```suggestion
        this.name = `${column}${row}`;
```

Base/res/js/Spreadsheet/runtime.js
```diff
@@ -1,8 +1,79 @@
...
+    }
+}
+
 function range(start, end, columnStep, rowStep) {
     columnStep = integer(columnStep ?? 1);
     rowStep = integer(rowStep ?? 1);
-    start = parse_cell_name(start) ?? { column: "A", row: 0 };
-    end = parse_cell_name(end) ?? start;
+    if (!(start instanceof Position)) start = parse_cell_name(start) ?? { column: "A", row: 0 };
+    if (!(end instanceof Position)) end = parse_cell_name(end) ?? start;
```
linusg, line 76: Thanks for using prettier lol, but I hate that it does this. Feel free to wrap the if body in {} to get a line break

Base/res/js/Spreadsheet/runtime.js
alimpfard: Very true!
I was honestly surprised when I saw nothing blowing up after adding this class.

Base/res/js/Spreadsheet/runtime.js
alimpfard: me two lol

awesomekling:
What is the purpose of the "Copy Strings out of the JS heap" commit? Those strings are refcounted so they will not be destroyed by GC.

alimpfard:
I thought so too, but I did hit a crash there where a `String` was trying to `ref()` a deleted `StringImpl`, and that copy made it disappear.

alimpfard:
I tested a bunch and that crash didn't come up, I'm gonna drop that commit and assume that everything is a-ok.

alimpfard:
Ah lovely, test-js strokes again (hehe bad pun): https://travis-ci.com/github/SerenityOS/serenity/builds/186992058

PR #3622, opened 2020-09-27
HackStudio: C++ Language Server

itamar8910:
This adds a C++ language server to HackStudio, inspired by the [LSP](https://microsoft.github.io/language-server-protocol/).

HackStudio notifies the language server when C++ source files are edited, and goes through it to get autocomplete suggestions.

There's lots of room for exciting functionality here, and also the possibility for language servers for other languages! :blush:


alimpfard:
Oh fancy.
Look, it's an excuse to make an incremental parser

itamar8910:
Looks like the buid fails because of a segfault in `test-js`?

DevTools/HackStudio/LanguageServers/Cpp/main.cpp
```diff
@@ -0,0 +1,81 @@
...
+#include <AK/LexicalPath.h>
+#include <DevTools/HackStudio/LanguageServers/Cpp/ClientConnection.h>
+#include <LibCore/EventLoop.h>
+#include <LibCore/File.h>
+#include <LibCore/LocalServer.h>
+#include <LibIPC/ClientConnection.h>
+#include <sys/stat.h>
+#include <unistd.h>
+
+static const char* local_socket_path = "/tmp/cpp-language";
```
awesomekling, line 36: This makes the server global which seems un-optimal if it's only intended to service a specific instance of HackStudio :)

tomuta: > This makes the server global which seems un-optimal if it's only intended to service a specific instance of HackStudio :)

I would expect multiple HackStudio instances with different projects to work with this architecture? And of course, global seems over-kill, I don't think it should be running at all times.

DevTools/HackStudio/LanguageServers/Cpp/AutoComplete.h
```diff
@@ -31,10 +31,12 @@
 #include <LibGUI/CppLexer.h>
 #include <LibGUI/TextPosition.h>
 
-namespace HackStudio {
-class CppAutoComplete {
+namespace LanguageServers {
+namespace Cpp {
```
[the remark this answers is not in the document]
itamar8910: oh, that's neat

DevTools/HackStudio/LanguageServers/Cpp/main.cpp
itamar8910: Yeah, one server could technically support connections from multiple instances of HackStudio, but it's definitely a better idea to have a separate instance of the language server per HackStudio instance (e.g if the language server crashes).
I changed the local socket path of the language server to be /tmp/language-servers/cpp/\<pid of hackstudio\>.

awesomekling: That path is racy since anybody could guess the PID and create a socket before you, intercepting your connection.
Paging @bugaevc, what do you think is a good approach for this sort of thing? Putting sockets in the home directory? Maybe it would be better to pass a file descriptor when forking off the language server and do the IPC over a pipe instead?

bugaevc: > /tmp/language-servers/cpp/<pid of hackstudio>

No no no, please don't do that sort of thing.

Why don't we just use SystemServer MultiInstance services for this again?

alimpfard: The only thing I can think of against using SystemServer would be that "this isn't a system-wide server".
imagine having to make a new entry for every language server we add, or the user decides to add.

alimpfard: Perhaps we should have a separate userland config file for system server?
it would be rw for some user (or all users?) and HackStudio could just register its language servers with that

bugaevc: > The only thing I can think of against using SystemServer would be that "this isn't a system-wide server".

But it is — in the same way that WebContent, ProtocolServer, and Clipboard are.

I get where you're coming from though; it would sure be nice to separate "actually system services" from "whatever the user wants to run on their own" — we could have per-user SystemServer configs, or even per-user SystemServer's, like launchd and systemd do.

But what I want to emphasize is that we've kind of punted on this distinction (system-wide vs user-specific) for a long time, and fixing/untangling that (if we even want to go there) is going to take way more than just the C++ language server. Or in other words, the language server is not any different from the other servers in this regard.

itamar8910: > Why don't we just use SystemServer MultiInstance services for this again?

That sounds good yeah, a service with `Lazy` and `MultiInstance` seems to be what we need here.

Having a per-user SystemServer config directory sounds really nice.
I think it would be out of the scope of this PR though, so for now I suggest we go with putting the C++ language server service in the global `SystemServer.ini`.

bugaevc: > I think it would be out of the scope of this PR though, so for now I suggest we go with putting the C++ language server service in the global `SystemServer.ini`.

Of course, I wasn't suggesting that needs to be done in the PR (or any time soon).

itamar8910: Okay, I pushed the changes. I'm marking this as resolved, but ofc feel free to reopen this conversation.

Libraries/LibCore/File.cpp
```diff
@@ -132,6 +134,30 @@ String File::real_path_for(const String& filename)
     return real_path;
 }
 
+bool File::ensure_parent_directories(const String& path)
+{
+    ASSERT(path[0] == '/');
```
bugaevc, line 139:
```suggestion
    ASSERT(path.starts_with('/'));
```

Libraries/LibCore/File.cpp
```diff
@@ -132,6 +134,30 @@ String File::real_path_for(const String& filename)
...
+
+    char* parent_buffer = strdup(path.characters());
+    const char* parent = dirname(parent_buffer);
+
+    int rc = 0;
+    while (true) {
+        int rc = mkdir(parent, 0755);
```
bugaevc, line 146: You're shadowing the `rc` from outer scope; so the outer scope `rc` will never get set to anything but 0.

Libraries/LibCore/File.cpp
```diff
@@ -132,6 +134,30 @@ String File::real_path_for(const String& filename)
...
+            break;
+
+        if (errno != ENOENT)
+            break;
+
+        ensure_parent_directories(parent);
+    };
+
+    free(parent_buffer);
+    return rc == 0;
```
bugaevc, line 158: By this time, `errno` is going to be changed by the `free()` call.

bugaevc:
Looks like all these were my fault all along? Well, please fix it nevertheless :slightly_smiling_face:

DevTools/HackStudio/LanguageClients/Cpp/CMakeLists.txt
```diff
@@ -0,0 +1,7 @@
+
+set(SOURCES
+    ServerConnection.cpp
+    ../../LanguageServers/Cpp/CppLanguageServerEndpoint.h
+    ../../LanguageServers/Cpp/CppLanguageClientEndpoint.h
```
bugaevc, line 5: You can't do that. Use `GENERATED_SOURCES` like this:
https://github.com/SerenityOS/serenity/blob/450a2a0f9ce2443bc06445ee018c9860d8cd5fc8/Libraries/LibAudio/CMakeLists.txt#L7-L10

Don't ask me why if you want to keep your sanity :)

DevTools/HackStudio/LanguageClients/Cpp/ServerConnection.cpp
```diff
@@ -0,0 +1,27 @@
...
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include "ServerConnection.h"
```
bugaevc, line 27: That's... weird?

DevTools/HackStudio/LanguageServers/Cpp/CMakeLists.txt
```diff
@@ -0,0 +1,14 @@
...
+set(SOURCES
+    ClientConnection.cpp
+    main.cpp
+    CppLanguageServerEndpoint.h
+    CppLanguageClientEndpoint.h
+    AutoComplete.cpp
+)
+
+serenity_bin(CppLanguageServer)
+target_link_libraries(CppLanguageServer LibIPC LibGUI)
```
bugaevc, line 14: How come a language server needs LibGUI? 😨

DevTools/HackStudio/LanguageServers/Cpp/ClientConnection.cpp
```diff
@@ -0,0 +1,185 @@
...
+#include <LibGUI/TextDocument.h>
+
+// #define DEBUG_CPP_LANGUAGE_SERVER
+// #define DEBUG_FILE_CONTENT
+
+namespace LanguageServers::Cpp {
+
+static HashMap<int, RefPtr<ClientConnection>> s_connections;
+
+ClientConnection::ClientConnection(NonnullRefPtr<Core::LocalSocket> socket, int client_id)
```
bugaevc, line 40: Is there a reason for it to take a `NonnullRefPtr<Core::LocalSocket>` and not a `Core::LocalSocket&`?

Libraries/LibCore/File.cpp
itamar8910: Any why does it matter? `rc` only tracks the return code of `mkdir`.

DevTools/HackStudio/LanguageServers/Cpp/ClientConnection.cpp
```diff
@@ -0,0 +1,185 @@
...
+void ClientConnection::handle(const Messages::CppLanguageServer::FileOpened& message)
+{
+    LexicalPath file_path(String::format("%s/%s", m_project_root.string().characters(), message.file_name().characters()));
+#ifdef DEBUG_CPP_LANGUAGE_SERVER
+    dbg() << "FileOpened: " << file_path.string();
+#endif
+
+    auto file = Core::File::construct(file_path.string());
+    if (!file->open(Core::IODevice::ReadOnly)) {
+        perror("open");
```
bugaevc, line 91: Does `Core::File::open()` set `errno`? I thought it saved the error to a member int that you had to get explicitly...

Libraries/LibCore/File.cpp
bugaevc: `rc` yes, but `errno` tracks the last call. If `mkdir()` fails, you'd return `false` form here, and then the caller would see that and use something like `perror("ensure_parent_directories")`, but would get the wrong error message.

Base/etc/SystemServer.ini
```diff
@@ -141,3 +141,11 @@ StdIO=/dev/tty1
 Environment=TERM=xterm
 KeepAlive=1
 BootModes=text
+
+[CppLanguageServer]
+Socket=/tmp/portal/cpp
```
bugaevc, line 146: If you're emphasizing that this is just the first of language servers (I'd anticipate a JS language server in the future, amirite?), why not put it in `/tmp/portal/language/cpp`?

DevTools/HackStudio/LanguageClients/CMakeLists.txt
```diff
@@ -0,0 +1,2 @@
+
```
bugaevc, line 1: Why leave all these empty lines at the start, in this file and elsewhere?

Libraries/LibCore/File.cpp
itamar8910: Do we expect `perror` to work with non-`LibC` functions?
For example, it would make using `AK` containers in LibCore functions problematic since they could call `free` in their destructor.

DevTools/HackStudio/LanguageServers/Cpp/CMakeLists.txt
itamar8910: Yeah it is a weird dependency - It's because we need `GUI::CppLexer`.
If we want to not depend on LibGUI here we could either:
1. Create `LibCpp`, similarly to the way we have the javascript lexer in `LibJS`. 
2. Move the c++ lexer to the language server itself, and make `GUI::SyntaxHighlighter` able to go through the c++ language server for lexing.
It would mean that we'd have to use the c++ language server from the TextEditor as well however, which seems wrong.

Libraries/LibCore/File.cpp
bugaevc: That depends on the API in question, but generally, if it's fallible, it should use one of the following styles:
* return `-1`/`nullptr`/`false` and set `errno`
* do the same and store the error somewhere inside the object; this is what `Core::IODevice` and stdio `FILE` do:
    https://github.com/SerenityOS/serenity/blob/9225bfa95e354dc12c6705d65ff33224095a8917/Libraries/LibCore/IODevice.h#L103
    https://github.com/SerenityOS/serenity/blob/9225bfa95e354dc12c6705d65ff33224095a8917/Libraries/LibC/stdio.cpp#L135
* return a `Result<>`

`perror()` only works with `errno`, so it's only suitable after a call to an API that sets `errno` (though you can set `errno` yourself if you're using an API that gives you the error in a different way).

DevTools/HackStudio/LanguageServers/Cpp/CMakeLists.txt
bugaevc: If you decide to continue linking to LibGUI, I would at least expect a comment here explaining the same thing you just said :slightly_smiling_face:

Base/etc/SystemServer.ini
alimpfard: I dunno about js, but you're gonna get a shell language server whether you want it or not :P

Libraries/LibCore/File.cpp
itamar8910: Okay, `ensure_parent_directories` should now return with an errno that reflects the reason for the failure. Let me know what you think.

DevTools/HackStudio/CMakeLists.txt
```diff
@@ -1,7 +1,10 @@
+
```
bugaevc, line 1: Extra blank line here too

DevTools/HackStudio/LanguageClients/Cpp/CMakeLists.txt
```diff
@@ -0,0 +1,9 @@
+
```
bugaevc, line 1: And here...

DevTools/HackStudio/LanguageServers/CMakeLists.txt
```diff
@@ -0,0 +1,2 @@
+
```
bugaevc, line 1: And here...

DevTools/HackStudio/LanguageServers/Cpp/CMakeLists.txt
```diff
@@ -0,0 +1,14 @@
+
```
bugaevc, line 1: And here...

Libraries/LibCore/File.cpp
bugaevc: This looks pretty good, assuming C++ guarantees that the destruction order is the reverse of the creation order.

DevTools/HackStudio/LanguageServers/Cpp/CMakeLists.txt
itamar8910: Actually there's another dependency we need from LibGUI, the `GUI::TextDocument`.
The language server uses it to update the content of a file when it receives updates over IPC on edit actions the user performs.
We could extract document-related logic to something like `LibDocument`, but it's a pretty big refactoring that I'll leave for a later time.
I ended up extracting the lexer to `LibCpp`, though.

DevTools/HackStudio/LanguageServers/Cpp/ClientConnection.cpp
itamar8910: Well, not really but it seems to be the convention among other `IPC::ClientConnection` subclasses from what I've seen.
I mean, the constructor of `IPC::ClientConnection` itself takes a `NonnullRefPtr<Core::LocalSocket>`.

bugaevc: Right...

Well, should we change that? @awesomekling?

awesomekling: Hmm, IIRC there was some reason behind this. I think it was because WebContent `main()` does this:

```cpp
    auto socket = Core::LocalSocket::take_over_accepted_socket_from_system_server();
    ASSERT(socket);
    IPC::new_client_connection<WebContent::ClientConnection>(socket.release_nonnull(), 1); 
    return event_loop.exec();
```

The constructor is invoked via forwarding from `new_client_connection`.
If we take the socket by raw reference, we now hold on to it in `main()` which means we have take extra care to unref the socket, or it will leak.

awesomekling: Or, not "leak" in the strict sense, but it will be forever kept alive by the stack variable in `main()`.

bugaevc: Well yeah, `main()` should just unref it if it doesn't want to keep the reference. It's not hard to do, we should just use a nested scope, or am I missing something?

```cpp
    {
        auto socket = Core::LocalSocket::take_over_accepted_socket_from_system_server();
        ASSERT(socket);
        IPC::new_client_connection<WebContent::ClientConnection>(socket.release_nonnull(), 1); 
    }
    return event_loop.exec();
```

itamar8910:
@bugaevc Thanks for reviewing this. Do you have any more comments?

bugaevc:
I'm going to re-review it sometime later, then we'll know 🙂

DevTools/HackStudio/EditorWrapper.h
```diff
@@ -50,7 +50,7 @@ class EditorWrapper : public GUI::Widget {
     void set_editor_has_focus(Badge<Editor>, bool);
 
 private:
-    explicit EditorWrapper(BreakpointChangeCallback);
+    explicit EditorWrapper();
```
bugaevc, line 53: Minor nit: no need for `explicit` here anymore

DevTools/HackStudio/LanguageServers/Cpp/ClientConnection.cpp
```diff
@@ -0,0 +1,186 @@
...
+}
+
+ClientConnection::~ClientConnection()
+{
+}
+
+void ClientConnection::die()
+{
+    s_connections.remove(client_id());
+    exit(0);
```
bugaevc, line 53: Hmm, not important, but I'd prefer the service to exit with an error in case the client disconnected abruptly (including when the client itself dies with an error). We can always do that later though; let's not block this PR on that.

DevTools/HackStudio/LanguageServers/Cpp/CppLanguageServer.ipc
```diff
@@ -0,0 +1,11 @@
+endpoint CppLanguageServer = 8001
+{
+    Greet(String project_root) => (i32 client_id)
+
+    FileOpened(String file_name) =|
+    FileEditInsertText(String file_name, String text, i32 start_line, i32 start_column) =|
+    FileEditRemoveText(String file_name, i32 start_line, i32 start_column, i32 end_line, i32 end_column) =|
+    SetFileContent(String file_name, String content) =|
+
+    AutoCompleteSuggestions(String file_name, i32 cursor_line, i32 cursor_column) => (Vector<String> suggestions)
```
bugaevc, line 10: I hope we're not calling this synchronously on the UI thread? 😨

DevTools/HackStudio/LanguageServers/Cpp/main.cpp
```diff
@@ -0,0 +1,51 @@
...
+{
+    Core::EventLoop event_loop;
+    if (pledge("stdio shared_buffer unix rpath", nullptr) < 0) {
+        perror("pledge");
+        return 1;
+    }
+
+    auto socket = Core::LocalSocket::take_over_accepted_socket_from_system_server();
+    IPC::new_client_connection<LanguageServers::Cpp::ClientConnection>(socket.release_nonnull(), 1);
+    if (pledge("stdio shared_buffer rpath", nullptr) < 0) {
```
bugaevc, line 46: Perhaps a stupid question, but what does it need shared buffers for?

bugaevc:
One other thing I would do differently is instead of passing a file pathand having the server open it, I'd pass an open file descriptor, now that we have the kernel support for that. (Though it probably needs LibIPC support to be nice? Or is this done already?)

Then we could remove `rparh`, and that'd be a major win for sandboxing. Parsers of any sort, especially parsers of something as complex as C++, are not to be trusted with complete read access to everything on the system :)

DevTools/HackStudio/LanguageServers/Cpp/main.cpp
itamar8910: Not at all a stupid question, it does not need this permission.

DevTools/HackStudio/LanguageServers/Cpp/CppLanguageServer.ipc
itamar8910: Oh, we definitely are :stuck_out_tongue_closed_eyes: 
Currently auto-completion is very simple and does only lexing, so there isn't any noticeable delay in the UI thread (I mean, we perform lexing on the program every time we edit).
Of course once it starts doing more complex stuff, we would not want to call this synchronously from the UI thread.
We could use server->client IPC messages for this, and then `AutoCompleteSuggestions` could be asynchronous and report the results back to the client when we have them.
I think having this call synchronous for now is OK to get the language server going, but I added a `FIXME` in `ClientConnection.cpp` about this (I don't think our .ipc compiler supports comment?),

bugaevc: Yeah, I guess we can get away with that for now (*somehow*, everything is quite fast for us to do on the UI thread — perhaps because we don't do animations and don't use the system under load a lot). But on other systems, it'd be a non-starter to do *any* blocking I/O on the UI thread, let alone RPCing into a remote process.

itamar8910:
> One other thing I would do differently is instead of passing a file path and having the server open it, I'd pass an open file descriptor

Yeah that would be very good, an (almost) completely sandboxed language server sounds great.

> (Though it probably needs LibIPC support to be nice? Or is this done already?)

I don't think that `sendfd` and `recvfd` are currently supported in the IPC system (@awesomekling  do you know otherwise?)
Just out of interest (because we definitely should have support for this in LibIPC) what would be the not-nice way of sending `fd`s over IPC?

awesomekling:
> > (Though it probably needs LibIPC support to be nice? Or is this done already?)
> 
> I don't think that `sendfd` and `recvfd` are currently supported in the IPC system (@awesomekling do you know otherwise?)
> Just out of interest (because we definitely should have support for this in LibIPC) what would be the not-nice way of sending `fd`s over IPC?

LibIPC currently does not have any high-level support for `sendfd` and `recvfd` at this moment. However, you could easily open an fd on one side and send it as an `i32` to the other side. In fact, I'm not really sure what a high level mechanism here would look like..

bugaevc:
> Just out of interest (because we definitely should have support for this in LibIPC) what would be the not-nice way of sending `fd`s over IPC?

Just call `sendfd()`/recvfd()` along with making/receiving a request that logically carries an fd 🙃

bugaevc:
> However, you could easily open an fd on one side and send it as an `i32` to the other side.

No, there's no point in sending the integer value, that's useless for the peer :) you don't have to send *anything* inline, but you do have to `sendfd()`/`recvfd()`. That would be what LibIPC would do on the low level as well.

> In fact, I'm not really sure what a high level mechanism here would look like..

Like this:

```
OpenFile(fd file) =|
```

Which translates to `int file_fd` on both sides, but actually passes an fd using `sendfd()`/`recvfd()` instead of an inline integer. Or, we could have some C++y wrapper type over a raw fd.

alimpfard:
I think I'm to blame for those conflicts, sowwy~

bugaevc:
@itamar8910 yeah, this looks good to me :)

itamar8910:
> I think I'm to blame for those conflicts, sowwy~

no worries, fixed them.

DevTools/HackStudio/Editor.cpp
```diff
@@ -462,6 +468,25 @@ void Editor::set_document(GUI::TextDocument& doc)
 {
     ASSERT(doc.is_code_document());
     GUI::TextEditor::set_document(doc);
+
+    CodeDocument& code_document = static_cast<CodeDocument&>(doc);
+    switch (code_document.language()) {
+    case Language::Cpp:
+        set_syntax_highlighter(make<GUI::CppSyntaxHighlighter>());
+        cpp_Language_server_connection().post_message(Messages::CppLanguageServer::FileOpened(code_document.file_path().string()));
```
alimpfard, line 476: @itamar8910 
I didn't really have the time to look at these until today, but...this is going to cause a rather big amount of duplication (also that it forces HackStudio to know about the different LSPs' it connects to statically, not great)

Take a look at https://github.com/alimpfard/serenity/commit/8837d87080571e2534ef063c6aaa1a6125fd6bbd for a preview of the duplication.
These languages servers should probably:
- have a single IPC definition
- have a base class (at least on the HackStudio side

itamar8910: Yeah I agree 100%, it wouldn't be good to have this much code duplication when using language servers.
Feel free to do this architecture change in your branch (or we could do it later which is also OK).

PR #3624, opened 2020-09-27
remove call to action comment to define SERENITY_ROOT

pfeilbr:
remove call to action comment to define SERENITY_ROOT


pfeilbr:
correct fix for #3620

PR #3625, opened 2020-09-27
AK: Add support for replacement fields to format.

asynts:
Now it's possible to specify the width with which numbers are formatted dynamically:

~~~c++
String::formatted("{:0{1}}", 1, 3) // 001
~~~

In the `std::format` version it is also possible to take it from the next format specifier implicitly, I do not support this syntax yet:

~~~c++
String::formatted("{:0{}}", 1, 3) // 001
~~~

---

In the future it might be smart to define the repetitive type information stuff using `ENUMERATE_BUILTIN_FORMATTER` macros, but I did not want to do this yet.

It also starts to get quite crowded in the header and implementation. I'll have to do some housekeeping soon.

### DIFF
--- a/AK/Format.cpp
+++ b/AK/Format.cpp
@@ -33,39 +33,111 @@
 
 namespace {
 
+constexpr size_t use_next_index = NumericLimits<size_t>::max();
+
 struct FormatSpecifier {
     StringView flags;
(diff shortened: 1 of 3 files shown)


awesomekling:
Oh whoops, this has conflicts

PR #3628, opened 2020-09-28
Shell+LibGUI: Syntax highlighting in text editors

alimpfard:
Title says all.

Draft for the time being, until I squash some bugs in it :P

### DIFF
--- a/Shell/AST.h
+++ b/Shell/AST.h
@@ -47,6 +47,16 @@ struct HighlightMetadata {
 struct Position {
     size_t start_offset { 0 };
     size_t end_offset { 0 };
+    struct Line {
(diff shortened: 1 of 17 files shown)


Shell/AST.h
```diff
@@ -47,6 +47,13 @@ struct HighlightMetadata {
 struct Position {
     size_t start_offset { 0 };
     size_t end_offset { 0 };
+    struct Line {
+        size_t line_number { 0 };
+        size_t line_column { 0 };
+
+        bool operator==(const Line&) const = default;
```
alimpfard, line 54: me: c++20 provides default-able comparison operators, nice
travis: heh, fool! as if I'd let you!

PR #3629, opened 2020-09-28
AK+Format: Support default index in replacement field.

asynts:
The following does now work:

~~~c++
outf("{:0{}}", 1, 3); // 001
~~~

(The nested braces tell the formatter to take the field width from the next argument.) Previously, it was necessary to specify the index explicitly.

### DIFF
--- a/AK/Format.cpp
+++ b/AK/Format.cpp
@@ -128,7 +128,7 @@ public:
             return false;
 
         if (!consume_number(index))
-            return use_next_index;
+            index = use_next_index;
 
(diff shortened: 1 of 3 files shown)


asynts:
Yay! The segmentation fault happened again.

https://travis-ci.com/github/SerenityOS/serenity/builds/186992145#L3167-L3169

awesomekling:
Yeah that segfault happens intermittently on Travis. Trying to track it down locally.

PR #3640, opened 2020-09-29
AK+Format: Support remaining format specifiers for integers and strings.

asynts:
With this pull request it's now possible to use align, width and precision specifiers with strings as well. I also added support for character casts, where other integer types are printed as character (just ASCII, nothing fancy.)

---

The code used for alignment starts to get really redundant, I'll need to clean this up. This is what I am planning to do next.


alimpfard:
I can finally clean up that alignment thing in `LibMarkdown/Table.cpp` with this!

PR #3642, opened 2020-09-29
LibWeb: Add cm, in, mm and Q CSS units

Lubrsi:
For example, this properly positions the boxes in this margin test: https://test.csswg.org/suites/css21_dev/20110323/html4/margin-001.htm

vmplayer_Q9yy2PNPrI

For some reason, the margin-right is way off, off by 116 pixels. Using margin-right: -116px on other browsers on the div2 element produces the same result.


Lubrsi:
Build failure seems to be the test-js segfault. Wish valgrind found it already, but no luck yet.

PR #3646, opened 2020-09-30
Shell: Wait for *a* child to change state when receiving a SIGCHLD

### LINKED ISSUE #3645: Shell: block_on_job() never unblocks when run from TelnetServer
Spinoff of #3637.
Repro:
```sh
# Easiest way to get telnet server running (under a new libline profile)
$ su
# Get a libline config for root, proof that this is not #3637 
$ mkdir -p ~/.config/lib
$ touch ~/.config/lib/line.ini
# Make libline use getline()
$ ini ~/.config/lib/line.ini behaviour operation_mode NoEscapeSequences

$ TelnetServer
# From host
$ telnet localhost 8823
# Run something, it will execute and wait forever
```

Observations from #3637: https://github.com/SerenityOS/serenity/issues/3637#issuecomment-701117842

alimpfard:
This really just works around the core issue, which is that we have no
reliable way to know exactly who raised the signal (yet).
Fixes #3645, in a very weird way.

This is not a particularly good way to deal with the problem, but it *does* fix the aforementioned issue for me.

cc @awesomekling - would be nice to get `SA_SIGINFO`, even if just `si_pid` is present.

Draft for now, I don't like this solution much.

### DIFF
--- a/Shell/main.cpp
+++ b/Shell/main.cpp
@@ -91,44 +91,71 @@ SavedFileDescriptors::~SavedFileDescriptors()
 void Shell::setup_signals()
 {
     Core::EventLoop::register_signal(SIGCHLD, [](int) {
+#ifdef SH_DEBUG
+        dbg() << "SIGCHLD!";
+#endif
         auto& jobs = s_shell->jobs;
         Vector<u64> disowned_jobs;
-        for (auto& it : jobs) {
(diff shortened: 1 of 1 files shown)


alimpfard:
Nevermind, one more test proved that this does not fix the issue either.

PR #3647, opened 2020-09-30
Shell: Wait for *a* child to change state when receiving a SIGCHLD

### LINKED ISSUE #3645: Shell: block_on_job() never unblocks when run from TelnetServer

alimpfard:
(take 2 - apparently you can't reopen PRs when you force-push to the branch)

This really just works around the core issue, which is that we have no
reliable way to know exactly who raised the signal (yet).
Fixes #3645, in a very weird way - although seems like bash also does something very similar (unless I'm just incapable of reading code with shit indentation) in its job control.

This is not a particularly good way to deal with the problem, but it does fix the aforementioned issue for me.

cc @awesomekling - would be nice to get `SA_SIGINFO`, even if just `si_pid` is present.

~Draft for now, I don't like this solution at all~ I can't really think of a better solution short of using `si_pid` from sigaction, but we don't have that.

### DIFF
--- a/Shell/Shell.cpp
+++ b/Shell/Shell.cpp
@@ -60,43 +60,10 @@ extern char** environ;
 void Shell::setup_signals()
 {
     Core::EventLoop::register_signal(SIGCHLD, [this](int) {
-        Vector<u64> disowned_jobs;
-        for (auto& it : jobs) {
-            auto job_id = it.key;
-            auto& job = *it.value;
-            int wstatus = 0;
-            auto child_pid = waitpid(job.pid(), &wstatus, WNOHANG | WUNTRACED);
-            if (child_pid < 0) {
-                if (errno == ECHILD) {
-                    // The child process went away before we could process its death, just assume it exited all ok.
-                    // FIXME: This should never happen, the child should stay around until we do the waitpid above.
-                    dbg() << "Child process gone, cannot get exit code for " << job_id;
-                    child_pid = job.pid();
-                } else {
-                    ASSERT_NOT_REACHED();
-                }
-            }
-#ifndef __serenity__
-            if (child_pid == 0) {
-                // Linux: if child didn't "change state", but existed.
-                continue;
-            }
+#ifdef SH_DEBUG
+        dbg() << "SIGCHLD!";
 #endif
-            if (child_pid == job.pid()) {
-                if (WIFSIGNALED(wstatus) && !WIFSTOPPED(wstatus)) {
-                    job.set_signalled(WTERMSIG(wstatus));
-                } else if (WIFEXITED(wstatus)) {
-                    job.set_has_exit(WEXITSTATUS(wstatus));
-                } else if (WIFSTOPPED(wstatus)) {
-                    job.unblock();
-                    job.set_is_suspended(true);
-                }
-            }
-            if (job.should_be_disowned())
-                disowned_jobs.append(job_id);
-        }
-        for (auto job_id : disowned_jobs)
-            jobs.remove(job_id);
+        notify_child_event();
     });
 
     Core::EventLoop::register_signal(SIGTSTP, [this](auto) {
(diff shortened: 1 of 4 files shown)


Shell/Shell.cpp
```diff
@@ -1359,6 +1360,69 @@ void Shell::custom_event(Core::CustomEvent& event)
...
+                disowned_jobs.append(job_id);
+        }
+        for (auto job_id : disowned_jobs)
+            jobs.remove(job_id);
+
+#ifndef SH_DEBUG
+        if (!found_child) {
+            dbg() << "DIDN'T FIND A CHILD?!?!?!?!?!?!?!?";
+            dbg() << "RETRY!";
+            Core::EventLoop::current().post_event(*this, make<Core::CustomEvent>(ShellEventType::ChildEvent));
```
alimpfard, line 1422: This spins for quite a while (some 100ms) before the child actually dies and shows up as dead through the `waitpid` above - possible finalisation bug? or maybe I'm just misunderstanding how this is supposed to work

tomuta: Waiting for *any* child is the correct approach. In absence of `SA_SIGINFO` there is just no way to know which child it was, so if you're doing a blocking `waitpid` on a particular PID you might end up deadlocking.

How long it may take, that depends. First, all threads need to be finalized, once all threads are finalized the process is finalized. All that assumes that the `FinalizerTask` had a chance to run. Whether that takes a little longer or not also depends on how the process was terminated, e.g. it may be necessary to briefly wake blocked threads to clean up the kernel stack before being able to flag it for finalization.

Shell/Shell.cpp
```diff
@@ -1359,6 +1360,69 @@ void Shell::custom_event(Core::CustomEvent& event)
...
+#endif
+            auto child_pid = waitpid(job.pid(), &wstatus, WNOHANG | WUNTRACED);
+#ifdef SH_DEBUG
+            dbgf("... = {} - {}", child_pid, wstatus);
+#endif
+
+            if (child_pid < 0) {
+                if (errno == ECHILD) {
+                    // The child process went away before we could process its death, just assume it exited all ok.
+                    // FIXME: This should never happen, the child should stay around until we do the waitpid above.
```
tomuta, line 1390: This is true, this should never happen, but it is only true for the first call to `waitpid`. Once `waitpid` is used on a terminated process, it is removed from the process table and subsequent calls will return `ECHILD`

Shell/Shell.cpp
```diff
@@ -1359,6 +1360,69 @@ void Shell::custom_event(Core::CustomEvent& event)
...
+        // This, of course, will mean that someone can send us a SIGCHILD and we'd be spinning here forever.
+        bool found_child = false;
+        for (auto& it : jobs) {
+            auto job_id = it.key;
+            auto& job = *it.value;
+            int wstatus = 0;
+#ifdef SH_DEBUG
+            dbgf("waitpid({}) = ...", job.pid());
+#endif
+            auto child_pid = waitpid(job.pid(), &wstatus, WNOHANG | WUNTRACED);
```
tomuta, line 1382: If you're waiting for any process in a process group, shouldn't it be `-job.pid()`?

Shell/Shell.cpp
alimpfard: > Waiting for any child is the correct approach

True, but that sounds really goofy, anyone can send us a `SIGCHLD`, and then we'd saturate the event loop waiting on nonexistent children.
Perhaps I should coalesce the events to at least limit the scope of this to "until the next child changes state" :thinking:

> Finalisation

I don't know the exact sequence of events here, but this signal is emitted in https://github.com/SerenityOS/serenity/blob/a9335eea1c96d50140e46b8e0361e01b87c09ce2/Kernel/Process.cpp#L622 - just before the process' memory regions are cleared and it's marked "dead".
so the finaliser _is_ run, but the process dies much later - or at least, that's the impression I'm getting.

tomuta: > True, but that sounds really goofy, anyone can send us a SIGCHLD, and then we'd saturate the event loop waiting on nonexistent children.
It's not really. `waitpid` will just return `ECHILD` in that case, telling you that there is no terminated child. So yes, the event loop will briefly execute to find a terminated child, but since it won't it should go back without doing anything else.

Shell/Shell.cpp
alimpfard: The entry in the job table should be removed if `waitpid` succeeds once though, I don't think this path would be taken twice for any one job

Shell/Shell.cpp
tomuta: > I don't know the exact sequence of events here, but this signal is emitted in
> 
> https://github.com/SerenityOS/serenity/blob/a9335eea1c96d50140e46b8e0361e01b87c09ce2/Kernel/Process.cpp#L622
> - just before the process' memory regions are cleared and it's marked "dead".
> so the finaliser _is_ run, but the process dies much later - or at least, that's the impression I'm getting.

Correct, this line is executed from `Process::finalize`, which is after all threads are finalized. Once finalized, its "shell" hangs around until you reap it with `waitpid`. Once reaped, it will be cleaned up entirely and can no longer be found with `waitpid`.

Shell/Shell.cpp
alimpfard: We're waiting for a specific process here, which should always be the process group leader.
Not sure why we'd want to look at the other members of a process group, am I missing something?

Shell/Shell.cpp
alimpfard: So then why the rather large delay between the signal being delivered and `waitpid` considering the process dead?

alimpfard:
Ah fantastic, a failing test!

Shell/Shell.cpp
tomuta: No I guess you're right. Sorry I'm not really familiar with all that ;-)

Shell/Shell.cpp
tomuta: Good question. I wonder if it's because `Thread::send_signal` doesn't seem to wake the `Thread` it's sending a signal to if `this != Thread::current()`.

tomuta: > Good question. I wonder if it's because `Thread::send_signal` doesn't seem to wake the `Thread` it's sending a signal to if `this != Thread::current()`.

`Scheduler::pick_next` does look at `Thread::has_unmasked_pending_signals`, but that probably needs to wait until the thread quantum (time slice) is depleted. But that shouldn't take 100ms.

Shell/Shell.cpp
tomuta: Ok, I guess I assumed it's somehow hitting this path. If it is then that would be a Kernel bug.

alimpfard: I'll add an assertion for it on the shell side (once I resolve this failing test).

alimpfard: We'll see if that assertion trips any time in the future :P

alimpfard:
_Theoretically_, test-js should not segfault.

alimpfard:
I suppose we can live with a 100ms delay in place of an infinity-ms delay in TelnetServer for the time being

PR #3648, opened 2020-09-30
AK+Format: Add formatters for boolean and pointer values and replace a few usages.

asynts:
This pull request adds formatters for boolean and pointer values which I somehow forgot about when creating #3640.

I also replaced almost all printf calls with format calls. There are a few missing, but they aren't trivially replaceable because `warnf` always appends a newline like `warn` or generally any `LogStream`. I am not 100% sure what is the best way to fix this so, I left it open for the future.

### DIFF
--- a/AK/Tests/TestFormat.cpp
+++ b/AK/Tests/TestFormat.cpp
@@ -147,4 +147,19 @@ TEST_CASE(boolean_values)
     EXPECT_EQ(String::formatted("{:#08x}", true), "0x000001");
 }
 
+TEST_CASE(pointers)
+{
+    void* ptr = reinterpret_cast<void*>(0x4000);
+
+    if (sizeof(void*) == 4) {
+        EXPECT_EQ(String::formatted("{:p}", 32), "0x00000020");
+        EXPECT_EQ(String::formatted("{:p}", ptr), "0x00004000");
+    } else if (sizeof(void*) == 8) {
+        EXPECT_EQ(String::formatted("{:p}", 32), "0x0000000000000020");
+        EXPECT_EQ(String::formatted("{:p}", ptr), "0x0000000000004000");
+    } else {
+        ASSERT_NOT_REACHED();
+    }
+}
+
 TEST_MAIN(Format)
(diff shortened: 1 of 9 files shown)


asynts:
This segmentation fault crash in `test-js` seems to get really common now.

AK/Tests/TestFormat.cpp
```diff
@@ -136,4 +136,15 @@ TEST_CASE(cast_integer_to_character)
...
 
+TEST_CASE(boolean_values)
+{
+    EXPECT_EQ(String::formatted("{}", true), "true");
+    EXPECT_EQ(String::formatted("{}", false), "false");
+    EXPECT_EQ(String::formatted("{:6}", true), "true  ");
+    EXPECT_EQ(String::formatted("{:>4}", false), "false");
+    EXPECT_EQ(String::formatted("{:d}", false), "0");
+    EXPECT_EQ(String::formatted("{:d}", true), "1");
+    EXPECT_EQ(String::formatted("{:#08x}", true), "0x000001");
```
awesomekling, line 147: Hmm, shouldn't this yield `"0x00000001"`?

asynts: [This is fmtlib's behavior exactly.][1]

I was thinking about this too, because when formatting numbers one is usually interested in the significant digits and not the total width. But fmtlib does considers the sign and the alternative form prefix in the width calculation.

In my mind it would probably be best to not consider it, because after all we have a "reserved" sign mode which puts a space for positive numbers, where the sign would go otherwise (e.g. `String::formatted("{: }", 1) == " 1"`.) This should be enough to ensure the exact width when required.

I am not completely certain if this is a good idea though, it would differ from how strings are formatted (since they never have any prefixes.)

  [1]: https://godbolt.org/z/noMEfn

asynts: On a second note, [this is printf's behavior too.][1]

  [1]: https://godbolt.org/z/8n5oMq

awesomekling: Huh, well okay :]

PR #3654, opened 2020-10-01
MenuApplets: Keymap Switcher

sppmacd:
This PR adds a new MenuApplet: "Keymap". It displays current keymap name (first 2 characters of it). If you click on it, it opens KeyboardSettings.

It's now updated every 5 seconds.

Screenshot:
Screenshot

Note: I think it will be better to handle keymaps in WindowServer instead of Kernel. We can then broadcast messages about keymap change (and then handle it properly e.g updating display in Keymap applet)

### DIFF
--- a/Kernel/Process.h
+++ b/Kernel/Process.h
@@ -319,6 +319,7 @@ public:
     int sys$realpath(Userspace<const Syscall::SC_realpath_params*>);
     ssize_t sys$getrandom(Userspace<void*>, size_t, unsigned int);
     int sys$setkeymap(Userspace<const Syscall::SC_setkeymap_params*>);
+    int sys$get_keymap_name(Userspace<char*> buffer, size_t buffer_size);
     int sys$module_load(Userspace<const char*> path, size_t path_length);
     int sys$module_unload(Userspace<const char*> name, size_t name_length);
     int sys$profiling_enable(pid_t);
--- a/MenuApplets/Keymap/main.cpp
+++ b/MenuApplets/Keymap/main.cpp
@@ -1,0 +1,139 @@
+/*
+ * Copyright (c) 2020, the SerenityOS developers.
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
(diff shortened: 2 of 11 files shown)


bugaevc:
> Note: I think it will be better to handle keymaps in WindowServer instead of Kernel. We can then broadcast messages about keymap change (and then handle it properly e.g updating display in Keymap applet)

Not arguing about where it's better to handle keymaps, but you could make the kernel notify the observers about keymap changes as well, instead of them having to poll the kernel all the time.

MenuApplets/Keymap/main.cpp
```diff
@@ -0,0 +1,136 @@
...
+        char buf[24];
+        if (get_keymap_name(buf, 24) < 0) {
+            perror("get_keymap_name");
+            m_keymap_name = "??";
+        } else {
+            m_keymap_name = String(buf).substring(0, 2).to_uppercase();
+        }
+        m_keymap_name_width = Gfx::Font::default_bold_font().width(m_keymap_name);
+
+        update();
```
bugaevc, line 94: It would be more efficient to not call `update()` (and subsequently draw, and send the new version to WS...) if the keymap name hasn't actually changed in the last 5 seconds :)

sppmacd: Fixed.

linusg:
18,892 line `CMakeLists.txt.user` file should probably be removed :)

Kernel/Process.h
```diff
@@ -319,6 +319,7 @@ class Process
     int sys$realpath(Userspace<const Syscall::SC_realpath_params*>);
     ssize_t sys$getrandom(Userspace<void*>, size_t, unsigned int);
     int sys$setkeymap(Userspace<const Syscall::SC_setkeymap_params*>);
+    int sys$get_keymap_name(Userspace<char*> buffer, size_t buffer_size);
```
awesomekling, line 322: We don't need a new syscall for this, you can get the current keymap name from `/proc/keymap` :)

awesomekling:
Closing this PR as "timed out" due to abandonment. Feel free to reopen if work continues. :)

PR #3656, opened 2020-10-01
HackStudio: Abstract away language-server details

alimpfard:
This commit moves all the logic that deals with the language server
(from HackStudio) into a LanguageClient class, provides some functions
to make constructing them easier, and makes all language servers use a
singular IPC definition.
Also fixes the FIXME about making autocompletion async.
This makes adding language servers in the future significantly less
duplicate-y, and significantly easier :^)

~~Note about async: For some reason, making it async creates a significant delay between the request (ctrl-space) and the response popping up - this is a TODO for this PR.~~

cc @itamar8910

### DIFF
--- a/DevTools/HackStudio/Editor.cpp
+++ b/DevTools/HackStudio/Editor.cpp
@@ -473,7 +473,7 @@ void Editor::set_document(GUI::TextDocument& doc)
     switch (code_document.language()) {
     case Language::Cpp:
         set_syntax_highlighter(make<GUI::CppSyntaxHighlighter>());
-        cpp_Language_server_connection().post_message(Messages::CppLanguageServer::FileOpened(code_document.file_path().string()));
+        m_language_client = get_language_client<LanguageClients::Cpp::ServerConnection>(project().root_directory());
         break;
     case Language::JavaScript:
         set_syntax_highlighter(make<GUI::JSSyntaxHighlighter>());
@@ -487,6 +487,9 @@ void Editor::set_document(GUI::TextDocument& doc)
     default:
         set_syntax_highlighter(nullptr);
     }
+
+    if (m_language_client)
+        m_language_client->open_file(code_document.file_path().string());
 }
 
 Optional<Editor::AutoCompleteRequestData> Editor::get_autocomplete_request_data()
@@ -517,22 +520,25 @@ Optional<Editor::AutoCompleteRequestData> Editor::get_autocomplete_request_data(
 
 void Editor::update_autocomplete(const AutoCompleteRequestData& data)
 {
-    if (code_document().language() != Language::Cpp)
+    if (!m_language_client)
         return;
-    auto autocomplete_response = cpp_Language_server_connection().send_sync<Messages::CppLanguageServer::AutoCompleteSuggestions>(
+
+    m_language_client->on_autocomplete_suggestions = [=, this](auto suggestions) {
+        if (suggestions.is_empty()) {
+            close_autocomplete();
+            return;
+        }
+
+        show_autocomplete(data);
+
+        m_autocomplete_box->update_suggestions(data.partial_input, move(suggestions));
+        m_autocomplete_in_focus = true;
+    };
+
+    m_language_client->request_autocomplete(
         code_document().file_path().string(),
         data.position.line(),
         data.position.column());
-    ASSERT(autocomplete_response);
-
-    auto suggestions = autocomplete_response->suggestions();
-    if (suggestions.is_empty()) {
-        close_autocomplete();
-        return;
-    }
-
-    m_autocomplete_box->update_suggestions(data.partial_input, move(suggestions));
-    m_autocomplete_in_focus = true;
 }
 
 void Editor::show_autocomplete(const AutoCompleteRequestData& data)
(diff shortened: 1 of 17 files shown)


DevTools/HackStudio/Editor.cpp
```diff
@@ -473,7 +473,7 @@ void Editor::set_document(GUI::TextDocument& doc)
     switch (code_document.language()) {
     case Language::Cpp:
         set_syntax_highlighter(make<GUI::CppSyntaxHighlighter>());
-        cpp_Language_server_connection().post_message(Messages::CppLanguageServer::FileOpened(code_document.file_path().string()));
+        m_language_client = create_language_client<LanguageClients::Cpp::ServerConnection>(project().root_directory());
```
itamar8910, line 476: Wouldn't this spawn a new language server process whenever we open a file?
Since a language server is supposed to handle multiple files, I think it would make more sense to have a single instance of the language server that serves the whole project (like we had before), with editors sharing the connection to it.

Also, previously we spawned a C++ language server on HackStudio startup regardless of whether there are even .cpp files in the project. We could improve on that by looking at the `.hsp` file for knowing which language servers to spawn (alternatively we could spawn language servers lazily as needed).

DevTools/HackStudio/LanguageClients/Cpp/ServerConnection.h
```diff
@@ -26,32 +26,37 @@
...
+#include <DevTools/HackStudio/LanguageServers/LanguageClientEndpoint.h>
+#include <DevTools/HackStudio/LanguageServers/LanguageServerEndpoint.h>
 #include <LibIPC/ServerConnection.h>
 
 namespace LanguageClients {
 namespace Cpp {
 
-class ServerConnection : public IPC::ServerConnection<CppLanguageClientEndpoint, CppLanguageServerEndpoint>
-    , public CppLanguageClientEndpoint {
+class ServerConnection : public HackStudio::ServerConnection
```
itamar8910, line 38: Now that we have all LSP servers and clients exposing the same IPC endpoints, I wonder if we still need a separate `ServerConnection.h` for every language?
We can have HackStudio::ServerConnection also inherit from `LanguageClientEndpoint` and implement `handshake()` & `handle(AutoCompleteSuggestions)` (And probably also get a path to the local socket of the server in its constructor).

We probably don't want to do something similar with langauge servers though, because we still want to allow flexability in the server-side e.g a language server knows how to update its internal representation of the codebase when `FileEditInsertText` is invoked.

DevTools/HackStudio/Editor.cpp
alimpfard: Hmm...I didn't think of that.
I'll make it share the servers.

> Lazily as needed

This is doing that though?

itamar8910: > This is doing that though?

Oh, I meant that e.g when we open a .sh file for the first time in a project we would spawn a shell language server, and when we open another .sh file we would use the connection to the already running shell language server.

Maybe we could have a global map between HackStudio::Language and a language server connection?

alimpfard: > Maybe we could have a global map between HackStudio::Language and a language server connection

Just a static local map per `ServerConnection` now.

DevTools/HackStudio/LanguageClients/Cpp/ServerConnection.h
alimpfard: Done.

DevTools/HackStudio/LanguageClients/ServerConnections.h
```diff
@@ -26,35 +26,28 @@
...
-        auto response = send_sync<Messages::CppLanguageServer::Greet>(m_project_path.string());
-        set_my_client_id(response->client_id());
-    }
-
-private:
-    ServerConnection(const String& project_path)
-        : IPC::ServerConnection<CppLanguageClientEndpoint, CppLanguageServerEndpoint>(*this, "/tmp/portal/language/cpp")
-        , m_project_path(project_path)
-    {
+#define LANGUAGE_CLIENT(namespace_, socket_name)                                               \
```
itamar8910, line 35: neat :)

itamar8910:
Okay! very nice changes. Working with language servers will be a lot nicer now.

DevTools/HackStudio/LanguageClient.h
```diff
@@ -0,0 +1,134 @@
...
+    virtual void request_autocomplete(const String& path, size_t cursor_line, size_t cursor_column);
+
+    void provide_autocomplete_suggestions(const Vector<String>&);
+
+    Function<void(Vector<String>)> on_autocomplete_suggestions;
+
+private:
+    ServerConnection& m_connection;
+    // Populated only if this object owns the server connection.
+    RefPtr<ServerConnection> m_server_connection;
```
awesomekling, line 124: Hmm this looks overly complicated :)
What if all `LanguageClient` objects simply have a `NonnullRefPtr<ServerConnection>` instead?

alimpfard: 🤔 I think the non-owning version isn't used anymore, so yeah, a NonnullRefPtr would work just fine.

PR #3657, opened 2020-10-02
Build: Emit paths in macros and debug sections relative to repo root

xTibor:
To avoid including paths from the build environment in the binaries. A step towards having reproducible builds.

This is kinda an experimental change, needs some testing before merge. (May cause issues with debuggers.)


nico:
Deterministic builds are cool. I wrote a long piece on them a while ago here https://blog.llvm.org/2019/11/deterministic-builds-with-clang-and-lld.html , the section "Getting debuggers to work well with locally deterministic builds" documents what the effect of this on debuggers is. (Doesn't affect printf debugging though!)

PR #3660, opened 2020-10-02
LibTar+Userland: Implement a un-tar only tar command

Petelliott:
Hello friends,

I've written a ustar compatible `tar(1)`. It can't create tar files yet, ~or operate on gzipped files~, but I'll be unable to work more on it for a few days, so I figured I'd check in what I have.


Libraries/LibTar/TarStream.cpp
```diff
@@ -0,0 +1,128 @@
...
+{
+    // verify that the stream has not advanced
+    ASSERT(m_tar_stream.m_generation == m_generation);
+
+    if (has_any_error())
+        return 0;
+
+    auto to_read = min(bytes.size(), m_tar_stream.header().size() - m_tar_stream.m_file_offset);
+
+    auto read = m_tar_stream.m_stream.read(bytes.slice(0, to_read));
```
asynts, line 47:
```suggestion
    auto read = m_tar_stream.m_stream.read(bytes.trim(to_read));
```

asynts: I would also call this variable `nread` because the method is already called `read`, but that's merely an esthetic.

Libraries/LibTar/TarStream.cpp
```diff
@@ -0,0 +1,128 @@
...
+void TarStream::advance()
+{
+    if (m_finished)
+        return;
+
+    m_generation++;
+    ASSERT(m_stream.discard_or_error(block_ceiling(m_header.size()) - m_file_offset));
+    m_file_offset = 0;
+
+    if (!m_stream.read_or_error(Bytes(static_cast<void*>(&m_header), sizeof(m_header)))) {
```
asynts, line 110:
```suggestion
    if (!m_stream.read_or_error(Bytes(&m_header, sizeof(m_header)))) {
```

asynts (commented):
I don't know enough about tar to tell if this is factually correct, but I am glad you picked up the `InputStream` thing, and the implementation seems good.

Note that I did not test the suggestions here, so make sure they compile (there is an include missing for the `Array` suggestion but GitHub is not flexible enough to make such suggestions.)

PR #3661, opened 2020-10-02
Base: Add missing options to the Shell(1) man page

linusg:
Basically an "improve docs" PR, but a serious one.

cc @alimpfard

### DIFF
--- a/Base/usr/share/man/man1/Shell.md
+++ b/Base/usr/share/man/man1/Shell.md
@@ -5,9 +5,10 @@
 ## Synopsis
 
 ```**sh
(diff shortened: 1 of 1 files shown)


Base/usr/share/man/man1/Shell.md
```diff
@@ -5,9 +5,10 @@
 ## Synopsis
 
 ```**sh
-$ Shell [--skip-shellrc]
+$ Shell [--skip-shellrc] [--live-formatting]
```
linusg, line 8: I added `--live-formatting` here as this is the invocation for REPL mode, and the option makes sense there.

PR #3662, opened 2020-10-02
Everywhere: Fix typos

nico:
Mostly in comments, but sprintf() now prints "August" instead of
"Auguest" so that's something.


awesomekling:
👕 👕 👕 👕 👕 👕 👕 👕 👕 👕 👕 👕 👕 👕 👕 👕

emanuele6:
What does the green t-shirt mean? is that a new inside joke?

PR #3664, opened 2020-10-02
Shell: IMPROVE HELP STRING - or something like that

alimpfard:
Why is this a separate PR? I dunno, your first guess is probably right.
(this actually annoyed me once I saw it lol)

### DIFF
--- a/Shell/main.cpp
+++ b/Shell/main.cpp
@@ -111,7 +111,7 @@ int main(int argc, char** argv)
     Core::ArgsParser parser;
     parser.add_option(command_to_run, "String to read commands from", "command-string", 'c', "command-string");
     parser.add_option(skip_rc_files, "Skip running shellrc files", "skip-shellrc", 0);
-    parser.add_option(format, "File to format", "format", 0, "file");
+    parser.add_option(format, "Format the given file into stdout and exit", "format", 0, "file");
     parser.add_option(should_format_live, "Enable live formatting", "live-formatting", 'f');
     parser.add_positional_argument(file_to_read_from, "File to read commands from", "file", Core::ArgsParser::Required::No);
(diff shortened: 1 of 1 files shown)


drunderscore:
idk, looks like spam to me ;)

Shell/main.cpp
```diff
@@ -111,7 +111,7 @@ int main(int argc, char** argv)
     Core::ArgsParser parser;
     parser.add_option(command_to_run, "String to read commands from", "command-string", 'c', "command-string");
     parser.add_option(skip_rc_files, "Skip running shellrc files", "skip-shellrc", 0);
-    parser.add_option(format, "File to format", "format", 0, "file");
+    parser.add_option(format, "Format the given file into stdout and exit", "format", 0, "file");
```
awesomekling, line 114: Hmm, wait, why is this better? This seems inconsistent with every other help string?

alimpfard: Is it? I personally thought that "File to format" isn't a description of what `--format` does, but rather the description of the value it takes?

Shell/main.cpp
```diff
@@ -111,7 +111,7 @@ int main(int argc, char** argv)
...
     parser.add_option(should_format_live, "Enable live formatting", "live-formatting", 'f');
     parser.add_positional_argument(file_to_read_from, "File to read commands from", "file", Core::ArgsParser::Required::No);
     parser.add_positional_argument(script_args, "Extra argumets to pass to the script (via $* and co)", "argument", Core::ArgsParser::Required::No);
```
TomMalbran, line 117: Since you are here. There is a tipo with "argumets"

Shell/main.cpp
awesomekling: Hm maybe you're right. Also the output from `--help` looks quite bad with the unaligned columns :(

PR #3668, opened 2020-10-02
Everywhere: Fix more typos

linusg:
Hate to do this in October (:shirt:) but @nico introduced me to `codespell` so I had to.

No idea why so many are left even after #3662?!

### DIFF
--- a/Base/etc/services
+++ b/Base/etc/services
@@ -213,7 +213,7 @@ pop3s		995/tcp				# POP-3 over SSL
 #> providing services to unknown callers, a service contact port is
 #> defined.  This list specifies the port used by the server process as its
 #> contact port.  While the IANA can not control uses of these ports it
-#> does register or list uses of these ports as a convienence to the
+#> does register or list uses of these ports as a convenience to the
 #> community.
 #
 socks		1080/tcp			# socks proxy server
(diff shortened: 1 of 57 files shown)


Base/etc/services
```diff
@@ -392,7 +392,7 @@ afs3-vlserver	7003/tcp			# volume location database
 afs3-vlserver	7003/udp
 afs3-kaserver	7004/tcp			# AFS/Kerberos authentication
 afs3-kaserver	7004/udp
-afs3-volser	7005/tcp			# volume managment server
```
xTibor, line 395: I'd rather see this fixed in the IANA registry.

linusg: True - could you point me to a repo or mailing list? Not exactly sure where this is coming from.

linusg: Oops, it says at the top of the file. https://www.iana.org/assignments/service-names-port-numbers/service-names-port-numbers.xhtml

xTibor: I have no idea about their contribution processes. Must be extremely bureaucratic though.

Base/usr/share/man/man2/pipe.md
```diff
@@ -29,7 +29,7 @@ writes some data to the pipe which the parent reads:
 ```c++
 #include <AK/Assertions.h>
 #include <stdio.h>
-#incldue <unistd.h>
+#include <unistd.h>
```
bugaevc, line 32: 🤦 and here we find out I didn't actually test the examples

awesomekling:
:shirt::shirt::shirt::shirt::shirt::shirt::shirt::shirt::shirt::shirt::shirt:

PR #3671, opened 2020-10-03
LibC: Remove unnecessary fallthrough comments.

emanuele6:
A "fallthrough comment" or `[[fallthrough]]` is only required after non-empty `case`s.

------------------------------

Neither `gcc` nor `clang` will report an `implicit-fallthrough` warning for these lines even if we remove these "fallthrough comments".

 There are already at least 840 instances of us not using a "fallthrough comment" or `[[fallthrough]]` in these situations:
```bash
$ git checkout master
Switched to branch 'master'
Your branch is up to date with 'origin/master'.
$ rg -UP $'case .*:\n(?=\\s*case)' | wc -l
842
```
NOTE: 2 of these matches are in `js` files.

### DIFF
--- a/Libraries/LibC/stdlib.cpp
+++ b/Libraries/LibC/stdlib.cpp
@@ -452,7 +452,6 @@ double strtod(const char* str, char** endptr)
                 digits_usable = true;
                 break;
             case DigitConsumeDecision::PosOverflow:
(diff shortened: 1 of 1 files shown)


emanuele6:
:shirt:

PR #3672, opened 2020-10-03
Add run.sh, to make it easy to run SerenityOS for the first time

xyproto:
Hi,

Some users might just want to give SerenityOS a spin, without being a developer. And some developers might want an easier and more streamlined way to check out the `master` branch and run the OS on a new computer.

This PR just adds a tiny little `run.sh` script in the root folder, that builds and runs SerenityOS in QEMU.

Dependencies still needs to be installed manually, but I believe this lowers the threshold for new users.

(I was not sure if I should submit this, since it's such a small thing, but I found it useful. I will fully understand if it's rejected).

Best regards!

### DIFF
--- a/run.sh
+++ b/run.sh
@@ -1,0 +1,14 @@
+#!/bin/sh
+#
+# Script to make it easy to build and run Serenity OS for the first time.
+# Old friends already know what to do.
+#
+set -e
+cd Toolchain
+./BuildIt.sh
+cd ../Build
+cmake -GNinja ..
+ninja
+ninja install
+ninja image
+ninja run


awesomekling:
I appreciate the effort, but I'm not going to merge this.

> this lowers the threshold for new users.

The threshold is intentionally where it is to avoid non-technical users and low-quality bug reports.
A competent developer can build the system in a few minutes without hand-holding scripts. :)

caelwithcats:
@awesomekling that's not a good idea. You should welcome new users, not discourage them. I noticed you are not using issue templates. It would help users from making bug reports without much explanation if you had an issue template like this:
```
** Describe the bug **

** Steps to reproduce **
1. Open the menu
2. In the demos menu item click Eyes

awesomekling:
Appealing to new users who are unwilling or unable to put in a minuscule amount of work to familiarize themselves with the project is a non-goal.

The SerenityOS project currently enjoys an extremely high signal-to-noise ratio, and this is one of the reasons why.

You are welcome to run your own projects as you please. :)

PR #3675, opened 2020-10-03
HackStudio: Add a Shell language server; Plus a whole lot of misc things

alimpfard:
In this PR:
- Bit of a language server architecture change (cc @itamar8910) - the C++ server should behave the same as before
- Some compiletime IPC overload issue detection
- A shell language server

Things I'm ~not~ somewhat sure about:
- What should the shell's working directory be? user's home directory seemed best
- What should the shell's PATH be? just the default places for now.

### DIFF
--- a/DevTools/HackStudio/LanguageServers/Shell/main.cpp
+++ b/DevTools/HackStudio/LanguageServers/Shell/main.cpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2020, Itamar S. <itamar8910@gmail.com>
+ * Copyright (c) 2020, the SerenityOS developers.
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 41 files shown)


dkoao:
>What should the shell's working directory be?
>What should the shell's PATH be?

I think that both of these should have the same values that you'd get by opening the shell as the current user, unless the user overrides them.

alimpfard:
That would mean running the shellrc though, I don't think embedding an entire functional shell in the language server is a good idea...

DevTools/HackStudio/LanguageServers/Shell/main.cpp
```diff
@@ -0,0 +1,51 @@
...
+{
+    Core::EventLoop event_loop;
+    if (pledge("stdio unix rpath", nullptr) < 0) {
+        perror("pledge");
+        return 1;
+    }
+
+    auto socket = Core::LocalSocket::take_over_accepted_socket_from_system_server();
+    IPC::new_client_connection<LanguageServers::Shell::ClientConnection>(socket.release_nonnull(), 1);
+    if (pledge("stdio rpath", nullptr) < 0) {
```
awesomekling, line 46: Would be cool to `unveil` the project root later on.

alimpfard: `unveil` is pretty questionable, we'd at least need to make everything readable for the shell, imagine trying to complete `/b` and getting nothing since the server couldn't stat() it

PR #3677, opened 2020-10-03
AK+Buffered: Call write instead of write_or_error.

### LINKED ISSUE #3666: LibGfx: Saving a font corrupts it
Demo:
juGlMixqne

Bisected to 5056d8bb200f9b9a8ee346738e8c2d95b4434bf4
cc @asynts

asynts:
Fixes #3666.

This pull request makes `Buffered` non-copyable, this did not cause any issues as far as I know but it shouldn't be copyable anyways.

The actual fix is the following:

~~~diff
diff --git a/AK/Buffered.h b/AK/Buffered.h
index 3c62862ef..d4a98cd15 100644
--- a/AK/Buffered.h
+++ b/AK/Buffered.h
@@ -182,7 +182,7 @@ public:
             flush();
 
             if (bytes.size() - nwritten >= Size)
-                nwritten += m_stream.write_or_error(bytes.slice(nwritten));
+                nwritten += m_stream.write(bytes.slice(nwritten));
 
             nwritten += write(bytes.slice(nwritten));
         }
~~~

Notice that `write_or_error` returns a `bool` which is converted into `size_t`.

---

Thanks @xTibor for pointing this out. https://github.com/SerenityOS/serenity/issues/3666#issuecomment-703130854

### DIFF
--- a/AK/Buffered.h
+++ b/AK/Buffered.h
@@ -178,7 +178,7 @@ public:
             flush();
 
             if (bytes.size() - nwritten >= Size)
-                nwritten += m_stream.write_or_error(bytes.slice(nwritten));
+                nwritten += m_stream.write(bytes.slice(nwritten));
 
             nwritten += write(bytes.slice(nwritten));
         }
--- a/AK/Buffered.h
+++ b/AK/Buffered.h
@@ -37,11 +37,20 @@ class Buffered;
 
 template<typename StreamType, size_t Size>
 class Buffered<StreamType, Size, typename EnableIf<IsBaseOf<InputStream, StreamType>::value>::Type> final : public InputStream {
+    AK_MAKE_NONCOPYABLE(Buffered);
+
 public:
     template<typename... Parameters>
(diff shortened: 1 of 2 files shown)


AK/Buffered.h
```diff
@@ -37,13 +37,24 @@ class Buffered;
...
     {
     }
 
+    Buffered(Buffered&& other)
+        : m_stream(move(other.m_stream))
+    {
+        other.buffer().copy_to(buffer());
+        m_buffered = other.m_buffered;
+
+        other.m_buffered = 0;
```
awesomekling, line 55: I like using `exchange` for this pattern:
```cpp
m_buffered = exchange(other.m_buffered, 0);
```

AK/Buffered.h
```diff
@@ -119,21 +130,33 @@ class Buffered<StreamType, Size, typename EnableIf<IsBaseOf<InputStream, StreamT
...
```
awesomekling, line 153: Same comment about `exchange`.

PR #3678, opened 2020-10-03
LibJS: Strict mode should be function- or file-level, not block-level

### LINKED ISSUE #3641: LibJS: Strict mode should be per-file or per-function, not per-block
cc @linusg

mattco98:
Fixes #3641

### DIFF
--- a/Libraries/LibJS/AST.cpp
+++ b/Libraries/LibJS/AST.cpp
@@ -89,6 +89,11 @@ Value ScopeNode::execute(Interpreter& interpreter, GlobalObject& global_object) 
     return interpreter.execute_statement(global_object, *this);
 }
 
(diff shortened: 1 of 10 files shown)


awesomekling:
Can we store this with the function or program node instead of the block node?

linusg:
This comment is about to become outdated and should be removed :)

https://github.com/SerenityOS/serenity/blob/e4bda2e1e70efed986745431e132b6608586b991/Libraries/LibJS/Interpreter.h#L89-L95

mattco98:
> Can we store this with the function or program node instead of the block node?
Yeah that'd probably be better, I'll do that instead :)

> This comment is about to become outdated and should be removed :)
True, I'll go ahead and remove that

awesomekling:
Excellent @mattco98, thanks for fixing this :)

PR #3683, opened 2020-10-04
WindowServer: Add transparency filters for windows

tomuta:
This implements blur effects for transparent windows. The amount
of blur for a transparent window can be changed using Ctrl+Logo+Wheel.

Still very much work in progress:

- [ ] Fix some glitches that cause windows to partially bleed through, e.g. when Snake is covered by an opaque window that is covered by a blurred window.
- [ ] Fix some glitches when moving a window over a blurred window.
- [ ] Fix `ASSERTION FAILED: !flush_rects.intersects(rect)` in `Compositor.cpp:311` being triggered sometimes
- [x] Figure out why `SpatialGaussianBlurFilter<N>` with `N != 5` distorts the picture
- [ ] Try Algorithm #4 from http://blog.ivank.net/fastest-gaussian-blur.html for better performance, maybe named `FastGaussianBlurFilter`
- [ ] Add ability to use transparency/blur for window frames (stretch goal)

Screenshot from 2020-10-04 11-33-51

Envisioned use case: Password prompt for privilege elevation purposes


awesomekling:
Very spicy! Maybe we can piggyback on some of this to implement basic window shadows as well :)

tomuta:
> Very spicy! Maybe we can piggyback on some of this to implement basic window shadows as well :)

I thought about this. What complicates this dramatically is the fact that the backing store for windows only covers the window content area, not the entire window. While I'd like to eventually double-buffer the window frames (which would be necessary for shadows), doing so right now would either require the frame to be split into 4 separate bitmaps (or I guess the top/bottom and left/right could be rendered to 2 bitmaps)...

alimpfard:
Neat, a second use-case for the filters!

> Figure out why SpatialGaussianBlurFilter<N> with N != 5 distorts the picture

I'd look into the kernels, they might be generated wrong?

tomuta:
> Neat, a second use-case for the filters!
> 
> > Figure out why SpatialGaussianBlurFilter with N != 5 distorts the picture
> 
> I'd look into the kernels, they might be generated wrong?

Possible, I just copy+pasted from PixelPaint. I'm not an expert in this area ;-) Basically that sets all values to 1 and then calls `normalize` on it. Maybe that's wrong?

alimpfard:
For a Gaussian filter? setting everything to 1 then calling `normalize()` will make a box filter

tomuta:
> For a Gaussian filter? setting everything to 1 then calling `normalize()` will make a box filter

Yeah you're right. Never mind, PixelPaint does some more sophisticated math. Looks like I copied the wrong one. But I guess then the question is why does the BoxFilter severely distort if `N != 5` ;-)

Applications/PixelPaint/FilterParams.h
```diff
@@ -0,0 +1,191 @@
...
+                auto r = sqrt(x * x + y * y);
+                kernel.elements()[x + 2][y + 2] = (exp(-(r * r) / s)) / (M_PI * s);
+            }
+        }
+
+        normalize(kernel);
+
+        return make<typename Gfx::GenericConvolutionFilter<N>::Parameters>(kernel);
+    }
+};
```
alimpfard, line 139: This generates a gaussian kernel.

Not sure what sort of distortion you're talking about, but a box blur shouldn't cause any particular distortions (maybe some darkening around the edges if it's set to not wrap?)

tomuta: At `3` colors get distorted in a weird way, at `9` it becomes really dark. Can't remember what it did at `7`. Haven't tried other values.

alimpfard: I'll pull this PR and take a looksie.

tomuta: So I copied this from PixelPaint into the `calc_matrix` function in `Compositor.cpp` and it crashes at startup trying to access `V0xffffffff`
```c++
    static Matrix<N, float> calc_matrix()
    {
        Matrix<N, float> kernel;
        auto sigma = 1.0f;
        auto s = 2.0f * sigma * sigma;

        for (auto x = -(ssize_t)N / 2; x <= (ssize_t)N / 2; x++) {
            for (auto y = -(ssize_t)N / 2; y <= (ssize_t)N / 2; y++) {
                auto r = sqrt(x * x + y * y);
                kernel.elements()[x + 2][y + 2] = (exp(-(r * r) / s)) / (M_PI * s);
            }
        }

        Gfx::normalize(kernel);
        return kernel;
    }
```

tomuta: That code crashes with `N == 7`

alimpfard: :thinking: weird...
the 5x5 gaussian filter crashes in pixelpaint too, but I assume it generates this ok before crashing in apply()

alimpfard: > That code crashes with N == 7

That makes too much sense, `-N/2` would be -3, so it's accessing elements at `-1,n`

alimpfard:
This should fix the colour distortions at least:

```diff
diff --git a/Libraries/LibGfx/Filters/GenericConvolutionFilter.h b/Libraries/LibGfx/Filters/GenericConvolutionFilter.h
index 73b74c45f..7477c7b20 100644
--- a/Libraries/LibGfx/Filters/GenericConvolutionFilter.h
+++ b/Libraries/LibGfx/Filters/GenericConvolutionFilter.h
@@ -73,6 +73,7 @@ public:
     class ApplyCache {
         template<size_t>
         friend class GenericConvolutionFilter;
+
     private:
         RefPtr<Gfx::Bitmap> m_target;
     };
@@ -150,7 +151,7 @@ public:
                     }
                 }

-                // The float->u8 overflow is intentional.
+                value.clamp(0, 255);
                 render_target_bitmap->set_pixel(i, j, Color(value.x(), value.y(), value.z(), source.get_pixel(i + source_delta_x, j + source_delta_y).alpha()));
             }
         }
diff --git a/Libraries/LibGfx/Vector3.h b/Libraries/LibGfx/Vector3.h
index a0030ad99..bf60f86cf 100644
--- a/Libraries/LibGfx/Vector3.h
+++ b/Libraries/LibGfx/Vector3.h
@@ -87,6 +87,16 @@ public:
         return *this * inv_length;
     }

+    void clamp(T min_value, T max_value)
+    {
+        m_x = max(min_value, m_x);
+        m_y = max(min_value, m_y);
+        m_z = max(min_value, m_z);
+        m_x = min(max_value, m_x);
+        m_y = min(max_value, m_y);
+        m_z = min(max_value, m_z);
+    }
+
     void normalize()
     {
         T inv_length = 1 / length();
```

Applications/PixelPaint/FilterParams.h
alimpfard: I'm not quite sure how I ever thought this convolution filter was _correct_.

alimpfard: I pushed a fix for this (and the distortions) in a [PR-PR](https://github.com/tomuta/serenity/pull/1) (\*giggle\*)

tomuta: Awesome, thanks!

alimpfard:
> Try Algorithm #4 from http://blog.ivank.net/fastest-gaussian-blur.html for better performance

Another option would be to just implement the _actual_ gaussian blur with FFT (Cooley-Tukey is O(nlgn), not quite the linear that algorithm offers, but FFT/iFFT would be a useful operation to have either way).

I could've sworn I had an implementation of that I wanted to use for the FFT multiplication algorithm, but I can't find it anywhere...:facepalm:

tomuta:
> Another option would be to just implement the _actual_ gaussian blur with FFT (Cooley-Tukey is O(nlgn), not quite the linear that algorithm offers, but FFT/iFFT would be a useful operation to have either way).

Well, in this case I think performance matters more than accuracy. Based on the samples on that page, that algorithm looks pretty darn close, and looks like it would be significantly more performant. And its linear nature makes it even more appealing for this particular use case.

scarlettbegonia:
Omg I am super sick and trying to take this all in. Travis? Travis Or Andrea may I please ask for a summary of what it is I am reading and why you are in my life. Please- I feel my soul searching for answers and truths and I am so ill. I found out I may have colon cancer and I am trying to now more than ever just get my clarity- my truth. Maybe from this, there will be a better future for my daughters- one I can say I was part of. I long so much to be part of something and be needed. To have value.

PR #3684, opened 2020-10-04
Replace more usages of printf with format.

asynts:


awesomekling:
*Carlender?* 📆  🤔

asynts:
My spelling really isn't the best. :smile:

PR #3686, opened 2020-10-04
LibJS: Unify syntax highlighting

linusg:
So far we have three different syntax highlighters for LibJS:

- `js`'s `Line::Editor` stylization
- `JS::MarkupGenerator`
- `GUI::JSSyntaxHighlighter`

This not only caused repetition of most token types in each highlighter but also a lot of inconsistency regarding the styling of certain tokens:

- `JSSyntaxHighlighter` was considering `TokenType::Period` to be an operator whereas `MarkupGenerator` categorized it as punctuation.
- `MarkupGenerator` was considering `TokenType::{Break,Case,Continue,Default,Switch,With}` control keywords whereas `JSSyntaxHighlighter` just disregarded them
- `MarkupGenerator` considered some future reserved keywords invalid and others not. `JSSyntaxHighlighter` and `js` disregarded most

Adding a new token type meant adding it to `ENUMERATE_JS_TOKENS` as well as each individual highlighter's `switch`/`case` construct.

I added a `TokenCategory` enum, and each `TokenType` is now associated to a certain category, which the syntax highlighters then can use for styling rather than operating on the token type directly. This also makes changing a token's category everywhere easier, should we need to do that (e.g. I decided to make `TokenType::{Period,QuestionMarkPeriod}` `TokenCategory::Operator` for now, but we might want to change them to `Punctuation`.

### DIFF
--- a/Libraries/LibGUI/JSSyntaxHighlighter.cpp
+++ b/Libraries/LibGUI/JSSyntaxHighlighter.cpp
@@ -35,110 +35,23 @@ namespace GUI {
 
 static TextStyle style_for_token_type(Gfx::Palette palette, JS::TokenType type)
 {
-    switch (type) {
-    case JS::TokenType::Invalid:
-    case JS::TokenType::Eof:
+    switch (JS::Token::category(type)) {
+    case JS::TokenCategory::Invalid:
         return { palette.syntax_comment() };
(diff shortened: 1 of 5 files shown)


awesomekling:
That's really nice! :)

Issue #3670, opened 2020-10-03
LibJS/AK: m_has_value assertion failure in AK::JsonParser::parse_number


Lubrsi:
Found while fuzzing.
[crash-f04da97ef8aa4111aa611c5ae72d95a1d4dae1fc.txt](https://github.com/SerenityOS/serenity/files/5321027/crash-f04da97ef8aa4111aa611c5ae72d95a1d4dae1fc.txt)

Line:
https://github.com/SerenityOS/serenity/blob/9dee140a9f3bc1c674ca32fd34d3eb7085e2f5d8/AK/JsonParser.cpp#L215

Trace:
```
USERSPACE(24) ASSERTION FAILED: m_has_value
.././AK/Optional.h:132
[js(24:24)]: Terminating js(24) due to signal 6
[FinalizerTask(3:3)]: CPU[0] NP(error) fault at invalid address V0x01601ccc
[FinalizerTask(3:3)]: copy_from_user(0xc094a9ac, 0x01601ccc, 4) failed at V0x01601ccc
[FinalizerTask(3:3)]: 0xdeadc0de  Kernel::Processor::switch_context(Kernel::Thread*&, Kernel::Thread*&) +181
0xdeadc0de  Kernel::Scheduler::context_switch(Kernel::Thread*) +181
0xdeadc0de  Kernel::Scheduler::pick_next() +1952
0xdeadc0de  Kernel::Processor::check_invoke_scheduler() +125
0xdeadc0de  Kernel::Thread::die_if_needed() +157
0xdeadc0de  syscall_handler +1414
0xdeadc0de  syscall_asm_entry +49
0x080bd280  raise +22
0x080480bd   +13
0x080b91a5  tolower +0
0x080dbab2  AK::JsonParser::parse_number() +520
0x080dbf70  AK::JsonParser::parse_helper() +130
0x080dc245  AK::JsonParser::parse() +25
0x080d1566  AK::JsonValue::from_string(AK::StringView const&) +64
0x0807aaa1  JS::JSONObject::parse(JS::VM&, JS::GlobalObject&) +135
0x0804bba6  AK::Function<JS::Value (JS::VM&, JS::GlobalObject&)>::CallableWrapper<JS::Value (*)(JS::VM&, JS::GlobalObject&)>::call(JS::VM&, JS::GlobalObject&) const +22
0x080a3749  JS::NativeFunction::call() +59
[...]

PR #3687, opened 2020-10-04
AK: check fractional string has_value() in JsonParser

tuckerpo:
Resolves edge case found via fuzzing in #3670

### DIFF
--- a/AK/JsonParser.cpp
+++ b/AK/JsonParser.cpp
@@ -212,7 +212,10 @@ Optional<JsonValue> JsonParser::parse_number()
             whole = number.value();
         }
 
-        int fraction = fraction_string.to_uint().value();
+        auto fraction_string_uint = fraction_string.to_uint();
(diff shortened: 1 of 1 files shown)


AK/JsonParser.cpp
```diff
@@ -212,6 +212,8 @@ Optional<JsonValue> JsonParser::parse_number()
             whole = number.value();
         }
 
+        if (!fraction_string.to_uint().has_value())
+            return {};
         int fraction = fraction_string.to_uint().value();
```
awesomekling, line 217: Good fix! But let's put the result of `to_uint()` in a local variable here so we don't do string-to-integer conversion twice :)

awesomekling (changes requested):

AK/JsonParser.cpp
tuckerpo: Done.

PR #3690, opened 2020-10-05
LibJS: Evaluate AssignmentExpression LHS before RHS according to the spec

### LINKED ISSUE #3689: LibJS: Assignment expression LHS/RHS evaluation is in wrong order
Consider the following:

```js
var a = [];
function b() { console.log("b()"); throw Error(); }
function c() { console.log("c()"); throw Error(); }
a[b()] = c();
```

(same with `+=` etc.)

Expected:

```
b()
Uncaught exception: [Error]
 -> b
 -> (global execution context)
```

Actual behaviour:

```
c()
Uncaught exception: [Error]
 -> c
 -> (global execution context)
```

Spec:

https://tc39.es/ecma262/#sec-assignment-operators-runtime-semantics-evaluation

linusg:
Fixes #3689.

### DIFF
--- a/Libraries/LibJS/AST.cpp
+++ b/Libraries/LibJS/AST.cpp
@@ -1198,84 +1198,67 @@ void ThisExpression::dump(int indent) const
 
 Value AssignmentExpression::execute(Interpreter& interpreter, GlobalObject& global_object) const
 {
-    auto rhs_result = m_rhs->execute(interpreter, global_object);
-    if (interpreter.exception())
-        return {};
+#define EXECUTE_LHS_AND_RHS()                                    \
+    do {                                                         \
+        lhs_result = m_lhs->execute(interpreter, global_object); \
+        if (interpreter.exception())                             \
+            return {};                                           \
+        rhs_result = m_rhs->execute(interpreter, global_object); \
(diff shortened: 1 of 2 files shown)


Libraries/LibJS/AST.cpp
```diff
@@ -1198,84 +1198,67 @@ void ThisExpression::dump(int indent) const
...
-        return {};
+#define EXECUTE_LHS_AND_RHS()                                    \
+    do {                                                         \
+        lhs_result = m_lhs->execute(interpreter, global_object); \
+        if (interpreter.exception())                             \
+            return {};                                           \
+        rhs_result = m_rhs->execute(interpreter, global_object); \
+        if (interpreter.exception())                             \
+            return {};                                           \
+    } while (0)
```
linusg, line 1209: I tried doing this with a two value struct the (like `compute_this_and_callee()`) but couldn't get it working in a concise way because of the `auto` in `auto [lhs_result, rhs_result] = ...`...

alimpfard: Perhaps we should have something akin to `std::bind`?
Though that particular function is...ugly imo

PR #3693, opened 2020-10-05
Move StringImpl::operator== implementation into StringImpl

nico:


AK/StringImpl.h
```diff
@@ -33,6 +33,12 @@
 #include <AK/Types.h>
 #include <AK/kmalloc.h>
 
+#ifdef KERNEL
+#    include <Kernel/StdLib.h>
+#else
+#    include <string.h>
+#endif
```
awesomekling, line 40: Instead of pulling in headers, we can also use `__builtin_memcmp` :]

nico: Done!

PR #3694, opened 2020-10-05
LibWeb: Add pc CSS unit and get a correct ex unit

Lubrsi:
LibWeb: Add pc CSS unit

-----

LibGfx+FontEditor+Fonts: Add "mean line" value to all fonts

The main inspiration behind this was to have a correct ex CSS unit.
The mean line is based off what it shows in the CSS Values and Units
Level 4 specification, section 6.1.1.

https://www.w3.org/TR/css-values-4/#font-relative-lengths


awesomekling:
Cool!

PR #3696, opened 2020-10-05
Use new format functions in FontEditor, Help and HexEditor.

asynts:


Libraries/LibGUI/ModelIndex.cpp
```diff
@@ -46,4 +46,14 @@ const LogStream& operator<<(const LogStream& stream, const ModelIndex& value)
     return stream << String::format("ModelIndex(%d,%d)", value.row(), value.column());
 }
 
+} // namespace GUI
```
emanuele6, line 49: We don't put these comments anywhere else, let's not put it here

Libraries/LibGUI/ModelIndex.cpp
```diff
@@ -46,4 +46,14 @@ const LogStream& operator<<(const LogStream& stream, const ModelIndex& value)
...
+
+void AK::Formatter<GUI::ModelIndex>::format(TypeErasedFormatParams& params, FormatBuilder& builder, const GUI::ModelIndex& value)
+{
+    Formatter<StringView> formatter { *this };
+
+    if (value.internal_data())
+        formatter.format(params, builder, String::formatted("ModelIndex({},{},{:p})", value.row(), value.column(), value.internal_data()));
+    else
+        formatter.format(params, builder, String::formatted("ModelIndex({},{})", value.row(), value.column()));
```
emanuele6, line 58: This function would be more readable at the top of the file imo.

emanuele6:
You need to include `AK/Format.h` in `Libraries/LibGUI/ModelIndex.cpp` or the code won't compile. (did you forget to compile before commiting/pushing?)

Applications/HexEditor/HexEditor.cpp
```diff
@@ -518,7 +516,7 @@ void HexEditor::paint_event(GUI::PaintEvent& event)
         };
 
         bool is_current_line = (m_position / bytes_per_row()) == i;
-        auto line = String::format("0x%08X", i * bytes_per_row());
+        auto line = String::formatted("{:#010X}", i * bytes_per_row());
```
emanuele6, line 519: I don't know how `String::formatted` work, but `0x%08X` and `%8X` can't both be equivalent to `{:#010X}`, you must have made a mistake in at least one of these lines.

Applications/HexEditor/HexEditorWidget.cpp
```diff
@@ -54,11 +54,11 @@ HexEditorWidget::HexEditorWidget()
     m_editor = add<HexEditor>();
 
     m_editor->on_status_change = [this](int position, HexEditor::EditMode edit_mode, int selection_start, int selection_end) {
-        m_statusbar->set_text(0, String::format("Offset: %8X", position));
-        m_statusbar->set_text(1, String::format("Edit Mode: %s", edit_mode == HexEditor::EditMode::Hex ? "Hex" : "Text"));
-        m_statusbar->set_text(2, String::format("Selection Start: %d", selection_start));
-        m_statusbar->set_text(3, String::format("Selection End: %d", selection_end));
-        m_statusbar->set_text(4, String::format("Selected Bytes: %d", abs(selection_end - selection_start) + 1));
+        m_statusbar->set_text(0, String::formatted("Offset: {:#010X}", position));
```
emanuele6, line 57: I don't know how `String::formatted` work, but `0x%08X` and `%8X` can't both be equivalent to `{:#010X}`, you must have made a mistake in at least one of these lines.

emanuele6 (changes requested):

Libraries/LibGUI/ModelIndex.cpp
asynts: What do you mean by that?

Applications/HexEditor/HexEditor.cpp
asynts: The `#` means "alternative form", for integers this means that prefixes `0x`, `0` or `0b` are added. (If the uppercase `X` is used then `0x` becomes `0X` and for `B` similarly.)

This prefix counts towards the width, so if we want to have 8 significant digits, then we need a width of 10 if we use the alternative form.

I replaced `8X` with `#010X` because this makes it more explicit that the offset is hexadecimal (this only matters for small numbers anyways.)

Applications/HexEditor/HexEditorWidget.cpp
asynts: Again, it's more explicit. (It's preference really.)

Applications/HexEditor/HexEditor.cpp
emanuele6: > I replaced 8X with #010X because this makes it more explicit that the offset is hexadecimal (this only matters for small numbers anyways.)

You could have wrote something about it in the commit message or, at least, in the PR message :/

asynts:
The build error occurred because I had no formatter defined for `KeyCode`. I assume that the Travis build defines `HEX_DEBUG` therefor it only occurs there.

In the future it might be better to define a formatter for any enum type (just print the integer) but that requires SFINAE magic, so I am just adding a cast:

~~~diff
@@ -346,7 +346,7 @@ ./Applications/HexEditor/HexEditor.cpp
 void HexEditor::keydown_event(GUI::KeyEvent& event)
 {
 #ifdef HEX_DEBUG
-    outln("HexEditor::keydown_event key={}", event.key());
+    outln("HexEditor::keydown_event key={}", static_cast<u8>(event.key()));
 #endif
 
     if (event.key() == KeyCode::Key_Up) {
~~~

Libraries/LibGUI/ModelIndex.cpp
asynts: They do appear in some places in the code base. Personally, I like to add them when multiple namespaces are involved to make the code more readable.

Applications/HexEditor/HexEditor.cpp
asynts: Added this to the commit message:

~~~
I changed the formatting for some numbers when it made sense, because I
believe '0x00000020' to be more expressive than '      20' for example.
~~~

Libraries/LibGUI/ModelIndex.cpp
emanuele6: It would be more clear that that function is not part of the GUI namespace.

Libraries/LibGUI/ModelIndex.cpp
emanuele6: > They do appear in some places in the code base.

To be fair, the only ones that I found are in `AK/Format.h` and `AK/Format.cpp`, and were added by you :/

Libraries/LibGUI/ModelIndex.cpp
asynts: That's what the `// namespace GUI` comment is for.

emanuele6: > That's what the // namespace GUI comment is for.

Doesn't really help since there is already a `}` at the end of the file and `// namespace xxx` is not required.

Libraries/LibGUI/ModelIndex.cpp
emanuele6: > [`CodingStyles.md`](https://github.com/SerenityOS/serenity/blob/52c31bb743ce5baa6358f1db34c5341833b8cdf1/Documentation/CodingStyle.md)

That isn't "codebase", and they aren't comments (just text); the only **comments** of that kind in the code are by you.

> so I'll leave it in.

Ok, that's fine.

Libraries/LibGUI/ModelIndex.cpp
emanuele6, line 58:
```suggestion
}

namespace AK {
void Formatter<GUI::ModelIndex>::format(TypeErasedFormatParams& params, FormatBuilder& builder, const GUI::ModelIndex& value)
{
    Formatter<StringView> formatter { *this };

    if (value.internal_data())
        formatter.format(params, builder, String::formatted("ModelIndex({},{},{:p})", value.row(), value.column(), value.internal_data()));
    else
        formatter.format(params, builder, String::formatted("ModelIndex({},{})", value.row(), value.column()));
}
```

emanuele6 (commented):
what about this? this makes it more clear that the function is not part of the `GUI` namespace and is consistent with the header file

Libraries/LibGUI/ModelIndex.cpp
emanuele6: also avoids this C++ weirdness:

```cpp
namespace Namespace {
    using Type = int;
}

void Namespace::SomeFunction()
{
    // Namespace:: not required
    Type a;
}

Libraries/LibGUI/ModelIndex.cpp
asynts: @awesomekling What's your opinion on this?

awesomekling: Let's just skip the `// namespace Foo`.

asynts:
I applied the following changes to resolve the issue:

~~~diff
@@ -46,9 +46,11 @@ ./Libraries/LibGUI/ModelIndex.cpp
     return stream << String::format("ModelIndex(%d,%d)", value.row(), value.column());
 }
 
-} // namespace GUI
+}
+
+namespace AK {
 
-void AK::Formatter<GUI::ModelIndex>::format(TypeErasedFormatParams& params, FormatBuilder& builder, const GUI::ModelIndex& value)
+void Formatter<GUI::ModelIndex>::format(TypeErasedFormatParams& params, FormatBuilder& builder, const GUI::ModelIndex& value)
 {
     Formatter<StringView> formatter { *this };
 
@@ -57,3 +59,5 @@ void AK::Formatter<GUI::ModelIndex>::format(TypeErasedFormatParams& params, Form
     else
         formatter.format(params, builder, String::formatted("ModelIndex({},{})", value.row(), value.column()));
 }
+
+}
~~~

I left the other `// namespace` comments in, because in `Format.cpp` there is a nested anonymous namespace which is difficult to spot otherwise.

https://github.com/SerenityOS/serenity/blob/be693e95ff39cf580ce8659c94acb26cb94ef138/AK/Format.cpp#L97-L101

PR #3699, opened 2020-10-06
AK+Format: Exclude prefix from width calculation, and the "crusade" continues.

asynts:
Recently, I said, that I wanted to create smaller pull requests... Hmm... :^)

There are really two interesting commits here, 18cd462ca014866395dcecf7e7643846b83e752f and 9716ffa7103b2e12df082ee9ab55631b4b152f9c. I explained my thoughts in the commits themselves.

The rest is just me replacing more usages.


awesomekling:
Totally agree with reasoning :)

PR #3701, opened 2020-10-06
LibGemini: Fix crash on Link lines without explicit link text

nico:
Link::Link() does:

    if (m_name.is_null())
        m_name = m_url.to_string();

This tries to set the link text to the link url if there's no
explicit link text, but m_url.to_string() returns a temporary
string object, and m_name was a StringView, so this ended
pointing to invalid memory. This made the converted HTML
contain garbage data as link text, which made LibWeb crash.

(LibWeb probably shouldn't crash on links with garbage link
text, but in this case it helped identify a bug, so let's keep
that crash around since real web pages won't trigger it and
it's kind of useful to find bugs like this one.)

Makes gemini://rawtext.club/ no longer crash Browser.

### DIFF
--- a/Libraries/LibGemini/Document.h
+++ b/Libraries/LibGemini/Document.h
@@ -89,7 +89,7 @@ public:
 
 private:
     URL m_url;
-    StringView m_name;
+    String m_name;
 };
 
 class Preformatted : public Line {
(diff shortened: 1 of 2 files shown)


awesomekling:
Yeah using StringView like that was clearly a mistake :)

PR #3708, opened 2020-10-07
VariadicFormatParams: Use initialized data to create parent class

ldm5180:
Problem:
- m_data is being passed to the constructor of the parent class before
  it is initialized. This is not really a problem because the compiler
  knows the location and it is only a span being constructed, but it
  triggers a warning in clang for use-before-init.

Solution:
- Initialize using a default constructed array and then overwrite it
  inside the constructor after the member is initialized.

### DIFF
--- a/AK/Format.h
+++ b/AK/Format.h
@@ -164,13 +164,9 @@ private:
 
 class TypeErasedFormatParams {
 public:
-    explicit TypeErasedFormatParams(Span<const TypeErasedParameter> parameters)
-        : m_parameters(parameters)
-    {
-    }
-
     Span<const TypeErasedParameter> parameters() const { return m_parameters; }
 
+    void set_parameters(Span<const TypeErasedParameter> parameters) { m_parameters = parameters; }
     size_t take_next_index() { return m_next_index++; }
 
     size_t decode(size_t value, size_t default_value = 0);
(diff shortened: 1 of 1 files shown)


ldm5180:
This is kinda ugly. Better ideas are welcome...

asynts:
You could remove the constructor of `TypeErasedFormatParams` entirely, then you won't need the `decltype` stuff:

~~~diff
 class TypeErasedFormatParams {
 public:
-    explicit TypeErasedFormatParams(Span<const TypeErasedParameter> parameters)
-        : m_parameters(parameters)
-    {
-    } 
- 
     Span<const TypeErasedParameter> parameters() const { return m_parameters; }
- 
     void set_parameters(Span<const TypeErasedParameter> parameters) { m_parameters = parameters; }
+ 
     size_t take_next_index() { return m_next_index++; }
 
     size_t decode(size_t value, size_t default_value = 0);
 
 private:
     Span<const TypeErasedParameter> m_parameters;
     size_t m_next_index { 0 };
 };
~~~

~~~diff
 template<typename... Parameters>
 class VariadicFormatParams : public TypeErasedFormatParams {
 public:
     static_assert(sizeof...(Parameters) <= max_format_arguments);
 
     explicit VariadicFormatParams(const Parameters&... parameters)
-        : TypeErasedFormatParams(decltype(m_data){})
         , m_data({ TypeErasedParameter { &parameters, TypeErasedParameter::get_type<Parameters>(), __format_value<Parameters> }... })
     {
       this->set_parameters(m_data);
     }
 
 private:
     Array<TypeErasedParameter, sizeof...(Parameters)> m_data;
 };
~~~

(I haven't put this in the compiler, but something like this should work.)

ldm5180:
@asynts Thanks. That is obvious now that you point it out. :) I thought that the `Span` might require explicit construction, but I guess not. Thanks!

It has been updated and that seems to work.

PR #3711, opened 2020-10-07
LibJS: Disallow 'return' outside of a function

### LINKED ISSUE #3607: LibJS: Parser allows return outside function
...and the AST even executes it.
```
$ js -A
> return "foo"
Program
  (Children)
    ReturnStatement
      StringLiteral "foo"
"foo"
> 
```

mattco98:
I'd like to include a test, however our test's `expect().toEval` works by calling `new Function(<string>)()`, which obviously isn't great for this. We can add a test once we have support for `eval`.

Fixes #3607


alimpfard:
Easy test: conditional formatting in the Spreadsheet should be broken now.

PR #3713, opened 2020-10-07
LibJS: Fix PropertyName::from_value() for negative and non-int numbers

### LINKED ISSUE #3712: LibJS: ObjectExpression with computed NaN/Infinity key asserts
this works:

```js
> o = {}
{  }
> o[Infinity] = "foo"
"foo"
> o[NaN] = "bar"
"bar"
> o
{ "Infinity": "foo", "NaN": "bar" }
```

this also works:

```js
> o = {}
{  }
> o[1/0] = "foo"
"foo"
> o[1/"x"] = "bar"
"bar"
> o
{ "Infinity": "foo", "NaN": "bar" }
```

this doesn't work:

```js
// Infinity
> o = {[1/0]: "foo"}
js: ../Libraries/LibJS/Runtime/PropertyName.h:60: JS::PropertyName::PropertyName(i32): Assertion `index >= 0' failed.
```

this doesn't work either:

```js
// NaN
> o = {[1/"x"]: "bar"}
js: ../Libraries/LibJS/Runtime/PropertyName.h:60: JS::PropertyName::PropertyName(i32): Assertion `index >= 0' failed.
```

---

Don't ask *why* I tried this, I don't have an answer.

linusg:
It was converting *any* number to an i32 index, which obviously is not correct for negative ints, doubles, infinity and nan.

Fixes #3712.

### DIFF
--- a/Libraries/LibJS/AST.cpp
+++ b/Libraries/LibJS/AST.cpp
@@ -1549,8 +1549,6 @@ Value ObjectExpression::execute(Interpreter& interpreter, GlobalObject& global_o
             continue;
         }
 
-        if (interpreter.exception())
-            return {};
         auto value = property.value().execute(interpreter, global_object);
         if (interpreter.exception())
             return {};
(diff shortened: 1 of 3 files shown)


Libraries/LibJS/AST.cpp
```diff
@@ -1549,8 +1549,6 @@ Value ObjectExpression::execute(Interpreter& interpreter, GlobalObject& global_o
             continue;
         }
 
-        if (interpreter.exception())
-            return {};
```
linusg, line 1553: This is not needed - if we enter the if branch above, this is never reached, and otherwise there's an exception check right before that.

awesomekling:
:star:

PR #3718, opened 2020-10-08
LibJS: Disallow 'continue' & 'break' outside of their respective scopes

### LINKED ISSUE #3608: LibJS: Parser allows break/continue outside loop/switch statement
```
$ js -A
> break
Program
  (Children)
    BreakStatement
undefined
> continue
Program
  (Children)
    ContinueStatement
undefined
> 
```

mattco98:
`continue` is no longer allowed outside of a loop, and an unlabeled `break` is not longer allowed outside of a loop or switch statement. Labeled `break` statements are still allowed everywhere, even if the label does not exist (which will be fixed in a PR after this one is merged since it conflicts).

Fixes #3608


awesomekling:
👍🏻

PR #3722, opened 2020-10-08
Endian: constexpr constructors and conversion operators

ldm5180:
Problem:
- Constructors and conversion operators are not `constexpr`,
  but they can be.
- `constexpr` is needed here so that other classes can add `constexpr`
  evaluation.

Solution:
- Add the `constexpr` keyword to the constructors and
  conversion operators.
- Add `static_assert` tests which ensure the capability works.

### DIFF
--- a/AK/Endian.h
+++ b/AK/Endian.h
@@ -32,7 +32,7 @@
 namespace AK {
 
 template<typename T>
-ALWAYS_INLINE T convert_between_host_and_little_endian(T value)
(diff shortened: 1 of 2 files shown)


asynts:
Why did you close #3719? The pull request seems to be exactly the same?

ldm5180:
@asynts Sorry about that. It is the same. I wanted to stop spamming with failing builds while I adjusted so I closed it and then I didn't see a way to reopen.

PR #3723, opened 2020-10-08
Add __portable_abort() with well defined signature.

asynts:
The signatures of `::abort` and `std::abort` are inconsistent among standard library implementations.

https://github.com/SerenityOS/serenity/commit/1d96d5eea44ea4813d947cca9bd84cf10733d927#r43069456

In some standard library implementations `::abort` is marked with `noexcept`, in others it is not. Similarly, the signature of `std::abort` sometimes is marked with `extern "C"` and sometimes not, this is how stdlibc++ does it:

~~~c++
extern "C" void abort(void) throw() __attribute((noreturn));

namespace std {
     // Implicitly inherits C linkage.
     using ::abort;
}
~~~

I am not 100% sure about the libstdc++ stuff because it's difficult to follow the include paths of these GNU-ish headers, but it's really weird for sure.

---

@sv-ochis Can you verify that this builds on MacOS?

### DIFF
--- a/AK/Platform.cpp
+++ b/AK/Platform.cpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2018-2020, Andreas Kling <kling@serenityos.org>
+ * Copyright (c) 2020, the SerenityOS developers.
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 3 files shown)


bugaevc:
* what's the point of `__portable_abort`, compared to just `#include`ing `<stdlib.h>` and using `abort()` (instead of redeclaring it)?
* why change the kernel at all?

asynts:
The problem is that we need to define the assert macros before anything else. Otherwise, they could be replaced incorrectly.

Although, we could get away with it in `stdlib.h`. Hmm.

I had to change some kernel code, because aborting in the kernel works differently than in userspace. I could have called `__assertion_failed` but I just wanted one function that is just implemented in both kernel and userspace.

asynts:
> Although, we could get away with it in `stdlib.h`. Hmm.

That would mean that the headers would no longer be commutative, `stdlib.h` would have to be included before anything else.

sv-ochis:
@asynts confirmed builds on macOS 👍

awesomekling:
I don't understand why this is necessary. The assertion failure codepath is as cold as it gets, so we can just move it out of line and call our own `assertion_failed()` which sits in some `.cpp` file that has included `<stdlib.h>`.

...Or no?

asynts:
Ugh... I hate it when I can't see a good solution for a problem.

Let's restate the problem:

  - We want to have `ASSERT`, `ASSERT_NOT_REACHED` and `TODO` macros which are callable from practically everywhere. (Including from any header file.)

  - We have to make this portable in three scenarios:

      - We are running userspace code in Serenity and want to assert, output should go to debug and stdout.

      - We are running unit tests in the host system and want to assert. In this case we have to use the abort function defined by the host libc, output should go to stdout.

      - We are running in kernel space and want to assert. Here we have to abort differently, output goes to klog.

  - We want these macros to behave differently when we are running the test suite. (The output messages have to include `FAIL` to be picked up by CMake.)

  - Aborting execution works differently in userspace and kernel.

      - We already have an abort function which is implemented either in kernel or userspace, but we can not forward declare it because the host libc can be a bit naughty.

I am not sure how to solve this, I'll have to think about this a bit more. For now you can just merge #3760 to unbreak the Mac OS users. I'll reopen this when I got an idea.

asynts:
I am just doing what @bugaevc suggested here: [#3760 (comment)](https://github.com/SerenityOS/serenity/pull/3760#issuecomment-708257134).

PR #3724, opened 2020-10-08
LibJS: break or continue with nonexistent label is a syntax error

### LINKED ISSUE #3714: LibJS: break/throw statement with nonexistent label should be a syntax error
Serenity:
```js
> for (let i = 0; i < 10; ++i) { continue loop; }
undefined
> for (let i = 0; i < 10; ++i) { break loop; }
undefined
```

Spidermonkey: 
```js
> for (let i = 0; i < 10; ++i) { continue loop; }
Uncaught SyntaxError: label not found
> for (let i = 0; i < 10; ++i) { break loop; }
Uncaught SyntaxError: label not found
```

mattco98:
Fixes #3714.

### DIFF
--- a/AK/StringView.h
+++ b/AK/StringView.h
@@ -187,6 +187,11 @@ private:
     size_t m_length { 0 };
 };
 
+template<>
+struct Traits<StringView> : public GenericTraits<String> {
+    static unsigned hash(const StringView& s) { return s.hash(); }
+};
+
 }
 
 using AK::StringView;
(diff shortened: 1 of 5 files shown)


AK/StringView.h
```diff
@@ -187,6 +187,11 @@ class StringView {
     size_t m_length { 0 };
 };
 
+template<>
+struct Traits<StringView> : public GenericTraits<String> {
+    static unsigned hash(const StringView& s) { return s.hash(); }
+};
+
```
linusg, line 194: Let's maybe put this in a separate commit?!

awesomekling:
Note that using `StringView` as a hash key may be less optimal than `String` since `String` will cache its hash while `StringView` may have to recompute it every time it's requested :)

mattco98:
Ahh interesting, didn't know that! I was just trying to avoid the `String` copies.

PR #3725, opened 2020-10-08
LibGUI: Make arrow-left/right with selection move cursor to edge of selection

nico:
This matches behavior on Linux, macOS, Windows, and it makes it possible
to hit ctrl-L arrow-right ctrl-backspace to edit the last component of
an URL in browser, or of the current path in File Manager.

Also make it so that ctrl-left/right does a word movement that starts
at the edge of the selection. This matches Linux and macOS, but not
Windows (which instead does a word movement relative to the cursor,
not the selection edge).


awesomekling:
Oh this makes so much sense

PR #3726, opened 2020-10-08
LibJS: Fix return statements not working properly in loops

### LINKED ISSUE #3604: LibJS: Returning from inside a loop is broken?
```js
function foo() {
    for (let i = 0; i < 10; ++i) {
        if (i == 4) return "FOO!"
    }
    return "BAR!"
}
```

`foo()` returns `undefined`:
```
> function foo() {
>     for (let i = 0; i < 10; ++i) {
>         if (i == 4) return "FOO!"
>     }
>     return "BAR!"
> }
undefined
> foo()
undefined
```

mattco98:
And as a bonus: return statements did nothing in while loops! So previously `while (true) { return }` would just loop forever. No clue how that one managed to go unnoticed until now lol.

Fixes #3604

### DIFF
--- a/Libraries/LibJS/AST.cpp
+++ b/Libraries/LibJS/AST.cpp
@@ -269,7 +269,7 @@ Value WhileStatement::execute(Interpreter& interpreter, GlobalObject& global_obj
                 interpreter.vm().stop_unwind();
                 break;
             } else {
-                return js_undefined();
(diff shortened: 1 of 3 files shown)


awesomekling:
> So previously `while (true) { return }` would just loop forever.

Oopsie!

PR #3727, opened 2020-10-08
Web::OutOfProcessWebView all the things

linusg:
Well, not *all* the things, but quite a few. Inspired by TextEditor crashing because of some LibWeb thing - not anymore. :^)


awesomekling:
Love it

PR #3728, opened 2020-10-08
FileManager: Permission checkboxes moved to a new tab

kessejones:
The checkboxes for changing permissions are now on a separate tab.


awesomekling:
Why though?

brendalf:
I like the ideia of moving to another table. We can start adding new features in the properties window, like details, size and etc.

kessejones:
I think it gets more organized like that.
We can add a SelectBox on this new tab to change the group as well.

tomuta:
I would argue that most of the time the properties are accessed to view/change permissions. If they're not on the initial page, that's yet another click. In my opinion we should strive to reduce the amount of clicks for most tasks (which is why I quite liked Windows 7 back in the day).

awesomekling:
Agree with @tomuta, this hides information and functionality between an unnecessary click. Let's not do this (unless there's a good reason, like having way too much information that we can't fit)

PR #3730, opened 2020-10-09
LibWeb: Move bulk of SVG layout logic into Layout nodes

mattco98:
This PR introduces layout nodes for each SVG element type, and moves all of the painting logic into those layout nodes. This brings the SVG API more inline with the rest of LibWeb.

Also, the inspector now shows the correct position and size for each SVG element:

(not sure why some of the elements don't have text in the inspector, but that's a problem for another PR)

This is a big change and it took me a long time to figure out how to do this, so I've probably not done it in the best way possible. I'm a LibWeb noob so I'd love all the feedback I can get to make this as good as possible.  Once I get this all figured out, I can finally start adding more SVG features :)

### DIFF
--- a/Libraries/LibWeb/Layout/LayoutSVGPath.cpp
+++ b/Libraries/LibWeb/Layout/LayoutSVGPath.cpp
@@ -1,0 +1,88 @@
+/*
+ * Copyright (c) 2020, Matthew Olsson <matthewcolsson@gmail.com>
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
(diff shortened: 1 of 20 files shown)


Libraries/LibWeb/Layout/LayoutSVGPath.cpp
```diff
@@ -0,0 +1,88 @@
...
+namespace Web {
+
+LayoutSVGPath::LayoutSVGPath(DOM::Document& document, SVG::SVGPathElement& element, NonnullRefPtr<CSS::StyleProperties> properties)
+    : LayoutSVGGraphics(document, element, properties)
+{
+}
+
+void LayoutSVGPath::layout(LayoutNode::LayoutMode mode)
+{
+    auto& bounding_box = downcast<SVG::SVGPathElement>(node()).get_path().bounding_box();
```
awesomekling, line 40: You can add a helper to `LayoutSVGPath` avoid the downcast:
```cpp
SVG::SVGPathElement& node() { return downcast<SVG::SVGPathElement>(LayoutSVGGraphics::node()); }
```

awesomekling:
Great work! And a much more flexible design, no longer constrained to life inside a bitmap 😄 
Please configure your editor to add a newline at the end of files.

mattco98:
> Please configure your editor to add a newline at the end of files.
I feel like you've told me this at least 10 times, and I also feel like I configure it every time... I should probably add it to `check-style.sh` one of these days

PR #3732, opened 2020-10-09
LibWeb: In the HTML tokenizer, pretty up ON_WHITESPACE a tiny bit

nico:
No behavior change.

--

I watched the "Let's write a HTML parser!" video this morning. It is one of my favorites so far, really nicely done.


awesomekling:
😄

PR #3733, opened 2020-10-09
CMake: Remove file globbing in AK/Tests

ldm5180:
Problem:
- File globbing is performed at the time of build system
  generation. Any files which are not there at that time are not
  included. So, when a new file is added it is not built unless the
  build system is recreated.

Solution:
- Remove globbing from AK/Tests directory in favor of explicitly
  listing the files.

### DIFF
--- a/AK/Tests/CMakeLists.txt
+++ b/AK/Tests/CMakeLists.txt
@@ -1,4 +1,36 @@
-file(GLOB AK_TEST_SOURCES "*.cpp")
+set(AK_TEST_SOURCES
+    TestArray.cpp
+    TestAtomic.cpp
+    TestBase64.cpp
+    TestBinarySearch.cpp
+    TestBitmap.cpp
+    TestByteBuffer.cpp
+    TestChecked.cpp
+    TestCircularDeque.cpp
+    TestCircularDuplexStream.cpp
+    TestCircularQueue.cpp
+    TestDistinctNumeric.cpp
+    TestEndian.cpp
+    TestFormat.cpp
+    TestHashMap.cpp
+    TestJSON.cpp
+    TestLexicalPath.cpp
+    TestMemoryStream.cpp
+    TestNonnullRefPtr.cpp
+    TestNumberFormat.cpp
+    TestOptional.cpp
+    TestQueue.cpp
+    TestRefPtr.cpp
+    TestSpan.cpp
+    TestString.cpp
+    TestStringUtils.cpp
+    TestStringView.cpp
+    TestTypedTransfer.cpp
+    TestURL.cpp
+    TestUtf8.cpp
+    TestVector.cpp
+    TestWeakPtr.cpp
+)
 
 foreach(source ${AK_TEST_SOURCES})
     get_filename_component(name ${source} NAME_WE)


nico:
Is the get_filename_component() still needed with this? Can we just add the sources to add_executable directly?

asynts:
If you just run `cmake ..` when adding a new file, it should find the new file, at least it does in my experience.

Explicitly listing every single file means that you can easily forget to tell CMake about the new file you just added.

ldm5180:
> If you just run `cmake ..` when adding a new file, it should find the new file, at least it does in my experience.
> 
> Explicitly listing every single file means that you can easily forget to tell CMake about the new file you just added.

@asynts The cmake community has deemed this a bit of an antipattern. There are a number of problems associated with globbing. One of the cmake devs included details in a talk at C++Now conference in 2016 https://www.youtube.com/watch?v=bsXLMQ6WgIk

ldm5180:
> Is the get_filename_component() still needed with this? Can we just add the sources to add_executable directly?

@nico Yes, that part is still needed to strip the `.cpp` from the name in order to create the name of the executable.

asynts:
From the CMake documentation about [file][1]:

  > **Note:** We do not recommend using `GLOB` to collect a list of source files from your source tree. If no CMakeLists.txt file changes when a source is added or removed then the generated build system cannot know when to ask CMake to regenerate. The `CONFIGURE_DEPENDS` flag may not work reliably on all generators, or if a new generator is added in the future that cannot support it, projects using it will be stuck. Even if `CONFIGURE_DEPENDS` works reliably, there is still a cost to perform the check on every rebuild. 

However, all the problems that they mention don't affect us. `ninja` and `make` both support this option, so let's just add `CONFIGURE_DEPENDS` instead.

  [1]: https://cmake.org/cmake/help/latest/command/file.html#glob

nico:
Sure, the first problem affects us, you have to manually rerun cmake when you add a file in this dir. There's other reasons why globbing is a bad idea, but we're also self-consistent about this: most of our build files don't glob.

asynts:
> [...] you have to manually rerun cmake when you add a file in this dir. [...]

This is just the default behavior, we can replace

~~~cmake
file(GLOB SOURCES *.cpp)
~~~

with

~~~cmake
file(GLOB SOURCES CONFIGURE_DEPENDS *.cpp)
~~~

Then Ninja or Make will re-evaluate the glob when you try to build. The only issue is that there might be other generators (other than Ninja or Make) that are not capable of this, but I don't consider this an issue.

ldm5180:
The documentation says that it does come with some cost to perform on each rebuild. This doesn't hurt much now because the number of files is small. Maybe I am missing something. What is the reason to do something different than following the best practices?

I am new to this community, so please help me out in understanding. I don't mean to be inflammatory.

asynts:
> [...] This doesn't hurt much now because the number of files is small. [...]

If we do get performance problems because of globs being re-evaluated all the time, we could make that change, but currently it seems unnecessary.

  > [...] What is the reason to do something different than following the best practices? [...]

A lot of these "best practices" are meant for projects where you can't easily change stuff. But with Serenity this is different, it's a very monolithic code base meaning that pretty much no one depends on it.

---

This is really just my opinion, Kling makes the decisions here. (Your message sounded a lot like you though I was "overruling" you, that is not the case.)

ldm5180:
@asynts I didn't think I was being overruled at all. Just making sure I am contributing in a way that is desired by the community.

nico:
Here's the full list of tradeoffs for and against using globs that I know about:

Pros of using globs:
- Don't have to explicitly list all cpp files in a directory in a CMakeLists file
- Can't forget to add cpp files to CMakeLists when adding new cpp files
-- This is especially useful for tests since it's easy to miss adding a new test cpp file

Cons of using globs:
- The generator (ie cmake) generally can't know that it needs to re-run when a new file is added, you have to rerun it manually somehow (explicit `cmake` call, touch a cmakelists file, ...) [1]. This makes the behavior of cmakelists with globs a bit surprising at times, unless you know about this.
- It's more implicit, in that you can't grep cmakelist files to see where a file is referenced
- git keeps untracked files around in the tree, so if you add a new file but forget to `git add` it, then switch branches, then the glob-based approach will silently keep the file in the build, while the explicit source list approach at least shows that you have a pending change in `git diff`. (This sounds obscure, but it's something we've run into surprisingly often in practice.)
- For large projects, evaluating globs makes the generator noticeably slower

Most larger projects I know of ban use of globs for these reasons -- the wins in predictability and understandability outweigh the convenience. (I'm not saying this is what serenity should do -- it's a smaller project and reasonable people can disagree here imho. This is just trying to make the tradeoff explicit.)


1: The CONFIGURE_DEPENDS suggestion above can't work in general. In practice, most OSs update the mtime of a directory node when a file is created in a directory, but iirc not all of them do, and I don't know if cmake adds directory paths to the configure_depends list for globs (...and all subdirs for recursive globs) -- and even if it does, this doesn't work on all systems iirc (maybe not on windows).

ldm5180:
To add another con to what @nico said... When updating to the upstream master after a file has been added, the developer needs to know to run cmake, otherwise the new file is not going to be built.

asynts:
On Linux:

~~~none
$ mkdir -p foo/bar
$ stat -c %y foo foo/bar
2020-10-13 19:52:43.113334461 +0200
2020-10-13 19:52:43.113334461 +0200
$ touch foo/bar/baz.cpp
$ stat -c %y foo foo/bar
2020-10-13 19:52:43.113334461 +0200
2020-10-13 19:53:04.483334539 +0200
~~~

So if I create a file then the modification time of the containing directory changes, but this is not propagated further. Now, is this enough?

This seems to suffice for both Ninja and Make, I created the following `CMakeLists.txt`:

~~~cmake
cmake_minimum_required(VERSION 3.18)
project(Example)

file(GLOB_RECURSE SOURCES CONFIGURE_DEPENDS src/*.cpp)

add_library(LibExample ${SOURCES})
~~~

Here is a directory tree:

~~~none
$ tree .
.
├── CMakeLists.txt
└── src
    └── foo
        └── bar.cpp

2 directories, 2 files
~~~

Now I do the following checks:

~~~none
$ mkdir Ninja && cd Ninja
$ cmake .. -GNinja
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /usr/bin/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Configuring done
-- Generating done
-- Build files have been written to: /tmp/tmp.PeibzJPNGJ/Ninja
$ ninja
[2/2] Linking CXX static library libLibExample.a
$ touch ../src/foo/baz.cpp
$ ninja
-- GLOB mismatch!
[1/2] Re-running CMake...
-- Configuring done
-- Generating done
-- Build files have been written to: /tmp/tmp.PeibzJPNGJ/Ninja
[2/2] Linking CXX static library libLibExample.a
$ rm ../src/foo/baz.cpp
~~~

~~~none
$ mkdir Make && cd Make
$ cmake .. -G"Unix Makefiles"
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /usr/bin/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Configuring done
-- Generating done
$ make
Scanning dependencies of target LibExample
[ 50%] Building CXX object CMakeFiles/LibExample.dir/src/foo/bar.cpp.o
[100%] Linking CXX static library libLibExample.a
[100%] Built target LibExample
$ touch ../src/foo/baz.cpp
$ make
-- GLOB mismatch!
-- Configuring done
-- Generating done
-- Build files have been written to: /tmp/tmp.PeibzJPNGJ/Make
Scanning dependencies of target LibExample
[ 33%] Building CXX object CMakeFiles/LibExample.dir/src/foo/baz.cpp.o
[ 66%] Linking CXX static library libLibExample.a
[100%] Built target LibExample
~~~

In both cases, the build tool prints "GLOB mismatch!" and rebuilds the new file. So this seems to be fine on Linux.

Can somebody run this on a Mac and on Windows?

nico:
I just remembered something I forgot to say in https://github.com/SerenityOS/serenity/pull/3733#issuecomment-707827051 : One downside of depending on the mtime of the directory (what we currently do) is that many editors, when saving files, first write to a temp file next to the final location, and then rename it to the actual file name (to save atomically, I think). This means every file save creates a file, which updates directory mtime in addition to saved file's mtime, which means the generator (cmake) needs to re-run on every save instead of only when new files are added.

ldm5180:
Resurrecting this one. I have noticed the same thing as @nico where cmake is getting re-run very frequently, even when I don't think things have changed in a way resulting in the need.

I have updated this PR. Any new thoughts?

PR #3734, opened 2020-10-09
Misc: String formatting fixes

linusg:


awesomekling:
D'oh!

nico:
Maybe the appendf family should assert if there's not a single % in the string and the appendff family if there's not a single {, so that things at least crash of you use the wrong one?

asynts:
The format stuff should crash when you do pretty much anything that isn't right. The only thing that I am aware of is that you can give unused parameters which is probably not ideal.

This assertion should trigger in that case if I am not mistaken:

https://github.com/SerenityOS/serenity/blob/fcd263f17ba73e44fdda6800cd2f7a3a8ee03f54/AK/Format.cpp#L84

Issue #3063, opened 2020-08-09
Filesystem is half a gigabyte: Is it time for shared objects?


BenWiederhake:
So our filesystem has grown a lot. Currently it looks like this:
```
~/workspace/serenity/Build$ dusch
12K	./mod
304K	./res
4.5M	./boot
52M	./usr
446M	./bin
503M	total
```
(`dusch` is just an alias I like to use to order everything by human-readable size. The long form is `find -maxdepth 1 ! -name . -print0 | xargs -0 du -sch0 | sort -hz | tr "\0" "\n"`, in case you're interested.)

That's half a gigabyte, just for various small applications!

(Also: Whoa, our kernel is only 4.5 MB? Awesome!)

```
~/workspace/serenity/Build/Libraries$ du -h ./LibWeb/libweb.a ./LibGUI/libgui.a
11M	./LibWeb/libweb.a
13M	./LibGUI/libgui.a
```

Because we link LibWeb and LibGUI into most applications, many binaries carry a copy of these around. LibWeb and especially LibGUI are used quite often, so any gains by LTO code elimination are offset by the sheer number of programs that link statically against (=carry a copy of) them. So most of those 446M are basically copies and copies of the same static libraries, in each binary.

A solution would be dynamic linking. For now, making only LibGUI dynamically The "manual kind" as demonstrated in `LinkDemo` already works nicely! However, it's "manual", and wrapping LibWeb or LibGUI into C-style functions sounds like a bad idea.

@awesomekling and everyone else: What do you think about this?
CC @ADKaster, because you implemented LinkDemo, and maybe I can interest you in writing more on dynamic linking? :)

ADKaster:
There's been some WIP work in this area already by myself and @itamar8910 

Issue:
https://github.com/SerenityOS/serenity/issues/1747

WIP MRs
https://github.com/SerenityOS/serenity/pull/1087
https://github.com/SerenityOS/serenity/pull/2461
https://github.com/SerenityOS/serenity/pull/2838

Always seems to get kind of close and then other things come up that cause the contributor to lose interest or get focused on other events. I know I've been focused on work and my master's recently so haven't been contributing much. 

If you'd like to pick it up, go for it! I don't have anything WIP and Itamar just closed his latest MR saying he was going to focuson on something else.

[...]

PR #3738, opened 2020-10-10
Dynamic Loader

itamar8910:
So this is another go at a dynamic loader (previous one - #2838). Relevant recent issue - #3063.
cc @ADKaster 

Notable differences from the previous version:
- TLS allocation from userspace.
Before loading the required libraries, the loader iterates them to see how much TLS we need, and requests the Kernel to allocate it.
To simplify things, we currently compile with `-ftls-model=initial-exec` which allows us to allocate only a single chunk of TLS. It means we currently can't support calling `dlopen` at runtime for libraries that need TLS, but I think it's okay for now.
- The loader compiles with the sources of `AK`, `LibELF` and `LibC` (Ideally we wouldn't want to use the whole of `LibC` in the loader, but turns out AK needs quite a lot of LibC stuff). The loader relocates itself on startup to be able to use these libraries.

See `DynamicObjectDemo` for an example shared object executable. It uses `SampleLib` and `LibC`, which both have TLS.

The next step would be trying to load something more complex, and see how the loader handles it.

Comments/thoughts on this are of course welcome.

Screenshot with memory map:
dynamic-loader


Applications/Loader/main.cpp
```diff
@@ -0,0 +1,252 @@
...
+    loader->m_global_symbol_lookup_func = global_symbol_lookup;
+
+    g_loaders.set(name, loader);
+
+    g_current_tls_offset += loader->tls_size();
+}
+
+static void map_library(const String& name)
+{
+    // TODO: Do we want to also look for libs in other paths too?
```
ADKaster, line 134: Room here for LD_LIBRARY_PATH, RUNPATH, RPATH, ... But definitely not first pass.

Applications/Loader/main.cpp
```diff
@@ -0,0 +1,252 @@
...
+{
+    dbg() << "mapping dependencies for: " << name;
+    auto lib = g_loaders.get(name).value();
+    lib->for_each_needed_library([](auto needed_name) {
+        dbg() << "needed library: " << needed_name;
+        String library_name = get_library_name(needed_name);
+
+        if (!g_loaders.contains(library_name)) {
+            map_library(library_name);
+            map_dependencies(library_name);
```
ADKaster, line 156: This looks quite recursive, might be an issue way wayyyy in the future for ports with deep library requirements.

Applications/Loader/math.cpp
```diff
@@ -0,0 +1,191 @@
...
+{
+    const int bits_in_dword_m1 = (int)(sizeof(int64_t) * 8) - 1;
+    int64_t s_a = a >> bits_in_dword_m1;                   // s_a = a < 0 ? -1 : 0
+    int64_t s_b = b >> bits_in_dword_m1;                   // s_b = b < 0 ? -1 : 0
+    a = (a ^ s_a) - s_a;                                   // negate if s_a == -1
+    b = (b ^ s_b) - s_b;                                   // negate if s_b == -1
+    s_a ^= s_b;                                            // sign of quotient
+    return (__udivmoddi4(a, b, (uint64_t*)0) ^ s_a) - s_a; // negate if s_a == -1
+}
+int64_t __moddi3(int64_t a, int64_t b)
```
ADKaster, line 180: Are these normally provided by libgcc_s?

Applications/Loader/misc.cpp
```diff
@@ -0,0 +1,39 @@
...
+
+extern "C" {
+const char* __cxa_demangle(const char*, void*, void*, int*)
+{
+    dbg() << "WARNING: __cxa_demangle not supported";
+    return "";
+}
+
+// FIXME: Is this correct?
+void* __dso_handle = nullptr;
```
ADKaster, line 38: I don't think it's quite correct? That's supposed to be a secret linker symbol. If anything, it needs a hearty `__attribute__((visibility("hidden"))` like in Kernel/init.cpp (or wherever that moved to)

Toolchain/Patches/gcc.patch
```diff
@@ -121,14 +121,14 @@ index 000000000..2692b8e57
...
++#define STARTFILE_SPEC "%{!shared:crt0.o%s} crti.o%s %{shared|pie:crtbeginS.o%s; :crtbegin.o%s}"
 +
 +/* Files that are linked after user code. */
 +#undef ENDFILE_SPEC
-+#define ENDFILE_SPEC "%{shared:crtendS.o%s; :crtend.o%s} crtn.o%s"
++#define ENDFILE_SPEC "%{shared|pie:crtendS.o%s; :crtend.o%s} crtn.o%s"
 +
 +#undef LINK_SPEC
-+#define LINK_SPEC "%{shared:-shared} %{static:-static} %{!shared: %{!static: %{rdynamic:-export-dynamic}}}"
++#define LINK_SPEC "%{shared:-shared} %{static:-static} %{!shared: %{!static: %{rdynamic:-export-dynamic} -dynamic-linker /usr/lib/Loader.so}}"
```
ADKaster, line 131: Don't forget to update the specs for the GCC port :)

Libraries/LibELF/DynamicLoader.cpp
```diff
@@ -57,36 +59,70 @@ static void* mmap_with_name(void* addr, size_t length, int prot, int flags, int
 
 namespace ELF {
 
-static bool s_always_bind_now = false;
+static bool s_always_bind_now = true;
```
ADKaster, line 62: Issues with PLT relocations? Those are annoyingly tricky...

Kernel/Syscalls/mmap.cpp
```diff
@@ -417,4 +417,46 @@ int Process::sys$munmap(void* addr, size_t size)
     return -EINVAL;
 }
 
+void* Process::sys$allocate_tls(size_t size)
+{
+    REQUIRE_PROMISE(stdio);
+
+    dbg() << "allocate TLS: " << size;
```
ADKaster, line 424: This syscall is clever, way better of a solution than I was thinking. I think other systems do the entirety of TLS allocation in userspace, but they don't have the privelidge of starting from scratch. Worth some thought whether doing full userspace only is worth it. One thing that will be annoying in the future is that for C++11 `thread_local`, GCC emits calls to `___tls_getaddr` I think. So, needs at least some help from the loader/libc to get the correct TLS location.

ADKaster (commented):
Looking good! I had some WIP work I spent like an hour or two on trying to tackle this (again...) starting from Auxilliary vector, but it kept crashing and I put it back on the shelf.

Libraries/LibELF/DynamicLoader.cpp
itamar8910: They sure are, I set this flag so we can not worry about these right now and make things easier to debug.
I'll switch back to doing lazy binding once things stabilize a bit, and it should only be a matter of flipping this flag and passing `RTLD_LAZY` to `ELF::DynamicLoader` I think, because you've already implemented everything needed there :smiley:

itamar8910:
So apparently upgrading to gcc 10.2.0 on master broke the build of dylib stuff. g++ complained about not having `-lc` and `-lm` for shared objects.
I added the cmake function `serenity_libc_shared`  to register the shared version of LibC as `-lc`, but now I get the following error when building:
```
serenity/Toolchain/Local/lib/gcc/i686-pc-serenity/10.2.0/../../../../i686-pc-serenity/bin/ld: Demos/DynamicObject/DynamicObjectDemo: hidden symbol `__divdi3' in serenity/Toolchain/Local/lib/gcc/i686-pc-serenity/10.2.0/libgcc.a(_divdi3.o) is referenced by DSO
```
So looks like we need a `.so` version of `libgcc`? I'm not sure how we can generate it.

ADKaster:
> So looks like we need a `.so` version of `libgcc`? I'm not sure how we can generate it.

I think there should be a way to do that in the patch file? I thought there was some stuff in there about how to build libgcc. I know at least on my ubuntu machines they have a `libgcc_s.so`.

itamar8910:
Okay, so we now build `libgcc_s.so`.
> I think there should be a way to do that in the patch file? 

Yeah I had to add the `t-slibgcc` target (and for some reason also patch its makefile).

But we're now getting weird exception-related compilation errors from `libstdc++.a`, so i'll need to look into those.
Example: 
```
serenity/Toolchain/Local/lib/gcc/i686-pc-serenity/10.2.0/../../../../i686-pc-serenity/bin/ld: serenity/Toolchain/Local/lib/gcc/i686-pc-serenity/10.2.0/../../../../i686-pc-serenity/lib/libstdc++.a(eh_terminate.o): in function `__cxxabiv1::__terminate(void (*)())':
serenity/Toolchain/Tarballs/gcc-10.2.0/libstdc++-v3/libsupc++/eh_terminate.cc:44: undefined reference to `_Unwind_Resume'
```

ADKaster:
> But we're now getting weird exception-related compilation errors from `libstdc++.a`, so i'll need to look into those.
> Example:
> 
> ```
> serenity/Toolchain/Local/lib/gcc/i686-pc-serenity/10.2.0/../../../../i686-pc-serenity/bin/ld: serenity/Toolchain/Local/lib/gcc/i686-pc-serenity/10.2.0/../../../../i686-pc-serenity/lib/libstdc++.a(eh_terminate.o): in function `__cxxabiv1::__terminate(void (*)())':
> serenity/Toolchain/Tarballs/gcc-10.2.0/libstdc++-v3/libsupc++/eh_terminate.cc:44: undefined reference to `_Unwind_Resume'
> ```

libgcc is supposed to include the Unwind ABI methods (the ones that I documented in LibUnwind/UnwindBase.h and then did nothing with yet :p). But, it seems odd to have libstdc++ expecting those methods if we compile everything with -fno-exceptions.  Unless libstdc++ isn't compiled that way?

Applications/Loader/math.cpp
itamar8910: They are, yes

itamar8910:
So, an update - we now use shared libraries everywhere, and the system boots and runs! :tada: 
Anybody that wants to try this out, remember to rebuild the Toolchain as there are some changes to it in this branch.

Some things I noticed that currently do not work after transitioning to shared libs:

- Crash reports - There are no symbols in backtraces.
There was some talk on IRC about having some userspace crash daemon process that gets notified about crashes and works with the loader to provide nice crash dumps.
- Profiler - Doesn't recognize symbols
- Userspace Emulator: Currently crashes because it wants to use `__cxa_demangle`, which we currently do not support. After resolving this, there is more work to be done to get it to know how to load shared libraries.

If anybody finds any other regressions, please comment here.

itamar8910:
I continued @Quaker762's work from #1917 to add core dump functionality in the kernel. Core dumps are written to the `/tmp/coredump/` directory when a process crashes.
I now plan to work on a `CrashDaemon` service which will listen for new coredump files and generate nice backtraces for us (+ maybe show a "this process crashed" dialog?)

Quaker762:
> I now plan to work on a `CrashDaemon` service which will listen for new coredump files and generate nice backtraces for us (+ maybe show a "this process crashed" dialog?)

Nice. This was my initial plan for that PR. Something like the Ubuntu Apport Window that pops up when a crash occurs.

itamar8910:
`CrashDaemon` can now print backtraces for new coredump files.
Example backtrace output: (Note that name demangling is currently disabled because we don't support it yet with shared libraries)
```
CrashDaemon(18): Backtrace for thread #0, tid=26
0x02943b24: [libc.so] kill +28
0x02943ba0: [libc.so] raise +32
0x0293c951: [libc.so]  +25
0x022a1fd4: [/bin/crash] _ZNK2AK8FunctionIFN5Crash7FailureEvEE15CallableWrapperIZ4mainEUlvE2_E4callEv +0
0x022a37cf: [/bin/crash] _ZZN5Crash3runENS_7RunTypeEENKUlvE_clEv +35
0x022a3892: [/bin/crash] _ZN5Crash3runENS_7RunTypeE +96
0x022a2ccb: [/bin/crash] main +1617
0x022a3a78: [/bin/crash] _start +36
```

itamar8910:
Okay, we now support c++ name demangling in libc.so.
Example backtrace for a browser crash:
```
CrashDaemon(18): Backtrace for thread #0, tid=25
0x051870b4: [libc.so] kill +28
0x05187130: [libc.so] raise +32
0x0517f499: [libc.so]  +25
0x0517fb9b: [libc.so] _start +0
0x05b977e7: [libweb.so] Web::HTML::HTMLTokenizer::HTMLTokenizer(AK::StringView const&, AK::String const&) +349
0x05b62542: [libweb.so] Web::HTML::HTMLDocumentParser::HTMLDocumentParser(AK::StringView const&, AK::String const&) +114
0x05bb5e18: [libweb.so] Web::FrameLoader::create_document_from_mime_type(AK::ByteBuffer const&, AK::URL const&, AK::String const&, AK::String const&) +90
0x05bb8119: [libweb.so] Web::FrameLoader::resource_did_load() +449
0x05bba8ad: [libweb.so] Web::Resource::for_each_client(AK::Function<void (Web::ResourceClient&)>) +329
0x05bbaf15: [libweb.so] Web::Resource::did_load(AK::Badge<Web::ResourceLoader>, AK::ByteBuffer const&, AK::HashMap<AK::String, AK::String, AK::CaseInsensitiveStringTraits> const&) +1257
0x05bbbbc4: [libweb.so] AK::Function<void (AK::ByteBuffer const&, AK::HashMap<AK::String, AK::String, AK::CaseInsensitiveStringTraits> const&)>::CallableWrapper<Web::ResourceLoader::load_resource(Web::Resource::Type, Web::LoadRequest const&)::{lambda(auto:1&, auto:2&)#1}>::call(AK::ByteBuffer const&, AK::HashMap<AK::String, AK::String, AK::CaseInsensitiveStringTraits> const&) const +60
0x05bbc58e: [libweb.so] AK::Function<void (bool, AK::ByteBuffer const&, AK::RefPtr<AK::SharedBuffer, AK::RefPtrTraits<AK::SharedBuffer> >, AK::HashMap<AK::String, AK::String, AK::CaseInsensitiveStringTraits> const&, AK::Optional<unsigned int>)>::CallableWrapper<Web::ResourceLoader::load(Web::LoadRequest const&, AK::Function<void (AK::ByteBuffer const&, AK::HashMap<AK::String, AK::String, AK::CaseInsensitiveStringTraits> const&)>, AK::Function<void (AK::String const&)>)::{lambda(bool, AK::ByteBuffer const&, auto:1, auto:2&, auto:3)#4}>::call(bool, AK::ByteBuffer const&, AK::RefPtr<AK::SharedBuffer, AK::RefPtrTraits<AK::SharedBuffer> >, AK::HashMap<AK::String, AK::String, AK::CaseInsensitiveStringTraits> const&, AK::Optional<unsigned int>) const +240
0x058f20a9: [libprotocol.so] Protocol::Download::did_finish(AK::Badge<Protocol::Client>, bool, AK::Optional<unsigned int>, unsigned int, int, IPC::Dictionary const&) +773
0x058ecb91: [libprotocol.so] Protocol::Client::handle(Messages::ProtocolClient::DownloadFinished const&) +275
0x058ee255: [libprotocol.so] ProtocolClientEndpoint::handle(IPC::Message const&) +69
0x058f1b3e: [libprotocol.so] IPC::Connection<ProtocolClientEndpoint, ProtocolServerEndpoint>::handle_messages() +128
0x058f1ba1: [libprotocol.so] AK::Function<void ()>::CallableWrapper<IPC::Connection<ProtocolClientEndpoint, ProtocolServerEndpoint>::Connection(ProtocolClientEndpoint&, AK::NonnullRefPtr<Core::LocalSocket>)::{lambda()#2}>::call() const +39
0x0524ccba: [libcore.so] Core::Notifier::event(Core::Event&) +78
0x0524d14f: [libcore.so] Core::Object::dispatch_event(Core::Event&, Core::Object*) +89
0x0524291e: [libcore.so] Core::EventLoop::pump(Core::EventLoop::WaitMode) +434
0x05242bc7: [libcore.so] Core::EventLoop::exec() +55
0x05750514: [libgui.so] GUI::Application::exec() +50
0x05cda272: [/bin/Browser] main +2348
0x05ced1b6: [/bin/Browser] _start +36
```

itamar8910:
Screenshot of the profile viewer with shared libs:
profiler-shared-libs2

Quaker762:
@itamar8910 Looking damn good. Can't wait to take this for a spin so I can port Quake II to the system (which loads the renderer at runtime via mapping functions from dynamic libs).

itamar8910:
An update:
`UserspaceEmulator` now works with dynamically loaded programs - including malloc tracing and symbolication in backtraces.

Screenshot of running `id` in the emulator (after intentionally inserting a memory leak):
ue-id-leak

So, unless I missed something, we're at a stage where everything in the core system works with shared libraries. Hooray!

I think this PR is really close to being ready for review, but first the following things need to be fixed:
- Ports (Currently fail to build with the updated toolchain patch)
- Lagom

I don't feel like working on these at the moment, so if anybody feels like doing that it would be great.
Otherwise I'll probably get to fixing these things soon-ish.

Applications/Loader/math.cpp
ADKaster: Can this file be removed now that you added libgcc_s.so to the build spec of libgcc? i.e. are there methods in here that the loader needs before it can possibly load libgcc_s?

DevTools/UserspaceEmulator/Emulator.cpp
```diff
@@ -125,54 +165,70 @@ void Emulator::setup_stack(const Vector<String>& arguments, const Vector<String>
...
+
+    ELF::Image executable_elf((const u8*)mapped_executable.data(), mapped_executable.size());
+
+    if (!executable_elf.is_dynamic()) {
+        // FIXME: Support static objects
+        ASSERT_NOT_REACHED();
+    }
+
+    String interpreter_path;
+    executable_elf.for_each_program_header([&interpreter_path](const ELF::Image::ProgramHeader& program_header) {
```
ADKaster, line 182: I think ELF::validate_program_headers will do this for you, and do some extra checking along the way as well.

itamar8910: That's nice, thanks!

itamar8910:
After fixing some things with the toolchain, ports are now OK.
I see that the CI failed, I'll look at it later.

itamar8910:
Okay, I marked this as ready for review - As far as I know everything in the system now works OK with shared libs.
@awesomekling - Good luck with reviewing this huge PR :sweat_smile: (~3K lines of diff are just from regenrating the gcc patch file, though!)

Applications/CMakeLists.txt
```diff
@@ -11,6 +11,7 @@ add_subdirectory(HexEditor)
 add_subdirectory(IRCClient)
 add_subdirectory(KeyboardMapper)
 add_subdirectory(KeyboardSettings)
+add_subdirectory(Loader)
```
awesomekling, line 15: The dynamic loader is not really an application :)
Let's put this somewhere else. `Userland/DynamicLoader/` perhaps?

Kernel/CoreDump.cpp
```diff
@@ -0,0 +1,274 @@
...
+{
+}
+
+CoreDump::~CoreDump()
+{
+}
+
+RefPtr<FileDescription> CoreDump::create_target_file(const Process& process)
+{
+    static constexpr const char* coredumps_directory = "/tmp/coredump";
```
awesomekling: We should definitely move away from this eventually, so we don't have hardcoded paths in the kernel like this.
However, I think it's okay at the moment.

Libraries/LibCore/DirectoryWatcher.h
```diff
@@ -0,0 +1,61 @@
...
+#include <AK/Noncopyable.h>
+#include <Kernel/API/InodeWatcherEvent.h>
+
+namespace Core {
+
+class DirectoryWatcher {
+    AK_MAKE_NONCOPYABLE(DirectoryWatcher);
+
+public:
+    DirectoryWatcher(const LexicalPath&);
```
awesomekling, line 40: Please make this constructor explicit.
Also let's handle paths in this class as `String`. `LexicalPath` is a heavy utility class for extracting pieces from paths, not really meant as a storage unit.

itamar8910: Okay. Also fixed this for `CoreDumpReader` and `CrashDaemon`.

awesomekling:
The kernel crashes if you try to generate a coredump with the same program twice.
Easy to reproduce in the shell: "asdf<enter>asdf<enter>"
(Also, why are we generating a coredump just because some command is not found!?)

awesomekling:
This is super cool :)

Kernel/Syscalls/exit.cpp
```diff
@@ -33,8 +33,10 @@ void Process::sys$exit(int status)
 {
     cli();
 
-    if (status != 0)
+    if (status != 0) {
         dump_backtrace();
+        set_dump_core(true);
```
alimpfard, line 38: why? I don't think `exit(whatever)` needs a coredump...?

itamar8910: Yeah I tend to agree. The current behavior is dumping a backtrace, which is not very useful for userspace without the core dump that `CrashDaemon` symbolicates.
But I don't think we need a backtrace here either, so I went ahead and disabled both printing backtrace & dumping core.

Lubrsi:
DynamicObjectDemo segfaults on startup. Full trace: https://gist.github.com/Lubrsi/47d0235e2b091219b135d9f5d2fee87a

itamar8910:
@Lubrsi - Okay, fixed that.

Lubrsi:
Screenshot from 2020-12-12 23-28-33

It appears userspace symbols don't appear in SystemMonitor's stack view.

awesomekling:
I don't think we should generate core when a process is terminated due to a signal. Only in the case of `DefaultSignalAction::DumpCore`

itamar8910:
@awesomekling 
> I don't think we should generate core when a process is terminated due to a signal. Only in the case of DefaultSignalAction::DumpCore

Sure, fixed that.

@Lubrsi 
> It appears userspace symbols don't appear in SystemMonitor's stack view.

Good catch.
Looking at the way it's implemented, supporting this feature looks to be non-trivial.
The way it currently works is the kernel generates it for us when SystemMonitor reads `/proc/<pid>/stacks/<tid>`.
I see two ways of implementing this:
1. Have a way to request the kernel to generate a coredump of a process from userspace 
2. Using ptrace

However, would you consider this a blocker for merging this PR? (Perhaps kling can comment on this).
I don't know that this feature it too widely used, and personally I think we can live with this regression until someone fixes it.

Lubrsi:
Yeah, that can be fixed at a later time. I don't use it too often either, just something I noticed.

awesomekling:
Let's fix that at a later time. And we can just *not* generate symbols in ProcFS and instead make SystemMonitor do it :D

awesomekling:
Ok let's do this.

Issue #3594, opened 2020-09-24
Userland/cp: Disallow copying directories into themselves


xTibor:
**Reproduction:**
```
$ cp -r . ./test
```

**Expected result:**
Something similar that GNU does:
```
cp: cannot copy a directory, '.', into itself, './test'
```

**Current results:**
![Screenshot_20200924_164244](https://user-images.githubusercontent.com/1627292/94171579-b8a8ba80-fe91-11ea-8837-f14e5b8203ae.png)

PR #3743, opened 2020-10-11
Userland/cp: Disallow copying directories into themselves

matthewcurry:
This patch causes cp to investigate whether a directory is being copied into a subdirectory of itself. It uses realpath(3) to ensure that links do not confound detection.

This addresses [Issue #3594](https://github.com/SerenityOS/serenity/issues/3594).

### DIFF
--- a/Userland/cp.cpp
+++ b/Userland/cp.cpp
@@ -29,6 +29,7 @@
 #include <AK/StringBuilder.h>
 #include <LibCore/ArgsParser.h>
 #include <LibCore/DirIterator.h>
+#include <LibCore/File.h>
 #include <assert.h>
 #include <fcntl.h>
 #include <stdio.h>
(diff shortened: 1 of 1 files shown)


Userland/cp.cpp
```diff
@@ -174,6 +174,16 @@ bool copy_directory(String src_path, String dst_path)
         perror("cp: mkdir");
         return false;
     }
+
+    char src_rp[PATH_MAX + 1], dst_rp[PATH_MAX + 1];
+    if (realpath(src_path.characters(), src_rp) != NULL && realpath(dst_path.characters(), dst_rp) != NULL) {
```
Petelliott, line 179: use `nullptr` instead of `NULL`

matthewcurry: Noted and addressed

awesomekling (changes requested):
Let's write C++ instead of C :)
You can use `Core::File::real_path_for()` to get the real paths as `String`.

PR #3745, opened 2020-10-11
IPCCompiler: Use a helper class for code generation and don't use LogStream.

asynts:
I originally started replacing the LogStream stuff with the new format stuff, but then I noticed that it even less readable than before.

So I added a `SourceGenerator` class instead. This works a lot like the configure stuff, you just write `@something@` and it gets replaced by something.

One difference is that `SourceGenerator` can be nested, in this case the "mapping" is scoped, meaning that any changes made by `SourceGenerator::set` are discarded when the object goes out of scope.

I've chosen `@` because it is less common in source code than `{` or `}`.

---

I created a diff comparing the generated endpoint files in Serenity before and after the change, there is quite a lot of changes regarding white space, but other than that I noticed no significant differences. (I did notice significant differences at first but that should be fixed now.)

Nevertheless, I am pretty sure that this pull request introduces a bug somewhere. :innocent:

### DIFF
--- a/DevTools/IPCCompiler/main.cpp
+++ b/DevTools/IPCCompiler/main.cpp
@@ -27,6 +27,7 @@
 #include <AK/Function.h>
 #include <AK/GenericLexer.h>
 #include <AK/HashMap.h>
+#include <AK/SourceGenerator.h>
 #include <AK/StringBuilder.h>
 #include <LibCore/File.h>
 #include <ctype.h>
@@ -64,13 +65,13 @@ struct Endpoint {
 int main(int argc, char** argv)
 {
     if (argc != 2) {
-        printf("usage: %s <IPC endpoint definition file>\n", argv[0]);
+        outln("usage: {} <IPC endpoint definition file>", argv[0]);
         return 0;
     }
 
     auto file = Core::File::construct(argv[1]);
     if (!file->open(Core::IODevice::ReadOnly)) {
-        fprintf(stderr, "Error: Cannot open %s: %s\n", argv[1], file->error_string());
+        warnln("Error: Cannot open {}: {}", argv[1], file->error_string());
         return 1;
     }
 
@@ -81,7 +82,7 @@ int main(int argc, char** argv)
 
     auto assert_specific = [&](char ch) {
         if (lexer.peek() != ch)
-            warn() << "assert_specific: wanted '" << ch << "', but got '" << lexer.peek() << "' at index " << lexer.tell();
+            warnln("assert_specific: wanted '{}', but got '{}' at index {}", ch, lexer.peek(), lexer.tell());
         bool saw_expected = lexer.consume_specific(ch);
         ASSERT(saw_expected);
     };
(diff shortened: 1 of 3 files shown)


asynts:
I did another diff and found this (I forgot to fix this after doing the first diff):

~~~diff
@@ -425,6 +425,7 @@ ./DevTools/IPCCompiler/main.cpp
 
     static int static_magic() { return @endpoint.magic@; }
     virtual int magic() const override { return @endpoint.magic@; }
+    static String static_name() { return "@endpoint.name@"; }
     virtual String name() const override { return "@endpoint.name@"; }
 
     static OwnPtr<IPC::Message> decode_message(const ByteBuffer& buffer, size_t& size_in_bytes)
~~~

DevTools/IPCCompiler/main.cpp
```diff
@@ -267,31 +282,58 @@ int main(int argc, char** argv)
...
-            out() << "    virtual i32 message_id() const override { return (int)MessageID::" << name << "; }";
-            out() << "    static i32 static_message_id() { return (int)MessageID::" << name << "; }";
-            out() << "    virtual const char* message_name() const override { return \"" << endpoint.name << "::" << name << "\"; }";
-            out() << "    static OwnPtr<" << name << "> decode(InputMemoryStream& stream, size_t& size_in_bytes)";
-            out() << "    {";
-
-            out() << "        IPC::Decoder decoder(stream);";
+                message_generator.append(R"~~~(
+   typedef class @message.response_type@ ResponseType;
+)~~~");
```
alimpfard, line 299: Is it just me, or does this somehow look worse?
I kinda like the `@foobar.blahblah@`, but the use of raw strings and weird indents makes it look...odd

perhaps we could instead do something like `source_generator << "typedef class @message.response_type@ ResponseType";`, and have it magically indent for us?

Something that comes to mind is `source_generator.indented(1) << "blahblah"`
and newlines could be inserted implicitly between `<<`'s now, as they cannot meaningfully delimit anything else given this `@foo@` thing?

alimpfard: Another option I can think of is what I did with the shell formatter [here](https://github.com/SerenityOS/serenity/blob/master/Shell/Formatter.cpp#L253-L256) and [here](https://github.com/SerenityOS/serenity/blob/ac8fe3d0625ffc95f9b042e990b122ccf5abe2d3/Shell/Formatter.cpp#L76-L79)
But that assumes stuff about the semantics of the language it's generating, so probably not a great "source generator" interface

asynts: > Is it just me, or does this somehow look worse?

It's surely not just you, but I actually like it. Previously, it was difficult to "see" the file that is being generated, now it looks like two "interleaved" source files.

(I notice now, that I was a bit inconsistent with this, but more towards the end of the file I always used the multi-line string literals and did the indentation correctly. I think that really benefits helps see the generated file at a glance.)

asynts: *(Now about the previous message.)*

I think if we did code generation on a much larger scale, then a more "functional" approach would be a really good idea, this is really just me trying to be create a simple solution.

alimpfard: That's fair, I suppose.
it looks odd to me, but I guess I could get used to it :shrug:

asynts: Another way of doing this would be like Go's text template thing. ([Go Documentation][1])

It's like `std::format` but with loops and conditions. For the simple stuff we are doing here that would totally suffice, but implementing something like that would really be overkill.

  [1]: https://golang.org/pkg/text/template/#hdr-Actions

alimpfard: Now that's definitely overkill.
I don't think there's any reason to go down that route, also given that it'll end up looking like one big string without good syntax highlighting... `(include python/fstring-syntax-fight)`

asynts: I might do some variation of

~~~c++
source_generator.indent(4) << "int foo = @foo@";
~~~

I'll have to sleep over that.

asynts: I don't think the `indent(4)` stuff is really needed, since the compiler doesn't care about it and we can always run it through clang format. (The indentation only truly aids readability if it is present in the source code as-is.)

I could go with the `generator << "int foo = @foo@;` syntax though, take the following snippet for example:

~~~c++
        endpoint_generator.append(R"~~~(
            return nullptr;
        }
        i32 message_id = 0;
        stream >> message_id;
        if (stream.handle_any_error()) {
)~~~");
~~~

I could change this to one of the following snippets:

~~~c++
        endpoint_generator << R"~~~(
            return nullptr;
        }
        i32 message_id = 0;
        stream >> message_id;
        if (stream.handle_any_error()) {
)~~~";
~~~

~~~c++
        endpoint_generator << "        return nullptr";
        endpoint_generator << "        }";
        endpoint_generator << "        i32 message_id = 0;"
        endpoint_generator << "        stream >> message_id;"
        endpoint_generator << "        if (stream.handle_any_error()) {"
~~~

I don't think that improves the code a lot, although in cases where just one line is printed it might. (But for consistency I would all format them the same way.)

alimpfard: What I thought `<<` could indicate was a logical statement, so newlines and indentation would be automatically inserted.
But that makes the generated source much less visible, so maybe your raw strings _are_ the same approach

asynts:
I noticed that I was a bit inconsistent with the formatting of the append stuff, I made a few changes similar to the following:

~~~diff
@@ -224,6 +224,8 @@ ./DevTools/IPCCompiler/main.cpp
         endpoint_generator.set("endpoint.name", endpoint.name);
         endpoint_generator.set("endpoint.magic", String::number(endpoint.magic));
 
-        endpoint_generator.append("namespace Messages::@endpoint.name@ {\n");
+        endpoint_generator.append(R"~~~(
+namespace Messages::@endpoint.name@ {
+)~~~");
 
         HashMap<String, int> message_ids;
~~~

DevTools/IPCCompiler/main.cpp
asynts: I'll just leave it as it is now, but I'll keep your thoughts in the back of my head, I usually come back to this sort of stuff.

I think we are doing some code generation in LibWeb too, so when I go and eradicate the printf and LogStream stuff from there, maybe a more "abstract" approach would be better there?

awesomekling:
This looks p nice :)

PR #3748, opened 2020-10-11
LibWeb: Add element namespace and initial implementation of foreign content parsing

Lubrsi:
LibWeb: Use modern namespaces and fix clang-format comments in tag names

-----

LibWeb: Add namespace to Element

-----

LibWeb: Add initial implementation of foreign content parsing

Plus sneak in a FIXME for the list of active formatting elements
and a test for Element.namespaceURI


awesomekling:
Instead of storing the namespaces as a separate string, let's implement [qualified names](https://en.wikipedia.org/wiki/QName) as its own class. A `QualifiedName` would have a `namespace`, `prefix` and `local_name`.

And Elements would then store a qualified name internally. :)

PR #3749, opened 2020-10-11
seq: Add help option

matthewcurry:
Added a help option to the seq command that gives usage information. Further, when a user gives an incorrect argument, usage is supplied on stderr.

### DIFF
--- a/Userland/seq.cpp
+++ b/Userland/seq.cpp
@@ -30,12 +30,26 @@
 #include <string.h>
 #include <unistd.h>
 
+const char* g_usage = R"(Usage:
+    seq [-h|--help]
+    seq LAST
+    seq FIRST LAST
+    seq FIRST INCREMENT LAST
+)";
+
+static void print_usage(FILE* stream)
+{
+    fprintf(stream, g_usage);
+    return;
+}
+
 static double get_double(const char* name, const char* d_string, int* number_of_decimals)
 {
     char* end;
     double d = strtod(d_string, &end);
     if (d == 0 && end == d_string) {
-        fprintf(stderr, "%s: invalid double value \"%s\"\n", name, d_string);
+        fprintf(stderr, "%s: invalid argument \"%s\"\n", name, d_string);
+        print_usage(stderr);
         exit(1);
     }
     if (char* dot = strchr(d_string, '.'))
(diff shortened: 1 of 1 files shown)


Userland/seq.cpp
```diff
@@ -30,12 +30,23 @@
 #include <string.h>
 #include <unistd.h>
 
+static void print_usage(FILE* stream)
+{
+    fprintf(stream, "Usage:\n"
```
nico, line 35: Maybe use a raw string?

matthewcurry: What are we trying to accomplish with the raw string?

Userland/seq.cpp
```diff
@@ -56,6 +67,13 @@ int main(int argc, const char* argv[])
         return 1;
     }
 
+    for (int i = 1; i < argc; i++) {
+        if (strcmp(argv[i], "--help") == 0 || strcmp(argv[i], "-h") == 0) {
+            print_usage(stdout);
+            exit(0);
+        }
+    }
+
```
linusg, line 76: We have `Core::ArgsParser` for this :)

alimpfard: Unfortunately, `ArgsParser` cannot really deal with the...odd argument format here (yet?)

linusg: I somehow thought one could give it a custom help text but I confused that with explicit calls of `args_parser.print_usage()` (which still shows the generated help text). Nevermind!

Userland/seq.cpp
nico: Then you don't need to put `\n`s at the line ends and you can have a single string literal instead of one per line. Makes editing the text a bit easier.

matthewcurry: Would you suggest something like this?

```
    fprintf(stream, R"(Usage:
    seq [-h|--help]
    seq LAST
    seq FIRST LAST
    seq FIRST INCREMENT LAST
)"); /* Could just put fprintf(stream, '\n') on a separate line */
```

I think it makes for odd-looking code where it is, since it wouldn't follow any indention conventions. I could pull it out into a global variable, too, which would kind of fix the indention (it might not match, but it would still look ok), but it really shouldn't be a global variable since I only use it in one place.

Anyhow, tell me which approach would make you happy, and I'll do it.

asynts: You can use 

~~~none
fprintf(stream, 1 + R"(
Usage:
    seq [-h|--help]
    seq LAST
    seq FIRST LAST
    seq FIRST INCREMENT LAST
)");
~~~

but it's a bit of a hack, raw string literals are really not useful if you care about the newlines exactly.

nico: @matthewcurry, yes, what you wrote is imho easier to read and edit than many string literals. I agree that the start looks a bit weird, but

* you rarely edit the first line
* hacks like the `1 +` etc don't seem worth it

Just my personal opinion though, and doesn't matter all that much :)

matthewcurry: I used a raw string as a global, which seems idiomatic with other uses within the project.

PR #3750, opened 2020-10-12
Userland/uniq: Add uniq utility

matthewcurry:
Add an implementation for uniq. While it will be nice in the future to make more hardened versions of sort and uniq in the future, this implementation of uniq is compatible with the current version of sort and its buffer sizes.

### DIFF
--- a/Userland/uniq.cpp
+++ b/Userland/uniq.cpp
@@ -1,0 +1,100 @@
+/*
+ * Copyright (c) 2020, Matthew L. Curry <matthew.curry@gmail.com>
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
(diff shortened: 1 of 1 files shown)


dkoao:
You're supposed to put your own info in the copyright notice.

matthewcurry:
> You're supposed to put your own info in the copyright notice. :)

Oops, right you are. Fixed.

Userland/uniq.cpp
```diff
@@ -0,0 +1,64 @@
...
+
+int main(int argc, char** argv)
+{
+    if (pledge("stdio", nullptr) > 0) {
+        perror("pledge");
+        return 1;
+    }
+
+    UNUSED_PARAM(argc);
+    UNUSED_PARAM(argv);
```
dkoao, line 42: The uniq utility should be able to take input and output file paths as arguments, among other things (but I wouldn't focus on the other switches for now)

https://pubs.opengroup.org/onlinepubs/9699919799/utilities/uniq.html

matthewcurry: I agree, I only avoided it in this initial version to get something that works alongside sort, which also does not take files yet. Should I go ahead and work on that?

matthewcurry: There are clearly improvements I can make here, I'll bring more soon.

matthewcurry: I have included positional arguments for input and output files according to The Open Group spec, along with the appropriate pledge. I've also cleaned up header files to just include the ones I'm really using.

dkoao: Looks good to me

matthewcurry:
Small update to this PR -- A stray NULL made it in, so I changed it to nullptr.

Userland/uniq.cpp
```diff
@@ -0,0 +1,90 @@
...
+        auto* str = fgets(current, sizeof(current), infile);
+        if (!str)
+            break;
+        if (!first_run && strncmp(current, previous, BUFSIZ) == 0)
+            continue;
+
+        fputs(current, outfile);
+        auto* tmp = current;
+        current = previous;
+        previous = tmp;
```
bugaevc, line 82: `AK::swap()`

Userland/uniq.cpp
```diff
@@ -0,0 +1,90 @@
...
+    const char* outpath = nullptr;
+    Core::ArgsParser args_parser;
+    args_parser.add_positional_argument(inpath, "Input file", "input", Core::ArgsParser::Required::No);
+    args_parser.add_positional_argument(outpath, "Output file", "output", Core::ArgsParser::Required::No);
+    args_parser.parse(argc, argv);
+
+    FILE* infile = get_stream(inpath, "r");
+    FILE* outfile = get_stream(outpath, "w");
+
+    char buffers[2][BUFSIZ];
```
bugaevc, line 68: But what if the line read doesn't fit into `BUFSIZ` bytes?

Userland/uniq.cpp
matthewcurry: Done

Userland/uniq.cpp
matthewcurry: I had used static buffers because sort did, and I figured there was a reason for it (e.g., readline implementation not ready). I've converted to using readline for lines of unlimited length (and opened a pull request on sort to achieve the same -- #3769).

bugaevc: It would be nice if we had an API to read a line as `AK::String`. Do we have such a thing somewhere already, perhaps in `Core::IODevice`? If we don't, we probably should have.

matthewcurry: Hi there,

I wasn't able to find a readline-esque method, unless one's been added in the last few days. I'll be happy to update this utility to use such a method (and the ensuing dependent classes) once such a method appears. I don't think it's correct to add core functionality in this PR for a utility, readline deserves its own discussion.

PR #3755, opened 2020-10-12
Use new format functions in remaining DevTools.

asynts:


awesomekling:
Aargh did not mean to press squash&merge. Ah well.

PR #3760, opened 2020-10-13
AK: Fix the build on macOS

nooga:
The `abort` function declaration was conflicting with macOS libc.

### DIFF
--- a/AK/TestSuite.h
+++ b/AK/TestSuite.h
@@ -28,7 +28,9 @@
 
 #define AK_TEST_SUITE
 
+#ifndef __APPLE__
 extern "C" __attribute__((noreturn)) void abort() noexcept;
+#endif
 
 namespace AK {
 


asynts:
I did something similar in #3723. I tried to keep the declaration around by adding a `__portable_abort` function which could be forward declared independent of the platform.

Removing the declaration entirely seems to work too, but this merely moves the issue. If we use assertions in any header before `stdlib.h` is included, this would no longer work. (We redefine the `ASSERT` to call abort.)

I don't like my solution at all, so this might be a way to patch it for now. Permanently, it would be better to replace the assertion macros with functions like this, but I am not entirely sure about how to do this best:

~~~c++
void __assert(bool condition, StringView name, StringView file = __builtin_FILE(), size_t line = __builtin_LINE());
#define ASSERT(condition) __assert(condition, #condition)
~~~

nooga:
@asynts Yeah, I'll admit that I just took the path of least resistance here. Got rid of it, it fixed the build on macOS and _doesn't seem to hurt anything_ so it's _good enough_ in my view. 

At the same time, after looking into it some more now, I think you're right about reworking the assertion macros.

bugaevc:
Why can't `AK/TestSuite.h` just `#include <stdlib.h>`?

asynts:
That would mean that the headers in `<AK/TestSuite.h>` would not be commutative. `stdlib.h` would have to be included first, before anything else, otherwise no declaration for `abort` would be available.

We already have some stuff like that going on elsewhere and I really don't like it. Similar to this pull request it's more of a stop gap but doesn't solve the underlying issue.

bugaevc:
But surely it's much better than defining `abort()` on our own, with a signature that might differ from what the system defines, and now even special-casing Apple systems because apparently it is known to break *at least* Apple systems in particular?

asynts:
I agree, I am being the perfectionist here. Including `<stdlib.h>` is a reasonable solution, let's do that.

asynts:
This can now be closed with #3764.

PR #3766, opened 2020-10-14
AudioApplet: Make the slider exponential for finer volume control

nooga:
The volume slider was linear, which is not ideal for an audio volume 
control. Perceived volume would not change much within 30-100% range
and would change in leaps within 0-30% range where the resolution is
not sufficient for fine grained control.

The simplest solution is to bring the value into 0.0-1.0 range and 
square it to obtain an exponential curve. This is a decent 
approximation of the logarithmic taper used in audio potentiometers.

### DIFF
--- a/MenuApplets/Audio/main.cpp
+++ b/MenuApplets/Audio/main.cpp
@@ -108,7 +108,8 @@ public:
         m_slider->set_size_policy(GUI::SizePolicy::Fill, GUI::SizePolicy::Fill);
         m_slider->on_value_changed = [&](int value) {
             int volume = clamp((20 - value) * 5, 0, 100);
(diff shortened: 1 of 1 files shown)


awesomekling:
Hm, does it make sense to do this in the applet instead of in the AudioServer?

nooga:
I would say it makes sense as the curve is usually a property of the instrument i.e. the slider. IMO The underlying DSP circuitry should not assume this and work with raw "voltages", if you know what I mean. This way we avoid confusion in case someone wishes to implement a different control scheme.

One thing I would change at the AudioServer, and the audio subsystem in general, is using floats in 0-1 range for things like volume instead of ints 0-100. It makes the math easier.

PR #3767, opened 2020-10-14
PixelPaint: Basic brush tool

BenJilks:
This patch adds a very basic implementation of the classic brush
tool. It features a wide brush with a falloff around the edges.


awesomekling:
Cool :)

PR #3769, opened 2020-10-15
Convert sort to use getline instead of a static buffer size

matthewcurry:


Userland/sort.cpp
```diff
@@ -44,9 +45,15 @@ int main(int argc, char** argv)
     Vector<String> lines;
 
     for (;;) {
-        char buffer[BUFSIZ];
-        auto* str = fgets(buffer, sizeof(buffer), stdin);
-        if (!str)
+        char* buffer = nullptr;
+        ssize_t buflen = 0;
+        errno = 0;
+        buflen = getline(&buffer, nullptr, stdin);
```
alimpfard, line 51: Doesn't this thing need a `free()`?

awesomekling: Also echoing @bugaevc from elsewhere; can we just invent some C++ API in LibCore that reads one line and returns it as an `AK::String`?

PR #3775, opened 2020-10-15
PixelPaint: Tool properties panel

BenJilks:
Each tool can have its own set of properties that can be modified
through a panel on the right side.

The tools I've added properties for are:

Pen:
	Thickness

Brush:
	Size
	Hardness

Spray:
	Thickness
	Density

Bucket:
	Threshold


awesomekling:
Very cool :)

PR #3776, opened 2020-10-15
Use new format functions in LibChess, LibAudio, Games and a bit of LibCore.

asynts:
9b03ad40d26b7e129a721dbaa189fd831d80440f is probably interesting, the rest is just replacing more `printf` and `LogStream` stuff.

### DIFF
--- a/AK/Format.cpp
+++ b/AK/Format.cpp
@@ -572,36 +572,17 @@ void Formatter<bool>::format(TypeErasedFormatParams& params, FormatBuilder& buil
 }
 
 #ifndef KERNEL
-void raw_out(StringView string)
-{
-    const auto retval = ::fwrite(string.characters_without_null_termination(), 1, string.length(), stdout);
-    ASSERT(retval == string.length());
-}
-void vout(StringView fmtstr, TypeErasedFormatParams params, bool newline)
+void vout(FILE* file, StringView fmtstr, TypeErasedFormatParams params, bool newline)
 {
     StringBuilder builder;
     vformat(builder, fmtstr, params);
 
-    if (newline && !builder.is_empty())
+    if (newline)
         builder.append('\n');
 
-    raw_out(builder.to_string());
-}
-
-void raw_warn(StringView string)
-{
-    const auto retval = ::write(STDERR_FILENO, string.characters_without_null_termination(), string.length());
+    const auto string = builder.string_view();
+    const auto retval = ::fwrite(string.characters_without_null_termination(), 1, string.length(), file);
     ASSERT(static_cast<size_t>(retval) == string.length());
-}
-void vwarn(StringView fmtstr, TypeErasedFormatParams params, bool newline)
-{
-    StringBuilder builder;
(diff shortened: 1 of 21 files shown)


AK/Format.cpp
```diff
@@ -572,37 +572,18 @@ void Formatter<bool>::format(TypeErasedFormatParams& params, FormatBuilder& buil
...
+    if (newline)
         builder.append('\n');
 
-    raw_out(builder.to_string());
-}
-
-void raw_warn(StringView string)
-{
-    const auto retval = ::write(STDERR_FILENO, string.characters_without_null_termination(), string.length());
+    const auto string = builder.build();
```
alimpfard, line 583: (here)

asynts:
I just noticed now that the build failed, I now build locally with `-DALL_THE_DEBUG_MACROS=ON`, the build should be fixed now.

~~~diff
@@ -194,10 +194,10 @@ ./AK/NonnullOwnPtr.h
 }
 
 template<typename T>
-struct Formatter<NonnullOwnPtr<T>> : Formatter<T*> {
+struct Formatter<NonnullOwnPtr<T>> : Formatter<const T*> {
     void format(TypeErasedFormatParams& params, FormatBuilder& builder, const NonnullOwnPtr<T>& value)
     {
-        Formatter<T*>::format(params, builder, value.ptr());
+        Formatter<const T*>::format(params, builder, value.ptr());
     }
 };
 
@@ -103,10 +103,10 @@ ./AK/WeakPtr.h
 }
 
 template<typename T>
-struct Formatter<WeakPtr<T>> : Formatter<T*> {
+struct Formatter<WeakPtr<T>> : Formatter<const T*> {
     void format(TypeErasedFormatParams& params, FormatBuilder& builder, const WeakPtr<T>& value)
     {
-        Formatter<T*>::format(params, builder, value.ptr());
+        Formatter<const T*>::format(params, builder, value.ptr());
     }
 };
 ~~~

AK/Format.cpp
asynts: I wasn't aware that this method existed, I'll update my code.

asynts:
~~~diff
@@ -580,8 +580,8 @@ ./AK/Format.cpp
     if (newline)
         builder.append('\n');
 
-    const auto string = builder.build();
-    const auto retval = ::fwrite(string.characters(), 1, string.length(), file);
+    const auto string = builder.string_view();
+    const auto retval = ::fwrite(string.characters_without_null_termination(), 1, string.length(), file);
     ASSERT(static_cast<size_t>(retval) == string.length());
 }
 #endif
@@ -617,9 +617,9 @@ ./AK/Format.cpp
     vformat(builder, fmtstr, params);
     builder.append('\n');
 
-    const auto string = builder.build();
+    const auto string = builder.string_view();
 
-    const auto retval = dbgputstr(string.characters(), string.length());
+    const auto retval = dbgputstr(string.characters_without_null_termination(), string.length());
     ASSERT(retval == 0);
 }
 ~~~

PR #3777, opened 2020-10-15
FileManager: Only set ~/Desktop as model root path in desktop mode

### LINKED ISSUE #3772: FileManager: Location box empty / tree collapsed when using Desktop menu's "Show in FileManager..."
image

:arrow_down: 

image

linusg:
Setting it as model root path in `DirectoryView::setup_model()` for windowed mode as well would cause an issue with the following:

- "open ~/Desktop"
- "FileManager ~/Desktop"
- "Show in FileManager..." from Desktop context menu

When viewing the Desktop as the initial path it would be the same and `on_path_change` wasn't called, leading to various widgets and window properties not being updated.

Fixes #3772.


awesomekling:
Great work tracking this down :)

PR #3778, opened 2020-10-16
Kernel: Fix sys$join_thread

itamar8910:
Previously, when we unblocked because the joinee has died, we didn't
copy its exit value back to the user.

running `tt` now works as expected, and prints out:
`Okay, joined and got retval=0xdeadbeef`


awesomekling:
Great catch!

tomuta:
Great catch indeed. I don't know what I was thinking. I guess I didn't realize that in this case a thread other than `Thread::current()` can return`Thread::BlockResult::InterruptedByDeath`. In most other cases this would be returned if the current thread is being killed.

PR #3780, opened 2020-10-16
Span: constexpr support

ldm5180:
Problem:
- `Span` is not `constexpr` aware.

Solution:
- Add `constexpr` support for all parts that do not require
  `reinterpret_cast`.
- Modify tests which use the `constexpr` functions.

### DIFF
--- a/AK/Tests/TestSpan.cpp
+++ b/AK/Tests/TestSpan.cpp
@@ -30,26 +30,26 @@
 #include <AK/Span.h>
 #include <AK/StdLibExtras.h>
 
-TEST_CASE(default_constructor_is_empty)
+TEST_CASE(constexpr_default_constructor_is_empty)
 {
-    Span<int> span;
-    EXPECT(span.is_empty());
+    constexpr Span<int> span;
+    static_assert(span.is_empty(), "Default constructed span should be empty.");
 }
 
 TEST_CASE(implicit_converson_to_const)
 {
-    Bytes bytes0;
-    [[maybe_unused]] ReadonlyBytes bytes2 = bytes0;
-    [[maybe_unused]] ReadonlyBytes bytes3 = static_cast<ReadonlyBytes>(bytes2);
+    constexpr Bytes bytes0;
+    [[maybe_unused]] constexpr ReadonlyBytes bytes2 = bytes0;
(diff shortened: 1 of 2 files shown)


AK/Tests/TestSpan.cpp
```diff
@@ -109,25 +109,25 @@ TEST_CASE(at_and_index_operator_return_same_value)
...
-    auto slice = span.slice(3, 2);
+    constexpr Span<const u16> span { buffer, 8 };
+    constexpr auto slice = span.slice(3, 2);
 
-    EXPECT_EQ(slice.size(), 2u);
-    EXPECT_EQ(slice[0], 4);
-    EXPECT_EQ(slice[1], 5);
+    static_assert(slice.size() == 2u);
+    static_assert(slice[0] == 4);
+    static_assert(slice[1] == 5);
```
emanuele6, line 130: Why are you replacing everywhere `EXPECT_EQ(lhs, rhs)` with `static_assert`?

This is a completely different behaviour: `EXPECT_EQ` just prints a message if `lhs != rhs`. `static_assert` will stop compilation if the condition is false.

ldm5180: The `static_assert` executes at compile-time which is needed to test that the `constexpr` works. The `static_assert` will also print a message, just a slightly different one. 

```
/AK/Tests/TestSpan.cpp:52:5: error: static_assert failed due to requirement 'bytes[2] == 4'                                                                                               
    static_assert(bytes[2] == 4);                                                                        
    ^             ~~~~~~~~~~~~~                                                                          
1 error generated.                                                                                       
make[2]: *** [AK/Tests/CMakeFiles/TestSpan.dir/build.make:63: AK/Tests/CMakeFiles/TestSpan.dir/TestSpan.cpp.o] Error 1                                                                                            
```

emanuele6: > The static_assert will also print a message, just a slightly different one.

Yes, but at compile time, not at runtime.

emanuele6: Ah, we also use `static_assert` in `AK/Tests/TestArray.cpp`, so I guess it's fine. I didn't know.

PR #3785, opened 2020-10-16
CircularQueue: constexpr support

ldm5180:
Problem:
- `CircularQueue` is not `constexpr` capable.

Solution:
- Add `constexpr` support to `AK::move()`.
- Add `constexpr` decoration to all member functions.
- Remove placement `new` in favor of simple `move` assignment.
- Remove explicit calls to destructor in favor of `move` assignment
  and let the underlying object call the destructor on subobjects if
  needed.
- Switch to `Array` instead of C-style array so that the type can be
  directly referenced without the need for `reinterpret_cast`.
- Add tests to ensure the compile-time evaluation.

### DIFF
--- a/AK/CircularQueue.h
+++ b/AK/CircularQueue.h
@@ -37,16 +37,19 @@ class CircularQueue {
     friend CircularDuplexStream<Capacity>;
 
 public:
-    CircularQueue()
+    constexpr CircularQueue() = default;
+    constexpr CircularQueue(const Array<T, Capacity * sizeof(T)>&& a)
+        : m_storage(a)
+        , m_size(Capacity)
     {
     }
 
-    ~CircularQueue()
+    constexpr ~CircularQueue()
     {
         clear();
     }
 
-    void clear()
+    constexpr void clear()
     {
         for (size_t i = 0; i < m_size; ++i)
             elements()[(m_head + i) % Capacity].~T();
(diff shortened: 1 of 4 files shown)


ldm5180:
Am I missing something with the changing away from placement-new and removing the explicit destructor calls?

AK/CircularQueue.h
```diff
@@ -118,17 +118,17 @@ class CircularQueue {
...
 
 protected:
-    T* elements() { return reinterpret_cast<T*>(m_storage); }
-    const T* elements() const { return reinterpret_cast<const T*>(m_storage); }
+    constexpr T* elements() { return &m_storage[0]; }
+    constexpr const T* elements() const { return &m_storage[0]; }
 
     friend class ConstIterator;
-    alignas(T) u8 m_storage[sizeof(T) * Capacity];
+    alignas(T) Array<T, sizeof(T) * Capacity> m_storage {};
```
alimpfard, line 131: This won't work, you're forcing all the elements to be default-constructed here.
The point of keeping it as untyped memory is to have the storage without the initialisation.

ldm5180: Thanks @alimpfard. Good call. I didn't think of that. I will have to think more about this.

ldm5180:
Closing for now.

PR #3788, opened 2020-10-17
AK::HashTable exception safety

danopernis:
Hello,

this PR hopefully improves `AK::HashTable` exception safety. I also included small refactor in the destructor, removing some writes to memory. I am not sure about the impact on performance. I would love to add some benchmark, but I am not sure how.

First contribution, I am a big fan :^)

### DIFF
--- a/AK/HashTable.h
+++ b/AK/HashTable.h
@@ -100,16 +100,12 @@ public:
 
     HashTable& operator=(const HashTable& other)
     {
-        if (this != &other) {
-            clear();
-            rehash(other.capacity());
-            for (auto& it : other)
-                set(it);
-        }
+        HashTable temporary(other);
+        swap(*this, temporary);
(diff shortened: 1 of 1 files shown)


awesomekling:
Hello! The system does not use exceptions, so `noexcept` is not necessary :)

danopernis:
I did not realize that. There are some `noexcept`s and `NOEXCEPT`s sprinkled in `math.h`and `Atomic.h`. However, everything is built with `-fno-rtti -fno-exceptions` so I see your point.

Still, I think this way of writing constructors and assignment operators is more idiomatic, although I understand this might be my personal point of view.

Merge or close this at will.

awesomekling:
Meh, it doesn't hurt anyone. And I like the other changes, so let's merge it ;)

PR #3792, opened 2020-10-18
Pixel paint saving

BenJilks:
This adds a few features to do with saving and loading in PixelPaint. It also fixes some issues with base64 decoding. In order to export the image to BMP, I had to create the BMPDumper class, to go along-side BMPLoader.


Applications/PixelPaint/CreateNewImageDialog.cpp
```diff
@@ -0,0 +1,65 @@
+#include "CreateNewImageDialog.h"
```
linusg, line 1: Missing license header, causing CI to fail

Applications/PixelPaint/CreateNewImageDialog.h
```diff
@@ -0,0 +1,23 @@
+#pragma once
```
linusg, line 1: Missing license header

linusg (changes requested):

awesomekling:
`BMPDumper` is a bit of a sad name.. how about we call it `BMPWriter` (and maybe rename `BMPLoader` to `BMPReader` as well)?

alimpfard:
I have a feeling (from just skimming the code) that applying filters does not report "did complete action"?

BenJilks:
@alimpfard You're right there, I forgot about them. I'll add them as actions.

Applications/PixelPaint/History.h
```diff
@@ -0,0 +1,54 @@
...
+#pragma once
+
+#include <AK/String.h>
+#include <AK/Vector.h>
+
+namespace PixelPaint {
+
+class Image;
+
+class History {
```
awesomekling, line 36: Please use `GUI::UndoStack` and `GUI::Command` instead of inventing something new :)
(they are modeled after Qt's `QUndoStack` and `QCommand`)

danboid:
It will be good to see this get finished and merged. I'm surprised pixelpaint and serenity have got this far with no image encoding/saving function.

See if you can get it done before GIMP 3.0 gets released!

danboid:
Can't wait to test this out with my latest masterpiece!

awesomekling:
This breaks undo in TextEditor (try opening it, typing something, and pressing Ctrl+Z)

danboid:
Whatchya tink Mr K?

awesomekling:
Spicy! 🌶️

PR #3793, opened 2020-10-18
Fix the build on FreeBSD

colfrog:
Hi, this is all it took to be able to build and run SerenityOS on FreeBSD 12-STABLE, with consideration for other Unix systems.

Have a good day!

### DIFF
--- a/AK/Random.h
+++ b/AK/Random.h
@@ -33,7 +33,7 @@
 #    include <stdlib.h>
 #endif
 
-#if defined(__linux__)
+#if defined(__unix__)
 #    include <unistd.h>
 #endif
 
@@ -47,7 +47,7 @@ inline void fill_with_random(void* buffer, size_t length)
 {
 #if defined(__serenity__)
     arc4random_buf(buffer, length);
-#elif defined(__linux__) or defined(__APPLE__)
+#elif defined(__unix__) or defined(__APPLE__)
     int rc = getentropy(buffer, length);
     (void)rc;
 #endif
--- a/Meta/build-root-filesystem.sh
+++ b/Meta/build-root-filesystem.sh
@@ -12,10 +12,11 @@ window_gid=13
 
 CP="cp"
 
-# cp on macOS does not support the -d option.
+# cp on macOS and BSD systems do not support the -d option.
 # gcp comes with coreutils, which is already a dependency.
-if [ "$(uname -s)" = "Darwin" ]; then
-	CP=gcp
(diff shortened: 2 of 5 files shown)


Meta/build-root-filesystem.sh
```diff
@@ -12,10 +12,11 @@ window_gid=13
 
 CP="cp"
 
-# cp on macOS does not support the -d option.
+# cp on macOS and FreeBSD do not support the -d option.
 # gcp comes with coreutils, which is already a dependency.
-if [ "$(uname -s)" = "Darwin" ]; then
-	CP=gcp
+OS="$(uname -s)"
+if [ "$OS" = "Darwin" ] || [ "$OS" = "FreeBSD" ] || [ "$OS" = "OpenBSD" ]; then
```
dkoao, line 18: Small nitpick, but it'd be cleaner to match for `*BSD` instead, since most (if not all) BSDs don't have this option. [NetBSD's `cp` doesn't](https://man.netbsd.org/NetBSD-9.0-STABLE/i386/cp.1), for example.

awesomekling: Agreed. That would be a nice tweak :)

awesomekling:
Looks all right! Please squash those commits, and format the commit message according to the [contribution guidelines](https://github.com/SerenityOS/serenity/blob/master/CONTRIBUTING.md)

AK/Random.h
```diff
@@ -47,7 +47,7 @@ inline void fill_with_random(void* buffer, size_t length)
 {
 #if defined(__serenity__)
     arc4random_buf(buffer, length);
-#elif defined(__linux__) or defined(__APPLE__)
+#elif defined(__unix__) or defined(__APPLE__)
     int rc = getentropy(buffer, length);
```
awesomekling, line 51: Actually I think NetBSD might not have `getentropy()` but would define `__unix__`

awesomekling: Not that I've ever heard from anyone building on NetBSD. 🤷🏻‍♂️

colfrog: Could we assume that getrandom would work fine on NetBSD, or would you rather either throw a preprocessor error if __NetBSD__ is defined, or leave it as is and let any eventual NetBSD user deal with the non-existent getentropy function?

Meta/build-root-filesystem.sh
colfrog: Good idea, it's done in my last commit!

PR #3796, opened 2020-10-18
LibJS: Fix arrow function parsing issues

linusg:
This fixes a total of 6 test262-parser-tests tests :^)

---

**LibJS: Share parameter parsing between regular and arrow functions**

This simplifies `try_parse_arrow_function_expression()` and fixes a few cases that should not produce an arrow function AST but did:

```
(a,,) => {}
(a b) => {}
(a ...b) => {}
(...b a) => {}
```

The new parsing logic checks whether parens are expected and uses `parse_function_parameters()` if so, rolling back if a new syntax error occurs during that. Otherwise it's just an identifier in which case we parse the single parameter ourselves.

---

**LibJS: Don't parse arrow function with newline between ) and =>**

If there's a newline between the closing paren and arrow it's not a valid arrow function, ASI should kick in instead (it'll then fail with "Unexpected token Arrow")

### DIFF
--- a/Libraries/LibJS/AST.cpp
+++ b/Libraries/LibJS/AST.cpp
@@ -1442,9 +1442,10 @@ void UpdateExpression::dump(int indent) const
     }
 
     ASTNode::dump(indent);
-    print_indent(indent + 1);
-    if (m_prefixed)
+    if (m_prefixed) {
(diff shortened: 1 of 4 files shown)


awesomekling:
This is all very awesome :)

PR #3798, opened 2020-10-19
Checked: constexpr capability

ldm5180:
Problem:
- Checked does not support `constexpr` construction and behaviors.

Solution:
- Implement `constexpr` support.
- Add tests to ensure compile-time behavior.

### DIFF
--- a/AK/Checked.h
+++ b/AK/Checked.h
@@ -112,111 +112,111 @@ inline constexpr bool is_within_range(Source value)
 template<typename T>
 class Checked {
 public:
-    Checked() = default;
+    constexpr Checked() = default;
 
-    Checked(T value)
+    constexpr Checked(T value)
         : m_value(value)
     {
     }
 
     template<typename U>
-    Checked(U value)
+    constexpr Checked(U value)
     {
         m_overflow = !is_within_range<T>(value);
         m_value = value;
     }
 
-    Checked(const Checked&) = default;
+    constexpr Checked(const Checked&) = default;
 
-    Checked(Checked&& other)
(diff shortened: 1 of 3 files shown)


AK/Checked.h
```diff
@@ -129,37 +126,19 @@ class Checked {
...
-    {
-    }
+    Checked(const Checked&) = default;
 
-    Checked(Checked&& other)
-        : m_value(exchange(other.m_value, 0))
-        , m_overflow(exchange(other.m_overflow, false))
-    {
-    }
+    Checked(Checked&& other) = default;
```
awesomekling, line 131: This is a behavior change.

AK/Checked.h
```diff
@@ -129,37 +126,19 @@ class Checked {
...
-    }
+    Checked& operator=(const Checked& other) = default;
 
-    Checked& operator=(Checked&& other)
-    {
-        m_value = exchange(other.m_value, 0);
-        m_overflow = exchange(other.m_overflow, false);
-        return *this;
-    }
+    Checked& operator=(Checked&& other) = default;
```
awesomekling, line 141: This too.

AK/Checked.h
ldm5180: Good catch.

AK/Checked.h
ldm5180: Good catch.

AK/Checked.h
ldm5180: @awesomekling It appears that it only affects the "moved from" object's state. Since the "moved from" object should be treated as only eligible for assignment or destruction, why does this matter?

PR #3799, opened 2020-10-19
Hackstudio: FormEditor changes

umasankar-yedida:
I have deleted the existing `FormEditor` logic and moved the `VisualBuilder` to `HackStudio` as a folder named `FormEditor`.

There are lot of things need to be done though. The following are the list of items that I want to work on which are much needed.

- [ ] Right now, the widgets are floating around the editor widget which is good. It would be nice to have a list of properties to control the editor widget behaviour like layout, spacing, margin, background color, etc.
- [ ] Only certain widgets can have children. We need to have a property to tell us that whether a particular widget can have other widgets.
- [ ] Enable only relevant toolbar actions in based on the Editing Mode
- [ ] Implement Copy & Paste of widgets
- [ ] Implement Undo & Redo
- [ ] Implement Save the form to a file and load from file to a file. I felt this requires a bit of refactoring because currently the `Editor` class extending the `GUI::TextEditor` class (now it should be other way around) or if it's possible to abstract common functionality to a new `Editor` class and extend it based on the need. A suggestion is very much needed here.
- [ ] Need to fix a crash that's happening upon closing the `HackStudio` when there are some widgets lying around in form editor (which is already there in `VisualBuilder`).
- [ ] UI/UX improvements

Please provide any suggestions/feedback on anything which I need to improve and take this forward. I'll pick one by one from the above list and update you.


awesomekling:
The existing `FormEditor` class in HackStudio is a slightly better design than the old VisualBuilder code IMO. I deleted the VisualBuilder code because it was old and worse than HackStudio, so let's not delete HS code and replace it with VB code :)

Form/widget editing should use the `Core::Object` property system as much as possible, which makes it easy to implement import/export.

umasankar-yedida:
Noted.

umasankar-yedida:
@awesomekling I'm trying to find the difference between `CursorTool` and `WidgetTool`. 

I found out your old commit https://github.com/SerenityOS/serenity/commit/e87756424d3fb0724930425e7cffa83fce073e7a.

Does that mean, we should drag the widget from toolbar to FormEditor using `WidgetTool`? and, move the widget(s) around the FormEditor using `CursorTool`? Specifically, What was your idea behind these when you initially wrote? :)

awesomekling:
The purpose of `CursorTool` is to select, move and resize existing widgets.
The purpose of `WidgetTool` is to create new widgets by drawing them onto the form editor area. :)

Basically, the idea was to recreate the basic workflow from Visual Basic.

umasankar-yedida:
Implemented `WidgetTool` and `CursorTool`. As we don't have `Cross Hair` cursor type, used `Move` for now. Will start working on the properties table.

umasankar-yedida:
@awesomekling I'm stuck at a place where, where would we maintain the properties(layout, width, height, x-pos, y-pos) of a widget that we want to show as a table? 

I thought of
- constructing the list of properties based on the selected widget every time we select a widget. I know, It's a very bad idea.
- If we maintain the properties for a widget when we add it to the form, then we need to wrap the widget and it's a cross-cutting change.
- If we somehow get the list of properties from the widget itself (LibGUI itself), then it would be easier and especially while saving this into a JSON and loading from JSON. I'm not sure if that's the right idea.

I'm unable to conclude this. Any suggestions?

awesomekling:
Let's abandon the FormEditor concept in favor of integrating GML into HackStudio.

PR #3802, opened 2020-10-19
Toolchain: Set CACHED_TOOLCHAIN_ARCHIVE after computing hash

linusg:
...oops. maybe let this one run to completion before hitting the green button :)


awesomekling:
🟢 🟢 🟢 
oops ok :)

PR #3803, opened 2020-10-19
LibJS: Fix parsing of numeric object keys

sunverwerth:
Numeric keys were interpreted as their source text, leading to
something like {0x10:true} to end up as {"0x10":true}
instead of {16:true}

### DIFF
--- a/AK/String.cpp
+++ b/AK/String.cpp
@@ -364,6 +364,15 @@ int String::replace(const String& needle, const String& replacement, bool all_oc
     return positions.size();
 }
 
+String String::reverse() const
+{
+    StringBuilder reversed_string;
+    for (size_t i = length(); i-- > 0;) {
+        reversed_string.append(characters()[i]);
+    }
+    return reversed_string.to_string();
+}
+
 String escape_html_entities(const StringView& html)
 {
     StringBuilder builder;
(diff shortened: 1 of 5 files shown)


awesomekling:
Is this meant to be a draft?

sunverwerth:
@awesomekling Yes, it's almost finished, I think. I still need to consult the standard to see if what I did is actually correct...

sunverwerth:
Quick update, I haven't been able to put much time into SerenityOS lately ='(
I hope I can find some more time after this week. ❤️

linusg (commented):
This looks great, thanks for fixing!

We can keep it on the TODO list for now, but eventually this should be able to replace most of this code:

https://github.com/SerenityOS/serenity/blob/a1029738fdd2bc7471f6db8ec4963734afc4575a/Libraries/LibJS/Runtime/NumberPrototype.cpp#L76-L141

https://tc39.es/ecma262/#sec-number.prototype.tostring

> Return the String representation of this Number value using the radix specified by radixMV. Letters a-z are used for digits with values 10 through 35. The precise algorithm is implementation-defined, however the algorithm should be a generalization of that specified in 6.1.6.1.20.

(the section number at the end is what you just implemented :slightly_smiling_face:)

sunverwerth:
@linusg It was a lot of work and I feel my brain is fried :D Someone who is an expert in floating point math should take a good look at what I just wrote.
Also there's a segfault in there somewhere, I'll need to iron that out and then I can remove the draft status. :+1:

awesomekling:
Very awesome :)

PR #3806, opened 2020-10-19
LibJS: Speed up IndexedPropertyIterator by computing non-empty indices

### LINKED ISSUE #3805: LibJS: interactive interpreter is slow printing objects with large integer keys
Bildschirmfoto 2020-10-19 um 22 59 47

linusg:
This provides a huge speed-up for objects with large numbers as property keys in some situation. Previously we would simply iterate from 0-\<max\> and check if there's a non-empty value at each index - now we're being smarter and compute a list of non-empty indices upfront, by checking each value in the packed elements vector and appending the sparse elements hashmap keys (for `GenericIndexedPropertyStorage`).

Consider this example, an object with a single own property, which is a number increasing by a factor of 10 each iteration:

```js
for (let i = 0; i < 10; ++i) {
    const o = {[10 ** i]: "foo"};
    const start = Date.now();
    Object.getOwnPropertyNames(o);  // <-- IndexedPropertyIterator
    const end = Date.now();
    console.log(`${10 ** i} -> ${(end - start) / 1000}s`);
}
```

Before this change:

    1 -> 0.0000s
    10 -> 0.0000s
    100 -> 0.0000s
    1000 -> 0.0000s
    10000 -> 0.0005s
    100000 -> 0.0039s
    1000000 -> 0.0295s
    10000000 -> 0.2489s
    100000000 -> 2.4758s
    1000000000 -> 25.5669s

After this change:

    1 -> 0.0000s
    10 -> 0.0000s
    100 -> 0.0000s
    1000 -> 0.0000s
    10000 -> 0.0000s
    100000 -> 0.0000s
    1000000 -> 0.0000s
    10000000 -> 0.0000s
    100000000 -> 0.0000s
    1000000000 -> 0.0000s

Fixes #3805.

### DIFF
--- a/Libraries/LibJS/Runtime/IndexedProperties.cpp
+++ b/Libraries/LibJS/Runtime/IndexedProperties.cpp
@@ -24,6 +24,7 @@
  * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
+#include <AK/QuickSort.h>
 #include <LibJS/Runtime/Accessor.h>
 #include <LibJS/Runtime/IndexedProperties.h>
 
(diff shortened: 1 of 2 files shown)


Libraries/LibJS/Runtime/IndexedProperties.cpp
```diff
@@ -266,6 +261,21 @@ ValueAndAttributes IndexedPropertyIterator::value_and_attributes(Object* this_ob
     return {};
 }
 
+void IndexedPropertyIterator::skip_empty_indices()
+{
+    auto indices = m_indexed_properties.indices();
```
awesomekling, line 266: This feels like a heavy thing to do on every iterator increment.
It's okay for now, but maybe later we should consider caching the indices in the iterator.

linusg: Yep there's definitely still room for improvements here

Issue #3754, opened 2020-10-12
FileManager: implement type-ahead filtering


diversys:
Type-ahead filtering is a feature of [Tracker](https://www.haiku-os.org/docs/userguide/en/tracker.html) file manager. It greatly improves user experience in locating files in current folder. It is one of the features i miss the most when working with other OSes.

PR #3808, opened 2020-10-20
LibGUI: Implement searching/jumping as you type in views

tomuta:
This allows the user to start typing and highlighting and jumping
to a match in ColumnsView, IconView, TableView and TreeView if
the model supports it.


Userland/test-fork.cpp
```diff
@@ -0,0 +1,42 @@
...
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include <LibCore/ArgsParser.h>
+#include <stdio.h>
+
+int main(int argc, char** argv)
```
awesomekling, line 30: This file feels unrelated to the commit :)

Libraries/LibGUI/AbstractView.cpp
```diff
@@ -517,7 +526,153 @@ void AbstractView::keydown_event(KeyEvent& event)
...
+        update();
+    }
+}
+
+bool AbstractView::is_searchable() const
+{
+    if (!m_searchable || !model())
+        return false;
+    if (!model()->is_searchable()) {
+		dbg() << "model is not searchable";
```
awesomekling, line 623: Please use `clang-format` :)

Userland/test-fork.cpp
tomuta: whoops indeed...

Libraries/LibGUI/AbstractView.cpp
tomuta: Ugh yeah... we need a pre-commit hook ;-)

awesomekling: Feel free to make one ;--)

awesomekling:
Bless you for consolidating the text drawing code! ❤️ 

Let's find a way to do the searching in `AbstractView` without requiring additional help from the model. All the information needed to perform a search should be already available, no?

tomuta:
> Bless you for consolidating the text drawing code! heart
> 
> Let's find a way to do the searching in `AbstractView` without requiring additional help from the model. All the information needed to perform a search should be already available, no?

I thought about this, but I think it is cleaner to do this in the model. What about remote sources, e.g. file shares where we may not want to download the whole list of data but only what we want to render. Or other very large data sets. I know that's future talk, but to me it seems like the model should be consulted. Just my opinion, though.

awesomekling:
> > Bless you for consolidating the text drawing code! heart
> > Let's find a way to do the searching in `AbstractView` without requiring additional help from the model. All the information needed to perform a search should be already available, no?
> 
> I thought about this, but I think it is cleaner to do this in the model. What about remote sources, e.g. file shares where we may not want to download the whole list of data but only what we want to render. Or other very large data sets. I know that's future talk, but to me it seems like the model should be consulted. Just my opinion, though.

Yeah okay, fair point. Although, we already have `data_matches()`, do we really need a new virtual?

tomuta:
Ok so the build now fails because of incorrect formatting in `Userland/tree.cpp`, but I'm not touching this file here...

awesomekling:
> Ok so the build now fails because of incorrect formatting in `Userland/tree.cpp`, but I'm not touching this file here...

Sorry about that, my bad!

tomuta:
> > > Bless you for consolidating the text drawing code! heart
> > > Let's find a way to do the searching in `AbstractView` without requiring additional help from the model. All the information needed to perform a search should be already available, no?
> > 
> > 
> > I thought about this, but I think it is cleaner to do this in the model. What about remote sources, e.g. file shares where we may not want to download the whole list of data but only what we want to render. Or other very large data sets. I know that's future talk, but to me it seems like the model should be consulted. Just my opinion, though.
> 
> Yeah okay, fair point. Although, we already have `data_matches()`, do we really need a new virtual?

It seems like `data_matches` is used tby `FilteringProxyModel`, but it only returns a `TriState` and doesn't have a way to return a `ModelIndex`, which means one has to iterate each item and call it for each. Thinking about remote searches again ;-)

tomuta:
Hm I guess `data_matches_searching` is really not a good name. How about `data_search`?

awesomekling:
> Hm I guess `data_matches_searching` is really not a good name. How about `data_search`?

Hm, or simply `Model::match()` ([like Qt does](https://doc.qt.io/qt-5/qabstractitemmodel.html#match).)

And if we make it return a `Vector<ModelIndex>` with all the matching indices, we could use it to implement the filtering model in a better way as well.. I think.

tomuta:
> > Hm I guess `data_matches_searching` is really not a good name. How about `data_search`?
> 
> Hm, or simply `Model::match()` ([like Qt does](https://doc.qt.io/qt-5/qabstractitemmodel.html#match).)
> 
> And if we make it return a `Vector<ModelIndex>` with all the matching indices, we could use it to implement the filtering model in a better way as well.. I think.

Sounds like a good and easy change. I think I'll also add some flags similar to Qt (but not all).

diversys:
Thanks a lot for working on this! :) As per #3754 type-ahead filtering should hide Tracker entries which don't match what was typed. Does it also implement the same in FileManager?

alimpfard:
> And if we make it return a `Vector<ModelIndex>` with all the matching indices, we could use it to implement the filtering model in a better way as well.. I think.

That's such an obvious thing to do, how did I miss that when I made the filtering proxy?

> filtering should hide Tracker entries which don't match what was typed

Out of curiousity, how would one cancel the search there? I imagine when some stuff is just selected, you can simply select something else and the search is now over

tomuta:
> Thanks a lot for working on this! :) As per #3754 type-ahead filtering should hide Tracker entries which don't match what was typed. Does it also implement the same in FileManager?

It does not hide non-matches. It implements jumping to entries that match (and highlight it), which is similar to what Windows and some Linux apps do. This enables fast navigation using the keyboard, and I feel that this is a better user experience. In fact Ubuntu does this out-of-the-box in Nautilus now too and it is the first feature I always disable on all my machines because I don't find it intuitive to use. I feel that a search-like experience should be triggered with an explicit Ctrl+F, which then would filter non-matches.

Maybe I should remove the "Fixes #3754" part as it doesn't implement that exact feature.

diversys:
> Out of curiousity, how would one cancel the search there? I imagine when some stuff is just selected, you can simply select something else and the search is now over

Nope, the stuff is not selected. In fact what was typed/matched becomes the only thing that is shown, the other entries get hidden. And you cancel it with Escape.

diversys:
> Maybe I should remove the "Fixes #3754" part as it doesn't implement that exact feature.

Please try this feature in Haiku for a few minutes to feel what it does. Of course It can be hidden behind an option like it's done in Tracker but IMHO when you try it for some time it feels so natural that you don't want to ever go back! :)

alimpfard:
Hmm...I don't really use graphical file managers, but I think type-ahead seems more useful as a "take me to this" than a full search?
Bonus points for globs @tomuta :P

tomuta:
> > Maybe I should remove the "Fixes #3754" part as it doesn't implement that exact feature.
> 
> Please try this feature in Haiku for a few minutes to feel what it does. Of course It can be hidden behind an option like it's done in Tracker but IMHO when you try it for some time it feels so natural that you don't want to ever go back! :)

I might give it a try later tonight. If it works like Nautilus then I don't like it. I generally browse file system architectures by typing the beginning of files/directories and it jumps right to these entries, then I usually either hit Enter to continue or use up/down/left/right to pick another entry (if needed). Nautilus hides all other entries (and is super slow as well because I believe it searches subdirectories as well), and I find that irritating when I don't know the exact search string. In this implementation you can start typing, if a character doesn't match it still highlights the last match, and you simply press the correct letter (which you probably find by seeing the entries in close proximity). This is what Windows did (not sure if it still does, but definitely up to and including Win 7) and personally I find this the most convenient way to quickly navigate file systems (and other lists). But that's just me I guess.

tomuta:
> > Maybe I should remove the "Fixes #3754" part as it doesn't implement that exact feature.
> 
> Please try this feature in Haiku for a few minutes to feel what it does. Of course It can be hidden behind an option like it's done in Tracker but IMHO when you try it for some time it feels so natural that you don't want to ever go back! :)

I don't see pressing ctrl+f and start typing too much of an inconvenience. That's what you do in almost all applications to find/filter stuff. Why not also in FileManager? Consistency isn't a strength of most UIs these days.

tomuta:
@diversys ok so I just tried Haiku and it does not hide non-matching entries, either. it selects the best match, similar to what this implementation does. Or am I doing something wrong?

diversys:
Right. It doesn't do that by default. You need to enable Type-ahead option in Tracker preferences.

tomuta:
> Right. It doesn't do that by default. You need to enable Type-ahead option in Tracker preferences.

I see. Well it seems to be better than what Nautilus does. Personally, I probably still wouldn't enable it but I guess I wouldn't be opposed to adding such an option as well. I guess I'll just remove the reference to your bug because it's not the exact feature you're asking for, I probably won't implement that as I don't see myself ever using/wanting it ;-)

tomuta:
@awesomekling Added `Model::matches` as discussed.

diversys:
> > Right. It doesn't do that by default. You need to enable Type-ahead option in Tracker preferences.
> 
> I see. Well it seems to be better than what Nautilus does. Personally, I probably still wouldn't enable it but I guess I wouldn't be opposed to adding such an option as well. I guess I'll just remove the reference to your bug because it's not the exact feature you're asking for, I probably won't implement that as I don't see myself ever using/wanting it ;-)

Eh :) That's because you didn't use this feature long enough  ;-) Give it a go for an hour or two and you'll see just how useful it really is :)

tomuta:
> > > Right. It doesn't do that by default. You need to enable Type-ahead option in Tracker preferences.
> > 
> > 
> > I see. Well it seems to be better than what Nautilus does. Personally, I probably still wouldn't enable it but I guess I wouldn't be opposed to adding such an option as well. I guess I'll just remove the reference to your bug because it's not the exact feature you're asking for, I probably won't implement that as I don't see myself ever using/wanting it ;-)
> 
> Eh :) That's because you didn't use this feature long enough ;-) Give it a go for an hour or two and you'll see just how useful it really is :)

Nah, I used it long enough in Nautilus (although that implementation is more obnoxious). I prefer just jumping to the entries, I don't need/want everything hidden. When I do I just hit ctrl+f

PR #3815, opened 2020-10-21
LibJS: Support all line terminators (LF, CR, LS, PS)

linusg:
You can now feed weirdo line terminators to LibJS. Fixes 7 test262-parser-tests tests.

https://tc39.es/ecma262/#sec-line-terminators


Libraries/LibJS/Lexer.cpp
```diff
@@ -30,8 +30,27 @@
...
+    } while (0)
+
+#define THREE_CHARS_VIEW()                                                       \
+    StringView view;                                                             \
+    do {                                                                         \
+        char second_char = m_source[m_position];                                 \
+        char third_char = m_source[m_position + 1];                              \
+        char three_chars[] { (char)m_current_char, second_char, third_char, 0 }; \
+        view = { three_chars };                                                  \
+    } while (0)
```
linusg, line 52: This is what I asked about on IRC this morning - made it two macros instead, I guess that works?!

Libraries/LibJS/Tests/parser-line-terminators.js
```diff
@@ -0,0 +1,41 @@
...
+    }).toThrowWithMessage(SyntaxError, "line: 3, column: 1");
+});
+
+test("PARAGRAPH SEPARATOR is a line terminator", () => {
+    expect(() => {
+        Function("\
\
@");
+    }).toThrowWithMessage(SyntaxError, "line: 3, column: 1");
+});
```
linusg, line 29: GitHub doesn't show it but there are LS/PS behind each backslash.

Libraries/LibJS/Tests/parser-line-terminators.js
```diff
@@ -0,0 +1,41 @@
...
+
+test("PARAGRAPH SEPARATOR is a line terminator", () => {
+    expect(() => {
+        Function(`

@`);
+    }).toThrowWithMessage(SyntaxError, "line: 3, column: 1");
+});
```
linusg, line 29: GitHub doesn't show it, but the template literal contains two LS/PS chars.

Libraries/LibJS/Lexer.cpp
alimpfard: Any particular reason why you can't just use `m_source`?
`StringView view { &m_source[m_position-1], 2 }` for the two chars one?
(What even is `m_source`? a String?)

linusg: > Any particular reason why you can't just use m_source?

'cause I didn't think of that, that's why :smile:

> (What even is m_source? a String?)

A StringView :)

I'll try this - thanks!

linusg: that totally worked lol. To not see the forest for the trees, or something... ty!

linusg:
A little JavaScript speciality I initially missed but fixed just now: LF/CR in string literal are not ok, LS/PS *are* :^)

PR #3821, opened 2020-10-22
LibJS: Various parser fixes

linusg:
My daily dose of LibJS improvements!

---

Fixes 40 (:smiley:) test262-parser-tests tests.

### DIFF
--- a/Libraries/LibJS/Parser.cpp
+++ b/Libraries/LibJS/Parser.cpp
@@ -241,7 +241,13 @@ NonnullRefPtr<Program> Parser::parse_program()
     bool first = true;
     m_parser_state.m_use_strict_directive = UseStrictDirectiveState::Looking;
     while (!done()) {
-        if (match_statement()) {
+        if (match_declaration()) {
+            program->append(parse_declaration());
+            if (first) {
+                first = false;
+                m_parser_state.m_use_strict_directive = UseStrictDirectiveState::None;
+            }
+        } else if (match_statement()) {
             program->append(parse_statement());
             if (first) {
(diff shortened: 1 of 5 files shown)


Libraries/LibJS/Parser.cpp
```diff
@@ -252,7 +258,7 @@ NonnullRefPtr<Program> Parser::parse_program()
                 m_parser_state.m_use_strict_directive = UseStrictDirectiveState::None;
             }
         } else {
-            expected("statement");
+            expected("statement or ");
```
emanuele6, line 261: "statement or declaration"?

emanuele6 (changes requested):

Libraries/LibJS/Parser.cpp
linusg: ...yes. thanks :)

Libraries/LibJS/Parser.cpp
```diff
@@ -314,11 +332,13 @@ NonnullRefPtr<Statement> Parser::parse_statement()
...
         }
         if (match_expression()) {
+            if (match(TokenType::Function))
+                syntax_error("Function declaration not allowed in single-statement context");
             auto expr = parse_expression(0);
             consume_or_insert_semicolon();
             return create_ast_node<ExpressionStatement>(move(expr));
         }
-        expected("statement (missing switch case)");
+        expected("statement");
```
linusg, line 341: I decided to get rid of these as this is not really the case anymore - it's not a "missing" switch case like in the early days, it's a syntax error, period.

PR #3822, opened 2020-10-22
LibJS: Check for exception after executing (do)while test expression

linusg:
Otherwise we crash the interpreter when an exception is thrown during evaluation of the while or do/while test expression - which is easily caused by a `ReferenceError` - e.g.:

```js
while (someUndefinedVariable) {
    // ...
}
```

### DIFF
--- a/Libraries/LibJS/AST.cpp
+++ b/Libraries/LibJS/AST.cpp
@@ -258,9 +258,12 @@ Value IfStatement::execute(Interpreter& interpreter, GlobalObject& global_object
 Value WhileStatement::execute(Interpreter& interpreter, GlobalObject& global_object) const
 {
     Value last_value = js_undefined();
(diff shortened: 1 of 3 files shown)


Libraries/LibJS/AST.cpp
```diff
@@ -258,9 +258,12 @@ Value IfStatement::execute(Interpreter& interpreter, GlobalObject& global_object
 Value WhileStatement::execute(Interpreter& interpreter, GlobalObject& global_object) const
 {
     Value last_value = js_undefined();
-    while (m_test->execute(interpreter, global_object).to_boolean()) {
+    for (;;) {
+        auto test_result = m_test->execute(interpreter, global_object);
         if (interpreter.exception())
             return {};
```
linusg, line 264: this apparently was an attempt to circumvent this very problem - but by the time we called `to_boolean()` on the empty value it was already too late.

PR #3823, opened 2020-10-23
AK: URL encode the components of a URL when stringifying it / LibWeb: Fix named character reference parsing inside attribute values 

### LINKED ISSUE #3633: Browser: HTTP request header injection through links with linebreaks
**Reproduction:**

```html
<!DOCTYPE html>
<html>
<head>
</head>
<body>
    <a href="http://192.168.1.103:9999/test HTTP/1.1
Abcd: efgh
Ijkl: mnop
">Click me</a>
</body>
</html>
```

**Request sent:**
```
GET /test HTTP/1.1
Abcd: efgh
Ijkl: mnop
 HTTP/1.1
Host: 192.168.1.103
User-Agent: Mozilla/4.0 (SerenityOS; x86) LibWeb+LibJS (Not KHTML, nor Gecko) LibWeb
Connection: close
```

alimpfard:
Fixes #3633

Also fixes `<a href="foo&amp;bar">named character references</a>` inside
attribute values (the standard says to check the matched reference, but
we checked the codepoints of the resulting entity).

### DIFF
--- a/AK/URLParser.cpp
+++ b/AK/URLParser.cpp
@@ -70,31 +70,30 @@ String urldecode(const StringView& input)
     return builder.to_string();
 }
 
-static inline bool in_c0_control_set(u32 c)
+String urlencode(const StringView& input, URLEncodeMode mode)
 {
-    return c <= 0x1f || c > '~';
+    switch (mode) {
+    case URLEncodeMode::PreserveSpecialCharacters:
+        return urlencode(input, "/?:@-._~!$&'()*+,;="); // https://www.ietf.org/rfc/rfc2396.txt section 2.2 and 2.3, "reserved characters" and "unreserved characters".
+    case URLEncodeMode::Full:
+        return urlencode(input, "-_.!~*'()"); // Just "unreserved characters", alphanums are never encoded.
+    }
+
+    ASSERT_NOT_REACHED();
 }
 
-static inline bool in_fragment_set(u32 c)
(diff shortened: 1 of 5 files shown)


AK/URLParser.cpp
```diff
@@ -70,31 +70,30 @@ String urldecode(const StringView& input)
...
-    return c <= 0x1f || c > '~';
-}
-
-static inline bool in_fragment_set(u32 c)
-{
-    return in_c0_control_set(c) || c == ' ' || c == '"' || c == '<' || c == '>' || c == '`';
-}
+    switch (mode) {
+    case URLEncodeMode::PreserveSpecialCharacters:
+        return urlencode(input, "/?:@-._~!$&'()*+,;="); // https://www.ietf.org/rfc/rfc2396.txt section 2.2 and 2.3, "reserved characters" and "unreserved characters".
```
awesomekling, line 77: Please use https://url.spec.whatwg.org/ as a reference for URL encoding and decoding :)

alimpfard: Oh my, a fancy specification that's not a text file written a few months after I was born!

alimpfard: That spec seems to be oddly silent about what to do when serialising query strings (it seems to assume that the query string is sourced from a valid url, and so the query string is definitely encoded correctly?)
I don't think I have the patience to read this spec fully, I'll just close this PR for now and revisit it later

PR #3831, opened 2020-10-24
UserspaceEmulator(+Kernel): Some stuff to make starting SystemServer itself from UE possible

alimpfard:
I saw some SystemServer crash and decided to run it from UE
imagine my surprise when I realised that the `init` kernel argument is just the program name.

> \<wezm> I added a large source tree to Base/home/anon/Source. The resulting disk image is 2.5Gb. Now Serenity crashes on boot. The particular error moves around between builds but it repeatable if the same image is re-run. Are there any tips for debuging kernel issues? I looking in Documentation and searched the IRC log already.
> \<wezm> This is the error I see: paste.gg/p/anonymous/3f250f2558b94f778cdb9bf608803af6

So this PR makes it possible (somewhat) to boot (textmode) with everything running under UE.

Please take a look at the `create_thread` impl, I am almost definitely certain that I've done something very wrong with it.

### DIFF
--- a/DevTools/UserspaceEmulator/Emulator.cpp
+++ b/DevTools/UserspaceEmulator/Emulator.cpp
@@ -353,6 +353,8 @@ u32 Emulator::virt_syscall(u32 function, u32 arg1, u32 arg2, u32 arg3)
         return virt$chmod(arg1, arg2, arg3);
     case SC_fchmod:
         return virt$fchmod(arg1, arg2);
+    case SC_fchown:
+        return virt$fchown(arg1, arg2, arg3);
     case SC_accept:
         return virt$accept(arg1, arg2, arg3);
     case SC_setsockopt:
@@ -503,6 +505,11 @@ int Emulator::virt$chmod(FlatPtr path_addr, size_t path_length, mode_t mode)
 int Emulator::virt$fchmod(int fd, mode_t mode)
 {
     return syscall(SC_fchmod, fd, mode);
+}
+
+int Emulator::virt$fchown(int fd, uid_t uid, gid_t gid)
+{
+    return syscall(SC_fchown, fd, uid, gid);
 }
 
 int Emulator::virt$setsockopt(FlatPtr params_addr)
--- a/DevTools/UserspaceEmulator/Emulator.cpp
+++ b/DevTools/UserspaceEmulator/Emulator.cpp
@@ -286,6 +286,8 @@ u32 Emulator::virt_syscall(u32 function, u32 arg1, u32 arg2, u32 arg3)
         return virt$shbuf_set_volatile(arg1, arg2);
     case SC_mmap:
         return virt$mmap(arg1);
+    case SC_mount:
+        return virt$mount(arg1);
     case SC_munmap:
         return virt$munmap(arg1, arg2);
     case SC_gettid:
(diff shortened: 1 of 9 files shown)


DevTools/UserspaceEmulator/Emulator.cpp
```diff
@@ -492,6 +494,12 @@ int Emulator::virt$dbgputstr(FlatPtr characters, int length)
     return 0;
 }
 
+int Emulator::virt$chmod(FlatPtr path_addr, size_t path_length, mode_t mode)
+{
+    auto path = String::copy(mmu().copy_buffer_from_vm(path_addr, path_length));
```
awesomekling, line 508: I don't think you need to make a `String` here as the syscall takes characters+length.

DevTools/UserspaceEmulator/Emulator.cpp
```diff
@@ -1450,4 +1455,47 @@ int Emulator::virt$sched_setparam(int pid, FlatPtr user_addr)
     return syscall(SC_sched_setparam, pid, &user_param);
 }
 
+void thread_entry_point(Emulator* emu, FlatPtr entry, FlatPtr esp)
+{
+    dbg() << "Thread entry point, emu=" << emu << ", entry=" << entry;
+    emu->m_cpu.set_eip((FlatPtr)entry);
+    // FIXME: Don't assume u32.
+    emu->m_cpu.set_esp(shadow_wrap_as_initialized((u32)esp));
+    emu->exec();
```
awesomekling, line 1473: How is this expected to work? The thread will clobber the CPU registers used by the emulator on the main thread.
Secondary threads should either have a separate CPU, or the emulator should have a thread scheduler that periodically switches between threads (while preserves the register state of each thread.)

DevTools/UserspaceEmulator/Emulator.cpp
alimpfard: Hmmm...I copied it to a string because I saw `realpath` and few other places doing the same, but now that you say it, I don't think any of them need this except maybe `execve`
Should I also replace the rest?

DevTools/UserspaceEmulator/Emulator.cpp
alimpfard: Oh.
I seem to have mistaken it for a separate process...oops

DevTools/UserspaceEmulator/Emulator.cpp
awesomekling: Sure that could be cool :)

DevTools/UserspaceEmulator/Emulator.cpp
alimpfard: I can think of two things to do about this (just your suggestions, really), and I don't like either of them:
- keep a bunch of `SoftCPU`'s in a hashmap or something, one for each active thread
- write another scheduler, as the one in Kernel is a bit too specialised

Both of these sound terrible, but I'm leaning towards option A - alternatively, dropping that commit entirely.
Any suggestions or alternatives you can think of?

alimpfard: Nevermind, putting a `SoftCPU` in a hashmap is annoying, that specific syscall can wait a little longer :shrug:

awesomekling:
This is highly awesome. Running the whole userspace in UE. :)

PR #3835, opened 2020-10-24
Shell: Brace expansions

### LINKED ISSUE #3832: Shell: foo_{bar,baz} should expand properly
This type of expression is so common people even use it human communication. It's silly not to support it in our shell :)

I realize that we have `foo_(bar baz)` but we should also support `{}` expansion.

alimpfard:
Adds brace expansions.
Closes #3832 

Note: The behaviour is quite dissimilar to bash in edge cases, but the test passes under both zsh and Shell.

cc @bugaevc, you might be interested
cc @awesomekling, you asked for this (yes, this cc is redundant, I understand the irony)


Shell/Tests/brace-exp.sh
```diff
@@ -0,0 +1,21 @@
...
+}
+
+test "$(echo {a,b,})" = "a b " || fail normal brace expansion with one empty slot
+test "$(echo {a,,b})" = "a  b" || fail normal brace expansion with one empty slot
+test "$(echo {a,,,b})" = "a   b" || fail normal brace expansion with two empty slots
+test "$(echo {a,b,,})" = "a b  " || fail normal brace expansion with two empty slots
+
+test "$(echo {a..c})" = "a b c" || fail range brace expansion, alpha
+test "$(echo {0..3})" = "0 1 2 3" || fail range brace expansion, number
+test "$(echo {س..ص})" = "س ش ص" || fail range brace expansion, unicode codepoint
```
alimpfard, line 17: First instance of these codepoints in these tests, might be smarter to use codepoints that our fonts have glyphs for? I'm not entirely sure.

alimpfard:
I should also point out that `({echo,foo,bar})` is now valid syntax...

Shell/Tests/brace-exp.sh
linusg: How about emojis, like in this LibJS test testing Unicode capabilities:

https://github.com/SerenityOS/serenity/blob/4fb96afafc6e87b75d2e310f949b0ca7b337b050/Libraries/LibJS/Tests/string-escapes.js#L9-L15

alimpfard: Do we have a consecutive range of emojis drawn?

Let's iterate some emojis instead!

linusg: From emoji.txt

```
😂 - U+1F602 FACE WITH TEARS OF JOY
😃 - U+1F603 SMILING FACE WITH OPEN MOUTH
😄 - U+1F604 SMILING FACE WITH OPEN MOUTH AND SMILING EYES
```

alimpfard: Even this test is happy about ranges of emojis :P

bugaevc: 😀

alimpfard: 😂 😃 😄

awesomekling:
Thank you! :D

PR #3836, opened 2020-10-24
LibGUI: Fix Splitter sometimes not working after widgets were resized

### LINKED ISSUE #3491: System Monitor: Splitter occasionally disappears
Not sure what's going on and I can't consistently trigger this, but try doing this:

1. Open System Monitor. Make sure to not resize the window
2. Select a process, try "Shell" (it may not reproduce with all)
3. Now maximize System Monitor. Notice that some times the splitter disappeared. Restoring the window sometimes makes it work again. If it still works, Close System Monitor and try again, or restore the window and try different processes until you find one.

tomuta:
We need to skip over widgets that are not visible as the layout does
not update their location. This fixes finding the correct widgets
surrounding the splitter.

Fixes #3491


awesomekling:
Very nice!

PR #3839, opened 2020-10-25
LibJS: Implement rules for duplicate function parameters

linusg:
- A regular function can have duplicate parameters except in strict mode or if its parameter list is not "simple" (has a default or rest parameter)
- An arrow function can never have duplicate parameters

Compared to other engines I opted for more useful syntax error messages than a generic "duplicate parameter name not allowed in this context":

```
"use strict"; function test(foo, foo) {}
                                 ^
Uncaught exception: [SyntaxError]: Duplicate parameter 'foo' not allowed in strict mode (line: 1, column: 34)
```

```
function test(foo, foo = 1) {}
                   ^
Uncaught exception: [SyntaxError]: Duplicate parameter 'foo' not allowed in function with default parameter (line: 1, column: 20)
```

```
function test(foo, ...foo) {}
                      ^
Uncaught exception: [SyntaxError]: Duplicate parameter 'foo' not allowed in function with rest parameter (line: 1, column: 23)
```

```
(foo, foo) => {}
      ^
Uncaught exception: [SyntaxError]: Duplicate parameter 'foo' not allowed in arrow function (line: 1, column: 7)

### DIFF
--- a/Libraries/LibJS/Parser.cpp
+++ b/Libraries/LibJS/Parser.cpp
@@ -359,10 +359,11 @@ RefPtr<FunctionExpression> Parser::try_parse_arrow_function_expression(bool expe
         // We have parens around the function parameters and can re-use the same parsing
         // logic used for regular functions: multiple parameters, default values, rest
         // parameter, maybe a trailing comma. If we have a new syntax error afterwards we
-        // know parsing failed and rollback the parser state.
+        // check if it's about a wrong token (something like duplicate parameter name must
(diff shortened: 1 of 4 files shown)


linusg:
I forgot to mention, of course: fixes 12 test262-parser-tests tests. :^)

awesomekling:
> I forgot to mention, of course: fixes 12 test262-parser-tests tests. :^)

Very rad!

PR #3845, opened 2020-10-25
LibJS: Parse line continuations in string literals properly

### LINKED ISSUE #3816: LibJS: Line continuation produces newline character in string
```js
"a\
b".length // 3
```

should be

```js
"a\
b".length // 2
```

https://tc39.es/ecma262/#sec-literals-string-literals

> <LF> and <CR> cannot appear in a string literal, except **as part of a LineContinuation to produce the empty code points sequence**. The proper way to include either in the String value of a string literal is to use an escape sequence such as \n or \u000A.

When fixing this we need to make sure a use strict directive is not recognised when containing a line continuation (it currently isn't due to the newline - `"use\n strict"`):

https://tc39.es/ecma262/#sec-directive-prologues-and-the-use-strict-directive

> A Use Strict Directive is an ExpressionStatement in a Directive Prologue whose StringLiteral is either of the exact code point sequences "use strict" or 'use strict'. A Use Strict Directive **may not contain an EscapeSequence or LineContinuation**.

nooga:
Newlines after line continuation were inserted into the string 
literals. This patch makes the parser ignore the newlines after \ and
also makes it so that "use strict" containing a line continuation is 
not a valid "use strict".

Fixes #3816

### DIFF
--- a/Libraries/LibJS/Token.cpp
+++ b/Libraries/LibJS/Token.cpp
@@ -155,6 +155,8 @@ String Token::string_value(StringValueStatus& status) const
             case '\\':
                 builder.append('\\');
                 break;
+            case '\n':
+                break;
             case 'x': {
                 if (i + 2 >= m_value.length() - offset)
                     return encoding_failure(StringValueStatus::MalformedHexEscape);
--- a/Libraries/LibJS/Tests/functions/function-strict-mode.js
+++ b/Libraries/LibJS/Tests/functions/function-strict-mode.js
@@ -30,6 +30,12 @@ test("use strict with double quotes after statement does not yield strict mode c
     expect(isStrictMode()).toBeFalse();
 });
 
+test("use strict interrupted by a line continuation does not yield strict mode code", () => {
+    "use \
+    strict";
+    expect(isStrictMode()).toBeFalse();
+});
+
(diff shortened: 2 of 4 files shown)


Libraries/LibJS/Token.cpp
```diff
@@ -155,6 +155,9 @@ String Token::string_value(StringValueStatus& status) const
             case '\\':
                 builder.append('\\');
                 break;
+            case '\n':
+                status = StringValueStatus::OkWithLineContinuation;
```
linusg, line 159: Do we need a new state for this, why not simply `Ok`?

nooga: Oh... it was to prevent
```
"use\
 strict"
``` 
from working but now I see that there is another mechanism for ensuring that. 
I'll make some changes.

linusg: oops I changed that recently, forgot to update the issue mentioning it - sorry about that :)

nooga: Haha, all good. Didn't notice your check when rebasing onto master - it made the additional state obsolete so I'm removing that.

linusg (approved):

PR #3846, opened 2020-10-25
Shell: Fix off-by-one in EOL mark printing

### LINKED ISSUE #3844: Shell: cat'ing a 1-character text file shows nothing (missing newline detection failure?)
Reproduce:
```bash
echo -n x > foo
cat foo
```

cc @alimpfard

alimpfard:
Fixes #3844.


alimpfard:
Not sure why that was there to begin with...?

PR #3847, opened 2020-10-25
LookupServer: Support multiple nameservers

### LINKED ISSUE #3843: LookupServer: Add suport for secondary DNS server
```
12:49 < kling> LookupServer feature idea: secondary DNS server ;)
```

linusg:
Best reviewed as individual commits as I also did some other things. Closes #3843.

### DIFF
--- a/Base/etc/LookupServer.ini
+++ b/Base/etc/LookupServer.ini
@@ -1,2 +1,2 @@
 [DNS]
-Nameserver=1.1.1.1
+Nameservers=1.1.1.1,1.0.0.1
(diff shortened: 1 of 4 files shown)


Base/etc/LookupServer.ini
```diff
@@ -1,2 +1,2 @@
 [DNS]
-Nameserver=1.1.1.1
+Nameservers=1.1.1.1,1.0.0.1
```
dkoao, line 2: Off-topic: Shouldn't these info be supplied by DHCP?

linusg: Not sure what the current state of DHCP in Serenity is, but yeah, eventually it should do that :)

awesomekling: Yeah that would be a cool thing to add to the `DHCPClient` service :)

PR #3850, opened 2020-10-25
Kernel: Set up and calibrate APIC timer, and enable timer on all CPUs

tomuta:
This enables the APIC timer on all CPUs, which means Scheduler::timer_tick
is now called on all CPUs independently. We still don't do anything on
the APs as it instantly crashes due to a number of other problems.

You asked for some work in that area, so here is a (small) step forward ;-)


awesomekling:
Very nice :)

PR #3851, opened 2020-10-25
LibTLS: Treat a close_notify before agreeing on a cipher suite as a h…

Lubrsi:
…andshake failure

Some TLS implementations (namely, AWS CloudFront) do this instead of
sending handshake_failure for some reason.

### DIFF
--- a/Libraries/LibTLS/Record.cpp
+++ b/Libraries/LibTLS/Record.cpp
@@ -315,6 +315,12 @@ ssize_t TLSv12::handle_message(const ByteBuffer& buffer)
                 res += 2;
                 alert(AlertLevel::Critical, AlertDescription::CloseNotify);
                 m_context.connection_finished = true;
+                if (!m_context.cipher_spec_set) {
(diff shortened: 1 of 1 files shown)


alimpfard:
The amount of spec non-conformance out there is amazing...(it's also always these $BIG_COMPANY services for some reason)

Lubrsi:
What's even better is that AWS S3 uses a different TLS implementation to CloudFront, as we can load the JS on GitHub which is hosted on it.

Lubrsi:
Or at least has a more accepting cipher suite

Lubrsi:
S3 and CloudFront really _do_ use different implementations:
```
ProtocolServer(25): We were alerted of a critical error: 40 (HandshakeFailure)
ProtocolServer(25): Possible failure cause(s): 
ProtocolServer(25): - No cipher suite in common with github.githubassets.com
ProtocolServer(25): Consumed an error: -1
```
I wonder why

awesomekling:
Huh, well okay :)

PR #3852, opened 2020-10-25
Shell: Various (mostly backgrounding) fixes

alimpfard:


Shell/AST.h
```diff
@@ -1256,3 +1257,18 @@ class WriteRedirection final : public PathRedirectionNode {
...
+template<>
+struct Formatter<Shell::AST::Command> : StandardFormatter {
+    Formatter() { }
+    explicit Formatter(StandardFormatter formatter)
+        : StandardFormatter(formatter)
+    {
+    }
+
+    void format(TypeErasedFormatParams&, FormatBuilder&, Shell::AST::Command value);
+};
```
alimpfard, line 1272: @asynts I tried to make this take a `const` reference, but that made the compiler very upset, is there some trick to this, or does `format()` incur a copy every time regardless?

asynts: There are very little restrictions about the format function, it just has to be callable from:

https://github.com/SerenityOS/serenity/blob/af056718437f23e895f45e9b6c473d37a3e5f699/AK/Format.h#L181-L188

If I make the following change against 7b4e1b2d23be3691fb1598481fc815db489fa61b (alimpfard:shell-fixes) it compiles on my machine:

~~~diff
diff --git a/Shell/AST.cpp b/Shell/AST.cpp
index b65ed3471..32942fdb2 100644
--- a/Shell/AST.cpp
+++ b/Shell/AST.cpp
@@ -37,7 +37,7 @@
 
 //#define EXECUTE_DEBUG
 
-void AK::Formatter<Shell::AST::Command>::format(TypeErasedFormatParams&, FormatBuilder& builder, Shell::AST::Command value)
+void AK::Formatter<Shell::AST::Command>::format(TypeErasedFormatParams&, FormatBuilder& builder, const Shell::AST::Command& value)
 {
     if (m_sign_mode != FormatBuilder::SignMode::Default)
         ASSERT_NOT_REACHED();
diff --git a/Shell/AST.h b/Shell/AST.h
index 53507e7d4..a10e8ed7f 100644
--- a/Shell/AST.h
+++ b/Shell/AST.h
@@ -1268,7 +1268,7 @@ struct Formatter<Shell::AST::Command> : StandardFormatter {
     {
     }
 
-    void format(TypeErasedFormatParams&, FormatBuilder&, Shell::AST::Command value);
+    void format(TypeErasedFormatParams&, FormatBuilder&, const Shell::AST::Command& value);
 };
 
 }
~~~

Maybe you forgot to change the signature in the `.cpp` file? Otherwise, what error message do you get?

alimpfard: oh I see...the formatter's first template argument is cv-stripped.

alimpfard: Much better!

PR #3853, opened 2020-10-25
LibLine+Shell: Support HISTCONTROL and HISTFILE env vars / js: load and save history

linusg:
Details in commit messages. cc @alimpfard

### DIFF
--- a/Base/usr/share/man/man7/Shell-vars.md
+++ b/Base/usr/share/man/man7/Shell-vars.md
@@ -15,10 +15,16 @@ Such variables can be changed or set by the user to tweak how the shell presents
 
 The value of this variable is used to join lists or split strings into lists, its default value is a newline (`\\n`).
 
+2. History
+
+`HISTFILE` (environment)
+
+The value of this variable is used as the Shell's history file path, both for reading history at startup and writing history on exit.
+Its default value is `~/.history`.
 
 ## Visual
 
-2. Prompting
+1. Prompting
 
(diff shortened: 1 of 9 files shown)


Base/usr/share/man/man7/Shell-vars.md
```diff
@@ -17,7 +17,17 @@ The value of this variable is used to join lists or split strings into lists, it
 
 2. History
 
-`HISTFILE` (environment)
+`HISTCONTROL` (environment)
+
+The value of this variable is used to determine which entries are kept in the Shell's history, both regarding the current active session and when writing the history to disk on exit. It is unset by default, meaning no entries will be excluded from history.
```
alimpfard, line 22: > it is unset by default

but it being unset isn't documented...

linusg: > meaning no entries will be excluded from history

isn't that enough documentation for this case? :smile:

alimpfard: I'd just move the "meaning no entries..." to the list below under 'unset'.
I read that part when reading through and it didn't register at all...

linusg: cool, should be more clear now.

alimpfard: It is, thanks!

awesomekling:
Very cool!

Issue #1124, opened 2020-01-23
Userland: expr (needed to run configure scripts)


danboid:
Today I have been trying to get an autoconf configure script to run under the Serenity bash port. To get configure to (almost) run, you need to replace /bin/sh with a link to /usr/local/bin/bash but even then it won't run as it requires sed and expr.

Whilst sed won't be trivial to reimplement, reimplementing enough of expr to run configure scripts should be a bit more doable, I'd hope.

Some example uses from a configure script:

```
as_echo_body='eval expr "X$1" : "X\\(.*\\)"'
expr "X$arg" : "X\\(.*\\)$as_nl";
if expr a : '\(a\)' >/dev/null 2>&1 &&
ac_optarg=`expr "X$ac_option" : '[^=]*=\(.*\)'`
```

It would be much nicer having our own expr implementation versus having to port, build and install GNU coreutils for one command .

danboid:
If we're not writing this from scratch but porting an existing implementation, the OpenBSD one seems like the one to go for:

https://github.com/openbsd/src/blob/master/bin/expr/expr.c

I have a quick go at building it under serenity and there are two main issues. First we don't have a regex.h. Could this be lifted and shifted unmodified from OpenBSD?

https://github.com/openbsd/src/blob/master/include/regex.h

Secondly, the errors (err.h) need to translated from OpenBSD to Serenity (errno.h), plus a couple of other changes like working around the lack of strtonum().

PR #3858, opened 2020-10-26
Snowball PR, Shell: Fix and finally test backgrounding / Userland: Add 'expr' / AK: Add `StringView::is_one_of()' / Shell: name parts of patterns in `match'

alimpfard:
> \< thakis_\> CxByte, [...] maybe the backgrounding stuff needs more automated tests?

Let's try this.
~I cannot guarantee that this will not randomly fail the builds (but it seems to be ok on my system)~

As expected of something with zero tests, it's completely broken :P

cc @nico.

### DIFF
--- a/AK/String.cpp
+++ b/AK/String.cpp
@@ -270,6 +270,11 @@ String String::repeated(char ch, size_t count)
     return *impl;
 }
 
+bool String::matches(const StringView& mask, Vector<MaskSpan>& mask_spans, CaseSensitivity case_sensitivity) const
+{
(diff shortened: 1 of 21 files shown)


alimpfard:
Well that's not nice :sad:

alimpfard:
Huh.
`dc` isn't installed by default? TIL.

I guess giving #1124 a little visit wouldn't be too bad of an idea...

alimpfard:
I hate it when the compiler doesn't default my comparison operators :mad:

alimpfard:
Oh fantastic.
something deadlocks or is racy

PR #3859, opened 2020-10-26
LibMarkdown: Don't parse lines starting with a space as heading

linusg:
This fixes a bug where lines starting with a space would get parsed as "level 0" headings - it would not find a "#" and therefore never increase the level counter (starting at zero), which then would cause the check for "space after #" pass (again, there is no "#").

Eventually we would get funny results like this:

    <h0>[n-1 spaces]oops!</h0>

Also `ASSERT(level > 0)` in the `Heading` constructor.


alimpfard:
(cc @bugaevc)

bugaevc:
LGTM 🙂

PR #3861, opened 2020-10-26
LibJS: Unterminated multi line comments are a syntax error

linusg:
This fixes the annoying fact that we used to just accept unterminated multi-line comments (also 10 more test262-parser-tests tests).

I hope this concept is sane, if you'd rather want these to be individual invalid tokens we can do that, but I'd prefer not to.


awesomekling:
Seems fine this way. :)

PR #3863, opened 2020-10-27
LibJS: Use ES6 do-while ASI rule // Don't parse numeric literals containing 8 or 9 as octal

linusg:
Two smol parser fixes.


Libraries/LibJS/Token.cpp
```diff
@@ -86,7 +86,8 @@ double Token::double_value() const
             return static_cast<double>(strtoul(value_string.characters() + 2, nullptr, 2));
         } else if (isdigit(value_string[1])) {
             // also octal, but syntax error in strict mode
-            return static_cast<double>(strtoul(value_string.characters() + 1, nullptr, 8));
+            if (!m_value.contains('8') && !m_value.contains('9'))
+                return static_cast<double>(strtoul(value_string.characters() + 1, nullptr, 8));
```
awesomekling, line 90: Very interesting!

PR #3864, opened 2020-10-28
More SMP related fixes

tomuta:
Not quite stable yet, crashes pretty easily still. But slowly getting there. This can somewhat schedule threads on all processors now, though for some odd reason it seems to prefer the first cpu and doesn't evenly distribute the workload. Baby steps...

Screenshot from 2020-10-27 21-02-21

EDIT: The bias for CPU#0 is because of this line, but removing it triggers the other crashes much more readily:
https://github.com/SerenityOS/serenity/blob/cff734dadd975b1d5ee5d321c00a7a4110a11662/Kernel/Scheduler.cpp#L872

### DIFF
--- a/AK/LogStream.cpp
+++ b/AK/LogStream.cpp
@@ -129,7 +129,7 @@ DebugLogStream dbg()
 #endif
 #if defined(__serenity__) && defined(KERNEL)
     if (Kernel::Processor::is_initialized() && Kernel::Thread::current())
-        stream << "\033[34;1m[" << *Kernel::Thread::current() << "]\033[0m: ";
+        stream << "\033[34;1m[#" << Kernel::Processor::current().id() << " " << *Kernel::Thread::current() << "]\033[0m: ";
     else
         stream << "\033[36;1m[Kernel]\033[0m: ";
 #endif
@@ -141,7 +141,7 @@ KernelLogStream klog()
 {
     KernelLogStream stream;
     if (Kernel::Processor::is_initialized() && Kernel::Thread::current())
-        stream << "\033[34;1m[" << *Kernel::Thread::current() << "]\033[0m: ";
+        stream << "\033[34;1m[#" << Kernel::Processor::current().id() << " " << *Kernel::Thread::current() << "]\033[0m: ";
     else
         stream << "\033[36;1m[Kernel]\033[0m: ";
     return stream;
--- a/Kernel/Arch/i386/CPU.cpp
+++ b/Kernel/Arch/i386/CPU.cpp
@@ -1739,6 +1739,57 @@ ProcessorMessage& Processor::smp_get_from_pool()
     return *msg;
 }
 
+Atomic<u32> Processor::s_idle_cpu_mask{ 0 };
+
+u32 Processor::smp_wake_n_idle_processors(u32 wake_count)
+{
+    ASSERT(Processor::current().in_critical());
+    ASSERT(wake_count > 0);
+    if (!s_smp_enabled)
+        return 0;
+
+    // Wake at most N - 1 processors
+    if (wake_count >= Processor::count()) {
+        wake_count = Processor::count() - 1;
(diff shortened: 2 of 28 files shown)


awesomekling:
Very interesting development! :star:

tomuta:
Most crashes on boot are now exactly the same as #3885, but in about 1 in 10 attempts it will actually boot to the desktop and is usable for a while until you run into the same bug. And in SystemManager (and in the kernel logs) you can see some threads on the other cpu occasionally now.

tomuta:
If anyone is bored, some help trying to track down #3885 would be much appreciated.

9/10 boots fail with this bug in this branch. It seems that somehow two kernel regions with the same range are created, causing a collision because the page directory is already mapped.

Lubrsi:
For me, it always fails to boot with this:
```
[#0 init_stage2(1:1)]: PATAChannel: PATA Controller found, ID (8086:7010)
[#0 init_stage2(1:1)]: PATAChannel: Bus master IDE: IO c080
[#0 init_stage2(1:1)]: PATAChannel: Name=QEMU HARDDISK, C/H/Spt=26/16/63
[#0 init_stage2(1:1)]: ASSERTION FAILED: !(m_bits & 1)
.././AK/NonnullRefPtr.h:87 in AK::NonnullRefPtr<T>::NonnullRefPtr(AK::NonnullRefPtr<T>::AdoptTag, T&) [with T = Kernel::PATADiskDevice]
[#0 init_stage2(1:1)]: c011ee8d (next: c0928338)
[#0 init_stage2(1:1)]: c012bed2 (next: c0928368)
[#0 init_stage2(1:1)]: c012b4a6 (next: c0928498)
[#0 init_stage2(1:1)]: c012ba04 (next: c09284e8)
[#0 init_stage2(1:1)]: c012baa6 (next: c0928548)
[#0 init_stage2(1:1)]: c01a91be (next: c0928ff8)
[#0 init_stage2(1:1)]: 0 (next: 0)
```

alimpfard:
> For me, it always fails to boot with this

That seems to be because `operator new()` for _eternal_ objects does not round the size up, so we end up with uneven pointers (as `kmalloc_eternal()` doesn't round it up either).

```diff
--- a/AK/kmalloc.h
+++ b/AK/kmalloc.h
@@ -33,7 +33,7 @@
 #ifdef KERNEL
 #    define AK_MAKE_ETERNAL                                               \
     public:                                                               \
-        void* operator new(size_t size) { return kmalloc_eternal(size); } \
+        void* operator new(size_t size) { return kmalloc_eternal(round_up_to_power_of_two(size, 2)); } \
                                                                           \
     private:
```

should fix _that_, I think

tomuta:
> For me, it always fails to boot with this:

Interesting, I've only run into this once so far and I couldn't replicate it.

@alimpfard shouldn't it rather align the pointer than the size? Or is this to fix pointers in arrays?

tomuta:
@alimpfard I created #3900 with your change, slightly different but almost the same.

tomuta:
@Lubrsi I added the fix for that crash to this PR.

tomuta:
Hah I found it... https://github.com/SerenityOS/serenity/pull/3864/commits/2f16e48e5a0813e30c6c03c8186f4d11faba6822 isn't quite right ;-) Has a subtle but fatal logic flaw...

tomuta:
Ok I think the PTE assertion is fixed now, at least I'm no longer running into it. Still either hangs occasionally due to some other odd bugs here and there, but it doesn't take 10 attempts anymore to boot it successfully :) And Browser now starts without crashing it due to this bug.

Lubrsi:
Confirmed fixed 👍 , though I'm not sure how to get multiple cores to show up, changing SMP in run.sh does nothing.

alimpfard:
@Lubrsi pass `smp=on` as a kernel param, e.g. `SERENITY_KERNEL_CMDLINE="smp=on"`

PR #3865, opened 2020-10-28
LibJS: Use GenericLexer for Token::string_value()

linusg:


Libraries/LibJS/Token.cpp
```diff
@@ -103,165 +103,111 @@ static u32 hex2int(char x)
...
+        if (!octal_str.is_null()) {
+            status = StringValueStatus::LegacyOctalEscapeSequence;
+            auto code_point = strtoul(octal_str.characters(), nullptr, 8);
+            ASSERT(code_point <= 255);
+            builder.append_code_point(code_point);
+            continue;
         }
+
+        lexer.retreat();
+        builder.append(lexer.consume_escaped_character('\\', "b\bf\fn\nr\rt\tv\v"));
```
nico, line 210: This feels a bit awkward. Maybe the start of the loop could instead do:

    char escaped = lexer.consume_escaped_character('\\', "b\bf\fn\nr\rt\tv\v")

instead of `lexer.ignore()` and the loop contents could look at `escaped` instead of `lexer.peek()` and this here could do `builder.append(escaped)`? Then you could also drop a few `.ignore()` calls, and we wouldn't need this `retreat()` api.

linusg: This doesn't work well with the octal number parsing though or am I missing something? `consume_escaped_character()` will have consumed the first digit so we'd still have to go back or get the possibly remaining digits and build a string for `strtoul()` - also awkward. 🤔

linusg: Ok, I tried this by moving the octal number parsing code further up, but it still does not work conveniently - because of this, which depends on the backslash being consumed:

https://github.com/SerenityOS/serenity/blob/755fda193bcd39ef96c4c300804e23dfadf33fae/Libraries/LibJS/Token.cpp#L131-L135

Suggestions welcome but I'd rather not overcomplicate this.

PR #3867, opened 2020-10-29
LibTLS(+ProtocolServer): Root CAs and certificate verification - pt1

alimpfard:
This PR finally makes LibTLS verify the certificate it's given (as in 'is this valid?'), and partially validate it against a very basic root CA store.
Issues with this PR:
- Anyone can claim to be signed by a root CA, we do _not_ yet verify this claim (i.e. no checking the signatures)
- The root CA store is likely incomplete
- The CA store format is very silly
- There are no blacklisted CAs, nor any support for them

but in general, I consider this a step forward~


Base/etc/ca_certs.ini
```diff
@@ -0,0 +1,465 @@
...
+country=US
+
+[Symantec Trust Network]
+country=US
+
+[Symantec Trust Network]
+country=US
+
+[Symantec Trust Network]
+country=US
```
dkoao, line 372: It'd be nice to not trust Symantec. They are a shady CA.
https://security.googleblog.com/2017/09/chromes-plan-to-distrust-symantec.html

(I know this doesn't matter right now because the signatures aren't verified, but it'll be relevant when Serenity starts doing that.)

alimpfard: Ideally, we'd run our own trust program!
I think, for the time being, we can afford to let this be, and when we implement a blacklist for these, add Symantec to that.

Services/ProtocolServer/HttpsProtocol.h
```diff
@@ -26,16 +26,20 @@
...
 class HttpsProtocol final : public Protocol {
 public:
-    HttpsProtocol();
+    HttpsProtocol(const Vector<Certificate>&);
     virtual ~HttpsProtocol() override;
 
     virtual OwnPtr<Download> start_download(ClientConnection&, const String& method, const URL&, const HashMap<String, String>& headers, const ByteBuffer& request_body) override;
+
+private:
+    const Vector<Certificate>& m_ca_certificates;
```
awesomekling, line 42: This feels a bit sloppy. How about we make a certificate manager that `HttpsProtocol` and `GeminiProtocol` can access through a singleton instance getter (or similar)?

alimpfard: For now, I went with something lower-level, a singleton `DefaultRootCACertificates`, and the *Protocol code remains mostly unchanged.
the Jobs just take an optional override param, or default to the...default.

If at some point we decide to allow the user to pick between different sets of root CAs, we'll change to your suggested certificate manager.

PR #3873, opened 2020-10-29
LibChess: Check for check by pawn on last rank

### LINKED ISSUE #3525: Chess: ChessEngine may just ignore a check
This is how the board was setup when this happened (using lichess to recreate it):
image

ChessEngine loves moving the rook between h8 and g8 every chance it gets to, completely ignoring everything else, even checks!
In this situation, ChessEngine moved from the rook from g8 to h8, ignoring the check. I was then allowed to capture the king.

image

Even when the king is captured, it still loves that rook move.

### LINKED ISSUE #3526: Chess: ChessEngine may resolve a check by checking itself
Here's the board setup:
image

Here's what ChessEngine did:
image

bcoles:
The King piece can no longer move into check from a pawn on the last
rank or remain in check from a pawn when on the last rank by playing
another move.

Resolves #3525
Resolves #3526


bcoles:
In retrospect, this breaks promotion. I'll take a look at this later.

Issue #3632, opened 2020-09-28
Browser: Disallow local URLs in XMLHttpRequest


xTibor:
I hope I'm not as controversial with this one as the port blacklisting one.

**Reproduction:**
```
<!DOCTYPE html>
<html>
<head>
<script>
function reqListener () {
    document.getElementById("logs").innerText = this.responseText;
}

var oReq = new XMLHttpRequest();
oReq.addEventListener("load", reqListener);
oReq.open("GET", "file:///etc/passwd");
oReq.send();
</script>
</head>
<body>
    <pre id="logs"></pre>
</body>
</html>
```

![Screenshot_20200928_171902](https://user-images.githubusercontent.com/1627292/94451693-0170b400-01af-11eb-8abf-d6e476ea62b3.png)

bcoles:
Dupe of #1615. Related #1616

xTibor:
Does making requests for local files from `file:/` URLs considered same origin? If so, there's a slight difference between #1615 and this.

bcoles:
> 
> 
> Does making requests for local files from `file:/` URLs considered same origin? If so, there's a slight difference between #1615 and this.

In the context of SOP, an origin is defined as a combination of URI scheme, host name, and port number. The URL scheme is part of the origin, although `file://` is special.

> If so, there's a slight difference between #1615 and this.

The same issue related to file theft is described in https://github.com/SerenityOS/serenity/issues/1615#issuecomment-610782263

xTibor:
I looked further into this. `file://` is not a special case, it's undefined behavior according to the specs (by not having meaningful hostname/port). Closing this in favor however you want to handle this is #1615.

bcoles:
> 
> 
> > Does making requests for local files from `file:/` URLs considered same origin? If so, there's a slight difference between #1615 and this.
> 
> In the context of SOP, an origin is defined as a combination of URI scheme, host name, and port number. The URL scheme is part of the origin, although `file://` is special.
> 
> > If so, there's a slight difference between #1615 and this.
> 
> The same issue related to file theft is described in [#1615 (comment)](https://github.com/SerenityOS/serenity/issues/1615#issuecomment-610782263)

None the less, URLs with schemes such as `file:` and `data:` are special.

There's value in tracking this in a separate issue to #1615, although any implemented mitigation should be performed with the SOP model in mind as part of the SOP implementation, rather than an ad-hoc patch.

[...]

PR #3880, opened 2020-10-31
LibWeb: Enforce Same-Origin Policy (SOP) for XMLHttpRequest requests

### LINKED ISSUE #1615: Browser: Same Origin Policy
The `Browser` application and associated `LibJS` do not yet enforce the Same Origin Policy. It is probably time to start thinking about it.

bcoles:
`DOM::XMLHttpRequest` now checks if the requested URL has the same
`Origin` as the requesting `Document`. If the requested URL is in
violation of SOP the request is rejected and an "error" `DOM::Event`
is dispatched.

Resolves #1615

---

This is a rudimentary implementation. It is restrictive. CORS requests are not supported.

I added a `dbg()` debugging statement as, given that `Browser` has not enforced SOP until now, I presume rejecting requests due to SOP violations will be unexpected and confusing for developers.

Perhaps most importantly, this change prevents random websites stealing files (https://github.com/SerenityOS/serenity/issues/1615#issuecomment-610782263 #3632).

image

```html
<html>
  <head>
    <title>hello friends</title>
  </head>
  <body>
    Logs 1: <pre id="logs1"></pre>
    <hr/>
    Logs 2: <pre id="logs2"></pre>
    <hr/>
    Logs 3: <pre id="logs3"></pre>
    <script>
    try {
      var xhr = new XMLHttpRequest();
      xhr.addEventListener("load", function() { document.getElementById("logs1").innerText = this.responseText; });
      xhr.open("GET", "http://172.16.191.165/allowed.txt");
      xhr.send();
    } catch(e) {
      alert('error:' + e.message);
    }

    try {
      var xhr = new XMLHttpRequest();
      xhr.addEventListener("load", function() { document.getElementById("logs2").innerText = this.responseText; });
      xhr.addEventListener("error", function(error) { document.getElementById("logs3").innerText = JSON.stringify(error); });
      xhr.open("GET", "file:///etc/passwd");
      xhr.send();
    } catch(e) {
      alert('error:' + e.message);
    }
    </script>
  </body>
</html>
```

### DIFF
--- a/Libraries/LibWeb/DOM/XMLHttpRequest.cpp
+++ b/Libraries/LibWeb/DOM/XMLHttpRequest.cpp
@@ -35,6 +35,7 @@
 #include <LibWeb/DOM/Window.h>
 #include <LibWeb/DOM/XMLHttpRequest.h>
 #include <LibWeb/Loader/ResourceLoader.h>
+#include <LibWeb/Origin.h>
 
 namespace Web {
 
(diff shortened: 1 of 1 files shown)


bcoles:
I forgot that `Document::origin()` has since been developed. I'll leave this PR open for discussion anyway.

I'm not sure if there's any reason we would want to use `is_same_origin` outside of `Document` context (?). In which case, perhaps it would make more sense to move the function from `ResourceLoader.is_same_origin(origin, destination)` to `Origin.is_same_origin(url)`.

In which case, there's an existing implementation in `Document::Origin.is_same()`:

```cpp
    bool is_same(const Origin& other) const
    {
        return protocol() == other.protocol()
            && host() == other.host()
            && port() == other.port();
    }
```

Although this policy is even more restrictive and has room for improvement.

bcoles:
Updated to make use of `Document::origin()` when determining the origin of the requested URL.

awesomekling:
Seems like a fine start :)

PR #3884, opened 2020-10-31
Kernel: Move timers from scheduler

tomuta:
In order to make more meaningful progress with #3864 there are two major things that need re-architecturing:

1. Timers/timeouts should not be handled in the `Scheduler`
2. Block conditions should not be evaluated in the `Scheduler`

The reason is that the scheduler needs to be as lean as possible. But not only that, doing these things in the Scheduler poses some major problems because they involve a lot of locks, which means there is high potential for deadlocks with multiple processors accessing them.

This PR aims to solve 1) only.

- [x] Figure out why timers (`select` timeouts?) don't fire as expected, e.g. clock applet doesn't update, Fire application doesn't render unless you move the mouse over the window, etc
- [x] Fix system time being off (0 epoch)


tomuta:
Pulled this into #4113, so one PR less to maintain...

PR #3887, opened 2020-10-31
Kernel: Map PCI devices only once during boot

supercomputer7:
In general, it makes more sense (at least to me) to not deal with mapping of a PCI device configuration space (when using the PCI MMIO method) each time we need to access it, but to have it already mapped somewhere in virtual memory space so we can access it almost immediately if needed.

Opinions, suggestions or fixes are very appreciated! :)


supercomputer7:
Just a note on this PR - We should be able to put a lock in the `get_device_configuration_space` method, so only CPU can modify/view a PCI device's configuration space at a time. For PCI IO access, we should have a similar thing, if the IO operation doesn't block the access to the PCI IO port from other CPUs.

PR #3888, opened 2020-10-31
Shell: Some bug fixes

alimpfard:
Just tiny bug fixes, nothing interesting.
- Don't crash when allocating more than 4 frames (!)
- Don't overwrite other frames' variables
- Don't leak ARGV into other function frames

Also names the frames, pretty handy when debugging, might be interesting to have decent stack traces

### DIFF
--- a/Shell/Shell.h
+++ b/Shell/Shell.h
@@ -103,8 +103,8 @@ public:
     RefPtr<AST::Value> get_argument(size_t);
     RefPtr<AST::Value> lookup_local_variable(const String&);
     String local_variable_or(const String&, const String&);
-    void set_local_variable(const String&, RefPtr<AST::Value>);
-    void unset_local_variable(const String&);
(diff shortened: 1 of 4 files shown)


Shell/Shell.h
```diff
@@ -247,7 +254,7 @@ class Shell : public Core::Object {
     };
 
     HashMap<String, ShellFunction> m_functions;
-    Vector<LocalFrame> m_local_frames;
+    Vector<OwnPtr<LocalFrame>> m_local_frames;
```
awesomekling, line 257: If you don't intend to store null values, use `NonnullOwnPtr` instead of `OwnPtr`.
You can also use `NonnullOwnPtrVector` for extra serenityness :)

alimpfard: Ah yes, that one pattern I always forget.

PR #3890, opened 2020-10-31
Travis: Update host GCC to version 10

linusg:
This allows us to use the latest C++20 features in programs which we compile and run as part of Lagom.

cc @alimpfard

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,9 +24,12 @@ notifications:
 before_install:
 - sudo apt-get update -qq
 # These packages are already part of the Travis image:
-#- sudo apt-get install -y g++-9 libstdc++-9-dev cmake shellcheck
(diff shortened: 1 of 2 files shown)


alimpfard:
One concern, I am not sure if everyone is using gcc 10, this might mean forcing them to upgrade or install gcc 10.
while I personally find that a good thing (why would someone keep using old compilers on the machines they use daily?), it's worth at least thinking about

awesomekling:
This seems like a good move, but let's also update the build instructions to say you need GCC >= 10 :)

linusg:
@alimpfard @awesomekling I think there's a middle ground here - I assume the majority of Lagom users are Serenity devs, so we shouldn't force GCC 10 on everyone unless required for building the toolchain. I added a note to Lagom's README instead. Ok?

PR #3898, opened 2020-11-01
SoundPlayer: Optionally allow playback to loop indefinitely

bcoles:
Now you can listen to your Piano creations on loop once exported.

image


awesomekling:
At last :)

PR #3918, opened 2020-11-02
HackStudio: Removes the file from the list of open files

kessejones:
When a file is deleted from the project it will also be removed from the list of open files


awesomekling:
We shouldn't close a file just because you remove it from the project. It's okay to have open files that are not part of the project.

kessejones:
I also have no problems opening files that are not part of the project. But for me, if I removed it from the project, it is because I no longer need it.

Issue #3906, opened 2020-11-01
Spreadsheet: Select, copy and paste cells


danboid:
At the moment it doesn't seem possible to select more than one cell in Spreadsheet. I just tried and it crashed.

When that works (also selecting entire columns and rows) it would be great, essential even, to be able to copy and paste ranges of cells.

WOULDN'T IT!?

Yes.

alimpfard:
It should be possible to select more than one cell, I guess the crash is due to some recent change in LibGUI or some such.

And copy-pasting _would_ be rather neat

danboid:
There are very few features I do need in spreadsheet but copy/paste is definitely one of them.

PR #3920, opened 2020-11-02
LibGUI: Shift-select / Spreadsheet: Copy/Paste and Drag'n'Drop

alimpfard:

### DIFF
--- a/Applications/Spreadsheet/Position.h
+++ b/Applications/Spreadsheet/Position.h
@@ -28,6 +28,7 @@
 
 #include <AK/String.h>
 #include <AK/Types.h>
+#include <AK/URL.h>
 
 namespace Spreadsheet {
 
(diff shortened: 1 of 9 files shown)


alimpfard:
Kinda sorta deals with #3906, but I'd like to keep that issue open until we can copy/paste ranges to other ranges too.

Applications/Spreadsheet/Position.h
```diff
@@ -44,6 +45,16 @@ struct Position {
     {
         return !(other == *this);
     }
+
+    URL to_uri() const
```
awesomekling, line 49: You are using URI and URL somewhat interchangeably here. Let's settle on one, how about just URL?

alimpfard: I just can't see a URL with a custom protocol and scheme as a URL, so I tend to think of it as a URI...
I'll swap over to URL.

PR #3921, opened 2020-11-02
Kernel: Implement an asynchronous device request stack

tomuta:
This allows issuing asynchronous requests for devices and waiting
on the completion of the request. The requests can cascade into
multiple sub-requests.

Change the PATA driver to use this model.

This is to solve the problem of having to call `Thread::wait_on` from IRQ handlers, which has been nothing but trouble (and as far as I know, no other OS allows preemption within irq handlers anyway). This is needed to be able to make more meaningful progress with #3864.

### DIFF
--- a/Kernel/Devices/AsyncDeviceRequest.cpp
+++ b/Kernel/Devices/AsyncDeviceRequest.cpp
@@ -1,0 +1,175 @@
+/*
+ * Copyright (c) 2020, The SerenityOS developers.
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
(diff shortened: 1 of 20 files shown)


Kernel/Devices/AsyncDeviceRequest.cpp
```diff
@@ -0,0 +1,169 @@
...
+
+AsyncDeviceRequest::~AsyncDeviceRequest()
+{
+    {
+        ScopedSpinLock lock(m_lock);
+        ASSERT(is_completed_result(m_result));
+        ASSERT(m_sub_requests_pending.is_empty());
+    }
+
+    for (auto& sub_request : m_sub_requests_complete) {
```
awesomekling, line 46: Wait, is it safe to iterate over `m_sub_requests_complete` without holding `m_lock`?
The locking here is a bit confusing.

tomuta: It should. I have not added a cancel operation yet, but the destructor should not be called until either `AsyncDeviceRequest::wait` or `AsyncDeviceRequest::cancel` (once it exists) returned, in either case no sub-requests should be marking any requests as complete anymore (they should have finished or failed, or if cancelled moved to the complete list), and all pending requests should have moved to the complete list. in other words, at that point nothing should be mocking with any of these lists and `m_result` should be final. I'll add a comment.

tomuta: Found another small issue (which I think would only manifest itself with more complex cascaded requests) and added a comment for this.

awesomekling:
This breaks booting without SMP:
```
[colonel(0:0)]: ASSERTION FAILED: handler.type() == HandlerType::SpuriousInterruptHandler
../Kernel/Interrupts/PIC.cpp:95 in virtual void Kernel::PIC::spurious_eoi(const Kernel::GenericInterruptHandler&) const
[colonel(0:0)]: c011eec3 (next: c0917e88)
[colonel(0:0)]: c014e37b (next: c0917ea8)
[colonel(0:0)]: c014e62f (next: c0917ee8)
[colonel(0:0)]: c014d089 (next: c0917f08)
[colonel(0:0)]: c011e023 (next: c0917ff8)
[colonel(0:0)]: 0 (next: 0)
kling@qncvs:~/src/serenity/Build$ 
```

tomuta:
Dang, I didn't think about testing that configuration...

PR #3923, opened 2020-11-02
LibJS: Implement 'new.target'

linusg:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/new.target

Also did some cleanup around syntax error line/columns.

### DIFF
--- a/Libraries/LibJS/Tests/functions/function-new-target.js
+++ b/Libraries/LibJS/Tests/functions/function-new-target.js
@@ -1,0 +1,26 @@
+test("basic functionality", () => {
+    function foo() {
+        return new.target;
+    }
+    expect(foo()).toBeUndefined();
+    expect(new foo()).toEqual(foo);
+
+    function bar() {
+        const baz = () => new.target;
+        return baz();
+    }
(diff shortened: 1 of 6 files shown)


Libraries/LibJS/Tests/functions/function-new-target.js
```diff
@@ -0,0 +1,26 @@
...
+            this.newTarget = new.target;
+        }
+    }
+    expect(new baz().newTarget).toEqual(baz);
+});
+
+// FIXME: This does not work as expected as toEval() places the code inside a function :|
+// test("syntax error outside of function", () => {
+//     expect("new.target").not.toEval();
+// });
```
linusg, line 26: does this violate the no commented-out code rule?! :D

awesomekling: Shouldn't this be a skipped test rather than a comment tho?

linusg: hah good point. fixed & resolves the question :)

awesomekling:
Nicely implemented!

PR #3927, opened 2020-11-03
LibGUI: Clear selection after focusin event

### LINKED ISSUE #3925: IconView: Focused item inadvertently selected
This seems a regression related to selecting the focused item. The fist item should not be selected when I start the rubber band with the mouse, which results in the first item being deselected.

iconview_selection

Normally, when a rubber band selection is started, unless the shift key is pressed, it should clear the selection. I suspect it does that, but because the mouse click also sets focus, it then adds the first item to the selection *after* clearing it.

### LINKED ISSUE #3926: IconView: Rubber band not clipped to widget content area
The rubber band is drawn over the scrollbar and also over the widget border.

Screenshot from 2020-11-02 16-52-33

AmusedNetwork:
After moving the cursor to the home position, clear the selection.

Fixes #3925.

### DIFF
--- a/Libraries/LibGUI/AbstractView.cpp
+++ b/Libraries/LibGUI/AbstractView.cpp
@@ -710,8 +710,10 @@ void AbstractView::focusin_event(FocusEvent& event)
 {
     ScrollableWidget::focusin_event(event);
 
-    if (model() && !cursor_index().is_valid())
+    if (model() && !cursor_index().is_valid()) {
         move_cursor(CursorMovement::Home, SelectionUpdate::None);
+        clear_selection();
+    }
 }
 
 }


Libraries/LibGUI/AbstractView.cpp
```diff
@@ -691,8 +691,10 @@ void AbstractView::focusin_event(FocusEvent& event)
 {
     ScrollableWidget::focusin_event(event);
 
-    if (model() && !cursor_index().is_valid())
-        move_cursor(CursorMovement::Home, SelectionUpdate::None);
+    if (model() && !cursor_index().is_valid()) {
+        move_cursor(CursorMovement::Home, SelectionUpdate::ClearIfNotSelected);
```
awesomekling, line 695: Why is it necessary to change the `SelectionUpdate` mode here? I would think that `SelectionUpdate::None` works fine, as long as we `clear_selection()` after, no?

awesomekling (changes requested):
The commit message makes it sound like we always clear the selection after focusing [some unspecified LibGUI widget].
It's good to be a bit more specific, e.g "LibGUI: Fix wonky AbstractView selection after initial focusin event"
This helps us months from now, when we're skimming through the git log and trying to understand what changed :)

Libraries/LibGUI/AbstractView.cpp
AmusedNetwork: Oh right, that was left in by accident, I’ll change it

PR #3935, opened 2020-11-03
HackStudio: Store and restore the scrollbar values of ProjectFiles wh…

supex0fan:
…en they are closed and reopened from the list of open files.


awesomekling:
Great idea :)

PR #3938, opened 2020-11-03
LibJS: Implement spec-compliant OrdinaryToPrimitive

linusg:
https://tc39.es/ecma262/#sec-ordinarytoprimitive

### DIFF
--- a/Libraries/LibJS/Runtime/Object.cpp
+++ b/Libraries/LibJS/Runtime/Object.cpp
@@ -836,48 +836,30 @@ bool Object::has_own_property(const PropertyName& property_name) const
     return shape().lookup(property_name.to_string_or_symbol()).has_value();
 }
 
-Value Object::to_primitive(Value::PreferredType preferred_type) const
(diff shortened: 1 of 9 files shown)


Libraries/LibJS/Runtime/Object.cpp
```diff
@@ -836,48 +836,28 @@ bool Object::has_own_property(const PropertyName& property_name) const
...
-        return string;
+
+    Vector<FlyString, 2> method_names;
+    if (preferred_type == Value::PreferredType::String)
+        method_names = { vm.names.toString, vm.names.valueOf };
+    else
+        method_names = { vm.names.valueOf, vm.names.toString };
+
+    for (auto& method_name : method_names) {
+        auto method = get(method_name);
```
awesomekling, line 852: Hmm.. do we need an exception check after this `get()`?

linusg: haha here I am always making sure we check for exceptions and now I'm missing one. :upside_down_face:

linusg:
CI failing for completely unrelated reasons now. :roll_eyes:

Issue #3644, opened 2020-09-29
Kernel: unveil shouldn't fail on non-existing paths when "c" permission is set


xTibor:
It currently fails with `ENOENT`. Don't know how it works on OpenBSD but common sense says to me that if you unveil a path for creating something there may not exist anything there yet. And I think that should be considered normal.

**Reproduction:**

You can hit this case with the Terminal app. For example if you want to reset its config.

1. Remove the Terminal config file from `/home/anon/.config/Terminal.ini`.
2. Open a new Terminal instance.
3. Fail on this `unveil`: https://github.com/SerenityOS/serenity/blob/a9335eea1c96d50140e46b8e0361e01b87c09ce2/Applications/Terminal/main.cpp#L376-L379
4. Terminal would recreate its config file if that `unveil` wouldn't fail above: https://github.com/SerenityOS/serenity/blob/a9335eea1c96d50140e46b8e0361e01b87c09ce2/Applications/Terminal/main.cpp#L383

xTibor:
Checked the behavior on OpenBSD. It allows unveiling non-existing files if their directories exist.

IdanHo:
This was fixed in #3955

PR #3942, opened 2020-11-04
Chess: Add missing default config file in .config :^)

Quaker762:
Following 9f8a8e07c2d25009268830c51f58c052a8f6936b, let's add a default
`.ini` to prevent `unveil` from failing.

### DIFF
--- a/Base/home/anon/.config/Chess.ini
+++ b/Base/home/anon/.config/Chess.ini
@@ -1,0 +1,7 @@
+[Display]
+size=512
+
+[Style]
+BoardTheme=Beige
+PieceSet=stelar7
+Coordinates=1


bcoles:
This is necessary due to #3644.

Quaker762:
> This is necessary due to #3644.

Ahh, thanks for the heads up on this one, I didn't even realise it'd been unveiled with `c` permissions.

bcoles:
This one is my bad. I didn't commit my `Chess.ini` config file in #3907 (files in `./Base/home/anon/.config` are ignored by git).

As it happens, I also botched the config loading, which I've fixed in #3945.

bcoles:
> I'll fix #3644 while I'm at it.

If you fix this issue, we can remove the default INI files for various Games from the `anon` home directory (and maybe some of the apps). There's no need to ship these, presuming that the application utilizes reasonable defaults.

Quaker762:
> > I'll fix #3644 while I'm at it.
> 
> If you fix this issue, we can remove the default INI files for various Games from the `anon` home directory (and maybe some of the apps). There's no need to ship these, presuming that the application utilizes reasonable defaults.

Isn't it good practice to supply default configuration files though? I think most applications shipped write a default conf file in `.config` on Linux, but this might be more arguing semantics. Just my 2c.

bcoles:
> I think most applications shipped write a default conf file in `.config` on Linux, but this might be more arguing semantics. Just my 2c.

Only when they're installed, and even then, the config file is usually not created until the application is run for the first time. Unsurprisingly, operating system distros generally do not ship config files for applications which are not installed.

There's (currently) no installation system inside SerenityOS. Removing applications from the build process will still include the config file in the user's home directory as `./Base` is unrelated to each application's `make` file.

PR #3955, opened 2020-11-05
Kernel: Prevent `unveil` returning ENOENT with cpath permissions

Quaker762:
This addresses the issue first enountered in #3644. If a path is
first unveiled with "c" permissions, we should NOT return ENOENT
if the node does not exist on the disk, as the program will most
likely be creating it at a later time.

### DIFF
--- a/Kernel/Syscalls/unveil.cpp
+++ b/Kernel/Syscalls/unveil.cpp
@@ -24,6 +24,7 @@
  * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
+#include <AK/LexicalPath.h>
 #include <AK/StringView.h>
 #include <Kernel/FileSystem/Custody.h>
 #include <Kernel/FileSystem/VirtualFileSystem.h>
@@ -58,18 +59,11 @@ int Process::sys$unveil(Userspace<const Syscall::SC_unveil_params*> user_params)
     if (path.value().is_empty() || path.value().characters()[0] != '/')
         return -EINVAL;
 
-    auto custody_or_error = VFS::the().resolve_path_without_veil(path.value(), root_directory());
-    if (custody_or_error.is_error())
-        // FIXME Should this be EINVAL?
-        return custody_or_error.error();
-
-    auto& custody = custody_or_error.value();
-    auto new_unveiled_path = custody->absolute_path();
-
     auto permissions = copy_string_from_user(params.permissions);
     if (permissions.is_null())
         return -EFAULT;
 
+    // Let's work out permissions first...
     unsigned new_permissions = 0;
     for (const char permission : permissions) {
         switch (permission) {
@@ -88,6 +82,24 @@ int Process::sys$unveil(Userspace<const Syscall::SC_unveil_params*> user_params)
         default:
             return -EINVAL;
         }
+    }
+
+    // Now, let's try and resolve the path and obtain custody of the inode on the disk, and if not, bail out with
+    // the error from resolve_path_without_veil()
+    // However, if the user specified unveil() with "c" permissions, we don't set errno if ENOENT is encountered,
+    // because they most likely intend the program to create the file for them later on.
+    // If this case is encountered, the parent node of the path is returned and the custody of that inode is used instead.
+    RefPtr<Custody> parent_custody; // Parent inode in case of ENOENT
+    String new_unveiled_path;
+    auto custody_or_error = VFS::the().resolve_path_without_veil(path.value(), root_directory(), &parent_custody);
+    if (!custody_or_error.is_error()) {
+        new_unveiled_path = custody_or_error.value()->absolute_path();
+    } else if (custody_or_error.error() == -ENOENT && parent_custody && (new_permissions & UnveiledPath::Access::CreateOrRemove)) {
+        String basename = LexicalPath(path.value()).basename();
+        new_unveiled_path = String::formatted("{}/{}", parent_custody->absolute_path(), basename);
+    } else {
+        // FIXME Should this be EINVAL?
+        return custody_or_error.error();
     }
 
     for (auto& unveiled_path : m_unveiled_paths) {


Quaker762:
Note that this also fixes the issues seen in some games as discussed in #3942. These default `.ini` files can probably be removed now, as the program should specify the default parameters itself without needing to read anything from the config file.

Kernel/Syscalls/unveil.cpp
```diff
@@ -90,6 +83,28 @@ int Process::sys$unveil(Userspace<const Syscall::SC_unveil_params*> user_params)
...
+    // If this case is encountered, the parent node of the path is returned and the custody of that inode is used instead.
+    RefPtr<Custody> custody_on_error; // Parent inode in case of ENOENT
+    auto custody_or_error = VFS::the().resolve_path_without_veil(path.value(), root_directory(), &custody_on_error);
+    if (custody_or_error.is_error())
+        if (!(new_permissions & UnveiledPath::Access::CreateOrRemove) && custody_or_error.error() == -ENOENT)
+            // FIXME Should this be EINVAL?
+            return custody_or_error.error();
+
+    RefPtr<Custody> custody;
+    if (custody_or_error.is_error())
```
Quaker762, line 99: @awesomekling I feel like this could probably be a bit neater, but I can't use a ternary due to one being a `RefPtr` and the other being a `NonNullRefPtr`

awesomekling:
cc @bugaevc

Kernel/Syscalls/unveil.cpp
```diff
@@ -90,6 +83,25 @@ int Process::sys$unveil(Userspace<const Syscall::SC_unveil_params*> user_params)
         }
     }
 
+    // Now, let's try and resolve the path and obtain custody of the inode on the disk, and if not, bail out with
+    // the error from resolve_path_without_veil()
+    // However, if the user specified unveil() with "c" permissions, we don't set errno if ENOENT is encountered,
+    // because they most likely intend the program to create the file for them later on.
+    // If this case is encountered, the parent node of the path is returned and the custody of that inode is used instead.
+    RefPtr<Custody> custody_on_error; // Parent inode in case of ENOENT
```
bugaevc, line 91: A `custody_or_error` should be of type `KResult<NonnullRefPtr<Custody>>`. Either rename the variable or change the type :)

bugaevc: Ah, it's custody *on* error!

That's very hard to see given how the `_or_error` names are used everywhere else.

Why not just name it `parent_custody` or something?

Quaker762: > Why not just name it parent_custody or something?

Good idea, I was having trouble coming up with a name that would describe what this variable actually related to, but that should make the most sense, considering someone can just look at the definition of `resolve_path_without_veil()`

bugaevc:
So with this, we effectively unveil the whole parent directory, not just the one file we wanted to create?

bugaevc:
If what you actually wanted to do was not that, but still unveiling that specific file even if it doesn't yet exist, you need to construct the path for the (nonexistent) file being unveiled from the parent custody's absolute path and the basename.

Quaker762:
> If what you actually wanted to do was not that, but still unveiling that specific file even if it doesn't yet exist, you need to construct the path for the (nonexistent) file being unveiled from the parent custody's absolute path and the basename.

So for `auto new_unveiled_path = custody->absolute_path();` I can just tack on the filename from the base path that the user specified via `unveil` (which would be the string after the last occurrence of `/`?

bugaevc:
Basically, yes. But please separate the codepath for this from the rest of the logic. I'd expect the code to go something like this:

```cpp
String new_unveiled_path;
RefPtr<Custody> parent_custody;
auto custody_or_error = resolve_path(path, base, &parent_custody, options);
if (!custody_or_error.is_error()) {
    // All went fine, and the file already exists.
    new_unveiled_path = custody_or_error.value()->absolute_path();
} else if (custody_or_error.error() == -ENOENT && parent_custody && (new_permissions & UnveiledPath::Access::CreateOrRemove)) {
    // The file does not exist yet, but its parent directory exists, and
    // we're being asked to allow creating this file. Take note of its future name.
    String basename = LexicalPath(path).basename();
    new_unveiled_path = String::formatted("{}/{}", parent_custody->absolute_path(), basename);
} else {
    return custody_or_error.error();
}
```

Quaker762:
> ```c++
> String::formatted("{}/{}"
> ```

Thanks for the heads up on that function, I didn't know it existed.

Quaker762:
@bugaevc Let's see how we go with this :P

Kernel/Syscalls/unveil.cpp
```diff
@@ -90,6 +84,33 @@ int Process::sys$unveil(Userspace<const Syscall::SC_unveil_params*> user_params)
...
+    // the error from resolve_path_without_veil()
+    // However, if the user specified unveil() with "c" permissions, we don't set errno if ENOENT is encountered,
+    // because they most likely intend the program to create the file for them later on.
+    // If this case is encountered, the parent node of the path is returned and the custody of that inode is used instead.
+    RefPtr<Custody> parent_custody; // Parent inode in case of ENOENT
+    String new_unveiled_path;
+    auto custody_or_error = VFS::the().resolve_path_without_veil(path.value(), root_directory(), &parent_custody);
+    if (!custody_or_error.is_error()) {
+        new_unveiled_path = custody_or_error.value()->absolute_path();
+    } else if (custody_or_error.error() == -ENOENT && (new_permissions & UnveiledPath::Access::CreateOrRemove)) {
```
bugaevc, line 97: You forgot to check that parent custody is not null here. That matters — if you ask for `/foo/bar/baz` and it doesn't exist, you get `ENOENT` and `parent_custody = /foo/bar`, but if `/foo/bar` (the parent) doesn't exist, you get `ENOENT` and `parent_custody = nullptr`.

Kernel/Syscalls/unveil.cpp
```diff
@@ -90,6 +84,33 @@ int Process::sys$unveil(Userspace<const Syscall::SC_unveil_params*> user_params)
...
+    }
+
+    RefPtr<Custody> custody;
+    if (custody_or_error.is_error()) {
+        custody = parent_custody;
+        new_unveiled_path = parent_custody->absolute_path();
+    } else {
+        custody = custody_or_error.value();
+        new_unveiled_path = custody->absolute_path();
+    }
```
bugaevc, line 112: Wait, you don't need to do any of this. The only thing you need is to work out `new_unveiled_path`, and that is already done above.

Quaker762: Whoops, forgot to remove that.

Kernel/Syscalls/unveil.cpp
Quaker762: @bugaevc Okay this one should be resolved now. I can't think of any other situations we'd need to check for outside of these (mb on missing the parent custody). Let me know if I've missed anything obvious.

bugaevc: Why leak the ref???

Other than that, it looks fine :slightly_smiling_face:. Except I'd combine this case with the error case below, just as I have written originally.

Quaker762: > Why leak the ref???

Ah, my bad. this was to circumvent a compile error I was getting with `!= nullptr` and I forgot `.ptr()` existed haha.

bugaevc (approved):

Issue #3893, opened 2020-10-31
Boot Freezes on A Black Screen in VMware Workstation.


daemonspudguy:
I don't know what it's doing.
![image](https://user-images.githubusercontent.com/24438557/97791472-3191e500-1ba9-11eb-8c52-572d0d4b30bc.png)

Lubrsi:
IIRC from when I tried it, there's two issues:
- The scheduler gets stuck in the idle loop
- We need to write a new video driver for VMWare SVGA 2

tomuta:
>     * The scheduler gets stuck in the idle loop

This sounds like exactly what happens when trying to boot it in Bochs. At least in Bochs it seems like the HPET timer never fires.

daemonspudguy:
Still completely borked.

tomuta:
> Still completely borked.

Yeah because of other bugs. PS/2 has major problems (I'm working on that), and then sounds like we don't have an appropriate svga driver for vmware. You're welcome to contribute ;-)

Lubrsi:
I noticed it does try to use MBVGA on VMware, not sure if that'll be enough before making an SVGA driver.

Lubrsi:
Plus there's also an IDE bug that needs to be found and fixed.

linusg:
Probably related to #3747

PR #3968, opened 2020-11-06
Kernel: Add I8042Controller to detect and manage PS/2 devices

tomuta:
Inspired by the problems I ran into with the HPET changes that I hoped would maybe help with #3893...

Rework the PS/2 keyboard and mouse drivers to use a common 8042
controller driver. Also, reset and reconfigure the 8042 controller
as they are not guaranteed to be in the state that we expect.

- [x] Detect and initialize keyboard/mouse in qemu, bochs and vmware
- [x] Keyboard not working at all
- [x] Pressing keyboard key not triggering irq and breaks mouse (qemu, bochs, vmware)
- [x] Mouse not moving at all in bochs

### DIFF
--- a/Kernel/Devices/I8042Controller.cpp
+++ b/Kernel/Devices/I8042Controller.cpp
@@ -1,0 +1,272 @@
+/*
+ * Copyright (c) 2020, The SerenityOS developers.
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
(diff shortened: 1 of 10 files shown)


Kernel/Devices/I8042Controller.cpp
```diff
@@ -0,0 +1,268 @@
...
+#include <Kernel/Devices/KeyboardDevice.h>
+#include <Kernel/IO.h>
+
+namespace Kernel {
+
+static I8042Controller* s_the;
+
+void I8042Controller::initialize()
+{
+    // TODO: Check ACPI for presence
```
supercomputer7, line 37: Technically, we could use the IA-PC Hardware flags to see if i8042 controller is present.
The problem is that QEMU doesn't support those flags correctly, even if the revision of the FADT table is 3 (which was declared in ACPI 2.0) when using the Q35 chipset. However, it could be that other hypervisors do actually support those flags, so maybe we should still consider reading those and for now, if QEMU tells that there's no i8042 - print it to the log and continue on initialization.
The next thing to do is to check presence is in the AML bytecode, and that requires an interpreter.

tomuta: So why not just check the FADT for the flag? [Table 5.11 on page 156](https://uefi.org/sites/default/files/resources/ACPI_Spec_6_3_A_Oct_6_2020.pdf#table.5.11) Why would we look at AML?

tomuta: Looks like we're already looking at that flag, but as you said QEMU doesn't appear to support it properly...

https://github.com/SerenityOS/serenity/blob/c9c3667ea7037467a0474548d2fe99d03679c3f8/Kernel/ACPI/Parser.cpp#L107

tomuta: OK so the problem is that qemu supplies a FADT table with `sdt->h.revision == 1`, in which case that field is supposed to be assumed to be set. So that line really needs to look more like this:

```c++
m_x86_specific_flags.keyboard_8042 = (sdt->h.revision < 2) || (sdt->ia_pc_boot_arch_flags & (u8)FADTFlags::IA_PC_Flags::PS2_8042);
```

supercomputer7: QEMU probably relies on the fact that Windows, Linux and other OSes (e.g. BSDs, Solaris) have an AML interpreter, so they could enumerate the ACPI namespace by decoding the AML bytecode, so there's no need to support those flags even though it should be that way.

I'm working on implementing an AML interpreter but that will take some time to have some results, therefore the best thing to do is to check presence - if the hardware says we have i8042 controller - go on, otherwise just print an "error" to the kernel log and continue. Later when we have the AML interpreter working - we could enumerate the ACPI namespace and tell if there's i8042 or not, and if there isn't, we shouldn't try to initialize anything.

If the FADT revision is not new enough, we shouldn't assume that we have an i8042 controller either although it's an acceptable assumption (especially on old machines). Again, the best thing to do is to enumerate the ACPI namespace so we can know for sure.

tomuta: I don't recall where I read it, but it seems that the consensus is that it should be assumed. That being said, we could (eventually) harden it by implementing timeouts so that if the controller truly doesn't exist or respond then we could just skip using it.

tomuta:
Looks like the shell back grounding test failed again

PR #3974, opened 2020-11-07
Kernel: Implement an asynchronous device request stack

tomuta:
This allows issuing asynchronous requests for devices and waiting
on the completion of the request. The requests can cascade into
multiple sub-requests.

Since IRQs may complete at any time, if the current process is no
longer the same that started the process, we need to swich the
paging context before accessing user buffers.

Change the PATA driver to use this model.

- [x] Figure out why this crashes with `smp=off`:
```
[colonel(0:0)]: ASSERTION FAILED: handler.type() == HandlerType::SpuriousInterruptHandler

#0  0xc011f41b in halt_this () at ../Kernel/Arch/i386/CPU.cpp:957
#1  Kernel::Processor::halt () at ../Kernel/Arch/i386/CPU.cpp:1934
#2  0xc011eec8 in __assertion_failed () at ../Kernel/Arch/i386/CPU.cpp:2143
#3  0xc014e349 in Kernel::PIC::spurious_eoi ()
    at ../Kernel/Interrupts/PIC.cpp:95
#4  Kernel::PIC::spurious_eoi () at ../Kernel/Interrupts/PIC.cpp:93
#5  0xc014e5fd in Kernel::PIC::eoi () at ../Kernel/Interrupts/PIC.cpp:143
#6  0xc014d057 in Kernel::IRQHandler::eoi ()
    at ../Kernel/Interrupts/IRQHandler.cpp:53
#7  0xc011e023 in interrupt_common_asm_entry ()
```

### DIFF
--- a/Kernel/CMakeLists.txt
+++ b/Kernel/CMakeLists.txt
@@ -14,6 +14,7 @@ set(KERNEL_SOURCES
     CMOS.cpp
     CommandLine.cpp
     Console.cpp
+    Devices/AsyncDeviceRequest.cpp
     Devices/BXVGADevice.cpp
     Devices/BlockDevice.cpp
     Devices/CharacterDevice.cpp
--- a/Kernel/Devices/AsyncDeviceRequest.cpp
+++ b/Kernel/Devices/AsyncDeviceRequest.cpp
@@ -1,0 +1,175 @@
+/*
+ * Copyright (c) 2020, The SerenityOS developers.
+ * All rights reserved.
+ *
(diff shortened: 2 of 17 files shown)


tomuta:
That crash ended up happening because I called `enable_irq`/`disable_irq` from within the IRQ handler, which caused `PIC::m_cached_irq_mask` to be changed during interrupt handling, and then `PIC::eoi` thought it should call `PIC::spurious_eoi` when it really shouldn't.

PR #3977, opened 2020-11-07
LibWeb: Load favicon.ico only for http/https URLs

bcoles:


awesomekling:
Yup this makes sense :)

PR #3979, opened 2020-11-07
LibWeb: Restrict HTML form submissions to permitted URL protocols

bcoles:
Form submissions to `file://` URLs are now permitted only if the
submitting document is also a `file://` URL and the form method is "get".

Form submissions to URLs with a `http(s)://` URL protocol are permitted.

Form submissions for all other URL protocols are rejected.


awesomekling:
We should allow GET forms from `file://` to `file://`, no?

bcoles:
> 
> 
> We should allow GET forms from `file://` to `file://`, no?

Good call. I'll fix it tomorrow.

bcoles:
> 
> 
> > We should allow GET forms from `file://` to `file://`, no?
> 
> Good call. I'll fix it tomorrow.

Fixed

Issue #3964, opened 2020-11-06
FileManager: Icons block full file name


Lubrsi:
For example:
![image](https://user-images.githubusercontent.com/25595356/98327689-09144b80-1fec-11eb-991b-648ee311be6e.png)

AmusedNetwork:
What would be best here? 

1. Only show the file name up to where it would overlap the icon, or 
2. Display the label over the top of another icon? 
Perhaps with the latter some kind of scrolling text, but that doesn’t seem ideal.

It looks like Windows shows up until the next icon down and then covers it when editing.

awesomekling:
The Windows policy in the screenshot makes sense to me.

linusg:
This was fixed in #3981.

PR #3981, opened 2020-11-07
LibGUI: Limit the height of item text in IconView

AmusedNetwork:
Set the max height of the text_rect to be the height difference
between two icons. Calculate the number of text lines that can be 
displayed in this height and display only that many.

One line
image
Two lines
image
Max lines:
image

Todo:
- [x] Add elision for the last line of the icon text if required.

Related to #3964

### DIFF
--- a/Libraries/LibGUI/IconView.cpp
+++ b/Libraries/LibGUI/IconView.cpp
@@ -523,23 +523,32 @@ void IconView::paint_event(PaintEvent& event)
 
         auto font = font_for_index(item_data.index);
 
-        painter.fill_rect(item_data.text_rect, background_color);
+        Gfx::IntRect text_rect = item_data.text_rect;
+        auto icon_translation = translation.y() - 12;
+        text_rect.set_height(text_rect.height() > icon_translation ? icon_translation : text_rect.height());
+
+        painter.fill_rect(text_rect, background_color);
         if (is_focused() && item_data.index == cursor_index()) {
-            painter.draw_rect(item_data.text_rect, widget_background_color);
-            painter.draw_focus_rect(item_data.text_rect, palette().focus_outline());
+            painter.draw_rect(text_rect, widget_background_color);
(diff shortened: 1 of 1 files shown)


Libraries/LibGUI/IconView.cpp
```diff
@@ -523,24 +523,33 @@ void IconView::paint_event(PaintEvent& event)
 
         auto font = font_for_index(item_data.index);
 
-        painter.fill_rect(item_data.text_rect, background_color);
+        Gfx::IntRect text_rect = item_data.text_rect;
+        auto icon_translation = translation.y() - 12;
```
AmusedNetwork, line 527: I need to work out a better way of finding the y distance between two icons. This seems to work currently, but I don't know if it will break things down the line.

Libraries/LibGUI/IconView.cpp
```diff
@@ -523,24 +523,33 @@ void IconView::paint_event(PaintEvent& event)
...
                 Gfx::IntRect line_rect;
                 line_rect.set_width(item_data.text_rect.width());
                 line_rect.set_height(font->glyph_height());
                 line_rect.center_horizontally_within(item_data.text_rect);
                 line_rect.set_y(2 + item_data.text_rect.y() + line_index * font->glyph_height());
                 line_rect.inflate(6, 0);
 
+                // Shrink the line_rect on the last line to apply elision if there are more lines.
+                if (number_of_text_lines - 1 == line_index && lines.size() > number_of_text_lines)
+                    line_rect.inflate(-(6 + 2 * font->max_glyph_width()), 0);
```
AmusedNetwork, line 551: I was unsure of how else to cause the final line to have `...` if there were more lines after it. So I shrink the text box by 6 + two times the max character width. There must be a better way to do it, perhaps passing a new boolean, `should_elide`, into the draw_item_text function. Then it could be passed into Painter.cpp to force an elision even if the text isn't too long for the rect.

PR #3983, opened 2020-11-07
Teach ArgsParser about options taking doubles; tweak epoch time update; remove unused methods

nico:
A bit of a grab bag. This is all useful for something I'm hacking on every now and then, but it's arguably (minorly) useful in itself, so I'm trying to reduce my local diff a bit :)


awesomekling:
>my local diff

🤔

PR #3984, opened 2020-11-07
LibGUI+WindowServer: MimeData in DragOperation / Spreadsheet: Copying ranges of cells

### LINKED ISSUE #3906: Spreadsheet: Select, copy and paste cells

alimpfard:
Finally closes #3906.
This messes with WindowServer and drag&drop, but it should not have caused any changes to its existing behaviour.

In action (pardon the crappy gif, this is my first time making one like this :P)
out

---

cc @awesomekling for the LibGUI changes, I'm sure there's _something_.

- [x] Figure out why that `!s_current_drag_operation` assertion sometimes trips

### DIFF
--- a/Applications/Spreadsheet/main.cpp
+++ b/Applications/Spreadsheet/main.cpp
@@ -27,6 +27,7 @@
 #include "HelpWindow.h"
 #include "Spreadsheet.h"
 #include "SpreadsheetWidget.h"
+#include <AK/ScopeGuard.h>
 #include <LibCore/ArgsParser.h>
 #include <LibCore/File.h>
 #include <LibGUI/AboutDialog.h>
@@ -153,11 +154,26 @@ int main(int argc, char* argv[])
 
     auto& edit_menu = menubar->add_menu("Edit");
     edit_menu.add_action(GUI::CommonActions::make_copy_action([&](auto&) {
+        /// text/x-spreadsheet-data:
+        /// - currently selected cell
+        /// - selected cell+
         auto& cells = spreadsheet_widget.current_worksheet().selected_cells();
         ASSERT(!cells.is_empty());
         StringBuilder text_builder, url_builder;
         bool first = true;
+        auto cursor = spreadsheet_widget.current_selection_cursor();
+        if (cursor) {
+            Spreadsheet::Position position { spreadsheet_widget.current_worksheet().column(cursor->column()), (size_t)cursor->row() };
+            url_builder.append(position.to_url().to_string());
+            url_builder.append('\n');
+        }
+
         for (auto& cell : cells) {
+            if (first && !cursor) {
+                url_builder.append(cell.to_url().to_string());
+                url_builder.append('\n');
+            }
+
             url_builder.append(cell.to_url().to_string());
             url_builder.append('\n');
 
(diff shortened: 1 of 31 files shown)


Libraries/LibIPC/Encoder.cpp
```diff
@@ -141,6 +142,13 @@ Encoder& Encoder::operator<<(const String& value)
     return *this << value.view();
 }
 
+Encoder& Encoder::operator<<(const ByteBuffer& value)
+{
+    *this << static_cast<i32>(value.size());
+    m_buffer.append(value.data(), value.size());
+    return *this;
+}
```
alimpfard, line 150: I briefly thought about using a `SharedBuffer` for this here, but I decided not to involve another layer of abstraction.
I could probably do that if we don't want to pass potentially big buffers around (kinda weird as we support passing them as `StringView`s anyway)

Applications/Spreadsheet/main.cpp
```diff
@@ -175,41 +191,30 @@ int main(int argc, char* argv[])
     },
         window));
     edit_menu.add_action(GUI::CommonActions::make_paste_action([&](auto&) {
+        ScopeGuard update_after_paste { [&] { spreadsheet_widget.current_worksheet().update(); } };
+
         auto& cells = spreadsheet_widget.current_worksheet().selected_cells();
         ASSERT(!cells.is_empty());
         const auto& data = GUI::Clipboard::the().data_and_type();
         if (auto spreadsheet_data = data.metadata.get("text/x-spreadsheet-data"); spreadsheet_data.has_value()) {
```
alimpfard, line 199: Looking at how this is used elsewhere, I don't think `Clipboard::DataAndType` is meant to contain different types of data - just one and its metadata.
while this works just fine for my purposes, I'm almost tempted to switch that up to use `MimeData` too, should be alright since there's a raw bitmap encoding now, I think?

awesomekling: Yes, the `Clipboard::DataAndType` is vestigial now and should be replaced with `MimeData` :)

Libraries/LibGUI/AbstractView.cpp
```diff
@@ -288,6 +288,14 @@ void AbstractView::mousemove_event(MouseEvent& event)
 
     ASSERT(!data_type.is_null());
 
+    if (m_is_dragging)
+        return;
+
+    // An event might sneak in between us constructing the drag operation and the
+    // event loop exec at the end of `drag_operation->exec()' if the user is fast enough.
+    // Prevent this by just ignoring later drag initiations (until the current drag operation ends).
+    TemporaryChange dragging { m_is_dragging, true };
```
awesomekling, line 297: This feels like we're papering over something unsound.

awesomekling: I suppose it's not a new issue though, but it's probably something we should design away :)

alimpfard: It did feel wrong, but I couldn't think of any better ways to prevent it.
definitely would be nice to make it impossible by design.

PR #3985, opened 2020-11-07
Lagom/Fuzzers: Add a Dockerfile for FuzzilliJs

linusg:
Based on Fedora. This allows building and running FuzzilliJs and Fuzzilli itself in a Docker/Podman container.

### DIFF
--- a/Meta/Lagom/Fuzzers/FuzzilliJs.dockerfile
+++ b/Meta/Lagom/Fuzzers/FuzzilliJs.dockerfile
@@ -1,0 +1,71 @@
+# Build the image:
+# $ podman build \
+#       --tag fuzzillijs \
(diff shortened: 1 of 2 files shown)


awesomekling:
Hmm, I don't understand why this needs to be part of the project source tree? It looks like a personal docker script(?)

awesomekling:
I suppose if it's useful to anyone :)

linusg:
> It looks like a personal docker script(?)

Well, it was! But I figured it might be useful to others, I have no desire to install a whole swift environment on my host for example.

> I don't understand why this needs to be part of the project source tree?

It absolutely doesn't! Please don't hesitate to reject such PRs if you feel it's unnecessary. I'd just put it in a personal repo then, no problem :)

PR #3996, opened 2020-11-08
LibWeb: Fix OOPWV crash when resizing to zero width or height

linusg:


linusg:
too slow :(

awesomekling:
Oops, your fixes are even nicer though. Let’s merge them :)

PR #4008, opened 2020-11-09
TextEditor: Go-to-line now shows line in middle of view

jgbyrne:
As requested in Andreas' recent Breakout video, this PR makes Go-to-line jump more nicely such that the desired line is displayed 1/3rd of the way down the view rather than right at the extremes.


awesomekling:
Thank you for working on this!

I feel like we should do this in most/all situations where we set the cursor programmatically, not just when using the "go to line" action. Or what do you think? :)

jgbyrne:
I did consider this but looking through the TextEditor.cpp source file I didn't see any other situations where I felt it was an unqualified benefit to have this auto-scrolling applied.

This might just be my personal preference though - I'm somewhat averse to IDEs that feel overly clever :-)

I can easily factor out the logic though, perhaps into a method called something like `set_cursor_and_focus_line`, so that it can be easily dropped in elsewhere as desired?

awesomekling:
That works :)

PR #4012, opened 2020-11-09
AK: Make RefPtr, NonnullRefPtr, WeakPtr thread safe

tomuta:
This makes most operations thread safe, especially so that they
can safely be used in the Kernel. This includes obtaining a strong
reference from a weak reference, which now requires an explicit
call to WeakPtr::strong_ref(). Another major change is that
Weakable::make_weak_ref() may require the explicit target type.
Previously we used reinterpret_cast in WeakPtr, assuming that it
can be properly converted. But WeakPtr does not necessarily have
the knowledge to be able to do this. Instead, we now ask the class
itself to deliver a WeakPtr to the type that we want.

Also, WeakLink is no longer specific to a target type. The reason
for this is that we want to be able to safely convert e.g. WeakPtr<T>
to WeakPtr<U>, and before this we just reinterpret_cast the internal
WeakLink<T> to WeakLink<U>, which is a bold assumption that it would
actually produce the correct code. Instead, WeakLink now operates
on just a raw pointer and we only make those constructors/operators
available if we can verify that it can be safely cast.

In order to guarantee thread safety, we now use the least significant
bit in the pointer for locking purposes. This also means that only
properly aligned pointers can be used.

This will facilitate with #3864 and may also help with the crash in #3990

### DIFF
--- a/AK/Forward.h
+++ b/AK/Forward.h
@@ -113,6 +113,9 @@ template<typename T>
 class Optional;
 
 template<typename T>
+class RefPtrTraits;
+
+template<typename T, typename PtrTraits = RefPtrTraits<T>>
 class RefPtr;
 
 template<typename T>
--- a/AK/NonnullOwnPtr.h
+++ b/AK/NonnullOwnPtr.h
@@ -35,7 +35,7 @@
 
 namespace AK {
 
-template<typename T>
+template<typename T, typename PtrTraits>
 class RefPtr;
 template<typename T>
 class NonnullRefPtr;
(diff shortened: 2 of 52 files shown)


tomuta:
Hm not sure why the build fails like that. `make test` works fine for me, but it also doesn't run as many unit tests (only 33). Not sure how to run those...

tomuta:
Ugh, it's because I only ran `cmake ..`, not `cmake .. -DBUILD_LAGOM=1`. This is so unintuitive, we really need to add a section to https://github.com/SerenityOS/serenity/blob/master/Documentation/BuildInstructions.md or consider making it the default.

tomuta:
Hm they all pass for me...

awesomekling:
Yeah I think that was just the shell backgrounding test acting up again.

The need to specify a type for `make_weak_ptr()` is quite sad :/

awesomekling:
Can we use the existing `IsBaseOf` instead of `IsStaticCastable`?

tomuta:
> Yeah I think that was just the shell backgrounding test acting up again.
> 
> The need to specify a type for `make_weak_ptr()` is quite sad :/

I suppose it could be done with a free function. The problem really is when a base class derives from `Weakable<T>` and `make_weak_ptr` is called from a derived class. I don't know of a way to automatically "adjust" the return type in such a case.

> Can we use the existing `IsBaseOf` instead of `IsStaticCastable`?

Yeah I think that works just as well. Changed it.

tomuta:
Or maybe just a special constructor and `operator=` for `WeakPtr`...

tomuta:
> Or maybe just a special constructor and `operator=` for `WeakPtr`...

Hm I just tried that and it seems a lot cleaner. So if you have a `WeakPtr<T>` you just assign a `T&`, `T*`, or `RefPtr<T>` and it internally calls `make_weak_ptr<T>`.

awesomekling:
> > Or maybe just a special constructor and `operator=` for `WeakPtr`...
> 
> Hm I just tried that and it seems a lot cleaner. So if you have a `WeakPtr<T>` you just assign a `T&`, `T*`, or `RefPtr<T>` and it internally calls `make_weak_ptr<T>`.

That sounds really nice :)

tomuta:
Ok I made those changes, I think this quite slick.

PR #4019, opened 2020-11-10
Kernel/USB: Flesh out USB subsystem

Quaker762:
Going off of @awesomekling 's work on the basic UHCI controller, it'd be nice to flesh this out so that, at the very least, we can have basic USB device detection when a user plugs/unplugs a device from the root controller. I'm also going to add some nice abstractions to the different 

If anybody is interested in learning more about USB, I'm currently going through [this](http://www.fysnet.net/the_universal_serial_bus.htm) book, which goes into some depth about the different USB standards and how to implement sections of the USB stack.

- [x] Move USB structures into USB folder/namespace
- [x] Correctly configure UHCI controller
- [x] Detect hotplugged USB devices
- [x] Create different UHCI structures (Queue Headers, Transfer Descriptor Pool etc)
- [ ] Implement some kind of USB Notification Device @bugaevc 
- [x] Send packets to devices


danboid:
This got a big yay from me! You can't overstate the importance of having support for USB HID and mass storage devices at least.

Quaker762:
> This got a big yay from me! You can't overstate the importance of having support for USB HID and mass storage devices at least.

Yeah it'll be v srs bzns once we have USB going tbh. Long way off from having anything usable though.

Kernel/init.cpp
```diff
@@ -246,7 +246,8 @@ void init_stage2(void*)
         }
     }
 
-    UHCIController::detect();
+    USB::UHCIController::detect();
+    USB::UHCIController::the().spawn_port_proc();
```
Lubrsi, line 250: Careful here, if no UHCIController is found, this causes a null deref :)

Quaker762: Yeah this is going to get a _major_ cleanup soon. I've moved into the constructor for now for debug purposes.

Quaker762:
uhci_successful_transfer

;)

@awesomekling Going to get this cleaned up before we merge. There's still a LOT of work to be done before we have actual USB support, but so far we can effectively plug and play and perform a transaction to a device. The next layer up is the USB Driver itself (in which the actual device drivers would sit on top of). This is going to take a bit of designing. 

I'll open up a draft PR for that once I get this tidied up. The structs may need to be "classified". I am a bit hesitant to do that though, as we're going to be using a fair bit of Kernel Heap by allocating so many objects for the QH and TD pool (which require 32-bytes for the descriptor, then all of the data for the actual container object).

Quaker762:
Let's see how we go with this :P

Kernel/Devices/USB/UHCIDescriptorTypes.h
```diff
@@ -0,0 +1,285 @@
...
+private:
+    u32 m_link_ptr { 0 };                  // Pointer to the next horizontal object that the controller will execute after this one
+    volatile u32 m_element_link_ptr { 0 }; // Pointer to the first data object in the queue (can be modified by hw)
+
+    // These values will be ignored by the controller, but we can use them for configuration/bookkeeping
+    // Any addresses besides `paddr` are assumed virtual and can be dereferenced
+    u32 m_paddr { 0 };                          // Physical address where this QueueHead is located
+    QueueHead* m_next_qh { nullptr };           // Next QH
+    QueueHead* m_prev_qh { nullptr };           // Previous QH
+    TransferDescriptor* m_first_td { nullptr }; // Pointer to first TD
```
Quaker762, line 280: I feel like these could be `OwnPtr<QueueHead>` instead?

Kernel/Devices/USB/UHCIDescriptorTypes.h
```diff
@@ -0,0 +1,285 @@
...
+        BabbleDetected = (1 << 20),
+        DataBufferError = (1 << 21),
+        Stalled = (1 << 22),
+        Active = (1 << 23)
+    };
+
+    enum class ControlBits : u32 {
+        InterruptOnComplete = (1 << 24),
+        IsochronousSelect = (1 << 25),
+    };
```
Quaker762, line 69: Should I add all of these at once or fill them out as they get used?

Kernel/Devices/USB/UHCIController.cpp
```diff
@@ -0,0 +1,470 @@
...
+struct setup_packet {
+    u8 bmRequestType;
+    u8 bRequest;
+    u16 wValue;
+    u16 wIndex;
+    u16 wLength;
+};
+
+void UHCIController::do_debug_transfer()
+{
```
Quaker762, line 348: Stuff like this will be removed next PR. More of a baseline for other people to see how this is done if they want to join in on working on USB stuff, because the spec does a very poor job at actually explaining how these are performed at all.

Kernel/Devices/USB/UHCIController.cpp
```diff
@@ -0,0 +1,470 @@
...
+        // Place a new Transfer Descriptor with a 1:1 in our region
+        // The pointer returned by `new()` lines up exactly with the value
+        // that we store in `paddr`, meaning our member functions directly
+        // access the raw descriptor (that we later send to the controller)
+        m_iso_td_list.at(i) = new (placement_addr) Kernel::USB::TransferDescriptor(paddr);
+        auto transfer_descriptor = m_iso_td_list.at(i);
+        transfer_descriptor->set_in_use(true); // Isochronous transfers are ALWAYS marked as in use (in case we somehow get allocated one...)
+        transfer_descriptor->set_isochronous();
+        transfer_descriptor->link_queue_head(m_interrupt_transfer_queue->paddr());
+        transfer_descriptor->print();
```
Quaker762, line 200: These I'll remove after a proper review (forgot they were here)

Kernel/Devices/USB/UHCIController.cpp
```diff
@@ -0,0 +1,470 @@
...
+    m_iso_td_list.resize(UHCI_NUMBER_OF_ISOCHRONOUS_TDS);
+    for (size_t i = 0; i < m_iso_td_list.size(); i++) {
+        auto placement_addr = reinterpret_cast<void*>(m_td_pool->vaddr().get() + (i * sizeof(Kernel::USB::TransferDescriptor)));
+        auto paddr = static_cast<u32>(m_td_pool->physical_page(0)->paddr().get() + (i * sizeof(Kernel::USB::TransferDescriptor)));
+
+        // Place a new Transfer Descriptor with a 1:1 in our region
+        // The pointer returned by `new()` lines up exactly with the value
+        // that we store in `paddr`, meaning our member functions directly
+        // access the raw descriptor (that we later send to the controller)
+        m_iso_td_list.at(i) = new (placement_addr) Kernel::USB::TransferDescriptor(paddr);
```
Quaker762, line 195: I feel like this is a neat trick to map these structures virtual->physical, which makes life a lot easier.

awesomekling:
For commits like "Fix incorrect dangling TD fields", please squash them into the appropriate earlier commit instead of leaving an earlier commit broken. It's a lot easier to review if we don't have known bugs in earlier commits that get removed later on. :)

awesomekling:
This breaks CI (some linting issues)

awesomekling:
Same goes for "OO'ify all USB structures".
We don't need to see the individual steps you took to arrive at the final data structures. If you have a better structure in the end, please squash away the earlier one.

Kernel/Devices/USB/UHCIDescriptorTypes.h
```diff
@@ -0,0 +1,285 @@
...
+
+    TransferDescriptor() = delete;
+    TransferDescriptor(u32 paddr)
+        : m_paddr(paddr)
+    {
+    }
+    //~TransferDescriptor() = delete; // Prevent anything except placement new on this object
```
awesomekling, line 76: Please don't add commented-out code.

Kernel/Devices/USB/UHCIDescriptorTypes.h
```diff
@@ -0,0 +1,285 @@
...
+        Terminate = 1,
+        QHSelect = 2,
+    };
+
+    QueueHead() = delete;
+    QueueHead(u32 paddr)
+        : m_paddr(paddr)
+    {
+    }
+    //~QueueHead() = delete; // Prevent anything except placement new on this object
```
awesomekling, line 211: Please don't add commented-out code

Kernel/Devices/USB/UHCIDescriptorTypes.h
awesomekling: I don't see you calling `delete` on them anywhere, so would be the purpose of wrapping them in an owning pointer?

Kernel/Devices/USB/UHCIController.h
```diff
@@ -63,7 +74,28 @@ class UHCIController final : public PCI::Device {
...
+
+    Vector<QueueHead*> m_free_qh_pool;
+    Vector<TransferDescriptor*> m_free_td_pool;
+    Vector<TransferDescriptor*> m_iso_td_list;
+
+    OwnPtr<QueueHead> m_interrupt_transfer_queue;
+    OwnPtr<QueueHead> m_lowspeed_control_qh;
+    OwnPtr<QueueHead> m_fullspeed_control_qh;
+    OwnPtr<QueueHead> m_bulk_qh;
+    OwnPtr<QueueHead> m_dummy_qh; // Needed for PIIX4 hack
```
awesomekling, line 93: Is this actually sane? `OwnPtr` is meant for objects that need to be `delete`d. These are allocated via `allocate_queue_head()` and come from a non-heap pool. It doesn't seem safe to `delete` pointers into that pool(?)

Kernel/Devices/USB/UHCIDescriptorTypes.h
Quaker762: I think this comes more from the mentality of "don't use raw pointers!!!" even though in this situation they're justified.

Quaker762:
> This breaks CI (some linting issues)

Sorry was this supposed to be in reference to a line of code???

Kernel/Devices/USB/UHCIController.h
Quaker762: Yeah these should probably be raw pointers too. I suppose unless explicitly performed otherwise, lifetime of these are guaranteed.

Dexesttp:
> 
> 
> > This breaks CI (some linting issues)
> 
> Sorry was this supposed to be in reference to a line of code???

Just for your information, here are the two linting errors that your PR triggers :
```
Running Meta/check-debug-flags.sh... 
ALL_THE_DEBUG_MACROS probably doesn't include UHCI_DEBUG
Some flags are missing for the ALL_THE_DEBUG_MACROS feature in CMakeLists.txt.
If you just added a new SOMETHING_DEBUG flag, that's great!
We want to enable all of these in automated builds, so that the code doesn't rot.
Please add it to the ALL_THE_DEBUG_MACROS section.
[FAIL]: Meta/check-debug-flags.sh
```

and :

```
Running Meta/check-style.sh... 
"#pragma once" should have a blank line before and after in these files: Kernel/Devices/USB/UHCIDescriptorTypes.h
[FAIL]: Meta/check-style.sh
```

Quaker762:
@Dexesttp Cheers mate I'll get them fixed up

awesomekling:
All right, let's get this merged and continue iterating in tree :)

PR #4023, opened 2020-11-10
LaunchServer: Add hsp=/bin/HackStudio file association to config

bcoles:
I've never seen `.hackstudio` used as a file extension for HackStudio projects. Let's replace the default file association in anon's LaunchServer config with `.hsp` which is currently used by the `little` example project.

**Edit:** Prior to this change, opening `.hsp` files in FileManager works, but `.hsp` files cannot be opened by clicking the file in Terminal. FileManager has its own hard-coded file associations outside of LaunchServer. (Perhaps this should be changed?)

https://www.youtube.com/watch?v=rRGu8DVrlHQ#t=44m15s

### DIFF
--- a/Base/home/anon/.config/LaunchServer.ini
+++ b/Base/home/anon/.config/LaunchServer.ini
@@ -11,7 +11,7 @@ html=/bin/Browser
 wav=/bin/SoundPlayer
 txt=/bin/TextEditor
 font=/bin/FontEditor
-hackstudio=/bin/HackStudio
+hsp=/bin/HackStudio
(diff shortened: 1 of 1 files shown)


linusg:
> I've never seen .hackstudio used as a file extension for HackStudio projects

https://github.com/SerenityOS/serenity/pull/3014/commits/2860d7b96c2994c6b8e896e6a25e4ad8434a2a7d

That was the change from the OG ".files" Hackstudio file extension to ".hackstudio" but then someone decided to rename it again (and forgot about LaunchServer, apparently) :(

PR #4029, opened 2020-11-10
Games: Add Pong

bcoles:
[Happy 48th birthday Pong!](https://en.wikipedia.org/wiki/Pong)

Pong!

Application boilerplate / setup code and physics code stolen from kling's `Breakout`.

It turns out that the screen size and object dimensions for classic Atari pong is [complicated](https://atariage.com/forums/topic/169128-what-is-the-atari-2600-screen-resolution/).

I opted to use a game size which will fit within a 800x600 screen resolution as the default. Eventually, this can be customised via config file.

[This image](https://www.arcade-history.com/images/game/2007_1.png) was used as a guide for size:

pong example

There are a few bugs:

* The final screen paint doesn't occur upon game over. The final score shown is `20` upon winning with a score of `21`.
* The physics are based on the `Breakout` game which has some quirks.
* Player 2 is the on the left hand side. Player 1 is on the right. Disputable whether this is a bug. Other clones also seem to set the AI on the left and the player or the right, although these clones don't use "player 1" / "player 2" terminology. Eventually we could add support for a second human player, instead of AI, so this terminology makes sense.
* There's no option to start a new game upon game over.
* There's no pause functionality. Traditionally, Pong did not support pausing. However, we may want to offer configuration options (change paddle color, game size, etc), and it would be silly to have the game keep playing while it is not in focus or while trying to navigate menus.
* The AI player plays like it is having a seizure. Not a bug. I assure you that the AI is fine and having fun.
* ... probably others


awesomekling:
*Vertical* pong!? Hmm... can we make it horizontal? :smile:

bcoles:
> 
> 
> _Vertical_ pong!? Hmm... can we make it horizontal? smile

Now that you mention it ...

I'll switch it to horizontal and add support for switching between two player / AI opponent.

tomuta:
> _Vertical_ pong!? Hmm... can we make it horizontal? smile

Both?

bcoles:
I've updated Pong to be horizontal as it should be.

> 
> 
> > _Vertical_ pong!? Hmm... can we make it horizontal? smile
> 
> Both?

Maybe later. Unfortunately the orientation of the game play area effects the game physics as we're using `x` and `y` coordinates for positions, velocity, etc. So any change to orientation would require associated physics changes. Not impossible, but not trivial either.

PR #4047, opened 2020-11-11
RefPtrTraits: struct/class mismatch in forward declaration

ldm5180:
Problem:
- Building with clang is broken because of the `struct` vs `class`
  mismatch between the definition and declaration.

Solution:
- Change `class` to `struct` in the forward declaration.


tomuta:
oops...

PR #4057, opened 2020-11-11
CMake: compile_commands.json output

ldm5180:
Problem:
- CMake is not outputting `compile_commands.json`.
- `compile_commands.json` is used by build integration tooling such as
  `clang-tidy`.

Solution:
- Enable `CMAKE_EXPORT_COMPILE_COMMANDS` option so that the file is
  output.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -12,6 +12,8 @@ set(CMAKE_INSTALL_MESSAGE NEVER)
 
 enable_testing()
 
+set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
+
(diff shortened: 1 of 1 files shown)


linusg:
hmm... I don't mind enabling this by default (I use it for clangd), but you can simply invoke CMake with `-DCMAKE_EXPORT_COMPILE_COMMANDS=ON` to solve this "problem".

ldm5180:
Right. I just though that since things like `clangd` use it there is a benefit to just having it enabled all the time. I could really go either way on this.

awesomekling:
Seems harmless :)

PR #4059, opened 2020-11-12
InputFileStream: Incorrectly defaulted constructor

ldm5180:
Problem:
- The default constructor is is deleted because NonnullRefPtr has no
  default constructor and there is a member variable of that type, but
  the function is set as `= default`.

Solution:
- Remove the code because the function is actually deleted implicitly.

### DIFF
--- a/Libraries/LibCore/FileStream.h
+++ b/Libraries/LibCore/FileStream.h
@@ -93,8 +93,6 @@ public:
     }
 
 private:
-    InputFileStream() = default;
-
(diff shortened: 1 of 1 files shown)


asynts:
That whole `InputFileStream` / `OutputFileStream` thing is a mess. I added it in the hope of making all stream operations implement a common interface, but in reality this proved difficult. I'll eventually have to come around and clean that up, but currently I have no idea how to proceed with that, so I'll just leave it as it is now.

PR #4068, opened 2020-11-13
TextEditor: Change cursor behaviour when clicking empty space

### LINKED ISSUE #4048: TextEditor: Strange cursor behaviour
Open TextEditor, enter some text then click within the TextEditor window on any line below the current one and the cursor gets moved to the start of the current line.

supex0fan:
When clicking empty space (beneath any used lines) in the TextEditor, the cursor would jump to the start of the last line, rather than the correct column, or the end of the line where appropriate. This was because the for_each_visual_line callback would return IterationDecision::Continue if the clicked point wasn't in the line's rect. Thus the callback would continue on each iteration and at the end, would set the cursor to the default column of 0. To fix this I added a bool to the callback's signature which tells the callback if the for_each_visual_line method is on the last visual line. The callback now does not return IterationDecision:Continue if for_each_visual_line method is on the last line and the correct column is then calculated with the line passed.

Resolves #4048

### DIFF
--- a/Libraries/LibGUI/TextEditor.cpp
+++ b/Libraries/LibGUI/TextEditor.cpp
@@ -170,8 +170,8 @@ TextPosition TextEditor::text_position_at(const Gfx::IntPoint& a_position) const
     size_t column_index = 0;
     switch (m_text_alignment) {
     case Gfx::TextAlignment::CenterLeft:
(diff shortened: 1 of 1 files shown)


awesomekling:
Please observe the contribution guidelines and keep commit messages at 72 characters or less per line :)

awesomekling (changes requested):
Commit message needs cleanup as mentioned.

PR #4069, opened 2020-11-13
LibWeb: Add initial implementation of document.implementation

Lubrsi:


awesomekling:
This leaves many code paths that create a `DOM::Document` without setting a content type.
Perhaps it would be better to make `DOM::Document::create()` take the type in an argument. That way we can't forget to set it.

Lubrsi:
I just checked the spec and I didn't realise that content type has a default value:
```
Unless stated otherwise, a document’s encoding is the utf-8 encoding, content type is "application/xml", etc.
```
Which would be better?

PR #4070, opened 2020-11-13
LibWeb: Use standardized encoding names, add encoding attribute to document

Lubrsi:


Libraries/LibWeb/Loader/FrameLoader.cpp
```diff
@@ -138,6 +139,7 @@ RefPtr<DOM::Document> FrameLoader::create_document_from_mime_type(const ByteBuff
         HTML::HTMLDocumentParser parser(data, encoding);
         parser.run(url);
         document = parser.document();
+        document->set_encoding(encoding);
```
awesomekling, line 142: Doesn't the `HTMLDocumentParser` already create the document with the correct encoding above?

Lubrsi: It only uses it to set the encoding for the tokenizer. I could move that line into ```HTMLDocumentParser``` though.

awesomekling: Yeah that'd be nice

PR #4073, opened 2020-11-13
ClipboardHistoryModel: Prevent duplicate rows

### LINKED ISSUE #4033: Clipboard: double clicking an entry duplicates it
The clipboard app seems pretty unusable at the moment because when you double click on an item in its window it duplicates it instead of making it the active clipboard buffer. You should only need to single-click on an entry to make it the active. Double clicking an item in the clipboard window should do nothing IMO.

supex0fan:
Prevents the adding of items to the ClipboardHistoryModel if the raw data and mime_type of the item being added is the same as another item already in the list.

Resolves #4033

### DIFF
--- a/MenuApplets/ClipboardHistory/ClipboardHistoryModel.cpp
+++ b/MenuApplets/ClipboardHistory/ClipboardHistoryModel.cpp
@@ -89,13 +89,9 @@ void ClipboardHistoryModel::update()
 
 void ClipboardHistoryModel::add_item(const GUI::Clipboard::DataAndType& item)
 {
-    for (size_t i = 0; i < m_history_items.size(); i++) {
-        auto& existing = m_history_items.at(i);
-        if (existing.data == item.data && existing.mime_type == item.mime_type) {
-            m_history_items.remove(i);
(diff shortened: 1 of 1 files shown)


MenuApplets/ClipboardHistory/ClipboardHistoryModel.cpp
```diff
@@ -89,8 +89,17 @@ void ClipboardHistoryModel::update()
...
 void ClipboardHistoryModel::add_item(const GUI::Clipboard::DataAndType& item)
 {
+    for (size_t i = 0; i < m_history_items.size(); i++) {
+        auto& existing = m_history_items.at(i);
+        if (existing.data == item.data && existing.mime_type == item.mime_type) {
+            m_history_items.remove(i);
+            break;
+        }
+    }
+
```
awesomekling, line 99: Please use `Vector::remove_first_matching()` here.

awesomekling (changes requested):
Please observe the contribution guidelines and keep your commit messages at 72 characters per line or less.

PR #4077, opened 2020-11-14
Userland: Add test-gfx-font for Gfx::Font tests

bcoles:
I had already written most of this while messing around with fonts. Figured I may as well clean it up and submit. Perhaps it will be useful for copypasta monkeys or as example code. It can't hurt to have more tests?

test-gfx-font


awesomekling:
Great initiative adding tests for this!
It's a bit sad that we have so many different tests that implement their own ad hoc "testing framework".
Would be something nice to unify going forward.

bcoles:
> 
> 
> Great initiative adding tests for this!
> It's a bit sad that we have so many different tests that implement their own ad hoc "testing framework".
> Would be something nice to unify going forward.

Created #4078 for discussion.

Issue #4072, opened 2020-11-13
False Toolchain cache hit when Toolchain is patched


BenWiederhake:
(This is kind of a note to myself, when I have time to deal with it.)

While scrolling through the Dynamic Linking PR #3738, I noticed that `Toolchain/Patches/gcc.patch` changed, but not `BuildIt.sh` or any of the "obvious" files.

This will break Github Actions, as it currently ignores changes to the `Toolchain/Patches/` directory. Obviously, this needs to be fixed.

So, to be overly clear:

Steps to reproduce: Change a file in `Toolchain/Patches/`, but not `Toolchain/BuildIt.sh` or any of the header files.
Expected behavior: Our Github Actions workflow rebuilds the Toolchain, and does *not* attempt to reuse the old toolchain.
Actual behavior: Because that directory is not in the `${{ hashFiles(…) }}` list, the has does not change, therefore it reuses the old toolchain, causing weird breakages down the line.

PR #4082, opened 2020-11-14
IRC bridge without joins/parts, and fix toolchain caching

BenWiederhake:
thakis runs the bridge, so if it goes down blame him instead of me :P

See #4072 for the toolchain topic.


awesomekling:
All right, let's see how long this works ;)

PR #4085, opened 2020-11-14
LibGUI/FileIconProvider: Add config file

linusg:
Also add filetype-json icons.


Base/etc/FileIconProvider.ini
```diff
@@ -16,3 +16,22 @@ python=*.py
 sound=*.wav
 spreadsheet=*.sheets
 text=*.txt
+
+[Overrides]
```
awesomekling, line 21: This feels very hackish. How about we just rename the files to `*.json` instead? :D

linusg: Indeed, it is quite hackish... but can we really do that without breaking stuff, same for the "text" files in `/proc`? `/proc/uptime.txt`? :thinking:

This is just a nice-to-have ofc, we can also leave it at the generic file icon.

awesomekling: Sure, it'll break some hard-coded paths here & there, but we can fix those :)
Everything the light touches is our kingdom!

alimpfard: All I see here is `br /proc/all.html`

PR #4091, opened 2020-11-15
UserspaceEmulator: Implement FISUBR_RM32

nico:
Fixes crash when playing `ue Breakout` with the mouse.


DevTools/UserspaceEmulator/SoftCPU.cpp
```diff
@@ -1672,11 +1672,17 @@ void SoftCPU::FISUB_RM32(const X86::Instruction& insn)
     ASSERT(!insn.modrm().is_register());
     auto m32int = (i32)insn.modrm().read32(*this, insn).value();
     // FIXME: Respect shadow values
-    auto f64 = (long double)m32int;
```
nico, line 1675: (removing the var here because a long double doesn't have 64 bits)

awesomekling: Oh right!

PR #4096, opened 2020-11-16
ClipboardHistory: Prevent crash when clicking empty row

supex0fan:
Added an index.is_valid() check to the on_activation method which
prevents vector.at(-1) from being called and triggering an assertion
failure.


awesomekling:
This is a bug in LibGUI. Views should not activate an invalid index, and we shouldn't work around it here :)
I have a fix for this that I will push, so I'm closing this.

PR #4100, opened 2020-11-16
LibGfx: Use gamma-corrected interpolation for color gradients

### LINKED ISSUE #1342: LibGfx: Gradients shouldn't be interpolated in sRGB
The sRGB color space that LibGfx uses to represent color values is a non-linear color space, interpolating between these gamma-compressed colors yields incorrect results. The effect of this is noticable around the center of LibGfx gradients as a large drop in color brightness.

It would be better to linearize the color space before doing any color transformations for the gradients.

ccapitalK:
Switch over to gamma-aware interpolation. This causes color gradients
to not look so dark in the middle.

Fixes #1342.

### DIFF
--- a/Libraries/LibGfx/Color.h
+++ b/Libraries/LibGfx/Color.h
@@ -89,10 +89,10 @@ public:
     static constexpr Color from_rgb(unsigned rgb) { return Color(rgb | 0xff000000); }
     static constexpr Color from_rgba(unsigned rgba) { return Color(rgba); }
 
-    u8 red() const { return (m_value >> 16) & 0xff; }
-    u8 green() const { return (m_value >> 8) & 0xff; }
-    u8 blue() const { return m_value & 0xff; }
-    u8 alpha() const { return (m_value >> 24) & 0xff; }
+    constexpr u8 red() const { return (m_value >> 16) & 0xff; }
+    constexpr u8 green() const { return (m_value >> 8) & 0xff; }
+    constexpr u8 blue() const { return m_value & 0xff; }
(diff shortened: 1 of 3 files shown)


ccapitalK:
As a simple test, I changed the title bar theme to be a simple rgb(1,0,0) to rgb(0,0,1) gradient, then took screenshots of the results with no gamma correction (the current state before this patch), accurate gamma correction (using slow pow() calls), and approximate gamma correction. I've uploaded them below.

From an aesthetic perspective, I'm kind of on the fence about whether we should have gamma corrected gradients. I kinda liked the gray effect that the current method has, it looks a bit more like windows 95. Thoughts?

ccapitalK:
gradient_linear

gradient_gamma_accurate

gradient_gamma_fast

ccapitalK:
I've also done some investigation into how close the approximate transformations are to the proper ones, here are some graphs:

ccapitalK:
I've done some preliminary benchmarking on how fast my machine can draw a 800x600 pixel gradient to a bitmap in memory. With the old code, my pc could pull off about 514 paints/second, with the approximate gamma transformation that becomes about 120 paints/second, and with ACCURATE_GAMMA_ADJUSTMENT enabled that becomes 0.25 paints/second.

ccapitalK:
I've updated the PR with an sse optimized implementation, which is enabled when built with `-msse`. This new implementation averages about 700 paints/second on the previous benchmark, which I'm going to classify as a 40% speed improvement.

ccapitalK:
I think this is ready for merge, unless you guys have a problem with it?

awesomekling:
Very nice work! :)

PR #4102, opened 2020-11-16
AK: Fix OOB access in DuplexMemoryStream::offset_of()

alimpfard:
This fixes an OOB access when the last read/written chunk is empty (as we _just_
started on a new chunk).
Found via human fuzzing in the shell:
```sh
for $(cat /dev/urandom) {
    clear
    match $it {
        ?* as (x) {
            echo $x
            sleep 1
        }
    }
}
```
would assert at some point.

### DIFF
--- a/AK/MemoryStream.h
+++ b/AK/MemoryStream.h
@@ -221,22 +221,26 @@ public:
         return true;
     }
 
+    // FIXME: Does not read across chunk boundaries
+    //        Perhaps implement AK::memmem() for iterators?
     Optional<size_t> offset_of(ReadonlyBytes value) const
     {
         if (value.size() > size())
             return {};
 
         // First, find which chunk we're in.
-        auto chunk_index = (m_read_offset - m_base_offset) / chunk_size;
+        auto chunk_index = min((m_read_offset - m_base_offset) / chunk_size, m_chunks.size() - 1);
         auto last_written_chunk_index = (m_write_offset - m_base_offset) / chunk_size;
         auto first_chunk_index = chunk_index;
         auto last_written_chunk_offset = m_write_offset % chunk_size;
         auto first_chunk_offset = m_read_offset % chunk_size;
         size_t last_chunk_offset = 0;
         auto found_value = false;
+        auto chunk_index_max_bound = last_written_chunk_offset > 0 ? last_written_chunk_index + 1 : last_written_chunk_index;
 
-        for (; chunk_index <= last_written_chunk_index; ++chunk_index) {
-            auto chunk_bytes = m_chunks[chunk_index].bytes();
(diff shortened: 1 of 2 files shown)


asynts:
That `offset_of` method is problematic in many ways. I've had it on my to-do list for a while to deal with it. It doesn't read across chunk boundaries and therefor isn't particularly reliable. Additionally, the logic is very overloaded and difficult to follow.

A solution could be to implement `memmem` with iterators but that seems though and probably won't be fast either. Another thing that I was thinking about was to remove the whole chunk stuff and just use a `Vector`, but that would mean storing the whole thing contiguously, which again could be very slow.

This fix will probably do for now, although, I would add a test case:

~~~diff
diff --git a/AK/Tests/TestMemoryStream.cpp b/AK/Tests/TestMemoryStream.cpp
index 789e9c488..766c1690c 100644
--- a/AK/Tests/TestMemoryStream.cpp
+++ b/AK/Tests/TestMemoryStream.cpp
@@ -194,4 +194,18 @@ TEST_CASE(new_output_memory_stream)
     EXPECT_EQ(stream.bytes().size(), 2u);
 }
 
+TEST_CASE(offset_of_out_of_bounds)
+{
+    Array<u8, 4> target { 0xff, 0xff, 0xff, 0xff };
+
+    Array<u8, DuplexMemoryStream::chunk_size> whole_chunk;
+    whole_chunk.span().fill(0);
+
+    DuplexMemoryStream stream;
+
+    stream << whole_chunk;
+
+    EXPECT(!stream.offset_of(target).has_value());
+}
+
 TEST_MAIN(MemoryStream)
~~~

This test case fails without the new commit and now passes. Maybe add a comment too:

~~~diff
diff --git a/AK/MemoryStream.h b/AK/MemoryStream.h
index 34af34ba0..484f78590 100644
--- a/AK/MemoryStream.h
+++ b/AK/MemoryStream.h
@@ -221,6 +221,7 @@ public:
         return true;
     }
 
+    // FIXME: Does not read across chunk boundaries. Maybe change AK::memmem to accept iterators?
     Optional<size_t> offset_of(ReadonlyBytes value) const
     {
         if (value.size() > size())
~~~

alimpfard:
I'll add the test case and the FIXME.

As for the many issues surrounding `offset_of()`...
I think the "fast" path in `memmem()` would be just fine with taking an iterator, but the naive loop will take a significant hit - might be interesting to think about, how common is `memmem()` with `needle_length >= 32` anyway?

asynts:
*(Sigh... GitHub... Ignore the previous "mentioned" message, I just just the issue tracker in my fork as to-do list.)*

asynts (approved):

PR #2363, opened 2020-05-24
Regex v2

lnzero1dev:
To be further discussed in #2352

Issue #2352, opened 2020-05-23
LibRegex: A regex engine for Serenity


nooga:
While browsing the code, I've noticed that we don't have `grep` yet and also that `RegExp` would be a nice addition to the "maturing" LibJS runtime. In short, Serenity could use a regex engine of its own and I would like to dedicate this issue to figuring out how to make it happen.

In my mind, the engine should ideally have the following features and properties:

- [ ] Accept POSIX regex syntaxes (BRE and ERE),
- [ ] Accept JS regex syntax (eventually), 
- [ ] Play nice with Unicode,
- [ ] Have a decent performance both for short strings and long streams,
- [ ] Be safe and easy to use in any Serenity application.

Given all this, my approach would be to structure the engine as follows:
```
Parser -[NFA]-> Compiler -[bytecode]-> VM
```
where:
1. Parser's job would be to construct a Nondeterministic Finite Automaton from the regex string,
2. Compiler's job would be to construct (and possibly minimize) a Deterministic Finite Automaton based on the NFA and generate a program for the VM,
3. VM's job would be to perform matching based on the supplied program.

This approach would allow us to:
- write parsers for different regex syntaxes independently,
- save on building ASTs for regexps,
- offload the heavy lifting to the generic compiler, 
- cache compiled regexps for rapid reuse, possibly even across threads.

Using the VM for matching would offer decent performance while not requiring executable memory which would be a deal breaker for apps like the Browser.

I'm not sure how the API would look like yet but it should probably provide common regexp related functionality such as:
- [ ] `compile` and `execute` (+ convenience function to do that in one step),
- [ ] `search` & `test`
- [ ] `match`
- [ ] `replace`

I'm curious about your ideas and input. I'd like to establish some kind of loose consensus over this idea before I start laying down the code :^)

linusg:
Related (previous attempt of adding regex capabilities to Serenity): https://github.com/SerenityOS/serenity/pull/1654

nooga:
@linusg missed that one 😅

devsh0:
Simulating a DFA would make back referencing extremely tricky. The engine will have to deal with a lot more states and the best known solution to minimize them is NP complete.

linusg:
> I believe LibJS needs to support back referencing @linusg?

Yes:

[...]

PR #4103, opened 2020-11-16
LibRegex: The continuation

alimpfard:
Continuing on from #2352 and #2363, this PR aims to:
- [x] Make it build and pass CI
- [x] Add 'match' to `expr`
- [x] Add (basic) support for ECMAScript262 patterns - things missing:
    - ~Lookarounds~ (a bit leaky, but we can maybe fix that later)
    - Unicode properties
    - ~negated character class escapes inside character classes (e.g. `[\W]`)~
- [x] Figure out the whole LibC/LibRegex dependency cycle
    > < kling> CxByte: we can live with -lregex for now and figure something out later?

We might decide not to have all of ECMAScript262's regex stuff happen in this PR, but I'd like to at least get its parser working, even if some paths are `TODO()`'d.


Libraries/LibRegex/RegexError.h
```diff
@@ -28,27 +28,32 @@
...
+    NoError = __Regex_NoError,
+    InvalidPattern = __Regex_InvalidPattern,                         // Invalid regular expression.
+    InvalidCollationElement = __Regex_InvalidCollationElement,       // Invalid collating element referenced.
+    InvalidCharacterClass = __Regex_InvalidCharacterClass,           // Invalid character class type referenced.
+    InvalidTrailingEscape = __Regex_InvalidTrailingEscape,           // Trailing \ in pattern.
+    InvalidNumber = __Regex_InvalidNumber,                           // Number in \digit invalid or in error.
+    MismatchingBracket = __Regex_MismatchingBracket,                 // [ ] imbalance.
+    MismatchingParen = __Regex_MismatchingParen,                     // ( ) imbalance.
+    MismatchingBrace = __Regex_MismatchingBrace,                     // { } imbalance.
+    InvalidBraceContent = __Regex_InvalidBraceContent,               // Content of {} invalid: not a number = __Regex_number, number too large = __Regex_large, more than two numbers = __Regex_numbers, first larger than second.
```
alimpfard, line 49: LOL vim `:s/.../.../g` makes things weird again

Libraries/LibJS/Runtime/RegExpObject.cpp
```diff
@@ -24,13 +24,66 @@
...
 #include <LibJS/Heap/Heap.h>
 #include <LibJS/Runtime/GlobalObject.h>
 #include <LibJS/Runtime/PrimitiveString.h>
 #include <LibJS/Runtime/RegExpObject.h>
 #include <LibJS/Runtime/Value.h>
 
+static regex::RegexOptions<ECMAScriptFlags> options_from(const String& flags, JS::VM& vm, JS::GlobalObject& global_object)
+{
+    bool g = false, i = false, m = false, s = false, u = false, y = false;
+    regex::RegexOptions<ECMAScriptFlags> options { (regex::ECMAScriptFlags)regex::AllFlags::Global }; // JS regexps are all 'global' as per our definition, but the "global" flag enables "stateful".
```
alimpfard, line 38: This is actually pretty weird (as expected of javascript!), all expressions are global by default (as in, they are not anchored), but the 'global' flag makes the instance 'stateful' in that it continues from whatever offset it last left off (if it was successful) and zero otherwise.

Future work (for `exec()` and co) should keep this fact in mind, also the fact that `Regex::match()` will find _all_ matches if `Global` is set; we might want to introduce another internal flag for this, something like `SingleMatch`.

Libraries/LibJS/Tests/builtins/RegExp/RegExp.prototype.test.js
```diff
@@ -0,0 +1,51 @@
...
+
+
+    expect(() => {
+        /foo/gg
+    }).toThrowWithMessage(SyntaxError, "Repeated RegExp flag 'g'");
+
+    expect(() => {
+        /foo/x
+    }).toThrowWithMessage(SyntaxError, "Invalid RegExp flag 'x'");
+});
```
alimpfard, line 51: cc @linusg, please review (at least the libjs parts) of this commit : )

Libraries/LibJS/Tests/builtins/RegExp/RegExp.prototype.test.js
```diff
@@ -0,0 +1,51 @@
+test("basic functionality", () => {
```
linusg, line 1: Format file with prettier please :)
Will only affect quotes, I think

Libraries/LibRegex/RegexMatcher.cpp
```diff
@@ -126,7 +139,9 @@ RegexResult Matcher<Parser>::match(const Vector<RegexStringView> views, Optional
     s_regex_dbg.print_header();
 #endif
 
-    bool continue_search = (input.regex_options & AllFlags::Global) || (input.regex_options & AllFlags::Multiline); // || (input.regex_options & AllFlags::MatchNotBeginOfLine) || (input.regex_options & AllFlags::MatchNotEndOfLine);
+    bool continue_search = input.regex_options.has_flag_set(AllFlags::Global) || input.regex_options.has_flag_set(AllFlags::Multiline); // || (input.regex_options & AllFlags::MatchNotBeginOfLine) || (input.regex_options & AllFlags::MatchNotEndOfLine);
```
linusg, line 142: commented code, mr kling will complain lol

Libraries/LibRegex/RegexMatcher.cpp
```diff
@@ -0,0 +1,365 @@
...
+#ifdef REGEX_DEBUG
+        dbg() << "[match] Starting match with view (" << view.length() << "): _" << view.to_string() << "_";
+#endif
+
+        auto view_length = view.length();
+        size_t view_index = m_pattern.start_offset;
+        for (; view_index < view_length; ++view_index) {
+            auto& match_length_minimum = m_pattern.parser_result.match_length_minimum;
+            // FIXME: More performant would be to know the remaining minimum string
+            //        lenght needed to match from the current position onwards within
```
linusg, line 157: s/lenght/length/

Libraries/LibJS/Runtime/RegExpPrototype.cpp
```diff
@@ -43,6 +43,19 @@ void RegExpPrototype::initialize(GlobalObject& global_object)
...
+    define_native_property(vm.names.dotAll, dotAll, nullptr, readable_attr);
+    define_native_property(vm.names.flags, flags, nullptr, readable_attr);
+    define_native_property(vm.names.global, global, nullptr, readable_attr);
+    define_native_property(vm.names.ignoreCase, ignore_case, nullptr, readable_attr);
+    define_native_property(vm.names.multiline, multiline, nullptr, readable_attr);
+    define_native_property(vm.names.source, source, nullptr, readable_attr);
+    define_native_property(vm.names.sticky, sticky, nullptr, readable_attr);
+    define_native_property(vm.names.unicode, unicode, nullptr, readable_attr);
+
+    define_native_property(vm.names.lastIndex, last_index, set_last_index, Attribute::Writable);
```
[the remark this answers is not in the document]
alimpfard: So...where should this be defined? In words that I could understand?
I tried `RegExp.prototype.lastIndex` in Firefox and it told me `0`, so I thought this is the right place - I guess not

Libraries/LibJS/Runtime/RegExpObject.h
```diff
@@ -42,12 +43,15 @@ class RegExpObject : public Object {
...
     const String& flags() const { return m_flags; }
+    const Regex<ECMAScript262>& regex() { return m_regex; }
+    const Regex<ECMAScript262>& regex() const { return m_regex; }
 
 private:
     virtual bool is_regexp_object() const override { return true; }
 
     String m_content;
     String m_flags;
+    Regex<ECMAScript262> m_regex;
```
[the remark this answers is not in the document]
alimpfard: I'll go with ECMA262 then~

Libraries/LibRegex/RegexMatcher.cpp
alimpfard: There's a whole bunch, I'll deal with them when the js stuff is happily where it should be :P

Libraries/LibJS/Runtime/RegExpPrototype.cpp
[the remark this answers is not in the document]
alimpfard: I don't quite understand how an instance property translates to LibJS's object system; should `RegexpObject::create()` just define a property on the object it allocates? :thinking:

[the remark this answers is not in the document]
linusg: > should RegexpObject::create() just define a property on the object it allocates?

Pretty much, yeah. This should work (but has a duplicated `regexp_object_from`):

```
diff --git a/Libraries/LibJS/Runtime/RegExpObject.cpp b/Libraries/LibJS/Runtime/RegExpObject.cpp
index d35edbd4f..6cef0fcbb 100644
--- a/Libraries/LibJS/Runtime/RegExpObject.cpp
+++ b/Libraries/LibJS/Runtime/RegExpObject.cpp
@@ -118,8 +118,53 @@ RegExpObject::RegExpObject(String content, String flags, Object& prototype)
     }
 }
 
+void RegExpObject::initialize(GlobalObject& global_object)
+{
+    auto& vm = this->vm();
+    Object::initialize(global_object);
+
+    define_native_property(vm.names.lastIndex, last_index, set_last_index, Attribute::Writable);
+}
+
 RegExpObject::~RegExpObject()
 {
 }
 
+static RegExpObject* regexp_object_from(VM& vm, GlobalObject& global_object)
+{
+    auto* this_object = vm.this_value(global_object).to_object(global_object);
+    if (!this_object)
+        return nullptr;
+    if (!this_object->is_regexp_object()) {
+        vm.throw_exception<TypeError>(global_object, ErrorType::NotA, "RegExp");
+        return nullptr;
+    }
+    return static_cast<RegExpObject*>(this_object);
+}
+
+JS_DEFINE_NATIVE_GETTER(RegExpObject::last_index)
+{
+    auto regexp_object = regexp_object_from(vm, global_object);
+    if (!regexp_object)
+        return {};
+
+    return Value((unsigned)regexp_object->regex().start_offset);
+}
+
+JS_DEFINE_NATIVE_SETTER(RegExpObject::set_last_index)
+{
+    auto regexp_object = regexp_object_from(vm, global_object);
+    if (!regexp_object)
+        return;
+
+    auto index = value.to_i32(global_object);
+    if (vm.exception())
+        return;
+
+    if (index < 0)
+        index = 0;
+
+    regexp_object->regex().start_offset = index;
+}
+
 }
diff --git a/Libraries/LibJS/Runtime/RegExpObject.h b/Libraries/LibJS/Runtime/RegExpObject.h
index c74eebf5c..5a0b4a63a 100644
--- a/Libraries/LibJS/Runtime/RegExpObject.h
+++ b/Libraries/LibJS/Runtime/RegExpObject.h
@@ -44,6 +44,7 @@ public:
     static RegExpObject* create(GlobalObject&, String content, String flags);
 
     RegExpObject(String content, String flags, Object& prototype);
+    virtual void initialize(GlobalObject&) override;
     virtual ~RegExpObject() override;
 
     const String& content() const { return m_content; }
@@ -55,6 +56,9 @@ public:
 private:
     virtual bool is_regexp_object() const override { return true; }
 
+    JS_DECLARE_NATIVE_GETTER(last_index);
+    JS_DECLARE_NATIVE_SETTER(set_last_index);
+
     String m_content;
     String m_flags;
     Flags m_active_flags;
diff --git a/Libraries/LibJS/Runtime/RegExpPrototype.cpp b/Libraries/LibJS/Runtime/RegExpPrototype.cpp
index d5bdf1341..04e7f1d2a 100644
--- a/Libraries/LibJS/Runtime/RegExpPrototype.cpp
+++ b/Libraries/LibJS/Runtime/RegExpPrototype.cpp
@@ -54,8 +54,6 @@ void RegExpPrototype::initialize(GlobalObject& global_object)
     define_native_property(vm.names.source, source, nullptr, readable_attr);
     define_native_property(vm.names.sticky, sticky, nullptr, readable_attr);
     define_native_property(vm.names.unicode, unicode, nullptr, readable_attr);
-
-    define_native_property(vm.names.lastIndex, last_index, set_last_index, Attribute::Writable);
 }
 
 RegExpPrototype::~RegExpPrototype()
@@ -117,31 +115,6 @@ JS_DEFINE_NATIVE_GETTER(RegExpPrototype::global)
     return Value(regexp_object->declared_options().has_flag_set(ECMAScriptFlags::Global)); // Note that this "Global" is actually "Global | Stateful"
 }
 
-JS_DEFINE_NATIVE_GETTER(RegExpPrototype::last_index)
-{
-    auto regexp_object = regexp_object_from(vm, global_object);
-    if (!regexp_object)
-        return {};
-
-    return Value((unsigned)regexp_object->regex().start_offset);
-}
-
-JS_DEFINE_NATIVE_SETTER(RegExpPrototype::set_last_index)
-{
-    auto regexp_object = regexp_object_from(vm, global_object);
-    if (!regexp_object)
-        return;
-
-    auto index = value.to_i32(global_object);
-    if (vm.exception())
-        return;
-
-    if (index < 0)
-        index = 0;
-
-    regexp_object->regex().start_offset = index;
-}
-
 JS_DEFINE_NATIVE_GETTER(RegExpPrototype::ignore_case)
 {
     auto regexp_object = regexp_object_from(vm, global_object);
diff --git a/Libraries/LibJS/Runtime/RegExpPrototype.h b/Libraries/LibJS/Runtime/RegExpPrototype.h
index 3a7c43ddd..1eaf86b20 100644
--- a/Libraries/LibJS/Runtime/RegExpPrototype.h
+++ b/Libraries/LibJS/Runtime/RegExpPrototype.h
@@ -50,9 +50,6 @@ private:
     JS_DECLARE_NATIVE_GETTER(sticky);
     JS_DECLARE_NATIVE_GETTER(unicode);
 
-    JS_DECLARE_NATIVE_GETTER(last_index);
-    JS_DECLARE_NATIVE_SETTER(set_last_index);
-
     JS_DECLARE_NATIVE_FUNCTION(test);
     JS_DECLARE_NATIVE_FUNCTION(to_string);
 };
```

> I tried RegExp.prototype.lastIndex in Firefox and it told me 0

`undefined` for me btw?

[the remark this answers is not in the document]
alimpfard: Thanks for the patch~

> `undefined` for me btw?

Interesting...maybe I got myself confused in my half-asleep state? :shrug:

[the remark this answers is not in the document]
alimpfard: I don't think the duplicated function is particularly troublesome, it doesn't look like it will ever need to have a logical change anyway.

Libraries/LibJS/Runtime/RegExpObject.cpp
```diff
@@ -24,13 +24,81 @@
...
 
+#include <AK/Function.h>
 #include <AK/StringBuilder.h>
 #include <LibJS/Heap/Heap.h>
 #include <LibJS/Runtime/GlobalObject.h>
 #include <LibJS/Runtime/PrimitiveString.h>
 #include <LibJS/Runtime/RegExpObject.h>
 #include <LibJS/Runtime/Value.h>
 
+static Flags options_from(const String& flags, JS::VM& vm, JS::GlobalObject& global_object)
```
linusg, line 35: What's the reason for this not being in the `JS` namespace? It looks like it would belong there to me, and we could drop a whole bunch of `JS::` prefixes.

Libraries/LibJS/Runtime/RegExpObject.cpp
```diff
@@ -24,13 +24,81 @@
...
+            return options;
+        }
+    }
+
+    return options;
+}
+
 namespace JS {
 
 RegExpObject* RegExpObject::create(GlobalObject& global_object, String content, String flags)
```
linusg, line 104: I know you didn't write this bit, but would you agree we rename "content" in every RegExp context (ctor, obj, proto) to "pattern"? That's what the spec and probably most people use for talking about "the string that makes up the regular expression".

Libraries/LibJS/Runtime/RegExpObject.cpp
alimpfard: Just something in my head saying "static helpers go in the global namespace or preferably an anonymous one in the global namespace"
I don't really mind putting it in the JS namespace, if you think that makes it better in any way, I'll move it there 🤷‍♂️

alimpfard: I went ahead and put it in the JS namespace.

linusg: ah, sorry I missed your comment! fwiw this was by no means an informed opinion, more of a gut feeling based on the (limited amount of) C++ code I have seen so far :smile:

Libraries/LibJS/Runtime/RegExpObject.cpp
linusg: thanks! :^)

alimpfard:
That's all the LibJS changes I had in mind for this PR.
it should be noted that this is only a part of what the spec says about `exec()` and `test()`, in particular, it completely ignores unicode stuff and the dynamic properties (such as using a user-provided `exec` function as opposed to the builtin) - this is noted as a FIXME.

Once the CI passes, I'll consider this ready for review (and the inevitable "what to do with LibC" discussion).

Libraries/LibJS/Runtime/RegExpPrototype.cpp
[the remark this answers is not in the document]
linusg: (Just for the record) I found how we did this with Array - `JS::Array` has `static Array* typed_this(VM&, GlobalObject&);`, and both `Array` and `ArrayPrototype` use `auto* array = Array::typed_this(vm, global_object);`.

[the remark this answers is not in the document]
alimpfard: Ah. well, that will probably come in handy once we decide to make `RegExpPrototype` respect the dynamic properties and stuff, I think that's a bit in the future still though :P

Libraries/LibRegex/Regex.hnew
```diff
@@ -0,0 +1,59 @@
+/*
```
linusg, line 1: This file should probably be deleted?

alimpfard: how did that even sneak in...?

Libraries/LibJS/Runtime/RegExpPrototype.cpp
```diff
@@ -43,6 +44,18 @@ void RegExpPrototype::initialize(GlobalObject& global_object)
     Object::initialize(global_object);
     u8 attr = Attribute::Writable | Attribute::Configurable;
     define_native_function(vm.names.toString, to_string, 0, attr);
+    define_native_function(vm.names.test, test, 1, attr);
+    define_native_function(vm.names.exec, exec, 1, attr);
+
+    u8 readable_attr = Attribute::Configurable;
+    define_native_property(vm.names.dotAll, dotAll, nullptr, readable_attr);
```
linusg, line 51: Please rename, for the sake of consistency:

```suggestion
    define_native_property(vm.names.dotAll, dot_all, nullptr, readable_attr);
```

Libraries/LibJS/Runtime/RegExpPrototype.cpp
```diff
@@ -61,12 +74,179 @@ static RegExpObject* regexp_object_from(VM& vm, GlobalObject& global_object)
     return static_cast<RegExpObject*>(this_object);
 }
 
+JS_DEFINE_NATIVE_GETTER(RegExpPrototype::dotAll)
```
linusg, line 77:
```suggestion
JS_DEFINE_NATIVE_GETTER(RegExpPrototype::dot_all)
```

Libraries/LibJS/Runtime/RegExpPrototype.h
```diff
@@ -39,6 +39,19 @@ class RegExpPrototype final : public RegExpObject {
     virtual ~RegExpPrototype() override;
 
 private:
+    static RegexResult do_match(const Regex<ECMA262>&, const StringView&);
+
+    JS_DECLARE_NATIVE_GETTER(dotAll);
```
linusg, line 44:
```suggestion
    JS_DECLARE_NATIVE_GETTER(dot_all);
```

linusg (changes requested):

alimpfard:
The individual commits show now build correctly (for everyone's bisecting pleasure).
@awesomekling (and anyone else with opinions), reviews, please - this is ready, imo

alimpfard:
:facepalm:

alimpfard:
At the cost of squashing all the commits to AK (it was breaking in all sorts of fun ways, no `std::initializer_list` because it's being built as part of the toolchain, fun name conflicts, and so on...), all commits are finally building correctly now (tested!).

Libraries/LibJS/Tests/builtins/RegExp/RegExp.prototype.test.js
```diff
@@ -0,0 +1,58 @@
...
+    expect(re.sticky).toBe(false);
+    expect(re.unicode).toBe(false);
+
+    expect(() => {
+        /foo/gg;
+    }).toThrowWithMessage(SyntaxError, "Repeated RegExp flag 'g'");
+
+    expect(() => {
+        /foo/x;
+    }).toThrowWithMessage(SyntaxError, "Invalid RegExp flag 'x'");
```
linusg, line 57: Tell me, JavaScript, why is this a SyntaxError during runtime? WHY?!

linusg (approved):
LibJS changes LGTM. There's a lot of spec-non-compliant stuff in there, but we can fix that gradually.

Userland/CMakeLists.txt
```diff
@@ -43,6 +44,7 @@ target_link_libraries(test-js LibJS LibLine LibCore)
 target_link_libraries(test-pthread LibThread)
 target_link_libraries(test-web LibWeb)
 target_link_libraries(tt LibPthread)
+target_link_libraries(grep LibRegex LibC)
```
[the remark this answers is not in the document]
alimpfard: Ah, that's true.
I'll fix that :)

Libraries/LibJS/Tests/builtins/RegExp/RegExp.prototype.test.js
alimpfard: JavaScript can't be _intuitive_, now can it?
that simply wouldn't work!

linusg: I had a look and this is what I found in the ES5.1 spec:

> https://www.ecma-international.org/ecma-262/5.1/#sec-7.8.5
> **Semantics**
> A regular expression literal evaluates to a value of the Object type that is an instance of the standard built-in constructor RegExp. This value is determined in two steps: first, the characters comprising the regular expression's RegularExpressionBody and RegularExpressionFlags production expansions are collected uninterpreted into two Strings Pattern and Flags, respectively. Then each time the literal is evaluated, a new object is created as if by the expression new RegExp(Pattern, Flags) where RegExp is the standard built-in constructor with that name. The newly constructed object becomes the value of the RegularExpressionLiteral. **If the call to new RegExp would generate an error as specified in 15.10.4.1, the error must be treated as an early error (Clause 16).**

In ES6+ that last part is nowhere to be found. At least that explains why some other engines make it an early error, but I think for the latest specs your implementations is correct.

alimpfard: Huh. that's interesting, looking at that clause 16:

> An implementation shall report all errors as specified, except for the following:
    - An implementation may extend program syntax and regular expression pattern or flag syntax. To permit this, all operations (such as calling eval, using a regular expression literal, or using the Function or RegExp constructor) that are allowed to throw SyntaxError are permitted to exhibit implementation-defined behaviour instead of throwing SyntaxError when they encounter an implementation-defined extension to the program syntax or regular expression pattern or flag syntax.

ES5.1 is pretty lenient about this too, "do whatever you want :shrug:"

awesomekling:
Let's do this!

PR #4110, opened 2020-11-17
Crypto: Remove foo.response file created by testing

ldm5180:
Problem:
- Test creates a file and leaves it in the source tree.

Solution:
- Remove the creation of the file since it is never checked.


alimpfard:
How about just not creating it?
It leaves the file because... well, it can't verify the response. But no one actually verifies the response anyway, so let's just not generate the file at all.

ldm5180:
@alimpfard Good point. Done.

PR #4111, opened 2020-11-17
Tests/Kernel: Remove redundant `if`

ldm5180:
Problem:
- If `fork()` fails the system tries to call `execl()`. That will
  either succeed and replace the running process image or it will fail
  and it needs to try again. The `if` is redundant because it will
  only be evaluated if `execl()` fails.

Solution:
- Remove the `if`.

### DIFF
--- a/Userland/Tests/Kernel/elf-execve-mmap-race.cpp
+++ b/Userland/Tests/Kernel/elf-execve-mmap-race.cpp
@@ -151,8 +151,7 @@ int main()
         if (!fork()) {
         try_again:
             printf("exec\n");
-            if (execl(path, "x", nullptr) < 0) {
-            }
+            execl(path, "x", nullptr);
             goto try_again;
         }
 


Userland/Tests/Kernel/elf-execve-mmap-race.cpp
```diff
@@ -152,8 +152,8 @@ int main()
         try_again:
             printf("exec\n");
             if (execl(path, "x", nullptr) < 0) {
+                goto try_again;
             }
-            goto try_again;
```
alimpfard, line 156: If `execl()` fails, we want it to try again, if it doesn't, then this `goto` can never be reached (short of a bug in `execve`).
this whole `if` is rather meaningless here, I think.

I suppose it's okay to be explicit about the jump though.

ldm5180: Removed the `if` completely.

PR #4113, opened 2020-11-19
Kernel: Move block condition evaluation out of the Scheduler

tomuta:
_This gets us another step closer to being able to make progress with #3864._

This makes the Scheduler a lot leaner by not having to evaluate
block conditions every time it is invoked. Instead evaluate them as
the states change, and unblock threads at that point.

- [x] Fix the various bugs that keep things from working
- [x] Figure out why Terminal only shows a "%" and doesn't accept input
- [x] Fix stopping/resuming processes
- [x] Fix g++ not exiting in HackStudio (`ReadBlocker` not unblocking)
- [x] Fix strace hanging (some problems with `WaitBlocker` and signal delivery/handling)
- [x] Validate that the 9p file system driver is still working
- [x] Reap unwaited child processes (zombies) when the parent exits

### DIFF
--- a/Kernel/Arch/i386/CPU.cpp
+++ b/Kernel/Arch/i386/CPU.cpp
@@ -63,6 +63,7 @@ static GenericInterruptHandler* s_interrupt_handler[GENERIC_INTERRUPT_HANDLERS_C
 extern "C" void enter_thread_context(Thread* from_thread, Thread* to_thread);
 extern "C" void context_first_init(Thread* from_thread, Thread* to_thread, TrapFrame* trap);
 extern "C" u32 do_init_context(Thread* thread, u32 flags);
+extern "C" void exit_kernel_thread(void);
 extern "C" void pre_init_finished(void);
 extern "C" void post_init_finished(void);
 extern "C" void handle_interrupt(TrapFrame*);
@@ -1457,6 +1458,11 @@ asm(
 "    jmp common_trap_exit \n"
 );
 
+void exit_kernel_thread(void)
+{
+    Thread::current()->exit();
+}
+
 u32 Processor::init_context(Thread& thread, bool leave_crit)
 {
     ASSERT(is_kernel_mode());
@@ -1468,7 +1474,7 @@ u32 Processor::init_context(Thread& thread, bool leave_crit)
         ASSERT(in_critical() == 1);
     }
 
-    const u32 kernel_stack_top = thread.kernel_stack_top();
+    u32 kernel_stack_top = thread.kernel_stack_top();
     u32 stack_top = kernel_stack_top;
 
     // TODO: handle NT?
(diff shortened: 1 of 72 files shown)


tomuta:
Hm seems like I have it *almost* working. Everything seems to work, except Terminal only shows a "%" for some reason. Everything else seems to be working, though. Probably just forgot kicking off blocking evaluation somewhere (maybe tty).

Screenshot from 2020-11-19 15-47-53

tomuta:
Hm, while I did indeed miss triggering notifications for TTY stuff, it didn't fix this problem...

tomuta:
Hm ok solved the terminal issue, missed some blocking evaluation checks.

tomuta:
@bugaevc I need some help validating that I didn't entirely break the 9P file system driver. I had to make some changes in there. I have a functioning diod server (I can mount my share with Linux), but I can't figure out how to mount it in Serenity. I patched my username and path in the source code, but I'm not having much luck with the `mount` command line.

bugaevc:
I'm glad to help, but I admit I don't quite understand the changes / the new blocking model (perhaps I should look at it again).

For Plan9FS, I originally planned to use a kernel thread like you've done it here, but then realized threads could cooperatively read messages for each other, and in the common case where there's one thread interacting with the FS, there wouldn't be an additional context switch. But I'm not opposed to using a kernel thread 🙂

I'm going to test this on my system and see if it still works then...

tomuta:
Well, I can test it if I can figure out how to get it mounted ;-)

Basically, prior to this change the scheduler would walk through all blocked threads and evaluate if they can be unblocked (e.g. by calling `can_read`/`can_write` on a file description. With this change, the scheduler no longer does this, instead when any state changes that could affect the return value of `can_read` or `can_write, the driver would call `evaluate_block_conditions`, which checks if there are any threads that are blocked and whether they should be unblocked.

It's not super stable yet, there are some cases where the system freezes, most likely because it didn't unblock someone.

bugaevc:
```
[mount9p(30:30)]: mount 9p: source fd 3 @ /mnt
[mount9p(30:30)]: ASSERTION FAILED: !PtrTraits::is_null(bits)
.././AK/RefPtr.h:458 in T* AK::RefPtr<T, PtrTraits>::as_nonnull_ptr(FlatPtr) const [with T = Kernel::Plan9FS::ReceiveCompletion; PtrTraits = AK::RefPtrTraits<Kernel::Plan9FS::ReceiveCompletion>; FlatPtr = unsigned int]
[mount9p(30:30)]: 0xc011f359  __assertion_failed(char const*, char const*, unsigned int, char const*) +185
[mount9p(30:30)]: 0xc0140654  Kernel::Plan9FS::post_message(Kernel::Plan9FS::Message&, AK::RefPtr<Kernel::Plan9FS::ReceiveCompletion, AK::RefPtrTraits<Kernel::Plan9FS::ReceiveCompletion> >) +290
[mount9p(30:30)]: 0xc01411d8  Kernel::Plan9FS::post_message_and_wait_for_a_reply(Kernel::Plan9FS::Message&, bool) +194
[mount9p(30:30)]: 0xc0142171  Kernel::Plan9FS::initialize() +121
[mount9p(30:30)]: 0xc018a3a4  Kernel::Process::sys$mount(AK::Userspace<Kernel::Syscall::SC_mount_params const*>) +3138
[mount9p(30:30)]: 0xc017d2d2  syscall_handler +1360
[mount9p(30:30)]: 0xc017cd23  syscall_asm_entry +49
[mount9p(30:30)]: 0x08048164  main +143
[mount9p(30:30)]: 0x0804831f  _start +94
```

bugaevc:
> but I'm not having much luck with the `mount` command line

Oh, `mount`(8) doesn't support mounting from sockets (or at least it haven't last time I looked at it). I use a tiny custom command (`mount9p`) whose source code you can find in https://github.com/SerenityOS/serenity/pull/2675

tomuta:
> > but I'm not having much luck with the `mount` command line
> 
> Oh, `mount`(8) doesn't support mounting from sockets (or at least it haven't last time I looked at it). I use a tiny custom command (`mount9p`) whose source code you can find in #2675

Dang, I overlooked the fact that the source code for that tool was in there... ;-)

tomuta:
Ok, was able to get it mounted with this command. This crasher is a fun one. DOH... 

```c++
m_completions.set(completion->tag, completion.release_nonnull());
```

tomuta:
I think I got most of the quirks worked out now :-)

awesomekling:
Great work :)

PR #4127, opened 2020-11-20
LibCore: Rename identifiers that can clash with libc macros

ericonr:
POSIX allows these default streams to be macros, which means that on
such systems (musl libc is one) building Lagom will fail due to the
File::stdin() functions. Undefining them at the top of the file is
enough to fix it.

### DIFF
--- a/Libraries/LibCore/File.cpp
+++ b/Libraries/LibCore/File.cpp
@@ -241,7 +241,7 @@ static RefPtr<File> stdin_file;
 static RefPtr<File> stdout_file;
 static RefPtr<File> stderr_file;
 
-NonnullRefPtr<File> File::stdin()
+NonnullRefPtr<File> File::standard_input()
 {
     if (!stdin_file) {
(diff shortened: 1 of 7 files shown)


awesomekling:
This doesn't feel like a very sturdy fix. What happens to code that includes `<LibCore/File.h>` after including such a libc header?
It seems better to just rename these functions on our side if it's a problem.
cc @bugaevc

bugaevc:
Agreed, undefining them is not great, let's rename ours instead.

ericonr:
Any suggestions for the names?

bugaevc:
Perhaps `std_in()`, or `standard_in()`, or `standard_input()`

ccapitalK:
I vote standard_input(), I feel like it makes the most sense.

ericonr:
Sorry for the time it took! Should follow the proposed solution now.

ericonr:
It also passes CI now :D

Issue #3643, opened 2020-09-29
LibIPC: Support passing file descriptors


itamar8910:
The kernel has support for passing open file descriptors over local sockets with the `sendfd` and `recvfd` syscalls, and it would be nice to integrate it in `LibIPC`.

@bugaevc suggested the following format in #3622:

 ```
OpenFile(fd file) =|
````
> Which translates to int file_fd on both sides, but actually passes an fd using sendfd()/recvfd() instead of an inline integer. Or, we could have some C++y wrapper type over a raw fd.

PR #4130, opened 2020-11-21
fd passing and unveil('b')

bugaevc:
* Prepend IPC messages with their sizes to make dealing with partial messages a little bit less involved
* `IPC::File` lets you easily pass an open fd over IPC (https://github.com/SerenityOS/serenity/issues/3643, cc @itamar8910)
    * I went with a real type instead of introducing a magical `fd` specifier
    * Using an `IPC::` type instead of say LibC's `FILE` seems more in line with other LibIPC types, and keeps the wrapper very thin (`IPC::File` is *just* a wrapper over an fd, it does not imply anything else). But I'm certainly open to arguments here :slightly_smiling_face:
* Use it in HackStudio language servers (as discussed in https://github.com/SerenityOS/serenity/issues/3643)
    * This means the C++ language server no longer needs any FS access :tada:
    * But the Shell language server wants to complete file paths
        * Add `unveil('b')` permission that lets you just browse directories, but not read any non-directories. To be exact, you can `open(O_RDONLY | O_DIRECTORY)` with *either* `r` or `b`, but the `open()` call will fail with `ENOTDIR` if it's not a directory because of `O_DIRECTORY`.
        * Use that in the Shell language server to browse (but not read!) the whole file system. cc @alimpfard

### DIFF
--- a/Base/usr/share/man/man2/unveil.md
+++ b/Base/usr/share/man/man2/unveil.md
@@ -28,6 +28,7 @@ include the following characters:
 * `w`: May write to a file at this path
 * `x`: May execute a program image at this path
 * `c`: May create or remove a file at this path
+* `b`: May browse directories at this path
 
 A single `unveil()` call may specify multiple permission characters at once.
 Subsequent `unveil()` calls may take away permissions from the ones allowed
@@ -77,6 +78,9 @@ unveil("/etc/WindowServer/WindowServer.ini", "rwc");
 
 // Allow the process to execute Calendar:
 unveil("/bin/Calendar", "x");
+
+// Allow the process to browse files from /usr/share:
+unveil("/usr/share", "b");
 
 // Disallow any further veil manipulation:
 unveil(nullptr, nullptr);
--- a/DevTools/HackStudio/Editor.cpp
+++ b/DevTools/HackStudio/Editor.cpp
@@ -48,6 +48,7 @@
 #include <LibWeb/DOM/Text.h>
 #include <LibWeb/HTML/HTMLHeadElement.h>
 #include <LibWeb/OutOfProcessWebView.h>
+#include <fcntl.h>
 
 // #define EDITOR_DEBUG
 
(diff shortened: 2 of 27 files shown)


bugaevc:
I'm not sure why the CI is failing. `recvfd()` is declared in `<sys/socket.h>`, which `Decoder.cpp` includes.

linusg:
> I'm not sure why the CI is failing. recvfd() is declared in <sys/socket.h>, which Decoder.cpp includes.

Presumably because it's the Lagom step that's failing and Linux doesn't have `recvfd()` in `sys/socket.h`?

DevTools/HackStudio/Editor.cpp
```diff
@@ -480,8 +481,17 @@ void Editor::set_document(GUI::TextDocument& doc)
         set_syntax_highlighter(nullptr);
     }
 
-    if (m_language_client)
-        m_language_client->open_file(code_document.file_path().string());
+    if (m_language_client) {
+        auto full_file_path = String::formatted("{}/{}", project().root_directory(), code_document.file_path());
+        dbg() << "Opening " << full_file_path;
+        int fd = open(full_file_path.characters(), O_RDONLY | O_NOCTTY);
```
alimpfard, line 487: do we actually set the controlling terminal to a terminal device when it's opened? I didn't expect to see `O_NOCTTY` here...?
for that matter, why is the user opening a non-regular file with hackstudio (and why are we allowing this)?

bugaevc:
> Presumably because it's the Lagom step that's failing and Linux doesn't have `recvfd()` in `sys/socket.h`?

Good point! I didn't expect Lagom to build LibIPC at all.

I think I'll just `#ifndef` fd passing out for non-Serenity then?

DevTools/HackStudio/Editor.cpp
bugaevc: Right, we don't seem to actually check it anywhere. We do have `Process::tty`, but that doesn't seem to influence anything either?

Perhaps let's just unimplement that and ignore controlling terminals the same way we ignore locales?

> for that matter, why is the user opening a non-regular file with hackstudio (and why are we allowing this)?

Perhaps some other code should do the check and disallow this (or does it already?), but the user can put a path to any kind of file into the project file...

alimpfard: I think I'd keep this specific `open()` behaviour out, at least.

Some other code _should_ check that and ensure that the user isn't trying to edit `/dev/urandom`, yeah :P

awesomekling:
> > Presumably because it's the Lagom step that's failing and Linux doesn't have `recvfd()` in `sys/socket.h`?
> 
> Good point! I didn't expect Lagom to build LibIPC at all.
> 
> I think I'll just `#ifndef` fd passing out for non-Serenity then?

Sure that makes sense. :)

Base/usr/share/man/man2/unveil.md
```diff
@@ -78,6 +79,9 @@ unveil("/etc/WindowServer/WindowServer.ini", "rwc");
 // Allow the process to execute Calendar:
 unveil("/bin/Calendar", "x");
 
+// Allow the process to browse files from /use/share:
```
awesomekling, line 82: Typo: `/usr/share`

awesomekling:
The `IPC::File` mechanism is so cool :D

PR #4131, opened 2020-11-21
LibWeb: Make event dispatcher spec-compliant and add the submit event to form

Lubrsi:
AK: Add first_matching and last_matching to Vector

first_matching returns the first item in the vector that matches
the given condition.

last_matching returns the last item in the vector that matches
the given condition.

-----

LibWeb: Make event dispatching spec-compliant

Specification: https://dom.spec.whatwg.org/#concept-event-dispatch

This also introduces shadow roots due to it being a requirement of
the event dispatcher.

However, it does not introduce the full shadow DOM, that can be
left for future work.

This changes some event dispatches which require certain attributes
to be initialised to a value.

-----

LibWeb: Expose ParentNode.{first,last}ElementChild

I needed these to write the event dispatcher test.

-----

LibWeb: Add a test for the new event dispatcher

-----

LibWeb: Add HTML::EventNames and UIEvents::EventNames

-----

LibWeb: Add the submit event to HTMLFormElement

Also adds the ability to submit from JavaScript.


Lubrsi:
It seems CI is hung on the shell-backgrounding test.

Libraries/LibWeb/DOM/EventListener.h
```diff
@@ -45,8 +45,28 @@ class EventListener
...
+    void set_passive(bool passive) { m_capture = passive; }
+
+    bool once() const { return m_once; }
+    void set_once(bool once) { m_once = once; }
+
+    bool removed() const { return m_removed; }
+    void set_removed(bool removed) { m_removed = removed; }
+
 private:
+    StringView m_type;
```
awesomekling, line 64: `StringView` should (almost) never be used as a member variable type, since there is no guarantee that the string data pointed to will remain valid. Instead, use a `FlyString` like we use for the event names elsewhere.

awesomekling (changes requested):

awesomekling:
This is some great work btw :)

PR #4148, opened 2020-11-23
FilePicker: Set Icon Based On Mode

### LINKED ISSUE #3910: LibGUI: GUI::CommonActions do not load GUI::FilePicker dialog icons
None of the `CommonActions` (open, save, save as, etc) load the correct icon for the file picker dialog. The default icon is loaded instead.

image

Despite the correct icon paths being defined for these actions :

https://github.com/SerenityOS/serenity/blob/46cd9157cde8b455907d212b5924ba59ba8f922f/Libraries/LibGUI/Action.cpp#L38-L51

supex0fan:
Resolves #3910.


supex0fan:
Just noticed that this line at the bottom of the commit:
`set_title(m_mode == Mode::Open ? "Open File" : "Save File");`
should be removed as it was replaced by the switch statement above it.

PR #4149, opened 2020-11-23
Spreadsheet: CSV save/load, copy one cell to many

### LINKED ISSUE #4010: Spreadsheet: cannot copy one cell to many
Currently you cannot copy the data from one cell and paste it into multiple other cells.

### LINKED ISSUE #4136: Spreadsheet: CSV support
The spreadsheet currently only handles its own json format but it would be handy to be able to export and import DOS and UNIX formatted csv files.

alimpfard:
This fixes #4010 and closes #4136.

The reader and writer can read (and write) whatever-separated-values, but Spreadsheet only allows saving to and loading from CSV (with normal quoting and quote escaping rules).
Spreadsheet does not try to guess the cell data type, so everything loaded from CSV is imported as literal strings, with no formatting applied.

### DIFF
--- a/Applications/Spreadsheet/CMakeLists.txt
+++ b/Applications/Spreadsheet/CMakeLists.txt
@@ -15,6 +15,7 @@ set(SOURCES
     CondFormattingViewUI.h
     HelpWindow.cpp
     JSIntegration.cpp
+    Readers/XSV.cpp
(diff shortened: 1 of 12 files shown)


alimpfard:
Uhhh...maybe pushing the 100,000-line benchmark file was a bad idea...

PR #4152, opened 2020-11-23
Kernel: Don't resume thread into Running state directly on SIGCONT

### LINKED ISSUE #4150: strace: Does not work at all anymore
Not sure when it broke. To reproduce: `strace id`

cc @tomuta

tomuta:
We should never resume a thread by directly setting it to Running state.
Instead, if a thread was in Running state when stopped, record the state
as Runnable.

Fixes #4150


awesomekling:
Thank you for fixing this :)

PR #4158, opened 2020-11-24
Once

bugaevc:
The implementation uses atomics and futexes (yay!) and is heavily based on the
implementation I did for my learning project named "Let's write synchronization
primitives" [0].

That project, in fact, started when I tried to implement pthread_once() for
Serenity (because it was needed for another project of mine, stay tuned :wink:) and
was not very sure I got every case right. So now, after learning some more about
code patterns around atomics and futexes, I am reasonably sure, and it's time to
contribute the implementation of pthread_once() to Serenity :^)

[0] To be published at https://github.com/bugaevc/lets-write-sync-primitives

cc @tomuta


tomuta:
I like it! :-)

PR #4162, opened 2020-11-25
Add more image loader fuzzers

nico:
Since we're now on oss-fuzz, it makes sense to do this sooner rather than later.

@DavidKorczynski Since oss-fuzz doesn't use upstream cmake files yet, I suppose this will require updating the oss-fuzz config after it goes in?


Meta/Lagom/Fuzzers/FuzzICOLoader.png
```diff
@@ -0,0 +1,35 @@
...
+
+#include <LibGfx/ICOLoader.h>
+#include <stddef.h>
+#include <stdint.h>
+
+extern "C" int LLVMFuzzerTestOneInput(const uint8_t* data, size_t size)
+{
+    Gfx::load_ico_from_memory(data, size);
+    return 0;
+}
```
linusg, line 35: `FuzzICOLoader.png` -> `FuzzICOLoader.cpp` :smile:

DavidKorczynski:
Yes, we should update OSS-Fuzz config files too. Ideally we should move the CMake stuff from OSS-Fuzz into here, would you like me to set that up?

Also for future references, it's not too much of an issue that causes OSS-Fuzz to break for a short period of time, so we can always push something here as long as we are ready to also do the OSS-Fuzz work shortly after.

PR #4174, opened 2020-11-26
Kernel: Reduce the scope we hold WaitQueue::m_lock during wake up

bgianfo:
Currently the spinlock of the wait queue is held for the extended
duration of the wake up, and then continuing to the scheduler yield.

This is excessive and will cause un-necessary contention on the
WaitQueue lock. The pattern used by most wait queues I've come
across is in production is to collect the targets to wake under
the lock, and then wakeup outside the lock. This gives you the
correctness desired, while still limiting the spin lock region
to the bare minimum.

I found the WAITQUEUE_DEBUG ifdefs really inhibit the readability
of the code,  but I did my best to preserve the semantics in case
someone finds this stuff useful.

### DIFF
--- a/Kernel/WaitQueue.cpp
+++ b/Kernel/WaitQueue.cpp
@@ -67,79 +67,117 @@ bool WaitQueue::dequeue(Thread& thread)
 
 void WaitQueue::wake_one(Atomic<bool>* lock)
 {
-    ScopedSpinLock queue_lock(m_lock);
-    if (lock)
+    if (lock) {
         *lock = false;
-    if (m_threads.is_empty()) {
-        // Save the fact that a wake was requested
-        m_wake_requested = true;
-#ifdef WAITQUEUE_DEBUG
-        dbg() << "WaitQueue " << VirtualAddress(this) << ": wake_one: nobody to wake, mark as pending";
-#endif
-        return;
     }
+
+    Thread* thread = nullptr;
+    {
+        ScopedSpinLock queue_lock(m_lock);
+        if (m_threads.is_empty()) {
+            // Save the fact that a wake was requested.
+            m_wake_requested = true;
 #ifdef WAITQUEUE_DEBUG
-    dbg() << "WaitQueue " << VirtualAddress(this) << ": wake_one:";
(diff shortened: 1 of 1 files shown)


bgianfo:
It's been a while since I sent a Serenity PR, so I'm not sure who to tag in this area anymore. 
Thanks 🙏

bgianfo:
Looks like you've been working in this area @tomuta ?

tomuta:
So I think to be able to reduce the lock duration you will need to place a `ScopedCritical` in place of where the original SpinLock region was. The reason for that is that leaving the `SpinLock` may trigger a call to `Scheduler::yield`, but you don't want to take the threads out of the list and implicitly yield, and when execution returns back you'd yield again. A ScopedCritical should postpone the yield until the function exits.

Another change to consider would be to change the list into a `NonnullRefPtr`. It won't be a problem now, but with this change I think it will be a problem once we actually run on multiple processors.

In regards to the `WAITQUEUE_DEBUG` ifdefs, I mainly added them because I found them *really* helpful tracking down wait-related bugs. Initially they were in a private branch only but after repeatedly trying to debug in that area I decided to just get them checked in for good.

Kernel/WaitQueue.cpp
```diff
@@ -67,79 +67,117 @@ bool WaitQueue::dequeue(Thread& thread)
...
+            m_wake_requested = true;
 #ifdef WAITQUEUE_DEBUG
-        dbg() << "WaitQueue " << VirtualAddress(this) << ": wake_all: nobody to wake, mark as pending";
+            dbg() << "WaitQueue " << VirtualAddress(this) << ": wake_all: nobody to wake, mark as pending";
 #endif
-        return;
+            return;
+        }
+
+        while (!m_threads.is_empty()) {
```
tomuta, line 159: This could be changed to a `do`...`while` loop.

tomuta:
@bgianfo, the `WaitQueue` class has since been converted to a `Thread::BlockCondition` and the implementation is now quite a bit different. I think this PR should be closed because it's no longer applicable.

bgianfo:
sure

PR #4176, opened 2020-11-26
Lagom: Various fixes to make Lagom run on OSS-Fuzz

DavidKorczynski:
Contains the various changes needed in the build environment to make OSS-Fuzz work.

There are no changes that break the regular build, i.e. it's only "additions" in case the OSS-Fuzz build is run. 

This PR will make things much easier to maintain from an OSS-Fuzz perspective, as very little - if any - will have to be maintained from the https://github.com/google/oss-fuzz repo. Once this PR is merged in then I will change the OSS-Fuzz repository to reflect the updates in this PR as well as the emails of the various maintainers that have asked to get in the configuration file. 

A few pointers on how to test the oss-fuzz set up locally:

### Build and run oss-fuzz fuzzers
Build and run serenity fuzzers by way of oss-fuzz
```
git clone https://github.com/google/oss-fuzz/
cd oss-fuzz
python3 infra/helper.py build_image serenity
python3 infra/helper.py build_fuzzers serenity
```
These commands will put the fuzzers in a folder in the oss-fuzz repo called: `build/out/serenity`. You can run the binaries in there individually, or simply type:
```
python3 infra/helper.py run_fuzzer serenity FUZZER_NAME
```

### asserts and aborts when fuzzing
Finally, a minor thing I noticed from the bug reports is that several `assert` statements are hit. This is naturally not necessarily a bug as the fuzzer may indicate. However, the fuzzer operates within a single process together with the library it is fuzzing, so if the code that is being fuzzed exits then the fuzzer will also exit and consider it a bug. As such, to make the code "fuzzing friendly" it is preferred to make the target code exit gracefully, i.e. instead of having asserts and aborts then simply return error codes. In this way the fuzzer can do more work - and you also get the benefit of potential malicious input not being able to trigger exits in a library (although this is completely your design decision to make).

### DIFF
--- a/AK/Random.h
+++ b/AK/Random.h
@@ -47,7 +47,7 @@ inline void fill_with_random(void* buffer, size_t length)
 {
 #if defined(__serenity__)
     arc4random_buf(buffer, length);
-#elif defined (OSS_FUZZ)
+#elif defined(OSS_FUZZ)
     (void)buffer;
     (void)length;
 #elif defined(__unix__) or defined(__APPLE__)
--- a/AK/Random.h
+++ b/AK/Random.h
@@ -48,16 +48,8 @@ inline void fill_with_random(void* buffer, size_t length)
 #if defined(__serenity__)
     arc4random_buf(buffer, length);
 #elif defined (OSS_FUZZ)
-    // Ugly hack to get OSS-Fuzz environment working, where we dont have the
-    // getentropy call. We use the conditional because we need buffer and
-    // length to be used.
-    int rc;
-    if (buffer && length) { 
-        rc = -1;
-    } 
-    else {
-        rc = -1;
-    }
+    (void)buffer;
+    (void)length;
 #elif defined(__unix__) or defined(__APPLE__)
     int rc = getentropy(buffer, length);
     (void)rc;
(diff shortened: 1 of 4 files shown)


AK/Random.h
```diff
@@ -47,6 +47,17 @@ inline void fill_with_random(void* buffer, size_t length)
...
+#elif defined (OSS_FUZZ)
+    // Ugly hack to get OSS-Fuzz environment working, where we dont have the
+    // getentropy call. We use the conditional because we need buffer and
+    // length to be used.
+    int rc;
+    if (buffer && length) { 
+        rc = -1;
+    } 
+    else {
+        rc = -1;
```
bugaevc, line 59: Wouldn't just
```suggestion
    (void)buffer;
    (void)length;
```
do the trick?

DavidKorczynski: It did - awesome, thanks!

DavidKorczynski:
@linusg @awesomekling Could you assist here? I am not entirely sure why the CI failed and if it is something I should fix

AK/Random.h
```diff
@@ -47,6 +47,9 @@ inline void fill_with_random(void* buffer, size_t length)
 {
 #if defined(__serenity__)
     arc4random_buf(buffer, length);
+#elif defined (OSS_FUZZ)
```
linusg, line 50:
```suggestion
#elif defined(OSS_FUZZ)
```

linusg (commented):
You'll need to run clang-format, otherwise CI will fail.

AK/Random.h
DavidKorczynski: Thanks a lot!

Issue #4191, opened 2020-11-27
find: find files by name


danboid:
find is missing a `-name` command to find files by their file name.

danboid:
@bugaevc

bugaevc:
Ah, yes, finding files by name would be cool. @alimpfard, how does one go about matching globs against file names, does the shell have an utility for this or something?

alimpfard:
globs are available as `StringUtils::matches()`
the Shell can expand globs given a base path and a pattern with `Shell::expand_globs(StringView path, StringView base)` (this is static).

alimpfard:
oh, did you mean that as a builtin? no, there's no dynamic glob resolution yet, I've had that on my wishlist though a "glob" builtin would be cool

bugaevc:
Oh right, how could I forget about `::matches`! Thanks, I'll submit an implementation soon then

PR #4197, opened 2020-11-28
find -name

bugaevc:
cc @alimpfard @danboid

https://github.com/SerenityOS/serenity/issues/4191


alimpfard:
Might as well implement `-iname`? it should just be an extra flag to `matches()`

bugaevc:
Indeed! That's what I get for only looking at POSIX manpages.

PR #4198, opened 2020-11-28
LibCore+Userland: Implement -I/-a (binary-mode) and -i (insensitive) for grep

alimpfard:
> LibCore: Add File::get_line()

cc @bugaevc - you've been asking for something like this since forever.

### DIFF
--- a/Libraries/LibCore/IODevice.cpp
+++ b/Libraries/LibCore/IODevice.cpp
@@ -200,8 +200,7 @@ ByteBuffer IODevice::read_line(size_t max_size)
             Vector<u8> new_buffered_data;
             new_buffered_data.append(m_buffered_data.data() + line_index, m_buffered_data.size() - line_index);
             m_buffered_data = move(new_buffered_data);
-            line[line_index] = '\0';
-            line.trim(line_index + 1);
+            line.trim(line_index);
             return line;
         }
     }
--- a/Userland/grep.cpp
+++ b/Userland/grep.cpp
@@ -35,6 +35,21 @@
 #include <stdio.h>
 #include <unistd.h>
 
+enum class BinaryFileMode {
+    Binary,
+    Text,
+    Skip,
+};
+
+template<typename... Ts>
+void fail(StringView format, Ts... args)
+{
+    fprintf(stderr, "\x1b[31m");
+    warnln(format, forward<Ts>(args)...);
+    fprintf(stderr, "\x1b[0m");
+    abort();
+}
+
 int main(int argc, char** argv)
 {
     if (pledge("stdio rpath", nullptr) < 0) {
@@ -47,11 +62,50 @@ int main(int argc, char** argv)
     bool recursive { false };
     bool use_ere { true };
     const char* pattern = nullptr;
+    BinaryFileMode binary_mode { BinaryFileMode::Binary };
+    bool case_insensitive = false;
 
     Core::ArgsParser args_parser;
     args_parser.add_option(recursive, "Recursively scan files starting in working directory", "recursive", 'r');
     args_parser.add_option(use_ere, "Extended regular expressions (default)", "extended-regexp", 'E');
     args_parser.add_option(pattern, "Pattern", "regexp", 'e', "Pattern");
+    args_parser.add_option(case_insensitive, "Make matches case-insensitive", nullptr, 'i');
+    args_parser.add_option(Core::ArgsParser::Option {
+        .requires_argument = true,
+        .help_string = "Action to take for binary files ([binary], text, skip)",
+        .long_name = "binary-mode",
+        .accept_value = [&](auto* str) {
+            if (StringView { "text" } == str)
+                binary_mode = BinaryFileMode::Text;
+            else if (StringView { "binary" } == str)
+                binary_mode = BinaryFileMode::Binary;
+            else if (StringView { "skip" } == str)
+                binary_mode = BinaryFileMode::Skip;
+            else
+                return false;
+            return true;
+        },
+    });
+    args_parser.add_option(Core::ArgsParser::Option {
+        .requires_argument = false,
+        .help_string = "Treat binary files as text (same as --binary-mode text)",
+        .long_name = "text",
+        .short_name = 'a',
+        .accept_value = [&](auto) {
+            binary_mode = BinaryFileMode::Text;
+            return true;
+        },
+    });
+    args_parser.add_option(Core::ArgsParser::Option {
+        .requires_argument = false,
+        .help_string = "Ignore binary files (same as --binary-mode skip)",
+        .long_name = nullptr,
+        .short_name = 'I',
+        .accept_value = [&](auto) {
+            binary_mode = BinaryFileMode::Skip;
+            return true;
+        },
+    });
     args_parser.add_positional_argument(files, "File(s) to process", "file", Core::ArgsParser::Required::No);
     args_parser.parse(argc, argv);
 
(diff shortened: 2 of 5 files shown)


bugaevc:
This is great! But there are a few issues with this approach:
* Are you `fclos`ing the `FILE`? And how would do that without closing the underlying fd? Right, you would `dup` it, which brings me to...
* Performance of creating a fresh `FILE` each time you read a line will be dismal. The who point (well, not the whole point, but a part of the point) is that LibC `FILE` does its own buffering, which brings me to...
* LibC *will* read more data from the underlying fd than it returns back to you from `getline()`, due to the buffering it does. This is transparent if you continue using the `FILE`, but not if you throw it away or forget about it
* Finally, this should not be specific to `Core::File`, but rather to `Core::IODevice`

alimpfard:
That's true...I kinda wanted to get it done so I could work on grep ^^'
So then I have two ideas to deal with this:
- put a `FILE*` in all `Core::IODevice`'s and always go through that 
- make this a static member function and just take a `FILE*`

I think I like the second idea

bugaevc:
A more promising approach would be to make `IODevice` wrap a `FILE` instead of a raw fd (keeping the getter working of course, to not break tons of code in event loop and elsewhere), and rewrite all of its existing I/O methods on top of stdio.

Ah, I see you're suggesting the same.

alimpfard:
but then there will be the whole stdio buffering thing, `Core::IODevice` has its own idea of how it should buffer things - I don't think mixing the two (or defining one on top of the other) would be a very interesting approach

bugaevc:
Honestly I'd just throw away all the buffering `IODevice` does and just use stdio. But I might be biased :slightly_smiling_face:

@awesomekling, what would you prefer?

alimpfard:
I'm not sure what depends on IODevice's buffering, at the very least, lots of things use `can_read_line()`, I don't think stdio has a direct equivalent of that?
let's see what @awesomekling prefers though :P

Libraries/LibCore/IODevice.cpp
```diff
@@ -302,6 +302,31 @@ int IODevice::printf(const char* format, ...)
     return ret;
 }
 
+Result<String, IODevice::GetLineError> IODevice::get_line(FILE* fp, bool should_chomp)
```
alimpfard, line 304: For reference, this is my second idea.

bugaevc:
`can_read_line()` is an interesting one (for one thing, it should be named `can_read_line_without_blocking()`), but indeed there's no direct equivalent in stdio, because stdio only does bounded buffering (which is probably a good thing).

On the other hand, I think any code that actually uses `can_read_line()` either doesn't realize it is does the non-blocking check and not just whether EOF is reached, or uses a fixed-sized buffer for reading the line (or both). Reading a line into a fixed-size buffer without blocking could be relatively easily implemented on top of `fgets()`.

Libraries/LibCore/IODevice.cpp
```diff
@@ -302,6 +302,31 @@ int IODevice::printf(const char* format, ...)
...
+{
+    size_t size = 0;
+    char* line = nullptr;
+    auto line_length = getline(&line, &size, fp);
+    // getline() returns -1 and sets errno=0 on EOF.
```
bugaevc, line 309: You're actually supposed to check `feof()` to detect EOF in case it returns -1.

alimpfard: There we go, already living up to the "supposed to" in the commit message lol
what happens if errno==0 and it returns -1, but `!feof()`?

Userland/grep.cpp
```diff
@@ -47,75 +62,108 @@ int main(int argc, char** argv)
...
+    args_parser.add_option(Core::ArgsParser::Option {
+        .requires_argument = false,
+        .help_string = "Ignore binary files (same as --binary-mode skip)",
+        .long_name = nullptr,
+        .short_name = 'I',
+        .accept_value = [&](auto) {
+            binary_mode_str = "skip";
+            return true;
+        },
+    });
```
bugaevc, line 93: Cool that you've mastered advanced args parser :smile:

Since you're doing that, perhaps get rid of `binary_mode_str` and just save to `binary_mode` directly everywhere? If you get a string you don't recognize, `return false` to args parser and it'll arrange to display an error and die.

alimpfard: That's a better idea, yeah!

Libraries/LibCore/IODevice.cpp
bugaevc: If stdio functions that work on a `FILE` fail, they don't typically even set errno, they set the internal error you can retrieve with `ferror()` instead.

`getline()` appears to be documented to set errno instead, which our implementation doesn't seem to do:
https://github.com/SerenityOS/serenity/blob/952c0dc2a02e92ba6380cba7fde8e4f9f00af257/Libraries/LibC/stdio.cpp#L675-L683
So you'll get some garbage in errno. If you're lucky, this will be the errno of the last syscall stdio performed, which *should* generally match the last error, but we should not rely on that.

alimpfard: Noted. I'll just ignore errno in that case then.
guess I'm also fixing everything else that is using `getline()` (wrong...)

alimpfard:
This poor PR is now transitioning to a LibRegex issue fix thing; I'll split it up if `grep`'s discussion lengthens.

Userland/mount.cpp
```diff
@@ -96,7 +96,7 @@ static bool mount_all()
     bool all_ok = true;
     while (fstab->can_read_line()) {
         ByteBuffer buffer = fstab->read_line(1024);
-        StringView line_view = (const char*)buffer.data();
+        StringView line_view { buffer.data(), buffer.size() };
```
alimpfard, line 99: what the heck @bugaevc :/

awesomekling:
> Honestly I'd just throw away all the buffering `IODevice` does and just use stdio. But I might be biased 
> 
> @awesomekling, what would you prefer?

I think a single buffering implementation makes sense. And keeping it in stdio for maximum reuse makes sense to me as well.

Userland/mount.cpp
bugaevc: Yeah, thanks for catching it :|

bugaevc:
I think we should also decide what those LibCore APIs actually do.

Does `read()` read exactly as much as asked (unless it hits and error or until EOF), like `fread()` does? Does it read as much as a single `read()` call returns? Does it read as much as it's possible to read without blocking?

Userland/sort.cpp
```diff
@@ -45,17 +46,15 @@ int main(int argc, char** argv)
...
-        ssize_t buflen = 0;
-        errno = 0;
-        buflen = getline(&buffer, nullptr, stdin);
-        if (buflen == -1 && errno != 0) {
+        auto result = Core::IODevice::get_line(stdin, false);
+        if (result.is_error()) {
+            if (result.error() == Core::IODevice::Eof)
+                break;
+            errno = result.error();
             perror("getline");
```
awesomekling, line 54: Not a fan of assigning to `errno` like this just so we can use `perror`.
How about this instead:

```cpp
warnln("getline: {}", strerror(result.error());
```

alimpfard: Eh, it did feel somewhat silly when I did that, I guess just using `strerror()` is better...until someone changes `perror` to do something different?

awesomekling: `perror` is a standard C library function though, so it shouldn't be changed to do something different :)

alimpfard: Fair enough.

alimpfard:
There's really only one thing I dislike about moving this `IODevice` stuff on top of libc, since there's no direct equivalent of `can_read_line()` (or even a safe equivalent of `can_read()`, I don't think `select` would work correctly with libc's buffers), we'll have to change other bits in the codebase (this is fine in isolation) and then somehow do a "read line non-blockingly" operation.
the only ways I can come up with for the second bit require buffering in `IODevice`, or direct use of libc (which is, I think, a step back in terms of code quality and readability)

bugaevc:
Again, we have to first decide what the relationship between `IODevice` and non-blocking I/O is.

alimpfard:
Its users certainly seem to be treating it as nonblocking reads.
take IRCClient for an example, a loop on `can_read_line()` as a result of `on_ready_to_read` (i.e. `select`) - sounds like nonblocking to me

bugaevc:
True, but there's also a lot of code that just uses it as a supposedly more convenient C++y I/O library for normal, blocking, reads. See the uses of the `read_all()` method as one example of this.

awesomekling:
The IODevice API feels really badly fitted. It collides conceptually with `IODevice::read_line()` and why is it in IODevice in the first place?

Like I said, it would be nice to unify buffering, but this feels like tacking on a new abstraction in a basically random place?

alimpfard:
honestly, I saw no "good" place to put it. I initially put it in `Core::File` (as `getline()` works on a `FILE*`, I thought "hey, that sorta fits"), but that's not really a good place to put it either.
what do you suggest we do with it? I definitely want this wrapper - `getline()` is far from the most friendly API around...

awesomekling:
I suggest we use `IODevice::read_line()` and improve on that API (internally and externally) as needed in a different PR :)
No need to turn this `grep` feature into a snowball

alimpfard:
Sure, I'll try using `IODevice::read_line()` in this PR.
but we'll someday have to deal with this `getline()` thing anyway :thinking:

alimpfard:
We'll just keep the whole `IODevice` talk to its own PR and its own time, then.

PR #4230, opened 2020-11-29
AudioServer: Mixer: limit max volume to 100

bcoles:
Prior to this PR it was possible to set system volume to maxint using `avol` or IPC. It hurt.


awesomekling:
> Prior to this PR it was possible to set system volume to maxint using `avol` or IPC. It hurt.

[laughs heartily]

Issue #4215, opened 2020-11-28
LibGfx: Vector OOB in decode_jpg when decoding bogus JPG


Lubrsi:
From OSS-Fuzz. https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=27870&sort=-id&q=proj%3Dserenity&can=2

https://oss-fuzz.com/download?testcase_id=6395000203771904

Crash State:
```
  i < m_size
  Gfx::decode_jpg
  Gfx::load_jpg_impl
```

Note: Apparently this is happening very frequently and is preventing JPG fuzzer from getting very far.

As their bot says:
```
This crash occurs very frequently on linux platform and is likely preventing the fuzzer FuzzJPGLoader from making much progress. Fixing this will allow more bugs to be found.
```

Issue #4212, opened 2020-11-28
LibGfx: Vector OOB in build_macroblocks when decoding bogus JPG


Lubrsi:
From OSS-Fuzz. https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=27878&sort=-id&q=proj%3Dserenity&can=2

https://oss-fuzz.com/download?testcase_id=5640937853943808

Crash State:
```
  i < m_size
  Gfx::build_macroblocks
  Gfx::decode_huffman_stream
```

Lubrsi:
Another file to test against: https://oss-fuzz.com/download?testcase_id=5640937853943808

Issue #4214, opened 2020-11-28
LibGfx: OOM when decoding bogus JPG


Lubrsi:
From OSS-Fuzz. https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=27883&sort=-id&q=proj%3Dserenity&can=2

https://oss-fuzz.com/download?testcase_id=6279404313116672

Lubrsi:
Note: Can't seem to reproduce it inside Serenity. Haven't tried with Lagom builds though.

linusg:
OSS-Fuzz considers this fixed (https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=27883#c3), issue can be closed.

PR #4237, opened 2020-11-29
Jpeg fixes

devsh0:
Fixes: #4212.
Fixes: #4215.

Besides, this patch makes the JPEG decoder slightly more spec-compliant than it used to be by allowing arbitrary component IDs in the interval [0-255]. Although the OOBs have been fixed, the program requesting the decode still crashes due to out-of-memory errors. File manager dies with the following message:

```
USERSPACE(30) ASSERTION FAILED: ptr != MAP_FAILED
../Libraries/LibC/malloc.cpp:158
```
@awesomekling 
I am not sure what to do with these OOMs. Valid image dimensions as per the JPEG spec is {width [0-65,535] x height (0-65,535]}. An image with (valid) maximum dimensions will require memory for 67,108,864 macroblocks but it looks like we are choking at some point between 270,000-280,000 macroblocks. Fortunately, images will rarely be the maximum size that's considered valid in JPEG. The "popular" image dimension that can fit under 270,000 macroblocks (which won't cause OOM) is 5100x3300 px. Some advice on what should be done to avoid these OOM errors will be highly appreciated :)

@Lubrsi 
I was unable to reproduce #4214 with these changes. Can you confirm?


Lubrsi:
Can't reproduce it inside Serenity. Haven't tested it with Lagom builds though.

PR #4240, opened 2020-11-29
LibVT: Make terminal scrollback max size configurable / Terminal: Allow the user to configure the maximum history size 

### LINKED ISSUE #4238: Terminal: Increase scrollback limit
The terminal is currently limited to displaying 500 lines of scrollback. It would be nice to see this increased to 10K+.

alimpfard:
(Also increase default scrollback size to 1K)
Closes #4238.

### DIFF
--- a/Applications/Terminal/main.cpp
+++ b/Applications/Terminal/main.cpp
@@ -157,7 +157,7 @@ static RefPtr<GUI::Window> create_settings_window(TerminalWidget& terminal)
     auto window = GUI::Window::construct();
     window->set_title("Terminal Settings");
     window->set_resizable(false);
-    window->resize(200, 185);
+    window->resize(200, 230);
     window->set_modal(true);
 
     auto& settings = window->set_main_widget<GUI::Widget>();
@@ -205,6 +205,19 @@ static RefPtr<GUI::Window> create_settings_window(TerminalWidget& terminal)
     spinbox.set_value(terminal.scroll_length());
     spinbox.on_change = [&terminal](int value) {
         terminal.set_scroll_length(value);
+    };
+
+    auto& history_size_spinbox_container = settings.add<GUI::GroupBox>("Maximum scrollback history lines");
+    history_size_spinbox_container.set_layout<GUI::VerticalBoxLayout>();
+    history_size_spinbox_container.layout()->set_margins({ 6, 16, 6, 6 });
+    history_size_spinbox_container.set_size_policy(GUI::SizePolicy::Fill, GUI::SizePolicy::Fixed);
+    history_size_spinbox_container.set_preferred_size(100, 46);
+
+    auto& history_size_spinbox = history_size_spinbox_container.add<GUI::SpinBox>();
+    history_size_spinbox.set_range(0, 40960);
+    history_size_spinbox.set_value(terminal.max_history_size());
+    history_size_spinbox.on_change = [&terminal](int value) {
+        terminal.set_max_history_size(value);
     };
 
     return window;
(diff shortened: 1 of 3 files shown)


Applications/Terminal/main.cpp
```diff
@@ -207,6 +207,19 @@ static RefPtr<GUI::Window> create_settings_window(TerminalWidget& terminal)
...
+    history_size_spinbox_container.layout()->set_margins({ 6, 16, 6, 6 });
+    history_size_spinbox_container.set_size_policy(GUI::SizePolicy::Fill, GUI::SizePolicy::Fixed);
+    history_size_spinbox_container.set_preferred_size(100, 46);
+
+    auto& history_size_spinbox = history_size_spinbox_container.add<GUI::SpinBox>();
+    history_size_spinbox.set_range(512, 40960);
+    history_size_spinbox.set_value(terminal.max_history_size());
+    history_size_spinbox.on_change = [&terminal](int value) {
+        terminal.set_max_history_size(value);
+    };
```
alimpfard, line 221: This is a little awkward, trying to type in the spinbox (because no one will click the buttons 9000 times) will:
- reset to 512 if you try to type a digit
- reset to 40960 if you try to clear
- *not* care if you go above 40960 (!?)

I'd rather these corrections happened on focus out instead...

Applications/Terminal/main.cpp
```diff
@@ -207,6 +207,19 @@ static RefPtr<GUI::Window> create_settings_window(TerminalWidget& terminal)
...
     };
 
+    auto& history_size_spinbox_container = settings.add<GUI::GroupBox>("Maximum scrollback history lines");
+    history_size_spinbox_container.set_layout<GUI::VerticalBoxLayout>();
+    history_size_spinbox_container.layout()->set_margins({ 6, 16, 6, 6 });
+    history_size_spinbox_container.set_size_policy(GUI::SizePolicy::Fill, GUI::SizePolicy::Fixed);
+    history_size_spinbox_container.set_preferred_size(100, 46);
+
+    auto& history_size_spinbox = history_size_spinbox_container.add<GUI::SpinBox>();
+    history_size_spinbox.set_range(512, 40960);
```
linusg, line 217: I guess a min value of 0 to disable scrollback would be nice? :)

Libraries/LibVT/Terminal.h
```diff
@@ -93,7 +93,24 @@ class Terminal {
...
+    size_t max_history_size() const { return m_max_history_lines; }
+    void set_max_history_size(size_t value)
+    {
+        if (m_max_history_lines > value) {
+            NonnullOwnPtrVector<Line> new_history;
+            new_history.ensure_capacity(value);
+            auto existing_line_count = min(m_history.size(), value);
+            for (size_t i = m_history.size() - existing_line_count; i < m_history.size(); ++i) {
+                auto j = (m_history_start + i) % m_history.size();
+                dbg() << "Copying index " << j << " into index " << i;
```
awesomekling, line 105: If you want to keep this debug logging, at least please use `dbgln()` in new code :)

Applications/Terminal/main.cpp
alimpfard: what sort of madman would do that...?

Libraries/LibVT/Terminal.h
alimpfard: Ah, sorry, don't need that

Applications/Terminal/main.cpp
alimpfard: are you honestly serious about this? I don't mind doing that, but it sounds like crazy talk to me

linusg: don't feel like you have to implement it if the current solution doesn't allow it, but enforcing a limit on the lower end seems rather arbitrary :shrug:

alimpfard: I'll let it do that if setting it to zero doesn't crash anything, I guess.
let's see...

alimpfard: Ah, whatever, if the user wants to disable scrollback, they should be able to :shrug:

PR #4245, opened 2020-11-29
LibRegex: Use match_ordinary_characters() in ECMA262Parser::parse_atom()

### LINKED ISSUE #4243: LibJS: /a-b/ should be a valid regular expression
cc @alimpfard :)

linusg:
Otherwise we would only match TokenType::Char, making all of these invalid:

- /foo,bar/
- /foo\\/bar/
- /foo=bar/
- /foo-bar/
- /foo:bar/

Fixes #4243.


Libraries/LibRegex/RegexParser.cpp
```diff
@@ -967,7 +967,7 @@ bool ECMA262Parser::parse_atom(ByteCode& stack, size_t& match_length_minimum, bo
         return false;
     }
 
-    if (match(TokenType::Char)) {
+    if (match_ordinary_characters()) {
```
alimpfard, line 970: What group of tokens aren't allowed here? I know a bunch of them are excluded above, and a few more are already checked above.
(this change is alright, I'm just wondering out loud)

alimpfard (approved):

PR #4204, opened 2020-11-28
Build *ALL* the debug code, and complain if some are forgotten

BenWiederhake:
Disadvantage: The command-line invocation gets longer and longer.

```
/usr/bin/ccache /home/eispin/workspace/serenity/Toolchain/Local/bin/i686-pc-serenity-g++  -I../Libraries
-I../. -I../Libraries/LibC -I../Libraries/LibM -I../Services -I. -IServices -ILibraries
-Wno-unknown-warning-option -Wall -Wextra -Werror -Wmissing-declarations -std=c++2a -fdiagnostics-color=always
-fconcepts -DACPI_DEBUG -DAPIC_DEBUG -DAPIC_SMP_DEBUG -DARP_DEBUG -DAWAVLOADER_DEBUG -DBBFS_DEBUG -DBMP_DEBUG
-DBXVGA_DEBUG -DCACHE_DEBUG -DCALLBACK_MACHINE_DEBUG -DCHTTPJOB_DEBUG -DCNETWORKJOB_DEBUG -DCOPY_DEBUG
-DCOMPOSE_DEBUG -DCONTEXT_SWITCH_DEBUG -DCONTIGUOUS_VMOBJECT_DEBUG -DCRYPTO_DEBUG -DCSOCKET_DEBUG
-DDEBUG_AUTOCOMPLETE -DDEBUG_CPP_LANGUAGE_SERVER -DDEBUG_CURSOR_TOOL -DDEBUG_DIFF -DDEBUG_FILE_CONTENT
-DDEBUG_GZIP -DDEBUG_HIGHLIGHT_FOCUSED_FRAME -DDEBUG_HUNKS -DDEBUG_IO -DDEBUG_ITEM_RECTS -DDEBUG_LOCAL_SOCKET
-DDEBUG_MAPPED_FILE -DDEBUG_MARKDOWN -DDEBUG_MENUS -DDEBUG_POLL_SELECT -DDEBUG_SH_LANGUAGE_SERVER -DDEBUG_SPAM
-DDEBUG_STRINGIMPL -DDEBUG_SYNTAX_HIGHLIGHTING -DDEBUG_TEXTEDITOR -DDEFERRED_INVOKE_DEBUG -DDHCPV4CLIENT_DEBUG
[...]

PR #4249, opened 2020-11-29
LibRegex: Give the bytecode a chance to run when there's no input

### LINKED ISSUE #4246: LibJS: /(?:)/.test("") === false
Should be `true` :)

alimpfard:
Fixes #4246. 

@BenWiederhake: if this is merged before you go back to working on #4204, it should fix your little regex test problem; if not, feel free to take 3412851, and I'll remove it from this PR.


Lubrsi:
May be a good idea to add a test case for this :)

alimpfard:
Will do, once I make it not fail the existing tests

PR #4261, opened 2020-11-30
TaskBar: Allow QuickLaunch apps to be added and removed from the taskbar graphically.

### LINKED ISSUE #4058: Add apps to taskbar
It would be useful if you could right-click apps in the menu and choose to add a shortcut for that app to the Taskbar and also remove shortcuts from the Taskbar via a context menu.

Petelliott:
This ended up being more complicated than I expected. Here is a quick rundown of the changes:

- `.af` files are now launched with `Core::App::launch()` which starts the application with the `AF_PATH` environment variable 
- `LibGUI` reads the value of `AF_PATH` and sends it to `WindowServer` when creating a window.
- Context menus for windows now have the option to pin the app to the taskbar.
- When the 'pin' option is clicked, `WindowServer` produces a WM event for WM listeners.
- Taskbar handles the event by adding the application to the taskbar, and saving it.

Closes #4058

### DIFF
--- a/Services/SystemMenu/main.cpp
+++ b/Services/SystemMenu/main.cpp
@@ -46,6 +46,7 @@ struct AppMetadata {
     String name;
     String icon_path;
     String category;
+    String af_path;
 };
 Vector<AppMetadata> g_apps;
 
@@ -112,7 +113,7 @@ Vector<String> discover_apps_and_categories()
         auto app_executable = af->read_entry("App", "Executable");
         auto app_category = af->read_entry("App", "Category");
         auto app_icon_path = af->read_entry("Icons", "16x16");
-        g_apps.append({ app_executable, app_name, app_icon_path, app_category });
+        g_apps.append({ app_executable, app_name, app_icon_path, app_category, af_path });
         seen_app_categories.set(app_category);
     }
     quick_sort(g_apps, [](auto& a, auto& b) { return a.name < b.name; });
@@ -164,7 +165,10 @@ NonnullRefPtr<GUI::Menu> build_system_menu()
             const auto& bin = g_apps[app_identifier].executable;
             pid_t child_pid;
             const char* argv[] = { bin.characters(), nullptr };
-            if ((errno = posix_spawn(&child_pid, bin.characters(), nullptr, nullptr, const_cast<char**>(argv), environ))) {
+
+            auto af_key = String::formatted("AF_PATH={}", g_apps[app_identifier].af_path);
+            const char* env[] = { af_key.characters(), nullptr };
+            if ((errno = posix_spawn(&child_pid, bin.characters(), nullptr, nullptr, const_cast<char**>(argv), const_cast<char**>(env)))) {
                 perror("posix_spawn");
             } else {
                 if (disown(child_pid) < 0)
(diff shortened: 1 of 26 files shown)


awesomekling:
cc @bugaevc

Services/SystemMenu/main.cpp
```diff
@@ -164,7 +165,10 @@ NonnullRefPtr<GUI::Menu> build_system_menu()
             const auto& bin = g_apps[app_identifier].executable;
             pid_t child_pid;
             const char* argv[] = { bin.characters(), nullptr };
-            if ((errno = posix_spawn(&child_pid, bin.characters(), nullptr, nullptr, const_cast<char**>(argv), environ))) {
+
+            auto af_key = String::formatted("AF_PATH={}", g_apps[app_identifier].af_path);
+            const char* env[] = { af_key.characters(), nullptr };
+            if ((errno = posix_spawn(&child_pid, bin.characters(), nullptr, nullptr, const_cast<char**>(argv), const_cast<char**>(env)))) {
```
bugaevc: Is this the only env we pass to our... wait, why is SystemMenu even spawning apps? Shouldn't LaunchServer do that?

Libraries/LibCore/App.h
```diff
@@ -0,0 +1,55 @@
...
+    const String& af_path() const { return m_af_path; }
+
+    int launch(bool chdir = true) const;
+
+    bool is_well_formed() const;
+
+private:
+    String m_af_path;
+    NonnullRefPtr<ConfigFile> m_config_file;
+};
```
bugaevc, line 53: I don't think this is a good fit for LibCore. LibDesktop perhaps?

Petelliott: makes sense

Services/SystemMenu/main.cpp
Petelliott: the env is now handled by `Core::App::Launch()`. SystemMenu and Taskbar both use it. I've been waning to add af files to LaunchServer anyway so that you can put them on your desktop.

Petelliott: I've added support for af apps in LaunchServer by default. However, I think that we should not use it to launch apps from SystemMenu, because SystemMenu only ever launches Apps, and I don't see why we would want the behavior of launching an AF app to be user defined.

danboid:
@bugaevc

awesomekling:
Some mixed feelings about this PR:

* I'm happy that you're working on this! It's something that currently feels overly rigid about the desktop.
* I'm unsure that the pin/unpin metaphor is a good fit for our windowing model. It feels better suited to "dock" style window management.
* I don't like that WindowServer has to know about any of this.

IIRC in Win9x this sort of thing worked by placing links in a special directory, and whatever you placed in there would then show up in the quicklaunch bar. I wonder if we could do something similar, and then allow you to drag/drop apps onto the bar from e.g FileManager/Desktop/SystemMenu. The SystemMenu could work in the same way.

Petelliott:
> I'm unsure that the pin/unpin metaphor is a good fit for our windowing model. It feels better suited to "dock" style window management.

It differs only in name. I can change all references to 'add to dock' or whatever

> IIRC in Win9x this sort of thing worked by placing links in a special directory, and whatever you placed in there would then show up in the quicklaunch bar. I wonder if we could do something similar, and then allow you to drag/drop apps onto the bar from e.g FileManager/Desktop/SystemMenu. The SystemMenu could work in the same way.

Do Serenity's FileSystem notifications work yet? we could have taskbar monitor `Taskbar.ini` for changes, but I think it's still a good idea to add an app from it's context menu, and that's handled in `WindowServer`.

PR #4271, opened 2020-11-30
Spreadsheet: Invert the drag-selection trigger

### LINKED ISSUE #4268: Spreadsheet: Drag select multiple cells
I thought this was supposed to be fixed by this mornings SS commits but it doesn't seem so.

LMB clicking and dragging a cell still copy/pastes it instead of initiating a rectangular selection of cells.

alimpfard:
Make drag-selection the default behaviour, allowing (almost) any part of
the cell to initiate a select.
a small 5x5 rect at the corners of a cell can be used to initiate a
drag-copy instead.
Fixes #4268.


awesomekling:
It would be nice to show this 5x5 rect visually too, something like what Excel does.

alimpfard:
I did try that, but it looked really cluttered when it was always drawn
Will probably make it draw that on hover or something (laterz)

PR #4284, opened 2020-12-01
FileManager: Call on_selection_change with the correct view

### LINKED ISSUE #4185: FileManager: copy/paste broken
copy/paste is always greyed out when you right-click on a file or directory. I'm sure this used to work.

supex0fan:
Fixes #4185

### DIFF
--- a/Applications/FileManager/DirectoryView.cpp
+++ b/Applications/FileManager/DirectoryView.cpp
@@ -483,7 +483,7 @@ void DirectoryView::handle_selection_change()
     m_force_delete_action->set_enabled(can_delete);
 
     if (on_selection_change)
-        on_selection_change(*m_table_view);
+        on_selection_change(current_view());
 }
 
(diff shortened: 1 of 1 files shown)


Applications/FileManager/DirectoryView.cpp
```diff
@@ -482,8 +482,21 @@ void DirectoryView::handle_selection_change()
...
+        case ViewMode::Columns:
+            on_selection_change(*m_columns_view);
+            break;
+        case ViewMode::Icon:
+            on_selection_change(*m_icon_view);
+            break;
+        default:
+            ASSERT_NOT_REACHED();
+        }
+    }
```
awesomekling, line 499: You can simplify this by using `current_view()` :)

```cpp
if (on_selection_change)
	on_selection_change(current_view());
```

awesomekling (changes requested):

PR #4288, opened 2020-12-01
Browser: Add DuckDuckGo to bookmarks

xspager:
Google can't have all the fun can it?

### DIFF
--- a/Base/home/anon/.config/bookmarks.json
+++ b/Base/home/anon/.config/bookmarks.json
@@ -12,6 +12,10 @@
         "url": "https://www.google.com/"
     },
     {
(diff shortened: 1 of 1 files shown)


danboid:
duckduckgo.com doesn't open in Browser for me.

xspager:
I hope this motivate fixing some of what's preventing it from open. But I'm new here, so you tell me.

danboid:
I don't think it'll get accepted as a bookmark until we can load and use DDG but it's up to @awesomekling

danboid:
I do prefer to use DDG myself so and agree DDG should be a priority site to get working in Browser.

PR #4292, opened 2020-12-01
LibGfx: Add some validation to BMPLoader

nico:
These changes fixed various asserts when I ran the fuzzer locally a
while ago.


nico:
Shell-sigpipe failure looks like a flake.

PR #4293, opened 2020-12-01
CI: Use Ninja for buildling

nico:
Makes watching build output on Actions on PRs slightly more pleasant.

---

No idea if this actually works, but I guess the PR CI Action will tell me.


nico:
@BenWiederhake

BenWiederhake:
You don't need my permission – and it looks nicer indeed! Yay! :D

PR #4295, opened 2020-12-01
LibAudio: Change the API to allow for multiple file formats

metalvoidzz:
This patch adds a generic interface for loading any kind of supported audio file format, instead of just WAV.

### DIFF
--- a/Libraries/LibAudio/Loader.cpp
+++ b/Libraries/LibAudio/Loader.cpp
@@ -1,0 +1,39 @@
+/*
+ * Copyright (c) 2018-2020, the SerenityOS developers.
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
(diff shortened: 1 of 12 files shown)


Libraries/LibAudio/WavLoader.cpp
```diff
@@ -33,23 +33,29 @@
...
-RefPtr<Buffer> WavLoader::get_more_samples(size_t max_bytes_to_read_from_input)
+bool WavLoaderPlugin::sniff()
 {
-#ifdef AWAVLOADER_DEBUG
+    return valid;
+}
+
+RefPtr<Buffer> WavLoaderPlugin::get_more_samples(size_t max_bytes_to_read_from_input)
+{
+#ifdef AWavLoaderPlugin_DEBUG
```
awesomekling, line 58: Please don't use lowercase letters in macros. Also, if you change a debug macro, you need to update CMakeLists.txt (see the CI failure.)

Libraries/LibAudio/Loader.cpp
```diff
@@ -0,0 +1,39 @@
...
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include <LibAudio/WavLoader.h>
+
+namespace Audio {
+
+Loader::Loader(const StringView& path)
```
awesomekling, line 31: At some point we should also support input not coming from a file on disk. (To support streaming etc.)

Libraries/LibAudio/WavLoader.cpp
metalvoidzz: Fixed it, sorry for that.

Libraries/LibAudio/Loader.cpp
metalvoidzz: You're right, I've added the option to load from a ByteBuffer.

awesomekling:
👍🏻

PR #4302, opened 2020-12-02
Games: Added level select and aditional level to Breakout

cooljohnny3:
Added a level select menu which is shown when the game start where
either of the 2 available levels may be selected.


awesomekling:
Very rad :)

PR #4311, opened 2020-12-02
LibAudio+SoundPlayer: Fix loading errors

metalvoidzz:
SoundPlayer will now accept dropped files and handle loading errors a bit better. Also, I screwed up in my last PR, WAV files can now be parsed correctly when reading from memory :)


danboid:
Pretty please add FLAC support! :unicorn:

PR #4316, opened 2020-12-03
Shell: "variable substitutions" (but actually readable)

alimpfard:
TL;DR:
```sh
# bashy:
${#foo[@]}
# better:
${count $foo}

# bashy
${foo%.txt}
# better:
${remove_suffix .txt $foo}
```

(these are also a little peculiar, as they can rewrite the AST as opposed to just returning new values)

### DIFF
--- a/Base/usr/share/man/man5/Shell.md
+++ b/Base/usr/share/man/man5/Shell.md
@@ -173,6 +173,44 @@ test -f foo.txt || touch foo.txt
 rm test && echo "deleted!" || echo "failed with $?"
 ```
 
+### Immediate Functions
+A series of functions provided by the shell that can operate on AST nodes (as opposed to string arguments).
+Such functions are only available in an _immediate function_ context, i.e. inside `${...}`.
+
+Functions in this context are only allowed to be named as barewords, no escaping or string parts are allowed in such function names.
+
+The following functions are currently implemented:
+| Name          | Effect |
+| :-            | :----- |
+| `count`       | Returns the counts of its input arguments (as lists) in a list |
+| `length`      | Returns the length of its input arguments (as strings) in a list |
+| `nth`         | Returns the nth elements of its input arguments (n = first argument) |
+| `substring`   | Returns the substrings starting at arg0 with length arg1 of its inputs (as strings) |
+| `slice`       | Returns the list slice starting at arg0 with length arg1 of its inputs (as lists) |
+| `remove_prefix`| Returns its input arguments, without the given prefix (arg0) - (i.e. strips the given prefix) |
+| `remove_suffix`| Similar to `remove_prefix`, but strips suffixes instead |
+| `regex_replace`  | Replaces all substrings matching the first arg using the second arg |
+| `filter_glob`  | Returns the items matching the given glob (arg0) |
+
+##### Examples
+
+```sh
+x=(1 2 3 4 5)
+# Get the count of elements in a list
+echo ${count $x}
+# => 5
+
+# Get the string length of the third element in the list
+echo ${length ${nth 2 $x}}
+# => 1
+
+# Rename .sh files to .txt files
+for *.sh {
+    mv $it ${removeSuffix .sh $it}.txt
+}
+```
+
+
 #### Control Structures
 
 ##### Conditionals
@@ -351,6 +389,7 @@ redirection :: number? '>'{1,2} ' '* string_composite
 list_expression :: ' '* expression (' '+ list_expression)?
 
 expression :: evaluate expression?
+            | immediate_brace_expression expression?
             | string_composite expression?
             | comment expression?
             | '(' list_expression ')' expression?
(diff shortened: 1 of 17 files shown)


AK/MemoryStream.h
```diff
@@ -222,55 +222,29 @@ class DuplexMemoryStream final : public DuplexStream {
...
-
-            return (chunk_index - first_chunk_index) * chunk_size + last_chunk_offset - first_chunk_offset;
+            bytes_to_search -= span.size();
+            spans.append(move(span));
+            ++chunk_index;
         }
 
-        // No dice.
-        return {};
+        return memmem(spans.begin(), spans.end(), value);
```
alimpfard, line 247: cc @asynts (this and the previous commit)

Shell/Shell.cpp
```diff
@@ -758,12 +766,12 @@ RefPtr<Job> Shell::run_command(const AST::Command& command)
         if (!m_is_subshell && command.should_wait)
             tcsetattr(0, TCSANOW, &default_termios);
 
-        Core::EventLoop mainloop;
-        setup_signals();
```
alimpfard, line 755: This works around #4326 somewhat, but the issue with LibCore remains.

Shell/Tests/loop.sh
```diff
@@ -42,9 +46,32 @@ for ((test 1 = 1) (test 2 = 2)) {
     $it || unset implicit_ok
 }
 
+# Infinite loop
+for {
+    infinite_ok=yes
+    break
+    unset break_ok
+}
```
alimpfard, line 54: I'm not 100% on this, `for { ... }` sounds...dumb...
but adding another keyword for almost zero gain sounds even more dumb - and I am therefore conflicted.
Ideas, anyone?

AK/MemoryStream.h
```diff
@@ -252,7 +252,7 @@ class DuplexMemoryStream final : public DuplexStream {
         size_t nread = 0;
         while (bytes.size() - nread > 0 && m_write_offset - m_read_offset - nread > 0) {
             const auto chunk_index = (m_read_offset - m_base_offset + nread) / chunk_size;
-            const auto chunk_bytes = m_chunks[chunk_index].bytes().slice(m_read_offset % chunk_size).trim(m_write_offset - m_read_offset - nread);
+            const auto chunk_bytes = m_chunks[chunk_index].bytes().slice((m_read_offset + nread) % chunk_size).trim(m_write_offset - m_read_offset - nread);
```
alimpfard, line 255: Also cc @asynts 
Test case is...not minimal: https://github.com/alimpfard/aoc2020-shell/blob/1036d3833b601401ae4848ffbcbb59700dd21346/4.sh#L76

Shell/Tests/loop.sh
awesomekling: \<bugaevc\> how about `loop { }` \</bugaevc\>

Base/usr/share/man/man5/Shell.md
```diff
@@ -173,6 +173,43 @@ test -f foo.txt || touch foo.txt
...
+| Name          | Effect |
+| :-            | :----- |
+| `count`       | Returns the counts of its input arguments (as lists) in a list |
+| `length`      | Returns the length of its input arguments (as strings) in a list |
+| `nth`         | Returns the nth elements of its input arguments (n = first argument) |
+| `substring`   | Returns the substrings starting at arg0 with length arg1 of its inputs (as strings) |
+| `slice`       | Returns the list slice starting at arg0 with length arg1 of its inputs (as lists) |
+| `removePrefix`| Returns its input arguments, without the given prefix (arg0) - (i.e. strips the given prefix) |
+| `removeSuffix`| Similar to `removePrefix`, but strips suffixes instead |
+| `filterGlob`  | Returns the items matching the given glob (arg0) |
```
awesomekling: Let's not introduce camelCase in something we have full control over.

alimpfard: Luckily these are still experimental.
`removeprefix` feels dumb, `remove_prefix` looks wrong in a shell, `remove-prefix` looks ok, `RemovePrefix` is too C# to exist, `rEmOvEpReFiX` is perfection (/s)
should we just go with kebab-case?

Shell/Tests/loop.sh
alimpfard: We discussed this in IRC, actually
I personally prefer `loop` too, but an extra keyword just for an infinite loop doesn't sound convincing.
Though I suppose we could use `loop` in place of `while`? I'm also not that interested in `while`, it seems to always be used for nefarious purposes (`whatever | while read...`  which could be better expressed as `for $(whatever) { match $it ... }` here to the same effect)

AK/MemoryStream.h
asynts: When I suggested that iterator thing, I thought that would be really messy with backtracking but this a really elegant solution.

AK/MemoryStream.h
asynts: I have trouble wrapping my head around all of these offset calculations but I think you are right. If `m_read_offset % chunk_size` is not zero and a second chunk is read, it would be sliced at the wrong offset.

This should be a test case, but for some reason `stream << 1 << input << 1` hangs, seems like it's totally broken. I'll debug that tomorrow.

~~~c++
TEST_CASE(offset_calculation_error_regression)
{
    [[maybe_unused]] Array<u8, DuplexMemoryStream::chunk_size> input, output;
    input.span().fill(0);

    DuplexMemoryStream stream;
    stream << 1 << input << 1;

    stream.discard_or_error(4);
    stream.read(output);

    EXPECT(compare(input, output));
}
~~~

Shell/Tests/loop.sh
alimpfard: `loop` it is.

AK/MemoryStream.h
asynts: (This was resolved in #4371.)

alimpfard: Oh yeah, silly me :)

PR #4317, opened 2020-12-03
LibVT: Auto-scroll the terminal

### LINKED ISSUE #4267: Terminal: Cannot select more than a windowful of text
It is not currently possible to click and drag to select and highlight more text than can be displayed in the terminal window without scrolling. The terminal should scroll to allow you to select more than a full terminal window of text.

metalvoidzz:
Fixes #4267

### DIFF
--- a/Libraries/LibVT/TerminalWidget.cpp
+++ b/Libraries/LibVT/TerminalWidget.cpp
@@ -101,6 +101,7 @@ TerminalWidget::TerminalWidget(int ptm_fd, bool automatic_size_policy, RefPtr<Co
     set_pty_master_fd(ptm_fd);
     m_cursor_blink_timer = add<Core::Timer>();
     m_visual_beep_timer = add<Core::Timer>();
+    m_auto_scroll_timer = add<Core::Timer>();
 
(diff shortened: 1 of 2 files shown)


danboid:
This'll be a pretty major upgrade to the terminal!

Thanks @metalvoidzz !

supex0fan:
I'm not a reviewer, but I was looking at this issue my self and didn't get around to it. That being said I think this issue should be resolved at the LibGUI level as almost every scrollable application should scroll when the cursor is dragged behind the window.

danboid:
@supex0fan That makes sense. TextEditor supports autoscrolling but currently Terminal and Browser don't. TE would prob need to be adapted if this is becoming a LibGUI thing.

@awesomekling

awesomekling:
>I'm not a reviewer

Everyone can review code :)

>I think this issue should be resolved at the LibGUI level as almost every scrollable application should scroll when the cursor is dragged behind the window.

Right, it would be nice to have this functionality in **GUI::ScrollableWidget** (and make **VT::TerminalWidget** inherit it from there.)

It's okay to only do this for the terminal in this PR though, we don't have to make everything perfect right away.

PR #4318, opened 2020-12-03
LibWeb: Improve contenteditable attribute.

asynts:
There doesn't seem to be a specification that dictates how exactly the browser should behave in the presence of "contenteditable".

I've tried to make the behavior feel intuitive. There are some edge cases where no behavior feels "intuitive". Here a list with a few nasty edge cases:

---

The user selects the text starting and ending with the uppercase letter and then presses backspace:

~~~html
<div contenteditable>
    <p>aBc</p>
    <b>cDe</b>
</div>
~~~

The user would expect the remainders of the bold text to stay bold.

Firefox & Chromium:
~~~html
<div contenteditable>
    <p>a<b>e</b></p>
</div>
~~~

Serenity:
~~~html
<div contenteditable>
    <p>ae</p>
</div>
~~~

---

The user selects the text starting and ending with the uppercase letter and then presses backspace:

~~~html
<div contenteditable>
    <p class=a>aBc</p>
    <p class=b>cDe</p>
</div>
~~~

The user would expect both paragraphs to be joined into one but what happens to the classes?

Firefox:
~~~html
<div contenteditable>
    <p class=b>ae</p>
</div>
~~~

Serenity & Chromium:
~~~html
<div contenteditable>
    <p class=a>ae</p>
</div>
~~~

### DIFF
--- a/Libraries/LibWeb/Bindings/RangeConstructor.cpp
+++ b/Libraries/LibWeb/Bindings/RangeConstructor.cpp
@@ -24,52 +24,35 @@
  * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
-#include <LibWeb/DOM/Document.h>
-#include <LibWeb/DOM/Node.h>
+#include <LibJS/Heap/Heap.h>
+#include <LibWeb/Bindings/RangeConstructor.h>
+#include <LibWeb/Bindings/RangeWrapper.h>
+#include <LibWeb/Bindings/WindowObject.h>
 #include <LibWeb/DOM/Range.h>
-#include <LibWeb/DOM/Window.h>
 
-namespace Web::DOM {
+namespace Web::Bindings {
 
-Range::Range(Window& window)
-    : m_start_container(window.document())
-    , m_start_offset(0)
-    , m_end_container(window.document())
-    , m_end_offset(0)
+RangeConstructor::RangeConstructor(JS::GlobalObject& global_object)
+    : NativeFunction(*global_object.function_prototype())
 {
 }
 
-Range::Range(Node& start_container, size_t start_offset, Node& end_container, size_t end_offset)
-    : m_start_container(start_container)
-    , m_start_offset(start_offset)
-    , m_end_container(end_container)
-    , m_end_offset(end_offset)
+void RangeConstructor::initialize(JS::GlobalObject& global_object)
 {
+    NativeFunction::initialize(global_object);
+
+    define_property("length", JS::Value(0), JS::Attribute::Configurable);
 }
 
-NonnullRefPtr<Range> Range::clone_range() const
+JS::Value RangeConstructor::call()
 {
-    return adopt(*new Range(const_cast<Node&>(*m_start_container), m_start_offset, const_cast<Node&>(*m_end_container), m_end_offset));
+    return construct(*this);
 }
 
-NonnullRefPtr<Range> Range::inverted() const
+JS::Value RangeConstructor::construct(Function&)
 {
-    return adopt(*new Range(const_cast<Node&>(*m_end_container), m_end_offset, const_cast<Node&>(*m_start_container), m_start_offset));
-}
-
-NonnullRefPtr<Range> Range::normalized() const
-{
-    if (m_start_container.ptr() == m_end_container.ptr()) {
-        if (m_start_offset <= m_end_offset)
-            return clone_range();
-
-        return inverted();
-    }
-
-    if (m_start_container->is_before(m_end_container))
-        return clone_range();
-
-    return inverted();
+    auto& window = static_cast<WindowObject&>(global_object());
+    return heap().allocate<RangeWrapper>(window, window, DOM::Range::create(window.impl()));
 }
 
 }
--- a/Libraries/LibWeb/Bindings/RangeConstructor.cpp
+++ b/Libraries/LibWeb/Bindings/RangeConstructor.cpp
@@ -26,6 +26,7 @@
 
 #include <LibJS/Heap/Heap.h>
 #include <LibWeb/Bindings/RangeConstructor.h>
+#include <LibWeb/Bindings/RangePrototype.h>
 #include <LibWeb/Bindings/RangeWrapper.h>
 #include <LibWeb/Bindings/WindowObject.h>
 #include <LibWeb/DOM/Range.h>
@@ -39,14 +40,17 @@ RangeConstructor::RangeConstructor(JS::GlobalObject& global_object)
 
 void RangeConstructor::initialize(JS::GlobalObject& global_object)
 {
+    auto& vm = this->vm();
     NativeFunction::initialize(global_object);
-
-    define_property("length", JS::Value(0), JS::Attribute::Configurable);
+    auto& window = static_cast<WindowObject&>(global_object);
+    define_property(vm.names.prototype, window.range_prototype(), 0);
+    define_property(vm.names.length, JS::Value(0), JS::Attribute::Configurable);
 }
 
 JS::Value RangeConstructor::call()
 {
-    return construct(*this);
+    vm().throw_exception<JS::TypeError>(global_object(), JS::ErrorType::ConstructorWithoutNew, "Range");
+    return {};
 }
 
 JS::Value RangeConstructor::construct(Function&)
--- a/Libraries/LibWeb/Bindings/RangePrototype.cpp
+++ b/Libraries/LibWeb/Bindings/RangePrototype.cpp
@@ -80,8 +80,8 @@ JS_DEFINE_NATIVE_FUNCTION(RangePrototype::set_start)
     if (vm.exception())
         return {};
 
-    if (!static_cast<NodeWrapper*>(arg0)->is_node_wrapper()) {
-        vm.throw_exception<JS::TypeError>(global_object, JS::ErrorType::NotA, "Range");
+    if (!arg0->is_node_wrapper()) {
+        vm.throw_exception<JS::TypeError>(global_object, JS::ErrorType::NotA, "Node");
         return {};
     }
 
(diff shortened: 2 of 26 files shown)


asynts:
I now see that I didn't do a good job explaining what this pull request does.

It makes it possible to delete selected text even if it goes across multiple nodes in the tree. It's also possible to navigate the cursor inside of text nodes with the arrow keys and delete with backspace or delete.

All of these operations are moved behind the `EditEventHandler` class which can be overwritten when I implement the writer application.

Libraries/LibWeb/Page/EditEventHandler.cpp
```diff
@@ -120,9 +100,12 @@ void EditEventHandler::handle_insert(DOM::Position position, u32 code_point)
         builder.append(node.data().substring_view(position.offset()));
         node.set_data(builder.to_string());
 
-        m_frame.cursor_position().set_offset(m_frame.cursor_position().offset() + 1);
         node.invalidate_style();
     }
+
+    // FIXME: We need to remove stale layout nodes when nodes are removed from the DOM. Currently,
+    //        this is the only way to get these to disappear.
```
awesomekling: This is the correct thing to do IMO. Trying to get fancy with this will lead to unpleasant levels of complexity.

Libraries/LibWeb/DOM/Position.h
```diff
@@ -62,6 +62,38 @@ class Position {
     unsigned m_offset { 0 };
 };
 
+class Range {
```
awesomekling, line 65: Let's make DOM::Range follow the specced Range type as closely as possible: https://developer.mozilla.org/en-US/docs/Web/API/Range :)

(Also note that it needs to be reference-counted since it will be exposed to JavaScript.)

Libraries/LibWeb/DOM/Position.h
```diff
@@ -62,6 +62,38 @@ class Position {
...
+    const Position& start() const { return m_start; }
+    Position& start() { return m_start; }
+
+    const Position& end() const { return m_end; }
+    Position& end() { return m_end; }
+
+    Range normalized() const;
+
+private:
+    Position m_start, m_end;
```
awesomekling, line 94: Please only declare one variable per line.

awesomekling:
Great work and I'm really glad you're taking this on!

Let's make sure we build this in the most sustainable and synergistic way possible by modeling things around specced Web APIs like [Range](https://dom.spec.whatwg.org/#range) and [Selection](https://w3c.github.io/selection-api/#selection-interface) :)

awesomekling:
Also, perhaps we could get rid of layout positions/ranges entirely and only use DOM positions/ranges. I'm not sure having both is really necessary since this state should always be representable in DOM terms anyway. (But maybe I'm missing something.)

Libraries/LibWeb/Page/EditEventHandler.cpp
asynts: This is on the entire document, not just that sub-tree. In my understanding, the correct behavior of `node.invalidate_style` would be to recompute the layout for that sub-tree at the next opportune moment.

asynts:
I've exposed `DOM::Range` to JavaScript that was an interesting learning experience.

I also started on a `DOM::Selection` class but that just got me frustrated. The API is *terrible* and it just wouldn't work. I got something that's almost working at abe4e23199e76ff705f407638df1d5901030c279 but I got enough of that for now. (That should work but the selection object isn't updated for some reason.)

Also it's not trivially possible to remove `LayoutRange` because it's used for the layout process, I didn't look into it too much, so it might be possible to just rewrite it a bit, but again, the whole selection thing just really got me frustrated.

awesomekling:
Please fix up the commit message for "Do the range thing with idl" :)

Libraries/LibWeb/Bindings/RangeConstructor.cpp
```diff
@@ -0,0 +1,58 @@
...
+void RangeConstructor::initialize(JS::GlobalObject& global_object)
+{
+    NativeFunction::initialize(global_object);
+
+    define_property("length", JS::Value(0), JS::Attribute::Configurable);
+}
+
+JS::Value RangeConstructor::call()
+{
+    return construct(*this);
```
linusg, line 49: I don't think `Range` is supposed to support construction without `new`.

```suggestion
    vm().throw_exception<JS::TypeError>(global_object(), JS::ErrorType::ConstructorWithoutNew, "Range");
    return {};
```

asynts:
Oops, I wanted to remove that commit during rebasing.

Libraries/LibWeb/Bindings/RangeConstructor.cpp
asynts: I just took that from here:

https://github.com/SerenityOS/serenity/blob/48d2545572f1281bd9526f64798e51ecc862bb93/Libraries/LibWeb/Bindings/XMLHttpRequestConstructor.cpp#L60

If I understand you correctly, the JavaScript engine calls `RangeConstructor::construct` when the user writes `new Range` and calls `RangeConstructor::call` if the user calls it directly somehow?

Lubrsi: Apparently it can be constructed with `new`, but it can't be called as a function:
Chrome:
chrome_BQeRmSXDsW

Firefox:
firefox_ujKusY64bR

linusg: Precisely :)

- `Foo()` -> `::call()`
- `new Foo()` -> `::construct()`

In some cases the desired behaviour is them being equal, that's why some objects simply call `construct` in their `call`.

Looks like XHR shouldn't support construction without `new` either though :smile:

asynts: Oh, that's what you mean, yes, it can be constructed with `new Range()` that just selects the entire document.

https://dom.spec.whatwg.org/#dom-range-range

Libraries/LibWeb/Bindings/RangePrototype.cpp
```diff
@@ -0,0 +1,161 @@
...
+    if (vm.exception())
+        return {};
+    auto arg1 = vm.argument(1).to_number(global_object);
+    if (vm.exception())
+        return {};
+
+    if (!static_cast<NodeWrapper*>(arg0)->is_node_wrapper()) {
+        vm.throw_exception<JS::TypeError>(global_object, JS::ErrorType::NotA, "Range");
+        return {};
+    }
```
linusg, line 86: This seems incorrect... (and crashes, when passing in a non-NodeWrapper object). Also the error message should probably say something like NotA "NodeWrapper", as it's not expecting a Range object?

Libraries/LibWeb/Bindings/RangePrototype.cpp
```diff
@@ -0,0 +1,161 @@
...
```
linusg, line 109: Same as above

Libraries/LibWeb/Bindings/RangePrototype.cpp
asynts: Oops. `static_cast<NodeWrapper*>(arg0)->is_node_wrapper()` is always true, because if it's not a node wrapper will still call the implementation of `NodeWrapper`.

awesomekling: This is a great reason to make all final "return true" override variants of `is_foo()` private :)

asynts: It now throws an exception and I updated the code generator to make the `is_X_wrapper` methods private.

Libraries/LibWeb/Bindings/RangeConstructor.cpp
asynts: Now `Range()` throws an exception while `new Range()` works normally.

asynts:
Oh, I just noticed that `new Range` puts `{}` into the console instead of `{"startOffset": 0, ...}`. Any idea why this happens?

linusg:
> Oh, I just noticed that new Range() puts {} into the console instead of {"startOffset": 0, ...}. Any idea why this happens?

Yeah, that's a bit unfortunate - since all the magic happens in `RangePrototype`, the `RangeWrapper` object allocated by the `Range` constructor doesn't have any "own properties", they're all inherited from its prototype - so the value printing functionality considers it to be "empty".

Same applies to XHR, let's fix this separately.

Libraries/LibWeb/Bindings/WindowObject.cpp
```diff
@@ -88,6 +90,11 @@ void WindowObject::initialize()
     m_xhr_constructor = heap().allocate<XMLHttpRequestConstructor>(*this, *this);
     m_xhr_constructor->define_property("prototype", m_xhr_prototype, 0);
     add_constructor("XMLHttpRequest", m_xhr_constructor, m_xhr_prototype);
+
+    m_range_prototype = heap().allocate<RangePrototype>(*this, *this);
+    m_range_constructor = heap().allocate<RangeConstructor>(*this, *this);
+    m_range_constructor->define_property("prototype", m_range_prototype);
+    add_constructor("Range", m_range_constructor, m_range_prototype);
```
linusg, line 97: Sorry, this needs to be updated to avoid the issue fixed in https://github.com/SerenityOS/serenity/pull/4364. You can follow the exact same pattern though :)

Also I noticed that `void WindowObject::visit_edges(Visitor& visitor)` needs to be updated to include `m_range_{constructor,prototype}`.

asynts: No idea what that `visit_edges` method is for, but I just added two lines like the XML thing.

asynts:
> [...] the `RangeWrapper` object allocated by the `RangeConstructor` doesn't have any "own properties", they're all inherited from its prototype [...]

My understanding is that they should still be displayed? Isn't that the point of prototype inheritance?

Libraries/LibWeb/Bindings/WindowObject.cpp
linusg: It's to prevent things from being garbage collected :)

linusg:
> My understanding is that they should still be displayed?

Should, yes, but the current implementations simply doesn't care about prototypes, only the object's (or, object shape's) own properties.

https://github.com/SerenityOS/serenity/blob/d2a2d19a86f64a36b3adfe59f78a91de5d557898/Libraries%2FLibJS%2FMarkupGenerator.cpp#L117-L145

PR #4319, opened 2020-12-04
Userland: Add an implementation of 'entr'

alimpfard:
This is like inotifywait, but it runs commands instead of printing
events out.

cc @nico - you suggested this; how would you suggest I word the commit?

### DIFF
--- a/Userland/entr.cpp
+++ b/Userland/entr.cpp
@@ -1,0 +1,191 @@
+/*
+ * Copyright (c) 2020, the SerenityOS developers.
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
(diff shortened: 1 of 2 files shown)


nico:
Hm, I'd probably put in an example,

```
To run a shell script every time it's saved:
  echo foo.sh | entr 'Shell foo.sh'

This is similar to
  while inotifywait -e modify -r foo.sh; do Shell foo.sh; done

but that misses events that happen during the body of the while loop. That can be fixed with

  inotifywait -m -e modify | while read dir action file; do Shell foo.sh; done

but that's harder to remember than the `entr` version.
```

Actually, maybe throw in a man page with some examples? (Or, if you don't like writing man pages, I can write on once this is in.)

Userland/entr.cpp
```diff
@@ -0,0 +1,203 @@
...
+                mutable_old_pid.clear();
+            }
+
+            Vector<const char*> argv;
+            argv.append(executable_name);
+            for (auto arg : str_arguments)
+                argv.append(arg.characters());
+            argv.append(nullptr);
+
+            posix_spawn_file_actions_t action;
```
nico, line 152: Doesn't look like you do anything with this other than create and destroy it -- can you just pass nullptr for the action instead?

alimpfard:
Since you're offering, go ahead and write the manpage :P 
good idea with the example in the commit, I shall do that

Userland/entr.cpp
```diff
@@ -0,0 +1,203 @@
...
+            started_job_ids.append(pid);
+            mutable_old_pid = pid;
+            if (verbose)
+                warnln("[exec] created instance for {} with pid={}", path, pid);
+        };
+
+        notifiers.empend(move(notifier));
+    }
+
+    Core::EventLoop::register_signal(SIGINT, [&](auto) {
```
nico, line 169: Out of interest, why is this needed? Shouldn't the default SIGINT handler do something similar?

Userland/entr.cpp
```diff
@@ -0,0 +1,203 @@
...
+            notifier.set_enabled(false);
+            close(notifier.fd());
+            notifier.close();
+        }
+        _exit(0);
+    });
+
+    Core::EventLoop::register_signal(SIGCHLD, [&](auto) {
+        Vector<pid_t> new_pids;
+        for (auto& pid : started_job_ids) {
```
nico, line 180: I think this handles coalesced SIGCHLDs correctly, but it feels "subtly correct" to me :) (No concrete suggestion here.)

Userland/entr.cpp
alimpfard: can I? I'm not too familiar with `posix_spawn()`, its manpage always talks about "the object pointed to by `attrp`" so I don't think it can be null?

Userland/entr.cpp
alimpfard: I guess, does it hurt to be sure?

Userland/entr.cpp
nico: Both Serenity's man page and https://pubs.opengroup.org/onlinepubs/009695399/functions/posix_spawn.html are fairly explicit that it can be nullptr.

Userland/entr.cpp
nico: If process teardown does this for you (and if it didn't, wouldn't it be a bug), why do it manually?

Userland/entr.cpp
alimpfard: I is bad at reading manpages, apparently.

awesomekling:
I can't understand this commit message, as it assumes I know exactly how inotifywait works. Which I don't :)

alimpfard:
oh this PR is still open? ^^'
let's close it for now, I remember finding some big holes in this idea and I don't feel like messing with it right now.

Userland/entr.cpp
tomuta: I know it's closed now, but for future reference: `WNOWAIT` is essentially just polling the information without un-waiting the child. So I'd expect zombies to hang around.

alimpfard: L142 should take care of that though?
IIRC the main problem I saw with this was the fact that it wouldn't notice a change while it's running something

tomuta: Possible, too complex logic for me to to comprehend ;) Also, I believe `WNOWAIT` will block if the process doesn't have waitable information (e.g. if it's still running) unless `WNOHANG` is specified. Not sure if that's intended since it's called in a loop.

alimpfard: Ah whatever, the future me that decides to reopen this PR will think about it :shrug:

Issue #3429, opened 2020-09-07
Clock loses 1.3-3.5 seconds per hour


nico:
Now that we can set time via `ntpquery -s` fairly accurately (it's a one-shot adjustment), we can evaluate how well serenity's clock keeps time.

I ran `su; ntpquery -s; for i in $(seq 1000) { sleep 1; echo $i; ntpquery }` to see how much the clock gains over 1000 seconds, and it gained 3.5s. Then I tried `su; ntpquery -s; for i in $(seq 12) { sleep 1800; echo $i; ntpquery }` to check every 30 min over 6 hours. At the end of the 6h, the clock was 13s late.

After 20h of uptime, I checked again, by then it was 32.8s late.

Eventually we'll have a long-running NTP service that keeps adjusting the clock, but maybe the OS should be able to keep time better even without an internet connection.

--

I read the timekeeping code a bit, and I think we use the HPETComparator code, which should tick at 1000 Hz even. TimeManagement::epoch_time() assumes 1000Hz exactly instead of using `ticks_per_second()` but at least for the HPET that's not a source of error. So not sure where exactly we lose time, but I figured I'd file at least a bug for it.

nico:
Oh, I should paste the raw data from my 6h experiment described above:

<details>

```
1
Offset: 1.256666
2
Offset: 2.393099
3
Offset: 3.388525
4
Offset: 4.455085
5
Offset: 5.387008
6
Offset: 6.299331
7
Offset: 7.236079
8
Offset: 8.239125
9
Offset: 9.368290
10
Offset: 10.399584
11
Offset: 11.694160
12
Offset: 13.049532

# later...
uptime
Up 20 hours, 3 minutes, 6 seconds
Offset: 32.787009
```
</details>

nico:
Now at `20 hours, 56 minutes, 11 seconds` uptime, at an offset of 42.54s -- so lost almost 10s in less than an hour. The system probably loses more time when it's getting used -- maybe we drop interrupts or something like that.

tomuta:
I don't think the "1000 Hz even" will be exactly that. All clock chips are subject to variation, differences due to manufacturing, quality, and temperature. Isn't that precisely the point of NTP so that the kernel can be periodically supplied with frequency adjustments to keep it roughly at the same rate as actual wall time?

nico:
Sure, but 1-10s per hour seems still kind of bad maybe? That's worse than even a mechanical watch.

tomuta:
[...]

PR #4320, opened 2020-12-04
Kernel: Improve time keeping and dramatically reduce interrupt load

tomuta:
This implements a number of changes related to time:
* If a HPET is present, it is now used only as a system timer, unless
  the Local APIC timer is used (in which case the HPET timer will not
  trigger any interrupts at all).
* If a HPET is present, the current time can now be as accurate as the
  chip can be, independently from the system timer. We now query the
  HPET main counter for the current time in CPU #0's system timer
  interrupt, and use that as a base line. If a high precision time is
  queried, that base line is used in combination with quering the HPET
  timer directly, which should give a much more accurate time stamp at
  the expense of more overhead. For faster time stamps, the more coarse
  value based on the last interrupt will be returned. This also means
  that any missed interrupts should not cause the time to drift.
* The default system interrupt rate is reduced to about 250 per second.
* Fix calculation of Thread CPU usage by using the amount of ticks they
  used rather than the number of times a context switch happened.
* Implement CLOCK_REALTIME_COARSE and CLOCK_MONOTONIC_COARSE and use it
  for most cases where precise timestamps are not needed.

For giggles, I tried changing `OPTIMAL_TICKS_PER_SECOND_RATE` to `100` and the system actually works just fine, but I figure `250` is what Linux uses by default, so that's probably a good default. Changing the way CPU usage is calulated became quite apparent the lower I dropped that number, because the perceived cpu usage went way up. With these changes it remains consistent regardless of whether we're running with `100`, `250`, or `1000` per second.

This might also help with #3429. At least theoretically this is now as accurate as HPET can be, and even missed interrupts should not cause any time drift.

### DIFF
--- a/Applications/SystemMonitor/ProcessModel.cpp
+++ b/Applications/SystemMonitor/ProcessModel.cpp
@@ -355,12 +355,12 @@ void ProcessModel::update()
     auto previous_pid_count = m_pids.size();
     auto all_processes = Core::ProcessStatisticsReader::get_all();
 
-    unsigned last_sum_times_scheduled = 0;
+    u64 last_sum_ticks_scheduled = 0;
     for (auto& it : m_threads)
-        last_sum_times_scheduled += it.value->current_state.times_scheduled;
+        last_sum_ticks_scheduled += it.value->current_state.ticks_user + it.value->current_state.ticks_kernel;
 
     HashTable<PidAndTid> live_pids;
-    unsigned sum_times_scheduled = 0;
+    u64 sum_ticks_scheduled = 0;
     for (auto& it : all_processes) {
         for (auto& thread : it.value.threads) {
             ThreadState state;
@@ -393,12 +393,14 @@ void ProcessModel::update()
             state.pgid = it.value.pgid;
             state.sid = it.value.sid;
             state.times_scheduled = thread.times_scheduled;
+            state.ticks_user = thread.ticks_user;
+            state.ticks_kernel = thread.ticks_kernel;
             state.cpu = thread.cpu;
             state.cpu_percent = 0;
             state.priority = thread.priority;
             state.effective_priority = thread.effective_priority;
             state.state = thread.state;
-            sum_times_scheduled += thread.times_scheduled;
+            sum_ticks_scheduled += thread.ticks_user + thread.ticks_kernel;
             {
                 auto pit = m_threads.find({ it.value.pid, thread.tid });
                 if (pit == m_threads.end())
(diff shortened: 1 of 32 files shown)


danboid:
Have you tested this with nesalizer or quake?

awesomekling:
This makes sound playback in the nesalizer port noticeably worse. Have not looked at quake.

danboid:
Gah! Thanks for checking AK!

tomuta:
Ah interesting. I wonder if it's because `CLOCK_MONOTONIC`/`CLOCK_REALTIME` by default return a coarse timestamp rather than a precise one. It's very easy to add the `CLOCK_MONOTONIC_COARSE`/`CLOCK_REALTIME_COARSE` variants but unless we change other places that heavily use timestamps (most of them do not need to be veryprecise) it does increase the system load...

tomuta:
I can replicate it with nesalizer. But it was kind of glitchy even without these changes, just a little less. So it looks like our audio stuff is in dire need of some upgrades. For example, there is no buffer. Normally, with e.g. ALSA one would configure a number of periods (e.g. dma transfers) and then also a playback trigger threshold of n number of periods. For example, a buffer size of 8 periods and a playback threshold of 1 period would allow user space to fill up to 8 periods with audio samples, but once it wrote 1 full period the driver triggers playback (dma transfers). I think this is what causes these glitches.

tomuta:
FWIW the reduced interrupt load seems to improve overall performance slightly, e.g. Browser shows about ~10% less time on the initial page (for me ~550 ms down to ~500 ms on average) with Qemu kvm disabled and `smp=on`.

tomuta:
I guess it's a good time to implement an audio device stack based on `AsyncDeviceRequest`... not sure how long it'll take me...

danboid:
You da man Tom!

awesomekling:
Still bad audio on NES fwiw :(

tomuta:
> Still bad audio on NES fwiw :(

Well, quite frankly the SB16 driver isn't working very well even without these changes. It's quite glitchy for me right now on `master` currently, even with just the Piano application. The way the SB16 driver is implemented it cannot produce smooth audio. I have the audio PR that is meant to address these glitches, but I haven't had time to look into why the dma transfer locks up the system.

But, how do you build nesalizer now? Mine is still failing to link, see #4457

awesomekling:
> > Still bad audio on NES fwiw :(
> 
> Well, quite frankly the SB16 driver isn't working very well even without these changes.

Touché. You are right of course :)

> But, how do you build nesalizer now? Mine is still failing to link, see #4457

I tweaked the nesalizer makefile locally by adding `-rpath-link` to the link arguments. Hoping that the upcoming GCC fixes from @itamar8910 makes that sort of hack unnecessary.

awesomekling:
Oops, sorry about the little conflict. With that resolved we can merge this :)

PR #4321, opened 2020-12-04
Added resign, flip board and export game as PGN to Chess game

AnicJov:
Started to add some features I wanted to the Chess game. This is just the start, I'm planing on adding a lot more in the future.
Hope that everything is okay, I'm willing to take any recommendations/advice and change the code accordingly if needed.

### DIFF
--- a/Games/Chess/ChessWidget.cpp
+++ b/Games/Chess/ChessWidget.cpp
@@ -288,3 +288,24 @@ void ChessWidget::maybe_input_engine_move()
         update();
     });
 }
+
+void ChessWidget::flip_board()
+{
+    m_side = Chess::opposing_colour(m_side);
+    update();
+}
+
+void ChessWidget::resign()
+{
+    if (m_engine && m_board.turn() != m_side) {
+        GUI::MessageBox::show(window(), "You can only resign on your turn.", "Resign", GUI::MessageBox::Type::Information);
+        return;
+    }
+
+    board().set_resigned(m_board.turn());
+
+    set_drag_enabled(false);
+    update();
+    const String msg = m_board.result_to_string(m_board.game_result());
+    GUI::MessageBox::show(window(), msg, "Game Over", GUI::MessageBox::Type::Information);
+}
--- a/Games/Chess/ChessWidget.cpp
+++ b/Games/Chess/ChessWidget.cpp
@@ -27,9 +27,12 @@
 #include "ChessWidget.h"
 #include "PromotionDialog.h"
 #include <AK/String.h>
+#include <LibCore/DateTime.h>
+#include <LibCore/File.h>
 #include <LibGUI/MessageBox.h>
 #include <LibGUI/Painter.h>
 #include <LibGfx/Font.h>
+#include <unistd.h>
 
 ChessWidget::ChessWidget(const StringView& set)
 {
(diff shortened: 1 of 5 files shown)


Games/Chess/ChessWidget.cpp
```diff
@@ -289,6 +292,58 @@ void ChessWidget::maybe_input_engine_move()
...
+        const String white = board().moves().at(i).to_algebraic();
+
+        if (i+1 < board().moves().size()) {
+            const String black = board().moves().at(i+1).to_algebraic();
+            file.write(String::formatted("{}. {} {} ", move_no, white, black));
+        }
+        else {
+            file.write(String::formatted("{}. {} ", move_no, white));
+        }        
+    }
```
emanuele6: You forgot to run `clang-format`.

emanuele6 (changes requested):

Libraries/LibChess/Chess.cpp
```diff
@@ -684,4 +687,40 @@ bool Board::operator==(const Board& other) const
...
+}
+
+String Board::result_to_string(Result r) const
+{
+    switch (r) {
+    case Result::CheckMate:
+        if (m_turn == Chess::Colour::White)
+            return "Black wins by Checkmate";
+        else
+            return "White wins by Checkmate";
```
emanuele6, line 772: Don't use `else` after `return`.

Also, you should probably assert that `m_turn` is not `Chess::Colour::None`:
```cpp
ASSERT(m_turn != Chess::Colour::None);
return m_turn == Chess:Colour::White ? "Black wins by Checkmate" : "White wins by Checkmate";
```

Libraries/LibChess/Chess.cpp
```diff
@@ -684,4 +757,71 @@ bool Board::operator==(const Board& other) const
...
+}
+
+String Board::result_to_points(Result r) const
+{
+    switch (r) {
+    case Result::CheckMate:
+        if (m_turn == Chess::Colour::White)
+            return "0-1";
+        else
+            return "1-0";
```
emanuele6, line 803: You should assert here too.

Libraries/LibChess/Chess.h
```diff
@@ -143,13 +154,17 @@ class Board {
         NotFinished,
     };
 
+    String result_to_string(Result) const;
+    String result_to_points(Result) const;
```
emanuele6, line 158: It's weird that these functions take a `Result`, but are not `static` since they use `m_turn`.

This means that to get the string representation of a board `b`, you have to call `result_to_string` function as:
```cpp
auto stringified_result = b.result_to_string(b.game_result());
```

Why don't you add to these functions a `Colour` (turn) parameter and make them `static` (or external helper functions)?

You could also just remove the `Result` argument and make these functions call `game_result()` internally.

AnicJov:
I rebased all of the changes. Hope I did it correctly.

Games/Chess/main.cpp
```diff
@@ -65,6 +67,16 @@ int main(int argc, char** argv)
         return 1;
     }
 
+    if (unveil("/home/anon/", "wcb") < 0) {
+        perror("unveil");
+        return 1;
+    }
```
bcoles, line 73: anon's home directory should not be hard coded.

A path is already unveiled for loading the user's config file a few lines above (using `Core::ConfigFile::get_for_app("Chess")`) :

```cpp
    if (unveil(config->file_name().characters(), "crw") < 0) {
        perror("unveil");
        return 1;
    }
```

So I'm guessing that you're unveiling the home directory to permit loading PGN files. In which case you can do something similar with `Core::StandardPaths::home_directory()` :

```suggestion
    if (unveil(Core::StandardPaths::home_directory().characters(), "wcb") < 0) {
        perror("unveil");
        return 1;
    }
```

Games/Chess/main.cpp
```diff
@@ -65,6 +67,16 @@ int main(int argc, char** argv)
...
+
+    if (unveil("/etc/passwd", "rb") < 0) {
+        perror("unveil");
+        return 1;
+    }
```
bcoles, line 78: `unevil` Browse `b` should not be necessary for `/etc/passwd` :

```suggestion
    if (unveil("/etc/passwd", "r") < 0) {
        perror("unveil");
        return 1;
    }
```

bcoles (changes requested):
Requires minor changes to use of `unveil`

bcoles:
I like this (but haven't tested).

FYI @Petelliott

Libraries/LibChess/Chess.h
```diff
@@ -105,6 +105,14 @@ struct Move {
     Square from;
     Square to;
     Type promote_to;
+    Piece piece;
+    bool is_check = false;
+    bool is_mate = false;
+    bool is_capture = false;
+    bool is_castle_short = false;
+    bool is_castle_long = false;
```
Petelliott, line 113: Castling can be determined by `from` `to` and `piece`. I think it would be better to make these methods based on those.

Games/Chess/ChessWidget.cpp
```diff
@@ -27,17 +27,20 @@
...
+#include <unistd.h>
 
 ChessWidget::ChessWidget(const StringView& set)
 {
     set_piece_set(set);
 }
 
 ChessWidget::ChessWidget()
-    : ChessWidget("test")
+    : ChessWidget("stelar7")
```
Petelliott, line 43: I take offense to this :^)

bcoles: You're welcome to override the default in your `Chess.ini` configuration file :P

Games/Chess/main.cpp
AnicJov: Yes, unveiling so that FilePicker can browse directories, and so PGN files can be written(/read eventually).
I agree, hard coding it was not the way to go.

What I wanted was the ability for FilePicker to have the same browse/read/write rights as the current user. Like what TextEditor does. But looking at TEs code, I didn't find an `unveil()` call and couldn't figure out how to replicate that behavior.
Do you have a solution for how to do this inside of Chess?

bcoles: > Do you have a solution for how to do this inside of Chess?

Does the code I provided work?

Libraries/LibChess/Chess.h
AnicJov: I didn't want to duplicate that castling logic, so I ended up setting those flags where castling would be calculated anyway.
It would be nice to debloat the `Move` struct by removing some flags though.
So I'm thinking of separating the castling check to it's own method and then call it from `apply_illegal_move()` and `Move::to_algebraic()`, removing the need for the bools.

Games/Chess/ChessWidget.cpp
AnicJov: I don't know the backstory here... I was just thinking from a first time user's perspective.
I can revert the change  : )

Games/Chess/main.cpp
AnicJov: > Does the code I provided work?

It does, but only for the user's home directory. It's fine like that I guess, but doesn't replicate TextEditor's behavior exactly. Which is maybe good for a chess program.
Either way I'm going with the code you provided for now.

bcoles: > 
> 
> > Does the code I provided work?
> 
> It does, but only for the user's home directory. It's fine like that I guess, but doesn't replicate TextEditor's behavior exactly. Which is maybe good for a chess program.
> Either way I'm going with the code you provided for now.

Ah yeah the TextEditor can browse the entire file system because it doesn't use `unveil`. Presumably the intention was to permit privileged users to use TextEditor for editing system config files which would require access to most of the system.

We have a new File IPC which was implemented along with the browse `b` unveil flag and seems pretty cool, but not something I've had a chance to look into yet.

The solution above is probably easiest and best for now.

Libraries/LibChess/Chess.h
AnicJov: Okay, since the logic in the other method is not quite the same, and the logic we need here is quite simple I just added the following at the start of `Move::to_algebraic()`:
```c++
if (piece.type == Type::King && from.file == 4) {
    if (to.file == 2)
        return "O-O-O";
    if (to.file == 6)
        return "O-O";
}
```

Games/Chess/ChessWidget.cpp
Petelliott: just kidding around. it's probably for the best long term

PR #4324, opened 2020-12-04
js: Improve pretty-printing of various objects

### LINKED ISSUE #4310: js: Most objects have useless representations
Primitive types and some objects (Array, Date, Error, Object, RegExp) have good representations, but many others don't:

```js
> new String("foo")
{  }
> new Number(42)
{  }
> new Boolean(true)
{  }
> new Proxy({}, { get() { return 42; } })
{  }
> new ArrayBuffer(10)
{  }
> new Uint8Array(10)
{  }
```

Node.js REPL for comparison:

```js
> new String("foo")
[String: 'foo']
> new Number(42)
[Number: 42]
> new Boolean(true)
[Boolean: true]
> new Proxy({}, { get() { return 42; } })
Proxy [ {}, { get: [Function: get] } ]
> new ArrayBuffer(10)
ArrayBuffer {
  [Uint8Contents]: <00 00 00 00 00 00 00 00 00 00>,
  byteLength: 10
}
> new Uint8Array(10)
Uint8Array(10) [
  0, 0, 0, 0, 0,
  0, 0, 0, 0, 0
]
```

I don't like the inconsistencies (`[Type: ...]`, `Type [ ... ]`, `Type { ... }`, `Type(...) [ ... ]`) there but it certainly can serve as inspiration.

linusg:
For many object types we only ever used the regular `print_object()` in the js REPL - resulting in a useless "{  }".

This patch adds more individual representations for the following types:
- `Boolean`/`Number`/`String` object: print wrapped value
- `ArrayBuffer`: print byteLength and hex-formatted bytes in chunks of 16
- `TypedArray`: print length, byteLength, buffer pointer and values
- `Proxy`: print target and handler

Also improve the existing pretty-printing output:
- More consistency, most objects now follow the format "[Type] ..."
- Some coloring tweaks
- No two spaces in empty arrays & objects

Possible future improvements:
- Add line breaks between elements of long arrays & objects

Closes #4310.

---

Old vs. new:

image
image

### DIFF
--- a/Libraries/LibJS/Runtime/Function.h
+++ b/Libraries/LibJS/Runtime/Function.h
@@ -50,8 +50,6 @@ public:
 
     virtual void visit_edges(Visitor&) override;
 
-    virtual bool is_script_function() const { return false; }
-
     BoundFunction* bind(Value bound_this_value, Vector<Value> arguments);
 
     Value bound_this() const { return m_bound_this; }
--- a/Libraries/LibJS/Runtime/Object.h
+++ b/Libraries/LibJS/Runtime/Object.h
@@ -117,6 +117,7 @@ public:
     virtual bool is_error() const { return false; }
     virtual bool is_function() const { return false; }
     virtual bool is_native_function() const { return false; }
+    virtual bool is_script_function() const { return false; }
     virtual bool is_bound_function() const { return false; }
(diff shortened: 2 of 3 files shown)


awesomekling:
Oh this is great :)

PR #4329, opened 2020-12-05
LibLine: Don't make Editor::load_history() cut off a character per line

linusg:
For some reason we were not considering the last *two* characters from the line's `ByteBuffer`, with the comment next to it talking about `\n` and `\0`. However the buffer doesn't contain a null-byte, so we were effectively removing the newline and the last character from each history line!


alimpfard:
This was some weirdness with the implementation of `IODevice::read_line()` sometimes (!) null-terminating the buffer for some reason, which was fixed in 129a58a.
I didn't see this instance when I was looking for what that might affect - sorry~

PR #4333, opened 2020-12-05
LibRegex: Minimal support for unicode escapes

alimpfard:


alimpfard:
uhhh...I can't see what failed

linusg:
https://github.com/SerenityOS/serenity/pull/4333/checks?check_run_id=1504422531#step:14:9112

> FAIL: ../Libraries/LibRegex/Tests/Regex.cpp:562: EXPECT_EQ(re.match(test.subject).success, test.matches) failed with lhs=false and rhs=true

PR #4338, opened 2020-12-06
AK: Add test for the will_be_destroyed and one_ref_left magic functions

ADKaster:
Fixes a regression introduced by 5c1b3ce. The commit description there
asserts that the changes allow calling will_be_destroyed and
one_ref_left, which are not required to be const qualified. The
implementation in fact does require the methods to be const qualified,
because we forgot to add the const_cast inside the decltypes :^)

### DIFF
--- a/AK/RefCounted.h
+++ b/AK/RefCounted.h
@@ -36,7 +36,7 @@
 namespace AK {
 
 template<class T>
-constexpr auto call_will_be_destroyed_if_present(const T* object) -> decltype(object->will_be_destroyed(), TrueType {})
+constexpr auto call_will_be_destroyed_if_present(const T* object) -> decltype(const_cast<T*>(object)->will_be_destroyed(), TrueType {})
 {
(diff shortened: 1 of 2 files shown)


ADKaster:
In practice, fixing this bug means... TmpFS Inodes are now actually unregistered when they are destroyed? Exercise for the reviewer to see if there's an exploit there ;)

ADKaster:
An alternative implementation might remove the const_cast from call_x_if_present and make unref() uglier with a static cast and a const_cast before calling call_x_if_present. Or, force those methods to be const? Not sure I understand the utility of will_be_destroyed, since the destructor will be called right after. Probably "something something large class hierarchies".

awesomekling:
> Not sure I understand the utility of will_be_destroyed, since the destructor will be called right after. Probably "something something large class hierarchies".

It allows you to call virtual functions on an object before it gets destroyed (and its vtable ruined) :)

awesomekling:
Great catch!

PR #4339, opened 2020-12-06
Add a AudioDevice class to be used as a base for sound cards

tomuta:
_This is still very much work in progress... Based on #4320_

- [ ] Figure out why `SB16` locks up the system as soon as playback is triggered

This implements a framework for negotiating audio features and various
settings between LibAudio and kernel drivers. By implementing most
features in the new abstract AudioDevice, implementing new sound card
drivers is greatly simplified while still maintaining compatibility
with LibAudio.

Also add an Audio::Device class to LibAudio that will take care
of the interaction with all drivers that are based on the AudioDevice
class.

Last but not least, update AudioServer to use the Audio::Device class.

### DIFF
--- a/Kernel/Devices/AudioDevice.cpp
+++ b/Kernel/Devices/AudioDevice.cpp
@@ -1,0 +1,471 @@
+/*
+ * Copyright (c) 2021, the SerenityOS developers.
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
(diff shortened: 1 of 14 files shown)


danboid:
Exciting! :D

Kernel/Devices/SB16.cpp
```diff
@@ -184,114 +225,156 @@ void SB16::set_irq_line(u8 irq_number)
...
 
-    sample_count -= 1;
+    // Write the buffer addresss
+    auto offset = (m_is16bit ? addr / 2 : addr) % 65536;
+    IO::out8(m_is16bit ? 0x8b : 0x83, addr >> 16);
+    IO::out8(m_is16bit ? 0xc4 : 0x02, (u8)offset);
+    IO::out8(m_is16bit ? 0xc4 : 0x02, (u8)(offset >> 8));
 
-    cli();
-    enable_irq();
```
awesomekling, line 286: Where are SB16 IRQ's supposed to be enabled? I would guess this is why the system locks up, if you're expecting an IRQ.

Kernel/Devices/AudioDevice.cpp
```diff
@@ -0,0 +1,454 @@
...
+    if (!card.current_request->buffer().read(dest_ptr, bytes_to_copy)) {
+        complete_current_request(card, AsyncDeviceRequest::RequestResult::MemoryFault);
+        return false;
+    }
+    
+    card.buffer_write_offset += bytes_to_copy;
+    card.request_buffer_offset += bytes_to_copy;
+
+    if (card.state == Card::State::Prepared && period_full && (card.current.periods_trigger == 0 || card.buffer_write_offset / card.bytes_per_period == card.current.periods_trigger)) {
+        if (!period_full && card.current.periods_trigger == 0) {
```
awesomekling, line 137: The inner `if` will never be entered since `period_full` is always `true` per the outer `if.

Kernel/Devices/AudioDevice.cpp
```diff
@@ -0,0 +1,454 @@
...
+        dbgln("in state {}", (int)card.state);
+    }
+    if (card.request_buffer_offset >= card.current_request->buffer_size()) {
+        complete_current_request(card, AsyncDeviceRequest::RequestResult::Success);
+        return false; // This doesn't indicate an error!
+    }
+    return true;
+}
+
+void AudioDevice::finished_playing_period(Card& card)
```
awesomekling, line 158: Where is this supposed to be called? The NOTE comment inside mentions being called from an interrupt handler but I don't see that happening.

Kernel/Devices/SB16.cpp
tbluemel-control4: Yeah that was a miss, fixed that.

PR #4340, opened 2020-12-06
LibC+Kernel: Implement SA_SIGINFO for signal actions

ADKaster:
This patch adds support for SA_SIGINFO signal handlers.

- [x] Add definitions for ucontext_t, stack_t, and siginfo_t::si_code values to `<signal.h>` and `"Kernel/UnixTypes.h"`
- [x] Write test/example program
- [x] Put ucontext_t and siginfo_t on stack in Thread::dispatch_signal(u8)
- [x] Push pointer to stashed ucontext and siginfo onto front(?) of stack next to signal number
- [x] Fumble around in the signal trampoline assembly for an hour or two (Black Magic R Us)
- [x] Implement siginfo in UserspaceEmulator
- [ ] Debug (or put off?) misaligned stack addresses that occur when sending a ton of signals  in rapid succession, to some applications. Especially likely to happen when running applications in UserspaceEmulator.

Note that the implementation here makes every single signal push all of this junk to the stack, instead of just the ones for which an SA_SIGINFO handler was registered. Doing that logic would probably require... even more changes to that signal trampoline assembly. Something I definitely do not want to mess with right now, that stuff is gnarly and confusing.

So, that's a doubling of the "signal stack" overhead (56 --> 116 bytes).

After this set of changes, then we come to terms with the fact that, at `dispatch_signal` time, we have lost a ton of useful information about the signal we're handling. The types of things we can use to fill in the `si_code` field of the siginfo_t. The types of things that could let us dump a lot of interesting backtrace information in userspace on crash 😄.

Reminder to self: make ucontext_t::uc_mcontext equivalent to RegisterDump so that those could possibly be merged? Sounds gnarly as well :/

### DIFF
--- a/Kernel/Process.cpp
+++ b/Kernel/Process.cpp
@@ -414,11 +414,15 @@ void signal_trampoline_dummy()
         "push ebp\n"
         "mov ebp, esp\n"
         "push eax\n"          // we have to store eax 'cause it might be the return value from a syscall
-        "sub esp, 4\n"        // align the stack to 16 bytes
+        "sub esp, 12\n"       // TODO: Is this the correct way to align the stack?
+        "mov eax, [ebp+20]\n" // push ucontext_t*
+        "push eax\n"
+        "mov eax, [ebp+16]\n" // push siginfo_t*
+        "push eax\n"
         "mov eax, [ebp+12]\n" // push the signal code
         "push eax\n"
         "call [ebp+8]\n" // call the signal handler
-        "add esp, 8\n"
+        "add esp, 16\n"  // Unroll the stack, such that the bottom of the stack is our register dump
         "mov eax, %P0\n"
         "int 0x82\n" // sigreturn syscall
         "asm_signal_trampoline_end:\n"
--- a/Kernel/Thread.cpp
+++ b/Kernel/Thread.cpp
@@ -703,8 +703,6 @@ DispatchSignalResult Thread::dispatch_signal(u8 signal)
     //    }
 
     auto& action = m_signal_action_data[signal];
-    // FIXME: Implement SA_SIGINFO signal handlers.
-    ASSERT(!(action.flags & SA_SIGINFO));
 
     // Mark this signal as handled.
     m_pending_signals &= ~(1 << (signal - 1));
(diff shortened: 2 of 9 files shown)


Kernel/Thread.cpp
```diff
@@ -710,56 +708,107 @@ DispatchSignalResult Thread::dispatch_signal(u8 signal)
...
+    // FIXME: Plumb sender information into this part. We need to know more about the signal.
+    //        Maybe need a struct SignalReason?
+    auto signal_info = siginfo_for_signal(signal, nullptr);
 
-        push_value_on_user_stack(stack, signal);
-        push_value_on_user_stack(stack, handler_vaddr.get());
-        push_value_on_user_stack(stack, 0); //push fake return address
+    ucontext_t user_context {};
+    user_context.uc_link = nullptr; // TODO: Are we expected to maintain a linked list of these things in the kernel? in userspace?
+    user_context.uc_mcontext = 0;   // TODO: What is this for anyway? Dr Posix pls
```
alimpfard, line 746: IIRC this is for the register dump.

Kernel/Thread.cpp
```diff
@@ -710,56 +708,107 @@ DispatchSignalResult Thread::dispatch_signal(u8 signal)
...
-        push_value_on_user_stack(stack, old_signal_mask);
+    // FIXME: Plumb sender information into this part. We need to know more about the signal.
+    //        Maybe need a struct SignalReason?
+    auto signal_info = siginfo_for_signal(signal, nullptr);
 
-        push_value_on_user_stack(stack, signal);
-        push_value_on_user_stack(stack, handler_vaddr.get());
-        push_value_on_user_stack(stack, 0); //push fake return address
+    ucontext_t user_context {};
+    user_context.uc_link = nullptr; // TODO: Are we expected to maintain a linked list of these things in the kernel? in userspace?
```
alimpfard, line 745: This guy is for `swapcontext()`, it's set up by the user after `makecontext()` to be the context to return to when the swapped context returns.

alimpfard: see: https://pubs.opengroup.org/onlinepubs/7908799/xsh/getcontext.html and https://pubs.opengroup.org/onlinepubs/7908799/xsh/makecontext.html

Kernel/Thread.cpp
ADKaster: Gotcha yeah, figured that out last night and made an edit to the PR description :D

Probably the proper way to do this would be to have the register dump that is used in `sys$sigreturn` be the one from the ucontext?

Kernel/Thread.cpp
ADKaster: Weird. So it's like old-style coroutines? That have to go through the kernel? And nobody in their right mind would use anymore? Or would the user context in a signal handler not include any kind of context data, and the save/return stuff would all be handled by the C library?

Kernel/Process.cpp
```diff
@@ -414,10 +414,14 @@ void signal_trampoline_dummy(void)
...
         "sub esp, 4\n"        // align the stack to 16 bytes
+        "mov eax, [ebp+20]\n" // push ucontext_t*
+        "push eax\n"
+        "mov eax, [ebp+16]\n" // push siginfo_t*
+        "push eax\n"
         "mov eax, [ebp+12]\n" // push the signal code
         "push eax\n"
         "call [ebp+8]\n" // call the signal handler
-        "add esp, 8\n"
+        "add esp, 16\n"
```
ADKaster, line 424: These changes to the asm here screw up the stack alignment or something. My test program works fine, but sending things like SIGINT to a process from the shell asserts in sigaction that the restored stack ptr == the calculated one (after removing the siginfo and ucontext). Not sure how to get it  correct, since I don't understand how the 16 byte alignment requirement comes into play nor how what the original code does makes things work?

Kernel/Thread.cpp
alimpfard: That, I don't know for sure - no one (right mind or not) uses that stuff.
re in the signal handler, I think you just get the context as `getcontext()` would give you, of wherever the thread was. not sure if this is used in the resumption though
I think this whole `swapcontext()` is just a libc thing, the kernel isn't involved here (again, I think), afaik it just sets up the registers, pushes an address and then `ret`'s to it. not entirely sure though?

Kernel/Thread.cpp
alimpfard: That's what I get for not reading the "Reminder to self" lol

Kernel/Thread.cpp
ADKaster: > SUSv2, POSIX.1-2001.  POSIX.1-2008 removes the specifications of
> makecontext() and swapcontext(), citing portability issues, and
> recommending that applications be rewritten to use POSIX threads
> instead.

Ah yes. pre-pthread "threads". Interesting that they were actually removed. Probably a better use of time to implement C++20 coroutine support than those crusty APIs.

alimpfard: I'm all for implementing C++20 coros over `swapcontext()` :P

Kernel/Process.cpp
ADKaster: Aha, I'm forgetting about the alignment at the top of setup_signal_stack. My assertion was wrong, not the code

ADKaster:
I'm gonna have to come back to this one... I really don't understand the signal stacks and the weird alignment errors threw me for a loop every time I tried to debug...

PR #4346, opened 2020-12-06
LibVT+Terminal: Enable Copy/paste actions conditionally / Add "Copy name" action to terminal link context menu

linusg:
See commit messages for details.

Screenshots:

image
image
image
image
image

### DIFF
--- a/Libraries/LibGUI/Action.h
+++ b/Libraries/LibGUI/Action.h
@@ -132,6 +132,9 @@ public:
     }
     void set_checked(bool);
 
+    bool swallow_key_event_when_disabled() const { return m_swallow_key_event_when_disabled; }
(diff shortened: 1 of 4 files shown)


Libraries/LibGUI/Action.h
```diff
@@ -132,6 +132,9 @@ class Action final : public Core::Object {
     }
     void set_checked(bool);
 
+    bool swallow_key_event_when_disabled() const { return m_swallow_key_event_when_disabled; }
+    void set_swallow_key_event_when_disabled(bool swallow) { m_swallow_key_event_when_disabled = swallow; }
```
awesomekling, line 136: This is a very niche situation but I think you've come up with a decent way of handling it.

Issue #4330, opened 2020-12-05
Kernel crashes on thread joins


BenWiederhake:
Steps to reproduce: Run `tt` (no arguments)

Expected behavior: Test passes, and outputs something like `Okay, joined and got retval=0xDEADBEEF`

Actual behavior: Kernel crash

```
[tt(43:44)]: ASSERTION FAILED: did_unblock
.././Kernel/Thread.h:433 in void Kernel::Thread::BlockCondition::do_unblock_all(UnblockOne) [with UnblockOne = Kernel::Thread::JoinBlockCondition::do_unblock_joiner()::<lambda(Kernel::Thread::Blocker&, void*)>]
[tt(43:44)]: 0xc011f355  __assertion_failed(char const*, char const*, unsigned int, char const*) +185
[tt(43:44)]: 0xc01f68d8  void Kernel::Thread::BlockCondition::do_unblock_all<Kernel::Thread::JoinBlockCondition::do_unblock_joiner()::{lambda(Kernel::Thread::Blocker&, void*)#1}>(Kernel::Thread::JoinBlockCondition::do_unblock_joiner()::{lambda(Kernel::Thread::Blocker&, void*)#1}) +110
[tt(43:44)]: 0xc019eac1  Kernel::Thread::exit(void*) +167
[tt(43:44)]: 0xc0194654  Kernel::Process::sys$exit_thread(AK::Userspace<void*>) +140
[tt(43:44)]: 0xc017b922  syscall_handler +702
[tt(43:44)]: 0xc017b65c  syscall_asm_entry +49
[tt(43:44)]: 0x08048d92  mutex_test()::{lambda(void*)#1}::_FUN(void*) +0
[tt(43:44)]: 0x0804a69d  pthread_create_helper +12
```

CC @tomuta, since this seems related to 046d6855f5e8a5039b319a47c3018a16d4c2f960.

tomuta:
I've seen this a few times but haven't gotten around to look at it yet, but I have some guesses as to what's wrong.

Lubrsi:
Also happens with `test-pthread`:
```
[test-pthread(24:25)]: ASSERTION FAILED: did_unblock
.././Kernel/Thread.h:433 in void Kernel::Thread::BlockCondition::do_unblock_all(UnblockOne) [with UnblockOne = Kernel::Thread::JoinBlockCondition::do_unblock_joiner()::<lambda(Kernel::Thread::Blocker&, void*)>]
[...]

PR #4347, opened 2020-12-07
Kernel: Fix some issues related to fixes and block conditions

### LINKED ISSUE #4336: Kernel: Cannot send signal to self
Consider the following program:

send-signal-to-self.cpp
```c++

#include <signal.h>
#include <stdio.h>
#include <unistd.h>

volatile sig_atomic_t saved_signal = -1;
volatile bool got_signal = false;

static void my_handler(int sig)
{
    saved_signal = sig;
    got_signal = true;
}

int main()
{
    if (SIG_ERR == signal(SIGUSR1, my_handler)) {
        perror("signal");
        return 1;
    }

    sleep(1);

    if (0 != kill(getpid(), SIGUSR1)) {
        perror("kill");
        return 2;
    }

    sleep(1);

    if (!got_signal) {
        fprintf(stderr, "Where's my signal, bro?\n");
        return 3;
    }

    printf("Got signal %d\n", saved_signal);

    return 0;
}
```

In serenity, this prints "Where's my signal, bro?"
signal-to-self

On Ubuntu 20.04, it prints, "Got signal 10".
```
andrew@DESKTOP-QQ5PM6D:~/testing$ g++ send-signal-to-self.cpp -o send-signal-to-self
andrew@DESKTOP-QQ5PM6D:~/testing$ ./send-signal-to-self 
Got signal 10
andrew@DESKTOP-QQ5PM6D:~/testing$ 
```

tomuta:
Kernel: Fix some issues related to fixes and block conditions

Fix some problems with join blocks where the joining thread block
condition was added twice, which lead to a crash when trying to
unblock that condition a second time.

Deferred block condition evaluation by File objects were also not
properly keeping the File object alive, which lead to some random
crashes and corruption problems.

Other problems were caused by the fact that the Queued state didn't
handle signals/interruptions consistently. To solve these issues we
remove this state entirely, along with Thread::wait_on and change
the WaitQueue into a BlockCondition instead.

Also, deliver signals even if there isn't going to be a context switch
to another thread.

Fixes #4336 and #4330


Kernel/FileSystem/File.h
```diff
@@ -142,8 +142,9 @@ class File
     {
         if (Processor::current().in_irq()) {
             // If called from an IRQ handler we need to delay evaluation
-            // and unblocking of waiting threads
-            Processor::deferred_call_queue([this]() {
+            // and unblocking of waiting threads. We need to make sure that
+            // this instance stays alive until then!
+            Processor::deferred_call_queue([this, self = NonnullRefPtr<File>(*this)]() {
```
awesomekling, line 147: Would this be even better with a `WeakPtr` to the `File`?
If all other references to the file have gone away, nobody should care about the block conditions anyway.

tomuta:
The downside to this is now all the code that uses `WaitQueue` in functions called `wait_for_irq` in various drivers, such as SB16. For SB16 specifically that should be fixed in #4339. The `PATAChannel` was already fixed for this some time ago, which is why the system is working well. There are a few other drivers that need fixing, such as networking but that seems to work ok (I'm guessing because of the kernel process `NetworkTask` doing the work rather with the device drivers?)

alimpfard:
Can confirm that c595b34 fixes the weird zombie process in the shell~

tomuta:
Ok, got `strace` to work with multi-threaded programs, too :)

tomuta:
I think this is in a good state now, seems to work quite well, at least the things I tested.

PR #4348, opened 2020-12-07
FileManager: Fix TreeView collapsing on file system changes

### LINKED ISSUE #4289: FileManager: Deleting or pasting files collapses tree view
Whenever you delete or paste any files or directories the filesystem tree view in the left pane of FileManager is collapsed so that only / is visible but in an unexpanded state.

supex0fan:
TreeViews using FileSystemModels collapse whenever the file system is changed in any way. This includes creating, dragging, deleting or pasting any files/folders. This commit updates the refresh_tree_view() lambda, which seems to have stopped working at some point, and calls it whenever any of the actions mentioned above are activated.

If it is possible to fix this problem at the treeview level, that is obviously preferred, I looked into it but couldn't find any solution.

Fixes #4289

### DIFF
--- a/Applications/FileManager/DirectoryView.cpp
+++ b/Applications/FileManager/DirectoryView.cpp
@@ -563,6 +563,7 @@ void DirectoryView::handle_drop(const GUI::ModelIndex& index, const GUI::DropEve
     if (!target_node.is_directory())
         return;
 
+    bool had_accepted_drop = false;
     for (auto& url_to_copy : urls) {
         if (!url_to_copy.is_valid() || url_to_copy.path() == target_node.full_path())
             continue;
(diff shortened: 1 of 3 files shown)


awesomekling:
Yeah all right. This should indeed eventually be fixed at the LibGUI level (with proper model index invalidation) but for now this is nice to have. :)

supex0fan:
> This should indeed eventually be fixed at the LibGUI level (with proper model index invalidation)

If you can expand on that, let me know and I'll take another look.

supex0fan:
Right... Ignore that, I found the problem in the TreeView's source. In `model_did_update()`, `m_view_metadata` (which stores which indexes are open) is cleared. If the clear is commented out, everything seems to work fine without `refresh_tree_view()`, the tree is updated, and it doesn't collapse. Do we need the `m_view_metadata.clear()`?

awesomekling:
After the model updates, none of the ModelIndexes are guaranteed to be valid. This is why we clear the metadata, since it could all be mapped using dangling pointers.

The proper solution is to make models notify the view whenever indexes are inserted, removed or moved around, so that it can keep any such mappings up to date. The LibGUI model/view classes are based on the ones in Qt, you can see how they've implemented it here: https://doc.qt.io/qt-5/qabstractitemmodel.html :)

supex0fan:
Thanks for the reference, I'll look into it.

PR #4350, opened 2020-12-07
LibJS: Implement String.prototype.split

### LINKED ISSUE #4349: LibJS: String.prototype.split is missing
I was looking at what I could do to get the DuckDuckGo page to load and noticed that code like ``"this string".split(" ")`` was failing. I have a pull request I'll send but it will probably take a few iterations to iron it out.

xspager:
The code is not good but it's a start. I need some assistance getting it good enough to get merged. This is supposed to fix #4349

### DIFF
--- a/Libraries/LibJS/Runtime/StringPrototype.cpp
+++ b/Libraries/LibJS/Runtime/StringPrototype.cpp
@@ -28,6 +28,7 @@
 #include <AK/Function.h>
 #include <AK/StringBuilder.h>
 #include <LibJS/Heap/Heap.h>
+#include <LibJS/Runtime/Array.h>
 #include <LibJS/Runtime/Error.h>
 #include <LibJS/Runtime/GlobalObject.h>
 #include <LibJS/Runtime/PrimitiveString.h>
@@ -90,6 +91,7 @@ void StringPrototype::initialize(GlobalObject& global_object)
     define_native_function(vm.names.includes, includes, 1, attr);
     define_native_function(vm.names.slice, slice, 2, attr);
     define_native_function(vm.names.lastIndexOf, last_index_of, 1, attr);
+    define_native_function(vm.names.split, split, 1, attr);
     define_native_function(vm.well_known_symbol_iterator(), symbol_iterator, 0, attr);
 }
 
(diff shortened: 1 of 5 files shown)


linusg (changes requested):
Please follow the spec closely, adding todo comments where necessary. Things like checking for arguments being missing rather than nullish introduce subtle non-standard behavior. Doing it right the first time (and skipping parts if necessary) is a lot easier than tracking those down retrospectively.

awesomekling:
> Please follow the spec closely, adding todo comments where necessary. Things like checking for arguments being missing rather than nullish introduce subtle non-standard behavior. Doing it right the first time (and skipping parts if necessary) is a lot easier than tracking those down retrospectively.

For reference: https://tc39.es/ecma262/#sec-string.prototype.split

Libraries/LibJS/Tests/builtins/String/String.js
```diff
@@ -7,3 +7,9 @@ test("typeof", () => {
     expect(typeof String()).toBe("string");
     expect(typeof new String()).toBe("object");
 });
+
+test("string split", () => {
+    expect("foo bar baz".split(" ")).toEqual(["foo", "bar", "baz"]);
+    expect("foo,bar,baz".split(",")).toEqual(["foo", "bar", "baz"]);
+    expect("foo||bar||baz".split("||")).toEqual(["foo", "bar", "baz"]);
+});
```
linusg, line 15: These belong in a separate file `String.prototype.split.js`, look at the other StringPrototype tests for examples.

xspager: Ok, I'll try to make this changes and work on some error checking. Thanks! :)

alimpfard:
`String.replace(RegExp)` \*wink wink nudge nudge\*

Libraries/LibJS/Runtime/StringPrototype.cpp
```diff
@@ -507,4 +509,27 @@ JS_DEFINE_NATIVE_FUNCTION(StringPrototype::symbol_iterator)
...
+JS_DEFINE_NATIVE_FUNCTION(StringPrototype::split)
+{
+    auto string = ak_string_from(vm, global_object);
+    if (string.is_null())
+        return {};
+
+    if (vm.argument_count() == 0)
+        return js_string(vm, string);
+
+    auto separator_string = vm.argument(0).to_string(global_object);
```
linusg, line 521: Missing exception check

linusg (commented):
hmm, not sure about this. as the tests prove it works for the basic case of one string argument, but that's about it. This is far from being spec compliant (missing support for `@@split` and `limit`) - we of course can (and will) iterate on it later, but at the very least add a comment in the code mentioning that.

My earlier comment wasn't addressed and still is true:

> Things like checking for arguments being missing rather than nullish introduce subtle non-standard behavior. Doing it right the first time (and skipping parts if necessary) is a lot easier than tracking those down retrospectively.

Specific example:

```cpp
if (vm.argument_count() == 0)
    return js_string(vm, string);
```

What the spec says:

> If separator is undefined, then
> 
>    Perform ! CreateDataPropertyOrThrow(A, "0", S).
>    Return A.

Which is something completely different. It's supposed to return a one-element array with the string, not the string itself.

Again, a non-complete first version is perfectly fine, but the code & commit messages should reflect that :)

linusg:
Superseded by #4872.

PR #4351, opened 2020-12-07
Fewer crashes, less attack surface, more features, mixed bag

BenWiederhake:
The only common theme here is "I ran into these problems while working on something else". As always: If you prefer I make these as multiple PRs, no problem.

- Initialize stuff in WindowServer correctly, which used to crash *other* programs (e.g. `pape -c`)
- Don't rely on the `bpp` metadata value for clipboard images. A misbehaving program was able to set bpps like `33` or `sevenbajillion lol`, which then would show up verbatim in the Clipboard History. (There's more PRs coming for that.)
- UserEmulator, realpath: Handle weird buffer requirements more reasonably. (For example, if the return buffer was too small, UE would just silently, truncate the result, whereas the Kernel would.)
- UserEmulator, readlink: Implement. Now we support such sophisticated programs such as `/bin/readlink`! (And `FileManager` runs a little longer until it runs into the unimplemented syscall `create_thread`.)

If anyone feels like it: `virt$realpath` and `virt$readlink` are very similar, and should probably be abstracted/generalized/put-together somehow.

### DIFF
--- a/DevTools/UserspaceEmulator/Emulator.cpp
+++ b/DevTools/UserspaceEmulator/Emulator.cpp
@@ -1137,16 +1137,20 @@ int Emulator::virt$realpath(FlatPtr params_addr)
     Syscall::SC_realpath_params params;
     mmu().copy_from_vm(&params, params_addr, sizeof(params));
 
+    if (params.path.length > PATH_MAX) {
+        return -ENAMETOOLONG;
+    }
     auto path = mmu().copy_buffer_from_vm((FlatPtr)params.path.characters, params.path.length);
     char host_buffer[PATH_MAX] = {};
+    size_t host_buffer_size = min(sizeof(host_buffer), params.buffer.size);
 
     Syscall::SC_realpath_params host_params;
     host_params.path = { (const char*)path.data(), path.size() };
-    host_params.buffer = { host_buffer, sizeof(host_buffer) };
+    host_params.buffer = { host_buffer, host_buffer_size };
     int rc = syscall(SC_realpath, &host_params);
     if (rc < 0)
         return rc;
-    mmu().copy_to_vm((FlatPtr)params.buffer.data, host_buffer, min(params.buffer.size, sizeof(host_buffer)));
+    mmu().copy_to_vm((FlatPtr)params.buffer.data, host_buffer, host_buffer_size);
     return rc;
 }
 
--- a/DevTools/UserspaceEmulator/Emulator.cpp
+++ b/DevTools/UserspaceEmulator/Emulator.cpp
@@ -404,6 +404,8 @@ u32 Emulator::virt_syscall(u32 function, u32 arg1, u32 arg2, u32 arg3)
         return virt$watch_file(arg1, arg2);
     case SC_clock_nanosleep:
         return virt$clock_nanosleep(arg1);
+    case SC_readlink:
+        return virt$readlink(arg1);
     default:
         reportln("\n=={}==  \033[31;1mUnimplemented syscall: {}\033[0m, {:p}", getpid(), Syscall::to_string((Syscall::Function)function), function);
         dump_backtrace();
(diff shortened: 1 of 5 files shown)


BenWiederhake:
readlink running in UserspaceEmulator

awesomekling:
s/UserEmulator/UserspaceEmulator/ ;)

BenWiederhake:
I agree! But historically, all commits call it UserEmulator, and now we have to stick with it D:

awesomekling:
Wha? I can't find a single mention of "UserEmulator" anywhere except this PR :)

BenWiederhake:
I could swear I checked the git history and every commit called it `UserEmulator`. Apparently I was wrong!

Changes since last push:
- Rebased for convenience
- Reworded to `User*space*Emulator`

DevTools/UserspaceEmulator/Emulator.cpp
```diff
@@ -1523,4 +1529,26 @@ int Emulator::virt$clock_nanosleep(FlatPtr params_addr)
...
+int Emulator::virt$readlink(FlatPtr params_addr)
+{
+    Syscall::SC_readlink_params params;
+    mmu().copy_from_vm(&params, params_addr, sizeof(params));
+
+    if (params.path.length > PATH_MAX) {
+        return -ENAMETOOLONG;
+    }
+    auto path = mmu().copy_buffer_from_vm((FlatPtr)params.path.characters, params.path.length);
+    char host_buffer[PATH_MAX] = {};
```
bugaevc, line 1541: Please, please, don't use `PATH_MAX`, except for checking syscall inputs against it (like it's done a few lines above — that one is OK). In particular, don't ever use it for any buffer sizes.

Using a buffer of size `PATH_MAX` here will both overallocate for the common case — most symlink paths are in fact very short, e.g. we have `/bin/sh` → `Shell` (5 bytes) and `/bin/[` → `test` (4 bytes) — and not work in case the symlink path is actually longer than `PATH_MAX`.

I have previously made `Core::File::read_link()` work properly for both short links (the common case) and arbitrarily long links, see https://github.com/SerenityOS/serenity/commit/47d83800e1c27c0a26d3ee432973d49a3f83f299 and https://github.com/SerenityOS/serenity/commit/d89843f96fa7673f409bb8c33e043e2355b257f1. We should make it work as well when UE-ed.

BenWiederhake: Regarding the LibC implementation: How is this better? Now in the case of overlong pathnames, the result is truncated. Taken from your linked commit:

```
__RETURN_WITH_ERRNO(rc, min((size_t)rc, size), -1);
```
That looks like a problem, because it can sometimes return something that is not an error code *and* not the target of the link.

I agree that over-allocation isn't particularly nice, but that's how `virt$realpath` does it, and it seems comparable to a `malloc`. So I went with the same thing, for consistency.

If we get rid of `PATH_MAX`, then we run into the issue that we need to allocate `params.buffer.size` amount of memory, without knowing the actual size yet that will be required. Since that variable is controlled by the program being emulated, this could be a malicious value like 2GiB. Where do you draw the line? `PATH_MAX` seemed like a good idea. We could also validate that the entire `params.buffer` region is mapped and writable, but that still leaves the question open how to behave if it doesn't work.

Please tell me (or implement) how to do that in a better way.

Also, have you experimented with extremely long paths? I run into trouble at a total path length of 2K already (`PATH_MAX = 4K`), so something fishy is going on anyways.

bugaevc: > Regarding the LibC implementation: How is this better? Now in the case of overlong pathnames, the result is silently truncated.

Yes, this is what POSIX seems to want:

> If the buf argument is not large enough to contain the link content, the first bufsize bytes shall be placed in buf. <...> Upon successful completion, these functions shall return the count of bytes placed in the buffer. Otherwise, these functions shall return a value of -1, leave the buffer unchanged, and set errno to indicate the error.

(So clearly they consider the case of a too-short buffer a "successful completion", not an error.)

The `File::read_link()` function, on the other hand, can be made to handle this properly, which I've done.

> I agree that over-allocation isn't particularly nice, but that's how `virt$realpath` does it

That one, too, needs fixing :slightly_smiling_face:

> Please tell me (or implement) how to do that in a better way.

I'd just use `params.buffer.size` here.

> If we get rid of PATH_MAX, then we run into the issue that we need to allocate params.buffer.size amount of memory, without knowing the actual size yet that will be required. Since that variable is controlled by the program being emulated, this could be a malicious value like 2GiB. Where do you draw the line?

I mean, this emulated program is not being sandboxed by us (UE) in any way, in fact we're running _in its process_. In the same way that an in-process library does not sandbox a potentially malicious program using it. If it asks for 2 GB, so be it. It's somebody else's job to impose memory limits on us IMO.

> Also, have you experimented with extremely long paths?

Not really. It'd be a good idea to do that and fix up all the bugs we encounter.

BenWiederhake: Let's move this to the "Discussions" tab that @awesomekling enabled recently: https://github.com/SerenityOS/serenity/discussions/4357

PR #4352, opened 2020-12-07
TCP: Fix double unlock in TCPSocket::protocol_connect

thatoddmailbox:
This fixes an issue where the TCPSocket's lock could be unlocked twice. The easiest way to cause this would be running something like `nc 127.0.0.1 1234` (without anything listening on port 1234) and then Control-C-ing `nc`.

The issue here was that we returned without holding the lock, and then the Locker's destructor would run, which would then try to unlock. I fixed it by relocking right before we return, but you could argue it's a bit of a hack...it might make more sense for the Locker's destructor to skip unlocking if we aren't holding the lock anymore? That felt like a far more disruptive change, though, so I just did this for now.


tomuta:
Feels like `Locker` should keep track of this internally.

PR #4354, opened 2020-12-08
LibGfx: SIMD optimized alpha blending

### LINKED ISSUE #69: Painter: SIMD optimized alpha blending
Alpha blending is currently pretty slow. It could be faster if it were written with SIMD.

ccapitalK:
Add SIMD optimized alpha blending. Also, make Gamma.h a bit nicer (and refactor some stuff that shouldn't be in there out of it).

Preliminary benchmarking shows that this makes Color::blend() about twice as fast on my machine.

Fixes #69.

### DIFF
--- a/AK/SIMD.h
+++ b/AK/SIMD.h
@@ -1,0 +1,76 @@
+/*
+ * Copyright (c) 2020, the SerenityOS developers.
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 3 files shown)


ccapitalK:
I think we'll eventually want blending to be gamma accurate as well, it isn't currently. It might be too much of a performance hit though.

awesomekling:
Oh that's really cool :)

PR #4364, opened 2020-12-09
LibWeb: Don't allocate XMLHttpRequestConstructor twice

linusg:
`add_constructor()` will already allocate an `XMLHttpRequestConstructor` and update `m_xhr_constructor` accordingly, we don't have to do it ourselves. This is now in line with how all the LibJS constructors work. Also make the XHR constructor responsible for setting its "prototype" property itself, for consistency and fail-proofing. Previously we would only set it on the constructor we allocated manually but which was then thrown away, leading to the property never being set properly.


awesomekling:
The big task here is to extend the WrapperGenerator in LibWeb to create these prototypes and constructor classes for us.
But it's really good that you're making this one correct first, since this can be our template that we base the code generation on :)

PR #4367, opened 2020-12-09
Userland: Fix segfault in sort

ccapitalK:
Currently, sort always segfaults since it passes nullptr as the second parameter to getline. This fixes that.


ccapitalK:
Our implementation of `getline(buf, n, file)` calls `getdelim(buf, n, '\n', file)` which assumes that `n != nullptr`. Glibc's getline also segfaults if passed null there, so I figured it would make more sense to make `n` point to an unused variable than to make getline handle `n` being null.

PR #4368, opened 2020-12-09
Userland: Fix overly-eager loop detection in cp

### LINKED ISSUE #4365: Userland: Over-eager detection of directory self-copying
To reproduce:

```
$ cp -r foo/ foo2/
```

Expected behaviour:

The folder is copied.

Actual behaviour:

```
cp: Cannot copy foo/ into itself (foo2/)
```

ccapitalK:
Fixes #4365.

### DIFF
--- a/Userland/cp.cpp
+++ b/Userland/cp.cpp
@@ -181,7 +181,9 @@ bool copy_directory(String src_path, String dst_path)
     }
 
     String src_rp = Core::File::real_path_for(src_path);
+    src_rp = String::format("%s/", src_rp.characters());
     String dst_rp = Core::File::real_path_for(dst_path);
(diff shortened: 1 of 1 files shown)


awesomekling:
What was the bug? It's not obvious from the changes what was actually going wrong.

ccapitalK:
The bug is that if you try to cp DIR_A to DIR_B where DIR_A's name is a prefix of DIR_B (e.g. foo -> foo2, bar -> barbar), it thinks that it's a subdirectory (since it checks if DIR_A's realpath is a prefix of DIR_B's realpath). The easiest solution is to put a path delimiter at the end before the comparison, since you can't have a '/' in the middle of a path component.

awesomekling:
Thank you for explaining, and for fixing the issue :)

PR #4370, opened 2020-12-09
Meta: Add .pre-commit-config.yaml

### LINKED ISSUE #2365: Meta: Some way to opt in to checking copyright  headers before a push

linusg:
This pre-commit config is quite simple: all it does for now is running the already existing `Meta/lint-ci.sh` script to ensure no changes that won't pass the CI linting step are committed.
Consequently `pass_filenames` is set to `false`, as these scripts will determine which files to check themselves - should this ever become a performance bottleneck, we can update them to accept an optional list of filenames.

It should be noted that using pre-commit for Serenity development is by no means a requirement, but doing so may decrease the number of CI builds failing due to forgotten license headers etc. - especially if you're pushing to `master` directly, rather than going through pull requests *hint hint*.

For info how to install & use pre-commit, see https://pre-commit.com/.

Fixes #2365.

---

Demo:

image

---

cc @alimpfard

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,8 @@
+repos:
+  - repo: local
+    hooks:
+      - id: meta-lint-ci
+        name: Running Meta/lint-ci.sh to ensure changes will pass linting on CI
+        entry: bash Meta/lint-ci.sh
+        language: system
+        pass_filenames: false


awesomekling:
👀

PR #4373, opened 2020-12-09
NotificationServer: Add a configurable notification timeout

metalvoidzz:


Services/NotificationServer/NotificationWindow.cpp
```diff
@@ -90,7 +91,8 @@ NotificationWindow::NotificationWindow(const String& text, const String& title,
     widget.layout()->set_margins({ 8, 8, 8, 8 });
     widget.layout()->set_spacing(6);
 
-    if (icon.is_valid()) {
+    auto config = Core::ConfigFile::get_for_system("NotificationServer");
+    if (icon.is_valid() && config->read_num_entry("Window", "ShowIcons", true)) {
```
linusg, line 95:
```suggestion
    if (icon.is_valid() && config->read_bool_entry("Window", "ShowIcons", true)) {
```

linusg (changes requested):

Services/NotificationServer/NotificationWindow.cpp
metalvoidzz: Thanks, overlooked that one

awesomekling:
This doesn't seem like a good default behavior to me.
If I went away from my computer for more than seconds, I still want to see my notifications when I return.

metalvoidzz:
Right, maybe we should have some kind of notification history applet, similar to clipboard history. It could change icons to tell you about unread notifications.

bcoles:
> 
> 
> Right, maybe we should have some kind of notification history applet, similar to clipboard history. It could change icons to tell you about unread notifications.

I like the idea of some sort of notification log. I also like the idea of having notifications disappear automatically.

Perhaps a notification applet icon could be displayed, with different icons depending on whether there are "new" notifications.

linusg:
@metalvoidzz do you still intend to make any changes here? Also, the PR has some merge conflicts that need to be resolved.

PR #4375, opened 2020-12-10
AK: Ensure dual_pivot_quick_sort does not copy the pivots

alimpfard:
Also add a test that would fail to compile if quick_sort tries to copy
anything :P

Sort of a dumb problem that showed up when I was trying to make `NonnullRefPtr<T>` refuse to be constructed from the result of `RefPtr<T>::operator*()` (which turned out to have more false positives than anything).

### DIFF
--- a/AK/Tests/TestQuickSort.cpp
+++ b/AK/Tests/TestQuickSort.cpp
@@ -1,0 +1,58 @@
+/*
+ * Copyright (c) 2020, the SerenityOS developers.
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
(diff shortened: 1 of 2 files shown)


ccapitalK:
Can you check that this doesn't break LibC's qsort? IIRC, that wraps AK's qsort in a kind of dodgy way, switching to rvalue references here may silently break it. If `/usr/Tests/LibC/qsort-sorts-and-copies` still passes with your patch, it should be fine.

alimpfard:
Seems to be passing.

AK/Tests/TestQuickSort.cpp
```diff
@@ -0,0 +1,58 @@
...
+#include <AK/Noncopyable.h>
+#include <AK/QuickSort.h>
+#include <AK/StdLibExtras.h>
+
+TEST_CASE(sorts_without_copy)
+{
+    struct NoCopy {
+        AK_MAKE_NONCOPYABLE(NoCopy);
+
+    public:
```
ccapitalK, line 39: Nit: not needed

ccapitalK (approved):

AK/Tests/TestQuickSort.cpp
alimpfard: Ah, no, `AK_MAKE_NONCOPYABLE` puts a `private:` at the top, see Noncopyable.h:30

ccapitalK: Oh, did not know that. Then yeah, disregard.

PR #4378, opened 2020-12-10
Shell: Restore the terminal PGID before printing out job status on exit

alimpfard:
> \< kling\> CxByte: i hav bug report for u
> \< kling\> CxByte: cat\<enter\>\<ctrl+c\>
> \< kling\> :)

This fixes the assert tripping when interrupting a foreground job.
Also make `bg` mark the job as 'should announce exit'.


ccapitalK:
This fixes it on my machine (also, it fixes a related issue where a program segfaulting in the foreground crashes the shell too).

ccapitalK (approved):

PR #4391, opened 2020-12-12
Spreadsheet: Reduce top bar default height

### LINKED ISSUE #3905: Spreadsheet: Oversized cell data input box
Hi Serenity land!

It's been a while but I've finally re-installed Ubuntu on my laptop and have qemu working again (and a new enough cmake) to test Serenity again! I broke it by installing qemu from source into a different prefix to what Ubuntu defaults too, then installing Ubuntu qemu at a later date or something.

I'm going to try to learn C first before I start contributing to Serenity but that doesn't mean I can't create more work for you all in the mean time!

Why is the spreadsheet cell data entry box so mahooosive?

Screenshot at 2020-11-01 19-42-36

linusg:
50px is a bit extreme, it's down to 26px high now. Still a bit larger than a regular `GUI::TextBox` but enough to look decent, even with the help button in there.

Closes #3905.

image

cc @alimpfard, does this look ok to you?


alimpfard:
It looks good, but now it needs a toolbar :]

PR #4395, opened 2020-12-12
AK::URL: Fix setting the port number in the case it was the last element of the URL

### LINKED ISSUE #4394: AK::URL construtor ignores the port if it is the last component of the URL
I noticed both in the Browser and on the pro command if a URL like http://example.com:8000/ will set the port to ``8000`` but http://example.com:8000 will get the default HTTP port ``80``. A pull request fixing the issue follows.

xspager:
Fixes #4394


awesomekling:
Great :)
You don't have to file a bug *and* a PR btw, just a PR is fine!

xspager:
Oh. Ok, will do next time Andreas.

PR #4407, opened 2020-12-13
LibTLS: Read subjectAltName from certificates and use it

alimpfard:
As quite a few certificates use this extension, reading and using it to
find matching certificates is fairly useful :^)


alimpfard:
_some day_ I will make `Crypto::ASN1` work, and _some day_, parsing these certificates will be much easier...
`++tech_debt;`

PR #4410, opened 2020-12-14
LibJS: Don't treat '?.' followed by decimal digit as QuestionMarkPeriod

### LINKED ISSUE #4409: LibJS: Question mark followed by a period always gets interpreted as QuestionMarkPeriod
This causes JQuery 1.12.4 minified to fail to parse: https://code.jquery.com/jquery-1.12.4.min.js
This is because it's using a ternary with a minified decimal number:
Snippet:
```js
?.01*parseFloat(RegExp.$1)+"":b?"1":""
```

linusg:
From the spec: https://tc39.es/ecma262/#sec-punctuators

    OptionalChainingPunctuator ::
        ?. [lookahead ∉ DecimalDigit]

We were missing the lookahead and therefore incorrectly treating any '?.' as `TokenType::QuestionMarkPeriod`.

Fixes #4409.

### DIFF
--- a/Libraries/LibJS/Lexer.cpp
+++ b/Libraries/LibJS/Lexer.cpp
@@ -598,10 +598,13 @@ Token Lexer::next()
             auto two_chars_view = m_source.substring_view(m_position - 1, 2);
             auto it = s_two_char_tokens.find(two_chars_view.hash(), [&](auto& entry) { return entry.key == two_chars_view; });
             if (it != s_two_char_tokens.end()) {
-                found_two_char_token = true;
-                consume();
-                consume();
-                token_type = it->value;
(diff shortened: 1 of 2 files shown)


alimpfard:
Is this an instance of the infamous shell-backgrounding failure? or is it something else?
maybe we should disable that test :thinking:

linusg:
holy sh*t I didn't even notice this never finished - 4h 0m 50s and counting, amazing :smile:

alimpfard:
why doesn't GA show the history of running jobs? that's so silly...
I think, for the time being, I'll just set a timeout on that test. failing fast is better than deadlocking for the rest of eternity :shrugging:

PR #4415, opened 2020-12-14
Shell: Make Subshell actually create a subshell / Shell: Set a timeout on Shell-backgrounding

alimpfard:
-> https://github.com/SerenityOS/serenity/pull/4410#issuecomment-744592981

Also makes `{ whatever }&` actually work.


alimpfard:
:sad:
when you set a timeout and even _that_ doesn't work

PR #4416, opened 2020-12-14
LibWeb: Add a simple StyleInvalidator class 

### LINKED ISSUE #4404: LibWeb: Setting an elements 'className' doesn't trigger style update of related elements
This is somewhat similar to #4403, but the provided fix doesn't help in this case.

Repro:

```html
<style>
.blue {
    color: blue;
}
.blue + .red {
    color: red;
}
</style>

<div>Blue</div>
<div class="red">Red</div>

<script>
setTimeout(() => {
    document.querySelector("div").className = "blue";
}, 1000);
</script>
```

Expected:

Both "Blue" and "Red" don't have color, but turn blue and red respectively after one second.

Actual:

Both "Blue" and "Red" don't have color, only "Blue" turns blue after one second. "Red" only changes its color when interacting with it.

---

The same applies to non-"class" attributes:

```html
<style>
[foo] {
    color: blue;
}
[foo] + [bar] {
    color: red;
}
</style>

<div>foo</div>
<div bar>bar</div>

<script>
setTimeout(() => {
    document.querySelector("div").setAttribute("foo", 123);
}, 1000);
</script>
```

In that case neither of the divs update until they're being hovered over.

linusg:
This patch adds a simple, naive & inefficient class for document-wide
style invalidation, e.g. after element attribute updates. During
construction it collects a HashMap of a document's elements and their
matching rules, during destruction it does the same and then compares
the results; dirtying all elements that have a different number or order
of matching rules afterwards.

Much room for improvement, but it solves the problem of stale element
styling after attribute updates for now :^)

Fixes #4404.

### DIFF
--- a/Libraries/LibWeb/CMakeLists.txt
+++ b/Libraries/LibWeb/CMakeLists.txt
@@ -23,6 +23,7 @@ set(SOURCES
     CSS/Selector.cpp
     CSS/SelectorEngine.cpp
     CSS/StyleDeclaration.cpp
+    CSS/StyleInvalidator.cpp
     CSS/StyleProperties.cpp
     CSS/StyleResolver.cpp
     CSS/StyleRule.cpp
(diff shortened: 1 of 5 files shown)


awesomekling:
Excellent! I love this kind of PR, it fixes a problem we have right now, *and* gives us something to optimize in the future :smile:

linusg:
spotted a small typo, now fixed.

PR #4423, opened 2020-12-15
EchoServer: Add a simple echo server

bcoles:
Maybe this will be useful for debugging networking / socket issues.

We already have `nc` netcat, but it doesn't support multiple concurrent clients and quits after the first client connection. `EchoServer` also conveniently prints input to console with `dbg()`, although perhaps we'll want to wrap this in a `#ifdef ECHOSERVER_DEBUG` later, maybe.

This code was largely stolen from kling's `TelnetServer`.

This server has issues echoing data when the socket input buffer exceeds the buffer size (1024). So don't do that.

The [RFC](https://tools.ietf.org/html/rfc862) doesn't offer any guidance on dealing with buffering. I looked at [some other implementations](https://rosettacode.org/wiki/Echo_server) which didn't offer much insight either.

EchoServer -p 1234

man EchoServer


Services/EchoServer/main.cpp
```diff
@@ -0,0 +1,104 @@
...
+    while ((opt = getopt(argc, argv, "p:")) != -1) {
+        switch (opt) {
+        case 'p':
+            port = atoi(optarg);
+            break;
+        default:
+            fprintf(stderr, "Usage: %s [-p port]", argv[0]);
+            exit(1);
+        }
+    }
```
alimpfard, line 65: `Core::ArgsParser`?

bugaevc: Yes please :slightly_smiling_face:

bcoles: done

monkey see, monkey do

https://github.com/SerenityOS/serenity/blob/eb7a94364188740d67278c74bc17d4223697becd/Services/TelnetServer/main.cpp#L114-L126

awesomekling:
Sure why not :)

Issue #4081, opened 2020-11-14
Userland: chroot executes only supplied program, not program with arguments


bcoles:
`chroot` executes only supplied program, not program with arguments.

Use case:

By default, chrooted environments have access to all environment variables available to the parent process.

![chroot](https://user-images.githubusercontent.com/434827/99145985-d93e0580-26c7-11eb-8a64-4cd6ef3e86ac.png)

This is standard default behavior for `chroot` (as opposed to [schroot](https://linux.die.net/man/1/schroot)). For example, on Ubuntu 20.04:

```
root@ubuntu:/home/user/Desktop/serenity# export SECRET=123
root@ubuntu:/home/user/Desktop/serenity# chroot / /bin/sh
# echo $SECRET
123
# 
root@ubuntu:/home/user/Desktop/serenity# chroot --userspec 1000:1000 / /bin/sh
$ echo $SECRET
123
$ 
```

Usually, to prevent leaking environment variables, the operator will invoke `chroot` with `env -i` to clear environment variables before executing the program / shell.

```
root@ubuntu:/home/user/Desktop/serenity# export SECRET=123
root@ubuntu:/home/user/Desktop/serenity# chroot --userspec 1000:1000 / env -i /bin/sh
$ echo $SECRET

$ 
```

This isn't possible with `chroot` on Serenity OS, as `chroot` accepts only a program to execute and does not accept program arguments.

Unfortunately, implementing this requires rewriting the option parsing, as `ArgParse` does not play nicely with positional arguments beginning with `-`, making this somewhat tedious to implement. ~~Also, env does not support `-i`.~~ (#4064) A few other utilities also have the same issue with arguments (ie, `profile`).

Alternatively, all the arguments could be quoted and shoved into a string as one argument, but that's hackish and is not in line with other implementations. Similarly, the optional positional arguments could be separated with ` -- `  and pre-processed before `args_parser.parse`, but that's also hackish and not in line with other implementations.

Test chroot:

```diff
diff --git a/Meta/build-root-filesystem.sh b/Meta/build-root-filesystem.sh
index e2a9addfe..c768f05ef 100755
--- a/Meta/build-root-filesystem.sh
+++ b/Meta/build-root-filesystem.sh
@@ -160,6 +160,27 @@ ln -s Shell mnt/bin/sh
 ln -s test mnt/bin/[
 echo "done"
 
+
+printf "creating (insecure) chroot... "
+mkdir mnt/var/chroot/
[...]

PR #4431, opened 2020-12-16
TelnetServer: replace getopt with LibCore ArgsParser

bcoles:


bugaevc (approved):
👍

It would be better to accept a sequence of command arguments (so, `Vector<const char*>`), but that's orthogonal

bcoles:
> 
> 
> +1
> 
> It would be better to accept a sequence of command arguments (so, `Vector<const char*>`), but that's orthogonal

Related: #4081

PR #4433, opened 2020-12-16
Userland: Add beep utility

bcoles:
beep

On my system, both unistd `sysbeep()` and `SC_beep` syscall only create an audible beep when invoked maybe one in 5 times.

If no audible beep is emitted, try opening `Piano`. The beep will play when `Piano` opens.

This issue may be caused by qemu and/or the underlying Ubuntu host OS.

Fortunately, we now have a handy `beep` utility to help debug this!

### DIFF
--- a/Base/usr/share/man/man1/beep.md
+++ b/Base/usr/share/man/man1/beep.md
@@ -1,0 +1,19 @@
+## Name
+
+beep - beep the pc speaker
+
+## Synopsis
+
+```sh
+$ beep
+```
+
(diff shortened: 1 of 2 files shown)


alimpfard:
10/10 manpage, here's what I saw at first glace:

## Beep

beep - beep the PC beep

## Synopsis

```sh
$ beep
```

## Description

beep beeps the beep to beep the PC beep.

## Examples

```sh
$ beep
```

PR #4434, opened 2020-12-16
Shell: Don't run commands with failing redirections

### LINKED ISSUE #3423: Shell: Commands with failing redirections should not be run
_As seen on TV_

```sh
$ echo foobar > /
error: blah blah
foobar
```

linusg:
Fixes #3423.

---

cc @alimpfard - this seems too easy to be true, but it also makes sense???


alimpfard:
That...looks right...
:facepalm:

PR #4436, opened 2020-12-16
Kernel: Fix Lock race causing infinite spinning between two threads

tomuta:
We need to account for how many shared lock instances the current
thread owns, so that we can properly release such references when
yielding execution.

We also need to release the process lock when donating.

I observed this bug with FileManager where two threads sometimes end up wasting all CPU (without locking up the entire system).


tomuta:
Hm just found that strace triggers an assertion...

tomuta:
Oh wow that was easy, just a bad assertion that shouldn't be there...

PR #4442, opened 2020-12-17
LibWeb: Include element attributes in innerHTML getter

linusg:


Libraries/LibWeb/DOM/Element.cpp
```diff
@@ -276,14 +276,24 @@ String Element::inner_html() const
...
+                auto& element = downcast<Element>(*child);
                 builder.append('<');
-                builder.append(downcast<Element>(*child).local_name());
+                builder.append(element.local_name());
+                element.for_each_attribute([&](auto& name, auto& value) {
+                    builder.append(' ');
+                    builder.append(name);
+                    builder.append('=');
+                    builder.append('"');
+                    builder.append(value);
```
awesomekling, line 287: Hmm,  do we need to do some escaping here?

linusg: You tell me! :smile:

This is copied from
https://github.com/SerenityOS/serenity/blob/169a9150cbf9d8e4402b223209c80a0d4007389f/Libraries%2FLibWeb%2FLayoutTreeModel.cpp#L142-L149

awesomekling: That's just for showing in the GUI though, incorrect output is kinda harmless :D

Here's the HTML fragment serialization algorithm:
https://html.spec.whatwg.org/multipage/parsing.html#html-fragment-serialisation-algorithm

So it looks like we need to escape characters like this (in the *attribute mode*):
https://html.spec.whatwg.org/multipage/parsing.html#escapingString

awesomekling:
Beautiful!

PR #4456, opened 2020-12-18
Kernel: Run clang-format on PCI/Definitions.h

supercomputer7:


supercomputer7:
This was a problem with my `clang-format` apparently.

PR #4459, opened 2020-12-19
AK: Generalize AK::String::to_int() for more types

### LINKED ISSUE #4377: Feature request: AK: Convert string to long long
It would be nice for there to be an equivalent of `String::to_int()` for 64 bit types.

ccapitalK:
Fixes #4377.

### DIFF
--- a/AK/FlyString.cpp
+++ b/AK/FlyString.cpp
@@ -85,10 +85,27 @@ FlyString::FlyString(const char* string)
 {
 }
 
-Optional<int> FlyString::to_int() const
+template<typename T>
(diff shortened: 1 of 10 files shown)


awesomekling:
Please add some unit tests under `AK/Tests/` so we know these work correctly :)

ccapitalK:
I've added tests and overflow checking, which broke the JsonParser test because the 4chan catalog test file had some integers that were larger than 32bits (previously, the parser stored garbage values there, but it wasn't checked so it was fine). I've partially fixed it, but eventually someone will have to write a proper json number parser.

PR #4461, opened 2020-12-19
MenuApplets: Add Network menu applet

bcoles:
Add a simple network applet.

It currently doesn't do much, but I think it's cute and makes the system feel more like a real OS.

Network.MenuApplet


awesomekling:
That's hecka cute! 🐱

bcoles:
A few bugs:

Sometimes upon boot the widget is loaded to left of the clock. I'm not sure why. I think this is a "bug" in the widget loading priority order, regardless of the order specified in the `SystemServer.ini` configuration.

The icon loading is currently very inefficient (`load_from_file` upon every timer increment (`5` seconds)). I looked into caching this once on load, but `ImageWidget` does not appear to expose a function to set the icon directly as a Bitmap, only from file. Unless I'm missing something. I've added it to my list of things to review.

More of a missing feature than a bug: handling of multiple adapters could be improved (optionally showing more than one widget per adapter). Given we currently only support one adapter, I'm ignoring this for now.

More of a missing feature than a bug: there's no right click context menu and clicking the widget pops the SystemManager - nothing related to network config, as we don't have a GUI app for network management. Given that it runs as the current user, and we don't have support for capabilities (`cap_net_admin`), there's not much value in adding something like this for now.

diversys:
It could show notifications on link state change.

diversys:
Also the monitor icon could change its color slightly to indicate network activity. It is done on Haiku with arrows in this 3d-party app:

image


https://github.com/HaikuArchives/NetPulse

bcoles:
> 
> 
> It could show notifications on link state change.

I've added simple notifications with a flag `--display-notifications` to toggle notifications on and off. Unfortunately, notifications aren't automatically dismissed (yet - #4373). 

Network notifications

PR #4462, opened 2020-12-19
Kernel: New Storage System #1

supercomputer7:
This PR includes many fixes and changes.

2 goals that this PR achieved for now:
1. To use drives on the secondary IDE channel. For example, if we connect a drive as Secondary-Master,
We could run `mount /dev/hdc /mnt` to mount that drive.

2. To boot with other drives on the PC. Simply specify the correct letter for the drive,
for example - `root=/dev/hdb` will boot the second connected harddrive*.

*Side note - the indexing scheme is very simple for now. Ideally all 4 connectors are connected to IDE drives so `hda` is the first drive, `hdb` is the second and so on... in case of missing drive, we simply give that name to the next available, so if the Primary-Slave drive is not connected, then Secondary-Master drive will be `hdb`, for example.

This PR lays down the foundations to my next goal - supporting NVMe.

Any suggestion or comment is highly appreciated :)


tomuta:
You beat me to it ;-) Just last week I even set up a vmware configuration for nvme ;-)

PR #4467, opened 2020-12-20
Kernel: Some futex improvements

tomuta:
Kernel: Some futex improvements

This adds support for FUTEX_WAKE_OP, FUTEX_WAIT_BITSET, FUTEX_WAKE_BITSET,
FUTEX_REQUEUE, and FUTEX_CMP_REQUEUE, as well well as global and private
futex and absolute/relative timeouts against the appropriate clock. This
also changes the implementation so that kernel resources are only used when
a thread is blocked on a futex.

Global futexes are implemented as offsets in VMObjects, so that different
processes can share a futex against the same VMObject despite potentially
being mapped at different virtual addresses.


Libraries/LibC/serenity.cpp
```diff
@@ -72,10 +72,38 @@ int set_process_boost(pid_t tid, int amount)
...
 
-int futex(int32_t* userspace_address, int futex_op, int32_t value, const struct timespec* timeout)
-{
-    Syscall::SC_futex_params params { userspace_address, futex_op, value, timeout };
-    int rc = syscall(SC_futex, &params);
+int futex(uint32_t* userspace_address, int futex_op, uint32_t value, const struct timespec* timeout, uint32_t* userspace_address2, uint32_t value3)
+{
+    int rc;
+    switch (futex_op & FUTEX_CMD_MASK) {
+    //case FUTEX_CMP_REQUEUE:
```
bgianfo, line 79: dead code?

tomuta: How so?

bugaevc:
Yay!

I still  haven't understood how blocking is working / supposed to work (in particular I don't think `m_wake_requested` can work reliably...), but if it appears to be working for you, awesome.

Can we do one thing different than Linux though? I just recently learned that on Darwin, `ulock_wait()` actually returns if there were other waiters still queued when you wake up. Initially I thought that was useless because new waiters can arrive at any point, so the info wouldn't be relevant by the time you get to use it, but then I understood how it can be used. Can we do the same and have `futex_wait()` return, upon a successful wake-up, the number of threads still sleeping?

tomuta:
`m_wake_requested` was taken form the `WaitQueue` implementation. To be honest, due to the lack of stuff exercising futexes I haven't really tested this.

As to the enhancement, yes we *could*, but I'm not sure how that is really useful because there is no mechanism to relay that information back to user mode. We do return if the queue is now empty after waking someone, which is used to delete the futex on the kernel side so it doesn't permanently occupy resources, after all the kernel doesn't know if a futex was permanently deleted.

bugaevc:
> `m_wake_requested` was taken from the `WaitQueue` implementation.

Yes, I get that. I don't think `m_wake_requested` can work reliably *at all*, not just for futexes — from my experience with userspace synchronization mechanisms, a single boolean is not enough for what is essentially a condition variable, where the condition being awaited can flip between being true and false a few times while you're preparing to sleep. Perhaps the new `WaitQueue`s / blockers have solved this issue in some way that I'm not seeing; that's why I'm saying I haven't figured out how it's supposed to work yet. Again, this is not specific to futexes, it's about me not-yet-understanding this new synchronization mechanisms in general :slightly_smiling_face:

tomuta:
> > `m_wake_requested` was taken from the `WaitQueue` implementation.
> 
> Yes, I get that. I don't think `m_wake_requested` can work reliably _at all_, not just for futexes — from my experience with userspace synchronization mechanisms, a single boolean is not enough for what is essentially a condition variable, where the condition being awaited can flip between being true and false a few times while you're preparing to sleep. Perhaps the new `WaitQueue`s / blockers have solved this issue in some way that I'm not seeing; that's why I'm saying I haven't figured out how it's supposed to work yet. Again, this is not specific to futexes, it's about me not-yet-understanding this new synchronization mechanisms in general slightly_smiling_face

Sorry I think I mixed that up with the bool in the blockers. The `m_wake_requested` is different, this means that if the queue is requested to wake a thread and if no thread is currently waiting, the next time a thread wants to wait it is told that it shouldn't wait immediately. This is absolutely necessary, actually not doing so would reintroduce bug 2693. Once a condition to wake waiters is triggered, there generally is no way to take it back. At least this is how other OSs including Windows and Linux work. It's a one-shot, once you wake a blocking condition, there is no way going back, though it's possible you may attempt to wake it more than once before it has a chance to run again.

bugaevc:
Oh, it's one-shot?

Then I don't think I understand at all why it is in fact necessary and how it's supposed to work. The `WaitQueue` itself is not one-shot, right? (or is it??)

In my understanding, most blocking in the kernel should behave like condition variables: we're waiting for some condition (say, an fd becoming readable), and the condition can become true, then false, then true again, and so on. So we atomically check that the condition is false and go to sleep, then someone (presumably the device driver if we were reading from a device) notifies us, then (some time later) we wake up, recheck the condition (it might have become false since then!) and if it is true now, we proceed, otherwise we go back to sleep and wait for someone to notify us again.

Two important points in this design:
* We must check the condition and go to sleep (or not) atomically, otherwise the condition can change to true before we go to sleep *and* we can miss a wake-up, which results in a hang.
* Somebody else notifying us does not directly cause us to stop sleeping, it only tells us to briefly wake and recheck. If we we find the condition to be false, we continue sleeping. Otherwise, yes, we stop sleeping and go on doing something useful.

Is that not how your blocking mechanisms work?

tomuta:
Yes, waking threads is a one-shot, there's no way to take it back. This particular feature is essentially preventing a race condition where let's say a driver uses a `WaitQueue` to have a thread wait until something happens. Let's say that "something happens" happens before another thread has a chance to wait for it, then that thread would add itself to the `WaitQueue` but because that event already happened, it would just remain blocked without ever being woken. Instead, because the `WaitQueue` is empty when that "something happens" occurs, it flags `m_wake_requested` so that next time a thread is trying to add itself, it is rejecting that and the thread knows that the "something happens" already happened. This basically fulfills the concept of triggering an unblock of at least one thread after a wake event.

The `WaitQueue` isn't really the same as `Thread::BlockCondition`, which is what most blocking is using. The `WaitQueue` is more simplistic, and maybe someday we could even retire it (maybe). For more complicated waits, such as waiting on processes, on sockets, files etc we now use `Thread::BlockCondition`, which implements your two points. The entire OS has switched to this model and I think it's working quite well since fixing some of the glaring problems/bugs.

tomuta:
Another important thing to remember is that condition variables and kernel blocks are not quite comparable for this one reason: condition variables may spuriously unblock, but kernel block conditions may never spuriously unblock. This is because if you wait on a block condition (e.g. using `select`), using a subsequent API to perform a potentially blocking operation (e.g. `read`) _must never_ block unexpectedly.

bugaevc:
Ah, so it sounds like `WaitQueue` is more like what I'd call *an event* (and not a *condition variable*), or what Linux calls *completions*, and a `Thread::BlockCondition` is what I should be focusing on.

But then we have to make very sure we're only using `WaitQueue`s for one-shot things. And it seems that we don't; for example I see the SB16 driver is apparently using it to wait for IRQ?

https://github.com/SerenityOS/serenity/blob/ba3488a6d521bbb2ee032f00c55b6f922c7e5fe6/Kernel/Devices/SB16.h#L76

Or the lock is using a queue to wait for someone unblocking it

https://github.com/SerenityOS/serenity/blob/c4176b0da1000d1cba663fc5cb0f34efb1df394a/Kernel/Lock.cpp#L126

Those don't seem one-shot...

> Meaning if select told you that you can read without blocking, then the first call to read must absolutely never block.

But that just doesn't work that way: by the time you wake up and attempt to read, someone else might have read the available bytes. Only way this can work is if you treat each wake-up as potentially spurious, and recheck the current state each time (and atomically go to sleep if that's not the state you wanted).

tomuta:
The SB16 driver specifically is severely broken. And no, it should not use a `WaitQueue`. Have a look at #4339 where I made some improvements (I could use some help figuring out why the DMA transfer locks up the system), but it's basically converting it to the `AsyncDeviceRequest` model that I converted the PATA driver to. Which does the signaling and blocking for you for the most part.

The reason for the loop in `Lock` is that if an unblock had been flagged, it's possible that there is a race condition where another thread may have grabbed the lock before us, so we need to loop until we win the race or we go to sleep again.

The guarantee for `select` and `read` is accurate, though. The guaranteed non-blocking nature of the subsequent `read` only applies to whoever did the first call to `read` after the `select`. In a single-threaded program this is generally easy, in case of a multi-threaded program care must be taken to track that fact in user mode, unless the code is designed to only have the same thread do the `read` who did the `select` prior to it. Duplicating a file descriptor makes this more complicated as well and that would have to be tracked/accounted for in user mode. This is really a core principle in order to make the posix APIs work the way they do, and Windows actually uses a very similar concept with pretty much the same guarantees (just different API scheme around it).

tomuta:
Of course `WNOHANG` is always a good idea when writing asynchronous software ;-)

bugaevc:
> The SB16 driver specifically is severely broken. And no, it should not use a `WaitQueue`.

Ah, I see.

> The reason for the loop in `Lock` is that if an unblock had been flagged, it's possible that there is a race condition where another thread may have grabbed the lock before us, so we need to loop until we win the race or we go to sleep again.

Yes, sure, I understand the reason for looping. What I don't understand is how it's OK to use a `WaitQueue` there for blocking, if it is, as you've just told me, one-shot. That would mean that once anyone is woken for the very first time in each given lock, any further attempts to sleep on the queue unblock immediately, no?

Moreover, I see that `m_wake_requested` can be reset back to `false` once it has been set to `true`:

https://github.com/SerenityOS/serenity/blob/780c64e1f09bf7400b94aab594e1993ac97e0557/Kernel/WaitQueue.cpp#L71

https://github.com/SerenityOS/serenity/blob/780c64e1f09bf7400b94aab594e1993ac97e0557/Kernel/WaitQueue.cpp#L39-L40

that can't possibly work... or I'm misunderstanding something big time.

> The guarantee for `select` and `read` is accurate, though.

Yes, in a single-threaded userspace process, sure (assuming no other processes/threads have the same fd). What I was talking about us that if you're implementing a blocking `sys$read()` in the Kernel, you cannot just wait *once*, you have to wait in a loop, because another thread may "steal" the data from you at any point. And since you have to use a loop anyway, the waiting primitive might as well do spurious wake-ups. Of course, none of this is visible to userspace; but userspace deals with very similar issues in single producer / multi consumer scenarios.

tomuta:
Yes, `m_wake_requested` will be reset to `false` if at least one thread was woken or attempted to block, that's intended. It means that the original unblock-condition was finally satisfied/"waited-for" by one thread either attempting to block or having been unblocked. `Thread::BlockCondition` allows more fine grained control over this, but the principle remains largely the same. `WaitQueue::should_add_blocker` is called when a thread attempts to block, which essentially translates to the old `WaitQueue`'s implementation here:
https://github.com/SerenityOS/serenity/blob/64cc3f51d0a1d4ddc68fe9e3b5da2834eac48dae/Kernel/WaitQueue.cpp#L48

In regards to
https://github.com/SerenityOS/serenity/blob/780c64e1f09bf7400b94aab594e1993ac97e0557/Kernel/WaitQueue.cpp#L39-L40
The purpose is to tell the `Thread::BlockCondition` that the first request to consume the wake-event is about to to consume this fact (which again translates to the quoted code from the old implementation above). Note that in this case `should_add_blocker` returns `false`, telling it to abort the block attempt and continue immediately, which returns `NotBlocked` from `Thread::block`. Since no two threads can signal a `Thread::BlockCondition` and trying to do a `Thread::block` on that same `Thread::BlockCondition` can happen simultaneously, there isn't a possibility for a race. Either it blocks and gets unblocked shortly after, or the `Thread::BlockCondition` is marked pending an unblock and the first thread trying to block after will be told not to block.

In response to your blocking `sys$read`, yes that could be a race and yes, there probably needs to be a loop. I haven't looked at that but I think your concern is legitimate. Right now it probably relies on the `Process`' big lock...

tomuta:
Note that such a loop is only necessary for when the `Thread::BlockCondition` doesn't maintain the data. It is for example not necessary for `waitpid` or joining threads, as the `Thread::BlockCondition` maintains all data, so there is no way to steal it through another channel.

tomuta:
Another approach could be imposing a lock on that data for the thread that was signalled, so that only that thread is able to read the data and anyone attempting to steal it would have to wait until the `sys$read` is done.

tomuta:
Or the `Thread::BlockCondition` could just hand the available data (or however much is requested) over to the thread that is being unblocked, that way it can't be stolen and there's no need for a loop.

bugaevc:
> Yes, `m_wake_requested` will be reset to false if at least one thread was woken or attempted to block, that's intended. It means that the original unblock-condition was finally satisfied/"waited-for" by one thread either attempting to block or having been unblocked.

At this point in time, I'm still not convinced that there can be a middle ground behind an "event" style primitive (one-shot, either already happened/completed or not, stores the state internally), and a "condition variable" style primitive (a condition is stored/evaluated externally to the primitive, and it can flip between being true/false, consumed/unconsumed repeatedly). I know that other implementations support reusable events, e.g. Linux has `reinit_completion()`, but I don't see how it can work.

See for yourself: if a condition we're waiting for can once again become false after becoming true (say, due to some other thread consuming it), that means we cannot assume that after we've been awoken, the condition will be true — exactly because some other thread can run in and consume/reset it while we're waiting/checking. So we need to loop, and we need some locking to be able to check & act on the condition without someone else stealing it from us in the middle. And we also want to release this lock and go to sleep atomically, in order not to miss any wake-ups. So we get a condition variable.

Actually, hmm, while I was writing that I got an idea of what you might be getting at. 🤔 If we don't need locking to check & act on condition — if, say, we just do an atomic pointer exchange and then act on the pointer if it's non-null — then we might be able to get away without using a lock at all. But we need to be very careful around lost wake-ups & stuff. A single boolean can't work for this purpose because:
* it's not even atomic! it's literally UB to do what `WaitQueue` does
* it can't differentiate between the previous wake intending to wake *one* thread and waking *everyone* — in both cases the waiting threads are there to process events of some sort, but in one case the events need to be *drained* or *consumed*, so it's fine to wake just one worker thread, and in the other case events are just announced/broadcasted, and we have to make sure everyone is woken / not put to sleep

I definitely need to think more about this.

> In response to your blocking sys$read, yes that could be a race and yes, there probably needs to be a loop. I haven't looked at that but I think your concern is legitimate. Right now it probably relies on the Process' big lock...

Right, but the big lock is only one process-wide; if different threads from different processes read from the same file simultaneously, we'll get in a weird state. The same logic *probably* applies to most other places we're blocking in the kernel...

> Another approach could be imposing a lock on that data for the thread that was signalled, so that only that thread is able to read the data and anyone attempting to steal it would have to wait until the `sys$read` is done.

Yes, that's how condition variables work

(By the way, if you haven't seen it yet, I have [a rather straightforward implementation](https://github.com/bugaevc/lets-write-sync-primitives/blob/master/src/condvar.cpp) of a userspace condition variable in LWSP that I hope to later upstream to Serenity — and `FUTEX_REQUEUE` is one of the things I need for that 😀)

> Or the `Thread::BlockCondition` could just hand the available data (or however much is requested) over to the thread that is being unblocked, that way it can't be stolen and there's no need for a loop.

You're thinking of a "channel" primitive, where you push stuff at one end and some worker thread picks it up on another thread. That only works when we know in advance how much data the given consumer thread is going to consume. E.g. with `sys$read()` we probably do, but it's easy to imagine cases where the consumer decides how much it's going to consume only when/after it sees the data.

tomuta:
> At this point in time, I'm still not convinced that there can be a middle ground behind an "event" style primitive (one-shot, either already happened/completed or not, stores the state internally), and a "condition variable" style primitive (a condition is stored/evaluated externally to the primitive, and it can flip between being true/false, consumed/unconsumed repeatedly). I know that other implementations support reusable events, e.g. Linux has `reinit_completion()`, but I don't see how it can work.
`Thread::BlockCondition` allows you to implement such things. Should be fairly straight forward.
 
> See for yourself: if a condition we're waiting for can once again become false after becoming true (say, due to some other thread consuming it), that means we cannot assume that after we've been awoken, the condition will be true — exactly because some other thread can run in and consume/reset it while we're waiting/checking. So we need to loop, and we need some locking to be able to check & act on the condition without someone else stealing it from us in the middle. And we also want to release this lock and go to sleep atomically, in order not to miss any wake-ups. So we get a condition variable.

I assume you're talking about `sys$read` here? If so, as I mentioned the loop can be avoided by passing the consumable data to the `Thread::BlockCondition`, which can pass it on to exactly one thread that is being woken. There would be no window to steal it in that case, and no loop would be necessary. I have not looked at `sys$read`, so your point of possibly having a race condition in this regard is valid and I'm going to look at it at some point (if I don't forget). This is really only a problem with the file/socket based waits where the data is disjoint from the wait condition, but it should be solvable relatively easily. Other blocks, such as waiting on processes or joining threads don't have this problem.

But conceptually, once a state change triggers an unblock, there is no way to revert that fact. Some threads may unblock (how many and which ones depends on the particular `Thread::BlockCondition` implementation.

> Actually, hmm, while I was writing that I got an idea of what you might be getting at. thinking If we don't need locking to check & act on condition — if, say, we just do an atomic pointer exchange and then act on the pointer if it's non-null — then we might be able to get away without using a lock at all. But we need to be very careful around lost wake-ups & stuff. A single boolean can't work for this purpose because:
> 
>     * it's not even atomic! it's literally UB to do what `WaitQueue` does
> 
>     * it can't differentiate between the previous wake intending to wake _one_ thread and waking _everyone_ — in both cases the waiting threads are there to process events of some sort, but in one case the events need to be _drained_ or _consumed_, so it's fine to wake just one worker thread, and in the other case events are just announced/broadcasted, and we have to make sure everyone is woken / not put to sleep

Well, that's where I'm getting it with there is some room for improvement. Especially with your `sys$read` example, I agree, you brought up a valid point for a window where another thread could steal the data. But that shouldn't be too difficult to close. In that particular case, the ThreadBlockCondition would just wake one thread, hand it the data it wants, and if anything is left over, wake the next etc until no data is left. Anyone else (if any) remains blocked. On the other hand, a `select` is more like a broadcast, but the guarantee still remains. If `select` states that a socket can be read from without blocking, the first call to `read` for that socket must never block. If anyone else is able to "steal" that data then this would be bad user mode code as it wouldn't account for that fact properly.

> Right, but the big lock is only one process-wide; if different threads from different processes read from the same file simultaneously, we'll get in a weird state. The same logic _probably_ applies to most other places we're blocking in the kernel...

I believe other OS such as Linux and Windows pose the same problem if they are shared among multiple processes, or if file descriptors are duplicated and shared with other ones. The fix for `sys$read` however would close a bug in that case where it would unblock but not have any data (there isn't even a way to relay that properly as e.g. returning `0` for a socket would indicate that the connection was closed. Again, I think there is a valid flaw but it should be fixable with the current framework.

> (By the way, if you haven't seen it yet, I have [a rather straightforward implementation](https://github.com/bugaevc/lets-write-sync-primitives/blob/master/src/condvar.cpp) of a userspace condition variable in LWSP that I hope to later upstream to Serenity — and `FUTEX_REQUEUE` is one of the things I need for that grinning)

I'll get to `FUTEX_REQUEUE` soon, thanks for the code, it'd actually be pretty awesome if you yould implement those bits and pieces into our pthread stuff so I have something to test against ;)

> You're thinking of a "channel" primitive, where you push stuff at one end and some worker thread picks it up on another end. That only works when we know in advance how much data the given consumer thread is going to consume. E.g. with `sys$read()` we probably do, but it's easy to imagine cases where the consumer decides how much it's going to consume only when/after it sees the data.

Correct, that would be purely for `read` but not for `select`. `Thread::BlockCondition` can only hand over data if it knows how much the consumer wants. In the `select` case it knows that there is no desire to consume anything, only to be notified that data could be consumed.

tomuta:
Btw, take a look at `FUTEX_WAKE_OP` which is designed for supporting condition variables. I haven't tested the implementation in this PR yet, so no idea if it works as intended...

tomuta:
Ok, added `FUTEX_REQUEUE` and `FUTEX_CMP_REQUEUE`. Now comes the fun part, testing if it actually works. I'm sure there's bugs all over... ;-)

bugaevc:
> I assume you're talking about `sys$read` here?

Not just `sys$read`, but presumably most of the places where we block in the Kernel. We're waiting for *something* (data, network packet, dead child process, lock) to become available, but then before we can act on it, someone else can front-run us and steal it from us. So we recheck and to back to sleep, etc.

> If so, as I mentioned the loop can be avoided by passing the consumable data to the `Thread::BlockCondition`, which can pass it on to exactly one thread that is being woken. There would be no window to steal it in that case, and no loop would be necessary.

Right, you have a point — perhaps in many (most?) of those cases we can just hand over the thing we're waiting for to a particular thread directly. Then, no need to check and to loop.

> But conceptually, once a state change triggers an unblock, there is no way to revert that fact.

Well, I'm not saying we have to revert that; the threads wake, and then (in a condition variable model) recheck the state, and voluntarily go back to sleep if what they were waiting for is still not there (either stolen by another thread, or it just was a spurious wake-up). But indeed we might not need that with a channel-like primitive.

> I believe other OS such as Linux and Windows pose the same problem if they are shared among multiple processes, or if file descriptors are duplicated and shared with other ones. The fix for `sys$read` however would close a bug in that case where it would unblock but not have any data (there isn't even a way to relay that properly as e.g. returning 0 for a socket would indicate that the connection was closed.

Not, that's certainly no problem with this in Linux. If another thread steals the data we were about to read, but more data might come later, `read()` goes to sleep again — it only returns to userspace when it has either read some data, on EOF, or on error.

Reading from a pipe is implemented in Linux with a loop, which uses locking and `wait_event*()` (essentially a condition variable), much like I'm saying: https://github.com/torvalds/linux/blob/8653b778e454a7708847aeafe689bce07aeeb94e/fs/pipe.c#L362

> In the `select` case it knows that there is no desire to consume anything, only to be notified that data could be consumed.

Yeah, though I'm guessing in the `sys$select()` case it's fine to return to userspace even if another thread concurrently reads all the available data; it's up to userspace to try, receive `-EWOULDBLOCK`, and repeat. Although it would still be faster to detect this in kernel space when possible, to avoid the whole roundtrip.

Anyways, today I'm going to try taking a proper look at `Thread::BlockCondition`, perhaps it'll just make sense :)

tomuta:
The problem with Linux I was mentioning was not referring to `read` but the combination of `select` and `read`. But yes, if we fix `read` as mentioned then we should behave the same as Linux, which should definitely be possible. Waiting on processes or joining threads does not have this problem, because the `Thread::WaitBlockCondition` already receives all the data associated with it so this race doesn't exist for that case. It's really only the `Thread::FileDescriptionBlocker` based blockers in combination with `read` or `write` where the data is disjoint. Essentially, I think we just have to pass data into `FileBlockCondition::unblock` and pull available data in `FileBlockCondition::should_add_blocker`. Anyway, a project for another day ;)

tomuta:
Ok I switched the pthread mutex and condvar code over to use futexes. Seems to work, but haven't tested it a whole lot (we don't have much using these)... I tried to come up with something leveraging `FUTEX_CMP_REQUEUE` and `FUTEX_WAKE_OP` but couldn't figure out a scheme that might work, but I also couldn't find any _readable_ real world implementation which obviously must exist _somewhere_. Anyway, the implementation here uses `FUTEX_REQUEUE` and `FUTEX_WAKE instead and seems to work.

Libraries/LibPthread/pthread.cpp
```diff
@@ -157,70 +167,233 @@ int pthread_sigmask(int how, const sigset_t* set, sigset_t* old_set)
...
+                    continue;
+                }
+#ifdef PTHREAD_DEBUG
+                dbg() << "futex_lock @ " << &futex_addr << ": wait error " << err;
+#endif
+                return err;
+            }
+            atomic_thread_fence(AK::MemoryOrder::memory_order_release);
+#ifdef PTHREAD_DEBUG
+            dbg() << "futex_lock @ " << &futex_addr << ": acquired after waiting";
```
awesomekling, line 211: Please use `dbgln()` in new code, the streaming variants are being phased out.

Libraries/LibPthread/pthread.cpp
```diff
@@ -157,70 +167,233 @@ int pthread_sigmask(int how, const sigset_t* set, sigset_t* old_set)
     return 0;
 }
 
-int pthread_mutex_init(pthread_mutex_t* mutex, const pthread_mutexattr_t* attributes)
+#define PTHREAD_FUTEX_UNLOCKED 0u
+#define PTHREAD_FUTEX_LOCKED 1u
+#define PTHREAD_FUTEX_LOCKED_WITH_WAITERS 2u
+
+static int futex_lock(uint32_t& futex_addr, uint32_t futex_flags, const timespec* abstime = nullptr, bool no_wait = false)
```
bugaevc, line 174: I don't think it's a good idea to wrap futexes like that to make them appear to be simple locks. Futexes are much more versatile: one can, and should, use the value of the futex word directly to encode the current state (such as "locked for writing, no waiting readers"), not just "locked" and "unlocked".

Libraries/LibPthread/pthread.cpp
```diff
@@ -157,70 +167,233 @@ int pthread_sigmask(int how, const sigset_t* set, sigset_t* old_set)
...
-    mutex->lock = 0;
-    mutex->owner = 0;
-    mutex->level = 0;
-    mutex->type = attributes ? attributes->type : PTHREAD_MUTEX_NORMAL;
-    return 0;
+    ASSERT(!(futex_flags & ~(FUTEX_PRIVATE_FLAG | FUTEX_CLOCK_REALTIME)));
+    for (;;) {
+        uint32_t expected = PTHREAD_FUTEX_UNLOCKED;
+        if (AK::atomic_compare_exchange_strong(&futex_addr, expected, PTHREAD_FUTEX_LOCKED, AK::MemoryOrder::memory_order_acquire)) {
+            atomic_thread_fence(AK::MemoryOrder::memory_order_release);
```
bugaevc, line 180: Why is this needed here?

Libraries/LibPthread/pthread.cpp
```diff
@@ -157,70 +167,233 @@ int pthread_sigmask(int how, const sigset_t* set, sigset_t* old_set)
...
 {
-    mutex->lock = 0;
-    mutex->owner = 0;
-    mutex->level = 0;
-    mutex->type = attributes ? attributes->type : PTHREAD_MUTEX_NORMAL;
-    return 0;
+    ASSERT(!(futex_flags & ~(FUTEX_PRIVATE_FLAG | FUTEX_CLOCK_REALTIME)));
+    for (;;) {
+        uint32_t expected = PTHREAD_FUTEX_UNLOCKED;
+        if (AK::atomic_compare_exchange_strong(&futex_addr, expected, PTHREAD_FUTEX_LOCKED, AK::MemoryOrder::memory_order_acquire)) {
```
bugaevc, line 179: This is inside a loop, so it makes sense to use the weak variant.

Libraries/LibPthread/pthread.cpp
```diff
@@ -157,70 +167,233 @@ int pthread_sigmask(int how, const sigset_t* set, sigset_t* old_set)
...
+            atomic_thread_fence(AK::MemoryOrder::memory_order_release);
+#ifdef PTHREAD_DEBUG
+            dbg() << "futex_lock @ " << &futex_addr << ": acquired @ " << &futex_addr;
+#endif
+            return 0;
+        }
+        if (no_wait)
+            return EBUSY;
+        if (expected == PTHREAD_FUTEX_LOCKED_WITH_WAITERS || AK::atomic_compare_exchange_strong(&futex_addr, expected, PTHREAD_FUTEX_LOCKED_WITH_WAITERS, AK::MemoryOrder::memory_order_acquire)) {
```
bugaevc, line 188: ditto weak variant

bugaevc: Also I don't think you need acquire here: if the cmp_xcg fails, you're continuing to the next iteration, where you already use acquire.

Libraries/LibPthread/pthread.cpp
```diff
@@ -157,70 +167,233 @@ int pthread_sigmask(int how, const sigset_t* set, sigset_t* old_set)
...
+                    dbg() << "futex_lock @ " << &futex_addr << ": waiting aborted, try locking again";
+#endif
+                    continue;
+                }
+#ifdef PTHREAD_DEBUG
+                dbg() << "futex_lock @ " << &futex_addr << ": wait error " << err;
+#endif
+                return err;
+            }
+            atomic_thread_fence(AK::MemoryOrder::memory_order_release);
```
bugaevc, line 209: ditto: why?

Libraries/LibPthread/pthread.cpp
```diff
@@ -157,70 +167,233 @@ int pthread_sigmask(int how, const sigset_t* set, sigset_t* old_set)
...
+        }
+    }
 }
 
-int pthread_mutex_destroy(pthread_mutex_t*)
+static int futex_unlock(uint32_t& futex_addr, uint32_t futex_flags)
 {
+    ASSERT(!(futex_flags & ~(FUTEX_PRIVATE_FLAG | FUTEX_CLOCK_REALTIME)));
+    uint32_t expected = PTHREAD_FUTEX_LOCKED;
+    if (AK::atomic_compare_exchange_strong(&futex_addr, expected, PTHREAD_FUTEX_UNLOCKED, AK::MemoryOrder::memory_order_acquire)) {
```
bugaevc, line 222: I usually just use `exchange()` (w/o comparing) on the release/unlock path

Also this needs release, not acquire.

Libraries/LibPthread/pthread.cpp
```diff
@@ -157,70 +167,233 @@ int pthread_sigmask(int how, const sigset_t* set, sigset_t* old_set)
...
+#ifdef PTHREAD_DEBUG
+                dbg() << "futex_lock @ " << &futex_addr << ": wait error " << err;
+#endif
+                return err;
+            }
+            atomic_thread_fence(AK::MemoryOrder::memory_order_release);
+#ifdef PTHREAD_DEBUG
+            dbg() << "futex_lock @ " << &futex_addr << ": acquired after waiting";
+#endif
+            return 0;
```
bugaevc, line 213: After we wake from sleeping on a futex, we need to loop again & try to take the lock, otherwise we leave it unlocked

Libraries/LibPthread/pthread.cpp
tomuta: Maybe I'll call the function `mutex_lock` then, it's an internal function only...

Libraries/LibPthread/pthread.cpp
tomuta: good point, yeah I think we don't need the release here

Libraries/LibPthread/pthread.cpp
tomuta: We don't have the weak variants (yet)...

Libraries/LibPthread/pthread.cpp
tomuta: Using `memory_order_relaxed` in a loop is not a good idea, either. Doing so can supposedly (some architectures) cause a extreme load on the memory bus, affecting performance much more than e.g. `memory_order_acquire`.

Libraries/LibPthread/pthread.cpp
tomuta: Dang you're right. BUT this is the spot to leverage `FUTEX_WAKE_OP` instead so this looping and taking the lock is not necessary ;-)

Libraries/LibPthread/pthread.cpp
bugaevc: My point is rather that each synchronization primitive should use futexes in its own special way that makes sense for that primitive — not simply as a lock.

I especially dislike it when various operations get implemented by locking an internal lock, checking some state, then unlocking the internal lock. That kills performance. Operations should not block each other from progressing any more than is required by the semantic of the operation — i.e., if we're talking about `Mutex::lock()` then surely it must block other threads executing `Mutex::lock()`, but `ConditionVariable::notify_all()` should not block other threads from doing the same. Also, "acquire" operations should not have to use any release-or-stronger atomics, and vice versa.

Libraries/LibPthread/pthread.cpp
bugaevc: Really??

I'm not an expert by any means, but I can't imagine why relaxed could be slower then acquire, since it seems that acquire does everything relax does and then some. So it'd always be correct (if inefficient) to replace any relaxed atomic op by an acquire op — so how can it be any slower?

Libraries/LibPthread/pthread.cpp
bugaevc: I don't think it's true though. You still need the thread to "see" the stuff the other thread has written, so it needs to perform an acquire op *after* waking from sleep. It can be just a fence though.

But there's another tricky thing here. The unlock/wake path has to actually change the value (from `locked_with_waiters` to `unlocked`, presumably) for the lock/wait path not to miss wakeups.

P.S. I assume you have read "futexes are tricky", right?

tomuta: Yeah I've read that. I think that at least the kernel support is in good shape (although largely untested at this point).

Thinking about it some more I don't think `FUTEX_WAKE_OP` will work in this particular instance, either.

Libraries/LibPthread/pthread.cpp
tomuta: Well the acquire would be slower but not saturate the memory bus as much as relaxed. It's not an issue on x86 but supposedly on some others (maybe even arm?). I can't remember where I read this, but it was essentially the exact same question: why one would not just use relaxed and when successful, do an acquire memory fence and bail the loop.

Libraries/LibPthread/pthread.cpp
tomuta: Yeah i get your point, but in the case of condition variables I can't figure out a way to do it without essentially a mutex. Even glibc uses a mutex for waiting, signaling, and broadcasting a condition variable. Every single time.

That being said, I would really like to leverage `FUTEX_WAIT_OP` and `FUTEX_CMP_REQUEUE` but I haven't been able to come up with a scheme that would work. Any ideas?

bugaevc: I've posted a link to my implementation of a condition variable (and other synchronization primitives, too!) somewhere above. [Here it is again](https://github.com/bugaevc/lets-write-sync-primitives/blob/master/src/condvar.cpp). It's supposed to be understandable, clean code, and it uses `FUTEX_REQUEUE` 😀 (without `CMP` though).

tomuta: Yeah I remember that one. This implementation also uses `FUTEX_REQUEUE`. Your implementation doesn't allow condvars choosing an arbitrary mutex after creation, which means you can't create a condvar and pick a mutex later (or change it). While condvars only allow the same mutex to be used at the same time, it's entirely possible to use a different mutex with the same condvar at a later point, so long as it's consistent. Hence the "mutex" that e.g. glibc uses in their condvar implementation.

bugaevc: Yes, I took the opportunity to pick somewhat nicer APIs there 🙂

TBH using a single condvar with multiple mutexes is a weird/broken thing to do to begin with (unlike using a single mutex with multiple condvars). But if we have to support that, it should be a matter of just doing relaxed reads/writes to a mutex pointer, no?

tomuta: Quite frankly, it'd be a nice LibCore thing ;) It'd be nice to have c++ mutexes, condvars etc, and they do not have to be built upon pthreads necessarily ;-)

bugaevc: That's what LibThread is for 🙂

But the issue with that is we want both LibThread to use LibPthread and the other way around, and that doesn't quite work.

tomuta: No, it's not broken. Yeah the API sucks, and if the mutex had to be specified when initializing the condvar then this particular model could work. The standard only says that you cannot use different mutexes at the same time, so you can start out with mutex A, and when no thread is waiting switch to using mutex B using the same condvar. But we only need pthread to support ports, we can have a better, more restrictive but also more performant condvar like what you came up with for our own stuff in LibCore.

bugaevc: > No, it's not broken.

Is there an example of how a condvar can be reasonably used with multiple mutexes? In my mental model, a condvar is supposed to guard a single condition/predicte that's also protected by a mutex (you can only evaluate or modify the condition while holding the mutex). So I can't really imagine a valid use case for different `wait()` calls supplying different mutexes.

> The standard only says that you cannot use different mutexes at the same time, so you can start out with mutex A, and when no thread is waiting switch to using mutex B using the same condvar.

Yes, I understand. I'm saying I don't understand why that would be useful for a program to use a condvar this way, because to me it sounds like breaking the ideas condvars are built on.

> Yeah the API sucks, and if the mutex had to be specified when initializing the condvar then this particular model could work. <...> we can have a better, more restrictive but also more performant condvar like what you came up with for our own stuff in LibCore.

But again, why can't my implementation be made to work with changing mutexes too? You'd replace the reference to the mutex with a pointer, set the mutex first thing in `.wait()`, and read it in `.notify_all()`. Both reading and writing the mutex pointer ~can be relaxed~ don't even have to be atomic (because the mutex pointer gets essentially protected by the mutex itself); no need for additional locking.

tomuta: I'm not sure how you can get this without races without a mutex even if you use an atomic variable for the mutex pointer. Think of two rapid broadcasts/notify_all where the waiters switch the mutex (while assuring they aren't using a different mutex at the same time). Depending on the sequence of operations in the wait that changes the mutex and the second broadcast/notify_all, you could end up using the previous mutex accidentally. The only way I can think of that you can do this is using a mutex. This is the line I'm referring to in this implementation:
https://github.com/SerenityOS/serenity/blob/3ff4d2349a416c011fee8f50ea048314834c2475/Libraries/LibPthread/pthread.cpp#L827

As you can see, it gets the mutex pointer and then resets it. Because this is synchronized with wait/signal/broadcast it should be race-free, even though it is using the pointer obtained while holding the lock after releasing the lock. This should still be still race-free as a subsequent wait with a different mutex between grabbing the pointer and actually waking/requeueing would just requeue to the previous mutex, which should still exist and would be the correct mutex to requeue to based on when broadcast was triggered. Any subsequent signal/broadcast would requeue to the new mutex, even if the first requeue hasn't happened yet.

bugaevc: Oh, I think I see what you mean.

That (using a single condvar with different mutexes *concurrently*) is *even more* broken than "just" using a single condvar with different mutexes. Surely we don't have to support that!

> would just requeue to the previous mutex, which should still exist and would be the correct mutex to requeue to based on when broadcast was triggered

I'm not sure there's such a thing as a thing as a "correct" mutex to use here, because this whole situation is very wrong. There can be no happens-before relationship between entering `wait()` and `notify()` calls (even if some external means of synchronization are used). And if there was, there's no guarantee that they'll acquire you internal lock in the same order.

I would argue that the only "correct" thing to do is for the wake/notify path to requeue waiting threads onto whichever mutex the condition variable is bound to "at that very moment", or in other words that fetching the mutex and requeuing must be done atomically wrt `wait()` calls changing the mutex. If that is done atomically and the `wait()` calls are somehow synchronized between themselves, all the waiters at the moment of requeuing use the same mutex — the one we'd atomically observe — and we'd wake/requeue them all "correctly".

That could be factored into you implementation, you just have to release the internal lock later after requeuing. That would slow it down further though, since you'll hold onto the lock for the duration of the syscall.

The other "correct" thing to do is to not use requeuing at all (i.e. pessimize) in case concurrent mutex modification is detected. Or, since we can only detect this *after* requeuing, wake everyone we've requeued up, effectively undoing the requeuing. My implementation could be patched the following way for that:

```cpp
void CondVar::notify_all() {
    // ...increment state...
    if (need to wake) {
        // ...unset need to wake bits...
        Mutex *mutex1 = m_mutex.load();
        futex_requeue(
            (const uint32_t *) &state, 1,
            (const uint32_t *) &mutex1->state, INT_MAX
        );
        Mutex *mutex2 = m_mutex.load();
        if (mutex1 != mutex2) {
            // Mutex has changed! Make sure
            // to wake everyone we may just have
            // requeued to the possibly wrong mutex,
            // and let them sort it out themselves.
            futex_wake(&mutex1->state, INT_MAX);
        }
    }
}
```

But again, IMO this is too broken to support. We don't have to follow POSIX 100%.

bugaevc: > `mutex1 != mutex2`

No, can't do that, that's the ABA problem.

bugaevc: Well, I guess if we can't figure that out *and* we decide we have to support this case — it'd be fine to use an internal lock on the slow path. For condvars, `wait()` is always the slow path, and `notify_all()`/`broadcast()` has a fast path where no one's waiting and a slow path where we do have to wake somebody; it would be OK to use the lock on the latter. We should make sure to drop the "have waiters"/"should take slow path" flag prior to taking the lock though, so that of two concurrent `notify_all()` calls only one has to enter the slow path.

tomuta: I don't think the re-queueing has to be done while holding the lock. But yeah there is a race where changing the mutex could cause requeuing to the wrong mutex. I wonder if this could be solved somehow with `FUTEX_CMP_REQUEUE` and maybe yet another atomic variable tracking mutex changes or something like that... Maybe I'll just drop the usermode changes from this PR for now, because I think the kernel implementation is conceptually matching that of Linux. I haven't tested it out thoroughly, though...

bugaevc: > I wonder if this could be solved somehow with `FUTEX_CMP_REQUEUE`

I've thought about that, but that wouldn't work because it compares the value of the same word it requeues from, not the value of a different word.

We could add a special kernel op for this; but I'm not sure it'd be much better than locking in userspace on the slow path.

Libraries/LibPthread/pthread.cpp
tomuta: Here's the quote I was referring to:

> This "optimization" is however a) useless and b) may in fact hurt: a) Since the thread will be busily spinning on a blocked spinlock, it does not matter if it will waste the CPU cycles with just "exchange" operations or with both useless "exchange" and "acquire" operations. b) A tight "exchange" loop without any memory-synchronizing instruction introduced through an "acquire" operation will on some systems monopolize the memory subsystem and degrade the performance of other system components.

As per the "Discussion" section of the [Spinlock example](https://www.boost.org/doc/libs/1_72_0/doc/html/atomic/usage_examples.html#boost_atomic.usage_examples.example_spinlock).

tomuta:
I marked all the pthread.cpp related conversations as resolved as I removed those changes from the PR for now. While I haven't tested all the features, maybe we can do that part once we have some good pthread.cpp changes that can leverage them.

Kernel/WaitQueue.cpp
```diff
@@ -71,50 +71,56 @@ void WaitQueue::wake_one()
     m_wake_requested = !did_unblock_one;
 }
 
-void WaitQueue::wake_n(u32 wake_count)
+u32 WaitQueue::wake_n(u32 wake_count)
 {
     if (wake_count == 0)
-        return; // should we assert instead?
+        return 0; // should we assert instaed?
```
emanuele6, line 77: you repleace "instead" with "instaed" here.

tomuta: whoops. Not sure how that happened...

emanuele6: @tomuta: ping. `lint-clang-format.sh` is failing.

awesomekling:
Do you have any other comments on this @bugaevc?

bugaevc:
Not really; and I don't claim to understand the actual Kernel-side futex implementation 🙂

Let's land this now, and if/when I figure it all out, we can discuss it further.

tomuta:
I'll happily fix problems as we start utilizing them more and more ;-)

awesomekling:
> I'll happily fix problems as we start utilizing them more and more ;-)

Grand! Please fix the merge conflict and we'll merge this :)

PR #4468, opened 2020-12-20
Ports: Add initial stress-ng port to find bugs in serenity

bgianfo:
This is a very WIP port bringing [stress-ng](https://kernel.ubuntu.com/~cking/stress-ng/) to SerenityOS.

stress-ng is great at doing multi-workload stress testing, this allows it to
find unique and interesting intermixed pairs of stressful operations which cause bugs.

This initial port just rips out all non applicable functionality to get the port to compile.


awesomekling:
Oh this sounds like it could get super interesting!

PR #4471, opened 2020-12-20
FileManager: Add links in file properties to open links in their real directory

lxea:
This adds a button if a file is a link to open a file manager in the original  files directory
20201221_14h05m18s_grim

### DIFF
--- a/Libraries/LibGUI/Link.cpp
+++ b/Libraries/LibGUI/Link.cpp
@@ -1,0 +1,89 @@
+/*
+ * Copyright (c) 2020, Alex McGrath <amk@amk.ie>
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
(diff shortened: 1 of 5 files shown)


diversys:
Maybe Location and Link target path sections can be linkified?

AlexKraak:
> Maybe Location and Link target path sections can be linkified?

I'll give this a try, that seems like itd be better

Libraries/LibGUI/Link.cpp
```diff
@@ -0,0 +1,50 @@
+/*
+ * Copyright (c) 2018-2020, Andreas Kling <kling@serenityos.org>
```
linusg, line 2: Andreas didn't write this, you did :)

Libraries/LibGUI/Link.h
```diff
@@ -0,0 +1,47 @@
+/*
+ * Copyright (c) 2018-2020, Andreas Kling <kling@serenityos.org>
```
linusg, line 2: Same as above

Libraries/LibGUI/Link.cpp
```diff
@@ -0,0 +1,50 @@
...
+    if (on_click) {
+        on_click();
+    }
+}
+
+void Link::paint_event(PaintEvent& event)
+{
+    Label::paint_event(event);
+    GUI::Painter painter(*this);
+    painter.draw_line({ 0, rect().bottom() }, { font().width(text()), rect().bottom() }, Color::MidGray);
```
linusg, line 47: this is wrong as it won't work properly with themes other than the default one, please use `Gfx::Palette` (accessible via `Widget::palette()`)

linusg (changes requested):

Libraries/LibGUI/Link.cpp
AlexKraak: Ah thank you, I wasnt sure what to look for for this!

diversys:
Looks nice! Just a few more thoughts:
How about making links coloured by default and underline them on mouse over? 
Is there a tooltip if a link is longer than window width?
There is something wrong with separators below Browser text field and above permission view.

diversys:
Also is scroll to selection after clicking links implemented?

lxea:
> How about making links coloured by default and underline them on mouse over?

20201224_13h11m06s_grim
I have this now, if you hover over it it shows the underline and the tooltip, 

> Is there a tooltip if a link is longer than window width?

im not sure how to make it so the tool tip only shows if its too long but ill give it another stab in a bit

> There is something wrong with separators below Browser text field and above permission view.

Not sure what you mean here, i dont think they should have changed

> Also is scroll to selection after clicking links implemented?

I dont think this exists yet,

diversys:
This looks nifty! :)

lxea:
Latest commit, it looks the same as last screenshot but itll only show the tooltip if the text is wider than the window

PR #4473, opened 2020-12-20
Everywhere: Switch from (void) to [[maybe_unused]]

ldm5180:
Problem:
- `(void)` simply casts the expression to void. This is understood to
  indicate that it is ignored, but this is really a compiler trick to
  get the compiler to not generate a warning.

Solution:
- Use the `[[maybe_unused]]` attribute to indicate the value is unused.

Note:
- Functions taking a `(void)` argument list have also been changed to
  `()` because this is not needed and shows up in the same grep
  command.

### DIFF
--- a/Libraries/LibC/assert.h
+++ b/Libraries/LibC/assert.h
@@ -41,7 +41,7 @@ __attribute__((noreturn)) void __assertion_failed(const char* msg);
         } while (0)
 #    define ASSERT_NOT_REACHED() assert(false)
 #else
-#    define assert(expr) ((void)0)
+#    define assert(expr) (void(0))
 #    define ASSERT_NOT_REACHED() CRASH()
 #endif
 
--- a/Libraries/LibC/stdlib.h
+++ b/Libraries/LibC/stdlib.h
@@ -41,13 +41,13 @@ __BEGIN_DECLS
 __attribute__((malloc)) __attribute__((alloc_size(1))) void* malloc(size_t);
 __attribute__((malloc)) __attribute__((alloc_size(1, 2))) void* calloc(size_t nmemb, size_t);
 size_t malloc_size(void*);
-void serenity_dump_malloc_stats(void);
+void serenity_dump_malloc_stats();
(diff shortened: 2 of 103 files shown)


Libraries/LibC/assert.h
```diff
@@ -41,7 +41,7 @@ __attribute__((noreturn)) void __assertion_failed(const char* msg);
         } while (0)
 #    define ASSERT_NOT_REACHED() assert(false)
 #else
-#    define assert(expr) ((void)0)
+#    define assert(expr) (void(0))
```
bugaevc, line 44: Is `void(0)` even correct C?

Libraries/LibC/stdlib.h
```diff
@@ -41,13 +41,13 @@ __BEGIN_DECLS
...
 size_t malloc_size(void*);
-void serenity_dump_malloc_stats(void);
+void serenity_dump_malloc_stats();
 void free(void*);
 __attribute__((alloc_size(2))) void* realloc(void* ptr, size_t);
 char* getenv(const char* name);
 int putenv(char*);
 int unsetenv(const char*);
-int clearenv(void);
+int clearenv();
```
bugaevc, line 50: It's incorrect to do this to C headers

In C, `void foo()` declares a function *without specifying anything* about its parameter count / types (somewhat like `void foo(...)`, but different); while `void foo(void)` declares a function taking no arguments.

ldm5180: Oops! You are right. I didn't think about these being used in C code, but there is no reason they shouldn't be. I just opened PR #4522 to address this. Thanks for catching it.

Libraries/LibC/assert.h
eteran: @bugaevc both are valid in C++, but I believe that `void(0)` is **not** valid C. Basically it is using C++'s "constructor like cast syntax". Just as you can do `(int)5.2` you can also do `int(5.2)`.

PR #4477, opened 2020-12-21
Userland: useradd: Add command line option to set user password

bcoles:
Prior to this PR, the `useradd` utility failed to initialize the `pw_passwd` member of the `passwd` struct resulting in unexpected behavior. This has been fixed.

Before (lol `DT_JMPREL`) :

useradd

After:

useradd


awesomekling:
hah!

PR #4486, opened 2020-12-22
Spreadsheet: Some cleanups and "stability improvements" :P

alimpfard:
In this PR:
- Spreadsheet can once again load whatever it has saved
- Stop treating `Cell` as a struct, it has clearly outlived that stage
- Renaming sheets
- Crash-no-mo (put stored `JS::Cell`s inside handles)
- make `here().left()` no longer try to make `(size_t)-1` columns when entered in the leftmost cell of a row
- Don't update unrelated cells just because they have a deep dependency chain

### DIFF
--- a/Applications/Spreadsheet/Cell.h
+++ b/Applications/Spreadsheet/Cell.h
@@ -38,21 +38,26 @@
 namespace Spreadsheet {
 
 struct Cell : public Weakable<Cell> {
+    enum Kind {
+        LiteralString,
+        Formula,
+    };
+
     Cell(String data, Position position, WeakPtr<Sheet> sheet)
-        : dirty(false)
-        , data(move(data))
-        , kind(LiteralString)
-        , sheet(sheet)
+        : m_dirty(false)
+        , m_data(move(data))
+        , m_kind(LiteralString)
+        , m_sheet(sheet)
         , m_position(move(position))
     {
     }
 
     Cell(String source, JS::Value&& cell_value, Position position, WeakPtr<Sheet> sheet)
-        : dirty(false)
-        , data(move(source))
-        , evaluated_data(move(cell_value))
-        , kind(Formula)
-        , sheet(sheet)
+        : m_dirty(false)
+        , m_data(move(source))
+        , m_evaluated_data(move(cell_value))
+        , m_kind(Formula)
+        , m_sheet(sheet)
         , m_position(move(position))
     {
     }
@@ -61,6 +66,12 @@ struct Cell : public Weakable<Cell> {
 
     void set_data(String new_data);
     void set_data(JS::Value new_data);
+    bool dirty() const { return m_dirty; }
+
+    const String& data() const { return m_data; }
+    const JS::Value& evaluated_data() const { return m_evaluated_data; }
+    Kind kind() const { return m_kind; }
+    const Vector<WeakPtr<Cell>>& referencing_cells() const { return m_referencing_cells; }
 
     void set_type(const StringView& name);
     void set_type(const CellType*);
@@ -69,15 +80,16 @@ struct Cell : public Weakable<Cell> {
     const Position& position() const { return m_position; }
     void set_position(Position position, Badge<Sheet>)
     {
-        dirty = true;
+        m_dirty = true;
         m_position = move(position);
     }
 
     const Format& evaluated_formats() const { return m_evaluated_formats; }
+    Format& evaluated_formats() { return m_evaluated_formats; }
     const Vector<ConditionalFormat>& conditional_formats() const { return m_conditional_formats; }
     void set_conditional_formats(Vector<ConditionalFormat>&& fmts)
     {
-        dirty = true;
+        m_dirty = true;
         m_conditional_formats = move(fmts);
     }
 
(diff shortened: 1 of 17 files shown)


alimpfard:
Of minor interest, I found this little `/proc/<pid>/coredump` very useful - maybe we should have something like this?
```diff
diff --git a/Kernel/FileSystem/ProcFS.cpp b/Kernel/FileSystem/ProcFS.cpp
index f223f40dd..53c0116f1 100644
--- a/Kernel/FileSystem/ProcFS.cpp
+++ b/Kernel/FileSystem/ProcFS.cpp
@@ -32,6 +32,7 @@
 #include <Kernel/Arch/i386/ProcessorInfo.h>
 #include <Kernel/CommandLine.h>
 #include <Kernel/Console.h>
+#include <Kernel/CoreDump.h>
 #include <Kernel/Devices/BlockDevice.h>
 #include <Kernel/Devices/KeyboardDevice.h>
 #include <Kernel/FileSystem/Custody.h>
@@ -53,6 +54,7 @@
 #include <Kernel/PCI/Access.h>
 #include <Kernel/Process.h>
 #include <Kernel/Profiling.h>
+#include <Kernel/RTC.h>
 #include <Kernel/Scheduler.h>
 #include <Kernel/StdLib.h>
 #include <Kernel/TTY/TTY.h>
@@ -122,6 +124,7 @@ enum ProcFileType {
     FI_PID_cwd,  // symlink
     FI_PID_root, // symlink
     FI_PID_fd,   // directory
+    FI_PID_coredump,
     __FI_PID_End,

     FI_MaxStaticFileIndex,
@@ -689,6 +692,24 @@ static OwnPtr<KBuffer> procfs$pid_root(InodeIdentifier identifier)
     return KBuffer::try_create_with_bytes(process->root_directory_relative_to_global_root().absolute_path().to_byte_buffer());
 }

+static OwnPtr<KBuffer> procfs$pid_coredump(InodeIdentifier identifier)
+{
+    auto process = Process::from_pid(to_pid(identifier));
+    if (!process)
+        return {};
+    dbgln("Generating coredump for pid: {}", process->pid().value());
+
+    auto coredump_path = String::formatted("/tmp/coredump/{}_{}_{}", process->name(), process->pid().value(), RTC::now());
+    auto coredump = CoreDump::create(*process, coredump_path);
+    if (coredump) {
+        coredump->write();
+        return KBuffer::try_create_with_bytes(coredump_path.to_byte_buffer());
+    } else {
+        dbgln("Could not create coredump");
+        return {};
+    }
+}
+
 static OwnPtr<KBuffer> procfs$self(InodeIdentifier)
 {
     char buffer[16];
@@ -1712,6 +1733,7 @@ ProcFS::ProcFS()
     m_entries[FI_PID_unveil] = { "unveil", FI_PID_unveil, false, procfs$pid_unveil };
     m_entries[FI_PID_root] = { "root", FI_PID_root, false, procfs$pid_root };
     m_entries[FI_PID_fd] = { "fd", FI_PID_fd, false };
+    m_entries[FI_PID_coredump] = { "coredump", FI_PID_coredump, false, procfs$pid_coredump };
 }

 ProcFS::ProcFSDirectoryEntry* ProcFS::get_directory_entry(InodeIdentifier identifier) const
```

alimpfard:
For extra fun, (part of) part 1 of AoC day 8, which was the test case for all of these fixes lol
[workbook.tar.gz](https://github.com/SerenityOS/serenity/files/5729025/workbook.tar.gz)

Applications/Spreadsheet/JSIntegration.cpp
```diff
@@ -34,6 +34,42 @@
 
 namespace Spreadsheet {
 
+bool can_be_safely_converted(const JS::Value& value)
+{
+    // FIXME: LibJS, please assert less on things you're handing out to your users.
```
linusg, line 39: I disagree with this, at least "from the outside" (JS code rather than LibJS-C++) it should be impossible to pass any of the value types below to a stringification function:

- Empty values should never be leaked, e.g. accessing an array hole returns undefined
- Accessors and native properties can exist in objects, but they always return a primitive or object when accessing them

So what am I missing here?

alimpfard: Those kinds of objects likely showed up because the GC was freeing things.
I guess that function has outlived its purpose

alimpfard: Hopefully we don't see this kind of GC mishap again :)

alimpfard:
/me forgets that he has random unstaged crap

Applications/Spreadsheet/Cell.h
```diff
@@ -92,30 +104,28 @@ struct Cell : public Weakable<Cell> {
...
+    void update_data(Badge<Sheet>);
 
-    enum Kind {
-        LiteralString,
-        Formula,
-    };
+    const Sheet& sheet() const { return *m_sheet; }
+    Sheet& sheet() { return *m_sheet; }
+
+    void copy_from(Cell&);
```
awesomekling, line 121: This intuitively feels like it should take a `const Cell&`

Applications/Spreadsheet/Cell.h
```diff
@@ -117,6 +122,8 @@ struct Cell : public Weakable<Cell> {
     bool m_evaluated_externally { false };
     String m_data;
     JS::Value m_evaluated_data;
+    JS::Handle<JS::Cell> m_evaluated_data_handle;
+    JS::Handle<JS::Exception> m_js_exception;
```
awesomekling, line 129: This does work but if we plan on holding on to a lot of these, it might be better to implement a `visit_edges` function that traces all the referenced JS cells instead of using handles (since handles have both a speed and space penalty in the GC.)

alimpfard: Sure, but wouldn't the the speed savings be nullified by having to go through the entire hashmap of sheet cells and conditionally visiting stuff?

awesomekling: Very possible. We can leave it using handles for now, no need to prematurely optimize :)
Perhaps something we'll come back to in the future

alimpfard: Too late, I've already switched to `visit_edges()` and it looks much cleaner.

PR #4489, opened 2020-12-22
Kernel/Net: E1000 interrupt rate register is 32-bit, not 16-bit

Lubrsi:
I looked at the spec sheet and noticed that it's 32-bit, not 16-bit.
This fixes E1000 causing an MMIO fault on VirtualBox.

Spec: https://www.intel.com/content/dam/doc/manual/pci-pci-x-family-gbe-controllers-software-dev-manual.pdf
Section 13.4.18


awesomekling:
Excellent catch! :)

Issue #4484, opened 2020-12-21
Applications launched by the SystemMenu do not inherit the user's login profile


lux01:
As a consequence of how the SystemMenu process is launched by SystemServer, the user's login profile is never run. Consequentially any applications launched from the SystemMenu have no environment variables set. Crucially this means that standard environment variables like `HOME` are not set.

This manifests itself when an application like Snake tries to access the per-user config file. Since the `HOME` environment variable is not set, `LibC` attempts to read `/etc/passwd` to find the home directory which fails because `/etc/passwd` has not been unveiled. This can be seen from a repeated `perror("open /etc/passwd")` message in the syslog:

![image](https://user-images.githubusercontent.com/565354/102822319-b4c4ff80-43d0-11eb-9ce8-303df4146934.png)

The `perror` call is from the following stack:
```
0x04b34bb9: [libc.so] setpwent +99 (pwd.cpp:60)
0x04b3523b: [libc.so] getpwuid +21 (pwd.cpp:82)
0x04c03426: [libcore.so] Core::StandardPaths::home_directory() +94 (StandardPaths.cpp:43)
0x04c0368e: [libcore.so] Core::StandardPaths::config_directory() +66 (StandardPaths.cpp:67)
0x04bee36b: [libcore.so] Core::ConfigFile::get_for_app(AK::String const&) +35 (StringView.h:55)
0x051c5da6: [/bin/Snake] SnakeGame::SnakeGame() +1816 (SnakeGame.cpp:45)
0x051c4d09: [/bin/Snake] SnakeGame& GUI::Window::set_main_widget<SnakeGame>() +39 (Atomic.h:158)
0x051c41a8: [/bin/Snake]  +1096 (kmalloc.h:59)
0x051c6ac1: [/bin/Snake] _start +36 (crt0_shared.cpp:50)
```

PR #4498, opened 2020-12-22
Shell+SystemMenu: Launch SystemMenu apps by execing them in the user's shell

lux01:
Currently the environment for the `SystemMenu` process is empty because the user's shell is not invoked at any point during its startup sequence. Consequentially any processes that are launched from the `SystemMenu` inherit an empty environment and do not run the user's shell profile. This can lead to issues such as #4484 where common environment variables such as `HOME` are absent from the program's environment.

This PR adds a very basic `exec` builtin to `Shell` and causes `SystemMenu` to launch all executables via the user's shell by running `$SHELL -c "exec <Executable>"`. Currently we don't support passing additional arguments to programs being launched `SystemMenu` so this is fine for now, but this will need rethinking if we ever support additional arguments.

This behaviour can be opted-out of by specifying `UseShell=false` in the `App` section of the appropriate `.af` file.

### DIFF
--- a/Services/SystemMenu/main.cpp
+++ b/Services/SystemMenu/main.cpp
@@ -36,9 +36,10 @@
 #include <LibGUI/Menu.h>
 #include <LibGUI/WindowServerConnection.h>
 #include <LibGfx/Bitmap.h>
+#include <pwd.h>
 #include <serenity.h>
 #include <spawn.h>
-
+#include <unistd.h>
 //#define SYSTEM_MENU_DEBUG
 
 struct AppMetadata {
@@ -46,6 +47,7 @@ struct AppMetadata {
     String name;
     String icon_path;
     String category;
+    bool use_shell;
 };
 Vector<AppMetadata> g_apps;
 
@@ -59,7 +61,9 @@ Color g_menu_selection_color;
 Vector<ThemeMetadata> g_themes;
 RefPtr<GUI::Menu> g_themes_menu;
 GUI::ActionGroup g_themes_group;
+String g_user_shell;
 
+static String discover_user_shell();
 static Vector<String> discover_apps_and_categories();
 static NonnullRefPtr<GUI::Menu> build_system_menu();
 
@@ -112,7 +116,8 @@ Vector<String> discover_apps_and_categories()
         auto app_executable = af->read_entry("App", "Executable");
         auto app_category = af->read_entry("App", "Category");
         auto app_icon_path = af->read_entry("Icons", "16x16");
-        g_apps.append({ app_executable, app_name, app_icon_path, app_category });
+        auto app_use_shell = af->read_bool_entry("App", "UseShell", true);
+        g_apps.append({ app_executable, app_name, app_icon_path, app_category, app_use_shell });
         seen_app_categories.set(app_category);
     }
     quick_sort(g_apps, [](auto& a, auto& b) { return a.name < b.name; });
@@ -126,8 +131,33 @@ Vector<String> discover_apps_and_categories()
     return sorted_app_categories;
 }
 
+String discover_user_shell()
+{
+    auto current_uid = getuid();
+
+    errno = 0;
+    auto pwent = getpwent();
+    for (; pwent != nullptr; errno = 0, pwent = getpwent()) {
+        if (pwent->pw_uid == current_uid) {
+            break;
+        }
+    }
+
+    auto user_shell = String {};
+    if (errno) {
+        perror("getpwent");
+    } else {
+        user_shell = pwent->pw_shell;
+    }
+
+    endpwent();
+    dbgln("User shell is '{}'", user_shell);
+    return user_shell;
+}
+
 NonnullRefPtr<GUI::Menu> build_system_menu()
 {
+    g_user_shell = discover_user_shell();
     const Vector<String> sorted_app_categories = discover_apps_and_categories();
     auto system_menu = GUI::Menu::construct("\xE2\x9A\xA1"); // HIGH VOLTAGE SIGN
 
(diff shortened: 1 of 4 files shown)


alimpfard:
but...`/bin/Shell` doesn't support `exec`...

Shell/Builtin.cpp
```diff
@@ -256,6 +256,15 @@ int Shell::builtin_dirs(int argc, const char** argv)
     return 0;
 }
 
+int Shell::builtin_exec(int argc, const char** argv)
+{
+    Vector<const char*> argv_vector;
+    argv_vector.append(argv + 1, argc - 1);
```
alimpfard, line 262: Huh. I didn't know this overload existed - neat.

Shell/Builtin.cpp
```diff
@@ -256,6 +256,15 @@ int Shell::builtin_dirs(int argc, const char** argv)
...
+    argv_vector.append(nullptr);
+
+    execute_process(std::move(argv_vector));
```
alimpfard, line 265: just `move(argv_vector)`.

Shell/Shell.cpp
```diff
@@ -785,42 +785,7 @@ RefPtr<Job> Shell::run_command(const AST::Command& command)
...
-            } else {
-                if (S_ISDIR(st.st_mode)) {
-                    fprintf(stderr, "Shell: %s: Is a directory\n", argv[0]);
-                    _exit(126);
-                }
-                fprintf(stderr, "execvp(%s): %s\n", argv[0], strerror(saved_errno));
-            }
-            _exit(126);
-        }
+        execute_process(std::move(argv));
```
alimpfard, line 788: Same note here.

Services/SystemMenu/main.cpp
```diff
@@ -161,10 +191,20 @@ NonnullRefPtr<GUI::Menu> build_system_menu()
...
             dbg() << "Activated app with ID " << app_identifier;
-            const auto& bin = g_apps[app_identifier].executable;
+            const auto& app = g_apps[app_identifier];
             pid_t child_pid;
-            const char* argv[] = { bin.characters(), nullptr };
-            if ((errno = posix_spawn(&child_pid, bin.characters(), nullptr, nullptr, const_cast<char**>(argv), environ))) {
+            if (app.use_shell) {
+                StringBuilder command_builder(app.executable.length() + 5);
+                command_builder.append("exec ");
+                command_builder.append(app.executable);
```
alimpfard, line 199: What if this stuff has spaces and special shell characters and whatnot?

alimpfard: Since this thing links with `LibShell` anyway (through `LibGUI`) you might want to use `Shell::escape_token()`

Shell/Builtin.cpp
lux01: Oh I didn't know about that, thanks. Is it worth maybe mentioning this in the CodingStyle document?

Services/SystemMenu/main.cpp
lux01: Ah if only I had seen your second comment sooner... I've just pushed rebased commits which wraps the command in a single quoted string and replaces any embedded single quotes with `'"'"'` which ends the current single quoted string, concatenates a new double quoted string containing a single quote, and then reopens the single quoted string.

I tested the above by creating a symlink `/bin/Hello $(whoami) 'Shell' Code` to `/bin/HelloWorld` and created an application entry for it in `/res/apps`. To my great surprise it actually works!

image

I was deliberately trying to avoid using `LibShell` in the implementation since the user may (eventually) not use `/bin/Shell` as their default shell. I'm happy to rework the commit if you would prefer the cleaner approach of `Shell::escape_token()` over a shell agnostic approach.

alimpfard: Nah, your solution is alright (and it has the advantage of _being implemented_ lol)

Shell/Builtin.cpp
alimpfard: I mean...`std::anything` is rather rare here, but mentioning it in the coding style could be useful, I suppose. 🤷

lux01:
Oops, accidentally hit close on the GitHub app... Turns out that closes the PR not the popup!

awesomekling:
cc @bugaevc 

I like the new `exec` shell builtin, but it feels like huge hack to use it in SystemMenu. Can we solve this by having SystemServer set up the environment nicely for spawned processes instead?

bugaevc:
+1, we should have `exec` in the Shell, but let's not launch apps through the Shell. Doesn't SystemServer set `HOME` for the services it launches already? If it doesn't, let me implement that...

lux01:
I guess there's a discussion that needs to be had around whether users/developers would expect applications to run with the user's shell profile at all, and if so when is that profile obtained? Our options are:

1. Applications don't run with the user's shell profile at all unless they are launched via shell scripts. This is the current behaviour and I would argue is a bug (or at least unexpected) when comparing Serenity to Linux or macOS desktops.
2. The user's shell profile is evaluated once and then all apps inherit this state. This could be done by having some service that runs as the user (SystemMenu seems like the easiest until we have some sort of login manager) running the `env` command once in the user's shell and then parsing the output into its own environment.
3. The user's shell profile is evaluated each time an application is launched. This can either be done by launching commands via the exec command in the user's shell, or by doing the `env` scraping described in the previous step but do it every time a command is launched.

Unless there's disagreement I'll spin off the `exec` builtin into a separate PR and then raise an issue to track the problem of if/when we run the user's profile.

bugaevc:
> Applications don't run with the user's shell profile at all unless they are launched via shell scripts. This is the current behaviour and I would argue is a bug (or at least unexpected) when comparing Serenity to Linux or macOS desktops.

Well, we're neither Linux not macOS 🙂 The Shell is not involved in startup or app launching at all.

awesomekling:
Closing this since #4484 was fixed elsewhere and the shell exec builtin was added separately.

PR #4500, opened 2020-12-22
Kernel: Consolidate the various BlockCondition::unblock variants

### LINKED ISSUE #4496: Kernel: Trying to connect to a server e.g. Google in VirtualBox crashes
E1000 now works in VirtualBox. DHCP succeeds and it gets an initial ARP table. However, when it tries to reach out to the internet, e.g. going to Google, it will crash with:
```
[NetworkTask(5:5)]: ASSERTION FAILED: did_unblock
.././Kernel/Thread.h:461 in bool Kernel::Thread::BlockCondition::do_unblock_all(UnblockOne) [with UnblockOne = Kernel::ARPTableBlockCondition::unblock(const AK::IPv4Address&, const MACAddress&)::<lambda(auto:15&, void*)>]
[NetworkTask(5:5)]: 0xc0119088  __assertion_failed(char const*, char const*, unsigned int, char const*) +185
[NetworkTask(5:5)]: 0xc01660e8  Kernel::update_arp_table(AK::IPv4Address const&, MACAddress const&) +296
[NetworkTask(5:5)]: 0xc0162611  Kernel::handle_arp(EthernetFrameHeader const&, unsigned long) +499
[NetworkTask(5:5)]: 0xc016307a  Kernel::NetworkTask_main(void*) +705
[NetworkTask(5:5)]: 0xc01190bc  exit_kernel_thread +0
```

Note, when testing in VirtualBox, set the network adapter to Intel PRO/1000 MT Server (82545EM). This is because Intel PRO/1000 MT Desktop (82540EM) shares an IRQ with the ACPI parser, which crashes with an assertion failure because the parser's IRQ handler is not implemented.

tomuta:
The unblock_all variant used to ASSERT if a blocker didn't unblock,
but it wasn't clear from the name that it would do that. Because
the BlockCondition already asserts that no blockers are left at
destruction time, it would still catch blockers that haven't been
unblocked for whatever reason.

Fixes #4496

### DIFF
--- a/Kernel/FileSystem/File.h
+++ b/Kernel/FileSystem/File.h
@@ -58,7 +58,7 @@ public:
     void unblock()
     {
         ScopedSpinLock lock(m_lock);
-        do_unblock([&](auto& b, void* data) {
+        do_unblock([&](auto& b, void* data, bool&) {
             ASSERT(b.blocker_type() == Thread::Blocker::Type::File);
             auto& blocker = static_cast<Thread::FileBlocker&>(b);
             return blocker.unblock(false, data);
--- a/Kernel/FileSystem/Plan9FileSystem.cpp
+++ b/Kernel/FileSystem/Plan9FileSystem.cpp
@@ -469,7 +469,7 @@ bool Plan9FS::Plan9FSBlockCondition::should_add_blocker(Thread::Blocker& b, void
 
 void Plan9FS::Plan9FSBlockCondition::unblock_completed(u16 tag)
 {
-    unblock([&](Thread::Blocker& b, void*) {
+    unblock([&](Thread::Blocker& b, void*, bool&) {
         ASSERT(b.blocker_type() == Thread::Blocker::Type::Plan9FS);
         auto& blocker = static_cast<Blocker&>(b);
         return blocker.unblock(tag);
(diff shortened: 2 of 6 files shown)


tomuta:
@Lubrsi can you try if this solves your issue?

Lubrsi:
Screenshot from 2020-12-23 06-53-21

It does fix the crashes. But for some reason, it seems that anything that uses the network will lock up after a connection terminates. For example, the SerenityOS banner doesn't appear, even though it appears the transfer was successful when looking at wireshark.

tomuta:
Hm no idea if that's related. Doesn't seem like it would. What VMware config are you using? I have never managed to boot into the GUI.

Lubrsi:
I'm using VirtualBox, VMware doesn't work currently. There's a guide here: https://github.com/SerenityOS/serenity/blob/master/Documentation/VirtualBox.md

Ignore the bit about the adapter type. Set it to Intel PRO/1000 MT Server (82545EM) instead.

tomuta:
Ah sorry, I mixed up the two. I'm not sure if I'll find the time the next few days to look into it, but maybe...

tomuta:
> I'm using VirtualBox, VMware doesn't work currently. There's a guide here: https://github.com/SerenityOS/serenity/blob/master/Documentation/VirtualBox.md
> 
> Ignore the bit about the adapter type. Set it to Intel PRO/1000 MT Server (82545EM) instead.

In regards to that page's  "Note on updating the disk image" section, The `VBoxManage convertfromraw` does have an optional `--uuid` argument that can take a hard-coded UUID. So you can tell it to use one and then update a disk image without having to mess with the VirtualBox configuration every time...

PR #4501, opened 2020-12-22
Chess: Use the hourglass cursor while waiting for the ChessEngine

tomuta:


Games/Chess/ChessWidget.cpp
```diff
@@ -409,7 +409,9 @@ void ChessWidget::input_engine_move()
     if (drag_was_enabled)
         set_drag_enabled(false);
 
+    set_override_cursor(Gfx::StandardCursor::Wait);
     m_engine->get_best_move(board(), 4000, [this, drag_was_enabled](Chess::Move move) {
+        set_override_cursor(Gfx::StandardCursor::Arrow);
```
bcoles, line 414: I'm not sure if it makes a difference, but I believe `Gfx::StandardCursor::None` is the preferred way to restore the default cursor, rather than explicitly forcing an arrow cursor.

awesomekling: Yes that's right.

tomuta: Learned something new :)

PR #4507, opened 2020-12-23
LibJS: Implement String.prototype.endsWith()

XavierCooney:
The relevant spec: https://tc39.es/ecma262/#sec-string.prototype.endswith.


Libraries/LibJS/Runtime/StringPrototype.cpp
```diff
@@ -185,6 +186,40 @@ JS_DEFINE_NATIVE_FUNCTION(StringPrototype::starts_with)
     return Value(string.substring(start, search_string_length) == search_string);
 }
 
+JS_DEFINE_NATIVE_FUNCTION(StringPrototype::ends_with)
+{
+    auto string = ak_string_from(vm, global_object);
+    if (string.is_null())
+        return {};
+    if (!vm.argument_count())
```
linusg, line 194: This is not part of the spec and therefore incorrect.

linusg (changes requested):
If you don't want to implement steps 3 and 4, please at least add a FIXME comment. :)

Libraries/LibJS/Runtime/StringPrototype.cpp
```diff
@@ -185,6 +186,40 @@ JS_DEFINE_NATIVE_FUNCTION(StringPrototype::starts_with)
...
+
+    auto search_string = vm.argument(0).to_string(global_object);
+    if (vm.exception())
+        return {};
+
+    auto string_length = string.length();
+    auto search_string_length = search_string.length();
+
+    size_t end = string_length;
+    if (vm.argument_count() > 1) {
```
linusg, line 205: This is differently in the spec, there's a difference between argument is undefined and argument doesn't exist

Libraries/LibJS/Runtime/StringPrototype.cpp
```diff
@@ -185,6 +186,40 @@ JS_DEFINE_NATIVE_FUNCTION(StringPrototype::starts_with)
...
+        auto number = vm.argument(1).to_number(global_object);
+        if (vm.exception())
+            return {};
+        end = min(string_length, number.to_size_t(global_object));
```
linusg, line 209: Please do not use to_size_t anymore, it's not part of the spec - use the function that's actually being used instead :)

XavierCooney:
Thanks, things should be corrected now, and I've added a few more tests for the cases I got wrong :).

Libraries/LibJS/Runtime/StringPrototype.cpp
```diff
@@ -185,6 +186,77 @@ JS_DEFINE_NATIVE_FUNCTION(StringPrototype::starts_with)
...
+static bool is_regexp(VM& vm, GlobalObject& global_object, Value possible_regexp)
+{
+    if (!possible_regexp.is_object())
+        return false;
+
+    auto matcher = possible_regexp.as_object().get(vm.well_known_symbol_match());
+    if (vm.exception())
+        return false;
+    if (!matcher.is_empty() && !matcher.is_undefined()) {
+        return !matcher.to_boolean();
```
linusg, line 198: When the spec says "If matcher is not undefined, return ! ToBoolean(matcher).", the "!" just means the operation cannot throw, and not exception check is needed (yes, it's a confusing way to express this):

https://tc39.es/ecma262/#sec-returnifabrupt-shorthands

> Similarly, prefix ! is used to indicate that the following invocation of an abstract or syntax-directed operation will never return an abrupt completion and that the resulting Completion Record's [[Value]] field should be used in place of the return value of the operation.

So:

```suggestion
        return matcher.to_boolean();
```

Libraries/LibJS/Runtime/StringPrototype.cpp
```diff
@@ -185,6 +186,77 @@ JS_DEFINE_NATIVE_FUNCTION(StringPrototype::starts_with)
...
+    }
+
+    auto regexp_as_object = possible_regexp.to_object(global_object);
+    if (vm.exception())
+        return false;
+    if (regexp_as_object->is_regexp_object()) {
+        return true;
+    }
```
linusg, line 206: We already checked that `possible_regexp` is an object, so we can use `as_object()` and don't need an exception check:

```suggestion
    if (possible_regexp.as_object(global_object).is_regexp_object())
        return true;
```

Libraries/LibJS/Runtime/StringPrototype.cpp
```diff
@@ -185,6 +186,77 @@ JS_DEFINE_NATIVE_FUNCTION(StringPrototype::starts_with)
...
+
+JS_DEFINE_NATIVE_FUNCTION(StringPrototype::ends_with)
+{
+    auto string = ak_string_from(vm, global_object);
+    if (string.is_null())
+        return {};
+
+    auto search_string_value = vm.argument(0);
+    if (vm.exception())
+        return {};
```
linusg, line 219: No need for an exception check here :)

```suggestion
    auto search_string_value = vm.argument(0);
```

Libraries/LibJS/Runtime/StringPrototype.cpp
```diff
@@ -185,6 +186,77 @@ JS_DEFINE_NATIVE_FUNCTION(StringPrototype::starts_with)
...
+
+    bool search_is_regexp = is_regexp(vm, global_object, search_string_value);
+    if (vm.exception())
+        return {};
+    if (search_is_regexp) {
+        vm.throw_exception<TypeError>(global_object, ErrorType::IsNotA, "searchString", "string, but a regular expression");
+        return {};
+    }
+
+    auto search_string = vm.argument(0).to_string(global_object);
```
linusg, line 229:
```suggestion
    auto search_string = search_string_value.to_string(global_object);
```

Libraries/LibJS/Runtime/StringPrototype.cpp
```diff
@@ -185,6 +186,77 @@ JS_DEFINE_NATIVE_FUNCTION(StringPrototype::starts_with)
...
+            return {};
+        pos = clamp(pos_as_double, static_cast<double>(0), static_cast<double>(string_length));
+    }
+
+    if (search_string_length == 0) {
+        return Value(true);
+    }
+    if (pos < search_string_length) {
+        return Value(false);
+    }
```
linusg, line 254: Not mandatory, but you can skip the curly braces here:

```suggestion
    if (search_string_length == 0)
        return Value(true);
    if (pos < search_string_length)
        return Value(false);
```

Libraries/LibJS/Runtime/StringPrototype.cpp
```diff
@@ -185,6 +186,77 @@ JS_DEFINE_NATIVE_FUNCTION(StringPrototype::starts_with)
     return Value(string.substring(start, search_string_length) == search_string);
 }
 
+static bool is_regexp(VM& vm, GlobalObject& global_object, Value possible_regexp)
```
linusg, line 189: It [looks like the `RegExp` constructor will use this as well](https://tc39.es/ecma262/#sec-regexp-pattern-flags), so let's move it to `Value.cpp`/`Value.h`, next to all the other `is_*` functions (even though those don't take a vm and GlobalObject).

linusg (commented):
Hey, this looks great! I got a couple more suggestions, but well done :)

linusg (approved):
Awesome!

PR #4510, opened 2020-12-23
CMake: Remove file globbing in AK/Tests

ldm5180:
Problem:
- File globbing is performed at the time of build system
  generation. Any files which are not there at that time are not
  included. So, when a new file is added it is not built unless the
  build system is recreated.

Solution:
- Remove globbing from AK/Tests directory in favor of explicitly
  listing the files.

### DIFF
--- a/AK/Tests/CMakeLists.txt
+++ b/AK/Tests/CMakeLists.txt
@@ -1,4 +1,44 @@
-file(GLOB AK_TEST_SOURCES CONFIGURE_DEPENDS "*.cpp")
+set(AK_TEST_SOURCES
+    TestAllOf.cpp
+    TestArray.cpp
+    TestAtomic.cpp
+    TestBase64.cpp
+    TestBinarySearch.cpp
+    TestBitmap.cpp
+    TestByteBuffer.cpp
+    TestChecked.cpp
+    TestCircularDeque.cpp
+    TestCircularDuplexStream.cpp
+    TestCircularQueue.cpp
+    TestDistinctNumeric.cpp
+    TestEndian.cpp
+    TestFormat.cpp
+    TestHashFunctions.cpp
+    TestHashMap.cpp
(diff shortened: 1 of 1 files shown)


ldm5180:
This was resurrected from #3733. The original branch was deleted, so github won't let me "reopen".

alimpfard:
Didn't we go around this issue using `CONFIGURE_DEPENDS`?

ldm5180:
@alimpfard New comment from @nico here: https://github.com/SerenityOS/serenity/pull/3733#issuecomment-750326946

alimpfard:
Re-resolving globs here and there doesn't seem like a particularly big issue to me, but either way works I guess.

asynts:
Currently, the C++ compiler is the bottleneck for build times, running CMake only takes a second or two.

ldm5180:
@asynts Yes, re-running cmake is not a bottleneck, but it will get progressively worse over time.

asynts:
@ldm5180 Well we could always do this in the future when it becomes a problem, but especially for the tests it is really helpful if all tests sources are found automatically.

I mean it's just a matter of time until someone adds a test source, forgets to add it to CMake and thinks that all the tests were run and passed. With other sources this isn't that important because there will be linker and compiler errors for that...

ldm5180:
@asynts I submitted PR #4523 which will check that all the test files are included in the CMakeLists.txt during CI. Does this alleviate that concern?

alimpfard:
Not gonna lie, this feels like needlessly going the long way just because "best practices" say so...
we can detect what files are missing, and we're completely sure that they are in fact, missing from the list - so why can't we just add the files automatically? 
which then turns into doing what cmake's glob does, but in a less conventional manner.

Still not convinced two/three seconds of build time is worth making the dev spend the same amount of time adding some name to some list

Issue #4418, opened 2020-12-15
SystemServer: crashes on boot when built on macOS


diversys:
This is bcabbbda8b0e4446c3d38b2772f3dee5b3dcc4cf, completely fresh build, running in qemu under macOS.

```
[init_stage2(1:1)]: Loading kernel symbol table...
[init_stage2(1:1)]: exec(/bin/SystemServer): Using program interpreter /usr/lib/Loader.so
[NetworkTask(5:5)]: NetworkTask: LoopbackAdapter network adapter found: hw=13:55:02:09:55:aa address=127.0.0.1 netmask=255.0.0.0 gateway=0.0.0.0
[NetworkTask(5:5)]: NetworkTask: E1000NetworkAdapter network adapter found: hw=52:54:00:12:34:56 address=10.0.2.15 netmask=255.255.255.0 gateway=10.0.2.2
[NetworkTask(5:5)]: NetworkTask: Enter main loop.
DynamicObject: DYNAMIC tag handling not implemented for DT_RPATH
USERSPACE(4) ASSERTION FAILED: m_data != nullptr
../Libraries/LibC/stdio.cpp:537
[SystemServer(4:4)]: Terminating SystemServer(4) due to signal 6
[FinalizerTask(3:3)]: 0xdeadc0de
0xdeadc0de
0xdeadc0de
0xdeadc0de
0xdeadc0de
0xdeadc0de
0xdeadc0de
0x080202c6
0x08000a1d
0x08019788
0x0802164a
0x08021993
0x08021c9a
0x08021cde
0x08021d05
0x0802a0d3
0x0802a10a
0x08016b14
0x08016c46
0x08016e93
0x0801618a
0x080011f0
0x08001611
0x08001da8

[FinalizerTask(3:3)]: Generating coredump for pid: 4
[FinalizerTask(3:3)]: RTC: Year: 2020, month: 12, day: 15, hour: 9, minute: 29, second: 21
[FinalizerTask(3:3)]: Could not create coredump
```

diversys:
It looks like I'm the only one (for now) hitting this. Perhaps it has something to do with macOS host gcc?

ChrisThrasher:
I'm experiencing an identical problem on macOS Catalina.

diversys:
macOS Mojave 10.14.6 here.

vkoskiv:
Hitting this too on Catalina.

ChrisThrasher:
[...]

PR #4514, opened 2020-12-23
LibELF: Add missing case for DT_RPATH

bromaniac:
SystemServer crashes on boot when built on macOS when
this is missing. See issue [#4418](https://github.com/SerenityOS/serenity/issues/4418).

This is the laziest way to fix this. DT_RPATH should probably be handled
correctly but I don't know enough about the ELF format to do this.

### DIFF
--- a/Libraries/LibELF/DynamicObject.cpp
+++ b/Libraries/LibELF/DynamicObject.cpp
@@ -160,6 +160,10 @@ void DynamicObject::parse()
         case DT_NEEDED:
             // We handle these in for_each_needed_library
             break;
+        case DT_RPATH:
+        case DT_RUNPATH:
+            // FIXME: Implement RPATH/RUNPATH handling in DynamicLoader
+            break;
         default:
             dbgprintf("DynamicObject: DYNAMIC tag handling not implemented for DT_%s\n", name_for_dtag(entry.tag()));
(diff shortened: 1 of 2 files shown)


Libraries/LibELF/DynamicObject.cpp
```diff
@@ -160,6 +160,8 @@ void DynamicObject::parse()
         case DT_NEEDED:
             // We handle these in for_each_needed_library
             break;
+        case DT_RPATH:
```
ADKaster, line 163: Pretty odd that the compiler/linker is generating DT_RPATH dynamic headers when built with macos? I would expect the configure flags to the gcc cross-compiler and binutils would be the same.

Should probably handle DT_RPATH and DT_RUNPATH in the same way. The proper way to implement this tag would be.. creating a list of strings for each string pointed to by entries in that section, and exposing it to the DynamicLoader for use in when it loads dependencies?

Anyway, I think if we want the "unblock MacOS" patch instead of the real impl it should look something like:
```
        case DT_RPATH:
        case DT_RUNPATH:
            // FIXME: Implement RPATH/RUNPATH handling in DynamicLoader
            break;
```

bromaniac:
I messed this PR up completely so I'll start over.

I will create a new PR to unblock macOS like you suggested and also look into fixing this proper.

PR #4516, opened 2020-12-23
Userland: Add pmap utility

bcoles:
The Linux [pmap man page](https://linux.die.net/man/1/pmap) helpfully states there is no existing standard output format:

> No standards apply, but pmap looks an awful lot like a SunOS command. 

So we shouldn't have to care too much about consistency with other systems (is any software likely to parse output from `pmap` ?)

pmap $$

Linux output format for comparison:

pmap $$

I opted to keep the default output simple (very similar to default Linux output format) and include everything that might be useful (similar to the `SystemMonitor` output) in an extended `-x` output format.

pmap -x $$

### DIFF
--- a/Base/usr/share/man/man1/pmap.md
+++ b/Base/usr/share/man/man1/pmap.md
@@ -1,0 +1,27 @@
+## Name
+
+pmap - print memory map of a process
+
+## Synopsis
+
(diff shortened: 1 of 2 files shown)


awesomekling:
Love it

Issue #4476, opened 2020-12-21
Shell (?): Redirections are broken inside Serenity


alimpfard:
repro:
```sh
$ echo hello > test
$ cat test
# <nothing here>
```

Note that this works on the lagom version on linux.

danboid:
gh needs an Ouch! response.

XavierCooney:
Not sure how much this helps, but for some reasoning making `echo` close `stdout` before it exits fixes this issue. The issue also doesn't appear if you try `yes > test` and then ctrl+c. The shell seems to make the correct `dup2` call, but there's some issue with closing the file or something related to that.

alimpfard:
Fixed by #4518.

PR #4518, opened 2020-12-24
DynamicLoader: Call libc's exit when exitting, to flush standard streams

ccapitalK:
This fixes a regression introduced with dynamic libraries, where the following command stopped working due to us no longer flushing standard streams on exit.

```sh
$ echo foo > bar
```


alimpfard:
Fixes #4476 ?

itamar8910:
ooh good fix, thanks

PR #4520, opened 2020-12-24
Meta: Set two minute timeout for CMake tests

Lubrsi:
CMake tests usually takes ~40 seconds. However, sometimes it deadlocks
and is only timed out after the 6 hour time limit.

Let's set a 2 minute timeout to make it fail sooner. 2 minutes instead
of 1 for good measure.

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -79,6 +79,7 @@ jobs:
     - name: Run CMake tests
       working-directory: ${{ github.workspace }}/Build
       run: CTEST_OUTPUT_ON_FAILURE=1 ninja test
+      timeout-minutes: 2
     - name: Run JS tests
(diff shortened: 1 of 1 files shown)


alimpfard:
Interesting...is this about Shell-backgrounding?
I set a 10m timeout or something to all shell tests, but apparently cmake isn't respecting that if it's still taking the whole 6h to timeout...

Lubrsi:
I believe so. It happened to this:
https://github.com/SerenityOS/serenity/runs/1603557880?check_suite_focus=true
It does timeout after 10 seconds, but then it tries to print the debug output, and it takes hours and hours and hours...

alimpfard:
I noticed that [here](https://github.com/SerenityOS/serenity/commit/eb7a94364188740d67278c74bc17d4223697becd#commitcomment-45159355) too...
Hopefully your patch fixes that.

PR #4522, opened 2020-12-24
Everywhere: void arguments to C functions

ldm5180:
Problem:
- C functions with no arguments require a single `void` in the argument list.

Solution:
- Put the `void` in the argument list of functions in C header files.

### DIFF
--- a/AK/Tests/TestSourceGenerator.cpp
+++ b/AK/Tests/TestSourceGenerator.cpp
@@ -44,9 +44,9 @@ TEST_CASE(generate_c_code)
     SourceGenerator generator { builder };
     generator.set("name", "foo");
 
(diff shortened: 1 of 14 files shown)


bugaevc:
You don't have to use `(void)` in C++ code though :slightly_smiling_face:

It's only the C headers (LibC, LibPthread, LibM) that need this treatment.

ldm5180:
I think I only changed back the ones that `extern C` and in those headers. Do you see anything I can remove?

bugaevc:
I don't think `extern "C"` affects it — only actual code that may be compiled with a C compiler need `(void)`.

That being said, I don't have anything against using `(void)` more either, so it doesn't matter.

PR #4526, opened 2020-12-24
LibProtocol/ProtocolServer/Browser/pro: Make ProtocolServer stream downloaded data when and where possible

alimpfard:
What:
- Currently, ProtocolServer buffers the downloaded data until the download is finished, and then passes it along (by copying it into a shared buffer no less!) to its user
- In a good deal of situations, this is wasteful, and in some other situations, it makes no sense (e.g. when downloading a huge file to disk via `DownloadWidget` or `pro`)
- This PR makes ProtocolServer shove data down a pipe, and its user read it from the pipe (as the download is progressing) - if possible

Why:
- pipe `pro`'s output to `/dev/null` and laugh like a maniac
- Actually download huge files
- Open up the path for people that want to have fun with streaming downloads
- Make the API better in general
- Copy less stuff!

When:
- After some cleanups and some more extensive tests :)

cc @bugaevc @awesomekling and maybe @deoxxa (you asked to see the resulting API - this is most likely what it looks like)

### DIFF
--- a/AK/ByteBuffer.h
+++ b/AK/ByteBuffer.h
@@ -123,6 +123,7 @@ public:
     static ByteBuffer create_uninitialized(size_t size) { return ByteBuffer(ByteBufferImpl::create_uninitialized(size)); }
     static ByteBuffer create_zeroed(size_t size) { return ByteBuffer(ByteBufferImpl::create_zeroed(size)); }
     static ByteBuffer copy(const void* data, size_t size) { return ByteBuffer(ByteBufferImpl::copy(data, size)); }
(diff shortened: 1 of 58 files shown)


tomuta:
I like this, also this would probably pave the way to support HTTP/2...

alimpfard:
HTTP/2 huh...it seems so far away, we don't even have support for persistent connections in LibHTTP.

alimpfard:
This thing feels stable enough to undergo review now.
pls no hate, pls many reviews

tomuta:
> HTTP/2 huh...it seems so far away, we don't even have support for persistent connections in LibHTTP.

HTTP/2 for the most part is quite straight forward. It's just binary frames and you have the concept of streams. And it's a persistent socket connection.

PR #4527, opened 2020-12-24
Kernel: Introducing basic DevFS & new partition subsystem

supercomputer7:
The first change (Get a reference of `UserOrKernelBuffer` instead of a pointer to it) is something I'm pretty sure we will want to merge.

Also, I introduce the DevFS - a solution for populating device nodes in `/dev` during bootup.

cc @awesomekling @bugaevc


Kernel/FileSystem/BlockBasedFileSystem.h
```diff
@@ -29,8 +29,15 @@
 #include <Kernel/FileSystem/FileBackedFileSystem.h>
 
 namespace Kernel {
+struct CacheEntry {
```
awesomekling, line 32: Please put this inside the `BlockBasedFS` class so we don't add a `CacheEntry` class in the `Kernel` namespace that everyone can see. :)

Kernel/Devices/Device.h
```diff
@@ -57,6 +57,9 @@ class Device : public File {
     uid_t uid() const { return m_uid; }
     uid_t gid() const { return m_gid; }
 
+    // ^Device
+    virtual mode_t permission_flags() const = 0;
```
bugaevc, line 61: No point in `// ^Device` inside `Device` itself 😀

Also perhaps call it just `mode()`?

supercomputer7: I guess it could be `mode`, or maybe `required_mode`?

supercomputer7:
I tested it after building an image with GRUB- DevFS exposes `hda1` block device which is the partition we booted from.
Next thing to do is to fix the `DiskPartition` class to use dynamic minor number for other partitions.

Services/SystemServer/main.cpp
```diff
@@ -84,6 +85,86 @@ static void parse_boot_mode()
...
+    // FIXME: Find a better way to all of this stuff, without hardcoding all of this!
+
+    int rc = mount(-1, "/dev", "dev", 0);
+    if (rc != 0) {
+        ASSERT_NOT_REACHED();
+    }
+    rc = mkdir("/dev/pts", 0755);
+    if (rc != 0) {
+        ASSERT_NOT_REACHED();
+    }
```
awesomekling, line 99: Can we avoid doing this in userspace? Can DevFS create this automatically instead?

Services/SystemServer/main.cpp
```diff
@@ -84,6 +85,86 @@ static void parse_boot_mode()
...
+    rc = symlink("/proc/self/fd/1", "/dev/stdout");
+    if (rc < 0) {
+        ASSERT_NOT_REACHED();
+    }
+    rc = symlink("/proc/self/fd/2", "/dev/stderr");
+    if (rc < 0) {
+        ASSERT_NOT_REACHED();
+    }
```
awesomekling, line 115: These are not necessary. ProcFS will generate them :)

Services/SystemServer/main.cpp
```diff
@@ -84,6 +85,86 @@ static void parse_boot_mode()
...
+    rc = symlink("/dev/random", "/dev/urandom");
+    if (rc < 0) {
+        ASSERT_NOT_REACHED();
+    }
```
awesomekling, line 119: Can we avoid doing this in userspace? Can DevFS create this automatically instead?

Services/SystemServer/main.cpp
supercomputer7: I talked with Sergey about this, and considering the fact that we already allow `SystemServer` to use `chown`, (so DevFS isn't read-only anymore), it was an idea to avoid doing such things in the kernel code also.
But sure, if you think we should do that in the kernel space, it can be done without any trouble :)

Services/SystemServer/main.cpp
supercomputer7: I guess we could create a symlink in the DevFS code. As I said in the comment above, it's up to you :)

Services/SystemServer/main.cpp
bugaevc: This creates `/dev/stdin -> /proc/self/fd/0`, not the other way around 🙂

supercomputer7: oh I see, so `build-root-filesystem.sh` created these links for nothing, right?

bugaevc: No, no, I'm replying to Andreas; `build-root-filesystem.sh` was doing everything correctly, and so does this code.

awesomekling: Oh yeah lol, my mistake :smile:

Kernel/FileSystem/BlockBasedFileSystem.cpp
```diff
@@ -174,6 +171,29 @@ int BlockBasedFS::write_block(unsigned index, const UserOrKernelBuffer& data, si
...
+    ASSERT(nread.value() == block_size());
+    entry.has_data = true;
+    return true;
+}
+
+BlockBasedFS::CacheEntry& BlockBasedFS::cache_block(unsigned index) const
+{
+    auto& entry = cache().get(index);
+    if (!entry.has_data) {
+        force_cache_block(index);
```
awesomekling, line 192: We're forgetting to check the return value of `force_cache_block` here. If there's an error we need to handle it.

Kernel/Storage/Partition/DiskPartitionMetadata.h
```diff
@@ -0,0 +1,55 @@
...
+    Optional<String> name() const;
+    Optional<ByteBuffer> partition_type() const;
+    Optional<ByteBuffer> unique_guid() const;
+
+private:
+    u64 m_start_block;
+    u64 m_end_block;
+    ByteBuffer m_partition_type;
+    ByteBuffer m_unique_guid;
+    u64 m_attributes;
```
awesomekling, line 53: Please always initialize primitive members explicitly.
```cpp
u64 m_attributes { 0 };
```
Notice how one of the constructors forgets to initialize this member :)

Kernel/FileSystem/BlockBasedFileSystem.cpp
supercomputer7: Oh right, if we can't cache here I'd assume that we should assert.

Kernel/Storage/Partition/EBRPartitionTable.h
```diff
@@ -33,39 +33,18 @@
...
-
+struct EBRPartitionHeader;
+class EBRPartitionTable : public MBRPartitionTable {
 public:
-    explicit EBRPartitionTable(NonnullRefPtr<BlockDevice>);
     ~EBRPartitionTable();
 
-    bool initialize();
-    RefPtr<DiskPartition> partition(unsigned index);
+    static NonnullOwnPtr<EBRPartitionTable> try_to_initialize(const StorageDevice&);
```
awesomekling, line 41: `try_to_initialize` sounds like something that can fail, yet this function returns a `NonnullOwnPtr`. So how is failure signalled to the caller? Normally, something that can fail would return an `OwnPtr` (with `nullptr` meaning failure.)

Kernel/Storage/Partition/DiskPartitionMetadata.h
supercomputer7: Thanks for catching this, I'll try to pay attention to such things later on :)

Kernel/FileSystem/BlockBasedFileSystem.cpp
awesomekling: Hmm.. if this is dealing with hardware, then hardware can fail at any time, so we need to handle it more gracefully than just asserting I think

Kernel/Storage/Partition/EBRPartitionTable.h
supercomputer7: The caller should check if the `PartitionTable` is valid by calling `table->is_valid()`. Also, if you try to get partitions from it by using the `partition` method, you will just get no value (I use the `Optional` container for such purpose) :)

supercomputer7: I changed the code to use `AK::Result` and `OwnPtr`. Seems fine now :)

Kernel/FileSystem/BlockBasedFileSystem.cpp
supercomputer7: I think I figured this out, I use `KResultOr` to indicate a failure.

PR #4532, opened 2020-12-25
LibC: Enable compiler warnings for printf/scanf/strftime format strings

ccapitalK:
This PR makes the compiler aware that our printf functions use printf format strings, and makes it a compile time error to have an incorrect format string (where possible).


ccapitalK:
Hmm, It appears this won't work because we have custom format types.

ADKaster:
> Hmm, It appears this won't work because we have custom format specifiers (%b).

Probably two solutions:
- 1, remove non-standard format specifiers to conform to C++20/C17 etc.
- 2, patch this file https://github.com/gcc-mirror/gcc/blob/releases/gcc-10/gcc/c-family/c-format.c to add %b for bool.

Libraries/LibGUI/BoxLayout.cpp
```diff
@@ -48,7 +48,7 @@ void BoxLayout::run(Widget& widget)
     should_log = true;
 #endif
     if (should_log)
-        dbgprintf("BoxLayout: running layout on %s{%p}, entry count: %d\n", widget.class_name(), &widget, m_entries.size());
+        dbgprintf("BoxLayout: running layout on %s{%p}, entry count: %lu\n", widget.class_name(), &widget, m_entries.size());
```
ADKaster, line 51: these .size() are size_t, so the format specifier should be `%zu`

ccapitalK:
> > Hmm, It appears this won't work because we have custom format specifiers (%b).
> 
> Probably two solutions:
> 
>     * 1, remove non-standard format specifiers to conform to C++20/C17 etc.
> 
>     * 2, patch this file https://github.com/gcc-mirror/gcc/blob/releases/gcc-10/gcc/c-family/c-format.c to add %b for bool.

I'm not particularly fond of (1), since I think serenity should retain the freedom to do stuff like the '%b' specifier, but I'm also not sure about (2), doesn't the CI use clang? I'm not sure if just patching that file would be enough, also it would require people to rebuild the toolchain.

ADKaster:
Looks like @awesomekling already chose option 1) and committed it yesterday :D

cb2c8f71f44f8c628f57e6ac3cadca57642c7682
ed5c26d698f586d0d5036b29cd88f6b91314314c

I like this patch, but I think it doesn't go far enough :). `-Wformat=2` is pretty sweet imo, and might want to add the format attributes for scanf and strftime as well. Using `-Wformat=2` would requires a these changes as well.

```
diff --git a/Userland/Tests/LibC/snprintf-correctness.cpp b/Userland/Tests/LibC/snprintf-correctness.cpp
index 9f9b43c49..3089419df 100644
--- a/Userland/Tests/LibC/snprintf-correctness.cpp
+++ b/Userland/Tests/LibC/snprintf-correctness.cpp
@@ -32,6 +32,8 @@
 #include <ctype.h>
 #include <stdio.h>
 
+#pragma GCC diagnostic ignored "-Wformat-nonliteral"
+
 struct Testcase {
     const char* dest;
     size_t dest_n;
diff --git a/Userland/cal.cpp b/Userland/cal.cpp
index e2d2f7ca6..df482c88e 100644
--- a/Userland/cal.cpp
+++ b/Userland/cal.cpp
@@ -155,13 +155,13 @@ int main(int argc, char** argv)
             insert_month_to_print(0, i++, year);
             insert_month_to_print(1, i++, year);
             insert_month_to_print(2, i, year);
-            printf(print_buffer);
+            printf("%s", print_buffer);
             printf("\n");
             clean_buffers();
         }
     } else {
         insert_month_to_print(0, month, year);
-        printf(print_buffer);
+        printf("%s", print_buffer);
         printf("\n\n");
         clean_buffers();
     }
diff --git a/Userland/seq.cpp b/Userland/seq.cpp
index e47d94fe0..a0cfe5787 100644
--- a/Userland/seq.cpp
+++ b/Userland/seq.cpp
@@ -30,7 +30,7 @@
 #include <string.h>
 #include <unistd.h>
 
-const char* g_usage = R"(Usage:
+const char* const g_usage = R"(Usage:
     seq [-h|--help]
     seq LAST
     seq FIRST LAST
```

Libraries/LibGUI/BoxLayout.cpp
ccapitalK: Good catch, I'll change these.

ccapitalK:
> Looks like @awesomekling already chose option 1) and committed it yesterday :D
> 
> [cb2c8f7](https://github.com/SerenityOS/serenity/commit/cb2c8f71f44f8c628f57e6ac3cadca57642c7682)
> [ed5c26d](https://github.com/SerenityOS/serenity/commit/ed5c26d698f586d0d5036b29cd88f6b91314314c)
> 
> I like this patch, but I think it doesn't go far enough :). `-Wformat=2` is pretty sweet imo, and might want to add the format attributes for scanf and strftime as well. Using `-Wformat=2` would requires a these changes as well.
> 
> ```
> diff --git a/Userland/Tests/LibC/snprintf-correctness.cpp b/Userland/Tests/LibC/snprintf-correctness.cpp
> index 9f9b43c49..3089419df 100644
> --- a/Userland/Tests/LibC/snprintf-correctness.cpp
> +++ b/Userland/Tests/LibC/snprintf-correctness.cpp
> @@ -32,6 +32,8 @@
>  #include <ctype.h>
>  #include <stdio.h>
>  
> +#pragma GCC diagnostic ignored "-Wformat-nonliteral"
> +
>  struct Testcase {
>      const char* dest;
>      size_t dest_n;
> diff --git a/Userland/cal.cpp b/Userland/cal.cpp
> index e2d2f7ca6..df482c88e 100644
> --- a/Userland/cal.cpp
> +++ b/Userland/cal.cpp
> @@ -155,13 +155,13 @@ int main(int argc, char** argv)
>              insert_month_to_print(0, i++, year);
>              insert_month_to_print(1, i++, year);
>              insert_month_to_print(2, i, year);
> -            printf(print_buffer);
> +            printf("%s", print_buffer);
>              printf("\n");
>              clean_buffers();
>          }
>      } else {
>          insert_month_to_print(0, month, year);
> -        printf(print_buffer);
> +        printf("%s", print_buffer);
>          printf("\n\n");
>          clean_buffers();
>      }
> diff --git a/Userland/seq.cpp b/Userland/seq.cpp
> index e47d94fe0..a0cfe5787 100644
> --- a/Userland/seq.cpp
> +++ b/Userland/seq.cpp
> @@ -30,7 +30,7 @@
>  #include <string.h>
>  #include <unistd.h>
>  
> -const char* g_usage = R"(Usage:
> +const char* const g_usage = R"(Usage:
>      seq [-h|--help]
>      seq LAST
>      seq FIRST LAST
> ```

Oh cool, I didn't notice that. Yeah, I'll just go ahead and the changes for the other format functions as well.

ccapitalK:
Build appears to pass now, I think it's ready to merge.

awesomekling:
Cool :)

Issue #4530, opened 2020-12-24
Kernel: Unveil picks first match, not best match 


ADKaster:
Terminal adds its unveils in this order:

```
[Terminal(22:22)]: Adding /res to unveiled paths with 1 permissions
[Terminal(22:22)]: Adding /bin to unveiled paths with 1 permissions
[Terminal(22:22)]: Adding /bin/Terminal to unveiled paths with 4 permissions
[Terminal(22:22)]: Adding /bin/utmpupdate to unveiled paths with 4 permissions
[Terminal(22:22)]: Adding /tmp/portal/launch to unveiled paths with 3 permissions
[Terminal(22:22)]: Adding /home/anon/.config/Terminal.ini to unveiled paths with 13 permissions
```
`/bin` is added with read perms, and /bin/utmpupdate is added with execute permissions.

During shutdown, it attempts to call utmpupdate, which it should be able to, it was unveiled with execute. However, /bin is selected as the matching unveiled path instead, and the execute fails.

```
Terminal(22): Exiting terminal, updating utmp
[Terminal(27:27)]: Found unveiled path /bin for /bin/utmpupdate
[Terminal(27:27)]: Unveiled path /bin requested with execute permissions, unveil perms: 1
[Terminal(27:27)]: Rejecting path '/bin/utmpupdate' since it hasn't been unveiled with 'x' permission.
[Terminal(27:27)]: 0xc014fb93  Kernel::VFS::validate_path_against_process_veil(AK::StringView, int) +1221
```

Seems the loop in VFS::find_matching_unveiled_paths is too naive, and always chooses the first match. It should pick the longest matching unveiled path instead. 

cc @bugaevc

bugaevc:
> Seems the loop in VFS::find_matching_unveiled_paths is too naive, and always chooses the first match. It should pick the longest matching unveiled path instead.

But on the other hand, perhaps the app should unveil paths in the order it wants them to be checked?

What does OpenBSD do?

ADKaster:
My C is a bit rusty, but it *looks* like they store the unveiled paths in a red-black tree and walk the unveil nodes from the exact path match to the root node to see if any of the parent nodes are unveiled.

https://github.com/openbsd/src/blob/master/sys/kern/kern_unveil.c#L281

alimpfard:
Perhaps it could build a prefix tree with path components as nodes?
it would also make insertion and matching even faster the rb-tree that openbsd is using?

[...]

PR #4544, opened 2020-12-26
(PoC) Kernel: Implement unveil() as a prefix-tree

alimpfard:
For #4530.

### DIFF
--- a/AK/Tests/TestTrie.cpp
+++ b/AK/Tests/TestTrie.cpp
@@ -1,0 +1,83 @@
+/*
+ * Copyright (c) 2020, the SerenityOS developers.
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ *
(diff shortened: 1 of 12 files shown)


bgianfo:
Does it make sense to add new unit tests targetting the new AK/Trie.h implementation directly?

alimpfard:
Ah definitely. thanks for reminding me~

awesomekling:
This looks very nice!

alimpfard:
I don't think there's anything wrong with it functionality-wise, basic-ish testing with the (included) `test-unveil` does show reasonable behaviour, so I don't think there's anything else blocking it from my side.

AK/Tests/TestTrie.cpp
```diff
@@ -0,0 +1,83 @@
...
+    Array<int, 64> input;
+    for (size_t i = 0; i < input.size(); ++i)
+        input[i] = i;
+
+    // Iteration order is preorder (order between adjacent nodes is not defined, but parents come before children)
+    // in this case, the tree is linear.
+    size_t i = 0;
+    for (auto& node : bunch_of_numbers) {
+        EXPECT_EQ(input[i], node.value());
+        ++i;
```
alimpfard, line 79: :facepalm:
This doesn't test anything damnit

PR #4545, opened 2020-12-26
Kernel: Tag more methods and types as [[nodiscard]]

bgianfo:
Tag methods at where not obvserving the return value is an obvious error
with [[nodiscard]] to catch potential future bugs.

### DIFF
--- a/Kernel/SpinLock.h
+++ b/Kernel/SpinLock.h
@@ -58,7 +58,7 @@ public:
         Processor::current().leave_critical(prev_flags);
     }
 
-    ALWAYS_INLINE bool is_locked() const
+    [[nodiscard]] ALWAYS_INLINE bool is_locked() const
     {
         return m_lock.load(AK::memory_order_relaxed) != 0;
     }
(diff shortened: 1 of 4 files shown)


Kernel/SpinLock.h
```diff
@@ -126,22 +126,23 @@ class RecursiveSpinLock {
...
+class [[nodiscard]] ScopedSpinLock
+{
     AK_MAKE_NONCOPYABLE(ScopedSpinLock);
 
 public:
     ScopedSpinLock() = delete;
     ScopedSpinLock& operator=(ScopedSpinLock&&) = delete;
 
-    ScopedSpinLock(LockType& lock)
+    ScopedSpinLock(LockType & lock)
```
alimpfard, line 137: :thinking: that be weird.
I do remember clang-format having issues when classes have [[attributes]], maybe `// clang-format off` the class decl?

bgianfo: Sure, I can try... auto formatters are great until they go haywire 😁🔫

bgianfo: I just added a new NO_DISCARD macro to be able to  hide this behind, clang-format off didn't work, and still mangled the rest of the file.

PR #4547, opened 2020-12-26
LibJS: Implement Array.prototype.sort()

XavierCooney:
An implementation for `Array.prototype.sort()`. Quicksort can't be used in all cases as the algorithm must be stable. There's a number of potential performance improvements which I've noted as `FIXME`s in the code. I've tried to adhere to the spec as closely as possible, but I may have missed something 😄.


Libraries/LibJS/Tests/builtins/Array/Array.prototype.sort.js
```diff
@@ -0,0 +1,117 @@
...
+        // sorting should be stable
+        arr = [{sorted_key: 2, other_property: 1}, {sorted_key: 2, other_property: 2}, {sorted_key: 1, other_property: 3}];
+        arr.sort((a, b) => a.sorted_key - b.sorted_key);
+        expect(arr[1].other_property == 1);
+        expect(arr[2].other_property == 2);
+    });
+
+    test("that it makes no unnecessary calls to compare function", () => {
+        // since we have no expect(...).not.toHaveBeenCalled(), just throw an error.
+        expect_no_call_compare_fn = function(a, b) { throw new Error('compare func called'); }
```
linusg, line 60: We have `expect().fail()` though :)

Also, please don't use snake_case for JS code.

```suggestion
        function expectNoCallCompareFn(a, b) { expect().fail(); }
```

Libraries/LibJS/Tests/builtins/Array/Array.prototype.sort.js
```diff
@@ -0,0 +1,117 @@
+describe("Array.prototype.sort", () => {
```
linusg, line 1: Please format this whole file with [prettier](https://github.com/prettier/prettier/) (notice the `.prettierrc` config file in the repository root)

XavierCooney:
Done fixing up the tests (and I added another one). The formatter makes the sparse arrays much easier to read :).

PR #4564, opened 2020-12-27
AK: Use direct-list-initialization for Vector::empend()

lanza:
clang trunk with -std=c++20 doesn't seem to properly look for an
aggregate initializer here when the type being constructed is a simple
aggregate (e.g. `struct Thing { int a; int b; };`). This template fails
to compile in a usage added 12/16/2020 in `AK/Trie.h`.

Both forms of initialization are supposed to call the
aggregate-initializers but direct-list-initialization delegating to
aggregate initializers is a new addition in c++20 that might not be
implemented yet.

The other changes are type narrowing issues brought to light by
switching to brace init. All these usages are templates which resolve
their templated arguments to `int` given integer literal inputs and
not the proper types for the type they are constructing.

### DIFF
--- a/AK/Vector.h
+++ b/AK/Vector.h
@@ -398,7 +398,7 @@ public:
     void empend(Args&&... args)
     {
         grow_capacity(m_size + 1);
-        new (slot(m_size)) T(forward<Args>(args)...);
+        new (slot(m_size)) T { forward<Args>(args)... };
         ++m_size;
     }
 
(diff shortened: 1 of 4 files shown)


lanza:
Note -- there are still other build blockers on Darwin atm and I haven't been able to finish the build nor run it.

alimpfard:
> be me
> use c++20 feature for extra comfort
> macos-build-breaks.jpg

lanza:
This has nothing to do with macOS, this is a clang issue.

alimpfard:
Which almost always shows up as the macOS build breaking, as clang is the default there :P

PR #4565, opened 2020-12-27
Meta: Remove catchall lines in .gitignore

lanza:
These two lines caused this local .gitignore to supersede all inputs
in any files specified by a user's `core.excludesFile` configuration
since the first match was going to be `!*.*` for any file with a
decimal or any directory. `git check-ignore -v somefile` can be used to
test this.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,4 @@
 *
-!*.*
-!*/
 !Makefile
 !LICENSE
 !Base/**


awesomekling:
All righty

awesomekling:
I reverted this in 45d1c083867ea73a62b030e2d39636f6c89464b2 since it caused us to gitignore *everything*
cc @linusg

PR #4570, opened 2020-12-27
Enforce prettier formatting for JavaScript files

linusg:
My bash is terrible, suggestions for improvement welcome. pre-commit now requires a working install of prettier.

---

Some background:
We've been using prettier for JavaScript files (mostly LibJS test files) [for a while now](https://freenode.logbot.info/serenityos/20200526#c3977901-c3977933), but it's never been automatically enforced like clang-format is, so it depended on a human reviewer catching inconsistencies - not anymore.


linusg:
draft until I figure this out :smile:

Issue #4569, opened 2020-12-27
libm: build is broken on macOS


diversys:
This is 815d39886f4f0e835bb6300682bafc7dc266ac2b with this patch applied https://github.com/SerenityOS/serenity/issues/4483#issuecomment-751423966

```
git pull && ninja install && ninja image && ninja run
Already up to date.
[0/2] Re-checking globbed directories...
[154/2338] Linking CXX static library Meta/Lagom/libLagomCore.a
/Library/Developer/CommandLineTools/usr/bin/ranlib: file: Meta/Lagom/libLagomCore.a(SharedBuffer.cpp.o) has no symbols
/Library/Developer/CommandLineTools/usr/bin/ranlib: file: Meta/Lagom/libLagomCore.a(Command.cpp.o) has no symbols
/Library/Developer/CommandLineTools/usr/bin/ranlib: file: Meta/Lagom/libLagomCore.a(DirectoryWatcher.cpp.o) has no symbols
/Library/Developer/CommandLineTools/usr/bin/ranlib: file: Meta/Lagom/libLagomCore.a(SharedBuffer.cpp.o) has no symbols
/Library/Developer/CommandLineTools/usr/bin/ranlib: file: Meta/Lagom/libLagomCore.a(Command.cpp.o) has no symbols
/Library/Developer/CommandLineTools/usr/bin/ranlib: file: Meta/Lagom/libLagomCore.a(DirectoryWatcher.cpp.o) has no symbols
[995/2338] Linking CXX shared library Libraries/LibM/libm.so
FAILED: Libraries/LibM/libm.so
: && /Users/diver/serenity/Toolchain/Local/bin/i686-pc-serenity-g++ -pie -fpic -nostdlib  -shared --sysroot /Users/diver/serenity/Build/Root -o Libraries/LibM/libm.so -install_name @rpath/libm.so Libraries/LibM/CMakeFiles/LibM.dir/math.cpp.o   && :
i686-pc-serenity-g++: error: rpath/libm.so: No such file or directory
i686-pc-serenity-g++: error: unrecognized command-line option '-install_name'
[1004/2338] Building CXX object Libraries/LibLine/CMakeFiles/LibLine.dir/Editor.cpp.o
ninja: build stopped: subcommand failed.
```

ADKaster:
Try one of these, docs suggest they will help:

```
set(CMAKE_INSTALL_NAME_DIR "")
set(CMAKE_MACOSX_RPATH FALSE)
```

[...]

Issue #4541, opened 2020-12-25
Broken build macOS 11.1 Apple M1 chip


eizyaev:
Hello,
It's my first time trying to build SerenityOS (So probably I am missing something obvious here), I am getting the following error:

Did anyone stumbled upon this by any chance and know how to solve it?

$ cd ..
$ cd Build
$ cmake .. -DCMAKE_C_COMPILER=gcc-10 -DCMAKE_CXX_COMPILER=g++-10
$ make

...
Scanning dependencies of target TestBinarySearch
[  6%] Building CXX object AK/Tests/CMakeFiles/TestBinarySearch.dir/TestBinarySearch.cpp.o
[  6%] Linking CXX executable TestBinarySearch
[  6%] Built target TestBinarySearch
Scanning dependencies of target Regex
[  6%] Building CXX object Libraries/LibRegex/Tests/CMakeFiles/Regex.dir/Regex.cpp.o
[  6%] Building CXX object Libraries/LibRegex/Tests/CMakeFiles/Regex.dir/__/C/Regex.cpp.o
[  6%] Building CXX object Libraries/LibRegex/Tests/CMakeFiles/Regex.dir/__/RegexByteCode.cpp.o
[  6%] Building CXX object Libraries/LibRegex/Tests/CMakeFiles/Regex.dir/__/RegexLexer.cpp.o
[  6%] Building CXX object Libraries/LibRegex/Tests/CMakeFiles/Regex.dir/__/RegexMatcher.cpp.o
[  6%] Building CXX object Libraries/LibRegex/Tests/CMakeFiles/Regex.dir/__/RegexParser.cpp.o
[  6%] Linking CXX executable Regex
[  6%] Built target Regex
Scanning dependencies of target RegexLibC
[  6%] Building CXX object Libraries/LibRegex/Tests/CMakeFiles/RegexLibC.dir/RegexLibC.cpp.o
[  6%] Building CXX object Libraries/LibRegex/Tests/CMakeFiles/RegexLibC.dir/__/C/Regex.cpp.o
[  7%] Building CXX object Libraries/LibRegex/Tests/CMakeFiles/RegexLibC.dir/__/RegexByteCode.cpp.o
[  7%] Building CXX object Libraries/LibRegex/Tests/CMakeFiles/RegexLibC.dir/__/RegexLexer.cpp.o
[...]

Issue #4483, opened 2020-12-21
#4320 broke macOS build


diversys:
#4320 broke build on macOS. 

```
[244/1268] Building CXX object Meta/Lagom/CMakeFiles/LagomCore.dir/__/__/Libraries/LibCore/ElapsedTimer.cpp.o
FAILED: Meta/Lagom/CMakeFiles/LagomCore.dir/__/__/Libraries/LibCore/ElapsedTimer.cpp.o
/usr/local/bin/g++-10  -I../Libraries -I../. -I../Meta/Lagom/../.. -I../Meta/Lagom/../../Libraries -Wno-unknown-warning-option -Wall -Wextra -Werror -Wmissing-declarations -std=c++2a -fdiagnostics-color=always -ftls-model=initial-exec -fconcepts -Wno-unknown-warning-option -O2 -Wall -Wextra -Werror -std=c++2a -fPIC -g -Wno-deprecated-copy -Wno-expansion-to-defined -isysroot /Library/Developer/CommandLineTools/SDKs/MacOSX10.14.sdk -MD -MT Meta/Lagom/CMakeFiles/LagomCore.dir/__/__/Libraries/LibCore/ElapsedTimer.cpp.o -MF Meta/Lagom/CMakeFiles/LagomCore.dir/__/__/Libraries/LibCore/ElapsedTimer.cpp.o.d -o Meta/Lagom/CMakeFiles/LagomCore.dir/__/__/Libraries/LibCore/ElapsedTimer.cpp.o -c ../Libraries/LibCore/ElapsedTimer.cpp
../Libraries/LibCore/ElapsedTimer.cpp: In member function 'void Core::ElapsedTimer::start()':
../Libraries/LibCore/ElapsedTimer.cpp:39:49: error: 'CLOCK_MONOTONIC_COARSE' was not declared in this scope; did you mean 'CLOCK_MONOTONIC_RAW'?
   39 |     clock_gettime(m_precise ? CLOCK_MONOTONIC : CLOCK_MONOTONIC_COARSE, &now_spec);
      |                                                 ^~~~~~~~~~~~~~~~~~~~~~
      |                                                 CLOCK_MONOTONIC_RAW
../Libraries/LibCore/ElapsedTimer.cpp: In member function 'int Core::ElapsedTimer::elapsed() const':
[...]

PR #4571, opened 2020-12-27
Meta: Add macOS workflow to CI

Lubrsi:
A good number of contributors use macOS. However, we have a bit of
a tendency of breaking the macOS build without realising it.

Luckily, GitHub Actions does actually supply macOS environments,
so let's use it.

Note that this doesn't currently have any fixes to the macOS build.
I'm not sure what to do with them.

Another thing to note is that this uses Big Sur for the build. They
also provide Catalina, but I just went for the latest. I'm not sure
if this would have any effect.

Plus, they don't seem to provide M1 variants, so this is sadly not
completely relevant to #4541.

Relevant issues: #4483 #4569

Marked as draft because I forgot to mark it as draft.

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -136,3 +136,48 @@ jobs:
     - name: Build Lagom with Fuzzers
       working-directory: ${{ github.workspace }}/Meta/Lagom/Build
       run: cmake --build .
+  build_and_test_on_macos:
+    runs-on: macos-11.0
+
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Install dependencies
+      run: brew install coreutils ninja
+    - name: Check versions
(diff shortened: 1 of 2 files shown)


.github/workflows/cmake.yml
```diff
@@ -136,3 +136,51 @@ jobs:
...
       run: cmake --build .
+  build_and_test_on_macos:
+    runs-on: macos-11.0
+
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Install dependencies
+      # These packages are already part of the ubuntu-20.04 image:
+      # clang-format-10 cmake gcc-10 g++-10 shellcheck libgmp-dev
```
alimpfard, line 147: But are they part of the macos-11.0 image?

Lubrsi: Whoops :sweat_smile:, good ol' copy and paste.

Lubrsi:
Taken in by #4582

PR #4574, opened 2020-12-27
LibGUI: Add GUI::AppFile, a simple wrapper around .af files

linusg:
This is an attempt to make the common task of opening and reading an `.af` file easier and more fail-proof.


awesomekling:
Great initiative! Perhaps this would fit in LibDesktop but I suppose LibGUI is fine. :)

PR #4575, opened 2020-12-27
DevTools: new project C++compiler

denis-campredon:
I created in DevTools a new project, a C++ compiler.

For now, it can only compile empty functions and functions that return a series of addition, subtraction or multiplication. The values in the expression must come from the function parameter. It only works with integer parameters. More will be added latter.

I've tried to mimic what was output by g++ and mostly clang++ via [godbolt](https://godbolt.org/z/8YTG38) with some minor modifications when needed

Of course, I'll try to add more and more stuff to it (after the holidays), so it'll be able to parse real programs, ... but for now I'd like to have some feedback, and maybe some help.

The program is split into 4 parts:
- **LibMiddleEnd**: 
  - SIR: Serenity Intermediate Representation. A tree structure that should look like a SSA (Static single assignment form) that should be language and platform agnostic, ...

- **LibBackEnd**: A lib to transform the middle end representation into assembly. (AT&T assembly for i386). It is really naive and only use `eax` and the function parameters from the stack. (Lot of useless `mov`, but the assembly looks like llvm one at -O0)

- **LibCpp**
  - The lexer has been a little bit modified, it now recognizes a lot less keywords and types.
  - A parser has been created. I've chosen to follow the names from the [ISO c++20 paper](https://isocpp.org/files/papers/N4860.pdf), so it'll be easier to add stuff to the parser 
  - some options, just to indicate what file to parse, ...
  - An AST, that extends the one from the LibMiddleEnd, to add more information about the constructs.
  - And something that flattens the tree `a +b + c;` into `-.1 = a + b; _.2 = _.1 + c`

- **test-c++-compiler**: a tester for the program. Basically, it lexes the file to get all special single or multiline `//$` or `/*$ */` comments, and then follow the instructions inside it. I've chosen to replace tabulations with `[[:whitespace:]]` because IDE do not like when tabulations are pasted, and the result can lead to confusion.

**Note**: This broke a little bit the syntax highlighting of Hackstudio, but it will go back when new keywords or types will be added to the compiler.


Base/res/apps/Playground.af
```diff
@@ -0,0 +1,4 @@
+[App]
```
denis-campredon, line 1: I know I added some older commits, but I don't really know/understand what happened And my git is not that great to fix that, so  maybe so help would be appreciated. :)

alimpfard: To fix that, you probably want to rebase on top of upstream master, `git rebase master` if your master branch is up-to-date.

AK/ScopeLogger.h
```diff
@@ -0,0 +1,61 @@
...
+        dbgln("\033[1;{}m{}leaving {}\033[0m", (m_depth + 1) % 8 + 30, sb.to_string(), m_fun);
+    }
+
+private:
+    static inline size_t m_depth = 0;
+    StringView m_fun;
+};
+}
+
+#    define SCOPE_LOGGER() auto tmp##__COUNTER__ = ScopeLogger(__PRETTY_FUNCTION__);
```
alimpfard: - Name in commit doesn't match actual name :) 
- Probably better to just `using AK::ScopeLogger;` right above this, that's what most of AK does too

DevTools/Compiler/C++Compiler/LibCpp/Driver.cpp
```diff
@@ -0,0 +1,46 @@
...
+
+namespace Cpp {
+void Cpp::CppCompiler::run(int ac, const char** av)
+{
+    auto options = Cpp::Option::parse_options(ac, av);
+    auto tu = Cpp::Parser::parse(options);
+    auto ir = Cpp::IR::to_internal_representation(tu);
+
+    SIR::process_internal_representation(ir);
+    auto backEnd = BackEnd::I386BackEnd(ir, options);
```
alimpfard, line 42: pls no camelCase in C++ code?

DevTools/Compiler/C++Compiler/LibCpp/Driver.cpp
```diff
@@ -0,0 +1,46 @@
...
+
+#include "Driver.h"
+#include "IR.h"
+#include "Option.h"
+#include "Parser.h"
+#include <LibBackEnd/I386BackEnd.h>
+#include <LibMiddleEnd/SIR.h>
+
+namespace Cpp {
+void Cpp::CppCompiler::run(int ac, const char** av)
```
alimpfard, line 35: This feels like a weird signature, maybe take the options instead?

DevTools/Compiler/C++Compiler/LibCpp/Option.cpp
```diff
@@ -0,0 +1,45 @@
...
+Cpp::Option Cpp::Option::parse_options(int ac, const char** av)
+{
+    assert(ac == 2 || ac == 4);
+
+    Cpp::Option options = {};
+
+    options.input_file = av[1];
+    if (ac == 2) {
+        options.output_file = options.input_file.isolated_copy();
+        assert(options.output_file.replace(".cpp", ".s") == 1);
```
alimpfard, line 39: I'm not sure what's going on with this `replace()` method, but it presumably modifies something?
Please don't put that sort of thing inside `assert()` - that will not be run when assertions are not enabled.
also, I think it's a little bit more idiomatic to use `ASSERT()`, but meh.

DevTools/Compiler/C++Compiler/LibCpp/Option.cpp
```diff
@@ -0,0 +1,45 @@
...
+
+#include "Option.h"
+#include <string.h>
+
+Cpp::Option Cpp::Option::parse_options(int ac, const char** av)
+{
+    assert(ac == 2 || ac == 4);
+
+    Cpp::Option options = {};
+
```
alimpfard, line 35: maybe, _just maybe_, you can make do with `Core::ArgsParser`?

DevTools/Compiler/C++Compiler/main.cpp
```diff
@@ -0,0 +1,34 @@
...
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include "LibCpp/Driver.h"
+
+int main(int ac, const char** av)
+{
+    Cpp::CppCompiler::run(ac, av);
```
alimpfard, line 31: Why not put that logic here? no one else will want to embed a C++ compiler with this interface - they'd fork/exec this thing otherwise?

DevTools/Compiler/LibBackEnd/I386BackEnd.cpp
```diff
@@ -0,0 +1,55 @@
...
+    auto input_file_name = m_options.input_file;
+    while (input_file_name.contains("/")) {
+        auto new_path = input_file_name.index_of("/");
+
+        assert(new_path.has_value());
+        input_file_name = input_file_name.substring(new_path.value() + 1);
+    }
+    m_output_file->printf("\t.file \"%s\"\n", input_file_name.characters());
+    m_output_file->printf("\t.ident \"Serenity-c++ compiler V0.0.0\"\n");
+    m_output_file->printf("\t.section \".note.GNU-stack\",\"\",@progbits\n");
```
alimpfard, line 110: Maybe a nitpick, but could you perhaps use `AK::SourceGenerator`?
it tends to make much more readable strings

DevTools/Compiler/C++Compiler/LibCpp/AST.cpp
```diff
@@ -0,0 +1,56 @@
...
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include "AST.h"
+#include <AK/StringBuilder.h>
+
+namespace Cpp {
+
+template<typename A, typename... Args>
+NonnullRefPtr<A> create(Args... args)
```
alimpfard, line 33: This guy is showing up in more places, maybe we should put it in AK?
Not a requirement for this PR, just a thought

DevTools/Compiler/C++Compiler/LibCpp/Parser.cpp
```diff
@@ -284,8 +502,8 @@ NonnullRefPtr<Function> Parser::parse_function_definition()
     auto declarator = parse_declarator();
 
     if (declarator.has_value()) {
-        parse_function_body();
-        return create_ast_node<Function>(return_type, declarator.value().name, declarator.value().parameters);
+        auto body = parse_function_body();
+        return create_ast_node<Function>(return_type, declarator.value().name, declarator.value().parameters, body);
```
alimpfard, line 533: maybe `move(body)`? not that important, just a nitpick

DevTools/Compiler/LibMiddleEnd/Utils.cpp
```diff
@@ -32,4 +32,14 @@ NonnullRefPtr<SIR::Variable> create_store(NonnullRefPtr<SIR::Type>& type, String
...
 }
 
+NonnullRefPtr<SIR::BinaryExpression> create_binary_operation(NonnullRefPtr<SIR::Variable>& left, NonnullRefPtr<SIR::Variable>& right, SIR::BinaryExpression::Kind operation)
+{
+    assert(left->node_type()->kind() == right->node_type()->kind());
+    assert(left->node_type()->size_in_bytes() == right->node_type()->size_in_bytes());
+    assert(left->node_type()->size_in_bits() == right->node_type()->size_in_bits());
+    auto result = RefPtr(create_store(left->node_type(), left->name()));
+
+    return create_ast_node<SIR::BinaryExpression>(operation, reinterpret_cast<NonnullRefPtr<SIR::ASTNode>&>(left), reinterpret_cast<NonnullRefPtr<SIR::ASTNode>&>(right), reinterpret_cast<RefPtr<SIR::Variable>&>(result));
```
alimpfard: what's going on over here? if I understand it right, `*left` should be fine, no need to `reinterpret_cast`?

DevTools/Compiler/C++Compiler/LibCpp/Parser.cpp
```diff
@@ -88,6 +87,35 @@ Token Parser::consume(Token::Type expected_type)
...
+
+        auto any_operator_match = [](auto const& expected, auto const&... rest) {
+            return ((expected == rest) || ...);
+        }(bi_right.binary_operation(), args...);
+
+        if (any_operator_match) {
+            auto right_left_node = bi_right.left();
+            result.set_right(right_left_node);
+            bi_right.set_left(result);
+            return bi_right;
```
alimpfard, line 113: why is this happening anyway? the parser shouldn't create the wrong AST to begin with, no?
if there's an actual need for this, could you at least put a comment about what's being fixed here?

alimpfard (commented):
Very cool to see a start on this!
I hope you're ready for a metric tonne of review comments :) 

Sidenote, please make the commit messages a bit more tidy?

linusg (commented):
The directory structure is strange, IMO:

- `DevTools/Compiler/C++Compiler/LibCpp`
- `DevTools/Compiler/LibMiddleEnd/`
- `DevTools/Compiler/LibBackEnd/`

---

- ~Why is `LibCpp` in a separate sub-directory but `LibMiddleEnd` and `LibBackEnd` are not?~ Edit: because they're supposed to be language agnostic, hmm...
- Why is LibCpp not in `Libraries` (anymore), where it belongs, IMO?
- Do we even need to separate LibMiddleEnd and LibBackEnd into different libraries?
- Since the code calls it `CppCompiler` maybe we can use that instead of `C++Compiler`?

---

I'd suggest something like this:

- `DevTools/CppCompiler/main.cpp`
- `Libraries/LibCpp/...`
- `Libraries/LibCpp/MiddleEnd/...`
- `Libraries/LibCpp/BackEnd/...`

That would also simplify the CMake files, which currently contain things like:

```
include_directories(../)
include_directories(../C++Compiler)
```

DevTools/Compiler/C++Compiler/main.cpp
denis-campredon: I put all the logic in CppCompiler (that I forgot to rename into Driver btw), so it could be called from the testsuite.
I don't know what other interface I should expose though. One with the options as parameter ?

Lubrsi:
Please configure your editor to add new lines to the end of the file when you save :)

DevTools/Compiler/LibMiddleEnd/Utils.cpp
denis-campredon: It's a problem with all LL parsers. When parsing two operator with the same precedence, it will produce a tree with the branches inverted.
`i + j - k` will have a tree like `i + (j -k)`. This method transorm that tree into `(i + j) -k`.

DevTools/Compiler/C++Compiler/main.cpp
alimpfard: Whichever works I suppose, but I feel like having an interface that's basically equivalent to `execl("the_compiler", ...)` isn't very flexible.
Though I suppose there's no particular need to embed the driver in something, so it's probably a non-issue

DevTools/Compiler/C++Compiler/LibCpp/Parser.cpp
denis-campredon: It's a problem with all LL parsers. When parsing two operator with the same precedence, it will produce a tree with the branches inverted.
`i + j - k` will have a tree like `i + (j -k)`. This method transorm that tree into `(i + j) -k`.

I'll add a comment explaining that.

alimpfard: Oh this is about + and -, I see.
For some reason I thought it was mixing up multiply and minus 🤔 
My bad~

denis-campredon:
@linusg Yeah maybe I should do something like that, and if someone wants to add a new language or whatever, we'll see how to make the structure better.

AK/ScopeLogger.h
denis-campredon: In my opinion this isn't supposed to be used without the macro, but I can still add the `using` directive.

asynts:
I think `LibMiddleEnd` is confusing and `LibBackEnd` is not very descriptive either. How about `LibIntermediate` and `LibAssembler` instead?

(Or `Libraries/LibCpp/Assembler` and `Libraries/LibCpp/Intermediate` with the new directory schema.)

Base/res/apps/Playground.af
denis-campredon: Should be fixed now. Thanks. :)

AK/ScopeLogger.h
denis-campredon: I added the using directive, and changed the macro to DEBUG_SPAM, since it was already existing.

DevTools/Compiler/C++Compiler/LibCpp/Option.cpp
denis-campredon: The assert have been removed. Maybe I should add it but in another line.

DevTools/Compiler/C++Compiler/LibCpp/Parser.cpp
denis-campredon: The compiler says I can't
`cannot bind non-const lvalue reference of type 'AK::NonnullRefPtrVector<SIR::ASTNode>&' to an rvalue of type 'AK::NonnullRefPtrVector<SIR::ASTNode>`
I haven't used a lot of move on the parser

DevTools/Compiler/LibMiddleEnd/Utils.cpp
denis-campredon: The function was not used anymore, I've removed it.

DevTools/Compiler/C++Compiler/LibCpp/Parser.cpp
denis-campredon: Comment has been added.

DevTools/Compiler/C++Compiler/LibCpp/Driver.cpp
denis-campredon: I've renamed that to `Driver::run`. I keep that signature so it can be called from the test driver.

DevTools/Compiler/LibBackEnd/I386BackEnd.cpp
denis-campredon: I don't understand what is the difference between that and `AK::StringBuilder`. Not modified it yet.

DevTools/Compiler/C++Compiler/LibCpp/Parser.cpp
alimpfard: Your constructor takes a `NonnullRefPtrVector<SIR::ASTNode>&` then copies it iirc, you should instead just take a `NonnullRefPtrVector<SIR::ASTNode>` and move the body from here.

DevTools/Compiler/C++Compiler/LibCpp/Option.cpp
alimpfard: There's nothing wrong with asserting, in fact, do more of that when you're assuming things!
just don't put state-changing code inside it is all I'm saying

denis-campredon:
I've recreated the structure suggested by @asynts and @linusg  (and added newlines in the missing files.)
I've also modified the commit message they should be more clear.

But now I have a problem, when the destructors from global objects are ran. (It did not happened before). Maybe I'll change the `HashMap` to a `HashMap*` to ignore the problem for now.

`
USERSPACE(27) ASSERTION FAILED: node == m_head   .././AK/InlineLinkedList.h:266

0x02d8d45a: [libc.so] kill +28 (Syscall.h:465)
0x02d8d4d6: [libc.so] raise +32 (signal.cpp:54)
0x02d85749: [libc.so] abort +25 (stdlib.cpp:230)
0x02d85fe7: [libc.so] tolower +0 (assert.cpp:42)
0x02d8901f: [libc.so] free +459 (malloc.cpp:312)
0x02e3a26a: [libcpp.so] AK::HashMap<AK::StringView, Cpp::Token::KnownKeyword, AK::Traits<AK::StringView> >::~HashMap() +34 (HashTable.h:104)
0x02d860a9: [libc.so] __cxa_finalize +72 (cxxabi.cpp:80)
0x02d9b5b8: [libc.so] exit +25 (stdlib.cpp:205)
0x029a2e08: [/bin/test-c++-compiler] __x86.get_pc_thunk.bx +0 (crt0.cpp:59)
`

DevTools/Compiler/LibBackEnd/I386BackEnd.cpp
alimpfard: IIRC, `SourceGenerator`s can store values for names, and they can be stacked on top of each other.
e.g. you'd go
```c++
auto generator = get_some_base_generator('{', '}');
generator.set("input.filename", input_file_name);
generator.append(R"~(
    .file "{input.filename}"
    .ident "Serenity-c++ compiler v0.0.0"
    .section ".note.GNU-stack","",@progbits
)~");
```

asynts:
Judging from that error message you are creating a `HashMap<StringView, X>`. While this could be correct in some contexts, you probably meant `HashMap<String, X>`?

denis-campredon:
The values are always `const char*`, so a `StringView` is fine there. The same problem arise when I'm using a `String` instead.

asynts:
The destructor is called twice for some reason, that's not your fault though. I've created an issue for this #4605.

If you wrap it into a function it's fine.

DevTools/Compiler/LibBackEnd/I386BackEnd.cpp
denis-campredon: Done, I did't use raw strings, because the IDE (CLion, was replacing the tabs with spaces.)

awesomekling:
Very interesting project!

Dexesttp:
You might have ended up with a slightly messed up commit history somehow ^^'

You can just run this (assuming that this remote repo is called "origin" for you and that your remote repo is called "denis") :
```shell
git fetch origin
git rebase --onto origin/master c5d857d4a2c6610f89ed1a2318b0345244231143 HEAD
```
And then resolve all the "false" merge conflicts you'll get (`git add .` and `git rebase --continue`).

I think everything should be there after the rebase ends - it seemed fine on my end at least. Please, check if it compiles & works properly before going forward, and if you're not missing any files. If it doesn't work properly / if you're missing something, you can `git checkout --force c++compiler` and try again, checking each step of the rebase instead of just blindly moving forward.

Then, once you're sure everything is "cleaned up" and works, you can set your branch to the clean state with `git branch -f c++compiler` and `git push -u --force denis c++compiler`

PR #4576, opened 2020-12-27
LibELF: Reject ELF with program header p_filesz larger than p_memsz

bcoles:
The `PT_LOAD` load section of the [Linux man page for elf](https://man7.org/linux/man-pages/man5/elf.5.html) states:

> The file size may not be larger than the memory size.

It seems to imply that this is only relevant for program headers with type `PT_LOAD`. ~~Apparently coredumps have a hardcoded memsz of zero (https://github.com/SerenityOS/serenity/pull/4576#issuecomment-751510304), so we can't apply this check to all program headers, even though that would probably make sense.~~ There's no immediately obvious reason why we can't apply this to all program headers. However, the tradeoff is that we won't be able to parse malformed ELF files, or coredumps files from other systems, which have `p_filesz` larger than `p_memsz`.

After this patch, the following program will now be rejected:

```cpp
#include <LibELF/exec_elf.h>
#include <fcntl.h>
#include <stdio.h>
#include <string.h>
#include <unistd.h>

#pragma GCC diagnostic ignored "-Wdeprecated-declarations"

int main()
{
    char buffer[0x2000];

    auto& header = *(Elf32_Ehdr*)buffer;
    header.e_ident[EI_MAG0] = ELFMAG0;
    header.e_ident[EI_MAG1] = ELFMAG1;
    header.e_ident[EI_MAG2] = ELFMAG2;
    header.e_ident[EI_MAG3] = ELFMAG3;
    header.e_ident[EI_CLASS] = ELFCLASS32;
    header.e_ident[EI_DATA] = ELFDATA2LSB;
    header.e_ident[EI_VERSION] = EV_CURRENT;
    header.e_ident[EI_OSABI] = ELFOSABI_SYSV;
    header.e_ident[EI_ABIVERSION] = 0;
    header.e_type = ET_EXEC;
    header.e_version = EV_CURRENT;
    header.e_ehsize = sizeof(Elf32_Ehdr);
    header.e_machine = EM_386;
    header.e_shentsize = sizeof(Elf32_Shdr);

    header.e_phnum = 1;
    header.e_phoff = 52;
    header.e_phentsize = sizeof(Elf32_Phdr);

    auto* ph = (Elf32_Phdr*)(&buffer[header.e_phoff]);
    ph[0].p_flags = PF_R | PF_X;
    ph[0].p_align = PAGE_SIZE;
    ph[0].p_vaddr = 0x000000d4;
    ph[0].p_type = PT_LOAD;
    ph[0].p_memsz = 0;  // crash
    ph[0].p_offset = 0x1000;
    ph[0].p_filesz = 0x1000;

    /* inaccurate and unrelated */
    header.e_shnum = 3;
    header.e_shoff = 1024;
    header.e_shstrndx = 2;
    header.e_entry = 1024;

    int fd = open("x", O_RDWR | O_CREAT, 0777);
    if (fd < 0) {
        perror("open");
        return 1;
    }

    int nwritten = write(fd, buffer, sizeof(buffer));
    if (nwritten < 0) {
        perror("write");
        return 1;
    }

    if (execl("/home/anon/x", "x", nullptr) < 0) {
        perror("execl");
        return 1;
    }

    return 0;
}
```

```
[elf-crash(27:27)]: Program header (0) PT_LOAD has p_filesz (4096) larger than p_memsz (0)
[elf-crash(27:27)]: exec(/home/anon/x): File has invalid ELF Program headers
elf-crash(27): perror(): execl: Exec format error
Shell(24): Job entry "elf-crash" deleted in 148 ms
```

### DIFF
--- a/Libraries/LibELF/Validation.cpp
+++ b/Libraries/LibELF/Validation.cpp
@@ -193,6 +193,13 @@ bool validate_program_headers(const Elf32_Ehdr& elf_header, size_t file_size, co
 
     for (size_t header_index = 0; header_index < num_program_headers; ++header_index) {
         auto& program_header = program_header_begin[header_index];
+
+        if (program_header.p_filesz > program_header.p_memsz) {
+            if (verbose)
+                dbgln("Program header ({}) has p_filesz ({}) larger than p_memsz ({})", header_index, program_header.p_filesz, program_header.p_memsz);
+            return false;
+        }
+
         switch (program_header.p_type) {
         case PT_INTERP:
             // We checked above that file_size was >= buffer size. We only care about buffer size anyway, we're trying to read this!


Libraries/LibELF/Validation.cpp
```diff
@@ -193,6 +193,13 @@ bool validate_program_headers(const Elf32_Ehdr& elf_header, size_t file_size, co
 
     for (size_t header_index = 0; header_index < num_program_headers; ++header_index) {
         auto& program_header = program_header_begin[header_index];
+
+        if (program_header.p_filesz > program_header.p_memsz) {
+            if (verbose)
+                dbgprintf("Program header (%zu) has p_filesz (%u) larger than p_memsz (%u)\n", header_index, program_header.p_filesz, program_header.p_memsz);
```
linusg, line 199: The latest and greatest would be:

```suggestion
                dbgln("Program header ({}) has p_filesz ({}) larger than p_memsz ({})", header_index, program_header.p_filesz, program_header.p_memsz);
```

bcoles:
Apparently the core dumps we generate have a large number of program headers, and some of these may have a file size larger than memory size. As a result, when core dumps are dropped in `/tmp/coredumps`, the `CrashDaemon` crashes (not that there's anything unusual about that).

```
CrashDaemon(19): New coredump file: /tmp/coredump/crash_27_1609099470
CrashDaemon(19): Program header (35) has p_filesz (1203) larger than p_memsz (0)
Image::parse(): ELF Program Headers not valid
USERSPACE(19) ASSERTION FAILED: m_valid
../Libraries/LibELF/Image.cpp:147
[CrashDaemon(19:19)]: Terminating CrashDaemon(19) due to signal 6
Shell(24): Job entry "crash -d" deleted in 919 ms
[FinalizerTask(3:3)]: 0xdeadc0de  Kernel::Processor::switch_context(Kernel::Thread*&, Kernel::Thread*&) +181
0xdeadc0de  Kernel::Scheduler::context_switch(Kernel::Thread*) +155
0xdeadc0de  Kernel::Scheduler::pick_next() +956
0xdeadc0de  Kernel::Processor::check_invoke_scheduler() +125
0xdeadc0de  Kernel::Thread::die_if_needed() +163
0xdeadc0de  syscall_handler +2195
0xdeadc0de  syscall_asm_entry +49
0x0248d512
0x02485759
0x02486023
0x024b648e
0x02041dd8
0x020420c7
0x02046d16
0x020478ae
0x1131802c

[FinalizerTask(3:3)]: Generating coredump for pid: 19
[FinalizerTask(3:3)]: RTC: Year: 2020, month: 12, day: 27, hour: 20, minute: 4, second: 31
SystemServer(4): Service CrashDaemon has exited with exit code 6
```

Libraries/LibELF/Validation.cpp
```diff
@@ -205,6 +205,12 @@ bool validate_program_headers(const Elf32_Ehdr& elf_header, size_t file_size, co
                 *interpreter_path = String((const char*)&buffer[program_header.p_offset], program_header.p_filesz - 1);
             break;
         case PT_LOAD:
+            if (program_header.p_filesz > program_header.p_memsz) {
+                if (verbose)
+                    dbgln("Program header ({}) PT_LOAD has p_filesz ({}) larger than p_memsz ({})", header_index, program_header.p_filesz, program_header.p_memsz);
+                return false;
+            }
+            break;
```
ADKaster, line 213: This change makes PT_LOAD headers skip the validation of p_offset and p_flags that used to be common to LOAD, DYNAMIC, NOTE, PHDR, and TLS. Should either muse `[[ fallthrough ]]` or consider if the validation you're should apply to more than just PT_LOAD

bcoles: > This change makes PT_LOAD headers skip the validation of p_offset and p_flags that used to be common to LOAD

How does it cause validation to be skipped?

This check is in the `validate_program_headers()` function. If this check fails it will return false. Presumably any consuming function should bail out if the header validation fails.

> or consider if the validation you're adding should apply to more than just PT_LOAD

I tried apply this check to more than `PT_LOAD` (all program headers) but ran into an issue with `PT_NOTE` in coredumps, as coredumps have a hardcoded `memsz` of zero.

Also, applying this check only to `PT_LOAD` is inline with the Linux [implementation](https://elixir.bootlin.com/linux/latest/source/fs/binfmt_elf.c#L663) and [documentation](https://man7.org/linux/man-pages/man5/elf.5.html).

ADKaster: The break statement. Before, the checks on line 218 through 227 were called for PT_LOAD. But with the new break here on line 213, if a PT_LOAD succeeds your new check, the old p_offset and p_flags checks are ignored for that program header.

bcoles: ugh thanks. I didn't realize that was intentional.

kling has merged my PR to set `memsz` to a non-zero value for coredumps, so I'll revert my changes back to what I had originally. That is, performing this check for every program header.

PR #4587, opened 2020-12-27
TextEditor: Added GML preview checkbox under 'View' and GML Preview

nanoproductions:
Added the ability to view GML files in the TextEditor and work on them!

### DIFF
--- a/Applications/TextEditor/TextEditorWidget.cpp
+++ b/Applications/TextEditor/TextEditorWidget.cpp
@@ -407,9 +407,16 @@ TextEditorWidget::TextEditorWidget()
         },
         this);
 
+    m_gml_preview_action = GUI::Action::create_checkable(
+        "GML preview", [this](auto&) {
+            set_preview_mode(PreviewMode::GML);
+        },
+        this);
+
     m_preview_actions.add_action(*m_no_preview_action);
     m_preview_actions.add_action(*m_markdown_preview_action);
     m_preview_actions.add_action(*m_html_preview_action);
+    m_preview_actions.add_action(*m_gml_preview_action);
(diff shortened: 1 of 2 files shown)


danboid:
Sounds v.cool!

Applications/TextEditor/TextEditorWidget.cpp
```diff
@@ -651,6 +668,13 @@ void TextEditorWidget::update_html_preview()
     m_page_view->scroll_into_view(current_scroll_pos, true, true);
 }
 
+void TextEditorWidget::update_GML_preview()
+{
+    auto current_scroll_pos = m_page_view->visible_content_rect();
+    m_page_view->load_from_gml(m_editor->text());
```
awesomekling, line 674: `m_page_view` is a web view, we should not be inserting random children into it.
This should be done with an empty `GUI::Widget` at the same level as `m_page_view`, possibly with both in a `GUI::StackWidget`.

Applications/TextEditor/TextEditorWidget.h
```diff
@@ -61,6 +62,7 @@ class TextEditorWidget final : public GUI::Widget {
     void update_preview();
     void update_markdown_preview();
     void update_html_preview();
+    void update_GML_preview();
```
awesomekling, line 65: `update_gml_preview()`

awesomekling (changes requested):

awesomekling:
I don't believe this feature is useful, since we already have a dedicated program (GML Playground) for editing GML. This seems like it would only result in having two instances of the same functionality, and this one always being weaker than what's available in GML Playground.

PR #4593, opened 2020-12-28
LibVT+Terminal: Implement a simple Find window for searching within the terminal

IdanHo:
A picture of it in use:
image


Libraries/LibVT/Range.h
```diff
@@ -0,0 +1,87 @@
+/*
+ * Copyright (c) 2018-2020, Andreas Kling <kling@serenityos.org>
```
awesomekling, line 2: I didn't write this :)

Libraries/LibVT/Range.h
```diff
@@ -0,0 +1,87 @@
...
+
+private:
+    VT::Position normalized_start() const { return m_start < m_end ? m_start : m_end; }
+    VT::Position normalized_end() const { return m_start < m_end ? m_end : m_start; }
+
+    VT::Position m_start;
+    VT::Position m_end;
+};
+
+};
```
awesomekling, line 87: Please configure your editor to put a newline at the end of files.

Libraries/LibVT/Range.h
IdanHo: Ah i assumed all files in serenity had to have that disclamer, should i just remove the copyright statement or replace it with my name?

awesomekling: Two options:
1) Put your own name
2) Put "the SerenityOS developers"

And please use 2020 as the copyright year instead of 2018-2020 :)

PR #4602, opened 2020-12-28
FileManager: Implement the cut file operation

IdanHo:
This exploits the comment lines in the text/uri-list specification: https://tools.ietf.org/rfc/rfc2483.txt (page 11), which might be a bit hackish, but the specification suggests doing the same for other uses ("In applications where one URI has been mapped to a list of URIs, the first line of the text/uri-list response SHOULD be a comment giving the original URI."), so it should be fine.

### DIFF
--- a/Applications/FileManager/FileUtils.cpp
+++ b/Applications/FileManager/FileUtils.cpp
@@ -38,6 +38,35 @@
 
 namespace FileUtils {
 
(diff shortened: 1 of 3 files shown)


IdanHo:
looks like the cmake tests timed out randomly (they pass locally), so im gonna try force pushing with no changes to force a re-run of the tests

awesomekling:
That's a nifty trick :)

PR #4606, opened 2020-12-28
LibJS: Track source position down to exceptions / Spreadsheet: Some fancy cell editor features :P 

alimpfard:
An image is worth some words, so here's an image :P

shot-2020-12-28_23-39-28

---

Now one might ask, "why not work on the multitude of issues about the Spreadsheet?", and to that I say "NOT FUN"

### DIFF
--- a/Applications/Spreadsheet/CellSyntaxHighlighter.cpp
+++ b/Applications/Spreadsheet/CellSyntaxHighlighter.cpp
@@ -53,6 +53,19 @@ void CellSyntaxHighlighter::rehighlight(Gfx::Palette palette)
         false,
         false,
         nullptr);
+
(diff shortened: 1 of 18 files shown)


alimpfard:
\*sighs\*
Stupid `Shell-backgrounding`.

Libraries/LibGfx/Rect.h
```diff
@@ -99,6 +99,20 @@ class Rect {
         m_location.move_by(delta);
     }
 
+    Rect moved_by(T dx, T dy) const
```
awesomekling, line 102: This already exists; it's called `translated()` :)

alimpfard: Ah. weird naming conflict, but ok.

Issue #4607, opened 2020-12-28
Shell: Wildcard globbing incorrect match


bcoles:
![no](https://user-images.githubusercontent.com/434827/103240850-926d4c00-49a5-11eb-8218-8dc6cd198823.png)

Simple reproducer:

![globbing match fail](https://user-images.githubusercontent.com/434827/103241367-17a53080-49a7-11eb-9311-f1afbfab5902.png)

The output makes it look like everything is in one directory - no concept of subdirectories. However, this is an issue with `ls` rather than `Shell`.

bcoles:
Same issue for `?` matching:

![no](https://user-images.githubusercontent.com/434827/103242357-0578c180-49aa-11eb-8ed8-449606f77fbd.png)

alimpfard:
`?` is supposed to match exactly a single character though
Does `/dev/pts?` actually match something in a...sane shell?

bcoles:
> `?` is supposed to match exactly a single character though

Yes.

> Does `/dev/pts?` actually match something in a...sane shell?

No. But the matching behavior is still unexpected. It should not match at all.

```
user@ubuntu:~/Desktop/serenity/Build$ ls -la /dev/pt?
total 0
drwxr-xr-x  2 root root      0 Nov 14 02:20 .
drwxr-xr-x 19 root root   4120 Dec 15 07:14 ..
crw--w----  1 user tty  136, 0 Dec 28 13:08 0
crw--w----  1 user tty  136, 1 Dec 28 13:05 1
crw--w----  1 user tty  136, 2 Dec 28 13:09 2
crw--w----  1 user tty  136, 3 Dec 24 13:13 3
crw--w----  1 user tty  136, 4 Dec 24 13:13 4
c---------  1 root root   5, 2 Nov 14 02:20 ptmx
user@ubuntu:~/Desktop/serenity/Build$ ls -la /dev/pts?
ls: cannot access '/dev/pts?': No such file or directory
user@ubuntu:~/Desktop/serenity/Build$ 
```

alimpfard:
It does _not match_, it just returns an empty list
Perhaps I should revisit #4316 a bit and pull out the exception stuff 🤔

bcoles:
> It does _not match_, it just returns an empty list

Good point. That makes sense.

bcoles:
The issue with shell wilcard globbing has been resolved. The remaining issues existing with `ls`. I may or may not create a separate issue for `ls` later.

Closing this issue.

PR #4608, opened 2020-12-28
AK: Allow trailing '*'s in a glob pattern to match nothing

alimpfard:
Fixes part of #4607.


bcoles:
Confirmed this resolves the issue for me.

fixed

PR #4618, opened 2020-12-29
Build: Support non-i686 toolchains

meme:
- Add `SERENITY_ARCH` option to CMake for selecting the target toolchain
- Port all build scripts but continue to use i686

This removes the `BuildIt_x86_64.sh` script in favour of an architecture-independent build process.

In this model, multiple toolchains can be present, and each CMake build directory has its own architecture.

The standard build process is completely unchanged, but editors and IDEs will need their paths updated and the documentation has been updated to reflect that.

You can attempt to build as x86_64 with:
```
cd Toolchains
ARCH=x86_64 ./BuildIt.sh
cd .. && cd Build
cmake -DSERENITY_ARCH=x86_64 ..
make
```

---

### Breaking Proposal

Differentiate the `Build` folder outputted by `BuildIt.sh` per architecture. For example: `Build/i686` and `Build/x86_64`. This way you can switch directories easily and swap between development of the architectures.


meme:
Looks like CI is failing due to cache issues. The binaries being copied are not placed in the architecture-differentiated path. Is there anyway I can bust the cache or update the binaries in this PR?

awesomekling:
cc @BenWiederhake

BenWiederhake:
I forgot to include `Toolchain/BuildIt.sh` into the set of files that trigger a Toolchain rebuild.

This is best solved as part of this PR, since the diff is so tiny. In the main workflow file `.github/workflows/cmake.yml`, you'll find `name: Toolchain cache` *twice*. In both cases, change this:
```diff
-        key: ${{ runner.os }}-toolchain-${{ hashFiles('Libraries/LibC/**/*.h', 'Toolchain/Patches/*.patch') }}
+        key: ${{ runner.os }}-toolchain-i686-${{ hashFiles('Libraries/LibC/**/*.h', 'Toolchain/Patches/*.patch', 'Toolchain/BuildIt.sh') }}
```

This has two effects:
- Automatically invalidate the cache when `BuildIt.sh` changes.
- Give the future people who add new architectures a good idea how to change the `cmake.yml`, and prevent certain types of toolchain issues.

I didn't explicitly test this, but I don't see (yet) how this can fail.

awesomekling:
Cool! :)

PR #4620, opened 2020-12-29
Userland: Add readelf utility

bcoles:
readelf help

headers

program headers

sections

symbol table

There are a few things marked `TODO`. The `LibELF` API requires a bit of a rework first.

### DIFF
--- a/Userland/readelf.cpp
+++ b/Userland/readelf.cpp
@@ -1,0 +1,451 @@
+/*
+ * Copyright (c) 2020, the SerenityOS developers.
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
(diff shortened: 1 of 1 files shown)


awesomekling:
Should we maybe call this something other than `readelf` since that name clashes with the tool from GNU binutils?

bcoles:
> 
> 
> Should we maybe call this something other than `readelf` since that name clashes with the tool from GNU binutils?

What do you propose?

* `elfinfo`
* `elfread`
* `elfinspect`
* something else

Native applications are preferred over ports. `readelf` is intuitive for users with a *nix background.

awesomekling:
> Native applications are preferred over ports. `readelf` is intuitive for users with a *nix background.

Yeah you're probably right. I was mostly thinking about stuff like autoconf scripts failing because they expect readelf to behave a certain way. But then they should check for GNU readelf..

PR #4623, opened 2020-12-29
AK: Add HashMap(std::initializer_list<Entry>) constructor

linusg:
This allows us to construct a HashMap from an initializer list like so:

    HashMap<K, V> hash_map = { { K, V }, { K, V } { K, V } };

### DIFF
--- a/AK/HashMap.h
+++ b/AK/HashMap.h
@@ -31,6 +31,13 @@
 #include <AK/StdLibExtras.h>
 #include <AK/Vector.h>
 
+// NOTE: We can't include <initializer_list> during the toolchain bootstrap,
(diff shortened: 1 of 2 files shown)


AK/HashMap.h
```diff
@@ -49,7 +56,19 @@ class HashMap {
...
+        ensure_capacity(list.size());
+        for (auto& item : list)
+            set(item.key, item.value);
+    }
+#endif
+
+    bool is_empty() const
+    {
+        return m_table.is_empty();
+    }
```
linusg, line 71: I tried to keep this on one line, clang-format disagrees - no idea why.

PR #4625, opened 2020-12-29
Build: Expose symbols SECTION_start and SECTION_size for embedded resources

lux01:
Resources embedded by the embed_resource() function will now also expose a SECTION_start and SECTION_size symbol so the embedded resource can be found by an application without having to parse its own ELF image which is not something applications can currently do from userspace.

This lays the groundwork for LibGUI to use the embedded icons in an application as the default icon, e.g. for a Window.


awesomekling:
This breaks the macOS build (see CI output)

lux01:
@awesomekling macOS is fixed now :)

PR #4626, opened 2020-12-29
Applications: Add CrashReporter :^)

linusg:
...and all the infrastructure required for it, mostly LibCore+LibGUI additions for GML and some LibCoreDump tweaks.

image

crash-reporter

Also use `dbgln()` for CrashDaemon's backtrace printing so it doesn't omit the process name and PID in the debug console:

image


Applications/CrashReporter/main.cpp
```diff
@@ -0,0 +1,165 @@
...
+        return 1;
+    }
+
+    auto backtrace = coredump->backtrace();
+
+    String executable_path;
+    // FIXME: Maybe we should just embed the process's executable path
+    // in the coredump by itself so we don't have to extract it from the backtrace.
+    // Such a process section could also include the PID, which currently we'd have
+    // to parse from the filename.
```
awesomekling, line 81: These are very cool ideas. Maybe we could add them in a separate note section?

Libraries/LibCoreDump/Backtrace.cpp
```diff
@@ -0,0 +1,127 @@
...
+        return nullptr;
+
+    MappedFile object_file(path);
+    if (!object_file.is_valid())
+        return nullptr;
+
+    auto info = make<ELFObjectInfo>(move(object_file), Debug::DebugInfo { make<ELF::Image>((const u8*)object_file.data(), object_file.size()) });
+    auto* info_ptr = info.ptr();
+    s_debug_info_cache.set(path, move(info));
+    return info_ptr;
```
devsh0, line 67: I assume the caller is allowed to use the pointer returned here? Is it okay to do a move and keep the pointer to the moved object around?

awesomekling: Yes, it's a heap-allocated object, now owned by `s_debug_info_cache`.
We use the `info_ptr` temporary because `info` is nulled out by the `move()` into the hash map :)

PR #4631, opened 2020-12-29
LibTTF: Finish work started by Srimanta Barua

### LINKED ISSUE #743: LibGUI: Implement TrueType and/or OpenType fonts

sunverwerth:
Screenshot von 2020-12-29 16 16 47

Closes #2512
Fixes #743

I rebased this on the most recent master and got it back to working.

ToDos:

- [x] ~~Abstract font interface for TTF and pixel fonts~~
- [x] Use ReadonlyBytes to avoid unnecessary buffer copies in the TTF::Font class
- [x] Cache rendered glyphs in the TTF::ScaledFont class

I'll leave it at that for now. TTF system fonts and font viewer apps are probably out of scope for this PR.


Libraries/LibTTF/Cmap.cpp
```diff
@@ -90,8 +90,7 @@ Optional<Cmap::Subtable> Cmap::subtable(u32 index) const
     u16 encoding_id = be_u16(m_slice.offset_pointer(record_offset + (u32)Offsets::EncodingRecord_EncodingID));
     u32 subtable_offset = be_u32(m_slice.offset_pointer(record_offset + (u32)Offsets::EncodingRecord_Offset));
     ASSERT(subtable_offset < m_slice.size());
-    // HACK: added const_cast because of new ByteBuffer::wrap behavior, should probably find another workaround
-    auto subtable_slice = ByteBuffer::wrap(const_cast<u8*>(m_slice.offset_pointer(subtable_offset)), m_slice.size() - subtable_offset);
+    auto subtable_slice = ByteBuffer::copy(m_slice.offset_pointer(subtable_offset), m_slice.size() - subtable_offset);
```
alimpfard, line 93: (if wrap was fine, you can definitely take `ReadonlyBytes` instead of copying)

alimpfard: I noticed that you had it as one of the TODOs, sorry for not looking harder lol

sunverwerth: Ah cool, that's an alias for Span<const u8>?

alimpfard: yep.

sunverwerth: Well, that was easy

tomuta:
Without having looked at the actual changes, I think we should get this merged sooner than later, even if it has imperfections. We can always improve upon it iteratively. Don't want it to die like the last PR ;)

sunverwerth:
@tomuta I'm determined to finish this now :smile: If not today, then tomorrow.

sunverwerth:
I couldn't help myself and test ttf as a system font. The results are... interesting :laughing: 

image

Clearly, there is a lot more work to do to transition from fixed dimension bitmaps to per character sizing.
I'm on my way there but there's still a lot to do.
I would like to split up the rest of the work into a seperate PR and merge what I have so far.

So, only the LibTTF and the last commit (move default_XXX_font() into FontDatabase)

Any objections?

sunverwerth:
@awesomekling I think you clicked edit instead of reply

Ok, I'll fix the linter issues.

Are there any change requests for the code in LibTTF itself?
I haven't touched most of it since the original PR.
Maybe someone can have a quick look. @alimpfard ?

alimpfard:
I shall review a bit then :)

awesomekling:
> @awesomekling I think you clicked edit instead of reply

Whoa oops, sorry about that!

alimpfard:
> I couldn't help myself and test ttf as a system font. The results are... interesting laughing
> 
> Clearly, there is a lot more work to do to transition from fixed dimension bitmaps to per character sizing.


I think those artifacts are mainly because it's not respecting kerning (or maybe GPOS, if that's in the font).
I had kerning in my old ttf branch, but I don't think it looked anywhere near good enough with just kerning either...

sunverwerth:
> > I couldn't help myself and test ttf as a system font. The results are... interesting laughing
> > Clearly, there is a lot more work to do to transition from fixed dimension bitmaps to per character sizing.
> 
> I think those artifacts are mainly because it's not respecting kerning (or maybe GPOS, if that's in the font).
> I had kerning in my old ttf branch, but I don't think it looked anywhere near good enough with just kerning either...

The main issue is that all text rendering functions until now expect all character bitmaps to be the same height.

awesomekling:
Would be good to check that each commit still makes sense. This one looks kinda unnecessary for example: `LibGfx: Clean up Painter and remove duplicated methods for drawing TTF text.`

alimpfard:
So far as the contents go, I don't think I have any nits to pick :)

sunverwerth:
I squashed the one commit and edited the commit messages that were a bit too long. Waiting for the CI to complete.

PR #4633, opened 2020-12-29
AK: Make binary_search signature more generic.

asynts:
There are quite a few template shenanigans here to unpack:

  - The return type depends on the container type and is quite difficult to write down exactly. Here it is deduced by `decltype(haystack[0])` which is a bit of a hack but seems to work reliable.

  - The `container` parameter has the type `Container&&` because we need to accept lvalue and rvalue references here. Same for the `needle` parameter.

  - Previously, the `needle` parameter was deduced from the `container` type and this caused problems with stuff like `NonnullOwnPtr` and `OwnPtr` being different types.

  - Previously, the default value for `compare` did not work because the compiler was not able to deduce the type of `Compare`. Now the default parameter is passed as default template parameter and the default constructor makes this work.

### DIFF
--- a/AK/BinarySearch.h
+++ b/AK/BinarySearch.h
@@ -32,14 +32,16 @@
 
 namespace AK {
 
-template<typename T>
-constexpr int integral_compare(const typename RemoveConst<T>::Type& a, const typename RemoveConst<T>::Type& b)
-{
-    return a - b;
-}
+struct IntegralComparator {
+    constexpr auto operator()(auto& lhs, auto& rhs) { return lhs - rhs; }
+};
 
-template<typename T, typename Compare>
-constexpr T* binary_search(Span<T> haystack, const typename RemoveConst<T>::Type& needle, Compare compare = integral_compare, size_t* nearby_index = nullptr)
+template<typename Container, typename Needle, typename Comparator = IntegralComparator>
+constexpr auto binary_search(
+    Container&& haystack,
+    Needle&& needle,
+    size_t* nearby_index = nullptr,
+    Comparator comparator = Comparator {}) -> decltype(&haystack[0])
 {
     if (haystack.size() == 0) {
         if (nearby_index)
@@ -51,7 +53,7 @@ constexpr T* binary_search(Span<T> haystack, const typename RemoveConst<T>::Type
     size_t high = haystack.size() - 1;
     while (low <= high) {
         size_t middle = low + ((high - low) / 2);
-        int comparison = compare(needle, haystack[middle]);
+        auto comparison = comparator(needle, haystack[middle]);
         if (comparison < 0)
             if (middle != 0)
                 high = middle - 1;
--- a/AK/Tests/TestBinarySearch.cpp
+++ b/AK/Tests/TestBinarySearch.cpp
@@ -38,24 +38,32 @@ TEST_CASE(vector_ints)
     ints.append(2);
     ints.append(3);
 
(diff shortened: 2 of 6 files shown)


AK/BinarySearch.h
```diff
@@ -32,14 +32,16 @@
...
-    return a - b;
-}
+struct IntegralComperator {
+    constexpr auto operator()(auto& lhs, auto& rhs) { return lhs - rhs; }
+};
 
-template<typename T, typename Compare>
-constexpr T* binary_search(Span<T> haystack, const typename RemoveConst<T>::Type& needle, Compare compare = integral_compare, size_t* nearby_index = nullptr)
+template<typename Container, typename Needle, typename Comperator = IntegralComperator>
+constexpr auto binary_search(
```
alimpfard, line 40: Does `decltype(auto)` work here?
it'd help the return type issue if the compiler doesn't get mad at it

asynts: I do get an error in that case:

~~~
inconsistent deduction for auto return type: 'std::nullptr_t' and then 'const unsigned int*'
~~~

I also tried leaving out the trailing return type entirely (just `auto` at the beginning) and get the same error.

alimpfard: oh it's also returning `nullptr` at some point...:sad:

AK/BinarySearch.h
```diff
@@ -32,14 +32,16 @@
 
 namespace AK {
 
-template<typename T>
-constexpr int integral_compare(const typename RemoveConst<T>::Type& a, const typename RemoveConst<T>::Type& b)
-{
-    return a - b;
-}
+struct IntegralComperator {
```
ADKaster, line 35: I think to be 100% grammatical this should be `IntegralComparator`, and `typename Comparator` below. No `e` 🙂. 

Could probably also constrain the types with some concepts, like BinaryOperator
? Did we write a set of concepts yet? 🤔 Might be a fun exercise

alimpfard: more concepts? yes please!

alimpfard: I've been sprinkling some here and there, but it'd be really nice to have some useful ones around at least

asynts: I fixed the spelling mistake but I wasn't able to do the concept stuff.

I got `clang-format 11.0.0`  installed and it doesn't seem to be able to handle concepts just yet. Also, all concept code so far is surrounded with `#ifdef` macros, not sure if this is so nice.

I guess sometime in early 2021 we have to eventually drop Clang and Coverity support and can use all the nice GCC 11 features unrestricted...

alimpfard: We can all dream of nice and shiny concepts in the meantime :)

ADKaster: https://godbolt.org/z/ac6Thz :eyes: clang 11 seems to handle them fine though. Just clang-format giving issues?

asynts: Here is a snippet I had trouble with:

~~~none
$ cat main.cpp
#include <concepts>

template <typename T>
concept Container = requires (T t) {
    { t.size() } -> std::same_as<std::size_t>;
};

int main() {}
$ clang-format main.cpp
#include <concepts>

template<typename T>
concept Container = requires(T t)
{
    {
        t.size()
    }
    ->std::same_as<std::size_t>;
};

int main() { }
~~~

ADKaster: Looks like this patch https://reviews.llvm.org/D79773 landed on December 4th. Hopefully next release  will have it, it looks like it should fix that formatting.

PR #4641, opened 2020-12-29
Applications: Added a new MouseSettings application

IdanHo:
image
This app allows the user to easily adjust his mouse's acceleration as well as the scrollwheel's global scroll length.
The mouse speed changes would not be noticeable in qemu as by default serenity uses VMWareBackdoor when available which lets the host handle mouse movement instead of the guest (Serenity), so in order to test this on a none-baremetal pc the VMWareBackdoor has to be disabled.
This pull also removes the scroll length setting from the terminal, as the global one is now available.


Services/WindowServer/WindowManager.cpp
```diff
@@ -163,6 +163,22 @@ Gfx::IntSize WindowManager::resolution() const
     return Screen::the().size();
 }
 
+void WindowManager::set_acceleration_factor(double factor)
+{
+    Screen::the().set_acceleration_factor(factor);
+    dbg() << "Saving acceleration factor: " << factor << " to config file at " << m_config->file_name();
```
linusg, line 169: Please use `dbgln()` going forward, the streaming variants are being phased out :)

```suggestion
    dbgln("Saving acceleration factor {} to config file at {}", factor, m_config->file_name());
```

Services/WindowServer/WindowManager.cpp
```diff
@@ -163,6 +163,22 @@ Gfx::IntSize WindowManager::resolution() const
...
+    m_config->write_entry("Mouse", "AccelFactor", String::formatted("{}", factor));
+    m_config->sync();
+}
+
+void WindowManager::set_scroll_scalar(unsigned scalar)
+{
+    Screen::the().set_scroll_scalar(scalar);
+    dbg() << "Saving scroll scalar: " << scalar << " to config file at " << m_config->file_name();
```
linusg, line 177:
```suggestion
    dbgln("Saving scroll scalar {} to config file at {}", scalar, m_config->file_name());
```

Services/WindowServer/WindowManager.cpp
IdanHo: fixed!

bcoles:
If the `ScrollLength` option is no longer valid for the `Terminal` application, it can also be removed from the default configuration file:

https://github.com/SerenityOS/serenity/blob/675bfb934bcd5a00999a566bb4a080184912c0e4/Base/home/anon/.config/Terminal.ini#L6

awesomekling:
Indeed, and let's preserve the current scroll length setting (4) :)

Services/WindowServer/WindowManager.cpp
```diff
@@ -163,6 +163,22 @@ Gfx::IntSize WindowManager::resolution() const
     return Screen::the().size();
 }
 
+void WindowManager::set_acceleration_factor(double factor)
+{
+    Screen::the().set_acceleration_factor(factor);
+    dbgln("Saving acceleration factor {} to config file at {}", factor, m_config->file_name());
+    m_config->write_entry("Mouse", "AccelFactor", String::formatted("{}", factor));
```
awesomekling, line 170: Let's call this setting `AccelerationFactor` instead of `AccelFactor`.

Services/WindowServer/WindowManager.cpp
```diff
@@ -163,6 +163,22 @@ Gfx::IntSize WindowManager::resolution() const
...
+    m_config->sync();
+}
+
+void WindowManager::set_scroll_scalar(unsigned scalar)
+{
+    Screen::the().set_scroll_scalar(scalar);
+    dbgln("Saving scroll scalar {} to config file at {}", scalar, m_config->file_name());
+    m_config->write_entry("Mouse", "ScrollScalar", String::number(scalar));
```
awesomekling, line 178: What's a "scalar" in this context? Can we find a better word? ScrollWheelStep? ScrollStepSize?

tomuta:
A setting to reverse left/right would be nice, too. One of my kids is a lefty so we actually use that setting ;-)

Services/WindowServer/WindowManager.cpp
IdanHo: i think "ScrollStepSize" is fitting, should i change it in the IPC def as well, or just here?

Services/WindowServer/WindowManager.cpp
IdanHo: done!

Services/WindowServer/WindowManager.cpp
IdanHo: also done

IdanHo:
> A setting to reverse left/right would be nice, too. One of my kids is a lefty so we actually use that setting ;-)

I'll create a seperate PR for that change, as that requires wider-scope changes

PR #4652, opened 2020-12-30
Playground: Add autocomplete

alimpfard:
Hello friends,
do you think that Playground is cool? of course you do!
do you think that it needs more autocomplete? of course you do!
don't you agree that it would be 100 times better if you could actually use it without cross-checking property names? OF COURSE YOU DO!

CxByte presents, _more autocomplete\*_
shot-2020-12-30_14-03-03
shot-2020-12-30_14-03-39

-------

<sub>\* Rules and ToS apply, please thoroughly read /dev/zero before accepting this pull request.</sub>


Libraries/LibGUI/AutocompleteProvider.cpp
```diff
@@ -64,11 +62,8 @@ class AutoCompleteSuggestionModel final : public GUI::Model {
                 return suggestion.completion;
             }
             if (index.column() == Column::Icon) {
-                // TODO: Have separate icons for fields, functions, methods etc
-                // FIXME: Probably should have different icons for the different kinds, rather than for "c++".
-                if (suggestion.kind == CompletionKind::Identifier)
-                    return *s_cplusplus_icon;
-                return *s_cplusplus_icon;
+                // TODO
+                return {};
```
alimpfard, line 66: I could either just return the C++ icon for `Language::Cpp`, or try to design a few (likely _bad_) icons, we currently really need two right now: `Cpp`+`Identifier`, `Unspecified`+`Identifier`

awesomekling: Make the best icon you can and we'll iterate :)

alimpfard: I'll try then.

DevTools/Playground/main.cpp
```diff
@@ -25,12 +25,139 @@
...
+                    should_push_state = false;
+                } else {
+                    state = AfterClassName;
+                }
+                break;
+            }
+        }
+
+        Vector<GUI::AutocompleteProvider::Entry> entries;
+        dbg() << "In state " << (int)state << ", have " << class_names.size() << " class names, last seen identifier " << identifier_string;
```
awesomekling, line 113: Please use `dbgln()` in new code

alimpfard: I've removed that, it's not necessary - just a leftover.

Libraries/LibGUI/AutocompleteProvider.cpp
alimpfard: Icons are hard, evidently.
I hope someone that doesn't entirely suck at pixel art makes a good-looking set one day lol

PR #4653, opened 2020-12-30
AK: Replace some SFINAE with requires clauses, clean up existing ones

ADKaster:
Add requires clauses to constraints on InputStream and OutputStream
operator<< / operator>>. Make the constraint on String::number a
requires clause instead of SFINAE. Also, fix some unnecessary IsSame in
Trie where specialized traits exist for the given use cases.

@alimpfard `IsSame<T, decltype(nullptr)>` is a weird way to spell `IsNullPtr<T>` 😉 

Pretty much all of these specific use cases scream "this should use a concept instead of a requires clause". Probably need `clang-format-12` for that to not make a mess of things, though.

### DIFF
--- a/AK/Stream.h
+++ b/AK/Stream.h
@@ -147,14 +147,14 @@ InputStream& operator>>(InputStream& stream, Optional<T>& value)
     return stream;
 }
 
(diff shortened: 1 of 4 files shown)


alimpfard:
Huh. I never knew that existed, I've never seen `std::is_null_pointer` before either...
TIL!

PR #4661, opened 2020-12-30
Allow arbitrary metadata in coredumps, use it to store assertion messages

linusg:
image

---

This is the first time I'm working on the Kernel & syscalls, so bear with me :)

### DIFF
--- a/Kernel/Syscalls/process.cpp
+++ b/Kernel/Syscalls/process.cpp
@@ -24,6 +24,7 @@
  * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
+#include <AK/Types.h>
 #include <Kernel/Process.h>
 #include <Kernel/SharedBuffer.h>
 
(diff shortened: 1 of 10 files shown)


Kernel/Syscalls/process.cpp
```diff
@@ -64,4 +64,17 @@ int Process::sys$set_process_name(Userspace<const char*> user_name, size_t user_
     return 0;
 }
 
+// FIXME: If we can get a syscall function with 5 arguments, we could pass key/value here.
```
linusg, line 67: Not sure what to do about this. Sending a single "key:value" string is probably an ok-ish solution, although the internet suggests syscalls with more than three arguments should be possible...

Kernel/Syscalls/process.cpp
```diff
@@ -64,4 +64,17 @@ int Process::sys$set_process_name(Userspace<const char*> user_name, size_t user_
...
+int Process::sys$set_coredump_metadata(Userspace<const char*> data, size_t data_length)
+{
```
linusg, line 69: I was thinking whether this needs a `REQUIRE_PROMISE()` but I figured not, otherwise all processes using pledge() would have to use it or their coredumps won't contain assertion messages...

Kernel/Syscalls/process.cpp
alimpfard: what's wrong with the existing way? just put all five into a struct and pass the struct's address to the kernel?

Kernel/Syscalls/process.cpp
alimpfard: more like they'd crash because they crashed, so they'd crash because they crashed, ...

Kernel/Syscalls/process.cpp
linusg: 🤯

You see, I have clearly no idea what I'm doing.

---

So, something like this? (names need some work, obvs)

```cpp
struct KeyAndValue {
    Userspace<const char*> key;
    size_t key_length;
    Userspace<const char*> value;
    size_t value_length;
};

int Process::sys$set_coredump_metadata(Userspace<const KeyAndValue*> key_and_value)
{
}
```

awesomekling: Right, look at `sys$mmap` for example. All the `SC_foo_params` structs are defined in `Kernel/API/Syscall.h`

linusg: Done, amazing!

Kernel/Syscalls/process.cpp
```diff
@@ -64,4 +64,23 @@ int Process::sys$set_process_name(Userspace<const char*> user_name, size_t user_
...
+        return -EINVAL;
+    if (params.value.length == 0)
+        return -EINVAL;
+    auto copied_key = copy_string_from_user(params.key.characters, params.key.length);
+    if (copied_key.is_null())
+        return -EFAULT;
+    auto copied_value = copy_string_from_user(params.value.characters, params.value.length);
+    if (copied_value.is_null())
+        return -EFAULT;
+    m_coredump_metadata.set(move(copied_key), move(copied_value));
```
awesomekling, line 82: Let's impose some limitations here so a userspace process can't fill the kernel heap with however much memory it wants.
Just some random numbers are fine, e.g maximum 16 keys, maximum 16KB per value (or whatever you think makes sense.)

linusg: Seems sensible :+1:
16 entries / 16 KiB each should be more than enough.

PR #4665, opened 2020-12-30
AK+ProtocolServer: Properly close download stream fd's

alimpfard:
```
< kling> CxByte: ProtocolServer(27): Protocol: pipe() failed: Too many open files
(snip)
< kling> max fds per process right now is uh.. 64 i think
(snip)
< kling> clicked on a link on google.com
< kling> the "privacy" link at the bottom
```

This PR exchanges that error for `assertion failed: child.is_inline()`
while it doesn't actually fix the matter at hand, it should be much harder to actually make a ProtocolServer instance use up all 64 available fds now.

Also re-enables favicons. oops.

### DIFF
--- a/AK/FileStream.h
+++ b/AK/FileStream.h
@@ -54,8 +54,9 @@ public:
     ~InputFileStream()
     {
         if (m_file) {
+            fflush(m_file);
             if (m_owned)
-                fflush(m_file);
+                fclose(m_file);
         }
     }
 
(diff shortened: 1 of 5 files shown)


awesomekling:
This `on_destruction` thing feels hackish but okay for now

alimpfard:
it _is_ a hack, I tried thinking of some other way to close that fd, but nothing came up
the only other more reasonable solution I thought of was closing it in `IPC::File`'s destructor, but that thing is copied around and treated like a value type

PR #4669, opened 2020-12-30
AK: Add operator* and operator-> overloads in Optional.

asynts:
This is a useful abbreviation because it can be quite cumbersome to always spell out `value()`. (When multiple `Optional` variables are involved.)

### DIFF
--- a/AK/Optional.h
+++ b/AK/Optional.h
@@ -155,6 +155,12 @@ public:
         return fallback;
     }
 
+    const T& operator*() const { return value(); }
+    T& operator*() { return value(); }
+
+    const T* operator->() const { return &value(); }
+    T* operator->() { return &value(); }
+
 private:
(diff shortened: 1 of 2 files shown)


linusg:
Do you possibly have clang-format 11 installed? I do, and it seems to fix some cases that clang-format 10 can't handle properly, as seen in this PR - I'm afraid you'll have to skip these for now and keep the old formatting, CI is not happy :confused:

asynts:
Ubuntu 20.10 seems to have it, can't we just update? https://packages.ubuntu.com/groovy/clang-format-11

tomuta:
Ubuntu 20.10 doesn't work nearly as stable/well as 20.04 on at least three of my machines :(

asynts:
I "fixed" the formatting for now.

awesomekling:
Hmm, all right

PR #4672, opened 2020-12-30
LibProtocol: Ensure download is finished before invoking on_finished

### LINKED ISSUE #4668: LibWeb/ProtocolServer: Certain redirects work inconsistently
For example, clicking on "Google Search" or "I'm Feeling Lucky" may show these pages instead:
Screenshot from 2020-12-30 19-28-54
But it sometimes work (it seems closing and reopening the browser helps with making it work):
Screenshot from 2020-12-30 19-33-03

~~But, if it works, the browser completely locks up, but doesn't go into the not responding state.~~ Separate issue with Google Doodles.
And certain redirects work consistently, for example, clicking the privacy link at the bottom of the google page.

alimpfard:
Fixes #4668

### DIFF
--- a/Libraries/LibWeb/Loader/ResourceLoader.cpp
+++ b/Libraries/LibWeb/Loader/ResourceLoader.cpp
@@ -184,6 +184,7 @@ void ResourceLoader::load(const LoadRequest& request, Function<void(ReadonlyByte
                     error_callback("HTTP load failed");
                 return;
             }
+            deferred_invoke([&](auto&) { const_cast<RefPtr<Protocol::Download>&>(download) = nullptr; });
             success_callback(payload, response_headers);
         };
         download->set_should_buffer_all_input(true);
--- a/Libraries/LibGemini/Job.cpp
+++ b/Libraries/LibGemini/Job.cpp
@@ -165,10 +165,12 @@ void Job::finish_up()
     m_state = State::Finished;
     flush_received_buffers();
     if (m_received_size != 0) {
-        // FIXME: What do we do? ignore it?
-        //        "Transmission failed" is not strictly correct, but let's roll with it for now.
+        // We have to wait for the client to consume all the downloaded data
+        // before we can actually call `did_finish`. in a normal flow, this should
+        // never be hit since the client is reading as we are writing, unless there
+        // are too many concurrent downloads going on.
         deferred_invoke([this](auto&) {
-            did_fail(Error::TransmissionFailed);
+            finish_up();
         });
         return;
     }
--- a/Libraries/LibHTTP/Job.cpp
+++ b/Libraries/LibHTTP/Job.cpp
@@ -376,10 +376,12 @@ void Job::finish_up()
 
     flush_received_buffers();
     if (m_buffered_size != 0) {
-        // FIXME: What do we do? ignore it?
-        //        "Transmission failed" is not strictly correct, but let's roll with it for now.
+        // We have to wait for the client to consume all the downloaded data
+        // before we can actually call `did_finish`. in a normal flow, this should
+        // never be hit since the client is reading as we are writing, unless there
+        // are too many concurrent downloads going on.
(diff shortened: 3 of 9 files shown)


alimpfard:
I can't think of any better way to do this, so that's that.
at least #4668 is fixed :shrug:

Libraries/LibWeb/Loader/ResourceLoader.cpp
```diff
@@ -184,6 +184,7 @@ void ResourceLoader::load(const LoadRequest& request, Function<void(ReadonlyByte
                     error_callback("HTTP load failed");
                 return;
             }
+            deferred_invoke([&](auto&) { const_cast<RefPtr<Protocol::Download>&>(download) = nullptr; });
```
awesomekling, line 187: Wait, what's going on here?

`deferred_invoke` will call this lambda when returning to the event loop. By that point, won't we be somewhere else entirely, and not have `download` on the stack anymore?

alimpfard: `download` is captured as a copy (of a `RefPtr`)
the lambda keeps the download alive until success_callback (or the error one) is called and done

it's such a convoluted mess, but I can't find a better way that doesn't involve making LibWeb support streaming downloads

awesomekling: Oh.. okay.

alimpfard: precisely my reaction when I realised _that_ was why the downloads weren't being deleted after they finished

tomuta: @alimpfard `download` is certainly not captured as a copy because of `[&]`! It probably works because nothing on the stack has trashed it, but this is definitely merely a reference capture to a local variable.

tomuta: This line looks really weird and looks like it will trash the stack at that location. This line should not be necessary, and the reason why it doesn't get freed is probably because `download` is copy-captured here:

https://github.com/alimpfard/serenity/blob/b87e92d70654509d69511c4923f2ca787f003d06/Libraries/LibWeb/Loader/ResourceLoader.cpp#L173

So, unless something resets `download->on_buffered_download_finish` you end up with a circular reference. So, I think all you really need is doing this instead to clear the circular reference:
```c++
deferred_invoke([download](auto&) { download->on_buffered_download_finish = nullptr; });
```

tomuta: The `success_callback` may keep the underlying download object alive, but not necessarily the RefPtr pointing to it.

alimpfard: Hmmm...
Here's the way I see that:
`download` is a RefPtr, and LibProtocol keeps it alive until it calls `on_finish`
`download->on_buffered_download_finish` is some object (internal to `Function`) on the heap.
giving a copy of `download` to `on_buffered_download_finish` creates a cyclic reference chain between `download` and `on_buffered_download_finish`, where `download` will end up with a refcount of 1.

taking a reference to `download` in the `deferred_invoke` lambda should be fine, as it's not pointing into stack - it's pointing to something kept around by `on_buffered_download_finish` (which is kept alive by `download` and is keeping `download` alive).
Clearing `download` will drop its refcount, which will in turn destroy `on_buffered_download_finished` and break the cycle.
it's too complicated and weird though.

where did that `success_callback` keeping download alive come from though?

tomuta: You're right, the reference probably points to the copy-captured value of the other lambda. So this code assumes that `on_buffered_download_finish` doesn't get cleared before the lambda referencing `download` finishes execution. Seems a little brittle but I guess works...

PR #4673, opened 2020-12-30
Upgrade clang-format to version 11

linusg:
Compared to version 10 this fixes a bunch of formatting issues, mostly around structs/classes with attributes like `[[gnu::packed]]`, and incorrect insertion of spaces in parameter types (`T &`/`T &&`).
I also removed a bunch of `// clang-format off/on` and `FIXME` comments that are no longer relevant - on the other hand it tried to destroy a couple of neatly formatted comments, so I had to add some as well.

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -21,13 +21,17 @@ jobs:
     # sudo apt-get update -qq
 
     - name: Purge interfering packages
-      # Remove GCC 9 (installed by default)
-      run: sudo apt-get purge -y gcc-9 g++-9 libstdc++-9-dev
+      # Remove GCC 9 and clang-format 10 (installed by default)
+      run: sudo apt-get purge -y gcc-9 g++-9 libstdc++-9-dev clang-format-10
     - name: Install dependencies
       # These packages are already part of the ubuntu-20.04 image:
-      # clang-format-10 cmake gcc-10 g++-10 shellcheck libgmp-dev
+      # cmake gcc-10 g++-10 shellcheck libgmp-dev
       # These aren't:
-      run: sudo apt-get install libstdc++-10-dev libmpfr-dev libmpc-dev ninja-build npm
+      run: |
+        wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key | sudo apt-key add -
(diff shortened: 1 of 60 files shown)


tomuta:
Does this force everyone to upgrade to 20.10? I have very annoying problems with 20.10 on at least 3 of my computers that work flawlessly with 20.04 (all three have quite different hardware). Or is there a way to get this onto 20.04?

linusg:
> Does this force everyone to upgrade to 20.10? I have very annoying problems with 20.10 on at least 3 of my computers that work flawlessly with 20.04 (all three have quite different hardware). Or is there a way to get this onto 20.04?

No, it just forces everyone to use clang-format 11 :)
I have no idea about Ubuntu, but you can probably grab a pre-built binary from https://releases.llvm.org/download.html (`https://github.com/llvm/llvm-project/releases/download/llvmorg-11.0.0/clang+llvm-11.0.0-x86_64-linux-gnu-ubuntu-20.04.tar.xz`).

I understand it's annoying for people using distributions that are "lacking behind", but it's equally annoying for people using distributions that are up to date. (I'm on Fedora 33, which has clang 11 by default).

linusg:
It's also just a test for now, as this was mentioned in https://github.com/SerenityOS/serenity/pull/4669#issuecomment-752753327 and that PR contains formatting changes where the current version produces weird results.

If CI passes and there's no major backlash, I'd be happy to have it merged. Otherwise :man_shrugging:

Lubrsi:
Huh, getting a macOS runner both Ubuntu runners, what a weird sight. I think anything they don't label as "latest" they have much less capacity for.

linusg:
That's concerning :smile:
If it slows down CI builds we might have to stick with Ubuntu 20.04 and get clang-format 11 from elsewhere.

Lubrsi:
I mean, macOS managed to do a full toolchain rebuild before getting even one ubuntu runner. That's pretty impressive

Lubrsi:
Apparently they were gonna put clang-11 and all that onto ubuntu 20.04 in december 14th, but it fell through because it didn't get released generally to 20.04 on time.

linusg:
@tomuta there's also https://apt.llvm.org/ which is probably even better than the tar file :)

tomuta:
> @tomuta there's also https://apt.llvm.org/ which is probably even better than the tar file :)

Nice, I can live with that :)

PR #4674, opened 2020-12-30
DynamicLoader: Handle Loader.so being invoked directly as an executable

lux01:
Loader.so is an actual executable, as well as the interpreter for dynamic
libraries. Currently launching Loader.so as a standalone executable results
in an obsucre crash as it tries to load itself over itself.

Now we at least print a helpful message saying that you're doing the wrong
thing and exit gracefully. In future we may wish to allow users to specify
additional options to learn more about what's going on during dynamic
linking, such as ld-linux.so.2 on Linux.

### DIFF
--- a/Userland/DynamicLoader/main.cpp
+++ b/Userland/DynamicLoader/main.cpp
@@ -250,6 +250,20 @@ static void clear_temporary_objects_mappings()
     g_loaders.clear();
 }
 
+static void display_help()
+{
+    const char message[] =
+        R"(You have invoked `Loader.so'. This is the helper program for programs that
+use shared libraries. Special directives embedded in executables tell the
+kernel to load this program.
+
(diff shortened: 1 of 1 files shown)


Userland/DynamicLoader/main.cpp
```diff
@@ -250,6 +250,20 @@ static void clear_temporary_objects_mappings()
...
+{
+    const char message[] =
+        R"(You have invoked `Loader.so'. This is the helper program for programs that
+use shared libraries. Special directives embedded in executables tell the
+kernel to load this program.
+
+This helper program loads the shared libraries needed by the program,
+prepares the program to run, and runs it. You do not need to invoke
+this helper program directly.
+)";
```
awesomekling, line 263: This is a very helpful and informative error message.

PR #4677, opened 2020-12-31
Kernel: Fix heap expansions deadlock

### LINKED ISSUE #4675: Kernel: When loading a web page, if a kmalloc heap extension happens when NetworkTask is scheduled, the system deadlocks
Keep trying to load webpages, e.g. DDG, if this happens:
```
[NetworkTask(5:5)]: kmalloc(): Adding memory to heap at V0xc0801000, bytes: 1048576
```
The kernel will deadlock. If it happens outside NetworkTask, it carries on normally.
GDB trace:
```
#0  0xc01dc2c7 in ?? ()
#1  0xc01b348d in ?? ()
#2  0xc01ad777 in ?? ()
#3  0xc02115f6 in ?? ()
#4  0xc0212089 in ?? ()
#5  0xc021210d in ?? ()
#6  0xc01cf256 in ?? ()
#7  0xc020b92f in ?? ()
#8  0xc020b9e1 in ?? ()
#9  0xc01b30af in ?? ()
#10 0xc01b34f1 in ?? ()
#11 0xc01ad777 in ?? ()
#12 0xc01e2c2d in ?? ()
#13 0xc01e2e0a in ?? ()
#14 0xc01f4e40 in ?? ()
#15 0xc0168303 in ?? ()
#16 0xc01190be in ?? ()
```
(It seems to be a different trace every time, so I don't know how useful this is)

tomuta:
If a heap expansion is triggered by allocating from e.g. the
RangeAllocator, which may be holding a spin lock, we cannot
immediately allocate another block of backup memory, which could
require the same locks to be acquired. So, defer allocating the
backup memory

Fixes #4675

### DIFF
--- a/Kernel/Heap/kmalloc.cpp
+++ b/Kernel/Heap/kmalloc.cpp
@@ -52,6 +52,8 @@
 
 static RecursiveSpinLock s_lock; // needs to be recursive because of dump_backtrace()
 
+static void kmalloc_allocate_backup_memory();
+
 struct KmallocGlobalHeap {
     struct ExpandGlobalHeap {
         KmallocGlobalHeap& m_global_heap;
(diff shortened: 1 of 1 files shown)


Kernel/Heap/kmalloc.cpp
```diff
@@ -94,7 +96,11 @@ struct KmallocGlobalHeap {
             // backup heap before returning. Otherwise we potentially lose
             // the ability to expand the heap next time we get called.
             ScopeGuard guard([&]() {
-                m_global_heap.allocate_backup_memory();
+                // We may need to defer allocating backup memory because the
+                // heap expansion may have been triggered while holding some
+                // other spinlock. We the expansion happens to need the same
```
awesomekling, line 101: Typo, `We the`

awesomekling:
Makes sense

PR #4686, opened 2020-12-31
Kernel: Boot with Partition UUID

supercomputer7:
Very simple, yet a neat feature that we all probably use in our Linux machines.
One can get the partition UUID (I quickly did it with an hex editor, but `cfdisk` also shows this value) of a GUID partition and specify it in a textual representation of a UUID in the kernel boot argument - `root=PARTUUID=123e4567-e89b-12d3-a456-426614174000`. If the kernel finds your partition, it will boot it :)
I used the [wikipedia article about UUIDs](https://en.wikipedia.org/wiki/Universally_unique_identifier) to understand what is expected from the implementation.

### DIFF
--- a/AK/UUID.cpp
+++ b/AK/UUID.cpp
@@ -1,0 +1,92 @@
+/*
+ * Copyright (c) 2020, Liav A. <liavalb@hotmail.co.il>
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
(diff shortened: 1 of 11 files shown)


AK/UUID.h
```diff
@@ -0,0 +1,63 @@
...
+    bool operator<=(const UUID&) const = delete;
+    bool operator>=(const UUID&) const = delete;
+    bool operator<(const UUID&) const = delete;
+    bool operator>(const UUID&) const = delete;
+
+    bool is_null() const;
+
+private:
+    void convert_string_view_to_uuid(const StringView&);
+    ByteBuffer m_uuid_buffer;
```
awesomekling, line 58: A UUID is always going to be the same length (128 bits), so let's use an array instead of a heap-allocating `ByteBuffer` :)

AK/UUID.h
```diff
@@ -0,0 +1,63 @@
...
+
+namespace AK {
+
+class UUID {
+public:
+    UUID();
+    explicit UUID(ByteBuffer);
+    explicit UUID(const String&);
+    explicit UUID(const StringView&);
+    explicit UUID(const char*);
```
awesomekling, line 41: Why do we need all these different constructors?

AK/UUID.h
```diff
@@ -0,0 +1,63 @@
...
+    ~UUID() { }
+
+    String as_string() const;
+    const ByteBuffer& as_bytebuffer() const;
```
awesomekling, line 45: `as_foo()` generally means "I know this is a `Foo` so just give me an appropriate reference type"
`to_foo()` generally means "I don't know what this is but give it to me in the form of a `Foo` even if you have to generate it"

awesomekling (changes requested):

AK/UUID.h
supercomputer7: You're right, the `UUID(const char*)` and `UUID(const String&)` constructors are not needed.

AK/UUID.cpp
```diff
@@ -0,0 +1,103 @@
...
+}
+
+bool UUID::is_null() const
+{
+    for (size_t index = 0; index < 16; index++) {
+        if (m_uuid_buffer.data()[index] != 0)
+            return false;
+    }
+    return true;
+}
```
linusg, line 101: 1. This looks like it could use `AK::all_of()` :)
2. Maybe `is_zero()` is a better name, as that's what it actually checks?

`AK::MACAddress` does both:

https://github.com/SerenityOS/serenity/blob/bbe787a0afd51307cdac2c023895c19df31519ca/AK/MACAddress.h#L80-L83

supercomputer7: I'll fix it right away, thanks! :)

linusg (commented):
two more small suggestions for UUID. I don't think I'm qualified to judge the kernel changes, but this is very cool!

AK/UUID.h
```diff
@@ -0,0 +1,64 @@
...
+namespace AK {
+
+class UUID {
+public:
+    UUID();
+    explicit UUID(ByteBuffer);
+    explicit UUID(const StringView&);
+    ~UUID() { }
+
+    String as_string() const;
```
awesomekling, line 43: Please call this `to_string()`

AK/UUID.h
```diff
@@ -0,0 +1,64 @@
...
+    ByteBuffer to_bytebuffer() const;
```
awesomekling, line 44: Do we actually need this API? It's always nice if we can avoid heap allocated buffers.

AK/UUID.cpp
```diff
@@ -0,0 +1,107 @@
...
+void UUID::convert_string_view_to_uuid(const StringView& uuid_string_view)
+{
+    ASSERT(uuid_string_view.length() == 36);
+    auto first_unit = decode_hex(uuid_string_view.substring_view(0, 8));
+    auto second_unit = decode_hex(uuid_string_view.substring_view(9, 4));
+    auto third_unit = decode_hex(uuid_string_view.substring_view(14, 4));
+    auto fourth_unit = decode_hex(uuid_string_view.substring_view(19, 4));
+    auto fifth_unit = decode_hex(uuid_string_view.substring_view(24, 12));
+
+    auto result = ByteBuffer::create_zeroed(16);
```
awesomekling, line 58: No need for a ByteBuffer here, a basic stack Array works fine :)

awesomekling (changes requested):

PR #4687, opened 2020-12-31
Implement shortest round-trip float to string conversion

XavierCooney:
This is currently very much WIP, but because it's kind of big I think it'd be good to get some feedback before I finish and polish everything. There's a big comment in `double_to_digits` which should explain most things. So far I've only actually added calls to this code in `Format.cpp`, but I'm halfway done with the adding to the JS engine, but there's also the printf implementation and the calculator which also have their own not-quite-correct float to string conversion. I'm also going to add another mode to `double_to_string` to ensure there are trailing zeroes up to precision, which should allow it to act more like the old formatter and the requirements for printf. The current implementations suffer from information loss, and sometimes an annoying '9999' at the end, but for the most part that probably doesn't matter a whole lot. However, a correct algorithm is needed to meet the JS spec, and also a number of the implementations don't seem to handle somewhat big floats, for example:
image
This could probably be correct in some way, but alas no simple multiply/divide by 10 algorithm can correctly get the shortest round-trip representation, thus this PR.

I also had to move `SignedBigInteger` and `UnsignedBigInteger` to AK (from LibCrypto), and I also added a simple `shift_left` method to `UnsignedBigInteger`.

### DIFF
--- a/AK/FloatToDigits.cpp
+++ b/AK/FloatToDigits.cpp
@@ -1,0 +1,376 @@
+/*
+ * Copyright (c) 2020, Xavier Cooney <xavier.cooney03@gmail.com>
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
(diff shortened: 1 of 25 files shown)


linusg:
For the first commit we should probably add a `using AK::SignedBigInteger;` and `using AK::UnsignedBigInteger;` so we can drop the `AK::` from every use, that's how most (all?) things in AK do it.

sunverwerth:
Very cool :+1:  I worked on JS::Value::to_string() recently but found out I was very much out of my element :D It has very specific rules as to when to switch to a scientific representation. Are these rules universal or specific to JS? Do you think we can unify all these implementations under one central API?

XavierCooney:
Closing this now because I don't have enough time to finish it right now, but I might be able in the future. If anyone else wants to finish this feel free.

PR #4691, opened 2020-12-31
WindowServer: Send WindowDeactivated event to windows blocked by modal

linusg:
When a new modal window is created, we still want to forward the `WindowDeactivated` event to its parent window, despite it being blocked by the newly created modal (which causes WindowServer's `Window::event()` to ignore all incoming events from WindowManager for that window).

This fixes the "terminal doesn't stop blinking when blocked by modal window" bug.


awesomekling:
Ohh, thanks for tracking that down :D

PR #4695, opened 2020-12-31
LibWeb: Clear circular download reference when download finished

tomuta:


Libraries/LibWeb/Loader/ResourceLoader.cpp
```diff
@@ -184,7 +184,10 @@ void ResourceLoader::load(const LoadRequest& request, Function<void(ReadonlyByte
                     error_callback("HTTP load failed");
                 return;
             }
-            deferred_invoke([&](auto&) { const_cast<RefPtr<Protocol::Download>&>(download) = nullptr; });
+            deferred_invoke([download](auto&) {
+                // Clear circular reference of `download` captured by copy
+                const_cast<Protocol::Download&>(*download).on_buffered_download_finish = nullptr;
```
alimpfard, line 189: Continuing on from https://github.com/SerenityOS/serenity/pull/4672#discussion_r550681708,

this is what I see here:
now what's happening is that `download` (a field in the anonymous class created by the lambda) is being copied into this other lambda, then the other side of the cyclic chain is cleared, the refcount of `download` drops to 1 (was 2 when we entered this lambda), and it's finally deleted when the object containing this lambda is deleted in the impl of `deferred_invoke`.

I think it's the same thing either way, both make sense to me

tomuta: Right, but if you don't copy it but instead just reference it using `[&]` you're potentially referencing a non-existing object. I think the code as it is right now probably references `download` captued by the other lambda, so if the other lamda were to be deleted it would point to a non-existing RefPtr. Or maybe even a reference to a local variable on the stack. Not sure how exactly this is defined in the c++ spec. To me, an explicit copy-capture is more explicit and probably less likely to break. Yes, it adds an additional reference for the lambda holding onto it.

alimpfard: There's no way the containing lambda would be deleted though, it's stuck in a reference cycle between itself and the download instance that contains it :thinking:

Either way is too damn convoluted though

tomuta: True. Yeah that's a lot of assumptions for a reference capture. Personally, I only like reference captures when I use them like functors.

alimpfard: Well, this change isn't any worse than what it replaces, so :+1: for that.
And yeah, it's much more comfortable to not have to think about something's lifetime when you capture it by reference, less room for weird action at a distance, so to speak

PR #4696, opened 2020-12-31
WindowServer+MouseSettings: Add a configurable primary mouse button setting

IdanHo:
primary
This PR adds a configurable (via MouseSettings) primary mouse button setting in WindowServer which will eventually allow
us to accommodate left handed serenity os users, but considerable work has to be done in the various user-facing parts of serenity to switch from hardcoded is MouseButton::Left checks to is_primary checks (Should i start work on that in this PR or another one?).

Note: i feel like some parts of the code in this one are a bit awkward (especially in the UI side), so i'd be happy to hear code quality improvement suggestions :)

### DIFF
--- a/Applications/FontEditor/GlyphEditorWidget.cpp
+++ b/Applications/FontEditor/GlyphEditorWidget.cpp
@@ -89,14 +89,14 @@ void GlyphEditorWidget::mousedown_event(GUI::MouseEvent& event)
 
 void GlyphEditorWidget::mousemove_event(GUI::MouseEvent& event)
 {
-    if (event.buttons() & (GUI::MouseButton::Left | GUI::MouseButton::Right))
+    if (event.buttons() & (GUI::MouseButton::Primary | GUI::MouseButton::Secondary))
         draw_at_mouse(event);
 }
 
 void GlyphEditorWidget::draw_at_mouse(const GUI::MouseEvent& event)
 {
-    bool set = event.buttons() & GUI::MouseButton::Left;
-    bool unset = event.buttons() & GUI::MouseButton::Right;
+    bool set = event.buttons() & GUI::MouseButton::Primary;
+    bool unset = event.buttons() & GUI::MouseButton::Secondary;
     if (!(set ^ unset))
         return;
     int x = (event.x() - 1) / m_scale;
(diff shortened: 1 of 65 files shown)


tomuta:
The only comment I have would be that maybe it'd be nice to convert it over to GML.

awesomekling:
Hmm, when I think more about this, do we really need a separate flag for primary? Maybe we can just rename Left/Right to Primary/Secondary instead, and switch what WindowServer sends based on the setting?

IdanHo:
That's also a valid option, the main problem is things like the mouse demo, since it relies on the actual buttons, instead of just primary/secondary actions (since it displays the actual hardware button pressed as blue)

awesomekling:
The mouse demo can cheat by reversing the buttons depending on the setting ;)

IdanHo:
that works for me :P we can always add a "physical button" field later on if needed, btw, additional mouse buttons should be left untouched right? (So like primary, secondary, middle, back, forward)

awesomekling:
Yeah I think that makes sense :)

IdanHo:
This should be ready for a re-review, the Mouse Demo hack is a bit awkward, but hopefully most other apps wont need similar functionality.

Dexesttp:
The code seems good :) Not a big fan of the `.send_sync` and static_casts directly in MouseSettings and the Mouse demo, but there's instances of that everywhere already, and there's not much to do about it.

Maybe a better abstraction would be to replace the `GetPrimaryMouseButton() => (u32)` / `SetPrimaryMouseButton(u32) => ()` API with a more generic `GetMouseButtonForRawMouseButton(u32) => (u32)` / `SetMouseButtonForRawMouseButton(u32, u32) => ()` API. This would allow a more generic mapping between the different `WindowServer::RawMouseButton` enum values and the `MouseButton` event enum values.

It allows more flexibility - maybe too much, as you could theorically remap the "primary" button to something like the middle click button. But it should be less "set in stone" than just allowing it for the left/right buttons. It will also make the mouse demo code more consistent : you would have to call the method for each physical button you need, and then "remap" _all_ the incoming buttons to their physical equivalent.

IdanHo:
This has become quite stale, so ill close it for now, might return to it in the future

PR #4705, opened 2021-01-01
LibThread/LibPthread: Key destruction and Thread::join

ADKaster:
LibPthread: Destroy pthread_keys on thread/program exit

Currently, keys are destroyed either via global destructors, with the
s_key_destroyer object, or in exit_thread. exit_thread is invoked by
pthread_exit, and transitively by pthread_create, via the
pthread_create_helper that ensures all threads created with the pthread
API properly clean up for themselves when they exit gracefully.

A future patch might make s_key_destroyer a C++11 thread_local instead,
assuming we get thread_local and thread_local destructors working.

LibThread: Make Thread::join return the ret from pthread_join, and take a void* for the thread's return value.

Change the only caller of pthread_join to use the new and improved Thread::join API.

cc @bugaevc

### DIFF
--- a/Libraries/LibThread/Thread.cpp
+++ b/Libraries/LibThread/Thread.cpp
@@ -34,13 +34,15 @@ LibThread::Thread::Thread(Function<int()> action, StringView thread_name)
     , m_action(move(action))
     , m_thread_name(thread_name.is_null() ? "" : thread_name)
 {
+    register_property("thread_name", [&] { return JsonValue { m_thread_name }; });
+    register_property("tid", [&] { return JsonValue { m_tid }; });
 }
 
 LibThread::Thread::~Thread()
 {
     if (m_tid) {
-        dbg() << "trying to destroy a running thread!";
-        join();
+        dbgln("Destroying thread \"{}\"({}) while it is still running!", m_thread_name, m_tid);
+        [[maybe_unused]] auto res = join();
     }
 }
 
@@ -51,7 +53,7 @@ void LibThread::Thread::start()
         nullptr,
         [](void* arg) -> void* {
             Thread* self = static_cast<Thread*>(arg);
-            size_t exit_code = self->m_action();
+            int exit_code = self->m_action();
             self->m_tid = 0;
             return (void*)exit_code;
         },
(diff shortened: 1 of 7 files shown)


Libraries/LibThread/Thread.cpp
```diff
@@ -62,22 +62,13 @@ void LibThread::Thread::start()
...
     }
-    dbg() << "Started a thread, tid = " << m_tid;
+    dbgln("Started thread \"{}\", tid = {}", m_thread_name, m_tid);
 }
 
-void LibThread::Thread::join()
+int LibThread::Thread::join(void* thread_return)
 {
-    int rc = pthread_join(m_tid, nullptr);
+    int rc = pthread_join(m_tid, thread_return ? &thread_return : nullptr);
```
tomuta, line 70: This looks wrong. You probably want `void** thread_return` as the `Thread::join` argument and pass it through.

tomuta (changes requested):

Libraries/LibThread/Thread.cpp
ADKaster: Oh duh yeah, otherwise it just writes it to the local varaible.

ADKaster: Nice catch!

bugaevc:
Perhaps we could have something more flexible as `.join()`s return type than a plain `int`? Like make it a `Thread<T>` and make the body and `.join()` return `T`? Or perhaps we don't need that.

ADKaster:
> Perhaps we could have something more flexible as `.join()`s return type than a plain `int`? Like make it a `Thread<T>` and make the body and `.join()` return `T`? Or perhaps we don't need that.

Well, currently the int is the return value from pthread_join, which is either 0, or an errno value from the underlying syscall.

So, could do some kind of `Expected<T, system_error>` thing with it (and other syscall wrapping classes...) but as of right now there's only one caller of join() outside of the Thread destructor, and that caller doesn't even look at the return value it took the time to allocate stack memory for. 

I'm not sure Thread<T> would be the right thing, not sure most users would be creating a thread for the purpose of getting the return value out . templated join sounds like a reasonable idea for some use cases? Maybe an overload.

tomuta:
I think it'd be really nice if `.join()` would return the return value of an arbitrary thread function, including `void`.

PR #4707, opened 2021-01-01
Memory purging improvements

tomuta:
This is the continuation of some of the work started in #3396, brought up to date.


smopucilowski (commented):
512MB or 1024MB?

tomuta:
> 512MB or 1024MB?

Forgot to update the commit message. I think 512 is OK for now.

@awesomekling So I still haven't figured out the odd crash when `execve` has to rollback the regions, but other than that I think it's working well. Maybe we can merge it anyway as it's a corner case, but eventually we'll obviously have to figure it out. Right now memory purging and low memory still causes crashes in user mode mainly because ProcFS doesn't handle it well. The original PR has some patches that improve/fix this, but I think I'll pull those separately from this PR. I think this itself is a good step forward. Not sure if 512 MB ram is a good choice, maybe 768 MB or 1 GB? We can drop it once we support swapping (the other PR has some work in that area, too).

tomuta:
Hm seems like noticably slower with KVM disabled :(

awesomekling:
Perhaps we can reorganize `sys$execve` so that it never needs to perform a rollback? Split it into a couple of passes, and once we commit to going forward, no rollback is needed.. very loosely:
1. Open ELF and parse headers
2. Reserve all the physical memory required
3. (this is where we commit)
4. Create the new page tables & update everything else in the process
5. Context switch

tomuta:
I suppose we could, but I don't really understand why the rollback is causing that weird assertion to be triggered. I don't see any reason why the rollback shouldn't work.

tomuta:
> Hm seems like noticably slower with KVM disabled :(

Switching the `KBuffer` allocation strategy from `AllocationStrategy::AllocateNow` to `AllocationStrategy::Reserve` looks like improves this a lot. But there might be more room for improvements.

PR #4708, opened 2021-01-01
Everywhere: Type-safer Widget GML

ADKaster:
This patch replaces `GUI::Widget::find_descendant_by_name` with `Core::Object::find_descendant_of_type_named<T>`.

It also leverages AK::Traits::is in order to do so. This requires a few changes:

- Add the finder helper to Core::Object
- Add a helper "default traits" macro to check if a Core::Object is-a T. Are we *sure* dynamic cast is bad? ;) 
- Spray the default trait macro all over the codebase
- ???
- Profit

I'm not sold on the name of the Core::Object helper, so lemme know if there's a better one. Everyone loves a good naming thread.

### DIFF
--- a/Applications/Browser/Tab.cpp
+++ b/Applications/Browser/Tab.cpp
@@ -90,10 +90,10 @@ Tab::Tab(Type type)
 {
     load_from_gml(tab_gml);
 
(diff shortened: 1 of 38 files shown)


ADKaster:
Superseded by #4714, which does the same change with compiler-generated RTTI instead.

PR #4711, opened 2021-01-01
Base: Add Coffee theme

bcoles:
Coffee


awesomekling:
Strong JDS vibes ☕

bcoles:
> 
> 
> Strong JDS vibes coffee

Purely coincidental >_>

bcoles:
I've been using (and tweaking) this theme for a couple of days. I'm mostly happy with it, but there's a couple of things I don't like, due to either shortcomings in the existing theme implementation or a fundamental misunderstanding on my part.

The most egregious is the window icons. Changing the `Button` color to control the window icon background color also sets the color for all other buttons. As a result, I've had to set the window icon images with a solid background color rather than full transparency. This results in two misaligned grey pixels during mouse down when clicking a window button icon.

There's also no way to set different custom window icons for active windows versus inactive windows. Upon close inspection of inactive windows there's a noticeable discrepancy between the window icon background color and the window title bar color.

Similarly, there's no easy way to control the window border without modifying `ThreedShadow1` and `ThreedShadow2` which effects all objects (ie, menus and the taskbar) not only application "windows". For this reason, I used grey borders but ideally they would have been purple.

PR #4712, opened 2021-01-01
LibTTF+LibGfx: Add system-wide truetype support

sunverwerth:
Add truetype support to font picker and gfx drawing routines. The system should ideally be totally oblivious to the fact that there are different types of fonts, except for special cases such as the FontEditor.

- [x] Parse name table from TTF files
- [x] Insert fonts into a more elaborate structure in FontDatabase
- [x] Make FontPicker use the new structure
- [x] Improve LibGfx to allow to render any Font implementation
- [x] Bonus: Create a SerenitySans.ttf :^)

~~Improve rasterizing~~


awesomekling:
>SerenitySans.ttf

That's the spirit!

sunverwerth:
Here's a sneaky:

image

:laughing:

tomuta:
Hey this font actually looks really cool!

sunverwerth:
TTF in terminal :partying_face: 

image

sunverwerth:
Ok, I think this is ready to be merged in.
Obviously a lot of work is still to be done. That just means more opportunities for you to earn some **Serenity points**:tm: 

The SerenitySans is just a quick scribble to get something into the repo to test with. If anyone feels artistically inclined, feel free to turn this into something really beautiful (@linusg ?). I made this in FontForge and can provide the project file. Maybe we can add this to the SerenityOS Github project.

awesomekling:
This is great progress but it looks a bit unhygienic. Please make an effort to fixup commits instead of stacking many sloppy commits and then doing a big "lint fixing" commit in the end. :)

I rather liked the OpenType-style numeric weights, can we have that instead of string-based "variants"?

sunverwerth:
> This is great progress but it looks a bit unhygienic. Please make an effort to fixup commits instead of stacking many sloppy commits and then doing a big "lint fixing" commit in the end. :)

Ok I will try my best. I hope my git fu good enough.


> I rather liked the OpenType-style numeric weights, can we have that instead of string-based "variants"?

I don't think that's enough info to identify the fonts. There can be different variants with the same weight, e.g Bold and BoldItalic

awesomekling:
> > This is great progress but it looks a bit unhygienic. Please make an effort to fixup commits instead of stacking many sloppy commits and then doing a big "lint fixing" commit in the end. :)
> 
> Ok I will try my best. I hope my git fu good enough.

Thank you! 🐱

> > I rather liked the OpenType-style numeric weights, can we have that instead of string-based "variants"?
> 
> I don't think that's enough info to identify the fonts. There can be different variants with the same weight, e.g Bold and BoldItalic

I feel like we're conflating two concepts here: style and weight. Italic is a font style, while medium/regular/bold/extra-bold/etc are font weights.

sunverwerth:
> I feel like we're conflating two concepts here: style and weight. Italic is a font style, while medium/regular/bold/extra-bold/etc are font weights.

Can you clarify what you meant wth the original comment? Maybe I just misunderstood.

awesomekling:
> > I feel like we're conflating two concepts here: style and weight. Italic is a font style, while medium/regular/bold/extra-bold/etc are font weights.
> 
> Can you clarify what you meant wth the original comment? Maybe I just misunderstood.

This PR removes the ability to select a weight and mixes it together with style. I think style and weight should be two separate concepts.

sunverwerth:
I'm not sure if this information is available in the ttf format. I'll need to take another look at it. Maybe it's available int the post table. Otherwise we need to extract it from the subfamily string somehow. I modeled it after the windows font picker where these styles are also compounded:

grafik

awesomekling:
If I'm reading this correctly, it's the `wght` tag in the `fvar` table:
https://docs.microsoft.com/en-us/typography/opentype/spec/fvar

CSS uses font weights, so we definitely need it there :)

awesomekling:
Or perhaps it's `usWeightClass` in the "OS/2 table"?
https://docs.microsoft.com/en-us/typography/opentype/spec/os2

sunverwerth:
Thanks, I'll have a look at this. The weight information is definitely still available in the font class. It's just not used in the FontPicker to display the different variants.

danboid:
What happened to this @awesomekling ?

sunverwerth:
Hey @danboid I've been pretty busy the last few weeks. I'll try to make time this weekend and finish the PR.

danboid:
Thanks for the update! I thought you were waiting on having your most recent commits reviewed.

sunverwerth:
@awesomekling I rewrote most of the commit history and applied linter fixes right where they popped up. The FontPicker is now back to the old style with family->weight->size selection. What doesn't work now is having multiple ttf font files of the same family since they are grouped under one font family in the picker. This needs to be fixed by parsing the relevant OpenType tables and extracting that information from there and then extending the font picker to allow more selection categories like italic, underlined, etc...

Also, strings in OpenType come in a wide variety of encodings. Currently ASCII is assumed although most fonts that I tested use UCS2 and thus show spaces between the characters. Still works, just looks funny in the FontPicker.

awesomekling:
Ok, kick-ass! Let's merge it :)

PR #4714, opened 2021-01-01
Everywhere: Type Safe Widget GML usage, with RTTI

ADKaster:
This patch replaces GUI::Widget::find_descendant_by_name with Core::Object::find_descendant_of_type_named<T>.

Elephant in the room, this patch enables RTTI for userspace programs. Some quick measurements showed that this is a small binary size increase, on the order of 3% for a class-rich library like `libgui.so`. 

This PR is an alternative to #4708, which does the AK::Traits thing to do the RTTI manually. Which, is kind of awkward and really annoying to have to copy paste the "opt-me-in to RTTI" macro into every header under the sun.

I'm not sold on the name of the Core::Object helper, so lemme know if there's a better one. Everyone loves a good naming thread.

### DIFF
--- a/Applications/Browser/Tab.cpp
+++ b/Applications/Browser/Tab.cpp
@@ -90,10 +90,10 @@ Tab::Tab(Type type)
 {
     load_from_gml(tab_gml);
 
-    m_toolbar_container = static_cast<GUI::ToolBarContainer&>(*find_descendant_by_name("toolbar_container"));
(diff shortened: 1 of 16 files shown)


awesomekling:
I like it. If RTTI proves to be a measurable problem, we can reconsider it then.
For now, the prospects for cleanup offered by this change is very attractive :)

PR #4717, opened 2021-01-01
LibGUI: Add an optional "automatic" autocomplete feature to TextEditor

alimpfard:
"automatic" autocomplete, where the editor shows suggestions unprompted (but tries to avoid being too obtrusive).

I briefly though about making only `<tab>` apply the highlighted suggestion in this mode, but it seemed weird (same for only `<return>` doing so)

Perhaps this should be the default?

### DIFF
--- a/Libraries/LibGUI/TextEditor.cpp
+++ b/Libraries/LibGUI/TextEditor.cpp
@@ -713,6 +713,7 @@ void TextEditor::sort_selected_lines()
 
 void TextEditor::keydown_event(KeyEvent& event)
 {
+    TemporaryChange change { m_should_keep_autocomplete_box, true };
     if (m_autocomplete_box && m_autocomplete_box->is_visible() && (event.key() == KeyCode::Key_Return || event.key() == KeyCode::Key_Tab)) {
         m_autocomplete_box->apply_suggestion();
         m_autocomplete_box->close();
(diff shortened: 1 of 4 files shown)


Libraries/LibGUI/TextEditor.cpp
```diff
@@ -1848,4 +1874,24 @@ void TextEditor::set_visualize_trailing_whitespace(bool enabled)
...
+        return;
+
+    if (value) {
+        ASSERT(m_autocomplete_provider);
+        m_autocomplete_timer = add<Core::Timer>(m_automatic_autocomplete_delay_ms, nullptr);
+        m_autocomplete_timer->set_single_shot(true);
+        m_autocomplete_timer->stop();
+        m_autocomplete_timer->on_timeout = [this] {
+            try_show_autocomplete();
+        };
```
awesomekling, line 1889:
```suggestion
    m_autocomplete_timer = Core::Timer::create_single_shot(m_automatic_autocomplete_delay_ms, [this] { try_show_autocomplete(); });
```

tomuta:
Slightly off-topic, but a LibSpell would be really cool. Would be nice for Browser, Spreadsheet, IRC client, and maybe even TextEditor...

Libraries/LibGUI/TextEditor.cpp
alimpfard: That changed so incrementally I didn't even think of using that factory function...oops

alimpfard:
> LibSpell

Where would we get the dictionary for that though?
We could generate it given some corpus, or just find one and grab it as a "port"?

It _would_ be nice, with this PR we'd even have autocorrect suggestions too

PR #4718, opened 2021-01-01
DisplaySettings: Remove `root_widget()` from DisplaySettings

Quaker762:
This fixes a weird dependency graph in DisplaySettings. The widget itself
(which is described in `gml` now), no longer contains `root_widget()`.

The widget itself has been moved into a tabbed pane, to get it ready
to add some more features and bring it more up to date with the current
UI code.

### DIFF
--- a/Applications/DisplaySettings/DisplaySettings.cpp
+++ b/Applications/DisplaySettings/DisplaySettings.cpp
@@ -90,12 +90,11 @@ void DisplaySettingsWidget::create_wallpaper_list()
 
 void DisplaySettingsWidget::create_frame()
 {
-    m_root_widget = GUI::Widget::construct();
-    m_root_widget->load_from_gml(display_settings_window_gml);
+    load_from_gml(display_settings_window_gml);
 
-    m_monitor_widget = *m_root_widget->find_descendant_of_type_named<DisplaySettings::MonitorWidget>("monitor_widget");
+    m_monitor_widget = *find_descendant_of_type_named<DisplaySettings::MonitorWidget>("monitor_widget");
 
-    m_wallpaper_combo = *m_root_widget->find_descendant_of_type_named<GUI::ComboBox>("wallpaper_combo");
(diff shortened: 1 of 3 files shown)


Applications/DisplaySettings/main.cpp
```diff
@@ -53,15 +54,17 @@ int main(int argc, char** argv)
...
+    tab_widget->set_fill_with_background_color(true); // No black backgrounds!
 
     auto window = GUI::Window::construct();
     dbgln("main window: {}", window);
     window->set_title("Display settings");
-    window->resize(360, 390);
+    window->resize(360, 410);
     window->set_resizable(false);
-    window->set_main_widget(instance->root_widget());
+    window->set_main_widget(reinterpret_cast<GUI::Widget*>(tab_widget.ptr()));
```
awesomekling, line 67: There should be no need for a cast here.

Applications/DisplaySettings/DisplaySettings.cpp
```diff
@@ -90,12 +90,11 @@ void DisplaySettingsWidget::create_wallpaper_list()
 
 void DisplaySettingsWidget::create_frame()
 {
-    m_root_widget = GUI::Widget::construct();
-    m_root_widget->load_from_gml(display_settings_window_gml);
+    load_from_gml(display_settings_window_gml);
 
-    m_monitor_widget = *m_root_widget->find_descendant_of_type_named<DisplaySettings::MonitorWidget>("monitor_widget");
+    m_monitor_widget = static_cast<DisplaySettings::MonitorWidget&>(*find_descendant_of_type_named<DisplaySettings::MonitorWidget>("monitor_widget"));
```
awesomekling, line 95: You don't need to cast the return values here, that's the whole point of passing the type as a template argument to find_descendant_of_type_named() :)

Quaker762: I see. So this is effectively `dynamic_cast<>()` right (if I was followign the IRC conversation correctly)

PR #4719, opened 2021-01-01
Games: Add Conway

bcoles:
Conway


Games/Conway/Game.cpp
```diff
@@ -0,0 +1,114 @@
...
+    painter.add_clip_rect(event.rect());
+    painter.fill_rect(event.rect(), m_dead_color);
+    auto game_rect = rect();
+    auto cell_size = Gfx::IntSize(game_rect.width() / m_columns, game_rect.height() / m_rows);
+
+    for (int y = 0; y < m_rows; y++) {
+        for (int x = 0; x < m_columns; x++) {
+            Gfx::IntRect rect {
+                y * cell_size.width(),
+                x * cell_size.height(),
```
linusg, line 107: Are x and y swapped on purpose here?

Games/Conway/Game.cpp
```diff
@@ -0,0 +1,114 @@
...
+            if (n == 3 || (n == 2 && cell))
+                new_universe[y][x] = true;
+            else
+                new_universe[y][x] = false;
+        }
+    }
+
+    for (int y = 0; y < m_rows; y++)
+        for (int x = 0; x < m_columns; x++)
+            m_universe[y][x] = new_universe[y][x];
```
emanuele6, line 86:
```suggestion
    for (int y = 0; y < m_rows; y++) {
        for (int x = 0; x < m_columns; x++)
            m_universe[y][x] = new_universe[y][x];
    }
```

Games/Conway/Game.cpp
```diff
@@ -0,0 +1,114 @@
...
+    int n;
+    for (int y = 0; y < m_rows; y++) {
+        for (int x = 0; x < m_columns; x++) {
+            n = 0;
+            int cell = m_universe[y][x];
+
+            for (int y1 = y - 1; y1 <= y + 1; y1++)
+                for (int x1 = x - 1; x1 <= x + 1; x1++)
+                    if (m_universe[(y1 + m_rows) % m_rows][(x1 + m_columns) % m_columns])
+                        n++;
```
emanuele6, line 72:
```suggestion
            for (int y1 = y - 1; y1 <= y + 1; y1++) {
                for (int x1 = x - 1; x1 <= x + 1; x1++) {
                    if (m_universe[(y1 + m_rows) % m_rows][(x1 + m_columns) % m_columns])
                        n++;
                }
            }
```

Games/Conway/Game.cpp
```diff
@@ -0,0 +1,114 @@
...
+    seed_universe();
+    start_timer(m_sleep);
+    update();
+}
+
+void Game::seed_universe()
+{
+    for (int y = 0; y < m_rows; y++)
+        for (int x = 0; x < m_columns; x++)
+            m_universe[y][x] = (arc4random() % 2) ? 1 : 0;
```
emanuele6, line 56:
```suggestion
    for (int y = 0; y < m_rows; y++) {
        for (int x = 0; x < m_columns; x++)
            m_universe[y][x] = (arc4random() % 2) ? 1 : 0;
    }
```

emanuele6 (changes requested):
We usually prefer not omitting braces in these cases.

Games/Conway/Game.cpp
```diff
@@ -0,0 +1,114 @@
...
+                    if (m_universe[(y1 + m_rows) % m_rows][(x1 + m_columns) % m_columns])
+                        n++;
+
+            if (cell)
+                n--;
+
+            if (n == 3 || (n == 2 && cell))
+                new_universe[y][x] = true;
+            else
+                new_universe[y][x] = false;
```
emanuele6, line 80: `cell` should be `cell != 0` to more explicit.
 
Also, can't this just be:
```suggestion
            new_universe[y][x] = n == 3 || (n == 2 && cell != 0);
```

bcoles: > 
> 
> I think you should use `cell != 0` instead of `cell` here to be more explicit. :)

why

> Also, can't this just be: [...]

It could, if you like mashing together declarations and conditionals into a single line with 6 equals signs and multiple parenthesis. Seems like a terrible decision for no benefit.

Games/Conway/Game.cpp
bcoles: No, no they weren't

Games/Conway/Game.cpp
```diff
@@ -0,0 +1,121 @@
...
+                    if (m_universe[(y1 + m_rows) % m_rows][(x1 + m_columns) % m_columns]) {
+                        n++;
+                    }
+                }
+            }
+
+            if (cell == true)
+                n--;
+
+            if (n == 3 || (n == 2 && cell == true))
```
emanuele6, line 82: `cell` is an `int`. This is the reason I suggested writing `cell != 0`; to avoid thinking it is a `bool`.

Games/Conway/Game.cpp
```diff
@@ -0,0 +1,121 @@
...
+
+void Game::update_universe()
+{
+    bool new_universe[m_rows][m_columns];
+
+    int n;
+    for (int y = 0; y < m_rows; y++) {
+        for (int x = 0; x < m_columns; x++) {
+            n = 0;
+            int cell = m_universe[y][x];
```
emanuele6, line 69: Oh, oops. `cell` should have been a `bool` here I think, I got confused. It's fine to just leave `cell` in the expressions then, no need to use `cell == true`-

But you should declare `cell` as a `bool`, not an `int`. also maybe use `auto` (?)

bcoles: Yeah it a `bool` from an array of `bool`, formerly `int`. I'll change that line to use `auto`.

Games/Conway/Game.cpp
emanuele6: I mean, it's just one one group of parenthesis.
```cpp
new_universe[y][x] = n == 3 || (cell && n == 2)
```
I just don't like using this personally:
```cpp
bool var;
if (EXPRESSION)
    var = true;
else
    var = false;
```
since `var` it is an indirect way of writing:
```cpp
bool var = EXPRESSION;
```
But no big deal if you prefer writing it that way. This was just a comment.

Games/Conway/Game.cpp
```diff
@@ -0,0 +1,119 @@
...
+}
+
+void Game::update_universe()
+{
+    bool new_universe[m_rows][m_columns];
+
+    int n;
+    for (int y = 0; y < m_rows; y++) {
+        for (int x = 0; x < m_columns; x++) {
+            n = 0;
```
emanuele6, line 66: `n` is not used outside of the inner loop. In my opinion, to make that clear, you should write this as:
```suggestion
    for (int y = 0; y < m_rows; y++) {
        for (int x = 0; x < m_columns; x++) {
            int n = 0;
```

awesomekling:
Mmm desktop games!

PR #4721, opened 2021-01-01
Lagom/Fuzzers: Add TTF fuzzer

Lubrsi:


Libraries/LibTTF/Font.cpp
```diff
@@ -208,6 +208,11 @@ RefPtr<Font> Font::load_from_file(const StringView& path, unsigned index)
         return nullptr;
     }
     auto buffer = file->read_all();
+    return load_from_memory(buffer, index);
+}
+
+RefPtr<Font> Font::load_from_memory(ByteBuffer& buffer, unsigned index)
```
awesomekling, line 213: Let's make this take a `ReadonlyBytes` instead of a `ByteBuffer&` :)

Libraries/LibTTF/Font.h
```diff
@@ -100,7 +100,7 @@ class Font : public RefCounted<Font> {
     }
 
     // This owns the font data
-    ByteBuffer m_buffer;
+    ReadonlyBytes m_buffer;
```
awesomekling, line 104: Oh wait back up. `ReadonlyBytes` is non-owning, so this doesn't seem safe.
I didn't realize that `Font` would take ownership of the `ByteBuffer`.

awesomekling:
Great initiative! Sorry about the flip-flop :)

PR #4722, opened 2021-01-01
AK: Deal with unsigned integers in binary search.

asynts:
I am really surprised that Serenity managed to boot with this. What is happening here is that if `lhs - rhs` becomes negative, it is interpreted as unsigned and then the binary search goes in the wrong direction.

I trust that the compiler will be able to optimize the 1 / -1 / 0 out so this should only make a difference for unsigned integers which were handled incorrectly before.

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -85,6 +85,9 @@ jobs:
     - name: Run JS tests
       working-directory: ${{ github.workspace }}/Build/Meta/Lagom
       run: DISABLE_DBG_OUTPUT=1 ./test-js
+    - name: Run LibCompress tests
+      working-directory: ${{ github.workspace }}/Build/Meta/Lagom
+      run: DISABLE_DBG_OUTPUT=1 ./test-compress
 
      # Run analysis last, so contributors get lint/test feedback ASAP.
     - name: Perform post build CodeQL Analysis
--- a/AK/BinarySearch.h
+++ b/AK/BinarySearch.h
@@ -32,11 +32,21 @@
 
 namespace AK {
 
-struct IntegralComparator {
-    constexpr auto operator()(auto& lhs, auto& rhs) { return lhs - rhs; }
+struct DefaultComparator {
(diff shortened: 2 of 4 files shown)


asynts:
I also added `test-compress` to the GitHub workflow, I noticed this bug because the `test-compress` would go into an infinite loop. Running that test would have prevented me from breaking this in #4633.

awesomekling:
Oh dang, this broke CI :(
```
/home/runner/work/_temp/5c6254e5-7042-454d-9263-807f47e148e3.sh: line 1: ./test-compress: No such file or directory
```

asynts:
~~~diff
diff --git a/Meta/Lagom/CMakeLists.txt b/Meta/Lagom/CMakeLists.txt
index 452d83c3c..dc51fda0f 100644
--- a/Meta/Lagom/CMakeLists.txt
+++ b/Meta/Lagom/CMakeLists.txt
@@ -121,6 +121,11 @@ if (BUILD_LAGOM)
             WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
         )
 
+        add_executable(test-compress_lagom ../../Userland/test-compress.cpp)
+        set_target_properties(test-compress_lagom PROPERTIES OUTPUT_NAME test-compress)
+        target_link_libraries(test-compress_lagom Lagom)
+        target_link_libraries(test-compress_lagom stdc++)
+
         add_executable(disasm_lagom ../../Userland/disasm.cpp)
         set_target_properties(disasm_lagom PROPERTIES OUTPUT_NAME disasm)
         target_link_libraries(disasm_lagom Lagom)
~~~

asynts:
You merged that rather quickly, I wasn't sure where to add the call to `test-compress` so I wanted to let the CI run through...

awesomekling:
Yeah that's my bad, sorry about that!

asynts:
Should I open a pull request or do you commit that?

awesomekling:
I just pushed it, thank you!

PR #4723, opened 2021-01-01
Conway: Distribute leftover space equally on all sides

linusg:
This draws the universe centered rather than just putting all leftover space on the right and bottom sides until the window is large enough to be completely filled with cells again.

So, this:

Peek 2021-01-01 23-14

Rather than:

image


asynts:
How about using `set_resize_aspect_ratio(1, 1)` instead? We use that for the Chess game already:

https://github.com/SerenityOS/serenity/blob/cf2c215defd448ca44523143827e3ba022f8dd71/Games/Chess/main.cpp#L90

linusg:
I know, but no ;)
The universe can be drawn stretched with cells having a different ratio than 1:1, and unlike in the chess game I think that's fine (reminds me of terminal implementations of conway's game of life, where cells are also stretched, obviously).

image

asynts:
Fair point.

PR #4726, opened 2021-01-01
Build + LibC: Enable -fstack-protector-strong in user space

bgianfo:
This PR modifies the user mode runtime to Insert stack canaries to find stack corruptions.
It looks like this was enabled in the past (842716a0b5eceb8db31416cd643720c1037032b2) but was lost during the CMake conversion. 

The `-fstack-protector-strong` variant was chosen because it catches more issues than `-fstack-protector`,
but doesn't have substantial performance impact like `-fstack-protector-all`.

Details:

    -fstack-protector enables stack protection for vulnerable functions that contain:

    * A character array larger than 8 bytes.
    * An 8-bit integer array larger than 8 bytes.
    * A call to alloca() with either a variable size or a constant size bigger than 8 bytes.

    -fstack-protector-strong enables stack protection for vulnerable functions that contain:

    * An array of any size and type.
    * A call to alloca().
    * A local variable that has its address taken.

Example of it catching corrupting in the `stack-smash` test:
```
courage ~ $ ./user/Tests/LibC/stack-smash
[+] Starting the stack smash ...
Error: Stack protector failure, stack smashing detected!
Shell: Job 1 (/usr/Tests/LibC/stack-smash) Aborted
```

### DIFF
--- a/Libraries/LibC/ssp.cpp
+++ b/Libraries/LibC/ssp.cpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2020, the SerenityOS developers.
+ * Copyright (c) 2021, Brian Gianforcaro <b.gianfo@gmail.com>
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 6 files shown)


Libraries/LibC/ssp.cpp
```diff
@@ -0,0 +1,47 @@
...
+
+#if defined __SSP__ || defined __SSP_ALL__
+#    error "file must not be compiled with stack protection enabled on it. Use -fno-stack-protector"
+#endif
+
+extern "C" {
+
+__attribute__((noreturn)) void __stack_chk_fail_local(void);
+
+void __stack_chk_fail_local(void)
```
ADKaster, line 40: Would it make sense to move __stack_chk_fail and __stack_chk_guard from cxxabi.cpp into this file?

bgianfo: I can try, the __stack_chk_fail_local symbol is kind of funky, as it needs to be injected into every application and gcc forces it to be a hidden symbol by default. (hence having to add ssp static library in CMake goo)

bgianfo: That worked, thanks for the suggestion, seems like a good change to me.

Libraries/LibC/ssp.cpp
```diff
@@ -0,0 +1,55 @@
...
+#include <unistd.h>
+
+#if defined __SSP__ || defined __SSP_ALL__
+#    error "file must not be compiled with stack protection enabled on it. Use -fno-stack-protector"
+#endif
+
+extern "C" {
+
+extern u32 __stack_chk_guard;
+u32 __stack_chk_guard = (u32)0xc6c7c8c9;
```
awesomekling, line 40: Next step: randomizing the canary value :)

bgianfo: yup, planning on it!

ADKaster: 16 good random bytes from the kernel should already be available in the ELF Auxiliary Vector for seeding all sorts of fun things :)

PR #4728, opened 2021-01-01
Kernel+LibELF: Use hex instead of decimal for stack offsets in back traces

bgianfo:
Hex is the de facto format for representing memory addresses, make backtraces conform to that convention.


awesomekling:
Yeah all right

PR #4732, opened 2021-01-02
LibELF+Loader.so: Support relocating weak symbols against global libraries

### LINKED ISSUE #4627: DynamicLoader: Weak symbols are not resolved correctly
The DynamicLoader currently does not support weak symbols correctly. At link time a weak symbol can and should be overridden by a strong symbol of the same name even if it is in a different library (and not even necessarily one of the current libraries dependencies). If there are two strong symbols with the same name then the one that was encountered first should be preferred. This allows applications to override library functions such as `malloc`. 

Here is a test that can show the issue. Consider a library with the following two headers:

**lib.h**
```
#pragma once

extern "C" {
    int __attribute__((visibility("default"))) print_icon();
}
```

**icons.h**
```
#pragma once


extern "C" {
    extern const char* __attribute__((weak, section("serenity_icon_s"))) serenity_icon_s_start;
    extern unsigned long __attribute__((weak, section("serenity_embedded_resource_info"))) serenity_icon_s_size;
}
```

and the following two source files:

**lib.cpp**
```
#include "icons.h"
#include "lib.h"

#include <cstdio>

int print_icon() {
	if (serenity_icon_s_start == nullptr) {
		printf("FAIL: serenity_icon_s_start is null\n");
		return 1;
	} else {
		printf("PASS: serenity_icon_s_start is not null and serenity_icon_s_size is %zu\n", serenity_icon_s_size);
		return 0;
	}
}
```

**default_icon.cpp**
```
#include "icons.h"

const char* serenity_icon_s_start = nullptr;
unsigned long serenity_icon_s_size = 0;
```

This library is consumed by the following executable
**main.cpp**
```
#include "lib.h"

int main([[maybe_unused]] int argc, [[maybe_unused]] char** argv) {
	return print_icon();
}
```

If this is compiled without embedding an application icon, e.g. using the `embed_resource()` CMake function then it will correctly print `FAIL: serenity_icon_s_start is null` on both Linux and Serenity. If the icon is embedded correctly then Linux will pass but Serenity will fail:

image

image

lux01:
This PR improves Loader.so so that it supports linking weak symbols correctly.

A strong symbol anywhere in an executable must override any weak symbol used in any library. This means that the weak symbol must be overridden by a strong symbol even if the strong symbol is in a dependent library. This means we need to perform relocations twice, and resolve weak symbols globally before attempting to resolve them locally. Consequentially we need to defer performing any initialisations until after we have performed the second round of relocations.

Fixes #4627.

image

### DIFF
--- a/Libraries/LibELF/DynamicLinker.cpp
+++ b/Libraries/LibELF/DynamicLinker.cpp
@@ -207,9 +207,25 @@ static void load_elf(const String& name)
     g_global_objects.append(*dynamic_object);
 
     VERBOSE("load_elf: done %s\n", name.characters());
-    if (name == "libc.so") {
-        initialize_libc(*dynamic_object);
+}
+
+static NonnullRefPtr<DynamicLoader> commit_elf(const String& name)
+{
+    auto loader = g_loaders.get(name).value();
+    for (const auto& needed_name : get_dependencies(name)) {
+        String library_name = get_library_name(needed_name);
+        if (g_loaders.contains(library_name)) {
+            commit_elf(library_name);
+        }
     }
+
+    auto object = loader->load_stage_3(RTLD_GLOBAL | RTLD_LAZY, g_total_tls_size);
+    ASSERT(object);
+    if (name == "libc.so") {
+        initialize_libc(*object);
+    }
+    g_loaders.remove(name);
+    return loader;
 }
 
 void ELF::DynamicLinker::linker_main(String&& main_program_name, int main_program_fd, int argc, char** argv, char** envp)
@@ -226,7 +242,7 @@ void ELF::DynamicLinker::linker_main(String&& main_program_name, int main_progra
     allocate_tls();
 
     load_elf(main_program_name);
-    auto main_program_lib = g_loaders.get(main_program_name).value();
+    auto main_program_lib = commit_elf(main_program_name);
 
     FlatPtr entry_point = reinterpret_cast<FlatPtr>(main_program_lib->image().entry().as_ptr());
     if (main_program_lib->is_dynamic())
--- a/Libraries/LibELF/DynamicLinker.cpp
+++ b/Libraries/LibELF/DynamicLinker.cpp
@@ -1,0 +1,250 @@
+/*
+ * Copyright (c) 2020, Itamar S. <itamar8910@gmail.com>
+ * Copyright (c) 2021, the SerenityOS developers.
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ *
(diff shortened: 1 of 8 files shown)


Libraries/LibELF/DynamicLinker.cpp
```diff
@@ -0,0 +1,266 @@
...
+{
+    auto loader = g_loaders.get(name).value();
+    for (const auto& needed_name : get_dependencies(name)) {
+        String library_name = get_library_name(needed_name);
+        if (g_loaders.contains(library_name)) {
+            commit_elf(library_name);
+        }
+    }
+
+    auto object = loader->load_stage_3(RTLD_GLOBAL | RTLD_LAZY, g_total_tls_size);
```
itamar8910, line 223: `load_stage_3` may return `nullptr` to indicate failure, let's at least assert here that it succeeded.

Libraries/LibELF/DynamicLoader.h
```diff
@@ -117,11 +114,15 @@ class DynamicLoader : public RefCounted<DynamicLoader> {
...
-    void do_relocations(size_t total_tls_size);
+    void do_main_relocations(size_t total_tls_size);
+
+    // Stage 3
+    void do_lazy_relocations(size_t total_tls_size);
     void setup_plt_trampoline();
     void call_object_init_functions();
 
     bool validate();
+    void do_relocation(size_t total_tls_size, DynamicObject::Relocation relocation, Vector<DynamicObject::Relocation>& lazy_relocation_queue);
```
itamar8910, line 125: Passing a vector of unresolved relocations feels inconvenient.
Maybe we could instead have `do_relocation` return an Enum type that reports whether the we succeeded to perform the relocation / failed / failed on a weak symbol ?
Then callers to the function could add the symbol to the lazy relocation list if they wish to do so.

Libraries/LibELF/DynamicLoader.h
```diff
@@ -144,6 +145,8 @@ class DynamicLoader : public RefCounted<DynamicLoader> {
 
     size_t m_tls_offset { 0 };
     size_t m_tls_size { 0 };
+
+    Vector<DynamicObject::Relocation> m_lazy_relocations;
```
itamar8910, line 149: Maybe a nitpick, but i think that this name could be confusing readers with lazy binding in PLT.
Perhaps `m_unresolved_relocations` or `m_late_relocations` would be better names?

itamar8910 (commented):
Looks good, left a few comments.
Also, thanks for putting things in `DynamicLinker.cpp`, good idea.

awesomekling:
```
/home/runner/work/serenity/serenity/Libraries/LibELF/DynamicLoader.cpp:247:21: error: loop variable 'relocation' of type 'const ELF::DynamicObject::Relocation' creates a copy from type 'const ELF::DynamicObject::Relocation' [-Werror,-Wrange-loop-construct]
    for (const auto relocation : m_unresolved_relocations) {
```

lux01:
CI is passing and I think I've resolved all the comments from @itamar8910 :)

Issue #4277, opened 2020-11-30
Spreadsheet: problems selecting multiple cells


danboid:
Currently it is only easy to select multiple vertically stacked cells. I have not been able to select more than 4 cells horizontally and rectangular sections don't really work unless you go across first then drag down.

danboid:
I have been able to select more than 4 cells horizontally but most of the time it stops at 3.

alimpfard:
Huh.
I didn't notice it was this bad - I tried it a few times and it seemed to work initially, but I do see this behaviour now.
current me wants to slap another "committed to doing a select" bool and call it okay, but I'm sure there's a better solution.

bcoles:
I'm guessing this is a new bug, but it fits with this issue title. Attempting to click and drag a cell causes Spreadsheet to crash.

```
USERSPACE(42) ASSERTION FAILED: i < m_size
.././AK/Vector.h:198
[Spreadsheet(42:42)]: Terminating Spreadsheet(42) due to signal 6
[FinalizerTask(3:3)]: 0xdeadc0de  Kernel::Processor::switch_context(Kernel::Thread*&, Kernel::Thread*&) +181
0xdeadc0de  Kernel::Scheduler::context_switch(Kernel::Thread*) +155
0xdeadc0de  Kernel::Scheduler::pick_next() +857
0xdeadc0de  Kernel::Processor::check_invoke_scheduler() +125
0xdeadc0de  Kernel::Thread::die_if_needed() +163
0xdeadc0de  syscall_handler +2181
0xdeadc0de  syscall_asm_entry +49
0x0384e214
0x03846309
0x03846c09
0x03fb48b0
0x03fb007a
0x03b86f15
0x03bd821b
0x03bd861f
0x03fb06bc
0x038d2bb9
0x03c0d146
0x03c1059f
0x038d2bb9
0x038c7a90
0x038c7d39
0x03b8c390
0x03f949e3
0x03f94b4c
0x0d9fe0cb

[FinalizerTask(3:3)]: Generating coredump for pid: 42
[FinalizerTask(3:3)]: RTC: Year: 2021, month: 1, day: 2, hour: 5, minute: 53, second: 30
WindowServer(15): ClientConnection{0x01b73f18}::post_message: Disconnected from peer
CrashDaemon(19): New coredump file: /tmp/coredump/Spreadsheet_42_1609566810
CrashDaemon(19): 0x0384e198: [libc.so] kill +0x1c (Syscall.h:479)
[...]

PR #4744, opened 2021-01-02
Spreadsheet: Tiny fixes

### LINKED ISSUE #4716: Spreadsheet: "click to view" example on sum help corrupts JS interpreter stack
Steps:

- Open Spreadsheets
- Open Function Help box from the ? next to the formula bar.
- Click on `sum`
- Click the `view example` buttom
- Close the example spreadsheet with the `x`
- Close the help box with the `x`
- Close the main spreadsheet window with the `x`
- 💥 

```
USERSPACE(29) ASSERTION FAILED: popped_interpreter == &interpreter
../Libraries/LibJS/Runtime/VM.cpp:87
```

```
CrashDaemon(19): New coredump file: /tmp/coredump/Spreadsheet_29_1609509863
Shell(24): Job entry "Spreadsheet" deleted in 160918 ms
CrashDaemon(19): 0x045fdf5e: [libc.so] kill +28 (Syscall.h:479)
CrashDaemon(19): 0x045fdfda: [libc.so] raise +32 (signal.cpp:54)
CrashDaemon(19): 0x045f60c9: [libc.so]  +25 (stdlib.cpp:230)
CrashDaemon(19): 0x045f69c9: [libc.so] tolower +0 (assert.cpp:51)
CrashDaemon(19): 0x0479f005: [libjs.so] JS::VM::pop_interpreter(JS::Interpreter&) +111 (VM.cpp:88)
CrashDaemon(19): 0x0479f035: [libjs.so] JS::VM::InterpreterExecutionScope::~InterpreterExecutionScope() +39 (VM.cpp:99)
CrashDaemon(19): 0x04d834d9: [/bin/Spreadsheet] Spreadsheet::SpreadsheetWidget::~SpreadsheetWidget() +67 (NonnullOwnPtr.h:157)
CrashDaemon(19): 0x04d83722: [/bin/Spreadsheet] Spreadsheet::SpreadsheetWidget::~SpreadsheetWidget() +28 (kmalloc.h:69)
CrashDaemon(19): 0x04681916: [libcore.so] AK::RefCounted<Core::Object>::unref() const [clone .isra.0] +58 (RefCounted.h:119)
CrashDaemon(19): 0x0468336f: [libcore.so] AK::Vector<AK::NonnullRefPtr<Core::Object>, 0ul>::clear_with_capacity() +105 (Atomic.h:268)
CrashDaemon(19): 0x046833a4: [libcore.so] AK::Vector<AK::NonnullRefPtr<Core::Object>, 0ul>::clear() +28 (Vector.h:138)
CrashDaemon(19): 0x04682c69: [libcore.so] Core::Object::~Object() +199 (NonnullPtrVector.h:34)
CrashDaemon(19): 0x049ca11b: [libgui.so] GUI::Window::~Window() +547 (Window.cpp:92)
CrashDaemon(19): 0x049ca16a: [libgui.so] GUI::Window::~Window() +28 (kmalloc.h:69)
CrashDaemon(19): 0x04d6d66f: [/bin/Spreadsheet] AK::RefCounted<Core::Object>::unref() const +63 (RefCounted.h:119)
CrashDaemon(19): 0x04d651a5: [/bin/Spreadsheet] main +4032 (NonnullRefPtr.h:56)
CrashDaemon(19): 0x04d6526c: [/bin/Spreadsheet] _start +36 (crt0_shared.cpp:50)
```

alimpfard:
This addresses https://github.com/SerenityOS/serenity/issues/4277#issuecomment-753436495 and fixes #4716.

### DIFF
--- a/Applications/Spreadsheet/HelpWindow.cpp
+++ b/Applications/Spreadsheet/HelpWindow.cpp
@@ -124,11 +124,13 @@ HelpWindow::HelpWindow(GUI::Window* parent)
                 return;
             }
 
-            auto dialog = GUI::Window::construct(this);
-            dialog->resize(size());
-            dialog->set_icon(icon());
-            dialog->set_title(String::formatted("Spreadsheet Help - Example {} for {}", name, entry));
-            auto& widget = dialog->set_main_widget<SpreadsheetWidget>(NonnullRefPtrVector<Sheet> {}, false);
+            auto window = GUI::Window::construct(this);
+            window->resize(size());
+            window->set_icon(icon());
+            window->set_title(String::formatted("Spreadsheet Help - Example {} for {}", name, entry));
+            window->on_close = [window = window.ptr()] { window->remove_from_parent(); };
+
+            auto& widget = window->set_main_widget<SpreadsheetWidget>(NonnullRefPtrVector<Sheet> {}, false);
             auto sheet = Sheet::from_json(value.as_object(), widget.workbook());
             if (!sheet) {
                 GUI::MessageBox::show_error(this, String::formatted("Corrupted example '{}' in '{}'", name, url.path()));
@@ -136,7 +138,7 @@ HelpWindow::HelpWindow(GUI::Window* parent)
             }
 
             widget.add_sheet(sheet.release_nonnull());
-            dialog->show();
+            window->show();
         } else if (url.host() == "doc") {
             auto entry = LexicalPath(url.path()).basename();
             m_webview->load(URL::create_with_data("text/html", render(entry)));
--- a/Applications/Spreadsheet/HelpWindow.h
+++ b/Applications/Spreadsheet/HelpWindow.h
@@ -27,6 +27,7 @@
 #pragma once
 
 #include <AK/JsonObject.h>
+#include <LibGUI/Dialog.h>
 #include <LibGUI/Widget.h>
 #include <LibGUI/Window.h>
 #include <LibWeb/OutOfProcessWebView.h>
@@ -37,12 +38,12 @@ class HelpWindow : public GUI::Window {
     C_OBJECT(HelpWindow);
 
 public:
-    static NonnullRefPtr<HelpWindow> the()
+    static NonnullRefPtr<HelpWindow> the(GUI::Window* window)
     {
         if (s_the)
             return *s_the;
 
-        return *(s_the = adopt(*new HelpWindow));
+        return *(s_the = adopt(*new HelpWindow(window)));
     }
 
     virtual ~HelpWindow() override;
--- a/Applications/Spreadsheet/SpreadsheetWidget.cpp
+++ b/Applications/Spreadsheet/SpreadsheetWidget.cpp
@@ -60,7 +60,7 @@ SpreadsheetWidget::SpreadsheetWidget(NonnullRefPtrVector<Sheet>&& sheets, bool s
     help_button.set_fixed_size(20, 20);
     help_button.on_click = [&](auto) {
         auto docs = m_selected_view->sheet().gather_documentation();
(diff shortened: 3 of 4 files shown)


Applications/Spreadsheet/HelpWindow.cpp
```diff
@@ -124,19 +124,26 @@ HelpWindow::HelpWindow(GUI::Window* parent)
...
                 return;
             }
 
             widget.add_sheet(sheet.release_nonnull());
-            dialog->show();
+            window->show();
+            window->on_close_request = [this, key = window.ptr()] {
+                auto window = m_all_example_windows.get(key).value();
+                m_all_example_windows.remove(key);
+                const_cast<GUI::Window*>(window)->remove_from_parent();
```
alimpfard, line 144: I feel like this is a bit hacky, not sure why `GUI::Window::construct(parent)` has to create a strong reference to the new object in the parent, but this is what I came up with :shrug:

awesomekling: Why do you need `m_all_example_windows` at all?

alimpfard: Another option is to copy `window` into this lambda to be able to call `remove_from_parent()` on it. but that's a reference cycle and I'd rather not introduce any of those

alimpfard: Am I missing an obvious solution here?

awesomekling: What I've done elswhere is capturing a raw pointer and using `on_close`:

```cpp
    properties->on_close = [properties = properties.ptr()] {
        properties->remove_from_parent();
    };
```

awesomekling: It's not the most beautiful thing, but meh. Another alternative might be to have `on_close` include  a `Window&` parameter.

alimpfard: That's what I was trying to avoid, but if it's there already, I'll change this to match.

Perhaps a better suggestion, why not just remove the window from its parent when it's "closed"?

awesomekling: I've specifically tried to avoid those kinds of magical ownership behaviors in LibCore/LibGUI. Some GUI toolkits work that way and it makes lifetime very confusing.

alimpfard: is it really magical though? I fully expect a window to be deleted when
- I have no references to it
- it's closed

if anything, I think the parent holding on to it is more surprising than not.

Libraries/LibGUI/Window.cpp
```diff
@@ -96,6 +96,8 @@ void Window::close()
     hide();
     if (on_close)
         on_close();
+    if (this->ref_count() == 1) // Only the parent is holding on to us
+        remove_from_parent();
```
alimpfard, line 100: Ooops

Applications/Spreadsheet/HelpWindow.cpp
awesomekling: Hm, maybe you're right. It does seem like a fairly intuitive behavior. Let's do it.

alimpfard: I tried this, but it doesn't look like we can reliably detect "only referenced by parent object and being closed".
a window with no external refs has a refcount of 3 in its `close()` (when invoked through the close button)
The same could also happen if it's called from outside the event handlers, where it would have refcount == 3.
Unconditionally removing from the parent child list isn't really an option either, it would also be really surprising.

I'll just leave it be the way it is for now

PR #4747, opened 2021-01-02
TextEditor: Add basics of a Vim emulator

rok-povsic:
This PR adds some basics of Vim emulation to Text Editor and Hack Studio. 

Enable it in the upper menu or via shortcut `CTRL+SHIFT+ALT+V`.

### DIFF
--- a/Libraries/LibGUI/EditingEngine.h
+++ b/Libraries/LibGUI/EditingEngine.h
@@ -1,0 +1,85 @@
+/*
+ * Copyright (c) 2021, the SerenityOS developers.
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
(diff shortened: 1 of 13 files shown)


Libraries/LibGUI/TextEditor.cpp
```diff
@@ -1024,6 +1035,17 @@ void TextEditor::keydown_event(KeyEvent& event)
...
+                        move_to_line_beginning();
+                        add_code_point(0x0A);
+                        move_one_up();
+                        switch_to_insert_mode();
+                    } else if (event.key() == KeyCode::Key_O) {
+                        move_one_down();
+                        move_to_line_beginning();
+                        add_code_point(0x0A);
+                        move_one_up();
+                        switch_to_insert_mode();
```
emanuele6, line 1048: Can't this just be
```cpp
move_to_line_end();
add_code_point(0x0A);
switch_to_insert_mode();
```

alimpfard:
Maybe I'm being silly, but can we have one library implementation of this (much like neovim)?
Then we could trivially slap it on anything (e.g. LibLine, a future termination text editor, etc) and magically give those vim capabilities
It could probably just be in-process for now.

Libraries/LibGUI/TextEditor.cpp
rok-povsic: Yes, done. :+1:

rok-povsic:
That's an option as well.

ccapitalK:
I feel like it would be best if we follow @alimpfard 's suggestion, and move it into its own library. VIM's behaviour can be quite complicated, we should factor all VIM specfic logic into its own VimEngine class, which then interacts with the text editor itself through a simple interface. Having it this tightly coupled to the texteditor would just end up making the TextEditor code unreadable in the long term.

If we do this, then down the line we could have built in vim navigation in the browser as well. :^)

rok-clarifai:
Makes sense, I'll do that. :)

rok-povsic:
I've added the `EditingEngine` abstraction that handles most of the text editing logic, with children `RegularEditingEngine` and `VimEditingEngine`.

Libraries/LibGUI/EditingEngine.h
```diff
@@ -0,0 +1,85 @@
...
+    AK_MAKE_NONCOPYABLE(EditingEngine);
+    AK_MAKE_NONMOVABLE(EditingEngine);
+
+public:
+    virtual ~EditingEngine();
+
+    virtual CursorWidth cursor_width() const { return NARROW; }
+
+    void attach(TextEditor& editor);
+    void detach();
```
alimpfard, line 50: Not gonna lie, this wasn't exactly what I had in mind when I suggested moving the vim logic to a library...what I had in mind was more of an engine that had its own set of commands (e.g. `MoveOneLeft`, `MoveOneRight`, `MoveUpSteps`, etc.) and had zero ties to any specific editor implementation - that's why I gave neovim as an example, but it seems I was ambiguous at best.

rok-povsic: Such decoupling would be quite hard to implement since the interaction surface is big. For instance, handling selections will require relatively tight integration. Is this worth the effort at this initial step?

alimpfard: I'm not entirely sure, if you think it won't hurt a future refactoring too much, you can leave it as it is.

(possibly for the future)
regarding the selections, it really does look like everything depends on the selection, maybe the engine should have its own idea of the selection, and the editor would just show said selection?

rok-povsic: > maybe the engine should have its own idea of the selection, and the editor would just show said selection?

I've thought about this, but it would require that mouse selection be handled by this module as well (or at least, forwarded there, from the text editor). Which could be very well be the case, but I thought it might be too big of a refactoring, just to add a vim emulation (at least in this step).

alimpfard: > mouse
> vim mode

disable that lol
(just a joke ^)

alimpfard: On a more serious note, it's alright to have it be dependent on TextEditor for now, we can always change it later.

rok-povsic:
Rebased & most commits squashed.

awesomekling:
This looks quite nice! Please configure your editor to put a newline at the end of files and we're good to go. :)

rok-povsic:
Whoops. Fixed. :)

colfrog (commented):
Very cool!

PR #4748, opened 2021-01-02
Kernel: VirtIO-based devices support

IdanHo:
This is a continuation of tomuta's work in #3236, this is still WIP but should soon allow for at least naive usage of split virtqueues which would allow us to stop using the considerably slower method of knocking on port 0xe9 for each character when outputting to qemu console by move to virtconsole and in the future also allow us to implement other virtio based devices and services (like porting the qemu agent which uses virtserial to allow clipboard sharing, or implementing a virtrng driver to fix rng starvation problems).

If anyone wants to contribute currently the 2 main problems are that buffers received from sys$write have to be memcpyed to contiguous physical memory pages that could then be handed off to the virtqueue (as the virtqueue requires physical addresses), and theres also a problem currently where the virt device sends us interrupts when the interrupt status is set to 0 (which is not possible as defined by the specification) so theres probably a bug in there somewhere as well.


Kernel/VirtIO/VirtIO.h
```diff
@@ -0,0 +1,251 @@
...
+                generation_before = config_read8(m_common_cfg, 0x15);
+                f();
+                generation_after = config_read8(m_common_cfg, 0x15);
+            } while (generation_before != generation_after);
+        } else {
+            f();
+        }
+    }
+
+    u8 config_read8(const Configuration*, u32);
```
supercomputer7, line 165: Please pass a const reference and not a pointer to that object.
Also, if this method doesn't change anything, mark it as `const` :)

Kernel/VirtIO/VirtIO.h
```diff
@@ -0,0 +1,251 @@
...
+
+    auto mapping_for_bar(u8) -> MappedMMIO&;
+
+    u8 read_status_bits();
+    void clear_status_bit(u8);
+    void set_status_bit(u8);
+    u64 get_device_features();
+    bool finish_init();
+
+    VirtIOQueue* get_queue(u16 queue_index)
```
supercomputer7, line 181: Either use the `OwnPtr` container, or return a reference and put in the first line of the method `ASSERT` statement.

Kernel/VirtIO/VirtIO.h
```diff
@@ -0,0 +1,251 @@
...
+                continue;
+            if (index > 0) {
+                index--;
+                continue;
+            }
+            return &cfg;
+        }
+        return nullptr;
+    }
+    const Configuration* get_common_config(u32 index = 0) const;
```
supercomputer7, line 147: Please use `OwnPtr` container here :)

Kernel/VirtIO/VirtIO.h
```diff
@@ -0,0 +1,251 @@
...
+    bool setup_queue(u16 queue_index);
+    void notify_queue(u16 queue_index);
+
+    void reset_device();
+
+    u8 isr_status();
+
+    NonnullOwnPtrVector<VirtIOQueue> m_queues;
+    Vector<Configuration> m_configs;
+    const Configuration* m_common_cfg { nullptr }; // Cached due to high usage
```
supercomputer7, line 237: Maybe `Configuration` should be a `RefCounted` object? If we want to keep a fast way to access them and also to store them in a `Vector`, this is the way to go.

Kernel/VirtIO/VirtIO.h
```diff
@@ -0,0 +1,251 @@
...
+#define DEVICE_CONFIG_INTERRUPT 0x2
+
+class VirtIO {
+public:
+    static void detect();
+};
+
+class VirtIODevice : public PCI::Device {
+public:
+    VirtIODevice(PCI::Address, const char*);
```
supercomputer7, line 96: Please don't pass a `const char*` in the constructor. Usually we have a virtual method of `class_name()` that each class that inherits from the parent class should implement and return the desirable string.

Kernel/VirtIO/VirtIO.h
```diff
@@ -0,0 +1,251 @@
...
+    bool is_feature_accepted(u64 feature) const
+    {
+        ASSERT(m_did_accept_features);
+        return is_feature_set(m_accepted_features, feature);
+    }
+
+    void supply_buffer_and_notify(u16 queue_index, const u8* buffer, u32 len, BufferType);
+
+    virtual void handle_irq(const RegisterState&) override;
+    virtual void handle_device_config_change() = 0;
```
supercomputer7, line 210: This method doesn't return any result to the user if it succeeded or not. Is there a mechanism inside such method to set a flag in case of failure?

Kernel/VirtIO/VirtIOQueue.h
```diff
@@ -0,0 +1,103 @@
...
+        VirtIOQueueDeviceItem rings[];
+    };
+
+    const u16 m_queue_size;
+    const u16 m_notify_offset;
+    u16 m_free_buffers;
+    u16 m_free_head { 0 };
+    u16 m_used_tail { 0 };
+
+    VirtIOQueueDescriptor* m_descriptors { nullptr };
```
supercomputer7, line 96: Please avoid using raw pointers.

Kernel/VirtIO/VirtIOQueue.cpp
```diff
@@ -0,0 +1,106 @@
...
+{
+    size_t size_of_descriptors = sizeof(VirtIOQueueDescriptor) * queue_size;
+    size_t size_of_driver = sizeof(VirtIOQueueDriver) + queue_size * sizeof(u16);
+    size_t size_of_device = sizeof(VirtIOQueueDevice) + queue_size * sizeof(VirtIOQueueDeviceItem);
+    m_region = MM.allocate_contiguous_kernel_region(PAGE_ROUND_UP(size_of_descriptors + size_of_driver + size_of_device), "VirtIO Queue", Region::Access::Read | Region::Access::Write);
+    if (m_region) {
+        // TODO: ensure alignment!!!
+        u8* ptr = m_region->vaddr().as_ptr();
+        memset(ptr, 0, m_region->size());
+        m_descriptors = reinterpret_cast<VirtIOQueueDescriptor*>(ptr);
```
supercomputer7, line 45: Let's make this piece of code more robust by not using `reinterpret_cast`, and to try to do all of these pointer arithmetic in a different way.

Kernel/VirtIO/VirtIOQueue.h
```diff
@@ -0,0 +1,103 @@
...
+    void disable_interrupts();
+
+    PhysicalAddress descriptor_area() const { return to_physical(m_descriptors); }
+    PhysicalAddress driver_area() const { return to_physical(m_driver); }
+    PhysicalAddress device_area() const { return to_physical(m_device); }
+
+    bool supply_buffer(const u8* buffer, u32 len, BufferType);
+    bool new_data_available() const;
+
+    bool handle_interrupt();
```
supercomputer7, line 57: if this class handles interrupts, maybe it should inherit from `IRQHandler` class?

supercomputer7 (commented):
This is definitely a useful kernel feature (for all of us, with our VMs)!
Feel free to ask questions about my comments :)

Kernel/VirtIO/VirtIOQueue.cpp
tomuta: I'm not sure how you would do that when the location of these objects is dictated by the standard. It's relatively simple pointer arithmetic, and there isn't really a way around the cast.

Kernel/VirtIO/VirtIO.h
IdanHo: Unfortunately we cant be sure there arent any side effects to reading a config value (for example, reading the interrupt status field in the isr capability config causes a de-assertion of the interrupt) so it cant be const

supercomputer7: Does it change the object though? if not, it should be `const`. If something else changes in kernel environment, the caller should be aware of that.

Kernel/VirtIO/VirtIOQueue.h
IdanHo: it handles PCI Interrupts, so it handles from PCI::Device (which inherits from IRQHandler), should it inherit it directly as well?

Kernel/VirtIO/VirtIO.h
IdanHo: i pass that since the base VirtIODevice doesnt inherit from File, so it cant just use the subclass's implemented class_name() when debug-logging

Kernel/VirtIO/VirtIO.h
IdanHo: ah looks like we currently lazily map the BARs to memory (if and when they are used, which config reading is an instance of such a use) so it does have changes to the object.

Kernel/VirtIO/VirtIOQueue.h
supercomputer7: I prefer that the model of `PCI::DeviceController` as a the controller of all VirtIOQueue(s), and a VirtIOQueue class that inherits directly from `IRQHandler`. This way, it will be much easier to enable MSIs or MSI-X later.
`PCI::Device` **should be used only** if you know that the device will never use any other source of interrupts except IRQs and also, the device will only handle one IRQ line, which is not the case with VirtIO.

Kernel/VirtIO/VirtIO.h
supercomputer7: Why do we need that string then? If you want to check the type of a `VirtIODevice` derived class, please a `enum class` with all defined types instead. 
Also, what's the lifetime model of that `const char*` that is passed in the constructor? Who owns it? It's better if you use `AK::String` container instead.

Kernel/VirtIO/VirtIOQueue.h
IdanHo: I was under the impression that VirtIO devices only use one IRQ line, in what scenarios do they use more than one?

supercomputer7: Well, the VirtIO specification allows devices to utilize MSI & MSI-X to reduce IRQ sharing. We will eventually want to support MSI (and MSI-X), because it makes things much easier and also much faster than IRQs, because MSIs bypass the IRQ controller and interrupt the targeted CPU directly.

IdanHo: I see, the virtio device itself has to receive IRQs itself as well (as one of the irqs is a device config change), so should the the virt device class be an irqhandler as well besides the devicecontroller?

Kernel/VirtIO/VirtIO.h
awesomekling: Please only make things reference counted if they have multiple logical owners. Reference counting makes lifetime much harder to understand and reason about, and should not be used unless necessary :)

IdanHo:
My local implementation of virtio queues is mostly done, but i've come across the following issue:
VirtIO Devices expect (physically-)contiguous memory in each descriptor, and as such we cant accept any arbitrary buffer.  There are several ways to overcome this:
1. Allocate a physical page for each virtqueue and then just memcpy incoming buffers into that page - this has the downside of only allowing the processing of one request at a time, defeating the "queue" property of the virtqueue, as well as the costs of memcpying each incoming buffer, this is the method i have implemented right now
2. Allocate several contiguous physical pages and create a sort-of make shift wrapping linear allocator (This is will allow O(1) allocations and frees due to the guaranteed order of deallocations) - this still incurs memcpy costs as well as still blocking if the whole pre-allocated range is used (although much less frequently)
3. Implementing Indirect-Descriptor support (see virtio spec) and use it as a scatter/gather list by finding and using the physical pages each buffer spans - this is (probably) the best way, as it doesnt require us to do any memcpying (besides ensuring the original backing-buffer is not modified until the request is processed) and it doesnt incur additional memory costs, but im not sure if theres a public facing method to get the backing physical pages of an arbitary (virtual) buffer

tomuta:
My vision was implementing 3), though I don't recall if this must be supported by the host, or if this is a fairly recent addition. Maybe we don't care.

IdanHo:
> My vision was implementing 3), though I don't recall if this must be supported by the host, or if this is a fairly recent addition. Maybe we don't care.

it does have to be supported by the host, but as far as i can tell its very old (and qemu supports it) so it shouldnt be an issue

IdanHo:
The last 2 commits (mostly) finish the base VirtIO device and queue implementations, as always im not very familiar with development at such a low level, so id be appreciate suggestions for changes/improvements :)
About hooking up VirtIOConsole to replace knocking on port 0xe9: im less familiar with how general printing in the kernel/userspace is implemented, but it seems like most of it is funneled down to Console::put_char (but not all of it), but that since that only takes one character at a time instead of whole strings it defeats the purpose of the optimization VirtIOConsole provides, so if anyone can help with that itd be great :)

Matrix89:
Hi, is there any major reason(besides review) preventing the infrastructure for writing VirtIO drivers from being merged? I'm not talking about the console or the RNG drivers, just the overall abstraction around VirtIO devices/virtqueues.

ccapitalK:
> Hi, is there any major reason(besides review) preventing the infrastructure for writing VirtIO drivers from being merged? I'm not talking about the console or the RNG drivers, just the overall abstraction around VirtIO devices/virtqueues.

A bunch of this code is now outdated. I am currently working on rebasing and fixing it up (APIs have changed, things like memory_barrier() in AK have been replaced with atomics), so there is still a bit of work to be done to get the Virtio devices into mainline.

I think it would be a good idea to get the console and rng drivers in with the pr though, you can't really be sure that the virtqueues implementation is correct without having some simple drivers to confirm it.

ccapitalK:
I've ported these commits in #6340 6340

IdanHo:
Closed due to continued work in #6340

Issue #4751, opened 2021-01-02
NotificationServer: Notifications aren't repositioned after screen resolution change


silvasur:
When you change the screen resolution to a smaller size, the existing notifications will be partly drawn outside the screen area (extra annoying, as the close button is now no longer clickable).

Here's a screen recording of the bug:

https://user-images.githubusercontent.com/357232/103466976-63aaf900-4d4a-11eb-86a0-f8f76e3f42e7.mp4

A similar thing happens when you increase the resolution, the notifications will then be weirdly positioned too far away from the screen border. But at least you can still click the close button :).

linusg:
Fixed by #4752.

PR #4752, opened 2021-01-02
NotificationServer: Reposition notifications on screen resolution change

silvasur:
Previously notifications were (partially) drawn outside the screen rect if
they were created before changing the screen resolution to smaller
dimensions. This prevented the user from dismissing the notification as the
close button was no longer clickable.

See also issue #4751.

I'm not sure if it's a good idea to simply expose the previously static `update_notification_window_locations` function? Perhaps this should be a static method of the `NotificationWindow` class instead? I don't have too much experience with C++, so I don't know if this would be an idiomatic way of doing this.


awesomekling:
Cool! Fix looks good :)

Issue #4703, opened 2021-01-01
WindowServer: seems like SetGlobalCursorTracking breaks double clicking


TheMorc:
I found a possible bug while I was messing with developing apps for SerenityOS.

If you're on the desktop and try double clicking the Browser app icon to open it, all should work just fine. But if you open an app(ex. Eyes Demo) that uses SetGlobalCursorTracking then it seems to break double clicking and you can only single click on icons and other stuff.
And closing the app also reverts all back to normal.

PR #4756, opened 2021-01-02
Demos: CatDog, my 2nd try at a Neko clone demo

TheMorc:
My first attempt to make a SerenityOS app, _but 2nd try_.. I modified the code a bit since then. I also tried making my own sprites so that's the reason why the cat sometimes looks more like a dog than like a cat. And I renamed it to CatDog because of those sprites.
There seems to be only one bug, it's the SetGlobalCursorTracking mentioned in #4703
So I hope that it's finally all as it should be. And i also hope that CI will build it without any clang-format or build errors..


awesomekling:
I *love* that you came back with an original cat! Awesome work :)

TheMorc:
> I _love_ that you came back with an original cat!

And if I have to be honest, I didn't expect that making a "few" 32x32 sprites would take me a lot more time compared to making the whole demo app.

tomuta:
Looks cute, one recommendation I have would be to merge all the png into one and then blit each sprite using a rectangle from one source bitmap.

TheMorc:
@tomuta I can try it but the harder thing for me would still be to not mess up the branch with more commits because it still needs to be squashed together.

awesomekling:
It's fine, we can merge it as-is for now and put more improvements on top.

PR #4759, opened 2021-01-02
LibGUI: Stop the automatic autocomplete timer when typing whitespace / Playground: Improve the autocompleted suggestions

alimpfard:
Hopefully now less in-your-way than before!

---

There's no reason to bring up autocomplete when the user is just
inserting indentation and spaces and such.

- Now sorted
- Add a "layout" property to GUI::Widget and GUI::Frame
- Only complete Layouts for the values of "layout"
- Don't suggest anything if the only suggestion is what the user has
  already typed

### DIFF
--- a/AK/StringUtils.cpp
+++ b/AK/StringUtils.cpp
@@ -31,6 +31,7 @@
 #include <AK/StringUtils.h>
 #include <AK/StringView.h>
 #include <AK/Vector.h>
+#include <ctype.h>
 
 namespace AK {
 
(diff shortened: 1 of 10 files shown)


AK/StringUtils.cpp
```diff
@@ -36,6 +36,16 @@ namespace AK {
...
 namespace StringUtils {
 
+static constexpr bool is_whitespace_character(char ch)
+{
+    return ch == '\t'
+        || ch == '\n'
+        || ch == '\v'
+        || ch == '\f'
+        || ch == '\r'
+        || ch == ' ';
```
ADKaster, line 46: `#include <ctype.h>` ? this looks like `isspace`

alimpfard: Not sure if I _can_ do that in that header, maybe there's a reason why it was like that?
I'll try

alimpfard: Build seems okay with it, I'll just use isspace :shrug:

awesomekling:
LGTM but let's move the layout registrations to BoxLayout.cpp per Andrew's change :)

PR #4763, opened 2021-01-03
Kernel: Fix ProcFS returning Inodes that are being deleted

tomuta:
There is a window between dropping the last reference and removing
a ProcFSInode from the lookup map. So, when looking up we need to
check if that Inode is being destructed.

_This solved almost all the odd crashes I kept running into..._

### DIFF
--- a/Kernel/FileSystem/ProcFS.cpp
+++ b/Kernel/FileSystem/ProcFS.cpp
@@ -1101,12 +1101,20 @@ RefPtr<Inode> ProcFS::get_inode(InodeIdentifier inode_id) const
 
     LOCKER(m_inodes_lock);
     auto it = m_inodes.find(inode_id.index());
-    if (it == m_inodes.end()) {
-        auto inode = adopt(*new ProcFSInode(const_cast<ProcFS&>(*this), inode_id.index()));
-        m_inodes.set(inode_id.index(), inode.ptr());
-        return inode;
+    if (it != m_inodes.end()) {
+        // It's possible that the ProcFSInode ref count was dropped to 0 or
+        // the ~ProcFSInode destructor is even running already, but blocked
+        // from removing it from this map. So we need to *try* to ref it,
+        // and if that fails we cannot return this instance anymore and just
+        // create a new one.
+        if (it->value->try_ref())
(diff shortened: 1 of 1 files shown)


Kernel/FileSystem/ProcFS.cpp
```diff
@@ -1101,12 +1101,20 @@ RefPtr<Inode> ProcFS::get_inode(InodeIdentifier inode_id) const
...
-    if (it == m_inodes.end()) {
-        auto inode = adopt(*new ProcFSInode(const_cast<ProcFS&>(*this), inode_id.index()));
-        m_inodes.set(inode_id.index(), inode.ptr());
-        return inode;
+    if (it != m_inodes.end()) {
+        // It's possible that the ProcFSInode ref count was dropped to 0 or
+        // the ~ProcFSInode destructor is even running already, but blocked
+        // from removing it from this map. So we need to *try* to ref it,
+        // and if that fails we cannot return this instance anymore and just
+        // create a new one.
```
awesomekling, line 1109: Would a better option be to have a custom deleter for `ProcFSInode` that takes `m_inodes_lock` before calling the destructor?

tomuta: That still wouldn't close the window between dropping the ref count to 0 and taking the `m_inodes_lock`, though.

tomuta: `try_ref` is pretty light-weight, so I think the overhead is minimal.

awesomekling: Oh yeah I'm not worried about overhead, was more thinking if there's a way we can come up with a safe pattern for situations like this.

But this is okay for now.

tomuta: Yeah I'm not sure, the only thing I can think of that might be better is the map value being an actual RefPtr and something other than the destructor would be in charge of removing it from the map. But even if it was a WeakPtr, it wouldn't clean up the map by itself, either. Maybe a specialized HashMap for WeakPtr values, but then find would have to obtain a `strong_ref` to be able to figure out if it should delete that entry and retrying before returning...

PR #4768, opened 2021-01-03
Meta: Add a script that makes sure files end in a newline.

emanuele6:
This script checks .js, .cpp, .h, .gml and .sh files.

It also makes sure that there are no black lines at the end of the files checked. (tested)

-------------------

NB: this commit should fail the checks of the CI because of these files:
```
Files missing with no newline at the end: Libraries/LibGfx/Font.cpp Ports/build_all.sh Ports/grep/package.sh Ports/less/package.sh Ports/libexpat/package.sh Ports/libiconv/package.sh Ports/m4/package.sh Ports/mandoc/package.sh Ports/mbedtls/package.sh Ports/rsync/package.sh Ports/sed/package.sh
```
https://github.com/SerenityOS/serenity/runs/1639694693?check_suite_focus=true


linusg:
THANK YOU! :smile:

> This script checks .js, .cpp, .h, .gml and .sh files.

I'd suggest to add CSS and HTML as well, we have a few of those.

Meta/lint-ci.sh
```diff
@@ -16,6 +16,7 @@ set +e
 for cmd in \
         Meta/check-ak-test-files.sh \
         Meta/check-debug-flags.sh \
+        Meta/check-newlines-at-eof.sh \
```
linusg, line 19: Separate script is fine as well but IMO this could just be part of check-stye.sh

emanuele6: > Separate script is fine as well but IMO this could just be part of check-stye.sh

I created a separate file because `check-style.sh` only checks .cpp and .h files; also all the checks there use regexps while these new checks do not.

linusg: > only checks .cpp and .h files

ah good point, I forgot. :+1:

Meta/check-newlines-at-eof.sh
```diff
@@ -0,0 +1,42 @@
...
+    ':!:Kernel/FileSystem/ext2_fs.h' \
+    ':!:Libraries/LibC/getopt.cpp' \
+    ':!:Libraries/LibCore/puff.h' \
+    ':!:Libraries/LibCore/puff.cpp' \
+    ':!:Libraries/LibELF/exec_elf.h' \
+)
+
+exit_status=0
+if (( ${#MISSING_NEWLINE_AT_EOF_ERRORS[@]} )); then
+    echo "Files missing with no newline at the end: ${MISSING_NEWLINE_AT_EOF_ERRORS[*]}"
```
emanuele6, line 35: oops, I need to fix this.

Meta/check-newlines-at-eof.sh
```diff
@@ -0,0 +1,42 @@
...
+
+MISSING_NEWLINE_AT_EOF_ERRORS=()
+MORE_THAN_ONE_NEWLINE_AT_EOF_ERRORS=()
+
+while IFS= read -r f; do
+    [ -s "$f" ] || continue
+
+    if [ "$(tail -n 1 "$f" | wc -l)" != 1 ]; then
+        MISSING_NEWLINE_AT_EOF_ERRORS+=( "$f" )
+    elif [[ "$(tail -n 2 "$f")" =~ ^[[:space:]]*$ ]]; then
```
emanuele6, line 14:
```suggestion
    elif [[ "$(tail -n 1 "$f")" =~ ^[[:space:]]*$ ]]; then
```

This should be a 1, not a 2!
I need to fix more files.
```
Files that have blank lines at the end: Applications/TextEditor/TextEditorWindow.gml Meta/text-to-cpp-string.sh Ports/indent/package.sh Shell/Tests/backgrounding.sh
```

emanuele6: https://github.com/SerenityOS/serenity/runs/1639805371

emanuele6:
It should all be good now.

PR #4770, opened 2021-01-03
Kernel+Loader.so: Allow dynamic executables without an interpreter

lux01:
Commit a3a9016701e487a5ca92d83b8cff179a190cdeb2 removed the PT_INTERP header
from Loader.so which cleaned up some kernel code in execve. Unfortunately
it prevents Loader.so from being run as an executable

### DIFF
--- a/Kernel/Syscalls/execve.cpp
+++ b/Kernel/Syscalls/execve.cpp
@@ -276,7 +276,7 @@ KResultOr<Process::LoadResult> Process::load_elf_object(FileDescription& object_
     };
 }
 
-KResultOr<Process::LoadResult> Process::load(NonnullRefPtr<FileDescription> main_program_description, RefPtr<FileDescription> interpreter_description)
+KResultOr<Process::LoadResult> Process::load(NonnullRefPtr<FileDescription> main_program_description, RefPtr<FileDescription> interpreter_description, bool is_dynamic)
 {
     RefPtr<PageDirectory> old_page_directory;
     NonnullOwnPtrVector<Region> old_regions;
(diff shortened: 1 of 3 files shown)


awesomekling:
Why is it useful to run Loader.so as an executable?

lux01:
It provides a convenient point for users to enable dynamic linker debugging options at runtime, rather than having to rebuild Loader.so and enable them for all programs. Specifically the next item on my list is to implement `ldd` which is usually a shell script wrapper around passing the `--list` option to the dynamic linker.

awesomekling:
All right, that makes sense :)

Kernel/Syscalls/execve.cpp
```diff
@@ -582,7 +583,7 @@ static KResultOr<Vector<String>> find_shebang_interpreter_for_executable(const c
     return KResult(-ENOEXEC);
 }
 
-KResultOr<NonnullRefPtr<FileDescription>> Process::find_elf_interpreter_for_executable(const String& path, char (&first_page)[PAGE_SIZE], int nread, size_t file_size)
+KResultOr<RefPtr<FileDescription>> Process::find_elf_interpreter_for_executable(const String& path, char (&first_page)[PAGE_SIZE], int nread, size_t file_size)
```
awesomekling, line 586: This return type has two failure states (`KResult` with an error, or null `RefPtr`.)
Let's keep the `KResultOr<NonnullRefPtr<FileDescription>>`.

lux01: Null `RefPtr` is no longer an error state unless I misread something, it's now an accepted state indicating a dynamic executable with no interpreter. The null value for `interpreter_descriptor` is then used in `Process::load()` to decide whether `interpreter_descriptor` or `main_program_descriptor`.

awesomekling: Ohh, I see. That's an unusual situation but yes it does make sense here :)

PR #4771, opened 2021-01-03
Meta: Rewrite the check-newlines-at-eof script in python :^)

emanuele6:
The bash version takes around 15 seconds to run; that is way too slow.
This python3 version should take less than one second to run. :^)

Additionally, the script will now also checks .py files.


Meta/check-newlines-at-eof.py
```diff
@@ -0,0 +1,73 @@
+#!/bin/python3
+
+import os
+import subprocess
+import sys
+
+try:
+    os.chdir(os.path.dirname(__file__) + "/..")
+except:
+    os.exit(1)
```
linusg, line 10: Raw `except` is almost never what you want, something like `except OSError` is more appropriate. I don't think the try/except is necessary at all though, just let it raise, that will exit the script as well (and shows the error to the user rather than exiting silently).

Meta/check-newlines-at-eof.py
```diff
@@ -0,0 +1,73 @@
...
+        "*.py",
+        ":!:Base",
+        ":!:Kernel/FileSystem/ext2_fs.h",
+        ":!:Libraries/LibC/getopt.cpp",
+        ":!:Libraries/LibCore/puff.h",
+        ":!:Libraries/LibCore/puff.cpp",
+        ":!:Libraries/LibELF/exec_elf.h"
+    ],
+    capture_output=True
+).stdout.decode().split('\n')
```
linusg, line 31: This would be much more likely to raise as it depends on an external command ("git") - if it fails it throws an error message at the user, which is what I mean in my previous comment :)

Meta/check-newlines-at-eof.py
```diff
@@ -0,0 +1,73 @@
...
+
+MISSING_NEWLINE_AT_EOF_ERRORS = []
+MORE_THAN_ONE_NEWLINE_AT_EOF_ERRORS = []
+
+did_fail = False
+for filename in files:
+    try:
+        f = open(filename, "r")
+    except:
+        continue
```
linusg, line 41: Same here, silently skipping files probably isn't a good idea.

Meta/check-newlines-at-eof.py
```diff
@@ -0,0 +1,73 @@
...
+
+    f.seek(0, os.SEEK_END)
```
linusg, line 43: I don't think a freshly opened file is ever not at seek position 0?

emanuele6:
https://github.com/SerenityOS/serenity/runs/1640412986?check_suite_focus=true

Meta/check-newlines-at-eof.py
```diff
@@ -0,0 +1,73 @@
...
+                MORE_THAN_ONE_NEWLINE_AT_EOF_ERRORS.append(filename)
+                break
+
+        f.close()
+        continue
+
+    did_fail = True
+    MISSING_NEWLINE_AT_EOF_ERRORS.append(filename)
+
+    f.close()
```
linusg, line 64: If you wrap this whole block in a with statement it will close the file automatically for you:

```python
for filename in files:
    with open(filename, "r") as f:
        f.seek(f.tell() - 1, os.SEEK_SET)
        if f.read(1) == '\n':
            while True:
                f.seek(f.tell() - 2, os.SEEK_SET)
                char = f.read(1)
                if not char.isspace():
                    break
                if char == '\n':
                    did_fail = True
                    MORE_THAN_ONE_NEWLINE_AT_EOF_ERRORS.append(filename)
                    break
        else:
            did_fail = True
            MISSING_NEWLINE_AT_EOF_ERRORS.append(filename)
```

Meta/check-newlines-at-eof.py
```diff
@@ -0,0 +1,73 @@
...
+
+if MISSING_NEWLINE_AT_EOF_ERRORS:
+    print("Files with no newline at the end:", " ".join(MISSING_NEWLINE_AT_EOF_ERRORS))
+if MORE_THAN_ONE_NEWLINE_AT_EOF_ERRORS:
+    print("Files that have blank lines at the end:", " ".join(MORE_THAN_ONE_NEWLINE_AT_EOF_ERRORS))
+
+if did_fail:
+    sys.exit(1)
+sys.exit(0)
```
linusg, line 73: not necessary, this is implied.

```suggestion
```

linusg (changes requested):

Meta/check-newlines-at-eof.py
emanuele6: Oh, I'm use to `bash` in which it isn't implied. Ok, I'll remove it

Meta/check-newlines-at-eof.py
emanuele6: This goes to the end of the file.

linusg: ...right. sorry. :facepalm:

Meta/check-newlines-at-eof.py
emanuele6: That's for the final element of files which is an empty line; I could also use `filename in files[:-1]` and remove the try block.

linusg: That'd be a lot better, it's not obvious (without a comment). `stdout.decode().trim().split('\n')` should avoid this altogether and only put files in `files` :)

Meta/check-newlines-at-eof.py
emanuele6: you are right.

Meta/check-newlines-at-eof.py
emanuele6: I get this error:
```
Traceback (most recent call last):
  File "/home/emanuele6/pro/serenityOS/serenity/Meta/check-newlines-at-eof.py", line 9, in <module>
    files = subprocess.run(
AttributeError: 'str' object has no attribute 'trim'
```

emanuele6: Oh, I think that method is called `strip`

linusg: Ugh, it's trim in JS but strip in Python, I always get those mixed.

emanuele6: Is there a method to only trim one `\n` from the end of the string? We probably don't want to remove spaces from the first and the last filename (just in case)

linusg: rstrip with argument:

```python
>>> "   aaa   \n\n".rstrip("\n")
'   aaa   '
```

emanuele6: `).stdout.decode()[:-1].split('\n')` we just use. but that will remove any character from the end which is not expressive.

emanuele6: @linusg oh, I didn't see your comment.

You are right, I was worrying about using `rstrip('\n')` because the last filename could end in a `\n`, but that is already and issue since we are using `split('\n')`. let's just use `rstrip('\n')`. :laughing:

emanuele6:
Now, it also checks `CMakeLists.txt` files.

linusg (approved):

Meta/check-newlines-at-eof.py
```diff
@@ -0,0 +1,61 @@
...
+        "*/CMakeLists.txt",
+        ":!:Base",
+        ":!:Kernel/FileSystem/ext2_fs.h",
+        ":!:Libraries/LibC/getopt.cpp",
+        ":!:Libraries/LibCore/puff.h",
+        ":!:Libraries/LibCore/puff.cpp",
+        ":!:Libraries/LibELF/exec_elf.h"
+    ],
+    capture_output=True
+).stdout.decode().rstrip('\n').split('\n')
```
emanuele6, line 30: Since we are using `split('\n')`, there is no reason to not use `strip('\n')`, I'll change it.

awesomekling:
Oh heck yeah :D

emanuele6:
I spotted a blank line at the end of Toolchain/CMakeToolchain.txt and I modified the script to make it check any txt file with a filename that begins with "CMake", instead of just txt files named `CMakeLists.txt`.

Meta/check-newlines-at-eof.py
```diff
@@ -0,0 +1,62 @@
...
+        "*.cpp",
+        "*.h",
+        "*.gml",
+        "*.html",
+        "*.js",
+        "*.css",
+        "*.sh",
+        "*.py",
+        "CMake[^/]*.txt",
+        "*/CMake[^/]*.txt",
```
emanuele6, line 21: These are not regular expressions. I should use this pattern instead:

```suggestion
        "CMake*.txt",
        "*/CMake*.txt",
```

> An asterisk "*" matches anything except a slash.

https://git-scm.com/docs/gitignore

Meta/check-newlines-at-eof.py
```diff
@@ -0,0 +1,62 @@
...
+        ":!:Libraries/LibC/getopt.cpp",
+        ":!:Libraries/LibCore/puff.h",
+        ":!:Libraries/LibCore/puff.cpp",
+        ":!:Libraries/LibELF/exec_elf.h"
+    ],
+    capture_output=True
+).stdout.decode().strip('\n').split('\n')
+
+MISSING_NEWLINE_AT_EOF_ERRORS = []
+MORE_THAN_ONE_NEWLINE_AT_EOF_ERRORS = []
```
ADKaster, line 33: Not sure it matters super much for this straight port of the bash script but seeing an all caps variable name that's not a constant in python code is surprising to me. I tend to follow either PEP-8 (with a checker like flake8) or the Google style guide for python files. https://google.github.io/styleguide/pyguide.html#s3.16-naming

emanuele6: ok, no problems.

emanuele6: > this straight port of the bash script

I wouldn't call it a straight port :^)

linusg (approved):

ADKaster (approved):

PR #4772, opened 2021-01-03
recommendation for clang-10 on Ubuntu 18.04.x

shantanugadgil:
Recommendation for clang-10 on Ubuntu 18.04.x.

The user doesn't have to enable any extra repos for this.


ADKaster:
Is this the correct recommendation? The CI has switched to using clang-format-11, so any contributor who installs clang-10 and not clang-11 will likely be using the wrong clang-format version.

shantanugadgil:
This worked for me a month ago. I was able to compile and boot up a test vm.
The latest code with AK/Tests and Demos fails with nullptr_t vs nullptr, so clang-10 may also be inadequate now.
I didn't know the CI switching to clang-11, we can drop this PR. 👍

PR #4773, opened 2021-01-03
Add GML formatter

linusg:
This adds a GML formatter and uses it in Playground as well as a new clang-format-like Userland program (`gml-format`).

gml-formatter

image

---

I already added `gml-format` to Lagom, so we could easily integrate it into the CI linting step. (in fact this PR was inspired by the GML format bikeshedding happening in another PR)


linusg:
dammit refined-github stop replacing my PR titles with the first commit message :roll_eyes:

alimpfard:
live formatting wen?

nit: don't mess the input up when it can't be formatted, just don't format at all (or do your best to make things right when there's no ambiguity?)

linusg:
> nit: don't mess the input up when it can't be formatted, just don't format at all (or do your best to make things right when there's no ambiguity?)

the only time it can't be formatted is when the GML is invalid and that's already being checked...

alimpfard:
I assumed that second picture was showing the output (I thought that `<-- missing ...` was being inserted by the formatter), I guess that's part of the input ^^'

linusg:
> I assumed that second picture was showing the output (I thought that <-- missing ... was being inserted by the formatter), I guess that's part of the input ^^'

ohh, yeah it totally looks like an error message, very confusing!
but you're right, it's already part of the input :)

DevTools/Playground/main.cpp
```diff
@@ -300,6 +301,35 @@ int main(int argc, char** argv)
...
+                    GUI::MessageBox::Type::Warning,
+                    GUI::MessageBox::InputType::OKCancel);
+                if (result == GUI::MessageBox::ExecCancel)
+                    return;
+                break;
+            }
+        }
+        auto formatted_gml = GUI::format_gml(source);
+        if (!formatted_gml.is_null()) {
+            editor.set_text(formatted_gml);
```
alimpfard, line 323: Future change idea: Maybe the editor could have a formatter (just like a highlighter and completion provider), and we could later enable "format on save" or "format on change"?

Obviously not a requirements for this PR :P

PR #4774, opened 2021-01-03
Kernel: Make GenericInterruptHandler(s) Ref-counted

supercomputer7:
This is another step into supporting hot-pluggable hardware (for example, AHCI & NVMe support hot-plugging).
I'm still not very sure how to handle everything correctly in an SMP mode (didn't enable that to test it).

cc @kling @tomuta for suggestions :)

### DIFF
--- a/Kernel/Arch/i386/CPU.cpp
+++ b/Kernel/Arch/i386/CPU.cpp
@@ -30,6 +30,7 @@
 #include <AK/Types.h>
 #include <Kernel/Arch/i386/CPU.h>
 #include <Kernel/Arch/i386/ISRStubs.h>
+#include <Kernel/Arch/i386/InterruptArray.h>
 #include <Kernel/Arch/i386/ProcessorInfo.h>
 #include <Kernel/IO.h>
 #include <Kernel/Interrupts/APIC.h>
@@ -56,8 +57,6 @@ namespace Kernel {
 
 static DescriptorTablePointer s_idtr;
 static Descriptor s_idt[256];
-
-static GenericInterruptHandler* s_interrupt_handler[GENERIC_INTERRUPT_HANDLERS_COUNT];
 
 // The compiler can't see the calls to these functions inside assembly.
 // Declare them, to avoid dead code warnings.
@@ -559,8 +558,8 @@ static void unimp_trap()
 
 GenericInterruptHandler& get_interrupt_handler(u8 interrupt_number)
 {
-    ASSERT(s_interrupt_handler[interrupt_number] != nullptr);
-    return *s_interrupt_handler[interrupt_number];
+    ASSERT(!InterruptArray::the().interrupt_handler(interrupt_number).is_null());
+    return *InterruptArray::the().interrupt_handler(interrupt_number);
 }
 
 static void revert_to_unused_handler(u8 interrupt_number)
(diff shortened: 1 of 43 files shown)


supercomputer7:
@tomuta, I think the current changes are too complicated and what I suggest to do is:
1. Move all raw pointers outside of `Kernel/Arch/i386/CPU.cpp` to `InterruptManagement` class, also, move all registering & unregistering methods to the `InterruptManagement` class.
2. Make a fixed `Array`, with `WeakPtr`s of all `GenericInterruptHandler`s. I;m not sure it's the right thing to do, maybe a raw pointer can be still used.
3. Make a fixed `Array` with same size of the above array, with `RecursiveSpinLock`s.
4. When a processor tries to register, unregister or call `handle_interrupt` of a `GenericInterruptHandler` in the specifed `interrupt_vector`, it has to acquire the Spin-lock with `ScopedSpinLock`. This way, if for example CPU0 tries to call `handle_interrupt` of IRQ12 handler, but CPU1 is in the process of removing it because removal of device, then CPU0 will wait until it can acquire the lock. One problem with this approach is to exclude interrupts like the `LocalAPICTimer`, but we can probably come up with a way to bypass the lock explicitly in this case. 

What do you say?

tomuta:
1. We should keep the IDT in CPP.cpp, and at least the immediate code messing with it, as it is i386 specific. 
2. Thinking about it some more, I think we should actually not use `WeakPtr` because it briefly needs to acquire a lock. But we can use the same atomic pointer logic as in `WeakLink::strong_ref` and `WeakLink::revoke` which is entirely lock-free to acquire, but revokation is potentially blocking. This would mean that running interrupts would never involve any block, but uninstalling an interrupt could briefly block if it is being used.
3. I don't really like the `RecursiveSpinLock` in there because only one CPU at at time can hold it. Which, for the LocalAPIC at least would be a disaster as this is running on each CPU potentially concurrently
4. See 2)

supercomputer7:
> 1. We should keep the IDT in CPP.cpp, and at least the immediate code messing with it, as it is i386 specific.
> 2. Thinking about it some more, I think we should actually not use `WeakPtr` because it briefly needs to acquire a lock. But we can use the same atomic pointer logic as in `WeakLink::strong_ref` and `WeakLink::revoke` which is entirely lock-free to acquire, but revokation is potentially blocking (Basically, the LSB bit of the ref count means it's been revoked). This would mean that running interrupts would never involve any block, but uninstalling an interrupt could briefly block if it is being used.
> 3. I don't really like the `RecursiveSpinLock` in there because only one CPU at at time can hold it. Which, for the LocalAPIC at least would be a disaster as this is running on each CPU potentially concurrently
> 4. See 2)

1. All specific i386 code remains at `CPU.cpp` :)
2. Hmm, so you want to implement a special `WeakPtr` (without locking) to hold Interrupt handlers?
3. I already came up with a solution to this - I created `ScopedInterruptSpinLock`, which briefly locks the the interrupt handler, but if it's a global interrupt handler,it releases the lock. It's a somewhat compromise, because I didn't want to have a `Vector` with values of global interrupts, and also didn't want to call the virtual method with a chance of a race condition (so someone else changes the object while I check it).

All relevant changes can be seen here:
https://github.com/supercomputer7/serenity/commit/ad42d3e474c291f123a1c3839050b83cc1f8d3b5

tomuta:
It seems overly complicated. like the simplicity of what we do in `WeakLink` and it's entirely lock-free except to revoke it. It's essentially just a bump of the ref count, and if the LSB was set it means that it can't use it. If the LSB wasn't set, instead of `try_ref`/`adopt` we'd just call the interrupt handler. If the LSB was set (meaning it is being deleted) then we don't and just drop the ref count again.

tomuta:
Deleting basically just sets the LSB and if the ref count is `!= 0` it spins until it drops to `0`.

supercomputer7:
Let's do it correctly then, and I think I have a better idea what to do instead.

PR #4783, opened 2021-01-04
AK: Add default memory order as template argument for Atomic<T>

tomuta:
This is useful for collecting statistics, e.g.
Atomic<unsigned, MemoryOrder::memory_order_relaxed> would allow
using operators such as ++ to use relaxed semantics throughout
without having to explicitly call fetch_add with the memory order.

### DIFF
--- a/AK/Atomic.h
+++ b/AK/Atomic.h
@@ -31,15 +31,6 @@
 
 namespace AK {
 
-enum MemoryOrder {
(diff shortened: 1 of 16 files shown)


jayv:
Nice to see alignment with `std::atomic...`

tomuta:
> Nice to see alignment with `std::atomic...`

`std::atomic` doesn't allow changing the default memory order. But yes, it's heavily inspired by `std::atomic` ;-)

PR #4785, opened 2021-01-04
LibLine: Don't overwrite stuff when moving origin around

alimpfard:
This fixes an issue (mainly) with multiline prompts, where a multiline
prompt would overwrite the lines before it when libline tries to display
it.
To reproduce, set `PROMPT="a\nb\nc> "` in the shell, then press return
a few times.

before:
shot-2021-01-04_16-41-02

after:
shot-2021-01-04_16-39-16

### DIFF
--- a/Libraries/LibLine/Editor.cpp
+++ b/Libraries/LibLine/Editor.cpp
@@ -545,8 +545,15 @@ auto Editor::get_line(const String& prompt) -> Result<String, Editor::Error>
 
     set_prompt(prompt);
     reset();
-    set_origin();
     strip_styles(true);
+
+    auto prompt_lines = max(current_prompt_metrics().line_lengths.size(), 1ul) - 1;
+    for (size_t i = 0; i < prompt_lines; ++i)
+        putc('\n', stderr);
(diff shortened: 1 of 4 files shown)


Libraries/LibLine/Editor.cpp
```diff
@@ -856,8 +863,7 @@ void Editor::handle_read_event()
             }
             reverse_tab = false;
 
-            auto completion_mode = m_times_tab_pressed == 1 ? SuggestionManager::CompletePrefix : m_times_tab_pressed == 2 ? SuggestionManager::ShowSuggestions
-                                                                                                                           : SuggestionManager::CycleSuggestions;
```
alimpfard, line 860: ???
When did this style change happen?

linusg: Greetings from clang-format-11, it's probably telling you to not nest ternary operators :P

alimpfard: Guess it's finally time to delete that clang-format-10-static binary then~

PR #4797, opened 2021-01-04
Playground: Try to keep cursor location when formatting GML

linusg:
Previously we would always reset the cursor to the start caused by `set_text()`. Now we try to keep the cursor as close to the original position, which still doesn't always make sense but it's a start.

### DIFF
--- a/DevTools/Playground/main.cpp
+++ b/DevTools/Playground/main.cpp
@@ -335,7 +335,26 @@ int main(int argc, char** argv)
         }
         auto formatted_gml = GUI::format_gml(source);
         if (!formatted_gml.is_null()) {
+            if (formatted_gml.is_empty()) {
+                editor.clear();
+                return;
+            }
+            // FIXME: This could be improved, but not moving the cursor in most cases is already better than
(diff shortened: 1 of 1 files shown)


alimpfard:
A possibly more sensible approach would be to track the token positions and restore to the closest token's new position (you could take a look at `Shell/Formatter.cpp` for one way of doing that, but that one tracks positions from the AST nodes)

linusg:
the absence of comment tokens in the AST and therefore the formatted source makes this difficult :)
we do not even have proper AST classes for GML, it's all just a bunch of JSON lol

alimpfard:
Yeah, the fact that all references to the original data are lost makes this quite a bit more difficult than it needs to be.
still though, it's possible if you accept the spaghetti:
- determine the closest token to the cursor (optionally also record the offset into the token)
- record where in the JSON that token influences (I'll advertise `JsonPath` for this)
- when generating the output for that specific path, note the offset into the output
- return the recorded offset when done

awesomekling:
Also an option: parse the GML into an AST ;)

linusg:
ok, let's do it properly then :)

PR #4807, opened 2021-01-05
SystemMonitor: Beef up CPU and memory graphs

tomuta:
We now show the total CPU usage as well as the kernel portion.
For the memory graphs we show the amount of committed memory,
actually allocated memory, and the portion of the kernel heap.

_I don't really like how the legend is difficult to read if the graph is right underneath the text, maybe someone else can find a nice solution..._

Screenshot from 2021-01-04 20-14-34

### DIFF
--- a/Applications/SystemMonitor/GraphWidget.cpp
+++ b/Applications/SystemMonitor/GraphWidget.cpp
@@ -27,6 +27,7 @@
 #include "GraphWidget.h"
 #include <LibGUI/Painter.h>
 #include <LibGfx/Font.h>
+#include <LibGfx/Path.h>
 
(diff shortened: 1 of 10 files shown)


Quaker762:
> 
> 
> We now show the total CPU usage as well as the kernel portion.
> For the memory graphs we show the amount of committed memory,
> actually allocated memory, and the portion of the kernel heap.
> 
> _I don't really like how the legend is difficult to read if the graph is right underneath the text, maybe someone else can find a nice solution..._
> 
> Screenshot from 2021-01-04 20-14-34

Hot Damn that looks cool.

smopucilowski:
akrellm when? :^)

Issue #4808, opened 2021-01-05
DynamicLoader: Re-enable lazy PLT symbol resolution


ADKaster:
As part of getting dynamic linking working globally, lazy resolution of R386_JMP_SLOT relocations was disabled in favor of always having "bind now" semantics.

There are some bugs in the implementation that need to be investigated and fixed to re-enable it.

Fixing this will likely help with #4609.

Issue #4609, opened 2020-12-28
[regression] longer boot time


diversys:
Last working build on macOS took 2 seconds to boot. Now that the build is fixed boot time is about 14 seconds.

sppmacd:
It may be related to #4490

ccapitalK:
I believe that this is more related to the Loader application being relatively slow. Application startup times have also regressed in general, with a large amount of the time needed to start up being taken by symbol resolution in the dynamic loader. We probably need to spend some time optimizing it.

awesomekling:
cc @itamar8910 @ADKaster 

Seeing a lot of time spent in ELF hashing. Perhaps we can recover some of the lost performance by using the GNU hash format instead of the SYSV one.

ADKaster:
This article seems like a good reference. 
https://blogs.oracle.com/solaris/gnu-hash-elf-sections-v2

Not sure I fully understand what a bloom filter is, but enabling the GNU hash "should" just be some linker options and implementing the two phase algorithm. Seems there's certain classes of symbols that it's known we won't find in a particular object (such as undef ones or ones with a PLT entry) and the GNU hash will tell you right away if you shouldn't bother walking the hash chain.

ccapitalK:
The GNU Hashing helped quite a bit, but I believe there is still room for improvement. I think we should re-enable lazy PLT symbol resolution, and try to optimize the relocation path a bit more.

diversys:
For me nothing changed subjectively.

ccapitalK:
> For me nothing changed subjectively.

Hmm, thats odd, I noticed roughly a 20% improvement on my machine, though I was mostly measuring individual application startup time instead of overall boot time. I do agree that it's still nowhere near the speed that it was before when we had static linking.

By the way, can you check the output of `objdump -h` when run on some of the shared objects and executables in the Build directory? The change I made to the CMakelists file was a bit finnicky, I want to make sure it is actually generating gnu hash sections on other people's machines. If the `.gnu.hash` section is there, it should be fine.

diversys:
```
objdump -h Applications/About/About

Applications/About/About:	file format ELF32-i386

Sections:
Idx Name          Size      Address          Type
  0               00000000 0000000000000000
  1 .interp       00000013 00000000000000d4 DATA
  2 .gnu.hash     0000004c 00000000000000e8
  3 .dynsym       00000290 0000000000000134
  4 .dynstr       000005ed 00000000000003c4
[...]

PR #4815, opened 2021-01-05
LibELF: Implement GNU Hashing algorithm for dynamic relocation table

ccapitalK:
This implements the gnu hashing algorithm, as an attempt to mitigate #4609. After running a few applications with this enabled, it appears that this doesn't cut down startup times massively as was hoped, but it does help a bit. I believe that the real performance improvements will come from re-enabling lazy PLT symbol resolution as per #4808.

As a side-note, this doesn't enable gnu hashing for every dynamic object in the SerenityOS build. libgcc_s.so still uses SYSV hashing, and I'm not sure how to change that, so the SYSV hashing code paths are still used for it in userspace programs.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -116,7 +116,8 @@ endforeach()
 
 set(CMAKE_INSTALL_NAME_TOOL "")
 set(CMAKE_SHARED_LIBRARY_SUFFIX ".so")
-set(CMAKE_SHARED_LIBRARY_CREATE_CXX_FLAGS "-shared")
+set(CMAKE_SHARED_LIBRARY_CREATE_CXX_FLAGS "-shared -Wl,--hash-style=gnu")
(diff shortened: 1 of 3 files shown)


ccapitalK:
There is still optimization room here, one thing I notice is that when performing a global lookup of a symbol, we have to calculate the hash O(n) times where n is the number of dynamic objects present, and these hash calculations may take a non-trivial amount of time when we have very long mangled symbol names (for example, `_ZNSt7__cxx1115basic_stringbufIcSt11char_traitsIcESaIcEE8_M_pbumpEPcS5_x`). We could probably try to hoist this hash calculation out when we detect that all objects use gnu hash, to make maximum use of the bloom filters.

Doing this would require us to build libgcc_s.so with a gnu hash for the optimization to ever actually do anything however, which I believe will require changes to the toolchain.

awesomekling:
Very cool :)

PR #4816, opened 2021-01-05
LIBGUI: Add GML properties to calendar widget

glenfordwilliams:
This pr introduces additional GML properties to the calendar widget.

- grid -bool
- date - string/array
- mode -  enum

image

```
@GUI::Widget{

    @GUI::Widget{
        layout: @GUI::HorizontalBoxLayout{}
        @GUI::Widget {
            //October 10 2020
            @GUI::Calendar{
                height : 200
                width: 300
                date: [10,10,2020]
            }
        }
      @GUI::Widget {
                //October 10 2020
                @GUI::Calendar{
                   height : 200
                   width: 300
                   date: "2020-10-10"
                }
            }
    }
    @GUI::Widget{
        layout: @GUI::HorizontalBoxLayout{}

        @GUI::Widget {
            //October 1 2020
            @GUI::Calendar{
               height : 200
               width: 300
               grid: false
            }
        }
        @GUI::Widget {
            //October 1 2020
            @GUI::Calendar{
               height : 200
               width: 300
               mode: "Month"
            }
        }
    }
}
```

edited  to show updated purpose

### DIFF
--- a/Libraries/LibCore/DateTime.cpp
+++ b/Libraries/LibCore/DateTime.cpp
@@ -27,8 +27,8 @@
 #include <AK/StringBuilder.h>
 #include <AK/Time.h>
 #include <LibCore/DateTime.h>
+#include <ctype.h>
 #include <sys/time.h>
-#include <time.h>
 
 namespace Core {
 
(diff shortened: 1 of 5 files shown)


alimpfard:
> also introduces a @GUI::GMLObject class for setting key value pairs to class properties

This...sounds a bit counterproductive? allowing arbitrary stuff (and expecting arbitrary stuff) is rather bad for a declarative language imo
Perhaps consider a `GUI::Calendar::Date`, or a more generic `GML::Date` object?

glenfordwilliams:
> > also introduces a @GUI::GMLObject class for setting key value pairs to class properties
> 
> This...sounds a bit counterproductive? allowing arbitrary stuff (and expecting arbitrary stuff) is rather bad for a declarative language imo
> Perhaps consider a `GUI::Calendar::Date`, or a more generic `GML::Date` object?

Is it that bad an idea?

alimpfard:
Bad? Not really, but imagine if everything accepted arbitrary objects and read out random fields from them
bye bye reliable autocomplete and static analysis

glenfordwilliams:
Oh right, didn't really think along those lines. Will have to rethink my approach then

glenfordwilliams:
updated to use a `@GUI::GMLDate` with properly defined properties. Api still feels clumsy but for now it should work until there is a better abstraction for components outside of the GUI namespace

awesomekling:
I'm not sure what's going on here, but GML is intended for UI declarations, not arbitrary object instantiation :)

awesomekling:
If you want to assign an initial date to a `GUI::Calendar` widget from GML, make it accept a date string. There's no need for special objects.

Also please don't add widget subclasses that are not actually widgets. :)

glenfordwilliams:
@awesomekling  updated to use  just a optional string/array for setting the date

awesomekling:
I don't really understand the purpose of this PR. This seems like a large amount of complexity with very little benefit. You are not even using the property for anything in the end?

awesomekling:
Also, please observe the [contribution guidelines](https://github.com/SerenityOS/serenity/blob/08190dd0ce5f23de6dffcd2c7521314202b05514/CONTRIBUTING.md) and squash fixups into earlier commits instead of stacking them.

glenfordwilliams:
@awesomekling  my apologies, i updated my top comment to make things clearer. I took into account your previous comments and updated my code

awesomekling:
Again, this seems like a large amount of complexity with very little benefit.

There's no code change here that makes use of the new property. What is this actually needed for?

Also, supporting both string and array properties seems excessive.

glenfordwilliams:
> Again, this seems like a large amount of complexity with very little benefit.
> 
> There's no code change here that makes use of the new property. What is this actually needed for?
> 

Which of the properties are we talking about here?

> Also, supporting both string and array properties seems excessive.
personally   thought it would be a useful thing to to, but since you are in disagreement i can remove the array option

linusg:
@glenfordwilliams this PR has various merge conflicts, please rebase :)

> > There's no code change here that makes use of the new property. What is this actually needed for?
>
> Which of the properties are we talking about here?

The `grid` and `mode` properties on `GUI::Calendar`.

awesomekling:
Closing this since it's basically adding dead code and increasing complexity for no reason.

Issue #4801, opened 2021-01-04
TextEditor: Should put a newline at the end of files


awesomekling:
When saving with TextEditor, the file should end in a newline (unless it's completely empty.)

PR #4820, opened 2021-01-06
TextEditor: puts a newline at the end of files

samkravitz:
- Doesn't write anything to empty files (though the file is still created)
 - Puts a newline at the end of files that do not have one
 - Does not put a newline at the end of files that already end in one

Addresses issue https://github.com/SerenityOS/serenity/issues/4801

### DIFF
--- a/Libraries/LibGUI/TextEditor.cpp
+++ b/Libraries/LibGUI/TextEditor.cpp
@@ -1055,6 +1055,11 @@ bool TextEditor::write_to_file(const StringView& path)
         perror("open");
         return false;
     }
+    
+    if (file_size == 0) {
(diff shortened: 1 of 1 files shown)


awesomekling:
Please observe the [contribution guidelines](https://github.com/SerenityOS/serenity/blob/edeec2974fa9cfbfaea6f70b2b940c950fb6ff05/CONTRIBUTING.md)
Specifically, squash your commits instead of stacking fixups.

Wrt the changes, do we really need two separate branches where we write out a `\n` character?

awesomekling:
Fixed this in 7f61644 since the PR was abandoned.

PR #4826, opened 2021-01-06
Give cp a `-l` flag, and let `ls -l` print hard link count

nico:
Inspired by frustration with macOS (https://github.com/nico/hack/blob/master/notes/copydir.md).

### DIFF
--- a/Userland/ls.cpp
+++ b/Userland/ls.cpp
@@ -288,6 +288,8 @@ static bool print_filesystem_object(const String& path, const String& name, cons
     else
         printf("%c", st.st_mode & S_IXOTH ? 'x' : '-');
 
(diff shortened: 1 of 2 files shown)


Userland/ls.cpp
```diff
@@ -288,6 +288,8 @@ static bool print_filesystem_object(const String& path, const String& name, cons
     else
         printf("%c", st.st_mode & S_IXOTH ? 'x' : '-');
 
+    printf(" %d", st.st_nlink);
```
awesomekling, line 291: `st_nlink` is unsigned :)

nico: tenor

Also, done.

Issue #3691, opened 2020-10-05
LibHTTP: Loading google analytics over HTTPS crashes with m_state == State::InBody assertion failure


Lubrsi:
Loading https://google-analytics.com/analytics.js crashes with this:
```
USERSPACE(117) ASSERTION FAILED: m_state == State::InBody
../Libraries/LibHTTP/Job.cpp:172
[ProtocolServer(117:117)]: Terminating ProtocolServer(117) due to signal 6
[FinalizerTask(3:3)]: CPU[0] NP(error) fault at invalid address V0x01c0260c
[FinalizerTask(3:3)]: copy_from_user(0xc094a9ac, 0x01c0260c, 4) failed at V0x01c0260c
[FinalizerTask(3:3)]: 0xdeadc0de  Kernel::Processor::switch_context(Kernel::Thread*&, Kernel::Thread*&) +181
0xdeadc0de  Kernel::Scheduler::context_switch(Kernel::Thread*) +181
0xdeadc0de  Kernel::Scheduler::pick_next() +1952
0xdeadc0de  Kernel::Processor::check_invoke_scheduler() +125
0xdeadc0de  Kernel::Thread::die_if_needed() +157
0xdeadc0de  syscall_handler +1414
0xdeadc0de  syscall_asm_entry +49
0x0807af7b  raise +22
0x080480bd  abort +13
0x08077de1  __cxa_atexit +0
0x0805f19f  AK::Function<void ()>::CallableWrapper<HTTP::Job::on_socket_connected()::{lambda()#2}>::call() const +1295
0x0805fe27  TLS::TLSv12::read_from_socket() +99
0x08050599  Core::EventLoop::pump(Core::EventLoop::WaitMode) +265
0x08050743  Core::EventLoop::exec() +43
0x080485af  main +384
0x08048a1e  _start +94
```

It loads fine over regular HTTP.

alimpfard:
Been a while since I've seen a LibHTTP bug
Exciting!

alimpfard:
Hmmm...I can't reproduce this, `pro https://google-analytics.com/analytics.js -O` succeeds and produces an ok-looking file.
opening it in the browser seems to be fine too, it loads and then "Failed to parse content".

that host resolves to 172.217.16.164 for me, not sure if that's significant.
the request produced via `pro` is:
```
GET /analytics.js HTTP/1.1
Host: google-analytics.com
Connection: close
```

Lubrsi:
[...]

PR #4831, opened 2021-01-06
LibHTTP: Take the final EOF read notification into account

alimpfard:
Fixes the HTTP part of #3691.

### DIFF
--- a/Libraries/LibHTTP/Job.cpp
+++ b/Libraries/LibHTTP/Job.cpp
@@ -125,6 +125,17 @@ void Job::on_socket_connected()
     register_on_ready_to_read([&] {
         if (is_cancelled())
             return;
+
+        if (m_state == State::Finished) {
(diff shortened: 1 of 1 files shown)


Libraries/LibHTTP/Job.cpp
```diff
@@ -125,6 +125,17 @@ void Job::on_socket_connected()
...
         if (is_cancelled())
             return;
+
+        if (m_state == State::Finished) {
+            // This is probably just a EOF notification, which means we should receive nothing
+            // and then get eof() == true.
+            [[maybe_unused]] auto payload = receive(64);
+            // These assertions are only correct if "Connection: close".
+            ASSERT(payload.is_empty());
+            ASSERT(eof());
```
awesomekling, line 135: We should aim towards handling this more gracefully at some point, but for now this is a progression. :)

awesomekling:
Thank you!

PR #4833, opened 2021-01-06
Utilities: Add a disk space usage analyzation program.

Maato:
space_analyzer_screenshot


Applications/SpaceAnalyzer/main.cpp
```diff
@@ -0,0 +1,308 @@
...
+    app_menu.add_action(GUI::Action::create("Analyze", [&](auto&) {
+        analyze(tree, treemapwidget, statusbar);
+    }));
+    app_menu.add_action(GUI::CommonActions::make_quit_action([&](auto&) {
+        app->quit();
+    }));
+    auto& help_menu = menubar->add_menu("Help");
+    help_menu.add_action(GUI::Action::create("About", [&](auto&) {
+        GUI::AboutDialog::show(APP_NAME, 0, window);
+    }));
```
awesomekling, line 282: Please use `GUI::CommonActions::make_about_action()`.

Applications/SpaceAnalyzer/TreeMapWidget.h
```diff
@@ -0,0 +1,77 @@
...
+
+private:
+    TreeMapWidget();
+    virtual void paint_event(GUI::PaintEvent&) override;
+    virtual void mousedown_event(GUI::MouseEvent&) override;
+    virtual void doubleclick_event(GUI::MouseEvent&) override;
+    virtual void mousewheel_event(GUI::MouseEvent&) override;
+
+    bool area_can_contain_children(Gfx::IntRect ext);
+    bool area_can_contain_label(Gfx::IntRect ext);
```
awesomekling, line 66: These should be `const` qualified. Also please pass large argument types such as `Gfx::IntRect` by const reference.

Applications/SpaceAnalyzer/TreeMapWidget.h
```diff
@@ -0,0 +1,77 @@
...
+    void paint_cell(GUI::Painter*, ITreeMapNode&, Gfx::IntRect ext, int depth, Gfx::IntPoint query_pos, Vector<int>& out_path);
+    void paint_node(GUI::Painter*, ITreeMapNode&, Gfx::IntRect ext, int depth, Gfx::IntPoint query_pos, Vector<int>& out_path);
+    Gfx::IntRect paint_cell_frame(GUI::Painter*, ITreeMapNode&, Gfx::IntRect ext, int depth);
```
awesomekling, line 69: Please pass `GUI::Painter` by reference, since `nullptr` is not a valid value for it.
Also, please pass `Gfx::IntPoint` by const reference.
Logically speaking, it feels weird for these to take the treemap node by mutable reference. I guess it's because you may need to sort it. Since the sorting is essentially a lazy operation, it might make sense to make sorting const-qualified and pass `const ITreeMapNode&` to these.

Applications/SpaceAnalyzer/TreeMapWidget.h
```diff
@@ -0,0 +1,77 @@
...
+};
+
+class TreeMapWidget final : public GUI::Frame {
+    C_OBJECT(TreeMapWidget)
+public:
+    virtual ~TreeMapWidget() override;
+    Function<void()> on_path_change;
+    size_t path_size();
+    ITreeMapNode* get_nth_path_node(size_t n);
+    size_t get_viewpoint();
```
awesomekling, line 54: These should be `const`.

Applications/SpaceAnalyzer/TreeMapWidget.h
```diff
@@ -0,0 +1,77 @@
...
+
+#include <LibGUI/Frame.h>
+#include <LibGfx/Rect.h>
+
+namespace TreeMap {
+
+struct ITreeMapNode {
+    virtual String name() = 0;
+    virtual int64_t area() = 0;
+    virtual size_t num_children() = 0;
```
awesomekling, line 37: These should be `const`.

Base/res/apps/SpaceAnalyzer.af
```diff
@@ -0,0 +1,4 @@
+[App]
+Name=SpaceAnalyzer
+Executable=/bin/SpaceAnalyzer
+Category=Utilities
```
awesomekling, line 4: Please configure your editor to put a newline at the end of files.

Applications/SpaceAnalyzer/main.cpp
```diff
@@ -0,0 +1,308 @@
...
+        return;
+    }
+
+    auto content = df->read_all();
+    auto json = JsonValue::from_string(content);
+    ASSERT(json.has_value());
+
+    json.value().as_array().for_each([&output](auto& value) {
+        auto fs_object = value.as_object();
+        MountInfo mi;
```
awesomekling, line 101: Please use full words instead of abbreviations in variable names. (See [coding style](https://github.com/SerenityOS/serenity/blob/master/Documentation/CodingStyle.md#names))

Applications/SpaceAnalyzer/TreeMapWidget.h
```diff
@@ -0,0 +1,77 @@
...
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#pragma once
+
+#include <LibGUI/Frame.h>
+#include <LibGfx/Rect.h>
+
+namespace TreeMap {
```
awesomekling, line 32: Please put this in the application's namespace since it's not part of a `LibTreeMap` :)

Applications/SpaceAnalyzer/main.cpp
```diff
@@ -0,0 +1,308 @@
...
+#include <LibGUI/Menu.h>
+#include <LibGUI/MenuBar.h>
+#include <LibGUI/StatusBar.h>
+#include <sys/stat.h>
+
+static const char* APP_NAME = "SpaceAnalyzer";
+
+struct TreeNode : public TreeMap::ITreeMapNode {
+    TreeNode(String name)
+        : m_name(name)
```
awesomekling, line 46: You can `move()` the `name` here.

Applications/SpaceAnalyzer/main.cpp
```diff
@@ -0,0 +1,308 @@
...
+        , m_area(0)
+        , m_children(nullptr) {};
```
awesomekling, line 48: Please initialize these as part of the member declaration instead of using the constructor initializer list.

Applications/SpaceAnalyzer/main.cpp
```diff
@@ -0,0 +1,308 @@
...
+    String m_name;
+    int64_t m_area;
+    OwnPtr<Vector<TreeNode>> m_children;
```
awesomekling, line 51: Is there a meaningful reason to allocate these vectors on the heap instead of making it simply `Vector<TreeNode> m_children;`?

Applications/SpaceAnalyzer/main.cpp
```diff
@@ -0,0 +1,308 @@
...
+    {
+        if (m_children != nullptr) {
+            quick_sort(*m_children, [](auto& a, auto& b) { return b.m_area < a.m_area; });
+        }
+    }
+};
+
+struct Tree : public TreeMap::ITreeMap {
+    Tree(String root_name)
+        : m_root(root_name) {};
```
awesomekling, line 72: You can `move()` the `root_name` here.

Applications/SpaceAnalyzer/main.cpp
```diff
@@ -0,0 +1,308 @@
...
+        MountInfo mi;
+        mi.mount_point = fs_object.get("mount_point").to_string();
+        mi.source = fs_object.get("source").as_string_or("none");
+        output.append(mi);
+    });
+}
+
+struct QueueEntry {
+    String path;
+    TreeNode* node;
```
awesomekling, line 110: Please always initialize primitive members, e.g `TreeNode* node { nullptr };`

Applications/SpaceAnalyzer/main.cpp
```diff
@@ -0,0 +1,308 @@
...
+            result += update_totals(node.m_children->at(k));
+        }
+        node.m_area = result;
+    } else {
+        result = node.m_area;
+    }
+    return result;
+}
+
+static int find_mount_for_path(String path, Vector<MountInfo>& mounts)
```
awesomekling, line 127: It could be nicer to return a `MountInfo*` here instead of an integer with a magic interpretation of `-1`.

Applications/SpaceAnalyzer/main.cpp
```diff
@@ -0,0 +1,308 @@
...
+    }
+
+    update_totals(root);
+}
+
+static void analyze(RefPtr<Tree> tree, TreeMap::TreeMapWidget& treemapwidget, GUI::StatusBar& statusbar)
+{
+    // Build an in-memory tree mirroring the filesystem and for each node
+    // calculate the sum of the file size for all its descendants.
+    //NonnullOwnPtr<TreeNode> root = make<TreeNode>(path);
```
awesomekling, line 215: Please don't add commented out code.

awesomekling (changes requested):
This looks super cool!

Applications/SpaceAnalyzer/TreeMapWidget.h
```diff
@@ -0,0 +1,77 @@
...
+
+struct ITreeMapNode {
```
alimpfard, line 34: nitpick: can we not `IJavaism`?

Applications/SpaceAnalyzer/main.cpp
Maato: I tried this after your suggestion, but it doesn't work because Vector<T> does sizeof(T) here: https://github.com/SerenityOS/serenity/blob/ddc255a6c0026af70a7ca685c2fed24941e7df9c/AK/Vector.h#L608
But T is an incomplete type at that point.

My original reasoning however, was to reduce the total size of the tree. I guessed (did not measure) that most nodes will be leaf nodes. Having a nullptr in those nodes would be smaller than an empty Vector.

Applications/SpaceAnalyzer/TreeMapWidget.h
Maato: Could you suggest an alternative that is more in line with serenity's code base?
I've seen Abstract as a prefix, but I think that doesn't apply since ITreeMapNode does not have any code or members.

alimpfard: I'd personally just call it `TreeMapNode`.
the concept of "interface" doesn't really exist in C++ anyway

Applications/SpaceAnalyzer/TreeMapWidget.h
Maato: Not sure what to do about this, the reason I use GUI::Painter *, is that these functions also get called by event handlers for mousedown and doubleclick in which case there is no painter.

I also either don't understand the last part of your comment or simply don't know how to do it. (Making sorting const-qualified that is.)

awesomekling: > Not sure what to do about this, the reason I use GUI::Painter *, is that these functions also get called by event handlers for mousedown and doubleclick in which case there is no painter.

Aha! Well that's extremely confusing :) Something called "paint_something" should not be called by a mouse event handler, since there's no painting happening.

Can we split those functions up into helpers that both paint and mouse events can call, without having to pass a painter in?

> I also either don't understand the last part of your comment or simply don't know how to do it. (Making sorting const-qualified that is.)

You are currently passing in the node as a `TreeMapNode&`. This looks very suspicious, as painting, logically speaking, is a read-only operation. So I would expect it to take `const TreeMapNode&`.

Applications/SpaceAnalyzer/main.cpp
awesomekling: Oh duh, my bad, I totally missed that it's a vector of TreeNode *inside* TreeNode :)

Applications/SpaceAnalyzer/TreeMapWidget.h
Maato: The layout algorithm is now split out into its own function, it is called by paint and mouse event handlers.
The lay_out_children function still takes a non const reference to TreeMapNode, because it needs to call sort_children_by_area. 

You mentioned making sorting const-qualified, I don't know how to do that.

Options I can think of:
- Do all sorting during tree construction, that would mean sorting nodes that are potentially never displayed (i.e. wasted work).
- Do sorting during layout into temporary storage, means no performance gains on subsequent layout calls.
- Keep as is.

awesomekling: You can make sorting const-qualified and then `const_cast<TreeMapWidget*>(this)->...` for mutable access inside `sort_children_by_area()` :)

awesomekling:
:star:

PR #4834, opened 2021-01-06
Userland: Rename cp's "-r" flag to "-R"

nico:
Linux accepts both -r and -R, but the BSDs only like -R, and
dR POSIX also only mentions -R. So make -R the canonical flag.

Keep -r available as an alias for -R.

### DIFF
--- a/Userland/cp.cpp
+++ b/Userland/cp.cpp
@@ -57,7 +57,8 @@ int main(int argc, char** argv)
 
     Core::ArgsParser args_parser;
     args_parser.add_option(link, "Link files instead of copying", "link", 'l');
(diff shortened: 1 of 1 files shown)


ccapitalK:
This would make it more compliant, but I feel like it would mess with muscle memory for most people, which IMO would make the terminal experience worse. I'm kind iffy on whether we should do this, I think it would be better to just accept both.

alimpfard:
How about accepting both and adding an extension (`--arr` maybe?) on top?

awesomekling:
Accepting both `-r` and `-R` seems like the way to go.

nico:
TIL that both help string and long flag name can be nullptr.

PR #4839, opened 2021-01-07
LibLine: Fix "^R<return>^C" / Shell: Tiny completion fix

alimpfard:


Libraries/LibCore/EventLoop.cpp
```diff
@@ -438,6 +440,14 @@ EventLoop::SignalHandlers::~SignalHandlers()
...
 void EventLoop::SignalHandlers::dispatch()
 {
+    Vector<int> removed;
+    for (auto& id : s_removed_signal_handler_ids)
+        if (m_handlers.remove(id))
+            removed.append(id);
+
+    for (auto& id : removed)
+        s_removed_signal_handler_ids.remove(id);
+
```
alimpfard, line 450: @tomuta I don't know if this is supposed to be this way, but unregistering signal handlers is super unreliable without this hackjob.
Also take a look at `DeletableBool` (ouch)

alimpfard:
On an unrelated sidenote, I wonder if the signal handlers not going away were at all related to Shell-backgrounding?
Let's try provisionally reenabling backgrounding :thinking:

alimpfard:
Hmmmmmmmmmmmmmmmmmmmmmmmmmmmm

I'll run it in a loop locally I guess.

Libraries/LibCore/EventLoop.cpp
tomuta: I'm not sure what problem this is trying to solve, but I can see a problem with unregistering a signal handler from within a signal callback. We should probably change `s_signal_handlers` to `HashMap<int, NonnullRefPtr<SignalHandlers>>` and in `EventLoop::dispatch_signal` make sure we bump the reference while calling the handler function. I suspect that this will solve any weird problem you're running into, and it shouldn't require any of these weird hacks.

alimpfard: the problem it's trying to solve is this:
- the signal handlers have static duration
- some object (possibly with static lifetime) registers a signal handler with references into itself
- Naturally, the object will try to unregister the handlers on deletion

if the object has static duration, no unregisteration is needed
if it doesn't, the unregisteration is required (as otherwise the handler will use dangling references.

just making the handlers refcounted won't help this, I don't think

tomuta: And of course that means removing `s_handling_signal` as well...

tomuta: So you're saying the problem is when the global destructors are run? Is there a way for me to repro this?

alimpfard: Sure, apply this diff and run some shell test:
```diff
diff --git a/Libraries/LibLine/Editor.cpp b/Libraries/LibLine/Editor.cpp
index 781fce17a..4beac9a8b 100644
--- a/Libraries/LibLine/Editor.cpp
+++ b/Libraries/LibLine/Editor.cpp
@@ -451,13 +451,13 @@ void Editor::initialize()
     for (auto& keybind : m_configuration.keybindings)
         register_key_input_callback(keybind);

-    Core::EventLoop::register_signal(SIGINT, [this](int) {
+    m_registered_signal_handlers.append(Core::EventLoop::register_signal(SIGINT, [this](int) {
         interrupted();
-    });
+    }));

-    Core::EventLoop::register_signal(SIGWINCH, [this](int) {
+    m_registered_signal_handlers.append(Core::EventLoop::register_signal(SIGWINCH, [this](int) {
         resized();
-    });
+    }));

     m_initialized = true;
 }
diff --git a/Libraries/LibLine/Editor.h b/Libraries/LibLine/Editor.h
index 3106debc8..4fb2596ee 100644
--- a/Libraries/LibLine/Editor.h
+++ b/Libraries/LibLine/Editor.h
@@ -39,6 +39,7 @@
 #include <AK/Utf8View.h>
 #include <AK/Vector.h>
 #include <LibCore/DirIterator.h>
+#include <LibCore/EventLoop.h>
 #include <LibCore/Notifier.h>
 #include <LibCore/Object.h>
 #include <LibLine/KeyCallbackMachine.h>
@@ -320,6 +321,8 @@ private:
     {
         ASSERT(m_initialized);
         tcsetattr(0, TCSANOW, &m_default_termios);
+        for (auto& id : m_registered_signal_handlers)
+            Core::EventLoop::unregister_signal(id);
         m_initialized = false;
     }

@@ -459,6 +462,7 @@ private:
     HashMap<u32, HashMap<u32, Style>> m_anchored_spans_ending;

     RefPtr<Core::Notifier> m_notifier;
+    Vector<int> m_registered_signal_handlers;

     bool m_initialized { false };
     bool m_refresh_needed { false };
```

tomuta: I applied the patch, but how would I trigger the bug? I tried a couple of things and seems to work fine?

alimpfard: Just run a shell test with the lagom version of Shell?
pretty sure it'll do the same inside serenity as well

tomuta: Sorry if I sound stupid, but how exactly do that? Not really familiar with lagom or what tests there may be or how to run them :(

alimpfard: Probably something like this:
```sh
$ cd Build
# assuming you're not building lagom
$ cmake .. -GNinja -DBUILD_LAGOM=1
$ ninja

# run some shell test
$ Meta/Lagom/shell ../Shell/Tests/brace-exp.sh
```

tomuta: That seems to work just fine for me? Not seeing any crashes.

tomuta: Oh I see, so the problem is really that `s_signal_handlers` probably was destructed before being used again...

alimpfard: Yeah.
Sorry, seems like I wasn't clear enough about what the actual problem was :P

alimpfard:
-> #4854 for the LibCore changes.

PR #4840, opened 2021-01-07
Userland: Add ddate utility

bcoles:
Today is Boomtime, day 7 of Chaos, in the YOLD 3187

ddate

### DIFF
--- a/Userland/ddate.cpp
+++ b/Userland/ddate.cpp
@@ -1,0 +1,132 @@
+/*
+ * Copyright (c) 2021, the SerenityOS developers.
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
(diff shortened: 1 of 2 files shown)


Userland/ddate.cpp
```diff
@@ -0,0 +1,132 @@
...
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include <AK/String.h>
+#include <LibCore/DateTime.h>
+#include <stdio.h>
+
+class DiscordianDate {
+public:
+    DiscordianDate(Core::DateTime date)
```
emanuele6, line 33: nitpick: I think this constructor should be `explicit`

awesomekling:
This doesn't seem very useful(?)

bcoles:
> 
> 
> This doesn't seem very useful(?)

I suppose that depends on whether you need to know the Discordian date. If so, this utility exceeds usefulness.

It could be improved by adding command line support for converting arbitrary Gregorian dates to Discordian dates and vice-versa. However, the existing (and arguably more important) `date` utility does not accept formatting user-supplied dates either (yet).

SerenityOS draws heavily on Unix for inspiration and design. The documentation specifically states features including "Unix-like libc and userland".

This utility does one thing and does it well, in line with Unix philosophy. 

`ddate` has been a part of the Unix userland for a long time. Similarly, `ddate` was part of `util-linux` for ~20 years. While not strictly part of the 90s aesthetic, it was installed by default on Linux systems in the 90s, and has cemented its place in computing history. As of early 2010s, it is no longer included in `util-linux`, but is available as an optional package.

Likewise, packages for `ddate` are available for most Unix flavors and derivatives, including BSDs and Solaris, and hobbyist systems such as TempleOS. Availability of packages such as `ddate`, along with `cowsay`, may also be perceived as a milestone for system maturity.

SerenityOS does not yet feature a package manager or repository for additional non-essential software. While a Ports system exists, native applications are preferred over Ports.

The essence of Serenity design is loosely defined as:

* take a classic Unix thing (eg. man pages),
* put a modern spin on it (eg. markdown),
* and display it in 90s-style UI (eg. Help app)

`ddate` satisfies only the first by virtue of longevity. Rendering the Discordian Calendar in a 90s style UI would satisfy both the second and third points. Whoever chooses to take on this task can leverage the `DiscordianDate` class provided in this PR, and possibly also the existing calendar code from the Calendar application.

awesomekling:
That's a respectably high-effort reply right there. All right then :)

PR #4842, opened 2021-01-08
LibGUI+SystemMonitor+MenuApplets: Open the correct tab on click

TheMorc:
I also added arguments parsing to SystemMonitor so that MenuApplets can open SystemMonitor with either Graphs or Network tab. And I'm blind or I just didn't notice any TabWidget active tab switching using a tab index, so I added it there.
I did it like this because I'm not really sure how should I switch between them using the set_active_widget method.

### DIFF
--- a/Applications/SystemMonitor/main.cpp
+++ b/Applications/SystemMonitor/main.cpp
@@ -35,6 +35,7 @@
 #include "ProcessUnveiledPathsWidget.h"
 #include "ThreadStackWidget.h"
 #include <AK/NumberFormat.h>
+#include <LibCore/ArgsParser.h>
 #include <LibCore/Timer.h>
 #include <LibGUI/Action.h>
 #include <LibGUI/ActionGroup.h>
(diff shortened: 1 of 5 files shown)


Applications/SystemMonitor/main.cpp
```diff
@@ -342,6 +349,23 @@ int main(int argc, char** argv)
...
+    else if (args_tab_view == "pci")
+        tabwidget.set_tab_index(3);
+    else if (args_tab_view == "devices")
+        tabwidget.set_tab_index(4);
+    else if (args_tab_view == "network")
+        tabwidget.set_tab_index(5);
+    else if (args_tab_view == "processors")
+        tabwidget.set_tab_index(6);
+    else if (args_tab_view == "interrupts")
+        tabwidget.set_tab_index(7);
```
awesomekling, line 367: Let's not hard-code the index of the tabs like this, or it's easy to make this buggy when adding/removing/moving tabs in the future.
Instead, let's hold on to pointers to the various tab page widgets when we create them and use `set_active_widget()` instead. :)

TheMorc: I got all to work with pointers now but the processes tab which is added using `tabwidget.add_tab` and not `add_widget` has it's pointer done a bit weirdly using `auto processes_tab = tabwidget.active_widget()` because i'm not sure rightnow how can i get a GUI::Widget instead of getting GUI::VerticalSplitter.

And i want to ask, should i squash it _(somehow)_ to the original SystemMonitor commit? _(i still suck at git so i'll rather try messing with squashing commits somewhere else so i don't mess it up later )_

awesomekling: **GUI::VerticalSplitter** inherits from **GUI::Widget**, so the pointers are already compatible :)

TheMorc:
So it should be _hopefully_ fixed now

PR #4851, opened 2021-01-08
UserspaceEmulator: Fix incorrect shadowing on mov sign extend

GalHorowitz:
Unlike zero-extend moves, the upper bytes are not just zeroed, but rather are based on the sign bit of the source, which means if the source is tainted, so should the upper bytes be.


awesomekling:
Oh, that's a great catch! :)

GalHorowitz:
Oh, I accidentally wrote `src.shadow()` instead of `src`. Weird that my IDE didn't complain. Should I open another pull request?

awesomekling:
No worries, I fixed it in 6a970611d65b1d410636f7155de1986021f22267

But please don't open a PR before even compiling your code next time ;)

PR #4852, opened 2021-01-08
Fix debugger after shared libraries change

itamar8910:
LibDebug & its users (sdb, HackStudio and functrace) have been pretty much non-functional ever since the shared libraries change.

This PR makes them function again :)


Libraries/LibELF/Image.h
```diff
@@ -76,6 +76,7 @@ class Image {
         unsigned type() const { return ELF32_ST_TYPE(m_sym.st_info); }
         unsigned bind() const { return ELF32_ST_BIND(m_sym.st_info); }
         const Section section() const { return m_image.section(section_index()); }
+        bool is_undefiend() const { return section_index() == 0; }
```
awesomekling, line 79: Typo, `is_undefined()`

Libraries/LibDebug/DebugInfo.cpp
```diff
@@ -124,6 +127,9 @@ void DebugInfo::prepare_lines()
             auto start_index = file_path.index_of(serenity_slash).value() + serenity_slash.length();
             file_path = file_path.substring(start_index, file_path.length() - start_index);
         }
+        if (file_path.starts_with("./") && !m_source_root.is_null()) {
+            file_path = LexicalPath(String::format("%s/%s", m_source_root.characters(), file_path.characters())).string();
```
awesomekling, line 131: Please use `String::formatted()` in new code. You can also use `LexicalPath::canonicalized_path()`	for a slight boost in expressiveness:
```cpp
file_path = LexicalPath::canonicalized_path(String::formatted("{}/{}", m_source_root, file_path));
```

Libraries/LibDebug/DebugSession.h
```diff
@@ -61,10 +61,26 @@ class DebugSession {
 
     struct BreakPoint {
         void* address;
```
awesomekling, line 63: Maybe be a good boi and initialize this with `{ nullptr }` since we're here.

Libraries/LibELF/DynamicLinker.cpp
```diff
@@ -231,6 +231,13 @@ static NonnullRefPtr<DynamicLoader> commit_elf(const String& name)
 void ELF::DynamicLinker::linker_main(String&& main_program_name, int main_program_fd, int argc, char** argv, char** envp)
 {
     g_envp = envp;
+    bool do_breakpoint_trap_before_entry = false;
+    for (char** env = envp; *env; ++env) {
+        if (StringView { *env } == "_LOADER_BREAKPOINT=1") {
+            do_breakpoint_trap_before_entry = true;
+        }
+    }
```
awesomekling, line 239: I think this particular feature is fine, but we do have to be very careful when using the environment in the dynamic loader, since it will deal with user-provided input to set-uid programs.

itamar8910:
Only fixed some of the comments (please do not merge yet)

awesomekling:
> Only fixed some of the comments (please do not merge yet)

You can mark the PR as a draft to prevent merging ;)

Libraries/LibELF/DynamicLinker.cpp
itamar8910: Agreed. Yeah, seems like the worst things a user could do in this case is make a set-uid program crash because of the uncaught SIGTRAP. The user won't be able to attach to the program because we disallow it ptrace.

ADKaster: There's the AT_SECURE auxiliary value (which is set for setuid programs, I think?), which could be read to set a `s_allowed_to_check_env_vars` flag. I believe that's what other dynamic loaders do.

itamar8910: That's neat, and it is indeed already set for setuid/gid programs.
I added a commit to this PR that does what you described here.

Libraries/LibELF/DynamicLinker.cpp
```diff
@@ -228,9 +231,23 @@ static NonnullRefPtr<DynamicLoader> commit_elf(const String& name)
     return loader;
 }
 
-void ELF::DynamicLinker::linker_main(String&& main_program_name, int main_program_fd, int argc, char** argv, char** envp)
+static void read_environemnt_variables()
```
ADKaster, line 234: Quick typo check: `environemnt` --> `environment`

itamar8910:
Okay @awesomekling, @ADKaster, I believe I addresses all comments.

PR #4854, opened 2021-01-08
LibCore: Allow adding/removing signal handlers while handling signals

tomuta:
This allows adding and removing of asynchronous signal handlers while
executing signal handlers, even if it is for the same signal that is
being handled right now.

@alimpfard In response to #4839 does this resolve the issue(s) you're running into? _Disclaimer: I did not test this code!_

### DIFF
--- a/Libraries/LibCore/EventLoop.cpp
+++ b/Libraries/LibCore/EventLoop.cpp
@@ -30,6 +30,7 @@
 #include <AK/JsonObject.h>
 #include <AK/JsonValue.h>
 #include <AK/NeverDestroyed.h>
+#include <AK/Singleton.h>
 #include <AK/TemporaryChange.h>
 #include <AK/Time.h>
 #include <LibCore/Event.h>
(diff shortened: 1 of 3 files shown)


alimpfard:
unfortunately, no.
it solves part of the problem, but the global destruction order still makes it crash.
Here's a backtrace (the diff I mentioned in #4839, running `serenity-shell Shell/Tests/brace-exp.sh`)
```
#0  0x00007ffff7a83615 in raise () from /usr/lib/libc.so.6
#1  0x00007ffff7a6c862 in abort () from /usr/lib/libc.so.6
#2  0x00007ffff7a6c747 in __assert_fail_base.cold () from /usr/lib/libc.so.6
#3  0x00007ffff7a7bbf6 in __assert_fail () from /usr/lib/libc.so.6
#4  0x00005555555bc4e6 in AK::NonnullRefPtr<Core::EventLoop::SignalHandlers>::as_nonnull_ptr (this=0x5555556d0fb0) at .././AK/NonnullRefPtr.h:257
#5  AK::NonnullRefPtr<Core::EventLoop::SignalHandlers>::operator* (this=0x5555556d0fb0) at .././AK/NonnullRefPtr.h:198
#6  Core::EventLoop::unregister_signal (handler_id=0x6) at ../Libraries/LibCore/EventLoop.cpp:539
#7  0x00005555555dff84 in Line::Editor::restore (this=0x5555556d1ba0) at ../Libraries/LibLine/Editor.h:325
#8  0x00005555555d509e in Line::Editor::~Editor (this=0x5555556d1ba0, __in_chrg=<optimized out>) at ../Libraries/LibLine/Editor.cpp:192
#9  0x00005555555d50e9 in Line::Editor::~Editor (this=0x5555556d1ba0, __in_chrg=<optimized out>) at ../Libraries/LibLine/Editor.cpp:193
#10 0x00007ffff7a85db7 in __run_exit_handlers () from /usr/lib/libc.so.6
#11 0x00007ffff7a85f5e in exit () from /usr/lib/libc.so.6
#12 0x00007ffff7a6e159 in __libc_start_main () from /usr/lib/libc.so.6
#13 0x00005555555a521e in _start ()
```

alimpfard:
(the CI is likely failing for the same reason)

tomuta:
@alimpfard Ok this ought to fix it. But there's some changes I want to make still, don't like the singleton stuff...

Libraries/LibLine/Editor.cpp
```diff
@@ -451,13 +451,13 @@ void Editor::initialize()
...
+    m_registered_signal_handlers.append(Core::EventLoop::register_signal(SIGINT, [this](int) {
         interrupted();
-    });
+    }));
 
-    Core::EventLoop::register_signal(SIGWINCH, [this](int) {
+    m_registered_signal_handlers.append(Core::EventLoop::register_signal(SIGWINCH, [this](int) {
         resized();
-    });
+    }));
```
alimpfard, line 460: Maybe mention this change in the commit message? (if you didn't forget about cleaning it up after the patch that is :)

alimpfard:
> don't like the singleton stuff

It's better than the hackery I made, that's for sure

Libraries/LibLine/Editor.cpp
tomuta: Sorry I did not mean to commit this, let alone take credit for it...

alimpfard: it's just a tiny patch, nothing to fight over

tomuta: No, it's a good catch because I didn't even notice ;) I had no intention to commit the other changes, I really don't want to ;)

Libraries/LibCore/EventLoop.cpp
```diff
@@ -82,13 +83,65 @@ static NeverDestroyed<IDAllocator> s_id_allocator;
...
+struct SignalHandlersInfo {
+    HashMap<int, NonnullRefPtr<SignalHandlers>> signal_handlers;
+    int next_signal_id { 0 };
+};
+
+template<bool create_if_null = true>
+inline SignalHandlersInfo* signals_info()
+{
+    static SignalHandlersInfo* s_signals;
+    return AK::Singleton<SignalHandlersInfo>::get(s_signals);
```
ADKaster, line 140: I'm curious, what benefit does creating an AK::Singleton object have over using a function-scoped static? I guess, it's storage is on the heap rather than the `.data` segment of the object? But all of the thread-safe only-init-once stuff is handled by __cxa_guard_acquire, __cxa_guard_release, and __cxa_guard_abort in the standard library, as shown in my failed attempt to remove libstdc++ as a dependency last year (dang ports!). https://github.com/SerenityOS/serenity/pull/2242/files#diff-adcd02a2365753c4ecc8545590bfe81a1a8b2285da29c595952951ca4ec06bf8R36

tomuta: It's essentially the same thing, except that you have the compiler generate the magic code rather than writing it explicitly. Did that get checked in?

ADKaster: It got merged in but reverted because the gcc port needed a fully featured libstdc++. Right now any function scope statics call into the __cxa_guard_foo functions provided by libgcc_s.

PR #4859, opened 2021-01-09
FileManager: Toggle location bar on breadcrumb bar double-click

nvella:
Howdy all. One thing that immediately struck me about the new breadcrumb bar in FileManager is that double-clicking did not result in toggling to a location bar - I'm not sure where my mind's picked it up, but clicking/double clicking whitespace to toggle seems like a 'natural' behavior that I more or less expect from a breadcrumb bar.

I thought it would be cool to implement this, but do be warned this would be my first time changing the interface of a LibGUI component, so please please be gentle. 🤓 I added an 'on_toggle' event to BreadcrumbBar which is fired when the underlying widget is double-clicked. Double-clicking a BreadcrumbBar Button does not fire the event, only double-clicking the whitespace in the bar does. In FileManager, I separated out the 'go to location' action from an inline instantiation into a local definition (`go_to_location_action`) in the `run_in_windowed_mode` method, following logically similar actions like `copy_action`. The `on_toggle` event on BreadcrumbBar is then attached to, calling `activate` on `go_to_location_action` when `on_toggle` is fired. 

Please do let me know if there's anything silly I've done :) Cheers

### DIFF
--- a/Libraries/LibGUI/BreadcrumbBar.cpp
+++ b/Libraries/LibGUI/BreadcrumbBar.cpp
@@ -139,4 +139,10 @@ void BreadcrumbBar::set_selected_segment(Optional<size_t> index)
     segment.button->set_checked(true);
 }
 
+void BreadcrumbBar::doubleclick_event(MouseEvent& event)
+{
+    if (on_doubleclick)
+        on_doubleclick(event);
+}
+
 }
--- a/Libraries/LibGUI/BreadcrumbBar.h
+++ b/Libraries/LibGUI/BreadcrumbBar.h
@@ -48,6 +48,7 @@ public:
     Function<void(size_t index)> on_segment_click;
     Function<void(size_t index, DropEvent&)> on_segment_drop;
     Function<void(size_t index, DragEvent&)> on_segment_drag_enter;
+    Function<void(MouseEvent& event)> on_doubleclick;
 
 private:
     BreadcrumbBar();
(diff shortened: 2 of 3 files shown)


Libraries/LibGUI/BreadcrumbBar.h
```diff
@@ -46,6 +46,7 @@ class BreadcrumbBar : public GUI::Widget {
     Optional<size_t> selected_segment() const { return m_selected_segment; }
 
     Function<void(size_t index)> on_segment_click;
+    Function<void()> on_toggle;
```
awesomekling, line 49: Let's call this `on_doubleclick` and pass along the `GUI::MouseEvent&` to the callback.
The fact that you want to use it for toggling something is specific to the File Manager changes. :)

nvella: All good 👍 I was thinking that since the underlying Widget doubleclick_event wasn't public I should somehow further specify its purpose before exposing an event handler, but I see now that my reasoning doesn't exactly make sense 😄

Libraries/LibGUI/BreadcrumbBar.cpp
```diff
@@ -139,4 +139,10 @@ void BreadcrumbBar::set_selected_segment(Optional<size_t> index)
     segment.button->set_checked(true);
 }
 
+void BreadcrumbBar::doubleclick_event(MouseEvent& event)
+{
+    if (on_doubleclick)
+        on_doubleclick(event);
```
awesomekling, line 145: Please update the commit message so it matches the new name for this hook :)

nvella: Whoops 😄 It should match now.

PR #4861, opened 2021-01-09
Meta: Add flake8 to lint python files

ADKaster:
Quick copy paste job from the prettier lint shell script to be a python one. Not sure how strictly required that is, because `flake8 .` will recursively find all python files itself? There's currently only like two python files anyway.

Make sure that flake8 is installed in the CI image.

Also, make sure to run the LibCompress tests in MacOS CI.

I have a branch on my fork that tries to do a matrix job, but something about my implementation broke the IRC notifications...
https://github.com/ADKaster/serenity/blob/ci-matrix/.github/workflows/cmake.yml#L135


ADKaster:
Closing in favor of #4862. If we don't like the matrix job changes i'll re-open this as a separate, useful PR.

PR #4862, opened 2021-01-09
Meta: Use a matrix job instead of separate job definitions

ADKaster:
This also enables linting of python using flake8.

This should make it easier to keep MacOS and Linux CI in lockstep as far as what features are there and what tests are run. Previously, the LibCompress tests were added to just the Linux CI, not MacOS. 

There might be a cleaner way to do this? Feels like maybe some strategic shell scripts as run steps would make the if/else magic less awkward.

cc @BenWiederhake

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -10,10 +10,24 @@ env:
 
 jobs:
   build_and_test:
-    runs-on: ubuntu-20.04
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - os: ubuntu-20.04
+            allow-test-failure: false
+            # There sure is a lot of logic here, is there a better way?
+            # TODO: Make IRC notifications its own job/workflow?
+            should-notify-irc: ${{ github.repository == 'SerenityOS/serenity' && (github.event_name == 'pull_request' || (github.event_name == 'push' && github.ref == 'refs/heads/master')) }}
+          - os: macos-10.15
+            allow-test-failure: true
+            should-notify-irc: false
 
     steps:
     - uses: actions/checkout@v2
+    # Set default Python to python 3.x, and set Python path such that pip install works properly
+    - uses: actions/setup-python@v2
 
     # === OS SETUP ===
 
@@ -23,7 +37,8 @@ jobs:
     - name: Purge interfering packages
       # Remove GCC 9 and clang-format 10 (installed by default)
       run: sudo apt-get purge -y gcc-9 g++-9 libstdc++-9-dev clang-format-10
-    - name: Install dependencies
+      if: ${{ runner.os == 'Linux' }}
+    - name: "Install Ubuntu dependencies"
       # These packages are already part of the ubuntu-20.04 image:
       # cmake gcc-10 g++-10 shellcheck libgmp-dev
       # These aren't:
@@ -34,17 +49,30 @@ jobs:
         sudo apt-get install clang-format-11 libstdc++-10-dev libmpfr-dev libmpc-dev ninja-build npm
       # If we ever do any qemu-emulation on Github Actions, we should re-enable this:
       # e2fsprogs qemu-system-i386 qemu-utils
-    - name: Install prettier
-      run: sudo npm install -g prettier
+      if: ${{ runner.os == 'Linux' }}
     - name: Use GCC 10 instead
       run: sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-10 60 --slave /usr/bin/g++ g++ /usr/bin/g++-10
+      if: ${{ runner.os == 'Linux' }}
+
+    - name: Install macOS dependencies
+      run: brew install coreutils ninja
+      if: ${{ runner.os == 'macOS' }}
+
+    - name: Install JS dependencies
+      run: sudo npm install -g prettier
+    - name: Install Python dependencies
+      # The setup-python action set default python to python3.x. Note that we are not using system python here.
+      run: |
+        python -m pip install --upgrade pip
+        pip install flake8 requests
     - name: Check versions
-      run: set +e; g++ --version; g++-10 --version; clang-format --version; clang-format-11 --version; prettier --version; python --version; python3 --version; ninja --version
+      run: set +e; g++ --version; g++-10 --version; clang-format --version; clang-format-11 --version; prettier --version; python --version; python3 --version; ninja --version; flake8 --version
 
     # === PREPARE FOR BUILDING ===
 
     - name: Lint (Phase 1/2)
       run: ${{ github.workspace }}/Meta/lint-ci.sh
+      if: ${{ runner.os == 'Linux' }}
     - name: Toolchain cache
       uses: actions/cache@v2
       with:
@@ -66,7 +94,7 @@ jobs:
       run: |
         mkdir -p Build
         cd Build
-        cmake .. -GNinja -DBUILD_LAGOM=ON -DENABLE_ALL_THE_DEBUG_MACROS=ON
+        cmake .. -GNinja -DBUILD_LAGOM=ON -DENABLE_ALL_THE_DEBUG_MACROS=ON -DCMAKE_C_COMPILER=gcc-10 -DCMAKE_CXX_COMPILER=g++-10
 
     # === ACTUALLY BUILD AND TEST ===
 
@@ -84,14 +112,14 @@ jobs:
       run: ./check-symbols.sh
     - name: Run CMake tests
       working-directory: ${{ github.workspace }}/Build
-      run: CTEST_OUTPUT_ON_FAILURE=1 ninja test
+      run: CTEST_OUTPUT_ON_FAILURE=1 ninja test || ${{ matrix.allow-test-failure }}
       timeout-minutes: 2
     - name: Run JS tests
       working-directory: ${{ github.workspace }}/Build/Meta/Lagom
-      run: DISABLE_DBG_OUTPUT=1 ./test-js
+      run: DISABLE_DBG_OUTPUT=1 ./test-js || ${{ matrix.allow-test-failure }}
     - name: Run LibCompress tests
       working-directory: ${{ github.workspace }}/Build/Meta/Lagom
-      run: DISABLE_DBG_OUTPUT=1 ./test-compress
+      run: ./test-compress
 
      # Run analysis last, so contributors get lint/test feedback ASAP.
     - name: Perform post build CodeQL Analysis
@@ -107,14 +135,14 @@ jobs:
         ${{ toJSON(github.event) }}
         EOF
     - name: Generate IRC message
-      # I really dislike putting so much logic here, but I can't come up with something better.
-      if: github.repository == 'SerenityOS/serenity' && !cancelled() && (github.event_name == 'pull_request' || (github.event_name == 'push' && github.ref == 'refs/heads/master'))
+      if: matrix.should-notify-irc == true && !cancelled()
       run: |
         ${{ github.workspace }}/Meta/notify_irc.py <<"EOF"
         ["${{ github.actor }}", ${{ github.run_id }}, "${{ job.status }}",
         ${{ toJSON(github.event) }}
         ]
         EOF
+
   build_lagom_with_fuzzers:
     runs-on: ubuntu-20.04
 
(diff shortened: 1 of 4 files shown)


.github/workflows/cmake.yml
```diff
@@ -107,14 +134,14 @@ jobs:
         ${{ toJSON(github.event) }}
         EOF
     - name: Generate IRC message
-      # I really dislike putting so much logic here, but I can't come up with something better.
-      if: github.repository == 'SerenityOS/serenity' && !cancelled() && (github.event_name == 'pull_request' || (github.event_name == 'push' && github.ref == 'refs/heads/master'))
+      if: matrix.should-notify-irc && !cancelled()
```
ADKaster, line 137: This generated an IRC notification message when I pushed to my fork and I don't understand why. It was the same commit as this PR though..

https://github.com/ADKaster/serenity/runs/1672988057?check_suite_focus=true

ADKaster: looks like I fixed this by checking `matrix.should-notify-irc == true` instead of `matrix.should-notify-irc`.

.github/workflows/cmake.yml
```diff
@@ -34,17 +47,29 @@ jobs:
...
+
+    - name: Install js dependencies
+      run: sudo npm install -g prettier
+    - name: Install python dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install flake8 requests
     - name: Check versions
-      run: set +e; g++ --version; g++-10 --version; clang-format --version; clang-format-11 --version; prettier --version; python --version; python3 --version; ninja --version
+      run: set +e; g++ --version; g++-10 --version; clang-format-11 --version; prettier --version; python --version; python3 --version; ninja --version; flake8 --version
```
BenWiederhake, line 66: Some script tries to determine on its own which `clang-format` it should use, so I'd prefer to log the versions of all binaries in each run. Note that a missing binary does not cause an error, that's what the `set +e` dooes. Suggestion: Keep the `clang-format --version`.

.github/workflows/cmake.yml
```diff
@@ -10,10 +10,22 @@ env:
...
-    runs-on: ubuntu-20.04
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - os: ubuntu-20.04
+            allow-test-failure: false
+            # There sure is a lot of logic here, is there a better way?
+            should-notify-irc: ${{ github.repository == 'SerenityOS/serenity' && (github.event_name == 'pull_request' || (github.event_name == 'push' && github.ref == 'refs/heads/master')) }}
```
BenWiederhake, line 21: TODO for the far future: Make IRC-notification a separate job that unconditionally runs after the build jobs.
Suggestion: None, I just wanted to make a note :)

.github/workflows/cmake.yml
```diff
@@ -84,14 +109,14 @@ jobs:
...
+      run: CTEST_OUTPUT_ON_FAILURE=1 ninja test || ${{ matrix.allow-test-failure }}
       timeout-minutes: 2
     - name: Run JS tests
       working-directory: ${{ github.workspace }}/Build/Meta/Lagom
-      run: DISABLE_DBG_OUTPUT=1 ./test-js
+      run: DISABLE_DBG_OUTPUT=1 ./test-js || ${{ matrix.allow-test-failure }}
     - name: Run LibCompress tests
       working-directory: ${{ github.workspace }}/Build/Meta/Lagom
-      run: DISABLE_DBG_OUTPUT=1 ./test-compress
+      run: ./test-compress
```
BenWiederhake, line 119: Good catch!

.github/workflows/cmake.yml
```diff
@@ -34,17 +47,29 @@ jobs:
...
       # If we ever do any qemu-emulation on Github Actions, we should re-enable this:
       # e2fsprogs qemu-system-i386 qemu-utils
-    - name: Install prettier
-      run: sudo npm install -g prettier
+      if: ${{ runner.os == 'Linux' }}
     - name: Use GCC 10 instead
       run: sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-10 60 --slave /usr/bin/g++ g++ /usr/bin/g++-10
+      if: ${{ runner.os == 'Linux' }}
+
+    - name: Install MacOS dependencies
```
linusg, line 55:
```suggestion
    - name: Install macOS dependencies
```

.github/workflows/cmake.yml
```diff
@@ -34,17 +47,29 @@ jobs:
...
+      run: brew install coreutils ninja
+      if: ${{ runner.os == 'macOS' }}
+
+    - name: Install js dependencies
```
linusg, line 59:
```suggestion
    - name: Install JS dependencies
```

.github/workflows/cmake.yml
```diff
@@ -34,17 +47,29 @@ jobs:
...
+      run: sudo npm install -g prettier
+    - name: Install python dependencies
```
linusg, line 61:
```suggestion
    - name: Install Python dependencies
```

.github/workflows/cmake.yml
```diff
@@ -34,17 +47,29 @@ jobs:
...
+      run: |
+        python -m pip install --upgrade pip
+        pip install flake8 requests
```
linusg, line 64:
```suggestion
        python3 -m pip install --upgrade pip
        pip3 install flake8 requests
```

linusg (changes requested):

.github/workflows/cmake.yml
```diff
@@ -10,10 +10,22 @@ env:
...
+            allow-test-failure: false
+            # There sure is a lot of logic here, is there a better way?
+            should-notify-irc: ${{ github.repository == 'SerenityOS/serenity' && (github.event_name == 'pull_request' || (github.event_name == 'push' && github.ref == 'refs/heads/master')) }}
+          - os: macos-10.15
+            allow-test-failure: true
+            should-notify-irc: false
 
     steps:
     - uses: actions/checkout@v2
+    - uses: actions/setup-python@v2
```
BenWiederhake, line 28: I don't quite see the point of this, since we don't use its main feature of selecting the exact python version. How about adding it to the list of expected-installed programs? We only ever need it on Ubuntu anyways, something like:

```diff
     - name: "Install Ubuntu dependencies"
       # These packages are already part of the ubuntu-20.04 image:
-      # cmake gcc-10 g++-10 shellcheck libgmp-dev
+      # cmake gcc-10 g++-10 python3 shellcheck libgmp-dev
```
(Package name is correct according to https://packages.ubuntu.com/focal/python3 )

.github/workflows/cmake.yml
ADKaster: The default `setup-python@v2` action will set `python` to `python3.9`. Should I make that step more explicit, and/or still go ahead with this change? i.e. pip will be pip3 as well

linusg: ok great, that's fine like this then :)

BenWiederhake (changes requested):
I like the idea of keeping both CI "scripts" in sync :)

@"strategic shell scripts": I believe `.github/workflows/cmake.yml` is the correct place to manage the CI lifecycle. I agree that it feels weird, maybe it helps to phrase it as "`cmake.yml` is the CI script". Suggestion: Keep it like this :)

@linusg: Doink! You beat me by 8 minutes :D

.github/workflows/cmake.yml
ADKaster: I added this because I'm also adding `flake8` to lint python, which is not part of the default ubuntu 20.04 image github gives us. Doing a `pip install` without that action makes pip install things in the wrong location.

```
Installing collected packages: mccabe, pyflakes, pycodestyle, flake8
  WARNING: The script pyflakes is installed in '/home/runner/.local/bin' which is not on PATH.
  Consider adding this directory to PATH or, if you prefer to suppress this warning, use --no-warn-script-location.
  WARNING: The script pycodestyle is installed in '/home/runner/.local/bin' which is not on PATH.
  Consider adding this directory to PATH or, if you prefer to suppress this warning, use --no-warn-script-location.
  WARNING: The script flake8 is installed in '/home/runner/.local/bin' which is not on PATH.
  Consider adding this directory to PATH or, if you prefer to suppress this warning, use --no-warn-script-location.
Successfully installed flake8-3.8.4 mccabe-0.6.1 pycodestyle-2.6.0 pyflakes-2.2.0
```

Which means that the flake8 version check exits with 127 because it could not be found.,

ADKaster: See this action log https://github.com/ADKaster/serenity/runs/1672776954?check_suite_focus=true

.github/workflows/cmake.yml
ADKaster: Yeah, unfortunately, in order to use the "nice" `on` syntax, it would have to be a separate workflow. which... ew.

BenWiederhake: *Shocked pikachu face*

Oh well.

.github/workflows/cmake.yml
ADKaster: That is to say, using the `setup-python` action ensures that a good version of python is in our $PATH, and that the version of python in our path is the one that pip installs things to. Also, apparently GitHub reserves the right to mess with what python in our path. https://github.com/actions/setup-python#using-python-without-setup-python

> The default version of Python in PATH vary between runners and can change unexpectedly so we recommend you always use setup-python.

BenWiederhake: If we want to continue using Python, then I agree that flake8 and therefore everything Pythony is a good idea.

However, I already feel bad for having committed such a ~language crossover crime~ file, and don't think we will add more Python to Serenity anyway. C++, JS, CMake, Sh, Bash, JS, and SerenityShell is already plenty of languages.

Your choice, both options are very reasonable I think.

ADKaster: > [I]  don't think we will add more Python to Serenity anyway.

What's this then? ;) https://github.com/SerenityOS/serenity/pull/4771

I  have nothing against python, it's way more approachable to me than shell scripts.

.github/workflows/cmake.yml
ADKaster: changed

.github/workflows/cmake.yml
ADKaster: uppercased

.github/workflows/cmake.yml
ADKaster: uppercased

.github/workflows/cmake.yml
ADKaster: I added a few comments to clarify this.

.github/workflows/cmake.yml
ADKaster: put the clang-format --version back. even though it's not a real thing :)

.github/workflows/cmake.yml
ADKaster: Added a todo in the code for this

ADKaster:
Ack, I should make sure the CI passes before hitting the re-review button :(

Meta/lint-python.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
+
+set -e
+
+script_path=$(cd -P -- "$(dirname -- "$0")" && pwd -P)
+cd "${script_path}/.." || exit 1
+
+if ! command -v flake8 >/dev/null 2>&1 ; then
+    echo "flake8 is not available. Either skip this script, or install flake8."
+    exit 1
```
BenWiederhake, line 10: It's fine to do `exit 1` here.
It's fine to run this as part of `lint-ci.sh`.
It's fine to run `lint-ci.sh` as part of `git commit`.
However, all three at the same time sounds like a bad idea.

We could:
- Require devs to install flake8 if they want to use the pre-commit hook
- Add an envvar `SKIP_MISSING` or something that is set for the pre-commit hook, and results in an `exit 0` here.
- Eat cupcakes? Dunno :)

ADKaster: `lint-prettier.sh` does the exact same thing 🤔 . So does `lint-clang-format.sh`, if the linter tool is not installed. I would say the options are more along the lines of...:

- Make all linters required to use the pre-commit hook
- Make no linters required to use the pre-commit hook
- Make just clang-format required to use the pre-commit hook

.github/workflows/cmake.yml
BenWiederhake: Alrighty! Then let's go with python :D

Meta/lint-python.sh
linusg: I think the most sensible option is to have the scripts check if there are any files to process first, and only then complain about the absence of the required commands.

- Someone who never touches JS files should not be required to install prettier (or flake8 for that matter). The pre-commit hook already passes staged files to these scripts to ensure they don't check all files on every commit
- CI always checks all files (git ls-files), but installing all required tools trivial

.github/workflows/cmake.yml
alimpfard: One can never have too many utility languages!
    - some guy that tends to make DSLs for every single problem he sees ahead of him

linusg (approved):

PR #4866, opened 2021-01-09
TextEditor: Implemented word wrapping

### LINKED ISSUE #4788: TextEditor: Implement word wrap
We should have a mode where we only wrap at word boundaries, not in the middle of words.

supex0fan:
Add a new wrapping mode to the TextEditor that will wrap lines at the
spaces between words.

Replace the previous menubar checkbox 'Wrapping Mode' in HackStudio and
the TextEditor with an exclusive submenu which allows switching between
'No wrapping', 'Wrap anywhere' and 'Wrap at words'. 'Wrap anywhere' (the
new 'Wrap lines') is still the default mode.

Setting the wrapping mode in the constructors of the TextEditorWidget
and HackStudio has been removed, it is now set when constructing the
menubar actions.

Closes #4788

### DIFF
--- a/Userland/Libraries/LibGUI/TextEditor.cpp
+++ b/Userland/Libraries/LibGUI/TextEditor.cpp
@@ -157,7 +157,7 @@ TextPosition TextEditor::text_position_at_content_position(const Gfx::IntPoint& 
 
     size_t line_index = 0;
 
-    if (is_line_wrapping_enabled()) {
+    if (is_wrapping_enabled()) {
         for (size_t i = 0; i < line_count(); ++i) {
             auto& rect = m_line_visual_data[i].visual_rect;
             if (position.y() >= rect.top() && position.y() <= rect.bottom()) {
@@ -198,7 +198,7 @@ TextPosition TextEditor::text_position_at_content_position(const Gfx::IntPoint& 
         break;
     case Gfx::TextAlignment::CenterRight:
         // FIXME: Support right-aligned line wrapping, I guess.
-        ASSERT(!is_line_wrapping_enabled());
+        ASSERT(!is_wrapping_enabled());
         column_index = (position.x() - content_x_for_position({ line_index, 0 }) + fixed_glyph_width() / 2) / fixed_glyph_width();
         break;
     default:
@@ -872,7 +872,7 @@ int TextEditor::content_x_for_position(const TextPosition& position) const
         return m_horizontal_content_padding + ((is_single_line() && icon()) ? (icon_size() + icon_padding()) : 0) + x_offset;
     case Gfx::TextAlignment::CenterRight:
         // FIXME
-        ASSERT(!is_line_wrapping_enabled());
+        ASSERT(!is_wrapping_enabled());
         return content_width() - m_horizontal_content_padding - (line.length() * fixed_glyph_width()) + (position.column() * fixed_glyph_width());
     default:
         ASSERT_NOT_REACHED();
@@ -952,7 +952,7 @@ Gfx::IntRect TextEditor::line_content_rect(size_t line_index) const
         line_rect.center_vertically_within({ {}, frame_inner_rect().size() });
         return line_rect;
     }
-    if (is_line_wrapping_enabled())
+    if (is_wrapping_enabled())
         return m_line_visual_data[line_index].visual_rect;
     return {
         content_x_for_position({ line_index, 0 }),
@@ -1286,7 +1286,7 @@ void TextEditor::did_update_selection()
     m_copy_action->set_enabled(has_selection());
     if (on_selection_change)
         on_selection_change();
-    if (is_line_wrapping_enabled()) {
+    if (is_wrapping_enabled()) {
         // FIXME: Try to repaint less.
         update();
     }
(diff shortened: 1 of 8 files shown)


Libraries/LibGUI/TextEditor.cpp
```diff
@@ -1406,13 +1406,28 @@ void TextEditor::recompute_visual_lines(size_t line_index)
     if (is_line_wrapping_enabled()) {
         int line_width_so_far = 0;
 
+        size_t last_whitespace_index = 0;
+        size_t line_width_since_last_whitespace = 0;
         auto glyph_spacing = font().glyph_spacing();
         for (size_t i = 0; i < line.length(); ++i) {
             auto code_point = line.code_points()[i];
+            if (code_point == 32) {
```
awesomekling, line 1414: Let's compare against `' '` instead of `32`.

Libraries/LibGUI/TextEditor.h
```diff
@@ -298,6 +301,7 @@ class TextEditor
     bool m_has_pending_change_notification { false };
     bool m_automatic_indentation_enabled { false };
     bool m_line_wrapping_enabled { false };
+    bool m_word_wrapping_enabled { false };
```
awesomekling, line 304: Wait, what if word wrapping is enabled, but line wrapping is not? Seems like this should be a single "wrapping mode" rather than two flags.

awesomekling (changes requested):

Libraries/LibGUI/TextEditor.cpp
bcoles: There's probably a bunch of whole bunch of whitespace characters which should be checked.

Space `0x20` and tab `0x09` are the most immediately obvious.

alimpfard: why not just use `isspace()` then?

Applications/TextEditor/TextEditorWidget.cpp
```diff
@@ -364,12 +364,12 @@ TextEditorWidget::TextEditorWidget()
...
-    m_line_wrapping_setting_action = GUI::Action::create_checkable("Line wrapping", [&](auto& action) {
-        m_editor->set_line_wrapping_enabled(action.is_checked());
+    m_wrapping_mode_action = GUI::Action::create_checkable("Wrapping Mode", [&](auto& action) {
+        m_editor->set_wrapping_mode_enabled(action.is_checked());
     });
-    m_line_wrapping_setting_action->set_checked(m_editor->is_line_wrapping_enabled());
+    m_wrapping_mode_action->set_checked(m_editor->is_wrapping_mode_enabled());
 
-    auto menubar = GUI::MenuBar::construct();
+   auto menubar = GUI::MenuBar::construct();
```
emanuele6, line 372: checks are failing because you removed a space here by mistake

emanuele6 (changes requested):

Applications/TextEditor/TextEditorWidget.cpp
supex0fan: Thanks for finding that

Libraries/LibGUI/TextEditor.h
awesomekling: Oops, what I meant by a single wrapping mode was something like:
```cpp
enum class WrappingMode {
	NoWrap,
	WrapLines,
	WrapWords,
};
```

Sorry about the confusion!!

supex0fan: It seems like in that system, you can only wrap either lines or words but not both. Is that correct?

supex0fan: Returning to your original comment, word wrapping without line wrapping does nothing in the current code and in my opinion, shouldn't do anything, thus it wasn't possible for word wrapping to be enabled without line wrapping due to the way the menubar actions were setup. If line wrapping was turned off, the action for word wrapping was disabled (not able to be enabled). 

That's why I'm a bit confused about the enum posted above, how do you want word wrapping without line wrapping to work?

awesomekling: Sorry, WrapLines was a dumb name. What I meant was having a way to choose between:
- no wrapping 
- the current mechanism (break at any character)
- word wrap (only break between full words)

supex0fan: Okay I get it now. Would you be okay with just two check boxes for setting the options? One saying 'Wrap Lines' and the other saying 'Wrap Words'. Or do you want a radio button style system.

Userland/Libraries/LibGUI/TextEditor.cpp
```diff
@@ -1469,13 +1484,13 @@ void TextEditor::for_each_visual_line(size_t line_index, Callback callback) cons
...
+void TextEditor::set_wrapping_mode(WrappingMode mode)
 {
-    if (m_line_wrapping_enabled == enabled)
+    if (m_wrapping_mode == mode)
         return;
 
-    m_line_wrapping_enabled = enabled;
-    horizontal_scrollbar().set_visible(!m_line_wrapping_enabled);
+    m_wrapping_mode = mode;
+    horizontal_scrollbar().set_visible(!m_wrapping_mode);
```
Dexesttp, line 1493: I think using either `horizontal_scrollbar().set_visible(m_wrapping_mode == WrappingMode::NoWrap)` or `horizontal_scrollbar().set_visible(!is_wrapping_enabled())` would be clearer for this ? The code works because `NoWrap == 0`, but is very unexpected.

supex0fan: Good catch, that wasn't intentional, it was leftover from the previous design of the commit.

PR #4870, opened 2021-01-09
Everywhere: Continue replacing dbg() dbgln().

asynts:
*Disclaimer: I simply replaced the statements without considering context and without proper testing. It would simply take to long otherwise. (There are about a thousand calls to `dbg()`.)*

This pull requests or the pull requests following this will probably introduce a few bugs, but I don't think this is to problematic, I've taken the following precautions:

 1. Split the changes into multiple commits and ensure that they build individually. This should make it possible to `git bisect` easily.

 2. Copy and paste the format strings as much as possible to avoid typos.

 3. Proof read the commits to avoid typos or similar errors.

 4. Ensure that the system boots with these changes and run every userland application that changed once.

If someone got an idea to catch more errors without spending too much time on every statement, please make suggestions.

---

Before:

~~~none
$ grep -rnI 'dbg()' | wc -l
1373
~~~

After:
~~~none
$ grep -rnI 'dbg()' | wc -l
1242
~~~

I am planing to write some scripts to deal with trivial cases but it would be too difficult to write a script to replace all uses.

---

I also added a `FormatString` helper class which simplifies the implementation of formatters and makes it possible to implement them in the headers without including `String.h`.


asynts:
I added another bundle where I used a simple `sed` command to replace many trivial cases. Now we have:

~~~none
$ grep -rnI 'dbg()' | wc -l
913
~~~

PR #4872, opened 2021-01-09
LibJS: Implement String.prototype.split

nooga:
This adds a String.prototype.split implementation modelled after 
ECMA262 specification. 

There is a tiny kludge for when the separator is an empty string. 
Basic tests and visiting google.com prove that this is working. 

I will work on the RegExp support tomorrow :^)

### DIFF
--- a/Libraries/LibJS/Runtime/StringPrototype.cpp
+++ b/Libraries/LibJS/Runtime/StringPrototype.cpp
@@ -28,6 +28,7 @@
 #include <AK/Function.h>
 #include <AK/StringBuilder.h>
 #include <LibJS/Heap/Heap.h>
+#include <LibJS/Runtime/Array.h>
 #include <LibJS/Runtime/Error.h>
 #include <LibJS/Runtime/GlobalObject.h>
 #include <LibJS/Runtime/PrimitiveString.h>
@@ -57,6 +58,17 @@ static String ak_string_from(VM& vm, GlobalObject& global_object)
     if (!this_object)
         return {};
     return Value(this_object).to_string(global_object);
+}
+
+static Optional<size_t> split_match(const String& haystack, size_t start, const String& needle)
+{
+    auto r = needle.length();
+    auto s = haystack.length();
+    if (start + r > s)
+        return {};
+    if (!haystack.substring_view(start).starts_with(needle))
+        return {};
+    return start + r;
 }
 
 StringPrototype::StringPrototype(GlobalObject& global_object)
@@ -90,6 +102,7 @@ void StringPrototype::initialize(GlobalObject& global_object)
     define_native_function(vm.names.substring, substring, 2, attr);
     define_native_function(vm.names.includes, includes, 1, attr);
     define_native_function(vm.names.slice, slice, 2, attr);
+    define_native_function(vm.names.split, split, 2, attr);
     define_native_function(vm.names.lastIndexOf, last_index_of, 1, attr);
     define_native_function(vm.well_known_symbol_iterator(), symbol_iterator, 0, attr);
 }
@@ -505,6 +518,74 @@ JS_DEFINE_NATIVE_FUNCTION(StringPrototype::slice)
     auto part_length = index_end - index_start;
     auto string_part = string.substring(index_start, part_length);
     return js_string(vm, string_part);
+}
+
+JS_DEFINE_NATIVE_FUNCTION(StringPrototype::split)
+{
+    // FIXME Implement the @@split part
+
+    auto string = ak_string_from(vm, global_object);
+    if (string.is_null())
+        return {};
+
+    auto* result = Array::create(global_object);
+    size_t result_len = 0;
+
+    auto limit = static_cast<u32>(MAX_U32);
+    if (!vm.argument(1).is_undefined()) {
+        limit = vm.argument(1).to_u32(global_object);
+        if (vm.exception())
+            return {};
+    }
+
+    auto separator = vm.argument(0).to_string(global_object);
+    if (vm.exception())
+        return {};
+
+    if (limit == 0)
+        return result;
+
+    if (vm.argument(0).is_undefined()) {
+        result->define_property(0, js_string(vm, string));
+        return result;
+    }
+
+    auto len = string.length();
+    auto separator_len = separator.length();
+    if (len == 0) {
+        if (separator_len > 0)
+            result->define_property(0, js_string(vm, string));
+        return result;
+    }
+
+    size_t start = 0;
+    auto pos = start;
+    if (separator_len == 0) {
+        for (pos = 0; pos < len; pos++)
+            result->define_property(pos, js_string(vm, string.substring(pos, 1)));
+        return result;
+    }
+
+    while (pos != len) {
+        auto e = split_match(string, pos, separator);
+        if (!e.has_value()) {
+            pos += 1;
+            continue;
+        }
+
+        auto segment = string.substring_view(start, pos - start);
+        result->define_property(result_len, js_string(vm, segment));
+        result_len++;
+        if (result_len == limit)
+            return result;
+        start = e.value();
+        pos = start;
+    }
+
+    auto rest = string.substring(start, len - start);
+    result->define_property(result_len, js_string(vm, rest));
+
+    return result;
 }
 
 JS_DEFINE_NATIVE_FUNCTION(StringPrototype::last_index_of)
--- a/Libraries/LibJS/Runtime/Value.cpp
+++ b/Libraries/LibJS/Runtime/Value.cpp
@@ -470,6 +470,12 @@ i32 Value::as_i32() const
     return static_cast<i32>(as_double());
 }
 
+u32 Value::as_u32() const
+{
+    ASSERT(as_double() >= 0);
+    return min((double)as_i32(), MAX_U32);
+}
+
 size_t Value::as_size_t() const
 {
     ASSERT(as_double() >= 0);
@@ -492,6 +498,19 @@ i32 Value::to_i32(GlobalObject& global_object) const
     if (number.is_nan() || number.is_infinity())
         return 0;
     return number.as_i32();
+}
+
+u32 Value::to_u32(GlobalObject& global_object) const
+{
+    // 7.1.7 ToUint32, https://tc39.es/ecma262/#sec-touint32
+    auto number = to_number(global_object);
+    if (global_object.vm().exception())
+        return INVALID;
+    if (number.is_nan() || number.is_infinity())
+        return 0;
+    if (number.as_double() <= 0)
+        return 0;
+    return number.as_u32();
 }
 
 size_t Value::to_size_t(GlobalObject& global_object) const
--- a/Libraries/LibJS/Runtime/Value.h
+++ b/Libraries/LibJS/Runtime/Value.h
@@ -36,6 +36,8 @@
 
 // 2 ** 53 - 1
 static constexpr double MAX_ARRAY_LIKE_INDEX = 9007199254740991.0;
+// 2 ** 32 - 1
+static constexpr double MAX_U32 = 4294967295.0;
 
 namespace JS {
 
@@ -241,6 +243,7 @@ public:
     Function& as_function();
 
     i32 as_i32() const;
+    u32 as_u32() const;
     size_t as_size_t() const;
 
     String to_string(GlobalObject&, bool legacy_null_to_empty_string = false) const;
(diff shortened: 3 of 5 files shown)


Libraries/LibJS/Runtime/StringPrototype.cpp
```diff
@@ -59,6 +60,17 @@ static String ak_string_from(VM& vm, GlobalObject& global_object)
     return Value(this_object).to_string(global_object);
 }
 
+static Optional<size_t> split_match(AK::String haystack, size_t start, AK::String needle)
```
linusg, line 63: We're usually not using the `AK::` prefix (no idea what that is actually called :smile:), also I think those strings can be const references, no copy needed:

```suggestion
static Optional<size_t> split_match(const String& haystack, size_t start, const String& needle)
```

Libraries/LibJS/Runtime/StringPrototype.cpp
```diff
@@ -507,6 +520,88 @@ JS_DEFINE_NATIVE_FUNCTION(StringPrototype::slice)
     return js_string(vm, string_part);
 }
 
+JS_DEFINE_NATIVE_FUNCTION(StringPrototype::split)
+{
```
linusg, line 524: If you don't want to implement the `@@split` part (step 2 in the spec), please leave a FIXME comment :)

Libraries/LibJS/Runtime/StringPrototype.cpp
```diff
@@ -507,6 +520,88 @@ JS_DEFINE_NATIVE_FUNCTION(StringPrototype::slice)
...
+    auto string = ak_string_from(vm, global_object);
+    if (string.is_null())
+        return {};
+
+    auto* result = Array::create(global_object);
+    size_t result_len = 0;
+
+    size_t limit = NumericLimits<size_t>::max();
+    if (!vm.argument(1).is_undefined())
+        limit = vm.argument(1).to_size_t(global_object);
```
linusg, line 534: `to_size_t` is non-standard and being phased out, please don't use it in new code. The spec says this should use the [`ToUint32()` abstract operation](https://tc39.es/ecma262/#sec-touint32), so let's add `Value::to_u32()` and use that instead. The implementation can be similar to `to_size_t`'s, but there are subtle differences - this would handle infinity incorrectly, for example.

Libraries/LibJS/Runtime/StringPrototype.cpp
```diff
@@ -507,6 +520,88 @@ JS_DEFINE_NATIVE_FUNCTION(StringPrototype::slice)
...
+JS_DEFINE_NATIVE_FUNCTION(StringPrototype::split)
+{
+    auto string = ak_string_from(vm, global_object);
+    if (string.is_null())
+        return {};
+
+    auto* result = Array::create(global_object);
+    size_t result_len = 0;
+
+    size_t limit = NumericLimits<size_t>::max();
```
linusg, line 532: IIRC `size_t` doesn't have a well defined size? Let's just hardcode 2^32 - 1 as the spec says, possibly `constexpr`'d (or just as 4294967295).

Libraries/LibJS/Runtime/StringPrototype.cpp
```diff
@@ -507,6 +520,88 @@ JS_DEFINE_NATIVE_FUNCTION(StringPrototype::slice)
...
+
+    if (limit == 0)
+        return result;
+
+    if (vm.argument(0).is_undefined()) {
+        if (!result->define_property(0, js_string(vm, string))) {
+            vm.throw_exception<TypeError>(global_object, ErrorType::ObjectDefinePropertyReturnedFalse);
+            return {};
+        }
+        return result;
```
linusg, line 548: You can skip this, we fully control the `result` array and there's no way `define_property()` will return false for this case :)

```suggestion
        result->define_property(0, js_string(vm, string));
        return result;
```

or, to be extra sure:

```suggestion
        auto success = result->define_property(0, js_string(vm, string));
        ASSERT(success);
        return result;
```

Libraries/LibJS/Runtime/StringPrototype.cpp
```diff
@@ -507,6 +520,88 @@ JS_DEFINE_NATIVE_FUNCTION(StringPrototype::slice)
...
+    }
+
+    size_t len = string.length();
+    size_t separator_len = separator.length();
+    if (len == 0) {
+        if (separator_len > 0) {
+            if (!result->define_property(0, js_string(vm, string))) {
+                vm.throw_exception<TypeError>(global_object, ErrorType::ObjectDefinePropertyReturnedFalse);
+                return {};
+            }
```
linusg, line 558: Same as above

Libraries/LibJS/Runtime/StringPrototype.cpp
```diff
@@ -507,6 +520,88 @@ JS_DEFINE_NATIVE_FUNCTION(StringPrototype::slice)
...
+    }
+
+    size_t start = 0;
+    size_t pos = start;
+    if (separator_len == 0) {
+        for (pos = 0; pos < len; pos++) {
+            if (!result->define_property(pos, js_string(vm, string.substring(pos, 1)))) {
+                vm.throw_exception<TypeError>(global_object, ErrorType::ObjectDefinePropertyReturnedFalse);
+                return {};
+            }
```
linusg, line 570: Same as above

Libraries/LibJS/Runtime/StringPrototype.cpp
```diff
@@ -507,6 +520,88 @@ JS_DEFINE_NATIVE_FUNCTION(StringPrototype::slice)
...
+            pos += 1;
+            continue;
+        }
+        if (separator_len == 0)
+            pos += 1;
+        auto segment = string.substring_view(start, pos - start);
+        if (!result->define_property(result_len, js_string(vm, segment))) {
+            vm.throw_exception<TypeError>(global_object, ErrorType::ObjectDefinePropertyReturnedFalse);
+            return {};
+        }
```
linusg, line 587: Same as above

Libraries/LibJS/Runtime/StringPrototype.cpp
```diff
@@ -507,6 +520,88 @@ JS_DEFINE_NATIVE_FUNCTION(StringPrototype::slice)
...
+        start = e.value();
+        pos = start;
+    }
+
+    auto rest = string.substring(start, len - start);
+
+    if (!result->define_property(result_len, js_string(vm, rest))) {
+        vm.throw_exception<TypeError>(global_object, ErrorType::ObjectDefinePropertyReturnedFalse);
+        return {};
+    }
```
linusg, line 600: Same as above

linusg (changes requested):

Libraries/LibJS/Runtime/StringPrototype.cpp
nooga: It was like that before but I changed it to what  it is in the last minute :D

Libraries/LibJS/Runtime/Value.cpp
```diff
@@ -470,6 +470,12 @@ i32 Value::as_i32() const
     return static_cast<i32>(as_double());
 }
 
+u32 Value::as_u32() const
+{
+    ASSERT(as_double() >= 0);
```
emanuele6, line 475: isn't it more appropriate to assert `as_i32() >= 0`?

Libraries/LibJS/Runtime/StringPrototype.cpp
```diff
@@ -507,6 +520,71 @@ JS_DEFINE_NATIVE_FUNCTION(StringPrototype::slice)
...
+    auto string = ak_string_from(vm, global_object);
+    if (string.is_null())
+        return {};
+
+    auto* result = Array::create(global_object);
+    size_t result_len = 0;
+
+    auto limit = static_cast<u32>(MAX_U32);
+    if (!vm.argument(1).is_undefined())
+        limit = vm.argument(1).to_u32(global_object);
```
linusg, line 536: The `to_number()` part in `to_u32()` can fail so this needs an exception check:

```suggestion
    if (!vm.argument(1).is_undefined()) {
        limit = vm.argument(1).to_u32(global_object);
        if (vm.exception())
            return {};
    }
```

Libraries/LibJS/Runtime/Value.cpp
```diff
@@ -494,6 +500,18 @@ i32 Value::to_i32(GlobalObject& global_object) const
     return number.as_i32();
 }
 
+u32 Value::to_u32(GlobalObject& global_object) const
+{
```
linusg, line 504:
```suggestion
u32 Value::to_u32(GlobalObject& global_object) const
{
    // 7.1.7 ToUint32, https://tc39.es/ecma262/#sec-touint32
```

(I just like putting links to the spec in these functions, makes it easier to look them up, where they're going to be used etc.)

Libraries/LibJS/Runtime/Value.cpp
```diff
@@ -494,6 +500,18 @@ i32 Value::to_i32(GlobalObject& global_object) const
...
+    auto number = to_number(global_object);
+    if (global_object.vm().exception())
+        return 0;
```
linusg, line 507:
```suggestion
        return INVALID;
```

that's just zero but hopefully makes it more clear that the return value of this function is to be discarded in the case of an exception.

Libraries/LibJS/Runtime/Value.h
```diff
@@ -36,6 +36,8 @@
 
 // 2 ** 53 - 1
 static constexpr double MAX_ARRAY_LIKE_INDEX = 9007199254740991.0;
+// 2 ** 32 -1
```
linusg, line 39:
```suggestion
// 2 ** 32 - 1
```

n i t p i c k

linusg (changes requested):

Libraries/LibJS/Runtime/StringPrototype.cpp
alimpfard: As a general rule of thumb I figured out while messing with these in Spreadsheet, _anything_ that takes a global object can (and likely will) throw.

linusg: That, as well as the `to_*` part (as opposed to `as_*`) :)

nooga: I was thinking, maybe we should somehow mark the methods that may throw? Coming back to LibJS after some time I realized that it has grown beyond the capabilities of my tiny brain.

linusg: The closest I can think of is `[[nodiscard]]`, but that won't help as exceptions are side effects...

linusg (approved):
LGTM, nicely done!

Libraries/LibJS/Runtime/StringPrototype.cpp
```diff
@@ -507,6 +520,74 @@ JS_DEFINE_NATIVE_FUNCTION(StringPrototype::slice)
...
+    size_t start = 0;
+    auto pos = start;
+    if (separator_len == 0) {
+        for (pos = 0; pos < len; pos++)
+            result->define_property(pos, js_string(vm, string.substring(pos, 1)));
+        return result;
+    }
+
+    while (pos != len) {
+        auto e = split_match(string, pos, separator);
```
alimpfard, line 570: Any reason to avoid `memmem` here?

nooga: I wanted to explicitly define this operation in terms of what the standard talks about. It uses AK's `starts_with()` under the hood, and that uses `memcmp` I think.

Libraries/LibJS/Runtime/StringPrototype.cpp
nooga: Forgot that you can't define custom attributes in C++. :(

Libraries/LibJS/Runtime/StringPrototype.cpp
alimpfard: I meant the "check equality for every character in string" part, `memmem` would probably be a fair bit faster than this for small strings, and it would give you the same-ish interface as `split_match` too.

if it's to follow the spec though, :shrug:

nooga: I just translated the pseudo code to C++ without getting smart about it. I think that there's nothing bad in following the spec to the letter even if it leads to sub-optimal performance at this stage. We can always iterate later and have the naive-yet-standard implementation as the baseline.

alimpfard: Following the spec is a fairly valid reason, I suppose.

PR #4874, opened 2021-01-10
Core::get_password(): don't return final newline character as part of the password.

emanuele6:
This pull request also changes the default password of anon to "foo" (it used to be "foo\n").

### DIFF
--- a/Libraries/LibCore/GetPassword.cpp
+++ b/Libraries/LibCore/GetPassword.cpp
@@ -1,5 +1,6 @@
 /*
  * Copyright (c) 2020, Peter Elliott <pelliott@ualberta.ca>
+ * Copyright (c) 2021, Emanuele Torre <torreemanuele6@gmail.com>
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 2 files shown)


Libraries/LibCore/GetPassword.cpp
```diff
@@ -48,13 +48,18 @@ Result<String, int> get_password(const StringView& prompt)
...
     size_t n = 0;
 
-    int ret = getline(&password, &n, stdin);
+    auto line_length = getline(&password, &n, stdin);
     tcsetattr(STDIN_FILENO, TCSAFLUSH, &original);
     putchar('\n');
-    if (ret < 0) {
+    if (line_length < 0) {
         return errno;
     }
```
emanuele6, line 56: I think this is a mistake: we are returning `errno` if ret/line_length is less than 0.

We probably mean to return `getline`'s `errno` here, not `tcsetattr`'s or `putchar`'s.

Is that the case?

emanuele6: I'm going to assume that is the case, I'll fix that too.

emanuele6: Oh, I noticed that we are returning the errno of the other `tcsetattr()`'s errno some lines before. I'm not sure what to do here since I don't know what the error value should be.

emanuele6: But I don't think it makes sense to bail out if we cannot restore the terminal setting since we already got the password.

PR #4883, opened 2021-01-10
LibLine: C-V<key> / LibVT: Make the "Negative" attribute work

alimpfard:
A diff of 163 lines just to do:
shot-2021-01-10_16-46-37

---

This commit adds support for inserting in a "verbatim" mode where a
single uninterpreted key is appended to the buffer.
As this allows the user to input control characters, all control
characters except \n (^M) are rendered in their caret form, with
reverse video (SGR 7) applied to it.
To not break cursor movement, the concept of "masked" characters is
introduced to the StringMetrics interface, which can be mostly ignored
by the rest of the system.

It should be noted that unlike some other line editing libraries,
LibLine does _not_ render a hard tab as a tab, but rather as '^I',
which greatly simplifies cursor handling.

### DIFF
--- a/Kernel/TTY/VirtualConsole.cpp
+++ b/Kernel/TTY/VirtualConsole.cpp
@@ -274,11 +274,11 @@ static inline u8 attribute_to_vga(const VT::Attribute& attribute)
 
     // Background color
     vga_attr &= ~0x70;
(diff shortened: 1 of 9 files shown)


awesomekling:
Rad!

PR #4887, opened 2021-01-10
Everywhere: Continue replacing dbg() dbgln(). (2)

asynts:
Followup to  #4870.

Before:
~~~none
$ grep -rnI 'dbg()' | wc -l
865
~~~

After:
~~~none
$ grep -rnI 'dbg()' | wc -l
731
~~~

---

The majority of the remaining debug statement is enclosed in `#ifdef FOO_DEBUG`, I haven't touched these yet. I am thinking about doing something like `dbgln<enable_foo_debug>()` in these cases but I haven't finished that thought just yet.

### DIFF
--- a/AK/Format.h
+++ b/AK/Format.h
@@ -396,6 +396,7 @@ template<typename... Parameters>
 void dbgln(StringView fmtstr, const Parameters&... parameters) { vdbgln(fmtstr, VariadicFormatParams { parameters... }); }
 template<typename... Parameters>
 void dbgln(const char* fmtstr, const Parameters&... parameters) { dbgln(StringView { fmtstr }, parameters...); }
+inline void dbgln() { dbgln(""); }
 
 template<typename T, typename = void>
 struct HasFormatter : TrueType {
--- a/Kernel/ACPI/Parser.cpp
+++ b/Kernel/ACPI/Parser.cpp
@@ -161,7 +161,7 @@ void Parser::access_generic_address(const Structures::GenericAddressStructure& s
             dbgln("ACPI Warning: Unknown access size {}", structure.access_size);
             ASSERT(structure.bit_width != (u8)GenericAddressStructure::BitWidth::QWord);
             ASSERT(structure.bit_width != (u8)GenericAddressStructure::BitWidth::Undefined);
-            dbg() << "ACPI: Bit Width - " << structure.bit_width << " bits";
+            dbgln("ACPI: Bit Width - {} bits", structure.bit_width);
             address.out(value, structure.bit_width);
             break;
         }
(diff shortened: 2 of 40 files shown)


BenWiederhake:
`dbgln<enable_foo_debug>()` sounds awesome!

alimpfard:
A downside to that (`dbgln<enable_foo_debug>()`) would be the fact that its arguments are evaluated.
if something effectful (or something with a performance hit) ends up there, it won't be optimised away AFAIK

BenWiederhake:
Just thinking aloud:
```
dbgln_if(DEBUG_WHATEVER, "quux: {}, frob: {}", m_quux, build_frob());
```
Where `dbgln_if` is a macro that transforms either to:
```
dbgln("quux: {}, frob: {}", m_quux, build_frob());
```
or:
```
(void)(decltype({m_quux, build_frob()}))0;
```

Or something along those lines.

alimpfard:
(also thinking out loud)

somehow transforming the arguments to a lambda sounds like a better option, less preprocessor magic

`dbgln<flag>(x,y,z)` -> `dbgln_p<flag>([&] { dbgln(x,y,z); })`
where `dbgln_p<true>` calls the lambda and `dbgln_p<false>` doesn't.

not sure if this will have any impact on performance, or the semantics of the debug log call though

BenWiederhake:
Ohh, nice! Just one suggestion: Don't name it `dbgln_p`, since now it's not just for `dbgln`.
Some names that I can come up with: `if_` (because it's damn close to `if` anyway), or `if_constexpr`.

Wait a second, does C++20 already contain something like that?

[Yes it does!](https://en.cppreference.com/w/cpp/language/if)

So we might write something like:

```
    if constexpr (DEBUG_WHATEVER)
        "quux: {}, frob: {}", m_quux, build_frob();
```

That's not a one-liner, but still much nicer than the indentation-skipping `#ifdef`s. And with the constexpr it's highly likely that the compiler will do the RightThing™. [Godbolt says yes.](https://godbolt.org/z/bKodcz)

asynts:
> Where `dbgln_if` is a macro that transforms either to:

Macros are generally bad because they don't recognize curly braces which is already annoying with `ASSERT` and such.

  > somehow transforming the arguments to a lambda sounds like a better option, less preprocessor magic

I don't quite understand what you mean by this. Do you mean something like this?

~~~c++
if_debug<debug_foo>([&] {
    dbgln("Hello, world!);
});
~~~

in that case, the following would be simpler:

~~~c++
if constexpr (debug_foo) {
    dbgln("Hello, world!);
}
~~~

This would prevent side effects, but I like the template stuff because it is much less annoying to type.

Maybe we can do the template stuff in most cases and then use an `if constexpr` to deal with weird performance relevant cases?

alimpfard:
> Do you mean something like this...

the idea was to get something that _expands_ to that, not sure if that's even possible.

PR #4894, opened 2021-01-10
AK: Find a value in any container offering iterators

ldm5180:
Problem:                                                              
- `find` is implemented inside of each container. This coupling       
  requires that each container needs to individually provide `find`.  
                                                                      
Solution:                                                             
- Decouple the `find` functionality from the container. This allows   
  provides a `find` algorithm which can work with all                 
  containers. Containers can still provide their own `find` in the    
  case where it can be optimized.                                     
- This also allows for searching sub-ranges of a container rather than
  the entire container as some of the container-specific member       
  functions enforced.                                                 
                                                                      
Note:                                                                 
- @davidstone's talk from 2015 C++Now conference entitled "Functions  
  Want to be Free" encourages this style:                             
  (https://www.youtube.com/watch?v=_lVlC0xzXDc), but it does come at  
  the cost of composability.                                          
- A logical follow-on to this is to provide a mechanism to use a      
  short-hand function which automatically searches the entire         
  container. This could automatically use the container-provided      
  version if available so that functions which provide their own      
  optimized version get the benefit.

### DIFF
--- a/AK/Find.h
+++ b/AK/Find.h
@@ -1,0 +1,57 @@
+/*
+ * Copyright (c) 2021, the SerenityOS developers.
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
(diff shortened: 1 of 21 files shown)


AK/Find.h
```diff
@@ -0,0 +1,57 @@
+/*
+ * Copyright (c) 2020-2021, the SerenityOS developers.
```
linusg, line 2:
```suggestion
 * Copyright (c) 2021, the SerenityOS developers.
```

AK/Tests/TestDoublyLinkedList.cpp
```diff
@@ -0,0 +1,65 @@
+/*
+ * Copyright (c) 2020, the SerenityOS developers.
```
linusg, line 2:
```suggestion
 * Copyright (c) 2021, the SerenityOS developers.
```

AK/Tests/TestFind.cpp
```diff
@@ -0,0 +1,76 @@
+/*
+ * Copyright (c) 2020, the SerenityOS developers.
```
linusg, line 2:
```suggestion
 * Copyright (c) 2021, the SerenityOS developers.
```

AK/Tests/TestSinglyLinkedList.cpp
```diff
@@ -0,0 +1,83 @@
+/*
+ * Copyright (c) 2020, the SerenityOS developers.
```
linusg, line 2:
```suggestion
 * Copyright (c) 2021, the SerenityOS developers.
```

awesomekling:
Yup this seems very nice. Please squash all those tiny fixups into the relevant preceding commit though.

Also note that you are free to add your own name to the copyright header. The "SerenityOS developers" name is used by people who for some reason don't wish to claim personal copyright. :)

ldm5180:
Squash and rebase complete.

Kernel/Devices/Device.h
```diff
@@ -75,7 +75,7 @@ class Device : public File {
         {
             ScopedSpinLock lock(m_requests_lock);
             was_empty = m_requests.is_empty();
-            m_requests.append(request);
+            m_requests.append(RefPtr<AsyncDeviceRequest>(request));
```
awesomekling, line 78: This change feels like an ergonomics regression. Can we avoid forcing clients to do this?

ldm5180: Yes. I was being overly prescriptive, I guess. The `static_assert` in `DoublyLinkedList`'s constructor and `append()` were requiring the type being passed in to be the same as the type in the container. This is not really required though, as long as it is convertible. I will change that `static_assert` to check for conversion rather than same-ness.

The `static_assert` *could* be removed completely, but in doing so I think the compiler errors might get obtuse. My reason for adding the `static_assert` is to ensure a good compiler error message as close to the problem site as possible.

ldm5180: New commits pushed and ready for re-review.

awesomekling: Cool :)

PR #4895, opened 2021-01-10
SystemMonitor: Make graphs look nice with dark themes

linusg:
image

image

image

The Default theme remains unaffected, although the line background opacity may be slightly different.

image

### DIFF
--- a/Applications/SystemMonitor/GraphWidget.cpp
+++ b/Applications/SystemMonitor/GraphWidget.cpp
@@ -27,7 +27,9 @@
 #include "GraphWidget.h"
 #include <LibGUI/Painter.h>
 #include <LibGfx/Font.h>
+#include <LibGfx/Palette.h>
 #include <LibGfx/Path.h>
+#include <LibGfx/SystemTheme.h>
 
 GraphWidget::GraphWidget()
 {
@@ -49,7 +51,7 @@ void GraphWidget::paint_event(GUI::PaintEvent& event)
     GUI::Painter painter(*this);
     painter.add_clip_rect(event.rect());
     painter.add_clip_rect(frame_inner_rect());
-    painter.fill_rect(event.rect(), m_background_color);
+    painter.fill_rect(event.rect(), palette().color(ColorRole::Base));
 
     auto inner_rect = frame_inner_rect();
     float scale = (float)inner_rect.height() / (float)m_max;
--- a/Applications/SystemMonitor/GraphWidget.h
+++ b/Applications/SystemMonitor/GraphWidget.h
@@ -39,8 +39,6 @@ public:
 
     void add_value(Vector<int, 1>&&);
 
-    void set_background_color(Color color) { m_background_color = color; }
-
     struct ValueFormat {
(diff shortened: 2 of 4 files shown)


ccapitalK:
Hmm, I think we need to improve the colour choices for the text here.The green text is very hard to read in the first example, as is the blue text in the third. I believe we should do one of the following: 

- Lean into making the graph look consistent with the system theme, by using colour properties from the themes to determine what colours are used to render the text and lines (adding new properties if needed)
- Stop treating those graphs as themable UI components altogether and instead make the rendering of the graph consistent between all themes, by hardcoding colour choices for everything in it

linusg:
> Hmm, I think we need to improve the colour choices for the text here.The green text is very hard to read in the first example, as is the blue text in the third.

Fair enough.

> Lean into making the graph look consistent with the system theme, by using colour properties from the themes to determine what colours are used to render the text and lines (adding new properties if needed)

Sounds good, feel free to add these new properties to all themes to improve legibility of these labels :)

> Stop treating those graphs as themable UI components altogether and instead make the rendering of the graph consistent between all themes, by hardcoding colour choices for everything in it

That's how it used to be. The black background we had for a long time was an ok solution IMO and works with light and dark themes, but a large white area just doesn't look good with dark themes at all.

tomuta:
I don't really have any opinion on this. I don't have any hard feelings if we don't want the white background color. Personally, I prefer it for fill graphs, it has more a graph on paper feel to it I guess. But yeah, it doesn't look good in the darker themes. But not sure what would look better. I like the screenshots above, fwiw

ccapitalK:
> > Lean into making the graph look consistent with the system theme, by using colour properties from the themes to determine what colours are used to render the text and lines (adding new properties if needed)
> 
> Sounds good, feel free to add these new properties to all themes to improve legibility of these labels :)

I looked through the themes, it appears that the `Syntax*` theme elements basically already fill this niche (being colors that are chosen to look nice on top of `ColorRole::Base`). You could for example use something like `SyntaxComment` instead of the hardcoded green color.

I do agree that the white graphs don't look as nice in dark themes

Applications/SystemMonitor/GraphWidget.cpp
```diff
@@ -49,7 +51,7 @@ void GraphWidget::paint_event(GUI::PaintEvent& event)
     GUI::Painter painter(*this);
     painter.add_clip_rect(event.rect());
     painter.add_clip_rect(frame_inner_rect());
-    painter.fill_rect(event.rect(), m_background_color);
+    painter.fill_rect(event.rect(), palette().color(ColorRole::Base));
```
ccapitalK, line 54: Nitpick: Wouldn't it make more sense to use `palette().base()` here?

ccapitalK:
I tried using the theme colors, they don't look too bad. This is a rough mockup:

2021-01-12-114506_2048x1590_scrot
2021-01-12-114655_2048x1590_scrot
2021-01-12-114751_2048x1590_scrot
2021-01-12-115233_2048x1590_scrot

PR #4897, opened 2021-01-11
LibGUI: Update TextEditor after triple-click to select

campital:
Previously, triple-clicking did not visually update `TextEditor`, so there was sometimes a delay in selecting the whole line. For some reason, this delay was only noticeable after triple-clicking the last word in a line. Now, the widget is updated after triple-clicking.


awesomekling:
Nice catch! But this is a LibGUI change, not LibGfx :)

campital:
Oops!

PR #4899, opened 2021-01-11
Shell: Add (basic) support for history event designators

alimpfard:
This is a (very basic) implementation of history events, up to what seemed reasonable to have - i.e. no modifiers.

~~There are quite a few surprises here that are worth mentioning:~~
- ~~Execute nodes (such as `$(...)`) are not well-behaved when expanding history events~~
- ~~Any side-effect (even outside the substituted range) may or may not happen, depending on whether it's lazy or not~~
- ~~redirections and such are not substituted (`echo foo > bar`, `sudo !!` -> `sudo echo foo` - the `> bar` is lost)~~

Very important note to remember, unlike bash, this does not split actual words, but rather singular expressions, so index 1 in `echo > foo` is `> foo`, not `>`.
Bash (and whatever copies it) isn't very consistent with this concept of "words", `$(foo bar baz)` is a "word" (and so is `$( foo bar baz )`), but `> foo` is two words. figure _that_ out.

With that said, the errors are really nice:
```
[test@turingcomplete] ~/Workspace/serenity/Build
❯ echo !yo
Shell Syntax Error: History event did not match any entry
  0| echo !yo
~~~~~~~~~~~^^
  1|


[test@turingcomplete] ~/Workspace/serenity/Build
❯ echo !!:0-100
History entry resolved to 'clear'
Shell Syntax Error: History word index out of bounds
  0| echo !!:0-100
~~~~~~~~~~~~~~~^^^
  1|


[test@turingcomplete] ~/Workspace/serenity/Build
❯ echo !10:^-10
History entry resolved to 'ninja'
Shell Syntax Error: History word index out of bounds
  0| echo !10:^-10
~~~~~~~~~~~~~~~~^^
  1|

```

### DIFF
--- a/Base/usr/share/man/man5/Shell.md
+++ b/Base/usr/share/man/man5/Shell.md
@@ -28,6 +28,10 @@ String of characters that are not _Special_ or _Syntactic Elements_
 ##### Glob
 String of characters containing at least one of `*?` in _bareword_ position
 
+##### History Events
+A _designator_ starting with `!` in _bareword_ position that describes a word or a range of words from a previously entered command.
+Please look at the section named 'History Event Designators' for a more thorough explanation. Only allowed in interactive mode.
+
 ##### Single Quoted String
 Any sequence of characters between two single quotes (`'`)
 
@@ -36,6 +40,7 @@ Any sequence of _Double Quoted String Part_ tokens:
 * Barewords
 * Single Quotes
 * Variable References
+* History Events
 * Evaluate expressions
 * Escaped sequences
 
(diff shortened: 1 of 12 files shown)


Shell/AST.cpp
```diff
@@ -1200,6 +1200,175 @@ Glob::~Glob()
...
+        if (it.is_end()) {
+            shell->raise_error(Shell::ShellError::EvaluatedSyntaxError, "History event did not match any entry", m_selector.event.text_position);
+            return create<AST::ListValue>({});
+        }
+        resolved_history = *it;
+        break;
+    }
+    }
+
+    dbg() << "History entry resolved to '" << resolved_history << "'";
```
alimpfard, line 1305: (yes, yes, `dbgln()` and such)

PR #4900, opened 2021-01-11
AK: Enable format string warnings for AK printf wrappers

ccapitalK:
This enables printf warnings for StringBuilder::appendf and String::format, which are both used quite a lot in the code base. A bunch of bugs have also been caught regarding incorrect printf usage, including some with omitted parameters, some with missing format specifiers and one call to String::format with a String::formatted format string.

### DIFF
--- a/AK/String.h
+++ b/AK/String.h
@@ -244,7 +244,7 @@ public:
         return String((const char*)buffer.data(), buffer.size(), should_chomp);
     }
 
-    static String format(const char*, ...);
+    static String format(const char*, ...) __attribute__((format(printf, 1, 2)));
 
     static String vformatted(StringView fmtstr, TypeErasedFormatParams);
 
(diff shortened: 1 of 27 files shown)


awesomekling:
While this is not bad on its own, do note that we are moving away from all printf-style formatting functions and these will be deleted as soon as we've ported the code to `String::formatted` et al :)

ccapitalK:
> While this is not bad on its own, do note that we are moving away from all printf-style formatting functions and these will be deleted as soon as we've ported the code to `String::formatted` et al :)

Oh, I was wondering why there were so many commits switching them over. Anyway, since it might take a while to get there (I'm counting around 500 usages of appendf and String::format), I figure this might be a useful stopgap.

asynts:
Adding my two cents:

`appendf` and `printf` will have to stay in the codebase, because sometimes we need the printf behavior, for example for the `printf(1)` utility.

The idea is to only use them when needed and probably to rename `StringBuilder::appendf` to `StringBuilder::printf` or something like that.

awesomekling:
Yes, of course we still need `printf` proper. But its use in our own C++ code should be extremely limited.

PR #4903, opened 2021-01-11
LibCore: Add line iterators to IODevice / LibLine: Handle history across multiple concurrent sessions better 

alimpfard:
- Store history entries as (timestamp)::(entry)\n\n
- Merge the entries together when saving to avoid loss of history
  entries

To ideally make having two concurrently open shells
(or `js` repls or whatever) not overwrite each others' history entries.

(Only applies to users of the lagom version of `shell` and `js`)
it should be noted that the old history format is not going to work anymore, so pls either manually insert newlines between the entries (missing timestamps are okay), or just drop the history file altogether.

### DIFF
--- a/Libraries/LibCore/IODevice.h
+++ b/Libraries/LibCore/IODevice.h
@@ -31,6 +31,28 @@
 
 namespace Core {
 
+// This is not necessarily a valid iterator in all contexts,
+// if we had concepts, this would be InputIterator, not Copyable, Movable.
+class LineIterator {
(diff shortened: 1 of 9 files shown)


Libraries/LibCore/IODevice.h
```diff
@@ -31,6 +31,28 @@
 
 namespace Core {
 
+// This is not necessarily a valid iterator in all contexts,
+// if we had concepts, this would be a
```
alimpfard, line 35: It seems my sentence just

PR #4906, opened 2021-01-11
Playground: Show placeholders for unregistered widgets

nico:
With this, Playground can be used to interactively edit e.g.
DisplaySettingsWindow.gml.

### DIFF
--- a/Libraries/LibGUI/Widget.cpp
+++ b/Libraries/LibGUI/Widget.cpp
@@ -954,13 +954,21 @@ void Widget::set_override_cursor(Gfx::StandardCursor cursor)
 
 bool Widget::load_from_gml(const StringView& gml_string)
 {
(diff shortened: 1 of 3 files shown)


Libraries/LibGUI/Widget.cpp
```diff
@@ -953,14 +953,22 @@ void Widget::set_override_cursor(Gfx::StandardCursor cursor)
 }
 
 bool Widget::load_from_gml(const StringView& gml_string)
+{
+    return load_from_gml(gml_string, [](const String& class_name) -> RefPtr<Widget> {
+        dbg() << "Class '" << class_name << "' not registered";
```
awesomekling, line 958: pls2use `dbgln()` in new code

nico: but but but it's old code just in a new location

nico: done

PR #4907, opened 2021-01-11
Everywhere: Continue replacing dbg() dbgln(). (3)

asynts:
Follow up to #4870 and #4887.

Before:
~~~none
$ grep -rnI 'dbg()' | wc -l
728
~~~

After:
~~~none
$ grep -rnI 'dbg()' | wc -l
673
~~~

### DIFF
--- a/Libraries/LibC/stdio.cpp
+++ b/Libraries/LibC/stdio.cpp
@@ -953,8 +953,8 @@ int snprintf(char* buffer, size_t size, const char* fmt, ...)
 void perror(const char* s)
 {
     int saved_errno = errno;
-    dbg() << "perror(): " << s << ": " << strerror(saved_errno);
-    fprintf(stderr, "%s: %s\n", s, strerror(saved_errno));
+    dbgln("perror(): {}: {}", s, strerror(saved_errno));
+    warnln("{}: {}", s, strerror(saved_errno));
 }
 
 static int parse_mode(const char* mode)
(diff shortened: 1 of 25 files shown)


asynts:
Not sure why the build fails, it seems my `clang-format` formatted that correctly. Maybe that file wasn't touched when we updated to `clang-format` 11?

~~~diff
diff --git a/Libraries/LibGfx/BitmapFont.cpp b/Libraries/LibGfx/BitmapFont.cpp
index f2cbb62..e5e24cb 100644
--- a/Libraries/LibGfx/BitmapFont.cpp
+++ b/Libraries/LibGfx/BitmapFont.cpp
@@ -43,8 +43,7 @@
 
 namespace Gfx {
 
-struct [[gnu::packed]] FontFileHeader
-{
+struct [[gnu::packed]] FontFileHeader {
     char magic[4];
     u8 glyph_width;
     u8 glyph_height;
~~~

linusg:
>  it seems my clang-format formatted that correctly.

No, it didn't :)

It changed

```
struct [[gnu::packed]] FontFileHeader {
```
(clang-format-11 formatting)

to
```
struct [[gnu::packed]] FontFileHeader
{
```
(clang-format-10 formatting because it can't handle attributes properly in some locations)

asynts:
@linusg Oops, I think I looked at the enum in `BitmapFont.h` instead of `BitmapFont.cpp`. That should be fixed now.

I do have clang format 11 installed though, which should do this correctly? Whatever.

PR #4908, opened 2021-01-11
AK: Simplify constructors and conversions from nullptr_t

ldm5180:
Problem:
- Many constructors are defined as `{}` rather than using the ` =
  default` compiler-provided constructor.
- Some types provide an implicit conversion operator from `nullptr_t`
  instead of requiring the caller to default construct. This violates
  the C++ Core Guidelines suggestion to declare single-argument
  constructors explicit
  (https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c46-by-default-declare-single-argument-constructors-explicit).

Solution:
- Change default constructors to use the compiler-provided default
  constructor.
- Remove implicit conversion operators from `nullptr_t` and change
  usage to enforce type consistency without conversion.

### DIFF
--- a/AK/ByteBuffer.h
+++ b/AK/ByteBuffer.h
@@ -27,6 +27,7 @@
 #pragma once
 
 #include <AK/NonnullRefPtr.h>
+#include <AK/OwnPtr.h>
 #include <AK/RefCounted.h>
 #include <AK/RefPtr.h>
 #include <AK/Span.h>
@@ -42,6 +43,7 @@ public:
     static NonnullRefPtr<ByteBufferImpl> create_zeroed(size_t);
     static NonnullRefPtr<ByteBufferImpl> copy(const void*, size_t);
 
+    ByteBufferImpl() = delete;
     ~ByteBufferImpl() { clear(); }
 
     void clear()
@@ -92,7 +94,6 @@ public:
 private:
     explicit ByteBufferImpl(size_t);
     ByteBufferImpl(const void*, size_t);
-    ByteBufferImpl() { }
 
     u8* m_data { nullptr };
     size_t m_size { 0 };
@@ -100,8 +101,7 @@ private:
 
 class ByteBuffer {
 public:
-    ByteBuffer() { }
-    ByteBuffer(std::nullptr_t) { }
+    ByteBuffer() = default;
     ByteBuffer(const ByteBuffer& other)
         : m_impl(other.m_impl)
     {
@@ -159,11 +159,35 @@ public:
     u8* data() { return m_impl ? m_impl->data() : nullptr; }
     const u8* data() const { return m_impl ? m_impl->data() : nullptr; }
 
-    Bytes bytes() { return m_impl ? m_impl->bytes() : nullptr; }
-    ReadonlyBytes bytes() const { return m_impl ? m_impl->bytes() : nullptr; }
+    Bytes bytes()
+    {
+        if (m_impl) {
+            return m_impl->bytes();
+        }
+        return {};
+    }
+    ReadonlyBytes bytes() const
+    {
+        if (m_impl) {
+            return m_impl->bytes();
+        }
+        return {};
+    }
 
-    Span<u8> span() { return m_impl ? m_impl->span() : nullptr; }
-    Span<const u8> span() const { return m_impl ? m_impl->span() : nullptr; }
+    Span<u8> span()
+    {
+        if (m_impl) {
+            return m_impl->span();
+        }
+        return {};
+    }
+    Span<const u8> span() const
+    {
+        if (m_impl) {
+            return m_impl->span();
+        }
+        return {};
+    }
 
     u8* offset_pointer(int offset) { return m_impl ? m_impl->offset_pointer(offset) : nullptr; }
     const u8* offset_pointer(int offset) const { return m_impl ? m_impl->offset_pointer(offset) : nullptr; }
--- a/AK/Span.h
+++ b/AK/Span.h
@@ -105,10 +105,7 @@ class Span : public Detail::Span<T> {
 public:
     using Detail::Span<T>::Span;
 
-    ALWAYS_INLINE constexpr Span(std::nullptr_t)
-        : Span()
-    {
-    }
+    constexpr Span() = default;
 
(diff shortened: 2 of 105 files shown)


AK/ByteBuffer.h
```diff
@@ -92,16 +94,14 @@ class ByteBufferImpl : public RefCounted<ByteBufferImpl> {
 private:
     explicit ByteBufferImpl(size_t);
     ByteBufferImpl(const void*, size_t);
-    ByteBufferImpl() { }
 
-    u8* m_data { nullptr };
+    u8* m_data {};
```
asynts, line 98: This looks weird. I mean, it probably does the right thing, but I wouldn't know that from looking at it.

AK/Span.h
```diff
@@ -105,10 +105,7 @@ class Span : public Detail::Span<T> {
 public:
     using Detail::Span<T>::Span;
 
-    ALWAYS_INLINE constexpr Span(std::nullptr_t)
```
asynts, line 108: I think `nullptr` is much more readable than `{}`. I'd keep this.

AK/ByteBuffer.h
ldm5180: It is just initialization using uniform initialization syntax. This one could be changed back... No change here. I think I changed this because I was experimenting with something else and forgot to undo it.

AK/Span.h
ldm5180: This constructor should at least be `explicit` since it takes only a single argument. However, it is unnecessary to even provide it. It is redundant with the default constructor and lets callers do implicit conversions. Removing this adds type safety (only a little bit though). 

 https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es23-prefer-the--initializer-syntax
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c46-by-default-declare-single-argument-constructors-explicit

Services/WindowServer/WindowManager.h
```diff
@@ -169,8 +169,21 @@ class WindowManager : public Core::Object {
...
     void start_window_resize(Window&, const MouseEvent&);
 
-    const Window* active_fullscreen_window() const { return (m_active_window && m_active_window->is_fullscreen()) ? m_active_window : nullptr; }
-    Window* active_fullscreen_window() { return (m_active_window && m_active_window->is_fullscreen()) ? m_active_window : nullptr; }
+    const Window* active_fullscreen_window() const
+    {
+        if (m_active_window && m_active_window->is_fullscreen()) {
+            return m_active_window;
+        }
+        return {};
```
emanuele6, line 177: doesn't this function return a regular pointer? why are we using `{}` instead of `nullptr`?

Services/WindowServer/WindowManager.h
```diff
@@ -169,8 +169,21 @@ class WindowManager : public Core::Object {
...
+    };
+
+    Window* active_fullscreen_window()
+    {
+        if (m_active_window && m_active_window->is_fullscreen()) {
+            return m_active_window;
+        }
+        return {};
```
emanuele6, line 185: doesn't this function return a regular pointer? why are we using `{}` instead of `nullptr`?

Services/WindowServer/WindowManager.h
ldm5180: Yes. I will change that.

The reason for the initial change was that the conversion operator from `std::nullptr_t` to `WeakPtr<T>` is removed so when the `?:` operator is used and returns different types (not convertible) the compiler complains. Note, the `m_active_window` is a `WeakPtr<Window>` and `nullptr` is a `std::nullptr_t`.

Services/WindowServer/WindowManager.h
ldm5180: Yes. I will change that.

The reason for the initial change was that the conversion operator from `std::nullptr_t` to `WeakPtr<T>` is removed so when the `?:` operator is used and returns different types (not convertible) the compiler complains. Note, the `m_active_window` is a `WeakPtr<Window>` and `nullptr` is a `std::nullptr_t`.

AK/ByteBuffer.h
ldm5180: Reverted.

Services/WindowServer/WindowManager.h
emanuele6: We can also just leave that file alone, there is no need to rewrite these functions without the ternary operator.

Services/WindowServer/WindowManager.h
emanuele6: @ldm5180: We can also just leave that file alone, there is no need to rewrite these functions without the ternary operator.

Services/WindowServer/WindowManager.h
ldm5180: It does not work using the ternary operator as I described above. This changeset removes the conversion operator from a `std::nullptr_t` to `WeakPtr<Window>`. Therefore, the ternary operator, which used to return a `std::nullptr_t` getting automatically converted to a `WeakPtr<Window>` no longer converts. A ternary operator must return the same type from each sub-expression, so that is why this needs to not be a ternary operator.

This is exactly one of the issues having the conversion operators causes. It leads to this confusion because the conversion from `std::nullptr_t` to `WeakPtr<Window>` is hidden.

The resulting code using the ternary operator (without optimization) would be to call the `WeakPtr<Window>(std::nullptr_t)` constructor and then call the `WeakPtr<Window>` conversion to a `Window*`, but this is hidden. The result now is that `return nullptr;` simply converts to a `Window*`. No construction of the `WeakPtr<Window>` at all.

Services/WindowServer/WindowManager.h
ldm5180: Same as above comment.

Services/WindowServer/WindowManager.h
emanuele6: Ah, that's very subtle. Thanks a lot for explaining it.

ldm5180: I agree it is subtle. IMO, this kind of automatic conversion is ugly and is why I want to remove these constructors.

PR #4911, opened 2021-01-11
VimEditingEngine: Add handling { and } to move between empty lines

rok-povsic:
This makes the Vim emulator handle SHIFT+LeftBrace  and SHIFT+RightBrace to jump to the previous / next empty lines block.

### DIFF
--- a/Userland/Libraries/LibGUI/VimEditingEngine.cpp
+++ b/Userland/Libraries/LibGUI/VimEditingEngine.cpp
@@ -204,6 +204,12 @@ bool VimEditingEngine::on_key_in_normal_mode(const KeyEvent& event)
                 move_one_up(event);
                 switch_to_insert_mode();
                 break;
+            case (KeyCode::Key_LeftBrace):
+                move_to_previous_empty_lines_block();
+                break;
+            case (KeyCode::Key_RightBrace):
+                move_to_next_empty_lines_block();
+                break;
             default:
                 break;
             }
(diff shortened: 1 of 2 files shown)


Libraries/LibGUI/VimEditingEngine.cpp
```diff
@@ -210,4 +216,42 @@ void VimEditingEngine::move_half_page_down(const KeyEvent& event)
     move_down(event, 0.5);
 };
 
+void VimEditingEngine::move_to_previous_empty_lines_block()
+{
+    size_t line_idx = m_editor->cursor().line();
```
emanuele6, line 221:
```suggestion
{
    ASSERT(m_editor != nullptr);
    size_t line_idx = m_editor->cursor().line();
```

Libraries/LibGUI/VimEditingEngine.cpp
```diff
@@ -210,4 +216,42 @@ void VimEditingEngine::move_half_page_down(const KeyEvent& event)
...
+    }
+
+    TextPosition new_cursor = { line_idx, 0 };
+
+    m_editor->set_cursor(new_cursor);
+};
+
+void VimEditingEngine::move_to_next_empty_lines_block()
+{
+    size_t line_idx = m_editor->cursor().line();
```
emanuele6, line 240:
```suggestion
{
    ASSERT(m_editor != nullptr);
    size_t line_idx = m_editor->cursor().line();
```

emanuele6 (changes requested):
Since we are assuming `m_editor != nullptr`, we should assert it.

Libraries/LibGUI/VimEditingEngine.cpp
emanuele6: You can also use:
```cpp
ASSERT(m_editor != nullptr);
auto& editor = *m_editor;
size_t line_idx = editor.cursor().line();
```
To get rid of the ugly `->`s.

emanuele6:
I notice that a lot of other method should `ASSERT(m_editor != nullptr)`, but do not.

Can we figure out a way to make `m_editor` a reference or a NonNull smart poiner? Does it really need to be a `WeakPtr`?

Libraries/LibGUI/VimEditingEngine.cpp
```diff
@@ -107,6 +107,12 @@ bool VimEditingEngine::on_key_in_normal_mode(const KeyEvent& event)
                 move_one_up(event);
                 switch_to_insert_mode();
                 break;
+            case (KeyCode::Key_LeftBrace):
+                move_to_previous_empty_lines_block();
+                break;
+            case (KeyCode::Key_RightBrace):
+                move_to_next_empty_lines_block();
+                break;
```
emanuele6, line 115: This relies too much on the keyboard layout in my opinion.
Can't we use `event.code_point() == '{'` et cetera here?

awesomekling: That seems worse. What does it mean to "rely too much on the keyboard layout"?

alimpfard: If (read: when) at some point we decide to split this off `TextEditor`, and `LibGUI` in general (like my original suggestion), either these key events will have to go, or we'll have to make fake key events somehow.
`KeyCode` seems better than raw characters though, I say we keep that but get rid of the key event?

emanuele6: I mean that this will only behaves like vim  if `{` is `shift + [` (and `ctrl` and `alt` are not held down).

Also, the other bindings that use an uppercase letter (e.g. `O`) will only work if you press `shift + o` and not if you only press `o` with CapsLock activated (unlike vim).

linusg:
@rok-povsic this PR has merge conflicts and the commits need to be squashed as per the [contribution guidelines](https://github.com/SerenityOS/serenity/blob/master/CONTRIBUTING.md).

PR #4913, opened 2021-01-11
wc: Remove a memory leak.

emanuele6:
Also replace uses of `printf`-functions with the new `String::formatted` print functions.

### DIFF
--- a/Userland/wc.cpp
+++ b/Userland/wc.cpp
@@ -85,6 +85,10 @@ static Count get_count(const String& file_specifier)
             count.words++;
         }
     }
+
+    if (file_pointer != stdin)
+        fclose(file_pointer);
+
     return count;
 }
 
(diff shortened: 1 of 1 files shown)


Userland/wc.cpp
```diff
@@ -85,6 +85,12 @@ static Count get_count(const String& file_specifier)
             count.words++;
         }
     }
+
+    if (file_pointer != stdin) {
+        if (fclose(file_pointer) != 0)
+            perror("fclose");
```
emanuele6, line 91: I think this is unnecessary. I'll remove it.

Userland/wc.cpp
```diff
@@ -85,6 +85,10 @@ static Count get_count(const String& file_specifier)
             count.words++;
         }
     }
+
+    if (file_pointer != stdin)
+        fclose(file_pointer);
```
awesomekling, line 90: I'd maybe call this a file leak, not just a memory leak :)
But it's not terribly important.

emanuele6: This is both a file descriptor leak and a memory leak I think. Because `file_pointer` is allocated in the heap.

I found this using `UserspaceEmulator` that called it a memory leak so I just used that term.

Issue #4078, opened 2020-11-14
Userland: Develop a testing framework


bcoles:
Thread for discussing a testing framework for tests located in `Userland` (ie, not `AK/Tests`).

Currently, there are tests for the kernel and libraries (`LibC` only) in `Userland/Tests`; and various test applications in `Userland/test[-_]*`. Each uses a different style for tests as there is no existing consensus on style.

```
$ ls -laR Userland/Tests/ Userland/test*
-rw-rw-r-- 1 user user  4523 Oct 25 01:17 Userland/test-bindtodevice.cpp
-rw-rw-r-- 1 user user  9562 Oct 25 01:17 Userland/test-compress.cpp
-rw-rw-r-- 1 user user 15078 Oct 25 01:17 Userland/test.cpp
-rw-rw-r-- 1 user user 93755 Oct 30 15:47 Userland/test-crypto.cpp
-rw-rw-r-- 1 user user  5877 Oct 25 01:17 Userland/test_efault.cpp
-rw-rw-r-- 1 user user  3503 Oct 25 01:17 Userland/test_env.cpp
-rw-rw-r-- 1 user user  5159 Nov 13 21:35 Userland/test-gfx-font.cpp
-rw-rw-r-- 1 user user 11190 Oct 25 13:01 Userland/test_io.cpp
-rw-rw-r-- 1 user user 24879 Nov 12 01:18 Userland/test-js.cpp
-rw-rw-r-- 1 user user 24630 Nov  1 22:16 Userland/test-web.cpp

Userland/Tests/:
total 20
drwxrwxr-x 4 user user 4096 Oct 25 01:17 .
drwxrwxr-x 3 user user 4096 Nov 14 00:38 ..
-rw-rw-r-- 1 user user   48 Oct 25 01:17 CMakeLists.txt
drwxrwxr-x 2 user user 4096 Nov 13 22:54 Kernel
drwxrwxr-x 2 user user 4096 Nov 11 10:23 LibC

Userland/Tests/Kernel:
total 104
drwxrwxr-x 2 user user 4096 Nov 13 22:54 .
drwxrwxr-x 4 user user 4096 Oct 25 01:17 ..
-rw-rw-r-- 1 user user 2146 Oct 25 01:17 bind-local-socket-to-symlink.cpp
-rw-rw-r-- 1 user user 3541 Oct 25 01:17 bxvga-mmap-kernel-into-userspace.cpp
[...]

PR #4921, opened 2021-01-12
Run on-target tests in QEMU in CI

ADKaster:
This PR  does a few things:

- Reorganize building of AK, Shell, and LibRegex tests.
   - Test/Tests directories are now always included by their parent. Make sure that  we have rules to install tests onto the Serenity rootfs target.
   - Tests that use CTest or `add_test` are moved to the Lagom CMakeLists.txt

- Remove some unnecessary waitid safety checks from Shell
- Fix a memory leak in shuf
- Add a "ci" run target that has minimal hardware such that it can be run in limited docker containers

- Rework CI script into 3 stages: BuildSerenity, BuildLagom, and NotifyIRC
    - BuildSerenity only runs for ubuntu, and has two variants: with and without ALL_DEBUG. Attempting to run on-target tests with ALL_DEBUG on is an exercise in extreme patience.
    - BuildLagom only builds and runs Lagom tests (and the Fuzzer enabled build). This stage runs on ubuntu and Mac.
    - NotifyIRC uses the action https://github.com/marketplace/actions/workflow-conclusion-action from technote-space to aggregate the matrix job results into the same format as before.
    
- Add documentation on how to run tests.

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -9,20 +9,13 @@ env:
   SERENITY_ROOT: ${{ github.workspace }}
 
 jobs:
-  build_and_test:
+  build_and_test_serenity:
     runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
       matrix:
-        include:
-          - os: ubuntu-20.04
-            allow-test-failure: false
-            # There sure is a lot of logic here, is there a better way?
-            # TODO: Make IRC notifications its own job/workflow?
-            should-notify-irc: ${{ github.repository == 'SerenityOS/serenity' && (github.event_name == 'pull_request' || (github.event_name == 'push' && github.ref == 'refs/heads/master')) }}
-          - os: macos-10.15
-            allow-test-failure: true
-            should-notify-irc: false
+        debug-macros: ['ALL_DEBUG', 'NORMAL_DEBUG']
+        os: [ubuntu-20.04]
 
     steps:
     - uses: actions/checkout@v2
@@ -37,7 +30,6 @@ jobs:
     - name: Purge interfering packages
       # Remove GCC 9 and clang-format 10 (installed by default)
       run: sudo apt-get purge -y gcc-9 g++-9 libstdc++-9-dev clang-format-10
-      if: ${{ runner.os == 'Linux' }}
     - name: "Install Ubuntu dependencies"
       # These packages are already part of the ubuntu-20.04 image:
       # cmake gcc-10 g++-10 shellcheck libgmp-dev
@@ -46,17 +38,9 @@ jobs:
         wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key | sudo apt-key add -
         sudo add-apt-repository "deb http://apt.llvm.org/focal/ llvm-toolchain-focal-11 main"
         sudo apt-get update
-        sudo apt-get install clang-format-11 libstdc++-10-dev libmpfr-dev libmpc-dev ninja-build npm
-      # If we ever do any qemu-emulation on Github Actions, we should re-enable this:
-      # e2fsprogs qemu-system-i386 qemu-utils
-      if: ${{ runner.os == 'Linux' }}
+        sudo apt-get install clang-format-11 libstdc++-10-dev libmpfr-dev libmpc-dev ninja-build npm e2fsprogs qemu-utils qemu-system-i386
     - name: Use GCC 10 instead
       run: sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-10 60 --slave /usr/bin/g++ g++ /usr/bin/g++-10
-      if: ${{ runner.os == 'Linux' }}
-
-    - name: Install macOS dependencies
-      run: brew install coreutils ninja
-      if: ${{ runner.os == 'macOS' }}
 
     - name: Install JS dependencies
       run: sudo npm install -g prettier
@@ -72,7 +56,6 @@ jobs:
 
     - name: Lint (Phase 1/2)
       run: ${{ github.workspace }}/Meta/lint-ci.sh
-      if: ${{ runner.os == 'Linux' }}
     - name: Toolchain cache
       uses: actions/cache@v2
       with:
@@ -86,6 +69,15 @@ jobs:
     # TODO: ccache
     # https://cristianadam.eu/20200113/speeding-up-c-plus-plus-github-actions-using-ccache/
     # https://github.com/cristianadam/HelloWorld/blob/master/.github/workflows/build_cmake.yml
+    - name: Create build environment with debug macros
+      working-directory: ${{ github.workspace }}
+      # Build the entire project with debug macros turned on, to prevent code rot.
+      # However, it is unweildy and slow to run tests with them enabled, so we will build twice.
+      run: |
+        mkdir -p Build
+        cd Build
+        cmake .. -GNinja -DBUILD_LAGOM=ON -DENABLE_ALL_THE_DEBUG_MACROS=ON -DCMAKE_C_COMPILER=gcc-10 -DCMAKE_CXX_COMPILER=g++-10
+      if: ${{ matrix.debug-macros == 'ALL_DEBUG' }}
     - name: Create build environment
       working-directory: ${{ github.workspace }}
       # Note that this needs to run *even if* the Toolchain was built,
@@ -93,9 +85,10 @@ jobs:
       run: |
         mkdir -p Build
         cd Build
-        cmake .. -GNinja -DBUILD_LAGOM=ON -DENABLE_ALL_THE_DEBUG_MACROS=ON -DCMAKE_C_COMPILER=gcc-10 -DCMAKE_CXX_COMPILER=g++-10
+        cmake .. -GNinja -DBUILD_LAGOM=ON -DCMAKE_C_COMPILER=gcc-10 -DCMAKE_CXX_COMPILER=g++-10
+      if: ${{ matrix.debug-macros == 'NORMAL_DEBUG' }}
 
-    # === ACTUALLY BUILD AND TEST ===
+    # === ACTUALLY BUILD ===
 
     - name: Build Serenity and Tests
       working-directory: ${{ github.workspace }}/Build
@@ -103,16 +96,106 @@ jobs:
     - name: Lint (Phase 2/2)
       working-directory: ${{ github.workspace }}/Meta
       run: ./check-symbols.sh
-    - name: Run CMake tests
+
+    - name: Create Serenity Rootfs
+      if: ${{ matrix.debug-macros == 'NORMAL_DEBUG'}}
       working-directory: ${{ github.workspace }}/Build
+      run: ninja install && ninja image
+
+    - name: Run On-Target Tests
+      if: ${{ matrix.debug-macros == 'NORMAL_DEBUG'}}
+      working-directory: ${{ github.workspace }}/Build
+      env:
+        SERENITY_KERNEL_CMDLINE: "boot_mode=self-test"
+        SERENITY_RUN: "ci"
+      run: ninja run
+      timeout-minutes: 10
+    - name: Print target logs
+      if: ${{ !cancelled() && matrix.debug-macros == 'NORMAL_DEBUG'}}
+      working-directory: ${{ github.workspace }}/Build
+      run: cat ./debug.log
+
+  build_and_test_lagom:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - with-fuzzers: FUZZ
+            os: ubuntu-20.04
+            allow-test-failure: false
+          - with-fuzzers: NO_FUZZ
+            os: ubuntu-20.04
+            allow-test-failure: false
+          - with-fuzzers: NO_FUZZ
+            os: macos-10.15
+            allow-test-failure: true
+
+    steps:
+    - uses: actions/checkout@v2
+
+    # === OS SETUP ===
+    #
+    - name: Install Ubuntu dependencies
+      run: sudo apt-get install ninja-build
+      if: ${{ runner.os == 'Linux' }}
+    - name: Install macOS dependencies
+      run: brew install ninja
+      if: ${{ runner.os == 'macOS' }}
+    - name: Check versions
+      run: set +e; clang --version; clang++ --version; ninja --version
+
+    # === PREPARE FOR BUILDING ===
+
+    # TODO: ccache
+    # https://cristianadam.eu/20200113/speeding-up-c-plus-plus-github-actions-using-ccache/
+    # https://github.com/cristianadam/HelloWorld/blob/master/.github/workflows/build_cmake.yml
+    - name: Create build environment (fuzz)
+      working-directory: ${{ github.workspace }}/Meta/Lagom
+      run: |
+        mkdir -p Build
+        cd Build
+        cmake -GNinja -DBUILD_LAGOM=ON -DENABLE_FUZZER_SANITIZER=ON -DENABLE_ADDRESS_SANITIZER=ON -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ ..
+      if: ${{ matrix.with-fuzzers == 'FUZZ' }}
+
+    - name: Create build environment (no fuzz)
+      working-directory: ${{ github.workspace }}/Meta/Lagom
+      run: |
+        mkdir -p Build
+        cd Build
+        cmake -GNinja -DBUILD_LAGOM=ON -DCMAKE_C_COMPILER=gcc-10 -DCMAKE_CXX_COMPILER=g++-10 ..
+      if: ${{ matrix.with-fuzzers == 'NO_FUZZ' }}
+
+    # === ACTUALLY BUILD AND TEST ===
+
+    - name: Build Lagom
+      working-directory: ${{ github.workspace }}/Meta/Lagom/Build
+      run: cmake --build .
+
+    - name: Run CMake tests
+      working-directory: ${{ github.workspace }}/Meta/Lagom/Build
       run: CTEST_OUTPUT_ON_FAILURE=1 ninja test || ${{ matrix.allow-test-failure }}
       timeout-minutes: 2
+      if: ${{ matrix.with-fuzzers == 'NO_FUZZ' }}
     - name: Run JS tests
-      working-directory: ${{ github.workspace }}/Build/Meta/Lagom
+      working-directory: ${{ github.workspace }}/Meta/Lagom/Build
+      if: ${{ matrix.with-fuzzers == 'NO_FUZZ' }}
       run: DISABLE_DBG_OUTPUT=1 ./test-js || ${{ matrix.allow-test-failure }}
     - name: Run LibCompress tests
-      working-directory: ${{ github.workspace }}/Build/Meta/Lagom
+      working-directory: ${{ github.workspace }}/Meta/Lagom/Build
       run: ./test-compress
+      if: ${{ matrix.with-fuzzers == 'NO_FUZZ' }}
+
+  notify_irc:
+    needs: [build_and_test_serenity, build_and_test_lagom]
+    runs-on: ubuntu-20.04
+    if: always() && github.repository == 'SerenityOS/serenity' && (github.event_name == 'pull_request' || (github.event_name == 'push' && github.ref == 'refs/heads/master'))
+
+    steps:
+    - uses: actions/checkout@v2
+    # Sets environment variable env.WORKFLOW_CONCLUSION to one of [neutral, success, skipped, cancelled, timed_out, action_required, failure]
+    # depending on result of all needs jobs
+    - uses: technote-space/workflow-conclusion-action@v2
 
     # === NOTIFICATIONS ===
 
@@ -122,43 +205,13 @@ jobs:
       run: |
         cat <<"EOF"
         ${{ toJSON(github.event) }}
+        ${{ toJSON(needs) }}
         EOF
     - name: Generate IRC message
-      if: matrix.should-notify-irc == true && !cancelled()
+      if: always()
       run: |
         ${{ github.workspace }}/Meta/notify_irc.py <<"EOF"
-        ["${{ github.actor }}", ${{ github.run_id }}, "${{ job.status }}",
+        ["${{ github.actor }}", ${{ github.run_id }}, "${{ env.WORKFLOW_CONCLUSION }}",
         ${{ toJSON(github.event) }}
         ]
         EOF
-
-  build_lagom_with_fuzzers:
-    runs-on: ubuntu-20.04
-
-    steps:
-    - uses: actions/checkout@v2
-
-    # === OS SETUP ===
-    #
-    - name: Install dependencies
-      run: sudo apt-get install ninja-build
-    - name: Check versions
-      run: set +e; clang --version; clang++ --version; ninja --version
-
-    # === PREPARE FOR BUILDING ===
-
-    # TODO: ccache
-    # https://cristianadam.eu/20200113/speeding-up-c-plus-plus-github-actions-using-ccache/
-    # https://github.com/cristianadam/HelloWorld/blob/master/.github/workflows/build_cmake.yml
-    - name: Create build environment
-      working-directory: ${{ github.workspace }}/Meta/Lagom
-      run: |
-        mkdir -p Build
-        cd Build
-        cmake -GNinja -DBUILD_LAGOM=ON -DENABLE_FUZZER_SANITIZER=ON -DENABLE_ADDRESS_SANITIZER=ON -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ ..
-
-    # === ACTUALLY BUILD ===
-
-    - name: Build Lagom with Fuzzers
-      working-directory: ${{ github.workspace }}/Meta/Lagom/Build
-      run: cmake --build .
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -110,6 +110,8 @@ jobs:
         SERENITY_RUN: "ci"
       run: ninja run
       timeout-minutes: 10
+      # FIXME: When stable, remove continue on error. (See issue #5541)
+      continue-on-error: true
     - name: Print target logs
       if: ${{ !cancelled() && matrix.debug-macros == 'NORMAL_DEBUG'}}
       working-directory: ${{ github.workspace }}/Build
--- a/Base/etc/SystemServer.ini
+++ b/Base/etc/SystemServer.ini
@@ -4,7 +4,7 @@ SocketPermissions=660
 Lazy=1
 Priority=low
 User=protocol
-BootModes=text,graphical
+BootModes=text,graphical,self-test
 MultiInstance=1
 AcceptSocketConnections=1
 
@@ -39,13 +39,13 @@ Lazy=1
 Priority=low
 KeepAlive=1
 User=lookup
-BootModes=text,graphical
+BootModes=text,graphical,self-test
 
 [DHCPClient]
 Priority=low
 KeepAlive=1
 User=root
-BootModes=text,graphical
+BootModes=text,graphical,self-test
 
 [NotificationServer]
 Socket=/tmp/portal/notify
@@ -175,3 +175,11 @@ AcceptSocketConnections=1
 Executable=/bin/CrashDaemon
 KeepAlive=1
 User=anon
+
+[TestRunner@ttyS0]
+Executable=/home/anon/tests/run-tests-and-shutdown.sh
+StdIO=/dev/ttyS0
+Environment=DO_SHUTDOWN_AFTER_TESTS=1 TERM=xterm PATH=/bin:/usr/bin:/usr/local/bin
+User=anon
+WorkingDirectory=/home/anon
+BootModes=self-test
--- a/Base/home/anon/tests/run-tests-and-shutdown.sh
+++ b/Base/home/anon/tests/run-tests-and-shutdown.sh
@@ -1,0 +1,30 @@
+#!/bin/sh
+
+echo "==== Running Tests on SerenityOS ===="
+
+run() {
+    test_cmd=($*)
+    echo "Running test -- $test_cmd"
+    if $test_cmd {
+        echo "::debug file=$test_cmd:: $test_cmd passed!"
+    } else {
+        echo "::error file=$test_cmd:: $test_cmd returned non-zero exit code, check logs!"
+    }
+}
+
+# TODO: It'd be nice to have a list+list op (as opposed to nest-on-in-another)
+# TODO: It'd be nice to have a list.length or enumerate(list) operation to allow terminal progress bar
+# TODO: test-web requires the window server
+system_tests=(test-js test-pthread test-compress (test-crypto bigint -t))
+# FIXME: Running too much at once is likely to run into #5541. Remove commented out find below when stable
+all_tests=($system_tests) #$(find /usr/Tests -type f | grep -v Kernel | grep -v .inc | shuf))
+
+for list in $all_tests {
+    for $list { run $it }
+}
+
+echo "==== Done running tests ===="
+
+if test $DO_SHUTDOWN_AFTER_TESTS {
+    shutdown -n
+}
--- a/Documentation/RunningTests.md
+++ b/Documentation/RunningTests.md
@@ -1,0 +1,87 @@
+## Running SerenityOS Tests
+
+There are two classes of tests built during a Serenity build: host tests and target tests. Host tests run on the build
+machine, and use [Lagom](../Meta/Lagom/README.md) to build Serenity userspace libraries for the host platform. Target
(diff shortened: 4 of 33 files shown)


.github/workflows/cmake.yml
```diff
@@ -87,14 +85,23 @@ jobs:
     # TODO: ccache
     # https://cristianadam.eu/20200113/speeding-up-c-plus-plus-github-actions-using-ccache/
     # https://github.com/cristianadam/HelloWorld/blob/master/.github/workflows/build_cmake.yml
+    - name: Create build environment with debug macros
+      working-directory: ${{ github.workspace }}
+      # Build the entire project with debug macros turned on, to prevent code rot.
+      # However, it is unweildy and slow to run tests with them enabled, so we will buld twice.
```
8infy, line 91: typo `buld`

Base/etc/SystemServer.ini
```diff
@@ -169,3 +169,11 @@ AcceptSocketConnections=1
...
 KeepAlive=1
 User=anon
+
+# Uncomment to test what CI does
+#[TestRunner@ttyS0]
+#Executable=/bin/Shell
+#StdIO=/dev/ttyS0
+#Arguments=-c /home/anon/tests/run-tests-and-shutdown.sh
+#Environment=TERM=xterm PATH=/bin:/usr/bin:/usr/local/bin
+#BootModes=text
```
linusg, line 179: This belongs in a documentation file (manpage or `Documentation/`) rather than commented out in the config IMO.

Userland/Utilities/test-js.cpp
```diff
@@ -713,6 +719,7 @@ int main(int argc, char** argv)
     args_parser.add_option(print_times, "Show duration of each test", "show-time", 't');
     args_parser.add_option(collect_on_every_allocation, "Collect garbage after every allocation", "collect-often", 'g');
     args_parser.add_option(test262_parser_tests, "Run test262 parser tests", "test262-parser-tests", 0);
+    args_parser.add_option(format_errors_for_ci, "Output error messages that CI understands", "ci-output", 'c');
```
linusg, line 722: This is GitHub Actions specific, so please reflect that in the description, flag name and/or variable name :)

ADKaster: Mmk, if I keep this change for test-js I'll make sure that's noted.

I'm considering whether to just make the shell script do the fancy notifications, or if it's worth the "ugliness" to embed some knowledge of github actions into the test apps?

```sh
#!/bin/sh

echo "==== Running Tests on SerenityOS ===="
test-js
if [ $? -ne 0];
    echo "::error file=test-js:: Test returned non-zero exit code, refer to logs"
fi

echo "==== Done running tests, shutting down.... ===="
shutdown -n
```

alimpfard: (Yes, I know this was 4 days ago, consider this a nudge)
Friendly reminder that the equivalent in /bin/Shell would be this:
```sh
echo "..."
if test-js {}
else {
    echo "::error file=blah:: 'if' needs negation"
}

echo "yoinks"
shutdown -n
```

alimpfard: TIL:
```sh
$ not ls 2>&1 >/dev/null; echo $?
1

$ pacman -Qo /usr/bin/not
/usr/bin/not is owned by llvm 11.0.1-1
```

ADKaster: I figured you'd chime in with the correct form eventually :p. I think that doing the shell thing is the way to go. But ideally, I'd want to create a list of "all tests" and just do this for all tests. Opened up 2 shell issues that either point out deficiencies in the shell's for loops/if statements, or in my reading of the docs before coffee as a result of messing around with this.

ADKaster: Something like the shell equivalent of this python snippet:

```python
print("Starting the tests yo")
all_tests=[ ??? ] # tbd method for aggregating tests

for idx, test in enumerate(all_tests):
    print("Running test {}/{} -- {}", idx, len(all_tests), test)
    if magic_exec(test):
        print("::info file={0}:: {0} passed!", test)
    else:
        print("::error file={0}:: {0} failed!", test)
        
print("done with the tests")
magic_exec("shutdown -n")
```

alimpfard: hmmm...indices in a for loop :thinking:
I guess you'll have to go the hard way for that, `i=0` before the loop and `i=$(expr $i + 1)` inside the loop.
alternatively implement `-n` for `cat` then use `match` to split the index (I've had thoughts about allowing `for match pattern as (names) in expr` and something similar for normal variable declarations, but I don't have anything concrete).

the rest should be fine though:
```sh
echo Starting the tests yoooooo
all_tests=( ??? )
# (list count doesn't exist yet, but you can cheese it:)
count=$(printf %s\n $all_tests | wc -l)
idx=0
for test in $all_tests {
    echo "Running test $idx/$count -- $test"
    if $test {
		echo whatever $test $test $test
	}
    else {
        echo on noes we broke it
    }
    idx=$(expr $idx + 1)
}
echo Done with the toasts
shutdown -n
```

ADKaster: Either I'm missing something (seems to be a trend ;) ) or there's some more shell bugs going on here... This crashes the shell with a segfault, even with only one test. Also the spacing on the count is super weird.

Note this is text mode though.

```sh
#!/bin/sh

echo "==== Running Tests on SerenityOS ===="

# TODO: Enable all tests when shell stops crashing :(
#IFS='\n'
#usr_tests_tests=($(find /usr/Tests -type f))
#unset IFS
#all_tests=("test-js" "test-web" "test-pthread" "test-crypto" "test-compress" $usr_tests_tests)
all_tests=("test-js")

# TODO: List size checking when?
count=$(printf %s\n $all_tests | wc -l)
idx=1

for test_name in $all_tests {
    echo "Running test $idx/$count -- $test_name"
    if $test_name {
        echo "::info file=$test_name:: $test_name passed!"
    } else {
        echo "::error file=$test_name:: $test_name returned non-zero exit code, check logs!"
    }
    idx=$(expr $idx + 1)
}

echo "==== Done running tests, shutting down.... ===="
shutdown -n
```

spacing
crashy-crash

ADKaster: Wait, wat is all this?
```
[expr(55:55)]: expr(55:55) Should already die
Shell(56): Job entry "expr 1 + 1" deleted in 251 ms
[Shell(56:56)]: File is not an ELF file.
[Shell(56:56)]: exec(/bin/): File has invalid ELF header
Shell(56): execvpe() failed on attempt (/bin/) with Exec format error
Shell(56): execvp() about to return -1 with errno=8
[Shell(56:56)]: Shell(56:56) Should already die
```

alimpfard: That's some weird looking error

alimpfard: Ohhh
Our `wc` is emitting extra spaces for some reason

alimpfard: I'm not seeing a segfault, but things are really weird alright

alimpfard: Something is fishy here, why is a subshell also being notified of these job completions, and why does it even have access to them? so many bugs

Also, I don't think we have support for nested lists yet, and you can't run a string like in bash (it won't split, this was a design decision - and I still think it's a good one).
you _can_ get around that by doing something really weird:
```sh
IFS=" "
if $($(echo $test_string)) { } else { }
IFS="\n"
```

alimpfard: Let's take a step back and fix bugs instead of making workarounds, silly me.

Base/etc/SystemServer.ini
ADKaster: That's a good point. I'll look around to see which documentation it belongs in, and if there's no obvious home i'll write up a short doc on how to run the tests in serenity in text mode locally.

Userland/Utilities/test-js.cpp
alimpfard: Okay, so with #4996, this works:
```sh
#!/bin/sh

echo "==== Running Tests on SerenityOS ===="

run() {
    test_cmd=($*)
    echo "Running test $idx/$count -- $test_cmd"
    if $test_cmd {
        echo "::info file=$test_cmd:: $test_cmd passed!"
    } else {
        echo "::error file=$test_cmd:: $test_cmd returned non-zero exit code, check logs!"
    }
    idx=$(expr $idx + 1)
}

# TODO: It'd be nice to have a list+list op (as opposed to nest-on-in-another)
system_tests=(test-js (test-crypto bigint -t)) #("test-js" "test-web" "test-pthread" "test-crypto" "test-compress" $usr_tests_tests)
usr_tests_tests=$(find /usr/Tests -type f)
all_tests=($system_tests $usr_tests_tests)

count=0
for list in $all_tests {
    for $list { count=$(expr $count + 1) }
}

idx=1
for list in $all_tests {
    for $list { run $it }
}

echo "==== Done running tests, shutting down.... ===="
shutdown -n
```

On one hand, I like the fact that lists aren't magically flattened until you pass them to a command, and on the other hand, we get these nested for loops...:thinking:

ADKaster: Looks like that does work, yeah. The loop to count all the files runs... very very slowly on my setup (No KVM or windows hyperv accel). Lotta shells to spawn to count all 80 tests. The tests don't all work though. Missing some way to either

```
$(find /usr/Tests -type f | grep -v Kernel)
```
or
```
$(find /usr/Tests -not -path /usr/Tests/Kernel -type f)
```
Now :D Not sure which would be easier to implement... probably `grep -v`

alimpfard: Yeah, I imagine it'd be quite slow, really should figure out the list length thing, but I've absolutely hated all the attempts at "variable substitution" so far

If you don't feel like grep -v, you could try using a shell function (and probably finding a few more bugs)

ADKaster: you're not wrong about the bugs :D

With my commits from #5089 applied... 

Runs ~80 tests, but some of them are not tests but instead proof of concept ad hoc feature integration programs that no longer work:
```sh
system_tests=(test-js (test-crypto bigint -t)) #("test-js" "test-web" "test-pthread" "test-crypto" "test-compress" $usr_tests_tests)
usr_tests_tests=$(find /usr/Tests -type f)
all_tests=($system_tests $usr_tests_tests)
```
Segmentation fault in the Shell:
```sh
system_tests=(test-js (test-crypto bigint -t)) #("test-js" "test-web" "test-pthread" "test-crypto" "test-compress" $usr_tests_tests)
usr_tests_tests=$(find /usr/Tests -type f | grep -v Kernel)
all_tests=($system_tests $usr_tests_tests)
```

Runs two tests, but the usr_test_tests variable refuses to turn into an array no matter what I try (Shell inexperience showing...):
```sh
system_tests=(test-js (test-crypto bigint -t)) #("test-js" "test-web" "test-pthread" "test-crypto" "test-compress" $usr_tests_tests)
usr_tests_tests=($(find /usr/Tests -type f | grep -v Kernel))
all_tests=($system_tests $usr_tests_tests)
```

ADKaster: Er ok I'm confused. Adding `#define ENSURE_WAITID_ONCE` at the top of Shell.cpp fixes the crash on the second one. but.. there's already an `#ifdef __serenity__ #define ENSURE_WAITID_ONCE` ...?

alimpfard: (while I rebuild things)

a segfault _inside_ waitid? that doesn't sound like a shell bug (wait.cpp:100 is the syscall)

> usr_test_tests variable refuses to turn into an array

it already is a list though? `$()` will always be a list, no need to put it in another list

Base/etc/SystemServer.ini
ADKaster: Added RunningTests.md

Userland/Utilities/test-js.cpp
ADKaster: Looks like the shell/kernel weirdness is stable enough to run the tests consistently on my machine

Documentation/RunningTests.md
```diff
@@ -0,0 +1,76 @@
+## Running SerenityOS Tests
+
+There are two classes of tests built during a Serenity build: host tests and target tests. Host tests run on the build
+machine, and use [Lagom](https://github.com/SerenityOS/serenity/blob/master/Meta/Lagom/README.md) to build Serenity
```
linusg, line 4: Let's link relative to this, so it works offline as well:

```suggestion
machine, and use [Lagom](../Meta/Lagom/README.md) to build Serenity
```

Documentation/RunningTests.md
```diff
@@ -0,0 +1,76 @@
...
+### Running Target Tests
+
+Tests built for the serenity target get installed either into `/usr/Tests` or `/bin`. `/usr/Tests` is preferred, but
+some system tests are installed into `/bin` for historical reasons.
+
+The easiest way to run all of the known tests in the system is to use the `run-tests-and-shutdown.sh` script that gets
+installed into `/home/anon/tests`. When running in CI, the environment variable `$DO_SHUTDOWN_AFTER_TESTS` is set, which
+will run `shutdown -n` after running all the tests.
+
+For completeness, a basic on-target test run will need the serenity image built and run via Qemu.
```
linusg, line 43: nitpick:

```suggestion
For completeness, a basic on-target test run will need the serenity image built and run via QEMU.
```

linusg (commented):
nice job documenting running the tests!

Documentation/RunningTests.md
ADKaster: sounds good. I think at least one of the other documentation pages has github links in it, so that's probably worth a docs fixup MR. Or at least, they did when I started this page :)

Documentation/RunningTests.md
ADKaster: Caps applied throughout to Serenity and QEMU

Meta/notify_irc.py
```diff
@@ -96,7 +96,8 @@ def run_on(json_string):
     wrapper = json.loads(json_string)
     line = compute_lines(wrapper)
     if line:
-        send_notification(line)
+        # send_notification(line)
+        print("About to send {}".format(line))
```
awesomekling, line 100: Hmm?

awesomekling: Ping. This reads like it would break the IRC notifications.

ADKaster:
Whoops yeah forgot to remove that when testing the notifications. I can
push a fix when I get off work today 😅

Meta/notify_irc.py
ADKaster: reverted, sorry bout the delay!

.github/workflows/cmake.yml
```diff
@@ -86,33 +69,68 @@ jobs:
...
+    - name: Print target logs
+      if: ${{ !cancelled() && matrix.debug-macros == 'NORMAL_DEBUG'}}
+      working-directory: ${{ github.workspace }}/Build
+      run: cat ./debug.log
+
+
+  notify_irc:
+    needs: [build_and_test_serenity, build_and_test_lagom]
+    runs-on: ubuntu-20.04
+    if: always() && github.repository == 'SerenityOS/serenity' && (github.event_name == 'pull_request' || (github.event_name == 'push' && github.ref == 'refs/heads/master'))
```
BenWiederhake, line 130: `always() && ` is redundant.

Have you replaced `github.repository == 'SerenityOS/serenity'` with your own repository name (and your own separate channel in `notify_irc.py`) to test whether this works? I never used the `needs:` feature before, and you reported that "not working as [you] hoped". Well, what does it do? What did your elf eyes see? :P

.github/workflows/cmake.yml
```diff
@@ -123,25 +141,51 @@ jobs:
         cat <<"EOF"
         ${{ toJSON(github.event) }}
         EOF
-    - name: Generate IRC message
-      if: matrix.should-notify-irc == true && !cancelled()
+    - name: Generate IRC message success
+      if: ${{ !cancelled() && needs.build_and_test_serenity.result && needs.build_and_test_lagom.result}}
+      run: |
+        ${{ github.workspace }}/Meta/notify_irc.py <<"EOF"
+        ["${{ github.actor }}", ${{ github.run_id }}, "success",
```
BenWiederhake, line 148: Suggestion: Replace `"success"` by `["${{needs.build_and_test_serenity.result}}", "${{needs.build_and_test_lagom.result}}"]`.

And now that I think about it: How does this interact with the build matrix? After all, there are three different jobs all named `build_and_test_lagom`. Maybe add a `${{toJSON(needs)}}` above in "Dump event info", just to make sure no information is lost.

Also, my gut feeling says that this job needs to be put below the jobs it references.

Base/home/anon/tests/run-tests-and-shutdown.sh
```diff
@@ -0,0 +1,34 @@
...
+}
+
+# TODO: It'd be nice to have a list+list op (as opposed to nest-on-in-another)
+system_tests=(test-js (test-crypto bigint -t)) #("test-js" "test-web" "test-pthread" "test-crypto" "test-compress" $usr_tests_tests)
+all_tests=($system_tests $(find /usr/Tests -type f | grep -v Kernel | grep -v .inc | shuf))
+
+count=0
+for list in $all_tests {
+    for $list { count=$(expr $count + 1) }
+}
```
BenWiederhake, line 23: Where is this count being used?

Base/home/anon/tests/run-tests-and-shutdown.sh
```diff
@@ -0,0 +1,34 @@
...
+        echo "::info file=$test_cmd:: $test_cmd passed!"
+    } else {
+        echo "::error file=$test_cmd:: $test_cmd returned non-zero exit code, check logs!"
+    }
+    idx=$(expr $idx + 1)
+}
+
+# TODO: It'd be nice to have a list+list op (as opposed to nest-on-in-another)
+system_tests=(test-js (test-crypto bigint -t)) #("test-js" "test-web" "test-pthread" "test-crypto" "test-compress" $usr_tests_tests)
+all_tests=($system_tests $(find /usr/Tests -type f | grep -v Kernel | grep -v .inc | shuf))
```
BenWiederhake, line 18: Shuffling them will reveal more bugs, at the cost of being non-deterministic.

I agree with you that we should reveal more bugs, but in order to track them down, the order needs to be recoverable. Maybe print the names of the tests as they are being executed, otherwise we will end up with randomly failing tests.

BenWiederhake, line 18: This runs *all* current tests, even though most of them are currently broken. How should we deal with this? I think it's a bad idea to just make all builds fail until every single "test" in this directory is repaired; this will take a while.

BenWiederhake (changes requested):
I love where this is going <3

.github/workflows/cmake.yml
BenWiederhake: Correction: [`always() &&` is apparently not redundant](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-not-requiring-dependent-jobs-to-be-successful)

Userland/Shell/Shell.cpp
```diff
@@ -1658,18 +1644,6 @@ void Shell::notify_child_event()
                     job.set_is_suspended(true);
                 }
                 found_child = true;
-#ifdef ENSURE_WAITID_ONCE
-                // NOTE: This check is here to find bugs about our assumptions about waitpid(),
-                //       it does not hold in general, and it definitely does not hold in the long run.
-                // Reasons that we would call waitpid() more than once:
-                // - PID reuse/wraparound: This will simply fail the assertion, ignored here.
-                // - Non-terminating unblocks:
-                //   - Suspension: (e.g. via ^Z)
```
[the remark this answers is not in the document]
alimpfard: "this" refers to the assertion itself :P

Base/home/anon/tests/run-tests-and-shutdown.sh
alimpfard: The names _are_ being printed, take a look at the `run()` function.

BenWiederhake: D'oh! Yes, you're right.

ADKaster: Well, fun fact, printing `::error` doesn't actually fail the build! (who knew.) Last I checked test-js doesn't pass and a few of the others abort instead of forking to check that a particular assertion happens. Seems a #TODO  to make the failing tests fail instead of printing unhappy things at the bottom of the actions run.

Base/home/anon/tests/run-tests-and-shutdown.sh
ADKaster: It's used to print "Running test $name ($idx/ $count )" in the run() function above. Gotta love globals that come after the function they're used in right? Anyway the entire reason this loop is here is b/c there's no easy way to get a list's length without .. whatever this is. spawning a subshell to increment a value. Takes forever without KVM on :)

.github/workflows/cmake.yml
ADKaster: I have not properly tested the IRC notifications with a real IRC server, you're right. It seems that for the matrix jobs, whichever one completes last is the one that gets the build status updated here when we check this job.

And yes, always is not redundant :/ . If there's a better way to say "please run this job always and give me a bool that says 'ALL previous jobs [succeeded|failed]'" using github actions I couldn't find one...

.github/workflows/cmake.yml
```diff
@@ -86,33 +69,68 @@ jobs:
...
-      working-directory: ${{ github.workspace }}/Build/Meta/Lagom
-      run: ./test-compress
+      run: |
+        echo -e "\n[TestRunner@ttyS0]\n" >> ../Base/etc/SystemServer.ini
+        echo -e "Executable=/home/anon/tests/run-tests-and-shutdown.sh\n" >> ../Base/etc/SystemServer.ini
+        echo -e "StdIO=/dev/ttyS0\n" >> ../Base/etc/SystemServer.ini
+        echo -e "Environment=DO_SHUTDOWN_AFTER_TESTS=1 TERM=xterm PATH=/bin:/usr/bin:/usr/local/bin\n" >> ../Base/etc/SystemServer.ini
+        echo -e "User=anon\n" >> ../Base/etc/SystemServer.ini
+        echo -e "WorkingDirectory=/home/anon\n" >> ../Base/etc/SystemServer.ini
+        echo -e "BootModes=text\n" >> ../Base/etc/SystemServer.ini
```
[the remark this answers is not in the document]
ADKaster: I'm not sure what you mean with `-n` ? I'm using `echo -e` which will interpret escape sequences. Testing locally though you're right about the double newline. Not sure which stack overflow post-fu failed me here.

.github/workflows/cmake.yml
ADKaster: From what I understand about how yaml parsers work, it doesn't "need" to be below, but It should be moved, yes. reads better that way.

I found this action that should do the trick? https://github.com/marketplace/actions/workflow-conclusion-action 

Assuming @awesomekling  would be ok adding a third party action.

Userland/Libraries/LibRegex/RegexDebug.h
```diff
@@ -129,8 +129,10 @@ class RegexDebug {
         for (size_t i = 0; i < length; ++i) {
             builder.append('=');
         }
+        auto str = builder.to_string();
+        VERIFY(!str.is_empty());
 
-        fprintf(m_file, "%s\n", builder.to_string().characters());
+        fprintf(m_file, "%s\n", str.characters());
```
ADKaster, line 135: Makes absolutely no sense to me how this could possibly be null, so I added a VERIFY. .... 

```
In file included from ../Userland/Libraries/LibRegex/Regex.h:30,
                 from ../Userland/Libraries/LibRegex/Tests/Regex.cpp:30:
../Userland/Libraries/LibRegex/RegexDebug.h: In member function 'void regex::RegexDebug::print_header()':
../Userland/Libraries/LibRegex/RegexDebug.h:133:16: error: '%s' directive argument is null [-Werror=format-overflow=]
  133 |         fprintf(m_file, "%s\n", builder.to_string().characters());
      |         ~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

alimpfard: It can't see the definition of `StringBuilder::to_string()`, and `String::characters()` _may_ return `nullptr` (even though `StringBuilder::to_string()` will never return that).

ADKaster: Gotcha. I'll keep the VERIFY then, since the compiler can't assume what to_string does and we can't un-confuse it without a check.

.github/workflows/cmake.yml
ADKaster: ayy looks like that action does the right thing. It even prints out all the good goods in some expandable tabs (jobs, conclusions, conclusion).

Failed run (one job failed): https://github.com/SerenityOS/serenity/runs/1992038457?check_suite_focus=true#step:3:9
Success run (all jobs passed): https://github.com/SerenityOS/serenity/runs/1992047005?check_suite_focus=true#step:3:9

Base/home/anon/tests/run-tests-and-shutdown.sh
BenWiederhake: Sounds like there should be shell built-ins "enumerate" (like in python) and "length" (of a list), CC @alimpfard :P

BenWiederhake:
There's many outdated/resolved discussions that only the author of the PR can close. Could you close them so the unresolved ones stand out more? Also, thanks for the request, I'll get right to it! :)

Userland/Utilities/shuf.cpp
```diff
@@ -54,6 +53,7 @@ int main([[maybe_unused]] int argc, [[maybe_unused]] char** argv)
             break;
         lines.append({ buffer, AK::ShouldChomp::Chomp });
     }
+    free(buffer);
```
BenWiederhake, line 56: Mea culpa! D:

BenWiederhake (changes requested):
The main problem with running the tests on Qemu is that Github Actions thinks it's "hanging" because there is no output. Is there a way to tell qemu to print to stderr instead of debug.log? This way, github actions should understand the output as, well, output, and cancel it only after a much longer time.

Furthermore, GA not running the tests in Serenity makes it a bit difficult to judge a PR about running the tests on Serenity :P

Meta/run.sh
```diff
@@ -126,6 +126,22 @@ elif [ "$SERENITY_RUN" = "qcmd" ]; then
...
+        $SERENITY_EXTRA_QEMU_ARGS \
+        -s -m $SERENITY_RAM_SIZE \
+        -cpu $SERENITY_QEMU_CPU \
+        -d guest_errors \
+        -smp 2 \
+        -drive file=${SERENITY_DISK_IMAGE},format=raw,index=0,media=disk \
+        -device ich9-ahci \
+        -nographic \
+        -display none \
+        -debugcon file:debug.log \
```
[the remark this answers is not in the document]
alimpfard: Should be the same place as `e1000.pcap`, in `Build/`

Base/home/anon/tests/run-tests-and-shutdown.sh
BenWiederhake: > Takes forever without KVM on :)

It doesn't fix the underlying problem, but it's a good idea anyway: Can we switch KVM on? I do realize it probably won't work (the GA image is, itself, already a VM, so running a VM inside a VM probably won't have hardware-acceleration available), but is there a way to try?

alimpfard: Yes, I am to blame for this.
it's not that the shell can't tell you the length of a list, it is that I cannot find the right syntax to add to the shell language.
I hate the sh `${#var}` syntax, (though I've had ideas about parsing this stuff to the existing AST)

I've experimented with `${length $var}` before (there's an open draft PR with that), but I didn't quite like that either.

re `enumerate`, not a _bad_ idea, but again, syntax how?

Base/home/anon/tests/run-tests-and-shutdown.sh
alimpfard: there's a `test_commons.sh.inc` or somesuch in Shell/Tests, it's just a pass/fail function (for now).

ADKaster: > Why exclude the Kernel tests?

At least two of them completely hang -- the futex test and the condition variable test.

> UE fails

Yeah not sure about that one.

There's also whole issue on running tests better in a consistent way here:  #4078 . Hopefully with a reasonable way to run tests we can go back and make a reasonable way to write them.

I also noticed that AK/TestSuite.h doesn't actually return anything from main? It just falls off the end and lets the compiler put the `return 0;`. I was planning to do a subsequent PR to have all the EXPECT and VERIFY macros in that header record to a TestCase::did_fail() thing.

>  I suggest making a blacklist of tests that cannot be run automatically yet.

I mean, this PR has been open a while... Can we close that item to an issue? 😅 

> I'm not sure what you mean by "Seems a #TODO to make the failing tests fail …", could you elaborate?

When a test fails, it will print an `::error` that gets read by actions and is visible in the actions overview. However, it doesn't fail the job. I'm not sure how to make github actions fail the job using that syntax.

Base/home/anon/tests/run-tests-and-shutdown.sh
ADKaster: > Can we switch KVM on?

Sadly no, because Github actions does not support nested virtualization.

Meta/run.sh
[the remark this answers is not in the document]
ADKaster: Re your intro comment: It's hanging in the actions because there legitimately is no output because the shell is hung in the kernel -- see #5541.  In a lucky run, the thing does complete all the way. And then dump the debug.log in the next step.

The way the qemu options work is whatever we write to /dev/ttyS0 will be sent to stdout, and the kernel log is sent to debug.log. So we see the test resuls scroll by, then then debug.log is cat-ed in the next !cancelled() step.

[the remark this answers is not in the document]
ADKaster: Ah and yes, it will drop debug.log in $PWD. Which will likely be Build/ so no gitignore needed imo.

.github/workflows/cmake.yml
```diff
@@ -86,62 +69,89 @@ jobs:
...
-        cat <<"EOF"
-        ${{ toJSON(github.event) }}
-        EOF
-    - name: Generate IRC message
-      if: matrix.should-notify-irc == true && !cancelled()
+    - name: Create Serenity Rootfs
+      if: ${{ matrix.debug-macros == 'NORMAL_DEBUG'}}
+      working-directory: ${{ github.workspace }}/Build
+      # See Documentation/RunningTests.md for copy-pasteable version of this addition
+      #    (Also please keep them in sync :) )
```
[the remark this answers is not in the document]
ADKaster: Well, boot mode is a global string in SystemServer, and all the services check whether their boot mode is contained w/in the boot_mode string before trying to start. So to do what you're suggesting we'd need to add "self-test" to enough services that the tests could actually run, but it would be possible yes. I'll look into that one. I never did like echoing a bunch of stuff into the file in the ci script.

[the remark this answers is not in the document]
ADKaster: All right I pushed a "self-test" mode that's an alias for text mode. Bit gnarly that setting up the framebuffer for graphical is handled in the kernel still but whatevs, it works as written.

Base/home/anon/tests/run-tests-and-shutdown.sh
BenWiederhake: > I mean, this PR has been open a while... Can we close that item to an issue?

I get what you mean – that's why I suggest to move the problem of failing tests to a different time. For example, blacklisting everything that fails, or only running known-working tests. If we only run two or three tests, manually invoked, that would also circumvent the Shell problems for a while, the PR can get accepted, and there's the opportunity to develop fixes independently. :)

Base/home/anon/tests/run-tests-and-shutdown.sh
ADKaster: All right, i've removed the count and left a todo in its place :)

Base/home/anon/tests/run-tests-and-shutdown.sh
ADKaster: That's a good point... would be nice to have things be loud and visible when they don't work properly, but that should be *after* we get to a known good state. I've pared down the set of tests to test-js, test-pthread, test-compress, and test-crypto bigint. These seem to somewhat reliably complete on my machine in non-KVM mode. Still prefer to leave the continue on error  in the run step until #5541 is resolved though.

BenWiederhake:
:partying_face: LGTM

Oh boy we should really figure out what's making Qemu spit out all those "Invalid access at addr 0xFEBB0000, size 1, region '(null)', reason: rejected" lines.

This hides failures like, oh, [test-js](https://github.com/SerenityOS/serenity/pull/4921/checks?check_run_id=1995437739#step:18:536) is already failing ^^'

PR #4927, opened 2021-01-12
Docs: Add design doc and implementation plan for highdpi

nico:
I have a local branch that gets us past implementation stage 1
in this doc, and it seems like a useful enough checkpoint to
upstream it and then iterate in tree.

### DIFF
--- a/Documentation/HighDPI.md
+++ b/Documentation/HighDPI.md
@@ -1,0 +1,37 @@
+HighDPI design
+==============
+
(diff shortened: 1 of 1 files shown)


nico:
@tomuta FYI. Should be fairly uncontroversial so far. Probably more interesting to talk about the actual code, once it's there.

tomuta:
Looks good to me

PR #4931, opened 2021-01-13
Everywhere: Continue replacing dbg() dbgln(). (4)

asynts:
Follow up to #4907.

I've started touching some `dbg()` statements which are in `#ifdef` blocks. (Only in the last commit.)

Before:
~~~none
$ grep -rnI 'dbg()' | wc -l
641
~~~

After:
~~~none
$ grep -rnI 'dbg()' | wc -l
565
~~~

### DIFF
--- a/Kernel/Scheduler.cpp
+++ b/Kernel/Scheduler.cpp
@@ -24,6 +24,7 @@
  * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
+#include <AK/Debug.h>
 #include <AK/QuickSort.h>
 #include <AK/ScopeGuard.h>
 #include <AK/TemporaryChange.h>
@@ -131,28 +132,46 @@ bool Scheduler::pick_next()
         // no longer want to schedule this thread. We can't wait until
         // Scheduler::enter_current because we don't want to allow it to
         // transition back to user mode.
-#ifdef SCHEDULER_DEBUG
-        dbg() << "Scheduler[" << Processor::current().id() << "]: Thread " << *current_thread << " is dying";
-#endif
+
+        if constexpr (debug_scheduler)
+            dbgln("Scheduler[{}]: Thread {} is dying", Processor::current().id(), *current_thread);
+
         current_thread->set_state(Thread::Dying);
     }
 
-#ifdef SCHEDULER_RUNNABLE_DEBUG
-    dbg() << "Scheduler[" << Processor::current().id() << "]: Non-runnables:";
-    Scheduler::for_each_nonrunnable([&](Thread& thread) -> IterationDecision {
-        if (thread.state() == Thread::Dying)
-            dbg() << "  " << String::format("%-12s", thread.state_string()) << " " << thread << " @ " << String::formatted("{:04x}:{:08x}", thread.tss().cs, thread.tss().eip) << " Finalizable: " << thread.is_finalizable();
-        else
-            dbg() << "  " << String::format("%-12s", thread.state_string()) << " " << thread << " @ " << String::formatted("{:04x}:{:08x}", thread.tss().cs, thread.tss().eip);
-        return IterationDecision::Continue;
-    });
+    if constexpr (debug_scheduler_runnable) {
+        dbgln("Scheduler[{}j]: Non-runnables:", Processor::current().id());
+        Scheduler::for_each_nonrunnable([&](Thread& thread) -> IterationDecision {
+            if (thread.state() == Thread::Dying) {
+                dbgln("  {:12} {} @ {:04x}:{:08x} Finalizable: {}",
+                    thread.state_string(),
+                    thread,
+                    thread.tss().cs,
+                    thread.tss().eip,
+                    thread.is_finalizable());
+            } else {
+                dbgln("  {:12} {} @ {:04x}:{:08x}",
+                    thread.state_string(),
+                    thread,
+                    thread.tss().cs,
+                    thread.tss().eip);
+            }
 
-    dbg() << "Scheduler[" << Processor::current().id() << "]: Runnables:";
-    Scheduler::for_each_runnable([](Thread& thread) -> IterationDecision {
-        dbg() << "  " << String::format("%3u", thread.effective_priority()) << "/" << String::format("%2u", thread.priority()) << " " << String::format("%-12s", thread.state_string()) << " " << thread << " @ " << String::formatted("{:04x}:{:08x}", thread.tss().cs, thread.tss().eip);
-        return IterationDecision::Continue;
-    });
-#endif
+            return IterationDecision::Continue;
+        });
+
+        dbgln("Scheduler[{}j]: Runnables:", Processor::current().id());
+        Scheduler::for_each_runnable([](Thread& thread) -> IterationDecision {
+            dbgln("  {:3}/{:2} {:12} @ {:04x}:{:08x}",
+                thread.effective_priority(),
+                thread.priority(),
+                thread.state_string(),
+                thread.tss().cs,
+                thread.tss().eip);
+
+            return IterationDecision::Continue;
+        });
+    }
 
     Thread* thread_to_schedule = nullptr;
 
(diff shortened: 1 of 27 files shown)


asynts:
Not sure what this build error is about. Seems like `<fcntl.h>` was not included? I didn't change anything in that regard.

~~~none
error: 'F_SETFD' was not declared in this scope
~~~

Kernel/Scheduler.cpp
```diff
@@ -211,9 +227,11 @@ bool Scheduler::pick_next()
...
         thread_to_schedule = Processor::current().idle_thread();
 
-#ifdef SCHEDULER_DEBUG
-    dbg() << "Scheduler[" << Processor::current().id() << "]: Switch to " << *thread_to_schedule << " @ " << String::format("%04x:%08x", thread_to_schedule->tss().cs, thread_to_schedule->tss().eip);
-#endif
+    if constexpr (debug_scheduler)
+        dbgln("Scheduler[{}]: Switch to {} @ {:04x}:{:08x}",
+            Processor::current().id(),
+            *thread_to_schedule,
+            thread_to_schedule->tss().cs, thread_to_schedule->tss().eip);
```
emanuele6, line 234: This is a multi-line statement, it should be wrapped in curlies.

But, again, any reason not to use `dbgln<debug_scheduler>()`?

Kernel/Scheduler.cpp
```diff
@@ -248,9 +264,9 @@ bool Scheduler::yield()
 
     if (!Scheduler::pick_next())
         return false;
-#ifdef SCHEDULER_DEBUG
-    dbg() << "Scheduler[" << Processor::current().id() << "]: yield returns to thread " << *current_thread << " in_irq: " << Processor::current().in_irq();
-#endif
+
+    if constexpr (debug_scheduler)
+        dbgln("Scheduler[{}]: yield returns to thread {} in_irq={}", Processor::current().id(), *current_thread, Processor::current().in_irq());
```
emanuele6, line 269: any reason not to use `dbgln<debug_scheduler>()`?

Kernel/Thread.cpp
```diff
@@ -61,9 +62,8 @@ Thread::Thread(NonnullRefPtr<Process> process)
     } else {
         m_tid = Process::allocate_pid().value();
     }
-#ifdef THREAD_DEBUG
-    dbg() << "Created new thread " << m_process->name() << "(" << m_process->pid().value() << ":" << m_tid.value() << ")";
-#endif
+    if constexpr (debug_thread)
+        dbgln("Created new thread {}({}:{})", m_process->name(), m_process->pid().value(), m_tid.value());
```
emanuele6, line 66: any reason not to use `dbgln<debug_thread>()`?

Kernel/Scheduler.cpp
asynts: `Processor::current().id()` is non trivial therefor I am wrapping it in `if constexpr`. I'll fix the braces.

asynts:
I updated the braces:

~~~diff
diff --git a/Kernel/Scheduler.cpp b/Kernel/Scheduler.cpp
index 7e07a30ab..44f65f06d 100644
--- a/Kernel/Scheduler.cpp
+++ b/Kernel/Scheduler.cpp
@@ -142,19 +142,20 @@ bool Scheduler::pick_next()
     if constexpr (debug_scheduler_runnable) {
         dbgln("Scheduler[{}j]: Non-runnables:", Processor::current().id());
         Scheduler::for_each_nonrunnable([&](Thread& thread) -> IterationDecision {
-            if (thread.state() == Thread::Dying)
+            if (thread.state() == Thread::Dying) {
                 dbgln("  {:12} {} @ {:04x}:{:08x} Finalizable: {}",
                     thread.state_string(),
                     thread,
                     thread.tss().cs,
                     thread.tss().eip,
                     thread.is_finalizable());
-            else
+            } else {
                 dbgln("  {:12} {} @ {:04x}:{:08x}",
                     thread.state_string(),
                     thread,
                     thread.tss().cs,
                     thread.tss().eip);
+            }
 
         });
@@ -227,11 +228,12 @@ bool Scheduler::pick_next()
     if (!thread_to_schedule)
         thread_to_schedule = Processor::current().idle_thread();
 
-    if constexpr (debug_scheduler)
+    if constexpr (debug_scheduler) {
         dbgln("Scheduler[{}]: Switch to {} @ {:04x}:{:08x}",
             Processor::current().id(),
             *thread_to_schedule,
             thread_to_schedule->tss().cs, thread_to_schedule->tss().eip);
+    }
 
     // We need to leave our first critical section before switching context,
     // but since we're still holding the scheduler lock we're still in a critical section
~~~

asynts:
I rebased onto master because I suspect #4936 fixed the build issue.

awesomekling:
Hmm, do we actually need the separate Debug.h header and the lowercase names?
Does `if constexpr (FOO_DEBUG)` not work?

asynts:
Personally, I just really hate macros, and would love to see them go entirely.

The biggest advantage and biggest disadvantage of using macros is the same: they can be defined in some compilation units only. This means that a `FOO_DEBUG` could be defined in `foo.cpp` but not in `bar.cpp` and thus only the debug statements in `foo.cpp` would apply. This could also decrease build time as only some compilation units would have to be recompiled.

On the other hand, this might not be the behavior that you expect and the same granularity could be achieved by defining multiple `debug_foo_x` macros. If we want to get really fancy we could use `if constexpr (debug_foo || debug_foo_x)` to get the same granularity and if we then define `debug_foo_x` in that compilation unit we can decrease build times as well.

I would have to change the way the macros are defined to allow them to be used in if statements, but that's not much of an issue:

https://github.com/SerenityOS/serenity/blob/30d515bb61d9458e3dbf80433c388443f8cd82d1/Meta/CMake/all_the_debug_macros.cmake#L1-L6

Another advantage of using macros that I can think of is that it is possible to use them in all contexts. There was one weird case where `LOCK_DEBUG` was used to conditionally define a member variable:

https://github.com/SerenityOS/serenity/blob/30d515bb61d9458e3dbf80433c388443f8cd82d1/Kernel/Thread.h#L1167-L1177

That's not really a big issue because we still need the `FOO_DEBUG` macros to communicate from CMake to C++ and can simply use `#ifdef` for this one case.

That being said, I still prefer the constexpr way. Should I change it to use macros or leave it as is?

asynts:
I've been thinking more about this.

  - Since debug macros would always have to be defined as either `true` or `false`, redefining them would require an `#undef`.

  - `Debug.h` can very easily be generated with CMake's `configure_file` and `#cmakedefine01`. ([reference][1])

    This would also provide a nice way of turning debug options on and of with `cmake .. -DFOO_DEBUG`.

  [1]: https://cmake.org/cmake/help/latest/command/configure_file.html

awesomekling:
> I've been thinking more about this:
> 
> * Since debug macros would always have to be defined as either `true` or `false`, redefining them would require an `#undef`. That's a bit ugly in my opinion.
> * `Debug.h` can very easily be generated with CMake's `configure_file` and `#cmakedefine01`. ([reference](https://cmake.org/cmake/help/latest/command/configure_file.html))
>   This would also provide a nice way of turning debug options on and of with `cmake .. -DFOO_DEBUG`.
> * We won't really need `ENABLE_ALL_THE_DEBUG_MACROS` anymore because this was only really used to ensure that that code was build which will be the case when all the `#ifdef` stuff is gone.

Yeah all righty. That last point is a great :)

asynts:
Still unsure about this. If we do use macros, we would create a `Debug.h.in` file like this:

~~~c++
#cmakedefine01 FOO_DEBUG
~~~

which would be replaced by one of the following:

~~~c++
#define FOO_DEBUG 0
#define FOO_DEBUG 1
~~~

then we could do `dbgln<FOO_DEBUG>(...)` and `#ifdef FOO_DEBUG` in the code base.

---

If we used the `constexpr` version, we would create a `Debug.h.in` file like this:

~~~c++
~~~

awesomekling:
> But replacing all the `#ifdef FOO_DEBUG` with `#if FOO_DEBUG == 1` will be another project to undertake. How about, we call this a "hack" and I'll come back to this?

I'm down with that.

PR #4935, opened 2021-01-13
DHCPClient: handle /proc/net/adapters invalid JSON gracefully

bcoles:
`DHCPClient` exits gracefully if `/proc/net/adapters` is unreadable. We don't need to `assert` a few lines later if `/proc/net/adapters` is empty or malformed. We can easily validate the JSON and return an error message instead.

Given that coredumps are expensive it makes sense to handle errors gracefully rather than delaying system boot by crashing (three times).

### DIFF
--- a/Userland/Services/DHCPClient/main.cpp
+++ b/Userland/Services/DHCPClient/main.cpp
@@ -69,13 +69,18 @@ int main([[maybe_unused]] int argc, [[maybe_unused]] char** argv)
 
     auto file = Core::File::construct("/proc/net/adapters");
     if (!file->open(Core::IODevice::ReadOnly)) {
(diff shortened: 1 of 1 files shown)


Userland/Services/DHCPClient/main.cpp
```diff
@@ -75,7 +75,12 @@ int main([[maybe_unused]] int argc, [[maybe_unused]] char** argv)
 
     auto file_contents = file->read_all();
     auto json = JsonValue::from_string(file_contents);
-    ASSERT(json.has_value());
+
+    if (!json.has_value() || !json.value().is_object()) {
+        fprintf(stderr, "Error: No network adapters available\n");
```
linusg, line 80:
```suggestion
        warnln("Error: No network adapters available");
```

Issue #4830, opened 2021-01-06
Kernel: Failed assertion when listing `/proc/<pid>/fd`


alimpfard:
Repro:
1. Attempting tab-completion of `/proc/<pid>/fd/` 
```
[Shell(31:31)]: ASSERTION FAILED: read_callback
../Kernel/FileSystem/ProcFS.cpp:1137 in Kernel::KResult Kernel::ProcFSInode::refresh_data(Kernel::FileDescription&) const
[Shell(31:31)]: 0xc01196a6  __assertion_failed(char const*, char const*, unsigned int, char const*) +0xe2
[Shell(31:31)]: 0xc014c874  Kernel::ProcFSInode::refresh_data(Kernel::FileDescription&) const +0xe4
[Shell(31:31)]: 0xc014c9ff  Kernel::ProcFSInode::attach(Kernel::FileDescription&) +0x31
[Shell(31:31)]: 0xc013f787  Kernel::FileDescription::attach() +0x5f
[Shell(31:31)]: 0xc0140fc8  Kernel::FileDescription::create(Kernel::Custody&) +0x2ac
[Shell(31:31)]: 0xc015aa75  Kernel::VFS::open(AK::StringView, int, unsigned short, Kernel::Custody&, AK::Optional<Kernel::UidAndGid>) +0xacb
[Shell(31:31)]: 0xc0198bf0  Kernel::Process::sys$open(AK::Userspace<Kernel::Syscall::SC_open_params const*>) +0x580
[Shell(31:31)]: 0xc01881be  syscall_handler +0x6be
[Shell(31:31)]: 0xc0187af8  syscall_asm_entry +0x31
```
2. `ls /proc/<pid>/fd`
```
[ls(27:27)]: ASSERTION FAILED: read_callback
../Kernel/FileSystem/ProcFS.cpp:1137 in Kernel::KResult Kernel::ProcFSInode::refresh_data(Kernel::FileDescription&) const
[ls(27:27)]: 0xc01196a6  __assertion_failed(char const*, char const*, unsigned int, char const*) +0xe2
[ls(27:27)]: 0xc014c874  Kernel::ProcFSInode::refresh_data(Kernel::FileDescription&) const +0xe4
[ls(27:27)]: 0xc014c9ff  Kernel::ProcFSInode::attach(Kernel::FileDescription&) +0x31
[ls(27:27)]: 0xc013f787  Kernel::FileDescription::attach() +0x5f
[...]

PR #4941, opened 2021-01-13
ProcFS: Ignore directories in refresh_data().

Maato:
Fixes: https://github.com/SerenityOS/serenity/issues/4830


bcoles:
This resolved the issue for me with both listing `/proc/<pid>/fd*` and expanding `/proc/` tree in File Manager.

There appears to be an issue with `/proc/self`; however, this issue existed prior to this PR.

/proc/self

PR #4943, opened 2021-01-13
Documentation: Recommend ninja by default

BenWiederhake:
I can only verify that this is the correct package name in Debian, but not for the other OSes. We should probably add at least the package names for MacOS and Ubuntu. Note that the `<!-- … -->` constructs will be detected by GitHub as HTML comments and thus not displayed.


nico:
for macos and windows i'd link to the releases on https://github.com/ninja-build/ninja/releases

linusg:
> I can only verify that this is the correct package name in Debian, but not for the other OSes

Helpful page: https://pkgs.org/download/ninja-build

TL;DR: Package name is `ninja-build` on Ubuntu, Debian, Fedora, CentOS.

Documentation/BuildInstructions.md
```diff
@@ -3,33 +3,41 @@
 ### Prerequisites
 
 #### Linux prerequisites
-Make sure you have all the dependencies installed:
+Make sure you have all the dependencies installed (`ninja` is optional, but is faster in practice):
 
 **Debian / Ubuntu**
 ```bash
-sudo apt install build-essential cmake curl libmpfr-dev libmpc-dev libgmp-dev e2fsprogs qemu-system-i386 qemu-utils
+sudo apt install build-essential cmake curl libmpfr-dev libmpc-dev libgmp-dev e2fsprogs ninja qemu-system-i386 qemu-utils
```
linusg, line 10: This is wrong! See:

https://github.com/SerenityOS/serenity/blob/13d7c09125f8eec703d0a43a9a87fc8aa08f7319/.github/workflows/cmake.yml#L155

and

https://packages.ubuntu.com/search?keywords=ninja

AndreasTNT:
Arch 
Sudo pacman -S ninja

Documentation/BuildInstructions.md
```diff
@@ -3,33 +3,41 @@
...
 **openSUSE**
 ```bash
 sudo zypper install curl cmake mpfr-devel mpc-devel gmp-devel e2fsprogs patch qemu-x86 qemu-audio-pa gcc gcc-c++ patterns-devel-C-C++-devel_C_C++
 ```
 
+<!-- FIXME: Add ninja -->
+
 **Arch Linux / Manjaro**
 ```bash
 sudo pacman -S --needed base-devel cmake curl mpfr libmpc gmp e2fsprogs qemu qemu-arch-extra
```
asynts, line 29: For Archlinux:

```suggestion
sudo pacman -S --needed base-devel cmake curl mpfr libmpc gmp e2fsprogs qemu qemu-arch-extra ninja
```

Not sure what Manjaro is, but a quick search shows that it's based on Archlinux so this should work.

Documentation/BuildInstructions.md
```diff
@@ -3,33 +3,41 @@
...
 ```
```
asynts, line 30: For Archlinux: (Not sure what Manjaro is but it seems to be based on Archlinux, so this is probably okay.)

~~~suggestion
**Arch Linux / Manjaro**
```bash
sudo pacman -S --needed base-devel cmake curl mpfr libmpc gmp e2fsprogs qemu qemu-arch-extra ninja
```
~~~

asynts:
*Hmm. GitHub doesn't seem to be able to handle multi-line suggestions.*

For ArchLinux this should be

~~~none
sudo pacman -S --needed base-devel cmake curl mpfr libmpc gmp e2fsprogs qemu qemu-arch-extra ninja
~~~

BenWiederhake:
I added all your comments :+1:

Now missing:
- openSUSE
- ALT Linux
- OpenBSD
- FreeBSD

Uhh. No idea why the build suddenly fails?

linusg:
> Uhh. No idea why the build suddenly fails?

That would be: things we have no control over falling apart.

> E: Failed to fetch http://apt.llvm.org/focal/dists/llvm-toolchain-focal-11/main/binary-amd64/Packages  File has unexpected size (8775 != 8802). Mirror sync in progress? [IP: 151.101.250.49 443]
> E: Some index files failed to download. They have been ignored, or old ones used instead.

ADKaster:
For macOS our CI gets ninja from homebrew I think? Unless it's pre-installed on the runners.

AndreasTNT:
Copied from ninja-build/ninja wiki 

Pre built Ninja packages

Linux

Arch: pacman -S ninja

Debian/Ubuntu: apt-get install ninja-build

Fedora: dnf install ninja-build

Gentoo: emerge dev-util/ninja

Opensuse: zypper in ninja

Alpine: apk add ninja

Mac

HomeBrew: brew install ninja

MacPorts: port install ninja

FreeBSD

pkg install ninja

Documentation/BuildInstructions.md
```diff
@@ -74,46 +78,53 @@ Notes:
...
 ```bash
 $ cd Toolchain
 $ ./BuildIt.sh
 ```
 
-Building the toolchain will also automatically create a `Build/` directory for the build to live in, and build cmake inside that directory.
+Building the toolchain will also automatically create a `Build/` directory for the build to live in.
 
-Once the toolchain and cmake have been built, go into the `Build/` directory and run the `make` and `make install` commands:
+Once the toolchain has been built, go into the `Build/` directory and run the commands. Note that `ninja` seems to be faster, you can also just use GNU make, by omitting `-g Ninja` and calling `make` instead of `ninja`:
```
xspager, line 113: the g in `-g Ninja` should be uppercase, right?

BenWiederhake: It's already correct, isn't it?

predmond, line 113: this should be `-GNinja`

nico: Also, s/Note that/Note that while/ (?)

xspager: no, it's using lowercase ``g``. Forgive me if I'm wrong.

BenWiederhake: Ah, now I see! I was too focused on `Ninja` itself. Yes, I'll change that!

BenWiederhake: Thanks!

BenWiederhake:
Updated based on all your comments, and rebased for convenience :)

BenWiederhake:
Sneaky scope addition: I also made `serenity.includes` more easily copyable, without any change of content.

PR #4954, opened 2021-01-15
Applications: Basic implementation of Run

nvella:
Howdy all. I decided to have a go at a basic Run dialog ala Win9x/2k. This PR can be seen as the initial basic implementation, with myself or potentially others adding more features later on (eg. history combo, etc)

Currently I've 'implemented' launching URIs (file, http, etc.) by unashamedly yoinking the code from the `open` utility. This allows launching any program by providing it's full path, or any URI by their relevant URI handlers. However, the other half of the utility would be launching commands with parameters, not unlike how the dialog exists in Windows. I figure I'd need to plug into the shell somehow via spawning it with the provided command via the `-c` parameter, but I'm unsure on how I'd go about checking that the shell has successfully started the command. The hacky solution would be waiting a few hundred milliseconds then checking that the shell process hasn't exited with an error code, but I'd prefer to know if the command string is valid before sending it to the shell for execution. Any ideas?

Here's what the dialog looks like so far :^)

image

Todo:

- [x] Basic dialog design via GML
- [x] Icon
- [x] 'Browse...' button launching file picker dialog
- [x] Add to system menu
- [x] Run string via URI launch
- [x] Run string via command shell

'Wishlist'/for future PRs:
- Global system keyboard shortcut? (CTRL+ALT+R?) - there doesn't seem to be an easily identifiable handler in the system at the moment for this.
- Replace text input with combo, remember history
- Custom actions configuration ala XFCE (https://twitter.com/XfceNation/status/1349839448477990917)
- Path autocompletion


alimpfard:
> I'd prefer to know if the command string is valid before sending it to the shell for execution

Shell is a library, you can pull it in and `if (auto node = Shell::Parser(your_input).parse(); !node || node->is_syntax_error()) ...`
Naturally this will be hit-or-miss if the user's shell isn't `Shell`, you might be better off just letting the shell run it and maybe fail, then just inspecting the exit code

nvella:
> > I'd prefer to know if the command string is valid before sending it to the shell for execution
> 
> Shell is a library, you can pull it in and `if (auto node = Shell::Parser(your_input).parse(); !node || node->is_syntax_error()) ...`

Cheers for that 👍 That'll solve the problem of whether or not it's a valid command string which will allow me to error out to the user if their input doesn't make any sense. 

> Naturally this will be hit-or-miss if the user's shell isn't `Shell`, you might be better off just letting the shell run it and maybe fail, then just inspecting the exit code

The problem is more or less that I'd have to wait for the process to return to inspect its exit code. If it actually is a valid command, Run will be waiting for the process to quit before closing itself. A workaround could be just to hide the Run window, and if the Shell returns a 'not found' error, display the error dialog before reshowing Run, otherwise just quit from there after a successful execution.

nvella:
Doing some digging, it seems that running a command with `Shell -c` always returns zero (Shell `main.cpp` line 180.) If I were to go down the route of inspecting the exit code I'd have to pass-through the return code of shell->run_command; this does bring us in-line with the behaviour of Bash. However, I do notice that we exit-code 126 on an unknown command, where-as Bash, and presumably other shells, exit 127. Is this the right exit-code for us to be emitting, or should I check for both 126 and 127 to ensure compatibility?

alimpfard:
I don't remember there being a standard on exit codes, but iirc 126 is the general "cannot exec" code in bash, and 127 is the "not found" one.
If we can figure out whether a command is "not found", there's no harm in exiting with 127, so feel free to change that if it helps.

nvella:
Ready for review 👍 In regards to the command execution problem as stated above, it currently works as follows:
- Tries to launch the user input as a URI, closes and exits if successful.
- Tries to run the user input as a command via `/bin/Shell -c`. Hides and waits for the process to exit, checks the exit code, closes and exits if successful.
- Failing these, it displays an error dialog containing a generic 'check input and try again' failure message.

diversys:
Nice! Does folder path autocompletion work there?

nvella:
> Nice! Does folder path autocompletion work there?

Not at the moment - this PR is essentially minimum-viable-product - but that does sound potentially quite useful. A command history drop-down box would be the first thing I'd consider for improving this further.

alimpfard:
fwiw, `TextEditor` has a suggestions API you can use (on that note, maybe the suggestion trigger should be configurable?), and I think `TextBox` (which is a `TextEditor`) has a history thing.

nvella:
> fwiw, `TextEditor` has a suggestions API you can use (on that note, maybe the suggestion trigger should be configurable?), and I think `TextBox` (which is a `TextEditor`) has a history thing.

That does look interesting, my original plan was to replace the TextBox with an editable ComboBox and populate it with the Run history, however using the TextEditor autocompletion may be another possible route. I feel the combo option would be the simplest to implement, and keeps us most true to a Windows-style Run, however I'd be open to the idea of possibly doing cool things with TextEditor suggestions.

By the way, I've just rebased the PR onto master and pledged `sendfd` so we should be working with the anonymous file changes :)

PR #4958, opened 2021-01-15
WindowServer: Make HighDPI aware

nico:
Almost all logic stays in "logical" (unscaled coordinates), which
means the patch is small and things like DnD, window moving and
resizing, menu handling, menuapplets, etc all work without changes.

Screen knows about phyiscal coordinates and mouse handling internally is
in physical coordinates (so that two 1 pixel movements in succession can
translate to one 1 logical coordinate mouse movement -- only a single
event is sent in this case, on the 2nd moved pixel).

Compositor also knows about physical pixels for its backbuffers. This is
a temporary state -- in a follow-up, I'll try to let Bitmaps know about
their intrinsic scale, then Compositor won't have to know about pixels
any longer. Most of Compositor's logic stays in view units, just
blitting to and from back buffers and the cursor save buffer has to be
done in pixels. The back buffer Painter gets a scale applied which
transparently handles all drawing. (But since the backbuffer and cursor
save buffer are also HighDPI, they currently need to be drawn using a
hack temporary unscaled Painter object. This will also go away once
Bitmaps know about their intrinsic scale.)

With this, editing WindowServer.ini to say

  Width=800
  Height=600
  ScaleFactor=2

and booting brings up a fully-functional HighDPI UI.
(Except for minimizing windows, which will crash the window server
until #4932 is merged. And I didn't test the window switcher since the
win-tab shortcut doesn't work on my system.) It's all pixel-scaled,
but it looks pretty decent :^)

### DIFF
--- a/Documentation/HighDPI.md
+++ b/Documentation/HighDPI.md
@@ -18,8 +18,8 @@ Integer scale factors are needed in any case so let's get that working first. Ac
 Desired end state
 -----------------
 
-- Window and Widget rects are in "logical" coordinates, which is the same as pixels at 1x scale
-- Same for mouse cursor
+- All rects  (Window and Widget rects, mouse cursor) are in "logical" coordinates, which is the same as pixels at 1x scale, as much as possible.
+- If something needs to be in pixels, its name starts with `physical_`. Physical coordinates should as much as possible not cross API boundaries.
 - Jury's still out if logical coordinates should stay ints. Probably, but it means mouse cursor etc only have point resolution, not pixel resolution
 - We should have something that can store a collection of (lazily-loaded?) bitmaps and fonts that each represent a single image / font at different scale levels, and at paint time the right representation is picked for the current scale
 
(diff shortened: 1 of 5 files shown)


nico:
whf

nico:
@tomuta

tomuta:
Looks awesome! I know we're not ready for this, but have you thought about multiple `Screen` cases where monitors my have different scale factors? Think about a old school monitor connected to a laptop with a 13" 4k display for example.

nico:
As far as I can tell ("`Screen::the()`"), there isn't any multi-monitor support in serenity yet. It'd be fun to add though! The interesting interaction with highdpi there is probably if a window spans several monitors -- does it get multiple backbuffers, or just one at the highest scale and it's then stretched on the other monitors?

awesomekling:
> As far as I can tell ("`Screen::the()`"), there isn't any multi-monitor support in serenity yet. It'd be fun to add though! The interesting interaction with highdpi there is probably if a window spans several monitors -- does it get multiple backbuffers, or just one at the highest scale and it's then stretched on the other monitors?

Just one at the highest scale seems like the sensible approach here. :)

tomuta:
> Just one at the highest scale seems like the sensible approach here. :)

I envisioned essentially the screen composition to do a `Screen::for_each` and render to each stream, which could apply different scale. I know one scale is going to be a really poor user experience. I had a laptop at work that had a 13" 4k screen, and connecting external displays with normal resolution just can't use the same scale. It would be waaaaay to big on those screens.

awesomekling:
The complication happens in the clients, not in the compositor. Now each client would need to draw into multiple back buffers on each paint event, one for each scale.

nico:
I mean not at framebuffer time. I mean the window's backing buffer could be 2x, and then when compositing to the low-res screen it could downsample at blit time, instead of having two backing buffers for the window.

nico:
> The complication happens in the clients, not in the compositor. Now each client would need to draw into multiple back buffers on each paint event, one for each scale.

(While I agree that I think just one backbuffer seems like the way to go, if the gui library handles all this like it should, the application wouldn't really notice, right? It'd just get its paint callback called once per backing buffer and everything should in theory Just Work.)

PR #4961, opened 2021-01-15
Utilities+WindowServer: Add "chres", give SetResolution a scale parameter, plumb it through

nico:
This adds a new "chres" utility and uses it to implement scale factor plumbing independently of implementing the UI for changing the scale factor in DisplaySettings.

Also, being able to set the resolution from the command line seems neat anyways.

Now, `chres 640 480 2` can set the UI to HighDPI 640x480 at runtime. A
real GUI for changing the display factor will come later.

(`chres 640 480 2` followed by `chres 1280 960` is very fast since
we don't have to re-allocate the framebuffer since both modes use
the exact same number of physical pixels.)


nico:
https://user-images.githubusercontent.com/3487/104775226-00f02f00-5746-11eb-80b3-a993ff4bc8d7.mp4

Userland/Utilities/chres.cpp
```diff
@@ -0,0 +1,50 @@
+/*
+ * Copyright (c) 2020, Nico Weber <thakis@chromium.org>
```
awesomekling, line 2: 2 0 2 0

Userland/Utilities/chres.cpp
```diff
@@ -0,0 +1,50 @@
...
+    int width = -1;
+    int height = -1;
+
+    Core::ArgsParser args_parser;
+    args_parser.set_general_help("Change the screen resolution.");
+    args_parser.add_positional_argument(width, "Width", "width");
+    args_parser.add_positional_argument(height, "Height", "height");
+    args_parser.parse(argc, argv);
+
+    // A Core::EventLoop is all we ened, but WindowServerConnection needs a full Application object.
```
awesomekling, line 44: Typo, need

PR #4963, opened 2021-01-15
CrashReporter: Implement kling's feature requests

linusg:
As promised. Crashing programs just got even more exciting!

---

Featuring:

- Tabs, because there's no way all of this will fit into the UI otherwise.
- Per-thread backtraces
  image
- CPU register states (also per-thread)
  image
- Process arguments and environment
  image

### DIFF
--- a/AK/JsonArray.h
+++ b/AK/JsonArray.h
@@ -47,6 +47,13 @@ public:
     {
     }
 
+    template<typename T>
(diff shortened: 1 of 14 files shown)


linusg:
> E: The repository 'download.opensuse.org/repositories/devel:/kubic:/libcontainers:/stable/xUbuntu_20.04  Release' no longer has a Release file.

:(

PR #4964, opened 2021-01-15
Ports: experimental MilkyTracker port

TheMorc:
I wanted to port MilkyTracker to SerenityOS, I also saw today that @danboid wants it. The only problem being is that it's more experimental and crashy than working.
But I managed to get it to build and run somehow but all of those patches seem to be a bit weird and I don't think I would be able to make it run myself. 
That's atleast why i'm creating a draft pull request so someone else can try messing with it.

_(And it seems that SDL Timer has some problems with pthread?)_


danboid:
Very cool! Is it working well enough to play a mod file?

I think we still have unresolved issues with audio which would prevent perfect playback atm, right @tomuta ?

TheMorc:
@danboid: it was working a bit "better" before dynamic linking was introduced. The problem back then was that SDL Timer wasn't working as it spammed pthread errors, config was also broken and opening files from it just crashed it. Nowadays it opens, shows a bad video mode message(due to simple anti-fullscreen patch) and then does a weird crash that repeats and eats 100% of CPU until you kill the process.
There's surely still a lot of work needed to get it to play but i'm not sure if i'll be able to finish it.

TheMorc:
I'll close this PR for now as there wasn't any movement done since creating it and it's still not working either.

PR #4965, opened 2021-01-15
Kernel+LibC+UserspaceEmulator: Handle PATH_MAX more gracefully

BenWiederhake:
This is the surprisingly short result of a surprisingly long discussion about [long paths](https://github.com/SerenityOS/serenity/discussions/4357). Now 

This covers all cases of that use `MutableBufferArgument` (`readlink` and `realpath`). If this gets general approval, we should implement a similar "retry dance" for other syscalls that can return arbitrarily long paths, at least `getcwd`.

CC @bugaevc, what do you think? :)

### DIFF
--- a/Userland/Libraries/LibC/stdlib.cpp
+++ b/Userland/Libraries/LibC/stdlib.cpp
@@ -1051,6 +1051,8 @@ char* realpath(const char* pathname, char* buffer)
     size_t size = PATH_MAX;
     bool self_allocated = false;
     if (buffer == nullptr) {
+        // Since we self-allocate, try to sneakily use a smaller buffer instead, in an attempt to use less memory.
+        size = 64;
         buffer = (char*)malloc(size);
         self_allocated = true;
     }
@@ -1061,6 +1063,29 @@ char* realpath(const char* pathname, char* buffer)
             free(buffer);
         errno = -rc;
         return nullptr;
+    }
+    if (self_allocated && static_cast<size_t>(rc) > size) {
+        // There was silent truncation, *and* we can simply retry without the caller noticing.
+        free(buffer);
+        size = static_cast<size_t>(rc);
+        buffer = (char*)malloc(size);
+        params.buffer = { buffer, size };
+        rc = syscall(SC_realpath, &params);
+        if (rc < 0) {
+            // Can only happen if we lose a race. Let's pretend we lost the race in the first place.
+            free(buffer);
+            errno = -rc;
+            return nullptr;
+        }
+        size_t new_size = static_cast<size_t>(rc);
+        if (new_size < size) {
+            // If we're here, the symlink has become longer while we were looking at it.
+            // There's not much we can do, unless we want to loop endlessly
+            // in this case. Let's leave it up to the caller whether to loop.
+            free(buffer);
+            errno = EAGAIN;
+            return nullptr;
+        }
     }
     errno = 0;
     return buffer;
--- a/Userland/Libraries/LibC/stdlib.cpp
+++ b/Userland/Libraries/LibC/stdlib.cpp
@@ -1049,11 +1049,16 @@ char* realpath(const char* pathname, char* buffer)
         return nullptr;
     }
     size_t size = PATH_MAX;
-    if (buffer == nullptr)
+    bool self_allocated = false;
+    if (buffer == nullptr) {
         buffer = (char*)malloc(size);
+        self_allocated = true;
+    }
     Syscall::SC_realpath_params params { { pathname, strlen(pathname) }, { buffer, size } };
     int rc = syscall(SC_realpath, &params);
     if (rc < 0) {
+        if (self_allocated)
+            free(buffer);
         errno = -rc;
         return nullptr;
     }
--- a/Userland/Libraries/LibCore/File.cpp
+++ b/Userland/Libraries/LibCore/File.cpp
@@ -209,7 +209,7 @@ String File::read_link(const StringView& link_path)
         return { large_buffer_ptr, new_size };
     // Otherwise, here's not much we can do, unless we want to loop endlessly
     // in this case. Let's leave it up to the caller whether to loop.
-    errno = -EAGAIN;
+    errno = EAGAIN;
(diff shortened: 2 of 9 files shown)


Userland/Libraries/LibC/stdlib.cpp
```diff
@@ -1049,11 +1049,15 @@ char* realpath(const char* pathname, char* buffer)
...
-    if (buffer == nullptr)
+    bool self_allocated = false;
+    if (buffer == nullptr) {
         buffer = (char*)malloc(size);
+        self_allocated = true;
+    }
     Syscall::SC_realpath_params params { { pathname, strlen(pathname) }, { buffer, size } };
     int rc = syscall(SC_realpath, &params);
     if (rc < 0) {
+        free(buffer);
```
emanuele6, line 1060: You forgot did not use `self_allocated`.

emanuele6: Oh, nevermind. you added it in ed3410f but only used it from 7f666635. I didn't notice.

emanuele6: Adding `bool self_allocated = false;` and `self_allocated = true;` in 7f66663 would make more sense in my opinion though.

bugaevc:
This is great! 😀

I was trying to do something similar with `Core::File::read_link()` — please take (another) look at that. So you think it makes sense to do this at the LibC level? Perhaps it does! What are the implications for `Core::File::read_link()` then?

But please don't allocate 4KB for the initial buffer, that's way too much for normal paths.

Userland/Libraries/LibC/stdlib.cpp
BenWiederhake: It's wrong nevertheless, since the new line 1060 doesn't check `self-allocated`. D'oh! :D

BenWiederhake:
I did find the `Core::File::read_link()` code, but initially only checked that it behaves correctly, not as a reference. Looking at it again, I see that:
- It (still) seems correct! (As in: It does the retry-dance properly)
- `If we're here, the symlink has changed while we were looking at it` → Oh! I forgot that corner case. I'll update "my" LibC implementation accordingly.

BenWiederhake:
Oh, sorry, I forgot to answer in full:

> What are the implications for Core::File::read_link() then?

None, since `readlink(3)` cannot reasonably retry, since the size is given as an argument. `File::read_link()` could use `readlink(3)` instead of `readlink(2), but that wouldn't change much.

`File::real_path_for` already uses `realpath(3)`.

> But please don't allocate 4KB for the initial buffer, that's way too much for normal paths.

Sure. What number is reasonable? You chose 64 bytes for a *symlink*, that makes sense, but absolute paths are usually longer than a simple (usually relative) link. What do you think about 128 bytes?

And finally: I just noticed that technically, POSIX says that [because we define `PATH_MAX`, we cannot return anything longer than that from `realpath(3)`.](https://pubs.opengroup.org/onlinepubs/9699919799/) However, that seems stupid, so let's ignore that.

bugaevc:
> What number is reasonable? You chose 64 bytes for a symlink, that makes sense, but absolute paths are usually longer than a simple (usually relative) link. What do you think about 128 bytes?

Actually more paths are reasonably short. Here are a few real-world Serenity paths and their lengths:
|path|length|
|-|-|
|`/bin/wc`|7|
|`/usr/share/man/man2/readlink.md`|31|
|`/res/themes/Coffee.ini`|22|
|`/res/fonts/CsillaRegular10.font`|31|
|`/home/anon/Documents/zip/archive.zip`|36|

BenWiederhake:
Changes since the last commit:
- Fix existing wrong errno in `File::read_link()`
- Fix existing memleak in getcwd(3)
- Fix memleak in this patch  as pointed out by emanuele6
- Fixed the corner-case of changing realpath as pointed out by bugaevc
- Reduce first guess in realpath to merely 64 characters as suggested by bugaevc
- Change `sys$getcwd` and functions surrounding it to do the same "retry-dance"
- Make `get(c)wd` more posixly correct (EINVAL, ERANGE, ENAMETOOLONG)
- Rebase for convenience
- Add a test that nicely demonstrates that it's working:

serenity-overlong

With this, *reading* arbitrarily long paths from the Kernel should always work as intended, especially when using `Core::File`.

In contrast, *writing* paths to the Kernel (e.g. open, move, copy, etc.) still runs into `PATH_MAX`. That is for another PR, if at all.

Userland/Libraries/LibCore/File.cpp
```diff
@@ -209,7 +209,7 @@ String File::read_link(const StringView& link_path)
         return { large_buffer_ptr, new_size };
     // Otherwise, here's not much we can do, unless we want to loop endlessly
     // in this case. Let's leave it up to the caller whether to loop.
-    errno = -EAGAIN;
+    errno = EAGAIN;
```
bugaevc, line 212: Sorry for messing this one up :cry:

Userland/Libraries/LibC/stdlib.cpp
```diff
@@ -1062,6 +1064,29 @@ char* realpath(const char* pathname, char* buffer)
...
+            return nullptr;
+        }
+        size_t new_size = static_cast<size_t>(rc);
+        if (new_size < size) {
+            // If we're here, the symlink has become longer while we were looking at it.
+            // There's not much we can do, unless we want to loop endlessly
+            // in this case. Let's leave it up to the caller whether to loop.
+            free(buffer);
+            errno = EAGAIN;
+            return {};
```
bugaevc, line 1087: `return nullptr;`

(`{}` is the idiomatic way to make a null `String`, but the idiomatic way to make a null `char*` is `nullptr`)

Userland/Libraries/LibCore/File.cpp
BenWiederhake: No problem <3

BenWiederhake:
Changes since the last push:
- default-initialize a `char*` with `nullptr` instead of `{}` (duh!)

PR #4966, opened 2021-01-16
DisplaySettings: Add UI for switching the scale factor

nico:
For now, only support 1x and 2x scale.

I tried doing something "smarter" first where the UI would try
to keep the physical resolution constant when toggling between
1x and 2x, but many of the smaller 1x resolutions map to 2x
logical resolutions that Compositor rejects (e.g. 1024x768 becomes
512x384, which is less than the minimum 640x480 that Compositor
wants) and it felt complicated and overly magical.

So this instead just gives you a 1x/2x toggle and a dropdown
with logical (!) resolutions. That is, 800x600 @ 2x gives you
a physical resolution of 1600x1200.

If we don't like this after trying it for a while, we can change
the UI then.

### DIFF
--- a/Base/etc/WindowServer/WindowServer.ini
+++ b/Base/etc/WindowServer/WindowServer.ini
@@ -1,6 +1,7 @@
 [Screen]
 Width=1024
 Height=768
+ScaleFactor=1
 
 [Theme]
 Name=Default
--- a/Documentation/HighDPI.md
+++ b/Documentation/HighDPI.md
@@ -31,8 +31,8 @@ The plan is to have all applications use highdpi backbuffers eventually. It'll t
 0. Add some scaling support to Painter. Make it do 2x nearest neighbor scaling of everything at paint time for now.
 1. Add scale factor concept to WindowServer. WindowServer has a scaled framebuffer/backbuffer. All other bitmaps (both other bitmaps in WindowServer, as well as everything WindowServer-client-side) are always stored at 1x and scaled up when they're painted to the framebuffer. Things will look fine at 2x, but pixely (but window gradients will be smooth already).
 2. Let DisplaySettings toggle it WindowServer scale. Now it's possible to switch to and from HighDPI dynamically, using UI.
-2. Come up with a system to have scale-dependent bitmap and font resources. Use that to use a high-res cursor bitmaps and high-res menu bar text painting in window server. Menu text and cursor will look less pixely. (And window frames too, I suppose.)
-3. Let apps opt in to high-res window framebuffers, and convert all apps one-by-one
-4. Remove high-res window framebuffer opt-in since all apps have it now.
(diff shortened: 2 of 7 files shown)


nico:
dpi1x
dpi2x

nico:
(fuzzer build is "/home/runner/work/serenity/serenity/Userland/Libraries/LibGfx/Bitmap.cpp:502: undefined reference to `Gfx::Bitmap::to_bitmap_backed_by_anon_fd() const'", not caused by this change, already broken at head.)

tomuta:
Hm that seems rather confusing. I would have expected things to be twice as big when I select that option. At least that's what Windows and Ubuntu do. It would never cross my mind that I'd have to select a resolution other than the monitor's native resolution, half the resolution. Or am I misunderstanding this?

nico:
Hm, I can't see a button for this in Ubuntu. (I'm using https://en.wikipedia.org/wiki/Cinnamon_(desktop_environment) ). Windows only has a ui scale slider and no pixels at all as far as I can tell – I guess it always uses the display's native resolution? (But a qemu window has no such thing as a native resolution.)

nico:
Oh, and I didn't click "Apply" yet in screenshot two. When you click "Apply" after switching from 100% to 200% without touching the resolution dropdown, the qemu window sizes doubles in each dimenson.

tomuta:
> Hm, I can't see a button for this in Ubuntu.

Maybe it requires certain hardware support to show up? Not sure, Here's how it looks like on 20.10:
Screenshot from 2021-01-15 19-30-50

> Windows only has a ui scale slider and no pixels at all as far as I can tell – I guess it always uses the display's native resolution? (But a qemu window has no such thing as a native resolution.)

Well in Windows 10 it's the madness of settings all over the place. I believe the resolution setting is in one place and then you have to click on advanced settings or something and then they have it there. Too lazy to boot into it to take a screenshot, but it's effectively the same as in Ubuntu: You pick the resolution as-is (in most cases you'd pick the native resolution of the display) and then a scale factor where 1x/100% is normal, 2x/200% everything is twice the size.

tomuta:
> Oh, and I didn't click "Apply" yet in screenshot two. When you click "Apply" after switching from 100% to 200% without touching the resolution dropdown, the qemu window sizes doubles in each dimenson.

So then it would use a display resolution of 2048x1536 in this particular example? That sounds confusing when it only shows 1024x768.

nico:
Right, that's what it'd do. As the commit message says, I tried the other behavior first but it gets a bit messy because the lower few resolutions at 2x are below Serenity's minimum resolution.

Exposing logical coordinates everywhere is self consistent with the code too. I think I'd like to try this for a bit and see how it feels. We can definitely iterate on the ui though.

nico:
Another reason why this makes some amount of sense is that when you resize a window, you get an overlay that shows the current window size in logical coordinates, and with this approach the numbers on the overlay match the numbers in the DisplaySettings dialog.

PR #4969, opened 2021-01-16
LibGUI: Don't bubble window events up to parent windows

### LINKED ISSUE #4967: LibGUI: Modal dialogs flow uncaught key events to parent window
Discovered this whilst trying to use TextEditor via keyboard navigation only. I attempted to close TextEditor having unsaved changes via Alt+F4. This opened the 'Unsaved Changes' modal. I struck the 'y' key, more or less expecting this to save changes, however it seems these keys haven't been bound yet in these dialogs (this is another issue.) To my surprise, my keypresses were still flowing to the text editor, even though the modal dialog was in the foreground. Enter/space still action the highlighted button in the dialog, so it seems that only the uncaught keypresses are flowing to the parent window.

### Screenshot 
image

tomuta:
Always accept the events so that they don't bubble up to the
parent object.

Fixes #4967

### DIFF
--- a/Userland/Libraries/LibGUI/Window.cpp
+++ b/Userland/Libraries/LibGUI/Window.cpp
@@ -27,6 +27,7 @@
 #include <AK/HashMap.h>
 #include <AK/JsonObject.h>
 #include <AK/NeverDestroyed.h>
+#include <AK/ScopeGuard.h>
 #include <LibCore/EventLoop.h>
 #include <LibCore/MimeData.h>
 #include <LibGUI/Action.h>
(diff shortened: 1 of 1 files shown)


tomuta:
@awesomekling Not sure if this is the right solution. Should maybe all `Window` events be accepted? Or are there legitimate cases where a parent window would want an event that was technically sent to another window? Seems like `EventLoop::pump` will always bubble them up unless they're accepted.

awesomekling:
> @awesomekling Not sure if this is the right solution. Should maybe all `Window` events be accepted? Or are there legitimate cases where a parent window would want an event that was technically sent to another window? Seems like `EventLoop::pump` will always bubble them up unless they're accepted.

Yeah I think we can accept all events in `Window::event()`. I don't think we ever want them to bubble, as you say :)

tomuta:
Ok, changed to accept all window events.

PR #4970, opened 2021-01-16
WindowServer, LibGUI, SystemMenu: Fix ShutdownDialog theming

nvella:
WindowServer seems to only notify clients of theme changes when they have a Window. As SystemMenu spends most of its time connected to WindowServer without displaying a window, it appears to miss most theme changes. This PR adds a new asynchronous IPC call to WindowServer (`RefreshSystemTheme`,) which causes WindowServer to immediately respond with an `UpdateSystemTheme` message. A relevant helper method is added to LibGUI::Window, and called in SystemMenu::ShutdownDialog's constructor so that it immediately receives and processes an `UpdateSystemTheme` message containing the current system theme shbuf ID.

### DIFF
--- a/Userland/Libraries/LibGUI/Window.cpp
+++ b/Userland/Libraries/LibGUI/Window.cpp
@@ -845,6 +845,11 @@ void Window::schedule_relayout()
     });
 }
 
+void Window::refresh_system_theme()
+{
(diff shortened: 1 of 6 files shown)


awesomekling:
Nice catch :)

PR #4973, opened 2021-01-16
Kernel: Fix inverted logic in KResultOr

BenWiederhake:
This silly inversion has survived so long (for about 9000 commits! 2f82d4fb31af551deeb7955073d0f6bf69b33d2a) because we don't exercise the 'unhappy paths' enough. :^)


awesomekling:
Oopsie!

Issue #3994, opened 2020-11-08
LaunchServer: executable files are executed silently and without warning - may lead to privilege escalation / command execution


bcoles:
When `LaunchServer` was added (#1958) it was decided that executing any file was acceptable as it was not possible to pass command line arguments. This poses several problems.

This service effectively undermines the security benefits enforced with `pledge()` and `unveil()` by allowing execution of arbitrary executable files in a new process (unsandboxed, with the privileges of the user). As such, care should be taken to ensure it is not abused.

Launching a file marked executable (+x) causes the file to be executed immediately, without warning, and silently (except for `dbg()` output in the console logs). This can be used to perform a privilege escalation attack through social engineering, or simply in the event of a misclick. The file is executed *regardless of the file extension*.

Although adoption of `LaunchServer` has been slow, it is currently used in a few places, such as the `Terminal`, `File Manager` and the `WebContent` renderer in `TextEditor`.

# Terminal

Clicking a file called `workbook.sheets` in `Terminal` will execute the file directly as an executable if it has execute +x permissions, rather than open the file in the program associated with the `.sheets` extension (`Spreadsheet`). Although the file is marked with a `*` and green text, this behaviour may still be surprising. Additionally, given that zip archive files preserve their file permissions, it's not uncommon to unzip a compressed file and be presented with a directory full of files with `777` permissions.

This behaviour can be abused to perform a privilege escalation attack through social engineering (`not-a-virus.txt`), or a misclick.

![not-a-virus.txt](https://user-images.githubusercontent.com/434827/98458474-25c7a380-21e5-11eb-8409-cf52656263d3.png)

Additionally, should you happen to click inside the Terminal window, perhaps some time after doing `ls /bin`, and happen to click on the `reboot` executable, the system will reboot without warning, potentially causing loss of data. All your in-progress Piano compositions and Spreadsheet sheets will be lost.

Similarly, the `/bin/TelnetServer` application does not require any command line arguments to start the server and binds on all interfaces `0.0.0.0` by default; however, the connection to `LaunchServer` is performed as `anon`, not `root`, so attempts to bind to the default telnet port 23 will fail. Else this would provide a trivial privilege escalation vector by enabling the server given that the `root` account does not have a password by default.

# File Manager

Similar to Terminal, however, `File Manager` does not use clickable links with `file://` URLs. As such, a double click is required, making this less of an issue.

[...]

PR #4974, opened 2021-01-16
Website: Bounty: Add rule for user interaction/social engineering

bcoles:
"unlikely" is a bit vague.

The spirit of this rule is to prevent claims such as "here, run this totally legit ~~trojan~~ application!" while still rewarding reports for client-side vulnerabilities, such as exploitation of file-format vulnerabilities in image rendering libraries exploited via loading a page in the web browser.

### DIFF
--- a/Meta/Websites/serenityos.org/bounty/index.html
+++ b/Meta/Websites/serenityos.org/bounty/index.html
@@ -21,6 +21,7 @@
         <li>The PoC exploit needs to work against the master branch at the time of claim.</li>
         <li>Max 5 bounties per person.</li>
         <li>No duplicates. If a bug is already reported, only the earliest reporter may claim the reward. This includes bugs found by continuous fuzzing systems.</li>
+        <li>No rewards for bugs that require unlikely user interaction or social engineering.</li>
         <li>Remote bugs must be exploitable with an unmodified "default setup" of SerenityOS. Bugs in programs that are not started by default don't qualify.</li>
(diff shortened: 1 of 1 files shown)


BenWiederhake:
Does this rule exclude privilege escalation?

For example: A user discovers (silly example!) that running `su` with an environment variable of length exactly 4096 somehow accidentally skips the password verification. That's unlikely, but still a security bug.

awesomekling:
That's definitely still a valid bug! This rule change was about excluding "hey, download this totally legit executable and run it on your system to double your RAM!" type issues.

bcoles:
> 
> 
> Does this rule exclude privilege escalation?
> 
> For example: A user discovers (silly example!) that running `su` with an environment variable of length exactly 4096 somehow accidentally skips the password verification. That's unlikely, but still a security bug.

Exploiting `su` with a malicious environment requires no social engineering and no user interaction. If this lead to bypassing authentication it would lead to privilege escalation which is included in the bounty program:

https://github.com/SerenityOS/serenity/blob/c90b7881a7313a9b36c7069769b5489c3be4f014/Meta/Websites/serenityos.org/bounty/index.html#L14-L16

Comparatively, client-side vulnerabilities, such as file format vulnerabilities, are a bit of a gray area. Exploiting file format vulnerabilities via the browser are eligible.

Where as command execution via markdown files opened in TextEditor and requiring the user to click a link in the markdown preview (#3994) are considered too unlikely to be eligible.

PR #4977, opened 2021-01-16
C++ parser-based autocomplete

itamar8910:
This is a draft PR for a C++ parser that's meant to be used by the C++ language server for autocomplete, go-to-symbol, etc.

The parser is designed to be error tolerant, and keeps track of the position spans of the AST nodes, which may come in handy later when we want to do incremental parsing.

There is currently no integration with the C++ language server, but that can come soon I think.

I'm aware of the @denis-campredon's parser in #4575. I took a different approach than denis's parser here (I did use his excellent `ScopeLogger` though :sweat_smile:).
I loosely based this parser on our LibJS parser, and this parser isn't really built around the spec, I'm just trying to parse more things as I go.
I think the parsers for the compiler & parser could merge together some day, but I think it's currently reasonable to develop them separately, because after all their requirements do differ.

Any comments on this would be welcome :smiley:


Base/home/anon/Source/little/test.cpp
```diff
@@ -0,0 +1,8 @@
+// hi
```
alimpfard, line 1: This is a curious commit
it's almost as if you want to blame me for this parser :P 

also, hi comment.

denis-campredon:
Nice work.
I don't really know if we should keep the work split or not.
 - The more we are waiting to merge our two implementations, the more pain and time it will take. 
 - But we are totally working of two things totally different with different strategies. (You are using a lot the positions, I did not touch that at all, ...)

alimpfard:
@denis-campredon tbh positioning information is simple to add from the start, and a very big pain to add when a big chunk of the parser is already written...

itamar8910:
I did some more work on the parser, and integrated it with the c++ language server.
Parser-based autocomplete can be enabled via the "project" menubar in `HackStudio`.

The autocomplete can:
- Lookup variables defined in the available scopes
- Deduce types
- Provide suggestions based on the defined properties of a type

This is still far from being really useful, but i think this is ready to get into the tree at this stage, so i'm marking this as ready for review.

Screenshot:
cpp-autocomplete

itamar8910:
> Nice work.
> I don't really know if we should keep the work split or not.
> 
>     * The more we are waiting to merge our two implementations, the more pain and time it will take.
> 
>     * But we are totally working of two things totally different with different strategies. (You are using a lot the positions, I did not touch that at all, ...)

Yeah.
I think right now it's okay to have two different parsers, because we're both making progress on our stuff simultaneously. 
Once both PRs get merged and are in-tree it might be a good time to think about merging the parsers.

Base/home/anon/Source/little/test.cpp
itamar8910: Yeah I'm not sure why the parser stuff was included in your commit.. fixed that :sweat_smile:

itamar8910:
Noticed that CI doesn't pass, converting this to Draft for the meanwhile

PR #4986, opened 2021-01-17
AK+Everywhere: Check dbgln() format at compiletime

alimpfard:
This is still _very_ WIP.
it's also fairly invasive, I think a better option would be to provide a macro `dbgln(fmt, ...)` -> `dbgln_helper<fmt>(...)` - but that would interfere with the `enabled` thing we currently have.

This caught two bugs as-is, but it seems to have a pretty big impact on compile time.

### DIFF
--- a/AK/Format.h
+++ b/AK/Format.h
@@ -26,6 +26,8 @@
 
 #pragma once
 
+#include <AK/AllOf.h>
+#include <AK/AnyOf.h>
 #include <AK/Array.h>
 #include <AK/GenericLexer.h>
 #include <AK/Optional.h>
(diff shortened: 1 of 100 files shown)


alimpfard:
On further inspection, this is not easily doable(\*) with clang-10 (or even 11) because they don't support class template argument deduction on NTTP parameters (I think clang-12 does).
We _could_ write a macro that produces the length too, but that's a little silly: `dbgln(fmt, ...)` -> `dbgln_helper<StringLiteral<string_size(fmt)>(fmt)>(...)` where `consteval size_t string_size(const char (&str)[N] { return N; }`

But I doubt that would do any favours to our compilation times

alimpfard:
We're basically stuck with either updating to clang-12 (which is probably not an option), or completely disallowing the use of non-literal strings with `dbgln` and using a macro (I found one such case).

alimpfard:
New plan:
- Forget about clang, just disable the check on clang
- Change the one non-literal case to use `vdbgln()`
- "Rename" `dbgln<enabled>(fmt, ...)` to `dbgln_if(enabled, fmt, ...)`

This is a lot less invasive, but it's not usable for most other cases of format (`out`, `outln`, etc.) since those have multiple overloads with incompatible parameter orders.

alimpfard:
cc @asynts, particularly for `dbgln_if` - any better ideas are very welcomed

asynts:
This looks very promising, especially when I get around and separate the parsing and formatting steps properly. (Then all the parsing could happen completely at compile time.)

Can we delay this though?

I've already been working on replacing the remaining `dbg()` statements with `dbgln` and if this were merged first, I would have to rebase the stuff that I already have onto that.

I don't think there is an easy way to do that rebase automatically, so this would be quite a bit of work. Instead we could wait until I finished that and then use a regex to replace all cases at once.

alimpfard:
We can absolutely delay this until the `dbg()`->`dbgln()` conversion is complete, no rush.

tomuta:
I propose we return back to the stream based approach, wrapped in a macro. Which allows something like this:
```c++
dbg("This is a value: " << value);
```
There is no way to mess up the stream formatting. There is also no need to check it at compile time. And last, but not least, it allows to internally wrap it into an `if` e.g. for log level, log facility etc, or it could even be compiled out entirely.

While I'm a big `libfmt` fan for string formatting purposes (e.g. `String::format`), I just haven't been able to appreciate using it for logging purposes. To me, it feels like a small improvement over `printf` style formatting, but the fact that we don't yet have the ability to validate at compile time, plus the fact that stream based logging just seems less error prone and more straight forward to me. There are some advantages for e.g. formatting numbers as hex vs decimal, but those can neatly be solved with stream based formatting as well (and no, I don't like the `std::hex` approach).

awesomekling:
Stream based logging is far more clunky and unpleasant than `dbgln()` so far IMO.
Being able to see the full output string in one chunk of the code instead of scattered between `<<`'s makes a huge difference in readability.
Compile-time validation will come, too. :)

asynts:
The debug macro situation should be resolved with #5183, unless someone comes up with another approach.

alimpfard:
So with that, I assume we're all okay with `dbgln_if(flag, format, args...)` and `dbgln(format, args...)`?

If so, what remains is the tale of `outln()`, `warnln()` and the many wrappers of `AK/Format`; the first two cannot be turned into a macro unless we reorder their arguments and the wrappers cannot be turned into a macro, period.

asynts:
> So with that, I assume we're all okay with dbgln_if(flag, format, args...) and dbgln(format, args...)?

:+1:

  > If so, what remains is the tale of outln(), warnln() and the many wrappers of AK/Format; the first two cannot be turned into a macro unless we reorder their arguments and the wrappers cannot be turned into a macro, period.

Honestly, I completely forgot about that issue. I mean technically it doesn't matter in which order the arguments are defined since we would be using the macro anyways, but it seems a bit hack-y, especially with `outln()` which doesn't take any parameters.

I mean we could easily do this when Clang finally catches up and supports NTTP but that might not happen for a while.

alimpfard:
In this set of patches:
- Let's ignore the other formatting functions for now, we can deal with them later
- Turns out we _also_ support explicit argument indices like `{1}`, so check those too
- somewhat make sure no commits fail to compile, for your bisecting pleasure (two commits left to complete this check)

Kernel/Arch/i386/CPU.cpp
```diff
@@ -71,56 +71,56 @@ extern "C" void pre_init_finished(void);
...
+                                                                                               "    subl $" __STRINGIFY(TRAP_FRAME_SIZE - 4) ", %esp \n"                     \
+                                                                                                                                             "    pushl %esp \n"             \
+                                                                                                                                             "    cld\n"                     \
+                                                                                                                                             "    call enter_trap_no_irq \n" \
+                                                                                                                                             "    call " #title "_handler\n" \
+                                                                                                                                             "    jmp common_trap_exit \n");
```
alimpfard, line 123: ???

whaaaaa?

alimpfard: I dunno why this was being filtered out from the clang-format lint, but the lint filter doesn't apply to the precommit hook.

(probably) cc @linusg and @BenWiederhake

asynts:
> [...] As this does add a fairly large build time overhead (~2m on my machine, == ~30% of previous build time), allow the check to be disabled by adding `-DDBGLN_NO_COMPILETIME_FORMAT_CHECK`

Maybe we should turn the switch around then, make it explicit to turn on. I mean this is mostly to prevent such commits from being merged, so running it in Github Actions should suffice for most uses.

alimpfard:
I thought about that, but it's rather rare to rebuild _everything_, the average change probably won't really notice it, but the errors would be very useful

AK/Format.h
```diff
@@ -35,6 +37,155 @@
...
+        state.saw_explicit_index = true;
+    }
+
+    if (!state.saw_explicit_index)
+        return next_implicit_argument_index++;
+
+    return state.index_value;
+}
+
+// FIXME: We should rather parse these format strings at compile-time if possible.
```
asynts, line 73: I was dreaming of moving the parsing out of `vformat` into `VariadicFormatParams`, that would make this "free". Maybe I'll get around and do this at some point.

asynts:
> I thought about that, but it's rather rare to rebuild everything, the average change probably won't really notice it, but the errors would be very useful

That's actually pretty common if changes are made to some really hot header, like `AK/Format.h`. That requires rebuilding almost everything.

alimpfard:
I'm gonna try make this a little more bearable on the build times, I have _some_ ideas.

On a sidenote, I'd also like to have _something_ show up in the IDE/editor (similar to printf-style warnings) - I wonder if that's at all possible

alimpfard:
| \    | Build time (avg of 2 consecutive builds), hot ccache - change in `AK/Format.h` (one comment)| Slowdown (factor) | Slowdown (offset)
| :--- | :--: | :-: | :-: |
| Without any comptime checks | 284164ms (284s) | 1 | 0s |
| Normal call | 385757ms (385s) | 1.355 | 101s |
| With `requires` | 369153ms (369s) | 1.299 | 85s |
| As concept with variadic parameter types | \<ICE\> | - | - | 
| As concept with just parameter count | 332761ms (332s) | 1.164 | 48s |

I say ship it :^)
(will push this in a bit)

alimpfard:
In this patchset:
- Rebased to get rid of (fake) merge conflicts
- Switched to a concept because `m_concepts_are_faster`

AK/Format.h
alimpfard: Perhaps it could be a standalone constexpr function?
For that matter, perhaps we could make `GenericLexer` constexpr, or write another constexpr one that could handle being used to parse format strings?

With this PR we have a constexpr-enabled string (even though it's extremely primitive), so building on that shouldn't be so hard; I don't think `GenericLexer` does anything that can't be done in compile-time.

I'll cc you if I get to doing this before you do ^^

Userland/Libraries/LibELF/DynamicLoader.cpp
```diff
@@ -433,7 +433,7 @@ DynamicLoader::RelocationResult DynamicLoader::do_relocation(size_t total_tls_si
         //     We could explicitly do them first using m_number_of_relocatoins from DT_RELCOUNT
         //     However, our compiler is nice enough to put them at the front of the relocations for us :)
         dbgln<DYNAMIC_LOAD_DEBUG>("Load address relocation at offset {:#08x}", relocation.offset());
-        dbgln<DYNAMIC_LOAD_DEBUG>("    patch ptr == %p, adding load base address ({:p}) to it and storing {:p}", *patch_ptr, m_dynamic_object->base_address().as_ptr(), *patch_ptr + m_dynamic_object->base_address().as_ptr());
+        dbgln<DYNAMIC_LOAD_DEBUG>("    patch ptr == {%p}, adding load base address ({:p}) to it and storing {:p}", *patch_ptr, m_dynamic_object->base_address().as_ptr(), *patch_ptr + m_dynamic_object->base_address().as_ptr());
```
BenWiederhake: Are you sure that `{%p}` makes sense?

alimpfard: Wow, that's pretty silly lol
I'm quite sure it doesn't make any sense! All the more reason to implement the parser as constexpr

AK/Format.h
alimpfard: I spent today trying to do this (while keeping the interface similar-ish), but it doesn't seem very doable.
I'm not sure how (or even _whether_) the folks over at `{fmt}` did this, since the API we're trying to emulate is not constexpr to begin with...

we could probably this a lot more simple by only having one "formatter" - i.e. the standard formatter, and completely separating the parse/format steps, which would then leave us with a metric tonne of string handling at compiletime. which is a painful experience...

BenWiederhake: If this is "only" about verification and consistency checking, we could make this a separate lint script.

alimpfard: @BenWiederhake not really, parsing at compiletime gives us two main things, one being that we don't have to parse at runtime (so faster formatting), and the other being format string validation and such.
we _could_ have two implementations of the parser, one being constexpr, which would entirely sidestep this issue - but that's just a pain.

I personally prefer to have the code not compile at all than to have a linter scream at me.

PR #4987, opened 2021-01-17
Random cleanups

linusg:


nico:
Don't we have a lint step that checks that code is Auto formatted? Does that not run on gml files?

linusg:
We have such lint steps, but not for GML files - `gml-format` can't handle comments yet (they're not in the JSON-"AST"), which is the only reason holding us back from Lagom-building it for CI use.

PR #4994, opened 2021-01-17
Shell: Allow newlines before `else`, tiny manpage clarification

alimpfard:


Base/usr/share/man/man5/Shell.md
```diff
@@ -182,12 +182,12 @@ rm test && echo "deleted!" || echo "failed with $?"
 
 ##### Conditionals
 Conditionals can either be expressed with the _Logical Relations_, or via explicit `if` expressions.
-An `if` expression contains at least a _condition_ and a _then clause_, and optionally the `else` keyword followed by an _else clause_.
+An `if` expression contains at least a _condition_ command and a _then clause_, and optionally the `else` keyword followed by an _else clause_.
```
alimpfard, line 185: @ADKaster I hope this is clear enough?
if not, could you suggest something that would be more clear?

ADKaster: I think most of my issue was unfamiliarity with how if expressions normally work in shells. The fact that we can do `if expr-that-is-cmd` instead of having to do `if [ expr -eq 0 ]` threw me off. This should be good enough to resolve my issue :D.

Side note, can't comment on the exact line for some reason, but the example block on line 242 is the same as the one on line 220.

alimpfard: :thinking:

alimpfard: Monkey see, monkey copy, monkey forget about extra copy

PR #4996, opened 2021-01-18
Shell: Fix some bugs encountered in #4921

alimpfard:
This PR fixes some issues surfaced by actual use of the Shell (!) in #4921.


alimpfard:
(and then the world crashed and the multiverse burst into flames)

alimpfard:
The only remaining issue is figuring out why this (v) happens, but that's not a big issue, so let's keep it separate.
```
Shell(28): Job entry "expr 15 + 1" deleted in 20 ms
# this subshell shouldn't know about its parent's jobs.
Shell(176): Job entry "expr 15 + 1" deleted in 20 ms
```

PR #5001, opened 2021-01-18
Ports: Add Python 3.9

linusg:
image

See commit messages for (lots of) details. :)


Userland/Libraries/LibM/math.cpp
```diff
@@ -563,4 +563,34 @@ double erfc(double x) NOEXCEPT
...
+
+double nextafter(double, double) NOEXCEPT
+{
+    TODO();
+}
+
+float nextafterf(float, float) NOEXCEPT
+{
+    TODO();
+}
```
alimpfard, line 575: `nextafter` isn't that hard to implement imperfectly, just read it as a u32 (if float) or u64 (if double), add the sign of the difference between the two args, if it's not normal, crap out, otherwise return as the respective type (this will over/underflow, but who cares)

Though I guess there's no reason to add an imperfect implementation instead of a stub?

linusg: sole use of this in CPython is [`math.nextafter()`](https://docs.python.org/3/library/math.html#math.nextafter) so my primary concern was getting it to compile :smile:

But I suppose I can put my brain into number mode and give it (and some other missing `math.h` functions) a try :)

alimpfard: if it's not used anywhere in cpython itself, why bother :shrug:
if someone needs it, they'll implement it correctly lol

Issue #4642, opened 2020-12-29
DynamicLoader: dlopen does not work


ADKaster:
An application attempting to use dlopen will get an assert on m_global_symbol_lookup_func.

dlfcn.cpp has its own map of elf objects, separate from the Loader. And, dlopen does not set the required global symbol lookup function like the loader does in map_library.

Ideally, dlopen in dlfcn.cpp would just be a weak reference, and DynamicLoader/main.cpp would create a strong reference to a method that it itself implements, probably one that calls map_library. 

Such an implementation would require #4627 first though :)

ADKaster:
Example from Iovah on IRC

![](https://usercontent.irccloud-cdn.com/file/wvAYd5PO/image.png)

alimpfard:
I was thinking about this, and wondering why we needed loader.so to define `dlopen` and co, is there any reason why `ELF::DynamicLinker` can't provide a `map_library` (and symbol lookup), which would then be called from libc/dlfcn? am I missing some important bit of information here?

itamar8910:
In the current implementation, the dynamic loader comes with its own copy of LibC & LibELF inside of it (which solves the issue of bootstrapping the dependencies of the loader, but makes `Loader.so` quite chubby at ~1MB).

Because of this, the shared libraries data structures from `ELF::DynamicLinker` do not exist in `libc.so` but rather in `Loader.so`.
So the way I see it, the implementation of dlopen etc would need to be inside `Loader.so`.

Regarding the implementation itself, there's a little caveat - I don't think we can currently support defining thread-local-storage data in libraries loaded with `dlopen`, because of the way we implement thread-local-storage.
If the loaded library defines TLS, we would need to call `allocate_tls` again, and i don't really see a way to do that in a thread-safe manner. But anyways I'm not sure supporting this is a must-have feature for us.

ADKaster:
map library needs to be available in the Loader before the app loads LibC. And when the app calls map_library, it needs to reference the same map_library that the loader uses. So the loader needs to take some special care that if there's a relocation for one of the symbols that it owns, the app will reference its implementation. So if we put a fully fledged version in LibC.so, it will be a waste of space, basically. Also, everything itamar said 😁.

alimpfard:
Hmm...in such a case, this seems like a big case of something-I-can't-wrap-my-head-around.
I tried a little and I couldn't even figure out how to make Loader.so substitute something from its own symbol namespace

ADKaster:
[...]

PR #5004, opened 2021-01-18
(WIP) Ports: Add a super barebones tcl port

alimpfard:
What it has:
- tclsh
- it _can_ load dynamic libs, but it has not really been tested (yet)

What it doesn't have:
- it's not a shared lib (`__tls_get_addr` doesn't exist? not sure)
- no threads


alimpfard:
What a tiny patch...maybe I should instead run `autoreconf` before the configure...(TODO)

ADKaster:
You're correct that ___tls_get_addr doesn't exist yet, need a LOT of work in the Loader and LibC to get proper thread local things going on. 

I wouldn't expect any of the scripting language ports' module loading to work, as `dlopen` hasn't been updated since the Loader was introduced -- it currently just crashes your program with an assert.

alimpfard:
Hmm...I'll try to statically link all of the modules together then. Seems like this PR won't get anywhere near where I hoped it would get.

ADKaster:
See https://github.com/SerenityOS/serenity/issues/4642 for my summary of dlopen. Needs some help to plug itself into the loader, probably with some weak symbols. Either glibc or openbsd that I was looking at last year makes the symbols empty ENOTSUPP stubs in LibC and has strong aliases in the RTLD that know how to hook into the loaders object maps

alimpfard:
sad things:
- `tk` has a fairly hard dependency on X11, it'll take more than a little work to port all of that to LibGUI/LibGfx, and even then, it's not clear whether we can statically link it in.
- `expect cannot be cross-compiled` \*sadness\* - anyone up for making a package manager that doesn't xcompile?

we could either close this PR until we have one of the two things above (`dlopen` and shared lib stuff, or a in-system package manager), or just merge the incomplete tcl port.

Ports/tcl/package.sh
```diff
@@ -0,0 +1,14 @@
...
+export version=8.6.11
+export workdir=tcl8.6.11/unix
+export useconfigure=true
+export configopts="--disable-threads --disable-64bit --disable-shared"
+export files="https://prdownloads.sourceforge.net/tcl/tcl8.6.11-src.tar.gz tcl8.6.11.tar.gz"
+
+pre_configure() {
+    pushd "$workdir" || exit 1
+    autoreconf -fi
+    popd || exit 1
```
linusg, line 13:
```suggestion
    run autoreconf -fi
```

linusg:
> or just merge the incomplete tcl port

For that pease add an entry to https://github.com/SerenityOS/serenity/blob/master/Ports/AvailablePorts.md :)

Ports/tcl/package.sh
alimpfard: ooooh nice

Userland/Libraries/LibELF/DynamicObject.cpp
```diff
@@ -507,6 +507,11 @@ Optional<DynamicObject::SymbolLookupResult> DynamicObject::lookup_symbol(const E
     if (symbol.is_undefined() || symbol.bind() == STB_WEAK)
         return DynamicLinker::lookup_global_symbol(symbol.name());
 
+    if (symbol.bind() == STB_GLOBAL) {
+        if (auto res = DynamicLinker::lookup_global_symbol(symbol.name()); res.has_value())
+            return res;
+    }
```
alimpfard, line 513: @ADKaster @itamar8910 I don't know if this is right (what I've read so far suggests that it could be), but it does resolve some issues I encountered.

In general, this commit is probably weird, but I think it's better to have a half-working implementation and fix it later - please tell me if FIXME's are needed in here somewhere

ADKaster: Oops, I made all those comments on your commit and not the PR 🤦‍♂️. Phone GitHub strikes again!

Anyway re RTLD_NEXT and RTLD_DEFAULT, they're actually reserved by POSIX with the expected behavior defined in the Application Usage section of dlsym. https://pubs.opengroup.org/onlinepubs/009696699/functions/dlsym.html

I think they're pretty useful for applications that load plugins for finding "the symbol foo as if there was a relocation for it" and for things like LibC replacements where you want to LD_PRELOAD a custom malloc or to override a symbol provided in another shared object when using it in this one.

alimpfard: Alright, I'll add a FIXME about it, why not.

alimpfard: (for future reference - mostly for myself lol - the comments are here: https://github.com/SerenityOS/serenity/commit/c0db5db28a829531be480b7f6dda5c067c03f440)

danboid:
What do you need tcl for Ali? I don't think I've installed tcl/tk since the 90's.

alimpfard:
A while back I thought "wouldn't it be cool if tk could use LibGfx+LibGUI for its windows and widgets?"
The first step to that is actually getting tcl, and the next step for that is getting dynamic modules working for it.
then I can attempt to port tk to LibGfx+LibGUI.

> I don't think I've installed tcl/tk since the 90's

Jokes on you, it comes with every python installation :P

danboid:
Oh yes! I'd totally forget about its inclusion in python! Maybe its not quite as obscure now as I was thinking earlier.

alimpfard:
(Closing this because I don't think I feel like working on dlfcn anytime soon)

PR #5005, opened 2021-01-18
Meta: Get building on NixOS

jonathandturner:
Managed to get this building on NixOS and figured I'd turn it into a PR so it didn't get lost. The gist is largely to not use `/bin/bash` or `/bin/mkdir`, since those aren't available on NixOS and instead use `env` to find where they are.

The rest of this is setting up the build dependencies so that both the bootstrap and Serenity can build successfully.

### DIFF
--- a/Ports/dropbear/package.sh
+++ b/Ports/dropbear/package.sh
@@ -1,4 +1,4 @@
-#!/bin/bash ../.port_include.sh
+#!/usr/bin/env -S bash ../.port_include.sh
 port=dropbear
 version=2019.78
 files="https://mirror.dropbear.nl/mirror/releases/dropbear-${version}.tar.bz2 dropbear-${version}.tar.bz2
--- a/Toolchain/BuildPython.sh
+++ b/Toolchain/BuildPython.sh
@@ -1,5 +1,4 @@
 #!/usr/bin/env bash
-
 set -e
 
 # This file will need to be run in bash, for now.
--- a/Userland/Libraries/LibWeb/CMakeLists.txt
+++ b/Userland/Libraries/LibWeb/CMakeLists.txt
@@ -390,7 +390,7 @@ add_custom_target(generate_PropertyID.h DEPENDS CSS/PropertyID.h)
 
 add_custom_command(
     OUTPUT CSS/PropertyID.cpp
-    COMMAND /bin/mkdir -p CSS
+    COMMAND /usr/bin/env -S mkdir -p CSS
     COMMAND ${write_if_different} CSS/PropertyID.cpp CodeGenerators/Generate_CSS_PropertyID_cpp ${CMAKE_CURRENT_SOURCE_DIR}/CSS/Properties.json
     VERBATIM
     DEPENDS Generate_CSS_PropertyID_cpp
@@ -408,7 +408,7 @@ add_custom_target(generate_ValueID.h DEPENDS CSS/ValueID.h)
 
 add_custom_command(
     OUTPUT CSS/ValueID.cpp
-    COMMAND /bin/mkdir -p CSS
+    COMMAND /usr/bin/env -S mkdir -p CSS
     COMMAND ${write_if_different} CSS/ValueID.cpp CodeGenerators/Generate_CSS_ValueID_cpp ${CMAKE_CURRENT_SOURCE_DIR}/CSS/Identifiers.json
     VERBATIM
     DEPENDS Generate_CSS_ValueID_cpp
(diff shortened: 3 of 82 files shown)


Ports/dropbear/package.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash ../.port_include.sh
+#!/usr/bin/env bash ../.port_include.sh
```
bugaevc, line 1: Does it actually work to specify multiple arguments in shebangs?

bugaevc:
Not to sound negative — being able to build Serenity on Guix would be great 😀 —

but the idea that `/bin/env` is a well-known path but `/bin/sh` is not one (and so all the `sh` invocations have to be proxied through `env`) sounds backwards to me. I'd expect any system that places its shell in an unusual location to provide the necessary symlinks or some other mechanism to make existing software work, rather than requiring everything to proxy binary invocations through `env`.

But I realize that ship has sailed, and the design of Nix/Guix is not gonna change because of my preferences.

Ports/dropbear/package.sh
asynts: I remember running into this when changing Serenity's `env`. Some `env` implementations allow that by default and others require a `-S` flag or something like that.

bugaevc: I see, so the kernel passes everything following the initial binary name as a single argument (in this case, `/usr/bin/env "bash ../.port_include.sh"`), and the `env` implementation can then optionally support splitting this as a shell would otherwise do it.

Toolchain/BuildPython.sh
```diff
@@ -1,5 +1,8 @@
 #!/usr/bin/env bash
+<<<<<<< HEAD
```
ADKaster, line 2: Whoops, looks like an unresolved merge conflict snuck in. I also noticed there's a merge commit in your commit list -- we like to follow a rebase + squash workflow so that commits can be directly placed in master from each PR.  https://github.com/SerenityOS/serenity/blob/master/CONTRIBUTING.md#code-submission-policy

Ports/dropbear/package.sh
jonathandturner: Sorry, I'm not quite following - is there a change you'd like me to do here?

Toolchain/BuildPython.sh
jonathandturner: Good catch!

Ports/dropbear/package.sh
bugaevc: You *might* need to add `-S` as @asynts says

jonathandturner: Updated

emanuele6: I think using `-S` breaks support for OpenBSD: https://man.openbsd.org/env.1

Userland/Libraries/LibWeb/CMakeLists.txt
```diff
@@ -390,7 +390,7 @@ add_custom_target(generate_PropertyID.h DEPENDS CSS/PropertyID.h)
 
 add_custom_command(
     OUTPUT CSS/PropertyID.cpp
-    COMMAND /bin/mkdir -p CSS
+    COMMAND /usr/bin/env -S mkdir -p CSS
```
emanuele6, line 393: Can't we somehow use `cmake` to make directories instead of using `/usr/bin/env -S mkdir -p ...`?

emanuele6:
```cmake
add_custom_command(
    ...
    COMMAND ${CMAKE_COMMAND} -E make_directory CSS
    ...
)
```
https://stackoverflow.com/a/3702233

Ports/dropbear/package.sh
jonathandturner: So no `-S`? 😅

emanuele6: @jonathanturner: I don't think it would work without.

linusg:
Why do some shebangs say `#!/usr/bin/env bash` and some say `#!/usr/bin/env -S bash` now?

emanuele6:
@linusg: because of how arguments are passed to programs with shebangs:
file
 
With `#!/usr/bin/env bash`, `env` will get `bash` and the filename as arguments which is fine.
With `#!/usr/bin/env bash ../.port_include.sh`, `env` will get `bash ../.port_include.sh` and the filename as arguments which is not fine because we don't want to run the file with a program called `bash ../.port_include.sh` which probably doesn't exist, we want to run it with `bash`.
To solve this issue, some implementations of `env` have added an `-S` (`--split-string`) "option" which tells `env` that if it gets an argument that looks like `-S  something somethingelse` it should treat it as two (or more, splitting spaces) seperate arguments: `something` and `somethingelse`.
Sadly, since `env` wasn't meant to be used in shebangs originally, this option is not defined in POSIX.

Userland/Libraries/LibWeb/CMakeLists.txt
emanuele6: Also, this didn't need `-S` since it is not a shebang.

PR #5011, opened 2021-01-19
WindowServer: Rename WallpaperMode::{Scaled => Stretch}

### LINKED ISSUE #5006: DisplaySettings: Wallpaper stretching broken
Looks like wallpapers no longer stretch when changing the resolution.

@nico Looks like bisect points to 699ba84bea7c79abab30547f6ac4e97e8cb363a3

Screenshot from 2021-01-19 09-54-00

linusg:
This option was renamed from scaled to stretch in DisplaySettings in 699ba84, but since WindowServer receives a plain string and was not updated, it wouldn't recognize the new renamed value as a valid option.
Turns out sending plain strings via IPC and only mapping them to enum values on the receiving end is brittle, we should probably update `Desktop::set_wallpaper_mode()` to use an enum as well at some point.

Fixes #5006.

### DIFF
--- a/Userland/Services/WindowServer/Compositor.cpp
+++ b/Userland/Services/WindowServer/Compositor.cpp
@@ -58,8 +58,8 @@ static WallpaperMode mode_to_enum(const String& name)
         return WallpaperMode::Tile;
     if (name == "center")
         return WallpaperMode::Center;
-    if (name == "scaled")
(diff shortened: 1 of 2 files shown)


nico:
Thanks!

awesomekling:
Oh yeah that's silly. Would be cool if the IPC code generator could understand enums somehow :)

PR #5012, opened 2021-01-19
Everywhere: Continue replacing dbg() dbgln(). (5)

asynts:
Follow up to #4931.

This should be almost every `dbg()` call. There are a few calls remaining that I couldn't remove yet, since `klog()` is still present in the code base. I'll have to introduce some `klogln` function but that will be done after dealing with the debug macros.

Before:
~~~none
$ grep -rnI 'dbg()' | wc -l
518
~~~

After:
~~~none
$ grep -rnI 'dbg()' | wc -l
18
~~~


AK/Format.cpp
```diff
@@ -652,8 +652,18 @@ void vout(FILE* file, StringView fmtstr, TypeErasedFormatParams params, bool new
 }
 #endif
 
+static bool is_debug_enabled = true;
+
+void set_debug_enabled(bool value)
+{
+    is_debug_enabled = value;
+}
```
alimpfard, line 660: How many of these disabling methods do we have with this? 3?

asynts: Well, `DebugLogStream::set_enabled` will soon go, what other functions are there?

alimpfard: I think we also had an environment variable, although thinking on it, that might've been for `dbg()`

asynts: Yes, we had that in `test-js`, I updated that:

https://github.com/asynts/serenity/blob/bf33690bb3d286200a0669f31c5e3e7fb3649b17/Userland/Utilities/test-js.cpp#L730-L732

asynts:
Not sure what is going on here, the CI reports tons of valid (!) build errors which don't happen when I build locally. I'll try to rebuild my toolchain.

asynts:
It seems the template parameters of `dbgln` aren't properly resolved when it is enclosed in `if constexpr (false)`. This means that it's currently possible to call `dbgln` with invalid parameters when the corresponding debug flag is enabled.

awesomekling:
Oops, this has some merge conflicts :(

PR #5016, opened 2021-01-20
LibGfxScaleDemo: Add coverage for blit_with_alpha

nico:


nico:
(I'm really liking this clip of buggie, we should use it in more places...)

PR #5019, opened 2021-01-20
SystemServer.ini: Default to not showing network change notifications

nico:
Until someone has time to implement something for not showing the
very first network change at boot, let's turn off notifications for
network changes by default altogether. Having to dismiss this
notification at every boot gets old fast.


bcoles:
Related to #4373 which has since been abandoned.

PR #5020, opened 2021-01-20
LibCore: Try to fix fuzzer build

nico:
This might fix https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=29675
See also `man memfd_create`.

### DIFF
--- a/Userland/Libraries/LibCore/AnonymousBuffer.cpp
+++ b/Userland/Libraries/LibCore/AnonymousBuffer.cpp
@@ -24,6 +24,10 @@
  * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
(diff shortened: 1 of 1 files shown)


ADKaster:
The fuzz job in CI passes, do we need to update it to match what oss fuzz is doing somehow?

nico:
CI just builds with `-DBUILD_LAGOM=ON -DENABLE_FUZZER_SANITIZER=ON -DENABLE_ADDRESS_SANITIZER=ON -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++` if I read .github/workflows/cmake.yaml right. oss-fuzz builds with a custom docker container and stuff, see "To run the oss-fuzz build locall" in Meta/Lagom/ReadMe.md. They're doing different things, the latter a superset of the former from what I understand.

PR #5022, opened 2021-01-20
Kernel+CrashReporter: Store pledge violations in coredumps

linusg:
I thought this might be useful :)

image


awesomekling:
Excellent!

PR #5026, opened 2021-01-20
oss-fuzz: Try harder to fix build

nico:
Apparently memfd_create() is newish in glibc, and oss-fuzz
uses Ubuntu 16.04 as base for its docker images, which doens't
yet have memfd_create(). But, not to worry, it does have the syscall
define and that's all we really need :/

---

With this, my local oss-fuzz-like build gets farther, but then complains about libmprf.so later on. Not sure if this is due to me holding things wrong or if that's genuinely newly broken in some way.


nico:
Looks like this did the trick \o/

PR #5034, opened 2021-01-21
Libraries: Add missing headers

BenWiederhake:
A C++ source file containing just
```
#include <LibFoo/Bar.h>
```
should always compile cleanly.
    
This patch adds missing header inclusions that could have caused weird error messages like this one if they were used without the expected context:

```
[5/821] Building CXX object Userland/HeaderCheck/CMakeFiles/H...land__Libraries__LibWeb__Painting__PaintContext.h__test.cpp.o
FAILED: Userland/HeaderCheck/CMakeFiles/HeaderCheck.dir/Userland__Libraries__LibWeb__Painting__PaintContext.h__test.cpp.o 
/usr/bin/ccache /home/eispin/workspace/serenity/Toolchain/Local/i686/bin/i686-pc-serenity-g++ -DDEBUG -DSANITIZE_PTRS -I../Userland/Libraries -I../. -I../Userland/Libraries/LibC -I../Userland/Libraries/LibM -I../Userland/Services -I../Userland -I. -IUserland/Services -IUserland/Libraries -IUserland -pie -fpic -Wno-unknown-warning-option -Wall -Wextra -Werror -Wmissing-declarations -Wformat=2 -fdiagnostics-color=always -ftls-model=initial-exec -fconcepts -Os -g1 -fno-exceptions -fstack-protector-strong -Wno-address-of-packed-member -Wundef -Wcast-qual -Wwrite-strings -Wimplicit-fallthrough -Wno-nonnull-compare -Wno-deprecated-copy -Wno-expansion-to-defined -ffile-prefix-map=/home/eispin/workspace/serenity=. -std=c++2a -MD -MT Userland/HeaderCheck/CMakeFiles/HeaderCheck.dir/Userland__Libraries__LibWeb__Painting__PaintContext.h__test.cpp.o -MF Userland/HeaderCheck/CMakeFiles/HeaderCheck.dir/Userland__Libraries__LibWeb__Painting__PaintContext.h__test.cpp.o.d -o Userland/HeaderCheck/CMakeFiles/HeaderCheck.dir/Userland__Libraries__LibWeb__Painting__PaintContext.h__test.cpp.o -c Userland/HeaderCheck/Userland__Libraries__LibWeb__Painting__PaintContext.h__test.cpp
In file included from .././Userland/Libraries/LibWeb/Painting/PaintContext.h:32,
                 from Userland/HeaderCheck/Userland__Libraries__LibWeb__Painting__PaintContext.h__test.cpp:1:
../Userland/Libraries/LibWeb/SVG/SVGContext.h:59:19: error: field 'm_states' has incomplete type 'AK::Vector<Web::SVGContext::State>'
   59 |     Vector<State> m_states;
      |                   ^~~~~~~~
In file included from ../Userland/Libraries/LibGfx/Palette.h:29,
                 from .././Userland/Libraries/LibWeb/Painting/PaintContext.h:30,
                 from Userland/HeaderCheck/Userland__Libraries__LibWeb__Painting__PaintContext.h__test.cpp:1:
.././AK/Forward.h:133:7: note: declaration of 'class AK::Vector<Web::SVGContext::State>'
  133 | class Vector;
      |       ^~~~~~
ninja: build stopped: subcommands failed.
```

Also, this used to confuse QtCreator; now it is no longer getting confused! :)

- Here's the tool that will compile many of Serenity's headers: https://github.com/BenWiederhake/serenity/tree/dev-header-check
  (I didn't include it in this PR because it's kinda ugly, slows down the build immensely, and is likely to generate false-positives. Feel free to use it if you're masochistic.)
- I excluded some tools (UserspaceEmulator, Shell), because they don't work with the tool cleanly.
- I dropped the diffs for HackStudio and WindowServer, because there the argument "Well, someone might want to use the header stand-alone!" doesn't make much sense. See 25dbfeb9818d7c320feae86fffb152993f0d64f3 if you're interested.
- I excluded the entire Kernel, because it *really* doesn't like being compiled in Userland context.

### DIFF
--- a/Userland/Applications/Run/main.cpp
+++ b/Userland/Applications/Run/main.cpp
@@ -28,6 +28,7 @@
 #include <AK/StringBuilder.h>
 #include <AK/Types.h>
 #include <AK/URL.h>
+#include <LibGUI/Application.h>
 #include <string.h>
 
 #include "RunWindow.h"
--- a/Userland/Demos/CatDog/main.cpp
+++ b/Userland/Demos/CatDog/main.cpp
@@ -26,6 +26,7 @@
  */
 
 #include <LibCore/ElapsedTimer.h>
+#include <LibGUI/Application.h>
 #include <LibGUI/BoxLayout.h>
 #include <LibGUI/Icon.h>
 #include <LibGUI/Menu.h>
--- a/Userland/Libraries/LibAudio/Loader.h
+++ b/Userland/Libraries/LibAudio/Loader.h
@@ -30,6 +30,7 @@
 #include <AK/RefCounted.h>
 #include <AK/RefPtr.h>
 #include <AK/StringView.h>
+#include <LibAudio/Buffer.h>
 #include <LibCore/File.h>
 
 namespace Audio {
--- a/Userland/Libraries/LibC/sys/un.h
+++ b/Userland/Libraries/LibC/sys/un.h
@@ -26,6 +26,7 @@
 
 #pragma once
 
+#include <bits/stdint.h>
 #include <sys/cdefs.h>
 
 __BEGIN_DECLS
--- a/Userland/Libraries/LibCrypto/ASN1/PEM.h
+++ b/Userland/Libraries/LibCrypto/ASN1/PEM.h
@@ -32,7 +32,7 @@
 
 namespace Crypto {
 
-static ByteBuffer decode_pem(ReadonlyBytes data_in, size_t cert_index = 0)
+static inline ByteBuffer decode_pem(ReadonlyBytes data_in, size_t cert_index = 0)
 {
     size_t i { 0 };
     size_t start_at { 0 };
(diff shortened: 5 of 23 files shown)


Userland/HeaderCheck/Counter.h
```diff
@@ -0,0 +1,39 @@
...
+class Counter {
+public:
+    Counter() { s_instances += 1; }
+
+    static int instances() { return s_instances; }
+    static int expected() { return s_expected; }
+
+private:
+    static int s_instances;
+    static int s_expected;
```
alimpfard, line 38: This probably belongs with 25dbfeb

awesomekling:
This is neat, can we do the reverse as well and remove not-actually-needed headers? :D

BenWiederhake:
I don't know how to approach that ^^'

One *could* in theory try to remove one and see if it still compiles, but that sounds like a bad idea. Or one could try to parse the cpp and h files and find any actual dependencies, but that's too close to writing a C++ compiler. That sounds like an even worse idea.

Nope, this PR will stick to *adding* missing headers. :P

Changes since last push: Fixed the spurious file as per @alimpfard's suggestion.

PR #5037, opened 2021-01-22
PixelPaint: Export PNG

phoffmeister:
This is a very simple implementation. Don't know if this has any advantages over the BMP format (other than the alpha) since I'm not using any real compression here.

In the end I was just annoyed that I couldn't export to PNG so I made this :D

### DIFF
--- a/Userland/Libraries/LibGfx/PNGWriter.cpp
+++ b/Userland/Libraries/LibGfx/PNGWriter.cpp
@@ -1,0 +1,228 @@
+/*
+ * Copyright (c) 2021, Pierre Hoffmeister
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include "PNGWriter.h"
+#include <AK/String.h>
+#include <LibCrypto/Checksum/CRC32.h>
+
+namespace Gfx {
+
+class PNGChunk {
+public:
+    PNGChunk(const String&);
+    const Vector<u8>& data() const { return m_data; };
+    const String& type() const { return m_type; };
+    void add_u8(u8);
+    void add_u16_big(u16);
+    void add_u32_big(u32);
+    void add_u16_little(u16);
+    void add_u32_little(u32);
+
+private:
+    Vector<u8> m_data;
+    String m_type;
+};
+
+class NonCompressibleBlock {
+public:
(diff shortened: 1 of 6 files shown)


Userland/Libraries/LibGfx/PNGWriter.cpp
```diff
@@ -0,0 +1,195 @@
...
+
+#include "PNGWriter.h"
+#include <AK/String.h>
+#include <LibCrypto/Checksum/CRC32.h>
+#include <stdio.h>
+
+namespace Gfx {
+
+PNGChunk::PNGChunk(const String& type)
+    : m_type { move(type) }
```
emanuele6, line 35: We usually always use parens-initialisers in constructors' member initialiaser lists.
```suggestion
    : m_type(move(type))
```

emanuele6 (changes requested):

Userland/Libraries/LibGfx/PNGWriter.h
```diff
@@ -0,0 +1,89 @@
...
+class PNGChunk {
+public:
+    PNGChunk(const String&);
+    const Vector<u8>& data() { return m_data; };
+    const String& type() { return m_type; };
+    void add_u8(const u8&);
+    void add_u16_big(const u16&);
+    void add_u32_big(const u32&);
+    void add_u16_little(const u16&);
+    void add_u32_little(const u32&);
```
emanuele6, line 45: These parameters can be passed by value, no?

Userland/Libraries/LibGfx/PNGWriter.cpp
phoffmeister: all right.

Userland/Libraries/LibGfx/PNGWriter.h
phoffmeister: Oopsi. Yeah sure. Done :)

Userland/Libraries/LibGfx/PNGWriter.h
```diff
@@ -0,0 +1,89 @@
...
+    u32 adler_s2() { return m_adler_s2; }
+
+private:
+    void add_block_to_chunk(PNGChunk&, bool);
+    void clear() { m_non_compressible_data.clear(); };
+    void update_adler(u8);
+    bool full() { return m_non_compressible_data.size() == 65535; }
+    Vector<u8> m_non_compressible_data;
+    u32 m_adler_s1;
+    u32 m_adler_s2;
```
emanuele6, line 73: I don't think this class needs a non-trivial constructor.
```suggestion
    void finalize(PNGChunk&);
    void add_byte_to_block(u8 data, PNGChunk&);

    u32 adler_s1() { return m_adler_s1; }
    u32 adler_s2() { return m_adler_s2; }

private:
    void add_block_to_chunk(PNGChunk&, bool);
    void clear() { m_non_compressible_data.clear(); };
    void update_adler(u8);
    bool full() { return m_non_compressible_data.size() == 65535; }
    Vector<u8> m_non_compressible_data;
    u32 m_adler_s1 { 1 };
    u32 m_adler_s2 { 0 };
```

phoffmeister: Done. :+1:

Userland/Libraries/LibGfx/PNGWriter.cpp
```diff
@@ -0,0 +1,195 @@
...
+    add_block_to_chunk(chunk, true);
+}
+
+void NonCompressibleBlock::update_adler(u8 data)
+{
+    m_adler_s1 = (m_adler_s1 + data) % 65521;
+    m_adler_s2 = (m_adler_s2 + m_adler_s1) % 65521;
+}
+
+void PNGWriter::add_chunk(PNGChunk& png_chunk)
```
awesomekling, line 112: Seems like this should take a `const PNGChunk&` since it's not mutating it.

Userland/Libraries/LibGfx/PNGWriter.cpp
```diff
@@ -0,0 +1,195 @@
...
+{
+    auto combined = Vector<u8> {};
```
awesomekling, line 114: `Vector<u8> combined;`

Userland/Libraries/LibGfx/PNGWriter.h
```diff
@@ -0,0 +1,82 @@
...
+    u32 adler_s2() { return m_adler_s2; }
+
+private:
+    void add_block_to_chunk(PNGChunk&, bool);
+    void update_adler(u8);
+    bool full() { return m_non_compressible_data.size() == 65535; }
+    Vector<u8> m_non_compressible_data;
+    u32 m_adler_s1 { 1 };
+    u32 m_adler_s2 { 0 };
+};
```
awesomekling, line 67: These classes are implementation details of the `PNGWriter` class, let's move them into the .cpp file.

Userland/Libraries/LibGfx/PNGWriter.h
```diff
@@ -0,0 +1,82 @@
...
+
+class PNGWriter {
+public:
+    void save_bitmap_to_file(const RefPtr<Bitmap>, const String&);
```
awesomekling, line 71: It would be better to have an API that generates an in-memory PNG. Then we can decide what to do with it, perhaps we don't want to save it to disk.

awesomekling (changes requested):

Userland/Libraries/LibGfx/PNGWriter.cpp
phoffmeister: Done.

Userland/Libraries/LibGfx/PNGWriter.cpp
phoffmeister: Done.

Userland/Libraries/LibGfx/PNGWriter.h
phoffmeister: Done.

Userland/Libraries/LibGfx/PNGWriter.h
phoffmeister: Sure. I applied the same logic as in BMPWriter now.

PR #5038, opened 2021-01-22
Ports: ProTracker 2 clone

TheMorc:
So after trying to port MilkyTracker and failing miserably, i tried to port something else. It's a port of pt2-clone by 8bitbubsy. I also need to thank @danboid because i was searching through issues here and this clone(or ft2-clone?) was mentioned in one of them.
This one atleast works and plays modules in comparison with the MilkyTracker port but has a small issue with mouse not being captured even after disabling HWCURSOR in protracker.ini

Here's a video if you want to see it in action.
https://www.youtube.com/watch?v=gxZfQc_plwI

(also note: SDL2 or this might not compile due to the recent changes in Gfx::Bitmap https://github.com/SerenityOS/SDL/pull/4)

Screen Shot 2021-01-22 at 14 43 29

### DIFF
--- a/Ports/pt2-clone/package.sh
+++ b/Ports/pt2-clone/package.sh
@@ -1,0 +1,16 @@
+#!/bin/bash ../.port_include.sh
+port=pt2-clone
+version=1.28
+workdir=pt2-clone-$version
+useconfigure=true
+files="https://github.com/8bitbubsy/pt2-clone/archive/v${version}.tar.gz v${version}.tar.gz"
+configopts="-DCMAKE_TOOLCHAIN_FILE=$SERENITY_ROOT/Toolchain/CMakeToolchain.txt"
+depends="SDL2"
+
+configure() {
+    run cmake $configopts
+}
+
+install() {
+    run make install
+}
--- a/Ports/pt2-clone/patches/PATH_MAX.patch
+++ b/Ports/pt2-clone/patches/PATH_MAX.patch
@@ -1,0 +1,78 @@
+Common subdirectories: pt2-clone-1.28/src/gfx and pt2-clone-1.28d/src/gfx
+diff -u pt2-clone-1.28/src/pt2_config.c pt2-clone-1.28d/src/pt2_config.c
+--- pt2-clone-1.28/src/pt2_config.c	2021-01-01 20:48:32.000000000 +0100
++++ pt2-clone-1.28d/src/pt2_config.c	2021-01-22 13:34:22.058489000 +0100
+@@ -21,6 +21,7 @@
+ #include "pt2_diskop.h"
+ #include "pt2_textout.h"
+ #include "pt2_sampler.h"
++#define PATH_MAX 4096
+ 
+ #ifndef _WIN32
+ static char oldCwd[PATH_MAX];
+diff -u pt2-clone-1.28/src/pt2_diskop.c pt2-clone-1.28d/src/pt2_diskop.c
+--- pt2-clone-1.28/src/pt2_diskop.c	2021-01-01 20:48:32.000000000 +0100
++++ pt2-clone-1.28d/src/pt2_diskop.c	2021-01-22 13:33:49.531212000 +0100
+@@ -36,6 +36,7 @@
+ #include "pt2_visuals.h"
+ #include "pt2_sample_loader.h"
+ #include "pt2_bmp.h"
++#define PATH_MAX 4096
+ 
+ typedef struct fileEntry_t
+ {
+diff -u pt2-clone-1.28/src/pt2_edit.c pt2-clone-1.28d/src/pt2_edit.c
+--- pt2-clone-1.28/src/pt2_edit.c	2021-01-01 20:48:32.000000000 +0100
++++ pt2-clone-1.28d/src/pt2_edit.c	2021-01-22 13:33:14.987979000 +0100
+@@ -29,6 +29,7 @@
+ #include "pt2_audio.h"
+ #include "pt2_sync.h"
+ #include "pt2_chordmaker.h"
++#define PATH_MAX 4096
+ 
+ const int8_t scancode2NoteLo[52] = // "USB usage page standard" order
+ {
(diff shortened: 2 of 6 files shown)


Ports/pt2-clone/package.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash ../.port_include.sh
+port=pt2-clone
+version=1.28
+workdir=pt2-clone-$version
+useconfigure=true
+files="https://github.com/8bitbubsy/pt2-clone/archive/v$version.tar.gz v$version.tar.gz"
```
linusg, line 6:
```suggestion
files="https://github.com/8bitbubsy/pt2-clone/archive/v${version}.tar.gz v${version}.tar.gz"
```

Ports/pt2-clone/package.sh
```diff
@@ -0,0 +1,18 @@
...
+configopts="-DCMAKE_TOOLCHAIN_FILE=$SERENITY_ROOT/Toolchain/CMakeToolchain.txt"
+##makeopts="-I../../SDL2/SDL-master-serenity/include/ -lSDL2 -j$(nproc)"
+installopts=""
```
linusg, line 9:
```suggestion
```

Ports/pt2-clone/patches/add-dummy-configure.patch
```diff
@@ -0,0 +1,7 @@
+diff --git a/configure b/configure
+new file mode 100644
+index 0000000..04bfb39
+--- /dev/null
++++ b/configure
+@@ -0,0 +1 @@
++#nop
```
linusg, line 7: What's that for? If this doesn't have a configure script, just omit the `useconfigure=true` (defaults to false).

Ports/pt2-clone/patches/PATH_MAX.patch
```diff
@@ -0,0 +1,78 @@
+Common subdirectories: pt2-clone-1.28/src/gfx and pt2-clone-1.28d/src/gfx
```
linusg, line 1: Never seen this in a patch before, what does it mean?

linusg (changes requested):

Ports/pt2-clone/patches/PATH_MAX.patch
TheMorc: i just used diff -u so i have a one big patch with all changed files and it seems like it left this there

Ports/pt2-clone/patches/add-dummy-configure.patch
TheMorc: I have it there so i can run cmake. Or is there some other way to run cmake without needing to use useconfigure?

linusg: ohh, you're running cmake in the configure step, and that expects a configure script. Let's do this instead:

```diff
--- a/Ports/.port_include.sh
+++ b/Ports/.port_include.sh
@@ -151,6 +151,7 @@ func_defined pre_configure || pre_configure() {
     :
 }
 func_defined configure || configure() {
+    chmod +x "${workdir}"/"$configscript"
     run ./"$configscript" --host=i686-pc-serenity $configopts
 }
 func_defined build || build() {
@@ -252,7 +253,6 @@ do_configure() {
     if [ "$useconfigure" = "true" ]; then
         echo "Configuring $port!"
         pre_configure
-        chmod +x "${workdir}"/"$configscript"
         configure
     else
         echo "This port does not use a configure script. Skipping configure step.
```

when `configure()` is overridden we should not assume that `configscript` will be run or even exists.

TheMorc: So this should be resolved now

TheMorc:
I added a fix for the mouse, so that's also working now

danboid:
Big thanks @TheMorc ! I've got it to build but I've not got it to load or play a mod file just yet. This is going to be very cool with a few tweaks.

Using the PT2 file browser is SUPER sluggish! On my recentish i7 laptop, it takes a few minutes to show the contents of a directory, every time you enter one.

```
DHCPClient(17): Received 548 bytes
DHCPClient(17): We were offered 10.0.2.15 for 2152792320
DHCPClient(17): Leasing the IP 10.0.2.15 for adapter e1k0
DHCPClient(17): Received 548 bytes
Terminal(23): Load config file from /home/anon/.config/Terminal.ini
FileManager (Desktop)(13): Watching /home/anon/Desktop for changes, m_watch_fd=11
[WindowServer(16:16)]: kmalloc(): Adding memory to heap at V0xc0801000, bytes: 1048576
[pt2-clone(27:27)]: CPU[0] NP(error) fault at invalid address V0x00000008
[pt2-clone(27:27)]: copy_from_user(0xc03e5eec, 0x00000008, 4) failed at V0x00000008
[pt2-clone(27:27)]: CPU[0] NP(error) fault at invalid address V0x00000008
[pt2-clone(27:27)]: copy_from_user(0xc03e5f88, 0x00000008, 4) failed at V0x00000008
SERENITY_CreateDevice 0
SDL2: Initialising SDL application
SDL2: Creating new window of size 640x510
SerenitySDLWidget::resize_event
w->m_window size = 640x510
Couldn't find matching render driverSDL2: Creating a new framebuffer of size 640x510
Created framebuffer 640x510
Cursors are not currently supported[pt2-clone(27:30)]: pt2-clone(27:30) Should already die
[pt2-clone(27:31)]: pt2-clone(27:31) Should already die
[pt2-clone(27:32)]: pt2-clone(27:32) Should already die
[pt2-clone(27:33)]: pt2-clone(27:33) Should already die
[pt2-clone(27:34)]: pt2-clone(27:34) Should already die
```

It would be nice to have a shortcut icon under the Sound menu and if we could tidy up the cursor so it we don't have two.

danboid:
I'm surprised at how well mod playback works under pt2-clone already. I can violently shake the window and playback doesn't suffer.

linusg:
Please add an entry to https://github.com/SerenityOS/serenity/blob/master/Ports/AvailablePorts.md :)

PR #5046, opened 2021-01-22
Kernel: Use system time as entropy if CPU doesn't support RDRAND

### LINKED ISSUE #4490: Kernel: Boot hangs on KernelRng::wait_for_entropy() when no RDRAND or RDSEED is supported
It seems that the init_stage2 process is stuck in `KernelRng::the().wait_for_entropy();`
more specifically, in
```
[init_stage2(1:1)]: 0xc0176361  Kernel::get_good_random_bytes(unsigned char*, unsigned long) +98
[init_stage2(1:1)]: 0xc017e650  Kernel::Process::load(AK::NonnullRefPtr<Kernel::FileDescription>, AK::RefPtr<Kernel::FileDescription, AK::RefPtrTraits<Kernel::FileDescription> >) +1390
[init_stage2(1:1)]: 0xc017ed6d  Kernel::Process::do_exec(AK::NonnullRefPtr<Kernel::FileDescription>, AK::Vector<AK::String, 0ul>, AK::Vector<AK::String, 0ul>, AK::RefPtr<Kernel::FileDescription, AK::RefPtrTraits<Kernel::FileDescription> >, Kernel::Thread*&, unsigned int&) +889
[init_stage2(1:1)]: 0xc0180bc7  Kernel::Process::exec(AK::String, AK::Vector<AK::String, 0ul>, AK::Vector<AK::String, 0ul>, int) +2519
[init_stage2(1:1)]: 0xc0173f84  Kernel::Process::create_user_process(AK::RefPtr<Kernel::Thread, AK::RefPtrTraits<Kernel::Thread> >&, AK::String const&, unsigned int, unsigned int, AK::DistinctNumeric<int, false, true, true, false, false, false, __ProcessID_tag>, int&, AK::Vector<AK::String, 0ul>&&, AK::Vector<AK::String, 0ul>&&, Kernel::TTY*) +3164
[init_stage2(1:1)]: 0xc01b3be1  Kernel::init_stage2(void*) +4107
[init_stage2(1:1)]: 0xc01190e0  exit_kernel_thread +
```
`SCHEDULER_RUNNABLE_DEBUG`:
```
[colonel(0:0)]: Scheduler[0]: Runnables:
[colonel(0:0)]: Scheduler[0]: Non-runnables:
[colonel(0:0)]:   FinalizerTask FinalizerTask(3:3) @ 0008:c0119c17
[colonel(0:0)]:   Invalid      SystemServer(4:4) @ 001b:0
[colonel(0:0)]:   KernelRng    init_stage2(1:1) @ 0008:c0119c17
[colonel(0:0)]:   Sleeping     SyncTask(2:2) @ 0008:c0119c17
```
Works after pressing a key on keyboard ~30 times.
Note: My CPU doesn't support RDRAND or RDSEED, it may be a reason. When disabled KVM, everything worked.

sppmacd:
We don't have anything better for these CPUs for now, but should be enough since it's better than hanging on black screen :).

Fixes #4490.

### DIFF
--- a/Kernel/Random.cpp
+++ b/Kernel/Random.cpp
@@ -30,6 +30,7 @@
 #include <Kernel/Devices/RandomDevice.h>
 #include <Kernel/Random.h>
 #include <Kernel/Time/HPET.h>
+#include <Kernel/Time/RTC.h>
 #include <Kernel/Time/TimeManagement.h>
 
 namespace Kernel {
@@ -65,13 +66,22 @@ KernelRng::KernelRng()
 
             this->resource().add_random_event(value, i % 32);
         }
-    } else {
+    } else if (TimeManagement::the().can_query_precise_time()) {
         // Add HPET as entropy source if we don't have anything better.
-        klog() << "KernelRng: Using HPET as entropy source (bad!)";
+        klog() << "KernelRng: Using HPET as entropy source";
 
         for (size_t i = 0; i < resource().pool_count * resource().reseed_threshold; ++i) {
             u64 hpet_time = HPET::the().read_main_counter();
             this->resource().add_random_event(hpet_time, i % 32);
+        }
+    } else {
+        // Fallback to RTC
+        klog() << "KernelRng: Using RTC as entropy source (bad!)";
+        time_t current_time = RTC::now();
+        for (size_t i = 0; i < resource().pool_count * resource().reseed_threshold; ++i) {
+            this->resource().add_random_event(current_time, i % 32);
+            current_time *= 0x574a;
+            current_time += 0x40b2;
         }
     }
 }
--- a/Kernel/Random.cpp
+++ b/Kernel/Random.cpp
@@ -29,6 +29,7 @@
 #include <Kernel/Arch/i386/CPU.h>
 #include <Kernel/Devices/RandomDevice.h>
 #include <Kernel/Random.h>
+#include <Kernel/Time/HPET.h>
 #include <Kernel/Time/TimeManagement.h>
 
 namespace Kernel {
@@ -45,6 +46,7 @@ KernelRng::KernelRng()
     bool supports_rdseed = Processor::current().has_feature(CPUFeature::RDSEED);
     bool supports_rdrand = Processor::current().has_feature(CPUFeature::RDRAND);
     if (supports_rdseed || supports_rdrand) {
+        klog() << "KernelRng: Using RDSEED or RDRAND as entropy source";
         for (size_t i = 0; i < resource().pool_count * resource().reseed_threshold; ++i) {
             u32 value = 0;
             if (supports_rdseed) {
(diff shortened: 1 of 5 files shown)


tomuta:
In absence of these, we could also fall back to read the HPET main counter (without looping, as we don't necessarily care about the 32 bit wrap around).

sppmacd:
> HPET main counter

Is it `TimeManagement::uptime_ms()` or something related?

tomuta:
I think you'd want to read it similar to `read_register_safe64` in HPET.cpp, except just read the `low` and `high` without doing a loop. `TimeManagement::uptime_ms` only returns a coarse timestamp that only gets updated when the timer interrupt fires. The HPET main counter continuously updates at much, much faster intervals.

sppmacd:
Now it uses `HPET::update_time` which uses `read_register_safe64` to read value from main counter.

Kernel/Random.cpp
```diff
@@ -63,6 +65,17 @@ KernelRng::KernelRng()
...
         }
+    } else {
+        // Add HPET as entropy source if we don't have anything better.
+        klog() << "KernelRng: Using HPET as entropy source (bad!)";
+
+        u64 seconds_since_boot;
+        u32 ticks_this_second;
+        u64 hpet_time = HPET::the().update_time(seconds_since_boot, ticks_this_second, true);
+        for (size_t i = 0; i < resource().pool_count * resource().reseed_threshold; ++i) {
+            this->resource().add_random_event(hpet_time, i % 32);
```
tomuta, line 76: Please do not call `HPET::update_time` for this purpose. Instead, add a new method that reads the main counter without a loop, and do this maybe inside this loop. Also, `% 32` essentially just discards the upper 64 bits, so you might as well just read the lower 32 bits. Or you just feed the entire 64 bit value...

tomuta (changes requested):

Kernel/Random.cpp
sppmacd: Done.

Kernel/Random.cpp
```diff
@@ -63,6 +65,14 @@ KernelRng::KernelRng()
 
             this->resource().add_random_event(value, i % 32);
         }
+    } else {
```
tomuta, line 68: You'll need to check that the system actually has a HPET timer, probably by checking `TimeManagement::m_can_query_precise_time`

tomuta (changes requested):

Kernel/Random.cpp
sppmacd: Now it uses HPET if it's found, RTC otherwise.

tomuta:
@awesomekling I think it looks good. Haven't tested it, though. The only thing I don't like, but we can change this later, is that `TimeManagement::can_query_precise_time` is used to check for the presence of HPET. Which right now is the only case, but maybe an API named `TimeManagement::have_hpet` or something makes more sense there. Again, minor detail that we can solve later. If it solves the issue I'd just merge it for now.

tomuta (approved):

awesomekling:
Ok cool. @sppmacd please fix the lint CI issue (clang-format) and we can merge this

sppmacd:
Done :)

PR #5058, opened 2021-01-22
Base: Move house on home-directory*.png icons to the left

linusg:
Otherwise it might be covered by the symlink overlay, which looks silly.

---

image image

image image


linusg:
ran `optipng -strip all` on all images to keep them at their previous tiny size, apparently GIMP adds a lot of garbage by default (confirmed they're still loading in Serenity ofc)

nico:
Maybe it could be in the upper right?

linusg:
@nico which of these would you prefer?

image image

Personally I think both of these look worse than bottom left/right...

Another option is just centered:

image

nico:
Hm, I guess they all need some getting used to. I retract my comment :)

PR #5059, opened 2021-01-23
Fix more wallpaper painting issues in highdpi

nico:


bcoles:
Confirmed this resolves #5017.

Issue #4135, opened 2020-11-22
SoundPlayer: Top of Open File window is obscured


danboid:
If you load SoundPlayer then choose its Open (File) menu option, the Open window is opened in a strange place just a little too high up in the top left corner so that you cannot see the window bar and a little bit below nor can you move the window by simply using the mouse.

supex0fan:
@awesomekling this is because `rect.center_within()` is trying to center the `FilePicker` within the much smaller `SoundPlayerWidget`. 

```
SoundPlayer(24): new_rect's current rect: [200,200 700x400]
SoundPlayer(24): parent_rect's current rect: [108,129 350x140]
SoundPlayer(24): new_rect.center_within(parent_window.rect() called
SoundPlayer(24): new_rect's new value: [-67,-1 700x400]

'x' value calculation:
(108 + 175) - (700 / 2)
283 - 350 = -67
```

How would you like this fixed? Easiest is to call `dialog.center_on_screen()` if the rect from `rect.center_within()` comes back negative, or you could add a `rect.center_over()` method which would center the rect over another rect as much as it can without placing it off screen.

bcoles:
> or you could add a `rect.center_over()` method which would center the rect over another rect as much as it can without placing it off screen.

If `rect.center_over()` can use the parent window then I think that's the best option.

This likely affects a lot of serenity applications - not only the SoundPlayer.

supex0fan:
An issue I have come across whilst implementing the `center_over()` method is that it is possible for the resulting rect to end up partially covered by the menubar or taskbar. This is easily solved assuming that the height of the menubar and taskbar is stored in a variable somewhere, but it would mean that I have to `#include` an (almost) entirely unrelated file into `Rect.h`, and write the necessary logic into a method that should really only be about centering one rect over another. It doesn't really sound right to me (with my minimal experience). 

On a related note, I've noticed that you can drag and drop a window under the menubar such that you can no longer drag it again as all of the draggable area is obscured by the menubar. In ubuntu for example, if you drag a window to the menubar, the top of the window will stick to the bottom of the menubar until the mouse cursor touches the bottom of the menubar in which case the window will fullscreen. So this issue might require a wider scoped change.

Issue #5052, opened 2021-01-22
WindowServer: New windows can be created outside the screen


BenWiederhake:
Steps to reproduce:
- Drag the Terminal to the left, but don't trigger the "left-tiles maximization" feature.
- Open settings via menu

Expected behavior: The settings dialog is now visible and accessible.

Actual behavior: The settings dialog is outside the screen, so even though it exists and is focused, it is neither visible nor accessible.

I intend to fix this and #4644 soon.

bcoles:
> I intend to fix this and #4644 soon.

Probably (?) also #4135.

BenWiederhake:
Ah, yes, that's the one.

tomuta:
FYI I also have changes in the works (almost complete) for #4910 that may fall into this area as well, which includes support for multiple monitors with any arbitrary sizes and arrangements, including the ability to find and align rectangles at another screen that is closest from the desired area.

BenWiederhake:
Is that a "dibs"? :)

So should I hold back or should I continue? (My fixes were very simple, I'm nearly done.)

tomuta:
Just proceed. FIFO ;-)

Issue #4644, opened 2020-12-30
It's possible to put windows off screen and therefore lose access to them


Lubrsi:
Demo:

https://user-images.githubusercontent.com/25595356/103326588-f1d06800-4a48-11eb-8936-fe2af00d713f.mp4

tomuta:
iirc Windows moves a window back into the screen lightly if it is (almost) entirely out of accessible screen estate. At least 10+ years ago. I think that would make sense for us as well.

diversys:
I can still reproduce it with Display Settings window by simply moving it to the very top.

PR #5060, opened 2021-01-23
WindowManager: Fix (un-)maximizing and (un-)tiling

BenWiederhake:
There were many things that could make a window inaccessible:
- A misbehaving program sending `CreateWindow`
- A perfectly upright program sending `CreateWindow` with `auto_position` set to true whose parent window is close to the border of the screen
- Unmaximizing (nearly always)
- Untiling (often, but not always)
- The user moving the window in an unfortunate way, so that the window is technically still accessible, but extremely hard to grab

Also, there was some funny logic around aspect ratio / tiling / maximization, where setting the aspect ratio did not cause the tiling/maximization to stop. And finally, when dragging a window upright, there was a bit of logic that was probably supposed to make the window stay maximized, but didn't work properly. This is fixed, too.

Closes issues #4135, #4644, #5052.

This also preserves a nice plug-in spot (`Window::normalize_rect`) that @tomuta probably needs for multi-monitor support :D

### DIFF
--- a/Userland/Libraries/LibGfx/Rect.cpp
+++ b/Userland/Libraries/LibGfx/Rect.cpp
@@ -141,6 +141,16 @@ void Rect<T>::align_within(const Rect<T>& other, TextAlignment alignment)
     }
 }
 
+template<typename T>
+void Rect<T>::set_size_around(const Size<T>& new_size, const Point<T>& fixed_point)
+{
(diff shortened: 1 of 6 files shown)


BenWiederhake:
Changes since the last push:
- Fixed an embarrassing logic error (windows juuust outside the desktop rect are meaningful – for example, the TaskBar, which is external. Also, the MenuBar if it were to call `normalize_rect`.)

PR #5066, opened 2021-01-23
Base+UE: Improvements around syscalls

BenWiederhake:
- UE and documentation are lacking many syscalls: Add a script that lets you more easily see which syscalls are missing where. Not part of `lint-ci.sh` because it's not an error, but we could change that later.
- The documentation for some LibC calls were in the man2 (syscalls) directory. Moved it.
- UE used to throw funny messages when the executable was not found (see screenshots). Replaced by proper error messages.
  before:
  Bildschirmfoto_2021-01-23_15-09-29
  after:
  Bildschirmfoto_2021-01-23_15-11-12
- Implement the `chown` syscall to demonstrate how easy it is. Also, a PR complaining about missing syscalls should implement at least one syscall :^)
  before:
  Bildschirmfoto_2021-01-23_15-19-59
  after:
  Bildschirmfoto_2021-01-23_15-46-46
- Demonstrate a few cornercases where UE fails to print a stacktrace (see #5067).


BenWiederhake:
Changes since last push:
- Remove evil BOM. I don't know who added that, could it be QtCreator? O.o

PR #5068, opened 2021-01-23
Deprecated dbg() and add kerneln().

asynts:
This should prevent anyone from adding new `dbg()` statements. In the meantime I can work on finishing up the debug macro stuff and removing `klog()` as well.

I've also added `kernelln()` which is the format version of `klog()`. I am open to suggestions if someone has better idea, here are a few names that I came up with:

  - `kernelln`, currently my favorite.
  - `klogln`, feels a bit overloaded to me especially if we look at `vklogln`.
  - `logln`, isn't clear enough.
  - `kernln`, abbreviating doesn't make the name better in my opinion.
  - `syslogln`, doesn't entirely match what it does.
  - `sysln`, abbreviating doesn't make the name better in my opinion.

### DIFF
--- a/AK/Debug.h
+++ b/AK/Debug.h
@@ -573,3 +573,15 @@ constexpr bool debug_lock_restore = true;
 #else
 constexpr bool debug_lock_restore = false;
 #endif
+
+#ifdef FUTEXQUEUE_DEBUG
+constexpr bool debug_futex_queue = true;
+#else
+constexpr bool debug_futex_queue = false;
+#endif
+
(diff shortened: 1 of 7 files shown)


awesomekling:
Perhaps the most accurate and directly understandable name would be `dmesgln()`?

asynts:
Personally, I don't like that name, but it seems to be pretty established in the Unix world and we do have a `dmesg` utility which reads the output of these messages.

I've changed it from `kernelln` to `dmesgln`.

PR #5069, opened 2021-01-23
Kernel: Add RTL8169 support

supercomputer7:
This is a basic driver for the known RTL8169 network chipset, written and tested with the help of PCI-passthrough a card to Serenity's VM and testing it.

What works:
* ping
* IRC

What doesn't work:
* Using the browser under heavy load of interrupts

### DIFF
--- a/Kernel/CMakeLists.txt
+++ b/Kernel/CMakeLists.txt
@@ -118,6 +118,7 @@ set(KERNEL_SOURCES
     Net/NetworkAdapter.cpp
     Net/NetworkTask.cpp
     Net/RTL8139NetworkAdapter.cpp
+    Net/RTL8169NetworkAdapter.cpp
     Net/Routing.cpp
     Net/Socket.cpp
     Net/TCPSocket.cpp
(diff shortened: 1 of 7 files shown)


supercomputer7:
@deoxxa If you can test this patch on your machine and report the results, it will be very appreciated! :)

deoxxa:
@supercomputer7 can do!

supercomputer7:
I tested this patch together with the ICH7 machine (and of course, with the card inserted into a PCI slot) and it seems to get stuck on initialization of the RTL8169, probably when doing the reset sequence.

supercomputer7:
Seems like after applying #6277 (rebasing on it), and trying to boot the kernel - it works but I can't access anything outside my local network.

supercomputer7:
I'll continue to work on this once I finish to work on #6277.

supercomputer7:
Let's reopen this, as I have more time to tackle this important patch now :)

supercomputer7:
Let's close this once again, next time I'll open this is when we have a working patch (and we're pretty close to it) :sweat_smile:

PR #5073, opened 2021-01-23
SMP: Implement thread priority queues

tomuta:
This is a continuation of #3864 and is based on #5100

For the purposes of implementing thread priority queues, I think this can be merged. I removed the commit to actually enable scheduling on all processors as there are at least two outstanding issues that prevent it from working properly:

- `KernelRng` needs some serious fixing as it almost instantly crashes the system, this is now tracked in #5132
- WindowServer hangs for a few seconds and resumes working normal, I suspect there is a race condition somewhere related to blockers.


tomuta:
Added a temporary `#define SCHEDULE_ON_ALL_PROCESSORS 0` to Scheduler.cpp so we can more easily turn it on. I find myself having to make these changes quite frequently.

PR #5079, opened 2021-01-23
CrashDaemon: Replace usleep() for 10ms with sleep() for 1s

bcoles:
The "Conforming To" section of the [usleep man page for Linux](https://www.man7.org/linux/man-pages/man3/usleep.3.html) suggests `usleep()` was removed from the 2008 POSIX standard and recommends using `nanosleep()` instead:

       4.3BSD, POSIX.1-2001.  POSIX.1-2001 declares this function
       obsolete; use nanosleep(2) instead.  POSIX.1-2008 removes the
       specification of usleep().

Compliance with POSIX in Serenity is generally desired, but not required, and it looks like the sleep funtcions all wrap the `sys$clock_nanosleep` syscall anyway:

* https://github.com/SerenityOS/serenity/pull/3337#discussion_r479940749
* fd29bed656ffdad79edd8d8a614195d4ae53ae58

However, this begs the question: does `CrashDaemon` really need to hammer the filesystem with `stat()` 100 times per second (every 10 milliseconds) in `wait_until_coredump_is_ready()` ?

Simply waiting for one full second seems reasonable.

### DIFF
--- a/Userland/Services/CrashDaemon/main.cpp
+++ b/Userland/Services/CrashDaemon/main.cpp
@@ -45,7 +45,7 @@ static void wait_until_coredump_is_ready(const String& coredump_path)
         if (statbuf.st_mode & 0400) // Check if readable
             break;
 
-        usleep(10000); // sleep for 10ms
+        sleep(1);
     }
 }
 


bcoles:
Perhaps `CrashDaemon` should also implement a timeout in `wait_until_coredump_is_ready()` rather than looping `while(true)` indefinitely. If the coredump is not "ready" (readable) within 10 minutes it is unlikely to be ready in 60 minutes, or 600 minutes, or 6000 minutes ...

linusg:
How about something like 100-200ms as a compromise (5-10 times / second)? 1s seems like it could noticeably increase the delay between crash and backtrace - e.g. if the crashdump is ready just after the first 1s sleep, it then takes another second for the backtrace to show up, instead of 1.2s.

alimpfard:
Am I missing something here? can you not just watch the directory?

bcoles:
> 
> 
> Am I missing something here? can you not just watch the directory?

Currently, the directory is watched in a `while(true)` loop:

https://github.com/SerenityOS/serenity/blob/397f432aed616aaf0b226095e630ba259c54f84d/Userland/Services/CrashDaemon/main.cpp#L92-L103

Which calls `wait_until_coredump_is_ready()` which then loops indefinitely waiting for the coredump to be "ready" (readable)"

https://github.com/SerenityOS/serenity/blob/397f432aed616aaf0b226095e630ba259c54f84d/Userland/Services/CrashDaemon/main.cpp#L37-L50

alimpfard:
by "watch" I meant the `watch_file` syscall, I'm pretty sure it sends notifications when the children of a directory change

bcoles:
> 
> 
> by "watch" I meant the `watch_file` syscall, I'm pretty sure it sends notifications when the children of a directory change

Right, but it doesn't send an event when something *doesn't change* (ie, the file is never "ready"). If we don't care about handling this case, then yes, simply watching the directory is sufficient.

alimpfard:
> ie, the file is never "ready"

You mean a timeout?
I'm sure we could have a timer next to it, `DirectoryWatcher` probably uses an event loop already

bcoles:
> 
> 
> > ie, the file is never "ready"
> 
> You mean a timeout?
> I'm sure we could have a timer next to it, `DirectoryWatcher` probably uses an event loop already

Yes a timeout seems like a good idea. https://github.com/SerenityOS/serenity/pull/5079#issuecomment-766186464

bcoles:
> How about something like 100-200ms as a compromise (5-10 times / second)?

Using a fractional sleep duration would require using `usleep()` or `nanosleep()`.

`nanosleep` is preferred over `usleep` due to unspecified behavior for `usleep` :

       The interaction of this function with the SIGALRM signal, and
       with other timer functions such as alarm(2), sleep(3),
       nanosleep(2), setitimer(2), timer_create(2), timer_delete(2),
       timer_getoverrun(2), timer_gettime(2), timer_settime(2),
       ualarm(3) is unspecified.

Serenity's sleep functions are identical (except for the return value) as they all wrap the `sys$clock_nanosleep` syscall, which makes this a non-issue (for now), beyond the surface level confusion when reading the code - "why is Serenity using a deprecated function?".

> 1s seems like it could noticeably increase the delay between crash and backtrace - e.g. if the crashdump is ready just after the first 1s sleep, it then takes another second for the backtrace to show up, instead of 1.2s.

Yes increased wait time is the price to be paid for increasing the sleep duration.

bcoles:
It seems that the preferred method for removing use of the deprecated `usleep()` function is to remove sleeping entirely and instead use a timeout on the directory watcher. Closing.

PR #5085, opened 2021-01-23
FileManager: Make DirectoryView open links in their real directory

DragonAlex98:
Previously it was possible to open a link like /home/anon/Desktop/Home,
leading to a folder with the same name. Now it correctly opens its real
path, which is /home/anon

This is my first time writing something in C++ and my first time contributing
to an open source project, I hope everything is fine with the pull request!

### DIFF
--- a/Userland/Applications/FileManager/DirectoryView.cpp
+++ b/Userland/Applications/FileManager/DirectoryView.cpp
@@ -29,6 +29,7 @@
 #include <AK/LexicalPath.h>
 #include <AK/NumberFormat.h>
 #include <AK/StringBuilder.h>
+#include <LibCore/File.h>
 #include <LibCore/MimeData.h>
 #include <LibCore/StandardPaths.h>
 #include <LibGUI/FileIconProvider.h>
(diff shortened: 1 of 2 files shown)


Userland/Applications/FileManager/DirectoryView.cpp
```diff
@@ -349,13 +349,19 @@ void DirectoryView::add_path_to_history(const StringView& path)
 
 void DirectoryView::open(const StringView& path)
 {
-    if (model().root_path() == path) {
+    String real_path;
+
+    char* buffer = realpath(path.characters_without_null_termination(), nullptr);
+    real_path = buffer;
+    free(buffer);
```
linusg, line 356: This is essentially a re-implementation of `Core::File::real_path_for()`, so let's use that :)

```suggestion
    auto real_path = Core::File::real_path_for(path);
```

DragonAlex98:
Done! I also used the same Core::File::real_path_for() function in the main.cpp file of FileManager

TheMorc:
You might also need to squash those two commits into one

PR #5089, opened 2021-01-24
Userland: Implement grep -v

ADKaster:
While doing so, make grep spin a little less on stdin, if that's where it's getting the input from.

### DIFF
--- a/Userland/Utilities/grep.cpp
+++ b/Userland/Utilities/grep.cpp
@@ -25,6 +25,7 @@
  */
 
 #include <AK/ByteBuffer.h>
+#include <AK/ScopeGuard.h>
 #include <AK/String.h>
 #include <AK/Utf8View.h>
 #include <AK/Vector.h>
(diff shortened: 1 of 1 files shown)


Userland/Utilities/grep.cpp
```diff
@@ -186,15 +187,19 @@ int main(int argc, char** argv)
     };
 
     if (!files.size() && !recursive) {
-        auto stdin_file = Core::File::standard_input();
-        while (!stdin_file->eof()) {
-            auto line = stdin_file->read_line();
```
alimpfard, line 191: I think the same bug exists in `Core::File::LineIterator` as well then (which I hope replaces the manual way of reading lines from a `Core::File` eventually)

seems like `IODevice::read_line()` should just block until it gets a newline instead of returning an empty string when it doesn't

ADKaster: It kinda feels weird that there's two completely separate implementations of buffered file I/O in the system. FILE* seems to be the more fleshed out one, at least recently. Shame it's got a crusty C API.

alimpfard: Yeah, I've tried making `IODevice` use a `FILE*` before, but it got rather boring to me so I gave up...

(just to clarify, I'm not saying these should be done in this PR)

ADKaster: Yeah it feels like the Core::File class works great for like, actual files that have a known beginning and end. When we start talking about streams it becomes less convincing.

If other parts of the system are actually using Core::File::standard_input (or Core::File::construct(STDIN_FILENO, ...) because the author didn't read the entire header file) it's likely they'll get surprising empty strings instead of full lines when using read_line.

ADKaster: I wrote up https://github.com/SerenityOS/serenity/issues/5093 to discuss this more. Seems like a bigger thing than grep -v for sure :)

PR #5094, opened 2021-01-24
Various minor tweaks to FontEditor

nico:


Userland/Applications/FontEditor/FontEditor.cpp
```diff
@@ -255,17 +255,17 @@ FontEditorWidget::FontEditorWidget(const String& path, RefPtr<Gfx::BitmapFont>&&
     fixed_width_checkbox.set_text("Fixed width");
     fixed_width_checkbox.set_checked(m_edited_font->is_fixed_width());
 
-    // Event hanglers
+    // Event handlers
```
awesomekling, line 258: >No behavior change

I'll take your word for it

PR #5096, opened 2021-01-24
Shell: Disable ENSURE_WAITID_ONCE

alimpfard:
Keeping it conditionally enabled seems to cause waitid to segfault for
unknown reasons.
```
 < CxByte> honestly...let's just remove ENSURE_WAITID_ONCE and pretend this weirdo behaviour never happened
 < andrewkaster> Problem for a few months from now, I like the way you think
 < CxByte> running just the preprocessor with the condition there, and the condition not there produces the exact same file
 < CxByte> so I declare spooky-action and will just remove that define.
```

If someone knows _why_ this is happening, please do tell!

### DIFF
--- a/Userland/Shell/Shell.cpp
+++ b/Userland/Shell/Shell.cpp
@@ -60,9 +60,11 @@ extern char** environ;
 
 namespace Shell {
 
+// Note: Conditionally enabling this with `ifdef __serenity__` produces some odd behaviour
+// https://github.com/SerenityOS/serenity/pull/4921#discussion_r563338358
+#if 0
 // FIXME: This should eventually be removed once we've established that
 //        waitpid() is not passed the same job twice.
-#ifdef __serenity__
 #    define ENSURE_WAITID_ONCE
 #endif
 


tomuta:
Does this mean that `waitid` or `waitpid` doesn't work as expected? I made a bunch of changes with the goal of implementing something similar to Linux, but it's possible that I missed something or something doesn't work quite as expected.

ADKaster:
When running without KVM, running 

```
system_tests=(test-js (test-crypto bigint -t))
usr_tests_tests=$(find /usr/Tests -type f | grep -v Kernel)
all_tests=($system_tests $usr_tests_tests)
```
Sometimes crashes the Shell in waitid, with a garbage backtrace...

ADKaster:
Definitely seems like something is funky -- with all of that code removed, I can seemingly reliably get it to crash EVERY time. But, if  I spam the command such that CPU usage spikes to 100 with crash reporter trying to dump 3 core dumps at once, it works. Race condition in shell?

ADKaster:
I can get the same crash by running all the tests in Shell/Tests on the system.

With this diff  https://pastebin.com/J6ZEgVv2

And running

```sh
for $(find /usr/Tests/Shell -type f) { if $it { echo "$it passed" } else { echo "$it failed" } }
```

It seems to crash in the control-structure-as-command.sh test every time?

alimpfard:
> Does this mean that waitid or waitpid doesn't work as expected?

maybe? maybe not? the crashes seem to happen only when `ENSURE_WAITPID_ONCE` is conditionally enabled behind `__serenity__`, it didn't crash with it always off, and it didn't crash with it always on.

> Race condition in shell?

Very likely, not sure where though

tomuta:
So the bug that causes this crash is caused when `Scheduler::enter_current` calls `Thread::dispatch_one_pending_signal`, which I believe causes `Thread::get_register_dump_from_stack` to return an invalid `RegisterState`. Thinking about it I think we need to track the "previous mode" (aka `%cs % 3`) and only call this if the previous mode is `!= 0` (meaning user mode).

alimpfard:
(likely) superceded by #5100, so I'll mark as draft until that one is merged

PR #5101, opened 2021-01-25
(WIP) Userland: Add an implementation of 'ed'

alimpfard:
We were missing a terminal-only editor, and 'ed' is simple enough to
implement.

---

A couple of commands aren't implemented yet, so this is still WIP :)


awesomekling:
This seems like a poor maintenance investment. `ed` is a terrible editor, even by late 1970s standards, and is unlikely to be used for anything other than "see, we have `ed`!".

I would strongly encourage making the effort to build a proper TUI editor instead. :)

alimpfard:
/me hides the fact that he uses `ed
the jump in complexity from a basic, simple line editor like `ed` to a TUI (even an extremely basic one) is quite drastic though

(I personally think it's a decent editor for quick and dirty edits, but alright)

Issue #5104, opened 2021-01-25
AK/Debug.h design means that any debug toggle needs a full rebuild


nico:
Apparently we put all debug toggles in AK/Debug.h. That means to toggle a debug macro you have these choices:

1. Pass a global define via a `-D` flag that affects AK/Debug.h. This requires rebuilding everything.
2. Locally edit AK/Debug.h to set some toggle to `true`. This requires rebuilding everything AK/Debug.h, which is almost everything.
3. In the cpp file that has the dbgln you want to change, add a `#define DEBUG_FOO` before including any header. That turns on debugging for just that file, but it has these disadvantages:
    - It's a bit questionable WRT ODR
    - It's non-obvious

Most `DEBUG_FOO` only affect a single file. This had better ergonomics with the pre AK/Debug.h design, and that old design also had much faster rebuilds.

@asynts

asynts:
I was thinking about splitting the debug defines into multiple files, one for global stuff `AK/Debug.h`, one for kernel stuff `Kernel/Debug.h` and one for userland stuff `Userland/Debug.h`. I believe this change should make the amount of stuff that needs to be rebuild manageable, while making it possible to change configuration options with CMake.

That's something that I still need to work out exactly.

asynts:
By the way, adding defines to CMake does not mean that everything needs to be rebuild. This is currently only the case because the `AK/Debug.h` file changes and thus many things have to be rebuild.

~~~none
$ ninja
[0/2] Re-checking globbed directories...
ninja: no work to do.
$ cmake .. -DTHIS_OPTIONS_ISNT_REFERENCED_ANYWHERE=ON
-- Configuring done
-- Generating done
CMake Warning:
  Manually-specified variables were not used by the project:

    THIS_OPTIONS_ISNT_REFERENCED_ANYWHERE


-- Build files have been written to: /home/me/dev/serenity/Build
$ ninja
[0/2] Re-checking globbed directories...
ninja: no work to do.
~~~

nico:
Why not have one header for each debug feature that occurs in more than one file, and no header at all for conditional dbgln()s that happen in just one file?

nico:
As of last night, step 3 is now:

- Move AK/Debug.h include to very top, and after it put `#undef FOO_DEBUG` `#define FOO_DEBUG 1`, even more non-obvious

tomuta:
> As of last night, step 3 is now:
> 
[...]

PR #5106, opened 2021-01-25
Meta: Split debug defines into multiple headers.

asynts:
This pull request splits the debug macros into two headers `AK/Debug.h` and `Kernel/Debug.h`. This makes it possible to toggle debug options without rebuilding almost everything.

Before if you toggled any debug option anywhere, you would have to rebuild 1925 files. Now, toggling a debug flag in `AK/Kernel.h` requires rebuilding 234 files. Sadly, toggling a debug flag in `AK/Debug.h` still requires rebuilding 1696 files. (I think I can cut this down a bit more by splitting again into `Userland/Debug.h` and `AK/Debug.h`, not sure how much of a benefit this would provide.)

This should improve the situation described in #5104.


asynts:
@nico Rebuilding 234 files for toggling a debug option seems reasonable to me, what do you think? (The other 1696 files are clearly not acceptable but I am working on that.)

nico:
234 is definitely much better than everything :)

It's still 233 files more than with the old approach, and I'm guessing as we add more files it'll become more than 234 files, while with the old approach it stayed 1 file.

Issue #2981, opened 2020-08-03
WindowServer: Occasional flickering


tomuta:
I noticed that sometimes windows from the background flicker, especially ones that draw frequently, especially Fire and Cube.

Steps to reproduce:
1. Open Fire (or Cube, but Fire seems to cause it more frequently)
2. Move some windows partially above that application, maybe even two windows
3. Observe that sometimes the entire Fire window will be drawn over the applications you moved above it

This really bothers me, but I don't really know where to look for in WindowServer. I suspect it's somewhere in `Compositor::compose`, maybe related to figuring out what rectangles to draw. But because it doesn't happen consistently, is it maybe related to `Compositor::flip_buffers` somehow where a buffer is flipped before the window stack is drawn entirely? But that doesn't make sense, either, because `Compositor::flip_buffers` isn't called until it's done drawing.

Thoughts?

awesomekling:
My hunch would be that we for some reason (incorrectly) believe that the dirty rect(s) don't intersect the window(s) above the flickering window.

tomuta:
I tried slowing down the compose timer but then it seemed like it never happened. Cut it in half (30fps) and caught it recording it as video:

Before:
![cap_flicker_00:00:45_01](https://user-images.githubusercontent.com/10320822/89224054-148f4b00-d595-11ea-8b89-abc6c91845f1.jpg)
Bad frame:
![cap_flicker_00:00:45_02](https://user-images.githubusercontent.com/10320822/89224084-1d801c80-d595-11ea-987c-b9dd6c9b6507.jpg)
After:
![cap_flicker_00:00:45_02](https://user-images.githubusercontent.com/10320822/89224101-2375fd80-d595-11ea-91a3-2a631f3e936d.jpg)

The pattern seems a little odd, doesn't seem like full rectangles (e.g. looking at the right outer frame of the Terminal window).

tomuta:
Another flicker:
Before:
![cap_flicker2_00:00:12_01](https://user-images.githubusercontent.com/10320822/89225015-c24f2980-d596-11ea-9836-76eaa55aa260.jpg)
Bad frame:
![cap_flicker2_00:00:12_02](https://user-images.githubusercontent.com/10320822/89225028-c7ac7400-d596-11ea-8ff9-5fa46d6575ba.jpg)
After:
[...]

PR #5107, opened 2021-01-25
WindowManager: Added a simple shadow effect

Cypher87:
I added a simple shadow effect for windows.


tomuta:
Ok so a couple thing here:

- We need to make sure the shadow is rendered to the temporary bitmap, otherwise it will cause excessive flickering when you have e.g. the Fire or Cube demo underneath. See #2994 and #2981 for some references as to why
- We **really** need to solve this a little bit differently: The Window frame needs to support transparency as well. This would allow us to render the shadow without having to account for it. Not only that, but it would allow arbitrary window frame shapes (e.g. rounded corners) depending on theme. The window frame then would have to be split into two bitmaps, one containing the top+bottom and one containing the left+right. And it would have to be treated similarly to the window content in regards to transparency: if it has transparency it would need to be rendered to the temporary bitmap.

Because of the reasons above I feel that we should not add all this logic at this point and instead focus on implementing window frame transparency properly. It will make it much more difficult to roll back these changes in order to implement it properly (at some point simply reverting this change to accomplish the above will not be easy due to conflicts and other related changes).

PR #5108, opened 2021-01-25
Documentation: Add instructions for network booting

boricj:
Here are some quick notes to make SerenityOS boot through PXE. While my setup works, it would probably be a good idea to check if this guide actually works before merging it.

### DIFF
--- a/Documentation/NetworkBoot.md
+++ b/Documentation/NetworkBoot.md
@@ -1,0 +1,102 @@
+## SerenityOS network booting
+
+By network booting, this guide implies a target computer (physical or virtual) that tries to boot over the network through PXE. The setup presented here is also diskless, as the bootloader and the root file system are supplied over the network.
+
+Note: it is recommended to boot a mainstream operating system through PXE on the target at least once before attempting this, if only to make sure that your setup works.
+
+### General notes
+
+This guide assumes several things:
+- The TFTP server root is `/srv/tftp/`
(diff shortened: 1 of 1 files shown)


Documentation/NetworkBoot.md
```diff
@@ -0,0 +1,102 @@
...
+### General notes
+
+This guide assumes several things:
+- The TFTP server root is `/srv/tftp/`
+- Bootloaders are located inside `/srv/tftp/boot/`
+- SerenityOS artefacts are located inside `/srv/tftp/serenity/`:
+    - The kernel is located at `/srv/tftp/serenity/kernel`
+        - You can find it at `Build/Kernel/Kernel`
+    - The ramdisk is located at `/srv/tftp/serenity/ramdisk`
+        - You can use the QEMU image at `Build/_disk_image_` as a ramdisk
```
TheMorc, line 16: isn't it only `_disk_image`?

danboid:
@deoxxa Does this get Serenity booting on your Athlon?

bcoles:
> it would probably be a good idea to check if this guide actually works before merging it.

TFTP+PXE setup instructions work for me on Ubuntu 20. Net booted VMware Player successfully.

VMware

TheMorc:
I tried a few machines and got it working currently only on Dell Studio XPS 1640
IMG_20210125_203117

Issue #5111, opened 2021-01-25
"window unresponsive" UI seems broken


nico:
Make some app unresponsive, e.g. by applying this patch, opening `fm` and opening columns view (ctrl-3):

```
diff --git a/Userland/Applications/FileManager/DirectoryView.cpp b/Userland/Applications/FileManager/DirectoryView.cpp
index 9e3595c5b..5671aa426 100644
--- a/Userland/Applications/FileManager/DirectoryView.cpp
+++ b/Userland/Applications/FileManager/DirectoryView.cpp
@@ -326,6 +326,7 @@ void DirectoryView::set_view_mode(ViewMode mode)
         return;
     }
     if (mode == ViewMode::Columns) {
+  sleep(40);
         set_active_widget(m_columns_view);
         return;
     }
```

Expected: "This application is unresponsive" title bar + greyed out window

Actual: Nothing changes

PR #5113, opened 2021-01-25
LibGfx: Implement scaling support for Painter::blit_filtered()

nico:
...and functions implemented in terms of it: blit_brightened(),
blit_dimmed(), blit_disabled().

In theory, this should stop the window server from asserting when
an application becomes unresponsive, but that feature seems to be
broken for unrelated reasons atm (#5111).


nico:
Failing macOS build says:

```
  Oops! A fatal internal error occurred.
  com.semmle.util.exception.CatastrophicError: We still have not reached idleness despite waiting for 30 minutes.
```

Has that graphql stuff ever found anything useful?

awesomekling:
> Failing macOS build says:
> 
> ```
>   Oops! A fatal internal error occurred.
>   com.semmle.util.exception.CatastrophicError: We still have not reached idleness despite waiting for 30 minutes.
> ```
> 
> Has that graphql stuff ever found anything useful?

I don't think so. All it does is crash occasionally. We should probably remove it.

PR #5118, opened 2021-01-26
EditingEngine: Fix move_to_previous_word not working on the last character in a file

supex0fan:
Code meant for the move_to_next_word functions which set the cursor to
the last character in the file if it was reached was copied into the
move_to_previous_word functions which lead them not moving when
the function was called from the end of the file.

### DIFF
--- a/Userland/Libraries/LibGUI/EditingEngine.cpp
+++ b/Userland/Libraries/LibGUI/EditingEngine.cpp
@@ -566,7 +566,7 @@ void EditingEngine::move_to_end_of_previous_word()
                 return;
             }
 
-            if (line_index == lines.size() - 1 && column_index == line.length() - 1) {
+            if (line_index == 0 && column_index == 0) {
                 m_editor->set_cursor({ line_index, column_index });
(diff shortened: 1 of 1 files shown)


Userland/Libraries/LibGUI/EditingEngine.cpp
```diff
@@ -616,7 +616,9 @@ void EditingEngine::move_to_beginning_of_previous_word()
 
             if (column_index == 0 && !is_first_iteration && (vim_isalnum(current_char) || vim_ispunct(current_char)))
                 return m_editor->set_cursor({ line_index, column_index });
-            else if (column_index == 0) {
+            else if (line_index == 0 && column_index == 0) {
+                return m_editor->set_cursor({ line_index, column_index });
+            } else if (column_index == 0 && is_first_iteration) {
                 is_first_iteration = false;
                 continue;
             }
```
awesomekling, line 624: Please don't mix braced and braceless in the same if/else block. If any of them have braces, they should all have braces :)

PR #5125, opened 2021-01-26
Swap confirm and cancel buttons on the file picker

nico:
And tweak a confirm dialog text.

Not sure if the button order is intentional, but it's inconsistent with the rest of the system as far as I can tell.


diversys:
I wonder what are the actual buttons here? OK/Cancel? If so wouldn't explicit wording be more helpful? As in Overwrite/Cancel.

nico:
@diversys Try it out :)

PR #5126, opened 2021-01-26
TextEditor: Add vim status indicators to the statusbar

supex0fan:
When using the VimEditingEngine in the TextEditor application, vim's
mode and the previous key are shown in the editor's status bar.

At the moment the UI doesn't look great because in the `GUI::StatusBar`, you can only set the number of labels in the constructor. And as far as I could see, you can't set constructor arguments in GML files. Correct me if I'm wrong there.

To get around this, I added a method to `GUI::StatusBar` that allows me to add labels after the constructor has been called, but then the resize corner wasn't updating with the new labels and I thought I'd better just stop there and ask what you think before changing more stuff.

### DIFF
--- a/Userland/Libraries/LibGUI/StatusBar.cpp
+++ b/Userland/Libraries/LibGUI/StatusBar.cpp
@@ -44,15 +44,15 @@ StatusBar::StatusBar(int label_count)
     layout()->set_margins({ 0, 0, 0, 0 });
     layout()->set_spacing(2);
 
-    if (label_count < 1)
-        label_count = 1;
+    if (label_count > 0) {
+        for (auto i = 0; i < label_count; i++)
+            m_labels.append(create_label());
 
-    for (auto i = 0; i < label_count; i++)
-        m_labels.append(create_label());
-
-    m_corner = add<ResizeCorner>();
+        m_corner = add<ResizeCorner>();
+    }
 
     REGISTER_STRING_PROPERTY("text", text, set_text);
+    REGISTER_INT_PROPERTY("label_count", label_count, set_label_count);
 }
 
 StatusBar::~StatusBar()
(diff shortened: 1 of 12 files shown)


awesomekling:
This commit touches both the `TextEditor` app and `LibGUI` so please indicate that in the commit message (or better yet, split it into separate commits if possible.)
Also, the application is called **TextEditor**, not **TextEditorApp** :)

Regarding the statusbar thing, I suppose we could default it to having no children, and then only auto-adding a label if you call the set_status_text API. I'm not completely sure what would be the best API here, but it would be nice if you could declare all the desired children in GML.

supex0fan:
The solution I went with is allowing GML files to set a label_count field which will create the set number of labels. The labels aren't created in the constructor (unless the constructor is called) as they previously were, but that doesn't cause any problems for the moment. I tried your other suggestion, creating labels when `set_text()` was called, but that felt really bad as you hinted it might.

That being said, explicitly setting the labels as children of the status bar in the GML is probably the best solution and worth revisiting.

Userland/Libraries/LibGUI/StatusBar.h
```diff
@@ -39,16 +39,23 @@ class StatusBar : public Widget {
...
     virtual void resize_event(ResizeEvent&) override;
 
 private:
     NonnullRefPtr<Label> create_label();
     NonnullRefPtrVector<Label> m_labels;
     RefPtr<ResizeCorner> m_corner;
+
+    // Used to initialize the number of labels that should
+    // be created from a GML file as opposed to the constructor.
+    int label_count() { return m_label_count; }
```
awesomekling, line 56: This should be const qualified.

Userland/Libraries/LibGUI/StatusBar.cpp
```diff
@@ -104,4 +104,18 @@ void StatusBar::resize_event(ResizeEvent& event)
     Widget::resize_event(event);
 }
 
+void StatusBar::set_label_count(int label_count)
```
awesomekling, line 107: This API is a bit awkward since Core::Object properties can be assigned multiple times, and assigning 5 to this twice would generate 10 labels.

awesomekling (changes requested):

Userland/Libraries/LibGUI/StatusBar.cpp
supex0fan: Yeah, I forgot to bring up the idea of only allowing the setter to create labels once. Does that sound like enough? Otherwise we could assert  `m_labels.is_empty()` or do something along those lines. What do you think?

awesomekling:
Reverted these changes in 31e04907b4ec0eadd12001694cc9961752e041ae and 775ae27a5579812a85957ac1154b6c2e9d2049b4 as they broke every app that uses **GUI::StatusBar**.

PR #5127, opened 2021-01-26
Kernel: Don't drop pledge()'d promises/execpromises when passing nullptr

linusg:
When passing `nullptr` for either `promises` or `execpromises` to `pledge()`, the expected behaviour is to not change their current value at all - we were accidentally resetting them to 0, effectively dropping previously `pledge()`'d promises.


awesomekling:
omg >_>

PR #5129, opened 2021-01-26
WindowManager: Make menus usable without a mouse

boricj:
I'm working on bare-metal without a mouse and SerenityOS is missing lots of keyboard shortcuts for essential features... Let's start with the most critical one.

### DIFF
--- a/Kernel/API/KeyCode.h
+++ b/Kernel/API/KeyCode.h
@@ -134,7 +134,8 @@
     __ENUMERATE_KEY_CODE(Pipe, "|")                  \
     __ENUMERATE_KEY_CODE(Tilde, "~")                 \
     __ENUMERATE_KEY_CODE(Backtick, "`")              \
-    __ENUMERATE_KEY_CODE(Logo, "Logo")
+    __ENUMERATE_KEY_CODE(Logo, "Logo")               \
+    __ENUMERATE_KEY_CODE(Menu, "Menu")
 
 enum KeyCode : u8 {
 #define __ENUMERATE_KEY_CODE(name, ui_name) Key_##name,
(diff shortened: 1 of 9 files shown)


awesomekling:
We had this in the past already, and it was extremely annoying since it kept opening the menu when you just wanted to move/resize windows.

boricj:
Normally, this should only pop up/close the menu with a straight press and release of the Logo key. Any event between those two will disengage it, so there should be no undesirable side-effects.

boricj:
I've revisited this PR and it now contains the following :
- Open the system menu with the Logo key ;
- Move around the top-level menus with the left/right key ;
- Default to the first enabled, non-separator menu item on key down ;
- Open context menu when 'Menu' key is pressed.

I believe this makes menus reasonably usable without a mouse. As far as I can tell, this doesn't introduce unwanted side-effects by itself.

awesomekling:
This too disruptive as-is. If I logo-drag a window around and then let go of logo, I don't want the menu to pop up.

boricj:
I *think* I've fixed it now, but I'm having trouble finding a window manager and X11/Wayland setup on my workstation that does not interfere when I attempt a logo-drag within QEMU's window, even with input grabbing... Hopefully it's corrected now.

tomuta:
> I _think_ I've fixed it now, but I'm having trouble finding a window manager and X11/Wayland setup on my workstation that does not interfere when I attempt a logo-drag within QEMU's window, even with input grabbing... Hopefully it's corrected now.

This works fine on stock Ubuntu 20.04. I really like this features so I'll give it a spin and report back in a few minutes

tomuta:
I can confirm this is working as intended now. I quite like it!

boricj:
I've also considered adding shortcuts like Logo+R to open up the "Run..." prompt and Logo+E to open up a file manager window, but these would require more in-depth modifications and that PR is big enough as it is.

One thing I do miss is something like KRunner on my KDE desktop to quickly launch something without going through a menu. The SerenityOS way would probably to add auto-complete to Run, but that's a story for another PR.

PR #5130, opened 2021-01-26
WindowServer: Vertically maximize window

### LINKED ISSUE #4123: Vertically maximize window
Under MATE I can vertically maximize a window by middle-clicking on the maximize window icon. I think we should steal this.

Camisul:
Button now can handle middle and right clicks.
Added 2 new handlers in button class: on_right_click for Right mouse
button and on_aux_click for middle mouse button.

Added functionality to vertically maximize window with middle mouse
click on the maximize window button

Closes #4123

### DIFF
--- a/Userland/Services/WindowServer/Button.cpp
+++ b/Userland/Services/WindowServer/Button.cpp
@@ -62,22 +62,36 @@ void Button::on_mouse_event(const MouseEvent& event)
 {
     auto& wm = WindowManager::the();
 
-    if (event.type() == Event::MouseDown && event.button() == MouseButton::Left) {
+    if (event.type() == Event::MouseDown && (event.button() == MouseButton::Left || event.button() == MouseButton::Right || event.button() == MouseButton::Middle)) {
         m_pressed = true;
         wm.set_cursor_tracking_button(this);
         m_frame.invalidate(m_relative_rect);
         return;
     }
 
-    if (event.type() == Event::MouseUp && event.button() == MouseButton::Left) {
+    if (event.type() == Event::MouseUp && (event.button() == MouseButton::Left || event.button() == MouseButton::Right || event.button() == MouseButton::Middle)) {
         if (wm.cursor_tracking_button() != this)
             return;
         wm.set_cursor_tracking_button(nullptr);
         bool old_pressed = m_pressed;
         m_pressed = false;
         if (rect().contains(event.position())) {
-            if (on_click)
-                on_click(*this);
+            switch (event.button()) {
+            case MouseButton::Left:
+                if (on_click)
+                    on_click(*this);
+                break;
+
+            case MouseButton::Right:
+                if (on_right_click)
+                    on_right_click(*this);
+                break;
+
+            default:
+                if (on_middle_click)
+                    on_middle_click(*this);
+                break;
+            }
         }
         if (old_pressed != m_pressed) {
             // Would like to compute:
--- a/Userland/Services/WindowServer/Button.h
+++ b/Userland/Services/WindowServer/Button.h
@@ -52,6 +52,8 @@ public:
     void on_mouse_event(const MouseEvent&);
 
     Function<void(Button&)> on_click;
+    Function<void(Button&)> on_right_click;
+    Function<void(Button&)> on_middle_click;
 
     bool is_visible() const { return m_visible; }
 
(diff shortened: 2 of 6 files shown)


nico:
Since we have this window edge snapping thing already, how about making it so that if you vertical-resize a window and hit the screen edge, it vertically maximizes then? Might be more discoverable than different click behaviors on buttons (which feels kinda early X11-y and weird).

Camisul:
> Since we have this window edge snapping thing already, how about making it so that if you vertical-resize a window and hit the screen edge, it vertically maximizes then? Might be more discoverable than different click behaviors on buttons (which feels kinda early X11-y and weird).

Yea, this sounds like a way better idea, I'll work on it.
Converting to draft for now.

Userland/Services/WindowServer/Button.h
```diff
@@ -52,6 +52,8 @@ class Button : public Weakable<Button> {
     void on_mouse_event(const MouseEvent&);
 
     Function<void(Button&)> on_click;
+    Function<void(Button&)> on_right_click;
+    Function<void(Button&)> on_aux_click;
```
awesomekling, line 56: Let's just call this `on_middle_click` as that names is used everywhere else.

Camisul: Done.

Userland/Services/WindowServer/Button.cpp
```diff
@@ -62,22 +62,36 @@ void Button::on_mouse_event(const MouseEvent& event)
...
-    if (event.type() == Event::MouseDown && event.button() == MouseButton::Left) {
+    if (event.type() == Event::MouseDown && event.button() > MouseButton::None && event.button() < MouseButton::Back) {
         m_pressed = true;
         wm.set_cursor_tracking_button(this);
         m_frame.invalidate(m_relative_rect);
         return;
     }
 
-    if (event.type() == Event::MouseUp && event.button() == MouseButton::Left) {
+    if (event.type() == Event::MouseUp && event.button() > MouseButton::None && event.button() < MouseButton::Back) {
```
awesomekling, line 72: Same here.

Userland/Services/WindowServer/Button.cpp
```diff
@@ -62,22 +62,36 @@ void Button::on_mouse_event(const MouseEvent& event)
 {
     auto& wm = WindowManager::the();
 
-    if (event.type() == Event::MouseDown && event.button() == MouseButton::Left) {
+    if (event.type() == Event::MouseDown && event.button() > MouseButton::None && event.button() < MouseButton::Back) {
```
awesomekling, line 65: This comparison feels really weird and depends on the enum values being in a certain order.
Let's instead test for each of the buttons we want explicitly (i.e Left, Right and Middle)

Userland/Services/WindowServer/WindowFrame.cpp
```diff
@@ -72,6 +72,10 @@ WindowFrame::WindowFrame(Window& window)
         auto button = make<Button>(*this, [this](auto&) {
             WindowManager::the().maximize_windows(m_window, !m_window.is_maximized());
         });
+        button->on_middle_click = [this](auto&) {
+            dbgln("on_middle_click");
```
awesomekling, line 76: Is this meant to be here?

awesomekling (changes requested):

Userland/Services/WindowServer/Button.cpp
```diff
@@ -62,22 +62,36 @@ void Button::on_mouse_event(const MouseEvent& event)
...
+            switch (event.button()) {
+            case MouseButton::Left:
+                if (on_click)
+                    on_click(*this);
+                break;
+
+            case MouseButton::Right:
+                if (on_right_click)
+                    on_right_click(*this);
+                break;
```
awesomekling, line 88: This is not used for anything?

Userland/Services/WindowServer/Button.cpp
```diff
@@ -62,22 +62,36 @@ void Button::on_mouse_event(const MouseEvent& event)
...
+
+            default:
+                if (on_middle_click)
+                    on_middle_click(*this);
+                break;
```
awesomekling, line 93: The commit message still mentions `on_aux_click`

awesomekling (changes requested):

Userland/Services/WindowServer/Button.cpp
Camisul: Yes, I thought adding only `on_middle_click` would be strange

PR #5136, opened 2021-01-27
HackStudio crashes when opening a new project or when adding a new file

masozzi:
When adding a new file in HackStudio, the program crashes.

This is caused by `FileSystemModel::row_count()` returning 0 if `node` is *not* a directory, causing the assertion to always fail.

### DIFF
--- a/Userland/DevTools/HackStudio/HackStudioWidget.cpp
+++ b/Userland/DevTools/HackStudio/HackStudioWidget.cpp
@@ -190,7 +190,6 @@ void HackStudioWidget::open_project(const String& root_path)
     ASSERT(m_project);
     if (m_project_tree_view) {
         m_project_tree_view->set_model(m_project->model());
-        m_project_tree_view->toggle_index(m_project_tree_view->model()->index(0, 0));
         m_project_tree_view->update();
     }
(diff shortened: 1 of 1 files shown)


awesomekling:
But FileSystemModel::row_count() *should* return 0 if the node is not a directory. This is just a symptom of something else.

masozzi:
I found another similira case when opening a new prject when HackStudio is already open.

The problem is the same, when calling `toggle_index()` the assertion fails.

I alreday solved the issue when opening a new project. I have simply removed `m_project_tree_view->toggle_index(m_project_tree_view->model()->index(0, 0));` in `HackStudioWidget::open_project()` because is redundant, since `m_project_tree_view->set_model()` already updates the indexes.

The other issue is here: `HackStudioWidget::create_new_action()`, and is also `m_project_tree_view->toggle_index(m_project_tree_view->model()->index(0, 0));`. Removing this line solves the issue, but i don't really understand why is there in the first place. Can this be removed, or is really needed?

awesomekling:
It's remains from the old times when we used to have project files instead of opening whole directories recursively. It's fine to remove it :)

PR #5137, opened 2021-01-27
Kernel: Make KernelRng thread-safe

### LINKED ISSUE #5132: Kernel: KernelRng not thread/smp safe and crashes with 4+ cores
The `KernelRng` implementation is neither safe to be called from interrupt handlers nor does it properly handle simultaneous access by multiple threads (on multiple processors). 

This can be easily reproduced on `master` by simply setting `smp 4` in Meta/run.sh. I remember the system booting just fine with 4 or even 8 cores, so this feels like a regression.

But it is also preventing me from making meaningful progress, e.g. with 4 cores and enabling scheduling on all processors (with #5073 by setting `SCHEDULE_ON_ALL_PROCESSORS` to `1` in Scheduler.cpp) odds are pretty much 100% that the system will crash in `KernelRng`. It is sometimes possible to boot it with 2 cores, but it may crash in there eventually, too.

The page faults probably happen somewhere here because `m_data_length` (and maybe others) are not atomic or implemented in a thread-safe manner:
https://github.com/SerenityOS/serenity/blob/7cf0c7cc0d74b84b0a2cc6f180914178cbb9ea20/Userland/Libraries/LibCrypto/Hash/SHA2.cpp#L97

This is an example of when it crashes trying to add entropy from an interrupt handler:
```
#0  Kernel::handle_crash () at ../Kernel/Arch/i386/CPU.cpp:166
#1  0xc011cdae in page_fault_handler () at ../Kernel/Arch/i386/CPU.cpp:319
#2  0xc0118c0f in page_fault_asm_entry ()
#3  0xc3551eec in ?? ()
#4  0xc011fe0c in Kernel::FortunaPRNG<Crypto::Cipher::AESCipher, Crypto::Hash::SHA256, 256>::add_random_event<Kernel::EntropySource::Event<unsigned char> > ()
    at .././Kernel/Random.h:87
#5  Kernel::EntropySource::add_random_event<unsigned char> ()
    at .././Kernel/Random.h:170
#6  handle_interrupt () at ../Kernel/Arch/i386/CPU.cpp:715
#7  0xc0118323 in interrupt_common_asm_entry ()
#8  0xc3551fac in ?? ()
```

Another example:
```
[#1 colonel(0:2)]: ASSERTION FAILED: data
../Kernel/WaitQueue.cpp:103 in Kernel::WaitQueue::wake_all()::<lambda(Kernel::Thread::Blocker&, void*, bool&)>
[#1 colonel(0:2)]: 0xc0119280  __assertion_failed(char const*, char const*, unsigned int, char const*) +0xe2
[#1 colonel(0:2)]: 0xc01d2086  Kernel::WaitQueue::wake_all() +0xc4
[#1 colonel(0:2)]: 0xc0184977  Kernel::KernelRng::wake_if_ready() +0x2d
[#1 colonel(0:2)]: 0xc011ffef  handle_interrupt +0xbd
[#1 colonel(0:2)]: 0xc0118323  interrupt_common_asm_entry +0x2b
[#1 colonel(0:2)]: 0xc0118ca7  exit_kernel_thread +0x0
```

tomuta:
This adds an optional argument to get_good_random_bytes that can be
used to only return randomness if it doesn't have to block.
Also add a SpinLock around using FortunaPRNG.

Fixes #5132

### DIFF
--- a/Kernel/Random.h
+++ b/Kernel/Random.h
@@ -55,8 +55,11 @@ public:
     {
     }
 
-    void get_random_bytes(u8* buffer, size_t n)
+    bool get_random_bytes(u8* buffer, size_t n)
     {
+        ScopedSpinLock lock(m_lock);
+        if (!is_ready())
+            return false;
         if (m_p0_len >= reseed_threshold) {
             this->reseed();
         }
@@ -75,6 +78,7 @@ public:
         // Extract a new key from the prng stream.
         Bytes key_span = m_key.bytes();
         cipher.key_stream(key_span, counter_span, &counter_span);
+        return true;
     }
 
     template<typename T>
@@ -94,8 +98,11 @@ public:
 
     [[nodiscard]] bool is_ready() const
     {
+        ASSERT(m_lock.is_locked());
         return is_seeded() || m_p0_len >= reseed_threshold;
     }
+
+    SpinLock<u8>& get_lock() { return m_lock; }
 
 private:
     void reseed()
@@ -121,6 +128,7 @@ private:
     size_t m_p0_len { 0 };
     ByteBuffer m_key;
     HashType m_pools[pool_count];
+    SpinLock<u8> m_lock;
 };
 
 class KernelRng : public Lockable<FortunaPRNG<Crypto::Cipher::AESCipher, Crypto::Hash::SHA256, 256>> {
@@ -133,6 +141,8 @@ public:
     void wait_for_entropy();
 
     void wake_if_ready();
+
+    SpinLock<u8>& get_lock() { return resource().get_lock(); }
 
 private:
     WaitQueue m_seed_queue;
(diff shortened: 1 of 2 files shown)


Kernel/Random.h
```diff
@@ -182,7 +194,7 @@ class EntropySource {
 //       The only difference is that get_fast_random is guaranteed not to block.
 
 void get_fast_random_bytes(u8*, size_t);
-void get_good_random_bytes(u8*, size_t);
+bool get_good_random_bytes(u8*, size_t, bool = true, bool = true);
```
awesomekling, line 197: Please give names to these arguments as it's super non-obvious what they are (and they are the same type, too!)

tomuta:
Not sure why the macos CI failed, maybe it's down or something.

PR #5138, opened 2021-01-28
WindowManager: Add option to maximize a window when it is dragged to the top of the screen

etaIneLp:


jacobmulquin:
Great idea. My only comment would be to move the configuration value into the [Input] section

etaIneLp:
I don't think it belongs in the "Input" group, it determines behavior of the window manager, not how it handles input. Maybe putting it in a "Behavior" group would be better?

jacobmulquin:
It's certainly more descriptive than Misc :^)

BenWiederhake:
Just for documentation: When I touched this code the last time, I was thinking about making the top of the screen two different zones. "Close to the top" would top-tile the window, and "top pixel" would maximize it. Then I disregarded the idea, because I didn't want to introduce behavior changes because [uhh](https://xkcd.com/1172/).

etaIneLp:
> It's certainly more descriptive than Misc :^)

I'll change it :^)

etaIneLp:
Changed & rebased

awesomekling:
Let's not add random behavioral settings to WindowServer, especially without GUI to toggle them.
If you feel strongly about this particular behavior, try finding a way to incorporate it as a nice default behavior instead. :)

PR #5144, opened 2021-01-28
Kernel: Retire SchedulerData and add Thread lookup table

tomuta:
This allows us to get rid of the thread lists in SchedulerData.
Also, instead of iterating over all threads to find a thread by id,
just use a lookup table. In the rare case of having to iterate over
all threads, just iterate the lookup table.

_This does change the output of `SCHEDULER_RUNNABLE_DEBUG` a bit for the worse, but I feel like it's still worth getting rid of this baggage._

### DIFF
--- a/Kernel/Thread.cpp
+++ b/Kernel/Thread.cpp
@@ -45,6 +45,14 @@
 
 namespace Kernel {
 
+SpinLock<u8> Thread::g_tid_map_lock;
+HashMap<ThreadID, Thread*>* Thread::g_tid_map;
+
+void Thread::initialize()
+{
+    g_tid_map = new HashMap<ThreadID, Thread*>();
+}
+
 Thread::Thread(NonnullRefPtr<Process> process)
     : m_process(move(process))
     , m_name(m_process->name())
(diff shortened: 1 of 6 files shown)


Kernel/Thread.cpp
```diff
@@ -45,6 +45,14 @@
 
 namespace Kernel {
 
+SpinLock<u8> Thread::g_tid_map_lock;
+HashMap<pid_t, Thread*>* Thread::g_tid_map;
```
awesomekling, line 49: Seems like we should use `ThreadID` instead of `pid_t` here.

tomuta: I tried that but I couldn't... No `hash` function it seems, though I don't quite understand the type magic for those...

tomuta: Maybe there's a way to implement a template `hash` function for those types?

tomuta: @BenWiederhake Can we add some templated `hash` function to `DistinctNumeric` so that we can use things like `ThreadID` as keys in `HashMap`?

tomuta: Tried to add something like this to DistinctNumeric.h but not sure what the syntax has to be like exactly:
```c++
template<typename T, typename... Args>
struct Traits<DistinctNumeric<T, Args...>> : public Traits<T> {};
```

alimpfard: I also took a look at this, the core problem is that a parameter pack declared with 'auto' will only accept NTTP's, and a parameter pack declared with 'typename' (or 'class') will only accept types.
taking an extra `typename X` at the end of that specialisation will somehow confuse the compiler, and make it unable to infer _any_ of the types in the template params, so that's out too.

The only way I can think of is to just move the last parameter of `DistinctNumeric` to the second parameter position - this will make reading the errors a little bit more difficult, but traits are more useful imho.

tomuta: @alimpfard thanks for looking into this! Do you mind if I cherry pick your change into this PR? That way @awesomekling can just merge this PR without merge conflicts (we had that issue yesterday when when I had one PR be based on another).

PR #5145, opened 2021-01-28
git: Tweak .gitignore file

nico:
- Push dirs to subdirectories. This is the main motivation -- to
  look at how ot fully rebuild the toolchain I always look in
  Toolchain/.gitignore first, and now that's enough
- Remove "Makefile" in a few places, no longer used
- Restore a comment that was lost in a commit that moved all
  gitignore content to the root gitignore file for some reason
- Push Meta/Lagom/build ignore to the local .gitignore file too

There's a few more things that look like they're unused
(eg all the CMakeCache.txt / CMakeFiles -- they should only be in the
Build dir), but let's leave that there so I have something to do for
next hacktoberfest.

### DIFF
--- a/Meta/Lagom/.gitignore
+++ b/Meta/Lagom/.gitignore
@@ -1,5 +1,5 @@
-liblagom.a
 CMakeCache.txt
 CMakeFiles
-Makefile
+build
 cmake_install.cmake
+liblagom.a
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,7 @@
+# Ignore all files without extension
 *
 !*.*
 !*/
-!Makefile
 !LICENSE
 !Base/**
 
@@ -22,13 +22,9 @@
 *.cflags
 *.cxxflags
 *.autosave
-Meta/Lagom/build
 Build
 build
 CMakeFiles
-Toolchain/Tarballs
-Toolchain/Build
-Toolchain/Local
 .vscode
 .ccls-cache
 .DS_Store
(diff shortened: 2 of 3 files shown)


Meta/Lagom/.gitignore
```diff
@@ -1,5 +1,5 @@
-liblagom.a
 CMakeCache.txt
 CMakeFiles
-Makefile
+build
```
linusg, line 3: Is that not covered by the top-level `.gitignore`'s `build`/`Build` already?

nico: Are `.gitignore` rules applied recursively? i.e. does `build` in toplevel .gitignore ignore ever/path/build?

linusg: yes :)

awesomekling: I don't understand the state of this PR, do you intend to make changes following this comment @nico?

awesomekling: Ping

nico: State is that I need to apparently read some docs on `.gitignore`. Like, if that's how it works, why do we have `Toolchain/Build` in toplevel .gitignore at all? Isn't that already redundant?

0xtrzy:
git not porting to serenity? Why?

Meta/Lagom/.gitignore
BenWiederhake: Yup, that was redundant :)

PR #5150, opened 2021-01-28
LibWeb: Flesh out prepare_script and execute_script

Lubrsi:
This fills in a bunch of the FIXMEs that was in prepare_script.

execute_script is almost finished, it's just missing the module side.

As an aside, let's not assert when inserting a script element with
innerHTML.


Userland/Libraries/LibWeb/HTML/HTMLScriptElement.cpp
```diff
@@ -56,16 +59,66 @@ void HTMLScriptElement::set_non_blocking(Badge<HTMLDocumentParser>, bool non_blo
...
+    if (m_from_an_external_file)
+        dispatch_event(DOM::Event::create(HTML::EventNames::load));
+}
+
+// https://mimesniff.spec.whatwg.org/#javascript-mime-type-essence-match
+static bool is_javascript_mime_type_essence_match(const String& mime_type)
+{
+    auto lowercase_mime_type = mime_type.to_lowercase();
+    return lowercase_mime_type.is_one_of("application/ecmascript", "application/javascript", "application/x-ecmascript", "application/x-javascript", "text/ecmascript", "text/javascript", "text/javascript1.0", "text/javascript1.1", "text/javascript1.2", "text/javascript1.3", "text/javascript1.4", "text/javascript1.5", "text/jscript", "text/livescript", "text/x-ecmascript", "text/x-javascript");
 }
```
linusg, line 112: This is not entirely correct, the "essence" of a mime type is:

> The essence of a MIME type mimeType is mimeType’s type, followed by U+002F (/), followed by mimeType’s subtype.

I.e. the parameters are stripped from the mime type - maybe worth a FIXME comment. Eventually we should move mime types from plain strings to an AK helper class like `URL` IMO (I did some work around this a while ago but never finished it), it's a lot nicer than operating on raw strings :)

Userland/Libraries/LibWeb/HTML/HTMLScriptElement.cpp
```diff
@@ -74,13 +127,40 @@ void HTMLScriptElement::prepare_script(Badge<HTMLDocumentParser>)
...
+        || (!has_type && has_language && attribute(HTML::AttributeNames::language).is_empty())
+        || (!has_type && !has_language)) {
+        script_block_type = "text/javascript";
+    } else if (has_type) {
+        script_block_type = attribute(HTML::AttributeNames::type).trim_whitespace();
+    } else if (!attribute(HTML::AttributeNames::language).is_empty()) {
+        StringBuilder builder;
+        builder.append("text/");
+        builder.append(attribute(HTML::AttributeNames::language));
+        script_block_type = builder.to_string();
```
linusg, line 153:
```suggestion
        script_block_type = String::formatted("text/{}", attribute(HTML::AttributeNames::language));
```

No need for a string builder here IMO.

Userland/Libraries/LibWeb/HTML/HTMLScriptElement.h
```diff
@@ -62,6 +62,9 @@ class HTMLScriptElement final : public HTMLElement {
     bool m_ready_to_be_parser_executed { false };
     bool m_failed_to_load { false };
 
+    // "false" is "classic". "true" is "module".
+    bool m_is_module { false };
```
linusg, line 66: Sounds like an enum :)

Userland/Libraries/LibWeb/HTML/HTMLScriptElement.cpp
Lubrsi: Yeah I agree, especially since there's a _[whole spec](https://mimesniff.spec.whatwg.org/)_ for it.

Issue #5149, opened 2021-01-28
Kernel: assertions fail size % PAGE_SIZE == 0 in VM/RangeAllocator triggered from unprivileged userspace


Jorropo:
# Backtrace :
```
[#0 jq(31:31)]: ASSERTION FAILED: (size % PAGE_SIZE) == 0
../Kernel/VM/RangeAllocator.cpp:98 in AK::Optional<Kernel::Range> Kernel::RangeAllocator::allocate_randomized(size_t, size_t)
[#0 jq(31:31)]: 0xc0119278  __assertion_failed(char const*, char const*, unsigned int, char const*) +0xe2
[#0 jq(31:31)]: 0xc01cda5b  Kernel::RangeAllocator::allocate_randomized(unsigned long, unsigned long) +0x8b
[#0 jq(31:31)]: 0xc019b242  Kernel::Process::sys$mmap(AK::Userspace<Kernel::Syscall::SC_mmap_params const*>) +0x24a
[#0 jq(31:31)]: 0xc0188876  syscall_handler +0x6eb
[#0 jq(31:31)]: 0xc0188183  syscall_asm_entry +0x31
```
# Steps to reproduce
- Get on the #5148 branch (if not merged yet)
- Build `jq` and reimage serenity
- Boot serenity
- input `jq` in the terminal (no need for proper json parsing, it crash earlier than that)
# Minimal reproducing code
Yet to be found (I'm new to this so I have some trouble with this :) ).

Jorropo:
I've made a tiny bit of progress, this issue bisects to : b6937e2560308379460a9f6d4a61bbf55e17dfde
And so seems to be related with ASLR.
(`jq` still doesn't work previous to b6937e2560308379460a9f6d4a61bbf55e17dfde tho, but it's a userland issue in the dynamic loader instead of a kernel panic.)

Jorropo:
Ohhhh, this is actually way easier than I first fought, it's just that `allocate_randomized` expect size to be sanitized but `mmap` don't do it before forwarding to `allocate_randomized` (which should be done I guess).

I'm not sure of what is the best approach, should `mmap` error if you don't require a whole number of page size ? Or should `mmap` give you back the size you asked rounded to the next page size ? (I would go for the second one, this seems way friendlier to me).

Jorropo:
Proposed fix: #5154

PR #5154, opened 2021-01-28
Kernel: sys$mmap PAGE_ROUND_UP size before calling allocate_randomized

Jorropo:
`allocate_randomized` assert an already sanitized size but `mmap` were just forwarding whatever the process asked so it was possible to trigger a kernel panic from an unpriviliged process just by asking some randomly placed memory and a size non alligned with the page size.
This fixes this issue by rounding up to the next page size before calling `allocate_randomized`.

This fixes: #5149


awesomekling:
Yup, great fix. No need for the signed-off-by btw :)

Jorropo:
> Yup, great fix. No need for the signed-off-by btw :)

Oh, we do it [at work](https://github.com/berty/berty) so it's in all my aliases.

Issue #5105, opened 2021-01-25
Windows can no longer be moved higher than top of screen


nico:
I used to very frequently move windows slightly off screen by super-dragging them up. It looks like it's now impossible to move a window higher than a y pos of 0. However, the window title bar still moves behind the system menu bar.

If the motivation is to prevent users from accidentally getting windows into a state where they can no longer move them back, they shouldn't be allowed behind the menu bar.

If the motivation is to make it impossible for windows to move completely off screen, they should be allowed to go higher than currently, just not completely off screen.

Maybe moving by super-drag should allow the larger range (since someone who knows that shortcut can get windows back as long as any pixels of a window are still draggable)?

This regression makes working on hidpi harder.

@BenWiederhake

BenWiederhake:
I forgot about super-dragging in Serenity, because that key is already assigned in my window manager to, uh, drag windows. I'll test around with super-drags this time around ^^

> the window title bar still moves behind the system menu bar.

The intention was to show "look, the window titlebar is slightly disappearing, you would get in trouble if I wouldn't stop you here", and stop the movement when the titlebar is just partially obscured but still grabbable.

> If the motivation is to prevent users from accidentally getting windows into a state where they can no longer move them back, they shouldn't be allowed behind the menu bar.

Does the titlebar vanish completely behind the menu bar? Weird, I tested precisely this, can you tell me the steps to reproduce this?

> Maybe moving by super-drag should allow the larger range (since someone who knows that shortcut can get windows back as long as any pixels of a window are still draggable)?

Sounds like a good idea! I'm afraid it'll take several days until I can do that.

> This regression makes working on hidpi harder.

I'm very sorry to hear that :/

Out of curiosity: What's the usecase of pushing a window through the top edge? And why does this interfere with work on hidpi?

nico:
> Does the titlebar vanish completely behind the menu bar? Weird, I tested precisely this, can you tell me the steps to reproduce this?

It does. Repro: Super-drag window as high as you can. It goes up to y=0, where the window's title bar is completely hidden.

> Out of curiosity: What's the usecase of pushing a window through the top edge? And why does this interfere with work on hidpi?

I do most testing in 640x480 (`chres 640 480 2`) and many windows don't completely fit the screen in that resolution, So I push them off the screen to the top to see their bottom bits. (Most of the time, `LibGfxScaleDemo`.)

BenWiederhake:
[...]

Issue #5063, opened 2021-01-23
WindowServer: Geometry label is on wrong layer


BenWiederhake:
![Bildschirmfoto_2021-01-22_21-29-12](https://user-images.githubusercontent.com/2690845/105567170-b287f500-5d30-11eb-84a1-1dea1c8d73e0.png)

This behavior is "correct", but a bit unexpected. Specifically, it feels like the TaskBar is squeezing itself between the window and the geometry label.

However, changing this will be non-trivial, as currently this label is drawn in `Compositor.cpp`, and I don't know how to make it appear in the correct layer.

(Note that this was already the case before #5060. Possibly related to #4910.)

nico:
Isn't this what you'd want? Where do you think the label should be? Below the task bar?

BenWiederhake:
Err, yes, I want it *below* the taskbar, like you said. And by "below" I mean "label is obscured by the taskbar". :)

Hmm, here's an entirely different observation: Dragging the window  off-screen far enough so that the center isn't visible anymore makes the label fall outside the screen.

How about we bound the label by the desktop? For example: The window starts in the middle of the screen, and is being dragged right. The geometry label stays centered on the window, *until* it would leave the screen. Then it stays there, even if the window is dragged further to the right.

tomuta:
Hm I'd expect it as-is... To me this is working as it should be in regards to the layer. But I agree, would make sense to be bound to the desktop.

PR #5156, opened 2021-01-28
WindowServer: Fix various annoying things

BenWiederhake:
In particular:
- Callers of `WindowManager::desktop_rect()` might have misbehaved during fullscreen (undocumented issue)
- Fix required visible region at the bottom (undocumented issue)
- Allow superdrags to hide the titlebar. #5105
- Fixes WindowServer crashing when a malicious program creates a window of size (0, 0) (undocumented issue)
- Constrain geometry label on move/resize to desktop #5063

I didn't know that other people were working on this, too D:

- Might conflict with PR #5130, if it fixes the issues around setting/reading `m_unmaximized_rect`
- Does *not* conflict with PR #5138
- Does *not* conflict with PR #5129

### DIFF
--- a/Userland/Services/WindowServer/WindowManager.cpp
+++ b/Userland/Services/WindowServer/WindowManager.cpp
@@ -554,8 +554,10 @@ bool WindowManager::process_ongoing_window_move(MouseEvent& event, Window*& hove
             } else if (m_move_window->tiled() == WindowTileType::None) {
                 Gfx::IntPoint pos = m_move_window_origin.translated(event.position() - m_move_origin);
                 m_move_window->set_position_without_repaint(pos);
-                // "Bounce back" the window it it would end up too far outside the screen:
-                m_move_window->normalize_rect();
+                // "Bounce back" the window if it would end up too far outside the screen.
+                // If the user has let go of Mod_Logo, maybe they didn't intentionally press it to begin with. Therefore, refuse to go into a state where knowledge about super-drags is necessary.
+                bool force_titlebar_visible = !(m_keyboard_modifiers & Mod_Logo);
+                m_move_window->normalize_rect(force_titlebar_visible);
             } else if (pixels_moved_from_start > 5) {
                 m_move_window->set_untiled(event.position());
                 m_move_origin = event.position();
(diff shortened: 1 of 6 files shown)


Userland/Services/WindowServer/WindowManager.cpp
```diff
@@ -1052,6 +1052,25 @@ Gfx::IntRect WindowManager::desktop_rect() const
     };
 }
 
+Gfx::IntRect WindowManager::arena_rect_for_type(WindowType type) const
```
nico, line 1056: s/arena/area/ (?)

nico: But even then I don't get what the name is supposed to mean. "screen_rect_for_window_type" (?)

BenWiederhake: That was not a typo: "Arena" as in "the space where you can move around". I'm not sure whether `screen_rect` is more intuitive than `arena`, but at least it looks less like a typo.

Userland/Services/WindowServer/WindowManager.cpp
```diff
@@ -554,8 +554,10 @@ bool WindowManager::process_ongoing_window_move(MouseEvent& event, Window*& hove
             } else if (m_move_window->tiled() == WindowTileType::None) {
                 Gfx::IntPoint pos = m_move_window_origin.translated(event.position() - m_move_origin);
                 m_move_window->set_position_without_repaint(pos);
-                // "Bounce back" the window it it would end up too far outside the screen:
-                m_move_window->normalize_rect();
+                // "Bounce back" the window it it would end up too far outside the screen.
```
emanuele6, line 557: > it it

BenWiederhake: Thanks!

BenWiederhake:
Changes since last push:
- Fixed a typo I did in FIXME, thanks to @emanuele6
- Fixed a typo in one of the commit messages themselves
- Rebased for convenience

I decided against @nico's suggestion to rename from "arena" to "screen_rect", as it does not refer to the entire screen, and the name is also not a typo.

PR #5157, opened 2021-01-28
Kernel: Make mount more resilient

linusg:


bugaevc:
Mounting on a file is not silly & a perfectly valid thing to do; it was intentionally supported. Why do you want to disallow it?

awesomekling:
Oh wait back up, I totally forgot that this is indeed supposed to be allowed >_>

linusg:
> supposed to be allowed

I know this isn't Linux, but that's definitely not a thing on Linux, nor on OpenBSD (if I understand their manpage correctly) - other systems as well, probably. It just seems unintuitive to me... Why would anyone want to do this, what was the reason for supporting it?

awesomekling:
It actually is a thing on Linux :)

You can bind-mount files anywhere you like with this method:
```
# touch foo
# mount -o bind /etc/passwd foo
```

linusg:
ohh, I wasn't thinking about bind mounts, that should indeed be possible. I still don't think this should be allowed, but whatever :)

```
# touch file
# mount /dev/hda file
```

awesomekling:
That we could disallow :)

bugaevc:
I don't think we should disallow that either. And whether this is a bind mount or not is orthogonal.

"Mounting" means replacing the given FS tree (inode) with another FS tree (another inode). You would typically mount onto empty files or empty directories, but in theory you should be able to mount onto anything. Then, in a non-bind mount, you're creating a new FS instance (from the supplied image) and use that as the FS tree to mount. In a bind mount you're reusing an already existing part of the FS tree.

PR #5158, opened 2021-01-28
Kernel: Prevent mmap-ing as both fixed and randomized

ccapitalK:
Just a small improvement I thought of while watching the video :^)


Jorropo (approved):
LGTM :)

Issue #2599, opened 2020-06-20
Kernel + Userland: Replace the DMIDecoder component with a raw blob in procfs


supercomputer7:
For now, we should remove the `DMIDecoder` code from the `Kernel/ACPI` folder. We used that component in the past to fast-determine if PCI is available, which did help because Bochs does set a flag of "I have PCI supported in the machine" in the SMBIOS tables. Currently we don't use this component anymore, because it's unreliable and I was mistaken when I used it in the kernel.
Later, we can just expose raw blobs of the SMBIOS tables and header in `/proc`, and use something like [`dmidecode`](https://github.com/mirror/dmidecode).

cc @awesomekling & @bugaevc

PR #5163, opened 2021-01-29
Ports: Add dmidecode

supercomputer7:
For this to work, I added two entries in `ProcFS`, and also added a new device called `MemoryDevice` (`/dev/mem`), which only allows currently to map the BIOS ROM between physical address `0xf0000` to `0x100000`.

### DIFF
--- a/Kernel/DMI.h
+++ b/Kernel/DMI.h
@@ -85,6 +85,8 @@ public:
     bool is_available() const { return m_available; }
     OwnPtr<KBuffer> entry_point() const;
     OwnPtr<KBuffer> structure_table() const;
+    size_t entry_point_length() const;
+    size_t structure_table_length() const;
 
 private:
     void set_64_bit_entry_initialization_values();
--- a/Kernel/DMI.h
+++ b/Kernel/DMI.h
@@ -1,0 +1,105 @@
+/*
+ * Copyright (c) 2021, Liav A. <liavalb@hotmail.co.il>
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#pragma once
+
+#include <AK/SinglyLinkedList.h>
+#include <AK/Types.h>
+#include <AK/Vector.h>
+#include <Kernel/KBuffer.h>
+#include <Kernel/PhysicalAddress.h>
+#include <Kernel/VM/Region.h>
+#include <Kernel/VirtualAddress.h>
+
+namespace Kernel::SMBIOS {
+
+struct [[gnu::packed]] LegacyEntryPoint32bit {
+    char legacy_sig[5];
+    u8 checksum2;
+    u16 smboios_table_length;
+    u32 smbios_table_ptr;
+    u16 smbios_tables_count;
+    u8 smbios_bcd_revision;
+};
+
+struct [[gnu::packed]] EntryPoint32bit {
+    char sig[4];
+    u8 checksum;
+    u8 length;
+    u8 major_version;
+    u8 minor_version;
+    u16 maximum_structure_size;
+    u8 implementation_revision;
+    char formatted_area[5];
+    LegacyEntryPoint32bit legacy_structure;
+};
+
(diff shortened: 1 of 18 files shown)


Ports/AvailablePorts.md
```diff
@@ -69,3 +69,4 @@ Please make sure to keep this list up to date when adding and updating ports. :^
 | [`vttest`](vttest/)            | vttest                                        | 20200610          | https://invisible-island.net/vttest/                  |
 | [`zlib`](zlib/)                | zlib                                          | 1.2.11            | https://www.zlib.net/                                 |
 | [`zstd`](zstd/)                | Zstandard                                     | 1.4.4             | https://facebook.github.io/zstd/                      |
+| [`dmidecode`](dmidecode/)                | dmidecode                                     | 3.1             | https://github.com/mirror/dmidecode                      |
```
linusg, line 72: Please keep this list sorted ;)

supercomputer7: Thanks for noticing :)

supercomputer7:
I made `dmidecode` to build with less patches than before, essentially now the patch files are semantic changes :)

supercomputer7:
It seems like `biosdecode` can use `/dev/mem` just fine, but `dmidecode` fails when using `/dev/mem`.

supercomputer7:
After couple of hours of debugging I finally figured it was a declaration of enum with wrong order of values... the `sysconf` syscall returned the wrong value, therefore a `mmap` called by `dmidecode` failed, what a journey!

supercomputer7:
I've added a better mechanism to determine if userspace is allowed to `mmap` physical memory with `/dev/mem` - in boot we will add all multiboot physical ranges to a Vector, then when trying to mmap, we will check if the `PhysicalAddress` is within a known range, and if so, allow to `mmap` for reserved ranges only. If we didn't find a range that corresponds to the checked address, we also allow to `mmap` it.

Also, this PR is related to https://github.com/SerenityOS/serenity/issues/2599, as I suggested to remove the older `DMIDecoder` component and expose the DMI blobs in `/proc/`.

Kernel/Devices/MemoryDevice.cpp
```diff
@@ -0,0 +1,83 @@
+/*
+ * Copyright (c) 2018-2020, Andreas Kling <kling@serenityos.org>
```
awesomekling, line 2: I didn't write this :)

supercomputer7: Oops, my bad!

awesomekling (changes requested):

Kernel/UnixTypes.h
```diff
@@ -60,8 +60,8 @@ enum {
     _SC_NPROCESSORS_CONF,
     _SC_NPROCESSORS_ONLN,
     _SC_OPEN_MAX,
-    _SC_PAGESIZE,
     _SC_TTY_NAME_MAX,
+    _SC_PAGESIZE,
```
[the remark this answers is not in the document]
supercomputer7: At least it works correctly now, `dmidecode` can `mmap` happily and I'm happy too :)

Kernel/DMI.h
```diff
@@ -0,0 +1,102 @@
...
+#include <AK/Types.h>
+#include <AK/Vector.h>
+#include <Kernel/KBuffer.h>
+#include <Kernel/PhysicalAddress.h>
+#include <Kernel/VM/Region.h>
+#include <Kernel/VirtualAddress.h>
+
+namespace Kernel {
+
+namespace SMBIOS {
```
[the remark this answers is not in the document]
supercomputer7: `DMIExpose` class is outside of this namespace on intent. But yeah, one empty line after opening a namespace is a good thing :)

Kernel/Devices/MemoryDevice.cpp
```diff
@@ -0,0 +1,83 @@
...
+{
+}
+
+MemoryDevice::~MemoryDevice()
+{
+}
+
+bool MemoryDevice::is_allowed_range(PhysicalAddress offset, const Range& range) const
+{
+    return MM.is_allowed_to_mmap_to_userspace(PhysicalAddress(offset.get())) && MM.is_allowed_to_mmap_to_userspace(PhysicalAddress(offset.get() + PAGE_ROUND_UP(range.size())));
```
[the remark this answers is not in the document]
supercomputer7: can we `mmap` with a range of not page size multiples? from what I understand we can't and therefore I round it with the macro.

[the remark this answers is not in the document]
awesomekling: The rounding is fine, but (base+size) is not the last byte of the range, it's one byte *after* the range.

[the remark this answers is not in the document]
supercomputer7: Oh I see, I was wrong about this, indeed it should take base+size so we don't have that "1 byte after" problem.

Ports/AvailablePorts.md
```diff
@@ -12,6 +12,7 @@ Please make sure to keep this list up to date when adding and updating ports. :^
 | [`curl`](curl/)                | curl                                          | 7.65.3            | https://curl.se/                                      |
 | [`dash`](dash/)                | DASH                                          | 0.5.10.2          | http://gondor.apana.org.au/~herbert/dash              |
 | [`diffutils`](diffutils/)      | GNU Diffutils                                 | 3.5               | https://www.gnu.org/software/diffutils/               |
+| [`dmidecode`](dmidecode/)      | dmidecode                                     | 3.3             | https://github.com/mirror/dmidecode                      |
```
bcoles, line 15: Fix table formatting alignment for consistency:

```suggestion
| [`dmidecode`](dmidecode/)      | dmidecode                                     | 3.3               | https://github.com/mirror/dmidecode                   |
```

Ports/dmidecode/package.sh
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env -S bash ../.port_include.sh
+port=dmidecode
+version=3.3
+useconfigure=false
+prefix=
+files="http://download.savannah.gnu.org/releases/dmidecode/dmidecode-${version}.tar.xz dmidecode-${version}.tar.xz"
```
bcoles, line 6: The archive file should be validated using the provided signature. The signature file (preferably both the archive and signature) should also be downloaded over HTTPS.

Untested, but this should do the trick:

```suggestion
files="https://download.savannah.gnu.org/releases/dmidecode/dmidecode-${version}.tar.xz dmidecode-${version}.tar.xz
https://download.savannah.gnu.org/releases/dmidecode/dmidecode-${version}.tar.xz.sig dmidecode-${version}.tar.xz.sig"
auth_type="sig"
auth_opts="--keyring ./gnu-keyring.gpg dmidecode-${version}.tar.xz.sig"
```

Ports/dmidecode/patches/dmiopt.h.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/dmiopt.h b/dmiopt.h
+index 2374637..d215aff 100644
+--- a/dmiopt.h
++++ b/dmiopt.h
+@@ -45,7 +45,7 @@ extern struct opt opt;
+ #define FLAG_QUIET              (1 << 3)
+ #define FLAG_DUMP_BIN           (1 << 4)
+ #define FLAG_FROM_DUMP          (1 << 5)
+-#define FLAG_NO_SYSFS           (1 << 6)
++#define FLAG_NO_PROCFS           (1 << 6)
```
bcoles, line 10:
```suggestion
+#define FLAG_NO_PROCFS          (1 << 6)
```

Ports/AvailablePorts.md
supercomputer7: Thanks for noticing! just asking - is there a tool to align these columns correctly so I can do that next time?

bcoles: None that I'm aware of.

PR #5166, opened 2021-01-29
Base: Make /www read-write by default

### LINKED ISSUE #5152: /www is read-only, even for root?
`mount` does not report `/www` as a mountpoint, but `touch /www/huh` fails with the error message `open: Read-only filesystem`. (Same with `cp` and `mv` or any other command.)

This makes it very difficult to extract files from Serenity :/

BenWiederhake:
Since it is owned by root anyway, there is no need for 'additional security' to prevent
modification of that directory. This makes it easier to quickly export files from Serenity.

Fixes #5152.

### DIFF
--- a/Base/etc/fstab
+++ b/Base/etc/fstab
@@ -7,6 +7,7 @@
 /home	/home	bind	bind,nodev,nosuid
 /root	/root	bind	bind,nodev,nosuid
 /var	/var	bind	bind,nodev,nosuid
+/www	/www	bind	bind,nodev,nosuid
 
 none	/proc	proc	nosuid
 none	/tmp	tmp	nodev,nosuid


alimpfard:
fwiw, `ws` takes a base directory argument.

off-topic (for this PR):
and now I'm reminded that it reads all the files into memory before sending them off, maybe we should stream that?

BenWiederhake:
Yes, but if we keep `/www` unusable even for root, then it seems better to just remove it entirely. Or to make it writable for root, like this PR suggests. :P

alimpfard:
that comment of mine was just an fyi, I think it's perfectly reasonable to mount /www as rw

awesomekling:
k sure

PR #5167, opened 2021-01-29
LibJS: Let RegExp.string get RegExp.prototype from the global object directly

linusg:
We can't assume that `RegExp` on the global object is still the original constructor, or an object at all.

This makes `--RegExp<</<</</,/</x/` work. :^)

Found by OSS-Fuzz: https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=29740


awesomekling:
> This makes `--RegExp<</<</</,/</x/` work. :^)

At last!

PR #5168, opened 2021-01-29
Kernel/Storage: Rewrite IDE disk detection and disk access

Lubrsi:
This replaces the current disk detection and disk access code with
code based on https://wiki.osdev.org/IDE

This allows the system to boot on VirtualBox with serial debugging
enabled and VMWare Player.

I believe there were several issues with the current code:
- It didn't utilise the last 8 bits of the LBA in 24-bit mode.
- {read,write}_sectors_with_dma was not setting the obsolete bits,
  which according to OSdev wiki aren't used but should be set.
- The PIO and DMA methods were using slightly different copy
  and pasted access code, which is now put into a single
  function called "ata_access"
- PIO mode doesn't work. This doesn't fix that and should
  be looked into in the future.
- The detection code was not checking for ATA/ATAPI.
- The detection code accidentally had cyls/heads/spt as 8-bit,
  when they're 16-bit.
- The capabilities of the device were not considered. This is now
  brought in and is currently used to check if the device supports
  LBA. If not, use CHS.

### DIFF
--- a/Kernel/Storage/IDEChannel.cpp
+++ b/Kernel/Storage/IDEChannel.cpp
@@ -100,9 +100,15 @@ namespace Kernel {
 #define ATA_REG_HDDEVSEL 0x06
 #define ATA_REG_COMMAND 0x07
 #define ATA_REG_STATUS 0x07
+#define ATA_REG_SECCOUNT1 0x08
+#define ATA_REG_LBA3 0x09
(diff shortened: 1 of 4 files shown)


awesomekling:
Oh this is great. It's amazing that the code worked at all :)

PR #5172, opened 2021-01-29
Meta: Expect sync-local.sh script at repository root

linusg:
This used to be in `Kernel/`, next to the `build-root-filesystem.sh` script, which was then moved to `Meta/` during the transition to CMake but has the working directory set to `Build/`, effectively expecting it there - which seems silly.

TL;DR: Very confusing. Use an explicit path relative to `SERENITY_ROOT` instead and update the `.gitignore` files.

---

cc @BenWiederhake, you mentioned this on IRC, so you probably use it & have an opinion :)

### DIFF
--- a/Kernel/.gitignore
+++ b/Kernel/.gitignore
@@ -1,4 +1,3 @@
 kernel.map
-sync-local.sh
 *.pcap
 eth_null*
--- a/Meta/build-root-filesystem.sh
+++ b/Meta/build-root-filesystem.sh
@@ -125,7 +125,22 @@ ln -s checksum mnt/bin/sha256sum
 ln -s checksum mnt/bin/sha512sum
 echo "done"
 
+if [ -f "${SERENITY_ROOT}/Kernel/sync-local.sh" ] || [ -f "${SERENITY_ROOT}/Build/sync-local.sh" ]; then
+    # TODO: Deprecated on 2021-01-30. In a few months, remove this 'if'.
+    tput setaf 1
+    echo
+    echo "   +-----------------------------------------------------------------------------+"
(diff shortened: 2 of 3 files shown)


Kernel/.gitignore
```diff
@@ -1,4 +1,3 @@
 kernel.map
-sync-local.sh
```
BenWiederhake, line 2: Maybe we could add a warning in the build script, and remove it in a few months?

Something like:
```sh
if [ -f "$SERENITY_ROOT"/Kernel/sync-local.sh ]; then
    # TODO: Deprecated on 2021-01-30. In a few months, remove this 'if'.
    echo "Warning: $SERENITY_ROOT/Kernel/sync-local.sh should move to $SERENITY_ROOT/sync-local.sh"
fi
```

or the likes.

linusg: Good call, done! (lots of red text :D)

Meta/build-root-filesystem.sh
```diff
@@ -125,7 +125,22 @@ ln -s checksum mnt/bin/sha256sum
...
+    echo "   |                                                                             |"
+    echo "   +-----------------------------------------------------------------------------+"
+    echo
+    tput sgr 0
+fi
+
 # Run local sync script, if it exists
-if [ -f sync-local.sh ]; then
-    sh sync-local.sh
+if [ -f "$SERENITY_ROOT}/sync-local.sh" ]; then
```
BenWiederhake, line 144: That opening curly brace looks suspiciously non-existing :D

linusg: ...oops. changed, tested, changed again, not tested again.

PR #5175, opened 2021-01-30
LibELF: Avoid quadratic memory usage weakness

BenWiederhake:
Section names are referred to by offset and length. We do not check
(and probably should not check) whether these names overlap in any way.
This opened the door to many sections (in this example: about 2700)
forcing ELF::Image::m_sections to contain endless copies of the same
huge string (in this case: 882K).

Fix this by loading only the first PAGE_SIZE bytes of each name.
Since section names are only relevant for relocations and debug
information and most section names are hard-coded (and far below 4096
bytes) anyway, this should be no restriction at all for 'normal'
executables.

Found by OSS-Fuzz:
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=29187


awesomekling:
Thank you so much for fixing fuzzer issues :D

PR #5176, opened 2021-01-30
WebServer: Stream the downloaded files / some other random tidbits

alimpfard:
Addresses my own comment at https://github.com/SerenityOS/serenity/pull/5166#issuecomment-769990034.

Some possibly dumb (but nice-sounding) ideas about streaming stuff
(cc @awesomekling and maybe @bugaevc?)

- Most of the time, the streaming is done from one fd to another
- Why can't the kernel handle this? instead of proxying down to userland and back to the kernel, just directly wire the write call of the source fd to the write handler of the target handler?
- For when it's not, we still have memfd (or rather, anon fds) - but this might be weird to do in the kernel, not all fds can be randomly accessed...

### DIFF
--- a/Userland/Libraries/LibProtocol/Download.cpp
+++ b/Userland/Libraries/LibProtocol/Download.cpp
@@ -57,7 +57,7 @@ void Download::stream_into(OutputStream& stream)
     };
 
     notifier->on_ready_to_read = [this, &stream, user_on_finish = move(user_on_finish)] {
-        constexpr size_t buffer_size = 1 * KiB;
+        constexpr size_t buffer_size = PAGE_SIZE;
         static char buf[buffer_size];
         auto nread = m_internal_stream_data->read_stream.read({ buf, buffer_size });
         if (!stream.write_or_error({ buf, nread })) {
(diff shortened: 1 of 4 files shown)


bugaevc:
> Why can't the kernel handle this? instead of proxying down to userland and back to the kernel, just directly wire the write call of the source fd to the write handler of the target handler?

Linux calls this `splice()`/`sendfile()`

alimpfard:
> Linux calls this splice()/sendfile()

Nice, somewhere to start! maybe it wasn't as dumb an idea as I thought it would be...

bugaevc:
> update

(Meta: please don't edit in new content into existing comments 😟 people are notified when you post the original comment, but not when you subsequently edit it. I only saw this new content by chance)

> what about a more permanent redirect-like thing?
e.g. we would just tell the kernel upfront that any data coming in to one fd is to be transferred to another fd, no need for reversal, and no need for intermediary userspace blocks?

No, I haven't heard of anything like that, and it sounds weird too. In that case, the userspace is expected to add this fd to the list of fds it watches over, and use splice or whatever to actually copy the data to the other fd.

alimpfard:
> please don't edit in new content into existing comments

sowwy~

> it sounds weird too

Figures. in my mind, it sorta makes sense, any writes to fd A would automagically be proxied to fd B without userland even knowing about it - I still like the weird idea tbh, but have no idea how it would be done :P

PR #5184, opened 2021-01-30
SMP related fixes

tomuta:
_This is very much work in progress and not necessarily meant to be merged at this point. These are a collection of fixes for various problems I observed trying to get closer to running on multiple cores. But some of these may have undesired side effects._

### DIFF
--- a/Kernel/Arch/i386/CPU.cpp
+++ b/Kernel/Arch/i386/CPU.cpp
@@ -1181,13 +1181,13 @@ UNMAP_AFTER_INIT void Processor::early_initialize(u32 cpu)
     m_in_irq = 0;
     m_in_critical = 0;
 
+    m_in_scheduler = false;
     m_invoke_scheduler_async = false;
     m_scheduler_initialized = false;
 
     m_message_queue = nullptr;
     m_idle_thread = nullptr;
     m_current_thread = nullptr;
-    m_scheduler_data = nullptr;
     m_mm_data = nullptr;
     m_info = nullptr;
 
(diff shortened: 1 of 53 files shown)


tomuta:
Looks like rewriting `Lock` as a `Thread::BlockCondition` helped with a lot of problems. The more cores the more threads would race, trying to grab the lock, including briefly blocking just to be unblocked and losing the race again. Well, with this new `Thread::BlockCondition` based approach those races are gone. Also, when a lock is released, ownership is transferred seamlessly to the next thread(s), and all they have to do is continue executing. Another benefit is much, much better logging. Also added (optional) detection for excessively long locks, in which case all owner as well as all those who are trying to acquire the lock are listed, including a backtrace for each. That should greatly help with figuring out (and noticing!) deadlocks.

There are still problems with the page fault handlers blocking, but with scheduling enabled on 4 cores it doesn't lock up nearly as often anymore. Unfortunately performance seems rather poor, especially with only one core. But that's also the case without this new `Lock` implementation, so probably not related. Not sure what the deal is, but I wonder if the problems with the page fault preemption are to blame.

tomuta:
And I just *completely* nuked `g_scheduler_lock`. So far the system seems to run just as well, and hopefully minus those deadlocking situations. Now a thread's `m_lock` is used for most things. But this also means that all the processors can switch contexts simultaneously.

There's still the intermittent hangs, so if anyone wants to try and figure out why that happens, your help is very much appreciated :)

Same for the intermittent `Lock` deadlocks in ext2 (this shows up with log messages and stack traces, once detected). I suspect it has to do with `m_paging_lock` and yielding from page fault handlers.

PR #5188, opened 2021-01-30
LibWeb: URL-encode/escape variables used in OOPWV's crash error page

linusg:
This fixes arbitrary HTML injections via the URL on OOPWV's crash error page - probably not a security issue, but annoying nonetheless.

| Before | After |
|--------|-------|
| image | image |


awesomekling:
Whoops 😄

PR #5191, opened 2021-01-30
Kernel: Add NE2000 network card driver

boricj:
Disclamer: this was seen as the path of least resistance for getting networking with SerenityOS on an Asus A7N8X-E Deluxe motherboard with what I have on hand. That was a colossal error in judgement, but what's done is done.

This has been tested on QEMU and it *seems* to work OK, but performance is downright horrible. That is not too surprising because this is an ancient, 16 bit PIO-only, 10 Mib/s Ethernet entry-level network card from the 80s. The current shortcomings of the SerenityOS network stack only compounds the issue.

~~I cannot test this driver with real cards because of unrelated issues (#5190). I do know however that the RAM test fails with real hardware.~~ Works fine on real hardware, performance is significantly better than on QEMU at least for packet reception.

This driver only supports PCI NE2000 cards. While I do have a couple of ISA NE2000 cards, SerenityOS is not about archeology.

Finally, a quote from "Using PCI NE2000 Cards with Linux":

> No document about this hardware can be written without an initial flame: PCI NE2000 clones are a bad idea. The original NE1000 and NE2000 boards were based around a minimal-component evaluation design described in the National Semiconductor LAN databook. That design was never intended to be mass-produced, it was only intended to demonstrate minimal functionality using the fewest, lowest cost external components. In doing so, performance and reliability features were omitted.

Remember, friends don't let friends use NE2000 network cards :^)


Kernel/Net/NE2000NetworkAdapter.cpp
```diff
@@ -0,0 +1,484 @@
+/*
+ * Copyright (c) 2018-2020, Andreas Kling <kling@serenityos.org>
```
drunderscore, line 2: you're _the_ Andreas Kling?
also let's not go back to 2020 :)

Meta/CMake/all_the_debug_macros.cmake
```diff
@@ -164,6 +165,7 @@ set(CPP_DEBUG ON)
 set(DEBUG_SPAM ON)
 set(DEBUG_CPP_LANGUAGE_SERVER ON)
 set(DEBUG_AUTOCOMPLETE ON)
+set(NE2000_DEBUG ON)
```
linusg, line 168: Duplicated, already set in line 149 :)

boricj:
So this works now on real hardware too (with #5196 as an additional fix).

Some benchmarks on an Athlon XP system with a RTL8029AS:
- HTTP downloads with `pro` starts at about 380 KiB/s, but it steadily drops to about ~250 KiB/s. The CPU is pegged at 100% during the download, but since the target has a wimpy Athlon XP CPU it's unclear what's the biggest bottleneck here.
- HTTP uploads with WebServer are pegged at ~11 KiB/s and CPU usage is ~40%. To be fair, the driver only has one transmission buffer and packet sending could be pipelined to speed this up.

It still works on QEMU, but download speed is around ~40 KiB/s. CPU usage is light, but the kernel continuously spits out `IPv4Socket(0xc08ae384): did_receive refusing packet since buffer is full.` about once per second. I haven't bothered checking upload sped.

For some reason, trying to ping stuff on bare metal ends badly, with processes becoming zombies. QEMU doesn't exhibit that, so I don't know where that could come from.

Overall, this driver should be in a good shape for a review. The biggest issue with it is that it cannot recover from a ring buffer overflow, but I haven't run into that problem so far. Hopefully that's good enough.

danboid:
Please add it to https://github.com/SerenityOS/serenity/blob/master/Documentation/HardwareCompatibility.md when it gets pushed.

Kernel/Net/NE2000NetworkAdapter.cpp
```diff
@@ -0,0 +1,487 @@
...
+}
+
+void NE2000NetworkAdapter::send_raw(ReadonlyBytes payload)
+{
+#if NE2000_DEBUG
+    klog() << "NE2000NetworkAdapter: Sending packet length=" << payload.size();
+#endif
+
+    if (payload.size() > NE2K_RAM_SEND_SIZE) {
+        klog() << "NE2000NetworkAdapter: Packet to send was too big; discarding";
```
awesomekling, line 401: Please do an early return here so we don't have to wrap most of the function in `else`.

Kernel/Net/NE2000NetworkAdapter.cpp
```diff
@@ -0,0 +1,487 @@
...
+#define BIT_INTERRUPTMASK_OVW (0b1 << 4)
+#define BIT_INTERRUPTMASK_CNT (0b1 << 5)
+#define BIT_INTERRUPTMASK_RDC (0b1 << 6)
+#define BIT_INTERRUPTMASK_RST (0b1 << 7)
+
+#define REG_RW_IOPORT 0x10
+
+// Page 1 registers
+#define REG_RW_PHYSICALADDRESS0 0x01
+#define REG_RW_CURRENT 0x07
```
awesomekling, line 129: How about we make all these things `static constexpr` instead of macros?
It seems nice to be explicit about the types used. :)

Kernel/Net/NE2000NetworkAdapter.cpp
```diff
@@ -0,0 +1,487 @@
...
+    });
+}
+
+NE2000NetworkAdapter::NE2000NetworkAdapter(PCI::Address address, u8 irq)
+    : PCI::Device(address, irq)
+    , m_io_base(PCI::get_BAR0(pci_address()) & ~3)
+{
+    set_interface_name("ne2k");
+
+    klog() << "NE2000: Found @ " << pci_address();
```
awesomekling, line 191: Please use `dmesgln()` instead of `klog()` in new code.

Kernel/Net/NE2000NetworkAdapter.cpp
```diff
@@ -0,0 +1,487 @@
...
+NE2000NetworkAdapter::~NE2000NetworkAdapter()
+{
+}
+
+void NE2000NetworkAdapter::handle_irq(const RegisterState&)
+{
+    u8 status = in8(REG_RW_INTERRUPTSTATUS);
+#if NE2000_DEBUG
+    klog() << "NE2000NetworkAdapter: Got interrupt, status=0x" << String::format("%02x", status);
+#endif
```
awesomekling, line 215: Instead of this construct, please do `dbgln<NE2000_DEBUG>(...)`
That way, the code will not bitrot :)

awesomekling (changes requested):
Cool :)

awesomekling:
> SerenityOS is not about archeology.

This is such a great sentence btw. 👍🏻

danboid:
T-shirt material.

PR #5192, opened 2021-01-30
Kernel+LibKeyboard+KeyboardMapper+keymap: All things keyboard!

BenWiederhake:
I tried to vaguely group the commits, but they're still pretty connected with each other.
- Kernel: new `getkeymap` syscall
- KeyboardMapper: Fix GUI and logic errors, start with current keymap by default (if no args given)
- keymap(1): fix unveil errors, better error reporting, support querying
- LibKeyboard: return error instead of asserting, support querying, don't force every application to carry the default keymap (which weighs 2.5 KiB)
- Actual keymaps: Fix some errors, extend `de.json` by a lot
- Meta: Add a script that checks the keymaps for consistency. Note that `jp.son` defines some exotic keycodes!
- Documentation :)


Documentation/BuildInstructions.md
```diff
@@ -169,3 +169,17 @@ Later on, when you `git pull` to get the latest changes, there's (usually) no ne
...
+#### Keymap
+
+Create a file with the exact name `sync-local.sh` in the project root (the same directory as `.clang-format`), with content like this:
+
+```
+#!/bin/bash
+
+set -e
+
+echo $'\n[keymap]\nExecutable=/bin/keymap\nArguments=de\nUser=anon' >> mnt/etc/SystemServer.ini
```
linusg, line 182: If you want increased readability:

```suggestion
cat <<EOT >> mnt/etc/SystemServer.ini

[keymap]
Executable=/bin/keymap
Arguments=de
User=anon
EOT
```

(this is what I use locally)

Documentation/BuildInstructions.md
```diff
@@ -169,3 +169,17 @@ Later on, when you `git pull` to get the latest changes, there's (usually) no ne
...
+```
+
+This will configure your keymap to German (`de`) instead of English. See `Base/res/keymaps/` for a full list.
```
linusg, line 185: Let's make it easier for the reader and link this:

```suggestion
This will configure your keymap to German (`de`) instead of English. See [`Base/res/keymaps/`](../Base/res/keymaps/) for a full list.
```

Documentation/BuildInstructions.md
```diff
@@ -169,3 +169,17 @@ Later on, when you `git pull` to get the latest changes, there's (usually) no ne
 
 #### Ports
 To add a package from the ports collection to Serenity, for example curl, go into `Ports/curl/` and run **./package.sh**. The sourcecode for the package will be downloaded and the package will be built. After that, run **make image** from the `Build/` directory to update the disk image. The next time you start Serenity with **make run**, `curl` will be available.
+
+#### Keymap
+
+Create a file with the exact name `sync-local.sh` in the project root (the same directory as `.clang-format`), with content like this:
+
+```
+#!/bin/bash
```
linusg, line 178: This is misleading as the interpreter is explicitly given here and the shebang therefore ignored:

https://github.com/SerenityOS/serenity/blob/ec57c6c83601d354e3103e4c668c3ef4a4cadf8f/Meta/build-root-filesystem.sh#L145

(`sh` may or may not be bash, we could make it say bash explicitly or just rely on the shebang)

Meta/lint-keymaps.py
```diff
@@ -0,0 +1,126 @@
...
+
+def report(filename, problem):
+    print('{}: {}'.format(filename, problem))
+
+
+def validate_single_map(filename, mapname, values):
+    all_good = True
+
+    if not isinstance(values, list):
+        report(filename, '"{}" is not a list'.format(mapname))
```
linusg, line 23: Let's use JSON terminology here:

```suggestion
        report(filename, '"{}" is not an array'.format(mapname))
```

Meta/lint-keymaps.py
```diff
@@ -0,0 +1,126 @@
...
+        return False  # Cannot continue other checks
+
+    if not any(values):
+        report(filename, 'no values set in {}'.format(mapname))
+        all_good = False
+
+    for i, c in enumerate(values):
+        if len(c) > 1:
+            report(filename, 'more than one character (“{}”) for charmap index {} of {}'.format(c, i, mapname))
```
linusg, line 32: Can we just use "normal" quotes here?

```suggestion
            report(filename, 'more than one character ("{}") for charmap index {} of {}'.format(c, i, mapname))
```

Meta/lint-keymaps.py
```diff
@@ -0,0 +1,126 @@
...
+        all_good = False
+
+    return all_good
+
+
+def validate_fullmap(filename, fullmap):
+    all_good = True
+
+    if not isinstance(fullmap, dict):
+        report(filename, 'is not a dict')
```
linusg, line 48: Same as above,

```suggestion
        report(filename, 'is not an object')
```

linusg (commented):
\\o/ for keymap linting!

alimpfard:
I was gonna ask, but why not stick a `keymap de` in your shellrc? the terminal runs at boot anyway?
the current way is definitely less hacky though~

Userland/Utilities/keymap.cpp
```diff
@@ -42,20 +42,45 @@ int main(int argc, char** argv)
...
+    }
 
-    Keyboard::CharacterMap character_map(path);
-    int rc = character_map.set_system_map();
-    if (rc != 0)
-        fprintf(stderr, "%s\n", strerror(-rc));
+    auto character_map = Keyboard::CharacterMap::load_from_file(path);
+    if (!character_map.has_value()) {
+        warnln("Cannot read keymap {}", path);
+        warnln("Hint: Must bei either a keymap name (e.g. 'en') or a full path.");
```
bcoles, line 76:
```suggestion
        warnln("Hint: Must be either a keymap name (e.g. 'en') or a full path.");
```

Documentation/BuildInstructions.md
emanuele6: you should use `<< 'EOT'` (with single quotes) to prevent variable substitution.

emanuele6:
I noticed that we are not being consistent with the style of keymap files.
Some of them only write non-ascii characters as backslash sequences (`it.json`, `fr.json`, ...) and some others write the non-ascii characters literally (`gb.json`, `de.json`, ...).

We should just pick one style and be consistent.

Documentation/BuildInstructions.md
BenWiederhake: Ah, yes! I like one-liners too much ^^'

Meta/lint-keymaps.py
BenWiederhake: But but but … *fancy*!

Yes, absolutely, I changed it :)

BenWiederhake:
> I was gonna ask, but why not stick a `keymap de` in your shellrc? the terminal runs at boot anyway?
> the current way is definitely less hacky though~

Hmm, we can't change it in `Base` because it's in version control. So we would need a command `sync-local.sh` that adds it to `.shellrc`. But that would be strictly worse than making the call *once* from SystemServer, because `.shellrc` will be executed multiple times, whenever the user opens a Terminal/Shell/Hack Studio/etc.

So for now, I'd like to stay with the SystemServer command.

Alternatively, we go the route that many system-utilities do: Make "the configuration file" a folder that gets executed alphabetically. This way we could tell git to ignore `Base/etc/SystemServer/*-local`, and the user only needs to create the file once. However, this is out of scope for this already-too-large-PR.

BenWiederhake:
> I noticed that we are not being consistent with the style of keymap files.
> Some of them only write non-ascii characters as backslash sequences (`it.json`, `fr.json`, ...) and some others write the non-ascii characters literally (`gb.json`, `de.json`, ...).
> 
> We should just pick one style and be consistent.

I agree! I suggest inserting the actual utf-8 things: It's easier to read, and for users who usually type it on their own keyboard, it's easier to write, too.

In that case, we would need to change `be`, `fr`, `it`, `trq`. Since both the lint script and `CharacterMapFile::load_from_file` do not see any difference, I'm too lazy to touch it, though.

BenWiederhake:
And because I forgot it when I created this PR: Please enjoy this tasteful screenshot.

Bildschirmfoto_2021-01-30_21-59-26

And in case anybody is interested: `KeyboardMapper` should also allow entering the Unicode codepoint somehow, because otherwise the only way to enter a "completely new" codepoint is to copy-paste it from somewhere else.

BenWiederhake:
Changes since first push:
- Fix silly python lint error
- Adjusted terminology in script
- More readable documentation
- Typos in the code and commit messages
- Rebased for convenience

BenWiederhake:
Changes since last push:
- Link in the documentation as linus suggested

emanuele6:
> I'm too lazy to touch it, though.

@BenWiederhake, you don't have to do it by hand~ :laughing:

```bash
#!/bin/bash

< <(grep -hoP -e '\\u[A-Za-z0-9]{4}' ./*.json) grep -i -ve '\\u001b' \
| sort -u \
| while read -r; do
    sed -i "s,\\$REPLY,$(eval "echo $'$REPLY'"),g" ./*.json
done
```
file

BenWiederhake:
Changes since last push:
- Unicode replacement through arcane incantation incantation (thanks @emanuele6!)
- Rebased for convenience

BenWiederhake:
Changes since last push:
- Fix	tabs	and	spaces	in	keymaps	by	just	spaces
- Rebased yet again

Kernel/Syscalls/keymap.cpp
```diff
@@ -66,4 +67,35 @@ int Process::sys$setkeymap(Userspace<const Syscall::SC_setkeymap_params*> user_p
     return 0;
 }
 
+int Process::sys$getkeymap(Userspace<const Syscall::SC_getkeymap_params*> user_params)
+{
+    // FIXME: Which pledge is appropriate here?
```
awesomekling, line 72: `stdio` seems fine

awesomekling: But a new `getkeymap` might be good too, since it's kind of a niche thing(?)

Userland/Utilities/keymap.cpp
```diff
@@ -49,13 +49,26 @@ int main(int argc, char** argv)
...
-    args_parser.add_positional_argument(path, "The mapping file to be used", "file");
+    args_parser.add_positional_argument(path, "The mapping file to be used", "file", Core::ArgsParser::Required::No);
     args_parser.parse(argc, argv);
 
+    dbgln("path is at {:p}, and contains {}", path, path);
+
+    if (!path) {
+        auto keymap = Keyboard::CharacterMap::fetch_system_map();
+        if (!keymap.has_value()) {
+            perror("getkeymap");
```
awesomekling, line 65: Please don't use `errno` to carry errors outside `LibC`. Instead, return a `Result<T, OSError>` or similar.

Documentation/BuildInstructions.md
```diff
@@ -169,3 +169,23 @@ Later on, when you `git pull` to get the latest changes, there's (usually) no ne
...
+cat << 'EOF' >> mnt/etc/SystemServer.ini
+
+[keymap]
+Executable=/bin/keymap
+Arguments=de
+User=anon
+EOF
+```
+
+This will configure your keymap to German (`de`) instead of English. See [`Base/res/keymaps/`](../Base/res/keymaps/) for a full list.
```
linusg, line 191:
```suggestion
This will configure your keymap to German (`de`) instead of US English. See [`Base/res/keymaps/`](../Base/res/keymaps/) for a full list.
```

Meta/lint-keymaps.py
```diff
@@ -0,0 +1,128 @@
...
+            if len(m) > index and m[index]:
+                c = m[index]
+        if c is None:
+            print('{}: None'.format(name))
+        else:
+            print('{}: "{}"'.format(name, c))
+
+
+if __name__ == '__main__':
+    os.chdir(os.path.dirname(__file__) + "/../Base/res/keymaps/")
```
linusg, line 121: No change required as far as I'm concerned, this is just a linting script... but you should try [`pathlib`](https://docs.python.org/3/library/pathlib.html) if you haven't yet :)

Documentation/BuildInstructions.md
emanuele6: we should also rename `en.json` to `us.json` or `en.json` to `enus.json` and `gb.json` to `engb.json` (for consistency with `ptpt` and `ptbr`), no?

linusg: Sounds good, yeah. I'd prefer `en-us`, `en-gb`, `pt-...` though.

emanuele6: agreed

Meta/lint-keymaps.py
BenWiederhake: It doesn't really help with going from `__file__` to the containing directory – at best it's another function. Simpler than addition is impossible (although I really like their `/` operator!), and `chdir` is a OS-level function, I don't really like to see it wrapped. So I'm sure `pathlib` is nice, but it solves a different problem :)

linusg: You shouldn't chdir in the first place, listdir (and the pathlib equivalent) works with explicitly given paths as well. Not here to argue though, if it works, it works

BenWiederhake:
Changes since last push:
- Call the default keyboard "US English"
- LibKeyboard no longer touches `errno`
- Remove resolved `//FIXME` (getkeymap(2) pledge)

Documentation/BuildInstructions.md
BenWiederhake: Okay, this is *definitely* out of scope for this PR. I mainly wanted to get the German layout complete and all tools working properly.

(Oh and: language-country codes are usually written like `pt_BR`, aren't they?)

emanuele6: @BenWiederhake: this is the full output of `locate keymaps` on my computer: https://gist.github.com/emanuele6/d89f513943fa26d67f0a8d1f6c6c294c

`pt-br` seems to be how is usually written. `pt_BR` looks like the name format of `locale` files.

BenWiederhake:
Changes since last push:
- Converted `sys$getkeymap` to be new pledge (new feature!)
- Added the pledge to keymap(1) and KeyboardMapper
- Convenience for rebased :P

Issue #4995, opened 2021-01-17
ProcFS: /proc/self/*: I/O error


bcoles:
`/proc/self/` started returning ` I/O error` some time in the past few months.

![/proc/self/](https://user-images.githubusercontent.com/434827/104855506-505e7700-5961-11eb-98fd-a22e1b0909ca.png)

bcoles:
Fixed in #5195

PR #5195, opened 2021-01-31
Kernel: Make /proc/self/ work again

BenWiederhake:
I have no idea when it broke.

Inspired by https://www.thanassis.space/bashheimer.html

### DIFF
--- a/Kernel/FileSystem/ProcFS.cpp
+++ b/Kernel/FileSystem/ProcFS.cpp
@@ -1449,7 +1449,11 @@ ssize_t ProcFSInode::write_bytes(off_t offset, ssize_t size, const UserOrKernelB
 
 KResultOr<NonnullRefPtr<Custody>> ProcFSInode::resolve_as_link(Custody& base, RefPtr<Custody>* out_parent, int options, int symlink_recursion_level) const
 {
-    // The only links are in pid directories, so it's safe to ignore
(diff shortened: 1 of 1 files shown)


Kernel/FileSystem/ProcFS.cpp
```diff
@@ -1449,7 +1449,13 @@ ssize_t ProcFSInode::write_bytes(off_t offset, ssize_t size, const UserOrKernelB
 
 KResultOr<NonnullRefPtr<Custody>> ProcFSInode::resolve_as_link(Custody& base, RefPtr<Custody>* out_parent, int options, int symlink_recursion_level) const
 {
-    // The only links are in pid directories, so it's safe to ignore
+    if (FI_Root_self == to_proc_file_type(identifier())) {
+        StringBuilder builder;
+        builder.appendf("%d", Process::current()->pid().value());
+        return VFS::the().resolve_path(builder.build(), base, out_parent, options, symlink_recursion_level);
```
awesomekling, line 1455: Instead of the `StringBuilder`:
```cpp
String::number(Process::current()->pid().value());
```

BenWiederhake:
Changes since last push:
- Use `String::number` for brevity
- Rebase for convenience

bcoles:
> I have no idea when it broke.

A few months ago. Confirmed this resolves #4995

PR #5197, opened 2021-01-31
Kernel: Gently torture the memory verification code

BenWiederhake:
This is a proof of concept (and can possibly evolve to become our own SerenitySyzkaller) for missing memory validation in the kernel. See #5198 for details.

### DIFF
--- a/Userland/Tests/Kernel/invalid-path-pointer-assert.cpp
+++ b/Userland/Tests/Kernel/invalid-path-pointer-assert.cpp
@@ -1,0 +1,56 @@
+/*
+ * Copyright (c) 2021, Ben Wiederhake <BenWiederhake.GitHub@gmx.de>
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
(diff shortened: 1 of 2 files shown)


Userland/Tests/Kernel/invalid-path-pointer-assert.cpp
```diff
@@ -0,0 +1,56 @@
...
+#include <errno.h>
+#include <stdio.h>
+#include <sys/stat.h>
+
+int main()
+{
+    struct stat statbuf;
+    // stat(3) would call strlen, and we can't have that.
+    Syscall::SC_stat_params params {
+        // Hey Kernel, please try to read the path from this totall valid location!
```
linusg, line 37:
```suggestion
        // Hey Kernel, please try to read the path from this totally valid location!
```

BenWiederhake:
Changes since last push:
- Fix typo – Gah! One of these days *someone* should implement an automatic spellchecker for the code changes.
- Rebased for convenience

Userland/Tests/Kernel/invalid-path-pointer-assert.cpp
BenWiederhake: Thanks!

PR #5200, opened 2021-01-31
AK: Make single pivot quick_sort guarantee a max stack depth of log(n)

Maato:
- The change to quick_sort requires SimpleIterator to support
assignment.

### DIFF
--- a/AK/QuickSort.h
+++ b/AK/QuickSort.h
@@ -86,35 +86,50 @@ void dual_pivot_quick_sort(Collection& col, int start, int end, LessThan less_th
 }
 
 template<typename Iterator, typename LessThan>
-void quick_sort(Iterator start, Iterator end, LessThan less_than)
+void single_pivot_quick_sort(Iterator start, Iterator end, LessThan less_than)
 {
-    int size = end - start;
-    if (size <= 1)
-        return;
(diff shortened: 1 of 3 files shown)


AK/QuickSort.h
```diff
@@ -88,27 +88,36 @@ void dual_pivot_quick_sort(Collection& col, int start, int end, LessThan less_th
...
-
-    if (pivot_point)
-        swap(*(start + pivot_point), *start);
+    for (;;) {
+        int size = end - start;
+        if (size <= 1)
+            return;
+
+        int pivot_point = size / 2;
+        auto pivot = *(start + pivot_point);
```
alimpfard, line 97: Pls no copying the pivot?
I thought I made a PR with that change a while back 🤔

Maato: That was already there, maybe you changed that in the dual_pivot_quick_sort?

alimpfard: Ooooh, you're right!
I missed the single pivot one, it seems ^^'

Well...please make that a forwarding reference while you're at it :)

Maato: Changed it, also included it in the test case.

PR #5205, opened 2021-01-31
Tests: Reorganize LibCompress unit tests

marcobiscaro2112:
Move LibCompress unit tests to LibCompress/Tests directory and register them with CMake's add_test. This allows us to run these tests with ninja test instead of running a separate executable.

Also split the existing tests in 3 test files that better follow the source code structure (inspired by AK tests).


alimpfard:
LibCrypto is already part of LagomCore IIRC

alimpfard:
Ah, only if `BUILD_LAGOM=1`.
if you're splitting it off, please remove it from `LAGOM_MORE_SOURCES` and link LagomCrypto where needed instead.

marcobiscaro2112:
Yes, it's part of Lagom (built wih ), but not LagomCore

>please remove it from LAGOM_MORE_SOURCES and link LagomCrypto where needed instead

On it

Userland/Libraries/LibCompress/Tests/TestDeflate.cpp
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2020, the SerenityOS developers.
+ * Copyright (c) 2021, the SerenityOS developers.
```
emanuele6, line 2:
```
 * Copyright (c) 2020-2021, the SerenityOS developers.
```

emanuele6 (changes requested):

marcobiscaro2112:
I think [LibCrypto](https://github.com/SerenityOS/serenity/blob/b48d8d1d6de5bcb0d2dc02590f5af3c0855dad9d/Userland/Utilities/test-crypto.cpp) tests would benefit of similar reorganization, moving the tests to LibCrypto/Tests and splitting the cases by component being tested (follow roughly the same structure as its source code). 

WDYT?

alimpfard:
`test-crypto` might be a bit special in that its tests are also expected to be run (and pass) inside serenity, and that it has its own ideas about what tests to run and how, so not sure if splitting it up has any particular value

marcobiscaro2112:
Oh, I see...

I think the value would be maintainability, because with < 10 crypto algorithms implemented `test-crypto.cpp` has 2700+ lines (and I think it also would be nice if all library unit tests follow the same pattern).

Just out of curiosity, do we have any automated tests that run inside serenity? Or these tests are only run locally on demand?

asynts:
*(Just dropping this here, since this might be relevant...)*

  > [...] (and I think it also would be nice if all library unit tests follow the same pattern).

I had a similar vision but quickly gave up on it because many of the tests are not really suited for it. The "unit" tests in `test-crypto` take too long to execute to be run frequently, and many other tests require are not really fully "automated". (At least the last time that I checked.)

marcobiscaro2112:
> tests in test-crypto take too long to execute to be run frequently

Yes, I noticed that too, but with it being a single executable it's difficult to run them in parallel (something we already have with ctest) and it's not immediately evident (in CI or with ninja test) which cases are taking too long (running test-crypto manually suggests RSA tests are the culprits)

ADKaster:
> Just out of curiosity, do we have any automated tests that run inside serenity? Or these tests are only run locally on demand?

https://github.com/SerenityOS/serenity/pull/4921

;)

marcobiscaro2112:
Trying to summarize, I think there are 3 topics we discussed here:

1. We must be able to run tests both inside serenity and in the host (using Lagom)
2. The tests must run fast
3. Ideally, the tests should be easy to maintain

About 1: we have #4921 that should enable us to run AK/Tests in both environments.

About 2: the existing tests ran with ctest (and also LibCompress tests that this PR reorganizes) are pretty fast already. The LibCrypto tests (`test-crypto` binary) are way too slow to run frequently (more about them below)

About 3: is what this PR does to LibCompress and what I suggested to do with LibCrypto, but we should probably solve 1 and 2 first.

Does this make sense?

--

About LibCrypto slow tests:

The tests are now slow per-se, but rather the library code is slow in some cases. Doing an early investigation, the main cause are functions from `Crypto::NumberTheory`.

`is_probably_prime` depends on a lot of [random number generation](https://github.com/SerenityOS/serenity/blob/a6ed034f89380a887c5c653e2b3f4336e25336b3/Userland/Libraries/LibCrypto/NumberTheory/ModularFunctions.cpp#L331-L332), which is not very efficient. It seems that a quick win here would be to decrease the number of iterations (maybe 16 is already enough, 1 in 4 billion error in worst case IIUC).

`ModularPower` [implementation is too naive](https://github.com/SerenityOS/serenity/blob/a6ed034f89380a887c5c653e2b3f4336e25336b3/Userland/Libraries/LibCrypto/NumberTheory/ModularFunctions.cpp#L137-L153) to work with big modulus. I tested some cases and it takes more than 7 seconds to calculate what the JVM BigInteger implementation takes ~120us.

alimpfard:
If you're looking for a cause, our bigint multiplication algorithm is particularly slow, and that's frequently used to calculate modexp
The modexp algorithm itself is fairly ok (it's O(lgn)) but with each op taking so long...

linusg:
@marcobiscaro2112 this PR has various merge conflicts, please rebase :)

marcobiscaro2112:
@linusg done!

Meta/Lagom/CMakeLists.txt
```diff
@@ -200,6 +191,17 @@ if (BUILD_LAGOM)
                 WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
             )
         endforeach()
+
+        foreach(source ${LIBCOMPRESS_TESTS})
+            get_filename_component(name ${source} NAME_WE)
+            add_executable(${name}_lagom ${source} ${LIBCOMPRESS_SOURCES})
+            target_link_libraries(${name}_lagom Lagom)
```
linusg, line 198: The AK and LibRegex tests above only link with LagomCore, would that be sufficient here as well?

marcobiscaro2112: We need full Lagom here because some LibCompress functions depend on LibCrypto. 

If we try to link only using LagomCore: 
```
/usr/bin/ld: serenity/Build/i686/../../Userland/Libraries/LibCompress/Gzip.cpp:241: undefined reference to `Crypto::Checksum::CRC32::digest()'
```

linusg: Thanks for clarifying!

PR #5213, opened 2021-02-01
HackStudio: New project dialog + Project templates

nvella:
~Decided to get this up as a draft to get some running feedback as I work on it, commits aren't squashed yet~ All features done, rebased onto master and tested working - ready for a proper review :smiley: 

### Summary
New Project dialog for HackStudio which reads simple templates from `/res/devel/templates`. Templates consist of an INI file (template manifest,) at the moment containing just name and description texts, a directory containing the template contents, and optionally a post-creation script which is ran at create-time with the project name and path as arguments. Four templates are included; empty, C++ basic (command-line), C++ GUI (a simple Hello World example), and C++ shared library. Icons are stored in `/res/icons/hackstudio/templates-32x32` and share the same name as the template manifest INI and content directory.

image

### What's done
- Dialog design and implementation
- `ProjectTemplate` class and `ProjectTemplateManager` singleton class
- `ProjectTemplate::create_project` method, where the magic happens.
- Empty, basic C++ command-line and C++ GUI templates
- C++ shared library template
- Automatically opening project on creation

### What I'm not entirely sure about
Still getting the hang of the general concept of smart pointers, I come from the land of managed memory (.NET) so go easy on me. :sweat_smile: There's probably a few mistakes around my pointer choices, object allocation methods and so on - constructive criticism is very much appreciated :)


Base/res/devel/templates/cpp-basic.ini
```diff
@@ -0,0 +1,4 @@
+[HSTemplate]
+Name=Command-line Application (C++)
+Description=Template for creating a basic C++ command-line application.
+Priority=95
```
linusg, line 4: Missing newline at end of file - please configure your editor to add one. :)
(affects many files in this PR)

Base/res/devel/templates/cpp-basic.ini
```diff
@@ -0,0 +1,4 @@
+[HSTemplate]
```
linusg, line 1: Perhaps we could just name this `HackStudioTemplate`? Doesn't seem *too* long to me, and less ambiguous.

Base/res/devel/templates/cpp-gui/main.cpp
```diff
@@ -0,0 +1,26 @@
...
+{
+    auto app = GUI::Application::construct(argc, argv);
+
+    auto window = GUI::Window::construct();
+    window->set_title("Hello friends!");
+    window->resize(200, 100);
+
+    auto button = GUI::Button::construct();
+    button->set_text("Click me!");
+    button->on_click = [&]() {
```
linusg, line 17: Does this actually work? I think this needs to be:

```suggestion
    button->on_click = [&](auto) {
```

As `on_click` will receive modifiers:

https://github.com/SerenityOS/serenity/blob/13d7c09125f8eec703d0a43a9a87fc8aa08f7319/Userland/Libraries/LibGUI/Button.h#L50

Userland/DevTools/HackStudio/CMakeLists.txt
```diff
@@ -11,6 +13,9 @@ set(SOURCES
     Debugger/DisassemblyWidget.cpp
     Debugger/RegistersModel.cpp
     Debugger/VariablesModel.cpp
+    Dialogs/NewProjectDialog.cpp
+    Dialogs/NewProjectDialog.gml
```
linusg, line 17: This should be

```suggestion
    Dialogs/NewProjectDialogGML.h
```

I think...

e.g. in Browser:

https://github.com/SerenityOS/serenity/blob/dc28c07fa526841e05e16161c74a6c23984f1dd5/Userland/Applications/Browser/CMakeLists.txt#L1-L16

Userland/DevTools/HackStudio/ProjectTemplate.cpp
```diff
@@ -0,0 +1,276 @@
...
+    auto bitmap_path_16 = String::formatted("/res/icons/hackstudio/templates-16x16/{}.png", id);
+
+    if (Core::File::exists(bitmap_path_32)) {
+        auto bitmap32 = Gfx::Bitmap::load_from_file(bitmap_path_32);
+        auto bitmap16 = Core::File::exists(bitmap_path_16)
+            ? Gfx::Bitmap::load_from_file(bitmap_path_16)
+            : Gfx::Bitmap::load_from_file("/res/icons/16x16/filetype-executable.png");
+
+        icon = GUI::Icon(move(bitmap16), move(bitmap32));
+    }
```
linusg, line 87: - Could we make this explicit and add the (32px) icon to the config file as well?
- Do we really need to attempt loading the 16px icon, that doesn't actually exist and is never shown in the widget anyway, as AFAICT?

linusg (commented):
*not a thorough review, just some feedback*

Base/res/devel/templates/cpp-gui/main.cpp
nvella: I have tested and interestingly it does, I'll make this change either way in the interest of the templates serving as good examples for developers :)

Base/res/devel/templates/cpp-basic.ini
nvella: I plan on running everything through clang-format before I squash and mark the PR as ready 😄

Userland/DevTools/HackStudio/ProjectTemplate.cpp
nvella: Scrapping the 16x loading is definitely a good idea. I included it initially in the interest of completeness, but in hindsight simpler would definitely be better here.

nvella:
Feedback has been applied @linusg :) I've also since added a rudimentary shared library template, and we're also now automatically opening the project after successful creation.

Userland/DevTools/HackStudio/ProjectTemplateManager.h
```diff
@@ -0,0 +1,48 @@
...
+
+namespace HackStudio {
+
+class ProjectTemplateManager {
+public:
+    ProjectTemplateManager();
+
+    static ProjectTemplateManager& the();
+
+    void rescan_templates();
```
Dexesttp, line 40: If I'm not mistaken, this is only called once (when creating the `ProjectTemplateManager` singleton), and then never again until HackStudio is closed.

Ideally, it would be nice to watch both the `/res/devel/templates/` directory for additions/deletions, and each identified `.ini` file into it for changes. However, after looking into it, these kind of utilities aren't really available in `Core` :
* The DirectoryWatcher is blocking and probably wouldn't play nice with the EventLoop. It seems like it was more created for [CrashDaemon](https://github.com/SerenityOS/serenity/blob/master/Userland/Services/CrashDaemon/main.cpp#L92) than for general purpose, and will need to be adapted for this use-case.
* There isn't any good wrapper for watching a file. The `watch_file` syscall is manually used for the GUI in [FileSystemModel.cpp line 142](https://github.com/SerenityOS/serenity/blob/master/Userland/Libraries/LibGUI/FileSystemModel.cpp#L142) via a Notifier, but not wrapped in any Core class.

Creating these utilities look very out of scope for this PR, though, and shouldn't block it on that.

It would however be nice to at least have one "refresh templates" button somewhere in the template UI (or maybe in a menu).

Userland/DevTools/HackStudio/ProjectTemplate.cpp
```diff
@@ -0,0 +1,279 @@
...
+    // Otherwise, create an empty directory at the project path.
+    if (Core::File::exists(content_path()) && Core::File::is_directory(content_path())) {
+        copy_directory(content_path(), path, false);
+    } else {
+        dbgln("No template content directory found for '{}', creating an empty directory for the project.", m_id);
+        int rc;
+        if ((rc = mkdir(path.characters(), 0777)) < 0) {
+            warnln("Failed to mkdir empty project directory, error: {}, rc: {}", strerror(errno), rc);
+            return false;
+        }
```
Dexesttp, line 106: Could `File::ensure_parent_directories()` work here ? Maybe if you absolutely want to create it as `777` instead of `755`, but I don't really see the reason to do that specifically in the rest of the code (and `777` looks wrong for a subdirectory of home).

Userland/DevTools/HackStudio/ProjectTemplate.cpp
```diff
@@ -0,0 +1,279 @@
...
+{
+    // Check if a file or directory already exists at the project path
+    if (Core::File::exists(path))
+        return false;
+
+    dbgln("Creating project at path '{}' with name '{}'", path, name);
+
+    // Verify that the template content directory exists. If it does, copy it's contents.
+    // Otherwise, create an empty directory at the project path.
+    if (Core::File::exists(content_path()) && Core::File::is_directory(content_path())) {
```
Dexesttp, line 98: I think failing `Core::File::is_directory` is going to be a hard error in any case ? `mkdir` will fail anyways with `EEXIST` and the method will return false correctly, but that seems like it could be misleading down the line.

Userland/DevTools/HackStudio/Dialogs/NewProjectDialog.cpp
```diff
@@ -0,0 +1,243 @@
...
+        return;
+    }
+
+    if (project_template->create_project(maybe_project_name.value(), maybe_project_full_path.value())) {
+        // Succesfully created, attempt to open the new project
+        m_created_project_path = maybe_project_full_path.value();
+        done(ExecResult::ExecOK);
+    } else {
+        GUI::MessageBox::show_error(this, "Could not create project: internal error. Please check your parameters and try again.");
+    }
```
Dexesttp, line 240: How about making `project_template` return an `AK::Result<String, String>` instead, with the first being the created project full path (for easier unwrapping below), and the second being a more detailed error string ? "Internal Error" is not as descriptive as we could be, and it seems like `create_project` could return a better message pretty easily.

Userland/DevTools/HackStudio/Dialogs/NewProjectDialog.cpp
```diff
@@ -0,0 +1,243 @@
...
+
+    auto maybe_project_path = get_project_full_path();
+
+    if (maybe_project_path.has_value()) {
+        m_full_path_label->set_text(maybe_project_path.value());
+    } else {
+        m_full_path_label->set_text("Invalid name or creation directory.");
+        m_input_valid = false;
+    }
```
Dexesttp, line 162: Same as for the project template, having the project path be an `Option<String>` instead of a `Result<String, String>` limits the message we can show to the user ("Invalid" isn't as useful as we could be, given that the `get_project_full_path()` method can fail in 6 different ways both by itself or in `get_available_project_name()`).

Dexesttp (commented):
I really like this - very good feature for both newcomers and quick project creation, using the shell for postcreate scripts seem like it's going to be very versatile, and the UI is sleek !

Just a few feedbacks, I don't think any of them are blocking, though :)

Userland/Libraries/LibGUI/IconView.h
```diff
@@ -53,6 +53,9 @@ class IconView : public AbstractView {
 
     Gfx::IntSize effective_item_size() const { return m_effective_item_size; }
 
+    bool always_wrap() const { return m_always_wrap; }
+    void set_always_wrap(bool value) { m_always_wrap = value; }
```
awesomekling: Let's be a little extra verbose here for clarity.. `always_wrap_item_labels` or `always_wrap_item_names`, something like that?

awesomekling:
Very cool work! Love this feature. :star:
Please fix conflicts, and also remove the merge commit from the PR :)

nvella:
> Very cool work! Love this feature. :star:
> 
> Please fix conflicts, and also remove the merge commit from the PR :)

No worries, and thanks for the review 😄 I'm usually quite busy during the week, I should be able to get around to all of it Friday evening (upside-down time 😆)

Userland/Libraries/LibGUI/IconView.h
nvella: I actually did fix this one up in a later commit (I think I went for `always_wrap_item_labels`), but accidentally squashed it into the main feature commit. I'll try and separate it out and squash it back into the LibGUI commit.

Userland/DevTools/HackStudio/ProjectTemplate.cpp
nvella: It seems `ensure_parent_directories` wouldn't work in this situation, it appears to only creates the parent directories in a path, but not the trailing directory itself. `755` is definitely the more correct permission, I'll make sure to change it :)

Userland/DevTools/HackStudio/Dialogs/NewProjectDialog.cpp
nvella: I may very well be missing something obvious, but it seems it's impossible to specify an error value for a `Result` with the same value and error type, which stops us from passing through the unwrapped project full path. It is possible to construct a `Result<void, String>` however, still allowing us to pass through an error string. It doesn't seem to be much functionally differently from a `Optional<string>`, but it seems like a more explicit way to pass an error than just using an `Optional`.

Userland/DevTools/HackStudio/Dialogs/NewProjectDialog.cpp
nvella: Referring to the previous conversation, it seems I'll have to pass on this for now until I find out how to specify an error value for `Result`s with the same value and error type 😄 Shouldn't be too much of a problem for now.

Dexesttp: Wow, that's right ! I hadn't even considered the possibility thiat it couldn't be done, and that seems like an unfortunate limitation of the `Result` type.
I don't have any issues with skipping it for this PR. It's just a small usability thing that can be improved later once the `Result<T, T>` thing is possible!

Dexesttp: An alternative would be to have a :
```cpp
struct TemplateCreationError {
    TemplateCreationError(String message)
        : message(message)
    {
    }
    String message;
}
```
and use it as the error type. This way, you could write `return TemplateCreationError("My message");`

nvella: Hmm... I've managed to pass along the template creation error with a `Result<void, String>` so far, do you think it would be more correct defining a specific error struct for this circumstance, even though only a String is being passed at the moment?

nvella: If you hold on just a few minutes I'll push my new changes up (unsquashed) :)

Userland/DevTools/HackStudio/ProjectTemplateManager.h
nvella: Consolidated the `ProjectTemplateManager` into the `ProjectTemplatesModel` for now; after a bit of a think I've deduced it to really only be added complexity, all template creation logic is in `ProjectTemplate` anyway. This makes using the new `FileWatcher` a lot easier, which I've gone ahead and done :)

nvella:
Just pushed my commits addressing the reviews :) Not yet squashed so it's a bit easier to distil my new changes, but otherwise will go ahead once all is OK 😄

Userland/DevTools/HackStudio/Dialogs/NewProjectDialog.cpp
Dexesttp: I don't actually know if creating a specific struct for errors is good practice in C++.
@awesomekling thoughts about using `Result<void, String>` vs `Result<void, TemplateCreationError>`, where `TemplateCreationError` is a wrapper struct around a string ?

PR #5216, opened 2021-02-01
Ports: Move packages.db from Ports/ to Build/

linusg:
When removing and recreating the Build directory, it's quite annoying having to edit/remove `Ports/packages.db` as the installer won't install previously installed port dependencies again if they're still listed.
This problem is easily solved by just considering `packages.db` a build-specific file.


awesomekling:
Well that's obviously a good idea!

PR #5222, opened 2021-02-01
Kernel: Make devices manage their own DevFS node

etaIneLp:


BenWiederhake:
I'm very sorry, please rebase against the new `master`.

etaIneLp:
> I'm very sorry, please rebase against the new `master`.

Rebased

Kernel/FileSystem/DevFS.h
```diff
@@ -113,7 +113,7 @@ class DevFSDeviceInode : public DevFSInode {
     virtual ssize_t write_bytes(off_t, ssize_t, const UserOrKernelBuffer& buffer, FileDescription*) override;
     virtual KResult chown(uid_t, gid_t) override;
 
-    NonnullRefPtr<Device> m_attached_device;
+    const Device* m_attached_device;
```
BenWiederhake, line 116: - Can `m_attached_device` now become nullptr? I don't think so, but in that case all accesses would now require checks; for example, `read_bytes` would crash otherwise.
- If it's guaranteed non-null, it should be a reference, NonnullOwnPtr, or NonnullRefPtr. Since `Device` is a `RefCounted` (through `File`), NonnullRefPtr makes the most sense.

So, what does this achieve? Note that your check `if (match->m_attached_device == &device)` can also be done for NonnullRefPtr.

supercomputer7:
Let's not make `DevFS` a singleton, as it makes it impossible to load different instances of it in other mountpoints.
Also, the nature of `DevFS` resembles much more of `devtmpfs` than a "real DevFS" (you may call it a hybrid "DevFS" because it's also writable by the user, and we set permissions of files in `SystemServer`) and personally I'd want to see a little userspace daemon taking care of removing unplugged devices.

supercomputer7:
Also, what is the overall benefit of these changes? We don't support hot-pluggable devices, so it seems to me that we will get any useful functionality (yet!) from this. Maybe we should do that after we support hardware like NVMe, AHCI or USB :)

etaIneLp:
> Also, what is the overall benefit of these changes? We don't support hot-pluggable devices currently, so it seems to me that we will not get any useful functionality (yet!) from this. Maybe we should do that after we support hardware like NVMe, AHCI or USB :)

I had planned on modifying the storage management code to allow partition tables to be reloaded while the system is running. That requires the ability to dynamically create and destroy device files and is the main reason I made these changes.

Kernel/FileSystem/DevFS.h
etaIneLp: It is guaranteed to be non-null. Smart pointers aren't helpful here since the `Device` constructor and destructor ensure that a `DevFSDeviceInode` and its accompanying `Device` have the exact same lifetime. A reference would probably be better than a pointer, however.

etaIneLp:
> Let's not make `DevFS` a singleton, as it makes it impossible to load different instances of it in other mountpoints.
> Also, the nature of `DevFS` resembles much more of `devtmpfs` than a "real DevFS" (you may call it a "hybrid" DevFS because it's also writable by the user, and we set permissions of files in `SystemServer`) and personally I'd want to see a little userspace daemon taking care of removing device files.

Alright. I'll close this then, as the changes relied on `DevFS` being a singleton.

supercomputer7:
We could make these changes to work without making `DevFS` a singleton. Still, I don't think the kernel should remove device files after `SystemServer` takes contol.
Don't get discouraged, if you want to, we can discuss the matter on IRC or within a github issue.

supercomputer7:
We could make these changes to work without making `DevFS` a singleton. Still, I don't think the kernel should remove device files after `SystemServer` takes contol.
If you want to, we can discuss the matter on IRC or within a github issue.

PR #5223, opened 2021-02-02
DisplaySettings: Show a confirmation message when applying changes

EvilHowl:
I felt like this needed some kind of confirmation message :)
image

### DIFF
--- a/Userland/Applications/DisplaySettings/DisplaySettings.cpp
+++ b/Userland/Applications/DisplaySettings/DisplaySettings.cpp
@@ -272,10 +272,41 @@ void DisplaySettingsWidget::load_current_settings()
 
 void DisplaySettingsWidget::send_settings_to_window_server()
 {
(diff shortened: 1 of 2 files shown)


awesomekling:
Great idea, but please make it a Yes/No message box instead of an OK/Cancel one :)

tomuta:
We need to do it like most other OSs, the cancel option needs to automatically happen after say 15 seconds or so. Because if your monitor doesn't properly work with that resolution/refresh rate you won't be seeing this message and you have no clue how to invoke the "cancel" option...

EvilHowl:
> We need to do it like most other OSs, the cancel option needs to automatically happen after say 15 seconds or so. Because if your monitor doesn't properly work with that resolution/refresh rate you won't be seeing this message and you have no clue how to invoke the "cancel" option...

You're indeed right! Now, if after 10 seconds the user doesn't react to the message box, it will just close the message box and revert the changes. Here's how it looks now:

image

PR #5231, opened 2021-02-04
Kernel: Use the resolved parent path when testing create veil

wbowling:
When opening a file with O_CREAT, if the [parent exists but the file does](https://github.com/SerenityOS/serenity/blob/91db36064f5da1d5d9f8e010271d7951937a9de5/Kernel/FileSystem/VirtualFileSystem.cpp#L266) not then `validate_path_against_process_veil` is called with the [unresolved path](https://github.com/SerenityOS/serenity/blob/91db36064f5da1d5d9f8e010271d7951937a9de5/Kernel/FileSystem/VirtualFileSystem.cpp#L386) which could allow symlinks to be used to bypass the veil.

Instead of using the original path, we can use the resolved path of the parent custody with the basename of the path.

Example hax.cpp to write a file to the home directory:
```cpp
#include <LibCore/File.h>
#include <serenity.h>

int main()
{
    if (pledge("stdio rpath wpath cpath", nullptr) < 0) {
        perror("pledge");
        return 1;
    }

    if (unveil("/tmp", "rwc")) {
        perror("unveil");
        return 1;
    }

    unveil(nullptr, nullptr);

    if (symlink("/home/anon/missing", "/tmp/missing") < 0) {
        perror("symlink");
    }

    auto file = Core::File::open("/tmp/missing/hax", Core::IODevice::OpenMode::ReadWrite);
    if (file.is_error()) {
        fprintf(stderr, "Error: %s", file.error().characters());
        return 1;
    }
    if (!file.value()->write("haxed")) {
        perror("write");
    }

    return 0;
}
```

---
This change is [Reviewable](https://reviewable.io/reviews/serenityos/serenity/5231)

### DIFF
--- a/Kernel/FileSystem/VirtualFileSystem.cpp
+++ b/Kernel/FileSystem/VirtualFileSystem.cpp
@@ -1031,6 +1031,10 @@ KResultOr<NonnullRefPtr<Custody>> VFS::resolve_path_without_veil(StringView path
             if (!safe_to_follow_symlink(*child_inode, parent_metadata))
                 return EACCES;
 
+            auto result = validate_path_against_process_veil(custody->absolute_path(), options);
+            if (result.is_error())
+                return result;
+
             auto symlink_target = child_inode->resolve_as_link(parent, out_parent, options, symlink_recursion_level + 1);
             if (symlink_target.is_error() || !have_more_parts)
                 return symlink_target;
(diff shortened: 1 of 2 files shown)


alimpfard:
Why not validate both?
imo accessing a symlink that is itself not unveiled also wrong

bugaevc:
Agreed, we should validate both

wbowling:
Yeah could do both, although then https://github.com/SerenityOS/serenity/blob/91db36064f5da1d5d9f8e010271d7951937a9de5/Kernel/FileSystem/VirtualFileSystem.cpp#L943 would need to change as well to check the original path (otherwise it's only create that would reject symlinks and all other operations would still allow it).

Not sure how that would work with relative paths as they need to be resolved first, and they you have lost the original symlink if there was one

wbowling:
Updated do prevent symlinks from being followed unless they have been unveiled, also added a quick test

PR #5232, opened 2021-02-04
LibTTF: Check if the given offset plus offset table size would overflow

Lubrsi:
If it does overflow, it would think there was enough data to read in
table information, when there isn't. This would cause read buffer
overflows when reading in the table information.

Found by: https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=29338&sort=-opened&can=1&q=proj%3Aserenity

### DIFF
--- a/Userland/Libraries/LibTTF/Font.cpp
+++ b/Userland/Libraries/LibTTF/Font.cpp
@@ -241,6 +241,11 @@ RefPtr<Font> Font::load_from_memory(ByteBuffer& buffer, unsigned index)
 // FIXME: "loca" and "glyf" are not available for CFF fonts.
 RefPtr<Font> Font::load_from_offset(ByteBuffer&& buffer, u32 offset)
 {
+    if (Checked<u32>::addition_would_overflow(offset, (u32)Sizes::OffsetTable)) {
+        dbgln("Invalid offset in font header");
+        return nullptr;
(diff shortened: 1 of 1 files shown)


Userland/Libraries/LibTTF/Font.cpp
```diff
@@ -246,6 +246,11 @@ RefPtr<Font> Font::load_from_offset(ByteBuffer&& buffer, u32 offset)
         return nullptr;
     }
 
+    if (buffer.size() < offset + (u32)Offsets::NumTables + sizeof(u16) - 1) {
```
awesomekling, line 249: `offset` is input-controlled here, right? So the expression `offset + (u32)Offsets::NumTables + sizeof(u16) - 1` could overflow

Lubrsi: Indeed. In fact, the only way this issue can happen in the first place is because the offset in the file was 0xFFFFFFFB, which caused an overflow on the check above (`offset + (u32)Sizes::OffsetTable`) making it think it had enough data when it didn't. Should I just add an overflow check for `offset + (u32)Sizes::OffsetTable` instead?

awesomekling: That sounds sensible :)

PR #5236, opened 2021-02-05
Kernel: Clear pending interrupts before enabling IRQs of IDE Channel

### LINKED ISSUE #5229: Boot with smp=on is broken (black screen)
I can't boot anymore with `smp=on`. No stack traces, doesn't seem like it's crashing, just sitting there with a black screen for me.

Git bisect points to 36f6351edcee67963e80363f9f9f94a1d7db1919

supercomputer7:
Calling detect_disks() can generate interrupts, so we must clear it to
allow proper function when booting with kernel argument smp=on.

Fixes #5229.


supercomputer7:
I tested this with QEMU version 4.1.0 that was built with `BuildQemu.sh` script, PIO and DMA seem to work with smp enabled.

PR #5241, opened 2021-02-06
SystemMonitor: Use system color themes for graph widgets

ccapitalK:
#4895 pointed out that the graphs in SystemMonitor look kind of bad with dark themes. This PR changes the graphs to use system theme colors, using the text widget background color and syntax highlighting colors to ensure that the graph is both readable and thematically appropriate.

Screenshots:

### DIFF
--- a/Userland/Applications/SystemMonitor/GraphWidget.cpp
+++ b/Userland/Applications/SystemMonitor/GraphWidget.cpp
@@ -27,7 +27,9 @@
 #include "GraphWidget.h"
 #include <LibGUI/Painter.h>
 #include <LibGfx/Font.h>
+#include <LibGfx/Palette.h>
 #include <LibGfx/Path.h>
+#include <LibGfx/SystemTheme.h>
 
 GraphWidget::GraphWidget()
 {
@@ -49,7 +51,7 @@ void GraphWidget::paint_event(GUI::PaintEvent& event)
     GUI::Painter painter(*this);
     painter.add_clip_rect(event.rect());
     painter.add_clip_rect(frame_inner_rect());
-    painter.fill_rect(event.rect(), m_background_color);
+    painter.fill_rect(event.rect(), palette().base());
 
     auto inner_rect = frame_inner_rect();
     float scale = (float)inner_rect.height() / (float)m_max;
--- a/Userland/Applications/SystemMonitor/GraphWidget.h
+++ b/Userland/Applications/SystemMonitor/GraphWidget.h
@@ -39,8 +39,6 @@ public:
 
     void add_value(Vector<int, 1>&&);
 
-    void set_background_color(Color color) { m_background_color = color; }
-
     struct ValueFormat {
         Color line_color { Color::Transparent };
         Color background_color { Color::Transparent };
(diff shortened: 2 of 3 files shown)


ccapitalK:
I think the 4th screenshot is a bit hard to read, but I tried opening a cpp file in TextEditor using that theme and that was hard to read  as well, so I'm ok with it :smile: .

nvella:
Looks nice :smile:, I wonder though if it'd make sense now to rename the syntax colour entries in the theme palette to something more general? Seems a bit silly that we're using `syntax_*` palette entries to draw things that aren't syntax :sweat_smile:

linusg:
> I think the 4th screenshot is a bit **hard to read**, but I tried opening a cpp file in TextEditor using that theme and that was hard to read as well, **so I'm ok with it**.

> Hmm, I think we **need to improve the colour choices** for the text here.The green text is very **hard to read** in the first example, as is the blue text in the third. ([source](https://github.com/SerenityOS/serenity/pull/4895#issuecomment-757564153))

:thinking:

ccapitalK:
> > I think the 4th screenshot is a bit **hard to read**, but I tried opening a cpp file in TextEditor using that theme and that was hard to read as well, **so I'm ok with it**.
> 
> > Hmm, I think we **need to improve the colour choices** for the text here.The green text is very **hard to read** in the first example, as is the blue text in the third. ([source](https://github.com/SerenityOS/serenity/pull/4895#issuecomment-757564153))
> 
> thinking

What I meant was that I feel like anyone using the 4th theme would be ok with that specific combination of colours, since they would also be using it in the text editor. I personally don't like that theme because of it, but other people seem to like it, so I don't really have any complaints.

ccapitalK:
> Looks nice :smile: , I wonder though if it'd make sense now to rename the syntax colour entries in the theme palette to something more general? Seems a bit silly that we're using `syntax_*` palette entries to draw things that aren't syntax :sweat_smile:

It would probably be a good idea later down the line, but for now I think it should be fine to keep them as they are, since this is currently the only other thing that uses it. I'm not sure what we would name them though, maybe something like `ForegroundAlternate1` - `ForegroundAlternateN`? Or we could do something like bootstrap, and call them something like `ForegroundWarning`, `ForegroundInformation`, etc.

ccapitalK:
Now that #5297 has been merged, the system monitor looks like this, which IMO is more readable:

2021-02-12-192859_2048x1590_scrot

I believe this should be merge-able now, since I don't think renaming the syntax_* entries would be worth it just for this use (I see this as more of a one-off re-use of the theme elements, not the start of a pattern). Does anyone else have any concerns?

tomuta:
Was the background color removed intentionally? It doesn't look anything like in the screenshots above?

ccapitalK:
> Was the background color removed intentionally? It doesn't look anything like in the screenshots above?

Wdym?

tomuta:
I only see lines anymore, regardless of theme. And the diff suggests the same ;)

Screenshot from 2021-02-12 22-25-32

ccapitalK:
> I only see lines anymore, regardless of theme. And the diff suggests the same ;)
> 
> Screenshot from 2021-02-12 22-25-32

Oh, I think I get what you mean, the thing stretching below the line to the bottom that's semi-transparent. Yeah, that wasn't intentional, my bad, I'm going to fix that one up too :sweat_smile:

tomuta:
Sorry for my poor wording ;)

PR #5242, opened 2021-02-06
Profiler: Display timing information in ProfileTimelineWidget

ccapitalK:
Currently, there is no way to know when in a profile's duration a
sample was taken. This commit adds a basic timestamp to the timeline
widget, and a black bar to show where the cursor is hovering over.

Example:
2021-02-06-143308_2048x1590_scrot


awesomekling:
Cool!

PR #5264, opened 2021-02-07
LibWeb: Implement Element.getElementsBy{Tag,Class}Name()

linusg:
Just like the `Document` variants, but using the given `Element` as `for_each_in_subtree_of_type()` root.

### DIFF
--- a/Userland/Libraries/LibWeb/DOM/Element.cpp
+++ b/Userland/Libraries/LibWeb/DOM/Element.cpp
@@ -46,6 +46,7 @@
 #include <LibWeb/Layout/TableRowGroupBox.h>
 #include <LibWeb/Layout/TreeBuilder.h>
 #include <LibWeb/Layout/WidgetBox.h>
+#include <LibWeb/Namespace.h>
 
 namespace Web::DOM {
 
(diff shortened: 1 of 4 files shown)


Userland/Libraries/LibWeb/DOM/Element.cpp
```diff
@@ -340,4 +340,26 @@ bool Element::is_focused() const
...
+        return IterationDecision::Continue;
+    });
+    return elements;
+}
+
+NonnullRefPtrVector<Element> Element::get_elements_by_class_name(const FlyString& class_name) const
+{
+    NonnullRefPtrVector<Element> elements;
+    for_each_in_subtree_of_type<Element>([&](auto& element) {
+        if (element.has_class(class_name))
```
Lubrsi, line 358: In quirks modes, this should be case insensitive. From the spec:
```
The comparisons for the classes must be done in an ASCII case-insensitive manner if root’s node document’s mode is "quirks", and in an identical to manner otherwise. 
```
`root` being `this` in this case.

emilio:
Just drive-by, but FYI unlike `querySelector{,All}`, these collections are live in browsers, so this is more complicated than what you're doing. Though I guess this is a pre-existing issue anyways.

linusg:
> Though I guess this is a pre-existing issue anyways.

It most certainly is, this & what we already have for document is like the bare minimum to get it working for the most common use cases :)

awesomekling:
Implementing `HTMLCollection` could be a fun task in itself :)

PR #5265, opened 2021-02-08
WindowServer: Window frame transparency and render caching

tomuta:
This only renders the window frame once until the size of the window
changes, or some other event requires re-rendering. It is rendered
to a temporary bitmap, and then the top and bottom part is stored
in one bitmap as well as the left and right part. This also adds
an opacity setting, allowing it to be rendered with a different
opacity.

This makes it easier to enhance window themes and allows using
arbitrary bitmaps with e.g. alpha channels for e.g. shadows.

_This should make implementing a shadow effect as attempted in #5107 a lot easier. It does not address hit-testing based on the alpha channel in the rendered window frame._

This is an example of 0.75 opacity for window frames:
Screenshot from 2021-02-07 22-25-34

### DIFF
--- a/Userland/Services/WindowServer/WindowFrame.h
+++ b/Userland/Services/WindowServer/WindowFrame.h
@@ -45,7 +45,9 @@ public:
     ~WindowFrame();
 
     Gfx::IntRect rect() const;
-    void paint(Gfx::Painter&);
+    void paint(Gfx::Painter&, const Gfx::IntRect&);
+    void render(Gfx::Painter&);
+    void render_to_cache();
     void on_mouse_event(const MouseEvent&);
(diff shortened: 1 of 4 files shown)


Userland/Services/WindowServer/WindowFrame.h
```diff
@@ -62,6 +62,17 @@ class WindowFrame {
...
     void start_flash_animation();
 
+    float opacity() const { return m_opacity; }
+
+    bool is_opaque() const
+    {
+        if (opacity() < 1.0f)
+            return false;
+        //if (has_alpha_channel())
+        //    return false;
```
awesomekling, line 74: Let's remove this check instead of commenting it out :)

PR #5269, opened 2021-02-08
Fix HackStudio/LibCpp crash

itamar8910:
The C++ language server was crashing at load time because of an unresolved symbol that belonged to LibSyntax (Which also made hackstudio crash when a .cpp file is opened).

Also, I noticed that (for some reason) using dbgln() in the dynamic loader crashes it now. This needs to be investigated,
but for now I moved the dbgln() error reporting that's enabled by default to use dbgprintf().

### DIFF
--- a/Userland/Libraries/LibCpp/CMakeLists.txt
+++ b/Userland/Libraries/LibCpp/CMakeLists.txt
@@ -6,4 +6,4 @@ set(SOURCES
 )
 
 serenity_lib(LibCpp cpp)
-target_link_libraries(LibCpp LibC)
+target_link_libraries(LibCpp LibC LibSyntax)


alimpfard:
> using dbgln() in the dynamic loader crashes it now

Got a stack trace for that? Or does it hit some assertion?
I doubt my latest dbgln changes could lead to a runtime crash, but it could've caused this

awesomekling:
> > using dbgln() in the dynamic loader crashes it now
> 
> Got a stack trace for that? Or does it hit some assertion?
> I doubt my latest dbgln changes could lead to a runtime crash, but it could've caused this

Fixed here 412cb66666f0827c75f182f626e2285d3fd051db

PR #5271, opened 2021-02-08
LibCore+LibGUI : Propose a new Core::FileWatcher class and use it in FileSystemModel

Dexesttp:
The idea of the FileWatcher class is based on [this comment](https://github.com/SerenityOS/serenity/pull/5213#discussion_r571575872) in #5213.

As part of the review, I found that the Core library didn't provide an easy way to watch a file, which made implementing requests such as the one above pretty hard. With this change, watching either a file or a directory in GUI applications should be easier.

`Core::FileWatcher` provides an event-based API via the `on_change` callback. This callback returns a `FileWatcher::Event` that can be either of type `Modified`, `ChildAdded` or `ChildRemoved` (mirroring the watch_file syscall API).
In addition, if the `ChildAdded` or `ChildRemoved` event is raised, an extra `child_path` member of the event is available to indicate the path of the child that was added or removed.

As a first use-case, this patch replaces the existing code in LibGUI::FileSystemModel to use the new FileWatcher API.

Note that this seems redundant with the `Core::DirectoryWatcher` class, but the major difference is that DirectoryWatcher is blocking while FileWatcher integrates in the event loop via the `Core::Notifier` class.


nvella:
Looking interesting, I'll definitely consider giving this a go in #5213 when this is merged.

Dexesttp:
Question for @itamar8910 : This new FileWatcher replaces and extends the DirectoryWatcher class in non-blocking contexts, but the CrashDaemon you wrote back in #3738 does not use an event loop and needs the blocking aspect of DirectoryWatcher.

Would it make sense to modify CrashDaemon so it uses an event loop, or is it better to leave it like that? I don't really know the constraints behind the original design of the Daemon.

If keeping both the blocking and non-blocking versions, do you see an alternative design that would allow to deduplicate the code more?

Dexesttp:
Thinking about it, if we keep both, it would be interesting to move them both to the FileWatcher.cpp/.h file, rename DirectoryWatcher to BlockingFileWatcher, and add handling for the Modified event in the BlockingFileWatcher. This way, the false FileWatcher/DirectoryWatcher dichotomy is not there anymore.

Dexesttp:
Ok, I did exactly what I said in the last message. This didn't need any change to the CrashDaemon except for some renaming, since there was already a check for ChildAdded in the service.

With this, and if there's no other remarks, I think this is ready for review / merge whenever !

Dexesttp:
The last push simply removes some unneeded imports (I needed Core::Object in an earlier design, but don't anymore so it's removed). There's no functional changes.

PR #5272, opened 2021-02-09
WindowServer: Implement simple window shadows

tomuta:
This implements simple window shadows (flicker-free), originally attempted in #5107

- [x] Fix glitch with shadow rendering on top right and bottom right
- [x] Fix glitches rendering shadow around small windows (e.g. tooltips)
- [x] Fix interaction with window frame elements (not yet accounting for shadow offset)

Screenshot from 2021-02-08 22-52-03


awesomekling:
This is gorgeous 🤩

PR #5277, opened 2021-02-09
Remove all text relocations and remove PaX holes that existed for it.

nico:
Also reduces memory use after boot from 178 MiB to 173 MiB, due to libc having no dirty pages.

The patch is obviously correct.

---

Actually, I have no idea why this works. I thought plt_trampoline.S is what makes @PLT relocations work, so not sure why things don't blow up here.

@ADKaster @itamar8910

### DIFF
--- a/Kernel/Syscalls/mmap.cpp
+++ b/Kernel/Syscalls/mmap.cpp
@@ -37,49 +37,6 @@
 
 namespace Kernel {
 
-static bool should_make_executable_exception_for_dynamic_loader(bool make_readable, bool make_writable, bool make_executable, const Region& region)
-{
-    // Normally we don't allow W -> X transitions, but we have to make an exception
-    // for the dynamic loader, which needs to do this after performing text relocations.
-
-    // FIXME: Investigate whether we could get rid of all text relocations entirely.
-
-    // The exception is only made if all the following criteria is fulfilled:
-
-    // The region must be RW
-    if (!(region.is_readable() && region.is_writable() && !region.is_executable()))
-        return false;
-
-    // The region wants to become RX
-    if (!(make_readable && !make_writable && make_executable))
-        return false;
-
-    // The region is backed by a file
-    if (!region.vmobject().is_inode())
-        return false;
-
-    // The file mapping is private, not shared (no relocations in a shared mapping!)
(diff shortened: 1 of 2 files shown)


ADKaster:
PLT worked and then with the full dynamic loader change it stopped working. The loader currently bypasses the lazy jump slot relocations happening at runtime by enforcing "bind now" behavior -- which does all the lazy relocs at program startup / lib load time.

ADKaster:
Err looking at this it's actually very strange. That plt trampoline doesn't need built into LibC at all. Nor the dynamic loader/ dynamic object classes. Or. I guess any part of LibELF? There's some opportunities for Library refactoring here.

awesomekling:
Can we just not support lazy binding? It has obvious security drawbacks and I'm not sure what benefit it actually provides.

nico:
As far as I know the main benefit is that it makes application startup faster. If you link to a new .so and don't do lazy binding, you have to process all relocations to that so at load time, instead of paying just for the symbols your app actually needs, and only when it needs them.

(Another good way to make startup / binding faster is to do two-level lookup keyed first by so and then by symbol like macos (pre dyld cache, I suppose), instead of having a single huge namespace for all symbols.)

Maybe relocation processing is fast enough these days that it's not needed. Since we currently don't do it, maybe it's not worth adding this mmap exception for it. /shruggie

awesomekling:
We still need the mmap exception for immediate binding. That only goes away if we ban text relocations entirely, unless I'm misunderstanding something(?)

nico:
Right, right now the libc.so textrel is the only textrel there is according to `for l in $(find . -name '*.so'); do echo $l; (readelf -d $l | grep TEXT); done`. The only way to get them is to use asm files that add them, which most projects don't do. So maybe not supporting textrels is fine. (I guess they should get a nicer diag if we do that though.)

awesomekling:
Okay, let's try not supporting them at all. With a better diagnostic, like you say.
This is not the kind of project that needs to make concessions for hypothetical future ports :smile:

awesomekling:
Ninja-merged the first part of this patch in 1e6d04c7462848bc47a09e51a2f375022bc3ed93 since it's just too good. ;)

PR #5280, opened 2021-02-09
WindowServer: Allow specifying different shadows for menus and tooltips

tomuta:
Also update the Redmond 2000 theme to drop shadows more Redmond-like.

Screenshot from 2021-02-09 14-14-23


Userland/Services/WindowServer/WindowFrame.cpp
```diff
@@ -56,12 +56,12 @@ static Gfx::Bitmap* s_minimize_icon;
...
 static Gfx::Bitmap* s_restore_icon;
 static Gfx::Bitmap* s_close_icon;
-static Gfx::Bitmap* s_window_shadow;
 
 static String s_last_title_button_icons_path;
 static int s_last_title_button_icons_scale;
 
-static String s_last_window_shadow_path;
+static Gfx::Bitmap *s_window_shadow, *s_menu_shadow, *s_tooltip_shadow;
+static String s_last_window_shadow_path, s_last_menu_shadow_path, s_last_tooltip_shadow_path;
```
awesomekling, line 64: pls no multiple variables on same line

tomuta: Will do. I was thinking about moving this stuff into some class instead, but that's an exercise for some other day...

PR #5282, opened 2021-02-09
Userland: Use INET_ADDRSTRLEN for inet_ntop() buffers

linusg:
There's no point in using different, seemingly randomly sized buffers as the required size for storing an IPv4 address representation is well known (16 bytes).


awesomekling:
Uh well that makes a lot of sense!

PR #5284, opened 2021-02-09
Kernel: Merge split function and data sections into one during linking

yeetari:
This makes the linker merge all `.text.*`, `.rodata.*` and `.data.*` sections into single `.text`, `.rodata` and `.data` sections to reduce the number of sections in the final binary. This reduces kernel binary size by ~0.5MiB, which is about a 20% reduction in a stripped build.

### DIFF
--- a/Kernel/linker.ld
+++ b/Kernel/linker.ld
@@ -11,16 +11,15 @@ SECTIONS
         $<TARGET_OBJECTS:boot>
         *(.multiboot)
         start_of_kernel_text = .;
-        *(.text)
-        *(.text.startup)
 
         start_of_safemem_text = .;
-        *(.text.safemem)
+        KEEP(*(.text.safemem))
         end_of_safemem_text = .;
         start_of_safemem_atomic_text = .;
-        *(.text.safemem.atomic)
+        KEEP(*(.text.safemem.atomic))
         end_of_safemem_atomic_text = .;
 
+        *(.text*)
         end_of_kernel_text = .;
     }
 
@@ -34,13 +33,13 @@ SECTIONS
         *(.ctors)
         end_ctors = .;
 
-        *(.rodata)
+        *(.rodata*)
     }
 
     .data ALIGN(4K) : AT (ADDR(.data) - 0xc0000000)
     {
         start_of_kernel_data = .;
-        *(.data)
+        *(.data*)
         end_of_kernel_data = .;
     }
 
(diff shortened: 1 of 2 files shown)


Kernel/linker.ld
```diff
@@ -11,7 +11,7 @@ SECTIONS
         $<TARGET_OBJECTS:boot>
         *(.multiboot)
         start_of_kernel_text = .;
-        *(.text)
+        *(.text*)
         *(.text.startup)
```
tomuta, line 15: Not an LD expert, but it seems like `*.text*` would suck in `.text.safemem` and `.text.safemem.atomic`? Or does this only apply to sections not explicitly mentioned?

yeetari: Yep, you're right. Moving `*(.text*)` after the safemem sections seems to fix this though, although I'm not sure if having the safemem stuff before the main `.text` is bad? I've also removed `*(.text.startup)` since `*(.text*)` includes that.

tomuta: Does it work by coincidence, or is it documented to work like this?

As for the safemem, moving that code into another section would affect the `start_of_safemem_atomic_text`/`end_of_safemem_atomic_text`, which are used by the page fault handler for a check whether a page fault needs to be handled in a special way. We probably should add a few assertions during startup checking that these functions reside within the range of these labels so that it doesn't get broken accidentally. I think the other changes are fine, but I didn't take a super close look.

yeetari: I believe it's documented to work like this. [Here](https://doc.ecoscentric.com/gnutools/doc/ld/Input-Section-Wildcards.html#Input-Section-Wildcards) it says
> If a file name matches more than one wildcard pattern, or if a file name appears explicitly and is also matched by a wildcard pattern, the linker will use the first match in the linker script.

In our case, the file name is always `*`, however `*(.text.safemem)` and `*(.text.safemem.atomic)` appear before `*(.text*)` so I think they are matched first. I think the assertions are a good idea, we could also assert that `&start_of_safemem_atomic_text != &end_of_safemem_atomic_text` which is what was happening before I moved `*(.text*)` after the safemem sections.

yeetari: Where would the best place to put the assertions be? Somewhere in `init.cpp` maybe?

tomuta: Probably, maybe after this line?
https://github.com/SerenityOS/serenity/blob/ef9bfbd38343cf9bfa708c0d714bd8556031deae/Kernel/init.cpp#L133

yeetari: Ok, I've added the assertions. I've also just wrapped the safemem sections with `KEEP()` to make sure the linker doesn't try to remove them.

awesomekling:
Cool

PR #5286, opened 2021-02-10
WindowServer: Fix compositing of fullscreen window

### LINKED ISSUE #4810: LibGUI/LibGfx: paint_events are ineffective on full screen windows
Once a window is set to full screen with `window->set_fullscreen(true)` only the first paint event takes effect. Subsequent paint events are triggered but nothing is drawn.

This used to work. Something changed in the graphics libraries about 6 months ago.

Reproduce:

Apply the following diff then run `Conway` :

```diff
diff --git a/Games/Conway/main.cpp b/Games/Conway/main.cpp
index ae57220f4..8507fe3b4 100644
--- a/Games/Conway/main.cpp
+++ b/Games/Conway/main.cpp
@@ -85,6 +85,7 @@ int main(int argc, char** argv)
     app->set_menubar(move(menubar));
 
     window->show();
+    window->set_fullscreen(true);
 
     return app->exec();
 }
```

linusg:
I honestly don't know the internals of all this and what exactly is going on, but this fixes compositing of the fullscreen window. By trial and error I found that specifically `m_invalidated_all` needs to be set to `false`, so it's probably different behaviour in `prepare_dirty_rects()`, which depends on that...
Either way, the code composing all windows in non-fullscreen mode calls `Window::clear_dirty_rects()` for each, so not doing that for the fullscreen window as well seems like an oversight.

Fixes #4810.

### DIFF
--- a/Userland/Services/WindowServer/Compositor.cpp
+++ b/Userland/Services/WindowServer/Compositor.cpp
@@ -408,6 +408,7 @@ void Compositor::compose()
     if (m_invalidated_window) {
         if (auto* fullscreen_window = wm.active_fullscreen_window()) {
             compose_window(*fullscreen_window);
+            fullscreen_window->clear_dirty_rects();
         } else {
             wm.for_each_visible_window_from_back_to_front([&](Window& window) {
                 compose_window(window);


tomuta:
This looks like an oversight, and the fix looks fine. We really need animated screen savers ;-) Anyone want to implement one that does color changing triangles or whatever it was back in the Win 3.1x days? ;)

tomuta (approved):

PR #5287, opened 2021-02-10
HackStudio: Show notification if 'make' is not available

itamar8910:
We previously popped a MessageBox for this, but a notification is less disruptive.


awesomekling:
Excellent use of notifications

PR #5288, opened 2021-02-10
WindowServer: Don't re-render the window frame when not needed

tomuta:
Merely moving a window shouldn't require re-rendering the window
frame anymore now that we cache the rendered frame in bitmaps. This
reduces CPU usage significantly when moving windows.


awesomekling:
That's just real nice :)

PR #5292, opened 2021-02-11
WindowServer: Fix clearing area not covered by backing bitmap

### LINKED ISSUE #5291: WindowServer: Area not covered by backing bitmap not cleared properly
Moving a window (with or without shadow) over an area not (yet) covered by a window backing store will causes parts of the window to be left on the screen without being cleared. When transparency is involved, this leaves behind images from the temporary transparency rendering bitmap. This is easiest reproduced by stopping a process, then enlarging that process' window and then moving other windows over the area not covered by the backing window.

Screenshot from 2021-02-10 21-41-35
Screenshot from 2021-02-10 21-41-56

tomuta:
We only cleared the area not covered by the backing bitmap if a
rendering rectangle intersected with the backing bitmap. But because
we are potentially calling the render function many times we need
to always clear the area not covered by the backing bitmap, whether
it intersects or not.

Fixes #5291


tomuta:
Added some changes on top of this to finally blit hung application content with transparency, if the window is transparent:

Screenshot from 2021-02-10 23-28-14

PR #5294, opened 2021-02-11
Kernel syscall fuzzing!

BenWiederhake:
This PR:
- Merges the previous two syscall-inconveniencers into the new program `/usr/Tests/Kernel/fuzz-syscalls`
- Makes `syscall`(1) *extremely* useful now (see screenshot)
- Fixes a stray `ASSERT()` in the Kernel that a user program could trigger (because it would indicate a program error, not a Kernel error)
- Avoids calling "deadly" syscalls during fuzzing – don't blame me for noticing that back in #5197, the Kernel crashed too early for me to notice that! :P
- Fixes a silly bug in `copy_string_from_user` in the Kernel: It would try to `klog()` the string that cannot be read. Which fails, of course, because the string cannot be read! :^)
- Fixes a silly kmem exhaustion bug in `writev`
- Fixes a too-early cast to enum of an arbitrary int (Found in cooperation with KUBSan!)

This fuzzer and the fixes it involves can be merged, but will probably introduce random build failures if we also merge #4921 (another reason why I don't think we should blindly run all tests during each build).

The fuzzer sometimes crashes, because, well, it does a bunch of random syscalls, what did you expect? :P

Ideally call it with `/usr/Tests/Kernel/fuzz-syscalls < /dev/null` to avoid getting stuck on `read(0, …)` calls.

This can also get stuck on `read(1, …)` for some reason.

And finally, some eyecandy:

Bildschirmfoto_2021-02-11_22-53-19


nico:
Yes, switching over the enum when it's still an int is the right fix for that particular problem.

PR #5295, opened 2021-02-11
WindowServer: Allow different shadows for active/inactive windows

tomuta:
Also allow specifying different shadows for the task and menu bar.

_It's subjective but I think it would make sense to make most shadows a little lighter and only use the darker shadow for the window currently active. Because this is now part of the theme configuration, it can easily be changed._

Screenshot from 2021-02-11 15-18-21


tomuta:
I was a little torn about whether the tooltips and menus should have a light or dark shadow...

awesomekling:
Looks nice :)

PR #5300, opened 2021-02-12
Themes: Make Redmond 2000 menu shadows a bit lighter

tomuta:
_I got the shadow a little too dark, I think this looks a little more Redmond like..._

Screenshot from 2021-02-11 20-51-38


nvella:
Apologies if this might be a bit of a nit-pick, but did W2K have shadows on its context menus? 😅

linusg:
> Apologies if this might be a bit of a nit-pick, but did W2K have shadows on its context menus?

Good point IMO - it did not:

image

(https://bellard.org/jslinux/vm.html?url=https://bellard.org/jslinux/win2k.cfg&mem=192&graphic=1&w=1024&h=768)

While in general shadows are a nice thing to have, it takes something away from the old Windows-y look and feel...

awesomekling:
Right, I think menu shadows were added in XP, which a lot of people ran in classic theme mode, and then retroactively misremember as Windows 2000 (myself included) :)

tomuta:
Dropping shadows should probably be configurable anyway, and then depending on power profile (e.g. laptop on battery in power savings mode) disabled automatically... ;) Or we can just make two themes. I don't really care one way or another ;)

nvella:
Perhaps an idea for the Theme Editor? Users could be able to clone and apply a custom version of their favourite theme, with their desired level of shadows :)

PR #5304, opened 2021-02-12
Kernel: Hold less locks when receiving ICMP packets

bugaevc:
* We don't have to lock the "all IPv4 sockets" in exclusive mode, shared mode is enough for just reading the list (as opposed to modifying it).
* We don't have to lock socket's own lock at all, the `IPv4Socket::did_receive()` implementation takes care of this.
* Most importantly, we don't have to hold the "all IPv4 sockets" across the `IPv4Socket::did_receive()` call(s). We can copy the current ICMP socket list while holding the lock, then release the lock, and then call `IPv4Socket::did_receive()` on all the ICMP sockets in our list.

These changes fix a deadlock triggered by receiving ICMP messages when using tap
networking setup (as opposed to QEMU's default user/SLIRP networking) on the host.

cc @alimpfard @tomuta

### DIFF
--- a/Kernel/Net/NetworkTask.cpp
+++ b/Kernel/Net/NetworkTask.cpp
@@ -250,13 +250,17 @@ void handle_icmp(const EthernetFrameHeader& eth, const IPv4Packet& ipv4_packet, 
 #endif
 
     {
-        LOCKER(IPv4Socket::all_sockets().lock());
-        for (RefPtr<IPv4Socket> socket : IPv4Socket::all_sockets().resource()) {
-            LOCKER(socket->lock());
-            if (socket->protocol() != (unsigned)IPv4Protocol::ICMP)
-                continue;
-            socket->did_receive(ipv4_packet.source(), 0, KBuffer::copy(&ipv4_packet, sizeof(IPv4Packet) + ipv4_packet.payload_size()), packet_timestamp);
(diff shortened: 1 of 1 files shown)


bugaevc:
I admit I have not investigated why exactly the deadlock happens and why this is enough to fix it.

tomuta:
We have some major problems in the FS and network stack, in particular the use of `Lock`. It is unsafe to call them from within deferred calls. The reason for this is that it is entirely unpredictable (by design) what `Thread::current` (and therefore also `Process:current`) may be when the call is executed. This is a lot more prone to happen on my SMP PR. The worst situation is when the idle thread blocks, because at that point we're in big trouble and end up stalling the system entirely. Most deadlocks related to `Lock` I've observed are that a deferred call using a `Lock` locks it while in a different `Thread::current`, which then means that thread "holds" a lock that will never be cleared. To solve some of these problems I added work queues (Linux has that concept as well), which are essentially just kernel threads that run functions similar to deferred calls, but in a predictable `Thread::current`, which means that `Lock` can be used. In fact my plan is to add an `ASSERT` to `Lock::lock` and `Lock::unlock` to ensure it's never called from a deferred function call.

I don't know if this is what's happening here, but it's given me a lot of grief with my SMP PR effort. I'm in the process of ripping out the use of `Lock` from the FS stack in my SMP PR for this and also other reasons.

PR #5305, opened 2021-02-12
Themes: Fix Basalt window title background colors to use full opacity

### LINKED ISSUE #5303: Themes/WindowServer: font trail on Basalt theme
It seems like a font trail appears only on the Basalt theme on window titlebar when you move it.
Screen Shot 2021-02-12 at 10 58 32

tomuta:
Fix #5303

_Not sure if this is really just masking the problem, maybe this really should work? Is it maybe a regression now that we cache the rendered frame in bitmaps?_


tomuta:
Closing this for now, working on enhancing the frame rendering to be able to use alpha channels in themes properly.

Issue #5317, opened 2021-02-13
LibCrypto: Read buffer overflow in Crypto::der_decode_sequence


Lubrsi:
Found with FuzzRSAKeyParsing.

[crash-f944dcd635f9801f7ac90a407fbc479964dec024.txt](https://github.com/SerenityOS/serenity/files/5975197/crash-f944dcd635f9801f7ac90a407fbc479964dec024.txt) (with txt extension to allow uploading to GH)

Trace:
```
==157609==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x602000008d92 at pc 0x00000056a24e bp 0x7fffa1e21750 sp 0x7fffa1e21748
READ of size 1 at 0x602000008d92 thread T0
    #0 0x56a24d in Crypto::der_decode_sequence(unsigned char const*, unsigned long, Crypto::ASN1::List*, unsigned long, bool) /home/lukew/Desktop/serenity-project/serenity/build/../Userland/Libraries/LibCrypto/ASN1/DER.h:345:18
    #1 0x566ba4 in Crypto::PK::RSA::parse_rsa_key(AK::Span<unsigned char const>) /home/lukew/Desktop/serenity-project/serenity/build/../Userland/Libraries/LibCrypto/PK/RSA.cpp:57:9
    #2 0x5623de in LLVMFuzzerTestOneInput /home/lukew/Desktop/serenity-project/serenity/build/../Meta/Lagom/Fuzzers/FuzzRSA.cpp:34:5
    #3 0x46a6d1 in fuzzer::Fuzzer::ExecuteCallback(unsigned char const*, unsigned long) (/home/lukew/Desktop/serenity-project/serenity/build/Meta/Lagom/Fuzzers/FuzzRSA+0x46a6d1)
    #4 0x469e15 in fuzzer::Fuzzer::RunOne(unsigned char const*, unsigned long, bool, fuzzer::InputInfo*, bool*) (/home/lukew/Desktop/serenity-project/serenity/build/Meta/Lagom/Fuzzers/FuzzRSA+0x469e15)
    #5 0x46c0b7 in fuzzer::Fuzzer::MutateAndTestOne() (/home/lukew/Desktop/serenity-project/serenity/build/Meta/Lagom/Fuzzers/FuzzRSA+0x46c0b7)
[...]

PR #5318, opened 2021-02-13
Lagom/Fuzzers: Add RSA key parser fuzzer

Lubrsi:
First issue: #5317


tomuta:
Just out of curiosity, is it possible to provide fuzzers with hints of what they might be fuzzing? Like for complex binary file formats to try to find more complex exploits?

Lubrsi:
You can provide dictionary files for text file formats: https://github.com/SerenityOS/serenity/blob/master/Meta/Lagom/Fuzzers/FuzzJs.dict
I haven't used dictionaries in a while though.
Not sure for binary file formats.

alimpfard:
The ASN.1 parser is neither robust nor complete...maybe this fuzzer will incentivize someone to finally work on fixing it...

PR #5320, opened 2021-02-13
AK+Format: Don't cast to size_t when you want u64.

### LINKED ISSUE #5308: Formatter broken for long long
Steps to reproduce:

`dbgln("Hello {} world", 0x123456789abcdef0LL);`

Actual result: `Hello 2596069104 world`
Expected result: `Hello 1311768467463790320 world`

Explanation (python):
```
>>> hex(1311768467463790320)
'0x123456789abcdef0'
>>> hex(2596069104)
'0x9abcdef0'
```

Found while fixing a weird syscall issue :)

asynts:
In Serenity, `size_t` is defined as `u32`, thus `static_cast<size_t>(value)` truncates the value.

Since the unit test only ever failed inside of Serenity, I quickly added another `TestFormat.cpp` file. When we finally run tests in Serenity, we'd have to merge these files.

Fixes #5308

### DIFF
--- a/Userland/Tests/CMakeLists.txt
+++ b/Userland/Tests/CMakeLists.txt
@@ -1,3 +1,4 @@
+add_subdirectory(AK)
 add_subdirectory(Kernel)
 add_subdirectory(LibC)
 add_subdirectory(LibGfx)
--- a/AK/Format.cpp
+++ b/AK/Format.cpp
@@ -368,7 +368,7 @@ void FormatBuilder::put_i64(
     const auto is_negative = value < 0;
     value = is_negative ? -value : value;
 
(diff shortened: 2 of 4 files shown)


Userland/Tests/CMakeLists.txt
```diff
@@ -3,3 +3,4 @@ add_subdirectory(LibC)
 add_subdirectory(LibGfx)
 add_subdirectory(LibM)
 add_subdirectory(UserspaceEmulator)
+add_subdirectory(AK)
```
BenWiederhake, line 6: Thanks for fixing this so fast! :D

Could you sort this alphabetically? I believe "AK" goes to the top.

Issue #5315, opened 2021-02-13
Kernel: addition overflow on long int in Thread::BlockTimeout constructor


Lubrsi:
Found with fuzz-syscalls.

Trace:
```
fuzz-syscalls(58): #6620: Calling poll(0x115a8000, 0xc0000000, 0xbffff000) with 0x115a8000 containing [0xffffffff, 0x00000000, 0xa405a98b, 0xc0000000, 0xbffff000, 0xbffff000, 0xc0000000, 0x00000001, 0x00000000]
[#0 H(58:58)]: KUBSAN: addition overflow, 'long int' (32-bit)
[#0 H(58:58)]: KUBSAN: at .././AK/Time.h, line 117, column: 32
[#0 H(58:58)]: 0xc022302a  print_location +0xfe
[#0 H(58:58)]: 0xc01e0744  Kernel::Thread::BlockTimeout::BlockTimeout(bool, timespec const*, timespec const*, int) +0x1a8
[#0 H(58:58)]: 0xc01f77eb  Kernel::Process::sys$poll(AK::Userspace<Kernel::Syscall::SC_poll_params const*>) +0x1b9
[#0 H(58:58)]: 0xc01ccda8  syscall_handler +0xa87
[#0 H(58:58)]: 0xc01cc2e9  syscall_asm_entry +0x31
```

BenWiederhake:
This seems already fixed in 62f0f73bf0e15e9830486a51c2114ae8399dbade, can you confirm?

BenWiederhake:
Fixed by #5323. You can verify this by running `syscall poll [ 0xffffffff 0 "Hello, world!" 0xc0000000 ]`.

IdanHo:
Closed as fixed (see above)

PR #5323, opened 2021-02-13
Fuzz-related fixes

BenWiederhake:
Most importantly: This fixes lots of timespec-related integer overflows! The Kernel assumes that all `timespec` structs have `.tv_nsec < 1_000_000_000`, as any sane caller would ensure. However, #5315 shows that we forgot this check in `sys$poll`. And in fact everywhere!

Instead of trying to find and fix all the places where we might receive an evil timespec, patch the implicit `copy_from_user<timespec>` instantiation to handle this for us.

Note that we prevent a similar overflow with `.tv_sec` by *rounding down* to **36533 million years**. I'm almost certain noone will notice this difference ;) . (If they do: Hello, very far future!)

Furthermore, fix a typo and reduce the number of accepted fds in `sys$poll` to the maximum open fds – trying to poll more than that is bound to contain invalid or duplicate fds.


Kernel/Syscalls/select.cpp
```diff
@@ -148,7 +148,7 @@ int Process::sys$poll(Userspace<const Syscall::SC_poll_params*> user_params)
         return -EFAULT;
 
     // This limit is just a number from the place where numbers come from.
```
awesomekling, line 150: This comment is no longer true. :)

BenWiederhake: Hmm, yes it is? The numbers (fds) come from `Process`, which contains the line `static const int m_max_open_file_descriptors { FD_SETSIZE };`, or did I misunderstand you?

awesomekling: The original meaning of the comment was "this is just a random number that I summoned from the universe's number generator". The new number is meaningful, relevant, and just an overall great choice!

BenWiederhake: Oh! I thought you actually meant something specific with "the place where numbers come from" xD

Kernel/StdLib.h
```diff
@@ -75,6 +76,23 @@ template<typename T>
     return copy_from_user(dest, src, sizeof(T));
 }
 
+template<>
+[[nodiscard]] inline bool copy_from_user<timespec>(timespec* dest, const timespec* src)
```
awesomekling, line 80: This feels quite weird. I wouldn't expect a low-level API like `copy_from_user` to interfere with the data being copied.

BenWiederhake: I agree that it's an unusual place, but as I mentioned in the first post: We use user-provided `timespec`s as-is in many places, and I think trying to hunt them all down is a bad idea.

What would be a better design? I considered`Optional<timespec> copy_timespec_from_user(const timespec*)` for a while, but disregarded it for the above reason.

tomuta:
I started working on this last night (didn't get as far though) and my solution was going to be a specialized `AK::Checked<timespec>`.

As for `Thread::BlockTimeout`, I think we've got two options:
- Set `m_infinite` to `true` if an overflow happens. In additions, in syscalls validated that `timespec::tv_nsec < 1000000000` and return `EINVAL`
- Make each caller check if `Thread::BlockTimeout` encountered an overflow and fail with `EINVAL`.

I kind of like the first solution better. If we overflow then we essentially will block indefinitely even if the overflow had been handled by some bignum operation. Infinite because it would exceed the lifetime of the equipment anyway. And it's not as messy to implement. I do NOT know how Linux handles these cases as a reference point.

awesomekling:
My preferred solution here would be a much fancier C++ type for time. It's silly that we restrict ourselves to using the crappy C `struct timespec`. Really, we should leave the POSIX types at the syscall boundary if there's any way to make it better with C++. :)

tomuta:
Maybe just nanoseconds in a `u64`? That should be enough for like 400 years. And would make overflow checking trivial as well.

Kernel/StdLib.h
bgianfo: Could you static assert that T is not timespec in the regular `copy_from_user`, with a message that people should use copy_timespec_from_user? 
Best of both worlds?

Hendiadyoin1:
> a much fancier C++ type for time.

Maybe port the stdc++ chrono library, its "fancy", wordy and supports all kinds of times

tomuta:
> > a much fancier C++ type for time.
> 
> Maybe port the stdc++ chrono library, its "fancy", wordy and supports all kinds of times

Not a big fan of it, i can never seem to remember how to use it properly... It's just too difficult and overkill for this purpose I think.

Kernel/StdLib.h
BenWiederhake: Turns out, static asserts *always* trigger, even if the template isn't instantiated. Huh! So I went with ` __attribute__((deprecated("use " #REPLACEMENT " instead")))` instead.

BenWiederhake:
Changes since first push are too numerous, so I'll just pretend I start from scratch:

- This fixes lots of timespec-related integer overflows! The Kernel assumes that all timespec structs have `.tv_nsec < 1'000'000'000`, as any sane caller would ensure. However, #5315 shows that we forgot this check in `sys$poll`. And in fact everywhere!
- Instead of trying to find and fix all the places where we might receive an evil timespec, deprecate and thus disable the implicit  `cppy_from_user<timespec>` instantiation (and in case anyone *does* manage to call it, fail with `VERIFY_NOT_REACHED`).
- In it's place, a new `copy_time_from_user` takes care of properly sanitizing the input.
- The new `AK::Time` type is currently only used to sanitize the input, and then the Kernel immediately falls back to timespec and timeval. This probably still hides some overflows, but now it's less likely and (hopefully) less severe. (I originally intended to do that in this PR but … uhhh, let's split that up.)
- I decided to go with 64-bit seconds + 32 bit nanoseconds to prevent any precision loss. This way, there is *absolutely no* argument to stick with timespec/timeval anywhere in the Kernel!
- Running `syscall poll [ 0xffffffff 0 "Hello, world!" 0xc0000000 ]` no longer triggers KUBSan, yay! :D
- Getting addition and subtraction with carry right in an overflow-free manner was an exercise in patience. Just look at `TestTime.cpp`. There's more tests and edge cases that SLOC in `Test.cpp`! D:
- Rebased for convenience, but thankfully there were no conflicts anyway.

ADKaster:
Re: Fancy c++ time: My vote would be a fancy class with a u64 nanoseconds representation, AK::Duration. I think absl::Duration has like, u32 quarter-nanoseconds and i64 seconds or something? Which seems like overkill.

BenWiederhake:
Changes since last description: Turns out, converting the Kernel to `AK::Time` was easier than expected, so I did that.
Pain points:
- I disabled the PR yesterday due to a `return EFAULT; return KSuccess;` typo that was surprisingly hard to find.
- Add headers in the correct commit (found while bisecting)
- `TimeManagement` does some weird locking around memory accesses. I don't dare touch it.
- I believe there's a bug in TCP sockets re-sending unacknowledged packages. This is likely to have never been discovered since Qemu probably does that, too. In particular, the following lines look suspicious to me:
  ```
  auto diff = packet.tx_time - now;
  if (diff <= Time::from_nanoseconds(500'000'000))
      continue;
  // re-send, and among other things:
  packet.tx_time = now;
  ```
  I am absolutely certain that I translated the signs here correctly, so this is what the code originally did. However, computing "tx_time - now" will always be a negative (or zero) amount of time, but not negative enough to start wrapping. Therefore, it always takes the `continue` branch in the old code. Hence, 846398a47ed54a09ad69aac230d138738c05472e.

AK/Tests/TestTime.cpp
```diff
@@ -0,0 +1,205 @@
+/*
+ * Copyright (c) 2018-2020, Andreas Kling <kling@serenityos.org>
```
awesomekling, line 2: I don't remember writing this file :)

Kernel/Net/TCPSocket.cpp
```diff
@@ -243,7 +243,7 @@ void TCPSocket::send_outgoing_packets()
     LOCKER(m_not_acked_lock, Lock::Mode::Shared);
     for (auto& packet : m_not_acked) {
         // FIXME: This doesn't look correct to me. Bug?
-        auto diff = packet.tx_time - now;
+        auto diff = now - packet.tx_time;
```
awesomekling: Hey let's remove the FIXME too :D

AK/Tests/TestTime.cpp
BenWiederhake: :man_facepalming:

AK/Time.h
```diff
@@ -66,6 +72,55 @@ inline int years_to_days_since_epoch(int year)
...
+    timeval to_timeval() const;
+
+    bool operator==(const Time& other) const { return this->m_seconds == other.m_seconds && this->m_nanoseconds == other.m_nanoseconds; }
+    bool operator!=(const Time& other) const { return !(*this == other); }
+    Time operator+(const Time& other) const;
+    Time operator-(const Time& other) const;
+    bool operator<(const Time& other) const;
+    bool operator<=(const Time& other) const;
+    bool operator>(const Time& other) const;
+    bool operator>=(const Time& other) const;
```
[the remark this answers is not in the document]
BenWiederhake: I already looked into this particular method for `DistinctNumeric`, and we can't easily support it because it is required to return a particular type in `std::…`, which we don't have.

PR #5324, opened 2021-02-13
Check mount_point isn't already mounted in mount and bind_mount 

BluuWasTaken:


bugaevc:
This would prevent us from stacking multiple mounts on top of each other, right?

Why?

linusg:
@BluuWasTaken there are linter errors an an unanswered question - do intend to make any changes here?

PR #5327, opened 2021-02-13
HackStudio: Handle LanguageServer crashing & use parser-based c++ autocomplete by default

itamar8910:
also some bug fixes in the c++ autocomplete engine.

Screenshot of a LanguageServer crashing:
hs-langserver-crash


awesomekling:
This is awesome. Finally we start to take advantage of being multiprocess ;)

PR #5329, opened 2021-02-13
Meta: Fix debug-flag detection

BenWiederhake:
Detection broke when we moved from '#ifdef DEBUG_FOO dbgln()' to 'dbgln<DEBUG_FOO>()'.
This patch makes detection more general, which sadly runs into more false-positives.

No rotten code was found, hooray! :^)

### DIFF
--- a/Meta/CMake/all_the_debug_macros.cmake
+++ b/Meta/CMake/all_the_debug_macros.cmake
@@ -170,7 +170,15 @@ set(SYSCALL_1_DEBUG ON)
 
 # False positive: DEBUG is a flag but it works differently.
 # set(DEBUG ON)
+# False positive: DT_DEBUG is a flag, but for a bitset, not a feature.
+# set(DT_DEBUG ON)
+# False positive: GUI_DND_DEBUG is a flag, but passed as an envvar.
+# set(GUI_DND_DEBUG ON)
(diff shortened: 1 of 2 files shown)


alimpfard:
why not look for `(if (?:constexpr )?|dbgln_if)\(\w+_DEBUG` ?

BenWiederhake:
Because that's a terribly long and complicated regex, and wouldn't even catch flags such as `DEBUG_SPAM`, `DEBUG_CPP_LANGUAGE_SERVER`, or `DEBUG_AUTOCOMPLETE`.

alimpfard:
How about this: `(if (?:constexpr )?|dbgln_if)\(\w+`
If something is used _as_ a debug flag, let's just say that it is a debug flag.

There's no `dbgln_if<flag>`, `dbgln_if` is a macro :)
Not sure if the codebase still has `dbgln<flag>(...)`, I don't think it does

BenWiederhake:
> There's no dbgln_if<flag>, dbgln_if is a macro :)

Nooo, my nit-picking was nit-picked xD

My point is: I'd rather add three false-positives and have a simple, robust scheme that *always* works than have to go back and fix the linter's regex every few months.

alimpfard:
Fair enough.

PR #5325, opened 2021-02-13
A lot of places: Add a CMake port

alimpfard:
Current status:
- Builds, and works (can configure Serenity!)
- Takes about ~two hours~ 80m to build
- ~Needs to NIH `scanf` (#90)~
- Probably needs code cleanup

Todos before ready to review:
- [x] Fix #90
- [x] Cleanup some messy changes (the "bump image size" commit)
- [x] Click that "Ready for review" button

PR #5330, opened 2021-02-13
Lint port table ('AvailablePorts.md')

BenWiederhake:
As requested by popular demand ;) https://github.com/SerenityOS/serenity/pull/5325#discussion_r575657614


Meta/lint-ports.py
```diff
@@ -0,0 +1,67 @@
+#!/usr/bin/env python3
+
+import os
+import re
+
+
+PORT_TABLE_REGEX = re.compile('^\\| \\[`([^`]+)`\\][^`]+$')
```
nico, line 7: nits:

1. Use a raw string (`r'^\| \[`([^`]+)`\][^`]+$'`)
2. Add a comment like `# Matches e.g. "| [``bash``]..." and captures "bash" in group 1

Meta/lint-ports.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+def read_port_table(filename):
+    ports = set()
+    with open(filename, 'r') as fp:
+        for line in fp:
+            match = PORT_TABLE_REGEX.match(line)
+            if not match:
+                continue
+            ports.add(match.groups()[0])
```
nico, line 19: nit: `match.group(1)`

But since it's a single group, I think the whole loop body can likely be replaced with `ports.add(PORT_TABLE_REGEX.findall(line))`

nico: If you use `re.M` in options, I think you don't even need to go line-by-line. Just

```
  return set(PORT_TABLE_REGEX.findall(open(filename, 'r').read()))
```

might be enough as function body.

BenWiederhake: I'll use `with open()`, to avoid leaking the fd. It doesn't really matter here, but [I like cleaning up fds](#5342). :)

BenWiederhake:
Changes since last push:
- Incorporated all the suggestions – thanks! I didn't know that `re` has such a convenient interface :D
- Rebased for convenience

PR #5331, opened 2021-02-13
LibGfx: Enable Painter::blit to apply opacity and mix alpha channels

tomuta:
This enables us to properly render windows that use both opacity and
alpha channels.


awesomekling:
`Gfx::Painter` consolidation is just :heart:

PR #5337, opened 2021-02-14
Fix build on OpenBSD

jcs:
Building on OpenBSD seems to get broken every few months.  This makes this a bit more portable and should hopefully not break on any other OSes.


awesomekling:
cool!

PR #5339, opened 2021-02-14
LibCrypto: memcmp() all bytes in UnsignedBigInteger::operator==

linusg:


Userland/Libraries/LibCrypto/BigInt/UnsignedBigInteger.cpp
```diff
@@ -281,7 +281,7 @@ bool UnsignedBigInteger::operator==(const UnsignedBigInteger& other) const
     if (length != other.trimmed_length())
         return false;
 
-    return !__builtin_memcmp(m_words.data(), other.words().data(), length);
+    return !__builtin_memcmp(m_words.data(), other.words().data(), length * (BITS_IN_WORD / 8));
```
awesomekling, line 284: Hey wait a minute.. how about `return m_words == other.m_words;`?

linusg: That, for whatever non-obvious reason, causes an infinite loop somewhere in the `BigInt.js` LibJS test...

awesomekling: Well that's really weird but ok for now then.

awesomekling: Oh I get it, because we should be comparing against the *trimmed* length. All good then :)

PR #5345, opened 2021-02-14
Adventures in LookupServer, pt. 3

bugaevc:
Lots of stuff here, culminating with our own DNS server!

A picture is worth a thousand words (aka 2KB):

image

Here, 192.168.122.1 is my host, 192.168.122.42 is SerenityOS VM. The .1 asks the .42, where's SerenityOS.org? LookupServer does not find it in the cache, randomizes the case and sends the question upstream. Upon receiving and validating and caching the answer, it replies back to .1 (notice that it replies back with the case used in the .1-to-.42 query!)

Then .1 asks .42 again, this time for serenityOS.ORG, LookupServer finds it in its cache, and replies back immediately, again using the case from the query. Finally, .1 asks .42 for serenityOS.ORGH, which is (was) not an existing domain. LookupServer asks its upstream servers, and finally returns a NXDOMAIN to .1.

Inverse lookups (get host by addr) are not demonstrated, but should also work.

cc @alimpfard @linusg @itamar8910

### DIFF
--- a/Base/etc/LookupServer.ini
+++ b/Base/etc/LookupServer.ini
@@ -1,2 +1,3 @@
 [DNS]
 Nameservers=1.1.1.1,1.0.0.1
+EnableServer=0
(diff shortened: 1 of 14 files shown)


awesomekling:
Super neat!

PR #5346, opened 2021-02-14
Ports: Add build_installed.sh script to rebuild installed ports

linusg:
Based on `build_all.sh`, but reading from `packages.db` and not hiding build output.

cc @alimpfard

### DIFF
--- a/Ports/build_all.sh
+++ b/Ports/build_all.sh
@@ -21,7 +21,7 @@ for file in *; do
             if $(./package.sh > /dev/null 2>&1 ); then
                 echo "Built ${dirname}."
             else
-                echo "ERROR: Built ${dirname} not succesful!"
+                echo "ERROR: Build of ${dirname} was not successful!"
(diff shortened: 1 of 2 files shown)


tomuta:
Hm can we add some documentation on how these are to be used? I only know to run `./package.sh` in each port, but this seems like it keeps track of them somehow?

linusg:
> Hm can we add some documentation on how these are to be used?

Will do!

> this seems like it keeps track of them somehow?

It does, and that's already documented - along with most other aspects of the ports system :)

> https://github.com/SerenityOS/serenity/tree/master/Ports#using-ports-scripts
> 
> Installed ports are being tracked in `Build/packages.db` (a simple text file). You can delete this file at any time, in fact it must be edited or removed when clearing the build directory as port dependencies may not be installed again otherwise.

tomuta:
Oh sorry, somehow overlooked this ;)

PR #5355, opened 2021-02-14
WindowServer: Add support for alpha channel based hit testing

tomuta:
This enables implementing non-rectangular window shapes, including
non-rectangular window frames.

### DIFF
--- a/Userland/Services/WindowServer/ClientConnection.cpp
+++ b/Userland/Services/WindowServer/ClientConnection.cpp
@@ -463,6 +463,7 @@ OwnPtr<Messages::WindowServer::CreateWindowResponse> ClientConnection::handle(co
         window->recalculate_rect();
     }
     window->set_opacity(message.opacity());
+    window->set_alpha_hit_threshold(message.alpha_hit_threshold());
     window->set_size_increment(message.size_increment());
     window->set_base_size(message.base_size());
     window->set_resize_aspect_ratio(message.resize_aspect_ratio());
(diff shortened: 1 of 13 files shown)


Userland/Services/WindowServer/Window.cpp
```diff
@@ -873,4 +873,21 @@ bool Window::is_descendant_of(Window& window) const
...
+bool Window::hit_test(const Gfx::IntPoint& point, bool include_frame) const
+{
+    if (!frame().rect().contains(point))
+        return false;
+    if (include_frame) {
+        if (!rect().contains(point))
+            return frame().hit_test(point);
+    } else if (!rect().contains(point)) {
+        return false;
+    }
```
awesomekling, line 885: I think this would look better "inverted":

```cpp
if (!rect().contains(point)) {
    if (include_frame)
        return frame().hit_test(point);
    return false;
}
```

Userland/Services/WindowServer/ClientConnection.cpp
```diff
@@ -636,6 +637,17 @@ OwnPtr<Messages::WindowServer::SetWindowHasAlphaChannelResponse> ClientConnectio
     return make<Messages::WindowServer::SetWindowHasAlphaChannelResponse>();
 }
 
+OwnPtr<Messages::WindowServer::SetWindowAlphaHitThresholdResponse> ClientConnection::handle(const Messages::WindowServer::SetWindowAlphaHitThreshold& message)
+{
+    auto it = m_windows.find(message.window_id());
+    if (it == m_windows.end()) {
+        did_misbehave("SetWindowAlphaHitThresholdResponse: Bad window ID");
```
awesomekling, line 644:
```suggestion
        did_misbehave("SetWindowAlphaHitThreshold: Bad window ID");
```

Userland/Services/WindowServer/WindowFrame.cpp
```diff
@@ -535,6 +535,40 @@ void WindowFrame::layout_buttons()
...
+    auto frame_rect = rect();
+    if (!frame_rect.contains(point))
+        return false;
+    auto window_rect = m_window.rect();
+    if (window_rect.contains(point))
+        return false;
+
+    u8 alpha_threshold = Gfx::WindowTheme::current().frame_alpha_hit_threshold(window_state_for_theme()) * 255;
+    if (alpha_threshold == 0)
+        return true;
```
awesomekling, line 551: What is the difference between threshold 0 and 1?

awesomekling (changes requested):

Userland/Services/WindowServer/WindowFrame.cpp
awesomekling: Oh wait, I guess 0 just means that no alpha hit testing should be done at all.

tomuta: That's right, a threshold of 0 would mean always hit, ignoring the alpha channel

PR #5356, opened 2021-02-15
CMake: Add 'setup-and-run' target to perform all prereqs and run the image

bgianfo:
Running 'ninja && ninja install && ninja image && ninja run` is mildly annoying.
I got tired, and came up with this instead, which does the right thing
and I don't have to type out the incantation every time.

I'm not in love with the name, any suggestions would be welcome.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -29,6 +29,18 @@ add_custom_target(run
     USES_TERMINAL
 )
 
+# This can currently only be implemented by ordered commands
+# as cmake doesn't support inter dependency ordering, and we
+# would like to avoid inject dependencies on the existing
+# custom commands to allow people to run commands adhoc with
+# out forcing re-builds when they might not want them.
+add_custom_target(setup-and-run
+    COMMAND ${CMAKE_MAKE_PROGRAM} install
+    COMMAND ${CMAKE_MAKE_PROGRAM} image
+    COMMAND ${CMAKE_MAKE_PROGRAM} run
+    USES_TERMINAL
+)
+
 add_custom_target(image
     DEPENDS qemu-image
 )


CMakeLists.txt
```diff
@@ -28,6 +28,11 @@ add_custom_target(run
     COMMAND ${CMAKE_SOURCE_DIR}/Meta/run.sh
     USES_TERMINAL
 )
+add_custom_target(all-and-run
+    DEPENDS all install qemu-image run
```
emanuele6, line 32: doesn't the `DEPENDS ... run` make it run run.sh twice?

emanuele6: I tried this and it made the image before finishing compiling.

emanuele6:
I tried this and it made the image before finishing compiling.

bgianfo:
Ok, I'll make sure it consistently works. I didn't test it exhaustively.
Thanks for the heads up!

alimpfard:
> Running 'ninja && ninja install && ninja image && ninja` is too much.

And that's why `^R&& ninja` is probably helpful.
(Also, the first `ninja` is extraneous and you can probably make a shell alias too)


Though it'd be nice to have a target that builds and runs, less things to type out when I mess with Shell's history and all goes poof.

CMakeLists.txt
bgianfo: Yup copy pasta bug, thanks!

bgianfo:
> (Also, the first `ninja` is extraneous and you can probably make a shell alias too)
I think way back when cmake support was first merged this didn't work (or I incorrectly convinced my self it didn't work) and I just kept typing it that way ever since. 
Thanks for the heads up.

bgianfo:
> I tried this and it made the image and started qemu before finishing compilation (I don't think it even ever started).
> Maybe custom targets cannot depend on `all` and/or `install` (?)

I confirmed your findings, after much googling and fighting cmake, it turns out cmake doesn't have any sort of way to force inter dependency ordering.
They assume that if you need ordering then you will declare dependencies in the targets them selves, which normally would solve the issues. However
when this came up during the initial cmake implementation folks stated they didn't want dependencies forced onto developers when sometimes
they might want to rebuild the image for instance without actually re-compiling the world and we should let that be possible.

That led me to the cursed implementation that is now rebased onto this PR, where we use COMMAND ordering to force the total ordering of
dependencies and everything just seems to work. I've tested ninja and make generators for cmake.

bgianfo:
`ThinkT510` suggested setup-and-run instead, which sounds better than all-and-run to me... so I'll switch to that.

> 23:54   bgianf | Anyone have a better name than 'all-and-run'? https://github.com/SerenityOS/serenity/pull/5356 
> 23:55   bgianf | as always, naming is hard
> 00:00   ThinkT510 | setup-and-run?
> 00:02    bgianf | Yea that sounds better :)
> 00:04   ThinkT510 | prepare-to-run might be more humorous

bgianfo:
I think this is good to merge now.

PR #5359, opened 2021-02-15
Kernel: Avoid some un-necessary copies coming from range based for loops

bgianfo:
- The irq_controller was getting add_ref/released needlessly during enumeration.

- Used ranges were also getting needlessly copied.

### DIFF
--- a/Kernel/Interrupts/InterruptManagement.cpp
+++ b/Kernel/Interrupts/InterruptManagement.cpp
@@ -117,7 +117,7 @@ RefPtr<IRQController> InterruptManagement::get_responsible_irq_controller(u8 int
     if (m_interrupt_controllers.size() == 1 && m_interrupt_controllers[0]->type() == IRQControllerType::i8259) {
         return m_interrupt_controllers[0];
     }
-    for (auto irq_controller : m_interrupt_controllers) {
+    for (auto& irq_controller : m_interrupt_controllers) {
         if (irq_controller->gsi_base() <= interrupt_vector)
             if (!irq_controller->is_hard_disabled())
                 return irq_controller;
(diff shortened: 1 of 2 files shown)


Kernel/Interrupts/InterruptManagement.cpp
```diff
@@ -117,7 +117,7 @@ RefPtr<IRQController> InterruptManagement::get_responsible_irq_controller(u8 int
     if (m_interrupt_controllers.size() == 1 && m_interrupt_controllers[0]->type() == IRQControllerType::i8259) {
         return m_interrupt_controllers[0];
     }
-    for (auto irq_controller : m_interrupt_controllers) {
+    for (const auto& irq_controller : m_interrupt_controllers) {
```
bugaevc, line 120: I *think* it's idiomatic Serenity style to leave out the `const` in `const auto`, because the compiler can infer it.

bgianfo: Is it? 

> $ rg "const auto&" | wc -l
> 173

alimpfard: Another cultured `rg` user, I see.
If you compare the counts instead...
```
$ rg --pcre2 '(?!=const )auto&' | wc -l
4390

# vs

rg --pcre2 'const auto&' | wc -l
174
```

bgianfo: 🤷 sure, "fixed".

alimpfard: I don't mind either way, just thought I'd add that bit lol

PR #5360, opened 2021-02-15
WindowServer+LibGUI: Variable minimum window sizes

nvella:
Minimum window size can now be customised and set at runtime  via the `SetWindowMinimumSize` `WindowServer` message and the `set_minimum_size` `LibGUI::Window` method. Windows are automatically resized if their minimum size is set larger than the current window size. The default minimum size remains at 50x50. Addresses #5357.

### DIFF
--- a/Userland/Libraries/LibGUI/Window.cpp
+++ b/Userland/Libraries/LibGUI/Window.cpp
@@ -149,6 +149,7 @@ void Window::show()
         m_opacity_when_windowless,
         m_base_size,
         m_size_increment,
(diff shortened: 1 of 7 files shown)


nvella:
Making this a draft for now as it's not currently possible to set the minimum size of a non-visible window, this does limit the usefulness a fair bit.

nvella:
This should also now fix `WindowServer` clients not being aware of their real size when they `CreateWindow` below the minimum size, a `WindowResized` message is posted if the minimum size is applied.

Issue #4412, opened 2020-12-14
DHCPClient: Doesn't work for class C networks


danboid:
I had my first successful bare metal Serenity boot in some time today and one of the main things I wanted to test was to see if I could get networking to work. No such luck unfortunately.

Yes, the machine I was testing on had a Realtek 8139 NIC in it and I could see in the boot log that it had detected it OK but DHCPClient hadn't fetched / assigned it a valid address and so networking didn't work. The NIC was attached to a class C network but serenity assigned it a class A address.

Also, if you run DHCPClient from the terminal as root you get the error in the attached screenshot.

![dhcp](https://user-images.githubusercontent.com/1429783/102116390-cf2d3500-3e34-11eb-80e3-e957f99c1f01.png)

I presume that means it is only intended to be run at startup?

linusg:
> Also, if you run DHCPClient from the terminal as root you get the error in the attached screenshot.

That doesn't look like a root shell in your screenshot :thinking:

> I presume that means it is only intended to be run at startup?

It's just non-existent error handling. Basically: something fails, and it doesn't like that and decides to crash :smile: 
Not sure what the error is though.

---

cc @alimpfard

danboid:
Ah yes, you're right @linusg but you get the exact same error if you run it as root.

danboid:
Its also worth mentioning that Browser freezed/crashed whe I clicked on a link - prob because networking wasn't actually working but serenity thought it was?

alimpfard:
> Not sure what the error is though

It's trying to bind to udp port 68 to listen to the incoming DHCP responses, and that's somehow failing.
as `anon`, this is expected (68 is a privileged port) but it shouldn't fail like this for `root`.

`DHCPClient` should log in the debug console if it gets an IP (or it fails to negotiate one), is there anything like that happening?

danboid:
Debug console on a bare metal boot?

alimpfard:
serial port?
if that's not accessible...ouch.
what's the "assigned" IP? 10.0.2.15?

danboid:
Unfortunately I won't be able to check anything for you until I'm next back in work which could be a few weeks off now. If I go into work again this week I'll do whatever testing is required.

alimpfard:
Time is perhaps the only thing we have a lot of :)

danboid:
@awesomekling does all the debug console stuff get output to the serial port by default, after booting?

danboid:
[...]

PR #5362, opened 2021-02-15
DHCPClient: Make it work on TAP interfaces (aka real interfaces) 

alimpfard:
cc @bugaevc 

Might also deal with #4412, should probably be checked.

### DIFF
--- a/Kernel/Net/Routing.cpp
+++ b/Kernel/Net/Routing.cpp
@@ -205,6 +205,12 @@ RoutingDecision route_to(const IPv4Address& target, const IPv4Address& source, c
         return { nullptr, {} };
     }
 
+    // If it's a broadcast, we already know everything we need to know.
+    // FIXME: We should also deal with the case where `target_addr` is
+    //        a broadcast to a subnet rather than a full broadcast.
+    if (target_addr == 0xffffffff && matches(adapter))
+        return { adapter, { 0xff, 0xff, 0xff, 0xff, 0xff, 0xff } };
+
     {
         LOCKER(arp_table().lock());
         auto addr = arp_table().resource().get(next_hop_ip);
--- a/Userland/Services/DHCPClient/DHCPv4.h
+++ b/Userland/Services/DHCPClient/DHCPv4.h
@@ -39,7 +39,7 @@
 #include <string.h>
 
 enum class DHCPv4Flags : u16 {
-    Broadcast = 1,
+    Broadcast = 1 << 15,
     /* everything else is reserved and must be zero */
 };
 
--- a/Userland/Services/DHCPClient/DHCPv4Client.cpp
+++ b/Userland/Services/DHCPClient/DHCPv4Client.cpp
@@ -276,6 +276,7 @@ void DHCPv4Client::dhcp_request(DHCPv4Transaction& transaction, const DHCPv4Pack
 
     DHCPv4Packet& packet = builder.peek();
     packet.set_op(DHCPv4Op::BootRequest);
+    packet.ciaddr() = offer.yiaddr();
     packet.set_htype(1); // 10mb ethernet
     packet.set_hlen(sizeof(MACAddress));
     packet.set_xid(offer.xid());
--- a/Userland/Services/DHCPClient/DHCPv4Client.cpp
+++ b/Userland/Services/DHCPClient/DHCPv4Client.cpp
@@ -53,7 +53,9 @@ static void send(const InterfaceDescriptor& iface, const DHCPv4Packet& packet, C
     dst.sin_addr.s_addr = IPv4Address { 255, 255, 255, 255 }.to_u32();
     memset(&dst.sin_zero, 0, sizeof(dst.sin_zero));
 
+    dbgln_if(DHCPV4CLIENT_DEBUG, "sendto({} bound to {}, ..., {} at {}) = ...?", fd, iface.m_ifname, dst.sin_addr.s_addr, dst.sin_port);
     auto rc = sendto(fd, &packet, sizeof(packet), 0, (sockaddr*)&dst, sizeof(dst));
+    dbgln_if(DHCPV4CLIENT_DEBUG, "sendto({}) = {}", fd, rc);
     if (rc < 0) {
         dbgln("sendto failed with {}", strerror(errno));
         // FIXME: what do we do here?
@@ -113,8 +115,8 @@ DHCPv4Client::DHCPv4Client(Vector<InterfaceDescriptor> ifnames)
     m_server->on_ready_to_receive = [this] {
         auto buffer = m_server->receive(sizeof(DHCPv4Packet));
         dbgln("Received {} bytes", buffer.size());
-        if (buffer.size() != sizeof(DHCPv4Packet)) {
-            dbgln("we expected {} bytes, this is a bad packet", sizeof(DHCPv4Packet));
+        if (buffer.size() < sizeof(DHCPv4Packet) - DHCPV4_OPTION_FIELD_MAX_LENGTH + 1 || buffer.size() > sizeof(DHCPv4Packet)) {
+            dbgln("we expected {}-{} bytes, this is a bad packet", sizeof(DHCPv4Packet) - DHCPV4_OPTION_FIELD_MAX_LENGTH + 1, sizeof(DHCPv4Packet));
             return;
         }
         auto& packet = *(DHCPv4Packet*)buffer.data();
--- a/Userland/Services/DHCPClient/DHCPv4Client.cpp
+++ b/Userland/Services/DHCPClient/DHCPv4Client.cpp
@@ -286,6 +286,8 @@ void DHCPv4Client::dhcp_request(DHCPv4Transaction& transaction, const DHCPv4Pack
 
     // set packet options
     builder.set_message_type(DHCPMessageType::DHCPRequest);
+    builder.add_option(DHCPOption::ServerIdentifier, sizeof(IPv4Address), &offer.siaddr());
+    builder.add_option(DHCPOption::RequestedIPAddress, sizeof(IPv4Address), &offer.yiaddr());
     auto& dhcp_packet = builder.build();
 
     // broadcast the "request" request
--- a/Userland/Services/DHCPClient/main.cpp
+++ b/Userland/Services/DHCPClient/main.cpp
@@ -25,6 +25,7 @@
  */
 
 #include "DHCPv4Client.h"
+#include <AK/Debug.h>
 #include <AK/JsonArray.h>
 #include <AK/JsonObject.h>
 #include <AK/String.h>
@@ -69,7 +70,7 @@ int main([[maybe_unused]] int argc, [[maybe_unused]] char** argv)
 
     auto file = Core::File::construct("/proc/net/adapters");
     if (!file->open(Core::IODevice::ReadOnly)) {
-        warnln("Error: {}", file->error_string());
+        dbgln("Error: {}", file->error_string());
         return 1;
     }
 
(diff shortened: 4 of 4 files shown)


Kernel/Net/Routing.cpp
```diff
@@ -205,6 +205,10 @@ RoutingDecision route_to(const IPv4Address& target, const IPv4Address& source, c
         return { nullptr, {} };
     }
 
+    // If it's a broadcast, we already know everything we need to know.
+    if (target_addr == 0xffffffff && matches(adapter))
+        return { adapter, { 0xff, 0xff, 0xff, 0xff, 0xff, 0xff } };
```
bugaevc, line 210: What should we do in case it's not a 255.255.255.255, but something like 192.168.1.255? So a broadcast, but not to *everyone* in the world, but just to a subnet?

alimpfard: I'm not very clear on the software side of this, but I think we'd have to implement a broadcast routing algorithm for that (assuming that ARP fails and no one around us knows how to broadcast to a subnet), probably convert it to a series of unicasts.

bugaevc:
Nitpick: I don't think "make it work on TAP interfaces" should be the commit title. The TAP exists (or does not) on the host, not in Serenity. This commit fixes stuff, and as a side effect (and as a goal) it makes it work if the host is using TAP+dhcpd. So it'd be appropriate to mention it in the commit message body at the end, e.g. "With this, a SerenityOS VM connected to a Linux host's TAP interface gets fully functioning networking :^)" or something, but not in the title — the title should be about the thing it fixes.

alimpfard:
Fair. let me try to phrase that better.

Kernel/Net/Routing.cpp
bugaevc: I'm nowhere near being an expert on this, but I would assume we should route such semi-broadcasts in much the same way, i.e. match the address against our adapters and then just send the broadcast there. No point in going through ARP since no individual peer has .255 anyway.

alimpfard: There is no "there" in this case, the only reason we can easily route network-wide broadcasts is because there's a reserved MAC address (ff:ff:ff:ff:ff:ff) for it.
To broadcast to a subnet, our best bet is to find someone that can do that for us - hence ARP.

I'm not an expert either, so I might be very wrong either way.

alimpfard:
Figured splitting it up is the best course of action, the changes weren't that related to each other anyway

Kernel/Net/Routing.cpp
bugaevc: It's my understanding that when broadcasting to a subnet, we also use the `ff:ff:ff:ff:ff:ff` MAC (as in, broadcast to everyone connected to this cable!) with a proper broadcast IP (.255). That is of course if we're connected to the subnet ourselves, otherwise we'd have to route it as usual to our gateway.

alimpfard: Hmmm...you might be right :thinking:
I'll slap a FIXME there, someone that knows about this more than I do can address that.

bugaevc:
So this doesn't quite work here.

The first issue is that evidently, DHCPClient tries to gain an IP address immediately on startup, not when the link becomes connected. In my case, I launch the VM with TAP networking first, and then later I configure connectivity/DHCP/whatever and bring up the TAP interface on the host. But DHCPClient doesn't notice the machine is now online.

Second, when I kill and re-run DHCPClient manually (as root, of course), it logs
```
DHCPClient(36): Received 300 bytes
DHCPClient(36): We were offered 192.168.122.76 for 269352960
DHCPClient(36): Leasing the IP 192.168.122.76 for adapter e1k0
```
which looks OK, but the adapter nevertheless is set to 10.0.2.15 (as seen in `ifconfig` and SystemMonitor).

alimpfard:
> not when the link becomes connected

We don't really have a way to be notified of that, do we?
if such a mechanism is in place, I could extend DHCPClient.

> which looks OK, but the adapter nevertheless is set to 10.0.2.15

it logs up to dhcp request, but I wonder if it's getting the ack at all? could you build with `DHCPV4CLIENT_DEBUG` and see if it's getting an ack at all?

bugaevc:
> could you build with `DHCPV4CLIENT_DEBUG` and see if it's getting an ack at all?

```
SystemServer(5): Service DHCPClient has exited with exit code 15
Shell(28): Job entry 'kill 18' deleted in 51 ms
DHCPClient(30): Trying to lease an IP for e1k0 with ID 16838
DHCPClient(30): sendto(8 bound to e1k0, ..., 4294967295 at 17152) = ...?
DHCPClient(30): sendto(8) = 548
DHCPClient(30): Received 300 bytes
DHCPClient(30): Here are the options: DHCP Options (9 entries)
	option 58 (4 bytes): 0  0  7  8 
	option 3 (4 bytes): 192  168  122  1 
	option 1 (4 bytes): 255  255  255  0 
	option 53 (1 bytes): 2 
	option 28 (4 bytes): 192  168  122  255 
	option 59 (4 bytes): 0  0  12  78 
	option 54 (4 bytes): 192  168  122  1 
	option 6 (4 bytes): 192  168  122  1 
	option 51 (4 bytes): 0  0  14  16 

DHCPClient(30): We were offered 192.168.122.76 for 269352960
DHCPClient(30): Leasing the IP 192.168.122.76 for adapter e1k0
DHCPClient(30): sendto(9 bound to e1k0, ..., 4294967295 at 17152) = ...?
DHCPClient(30): sendto(9) = 548
```

alimpfard:
So...there's no ack from the dhcp server :thinking:

alimpfard:
Three things I can think of:
- the dhcp server wants something (an option, maybe) it's not getting
- the dhcp server is not broadcasting the ack (expecting the client to set its IP before getting an ack? that's weird)
- our udp stack is somehow dropping the ack packet?

bugaevc:
image
Looks like there isn't one, indeed

bugaevc:
Interestingly, DHCP server logs do mention the ack:

```
Feb 16 12:51:53 bugaevc-laptop dnsmasq-dhcp[5587]: DHCPDISCOVER(virbr0) 52:54:00:12:34:56
Feb 16 12:51:53 bugaevc-laptop dnsmasq-dhcp[5587]: DHCPOFFER(virbr0) 192.168.122.76 52:54:00:12:34:56
Feb 16 12:51:53 bugaevc-laptop dnsmasq-dhcp[5587]: DHCPREQUEST(virbr0) 192.168.122.76 52:54:00:12:34:56
Feb 16 12:51:53 bugaevc-laptop dnsmasq-dhcp[5587]: DHCPACK(virbr0) 192.168.122.76 52:54:00:12:34:56
```

alimpfard:
I was thinking that maybe the server wants `siaddr` set, but it seems to be responding just fine.
Maybe a firewall issue?

bugaevc:
Let me actually check how this looks w/ other VMs.

bugaevc:
image

With systemd-networkd as the DHCP client, it looks like this:
* the source address is 0.0.0.0
* the ACK is directed to the new address

alimpfard:
> the ACK is directed to the new address

is the BOOTP Broadcast bit set in the dhcp request packet? it'd be pretty weird if it was, and the dhcp still opted to unicast

bugaevc:
image
And here's the ISC DHCP client

bugaevc:
> is the BOOTP Broadcast bit set in the dhcp request packet?

In the sd-nd case, no. In the ISC DHCP client case, yes.

alimpfard:
So at least the server is acting to-spec, that's good.
That with the fact that the server logs an ack makes me really suspicious of a firewall of sorts; unless these clients are also in a VM, in which case I'm out of ideas

bugaevc:
These clients are all in VMs (sd-nd on a Debian GNU/Linux VM,  ISC DHCP on a Debian GNU/Hurd VM :wink:)

The difference must be in the source address being or not being 0.0.0.0. The spec says

> DHCP messages broadcast by a client prior to that client obtaining its IP address must have the source address field in the IP header set to 0.

alimpfard:
:thinking: in the IP header? That sounds like an odd thing to ask for.
Even then, why is it responding to our DISCOVER, but not to our REQUEST? our IP header has whatever as its source address

bugaevc:
> Even then, why is it responding to our DISCOVER, but not to our REQUEST?

¯\\\_(ツ)\_/¯, but we can't complain if we are the ones who doesn't follow the spec :slightly_smiling_face:

alimpfard:
Let's try setting the server identifier option. I really doubt anyone cares about the source IP

bugaevc:
It says "must", not "may" or "should". It's not capital-case "MUST", but we should still abide.

alimpfard:
we can't really do that without dropping down a layer though, we'd either have to start with the interface IP set to zero, or make custom ethernet packets.

In any case, could you please try with this patch:
```diff
diff --git a/Userland/Services/DHCPClient/DHCPv4Client.cpp b/Userland/Services/DHCPClient/DHCPv4Client.cpp
index 61ee0c60f..6c427cbdc 100644
--- a/Userland/Services/DHCPClient/DHCPv4Client.cpp
+++ b/Userland/Services/DHCPClient/DHCPv4Client.cpp
@@ -286,6 +286,8 @@ void DHCPv4Client::dhcp_request(DHCPv4Transaction& transaction, const DHCPv4Pack

     // set packet options
     builder.set_message_type(DHCPMessageType::DHCPRequest);
+    builder.add_option(DHCPOption::ServerIdentifier, sizeof(IPv4Address), &offer.siaddr());
+    builder.add_option(DHCPOption::RequestedIPAddress, sizeof(IPv4Address), &offer.yiaddr());
     auto& dhcp_packet = builder.build();

     // broadcast the "request" request
```

bugaevc:
> we can't really do that without dropping down a layer though, we'd either have to start with the interface IP set to zero, or make custom ethernet packets.

Normally you'd do that by using `IP_PKTINFO` and specifying any source IP address you want (INADDR_ANY in this case).

alimpfard:
Seems like even `IP_PKTINFO` can't set it to zero:

> If IP_PKTINFO is passed to sendmsg(2) **and ipi_spec_dst is not zero**, then it is used as the local source address for the routing table lookup and for setting up IP source route options.

```
 IP_PKTINFO (since Linux 2.2)
              Pass an IP_PKTINFO ancillary message that contains a
              pktinfo structure that supplies some information about the
              incoming packet.  This works only for datagram oriented
              sockets.  The argument is a flag that tells the socket
              whether the IP_PKTINFO message should be passed or not.
              The message itself can be sent/retrieved only as a control
              message with a packet using recvmsg(2) or sendmsg(2).

                  struct in_pktinfo {
                      unsigned int   ipi_ifindex;  /* Interface index */
                      struct in_addr ipi_spec_dst; /* Local address */
                      struct in_addr ipi_addr;     /* Header Destination
                                                      address */
                  };

              ipi_ifindex is the unique index of the interface the
              packet was received on.  ipi_spec_dst is the local address
              of the packet and ipi_addr is the destination address in
              the packet header.  If IP_PKTINFO is passed to sendmsg(2)
              and ipi_spec_dst is not zero, then it is used as the local
              source address for the routing table lookup and for
              setting up IP source route options.  When ipi_ifindex is
              not zero, the primary local address of the interface
              specified by the index overwrites ipi_spec_dst for the
              routing table lookup.
```

bugaevc:
> In any case, could you please try with this patch

That totally worked! :tada:

supercomputer7:
So that's why I didn't have DHCP working with the [RTL8169 patch](https://github.com/SerenityOS/serenity/pull/5069)?
If that's so, I should try to test this patch with it and see what happens then.

PR #5369, opened 2021-02-16
Run, LibGUI: Run command history + ComboBox list window size fixes

nvella:
Rolling these two commits into one PR as I came across the interesting `ComboBox` behaviour whilst implementing the Run history, and more or less worked on the two simultaneously. Can split into separate PRs if desired :)

- Run now stores successfully ran command history in a simple text file under `.config/RunHistory.txt` and presents it as a `ComboBox`. The last 25 successfully ran inputs are stored, with the newest entries being inserted at the top. If the command ran is already in the history, it is moved to the top.
- `LibGUI::ComboBox` now explicitly sets a minimum list window size using the new mechanism established in #5360. The previous behaviour was a minimum list window size of 50px, even though the inner list widget may've not been that tall - this resulted in unpainted area in the list window. The minimum height is now set to the total height of all the items in the list, or the height of three items, whichever is smaller. This calculation for minimum size, rather than just setting to `{1, 1}`, is needed to prevent the list window from becoming infinitesimally small when opened near the screen edge. 😁

### DIFF
--- a/Userland/Applications/Run/RunWindow.cpp
+++ b/Userland/Applications/Run/RunWindow.cpp
@@ -25,10 +25,12 @@
  */
 
 #include "RunWindow.h"
+#include <AK/LexicalPath.h>
 #include <AK/URL.h>
 #include <AK/URLParser.h>
 #include <Applications/Run/RunGML.h>
 #include <LibCore/File.h>
+#include <LibCore/StandardPaths.h>
 #include <LibDesktop/Launcher.h>
 #include <LibGUI/Button.h>
 #include <LibGUI/Event.h>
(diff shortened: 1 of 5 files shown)


Userland/Applications/Run/RunWindow.cpp
```diff
@@ -168,3 +180,35 @@ bool RunWindow::run_via_launch(const String& run_input)
 
     return true;
 }
+
+String RunWindow::history_file_path()
+{
+    return AK::LexicalPath::canonicalized_path(String::formatted("{}/{}", Core::StandardPaths::config_directory(), "RunHistory.txt"));
```
awesomekling, line 186: No need to qualify things with the `AK::` prefix

nvella: Not sure why it never dawned on me that _everything_ in `AK` is in the global namespace, considering how often I use `String`s and `Vector`s without qualifying the namespace 😄 Will give everything a look-over tonight (7/8 hrs or so) 👍

PR #5370, opened 2021-02-16
LibGUI: Improve IconView rubberband performance

tomuta:
Rather than invalidating the entire window, which is very expensive on
the transparent desktop widget, just invalidate the areas that actually
need updating.

_This reduces CPU usage by more than 50%_

_Before:_
iconview_performance_before

_After:_
iconview_performance


tomuta:
@awesomekling When I add a weird hack like this https://github.com/tomuta/serenity/commit/1b486a891619f68e43cfecc1ed97dd0634ba48e7 where I essentially keep "dirty rects" in sync with the front/back buffer then it dramatically helps (though still not perfect, but that may be another bug). Of course it feels like this shouldn't be needed...

tomuta:
Looks like the alpha fix in LibGUI's `Window` ended up fixing this problem...

PR #5381, opened 2021-02-16
LibJS: Implement Array.prototype.flat

kessejones:


Userland/Libraries/LibJS/Runtime/ArrayPrototype.cpp
```diff
@@ -1034,4 +1035,42 @@ JS_DEFINE_NATIVE_FUNCTION(ArrayPrototype::values)
...
+            if (depth_argument.is_negative_infinity()) {
+                depth = 0;
+            } else if (depth_argument.is_positive_infinity()) {
+                is_infinity = true;
+            } else {
+                auto argumento_depth = depth_argument.to_i32(global_object);
+                depth = max(argumento_depth, 0);
+            }
+        }
+    }
```
linusg, line 1071: This not spec compliant, and for no good reason :)

```
Let depthNum be 1.
If depth is not undefined, then
    Set depthNum to ? ToIntegerOrInfinity(depth).
    If depthNum < 0, set depthNum to 0.
```

- `if (vm.argument_count() == 1)` should not exist here, if you call the function with two or more arguments this breaks
- `depth_argument.is_negative_infinity()` doesn't take into account that `to_integer_or_infinity()` would coerce things like `"Infinity"` to +Infinity, but is not used here

Userland/Libraries/LibJS/Runtime/ArrayPrototype.cpp
```diff
@@ -1034,4 +1035,42 @@ JS_DEFINE_NATIVE_FUNCTION(ArrayPrototype::values)
     return ArrayIterator::create(global_object, this_object, Object::PropertyKind::Value);
 }
 
+static void recursive_array_flat(GlobalObject& global_object, JS::Array& new_array, JS::Array& array, i32 depth, bool is_infinity)
```
linusg, line 1038: No need for `JS::` namespace prefix here, `array` should be an `Object` (e.g. `array_like`)

Userland/Libraries/LibJS/Runtime/ArrayPrototype.cpp
```diff
@@ -1034,4 +1035,42 @@ JS_DEFINE_NATIVE_FUNCTION(ArrayPrototype::values)
...
+{
+    auto array_length = length_of_array_like(global_object, array);
```
linusg, line 1040: Missing exception check (considering the above comment)

Userland/Libraries/LibJS/Runtime/ArrayPrototype.cpp
```diff
@@ -1034,4 +1035,42 @@ JS_DEFINE_NATIVE_FUNCTION(ArrayPrototype::values)
...
+    for (size_t j = 0; j < array_length; ++j) {
+        auto value = array.get(j);
```
linusg, line 1042: Missing exception check (considering the above)

linusg (changes requested):

Userland/Libraries/LibJS/Runtime/ArrayPrototype.cpp
brendalf: While making the changes you've suggested. We realized that `Value::to_integer_or_infinity` returns -9223372036854775808.000000 for both `-Infinity` and `Infinity`, but the `flat` has to return different results for each case.

If depth is equals to Infinity, the function needs to flat all arrays and if depth is equals to -Infinity, the function needs to return a copy of the array without any changes (`depth = 0`).

linusg: > We realized that Value::to_integer_or_infinity returns -9223372036854775808.000000 for both -Infinity and Infinity

`to_integer_or_infinity()` is working just fine, that's a `dbgln()` bug :)

Proof:

```diff
--- a/Userland/Libraries/LibJS/Runtime/GlobalObject.cpp
+++ b/Userland/Libraries/LibJS/Runtime/GlobalObject.cpp
@@ -119,6 +119,7 @@ void GlobalObject::initialize()
     define_native_function(vm.names.isFinite, is_finite, 1, attr);
     define_native_function(vm.names.parseFloat, parse_float, 1, attr);
     define_native_function(vm.names.parseInt, parse_int, 1, attr);
+    define_native_function("toIntegerOrInfinity", to_integer_or_infinity, 1, attr);
 
     define_property(vm.names.NaN, js_nan(), 0);
     define_property(vm.names.Infinity, js_infinity(), 0);
@@ -286,6 +287,15 @@ JS_DEFINE_NATIVE_FUNCTION(GlobalObject::parse_int)
     return Value(sign * number);
 }
 
+JS_DEFINE_NATIVE_FUNCTION(GlobalObject::to_integer_or_infinity)
+{
+    auto integer_or_infinity = vm.argument(0).to_integer_or_infinity(global_object);
+    if (vm.exception())
+        return {};
+    dbgln("{}", integer_or_infinity);
+    return Value(integer_or_infinity);
+}
+
 Optional<Variable> GlobalObject::get_from_scope(const FlyString& name) const
 {
     auto value = get(name);
--- a/Userland/Libraries/LibJS/Runtime/GlobalObject.h
+++ b/Userland/Libraries/LibJS/Runtime/GlobalObject.h
@@ -81,6 +81,7 @@ private:
     JS_DECLARE_NATIVE_FUNCTION(is_finite);
     JS_DECLARE_NATIVE_FUNCTION(parse_float);
     JS_DECLARE_NATIVE_FUNCTION(parse_int);
+    JS_DECLARE_NATIVE_FUNCTION(to_integer_or_infinity);
 
     NonnullOwnPtr<Console> m_console;
```

```
> toIntegerOrInfinity(-Infinity)
-9223372036854775808.000000
-Infinity
> toIntegerOrInfinity(Infinity)
-9223372036854775808.000000
Infinity
```

linusg (approved):
Nice!

PR #5394, opened 2021-02-17
WindowServer: Apply the backing bitmap's scale when alpha hit-testing

### LINKED ISSUE #5390: WindowServer: window titlebar mouse dead zone on HighDPI
I don't really know how to explain this issue properly. Corners and the top half of titlebar is not clickable on HighDPI and it passes through to the window/desktop under it. It also makes titlebar buttons kinda impossible to press unless you put the mouse on the bottom half of the titlebar. Other controls of window like buttons or the menubars work without problems

https://user-images.githubusercontent.com/13377926/108271330-ed831a00-7170-11eb-81aa-9d3f6bc0b7ef.mov

tomuta:
Fixes #5390

### DIFF
--- a/Userland/Services/WindowServer/Window.cpp
+++ b/Userland/Services/WindowServer/Window.cpp
@@ -916,7 +916,8 @@ bool Window::hit_test(const Gfx::IntPoint& point, bool include_frame) const
     u8 threshold = alpha_hit_threshold() * 255;
     if (threshold == 0 || !m_backing_store || !m_backing_store->has_alpha_channel())
         return true;
-    auto color = m_backing_store->get_pixel(point.translated(-rect().location()));
+    auto relative_point = point.translated(-rect().location()) * m_backing_store->scale();
+    auto color = m_backing_store->get_pixel(relative_point);
     return color.alpha() >= threshold;
 }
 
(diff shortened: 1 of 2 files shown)


Userland/Services/WindowServer/Window.cpp
```diff
@@ -916,7 +916,8 @@ bool Window::hit_test(const Gfx::IntPoint& point, bool include_frame) const
     u8 threshold = alpha_hit_threshold() * 255;
     if (threshold == 0 || !m_backing_store || !m_backing_store->has_alpha_channel())
         return true;
-    auto color = m_backing_store->get_pixel(point.translated(-rect().location()));
+    auto relative_point = point.translated(-rect().location()) * Screen::the().scale_factor();
```
nico, line 919: afaik no windows have scaled backing stores yet. i'd expect this to do the wrong thing. if you want to future-proof, you can get the scale factor off m_backing_store itself though (`m_backing_store->scale()`).

tomuta: Hm good point. I was also wondering why Bitmap::get_pixel didn't use the scale factor it had.

nico: Maybe it should! My thinking was that it returns a single pixel, so it kind of has to be pixel-based. Maybe we should have a "average_rect_sample()" or what that gets a sample rect size defaulting to 1x1 and that could then work in logical coords and return the 2x2 avg on 2x bitmaps.

PR #5396, opened 2021-02-18
Kernel: Gracefully handle kernel stack overflow

### LINKED ISSUE #67: Kernel: dump_backtrace() causes triple fault in ELF symbolication
We don't know what's happening yet, but not calling dump_backtrace() avoids triple faulting.
This seems to happen somewhere in the ELF symbolication code.

Petelliott:
- [x] Add guard page to the bottom of thread kernel stacks.
- [ ] Handle double fault when kernel stack overflows

closes #67

### DIFF
--- a/Kernel/VM/MemoryManager.cpp
+++ b/Kernel/VM/MemoryManager.cpp
@@ -479,11 +479,11 @@ OwnPtr<Region> MemoryManager::allocate_contiguous_kernel_region(size_t size, Str
     return allocate_kernel_region_with_vmobject(range.value(), vmobject, move(name), access, cacheable);
 }
 
-OwnPtr<Region> MemoryManager::allocate_kernel_region(size_t size, String name, u8 access, AllocationStrategy strategy, Region::Cacheable cacheable)
+OwnPtr<Region> MemoryManager::allocate_kernel_region(size_t size, String name, u8 access, AllocationStrategy strategy, Region::Cacheable cacheable, RangeAllocator::GuardSide guard_side)
 {
     ASSERT(!(size % PAGE_SIZE));
     ScopedSpinLock lock(s_mm_lock);
-    auto range = kernel_page_directory().range_allocator().allocate_anywhere(size);
+    auto range = kernel_page_directory().range_allocator().allocate_anywhere(size, PAGE_SIZE, guard_side);
     if (!range.has_value())
         return {};
     auto vmobject = AnonymousVMObject::create_with_size(size, strategy);
--- a/Kernel/VM/MemoryManager.cpp
+++ b/Kernel/VM/MemoryManager.cpp
@@ -82,7 +82,7 @@ bool MemoryManager::is_initialized()
 MemoryManager::MemoryManager()
 {
     ScopedSpinLock lock(s_mm_lock);
-    m_kernel_page_directory = PageDirectory::create_kernel_page_directory();
+    m_kernel_page_directory = &PageDirectory::kernel_page_directory();
     parse_memory_map();
     write_cr3(kernel_page_directory().cr3());
     protect_kernel_image();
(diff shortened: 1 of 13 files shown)


tomuta:
Hm why add a `GuardVMObject`? Can't we just expose what `mprotect` just mark the first page of the kernel stack region appropriately?

Petelliott:
> Hm why add a `GuardVMObject`? Can't we just expose what `mprotect` does and just mark the first page of the kernel stack region appropriately?

`mprotect` relies on `Space`, which only applies to userspace regions. space does basically the same thing, and splits the region into smaller regions. Since we have to manually split the range anyways, I figured that we might as well give a better error message and region names, rather than just setting an anonymous memory region to no-write/no-read/no-exec

tomuta:
What I meant with "expose what `mprotect` is: Expose the functionality as a kernel api so we can accomplish the same on kernel regions. There is really no need to have multiple VM objects to do this.  This isn't needed on neither Windows nor Linux, both of which simply mark the first page with no access using `mprotect` or `VirtualProtect(Ex)`, or a kernel api for use by the kernel.

Petelliott:
You are right. I misunderstood the relation between `Region` and `VMObject`

Kernel/VM/MemoryManager.cpp
```diff
@@ -537,6 +542,50 @@ OwnPtr<Region> MemoryManager::allocate_kernel_region_with_vmobject(VMObject& vmo
...
+        return {};
+    auto vmobject = maybe_vmobject.release_nonnull();
+
+    Range real_range = range.value();
+
+    OwnPtr<Region> low_region;
+    if (side == GuardSide::Both || side == GuardSide::Low) {
+        Range low_guard = { real_range.base(), PAGE_SIZE };
+        real_range = real_range.carve(low_guard)[0];
+        low_region = allocate_kernel_region_with_vmobject(low_guard, 0, vmobject, String::formatted("{} lower guard page", name), 0, cachable);
```
tomuta, line 571: All it should take is mark a single page at the beginning or end of the region with no-access permissions. This is a lot of unnecessary overhead for such a simple operation. We really don't want to maintain child regions.

tomuta (changes requested):

Kernel/VM/MemoryManager.cpp
Petelliott: Are you suggesting that Region should just support non-accessible pages at the beginning specifically for kernel stacks? you kernel developers never let me have any fun :^)

tomuta: No, that's not what I suggest. There doesn't really have to be any additional accounting for supporting this. All it takes is marking a PTE in a kernel region as inaccessible.

The reason I'm pushing back is that `Region` as it is is already heavier than it should be. Same technically goes for `AnonymousVMObject` and there really isn't any reason to add more objects to account for merely a few bits for a page table entry. Something we already implement for `mprotect` anyway.

tomuta: I think I see why you're doing this. Looking at `mprotect` it looks like we're doing a lot more than what we should have to, splitting regions etc. All we really should have to do is modifying the PTEs accordingly. Maybe for now doing something similar does make sense, but eventually I think we should get away from all that extra overhead that shouldn't be needed.

awesomekling: For userspace VM, this is done by `RangeAllocator`. When generating a new VM range, we simply expand the range by one page in both directions and return the unpadded range to the allocator's client. Can we use the same approach here?

tomuta: @awesomekling But why split regions? Shouldn't we just take a region as-is and just modify the PTEs in question? AFAIK that is what Linux and Windows do. They do not create sub-regions or split them in any way. What if you were to do a checker pattern where you want to `mprotect` every second page? Example would be a heap debugger that would place allocations at the beginning or ending of a page and mark the adjacent page as no-access, with the goal to find a under/overflow. Would you want to create a region for each and every single page? On Linux/Windows this doesn't incur any additional overhead, the memory manager simply changes the permissions on the PTE for those pages.

awesomekling: I'm not super familiar with Linux internals but it looks to me like they split VMA's on mprotect/madvise/munmap/etc.

My general idea with `Region` (and `VMObject`) was to have an abstraction instead of tying ourselves directly to x86 page table layouts. Since it's pretty likely we'll want to do a port to something else some day. But maybe there's a better approach that's still port-friendly?

tomuta: Ah I guess you're right, on Linux this is probably `struct vm_area_struct`.

Petelliott: @awesomekling looks like all range allocations are actually guarded by default. I'll change the API though so that an individual requester can ensure that a range is guarded.

PR #5400, opened 2021-02-18
LibC: Remove serenity.h.rej

linusg:


tomuta:
Whoops... The joys of manually resolving conflicts when files are are moved around... ;)

PR #5403, opened 2021-02-18
DisplaySettings: Show revert dialog only for resolution/dpi changes.

belginul:
There's no point in showing the revert dialog for strictly wallpaper changes.


tomuta:
I've been meaning to write this up as a bug for a while now...

PR #5404, opened 2021-02-18
UserspaceEmulator: SoftCPU: Add support for several x86 instructions

bcoles:
Adds:

* FLDL2T
* FLDLG2
* FLDLN2
* FRNDINT
* FCOS


awesomekling:
I think that existing code looks right and your new functions look weird.
The instruction's RM byte contains an FPU register index in the lower 3 bits, that's what we're masking off with `&7`. We shouldn't be using the instruction's condition code to mask the RM byte.

PR #5411, opened 2021-02-19
Toolchain: Use -ftls-model=initial-exec by default

### LINKED ISSUE #5366: Ports: Trying to compile non-existant file with gcc provides the wrong error number and also increments it
For example:
Screenshot from 2021-02-16 10-31-16

And eventually:
Screenshot from 2021-02-16 10-30-44

itamar8910:
Closes #5366


awesomekling:
Does that mean we can remove it from the root CMakeLists.txt?

itamar8910:
> Does that mean we can remove it from the root CMakeLists.txt?

It does yeah, I forgot to commit that change :sweat_smile:

awesomekling:
Thanks so much @itamar8910 for tracking that down :)

PR #5414, opened 2021-02-19
LibWeb: Add DOMExceptions

linusg:
image


Userland/Libraries/LibWeb/CodeGenerators/WrapperGenerator.cpp
```diff
@@ -1401,6 +1410,80 @@ JS_DEFINE_NATIVE_SETTER(@prototype_class@::@attribute.setter_callback@)
...
+        // And yet this blows up when that's not the case and this branch is unused:
+        if (result.is_exception())
+            ...
+    } else {
+        // Regular return value with no exception
+    }
+
+So we use template magic instead and let the compiler figure this out. The unspecialized templates won't compile as a safeguard.
+FIXME: This is not the way.
+*/
```
linusg, line 1430: Need help from a C++ wizard to make this nice, it's way outside my comfort zone already :mage:

alimpfard: Not a C++ wizard, but...
where exactly is this snippet supposed to go?

linusg: This snippet is the essence of the generated code, starting at line 1501 :)

alimpfard: [This patch](https://gist.github.com/alimpfard/652b05a05a8da76601a7664582e7a276) is the only "better" thing I can come up with.

`if constexpr` doesn't do what you think it does inside non-templated entities, so you have to always move it inside a template to get the "let me call nonexistent members in the false branch" effect.

The fact that the method call can return `void` really messes it up (good thinking on wrapping it in a lambda to delay execution!).

linusg: > if constexpr doesn't do what you think it does inside non-templated entities, so you have to always move it inside a template to get the "let me call nonexistent members in the false branch" effect.

I knew it's supposed to *somehow* do that, thanks for clarifying!

> The fact that the method call can return void really messes it up

yup...

> good thinking on wrapping it in a lambda to delay execution!

I'll pass that on to whoever had this idea on Stack Overflow :smile:

This looks great, thank you! Currently rebuilding a toolchain, will report back...

linusg: Works like a charm (not that I expected otherwise!)

PR #5418, opened 2021-02-20
Ports: MilkyTracker v1.03 port

TheMorc:
Before i made my pt2-clone port I had a "dream" to see MilkyTracker working on Serenity. And even if i was able to get MilkyTracker to compile and sorta run, it was just more crashing than working back then. So I just gave up for a while and made the pt2-clone port.
But I came back to it and I think that @danboid will be excited to see this.

The port now works, plays, saves settings all as it should.
It's _nearly_ flawless, doesn't have any weird issues like pthread spam, broken settings, instant continuous backup crashes etc, but the only problem now is that the UI isn't refreshing and stays still, but it atleast plays now.

Screen Shot 2021-02-20 at 02 34 11

There are still a few things i want to improve (patches and old mistakes from before) so i'm drafting it for now.

Edit: there's a bit of noticable crackling, but it was also present on the pt2-clone

### DIFF
--- a/Ports/milkytracker/patches/DisplayDevice_SDL.patch
+++ b/Ports/milkytracker/patches/DisplayDevice_SDL.patch
@@ -1,0 +1,73 @@
+--- MilkyTracker_orig/src/ppui/sdl/DisplayDevice_SDL.cpp	2020-12-09 23:58:14.000000000 +0100
++++ MilkyTracker-1.03.00/src/ppui/sdl/DisplayDevice_SDL.cpp	2021-02-20 10:24:50.816107401 +0100
+@@ -26,30 +26,8 @@
+ SDL_Window* PPDisplayDevice::CreateWindow(pp_int32& w, pp_int32& h, pp_int32& bpp, Uint32 flags)
+ {
+ 	size_t namelen = 0;
+-	char rendername[256] = { 0 };
+-	PFNGLGETSTRINGPROC glGetStringAPI = NULL;
+-
+-	for (int it = 0; it < SDL_GetNumRenderDrivers(); it++)
+-	{
+-		SDL_RendererInfo info;
+-		SDL_GetRenderDriverInfo(it, &info);
+-
+-		namelen += strlen(info.name) + 1;
+-		strncat(rendername, info.name, sizeof(rendername) - namelen);
+-		strncat(rendername, " ", sizeof(rendername) - namelen);
+-
+-		if (strncmp("opengles2", info.name, 9) == 0)
+-		{
+-			drv_index = it;
+-			SDL_GL_SetAttribute(SDL_GL_CONTEXT_PROFILE_MASK, SDL_GL_CONTEXT_PROFILE_ES);
+-			SDL_GL_SetAttribute(SDL_GL_CONTEXT_MAJOR_VERSION, 2);
+-			SDL_GL_SetAttribute(SDL_GL_CONTEXT_MINOR_VERSION, 0);
+-			SDL_GL_SetAttribute(SDL_GL_ACCELERATED_VISUAL, 1);
+-		}
+-	}
+-
+ 	// Create SDL window
+-	SDL_Window* theWindow = SDL_CreateWindow("MilkyTracker", SDL_WINDOWPOS_UNDEFINED, SDL_WINDOWPOS_UNDEFINED, w, h, SDL_WINDOW_OPENGL | flags);
++	SDL_Window* theWindow = SDL_CreateWindow("MilkyTracker", SDL_WINDOWPOS_UNDEFINED, SDL_WINDOWPOS_UNDEFINED, w, h, flags);
+ 
+ 	if (theWindow == NULL)
+ 	{
+@@ -59,7 +37,7 @@
+ 		w = getDefaultWidth();
+ 		h = getDefaultHeight();
+ 		
+-		theWindow = SDL_CreateWindow("MilkyTracker", SDL_WINDOWPOS_UNDEFINED, SDL_WINDOWPOS_UNDEFINED, w, h, SDL_WINDOW_OPENGL | flags);
++		theWindow = SDL_CreateWindow("MilkyTracker", SDL_WINDOWPOS_UNDEFINED, SDL_WINDOWPOS_UNDEFINED, w, h, flags);
+ 		
+ 		if (theWindow == NULL)
+ 		{
+@@ -69,21 +47,6 @@
+ 		}
+ 	}
+ 
+-	SDL_GLContext ctx = SDL_GL_CreateContext(theWindow);
+-	SDL_GL_MakeCurrent(theWindow, ctx);
+-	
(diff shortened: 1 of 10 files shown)


danboid:
Excellent news!

tomuta:
The crackling is probably our SB16 driver's fault. It's using a one-shot dma that is documented to be unsuitable for smooth playback...

Ports/milkytracker/patches/DisplayDevice_SDL.patch
```diff
@@ -0,0 +1,73 @@
+--- MilkyTracker_orig/src/ppui/sdl/DisplayDevice_SDL.cpp	2020-12-09 23:58:14.000000000 +0100
```
emanuele6, line 1: Shouldn't these files be named `.cpp.patch` instead of just `.patch`?

emanuele6: Ah, it doesn't matter. nevermind.

TheMorc: I was just using `diff -u <original> <patched>` to compare originals with modified files and it seems like `patch` doesn't care about this line and modifies the correct file

TheMorc:
So i'm still not sure why it doesn't update, the same patched codebase works just fine on Linux so the problem could be somewhere in Serenity or in the SDL2 port. I'll leave it as is for now, maybe someone can figure it out.

danboid:
I wanted to see if I could get this to build but I obviously still no nothing about git:

```
git clone https://github.com/TheMorc/serenity.git
cd serenity/
git checkout -b milkytracker
```

but there is no milktracker dir in Ports?

linusg:
@danboid `-b` creates a new branch - now do:

```console
$ git checkout master
Switched to branch 'master'
Your branch is up to date with 'origin/master'.
$ git branch -D milkytracker
Deleted branch milkytracker (was a525d0271).
$ git checkout milkytracker
Branch 'milkytracker' set up to track remote branch 'milkytracker' from 'origin'.
Switched to a new branch 'milkytracker'
$ ls -la Ports/milkytracker
...
```

danboid:
Thanks Linus!

I've successfully built this port and got it to play mods. I tried adjusting various settings (scrolling, scopes, editing mode etc) but had no luck getting it to scroll during playback as @TheMorc says.

One interesting thing to note is that the file browser in Milkytracker does not suffer the same delays as pt2-clone does, but pt2-clone does scroll fine during playback :|

danboid:
@TheMorc  - have you already tried disabling SDL_RENDERER_TARGETTEXTURE in src/ppui/sdl/DisplayDeviceFB_SDL.cpp ?

TheMorc:
> @TheMorc - have you already tried disabling SDL_RENDERER_TARGETTEXTURE in src/ppui/sdl/DisplayDeviceFB_SDL.cpp ?

I don't remember actually to be honest.

danboid:
I've not looked into it so I dunno if this is something that would need to be disabled in SDL but I did notice:

SDL: Renderer supports rendering to texture

printed to the console when starting Milktracker so maybe if we disable this it may fix the display update?

metalvoidzz:
Hi, I've looked into porting MilkyTracker some time ago too, and ran into the same refresh issue. My theory is that it involves the SDL_Timer callback, as seen in `SDL_Main.cpp`.

In SDL, timer events are implemented with a loop in a separate thread. That loop depends on the LibC Semaphore functions, which are just stubbed out at the moment.

TheMorc:
> Hi, I've looked into porting MilkyTracker some time ago too, and ran into the same refresh issue. My theory is that it involves the SDL_Timer callback, as seen in `SDL_Main.cpp`.
> 
> In SDL, timer events are implemented with a loop in a separate thread. That loop depends on the LibC Semaphore functions, which are just stubbed out at the moment.

Yeah, i was thinking that it also could be an issue with timers. Nothing that I can do with because I really don't want to mess with kernel stuff.

metalvoidzz:
I'm thinking we could circumvent this issue for now by using `LibCore::Timer` instead, either in SDL2 or MilkyTracker directly.

Ports/milkytracker/patches/tracker-CMakeLists.patch
```diff
@@ -0,0 +1,10 @@
+--- MilkyTracker-1.03.00d/src/tracker/CMakeLists.txt	2021-01-04 15:49:14.173940007 +0100
++++ MilkyTracker-1.03.00/src/tracker/CMakeLists.txt	2021-01-04 15:48:03.179731481 +0100
+@@ -374,4 +375,7 @@
+     set(INSTALL_DEST ${CMAKE_INSTALL_BINDIR})
+ endif()
+ 
++target_link_libraries(tracker -L../../SDL2/SDL-master-serenity/ -lSDL2 -lgui -lipc -lm)
```
awesomekling, line 7: We should link with the libSDL in the `Build/Root/` subtree, not with the one from the `Ports/SDL2/` subtree.

Ports/milkytracker/patches/tracker-CMakeLists.patch
```diff
@@ -0,0 +1,10 @@
...
++set(INSTALL_DEST ../../../../../../usr/bin/)
```
awesomekling, line 8: This looks highly suspicious. Is this not my host system's `/usr/bin`?

Ports/milkytracker/patches/SampleEditor-fabsf.patch
```diff
@@ -0,0 +1,11 @@
+--- MilkyTracker-1.03.00/src/tracker/SampleEditor.cpp	2020-12-09 23:58:14.000000000 +0100
++++ SampleEditor.cpp	2021-01-07 21:31:05.830204543 +0100
+@@ -1620,7 +1620,7 @@
+ 
+ 			f = (1.0f-frac)*f1 + frac*f2;
+ 
+-			step = powf(16.0f,fabsf(fi));
++			step = powf(16.0f,fi);
```
awesomekling, line 8: `fabsf()` is available in LibM so there's no need for this patch. :)

Ports/milkytracker/patches/SDL_Main-PATH_MAX.patch
```diff
@@ -0,0 +1,10 @@
+--- MilkyTracker-1.03.00/src/tracker/sdl/SDL_Main.cpp	2020-12-09 23:58:14.000000000 +0100
++++ MilkyTracker-1.03.00DD/src/tracker/sdl/SDL_Main.cpp	2021-01-04 15:26:23.821172703 +0100
+@@ -51,6 +51,7 @@
+  *    - Fix for french azerty keyboards (their number keys are shifted)
+  *
+  */
++#define PATH_MAX 4096
```
awesomekling, line 7: To get `PATH_MAX`, the correct fix is to `#include <limits.h>`

linusg:
@TheMorc do you intend to make any changes to address the review comments above?

awesomekling:
Superseded by #6685. Closing this since it appears to be abandoned

PR #5423, opened 2021-02-20
LibWeb: Implement Window.prompt()

linusg:


Userland/Libraries/LibGUI/InputBox.h
```diff
@@ -38,7 +38,7 @@ class InputBox : public Dialog {
     static int show(String& text_value, Window* parent_window, const StringView& prompt, const StringView& title);
 
 private:
-    explicit InputBox(Window* parent_window, const StringView& prompt, const StringView& title);
+    explicit InputBox(String& text_value, Window* parent_window, const StringView& prompt, const StringView& title);
```
awesomekling, line 41: Let's keep the parent window as the first argument to stay consistent with other dialog classes

PR #5424, opened 2021-02-20
LanguageServers/Cpp: Add 'FindDeclaration' capability

itamar8910:
The C++ LanguageServer can now find the matching declaration for variable names, function calls, struct/class types and properties.

When clicking on one of the above with Ctrl pressed, HackStudio will ask the language server to find a matching declaration, and navigate to the result in the Editor

Example usage:

goto_declaration

### DIFF
--- a/Userland/DevTools/HackStudio/Editor.cpp
+++ b/Userland/DevTools/HackStudio/Editor.cpp
@@ -1,5 +1,6 @@
 /*
  * Copyright (c) 2018-2021, Andreas Kling <kling@serenityos.org>
+ * 2018-2021, the SerenityOS developers
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
@@ -153,9 +154,7 @@ void Editor::show_documentation_tooltip_if_available(const String& hovered_token
 {
     auto it = man_paths().find(hovered_token);
     if (it == man_paths().end()) {
-#if EDITOR_DEBUG
-        dbgln("no man path for {}", hovered_token);
-#endif
+        dbgln_if(EDITOR_DEBUG, "no man path for {}", hovered_token);
         m_documentation_tooltip_window->hide();
         return;
     }
@@ -164,9 +163,7 @@ void Editor::show_documentation_tooltip_if_available(const String& hovered_token
         return;
     }
 
-#if EDITOR_DEBUG
-    dbgln("opening {}", it->value);
-#endif
+    dbgln(EDITOR_DEBUG, "opening {}", it->value);
     auto file = Core::File::construct(it->value);
     if (!file->open(Core::File::ReadOnly)) {
         dbgln("failed to open {}, {}", it->value, file->error_string());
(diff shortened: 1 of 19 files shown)


Userland/DevTools/HackStudio/Editor.cpp
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2018-2020, Andreas Kling <kling@serenityos.org>
+ * Copyright (c) 2018-2021, Andreas Kling <kling@serenityos.org>
```
awesomekling, line 2: 🤔

Userland/DevTools/HackStudio/Editor.cpp
```diff
@@ -525,4 +517,16 @@ void Editor::flush_file_content_to_langauge_server()
...
+
+void Editor::on_navigatable_link_click(const GUI::TextDocumentSpan& span)
+{
+    auto adjusted_range = span.range;
+    adjusted_range.end().set_column(adjusted_range.end().column() + 1);
+    auto span_text = document().text_in_range(adjusted_range);
+    auto header_path = span_text.substring(1, span_text.length() - 2);
+#if EDITOR_DEBUG
+    dbgln("Ctrl+click: {} \"{}\"", adjusted_range, header_path);
+#endif
```
awesomekling, line 532: Please use `dbgln_if(EDITOR_DEBUG, ...)`

Userland/DevTools/HackStudio/HackStudioWidget.cpp
```diff
@@ -210,8 +210,13 @@ Vector<String> HackStudioWidget::selected_file_names() const
     return files;
 }
 
-void HackStudioWidget::open_file(const String& filename)
+void HackStudioWidget::open_file(const String& full_filename)
 {
+    String filename = full_filename;
+    if (full_filename.starts_with(project().root_path())) {
+        filename = full_filename.substring(LexicalPath { project().root_path() }.string().length() + 1);
```
awesomekling, line 217: Can we create a helper for this transformation and call it something helpful so you understand what's going on here immediately? :)

Userland/Libraries/LibCpp/AST.h
```diff
@@ -60,6 +61,10 @@ class ASTNode : public RefCounted<ASTNode> {
         ASSERT(m_end.has_value());
         return m_end.value();
     }
+    StringView filename() const
+    {
+        return m_filename.view();
+    }
```
awesomekling, line 67: If we're already storing a `String` we might as well return a `const String&` here, no point in only exposing a `StringView`.

Userland/DevTools/HackStudio/LanguageServers/Cpp/ParserAutoComplete.cpp
```diff
@@ -35,7 +35,7 @@
 
 // #define AUTOCOMPLETE_VERBOSE
 
-#ifdef AUTOCOMPLETE_VERBOSE
+#ifdef CPP_LANGUAGE_SERVER_DEBUG
 #    define VERBOSE(fmt, ...) dbgln(fmt, ##__VA_ARGS__)
 #else
 #    define VERBOSE(fmt, ...) \
```
awesomekling, line 41: Let's get rid of the `VERBOSE` business and use `dbgln_if(` :)

Userland/Libraries/LibCpp/AST.h
itamar8910: Well we store`m_filename` as a `FlyString` (to avoid having copies of the same filename in all of the `ASTNode`s). Is there an alternative to exposing a `StringView` here?

awesomekling: Sure, return a `const FlyString&` :)

Userland/DevTools/HackStudio/HackStudioWidget.cpp
itamar8910: Sure, added it as `LexicalPath::relative_path()`

PR #5425, opened 2021-02-20
LibWeb: Move ExceptionOr bindings utils into own header

linusg:


awesomekling:
Downright tasteful :)

Issue #4979, opened 2021-01-16
SpaceAnalyzer: Context menu


linusg:
It would be nice if Space Analyzer would have a context menu for each entry (file or directory), allowing to either `Delete` it or `Open in File Manager`, for example..

GNOME's Disk Usage Analyzer for reference:

![image](https://user-images.githubusercontent.com/19366641/104819669-40df1100-582f-11eb-864f-4cfc2c41e9e3.png)

linusg:
This was implemented in #5428.

PR #5428, opened 2021-02-20
SpaceAnalyzer: Added context menu for nodes

speles:
Currently supports 3 actions: Open, Copy Path, and Delete (as described in  #4979).
Screenshot_20210220_175347


RealKC:
Consider using the `Core::File::remove` introduced by #5444 instead of duplicating the logic yourself

speles:
Updated to use `Core::File::remove`

PR #5429, opened 2021-02-20
WindowServer: Add shortcut to quit windows.

daubaris:
An addition of simple Ctrl-Q for quitting windows.

### DIFF
--- a/Userland/Services/WindowServer/WindowManager.cpp
+++ b/Userland/Services/WindowServer/WindowManager.cpp
@@ -1138,6 +1138,10 @@ void WindowManager::event(Core::Event& event)
             }
         }
 
+        if (key_event.type() == Event::KeyDown && key_event.key() == Key_Q) {
+            m_active_window->request_close();
+        }
+
         if (MenuManager::the().current_menu()) {
             MenuManager::the().dispatch_event(event);
             return;


tomuta:
This seems like a rather odd short cut, which is likely used by some ports. Why not just the universal Alt+F4?

daubaris:
@tomuta Well, I guess there were different suggestions, but the present did not seem to be an unreasonable choice. Relevant discussion can be seen [here](https://freenode.logbot.info/serenityos/20210220#c6983191).

awesomekling:
Either way, this is much too simplistic.

First off, WindowServer uses Logo+ for all of its shortcuts, so if this is a WindowServer thing, Logo+ would be a better prefix.

There are two logical functions here, and I think they might be best implemented on the client side. There's "close window" (Ctrl+W) and "quit application" (Ctrl+Q)

daubaris:
Alright, I guess then I will refactor it to be on the client side.

alimpfard:
@awesomekling while I do agree that this is on the simplistic side, handling this on windowserver isn't that bad of an idea (if it was configurable). it's much better imo to have a global known binding that always does what you expect, if you allow the clients to have their own bindings, they will either have nothing, or have the wrong defaults (compared to the system, and each other).

e.g. we've internalised this, but C-w is "kill word" in a terminal more often than not, but it's "close window" (whatever the meaning of "window" is to that specific application) on many other apps, and it simply does nothing on many others.

daubaris:
I just realized that Alt+F4 already existed. As a result, I am closing the PR as it is completely irrelevant.

PR #5431, opened 2021-02-20
LibLine: Avoid refreshing the entire line when inserting at the end

### LINKED ISSUE #5296: LibLine: Could do with much less repainting
Currently, if the lines or the styles change, and the user has actually inserted a character somewhere, the entirety of the line is repainted.

The most simple change could be to note when the insertions are only at the end of the buffer (and cache the ending styles), to only append to the previous line when typing consecutive characters.

Going further, we could compute deltas between the previous and the current line, and only modify what needs to be modified.

alimpfard:
This patchset allows the editor to avoid redrawing the entire line when
the changes cause no unrecoverable style updates, and are at the end of
the line (this applies to most normal typing situations).
Cases that this does not resolve:
- When the cursor is not at the end of the buffer
- When a display refresh changes the styles on the already-drawn parts
  of the line
- When the prompt has not yet been drawn, or has somehow changed

Fixes #5296.

cc @awesomekling, you ask for faster libline redraws, you get more convoluted redraw logic

### DIFF
--- a/Userland/Libraries/LibLine/Editor.cpp
+++ b/Userland/Libraries/LibLine/Editor.cpp
@@ -26,6 +26,7 @@
  */
 
 #include "Editor.h"
+#include <AK/Debug.h>
 #include <AK/GenericLexer.h>
 #include <AK/JsonObject.h>
 #include <AK/ScopeGuard.h>
@@ -355,7 +356,7 @@ void Editor::insert(const u32 cp)
     }
 
     m_buffer.insert(m_cursor, cp);
-    ++m_chars_inserted_in_the_middle;
+    ++m_chars_touched_in_the_middle;
     ++m_cursor;
     m_inline_search_cursor = m_cursor;
 }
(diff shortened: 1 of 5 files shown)


Userland/Libraries/LibLine/Editor.cpp
```diff
@@ -1252,30 +1246,87 @@ void Editor::refresh_display()
...
+        VT::apply_style(Style::reset_style());
+        m_pending_chars.clear();
+        m_refresh_needed = false;
+        m_cached_buffer_metrics = actual_rendered_string_metrics(buffer_view());
+        m_chars_touched_in_the_middle = 0;
+        m_drawn_end_of_line_offset = m_buffer.size();
+
+        // No need to reposition the cursor, the cursor is already where it needs to be.
+        return;
+    } else {
```
awesomekling, line 1271: No need for `else` after `return`.

alimpfard: Ehhh, fair enough :)

Issue #5299, opened 2021-02-12
FileManager: crashes opening bmpsuite_files


diversys:
To repro:

`fm /res/html/misc/bmpsuite_files`

bcoles:
```
USERSPACE(27) ASSERTION FAILED: bitmap.format() == Gfx::BitmapFormat::RGB32 || bitmap.format() == Gfx::BitmapFormat::RGBA32
../Userland/Libraries/LibGfx/Painter.cpp:76
```

<details><p>

```
USERSPACE(27) ASSERTION FAILED: bitmap.format() == Gfx::BitmapFormat::RGB32 || bitmap.format() == Gfx::BitmapFormat::RGBA32
../Userland/Libraries/LibGfx/Painter.cpp:76
[#0 FileManager(27:28)]: 0x00000000  (?)

[#0 FileManager(27:28)]: 0xc01c25b6  Kernel::Process::crash(int, unsigned int, bool) +0x2e6
[#0 FileManager(27:28)]: 0xc01ce5e7  Kernel::Process::sys$abort() +0x1f
[#0 FileManager(27:28)]: 0xc01cc8e0  syscall_handler +0xa87
[#0 FileManager(27:28)]: 0xc01cbe21  syscall_asm_entry +0x31
[#0 FileManager(27:28)]: Process regions:
[#0 FileManager(27:28)]: BEGIN       END         SIZE        ACCESS  NAME
[FileManager(27:28)]: 01c01000 -- 02000fff 00400000 RW  T  
[FileManager(27:28)]: 02002000 -- 02002fff 00001000 R      
[FileManager(27:28)]: 02004000 -- 02004fff 00001000 RW S   
[FileManager(27:28)]: 02006000 -- 02009fff 00004000 R  S   
[FileManager(27:28)]: 0200d000 -- 0200dfff 00001000 RW     
[FileManager(27:28)]: 02010000 -- 0201ffff 00010000        
[FileManager(27:28)]: 02021000 -- 02025fff 00005000 R  S   
[FileManager(27:28)]: 02027000 -- 02027fff 00001000 RW     
[FileManager(27:28)]: 0202b000 -- 0202bfff 00001000 RW     
[FileManager(27:28)]: 0202d000 -- 0202dfff 00001000 RW     
[FileManager(27:28)]: 02030000 -- 0203ffff 00010000 RW     
[FileManager(27:28)]: 02041000 -- 02113fff 000d3000 R      
[...]

PR #5438, opened 2021-02-20
Documentation, access to the fuzz targets inside serenity, crash in file manager

BenWiederhake:
- I had to set up serenity from scratch and ironed out a few shortcomings in the QtCreator setup documentation
- I *again* ran into the issue of wanting to "just" run the FuzzBMPLoader code inside Serenity. There used to be no way to do this, now there is.
- The bug wasn't in the BMP loader at all, instead it was in LibGUI: When making a thumbnail, don't use the exact same format unconditionally, especially since the later stages assume (A)RGB32.


Userland/Utilities/test-fuzz.cpp
```diff
@@ -0,0 +1,130 @@
...
+//    T(TTF) \
+//    T(URL)
+*/
+
+#undef __ENUMERATE_TARGET
+#define __ENUMERATE_TARGET(x) extern "C" int Test##x(const uint8_t*, size_t);
+ENUMERATE_TARGETS(__ENUMERATE_TARGET)
+#undef __ENUMERATE_TARGET
+
+// FIXME: How to automate this?
```
alimpfard, line 60: Generate this particular file, maybe?

Userland/Utilities/test-fuzz.cpp
```diff
@@ -0,0 +1,130 @@
...
+//    T(PBMLoader) \
+//    T(PGMLoader) \
+//    T(PNGLoader) \
+//    T(PPMLoader) \
+//    T(RegexECMA262) \
+//    T(RegexPosixExtended) \
+//    T(Shell) \
+//    T(TTF) \
+//    T(URL)
+*/
```
alimpfard, line 53: Any particular reason why these are disabled?

Userland/Utilities/test-fuzz.cpp
```diff
@@ -0,0 +1,130 @@
...
+}
+
+int main(int argc, char** argv)
+{
+    const char* type;
+    const char* filename;
+
+    Core::ArgsParser args_parser;
+    args_parser.add_positional_argument(type, "Which type of fuzzer you'd like to try (use \"list\" to list all existing)", "fuzz-kind");
+    args_parser.add_positional_argument(filename, "Which file to use as input", "filename");
```
alimpfard, line 116: I think it's more common to write something like "Type of fuzzer to run" and "Fuzzer input file" instead?
Just a nit

Userland/Utilities/test-fuzz.cpp
```diff
@@ -0,0 +1,130 @@
...
+#include <Meta/Lagom/Fuzzers/FuzzHttpRequest.cpp>
+#undef LLVMFuzzerTestOneInput
+
+#define LLVMFuzzerTestOneInput TestICOLoader
+#include <Meta/Lagom/Fuzzers/FuzzICOLoader.cpp>
+#undef LLVMFuzzerTestOneInput
+
+typedef int (*TestOneInputFn)(const uint8_t*, size_t);
+
+static TestOneInputFn parse_target_name(const String& name)
```
alimpfard, line 87: `auto` should be able to infer this, I think.
Below too, so you can very likely omit the typedef above

Userland/Utilities/test-fuzz.cpp
```diff
@@ -0,0 +1,130 @@
...
+{
+    const char* type;
+    const char* filename;
+
+    Core::ArgsParser args_parser;
+    args_parser.add_positional_argument(type, "Which type of fuzzer you'd like to try (use \"list\" to list all existing)", "fuzz-kind");
+    args_parser.add_positional_argument(filename, "Which file to use as input", "filename");
+    args_parser.parse(argc, argv);
+
+    TestOneInputFn fn = parse_target_name(type);
```
alimpfard, line 119: (`auto` here)

Userland/Utilities/test-fuzz.cpp
BenWiederhake: I didn't know that `auto` works for return types!

Userland/Utilities/test-fuzz.cpp
BenWiederhake: Hmm, that would involve non-trivial amounts of cmake magic. Also, we might exclude some fuzzers, for example FuzziliJS, so it shouldn't be anything that tries to take *all* fuzzers.

Ironically, `#include` *does* work with macro-built filenames, but the `#define` doesn't. Just in case that sparks an idea.

Userland/Utilities/test-fuzz.cpp
BenWiederhake: Nope, just laziness.

BenWiederhake: Converted to draft until I finish that part.

Userland/Utilities/test-fuzz.cpp
alimpfard: Why build it with cmake? we have a bunch of generators written in C++ already.

Hmm...sure, switch to this (not sure if `#include` works like this, but worth a try)
```
#define INCLUDE_FUZZER(name) \
    #define LLVMTestOneInput run
    namespace name { \
    #include wherever/name/is \
    }
```

Userland/Utilities/test-fuzz.cpp
alimpfard: No real reason to block on that, just wondering if there was some issue with those specific fuzzers.

BenWiederhake:
Changes since first push:
- Many nits from @alimpfard – thanks!
- Extended it to *all* existing fuzzing targets (except FuzziliJS which works very differently)

I'm still uncertain about the many `#include`s. Compiling them as separate unit asks for trouble (how to resolve the obvious naming conflict?), and handling the looping by cmake code generation spreads out the definition even further – after all, the enumeration and the actually included fileset needs to be kept in sync.I can't think of anything better than manually listing all the included files, so that's what I'll go with. Suggestions welcome.

Userland/Utilities/test-fuzz.cpp
BenWiederhake: This will fail for three reasons:

1. A macro can't include a `#`, because it will be interpreted as "stringify the macro argument with the following name", and then complain about `include` not being a macro argument. That's why I keep saying that a macro can't `#include`. Do note that a `#include` can make use of macros.
2. `namespace` will cause lots of trouble, because the fuzz targets in turn `include` a bunch of headers, which cause the fuzz target try to link against a symbol `name::Gfx::BMPLoader::…`, which doesn't exist.
3. `namespace` doesn't even solve the problem, because the fuzzers are (and have to be) `extern "C"`, so the namespacing is effectively ignored.

alimpfard: oh yikes. 
It seems I misunderstood your "#include does work with macro-built filenames", my bad.

Then I think it's either listing them all, or accepting the downsides and generating them (at least partially, maybe just `#define INCLUDED_AS_UTILITY` and just put all of FuzziliJs behind an `#ifndef INCLUDED_AS_UTILITY`)

BenWiederhake: That doesn't work as easily either, because the decision made in FuzziliJS would somehow need to travel to the output of `test-fuzz list all`, or else that command would list FuzziliJS as a possible target.

alimpfard: Fair enough. so it's either a generator that can make the decisions or just listing them all.
these aren't that frequently updated, so maybe it being manual isn't _that_ bad...

awesomekling:
Is that file manager bug the same as #5299 perhaps?

BenWiederhake:
Technically yes, because now it crashes somewhere else.

BenWiederhake:
Changes since the last push:
- Sneakily remove a "FIXME" in `Userland/Utilities/test-fuzz.cpp`, because apparently there's no way to do this any better.
- Rebase for convenience.

awesomekling:
> Technically yes, because now it crashes somewhere else.

Does that make "Undocumented bug" true or false?

BenWiederhake:
Right, I should change that, too ^^'

Changes since last push:
- Reword that commit message
- Rebase

PR #5439, opened 2021-02-20
LibGUI: Let ScrollableWidget handle the wheel events of its ScrollBars

### LINKED ISSUE #5419: LibGUI: Different scroll speed when mouse is hovering the scrollbar
Steps to reproduce:
1. Open File Manager
2. Navigate to a folder with many files, e.g. /bin
3. Move the mouse cursor inside the IconView widget, but not over the scrollbar
4. Scroll up and down to get a feel for how much it should scroll with each wheel turn
5. While scrolling up and down, move the mouse cursor over the scroll bar and continue moving the mouse wheel at the same speed
6. Observe that the widget's content moves a lot slower while the mouse is hovering the scroll bar

tomuta:
Route the ScrollBar's wheel event to the ScrollableWidget so it can
handle it itself. This allows it to handle it consistently (e.g.
speed) when the cursor is hovering the scroll bars rather than the
widget's contents.

Fixes #5419

### DIFF
--- a/Userland/Libraries/LibGUI/ScrollBar.h
+++ b/Userland/Libraries/LibGUI/ScrollBar.h
@@ -31,7 +31,7 @@
 
 namespace GUI {
 
-class ScrollBar final : public AbstractSlider {
+class ScrollBar : public AbstractSlider {
     C_OBJECT(ScrollBar);
 
 public:
@@ -49,7 +49,7 @@ public:
         Scrubber,
     };
 
-private:
+protected:
     explicit ScrollBar(Gfx::Orientation = Gfx::Orientation::Vertical);
 
     virtual void paint_event(PaintEvent&) override;
@@ -60,6 +60,7 @@ private:
     virtual void leave_event(Core::Event&) override;
     virtual void change_event(Event&) override;
 
+private:
     int default_button_size() const { return 16; }
     int button_size() const { return length(orientation()) <= (default_button_size() * 2) ? length(orientation()) / 2 : default_button_size(); }
     int button_width() const { return orientation() == Orientation::Vertical ? width() : button_size(); }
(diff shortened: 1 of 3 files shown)


awesomekling:
If you call `ignore()` on the wheel event, it will bubble up the widget parent chain. So a more idiomatic approach would be:

* Make a subclass of `ScrollBar` in ScrollableWidget.cpp, e.g `ScrollableWidgetScrollBar`.
* Override its `wheel_event() to simply ignore the events.
* Then install those kinds of scroll bars in the ScrollableWidget :)

tomuta:
That sounds like a better solution.

tomuta:
Hm, `ScrollBar` is final, though...

awesomekling:
> Hm, `ScrollBar` is final, though...

Just remove the `final`, there's no reason for it other than "hey nobody was inheriting this" :)

tomuta:
Sounds good.

tomuta:
However, how would I handle the special horizontal scrollbar case? How do I know that the mouse was over it? Just check if the event position is in the child widget's area?

awesomekling:
Actually maybe it's easier to still subclass, but instead of bubbling the event, just have `ScrollableWidgetScrollBar::wheel_event()` call `ScrollableWidget::handle_wheel_event()` as you did in the original patch.

awesomekling:
Bubbling the event indeed has the problem of the coordinates still being in child widget space.

tomuta:
Well that problem was the same with my initial approach, but I ignored it as the event handler didn't care about the coordinates. Seems like if events are bubbled up we should translate coordinates accordingly, but that'd be a separate issue.

tomuta:
Ok, how's this looking?

awesomekling:
Gorgeous

PR #5441, opened 2021-02-21
LibLine: Update the lazy refresh data and flags in some more places

### LINKED ISSUE #5440: LibLine: Goofy behavior with up/down
Repro:

```
foo<ENTER>
<UP>
<DOWN>
<UP>
<DOWN>
<UP>
<DOWN>
```

It now says "foofoofoo". :(

cc @alimpfard

alimpfard:
These were forgotten in the last LibLine commit, any changes to m_buffer
not going through insert() and remove_at_index() should also be updating
these.

Fixes #5440.

LibLine: That one library in dire need of tests...a VT test thing would be nice...

### DIFF
--- a/Userland/Libraries/LibLine/Editor.cpp
+++ b/Userland/Libraries/LibLine/Editor.cpp
@@ -1531,12 +1531,16 @@ void VT::apply_style(const Style& style, bool is_starting)
 
 void VT::clear_lines(size_t count_above, size_t count_below)
 {
-    // Go down count_below lines.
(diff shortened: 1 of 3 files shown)


alimpfard:
ffs ^L is broken too

alimpfard:
Things now tested (and fixed if broken):
- up/down
- search
- ^L, ^W, ^T, Alt+T, and a bunch more combinations
- suggestions

I hope I'm not missing anything else

Issue #5209, opened 2021-02-01
LibCore: general purpose file management helpers


nvella:
It seems we are often duplicating code across the system for copying, moving and deleting file hierarchies - eg. code duplication between FileManager and `cp`, `rm`, and the likes. It'd be a nice if we had some general-purpose file management helper functions in LibCore which could fulfill all these use-cases, so we don't repeat ourselves.

linusg:
Implemented in #5444.

PR #5444, opened 2021-02-21
LibCore+FileManager+Utilities: Add file management helpers to reduce code duplication

RealKC:
I'd say this addresses #5209

Perhaps of note is the fact that the functions that existed previously were not consistent wrt the order of the destination and source paths parameters, the newly added functions are, (and I think there should be no bugs related to that either) they take them in `(dst, src)` order, tho I have no issues with switching to `(src, dst)` if that is desired


Userland/Applications/FileManager/FileUtils.h
```diff
@@ -41,10 +41,5 @@ enum class FileOperation {
 void delete_path(const String&, GUI::Window*);
 void delete_paths(const Vector<String>&, bool should_confirm, GUI::Window*);
 int delete_directory(String directory, String& file_that_caused_error);
```
IdanHo, line 43: You might want to replace these with LibCore utils as well (similarly to your rm commit)

RealKC: Yeah, initially I wanted to, then I thought it to be a bit too complex, but looking back to it, I think it shouldn't be that bad.

I'd need to make `remove` return the file path that caused the error alongside the OSError I think. Will look into it in an hour or so

IdanHo: just replacing delete_path with the LibCore equivalent might be enough i think, the whole delete_path and delete_directory stuff might be too customized for it

Userland/Libraries/LibCore/File.cpp
```diff
@@ -259,4 +266,213 @@ NonnullRefPtr<File> File::standard_error()
...
+}
+
+Result<void, String> File::copy_directory(const String& dst_path, const String& src_path, const struct stat& src_stat, LinkMode link)
+{
+    int rc = mkdir(dst_path.characters(), 0755);
+    if (rc < 0)
+        return String(strerror(errno));
+
+    String src_rp = Core::File::real_path_for(src_path);
+    src_rp = String::format("%s/", src_rp.characters());
```
awesomekling, line 394: Please don't use `String::format()` in new code as it is being removed. Use `String::formatted()` instead.

Userland/Libraries/LibCore/File.cpp
```diff
@@ -259,4 +266,213 @@ NonnullRefPtr<File> File::standard_error()
...
+    String dst_rp = Core::File::real_path_for(dst_path);
+    dst_rp = String::format("%s/", dst_rp.characters());
+
+    if (!dst_rp.is_empty() && dst_rp.starts_with(src_rp)) {
+        fprintf(stderr, "cp: Cannot copy %s into itself (%s)\n",
+            src_path.characters(), dst_path.characters());
```
awesomekling, line 400: Please use `warnln()` instead of `fprintf(stderr, ...)`

Userland/Utilities/cp.cpp
```diff
@@ -51,13 +51,22 @@ int main(int argc, char** argv)
...
-        if (!ok)
+        auto rc = Core::File::copy_file_or_directory(
+            destination, source,
+            recursion_allowed ? Core::File::RecursionMode::Allowed : Core::File::RecursionMode::Disallowed,
+            link ? Core::File::LinkMode::Allowed : Core::File::LinkMode::Disallowed,
+            Core::File::AddDuplicateFileMarker::No);
+
+        if (rc.is_error()) {
+            // FIXME: This feels a bit hacky, but I am not sure what a better way would be
+            if (rc.error() == "Tried to copy directory with RecursionMode::Disallowed")
```
awesomekling, line 63: Way too hacky indeed; either use `OSError` or if we want to provide even more info then let's define a richer error struct type and return that from `copy_file_or_directory`.

Userland/Applications/FileManager/DirectoryView.cpp
```diff
@@ -580,8 +580,8 @@ void DirectoryView::handle_drop(const GUI::ModelIndex& index, const GUI::DropEve
         if (url_to_copy.path() == new_path)
             continue;
 
-        if (!FileUtils::copy_file_or_directory(url_to_copy.path(), new_path)) {
-            auto error_message = String::formatted("Could not copy {} into {}.", url_to_copy.to_string(), new_path);
+        if (auto rc = Core::File::copy_file_or_directory(new_path, url_to_copy.path()); rc.is_error()) {
+            auto error_message = String::formatted("Could not copy {} into {}: \n", url_to_copy.to_string(), new_path, rc.error());
```
awesomekling, line 584: `rc` is conventionally used to store syscall return codes.
Since we're not calling a syscall here, I'd call it `result` instead.

Also the format string looks wrong, missing a "{}"
And we don't need to end it with `\n` if we're showing it in a **GUI::MessageBox** :)

awesomekling (changes requested):

Userland/Applications/FileManager/FileUtils.h
RealKC: I've rewritten delete_path and delete_paths in terms of Core::File::remove

Userland/Libraries/LibCore/File.cpp
RealKC: These were actually the result of me forgetting to remove them when copying the code from the existing implementation, so I removed them

Userland/Utilities/cp.cpp
RealKC: Is the richer error type I've written a good fit in your opinion?

PR #5448, opened 2021-02-21
AK + Kernel: Remove unneeded Thread::set_default_signal_dispositions

bgianfo:
* Add `Span<T>` constructor for arrays. 

* Add `Array<T, Size>::fill(...)` 

* The `default_signal_action(u8 signal)` function already has the
  full mapping. The only caveat being that now we need to make
  sure the thread constructor and `clear_signals()` method do the work
  of resetting the `m_signal_action_data` array, instead or relying on
  the previous logic in `Thread::set_default_signal_dispositions`.

* Switch `Thread::m_signal_action_data` to use `Array<T, Size>`

### DIFF
--- a/Kernel/Thread.h
+++ b/Kernel/Thread.h
@@ -47,6 +47,7 @@
 #include <Kernel/TimerQueue.h>
 #include <Kernel/UnixTypes.h>
 #include <LibC/fd_set.h>
+#include <LibC/signal_numbers.h>
 
 namespace Kernel {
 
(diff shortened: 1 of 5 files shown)


Kernel/Thread.h
```diff
@@ -1241,7 +1239,7 @@ class Thread
     u32 m_kernel_stack_top { 0 };
     OwnPtr<Region> m_kernel_stack_region;
     VirtualAddress m_thread_specific_data;
-    SignalActionData m_signal_action_data[32];
+    SignalActionData m_signal_action_data[32] { };
```
awesomekling, line 1242: Suggest making this an `Array<SignalActionData, 32>` instead and adding `Array::fill()`

awesomekling: `Array<SignalActionData, NSIG>` even :)

PR #5449, opened 2021-02-21
AK + Kernel + Userland: Avoid raw memset where possible, as it's easy to introduce bugs. 

bgianfo:
Raw `memset` is relatively easy to mess up, avoid it when there are
better alternatives provided by the compiler in modern C++.

This change addresses this by using uniform initialization instead of `memset` for stack variables

The other change is to enlighten `ByteBuffer` with a new `zero_buffer` function which performs the
`memset` internally with the known pointer / data size.

### DIFF
--- a/AK/ByteBuffer.h
+++ b/AK/ByteBuffer.h
@@ -89,6 +89,8 @@ public:
 
     void grow(size_t size);
 
+    void zero_fill();
+
 private:
     explicit ByteBufferImpl(size_t);
     ByteBufferImpl(const void*, size_t);
@@ -248,6 +250,11 @@ public:
         __builtin_memcpy(this->data() + offset, data, data_size);
     }
 
+    void zero_fill()
+    {
+        m_impl->zero_fill();
+    }
+
     operator Bytes() { return bytes(); }
     operator ReadonlyBytes() const { return bytes(); }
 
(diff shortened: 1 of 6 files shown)


Kernel/FileSystem/Ext2FileSystem.cpp
```diff
@@ -370,7 +370,7 @@ KResult Ext2FS::write_block_list_for_inode(InodeIndex inode_index, ext2_inode& e
 
             auto ind_block_contents = ByteBuffer::create_uninitialized(block_size());
             if (ind_block_new) {
-                memset(ind_block_contents.data(), 0, dind_block_contents.size());
```
bgianfo, line 373: Note: Here is one such bug this addresses, we are passing the wrong data size.

awesomekling: Great catch!

AK/ByteBuffer.h
```diff
@@ -89,6 +89,8 @@ class ByteBufferImpl : public RefCounted<ByteBufferImpl> {
 
     void grow(size_t size);
 
+    void zero_buffer();
```
awesomekling, line 92: Let's call it `zero_fill()`

Kernel/FileSystem/Ext2FileSystem.cpp
awesomekling: Typo in commit message here btw, s/BlockBuffer/ByteBuffer/

AK/ByteBuffer.h
bgianfo: Sure, done.

PR #5456, opened 2021-02-21
LibWeb: Add simple css (at)import handling

speles:
This PR adds basic (at)import rule handling.
Only the loading of the CSS is supported, and the meta queries are just ignored.
The reason for this PR is w3.org CSS1 tests love the import rules, and having them implemented will ease the future testing and development.

PS. All these commits leave LibWeb in a buildable and working state, but they don't have much meaning separately. I'm not sure if I had to squash them per contribution guidelines.
But it would've been an absolutely huge commit, so I decided to separate them for reviewing convenience.


Userland/Libraries/LibWeb/CSS/ImportRule.h
```diff
@@ -0,0 +1,57 @@
...
+ */
+
+#pragma once
+
+#include <AK/URL.h>
+#include <LibWeb/CSS/CSSRule.h>
+
+namespace Web::CSS {
+
+class ImportRule : public CSSRule {
```
awesomekling, line 34: Let's call it `CSSImportRule` to match the CSSOM.

Userland/Libraries/LibWeb/CSS/CSSRule.h
```diff
@@ -0,0 +1,53 @@
...
+#include <LibWeb/CSS/Selector.h>
+#include <LibWeb/CSS/StyleDeclaration.h>
+
+namespace Web::CSS {
+
+class CSSRule : public RefCounted<CSSRule> {
+public:
+    virtual ~CSSRule();
+
+    enum class Kind : u32 {
```
awesomekling, line 40: Let's call this `Type`

Userland/Libraries/LibWeb/CSS/CSSRule.h
```diff
@@ -0,0 +1,53 @@
...
+        Style,
+        Import,
+        Media,
+        __Count,
+    };
+
+    virtual String class_name() const = 0;
```
awesomekling, line 47: Let's return `StringView` instead of `String` here so we don't have to pay for a heap allocation when calling this.

Userland/Libraries/LibWeb/CSS/ImportRule.h
```diff
@@ -0,0 +1,57 @@
...
+
+    ~ImportRule();
+
+    const AK::URL& url() const { return m_url; }
+
+    virtual String class_name() const { return "Import Rule"; };
+    virtual Kind kind() const { return Kind::Import; };
+
+private:
+    ImportRule(AK::URL&& url);
```
awesomekling, line 57: No need for the `&&` (nor the parameter name)

Userland/Libraries/LibWeb/CSS/ImportRule.h
```diff
@@ -0,0 +1,57 @@
...
+    static NonnullRefPtr<ImportRule> create(AK::URL&& url)
+    {
+        return adopt(*new ImportRule(move(url)));
+    }
+
+    ~ImportRule();
+
+    const AK::URL& url() const { return m_url; }
+
+    virtual String class_name() const { return "Import Rule"; };
```
awesomekling, line 53: Please return the actual class name, i.e `"CSSImportRule"` rather than `"Import Rule"`.

Userland/Libraries/LibWeb/CSS/ImportRule.h
```diff
@@ -0,0 +1,57 @@
...
+#include <LibWeb/CSS/CSSRule.h>
+
+namespace Web::CSS {
+
+class ImportRule : public CSSRule {
+    AK_MAKE_NONCOPYABLE(ImportRule);
+    AK_MAKE_NONMOVABLE(ImportRule);
+
+public:
+    static NonnullRefPtr<ImportRule> create(AK::URL&& url)
```
awesomekling, line 39: No need for the `&&` here.

Userland/Libraries/LibWeb/CSS/ImportRule.h
```diff
@@ -0,0 +1,57 @@
...
+
+    const AK::URL& url() const { return m_url; }
+
+    virtual String class_name() const { return "Import Rule"; };
+    virtual Kind kind() const { return Kind::Import; };
+
+private:
+    ImportRule(AK::URL&& url);
+
+    AK::URL m_url;
```
awesomekling, line 59: Please omit the `AK::` prefix everywhere, these are `using`'ed automatically so no qualification is required.

Userland/Libraries/LibWeb/CSS/Parser/CSSParser.cpp
```diff
@@ -820,7 +808,129 @@ class CSSParser {
...
+        buffer.clear();
+        while (peek() && peek() != ')')
+            buffer.append(consume_one());
+
+        String url_value(String::copy(buffer));
+        buffer.clear();
+
+        if (peek() == ')')
+            return Optional(url_value);
+        return Optional<String>();
```
awesomekling, line 856: `return {};` would be more idiomatic here.

Userland/Libraries/LibWeb/CSS/Parser/CSSParser.h
```diff
@@ -38,6 +39,8 @@ class ParsingContext {
 
     bool in_quirks_mode() const;
 
+    AK::URL complete_url(const String& addr) const;
```
awesomekling, line 42: Please omit the `addr` parameter name as it's implied from context anyway.

Userland/Libraries/LibWeb/CSS/ImportRule.h
```diff
@@ -45,13 +45,19 @@ class ImportRule : public CSSRule {
 
     const AK::URL& url() const { return m_url; }
 
+    bool has_import_result() const { return !m_style_sheet.is_null(); }
+    RefPtr<StyleSheet>& loaded_style_sheet() { return m_style_sheet; }
+    const RefPtr<StyleSheet>& loaded_style_sheet() const { return m_style_sheet; }
```
awesomekling, line 50: These are a bit awkward, I don't think we want to vend a reference to the internal `RefPtr` member. Instead, either return `RefPtr<StyleSheet>` or simply `StyleSheet*`.

awesomekling:
Cool stuff, I like the consolidation of sheet loading into a CSSLoader thing :)

speles:
Fixed all review comments (hopefully haven't missed any), and rebased on the latest master.

awesomekling:
:star:

PR #5460, opened 2021-02-21
Toolchain: Ensure that the correct patches are applied when building

alimpfard:
No one seemed to be addressing this, so I decided to give it a shot.
Things it does:
- nuke Toolchain/Tarballs/\<relevant directory>
- nuke Toolchain/Build/\<relevant directory>
- clean the previously built artifacts in Build

Things it *doesn't* do:
- Remove stuff from Toolchain/Local: Not entirely convinced if this is necessary
- Clean ccache cache: While this did fix my problem once, I think it should be dealt with on a case-by-case basis, it's too weird to have as part of a build script.
- Mess with the repo checkout: Let's leave that as a last resort.

cc @tomuta (your idea, mostly)

### DIFF
--- a/Toolchain/BuildIt.sh
+++ b/Toolchain/BuildIt.sh
@@ -140,47 +140,50 @@ pushd "$DIR/Tarballs"
         echo "Skipped downloading gcc"
     fi
 
-    if [ ! -d ${BINUTILS_NAME} ]; then
-        echo "Extracting binutils..."
-        tar -xzf ${BINUTILS_PKG}
-
-        pushd ${BINUTILS_NAME}
-            if [ "$git_patch" = "1" ]; then
-                git init > /dev/null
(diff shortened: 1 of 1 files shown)


awesomekling:
How about we don't try to get fancy, and simply make BuildIt.sh nuke the Build, Local and Tarballs/{gcc,binutils}-VERSION directories?

alimpfard:
Every time? I guess we could, but is there any real added complexity here? we already check hashes for a whole bunch of things there

awesomekling:
If someone is electing to rebuild the toolchain, I would think a "clean build" is preferred over a "possibly correct but totally incremental" build

alimpfard:
Alright. so long as this nukes the right dirs, I have nothing against it :shrug:

PR #5461, opened 2021-02-21
WindowServer: Re-evaluate the mouse cursor when alpha hit-testing

tomuta:
A window repaint may change the alpha value, resulting in a different
hit test outcome. In those cases, re-evaluate the cursor hit testing
after a window was painted, and update the cursor if needed.

_Notice how this properly updates the cursor even though the mouse is not being moved:_
hittest_cursor

### DIFF
--- a/Userland/Services/WindowServer/WindowManager.cpp
+++ b/Userland/Services/WindowServer/WindowManager.cpp
@@ -1043,6 +1043,52 @@ void WindowManager::process_mouse_event(MouseEvent& event, Window*& hovered_wind
         clear_resize_candidate();
 }
 
+void WindowManager::reevaluate_hovered_window(Window* updated_window)
+{
+    if (m_dnd_client || m_resize_window || m_move_window || m_cursor_tracking_button || MenuManager::the().has_open_menu())
+        return;
+
+    auto cursor_location = Screen::the().cursor_location();
+    auto* currently_hovered = hovered_window();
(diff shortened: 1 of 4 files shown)


tomuta:
Hm just noticed it still needs some work. Doesn't properly hit-test against window frames...

tomuta:
> Hm just noticed it still needs some work. Doesn't properly hit-test against window frames...

Should be fixed now.

Userland/Services/WindowServer/WindowManager.cpp
```diff
@@ -1043,6 +1043,52 @@ void WindowManager::process_mouse_event(MouseEvent& event, Window*& hovered_wind
         clear_resize_candidate();
 }
 
+void WindowManager::reevaluate_hovered_window(Window* updated_window)
```
awesomekling, line 1046: Can we come up with a better name for this function?
Re-evaluate what?

tomuta: This function re-evaluate the hovered window, and changes it to another one if needed. I'm open to suggestions ;-) Can't think of a good name.

awesomekling: Ohh I get it. I was reading it in a weird way. Yes, we are re-evaluating which window is hovered. Makes sense :)

PR #5464, opened 2021-02-21
Ports: Update Python to 3.9.2

linusg:
Released on 2021-02-19, no patch changes required. https://www.python.org/downloads/release/python-392/


awesomekling:
\>"no patch changes required"
\>still updates every patch with new timestamps
🤔

linusg:
> still updates every patch with new timestamps

heh, I just re-applied all patches to make sure the markers are still up to date :smile:

PR #5468, opened 2021-02-21
AK+Everywhere: EVERYBODY GETS A COMPILETIME FORMAT STRING CHECK

alimpfard:
Adds compiletime checks for `outln()`, `warnln()` and `dmesgln()`.

Unlike `dbgln()`, this one is quite a bit more involved, and also shows how this macro-based approach isn't exactly the best approach. but I still have nothing better that doesn't involve changing how format strings should be handled (though one of those options might be the right one, if we sacrifice call-site simplicity)

Draft for now, since I think it's quite a mess as it is.

### DIFF
--- a/AK/Assertions.h
+++ b/AK/Assertions.h
@@ -26,18 +26,14 @@
 
 #pragma once
 
-#ifndef AK_TEST_SUITE
-
-#    if defined(KERNEL)
-#        include <Kernel/Assertions.h>
-#    else
-#        include <assert.h>
-#        ifndef __serenity__
-#            define ASSERT assert
-#            define ASSERT_NOT_REACHED() assert(false)
-#            define RELEASE_ASSERT assert
-#            define TODO ASSERT_NOT_REACHED
-#        endif
+#if defined(KERNEL)
+#    include <Kernel/Assertions.h>
+#else
+#    include <assert.h>
+#    ifndef __serenity__
+#        define ASSERT assert
+#        define ASSERT_NOT_REACHED() assert(false)
+#        define RELEASE_ASSERT assert
+#        define TODO ASSERT_NOT_REACHED
 #    endif
-
 #endif
--- a/AK/CheckedFormatString.h
+++ b/AK/CheckedFormatString.h
@@ -1,0 +1,242 @@
+/*
+ * Copyright (c) 2021, the SerenityOS developers.
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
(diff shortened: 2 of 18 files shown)


asynts:
Hmm, that macro stuff starts to get a bit nasty.

I think I have a solution though: (Just got that idea while looking though the code.)

https://godbolt.org/z/Kb7TMM

If I understand things correctly, passing a `const char*` as template parameter does not require compiler support for NTTP.

This means that we can do things the C++ way: `dbgln<"Magic number: {}">(42)` and then just change the function signature to disable checks entirely.

alimpfard:
Yeah, the macros were bound to get really ugly from the start...

I really don't want to pass the format string through the template parameters, for two main reasons:
- it's ugly.
- clang does not support it

But (and this is a big one) if we are okay with changing how these are called, I think UDLs would be much better (little demo [here](https://gcc.godbolt.org/z/3KPY9q), I have a little test branch I'll push in a bit), we can also keep the `StringView` overloads for when the compiletime checks are disabled (or not supported), and also keep the more pleasant `dbgln("My little {} string"_f, "format")`.

alimpfard:
Here's the branch I mentioned: https://github.com/alimpfard/serenity/tree/format-udl (In particular, this commit: https://github.com/alimpfard/serenity/commit/a389af54526b8aa9e69834f174c52f86cefa2e61)

Seems like the effort needed to change (either to `dbgln<fmt>` or `dbgln(fmt_f)`) is quite high, whichever feels better is fine by me.

trflynn89:
Hey! I've been following this project for a bit and I've seen this discussion about compile-time checking of format strings a couple times. Just wanted to chime in with a potential solution that doesn't change call sites or involve macros.

C++20 added the `consteval` keyword. You can use that here to implicitly construct some `FormatString` object (whose constructor is `consteval`) from the raw format string literal when invoking your `format` or `dbgln` methods. That constructor will be evaluated at compile time, during which you can parse the format string.

To raise a compile error, the simplest thing to do is define some non-`constexpr` method like `on_error(const char *)` in the `FormatString` class, and attempt to invoke that method with some error message. The compiler will raise an error when the invocation is reached due to trying to execute a non-compile-time method from a compile-time context.

Probably easiest to show how this works by example: https://godbolt.org/z/h4odEa

That is a formatter that just replaces any `{}` instances with the provided format parameters in order. It will raise a compile error if the number of `{}` does not match the number of format parameters, or if it encountered an opening '{' without an immediate closing '}'.

Hope this helps!

alimpfard:
This is _very_ interesting, I had never tried using an implicit cast to an object with a `consteval` ctor...
Thanks for showing me this - it helps a whole lot!

alimpfard:
_You get a compiletime check, you get a compiletime check; everybody gets a compiletime check_

In this change:

- We get to have our cake and eat it too, thanks to @trflynn89 
- All the format functions get compiletime checks
- A few more formatting fixes
-

alimpfard:
Seems like this has broken _something_ in the kernel - needs a little more investigation still.

alimpfard:
Build time impacts:

| | Cold ccache | Hot ccache with a single comment changed in Format.h |
| :- | :-: | :-: |
| (previous) concepts with a buncha macros | ? | 332761ms (332s) |
| (this PR) `consteval` check on the format string ctor | 452028ms (452s) | 412021ms (412s) |

asynts:
This is amazing!

PR #5470, opened 2021-02-22
TextEditor: Add basic case insensitivity to Find/Replace and reorganize its layout

thankyouverycool:
Makes find a bit more useful until it gets proper code point matching
find-replace


awesomekling:
Damn, that GroupBox really pulls it together 🤩

PR #5477, opened 2021-02-22
Browser: Implement view source for out of process mode.

xeons:
Implements the view source functionality for out of process mode.

### DIFF
--- a/Userland/Services/WebContent/WebContentServer.ipc
+++ b/Userland/Services/WebContent/WebContentServer.ipc
@@ -20,4 +20,5 @@ endpoint WebContentServer = 89
     KeyDown(i32 key, unsigned modifiers, u32 code_point) =|
 
     DebugRequest(String request, String argument) =|
+    GetSource() =|
 }
--- a/Userland/Applications/Browser/Tab.cpp
+++ b/Userland/Applications/Browser/Tab.cpp
@@ -85,6 +85,20 @@ static void start_download(const URL& url)
     [[maybe_unused]] auto& unused = window.leak_ref();
 }
 
+static void view_source(const String& url, const String& source)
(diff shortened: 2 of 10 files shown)


Userland/Services/WebContent/WebContentServer.ipc
```diff
@@ -20,4 +20,6 @@ endpoint WebContentServer = 89
     KeyDown(i32 key, unsigned modifiers, u32 code_point) =|
 
     DebugRequest(String request, String argument) =|
+
+    GetSource() => (String source)
```
awesomekling, line 24: All IPC API's provided by WebContentServer should be async (`=|`) so the Browser never has to block while waiting for a response :)

awesomekling (changes requested):

Userland/Services/WebContent/WebContentServer.ipc
xeons: Is getting the source going to be a blocking operation though? Shouldn't it return null If the document isn't loaded yet?

awesomekling: Since it's done over IPC, we have no guarantees that the request will be serviced immediately. Hence we must be asynchronous or risk blocking the browser GUI

xeons: So would I create method to request the sourcecode, and then have something like DidRetrievedSourcecode which would get fired back from the handle method?

awesomekling: Right :)

xeons: Everything should be async now.

PR #5486, opened 2021-02-23
Spreadsheet: Store the column index in a Position instead of its name / LibGUI: Don't paint all the table when headers are visible

### LINKED ISSUE #5483: Spreadsheet: Scrolling gets more and more sluggish the further down a spreadsheet
Steps to reproduce:
1. Scroll down until you have a few thousand lines
2. Slowly, but steadily, scroll up and down the spreadsheet using the scrollbar thumb
3. Observe that the higher the line numbers, the more sluggish/slower the scrolling

Interesting tidbit: If you horizontally scroll far enough to make the "A" column (mostly) out of view the problem goes away. The cut off doesn't appear to be exactly the entire "A" column being out of view, a few pixels (20 maybe) can remain visible for the problem to disappear.

spreadsheet_slow

alimpfard:
Fixes #5483.

### DIFF
--- a/Userland/Applications/Spreadsheet/JSIntegration.cpp
+++ b/Userland/Applications/Spreadsheet/JSIntegration.cpp
@@ -125,7 +125,7 @@ JS::Value SheetGlobalObject::get(const JS::PropertyName& name, JS::Value receive
 
             return JS::js_undefined();
         }
-        if (auto pos = Sheet::parse_cell_name(name.as_string()); pos.has_value()) {
+        if (auto pos = m_sheet.parse_cell_name(name.as_string()); pos.has_value()) {
             auto& cell = m_sheet.ensure(pos.value());
             cell.reference_from(m_sheet.current_evaluated_cell());
             return cell.typed_js_data();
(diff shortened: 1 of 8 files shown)


Userland/Applications/Spreadsheet/JSIntegration.cpp
```diff
@@ -254,6 +254,17 @@ JS_DEFINE_NATIVE_FUNCTION(SheetGlobalObject::set_real_cell_contents)
 
 JS_DEFINE_NATIVE_FUNCTION(SheetGlobalObject::parse_cell_name)
 {
+    auto* this_object = vm.this_value(global_object).to_object(global_object);
+    if (!this_object)
+        return JS::js_null();
+
+    if (StringView("SheetGlobalObject") != this_object->class_name()) {
```
awesomekling, line 261:
```cpp
if (!is<SheetGlobalObject>(*this_object))
```

Userland/Applications/Spreadsheet/Position.h
```diff
@@ -46,15 +62,13 @@ struct Position {
...
-        url.set_protocol("spreadsheet");
-        url.set_host("cell");
-        url.set_path(String::formatted("/{}", getpid()));
-        url.set_fragment(String::formatted("{}{}", column, row));
-        return url;
-    }
+    URL to_url(const Sheet& sheet) const;
+
+    size_t column;
+    size_t row { 0 };
```
awesomekling, line 68: Let's be good and always initialize member variables.

Userland/Applications/Spreadsheet/JSIntegration.cpp
alimpfard: Oh yeah, this file needs some _modernisation_

PR #5488, opened 2021-02-23
AK: Rename {DBGLN_NO => ENABLE}_COMPILETIME_FORMAT_CHECK

linusg:
This is no longer limited to `dbgln()`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -17,13 +17,14 @@ set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
 set(SERENITY_ARCH "i686" CACHE STRING "Target architecture for SerenityOS.")
 
 # Central location for all custom options used in the Serenity build.
-option(ENABLE_ADDRESS_SANITIZER "Enable address sanitizer testing in gcc/clang" FALSE)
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -164,8 +164,8 @@ add_compile_definitions(DEBUG SANITIZE_PTRS)
 set(CMAKE_CXX_FLAGS_STATIC "${CMAKE_CXX_FLAGS} -static")
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -pie -fpic")
 
-if (DBGLN_NO_COMPILETIME_FORMAT_CHECK)
-    add_compile_definitions(DBGLN_NO_COMPILETIME_FORMAT_CHECK)
+if (DISABLE_COMPILETIME_FORMAT_CHECK)
```
bgianfo, line 167: tangentially related... this should probably surfaced as a documented option(..) in the CMake files (see the top of this file for example).

awesomekling:
Hey since we're here.. how about..

`#ifndef DISABLE_` -> `#ifdef ENABLE`

🤔

linusg:
d o n e

PR #5489, opened 2021-02-24
AK: Make dbgln_if() avoid evaluating the arguments when disabled

alimpfard:


AK/Format.h
```diff
@@ -488,4 +486,4 @@ using AK::CheckedFormatString;
 using AK::FormatIfSupported;
 using AK::FormatString;
 
-#define dbgln_if(flag, fmt, ...) dbgln<flag>(fmt, ##__VA_ARGS__)
+#define dbgln_if(flag, fmt, ...) ((flag) ? dbgln(fmt, ##__VA_ARGS__) : (void)0)
```
bgianfo, line 489: nit pick: It might be arguably more clear if it was just wrapped in constexpr inside the macro? 

```suggestion
#define dbgln_if(flag, fmt, ...) \ 
  if constexpr (flag) { dbgln(fmt, ##__VA_ARGS__) } 
```

alimpfard: This would mess up following `else` clauses, fixing that would turn it to something like `do { if constexpr (...) ... } while(0)`

Either would work, but I like the ternary better :shrug:

alimpfard: I guess it could be more clear, why not.

PR #5502, opened 2021-02-24
LibC: Implement fenv.h

### LINKED ISSUE #853: LibC: fenv.h and fseeko missing (needed for hatari Atari emulator)
hatari emulates the Atari 16/32 bit machines such as the Atari ST and Falcon. Its only real dependencies are SDL2 and zlib so there is a good chance it will run under Serenity with our existing set of ports. I've just spent the best part of a day trying to get it to build only to discover it requires fenv.h (floating point data types) and fseeko. Serenity has fseek but not fseeko. Hopefully those are the only missing headers.

This port doesn't build yet but I thought it was worth sharing here as an example of how to get a cmake app to build using ports and with any luck we will be able to run it soon.

[hatari-serenity.tar.gz](https://github.com/SerenityOS/serenity/files/3929669/hatari-serenity.tar.gz)

RealKC:
This is my first time ever doing inline assembly, so please tell me if it's all correct.

Also the C standard says that the floating point environment must be thread local, but this patch does not implement that. Reading the intel programmer's manual, it is suggested to track FPU ownership in the kernel, I'm not sure if we do that already, if we do, we get the localness of the environment for free, if not, we will have to implement that (could be done at a latter date).

On another note, I know that it is recommended to use the `u<size>` types in serenity code, if I keep the definition of `struct fenv_t` in the header, I have to use the `uint<size>_t` types, should I move it in the source file and side-step that issue entirely, or keep it as is?

Closes #853

Edit: this also does not touch anything related to SSE, though I think it should? not sure

### DIFF
--- a/Userland/Libraries/LibC/CMakeLists.txt
+++ b/Userland/Libraries/LibC/CMakeLists.txt
@@ -6,6 +6,7 @@ set(LIBC_SOURCES
     dirent.cpp
     dlfcn.cpp
     fcntl.cpp
+    fenv.cpp
     getopt.cpp
     grp.cpp
     ioctl.cpp
--- a/Userland/Libraries/LibC/fenv.cpp
+++ b/Userland/Libraries/LibC/fenv.cpp
@@ -1,0 +1,229 @@
+/*
+ * Copyright (c) 2021, Mițca Dumitru <dumitru0mitca@gmail.com>
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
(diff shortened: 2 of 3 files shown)


RealKC:
Updates:

> Also the C standard says that the floating point environment must be thread local, but this patch does not implement that. Reading the intel programmer's manual, it is suggested to track FPU ownership in the kernel, I'm not sure if we do that already, if we do, we get the localness of the environment for free, if not, we will have to implement that (could be done at a latter date).

This is done already by the kernel

> On another note, I know that it is recommended to use the u<size> types in serenity code, if I keep the definition of struct fenv_t in the header, I have to use the uint<size>_t types, should I move it in the source file and side-step that issue entirely, or keep it as is?

Can't move the struct definition to the source file due to there being no way to allocate it otherwise. We could just use raw types(`unsigned short`, `unsigned int`) but I like having explicitly sized types more in this case. (Also as a why I can't use `uN` types, `AK/Types.h` doesn't compile in C)

> Edit: this also does not touch anything related to SSE, though I think it should? not sure

I've made it keep track of mxscr :)

Also is my use of `asm volatile` too excessive? I wasn't entirely sure if I should just use `asm` or `asm volatile` so I went with the stronger option

ADKaster:
This is pretty sweet and does actually look like the correct implementation for x86. However, comma, 

Looking around it seems like clang and gcc are.. less than feature complete with regards to how user programs are supposed to interact with this header?

https://bugs.llvm.org/show_bug.cgi?id=8100
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=34678

It seems that the standard C99 pragmas, `#pragma FENV_ACCESS ON` and `#pragma FENV_ACCESS OFF` are not particularly supported by the compilers we're using with serenity's codebase. So, any userspace code intending to manipulate the floating point environment behind the back of the compiler using these functions will need a massive caveat emptor to not have its constants propagated incorrectly or its possibly trapping code deleted for having undefined behavior in the default rounding mode. Of course, with a homegrown compiler and some enthusiastic floating point folks that might become moot in the future :).

RealKC:
> Looking around it seems like clang and gcc are.. less than feature complete with regards to how user programs are supposed to interact with this header?

Yeah reading the cppreference page on this part of the stdlib I noticed that GCC and Clang didn't support the `#pragma`, but it mentions that they offer the header anyway (which is true, it's installed on my linux box)

PR #5505, opened 2021-02-24
LibELF: Convert more string literals to StringView literals.

bgianfo:
Most of these won't have perf impact, but the optimization is
practically free, so no harm in fixing these up.

### DIFF
--- a/Userland/Libraries/LibELF/DynamicLinker.cpp
+++ b/Userland/Libraries/LibELF/DynamicLinker.cpp
@@ -161,19 +161,19 @@ static void initialize_libc(DynamicObject& libc)
     // we have to initialize libc just after it is loaded.
     // Also, we can't just mark `__libc_init` with "__attribute__((constructor))"
     // because it uses getenv() internally, so `environ` has to be initialized before we call `__libc_init`.
-    auto res = libc.lookup_symbol("environ");
+    auto res = libc.lookup_symbol("environ"sv);
     VERIFY(res.has_value());
     *((char***)res.value().address.as_ptr()) = g_envp;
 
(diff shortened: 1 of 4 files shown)


alimpfard:
Before we go on an `operator""sv` spree, why not provide a `StringView(const char(&)[N])`?

awesomekling:
Oh sure, that'd be nicer.

bgianfo:
You can't because then you end up converting arrays which aren't necessarily filled to the size of the storage... so it's impossible to calculate length without strlen.
I already went down this hole :)

bgianfo:
Take this code in the kernel for an example, I had an assert in my StringView constructor (what you suggested @alimpfard) that validate strlen vs size. 

https://github.com/SerenityOS/serenity/blob/7db8ccc0e47e09e6c5d015a042c374b43cdd43ed/Kernel/Net/LocalSocket.cpp#L151-L161

alimpfard:
Hmmm...but an array ref isn't constexpr - maybe we can use that somehow...:thinking:
I'll take a little look down the hole too :)

bgianfo:
> I'll take a little look down the hole too :)

I hope there is a way that I missed, but after staring at crashes for a while I realized what I was doing wasn't going to work. 
I'm sure you can prove me wrong 😁

alimpfard:
Appears that it's not as straightforward as I thought, there's no real way to distinguish between the two as 'constexpr'-ness seems to disappear when you're inside a non-constexpr function (reasonable, maybe) - but if you're inside a constexpr function, looking at the array that was passed to you will blow up unconditionally if it were to be not a constant expression...

TL;DR, I guess we're going on an `operator""sv` spree.

PR #5506, opened 2021-02-24
KUBSAN: Condense all options down to "undefined"

Hendiadyoin1:
Now that there is enough memory and all fsanitize flags are working
we can condense all options back to one

### DIFF
--- a/Kernel/CMakeLists.txt
+++ b/Kernel/CMakeLists.txt
@@ -273,36 +273,6 @@ set(SOURCES
     ${CRYPTO_SOURCES}
     ${C_SOURCES}
 )
-set(KERNEL_FSANITIZE_FLAGS
-    "shift"
-    "shift-exponent"
-    "shift-base"
-    "integer-divide-by-zero"
-    "unreachable"
-    "vla-bound"
-    "return"
-    "signed-integer-overflow"
-    "bounds"
-    "bounds-strict"
-    "object-size"
-    "float-divide-by-zero"
-    "float-cast-overflow"
-    "nonnull-attribute"
-    "returns-nonnull-attribute"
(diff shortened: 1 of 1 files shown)


Kernel/CMakeLists.txt
```diff
@@ -273,46 +273,17 @@ set(SOURCES
...
-    endforeach()
-endif()
 
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-unknown-warning-option")
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -pie -fPIE -fno-rtti -ffreestanding -fbuiltin")
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -mno-80387 -mno-mmx -mno-sse -mno-sse2")
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fno-asynchronous-unwind-tables")
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fstack-protector-strong")
 if (NOT ${CMAKE_HOST_SYSTEM_NAME} MATCHES SerenityOS)
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -nostdlib -nostdinc -nostdinc++")
```
awesomekling, line 313: Why this indentation change?

Hendiadyoin1: I don't know, but I'll fix that probably today

Hendiadyoin1: Should be fixed now,  
Additionally I've readded the compiler check for fsanitize

PR #5508, opened 2021-02-24
Calculator: Add support for copy+paste using system clipboard

bcoles:


Userland/Applications/Calculator/CalculatorWidget.cpp
```diff
@@ -147,6 +147,18 @@ void CalculatorWidget::add_digit_button(GUI::Button& button, int digit)
...
 }
 
+String CalculatorWidget::get_entry()
+{
+    return m_entry->text();
+}
+
+void CalculatorWidget::set_entry(const StringView& text)
+{
+    double new_entry = atof(text.characters_without_null_termination());
```
awesomekling, line 157: As the name says, `characters_without_null_termination()` is not guaranteed to be null-terminated. Hence it's not safe to pass it to an API like `atof()` that expects a null-terminated string.
Avoid this by making the function take a `String` instead of a `StringView` :)

PR #5510, opened 2021-02-24
Kernel: allow munmap to unmap multiple pages

### LINKED ISSUE #175: Kernel: munmap() should handle partial and multiple regions
If I understand POSIX correctly, `munmap()` should be able to both unmap part of an existing region, possibly turning it into two regions with a hole in the middle.

It should also be able to unmap multiple regions if the specified range spans multiple regions.

See: http://pubs.opengroup.org/onlinepubs/9699919799/functions/munmap.html

Hendiadyoin1:
This should fix #175, wich is lingering in the issue box since mid 2019

The implemetation is probably not efficient, because I dont ~know if the regions are sorted by adress in their list, if yes we could make it O(n) instead of  probably O(n^2)~ use the cache

Please check my logic and standard complience.

### DIFF
--- a/Kernel/Syscalls/mmap.cpp
+++ b/Kernel/Syscalls/mmap.cpp
@@ -493,9 +493,6 @@ KResultOr<int> Process::sys$munmap(void* addr, size_t size)
     // Try again while checkin multiple regions at a time
     // slow: without caching
     const auto& regions = space().find_regions_intersecting(range_to_unmap);
-    // if there still no regions found error out
-    if (!regions.size())
-        return EINVAL;
 
     // check if any of the regions is not mmaped, to not accientally
     // error-out with just half a region map left
@@ -528,6 +525,7 @@ KResultOr<int> Process::sys$munmap(void* addr, size_t size)
     for (auto* new_region : new_regions) {
         new_region->map(space().page_directory());
     }
+
     return 0;
 }
 
--- a/Kernel/Syscalls/mmap.cpp
+++ b/Kernel/Syscalls/mmap.cpp
@@ -1,5 +1,6 @@
 /*
  * Copyright (c) 2018-2021, Andreas Kling <kling@serenityos.org>
+ * Copyright (c) 2021, Leon Albrecht <leon2002.la@gmail.com>
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
@@ -489,9 +490,45 @@ KResultOr<int> Process::sys$munmap(void* addr, size_t size)
         return 0;
     }
 
-    // FIXME: We should also support munmap() across multiple regions. (#175)
+    // Try again while checkin multiple regions at a time
+    // slow: without caching
+    const auto& regions = space().find_regions_intersecting(range_to_unmap);
+    // if there still no regions found error out
+    if (!regions.size())
+        return EINVAL;
 
-    return EINVAL;
+    // check if any of the regions is not mmaped, to not accientally
+    // error-out with just half a region map left
+    for (auto* region : regions) {
+        if (!region->is_mmap())
+            return EPERM;
+    }
+
+    Vector<Region*, 2> new_regions;
+
+    for (auto* old_region : regions) {
+        // if it's a full match we can delete the complete old region
+        if (old_region->range().intersect(range_to_unmap).size() == old_region->size()) {
+            bool res = space().deallocate_region(*old_region);
+            VERIFY(res);
+            continue;
+        }
+
+        // otherwise just split the regions and collect them for future mapping
+        new_regions.append(space().split_region_around_range(*old_region, range_to_unmap));
+
+        // We manually unmap the old region here, specifying that we *don't* want the VM deallocated.
+        old_region->unmap(Region::ShouldDeallocateVirtualMemoryRange::No);
+        bool res = space().deallocate_region(*old_region);
+        VERIFY(res);
+    }
+    // Instead we give back the unwanted VM manually at the end.
+    space().page_directory().range_allocator().deallocate(range_to_unmap);
+    // And finally we map the new region(s) using our page directory (they were just allocated and don't have one).
+    for (auto* new_region : new_regions) {
+        new_region->map(space().page_directory());
+    }
+    return 0;
 }
 
 KResultOr<FlatPtr> Process::sys$mremap(Userspace<const Syscall::SC_mremap_params*> user_params)
--- a/Kernel/VM/Range.cpp
+++ b/Kernel/VM/Range.cpp
@@ -1,5 +1,6 @@
 /*
  * Copyright (c) 2018-2021, Andreas Kling <kling@serenityos.org>
+ * Copyright (c) 2021, Leon Albrecht <leon2002.la@gmail.com>
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
@@ -42,6 +43,16 @@ Vector<Range, 2> Range::carve(const Range& taken) const
     if (taken.end() < end())
         parts.append({ taken.end(), end().get() - taken.end().get() });
     return parts;
+}
+Range Range::intersect(const Range& other) const
+{
+    if (*this == other) {
+        return *this;
+    }
+    auto new_base = max(base(), other.base());
+    auto new_end = min(end(), other.end());
+    VERIFY(new_base < new_end);
+    return Range(new_base, (new_end - new_base).get());
 }
 
 }
--- a/Kernel/VM/Space.cpp
+++ b/Kernel/VM/Space.cpp
@@ -1,5 +1,6 @@
 /*
  * Copyright (c) 2021, Andreas Kling <kling@serenityos.org>
+ * Copyright (c) 2021, Leon Albrecht <leon2002.la@gmail.com>
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
@@ -156,6 +157,27 @@ Region* Space::find_region_containing(const Range& range)
             return &region;
     }
     return nullptr;
+}
+
+Vector<Region*> Space::find_regions_intersecting(const Range& range)
+{
+    Vector<Region*> regions = {};
+    size_t total_size_collected = 0;
+
+    ScopedSpinLock lock(m_lock);
+
+    // FIXME: Maybe take the cache from the single lookup?
+    for (auto& region : m_regions) {
+        if (region.range().base() < range.end() && region.range().end() > range.base()) {
+            regions.append(&region);
+
+            total_size_collected += region.size() - region.range().intersect(range).size();
+            if (total_size_collected == range.size())
+                break;
+        }
+    }
+
+    return regions;
 }
 
 Region& Space::add_region(NonnullOwnPtr<Region> region)
--- a/Kernel/VM/Space.h
+++ b/Kernel/VM/Space.h
@@ -1,5 +1,6 @@
 /*
  * Copyright (c) 2018-2021, Andreas Kling <kling@serenityos.org>
+ * Copyright (c) 2021, Leon Albrecht <leon2002.la@gmail.com>
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
@@ -27,6 +28,7 @@
 #pragma once
 
 #include <AK/NonnullOwnPtrVector.h>
+#include <AK/Vector.h>
 #include <AK/WeakPtr.h>
 #include <Kernel/UnixTypes.h>
 #include <Kernel/VM/AllocationStrategy.h>
@@ -62,6 +64,8 @@ public:
 
     Region* find_region_from_range(const Range&);
     Region* find_region_containing(const Range&);
+
+    Vector<Region*> find_regions_intersecting(const Range&);
 
     bool enforces_syscall_regions() const { return m_enforces_syscall_regions; }
     void set_enforces_syscall_regions(bool b) { m_enforces_syscall_regions = b; }
--- a/Userland/Tests/Kernel/munmap-multi-region-unmapping.cpp
+++ b/Userland/Tests/Kernel/munmap-multi-region-unmapping.cpp
@@ -1,0 +1,126 @@
+/*
+ * Copyright (c) 2021, the SerenityOS developers.
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include <AK/Types.h>
+#include <fcntl.h>
+#include <stdio.h>
+#include <string.h>
+#include <sys/mman.h>
+#include <unistd.h>
+
+int main()
+{
+    {
+        printf("Testing full unnmap\n");
+        auto* map1 = mmap(nullptr, 2 * PAGE_SIZE, PROT_READ | PROT_WRITE, MAP_ANONYMOUS | MAP_PRIVATE | MAP_FIXED, 0, 0);
+        if (map1 == MAP_FAILED) {
+            perror("mmap 1");
+            return 1;
+        }
+        auto* map2 = mmap((void*)((FlatPtr)map1 + 2 * PAGE_SIZE), 2 * PAGE_SIZE, PROT_READ | PROT_WRITE, MAP_ANONYMOUS | MAP_PRIVATE | MAP_FIXED, 0, 0);
+        if (map2 == MAP_FAILED) {
+            perror("mmap 2");
+            return 1;
+        }
+
+        ((u32*)map1)[0] = 0x41414141;
+        ((u32*)map1)[PAGE_SIZE / sizeof(u32)] = 0x42424242;
+
+        ((u32*)map2)[0] = 0xbeefbeef;
+        ((u32*)map2)[PAGE_SIZE / sizeof(u32)] = 0xc0dec0de;
+
+        if (((u32*)map1)[0] != 0x41414141 || ((u32*)map1)[PAGE_SIZE / sizeof(u32)] != 0x42424242
+            || ((u32*)map2)[0] != 0xbeefbeef || ((u32*)map2)[PAGE_SIZE / sizeof(u32)] != 0xc0dec0de) {
+            perror("write");
+            return 1;
+        }
+
+        int res = munmap(map1, 4 * PAGE_SIZE);
+        if (res < 0) {
+            perror("unmap");
+            return 1;
+        }
+    }
+    {
+        printf("Testing partial unmaping\n");
+        auto* map1 = mmap(nullptr, 2 * PAGE_SIZE, PROT_READ | PROT_WRITE, MAP_ANONYMOUS | MAP_PRIVATE | MAP_FIXED, 0, 0);
+        if (map1 == MAP_FAILED) {
+            perror("mmap 1");
+            return 1;
+        }
+        auto* map2 = mmap((void*)((FlatPtr)map1 + 2 * PAGE_SIZE), 2 * PAGE_SIZE, PROT_READ | PROT_WRITE, MAP_ANONYMOUS | MAP_PRIVATE | MAP_FIXED, 0, 0);
+        if (map2 == MAP_FAILED) {
+            perror("mmap 2");
+            return 1;
+        }
+
+        ((u32*)map1)[0] = 0x41414141;
+        ((u32*)map1)[PAGE_SIZE / sizeof(u32)] = 0x42424242;
+
+        ((u32*)map2)[0] = 0xbeefbeef;
+        ((u32*)map2)[PAGE_SIZE / sizeof(u32)] = 0xc0dec0de;
+
+        if (((u32*)map1)[0] != 0x41414141 || ((u32*)map1)[PAGE_SIZE / sizeof(u32)] != 0x42424242
+            || ((u32*)map2)[0] != 0xbeefbeef || ((u32*)map2)[PAGE_SIZE / sizeof(u32)] != 0xc0dec0de) {
+            perror("write");
+            return 1;
+        }
+
+        int res = munmap((void*)((FlatPtr)map1 + PAGE_SIZE), 2 * PAGE_SIZE);
+        if (res < 0) {
+            perror("unmap");
+            return 1;
+        }
+
+        auto* map3 = mmap((void*)((FlatPtr)map1 + PAGE_SIZE), PAGE_SIZE, PROT_READ | PROT_WRITE, MAP_ANONYMOUS | MAP_PRIVATE | MAP_FIXED, 0, 0);
+        if (map3 == MAP_FAILED) {
+            perror("remap 1");
+            return 1;
+        }
+        auto* map4 = mmap(map2, PAGE_SIZE, PROT_READ | PROT_WRITE, MAP_ANONYMOUS | MAP_PRIVATE | MAP_FIXED, 0, 0);
+        if (map4 == MAP_FAILED) {
+            perror("remap 2");
+            return 1;
+        }
+        ((u32*)map3)[0] = 0x13371337;
+        ((u32*)map4)[0] = 0x1b1b1b1b;
+        if (((u32*)map1)[0] != 0x41414141 || ((u32*)map2)[PAGE_SIZE / sizeof(u32)] != 0xc0dec0de
+            || ((u32*)map3)[0] != 0x13371337 || ((u32*)map4)[0] != 0x1b1b1b1b
+            || ((u32*)map1)[PAGE_SIZE / sizeof(int)] != ((u32*)map3)[0] || ((u32*)map2)[0] != ((u32*)map4)[0]) {
+            perror("read at old map and write at remap");
+            return 1;
+        }
+
+        res = munmap(map1, PAGE_SIZE * 4);
+        if (res < 0) {
+            perror("cleanup");
+            return 1;
+        }
+    }
+
+    printf("PASS\n");
+    return 0;
+}
--- a/Kernel/VM/Range.h
+++ b/Kernel/VM/Range.h
@@ -1,5 +1,6 @@
 /*
  * Copyright (c) 2018-2021, Andreas Kling <kling@serenityos.org>
+ * Copyright (c) 2021, Leon Albrecht <leon2002.la@gmail.com>
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
@@ -67,6 +68,7 @@ public:
     }
 
     Vector<Range, 2> carve(const Range&) const;
+    Range intersect(const Range&) const;
 
 private:
     VirtualAddress m_base;


Kernel/VM/Space.cpp
```diff
@@ -158,6 +158,63 @@ Region* Space::find_region_containing(const Range& range)
...
+                    // If yes than we do not need to break here
+                    found_one = true;
+                    break;
+                } else
+                    return {};
+        }
+    }
+    return regions;
+}
+Vector<Region*> Space::find_regions_containing(Range* range)
```
BenWiederhake, line 194: This duplicates a lot of code. Is this really necessary?

BenWiederhake:
Even if they aren't sorted, I don't see why this should take O(n²). Filter all Regions intersecting with the desired range into a vector, sort the vector, then check whether the entire range is covered. If the regions already are sorted, then obviously the sorting step becomes superfluous. Also, why does `sys$unmap` have two separate paths for that?

Finally, a nit: "*Keen*el" in the first commit should probably be "Kernel" :P

Kernel/VM/Space.cpp
Hendiadyoin1: maybe, maybe not it was the one thing that came to my mind looking at the old code, surely you could try to break it out into a lot of samller funtions, but that would be spaghtti code.

If you have another idea, how to solve this I'd like to here it, I'm not exactly a CS student..

Hendiadyoin1:
Ok I've fixed the typo, and will look into your implementation idea

Kernel/VM/Space.cpp
Hendiadyoin1: The implementation should now be shorter and as you have discibed in your comment

Hendiadyoin1:
This with the two paths we can make it one, because in the second one we still have to check if we unmap the whole Region and than take the same path

Kernel/VM/Space.cpp
tomuta: Why not just a `for_each_intersected` template function that just iterates over them and only handles the intersecting regions?

Hendiadyoin1: because with that we could not check if we bailed out with EPERM or if there actually were some regions to unmap

tomuta: Not following. Just capture a local variable by reference in the iteration lambda?

Hendiadyoin1: Okay ive changed it to a lambda based aproach, though I don't think the code readability nor the performance of that code really changed thourgh that

Kernel/VM/Space.cpp
```diff
@@ -202,6 +202,14 @@ Vector<Region*> Space::find_regions_containing(const Range& range)
 
     return regions;
 }
+template<typename Callback>
+void Space::for_each_region_intersecting(const Range& range, Callback callback)
+{
+    for (auto* region : find_regions_intersecting(range)) {
```
tomuta, line 208: The idea for using a `for_each_...` was to essentially to not implement `find_regions_intersecting` so that instead of putting the regions in a vector you just iterate them, and for those that do intersect you simply call `callback`.

tomuta (changes requested):

BenWiederhake (changes requested):
Hi! You're awesome! Please don't get demotivated by this (hopefully constructive) criticism of the code. You tried, and it shows! :D

Most importantly, I would like to suggest that you read up on `git-rebase`, because that can make your commit history much more readable than the current "okay here's two functions wait nvm only one function actually lemme rename it wait now it does something different" which is a bit hard to follow.

I think the current approach of the PR is in the wrong direction: Two different, but mostly duplicated code paths to determine two different, but mostly similar results, and eventually two different, but mostly duplicated for-loops to deallocate everything.

How about a slightly different approach: There must exist code anyway has to deal with "arbitrary worst-case scenarios" anyway. So let's write that code path first. And *then*, if you're worried about performance, only then write the code that checks for the happy case, the "munmap a range and there's only a single region of exactly that size anyway" case.

So that would probably be something like:
- Find all regions that intersect with the range in question
- If any of them cause `EINVAL`, do it now
- If any regions straddle the ends of `range`, do a `split_region_around_range` first
- Deallocate regions that are wholly inside the range

That handles all cases the same. Note that this handles the "single region exactly matching the range" case, without it being special in any way! :D
Having no corner cases tends to be a really good idea, especially in contrast to the current "detect 4 different scenarios" approach.

I can totally understand if that's not you want to do though, so see below for detailed feedback on why certain things are likely to go wrong.

Kernel/Syscalls/mmap.cpp
```diff
@@ -488,8 +488,41 @@ int Process::sys$munmap(void* addr, size_t size)
...
+
+            auto new_regions = space().split_region_around_range(*old_region, range_to_unmap);
+
+            // We manually unmap the old region here, specifying that we *don't* want the VM deallocated.
+            old_region->unmap(Region::ShouldDeallocateVirtualMemoryRange::No);
+            space().deallocate_region(*old_region);
 
-    // FIXME: We should also support munmap() across multiple regions. (#175)
+            // Instead we give back the unwanted VM manually.
+            space().page_directory().range_allocator().deallocate(range_to_unmap);
```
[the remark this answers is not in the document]
Hendiadyoin1: Thats a part I've copied staight out of the implementation of the earlyer ones, so "like I'd know!"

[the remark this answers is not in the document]
BenWiederhake: Yup, in the original implementation it was only called once, so why is it now called multiple times? I don't exactly know either what it does, but it seems unwise to change it without having a clue what it does :P

Kernel/VM/Space.cpp
Hendiadyoin1: ...

Kernel/VM/RangeAllocator.cpp
```diff
@@ -75,8 +75,12 @@ Vector<Range, 2> Range::carve(const Range& taken) const
         return {};
     if (taken.base() > base())
         parts.append({ base(), taken.base().get() - base().get() });
-    if (taken.end() < end())
-        parts.append({ taken.end(), end().get() - taken.end().get() });
+    if (taken.end() < end()) {
```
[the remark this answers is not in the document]
Hendiadyoin1: I've changed that to a VERIFY (a just wanted to write Assertion, but thats not "true" any more...), that the taken range is in range

Kernel/VM/Space.cpp
```diff
@@ -158,61 +158,44 @@ Region* Space::find_region_containing(const Range& range)
...
 
-    bool found_one = true;
-    while (range && found_one) {
-        found_one = false;
-        for (auto& region : m_regions) {
-            if (region.contains(*range)) {
+    // FIXME: Maybe take the cache from the single lookup?
+    for (auto& region : m_regions) {
+        if ((region.vaddr() >= range.base() && region.vaddr() <= range.end())
+            || (region.range().end() <= range.end() && region.range().end() >= range.base())) {
```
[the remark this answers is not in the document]
Hendiadyoin1: Yep, that makes sense...

Kernel/VM/Space.cpp
```diff
@@ -158,61 +158,44 @@ Region* Space::find_region_containing(const Range& range)
     return nullptr;
 }
 
-Vector<Region*> Space::find_regions_from_range(Range* range)
+Vector<Region*> Space::find_regions_from_range(const Range& range)
 {
     Vector<Region*> regions = {};
+    size_t size = page_round_up(range.size());
+    size_t total_size_collected = 0;
```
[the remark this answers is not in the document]
Hendiadyoin1: I've squashed them (was wierd I needed like 3 attempts...)

Kernel/VM/RangeAllocator.cpp
```diff
@@ -84,6 +84,17 @@ Vector<Range, 2> Range::carve(const Range& taken) const
...
 
+Range Range::intersect(const Range& other) const
+{
+    if (*this == other) {
+        return *this;
+    }
+    auto new_base = max(base(), other.base());
+    auto new_end = min(min(end(), other.end()), base());
+    VERIFY(new_base < new_end);
+    return Range(new_base, (new_end - new_end).get());
```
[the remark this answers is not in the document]
Hendiadyoin1: I've even found these typos my self, remembering I wrote two min's and they should be fixed now

Kernel/Syscalls/mmap.cpp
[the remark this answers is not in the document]
Hendiadyoin1: It could complain that its deallocating a range that does not exist...

Kernel/Syscalls/mmap.cpp
```diff
@@ -489,25 +489,14 @@ int Process::sys$munmap(void* addr, size_t size)
...
-        }
-        return 0;
-    }
-
+    // slow: without caching
     auto old_regions = space().find_regions_containing(range_to_unmap);
     if (old_regions.size()) {
         for (auto* old_region : old_regions) {
             if (!old_region->is_mmap())
                 return -EPERM;
```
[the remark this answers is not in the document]
Hendiadyoin1: that we should do...

Kernel/VM/Space.cpp
```diff
@@ -158,41 +158,45 @@ Region* Space::find_region_containing(const Range& range)
...
         if ((region.vaddr() >= range.base() && region.vaddr() <= range.end())
             || (region.range().end() <= range.end() && region.range().end() >= range.base())) {
-            if (region.size() == size) {
-                regions.append(&region);
-            }
+            regions.append(&region);
+
+            total_size_collected += region.size() - region.range().intersect(range).size();
+            if (total_size_collected == range.size())
+                break;
```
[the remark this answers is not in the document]
Hendiadyoin1: 1st _if_, we should ask someone more expirienced with this code about it though

2nd thats why I'm using the intersection size or is my logic wrong

Kernel/VM/Space.cpp
```diff
@@ -158,6 +158,63 @@ Region* Space::find_region_containing(const Range& range)
...
+    while (range && found_one) {
+        found_one = false;
+
+        for (auto& region : m_regions) {
+            if (region.vaddr() == range->base())
+                if (region.size() == size) {
+                    regions.append(&region);
+
+                    auto split_range = range->carve(region.range());
+                    range = split_range.is_empty() ? &(split_range.last()) : nullptr;
```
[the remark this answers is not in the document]
Hendiadyoin1: I think that piece of code was for a exit condition to tell the while loop to breakout, because we used up all of the range to unmap, but this pice of code isnt used anymore

Kernel/VM/Space.cpp
```diff
@@ -158,41 +158,45 @@ Region* Space::find_region_containing(const Range& range)
     return nullptr;
 }
 
+/**
+* @brief returns all whole regions in range and only when all matches are full matches
```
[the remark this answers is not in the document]
Hendiadyoin1: I've removed this helper, because I didn't need it anymore...

And I've renamed the second more general helper to something simmilar to your suggestion

Kernel/VM/Space.cpp
Hendiadyoin1: This isnt viable anymore due to needing two loops in the new implementation due to safety reasons

Kernel/Syscalls/mmap.cpp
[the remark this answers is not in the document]
Hendiadyoin1: This should be fixed now, we do that now at the end of the loop and only after that we remap the newly created regions

Kernel/VM/Space.cpp
```diff
@@ -304,5 +361,4 @@ size_t Space::amount_purgeable_nonvolatile() const
     }
     return amount;
 }
-
```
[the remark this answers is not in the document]
linusg: > This should trip the clang-format lint

No, (sadly) clang-format (or just our config?) doesn't care about having an empty line around namespaces.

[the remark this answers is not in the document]
BenWiederhake: More importantly: Does anyone know why CI doesn't run on this PR?

Kernel/VM/Space.cpp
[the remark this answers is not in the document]
BenWiederhake: Yeah, it took me lots of mistakes until I finally got the hang of it. It gets easier over time, it'll end up feeling like second nature :)

Kernel/Syscalls/mmap.cpp
```diff
@@ -1,5 +1,6 @@
 /*
  * Copyright (c) 2018-2021, Andreas Kling <kling@serenityos.org>
+ * Copyright (c) 2018-2021, Leon Albrecht <leon2002.la@gmail.com>
```
BenWiederhake, line 3: You were'nt working on this in the years 2018, 2019, nor 2020. Please adjust accordingly.

Hendiadyoin1: Ooops I thought IÄve adjusted that,
gonna move that into the right commits

BenWiederhake:
"KERNEL: make munmap safer and propperly behaving" has two typos in the commit message.

Let's exercise git-rebase here:
- `git rebase origin/master -i`
- Wait for the editor to open
- Edit the `pick abc1234 KERNEL: make munmap safer and propperly behaving` to say "r" instead of "pick". "r" is short for "reword", see explanations in the comments beneath the commit commands.
- Close the editor
- Wait until it opens up again (it runs the commands in the previous file sequentially)
- Fix both typos
- Save and quit
- Wait for git-rebase to finish
- Profit!

Kernel/Syscalls/mmap.cpp
Hendiadyoin1: so, ehh fixed

Hendiadyoin1:
> 
> 
> "KERNEL: make munmap safer and propperly behaving" has two typos in the commit message.
> 
> Let's exercise git-rebase here:
> 
>     * `git rebase origin/master -i`
> 
>     * Wait for the editor to open
> 
>     * Edit the `pick abc1234 KERNEL: make munmap safer and propperly behaving` to say "r" instead of "pick". "r" is short for "reword", see explanations in the comments beneath the commit commands.
> 
>     * Close the editor
> 
>     * Wait until it opens up again (it runs the commands in the previous file sequentially)
> 
>     * Fix both typos
> 
>     * Save and quit
> 
>     * Wait for git-rebase to finish
> 
>     * Profit!

Squashed it into the previous one instead...

Kernel/Syscalls/mmap.cpp
```diff
@@ -489,9 +490,40 @@ int Process::sys$munmap(void* addr, size_t size)
         return 0;
     }
 
-    // FIXME: We should also support munmap() across multiple regions. (#175)
+    // Try again while checkin multiple regions at a time
+    // slow: without caching
+    const auto& regions = space().find_regions_intersecting(range_to_unmap);
+    // if there still no regions found error out
+    if (!regions.size())
+        return -EINVAL;
```
BenWiederhake, line 498: This decides *against* what POSIX says. I think this is perfectly reasonable, and we should do this in Serenity, so: Yeah, let's do this.

Hendiadyoin1: This is a reamnant from the original implementation, but ok...

Kernel/VM/Space.h
```diff
@@ -63,6 +65,11 @@ class Space {
     Region* find_region_from_range(const Range&);
     Region* find_region_containing(const Range&);
 
+    Vector<Region*> find_regions_intersecting(const Range&);
+
+    template<typename Callback>
+    void for_each_region_intersecting(const Range&, Callback);
```
BenWiederhake, line 71: I believe this cannot work, since the compiler can't see the template definition from here. And since it's not used anyway, let's delete it entirely.

(@tomuta suggested templates to generalize the duplicated code away. Now there is no duplication anymore, so there is no need for a template.)

Hendiadyoin1: Done, should I squash it into its creation?

Kernel/VM/Range.cpp
```diff
@@ -44,5 +45,17 @@ Vector<Range, 2> Range::carve(const Range& taken) const
         parts.append({ taken.end(), end().get() - taken.end().get() });
     return parts;
 }
+Range Range::intersect(const Range& other) const
+{
+    if (*this == other) {
+        return *this;
+    }
+    auto new_base = max(base(), other.base());
+    auto new_end = max(min(end(), other.end()), base());
```
BenWiederhake, line 54: This will crash sometimes.

Let `this` be the Range from `0x1000` to `0x2000`, and let `other` be the Range from `0x4000` to `0x7000`.
Then your code computes:
```
new_base = max(0x1000, 0x4000) = 0x4000
new_end = max(min(0x2000, 0x7000), 0x1000) = max(0x2000, 0x1000) = 0x2000
VERIFY(false);
```
Oops!

Here's a tip for life: If something is symmetric in mathematics (for example, computing the intersection of two intervals, "A intersect B" = "B intersect A"), then code that computes this symmetric operation should also be symmetric. That's not a hard rule, but it certainly raises an eyebrow if the code looks asymmetric. Case in point: `auto new_base = max(base(), other.base());` is obviously symmetric: If "this" and "other" are swapped, it obviously should produce the same result. Same for `min(end(), other.end())`. But `max(…, base());`?

This can be fixed, yes, but the only use is an incomplete micro-optimization in `find_regions_intersecting` as I already explained. (Either there is a much better one when it's sorted, or it cannot possibly save any significant amount of time.)

BenWiederhake:
"Kernel: Remove unused helper in VM::Space" → squash

"KERNEL: make munmap posix more complient" → Oh, alright. Either way, the spelling is still "Kernel", "Make", and "compliant", like all the other commits. Also, the word order got mixed up ("more posix compliant").

I realize that iterating on a PR is frustrating, but let's keep the commit messages readable please.

Kernel/VM/Range.cpp
Hendiadyoin1: Ok, I fixed that, and I realized, with a zero intersection, we would Crash, so the Verify should be  a `<=`

And I haven't made that function for preformance sake, but for readability and ease of use,
I think its easier to understand when I say that the intersection is eaqual, than saying are the bounds contained, or to calculate the intersection size at onepoint it gets down to an understandable oneliner

Hendiadyoin1:
> "KERNEL: make munmap posix more complient" → Oh, alright. Either way, the spelling is still "Kernel", "Make", and "compliant", like all the other commits. Also, the word order got mixed up ("more posix compliant").

Well its getting late, so my brain starts todo stuff weirder than usual

Hendiadyoin1:
Should be all fixed now, sadly nothing I've found may be able to use this feature at this point (atleast I havent found any workarounds), so a manual testscript may be of use

Kernel/VM/Space.cpp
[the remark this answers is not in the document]
Hendiadyoin1: weirdly enough it finds all kinds of stuff in AK and other locations I haven't touched... (atleast on my mashine)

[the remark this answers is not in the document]
Hendiadyoin1: Both occurences should be fixed

Kernel/VM/Range.cpp
BenWiederhake: So what exactly does the function assume now?

- If the Ranges are "very" disjoint, it crashes in `VERIFY`.
- If the Ranges are disjoint but "touch" (i.e. `end() == other.base()`), it passes, and (uhh, see below)

So does the function assume they intersect or not?

Also, there's a typo in the "return" line: `new_end - end()` should probably be `new_end - new_base`.

BenWiederhake:
> so a manual testscript may be of use

I'd love to persuade you to add another test, *especially* if this case is hard to hit. And apparently it's easy to make typos in the logic, as we have seen three times now. Sorry for the extra work, but if it already went wrong three times, I wouldn't want to trust the code without at least running it a few times.

Kernel/VM/Range.cpp
Hendiadyoin1: So it should always return actually even when the ranges are very disjoint all values get clamped to end(), 
if I'm not mistaken and thenn theres a 0-Range being returned...

And thanks for hinting at the typo, it would probably cause a very weird bug somewhere else...

Hendiadyoin1: And should we check for page alignment like carve does?

Hendiadyoin1:
So there's now a testscript and it only apears to run correctly on the new branch

Kernel/VM/Range.cpp
Hendiadyoin1: I've changed it, that incase of no match it just returns an invalid region

BenWiederhake: Yup, I think this code works. I would have written it differently, but it works, so I don't complain :D

Kernel/Syscalls/mmap.cpp
```diff
@@ -493,9 +493,6 @@ int Process::sys$munmap(void* addr, size_t size)
     // Try again while checkin multiple regions at a time
     // slow: without caching
     const auto& regions = space().find_regions_intersecting(range_to_unmap);
-    // if there still no regions found error out
-    if (!regions.size())
-        return -EINVAL;
```
BenWiederhake: Squash this into the previous commit (43fefc6d76c56a0965ba90b237d8513547860d23), which adds these lines.

(Or leave it out completely.)

Userland/Tests/Kernel/munmap-multi-region-unmapping.cpp
```diff
@@ -0,0 +1,80 @@
...
+        perror("open");
+        return 1;
+    }
+    {
+        auto* map1 = mmap(nullptr, size, PROT_READ, MAP_FILE | MAP_SHARED, fd, 0);
+        if (map1 == MAP_FAILED) {
+            perror("mmap11");
+            return 1;
+        }
+        auto* map2 = mmap((void*)((FlatPtr)map1 + size), size, PROT_READ, MAP_FILE | MAP_SHARED, fd, 0);
```
BenWiederhake, line 48: To make sure this doesn't bite us later on: Please add `MAP_FIXED`. Other than that, it looks good to me.

Userland/Tests/Kernel/munmap-multi-region-unmapping.cpp
```diff
@@ -0,0 +1,80 @@
...
+            perror("mmap12");
+            return 1;
+        }
+        auto* map2 = mmap((void*)((FlatPtr)map1 + size), size, PROT_READ, MAP_FILE | MAP_SHARED, fd, 0);
+        if (map2 == MAP_FAILED) {
+            perror("mmap22");
+            return 1;
+        }
+
+        int res = munmap((void*)((FlatPtr)map1 + size / 2), size);
```
BenWiederhake, line 72: I don't think this does what you think this does. [posix(ish) says](https://linux.die.net/man/3/munmap):

> The munmap() function shall remove any mappings for those entire pages containing any part of the address space of the process starting at addr and continuing for len bytes.

So, this should remove all mappings that contain any part of the [map1 + half-page, map1 + page + half-page] interval. Therefore the very first thing that munmap should do is rounding the start and end to [map1, map1 + 2 * page], so this ends up testing the exact same thing as the previous testcase.

It's good that you test this scenario: After all, this is the tricky case that would have triggered the previous bugs.

One approach would be to just mmap 2 pages each; it doesn't need a filedescriptor, so I wouldn't use one here. For the first mmap, the invocation should probably look like this: `mmap(nullptr, 2 * PAGE_SIZE, PROT_READ, MAP_ANONYMOUS | MAP_PRIVATE, 0, 0)`. Make sure to write to the pages to force the Kernel to actually allocate the memory, and read from the valid pages after deallocating. Only this way we can verify that the memory that should still be valid (the first and last page) is actually still valid.

Something like this:
```
map1 = mmap(nullptr, 2 * PAGE_SIZE, PROT_READ, MAP_ANONYMOUS | MAP_PRIVATE, 0, 0)
map2 = mmap(map2 + 2 * PAGE_SIZE, 2 * PAGE_SIZE, PROT_READ, MAP_ANONYMOUS | MAP_PRIVATE | MAP_FIXED, 0, 0)
((int*)map1)[0] = 0x12345678;
((int*)map1)[PAGE_SIZE / sizeof(int)] = 0xabcdef;
((int*)map2)[0] = 0x87654321;
((int*)map2)[PAGE_SIZE / sizeof(int)] = 0xfdebca;
munmap(map1 + PAGE_SIZE, PAGE_SIZE * 2)
VERIFY(((int*)map1)[0] == 0x12345678);
VERIFY(((int*)map2)[PAGE_SIZE / sizeof(int)] == 0xfdebca);
```

(Bonus points if you figure out how to check that the munmap'd space has actually been unmapped. I don't know how to do that, so it's fine to leave it unchecked.)

Kernel/Syscalls/mmap.cpp
BenWiederhake: Hmm, Github says my comment is "Outdated", but I'm pretty sure I commented on the right commit. If I just got confused, please ignore my comment.

Userland/Tests/Kernel/munmap-multi-region-unmapping.cpp
Hendiadyoin1: So we should round to pages when checking for intersections or when we lookup the regions intersecting? (if yes, wich ifnot both)

Hendiadyoin1: To check if they were unmapped we could just try to remap it, can't we?

BenWiederhake: > we could just try to remap it

Good idea!

Hendiadyoin1: > So we should round to pages when checking for intersections or when we lookup the regions intersecting? (if yes, wich ifnot both)

So well actually we expand the range_to_unmap before we even start looking for regions

Hendiadyoin1: The Tets does now checks all those things, and it worked without a hitch

BenWiederhake: @rounding: The Kernel already does, it's just your test program that forgot about it. :P

In the Kernel, this job is done by the `auto range_or_error = expand_range_to_page_boundaries((FlatPtr)addr, size);` line.

Userland/Tests/Kernel/munmap-multi-region-unmapping.cpp
```diff
@@ -0,0 +1,126 @@
...
+        }
+
+        int res = munmap(map1, 4 * PAGE_SIZE);
+        if (res < 0) {
+            perror("unmap");
+            return 1;
+        }
+    }
+    {
+        printf("Testing partial unmaping\n");
```
BenWiederhake, line 68: :+1:

Userland/Tests/Kernel/munmap-multi-region-unmapping.cpp
```diff
@@ -0,0 +1,126 @@
...
+        auto* map1 = mmap(nullptr, 2 * PAGE_SIZE, PROT_READ | PROT_WRITE, MAP_ANONYMOUS | MAP_PRIVATE | MAP_FIXED, 0, 0);
```
BenWiederhake, line 69: Does nullptr and `MAP_FIXED` actually work? That sounds like a bug in the Kernel, so let's not rely on it. (This affects the above `map1 = mmap(…` line, too.)

Userland/Tests/Kernel/munmap-multi-region-unmapping.cpp
```diff
@@ -0,0 +1,126 @@
...
+        if (map1 == MAP_FAILED) {
+            perror("mmap 1");
+            return 1;
+        }
+        auto* map2 = mmap((void*)((FlatPtr)map1 + 2 * PAGE_SIZE), 2 * PAGE_SIZE, PROT_READ | PROT_WRITE, MAP_ANONYMOUS | MAP_PRIVATE | MAP_FIXED, 0, 0);
```
BenWiederhake, line 74: :+1:

Userland/Tests/Kernel/munmap-multi-region-unmapping.cpp
Hendiadyoin1: I don't actually know how these parameters are supposed to work, but it runs without complaining

Hendiadyoin1: The [man 2](https://man7.org/linux/man-pages/man2/mmap.2.html) page says that you should omit the MAP_FIXED flag,
but it also states, that MAP_FIXED can be really forcefull and overwrite old maps..

BenWiederhake: Turns out, nullptr and MAP_FIXED triggers the `allocate_anywhere` branch, so this code is perfectly fine. I was wrong.

(Also, for future reference, our Kernel seems to reject trying to overwrite old maps.)

BenWiederhake:
From my point of view: This PR can be merged :)

Hendiadyoin1:
Then thanks for your advices, I've learned a lot!

Hendiadyoin1:
@awesomekling I think we only need your stamp of approval now!

Hendiadyoin1:
How nice its failing at a place I havent touched!

Kernel/VM/Space.cpp
```diff
@@ -158,6 +159,27 @@ Region* Space::find_region_containing(const Range& range)
     return nullptr;
 }
 
+Vector<Region*> Space::find_regions_intersecting(const Range& range)
```
awesomekling, line 162: This would be better written as `template<typename Callback> void Space::for_each_region_intersecting(const Range&, Callback)`. That way we don't incur any heap allocations (which we do here due to the use of `Vector`.)

Kernel/Syscalls/mmap.cpp
```diff
@@ -489,9 +490,42 @@ KResultOr<int> Process::sys$munmap(Userspace<void*> addr, size_t size)
...
+    for (auto* region : regions)
+        if (!region->is_mmap())
+            return EPERM;
+
+    Vector<Region*, 2> new_regions;
+
+    for (auto* old_region : regions) {
+        // if it's a full match we can delete the complete old region
+        if (old_region->range().intersect(range_to_unmap).size() == old_region->size()) {
+            VERIFY(space().deallocate_region(*old_region));
```
awesomekling, line 508: Please don't put effectful expressions inside `VERIFY()`.
While they are always compiled, it makes the code unnecessarily hard to read.

Kernel/Syscalls/mmap.cpp
```diff
@@ -489,9 +490,42 @@ KResultOr<int> Process::sys$munmap(Userspace<void*> addr, size_t size)
...
+            continue;
+        }
+
+        // otherwise just split the regions and collect them for future mapping
+        new_regions.append(space().split_region_around_range(*old_region, range_to_unmap));
+
+        // We manually unmap the old region here, specifying that we *don't* want the VM deallocated.
+        old_region->unmap(Region::ShouldDeallocateVirtualMemoryRange::No);
+        VERIFY(space().deallocate_region(*old_region));
```
awesomekling, line 517: Same comment here.

Kernel/Syscalls/mmap.cpp
```diff
@@ -489,9 +490,42 @@ KResultOr<int> Process::sys$munmap(Userspace<void*> addr, size_t size)
...
+    // Try again while checkin multiple regions at a time
+    // slow: without caching
+    const auto& regions = space().find_regions_intersecting(range_to_unmap);
+    // if there still no regions found error out
+    if (!regions.size())
+        return EINVAL;
 
-    return EINVAL;
+    // check if any of the regions is invalid in somesort, to not accientally
+    // crash with just half a region map left
```
awesomekling, line 498: I don't understand this comment. Why would we "crash with just half a region map left" if there's a non-mmap region in the range?

It's correct to check for this, as we don't want `munmap` to unmap the signal trampoline page (which is non-mmap.)

Kernel/Syscalls/mmap.cpp
```diff
@@ -489,9 +490,42 @@ KResultOr<int> Process::sys$munmap(Userspace<void*> addr, size_t size)
...
+    for (auto* region : regions)
+        if (!region->is_mmap())
+            return EPERM;
```
awesomekling, line 501: Please wrap the outer `for` block in `{}`

awesomekling (changes requested):

Kernel/VM/Space.cpp
BenWiederhake: In this particular case it can't be avoided: We first have to collect all relevant Regions, check whether they might cause `EINVAL` in munmap, and only afterwards deallocate them. One *could* replace that by calling `for_each_region_intersecting` twice, but that extra processing time seems silly to avoid a few bytes of temporary heap memory.

Hendiadyoin1: I recieved this advice before, but we decided to loop over the data two times, to avoid returning an error half way through reallocation, which could leave the memory pages in a bad state

Kernel/Syscalls/mmap.cpp
Hendiadyoin1: we check beforehand if any of the pages are not mmap, to avoid an error halfway through, at which point some pages would have been unmapped and others won't, or worse we split a range and haven't set it up correctly again.

If we now at which range the signal trampoline is wecould check with intersection and then use the lambda aproach as stated above

Kernel/VM/Space.cpp
awesomekling: Right, looping is performed twice anyway. Also `kmalloc()` memory, once owned, is cheap, but calling `kmalloc()` and `kfree()` is quite slow and both take a global spinlock for the duration.

Kernel/Syscalls/mmap.cpp
awesomekling: This loop is fine, I just didn't understand the comment since it talks about crashing which is a specific thing in this context (`Process::crash()`) and it didn't seem to apply :)

Hendiadyoin1: Is the signal trampoline the only not mmap region in a process?

awesomekling: No, there is also the original executable/loader mappings, as well as the main thread stack.

Kernel/VM/Range.cpp
```diff
@@ -43,5 +44,20 @@ Vector<Range, 2> Range::carve(const Range& taken) const
...
+    if (*this == other) {
+        return *this;
+    }
+    auto new_base = min(max(base(), other.base()), end());
+    auto new_end = max(min(end(), other.end()), base());
+    auto new_size = (new_end - new_base).get();
+
+    // if we have no match, return an invalid range
+    if (!new_size)
+        return Range(VirtualAddress(nullptr), 0ULL);
```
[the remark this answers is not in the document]
Hendiadyoin1: Done.

Kernel/Syscalls/mmap.cpp
Hendiadyoin1: Done.

Kernel/Syscalls/mmap.cpp
Hendiadyoin1: Done

Kernel/Syscalls/mmap.cpp
Hendiadyoin1: So explicitly checking for these wouldnt be woth it?

Do we even need the original executable/loader mappings at whis point?

Kernel/Syscalls/mmap.cpp
Hendiadyoin1: Done.

Kernel/Syscalls/mmap.cpp
Hendiadyoin1: At least I've adjusted that wording now to some degree to
>  check if any of the regions is not mmaped, to not accientally error-out with just half a region map left

Kernel/VM/Space.cpp
Hendiadyoin1: So do I need to change something?

If you are referring to this line
```cpp
// otherwise just split the regions and collect them for future mapping new_regions.append(space().split_region_around_range(*old_region, range_to_unmap));

```
split_region is stack allocated and guaranteed to be of size one, and the new_region vector is also stack allocated and has enough stack-capacity to hold both possible split regions.

Hendiadyoin1:
I'll squash evrything down again, when the current Conversations are resolved

Kernel/VM/Range.cpp
```diff
@@ -55,7 +55,7 @@ Range Range::intersect(const Range& other) const
 
     // if we have no match, return an invalid range
     if (!new_size)
-        return Range(VirtualAddress(nullptr), 0ULL);
+        return {};
```
BenWiederhake: Given that this function is now only used in a context where we expect the ranges to have non-zero intersection and would actually crash the kernel (`Optional::operator->` for `total_size_collected`), one could just move the VERIFY here, and simplify some other stuff along the way:

```
    auto new_base = max(base(), other.base());
    auto new_end = min(end(), other.end());
    VERIFY(new_base < new_end);
    return Range(new_base, new_end - new_size);
```

(And therefore no `Optional` anymore.)

Kernel/VM/Space.cpp
BenWiederhake: Implementing `for_each_region_intersecting` will be non-trivial if you want to be able to loop over it and remove from the underlying `NonnullOwnPtrVector<Region> m_regions` at the same time. @awesomekling

Kernel/VM/Range.cpp
Hendiadyoin1: Done that

Kernel/VM/Space.cpp
Hendiadyoin1: So nothing to do for me here?

BenWiederhake:
Hmm, it says "1 change requested", but kling's request has already been addressed. Is this a glitch in the Github UI?

Hendiadyoin1:
> 
> 
> Hmm, it says "1 change requested", but kling's request has already been addressed. Is this a glitch in the Github UI?

On the preview it says Tomuta and on the summary on the bottom it just says Awesomekling, it apears that didnt press the request re-revie button hard enough last time

PR #5511, opened 2021-02-24
PixelPaint: Bug fixes

### LINKED ISSUE #5475: PixelPaint: Saving project with invisible layer causes crash when reopening
**To reproduce:**
1. Open PixelPaint.
2. Using the default project with the two squares, select the background layer and uncheck "Visible".
3. _Save As..._ (.pp).
4. _Open..._ the file you just saved.

The steps above cause the assertion in `../Userland/Applications/PixelPaint/Image.cpp:209` to fail.

I'm pretty sure this issue is related to the same bug which causes #5377 which I am also working on, trying to fix.

IvanHansgaardHansen:
This PR fixes #5475 and another bug, where opened .pp projects would be uninteractable before a layer was actively (re)selected, even though the last used layer (from when the .pp project was saved) was highlighted in the list on the right.

### DIFF
--- a/Userland/Applications/PixelPaint/Image.cpp
+++ b/Userland/Applications/PixelPaint/Image.cpp
@@ -91,7 +91,10 @@ RefPtr<Image> Image::create_from_file(const String& file_path)
         auto width = json_layer_object.get("width").to_i32();
         auto height = json_layer_object.get("height").to_i32();
         auto name = json_layer_object.get("name").as_string();
+
         auto layer = Layer::create_with_size(*image, { width, height }, name);
+        image->add_layer(*layer);
+
         layer->set_location({ json_layer_object.get("locationx").to_i32(), json_layer_object.get("locationy").to_i32() });
         layer->set_opacity_percent(json_layer_object.get("opacity_percent").to_i32());
         layer->set_visible(json_layer_object.get("visible").as_bool());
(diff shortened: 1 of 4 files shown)


IvanHansgaardHansen:
Any specific reason this is not getting merged? Or are we just busy? 😄

bcoles:
> 
> 
> Any specific reason this is not getting merged? Or are we just busy? smile

Probably just busy and no one has reviewed yet. The code looks fine at first glance (untested) and this PR allegedly (untested) does something useful (fixes a bug).

However the commit messages are not in line with the [contribution guidelines](https://github.com/SerenityOS/serenity/blob/master/CONTRIBUTING.md):

> Wrap your commit messages at 72 characters.

Additionally, the first line of the first commit "`PixelPaint: Fix issue #5475 (invisible layers causing crashes)`" is arguably in violation of the following guidance:

> The first line of the commit message should have the format "Category: Brief description of what's being changed". The "category" can be a subdirectory, but also something like "POSIX compliance" or "ClassName". Whatever seems logical.

Something like this would be more appropriate:

```
PixelPaint: Always add a layer to the image before accessing layer methods
```

linusg:
@IvanHansgaardHansen do you intend to make any changes following the above comment? Also, there's a merge conflict that needs to be resolved.

PR #5514, opened 2021-02-25
LibGUI: Paint all widgets as toolbars in ToolBarContainer

thankyouverycool:
A while ago ToolBarContainer started checking for child widget type and stopped drawing the bookmark bar with proper threading. At first I thought it looked janky but then I went dark-theme-sicko-mode and realized it might look kinda nice recessed below the main toolbar. This makes that depth explicit in all themes. Up to you!
toolbarcontainer


awesomekling:
Hmm. Not a huge fan of the receded look. How about we fix the bug instead :)

Issue #5487, opened 2021-02-23
Enable more Warnings!


Domkeykong:
`-Wall -Wextra` are the bare minumun of warnings to enable
`-Wmissing-declarations -Wformat=2` are a good step in the right direction, but
gcc is known to have bad default warning options.
Here is an incomprehansive list of warnings to consider enabeling:
```cmake
-Wshadow # warn the user if a variable declaration shadows one from a parent context
-Wnon-virtual-dtor # warn the user if a class with virtual functions has a non-virtual destructor.
                   # This helps catch hard to track down memory errors
-Wold-style-cast # warn for c-style casts
-Wcast-align # warn for potential performance problem casts
-Wunused # warn on anything being unused
-Woverloaded-virtual # warn if you overload (not override) a virtual function
-Wpedantic # warn if non-standard C++ is used
-Wconversion # warn on type conversions that may lose data
-Wsign-conversion # warn on sign conversions
-Wnull-dereference # warn if a null dereference is detected
-Wdouble-promotion # warn if float is implicit promoted to double
-Wformat=2 # warn on security issues around functions that format output (ie printf)
-Wmisleading-indentation # warn if indentation implies blocks where blocks do not exist
-Wduplicated-cond # warn if if / else chain has duplicated conditions
-Wduplicated-branches # warn if if / else branches have duplicated code
-Wlogical-op # warn about logical operations being used where bitwise were probably wanted
-Wuseless-cast # warn if you perform a cast to the same type
```

Hendiadyoin1:
I just tested with all the flags you specified and in the Kernel accessed code there are like 220 warning, most of them are old-style casts (129) and unsafe casts of integers of diffrerent sizes and signs(53)...  
And a zero-sized array in StringImpl?  
And zero-sized arrays in Array and Vector

awesomekling:
Yes, compilers have many warnings. Do you have a specific plan to enable one of these warnings and deal with the fallout?

Hendiadyoin1:
I'll look into the oldstyle casts a bit and try to fix that 0 sized array in StringImpl

Nicholas-Baron:
I have started looking for the free warnings (ones which can be enabled without any new warnings occuring).

A few things to report:
1. `-Wdouble-precision` occurs a lot. Any call to `String::format` seems to cause double promotion.
[...]

PR #5524, opened 2021-02-25
Everywhere: Add unused and old-style-cast warnings to AK, Kernel, and Libraries

Hendiadyoin1:
this is work towards #5487

During whis I've found some other (related things) things:
* we were casting m_raw to u32 to just cast it back to u64
* somewhere we were using size_t as a ptr_t instead of FlatPtr
* why are we casting nulltptrs?
* why are we casting constants over two corners and not just saying the constant directly?
* someone else should take a look at IntrusiveList<T, member>::node_to_value its horrifying
* God help me are these a lot of casts

### DIFF
--- a/AK/IntrusiveList.h
+++ b/AK/IntrusiveList.h
@@ -273,7 +273,11 @@ inline typename IntrusiveList<T, member>::Iterator IntrusiveList<T, member>::end
 template<class T, IntrusiveListNode T::*member>
 inline T* IntrusiveList<T, member>::node_to_value(IntrusiveListNode& node)
 {
-    return (T*)((char*)&node - ((char*)&(((T*)nullptr)->*member) - (char*)nullptr));
+// FIXME: WTF is even going on here, why are we casting nullptrs?
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wold-style-cast"
+    return reinterpret_cast<T*>((char*)&node - ((char*)&(((T*)nullptr)->*member) - (char*)nullptr));
+#pragma GCC diagnostic pop
 }
 
 inline IntrusiveListNode::~IntrusiveListNode()
--- a/AK/Bitmap.h
+++ b/AK/Bitmap.h
@@ -126,19 +126,19 @@ public:
             byte &= bitmask_last_byte[(start + len) % 8];
             count += __builtin_popcount(byte);
             if (++first < last) {
-                const u32* ptr32 = (const u32*)(((FlatPtr)first + sizeof(u32) - 1) & ~(sizeof(u32) - 1));
-                if ((const u8*)ptr32 > last)
-                    ptr32 = (const u32*)last;
-                while (first < (const u8*)ptr32) {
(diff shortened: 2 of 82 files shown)


alimpfard:
> Everywhere: change c-castings to corresponding c++ casting

Not that I'm saying we shouldn't do this, but is there any real benefit to this? it feels rather silly to go to all this trouble (and further trouble) just to appease a warning that doesn't really do much in a project of this scope?

I've always found `reinterpret_cast<T>(expr)` to be seriously chunky for no reason `(T)expr` communicates the same thing in a C++ project anyway?

Domkeykong:
A C-style cast can perform any combination of the following casts:
- `const_cast`
- `static_cast`
- `reinterpret_cast`

By explicitly telling the compiler what we want a `static_cast` we are still `const` correct.
Doing a `reinterpret_cast` is still Undefined Behavior, but you already knew that ;)
Instead you should looking into doing a `std::bit_cast`

AK/IntrusiveList.h
```diff
@@ -273,7 +273,11 @@ inline typename IntrusiveList<T, member>::Iterator IntrusiveList<T, member>::end
 template<class T, IntrusiveListNode T::*member>
 inline T* IntrusiveList<T, member>::node_to_value(IntrusiveListNode& node)
 {
-    return (T*)((char*)&node - ((char*)&(((T*)nullptr)->*member) - (char*)nullptr));
+// FIXME: WTF is even going on here, why are we casting nullptrs?
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wold-style-cast"
+    return reinterpret_cast<T*>((char*)&node - ((char*)&(((T*)nullptr)->*member) - (char*)nullptr));
```
ADKaster, line 279: This is a hand rolled `offsetof`. I believe we can use `__builtin_offsetof` instead, to be more standards compliant.

ADKaster: Actually it's not that simple. I don't know how to implement this particular operation with the pointer-to-member using offsetof. But the comment should be more like `// arcane magic here, get pointer to T given the list node pointer-to-member in the CRTP'd type T.`

Nicholas-Baron:
`-Wunused` is enabled project wide due to #6342. 

Maybe `-Wuseless-cast` could lighten the work by picking out redundant casts.

Hendiadyoin1:
I think if I start work on thiis again I should drop the old-style-cast checks, because that's way to tedious and some casts are just weird...

alimpfard:
If you're talking about the manual resolution of indirect class member pointers, I need to fix them for a little thing I'm working on.
Though it'll probably just be riddled with `reinterpret_cast` anyway.

Hendiadyoin1:
In general, they are every where, and sometimes its logical what it is supposed to do and sometimes you are wondering, why we cast at that point (enums in switch statements casted to integers to handle a default case, was a case if I remember correctly)

And I started so far back, I would have to restart with that flag again...

PR #5526, opened 2021-02-25
Spreadsheet: Fix cell identifier label

GMTA:
Commit 6a6f19a broke the cell position display in the top left of the Spreadsheet window and the title of the cell type dialog, causing the application to crash when interacting with cells beyond column FE.


alimpfard (changes requested):
Typical me, always forgetting to test things...

Please properly format the commit message (as per https://github.com/SerenityOS/serenity/blob/master/CONTRIBUTING.md#code-submission-policy) - at the very least, titlecase "Fix".

I believe this identifier is generated and used in other places too, `CellTypeDialog` being one I can immediately think of, please un-break those too while you're at it :)

GMTA:
@alimpfard Good points; fixed them as well.

alimpfard (approved):

PR #5528, opened 2021-02-25
Meta: Enable qemu virtualization acceleration on mac

Safffatttu:
As Mac OS does not support kvm, QEMU virtualization is really slow.
To address this, I changed run script so that it check whether we can Mac virtualization acceleration.
This change drastically improves QEMU performance on Mac machines.

### DIFF
--- a/Meta/run.sh
+++ b/Meta/run.sh
@@ -10,9 +10,21 @@ die() {
 
 #SERENITY_PACKET_LOGGING_ARG="-object filter-dump,id=hue,netdev=breh,file=e1000.pcap"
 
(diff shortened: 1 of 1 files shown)


emanuele6:
Please, make sure to follow the [contributing guidelines](https://github.com/SerenityOS/serenity/blob/master/CONTRIBUTING.md).

Mainly:
* The first line of the commit message should have the format "Category: Brief description of what's being changed". The "category" can be a subdirectory, but also something like "POSIX compliance" or "ClassName". Whatever seems logical.

(In this case, the *category* would be "Meta")

Also, prefer imperative verbs in commit titles; i.e. "enable" instead of "enabled".

PR #5529, opened 2021-02-26
Kernel: Add AHCI functionality

supercomputer7:
This patch allows us to utilize AHCI, so we can boot and use SATA drives.
Also, some of this code was tested with a real AHCI controller (that was passed to the VM with IOMMU), I still need to check it again.

Still a work in progress, I have a few things to fix, but it's mostly finished now :)


Kernel/Arch/i386/CPU.h
```diff
@@ -295,6 +295,8 @@ void load_task_register(u16 selector);
                                : "memory")
 #define sti() asm volatile("sti" :: \
                                : "memory")
+#define memory_barrier() asm volatile("" :: \
+                                          : "memory")
```
alimpfard, line 299: The name sounds... misleading?
I was expecting `__builtin_ia32_mfence()` here tbh

Maybe name it `compiler_memory_barrier` instead?

Kernel/Storage/AHCIController.cpp
```diff
@@ -0,0 +1,204 @@
...
+    hba().control_regs.ghc = 1;
+    memory_barrier();
+    size_t retry = 0;
+    while (true) {
+        if (retry > 1000)
+            return false;
+        if (!(hba().control_regs.ghc & 1))
+            break;
+        // wait 1 second
+        IO::delay(1000);
```
alimpfard, line 52: This is 1ms, right?
The comment below seems more feasible than the one above ("wait 1s")

__/me wonders if some basic units-of-measure API would be morr helpful__

Kernel/Arch/i386/CPU.h
supercomputer7: Maybe I need to do this in a better way, and I should use the `__builtin_ia32_mfence()` with the `compiler_memory_barrier` macro somehow...

alimpfard: if I'm not mistaken, you don't _really_ need to stop the CPU from reordering this (there's no other nondependent memory ops around it anyway), you just want to stop the compiler from reordering it, so an `mfence` sounds pretty overkill - I think just naming the macro in a way that doesn't imply "this is an mfence" would be more than enough

supercomputer7: Well, I do also want to ensure that the CPU doesn't reorder on its own the access to memory, because the order of setting some of the registers in the HBA is very important.

Lubrsi:
Doesn't appear to work on VMware Player:
```
[34;1m[init_stage2(1:1)][0m: PCI [0000:02:03:00]: AHCI controller reset
[34;1m[#0 init_stage2(1:1)][0m: PCI [0000:02:03:00]: AHCI command list entries count - 32
[34;1m[init_stage2(1:1)][0m: AHCI Port 1: Device detected, Phy disabled
[34;1m[#0 init_stage2(1:1)][0m: AHCI Port 1: Identify failed, SError 67174400
[34;1m[#0 init_stage2(1:1)][0m: Spurious Interrupt, vector 7
[34;1m[init_stage2(1:1)][0m: AHCI Port 2: Device detected, Phy disabled
[34;1m[#0 init_stage2(1:1)][0m: AHCI Port 2: Identify failed, SError 67174400
```
But not really a blocker. It works with IDE.

Lubrsi:
Also confirmed that it works on VirtualBox :+1:

supercomputer7:
> Doesn't appear to work on VMware Player:
> 
> ```
> �[34;1m[init_stage2(1:1)]�[0m: PCI [0000:02:03:00]: AHCI controller reset
> �[34;1m[#0 init_stage2(1:1)]�[0m: PCI [0000:02:03:00]: AHCI command list entries count - 32
> �[34;1m[init_stage2(1:1)]�[0m: AHCI Port 1: Device detected, Phy disabled
> �[34;1m[#0 init_stage2(1:1)]�[0m: AHCI Port 1: Identify failed, SError 67174400
> �[34;1m[#0 init_stage2(1:1)]�[0m: Spurious Interrupt, vector 7
> �[34;1m[init_stage2(1:1)]�[0m: AHCI Port 2: Device detected, Phy disabled
> �[34;1m[#0 init_stage2(1:1)]�[0m: AHCI Port 2: Identify failed, SError 67174400
> ```
> 
> But not really a blocker. It works with IDE.

I checked what is error `67174400`, and it appears to be the "Exchanged (X)" bit in `SError`. I don't have a VMWare Player on my machine, but I bet it's something I have done wrong in the reset sequence.

Lubrsi:
Oh, it appears to only be exposing the IDE controller for some reason, even though I've told it to use SATA...

supercomputer7:
I currently work on improving the code, so it can deal with scattered DMA regions and reads and writes that have block count bigger than 8.
If I have time and it will be necessary, I'll try to test again on real hardware.

supercomputer7:
> Oh, it appears to only be exposing the IDE controller for some reason, even though I've told it to use SATA...

That explains why it says `Device detected, Phy disabled`, which normally shouldn't happen after a reset sequence if the device is attached.

Kernel/Arch/i386/CPU.h
```diff
@@ -296,6 +296,10 @@ void load_task_register(u16 selector);
 #define sti() asm volatile("sti" :: \
                                : "memory")
 
+#define full_memory_barrier() __sync_synchronize()
+#define compiler_memory_barrier() asm volatile("" :: \
```
tomuta, line 300: We should just add the memory fences that are standardized to AK::Atomic.h, which will emit the appropriate instructions.

supercomputer7: I don't think I care about atomic operations (should I need to?), I do care that the CPU will not reorder memory accesses because it can result in undefined behavior in some cases.

tomuta: It's the same thing. The memory fences aren't really anything that have anything to do with atomic operations. It's all about reordering.

tomuta: https://www.modernescpp.com/index.php/fences-as-memory-barriers

tomuta: A full barrier is simply `memory_order_acq_rel`

Kernel/Storage/AHCIPort.cpp
```diff
@@ -0,0 +1,674 @@
...
+    if (!(m_port_registers.cmd & (1 << 20)))
+        return;
+    m_port_registers.cmd = m_port_registers.cmd | (1 << 2);
+}
+
+void AHCIPort::spin_up() const
+{
+    VERIFY(m_lock.is_locked());
+    dbgln_if(AHCI_DEBUG, "AHCI Port {}, staggered spin up? {}", representative_port_index(), m_parent_handler->hba_capabilities().staggered_spin_up_supported);
+    if (m_parent_handler->hba_capabilities().staggered_spin_up_supported)
```
8infy, line 597: probably forgot to negate here

supercomputer7: Thanks for noticing :)

Kernel/Arch/i386/CPU.h
supercomputer7: OK, thanks for the explanation :)

supercomputer7: I fixed the code to use `atomic_thread_fence(AK::MemoryOrder::memory_order_acq_rel)`, I hope it's enough for ensuring memory order :)

supercomputer7:
> Doesn't appear to work on VMware Player:
> 
> ```
> �[34;1m[init_stage2(1:1)]�[0m: PCI [0000:02:03:00]: AHCI controller reset
> �[34;1m[#0 init_stage2(1:1)]�[0m: PCI [0000:02:03:00]: AHCI command list entries count - 32
> �[34;1m[init_stage2(1:1)]�[0m: AHCI Port 1: Device detected, Phy disabled
> �[34;1m[#0 init_stage2(1:1)]�[0m: AHCI Port 1: Identify failed, SError 67174400
> �[34;1m[#0 init_stage2(1:1)]�[0m: Spurious Interrupt, vector 7
> �[34;1m[init_stage2(1:1)]�[0m: AHCI Port 2: Device detected, Phy disabled
> �[34;1m[#0 init_stage2(1:1)]�[0m: AHCI Port 2: Identify failed, SError 67174400
> ```
> 
> But not really a blocker. It works with IDE.

Is it possible that VMWare Player supports staggered spin up feature? I added fix for such case, because we should spin-up the device before trying to set `PxSCTL.DET` to 1, if staggered spin up feature is supported.

Lubrsi:
> Is it possible that VMWare Player supports staggered spin up feature? I added fix for such case, because we should spin-up the device before trying to set `PxSCTL.DET` to 1, if staggered spin up feature is supported.

It doesn't appear to:
```
[34;1m[#0 init_stage2(1:1)][0m: AHCI Controller Capabilities: c1341f1d
[34;1m[init_stage2(1:1)][0m: PCI [0000:02:03:00]: AHCI controller reset
[34;1m[#0 init_stage2(1:1)][0m: PCI [0000:02:03:00]: AHCI command list entries count - 32
[34;1m[#0 init_stage2(1:1)][0m: AHCI Port 1, staggered spin up? false
[34;1m[init_stage2(1:1)][0m: AHCI Port 1: Device detected, Phy disabled
[34;1m[#0 init_stage2(1:1)][0m: AHCI Port 1, staggered spin up? false
[34;1m[#0 init_stage2(1:1)][0m: AHCI Port 1: unused command header at index 0
[34;1m[#0 init_stage2(1:1)][0m: AHCI Port Handler: Get identify metadata physical address of port 0 - 0
[34;1m[#0 init_stage2(1:1)][0m: AHCI Port 1: Identify failed, SError 67174400
[34;1m[#0 colonel(0:0)][0m: Spurious Interrupt, vector 7
[34;1m[#0 init_stage2(1:1)][0m: AHCI Port 2, staggered spin up? false
[34;1m[init_stage2(1:1)][0m: AHCI Port 2: Device detected, Phy disabled
[34;1m[#0 init_stage2(1:1)][0m: AHCI Port 2, staggered spin up? false
[34;1m[#0 init_stage2(1:1)][0m: AHCI Port 2: unused command header at index 0
[34;1m[#0 init_stage2(1:1)][0m: AHCI Port Handler: Get identify metadata physical address of port 1 - 0
[34;1m[#0 init_stage2(1:1)][0m: AHCI Port 2: Identify failed, SError 67174400
```

supercomputer7:
@awesomekling I looked at the code again, it seems to work fine. I also tested this with SMP and the system boots just fine.
If you have any notes on the changes, I'd be happy to read those and fix accordingly :)

Kernel/Arch/i386/CPU.h
tomuta: Actually, you probably also need to pair it with `atomic_signal_fence` to prevent compiler reordering.

supercomputer7: But we don't have such function in AK/Atomic.h, so I think we should add a new function for it and pair these functions together in one `full_memory_barrier` function in Atomic.h :)

Kernel/Arch/i386/CPU.h
tomuta: Hm didn't see this until now but generally for memory-mapped registers you just need to mark the fields in the structure that represents them as volatile. That ensures that the compiler generates code that reads/writes them in the specified order. But it doesn't imply a memory fence.

Kernel/Storage/AHCIPort.cpp
```diff
@@ -305,32 +318,63 @@ void AHCIPort::set_sleep_state() const
     m_port_registers.cmd = (m_port_registers.cmd & 0x0ffffff) | (0b1000 << 28);
 }
 
-void AHCIPort::start_request(AsyncBlockDeviceRequest& request)
+Optional<size_t> AHCIPort::calculate_descriptors_count(size_t block_count) const
```
awesomekling, line 321: This returns `Optional<size_t>` but never fails? In what situation should this fail?

Kernel/Storage/AHCIPort.h
```diff
@@ -83,6 +99,8 @@ class AHCIPort : public RefCounted<AHCIPort> {
     void start_request(AsyncBlockDeviceRequest&);
     void complete_current_request(AsyncDeviceRequest::RequestResult);
     bool access_device(AsyncBlockDeviceRequest::RequestType, u64 lba, u8 block_count);
+    Optional<size_t> calculate_descriptors_count(size_t block_count) const;
+    NO_DISCARD Optional<AsyncDeviceRequest::RequestResult> prepare_and_set_scatter_list(AsyncBlockDeviceRequest& request);
```
awesomekling, line 103: Please use `[[nodiscard]]`, `NO_DISCARD` is a hack that is only used with `class` declarations to prevent `clang-format` from messing up the file. :)

Kernel/Storage/AHCIPort.h
```diff
@@ -50,6 +51,21 @@ class AHCIPort : public RefCounted<AHCIPort> {
     friend class AHCIPortHandler;
     friend class SATADiskDevice;
 
+private:
+    class ScatterList : public RefCounted<ScatterList> {
+    public:
+        static NonnullRefPtr<ScatterList> create(AsyncBlockDeviceRequest&, NonnullRefPtrVector<PhysicalPage> allocated_pages, size_t device_block_size);
+        NonnullRefPtr<AnonymousVMObject> vmobject() const { return m_vm_object; }
```
awesomekling, line 58: Since the `VMObject` is owned by the `ScatterList`, let's return it as a `const VMObject&` or `VMObject&`.

Kernel/Storage/AHCIPort.cpp
supercomputer7: The idea was to not return anything with the `Optional` container if we don't have enough DMA pages for a request, but for now I prefer to simply assert on that, so for now we can return `size_t` :)

PR #5536, opened 2021-02-26
Net: use KBuffer::capacity() when checking if we can reuse buffer

adamse:


awesomekling:
Well that just straight-up makes sense.

PR #5538, opened 2021-02-26
LibRegex: Implement section B.1.4. of the ECMA262 spec / LibRegex: some tiny fixes

### LINKED ISSUE #5517: LibRegex: Parsing error in prettify.js
From: https://getbootstrap.com/2.0.0/assets/js/google-code-prettify/prettify.js
Error:
```
WebContent(30): HTMLScriptElement: Running script assets/js/google-code-prettify/prettify.js
WebContent(30): Throwing JavaScript Error: SyntaxError, RegExp compile error: Error during parsing of regular expression:
    ^(?:^^\.?|[!+-]|!=|!==|#|%|%=|&|&&|&&=|&=|\(|\*|\*=|\+=|,|-=|->|\/|\/=|:|::|;|<|<<|<<=|<=|=|==|===|>|>=|>>|>>=|>>>|>>>=|[?@[^]|\^=|\^\^|\^\^=|{|\||\|=|\|\||\|\|=|~|break|case|continue|delete|do|else|finally|instanceof|return|throw|try|typeof)\s*(\/(?=[^*/])(?:[^/[\\]|\\[\S\s]|\[(?:[^\\\]]|\\[\S\s])*(?:]|$))+\/)
                                                                                                                                                  ^---- Invalid regular expression.
```

### LINKED ISSUE #5518: LibRegex: Parsing error with unlimited quantifiers
For example, `([0-9a-f]{3,})` should match 3 _or more_ times.
The parser doesn't handle this:
```
WebContent(30): Throwing JavaScript Error: SyntaxError, RegExp compile error: Error during parsing of regular expression:
    (?:#|\^)([0-9a-f]{3,})\:(?:#|\^)([0-9a-f]{3,})
                        ^---- Content of {} invalid: not a number, number too large, more than two numbers, first larger than second.
```
(From https://getbootstrap.com/docs/3.3/assets/js/docs.min.js)

alimpfard:
This allows the parser to deal with crazy patterns like the one in #5517.

I decided to _conditionally_ bite the bullet (in the regex library).
As discussed, this enables that flag by default in LibJS (+a FIXME). we can later disable it as needed.

Fixes #5517.
Fixes #5518.

### DIFF
--- a/Userland/Libraries/LibRegex/RegexByteCode.cpp
+++ b/Userland/Libraries/LibRegex/RegexByteCode.cpp
@@ -701,21 +701,35 @@ const Vector<String> OpCode_Compare::variable_arguments_to_string(Optional<Match
 
     for (size_t i = 0; i < arguments_count(); ++i) {
         auto compare_type = (CharacterCompareType)m_bytecode->at(offset++);
-        result.empend(String::format("type=%lu [%s]", (size_t)compare_type, character_compare_type_name(compare_type)));
+        result.empend(String::formatted("type={} [{}]", (size_t)compare_type, character_compare_type_name(compare_type)));
 
         auto compared_against_string_start_offset = state().string_position > 0 ? state().string_position - 1 : state().string_position;
 
         if (compare_type == CharacterCompareType::Char) {
-            char ch = m_bytecode->at(offset++);
-            result.empend(String::format("value='%c'", ch));
-            if (!view.is_null() && view.length() > state().string_position)
-                result.empend(String::format(
-                    "compare against: '%s'",
-                    view.substring_view(compared_against_string_start_offset, state().string_position > view.length() ? 0 : 1).to_string().characters()));
+            auto ch = m_bytecode->at(offset++);
+            auto is_ascii = isascii(ch) && isprint(ch);
+            if (is_ascii)
+                result.empend(String::formatted("value='{:c}'", static_cast<char>(ch)));
+            else
+                result.empend(String::formatted("value={:x}", ch));
+
+            if (!view.is_null() && view.length() > state().string_position) {
+                if (is_ascii) {
+                    result.empend(String::formatted(
+                        "compare against: '{}'",
+                        view.substring_view(compared_against_string_start_offset, state().string_position > view.length() ? 0 : 1).to_string()));
+                } else {
+                    auto str = view.substring_view(compared_against_string_start_offset, state().string_position > view.length() ? 0 : 1).to_string();
+                    u8 buf[8] { 0 };
+                    __builtin_memcpy(buf, str.characters(), min(str.length(), sizeof(buf)));
+                    result.empend(String::formatted("compare against: {:x},{:x},{:x},{:x},{:x},{:x},{:x},{:x}",
+                        buf[0], buf[1], buf[2], buf[3], buf[4], buf[5], buf[6], buf[7]));
+                }
+            }
         } else if (compare_type == CharacterCompareType::NamedReference) {
             auto ptr = (const char*)m_bytecode->at(offset++);
             auto length = m_bytecode->at(offset++);
-            result.empend(String::format("name='%.*s'", (int)length, ptr));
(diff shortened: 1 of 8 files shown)


alimpfard:
cc @linusg for JS stuff

Userland/Libraries/LibRegex/RegexParser.h
```diff
@@ -197,6 +201,17 @@ class ECMA262Parser final : public Parser {
...
+    // Used only by B.1.4, Regular Expression Patterns (Extended for use in browsers)
+    bool parse_quantifiable_assertion(ByteCode&, size_t&, bool named);
+    bool parse_extended_atom(ByteCode&, size_t&, bool named);
+    bool parse_inner_disjunction(ByteCode& bytecode_stack, size_t& length, bool unicode, bool named);
+    bool parse_invalid_braced_quantifier(); // Note: This function either parses and *fails*, or doesn't parse anything and returns false.
+    bool parse_legacy_octal_escape_sequence(ByteCode& bytecode_stack, size_t& length);
+    Optional<u8> parse_legacy_octal_escape();
+
+    // Let's keep them separate, for now?
+    bool m_should_use_browser_extended_grammar { false };
```
linusg, line 214: Yes, definitely. What I'd like to do, and I might have mentioned this on IRC before, is adding an Annex B flag to the JS parser and engine, so you can conditionally enable/disable all the legacy stuff - I'd say for now just always enable this and I'll come back to it, I'm sure there are other Annex B features already built in unconditionally.

The browser would have that always on, and the JS repl by default with an option to disable, or something like that.

linusg (commented):
great stuff!

Userland/Libraries/LibRegex/RegexParser.h
alimpfard: Then I guess we can always pass the `Internal_BrowserExtended` flag from LibJS here, and any further changes can be done in LibJS.

Userland/Libraries/LibRegex/RegexByteCode.cpp
```diff
@@ -701,21 +701,41 @@ const Vector<String> OpCode_Compare::variable_arguments_to_string(Optional<Match
...
+            if (!view.is_null() && view.length() > state().string_position) {
+                if (is_ascii) {
+                    result.empend(String::formatted(
+                        "compare against: '{}'",
+                        view.substring_view(compared_against_string_start_offset, state().string_position > view.length() ? 0 : 1).to_string()));
+                } else {
+                    auto str = view.substring_view(compared_against_string_start_offset, state().string_position > view.length() ? 0 : 1).to_string();
+                    u64 value = 0;
+                    if (str.length() >= sizeof(u64))
+                        value = *reinterpret_cast<const u64*>(str.characters());
```
tomuta, line 725: Isn't this going to cause potentially unaligned access? Shouldn't this be done using `memcpy`?

alimpfard: Honestly, it's in debug code...
I was thinking this as I wrote it, then I thought "well it's debug code, so who cares if it blows up"...

I'll fix it when I feel less lazy, draft until then I guess

alimpfard: Hmmm, actually, why even bother, let's just dump out some hex instead.

tomuta: I didn't even realize it was debug code. Anyway, on x86 this works with a slight speed penalty, though we should probably enable the alignment check at some point, which would then also lead to a crash. But it's not portable and not all architectures allow unaligned access. Older ARMs for example didn't allow it.

alimpfard: if we're debugging LibRegex on a non-x86 arch, I'll consider even the crash a win lol

tomuta: Well wait until @awesomekling throws in the alignment check bit and it'll crash on x86 as well ;)

PR #5540, opened 2021-02-26
LibGfx: Fix read buffer overflow in interlaced GIF decode

Lubrsi:
Unfortunately 10420dee7e48c818a7b1c5386b8fcebc587825f0 didn't quite fix it,
as the buffer overflow was actually happening here:
https://github.com/SerenityOS/serenity/blob/af2220448834fb0bff5132bf68104719819862ce/Userland/Libraries/LibGfx/GIFLoader.cpp#L402

Found by oss-fuzz: https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=30507


awesomekling:
Well duh :)

PR #5547, opened 2021-02-27
LibWeb: Display filename:line:column in JavaScript call traces

boricj:
This makes JavaScript errors a bit more useful in the logs:
```
WebContent(31): Throwing JavaScript Error: TypeError, undefined is not a function (evaluated from '<object>.replace')
WebContent(31):   <anonymous> at jquery-3.5.1.js:328:34
WebContent(31):   <anonymous> at jquery-3.5.1.js:36:12
WebContent(31):   (global execution context) at jquery-3.5.1.js:40:53
WebContent(31): Throwing JavaScript Error: ReferenceError, '$' is not defined
WebContent(31):   (global execution context) at (document):2:9
```
Note that this is a rather bare-bones implementation. Of particular note, forwarding the line and column numbers from the document to the JavaScript interpreter for inline script tags is not implemented, so those are displayed relative to the inline script contents.

### DIFF
--- a/Userland/Libraries/LibJS/Interpreter.cpp
+++ b/Userland/Libraries/LibJS/Interpreter.cpp
@@ -62,6 +62,7 @@ Value Interpreter::run(GlobalObject& global_object, const Program& program)
     VM::InterpreterExecutionScope scope(*this);
 
     CallFrame global_call_frame;
+    global_call_frame.current_node = &program;
     global_call_frame.this_value = &global_object;
     static FlyString global_execution_context_name = "(global execution context)";
     global_call_frame.function_name = global_execution_context_name;
@@ -140,6 +141,7 @@ void Interpreter::exit_scope(const ScopeNode& scope_node)
 
 void Interpreter::enter_node(const ASTNode& node)
 {
+    vm().call_frame().current_node = &node;
     vm().push_ast_node(node);
 }
 
--- a/Userland/Libraries/LibJS/Lexer.h
+++ b/Userland/Libraries/LibJS/Lexer.h
@@ -36,11 +36,12 @@ namespace JS {
 
 class Lexer {
 public:
-    explicit Lexer(StringView source);
+    explicit Lexer(StringView source, StringView filename = "(unknown)", size_t line_number = 1, size_t line_column = 0);
 
     Token next();
 
     const StringView& source() const { return m_source; };
+    const StringView& filename() const { return m_filename; };
 
 private:
     void consume();
@@ -65,6 +66,8 @@ private:
     size_t m_position { 0 };
     Token m_current_token;
     char m_current_char { 0 };
+
+    StringView m_filename;
     size_t m_line_number { 1 };
     size_t m_line_column { 0 };
 
--- a/Userland/Libraries/LibJS/AST.cpp
+++ b/Userland/Libraries/LibJS/AST.cpp
@@ -225,6 +225,7 @@ Value CallExpression::execute(Interpreter& interpreter, GlobalObject& global_obj
         }
     }
 
+    vm.call_frame().current_node = vm.node_stack().last();
     Object* new_object = nullptr;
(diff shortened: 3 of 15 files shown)


Userland/Libraries/LibJS/Lexer.h
```diff
@@ -62,11 +63,12 @@ class Lexer {
...
 
     StringView m_source;
+    StringView m_filename;
     size_t m_position { 0 };
-    Token m_current_token;
     char m_current_char { 0 };
-    size_t m_line_number { 1 };
-    size_t m_line_column { 0 };
+    size_t m_line_number;
+    size_t m_line_column;
```
alimpfard, line 70: \<kling\> Please initialize primitive types \</kling\>

Userland/Libraries/LibJS/Interpreter.cpp
```diff
@@ -140,6 +140,7 @@ void Interpreter::exit_scope(const ScopeNode& scope_node)
 
 void Interpreter::enter_node(const ASTNode& node)
 {
+    vm().call_frame().source_range = node.source_range();
```
linusg, line 143: why do we need to store the current source range on the call frame - we already store the node on the VM, can't we get it from there?
Also, the name is probably not ideal, since this isn't the "call frame's source range" but merely the current AST node's source range stored on the current call frame.

boricj: I'm not familiar with LibJS internals, but the call stack and the node stack are two independent data structures. It's not obvious to me that an AST node is guaranteed to outlive a particular call frame unless it takes an `AK::RefPtr` to it, which would be very expensive here right inside the interpreter loop...

I'd be happy to stick a raw pointer to an AST node here instead, but only if I'm explicitly told it's OK to do that.

linusg: > It's not obvious to me that an AST node is guaranteed to outlive a particular call frame

They will, they're all children of `JS::Program` which outlives execution of itself.

What I meant is, we already push/pop AST nodes as they are executed, and you can reach the current one via `vm().node_stack().last()`, but I now realise why that on its own with no link to individual call frames is insufficient.

FWIW it's a `Vector<const ASTNode*>` though so yes, that's ok. This is really cool btw :)

linusg:
Minor flaw, since function argument AST nodes are executed after the CallExpression node, something like this:

```js
function foo() {
    throw Error("foo() did throw!");
}

foo(
    1,
    2,
    3 
)
```

Produces the following error log:

```
Throwing JavaScript Error: Error, foo() did throw!
  foo at (unknown):2:11
  (global execution context) at (unknown):8:5
```

where `8:5` is `3`.

boricj:
I assume there are other nodes that might exhibit this behavior. I'm not a JavaScript programmer, so while I can attempt to fix these types of issues that are pointed out whack-a-mole style, I most likely can't fix those proactively... This is probably going to be one of these "meh, close enough" PRs that should be improved upon later on.

Userland/Libraries/LibJS/Interpreter.cpp
```diff
@@ -140,6 +141,7 @@ void Interpreter::exit_scope(const ScopeNode& scope_node)
 
 void Interpreter::enter_node(const ASTNode& node)
 {
+    vm().call_frame().current_node = &node;
```
linusg, line 144: This seems redundant now that `current_node` is set right after each `CallFrame call_frame;`

boricj: We need to update the current call frame at each node to track execution progress within a frame, not just on each new call frame.

boricj:
I've fixed the function call flaw and similar code sites, hopefully that's enough to have correct line and column numbers in all cases (or close enough anyway).

awesomekling:
Very neat feature :)

PR #5551, opened 2021-02-27
LibGUI: Faster spanned text

Torbenx:
Drawing text with spans was previously very inefficient.
To see the full difference disable kvm, open up a large c++ File in HackStudio and drag the scroll bar around.


awesomekling:
It's nice to make things faster but please explain the algorithm a little bit more than "it's faster" :)

Torbenx:
Previously this was going character by character and for each one it went through all possible spans and checked whether it was contained in its range.
This instead uses that the spans provided are sorted and goes though them in order and draws all the contained text at once.
That is the idea at least. It got a little longer then expected because of all the edge cases and spans that are longer then one line.
And I really wanted to at least print a warning when ignoring a span so this doesn't cause any confusion later.
Also: Should this go in a commit message?

awesomekling:
Right, yes, please put this in a commit message!

PR #5553, opened 2021-02-28
Userland: Add a simple mktemp(1) utility

IdanHo:
This was needed for a port

### DIFF
--- a/Userland/Utilities/mktemp.cpp
+++ b/Userland/Utilities/mktemp.cpp
@@ -1,0 +1,137 @@
+/*
+ * Copyright (c) 2021, Idan Horowitz <idan.horowitz@gmail.com>
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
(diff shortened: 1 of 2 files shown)


Userland/Utilities/mktemp.cpp
```diff
@@ -0,0 +1,136 @@
...
+    for (auto i = pattern_length - 1; i >= 0; --i) {
+        if (pattern[i] != 'X')
+            break;
+        new_filename[i] = random_characters[(arc4random() % (sizeof(random_characters) - 1))];
+    }
+
+    return new_filename;
+}
+
+static const char* make_temp(const char* pattern, bool directory, bool dry_run)
```
ADKaster, line 51: Is there anything special about the `mktemp(1)` utility that means you can't use `mktemp(3)` and `mkdtemp` from stdlib.h? I guess maybe the dry run option?

Userland/Utilities/mktemp.cpp
```diff
@@ -0,0 +1,136 @@
...
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <sys/stat.h>
+
+constexpr const char* default_template = "tmp.XXXXXXXXXX";
+
+static const char* generate_random_filename(const char* pattern)
+{
+    char* new_filename = strdup(pattern);
```
ADKaster, line 38: This string is never freed

Userland/Utilities/mktemp.cpp
bcoles: > 
> 
> Is there anything special about the `mktemp(1)` utility that means you can't use `mktemp(3)` and `mkdtemp` from stdlib.h? I guess maybe the dry run option?

`mktemp` can be used to perform a dry run. It does not create the associated file.

Userland/Utilities/mktemp.cpp
```diff
@@ -0,0 +1,136 @@
...
+#include <AK/LexicalPath.h>
+#include <LibCore/ArgsParser.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <sys/stat.h>
+
+constexpr const char* default_template = "tmp.XXXXXXXXXX";
+
+static const char* generate_random_filename(const char* pattern)
```
bcoles, line 36: `stdlib` also includes `generate_unique_filename` for this purpose.

bgianfo:
Consider adding a man page for your nice new shiny utility! 👍

Userland/Utilities/mktemp.cpp
IdanHo: mktemp(3) only allows templates with exactly 6 Xs (6 random characters), while mktemp(1) is specified to allow any amount larger than 3, but yes the dry run option as well

Userland/Utilities/mktemp.cpp
IdanHo: same problem as using mktemp(3) directly, mktemp(3) only allows templates with exactly 6 Xs (6 random characters), and as a result the implementation of generate_unique_filename in stdlib does as well, while mktemp(1) is specified to allow any amount larger than 3

Userland/Utilities/mktemp.cpp
IdanHo: oops! fixed

PR #5554, opened 2021-02-28
LibWeb: Handle scrolling an OutOfProcessWebView

### LINKED ISSUE #5515: Out-of-process web view doesn't scroll using mouse wheel
Steps to reproduce:
1. Open e.g. Help
2. Select a page that is big enough to show a vertical scroll bar
3. Move the mouse cursor to the web view, but not above a scroll bar
4. Try to scroll using the mouse wheel. It won't scroll as expected.

angus-g:
Closes #5515.

### DIFF
--- a/Userland/Libraries/LibWeb/OutOfProcessWebView.cpp
+++ b/Userland/Libraries/LibWeb/OutOfProcessWebView.cpp
@@ -245,6 +245,11 @@ void OutOfProcessWebView::notify_server_did_change_title(Badge<WebContentClient>
         on_title_change(title);
 }
 
+void OutOfProcessWebView::notify_server_did_request_scroll(Badge<WebContentClient>, int wheel_delta)
+{
+    vertical_scrollbar().set_value(vertical_scrollbar().value() + wheel_delta * 20);
+}
+
 void OutOfProcessWebView::notify_server_did_request_scroll_into_view(Badge<WebContentClient>, const Gfx::IntRect& rect)
 {
     scroll_into_view(rect, true, true);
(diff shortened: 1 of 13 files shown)


tomuta:
Hm somehow this doesn't feel like the right thing to do. Shouldn't that be handled on the receiving side only?

angus-g:
Yep, you're definitely right, I misunderstood the architecture a bit! Maybe it could be done by doing the normal client-side handling (see if there's a node that'll scroll), but if nothing responds (i.e. we're scrolling the main page), that should be sent back up to the `ScrollableWidget` with the usual `notify_server_should_...` pattern? I did at least get scrolling doing things this way, but I hacked it in front of the client-side test because currently that doesn't report whether the scroll event was actually handled...

Userland/Libraries/LibWeb/OutOfProcessWebView.cpp
```diff
@@ -245,6 +245,11 @@ void OutOfProcessWebView::notify_server_did_change_title(Badge<WebContentClient>
         on_title_change(title);
 }
 
+void OutOfProcessWebView::notify_server_did_request_scroll(Badge<WebContentClient>, int wheel_delta)
+{
+    vertical_scrollbar().set_value(vertical_scrollbar().value() + wheel_delta * 20);
```
angus-g, line 250: This is a bit yucky... just copied from `ScrollableWidget::handle_wheel_event`, but maybe this should be factored out (and that way the multiplication constant can be more sensibly defined)?

Also, is it sufficient to just give a wheel delta? I guess we don't have horizontal/vertical wheel scrolling to differentiate right now?

PR #5555, opened 2021-02-28
Documentation: Added GTK3-Devel in build instructions

r0x0d:
The dependency `gtk3-devel` was missing in the Fedora build instruction

### DIFF
--- a/Documentation/BuildInstructions.md
+++ b/Documentation/BuildInstructions.md
@@ -12,7 +12,7 @@ sudo apt install build-essential cmake curl libmpfr-dev libmpc-dev libgmp-dev e2
 
 **Fedora**
 ```bash
-sudo dnf install curl cmake mpfr-devel libmpc-devel gmp-devel e2fsprogs ninja-build patch @"C Development Tools and Libraries" @Virtualization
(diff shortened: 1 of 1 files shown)


deoxxa:
What exactly is GTK needed for?

awesomekling:
Please format your commit messages according to the [contribution guidelines](https://github.com/SerenityOS/serenity/blob/2830ce538360f12ee5fe8f7a922c2c2b7d5295a7/CONTRIBUTING.md)

r0x0d:
> What exactly is GTK needed for?

Hi, @deoxxa. I'm new to the project and not entirely sure in what manner the `gtk3-devel` is needed, but I was trying to build the project using the shell script `Toolchains/BuildQemu.sh` and suddenly it start to fails because of this dep. Thought it was something that was missing from the docs.

linusg:
Would probably good to mention that explicitly then as people using their distro's QEMU won't need it :)

r0x0d:
Should have open an issue first to talk about this, not an PR. Sorry. Should i close this PR?

ADKaster:
For what it's worth, `libgtk-3-dev` is required to run the BuildQemu script on Ubuntu. The only reference in BuildInstructions.md to that script is in the macOS section. So maybe there should be a more detailed explanation of when and why you'd want to build your own QEMU binaries? (And what extra packages might be required).

PR #5559, opened 2021-02-28
Spreadsheet: Implement the cut operation for cells

### LINKED ISSUE #4275: Spreadsheet: cut
Spreadsheet has copy and paste but no cut.

IdanHo:
This is done by adding another field to our custom text/x-spreadsheet-data mime-type that specifies the action (just copy/cut for now).

This closes #4275


IdanHo:
About macos tests failing: looks like gnu's mirror is down and as such the workflow cant download its dependencies

PR #5562, opened 2021-02-28
Meta: Allow specifying custom grub config file for build-image-grub.sh

tomuta:
This allows passing a custom grub config file as second argument to
build-image-grub.sh.

_Since grub is only displayed for 1 second I find myself missing that time frame often, and booting without serial debug. This way I can simply specify my own custom grub config._


linusg:
> Since grub is only displayed for 1 second I find myself missing that time frame often

Same - do we control that 1s and can make it wait a little longer?

tomuta:
Yeah but it's also convenient to be able to just point to your custom grub config. E.g. I like booting with smp=on, but if you're working on other flag-driven features it also helps without having to modify the files (which can easily end up in a commit by accident)

Issue #5539, opened 2021-02-26
Kernel: Crash with pure virtual call if the APIC timer fails to calibrate


Lubrsi:
When booting in VMware Player with SMP enabled, the APIC timer will fail to calibrate and will crash with a pure virtual call:
```
[36;1m[Kernel][0m: APICTimer: Using HPET as calibration source
[36;1m[Kernel][0m: KernelRng: Using RDSEED or RDRAND as entropy source
[36;1m[Kernel][0m: APICTimer: Frequency too slow!
[34;1m[Kernel][0m: ASSERTION FAILED: false
[34;1m[Kernel][0m: ../Kernel/StdLib.cpp:381 in void __cxa_pure_virtual()
[34;1m[#0 Kernel][0m: 0xc01237db (next: 0xc0107f38)
[34;1m[#0 Kernel][0m:  (next: 0xc0107f58)
[34;1m[#0 Kernel][0m: 0xc0125321 (next: 0xc0107fb8)
[34;1m[#0 Kernel][0m: 0xc02a9adc (next: 0xc0107ff8)
[34;1m[#0 Kernel][0m: 0xc046d622 (next: 0xc0108038)
[34;1m[#0 Kernel][0m: 0xc046d7fe (next: 0xc0108058)
[34;1m[#0 Kernel][0m: 0xc05ba11e (next: 0xc01080a8)
[34;1m[#0 Kernel][0m: 0xc05a15c6 (next: 0xc01080d8)
[34;1m[#0 Kernel][0m: 0xc05c0b54 (next: 0xc01081a8)
[34;1m[#0 Kernel][0m: 0xc05ca99c (next: 0xc0108298)
[34;1m[#0 Kernel][0m: 0xc0100145 (next: 0x00000000)
```
This points to:
https://github.com/SerenityOS/serenity/blob/af2220448834fb0bff5132bf68104719819862ce/Kernel/Time/APICTimer.cpp#L43

tomuta:
Hm seems odd, not sure how it could possibly be a pure virtual?

At any rate, this seems questionable:
https://github.com/SerenityOS/serenity/blob/af2220448834fb0bff5132bf68104719819862ce/Kernel/Time/APICTimer.cpp#L118

Lubrsi:
If it helps at all, addr2line does specify `(discriminator 1)`: `./Build/../Kernel/Time/APICTimer.cpp:43 (discriminator 1)`

tomuta:
[...]

PR #5563, opened 2021-02-28
Kernel: Fix APIC timer calibration to be more accurate

tomuta:
We were calibrating it to 260 instead of 250 ticks per second (being
off by one for the 1/10th second calibration time), resulting in
ticks of only ~3.6 ms instead of ~4ms. This gets us closer to ~4ms,
but because the APIC isn't nearly as precise as e.g. HPET, it will
only be a best effort. Then, use the higher precision reference
timer to more accurately calculate how many ticks we actually get
each second.

Also the frequency calculation was off, causing a "Frequency too slow"
error with VMware.

Fixes some problems observed in #5539

_@Lubrsi Looks like this also fixes the "Frequency too slow" problem on VMware_


Lubrsi:
Can confirm :+1:

PR #5564, opened 2021-02-28
FileManager+Everywhere: Add starting filemanager with focus on file

### LINKED ISSUE #4647: FileManager: implement ScrollToSelection()
So clicking links in Properties dialog could be useful.

speles:
Closes  #4647


awesomekling:
>Using to_string, and then constructing the URL from that string
>not always produces completely equal URL object.

We should fix that bug, instead of working around it in the IPC layer :)

speles:
Yes, you're probably correct :)
I've removed that commit from this PR, as it's not strictly necessary. And will open separate one for the URL fix as it's not that trivial.

PR #5577, opened 2021-02-28
LibTest + test-js: Add initial skeleton of LibTest and migrate code there

bgianfo:
The `test-js` reporter is arguably the nicest test runner / reporter that
exists in the serenity code base. To the goal of leveling up all the
other unit test environments, start a new `LibTest` library so that we
can share code and reporting utilities to make all the test systems
look and behave similarly.

The goal of this PR obviously isn't a huge code movement, it's more 
about a first babystep towards the goal of having a common test framework
(see: https://github.com/SerenityOS/serenity/issues/4078) if people like the proposed approach we can move forward migrating stuff
here and establishing common patterns for unit tests / crashing tests / benchmarks / etc, etc. 

cc: @ADKaster / @BenWiederhake  / @linusg


linusg (commented):
Awesome! Would you mind applying the same changes to test-web? It's largely a copy of test-js (another reason to factor out this stuff), so it'd be nice to keep them in sync. :)

bgianfo:
> Would you mind applying the same changes to test-web? It's largely a copy of test-js (another reason to factor out this stuff), so it'd be nice to keep them in sync. :)

Sure, done.

PR #5582, opened 2021-03-01
TextEditor: Add a command line option to jump to a specific line number

ry755:
This adds a command line option (-l or --line-number) to specify a line number to start at when opening a file in TextEditor.


diversys:
Maybe it's better to have `te` to open line numbers directly like `TextEditor ReadMe.md:10`?

awesomekling:
> Maybe it's better to have `te` to open line numbers directly like `TextEditor ReadMe.md:10`?

Let's support both. :)

ry755:
Good idea! I can try implementing that when I have some free time later today.

PR #5593, opened 2021-03-01
Line number integration through URL query

ElDonad:
Hi there ! 
This small PR aims to allow files to be open from the terminal at a specific line number. I tried to do it by making use of the URL query functionnality, since it is not usable in a traditionnal file path anyway. That's probably a bad idea, but I wanted to give it a try. 

I also implemented the integration for the small `bt` utility, since it's what motivated me to do it in the first place. 

(_I'd like to take this opportunity to thank you guys, and @awesomekling for your brillant work on this operating system. I'm a long time follower and enjoyer of the youtube coding sessions, and always wanted to add my little brick of my own, within the boundaries of my capacities...)_

### DIFF
--- a/Userland/Services/LaunchServer/Launcher.cpp
+++ b/Userland/Services/LaunchServer/Launcher.cpp
@@ -176,7 +176,7 @@ bool Launcher::open_url(const URL& url, const String& handler_name)
     if (url.protocol() == "file")
         return open_file_url(url);
 
-    return open_with_user_preferences(m_protocol_handlers, url.protocol(), url.to_string());
+    return open_with_user_preferences(m_protocol_handlers, url.protocol(), { url.to_string() });
 }
 
 bool Launcher::open_with_handler_name(const URL& url, const String& handler_name)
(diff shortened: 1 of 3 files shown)


ElDonad:
Also, food for tough, if this proof of concept doesn't seems unreasonable : 

- The URL parsing probably should be deferred to the software itself. This way we could scale the concept more easily to other file formats (maybe a page_number for a word document, or a function_name for a js file ? dunno)
- My main concern is that this would be in essence a parallel system to the traditionnal launch arguments, albeit associated with a file type instead of a specific application. Maybe a bad idea in itself. Or we could come up with some kind of mapping between url query parameters and CLI parameters...

Userland/Services/LaunchServer/Launcher.cpp
```diff
@@ -312,6 +312,19 @@ bool Launcher::open_file_url(const URL& url)
...
         extension = extension_parts.last();
-    return open_with_user_preferences(m_file_handlers, extension, url.path(), "/bin/TextEditor");
+
+    // Additionnal parameters parsing, specific for the file protocol
+    Vector<String> additional_parameters;
+    additional_parameters.append(url.path());
+    auto parameters = url.query().split('&');
+    for (auto parameter = parameters.begin(); parameter != parameters.end(); ++parameter) {
+        auto pair = parameter->split('=');
+        if (pair[0] == "line_number") {
```
linusg, line 322:
```suggestion
        if (pair.size() == 2 && pair[0] == "line_number") {
```

otherwise this would crash with something like "?line_number="

Userland/Services/LaunchServer/Launcher.cpp
```diff
@@ -312,6 +312,19 @@ bool Launcher::open_file_url(const URL& url)
...
+            auto line = pair[1].to_int();
+            if (line.has_value())
+                additional_parameters.prepend(String::format("-l%i", *line));
```
linusg, line 325: Please use `String::formatted()` in new code, also I think `value()` is more clear than `operator*()` here

```suggestion
                additional_parameters.prepend(String::formatted("-l {}", line.value()));
```

linusg (changes requested):

ElDonad:
Thanks for the review ! Sorry for the long wait, I didn't see notice it happened until this morning (still quite new at github collaboration). So, namespace mention dropped, const references where possible, and explicit function names over overloaded operators. Will try to remember that for the next time !

linusg (approved):

PR #5594, opened 2021-03-01
Shell+HackStudio: Generalise some LanguageServer functionality and implement new HS features in ShellLanguageServer

alimpfard:
In this PR:

- Move out `AutoCompleteEngine` and `FileDB` for reuse
- Fix busted Shell syntax highlighter
- Handle some edge cases in the spanned text drawing code
- Implement go-to-definition and listing of available functions in ShellLanguageServer

Some more bugs to fix:

- [x] Autocomplete inserts the entire word instead of the right portion
- [x] Strings are highlighted weirdly


itamar8910:
Looks good, definitely nice to have these components generalized out of the cpp langserver :+1:

PR #5595, opened 2021-03-01
FileManager+Everywhere: Improvements for starting FileManager with pre-selected files

### LINKED ISSUE #5586: FileManager: clicking Link to target doesn't select target
To repro:

* Right click Browser link on the desktop
* Select Properties
* Click Link target

FileManager will open /bin but won't select Browser binary.

speles:
This PR adds opening the folder in FileManager with a pre-selected file into more places (closes #5586), and a bit of related refactoring and fixing.

I've added that to all the places that I could've to think of, but the list probably is still incomplete.
So if you know more places where it makes sense, feel free to mention them :^)

Also, the LibGUI commit feels very hackish to me, but I'm struggling to come up with a better solution.
Maybe anyone has some ideas of a better fix?

### DIFF
--- a/AK/URL.cpp
+++ b/AK/URL.cpp
@@ -444,11 +444,12 @@ u16 URL::default_port_for_protocol(const String& protocol)
     return 0;
 }
 
-URL URL::create_with_file_protocol(const String& path)
+URL URL::create_with_file_protocol(const String& path, const String& fragment)
 {
     URL url;
     url.set_protocol("file");
     url.set_path(path);
+    url.set_fragment(fragment);
     return url;
 }
 
--- a/AK/URL.h
+++ b/AK/URL.h
@@ -77,7 +77,7 @@ public:
     const String& data_payload() const { return m_data_payload; }
 
     static URL create_with_url_or_path(const String& url_or_path);
-    static URL create_with_file_protocol(const String& path);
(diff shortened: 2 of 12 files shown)


linusg:
> Now that we use fragment for specifying starting selection in FileManager

Why was it implemented like that? Feels like a hack IMO, and quite unintuitive too. A regular path should be just fine - `FileManager /foo/bar.txt` would confirm that `bar.txt` is a file and then open `/foo` with `bar.txt` highlighted...

speles:
Yes, the fragment is only used for communicating with LaunchServer to tell it that we want to open the folder instead of the file :)

About the arguments: I agree that having 2 arguments might be not the greatest solution, but it allows us to support highlighting not only the "files inside a folder" but the "folder inside a folder" too (for example Location of a folder from Properties).
For example, should calling `FileManager /home/anon` open a `/home` with `anon` selected, or actually open `/home/anon`? :)

Though, we certainly can try to support the invocation of `FileManager /path/to/file` to be equal to `FileManager /path/to file`, but I'm not sure if it will make the FileManager usage less or more confusing.

speles:
Just noticed a great example @linusg left in #5597, and decided that that's indeed a much better way to handle this whole thing.
Added a similar implementation, so now this PR also closes #5597 :^)

PR #5596, opened 2021-03-01
Kernel: Enable AVX flags if supported by cpu

ashnewmanjones:
This enables AVX and its dependency CR4.OSXSAVE when both are available on the cpu.

I have tested this with and without AVX cpu flags in QEMU and it appears to be working (boots to OS), however I haven't figured out how to dump the xcr0 register from userspace to double check that the flags are indeed set correctly. If someone can point me in the right direction I'm happy to test that also.

### DIFF
--- a/Kernel/Arch/i386/CPU.cpp
+++ b/Kernel/Arch/i386/CPU.cpp
@@ -829,6 +829,24 @@ FlatPtr read_dr6()
     return dr6;
 }
 
+#define XCR_XFEATURE_ENABLED_MASK 0
+
+UNMAP_AFTER_INIT u64 read_xcr0()
+{
+    u32 eax, edx;
+    asm volatile("xgetbv"
+                 : "=a"(eax), "=d"(edx)
+                 : "c"(XCR_XFEATURE_ENABLED_MASK));
+    return eax + ((u64)edx << 32);
+}
+
+UNMAP_AFTER_INIT void write_xcr0(u64 value)
+{
+    u32 eax = value;
+    u32 edx = value >> 32;
+    asm volatile("xsetbv" ::"a"(eax), "d"(edx), "c"(XCR_XFEATURE_ENABLED_MASK));
+}
+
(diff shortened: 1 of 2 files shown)


Kernel/Arch/i386/CPU.cpp
```diff
@@ -829,6 +829,22 @@ FlatPtr read_dr6()
...
+    asm volatile("xgetbv" : "=a"(eax), "=d"(edx) : "c" (XCR_XFEATURE_ENABLED_MASK));
+    return eax + ((u64)edx << 32);
+}
+
+void write_xcr0(u64 value)
+{
+    u32 eax = value;
+    u32 edx = value >> 32;
+    asm volatile("xsetbv" :: "a"(eax), "d" (edx), "c" (XCR_XFEATURE_ENABLED_MASK));
+}
```
awesomekling, line 846: Since neither of these functions are needed after the kernel has booted into userspace, please mark them as `UNMAP_AFTER_INIT` :)

ashnewmanjones: Thanks, updated the PR.

ashnewmanjones:
I realised I can dump the xcr0 register during boot and it looks like it is being set correctly.

Kernel/Arch/i386/CPU.cpp
```diff
@@ -989,6 +1009,13 @@ UNMAP_AFTER_INIT void Processor::cpu_setup()
     if (has_feature(CPUFeature::TSC)) {
         write_cr4(read_cr4() | 0x4);
     }
+
+    if (has_feature(CPUFeature::XSAVE) & has_feature(CPUFeature::AVX)) {
```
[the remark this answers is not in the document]
ashnewmanjones: Thanks.

awesomekling:
Please squash fixups into the commit they are fixing instead of stacking commit on top of commits. :)

PR #5607, opened 2021-03-02
LibJS: Implement and use GetMethod abstract operation

linusg:


awesomekling:
Oh wow this is just really nice :)

PR #5614, opened 2021-03-02
Meta: Use and cache ccache on Github Actions

### LINKED ISSUE #5561: CI: CCache output should be cached across test runs
Since pulls usually change just a small subset of serenity we could probably speed up CI quite a bit by caching the ccache folder across runs (using something similar to our toolchain caching) and only building the changed sources.

BenWiederhake:
A test with sample-size 1 shows a reduction of the total build time of around [24 minutes](https://github.com/BenWiederhake/serenity/actions/runs/615413408) down to around [5 minutes](https://github.com/BenWiederhake/serenity/actions/runs/615490316). Holy crap, I didn't expect that!

Contains a cache-killswitch (`matrix.ccache-mark` near the top of `cmake.yml`) that can be used to restart the ccache cache directory from scratch. This *really* shouldn't ever be necessary, but @bgianfo apparently ran into issues with ccache, so here's an easy way to overcome this problem should it ever arise.

Fixes #5561.

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -16,6 +16,8 @@ jobs:
       matrix:
         debug-macros: ['ALL_DEBUG', 'NORMAL_DEBUG']
         os: [ubuntu-20.04]
+        # If ccache is broken and you would like to bust the ccache cache on Github Actions, increment this:
+        ccache-mark: [0]
 
     steps:
     - uses: actions/checkout@v2
@@ -38,7 +40,7 @@ jobs:
         wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key | sudo apt-key add -
         sudo add-apt-repository "deb http://apt.llvm.org/focal/ llvm-toolchain-focal-11 main"
         sudo apt-get update
-        sudo apt-get install clang-format-11 libstdc++-10-dev libmpfr-dev libmpc-dev ninja-build npm e2fsprogs qemu-utils qemu-system-i386
+        sudo apt-get install clang-format-11 libstdc++-10-dev libmpfr-dev libmpc-dev ninja-build npm e2fsprogs qemu-utils qemu-system-i386 ccache
     - name: Use GCC 10 instead
       run: sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-10 60 --slave /usr/bin/g++ g++ /usr/bin/g++-10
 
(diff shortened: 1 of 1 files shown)


tomuta:
While I personally love ccache and rarely ever have a problem with it, I don't know if it would pick up subtle toolchain changes appropriately. It would be a problem if it returns cached binaries that should have been rebuilt but ccache doesn't see different command line arguments etc so it thinks nothing has changed. Don't know, just want to make sure we accounted for this.

tomuta:
> While I personally love ccache and rarely ever have a problem with it, I don't know if it would pick up subtle toolchain changes appropriately. It would be a problem if it returns cached binaries that should have been rebuilt but ccache doesn't see different command line arguments etc so it thinks nothing has changed. Don't know, just want to make sure we accounted for this.

I guess that's what `ccache-mark` is for...

BenWiederhake:
I personally never had to call `ccache -C`, it always picked up on these things. I *could* invalidate the ccache dir with every LibC header change (that's the criterion we currently use to rebuild the toolchain). However, that seems like overkill, don't you think?

tomuta:
> I personally never had to call `ccache -C`, it always picked up on these things. I _could_ invalidate the ccache dir with every LibC header change (that's the criterion we currently use to rebuild the toolchain). However, that seems like overkill, don't you think?

I'm just thinking about the problems we kept running into when the toolchain was slightly modified. But again, don't know if ccache would detect this or not. Header changes yes, but applying a compiler patch or something similar, something that doesn't result in the compiled preprocessed source changing.

BenWiederhake:
Things like this are exactly what `ccache-mark` is for, indeed :)

bgianfo:
> Things like this are exactly what `ccache-mark` is for, indeed :)

Can we also include the toolchain in the cache key calculation, so it just happens automatically for the toolchain?
It's nice we have an escape hatch, but the system should just work, and we shouldn't rely on the fact that folks will know where the knob is and how to use it when the time comes.

BenWiederhake:
> so it just happens automatically for the toolchain?

The toolchain already automatically rebuilds whenever anything changes that might affect it? That includes the patch directory, the `BuildIt.sh` script, and all LibC headers. I assume you wanted to suggest to reset the ccache cache whenever we rebuild the toolchain? I'm currently implementing that.

BenWiederhake:
Done, it now discards the ccache whenever we rotate the toolchain. (I forgot to mark it as "ready for review" yesterday.)

bgianfo:
FYI It looks like the ccache isn't being found for any of the recent runs on cache restore.

BenWiederhake:
That's because the 5 builds were started in "quick" succession, at least quick enough to not be able to see each other's uploaded cache.

bgianfo:
Sure, makes sense. Thanks Ben!

BenWiederhake:
Turns out, you are right. :facepalm:

PR #5615, opened 2021-03-02
Better error reporting for failing "test-math"

BenWiederhake:
Since this seems to be a blocker for better on-target testing, here's some better error reporting about it.

Looks like the results are actually "fine", given that we have our own, wonky, self-made implementation for everything like `asin`. The only outlier seems to be `exp`, but that is only because `exp` is particularly sensitive to any errors.

This PR only sheds light onto what exactly is failing, but does not fix any of the issues. My hope is to either convince someone to do a better job at writing `asin`, or maybe we'll just lower the sensitivity of `test-math` and move on. :)

### DIFF
--- a/AK/TestSuite.h
+++ b/AK/TestSuite.h
@@ -325,13 +325,13 @@ using AK::TestSuite;
         }                                                                                      \
     } while (false)
 
-#define EXPECT_CLOSE(a, b)                                                                          \
+#define EXPECT_APPROXIMATE(a, b)                                                                    \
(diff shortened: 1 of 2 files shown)


bgianfo:
While we are here, maybe `EXPECT_APPROXIMATE` or some alternative would be a better than then `EXPECT_CLOSE`? The dual defiintion of close doesn't make it a great name for an assertion.

BenWiederhake:
Done

PR #5616, opened 2021-03-02
Applications: Add userspace UBSAN w/kernel signal

ForLoveOfCats:
This is an *extremely* simple port of the KUBSAN logic to userland. It is currently only enabled and linked to the programs under `Userland/Applications`. The types header for KUBSAN was moved into `AK` for use in both KUBSAN and userpace UBSAN. That may not be the best location for it? I kept the original copyright header from `Kernel/UBSanitizer.cpp` for `Userland/Libraries/LibSanitizer/UndefinedSanitizer.cpp` as is mostly the same contents. It may be nice to unify these two files.

The UBSAN option `vptr` had to be explicitly disabled as the existing logic on its own is not capable of handling this check. I made some headway on starting to implement the things necessary for this check but it is sufficiently complicated enough that it is probably more reasonable to figure out what the kernel is being linked to that these programs are not and provides the logic required for this check.

I added a kernel signal `SIGUBSAN` for UBSAN to kill the process so that there is a fairly nice message in the crash reporter. A number of locations in the kernel had the signal count hardcoded at 32 which caused the signal to be swallowed in several locations. These have been corrected(?) to use `NSIG` where I've spotted them. It would be awesome to add a new syscall specifically for sanitizers to call passing in an id for the specific failure which could be stored in the coredump for extra information post-crash however this seemed overly involved for this patch.

Demonstration where I've modified `About` to purposefully perform integer overflow
image


ForLoveOfCats:
The build is failing in CI with link errors but succeeding locally (even after a fully clean build on a fresh clone). I'm going to take a closer look at it tomorrow but I welcome any pointers to what might be causing this issue or what may differ between a local build and the CI build :)

Userland/Libraries/LibSanitizer/CMakeLists.txt
```diff
@@ -0,0 +1,6 @@
+set(SOURCES
+    UndefinedSanitizer.cpp
+)
+
+serenity_lib(LibSanitizer sanitizer)
+target_link_libraries(LibSanitizer LibC)
```
bgianfo, line 6: Maybe `LibSanitizer` shoud just be part of `LibC`, so we don't need to go make it a dependency everwhere?
Or we could just inject the depenedency in the `serenity_bin(..)` / `serenity_app(..)` cmake functions? 

It might also be nice to add a new `option(ENABLE_USER_SPACE_UBSAN, ...)` in the top level CMakeLists.txt defauling to ON, so folks can control this if they want?

bgianfo:
> The build is failing in CI with link errors but succeeding locally (even after a fully clean build on a fresh clone). I'm going to take a closer look at it tomorrow but I welcome any pointers to what might be causing this issue or what may differ between a local build and the CI build :)

It looks like the linker errors are coming from the `Serendipity` app which was just merged (see 71a123148a685e463bdb5fcb40c5e6571c2aaa36), looks like you just need to add `LibSanitizer` as a link dependency.

Userland/Libraries/LibSanitizer/CMakeLists.txt
ForLoveOfCats: What great timing as the CI failure is due to something just like this (new app in master not having LibSanitizer as a dep).

Putting it in LibC feels weird to me for some reason(?) but I am so far removed from an expert to make a decision on something like that. At the very least making it a dependency for programs automatically very much seems like a good idea. I will also add in the requested option :)

ForLoveOfCats:
Yup, I *literally* just figured that out

awesomekling:
Hmm, is there some way we could use the `libsanitizer` that is part of GCC instead of making our own? I'd consider this part of the toolchain which is basically the one area of SerenityOS that we don't aggressively NIH ;)

The new signal breaks the `sigset_t` type which is a 32-bit unsigned int that assumes there are only 32 signals. We need to either

(a) fix all the code that uses `sigset_t` to deal with >32 signal numbers
(b) remove an existing unused signal and replace it with `SIGUBSAN`
(c) third option I didn't think of :)

ADKaster:
One option for (c) would be to just use SIGABRT instead of a fancy new signal 🤷

tomuta:
> One option for (c) would be to just use SIGABRT instead of a fancy new signal shrug

Agreed. Somewhere on my list of things is adding support for Intel CET (just waiting on linux/qemu/kvm support so I don't have to attempt to do this on bare metal). And there's probably plenty more protection mechanisms that in theory could warrant their own signal. But in the end, it's really just a `SIGABRT`...

ForLoveOfCats:
I was envisioning that in the future a coredump of a program killed by UBSAN or ASAN would include all the information that the sanitizer would use to generate the error message and thusly the dump viewer would be able to recreate the error message and tell you exactly what went wrong. Adding a signal seemed like the simplest way to get the coredump to be aware of the sanitizer but definitely seems like the wrong choice now.

I do not have the time to commit to working toward that goal so I'll go ahead and close this PR. Because this was originally just me playing around to see what it would take to start running userland applications with a sanitizer and then deciding to submit a PR once it worked I am glad that it was able to prompt some discussion on the topic even if this work will not directly contribute to the eventual implementation that I am sure will come to exist. Thanks for the friendly environment and cheers!

PR #5618, opened 2021-03-03
WindowServer: Mark screen dirty when changing highlighted window

### LINKED ISSUE #5599: WindowServer(?): Terminal draws over other windows' frames when in focus-change preview (after super-tab key combo)
Ah yes, hidden shortcuts and their hidden bugs:
2021-03-02-003950_2560x2160_scrot

(open `sm` and `br`, switch focus a bit and see Terminal rendering partially above SystemMonitor or Browser)

_Originally posted by @alimpfard in https://github.com/SerenityOS/serenity/pull/5590#discussion_r585048643_

tomuta:
Because the z-order changes we not only need to recompute occlusions
but we also need to mark the screen area of the previous and new
highlighted window as dirty.

Fixes #5599

### DIFF
--- a/Userland/Services/WindowServer/Window.cpp
+++ b/Userland/Services/WindowServer/Window.cpp
@@ -576,6 +576,14 @@ void Window::prepare_dirty_rects()
             m_dirty_rects = rect();
     } else {
         m_dirty_rects.move_by(frame().render_rect().location());
+        if (m_invalidated_frame) {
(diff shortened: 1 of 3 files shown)


tomuta:
It feels like the calls to `Compositor::invalidate_screen` shouldn't be necessary, but they are right now. But that's an investigation I'm probably going to do some other day.

PR #5624, opened 2021-03-03
AK+Kernel: Remove NO_DISCARD macro hack

linusg:
This was added as clang-format would mess up the formatting when using `[[nodiscard]]` on a class, which is no longer the case.


awesomekling:
Oh hey nice

PR #5628, opened 2021-03-03
Base: Add man pages for null/zero/full character devices

bcoles:


Base/usr/share/man/man4/full.md
```diff
@@ -0,0 +1,35 @@
...
+
+Reading from `/dev/full` returns '\0' bytes and exits with status 0.
+
+Writing to `/dev/full` fails with ENOSPC error.
+
+To create it manually:
+
+```sh
+mknod /dev/full c 1 7
+chmod 666 /dev/full
```
bcoles, line 17: Currently, for no adequately explained reason, `/dev/full` is created with `600` permissions by default (not `666`).

The intention of this device is to allow programs (including userland programs) to test error handling when writing to devices which are full. Limiting this to root only by default seems like an oversight.

Base/usr/share/man/man4/zero.md
```diff
@@ -0,0 +1,26 @@
+## Name
+
+zero - data sink
+
+## Description
+
+`/dev/zero` is a character device file which discards input.
```
angus-g, line 7: data source?

bcoles: yes

PR #5639, opened 2021-03-04
LibC: Fix typo in fenv.h

RealKC:
cc @danboid

Don't think this is enough for hatari though, tried building it and it complained about _SC_CLK_TCK not being defined.

### DIFF
--- a/Userland/Libraries/LibC/fenv.cpp
+++ b/Userland/Libraries/LibC/fenv.cpp
@@ -157,7 +157,7 @@ int fegetround()
 
 int fesetround(int rounding_mode)
 {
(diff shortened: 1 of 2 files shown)


danboid:
Yep, a bunch of math stuff seems to be missing too. Hatari has been ported to some pretty niche platforms already such as RISCOS and only really needs SDL2 so I don't think we'll far off having a nice 16 bit emu.

RealKC:
Think that typo might've been the last of the math stuff missing? iirc after looking at your original issue, the <math.h> functions were either implemented by PRs or through patches from anonymous people:tm:, and <fenv.h> was the last thing

PR #5643, opened 2021-03-04
Shell: Fix and document prompt

BenWiederhake:


Base/usr/share/man/man7/Shell-vars.md
```diff
@@ -40,10 +40,13 @@ Its default value is `~/.history`.
...
-- `\\h` : the current hostname
-- `\\w` : a collapsed path (relative to home) to the current directory
-- `\\p` : the string '$' (or '#' if the user is 'root')
+- `\\a`: bell character (behavior depends on terminal)
+- `\\e`: escape character (`0x1b`)
+- `\\h`: the current hostname
+- `\\p`: the string '$' (or '#' if the user is 'root')
+- `\\u`: the current username
+- `\\w`: a collapsed path (relative to home) to the current directory
+- `\\X`: reset style (foreground and background color, etc)
```
alimpfard, line 49: I've always thought that some of these feel like unnecessary duplication, why have `\\e` when the string can contain the escape `\e`? why add `\\X` when you can just `\e[0m`?
the only ones I see that are actually needed are h,p,u,w - the rest can just be written as normal escapes

Not a problem with this PR, just ranting about this thing

awesomekling (changes requested):
The absence of username is intentional.

alimpfard:
@awesomekling Any particular reason for that? if we don't want the username to show up, maybe we could also remove the `@`?

awesomekling:
No reason other than changing up the look a little bit. If you don’t like it feel free to make a local override :)

alimpfard:
I don't mind either way, I just thought that `@courage:~ $` looked a bit odd :shrug:

BenWiederhake:
Especially since the username *does* change while using the shell, whereas the host does not.

BenWiederhake:
Alrighty, then I guess this PR won't be accepted. I'll put the other commit in a seperate PR.

In case someone's reading this: Put this in your `sync-local.sh` in the root directory (above `Build/`), and make sure to make it executable:

```
#!/bin/bash

set -e

# Make the prompt include the user
cat << 'EOF' >> mnt/etc/shellrc
export PROMPT="\\X\\u@\\h:\\w\\a\\e[32;1m\\u@\\h\\e[0m:\\e[34;1m\\w\\e[0m \\p "
EOF
```

PR #5646, opened 2021-03-04
Profiler: Get perfcore file from ArgsParser

### LINKED ISSUE #5641: Profiler: Profiler has wrong/incomplete footprint
At the momnent the Profiler/pv has the footprint  
`Profiler [--pid PID]`  
which is of course wrong, because we can pass in a Profile our selfs, wich would probably look like  
`Profiler [FILE] [--pid PID]`

Because of this issue, every time we start the Profiler with a previously recorded Profile (ie /proc/profile) we get its"wrong arguments" output, but it just keeps running and doing what we want it to...

linusg:
Fixes #5641.


bgianfo:
Thank you for fixig this, it was bugging me the other day 👍

PR #5650, opened 2021-03-05
WebServer: Serve X-Frame-Options and X-Content-Type-Options HTTP headers

bcoles:
* X-Frame-Options: SAMEORIGIN
* X-Content-Type-Options: nosniff

Help prevent client side attacks against clients using the web server. (Also lets us test these features once they're implemented in `Browser`).

### DIFF
--- a/Userland/Services/WebServer/Client.cpp
+++ b/Userland/Services/WebServer/Client.cpp
@@ -137,6 +137,8 @@ void Client::send_response(InputStream& response, const HTTP::HttpRequest& reque
     StringBuilder builder;
     builder.append("HTTP/1.0 200 OK\r\n");
     builder.append("Server: WebServer (SerenityOS)\r\n");
+    builder.append("X-Frame-Options: SAMEORIGIN\r\n");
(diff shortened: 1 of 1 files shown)


Userland/Services/WebServer/Client.cpp
```diff
@@ -137,6 +137,9 @@ void Client::send_response(InputStream& response, const HTTP::HttpRequest& reque
     StringBuilder builder;
     builder.append("HTTP/1.0 200 OK\r\n");
     builder.append("Server: WebServer (SerenityOS)\r\n");
+    builder.append("X-Frame-Options: SAMEORIGIN\r\n");
+    builder.append("X-Xss-Protection: 1; mode=block\r\n");
```
awesomekling, line 141: I think the canonical name of this header is `X-XSS-Protection` but it probably doesn't matter.

bcoles: Correct. We probably shouldn't serve this header anyway. It sucks. And `Browser` shouldn't support it either.

PR #5655, opened 2021-03-05
LibM: optimized (branchless) copysign

nickwanninger:
This version just copies the sign bit from y into x. This is just an optimization after @awesomekling mentioned there is likely a more clever way to do it in his recent video. I think this is the clever way :^)

### DIFF
--- a/Userland/Libraries/LibM/math.cpp
+++ b/Userland/Libraries/LibM/math.cpp
@@ -745,12 +745,11 @@ long double nexttowardl(long double, long double) NOEXCEPT
 
 double copysign(double x, double y)
 {
-    if (x < 0 && y < 0)
-        return x;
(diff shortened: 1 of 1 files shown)


TheMorc:
A minor thing, please squash those commits into a single commit.

nickwanninger:
Sorry, having some trouble with git on my end, I'll see what I can do

TheMorc:
> Sorry, having some trouble with git on my end, I'll see what I can do

No problem, take your time. I also had problems with getting it to work on the first time. 
You can usually do `git rebase -i HEAD~3` and then you change pick to fixup for the commit you want it to merge with.

nickwanninger:
Alright that should do it

awesomekling:
Oh that's super slick :)

PR #5656, opened 2021-03-05
LibGfx: Fix draw_triangle() precision issues

### LINKED ISSUE #5647: Cube: random artifacts
I've noticed them forever, but I wonder if this is a Cube bug or maybe some odd kernel fpu problem? They seem to be somewhat random, sometimes more pronounced than other times. Recorded a screen video and highlighted some of these artifacts.

Screenshot from 2021-03-04 18-51-38
Screenshot from 2021-03-04 18-51-16
Screenshot from 2021-03-04 18-51-28
Screenshot from 2021-03-04 18-51-06

sunverwerth:
This fixes some precision issues in Painter::draw_triangle()
that caused the Cube demo to leave pixels empty between triangles
under certain angles.

Also adds some extra comments for clarity and early returns when
parts of the triangle do not need to be drawn.

fixes #5647

### DIFF
--- a/Userland/Libraries/LibGfx/Painter.cpp
+++ b/Userland/Libraries/LibGfx/Painter.cpp
@@ -456,12 +456,11 @@ void Painter::draw_triangle(const IntPoint& a, const IntPoint& b, const IntPoint
 {
     VERIFY(scale() == 1); // FIXME: Add scaling support.
 
-    RGBA32 rgba = color.value();
-
     IntPoint p0(a);
     IntPoint p1(b);
     IntPoint p2(c);
 
+    // sort points from top to bottom
     if (p0.y() > p1.y())
         swap(p0, p1);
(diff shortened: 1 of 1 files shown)


Userland/Libraries/LibGfx/Painter.cpp
```diff
@@ -452,58 +452,68 @@ void Painter::draw_bitmap(const IntPoint& p, const GlyphBitmap& bitmap, Color co
     }
 }
 
-void Painter::draw_triangle(const IntPoint& a, const IntPoint& b, const IntPoint& c, Color color)
+void Painter::draw_triangle(const IntPoint& a, const IntPoint& b, const IntPoint& c, const Color& color)
```
tomuta, line 455: You can just pass `Color` by value, it's only a 32 bit value. I believe we pass it by value almost everywhere.

PR #5659, opened 2021-03-06
LibWeb: Make CSS background image painting respect destination origin and transparency

mihaip:
It was previously using draw_tiled_bitmap, which always aligns the
tiles with the global origin and does not respect the alpha of the
source. Switch to a new Painter::blit_tiled helper which uses
Painter::blit under the hood, which has more correct behavior.


mihaip:
There's a lot more that we could do for CSS background image (respect `background-position`, `background-repeat`, etc.), but this fixes a rendering glitch on my blog's homepage:

| Before | After |
| - | - |
| Screen Shot 2021-03-05 at 4 21 00 PM |  Screen Shot 2021-03-05 at 4 32 37 PM |

awesomekling:
Kickass :)

PR #5661, opened 2021-03-06
Base: Add egrep and rgrep aliases to /etc/shellrc

bcoles:
Common aliases. Available by default on Debian systems (and derivatives such as Ubuntu) as simple shell wrapper scripts.

### DIFF
--- a/Base/etc/shellrc
+++ b/Base/etc/shellrc
@@ -27,6 +27,8 @@ alias sl=Solitaire
 alias ue=UserspaceEmulator
 alias welcome=Serendipity
 
+alias rgrep="grep -r"
+alias egrep="grep -E"
 alias ll='ls -l'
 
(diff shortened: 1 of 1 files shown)


awesomekling:
Curiously, it seems that `grep -E` is the default operating behavior of our `grep` and there's no way (?) to opt out.

bcoles:
> 
> 
> Curiously, it seems that `grep -E` is the default operating behavior of our `grep` and there's no way (?) to opt out.

I believe that is the default for GNU grep too.

```
$ grep r..t /etc/passwd
root:x:0:0:root:/root:/bin/bash
```

The traditional approach to opt-out is to support a `-F` flag. Presumably, the best approach to resolve this issue would be to copy the functionality from `fgrep` into `grep` as `-F`. This also means that the `fgrep` utility could be removed entirely and instead replaced with an alias `fgrep="grep -F"`.

PR #5662, opened 2021-03-06
LibM: Less naive implementation of copysign() :)

Jorropo:
I just wanted to do that after watching the video. :)
This change from a variable `O() = 2` at best to `O() = 5` at worst to a constant `O() = 2` (counting the comparisions).

Honestly the difference isn't that big and almost nobody use it but whatever.

I'm pretty sure a better implementation can be done in 2 or 3 cycles in asm.


Jorropo:
I've just seen #5655

Way better :) I'm closing this one.

PR #5664, opened 2021-03-06
Ports/git: Forcing NO_IPV6 option.

Jorropo:
This fixes the build, how ever I don't know why the auto detection broke in the first place.


alimpfard:
Probably because we have _some_ of the ipv6-related structures, and it just so happened that git was using one of those to check for compat

Jorropo:
> Probably because we have _some_ of the ipv6-related structures, and it just so happened that git was using one of those to check for compat

Yes, I know that :) I was unclear, I wish to know the exact one and find a cleaner solution than `CFLAGS` (remove the struct, make it or making it optional, ... idk)).

I tried a bisect to find this but this was taking ages to rebuild each time.

Issue #5589, opened 2021-03-01
TextEditor: Accept "file:line:col" argument format


BenWiederhake:
Many tools like g++, grep, shellcheck, etc. print file locations as file:line:col, for example `Meta/CMake/utils.cmake:89` or:

```
../Kernel/Net/TCPSocket.cpp:245:34: error: 'struct Kernel::TCPSocket::OutgoingPacket' has no member named 'tx_timexx'; did you mean 'tx_time'?
```

Many editors like QtCreator, Geany, or CLion understand this format, and open the file with the text cursor ready to go at the desired location. b474f4916479fbf64e6fb5a66cb25b8496e153b3 inspired me to request that our TextEditor (and maybe HackStudio?) should do the same. :)

Maybe this could be a LibCore feature? Dunno.

CC @ry755

tomuta:
What if the filename contains a colon? It's a valid file name character.

ry755:
What if there was a command line option that enables this behavior, and otherwise the colon would be considered part of the filename? (possibly replacing the command line option added in b474f4916479fbf64e6fb5a66cb25b8496e153b3)
For example, Visual Studio Code has this option:
`-g --goto <file:line[:character]> Open a file at the path on the specified line and character position.`
`code -g example.txt:2` will open `example.txt` to line 2, while `code example.txt:2` will open a file called `example.txt:2`

Something similar could be used here in case the user wants colons to be interpreted this way, and the default behavior would stay how it currently is.

ElDonad:
Hello there ! 

Yeah, I thought of that solution too. I settled initially for the query parameters because the LaunchServer uses file extension to infer which application to use, and we would need to make extra checks to ensure the line number isn't mistaken for part of the extension, when query parameters were already sanitized. But either way would be totally fine, I just took the path of least resistance 'cause I'm lazy...

My assumption is that HackStudio probably shouldn't rely on the LaunchServer to open files, since I assume we want them to be opened in HackStudio whatever the default behaviour is. 

Also the Terminal relies on a hidden "hyperlink", at least to know where links point to, and since gcc and the like probably don't do that, either we could make them do it, or we could implement some sort of custom pattern detection in the HackStudio terminal to handle this kind of links. Or something else, that's all I can think of at the moment.

As for the system Terminal, I'm not sure what to think...

willmcpherson2:
Why not just try to open the file and if it doesn't exist, try to interpret it as a file + line number?

PR #5665, opened 2021-03-06
TextEditor: Specify the starting line and column number using colons

ry755:
This allows the user to specify a specific line and column number to start at when opening a file in TextEditor through the terminal, by adding colon(s) after the file name.

For example, `TextEditor ReadMe.md:10:5` will open ReadMe.md and put the cursor on line 10 at column 5.

To ensure that the user isn't trying to open a file that actually has colons in its name, it checks if the file exists before parsing it.

This replaces the -l/--line-number command line option added in b474f4916479fbf64e6fb5a66cb25b8496e153b3 and closes issue #5589

I'm still very new to SerenityOS development (and C++ in general) so any tips would be appreciated!

### DIFF
--- a/Userland/Applications/TextEditor/main.cpp
+++ b/Userland/Applications/TextEditor/main.cpp
@@ -4,6 +4,7 @@
  * SPDX-License-Identifier: BSD-2-Clause
  */
 
+#include "FileArgument.h"
 #include "MainWidget.h"
 #include <LibCore/ArgsParser.h>
 #include <LibGUI/Menubar.h>
@@ -27,12 +28,10 @@ int main(int argc, char** argv)
     }
 
     const char* preview_mode = "auto";
-    int initial_line_number = 0;
     const char* file_to_edit = nullptr;
     Core::ArgsParser parser;
     parser.add_option(preview_mode, "Preview mode, one of 'none', 'html', 'markdown', 'auto'", "preview-mode", '\0', "mode");
-    parser.add_option(initial_line_number, "Start at line number", "line-number", 'l', "line");
-    parser.add_positional_argument(file_to_edit, "File to edit", "file", Core::ArgsParser::Required::No);
+    parser.add_positional_argument(file_to_edit, "File to edit, with optional starting line and column number", "file[:line[:column]]", Core::ArgsParser::Required::No);
 
     parser.parse(argc, argv);
 
@@ -71,14 +70,13 @@ int main(int argc, char** argv)
     window->set_menubar(menubar);
 
     if (file_to_edit) {
-        if (!text_widget.open_file(file_to_edit))
+        // A file name was passed, parse any possible line and column numbers included.
+        FileArgument parsed_argument(file_to_edit);
+        if (!text_widget.open_file(parsed_argument.file_name()))
             return 1;
+        text_widget.editor().set_cursor_and_focus_line(parsed_argument.line().value_or(1) - 1, parsed_argument.column().value_or(0));
     }
-
     text_widget.update_title();
-
-    if (initial_line_number != 0)
-        text_widget.editor().set_cursor_and_focus_line(initial_line_number - 1, 0);
 
     window->show();
     window->set_icon(app_icon.bitmap_for_size(16));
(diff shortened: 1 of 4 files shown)


Userland/Applications/TextEditor/main.cpp
```diff
@@ -88,13 +88,51 @@ int main(int argc, char** argv)
...
+            if (Core::File::exists(file_to_edit)) {
+                // A file exists with a name including the colon(s), open it.
+                text_widget.open_file(file_to_edit);
+            } else {
+                // A file doesn't exist with a name including the colon(s), maybe the user entered line/column coordinates?
+                auto file_name_vector = StringView(file_to_edit).split_view(':');
+                auto file_name_vector_size = file_name_vector.size();
+                if (!file_name_vector.is_empty()) {
+                    auto file_name = file_name_vector.at(0);
+                    if (file_name_vector_size >= 3) {
```
alimpfard, line 106: What about the other components?
`foo:bar.cpp:10:42` would open `foo` here.

Maybe invent a `split_end_limit()` and use that? Or just join the parts back with a colon

ry755: Sorry for taking so long to get back to this! I rewrote it to use regex instead, which solves this issue. Now you're able to have colons as part of the file name without any issues.
I think regex is a better solution anyways compared to the somewhat hacky if-else statements that I had before.

alimpfard: I personally find the regex solution more clear (and indeed, it does handle this edge case correctly), so that's a lgtm from me.

Userland/Applications/TextEditor/main.cpp
```diff
@@ -88,13 +88,60 @@ int main(int argc, char** argv)
...
+            } else {
+                // A file doesn't exist with the full specified name, maybe the user entered line/column coordinates?
+                Regex<PosixExtended> line_and_column_number("^(.+?)(:([0-9]+):?([0-9]+)?)?$");
+                RegexResult result = match(file_to_edit, line_and_column_number, PosixFlags::Global | PosixFlags::Multiline | PosixFlags::Ungreedy);
+
+                // Match 0 group 0: file name
+                // Match 0 group 1: line number
+                // Match 0 group 2: column number
+
+                if (result.capture_group_matches.at(0).size() > 3) {
```
alimpfard, line 109: Just a nitpick,
Maybe extract this `result.capture_group_matches.at(0)` above just once? `.at(0).at(0)` looks rather odd

```c++
auto& groups = result.capture_group_matches.at(0);
if (groups.size() > 3)
    ...
```

ry755: Right, thanks!

alimpfard (approved):

Userland/Applications/TextEditor/main.cpp
```diff
@@ -88,13 +88,61 @@ int main(int argc, char** argv)
...
+    // If a colon exists in the file name argument, the user may have specified line and column numbers.
+    // First check if the file exists before doing any further processing. If so, open it.
+    // Otherwise, split apart the line and column numbers from the file name, open the file, and jump to the specified line/column.
+
+    if (file_to_edit) {
+        if (StringView(file_to_edit).contains(':')) {
+            if (Core::File::exists(file_to_edit)) {
+                // A file exists with the full specified name, open it without doing any further processing.
+                text_widget.open_file(file_to_edit);
+            } else {
```
awesomekling, line 100: This is a lot of nesting. Let's find a way to write this code so we don't have to nest as much :)

How about we move this logic to a function that takes the `file_to_edit` string and turns it into a `{ file_name, Optional<size_t> line, Optional<size_t> column }`?

ry755: :thumbsup:
I've turned this into a `FileArgument` class which takes the `file_to_edit` string as does as you described, does this look ok?

awesomekling:
This looks nice, but please remove the merge commit from the branch since that's preventing me from accepting the changes :)

ry755:
Sorry about that, I merged it in an attempt to resolve a conflict. That conflict is still there now, is there something else I should do to resolve it?

IdanHo:
> Sorry about that, I merged it in an attempt to resolve a conflict. That conflict is still there now, is there something else I should do to resolve it?

I should look into rebasing your changes instead of merging them :)

linusg:
@ry755 ping, please rebase this PR :)

ry755:
Oh crap, thank you for pinging me again! It has been rebased and should be good to merge now :thumbsup:

ry755:
Rebased again to resolve a few new conflicts (ping @linusg)

PR #5669, opened 2021-03-06
LibWeb: Add a couple child node operations to Node and add node types

Lubrsi:


awesomekling:
`Element.childNodes` should actually return a live `NodeList` but ofc we don't have the facilities to support that at the moment so this is still a step up for web compat :)

PR #5673, opened 2021-03-07
AK + Kernel + LibGUI:  Add AK_ENUM_BITWISE_OPERATORS(..) to enable type-safe, enum bitwise operations and start using it

bgianfo:
This change introduces `AK_ENUM_BITWISE_OPERATORS(..)` which when
enabled for an enum, will automatically declare all the necessary
bitwise operators for that enum.
    
This allows bit masks enums to be used as first class, type safe, citizens and
starts switching some Kernel and LibGUI code to start using it.

### DIFF
--- a/AK/EnumBits.h
+++ b/AK/EnumBits.h
@@ -1,0 +1,101 @@
+/*
+ * Copyright (c) 2021, Brian Gianforcaro <b.gianfo@gmail.com>
+ * All rights reserved.
+ *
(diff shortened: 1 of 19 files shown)


bgianfo:
Just noticed UI isn't booting after this change... debugging.

bgianfo:
Bug fixed :)

awesomekling:
This makes the code look nicer indeed. :)

PR #5674, opened 2021-03-07
Kernel: Fix unaligned read inside map_ebda()

boricj:
Fixes this KUBSAN warning:
```
[#0 Kernel]: Early access to ACPI tables for interrupt setup
[#0 Kernel]: KUBSAN: reference binding to misaligned address 0xc1104413 of type 'short unsigned int'
[#0 Kernel]: KUBSAN: at .././Kernel/VM/TypedMapping.h, line 41, column: 33
[#0 Kernel]: 0xc04bb7c7 (next: 0xc0107f38)
[#0 Kernel]: 0xc04bd866 (next: 0xc0107fd8)
[#0 Kernel]: 0xc0120d54 (next: 0xc0108058)
[#0 Kernel]: 0xc05c31fe (next: 0xc01080c8)
[#0 Kernel]: 0xc05e8314 (next: 0xc0108128)
[#0 Kernel]: 0xc05e86ef (next: 0xc0108178)
[#0 Kernel]: 0xc05e87a1 (next: 0xc01081b8)
[#0 Kernel]: 0xc060ea46 (next: 0xc0108298)
[#0 Kernel]: 0xc0100145 (next: 0x00000000)
[#0 Kernel]: KUBSAN: load of misaligned address 0xc1104413 of type 'short unsigned int'
[#0 Kernel]: KUBSAN: at ../Kernel/Arch/PC/BIOS.cpp, line 49, column: 24
[#0 Kernel]: 0xc04bb7c7 (next: 0xc0107f38)
[#0 Kernel]: 0xc04bd866 (next: 0xc0107fd8)
[#0 Kernel]: 0xc0120d72 (next: 0xc0108058)
[#0 Kernel]: 0xc05c31fe (next: 0xc01080c8)
[#0 Kernel]: 0xc05e8314 (next: 0xc0108128)
[#0 Kernel]: 0xc05e86ef (next: 0xc0108178)
[#0 Kernel]: 0xc05e87a1 (next: 0xc01081b8)
[#0 Kernel]: 0xc060ea46 (next: 0xc0108298)
[#0 Kernel]: 0xc0100145 (next: 0x00000000)
[Kernel]: Interrupts: Switch to Legacy PIC mode
[Kernel]: PIC(i8259): cascading mode, vectors 0x50-0x5f
[#0 Kernel]: Trying to unregister unused handler (?)
[#0 Kernel]: Trying to unregister unused handler (?)
[#0 Kernel]: Interrupts: Detected Dual i8259
[...]

PR #5677, opened 2021-03-07
Kernel+Fuzzing: Harden Kernel just a tad more, add sysctl for deadliness of UBSan

BenWiederhake:
This PR:
- Makes `fuzz-syscalls` less brittle, and fixes what may be the last bug findable by such a simple program (`sys$create_thread`, a silly technicality)
- Adds an "obviously missing" Test for `AK::Checked<unsigned>`
- Adds a sysctl (`/proc/sys/ubsan_is_deadly`) to immediately panic the Kernel on UB.
- Perhaps most controversially: This PR makes `ubsan_is_deadly` enabled by default. This is possible since @boricj [fixed](https://github.com/SerenityOS/serenity/pull/5674) what seems to be the last UB during boot. Hooray! :partying_face: 

I'm really not sure whether `ubsan_is_deadly` should be enabled by default, but this flag helped me during fuzzing and will probably be interesting during future testing/fuzzing, too.

### DIFF
--- a/AK/Tests/TestChecked.cpp
+++ b/AK/Tests/TestChecked.cpp
@@ -118,6 +118,27 @@ TEST_CASE(detects_signed_overflow)
     EXPECT((Checked<i64>(-0x4000000000000000) - Checked<i64>(0x4000000000000001)).has_overflow());
 }
 
+TEST_CASE(detects_unsigned_overflow)
+{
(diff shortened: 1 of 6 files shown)


BenWiederhake:
(Changes since last push: Attribution to fuzz-syscalls in commit message.)

boricj:
Nickpicking the PR's description and the first commit's message: my GitHub handle is @boricj, not boric.

BenWiederhake:
Changes since last push: Fix @boricj's username :P

PR #5679, opened 2021-03-07
Everywhere: Add some more x86_64 support

Hendiadyoin1:
~- [ ] Kernel~
   ~- [?] Boot~
    ~- Was done before~
    ~- Have we enabled 64 bit (long) mode yet?~
  ~- [ ] CPU~
    ~- [ ] Interrupts~
      ~- [ ] New IDTGate struct~
        ~- [x] Write It~
        ~- [ ] Use It~
      ~- [ ] Fix Interrupt asm_entry~
    ~- [ ] ProcInfo~
    ~- [ ] SafeMem~
  ~- [ ] signal trampoline~
  ~- [ ] systemcalls~
    ~- [ ] sys$sigaction~
    ~- [ ] sys$sigreturn~
    ~- [ ] system call entry~
  ~- [x] UHCI/USB Driver~
~- [ ] Registers~
  ~- [X] R...~
  ~- [ ] FPU~
  ~- [ ] MMX XMM YMM...~
~- [ ] Dynamic Loader~
  ~- [ ] el64 binaries~
  ~- [ ] elf64 symbols~
  ~- [ ] elf64 relocations~
~- [ ] LibC + LibM~
  ~- [x] Setjump / Longjump~
  ~- [ ] asm optimized functions~
~- [ ] Userspace Cleanup~
~- [ ] Userspace Emulator (this **will** break)~
~- [X] Fix Copyright statements~
<hr/>

So I'm breaking short at this point.  
What I have done:
* Fix the UHCI Controller
* Theoretically enter long-mode
* "Improve" Kernel structure to make it easier to port and reduce code duplication
  * Someone should try to demangle CPU.cpp a bit more (it's a mess)
* Some other Build errors I hve encountered in x86_64 mode
* IDT-struct for 64 bit
* TSS-struct for 64 bit (how ever you are supposed to use them)
* Work on the Register struct in LibC (Is the order important, ie standardized in posix, I could only find the manpage to ptrace and that said "look at sys/user.h")


ADKaster:
Don't forget to add to your list:

- Signal trampolines in kernel (asm)
- sigaction /sigreturn syscalls
- syscall entry (sysenter instruction?)
- dynamic loader + kernel + libelf
    - handle elf64 class elf binaries
    - handle elf64 symbols
    - handle elf64 relocations and relocation types
- libc + libm
    - setjump / longjump
    - any asm optimized functions
- Other things I forgot that have hard coded 32 bit x86 assumptions.

Many of these still apply even if the kernel is 64 bit with a 32 bit user space. 🙂

Userland/Libraries/LibELF/Arch/x86_64/plt_trampoline.S
```diff
@@ -0,0 +1,58 @@
...
+	.type	_plt_trampoline,@function
+_plt_trampoline:	# (obj, reloff)
+	pushf				# save registers
+	push	%rax
+	push	%rcx
+	push	%rdx
+
+	push	20(%esp)		 # Copy of reloff
+	push	20(%esp)		 # Copy of obj
+	call	_fixup_plt_entry@PLT # Call the binder
```
ADKaster, line 48: I would verify that this matches the reference asm in netbsd for the x86_64 arch. I wouldn't be surprised if it needs 8 byte alignment for example. FWIW, the 32-bit plt trampoline part is known to not work -- neither itamar nor myself have taken the time to re-enable lazy relocations.

Hendiadyoin1: I just copied the i386 code and fixed push and pop nothing more so there will probably be some more changes needed....

Userland/Libraries/LibELF/Arch/x86_64/plt_trampoline.S
```diff
@@ -0,0 +1,58 @@
...
+	addl	$8,%esp			 # pop binder args
+	movl	%eax,20(%esp)	 # Store function to be called in obj
+
+	pop	%rdx
+	pop	%rcx
+	pop	%rax
+	popf
+
+	leal	4(%esp),%esp		# Discard reloff, do not change eflags
+	ret
```
linusg, line 58: s/tabs/spaces/

Userland/Libraries/LibELF/Arch/x86_64/plt_trampoline.S
Hendiadyoin1: looking at the netbsd implimentation, I have no clue what to change  
Do you have any expertise of this topic?

ref:  
http://ftp.jp.freebsd.org/pub/NetBSD/NetBSD-current/src/libexec/ld.elf_so/arch/x86_64/rtld_start.S

Userland/Libraries/LibELF/Arch/x86_64/plt_trampoline.S
Hendiadyoin1: >     * setjump / longjump

that was already implemented as it seems

Hendiadyoin1:
>     * setjump / longjump

That was already implemented as it seems

Userland/Libraries/LibELF/Arch/x86_64/plt_trampoline.S
ADKaster: So essentially what's happening is this function is "_rtld_bind". In the procedure linkage table (PLT) of each loaded ELF object we have entries for every external symbol we found during the build-time link. The table has something along the lines of a relative jump to a certain point in the global offset table, and a relocation index. The dimension dynamic loader puts this procedure's address in the GOT as a function pointer for all the PLT entries to jump up so that we can call into the dynamic object or dynamic loader for that object (library) to perform symbol lookup and patch the PLT at runtime. 

Which is all called "lazy symbol resolution" and has the purpose of speeding up application load times at the cost of a once per symbol relocation during the execution of the program. Which has security implications and such. The implementation in NetBSD essentially does a manual procedure call prologue/epilogue for all the caller saved registers in the x86_64 ABI. It also does some adjustments to make sure the values that were pushed by the stub code in the PLT are in the proper registers for the C linkage function it ends up calling to do the work of the lazy relocation.

Hendiadyoin1: You know what? I gave you access to the fork.  
The branch is called Hendiadyoin1:x86_64-Support!  
Feel free to do that part and everything that catches your eye

Hendiadyoin1:
Looking at the IRQ implementation I find
https://github.com/SerenityOS/serenity/blob/bac0dd5e3d8a6b5d4d9fbb17fb7b346c266ba3c6/Kernel/Arch/i386/CPU.cpp#L485-L495
where the the 0x00080000 and the 0x8e00 / 0xef00 bug me a bit
I can guess that the 0x00080000 maps to the selector-part, but what does the selector actually mean?  
And the other one are probably the attribute flags, where one sets Ring3 access, but following my guess, why do they both have different GateTypes; Is it important, that Ring0 handles are interrupt Gates and Userspace/Ring3 handles are Trapgates; Is that always the case?

Userland/Libraries/LibELF/Arch/x86_64/plt_trampoline.S
Hendiadyoin1: Ohh... You mean that that file uses Tabs instead of spaces....

This is nearly 1:1 the original File and may or may not be totally rewritten.  
The Tabs are used in the old file too though...

Hendiadyoin1:
So Im working at context switching at the moment and according to the [osDevWiki](https://wiki.osdev.org/Task_State_Segment) the 64 bit version of the TSS does not support a place for the stackPtr or the instructionPtr, we theoretically have some unused space in the struct, but only one of these seems to be usable (the other big enough space seems to be reserved for a non existent IST0) has anyone an idea how to go at this topic?

Hendiadyoin1:
So I think I gonna cut the scope of this pull request down and finish up what I was doing, due to me not knowing enough about the inner workings of a lot of the lowest level stuff.

Hendiadyoin1:
So apperently clang-format 10 is the standard on my ubuntu and it is wierd in comparison to clang-format 11...

awesomekling:
Thank you for moving this forward :)
It's still all very clunky and fuzzy but it's progress.

PR #5680, opened 2021-03-07
LibC: Misc additions and a typo fix

RealKC:


awesomekling:
Actually, `stdbool.h`, `stdnoreturn.h` and `stdalign.h` are all provided by the compiler and should not be part of **LibC**. So I think the correct fix here is to remove our `stdbool.h` :)

RealKC:
Oh that would also work then, so drop the stdbool, stdnoreturn and stdalign commits and remove stdbool.h?

awesomekling:
Yeah :)

RealKC:
Think that should do it

PR #5683, opened 2021-03-07
LibCompress: Rename libcompression.so to libcompress.so

GMTA:
Andreas mentioned this small discrepancy in his Diablo vid, so let's rename the library file.


linusg:
At least the SDL2_ttf port (30b957e46cfc3688754abcd1bc6e6e59c6a6552f) needs to be updated as well

GMTA:
> At least the SDL2_ttf port ([30b957e](https://github.com/SerenityOS/serenity/commit/30b957e46cfc3688754abcd1bc6e6e59c6a6552f)) needs to be updated as well

Yeah, that dropped in `master` around the same time as this PR. Fixed!

PR #5686, opened 2021-03-07
Meta: Add a build helper script

tomuta:
This script automatically manages the toolchain and cmake/ninja folders
making it easier to manage the different target architectures.

```
Usage: ./Meta/serenity.sh COMMAND [TARGET] [ARGS...]
  Supported TARGETs: i686 (default), x86_64, lagom
  Supported COMMANDs:
    build:      Compiles the target binaries, [ARGS...] are passed through to ninja
    install:    Installs the target binary
    image:      Creates a disk image with the installed binaries
    run:        TARGET lagom [specific_test]: Runs the unit tests on the build host,
                    or if specific_test is specified only this test.
                All other TARGETs: ./Meta/serenity.sh run [TARGET] [kernel_cmd_line]
                    Runs the built image in QEMU, and optionally passes the
                    kernel_cmd_line to the Kernel
    delete:     Removes the build environment for the TARGET
    recreate:   Deletes and re-creates the build environment for TARGET
    rebuild:    Deletes and re-creates the build environment, and compiles for TARGET
    kaddr2line: ./Meta/serenity.sh kaddr2line TARGET ADDRESS
                    Resolves the ADDRESS in the Kernel/Kernel binary to a file:line
    addr2line:  ./Meta/serenity.sh addr2line TARGET BINARY_FILE ADDRESS
                    Resolves the ADDRESS in BINARY_FILE to a file:line. It will
                    attempt to find the BINARY_FILE in the appropriate build directory

    rebuild-toolchain: Deletes and re-builds the TARGET's toolchain

  Examples:
    ./Meta/serenity.sh run i686 smp=on
        Runs the image in QEMU passing "smp=on" to the kernel command line
    ./Meta/serenity.sh run
        Runs the image for the default TARGET i686 in QEMU
    ./Meta/serenity.sh run lagom
        Runs the unit tests on the build host
    ./Meta/serenity.sh kaddr2line i686 0x12345678
        Resolves the address 0x12345678 in the Kernel binary
    ./Meta/serenity.sh addr2line i686 WindowServer 0x12345678
        Resolves the address 0x12345678 in the WindowServer binary
```

### DIFF
--- a/Meta/serenity.sh
+++ b/Meta/serenity.sh
@@ -1,0 +1,206 @@
+#!/bin/env bash
+set -e
+
+ARG0=$0
+print_help() {
+    cat <<EOF
+Usage: $ARG0 COMMAND [TARGET] [ARGS...]
+  Supported TARGETs: i686 (default), x86_64, lagom
+  Supported COMMANDs:
+    build:      Compiles the target binaries, [ARGS...] are passed through to ninja
+    install:    Installs the target binary
+    image:      Creates a disk image with the installed binaries
+    run:        TARGET lagom [specific_test]: Runs the unit tests on the build host,
+                    or if specific_test is specified only this test.
+                All other TARGETs: $ARG0 run [TARGET] [kernel_cmd_line]
+                    Runs the built image in QEMU, and optionally passes the
+                    kernel_cmd_line to the Kernel
+    delete:     Removes the build environment for the TARGET
+    recreate:   Deletes and re-creates the build environment for TARGET
+    rebuild:    Deletes and re-creates the build environment, and compiles for TARGET
+    kaddr2line: $ARG0 kaddr2line TARGET ADDRESS
+                    Resolves the ADDRESS in the Kernel/Kernel binary to a file:line
+    addr2line:  $ARG0 addr2line TARGET BINARY_FILE ADDRESS
+                    Resolves the ADDRESS in BINARY_FILE to a file:line. It will
+                    attempt to find the BINARY_FILE in the appropriate build directory
+
+    rebuild-toolchain: Deletes and re-builds the TARGET's toolchain
+
+  Examples:
+    $ARG0 run i686 smp=on
+        Runs the image in QEMU passing "smp=on" to the kernel command line
+    $ARG0 run
+        Runs the image for the default TARGET i686 in QEMU
+    $ARG0 run lagom
+        Runs the unit tests on the build host
+    $ARG0 kaddr2line i686 0x12345678
+        Resolves the address 0x12345678 in the Kernel binary
+    $ARG0 addr2line i686 WindowServer 0x12345678
+        Resolves the address 0x12345678 in the WindowServer binary
+EOF
+}
+
+die() {
+    >&2 echo "die: $*"
+    exit 1
+}
+
+usage() {
+    >&2 print_help
+    exit 1
+}
+
+CMD=$1
+[ -n "$CMD" ] || usage
+shift
+if [ "$CMD" = "help" ]; then
+    print_help
+    exit 0
+fi
+
+if [ -n "$1" ]; then
+    TARGET="$1"; shift
+else
+    TARGET="i686"
+fi
+CMD_ARGS=( "$@" )
+CMAKE_ARGS=()
+
+get_top_dir() {
+    git rev-parse --show-toplevel
+}
+
+is_valid_target() {
+    if [ "$TARGET" = "lagom" ]; then
+        CMAKE_ARGS+=("-DBUILD_LAGOM=ON")
+        return 0
+    fi
+    [[ "$TARGET" =~ ^(i686|x86_64|lagom)$ ]] || return 1
+}
+
+create_build_dir() {
+    mkdir -p "$BUILD_DIR"
+    cmake -GNinja "${CMAKE_ARGS[@]}" -S . -B "$BUILD_DIR"
+}
+
+cmd_with_target() {
+    is_valid_target || ( >&2 echo "Unknown target: $TARGET"; usage )
+    SERENITY_ROOT="$(get_top_dir)"
+    export SERENITY_ROOT
+    BUILD_DIR="$SERENITY_ROOT/Build/$TARGET"
+    if [ "$TARGET" = "lagom" ]; then
+        TOOLCHAIN=i686
+    else
+        TOOLCHAIN="$TARGET"
+    fi
+    export SERENITY_ARCH="TOOLCHAIN"
+    TOOLCHAIN_DIR="$SERENITY_ROOT/Toolchain/Build/$TOOLCHAIN"
+}
+
+ensure_target() {
+    [ -d "$BUILD_DIR" ] || create_build_dir
+}
+
+run_tests() {
+    local TEST_NAME="$1"
+    export CTEST_OUTPUT_ON_FAILURE=1
+    if [ -n "$TEST_NAME" ]; then
+        ( cd "$BUILD_DIR" && ctest -R "$TEST_NAME" )
+    else
+        ( cd "$BUILD_DIR" && ctest )
+    fi
+}
+
+build_target() {
+    ninja -C "$BUILD_DIR" -- "$@"
+}
+
+delete_target() {
+    [ ! -d "$BUILD_DIR" ] || rm -rf "$BUILD_DIR"
+}
+
+build_toolchain() {
+    ( cd Toolchain && ARCH="$TOOLCHAIN" ./BuildIt.sh )
+}
+
+ensure_toolchain() {
+    [ -d "$TOOLCHAIN_DIR" ] || build_toolchain
+}
+
+delete_toolchain() {
+    [ ! -d "$TOOLCHAIN_DIR" ] || rm -rf "$TOOLCHAIN_DIR"
+}
+
+if [[ "$CMD" =~ ^(build|install|image|run|rebuild|recreate|kaddr2line|addr2line|setup-and-run)$ ]]; then
+    cmd_with_target
+    [[ "$CMD" != "recreate" && "$CMD" != "rebuild" ]] || delete_target
+    # FIXME: We should probably call ensure_toolchain first, but this somehow causes
+    # this error after the toolchain finished building:
+    # ninja: error: loading 'build.ninja': No such file or directory
+    ensure_target
+    ensure_toolchain
+    case "$CMD" in
+        build)
+            build_target "$@"
+            ;;
+        install)
+            [ "$TARGET" != "lagom" ] || die "Command '$CMD' not supported for the lagom target"
+            build_target
+            build_target install
+            ;;
+        image)
+            [ "$TARGET" != "lagom" ] || die "Command '$CMD' not supported for the lagom target"
+            build_target
+            build_target install
+            build_target image
+            ;;
+        run)
+            build_target
+            if [ "$TARGET" = "lagom" ]; then
+                run_tests "${CMD_ARGS[0]}"
+            else
+                build_target install
+                build_target image
+                if [ -n "${CMD_ARGS[0]}" ]; then
+                    export SERENITY_KERNEL_CMDLINE="${CMD_ARGS[0]}"
+                fi
+                build_target run
+            fi
+            ;;
+        rebuild)
+            build_target "$@"
+            ;;
+        recreate)
+            ;;
+        kaddr2line)
+            build_target
+            [ $# -ge 1 ] || usage
+            "$TOOLCHAIN_DIR/binutils/binutils/addr2line" -e "$BUILD_DIR/Kernel/Kernel" "$@"
+            ;;
+        addr2line)
+            build_target
+            [ $# -ge 2 ] || usage
+            BINARY_FILE="$1"; shift
+            BINARY_FILE_PATH="$BUILD_DIR/$BINARY_FILE"
+            if [ -x "$BINARY_FILE_PATH" ]; then
+                "$TOOLCHAIN_DIR/binutils/binutils/addr2line" -e "$BINARY_FILE_PATH" "$@"
+            else
+                find "$BUILD_DIR" -name "$BINARY_FILE" -executable -type f -exec "$TOOLCHAIN_DIR/binutils/binutils/addr2line" -e {} "$@" \;
+            fi
+            ;;
+        *)
+            build_target "$CMD" "$@"
+            ;;
+    esac
+elif [ "$CMD" = "delete" ]; then
+    cmd_with_target
+    delete_target
+elif [ "$CMD" = "rebuild-toolchain" ]; then
+    cmd_with_target
+    [ "$TARGET" != "lagom" ] || die "The lagom target uses the host toolchain"
+    delete_toolchain
+    ensure_toolchain
+else
+    >&2 echo "Unknown command: $CMD"
+    usage
+fi
--- a/Meta/debug-kernel.sh
+++ b/Meta/debug-kernel.sh
@@ -9,6 +9,6 @@
 # the debugger which binary to load symbols, etc from.
 #
 $SERENITY_KERNEL_DEBUGGER \
-    -ex "file $(dirname "$0")/../Build/Kernel/Kernel" \
+    -ex "file $(dirname "$0")/../Build/${SERENITY_ARCH:-i686}/Kernel/Kernel" \
     -ex 'set arch i386:intel' \
     -ex 'target remote localhost:1234'
--- a/Toolchain/BuildIt.sh
+++ b/Toolchain/BuildIt.sh
@@ -12,7 +12,7 @@ echo "$DIR"
 ARCH=${ARCH:-"i686"}
 TARGET="$ARCH-pc-serenity"
 PREFIX="$DIR/Local/$ARCH"
-BUILD="$DIR/../Build"
+BUILD="$DIR/../Build/$ARCH"
 SYSROOT="$BUILD/Root"
 
 MAKE="make"


Meta/serenity.sh
```diff
@@ -0,0 +1,190 @@
...
+    *)
+      build_target "$CMD" "$@"
+      ;;
+  esac
+elif [ $CMD = "delete" ]; then
+  cmd_with_target
+  delete_target
+elif [ $CMD = "rebuild-toolchain" ]; then
+  cmd_with_target
+  [ "$TARGET" != "lagom" ] || ( >&2 echo "The lagom target uses the host toolchain"; exit 1 )
```
emanuele6, line 184: Running `exit 1` inside `()` is basically the same thing as running `false`; this is probably what you want instead.

```suggestion
  [ "$TARGET" != "lagom" ] || { echo "The lagom target uses the host toolchain" >&2; exit 1 ;}
```

I think it would be better to just create a `die` helper function that prints an error message and exits like we do in all the other shell scripts though.

Also please indent with 4 spaces and define functions as `function_name() {\ncode\n}` since that's the code style we are already using for shell scripts.

emanuele6: I think it will also work as you wrote it, but only because of `set -e`.

Meta/serenity.sh
```diff
@@ -0,0 +1,190 @@
...
+{
+  >&2 echo "Command '$CMD' not supported for the lagom target"
+  exit 1
+}
+
+if [[ $CMD =~ ^(build|install|image|run|rebuild|recreate|addr2line|setup-and-run)$ ]]; then
+  cmd_with_target
+  [ $CMD != "recreate" -a $CMD != "rebuild" ] || delete_target
+  ensure_target
+  ensure_toolchain
```
BenWiederhake, line 135: If the toolchain doesn't exist yet, I think this will go wrong: Telling cmake to use a compiler that doesn't exist yet sounds like a bad idea.

Suggestion:
```
ensure_toolchain
ensure_target
```
Or does this cause other issues?

Meta/serenity.sh
```diff
@@ -0,0 +1,190 @@
...
+
+function get_top_dir()
+{
+  git rev-parse --show-toplevel
+}
+
+function is_valid_target()
+{
+  if [ $TARGET = "lagom" ]; then
+    echo 'target is lagom';
```
BenWiederhake, line 67: Is this a forgotten debug statement? :)

Meta/serenity.sh
```diff
@@ -0,0 +1,190 @@
...
+  ( cd "$BUILD_DIR" && cmake -GNinja "${CMAKE_ARGS[@]}" ../..)
+}
+
+function cmd_with_target()
+{
+  is_valid_target || ( >&2 echo "Unknown target: $TARGET"; usage )
+  export SERENITY_ARCH="$TARGET"
+  SERENITY_ROOT=$(get_top_dir)
+  export SERENITY_ROOT
+  BUILD_DIR="$SERENITY_ROOT/Build/$TARGET"
```
BenWiederhake, line 86: I like the new directory layout, but I think the Toolchain-builder has clashing assumptions built into it. Currently, `BuildIt.sh` contains the line:
```
BUILD="$DIR/../Build"
```

Maybe it is sufficient to change this to:
```
BUILD="$DIR/../Build/$ARCH"
```

I'm not sure whether this is all it takes, so it might be a good idea to run the commands locally, which ensures that your changes really work. :)

BenWiederhake (changes requested):
Awesome! And then there's the [shellcheck stuff](https://github.com/SerenityOS/serenity/pull/5686/checks?check_run_id=2051816486#step:10:25) :)

Meta/serenity.sh
```diff
@@ -0,0 +1,190 @@
...
+    build:     Compiles the target binaries, [ARGS...] are passed through to ninja
+    install:   Installs the target binary
+    image:     Creates a disk image with the installed binaries
+    run:       TARGET lagom:      Runs the unit tests on the build host
+               All other TARGETs: $ARG0 run [TARGET] [kernel_cmd_line]
+                   Runs the built image in QEMU, and optionally passed the
+    delete:    Removes the build environment for the TARGET
+    recreate:  Deletes and re-creates the build environment for TARGET
+    rebuild:   Deletes and re-creates the build environment, and compiles for TARGET
+    addr2line: $ARG0 addr2line TARGET BINARY_FILE ADDRESS
```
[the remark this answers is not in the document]
tomuta: I'm still hoping for a kernel cmd line flag to disable the randomization for debugging purposes ;) I used to use addr2line for user mode stuff routinely. But yeah a kaddr2line command sounds convenient!

Meta/serenity.sh
tomuta: I'll have to check it out. When I wrote my scripts I had to do it this way or the BuildIt.sh of the toolchain would run cmake for me or something weird like this. Maybe it's no longer necessary.

Meta/serenity.sh
tomuta: Indeed

Meta/serenity.sh
tomuta: Hm interesting. I like your suggestion.

Meta/serenity.sh
tomuta: Doesn't this apply to `/bin/false` either? Only works due to `set -e`? I'm not a shell expert, if `false` is the right thing to use I'm happy to change it.

tomuta:
Hm, how do you fix this?
```bash
  [ "$CMD" != "recreate" -a "$CMD" != "rebuild" ] || delete_target
                         ^-- SC2166: Prefer [ p ] && [ q ] as [ p -a q ] is not well defined.
```
I tried the suggestion, but then obviously A && B || C isn't great, either, but trying to nest it in additional `[ ... ]` doesn't work.

Meta/serenity.sh
emanuele6: No, the correct thing would be to use `{}` instead of `()`.
`()` runs commands inside a subshell and, if you run `exit 1` in the subshell, you will only exit (with exit status 1) from the subshell not the main script (hence equivalent to running `false`). Since `set -e` is activated and `(...; exit 1)` returns a non-0 exit status, the main script will exit anyway, but only because the script automatically kills itself when a command returns non-0 and `set -e` is on, not because `exit` actually made it exit.
`{}` just groups commands and runs them in the current shell (not in a subshell), that is probably what you want here.

Meta/serenity.sh
tomuta: Changing it to the more correct order somehow causes this error after building the toolchain:

```
ninja: error: loading 'build.ninja': No such file or directory
```

Not sure why the other order works, maybe it's picking up the wrong compiler indeed.

emanuele6:
@tomuta you either fix it using an if statement:
```sh
if [ "$CMD" = "recreate" ] || ["$CMD" = "rebuild" ]; then
    delete_target
fi
```

writing it like so using `{}`:
```sh
{ [ "$CMD" != "recreate" ] && [ "$CMD" != "rebuild" ] ;} || delete_target
```

Or, since this is a `bash` script, using bashisms:

```bash
[[ "$CMD" != "recreate" && "$CMD" != "rebuild" ]] || delete_target
```

tomuta:
Thanks. The last one looks the cleanest and is *almost* what I had, except that it works ;)

Meta/serenity.sh
tomuta: So it appears to work in the wrong order because it doesn't look like cmake is actually looking for the toolchain. When I have the toolchain built and manually run cmake after, I get this:
```
-- The ASM compiler identification is GNU
-- Found assembler: /usr/bin/cc
-- Check for working C compiler: /usr/bin/cc
-- Check for working C compiler: /usr/bin/cc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler: /usr/bin/c++
-- Check for working CXX compiler: /usr/bin/c++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
```
Why it doesn't work in the correct order, I'm not sure. I've been using the wrong (though working) order for a long, long time, so I think for now I'll just leave it as is.

Meta/serenity.sh
emanuele6: But I think it's better to add an helper `die` function like the one that can be found in `Meta/run.sh` (and other shell scripts):
```
die() {
    echo "die: $*"
    exit 1
}
```

And then use that:

```suggestion
  [ "$TARGET" != "lagom" ] || die 'The lagom target uses the host toolchain'
```

Meta/serenity.sh
```diff
@@ -0,0 +1,205 @@
...
+
+function ensure_target()
+{
+  [ -d "$BUILD_DIR" ] || create_build_dir
+}
+
+function build_target()
+{
+  ( cd "$BUILD_DIR" && ninja "$@" )
+}
```
emanuele6, line 108: This can be rewritten as:

```
build_target() {
    ninja -C "$BUILD_DIR" -- "$@"
}
```

Meta/serenity.sh
```diff
@@ -0,0 +1,213 @@
...
+function delete_toolchain()
+{
+  [ ! -d "$TOOLCHAIN_DIR" ] || rm -rf "$TOOLCHAIN_DIR"
+}
+
+function not_supported_lagom()
+{
+  >&2 echo "Command '$CMD' not supported for the lagom target"
+  exit 1
+}
```
emanuele6, line 139: This function can be replaced by `die` :^)

Meta/serenity.sh
emanuele6: https://github.com/SerenityOS/serenity/pull/5686#discussion_r589078701

> Also please indent with 4 spaces and define functions as `function_name() {\ncode\n}` since that's the code style we are already using for shell scripts.

@tomuta, I think you also missed this part. :^)

Meta/serenity.sh
```diff
@@ -0,0 +1,207 @@
...
+CMAKE_ARGS=()
+
+function get_top_dir()
+{
+  git rev-parse --show-toplevel
+}
+
+function is_valid_target()
+{
+  if [ $TARGET = "lagom" ]; then
```
emanuele6, line 77:
```suggestion
  if [ "$TARGET" = lagom ]; then
```

I don't understand why shellcheck isn't complaining about this.

You need to quote `$TARGET` here.

Try running this script with `./serenity.sh run 'a b'` and you will see that not quoting this variable causes problems; it will output:
```
./serenity.sh: line 77: [: too many arguments
Unknown target: a b
Usage: ./serenity.sh COMMAND [TARGET] [ARGS...]
  Supported TARGETs: i686 (default), x86_64, lagom
  Supported COMMANDs:
    build:      Compiles the target binaries, [ARGS...] are passed through to ninja
    install:    Installs the target binary
    image:      Creates a disk image with the installed binaries
    run:        TARGET lagom:      Runs the unit tests on the build host
                All other TARGETs: ./serenity.sh run [TARGET] [kernel_cmd_line]
                    Runs the built image in QEMU, and optionally passes the
                    kernel_cmd_line to the Kernel
    delete:     Removes the build environment for the TARGET
    recreate:   Deletes and re-creates the build environment for TARGET
    rebuild:    Deletes and re-creates the build environment, and compiles for TARGET
    kaddr2line: ./serenity.sh kaddr2line TARGET ADDRESS
                    Resolves the ADDRESS in the Kernel/Kernel binary to a file:line
    addr2line:  ./serenity.sh addr2line TARGET BINARY_FILE ADDRESS
                    Resolves the ADDRESS in BINARY_FILE to a file:line. It will
                    attempt to find the BINARY_FILE in the appropriate build directory

    rebuild-toolchain: Deletes and re-builds the TARGET's toolchain

  Examples:
    ./serenity.sh run i686 smp=on
        Runs the image in QEMU passing "smp=on" to the kernel command line
    ./serenity.sh run
        Runs the image for the default TARGET i686 in QEMU
    ./serenity.sh run lagom
        Runs the unit tests on the build host
    ./serenity.sh kaddr2line i686 0x12345678
        Resolves the address 0x12345678 in the Kernel binary
    ./serenity.sh addr2line i686 WindowServer 0x12345678
        Resolves the address 0x12345678 in the WindowServer binary

```

Meta/serenity.sh
```diff
@@ -0,0 +1,207 @@
...
+}
+
+function cmd_with_target()
+{
+  is_valid_target || ( >&2 echo "Unknown target: $TARGET"; usage )
+  export SERENITY_ARCH="$TARGET"
+  SERENITY_ROOT=$(get_top_dir)
+  export SERENITY_ROOT
+  BUILD_DIR="$SERENITY_ROOT/Build/$TARGET"
+  if [ $TARGET = "lagom" ]; then
```
emanuele6, line 97: same here

Meta/serenity.sh
```diff
@@ -0,0 +1,207 @@
...
+  # this error after the toolchain finished building:
+  # ninja: error: loading 'build.ninja': No such file or directory
+  ensure_target
+  ensure_toolchain
+  case "$CMD" in
+    build)
+      build_target "$@"
+      ;;
+    install)
+      [ $TARGET != "lagom" ] || die "Command '$CMD' not supported for the lagom target"
```
emanuele6, line 148: same here

Meta/serenity.sh
```diff
@@ -0,0 +1,207 @@
...
+      build_target
+      build_target install
+      ;;
+    image)
+      [ $TARGET != "lagom" ] || die "Command '$CMD' not supported for the lagom target"
```
emanuele6, line 153: same here

Meta/serenity.sh
```diff
@@ -0,0 +1,207 @@
...
+      build_target
+      build_target install
+      build_target image
+      ;;
+    run)
+      build_target
+      if [ $TARGET = "lagom" ]; then
```
emanuele6, line 160: same here

Meta/serenity.sh
tomuta: I did miss this. Now if github would only expand those collapsed resolved conversations when comments are added... ;)

emanuele6 (changes requested):

Meta/serenity.sh
emanuele6: If you run it as `./serenity.sh run '*'` it will get even messier.

tomuta:
Hm looks like some of the changes now causes the toolchain to be rebuilt every time when running `run lagom`...

Meta/serenity.sh
```diff
@@ -0,0 +1,195 @@
...
+    export SERENITY_ARCH="$TARGET"
+    SERENITY_ROOT=$(get_top_dir)
+    export SERENITY_ROOT
+    BUILD_DIR="$SERENITY_ROOT/Build/$TARGET"
+    if [ "$TARGET" = "lagom" ]; then
+        TOOLCHAIN=i686
+    else
+        TOOLCHAIN="$TARGET"
+    fi
+    TOOLCHAIN_DIR="$SERENITY_ROOT/Toolchain/Build/$TARGET"
```
emanuele6, line 96: Shouldn't this be `$SERENITY_ROOT/Toolchain/Build/$TOOLCHAIN` ?

tomuta: That would make too much sense.... Good catch, that's probably why it keeps building the toolchain over and over again.

Meta/serenity.sh
```diff
@@ -0,0 +1,195 @@
+#!/bin/env bash
+set -e
+
+ARG0=$0
+function print_help() {
```
emanuele6, line 5: You forgot to remove the `function` keywords

Meta/serenity.sh
```diff
@@ -0,0 +1,195 @@
...
+function create_build_dir() {
+    mkdir -p "$BUILD_DIR"
+    cmake -GNinja "${CMAKE_ARGS[@]}" -S . -B "$BUILD_DIR"
+}
+
+function cmd_with_target() {
+    is_valid_target || ( >&2 echo "Unknown target: $TARGET"; usage )
+    export SERENITY_ARCH="$TARGET"
+    SERENITY_ROOT=$(get_top_dir)
+    export SERENITY_ROOT
```
emanuele6, line 89:
```suggestion
    export SERENITY_ROOT=$(get_top_dir)
```

Meta/serenity.sh
```diff
@@ -0,0 +1,195 @@
...
+    >&2 echo "die: $*"
+    exit 1
+}
+
+function usage() {
+    >&2 print_help
+    exit 1
+}
+
+CMD=$1;
```
emanuele6, line 52:
```suggestion
CMD=$1
```

Meta/serenity.sh
tomuta: Unfortunately that didn't fix the problem...

tomuta:
Geez, that's why I don't like writing shell scripts ;)

Meta/serenity.sh
tomuta: Well maybe if I had fixed it correctly it would have... ;)

Meta/serenity.sh
tomuta: Linter doesn't like this:
```
    export SERENITY_ROOT=$(get_top_dir)
           ^-----------^ SC2155: Declare and assign separately to avoid masking return values.
```

tomuta:
> Hm looks like some of the changes now causes the toolchain to be rebuilt every time when running `run lagom`...

Should be resolved now.

PR #5688, opened 2021-03-07
LibC+LibM: Implement more math functionality

### LINKED ISSUE #4282: LibM: Missing proper implementation of frexp and other functions
I was having a play with the Lua port and noticed a commit that defines a bunch of math.h functions without implement them. Is there a plan to have a task force to implement all this functions? Many are pretty exoteric and my attempt to implement ``frexp`` left me with nothing that I can claim I wrote myself.

I was trying to run this simple Lua snippet https://rosettacode.org/wiki/Mandelbrot_set#ASCII_2 when I got a ASSERT_NOT_REACHED

RealKC:
Fixes #4282


Userland/Libraries/LibM/math.cpp
```diff
@@ -58,15 +61,36 @@ template<size_t value>
...
+    ToZero = FE_TOWARDZERO,
+    Up = FE_UPWARD,
+    Down = FE_DOWNWARD,
+    ToEven = FE_TONEAREST
 };
 
 template<typename T>
 union FloatExtractor;
 
+// This assumes long double is 80 bits, which should be the default with GCC
```
ADKaster, line 73: This assumption only true on intel platforms. On MIPS/SPARC/PowerPC/aarch64, it might be IEEE 754 Quad Double, which is 128 bits.

awesomekling: Indeed, let's wrap this assumption in `#if ARCH(I386) || ARCH(X86_64)`

danboid:
Yay! Thanks @RealKC

Userland/Libraries/LibM/math.cpp
```diff
@@ -81,6 +83,11 @@ union FloatExtractor<long double> {
         unsigned sign : 1;
     };
     long double d;
+
+    bool is_subnormal() const
+    {
+        return exponent == exponent_bias;
```
BenWiederhake, line 92: This (and the other implementations) seem wrong to me. Subnormals are numbers extremely close to zero, so the real exponent is as negative as it can be, e.g. 2^-308 or something. This means the biased exponent would be 0, so we should check for `0` instead of `exponent_bias`.

You might be interested in writing a test or three to make sure there are no further bugs lurking around. :)

Given that you only need it for `internal_scalbn` and that there are no plans anyway to make `FloatExtractor` accessible from `math.h`, I think it's nicer to not implement it four times, but rather once (namely in `internal_scalbn`).

RealKC:
Let's hope I didn't mess the commits up...

Userland/Libraries/LibM/math.cpp
RealKC: Should be done now

BenWiederhake (changes requested):
I spy with my little eye … could that be a bug? :)

Userland/Libraries/LibM/math.cpp
RealKC: Ah yes, I misread the wikipedia page on subnormals when implementing that.

Tests are a good idea, how should I go about writing them. Do I just pick numbers that should exercise different parts of the implementation, run them through my host's implementation, and then write tests expecting the values returned by my implementation to match/be close to the values returned by the host's implementation, or is there some better way?

BenWiederhake: Regarding how to physically implement it: In `Userland/Tests/LibM` and `Userland/Tests/LibC` you can already find a few things. Of course, you have lots of leeway regarding how exactly to implement it. The only real important thing is "exit code 0 if and only if it passes", and the output "FAIL" and "PASS" would be nice. (You probably also want more info for yourself, to understand what exactly went wrong if something goes wrong, but that's optional.)

Regarding which cases to test: Well, that's the magic sauce of testing, and noone can give you a definite answer. Your method certainly is better than no testing at all, but I wouldn't spend too many lines/effort on that. I would rather recommend writing one or two sanity checks ("Does 1+1 still equal 2?" caliber), and then writing up the corner cases that immediately come to your mind. If you want to know how that looks like when done *excessively*, take a glance here: https://github.com/SerenityOS/serenity/blob/master/AK/Tests/TestTime.cpp

It's totally up to you which corner cases you choose and how intensely you test each of them, but do include one test that would detect this bug.

RealKC: I've fixed the bug, and added some tests. Does everything look good to you?

BenWiederhake: Lovely! :)

BenWiederhake:
Fixes #4282.

RealKC:
I've added that to the PR description

PR #5690, opened 2021-03-08
AK: Take advantage of constexpr in Time and add time conversion methods

tomuta:
By making the Time constructor constexpr we can optimize creating a
Time instance from hardcoded values.

Also add more functions to convert between Time and various time units.

Since we tell the compiler to provide a default constructor we
need to initialize the member variables.

_I found myself missing these functions for some Time calculations_

### DIFF
--- a/AK/Time.h
+++ b/AK/Time.h
@@ -117,8 +117,8 @@ private:
 
     static Time from_half_sanitized(i64 seconds, i32 extra_seconds, u32 nanoseconds);
 
-    i64 m_seconds;
-    u32 m_nanoseconds; // Always less than 1'000'000'000
+    i64 m_seconds { 0 };
+    u32 m_nanoseconds { 0 }; // Always less than 1'000'000'000
 };
 
 template<typename TimevalType>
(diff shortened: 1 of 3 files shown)


AK/Time.h
```diff
@@ -84,9 +84,35 @@ class Time {
...
+    constexpr static Time from_seconds(i64 seconds) { return Time(seconds, 0); }
+    constexpr static Time from_nanoseconds(i64 nanoseconds)
+    {
+        return Time(nanoseconds / 1'000'000'000, nanoseconds % 1'000'000'000);
+    }
+    constexpr static Time from_microseconds(i64 microseconds)
+    {
+        return Time(microseconds / 1'000'000, (microseconds % 1'000'000) * 1'000);
+    }
+    constexpr static Time from_milliseconds(i64 milliseconds)
```
alimpfard, line 112: Perhaps a silly suggestion, but how about a uom-like interface for these?
`1ms`? it'd probably just call into `Time`.

tomuta: Go for it ;)

alimpfard: _/me adds it to the TODO list_

BenWiederhake:
There are several problems with this, see https://github.com/SerenityOS/serenity/commit/183b2e71ba8d85293db493cab27b8adb4af54981

PR #5696, opened 2021-03-08
Ports: Add libogg

AnicJov:
This library will allow us to port more multimedia programs and games,
since it's a common dependency of such projects. I was looking at
Scummvm's dependencies in particular, with a goal of porting them all
eventually, so that we can have Myst running in Serenity!

It's my first time trying to port something, so hopefully I did it somewhat correctly. 😅


devsh0:
Ogg is just the container though, not a codec standard. Pretty sure applications use an ogg-compatible codec with ogg.

AnicJov:
> Ogg is just the container though, not a codec standard. Pretty sure applications use an ogg-compatible codec with ogg.

Yes, and the plan is to port libFLAC and libtheora, for example, which require libogg.
My wording was inaccurate, libogg won't allow us to run games, but it's a root of a dependency tree that will.
Like I said, I was just going through Scummvm's dependencies, and got my feet wet porting this.

PR #5699, opened 2021-03-08
AK: Time over/underflow fixes

tomuta:


AK/Tests/TestTime.cpp
```diff
@@ -238,6 +238,11 @@ TEST_CASE(truncation)
     EXPECT_EQ(TIME(-2, -800'800'800).to_truncated_milliseconds(), -2'800);
     EXPECT_EQ(TIME(-2, -800'800'800).to_truncated_microseconds(), -2'800'800);
 
+    EXPECT_EQ(TIME(-9223372036854776, 193'000'000).to_truncated_milliseconds(), -0x7fff'ffff'ffff'ffff);
+    //EXPECT_EQ(TIME(-9223372036854776, 193'000'000).to_truncated_microseconds(), -0x7fff'ffff'ffff);
```
tomuta, line 242: @BenWiederhake not sure why this one doesn't pass, it's off by one. My brain needs a reboot ;)

AK/Time.h
```diff
@@ -151,6 +155,24 @@ class Time {
     {
     }
 
+    template<typename T>
+    ALWAYS_INLINE static constexpr i32 sane_mod(T& numerator, i32 denominator)
```
BenWiederhake, line 159: Making this a template means making overflow-bugs harder to see, because you're effectively reading two functions at the same time.

Case in point: `from_milliseconds(i64 milliseconds)` instantiates this with `i64`, and might call this with `(0x7fff'ffff'ffff'ffff, 1000)`. Then it ends up trying to write `0x20'c49b'a5e3'53f7` into `i32 dividend`, which overflows.

I *think* this can be fixed by making `dividend` and the return type `T`, especially since the callsites already assume a i64 return value.

Is there a way to say "Instantiate the template like this but only for these specific, hand-selected types"? Something like `static_assert(any_of<i32, i64>(T), "Fragile code ahead! Read it first, and make sure it is overflow-free for your new type before adding to the list.");`

AK/Time.cpp
```diff
@@ -106,49 +89,56 @@ i64 Time::to_truncated_seconds() const
...
+                // we only multiplied by one second less before to avoid an
+                // overflow that may not necessarily happen depending on add_ms
+                microseconds -= 1'000'000;
+            }
         }
+        if (!microseconds.has_overflow())
+            return microseconds.value();
     }
-    return m_seconds < 0 ? -0x8000'0000'0000'0000LL : 0x7fff'ffff'ffff'ffffLL;
+    return m_seconds < 0 ? -0x8000'0000'0000LL : 0x7fff'ffff'ffffLL;
```
BenWiederhake, line 141: This is too specific to be a typo, but too meaningless to be intentional. What happened here? :D

The correct values are `I64_MIN` and `I64_MAX`, respectively. No need to divide by 65536. Int saturation is int saturation.

AK/Tests/TestTime.cpp
```diff
@@ -238,6 +238,11 @@ TEST_CASE(truncation)
...
+    EXPECT_EQ(TIME(9223372036854776, 193'000'000).to_truncated_milliseconds(), 0x7fff'ffff'ffff'ffff);
```
BenWiederhake, line 243: It's rather uninteresting to check for the return value `0x7fff'ffff'ffff'ffff`, since any bug that causes one overflow too many will trigger this return value easily.

Let's also test this:
```
EXPECT_EQ(TIME(9223372036854776, 192'000'000).to_truncated_milliseconds(), 0x7fff'ffff'ffff'fffe);
EXPECT_EQ(TIME(9223372036854776, 193'000'000).to_truncated_milliseconds(), 0x7fff'ffff'ffff'ffff);
EXPECT_EQ(TIME(9223372036854776, 194'000'000).to_truncated_milliseconds(), 0x7fff'ffff'ffff'ffff);
```

… and now you know why many tests here occur in triplets :)
(One before the overflow, one that causes the extremal value as the correct result anyway, and one that overflows and causes the extremal value by saturation.)

BenWiederhake (changes requested):
:hugs: I'm glad I'm not the only one having trouble writing correct code around overflows. Here are some things that need improving :)

I feel justified in writing so many tests. Now imagine you have to deal with `m_nanoseconds` being an arbitrary `i32`, which happens when parsing a `struct timespec`! Gosh that took me a long time to come up with something that works.

BenWiederhake:
… and apparently clang wants you to declare the template before you use it. Sounds fair.

tomuta:
@BenWiederhake do you mind taking over this PR and fix it to your liking? Feel free to take these changes and fix them up, start over or whatever you prefer. I probably won't get to it in the next few days.

BenWiederhake:
Looks like I have to ^^

Issue #5698, opened 2021-03-08
Quotes page: Permission to be quoted


BenWiederhake:
Even though the IRC channel is publicly logged, I still wanna ensure there is consent before we start quoting people randomly out of context. Please refer to this issue when deciding whether someone can be quoted or not. In particular, this is preparation for our very own `fortunes` program.

People that already agreed on IRC:
- [X] BenW
- [X] CxByte
- [X] kling
- [X] linusg
- [X] nooga
- [X] thakis

And probably other people that I forgot to mention ^^

EDIT: You guys were too fast for me!

BenWiederhake:
@boricj I found a nice quote from you – and this time I will make extra double-sure to spell your name correctly! :D

boricj:
I'm OK with that, but I'll need to ramp up my quote game then...

tomuta:
Just for the record, any of my statements are subject to this project's BSD 2-clause license, so don't forget to reproduce that license notice.

_Just kidding._

danboid:
I'd like my glorious typos immortalised too.

Quaker762:
Lol, yeah sure

Tag @deoxxa

deoxxa:
I accept!

I'm not sure whose reputation is more at stake by me saying yes though - mine or Serenity's as a whole.

emanuele6:
I missed this; I also accept. :^)

bgianfo:
@BenWiederhake can this be closed?

BenWiederhake:
Sure. I still think it's reasonable to use this issue if you want to document that one is definitely okay with being quoted, but that can also be done while the issue is closed.

kleinesfilmroellchen:
Add me just in case

AtkinsSJ:
I'm fine being quoted, in case someone wants to.

PR #5701, opened 2021-03-08
Utilities: Add a quotes database and 'fortunes' program

BenWiederhake:
I told you, we need a quotes ~~page~~ file! :D

This even comes with fancy formatting and a tasteful selection ready to go! :^)
And look at this wonderful integration: It even links directly to the quote on logbot! (However, it can't be opened in Browser yet, probably a TLS thing is missing I guess.)

Bildschirmfoto_2021-03-08_23-05-13
Bildschirmfoto_2021-03-08_23-05-29

CC everyone in #5698, especially @linusg

Even though the IRC channel is publicly logged, I would still think it is nice to first ask before you quote someone out of context, so either check whether they have agreed to do so in #5698, or just ask them in person. Thanks!

### DIFF
--- a/Userland/Utilities/fortune.cpp
+++ b/Userland/Utilities/fortune.cpp
@@ -1,0 +1,156 @@
+/*
+ * Copyright (c) 2021, Ben Wiederhake <BenWiederhake.GitHub@gmx.de>
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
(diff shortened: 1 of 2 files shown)


Userland/Utilities/fortunes.cpp
```diff
@@ -0,0 +1,156 @@
+/*
+ * Copyright (c) 2018-2020, Andreas Kling <kling@serenityos.org>
```
linusg, line 2: you know... :wink:

<sub>\<kling\>I didn't write this!\</kling\></sub>

jcs:
IMHO, it should be called `fortune` to match the [historical Unix command](https://en.wikipedia.org/wiki/Fortune_(Unix)).

linusg (commented):
minor nitpick: can we call the `link` entry `url`?

BenWiederhake:
Changes since last push: ALL the things!

Userland/Utilities/fortune.cpp
```diff
@@ -0,0 +1,156 @@
...
+    if (quotes.is_empty()) {
+        warnln("{} does not contain any valid quotes", path);
+        return 1;
+    }
+
+    u32 i = arc4random_uniform(quotes.size());
+    const auto& chosen_quote = quotes[i];
+    auto datetime = Core::DateTime::from_timestamp(chosen_quote.utc_time());
+
+    outln(); // Tasteful spacing
```
jcs, line 141: While tasteful, this seems non-standard

BenWiederhake: Well, we don't try to implement POSIX fortune here; our storage format is also very non-standard. Is there any downside on the spacing? I find it makes it more readable, especially for larger quotes; see screenshots.

BenWiederhake:
The broken build seems to be due to a broken `Userland/Libraries/LibWeb/CSS/Parser/Token.h` on `master`.

Userland/Utilities/fortune.cpp
jcs: It just seems unnecessary and arbitrary, but you're the author so do what you prefer.

(It's only more readable because there's so much color being puked onto the terminal everywhere \*ducks\*)

BenWiederhake:
Changes since last push:
- Sneakily added a quote
- Rebased to build it on CI

@jcs: Thanks for the feedback! But I think I'll leave those extra newlines in for now, until someone implements `--fancy=off` or something. :)

PR #5702, opened 2021-03-09
LibWeb: Expose new CSS interfaces on the window object

Lubrsi:


awesomekling:
Oh right totally forgot about that. Thanks :D
Would be nice to auto-generate this entire thing.

PR #5712, opened 2021-03-09
LibCore+TextEditor: Introduce LibCore::MimeType, and use it to check when opening files in TextEditor

drunderscore:
This will prompt the user when opening a file in TextEditor that doesn't have a MIME type of `text`.

Currently, opening binary files in text editor will crash simply because the UTF8 parser doesn't fail gracefully. So at least the user will now know there's a good chance something will go wrong.

Some file types have very bad MIME types assigned to them (_cough_ `application/json` _cough_), and files without any MIME type assigned to them will also ask the user. If _may_ be too annoying, so we might want to only ask for files we know for certain aren't text .

### DIFF
--- a/Userland/Applications/TextEditor/TextEditorWidget.cpp
+++ b/Userland/Applications/TextEditor/TextEditorWidget.cpp
@@ -1,5 +1,6 @@
 /*
  * Copyright (c) 2018-2020, Andreas Kling <kling@serenityos.org>
+ * Copyright (c) 2021, James Puleo <james@jame.xyz>
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
@@ -34,6 +35,7 @@
 #include <LibCore/ConfigFile.h>
 #include <LibCore/File.h>
 #include <LibCore/MimeData.h>
+#include <LibCore/MimeType.h>
 #include <LibCpp/SyntaxHighlighter.h>
 #include <LibDesktop/Launcher.h>
 #include <LibGUI/Action.h>
@@ -584,6 +586,23 @@ void TextEditorWidget::update_title()
 
 void TextEditorWidget::open_file(const String& path)
 {
+    if (Core::File::exists(path)) {
+        auto guessed_mime = Core::guess_mime_type_based_on_filename(path);
+        if (!guessed_mime.has_value() || guessed_mime->type() != Core::MimeType::Type::text) {
+            StringBuilder error_builder;
+            if (!guessed_mime.has_value())
+                error_builder.append("This file does not have a MIME type, and so it may not be a text file. ");
+            else
+                error_builder.appendff("This file has the MIME type of '{}', and so it may not be a text file. ", guessed_mime->as_string());
+            error_builder.append("Are you sure you want to open this?");
+
+            if (GUI::MessageBox::show(window(), error_builder.to_string(), "Not a text file", GUI::MessageBox::Type::Warning,
+                    GUI::MessageBox::InputType::YesNo)
+                == GUI::Dialog::ExecResult::ExecNo)
+                return;
+        }
+    }
+
     auto file = Core::File::construct(path);
     if (!file->open(Core::IODevice::ReadOnly) && file->error() != ENOENT) {
         GUI::MessageBox::show(window(), String::formatted("Opening \"{}\" failed: {}", path, strerror(errno)), "Error", GUI::MessageBox::Type::Error);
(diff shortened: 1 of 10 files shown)


awesomekling:
Great idea! But let's put this in LibCore; AK is intended for things that are kernel/userspace agnostic :)

drunderscore:
@awesomekling Alright, I moved it into LibCore, and made the linter happy.

Userland/Libraries/LibCore/MimeData.cpp
```diff
@@ -70,33 +71,45 @@ void MimeData::set_text(const String& text)
...
     if (path.ends_with(".html", CaseSensitivity::CaseInsensitive) || path.ends_with(".htm", CaseSensitivity::CaseInsensitive))
-        return "text/html";
+        return Core::MimeType(Core::MimeType::m_type_text, "html");
     if (path.ends_with("/", CaseSensitivity::CaseInsensitive))
-        return "text/html";
+        return Core::MimeType(Core::MimeType::m_type_text, "html");
     if (path.ends_with(".csv", CaseSensitivity::CaseInsensitive))
-        return "text/csv";
-    return "text/plain";
+        return Core::MimeType(Core::MimeType::m_type_text, "csv");
```
alimpfard, line 100: It feels rather odd to have `m_*` as part the public API of this thing
Maybe put that in a `Type` struct and just name it `text`?

Userland/Libraries/LibCore/MimeType.h
```diff
@@ -0,0 +1,77 @@
...
+
+    static constexpr const char* m_type_text = "text";
+    static constexpr const char* m_type_image = "image";
+    static constexpr const char* m_type_audio = "audio";
+    static constexpr const char* m_type_video = "video";
+    static constexpr const char* m_type_application = "application";
+
+private:
+    MimeType(const String& mime, const StringView& type, const StringView& subtype)
+        : m_mime(move(mime))
```
awesomekling, line 59: The `move` here is ineffective since `mime` is a const reference.
```suggestion
    MimeType(String mime, const StringView& type, const StringView& subtype)
        : m_mime(move(mime))
```

Userland/Applications/TextEditor/main.cpp
```diff
@@ -89,7 +89,10 @@ int main(int argc, char** argv)
     app->set_menubar(menubar);
 
     if (file_to_edit)
-        text_widget.open_file(file_to_edit);
+    {
+        if(!text_widget.open_file(file_to_edit))
+            text_widget.update_title();
+    }
```
awesomekling: Please use `clang-format`

Userland/Libraries/LibCore/MimeData.cpp
awesomekling: Agreed. `Core::MimeType::m_type_text` => `Core::MimeType::Type::text`

awesomekling (changes requested):

drunderscore:
Everything suggested was implemented, as well as making the MIME checking only occur if the file actually exists.

Userland/Applications/TextEditor/TextEditorWidget.cpp
```diff
@@ -584,6 +586,23 @@ void TextEditorWidget::update_title()
...
+                error_builder.append("This file does not have a MIME type, and so it may not be a text file. ");
+            else
+                error_builder.appendff("This file has the MIME type of '{}', and so it may not be a text file. ", guessed_mime->as_string());
+            error_builder.append("Are you sure you want to open this?");
+
+            if (GUI::MessageBox::show(window(), error_builder.to_string(), "Not a text file", GUI::MessageBox::Type::Warning,
+                    GUI::MessageBox::InputType::YesNo)
+                == GUI::Dialog::ExecResult::ExecNo)
+                return;
+        }
```
awesomekling, line 603: This feels like we're being unnecessarily obnoxious to the user: they've asked us to open a file, and instead of doing that, we first get in their way and say "uh-huh, well are you suuure you want to open this file though?" even though opening a file a non-destructive operation.

Can we find a less obstructive approach? Perhaps a dismissable banner above the text editor widget with some text e.g "NOTE: This file has non-textual MIME type abc/def"

drunderscore: Yeah, I was worried it was a bit annoying, and you're right -- it's non-destructive so who cares if we crash!

I like the banner idea, unfortunately I'm not sure if I know how to implement that very well... I'll take a look notifications for some :sparkles:inspiration:sparkles:.

drunderscore:
Closing this for the time being -- I'm not sure how to implement what Andreas asked, and quite honestly, this is more of an issue pertaining to UTF-8 parsing not failing gracefully.

If you still would like the `MimeType` type, then I'll gladly cherry pick it out and have a separate PR for it.

PR #5713, opened 2021-03-10
Userland: Fix tar/unzip directory traversal vulnerability

### LINKED ISSUE #3991: tar: Directory traversal vulnerability may lead to privilege escalation
```
$ cat shellrc 
/bin/id
$ ./evilarc.py -f evil.tar.gz -o unix -p etc shellrc
Creating evil.tar.gz containing ../../../../../../../../etc/shellrc
```

lol

Most of the file system is mounted read-only. However, we can overwrite `/etc/shellrc` to gain privileges next time `root` uses `/bin/sh`.

### LINKED ISSUE #3992: unzip: Directory traversal vulnerability may lead to command execution / privilege escalation
Same issue as #3991 in `tar`.

lol

This could also be used to gain command execution on the host, or elevate privileges to the `anon` user from a lower privileged user, in the event that `anon` extracts a malicious tar file.

Command execution via `/etc/shellrc` (as `root`) or `/home/anon/.config/Terminal.ini` (as `anon`).

bblenard:
This change validates the filenames within a tar/zip archive during
extraction. If the filename within a the archive is outside of the
current working directory the file will be skipped and not extracted
onto the host system.

Closes #3991
Closes #3992

### DIFF
--- a/Userland/Utilities/unzip.cpp
+++ b/Userland/Utilities/unzip.cpp
@@ -24,12 +24,14 @@
  * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
+#include <AK/LexicalPath.h>
 #include <AK/MappedFile.h>
 #include <AK/NumberFormat.h>
 #include <LibCore/ArgsParser.h>
 #include <LibCore/File.h>
 #include <string.h>
 #include <sys/stat.h>
+#include <unistd.h>
 
 static const u8 central_directory_file_header_sig[] = "\x50\x4b\x01\x02";
 
@@ -62,7 +64,7 @@ static bool find_next_central_directory(off_t file_size, const MappedFile& file,
     return false;
 }
 
-static bool unpack_file_for_central_directory_index(off_t central_directory_index, const MappedFile& file)
+static bool unpack_file_for_central_directory_index(off_t central_directory_index, const MappedFile& file, const AK::String target_path)
 {
     enum CentralFileDirectoryHeaderOffsets {
         CFDHCompressionMethodOffset = 10,
@@ -120,6 +122,25 @@ static bool unpack_file_for_central_directory_index(off_t central_directory_inde
         return false;
     file_name[file_name_length] = '\0';
 
+    // determine where the archive intends to write a file
+    StringBuilder destination_path_raw;
+    if (file_name[0] != '/') {
+        destination_path_raw.append(target_path);
+        destination_path_raw.append("/");
+    }
+    destination_path_raw.append(file_name);
+    LexicalPath destination_path(destination_path_raw.to_string());
+    if (!(destination_path.string().starts_with(target_path))) {
+        fprintf(stderr, "File %s path is outside of current working directory, skipping\n", file_name);
+        return false;
+    }
+
+    // Ignore zips containing target path.
+    // This will prevent mkdir from failing later when it probably shouldn't
+    if (destination_path.string() == target_path) {
+        return true;
+    }
+
     if (file_name[file_name_length - 1] == '/') {
         if (mkdir(file_name, 0755) < 0) {
             perror("mkdir");
@@ -162,6 +183,29 @@ int main(int argc, char** argv)
     args_parser.add_positional_argument(path, "File to unzip", "path", Core::ArgsParser::Required::Yes);
     args_parser.parse(argc, argv);
 
+    StringBuilder target_path_raw;
+    // Currently unzip doesn't support extracting to a specified location. If that is implemented the target_path
+    // should be replaced with the target path from the command line arguements instead of "."
+    target_path_raw.append(realpath(".", nullptr));
+    target_path_raw.append("/");
+    LexicalPath target_path(target_path_raw.to_string());
+    auto target_path_string = target_path.string();
+    if (unveil(target_path_string.characters(), "rwc") < 0) {
+        perror("unveil");
+        return 1;
+    }
+
+    auto zip_path = realpath(path, nullptr);
+    if (unveil(zip_path, "r") < 0) {
+        perror("unveil");
+        return 1;
+    }
+
+    if (unveil(nullptr, nullptr) < 0) {
+        perror("unveil");
+        return 1;
+    }
+
     String zip_file_path { path };
 
     struct stat st;
@@ -192,7 +236,7 @@ int main(int argc, char** argv)
 
     off_t index = 0;
     while (find_next_central_directory(st.st_size, mapped_file, index, index)) {
-        bool success = unpack_file_for_central_directory_index(index, mapped_file);
+        bool success = unpack_file_for_central_directory_index(index, mapped_file, target_path_string);
         if (!success) {
             printf("Could not find local file header for a file.\n");
             return 4;
--- a/AK/LexicalPath.cpp
+++ b/AK/LexicalPath.cpp
@@ -46,6 +46,7 @@ void LexicalPath::canonicalize()
     }
 
     m_is_absolute = m_string[0] == '/';
+    m_is_dir = m_string.ends_with('/');
     auto parts = m_string.split_view('/');
 
     size_t approximate_canonical_length = 0;
@@ -104,6 +105,10 @@ void LexicalPath::canonicalize()
         if (m_is_absolute || i != 0)
             builder.append('/');
         builder.append(canonical_part);
+    }
+    if (m_is_dir) {
+        builder.append('/');
+        m_dirname = builder.to_string();
     }
     m_parts = move(canonical_parts);
     m_string = builder.to_string();
--- a/AK/LexicalPath.h
+++ b/AK/LexicalPath.h
@@ -63,6 +63,7 @@ private:
     String m_extension;
     bool m_is_valid { false };
     bool m_is_absolute { false };
+    bool m_is_dir { false };
 };
 
 template<>
--- a/Userland/Utilities/tar.cpp
+++ b/Userland/Utilities/tar.cpp
@@ -59,6 +59,18 @@ int main(int argc, char** argv)
     args_parser.add_positional_argument(paths, "Paths", "PATHS", Core::ArgsParser::Required::No);
     args_parser.parse(argc, argv);
 
+    // Currently tar doesn't support extracting to a specified location. If that is implemented the target path
+    // should be unveiled with rwc instead of the current directory.
+    StringBuilder target_path_raw;
+    target_path_raw.append(realpath(".", nullptr));
+    target_path_raw.append("/");
+    LexicalPath target_path(target_path_raw.to_string());
+    auto target_path_string = target_path.string();
+    if (unveil(target_path_string.characters(), "rwc") < 0) {
+        perror("unveil");
+        return 1;
+    }
+
     if (create + extract + list != 1) {
         warnln("exactly one of -c, -x, and -t can be used");
         return 1;
@@ -68,6 +80,16 @@ int main(int argc, char** argv)
         auto file = Core::File::standard_input();
 
         if (archive_file) {
+            // make sure we can read the archive path
+            auto archive_file_path = realpath(archive_file, nullptr);
+            if (unveil(archive_file_path, "r") < 0) {
+                perror("unveil");
+                return 1;
+            }
+            if (unveil(nullptr, nullptr) < 0) {
+                perror("unveil");
+                return 1;
+            }
             auto maybe_file = Core::File::open(archive_file, Core::IODevice::OpenMode::ReadOnly);
             if (maybe_file.is_error()) {
                 warnln("Core::File::open: {}", maybe_file.error());
@@ -86,14 +108,38 @@ int main(int argc, char** argv)
             warnln("the provided file is not a well-formatted ustar file");
             return 1;
         }
+
         for (; !tar_stream.finished(); tar_stream.advance()) {
+            auto file_name = tar_stream.header().file_name();
+
             if (list || verbose)
-                outln("{}", tar_stream.header().file_name());
+                outln("{}", file_name);
 
             if (extract) {
                 Tar::TarFileStream file_stream = tar_stream.file_contents();
 
                 const Tar::Header& header = tar_stream.header();
+
+                // determine where the archive intends to write a file
+                StringBuilder destination_path_raw;
+                if (file_name[0] != '/') {
+                    destination_path_raw.append(target_path_string);
+                    destination_path_raw.append("/");
+                }
+                destination_path_raw.append(file_name);
+                LexicalPath destination_path(destination_path_raw.to_string());
+
+                if (!(destination_path.string().starts_with(target_path_string))) {
+                    fprintf(stderr, "File %s path is outside of current working directory, skipping\n", file_name.to_string().characters());
+                    return false;
+                }
+
+                // Ignore tars containing target path.
+                // This will prevent mkdir from failing later when it probably shouldn't
+                if (destination_path.string() == target_path_string) {
+                    continue;
+                }
+
                 switch (header.type_flag()) {
                 case Tar::NormalFile:
                 case Tar::AlternateNormalFile: {


bcoles:
I had a quick read over the code, but haven't tested.

In the `tar` patch, it looks like there's no validation on the file name beyond this check:

```cpp
if (!file_name.starts_with(current_path.to_string()))
```

Wouldn't this still allow traversal, as a filename of `/my/current/path/../../../../../something/else` still technically starts with `/my/current/path` ?

The approach for `unzip` looks better; however, it uses `realpath`.

I started on a patch for this ages ago (see testing code below). The issue I ran into is that `realpath` only work on paths which exist, and some paths won't exist prior to extraction. IIRC, this was an issue when the compressed archive contained directories as the directory had not yet been created.

```diff
diff --git a/Userland/unzip.cpp b/Userland/unzip.cpp
index 2617c353d..bb67a83bc 100644
--- a/Userland/unzip.cpp
+++ b/Userland/unzip.cpp
@@ -28,9 +28,13 @@
 #include <AK/NumberFormat.h>
 #include <LibCore/ArgsParser.h>
 #include <LibCore/File.h>
+#include <limits.h>
+#include <stdlib.h>
 #include <string.h>
 #include <sys/stat.h>
+#include <unistd.h>
 
+bool verbose = false;
 static const u8 central_directory_file_header_sig[] = "\x50\x4b\x01\x02";
 
 static bool seek_and_read(u8* buffer, const MappedFile& file, off_t seek_to, size_t bytes_to_read)
@@ -120,31 +124,54 @@ static bool unpack_file_for_central_directory_index(off_t central_directory_inde
         return false;
     file_name[file_name_length] = '\0';
 
-    if (file_name[file_name_length - 1] == '/') {
-        if (mkdir(file_name, 0755) < 0) {
+    if (strlen(file_name) > PATH_MAX) {
+        fprintf(stderr, "Warning: skipped file path (length %lu exceeds %d): %s\n", strlen(file_name), PATH_MAX, file_name);
+        return false;
+    }
+
+    char output_directory[PATH_MAX];
+    // FIXME: change once output directory (-d) is supported
+    getcwd(output_directory, sizeof(output_directory));
+
+    char file_path[PATH_MAX];
+    //realpath(String::format("%s/%s", output_directory, file_name).characters(), file_path);
+    realpath(file_name, file_path);
+    printf("OUT: %s\n", output_directory);
+    printf("path: %s\n", file_path);
+
+    if (strncmp(output_directory, file_path, strlen(output_directory)) != 0) {
+        fprintf(stderr, "Warning: ignored path traversal: %s\n", file_path);
+        //fprintf(stderr, "Warning: ignored path traversal: %s\n", file_name);
+        return false;
+    }
+
+    if (verbose)
+        printf(" extracting: %s\n", file_path);
+
+    if (file_path[strlen(file_path) - 1] == '/') {
+        if (mkdir(file_path, 0755) < 0) {
             perror("mkdir");
             return false;
         }
     } else {
-        auto new_file = Core::File::construct(String { file_name });
+        auto new_file = Core::File::construct(String { file_path });
         if (!new_file->open(Core::IODevice::WriteOnly)) {
-            fprintf(stderr, "Can't write file %s: %s\n", file_name, new_file->error_string());
+            fprintf(stderr, "Can't write file %s: %s\n", file_path, new_file->error_string());
             return false;
         }
 
-        printf(" extracting: %s\n", file_name);
         u8 raw_file_contents[compressed_file_size];
-        if (!seek_and_read(raw_file_contents, file, local_file_header_index + LFHFileNameBaseOffset + file_name_length + extra_field_length, compressed_file_size))
+        if (!seek_and_read(raw_file_contents, file, local_file_header_index + LFHFileNameBaseOffset + strlen(file_path) + extra_field_length, compressed_file_size))
             return false;
 
         // FIXME: Try to uncompress data here. We're just ignoring it as no decompression methods are implemented yet.
         if (!new_file->write(raw_file_contents, compressed_file_size)) {
-            fprintf(stderr, "Can't write file contents in %s: %s\n", file_name, new_file->error_string());
+            fprintf(stderr, "Can't write file contents in %s: %s\n", file_path, new_file->error_string());
             return false;
         }
 
         if (!new_file->close()) {
-            fprintf(stderr, "Can't close file %s: %s\n", file_name, new_file->error_string());
+            fprintf(stderr, "Can't close file %s: %s\n", file_path, new_file->error_string());
             return false;
         }
     }
@@ -159,6 +186,7 @@ int main(int argc, char** argv)
 
     Core::ArgsParser args_parser;
     args_parser.add_option(map_size_limit, "Maximum chunk size to map", "map-size-limit", 0, "size");
+    args_parser.add_option(verbose, "Verbose output", "verbose", 'v');
     args_parser.add_positional_argument(path, "File to unzip", "path", Core::ArgsParser::Required::Yes);
     args_parser.parse(argc, argv);
 
```

bcoles:
> IIRC, this was an issue when the compressed archive contained directories as the directory had not yet been created.

Confirmed. I tested the `unzip` patch in this PR and ran into the same issue when the compressed zip file contains a directory, even when there's no traversal shenanigans at play.

unzip

```
user@ubuntu:~/Desktop/serenity$ unzip -l Base/home/anon/asdf.zip 
Archive:  Base/home/anon/asdf.zip
  Length      Date    Time    Name
---------  ---------- -----   ----
        0  2020-11-15 09:42   test/
       77  2020-11-15 09:42   test/Terminal.ini
---------                     -------
       77                     2 files
user@ubuntu:~/Desktop/serenity$ 
```

bblenard:
:facepalm: While cleaning up my code I must have accidentally removed the realpath call in tar. So yes you are right tar is still borked.

My testing wasn't good enough to catch the other issue you pointed out so thanks for mentioning it. I'll have to take this one back to the drawing board and see how to fix this.

bcoles:
> My testing wasn't good enough to catch the other issue you pointed out so thanks for mentioning it. I'll have to take this one back to the drawing board and see how to fix this.

Using `realpath` is good and the correct approach (better than trying to perform string matching on `..` or using `starts_with`, both of which are doomed to failure).

Unfortunately this issue is more complex than it appears on the surface. It [the unzip patch] definitely resolves the issue, but it does so by breaking existing functionality. I got about as far as you did before I decided it was someone else's problem - good luck :P

bblenard:
We might be in luck! I have to look this over but implementing this was going to be my next idea. Looks like its already been done though (maybe still have to review)
https://github.com/SerenityOS/serenity/blob/54f643659895ffbfc4b961f729ed5ed8f0cce37a/AK/LexicalPath.cpp#L41

bblenard:
@bcoles although I think using Lexical path could work I'm wondering if a cleaner solution would be to utilize `unveil` instead. If we just unveil our current working directory for `rwc` I would think unzip / tar should fail if there is a funky archive. I'll give that a try too but I'm curious on your thoughts?

bcoles:
> 
> 
> @bcoles although I think using Lexical path could work I'm wondering if a cleaner solution would be to utilize `unveil` instead. If we just unveil our current working directory for `rwc` I would think unzip / tar should fail if there is a funky archive. I'll give that a try too but I'm curious on your thoughts?

`unveil` is a good idea but I don't consider that the cleanest approach.

The paths should still be canonicalized appropriately. A combination would be best.

I forsee no immediate issues with using `unveil`. This will need to support writing to a user-specified output directory if one is provided (not sure if `tar` and `unzip` support these yet). There may also need to be support for temporary files created during extraction (maybe).

bblenard:
Hopefully that fixes the issue properly now. I wasn't able to verify with a non malicious zip because I couldn't figure out how to make one that passed `VERIFY(compression_method == None);` :facepalm: not sure what I was doing wrong but tested with a non malicious tar archive so I'm fairly confident. Below is what I did to test

serenity

bblenard:
Also I'm not 100% sold on the design of my code but wasn't sure how I would make it better so feel free to rip that apart :)

bcoles:
> I wasn't able to verify with a non malicious zip because I couldn't figure out how to make one

You'll need to create the ZIP on the host then copy it to the filesystem. Everything in `Base` is copied to the filesystem during `ninja image` as part of `build-root-filesystem.sh`.

```
mkdir test
echo 'hello friends' > test/hello
zip -r hello.zip test
me hello.zip $SERENITY_ROOT/Base/home/anon/hello.zip
```

bblenard:
> > I wasn't able to verify with a non malicious zip because I couldn't figure out how to make one
> 
> You'll need to create the ZIP on the host then copy it to the filesystem. Every directory in `Base` is copied to the filesystem during `ninja image` as part of `build-root-filesystem.sh`.
> 
> ```
> mkdir test
> echo 'hello friends' > test/hello
> zip -r hello.zip test
> mv hello.zip $SERENITY_ROOT/Base/home/anon/hello.zip
> ```

Thats good to know! I've been using a simple http server to download things with `pro`. That being said I tried to create a zip on my host with `zip -r....` but it `unzip` on serenity complained, which is strange since thats how you are creating the zips

bcoles:
> 
> 
> > > I wasn't able to verify with a non malicious zip because I couldn't figure out how to make one
> > 
> > 
> > You'll need to create the ZIP on the host then copy it to the filesystem. Every directory in `Base` is copied to the filesystem during `ninja image` as part of `build-root-filesystem.sh`.
> > ```
> > mkdir test
> > echo 'hello friends' > test/hello
> > zip -r hello.zip test
> > mv hello.zip $SERENITY_ROOT/Base/home/anon/hello.zip
> > ```
> 
> Thats good to know! I've been using a simple http server to download things with `pro`. That being said I tried to create a zip on my host with `zip -r....` but it `unzip` on serenity complained, which is strange since thats how you are creating the zips

What does "complained" mean ? I suggest that the complaint is a result of your changes to `unzip`.

unzip hello.zip

bblenard:
It happens on a clean master build too

clean-master

Userland/Utilities/unzip.cpp
```diff
@@ -162,6 +178,20 @@ int main(int argc, char** argv)
...
+    if (unveil(cwd, "rwc") < 0) {
+        fprintf(stderr, "CWD unveil\n");
+        perror("unveil");
+        return 1;
+    }
+
+    auto zip_path = realpath(path, nullptr);
+    if (unveil(zip_path, "r") < 0) {
+        fprintf(stderr, "PATH unveil\n");
+        perror("unveil");
```
bcoles, line 191: Using both `fprintf` and `perror` seems redundant. Simply `perror("unveil")` is sufficient.

Userland/Utilities/unzip.cpp
```diff
@@ -162,6 +178,20 @@ int main(int argc, char** argv)
...
+        return 1;
+    }
+
```
bcoles, line 194: The `unveil` state should be locked once all permitted paths have been set. Currently it is set but not locked.

```cpp
unveil(nullptr, nullptr);
```

Userland/Utilities/unzip.cpp
```diff
@@ -191,8 +221,14 @@ int main(int argc, char** argv)
     printf("Archive: %s\n", zip_file_path.characters());
 
     off_t index = 0;
+    StringBuilder target_path;
+    // Currently unzip doesn't support extracting to a specified location. If that is implemented the target_path
+    // should be replaced with the target path from the command line arguements instead of "."
+    target_path.append(realpath(".", nullptr));
+    target_path.append("/");
+    auto target_path_string = target_path.to_string();
```
bcoles, line 229: The `target_path` is effectively the same as `cwd` which is defined and unveiled earlier. `target_path` is a more future-proof name and it would make sense to group these together and use the same variable.

This way, if someone adds support for decompressing to a user-specified directory (ie, `unzip -d /output/path`) in the future, they only need to change one line as there's no hard coded assumption of operating in the working directory.

Userland/Utilities/unzip.cpp
```diff
@@ -162,6 +178,20 @@ int main(int argc, char** argv)
     args_parser.add_positional_argument(path, "File to unzip", "path", Core::ArgsParser::Required::Yes);
     args_parser.parse(argc, argv);
 
+    auto cwd = realpath(".", nullptr);
+    if (unveil(cwd, "rwc") < 0) {
+        fprintf(stderr, "CWD unveil\n");
+        perror("unveil");
```
bcoles, line 184: Using both `fprintf` and `perror` seems redundant. Simply `perror("unveil")` is sufficient.

bcoles:
For reference, this was [discussed on IRC](https://freenode.logbot.info/serenityos/20210312#c7224779) and the issue with decompression was resolved (and was unrelated to the patches in this PR).

bcoles (commented):
I had a look at the `unzip` patch only. My commentary may also apply to `tar`.

bcoles:
I tried the patched `tar` utility on an old `.tar.gz` file I had lying around (after first decompressing with `gunzip`) which results in some unexpected output.

tar -v -x -f php.tar

An archive containing `./` is certainly unusual, but the output stating that it is not within the target directory is also a lie. Perhaps there should be a hard coded rejection for `./` entries with an appropriate error message.

On the other hand, `tar` on Ubuntu silently ignores it. For reference, here's the output for Ubuntu:

```
user@ubuntu:~/test$ tar --list -f php.tar 
./
./README
./index.php
./lang/
./lang/english.php
./lang/german.php
./lang/polish.php
./lang/rus-1251.php
./lang/rus-koi8r.php
./themes/
./themes/black-green.php
./themes/black-orange.php
./themes/default.php
user@ubuntu:~/test$ tar xvf php.tar 
./
./README
tar: ./README: implausibly old time stamp -9223372036854775808
./index.php
tar: ./index.php: implausibly old time stamp -9223372036854775808
./lang/
./lang/english.php
tar: ./lang/english.php: implausibly old time stamp -9223372036854775808
./lang/german.php
tar: ./lang/german.php: implausibly old time stamp -9223372036854775808
./lang/polish.php
tar: ./lang/polish.php: implausibly old time stamp -9223372036854775808
./lang/rus-1251.php
tar: ./lang/rus-1251.php: implausibly old time stamp -9223372036854775808
./lang/rus-koi8r.php
tar: ./lang/rus-koi8r.php: implausibly old time stamp -9223372036854775808
./themes/
tar: ./lang: implausibly old time stamp -9223372036854775808
./themes/black-green.php
tar: ./themes/black-green.php: implausibly old time stamp -9223372036854775808
./themes/black-orange.php
tar: ./themes/black-orange.php: implausibly old time stamp -9223372036854775808
./themes/default.php
tar: ./themes/default.php: implausibly old time stamp -9223372036854775808
tar: ./themes: implausibly old time stamp -9223372036854775808
tar: .: implausibly old time stamp -9223372036854775808
user@ubuntu:~/test$ ls -la
total 96
drwxr-xr-x  4 user user  4096 Dec 13  1901 .
drwxr-xr-x 21 user user  4096 Mar 12 06:07 ..
-rw-r--r--  1 user user 24508 Dec 13  1901 index.php
drwxr-xr-x  2 user user  4096 Dec 13  1901 lang
-rw-r--r--  1 user user 51200 Mar 12 05:57 php.tar
-rw-r--r--  1 user user  1300 Dec 13  1901 README
drwxr-xr-x  2 user user  4096 Dec 13  1901 themes
user@ubuntu:~/test$ 
```

On the other other hand, perhaps this is outside the scope of this PR.

bblenard:
I think its in scope. The issue you saw was from some sloppy programming on my part. The issue happens because I use the following check to determine if a file is in the target path

```
if (!destination_path.string().starts_with(target_path_string)) {
```

The reason you see this error is because for `./` the destination path ends up being `/path/to/some/folder` and the `target_path_string` is `/path/to/some/folder/`. I intentionally add the `/` to the target path because I wanted a way to ensure you were extracting under the target directory and not to a path that shares the same prefix but I don't know if that is a realistic thing to worry about.

bcoles:
> The reason you see this error is because for `./` the destination path ends up being `/path/to/some/folder` and the `target_path_string` is `/path/to/some/folder/`. I intentionally add the `/` to the target path because I wanted a way to ensure you were extracting under the target directory and not to a path that shares the same prefix but I don't know if that is a realistic thing to worry about.

It is a realistic thing to worry about, but you should be able to use the canonicalized path for the `starts_with` comparison, ensuring `/some/path/./ == /some/path/`.

Userland/Utilities/unzip.cpp
bblenard: Good call :+1:

bblenard:
Hopefully I'm getting close with this now. I did make an ease of use change to LexicalPath. If that isn't an acceptable change I would love to hear some ideas on how to handle archive file paths that share prefixes with the target path but are not the same directory. For example

`archive file path`: `/some/path_right_here`
`target file path`:`/some/path`

serenity

linusg:
@bblenard there are two merge conflicts, please rebase this PR :)

PR #5714, opened 2021-03-10
QuoteServer: Add a simple QOTD service

bcoles:
Add a simple [QOTD](https://en.wikipedia.org/wiki/QOTD) service. Much of the code (including the `Quote` class) was stolen from BenW (#5701). This class could be abstracted away into a library and even integrated into `ProtocolServer`; however, this feels like overkill for such a simple protocol/service/feature.

QOTD

The [QOTD RFC](https://tools.ietf.org/html/rfc865) imposes very few limitations on the protocol implementation beyond defining the default port (17).

The only real limitation on the quote format relates to the use of ASCII and quote length, both of which are ignored in this implementation.

```
Quote Syntax

   There is no specific syntax for the quote.  It is recommended that it
   be limited to the ASCII printing characters, space, carriage return,
   and line feed.  The quote may be just one or up to several lines, but
   it should be less than 512 characters.
```


tomuta:
This seems... way overkill? Another unnecessary attack vector?

bcoles:
> 
> 
> This seems... way overkill? Another unnecessary attack vector?

Like `TelnetServer` ? `QuoteServer` does not run by default.

awesomekling:
TelnetServer provides something very useful (remote login) that nothing else in the base system offers yet.

Let's be honest: this is bloat. The system is growing fast enough as it is, and we should focus on adding things that one would actually use, instead of doing things just because they are possible. :)

PR #5717, opened 2021-03-10
Documentation: Add Serenity installation guide for VMware

bcoles:


diversys:
It would be nice to be able to build .vmdk images directly. Haiku has a tool to do exactly that https://git.haiku-os.org/haiku/log/?qt=grep&q=vmdk

bcoles:
> 
> 
> It would be nice to be able to build .vmdk images directly. Haiku has a tool to do exactly that https://git.haiku-os.org/haiku/log/?qt=grep&q=vmdk

`raw2vmdk` and `dd2vmdk` also

PR #5720, opened 2021-03-10
Mixed small fixes (memleak, typo'd logic, unused stuff)

BenWiederhake:
- The xargs memory leak was detected by UserlandEmulator, after applying #5719. :partying_face: 
- No idea what the `kmalloc.h` includes where doing in Userland programs, but apparently they weren't necessary.
- I ran into the `crash(1)` problems while experimenting around with it and UE.
- The unused `config.host.orig` file came up when trying to investigate something else entirely.

I'll keep this as draft until the CI passes, to make extra-double-triple-sure that the `config.host.orig` really isn't necessary.


BenWiederhake:
Build passed, this can be merged now.

Issue #5710, opened 2021-03-09
Meta: serenity.sh should work with ports


linusg:
Currently many ports expect the build root to be `${SERENITY_ROOT}/Build`. When using the new `serenity.sh` script (:heart: it), the build root is something like `${SERENITY_ROOT}/Build/i686`, and installing ports won't work as expected.

Possible options:

- use something like a `SERENITY_BUILD_ROOT` environment variable in ports instead of hardcoded paths
- use `${SERENITY_ROOT}/Build/${SERENITY_ARCH}` as the new default location everywhere and update docs + port scripts

PR #5726, opened 2021-03-11
Change Toolchain/BuildIt.sh BUILD dir

basile-henry:
The ports seem to rely on this location for the toolchain's `BUILD`.

More specifically the `zlib` port wouldn't compile with the other location for `BUILD`. `ld` was complaining that it couldn't find `crt0_shared.o`.

I git bisected the issue back to 389dddd4b372b3d29e1629a97315e7432e3f04f1, and this is my attempt at fixing it. @tomuta Was changing `BUILD` necessary for something else? I don't want to break the new `Meta/serenity.sh`.

### DIFF
--- a/Toolchain/BuildIt.sh
+++ b/Toolchain/BuildIt.sh
@@ -12,7 +12,7 @@ echo "$DIR"
 ARCH=${ARCH:-"i686"}
 TARGET="$ARCH-pc-serenity"
 PREFIX="$DIR/Local/$ARCH"
-BUILD="$DIR/../Build/$ARCH"
+BUILD="$DIR/../Build"
 SYSROOT="$BUILD/Root"
 
 MAKE="make"


BenWiederhake:
> Was changing BUILD necessary for something else? I don't want to break the new Meta/serenity.sh.

The idea is to split up the various targets into separate build directories. This has to be done at some point anyway, if we ever want to support x86_64, so `serenity.sh` was written with the hardcoded assumption of a new layout. `BuildIt.sh` now also follows this new layout.

You're right that it breaks Ports, and there [is already an issue for that](https://github.com/SerenityOS/serenity/issues/5710). But I think the way forward is to fix `Ports/.include_package.sh`, and not `BuildIt.sh`.

basile-henry:
Alright, that makes sense. I missed that issue. I'll see if I can change `Ports/.port_include.sh` to take that change into account :+1:

PR #5732, opened 2021-03-11
LibGUI+NotificationServer: Support mutable notifications

Sylvyrfysh:
This commit aims to add a way to interact with notifications once they have been published. This switches notifications to use a lifetime connection that handles a single notification.

The new API returns a boolean as to if the notification window is still open.

This is meant for applications like chat clients or email clients that could send a notification about a message, but may wish to update the text, title, or image if a second message is received. I'd like to eventually add the option to focus an application and send some data upon a click, but figured I'd get the hang of Serenity by starting here.

### DIFF
--- a/Userland/Libraries/LibGUI/Notification.cpp
+++ b/Userland/Libraries/LibGUI/Notification.cpp
@@ -34,23 +34,28 @@ namespace GUI {
 class NotificationServerConnection : public IPC::ServerConnection<NotificationClientEndpoint, NotificationServerEndpoint>
     , public NotificationClientEndpoint {
     C_OBJECT(NotificationServerConnection)
+
+    friend class Notification;
+
 public:
     virtual void handshake() override
     {
         send_sync<Messages::NotificationServer::Greet>();
     }
 
-    virtual void die() override { m_connected = false; }
-
-    bool is_connected() const { return m_connected; }
+    virtual void die() override
+    {
+        m_notification->connection_closed();
+    }
 
 private:
-    NotificationServerConnection()
+    explicit NotificationServerConnection(Notification* notification)
         : IPC::ServerConnection<NotificationClientEndpoint, NotificationServerEndpoint>(*this, "/tmp/portal/notify")
+        , m_notification(notification)
     {
     }
     virtual void handle(const Messages::NotificationClient::Dummy&) override { }
-    bool m_connected { true };
+    Notification* m_notification;
 };
 
 Notification::Notification()
@@ -63,25 +68,27 @@ Notification::~Notification()
 
 void Notification::show()
 {
-    VERIFY(!m_connection);
+    VERIFY(!m_shown && !m_destroyed);
     auto icon = m_icon ? m_icon->to_shareable_bitmap() : Gfx::ShareableBitmap();
-    m_connection = NotificationServerConnection::construct();
+    m_connection = NotificationServerConnection::construct(this);
     m_connection->send_sync<Messages::NotificationServer::ShowNotification>(m_text, m_title, icon);
+    m_shown = true;
 }
 
 void Notification::close()
 {
-    VERIFY(m_connection);
-    if (!m_connection->is_connected()) {
+    VERIFY(m_shown);
+    if (!m_destroyed) {
+        m_connection->send_sync<Messages::NotificationServer::CloseNotification>();
+        connection_closed();
         return;
     }
-    m_connection->send_sync<Messages::NotificationServer::CloseNotification>();
 }
 
 bool Notification::update()
 {
-    VERIFY(m_connection);
-    if (!m_connection->is_connected()) {
+    VERIFY(m_shown);
+    if (m_destroyed) {
         return false;
     }
 
@@ -97,6 +104,12 @@ bool Notification::update()
     }
 
     return true;
+}
+
+void Notification::connection_closed()
+{
+    m_connection.clear();
+    m_destroyed = true;
 }
 
 }
--- a/Userland/Libraries/LibGUI/Notification.cpp
+++ b/Userland/Libraries/LibGUI/Notification.cpp
@@ -49,6 +49,7 @@ private:
 };
 
 Notification::Notification()
+    : m_connection(NotificationServerConnection::construct())
 {
 }
 
@@ -58,9 +59,8 @@ Notification::~Notification()
 
 void Notification::show()
 {
-    auto connection = NotificationServerConnection::construct();
     auto icon = m_icon ? m_icon->to_shareable_bitmap() : Gfx::ShareableBitmap();
-    connection->send_sync<Messages::NotificationServer::ShowNotification>(m_text, m_title, icon);
+    m_connection->send_sync<Messages::NotificationServer::ShowNotification>(m_text, m_title, icon);
 }
 
 }
--- a/Userland/Libraries/LibGUI/Notification.cpp
+++ b/Userland/Libraries/LibGUI/Notification.cpp
@@ -40,12 +40,17 @@ public:
         send_sync<Messages::NotificationServer::Greet>();
     }
 
+    virtual void die() override { m_connected = false; }
+
+    bool is_connected() const { return m_connected; }
+
 private:
     NotificationServerConnection()
         : IPC::ServerConnection<NotificationClientEndpoint, NotificationServerEndpoint>(*this, "/tmp/portal/notify")
     {
     }
     virtual void handle(const Messages::NotificationClient::Dummy&) override { }
+    bool m_connected { true };
 };
 
 Notification::Notification()
@@ -60,15 +65,26 @@ Notification::~Notification()
 void Notification::show()
 {
     VERIFY(!m_showing);
+    VERIFY(!m_disposed);
+    if (!m_connection->is_connected()) {
+        // This would imply that the NotificationServer crashed before we could send it any data.
+        VERIFY_NOT_REACHED();
+    }
     auto icon = m_icon ? m_icon->to_shareable_bitmap() : Gfx::ShareableBitmap();
     m_connection->send_sync<Messages::NotificationServer::ShowNotification>(m_text, m_title, icon);
     m_showing = true;
 }
+
 void Notification::close()
 {
     VERIFY(m_showing);
-    m_connection->send_sync<Messages::NotificationServer::CloseNotification>();
+    if (m_connection->is_connected()) {
+        m_connection->send_sync<Messages::NotificationServer::CloseNotification>();
+    }
+
     m_showing = false;
+    m_disposed = true;
+}
 }
 
 }
--- a/Userland/Libraries/LibGUI/Notification.cpp
+++ b/Userland/Libraries/LibGUI/Notification.cpp
@@ -54,7 +54,6 @@ private:
 };
 
 Notification::Notification()
-    : m_connection(NotificationServerConnection::construct())
 {
 }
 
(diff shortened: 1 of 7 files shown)


Userland/Libraries/LibGUI/Notification.cpp
```diff
@@ -58,9 +58,56 @@ Notification::~Notification()
...
+    m_showing = true;
+}
+
+bool Notification::update()
+{
+    VERIFY(m_id != -1 && m_showing);
+
+    bool is_checked = false;
+
+    auto connection = NotificationServerConnection::construct();
```
BenWiederhake, line 81: So you create a new connection for every single update? How does the NotificationServer check whether updates come from the same sender? If it doesn't, then other (malfunctioning/malicious) programs might interfere with the notification. That doesn't sound like a good idea.

Userland/Services/NotificationServer/ClientConnection.cpp
```diff
@@ -57,7 +57,41 @@ OwnPtr<Messages::NotificationServer::ShowNotificationResponse> ClientConnection:
...
+}
+
+OwnPtr<Messages::NotificationServer::RemoveNotificationResponse> ClientConnection::handle(const Messages::NotificationServer::RemoveNotification& message)
+{
+    auto window = NotificationWindow::get_window_by_id(message.id());
+    if (window) {
+        window->close();
+    }
+    return make<Messages::NotificationServer::RemoveNotificationResponse>();
+}
```
BenWiederhake, line 89: Good idea! Can you add a way in `GUI::Notification` to call this? I currently don't see anything that calls this.

Userland/Services/NotificationServer/NotificationServer.ipc
```diff
@@ -3,5 +3,13 @@ endpoint NotificationServer = 95
...
-    ShowNotification([UTF8] String text, [UTF8] String title, Gfx::ShareableBitmap icon) => ()
+    ShowNotification([UTF8] String text, [UTF8] String title, Gfx::ShareableBitmap icon) => (i64 id)
+
+    UpdateNotificationText(i64 id, [UTF8] String text, [UTF8] String title) => (bool still_showing)
+
+    UpdateNotificationIcon(i64 id, Gfx::ShareableBitmap icon) => (bool still_showing)
+
+    RemoveNotification(i64 id) => ()
+
+    IsShowing(i32 id) => (bool still_showing)
```
BenWiederhake, line 14: All other RPCs use `i64`. Typo?

Userland/Services/NotificationServer/NotificationWindow.cpp
```diff
@@ -60,14 +62,28 @@ void update_notification_window_locations()
 
 NotificationWindow::NotificationWindow(const String& text, const String& title, const Gfx::ShareableBitmap& icon)
 {
-    s_windows.append(this);
+    while (true) {
+        m_window_id = get_random<i64>();
+        if (m_window_id < 0) {
+            if (m_window_id == (i64)0xFFFFFFFFFFFFFFFF) {
```
BenWiederhake, line 68: `(i64)0xFFFFFFFFFFFFFFFF` is the same as `-1`.

Userland/Services/NotificationServer/NotificationWindow.cpp
```diff
@@ -60,14 +62,28 @@ void update_notification_window_locations()
...
+                m_window_id = 0;
+            } else {
+                m_window_id = -m_window_id;
```
BenWiederhake, line 71: Hmm? Why not use `u64` directly? It seems like the only place that uses `-1` as a special value is `GUI::Notification`, and there it is redundant to `m_showing`. Allowing `-1` (or switching to `u64`) would get rid of all the special cases.

Userland/Services/NotificationServer/NotificationWindow.cpp
```diff
@@ -60,14 +62,28 @@ void update_notification_window_locations()
...
+    }
+    s_windows.set(m_window_id, this);
 
     set_window_type(GUI::WindowType::Notification);
     set_resizable(false);
     set_minimizable(false);
 
     Gfx::IntRect lowest_notification_rect_on_screen;
-    for (auto& window : s_windows) {
+    for (auto& window_ent : s_windows) {
```
BenWiederhake, line 85: Nit: `entry`, to avoid contusion with `entity`? Or maybe I just read LibC for too long ^^'

Userland/Services/NotificationServer/NotificationWindow.cpp
```diff
@@ -92,36 +108,59 @@ NotificationWindow::NotificationWindow(const String& text, const String& title,
...
+    on_close_request = [&] {
+        s_windows.remove(m_window_id);
         update_notification_window_locations();
         return CloseRequestDecision::Close;
     };
 }
 
 NotificationWindow::~NotificationWindow()
 {
+    s_windows.remove(m_window_id);
```
BenWiederhake, line 141: Does this really make sense? `s_windows` has a RefPtr to `this`. So `~NotificationWindow` can only be called if `this` was already removed from `s_windows`, or has never been added for some reason. Did I miss something, or can this be removed?

Userland/Services/NotificationServer/ClientConnection.cpp
```diff
@@ -57,7 +57,41 @@ OwnPtr<Messages::NotificationServer::ShowNotificationResponse> ClientConnection:
...
+        window->set_title_widget(message.title());
+    }
+    return make<Messages::NotificationServer::UpdateNotificationTextResponse>(window);
+}
+
+OwnPtr<Messages::NotificationServer::RemoveNotificationResponse> ClientConnection::handle(const Messages::NotificationServer::RemoveNotification& message)
+{
+    auto window = NotificationWindow::get_window_by_id(message.id());
+    if (window) {
+        window->close();
```
BenWiederhake, line 86: I think this leaks memory by not removing the window from `s_windows`:
- `window->close()` is only implemented by `GUI::Window::close`
- `GUI::Window::on_close` is never set, and does not exist by default (and will not call `on_close_request`)

Is that the case? If so, maybe changing `NotificationWindow` to hook `on_close` instead of `on_close_request` makes sense?

Userland/Services/NotificationServer/NotificationWindow.cpp
```diff
@@ -92,36 +108,59 @@ NotificationWindow::NotificationWindow(const String& text, const String& title,
...
+}
+
+void NotificationWindow::set_title_widget(const String& text)
+{
+    m_title_label->set_text(text);
+}
+
+void NotificationWindow::set_image(const Gfx::ShareableBitmap& icon)
+{
+    m_image->set_bitmap(icon.bitmap());
```
BenWiederhake, line 163: This will crash if the original notification didn't have an image (see `if (icon.is_valid()) {`), because in that case `m_image` is `nullptr`.

BenWiederhake (changes requested):
Hell yeah, mutable notifications! :D

Please fix the crash, the memory leak, and the type-typo ("type-o"?) first. Also, I hope I could convince you to switch away from making `-1` a special case.

Also: Please take another look at the [contributing guidelines](https://github.com/SerenityOS/serenity/blob/master/CONTRIBUTING.md#code-submission-policy), in particular the thing about squashing your iterations into easily-readable commits.

And finally: The commit body is great! Could you improve the commit title a bit? After all, the notification feature already existed previously, you didn't add it. I would probably write something like `LibGUI+NotificationServer: Support mutable notifications`, because that's the actual feature you're adding. (Which is an awesome feature!)

Userland/Libraries/LibGUI/Notification.cpp
Sylvyrfysh: That was the idea behind the 63-bit notification ID's. I toyed with the idea of a continuous connection, but decided to try to keep as much code the same as possible. If you think it'd be better, we can go to a continuous connection model.

Userland/Services/NotificationServer/NotificationWindow.cpp
Sylvyrfysh: You're correct. I'm used to a custom c++03 implementation of smart pointers in my company, where circular references are automatically handled.

Userland/Libraries/LibGUI/Notification.cpp
BenWiederhake: I'm not sure how to fix this.
- One might think about keeping the connection open, and converting `s_windows` to a member of each connection, somehow, dunno whether that's possible. The client would need to keep a connection to the NotificationServer open, which sounds like a good idea anyway.
- Another way is to have some form of authentication: In order to modify a notification, the client needs to send some kind of "secret" that proves they're still the same.
- Or you say that the ID is randomized for precisely that purpose. After all, guessing a "`u63`" correctly is pretty unlikely to happen accidentally.
- Or something else entirely?

BenWiederhake: Oh, sorry, I didn't see your reply! :D

I think the continuous connection is the way to go in the long run, but the more I think about it the more effort this seems to be. So I guess I'm fine with the ID-as-nonce idea. Could you add a TODO in the `.ipc` about that? After all, abusing IDs to also do verification seems documentation-worthy, and maybe this is a good place. What do you think?

Userland/Services/NotificationServer/NotificationWindow.cpp
BenWiederhake: confusion* – Oh the irony of making a typo in a comment that's complaining about spelling. XD

Userland/Services/NotificationServer/NotificationWindow.cpp
BenWiederhake: Hmm, that's an interesting philosophical question. But I think here it's not even a circular reference, because `s_windows` isn't under a `RefPtr`. :P

Userland/Libraries/LibGUI/Notification.cpp
Sylvyrfysh: I'll just rework this to use a continuous model I think. I played around with it and it removes a lot of smells, is easy to implement, and also lowers the number of required function calls. I'll plan to go this way.

Sylvyrfysh: The biggest issue that this has is that ClientConnection.h is autogenerated, and I cannot store a window in that as that variable would be removed upon regeneration. I will have to keep the static hashmap.

BenWiederhake: That sounds good.

Alternatively, in theory, if you really wanna do it: You could also extend the IPC description format, to allow injecting a member somehow. No idea whether that's a good idea or how that would look :D

Userland/Services/NotificationServer/NotificationWindow.cpp
Sylvyrfysh: Changed to the client id with the lifetime connection, should never rollover (unless you wanna make 4 billion notifications) now :)

Sylvyrfysh:
P.S.: Not sure if I went overboard on the commit separation. Let me know if I went too far!

Userland/Services/NotificationServer/NotificationWindow.cpp
```diff
@@ -95,23 +96,21 @@ NotificationWindow::NotificationWindow(i32 client_id, const String& text, const
     widget.layout()->set_margins({ 8, 8, 8, 8 });
     widget.layout()->set_spacing(6);
 
-    if (icon.is_valid()) {
-        auto& image = widget.add<GUI::ImageWidget>();
-        image.set_bitmap(icon.bitmap());
-    }
+    m_image = &widget.add<GUI::ImageWidget>();
+    m_image->set_bitmap(icon.is_valid() ? icon.bitmap() : default_image);
```
BenWiederhake, line 97: Will this *always* show the ladybug icon by default? That sounds like it might get old fast. I think it's perfectly fine to have a notification without an icon. Alternatively, if you really want icons, what do you think about *requiring* an icon from the caller?

The thing I want to avoid is putting uninformative "information" in a notification (which by definition is always low on screenspace).

Userland/Services/NotificationServer/ClientConnection.cpp
```diff
@@ -55,7 +55,7 @@ OwnPtr<Messages::NotificationServer::GreetResponse> ClientConnection::handle(con
 
 OwnPtr<Messages::NotificationServer::ShowNotificationResponse> ClientConnection::handle(const Messages::NotificationServer::ShowNotification& message)
 {
-    auto window = NotificationWindow::construct(message.text(), message.title(), message.icon());
+    auto window = NotificationWindow::construct(client_id(), message.text(), message.title(), message.icon());
```
BenWiederhake, line 58: I like using `client_id()` as workaround!

Userland/Libraries/LibGUI/Notification.h
```diff
@@ -30,6 +30,7 @@
 #include <LibGfx/Bitmap.h>
 
 namespace GUI {
+    class NotificationServerConnection;
```
BenWiederhake: Formatting: No leading spaces, and while you're add it please add a newline between `namespace GUI {` and `class NotificationServerConnection;`.

Userland/Libraries/LibGUI/Notification.cpp
```diff
@@ -60,15 +65,26 @@ Notification::~Notification()
 void Notification::show()
 {
     VERIFY(!m_showing);
+    VERIFY(!m_disposed);
+    if (!m_connection->is_connected()) {
+        // This would imply that the NotificationServer crashed before we could send it any data.
+        VERIFY_NOT_REACHED();
+    }
```
BenWiederhake, line 72: This edge case can be avoided by merging `m_showing`, `m_disposed`, and `m_connection` into `RefPtr<> m_connection` that contains a connection if and only if it's currently showing. In other words: The connection is only created on-demand.

Userland/Services/NotificationServer/ClientConnection.cpp
```diff
@@ -69,4 +69,29 @@ OwnPtr<Messages::NotificationServer::CloseNotificationResponse> ClientConnection
...
+{
+    auto window = NotificationWindow::get_window_by_id(client_id());
+    if (window) {
+        window->set_text(message.text());
+        window->set_title(message.title());
+    }
+    return make<Messages::NotificationServer::UpdateNotificationTextResponse>(window);
+}
+
+OwnPtr<Messages::NotificationServer::IsShowingResponse> ClientConnection::handle([[maybe_unused]] const Messages::NotificationServer::IsShowing& message)
```
BenWiederhake, line 91: I believe this should work:
```suggestion
OwnPtr<Messages::NotificationServer::IsShowingResponse> ClientConnection::handle(const Messages::NotificationServer::IsShowing&)
```

BenWiederhake (changes requested):
Now it's only two small syntax things, and a matter of taste (default icon and whether the two extra `boolean`s are really necessary). Feel free to ignore me on the "taste" things if you disagree ^^

Userland/Services/NotificationServer/ClientConnection.cpp
Sylvyrfysh: Nice side effect of using a single connection!

Sylvyrfysh:
Ran into an issue where closing a notification causes CPU usage to spike, and I can get a profile trace, but I can't see any symbols with a simple test app. Marking as a draft until then, could someone provide instructions on how to view the symbols?

Userland/Services/NotificationServer/NotificationWindow.cpp
Sylvyrfysh: This was removed by hiding the image if it was not valid.

Sylvyrfysh:
@BenWiederhake While I liked the RefPtr approach, it caused the core event loop to use 100% CPU when it was closed. I had to add state variables back, let me know if you see a better solution!

awesomekling:
This looks very cool :)

Please squash the "Formatting" commit into the relevant preceding commit.
Also please wrap commit messages at 72 characters per line.

awesomekling:
Oh and also please squash "Update Userland/Services/NotificationServer/ClientConnection.cpp" into the relevant preceding commit.

BenWiederhake:
> @BenWiederhake While I liked the RefPtr approach, it caused the core event loop to use 100% CPU when it was closed. I had to add state variables back, let me know if you see a better solution!

That sounds like a separate bug. You are right, working around that is a good idea. I'm unable to investigate this due to hardware issues.

Sylvyrfysh:
All changes requested have been made :)
Let me know if there's anything else to be done!

PR #5734, opened 2021-03-11
Kernel: Change AHCI port reset policy

supercomputer7:
The intention is to make the boot to be faster, therefore we should
decrease the time deltas in timeout loops to allow earlier break
from these.


tomuta:
We could also spawn a kernel thread. In theory, we could spawn a separate thread for each device, but maybe just one for this might make sense.

PR #5735, opened 2021-03-11
Kernel: Add support for VMWare graphics adapter

supercomputer7:
This is not a patch ready for merging yet, I still need to fix a few things.

@Lubrsi can you test this patch on VMWare Player and see if the kernel detects the adapter and can set a resolution of 1024x768?


Lubrsi:
> @Lubrsi can you test this patch on VMWare Player and see if the kernel detects the adapter and can set a resolution of 1024x768?

It does :+1:

supercomputer7:
> > @Lubrsi can you test this patch on VMWare Player and see if the kernel detects the adapter and can set a resolution of 1024x768?
> 
> It does +1

Cool! Does it work correctly though? on QEMU the screen is not updated (it does so only if I switch between QEMU and other window on my system).

Lubrsi:
> Cool! Does it work correctly though? on QEMU the screen is not updated (it does so only if I switch between QEMU and other window on my system).

It appears to just be stuck on a black screen, but the resolution is correct.

supercomputer7:
> > Cool! Does it work correctly though? on QEMU the screen is not updated (it does so only if I switch between QEMU and other window on my system).
> 
> It appears to just be stuck on a black screen, but the resolution is correct.

ah, OK, so I have a problem in my code, thanks for the help :)

Lubrsi:
IIRC, when I looked at the sample code, the driver has to send an update command via the FIFO to update the screen.

supercomputer7:
> IIRC, when I looked at the sample code, the driver has to send an update command via the FIFO to update the screen.

I've seen that too, I just hope we can somehow make it optimized.

Lubrsi:
It does allow us to update portions of the screen instead of the entire screen, but that'd need integrating into WindowServer and such.

tomuta:
> It does allow us to update portions of the screen instead of the entire screen, but that'd need integrating into WindowServer and such.

Hm that'd actually be fairly easy, the Compositor has a set of rectangles it knows that updated. So my guess is instead of flipping the frame buffer we'd essentially just send down a vector of rectangles so that vmware knows what needs to be repainted?

supercomputer7:
I'm closing this pull request for now, as I'm working on other aspect of the graphics subsystem. This PR will be re-opened when I'll be back to work on this (particularly the VMWare SVGA card).

PR #5738, opened 2021-03-11
Kernel: Implement triply indirect block support in Ext2FSInode

boricj:
Work in progress, but seems to works if you don't breathe on it too hard. I'm having trouble properly testing this because this is pushing several limits of SerenityOS (both architecturally and practically)... But what I do know is that according to `fsck.ext2` it's leaking metadata blocks yet according to the kernel logs they are freed, so this is probably not production-ready yet.


boricj:
Closing and reopening because GitHub did not realize I've force-pushed new commits.

boricj:
Reverting to the last HEAD commit when this PR was closed because GitHub does not allow reopening PRs otherwise.

boricj:
And we're back in business. GitHub had a pretty substantial outage according to GitHub Status, hence the rain dance to unwedge this PR.

The good news is that I believe this PR should be essentially good to go. I've stress-tested this particular section of code with a newly-created `stress-truncate` program that resizes a file in a loop with `truncate()` and given the amount of `VERIFY()` in this new code, I'm reasonably sure this won't immediately crash and burn upon merging.

The bad news is that there does not appear to be an internal filesystem test suite for SerenityOS, so I don't really have a way to check for regressions in non-obvious cases. I'm also pretty sure Ext2FS leaks indirect metadata blocks when a file gets unlinked, because I don't see `Ext2FSInode::flush_block_list()` being called at that point. However, as far as I can tell, this is not the result of changes in this PR.

So... Good to merge I suppose?

PR #5742, opened 2021-03-11
Toolchain: Always strip, make code less confusing

BenWiederhake:
I always manually strip the executables, simply because a 1.2 GB toolchain is ridiculous and might cause some slowdowns during build.

So I thought: Hey, we have the code here anyway, why not enable it for everyone(*)? So that's what I did.

(*): Except MacOS users. Apparently there's something wonky, but I can't test it.

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -143,14 +143,22 @@ jobs:
       env:
         SERENITY_KERNEL_CMDLINE: "boot_mode=self-test"
         SERENITY_RUN: "ci"
-      run: ninja run
+      run: |
+        echo "::group::ninja run # Qemu output"
+        run_worked=y
+        ninja run || run_worked=n
+        echo "::endgroup::"
+        echo "::group::cat debug.log # Serenity output"
+        if [ "y" = "${run_worked}" ] ; then
+          cat debug.log
+        else
+          echo "skipped (qemu had non-zero exit-code)"
+        fi
+        echo "::endgroup::"
+        [ "y" = ${run_worked} ]
       timeout-minutes: 10
       # FIXME: When stable, remove continue on error. (See issue #5541)
       continue-on-error: true
-    - name: Print target logs
-      if: ${{ !cancelled() && matrix.debug-macros == 'NORMAL_DEBUG'}}
-      working-directory: ${{ github.workspace }}/Build
-      run: cat ./debug.log
 
   build_and_test_lagom:
     runs-on: ${{ matrix.os }}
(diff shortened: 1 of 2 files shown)


IdanHo:
Looks like you accidentally have #5740's commits included in this branch as well

BenWiederhake:
That's why I marked it as "draft" before I ran out of time for the day :)

BenWiederhake:
Changes since the first push:
- Fix build-breaking mistake, doink!
- Make build-error analysis less irritating

BenWiederhake:
Changes since the first push:
- Fix *other* build-breaking mistake, doink²!

.github/workflows/cmake.yml
```diff
@@ -150,7 +150,7 @@ jobs:
     - name: Print target logs
       if: ${{ !cancelled() && matrix.debug-macros == 'NORMAL_DEBUG'}}
       working-directory: ${{ github.workspace }}/Build
-      run: cat ./debug.log
+      run: '[ ! -e debug.log ] || cat debug.log'
```
ADKaster, line 153: Hmm the point of this step is to make sure that we can see the boot dbg log even if the previous step which runs on target tests fails. If we don't generate the file at all because of "reasons" (maybe someone changes the run.sh step but forgot to update CI?) then this change would hide that, right? Maybe we could ferry some input/output from the `run tests` step? Not sure that's very elegant...

BenWiederhake: You're right that this might "hide" an error if it fails at just the right moment. However, `run.sh` calls qemu with `-debugcon file:debug.log`, so this step has to fail extremely early for the file to be missing; in which case the emulation step will have generated interesting output and resulted in nonzero exit-code.

I'm not aware of a much better solution:
- If we fail here, then it "hides" the problem by being duplicious.
- If we add `touch debug.log` before the emulation, then this step will also "fail silently".

What do you suggest?

ADKaster: Hmm. Yeah we'd have to be calling qemu incorrectly in order for it to hide the failure. Could make it not a separate step but just `ninja run || cat ./debug.log` with some `echo startgroup` magic in the middle? It's probably fine as is, that kind of error should be caught pretty quickly by a reviewer or by someone checking the logs.

BenWiederhake: I like that idea!

BenWiederhake:
Changes since last push:
- Rebased
- Changed the "ninja run qemu magic" to fail if and only if qemu failed.

CC @ADKaster, now it uses the "group" magic, what do you think? :)

ADKaster (approved):
Looks like it'll do the trick! No annoying errors this way.

PR #5746, opened 2021-03-12
LibJS: Implement 'Relative Indexing Method' proposal (.at())

linusg:
Still stage 3, but already implemented in major engines and unlikely to
change - there isn't much to change here anyway. :^)

See:

- https://github.com/tc39/proposal-relative-indexing-method
- https://tc39.es/proposal-relative-indexing-method/
- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/at

---

h/t @wongmjane, only [learned about this](https://twitter.com/wongmjane/status/1370306283610411010) today!


awesomekling:
Very neat! :)

PR #5747, opened 2021-03-12
Meta: serenity.sh tweaks and better Lagom support

linusg:
cc @tomuta, hope you agree with these :)


tomuta:
Seems sensible ;) Haven't tried it but looks like good changes.

PR #5749, opened 2021-03-12
Shell: Add support for indexing into variables

alimpfard:
Now a variable may have an optional slice (only _one_ slice), which can
also use negative indices to index from the end.
This works on both lists and strings.
The contents of the slice have the same semantics as brace expansions.
For example:
```sh
$ x=(1 2 3 4 5 6)
$ echo $x[1..3] # select indices 1, 2, 3
2 3 4
$ echo $x[3,4,1,0] # select indices 3, 4, 1, 0 (in that order)
4 5 2 1
$ x="Well Hello Friends!"
$ echo $x[5..9]
Hello
```


alimpfard:
I'm not sure if I want to allow slices on only variables, or also allow them on any expression.
Also not sure if I want to allow slicing into strings, or only lists.

(Hence why this is still a draft)

Issue #4601, opened 2020-12-28
Kernel: Coredumps waste lots of space and should probably be compressed


linusg:
A coredump of `crash -r` is 6359399 bytes (6 MiB), out of which 5821571 (5.5 MiB) are null bytes occurring in chunks of more than 100 consecutive null bytes - that's 91% (!). The largest chunk is 4195727 bytes (4 MiB) of nothing but null bytes.

A coredump of Browser viewing the default page (`SIGSEGV`'d with `kill`) is 16346680 bytes (15.6 MiB), out of which 7553781 (7.2 MiB) are null bytes occurring in chunks of more than 100 consecutive null bytes - that's 46%. The largest chunk is 4148707 bytes (3.9 MiB) of nothing but null bytes.

I haven't looked at coredump generation in detail and don't know the underlying file format, but we can probably compress them easily.

PR #5750, opened 2021-03-12
LibCompress+LibTar: Implement DEFLATE compression, gzip compression and tar archive creation

### LINKED ISSUE #3989: tar: the provided file is not a well-formated ustar file
I tried extracting the .tar.gz for the latest stable Milkytracker release:

https://github.com/milkytracker/MilkyTracker/releases/download/v1.02.00/milkytracker-1.02.00.tar.gz

It extracts fine using gnu tar under Ubuntu but when I try to extract the same archive with serenity's tar it says:

```
the provided file is not a well-formatted ustar file
```

IdanHo:
This PR implements several things:
- a bit-wise output stream and a minimum binary heap in AK (needed for deflate)
- a stream compressor for the DEFLATE compression format (RFC 1951) - A large part of of which is based on Hans Wennborg's great article about the DEFLATE and zip specifications: https://www.hanshq.net/zip.html
- a stream compressor for the GZip compression format (RFC 1952)
- a tar archive creation stream - while working on LibTar i also made it gnu/ustar agnostic which fixes #3989
- tar archive creation (using the above 3) in our tar utility

reviewer's note: due to serenity's coding style choice of discouraging the use of abbreviations combined with the devilishly intricate double-layer deep huffman coding in the deflate dynamic huffman blocks some of the variable names and function prototypes are comically long, so suggestions on tidying that up would be appreciated

i originally started working on this to provide the needed compression functionality for #4601, but i am not very familiar with that part of serenity so ill leave actually implementing that to someone else :)

### DIFF
--- a/AK/BinaryHeap.h
+++ b/AK/BinaryHeap.h
@@ -1,0 +1,128 @@
+/*
+ * Copyright (c) 2021, Idan Horowitz <idan.horowitz@gmail.com>
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#pragma once
+
+namespace AK {
+
+template<typename K, typename V, size_t Capacity>
+class BinaryHeap {
+public:
+    BinaryHeap() = default;
+    ~BinaryHeap() = default;
+
+    // This constructor allows for O(n) construction of the heap (instead of O(nlogn) for repeated insertions)
+    BinaryHeap(K keys[], V values[], size_t size)
+    {
+        VERIFY(size <= Capacity);
+        m_size = size;
+        __builtin_memcpy(m_keys, keys, size * sizeof(K));
+        __builtin_memcpy(m_values, values, size * sizeof(V));
+
+        for (ssize_t i = size / 2; i >= 0; i--) {
+            heapify_down(i);
+        }
+    }
+
+    [[nodiscard]] size_t size() const { return m_size; }
+    [[nodiscard]] bool is_empty() const { return m_size == 0; }
+
+    void insert(K key, V value)
+    {
+        VERIFY(m_size < Capacity);
+        auto index = m_size++;
+        m_keys[index] = key;
+        m_values[index] = value;
+        heapify_up(index);
+    }
+
+    V pop_min()
+    {
+        VERIFY(!is_empty());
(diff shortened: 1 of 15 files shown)


Userland/Libraries/LibCompress/Deflate.cpp
```diff
@@ -90,14 +84,17 @@ Optional<CanonicalCode> CanonicalCode::from_bytes(ReadonlyBytes bytes)
...
         }
     }
 
-    if (next_code != (1 << 15)) {
-        return {};
-    }
+    // FIXME: I'm not quite sure what this is checking for, but it fails for certain (seemingly) valid codes
+    // if (next_code != (1 << 15)) {
+    //     return {};
+    // }
```
bcoles, line 97: > FIXME: I'm not quite sure what this is checking for, but it fails for certain (seemingly) valid codes

Me neither. The [tinflate.c](https://achurch.org/tinflate.c) DEFLATE implementation is simplistic and makes for a good comparison.

It has a similar code pattern including this check in `gen_huffman_table`.

```
    if (first_code[15] + length_count[15] != (unsigned short)1<<15) {
        /* The Huffman tree is incomplete and thus invalid. */
        return 0;
    }
```

The calculation of `first_code` (`next_code` in Serenity's implementation) is also different.

```
first_code[i] = (first_code[i-1] + length_count[i-1]) << 1;
```

Userland/Libraries/LibCompress/Deflate.cpp
```diff
@@ -448,4 +449,571 @@ void DeflateDecompressor::decode_codes(CanonicalCode& literal_code, Optional<Can
...
+                encoded_lengths[encoded_count++].symbol = 0;
+                i++;
+                continue;
+            }
+
+            if (zero_count <= 10) {
+                encoded_lengths[encoded_count].symbol = 17;
+                encoded_lengths[encoded_count++].count = zero_count;
+            } else {
+                encoded_lengths[encoded_count].symbol = 18;
```
alimpfard, line 823: Maybe these "symbols" (0, 17, 18, ...) could use some names?
e.g. an enum?
I dunno what it's referring to though, so feel free to ignore this if it makes no sense

IdanHo: Yep they do have names and theyre used in several places (including the decompressor) so ill turn them into an enum :+1:

Userland/Libraries/LibCompress/Deflate.cpp
IdanHo: While i'm still not entirely sure what that checks for, that library has a special case for canonical codes with exactly 1 symbol, and in fact the valid codes which this check failed for are codes with exactly 1 symbol, so ill just add a similar special case for 1 symbol and re-enable the check :)

AK/BinaryHeap.h
```diff
@@ -0,0 +1,128 @@
...
+            if (m_keys[index] >= m_keys[parent])
+                break;
+            swap(m_keys[index], m_keys[parent]);
+            swap(m_values[index], m_values[parent]);
+            index = parent;
+        }
+    }
+
+    K m_keys[Capacity];
+    V m_values[Capacity];
```
zmodem, line 122: Instead of using two arrays, one idea would be have an array of two-element structs:

```
struct {
  K key;
  V value;
} m_elements[Capacity];
```

This way, instead of separate swaps for the keys and values, you could swap them both at once:

`swap(m_elements[x], m_elements[y]);`

and for example the key and values are 32 bits, they could both be swapped with a 64-bit load and store.

AK/BitStream.h
```diff
@@ -121,6 +122,85 @@ class InputBitStream final : public InputStream {
...
 
+class OutputBitStream final : public OutputStream {
+public:
+    explicit OutputBitStream(OutputStream& stream)
+        : m_stream(stream)
+    {
+    }
+
+    // WARNING: write aligns to the next byte boundary before writing, if unaligned writes are needed this should be rewritten
+    size_t write(ReadonlyBytes bytes) override
```
zmodem, line 133: Perhaps write_aligned would be a better name?

Userland/Libraries/LibCompress/Deflate.cpp
```diff
@@ -448,4 +464,571 @@ void DeflateDecompressor::decode_codes(CanonicalCode& literal_code, Optional<Can
...
+        size_t bit_length = 1;
+        while (link != 2) {
+            bit_length++;
+            link = huffman_links[link];
+        }
+
+        if (bit_length > max_bit_length) {
+            VERIFY(frequency_cap != 1);
+            frequency_cap /= 2;
+            goto try_again; // FIXME: gotos are ugly, but i cant think of a good way to flatten this
```
zmodem, line 648: Sorry about that :-p

I think one way would be if generate_huffman_lengths took frequency_cap as a parameter (with default value UINT16_MAX). Then you could tail call here:

return generate_huffman_lengths(lengths, frequencies, max_bit_length, frequency_cap / 2)

AK/BitStream.h
```diff
@@ -121,6 +122,85 @@ class InputBitStream final : public InputStream {
...
+        VERIFY(count <= 32);
+        size_t n_written = 0;
+        while (n_written < count) {
+            if (m_stream.has_any_error()) {
+                set_fatal_error();
+                return;
+            }
+
+            if (m_next_byte.has_value()) {
+                m_next_byte.value() |= ((bits >> n_written) & 1) << m_bit_offset;
```
zmodem, line 163: Optimizing this routine to do more than a bit at a time would probably give a good speed improvement.

zmodem (commented):
Very cool! I skimmed the code and left some comments. Please feel free to ignore all of them :-)

AK/BitStream.h
IdanHo: unfortunately write is a pure virtual method im implementing from the abstract OutputStream, write_aligned would have been a better name if we were not constrained

IdanHo:
@zmodem: thanks for the great suggestions, theyre up for review in #5767

zmodem:
> @zmodem: thanks for the great suggestions, theyre up for review in #5767

Looks great!

PR #5757, opened 2021-03-13
Kernel: Introduce the new SysFS

supercomputer7:
The final goal of this PR is to add support for userspace events via the SysFS.
Currently there are 2 folders - `acpi` with all available ACPI tables on the machine and `bios` which exposes 2 DMI blobs.
I still don't know how I want to implement it, so if anyone has a suggestion, I'd be happy to see your comments :)

I should also fix the dmidecode port in this PR too, because it relies on the DMI blobs to be exposed in `/proc`.


supercomputer7:
I'll continue to work on this once I finish to work on #6277.

PR #5765, opened 2021-03-13
Ports: Fix c-ray

linusg:


Userland/Libraries/LibM/math.cpp
```diff
@@ -445,6 +445,11 @@ double tan(double angle) NOEXCEPT
     return ampsin(angle) / ampsin(M_PI_2 + angle);
 }
 
+float tanf(float angle) NOEXCEPT
+{
+    return (float)tan((double)angle);
```
awesomekling, line 450: 🤷🏻

linusg: what can I say, it has the CxByte stamp of approval though! :smile:

BenWiederhake: Are the manual casts really necessary? For me it was sufficient to write `return tan(angle);`

linusg: not necessary, but at least this way it's very clear that there's casts both ways :)

PR #5766, opened 2021-03-13
Everywhere: working towards 64 bits storage

boricj:
I wanted to properly test out my triply indirect block support for Ext2FS, but it turns out the storage stack of SerenityOS has trouble dealing with disks, files and file systems bigger than 2 GiB... Since this is not the 1990s anymore, bring out the storage stack into the 2000s, kicking and screaming.

Suffice to say, this is snowballing out of control pretty badly since this requires changes to syscalls, system types, device drivers and probably other things I've yet to discover... Note that while this branch appears to work, Ext2FS so far has yet to survive an encounter with a 8 GiB root filesystem and live to tell the tale.

### DIFF
--- a/Kernel/Storage/PATADiskDevice.h
+++ b/Kernel/Storage/PATADiskDevice.h
@@ -57,19 +57,18 @@ public:
     };
 
 public:
-    static NonnullRefPtr<PATADiskDevice> create(const IDEController&, IDEChannel&, DriveType, InterfaceType, u16, u16, u16, u16);
+    static NonnullRefPtr<PATADiskDevice> create(const IDEController&, IDEChannel&, DriveType, InterfaceType, u16, u64);
     virtual ~PATADiskDevice() override;
 
     // ^StorageDevice
     virtual Type type() const override { return StorageDevice::Type::IDE; }
-    virtual size_t max_addressable_block() const override;
 
     // ^BlockDevice
     virtual void start_request(AsyncBlockDeviceRequest&) override;
     virtual String device_name() const override;
 
 private:
-    PATADiskDevice(const IDEController&, IDEChannel&, DriveType, InterfaceType, u16, u16, u16, u16);
+    PATADiskDevice(const IDEController&, IDEChannel&, DriveType, InterfaceType, u16, u64);
 
     // ^DiskDevice
     virtual const char* class_name() const override;
--- a/Kernel/Storage/SATADiskDevice.cpp
+++ b/Kernel/Storage/SATADiskDevice.cpp
@@ -33,12 +33,12 @@
 
 namespace Kernel {
 
-NonnullRefPtr<SATADiskDevice> SATADiskDevice::create(const AHCIController& controller, const AHCIPort& port, size_t sector_size, size_t max_addressable_block)
+NonnullRefPtr<SATADiskDevice> SATADiskDevice::create(const AHCIController& controller, const AHCIPort& port, size_t sector_size, u64 max_addressable_block)
 {
     return adopt(*new SATADiskDevice(controller, port, sector_size, max_addressable_block));
 }
 
-SATADiskDevice::SATADiskDevice(const AHCIController& controller, const AHCIPort& port, size_t sector_size, size_t max_addressable_block)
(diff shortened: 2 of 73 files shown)


Lubrsi:
Potentially relevant issue: #5376

Kernel/Storage/SATADiskDevice.cpp
```diff
@@ -60,7 +60,7 @@ void SATADiskDevice::start_request(AsyncBlockDeviceRequest& request)
 
 String SATADiskDevice::device_name() const
 {
-    return String::formatted("hd{:c}", 'a' + minor());
+    return String::formatted("sd{:c}", 'a' + minor());
```
supercomputer7, line 63: This is not a typo, but a name that was decided because it represents "hard drive" for the letters `hd`.
As far as I know, `sd` represents a compatibility with the `SCSI`  subsystem in Linux (which means that almost everything besides NVMe emulates SCSI).

boricj: What happens if a system has both IDE and AHCI drives at the same time?

danboid:
This is prob the biggest single limitation for Serenity currently IMO so nice to see it being tackled.

Kernel/Storage/SATADiskDevice.cpp
supercomputer7: Ideally, the kernel shouldn't worry about naming of storage devices, so we should have a userspace `DeviceMapper` (as a service in `SystemServer`) for such purpose (look at #5757 for this).
for the time being, we just call a method from the `StorageManagement` singleton to handle the decision of the "minor number", so there's no way that both drives will have the same name.

Kernel/Storage/PATADiskDevice.h
```diff
@@ -80,6 +80,7 @@ class PATADiskDevice final : public StorageDevice {
     u16 m_cylinders { 0 };
     u16 m_heads { 0 };
     u16 m_sectors_per_track { 0 };
+    u64 m_max_lba { 0 };
```
supercomputer7, line 83: Instead of creating a new field in `PATADiskDevice`, you could pass the max LBA value for the `StorageDevice` constructor.

boricj:
Making progress. The next blocker is the `lseek` syscall because the kernel syscall ABI deals in `uintptr_t` quantities, so garbage is returned to the userland if the offset goes beyond 2 GiB. The least painful way to deal with this is probably to use a pointer to `off_t` to pass and return the offset on 32-bit systems.

awesomekling:
> Making progress. The next blocker is the `lseek` syscall because the kernel syscall ABI deals in `uintptr_t` quantities, so garbage is returned to the userland if the offset goes beyond 2 GiB. The least painful way to deal with this is probably to use a pointer to `off_t` to pass and return the offset on 32-bit systems.

We can also change the syscall to take a pointer to a result struct where the 64-bit offset would be written to. :)

IdanHo:
maybe we should introduce a lseek64 to keep posix compatibility?

boricj:
> maybe we should introduce a lseek64 to keep posix compatibility?

We don't have to provide POSIX features exclusively via the kernel syscall table, lseek64 can be provided as a userland shim.

Speaking of, we also don't need to name functions in the kernel syscall tables with names dating back from the 1970s. We should probably rename sys$lseek to just sys$seek (obviously we should not rename the POSIX API functions themselves).

boricj:
> We can also change the syscall to take a pointer to a result struct where the 64-bit offset would be written to. :)

We would then have to deal with a userland pointer on both 32 and 64 bit architectures. I'll probably just turn the offset into a `off_t*`, there's no other value to return except an error and there's no 64 bit architecture supported to micro-optimize anyway.

Lubrsi:
In IDEChannel::ata_access, now that you've upgraded lba to be u64, you also need to pass in the bits 32-39 and bits 40-47 of the LBA here:
https://github.com/SerenityOS/serenity/blob/d792200a5531bf9872e19c7c7a3ba4c865ad43b6/Kernel/Storage/IDEChannel.cpp#L432-L433

boricj:
There's simply too much stuff to fix at once. While the system itself should be 64 bit clean by now, the rest of the system is definitely not. I especially suspect the virtual memory system to be partially at fault.

Let's get this merged for now.

awesomekling:
Pure awesome :)

PR #5771, opened 2021-03-13
(WIP) LibWeb+LibJS: Make WrapperGenerator (i.e. our IDL compiler) less LibWeb-centric

alimpfard:
This PR tries to make the IDL compiler more general, by allowing the IDL files declare:
- Their toplevel namespace (`module`)
- The headers they need, and any inter-IDL dependencies they have (`import`)

If and when this PR is merged, we should be able to compile IDLs (still for LibJS) that are completely unrelated to LibWeb, and contain no references to it.

To do this, I have to figure out an annoyingly large number of dependencies, so this PR will be a WIP for a while.

---

Status: This is currently b0rked.


linusg:
hmm, I thought the IDL impl in serenity was supposed to follow https://heycam.github.io/webidl/ - so we're deviating from that now?

alimpfard:
Huh.
I didn't know we were following that, I'll take a look at it and see if we can decouple the IDL stuff from LibWeb and still comply with that spec.

alimpfard:
Parsing the IDL isn't particularly interesting, the part this is trying to address is where any bindings generates are specific to LibWeb.
That can map to the namespaces in the spec you mentioned, but nothing in that spec immediately jumps out for imports 

Resolving the types and interfaces to a minimal set of imports (without explicitly declared dependencies) would require the code generator to load every single IDL file, and then resolve every single dependency - not a very nice solution tbh

alimpfard:
Seems like webidl wants to have all dependencies remain implicit, in which case, how are we supposed to resolve them?
Clearly loading every single IDL file to check if we depend on it, or need it for anything for every single IDL file is silly...
Yet it seems this is _exactly_ what Blink does.


I guess we can do that...maybe make a big interface database first, then resolve all the IDL files with that :thinking:

alimpfard:
This is far too big an architecture change for my tastes.
I'll reopen the PR if I get somewhere with this.

awesomekling:
At first glance, adding import statements seems reasonable, although I suppose there might be circular dependencies in some interfaces.. so maybe the "load errything" approach is ultimately the sane one.

alimpfard:
Yeah, but then the compiler has to have multiple passes (unless we want quadratic codegen time - eww), and honestly, (de)serialising in C++ is one of the absolute worst things one can attempt, so...

PR #5777, opened 2021-03-14
readelf: Add support for core notes and relocations

bcoles:
`readelf` now supports core notes and relocations. Perhaps the must immediately useful feature is the ability to read notes from core dump files.

read notes from core dump file

This also adds minimal output for the dynamic section (simply printing the number of dynamic entries, but not displaying these entries).

This also adds output for unwind which simply prints a message stating that unwind info is not supported. This is the same as GNU `readelf` which supports unwind info only for a very limited number of architectures (ie, not x86 and x86_64). However, GNU `readelf` attempts to parse unwind information as part of `-a` (all) (even though it will usually fail) and so does our `readelf`. Thus adding a message related to unwind info is included for consistency.

### DIFF
--- a/Userland/Utilities/readelf.cpp
+++ b/Userland/Utilities/readelf.cpp
@@ -237,6 +237,36 @@ static const char* object_symbol_binding_to_string(Elf32_Word type)
     }
 }
 
+static const char* object_relocation_type_to_string(Elf32_Word type)
+{
+    switch (type) {
+    case R_386_NONE:
+        return "R_386_NONE";
+    case R_386_32:
+        return "R_386_32";
+    case R_386_PC32:
+        return "R_386_PC32";
+    case R_386_GOT32:
+        return "R_386_GOT32";
+    case R_386_PLT32:
+        return "R_386_PLT32";
+    case R_386_COPY:
+        return "R_386_COPY";
+    case R_386_GLOB_DAT:
+        return "R_386_GLOB_DAT";
+    case R_386_JMP_SLOT:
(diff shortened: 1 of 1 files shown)


Dexesttp:
Rather than having two `has_section` and `lookup_section` methods, could you make `lookup_section` return an "Optional<Section>" ? I don't think it would slow things down since the `lookup_section` method (and the new `has_section`) are already creating a bunch of section objects  to look up their names.

bcoles:
> 
> 
> Rather than having two `has_section` and `lookup_section` methods, could you make `lookup_section` return an `Optional<Section>` ? I don't think it would slow things down since the `lookup_section` method (and the new `has_section`) are already creating a bunch of section objects to look up their names.

Perhaps. I'm not interested in reworking the ELF image and relocation image API for the moment.

I've removed the changes to LibELF from this PR. It now only modifies `readelf`.

awesomekling:
Nice!

bcoles:
> 
> 
> > Rather than having two `has_section` and `lookup_section` methods, could you make `lookup_section` return an `Optional<Section>` ? I don't think it would slow things down since the `lookup_section` method (and the new `has_section`) are already creating a bunch of section objects to look up their names.
> 
> Perhaps. I'm not interested in reworking the ELF image and relocation image API for the moment.
> 
> I've removed the changes to LibELF from this PR. It now only modifies `readelf`.

I finally circled back to look at this, only to discover that it was implemented by kling in 16221305ade1e37d68c95d9bf674021afc18570f yesterday.

PR #5782, opened 2021-03-14
Ports: Allow verbose argument in build_all.sh

### LINKED ISSUE #5783: Bash port fails to build with error configure: cross-compiling for i686-pc-serenity is not supported
I've been watching Serenity development on YouTube for a while now, and finally took it for a spin myself and I hit some errors building the bash port. I am getting the following error configure: cross-compiling for i686-pc-serenity is not supported. Is this a known issue or am i doing something wrong?

Hitting this issue inspired me to open #5782

06needhamt:
This patch allows for a verbose argument to be passed
so that the build output of the individual builds
is printed to stdout instead of /dev/null to help with diagnosing errors

If the verbose argument is not passed the old behaviour is preserved
and the build output is printed to /dev/null

Also fixes #5783

### DIFF
--- a/Ports/build_all.sh
+++ b/Ports/build_all.sh
@@ -1,9 +1,25 @@
 #!/usr/bin/env bash
 
 clean=false
+verbose=false
+
 case "$1" in
     clean)
         clean=true
+        ;;
+    verbose)
+        verbose=true
+        ;;
+    *)
+        ;;
+esac
+
+case "$2" in
+    clean)
+        clean=true
+        ;;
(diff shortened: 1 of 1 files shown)


06needhamt:
Hello

I've been watching Serenity development on YouTube for a while now, and finally took it for a spin myself and I hit some errors building some of the ports so I went and removed the output redirect to /dev/null and then thought "This would be a perfect first patch". So here it is.

Keep up the good work.

PR #5762, opened 2021-03-13
Make some ports build, document shortcomings

BenWiederhake:
This fixes / touches up some ports that now build:
- bc
- git
- (In some sense: libogg, genemu)

These ports already worked:
- bash
- binutils
- bison
- byacc
- bzip2
- cmake
- curl
- dmidecode
- doom
- ed
- figlet
- flex
- freetype
- gcc
- genemu
- grep
- indent
- jot
- jq
- klong
- less
- libarchive
- libffi
- libogg
- ncurses
- pcre2
- python3
- quake
- rsync
- SDL2
- SDL2_mixer
- SDL2_ttf - depends on libfreetype-dev of host
- sed
- sl
- termcap
- tinycc
- tinyscheme
- vim
- vttest
- zlib
- zstd

The following ports currently are broken, and I don't know how to fix them:
- **dash and dropbear:** static linkage is broken, will probably be fixed by #5758
- **c-ray and frotz:** desync from upstream
- **diffutils, libiconv, nano:** conflicting definitions of lstat?!
- **libexpat:** "hidden symbol `main' in outline.o is referenced by DSO"
- **mrsh:** "local symbol `main' in /tmp/cchQaYBA.o is referenced by DSO" during configure ("Checking for exported symbol restrictions")
- **ninja:** "hidden symbol `main' in build/ninja.o is referenced by DSO"
- **gnuplot:** Depends on tex?!
- **openssh:** Depends on broken openssl
- **openssl:** Apparently we're missing AI_PASSIVE, NI_MAXHOST, NI_MAXSERV, NI_NUMERICHOST, NI_NUMERICSERV. Maybe due to EAI_FAMILY?
- **stress-ng:** Something about zlib and feclearexcept, I don't get it.
- **tr:** How could this possibly work? conflicting definition of ctypes, doesn't use configure

I took the liberty to remove the "editline" port, as it has never worked. See #2532 and #2677.

CC @alimpfard and @linusg, and everyone else who cares about ports. I'm currently implementing some kind of cronjob that automatically checks the state of the ports every week, that might help in keeping bitrot in check.

PR #5792, opened 2021-03-14
LibM: Add more math functionality

RealKC:
@danboid @aabajyan does this help with the building hatari?

@BenWiederhake I implemented lgamma+signnam in this, as you said gnuplot wants them in https://github.com/SerenityOS/serenity/pull/5762#issuecomment-798790155, does this help in building it? I tried building gnuplot myself, but our gcc complained about not recognizing -pthread

Also can someone check I didn't declare any functions in math.h but forgot to actually implement them in math.cpp? I first wrote the missing functions in the header and then implemented them, and I'm not sure that was the best of ideas :sweat_smile:


IdanHo:
looks like the remainder family is slightly different compared to fmod: https://ftp.gnu.org/old-gnu/Manuals/glibc-2.2.3/html_node/libc_406.html, but just leaving it as is and adding a FIXME might be good enough for now

RealKC:
> looks like the remainder family is slightly different compared to fmod: [ftp.gnu.org/old-gnu/Manuals/glibc-2.2.3/html_node/libc_406.html](https://ftp.gnu.org/old-gnu/Manuals/glibc-2.2.3/html_node/libc_406.html), but just leaving it as is and adding a FIXME might be good enough for now

Ah yeah, that's a good idea

aabajyan:
Thank you! It compiled successfully! :D

PR #5801, opened 2021-03-15
UserspaceEmulator: downgrade TODO to dbgln for invalid fcntl cmd arg

bcoles:
Test: `/usr/Tests/Kernel/crash-fcntl-invalid-cmd`


Userland/DevTools/UserspaceEmulator/Emulator_syscalls.cpp
```diff
@@ -668,7 +668,7 @@ u32 Emulator::virt$fcntl(int fd, int cmd, u32 arg)
     case F_ISTTY:
         break;
     default:
-        TODO();
+        dbgln("Invalid fcntl cmd: {}\n", cmd);
```
awesomekling, line 671: No need for `\n` when using `dbgln()` :)

PR #5804, opened 2021-03-15
Base: Change path to both `modload` and `modunload`

r0ei:
Seems like the `MD` file of `modload` and `modunload` has moved to `man8` instead of `man1`.

### DIFF
--- a/Base/usr/share/man/man7/kernel_modules.md
+++ b/Base/usr/share/man/man7/kernel_modules.md
@@ -59,7 +59,7 @@ extern "C" void module_fini()
 
 ## See also
 
-* [`modload`(1)](../man1/modload.md)
-* [`modunload`(1)](../man1/modunload.md)
+* [`modload`(1)](../man8/modload.md)
+* [`modunload`(1)](../man8/modunload.md)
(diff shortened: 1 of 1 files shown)


Base/usr/share/man/man7/kernel_modules.md
```diff
@@ -59,7 +59,7 @@ extern "C" void module_fini()
 
 ## See also
 
-* [`modload`(1)](../man1/modload.md)
-* [`modunload`(1)](../man1/modunload.md)
+* [`modload`(1)](../man8/modload.md)
+* [`modunload`(1)](../man8/modunload.md)
```
awesomekling, line 63: The labels should be updated too: `modload(8)` and `modunload(8)`.

Also, please observe the [contribution guidelines](https://github.com/SerenityOS/serenity/blob/master/CONTRIBUTING.md) and add a category to your commit message. "Base" would be fine.

r0ei: _The labels should be updated too: modload(8) and modunload(8)_ - Sorry, my bad, didn't see that at first.

PR #5807, opened 2021-03-15
LibM: Fix issues and add fmax/fmin

RealKC:


Userland/Libraries/LibM/math.cpp
```diff
@@ -1364,4 +1364,64 @@ long double scalblnl(long double x, long exponent) NOEXCEPT
 {
     return internal_scalbn(x, exponent);
 }
+
+long double fmaxl(long double x, long double y) NOEXCEPT
+{
+    if (isnan(x))
+        return y;
+    else if (isnan(y))
```
awesomekling, line 1372: No need for `else` after `return` :)

PR #5808, opened 2021-03-15
Kernel+LibC: Make libc.a self contained and fix TLS allocation for static programs

### LINKED ISSUE #5758: LibCStatic is not self-contained (which causes some ports to fail)
Currently, the following command fails:

`./Toolchain/Local/i686/bin/i686-pc-serenity-gcc -o mytest -static mytest.c`

For the following input:

`int main () { return 0; }`

This is because `libc.a` (cmake-target "LibCStatic") does not contain ssp.c or LibSystem:
```
/home/user/workspace/serenity/Toolchain/Local/i686/lib/gcc/i686-pc-serenity/10.2.0/../../../../i686-pc-serenity/bin/ld: /home/user/workspace/serenity/Toolchain/../Build/i686/Root/usr/lib/libc.a(Time.cpp.o): in function `AK::Time::operator-(AK::Time const&) const':
./Build/i686/../../AK/Time.cpp:284: undefined reference to `__stack_chk_guard'

/home/user/workspace/serenity/Toolchain/Local/i686/lib/gcc/i686-pc-serenity/10.2.0/../../../../i686-pc-serenity/bin/ld: /home/user/workspace/serenity/Toolchain/../Build/i686/Root/usr/lib/libc.a(times.cpp.o): in function `unsigned int syscall<Kernel::Syscall::Function, tms*>(Kernel::Syscall::Function, tms*)':
./Build/i686/../../Userland/Libraries/LibSystem/syscall.h:49: undefined reference to `syscall1'
```

A quick look with `objdump -t Build/i686/Root/usr/lib/libc.a | grep stack` reveals that these symbols really are missing.

This is an actual problem, not just a hypothetical one. It causes some ports to fail building, for example `dash`. We might choose to make dash link dynamically, but it seems like a bad idea to not support static linking.

I can't fix the problem by myself: The `LibC/CMakeLists.txt` already contains this line:
```
target_link_libraries(LibCStatic crt0 ssp system)
```
… so I don't know why it still doesn't include those symbols.

itamar8910:
Closes #5758.

Apparently simply using `target_link_libraries` with static archives does not actually include the contents of the dependencies in the target archive.
So previously, `libc.a` did not contain symbols from `ssp` and `LibSystem` which caused compilation of statically linked programs to fail.

This PR fixes this by first building `LibCStaticWithoutDeps`, which is an archive that consists of all object files of libc without any other required dependencies. 

We then build `libc.a` with a custom CMake rule that first extracts all object files out of the archives of `LibCStaticWithoutDeps` and the required dependencies, and then combines them all into a single libc.a.

Doing this with a custom CMake target is not the prettiest, but I didn't find any built-in way to do this with CMake. ~~And hardcoding the path to `libstdc++` is meh~~ (We actually don't need to link libcstdc++ here). So suggestions to improve the way it's done here are welcome.

I also fixed some loading code in the kernel to actually setup a TLS region for statically linked programs (I'm pretty sure it used to work right after we switched to dynamic linking and we just didn't notice when it broke).

Compiling & running a test program with `-static` now works. 

cc @BenWiederhake @ADKaster

### DIFF
--- a/Kernel/Syscalls/execve.cpp
+++ b/Kernel/Syscalls/execve.cpp
@@ -310,6 +310,7 @@ static KResultOr<LoadResult> load_elf_object(NonnullOwnPtr<Space> new_space, Fil
                 ph_load_result = region_or_error.error();
                 return IterationDecision::Break;
             }
+
             master_tls_region = region_or_error.value();
             master_tls_size = program_header.size_in_memory();
             master_tls_alignment = program_header.alignment();
@@ -444,6 +445,11 @@ KResultOr<LoadResult> Process::load(NonnullRefPtr<FileDescription> main_program_
         auto result = load_elf_object(new_space.release_nonnull(), main_program_description, FlatPtr { 0 }, ShouldAllocateTls::Yes);
         if (result.is_error())
             return result.error();
+
+        m_master_tls_region = result.value().tls_region;
+        m_master_tls_size = result.value().tls_size;
+        m_master_tls_alignment = result.value().tls_alignment;
+
         return result;
     }
 
--- a/Kernel/Thread.cpp
+++ b/Kernel/Thread.cpp
@@ -1005,7 +1005,7 @@ size_t Thread::thread_specific_region_size() const
 
 KResult Thread::make_thread_specific_region(Badge<Process>)
 {
-    // The process may not require a TLS region
+    // The process may not require a TLS region, or allocate TLS later with sys$allocate_tls (which is what dynamically loaded programs do)
     if (!process().m_master_tls_region)
         return KSuccess;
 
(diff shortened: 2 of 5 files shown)


BenWiederhake:
Because I forgot to write it in the issue: DynamicLoader currently hand-picks the `.cpp` files that it includes into itself, so as soon as our LibCStatic works again that would be a great opportunity/sanitycheck to switch it from "hand-picked selection of files" to "properly link against the static LibC".

itamar8910:
@BenWiederhake well for the dynamic loader we need to compile the `libc` sources with the `NO_TLS` preprocessor flag defined, because we do not support allocating TLS for loader itself (and we do not need it since the loader is single-threaded).

What we could do is compile a separate target a-la `LibCNoTLS` and link that with the loader, but that's better done outside this PR.

BenWiederhake:
Oh, interesting. Sure, I thought the DynamicLoader should "just work" with a LibCStatic. I was wrong ^^'

awesomekling:
This fails CI:

```
Run ./check-symbols.sh
         U __cxa_guard_acquire
         U __cxa_guard_acquire
         U __cxa_guard_acquire
         U __cxa_guard_acquire
         U __cxa_guard_acquire
         U __cxa_guard_acquire
         U __cxa_guard_acquire
         U __cxa_guard_acquire
Forbidden undefined symbol in LibC: __cxa_guard_acquire
See comment in Meta/check-symbols.sh for more info
```

ADKaster:
__cxa_guard_acquire is a symbol from libgcc.a and/or libgcc_s.so required
for function-scope statics to work.

Probably need to statically link that guy too.

itamar8910:
> This fails CI:

This CI check guards us from having undefined references to those functions in libc,
but `__cxa_guard_acquire` is actually already defined in `libc.a` (definition is in `guard.o`), but symbols-check didn't realize this. I fixed symbols-check to not fail when this is the case.

Issue #3112, opened 2020-08-12
QuickShow: Zooming to cursor is broken


xTibor:
Zooming to cursor in QuickShow has recently regressed in two different ways.

**First:** When the zoom level crosses the 100% mark while zooming in or out, the canvas position resets itself to the center.
Repro: https://imgur.com/dlz9IIp
Regressed by: https://github.com/SerenityOS/serenity/commit/ea1ebe866299ef75376dc79c9406faa35c431d24

**Second:** When the app automatically resizes the window on zoom it doesn't shift the window or canvas position to maintain the cursor target. This auto-resizing also results in ridiculously large windows which hurts usability.
Repro: https://imgur.com/JoCf2b6
Regressed by: https://github.com/SerenityOS/serenity/commit/0001bbf1824c17d048f0ec8dde75eccf91a86f26

PR #5809, opened 2021-03-15
More intuitive zooming for QuickShow

olegv11:
Make it so that you always zoom "into" the pixel the mouse points to, which mimics behavior of Irfanview or Gwenview. Before it frequently over- and undershot which made the zoom hard to use.


awesomekling:
This looks great! :star:

olegv11:
Forgot to check the issue tracker beforehand, but it partially fixes #3112. There's still some weirdness going on when the window resizes after zoom (maybe we can shift the mouse cursor like IrfanView does it and also cap the window size), but I guess that's for another pull request.

PR #5816, opened 2021-03-16
UserspaceEmulator: Print PID in mmap nullptr with MAP_FIXED reportln()

bcoles:
Unbalanced `{}` were causing an unrelated crash in `AK/Span.h`.

Test: `ue /usr/Tests/Kernel/munmap-multi-region-unmapping`

### DIFF
--- a/Userland/DevTools/UserspaceEmulator/Emulator_syscalls.cpp
+++ b/Userland/DevTools/UserspaceEmulator/Emulator_syscalls.cpp
@@ -750,7 +750,7 @@ u32 Emulator::virt$mmap(u32 params_addr)
         else {
             // mmap(nullptr, …, MAP_FIXED) is technically okay, but tends to be a bug.
             // Therefore, refuse to be helpful.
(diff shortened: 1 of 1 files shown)


alimpfard:
Maybe slap `reportln` with the CheckedFormatString stick?
Might find a bug or two 🤷

bcoles:
> 
> 
> Maybe slap `reportln` with the CheckedFormatString stick?
> Might find a bug or two shrug

I manually checked for other instances of unbalanced `{}` and didn't find any other violations.

I'm not sure how to run `CheckedFormatString` ?

alimpfard:
Oh, it's supposed to replace the format string type
So for instance, `template<typename... Args> reportln(StringView fmt, Args... args)` turns into `template<typename... Args> reportln(CheckedFormatString<Args...> fmt, Args... args)`.

BenWiederhake:
D:

/me hides in shame

bcoles:
> 
> 
> Oh, it's supposed to replace the format string type
> So for instance, `template<typename... Args> reportln(StringView fmt, Args... args)` turns into `template<typename... Args> reportln(CheckedFormatString<Args...> fmt, Args... args)`.

`CheckedFormatString` flagged no issues.

PR #5819, opened 2021-03-16
Feature/invalid date

leggenda47:


Userland/Libraries/LibJS/Runtime/Date.cpp
```diff
@@ -32,15 +32,16 @@
 
 namespace JS {
 
-Date* Date::create(GlobalObject& global_object, Core::DateTime datetime, u16 milliseconds)
+Date* Date::create(GlobalObject& global_object, Core::DateTime datetime, u16 milliseconds, bool invalid_date)
```
linusg, line 35: Let's call this the same as the member:

```suggestion
Date* Date::create(GlobalObject& global_object, Core::DateTime datetime, u16 milliseconds, bool is_invalid)
```

Userland/Libraries/LibJS/Runtime/Date.cpp
```diff
@@ -32,15 +32,16 @@
...
 {
-    return global_object.heap().allocate<Date>(global_object, datetime, milliseconds, *global_object.date_prototype());
+    return global_object.heap().allocate<Date>(global_object, datetime, milliseconds, invalid_date, *global_object.date_prototype());
```
linusg, line 37:
```suggestion
    return global_object.heap().allocate<Date>(global_object, datetime, milliseconds, is_invalid, *global_object.date_prototype());
```

Userland/Libraries/LibJS/Runtime/Date.cpp
```diff
@@ -32,15 +32,16 @@
...
 }
 
-Date::Date(Core::DateTime datetime, u16 milliseconds, Object& prototype)
+Date::Date(Core::DateTime datetime, u16 milliseconds, bool invalid_date, Object& prototype)
```
linusg, line 40:
```suggestion
Date::Date(Core::DateTime datetime, u16 milliseconds, bool is_invalid, Object& prototype)
```

Userland/Libraries/LibJS/Runtime/Date.cpp
```diff
@@ -32,15 +32,16 @@
...
     : Object(prototype)
     , m_datetime(datetime)
     , m_milliseconds(milliseconds)
+    , m_is_invalid(invalid_date)
```
linusg, line 44:
```suggestion
    , m_is_invalid(is_invalid)
```

Userland/Libraries/LibJS/Runtime/Date.h
```diff
@@ -35,9 +35,9 @@ class Date final : public Object {
     JS_OBJECT(Date, Object);
 
 public:
-    static Date* create(GlobalObject&, Core::DateTime, u16 milliseconds);
+    static Date* create(GlobalObject&, Core::DateTime, u16 milliseconds, bool invalid_date = false);
 
-    Date(Core::DateTime datetime, u16 milliseconds, Object& prototype);
+    Date(Core::DateTime datetime, u16 milliseconds, bool invalid_date, Object& prototype);
```
linusg, line 40: Same as above

Userland/Libraries/LibJS/Runtime/Date.h
```diff
@@ -54,6 +54,9 @@ class Date final : public Object {
     double time() const { return datetime().timestamp() * 1000.0 + milliseconds(); }
     int year() const { return datetime().day(); }
 
+    bool is_invalid() const { return m_is_invalid; }
+    void set_invalid(bool value) { m_is_invalid = value; }
```
linusg, line 58:
```suggestion
    void set_is_invalid(bool value) { m_is_invalid = value; }
```

Userland/Libraries/LibJS/Runtime/Date.h
```diff
@@ -68,6 +71,9 @@ class Date final : public Object {
     String time_string() const { return m_datetime.to_string("%T GMT+0000 (UTC)"); }
     String string() const
     {
+        if (is_invalid())
+            return String { "Invalid Date" };
+
```
linusg, line 76: You can use the implicit variant here:

```suggestion
        if (is_invalid())
            return "Invalid Date";
```

Userland/Libraries/LibJS/Runtime/DateConstructor.cpp
```diff
@@ -176,22 +176,43 @@ Value DateConstructor::construct(Function&)
         auto milliseconds = static_cast<u16>(tv.tv_usec / 1000);
         return Date::create(global_object(), datetime, milliseconds);
     }
+
+    auto invalid_arg = [&vm](size_t i) { return vm.argument(i).is_nan() || vm.argument(i).is_undefined(); };
```
linusg, line 180: This is incorrect, the arguments should be checked after being coerced to numbers - e.g. `new Date("")` is not matched by this but returns an invalid date.

Userland/Libraries/LibJS/Runtime/DatePrototype.cpp
```diff
@@ -101,6 +102,11 @@ JS_DEFINE_NATIVE_FUNCTION(DatePrototype::get_date)
     auto* this_object = typed_this(vm, global_object);
     if (!this_object)
         return {};
+
+    if (this_object->is_invalid()) {
+        return Value(NAN);
+    }
+
```
linusg, line 109: - Please use the `js_nan()` function instead of `Value(NAN)`
- Please omit the curly braces for an if-body with only one line

(everywhere in this PR)

Userland/Libraries/LibJS/Runtime/DatePrototype.cpp
```diff
@@ -125,22 +141,48 @@ JS_DEFINE_NATIVE_FUNCTION(DatePrototype::set_full_year)
...
-    // FIXME: Support specifying new month and new day as well.
     auto& datetime = this_object->datetime();
-    auto new_month = datetime.month();
-    auto new_day = datetime.day();
-    datetime.set_time(new_year, new_month, new_day, datetime.hour(), datetime.minute(), datetime.second());
-    return Value { this_object->time() };
+
+    auto arg_or = [&vm, &global_object](size_t i, i32 fallback) { return vm.argument_count() > i ? vm.argument(i).to_i32(global_object) : fallback; };
+
+    auto new_year = arg_or(0, 1970);
```
linusg, line 148: The default value 1970 is not part of the spec :thinking:

https://tc39.es/ecma262/#sec-date.prototype.setfullyear

Userland/Libraries/LibJS/Runtime/DatePrototype.cpp
```diff
@@ -125,22 +141,48 @@ JS_DEFINE_NATIVE_FUNCTION(DatePrototype::set_full_year)
...
+    auto new_day = arg_or(2, datetime.day());
+
+    auto invalid_arg = [&vm](size_t i) { return vm.argument(i).is_nan() || vm.argument(i).is_undefined(); };
+    auto has_invalid_args = [&vm, &invalid_arg]() {
+        bool has_invalid_arg = false;
+        for (size_t i = 0; i < vm.argument_count(); ++i) {
+            has_invalid_arg |= invalid_arg(i);
+        }
+        return has_invalid_arg;
+    };
```
linusg, line 159: Same as before, values must be checked *after* coercion.

Userland/Libraries/LibJS/Runtime/DatePrototype.cpp
```diff
@@ -125,22 +141,48 @@ JS_DEFINE_NATIVE_FUNCTION(DatePrototype::set_full_year)
...
+    } else {
+        datetime.set_time(new_year, new_month, new_day, datetime.hour(), datetime.minute(), datetime.second());
+        this_object->set_invalid(false);
+    }
+
+    if (this_object->is_invalid()) {
+        return Value(NAN);
+    } else {
+        return Value { this_object->time() };
+    }
```
linusg, line 173: No need for else after return

Userland/Libraries/LibJS/Runtime/DatePrototype.cpp
```diff
@@ -271,6 +396,12 @@ JS_DEFINE_NATIVE_FUNCTION(DatePrototype::to_iso_string)
     auto* this_object = typed_this(vm, global_object);
     if (!this_object)
         return {};
+
+    if (this_object->is_invalid()) {
+        vm.throw_exception<RangeError>(global_object, ErrorType::InvalidTimeValue, "Date.toISOString (<anonymous>)");
```
linusg, line 401: The function name should *not* be part of the error message, you've misinterpreted output from NodeJS's stack trace here :smile:

Userland/Libraries/LibJS/Runtime/ErrorTypes.h
```diff
@@ -190,5 +191,4 @@ class ErrorType {
 
     const char* m_message;
 };
-
```
linusg, line 193: Please add this back

Userland/Libraries/LibJS/Tests/builtins/Date/Date.js
```diff
@@ -24,6 +24,11 @@ test("timestamp constructor", () => {
     expect(date.getSeconds()).toBe(1);
     expect(date.getFullYear()).toBe(1970);
     expect(date.getMonth()).toBe(1); // Feb
+
+    let date = new Date(NaN);
+    expect(date.getTime()).toBe(NaN);
+    let date = new Date(undefined);
+    expect(date.getTime()).toBe(NaN);
```
linusg, line 31: Even if LibJS doesn't complain about it, declaring a variable with `let` in in the same scope multiple times is invalid syntax - don't do it (same below and in other tests)

linusg (changes requested):
Nice change, got a bit of feedback below :)

Please prefix the commits with "LibJS" as it's not clear what "Date" refers to otherwise. Also squash the linter fixes and subsequent changes into the appropriate commits as per https://github.com/SerenityOS/serenity/blob/master/CONTRIBUTING.md

Userland/Libraries/LibJS/Runtime/DatePrototype.cpp
```diff
@@ -125,14 +138,32 @@ JS_DEFINE_NATIVE_FUNCTION(DatePrototype::set_full_year)
...
-    auto new_month = datetime.month();
-    auto new_day = datetime.day();
+
+    auto arg_or = [&vm, &global_object](size_t i, i32 fallback) { return vm.argument_count() > i ? vm.argument(i).to_i32(global_object) : fallback; };
+
+    auto new_year = vm.argument(0).to_i32(global_object);
+    auto new_month = arg_or(1, datetime.month());
+    auto new_day = arg_or(2, datetime.day());
+
+    auto invalid_arg = [&vm](size_t i) { return !vm.argument(i).is_finite_number(); };
```
linusg, line 150: same as before, this checks the argument before coercion

linusg (changes requested):
almost there! please don't forget to squash the fixup commits :)

awesomekling:
Please remove the merge commit. Also, when making fixups in response to review comments, please squash those fixups into the relevant preceding commit instead of stacking additional commits.

Userland/Libraries/LibJS/Runtime/DateConstructor.cpp
```diff
@@ -176,25 +176,46 @@ Value DateConstructor::construct(Function&)
...
+            auto datetime = Core::DateTime::from_timestamp(static_cast<time_t>(0));
+            auto milliseconds = static_cast<u16>(0);
+            return Date::create(global_object(), datetime, milliseconds, true);
+        }
+
         // A timestamp since the epoch, in UTC.
-        // FIXME: This doesn't construct an "Invalid Date" object if the argument is NaN.
-        // FIXME: Date() probably should use a double as internal representation, so that NaN arguments and larger offsets are handled correctly.
-        double value_as_double = value.to_double(global_object());
+        double value_as_double = value.as_double();
```
linusg, line 192: This makes the constructor crash, the above checks the result of `value.to_number()`, that doesn't make `value` a number itself (which is required to call `as_double()` on it)

linusg (changes requested):
@leggenda47 my previous comment hasn't been addressed, so I'll just repeat it:

> This is still incorrect, `arg_or()` below calls `to_i32()` on the argument, and since `to_number()` can have side effects we cannot do that twice.
> 
> I suggest you take a look at the spec, they have good pseudo code: [tc39.es/ecma262/#sec-date](tc39.es/ecma262/#sec-date)

Here's a test case to check if this is fixed:

```js
new Date().setFullYear({
    valueOf() {
        console.log("valueOf() has been called");
        return 1970;
    }
});
```

This must log only once.

linusg (approved):
LGTM!

PR #5824, opened 2021-03-16
LibAudio: decrease WavLoader's size limit to a more reasonable size

IdanHo:
A 4 GiB wav (current size limit) is very unreasonable, and larger than oss-fuzz's 2.5 GiB per-process memory limit.
Found as a result of investigating https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=31658 and should fix it.

### DIFF
--- a/Userland/Libraries/LibAudio/WavLoader.cpp
+++ b/Userland/Libraries/LibAudio/WavLoader.cpp
@@ -34,6 +34,8 @@
 
 namespace Audio {
 
+static constexpr size_t maximum_wav_size = 1 * GiB; // FIXME: is there a more appropriate size limit?
+
 WavLoaderPlugin::WavLoaderPlugin(const StringView& path)
     : m_file(Core::File::construct(path))
 {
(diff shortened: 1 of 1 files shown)


Userland/Libraries/LibAudio/WavLoader.cpp
```diff
@@ -34,6 +34,8 @@
 
 namespace Audio {
 
+static constexpr size_t maximum_wav_size = 50 * MiB; // FIXME: is there a more appropriate size limit?
```
awesomekling, line 37: 1 minute of 44100Hz 16-bit WAV data is ~10MB so this feels way too short.

IdanHo: Whats the current per process memory limit? (and would that be an acceptable size limit)

awesomekling: There isn't currently a per-process memory limit, but how about 1GB here? :shrug:

IdanHo: done :+1:

PR #5812, opened 2021-03-15
Ports: Add Hatari Atari ST/STE/TT/Falcon emulator

danboid:

PR #5830, opened 2021-03-16
Kernel: Add _SC_CLK_TCK to sysconf.

thatdutchguy:
Makes patching unnecessary for the Hatari port (see https://github.com/SerenityOS/serenity/pull/5812)


thatdutchguy:
Updated to fix linting issues

Kernel/Syscalls/sysconf.cpp
```diff
@@ -44,6 +44,9 @@ KResultOr<long> Process::sys$sysconf(int name)
         return TTY_NAME_MAX;
     case _SC_GETPW_R_SIZE_MAX:
         return 4096; // idk
+    case _SC_CLK_TCK:
+        // FIXME: should return Number of clock intervals per second for times()
+        return 100; // seems to be 100 on x86_64
```
RealKC, line 49: Not really a kernel programmer, but wouldn't calling https://github.com/SerenityOS/serenity/blob/master/Kernel/Time/TimeManagement.h#L66 be more appropriate?

thatdutchguy: Thanks, that does seem more appropriate, let me give it a go.

thatdutchguy:
That seems to work, and now it returns 250 👍

PR #5843, opened 2021-03-17
UE: Add killpg, getsockname, getpeername, shutdown, sync syscalls

bcoles:
Support for these syscalls allows us to run more `/usr/Tests/` in `UserspaceEmulator`. Most tests can now be run in `ue`, except those which make use of threading and networking.

This also allows us to perform some networking, such as running `nc` in `ue` :

nc -l 0.0.0.0 8000

### DIFF
--- a/Userland/DevTools/UserspaceEmulator/Emulator_syscalls.cpp
+++ b/Userland/DevTools/UserspaceEmulator/Emulator_syscalls.cpp
@@ -192,10 +192,16 @@ u32 Emulator::virt_syscall(u32 function, u32 arg1, u32 arg2, u32 arg3)
         return virt$accept(arg1, arg2, arg3);
     case SC_setsockopt:
         return virt$setsockopt(arg1);
+    case SC_getsockname:
+        return virt$getsockname(arg1);
+    case SC_getpeername:
+        return virt$getpeername(arg1);
     case SC_bind:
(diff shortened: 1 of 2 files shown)


Userland/DevTools/UserspaceEmulator/Emulator_syscalls.cpp
```diff
@@ -689,6 +710,20 @@ int Emulator::virt$getsockopt(FlatPtr params_addr)
...
+    mmu().copy_from_vm(&params, params_addr, sizeof(params));
+    return getsockname(params.sockfd, params.addr, params.addrlen);
+}
+
+int Emulator::virt$getpeername(FlatPtr params_addr)
+{
+    Syscall::SC_getpeername_params params;
+    mmu().copy_from_vm(&params, params_addr, sizeof(params));
+    return getpeername(params.sockfd, params.addr, params.addrlen);
+}
```
awesomekling, line 725: We're not copying the names back into VM memory here.

bcoles: Agreed. That would help.

bcoles: Fixed.

PR #5853, opened 2021-03-18
LibGUI: Support double-click resizing column headers 

thankyouverycool:
keep your columns comfy lads


diversys:
Nice! How about also using middle-clicking for that?

PR #5856, opened 2021-03-18
readelf: Add --checksec flag to display security hardening info

bcoles:
readelf checksec

Based on [checksec](https://github.com/slimm609/checksec.sh) utility.

checksec

### DIFF
--- a/Userland/Utilities/readelf.cpp
+++ b/Userland/Utilities/readelf.cpp
@@ -286,18 +286,20 @@ int main(int argc, char** argv)
     static bool display_relocations = false;
     static bool display_unwind_info = false;
     static bool display_dynamic_section = false;
+    static bool display_hardening = false;
 
     Core::ArgsParser args_parser;
     args_parser.add_option(display_all, "Display all", "all", 'a');
     args_parser.add_option(display_elf_header, "Display ELF header", "file-header", 'h');
     args_parser.add_option(display_program_headers, "Display program headers", "program-headers", 'l');
     args_parser.add_option(display_section_headers, "Display section headers", "section-headers", 'S');
-    args_parser.add_option(display_headers, "Equivalent to: -h -l -S -s -r -d -n -u", "headers", 'e');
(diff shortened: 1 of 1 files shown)


Userland/Utilities/readelf.cpp
```diff
@@ -579,5 +582,68 @@ int main(int argc, char** argv)
...
+
+        if (canary)
+            printf("\033[0;32m%-12s\033[0m ", "Canary found");
+        else
+            printf("\033[0;31m%-12s\033[0m ", "No canary");
+
+        bool nx = false;
+        interpreter_image.for_each_program_header([&nx](const ELF::Image::ProgramHeader& program_header) {
+            if (program_header.type() == PT_GNU_STACK) {
+                if (program_header.flags() == 0x07)
```
bcoles, line 621: Checking for RWX/RWE (`0x07`) is consistent with checksec. There's an [open issue](https://github.com/slimm609/checksec.sh/issues/132) discussing the shortcomings of this approach.

I also need to rewatch the [OS hacking: Stronger W^X memory protection (like PaX MPROTECT) video](https://www.youtube.com/watch?v=7e3qH0HZ1Eo) to check how R^X was implemented on SerenityOS given that we don't have a `GNU_STACK` header.

awesomekling: Stacks are simply never executable on SerenityOS.

bcoles: I've modified this to check only if the stack is executable, regardless of whether it is read/write.

```
if (program_header.flags() & PF_X)
```

This is in line with the ELF validation in `LibELF` which calls shenanigans if the stack is executable.

https://github.com/SerenityOS/serenity/blob/0d8c9024eec12d57270cca1cf6bfc7d03c0ca229/Userland/Libraries/LibELF/Validation.cpp#L293-L298

I've also opened a PR #5867 to build with `--noexecstack` so all ELF files are build with the `GNU_STACK` ELF program header.

PR #5865, opened 2021-03-18
Kernel/AHCI: Prepare for working on real hardware and VMWare

supercomputer7:
Thanks to @tomuta, and his great work in #5184, I'm able to provide interrupt-based solution for the AHCI code, so performance is not so bad and things seem to work on real hardware just fine :)

### DIFF
--- a/Kernel/Storage/AHCIPort.cpp
+++ b/Kernel/Storage/AHCIPort.cpp
@@ -32,6 +32,7 @@
 #include <Kernel/VM/AnonymousVMObject.h>
 #include <Kernel/VM/MemoryManager.h>
 #include <Kernel/VM/TypedMapping.h>
+#include <Kernel/WorkQueue.h>
 
 namespace Kernel {
 
@@ -96,17 +97,49 @@ void AHCIPort::handle_interrupt()
     if (m_interrupt_status.raw_value() == 0) {
         return;
     }
-    if (m_interrupt_status.is_set(AHCI::PortInterruptFlag::PRC) || m_interrupt_status.is_set(AHCI::PortInterruptFlag::PC)) {
-        reset();
-        return;
+    if (m_interrupt_status.is_set(AHCI::PortInterruptFlag::PRC) && m_interrupt_status.is_set(AHCI::PortInterruptFlag::PC)) {
+        clear_sata_error_register();
+        m_wait_connect_for_completion = true;
     }
     if (m_interrupt_status.is_set(AHCI::PortInterruptFlag::INF)) {
         reset();
         return;
     }
     if (m_interrupt_status.is_set(AHCI::PortInterruptFlag::IF) || m_interrupt_status.is_set(AHCI::PortInterruptFlag::TFE) || m_interrupt_status.is_set(AHCI::PortInterruptFlag::HBD) || m_interrupt_status.is_set(AHCI::PortInterruptFlag::HBF)) {
-        recover_from_fatal_error();
+        g_io_work->queue([this]() {
+            recover_from_fatal_error();
+        });
+        return;
     }
+    if (m_interrupt_status.is_set(AHCI::PortInterruptFlag::DHR) || m_interrupt_status.is_set(AHCI::PortInterruptFlag::PS)) {
+        m_wait_for_completion = false;
+
+        // Now schedule reading/writing the buffer as soon as we leave the irq handler.
+        // This is important so that we can safely access the buffers, which could
+        // trigger page faults
+        if (!m_current_request) {
+            dbgln_if(AHCI_DEBUG, "AHCI Port {}: Request handled, probably identify request", representative_port_index());
+        } else {
+            g_io_work->queue([this]() {
+                dbgln_if(AHCI_DEBUG, "AHCI Port {}: Request handled", representative_port_index());
+                LOCKER(m_lock);
+                VERIFY(m_current_request);
+                VERIFY(m_current_scatter_list);
+                if (m_current_request->request_type() == AsyncBlockDeviceRequest::Read) {
+                    if (!m_current_request->write_to_buffer(m_current_request->buffer(), m_current_scatter_list->dma_region().as_ptr(), m_connected_device->block_size() * m_current_request->block_count())) {
+                        dbgln_if(AHCI_DEBUG, "AHCI Port {}: Request failure, memory fault occurred when reading in data.", representative_port_index());
+                        m_current_scatter_list = nullptr;
+                        complete_current_request(AsyncDeviceRequest::MemoryFault);
+                        return;
+                    }
+                }
+                m_current_scatter_list = nullptr;
+                dbgln_if(AHCI_DEBUG, "AHCI Port {}: Request success", representative_port_index());
+                complete_current_request(AsyncDeviceRequest::Success);
+            });
+        }
+    }
+
     m_interrupt_status.clear();
 }
 
@@ -117,8 +150,10 @@ bool AHCIPort::is_interrupts_enabled() const
 
 void AHCIPort::recover_from_fatal_error()
 {
-    ScopedSpinLock lock(m_lock);
+    LOCKER(m_lock);
+    ScopedSpinLock lock(m_hard_lock);
     dmesgln("{}: AHCI Port {} fatal error, shutting down!", m_parent_handler->hba_controller()->pci_address(), representative_port_index());
+    dmesgln("{}: AHCI Port {} fatal error, SError {}", m_parent_handler->hba_controller()->pci_address(), representative_port_index(), (u32)m_port_registers.serr);
     stop_command_list_processing();
     stop_fis_receiving();
     m_interrupt_enable.clear();
@@ -192,8 +227,6 @@ void AHCIPort::eject()
             try_disambiguate_sata_error();
             VERIFY_NOT_REACHED();
         }
-        if ((m_port_registers.ci & (1 << unused_command_header.value())) == 0)
-            break;
     }
     dbgln("AHCI Port {}: Eject Drive", representative_port_index());
     return;
@@ -201,7 +234,8 @@ void AHCIPort::eject()
 
 bool AHCIPort::reset()
 {
-    ScopedSpinLock lock(m_lock);
+    LOCKER(m_lock);
+    ScopedSpinLock lock(m_hard_lock);
 
     dbgln_if(AHCI_DEBUG, "AHCI Port {}: Resetting", representative_port_index());
 
@@ -217,20 +251,21 @@ bool AHCIPort::reset()
     full_memory_barrier();
     clear_sata_error_register();
     full_memory_barrier();
-    if (!initiate_sata_reset()) {
+    if (!initiate_sata_reset(lock)) {
         return false;
     }
-    return initialize();
+    return initialize(lock);
 }
 
 bool AHCIPort::initialize_without_reset()
 {
-    ScopedSpinLock lock(m_lock);
+    LOCKER(m_lock);
+    ScopedSpinLock lock(m_hard_lock);
     dmesgln("AHCI Port {}: {}", representative_port_index(), try_disambiguate_sata_status());
-    return initialize();
+    return initialize(lock);
 }
 
-bool AHCIPort::initialize()
+bool AHCIPort::initialize(ScopedSpinLock<SpinLock<u8>>& main_lock)
 {
     VERIFY(m_lock.is_locked());
     dbgln_if(AHCI_DEBUG, "AHCI Port {}: Initialization. Signature = 0x{:08x}", representative_port_index(), static_cast<u32>(m_port_registers.sig));
@@ -255,7 +290,7 @@ bool AHCIPort::initialize()
     size_t logical_sector_size = 512;
     size_t physical_sector_size = 512;
     u64 max_addressable_sector = 0;
-    if (identify_device()) {
+    if (identify_device(main_lock)) {
         auto identify_block = map_typed<ATAIdentifyBlock>(m_parent_handler->get_identify_metadata_physical_region(m_port_index));
         // Check if word 106 is valid before using it!
         if ((identify_block->physical_sector_size_to_logical_sector_size >> 14) == 1) {
@@ -361,6 +396,7 @@ void AHCIPort::try_disambiguate_sata_error()
 void AHCIPort::rebase()
 {
     VERIFY(m_lock.is_locked());
+    VERIFY(m_hard_lock.is_locked());
     VERIFY(!m_command_list_page.is_null() && !m_fis_receive_page.is_null());
     dbgln_if(AHCI_DEBUG, "AHCI Port {}: Rebasing.", representative_port_index());
     full_memory_barrier();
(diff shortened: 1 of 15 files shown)


supercomputer7:
I've tested these changes with an old 2.5 Inch SATA1 hard drive (capacity is 60GB), and it seems to work quite fine for a hard drive, a little bit slow but still fast enough (booting in about 12 seconds).

Kernel/WorkQueue.h
```diff
@@ -0,0 +1,94 @@
...
+class WorkQueue {
+    AK_MAKE_NONCOPYABLE(WorkQueue);
+    AK_MAKE_NONMOVABLE(WorkQueue);
+
+public:
+    static void initialize();
+
+    WorkQueue(const char*);
+
+    void queue(void (*function)(void*), void* data = nullptr, void (*free_data)(void*) = nullptr)
```
bgianfo, line 45: Instead of using a C API it seems more "serenity like" to queue an object (Maybe called `WorkItem`?) that has an virtual `execute()`/`do_work()` method on it instead? Then callers can just derive their workitem types in a normal C++ fashion?

Kernel/Storage/AHCIPort.cpp
```diff
@@ -704,7 +744,7 @@ void AHCIPort::stop_fis_receiving() const
     m_port_registers.cmd = m_port_registers.cmd & 0xFFFFFFEF;
 }
 
-bool AHCIPort::initiate_sata_reset()
+bool AHCIPort::initiate_sata_reset(ScopedSpinLock<SpinLock<u8>>& scoped_locking)
```
bgianfo, line 747: `scoped_locking` should probably describe the lock / what it's locking instead of the type of the lock?

Kernel/Storage/AHCIPort.cpp
```diff
@@ -609,19 +638,28 @@ bool AHCIPort::identify_device()
...
+                try_disambiguate_sata_error();
+                return false;
+            }
+            if (!m_wait_for_completion)
+                break;
         }
-        if ((m_port_registers.ci & (1 << unused_command_header.value())) == 0)
-            break;
+        scoped_locking.lock();
+        VERIFY(scoped_locking.have_lock());
```
bgianfo, line 660: Why assert something we just did?

Lubrsi:
Fixes AHCI boot on VMware (given if you disable spurious interrupt handler on IRQ 7):
Screenshot from 2021-03-19 01-41-17

Kernel/WorkQueue.h
tomuta: Since work queues are generally used to execute arbitrary code queuing a function or functor (see the overload below this) makes more sense to me. This way you don't have to come up with objects just to be able to execute some code. This allows you to queue lambdas very easily.

supercomputer7: I just made a commit with fixes to `WorkQueue` code. Hopefully it will be enough for merging, but suggestions are always appreciated :)

supercomputer7:
> Fixes AHCI boot on VMware (given if you disable spurious interrupt handler on IRQ 7):
> Screenshot from 2021-03-19 01-41-17

Let's fix it by asking the PCI IRQ line to be something else if it's set by the firmware to line 7 :)

supercomputer7:
As a compromise, if the fimrware decided to set the IRQ line to be 7, or something else we can't deal with,
the user can simply force the code to work with IRQ 11.
For this to happen, set the boot argument `force_ahci_irq_11=on` and then try again.

supercomputer7:
@Lubrsi can you please test this last commit and see if it works for you?

Lubrsi:
It does appear to be setting the IRQ correctly:
```
[34;1m[#0 init_stage2(2:2)][0m: AHCI Port Handler: IRQ 11
[34;1m[#0 init_stage2(2:2)][0m: Trying to unregister unused handler (?)
[34;1m[init_stage2(2:2)][0m: AHCI Port 1: Device detected, Phy enabled
[34;1m[#0 init_stage2(2:2)][0m: AHCI Port Handler: Get identify metadata physical address of port 0 - 0
```

But it hangs after device detection.

supercomputer7:
> It does appear to be setting the IRQ correctly:
> 
> ```
> �[34;1m[#0 init_stage2(2:2)]�[0m: AHCI Port Handler: IRQ 11
> �[34;1m[#0 init_stage2(2:2)]�[0m: Trying to unregister unused handler (?)
> �[34;1m[init_stage2(2:2)]�[0m: AHCI Port 1: Device detected, Phy enabled
> �[34;1m[#0 init_stage2(2:2)]�[0m: AHCI Port Handler: Get identify metadata physical address of port 0 - 0
> ```
> 
> But it hangs after device detection.
> 
> `�[34;1m[#0 init_stage2(2:2)]�[0m: Trying to unregister unused handler (?)` doesn't appear when using the original IRQ:
> 
> ```
> �[34;1m[#0 init_stage2(2:2)]�[0m: AHCI Port Handler: IRQ 7
> �[34;1m[init_stage2(2:2)]�[0m: AHCI Port 1: Device detected, Phy enabled
> �[34;1m[#0 init_stage2(2:2)]�[0m: AHCI Port Handler: Get identify metadata physical address of port 0 - 0
> �[34;1m[#0 init_stage2(2:2)]�[0m: Spurious interrupt, vector 7
> ```

This is normal, as we try to delete the `UnhandledInterruptHandler` and replace it with the new handler.

Kernel/WorkQueue.h
```diff
@@ -42,52 +42,33 @@ class WorkQueue {
...
+    public:
+        static NonnullOwnPtr<WorkItem> create(Function<void(void*)>, Function<void(void*)>, void*);
+        void callback() const;
+        void free_data_callback() const;
+
+    private:
+        WorkItem(Function<void(void*)>, Function<void(void*)>, void*);
+        Function<void(void*)> m_callback;
+        Function<void(void*)> m_free_data_callback;
+        void* m_data { nullptr };
```
tomuta, line 61: I'm not sure what the point of `m_free_data_callback` and `m_data` is if you use `Function<>`. The reason I used these with the c function pointers is simply to support lambdas, and to support them efficiently. And it gives you the flexibility to use lambdas or static function handlers without any allocation overhead if you don't have a need for it. To acoomodate lambdas the data pointer is set to the lambda instance and the free callback is set to the deleter of that object. It also allowed for in-place lambdas to save on unnecessary memory allocations, so if you only capture one pointer such as `this` it could be done with one allocation less.

To be honest I don't really see any advantage here, it's more overhead and not as nice to use. I modeled this after the deferred functions in `Processor`, it's lightweight and supports lambdas just fine. Note that `Function` currently even uses a `RefPtr` internally which makes it even more heavy than it should be.

tomuta (changes requested):

Kernel/Storage/AHCIPort.cpp
```diff
@@ -117,7 +117,7 @@ void AHCIPort::handle_interrupt()
         if (!m_current_request) {
             dbgln_if(AHCI_DEBUG, "AHCI Port {}: Request handled, probably identify request", representative_port_index());
         } else {
-            g_io_work->queue([this]() {
+            g_io_work->queue([this](void*) {
```
tomuta, line 120: This should not be needed, if we use lambdas we should just capture what we need and there should be no need for arguments unless you pass a static free function to it.

tomuta (changes requested):

Kernel/WorkQueue.h
tomuta: Also, I had plans to use a `WorkItem` pool so that we don't it `kmalloc` every time we want to queue something.

supercomputer7: I don't see where `Function` uses a `RefPtr`, but I do see your point, the code should be lightweight if we happen to do lots of IO operations with it. Still, all the `reinterpret_cast`s look suspicious, but maybe we should accommodate them.

tomuta: Sorry, my bad, a `OwnPtr`:
https://github.com/SerenityOS/serenity/blob/0d8c9024eec12d57270cca1cf6bfc7d03c0ca229/AK/Function.h#L117

Either way, that shouldn't really matter for the scope of this PR, it was just one thing I don't like about `Function` and have been meaning to fix it.

Kernel/WorkQueue.h
tomuta: FWIW I prefer the implementation I had before. I purposely added the c-style API for the free function/lambda case that doesn't require data, similar to the deferred call api.

supercomputer7: Let's revert this back then to what it was. If we need to, we can always come up with a better idea what to do :)

supercomputer7:
I also tested this on a new 2TB hard-drive, and it seems to work just fine with it :)
That harddrive is also the first harddrive I encountered that reported a logical sector size of 512 bytes, and physical sector size of 4096 byes, exciting! :)
Still, doing `cat /dev/hda` from the disk asserts due to bad locking, I still need to figure this out.

supercomputer7:
> I also tested this on a new 2TB hard-drive, and it seems to work just fine with it :)
> That harddrive is also the first harddrive I encountered that reported a logical sector size of 512 bytes, and physical sector size of 4096 bytes (so we should start caring about optimizing writes to block boundaries probably), exciting! :)
> Still, doing cat /dev/hda from the disk asserts due to bad locking, I still need to figure this out.

A visual result:

tomuta:
> So what do you suggest to do? I must hold a lock here, but maybe a `SpinLock` is not a good choice?

Well that's kind of the purpose of the `WorkQueue`. You'll have to do some different locking, such as `Lock`, or if the design allows it e.g. pass ownership to the pages to the function that the `WorkQueue` executes. I'm not sure how drivers in Linux do this type of thing, but conceptually whenever you hold a `SpinLock` interrupts are disabled so you can't do anything that could either preempt (e.g. a `Lock`) or page fault in some way while that lock is held.

supercomputer7:
> > So what do you suggest to do? I must hold a lock here, but maybe a `SpinLock` is not a good choice?
> 
> Well that's kind of the purpose of the `WorkQueue`. You'll have to do some different locking, such as `Lock`, or if the design allows it e.g. pass ownership to the pages to the function that the `WorkQueue` executes. I'm not sure how drivers in Linux do this type of thing, but conceptually whenever you hold a `SpinLock` interrupts are disabled so you can't do anything that could either preempt (e.g. a `Lock`) or page fault in some way while that lock is held.

I tried to use `Lock`, and even tried to create a method to pass ownership of the lock if we lock it exclusively to the `WorkQueue` thread later so it can unlock it. It worked, but there's a problem with other locks, because we release other locks and the scheduler just goes on to the next process but we still hold the `AHCIPort` lock so we can't request it to get any data, therefore it just halts.

tomuta:
Passing ownership of a SpinLock lock isn't going to solve this problem, interrupts would still remain disabled if you were to do that. Passing ownership of `Lock` also won't work because `WorkQueue` will run in a different thread, which means releasing the lock won't work (probably trigger a `VERIFY`).

tomuta:
So in this particular instance, do you even need the lock at all? Because `m_current_request` won't change until you complete it (which for performance isn't exactly great if the work queue is being blocked/preempted). So no other request can be started (though they can be queued) until you complete it.

tomuta:
> So in this particular instance, do you even need the lock at all? Because `m_current_request` won't change until you complete it (which for performance isn't exactly great if the work queue is being blocked/preempted). So no other request can be started (though they can be queued) until you complete it.

Which maybe we need another semi-complete state for async requests, some that would allow the next request to be processed while this one is finishing up. Then we could possibly e.g. transfer ownership of the dma page to the `WorkQueue` to finish copying to user (and potentially being preempted doing so), while the next request can be processed in the meanwhile.

supercomputer7:
> > So in this particular instance, do you even need the lock at all? Because `m_current_request` won't change until you complete it (which for performance isn't exactly great if the work queue is being blocked/preempted). So no other request can be started (though they can be queued) until you complete it.
> 
> Maybe we need another semi-complete state for async requests, some that would allow the next request to be processed while this one is finishing up. Then we could possibly e.g. transfer ownership of the dma page to the `WorkQueue` to finish copying to user (and potentially being preempted doing so), while the next request can be processed in the meanwhile.

Maybe, we can create a queue of requests, and if the `m_current_request` was completed and was set to `nullptr`, just get the first from the queue and handle it. Then, from the `WorkQueue` we can just signal for finishing of ending the current request by setting `m_current_request` to `nullptr` and and let another request to be fetched from the queue.

tomuta:
> > > So in this particular instance, do you even need the lock at all? Because `m_current_request` won't change until you complete it (which for performance isn't exactly great if the work queue is being blocked/preempted). So no other request can be started (though they can be queued) until you complete it.
> > 
> > 
> > Maybe we need another semi-complete state for async requests, some that would allow the next request to be processed while this one is finishing up. Then we could possibly e.g. transfer ownership of the dma page to the `WorkQueue` to finish copying to user (and potentially being preempted doing so), while the next request can be processed in the meanwhile.
> 
> Maybe, we can create a queue of requests, and if the `m_current_request` was completed and was set to `nullptr`, just get the first from the queue and handle it. Then, from the `WorkQueue` we can just signal for finishing of ending the current request by setting `m_current_request` to `nullptr` and and let another request to be fetched from the queue.

It already is a queue, but the model is designed to serve device driver hardware requests, but obviously copying data back to user mode usually falls outside of that scope and could be done asynchronously without blocking further requests while that's going on.

supercomputer7:
So, now we can do `cat /dev/hda` without any problems. Still, we may want to improve things before merging this.

supercomputer7:
@tomuta It seems to be ready, I rebased the commits and removed a few. Hopefully it's OK from the SMP aspect.

Kernel/Storage/AHCIPort.cpp
```diff
@@ -609,25 +651,34 @@ bool AHCIPort::identify_device()
...
-    full_memory_barrier();
-
-    while (1) {
-        if (m_port_registers.serr != 0) {
-            dbgln("AHCI Port {}: Identify failed, SError 0x{:08x}", representative_port_index(), (u32)m_port_registers.serr);
-            try_disambiguate_sata_error();
-            return false;
+    // FIXME: Find a better way to send IDENTIFY DEVICE and getting an interrupt!
+    {
+        main_lock.unlock();
```
tomuta, line 656: Since `main_lock.unlock()` is supposed to re-enable interrupts, we should probably `VERIFY_INTERRUPTS_ENABLED` here to make sure this actually happened.

Kernel/Storage/AHCIPort.cpp
```diff
@@ -725,17 +786,23 @@ bool AHCIPort::initiate_sata_reset()
...
-    retry = 0;
-    while (retry < 5000) {
-        if (!((m_port_registers.ssts & 0xf) == 0))
-            break;
-
-        IO::delay(10);
-        retry++;
+    // FIXME: Find a better way to opt-out temporarily from Scoped locking!
+    {
+        main_lock.unlock();
```
tomuta, line 791: Same as above, we should verify that interrupts are indeed enabled.

tomuta (changes requested):

PR #5867, opened 2021-03-19
Build: Enable --noexecstack

bcoles:
Build ELF executables with a zero length `GNU_STACK`
program header flagged non-executable.

The stack is never executable on SerenityOS regardless
of whether the `GNU_STACK` header is specified.

Specifically defining this header is more explicit,
as absence of this header implies an executable stack
on other systems (Linux).

---

This might also be beneficial when executing ELF
executables outside of SerenityOS (Lagom).

Also, checking for a non-executable stack by checking for
a `GNU_STACK` header is common practice. Using this header
lets us show off how securely built our executables are.


awesomekling:
Sure, why not :)

PR #5876, opened 2021-03-19
MouseSettings: add ability to configure double-click speed.

thatdutchguy:
Also adds labels to display the values for to each slider.

Before:
MouseSettings-before

After:
MouseSettings-after

### DIFF
--- a/Userland/Services/WindowServer/WindowManager.cpp
+++ b/Userland/Services/WindowServer/WindowManager.cpp
@@ -172,6 +172,20 @@ void WindowManager::set_scroll_step_size(unsigned step_size)
     m_config->sync();
 }
 
+void WindowManager::set_double_click_speed(int speed)
+{
+    VERIFY(speed >= double_click_speed_min && speed <= double_click_speed_max);
+    m_double_click_speed = speed;
+    dbgln("Saving double-click speed {} to config file at {}", speed, m_config->file_name());
+    m_config->write_entry("Input", "DoubleClickSpeed", String::number(speed));
+    m_config->sync();
+}
+
+int WindowManager::double_click_speed() const
+{
+    return m_double_click_speed;
+}
+
 int WindowManager::scale_factor() const
 {
     return Screen::the().scale_factor();
--- a/Userland/Applications/MouseSettings/CMakeLists.txt
+++ b/Userland/Applications/MouseSettings/CMakeLists.txt
@@ -1,5 +1,10 @@
+compile_gml(MouseSettingsWindow.gml MouseSettingsWindowGML.h mouse_settings_window_gml)
+
 set(SOURCES
     main.cpp
+    MouseSettingsWindow.cpp
+    MouseSettingsWindow.h
+    MouseSettingsWindowGML.h
 )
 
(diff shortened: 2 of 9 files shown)


Userland/Applications/MouseSettings/main.cpp
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2020, Idan Horowitz <idan.horowitz@gmail.com>
+ * Copyright (c) 2020-2021, the SerenityOS developers.
```
ADKaster, line 2: Kinda gauche to remove someone else's copyright, I think you meant to add The SerenityOs developers 2021 below this like?

thatdutchguy: Totally - I did not mean to stomp over that!

thatdutchguy: Updated the license files so credit is given where due.

linusg: Please drop the trailing commas though, this is how it's suppose to look:

https://github.com/SerenityOS/serenity/blob/0a911178cea4fcfb462eb7e256c51ed75756c7c8/Userland/Libraries/LibJS/Runtime/ArrayPrototype.cpp#L2-L4

thatdutchguy: What about the single trailing dot at the end of "the SerenityOS deverlopers."? I've noticed that it's present most of the time.

thatdutchguy: Dropped the dot as well, it's now following this style:
https://github.com/SerenityOS/serenity/blob/0a911178cea4fcfb462eb7e256c51ed75756c7c8/Userland/Libraries/LibCompress/Deflate.h#L2-L3

thatdutchguy:
Noticed some small UI update issues that I will address when I get a chance later today.

thatdutchguy:
All good & rebased on top of master

Userland/Services/WindowServer/WindowManager.cpp
```diff
@@ -172,6 +172,20 @@ void WindowManager::set_scroll_step_size(unsigned step_size)
     m_config->sync();
 }
 
+void WindowManager::set_double_click_speed(int speed)
+{
+    VERIFY(speed >= double_click_speed_min && speed <= double_click_speed_max);
```
awesomekling, line 177: This allows anyone to crash the window server by connecting to it and setting an out-of-range doubleclick speed.
Instead of `VERIFY`, we should validate the value and reject invalid ones.

awesomekling: Oh wait nevermind, you already do this in the ClientConnection handler :)

awesomekling:
This seems fine, but the commit message is deceptively small: you are adding new IPC API to WindowServer  yet the message only mentions the MouseSettings app. Please make a little more effort to explain what the commit is doing :)

thatdutchguy:
Updated commit message. Funny, before I squashed it it had some pretty lengthy comments which seemed a little verbose. At the time it seemed so obvious ;-)

PR #5878, opened 2021-03-19
AK+LibAudio+AudioPlayer: Added a complex number library, revised type names in LibAudio and extended AudioPlayer

ngc6302h:
Hello!

Updated the audio player a bit, it needed some love. I added several new features to it, and in the process updated LibM and LibAudio to make this possible.
AudioPlayer still a work-in-progress, and the audio loading & feeding process to AudioServer is in dire need of a rewrite if we want it to really be extensible and augmentable for more features and audio formats. I'll be working on this but if someone wants to do it first, go ahead. I'd also like to eventually turn it into a media player, not just an audio player :D

image


Userland/Applications/SoundPlayer/AudioAlgorithms.cpp
```diff
@@ -0,0 +1,78 @@
...
+// of the smallest frequency to be resolved.
+//
+// For example, to resolve a 10 KHz and a 2 Hz sine waves we need at least
+// a samplerate of 20 KHz and a window of 0.5 seconds
+//
+// If invert is true, this function computes the inverse discrete fourier transform.
+//
+// The data vector must be a power of 2
+// Adapted from https://cp-algorithms.com/algebra/fft.html
+void fft(complex<double>* data, int sample_count, bool invert)
```
alimpfard, line 46: I feel like this should be in a more accessible place, since its only use-case isn't audio processing...but maybe it's okay for now?

Either way though, it's a bit sad to have a C-style API for it, maybe use `Span<const complex<double>>`, or just take the vector by const reference?

Userland/Applications/SoundPlayer/BarsVisualizationWidget.cpp
```diff
@@ -0,0 +1,147 @@
...
+    painter.fill_rect(frame_inner_rect(), Color::Black);
+
+    if (m_sample_buffer.is_null() || m_sample_buffer.is_empty())
+        return;
+
+    fft(m_sample_buffer.data(), m_sample_count, false);
+    double max = sqrt(m_sample_count * 2);
+
+    //TODO: don't hardcode this!
+    double freqBin = 44100 / m_sample_count;
```
alimpfard, line 52: please no camelCase in C++ code.

Userland/Applications/SoundPlayer/Common.h
```diff
@@ -0,0 +1,55 @@
...
+ */
+
+#pragma once
+
+#include <LibGUI/Slider.h>
+
+class Slider final : public GUI::Slider {
+    C_OBJECT(Slider)
+public:
+    ~Slider() override {};
```
alimpfard, line 34: `= default;` ?

Userland/Libraries/LibM/complex.h
```diff
@@ -0,0 +1,307 @@
...
+{
+    const auto x = const_cast<complex<T>&>(a) - const_cast<complex<U>&>(b);
+    return x.srqMagnitude() <= margin;
+}
+
+//complex version of exp()
+template<typename T>
+requires AK::IsArithmetic<T>::value static constexpr complex<T> exp(const complex<T>& a)
+{
+    // this can probably be faster and not use expensive trigonometric functions
```
alimpfard, line 301: `FIXME` this, maybe?

Userland/Applications/SoundPlayer/AudioAlgorithms.cpp
ngc6302h: You're right. I put it there because didn't want to extend the scope of the PR too much, but it has much more applications and it'd be worth moving it. I accept suggestions about where to move it :)

alimpfard: I think the placement is okay for now, we can move it to a common library when we need it elsewhere.

ngc6302h: Alright. About the fft using a raw pointer, you're right in your comment. I'll make it more cppish.

devsh0:
This is great. I so wanted these features in the SoundPlayer so started working on these off serenity here: https://github.com/devsh0/fft-playground

...but something went wrong with sound when I rebased to latest master. I am very happy that you worked on this.

tomuta:
> ...but something went wrong with sound when I rebased to latest master. I am very happy that you worked on this.

Our SB16 driver doesn't implement playback properly, it's using a mechanism that is almost guaranteed to produce glitches and/or crackling as that mechanism is documented to do so. Not sure if that's what you're referring to, but currently there is no driver that can produce a smooth audio stream.

devsh0:
Yeah, Kling mentioned that but that wasn't it.  I wasn't able to play any sound with latest QEMU build. First I thought something was wrong with my build but I tried a linux mint image (32-bit) with -soundhw ac97 and I was able to play sound with it. OTOH, using QEMU 4.2.1 or earlier releases only played noise.

Userland/Libraries/LibM/complex.h
```diff
@@ -0,0 +1,307 @@
...
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#pragma once
+
+#include <AK/Concepts.h>
+#include <math.h>
+
+template<typename T>
+requires AK::IsArithmetic<T>::value class [[gnu::packed]] complex {
```
RealKC, line 33: The Serenity coding style says to use CamelCase(first letter is capital) names for classes, i.e. this should be called `Complex`.

Also I think clang-format is formatting the `requires` clause really weird, maybe it'd be worth adding an `Arithmetic` concept to `AK/Concepts.h` so we could write `template<Arithmetic T>` instead? Something like
```cpp
template<typename T>
concept Arithmetic = IsArithmetic<T>::value;
```
would work, I think.

Finally there's no reason to qualify things from AK with the `AK` namespace

Userland/Libraries/LibM/complex.h
```diff
@@ -0,0 +1,307 @@
...
+    }
+    constexpr complex(T real, T imaginary)
+        : m_real(real)
+        , m_imag(imaginary)
+    {
+    }
+
+    constexpr T real() const NOEXCEPT { return m_real; }
+    constexpr T imag() const NOEXCEPT { return m_imag; }
+    constexpr T srqMagnitude() const NOEXCEPT { return m_real * m_real + m_imag * m_imag; }
```
RealKC, line 53: I think magnitude_squared would be a better name for this

RealKC (changes requested):

RealKC:
On a different note, the C standard does actually define a headers called `complex.h` for its own set of complex numbers arithmetic, and while I'm not opposed to us putting our complex number class in it, I think it should be fenced by `#ifdef __cplusplus ... #endif`. An alternative would be putting it in a header called `Complex.h`, but I'm not sure I like the only difference between those two headers being capitalization

Userland/Libraries/LibM/complex.h
ngc6302h: I put it in lowercase in concordance with other types like the ones under AK::SIMD. I think it's worth making an exception for the complex number type, since having it look like as if it was a native type I think makes reading and comprehending algorithms easier, since you don't have to think that it is another class but just a complex number.
If it can not be avoided then I'll make it camelcase.

RealKC: That makes sense, in that case, I think it's fine.

ngc6302h:
> On a different note, the C standard does actually define a headers called `complex.h` for its own set of complex numbers arithmetic, and while I'm not opposed to us putting our complex number class in it, I think it should be fenced by `#ifdef __cplusplus ... #endif`. An alternative would be putting it in a header called `Complex.h`, but I'm not sure I like the only difference between those two headers being capitalization
I'm gonna go with #ifdef for now.

awesomekling:
Looks cool. Some notes:

Please don't add `LibM/complex.h` unless you are implementing the standards compliant `complex.h` as @RealKC mentions.

This would fit better as `AK/Complex.h`. Also it should be named `Complex`, not `complex`.

The patch with "add lots of features" is extremely crowded. Please make an effort to separate features into separate commits :)

ngc6302h:
>Please don't add LibM/complex.h unless you are implementing the standards compliant complex.h as @RealKC mentions.


Not really thinking about implementing the standard now so I'll move it to AK.

>Also it should be named Complex, not complex.

I answered that here https://github.com/SerenityOS/serenity/pull/5878#discussion_r598119342 but if someone's eventually going to implement the C standard one then I agree to make it CamelCase.

awesomekling:
> > Also it should be named Complex, not complex.
> 
> I answered that here [#5878 (comment)](https://github.com/SerenityOS/serenity/pull/5878#discussion_r598119342) but if someone's eventually going to implement the C standard one then I agree to make it CamelCase.

The SIMD types are primitive types, this is a class. Classes are PascalCase.

> > The patch with "add lots of features" is extremely crowded. Please make an effort to separate features into separate commits :)
> 
> Working on it rn :D

👍🏻

ngc6302h:
An update of what it's done and what it's left:
- [x] abstract the player functionality from the gui (not pushed yet)
- [ ] an M3U(8) parser (in progress)
- [ ] playlist support (in progress)
- [ ] mp3 decoder (to be documented)
- [ ] mkv/webm parser (in progress)
- [ ] av1 decoder (in progress)
- [ ] h264 decoder (to be documented)
- [ ] rewrite and extend playbackmanager

ngc6302h:
I think the player is a reasonable better state now. Since the original goals of this PR are fulfilled, I think further work on https://github.com/SerenityOS/serenity/pull/5878#issuecomment-805431667 should be done on future PRs.
image

Userland/Applications/SoundPlayer/M3UParser.cpp
```diff
@@ -0,0 +1,130 @@
...
+    VERIFY(file != nullptr);
+    fseek(file, 0, SEEK_END);
+    size_t file_size = ftell(file);
+    fseek(file, 0, SEEK_SET);
+    VERIFY(file_size > 0);
+    Vector<u8> temp_buffer;
+    temp_buffer.resize(file_size);
+    auto bytes_read = fread(temp_buffer.data(), 1, file_size, file);
+    VERIFY(bytes_read == file_size);
+    parser->m_m3u_raw_data = *new String(temp_buffer.span(), NoChomp);
```
Dexesttp, line 55: I _feel_ like you should be able to use [a `Core::File` object](https://github.com/SerenityOS/serenity/blob/master/Userland/Libraries/LibCore/File.h) here, but it doesn't seem like it has an easy `read_all_to_string() -> String` method...
Maybe there's something that could be done with [`Core::File::read_all() -> ByteBuffer`](https://github.com/SerenityOS/serenity/blob/master/Userland/Libraries/LibCore/IODevice.h#L84) ?

Userland/Applications/SoundPlayer/PlaylistWidget.cpp
```diff
@@ -0,0 +1,127 @@
...
+        case 4:
+            return m_playlist_items[index.row()].extended_info->album_artist.value_or("");
+        case 5:
+            return format_filesize(m_playlist_items[index.row()].extended_info->file_size_in_bytes.value_or(0));
+        }
+    }
+    if (role == GUI::ModelRole::Sort)
+        return data(index, GUI::ModelRole::Display);
+    if ((int)role == 0x101) //path
+        return m_playlist_items[index.row()].path;
```
Dexesttp, line 75: That feels a bit weeird, hardcoded like that. [`AutocompleteProvider.cpp`](https://github.com/SerenityOS/serenity/blob/master/Userland/Libraries/LibGUI/AutocompleteProvider.cpp#L84) seems to be the only other file that uses extended roles, and it still adds [a new `InternalRoles` enum](https://github.com/SerenityOS/serenity/blob/master/Userland/Libraries/LibGUI/AutocompleteProvider.cpp#L52). How about doing something like this ?

Userland/Applications/SoundPlayer/M3UParser.cpp
```diff
@@ -0,0 +1,130 @@
...
+                //TODO: Support M3A files (M3U files with embedded mp3 files)
+            } else {
+                entry.path = line;
+                entry.extended_info = metadata_for_next_file;
+                vec->append(entry);
+            }
+        }
+    } else {
+        //TODO: Implement M3U8 parsing
+        VERIFY_NOT_REACHED();
```
Dexesttp, line 127: Nit: this could be a [`TODO();`](https://github.com/SerenityOS/serenity/blob/master/AK/Assertions.h#L36) instead of a `VERIFY_NOT_REACHED();`. It seems more contextually appropriate :^)

Userland/Applications/SoundPlayer/BarsVisualizationWidget.cpp
```diff
@@ -0,0 +1,151 @@
...
+// black magic from Hacker's delight
+u32 round_previous_power_of_2(u32 x)
+{
+    x = x | (x >> 1);
+    x = x | (x >> 2);
+    x = x | (x >> 4);
+    x = x | (x >> 8);
+    x = x | (x >> 16);
+    return x - (x >> 1);
+}
```
Dexesttp, line 151: How about making this either a local method for this .cpp file, or moving it into AudioAlgorithms.h ?

Userland/Applications/SoundPlayer/BarsVisualizationWidget.cpp
```diff
@@ -0,0 +1,151 @@
...
+        m_adjust_frequencies = action.is_checked();
+    }));
+}
+
+void BarsVisualizationWidget::set_buffer(RefPtr<Audio::Buffer> buffer, int samples_to_use)
+{
+    if (m_is_using_last)
+        return;
+    m_is_using_last = true;
+    // should probably do a check to see if the buffer contains a meaningful number of samples
```
Dexesttp, line 109: Given that `buffer->samples()[i]` below will crash if it doesn't have enough samples, then you might want to VERIFY(); that right away indeed !

Dexesttp (commented):
I really like this code, it's pretty simple and easy to follow ! Great work :)

These are mostly small things that shouldn't change your code too much, except for the `Core::File` and `String => ByteBuffer` thing, so feel free to ignore this one if you can't make it work.

A small style-related nitpick is that you use both the `//comment` and `// comment` syntax, but that's already true for the codebase as a whole (even though `// comment` seems to be used more often), so it might be okay to leave it all like this.

Userland/Applications/SoundPlayer/M3UParser.cpp
ngc6302h: Good point, it makes the code shorter too.

Userland/Applications/SoundPlayer/PlaylistWidget.cpp
ngc6302h: Yeah, fixed :)

Userland/Applications/SoundPlayer/M3UParser.cpp
ngc6302h: That's a nice alias for contextual info :+1:

Userland/Applications/SoundPlayer/BarsVisualizationWidget.cpp
ngc6302h: Yeah it didn't really make sense for it to be a member function. Made it a local func.

Userland/Applications/SoundPlayer/BarsVisualizationWidget.cpp
ngc6302h: Done

bcoles:
This PR reverts #4230 by removing the max 100% playback volume level restriction. Allowing 100%+ makes sense but there should be an upper limit.

Issue #5881, opened 2021-03-20
LibGUI: Button doesn't show depressed state when quickly pressing it many times per second


tomuta:
I noticed this, same happens in every application, including the widget gallery. Notice how the tips update every time I press the button, but the button seems to only show the depressed state once every N number of clicks (or so it seems).

~~Might be related to me clicking fast enough for it to consider it a double-click.~~

![button](https://user-images.githubusercontent.com/10320822/111856890-78337200-88f3-11eb-9e5a-bb905ecb33de.gif)

Added some log statements, seems like by the time the paint event is dispatched the mouse-up event had already been dispatched:
```
Serendipity(31): down: 0x04684818
Serendipity(31): button paint 0x04684818 pressed: true
Serendipity(31): up: 0x04684818
Serendipity(31): button paint 0x04684818 pressed: false
Serendipity(31): down: 0x04684818
Serendipity(31): up: 0x04684818
Serendipity(31): button paint 0x04684818 pressed: false
Serendipity(31): down: 0x04684818
Serendipity(31): up: 0x04684818
Serendipity(31): button paint 0x04684818 pressed: false
Serendipity(31): button paint 0x04684818 pressed: false
```

thatdutchguy:
I currently cannot reproduce this, no matter what delay I set for DoubleClickSpeed.

![button_clicks](https://user-images.githubusercontent.com/209150/111917642-6e6a5580-8a3e-11eb-9e4e-4936d08b6a2b.gif)

tomuta:
> I currently cannot reproduce this, no matter what delay I set for DoubleClickSpeed.
> 
> ![button_clicks](https://user-images.githubusercontent.com/209150/111917642-6e6a5580-8a3e-11eb-9e4e-4936d08b6a2b.gif)

It seemed related to double clicks, but I verified that it has nothing to do with it.

GMTA:
So I'm fairly certain that this issue is caused by too many mouse events - this happens with an unfinished double-click and scroll wheel usage. I'm not seeing issues with our code though - it seems like Qemu is generating superfluous mouse events somewhere in its input stack. If I add debugging statements to `ui/input-legacy.c`, I see the same (incorrect) amount of mouse events.

PR #5882, opened 2021-03-20
LibGUI: Process pending window updates after handling input events

tomuta:
Since input events may trigger window portions to be invalidated,
rather than making a round trip to WindowServer to get paint events
we can simply fake an immediate paint event and update the window
contents more quickly.

Improves #5881

_Not sure I like this solution, but it makes sense to more rapidly trigger paint events after processing input events if they may require the screen to be updated. It's not perfect, though._


awesomekling:
This feels like a highly specific band-aid and less like a general improvement.. I'm not sure what would make a general improvement though.

tomuta:
> This feels more like a highly specific band-aid and less like a general improvement.. I'm not sure what would make a general improvement though.

We could do it after all events. For input events latency is more important, so it makes sense to process pending paint events right away. This cuts out a round trip to WindowServer for those window invalidations. I believe in Windows they give threads that have GUI input higher priority, but it's also a quite different architecture iirc.

Issue #5055, opened 2021-01-22
WindowServer: cursor alpha channel issue


TheMorc:
I found out after tinkering with SDL2 apps that the mouse cursor at 100% has some kind of alpha channel problem (?)
![Screen Shot 2021-01-22 at 22 01 35](https://user-images.githubusercontent.com/13377926/105546478-bbab9e80-5cfd-11eb-878b-e807c5286225.png)

But interestingly, it doesn't happen if you use 200% scale.
![Screen Shot 2021-01-22 at 22 01 22](https://user-images.githubusercontent.com/13377926/105546526-c9f9ba80-5cfd-11eb-8e43-22684be7706c.png)

And iirc, it also happened the other way on some site, the alpha channel of the png picture there was black and the cursor was "exposing" it so you could see through the cursor shadow to the site

Lubrsi:
This also happens with window shadows, and also isn't a problem at 200% scale:

![Screenshot from 2021-02-13 23-21-30](https://user-images.githubusercontent.com/25595356/107864252-404b9180-6e52-11eb-950f-ac958c2766e4.png)

tomuta:
@Lubrsi I'm running into the same issues with some changes I'm working on in `Painter::blit`. It seems like one problem may be that it doesn't seem to respect the bitmap's `StorageFormat`. I'm still not entirely sure I understand the code but it sure seems like it's taking some shortcuts it shouldn't be taking.

linusg:
Fixed in #5888.

PR #5888, opened 2021-03-20
LibGfx: Don't use target's alpha if it doesn't have one

olegv11:
Fixes the white column when rendering the cursor in a few places, like LibGfxDemo. Might be related to #5055.
Before:
bug1
After:
bug2

Here's how the bug manifested:
The background (BGRx) is set to {0, 0, 0, 0} and is read as BGRA. The last column of pixels in cursor's png file (BGRA) is set to {255, 255, 255, 0}. After blend you get {255, 255, 255, 0}, which is correct - but then it gets written as is to BGRx8888 bitmap, basically losing all information about alpha and making the pixel pure white.

Currently this PR also breaks the rendering of SDL applications like pt2-clone, so I'll mark it as a draft until I find out why.

### DIFF
--- a/Userland/Libraries/LibGfx/Painter.cpp
+++ b/Userland/Libraries/LibGfx/Painter.cpp
@@ -565,28 +565,37 @@ void Painter::blit_with_opacity(const IntPoint& position, const Gfx::Bitmap& sou
     const RGBA32* src = source.scanline(src_rect.top() + first_row) + src_rect.left() + first_column;
     const unsigned src_skip = source.pitch() / sizeof(RGBA32);
 
-    auto do_blit = [&]<bool with_src_alpha>() {
+    auto do_blit = [&]<bool with_src_alpha, bool with_dst_alpha>() {
         for (int row = first_row; row <= last_row; ++row) {
             for (int x = 0; x <= (last_column - first_column); ++x) {
+                Color dest_color = with_dst_alpha ? Color::from_rgba(dst[x]) : Color::from_rgb(dst[x]);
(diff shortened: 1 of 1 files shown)


Userland/Libraries/LibGfx/Painter.cpp
```diff
@@ -565,28 +565,37 @@ void Painter::blit_with_opacity(const IntPoint& position, const Gfx::Bitmap& sou
...
+        if (m_target->has_alpha_channel())
+            do_blit.template operator()<true, true>();
+        else
+            do_blit.template operator()<true, false>();
+    } else {
+        if (m_target->has_alpha_channel())
+            do_blit.template operator()<false, true>();
+        else
+            do_blit.template operator()<false, false>();
+    }
```
alimpfard, line 598: These look painful to read...maybe take locally defined enums instead of booleans?
Or maybe also extract it to a normal member/helper function and also get rid of `template operator()`?

olegv11:
Alright, the reason was that SDL backend was creating a bitmap with alpha without telling (pushed a PR there). Now I can confirm that this commit does in fact fix #5055:

bug3

Userland/Libraries/LibGfx/Painter.cpp
olegv11: Alright, extracted it to a helper function.

linusg:
Nitpick: it's LibGfx.

olegv11:
Thanks, @linusg. Reworded commit messages.

PR #5895, opened 2021-03-21
readelf: Use Loader.so interpreter for dynamic section and symbols

bcoles:
readelf checksec

A few `FIXME` comments:

* `init` / `fini` / `start` / `main` entries are not displayed in dynamic symbols output.
* ~~`RPATH` and `RUNPATH` entry offset is printed (not the value).~~
* The dynamic section entry count is incorrect (off by +4). The `ELF_DYNAMIC` section entry count is populated by `LibELF` (`ELF::Image`).

`ELF::DynamicObject` also drops a final `NULL` (`0x00`) dynamic entry from the dynamic section for some reason.

image

Compare with GNU readelf on Ubuntu:

```
user@ubuntu:~/Desktop/serenity$ readelf --dyn-syms Build/Root/bin/tty

Symbol table '.dynsym' contains 15 entries:
   Num:    Value  Size Type    Bind   Vis      Ndx Name
     0: 00000000     0 NOTYPE  LOCAL  DEFAULT  UND 
     1: 00000000     0 OBJECT  GLOBAL DEFAULT  UND __stack_chk_guard
     2: 00000000     0 FUNC    GLOBAL DEFAULT  UND perror
     3: 00000000     0 FUNC    GLOBAL DEFAULT  UND puts
     4: 00000000     0 FUNC    WEAK   DEFAULT  UND __cxa_finalize
     5: 00000000     0 FUNC    WEAK   DEFAULT  UND __deregister_frame_info
     6: 00000000     0 FUNC    WEAK   DEFAULT  UND _ITM_registerTMCloneTable
     7: 00000000     0 FUNC    WEAK   DEFAULT  UND _ITM_deregisterTMCloneTab
     8: 00000000     0 FUNC    GLOBAL DEFAULT  UND ttyname
     9: 00000000     0 FUNC    GLOBAL DEFAULT  UND arc4random_buf
    10: 00000000     0 FUNC    WEAK   DEFAULT  UND __register_frame_info
    11: 000003b0     0 FUNC    GLOBAL DEFAULT    7 _init
    12: 00000490    79 FUNC    GLOBAL DEFAULT   10 _start
    13: 00000430    92 FUNC    GLOBAL DEFAULT   10 main
    14: 000006e1     0 FUNC    GLOBAL DEFAULT   11 _fini
```

```
user@ubuntu:~/Desktop/serenity$ readelf -d Build/Root/bin/tty

Dynamic section at offset 0xedc contains 26 entries:
  Tag        Type                         Name/Value
 0x00000001 (NEEDED)                     Shared library: [libgcc_s.so]
 0x00000001 (NEEDED)                     Shared library: [libcore.so]
 0x00000001 (NEEDED)                     Shared library: [libcrypt.so]
 0x00000001 (NEEDED)                     Shared library: [libcrypto.so]
 0x00000001 (NEEDED)                     Shared library: [libc.so]
 0x00000001 (NEEDED)                     Shared library: [libsystem.so]
 0x00000001 (NEEDED)                     Shared library: [libm.so]
 0x0000000c (INIT)                       0x3b0
 0x0000000d (FINI)                       0x6e1
 0x6ffffef5 (GNU_HASH)                   0x128
 0x00000005 (STRTAB)                     0x248
 0x00000006 (SYMTAB)                     0x158
 0x0000000a (STRSZ)                      270 (bytes)
 0x0000000b (SYMENT)                     16 (bytes)
 0x00000015 (DEBUG)                      0x0
 0x00000003 (PLTGOT)                     0x1fcc
 0x00000002 (PLTRELSZ)                   32 (bytes)
 0x00000014 (PLTREL)                     REL
 0x00000017 (JMPREL)                     0x390
 0x00000011 (REL)                        0x358
 0x00000012 (RELSZ)                      56 (bytes)
 0x00000013 (RELENT)                     8 (bytes)
 0x00000018 (BIND_NOW)                   
 0x6ffffffb (FLAGS_1)                    Flags: NOW PIE
 0x6ffffffa (RELCOUNT)                   1
 0x00000000 (NULL)                       0x0
```

### DIFF
--- a/Userland/Utilities/readelf.cpp
+++ b/Userland/Utilities/readelf.cpp
@@ -29,6 +29,9 @@
 #include <AK/StringBuilder.h>
 #include <AK/StringView.h>
 #include <LibCore/ArgsParser.h>
+#include <LibCore/File.h>
+#include <LibELF/DynamicLoader.h>
+#include <LibELF/DynamicObject.h>
 #include <LibELF/Image.h>
 #include <LibELF/Validation.h>
 #include <ctype.h>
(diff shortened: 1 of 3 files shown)


ADKaster:
> ELF::DynamicObject also drops a final NULL (0x00) dynamic entry from the dynamic section for some reason.

This one is straightforward, in for_each_dynamic_entry we check for DT_NULL to break out, but don't call func() on that entry. I think it's that way because the actual loading code will never care about the null entry.

Userland/Utilities/readelf.cpp
```diff
@@ -450,23 +578,49 @@ int main(int argc, char** argv)
...
+            object->for_each_dynamic_entry([&library_index, &libraries](const ELF::DynamicObject::DynamicEntry& entry) {
+                printf("  0x%08x ", entry.tag());
+                printf("%-17s ", object_tag_to_string(entry.tag()));
+
+                if (entry.tag() == DT_NEEDED) {
+                    printf("Shared library: %s\n", String(libraries[library_index]).characters());
+                    library_index++;
+                } else if (entry.tag() == DT_RPATH) {
+                    // FIXME: print the rpath
+                    printf("Library rpath: 0x%08x\n", entry.val());
```
ADKaster, line 614: Should also be able to print SONAME here-- there's a helper on Dynamic object to get it from the .dynstr section. The same strategy for soname() could be used for runpath/rpath.

bcoles:
> This one is straightforward, in for_each_dynamic_entry we check for DT_NULL to break out, but don't call func() on that entry. I think it's that way because the actual loading code will never care about the null entry.

Seems like that should be patched in LibELF (if at all). I'd rather not add hard coded presumptions to `readelf`, especially when `readelf` will (ideally) be able to parse any ELF file regardless of whether it was compiled for Serenity.

Userland/Utilities/readelf.cpp
bcoles: I've added LibELF helpers for `rpath` and `runpath` and added these and `soname` to readelf for now.

The helpers will lie in the event of duplicate entries.

Using helpers isn't ideal. Eventually `readelf` should resolve entries directly via offset rather than relying on helpers. There's still a lot work to be done on LibELF.

PR #5897, opened 2021-03-21
LibCompress: Convert GzipDecompressor from recursive to iterative

IdanHo:
This way a gzip compressed file that contains a large amount of small blocks wont cause a stack overflow.
(This is similar to #5872 and found by/fixes https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=32298)
Sorry for the large amount of LibCompress deflate/gzip decompresion fixes lately, looks like our existing decompressors werent as robust as i'd like them to be, but all of these fixes should help with that :)

### DIFF
--- a/Userland/Libraries/LibCompress/Gzip.cpp
+++ b/Userland/Libraries/LibCompress/Gzip.cpp
@@ -70,86 +70,94 @@ GzipDecompressor::~GzipDecompressor()
 // FIXME: Again, there are surely a ton of bugs because the code doesn't check for read errors.
 size_t GzipDecompressor::read(Bytes bytes)
 {
-    if (has_any_error() || m_eof)
-        return 0;
+    size_t total_read = 0;
+    while (total_read < bytes.size()) {
+        if (has_any_error() || m_eof)
+            break;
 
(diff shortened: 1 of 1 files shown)


asynts:
> [...]  looks like our existing decompressors werent as robust as i'd like them to be [...]

That's my fault, I kinda just started working on it without much knowledge about the topic. When I finally discovered https://www.hanshq.net/zip.html I had already lost interest. Good that someone picked that up, otherwise, I would have at some point.

IdanHo:
> > [...]  looks like our existing decompressors werent as robust as i'd like them to be [...]
> 
> That's my fault, I kinda just started working on it without much knowledge about the topic. When I finally discovered https://www.hanshq.net/zip.html I had already lost interest. Good that someone picked that up, otherwise, I would have at some point.

Not blaming you at all! The reason i even started working on the compression side was seeing your decompression implementation, it made the idea much less daunting :)

awesomekling:
> > [...]  looks like our existing decompressors werent as robust as i'd like them to be [...]
> 
> That's my fault, I kinda just started working on it without much knowledge about the topic. 

Much of this system is built that way <3

Issue #3549, opened 2020-09-19
LibJS: m_ptr[offset] >> 6 == 2 assertion failed in Utf8CodepointIterator in less than comparison


Lubrsi:
Found whilst fuzzing
[crash.txt](https://github.com/SerenityOS/serenity/files/5249012/crash.txt)

Trace:
```
USERSPACE(24) ASSERTION FAILED: m_ptr[offset] >> 6 == 2
../AK/Utf8View.cpp:210
[js(24:24)]: Terminating js(24) due to signal 6
[FinalizerTask(3:3)]: CPU[0] NP(error) fault at invalid address V0x01801d9c
[FinalizerTask(3:3)]: copy_from_user(0xc094a9ac, 0x01801d9c, 4) failed at V0x01801d9c
[FinalizerTask(3:3)]: 0xdeadc0de  Kernel::Processor::switch_context(Kernel::Thread*&, Kernel::Thread*&) +181
0xdeadc0de  Kernel::Scheduler::context_switch(Kernel::Thread*) +183
0xdeadc0de  Kernel::Scheduler::pick_next() +1593
0xdeadc0de  Kernel::Processor::check_invoke_scheduler() +125
0xdeadc0de  Kernel::Thread::die_if_needed() +157
0xdeadc0de  syscall_handler +1414
0xdeadc0de  syscall_asm_entry +49
0x080baa78  raise +22
0x080480bd  abort +13
0x080b6acb  tolower +0
0x080d5468  AK::Utf8CodepointIterator::operator*() const +180
0x0809384a  JS::abstract_relation(JS::Interpreter&, bool, JS::Value, JS::Value) +591
0x08093dd9  JS::less_than(JS::Interpreter&, JS::Value, JS::Value) +49
0x0805a0ce  JS::BinaryExpression::execute(JS::Interpreter&, JS::GlobalObject&) const +644
0x08059e81  JS::BinaryExpression::execute(JS::Interpreter&, JS::GlobalObject&) const +55
0x08059c3a  JS::ExpressionStatement::execute(JS::Interpreter&, JS::GlobalObject&) const +44
0x08066716  JS::Interpreter::execute_statement(JS::GlobalObject&, JS::Statement const&, AK::Vector<JS::Argument, 8ul>, JS::ScopeType) +414
0x0805c120  JS::ScopeNode::execute(JS::Interpreter&, JS::GlobalObject&) const +64
[...]

PR #5901, opened 2021-03-21
LibJS: Handle overlong encoded utf8 sequences in abstract relations

IdanHo:
This adds Utf8View::starts_with and uses it in LibJS's abstract relation, which unlike String/StringView::starts_with this compares utf8 code points instead of "characters" (bytes), which is important when handling aribtary utf-8 input that could include overlong characters. This was found by/fixes https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=31134.

NOTE: while this fixes this specific bug i think there is more discussion to be had about the handing of malformed utf8 source code in libjs, as currently we do no validation to the input source code (a lot of invalid utf8 codes like 0xFE can crash libjs). overlong encoded utf8 sequences are a seperate issue as well, since they are technically valid, but usually are a sign of bad data (and as such rejecting them/and or replacing them with valid counterparts at/before the lexer stage might be a good idea). (relevant issue: #3549)

### DIFF
--- a/AK/Utf8View.cpp
+++ b/AK/Utf8View.cpp
@@ -144,6 +144,24 @@ size_t Utf8View::calculate_length() const
     return length;
 }
 
+bool Utf8View::starts_with(const Utf8View& start) const
+{
+    if (start.is_empty())
+        return true;
+    if (is_empty())
+        return false;
+    if (start.length() > length())
+        return false;
+    if (begin_ptr() == start.begin_ptr())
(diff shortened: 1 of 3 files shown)


IdanHo:
RE: the note above, i looked into v8, and it looks like their approach is to replace both illegal characters (like 0xFE) and semi-legal characters (like overlong encoded ones) with the unicode replacement character (U+FFFD)

PR #5903, opened 2021-03-21
Documentation: Fix missing subdirectory in build path

infi:
The build path moved from `Build/` into `Build/i686/`, but the `cmake` command in the documentation currently does not accommodate for that, as the `CMakeLists.txt` is in the root rather than in `Build/`, which causes `cmake` to be unable to find the file.

### DIFF
--- a/Documentation/BuildInstructions.md
+++ b/Documentation/BuildInstructions.md
@@ -187,7 +187,7 @@ Once the toolchain has been built, go into the `Build/i686/` directory and run t
 
 ```console
 $ cd ../Build/i686
-$ cmake ../.. -G Ninja
+$ cmake ../../.. -G Ninja
 $ ninja install
 ```
 


linusg:
This seems wrong. Let's break it down:

- We're starting in `$SERENITY_ROOT/Toolchain`
- `$ cd ../Build/i686`
- now we're in `$SERENITY_ROOT/Build/i686`
- Previously we'd do `$ cmake ../.. -G Ninja`, i.e. `$ cmake $SERENITY_ROOT -G Ninja`. That's where `CMakeLists.txt` is
- However, now we're doing `$ cmake ../../.. -G Ninja`, i.e. going past the repository root... :thinking:

linusg:
In fact this was updated for that very change you mention, here: https://github.com/SerenityOS/serenity/commit/d0427b610c52a88f2e0a13f76fbe4c7c26769999

```diff
-$ cd ../Build
-$ cmake .. -G Ninja
+$ cd ../Build/i686
+$ cmake ../.. -G Ninja
```

infi:
Must've been my issue then, although I have double-checked my logic twice. However, it does make sense now that you broke it down. Closing the PR since it's invalid.

PR #5915, opened 2021-03-22
LibJS: Simplify positive/negative zero checks

olegv11:
I thought GCC would magically optimize the division by infinity away, but apparently it would not. See https://godbolt.org/z/8z81PYfTW

### DIFF
--- a/AK/BitCast.h
+++ b/AK/BitCast.h
@@ -1,0 +1,43 @@
+/*
+ * Copyright (c) 2021, the SerenityOS developers.
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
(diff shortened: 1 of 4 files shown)


RealKC:
Any reason not to simply use [signbit](https://en.cppreference.com/w/c/numeric/math/signbit) which our LibM has?

IdanHo:
> Any reason not to simply use [signbit](https://en.cppreference.com/w/c/numeric/math/signbit) which our LibM has?

readability wise thats probably a good idea, but godbolt tells me that signbit (even after optimizations) emits 15 floating point instructions for this case, while olegv11's solution emits 5 "normal" (alu/memory) instructions, so for the value constructor (which is called quite a bit) that would likely be preferable

RealKC:
That's a fair point, maybe we could extract the FloatExtractor from LibM and see how that fares?

PR #5916, opened 2021-03-22
Kernel: Make VMWare to boot with AHCI

supercomputer7:
@Lubrsi can you please test this patch and tell if it works for you?


Lubrsi:
Yep:
Screenshot from 2021-03-23 01-46-15

(as a sidenote and probably for a separate PR, the name is still "Spurious Interrupt Handler")

supercomputer7:
> Yep:
> Screenshot from 2021-03-23 01-46-15
> 
> (as a sidenote and probably for a separate PR, the name is still "Spurious Interrupt Handler")

Well, tha'ts really a tiny fix, so let's do it now :)

PR #5920, opened 2021-03-23
HackStudio: Added ability to rename files in project tree.

sagefarrenholz:
First commit to Serenity, still learning the ins and outs of the system. This change makes working in HackStudio a little less painful.  Not sure about the key shortcut, wanted to do Shift + Return but that could probably be used within the editor in the future.

### DIFF
--- a/Userland/DevTools/HackStudio/HackStudioWidget.cpp
+++ b/Userland/DevTools/HackStudio/HackStudioWidget.cpp
@@ -302,6 +302,7 @@ NonnullRefPtr<GUI::Menu> HackStudioWidget::create_project_tree_view_context_menu
     m_open_selected_action = create_open_selected_action();
     m_new_file_action = create_new_file_action();
     m_new_directory_action = create_new_directory_action();
+    m_rename_action = create_rename_action();
     m_delete_action = create_delete_action();
     auto project_tree_view_context_menu = GUI::Menu::construct("Project Files");
     project_tree_view_context_menu->add_action(*m_open_selected_action);
@@ -309,6 +310,7 @@ NonnullRefPtr<GUI::Menu> HackStudioWidget::create_project_tree_view_context_menu
     project_tree_view_context_menu->add_separator();
     project_tree_view_context_menu->add_action(*m_new_file_action);
     project_tree_view_context_menu->add_action(*m_new_directory_action);
+    project_tree_view_context_menu->add_action(*m_rename_action);
     project_tree_view_context_menu->add_action(*m_delete_action);
     return project_tree_view_context_menu;
 }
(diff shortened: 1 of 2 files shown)


Userland/DevTools/HackStudio/HackStudioWidget.cpp
```diff
@@ -422,6 +424,25 @@ NonnullRefPtr<GUI::Action> HackStudioWidget::create_new_project_action()
...
+            return;
+
+        auto pre_filename = files[0];
+
+        String new_name;
+        if (GUI::InputBox::show(window(), new_name, "Enter the new file name:", "Rename file") != GUI::InputBox::ExecOK) {
+            return;
+        }
+        if (!new_name.is_empty()) {
+            rename(pre_filename.characters(), new_name.characters());
```
ADKaster, line 441: Unless I'm misreading this, you're calling `::rename` here from the C library, which will simply change the name of the file in the filesystem. 

If the file is currently open in an editor window and you rename it, what happens to that editor? I'm assuming that if you save the file, the old file will magically re-appear? I think this action probably wants to do something interesting to the project tree view (if the file exists in the view, anyway).

Userland/DevTools/HackStudio/HackStudioWidget.cpp
```diff
@@ -422,6 +424,25 @@ NonnullRefPtr<GUI::Action> HackStudioWidget::create_new_project_action()
...
+{
+    return GUI::Action::create("Rename", { Mod_Alt, Key_Return }, [this](const GUI::Action&) {
+        auto files = selected_file_names();
+        if (files.is_empty() || files.size() > 1)
+            return;
+
+        auto pre_filename = files[0];
+
+        String new_name;
+        if (GUI::InputBox::show(window(), new_name, "Enter the new file name:", "Rename file") != GUI::InputBox::ExecOK) {
```
ADKaster, line 437: It looks like the way this is written, you have to supply the entire path in the input box? I think it would probably be more natural to only rename the file relative to its parent.

i.e. If I have `/home/anon/MyProject/foo.cpp`, I would expect that typing `bar.cpp` into the rename dialog would rename it to `/home/anon/MyProject/bar.cpp`, rather than `/path/where/I/opened/hackstudio/from/bar.cpp`. 

Or, maybe the dialog could fill in the input box with `/home/anon/MyProject/foo.cpp` already and all you have to do is edit the string in place.

`LexicalPath::relative_path` would probably help for the first suggestion.

ADKaster (commented):
Hi there! Welcome to the project!

Make sure you squash your fixup commits per the [contribution guidelines](https://github.com/SerenityOS/serenity/blob/master/CONTRIBUTING.md#code-submission-policy).

Looks like this will definitely rename a file in the filesystem from hack studio. Few comments though..

Userland/DevTools/HackStudio/HackStudioWidget.cpp
sagefarrenholz: I went ahead and made it update the open files list and reopen the renamed file (if it is open in the editor). You were right though, before this change saving the previously named file after the rename would create a duplicate.

Userland/DevTools/HackStudio/HackStudioWidget.cpp
sagefarrenholz: Yup, trying to rename any file within a directory in the tree would fail as rename expects the full relative path. Went ahead and fixed that with some string hacking.

Userland/DevTools/HackStudio/HackStudioWidget.cpp
```diff
@@ -422,6 +424,57 @@ NonnullRefPtr<GUI::Action> HackStudioWidget::create_new_project_action()
...
+                m_open_files_vector.remove(file_list_index.value());
+                if (currently_open_file().equals_ignoring_case(old_relative_path)) {
+                    open_file(new_relative_path);
+                } else {
+                    m_open_files_vector.append(new_relative_path);
+                }
+                auto project_file = m_open_files.get(old_relative_path).value();
+                m_open_files.remove(old_relative_path);
+                m_open_files.set(new_full_path, project_file);
+                m_open_files_view->model()->update();
```
ADKaster, line 472: Is this update sufficient? The existing ProjectFile has an m_name property associated with the old filename. And it might have a CodeDocument member associated with the old file as well? There's probably a way to safely re-use the properties of the old file (such as scroll position, highlights, etc), but I think the "safest" and quickest correct update would be to throw away the old file and ProjectfFile::create() one with new path.

For example: What happens if you do a rename on a file that you once had open, that is not the currently open file, then try to FindInFiles? Will it use the proper new filename? I suspect the result will properly search the new file, but the search result view will show the old filename.

PR #5924, opened 2021-03-23
LibCpp + Language Server: Support Namespaces

itamar8910:
We can now parse c++ namespaces, and include the scope of symbols in the Locator.

Example:
hs-namespace


awesomekling:
Does this support `namespace Foo::Bar {`?

itamar8910:
> Does this support `namespace Foo::Bar {`?

It does now :wink: 
So now we can technically say we support parsing _some_ C++17 features :laughing:

PR #5927, opened 2021-03-23
LibJS: Don't static_cast<double>() various Date getter values

linusg:
Since we have `Value::Type::Int32` now, let's use the `Value(i32)` constructor here directly by not casting these i32 values to doubles. The `Value(double)` would also figure out that these can be stored as integers, but needs to do extra work which is not needed here. :^)

### DIFF
--- a/Userland/Libraries/LibJS/Runtime/DatePrototype.cpp
+++ b/Userland/Libraries/LibJS/Runtime/DatePrototype.cpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2020, Linus Groh <mail@linusgroh.de>
+ * Copyright (c) 2020-2021, Linus Groh <mail@linusgroh.de>
  * Copyright (c) 2021, Petróczi Zoltán <petroczizoltan@tutanota.com>
  * All rights reserved.
  *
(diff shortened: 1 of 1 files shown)


Userland/Libraries/LibJS/Runtime/DatePrototype.cpp
```diff
@@ -178,7 +178,7 @@ JS_DEFINE_NATIVE_FUNCTION(DatePrototype::set_full_year)
     datetime.set_time(new_year, new_month, new_day, datetime.hour(), datetime.minute(), datetime.second());
     this_object->set_is_invalid(false);
 
-    return Value { this_object->time() };
+    return Value(this_object->time());
```
awesomekling, line 181: :eyes:

linusg: for the sake of consistencyyyyy

(only use of `Value { foo }` in all of LibJS, can't have that)

awesomekling: Yeah yeah all right then Linus 😄

PR #5932, opened 2021-03-23
LibGUI+Playground: GML improvements

edgararaj:


alimpfard:
I _think_ this'll break autocomplete suggestions?
Pretty sure it won't like having two namea next to each other

edgararaj:
Good catch! This will fix it

PR #5933, opened 2021-03-24
jp: add variable indentation level and use stdin input by default

ngc6302h:
This PR was originally for a new utility, but upon realizing it was sort of duplicated, turned into extending an existing one.

Pass -i N where N is an integer to change the indentation to N spaces.

### DIFF
--- a/Base/usr/share/man/man1/jp.md
+++ b/Base/usr/share/man/man1/jp.md
@@ -1,0 +1,35 @@
+## Name
+
+jp - pretty-print a JSON file with syntax-coloring and indentation
+
+## Synopsis
+
+```**sh
+$ jp [options...] [path...]
+```
+
+## Description
+
+`jp` pretty-prints a JSON file with syntax-coloring and indentation.
+
+If no *path* argument is provided stdin is used.
+
+## Options
+
+* `--help`: Display this message
+* `-i`, `--indent-size`: Size of indentations in spaces
+
+## Arguments
+
+* path: file to pretty-print
+
+## Examples
+
+```sh
+# Pretty-print stdin
+$ cat /proc/all | jp
+# Pretty-print a file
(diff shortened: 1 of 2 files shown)


bcoles:
This looks similar to the existing [jp](https://github.com/SerenityOS/serenity/blob/master/Userland/Utilities/jp.cpp) tool. Although `jp` does not accept stdin by default, instead requiring `cat /proc/all | jp /dev/stdin` and does not support customizable indentation.

ngc6302h:
> This looks similar to the existing [jp](https://github.com/SerenityOS/serenity/blob/master/Userland/Utilities/jp.cpp) tool. Although `jp` does not accept stdin by default, instead requiring `cat /proc/all | jp /dev/stdin` and does not support customizable indentation.

Whoops, I didn't realize. What do you recommend I do now?

bcoles:
> 
> 
> > This looks similar to the existing [jp](https://github.com/SerenityOS/serenity/blob/master/Userland/Utilities/jp.cpp) tool. Although `jp` does not accept stdin by default, instead requiring `cat /proc/all | jp /dev/stdin` and does not support customizable indentation.
> 
> Whoops, I didn't realize. What do you recommend I do now?

Perhaps augment `jp` to support `stdin` and add an indentation command line option?

ngc6302h:
> > > This looks similar to the existing [jp](https://github.com/SerenityOS/serenity/blob/master/Userland/Utilities/jp.cpp) tool. Although `jp` does not accept stdin by default, instead requiring `cat /proc/all | jp /dev/stdin` and does not support customizable indentation.
> > 
> > 
> > Whoops, I didn't realize. What do you recommend I do now?
> 
> Perhaps augment `jp` to support `stdin` and add an indentation command line option?

Very well :)

PR #5944, opened 2021-03-25
Kernel: Add barebones ACPI AML parsing support

tomuta:
_This is *very* early and incomplete, so don't waste your time on reviewing the code at this point, it will likely change significantly!_

- [ ] Get the included test.aml to run in the unit test


Kernel/ACPI/Parser.h
```diff
@@ -90,6 +97,8 @@ class Parser {
     Vector<PhysicalAddress> m_sdt_pointers;
     PhysicalAddress m_fadt;
     PhysicalAddress m_facs;
+    u64 m_dsdt { 0 };
```
supercomputer7, line 100: Is `m_dsdt` supposed to indicate the physical address of the the DSDT table? if so, why not to use `PhysicalAddress` instead?
Also, remember that the FADT holds this value twice (one for ACPI 1, and an extended pointer for later revisions).

supercomputer7:
I also wanted to mention the decision of making a `LibAML`, which is imho a very good one. Maybe we can write a userspace parser, so we could compile ASL files into AML blobs, but that's a feature that we will think about for a long time to come  :)

Kernel/ACPI/Parser.h
tomuta: Yeah eventually. The problem is that you can have 64 bit physical addresses on a 32 bit machine, but we use `FlatPtr` only. In fact, there's two DSDT pointers, an optional 64 bit physical address and the legacy 32 bit physical address.

tomuta:
Writing a "decompiler" with this should be trivial, it's essentially just dumping each node as a string...

Kernel/ACPI/Parser.h
supercomputer7: Then I think we should make `PhysicalAddress` to wrap a `u64` value as it represents a physical address which can be as you said 64 bits on 32 bit machine.
About the DSDT pointer, AFAIK most machines use both DSDT pointers to point to the same physical address as it doesn't make sense to make two copies of the same data. On my machine I decompiled the FADT blob and saw that this is true.

PR #5946, opened 2021-03-25
Remove unused private fields.

mhermier:
Hi,
Trying to compile Serenity on my system using clang triggered a bunch of warnings promoted to errors because of -Wunused-private-fields. The following change set remove these entries.

Please note that I can only compile test using clang for now since my distro is still stuck on gcc 9.2.1 for now.
Please review.

### DIFF
--- a/Kernel/FileSystem/File.cpp
+++ b/Kernel/FileSystem/File.cpp
@@ -31,7 +31,6 @@
 namespace Kernel {
 
 File::File()
-    : m_block_condition(*this)
 {
 }
 
--- a/Kernel/FileSystem/File.h
+++ b/Kernel/FileSystem/File.h
@@ -43,10 +43,7 @@ class File;
 
 class FileBlockCondition : public Thread::BlockCondition {
 public:
-    FileBlockCondition(File& file)
-        : m_file(file)
-    {
-    }
+    FileBlockCondition() { }
 
     virtual bool should_add_blocker(Thread::Blocker& b, void* data) override
     {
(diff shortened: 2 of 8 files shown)


ADKaster:
What do you mean by compiling with clang? Serenity must be compiled with a the cross compiler toolchain which you must build from source in the Toolchain directory. Unless you mean the Lagom/host build parts, but the Kernel is definitely not supposed to be built with a host compiler.

mhermier:
Unless I miss-understood the Documentation/BuildInstructions.md / CMakeLists.txt at no point the environment is twicked to use the Toolchain. So when I did (as peer Documentation):
```
$ cd Toolchain
$ ./BuildIt.sh
$ cd ../Build/i686
$ cmake ../.. -G Ninja
```
It started using my host compiler.

So I thought a first stage was required. So I give it a try adding options to cmake to use clang since it is my only distro available compiler that is enough near of C++20, since it was using my host compiler anyway... I don't complain here that it is not compiling with clang. But compiling with it showed some warning like the one I propose to resolve in the patch set.

ADKaster:
Yep these fixes seen fine, just curious what your setup was that it started trying to build with your first compiler for the serenity part. 

The CMakeLists changes a lot of variables halfway through to trick CMake into using the cross complier, we have some issues tracking maybe a better way to set all that up. If your environment has found a way to avoid the games we're playing with setting the CMake variables for all the compiler settings it might be time to revisit that topic.

mhermier:
Gave a try at various things, and it seems the variables are set correctly but not stored even by using `mark_as_advanced`.

mhermier:
For the record since it is highly seems to matter, I have cmake 3.19.7 (on Frugalware but I doubt that the distro matters here).

PR #5948, opened 2021-03-25
LibGUI: Show human readable size in filemanager

### LINKED ISSUE #5938: FileManager: Display file size in a human readable format
It would be nice to display the file size in a human readable format.

I took a quick look at it and made `FileSystemModel::data` return a formatted `String` instead of an `int` which would solve the issue. But the range selection size display use the same function to then calculate the total size of the files in the selection, so such a change would break the UI. There is maybe a better way to display the size in the `FileSystemModel`?

Current implementation:

```c++
Variant FileSystemModel::data(const ModelIndex& index, ModelRole role) const
...
    if (role == ModelRole::Display) {
...
        case Column::Size:
            return (int)node.size;
```

image

Proposal (with broken range select):

```c++
Variant FileSystemModel::data(const ModelIndex& index, ModelRole role) const
...
    if (role == ModelRole::Display) {
...
        case Column::Size:
            return human_readable_size(node.size);
```

image

guerinoni:
Th size is visible only in TableView mode.
This fixes #5938.

### DIFF
--- a/Userland/Libraries/LibGUI/FileSystemModel.cpp
+++ b/Userland/Libraries/LibGUI/FileSystemModel.cpp
@@ -25,6 +25,7 @@
  */
 
 #include <AK/LexicalPath.h>
+#include <AK/NumberFormat.h>
 #include <AK/QuickSort.h>
 #include <AK/StringBuilder.h>
 #include <LibCore/DirIterator.h>
@@ -439,7 +440,7 @@ Variant FileSystemModel::data(const ModelIndex& index, ModelRole role) const
         case Column::Name:
             return node.name;
         case Column::Size:
-            return (int)node.size;
+            return human_readable_size(node.size);
         case Column::Owner:
             return name_for_uid(node.uid);
         case Column::Group:
--- a/Userland/Applications/FileManager/DirectoryView.cpp
+++ b/Userland/Applications/FileManager/DirectoryView.cpp
@@ -413,10 +413,8 @@ void DirectoryView::update_statusbar()
     size_t selected_byte_count = 0;
 
     current_view().selection().for_each_index([&](auto& index) {
-        auto& model = *current_view().model();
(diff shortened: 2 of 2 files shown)


linusg:
As mentioned in #5938, this breaks the size of selected files information in the status bar.

linusg:
Here's how to fix it:

```diff
--- a/Userland/Applications/FileManager/DirectoryView.cpp
+++ b/Userland/Applications/FileManager/DirectoryView.cpp
@@ -413,10 +413,8 @@ void DirectoryView::update_statusbar()
     size_t selected_byte_count = 0;
 
     current_view().selection().for_each_index([&](auto& index) {
-        auto& model = *current_view().model();
-        auto size_index = model.index(index.row(), GUI::FileSystemModel::Column::Size, model.parent_index(index));
-        auto file_size = size_index.data().to_i32();
-        selected_byte_count += file_size;
+        auto& node = this->node(index);
+        selected_byte_count += node.size;
     });
 
     StringBuilder builder;
```

Userland/Libraries/LibGUI/FileSystemModel.cpp
```diff
@@ -439,7 +440,7 @@ Variant FileSystemModel::data(const ModelIndex& index, ModelRole role) const
         case Column::Name:
             return node.name;
         case Column::Size:
-            return (int)node.size;
+            return AK::human_readable_size(node.size);
```
linusg, line 443: please drop the AK namespace prefix here, this has a `using` directive:

```suggestion
            return human_readable_size(node.size);
```

guerinoni: Thanks for the suggestion! I fixed it!

guerinoni:
@linusg many thanks for the accurate review and suggestions. If this is ok I will autosquash in 1 commit

linusg:
Yes, please squash both fixup commits into the initial one :)

linusg (approved):

PR #5954, opened 2021-03-26
Kernel: Implements fastbins for heap

Jorropo:
This makes kmalloc and realateds to use a slab allocator instead of the variable heap if praticable.

This is somewhat faster but not by much (`<1%` on overall system I think).

Idealy the compiler would "rewrite" all calls to `kmalloc` where size is known at build time and would be better if used in a slab allocator into a slab allocator call else for when the size is unknown the current implementation is fine. But I really don't know how to do this kind of compiler instrumentation.

### DIFF
--- a/Kernel/Heap/SlabAllocator.cpp
+++ b/Kernel/Heap/SlabAllocator.cpp
@@ -36,99 +36,76 @@
 namespace Kernel {
 
 template<size_t templated_slab_size>
-class SlabAllocator {
-public:
-    SlabAllocator() = default;
-
-    void init(size_t size)
-    {
-        m_base = kmalloc_eternal(size);
-        m_end = (u8*)m_base + size;
-        FreeSlab* slabs = (FreeSlab*)m_base;
-        m_slab_count = size / templated_slab_size;
-        for (size_t i = 1; i < m_slab_count; ++i) {
-            slabs[i].next = &slabs[i - 1];
-        }
-        slabs[0].next = nullptr;
-        m_freelist = &slabs[m_slab_count - 1];
-        m_num_allocated = 0;
+void SlabAllocator<templated_slab_size>::init(void* base, size_t size)
+{
+    m_base = base;
+    m_end = (u8*)m_base + size;
+    FreeSlab* slabs = (FreeSlab*)m_base;
+    m_slab_count = size / templated_slab_size;
+    for (size_t i = 1; i < m_slab_count; ++i) {
+        slabs[i].next = &slabs[i - 1];
     }
+    slabs[0].next = nullptr;
+    m_freelist = &slabs[m_slab_count - 1];
+    m_num_allocated = 0;
+}
 
-    constexpr size_t slab_size() const { return templated_slab_size; }
-    size_t slab_count() const { return m_slab_count; }
-
-    void* alloc()
+template<size_t templated_slab_size>
+void* SlabAllocator<templated_slab_size>::alloc()
+{
+    FreeSlab* free_slab;
     {
-        FreeSlab* free_slab;
-        {
-            // We want to avoid being swapped out in the middle of this
-            ScopedCritical critical;
-            FreeSlab* next_free;
-            free_slab = m_freelist.load(AK::memory_order_consume);
-            do {
-                if (!free_slab)
-                    return kmalloc(slab_size());
-                // It's possible another processor is doing the same thing at
-                // the same time, so next_free *can* be a bogus pointer. However,
-                // in that case compare_exchange_strong would fail and we would
-                // try again.
-                next_free = free_slab->next;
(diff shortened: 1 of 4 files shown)


alimpfard:
Try inlining the `kmalloc` function, if the size is known at compiletime, the compiler will most likely emit only the taken branches that depend on the size.
(please verify that it does so, if you try this)

Jorropo:
> 
> 
> Try inlining the `kmalloc` function, if the size is known at compiletime, the compiler will most likely emit only the taken branches that depend on the size.
> (please verify that it does so, if you try this)

Yes I've tryed, but I can't get it working, this somehow creates an infinite template loop in `AK`. (It's not releated to the inline directly, but including `SlabAllocator.h` in `kmalloc.h` somehow mess everything up).
Plus it's not an ideal solution, the slab sellection code is a bit big and `kmalloc` is called a lot, I think this would blot up the binary more than we want. (I'm searching for a solution to inline `kmalloc` only if the size is known at compile time).

alimpfard:
Marking only `kmalloc` as always_inline will do exactly what you want (it won't necessarily inline `kmalloc_slowbin` or the various slab allocator functions).

You don't need to include the allocator header, just declare them in `kmalloc.h` and put `kmalloc` in the header

Jorropo:
> Marking only `kmalloc` as always_inline will do exactly what you want (it won't necessarily inline `kmalloc_slowbin` or the various slab allocator functions).

Wait I want to be sure :
- if I mark kmalloc as always_inline it will inline only if the size is compile time known ? (I don't want to inline if the size is unkown)
- Where should I put this inline ? On the declaration in `.h` or in the implementation in `.cpp` ?

alimpfard:
I don't think conditionally controlling how it's inlined is possible, if you mark it always_inline, `kmalloc` will _always_ be inlined, but as it's a pretty small function, I don't think this is gonna be a big problem.

Though I don't think inlining it is gonna have any particularly interesting result, the main user of `kmalloc` is `operator new`, and that guy is out of line.

> Where should I put this inline?

Move the implementation of `kmalloc` from the cpp file to the header, and slap `ALWAYS_INLINE` before it.

The only way I can think of to get the effect you're looking for is to move `kmalloc` to the header as `inline void* kmalloc(size_t)`, and just hope that the compiler would decide to inline it correctly.

Jorropo:
> Move the implementation of `kmalloc` from the cpp file to the header, and slap `ALWAYS_INLINE` before it.

I've tryed, I really can't get it working, this spawns issues with `AK` somehow.

> Though I don't think inlining it is gonna have any particularly interesting results, the main user of kmalloc is operator new, and that guy is out of line.

WDYM ? Are you saying that the new operator is never gonna be inlined or that it can't inline anything ?
Because for the compiler to understand what I want it would only need to inline `kmalloc` into new, not new.

alimpfard:
`operator new` won't be inlined because it's in another translation unit (`kmalloc.cpp`), and so everything will be a runtime value as far as `kmalloc` is concerned

awesomekling:
How did you performance test this, and what was the improvement?

All the inlining looks kinda gratuitous, how much does it actually help?

Jorropo:
> How did you performance test this, and what was the improvement?

I used the Userland Emulator to "multiply" syscalls and run various programs (`jq`, `id`, ...)

Running things like `time /bin/UserlandEmulator /bin/UserlandEmulator /bin/id` ... and comparing results before and after the patch (it was constantly faster without but not by much (`~0.2%` I think, but it's hard to say that because this change a lot depending of the program being runned, never more than `0.7%`)).

I think a better kernel performance tester would be better for it but since it's constently faster but not by much, fast bins aren't a new thing and this is mostly only new compares I just didn't felt the need to more properly check that.
Actualy I noticed that it was rare for slabs to overflow and fallback into kmalloc, this isn't the case anymore (wouldn't variable size slabs backed by kmalloc memory needed ?)

> All the inlining looks kinda gratuitous, how much does it actually help?

Which inlining ? I couldn't get it working (and I don't think it's actually what would be needed).

I'm thinking about a kmalloc rewrite with variable pool size (reserving memory using MM) anyway so you might want to wait before merging this.

tomuta:
Our kmalloc already supports expanding the heap using MM. Or what do you mean?

Jorropo:
> Our kmalloc already supports expanding the heap using MM. Or what do you mean?

What is :
https://github.com/SerenityOS/serenity/blob/0f468a5013f125c467e491cf58e74f769efc96e1/Kernel/Heap/kmalloc.cpp#L48

Then ? (the blocksize to pass to MM ?)

tomuta:
That's the initial pool size that the system is limited to until MM is initialized

tomuta:
Once MM is initialized we allocate new regions and expand the heap as needed.

PR #5958, opened 2021-03-26
LibWeb+WebContent: Support displaying favicons in OOPWV

trflynn89:
In single process mode, the browser will display a page's favicon in
both the location bar and tab. This adds the same support for multi-
process mode.


awesomekling:
Excellent patch!

PR #5959, opened 2021-03-26
gunzip: Fail gracefully on bad input

IdanHo:
Several small fixes to the gunzip utility:
- Fail gracefully instead of asserting on a missing input file
- Fail gracefully instead of looping forever on a bad (not gzip compressed) input file
- Fail gracefully instead of asserting on a permission problem in the output file

### DIFF
--- a/Userland/Utilities/gunzip.cpp
+++ b/Userland/Utilities/gunzip.cpp
@@ -29,16 +29,18 @@
 #include <LibCore/FileStream.h>
 #include <unistd.h>
 
-static void decompress_file(Buffered<Core::InputFileStream>& input_stream, Buffered<Core::OutputFileStream>& output_stream)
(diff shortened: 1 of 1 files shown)


Userland/Utilities/gunzip.cpp
```diff
@@ -65,17 +67,35 @@ int main(int argc, char** argv)
 
         auto input_stream_result = Core::InputFileStream::open_buffered(input_filename);
 
+        if (input_stream_result.is_error()) {
+            dbgln("Failed opening input file for reading: {}", input_stream_result.error());
```
awesomekling, line 71: For command-line utilities, we should report errors to `stderr`, so `warnln()` rather than `dbgln()` :)

IdanHo: i wanted to do that, but i noticed that one of the optional flags enables piping to stdout

IdanHo: oh, nvm, warnln writes to stderr, so it should still be just fine, will fix :)

PR #5960, opened 2021-03-26
LibCoreDump+CrashDaemon: Compress coredumps

### LINKED ISSUE #4601: Kernel: Coredumps waste lots of space and should probably be compressed

IdanHo:
Most coredumps contain large amounts of consecutive null bytes and as such are a prime candidate for compression.
This commit makes CrashDaemon compress files once the kernel finishes emitting them, as well as adds the functionality needed in LibCoreDump to then parse them. (this fixes #4601)

I'm not entirely sure if this is the right place to compress the core dumps (aka, if it should be done in the kernel, or someplace else), but it works for now. Also, another thing to consider is the (small) increase in startup time for CrashReporter due to the decompression that is now done. (especially the initial launch right after the crash, as that now includes the time needed to compress as well)

### DIFF
--- a/Kernel/CoreDump.cpp
+++ b/Kernel/CoreDump.cpp
@@ -72,7 +72,7 @@ RefPtr<FileDescription> CoreDump::create_target_file(const Process& process, con
         return nullptr;
     }
     auto dump_directory_metadata = dump_directory.value()->inode().metadata();
-    if (dump_directory_metadata.uid != 0 || dump_directory_metadata.gid != 0 || dump_directory_metadata.mode != 040755) {
+    if (dump_directory_metadata.uid != 0 || dump_directory_metadata.gid != 0 || dump_directory_metadata.mode != 040777) {
         dbgln("Refusing to put core dump in sketchy directory '{}'", output_directory);
         return nullptr;
     }
@@ -329,7 +329,7 @@ KResult CoreDump::write()
     if (result.is_error())
         return result;
 
-    return m_fd->chmod(0400); // Make coredump file readable
+    return m_fd->chmod(0600); // Make coredump file read/writable
 }
 
 }
--- a/Userland/Applications/CrashReporter/main.cpp
+++ b/Userland/Applications/CrashReporter/main.cpp
@@ -30,6 +30,7 @@
 #include <AK/URL.h>
 #include <Applications/CrashReporter/CrashReporterWindowGML.h>
 #include <LibCore/ArgsParser.h>
+#include <LibCore/File.h>
 #include <LibCoreDump/Backtrace.h>
 #include <LibCoreDump/Reader.h>
 #include <LibDesktop/AppFile.h>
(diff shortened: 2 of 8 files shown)


itamar8910:
Definitely nice to have the coredump files not take so much memory after we use them.

Also I agree with the approach taken here that compression should be done in userspace and not kernel. Doing less in the kernel is always good.

However compressing the coredump file and then passing it to the crash reporter which decompresses it is not really ideal.
An improvement (that could be done in the future, not a must for now IMO) could be deleting the coredump file after the crash reporter parses it, and offering an option in the crash reporter to save a compressed coredump file to persistent storage if the user wants to keep it.

IdanHo:
> However compressing the coredump file and then passing it to the crash reporter which decompresses it is not really ideal.
> An improvement (that could be done in the future, not a must for now IMO) could be deleting the coredump file after the crash reporter parses it, and offering an option in the crash reporter to save a compressed coredump file to persistent storage if the user wants to keep it.

Compressing and then immediately decompressing the core dump does feel a bit clunky, another option is to have CrashDaemon/CrashReporter compress the core dump *after* the first launch of crash reporter following the crash (since LibCoreDump can parse both compressed and uncompressed core dumps with this patch), that way we dont have any overhead in the initial post-crash launch. (not sure which of the options we should go with)

itamar8910:
> compress the core dump _after_ the first launch of crash reporter following the crash

Yeah that's also a good option. Just need to take care not to perform the compression in-place on the coredump file because that could be racy with CrashReporter reading it.
Also it would be nice to add a `.gz` extension to the compressed coredump file

IdanHo:
> Yeah that's also a good option. Just need to take care not to perform the compression in-place on the coredump file because that could be racy with CrashReporter reading it.
> Also it would be nice to add a `.gz` extension to the compressed coredump file

Writing the file to a .gz extension will essentially fix that race, but then theres the problem of the deleting the old file (since then its a race of who finishes first, CrashReporter's mmap and CrashDaemon's unlink)

IdanHo:
Hmm, looks like /tmp/coredump/ is owned by root as well (and is read-only for others), so we have to change the permission of the dir if we want to rename the files to .gz

itamar8910:
We could do:
1. CrashDaemon reads the file to memory
2. Invokes CrashReporter
3. CrashDaemon creates the .gz file
4. CrashReporter reads the original coredump file and then unlinks it (we could add a --unlink to CrashReporter for this)

> Hmm, looks like /tmp/coredump/ is owned by root as well (and is read-only for others), so we have to change the permission of the dir if we want to rename the files to .gz

Hmm I think it's okay to have it writable by everyone for now, and maybe change it back to readonly when we add the "export coredump" feature to CrashReporter.

IdanHo:
> We could do:
> 
> 1. CrashDaemon reads the file to memory
> 2. Invokes CrashReporter
> 3. CrashDaemon creates the .gz file
> 4. CrashReporter reads the original coredump file and then unlinks it (we could add a --unlink to CrashReporter for this)

This looks simple enough so ill implement it in this PR and leave the export coredump stuff to a possible future one.

IdanHo:
Ok, implemented that, but as a side effect of making /tmp/coredump writeable anyone can trigger CrashDaemon by creating a file in the directory, not sure if we want/can do anything about that

itamar8910:
Okay, looks good to me :smile:

Issue #5893, opened 2021-03-20
Unable to boot on bare metal (Intel ICH7 chipset, 2009 era machine)


supercomputer7:
The log from serial port:
```
[#0 init_stage2(2:2)]: UHCI: Allocated framelist at physical address P0x00e06000
[#0 init_stage2(2:2)]: UHCI: Framelist is at virtual address V0xc115d000
[#0 init_stage2(2:2)]: UHCI: QH(0xc115f000) @ 14708736: link_ptr=14708770, element_link_ptr=1
[#0 init_stage2(2:2)]: UHCI: QH(0xc115f020) @ 14708768: link_ptr=14708802, element_link_ptr=1
[#0 init_stage2(2:2)]: UHCI: QH(0xc115f040) @ 14708800: link_ptr=14708834, element_link_ptr=1
[#0 init_stage2(2:2)]: UHCI: QH(0xc115f060) @ 14708832: link_ptr=14708866, element_link_ptr=1
[#0 init_stage2(2:2)]: UHCI: QH(0xc115f080) @ 14708864: link_ptr=14721025, element_link_ptr=1
[#0 init_stage2(2:2)]: UHCI: Reset completed
[#0 init_stage2(2:2)]: UHCI: Started
[#0 init_stage2(2:2)]: DMIExpose: SMBIOS 32bit Entry point @ P0x000f0ce0
[#0 init_stage2(2:2)]: DMIExpose: Data table @ P0x000f0100
[#0 init_stage2(2:2)]: IDEChannel: Primary master ATA device found: Name=WDC WD20EZAZ-00GGJB0, Capacity=137438952960, C/H/Spt=16383/16/63, Capabilities=0x2f00
[#0 init_stage2(2:2)]: Trying to unregister unused handler (?)
[#0 init_stage2(2:2)]: KUBSAN: multiplication overflow, 'int' (32-bit)
[#0 init_stage2(2:2)]: KUBSAN: at ../Kernel/Storage/IDEChannel.cpp, line 392, column: 50
[#0 init_stage2(2:2)]: 0xc04caa4b (next: 0xc11398d0)
[#0 init_stage2(2:2)]: 0xc05ef3e3 (next: 0xc1139a20)
[#0 init_stage2(2:2)]: 0xc05f1353 (next: 0xc1139a80)
[#0 init_stage2(2:2)]: 0xc05f14e9 (next: 0xc1139ae0)
[...]

PR #5971, opened 2021-03-27
Kernel: Bare metal fixes

supercomputer7:
The purpose is to fix the boot problem indicated in #5893.
When booting with the kernel argument `force_pio`, the system boots just fine, the problem is with the DMA implementation.


supercomputer7:
One last test before I can confirm this is working on the test machine :)

supercomputer7:
It works and boots up :)

supercomputer7:
And this is a image I took, the lighting condition is not go great, but we have something :)

Issue #5975, opened 2021-03-27
LibArchive: Buffer overflow in EndOfCentralDirectory::read


Lubrsi:
Found with FuzzZip.

[crash-a8b0cfbaa006f4fa95bdcf308ba014e7e95a59de.zip](https://github.com/SerenityOS/serenity/files/6215763/crash-a8b0cfbaa006f4fa95bdcf308ba014e7e95a59de.zip)

Trace:
```
=================================================================
==57861==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x60300002c6c4 at pc 0x00000052a6aa bp 0x7ffd4de7fb50 sp 0x7ffd4de7f318
READ of size 18 at 0x60300002c6c4 thread T0
    #0 0x52a6a9 in __asan_memcpy (/home/lukew/Desktop/serenity-project/serenity/build/Meta/Lagom/Fuzzers/FuzzZip+0x52a6a9)
    #1 0x5645e5 in Archive::EndOfCentralDirectory::read(AK::Span<unsigned char const>) /home/lukew/Desktop/serenity-project/serenity/build/../Userland/Libraries/LibArchive/Zip.h:59:9
    #2 0x5627a2 in Archive::Zip::try_create(AK::Span<unsigned char const> const&) /home/lukew/Desktop/serenity-project/serenity/build/../Userland/Libraries/LibArchive/Zip.cpp:54:35
    #3 0x55d410 in LLVMFuzzerTestOneInput /home/lukew/Desktop/serenity-project/serenity/build/../Meta/Lagom/Fuzzers/FuzzZip.cpp:34:21
    #4 0x4656f1 in fuzzer::Fuzzer::ExecuteCallback(unsigned char const*, unsigned long) (/home/lukew/Desktop/serenity-project/serenity/build/Meta/Lagom/Fuzzers/FuzzZip+0x4656f1)
    #5 0x464e35 in fuzzer::Fuzzer::RunOne(unsigned char const*, unsigned long, bool, fuzzer::InputInfo*, bool*) (/home/lukew/Desktop/serenity-project/serenity/build/Meta/Lagom/Fuzzers/FuzzZip+0x464e35)
    #6 0x4670d7 in fuzzer::Fuzzer::MutateAndTestOne() (/home/lukew/Desktop/serenity-project/serenity/build/Meta/Lagom/Fuzzers/FuzzZip+0x4670d7)
[...]

PR #5976, opened 2021-03-27
Lagom/Fuzzers: Add fuzzer for zip file parser

Lubrsi:
First issue: #5975


Meta/Lagom/Fuzzers/FuzzZip.cpp
```diff
@@ -0,0 +1,43 @@
...
+#include <LibArchive/Zip.h>
+#include <stddef.h>
+#include <stdint.h>
+
+extern "C" int LLVMFuzzerTestOneInput(const uint8_t* data, size_t size)
+{
+    ByteBuffer zip_data = ByteBuffer::copy(data, size);
+    auto zip_file = Archive::Zip::try_create(zip_data);
+    if (!zip_file.has_value())
+        return 1;
```
linusg, line 36: From https://releases.llvm.org/3.8.1/docs/LibFuzzer.html:

> Build a test driver using the same options as the library. The test driver is a C/C++ file containing interesting calls to the library inside a single function `extern "C" int LLVMFuzzerTestOneInput(const uint8_t *Data, size_t Size);`. **Currently, the only expected return value is 0, others are reserved for future.**

PR #5982, opened 2021-03-27
Utilities: Make sure not to overwrite existing file (shot)

r0ei:
A file might get overwritten if you make another screenshot and provide a name of a file that already exists. If a file already exists the patch will just add a number after the name, for example, `file(1).png`, `file(2).png` and so on. I am also checking if `output_path` contains `.png` at the end, and if not I automatically add it.

### DIFF
--- a/Userland/Utilities/shot.cpp
+++ b/Userland/Utilities/shot.cpp
@@ -49,6 +49,9 @@ int main(int argc, char** argv)
     if (output_path.is_empty())
         output_path = Core::DateTime::now().to_string("screenshot-%Y-%m-%d-%H-%M-%S.png");
     
+    if (!output_path.ends_with(".png"))
+        output_path = String::formatted("{}.png", output_path);
+    
     while (Core::File::exists(output_path)) {
         file_append += 1;
         if (!Core::File::exists(String::formatted("{}({}).png", output_path.split('.')[0], file_append)))
--- a/Userland/Utilities/shot.cpp
+++ b/Userland/Utilities/shot.cpp
@@ -38,6 +38,7 @@ int main(int argc, char** argv)
     Core::ArgsParser args_parser;
 
     String output_path;
+    int file_append = 0;
     bool output_to_clipboard = false;
 
     args_parser.add_positional_argument(output_path, "Output filename", "output", Core::ArgsParser::Required::No);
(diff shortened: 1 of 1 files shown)


Userland/Utilities/shot.cpp
```diff
@@ -38,17 +38,26 @@ int main(int argc, char** argv)
...
 
-    if (output_path.is_empty()) {
+    if (output_path.is_empty())
         output_path = Core::DateTime::now().to_string("screenshot-%Y-%m-%d-%H-%M-%S.png");
+    
+    while (Core::File::exists(output_path)) {
+        file_append += 1;
+        if (!Core::File::exists(String::formatted("{}({}).png", output_path.split('.')[0], file_append)))
+            break;
     }
```
awesomekling, line 59: This is okay for auto-generated file names but if I'm specifying a name, I don't want the program to mess with it.

Userland/Utilities/shot.cpp
```diff
@@ -49,6 +49,9 @@ int main(int argc, char** argv)
     if (output_path.is_empty())
         output_path = Core::DateTime::now().to_string("screenshot-%Y-%m-%d-%H-%M-%S.png");
     
+    if (!output_path.ends_with(".png"))
+        output_path = String::formatted("{}.png", output_path);
```
awesomekling, line 53: If the user didn't append `.png` we should assume the user didn't want `.png` :)

r0ei: @awesomekling  Hmmm thought it only supports PNG format..

Userland/Utilities/shot.cpp
r0ei: @awesomekling exactly the opposite, auto-generated names is the date in which the file was created, but if you specify a name let's say `shot file.png`, and then you do it again `shot file.png`, with my patch the second file will be renamed to `file(1).png` to not accidentally overwrite the first screenshot.

awesomekling: If the user runs `shot x.png`, we should respect that and create `x.png`. Changing the name automatically would violate expectations and does not match the behavior of any other program.

Userland/Utilities/shot.cpp
awesomekling: Yes it does, but if the user wants to create a file named `x` we should respect that.

r0ei: @awesomekling Okay, sounds fair. I like your way of thinking, like, giving the user freedom to do what he wants. Thank you for reviewing my pull request!

PR #5984, opened 2021-03-27
Utilities:  Added my version of 'dd'

g0mb4:
Recently started to work on Serenity and really missed this tool, so I implemented it:
Screenshot from 2021-03-29 10-13-11

### DIFF
--- a/Userland/Utilities/dd.cpp
+++ b/Userland/Utilities/dd.cpp
@@ -1,0 +1,259 @@
+/*
+ * Copyright (c) 2021, János Tóth <toth-janos@outlook.com>
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ *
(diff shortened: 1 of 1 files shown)


Userland/Utilities/dd.cpp
```diff
@@ -0,0 +1,118 @@
...
+
+int main(int argc, char** argv)
+{
+    int input_fd = 0;
+    int output_fd = 1;
+    int block_size = 512;
+    int count = -1; // infinite
+
+    // Handling heretic command-line options.
+    for (int a = 1; a < argc; a++) {
```
chengchenwish, line 150: better to use  Core::ArgsParser to parse args.

g0mb4: Ohh, I thought that ArgParser only deals with arguments in the format of "-a value" or "--arg value" and dd uses a different format.

chengchenwish: You are right, I didn't aware of this😄.

awesomekling:
Hello and thank you for working on this! Please squash little fixes into the relevant preceding commit instead of adding more commits, we don't need to preserve the fixup history and it just makes it harder to review the PR anyway :)

g0mb4:
Squashed my commits, sorry for the inconvenience :)

PR #5987, opened 2021-03-27
Kernel: More bare metal fixes, new HID subsystem

supercomputer7:
@awesomekling can you test this and see if it works on your test machine?


Kernel/Storage/BMIDEChannel.h
```diff
@@ -46,12 +46,14 @@ class BMIDEChannel final : public IDEChannel {
 
 public:
     static NonnullRefPtr<BMIDEChannel> create(const IDEController&, IDEChannel::IOAddressGroup, IDEChannel::ChannelType type);
+    static NonnullRefPtr<BMIDEChannel> create(const IDEController&, u8 irq, IDEChannel::IOAddressGroup, IDEChannel::ChannelType type);
```
Hendiadyoin1, line 49: for sake of consistency I would put the irq last, that the rest is symmetric  
same goes for line 56 and the constructor

supercomputer7: I do see your point, however, there's no real issue with doing it this way.
Also, I want to ensure that the reader of this code clearly understands that this is a different way to create a special `IDEChannel` object, because of the IRQ line being set to something different. It seemed logical to put `u8 irq` just after `const IDEController&` as it is "more important" than the other parameters after it.

danboid:
I should get to test this on Thursday.

supercomputer7:
> I should get to test this on Thursday.

Great! Thanks for the help :)

supercomputer7:
I added a few more commits to address #6074. We rely on the fact that PS/2 exists on QEMU and many other machines, but some machines (like mine) can disable the PS/2 controller entirely on boot if no device is connected to it.

awesomekling (changes requested):

Kernel/Devices/HID/HIDManagement.h
```diff
@@ -0,0 +1,84 @@
...
+
+    const String keymap_name() { return m_character_map.character_map_name(); }
+    const Keyboard::CharacterMapData& character_maps() const { return m_character_map.character_map_data(); }
+    const Keyboard::CharacterMap& character_map() const { return m_character_map; }
+    void set_client(KeyboardClient* client) { m_client = client; }
+    void set_maps(const Keyboard::CharacterMapData& character_map, const String& character_map_name);
+
+private:
+    size_t mouse_minor_number();
+    size_t keyboard_minor_number();
```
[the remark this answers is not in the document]
supercomputer7: Seems like good names for me :)

PR #5988, opened 2021-03-27
WindowServer: Add menu item fade-out animation upon activation

tomuta:
_Always felt we were missing this..._

menu_animation

### DIFF
--- a/Userland/Services/WindowServer/Menu.cpp
+++ b/Userland/Services/WindowServer/Menu.cpp
@@ -456,10 +456,60 @@ void Menu::clear_hovered_item()
     redraw();
 }
 
+void Menu::start_activation_animation(MenuItem& item)
+{
+    VERIFY(menu_window());
+    VERIFY(menu_window()->backing_store());
+    auto window = Window::construct(*this, WindowType::Menu);
+    window->set_frameless(true);
+    window->set_hit_testing_enabled(false);
+    window->set_opacity(0.8f); // start out transparent so we don't have to recompute occlusions
+    window->set_rect(item.rect().translated(m_menu_window->rect().location()));
+    window->set_event_filter([](Core::Event&) {
+        // ignore all events
+        return false;
+    });
+
+    VERIFY(window->backing_store());
+    Gfx::Painter painter(*window->backing_store());
+    painter.blit({}, *menu_window()->backing_store(), item.rect(), 1.0f, false);
+    window->invalidate();
+
+    struct AnimationInfo {
+        RefPtr<Core::Timer> timer;
+        RefPtr<Window> window;
+        u8 step { 8 }; // Must be even number!
+
+        AnimationInfo(NonnullRefPtr<Window>&& window)
+            : window(move(window))
+        {
+        }
+    };
+    auto animation = adopt_own(*new AnimationInfo(move(window)));
+    auto& timer = animation->timer;
+    timer = Core::Timer::create_repeating(50, [this, animation = animation.ptr(), animation_ref = move(animation)] {
+        VERIFY(animation->step % 2 == 0);
+        animation->step -= 2;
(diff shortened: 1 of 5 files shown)


Userland/Services/WindowServer/Window.h
```diff
@@ -146,6 +146,10 @@ class Window final : public Core::Object
     float opacity() const { return m_opacity; }
     void set_opacity(float);
 
+    void disable_hit_tests(bool value)
```
awesomekling, line 149: Please use `set_` prefix for setters and prefer addressing the positive (enabled) state: `set_hit_testing_enabled(bool)`, alternatively `set_hit_testing_disabled()` (with no way of going back)

Userland/Services/WindowServer/Menu.cpp
```diff
@@ -456,11 +456,64 @@ void Menu::clear_hovered_item()
...
+        RefPtr<Core::Timer> timer;
+        RefPtr<Window> window;
+        u8 step { 8 }; // Must be even number!
+
+        AnimationInfo(NonnullRefPtr<Window>&& window)
+            : window(move(window))
+        {
+        }
+    };
+    auto* animation = new AnimationInfo(move(window));
```
awesomekling, line 488: Can we put this in an `OwnPtr` (or `NonnullOwnPtr`) instead? Naked new is so C++98..

Userland/Services/WindowServer/Window.h
```diff
@@ -336,6 +336,12 @@ class Window final : public Core::Object
     const MenuBar* menubar() const { return m_menubar; }
     void set_menubar(MenuBar*);
 
+    template<typename EventHandler>
+    void set_event_handler(EventHandler&& event_handler)
+    {
+        m_event_handler = move(event_handler);
+    }
```
awesomekling, line 343: I added a general mechanism to `Core::Object` for this so we don't have to specialize the behavior in Window :)

`Core::Object::set_event_filter(Function<bool(Event&)>)`

You set a filter on the target object, and then the filter gets consulted during event dispatch. Return `false` from the filter and the event is not delivered.

tomuta: Ah, didn't know, will look into it, thx.

Userland/Services/WindowServer/Menu.cpp
tomuta: I tried that, and it ended up being very ugly. What happens is that when you capture a `NonnullOwnPtr` in the lambda, you end up not being able to use it without a `const_cast` or capturing an additional raw pointer and only using the captured object for lifetime purposes. I can change it back, but this looked a lot cleaner, though less elegant indeed.

Userland/Services/WindowServer/Menu.cpp
```diff
@@ -456,11 +456,61 @@ void Menu::clear_hovered_item()
...
+    };
+    auto animation = adopt_own(*new AnimationInfo(move(window)));
+    auto& timer = animation->timer;
+    timer = Core::Timer::create_repeating(50, [this, animation = animation.ptr(), animation_ref = move(animation)] {
+        VERIFY(animation->step % 2 == 0);
+        animation->step -= 2;
+        if (animation->step == 0) {
+            animation->window->set_visible(false);
+            animation->timer->stop();
+            animation->timer = nullptr; // break circular reference
```
tomuta, line 496: @awesomekling I assume this is legal to do from within the timer event handler? Seems to work, but I'm not sure if it just happens to work by accident. I know that `std::function` doesn't like being assigned/freed while it's in `operator()`...

awesomekling: It's fine since the event loop always holds a ref on the target object during event dispatch :)

PR #6002, opened 2021-03-28
LibC: Implement utimes

GMTA:
Implement the `utimes` function as part of `<sys/time.h>`. This is basically a `utime` alternative with sub-second precision, which our VFS does not yet support. Left the implementation of `tv_usec` as a FIXME.

Manpage: https://linux.die.net/man/2/utimes

Adding this function removes the necessity of an SQLite compilation flag, so also remove that.


alimpfard:
`utimes` is deprecated in POSIX, maybe instead implement `utimensat`?

https://pubs.opengroup.org/onlinepubs/9699919799/functions/utime.html

GMTA:
@alimpfard This was the minimum change to get rid of the patch and it's not deprecated indeed. I suggest leaving the `utimensat` implementation for another time :-)

Issue #6004, opened 2021-03-29
HackStudio: wrnline() causes Assert in vout()


ElectrodeYT:
When opening a .html file in /www this assert can be seen:

``
repeat alot...
HackStudio(35): vout: file: 0x1cc487d8 newline true static_cast<size_t>(retval) 0 string.length() 89
HackStudio(35): USERSPACE(35) ASSERTION FAILED: static_cast<size_t>(retval) == string.length()
../../AK/Format.cpp:650
repeat alot...
``

CrashDaemon:

``
repeat alot...
CrashDaemon(25): 0x4ee2dc6f: [libc.so] AK::vout(FILE*, AK::StringView, AK::TypeErasedFormatParams, bool) +0x1ff (Format.cpp:651)
CrashDaemon(25): 0x4edfbeb1: [libc.so] __assertion_failed +0x171 (Format.h:375)
repeat alot...
``

![image](https://user-images.githubusercontent.com/49659190/112831007-9b001c00-9093-11eb-9b01-2aa9bf8a9e99.png)

ElectrodeYT:
I tried to reference the code in ProjectFile.cpp that calls wrnline that causes this, but it didnt seem to work.

Its here:
void ProjectFile::create_document_if_needed() const
{
    if (m_document)
        return;

    m_document = CodeDocument::create(m_name);
    auto file_or_error = Core::File::open(m_name, Core::File::ReadOnly);
    if (file_or_error.is_error()) {

        warnln("Couldn't open '{}': {}", m_name, file_or_error.error()); << Assert in vout() comes from here
        // This is okay though, we'll just go with an empty document and create the file when saving.
        return;
    }

    auto& file = *file_or_error.value();
    m_could_render_text = m_document->set_text(file.read_all());
}


There is only one assert in vout:


#ifndef KERNEL
void vout(FILE* file, StringView fmtstr, TypeErasedFormatParams params, bool newline)
{
    StringBuilder builder;
    vformat(builder, fmtstr, params);

    if (newline)
        builder.append('\n');

    const auto string = builder.string_view();
    const auto retval = ::fwrite(string.characters_without_null_termination(), 1, string.length(), file);
    VERIFY(static_cast<size_t>(retval) == string.length()); << Assert here
}
#endif

ElectrodeYT:
I played around a bit, and i found out that HackStudio is a lot glitchier then i thought; for example, opening HackStudio  while SystemMonitor is open will crash systemmonitor.

[...]

Issue #6003, opened 2021-03-28
GCC/G++: libstdc++.a/libgcc.c file format not recognized


ElectrodeYT:
Compiling the files in the image causes the linker to not recognize the format of the file libstdc++ or libgcc.a.
![image](https://user-images.githubusercontent.com/49659190/112770922-73746980-9029-11eb-88da-41c3b8ec3592.png)

alimpfard:
I remember this happening before, that time it was `scanf` being broken: #5365.

ElectrodeYT:
Above pull request fixes it to the extent that the error does not get printed, but causes a Syscall from non-syscall region.
https://gist.github.com/ElectrodeYT/b401ab3aa6af1d98d87fea8b012f23d8

alimpfard:
Looks like something in Loader.so is trying to make a syscall?
Maybe this https://github.com/alimpfard/serenity/commit/b68f25d289880150c6689ecff901e221b96a1c3a#diff-785fb88919ba484bcd7d691028c4358ef774e29ca93df8bc9d96d540feb1dd61R199 could help a little by at least saying what the syscall is - though I suspect it is either `dbgputstr` or `get_process_name`, used by AK::Format.

Currently the loader cannot write anything to debug after it has `msyscall`'d libsystem.so (so basically after calling `main`)

IdanHo:
> Above pull request fixes it to the extent that the error does not get printed, but causes a Syscall from non-syscall region.

I cant reproduce the syscall problem, im still just getting `file format not recognized`

ElectrodeYT:
I'll recompile serenity, I wonder if something broke on my end?

ElectrodeYT:
Yeah, sorry about that syscall thing, seems like it was something about the branch that messed it up. It still has the file format error though. Makes me wonder if there is a bug in Loader.so though?

Hendiadyoin1:
can you try to build these libraries without stripping them, there are some issues apperantly caused by that

awesomekling:
Looks like this broke somehow with @boricj's recent changes for 64-bit file offsets.

boricj:
I'd first make sure that libgcc.a and libstdc++.a have the same contents inside SerenityOS and outside with SHA256, just in case. It could be a bug in the storage stack where we return garbage inside either the device driver or the file system.

awesomekling:
[...]

PR #6006, opened 2021-03-29
LibC+Kernel: Fix make port

itamar8910:
This fixes the `make` port.
The regression in the kernel was probably introduced in d48666489c83b7f0922537ff2f3123644d17201e


ElectrodeYT:
Fixes #6004. Partially fixes #6003, but just causes a Segfault with a bunch of write calls in the Backtrace (could also be me disabling SMP while trying to diagnose #6004

PR #6011, opened 2021-03-29
LibWeb: Compute position of relative block elements

trflynn89:
Another rending fix for xkcd.com - the div containing the comic had an incorrect 'left' position.

Before:
before

After:
after

You might notice the width of the comic div is a few px wider than the row above it. Turns out that isn't a LibWeb issue, that's how it looks on other browsers as well if you remove rounded corners from the CSS :)

### DIFF
--- a/Userland/Libraries/LibWeb/Layout/BlockFormattingContext.cpp
+++ b/Userland/Libraries/LibWeb/Layout/BlockFormattingContext.cpp
@@ -351,6 +351,40 @@ void BlockFormattingContext::compute_height(Box& box)
     box.set_height(height);
 }
 
+void BlockFormattingContext::compute_position(Box& box)
+{
+    // 9.4.3 Relative positioning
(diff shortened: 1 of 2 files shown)


Userland/Libraries/LibWeb/Layout/BlockFormattingContext.cpp
```diff
@@ -351,6 +351,40 @@ void BlockFormattingContext::compute_height(Box& box)
     box.set_height(height);
 }
 
+void BlockFormattingContext::compute_position(Box& box)
+{
+    // 9.4.3 Relative positioning
```
awesomekling, line 356: We're not checking that `box.computed_values().position() == CSS::Position::Relative`. This sounds like it should only apply in those cases(?)

trflynn89: Ah that's right - I misread the text in 10.3 that refers to 9.4.3, and thought it applied to more contexts. Will fix.

PR #6012, opened 2021-03-29
LibGUI: Add transient option to show dotfiles in FilePicker

trflynn89:
This is particularly useful when wanting to open files in ~/.config
from the Text Editor. The option is currently not persistent, but could
be hooked into File Manager's configuration.


awesomekling:
Great idea!

PR #6018, opened 2021-03-29
Check that new_offset in FileDescription::seek is not larger than file size

BluuWasTaken:


IdanHo:
wouldn't this fail for non inode-backed files? (aka block devices)

awesomekling:
Seeking past the end of a regular file is allowed by POSIX.

PR #6025, opened 2021-03-30
LibWeb: Compute space for inline blocks with absolute positions

trflynn89:
Totally up for discussion on whether this is the right approach. Details on this change are in the commit message. This change didn't seem to hurt the existing floating and inline box tests linked on welcome.html, and it quite nicely improves xkcd.com, but of course there may be something I missed :)

Before:
before

After:
after

### DIFF
--- a/Userland/Libraries/LibWeb/Layout/BlockFormattingContext.cpp
+++ b/Userland/Libraries/LibWeb/Layout/BlockFormattingContext.cpp
@@ -406,16 +406,17 @@ void BlockFormattingContext::layout_block_level_children(Box& box, LayoutMode la
         }
 
         compute_width(child_box);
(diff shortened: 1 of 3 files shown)


trflynn89:
Yeah that makes sense. Though I was looking at this particular layout a bit more earlier, I think the issue here is actually much simpler than I thought!

The "top" container on xkcd has an `overflow` value of `hidden`, and it turns out \*that\* is what makes the "bottom" container get laid out beneath the top container, rather than inline. Implementing the `FIXME: Block elements where overflow has a value other than visible and clip.` in `FormattingContext::creates_block_formatting_context` fixes this particular layout :)

I'll open another PR and close this one

Issue #184, opened 2019-06-04
Userland: Implement "traceroute"


awesomekling:
It would be cool to have a traceroute program. If I understand correctly, the way it works is that you send one ICMP packet at a time to your target, incrementing the TTL by 1 each time. This means that every packet will get rejected one hop further into the route. The rejecting router will send back a "TE" packet and that's how we identify that hop.

PR #6030, opened 2021-03-30
Utilities: Add simple traceroute utility

IdanHo:
This utility traces the route packets take to a user specified host.

QEMU user networking (the type of networking we currently have setup) does not support sending "real" raw IPv4 packets, and as such we cant specify a custom TTL value. As a result, this utility will only work on real hardware, qemu setup with tun networking (requires root and setting up a tun adapter) and other hypervisors that support bridged adapters (VirtualBox/VMWare).

image


linusg:
"Fixes #184" :^)

Userland/Utilities/traceroute.cpp
```diff
@@ -0,0 +1,175 @@
...
+        warnln("Failed to drop privileges");
+        return 1;
+    }
+
+    if (pledge("stdio inet unix", nullptr) < 0) {
+        perror("pledge");
+        return 1;
+    }
+
+    timeval timeout { 5, 0 };
```
linusg, line 94: Can we make this an option? should be easy enough

Userland/Utilities/traceroute.cpp
```diff
@@ -0,0 +1,175 @@
...
+
+    // 1: reached target
+    // 0: got ttl exhausted response
+    // -1: error or no response
+    auto try_reach_host = [&](int ttl) -> int {
+        Core::ElapsedTimer m_timer { true };
+        for (auto i = 0; i < max_retries; i++) {
+            icmp_request request {};
+            request.header = { ICMP_ECHO, 0, 0, { { 0, 0 } } };
+            VERIFY(String::number(ttl).copy_characters_to_buffer(request.msg, sizeof(request.msg)));
```
ADKaster, line 119: I know it's not kosher to do things with side effects in an assertion, but does the same rule apply to VERIFY?

awesomekling: It technically doesn't apply since `VERIFY` never gets compiled out.
However, it looks really odd so let's not do it :)

Userland/Utilities/traceroute.cpp
```diff
@@ -0,0 +1,175 @@
...
+                continue;
+
+            auto* peer = gethostbyaddr(&peer_address.sin_addr, sizeof(peer_address.sin_addr), AF_INET);
+
+            const char* peer_name;
+            if (peer) {
+                peer_name = peer->h_name;
+            } else {
+                char addr_buf[INET_ADDRSTRLEN];
+                peer_name = inet_ntop(AF_INET, &peer_address.sin_addr, addr_buf, sizeof(addr_buf));
```
ADKaster, line 147: > The inet_ntop() function shall return a pointer to the buffer containing the text string if the conversion succeeds, and NULL otherwise, and set errno to indicate the error.

The else branch here returns a pointer into the extra-scoped addr_buf whose lifetime ends at the end of the else statement. I suggest moving the declaration of that buffer up next to peer_name, or making peer_name an AK::String.

IdanHo: good catch, fixed :)

PR #6031, opened 2021-03-30
LibWeb: Create a new BFC when the overflow is neither visible nor clip

trflynn89:
This fixes the layout of the middle box containing the comic on xkcd.com.

Before:
before

After:
after


awesomekling:
Downright surgical :star:

PR #6033, opened 2021-03-31
shot: Improved Error Codes and Option to Screenshot Active Window

fraang:


Userland/Services/WindowServer/WindowFrame.cpp
```diff
@@ -383,12 +383,14 @@ void WindowFrame::paint(Gfx::Painter& painter, const Gfx::IntRect& rect)
     }
 }
 
-void WindowFrame::paint_completely(Gfx::Painter& painter) {
+void WindowFrame::paint_completely(Gfx::Painter& painter)
+{
```
awesomekling: Tons of unrelated formatting changes in this commit. Please fix formatting in the earlier commit instead :)

fraang: Is fixed now. :-)

Userland/Services/WindowServer/Compositor.h
```diff
@@ -110,6 +111,9 @@ class Compositor final : public Core::Object {
     OwnPtr<Gfx::Painter> m_front_painter;
     OwnPtr<Gfx::Painter> m_temp_painter;
 
+    RefPtr<Gfx::Bitmap> m_active_window_bitmap;
+    OwnPtr<Gfx::Painter> m_active_window_painter;
```
awesomekling, line 115: Hmm, I don't think we need to keep these images around here in WindowServer.

Instead, let's create them on the fly and have `bitmap_of_active_window()` return a `RefPtr<Gfx::Bitmap>`.

Userland/Services/WindowServer/Compositor.cpp
```diff
@@ -120,6 +120,38 @@ void Compositor::did_construct_window_manager(Badge<WindowManager>)
...
+    const auto active_window = wm.active_window();
+
+    auto& active_window_frame = active_window->frame();
+    const auto active_window_frame_render_rect = active_window_frame.render_rect();
+    const auto size = Gfx::IntSize { active_window_frame_render_rect.width(), active_window_frame_render_rect.height() };
+
+    auto& screen = Screen::the();
+    const auto scale_factor = screen.scale_factor();
+
+    m_active_window_bitmap = Gfx::Bitmap::create(Gfx::BitmapFormat::BGRA8888, size, scale_factor);
```
awesomekling, line 136: Creating a bitmap can fail if we run out of memory. We need to null-check the result of `Gfx::Bitmap::create()`

PR #6034, opened 2021-03-31
HackStudio: Actions on project tree selections improved

### LINKED ISSUE #5935: HackStudio: A new file or directory in HackStudio is always in the top of the project.
### Reproduce
Open HackStudio and select a sub directory in the project tree view. Select the new file action or new directory action. After creation, the new file or directory will be in the root of the project neglecting the selected sub directory. This is not intuitive. I believe it has to do with relative paths, the created file is not respecting the relative path of the selected directory. 

Also, I am unsure if this should get its own issue, or perhaps it is an unimplemented feature but dragging files between folders in the project view does not work.

### Proposed solution
Update new_file_action & new_directory_action to respect tree selection. Maybe we create a new private function in HackStudioWidget that retrieves the relative directory of the current project tree selection as this code will probably be used often.

AlgotAxelzon:
Creating a file while having a file or directory selected will now
create the file in the selected directory, or in same directory as
a selected file.
Creating directories works the same way.

Right click -> Open on selected files will now open the selected files.

Deleting selected files will now delete the selected files.

As suggested by sagefarrenholz this was accomplished by adding a private
function HackStudioWidget::selected_file_paths() that returns a
Vector<LexicalPath> of the current project tree selection.

(Fixes #5935)

This is my first contribution so there is probably something that is in need
of change!

### DIFF
--- a/Userland/DevTools/HackStudio/HackStudioWidget.cpp
+++ b/Userland/DevTools/HackStudio/HackStudioWidget.cpp
@@ -202,11 +202,21 @@ void HackStudioWidget::open_project(const String& root_path)
     }
 }
 
-Vector<String> HackStudioWidget::selected_file_names() const
+Vector<String> HackStudioWidget::selected_file_paths() const
 {
     Vector<String> files;
     m_project_tree_view->selection().for_each_index([&](const GUI::ModelIndex& index) {
-        files.append(index.data().as_string());
+        String sub_path = index.data().as_string();
+
+        GUI::ModelIndex parent_or_invalid = index.parent();
+
+        while (parent_or_invalid.is_valid()) {
+            sub_path = String::formatted("{}/{}", parent_or_invalid.data().as_string(), sub_path);
+
+            parent_or_invalid = parent_or_invalid.parent();
+        }
+
+        files.append(sub_path);
     });
     return files;
 }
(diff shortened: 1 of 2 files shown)


Userland/DevTools/HackStudio/HackStudioWidget.h
```diff
@@ -73,6 +73,7 @@ class HackStudioWidget : public GUI::Widget {
 
 private:
     static String get_full_path_of_serenity_source(const String& file);
+    Vector<LexicalPath> selected_file_paths() const;
```
itamar8910, line 76: `LexcialPath` is meant to be used as a utility class for parsing & validation paths, not really as a storage class for paths.
For simply storing paths & passing them around, please prefer to use a simple `String` class.

itamar8910: Also, after adding `selected_file_paths`, it seems that `selected_file_names` is no longer used. If this is the case, we can delete it.

Userland/DevTools/HackStudio/HackStudioWidget.cpp
```diff
@@ -211,6 +211,30 @@ Vector<String> HackStudioWidget::selected_file_names() const
...
+{
+    Vector<LexicalPath> files;
+    m_project_tree_view->selection().for_each_index([&](const GUI::ModelIndex& index) {
+        String sub_path = index.data().as_string();
+
+        GUI::ModelIndex parent_or_invalid = index.parent();
+
+        while (parent_or_invalid.is_valid()) {
+            StringBuilder path_from_parent;
+            path_from_parent.appendff("{}/{}", parent_or_invalid.data().as_string(), sub_path);
```
itamar8910, line 224: Instead of using a `StringBuilder` here, we can simply use `String::formatted`

Userland/DevTools/HackStudio/HackStudioWidget.cpp
```diff
@@ -211,6 +211,30 @@ Vector<String> HackStudioWidget::selected_file_names() const
...
+
+            sub_path = path_from_parent.build();
+
+            parent_or_invalid = parent_or_invalid.parent();
+        }
+        StringView sub_path_sv(sub_path);
```
itamar8910, line 230: `StringView` can be implicitly constructed from a `String`, so there's no need to create the `sub_path_sv` local variable, you can simply do `LexicalPath path(sub_path)`

Userland/DevTools/HackStudio/HackStudioWidget.cpp
```diff
@@ -334,6 +380,23 @@ NonnullRefPtr<GUI::Action> HackStudioWidget::create_new_directory_action()
...
+        if (!path_to_selected.is_empty()) {
+            String dir_path;
+            StringBuilder builder;
+
+            if (Core::File::is_directory(path_to_selected.first().string()))
+                dir_path = path_to_selected.first().string();
+            else
+                dir_path = path_to_selected.first().dirname();
+
+            builder.appendff("{}/{}", dir_path, directory_name);
```
itamar8910, line 395: We can use `String::formatted` here

itamar8910 (changes requested):
Thank you for contributing and improving HackStudio!
Please see my comments.

Also, a small nitpick regarding the commit message, we prefer to use imperative language in commit messages. So a more appropriate title for the commit message would be: `HackStudio: Make actions with project tree selections work as expected `

AlgotAxelzon:
Thanks a lot for the feedback, it made me learn a few things!
I have made changes related to your comments; using less StringBuilders, using Strings to pass around paths, and removed unused functions.

After a bit of struggling I also managed to make the git history look decent 😉 

Please tell me if there is something more that is needed to be done.

itamar8910 (approved):
Okay, looks good to me :+1:

PR #6037, opened 2021-03-31
TTY+Kernel: Added TIOCSTI ioctl placeholder.

Baitinq:
Added the TIOCSTI ioctl, which inserts a byte in the TTY's input queue.
This can be used to simulate input (Could have some security
implications). Initially added it because cmatrix uses it and I am
planning to port it.


awesomekling:
We should not support `TIOCSTI` due to the huge security problems. However, it would be fine to add the `#define` and make it fail with `EIO` for compatibility.

Baitinq:
Okay, that seems like the best thing to do. Will update my commit :)

Baitinq:
@awesomekling what about a promise for this ioctl? would that even make sense?

awesomekling:
> @awesomekling what about a promise for this ioctl? would that even make sense?

No, we should simply not implement it as it's inherently insecure :)

PR #6050, opened 2021-04-01
Ports: Add flatbuffers library

arthurmco:
The flatbuffers library is a serialization library, created by Google for game development and performance-critical applications.
It aims to be fast and efficient.

This commit creates a port of it to SerenityOS. 
Although this library is not much popular among the bigger opensource projects, there are [some uses of it here on Github](https://github.com/search?l=C%2B%2B&q=FlatBufferBuilder&type=Code). If someone happens to port any small game that uses flatbuffers, it would be one less dependency for the person to port

The flatbuffers build process generates three things: some header files, a library (libflatbuffers) and a schema compiler (flatc).

------

A small caveat of this port: the library comes with a test suite, but we cannot run it, because it attempts to run the flatc executable we are compiling to SerenityOS in the host system. The host system does not recognize the file format, so it fails.
Nevertheless, the port compiles and runs: the schema compiler runs and the library links successfully

### DIFF
--- a/Ports/flatbuffers/patches/flatbuffers-1.12.0.patch
+++ b/Ports/flatbuffers/patches/flatbuffers-1.12.0.patch
@@ -1,0 +1,35 @@
+diff -ruN flatbuffers-1.12.0/src/code_generators.cpp flatbuffers-1.12.0-serenity/src/code_generators.cpp
+--- flatbuffers-1.12.0/src/code_generators.cpp	2020-03-12 19:33:39.000000000 -0300
++++ flatbuffers-1.12.0-serenity/src/code_generators.cpp	2021-03-31 20:39:12.000000000 -0300
+@@ -23,6 +23,17 @@
+ #include "flatbuffers/base.h"
+ #include "flatbuffers/util.h"
+ 
++#if defined(__serenity__)
++  // We do not have those functions inside std namespace...
++
++namespace std {
++  auto isnan(double x) { return ::isnan(x); }
++  auto isinf(double x) { return ::isinf(x); }
++}
++
++#endif
++
++
+ #if defined(_MSC_VER)
+ #  pragma warning(push)
+ #  pragma warning(disable : 4127)  // C4127: conditional expression is constant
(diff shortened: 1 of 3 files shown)


Ports/flatbuffers/patches/flatbuffers-1.12.0.patch
```diff
@@ -0,0 +1,35 @@
...
+diff -ruN flatbuffers-1.12.0/src/util.cpp flatbuffers-1.12.0-serenity/src/util.cpp
+--- flatbuffers-1.12.0/src/util.cpp	2020-03-12 19:33:39.000000000 -0300
++++ flatbuffers-1.12.0-serenity/src/util.cpp	2021-03-31 20:28:59.000000000 -0300
+@@ -189,6 +189,10 @@
+ std::string AbsolutePath(const std::string &filepath) {
+   // clang-format off
+ 
++  #ifdef __serenity__
++  #define PATH_MAX 4096
++  #endif
```
awesomekling, line 31: The correct way (according to Dr. POSIX) to get `PATH_MAX` is to make sure `<limits.h>` is included.

arthurmco: Oh, sorry, I forgot about `<limits.h>`...

But a weird thing: the file I patched already includes this header

```cpp
#ifdef _WIN32
#  ifndef WIN32_LEAN_AND_MEAN
#    define WIN32_LEAN_AND_MEAN
#  endif
#  ifndef NOMINMAX
#    define NOMINMAX
#  endif
#  ifdef _MSC_VER
#    include <crtdbg.h>
#  endif
#  include <windows.h> 
#  include <direct.h>
#  include <winbase.h>
#  undef interface 
#else
#  define _XOPEN_SOURCE 600  
#  include <limits.h>  // <----------
#endif
// clang-format on
```
(the complete file is [here](https://github.com/google/flatbuffers/blob/v1.12.0/src/util.cpp#L20-L38) )

------

I will try to compile a test program to investigate

arthurmco: I investigated. It was something with the toolchain
I recompiled it and it worked

PR #6054, opened 2021-04-01
AK::Result change release_error return value

vcollette:
The function was not used anywhere so the error was unoticed


tomuta:
Woah, this feels like some g++ warning option should catch?

ADKaster:
Unless someone added another use I think the only place the void specialization of Result is used is in LibThread? Don't think anyone calls release_error in that context, so it's probably never been instantiated.

PR #6055, opened 2021-04-01
Ports: Added a SDL2_image port

Baitinq:


Baitinq:
EDIT: Fixed the install location of the headers. It was being install in /usr/local/SDL2 when we use /usr/SDL2

Baitinq:
EDIT: Changed the install location of the library so it installs under /usr/include/SDL2 instead of /usr/local/include/SDL2, as that is where all of serenity's SDL2 headers are located.

PR #6059, opened 2021-04-01
Ports: Build Python with --disable-ipv6

linusg:
The addition of some IPv6 related things makes the configure script think we support it now. We don't.


Jorropo (approved):

awesomekling:
c o n f l i c t

linusg:
r e s o l v e d

Issue #964, opened 2019-12-30
Kernel: mprotect should allow page-level access bits


ADKaster:
The current implementation of mmap/mprotect allocates one Region for the mmap call, and RWX can only be set on the entire region at once.

For uses like loading an ELF image into contiguous memory, it might be useful to mprotect with page granularity instead of regions. In fact, the man page for mprotect on linux (for glibc anyway) shows that Linux supports this type of protection in its example.

This would probably require re-factoring the entire region code though :/

awesomekling:
For what it's worth, `munmap()` already supports region splitting if you call it to unmap a partial range.

It probably wouldn't be terribly horrible to take that code and factor it into something reusable :)

ADKaster:
Related to #175

ADKaster:
Partially implemented with #965

Hendiadyoin1:
Is this fixed now?
Or do we still need to do more stuff here?

IdanHo:
This is fully implemented as far as I can tell

PR #6064, opened 2021-04-01
Kernel: Implement multiple Region mprotect

### LINKED ISSUE #177: Kernel: Implement mprotect()
We should support altering the protection of VM ranges after they've been created. This will get a bit complicated, since protection bits are currently per-region, so `mprotect()` will need to know how to merge and split regions.

See: http://pubs.opengroup.org/onlinepubs/9699919799/functions/mprotect.html

Hendiadyoin1:
We were able to do this with munmap already, and for mprotect its just copy and paste from there

Fixes #177 and maybe the rest of #964

### DIFF
--- a/Kernel/Syscalls/mmap.cpp
+++ b/Kernel/Syscalls/mmap.cpp
@@ -357,7 +357,67 @@ KResultOr<int> Process::sys$mprotect(Userspace<void*> addr, size_t size, int pro
         return 0;
     }
 
-    // FIXME: We should also support mprotect() across multiple regions. (#175) (#964)
+    if (const auto& regions = space().find_regions_intersecting(range_to_mprotect); regions.size()) {
+        size_t full_size_found = 0;
+        // first check before doing anything
+        for (const auto* region : regions) {
+            if (!region->is_mmap())
+                return EPERM;
+            if (!validate_mmap_prot(prot, region->is_stack(), region->vmobject().is_anonymous(), region))
+                return EINVAL;
+            if (region->access() == prot_to_region_access_flags(prot))
+                return 0;
+            if (region->vmobject().is_inode()
+                && !validate_inode_mmap_prot(*this, prot, static_cast<const InodeVMObject&>(region->vmobject()).inode(), region->is_shared())) {
+                return EACCES;
+            }
+            full_size_found += region->range().intersect(range_to_mprotect).size();
+        }
+
+        if (full_size_found != range_to_mprotect.size())
+            return ENOMEM;
+
+        // then do all the other stuff
+        for (auto* old_region : regions) {
+            const auto intersection_to_mprotect = range_to_mprotect.intersect(old_region->range());
+            // full sub region
+            if (intersection_to_mprotect == old_region->range()) {
+                old_region->set_readable(prot & PROT_READ);
+                old_region->set_writable(prot & PROT_WRITE);
+                old_region->set_executable(prot & PROT_EXEC);
+
+                old_region->remap();
+                continue;
+            }
+            // Remove the old region from our regions tree, since were going to add another region
+            // with the exact same start address, but dont deallocate it yet
+            auto region = space().take_region(*old_region);
+            VERIFY(region);
+
+            // Unmap the old region here, specifying that we *don't* want the VM deallocated.
+            region->unmap(Region::ShouldDeallocateVirtualMemoryRange::No);
+
+            // This vector is the region(s) adjacent to our range.
+            // We need to allocate a new region for the range we wanted to change permission bits on.
+            auto adjacent_regions = space().split_region_around_range(*old_region, intersection_to_mprotect);
+            // there should only be one
+            VERIFY(adjacent_regions.size() == 1);
+
+            size_t new_range_offset_in_vmobject = old_region->offset_in_vmobject() + (intersection_to_mprotect.base().get() - old_region->range().base().get());
+            auto& new_region = space().allocate_split_region(*region, intersection_to_mprotect, new_range_offset_in_vmobject);
+            new_region.set_readable(prot & PROT_READ);
+            new_region.set_writable(prot & PROT_WRITE);
+            new_region.set_executable(prot & PROT_EXEC);
+
+            // Map the new region using our page directory (they were just allocated and don't have one) if any.
+            if (adjacent_regions.size())
+                adjacent_regions[0]->map(space().page_directory());
+
+            new_region.map(space().page_directory());
+        }
+
+        return 0;
+    }
 
     return EINVAL;
 }
--- a/Tests/Kernel/mprotect-multi-region-mrotect.cpp
+++ b/Tests/Kernel/mprotect-multi-region-mrotect.cpp
@@ -1,0 +1,75 @@
+/*
+ * Copyright (c) 2021, the SerenityOS developers.
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+#include <AK/Format.h>
+#include <AK/Types.h>
+#include <fcntl.h>
+#include <stdio.h>
+#include <string.h>
+#include <sys/mman.h>
+#include <unistd.h>
+
+int main()
+{
+    printf("Testing full unnmap\n");
+    auto* map1 = mmap(nullptr, 2 * PAGE_SIZE, PROT_READ | PROT_WRITE, MAP_ANONYMOUS | MAP_PRIVATE | MAP_FIXED, 0, 0);
+    if (map1 == MAP_FAILED) {
+        perror("mmap 1");
+        return 1;
+    }
+    auto* map2 = mmap((void*)((FlatPtr)map1 + 2 * PAGE_SIZE), 2 * PAGE_SIZE, PROT_READ | PROT_WRITE, MAP_ANONYMOUS | MAP_PRIVATE | MAP_FIXED, 0, 0);
+    if (map2 == MAP_FAILED) {
+        perror("mmap 2");
+        return 1;
+    }
+    auto* map3 = mmap((void*)((FlatPtr)map1 + 4 * PAGE_SIZE), 2 * PAGE_SIZE, PROT_READ | PROT_WRITE, MAP_ANONYMOUS | MAP_PRIVATE | MAP_FIXED, 0, 0);
+    if (map3 == MAP_FAILED) {
+        perror("mmap 3");
+        return 1;
+    }
+
+    // really allocating pages
+    memset(map1, 0x01, 6 * PAGE_SIZE);
+
+    int rc;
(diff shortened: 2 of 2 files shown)


Hendiadyoin1:
oops messed up the history

Kernel/Syscalls/mmap.cpp
```diff
@@ -363,7 +363,48 @@ KResultOr<int> Process::sys$mprotect(Userspace<void*> addr, size_t size, int pro
...
+            // Unmap the old region here, specifying that we *don't* want the VM deallocated.
+            old_region->unmap(Region::ShouldDeallocateVirtualMemoryRange::No);
+            space().deallocate_region(*old_region);
+
+            // Map the new region using our page directory (they were just allocated and don't have one) if any.
+            if (adjacent_regions.size())
+                adjacent_regions[0]->map(space().page_directory());
+
+            new_region.map(space().page_directory());
+        }
```
ADKaster, line 406: I think you're missing a return 0 here?

Hendiadyoin1: ohh ehm yes, I will fix tomorow

Hendiadyoin1: should be fixed

Hendiadyoin1:
some things before merge:  
1. Do we need to check first, or are we allowed to fail half way through?
2. Do we need to assert full coverage, or is it allowed, that there is a not mapped region in between?
3. the issue talked about merging, what is meant with that, are we supposed to merge these multiple regions if we can or do we need to assert if we can etc...

Kernel/Syscalls/mmap.cpp
```diff
@@ -363,7 +363,50 @@ KResultOr<int> Process::sys$mprotect(Userspace<void*> addr, size_t size, int pro
...
+            if (region->access() == prot_to_region_access_flags(prot))
+                return 0;
+            if (region->vmobject().is_inode()
+                && !validate_inode_mmap_prot(*this, prot, static_cast<const InodeVMObject&>(region->vmobject()).inode(), region->is_shared())) {
+                return EACCES;
+            }
+        }
+
+        // then do all the other stuff
+        for (auto* old_region : regions) {
```
ADKaster, line 384: I'm not convinced this is correct.. some MAP_FIXED tests are in order I think.  I Would want to verify that all the edge cases work out, such as:

1) middle of one region into the middle of another region 
2) middle of one region , past a second region , into the middle of a third region
3) beginning of one region into the middle of a second region
4) begining of one region to the end of a second region
5) beginning of one region, through a second region, to the end of a third region
...
etc.

ADKaster (commented):
I could be overthinking this, but I definitely think you'll want to write some tests to verify the behavior the manpage is describing.

I think the munmap code is mostly concerned with creating holes. But the mprotect code is concerned with changing the properties of regions in-place, so I'm not sure all the same assumptions hold as the munmap loop over regions.

Kernel/Syscalls/mmap.cpp
```diff
@@ -363,7 +363,50 @@ KResultOr<int> Process::sys$mprotect(Userspace<void*> addr, size_t size, int pro
         return 0;
     }
 
-    // FIXME: We should also support mprotect() across multiple regions. (#175) (#964)
+    const auto& regions = space().find_regions_intersecting(range_to_mprotect);
+    // FIXME: what do we do if we have a not mapped page in the range?
```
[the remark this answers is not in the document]
Hendiadyoin1: With the missing region and a non existing region I have checekd with [this](https://pastebin.com/FgU3ajMm) and bith state "Cannot allocate Memmory" the code is -1

Kernel/Syscalls/mmap.cpp
```diff
@@ -363,7 +363,50 @@ KResultOr<int> Process::sys$mprotect(Userspace<void*> addr, size_t size, int pro
...
+    if (regions.size()) {
+        // first check before doing anything
+        for (auto* region : regions) {
+            if (!region->is_mmap())
+                return EPERM;
+            if (!validate_mmap_prot(prot, region->is_stack(), region->vmobject().is_anonymous(), region))
+                return EINVAL;
+            if (region->access() == prot_to_region_access_flags(prot))
+                return 0;
+            if (region->vmobject().is_inode()
```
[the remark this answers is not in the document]
Hendiadyoin1: mprotect does not merge (at this point Idk how other systems handle it and It probably is difficult to check)  
and when it is allowed to merge there should be a logicm that it only allows some types of merges  
and munmap carves out what it is supposed to unmap, so the rest of a mapping stays intact

PS:
munmap(ptr, size, prot, flag, fd, off)

Kernel/Syscalls/mmap.cpp
Hendiadyoin1: Just reading the code again, There is an Assertion error on a full match while carving everything out

Hendiadyoin1: I've fixed that and may write tests later

Kernel/Syscalls/mmap.cpp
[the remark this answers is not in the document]
Hendiadyoin1: I check the match size now and if it does not add up return EINVAL

Hendiadyoin1:
Gonna hve to rebase to make it work with the red and black tree that is now in place

Kernel/Syscalls/mmap.cpp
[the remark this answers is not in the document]
Hendiadyoin1: There is now a test for it all (no merging still)

Kernel/Syscalls/mmap.cpp
Hendiadyoin1: There is now a test implemented

Hendiadyoin1:
Everything should be working now

Userland/Tests/Kernel/mprotect-multi-region-mrotect.cpp
```diff
@@ -0,0 +1,95 @@
...
+#include <fcntl.h>
+#include <stdio.h>
+#include <string.h>
+#include <sys/mman.h>
+#include <unistd.h>
+
+int main()
+{
+    printf("Testing full unnmap\n");
+    auto* map1 = mmap(nullptr, 2 * PAGE_SIZE, PROT_READ | PROT_WRITE, MAP_ANONYMOUS | MAP_PRIVATE | MAP_FIXED, 0, 0);
```
ADKaster, line 38: This file will need moved to Tests/Kernel, after Brian's changes earlier this week.

Kernel/Syscalls/mmap.cpp
```diff
@@ -367,7 +367,67 @@ KResultOr<int> Process::sys$mprotect(Userspace<void*> addr, size_t size, int pro
         return 0;
     }
 
-    // FIXME: We should also support mprotect() across multiple regions. (#175) (#964)
+    if (const auto& regions = space().find_regions_intersecting(range_to_mprotect); regions.size()) {
+        size_t full_size_found = 0;
+        // first check before doing anything
+        for (const auto* region : regions) {
+            if (!region->is_mmap())
```
ADKaster, line 374: I would expect to see a check similar to:

```c++
        // first check before doing anything that all the intersecting regions are mapped from the same vmobject
        const VMObject* vm_obj = nullptr;
        for (const auto* region : regions) {
            if (!vm_obj) {
                vm_obj = &region->vmobject();
            } else if (&region->vmobject() != vm_obj) {
                return EINVAL;
            }
```
to make sure that all the regions we're trying to potentially merge or change perms of were all from the same vmobject originally.

Userland/Tests/Kernel/mprotect-multi-region-mrotect.cpp
```diff
@@ -0,0 +1,95 @@
...
+    if (map1 == MAP_FAILED) {
+        perror("mmap 1");
+        return 1;
+    }
+    auto* map2 = mmap((void*)((FlatPtr)map1 + 2 * PAGE_SIZE), 2 * PAGE_SIZE, PROT_READ | PROT_WRITE, MAP_ANONYMOUS | MAP_PRIVATE | MAP_FIXED, 0, 0);
+    if (map2 == MAP_FAILED) {
+        perror("mmap 2");
+        return 1;
+    }
+    auto* map3 = mmap((void*)((FlatPtr)map1 + 4 * PAGE_SIZE), 2 * PAGE_SIZE, PROT_READ | PROT_WRITE, MAP_ANONYMOUS | MAP_PRIVATE | MAP_FIXED, 0, 0);
```
ADKaster, line 48: This isn't quite the test I was thinking of. I would expect this test with the 3 discrete mmap call to fail.  If the 6 pages were originally from the same 6-page MAP_ANONYMOUS and mprotect calls had split up the regions in the kernel, then I would expect the mprotects below to pass. (once my other comment is in).

ADKaster (changes requested):
Sorry for letting this almost go stale! Few things that I think should happen before this goes in.

Kernel/Syscalls/mmap.cpp
Hendiadyoin1: I just looked at the [manpage](https://man7.org/linux/man-pages/man2/mprotect.2.html) again, and it doesn't really seem to specifiy, that you aren't allowed to do that, and due to us not merging anything we don't have to really care, can you look up if we have to have the same vmobject across all regions?

Sidenote: We are returning the wrong code on one or more not mapped pages (EINVAL->ENOMEM)

Userland/Tests/Kernel/mprotect-multi-region-mrotect.cpp
Hendiadyoin1: I've tried it on Linux and the test passes (after some compatability changes and a sudo (for some reason I wasn't allowed to mmap)) so it seems we can hav multiple vm objects or Linux does some optimizations in the background and unifies the vm objects silently...

Kernel/Syscalls/mmap.cpp
ADKaster: All right, I think it's probably up to the OS then. Doing what Linux does until there's a good reason not to seems fine to me.

Userland/Tests/Kernel/mprotect-multi-region-mrotect.cpp
Hendiadyoin1: just rebased and moved it to the corect spot (fixed that return too)

Hendiadyoin1:
If theres nothing else feel free to ping a someone with write access

ADKaster (approved):

Hendiadyoin1:
@awesomekling @linusg This is ready for a final review now as it seems

PR #6066, opened 2021-04-01
LibJS: Implement most of String.prototype.replace

trflynn89:
"Eh the spec for `replace` can't be that bad" - me, living in a simpler past :)

### DIFF
--- a/Userland/Libraries/LibJS/Runtime/RegExpPrototype.cpp
+++ b/Userland/Libraries/LibJS/Runtime/RegExpPrototype.cpp
@@ -50,6 +50,7 @@ void RegExpPrototype::initialize(GlobalObject& global_object)
     define_native_function(vm.names.exec, exec, 1, attr);
 
     define_native_function(vm.well_known_symbol_match(), symbol_match, 1, attr);
+    define_native_function(vm.well_known_symbol_replace(), symbol_replace, 2, attr);
 
     u8 readable_attr = Attribute::Configurable;
     define_native_property(vm.names.flags, flags, {}, readable_attr);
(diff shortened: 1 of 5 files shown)


Lubrsi:
Very nice to have :smile: , whenever I went to random websites this was almost guaranteed to pop up as missing.

Userland/Libraries/LibJS/Runtime/RegExpPrototype.cpp
```diff
@@ -281,4 +282,119 @@ JS_DEFINE_NATIVE_FUNCTION(RegExpPrototype::symbol_match)
...
+        if (result.is_null()) {
+            break;
+        }
+
+        results.append(move(result));
+        if (!global) {
+            break;
+        }
+
+        String match_str = results.last().as_array().get(0).to_string(global_object);
```
linusg, line 324: This blows up when messing with the RegExp object's `exec` method, e.g.:

```js
> r = /foo/
[RegExp] /foo/
> r.exec = () => 42
[ScriptFunction] exec
> r[Symbol.replace]()
js: ../../Userland/Libraries/LibJS/Runtime/Value.cpp:225: JS::Array& JS::Value::as_array(): Assertion `is_array()' failed.
```

spec says `Let matchStr be ? ToString(? Get(result, "0")).`, which works with any object (but depends on `RegExpExec` throwing for type other than object or null).

Userland/Libraries/LibJS/Runtime/RegExpPrototype.cpp
```diff
@@ -281,4 +282,119 @@ JS_DEFINE_NATIVE_FUNCTION(RegExpPrototype::symbol_match)
...
+            //        Once implemented, step (8a) of the @@replace algorithm must also be implemented.
+            rx->regex().start_offset += 1;
+        }
+    }
+
+    String accumulated_result;
+    size_t next_source_position = 0;
+
+    for (auto& result : results) {
+        const auto& result_array = result.as_array();
```
linusg, line 338: Same as above, `length_of_array_like()` works with any object

Userland/Libraries/LibJS/Runtime/RegExpPrototype.cpp
```diff
@@ -281,4 +282,119 @@ JS_DEFINE_NATIVE_FUNCTION(RegExpPrototype::symbol_match)
...
+
+        size_t result_length = length_of_array_like(global_object, result_array);
+        size_t n_captures = result_length == 0 ? 0 : result_length - 1;
+
+        Value matched = result_array.get(0);
```
linusg, line 343: - Missing exception check, possible empty value leak when calling `replace_value.as_function()` below
- you can use `auto` here :)

Userland/Libraries/LibJS/Runtime/RegExpPrototype.cpp
```diff
@@ -281,4 +282,119 @@ JS_DEFINE_NATIVE_FUNCTION(RegExpPrototype::symbol_match)
...
+
+        double position = result_array.get(vm.names.index).to_integer_or_infinity(global_object);
```
linusg, line 345: Missing exception check

Userland/Libraries/LibJS/Runtime/RegExpPrototype.cpp
```diff
@@ -281,4 +282,119 @@ JS_DEFINE_NATIVE_FUNCTION(RegExpPrototype::symbol_match)
...
+        position = clamp(position, static_cast<double>(0), static_cast<double>(string.length()));
+
+        Vector<Value> captures;
+        for (size_t n = 1; n <= n_captures; ++n) {
+            captures.append(result_array.get(n));
```
linusg, line 350: Missing exception check

Userland/Libraries/LibJS/Runtime/RegExpPrototype.cpp
```diff
@@ -281,4 +282,119 @@ JS_DEFINE_NATIVE_FUNCTION(RegExpPrototype::symbol_match)
...
+        }
+
+        auto named_captures = result_array.get(vm.names.groups);
```
linusg, line 353: Missing exception check

linusg (changes requested):

linusg:
Nice to have this! TL;DR: the `result_array` is an arbitrary object we cannot trust, any property may not exist or accessing it throw an exception.

Userland/Libraries/LibJS/Runtime/RegExpPrototype.cpp
trflynn89: Yeah I wasn't sure about the `ToString` here. We have to store the `captures` list as a list of `Value`, so we'd basically be converting from a `Value` to a `String` then immediately back to a `Value`. Seemed wasteful since all we do with this information currently is pass it to the function callback. Would just adding an `is_string` check suffice?

trflynn89:
Thanks for the feedback! I \*think\* I've addressed your comments :)

Userland/Libraries/LibJS/Runtime/RegExpPrototype.cpp
linusg: Calling `to_string()` on a `Value` can also side effects, so we can't just skip it - it is what it is 🤷
We get an `AK::String` and put that into a `JS::PrimitiveString` which is then wrapped in a `JS::Value` (there's no conversion for that) - it's probably not as wasteful as it may seem.

Userland/Libraries/LibJS/Runtime/RegExpPrototype.cpp
```diff
@@ -281,4 +282,138 @@ JS_DEFINE_NATIVE_FUNCTION(RegExpPrototype::symbol_match)
...
+    bool global = global_value.to_boolean();
+    if (global)
+        rx->regex().start_offset = 0;
+
+    // FIXME: Implement and use RegExpExec - https://tc39.es/ecma262/#sec-regexpexec
+    auto* exec = get_method(global_object, rx, vm.names.exec);
+    if (!exec)
+        return {};
+
+    Vector<Value> results;
```
linusg, line 311: Let's make this a `Vector<Object&>` instead - the spec makes that assumption too (`Let result be ? RegExpExec(rx, S).` returns either object or null or throws). That way we also won't need to call `to_object()` on the value multiple times.

Userland/Libraries/LibJS/Runtime/RegExpPrototype.cpp
```diff
@@ -281,4 +282,138 @@ JS_DEFINE_NATIVE_FUNCTION(RegExpPrototype::symbol_match)
...
+
+    while (true) {
+        auto result = vm.call(*exec, rx, string_value);
```
linusg, line 314: one more missing exception check here

Userland/Libraries/LibJS/Runtime/RegExpPrototype.cpp
trflynn89: :+1: got it. Updated with that step added.

Userland/Libraries/LibJS/Runtime/RegExpPrototype.cpp
```diff
@@ -281,4 +282,148 @@ JS_DEFINE_NATIVE_FUNCTION(RegExpPrototype::symbol_match)
...
+        results.append(move(result));
+        if (!global) {
+            break;
+        }
+
+        auto* match_object = results.last().to_object(global_object);
+        if (!match_object)
+            return {};
+
+        String match_str = match_object->get(0).to_string(global_object);
```
linusg, line 328: needs to be split into separate steps unfortunately, when the get() throws it returns an empty value we can't call to_string() on

Userland/Libraries/LibJS/Runtime/RegExpPrototype.cpp
```diff
@@ -281,4 +282,148 @@ JS_DEFINE_NATIVE_FUNCTION(RegExpPrototype::symbol_match)
...
+
+    for (auto& result : results) {
+        auto* result_object = result.to_object(global_object);
+        if (!result_object)
+            return {};
+
+        size_t result_length = length_of_array_like(global_object, *result_object);
+        size_t n_captures = result_length == 0 ? 0 : result_length - 1;
+
+        auto matched = result_object->get("0");
```
linusg, line 349:
```suggestion
        auto matched = result_object->get(0).value_or(js_undefined());
```

(creating a numeric `PropertyName` in the first place is slightly faster as `get()` would try to convert the string to a number otherwise, and if no property `0` exists we get an empty value back which we don't want to give to the "outside" - IIRC the spec defaults to undefined for `Get()`, so I see why that is not obvious)

Userland/Libraries/LibJS/Runtime/RegExpPrototype.cpp
```diff
@@ -281,4 +282,148 @@ JS_DEFINE_NATIVE_FUNCTION(RegExpPrototype::symbol_match)
...
+        if (vm.exception())
+            return {};
+
+        auto position_value = result_object->get(vm.names.index);
+        if (vm.exception() || position_value.is_empty())
+            return {};
```
linusg, line 355: I think this would be

```suggestion
        auto position_value = result_object->get(vm.names.index).value_or(js_undefined());
        if (vm.exception())
            return {};
```

undefined `to_integer_or_infinity()`'d results in `0`, but we can continue with that

Userland/Libraries/LibJS/Runtime/RegExpPrototype.cpp
```diff
@@ -281,4 +282,148 @@ JS_DEFINE_NATIVE_FUNCTION(RegExpPrototype::symbol_match)
...
+
+        double position = position_value.to_integer_or_infinity(global_object);
+        if (vm.exception())
+            return {};
+
+        position = clamp(position, static_cast<double>(0), static_cast<double>(string.length()));
+
+        Vector<Value> captures;
+        for (size_t n = 1; n <= n_captures; ++n) {
+            auto capture = result_object->get(n);
```
linusg, line 365:
```suggestion
            auto capture = result_object->get(n).value_or(js_undefined());
```

Userland/Libraries/LibJS/Runtime/RegExpPrototype.cpp
```diff
@@ -281,4 +282,148 @@ JS_DEFINE_NATIVE_FUNCTION(RegExpPrototype::symbol_match)
...
+
+                capture = Value(js_string(vm, capture_string));
+                if (vm.exception())
+                    return {};
+            }
+
+            captures.append(move(capture));
+        }
+
+        auto named_captures = result_object->get(vm.names.groups);
```
linusg, line 382:
```suggestion
        auto named_captures = result_object->get(vm.names.groups).value_or(js_undefined());
        if (vm.exception())
            return {};
```

linusg (commented):
some more feedback - sorry I missed some things earlier, it's late :sleeping:

Userland/Libraries/LibJS/Runtime/RegExpPrototype.cpp
trflynn89: It looks like `Vector` doesn't like holding references. Just declaring `Vector<Object&>` results in:
```
../.././AK/Vector.h: In instantiation of 'class AK::Vector<JS::Object&>':
../../Userland/Libraries/LibJS/Runtime/RegExpPrototype.cpp:311:21:   required from here
../.././AK/Vector.h:179:8: error: forming pointer to reference type 'JS::Object&'
  179 |     T* data()
```

Using `Vector<Object *>` instead seems to work fine. Is that safe here?

trflynn89:
> some more feedback - sorry I missed some things earlier, it's late sleeping

No worries! Updated - thanks for the detailed review!

Userland/Libraries/LibJS/Runtime/RegExpPrototype.cpp
linusg: Well, I was 100% convinced that's a thing, but you're right - it's not. This is fine :+1:

linusg (approved):

PR #6067, opened 2021-04-01
Ports: Added a Super-Mario clone port.

Baitinq:
Added a minimal dependency super mario port. (Only depends on SDL, SDL_image and SDL_mixer).

I do have some questions about this port tho. Currently I created an $install_location variable in the package.sh, which is where all the game files and executable will be installed, as currently the game seems to only work when the game files are in the same place as the executable.

Is there any better way to do this? Currently it feels kinda hackish in the bad way :^(

### DIFF
--- a/Ports/AvailablePorts.md
+++ b/Ports/AvailablePorts.md
@@ -80,6 +80,7 @@ Please make sure to keep this list up to date when adding and updating ports. :^
 | [`sl`](sl/)                    | Steam Locomotive (SL)                         |                   | https://github.com/mtoyoda/sl                         |
 | [`sqlite`](sqlite/)            | SQLite                                        | 3350300           | https://www.sqlite.org/                               |
 | [`stress-ng`](stress-ng/)      | stress-ng                                     | 0.11.23           | https://github.com/ColinIanKing/stress-ng             |
+| [`Super-Mario`](Super-Mario/)  | Super-Mario Clone                             |                   |https://github.com/Bennyhwanggggg/Super-Mario-Clone-Cpp|
 | [`termcap`](termcap/)          | GNU termcap                                   | 1.3.1             | https://www.gnu.org/software/termutils/               |
 | [`tinycc`](tinycc/)            | Tiny C Compiler (TinyCC)                      | dev               | https://github.com/TinyCC/tinycc                      |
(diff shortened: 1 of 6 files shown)


linusg:
This looks awesome! I tried building it, are you sure the dependency finding is set up correctly? Even with the SDL2, SDL2_image and SDL2_mixer ports installed I would get errors telling me it couldn't find either, and those errors did go away (or rather changed into compiler errors, no i368-specific headers found) once I `dnf install`ed the development packages on my host machine...

> Is there any better way to do this? Currently it feels kinda hackish in the bad way :^(

This seems fine tbh, many ports just `cp` their stuff where ever they want - some even into `$HOME`. My only suggestion would be installing into `/urs/local/bin`, not `/usr/bin`.

Baitinq:
@linusg  thanks! Yeah it seems like the dependency checking scripts were a bit broken, I only had to export the SDL2DIR variable to fix it (is there any better place to do that than in the package.sh?). I dont know whats up with the "no i368-specific headers found" error, as it would build for me if I had SDL2_mixer and SDL2_image installed in my system. Maybe it somehow it magically works now?

linusg:
> I dont know whats up with the "no i368-specific headers found" error

To clarify, I only had the x86_64 SDL2 packages installed on my machine and it seemed to have used those (unsuccessfully) - sorry for the confusion. No longer an issue, works perfectly now!

Baitinq:
Cool!

Baitinq:
> This seems fine tbh, many ports just `cp` their stuff where ever they want - some even into `$HOME`. My only suggestion would be installing into `/urs/local/bin`, not `/usr/bin`.

@linusg Maybe an /opt directory for software like this would be nice, that way it wouldnt have to share a place amongst binary only programs and could have its own directory inside /opt

linusg:
sure, that's probably even better :)

Baitinq:
EDIT: now the port installs under a new /opt directory

Baitinq:
EDIT: now it installs under /opt

thatdutchguy:
Neat! It does take close to 400MB of storage (uncompressed image assets), which is hilarious considering the original game fit in 32KiB :-)

PR #6071, opened 2021-04-02
Toolchain: Build gcc with --enable-threads=posix

### LINKED ISSUE #2485: LibC GTHREADS missing.
From the looks of it gthr-default.h doesn't tie into the PThreads implementation and __GTHREADS is not defined as 1. As a result _GLIBCXX_HAS_GTHREADS is not defined. That causes std::mutex and std::condition_variable to not be available.

RealKC:
This should allow us to port programs that use <mutex> or <conditional_variable> with less patches now.

Closes #2485 I think

@alimpfard do you want to check if this allows us to remove`0022-remove-mutex.patch` and `0024-shared-mutex.patch` from the cmake port?

Also is defining abort in the kernel as I did in https://github.com/RealKC/serenity/commit/32b8b6811baa4527216ad2c9cbcc749930f59be4 the right approach? Not doing so would require us to patch libstdc++ I believe

### DIFF
--- a/Kernel/init.cpp
+++ b/Kernel/init.cpp
@@ -355,3 +355,10 @@ multiboot_info_t* multiboot_info_ptr;
 void* __dso_handle __attribute__((visibility("hidden")));
 
 }
+extern "C" {
+[[noreturn]] void abort();
+[[noreturn]] void abort()
+{
+    VERIFY_NOT_REACHED();
+}
+}
--- a/Toolchain/BuildIt.sh
+++ b/Toolchain/BuildIt.sh
@@ -237,29 +237,33 @@ pushd "$DIR/Build/$ARCH"
                                             --enable-languages=c,c++ \
                                             --enable-default-pie \
                                             --enable-lto \
+                                            --enable-threads=posix \
                                             ${TRY_USE_LOCAL_TOOLCHAIN:+"--quiet"} || exit 1
 
-        echo "XXX build gcc and libgcc"
+        echo "XXX build gcc"
         "$MAKE" -j "$MAKEJOBS" all-gcc || exit 1
         if [ "$(uname -s)" = "OpenBSD" ]; then
             ln -sf liblto_plugin.so.0.0 gcc/liblto_plugin.so
         fi
-        "$MAKE" -j "$MAKEJOBS" all-target-libgcc || exit 1
-        echo "XXX install gcc and libgcc"
-        "$MAKE" install-gcc install-target-libgcc || exit 1
+        echo "XXX install gcc"
+        "$MAKE" install-gcc || exit 1
 
-        echo "XXX serenity libc and libm headers"
+        echo "XXX serenity libc, libm, and pthread headers"
         mkdir -p "$BUILD"
         pushd "$BUILD"
-            mkdir -p Root/usr/include/
             SRC_ROOT=$($REALPATH "$DIR"/..)
-            FILES=$(find "$SRC_ROOT"/Userland/Libraries/LibC "$SRC_ROOT"/Userland/Libraries/LibM -name '*.h' -print)
+            FILES=$(find "$SRC_ROOT"/Userland/Libraries/LibC "$SRC_ROOT"/Userland/Libraries/LibM "$SRC_ROOT"/Userland/Libraries/LibPthread -name '*.h' -print)
             for header in $FILES; do
-                target=$(echo "$header" | sed -e "s@$SRC_ROOT/Userland/Libraries/LibC@@" -e "s@$SRC_ROOT/Userland/Libraries/LibM@@")
+                target=$(echo "$header" | sed -e "s@$SRC_ROOT/Userland/Libraries/LibC@@" -e "s@$SRC_ROOT/Userland/Libraries/LibM@@" -e "s@$SRC_ROOT/Userland/Libraries/LibPthread@@")
                 $INSTALL -D "$header" "Root/usr/include/$target"
             done
             unset SRC_ROOT
         popd
+
+        echo "XXX build libgcc"
+        "$MAKE" -j "$MAKEJOBS" all-target-libgcc || exit 1
+        echo "XXX install libgcc"
+        "$MAKE" install-target-libgcc || exit 1
 
         echo "XXX build libstdc++"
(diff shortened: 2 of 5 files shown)


alimpfard:
I don't think I can check that, not until #6003 is resolved

RealKC:
Oh yeah that seems like it'd cause an issue. I'll check the gcc port to see if anything needs to be done to it related to the thread model

RealKC:
Marking this as a draft because it seems to have broken booting the system

RealKC:
So I looked into the error and I'm not sure how I'd go about fixing it.

With this patch, libgcc now requires `pthread` symbols, so I guess we'd need to link every binary with pthread? I *did* try linking SystemServer with LibPthread and the dynamic linker complains about us not having `pthread_cancel` defined. So we can't get past SystemServer for now

Here's the relevant part of the logs \_without\_ linking SystemServer with LibPthread: https://wastebin.party/ejisoyolir.md

alimpfard:
Seems like glibc has weak pthread API stubs (probably to get around this issue).
We could also do the same, though some patches to Loader.so are required to make it resolve global weak symbols correctly (at least https://github.com/alimpfard/serenity/commit/b68f25d289880150c6689ecff901e221b96a1c3a#diff-2ed11012351e5c369299541e72f65d9bc5c00886006549033ec43c8eb01d7073R417 and the other changes in that file)

RealKC:
> Seems like glibc has weak pthread API stubs (probably to get around this issue).

Makes sense

> We could also do the same, though some patches to Loader.so are required to make it resolve global weak symbols correctly (at least [alimpfard@b68f25d#diff-2ed11012351e5c369299541e72f65d9bc5c00886006549033ec43c8eb01d7073R417](https://github.com/alimpfard/serenity/commit/b68f25d289880150c6689ecff901e221b96a1c3a#diff-2ed11012351e5c369299541e72f65d9bc5c00886006549033ec43c8eb01d7073R417) and the other changes in that file)

Hm, not exactly knowledgeable on linkers, think I might do some reading.

ADKaster:
> Hm, not exactly knowledgeable on linkers, think I might do some reading

Can't go wrong with the classic Linkers and Loaders by John R. Levine. Manuscript is actually viewable as a pdf on his website (or like, physical copies exist)

https://www.iecc.com/linker/

RealKC:
So with the new patches, the system still doesn't boot. Linking `mount` and `SystemServer` with `pthread` gets us further, when the applets and possibly other services crash, at least some are due to missing the pledge for thread. (There's a lot of debug log spam caused by SystemServer trying to restart services that keep crashing I guess?) It seems like `futex` is being called somewhere, but I can't seem to get nice backtraces with that. Though maybe I'm doing something wrong, I'll continue investigating

RealKC:
Here's my patches that get us further
```
╰─>$ git diff -u
diff --git a/Kernel/Process.h b/Kernel/Process.h
index 7848baa31..b3e6171d8 100644
--- a/Kernel/Process.h
+++ b/Kernel/Process.h
@@ -736,6 +736,7 @@ inline ProcessID Thread::pid() const
             dbgln("Has not pledged {}", #promise);                   \
             Process::current()->coredump_metadata().set(             \
                 "pledge_violation", #promise);                       \
+            (void)Process::current()->sys$dump_backtrace();                                                         \
             Process::current()->crash(SIGABRT, 0);                   \
             VERIFY_NOT_REACHED();                                    \
         }                                                            \
diff --git a/Userland/Services/SystemServer/CMakeLists.txt b/Userland/Services/SystemServer/CMakeLists.txt
index fda6eb1bc..c26b74a32 100644
--- a/Userland/Services/SystemServer/CMakeLists.txt
+++ b/Userland/Services/SystemServer/CMakeLists.txt
@@ -4,4 +4,4 @@ set(SOURCES
 )
 
 serenity_bin(SystemServer)
-target_link_libraries(SystemServer LibCore)
+target_link_libraries(SystemServer LibCore LibPthread)
diff --git a/Userland/Services/SystemServer/main.cpp b/Userland/Services/SystemServer/main.cpp
index 17e52b9da..e305a46c2 100644
--- a/Userland/Services/SystemServer/main.cpp
+++ b/Userland/Services/SystemServer/main.cpp
@@ -203,7 +203,7 @@ int main(int, char**)
 {
     prepare_devfs();
 
-    if (pledge("stdio proc exec tty accept unix rpath wpath cpath chown fattr id sigaction", nullptr) < 0) {
+    if (pledge("stdio proc exec tty accept unix rpath wpath cpath chown fattr id sigaction thread", nullptr) < 0) {
         perror("pledge");
         return 1;
     }
diff --git a/Userland/Utilities/CMakeLists.txt b/Userland/Utilities/CMakeLists.txt
index 41b016336..52fe140ad 100644
--- a/Userland/Utilities/CMakeLists.txt
+++ b/Userland/Utilities/CMakeLists.txt
@@ -32,6 +32,7 @@ target_link_libraries(lspci LibPCIDB)
 target_link_libraries(man LibMarkdown)
 target_link_libraries(md LibMarkdown)
 target_link_libraries(misbehaving-application LibCore)
+target_link_libraries(mount LibPthread)
 target_link_libraries(notify LibGUI)
 target_link_libraries(open LibDesktop)
 target_link_libraries(pape LibGUI)
```
and the debug log I get when trying to boot with them: <https://wastebin.party/fenodogago.md>, if anyone would like to lend me a hand 😅

Hendiadyoin1:
The current Errors seem to be
> [#0 ResourceGraph.Applet(9:9)]: Has not pledged thread
> [#0 ClipboardHistory.Applet(17:17)]: Has not pledged thread

and diverse calls of the LibC pthread stubs  
just use [Cntr]+F to find them all:
* DHCPClient(13)
* Clipboard(14)
* Somthing (probably ResourceGraph.Applet)
so they have to be linked with pthread

RealKC:
Yeah I think I'll just progressively link everything that now needs it to LibPthread.

I hope this won't require us pledging `thread` everywhere, as it makes the promise meaningless. I think that's because our mutexes/locks are implemented using `futex` and that's a syscall requiring `thread`?

RealKC:
I'm closing this in favor of #6343 as I think that should have the same effect.

PR #6078, opened 2021-04-02
LibWeb: Add a line box fragment for <br> nodes

trflynn89:
This fixes (some of) the rendering of http://serenityos.org/dromaeo/ :)

The DOM for this page has some `<li>` nodes with a number of `<br>` nodes as their last element. When `BlockFormattingContext` tried to figure out the height of those `<li>` tags, it decided on 0 because the last `<br>` child didn't have any line box fragments.

Before:
before

After:
after


awesomekling:
This feels slightly less elegant than adapting the inline layout algorithm to check for zero-fragment line boxes.
But I think that's okay. And as I mentioned on some other bug, inline layout needs to change substantially to better support floats anyway :)

PR #6080, opened 2021-04-02
Shot + LibGfx: Added the ability for shot to take partial screenshots

Baitinq:
cropped(Gfx::Point top_left, Gfx::Point bottom_right) is a function that can be used on a bitmap to return a crop from another bitmap (aka another bitmap). It requires two points as arguments (the top-left and bottom-right parts of the crop). This could be used to implement a screenshot section-feature for shot :)

### DIFF
--- a/Userland/Libraries/LibGfx/Bitmap.cpp
+++ b/Userland/Libraries/LibGfx/Bitmap.cpp
@@ -409,6 +409,27 @@ RefPtr<Gfx::Bitmap> Bitmap::flipped(Gfx::Orientation orientation) const
     return new_bitmap;
 }
 
+RefPtr<Gfx::Bitmap> Bitmap::cropped(const Gfx::IntRect& rect) const
+{
+    auto crop_width = rect.width();
+    auto crop_height = rect.height();
+    auto start_x = rect.top_left().x();
+    auto start_y = rect.top_left().y();
+
+    auto new_bitmap = Gfx::Bitmap::create(this->format(), { crop_width, crop_height }, scale());
+    if (!new_bitmap)
+        return nullptr;
+
+    for (int i = 0; i < crop_width; i++) {
+        for (int j = 0; j < crop_height; j++) {
+            auto original_pixel = get_pixel(i + start_x, j + start_y);
+            new_bitmap->set_pixel(i, j, original_pixel);
+        }
+    }
+
+    return new_bitmap;
+}
+
 #ifdef __serenity__
 RefPtr<Bitmap> Bitmap::to_bitmap_backed_by_anon_fd() const
 {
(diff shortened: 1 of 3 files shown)


awesomekling:
While this is neat functionality, you're effectively just adding unused code.
Please include the feature as well. :)

Also, the API signature is quite odd. Let's instead do:

```cpp
RefPtr<Gfx::Bitmap> cropped(const Gfx::IntRect& rect) const;
```

Baitinq:
@awesomekling Sure, was planning to do it in a separate pull request but ill add it in with this one :^)

Userland/Libraries/LibGfx/Bitmap.cpp
```diff
@@ -409,6 +409,25 @@ RefPtr<Gfx::Bitmap> Bitmap::flipped(Gfx::Orientation orientation) const
...
+RefPtr<Gfx::Bitmap> Bitmap::cropped(Gfx::Point<int> top_left, Gfx::Point<int> bottom_right) const
+{
+    auto crop_width = top_left.absolute_relative_distance_to(bottom_right).x();
+    auto crop_height = top_left.absolute_relative_distance_to(bottom_right).y();
+
+    auto new_bitmap = Gfx::Bitmap::create(this->format(), { crop_width, crop_height }, scale());
+    if (!new_bitmap)
+        return nullptr;
+
+    for (int i = top_left.x(); i < crop_width; i++) {
```
tomuta, line 421: Why not `blit()`? That way we can use fast paths or fall back to slower paths (if we ever add e.g. cropping to a different format).

Baitinq: Im not at all familiar with the blit function. At the moment i am just providing a basic way of achieving this functionality but over the next few days ill see if I can improve it using blit :). Thanks for the heads up

Baitinq:
@awesomekling I have added the very basic partial-screenshoting functionality to the shot utility. Please let me know if this is fine as I feel like I might have oversimplified it

Userland/Utilities/shot.cpp
```diff
@@ -39,10 +39,18 @@ int main(int argc, char** argv)
...
+    const char* width = nullptr;
+    const char* height = nullptr;
     bool output_to_clipboard = false;
     int delay = 0;
 
     args_parser.add_positional_argument(output_path, "Output filename", "output", Core::ArgsParser::Required::No);
+    args_parser.add_option(start_x, "Pixels from where the screenshot gets taken (x)", "starting-x", 'x', "pixels");
+    args_parser.add_option(start_y, "Pixels from where the screenshot gets taken (y)", "starting-y", 'y', "pixels");
+    args_parser.add_option(height, "Screenshot height in pixels", "height", 'h', "pixels");
+    args_parser.add_option(width, "Screenshot width in pixels", "width", 'w', "pixels");
```
awesomekling, line 53: This feels very clunky. It would be a lot nicer if you would provide the screenshot rect as a string e.g `shot -r 100,100:200x300` or similar.

PR #6083, opened 2021-04-02
LibWeb: Initial support for background-repeat

trflynn89:
This supports a subset of single-value background-repeat properties (`repeat`, `repeat-x`, `repeat-y`, `no-repeat`). Previously, `repeat` was implicitly used.

repeat

This makes the background of http://serenityos.org/dromaeo/ look (more) as intended:
bg


awesomekling:
It would be good to have background painting logic in a single location. But for now this is great :)

PR #6088, opened 2021-04-03
Kernel: Enable PCI ECAM method again if available

supercomputer7:
Apparently we don't enable PCI ECAM (MMIO access to the PCI
configuration space) even if we can. This is a regression, as it was
enabled in the past and in unknown time it was regressed.

The CommandLine::is_mmio_enabled method was renamed to
CommandLine::is_pci_ecam_enabled to better represent the meaning
of this method and what it determines.

Also, an UNMAP_AFTER_INIT macro was removed from a method
in the MMIOAccess class as it halted the system when the kernel
tried to access devices after the boot process.

### DIFF
--- a/Kernel/CMakeLists.txt
+++ b/Kernel/CMakeLists.txt
@@ -115,8 +115,9 @@ set(KERNEL_SOURCES
     PCI/Device.cpp
     PCI/DeviceController.cpp
     PCI/IOAccess.cpp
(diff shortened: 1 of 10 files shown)


supercomputer7:
I converted this back to be a draft because I noticed it takes too much time to invoke `lspci` when booting with the Q35 machine.
We need to map things in a better way to make things more efficient.

PR #6090, opened 2021-04-03
LibJS+LibWeb: Make Dromaeo work, part 1

linusg:
A bunch of LibJS and LibWeb changes towards the goal of making the [Dromaeo JavaScript Performance Test Suite](https://wiki.mozilla.org/Dromaeo) work in the browser.

It now gets as far as loading tests in the background via XHR, but then crashes with:

```
WebContent(32): USERSPACE(32) ASSERTION FAILED: !node->m_parent
../../Userland/Libraries/LibWeb/TreeNode.h:357
```

Which I can circumvent like this (not included here as that's definitely not the right solution):

```diff
--- a/Userland/Libraries/LibWeb/DOM/Node.cpp
+++ b/Userland/Libraries/LibWeb/DOM/Node.cpp
@@ -181,6 +181,11 @@ const Element* Node::parent_element() const
 
 RefPtr<Node> Node::append_child(NonnullRefPtr<Node> node, bool notify)
 {
+    if (node->parent_node()) {
+        dbgln("FIXME: Trying to append_child() a node that already has a parent");
+        // Can't return nullptr here, appendChild() expects a non-null return value. No idea why is is a plain RefPtr.
+        return node;
+    }
     if (&node->document() != &document())
         document().adopt_node(node);
     TreeNode<Node>::append_child(node, notify);
```

After that it just hangs indefinitely (Browser window becomes non-responsive in single process mode / webview stops responding and updating in multi-process mode), which is where I'm stuck and why this is "part 1".


awesomekling:
Typo in commit message (so annoying that GitHub doesn't let you comment on commit messages): Contructor

Lubrsi:
> It now gets as far as loading tests in the background via XHR, but then crashes with:

It's because we don't quite follow the spec for node insertion. Spec: https://dom.spec.whatwg.org/#concept-node-insert
Step 7.1 of node insertion wants us to "Adopt node into parent's node document."
Step 2 of the adoption removes the node's parent if it has one. https://dom.spec.whatwg.org/#concept-node-adopt

PR #6092, opened 2021-04-03
LibGfx: Use integer version of Bresenham's algorithm.

olegv11:
Multiply the original floating-point version by 2dx (or 2dy in another branch) and you get an integer version of the same line-drawing algorithm. My quick benchmark showed that diagonal line drawing is 20% faster with this version. However, it's not exactly equivalent:

<details>
<summary>Floating point version</summary>

float_version
</details>

<details>
<summary>Integer version</summary>

int_version
</details>

<details>
<summary>Imagemagick's "compare"</summary>

diff
</details>

Notice few pixels on the the left Bezier curve. Still, because nothing else changed I think it was just floating point weirdness for the short lines the curve was made of, and it's safe to use the new version.


alimpfard (approved):
LGTM

PR #6093, opened 2021-04-03
Ports: Added a Prince of Persia (SDLPoP) port.

Baitinq:
This is a port of the dissasembly version of the DOS prince of persia
game :)

### DIFF
--- a/Ports/AvailablePorts.md
+++ b/Ports/AvailablePorts.md
@@ -65,6 +65,7 @@ Please make sure to keep this list up to date when adding and updating ports. :^
 | [`patch`](patch/)              | patch (OpenBSD)                               | 6.6               | https://github.com/ibara/libpuffy                     |
 | [`pcre2`](pcre2/)              | Perl-compatible Regular Expressions (PCRE2)   | 10.34             | https://www.pcre.org/                                 |
 | [`pkgconf`](pkgconf/)          | pkgconf                                       | 1.7.3             | https://github.com/pkgconf/pkgconf                    |
+| [`SDLPoP`](SDLPoP/)            | Prince of Persia game                         |                   | https://github.com/NagyD/SDLPoP                       |
 | [`printf`](printf/)            | printf (OpenBSD)                              | 6.6               | https://github.com/ibara/libpuffy                     |
 | [`pt2-clone`](pt2-clone/)      | ProTracker 2 clone                            | 1.28              | https://github.com/8bitbubsy/pt2-clone                |
 | [`python3`](python3/)          | Python                                        | 3.9.2             | https://www.python.org/                               |
--- a/Ports/SDLPoP/package.sh
+++ b/Ports/SDLPoP/package.sh
@@ -1,0 +1,18 @@
+#!/usr/bin/env -S bash ../.port_include.sh
+port=PrinceOfPersia
+useconfigure=true
+version=git
+depends="SDL2 SDL2_image"
+workdir=SDLPoP-master
+configopts="-DCMAKE_TOOLCHAIN_FILE=$SERENITY_ROOT/Toolchain/CMakeToolchain.txt"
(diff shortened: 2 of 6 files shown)


awesomekling:
Could you explain the patches and why they are necessary?

Baitinq:
@awesomekling Sure!

`disable_extra_features.patch ` this just disables some extra features the game has such as screenshots, hardware acceleration, etc

`disable_options_loading.patch ` this disables loading game options from the SDLPoP.ini file. This is because the system they use to load the config uses a conversion specifier for fscanf that we dont support.

`disable_sound_loading.patch ` This disables the sound loading as it has the same fscanf conversion specifier problem.

`fix_cmakelists.patch ` This just links against serenity's base libraries.

Ports/SDLPoP/patches/disable_options_loading.patch
```diff
@@ -0,0 +1,49 @@
...
+--- a/src/options.c
++++ b/src/options.c
+@@ -55,24 +55,24 @@ int ini_load(const char *filename,
+ 		return -1;
+ 	}
+ 
+-	while (!feof(f)) {
+-		if (fscanf(f, "[%127[^];\n]]\n", section) == 1) {
++	//while (!feof(f)) {
++		/*if (fscanf(f, "[%127[^];\n]]\n", section) == 1) {
```
Dexesttp, line 12: I'm guessing that the formatting issue that caused this patch is due to formatting strings of the form `fscanf(f, "%127s")`.

If I'm not mistaken, this is the "optional specifier for maximum field width" (as specified in C99 chapter 7.19.6.2 section 3). It would make sense that this doesn't work, since [`LibC/scanf.cpp`](https://github.com/SerenityOS/serenity/blob/master/Userland/Libraries/LibC/scanf.cpp#L387) doesn't seem to support this feature at all.

Is that right?
In which case, instead of commenting everything out, it would be cleaner to simply remove the numbers :
```suggestion
+		if (fscanf(f, "[%[^];\n]]\n", section) == 1) {
```

Dexesttp: Right, nevermind, this is a bit more complex than handling that, since formatting strings like `fscanf(f, "%[]qwe]")` or `fscanf(f, "%[^]qwe]")` don't seem supported by the code either (it should be matched as a scan list respectively containing and not containing ']', 'q', 'w' and 'e'). In this case, the only clean option seems to comment it out indeed.
I'll check if that's indeed the case once my serenity compiles again, and open a few issues about these if needed.

Baitinq: Ive just tested the code with your suggestion and it compiles and runs fine if you only remove those numbers, will submit the updated patch in a bit. Thanks for the help!

Baitinq:
EDIT: Implemented @Dexesttp suggestion and renamed the patches to make their purpose more clear.

Ports/SDLPoP/patches/disable_options_loading.patch
Baitinq: > Right, nevermind, this is a bit more complex than handling that, since formatting strings like `fscanf(f, "%[]qwe]")` or `fscanf(f, "%[^]qwe]")` don't seem supported by the code either (it should be matched as a scan list respectively containing and not containing ']', 'q', 'w' and 'e'). In this case, the only clean option seems to comment it out indeed.
> I'll check if that's indeed the case once my serenity compiles again, and open a few issues about these if needed.

tbh I have no idea what the `[^]qwe` format specifer would even do, never seen something like that haha. ill try and read up a bit on this type of format specifiers. Atm the code works pretty good so if it doesnt work its not trivial for the games functionality :^)

awesomekling:
This looks good! The long-term goal ofc is to remove the need for the patches, but it's cool for now :)

PR #6094, opened 2021-04-03
Browser: Fix crash of the browser when hitting del key on edittable element al…

mshojatalab:
…l the way up to the EOL and then again, as well as hitting backspace all the way up to the start of the line and then again.

### DIFF
--- a/Userland/Libraries/LibWeb/Page/EditEventHandler.cpp
+++ b/Userland/Libraries/LibWeb/Page/EditEventHandler.cpp
@@ -45,8 +45,8 @@ void EditEventHandler::handle_delete(DOM::Range& range)
     if (start == end) {
         StringBuilder builder;
         builder.append(start->data().substring_view(0, range.start_offset()));
-        builder.append(end->data().substring_view(range.end_offset()));
-
+        auto str = end->data();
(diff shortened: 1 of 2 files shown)


awesomekling (changes requested):
This looks like you are just papering over the `TODO()` statements with returns and not actually implementing the missing functionality.

This is not an appropriate approach to development. Instead, please work out what the correct behavior should be and implement that. :)

mshojatalab:
My thinking was that retuning and not doing anything further is the correct behaviour, no?

awesomekling:
> EDIT:
> My thinking was that retuning and not doing anything further is the correct behaviour, I mean for a non cross boundary delete action, don't you agree?
> We can still add a comment //FIXME for cross boundary delete, but with this change at least we don't crash.

If the forward/backward delete action doesn't cross a node boundary, then yes there is nothing more to do. So if this patch is only about fixing the most trivial cases, we should *check* for the cross-boundary case and `TODO()` in that case.

Hitting a `TODO()` crash is not a bug, it's a strong reminder that we need to implement something. :)

Before continuing, please read the [contribution guidelines](https://github.com/SerenityOS/serenity/blob/83b5655977787ebd3410c6d2562fc18e97ae0a91/CONTRIBUTING.md) and format your commit message properly.

PR #6095, opened 2021-04-03
Spreadsheet: Declare the extern GML variable as an array, not a pointer

d1823:
The old "array is not a pointer" strikes again. :+1:


alimpfard:
Well that's an interesting dumb mistake lol

PR #6098, opened 2021-04-03
 LibC + Ports: Added strtoimax() and strtoumax() to make unnecesary a Prince of Persia patch

Baitinq:
They are the same as strtol() and strtoul() but if there is overflow/underflow then the maximum integer val/lower integer val/maximum uint val will be returned while also setting errno to ERANGE.

This makes the `remove_strtoimax.patch` obsolete.


Baitinq:
This is probably a naive implementation so please let me know if I should change anything :D

Baitinq:
EDIT: Fixed the format in accordance to the clang format.

RealKC:
Some things I noticed:

1. You're checking if the value is in the range of (unsigned) int, but you should check if it's in the range of (u)intmax_t

2. I'd use stro(u)ll instead, to get a bigger range

3. I believe these functions should actually be defined in <inttypes.h>

Baitinq:
> Some things I noticed:
> 
>     1. You're checking if the value is in the range of (unsigned) int, but you should check if it's in the range of (u)intmax_t
> 
>     2. I'd use stro(u)ll instead, to get a bigger range
> 
>     3. I believe these functions should actually be defined in <inttypes.h>

Ive added all your suggestions, thanks :D

awesomekling:
Oh heck yeah :D

PR #6099, opened 2021-04-03
LibWeb: Implement the Screen interface (and everything needed to make it happen)

linusg:
Don't ask me why they want signed integers for screen width and height, but that IDL is straight from the [spec](https://drafts.csswg.org/cssom-view/#screen).


tomuta:
Why signed? So they can represent imaginary screens.

PR #6101, opened 2021-04-03
AK: Update LexicalPath::relative_path to work for '/' prefix

caffineehacker:
If the prefix path ends in a slash the LexicalPath was removing too many characters. Now only remove an extra character if the prefix does not end in a slash.

### DIFF
--- a/AK/LexicalPath.cpp
+++ b/AK/LexicalPath.cpp
@@ -127,7 +127,9 @@ String LexicalPath::relative_path(const String absolute_path, const String& pref
     if (!absolute_path.starts_with(prefix))
         return absolute_path;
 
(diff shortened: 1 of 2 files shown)


awesomekling:
Please add a test to `AK/Tests/TestLexicalPaths.cpp` to cover this :)
Also please format the commit message per the [contribution guidelines](https://github.com/SerenityOS/serenity/blob/ade97d409444cceecfab812cb497a9559b7bd412/CONTRIBUTING.md)

caffineehacker:
Thanks, updated the commit description to reflect the component being changed and added tests (which actually caught an issue in my change!)

PR #6110, opened 2021-04-04
PixelPaint: Avoid notifying image when creating a layer's snapshot

mrkct:
This fixes a bug where the application would crash if the user changed the default values for opacity or visibility of a layer and then tried to draw on it. 

Calling `set_opacity_percent`/`set_visible` also calls `m_image.layer_did_modify_properties({}, *this)` which tries (and fails) to lookup the newly created layer. It worked for default values because those two methods do nothing if the new value is the same as the current one.

### DIFF
--- a/Userland/Applications/PixelPaint/Layer.cpp
+++ b/Userland/Applications/PixelPaint/Layer.cpp
@@ -55,10 +55,17 @@ RefPtr<Layer> Layer::create_with_bitmap(Image& image, const Gfx::Bitmap& bitmap,
 RefPtr<Layer> Layer::create_snapshot(Image& image, const Layer& layer)
 {
     auto snapshot = create_with_bitmap(image, *layer.bitmap().clone(), layer.name());
-    snapshot->set_opacity_percent(layer.opacity_percent());
-    snapshot->set_visible(layer.is_visible());
+    /*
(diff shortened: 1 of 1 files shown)


Userland/Applications/PixelPaint/Layer.cpp
```diff
@@ -55,8 +55,9 @@ RefPtr<Layer> Layer::create_with_bitmap(Image& image, const Gfx::Bitmap& bitmap,
 RefPtr<Layer> Layer::create_snapshot(Image& image, const Layer& layer)
 {
     auto snapshot = create_with_bitmap(image, *layer.bitmap().clone(), layer.name());
-    snapshot->set_opacity_percent(layer.opacity_percent());
-    snapshot->set_visible(layer.is_visible());
+    snapshot->m_opacity_percent = layer.opacity_percent();
+    snapshot->m_visible = layer.is_visible();
+
     snapshot->set_selected(layer.is_selected());
     snapshot->set_location(layer.location());
```
awesomekling, line 62: It's not obvious while reading this code why some things are assigned directly and others go through setters. At the very least let's add a comment explaining that these must be set directly to avoid notifying the `Image`.

mrkct: Done

PR #6112, opened 2021-04-04
Ports: fix libvorbis installtion prefix

ElectrodeYT:
Accidentally left in a --prefix command, causing the port to install to the wrong location.
Sorry!


ElectrodeYT:
It works now:
image

PR #6114, opened 2021-04-04
LibGfx: Speed up  fill_rect_with_gradient

olegv11:
I was dragging windows around with enabled profiler and saw that fill_rect_with_gradient took a lot of samples in WindowFrame::Paint, so I decided to see what's up with it. Turns out it spent most of the time converting between double and single floating point formats. Since all other gamma correction calculations are done in floats there's no reason to use the double-version of square root. This gave a very nice and unexpected 5x boost, at least on a large square bitmap. 

Also add benchmarks for posterity.


awesomekling:
Oh that's a great find!

awesomekling:
With a 5x speedup for 1 character changed, this patch belongs in the patch museum. :)

PR #6076, opened 2021-04-02
Kernel: Add driver for RTL8168 & RTL8111 NICs

IdanHo:
These are pretty common on older LGA1366 & LGA1150 motherboards. I havent had the chance to test the driver with the real NIC, waiting for the PSU for this motherboard to arrive in the mail :)
If anyone has a motherboard with a RTL8168/RTL8111 variant and wants to test it while im waiting, it would be greatly appreciated!

PR #6120, opened 2021-04-04
DHCPClient: Parse MacAddress parts using strtoul

IdanHo:
The current parsing code assumed the ascii lowercase letters came after the ascii numbers, which is not the case, and as such corrupted any mac address that included hex letters (a-f). We likely did not notice this as QEMU's emulated MAC is made up of only hex digits.
@Lubrsi and i noticed this while testing #6076

### DIFF
--- a/Userland/Services/DHCPClient/main.cpp
+++ b/Userland/Services/DHCPClient/main.cpp
@@ -29,6 +29,7 @@
 #include <AK/JsonArray.h>
 #include <AK/JsonObject.h>
 #include <AK/String.h>
+#include <AK/StringUtils.h>
 #include <AK/Types.h>
 #include <LibCore/EventLoop.h>
 #include <LibCore/File.h>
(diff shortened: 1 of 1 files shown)


Userland/Services/DHCPClient/main.cpp
```diff
@@ -40,7 +40,7 @@
 static u8 mac_part(const Vector<String>& parts, size_t index)
 {
     auto chars = parts.at(index).characters();
-    return (chars[0] - '0') * 16 + (chars[1] - '0');
+    return strtoul(chars, nullptr, 16);
```
awesomekling, line 43: Maybe use `StringUtils::convert_to_uint_from_hex()` instead of `strtoul`?

IdanHo: done :)

awesomekling: Commit message still references `strtoul` ;)

IdanHo: whoops! fixed

IdanHo: whoops! fixed

Issue #6116, opened 2021-04-04
Problems with the new PCI ECAM method


supercomputer7:
The log on my ICH7 machine, still not sure why it complains:

<pre><font color="#3B8EEA"><b>[Kernel]</b></font>: CPU[0]: Supported features: nx pae pge sse tsc constant_tsc sep mmx sse2 sse3 ssse3 xsave
<font color="#3B8EEA"><b>[Kernel]</b></font>: CPU[0]: No RDRAND support detected, randomness will be poor
<font color="#3B8EEA"><b>[Kernel]</b></font>: CPU[0]: Physical address bit width: 36
<font color="#3B8EEA"><b>[#0 Kernel]</b></font>: Installing Unhandled Handlers
<font color="#3B8EEA"><b>[Kernel]</b></font>: MM: boot_pdpt @ P0x0010a000
<font color="#3B8EEA"><b>[Kernel]</b></font>: MM: boot_pd0 @ P0x0010b000
<font color="#3B8EEA"><b>[Kernel]</b></font>: MM: boot_pd3 @ P0x0010c000
<font color="#3B8EEA"><b>[Kernel]</b></font>: MM: Low memory range @ P0x00000000 - P0x00100000
<font color="#3B8EEA"><b>[Kernel]</b></font>: MM: Kernel range @ P0x00100000 - P0x00f26000
<font color="#3B8EEA"><b>[Kernel]</b></font>: MM: Multiboot mmap: address=0x00000000, length=653312, type=1
<font color="#3B8EEA"><b>[Kernel]</b></font>: MM: Got an unaligned physical_region from the bootloader; correcting length 653312 by 2048 bytes
<font color="#3B8EEA"><b>[Kernel]</b></font>: MM: Multiboot mmap: address=0x0009f800, length=2048, type=2
<font color="#3B8EEA"><b>[Kernel]</b></font>: MM: Multiboot mmap: address=0x000f0000, length=65536, type=2
<font color="#3B8EEA"><b>[Kernel]</b></font>: MM: Multiboot mmap: address=0x00100000, length=2135818240, type=1
<font color="#3B8EEA"><b>[Kernel]</b></font>: MM: Multiboot mmap: address=0x7f5e0000, length=12288, type=4
<font color="#3B8EEA"><b>[Kernel]</b></font>: MM: Multiboot mmap: address=0x7f5e3000, length=53248, type=3
[...]

PR #3, opened 2019-04-09
Thread Safety (WIP)

ghostslayer989:
Consider using EX:  _**getlogin_r(...) instead of getlogin(...)**_ for improved thread safety.

Much code has been originally written without consideration of multi-threading. Also, engineers were relying on their old experience; They had learned posix before threading extensions were added. This change guides the program to use thread-safe functions.

PR #6121, opened 2021-04-04
Kernel: More bare metal fixes #3

supercomputer7:
Still a draft, but getting closer to fix issue #6116.

### DIFF
--- a/Kernel/Arch/x86/CPU.h
+++ b/Kernel/Arch/x86/CPU.h
@@ -195,6 +195,50 @@ public:
 class GenericInterruptHandler;
 struct RegisterState;
 
+template<typename T>
+void read_possibly_unaligned_data(u8* where, T& data)
+{
+    // FIXME: Implement 64 bit unaligned data access
+    VERIFY(sizeof(T) == 8 || sizeof(T) == 4 || sizeof(T) == 2);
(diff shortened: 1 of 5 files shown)


supercomputer7:
Until I figure out why PCI ECAM doesn't work on my ICH7 machine, we should disable it, in case it can cause boot problems on other machines as well.

Kernel/Arch/x86/CPU.h
```diff
@@ -195,6 +195,50 @@ class PageDirectoryPointerTable {
 class GenericInterruptHandler;
 struct RegisterState;
 
+template<typename T>
+void read_possibly_unaligned_data(u8* where, T& data)
+{
+    // FIXME: Implement 64 bit unaligned data access
+    VERIFY(sizeof(T) == 8 || sizeof(T) == 4 || sizeof(T) == 2);
```
tomuta, line 202: Sorry, a little late to the party, but I think we should consider doing a `static_assert(sizeof(T) != 8)` here until it's implemented.

supercomputer7: But then you won't be able to access aligned 64 bit pointers, so the only thing you've got here is to disable all 64 bit accesses even though it should work just fine for many cases.

PR #6123, opened 2021-04-04
LibGUI+Calendar+Taskbar: Enable adding calendar events

edgararaj:


awesomekling:
Hmm, it feels a bit messy to start pulling calendar event logic into the taskbar program.
Maybe we should bring back the clock applet after all :]

Anyway, that doesn't prevent us from moving forward here.

PR #6129, opened 2021-04-04
Ports+LibC+LibM: Make ScummVM run games

AnicJov:
By adding these libs we make ScummVM capable of running a lot of games :)
There are still a few patches to the actual scummvm port that need to be made, which I will submit in a separate PR (#6134).

Hooray for Myst on Serenity!
image


Ports/faad2/package.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env -S bash ../.port_include.sh
+port=faad2
+version=2_10_0
+useconfigure=true
+files="https://github.com/knik0/faad2/archive/refs/tags/${version}.tar.gz faad2-${version}.tar.gz"
+
+post_fetch() {
+    cd "faad2-${version}"
+    autoreconf -vif
+    cd ..
```
linusg, line 10: That's exactly what `run` does :)

https://github.com/SerenityOS/serenity/blob/15a820e9ad47fa61b6010fac8dbd58c6ea853f81/Ports/.port_include.sh#L41-L44

```suggestion
    run autoreconf -vif
```

Ports/fribidi/package.sh
```diff
@@ -0,0 +1,12 @@
...
+port=fribidi
+version=1.0.10
+useconfigure=true
+files="https://github.com/fribidi/fribidi/releases/download/v${version}/fribidi-${version}.tar.xz fribidi-${version}.tar.xz"
+makeopts="CFLAGS=-std=c99"
+
+pre_configure() {
+    cd "fribidi-${version}"
+    aclocal
+    cd ..
```
linusg, line 11:
```suggestion
    run aclocal
```

Ports/readline/package.sh
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env -S bash ../.port_include.sh
+port=readline
+version=8.1
+useconfigure=true
+files="ftp://ftp.gnu.org/gnu/readline/readline-${version}.tar.gz readline-${version}.tar.gz"
```
linusg, line 5: Sweet, I'll update the Python port to build with readline which should make the REPL experience a lot better :D

Userland/Libraries/LibC/sys/timeb.cpp
```diff
@@ -0,0 +1,44 @@
+/*
+ * Copyright (c) 2021, Andreas Kling <kling@serenityos.org>
```
linusg, line 2: This file was written by you!

Userland/Libraries/LibC/sys/timeb.h
```diff
@@ -0,0 +1,43 @@
+/*
+ * Copyright (c) 2021, Andreas Kling <kling@serenityos.org>
```
linusg, line 2: same here

linusg (changes requested):

Ports/faad2/package.sh
AnicJov: Did not know this! Will fix

awesomekling:
The "Ports: Add flac" commit is full of unrelated changes.

GMTA:
@AnicJov this is a pretty big PR; why do the datafiles from a game released in 1993 require flac and libtheora?

PR #6132, opened 2021-04-05
VimEditingEngine: basic support for multiplied commands

Szune:
Add basic support for multiplied commands - only movement keys to begin with

cepi @ irc here, like I said I'm very new to C++ but the second I noticed there was a vim emulation mode in TextEditor and my usual movements didn't work (5j5j5j5j...), I figured that might be an okay place to start contributing.

I imagine the code isn't up to par yet, but I'd like to get there eventually :-)

Currently there's only support for HJKL/arrow keys in normal mode (figured it was best not to make too much of a mess before looking for feedback)

### DIFF
--- a/Userland/Libraries/LibGUI/VimEditingEngine.cpp
+++ b/Userland/Libraries/LibGUI/VimEditingEngine.cpp
@@ -65,8 +65,7 @@ bool VimEditingEngine::on_key_in_insert_mode(const KeyEvent& event)
 
 char VimEditingEngine::numeric_key_value(const KeyCode key)
 {
-    switch(key)
-    {
+    switch(key) {
     case (KeyCode::Key_0):
         return '0';
     case (KeyCode::Key_1):
@@ -97,7 +96,7 @@ void VimEditingEngine::clear_command_multiplier()
     m_command_multiplier.clear();
 }
 
-void VimEditingEngine::vim_move_down(const KeyEvent &event)
+void VimEditingEngine::vim_move_down(const KeyEvent& event)
 {
     // FIXME: probably not an efficient way to do this
     if (m_command_multiplier.is_empty()) {
@@ -112,7 +111,7 @@ void VimEditingEngine::vim_move_down(const KeyEvent &event)
     }
 }
 
-void VimEditingEngine::vim_move_up(const KeyEvent &event)
+void VimEditingEngine::vim_move_up(const KeyEvent& event)
 {
     if (m_command_multiplier.is_empty()) {
         move_one_up(event);
@@ -126,7 +125,7 @@ void VimEditingEngine::vim_move_up(const KeyEvent &event)
     }
 }
 
-void VimEditingEngine::vim_move_left(const KeyEvent &event)
+void VimEditingEngine::vim_move_left(const KeyEvent& event)
 {
     if (m_command_multiplier.is_empty()) {
         move_one_left(event);
@@ -140,7 +139,7 @@ void VimEditingEngine::vim_move_left(const KeyEvent &event)
     }
 }
 
-void VimEditingEngine::vim_move_right(const KeyEvent &event)
+void VimEditingEngine::vim_move_right(const KeyEvent& event)
 {
     if (m_command_multiplier.is_empty()) {
         move_one_right(event);
--- a/Userland/Libraries/LibGUI/VimEditingEngine.cpp
+++ b/Userland/Libraries/LibGUI/VimEditingEngine.cpp
@@ -86,15 +86,15 @@ void VimEditingEngine::clear_command_multiplier()
 }
 
 template<typename F>
-void VimEditingEngine::multiply_command(F&& move_function)
+void VimEditingEngine::multiply_command(F&& command)
 {
     if (m_command_multiplier > 0) {
         for (uint32_t i = 0; i < m_command_multiplier; ++i) {
-            move_function();
+            command();
         }
         clear_command_multiplier();
     } else {
-        move_function();
+        command();
     }
 }
 
--- a/Userland/Libraries/LibGUI/VimEditingEngine.cpp
+++ b/Userland/Libraries/LibGUI/VimEditingEngine.cpp
@@ -63,110 +63,82 @@ bool VimEditingEngine::on_key_in_insert_mode(const KeyEvent& event)
     return false;
 }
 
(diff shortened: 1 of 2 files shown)


Szune:
Realized right before going to bed that I made a mistake and broke a lot of commands by adding the new code before any modifier key checks, will fix when I wake up!

Userland/Libraries/LibGUI/VimEditingEngine.cpp
```diff
@@ -63,8 +63,111 @@ bool VimEditingEngine::on_key_in_insert_mode(const KeyEvent& event)
...
+        move_one_up(event);
+    } else {
+        String multiplier_string = m_command_multiplier.build();
+        uint32_t multiplier = multiplier_string.to_uint().value();
+        for (uint32_t i = 0; i < multiplier; ++i) {
+            move_one_up(event);
+        }
+        clear_command_multiplier();
+    }
+}
```
Petelliott, line 126: this logic shouldn't be repeated this many times. I would make a helper function that takes a lambda expression and evaluates it `m_command_multiplier` times, then resets `m_command_multiplier`. then I would wrap the actions in `on_key_in_normal_mode`

Userland/Libraries/LibGUI/VimEditingEngine.cpp
```diff
@@ -63,8 +63,111 @@ bool VimEditingEngine::on_key_in_insert_mode(const KeyEvent& event)
...
+
 bool VimEditingEngine::on_key_in_normal_mode(const KeyEvent& event)
 {
+    char numeric_value = numeric_key_value(event.key());
+    if (numeric_value != '\0') {
+        add_to_command_multiplier(numeric_value);
+        return true;
+    }
```
ADKaster, line 170: `on_key_in_normal_mode` is already really chonky, but it feels like the specific implementation you're going for, moving N times in a direction, only applies 1) if the previous key was not d (in that case we should be looking for more input for "what" to `delete`) and  2) if there are no modifiers pressed.

so, I'd recommend moving this check to the !ctrl, !alt and !shift check that starts on line 337 below. Otherwise if you do, `5dd5` or something you'll store `55` in the m_command_multiplier instead of `5`.

ADKaster: actually maybe i'm wrong about the recommendation. But the m_previous_key logic might screw you up if you don't clear the command multiplier in more situations.

Userland/Libraries/LibGUI/VimEditingEngine.cpp
```diff
@@ -63,8 +63,111 @@ bool VimEditingEngine::on_key_in_insert_mode(const KeyEvent& event)
...
+        for (uint32_t i = 0; i < multiplier; ++i) {
+            move_one_right(event);
+        }
+        clear_command_multiplier();
+    }
+}
+
+void VimEditingEngine::add_to_command_multiplier(char value)
+{
+    // FIXME: there are more efficient ways to do this
```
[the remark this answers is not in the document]
Szune: That's what it was, thanks a ton! I tried another convoluted way before I ended up with the string "solution", probably should have gone to bed :)

Userland/Libraries/LibGUI/VimEditingEngine.h
```diff
@@ -51,6 +51,15 @@ class VimEditingEngine final : public EditingEngine {
 
     VimMode m_vim_mode { VimMode::Normal };
 
+    void vim_move_down(const KeyEvent&);
+    void vim_move_left(const KeyEvent&);
+    void vim_move_right(const KeyEvent&);
+    void vim_move_up(const KeyEvent&);
```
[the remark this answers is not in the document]
Szune: Good note as well, I did that because move_down() etc would shadow/hide the move_down() that is inherited from EditingEngine, is there a precedent on doing this? :)

Userland/Libraries/LibGUI/VimEditingEngine.cpp
Szune: Super! I'll have a look at how lambdas are implemented in C++ and fix that :)

Userland/Libraries/LibGUI/VimEditingEngine.cpp
Szune: Good note! You're absolutely right that I got those things wrong, I'll start by taking your recommendation and moving it to the no-modifier part of the code and then think of some scenarios when the multiplier needs to be cleared and when it is okay to add to it (and capture the key) :)

Userland/Libraries/LibGUI/VimEditingEngine.cpp
```diff
@@ -63,8 +63,112 @@ bool VimEditingEngine::on_key_in_insert_mode(const KeyEvent& event)
     return false;
 }
 
+char VimEditingEngine::numeric_key_value(const KeyCode key)
```
awesomekling: Instead of this, you can get the numeric character from the `KeyEvent::code_point()`

awesomekling (changes requested):
Please squash fixups instead of stacking commits. We don't need to preserve your exact editing 
 history :)

Userland/Libraries/LibGUI/VimEditingEngine.cpp
Szune: Ah, that's an older commit, sorry!

I changed it to use a new method called "is_multiplied_key" in VimEditingEngine.cpp to apply it on the correct commands.
And "key_code_to_numeric_key_value" (returns Optional<T>) in VimEditingEngine.h (not sure if this one already exists, don't think I found it in the KeyEvent source?) to get the numeric value of the key code (if any)

If I'm understanding you correctly about the squashing, the first 3 commits should have been squashed into a first commit and the last 2 should have been squashed into a second commit? I'll do better next time, thanks for the advice :)

sin-ack:
Hi @Szune, I worked on my own implementation of motions without realizing you were also working on it. Would you like to help out with #6437 or should I close mine so you can work on yours instead?

Szune:
Hi @sin-ack!
From a cursory glance yours seems to be more comprehensive and I'd be perfectly fine closing mine instead :)
However, I'm not certain I will be available to work on yours because of time constraints, but it seems like you've already got most things done anyhow (I like your separation into units btw!)

Szune:
Closing in favor of #6437

PR #6144, opened 2021-04-05
LibC: implement sigsuspend in signal.cpp

mrkmg:
Implemented to support port of zsh

### DIFF
--- a/Userland/Libraries/LibC/signal.cpp
+++ b/Userland/Libraries/LibC/signal.cpp
@@ -182,10 +182,21 @@ void siglongjmp(jmp_buf env, int val)
     longjmp(env, val);
 }
 
-int sigsuspend(const sigset_t*)
+int sigsuspend(const sigset_t* set)
 {
-    dbgln("FIXME: Implement sigsuspend()");
-    TODO();
+    assert(set);
+    sigset_t o_sigset;
+
+    if (sigprocmask(SIG_SETMASK, set, &o_sigset) < 0)
+        return -1;
+
+    auto block = sigprocmask(SIG_BLOCK, nullptr, nullptr);
+
+    if (sigprocmask(SIG_SETMASK, &o_sigset, nullptr) < 0)
+        return -1;
+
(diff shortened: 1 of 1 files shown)


awesomekling:
Hmm? `sigsuspend()` is supposed to suspend the calling thread until a signal is received.
There's nothing in this implementation that will suspend the thread.

We probably need a kernel-side implementation of [pause()](https://pubs.opengroup.org/onlinepubs/9699919799/functions/pause.html) for this.

mrkmg:
@awesomekling

I read through how glibc implements it, and they do use pause(), but that (from what I can tell), just calls sigprocmask(SIG_BLOCK) to block the process until a signal is received. I likely am not interpreting it correctly though.

References: 
https://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/posix/sigsuspend.c
https://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/posix/pause.c

bgianfo:
This PR should probably add a new test to validate the implementation works as expected... 😁

Userland/Libraries/LibC/signal.cpp
```diff
@@ -182,10 +182,21 @@ void siglongjmp(jmp_buf env, int val)
...
+    assert(set);
+    sigset_t o_sigset;
+
+    if (sigprocmask(SIG_SETMASK, set, &o_sigset) < 0)
+        return -1;
+
+    auto block = sigprocmask(SIG_BLOCK, nullptr, nullptr);
+
+    if (sigprocmask(SIG_SETMASK, &o_sigset, nullptr) < 0)
+        return -1;
```
awesomekling, line 196: None of these system calls are blocking, so this implementation will not block.

I think this could be implementing by blocking the relevant signals, and then doing `select()` with empty fd sets to effectively block indefinitely until an unblocked signal is delivered.

awesomekling (changes requested):

PR #6151, opened 2021-04-05
DHCPClient: Retry DISCOVER for interfaces that fail

### LINKED ISSUE #6125: DHCPClient: Only attempt to perform DHCP on connected interfaces
Found when testing #6076, where a real NIC takes a while to report that the link is up (where userland has already finished loading). We attempt DHCP anyway and the packets are lost, requiring a manual restart of DHCPClient.

### LINKED ISSUE #6126: DHCPClient: Perform DHCP on interfaces that have just become connected
An extension of #6125. When an interface comes online, DHCPClient is none the wiser and the interface goes without an IP. It currently requires you to restart DHCPClient to get an IP.

alimpfard:
Also avoid trying to acquire an IP on interfaces that aren't up.
Fixes #6126.
Fixes #6125.

I'm not too happy about the polling style, but it's not frequent enough to be an issue...probably.
cc @Lubrsi

### DIFF
--- a/Userland/Services/DHCPClient/DHCPv4Client.cpp
+++ b/Userland/Services/DHCPClient/DHCPv4Client.cpp
@@ -29,22 +29,43 @@
 #include <AK/Debug.h>
 #include <AK/Endian.h>
 #include <AK/Function.h>
+#include <AK/JsonArray.h>
+#include <AK/JsonObject.h>
+#include <AK/JsonParser.h>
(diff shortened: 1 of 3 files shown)


Lubrsi:
Yep, it works. Though a minute does feel a bit long

alimpfard:
Hmm...we could try an exponential backoff, say, start at 1s and grow to 10m or something

Issue #6075, opened 2021-04-02
LibWeb: ExceptionOr doesn't work with values that JS::Value doesn't accept


Lubrsi:
I'm currently rewriting the node mutation algorithms to be spec compliant and this is a common return type: `ExceptionOr<NonnullRefPtr<Node>>`

However, this doesn't quite work:
```
In file included from Userland/Libraries/LibWeb/Bindings/NodePrototype.cpp:19:
../../Userland/Libraries/LibWeb/Bindings/ExceptionOrUtils.h: In instantiation of 'Ret Web::Bindings::throw_dom_exception_if_needed(auto:19&&, auto:20&&, F&&) [with F = Web::Bindings::NodePrototype::append_child(JS::VM&, JS::GlobalObject&)::<lambda()>; T = Web::DOM::ExceptionOr<AK::NonnullRefPtr<Web::DOM::Node> >; Ret = JS::Value; auto:19 = JS::VM&; auto:20 = JS::GlobalObject&]':
Userland/Libraries/LibWeb/Bindings/NodePrototype.cpp:350:108:   required from here
../../Userland/Libraries/LibWeb/Bindings/ExceptionOrUtils.h:62:33: error: could not convert '(& result)->Web::DOM::ExceptionOr<AK::NonnullRefPtr<Web::DOM::Node> >::value()' from 'AK::NonnullRefPtr<Web::DOM::Node>' to 'JS::Value'
   62 |             return result.value();
      |                                 ^
      |                                 |
      |                                 AK::NonnullRefPtr<Web::DOM::Node>
../../Userland/Libraries/LibWeb/Bindings/ExceptionOrUtils.h: In instantiation of 'Ret Web::Bindings::throw_dom_exception_if_needed(auto:19&&, auto:20&&, F&&) [with F = Web::Bindings::NodePrototype::insert_before(JS::VM&, JS::GlobalObject&)::<lambda()>; T = Web::DOM::ExceptionOr<AK::NonnullRefPtr<Web::DOM::Node> >; Ret = JS::Value; auto:19 = JS::VM&; auto:20 = JS::GlobalObject&]':
Userland/Libraries/LibWeb/Bindings/NodePrototype.cpp:395:113:   required from here
[...]

PR #6158, opened 2021-04-06
LibWeb: Make the node mutation algorithms more spec compliant 

Lubrsi:
From the "LibWeb: Make the node mutation algorithms more spec compliant" commit. More details in the individual commits.

The mutation algorithms now more closely follow the spec and
fixes some assertion failures in tests such as Acid3 and Dromaeo.

The main thing that is missing right now is passing exceptions to the
bindings layer. This is because of issue #6075. I spent a while trying
to work it out and got so frustrated I just left it as a FIXME. Besides
that, the algorithms bail at the appropriate points.

This also makes the adopting steps in the document more spec compliant
as it's needed by the insertion algorithm. While I was at it, I added
the adoptNode IDL binding.

This adds a bunch of ancestor/descendant checks to TreeNode as well.
I moved the "remove_all_children" function to Node as it needs to use
the full remove algorithm instead of simply removing it from
the child list.


Lubrsi:
Why do I only notice issues once I submit the PR? Anyway, that should be the last push before review.

awesomekling:
Beautiful work @Lubrsi :)

PR #6160, opened 2021-04-06
LibJS: Support mapFn argument of Array.from

tuqqu:


Userland/Libraries/LibJS/Runtime/ArrayConstructor.cpp
```diff
@@ -54,7 +54,7 @@ void ArrayConstructor::initialize(GlobalObject& global_object)
     define_property(vm.names.length, Value(1), Attribute::Configurable);
 
     u8 attr = Attribute::Writable | Attribute::Configurable;
-    define_native_function(vm.names.from, from, 1, attr);
+    define_native_function(vm.names.from, from, 2, attr);
```
linusg, line 57: This is wrong, the length of `Array.from` is still 1.

Userland/Libraries/LibJS/Runtime/ArrayConstructor.cpp
```diff
@@ -95,13 +95,29 @@ JS_DEFINE_NATIVE_FUNCTION(ArrayConstructor::from)
 
     auto* array = Array::create(global_object);
 
+    Function* map_fn = nullptr;
+    if (vm.argument_count() >= 2) {
```
linusg, line 99: This should not be based on the number of arguments:

```
If mapfn is undefined, let mapping be false.
3. Else,
    a. If IsCallable(mapfn) is false, throw a TypeError exception.
    b. Let mapping be true.
```

Userland/Libraries/LibJS/Runtime/ArrayConstructor.cpp
```diff
@@ -95,13 +95,29 @@ JS_DEFINE_NATIVE_FUNCTION(ArrayConstructor::from)
...
+
     // Array.from() lets you create Arrays from:
     if (auto size = object->indexed_properties().array_like_size()) {
         // * array-like objects (objects with a length property and indexed elements)
         MarkedValueList elements(vm.heap());
         elements.ensure_capacity(size);
         for (size_t i = 0; i < size; ++i) {
-            elements.append(object->get(i));
+            if (map_fn) {
+                auto map_fn_result = vm.call(*map_fn, value, object->get(i));
```
linusg, line 115: Missing exception check, when `object->get()` throws, the subsequent `call()` will blow up. Needs to be split into separate steps.

Userland/Libraries/LibJS/Runtime/ArrayConstructor.cpp
```diff
@@ -111,14 +127,20 @@ JS_DEFINE_NATIVE_FUNCTION(ArrayConstructor::from)
         get_iterator_values(global_object, value, [&](Value element) {
             if (vm.exception())
                 return IterationDecision::Break;
-            array->indexed_properties().append(element);
+
+            if (map_fn) {
+                auto map_fn_result = vm.call(*map_fn, value, element);
```
linusg, line 132: Missing exception check

linusg (changes requested):

tuqqu:
@linusg thanks! fixed

Userland/Libraries/LibJS/Runtime/ArrayConstructor.cpp
```diff
@@ -95,30 +95,61 @@ JS_DEFINE_NATIVE_FUNCTION(ArrayConstructor::from)
 
     auto* array = Array::create(global_object);
 
+    Function* map_fn = nullptr;
+    if (vm.argument_count() >= 2 && !vm.argument(1).is_undefined()) {
```
linusg, line 99: Just this is fine :)

```suggestion
    if (!vm.argument(1).is_undefined()) {
```

linusg (approved):
LGTM!

Userland/Libraries/LibJS/Runtime/ArrayConstructor.cpp
tuqqu: oh, that makes sense. Fixed

PR #6161, opened 2021-04-06
FontEditor: Improved interface and edit commands

thankyouverycool:
font-editor


awesomekling:
Would be cool to copy between instances too, not just within one app :)

PR #6162, opened 2021-04-06
Icons: Another batch

thankyouverycool:


awesomekling:
Oh my goodness, that chess pawn!

\*imagines an entire chess piece set in that style\* :smile:

PR #6163, opened 2021-04-06
LibJS: Implement Object.freeze(), Object.seal(), Object.isFrozen() and Object.isSealed()

linusg:
(not related to Dromaeo :smile:)

### DIFF
--- a/Userland/Libraries/LibJS/Runtime/Object.cpp
+++ b/Userland/Libraries/LibJS/Runtime/Object.cpp
@@ -160,6 +160,54 @@ bool Object::prevent_extensions()
     return true;
 }
 
+// 7.3.15 SetIntegrityLevel, https://tc39.es/ecma262/#sec-setintegritylevel
+bool Object::set_integrity_level(IntegrityLevel level)
+{
+    // FIXME: This feels clunky and should get nicer abstractions.
+    auto update_property = [this](auto& key, auto attributes) {
+        auto property_name = PropertyName::from_value(global_object(), key);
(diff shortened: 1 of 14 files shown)


Userland/Libraries/LibJS/Runtime/Object.cpp
```diff
@@ -187,64 +187,61 @@ Value Object::get_own_property(const PropertyName& property_name, Value receiver
     return value_here;
 }
 
-Value Object::get_own_properties(PropertyKind kind, bool only_enumerable_properties, GetOwnPropertyReturnType return_type) const
+Vector<Value> Object::get_own_properties(PropertyKind kind, bool only_enumerable_properties, GetOwnPropertyReturnType return_type) const
 {
-    auto* properties_array = Array::create(global_object());
+    Vector<Value> properties;
```
awesomekling, line 266: These values are not protected from GC which may happen later in this function. This is why we have `MarkedValueList` :)

linusg: Hmm, but nothing was protecting the array (and therefore all the elements added to it) from being GC'd either, right?

awesomekling: A pointer to the array was on the stack, and our conservative GC scans everything on the stack that looks like a pointer. :) `Vector` storage is on the malloc heap which is not scanned by GC. See implementation in `Heap::gather_conservative_roots`.

linusg: > scans everything on the stack

that's the part I wasn't aware of :)
Updated!

PR #6166, opened 2021-04-06
move directory down for build

sklei4:
Needs to go down one more directory from ../Build/i686 to reach CMakeLists.txt.


linusg:
This is wrong, the same change was attempted in #5903 - see explanation there.

sklei4:
I see, so the issue likely was that I ran `cd Build/i686` instead of `cd ../Build/i686`, which is a pretty common mistake to make, I presume, seeing that someone already has done this. Thanks for the point. I'll close PR and think of a better resolution.

PR #6167, opened 2021-04-06
Kernel: Replace process' regions vector with a Red Black tree

IdanHo:
This should provide some speed up, as currently searches for regions containing a given address were performed in O(n) complexity, while this container allows us to do those in O(logn).

I haven't done any rigorous benchmarking yet (mainly because im not sure what the best way to do that would be, help with this part would be appreciated), so its possible this doesnt produce a tangible speed-up. (since most proccess' region counts are in the low 10s, and due to the large constant factor of red black tree algorithm, its almost equivalent to vectors in low Ns)

### DIFF
--- a/Kernel/VM/MemoryManager.cpp
+++ b/Kernel/VM/MemoryManager.cpp
@@ -433,13 +433,8 @@ Region* MemoryManager::kernel_region_from_vaddr(VirtualAddress vaddr)
 
 Region* MemoryManager::user_region_from_vaddr(Space& space, VirtualAddress vaddr)
 {
-    // FIXME: Use a binary search tree (maybe red/black?) or some other more appropriate data structure!
     ScopedSpinLock lock(space.get_lock());
-    for (auto& region : space.regions()) {
-        if (region.contains(vaddr))
-            return &region;
-    }
-    return nullptr;
+    return space.find_region_containing({ vaddr, 1 });
 }
 
 Region* MemoryManager::find_region_from_vaddr(Space& space, VirtualAddress vaddr)
--- a/AK/IntrusiveRedBlackTree.h
+++ b/AK/IntrusiveRedBlackTree.h
@@ -1,0 +1,195 @@
+/*
+ * Copyright (c) 2021, Idan Horowitz <idan.horowitz@gmail.com>
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#pragma once
+
+#include <AK/RedBlackTree.h>
+
+namespace AK {
+
+template<Integral K>
+class IntrusiveRedBlackTreeNode;
+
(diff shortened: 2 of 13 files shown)


bgianfo:
Can we made this an Intrusive RedBlack Tree? The fact that we are blindly new'ing and not worrying if the allocation can fail in kernel mode is just making the lack of OOM handling in the kernel even worse than it is today. It would be great if the replacement for the current Vector storage also addressed this shortcoming.

Kernel/VM/MemoryManager.cpp
```diff
@@ -433,13 +433,8 @@ Region* MemoryManager::kernel_region_from_vaddr(VirtualAddress vaddr)
...
 Region* MemoryManager::user_region_from_vaddr(Space& space, VirtualAddress vaddr)
 {
-    // FIXME: Use a binary search tree (maybe red/black?) or some other more appropriate data structure!
     ScopedSpinLock lock(space.get_lock());
-    for (auto& region : space.regions()) {
-        if (region.contains(vaddr))
-            return &region;
-    }
-    return nullptr;
+    return space.find_region_containing({ vaddr, 1 });
```
bgianfo, line 437: Can we do this for kernel_regions as well above? See: `MemoryManager::kernel_region_from_vaddr(VirtualAddress vaddr)`

tomuta:
> Can we made this an Intrusive RedBlack Tree? The fact that we are blindly new'ing nodes on insert and not worrying if the allocation can fail in kernel mode is just making the lack of OOM handling in the kernel even worse than it is today. The current Vector storage also has this shortcoming, it would be great if we could fix this as well with this change.

How about both? Then you can pick whatever is best for the use case.

bgianfo:
> How about both? Then you can pick whatever is best for the use case.

Yea, that's a great idea!

Although you do have the same problem for user space since serenity doesn't compile with exceptions... but most people I encounter claim handling OOM in user space is useless (I am not in this camp) but no point in having that argument here and now. 😄

Kernel/VM/MemoryManager.cpp
IdanHo: Kernel regions are stored as a simple linked list, so i cant run this method on them, unless youre suggesting to replace the kernel regions container itself?

IdanHo:
> Can we made this an Intrusive RedBlack Tree? The fact that we are blindly new'ing nodes on insert and not worrying if the allocation can fail in kernel mode is just making the lack of OOM handling in the kernel even worse than it is today. The current Vector storage also has this shortcoming, it would be great if we could fix this as well with this change.

I'm not very familiar with the idiomatic way to implement Intrusive data structures, would simply exposing the (currently private) Node class and the accompanying (currently private) insert & remove methods be enough? or is this supposed to be a subclassing thing (a la weakable\<T>)

IdanHo:
> I haven't done any rigorous benchmarking yet (mainly because im not sure what the best way to do that would be, help with this part would be appreciated), so its possible this doesnt produce a tangible speed-up. (since most proccess' region counts are in the low 10s, and due to the large constant factor of red black tree algorithm, its almost equivalent to vectors in low Ns)

Another note about this, its possible to reduce the constant factor of the current red black tree implementation (cutting down branches by half by using pointer arrays, using a sentinel node instead of null checking, caching children colors inside the lsb bit of the pointers, etc etc) but i think these more complex optimizations that we can do in future PRs where we iterate on this implementation

bgianfo:
> I'm not very familiar with the idiomatic way to implement Intrusive data structures, would simply exposing the (currently private) Node class and the accompanying (currently private) insert & remove methods be enough? or is this supposed to be a subclassing thing (a la weakable<T>)

You should be able to follow the existing pattern used by [AK/IntrusiveList.h](https://github.com/SerenityOS/serenity/blob/master/AK/IntrusiveList.h)

Kernel/VM/MemoryManager.cpp
bgianfo: Yea I was suggesting we unify the implementation so they both use the same data structure. For example the page fault handler will benefit from the faster lookup.

IdanHo: Hmm, yeah, i guess so, but it looks like memory manager sometimes just wants to iterate over all the kernel and user regions regardless of correct order, so that would be slowed down if we replace it with a red black tree, so we need to store both a red black tree and the current linked list (and use them for those 2 seperate cases), if thats fine

awesomekling: Is there something particularly performance-sensitive that would benefit from unordered iteration? Otherwise I feel like we could go with simplicity first and then add on something like a linked list of regions as an optimization if this becomes an issue in the future.

IdanHo: theres a FIXME in VMObject::for_each_region which says we unfortunately iterate through *all* regions (all user regions of all processes, and all kernel regions) every time an inode changes (but im not sure if thats still correct)

awesomekling: Oh right, that one! I think we can improve that by finding a less generic way to locate all the `Region` objects that reference a given `InodeVMObject` :)

Perhaps `Region` could participate in an intrusive list of `InodeVMObject` clients, for example.

IdanHo: that might be a bit out of scope for this PR (and for my current level of understanding of serenity's VM implementation :P), so would you prefer that i still switch over to an RB tree for kernel regions and just leave the current FIXME, or keep it a linked list for now and add a FIXME so the future PR that optimizes that unnecessary iteration can switch to an RB tree as well?

IdanHo:
I've implemented an Intrusive version of the red black tree now (after trudging through template inheritence hell in order to not completely duplicate the code), but i now realised that in this case the RedBlackTree is also responsible for the ownership of the regions, so replacing it with an intrusive one would result in all of them being dropped by their smart pointers. so im not sure what exactly the best way forward is?

IdanHo:
> I've implemented an Intrusive version of the red black tree now (after trudging through template inheritence hell in order to not completely duplicate the code), but i now realised that in this case the RedBlackTree is also responsible for the ownership of the regions, so replacing it with an intrusive one would result in all of them being dropped by their smart pointers. so im not sure what exactly the best way forward is?

The first way that comes to mind for me is just leaking them when theyre stored in the tree and then re-adopting them when removing them in Space's remove region method, but that sounds weird to me

IdanHo:
Well im not entirely sure what to do about the ownership problem. Since we were already using Vectors before this PR (which has the same OOM problem), i think it should be fine to use the non intrusive version for now, so i just pushed the commit that adds the IntrusiveRedBlackTree, and if I (or someone else) can think of a good way to use it here, it can be done in a future PR

awesomekling:
Ok let's do it.

PR #6170, opened 2021-04-07
LibGUI: Update the autocomplete suggestions after processing keys

alimpfard:
Previously, this was updating the suggestions before processing the keys
and whatever changes they caused, making the suggestions stale until the
periodic auto-autocomplete timer fired (if enabled).

> Okay, autocomplete is basically unusable

Now it still is unusable, but at least it's not incorrect!


awesomekling:
Nice :smile:

PR #6175, opened 2021-04-07
LibJS: Added construction of TypedArray from another

Amjad50:
A fix for `TODO` for `TypedArray` constructor from another `TypedArray`.

This implementation might not be the best, but not sure how else we can get the type of the array when we have `TypedArrayBase`.

If the implementation is good, I'll start adding js tests.

### DIFF
--- a/Userland/Libraries/LibJS/Runtime/TypedArray.cpp
+++ b/Userland/Libraries/LibJS/Runtime/TypedArray.cpp
@@ -93,6 +93,48 @@ static void initialize_typed_array_from_array_buffer(GlobalObject& global_object
     typed_array.set_byte_offset(offset);
     typed_array.set_array_length(new_byte_length.value() / element_size);
 }
+template<typename T>
+static void initialize_typed_array_from_typed_array(GlobalObject& global_object, TypedArray<T>& dest_array, TypedArrayBase& src_array)
+{
+    auto& vm = global_object.vm();
+    if (vm.exception())
+        return;
+
+    // FIXME: 4. If IsDetachedBuffer(src_data) is true, throw a TypeError exception.
+    VERIFY(src_array.viewed_array_buffer());
+
+    auto src_array_length = src_array.array_length();
+    auto dest_element_size = dest_array.element_size();
+    Checked byte_length = src_array_length * dest_element_size;
+    if (byte_length.has_overflow()) {
+        vm.throw_exception<RangeError>(global_object, ErrorType::InvalidLength, "typed array");
+        return;
+    }
+
+    // FIXME: 17.b If IsDetachedBuffer(array_buffer) is true, throw a TypeError exception.
+    // FIXME: 17.c If src_array.[[ContentType]] != dest_array.[[ContentType]], throw a TypeError exception.
+    auto array_buffer = ArrayBuffer::create(global_object, byte_length.value());
+    dest_array.set_array_length(src_array_length);
+    dest_array.set_viewed_array_buffer(array_buffer);
+    dest_array.set_byte_offset(0);
+    dest_array.set_byte_length(array_buffer->byte_length());
+
+    for (u32 i = 0; i < src_array_length; i++) {
+        Value v;
+#undef __JS_ENUMERATE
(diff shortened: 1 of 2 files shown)


linusg (commented):
No need for the new enum, you can get the type at runtime via `is<T>()` - e.g. like this, in the js REPL:

https://github.com/SerenityOS/serenity/blob/96121ddb119e14d36398a4c9589bfecbfd4d7738/Userland/Utilities/js.cpp#L339-L354

linusg:
More importantly though, this assumes that the source and target array have the same type, which we cannot do. See how the [spec](https://tc39.es/ecma262/#sec-initializetypedarrayfromtypedarray) handles this.

Amjad50:
I used `get_by_index` and `put_by_index` for this case,
as when two different types are used. it should cast from one type to the other

Amjad50:
but looks like the specs has more stuff and error handling, I'll look at it, thanks

linusg:
> I used get_by_index and put_by_index for this case,
> as when two different types are used. it should cast from one type to the other

Sorry, I misread the cast - all good :+1:

Userland/Libraries/LibJS/Runtime/TypedArray.cpp
```diff
@@ -93,6 +93,32 @@ static void initialize_typed_array_from_array_buffer(GlobalObject& global_object
...
     typed_array.set_array_length(new_byte_length.value() / element_size);
 }
+template<typename T, TypedArrayType type>
+static void initialize_typed_array_from_typed_array(GlobalObject& global_object, TypedArray<T, type>& typed_array_to, TypedArrayBase& typed_array_from)
+{
+    auto array_length = typed_array_from.array_length();
+    auto element_size = typed_array_to.element_size();
+
+    typed_array_to.set_array_length(array_length);
+    typed_array_to.set_viewed_array_buffer(ArrayBuffer::create(global_object, array_length * element_size));
```
awesomekling, line 103: Looks like `array_length * element_size` can overflow here, no?

Userland/Libraries/LibJS/Runtime/TypedArray.cpp
```diff
@@ -93,6 +93,32 @@ static void initialize_typed_array_from_array_buffer(GlobalObject& global_object
...
+    typed_array_to.set_byte_length(typed_array_to.viewed_array_buffer()->byte_length());
+
+    for (u32 i = 0; i < array_length; i++) {
+        Value v;
+        switch (typed_array_from.array_type()) {
+#undef __JS_ENUMERATE
+#define __JS_ENUMERATE(ClassName, snake_name, PrototypeName, ConstructorName, ArrayType)             \
+    case TYPED_ARRAY_ENUM_TYPE(ArrayType):                                                           \
+        v = dynamic_cast<TypedArray<ArrayType, TYPED_ARRAY_ENUM_TYPE(ArrayType)>&>(typed_array_from) \
+                .get_by_index(i);                                                                    \
```
awesomekling, line 113: `dynamic_cast` can be quite slow, so it's sad to do it for every step of the iteration.
Can we hoist the casting out of the loop?

awesomekling (changes requested):

Userland/Libraries/LibJS/Runtime/TypedArray.cpp
Amjad50: yes, working on it now

Userland/Libraries/LibJS/Runtime/TypedArray.cpp
Amjad50: Agree, but not sure if there is a way to do it, since we can't have an outside variable of type `TypedArray<T>` or is there?

Amjad50: now replaced it with static cast, so I think should be better

linusg (changes requested):
Please add some tests.

Amjad50:
Any other tests should we add?

linusg:
Looks ok, but please use `pascalCase` for JavaScript code

PR #6179, opened 2021-04-07
UE: Some bug fixes

Hendiadyoin1:
Trying to fix everything I found in #5929

Problems:
1. ~the "fix" to exclude malloc and free in the malloc tracker does not seems to work~ It throws complaints in malloc and free, but malloc and free are specifically exempt form these "errors"
2. ~Hashmap does seem to use "uninitialized data" even though everything should be memeset to 0~

regarding 2:  
the error apears in test-js->Heap.cpp and seems to points at
https://github.com/SerenityOS/serenity/blob/2f9321a0d43f3543a50aabb75bb5b6f601ac7b94/AK/HashTable.h#L340
but again every thing should be set to 0 or is a reference to a value, which should be set

### DIFF
--- a/Kernel/StdLib.cpp
+++ b/Kernel/StdLib.cpp
@@ -282,9 +282,7 @@ void* memset(void* dest_ptr, int c, size_t n)
     // FIXME: Support starting at an unaligned address.
     if (!(dest & 0x3) && n >= 12) {
         size_t size_ts = n / sizeof(size_t);
-        size_t expanded_c = (u8)c;
-        expanded_c |= expanded_c << 8;
-        expanded_c |= expanded_c << 16;
+        size_t expanded_c = explode_byte((u8)c);
         asm volatile(
             "rep stosl\n"
             : "=D"(dest)
--- a/Userland/DevTools/UserspaceEmulator/Emulator_syscalls.cpp
+++ b/Userland/DevTools/UserspaceEmulator/Emulator_syscalls.cpp
@@ -880,13 +880,12 @@ u32 Emulator::virt$mmap(u32 params_addr)
         mmu().add_region(MmapRegion::create_anonymous(final_address, final_size, params.prot, move(name_str)));
     } else {
         auto region = MmapRegion::create_file_backed(final_address, final_size, params.prot, params.flags, params.fd, params.offset, move(name_str));
-        if (region->name() == "libc.so: .text") {
+        if (region->name() == "libc.so: .text" && !m_libc_start) {
             m_libc_start = final_address;
             m_libc_end = final_address + final_size;
             bool rc = find_malloc_symbols(*region);
             VERIFY(rc);
-        }
-        if (region->name() == "libsystem.so: .text") {
+        } else if (region->name() == "libsystem.so: .text" && !m_libsystem_start) {
             m_libsystem_start = final_address;
             m_libsystem_end = final_address + final_size;
         }
--- a/Userland/DevTools/UserspaceEmulator/SoftCPU.cpp
+++ b/Userland/DevTools/UserspaceEmulator/SoftCPU.cpp
@@ -27,6 +27,7 @@
 #include "SoftCPU.h"
 #include "Emulator.h"
 #include <AK/Assertions.h>
+#include <AK/BitCast.h>
 #include <AK/Debug.h>
 #include <math.h>
 #include <stdio.h>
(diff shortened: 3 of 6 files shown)


Hendiadyoin1:
OK, I think the Error in Heap.cpp/HashTable comes from
https://github.com/SerenityOS/serenity/blob/2f9321a0d43f3543a50aabb75bb5b6f601ac7b94/Userland/Libraries/LibJS/Heap/Heap.cpp#L134-L142
maybe we don't really initialize the jump buffer

awesomekling:
> OK, I think the Error in Heap.cpp/HashTable comes from
> https://github.com/SerenityOS/serenity/blob/2f9321a0d43f3543a50aabb75bb5b6f601ac7b94/Userland/Libraries/LibJS/Heap/Heap.cpp#L134-L142

The JS garbage collector does legitimate access of uninitialized memory while performing the conservative stack scan. We take every value on the stack and see if it looks like it might be a pointer to a GC-allocated object. If so, we protect it from collection. This is how the C++ stack keeps JS objects alive.

Hendiadyoin1:
So this is just a bad choice to test?

Hendiadyoin1:
then I'll try to test ue with ue, id and maybe sh again, any other good tests, that don't do to weird stuff?

Hendiadyoin1:
And any idea, why the test circumvention in malloc etc does not seem to work?

Hendiadyoin1:
well looking closer it already breaks out when in malloc or free so the patch is useless...

Hendiadyoin1:
Ok I have an Idea whats going on with the malloc and free

It seems that ue loads libc two times, one for its own and one for the vm, but when the child ue loads libc for its vm the outer one may forget where ue's libc was and now only knows the new one, so we may just need to prevent overriding these addresses, when we have them already

AK/HashTable.h
```diff
@@ -31,6 +31,12 @@
 #include <AK/Types.h>
 #include <AK/kmalloc.h>
 
+#ifdef KERNEL
+#    include <Kernel/StdLib.h>
+#else
+#    include <string.h>
+#endif
```
awesomekling, line 38: The point of using `__builtin_memset` in AK headers is precisely to avoid the `#ifdef KERNEL` stuff. The compiled output is exactly the same, `__builtin_memset` becomes a call to `memset` anyway.

Kernel/StdLib.cpp
```diff
@@ -282,9 +282,7 @@ void* memset(void* dest_ptr, int c, size_t n)
     // FIXME: Support starting at an unaligned address.
     if (!(dest & 0x3) && n >= 12) {
         size_t size_ts = n / sizeof(size_t);
-        size_t expanded_c = (u8)c;
-        expanded_c |= expanded_c << 8;
-        expanded_c |= expanded_c << 16;
+        size_t expanded_c = explode_byte((u8)c);
```
awesomekling, line 285: Whoops! What a silly mistake :)

awesomekling (changes requested):
Some typo in commit message:
`dbglnif -> dbgln_if`
`lic -> libc`

AK/HashTable.h
Hendiadyoin1: shall I drop these commits then?

Kernel/StdLib.cpp
Hendiadyoin1: no body uses that upper byte anyway

AK/HashTable.h
awesomekling: Yes please :)

Userland/DevTools/UserspaceEmulator/Emulator.cpp
```diff
@@ -532,6 +532,14 @@ bool Emulator::find_malloc_symbols(const MmapRegion& libc_text)
     m_calloc_symbol_end = m_calloc_symbol_start + calloc_symbol.value().size();
     m_malloc_size_symbol_start = malloc_size_symbol.value().value() + libc_text.base();
     m_malloc_size_symbol_end = m_malloc_size_symbol_start + malloc_size_symbol.value().size();
+
+    reportln("=={}==  Mapped Malloc and Free", getpid());
+    reportln("    malloc: {:p} - {:p}", m_malloc_symbol_start, m_malloc_symbol_end);
+    reportln("    free: {:p} - {:p}", m_free_symbol_start, m_free_symbol_end);
+    reportln("    realloc: {:p} - {:p}", m_realloc_symbol_start, m_realloc_symbol_end);
+    reportln("    calloc: {:p} - {:p}", m_calloc_symbol_start, m_calloc_symbol_end);
+    reportln("    malloc_size: {:p} - {:p}", m_malloc_size_symbol_start, m_malloc_size_symbol_end);
```
awesomekling, line 541: This seems like `dbgln()` material, not `reportln()` material.

Hendiadyoin1: Should I flag it with something too?

Hendiadyoin1: Do we even need it at this point, after the error this was realted to was resoved?

awesomekling: Oh good point. Yeah then we can just drop it :)

Hendiadyoin1: Ok done that then

Hendiadyoin1:
Anything else?

awesomekling:
> Anything else?

No, I think we're good here :) Thanks for working on this!

PR #6187, opened 2021-04-07
Userland: Implement tac command

guerinoni:
Nobody care about tac :^).

### DIFF
--- a/Userland/Utilities/tac.cpp
+++ b/Userland/Utilities/tac.cpp
@@ -1,0 +1,69 @@
+/*
+ * Copyright (c) 2021, Federico Guerinoni <guerinoni.federico@gmail.com>
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include <AK/Vector.h>
+#include <LibCore/ArgsParser.h>
+#include <LibCore/File.h>
+#include <unistd.h>
+
+void print_lines(const Vector<String>& vec);
+
+int main(int argc, char** argv)
+{
+    if (pledge("stdio rpath", nullptr) < 0) {
+        perror("pledge");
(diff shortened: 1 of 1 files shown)


Userland/Utilities/tac.cpp
```diff
@@ -0,0 +1,74 @@
...
+{
+    if (pledge("stdio rpath", nullptr) < 0) {
+        perror("pledge");
+        return 1;
+    }
+
+    Vector<const char*> paths;
+
+    Core::ArgsParser args_parser;
+    args_parser.set_general_help("Concatenate files or pipes to stdout.");
```
awesomekling, line 49: Seems like this should mention the "in reverse" aspect of the program! :)

Userland/Utilities/tac.cpp
```diff
@@ -0,0 +1,74 @@
...
+    args_parser.add_positional_argument(paths, "File path", "path", Core::ArgsParser::Required::No);
+    args_parser.parse(argc, argv);
+
+    if (!paths.is_empty()) {
```
awesomekling, line 53: If `paths` is empty we should do `stdin` instead of nothing.

Userland/Utilities/tac.cpp
```diff
@@ -0,0 +1,74 @@
...
+            auto file = Core::File::construct(path);
+            if (!file->open(Core::IODevice::ReadOnly)) {
+                warnln("Failed to open {}: {}", path, file->error_string());
+                return false;
+            }
+
+            Vector<String> lines;
+            while (file->can_read_line()) {
+                auto line = file->read_line();
+                lines.insert(0, line);
```
awesomekling, line 64: This becomes very slow as the `lines` vector grows. Every time you `insert` at `0`, the vector storage must be shifted one step.

Instead, we can simply `append` to the vector when reading the file, and then iterate the lines backwards when printing them.

awesomekling (changes requested):

Userland/Utilities/tac.cpp
guerinoni: Right!!

Userland/Utilities/tac.cpp
guerinoni: Fixed, :D

Userland/Utilities/tac.cpp
guerinoni: Yes, maybe next time I will add `rbegin` for Vector :)

Userland/Utilities/tac.cpp
```diff
@@ -0,0 +1,83 @@
...
+        }
+    } else {
+        char buf[32768];
+        ssize_t nread = read(0, buf, sizeof(buf));
+        if (nread == 0)
+            return 1;
+        if (nread < 0) {
+            perror("read");
+            return 2;
+        }
```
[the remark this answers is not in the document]
guerinoni: Ehehe I don't know, the implementation on Linux does this :| ... What do you expect after?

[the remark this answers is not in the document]
awesomekling: Well, I would expect it to print the lines it read from `stdin` in reverse order. ;)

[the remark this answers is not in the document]
guerinoni: I think now it's ok

awesomekling (changes requested):
This obviously will not work. Please try your program :)

```
tac /etc/motd

cat /etc/motd | tac
```

Both of these should give the same result.

Userland/Utilities/tac.cpp
```diff
@@ -0,0 +1,95 @@
...
+        for (auto&& s : vec) {
+            lines.append(move(s));
+        }
+    }
+
+    if (!lines.is_empty()) {
+        print_lines(lines);
+    }
+
+    if (!paths.is_empty()) {
```
IdanHo, line 69: instead of a separate if, you can just use the else branch of the first if

Userland/Utilities/tac.cpp
```diff
@@ -0,0 +1,95 @@
...
+            return 2;
+        }
+
+        auto vec = String(buf, Chomp).split('\n', true);
+        for (auto&& s : vec) {
+            lines.append(move(s));
+        }
+    }
+
+    if (!lines.is_empty()) {
```
IdanHo, line 65: print_lines() is a noop with an empty lines, so theres no need to check it length

Userland/Utilities/tac.cpp
```diff
@@ -0,0 +1,95 @@
...
+            }
+
+            print_lines(lines);
+        }
+    }
+
+    return 0;
+}
+
+void print_lines(const Vector<String>& vec)
```
IdanHo, line 90: instead of extracting this for loop into a seperate function, you can just unconditionally do it at the end of the main function, and remove both the calls

Userland/Utilities/tac.cpp
```diff
@@ -0,0 +1,82 @@
...
+            return 1;
+        if (nread < 0) {
+            perror("read");
+            return 2;
+        }
+
+        auto vec = String(buf, Chomp).split('\n', true);
+        for (auto&& s : vec) {
+            lines.append(move(s));
+        }
```
awesomekling, line 62: This code still does not make sense. What if there's more than 32768 bytes coming on  `stdin`? We will not process any more than that.

Instead of having two separate code paths for `paths.is_empty()` being true/false, I suggest doing everything on one code path, and simply using `Core::File::standard_input()` in the stdin case, which gives you a `Core::File` wrapper around stdin.

awesomekling (changes requested):

awesomekling:
👍🏻

PR #6189, opened 2021-04-08
copy/clip: Rename the 'copy' utility to 'clip', and allow output from the

drunderscore:
clipboard to stdout.

Keeping the name as "copy" didn't make much sense to me, especially since outputting the clipboard is usually called "paste"ing...


awesomekling:
This functionality is already implemented by `paste`, which is a complement to the `copy` program.

drunderscore:
@awesomekling oh my... I wish you were joking!

PR #6191, opened 2021-04-08
Shell: Allow newlines between the function decl and its body

alimpfard:
All other control structures are fine with this, so let's keep the
behaviour consistent.


bcoles:
works for me

PR #6195, opened 2021-04-08
AK/Format: Strip trailing zeroes from floating point value output

GMTA:
Before:
image

After:
image

This PR also resolves a name conflict for the test binary `TestFormat` by merging its two sources.


awesomekling:
Hey that's quite nice :)

PR #6205, opened 2021-04-09
LibTLS: Support empty SNI in ServerHello messages

### LINKED ISSUE #6201: ProtocolServer: Encountered unknown extension 0 with length 0
HTML loads fine when doing `pro https://jcs.org`, however when loading the website through a browser it spams `dmesg` with messages since the initial HTML triggers more stuff to be downloaded.

## dmesg
```
pro(26): started download with id 1
ProtocolServer(27): Encountered unknown extension 0 with length 0
ProtocolServer(27): not enough data for certificate size
ProtocolServer(27): Alert: 0
ProtocolServer(27): perror(): IODevice::write: write: Try again
ProtocolServer(27): perror(): IODevice::write: write: Try again
```

GMTA:
According to [RFC6066](https://tools.ietf.org/html/rfc6066), empty extension_data for an SNI extension is absolutely one of the possibilities - so let's support this instead of spamming the debug log.

Fixes #6201.

### DIFF
--- a/Userland/Libraries/LibTLS/ClientHandshake.cpp
+++ b/Userland/Libraries/LibTLS/ClientHandshake.cpp
@@ -117,27 +117,24 @@ ssize_t TLSv12::handle_hello(ReadonlyBytes buffer, WritePacketStage& write_packe
     // The handshake hash function is _always_ SHA256
     m_context.handshake_hash.initialize(Crypto::Hash::HashKind::SHA256);
 
-    if (buffer.size() - res < 1) {
-        dbgln("not enough data for compression spec");
+    // Compression method
+    if (buffer.size() - res < 1)
         return (i8)Error::NeedMoreData;
-    }
     u8 compression = buffer[res++];
-    if (compression != 0) {
-        dbgln("Server told us to compress, we will not!");
+    if (compression != 0)
         return (i8)Error::CompressionNotSupported;
+
+    if (m_context.connection_status != ConnectionStatus::Renegotiating)
+        m_context.connection_status = ConnectionStatus::Negotiating;
+    if (m_context.is_server) {
+        dbgln("unsupported: server mode");
+        write_packets = WritePacketStage::ServerHandshake;
     }
 
-    if (res > 0) {
-        if (m_context.connection_status != ConnectionStatus::Renegotiating)
-            m_context.connection_status = ConnectionStatus::Negotiating;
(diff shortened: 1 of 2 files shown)


Userland/Libraries/LibTLS/ClientHandshake.cpp
```diff
@@ -146,60 +143,70 @@ ssize_t TLSv12::handle_hello(ReadonlyBytes buffer, WritePacketStage& write_packe
...
-                                dbgln("negotiated alpn: {}", alpn_str);
-                                break;
-                            }
-                            alpn_position += alpn_length;
-                            if (!m_context.is_server) // server hello must contain one ALPN
-                                break;
+                // Exactly one ServerName should be present
+                if (buffer.size() - res < 3)
+                    return (i8)Error::NeedMoreData;
+                auto sni_name_type = (NameType) * (const u8*)buffer.offset_pointer(res++);
```
alimpfard, line 163: This deref looks really weird...as if you're trying to multiply a pointer with a value called `NameType`...
is this clang-format's doing?

awesomekling: Yeah I've seen `clang-format` do that sort of thing somewhere else too :(

alimpfard: well...I guess one extra set of parenthesis is a small price to pay for more readable code (I had to think for a while whether something fishy was going on here, just because of _one_ space...)

@GMTA how does this look:
```diff
+ auto sni_name_type = (NameType) (*(const u8*)buffer.offset_pointer(res++));
- auto sni_name_type = (NameType) * (const u8*)buffer.offset_pointer(res++);
```

GMTA: Thanks @alimpfard, that looks more comprehensible - applied! I was checking to see if there were more appropriate C++-style casts I could apply but couldn't figure them out.

GMTA: Although `clang-format` really did not like that space between  `(NameType)` and `(*`...

alimpfard: That's a bummer, but not a big problem 🤷‍♂️

alimpfard (approved):

PR #6207, opened 2021-04-09
UserspaceEmulator: Improve detection of memory leaks

TobyAsE:
I was watching one of your older videos about the UserspaceEmulator and thought why not help?

I replaced the logic determining whether a mallocation is reachable with an implementation that puts all of the mallocations into a graph and find out that way. Doing it like this allows us to also catch leaks where two (or more) mallocations point to each other but nothing from outside has any pointer into any of these segments. Previously these mallocations would've been considered reachable.

This is my first contribution to this project and apart from some Hacktoberfest projects also the first contribution to a real open source project :) Also the first time touching C++, so hopefully I didn't make a complete fool out of me.

### DIFF
--- a/Userland/DevTools/UserspaceEmulator/MallocTracer.cpp
+++ b/Userland/DevTools/UserspaceEmulator/MallocTracer.cpp
@@ -1,5 +1,6 @@
 /*
  * Copyright (c) 2020, Andreas Kling <kling@serenityos.org>
+ * Copyright (c) 2021, Tobias Christiansen <tobi@tobyase.de>
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
@@ -306,36 +307,40 @@ void MallocTracer::audit_write(const Region& region, FlatPtr address, size_t siz
     }
 }
 
-bool MallocTracer::is_reachable(const Mallocation& mallocation) const
+void MallocTracer::populate_memory_graph()
 {
-    VERIFY(!mallocation.freed);
-
-    bool reachable = false;
-
-    // 1. Search in active (non-freed) mallocations for pointers to this mallocation
-    for_each_mallocation([&](auto& other_mallocation) {
-        if (&mallocation == &other_mallocation)
+    // Create Node for each live Mallocation
+    for_each_mallocation([&](auto& mallocation) {
+        if (mallocation.freed)
             return IterationDecision::Continue;
-        if (other_mallocation.freed)
+        m_memory_graph.set(mallocation.address, {});
+        return IterationDecision::Continue;
+    });
+
+    // Find pointers from each memory region to another
+    for_each_mallocation([&](auto& mallocation) {
+        if (mallocation.freed)
             return IterationDecision::Continue;
-        size_t pointers_in_mallocation = other_mallocation.size / sizeof(u32);
+
+        size_t pointers_in_mallocation = mallocation.size / sizeof(u32);
+
+        auto& edges_from_mallocation = m_memory_graph.find(mallocation.address)->value;
+
         for (size_t i = 0; i < pointers_in_mallocation; ++i) {
-            auto value = m_emulator.mmu().read32({ 0x23, other_mallocation.address + i * sizeof(u32) });
-            if (value.value() == mallocation.address && !value.is_uninitialized()) {
+            auto value = m_emulator.mmu().read32({ 0x23, mallocation.address + i * sizeof(u32) });
+            auto other_address = value.value();
+            if (!value.is_uninitialized() && m_memory_graph.contains(value.value())) {
 #if REACHABLE_DEBUG
-                reportln("mallocation {:p} is reachable from other mallocation {:p}", mallocation.address, other_mallocation.address);
+                reportln("region/mallocation {:p} is reachable from other mallocation {:p}", other_address, mallocation.address);
 #endif
-                reachable = true;
-                return IterationDecision::Break;
+                edges_from_mallocation.edges_from_node.append(other_address);
             }
         }
         return IterationDecision::Continue;
     });
 
-    if (reachable)
-        return true;
-
-    // 2. Search in other memory regions for pointers to this mallocation
+    // Find mallocations that are pointed to by other regions
+    Vector<FlatPtr> reachable_mallocations = {};
     m_emulator.mmu().for_each_region([&](auto& region) {
         // Skip the stack
         if (region.is_stack())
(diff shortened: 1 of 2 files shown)


Userland/DevTools/UserspaceEmulator/MallocTracer.h
```diff
@@ -39,6 +39,23 @@ namespace UserspaceEmulator {
...
+};
+
+struct MemoryGraph {
+    HashMap<FlatPtr, GraphNode> map;
+
+    GraphNode get(FlatPtr address)
+    {
+        auto value_or_error = map.get(address);
+        VERIFY(value_or_error.has_value());
+        return value_or_error.value();
```
ADKaster, line 55: Calling `Optional<T>::value()` does a VERIFY that there is a value, so I'm not sure that this helper gets you much at the moment. Seems you could replace this struct with a type alias until more graph-like operations on the Memory graph need specific member functions.

```
using MemoryGraph = HashMap<FlatPtr, GraphNode>:
```

ADKaster (commented):
This looks like a pretty sweet data structure upgrade! Few more copies than are probably necessary though.

Userland/DevTools/UserspaceEmulator/MallocTracer.cpp
```diff
@@ -306,36 +307,43 @@ void MallocTracer::audit_write(const Region& region, FlatPtr address, size_t siz
...
+
+    // Find pointers from each memory region to another
+    for_each_mallocation([&](auto& mallocation) {
+        if (mallocation.freed)
             return IterationDecision::Continue;
-        size_t pointers_in_mallocation = other_mallocation.size / sizeof(u32);
+
+        size_t pointers_in_mallocation = mallocation.size / sizeof(u32);
+
+        auto edges_from_mallocation = graph->get(mallocation.address);
```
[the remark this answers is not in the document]
TobyAsE: I thought that my way of doing it was a bit clunky...
Your suggestion is of course way more elegant, thank you!

Userland/DevTools/UserspaceEmulator/MallocTracer.h
TobyAsE: Ah, that makes a lot of sense!

Userland/DevTools/UserspaceEmulator/MallocTracer.cpp
```diff
@@ -306,36 +307,43 @@ void MallocTracer::audit_write(const Region& region, FlatPtr address, size_t siz
     }
 }
 
-bool MallocTracer::is_reachable(const Mallocation& mallocation) const
+MemoryGraph* MallocTracer::build_memory_graph()
 {
-    VERIFY(!mallocation.freed);
+    auto graph = new MemoryGraph;
```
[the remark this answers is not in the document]
TobyAsE: I have to learn more about this... ;)
I'll just follow your recommendation for now, thanks!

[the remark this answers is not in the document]
IdanHo: Small nitpick: if you do choose to go the OwnPtr route, NonnullOwnPtr will be a better match (since your generation method cant fail)

Userland/DevTools/UserspaceEmulator/MallocTracer.cpp
```diff
@@ -349,19 +354,51 @@ bool MallocTracer::is_reachable(const Mallocation& mallocation) const
...
+    for (size_t i = 0; i < reachable_mallocations.size(); ++i) {
+        auto reachable = reachable_mallocations.at(i);
+        if (visited.contains_slow(reachable))
+            continue;
+        visited.append(reachable);
+        auto& mallocation_node = m_memory_graph.find(reachable)->value;
+
+        if (!mallocation_node.is_reachable) {
+            mallocation_node.is_reachable = true;
+            m_memory_graph.set(reachable, mallocation_node);
```
ADKaster, line 384: This set here is also unnecessary since your mallocation node is a reference to a GraphNode in the memory graph.

Userland/DevTools/UserspaceEmulator/MallocTracer.cpp
```diff
@@ -349,19 +354,51 @@ bool MallocTracer::is_reachable(const Mallocation& mallocation) const
...
+        }
+
+        for (auto& edge : mallocation_node.edges_from_node) {
+            reachable_mallocations.append(edge);
+        }
+    }
+}
+
+void MallocTracer::dump_memory_graph(MemoryGraph graph)
```
ADKaster, line 393: You're passing the graph here by value, which (in this case) results in a copy of the entire memory graph.

Few options to avoid that:

1) Pass by const reference (generally a good idea)
2) don't pass anything and use m_memory_graph directly (good idea in this situation)
3) re-invent your MemoryGraph struct and add a `dump()` method to it. Or `show()`, but I think the pattern in serenity is more towards `dump()`. (maybe excessive, there's not any other specific operations on that graph that are particularly interesting at the moment)

ADKaster (commented):
Looks like much better C++ now! :) Two small things left that I would change

Userland/DevTools/UserspaceEmulator/MallocTracer.cpp
ADKaster: And actually, removing the set makes it more clear that this is almost the exact same thing you're doing on line 365.

A future change set might make `MemoryGraph::set_reachable(FlatPtr, bool)` or similar a thing that should exist.

Could also iterate on this all day and do optimizations such as:
- Make `reachable_mallocations` a vector of `GraphNode*`, since we know all the mallocations will be in the hashmap and have stable addresses. This would let you avoid the hash table lookup when walking your ever-growing reachable_mallocations vector. Would require storing the address in the graph node too, though.
- Make the malloc tracer use the a graph/hashmap structure for all mallocation tracking, not just at report generation time, which would let you avoid the first pass through for_all_mallocations (dubious claim from me, because you are filtering out freed mallocations as well)
- Give reachable_mallocations a quick `reserve` to the number of  nodes in the memory graph, since we know it will need to be at least that big by the end. Same with visited. Maybe number of nodes * some arbitrary scale factor? (1.2? 1.4?)

Buuuut with the two changes I mentioned already it's good enough to merge probably, with the optimizations above to be investigated later :)

Userland/DevTools/UserspaceEmulator/MallocTracer.cpp
TobyAsE: It would be nice to have a generic memory dumping utility, if there arises the need for other memory graphs. For nopw though I just use the m_memory_graph.

ADKaster (approved):

TobyAsE:
Thanks for your help & comments, @ADKaster !

awesomekling:
Very awesome stuff, thank you so much for improving this :)

PR #6210, opened 2021-04-10
Ports: Download GNU port archives and signatures using HTTPS

bcoles:


asynts:
You don't have do download the files themselves via HTTPS, downloading the signatures via HTTPS should be sufficient.

bcoles:
> 
> 
> You don't have do download the files themselves via HTTPS, downloading the signatures via HTTPS should be sufficient.

Prior to this PR we were doing the opposite - only downloading the archives using HTTPS and not the signatures.

No harm in using HTTPS for both.

asynts:
My understanding is that downloading the actual file over HTTPS is more work for the server, because it can't cache it properly. (The encryption has to be done for each individual client.) Not sure how accurate this is today though, I think AES got pretty fast.

I don't think this is a major thing though, just leave it as is.

bcoles:
> My understanding is that downloading the actual file over HTTPS is more work for the server

It is, but the extra server-side processing overhead is insignificant to the point of irrelevance.

> because it can't cache it properly. (The encryption has to be done for each individual client.)

Unless GNU have explicitly stated that HTTP is preferred over HTTPS due to caching I consider this to be a non-issue.

For per-client caching, this is irrelevant, as we download the archive file only once and keep a local copy of the file. The file won't be downloaded again unless the user deletes the archive file (`./package.sh clean_all`). This is possible during development, but unlikely during deployment.

For network caching, GNU do not use intermediary third-party distribution and caching services such as Cloudflare, instead preferring to use mirrors.

We're downloading from the primary GNU server as we don't have any sort of functionality for supporting mirrors yet. The [mirror documentation](https://www.gnu.org/server/mirror.html) expresses no preference for HTTP vs HTTPS, beyond:

    It is very useful to make your mirror available to users via HTTP or HTTPS, so please support this if you can.
    We highly encourage new mirrors to support HTTPS as the most secure option.

Obviously GNU want mirrors to support HTTPS so signatures can be downloaded securely, but this says nothing of the archive files.

> I don't think this is a major thing though

It is not a major issue.

> just leave it as is.

No.

There are risks in using HTTP rather than HTTPS. The risks are all minor and unlikely. Using HTTPS makes them all go away.

The pros of using HTTPS outweigh the cons. The SerenityOS user base would have to increase a thousand fold before I would even begin to consider the cons.

What I download from GNU's servers is between me and GNU. It is nobody else's business. (Admittedly whichever file was downloaded can be inferred with minimal effort based on the size of the transmission).

In an active interception scenario, if my network connections are being middled, then installing the port will fail due to a signature mismatch. By using HTTPS we fail as soon as possible, with a clear indication that the SSL/TLS connection was at fault, not the key or signature.

If my network connections are being middled I'd want to know immediately, *before* the files were downloaded; rather than:

* after I've downloaded 1TB of compressed goatse or junk.
* after my `wget` client followed a HTTP redirect to a website hosting contraband, then downloaded said contraband.
* after `gpg` attempted to parse a potentially spoofed or malicious signature file. [[1](https://en.wikipedia.org/wiki/SigSpoof)] [[2](https://www.coresecurity.com/core-labs/advisories/gnupg-and-gnupg-clients-unsigned-data-injection-vulnerability)] [[3](https://nvd.nist.gov/vuln/detail/CVE-2020-10759)] [[4](https://lists.zx2c4.com/pipermail/password-store/2018-June/003308.html)] [[5](https://nvd.nist.gov/vuln/detail/CVE-2006-0455)]

In some instances we were downloading both the archive and the signature over clear text HTTP, and downloading the key using HTTPS. This allows installation of any signed archive so long as the signature was included in the GNU keyring. This includes a lot of keys:

```
$ gpg --keyring ./gnu-keyring.gpg --list-keys | grep @ | fgrep -v gnu.org | sort -u | wc -l
769
```

Currently we download `gnu-keyring.gpg` using HTTPS for all GNU ports, which is good. However, the current signature check functionality ignores warnings:

```
URL: https://ftp.gnu.org/gnu/gnu-keyring.gpg
+ curl https://ftp.gnu.org/gnu/gnu-keyring.gpg -L -o gnu-keyring.gpg (nocd)
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 2567k  100 2567k    0     0   773k      0  0:00:03  0:00:03 --:--:--  772k
Note: no case for file gnu-keyring.gpg.
gpg: assuming signed data in 'indent-2.2.11.tar.gz'
gpg: Signature made Fri 16 Nov 2018 02:38:44 PM PST
gpg:                using RSA key 7AE4BD64BF00D23B3434688E5E408CD91CD839D2
gpg:                issuer "andrewsh@debian.org"
gpg: Good signature from "Andrej Shadura <andrew@shadura.me>" [expired]
gpg:                 aka "Andrew Shadura <andrew@shadura.me>" [expired]
gpg:                 aka "Andrew O. Shadura <bugzilla@tut.by>" [expired]
gpg:                 aka "Andrew O. Shadoura <bugzilla@tut.by>" [expired]
gpg:                 aka "Andrei Shadura <bugzilla@tut.by>" [expired]
gpg:                 aka "Andrew O. Shadoura <Andrew.Shadoura@gmail.com>" [expired]
gpg:                 aka "Andrew Shadura <andrewsh@debian.org>" [expired]
gpg:                 aka "Andrej Shadura <andrewsh@debian.org>" [expired]
gpg: Note: This key has expired!
Primary key fingerprint: 7821 30B4 C994 4247 977B  82FD 6EA4 D231 1A2D 268D
     Subkey fingerprint: 7AE4 BD64 BF00 D23B 3434  688E 5E40 8CD9 1CD8 39D2
- Signature check OK.
```

```
URL: https://ftp.gnu.org/gnu/gnu-keyring.gpg
+ curl https://ftp.gnu.org/gnu/gnu-keyring.gpg -L -o gnu-keyring.gpg (nocd)
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 2567k  100 2567k    0     0   259k      0  0:00:09  0:00:09 --:--:--  311k
Note: no case for file gnu-keyring.gpg.
gpg: assuming signed data in 'ncurses-6.2.tar.gz'
gpg: Signature made Wed 12 Feb 2020 01:45:03 AM PST
gpg:                using DSA key C52048C0C0748FEE227D47A2702353E0F7E48EDB
gpg: Good signature from "Thomas Dickey <dickey@invisible-island.net>" [unknown]
gpg: WARNING: This key is not certified with a trusted signature!
gpg:          There is no indication that the signature belongs to the owner.
Primary key fingerprint: C520 48C0 C074 8FEE 227D  47A2 7023 53E0 F7E4 8EDB
- Signature check OK.
```

```
$ gpg --keyring ./gnu-keyring.gpg --list-keys | grep expired | wc -l
300
```

asynts:
Wow, I did not know that there was so difficult/unreliable to check a signature. I always thought that `gpg --verify` is bulletproof but it looks like this is far from the truth.

Thanks for pointing this out to me, I'll definitely do more research when I need to check signatures.

NightCheese:
> * after I've downloaded 1TB of compressed **goatse**

Haha. Even the hackers in this community are old school!

PR #6212, opened 2021-04-10
Userland: Locateutils

### LINKED ISSUE #6097: Userland: Add locate/updatedb
So I noticed there was no locate/updatedb utility, and I might be the last person to still use locate, but I thought it would be rather useful (especially if running on spinning disks, which I assume might be the case for some of the SerenityOS installations on real hardware).

After studying the mlocate implementation a bit, I started writing an indexer with a corresponding "db" (it's quite a stretch to call it a db tbh), however, it would be good to know if I should go ahead and complete this or if I'm the only one still using locate.

aowron:
Closes #6097 

Update: should have basic database + locate functionality now.

Opening the pull request early to get some eyes on it; it is in other words ~~not~~ ready for merge yet, but since I'm rather new to C++, I'd rather get some feedback early due to that I probably have done something wrong (and I'd rather avoid accumulating bad coding habits), especially when it comes to the smart pointers.

Anyways, this is a really basic version of updatedb and locate that can traverse hard drive of a serenity OS installation to generate a file database and search it.

What is missing before I'm done with this PR:
* ~~Following symlinks (will need to look into whether this should just be done when indexing or being possible to toggle in locate like in other locate implementations)~~ Seems like this is a legacy feature of most locate clones and not really needed (can neither really think of any purpose of this)
- [x] Verify that the database exists before opening in read-mode
- [x] Location of file database; currently stored in /home/anon for testing purposes. Should be stored in /var/lib (thus, updatedb should also require root privileges).
- [x] Implemented pledges

Stuff that possibly should be changed in the PR:
* Since handling of the file DB was implemented as a separate class (thus delegating all responsibility of handling the file db to a specific entity), I had to put it somewhere. Putting it in LibCore doesn't really feel like the best place to put it, but I couldn't come up with a better location.
* The LocateDb::parse_line implementation might be rather stupid; could be a better idea to use the String-class here instead.
* Pointers; I did read up a bit on smart pointers, but there's probably quite some details I haven't grasped properly, thus, I might have used them rather inappropriately.

Also, the commit log will get squashed once I'm satisfied with the implementation.


Userland/Libraries/LibCore/LocateDb.cpp
```diff
@@ -0,0 +1,43 @@
...
+LocateDb::~LocateDb()
+{
+    fclose(file_handle);
+}
+
+void LocateDb::write_header()
+{
+    // Format: magic string + version.
+    String header = String::format("%s|%d\n", "locatedb", 1);
+    fwrite(header.characters(), sizeof(char), header.length(), file_handle);
```
alimpfard, line 46: You can replace these two lines with this:
```c++
outln(file_handle, "{}|{}", "locatedb", 1);
```

Also, please prefer the newer AK/Format functions over the printf-style ones (this isn't a hard requirement, so if at some point you feel like using printf-style formatting would be more clear, feel free to use it)

Userland/Libraries/LibCore/LocateDb.cpp
```diff
@@ -0,0 +1,43 @@
...
+void LocateDb::write_directory(DirectoryInfo& dir_info)
+{
+    StringBuilder dir_chunk;
+    dir_chunk.append(String::format("%d|%s\n", (uint8_t)RowType::dir_start, dir_info.path.characters()).characters());
+    for (auto& child : dir_info.children) {
+        dir_chunk.append(String::format("%d|%s\n", (uint8_t)(child->type == FileType::file ? RowType::file : RowType::dir), child->name.characters()).characters());
+    }
+    dir_chunk.append(String::format("%d|%s\n", (uint8_t)RowType::dir_end, dir_info.path.characters()).characters());
+    dir_chunk.to_byte_buffer();
+    fwrite(dir_chunk.to_string().characters(), sizeof(char), dir_chunk.to_string().length(), file_handle);
```
alimpfard, line 58: This looks like it could potentially allocate a huge buffer, maybe write directly to the output file?
if corruption is possible, you could also first output to a temporary file then `rename()` it (with the typical `EXDEV` dance if needed)

Userland/Utilities/updatedb.cpp
```diff
@@ -0,0 +1,75 @@
...
+
+    NonnullOwnPtr<Crawler> crawler = make<Crawler>(String("/"));
+    NonnullOwnPtr<Core::LocateDb> file_db = make<Core::LocateDb>("/home/anon/testdb.db", Core::LocateDbMode::write);
+    file_db->write_header();
+
+    while (crawler->dirs_in_queue() > 0) {
+        auto dir_info = crawler->index_next_directory();
+        file_db->write_directory(*dir_info);
+    }
+}
```
alimpfard: Missing newline at the end of file : )

Userland/Libraries/LibCore/LocateDb.cpp
```diff
@@ -32,12 +58,86 @@ void LocateDb::write_directory(DirectoryInfo& dir_info)
...
+        break;
+    case (RowType::dir):
+        found_type = RowType::dir;
+        break;
+    case (RowType::file):
+        found_type = RowType::file;
+        break;
+    default:
+        dbgln("Buffer length: %zu\n", String(buffer).length());
+        dbgln("Buffer on fail: %s\n", buffer);
```
alimpfard, line 88: This isn't going to work, `dbgln()`'s placeholders are braces (e.g. `"Buffer on fail: {}"`), and it will automatically append a newline, no need for `\n`.

Userland/Libraries/LibCore/LocateDb.cpp
```diff
@@ -32,12 +58,86 @@ void LocateDb::write_directory(DirectoryInfo& dir_info)
...
+        found_type = RowType::dir_end;
+        break;
+    case (RowType::dir):
+        found_type = RowType::dir;
+        break;
+    case (RowType::file):
+        found_type = RowType::file;
+        break;
+    default:
+        dbgln("Buffer length: %zu\n", String(buffer).length());
```
alimpfard, line 87: This will allocate a string on the heap and copy the buffer into it, just because you want the length :P 
Please just use `strlen()`.

Userland/Utilities/locate.cpp
```diff
@@ -0,0 +1,62 @@
...
+    args_parser.parse(argc, argv);
+
+    auto locate_db = Core::LocateDb("/home/anon/testdb.db", Core::LocateDbMode::read);
+    VERIFY(locate_db.verify_header());
+    String keyword = String(keyword_holder[0]);
+    CaseSensitivity case_sensitivity = o_case_insensitive ? CaseSensitivity::CaseInsensitive : CaseSensitivity::CaseSensitive;
+
+    while (auto directory = locate_db.get_next_directory()) {
+        if (directory->path.contains(keyword, case_sensitivity)) {
+            printf("%s\n", directory->path.characters());
```
alimpfard, line 49: You could also use `outln("{}", directory->path);` if you want.
one less `strlen` :shrug:

Userland/Libraries/LibCore/LocateDb.cpp
aowron: The idea behind writing in chunks was to avoid hammering the disk with write requests; however, I might be trying to solve a problem that doesn't exist by doing it this way? Maybe it would be a good idea to at least put the StringBuilder on the heap?

Thx for the comments by the way, this was exactly the type of feedback I was looking for :) (will obviously traverse the other issues you had noticed as well)

alimpfard: stdio has a bunch of buffers already.
At this level, I don't think you need to worry about disk requests, let the kernel and libc worry about that :P 

As far as StringBuilder goes, it will start with a tiny buffer on the stack for small operations, and switch to heap allocations whenever needed, so...

aowron:
Something else that goes into the "TODO" list: serialize the directory paths and file names in some way; apparently, "|" is a valid character to use for both files and directories (both in Serenity and Linux; and guess in any kind of *nix in general).

aowron:
Should finally be done for review now. Squashed quite a bit of commits, although, are there any additional commits I should squash except for the last three linting fixes?

Userland/Libraries/LibLocate/Crawler.cpp
```diff
@@ -0,0 +1,90 @@
...
+
+namespace Locate {
+
+Crawler::Crawler(String path)
+{
+    m_dir_queue.enqueue(make<Locate::DirectoryInfo>(Locate::DirectoryInfo {
+        path,
+        m_id_counter,
+        0,
+        SinglyLinkedListWithCount<NonnullOwnPtr<Locate::ChildInfo>>() }));
```
alimpfard, line 37: You're invoking `DirectoryInfo`'s move ctor here (if there is one, otherwise this is a copy), is there any reason not to do this?
```c++
m_dir_queue.enqueue(make<Locate::DirectoryInfo>(
        path,
        m_id_counter,
        0,
        SinglyLinkedListWithCount<NonnullOwnPtr<Locate::ChildInfo>>()));
```

Userland/Libraries/LibLocate/Crawler.cpp
```diff
@@ -0,0 +1,90 @@
...
+
+    while (dir_iterator.has_next()) {
+        NonnullOwnPtr<Locate::ChildInfo> child = make<Locate::ChildInfo>();
+        child->name = dir_iterator.next_path();
+
+        String fullpath;
+        if (cur_path == "/")
+            fullpath = String(String::format("%s%s", cur_path.characters(), child->name.characters()));
+        else
+            fullpath = String(String::format("%s/%s", cur_path.characters(), child->name.characters()));
```
alimpfard, line 55: `String::formatted()` :eyes:

Userland/Libraries/LibLocate/Crawler.cpp
```diff
@@ -0,0 +1,90 @@
...
+        lstat(fullpath.characters(), &stat_info);
+        child->db_id = m_id_counter;
+        child->parent_db_id = cur_dir_info->db_id;
+        child->user_owner = stat_info.st_uid;
+        child->group_owner = stat_info.st_gid;
+        child->st_mode = stat_info.st_mode & ~0xFE00; //Only keep the permission bits
+
+        if (S_ISDIR(stat_info.st_mode)) {
+            child->type = Locate::FileType::dir;
+            m_dir_queue.enqueue(make<Locate::DirectoryInfo>(Locate::DirectoryInfo {
```
alimpfard, line 67: Same make/move or make/copy thing here

Userland/Libraries/LibLocate/Crawler.h
```diff
@@ -0,0 +1,47 @@
...
+namespace Locate {
+
+class Crawler {
+public:
+    Crawler(String path);
+    NonnullOwnPtr<Locate::DirectoryInfo> index_next_directory();
+    size_t dirs_in_queue();
+
+private:
+    Queue<NonnullOwnPtr<Locate::DirectoryInfo>> m_dir_queue;
```
alimpfard, line 44: Any reason why you're putting this in a `NonnullOwnPtr` (over just a normal `DirectoryInfo`)?

Userland/Libraries/LibLocate/LocateDb.cpp
```diff
@@ -0,0 +1,221 @@
...
+ */
+
+#include <LibLocate/LocateDb.h>
+
+namespace Locate {
+
+LocateDb::LocateDb(String path, LocateDbMode mode)
+{
+    if (!Core::File::exists(LOCATE_DB_PATH)) {
+        VERIFY(Core::File::ensure_parent_directories(LOCATE_DB_PATH));
```
alimpfard, line 34: While things in `VERIFY()` are always run, please try avoid putting things with side effects in there.

Userland/Libraries/LibLocate/LocateDb.cpp
```diff
@@ -0,0 +1,221 @@
...
+
+void LocateDb::write_directory(DirectoryInfo& dir_info)
+{
+    struct DirStartEndHeader dir_header {
+        .type = (uint8_t)ChunkType::dir_start,
+        .db_id = dir_info.db_id,
+        .parent_db_id = dir_info.parent_db_id,
+        .path_size = dir_info.path.length() + 1
+    };
+    fwrite((char*)&dir_header, sizeof(struct DirStartEndHeader), 1, m_file_handle);
```
alimpfard, line 70: Now that we have `bit_cast`, pls `bit_cast<char*>(&dir_header)`.

Userland/Libraries/LibLocate/LocateDb.cpp
```diff
@@ -0,0 +1,221 @@
...
+    fwrite(dir_info.path.characters(), sizeof(uint8_t), dir_info.path.length() + 1, m_file_handle);
+
+    for (auto& child : dir_info.children) {
+        struct ChildFileData cur_child {
```
alimpfard, line 74: `struct` is superfluous.
Not an issue, just a suggestion.

Userland/Libraries/LibLocate/LocateDb.cpp
```diff
@@ -0,0 +1,221 @@
...
+            .type = (uint8_t)(child->type == FileType::file ? ChunkType::file : ChunkType::dir),
+            .db_id = child->db_id,
+            .parent_db_id = child->parent_db_id,
+            .user_owner = child->user_owner,
+            .group_owner = child->group_owner,
+            .st_mode = child->st_mode,
+            .name_size = child->name.length() + 1
+        };
+
+        fwrite((char*)&cur_child, sizeof(struct ChildFileData), 1, m_file_handle);
```
alimpfard, line 84: `bit_cast` here too? (and in multiple other places in this file)

Userland/Libraries/LibLocate/LocateDb.cpp
```diff
@@ -0,0 +1,221 @@
...
+    fwrite((char*)&dir_header, sizeof(struct DirStartEndHeader), 1, m_file_handle);
+    fwrite(dir_info.path.characters(), sizeof(uint8_t), dir_info.path.length() + 1, m_file_handle);
+}
+
+bool LocateDb::verify_header()
+{
+    char header[] = "locatedb|1";
+    char buffer[strlen(header) + 1];
+    size_t bytes_read = fread(&buffer, sizeof(char), strlen("locatedb|1") + 1, m_file_handle);
+    VERIFY(bytes_read > 0);
```
alimpfard, line 98: what if it `== 0`?
why not just return `false`?

Userland/Libraries/LibLocate/LocateDb.cpp
```diff
@@ -0,0 +1,221 @@
...
+    if (String(buffer) == "locatedb|1")
```
alimpfard, line 99: Please avoid constructing a temporary string like this, instead try `if ("locatedb|1"sv == buffer)`.

Userland/Libraries/LibLocate/LocateDb.cpp
```diff
@@ -0,0 +1,221 @@
...
+
+    dir_header.type = (uint8_t)ChunkType::dir_end;
+    fwrite((char*)&dir_header, sizeof(struct DirStartEndHeader), 1, m_file_handle);
+    fwrite(dir_info.path.characters(), sizeof(uint8_t), dir_info.path.length() + 1, m_file_handle);
+}
+
+bool LocateDb::verify_header()
+{
+    char header[] = "locatedb|1";
+    char buffer[strlen(header) + 1];
```
alimpfard, line 96: No VLAs please.

Userland/Libraries/LibLocate/LocateDb.cpp
```diff
@@ -0,0 +1,221 @@
...
+{
+    uint8_t type_byte;
+    size_t bytes_read = fread(&type_byte, sizeof(uint8_t), 1, m_file_handle);
+    fseek(m_file_handle, -1, SEEK_CUR);
+    VERIFY(feof(m_file_handle) == 0);
+    VERIFY(bytes_read == 1);
+    ChunkType found_type;
+    switch (ChunkType(type_byte)) {
+    case (ChunkType::dir_start):
+        found_type = ChunkType::dir_start;
```
alimpfard, line 114: Not a problem, but most enums are PascalCase here, I think.

Userland/Libraries/LibLocate/LocateDb.cpp
```diff
@@ -0,0 +1,221 @@
...
+
+void LocateDb::parse_dir_info(DirectoryInfo& dir_info)
+{
+    struct DirStartEndHeader header;
+    size_t bytes_read = fread(&header, sizeof(struct DirStartEndHeader), 1, m_file_handle);
+    VERIFY(bytes_read > 0);
+    dir_info.db_id = header.db_id;
+    dir_info.parent_db_id = header.parent_db_id;
+
+    char path_from_file[header.path_size];
```
alimpfard, line 140: Please avoid VLAs.
(occurs many times in this file)

Userland/Libraries/LibLocate/LocateDb.h
```diff
@@ -0,0 +1,94 @@
...
+#include <AK/String.h>
+#include <AK/StringBuilder.h>
+#include <LibC/string.h>
+#include <LibC/unistd.h>
+#include <LibCore/File.h>
+#include <LibLocate/Types.h>
+#include <stdio.h>
+#include <sys/stat.h>
+
+#define LOCATE_DB_PATH "/var/lib/locate/locate.db"
```
alimpfard, line 44: `constexpr static auto`?

alimpfard (changes requested):

Userland/Libraries/LibLocate/Crawler.cpp
aowron: Since DirectoryInfo is a struct, it doesn't (afaik) have constructor. Did btw. try your suggestion and it will indeed not work (due to the lack of a constructor in DirectoryInfo). Since DirectoryInfo is just a container, I thought that it would be overkill to make a complete class out of it.

I did also notice now that I hadn't used the .attribute = value notation, which hurts my eyes (well, rather, the lack of it), so fixed that was well while looking at this.

alimpfard: Alright, it's probably gonna be optimised away anyway

aowron: Still a bit new to smart pointers, so I might have done something stupid here, but the idea was basically only to have the DirectoryInfo stored in one place and thus have it moved when dequeued rather than copied (which is what I understood would happen from when reading up on the workings of smart pointers). I.e. from what I undestood, this should also result in the ownership changing when getting dequeued (and thus, the DirectoryInfo object after being handled in the locate.cpp being cleaned up at the end of the main loop due to going out of scope).

Userland/Libraries/LibLocate/Crawler.h
aowron: Probably habit from C/old c++ (thinking along the lines of "creating an object => I should obviously put this on the heap with a "new" so that I don't fill up the stack and cause a stack overflow").

Userland/Libraries/LibLocate/Crawler.cpp
alimpfard: I'm not too sure which would be better, the members of this struct look rather cheap to move() out (which is what dequeue should be doing), maybe except the String member which has to go through an atomic op...
I guess it's alright to keep this as-is (so feel free to disregard the question below about `NonnullOwnPtr`)

aowron: > Alright, it's probably gonna be optimised away anyway

I'd rather redo it properly though :)

Userland/Libraries/LibLocate/Crawler.h
alimpfard: Well, what you have here makes it cheap to move these around, so it really depends on which has more impact in this case, the heap churn, or a shallow copy of the struct (which involves an atomic op, so it's not very cheap)

aowron: My thinking was basically "move: cheap, copy: expensive", although, from what I am noticing now is though that putting DirectoryInfo objects directly into the queue will result in the object being returned, so still might be better to store NonnullOwnPtr:s in it after all.

alimpfard: Sure, that's fine :)

Userland/Libraries/LibLocate/LocateDb.cpp
aowron: ...and I just realized that I didn't read through the code of "ensure_parent_directories" properly, thinking that it only would check whether the path to the file existed, while it in fact creates directories; that might explain why I had to add so many additional pledges then.

Userland/Libraries/LibLocate/LocateDb.cpp
aowron: Might have been a leftover from debugging; although, the CI won't like freads where the read_bytes value isn't read (since gcc will generate warnings for functions where a value is being returned but that value isn't stored in a var, which in turn has to be used in some way to not generate a warning). Anyway, moving that check into the "if"-check instead.

aowron:
Suggested changes made; once I know it's good enough, I'll squash the commit log (from what I've understood from Contributing.md, this should be done after review; it has been squashed a bit before though).

alimpfard (approved):
LGTM : )

aowron:
Commits squashed; although I'm worried they might not be atomic enough (however, it might be hard to make them atomic due to that fixes and changes to the various aspects of the locate utils were changed over the course of the development). Also, please let me know if I should squash them further, might be too much of a "implementation process" (I'd rather make proper commits than have many commits :) ).

alimpfard (approved):
I think the commits are fine.
As a finishing touch, please put "Closes #6097" in the PR message (you've written "Implements") to make github auto-close that issue when this is merged : )

aowron:
> I think the commits are fine.
> As a finishing touch, please put "Closes #6097" in the PR message (you've written "Implements") to make github auto-close that issue when this is merged : )

Thanks! Added that to the message now. Not really that used to github either (used Bitbucket at my previous workplace)

Userland/Libraries/LibLocate/LocateDb.h
```diff
@@ -0,0 +1,94 @@
...
+    long m_file_size;
+    FILE* m_file_handle;
+    LocateDbMode m_mode;
+    HashMap<uint32_t, NonnullOwnPtr<PermissionInfo>> m_path_relations;
+
+    enum class ChunkType {
+        DirStart = 1,
+        DirEnd = 2,
+        File = 3,
+        Dir = 4
```
linusg, line 69: No need to explicitly assign all values, please add a comma to the last enum member, and I don't think we should abbreviate directory here (https://github.com/SerenityOS/serenity/blob/master/Documentation/CodingStyle.md#names).

```suggestion
        DirectoryStart,
        DirectoryEnd,
        File,
        Directory,
```

Userland/Libraries/LibLocate/LocateDb.cpp
```diff
@@ -0,0 +1,220 @@
...
+        found_type = ChunkType::Dir;
+        break;
+    case (ChunkType::File):
+        found_type = ChunkType::File;
+        break;
+    default:
+        VERIFY_NOT_REACHED();
+    }
+
+    return found_type;
```
linusg, line 126: What's the reason for this to be a switch/case? Isn't this just a lengthy way of doing

```suggestion
    VERIFY(type_byte < ChunkType::__Count);
    return type_byte;
```

Userland/Libraries/LibLocate/LocateDb.cpp
```diff
@@ -0,0 +1,220 @@
...
+
+    return child_info;
+}
+
+OwnPtr<DirectoryInfo> LocateDb::get_next_directory()
+{
+    OwnPtr<DirectoryInfo> cur_dir = make<DirectoryInfo>();
+
+    while (true) {
+
```
linusg, line 172:
```suggestion
```

Userland/Libraries/LibLocate/Types.h
```diff
@@ -0,0 +1,67 @@
...
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include <AK/SinglyLinkedListWithCount.h>
+
+#pragma once
```
linusg, line 29: This should go before the include

Userland/Libraries/LibLocate/Crawler.h
```diff
@@ -0,0 +1,47 @@
...
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include <AK/Queue.h>
+#include <AK/String.h>
+#include <LibCore/DirIterator.h>
+#include <LibLocate/Types.h>
+#include <sys/stat.h>
+
+#pragma once
```
linusg, line 33: This should go before the includes

Userland/Utilities/updatedb.cpp
```diff
@@ -0,0 +1,53 @@
...
+ */
+
+#include <AK/NonnullOwnPtrVector.h>
+#include <AK/String.h>
+#include <LibLocate/Crawler.h>
+#include <LibLocate/LocateDb.h>
+
+int main()
+{
+
```
linusg, line 34:
```suggestion
```

Userland/Libraries/LibLocate/LocateDb.cpp
aowron: Yeah, I clearly overenginereed that one now when you mention it... your solution is way more elegant :)

linusg:
Please squash the code review changes into the individual commits instead of all adding them in one fixup commit.

aowron:
> Please squash the code review changes into the individual commits instead of all adding them in one fixup commit.

Do you mean every single fix into an individual commit (that would result in 10+ additional commits) or just the changes made after alimpfard's review comments in one commit, and the changes from your review in a separate commit?

linusg:
neither, each fix should be squashed into the commit where the reviewed piece of code was added or changed.

aowron:
So after fighting git for a few hours trying to split those changes by interactively rebasing and resetting the head, I can't really get them to fit nicely into older commits without ruining the whole pull request (well, getting a mess of old and new changes). I could  just throw away the commit history of the pull request, commit LibLocate and the utils in two or three separate commits, thus getting a synthetic albeit nicer commit history; would that be ok?

linusg:
sounds good, for example you could add the library and CLI utilities each separately, but do whatever you think works best :)

aowron:
> sounds good, for example you could add the library and CLI utilities each separately, but do whatever you think works best :)

Great! Hopefully, the commit log should be quite a bit neater now. I did indeed end up going for separate commits for the CLI utils, and one separate commit for the LibLocate library itself.

aowron:
@linusg Would you mind re-reviewing this one with the cleaned up commit history? This PR should be done for now unless there is something else I should improve/fix in it. I will probably need to revisit LibLocate and update it once PR #6513 is merged though.

Meta/build-root-filesystem.sh
```diff
@@ -81,6 +85,11 @@ chown 0:$utmp_gid mnt/var/run/utmp
 chmod 664 mnt/var/run/utmp
 echo "done"
 
+printf "create locate folder..."
```
linusg, line 88:
```suggestion
printf "createing locate folder... "
```

trailing space is on purpose as this doesn't print a newline, so the "done" below will go after the "...".

Userland/Libraries/LibLocate/CMakeLists.txt
```diff
@@ -0,0 +1,7 @@
+set(SOURCES
+    LocateDB.cpp
+    Crawler.cpp
```
linusg, line 3: Maybe not super important for such small files, but we generally keep `SOURCES` in alphabetical order.

```suggestion
    Crawler.cpp
    LocateDB.cpp
```

Userland/Libraries/LibLocate/Crawler.cpp
```diff
@@ -0,0 +1,90 @@
...
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include <LibLocate/Crawler.h>
+
+namespace Locate {
+
+Crawler::Crawler(String path)
```
linusg, line 31: I think this should either take a `String const&` here or `move()` the path below.

Userland/Libraries/LibLocate/Crawler.cpp
```diff
@@ -0,0 +1,90 @@
...
+        .parent_db_id = 0,
+        .children = SinglyLinkedListWithCount<NonnullOwnPtr<Locate::ChildInfo>>() }));
+    m_id_counter++;
+}
+
+NonnullOwnPtr<Locate::DirectoryInfo> Crawler::index_next_directory()
+{
+    auto cur_dir_info = m_dir_queue.dequeue();
+    String cur_path = cur_dir_info->path.characters();
+    Core::DirIterator dir_iterator(cur_path, Core::DirIterator::SkipParentAndBaseDir);
```
linusg, line 45: "DirIterator" is a bit of an exception (not sure why tbh), but generally please avoid abbreviating names, unless the abbreviation is more clear/common than the regular word. `cur` -> `current`, and ideally `dir` -> `directory`. It's also called `Locate::DirectoryInfo` so that makes it more consistent :)

Applies to more parts of this PR, I won't mark all of them individually.

Userland/Libraries/LibLocate/Crawler.cpp
```diff
@@ -0,0 +1,90 @@
...
+
+    while (dir_iterator.has_next()) {
+        NonnullOwnPtr<Locate::ChildInfo> child = make<Locate::ChildInfo>();
```
linusg, line 48:
```suggestion
        auto child = make<Locate::ChildInfo>();
```

Userland/Libraries/LibLocate/Crawler.cpp
```diff
@@ -0,0 +1,90 @@
...
+        child->name = dir_iterator.next_path();
+
+        String fullpath;
+        if (cur_path == "/")
+            fullpath = String(String::formatted("{}{}", cur_path.characters(), child->name.characters()));
+        else
+            fullpath = String(String::formatted("{}/{}", cur_path.characters(), child->name.characters()));
```
linusg, line 55: No need for `.characters()` when using `{}` formatting.

```suggestion
            fullpath = String(String::formatted("{}{}", cur_path, child->name));
        else
            fullpath = String(String::formatted("{}/{}", cur_path, child->name));
```

Userland/Libraries/LibLocate/Crawler.cpp
```diff
@@ -0,0 +1,90 @@
...
+
+        struct stat stat_info;
+        lstat(fullpath.characters(), &stat_info);
+        child->db_id = m_id_counter;
+        child->parent_db_id = cur_dir_info->db_id;
+        child->user_owner = stat_info.st_uid;
+        child->group_owner = stat_info.st_gid;
+        child->st_mode = stat_info.st_mode & ~0xFE00; //Only keep the permission bits
```
linusg, line 63:
```suggestion
        child->st_mode = stat_info.st_mode & ~0xFE00; // Only keep the permission bits
```

Userland/Libraries/LibLocate/LocateDB.cpp
```diff
@@ -0,0 +1,203 @@
...
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include <LibLocate/LocateDB.h>
+
+namespace Locate {
+
+LocateDB::LocateDB(String path, LocateDbMode mode)
```
linusg, line 31: path can be a `String const&`, no need to copy

Userland/Libraries/LibLocate/LocateDB.cpp
```diff
@@ -0,0 +1,203 @@
...
+        auto file_handle = Core::File::open(locate_db_path, Core::IODevice::WriteOnly, 0640);
+        if (file_handle.is_error()) {
+            outln("Got error while trying to create locate db file: {}", file_handle.error());
+            VERIFY_NOT_REACHED();
+        }
+        file_handle.value()->close();
+        chown(locate_db_path, 0, 18);
+    }
+
+    m_file_handle = fopen(path.characters(), (mode == LocateDbMode::Write ? "wb" : "rb"));
```
linusg, line 43:
```suggestion
    m_file_handle = fopen(path.characters(), mode == LocateDbMode::Write ? "wb" : "rb");
```

Userland/Libraries/LibLocate/LocateDB.cpp
```diff
@@ -0,0 +1,203 @@
...
+
+LocateDB::~LocateDB()
+{
+    fclose(m_file_handle);
+}
+
+void LocateDB::write_header()
+{
+    // Format: magic string + version.
+    String db_header = String::formatted("{}|{}", "locatedb", 1);
```
linusg, line 57: Why does this need formatting, isn't this sufficient?

```suggestion
    String db_header("locatedb|1");
```

Userland/Libraries/LibLocate/LocateDB.cpp
```diff
@@ -0,0 +1,203 @@
...
+
+    dir_header.type = (uint8_t)ChunkType::DirectoryEnd;
+    fwrite(bit_cast<char*>(&dir_header), sizeof(struct DirStartEndHeader), 1, m_file_handle);
+    fwrite(dir_info.path.characters(), sizeof(uint8_t), dir_info.path.length() + 1, m_file_handle);
+}
+
+bool LocateDB::verify_header()
+{
+    char buffer[BUFSIZ];
+    size_t bytes_read = fread(&buffer, sizeof(char), strlen("locatedb|1") + 1, m_file_handle);
```
linusg, line 95: As this is hardcoded a second time here, and then a third time below, can we make it a constant instead? Something like

```cpp
constexpr const StringView MAGIC_HEADER { "locatedb|1" };
```

Userland/Libraries/LibLocate/LocateDB.cpp
```diff
@@ -0,0 +1,203 @@
...
+            parse_dir_info(verification_dir);
+            VERIFY(cur_dir->path.length() != 0 && cur_dir->path == verification_dir.path);
+            return cur_dir;
+        } else {
+            auto child = parse_child_info();
+            String path_string;
+            if (cur_dir->path == "/")
+                path_string = String::formatted("{}{}", cur_dir->path, child->name);
+            else
+                //Don't waste memory on file names in the path hierarchy
```
linusg, line 183:
```suggestion
                // Don't waste memory on file names in the path hierarchy.
```

Userland/Utilities/locate.cpp
```diff
@@ -0,0 +1,97 @@
...
+#include <unistd.h>
+
+bool check_permissions(uint32_t db_id, Locate::LocateDB& locate_db);
+
+bool check_permissions(uint32_t db_id, Locate::LocateDB& locate_db)
+{
+    uint32_t next_id = db_id;
+    while (true) {
+        const Locate::PermissionInfo* permission_info = locate_db.get_permission_info(next_id);
+        //Reached /, thus, the readable path for the user is not broken along the way, thus, the file can be shown
```
linusg, line 40:
```suggestion
        // Reached /, thus, the readable path for the user is not broken along the way, thus, the file can be shown.
```

linusg (commented):
Thanks for the ping, some more feedback :)

Userland/Libraries/LibLocate/LocateDB.cpp
```diff
@@ -0,0 +1,203 @@
...
+LocateDB::LocateDB(String path, LocateDbMode mode)
+{
+    if (!Core::File::exists(locate_db_path)) {
+        auto file_handle = Core::File::open(locate_db_path, Core::IODevice::WriteOnly, 0640);
+        if (file_handle.is_error()) {
+            outln("Got error while trying to create locate db file: {}", file_handle.error());
+            VERIFY_NOT_REACHED();
+        }
+        file_handle.value()->close();
+        chown(locate_db_path, 0, 18);
```
awesomekling, line 40: Missing error check.

Userland/Libraries/LibLocate/LocateDB.cpp
```diff
@@ -0,0 +1,203 @@
...
+    }
+
+    m_file_handle = fopen(path.characters(), (mode == LocateDbMode::Write ? "wb" : "rb"));
+    fseek(m_file_handle, 0, SEEK_END);
+    m_file_size = ftell(m_file_handle);
+    fseek(m_file_handle, 0, SEEK_SET);
```
awesomekling, line 46: Missing error checks on all these calls.

Userland/Libraries/LibLocate/LocateDB.cpp
```diff
@@ -0,0 +1,203 @@
...
+LocateDB::~LocateDB()
+{
+    fclose(m_file_handle);
+}
+
+void LocateDB::write_header()
+{
+    // Format: magic string + version.
+    String db_header = String::formatted("{}|{}", "locatedb", 1);
+    fwrite(db_header.characters(), sizeof(uint8_t), db_header.length() + 1, m_file_handle);
```
awesomekling, line 58: Missing error check.

Userland/Libraries/LibLocate/LocateDB.cpp
```diff
@@ -0,0 +1,203 @@
...
+void LocateDB::write_directory(DirectoryInfo& dir_info)
+{
+    struct DirStartEndHeader dir_header {
+        .type = (uint8_t)ChunkType::DirectoryStart,
+        .db_id = dir_info.db_id,
+        .parent_db_id = dir_info.parent_db_id,
+        .path_size = dir_info.path.length() + 1
+    };
+    fwrite(bit_cast<char*>(&dir_header), sizeof(struct DirStartEndHeader), 1, m_file_handle);
+    fwrite(dir_info.path.characters(), sizeof(uint8_t), dir_info.path.length() + 1, m_file_handle);
```
awesomekling, line 70: Missing error checks.

Userland/Libraries/LibLocate/LocateDB.cpp
```diff
@@ -0,0 +1,203 @@
...
+            .db_id = child->db_id,
+            .parent_db_id = child->parent_db_id,
+            .user_owner = child->user_owner,
+            .group_owner = child->group_owner,
+            .st_mode = child->st_mode,
+            .name_size = child->name.length() + 1
+        };
+
+        fwrite(bit_cast<char*>(&cur_child), sizeof(struct ChildFileData), 1, m_file_handle);
+        fwrite(child->name.characters(), sizeof(uint8_t), child->name.length() + 1, m_file_handle);
```
awesomekling, line 84: Missing error checks.

Userland/Libraries/LibLocate/LocateDB.cpp
```diff
@@ -0,0 +1,203 @@
...
+    }
+
+    dir_header.type = (uint8_t)ChunkType::DirectoryEnd;
+    fwrite(bit_cast<char*>(&dir_header), sizeof(struct DirStartEndHeader), 1, m_file_handle);
+    fwrite(dir_info.path.characters(), sizeof(uint8_t), dir_info.path.length() + 1, m_file_handle);
```
awesomekling, line 89: Missing error checks.

Meta/build-root-filesystem.sh
IdanHo: small typo: "createing" -> "creating"

Userland/Libraries/LibLocate/LocateDB.cpp
awesomekling, line 95: Missing error check.

Userland/Libraries/LibLocate/LocateDB.cpp
```diff
@@ -0,0 +1,203 @@
...
+    if (bytes_read > 0 && "locatedb|1"sv == buffer)
+        return true;
+    return false;
+}
+
+LocateDB::ChunkType LocateDB::identify_chunk()
+{
+    uint8_t type_byte;
+    size_t bytes_read = fread(&type_byte, sizeof(uint8_t), 1, m_file_handle);
+    fseek(m_file_handle, -1, SEEK_CUR);
```
awesomekling, line 105: Missing error checks.

...you get the idea by now :^)

awesomekling (changes requested):

Meta/build-root-filesystem.sh
linusg: whoops, thanks!

Userland/Libraries/LibLocate/Crawler.cpp
aowron: Well, this is certainly a case of "why did I even implement that?". So, initially, I thought of storing the permissions and checking them upon loading the directory, but later noticed the "access"-function, which made life a lot easier. Anyway, I have some cleaning up to do here (the whole permission bit masking and storing it in the db is totally unnecessary), since this will result in the locatedb storing a lot of unnecessary data.

aowron:
Should have covered the changes mentioned in the reviews now; the commits have been squashed, so please let me know if I should have squashed them after the re-review rather than before.

Couldn't come up with a better way of doing the error checking than using a lot of "VERIFY"'s (I was consdering adding methods for reading and writing within the LocateDB class, but since that would have obfuscated line number mentioned by the assertion, I assume that's not a good idea). And yes, I was itching for exceptions there, but to keep in line with the coding standards, I won't use them.

aowron:
@linusg @awesomekling Could i please get approval for running the CI-workflow and a re-review on the PR after making the changes you requested?

awesomekling:
I'm not sure this is a good addition to the system, at least not in such a simplistic form.

I would much rather see us build a generic indexing and locating service like Spotlight on macOS, for rich metadata indexing of arbitrary content (including files).
`locate` could then be implemented as a simple query against this index.

aowron:
> I'm not sure this is a good addition to the system, at least not in such a simplistic form.
> 
> I would much rather see us build a generic indexing and locating service like Spotlight on macOS, for rich metadata indexing of arbitrary content (including files).
> `locate` could then be implemented as a simple query against this index.

Yeah, that sounds sensible; especially once the libsql library has reached a usable status. Well, I did at least learn quite a lot from this exercise :).

PR #6214, opened 2021-04-10
AK+Everywhere: Make StdLibExtras templates less wrapper-y

alimpfard:
This commit makes the user-facing StdLibExtras templates and utilities
arguably more nice-looking by removing the need to reach into the
wrapper structs generated by them to get the value/type needed.
The C++ standard library had to invent `_v` and `_t` variants (likely
because of backwards compat), but we don't need to cater to any codebase
except our own, so might as well have good things for free. :^)

Still draft because I'm not sure if I missed any others.


alimpfard:
Looks like that's all of them.

PR #6216, opened 2021-04-10
HackStudio: Add ClassView

itamar8910:
This enables the user to view and navigate classes with a TreeView that is updated by the LanguageServer as it parses the code.
Thanks to @awesomekling for suggesting this :)

class-view


awesomekling:
Oh heck yes! Can we move it to the left hand side though? (And make it an alternative to the file tree view :))

itamar8910:
> Oh heck yes! Can we move it to the left hand side though? (And make it an alternative to the file tree view :))

Sure yeah, will do that :+1:

itamar8910:
Okay, ClassView is now an alternative tab to files TreeView.
@awesomekling is this what you had in mind?

(Also, CI fails because of the homebrew bintray issue)

hs-classview-side

PR #6217, opened 2021-04-10
Build fixes for a couple of ports

gunnarbeutner:
This fixes building the following ports:

* tr: This port expects the `<ctype.h>` functions to also be available as macros (as per https://pubs.opengroup.org/onlinepubs/7908799/xsh/ctype.h.html).
* c-ray: Expects `sockaddr_in` to be defined after including `<arpa/inet.h>`. Ideally c-ray should include `<netinet/in.h>` on its own but https://pubs.opengroup.org/onlinepubs/7908799/xns/arpainet.h.html says that <arpa/inet.h> may optionally include `<netinet/in.h>` and `<inttypes.h>` so I added those there.
* dash: Partially fixed: Building statically linked executables is broken at the moment (because `libc.a` doesn't provide some `cxxabiv1::*` symbols while `libc.so` has those) so I removed the configure option `--enable-static`
* bash: Re-installing the port failed because the symlink already exists, so I used `ln -sf` instead to overwrite it.

Building dash still fails because the `config.h` file is used to build executables for both the cross-compilation target as well as the host system (see `HELPERS` in the Makefile). This is a problem if the host system has `stat64` (which Linux does) while the target doesn't (SerenityOS doesn't). I don't have a fix for that yet.

### DIFF
--- a/Ports/bash/package.sh
+++ b/Ports/bash/package.sh
@@ -17,5 +17,5 @@ build() {
 
 post_install() {
     mkdir -p "${SERENITY_BUILD_DIR}/Root/bin"
-    ln -s /usr/local/bin/bash "${SERENITY_BUILD_DIR}/Root/bin/bash"
+    ln -sf /usr/local/bin/bash "${SERENITY_BUILD_DIR}/Root/bin/bash"
 }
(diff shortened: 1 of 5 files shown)


awesomekling:
Very nice fixes :)

PR #6218, opened 2021-04-10
LibJS+LibWeb: Implement various missing things to get a Vue.js hello world running

linusg:
Not quite there yet, it currently chokes on missing `cloneNode()` - but this is getting crowded already, so here's a first PR.

### DIFF
--- a/Userland/Libraries/LibWeb/DOM/Document.cpp
+++ b/Userland/Libraries/LibWeb/DOM/Document.cpp
@@ -1,5 +1,6 @@
 /*
  * Copyright (c) 2018-2021, Andreas Kling <kling@serenityos.org>
+ * Copyright (c) 2021, Linus Groh <mail@linusgroh.de>
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
@@ -63,6 +64,7 @@
 #include <LibWeb/Origin.h>
 #include <LibWeb/Page/Frame.h>
 #include <LibWeb/SVG/TagNames.h>
+#include <LibWeb/UIEvents/MouseEvent.h>
 #include <ctype.h>
 
 namespace Web::DOM {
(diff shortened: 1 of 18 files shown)


Userland/Libraries/LibWeb/DOM/Document.cpp
```diff
@@ -628,6 +630,60 @@ NonnullRefPtr<Range> Document::create_range()
...
+        event = Event::create(""); // FIXME: Create FocusEvent
+    } else if (interface.matches("hashchangeevent")) {
+        event = Event::create(""); // FIXME: Create HashChangeEvent
+    } else if (interface.matches("htmlevents")) {
+        event = Event::create("");
+    } else if (interface.matches("keyboardevent")) {
+        event = Event::create(""); // FIXME: Create KeyboardEvent
+    } else if (interface.matches("messageevent")) {
+        event = Event::create(""); // FIXME: Create MessageEvent
+    } else if (interface.matches("mouseevent") || interface.matches("mouseevents")) {
```
Lubrsi, line 661: As a shorthand for this, we can do:
```suggestion
    } else if (interface.is_one_of("mouseevent", "mouseevents")) {
```

Userland/Libraries/LibWeb/DOM/Document.cpp
```diff
@@ -628,6 +630,60 @@ NonnullRefPtr<Range> Document::create_range()
...
+        event = UIEvents::MouseEvent::create("", 0, 0);
+    } else if (interface.matches("storageevent")) {
+        event = Event::create(""); // FIXME: Create StorageEvent
+    } else if (interface.matches("svgevents")) {
+        event = Event::create("");
+    } else if (interface.matches("textevent")) {
+        event = Event::create(""); // FIXME: Create CompositionEvent
+    } else if (interface.matches("touchevent")) {
+        event = Event::create(""); // FIXME: Create TouchEvent
+    } else if (interface.matches("uievent") || interface.matches("uievents")) {
```
Lubrsi, line 671: Ditto

Userland/Libraries/LibWeb/DOM/Document.cpp
```diff
@@ -628,6 +630,60 @@ NonnullRefPtr<Range> Document::create_range()
     return Range::create(*this);
 }
 
+// https://dom.spec.whatwg.org/#dom-document-createevent
+NonnullRefPtr<Event> Document::create_event(const String& interface)
```
Lubrsi, line 634: `interface` should be made all lowercase, as it needs to be a case-insensitive match.

linusg: `matches()` defaults to `CaseSensitivity::CaseInsensitive`

Lubrsi: Oh, didn't know that :)
I just realised is_one_of is case sensitive so it's probably better to leave it all as matches then.

linusg: now that I think about it, matches is probably the wrong method to use here as it's really intended for globs. `equals_ignoring_case()` would do the trick but just uses `to_lowercase` under the hood, so we'd call that a bunch - so we're back to your initial suggestion :smile:

Lubrsi:
With this, web platform tests can now finish setting up and gives some output:
Screenshot from 2021-04-10 18-25-53

Userland/Libraries/LibWeb/DOM/Document.cpp
linusg: Now implemented using explicit `to_lowecase` and `is_one_of` where applicable. Thanks for the suggestions!

Userland/Libraries/LibWeb/DOM/Document.cpp
```diff
@@ -628,6 +630,61 @@ NonnullRefPtr<Range> Document::create_range()
...
+{
+    auto interface_lowecase = interface.to_lowercase();
```
Lubrsi, line 636: Typo:
```suggestion
    auto interface_lowercase = interface.to_lowercase();
```

Userland/Libraries/LibWeb/DOM/Document.cpp
```diff
@@ -628,6 +630,61 @@ NonnullRefPtr<Range> Document::create_range()
...
+        event = Event::create(""); // FIXME: Create BeforeUnloadEvent
+    } else if (interface_lowecase == "compositionevent") {
+        event = Event::create(""); // FIXME: Create CompositionEvent
+    } else if (interface_lowecase == "customevent") {
+        event = Event::create(""); // FIXME: Create CustomEvent
+    } else if (interface_lowecase == "devicemotionevent") {
+        event = Event::create(""); // FIXME: Create DeviceMotionEvent
+    } else if (interface_lowecase == "deviceorientationevent") {
+        event = Event::create(""); // FIXME: Create DeviceOrientationEvent
+    } else if (interface.equals_ignoring_case("dragevent")) {
```
Lubrsi, line 648: Accidentally forgot to use the new `interface_lowercase`:
```suggestion
    } else if (interface_lowercase == "dragevent")) {
```

Userland/Libraries/LibWeb/DOM/Document.cpp
```diff
@@ -628,6 +630,61 @@ NonnullRefPtr<Range> Document::create_range()
...
+        event = Event::create(""); // FIXME: Create DragEvent
+    } else if (interface.is_one_of("event", "events")) {
```
Lubrsi, line 650: Ditto:
```suggestion
    } else if (interface_lowercase.is_one_of("event", "events")) {
```

Userland/Libraries/LibWeb/DOM/Document.cpp
```diff
@@ -628,6 +630,61 @@ NonnullRefPtr<Range> Document::create_range()
...
+        event = Event::create(""); // FIXME: Create FocusEvent
+    } else if (interface_lowecase == "hashchangeevent") {
+        event = Event::create(""); // FIXME: Create HashChangeEvent
+    } else if (interface_lowecase == "htmlevents") {
+        event = Event::create("");
+    } else if (interface_lowecase == "keyboardevent") {
+        event = Event::create(""); // FIXME: Create KeyboardEvent
+    } else if (interface_lowecase == "messageevent") {
+        event = Event::create(""); // FIXME: Create MessageEvent
+    } else if (interface.is_one_of("mouseevent", "mouseevents")) {
```
Lubrsi, line 662: Ditto:
```suggestion
    } else if (interface_lowercase.is_one_of("mouseevent", "mouseevents")) {
```

Userland/Libraries/LibWeb/DOM/Document.cpp
```diff
@@ -628,6 +630,61 @@ NonnullRefPtr<Range> Document::create_range()
...
+        event = UIEvents::MouseEvent::create("", 0, 0);
+    } else if (interface_lowecase == "storageevent") {
+        event = Event::create(""); // FIXME: Create StorageEvent
+    } else if (interface_lowecase == "svgevents") {
+        event = Event::create("");
+    } else if (interface_lowecase == "textevent") {
+        event = Event::create(""); // FIXME: Create CompositionEvent
+    } else if (interface_lowecase == "touchevent") {
+        event = Event::create(""); // FIXME: Create TouchEvent
+    } else if (interface.is_one_of("uievent", "uievents")) {
```
Lubrsi, line 672: Ditto:
```suggestion
    } else if (interface_lowercase.is_one_of("uievent", "uievents")) {
````

awesomekling:
This looks awesome. :)

PR #6220, opened 2021-04-10
HackStudio: Allow users to drag and drop files into the editor

RickySeverino:
Add drag and drop functionality to the editor. Right now this wont work if there is an empty widget, i.e when HackStudio is first opened, but when the editor is present. An error message will show the user tries to drag and drop multiple files but this can change when HackStudio allows multiple panes to open.


RickySeverino:
oops, looks like I forgot to include MessageBox.h. I'll fix that

PR #6224, opened 2021-04-10
Meta: Add discord notifications for pushes and pull requests

IdanHo:
These are similar to the existing irc notifications.
(the accompanying discord channel + webhook + github secret has to be created as well)


bgianfo:
Since when is there a discord? 😁

IdanHo:
> Since when is there a discord?

see andreas' twitter :)

PR #6225, opened 2021-04-10
UE: More Bits!

Hendiadyoin1:
* Adds support for 128/256 bit reads and writes
* ~Adds support for Packed Data and operations upon them~
* Adds FLD_RM80 and FSTP_RM80 instruction support

Known Issues:
* ~u256 is not constexpr'nable, at least not in a neat way due to u128 not being "trivially constructible"~ u128 and u256 are not trivially contructable
* narrowing conversions from u128/u256 down are not marked by the compiler
* ~the name of the Packed data containers is not that good (is it even in the right place?)~ deleted for now
* We could do two reads/writes for f80 instead of one too big one

Side Note:
* long double's are weird with their supposed size

### DIFF
--- a/Userland/DevTools/UserspaceEmulator/SoftCPU.cpp
+++ b/Userland/DevTools/UserspaceEmulator/SoftCPU.cpp
@@ -1798,7 +1798,18 @@ void SoftCPU::FNDISI(const X86::Instruction&) { TODO_INSN(); }
 void SoftCPU::FNCLEX(const X86::Instruction&) { TODO_INSN(); }
 void SoftCPU::FNINIT(const X86::Instruction&) { TODO_INSN(); }
 void SoftCPU::FNSETPM(const X86::Instruction&) { TODO_INSN(); }
-void SoftCPU::FLD_RM80(const X86::Instruction&) { TODO_INSN(); }
+
+void SoftCPU::FLD_RM80(const X86::Instruction& insn)
+{
+    VERIFY(!insn.modrm().is_register());
+
+    // long doubles can be up to 128 bits wide in memory for reasons (alignment) and only uses 80 bits of precision
+    // gcc uses 12 byte in 32 bit and 16 byte in 64 bit mode
+    // so in the 32 bit case we read a bit to much, but that shouldnt be that bad
+    auto new_f80 = insn.modrm().read128(*this, insn);
+    // FIXME: Respect shadow values
+    fpu_push(*(long double*)new_f80.value().bytes());
+}
 
 void SoftCPU::FUCOMI(const X86::Instruction& insn)
 {
@@ -1835,7 +1846,24 @@ void SoftCPU::FCOMI(const X86::Instruction& insn)
     m_flags_tainted = false;
 }
 
-void SoftCPU::FSTP_RM80(const X86::Instruction&) { TODO_INSN(); }
+void SoftCPU::FSTP_RM80(const X86::Instruction& insn)
+{
+    if (insn.modrm().is_register()) {
+        fpu_set(insn.modrm().register_index(), fpu_pop());
+    } else {
+        // FIXME: Respect shadow values
+        // long doubles can be up to 128 bits wide in memory for reasons (alignment) and only uses 80 bits of precision
+        // gcc uses 12 byte in 32 bit and 16 byte in 64 bit mode
+        // so in the 32 bit case we have to read first, to not override data on the overly big write
+        u128 f80 {};
+        if constexpr (sizeof(long double) == 12)
+            f80 = insn.modrm().read128(*this, insn).value();
+
+        *(long double*)f80.bytes() = fpu_pop();
+
+        insn.modrm().write128(*this, insn, shadow_wrap_as_initialized(f80));
+    }
+}
 
 void SoftCPU::FADD_RM64(const X86::Instruction& insn)
 {
--- a/Userland/DevTools/UserspaceEmulator/SoftCPU.cpp
+++ b/Userland/DevTools/UserspaceEmulator/SoftCPU.cpp
@@ -133,6 +133,25 @@ ValueWithShadow<u64> SoftCPU::read_memory64(X86::LogicalAddress address)
     return value;
 }
 
+ValueWithShadow<u128> SoftCPU::read_memory128(X86::LogicalAddress address)
+{
+    VERIFY(address.selector() == 0x1b || address.selector() == 0x23 || address.selector() == 0x2b);
+    auto value = m_emulator.mmu().read128(address);
+#if MEMORY_DEBUG
+    outln("\033[36;1mread_memory128: @{:04x}:{:08x} -> {:032x} ({:032x})\033[0m", address.selector(), address.offset(), value, value.shadow());
+#endif
+    return value;
+}
+ValueWithShadow<u256> SoftCPU::read_memory256(X86::LogicalAddress address)
+{
+    VERIFY(address.selector() == 0x1b || address.selector() == 0x23 || address.selector() == 0x2b);
+    auto value = m_emulator.mmu().read256(address);
+#if MEMORY_DEBUG
+    outln("\033[36;1mread_memory256: @{:04x}:{:08x} -> {:064x} ({:064x})\033[0m", address.selector(), address.offset(), value, value.shadow());
+#endif
+    return value;
+}
+
 void SoftCPU::write_memory8(X86::LogicalAddress address, ValueWithShadow<u8> value)
 {
     VERIFY(address.selector() == 0x23 || address.selector() == 0x2b);
@@ -159,6 +178,24 @@ void SoftCPU::write_memory64(X86::LogicalAddress address, ValueWithShadow<u64> v
     VERIFY(address.selector() == 0x23 || address.selector() == 0x2b);
     outln_if(MEMORY_DEBUG, "\033[36;1mwrite_memory64: @{:04x}:{:08x} <- {:016x} ({:016x})\033[0m", address.selector(), address.offset(), value, value.shadow());
     m_emulator.mmu().write64(address, value);
+}
+
+void SoftCPU::write_memory128(X86::LogicalAddress address, ValueWithShadow<u128> value)
+{
+    VERIFY(address.selector() == 0x23 || address.selector() == 0x2b);
+#if MEMORY_DEBUG
+    outln("\033[36;1mwrite_memory128: @{:04x}:{:08x} <- {:032x} ({:032x})\033[0m", address.selector(), address.offset(), value, value.shadow());
+#endif
+    m_emulator.mmu().write128(address, value);
+}
+
+void SoftCPU::write_memory256(X86::LogicalAddress address, ValueWithShadow<u256> value)
+{
+    VERIFY(address.selector() == 0x23 || address.selector() == 0x2b);
+#if MEMORY_DEBUG
+    outln("\033[36;1mwrite_memory256: @{:04x}:{:08x} <- {:064x} ({:064x})\033[0m", address.selector(), address.offset(), value, value.shadow());
+#endif
+    m_emulator.mmu().write256(address, value);
 }
 
 void SoftCPU::push_string(const StringView& string)
--- a/Userland/DevTools/UserspaceEmulator/ValueWithShadow.h
+++ b/Userland/DevTools/UserspaceEmulator/ValueWithShadow.h
@@ -4,12 +4,18 @@
  * SPDX-License-Identifier: BSD-2-Clause
  */
 
-#include <AK/Format.h>
-#include <AK/Platform.h>
-
 #pragma once
 
+#include <AK/Format.h>
+#include <AK/Platform.h>
+#include <LibX86/Types.h>
+#include <string.h>
+
 namespace UserspaceEmulator {
+
+constexpr u64 _inititalized_64 = 0x01010101'01010101LLU;
+constexpr u128 _initialized_128 = u128(_inititalized_64, _inititalized_64);
+constexpr u256 _initialized_256 = u256(_initialized_128, _initialized_128);
 
 template<typename T>
 class ValueAndShadowReference;
@@ -32,6 +38,12 @@ public:
 
     bool is_uninitialized() const
     {
+        if constexpr (sizeof(T) == 32)
+            return (m_shadow & _initialized_256) != _initialized_256;
+        if constexpr (sizeof(T) == 16)
+            return (m_shadow & _initialized_128) != _initialized_128;
+        if constexpr (sizeof(T) == 8)
+            return (m_shadow & _inititalized_64) != _inititalized_64;
         if constexpr (sizeof(T) == 4)
             return (m_shadow & 0x01010101) != 0x01010101;
         if constexpr (sizeof(T) == 2)
@@ -42,6 +54,12 @@ public:
 
     void set_initialized()
     {
+        if constexpr (sizeof(T) == 32)
+            m_shadow = _initialized_256;
+        if constexpr (sizeof(T) == 16)
+            m_shadow = _initialized_128;
+        if constexpr (sizeof(T) == 8)
+            m_shadow = _inititalized_64;
         if constexpr (sizeof(T) == 4)
             m_shadow = 0x01010101;
         if constexpr (sizeof(T) == 2)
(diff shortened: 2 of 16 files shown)


bcoles (commented):
Copyright headers on all new files say 2020 (not 2021).

Hendiadyoin1:
We now support printing the containers and a whole lot more actions with them

Hendiadyoin1:
regarding the 4th known issue (to big reads), One Idea would be:
``` cpp
// writing
{       
    [...]
    auto address = insn.modrm().resolve(this,insn);
    u8 data[sizeof(long double)];
    *(long double*)data = fpu_pop();
    
    write_memory64(address, {data, 0x01010101'01010101ULL});
    if constexpr (sizeof(long double) == 12)
        write_memory32(address, {*(u32*)&data[8], 0x01010101});
    else if constexpr (sizeof(long double) == 16)
        write_memory64(address, {*(u64*)&data[8],0x01010101'01010101ULL});
    else VERIFY_NOT_REACHED();
}
//reading
{
    auto address = insn.modrm().resolve(this,insn);

    u8 data[sizeof(long double)];
    *(u64*)data =  read_memory64(address).value();
    if constexpr (sizeof(long double) == 12)
        *(u32*)&data[8] =  read_memory32(address).value();
    else if constexpr (sizeof(long double) == 16)
        *(u64*)&data[8] =  read_memory64(address).value();
    else VERIFY_NOT_REACHED();

    fpu_push(*(long double*)data);
}
```

Userland/DevTools/UserspaceEmulator/SoftCPU.cpp
```diff
@@ -1834,7 +1834,18 @@ void SoftCPU::FNDISI(const X86::Instruction&) { TODO_INSN(); }
 void SoftCPU::FNCLEX(const X86::Instruction&) { TODO_INSN(); }
 void SoftCPU::FNINIT(const X86::Instruction&) { TODO_INSN(); }
 void SoftCPU::FNSETPM(const X86::Instruction&) { TODO_INSN(); }
-void SoftCPU::FLD_RM80(const X86::Instruction&) { TODO_INSN(); }
+
+void SoftCPU::FLD_RM80(const X86::Instruction& insn)
+{
+    VERIFY(!insn.modrm().is_register());
+
+    // long doubles can be up to 128 bits wide in memmory for reasons (alignment) and only uses 80 bits of precision
```
awesomekling, line 1842: memmory -> memory

Userland/DevTools/UserspaceEmulator/SoftCPU.cpp
```diff
@@ -1834,7 +1834,18 @@ void SoftCPU::FNDISI(const X86::Instruction&) { TODO_INSN(); }
...
+    // bit_cast seems to see the 12 bytes, but godbolt says 16 bytes in x86-64 or 12 with -m32
```
awesomekling, line 1843: godbolt is a website. Are you saying that a compiler is telling you something? Please be more clear.

Userland/DevTools/UserspaceEmulator/SoftCPU.cpp
```diff
@@ -1834,7 +1834,18 @@ void SoftCPU::FNDISI(const X86::Instruction&) { TODO_INSN(); }
...
+    // and writing a 12 byte read just for this would be ineficient
```
awesomekling, line 1844: ineficient -> inefficient
Also, why would it be inefficient?

Userland/DevTools/UserspaceEmulator/SoftCPU.cpp
```diff
@@ -1871,7 +1882,22 @@ void SoftCPU::FCOMI(const X86::Instruction& insn)
...
-void SoftCPU::FSTP_RM80(const X86::Instruction&) { TODO_INSN(); }
+void SoftCPU::FSTP_RM80(const X86::Instruction& insn)
+{
+    if (insn.modrm().is_register()) {
+        fpu_set(insn.modrm().register_index(), fpu_pop());
+    } else {
+        // FIXME: Respect shadow values
+        // long doubles can be up to 128 bits wide in memmory for reasons (alignment) and only uses 80 bits of precision
+        // bit_cast seems to see the 12 bytes, but godbolt says 16 bytes in x86-64 or 12 with -m32
+        // and writing a 12 byte write just for this would be ineficient
```
awesomekling, line 1893: Same comments as above.

Userland/DevTools/UserspaceEmulator/SoftCPU.cpp
```diff
@@ -1871,7 +1882,22 @@ void SoftCPU::FCOMI(const X86::Instruction& insn)
...
+        // and for that reason we need to read first to preserve the other few bytes,
+        // although it taints the memmory graph
```
awesomekling, line 1895: memmory -> memory

AK/BitCast.h
```diff
@@ -34,7 +40,7 @@ inline T bit_cast(const U& a)
     static_assert(sizeof(T) == sizeof(U));
 
     T result;
-    __builtin_memcpy(&result, &a, sizeof(T));
+    memcpy((void*)&result, &a, sizeof(T));
```
awesomekling, line 43: Why is this necessary? Does it change the generated code at all?

awesomekling (changes requested):
What is IdentMap? What is "Ident" in the X86 context?
Please explain what you are doing :)

Userland/Libraries/LibX86/Types/Formatter.cpp
```diff
@@ -0,0 +1,133 @@
...
+    if (value.high() == 0) {
+        AK::Formatter<u64> formatter { *this };
+        return formatter.format(builder, value.low());
+    }
+
+    if (m_precision.has_value())
+        VERIFY_NOT_REACHED();
+
+    if (m_mode == Mode::Pointer) {
+        // this is way to beig for a pointer
```
bcoles, line 47: big?

```suggestion
        // this is way to big for a pointer
```

Userland/Libraries/LibX86/Types/Formatter.cpp
```diff
@@ -0,0 +1,133 @@
...
+    if (m_mode == Mode::Binary) {
+        base = 2;
+    } else if (m_mode == Mode::BinaryUppercase) {
+        base = 2;
+        upper_case = true;
+    } else if (m_mode == Mode::Octal) {
+        base = 8;
+    } else if (m_mode == Mode::Decimal || m_mode == Mode::Default) {
+        // FIXME: implement this
+        VERIFY_NOT_REACHED();
```
bcoles, line 62:
```suggestion
        TODO();
```

Userland/Libraries/LibX86/Types/Formatter.cpp
```diff
@@ -0,0 +1,133 @@
...
+    if (value.high() == 0) {
+        AK::Formatter<u128> formatter { *this };
+        return formatter.format(builder, value.low());
+    }
+
+    if (m_precision.has_value())
+        VERIFY_NOT_REACHED();
+
+    if (m_mode == Mode::Pointer) {
+        // this is way to beig for a pointer
```
bcoles, line 92:
```suggestion
        // this is way to big for a pointer
```

Userland/Libraries/LibX86/Types/Formatter.cpp
```diff
@@ -0,0 +1,133 @@
...
+    if (m_mode == Mode::Binary) {
+        base = 2;
+    } else if (m_mode == Mode::BinaryUppercase) {
+        base = 2;
+        upper_case = true;
+    } else if (m_mode == Mode::Octal) {
+        base = 8;
+    } else if (m_mode == Mode::Decimal || m_mode == Mode::Default) {
+        // FIXME: implement this
+        VERIFY_NOT_REACHED();
```
bcoles, line 107:
```suggestion
        TODO();
```

Userland/Libraries/LibX86/Types/IdentView.h
```diff
@@ -0,0 +1,154 @@
...
+    template<Integral R, Integral U>                                       \
+    ALWAYS_INLINE IdentMap<R> operator op(const IdentView<U>& other) const \
+    {                                                                      \
+        VERIFY(m_size == other.size());                                    \
+        IdentMap<R> ret(m_size);                                           \
+        for (off_t i = 0; i < m_size; ++i) {                               \
+            ret[i] = m_values[i] op other[i];                              \
+        }                                                                  \
+        return ret;                                                        \
+    }                                                                      \
```
bcoles, line 64: Both `IdentView.h` and `IdentMap.h` are very similar and `#include` each other.

These highlighted functions are identical in both files. Are these supposed to be `ALWAYS_INLINE IdentView<R> ...` rather than `ALWAYS_INLINE IdentMap<R> ...` ?

Userland/Libraries/LibX86/Types/u128.h
```diff
@@ -0,0 +1,297 @@
...
+    ALWAYS_INLINE u128 operator|(const u128& other) const
+    {
+        return { low() | other.low(), high() | other.high() };
+    }
+    ALWAYS_INLINE u128 operator^(const u128& other) const
+    {
+        return { low() ^ other.low(), high() ^ other.high() };
+    }
+
+    // bitwise asign
```
bcoles, line 220:
```suggestion
    // bitwise assign
```

Userland/Libraries/LibX86/Types/u256.h
```diff
@@ -0,0 +1,299 @@
...
+    ALWAYS_INLINE u256 operator|(const u256& other) const
+    {
+        return { low() | other.low(), high() | other.high() };
+    }
+    ALWAYS_INLINE u256 operator^(const u256& other) const
+    {
+        return { low() ^ other.low(), high() ^ other.high() };
+    }
+
+    // bitwise asign
```
bcoles, line 228:
```suggestion
    // bitwise assign
```

Userland/Libraries/LibX86/Types/IdentView.h
Hendiadyoin1: In this  its a map, because it uses its own copy of data, instead of acting upon data provided to it

AK/BitCast.h
Hendiadyoin1: haven't checked that, but I just happened  to stumble on that while trying to use bit_cast's for long double conversion and I thought the "zen" was to try write every method on our own, and not rely on compilers and their libraries if possible

the (void*) cast is useless though

Userland/DevTools/UserspaceEmulator/SoftCPU.cpp
Hendiadyoin1: code wise  
it would be another function that would have to be implemented across multiple classes like the other reads/writes  
in a previous comment I have described a way to do it in 2 writes dependent on long double size
In the newer version its described and handled better

Userland/DevTools/UserspaceEmulator/SoftCPU.cpp
Hendiadyoin1: Changed it to refer to gcc's behavior

Hendiadyoin1:
> 
> 
> What is IdentMap? What is "Ident" in the X86 context?
> Please explain what you are doing :)


So these are supposed to be views on packed data, with which you can perform the same arythmatic across all values at once.
I thought it may be a good way to implement the structure backing ymm and xmm vector registers

Userland/Libraries/LibX86/Types/u256.h
Hendiadyoin1: should be fixed

Userland/Libraries/LibX86/Types/u128.h
Hendiadyoin1: should be fixed

Userland/Libraries/LibX86/Types/Formatter.cpp
Hendiadyoin1: should be fixed

Userland/Libraries/LibX86/Types/Formatter.cpp
Hendiadyoin1: should be fixed

Userland/Libraries/LibX86/Types/Formatter.cpp
Hendiadyoin1: should be fixed

Userland/Libraries/LibX86/Types/Formatter.cpp
Hendiadyoin1: should be fixed

Userland/DevTools/UserspaceEmulator/SoftCPU.cpp
Hendiadyoin1: should be fixed

Userland/DevTools/UserspaceEmulator/SoftCPU.cpp
Hendiadyoin1: same answer as above.

Userland/DevTools/UserspaceEmulator/SoftCPU.cpp
Hendiadyoin1: should be fixed

AK/BitCast.h
Hendiadyoin1: And looking at [this](https://godbolt.org/z/9jqd5GfWc) it seems that buildins are better due to being optimized better to just to movs instead of a rep movsb and allways inlined, maybe we can apply some of these optimizations too in out libC implementation

Userland/Libraries/LibX86/Types/IdentView.h
Hendiadyoin1: And when doing operations we do not want to alter the data directly so its a Map there

AK/BitCast.h
Hendiadyoin1: guess I'll drop that change for now

Userland/DevTools/UserspaceEmulator/ValueWithShadow.h
```diff
@@ -24,11 +24,16 @@
...
 #include <AK/Format.h>
 #include <AK/Platform.h>
+#include <LibX86/Types.h>
+#include <string.h>
 
-#pragma once
-
+#define INITIALIZED_64 0x01010101'01010101LLU
+#define INITIALIZED_128 u128(INITIALIZED_64, INITIALIZED_64)
+#define INITIALIZED_256 u256(INITIALIZED_128, INITIALIZED_128)
```
awesomekling, line 36: Can we make these `constexpr` values instead?

Userland/DevTools/UserspaceEmulator/SoftCPU.cpp
```diff
@@ -1871,7 +1882,23 @@ void SoftCPU::FCOMI(const X86::Instruction& insn)
...
+void SoftCPU::FSTP_RM80(const X86::Instruction& insn)
+{
+    if (insn.modrm().is_register()) {
+        fpu_set(insn.modrm().register_index(), fpu_pop());
+    } else {
+        // FIXME: Respect shadow values
+        // long doubles can be up to 128 bits wide in memory for reasons (alignment) and only uses 80 bits of precision
+        // gcc uses 12 byte in 32 bit and 16 byte in 64 bit mode
+        // so in the 32 bit case we have to read first, to not override data on the overly big write
+        u128 f80();
```
awesomekling, line 1894: This is a very strange declaration. please use brace initialization instead, i.e `u128 f80 {};`

awesomekling (changes requested):
> > What is IdentMap? What is "Ident" in the X86 context?
> > Please explain what you are doing :)
> 
> So these are supposed to be views on packed data, with which you can perform the same arythmatic across all values at once.
> I thought it may be a good way to implement the structure backing ymm and xmm vector registers
> 
> I know the name is bad how about "DataVector[...]"

Since you are not using them for anything here, it's just dead code right now, so I think it's better to delay introducing them until they're actually needed for something.

Userland/DevTools/UserspaceEmulator/SoftCPU.cpp
Hendiadyoin1: I thought i changed that, but that may explain the build error

Userland/DevTools/UserspaceEmulator/ValueWithShadow.h
Hendiadyoin1: the byte access in in u256 would be a cast of `this` to the corresponding type, but yes

this will be fixed in the next commit iteration

Hendiadyoin1:
> 
> 
> > > What is IdentMap? What is "Ident" in the X86 context?
> > > Please explain what you are doing :)
> > 
> > 
> > So these are supposed to be views on packed data, with which you can perform the same arythmatic across all values at once.
> > I thought it may be a good way to implement the structure backing ymm and xmm vector registers
> > I know the name is bad how about "DataVector[...]"
> 
> Since you are not using them for anything here, it's just dead code right now, so I think it's better to delay introducing them until they're actually needed for something.

Then I'll stash them away for now

Hendiadyoin1:
Was that everything?

PR #6227, opened 2021-04-10
Ports: Begin porting ports tree to serenity

Petelliott:
This PR adds the `install-ports` make target, which copies the ports tree into serenity. I also fixed some command line utilities so that `./port_include.sh` would run on serenity. I still havn't built any ports on serenity due to the file-system running out of space (even though system monitor shows plenty of space).


Ports/.port_include.sh
```diff
@@ -68,6 +68,13 @@ fetch() {
         IFS=$OLDIFS
         read url filename auth_sum<<< $(echo "$f")
         echo "URL: ${url}"
+
+        # FIXME: Serenity's curl port does not support https, even with openssl installed.
```
alimpfard, line 72: Failing this, you _could_ try using `pro`

Base/usr/bin/env
```diff
@@ -0,0 +1 @@
+/bin/env
```
bcoles, line 1: This should probably be a symlink to `/bin/env` rather than a shell script (without a shebang?)

```diff
$ gitd Meta/build-root-filesystem.sh 
diff --git a/Meta/build-root-filesystem.sh b/Meta/build-root-filesystem.sh
index 71785bb51..1814ca92c 100755
--- a/Meta/build-root-filesystem.sh
+++ b/Meta/build-root-filesystem.sh
@@ -117,6 +117,7 @@ echo "done"
 printf "installing shortcuts... "
 ln -s Shell mnt/bin/sh
 ln -s test mnt/bin/[
+ln -s /bin/env mnt/usr/bin/env
 echo "done"
 
 printf "installing 'checksum' variants... "
```

Petelliott: This actually is a symlink. symlinks are just rendered weirdly on github.

Userland/Utilities/env.cpp
```diff
@@ -36,40 +37,57 @@ int main(int argc, char** argv)
...
-                continue;
-            }
-        }
-        if (StringView { argv[idx] }.contains('=')) {
-            putenv(argv[idx]);
+    Core::ArgsParser args_parser;
+    args_parser.add_option(ignore_env, "Start with an empty environment", "ignore-environment", 'i');
+    args_parser.add_option(split_string, "process and split S into separate arguments; used to pass multiple arguments on shebang lines", "split-string", 'S', "S");
+
+    args_parser.add_positional_argument(values, "environment and commands", "env/command", Core::ArgsParser::Required::No);
```
bcoles, line 48: Capitalization:

```suggestion
    args_parser.add_option(ignore_env, "Start with an empty environment", "ignore-environment", 'i');
    args_parser.add_option(split_string, "Process and split S into separate arguments; used to pass multiple arguments on shebang lines", "split-string", 'S', "S");
    args_parser.add_positional_argument(values, "Environment and commands", "env/command", Core::ArgsParser::Required::No);
```

Ports/.port_include.sh
```diff
@@ -122,16 +134,20 @@ fetch() {
...
+            fi
         else
-            echo "- Signature check NOT OK"
-            for f in $files; do
-                rm -f $f
-            done
-            rm -rf "$workdir"
-            echo "  Signature mismatching, removed erronous download. Please run script again."
-            exit 1
+            echo "gpg not installed; skipping signature validation."
```
bcoles, line 150: oof

Petelliott: Whats the complaint here? I think it's reasonable to not check signatures if the user does not have `gpg`.

bcoles: This change effects ports both inside Serenity and on the build host.

This change has security implications and isn't mentioned anywhere in the PR description or commits.

For all other other auth types (md5, sha1, etc) a failed signature match is considered a fatal error and the installation is abandoned.

Ports/.port_include.sh
Petelliott: pro doesn't have the ciphers in common with ftp.gnu.org, so it's not a great solution either. I've added it as a fallback when curl isn't installed though.

Ports/.port_include.sh
bcoles: I'm not sure the best approach here if we want to install ports from inside Serenity. Serenity doesn't have a `gpg` port yet. And even if it did, it would likely be a cyclic dependency requiring gpg to verify the key to install gpg.

When building ports on the build host, failed signature validation due to missing `gpg` should be fatal.

When building ports on the Serenity host, print a warning due to missing `gpg`. Although I'm generally not keen on undermining package validation on an adhoc basis with a single line warning.

Petelliott: what do you sugguest I do about this. I can remove this, or I can add a `noverify` flag to `./package.sh` so it can be used in serenity with some caution.

bcoles: > 
> 
> what do you suggest I do about this? I can remove this, or I can add a `noverify` flag to `./package.sh` so it can be used in serenity with some caution.

Something akin to an optional `noverify` flag would likely be the best long term solution.

Ports/.port_include.sh
alimpfard: shame.
Eventually it will have _some_ modern ciphersuite...eventually.

Ports/.port_include.sh
Petelliott: i've added the `--no-gpg-verification` flag

PR #6231, opened 2021-04-11
Shell: add `type` builtin

g-w1:
This is my first pr, so please tell me if something is non-idomatic or wrong :^).
image
In bash, type on a function shows the source (I quite like this feature), but I couldn't figure out how to get source of a function from a `Node` so I left it as a fixme. If anyone knows, please tell me and I'll add it!

### DIFF
--- a/Userland/Shell/Builtin.cpp
+++ b/Userland/Shell/Builtin.cpp
@@ -1010,6 +1010,8 @@ bool Shell::run_builtin(const AST::Command& command, const NonnullRefPtrVector<A
         retval = builtin_##builtin(argv.size() - 1, argv.data());        \
         if (!has_error(ShellError::None))                                \
             raise_error(m_error, m_error_description, command.position); \
+        fflush(stdout);                                                  \
+        fflush(stderr);                                                  \
         return true;                                                     \
     }
 
--- a/Userland/Shell/Builtin.cpp
+++ b/Userland/Shell/Builtin.cpp
@@ -26,6 +26,7 @@
 
 #include "AST.h"
 #include "Shell.h"
+#include "Shell/Formatter.h"
 #include <AK/LexicalPath.h>
 #include <AK/ScopeGuard.h>
 #include <LibCore/ArgsParser.h>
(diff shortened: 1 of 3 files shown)


alimpfard:
You can't get the original source, but you can make a Formatter and pretty print the source from the AST.

Userland/Shell/Builtin.cpp
```diff
@@ -127,6 +127,57 @@ int Shell::builtin_bg(int argc, const char** argv)
...
+        if (found)
+            continue;
+
+        // check if its an executable in PATH
+        auto fullpath = Core::find_executable_in_path(command);
+        if (!fullpath.is_null()) {
+            printf("%s is %s\n", command, escape_token(fullpath).characters());
+            continue;
+        }
+        printf("type: %s not found\n", command);
```
alimpfard, line 175: Should this case make it exit with nonzero?

Userland/Shell/Builtin.cpp
```diff
@@ -127,6 +127,57 @@ int Shell::builtin_bg(int argc, const char** argv)
...
+    parser.set_general_help("Display information about commands.");
+    parser.add_positional_argument(commands, "Command(s) to list info about", "command");
+
+    if (!parser.parse(argc, const_cast<char**>(argv), false))
+        return 1;
+
+    for (auto& command : commands) {
+        // check if it is an alias
+        auto alias = m_aliases.get(command);
+        if (alias.has_value()) {
```
alimpfard, line 145: Just a suggestion, you could instead use 
```c++
if (auto alias = m_aliases.get(command); alias.has_value())
    ...
```

(This one and the one below)

Userland/Shell/Builtin.cpp
```diff
@@ -127,6 +127,57 @@ int Shell::builtin_bg(int argc, const char** argv)
...
+        // check if its a builtin
+        bool found = false;
+        for (auto& builtin_name : builtin_names) {
+            if (strcmp(builtin_name, command) == 0) {
+                printf("%s is a shell builtin\n", builtin_name);
+                found = true;
+            }
+        }
+        if (found)
+            continue;
```
alimpfard, line 167:
```c++
if (has_builtin(command)) {
    ....
    continue;
}
```

alimpfard (commented):
Also please add some tests : )

alimpfard:
Also also, please property capitalise the commit message (e.g. "Shell: Add a `type` builtin")

Userland/Shell/Builtin.cpp
```diff
@@ -127,6 +128,72 @@ int Shell::builtin_bg(int argc, const char** argv)
...
+
+        // check if it is a function
+        if (auto function = m_functions.get(command); function.has_value()) {
+            auto fn = function.value();
+            printf("%s is a function\n", command);
+            if (!dont_show_function_source) {
+                StringBuilder builder;
+                builder.append(fn.name);
+                builder.append("(");
+                for (int i = 0; i < (int)fn.arguments.size(); i++) {
```
alimpfard, line 162: `size_t i = 0` and avoid all the casts?

Userland/Shell/Tests/builtin-test.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/sh
+
+source $(dirname "$0")/test-commons.inc
+
+
+if not [ "$(type ls)" = "ls is /bin/ls" ] { fail "'type' on a binary not working" }
```
alimpfard, line 6: This might not be the case on linux (we run the tests there), try `"ls is $(which ls)"` instead?

g-w1: Its still failing, any clue why?

alimpfard: hmm...no idea, just run it again in the if body and let's see what the output is?

g-w1: Even better, is there a way to build the Shell on linux?

alimpfard: Sure, just pass `-DBUILD_LAGOM=1` to your cmake invocation, and rebuild.
the shell is at `Build/$ARCH/Meta/Lagom/shell`

alimpfard: https://github.com/alimpfard/serenity/commit/54ebae6261106b502bd8cdd7c3c707c1ef06deb9 should fix the capturing-builtin-output issue. you could either cherry-pick that commit, ignore the issue in this PR, or commit it yourself - whichever you're more comfortable with is fine.

Userland/Shell/Builtin.cpp
```diff
@@ -127,6 +128,72 @@ int Shell::builtin_bg(int argc, const char** argv)
...
+                StringBuilder builder;
+                builder.append(fn.name);
+                builder.append("(");
+                for (size_t i = 0; i < fn.arguments.size(); i++) {
+                    builder.append(fn.arguments[i]);
+                    if (!(i == fn.arguments.size() - 1))
+                        builder.append(" ");
+                }
+                builder.append(") {\n");
+                auto formatter = Formatter(*fn.body);
```
alimpfard, line 168: Minor issue: `fn.body` can be null (in case of a function with an empty body), could you add a check for that?

Userland/Shell/Tests/builtin-test.sh
```diff
@@ -0,0 +1,24 @@
...
+source $(dirname "$0")/test-commons.inc
+
+
+if not [ "$(type ls)" = "ls is $(which ls)" ] { fail "'type' on a binary not working" }
+
+if not [ "$(type pwd)" = "pwd is a shell builtin" ] { fail "'type' on a builtin not working" }
+
+f() { ls }
+
+if not [ "$(type f)" = "f is a function\nf() {\nls\n}" ] { fail "'type' on a function not working" }
```
g-w1, line 12: Now this is failing, I think it must be a bug in `[` or something else :thinking:.

Userland/Shell/Builtin.cpp
```diff
@@ -939,6 +1010,8 @@ bool Shell::run_builtin(const AST::Command& command, const NonnullRefPtrVector<A
         retval = builtin_##builtin(argv.size() - 1, argv.data());        \
         if (!has_error(ShellError::None))                                \
             raise_error(m_error, m_error_description, command.position); \
+        fflush(stdout);                                                  \
+        fflush(stderr);                                                  \
```
alimpfard, line 1014: I wish you had put these in a separate commit...or at least mentioned them in the description of this commit :P

Userland/Shell/Builtin.cpp
```diff
@@ -127,6 +128,76 @@ int Shell::builtin_bg(int argc, const char** argv)
...
+int Shell::builtin_type(int argc, const char** argv)
+{
+
+    Vector<const char*> commands;
+    bool dont_show_function_source = false;
+
+    Core::ArgsParser parser;
+    parser.set_general_help("Display information about commands.");
+    parser.add_positional_argument(commands, "Command(s) to list info about", "command");
+    parser.add_option(dont_show_function_source, "Do now show functions source.", "no-fn-source", 'f');
```
rajiv-k, line 140: s/now/not

alimpfard (approved):

PR #6232, opened 2021-04-11
Toolchain: Update to binutils-2.36.1 and gcc-10.3.0

bgianfo:
gcc-10.3.0 is a bugfix release, with 178 bugs fixed.
See: https://gcc.gnu.org/bugzilla/buglist.cgi?bug_status=RESOLVED&resolution=FIXED&target_milestone=10.3

binutils-2.36.1 is also a minor release with only a few real changes.
See: https://fossies.org/linux/binutils/binutils/ChangeLog

The binutils release had some changes which shuffled code around,
so I had to fixup our patch to apply cleanly again.


bgianfo:
Looking into the weird initialize_list failure...

bgianfo:
Ok, this is building cleanly... I left the gcc port at 10.2.0 as it will likely be more work to rebase the patches, etc to 10.3.0

bgianfo:
@awesomekling thoughts?

awesomekling:
Let's do it. Great work :)

PR #6233, opened 2021-04-11
Fix a few more ports

gunnarbeutner:
This PR fixes:

* frotz: The include directory for ncurses was incorrect which caused the port to fail to build when ncurses wasn't installed on the host system.
* dmidecode: Reinstalling the port failed because `ln -s` doesn't overwrite existing symlinks.
* dropbear: Disabled static build because `gcc -static` is broken.
* nesalizer: Fixed building the port when SDL2 wasn't installed on the host system.
* libexpat: Made sure the `main` symbol has default visibility.
* flatbuffers: `PATH_MAX` (or rather, anything in SerenityOS' `<limits.h>`) wasn't available because GCC installed its own "fixed" version of the `<limits.h>` header. This change requires a rebuild of the toolchain.
* libiconv/diffutils/nano: Fixed build errors for `rpl_stat`. These ports provide their own overrides for `<sys/stat.h>` and `<fcntl.h>` and didn't expect the system's `<sys/stat.h>` to include `<fcntl.h>`.
* diffutils: `getopt` and `getopt_long` shouldn't modify `argv`.
* diffutils: Commented out an unused function in gnulib (`getdtablesize`) which allows the port to build properly now.
* git: This wasn't broken initially, but now that `libiconv` is available `configure` failed when checking some `libiconf` stuff.


awesomekling:
Looks great :)

My one note is: you keep mentioning SUSv2 which is a very old document.
It doesn't look like anything has changed in later versions, but please always use the latest POSIX specification (SUSv4-2018 at the time of writing.)

gunnarbeutner:
Aye, good point. SUSv2 just so happened to be the first thing I found on Google that was "not Linux". Will keep that in mind for the future. :+1:

PR #6235, opened 2021-04-11
Ports: Fix Ports/README.md to match lint rule

bgianfo:
Looks like https://github.com/SerenityOS/serenity/commit/251018676f47fb34944b1ac195a084573e960656 broke the build?


awesomekling:
D'oh!

PR #6240, opened 2021-04-11
Browser+LibWeb+WebContent: Intial support for cookies

trflynn89:
This adds a (currently ephemeral) backend storage for cookies set via `document.cookie` or the `Set-Cookie` HTTP response header. We can parse the name-value pair from cookies, but are currently ignoring attributes.


awesomekling:
Very sensible design 👍🏻

PR #6244, opened 2021-04-11
LibJS: Removed a fixme in a test of BigInt.prototype.valueOf

tuqqu:


linusg (changes requested):
LGTM but please update the commit message, this was a FIXME in a test for that function, not the function itself :)

tuqqu:
@linusg that's true, changed it :)

linusg (approved):

PR #6247, opened 2021-04-11
HackStudio: Allow users to drag and drop files into the editor

ry-sev:
This drag and drop feature will not currently work with the placeholder widget, i.e when HackStudio first launches. There will be an error message box if the user attempts to drop multiple files at once. I suspect this will change when HackStudio allows multiple editors to be open at once in the form of separate panes.


itamar8910 (approved):
LGTM.
Also just so you know HackStudio does have support for opening multiple editors panes (view->"Add new editor") but we do not have to support this with drag and drop right now.

PR #6248, opened 2021-04-11
LibJS: Array.from mapFn fixes + thisArg support

tuqqu:
This MR addresses several issues of `Array.from`:

* Callback `mapFn` now properly supports its second argument, which is the current iteration index.
* ~Global object is passed as `this` value in `vm.call`, instead of the array-like value, which was incorrect.~
* Support of the third argument `thisArg` to be passed as `this` in `vm.call`

Also added tests for all cases.
~With this MR merged it looks like `Array.from` can be considered fully implemented in LibJS.~

### DIFF
--- a/Userland/Libraries/LibJS/Runtime/ArrayConstructor.cpp
+++ b/Userland/Libraries/LibJS/Runtime/ArrayConstructor.cpp
@@ -105,6 +105,8 @@ JS_DEFINE_NATIVE_FUNCTION(ArrayConstructor::from)
         map_fn = &callback.as_function();
     }
 
+    auto this_arg = vm.argument(2);
+
     // Array.from() lets you create Arrays from:
     if (auto size = object->indexed_properties().array_like_size()) {
         // * array-like objects (objects with a length property and indexed elements)
@@ -116,7 +118,7 @@ JS_DEFINE_NATIVE_FUNCTION(ArrayConstructor::from)
                 if (vm.exception())
                     return {};
 
-                auto map_fn_result = vm.call(*map_fn, value, element);
+                auto map_fn_result = vm.call(*map_fn, this_arg, element, Value((i32)i));
                 if (vm.exception())
                     return {};
 
@@ -130,12 +132,14 @@ JS_DEFINE_NATIVE_FUNCTION(ArrayConstructor::from)
         array->set_indexed_property_elements(move(elements));
     } else {
         // * iterable objects
+        i32 i = 0;
         get_iterator_values(global_object, value, [&](Value element) {
             if (vm.exception())
                 return IterationDecision::Break;
 
             if (map_fn) {
-                auto map_fn_result = vm.call(*map_fn, value, element);
+                auto map_fn_result = vm.call(*map_fn, this_arg, element, Value(i));
+                i++;
                 if (vm.exception())
                     return IterationDecision::Break;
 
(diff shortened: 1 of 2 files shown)


Userland/Libraries/LibJS/Runtime/ArrayConstructor.cpp
```diff
@@ -105,6 +105,12 @@ JS_DEFINE_NATIVE_FUNCTION(ArrayConstructor::from)
         map_fn = &callback.as_function();
     }
 
+    // thisArg argument
+    auto this_value = vm.argument(2);
+    if (this_value.is_undefined()) {
+        this_value = global_object.value_of();
+    }
```
linusg, line 112: This is not part of the spec... they just pass whatever the third argument is directly to the callback.

Also, `this_value` is usually reserved for the actual this value within the function (`vm.this_value()`), so let's call it `this_arg` just like the spec.

```suggestion
    auto this_arg = vm.argument(2);
```

linusg (changes requested):

linusg:
> With this MR merged it looks like Array.from can be considered fully implemented in LibJS.

Not yet :)

For example, this part is missing:

> The from function is an intentionally generic factory method; it does not require that its this value be the Array constructor. Therefore it can be transferred to or inherited by any other constructors that may be called with a single numeric argument.

Or, as code:

```js
class MyArray extends Array {}
MyArray.from([]) instanceof MyArray
// false, should be true. https://tc39.es/ecma262/#sec-array.from, step 9 and 10.
```

Userland/Libraries/LibJS/Runtime/ArrayConstructor.cpp
tuqqu: tbh I don't think that's correct...
If `thisArg` is not passed or passed as `undefined` the most reasonable choice is to pass `global_object` to `vm.call`.

With your suggestion, we would always give `vm.call` `undefined` should `thisArg` be omitted. And this will trigger `TypeError: ToObject on null or undefined` each time function tries to address `this`

Consider these examples:

should give [ NaN, NaN, NaN ], but will be TypeError
```
Array.from([1, 2, 3], function (n, i) { return n + this.e });   
```

should give [ 101, 102, 103 ], but will be TypeError
```
e = 100;
Array.from([1, 2, 3], function (n, i) { return n + this.e });
```

should give [ 101, 102, 103 ], but will be TypeError
```
e = 100;
Array.from([1, 2, 3], function (n, i) { return n + this.e }, undefined);
```

linusg: > tbh I don't think that's correct...

It is, see spec. :)

>  If thisArg is not passed or passed as undefined the most reasonable choice is to pass global_object to vm.call.

No.

> With your suggestion, we would always give vm.call undefined should thisArg be omitted.

Yes.

> And this will trigger TypeError: ToObject on null or undefined each time function tries to address this

Then that's a bug in out call implementation. The relevant part of the spec for that is here: https://tc39.es/ecma262/#sec-ordinarycallbindthis
Which gets called by https://tc39.es/ecma262/#sec-ecmascript-function-objects-call-thisargument-argumentslist (step 6). You'll see it has behaviour depending on whether we're currently in strict mode or not - `"use strict"; Array.from([1, 2, 3], function () { console.log(this) });` will print `undefined` three times (in compliant engines, that is), but not if we explicitly pass in the global object.

tuqqu: fair enough. You convinced me :) 
I made appropriate changes as well as changes to the PR message.

linusg (approved):

Issue #6086, opened 2021-04-03
LibGUI: Horizontal/VerticalBoxLayout are registered as widgets but do not derive from GUI::Widget


tomuta:
While changing `RefPtr` back to a non-threadsafe implemenation to address #4902 I ran into the problem that we apparently didn't properly check type conversions.

```
../.././AK/NonnullRefPtr.h: In instantiation of 'AK::NonnullRefPtr<T>::NonnullRefPtr(AK::NonnullRefPtr<U>&&) [with U = GUI::VerticalBoxLayout; T = GUI::Widget]':
../.././AK/Function.h:103:53:   required from 'Out AK::Function<Out(In ...)>::CallableWrapper<CallableType>::call(In ...) const [with CallableType = {anonymous}::<lambda()>; Out = AK::NonnullRefPtr<GUI::Widget>; In = {}]'
../.././AK/Function.h:100:13:   required from here
../.././AK/NonnullRefPtr.h:97:17: error: invalid 'static_cast' from type 'GUI::VerticalBoxLayout*' to type 'GUI::Widget*'
   97 |         : m_ptr(static_cast<T*>(&other.leak_ref()))
```

Anyway, here is the offending line:

https://github.com/SerenityOS/serenity/blob/22d13d8b1abb7210fe4091c23eb0013e64e02d96/Userland/Libraries/LibGUI/BoxLayout.cpp#L33

Without these changes adding a`static_assert` to the `REGISTER_WIDGET` macro can also catch it:
```c++
static_assert(IsBaseOf<GUI::Widget, namespace_::class_name>::value);
```

Other offenders:
```
../.././AK/NonnullRefPtr.h: In instantiation of 'AK::NonnullRefPtr<T>::NonnullRefPtr(AK::NonnullRefPtr<U>&&) [with U = FileIconsModel; T = MouseCursorModel]':
../../Userland/Demos/WidgetGallery/GalleryModels.h:112:87:   required from here
../.././AK/NonnullRefPtr.h:97:17: error: invalid 'static_cast' from type 'FileIconsModel*' to type 'MouseCursorModel*'
   97 |         : m_ptr(static_cast<T*>(&other.leak_ref()))
```
```
../.././AK/RefPtr.h: In instantiation of 'AK::RefPtr<T>& AK::RefPtr<T>::operator=(const AK::RefPtr<U>&) [with U = Web::CSS::StyleSheet; T = Web::CSS::CSSStyleSheet]':
[...]

Issue #4902, opened 2021-01-11
AK: Atomic ref-counting penalty is way too high in single-threaded code


awesomekling:
`RefPtr` locking shows up near the top in a lot of our profiles.

It's extremely visible for any code that does `quick_sort` on anything containing `String`s where each `swap` ends up doing three `PtrTraits::exchange()`.

Such sorting accounts for ~55% of `CrashReporter` runtime according to some local profiling.

We need a way to reduce the impact of thread-safe ref-counting on code that doesn't even run in a multi-threaded context.

One basic idea would be to have `RefCounted` and `ThreadUnsafeRefcounted`, `String` and `ThreadUnsafeString` etc. And then performance-sensitive single-threaded code could opt in to non-atomic reference counting.

Suggestions welcome!

cc @tomuta

ccapitalK:
I'm not sure I like the ThreadUnsafeRefCounted option, it would require a similar split in all structures and functions that use them that could be used in threaded code, I think it would result in a lot of code duplication. I have a couple of other ideas, but all of them seem to have drawbacks of some form.

Maybe we could do something like keeping track of the "threaded-ness" of RefPtrs, with all RefPtrs starting out as single-threaded and only switching over to a multi-threaded profile when accessed by a different thread? I feel like making this work in a way that turns out to actually being faster would be really difficult, and we would need to be very careful in how we construct accesses to RefPtrs to make sure it is thread safe in all access orderings, but I believe that this would give us the best of both worlds.

Another way would be to still encoding the threadedness, but having all objects intended to be used across threads be explicitly promoted to being MT-capable by the user around the time they create it (something like `this->set_threading_profile(Threading::MultiThreaded)`). We'd just need to make sure to do this for class members as well (make sure all members have something like `m_object->set_threading_profile(this->get_threading_profile())` run shortly after construction), but then at least the threading will be abstracted away a bit and we wouldn't need as much duplication.

Finally (and this idea is kinda half baked at this point, I'm just spit-balling), we may be able to do this by keeping a flag in thread local storage determining if whether we are constructing MT aware objects or not, which would be checked in RefPtr constructors. If we make it so users have to opt-in to disabling Thread safety in their ST hot-paths, we should be able to mitigate the worst performance pains, and it wouldn't require as much code duplication as the ThreadUnsafe approach.

Again, these are just rough ideas for how we might approach it, I don't really have any preference here.

tomuta:
[...]

PR #6249, opened 2021-04-11
AK: Simplify RefPtr again

tomuta:
Let's make `RefPtr` not thread-safe itself again (similar to the std library) as the overhead is just too much, considering most of our code is single-threaded and doesn't need these features.

- [ ] Fix weird linker error:
```
i686-pc-serenity/bin/ld: Userland/Libraries/LibGUI/CMakeFiles/LibGUI.dir/Button.cpp.o: relocation R_386_GOTOFF against undefined symbol `_ZN12_GLOBAL__N_119registration_WidgetE' can not be used when making a shared object
i686-pc-serenity/bin/ld: final link failed: bad value
```
- [ ] Figure out a solution for #6086

### DIFF
--- a/Userland/Libraries/LibGUI/Widget.h
+++ b/Userland/Libraries/LibGUI/Widget.h
@@ -39,9 +39,13 @@
 #include <LibGfx/Rect.h>
 #include <LibGfx/StandardCursor.h>
 
+namespace {
+    extern Core::ObjectClassRegistration registration_Widget;
+}
+
 #define REGISTER_WIDGET(namespace_, class_name)                                                                                                 \
     namespace {                                                                                                                                 \
-    GUI::WidgetClassRegistration registration_##class_name(#namespace_ "::" #class_name, []() { return namespace_::class_name::construct(); }); \
+    Core::ObjectClassRegistration registration_##class_name(#namespace_ "::" #class_name, []() { return static_ptr_cast<Core::Object>(namespace_::class_name::construct()); }, &registration_Widget); \
     }
 
 namespace GUI {
@@ -53,25 +57,6 @@ enum class HorizontalDirection {
 enum class VerticalDirection {
     Up,
     Down
-};
-
-class WidgetClassRegistration {
-    AK_MAKE_NONCOPYABLE(WidgetClassRegistration);
-    AK_MAKE_NONMOVABLE(WidgetClassRegistration);
-
-public:
-    WidgetClassRegistration(const String& class_name, Function<NonnullRefPtr<Widget>()> factory);
-    ~WidgetClassRegistration();
-
-    String class_name() const { return m_class_name; }
-    NonnullRefPtr<Widget> construct() const { return m_factory(); }
-
-    static void for_each(Function<void(const WidgetClassRegistration&)>);
-    static const WidgetClassRegistration* find(const String& class_name);
-
-private:
-    String m_class_name;
-    Function<NonnullRefPtr<Widget>()> m_factory;
 };
 
 enum class FocusPolicy {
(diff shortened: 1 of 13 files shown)


bgianfo:
If the overhead of RefPtr is too much of a perf hit, isn't that a sign that we are missing important moves(..) / passing by RefPtr by value in too many places?

tomuta:
> If the overhead of RefPtr is too much of a perf hit, isn't that a sign that we are missing important moves(..) / passing by RefPtr by value in too many places?

I don't think that's a problem, it's the loops with atomic locking of the pointer value inside that kills performance (see #4902)

Userland/Libraries/LibGUI/Widget.h
```diff
@@ -39,9 +39,13 @@
 #include <LibGfx/Rect.h>
 #include <LibGfx/StandardCursor.h>
 
+namespace {
+    extern Core::ObjectClassRegistration registration_Widget;
```
tomuta, line 43: Not sure why this is causing the link error:
```
i686-pc-serenity/bin/ld: Userland/Libraries/LibGUI/CMakeFiles/LibGUI.dir/Button.cpp.o: relocation R_386_GOTOFF against undefined symbol `_ZN12_GLOBAL__N_119registration_WidgetE' can not be used when making a shared object
i686-pc-serenity/bin/ld: final link failed: bad value
```

bgianfo:
> > If the overhead of RefPtr is too much of a perf hit, isn't that a sign that we are missing important moves(..) / passing by RefPtr by value in too many places?
> 
> I don't think that's a problem, it's the loops with atomic locking of the pointer value inside that kills performance (see #4902)

I looked at this some more... referencing the issue you mentioned and I think move(...) of String is just completely messed up? Each `swap(..)` is just calling `String`'s move constructor / move operator which just forward the move to the internal `StringImpl` ... `StringImpl` doesn't implement a move constructor / move operator. `StringImpl` derives form `RefCounted<T>` which also doesn't implement move constructor / move operator. `RefCounted<T>` derives from `RefcountedBase`.. which is marked as `AK_MAKE_NONCOPYABLE` / `AK_MAKE_NONMOVABLE`. Is there really a reason why we can't implement a more optimized `move(..)` for strings? I don't see why it's not just detach from one object and attach to the other? The one already has a stable refcount.. why do we need to increment / decrement at all... we just need to move owership of the ref?

bgianfo:
cc: @alimpfard in case they have ideas / I'm missing something.

alimpfard:
`RefPtr<T>` seems to have a move ctor though? `RefPtr.h:165`
The actual cost here seems to be in `RefPtrTraits<T>::exchange`, which forces an strong CAS on everything, which is used in that move ctor

alimpfard:
Honestly, I don't think we need to actually worry about the state of the moved-from RefPtr, how about a `is_moved_from` (or equivalent) field that we set on move instead of actually setting it to null?

awesomekling:
> Honestly, I don't think we need to actually worry about the state of the moved-from RefPtr, how about a `is_moved_from` (or equivalent) field that we set on move instead of actually setting it to null?

Moving from `RefPtr` is a well-defined operation that nulls out the pointer. We do this in tons of places. Perhaps you are thinking of `NonnullRefPtr`?

alimpfard:
What I'm saying is, I don't think we need to _actually_ set the bits to zero, we could instead set a flag and treat said RefPtr as null, at least, not in `leak_ref_raw()`

awesomekling:
> What I'm saying is, I don't think we need to _actually_ set the bits to zero, we could instead set a flag and treat said RefPtr as null, at least, not in `leak_ref_raw()`

Where do you store that flag though? (without increasing the size of `RefPtr`)

awesomekling:
Regardless, having implicit synchronization built into a smart pointer is weird and extremely heavy. As Tom indicates in the commit, we should simply use external synchronization like any other code would.

alimpfard:
well if we're going to ignore the thread-safety, then that's fine, no need for safety guarantees when it's not meant to be safe lol
if not though, I don't think making it a bit larger is a big loss if we can gain some perf benefit from it by not synchronising the null state

PR #6253, opened 2021-04-11
Even more ports changes

gunnarbeutner:
This PR:

* Ensures that `pkg-config` doesn't pick up packages from the host system.
* Fixes build errors for SDL2_ttf when X11 headers are installed on the host.
* Adds support for running (some) port build/configure commands with the host toolchain.
* Adds a port for libicu (not thread-safe and mostly untested).
* Fixes the return type for `sleep()` - it should be `unsigned int`.
* Fixes the declaration for `regex_t` and `regmatch_t` - these should be typedefs rather than structs.
* Adds a port for nethack. Because, well, why not.
* Fixes the ninja port (similar problem like libexpat).
* Adds a missing macro for _SC_CLK_TCK.
* Fixes the hatari port by removing an obsolete patch.
* Adds a port for pcre.
* Links libarchive against pcre. I don't see how this could've previously worked because the target's libregex doesn't have the re_nsub field. It probably used the host's header file and would've crashed at runtime due to mismatching struct layouts.
* Fixes the nano port by adding a missing macro.
* Fixes building the dash port.
* Adds support for `isblank()`.
* Removes duplicate code in `isascii()`.
* Fixes the mrsh port.


gunnarbeutner:
This currently breaks the changes from https://github.com/SerenityOS/serenity/commit/938924f36de1f335903608c28c4ba5284d5dd883 - I'll fix that in a bit.

danboid:
Oooh!

I'm liking:

* Ensures that pkg-config doesn't pick up packages from the host system.

We've been moaning about that one for years and

* Adds support for running (some) port build/configure commands with the host toolchain.

I've wanted to see that in serenity for some time, if it means what I hope it means.

Great work @gunnarbeutner !

gunnarbeutner:
Thank you!

I needed both host and target toolchains for some ports. For example, dash tries to compile and run some utilities during build time which obviously fails when those utilities are built with the target toolchain. I've decided to just build dash twice - once for the host and once for the target. That way I can copy over the host version of the utilities into the target build tree and use them there.

gunnarbeutner:
I've decided to move some of the changes into their own PRs. Also, I've removed the `getaddrinfo()` stub because I have a fully functional version of that in another branch.

The problem with the recent `.hosted_defs.sh` changes should be resolved with the latest push.

gunnarbeutner:
I've rebased the PR against the latest master. Should be good to go as far as I'm concerned. :)

PR #6258, opened 2021-04-12
HackStudio+LibDebug: Various debugger fixes

FalseHonesty:
HackStudio's debugger instance has its source root property updated when switching projects, and breakpoints will properly canonicalize their file paths as the Debugger now expects.

This fixes breakpoints not activating in the "little" project (~~although actually trying to debug the program still trips an assertion in LibDebug, probably tackling that issue next~~ fixed!).

cc @itamar8910


itamar8910 (approved):
Great fixes :smile:

PR #6265, opened 2021-04-12
VirtualFileSystem: Added checks whether mount points are already in use

paulmetzger:
Hello,

I implemented two "FIXMEs" in VirtualFileSystem.cpp that proposed to check whether a mount point is already in use.
The _mount_ syscall now returns EBUSY if the requested mount point is in use.
This pull request also includes a fix for a segfault in the DiskCache class in BlockedFileSystem.cpp, which was triggered by these checks.
When DiskCache was deconstructed the fields m_cached_block_data and m_entries were deconstructed before m_clean_list and m_dirty_list, which, without this fix, attempt to use the already deallocated buffers m_cached_block_data and m_entries in their deconstructors.
I fixed this by changing the order in which these fields are deconstructed.
This is only a small pull request to familiarise myself with the code.

Paul

PS: This is my first PR. Please let me know if I am doing something wrong.

### DIFF
--- a/Kernel/FileSystem/BlockBasedFileSystem.cpp
+++ b/Kernel/FileSystem/BlockBasedFileSystem.cpp
@@ -117,10 +117,10 @@ private:
     BlockBasedFS& m_fs;
     size_t m_entry_count { 10000 };
     mutable HashMap<BlockBasedFS::BlockIndex, CacheEntry*> m_hash;
-    mutable IntrusiveList<CacheEntry, &CacheEntry::list_node> m_clean_list;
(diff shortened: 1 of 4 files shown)


Granddave:
I'm not really qualified to review the _changes_ in this PR, but I would recommend reading through the [code submission policy](https://github.com/SerenityOS/serenity/blob/master/CONTRIBUTING.md#code-submission-policy). I'm specifically thinking about commit category and commit message length. :slightly_smiling_face: 
I believe that either "Kernel" or "VFS" would be appropriate as category.

paulmetzger:
Thanks for you response! I'll have a look.

PR #6268, opened 2021-04-12
LibCore+LibGUI: add StandardPaths::documents_directory and add Documents and Downloads directories to tray shortcuts in FilePicker

ZikiFlicky:


awesomekling (changes requested):
This is not used by anything? Please don't add dead code. Instead, add this in a PR that uses it for something.

ZikiFlicky:
I was planning to add a Documents shortcut in the left toolbar in the FilePicker widget that will use the method, but I wasn't sure if it belongs in this pull request. Sorry for not mentioning it 😬

ZikiFlicky:
I added Documents and Downloads folders to FilePicker. can you review the changes?

Userland/Libraries/LibGUI/FilePicker.cpp
```diff
@@ -243,12 +243,18 @@ FilePicker::FilePicker(Window* parent_window, Mode mode, const StringView& file_
...
 
     m_model->on_complete = [&] {
         if (m_model->root_path() == Core::StandardPaths::home_directory()) {
             home_button.set_checked(true);
         } else if (m_model->root_path() == Core::StandardPaths::desktop_directory()) {
             desktop_button.set_checked(true);
+        } else if (m_model->root_path() == Core::StandardPaths::downloads_directory()) {
+            downloads_button.set_checked(true);
+        } else if (m_model->root_path() == Core::StandardPaths::documents_directory()) {
+            documents_button.set_checked(true);
```
Dexesttp, line 257: You will need to uncheck these buttons when another path is selected, which is done right below :) Just add the `downloads_button.set_checked(false);` and `documents_button.set_checked(false);` to the else branch.

PR #6271, opened 2021-04-12
WindowServer: Fix notification placement

### LINKED ISSUE #6262: WindowServer: Notification placement looks off now that the menu bar is gone
As mentioned by Clovqr on Discord. Could either move them up a bit, or down to the task bar, stacking upwards rather than downwards.

image

vkoskiv:
No more top bar, no more extra padding. Spacing is equal on the top and
on the left, so it looks pretty tidy.

Resolves #6262

### DIFF
--- a/Userland/Services/NotificationServer/NotificationWindow.cpp
+++ b/Userland/Services/NotificationServer/NotificationWindow.cpp
@@ -46,7 +46,7 @@ static void update_notification_window_locations(const Gfx::IntRect& screen_rect
         auto& window = window_entry.value;
         Gfx::IntPoint new_window_location;
         if (last_window_rect.is_null())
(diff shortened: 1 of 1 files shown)


Userland/Services/NotificationServer/NotificationWindow.cpp
```diff
@@ -46,7 +46,7 @@ static void update_notification_window_locations(const Gfx::IntRect& screen_rect
         auto& window = window_entry.value;
         Gfx::IntPoint new_window_location;
         if (last_window_rect.is_null())
-            new_window_location = screen_rect.top_right().translated(-window->rect().width() - 24, 26);
+            new_window_location = screen_rect.top_right().translated(-window->rect().width() - 24, 7);
```
vkoskiv, line 49: Maybe these magic numbers could be documented with a constexpr variable somewhere?

PR #6272, opened 2021-04-12
Browser: Process Expires cookie attribute, add command to dump cookies

trflynn89:


AK/StringView.cpp
```diff
@@ -295,20 +294,20 @@ Optional<size_t> StringView::find_first_of(const StringView& view) const
 
 Optional<size_t> StringView::find_last_of(char c) const
 {
-    for (size_t pos = m_length; --pos > 0;) {
-        if (m_characters[pos] == c)
-            return pos;
+    for (size_t pos = m_length; pos != 0; --pos) {
+        if (m_characters[pos - 1] == c)
+            return pos - 1;
```
GMTA, line 299: Couldn't the fix for this just be simply replacing `--pos > 0` with `--pos >= 0`? Unless I'm missing something

trflynn89: `size_t` is unsigned so it's always >= 0

GMTA: Right, missed that! Test post please ignore :-)

PR #6277, opened 2021-04-12
Kernel: Introduce a new graphics subsystem + add basic support for Intel native graphics

supercomputer7:
There's still a lot to fix (and to change probably), but it works on my ICH7 (PCI ID of the graphics adapter is `8086:29c2`).
For now, the patch just makes the kernel to check what is the preferred resolution indicated by the EDID and it tries to set it after performing some calculations.

If anyone wants to test this on his old Intel machine, you need to add the PCI ID in the "whitelist" of the driver and then you have 2 options:
1. Use the iPXE bootloader (which doesn't set a resolution like GRUB) and try to boot the kernel with a ramdisk over the network
2. Remove the flag in the `boot.S` file and then try to boot with GRUB (either with a USB GRUB rescue drive booting so you will need a ramdisk like with the PXE method or just `dd` the image to a harddrive).


supercomputer7:
Although the driver doesn't check if the resolution is valid, it's ready and it is working on my ICH7 test machine. I tested first a old display (which supported 1600x900 resolution), and then another display which supports 1920x1080. Both worked just fine :)

danboid:
Very cool stuff @supercomputer7 !

supercomputer7:
> Very cool stuff @supercomputer7 !

Thank you, and I hope this patch will help people to use Serenity more on bare metal :)

danboid:
I'll test this out soon. I've got a mini ATX machine (a Celeron) with Intel graphics that I tried to network boot serenity recently but I didn't get a picture so hopefully these patches will fix that.

supercomputer7:
I tested this on another monitor which its native resolution is 1280x1024 and it worked OK with it :)

Kernel/Graphics/Definitions.h
```diff
@@ -0,0 +1,105 @@
...
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#pragma once
+
+#include <AK/Types.h>
+
+namespace Kernel {
+
+struct Timings {
```
awesomekling, line 33: `Kernel::Timings` feels like an overly generic name here.
Since we are doing this in a new directory, how about we put it in a `Kernel::Graphics` namespace?

Kernel/Graphics/IntelNativeGraphicsAdapter.h
```diff
@@ -0,0 +1,162 @@
...
+#include <Kernel/Graphics/Definitions.h>
+#include <Kernel/Graphics/GraphicsDevice.h>
+#include <Kernel/Graphics/RawFramebufferDevice.h>
+#include <Kernel/PCI/DeviceController.h>
+#include <Kernel/PhysicalAddress.h>
+
+namespace Kernel {
+
+class IntelNativeGraphicsAdapter final : public GraphicsDevice
+    , public PCI::DeviceController {
```
awesomekling, line 40: `clang-format` is not great at these, but I usually format multiple inheritance like this:

```cpp
class IntelNativeGraphicsAdapter final
    : public GraphicsDevice
    , public PCI::DeviceController {
```

Kernel/Graphics/Definitions.h
supercomputer7: Sounds great :)

supercomputer7:
I pushed a commit that explains in detail why I want to not use GRUB VBE modesetting.
What I suggest is that if a user tries to boot Serenity on real hardware and sees that text-mode is working OK, he could compile a new kernel that enables early modesetting with the flag I removed.

etaIneLp:
> The second reason is that we now have a native driver for Intel
graphics and although it doesn't support most Intel graphics cards out
there, it's a good starting point to support more cards.

GRUB can set the video mode on most, if not all graphics cards ever made. The Intel driver supports a small subset of Intel GPUs. Even Linux still supports a simple framebuffer, despite having native drivers for Intel, AMD, and Nvidia GPUs.

> For such purpose, setting a framebuffer early on boot can be problematic
especially on hardware that doesn't function correctly with our kernel.

How so?

> The fourth reason is that we want to support VGA text mode. Because
GRUB uses the VBE interface to change modes, it's not possible to drop
back to VGA text mode without having a native driver for a specific
card.

> What I suggest is that if a user tries to boot Serenity on real hardware and sees that text-mode is working OK, he could compile a new kernel that enables early modesetting with the flag I removed.

Computers booted through UEFI do not support text mode at all. I think that we should default to relying on GRUB to set a video mode. If someone wants to use text mode, they can remove the flag and recompile (though I'm not sure why you would want to do that, text mode on modern screens is horrendous).

supercomputer7:
> GRUB can set the video mode on most, if not all graphics cards ever made. The Intel driver supports a small subset of Intel GPUs. Even Linux still supports a simple framebuffer, despite having native drivers for Intel, AMD, and Nvidia GPUs.

We are not aiming at supporting all display hardware. Also, if we want to modeset the resolution with VBE, **we should do it with our code, and not to rely on GRUB to do this for us.**

> How so?
Many times when people are trying to boot on real hardware (me included), we can't get a picture on the screen due to a bug in the kernel. Also, we support bochs-display device so QEMU should work anyway. This change is meant to fix the situation on bare-metal.

> Computers booted through UEFI do not support text mode at all. I think that we should default to relying on GRUB to set a video mode. If someone wants to use text mode, they can remove the flag and recompile (though I'm not sure why you would want to do that, text mode on modern screens is horrendous).

We don't support UEFI yet. And even when we get there, the kernel is not dropping support of using preset framebuffers (look at my code to see that this is correct), it just doesn't request it from the GRUB multiboot protocol.

`If someone wants to use text mode, they can remove the flag and recompile (though I'm not sure why you would want to do that, text mode on modern screens is horrendous).`
Sorry, but the same can said otherwise - because the kernel still supports preset framebuffers, after you verified that the system boots up in text mode correctly, just add this flag and compile again :)

etaIneLp:
> We don't support UEFI yet. And even when we get there, the kernel is not dropping support of using preset framebuffers (look at my code to see that this is correct), it just doesn't request it from the GRUB multiboot protocol.

I had a look at UEFI GOP (and multiboot2, so we can tell GRUB that we want to use UEFI boot services) and they don't seem very complicated. I'll try to add support for those when this is merged, then it won't matter whether GRUB sets the video mode or not. :)

supercomputer7:
> > We don't support UEFI yet. And even when we get there, the kernel is not dropping support of using preset framebuffers (look at my code to see that this is correct), it just doesn't request it from the GRUB multiboot protocol.
> 
> I had a look at UEFI GOP (and multiboot2, so we can tell GRUB that we want to use UEFI boot services) and it doesn't seem very complicated. I'll try to add support for those when this is merged, then it won't matter whether GRUB sets the video mode or not. :)

If you want to do this, I'm all for it. Just ensure that the user can disable the UEFI GOP somehow, or ensure that we emulate text mode with the framebuffer. For the time being, UEFI is really not a goal (at least not my goal), as I'm almost sure we need x86-64 support before this can happen... if you feel like you want to work on it, then do it :)

supercomputer7:
I'm looking now to add better support of the text mode when doing bare-metal debugging.
I had an idea to allow GRUB to modeset the resolution (partially-reverting the last commit) so we won't require native drivers on bare metal for now (though we should do that eventually when we have enough native drivers, or VESA VBE modesetting code in the kernel).

If we want to enable text mode we will create another type of console called FramebufferConsole. On boot, if the user requests text mode, we will attach the appropriate console, which depends on whether the bootloader provided a framebuffer or not before booting our kernel. The other consoles could be `VGAConsole`, so we could allow the user to change fonts of the tty, and a basic 80x25 text mode console like now.

The inspiration of the idea is based on the `fbcon` code from Linux, which is what I described in the previous paragraph. The framebuffer console in Linux prints fonts without the help of userspace, from what I can tell after looking at the code.
The question is - do we want to print the font within the kernel code or not... Linux does that (and that's probably fine), and maybe we should too allow printing debug messages, even when we have a high resolution framebuffer. I know there are some disadvantages to this approach, but if we want to neatly implement text consoles (so we don't ask people to recompile without the multiboot resolution setting flag) and still be able to debug problems on bare metal, this is probably the finest solution (I could think of).
Any suggestion or comment is appreciated :)

cc @awesomekling

awesomekling:
Rendering basic bitmap fonts in the kernel should be fine. But is it really valuable to add a high-resolution framebuffer console? The text console is really only useful for debugging why you are unable to get into full screen graphics mode, it's not really an intended way to run this system.

supercomputer7:
> Rendering basic bitmap fonts in the kernel should be fine. But is it really valuable to add a high-resolution framebuffer console? The text console is really only useful for debugging why you are unable to get into full screen graphics mode, it's not really an intended way to run this system.

I do think there's a value to adding such mode. There are mainly 2 reasons for this.
When it was decided that the SerenityOS project will set a goal of keeping away non-technical users (so we don't provide pre-built images) , and that definition is really on the margin of saying "people that don't care to read the documentation for about 5 minutes and to figure out how compile the system", I thought that this is indeed the right way to keep the project safe from invaluable bug reports, lots of feature requests by non-contributors and other problems that arise from such situation. 

What I don't think is right for the project is to expect people to start figuring out why and how the kernel requests a framebuffer from GRUB. 
The problem is with how flexible the multiboot protocol is, but since this is our reality and we don't have other options, we should be flexible as much as we can (so we are trying to help people to build and use the system without compromising on quality of what we do). So, if iPXE, QEMU and SYSLINUX don't set a framebuffer, we should either use text mode or have a native driver (because really you don't have anything else better to offer).
However, if GRUB supplies a framebuffer, we can still detect it and support it. Same thing for UEFI GOP when that will be supported.

Second reason is that I do see it valuable to have a framebuffer console. Let's say that you booted the system, everything works as you expected, but the WindowServer got stuck (or something else), you could just switch to console mode (like how it works with multiple ttys on Linux). You could even make the `Alt+Shift+F12` combination to print the kernel / scheduler state on the framebuffer console and switch to it so the user can view it even without having a Serial connection on bare-metal :)

awesomekling:
> > Rendering basic bitmap fonts in the kernel should be fine. But is it really valuable to add a high-resolution framebuffer console? The text console is really only useful for debugging why you are unable to get into full screen graphics mode, it's not really an intended way to run this system.
> 
> I do think there's a value to adding such mode. There are mainly 2 reasons for this.
> When it was decided that the SerenityOS project will set a goal of keeping away non-technical users (so we don't provide pre-built images) , and that definition is really on the margin of saying "people that don't care to read the documentation for about 5 minutes and to figure out how compile the system", I thought that this is indeed the right way to keep the project safe from invaluable bug reports, lots of feature requests by non-contributors and other problems that arise from such situation.
> 
> What I don't think is right for the project is to expect people to start figuring out why and how the kernel requests a framebuffer from GRUB.
> The problem is with how flexible the multiboot protocol is, but since this is our reality and we don't have other options, we should be flexible as much as we can (so we are trying to help people to build and use the system without compromising on quality of what we do). So, if iPXE, QEMU and SYSLINUX don't set a framebuffer, we should either use text mode or have a native driver (because really you don't have anything else better to offer).
> However, if GRUB supplies a framebuffer, we can still detect it and support it. Same thing for UEFI GOP when that will be supported.
> 
> Second reason is that I do see it valuable to have a framebuffer console. Let's say that you booted the system, everything works as you expected, but the WindowServer got stuck (or something else doesn't work, as it happens sometimes on @linusg machine), you could just switch to console mode (like how it works with multiple ttys on Linux). You could even make the `Alt+Shift+F12` combination to print the kernel / scheduler state on the framebuffer console and switch to it so the user can view it even without having a Serial connection on bare-metal :)

Okay, you have convinced me :)

supercomputer7:
Last commit seems to work OK on QEMU but not on my ICH7 test machine. On QEMU I can get either a framebuffer console or graphical environment. On bare metal I get `kmalloc: Out of memory` errors, but I have 2GB RAM on that machine, so I really don't know what's wrong with the kernel.

@awesomekling

supercomputer7:
Apparently having a `VirtualConsole` with width x height bigger than 80x25 is what causing the stack overflow and the triple fault.
I need to check why this happens, probably it's a fault in my change.

supercomputer7:
This is not yet clear, but it seems like we triple-fault due to faulting at double-faulting handler. This can happen if we try to print what's going on, but printing is making the system to fault again.

supercomputer7:
We are nearly done! A few more fixes and it's ready for review :)

supercomputer7:
We should still minimize the code size of `VirtualConsole`. I might have forgotten commented code, so I'll search for this and will delete where I find it. Other than that, it's ready for review (but not for merging probably).

Kernel/Graphics/FramebufferDevice.cpp
```diff
@@ -0,0 +1,156 @@
...
+{
+    m_real_framebuffer_vmobject = AnonymousVMObject::create_for_physical_range(m_framebuffer_address, page_round_up(framebuffer_size_in_bytes()));
+    VERIFY(m_real_framebuffer_vmobject);
+    m_real_framebuffer_region = MM.allocate_kernel_region_with_vmobject(*m_real_framebuffer_vmobject, framebuffer_size_in_bytes(), "Framebuffer", Region::Access::Read | Region::Access::Write);
+    VERIFY(m_real_framebuffer_region);
+    m_swapped_framebuffer_vmobject = AnonymousVMObject::create_with_size(page_round_up(framebuffer_size_in_bytes()), AllocationStrategy::AllocateNow);
+    VERIFY(m_swapped_framebuffer_vmobject);
+    m_swapped_framebuffer_region = MM.allocate_kernel_region_with_vmobject(*m_swapped_framebuffer_vmobject, page_round_up(framebuffer_size_in_bytes()), "Framebuffer Swap (Blank)", Region::Access::Read | Region::Access::Write);
+    VERIFY(m_swapped_framebuffer_region);
+}
```
ElectrodeYT, line 96: Do we need to allocate them into kernel space permenantly?

Kernel/Graphics/GraphicsDevice.h
```diff
@@ -0,0 +1,38 @@
...
+namespace Kernel {
+class FramebufferDevice;
+class GraphicsDevice : public RefCounted<GraphicsDevice> {
+public:
+    enum class Type {
+        VGACompatible,
+        Bochs,
+        SVGA,
+        Raw
+    };
```
ElectrodeYT, line 24: Is this required?

supercomputer7: Technically yes, because `GraphicsManagement` holds `m_vga_adapter` and it determines the appropriate device by looking at the type of it. Later on, we could try to make VGA compatible adapters to work (or to not work) in VGA mode manually.

Kernel/Graphics/FramebufferDevice.cpp
supercomputer7: Yes. When we switch between graphical mode and textual mode, we need to ensure we restore everything (especially in graphical mode).

Kernel/TTY/VirtualConsole.cpp
```diff
@@ -1,129 +1,58 @@
...
-    , m_terminal(*this)
 {
-    VERIFY(index < s_max_virtual_consoles);
-
     m_tty_name = String::formatted("/dev/tty{}", m_index);
-    m_terminal.set_size(80, 25);
+    VERIFY(GraphicsManagement::the().console());
+    change_size(GraphicsManagement::the().console()->max_column(), GraphicsManagement::the().console()->max_row());
+    // 2 mebibytes seems like a reasonable amount of Cells to store within...
+    m_cells = MM.allocate_kernel_region(2 * MiB, "Virtual Console Cells", Region::Access::Read | Region::Access::Write, AllocationStrategy::AllocateNow);
```
ElectrodeYT, line 41: Couldn't this be calculated?

Kernel/Graphics/Console/FramebufferConsole.cpp
```diff
@@ -0,0 +1,352 @@
...
+    { 0x00, 0x00, 0x63, 0x6B, 0x7F, 0x7F, 0x36, 0x00 }, // U+0077 (w)
+    { 0x00, 0x00, 0x63, 0x36, 0x1C, 0x36, 0x63, 0x00 }, // U+0078 (x)
+    { 0x00, 0x00, 0x33, 0x33, 0x33, 0x3E, 0x30, 0x1F }, // U+0079 (y)
+    { 0x00, 0x00, 0x3F, 0x19, 0x0C, 0x26, 0x3F, 0x00 }, // U+007A (z)
+    { 0x38, 0x0C, 0x0C, 0x07, 0x0C, 0x0C, 0x38, 0x00 }, // U+007B ({)
+    { 0x18, 0x18, 0x18, 0x00, 0x18, 0x18, 0x18, 0x00 }, // U+007C (|)
+    { 0x07, 0x0C, 0x0C, 0x38, 0x0C, 0x0C, 0x07, 0x00 }, // U+007D (})
+    { 0x6E, 0x3B, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00 }, // U+007E (~)
+    { 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00 }  // U+007F
+};
```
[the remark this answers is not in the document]
supercomputer7: Why? it's not like some other code will use it, right?

Kernel/Graphics/Console/FramebufferConsole.cpp
```diff
@@ -0,0 +1,352 @@
...
+        m_x = 0;
+        m_y += 1;
+        if (m_y >= max_row())
+            m_y = 0;
+        return;
+    }
+    if ((int)ch < 0x20 || (int)ch == 0x7f) {
+        // There's no point in printing empty glyphs...
+        return;
+    }
```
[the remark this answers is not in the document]
supercomputer7: So it's more like a "FIXME", right? The current glyphs below 0x20 are empty, so we need to create those first.

Kernel/TTY/VirtualConsole.cpp
supercomputer7: Probably yes. Let me look into a solution :)

Kernel/Graphics/Console/FramebufferConsole.cpp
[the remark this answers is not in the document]
ElectrodeYT: Yeah, true

Kernel/Graphics/Console/FramebufferConsole.cpp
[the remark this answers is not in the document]
ElectrodeYT: One could reprogram the VGA console font (at least I think), one could use it for that.
I was mostly thinking about neatness and organisation, but I'm not sure I'm the best at it anyway :)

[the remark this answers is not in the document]
supercomputer7: When we come to do this, I'm sure that person will take this into consideration and will take this array into a separate file :)

supercomputer7:
So we have one implementation of VT consoles, both for kernel and userspace. We just need to ensure we can scroll back on the Virtual console and I can call it a day.

ElectrodeYT:
Doesn't work on VMWare and bare metal with MULTIBOOT_VIDEO_MODE set and a framebuffer:
VMWare log:
```
[#0 Kernel]: Forcing no initialization of framebuffer devices
[Kernel]: Graphics: Using a preset resolution from the bootloader
[#0 Kernel]: Framebuffer Console: taking 1310720 bytes
[#0 Kernel]: Graphics adapter @ PCI [0000:00:0f:00] is operating in VGA mode
[#0 Kernel]: KUBSAN: member call on null pointer of type 'struct Thread'
[#0 Kernel]: KUBSAN: at ../../Kernel/Arch/i386/CPU.cpp, line 280, column: 51
[#0 Kernel]: 0xc052a98d (next: 0xc0107e24)
[#0 Kernel]: 0xc052c7fd (next: 0xc0107ea4)
[#0 Kernel]: 0xc05b1cb3 (next: 0xc0107f64)
[#0 Kernel]: 0xc05a872a (next: 0xc0108008)
[#0 Kernel]: 0xc0192b2d (next: 0xc01080b8)
[#0 Kernel]: 0xc0395bb3 (next: 0xc0108108)
[#0 Kernel]: 0xc03942bd (next: 0xc0108188)
[#0 Kernel]: 0xc039431f (next: 0xc01081a8)
[#0 Kernel]: 0xc065c9b4 (next: 0xc0108218)
[#0 Kernel]: 0xc06a36b3 (next: 0xc0108298)
[#0 Kernel]: 0xc0100145 (next: 0x00000000)
[Kernel]: KERNEL PANIC! :^(
[Kernel]: UB is configured to be deadly.
[Kernel]: at ../../Kernel/UBSanitizer.cpp:28 in void print_location(const Kernel::UBSanitizer::SourceLocation&)
[#0 Kernel]: 0xc039f9fe (next: 0xc0107d84)
[#0 Kernel]: 0xc052a9ff (next: 0xc0107e24)
[#0 Kernel]: 0xc052c7fd (next: 0xc0107ea4)
[#0 Kernel]: 0xc05b1cb3 (next: 0xc0107f64)
[#0 Kernel]: 0xc05a872a (next: 0xc0108008)
[#0 Kernel]: 0xc0192b2d (next: 0xc01080b8)
[#0 Kernel]: 0xc0395bb3 (next: 0xc0108108)
[#0 Kernel]: 0xc03942bd (next: 0xc0108188)
[#0 Kernel]: 0xc039431f (next: 0xc01081a8)
[#0 Kernel]: 0xc065c9b4 (next: 0xc0108218)
[#0 Kernel]: 0xc06a36b3 (next: 0xc0108298)
[#0 Kernel]: 0xc0100145 (next: 0x00000000)
```

IdanHo:
Do you mind cherry picking d4fb2c24a51a358309e3bc6923d8a68652a640d6 into a separate PR? (My testing machine is crashing on the same bug i assume you fixed there)

Kernel/Graphics/GraphicsManagement.h
```diff
@@ -0,0 +1,38 @@
...
+
+class GraphicsManagement {
+    AK_MAKE_ETERNAL;
+
+public:
+    static GraphicsManagement& the();
+    static bool is_initialized();
+    bool initialize();
+
+    unsigned current_minor_number() { return m_current_minor_number++; };
```
awesomekling, line 35: This name makes it sound like it simply returns the `current minor number`, whatever that is.

Let's call it something a bit more accurate, how about `allocate_minor_device_number()`?

PR #6278, opened 2021-04-12
Kernel: Mark s_syscall_table const so it ends up in ro_data.

bgianfo:


awesomekling:
Hmm wait, do we write to this *during* init? Otherwise we could just make it `const` and move it into rodata :)

bgianfo:
Yea, you're right

PR #6280, opened 2021-04-12
Meta: Enable pull_request_target context for the notify_discord action

IdanHo:
While we did allow the notify_discord job to run inside the context, we didnt ask github to run in that context (resulting in no notifications about PRs being posted), so this commit enables that. While i was there i also noticed a flaw in the implementation of the build status posts for PRs: since we run the discord posts in a seperate context for PRs their build status would always be success (as the reported build status is only for the jobs in the current context), so this commit uses the "action-wait-for-check" sub-action (which i forked from the original to add missing functionality https://github.com/IdanHo/action-wait-for-check) to ensure the posted build results are accurate for pull requests (since the build workflow is done in a separate context for PRs).
The logic turned out quite complex, if someone has any good ideas on how to simplify it i would be glad to hear about it :)

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -253,31 +253,3 @@ jobs:
         ${{ toJSON(github.event) }}
         ]
         EOF
-
-  notify_discord:
-    needs: [build_and_test_serenity, build_and_test_lagom]
-    runs-on: ubuntu-20.04
-    if: always() && github.repository == 'SerenityOS/serenity' && (github.event_name == 'pull_request_target' || (github.event_name == 'push' && github.ref == 'refs/heads/master'))
-
-    steps:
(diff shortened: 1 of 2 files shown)


IdanHo:
oh, i have a small mistake in the action file, so dont merge this yet

IdanHo:
Well i had to solve another problem: discord's github webhook parser didnt know how to parse an event with the pull_request_target context, so i had to fork the discord sub-action and add a way to lie to discord about it being in the normal pull_request context (https://github.com/IdanHo/action-discord/commit/754598254f288e6d8e9fca637832e3c163515ba8), and i also split the discord notifications into a seperate workflow, since if i didnt we would have one duplicated skipped job for every test job (as they would be skipped in the pull_request_target context), but everything should "just work"™ now :D

PR #6290, opened 2021-04-13
LibWeb: Various improvements

Cleverking2003:
Some improvements to make acid2 render more correctly


Userland/Libraries/LibWeb/CSS/StyleResolver.cpp
```diff
@@ -177,6 +177,28 @@ static Vector<String> split_on_whitespace(const StringView& string)
     return v;
 }
 
+static Vector<String> split_on_character(const StringView& string, const char character)
```
awesomekling, line 180: Can we use `String::split_view()` here instead of implementing yet anther split function?

Cleverking2003: Fixed

PR #6291, opened 2021-04-13
Ports: Fix building the stress-ng port

gunnarbeutner:
This fixes a number of things:

* `mremap()` incorrectly resets the mapping's offset to zero when a non-zero offset was originally specified in `mmap()`.
* Add support for ELF objects where the text segment doesn't start at offset zero. For this to work `should_make_executable_exception_for_dynamic_loader()` now reads from the inode rather than the mapping.
* Add support for loading ELF objects which have multiple text and data segments.
* Build fixes for the stress-ng port.

Screenshot from 2021-04-13 19-46-12


bgianfo:
This is awesome, thanks for the fixes and bringing this forward!

bgianfo (approved):

PR #6295, opened 2021-04-13
All: Various fixes and cleanups

boricj:
Mostly stuff that I've stumbled upon while browsing the code and I'm too lazy to open three PRs for these.


awesomekling:
OMG that Minesweeper bug..
We really need to improve the palette system. But thank you for patching it up in the meantime :)

PR #6299, opened 2021-04-13
HackStudio+LibDebug: Improve the debugger "lifecycle"

### LINKED ISSUE #4393: HackStudio: Debugger leaks a thread for each session
It looks like HackStudio leaves behind a thread for every debug session. Upon closer look it looks like it creates a new thread but never detaches/joins it unless the `HackStudioWidget` is destroyed, in which case it will only join the last thread created. This is where the thread gets created:

https://github.com/SerenityOS/serenity/blob/ec89302fbe42987f037eb725d34a59a4936c5925/DevTools/HackStudio/HackStudioWidget.cpp#L515

FalseHonesty:
Adds support for stopping a currently debugged process. Also implements cleaning up the debugger thread when debugging is finished, which fixes #4393.

### DIFF
--- a/Userland/DevTools/HackStudio/Debugger/Debugger.cpp
+++ b/Userland/DevTools/HackStudio/Debugger/Debugger.cpp
@@ -110,6 +110,11 @@ int Debugger::start_static()
     return 0;
 }
 
+void Debugger::stop()
+{
+    set_requested_debugger_action(DebuggerAction::Exit);
+}
+
 void Debugger::start()
 {
     m_debug_session = Debug::DebugSession::exec_and_attach(m_executable_path, m_source_root);
@@ -190,11 +195,9 @@ int Debugger::debugger_loop()
             do_step_over(regs);
             return Debug::DebugSession::DebugDecision::Continue;
         case DebuggerAction::Exit:
-            // NOTE: Is detaching from the debuggee the best thing to do here?
-            // We could display a dialog in the UI, remind the user that there is
-            // a live debugged process, and ask whether they want to terminate/detach.
             dbgln("Debugger exiting");
-            return Debug::DebugSession::DebugDecision::Detach;
+            m_on_exit_callback();
+            return Debug::DebugSession::DebugDecision::Kill;
         }
         VERIFY_NOT_REACHED();
     });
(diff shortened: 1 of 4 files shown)


FalseHonesty:
cc @tomuta and @itamar8910 :)

Userland/DevTools/HackStudio/Debugger/Debugger.cpp
```diff
@@ -190,11 +195,11 @@ int Debugger::debugger_loop()
             do_step_over(regs);
             return Debug::DebugSession::DebugDecision::Continue;
         case DebuggerAction::Exit:
-            // NOTE: Is detaching from the debuggee the best thing to do here?
-            // We could display a dialog in the UI, remind the user that there is
-            // a live debugged process, and ask whether they want to terminate/detach.
             dbgln("Debugger exiting");
-            return Debug::DebugSession::DebugDecision::Detach;
+            // In case the debug session is exited while execution is already paused.
+            m_on_continue_callback();
```
itamar8910, line 200: I may be missing something - why do we call `m_on_continue_callback` here?

itamar8910 (approved):

Userland/DevTools/HackStudio/Debugger/Debugger.cpp
FalseHonesty: If the debug session is stopped while execution is paused, the hack studios debugger state doesn't properly clean up (namely, current execution point arrow). I could have either copied that code from the continue callback to the stop block or ran the continue callback, which IMO feels more correct.

itamar8910: Oh okay, thanks for clarifying that for me.
Personally I would prefer to copy the relevant code that resets the UI into the exit callback (its only 3 lines so still feels reasonable).
Calling the continue callback here seems kinda confusing and could also cause future trouble if we change the logic of the continue callback to include something we would not want to do when the debugger exits.

FalseHonesty: Okay, sounds reasonable :)

PR #6303, opened 2021-04-13
Browser: Add bookmark cap and entry for it in config file

one-some:
I thought this might be important as we currently load all bookmarks into memory.
The limit of 300 is rather arbitrary, so other numbers are welcome.

This is my first (relatively) "big" pr, so please excuse any obvious mistakes I may make. :)

### DIFF
--- a/Base/home/anon/.config/Browser.ini
+++ b/Base/home/anon/.config/Browser.ini
@@ -1,2 +1,3 @@
 [Preferences]
 Home=file:///res/html/misc/welcome.html
+MaxBookmarks=300
--- a/Userland/Applications/Browser/BookmarksBarWidget.cpp
+++ b/Userland/Applications/Browser/BookmarksBarWidget.cpp
@@ -33,6 +33,7 @@
 #include <LibGUI/Event.h>
 #include <LibGUI/JsonArrayModel.h>
 #include <LibGUI/Menu.h>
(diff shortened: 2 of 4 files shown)


awesomekling:
This seems like a totally arbitrary limitation for no reason. Bookmarks use a tiny amount of disk space.
If anything, it seems like a GUI problem instead -- how would we display that many bookmarks in a useful way?

one-some:
> This seems like a totally arbitrary limitation for no reason. Bookmarks use a tiny amount of disk space.

This is not intended to save disk space, but rather to ensure browser responsiveness and prevent OOM when loading bookmarks.
From my tests:
- At ~300 bookmarks the browser becomes significantly slow to resize.
- At ~10,000 bookmarks the browser becomes heavily unresponsive and very slow to start.
- At ~15,000 bookmarks, whenever you launch the browser, OOM errors occur.

In the future, the resizing slowdown from lower numbers may be fixed, however I still feel it is important to make sure the user doesn't accidentally run out of memory entirely just from having too many bookmarks. If someone really wanted to have a large number of bookmarks, the config file could be edited.

awesomekling:
Performance problems with a large number of bookmarks should be addressed at a UI level, not by enforcing arbitrary limitations on the number of bookmarks.

PR #6306, opened 2021-04-14
LibDebug: Support for parsing arrays

FalseHonesty:
When debugging variables with array types they will now display actual data instead of crashing.

preview

cc @itamar8910


Userland/Libraries/LibDebug/DebugInfo.cpp
```diff
@@ -215,7 +215,7 @@ static Optional<Dwarf::DIE> parse_variable_type_die(const Dwarf::DIE& variable_d
         variable_info.type_name = type_name.value().data.as_string;
     } else {
         dbgln("Unnamed DWARF type at offset: {}", type_die.offset());
-        variable_info.name = "[Unnamed Type]";
+        variable_info.type_name = "[Unnamed Type]";
```
awesomekling, line 218: Oopsie

PR #6311, opened 2021-04-14
Meta: lint-ports: Check package files for required properties

bcoles:
The intention here is to eventually support checking Ports for `auth_type`. This hasn't been implemented yet, as multiple packages do not yet use auth, so the build will show as failing.

### DIFF
--- a/Meta/lint-ports.py
+++ b/Meta/lint-ports.py
@@ -36,7 +36,7 @@ def read_port_dirs():
     """Check Ports directory for unexpected files and check each port has a package.sh file.
 
     Returns:
-        list: all ports (set), errors encountered (bool)
+        list: all ports (set), no errors encountered (bool)
     """
 
     ports = set()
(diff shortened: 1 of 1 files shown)


Meta/lint-ports.py
```diff
@@ -57,24 +57,64 @@ def read_port_dirs():
...
+                print('"Ports/{}" is missing "port" property'.format(package))
+                all_good = False
+
+            if 'version=' not in data:
+                print('"Ports/{}" is missing "version" property'.format(package))
+                all_good = False
+
+            if 'files=' not in data:
+                print('"Ports/{}" is missing "files" property'.format(package))
+                all_good = False
```
linusg, line 92: Checking if these appear *anywhere* in the file seems error-prone, maybe we can iterate over individual lines and check if they *start with* one of the prefixes instead?

linusg: Also, Python 3.5 had EOL and I don't think we need (or should) support it anymore - it's therefore safe to use f-strings rather than `.format()`.

bcoles: > 
> 
> Also, Python 3.5 had EOL and I don't think we need (or should) support it anymore - it's therefore safe to use f-strings rather than `.format()`.

The entirety of this file uses `format`. As do the other Python scripts:

```
$ grep -rn "format(" Meta/ | wc -l
24
```

And the contributing guidelines stipulate:

> Don't:
> Touch anything outside the stated scope of the PR.

bcoles:
Continued in #6380.

Meta/lint-ports.py
bcoles: > 
> 
> Checking if these appear _anywhere_ in the file seems error-prone, maybe we can iterate over individual lines and check if they _start with_ one of the prefixes instead?

Resolved in #6380 without iterating over lines.

PR #6315, opened 2021-04-14
SystemServer: Improve API and add service management interface

sppmacd:
This PR moves everything service-related to the new ServiceManagement class. It also adds a new command-line utility, `service`, which allows superuser to manage services.

For now, it supports `status`, `start`, `stop`, `enable` and `disable`.

Note that it should already be compatible with new socket subsystem made by @sin-ack.


linusg:
This has some merge conflicts, please rebase :)

sppmacd:
> This has some merge conflicts, please rebase :)

Is it OK now?

awesomekling:
cc @bugaevc

bugaevc:
I'm very sorry for still not having time to help reviewing this 😢

I'll try to make some time today.

Userland/Utilities/service.cpp
```diff
@@ -0,0 +1,92 @@
...
+
+    Core::EventLoop loop;
+    auto connection = Client::construct();
+
+    dbgln("$ service {} {}", service_name, command);
+
+    if (command_string == "status"sv) {
+        auto status = connection->send_sync<Messages::ServiceManagementServer::ServiceStatus>(service_name);
+        outln("{}", status->status());
+        loop.quit(0);
```
bugaevc: Does it ever actually run the loop? It appears to always `return 1` below.

Userland/Services/SystemServer/ClientConnection.cpp
```diff
@@ -59,8 +59,12 @@ OwnPtr<Messages::ServiceManagementServer::ServiceStatusResponse> ClientConnectio
 {
     auto service = ServiceManagement::the().find_service_by_name(message.service_name());
     dbgln("{}", service);
-    if (service)
-        return make<Messages::ServiceManagementServer::ServiceStatusResponse>(String::formatted("{}", service->is_enabled() ? "enabled" : "disabled"));
+    if (service) {
+        // FIXME: Handle multi-instance processes
+        auto enabled = service->is_enabled() ? "enabled" : "disabled";
+        auto running = service->is_running() ? "running" : "";
+        return make<Messages::ServiceManagementServer::ServiceStatusResponse>(String::formatted("{} {}", enabled, running));
```
bugaevc: I'm not sure if I like formatting these into a string server-side. Perhaps it could return typed info (here, two booleans), and format them in the `service` utility?

bugaevc: Ah, I see the next commit changes this to JSON. That's better, but perhaps explicit types at the IPC level would be better still.

Userland/Services/SystemServer/ClientConnection.cpp
```diff
@@ -56,6 +56,18 @@ OwnPtr<Messages::ServiceManagementServer::GreetResponse> ClientConnection::handl
...
 }
 
+OwnPtr<Messages::ServiceManagementServer::ServiceListResponse> ClientConnection::handle(const Messages::ServiceManagementServer::ServiceList&)
+{
+    // FIXME: Provide information about status
+    auto& services = ServiceManagement::the().services();
+    Vector<String> service_names;
+
+    for (auto& service : services)
+        service_names.append(String::formatted("{}{}", service.is_running() ? "+" : "-", service.name()));
```
bugaevc, line 46: This, too, formats info server-side; even after you've switched to JSON in the other place.

Userland/Services/SystemServer/ServiceManagementServer.ipc
```diff
@@ -4,4 +4,7 @@ endpoint ServiceManagementServer = 4151813
 
     ServiceList() => (Vector<String> services)
     ServiceStatus(String service_name) => (String status)
+    ServiceSetEnabled(String service_name, bool enabled) => ()
+    ServiceStart(String service_name) => ()
+    ServiceStop(String service_name) => ()
```
bugaevc, line 9: It makes sense to either make both accept booleans (`SetEnabled`, `SetRunning`), or make 4 distinct methods.

Userland/Services/SystemServer/Service.h
```diff
@@ -75,6 +76,12 @@ class Service final : public Core::Object {
     int m_priority { 1 };
     // Whether we should re-launch it if it exits.
     bool m_keep_alive { false };
+    // Whether it was stopped by `service XXX stop`
+    bool m_force_stopped { false };
+    // Path to the socket to create and listen on on behalf of this service.
+    String m_socket_path;
+    // File system permissions for the socket.
+    mode_t m_socket_permissions { 0 };
```
bugaevc, line 64: Huh? Why does this commit add those two fields?

Userland/Services/SystemServer/Service.h
```diff
@@ -75,6 +76,12 @@ class Service final : public Core::Object {
     int m_priority { 1 };
     // Whether we should re-launch it if it exits.
     bool m_keep_alive { false };
+    // Whether it was stopped by `service XXX stop`
+    bool m_force_stopped { false };
```
bugaevc, line 60: Perhaps `m_explicitly_stopped`?

And the comment should be <code>Whether it has been explicitly stopped, such as by &#96;service XXX stop&#96;</code>.

Userland/Services/SystemServer/Service.cpp
```diff
@@ -273,7 +310,7 @@ void Service::did_exit(int exit_code)
     s_service_map.remove(m_pid);
     m_pid = -1;
 
-    if (!m_keep_alive)
+    if (!m_keep_alive || m_force_stopped)
```
bugaevc, line 296: This should probably check whether the service is enabled rather than whether it has been explicitly stopped?

Userland/Services/SystemServer/Service.cpp
```diff
@@ -144,14 +157,38 @@ void Service::handle_socket_connection()
...
 
     if (m_lazy)
         setup_notifier();
     else
         spawn();
 }
 
+void Service::deactivate()
+{
+    // Already stopped
```
bugaevc, line 158: Nitpick: here and elsewhere, need periods at the ends of sentences.

Userland/Services/SystemServer/Service.cpp
```diff
@@ -144,14 +157,38 @@ void Service::handle_socket_connection()
...
+    // It makes that the service won't restart automatically.
+    m_force_stopped = true;
+
+    if (m_lazy)
+        m_socket_notifier = nullptr;
+    else {
+        // FIXME: Maybe add some configuration option for it ('AllowDeactivate')
+        // FIXME: Check for errors
+        kill(m_pid, SIGTERM);
+        did_exit(0);
```
bugaevc, line 171: No no no, you don't call `did_exit()` yourself, it gets called when you receive a notification from the kernel that it did, in fact, exit — via SIGCHLD, or potentially via other means.

Also this is not how you kill services. You kill with SIGTERM, yes, then you wait for some timeout and then you kill with SIGKILL. And importantly, you make super sure this does not race with you reaping the child, so that you never kill another process that happens to reuse the same PID. This is not a trivial task.

Userland/Services/SystemServer/Service.cpp
```diff
@@ -144,14 +157,38 @@ void Service::handle_socket_connection()
 
 void Service::activate()
 {
-    VERIFY(m_pid < 0);
+    // Already running
+    if (m_pid >= 0)
+        return;
+    
+    // It makes that the service won't restart automatically.
+    m_force_stopped = false;
```
bugaevc, line 148: This causes the service to be restarted automatically, not the other way around, doesn't it?

Userland/Services/SystemServer/Service.cpp
```diff
@@ -56,13 +56,16 @@ Service* Service::find_by_pid(pid_t pid)
 void Service::set_enabled_for_current_boot_mode(bool enabled)
 {
     extern String g_boot_mode;
-    
+
```
bugaevc, line 39: This should be fixed in the previous commit then.

Userland/Services/SystemServer/Service.cpp
```diff
@@ -56,13 +56,16 @@ Service* Service::find_by_pid(pid_t pid)
...
     if (!is_enabled() && enabled) {
         m_boot_modes.append(g_boot_mode);
     } else {
         m_boot_modes.remove_first_matching([&](auto& mode) { return mode == g_boot_mode; });
     }
     dbgln("Service::set_enabled_for_current_boot_mode({}) {}: {}", enabled, g_boot_mode, m_boot_modes.size());
+
+    auto config = Core::ConfigFile::get_for_system("SystemServer");
+    config->write_entry(name(), "BootModes", String::join(',', m_boot_modes));
```
bugaevc, line 48: I'm not sure I like the server modifying its own config file. Do our other servers do this?

A config file sounds like something the server reads, but never writes. It's the administrator who edits the file. Perhaps the `service` utility should write to the file directly, and then notify SystemServer about the changes? (That is also how `systemctl edit` works.)

Userland/Services/SystemServer/Service.cpp
```diff
@@ -175,7 +178,7 @@ void Service::deactivate()
     // Already stopped
     if (m_pid <= 0)
         return;
-    
+
```
bugaevc, line 161: Ditto

Userland/Utilities/service.cpp
sppmacd: This is removed in some next commit.

Userland/Services/SystemServer/Service.cpp
sppmacd: > No no no, you don't call `did_exit()` yourself, it gets called when you receive a notification from the kernel that it did, in fact, exit — via SIGCHLD, or potentially via other means.

This wasn't called then for some reason.

sppmacd: Should it be something like this?
```c++
Service::Service()
{
    ...
    m_sigterm_timer = Core::Timer::create_single_shot(5000, [this]() {
        if (m_pid == m_last_pid) {
            dbgln("Force killing service {} ({})", name(), m_pid);
            kill(m_pid, SIGKILL);
        }
    });
    ....
}
...
void Service::deactivate()
{
    ....
    m_sigterm_timer->start();
    kill(m_pid, SIGTERM);
    ....
}
```
(`m_last_pid` is set to `m_pid` in `did_exit` just before `m_pid` is set to -1)

bugaevc: `m_last_pid` doesn't sound too good. Can't we simply disarm the timer if the service exits before the time runs out?

Also, you have to check whether the `kill()` calls in fact succeed, and at least log an error otherwise.

sppmacd: Simply
```
m_sigterm_timer->stop();
m_pid = -1;
```
in did_exit and
```
kill(m_pid, SIGKILL);
```
in m_sigterm_timer?

bugaevc: Yeah I guess.

But again, please make very sure that killing a process and reaping it cannot race. That mainly means we should never reap from inside the `SIGCHLD` handler, only inside the event loop.

sppmacd:
I will close the PR for now. Maybe I will return to working on it, for now if someone wants to take over  it, please do :)

PR #6317, opened 2021-04-14
Browser+LibWeb: Add support for spoofing the browser user agent

IdanHo:
This is helpful when testing certain sites like twitter.com which display differently based on the user agent.


IdanHo:
Changes in the push: added a couple more default user agents (instead of just chrome), and added a "custom" option

linusg:
Random idea: do menu items support tooltips? Would be neat to show the exact UA string when hovering the various options :)

IdanHo:
> Random idea: do menu items support tooltips? Would be neat to show the exact UA string when hovering the various options :)

That would be very neat indeed, but it doesnt look like we support that right now (unless i missed something) :(

PR #6319, opened 2021-04-14
Userland: Implement `pls`, a sudo clone

Quaker762:
Well, more of a hybrid between `sudo` and `doas` from BSD (due to the power of unveil). This is a pretty basic clone for now, and should give us enough functionality to provide the user with some level of privilege elevation. 

Also note for now, due to the way the argument parser functions, any commands that require `-` arguments will need to be run as follows (e.g)

`pls -- ls -la`

Screenshot_2021-04-14_22-07-56


Quaker762:
Btw I still need to write a `man` page for this (probably tomorrow :^) )

nico:
@alimpfard needs to add a shell feature to allow writing a program's args before the program name (maybe `,`?: `ls foo` <=> `foo, ls`), so that this can be used as `whoami, pls`.

Userland/Utilities/pls.cpp
```diff
@@ -0,0 +1,196 @@
...
+#include <string.h>
+#include <unistd.h>
+
+static constexpr int line_buff_length = 256;
+
+// Function Definitions
+extern "C" int main(int arch, char** argv);
+int unveil_paths(const char*);
+
+// Unveil paths, given that the current
```
nico, line 46: THE SUSPENSE

danboid:
Nice work Q!

More importantly, have you seen AK's range of :^) clothing?

danboid:
https://xkcd.com/149/

Userland/Utilities/pls.cpp
Quaker762: I like to keep all of us on our toes

Quaker762:
> Nice work Q!
> 
> More importantly, have you seen AK's range of :^) clothing?

I have! Pretty meta!

Userland/Utilities/pls.cpp
```diff
@@ -0,0 +1,196 @@
...
+int unveil_paths(const char* command)
+{
+    // Get the system path, split it and attempt to unveil all the paths.
+    // We do NOT error out on an invalid path
+    auto paths = String(getenv("PATH")).split(':');
+    int num_unveils = 0;
+    char path_buf[256];
+
+    // Unveil each path
+    for (const auto path : paths) {
```
Quaker762, line 56: Whoops this should be `const auto&`

Userland/Utilities/pls.cpp
```diff
@@ -0,0 +1,196 @@
...
+        perror("pledge");
+        return 1;
+    }
+
+    if (seteuid(0) < 0) {
+        perror("seteuid");
+        return 1;
+    }
+
+    if (unveil("/etc", "r") < 0) {
```
Quaker762, line 92: Oops, this should probably just be `unveil("/etc/sudoers,...)`

Userland/Utilities/pls.cpp
```diff
@@ -0,0 +1,196 @@
...
+        // Ignore comments
+        if (buffer[0] == '#')
+            continue;
+
+        buffer[strcspn(buffer, "\n")] = '\0';
+        sudoers.append(String(buffer));
+    }
+
+    if (!sudoers.contains_slow(username)) {
+        warnln("{} is not in the sudoers file. This incident will be reported :^)", username);
```
g-w1, line 118: Where will this be reported? Or is it left in for the memes? :P

Userland/Utilities/pls.cpp
```diff
@@ -0,0 +1,196 @@
...
+}
+
+// <kling> linusg: quaker: "please" feels quite long, how about "pls" :P
+// <kling> "pls rm -r crap" has a nice ring to it
+// lol
+int main(int argc, char** argv)
+{
+    Vector<const char*> command;
+    Core::ArgsParser args_parser;
+    args_parser.add_positional_argument(command, "Name of command to run at elevated privilege level", "cmd");
```
linusg, line 78:
```suggestion
    args_parser.add_positional_argument(command, "Command to run at elevated privilege level", "command");
```

(IMO)

Userland/Utilities/pls.cpp
```diff
@@ -0,0 +1,196 @@
...
+    }
+
+    // The user was in the sudoers file, now let's ask for their password to ensure that it's actually them...
+    uid_t current_uid = getuid();
+    auto account_or_error = (username)
+        ? Core::Account::from_name(username)
+        : Core::Account::from_uid(current_uid);
+
+    if (account_or_error.is_error()) {
+        fprintf(stderr, "Core::Account::from_name: %s\n", account_or_error.error().characters());
```
linusg, line 129:
```suggestion
        warnln("Core::Account::from_name: {}", account_or_error.error());
```

Userland/Utilities/pls.cpp
```diff
@@ -0,0 +1,196 @@
...
+    auto unveil_count = unveil_paths(command.at(0));
+    if (unveil_count == 0) {
+        warnln("Failed to unveil paths!");
+        return 3;
+    }
+
+    unveil(nullptr, nullptr);
+
+    // Build the arguments list passed to `execvpe`
+    char* exec_args[command.size() + 1];
```
bgianfo, line 161: Don't use VLA's, use a Vector<String> instead?

Userland/Utilities/pls.cpp
```diff
@@ -0,0 +1,196 @@
...
+            warnln("Incorrect or disabled password.");
+            return 1;
+        }
+    }
+
+    // TODO: Support swapping users instead of just defaulting to root
+    setuid(0);
+    setgid(0);
+
+    // Unveil all paths in the user's PATH, as well as the command they've specified.
```
bgianfo, line 151: Can we move all of the processing to the top so we can lock the veil as soon as possible?

Userland/Utilities/pls.cpp
bcoles: `/etc/sudoers`, `/etc/passwd`, and `/etc/shadow`.

Userland/Utilities/pls.cpp
```diff
@@ -0,0 +1,196 @@
...
+        ? Core::Account::from_name(username)
+        : Core::Account::from_uid(current_uid);
+
+    if (account_or_error.is_error()) {
+        fprintf(stderr, "Core::Account::from_name: %s\n", account_or_error.error().characters());
+        return 1;
+    }
+
+    const auto& account = account_or_error.value();
+    if (getuid() != 0 && account.has_password()) {
```
bcoles, line 134: You already have the uid. No need to call `getuid()` again.

```suggestion
    if (current_uid != 0 && account.has_password()) {
```

Base/etc/sudoers
```diff
@@ -0,0 +1 @@
+anon
```
bcoles, line 1:
```suggestion
root
anon
```

Userland/Utilities/pls.cpp
bcoles: Perhaps eventually? We have `syslog()` but it is currently unused.

awesomekling: Please abstain from memeing in user-facing parts of the system.

Userland/Utilities/pls.cpp
Hendiadyoin1: or AK::Array, its constant size

Userland/Utilities/pls.cpp
Quaker762: Good idea, I'll add in the unveil for `/etc/sudoers` etcetera to this function too

Userland/Utilities/pls.cpp
Quaker762: Sorry I wasn't aware this was memeing as it gives the same output as sudo sans :^)

Userland/Utilities/pls.cpp
Quaker762: @bgianfo @Hendiadyoin1 Are `AK::Vector` and `AK::Array` both expected to be sequentially allocated?

Userland/Utilities/pls.cpp
Quaker762: Does this specifically print into `stderr`? I feel an error of this kind should be directed there.

bgianfo:
Consider adding a man page as well? (under `/Base/usr/share/man/man1`)

bcoles:
> 
> 
> Consider adding a man page as well? (under `/Base/usr/share/man/man1`)

https://github.com/SerenityOS/serenity/pull/6319#issuecomment-819532850

Userland/Utilities/pls.cpp
asynts: These functions work a lot like `std::format`. We got `dbgln` which writes debug output, `outln` which writes to `stdout` and `warnln` which writes to `stderr`. There are also a few variants like `out` and `warn` which do not append newlines. Additionally, there is `dmesgln` which is for the kernel.

Userland/Utilities/pls.cpp
```diff
@@ -0,0 +1,196 @@
...
+
+        if (!account.authenticate(password.value().characters())) {
+            warnln("Incorrect or disabled password.");
+            return 1;
+        }
+    }
+
+    // TODO: Support swapping users instead of just defaulting to root
+    setuid(0);
+    setgid(0);
```
ADKaster, line 149: This PR seems like quite a party 🎉. I believe you're supposed to set group before user, normally 😁.

Userland/Utilities/pls.cpp
```diff
@@ -0,0 +1,196 @@
...
+    auto sudoers_fd = fopen("/etc/sudoers", "r");
+    if (!sudoers_fd) {
+        warnln("couldn't open /etc/sudoers!");
+        return 1;
+    }
+
+    // Okay, so we have a handle to /etc/sudoers, now let's read it in line by line to get a
+    // list of all of the users contained within
+    char buffer[line_buff_length];
+    while (fgets(buffer, line_buff_length, sudoers_fd)) {
```
ADKaster, line 108: use_core

Quaker762: REEEEEE

alimpfard: Did you know you can get a line iterator from a Core::File?
It had issues with streams, but it should probably work just fine for normal files, you could use that too
`for (auto it = file.line_begin(); it < file.line_end(); ++it)`

(I wish there was a way to make the ranged loops call some other method...)

Quaker762:
Alright! All tidied up and changes made! Everyone is about to get a rereview notification, so enjoy!

Userland/Utilities/pls.cpp
Quaker762: `for (auto line = sudoers_file_or_error.value()->line_begin(); !line.at_end(); ++line)` seems to more specifically worked for me in this instance.

Userland/Utilities/pls.cpp
danboid: Popcorn, anyone?

alimpfard:
Other than that, LGTM.

Documentation/BuildInstructions.md
```diff
@@ -209,6 +209,8 @@ $ ninja run
 
 Note that the `anon` user is able to become `root` without password by default, as a development convenience.
 To prevent this, remove `anon` from the `wheel` group and he will no longer be able to run `/bin/su`.
+`anon` is also, by default, located in `/etc/sudoers`, meaning that they will be able to execute with root permission using `pls`.
+To prevent this, please remove them `/etc/sudoers`.
```
linusg, line 213:
```suggestion
To prevent this, remove them from `/etc/sudoers`.
```

Userland/Utilities/pls.cpp
```diff
@@ -0,0 +1,220 @@
...
+}
+
+// <kling> linusg: quaker: "please" feels quite long, how about "pls" :P
+// <kling> "pls rm -r crap" has a nice ring to it
+// lol
+int main(int argc, char** argv)
+{
+    Vector<const char*> command;
+    Core::ArgsParser args_parser;
+    args_parser.add_positional_argument(command, "Name of command to run at elevated privilege level", "command");
```
linusg, line 78: I still think this should say "command", not "name of command" - at least in my view, the "name" of a command with arguments is just the binary...

Userland/Utilities/pls.cpp
```diff
@@ -0,0 +1,220 @@
...
+        perror("pledge");
+        return 1;
+    }
+
+    if (seteuid(0) < 0) {
+        perror("seteuid");
+        return 1;
+    }
+
+    // Unveil sudoers file
```
linusg, line 92: This comment (and the ones below) don't add any value. I can already see that from the function name (`unveil`) and argument (`/etc/sudoers`) :^)

Userland/Utilities/pls.cpp
```diff
@@ -0,0 +1,220 @@
...
+
+    // The user was in the sudoers file, now let's ask for their password to ensure that it's actually them...
+    uid_t current_uid = getuid();
+    auto account_or_error = (username)
+        ? Core::Account::from_name(username)
+        : Core::Account::from_uid(current_uid);
+
+    if (account_or_error.is_error()) {
+        fprintf(stderr, "Core::Account::from_name: %s\n", account_or_error.error().characters());
+        warnln("Core::Account::from_name: {}", account_or_error.error().characters());
```
linusg, line 162: You can directly pass a string to warnln, no need for `.characters()`. Also, this logs twice, the above line still needs to be removed!

linusg (changes requested):

Userland/Utilities/pls.cpp
Quaker762: Yeah fair point, I'll change it.

Userland/Utilities/pls.cpp
Quaker762: Err, yeah, I missed that one whoops D:

Documentation/BuildInstructions.md
Quaker762: `pls remove_from /etc/sudoers` :^)

linusg (approved):

Meta/build-root-filesystem.sh
```diff
@@ -55,6 +55,7 @@ chmod 0400 mnt/res/kernel.map
 chmod 0400 mnt/boot/Kernel
 chmod 4750 mnt/bin/su
 chmod 4755 mnt/bin/passwd
+chmod 4755 mnt/bin/pls
```
bcoles, line 58: Not a big deal, but read permissions for everyone isn't necessary - only execute:

```suggestion
chmod 4751 mnt/bin/pls
```

In the future, it may be a good idea to also add a `sudo` or `pls` group. `/bin/pls` could then have `4750` permissions and `root:pls` ownership to prevent execution by users not in the correct group.

Base/usr/share/man/man8/pls.md
```diff
@@ -0,0 +1,41 @@
...
+$ pls [command]
+```
+
+## Description
+
+Executes a command as the root user (uid and gid 0), given that the user executing `pls` is located in
+the sudoers file. 
+
+It is possible to execute commands that contain hyphenated options via the use of `--`, which signifies the. 
+end of command options. For example:
```
bcoles, line 17:
```suggestion
It is possible to execute commands that contain hyphenated options via the use of `--`, which signifies
the end of command options. For example:
```

Base/usr/share/man/man8/pls.md
```diff
@@ -0,0 +1,41 @@
...
+## Synopsis
+
+```**sh
+$ pls [command]
+```
+
+## Description
+
+Executes a command as the root user (uid and gid 0), given that the user executing `pls` is located in
+the sudoers file.
```
bcoles, line 14:
```suggestion
the sudoers file.
```

Quaker762: > Me nitpicking about this at work and doing it off the clock
https://www.youtube.com/watch?v=INCSBwpJBQk

Meta/build-root-filesystem.sh
Quaker762: Alright I'll get that changed. I'll keep iterating on this once it's in

alimpfard:
Something to think about, currently, /etc/sudoers is readable, but I don't think that's the case in (at least) most Linux distros.
Though I'm not sure if that provides any meaningful security (what can you do with knowing who can sudo?)

bcoles (approved):

alimpfard (approved):

Quaker762:
> Something to think about, currently, /etc/sudoers is readable, but I don't think that's the case in (at least) most Linux distros.
> Though I'm not sure if that provides any meaningful security (what can you do with knowing who can sudo?)

Some justifications are here: https://unix.stackexchange.com/questions/494435/why-is-etc-sudoers-not-world-readable

Perhaps we `chmod` `/etc/sudoers` to a more appropriate permission set?

bcoles:
> 
> 
> Something to think about, currently, /etc/sudoers is readable, but I don't think that's the case in (at least) most Linux distros.
> Though I'm not sure if that provides any meaningful security (what can you do with knowing who can sudo?)

I was drafting up something on the topic. Currently the sudoers file only contains usernames. Allowing this information to be read doesn't pose much of an issue.

However...

A potential hardening measure is to ensure the `/etc/sudoers` file has permissions `600` and is owned by root. Otherwise exit with an error message.

This defense is enforced by modern `sudo` and has some merit to hinder privilege escalation in some scenarios. 

Often in arbitrary file write scenarios it is possible to control the file path but only part of the data which is written to file. This is extremely common in situations where a suid binary writes data to a user-specified file (for example, the XOrg privesc logfile bug from about a year ago). In these situations, it is possible to control only some of the data written to file. Because the sudoers file parsing is quite lax, all junk data is ignored. An attacker `hacker` needs only to be able to inject "\nhacker\n" anywhere in the file in order to gain root privileges with `pls`.

Similarly, ensuring that sudoers has permissions `600` will hinder exploitation via arbitrary `chown` vulnerabilities.

Typically an attacker will choose other methods, like overwriting LDPRELOAD (which we don't support), overwriting cron jobs (which we don't support, and cron have tightened up their cron job parsing to prevent these attacks), or overwriting shellrc scripts (which requires waiting until the user logs in to execute the privesc payload, rather than instant gratification).

Quaker762:
> > Something to think about, currently, /etc/sudoers is readable, but I don't think that's the case in (at least) most Linux distros.
> > Though I'm not sure if that provides any meaningful security (what can you do with knowing who can sudo?)
> 
> I was drafting up something on the topic. Currently the sudoers file only contains usernames. Allowing this information to be read doesn't pose much of an issue.
> 
> However...
> 
> A potential hardening measure is to ensure the `/etc/sudoers` file has permissions `600` and is owned by root. Otherwise exit with an error message.
> 
> This defense is enforced by modern `sudo` and has some merit to hinder privilege escalation in some scenarios.
> 
> Often in arbitrary file write scenarios it is possible to control the file path but only part of the data which is written to file. This is extremely common in situations where a suid binary writes data to a user-specified file (for example, the XOrg privesc logfile bug from about a year ago). In these situations, it is possible to control only some of the data written to file. Because the sudoers file parsing is quite lax, all junk data is ignored. An attacker `hacker` needs only to be able to inject "\nhacker\n" anywhere in the file in order to gain root privileges with `pls`.
> 
> Similarly, ensuring that sudoers has permissions `600` and owned by `root` will hinder exploitation via arbitrary `chown` / `chmod` vulnerabilities.
> 
> Typically an attacker will choose other methods, like overwriting LDPRELOAD (which we don't support), overwriting cron jobs (which we don't support, and cron have tightened up their cron job parsing to prevent these attacks), or overwriting shellrc scripts (which requires waiting until the user logs in to execute the privesc payload, rather than instant gratification).

Good enough for me, I'll `chmod` the sudoers file during `build-root-filesystem`

Quaker762:
I'm adding changes w.r.t @bcoles comments as a separate commit so we have a log of the discussion.

Base/usr/share/man/man8/pls.md
```diff
@@ -0,0 +1,41 @@
...
+
+It is possible to execute commands that contain hyphenated options via the use of `--`, which signifies the
+end of command options. For example:
+
+```sh
+$ pls -- ls -la
+```
+
+## Files
+/etc/sudoers - List of users that can run `pls`
```
awesomekling, line 24: Why is the file called "sudoers" if we're not using the "sudo" name?

Userland/Utilities/pls.cpp
```diff
@@ -0,0 +1,213 @@
...
+        if (unveil(command_path, "x") == 0)
+            num_unveils++;
+    }
+
+    return num_unveils;
+}
+
+// <kling> linusg: quaker: "please" feels quite long, how about "pls" :P
+// <kling> "pls rm -r crap" has a nice ring to it
+// lol
```
awesomekling, line 73: Please tone down the memeing.

Userland/Utilities/pls.cpp
```diff
@@ -0,0 +1,213 @@
...
+    }
+
+    // Lock veil
+    unveil(nullptr, nullptr);
+
+    const char* username = getlogin();
+    auto sudoers_file_or_error = Core::File::open("/etc/sudoers", Core::IODevice::ReadOnly);
+    bool user_found = false;
+    if (sudoers_file_or_error.is_error()) {
+        warnln("couldn't open /etc/sudoers!");
```
awesomekling, line 143: There's an error string in `sudoers_file_or_error.error()`, let's add it to the warning.

Userland/Utilities/pls.cpp
```diff
@@ -0,0 +1,213 @@
...
+        }
+
+        if (!account.authenticate(password.value().characters())) {
+            warnln("Incorrect or disabled password.");
+            return 1;
+        }
+    }
+
+    // TODO: Support swapping users instead of just defaulting to root
+    setgid(0);
```
awesomekling, line 193: What if this fails?

Userland/Utilities/pls.cpp
```diff
@@ -0,0 +1,213 @@
...
+    setuid(0);
```
awesomekling, line 194: What if this fails?

Userland/Utilities/pls.cpp
```diff
@@ -0,0 +1,213 @@
...
+    builder.clear();
+
+    // Build SUDO_COMMAND envvar
+    builder.append("SUDO_COMMAND=");
+    builder.append(command.at(0));
+    auto sudo_command = builder.build();
+    builder.clear();
+
+    const char* envs[] = { "PROMPT=\\X\\u@\\h:\\w\\a\\e[33;1m\\h\\e[0m \\e[34;1m\\w\\e[0m \\p ",
+        "TERM=xterm", "PAGER=more", "PATH=/bin:/usr/bin:/usr/local/bin",
```
awesomekling, line 221: It doesn't seem appropriate for this program to hard-code `PROMPT`, `TERM` and `PAGER`.
For `TERM`, we should probably just pass through whatever the current `TERM` is.
Do we need the other two at all?

Userland/Utilities/pls.cpp
```diff
@@ -0,0 +1,213 @@
...
+        sudo_user.characters(), sudo_command.characters(), nullptr };
+
+    // Execute the desired command
+    int rc = execvpe(command.at(0), const_cast<char**>(exec_args.data()), const_cast<char**>(envs));
+    if (rc < 0) {
```
awesomekling, line 226: Since you don't need the `rc` value for anything, I would suggest tightening its scope:
```cpp
if (execvpe(...) < 0)) {
```

Userland/Utilities/pls.cpp
```diff
@@ -0,0 +1,213 @@
...
+        }
+    }
+
+    // TODO: Support swapping users instead of just defaulting to root
+    setgid(0);
+    setuid(0);
+
+    // Build the arguments list passed to `execvpe`
+    Vector<const char*> exec_args;
+    for (size_t i = 0; i < command.size(); i++) {
```
awesomekling, line 198: Use range-for:
```cpp
for (auto& arg : command) {
    ...
```

Userland/Utilities/pls.cpp
```diff
@@ -0,0 +1,213 @@
...
+    if (!user_found) {
+        warnln("{} is not in the sudoers file!", username);
+        return 2;
+    }
+
+    // The user was in the sudoers file, now let's ask for their password to ensure that it's actually them...
+    uid_t current_uid = getuid();
+    auto account_or_error = (username)
+        ? Core::Account::from_name(username)
+        : Core::Account::from_uid(current_uid);
```
awesomekling, line 171: Under what circumstances would `username` be null here?
And if we can get here with `!username`, what did we even check in the sudoers file?

awesomekling (changes requested):

Userland/Utilities/pls.cpp
Quaker762: Hrmm, I believe this block of code was taken from `passwd.cpp` when I was initially working out how to use LibCore.

I'll reactor this then look at cleaning that up too.

Base/usr/share/man/man8/pls.md
Quaker762: It made sense at the time because I couldn't think of a better name, though I agree it would confuse people that now expect `sudo` to work on the system.

Suggestions for a better? "Pleasers" doesn't quite have the same ring to it. Perhaps `plsusers` or just `pls.conf`?

Userland/Utilities/pls.cpp
Quaker762: Whoops missed this one when I removed the other one...

Userland/Utilities/pls.cpp
Quaker762: @awesomekling The `sudo user` and `sudo command` is a  feature that comes included with `sudo`. This way we can see who invoked `SUDO`. According to the Ubuntu user manual:

```
       If sudo is run by root and the SUDO_USER environment variable is set, the sudoers policy
       will use this value to determine who the actual user is.  This can be used by a user to
       log commands through sudo even when a root shell has been invoked.  It also allows the -e
       option to remain useful even when invoked via a sudo-run script or program.  Note,
       however, that the sudoers lookup is still done for root, not the user specified by
       SUDO_USER.
```

Userland/Utilities/pls.cpp
```diff
@@ -0,0 +1,213 @@
...
+    char path_buf[256];
+
+    // Unveil each path
+    for (const auto& path : paths) {
+        if (unveil(path.characters(), "x") == 0)
+            num_unveils++;
+    }
+
+    // Now unveil the command
+    auto command_path = realpath(command, &path_buf[0]);
```
awesomekling, line 62: This looks extremely suspicious! `path_buf` is 256 bytes, but `realpath` might write up to `PATH_MAX` byte into it, if I understand correctly.

This attempt to unveil everything in `PATH` seems rather odd in general.

awesomekling: If anything, we could resolve the command path *before* calling unveil, and then unveiling only that one path.

Quaker762: > This attempt to unveil everything in PATH seems rather odd in general.

This actually came from the source of `doas` when I doing some research on how they handled unveil for this stuff. I think the command path is probably a lot safer, I'll get that done instead :P

Userland/Utilities/pls.cpp
```diff
@@ -32,43 +32,46 @@
...
-
-    // Unveil each path
-    for (const auto& path : paths) {
-        if (unveil(path.characters(), "x") == 0)
-            num_unveils++;
-    }
 
-    // Now unveil the command
-    auto command_path = realpath(command, &path_buf[0]);
+    // Attempt to unveil the command first...
```
Quaker762, line 52: Let's see how we go with this. Not entirely sure if this is considered "awkward" in how I'm doing it, though it should take into consideration most cases (i.e someone supplies a direct path e.g `pls ./quake` or a system program `pls whoami`)

linusg (changes requested):

Userland/Utilities/pls.cpp
```diff
@@ -0,0 +1,248 @@
...
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
```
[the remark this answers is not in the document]
Quaker762: Done :)

awesomekling (changes requested):
There's a concerning lack of attention to detail in this PR, for something this security-sensitive.

Userland/Utilities/pls.cpp
```diff
@@ -0,0 +1,228 @@
...
+        exec_args.append(arg);
+    }
+
+    // Always terminate with a NULL (to signal end of args list)
+    exec_args.append(nullptr);
+
+    // Build the environment arguments
+    StringBuilder builder;
+
+    // Build SUDO_USER envvar
```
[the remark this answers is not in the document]
bcoles: > Also why is it necessary?

There are instances where a program invoked by `pls` and running as `root` may wish to know the original user ID (usually for logging/auditing, instead of logging all actions as `root`).

`sudo` does something similar:

```
$ whoami && sudo sh -c 'echo $SUDO_USER'
user
user
```

As an example, consider a `sudo` configuration which permits users to run only a limited set of commands, one of which is a program which requires elevated privileges and stores information related to the user. This program has no method to determine the user name/uid/gid/etc of the calling user, only the name/uid/gid/etc of `root`.

Currently, unlike `sudo`, `pls` does not support the same level of control over which commands can be executed by authorised users - it allows either all or nothing. Presumably eventually the `plsusers` configuration will be updated to allow this level of granularity.

Related:

* https://unix.stackexchange.com/questions/190221/sudo-user-undefined-in-sudo-manual-clarification-would-be-lovely
* https://stackoverflow.com/questions/10272784/how-do-i-get-the-users-real-uid-if-the-program-is-run-with-sudo

Quaker762:
> There's a concerning lack of attention to detail in this PR, for something this security-sensitive.

I'm going to refrain from opening multiple PRs of such large volume in the future as it's obviously impossible to maintain code quality across all of them.

Userland/Utilities/pls.cpp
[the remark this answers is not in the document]
Quaker762: I'm removing these for now until we have something a bit more solid.

Userland/Utilities/pls.cpp
```diff
@@ -198,13 +203,23 @@ int main(int argc, char** argv)
...
+    char* env_term = getenv("TERM");
+    builder.append("TERM=");
+
+    if (env_term == NULL) {
+        builder.append("xterm");
+    } else {
+        builder.append(env_term);
+    }
+    auto term = builder.build();
+    builder.clear();
```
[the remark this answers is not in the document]
Quaker762: In that instance, would `TERM` just not be passed in at all via a blank string? From the previous comment: 
`For TERM, we should probably just pass through whatever the current TERM is.`

Userland/Utilities/pls.cpp
```diff
@@ -0,0 +1,228 @@
...
+        if (unveil(command_path, "x") == 0)
+            num_unveils++;
+
+        free(command_path);
+        return num_unveils;
+    }
+
+    // Okay, so we couldn't find the actual file specified by the user, let's
+    // instead search PATH for it...
+    auto command_path_system = Core::find_executable_in_path(command);
```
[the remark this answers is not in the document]
Quaker762: Sorry I don't quite understand what you mean here. Will `Core::find_executable_in_path()` not work before I unveil directories in the path? `which` seems to use this without any calls to unveil? Unveil is only ever gets called once here. 

The first check in this function is to check for a local command, i.e `pls ./sdlquake2` that can be resolved via `realpath`. If that fails, we then use `find_executable_in_path` to get the full path of the command typed in, if any, as something in the path has probably been specified, i.e `pls whoami`, `pls ls`. Perhaps I'm misunderstanding the correct order of operations that should be occurring here. 

Also the return value/use of `num_unveils` is misleading. I've changed this to `bool` to signify that it either unveils the command or it doesn't, meaning it's probably non-existent.

Userland/Utilities/pls.cpp
```diff
@@ -0,0 +1,222 @@
...
+    exec_args.append(nullptr);
+
+    // Build the environment arguments
+    StringBuilder builder;
+    Vector<String> env_args_str;
+
+    // TERM envvar
+    char* env_term = getenv("TERM");
+
+    if (env_term == nullptr) {
```
Quaker762, line 201: Oops, I'll get this fixed right now

Userland/Utilities/pls.cpp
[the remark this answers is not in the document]
awesomekling: After you have called `unveil()` at least once, your view of file system is now constrained to whatever paths you have unveiled. In this program, you first `unveil` a handful of paths in `main()`, and then call `unveil_paths()`.

This means that `Core::find_executable_in_path()` cannot see into `/bin`, for example. Since the process has a file system veil, the only syscall that can see `/bin` is a call to `unveil()` to unveil `/bin`. If you do that, `/bin` is now visible to all other syscalls as well.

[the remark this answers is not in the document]
Quaker762: So it would be best to `unveil` the path to the program first, unveil `/etc/` to unveil the files required (`plsusers`, `passwd` etc), then lock the veil?

Meta/build-root-filesystem.sh
```diff
@@ -50,19 +50,24 @@ chmod 660 mnt/etc/WindowServer.ini
 chown $window_uid:$window_gid mnt/etc/WindowServer.ini
 echo "/bin/sh" > mnt/etc/shells
 
+chmod 0400 mnt/etc/plsusers
+chown 0:0 mnt/etc/plsusers
```
bcoles, line 54:
```suggestion
chown 0:0 mnt/etc/plsusers
```

Userland/Utilities/pls.cpp
bugaevc: > `path_buf` is 256 bytes, but `realpath` might write up to `PATH_MAX` byte into it

TRIGGERED 🤬

Never, ever use `realpath()` with a preallocated buffer. @BenWiederhake has done a fantastic job at fixing it in https://github.com/SerenityOS/serenity/pull/4965; so you can use `realpath(foo, nullptr)` and get a properly sized buffer, or `Core::File::real_path_for(foo)` and get a `String`.

P.S. Thanks for linking me here 😂

alimpfard:
Tracked in #7567.

Meta/build-root-filesystem.sh
alimpfard: Resolved in #7567.

Issue #5380, opened 2021-02-16
FontPicker: Ability to pick arbitrary sizes for TrueType fonts


danboid:
Currently, FontPicker only gives the user a limited number of font sizes to choose from, between 8 and 36. Ideally the user would be able to enter their desired size of TrueType fonts within the FontPicker window.

SHUBHANSHU-creator:
I would like to work on this issue

linusg:
Implemented in #6320.

PR #6320, opened 2021-04-14
LibGUI: Allow arbitrary font size for TTF fonts in FontPicker

JTriantafylos:
This commit adds a SpinBox to the FontPicker size selection portion to allow users to set arbitrary font sizes (1 to 36 inclusive currently) for TTF fonts.

This enhancement was suggested in #5380.

When a font size is set in the SpinBox and that font size exists in the size list, the size will be selected in the list:

FontPicker_size_SpinBox

Otherwise, if the font size set in the SpinBox does not exist in the size list, selection will be disabled for the list:

FontPicker_size_SpinBox2

The SpinBox is also hidden when a fixed size font is selected:

FontPicker_size_SpinBox3

Note: This is my first commit to any open source project so any feedback is greatly appreciated.

### DIFF
--- a/Userland/Libraries/LibGUI/FontPicker.cpp
+++ b/Userland/Libraries/LibGUI/FontPicker.cpp
@@ -32,6 +32,7 @@
 #include <LibGUI/Label.h>
 #include <LibGUI/ListView.h>
 #include <LibGUI/Scrollbar.h>
+#include <LibGUI/SpinBox.h>
 #include <LibGUI/Widget.h>
 #include <LibGfx/FontDatabase.h>
 
@@ -56,6 +57,9 @@ FontPicker::FontPicker(Window* parent_window, const Gfx::Font* current_font, boo
     m_weight_list_view = *widget.find_descendant_of_type_named<ListView>("weight_list_view");
     m_weight_list_view->set_model(adopt(*new FontWeightListModel(m_weights)));
     m_weight_list_view->horizontal_scrollbar().set_visible(false);
+
+    m_size_spin_box = *widget.find_descendant_of_type_named<SpinBox>("size_spin_box");
+    m_size_spin_box->set_range(1, 255);
 
     m_size_list_view = *widget.find_descendant_of_type_named<ListView>("size_list_view");
     m_size_list_view->set_model(ItemListModel<int>::create(m_sizes));
@@ -93,6 +97,7 @@ FontPicker::FontPicker(Window* parent_window, const Gfx::Font* current_font, boo
     };
 
     m_weight_list_view->on_selection = [this](auto& index) {
+        bool font_is_fixed_size = false;
         m_weight = index.data(ModelRole::Custom).to_i32();
         m_sizes.clear();
         dbgln("Selected weight: {}", m_weight.value());
@@ -100,11 +105,16 @@ FontPicker::FontPicker(Window* parent_window, const Gfx::Font* current_font, boo
             if (m_fixed_width_only && !typeface.is_fixed_width())
                 return;
             if (typeface.family() == m_family.value() && (int)typeface.weight() == m_weight.value()) {
-                if (typeface.is_fixed_size()) {
+                font_is_fixed_size = typeface.is_fixed_size();
+                if (font_is_fixed_size) {
+                    m_size_spin_box->set_visible(false);
+
                     typeface.for_each_fixed_size_font([&](auto& font) {
                         m_sizes.append(font.presentation_size());
                     });
                 } else {
+                    m_size_spin_box->set_visible(true);
+
                     m_sizes.append(8);
                     m_sizes.append(10);
                     m_sizes.append(12);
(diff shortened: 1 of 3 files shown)


danboid:
Why a limit of 36?

JTriantafylos:
> Why a limit of 36?

Frankly I don't see why it couldn't be changed to have a higher limit.
I just used 36 as that is the current maximum able to be selected through the size list.

Userland/Libraries/LibGUI/FontPicker.cpp
```diff
@@ -57,6 +58,9 @@ FontPicker::FontPicker(Window* parent_window, const Gfx::Font* current_font, boo
     m_weight_list_view->set_model(adopt(*new FontWeightListModel(m_weights)));
     m_weight_list_view->horizontal_scrollbar().set_visible(false);
 
+    m_size_spin_box = widget.find_descendant_of_type_named<SpinBox>("size_spin_box");
+    m_size_spin_box->set_range(1, 36);
```
linusg, line 62: A limit of 36 indeed seems rather arbitrary - a limit itself is probably a good idea but how about a higher (still arbitrary) number somewhere > 100?

@danboid what would you suggest?

Userland/Libraries/LibGUI/FontPicker.cpp
```diff
@@ -93,18 +97,24 @@ FontPicker::FontPicker(Window* parent_window, const Gfx::Font* current_font, boo
     };
 
     m_weight_list_view->on_selection = [this](auto& index) {
+        bool m_font_is_fixed_size;
```
linusg, line 100: The `m_` prefix should only be used for member variables, this would be just `font_is_fixed_size`. Please also initialize primitive types, i.e.:

```suggestion
        bool font_is_fixed_size = false;
```

linusg (changes requested):

Userland/Libraries/LibGUI/FontPicker.cpp
danboid: LO Writer seems to max out at 999.9 so maybe we should go with that too?

JTriantafylos:
@linusg Thank you for the advice. After taking a second look at this I have realized I started my initial commit title with "Libraries:" when "LibGUI:" would likely be more appropriate.
Do you have any advice on correcting this before I go further with your requested changes?

linusg:
Sure - just amend the previous commit or do an interactive rebase (https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History), both will allow you to change the title. You'll have to use either of these to incorporate code changes into the current commit :)

Userland/Libraries/LibGUI/FontPicker.cpp
JTriantafylos: It looks like font sizes greater than 255 are not supported by `LibGFX::Font` since `LibGFX::Font::presentation_size()` is defined to only return a `u8`.
This seems to cause the maximum safe font size to be 255 which is what I propose we set the maximum to.
Thoughts?

danboid: 255 will be fine. I've rarely used fonts greater than ~150 when doing WP/DTP sorta stuff so 255 should be fine for most cases.

Userland/Libraries/LibGUI/FontPicker.cpp
JTriantafylos: Thanks Linus, I have added these changes to the amended commit where I also corrected the commit title/message.

Userland/Libraries/LibGUI/FontPicker.cpp
```diff
@@ -119,18 +129,44 @@ FontPicker::FontPicker(Window* parent_window, const Gfx::Font* current_font, boo
...
+    m_size_spin_box->on_change = [this](int value) {
+        m_size = value;
+
+        Optional<size_t> index_of_new_size_in_list = m_sizes.find_first_index(m_size.value());
+
+        m_size_list_view->set_selection_mode(GUI::AbstractView::SelectionMode::SingleSelection);
+        if (index_of_new_size_in_list.has_value())
+            m_size_list_view->set_cursor(m_size_list_view->model()->index(index_of_new_size_in_list.value()), GUI::AbstractView::SelectionUpdate::Set);
+        else
+            m_size_list_view->set_selection_mode(GUI::AbstractView::SelectionMode::NoSelection);
```
linusg, line 168:
```suggestion
        if (index_of_new_size_in_list.has_value()) {
            m_size_list_view->set_selection_mode(GUI::AbstractView::SelectionMode::SingleSelection);
            m_size_list_view->set_cursor(m_size_list_view->model()->index(index_of_new_size_in_list.value()), GUI::AbstractView::SelectionUpdate::Set);
        } else {
            m_size_list_view->set_selection_mode(GUI::AbstractView::SelectionMode::NoSelection);
        }
```

linusg (commented):
Awesome, thanks for making these changes! One more request: please squash the second commit into the first one as per the [contribution guidelines](https://github.com/SerenityOS/serenity/blob/master/CONTRIBUTING.md). No need to preserve such code review changes in a separate commit :)

JTriantafylos:
> Awesome, thanks for making these changes! One more request: please squash the second commit into the first one as per the [contribution guidelines](https://github.com/SerenityOS/serenity/blob/master/CONTRIBUTING.md). No need to preserve such code review changes in a separate commit :)

Okay I believe I did everything correctly with squashing those two commits into one.
Thank you so much for your help through all of this :)

PR #6287, opened 2021-04-13
LibC: getaddrinfo: Set addrinfo sin_port to 0 if service arg is NULL

bcoles:
This can be tested with the `links` port with `g http://example.com/` (or any host which requires lookup - ie, a domain not an IP address).

Prior to this PR, hostname resolution would crash in `getservbyname()` due to a null pointer. This PR adds a guard clause for null `service` argument to `getservbyname()`, and ensures `getaddrinfo()` sets the port to zero in each returned `addrinfo` struct if `service` is null.

```
CrashDaemon(15): New coredump file: /tmp/coredump/links_27_1618287824
CrashDaemon(15): --- Backtrace for thread #0 (TID 27) ---
CrashDaemon(15): 0x32229b30: [libc.so] strtoll +0x20 (stdlib.cpp:55)
CrashDaemon(15): 0x32229fd0: [libc.so] strtol +0x20 (stdlib.cpp:905)
CrashDaemon(15): 0x3220feec: [libc.so] getaddrinfo +0x39c (netdb.cpp:706)
CrashDaemon(15): 0x0188b2fb: [/usr/local/bin/links] do_real_lookup +0xdb (dns.c:306)
CrashDaemon(15): 0x0188b4fc: [/usr/local/bin/links] lookup_fn +0x2c (dns.c:460)
CrashDaemon(15): 0x018b9e11: [/usr/local/bin/links] start_thread +0x91 (os_dep.c:2658)
CrashDaemon(15): 0x0188b5d4: [/usr/local/bin/links] find_host_no_cache +0xb4 (dns.c:492)
CrashDaemon(15): 0x0188b714: [/usr/local/bin/links] find_host +0x64 (dns.c:624)
CrashDaemon(15): 0x01885d17: [/usr/local/bin/links] make_connection +0x1f7 (connect.c:246)
CrashDaemon(15): 0x018a8dca: [/usr/local/bin/links] http_func +0x4a (http.c:181)
CrashDaemon(15): 0x018bb86b: [/usr/local/bin/links] run_connection +0xfb (sched.c:538)
CrashDaemon(15): 0x018bbb87: [/usr/local/bin/links] try_connection +0x67 (sched.c:616)
CrashDaemon(15): 0x018bbd87: [/usr/local/bin/links] check_queue +0x1e7 (sched.c:677)
CrashDaemon(15): 0x018be1c5: [/usr/local/bin/links] check_bottom_halves +0x65 (select.c:227)
[...]

PR #6324, opened 2021-04-14
[WIP] Port FFmpeg/libav* to Serenity!

06needhamt:
This PR contains my current efforts to port FFmpeg/libav* to SerenityOS. 

This is the first step to me achieving my dream of implementing a simple audio/video player for Serenity. In the end, I would like to have all the video decoding done natively without any external dependencies.

The Port is not currently functional due to some issues with format strings which I hope to resolve soon.

Any comments on my work so far would be greatly appreciated


Ports/FFmpeg/package.sh
```diff
@@ -0,0 +1,16 @@
...
+makeopts="HOSTCC=${CC} HOSTLD=${CC}"
+depends="SDL2"
+
+configure() {
+	run ./configure --enable-cross-compile --target-os=serenity --arch=i686 \
+	    --enable-ffmpeg --enable-ffprobe --enable-ffplay --cross-prefix=${CC%gcc} \
+	    --disable-txtpages --disable-manpages --disable-podpages \
+	    --disable-static --enable-shared \
+	    --enable-debug --disable-stripping # Remove Before PR!
+}
```
bcoles, line 16: You should be able to move all of these into a `configopts` parameter.

```suggestion
configopts="--enable-cross-compile --enable-ffmpeg --enable-ffprobe --enable-ffplay --cross-prefix=${CC%gcc} --disable-txtpages --disable-manpages --disable-podpages --disable-static --enable-shared --enable-debug --disable-stripping"
```

`configopts` is already defined in `.port_include.sh` and sets the `host` flag automatically:

```
--host="${SERENITY_ARCH}-pc-serenity"
```

I think you can also drop the ` --target-os=serenity --arch=i686` flags. This should be picked up by the configure script automatically (maybe).

Also, unless generating the man pages is causing issues, you can remove `--disable-manpages`. It doesn't hurt to have the man pages installed. Many other ports do this, and maybe we'll one day have a port of `man` to read man pages in groff format.

Userland/Libraries/LibDebug/Dwarf/DIE.cpp
```diff
@@ -139,6 +139,22 @@ DIE::AttributeValue DIE::get_attribute_value(AttributeDataForm form,
...
+        break;
+    }
+    case AttributeDataForm::Ref8: {
+        u64 data;
+        debug_info_stream >> data;
+            VERIFY(!debug_info_stream.has_any_error());
+        value.type = AttributeValue::Type::DieReference;
+        value.data.as_u32 = data + m_compilation_unit.offset(); // FIXME: Do we need to properly handle 64-Bit values?
+        break;
+    }
```
bcoles, line 157: > FIXME: Do we need to properly handle 64-Bit values?

No, we don't. #6036

If you're interested in implementing this, there's a patch here which adds support for 64 bit values to LibDebug.

https://github.com/SerenityOS/serenity/pull/6287#issuecomment-818505818

However, I found that this generated corrupted ELF crash dump files, so I never merged it investigated further.

Userland/Libraries/LibC/sys/socket.cpp
```diff
@@ -112,6 +112,12 @@ ssize_t recvfrom(int sockfd, void* buffer, size_t buffer_length, int flags, stru
     return rc;
 }
 
+int getnameinfo(const struct sockaddr*, socklen_t, char*, socklen_t, char*, socklen_t, int)
+{
+    // FIXME: Actually implement this
```
Nicholas-Baron, line 117:
```suggestion
    VERIFY_NOT_REACHED();
```
More like this?

linusg: if anything, more like this:

```suggestion
    TODO();
```

PR #6329, opened 2021-04-14
Tests: fixed test-crypto wrong argument order that failed - run-tests-and-shutdown.sh

DanielEliad:
run-tests-and-shutdown.sh failed on startup of `test-crypto` because of wrong parameter order
with the new parameter order the test actually runs (and succeeds)
:)


awesomekling:
Oopsie!

PR #6330, opened 2021-04-14
LibGfx+LibWeb: CanvasRenderingContext2D Path fixes and tweaks

IdanHo:
Several fixes to CanvasRenderingContext2D i found while trying to get MDN's simple breakout tutorial game to run in Serenity's Browser (http://breakout.enclavegames.com/lesson10.html) as well as a missing path method it needed.
Unfortunately i was not able to get it fully working, as it also requires the arc path method, of which i do not have the needed mathematical understanding to implement.

### DIFF
--- a/Userland/Libraries/LibWeb/HTML/CanvasRenderingContext2D.cpp
+++ b/Userland/Libraries/LibWeb/HTML/CanvasRenderingContext2D.cpp
@@ -195,6 +195,7 @@ void CanvasRenderingContext2D::stroke()
         return;
 
     painter->stroke_path(m_path, m_stroke_style, m_line_width);
+    did_draw(m_path.bounding_box());
 }
 
 void CanvasRenderingContext2D::fill(Gfx::Painter::WindingRule winding)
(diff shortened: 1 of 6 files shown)


alimpfard:
> arc

We already have a `EllipticalArcTo`, just give both its radii the value of the circle's radius :P

Userland/Libraries/LibWeb/HTML/CanvasRenderingContext2D.cpp
```diff
@@ -195,6 +195,7 @@ void CanvasRenderingContext2D::stroke()
         return;
 
     painter->stroke_path(m_path, m_stroke_style, m_line_width);
+    did_draw(m_path.bounding_box());
```
awesomekling, line 209: Nitpick about commit message: the layout (position and geometry of each `Web::Layout::Node`) is not being invalidated here, `did_draw()` only calls `set_needs_display()` on the canvas element's layout node (a `Web::Layout::CanvasBox`), meaning that it needs to be repainted.

awesomekling (changes requested):

Userland/Libraries/LibWeb/HTML/CanvasRenderingContext2D.cpp
IdanHo: my bad, fixed!

IdanHo:
> > arc
> 
> We already have a `EllipticalArcTo`, just give both its radii the value of the circle's radius :P

I tried using that, but i wasnt entirely sure what point was (first parameter), as the only coordinates of the ellipse i have is the center (second parameter)

alimpfard:
Hmmm...that's pretty awkward, seems to be one end of the arc, with the other end being the current point...kinda odd, given that the other points uniquely identify it

PR #6332, opened 2021-04-14
LibPthread: Improvements for semaphores

gunnarbeutner:
This PR improves the `sem_*()` functions:

* Ensures we're not overflowing the semaphore's value.
* Improves error handling for the semaphore functions by returning the error codes from `pthread_*()` in errno.
* Makes sure we're not holding sem->mtx after sem_wait()/sem_trywait().
* Implements sem_getvalue().

### DIFF
--- a/Userland/Libraries/LibPthread/semaphore.cpp
+++ b/Userland/Libraries/LibPthread/semaphore.cpp
@@ -51,9 +51,25 @@ int sem_destroy(sem_t* sem)
     return 0;
 }
 
-int sem_getvalue(sem_t*, int*)
+int sem_getvalue(sem_t* sem, int* sval)
 {
-    VERIFY_NOT_REACHED();
+    auto rc = pthread_mutex_trylock(&sem->mtx);
+
+    if (rc == EBUSY) {
+        *sval = 0;
+        return 0;
+    }
+
+    if (rc != 0) {
+        errno = rc;
+        return -1;
+    }
+
+    *sval = sem->value;
+
+    pthread_mutex_unlock(&sem->mtx);
+
+    return 0;
 }
 
 int sem_init(sem_t* sem, int shared, unsigned int value)
--- a/Userland/Libraries/LibPthread/semaphore.cpp
+++ b/Userland/Libraries/LibPthread/semaphore.cpp
@@ -36,8 +36,18 @@ int sem_close(sem_t*)
 
 int sem_destroy(sem_t* sem)
 {
-    pthread_mutex_destroy(&sem->mtx);
-    pthread_cond_destroy(&sem->cv);
+    auto rc = pthread_mutex_destroy(&sem->mtx);
+    if (rc != 0) {
+        errno = rc;
+        return -1;
+    }
+
+    rc = pthread_cond_destroy(&sem->cv);
+    if (rc != 0) {
+        errno = rc;
+        return -1;
+    }
+
     return 0;
 }
 
@@ -58,11 +68,17 @@ int sem_init(sem_t* sem, int shared, unsigned int value)
         return -1;
     }
 
-    if (pthread_mutex_init(&sem->mtx, nullptr) != 0)
+    auto rc = pthread_mutex_init(&sem->mtx, nullptr);
+    if (rc != 0) {
+        errno = rc;
         return -1;
+    }
 
-    if (pthread_cond_init(&sem->cv, nullptr) != 0)
+    rc = pthread_cond_init(&sem->cv, nullptr);
+    if (rc != 0) {
+        errno = rc;
         return -1;
+    }
 
     sem->value = value;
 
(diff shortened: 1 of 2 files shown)


Userland/Libraries/LibPthread/semaphore.cpp
```diff
@@ -47,9 +47,20 @@ int sem_destroy(sem_t* sem)
...
 
-int sem_getvalue(sem_t*, int*)
+int sem_getvalue(sem_t* sem, int* sval)
 {
-    VERIFY_NOT_REACHED();
+    errno = pthread_mutex_trylock(&sem->mtx);
+    if (errno == EBUSY) {
+        *sval = 0;
+        return 0;
+    } else if (errno != 0)
```
alimpfard, line 56: No need for `else` after `return`.

Userland/Libraries/LibPthread/semaphore.cpp
```diff
@@ -81,16 +92,23 @@ int sem_post(sem_t* sem)
 
     sem->value++;
 
-    pthread_cond_signal(&sem->cv);
+    errno = pthread_cond_signal(&sem->cv);
```
alimpfard, line 110: Not an issue with your impl, but these `pthread_cond_*` functions will blow up if something goes wrong, so control won't reach here anyway.
Also, they don't seem to return `errno` anyway, `futex()` sets `errno` and `pthread_cond_signal` just blows up if `ret < 0`.

Userland/Libraries/LibPthread/semaphore.cpp
gunnarbeutner: In theory `pthread_mutex_trylock()` could return something else other than `EBUSY` and `0` though, right? I know our `pthread_mutex_trylock()` doesn't currently, but maybe later on.

Userland/Libraries/LibPthread/semaphore.cpp
gunnarbeutner: I'm OK with things blowing up when `sem->cv` isn't a valid condvar or if something else bad happens. As for `errno`, I know that `pthread_*` doesn't set errno and just returns the error number instead. However, the `sem_post` caller expects the error code in `errno` so that's why I'm copying it to `errno` for them.

Userland/Libraries/LibPthread/semaphore.cpp
alimpfard: I just meant that you can write this as
```c++
if (errno == EBUSY) {
    // EBUSY things
}
if (errno != 0)
    // not EBUSY things
```

gunnarbeutner: Oh right, good point. I'll change that. :+1:

Userland/Libraries/LibPthread/semaphore.cpp
alimpfard: it doesn't look like `pthread_cond_signal()` (for example) is returning the error.
It asserts that the return value of `futext()` is not -1 (`futex()` returns -1 and sets `errno` if something goes wrong), and then just returns zero.

see `pthread.cpp:519` (for `pthread_cond_signal()`), and `serenity.cpp:97` (for `futex()`)

gunnarbeutner: Aye, you're right - that's what it does at the moment. But, according to POSIX.1 it _could_ return an error code at some later point, like `EINVAL` if someone decides to implement that (not that I see why anyone would want to - for performance reasons). So I guess my code is overly cautious.

Userland/Libraries/LibPthread/semaphore.cpp
```diff
@@ -36,25 +36,53 @@ int sem_close(sem_t*)
...
 int sem_init(sem_t* sem, int shared, unsigned int value)
 {
     if (shared)
         return ENOSYS;
 
-    if (pthread_mutex_init(&sem->mtx, nullptr) != 0)
+    if (value > SEM_VALUE_MAX)
+        return EINVAL;
+
+    if (value > SEM_VALUE_MAX)
```
ADKaster, line 77: Dr. Posix says to return 0, or -1, not an errno value. Pre-existing though.
https://www.man7.org/linux/man-pages/man3/sem_init.3p.html

gunnarbeutner: Yup, good catch. I've updated the PR. :+1:

Userland/Libraries/LibPthread/semaphore.cpp
tomuta: @alimpfard Does this get optimized as expected? Meaning is it going to essentially do two `cmp` if the first condition is true? Just wondering because `errno` is a special (tls?) variable. I usually use `else if` for those reasons and only do two `if` if I'm e.g. `return`ing in the first condition.

gunnarbeutner: I just disassembled `sem_getvalue` for the fun of it and it does appear to get optimized properly. The compiler stores the result from the call into the TLS variable (which arguably isn't necessary if the result is 0 - but I'm not sure if that's worth optimizing) but keeps using the value it has in a register for the two CMPs, i.e. it does _not_ re-load the value from the TLS segment.

gunnarbeutner: Well, I decided that I do care after all, so I added a temporary variable instead of just storing the result right into `errno`.

PR #6334, opened 2021-04-14
SystemServer+LibCore: Allow service to request multiple sockets

sin-ack:
SystemServer only allowed a single socket to be created for a service
before this.  Now, SystemServer will allow any amount of sockets.  The
sockets can be defined like so:

```ini
[SomeService]
Socket=/tmp/portal/socket1,/tmp/portal/socket2,/tmp/portal/socket3
SocketPermissions=660,600
```

The last item in SocketPermissions is applied to the remainder of the
sockets in the Socket= line, so multiple sockets can have the same
permissions without having to repeat them.

Defining multiple sockets is not allowed for socket-activated services
at the moment, and wouldn't make much sense anyway.

This patch also makes socket takeovers more robust by removing the
assumption that the socket will always be passed in fd 3.  Now, the
SOCKET_TAKEOVER environment variable carries information about which
endpoint corresponds to which socket, like so:

```console
SOCKET_TAKEOVER=/tmp/portal/socket1:3 /tmp/portal/socket2:4
```
and LocalServer/LocalService will parse this automatically and select
the correct one.  The old behavior of getting the default socket is
preserved so long as the service only requests a single socket in
SystemServer.ini.

### DIFF
--- a/Userland/Libraries/LibCore/LocalServer.cpp
+++ b/Userland/Libraries/LibCore/LocalServer.cpp
@@ -50,28 +50,39 @@ LocalServer::~LocalServer()
         ::close(m_fd);
 }
 
-bool LocalServer::take_over_from_system_server()
+bool LocalServer::take_over_from_system_server(String const& socket_path)
 {
     if (m_listening)
         return false;
 
-    constexpr auto socket_takeover = "SOCKET_TAKEOVER";
+    if (!LocalSocket::s_overtaken_sockets_parsed)
+        LocalSocket::parse_sockets_from_system_server();
 
-    if (getenv(socket_takeover)) {
+    int fd = -1;
+    if (socket_path.is_null()) {
+        // We want the first (and only) socket.
+        if (LocalSocket::s_overtaken_sockets.size() == 1) {
+            fd = LocalSocket::s_overtaken_sockets.begin()->value;
+        }
+    } else {
+        auto it = LocalSocket::s_overtaken_sockets.find(socket_path);
+        if (it != LocalSocket::s_overtaken_sockets.end()) {
+            fd = it->value;
+        }
+    }
+
+    if (fd >= 0) {
         // Sanity check: it has to be a socket.
         struct stat stat;
-        int rc = fstat(3, &stat);
+        int rc = fstat(fd, &stat);
         if (rc == 0 && S_ISSOCK(stat.st_mode)) {
-            // The SystemServer has passed us the socket as fd 3,
-            // so use that instead of creating our own.
-            m_fd = 3;
+            // The SystemServer has passed us the socket, so use that instead of
+            // creating our own.
+            m_fd = fd;
             // It had to be !CLOEXEC for obvious reasons, but we
             // don't need it to be !CLOEXEC anymore, so set the
             // CLOEXEC flag now.
             fcntl(m_fd, F_SETFD, FD_CLOEXEC);
-            // We wouldn't want our children to think we're passing
-            // them a socket either, so unset the env variable.
-            unsetenv(socket_takeover);
 
             m_listening = true;
             setup_notifier();
--- a/Userland/Libraries/LibCore/LocalSocket.cpp
+++ b/Userland/Libraries/LibCore/LocalSocket.cpp
@@ -28,6 +28,7 @@
 #include <errno.h>
 #include <fcntl.h>
 #include <stdio.h>
+#include <stdlib.h>
 #include <sys/socket.h>
 #include <sys/stat.h>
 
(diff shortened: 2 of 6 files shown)


ADKaster:
Defining multiple sockets sure looks like the ini file for service definitions will eventually not be sufficient. XML anyone? 😜

```xml
<Service name=SomeService>
<Sockets>
<Socket path=/tmp/portal/socket1 fd=3 />
<Socket path=/tmp/portal/socket2 fd=4 />
</Sockets>
</Service>
```

(Forreal though, if it gets much more complex we might want JSON)

sin-ack:
> Defining multiple sockets sure looks like the ini file for service definitions will eventually not be sufficient. XML anyone? stuck_out_tongue_winking_eye
> 
> ```
> <Service name=SomeService>
> <Sockets>
> <Socket path=/tmp/portal/socket1 fd=3 />
> <Socket path=/tmp/portal/socket2 fd=4 />
> </Sockets>
> </Service>
> ```
> 
> (Forreal though, if it gets much more complex we might want JSON)

We already have .ini, so might as well create service files. :^)

Userland/Services/SystemServer/Service.h
```diff
@@ -60,10 +60,12 @@ class Service final : public Core::Object {
...
-    // Path to the socket to create and listen on on behalf of this service.
-    String m_socket_path;
-    // File system permissions for the socket.
-    mode_t m_socket_permissions { 0 };
+    // A list of each socket path that is required by the service.
+    Vector<String> m_socket_paths;
+    // File descriptors for each socket.
+    Vector<int> m_socket_fds;
+    // File system permissions for each socket.
+    Vector<mode_t> m_socket_permissions;
```
awesomekling, line 68: Seems like we could bundle three socket attributes in a struct instead of having three separate vectors.

Userland/Libraries/LibCore/LocalServer.cpp
```diff
@@ -50,28 +50,39 @@ LocalServer::~LocalServer()
         ::close(m_fd);
 }
 
-bool LocalServer::take_over_from_system_server()
+bool LocalServer::take_over_from_system_server(String const& socket_path)
```
awesomekling, line 53: Normally this project uses west const style, but I never though much about it.
Now that I do, east const style does seem more logical and I think I'll adopt it too. :)

awesomekling (changes requested):

awesomekling:
> > Defining multiple sockets sure looks like the ini file for service definitions will eventually not be sufficient. XML anyone? stuck_out_tongue_winking_eye
> > ```
> > <Service name=SomeService>
> > <Sockets>
> > <Socket path=/tmp/portal/socket1 fd=3 />
> > <Socket path=/tmp/portal/socket2 fd=4 />
> > </Sockets>
> > </Service>
> > ```
> > 
> > 
> > (Forreal though, if it gets much more complex we might want JSON)
> 
> We already have .ini, so might as well create service files. :^)

Yes, a directory with separate files per service might be nice actually :)

Userland/Libraries/LibCore/LocalServer.cpp
sin-ack: That's awesome! East const makes more sense when you consider that C style types are read right-to-left, I think. I always write it as `char const *` on my own code. I try to follow the code styles of other projects in PRs, but I accidentally slipped that one in :^)

Userland/Services/SystemServer/Service.h
sin-ack: What would be a good place to put this? Just above the `Service`?

awesomekling: Sure, or as a private nested struct within `Service` if you don't need it outside the class.

Userland/Libraries/LibCore/LocalSocket.cpp
```diff
@@ -73,15 +74,49 @@ LocalSocket::~LocalSocket()
 {
 }
 
-RefPtr<LocalSocket> LocalSocket::take_over_accepted_socket_from_system_server()
+HashMap<String, int> LocalSocket::s_sockets {};
+bool LocalSocket::s_parsed { false };
+
+void LocalSocket::parse_sockets()
```
awesomekling, line 80: Since this stuff is specific to socket takeover via the `SOCKET_TAKEOVER` environment variable, let's not use generic names like `s_parsed` and `parse_sockets`.

The old name, `take_over_accepted_socket_from_system_server` was much better here. If we want to change the names, we should find something of similar specificity so that we don't lose clarity here :)

Userland/Services/SystemServer/Service.h
```diff
@@ -96,7 +103,8 @@ class Service final : public Core::Object {
     // times where it has exited unsuccessfully and too quickly.
     int m_restart_attempts { 0 };
 
-    void setup_socket();
+    void setup_socket(unsigned socket_index);
```
awesomekling, line 106: Let's make this take a `SocketDescriptor&` directly since `setup_socket` has no particular need to know the index of it in the vector.

Userland/Libraries/LibCore/LocalSocket.cpp
sin-ack: I see. Since this interface is private, I just used a shorter name, but how does `parse_sockets_from_system_server` sound? Note that the public interface didn't change.

awesomekling (approved):
Ok cool. We should also update the SystemServer(5) man page, but let's merge this in the meantime :)

PR #6335, opened 2021-04-14
LibGUI+WindowServer: Separate window manager IPC from regular IPC

sin-ack:
With this patch the window manager related functionality is split out
onto a new endpoint pair named WindowManagerServer/Client.  This allows
window manager functionality to be potentially privilege separated in
the future.  To this end, a new client named WMConnectionClient
is used to maintain a window manager connection.  When a process
connects to the endpoint and greets the WindowServer as a window manager
(via Window::make_window_manager(int)), they're subscribed to the events
they requested via the WM event mask.

This patch also removes the hardcoding of the Taskbar WindowType to
receive WM events automatically.  However, being a window manager still
requires having an active window, at the moment.

~~This patch currently depends on #6334, and will be a draft until that is merged.~~
Since #6334 is merged I've un-drafted it.

### DIFF
--- a/Userland/Services/Taskbar/main.cpp
+++ b/Userland/Services/Taskbar/main.cpp
@@ -36,7 +36,9 @@
 #include <LibGUI/ActionGroup.h>
 #include <LibGUI/Application.h>
 #include <LibGUI/Menu.h>
+#include <LibGUI/WindowManagerServerConnection.h>
 #include <LibGUI/WindowServerConnection.h>
+#include <WindowServer/Window.h>
 #include <serenity.h>
 #include <signal.h>
 #include <spawn.h>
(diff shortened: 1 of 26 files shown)


Userland/Services/Taskbar/main.cpp
```diff
@@ -72,6 +77,12 @@ int main(int argc, char** argv)
     auto window = TaskbarWindow::construct(move(menu));
     window->show();
 
+    // XXX: without the line comments at the end this line goes to 200 chars.
```
awesomekling, line 80: If you want to avoid the long line, put the `|` characters at the start of each line instead of at the end :)

awesomekling:
This looks really nice.

Userland/Services/Taskbar/main.cpp
sin-ack: Done. :^)

PR #6336, opened 2021-04-14
Ports: Make sure we're building libvorbis before SDL2_mixer

gunnarbeutner:
This hasn't been an issue for me so far because I was using `./build_all.sh` which builds the ports in alphabetical order (so `libvorbis` comes before `SDL2_mixer`).


bcoles (approved):
seems legit

PR #6337, opened 2021-04-14
LibJS: Implement String.prototype.substr according to the spec

### LINKED ISSUE #6325: LibJS: Crash in String.prototype.substr on HTML spec
Seeing the following crash on the [HTML spec bookmark](https://html.spec.whatwg.org/multipage/):

```
NotPresent fault on Write at address 0x055f0000

0x260d60e0: [libc.so] memcpy +0x10 (string.cpp:138)
0x260f728e: [libc.so] AK::StringImpl::create(char const*, unsigned long, AK::ShouldChomp) +0xee (Atomic.h:266)
0x260efda9: [libc.so] AK::String::substring(unsigned long, unsigned long) const +0x99 (String.h:85)
0x8449e386: [libjs.so] JS::StringPrototype::substr(JS::VM&, JS::GlobalObject&) +0x3b6 (RefPtr.h:189)
0x8441a18f: [libjs.so] AK::Function<JS::Value (JS::VM&, JS::GlobalObject&)>::CallableWrapper<JS::Value (*)(JS::VM&, JS::GlobalObject&)>::call(JS::VM&, JS::GlobalObject&) const +0x2f (Function.h:111)
0x84461e9b: [libjs.so] JS::NativeFunction::call() +0x4b (NativeFunction.cpp:64)
```

It looks like `JS::StringPrototype::substr` tries to invoke `AK::String::substring` on an empty string with `start=0` and `length=-1`. 

[The `substr` spec](https://tc39.es/ecma262/#sec-string.prototype.substr) notes that if `length <= 0` we should return the empty string, so a super simple fix is to just do that (we currently only check `length == 0`). But I figured while I was at it, might as well fully implement `substr` to be spec compliant. But I'm confused on one part in the spec, hoping someone can help me read this :)

The spec notes:
> 4. Let intStart be ? ToIntegerOrInfinity(start).
> 5. If intStart is -∞, set intStart to 0.
> 6. Else if intStart < 0, set intStart to max(size + intStart, 0).

My confusion is how `intStart < 0` could be true here if it is the result of `ToIntegerOrInfinity`? The spec of `ToIntegerOrInfinity` says `If number < +0𝔽, set integer to -integer`. So it seems the result will always be non-negative? That is what I'm seeing after implementing the `substr` spec: if `start` is -1, then `ToIntegerOrInfinity` converts it to 1 and we get bad results out of `substr` because that step 6 doesn't take effect.

trflynn89:
Fixes #6325

The JavaScript on the HTML Spec site that caused the crash is:
    window.location.hash.substr(1)

Of course, window.location.hash can be the empty string. The spec allows
for calling substr(1) on an empty string, but our partial implementation
wasn't handling it properly.

### DIFF
--- a/Userland/Libraries/LibJS/Runtime/StringPrototype.cpp
+++ b/Userland/Libraries/LibJS/Runtime/StringPrototype.cpp
@@ -424,28 +424,31 @@ JS_DEFINE_NATIVE_FUNCTION(StringPrototype::substr)
         return js_string(vm, string);
 
     // FIXME: this should index a UTF-16 code_point view of the string.
-    auto string_length = (i32)string.length();
+    auto size = (i32)string.length();
 
-    auto start_argument = vm.argument(0).to_i32(global_object);
+    auto int_start = vm.argument(0).to_integer_or_infinity(global_object);
+    if (vm.exception())
+        return {};
+    if (Value(int_start).is_negative_infinity())
+        int_start = 0;
+    if (int_start < 0)
+        int_start = max(size + (i32)int_start, 0);
+
+    auto length = vm.argument(1);
+
+    auto int_length = length.is_undefined() ? size : length.to_integer_or_infinity(global_object);
     if (vm.exception())
         return {};
 
-    auto start = start_argument < 0 ? (string_length - -start_argument) : start_argument;
-
-    auto length = string_length - start;
-    if (vm.argument_count() >= 2) {
-        auto length_argument = vm.argument(1).to_i32(global_object);
-        if (vm.exception())
-            return {};
-        length = max(0, min(length_argument, length));
-        if (vm.exception())
-            return {};
-    }
-
-    if (length == 0)
+    if (Value(int_start).is_positive_infinity() || (int_length <= 0) || Value(int_length).is_positive_infinity())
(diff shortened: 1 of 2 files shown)


Userland/Libraries/LibJS/Runtime/StringPrototype.cpp
```diff
@@ -424,28 +424,33 @@ JS_DEFINE_NATIVE_FUNCTION(StringPrototype::substr)
...
         return {};
+    if (Value(int_start).is_negative_infinity())
+        int_start = 0;
+    else if (int_start < 0)
+        int_start = max(size + (i32)int_start, 0);
 
-    auto start = start_argument < 0 ? (string_length - -start_argument) : start_argument;
+    auto length = vm.argument(1).value_or(js_undefined());
+    if (vm.exception())
+        return {};
```
linusg, line 439: This can't throw, and `vm.argument()` OOB access already returns undefined

```suggestion
    auto length = vm.argument(1);
```

trflynn89: :+1:

linusg (approved):

PR #6340, opened 2021-04-15
Kernel: Virtio attempt 3

ccapitalK:
This is a fast-forward of tomuta's and idanho's work in #4748, which implemented most of the virtio api but is currently several months out of date and uses some kernel APIs that have been removed. They implemented an interface that covers most of the Virtio API, the previous PR was only missing some work on split virtqueues and switching over the main console from port 0xe9, which aren't really necessary for basic virtio support.

What I want to do with this PR is to get the VirtIO MVP into the repo proper, so that other people can start developing on it (and hopefully fixing all the remaining issues). There are still a couple of sharp edges here, for instance writing too quickly to the virtio console will assert-fail instead of blocking, but I think it's good enough to get into the kernel.


ccapitalK:
#3236

ccapitalK:
@Matrix89

supercomputer7:
I see you know, there's already a PR about it (#4748). Maybe we should decide what to keep open and then to continue with it.
Also, I already implemented `ScatterGatherList` in the AHCI code, so please use it.
Don't forget to squash-rebase your commits. The changes in https://github.com/SerenityOS/serenity/pull/6340/commits/2d3cdc55aafe75f4ed99c3b26f1c5fb705455115 should be propagated to the relevant places and not be a separate commit.

ccapitalK:
> I see you know, there's already a PR about it (#4748). Maybe we should decide what to keep open and then to continue with it.
> Also, I already implemented `ScatterGatherList` in the AHCI code, so please use it.
> Don't forget to squash-rebase your commits. The changes in [2d3cdc5](https://github.com/SerenityOS/serenity/commit/2d3cdc55aafe75f4ed99c3b26f1c5fb705455115) should be propagated to the relevant places and not be a separate commit.

I've squash-rebased it so that the changes are spread through the history of commits. I've added myself as a co-author of any commit that needed significant changes to compile.

I would refactor this to use AHCIPort::ScatterList, but that would require refactoring that implementation to not be an internal class of the AHCI subsystem. To avoid PR creep, I think we should have a separate PR to do that refactoring.

IdanHo:
I haven't touched virtio in a while, so i cant provide much input, but your fixups look sane to me :), i agree with supercomputer7 that we shouldnt have 2 seperate implementions of scatter gather lists, but if the 2 will get merged in a seperate PR that is probably fine as well.

ccapitalK:
Alright, so what are people's thoughts on this? I think it's ready to merge, with a follow-up task to refactor out the scatter gather list to be a common one.

PR #6342, opened 2021-04-15
Everywhere: Add "free" warnings

Nicholas-Baron:
The following warnings do not occur anywhere in the codebase and so enabling them is effectively free:
 * `-Wcast-align`
 * `-Wduplicated-cond`
 * `-Wformat=2`
 * `-Wlogical-op`
 * `-Wmisleading-indentation`
 * `-Wunused`

These are taken as a strict subset of the list in #5487.


awesomekling:
Sounds good to me!

PR #6343, opened 2021-04-15
Toolchain: Thread support for the C++ standard library

gunnarbeutner:
This enables thread support for the C++ standard library which will enable porting some software. I want to port openttd at some point but that'll also require C++ exception handling, so I guess this is part 1. There are a few things I'm not quite happy with:

Static initializers are now thread safe, however - for that to work GCC inserts calls to `pthread_*` functions in programs which did not otherwise expect that they'd make calls to libpthread.

As a workaround I've linked libc against libpthread for now and hacked `pledge()` to always add the `Pledge::thread` flag. Before this gets merged I'll definitely need to figure out how to tell GCC not to make static initializers thread-safe somehow.

Also, there are a bunch of warnings when rebuilding the toolchain because the build script tries to run `strip` against header files.


RealKC:
I've also done work on this, but real life got in the way, you could take a look at #6071 if any of the patches there can help you with this.

I think you should build the gcc port with this option too

gunnarbeutner:
Ah, that goes to show I need to use GitHub's search feature more. I'll have a closer look in a bit. :+1:

gunnarbeutner:
@RealKC If you want you can give this a try. Together with the changes from #6480, #6481 and various other PRs this should work properly now. :tada:

gunnarbeutner:
GitHub seems to be using the cached toolchain for this which doesn't install `pthread.h` in the build dir. Not sure what to do about that. :man_shrugging:

awesomekling:
This has conflicts. Looks cool otherwise :)

PR #6346, opened 2021-04-15
Documentation: Describe how to use the iPXE bootloader for network boot

supercomputer7:
This method seems useful for bare-metal debugging.

### DIFF
--- a/Documentation/NetworkBoot.md
+++ b/Documentation/NetworkBoot.md
@@ -1,4 +1,4 @@
-## SerenityOS network booting
+## SerenityOS network booting via TFTP and DHCP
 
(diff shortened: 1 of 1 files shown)


Documentation/NetworkBoot.md
```diff
@@ -143,3 +143,37 @@ echo 0 > /sys/devices/virtual/net/br0/bridge/multicast_querier
...
+dhcp
+kernel http://X.Y.Z.W/Kernel serial_debug root=/dev/ramdisk0
+imgfetch http://X.Y.Z.W/ramdisk
+boot
+```
+This file can be called in any name you'd want. For the sake of simplicity in this guide,
+this file is named `script.ipxe` from now on.
+Don't forget to replace `X.Y.Z.W` with your HTTP server IP address.
+For troubleshooting purposes, you could add `disable_physical_storage` and `disable_ps2_controller` 
+If you suspect our implementation fails to work with your hardware.
```
rajiv-k, line 170: s/If/if ?

PR #6347, opened 2021-04-15
LibGfx+LibWeb: CRC2D.ellipse() and .arc(), plus less recursion in drawing algorithms

alimpfard:
Here's how MDN's example is supposed to look:
shot-2021-04-15_18-50-21

and how it looks:
shot-2021-04-15_18-50-38

As is obvious, I failed to make it work to-spec, but hey, it's a start.

Continuing #6330: 
shot-2021-04-15_18-54-34
(it's far more jittery than it looks)

cc @IdanHo, @mattco98.


IdanHo:
very nice work! :)

PR #6348, opened 2021-04-15
WIP: UserspaceEmulator: Add a callgrind-like CallTracer

TobyAsE:
I wanted to add a utility to the UserspaceEmulator that somehow mimics what callgrind does.
What I think it should do:
- Show you the hottest/coldest functions in the program
- Show how many calls to these functions there were
- Show who called who and how often

It would be cool to have a tool like KCacheGrind in the future but let's start slow ;)

You enable the call-tracer by passing `-c` to the UserspaceEmulator.
I hooked into the call instructions and kept track who called who. There for sure are a lot of optimizations/edge cases but for now this approach works somewhat.

### DIFF
--- a/Userland/DevTools/UserspaceEmulator/CallTracer.cpp
+++ b/Userland/DevTools/UserspaceEmulator/CallTracer.cpp
@@ -1,0 +1,109 @@
+/*
+ * Copyright (c) 2021, Tobias Christiansen <tobi@tobyase.de>
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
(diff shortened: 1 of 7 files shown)


Userland/DevTools/UserspaceEmulator/CallTracer.cpp
```diff
@@ -0,0 +1,109 @@
...
+namespace UserspaceEmulator {
+
+CallTracer::CallTracer(Emulator& emulator)
+    : m_emulator(emulator)
+{
+}
+
+void CallTracer::prepare_call_data()
+{
+    if (m_has_been_sorted)
```
Hendiadyoin1, line 43: maybe call it m_has_bee_prepared

PR #6349, opened 2021-04-15
Browser+LibWeb+WebContent: Changes for cookie safety

trflynn89:
Two changes here to protect the browser from bad cookies:
1. Limit the max size of cookies to 4KiB (a recommendation from the spec).
2. Parse cookies in the OOP tab, and send the parsed result over IPC to the main process for storage.

### DIFF
--- a/AK/Optional.h
+++ b/AK/Optional.h
@@ -36,6 +36,8 @@ namespace AK {
 template<typename T>
 class alignas(T) [[nodiscard]] Optional {
 public:
+    using ValueType = T;
+
     ALWAYS_INLINE Optional() = default;
 
     ALWAYS_INLINE Optional(const T& value)
--- a/Userland/Libraries/LibCore/DateTime.h
+++ b/Userland/Libraries/LibCore/DateTime.h
@@ -27,6 +27,7 @@
 #pragma once
 
 #include <AK/String.h>
+#include <LibIPC/Forward.h>
 #include <time.h>
 
 namespace Core {
@@ -68,5 +69,12 @@ private:
     unsigned m_minute { 0 };
     unsigned m_second { 0 };
 };
+
+}
+
+namespace IPC {
+
+bool encode(IPC::Encoder&, const Core::DateTime&);
+bool decode(IPC::Decoder&, Core::DateTime&);
 
 }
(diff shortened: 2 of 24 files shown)


AK/Optional.h
```diff
@@ -36,6 +36,8 @@ namespace AK {
 template<typename T>
 class alignas(T) [[nodiscard]] Optional {
 public:
+    using value_type = T;
```
awesomekling, line 39: Please use `TitleCase` for type names.

Userland/Libraries/LibCore/DateTime.h
```diff
@@ -70,3 +71,10 @@ class DateTime {
 };
 
 }
+
+namespace IPC {
+
+bool encode(IPC::Encoder& encoder, const Core::DateTime& datetime);
+bool decode(IPC::Decoder& decoder, Core::DateTime& datetime);
```
awesomekling, line 78: Please omit parameter names when they are already part of the type name.
```cpp
bool encode(IPC::Encoder&, Core::DateTime const&);
bool decode(IPC::Decoder&, Core::Datetime&);
```

AK/Optional.h
trflynn89: gotcha, wasn't sure of the casing on this one since I just copied the alias from `Vector`

Userland/Libraries/LibWeb/Cookie/ParsedCookie.cpp
```diff
@@ -351,3 +354,69 @@ Optional<Core::DateTime> parse_date_time(StringView date_string)
...
+
+    encoder << cookie.secure_attribute_present;
+    encoder << cookie.http_only_attribute_present;
+
+    return true;
+}
+
+bool IPC::decode(IPC::Decoder& decoder, Web::Cookie::ParsedCookie& cookie)
+{
+    auto decode_optional = [&](auto& optional) {
```
awesomekling, line 387: Hmm, hold up, can we not `decode()` directly onto an `Optional<T>`?
`IPC::Decoder` has `template<typename T> bool decode(Optional<T>&)`

awesomekling (changes requested):

Userland/Libraries/LibWeb/Cookie/ParsedCookie.cpp
trflynn89: Oh maybe, I'll check - I looked around for an `Optional` decoder but apparently not hard enough :)

trflynn89: holy moly yeah `Encoder` and `Decoder` already do all this...will fix

PR #6350, opened 2021-04-15
Meta: Use basename for serenity.sh help message

bgianfo:
I have this symlinked into ~/bin, when looking at the help/usage
it would previously print the fully qualified path to the script
making the help very difficult to read.

### DIFF
--- a/Meta/serenity.sh
+++ b/Meta/serenity.sh
@@ -3,56 +3,57 @@ set -e
 
 ARG0=$0
 print_help() {
+    NAME=$(basename "$ARG0")
     cat <<EOF
-Usage: $ARG0 COMMAND [TARGET] [ARGS...]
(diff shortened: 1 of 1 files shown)


Meta/serenity.sh
```diff
@@ -3,56 +3,57 @@ set -e
 
 ARG0=$0
 print_help() {
+    NAME=$(basename "$ARG0")
     cat <<EOF
-Usage: $ARG0 COMMAND [TARGET] [ARGS...]
+Usage: $SCRIPT COMMAND [TARGET] [ARGS...]
```
awesomekling, line 8: ...`$SCRIPT`?

bgianfo: whooops, thanks!

PR #6351, opened 2021-04-15
Everything: Add `-Wnon-virtual-dtor` flag

Nicholas-Baron:
This flag warns on classes which have `virtual` functions but do not have a `virtual` destructor.

This patch adds both the flag and missing destructors. The access level
of the destructors was determined by two rules of thumb:
1. A destructor should have a similar or lower access level to that of a
   constructor.
2. Having a `private` destructor implicitly deletes the default
   constructor, which is probably undesirable for "interface" types
   (classes with only virtual functions and no data).

In short, most of the added destructors are `protected`, unless the
compiler complained about access.


Nicholas-Baron:
Work towards #5487.

PR #6352, opened 2021-04-15
LibWeb: Small additions and improvements

IdanHo:
This is continuing #6347, the game is now fully playable, but still not completely functional:

- document.location.reload() is not implemented, so the game does not restart on a lose/win
- keyboard events are not implemented, so the paddle can only be moved via the mouse

breakout


IdanHo:
NOTE: looks like something in 32aae3067ba326e6a0afae0345aecf0164920321 caused a regression (accessing some websites crashes the browser in multiprocess mode with an unveil problem), im not sure how the hell that change can cause that, but investigating.

awesomekling:
Implementing formerly missing API can easily cause new failures by allowing scripts to progress further and uncover new bugs :)

PR #6353, opened 2021-04-15
Kernel+LibDebug+HackStudio: Implement variable watchpoints

FalseHonesty:
This PR adds the ability to place a hardware watchpoint on a debugged variable. To do this from HackStudio, just right click on a variable in the variables view and select add watchpoint. Under the hood this will use ptrace to communicate with the kernel to modify the proper x86 debug registers.

This is my first time ever looking at the Kernel in depth, much less modifying it, so I'm definitely open to suggestions there.

### DIFF
--- a/Kernel/Arch/i386/CPU.cpp
+++ b/Kernel/Arch/i386/CPU.cpp
@@ -362,14 +362,15 @@ void debug_handler(TrapFrame* trap)
         PANIC("Debug exception in ring 0");
     }
     constexpr u8 REASON_SINGLESTEP = 14;
-    bool is_reason_singlestep = (read_dr6() & (1 << REASON_SINGLESTEP));
-    if (!is_reason_singlestep)
+    auto debug_status = read_dr6();
+    auto should_trap_mask = (1 << REASON_SINGLESTEP) | 0b1111;
+    if ((debug_status & should_trap_mask) == 0)
         return;
-
     if (auto tracer = process.tracer()) {
         tracer->set_regs(regs);
     }
     current_thread->send_urgent_signal_to_self(SIGTRAP);
+    write_dr6(debug_status & ~(should_trap_mask));
 }
 
 EH_ENTRY_NO_CODE(3, breakpoint);
--- a/Kernel/Arch/i386/CPU.cpp
+++ b/Kernel/Arch/i386/CPU.cpp
@@ -854,18 +854,69 @@ FlatPtr read_cr4()
     return cr4;
 }
 
-FlatPtr read_dr6()
+void read_debug_registers_into(DebugRegisterState& state)
 {
-    FlatPtr dr6;
-#if ARCH(I386)
-    asm("mov %%dr6, %%eax"
-        : "=a"(dr6));
-#else
-    asm("mov %%dr6, %%rax"
-        : "=a"(dr6));
-#endif
-    return dr6;
+    state.dr0 = read_dr0();
+    state.dr1 = read_dr1();
+    state.dr2 = read_dr2();
+    state.dr3 = read_dr3();
+    state.dr6 = read_dr6();
+    state.dr7 = read_dr7();
 }
+
+void write_debug_registers_from(const DebugRegisterState& state)
+{
+    write_dr0(state.dr0);
+    write_dr1(state.dr1);
+    write_dr2(state.dr2);
+    write_dr3(state.dr3);
+    write_dr6(state.dr6);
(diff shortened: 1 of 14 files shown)


FalseHonesty:
cc @itamar8910

Kernel/Arch/i386/CPU.cpp
```diff
@@ -1390,6 +1530,9 @@ extern "C" void enter_thread_context(Thread* from_thread, Thread* to_thread)
     set_fs(to_tss.fs);
     set_gs(to_tss.gs);
 
+    read_debug_registers_into(from_thread->debug_register_state());
```
bgianfo, line 1533: Should we guard this case so it's only performed if the processes involved are being traced?

Kernel/Arch/i386/CPU.cpp
```diff
@@ -854,6 +855,114 @@ FlatPtr read_cr4()
...
+#else
+    asm volatile("mov %%rax, %%dr2" ::"a"(value));
+#endif
+}
+
+FlatPtr read_dr3()
+{
+    FlatPtr dr3;
+#if ARCH(I386)
+    asm("mov %%dr3, %%eax"
```
bgianfo, line 948: Given these are all the same besides the debug register name, perhaps we should generate their definition via macros?

Kernel/Syscalls/ptrace.cpp
```diff
@@ -229,4 +242,56 @@ KResult Process::poke_user_data(Userspace<u32*> address, u32 data)
...
+        break;
+    case 3:
+        m_debug_register_state.dr3 = data;
+        break;
+    case 7:
+        m_debug_register_state.dr7 = data;
+        break;
+    default:
+        return EFAULT;
+    }
```
bgianfo, line 293: Should we return `EINVAL`? As we don't really fault here?

Kernel/Arch/i386/CPU.cpp
```diff
@@ -854,6 +855,114 @@ FlatPtr read_cr4()
...
+{
+    debug_register_state.dr0 = read_dr0();
+    debug_register_state.dr1 = read_dr1();
+    debug_register_state.dr2 = read_dr2();
+    debug_register_state.dr3 = read_dr3();
+    debug_register_state.dr6 = read_dr6();
+    debug_register_state.dr7 = read_dr7();
+}
+
+void write_debug_registers_from(DebugRegisterState& state)
```
bgianfo, line 868: This should be a const reference?

Userland/Libraries/LibDebug/DebugSession.h
```diff
@@ -223,6 +238,45 @@ void DebugSession::run(DesiredInitialDebugeeState initial_debugee_state, Callbac
             state = State::FreeRun;
 
         auto regs = get_registers();
+
+        auto debug_status = peek_debug(6);
```
bgianfo, line 242: Is there a constant we can use instead of `6`?

FalseHonesty: I've moved them to constants in the ptrace header, that feels like the most applicable place to put them?

Kernel/Syscalls/ptrace.cpp
```diff
@@ -229,4 +242,56 @@ KResult Process::poke_user_data(Userspace<u32*> address, u32 data)
...
+        data = m_debug_register_state.dr3;
+        break;
+    case 6:
+        data = m_debug_register_state.dr6;
+        break;
+    case 7:
+        data = m_debug_register_state.dr7;
+        break;
+    default:
+        return EFAULT;
```
bgianfo, line 268: EINVAL here as well?

bgianfo (approved):

awesomekling:
Hmm wait, how does the kernel preserve per-thread debug registers when a context switch occurs?

FalseHonesty:
> Hmm wait, how does the kernel preserve per-thread debug registers when a context switch occurs?

Relevant code is [here](https://github.com/SerenityOS/serenity/pull/6353/files#diff-05c65ddc0798c85b83113c2bd6c1285bcd8f7cf42843a7c79faa8cf86bc65acfR1445). When I first implemented debug registers, I didn't realize I had to manually store/load on context switch, so everything was crashing. After looking for a while where to put it, I found this method (`enter_thread_context`) where other states were loaded/saved and realized it had a comment saying to add support for debug registers, so I figure it's the right location?

PR #6355, opened 2021-04-15
Everywhere: Add `-Wdouble-promotion` warning

Nicholas-Baron:
This warning informs of float-to-double conversions. The best solution
seems to be to do math *either* in 32-bit *or* in 64-bit, and only to
cross over when absolutely necessary.

It seems like some old formatting functions 
(e.g. `String::format`, `StringBuilder::appendf`) inherently cause 
float-to-double promotions. 
This warning may lead to further avoidance of those functions.


Nicholas-Baron:
Work towards #5487

PR #6357, opened 2021-04-15
Everywhere: Add outln_if and warnln_if

Hendiadyoin1:
These are pendants to dbgln_if and should be compiled away when they are dissabled

### DIFF
--- a/AK/Format.h
+++ b/AK/Format.h
@@ -384,13 +384,29 @@ void outln(CheckedFormatString<Parameters...>&& fmtstr, const Parameters&... par
 
 inline void outln() { outln(stdout); }
 
+#    define outln_if(flag, fmt, ...)       \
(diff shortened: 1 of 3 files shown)


Hendiadyoin1:
Need to set semicolons....

linusg (changes requested):
Seems fine for UserspaceEmulator, but the HexEditor changes should be using `dbgln_if()`, there's no reason to print this stuff to stdout.

Hendiadyoin1:
changed that

linusg:
Please remove the `HexEditor: Use outln_if` commit, that's not needed - we can go straight from `#if, outln()` to `dbgln_if()` :)

Hendiadyoin1:
Then when the ci finishes this should be ready?

PR #6359, opened 2021-04-15
Toolchain+LibC: Implement support for -Wl,--no-undefined and -nostartfiles

gunnarbeutner:
The short version is that this fixes how `crt0` gets linked which enables support for `-Wl,--no-undefined` and `-nostartfiles`. The commit messages have a lot more details.

Unfortunately this requires a complete rebuild because the toolchain changes and cmake doesn't pick up the crt0 dependency either (i.e. doesn't realize it needs to rebuild everything).


awesomekling:
Thank you so much for working on this. Super cool! :smile:

PR #6360, opened 2021-04-15
WindowServer: Replace window menu action magic numbers with enum

linusg:
This makes it a lot easier to find where the action handling is happening, just 1, 2, 3, 4 isn't very discoverable. :^)


sin-ack:
Would it be possible to get rid of the casts by just making `MenuItem` take an enum? Or would that affect some other code?

linusg:
The `AddMenuItem` IPC call still uses arbitrary numbers (that's how regular menus are created from LibGUI).

sin-ack:
I see, me or someone else can pick up fixing that once this is merged I suppose.

awesomekling:
Yes this was some really weak code. :^)

PR #6363, opened 2021-04-15
HackStudio+TextEditor: Add statusbar with selected text information

ry-sev:
This removes the cursor label located at the top right of editor
and will implement a statusbar which displays the line
and column position of the cursor. When the user selects (highlights)
text, the statusbar will also show the number of selected lines,
words, and characters.

### DIFF
--- a/Userland/DevTools/HackStudio/HackStudioWidget.cpp
+++ b/Userland/DevTools/HackStudio/HackStudioWidget.cpp
@@ -71,6 +71,7 @@
 #include <LibGUI/RegularEditingEngine.h>
 #include <LibGUI/Splitter.h>
 #include <LibGUI/StackWidget.h>
+#include <LibGUI/Statusbar.h>
 #include <LibGUI/TabWidget.h>
 #include <LibGUI/TableView.h>
 #include <LibGUI/TextBox.h>
@@ -156,6 +157,8 @@ HackStudioWidget::HackStudioWidget(const String& path_to_project)
     initialize_debugger();
 
     create_toolbar(toolbar_container);
+
+    m_statusbar = add<GUI::Statusbar>(3);
 }
 
 void HackStudioWidget::update_actions()
@@ -280,6 +283,8 @@ void HackStudioWidget::open_file(const String& full_filename)
     m_project_tree_view->update();
 
     current_editor_wrapper().filename_label().set_text(filename);
+
+    current_editor().on_cursor_change = [this] { update_statusbar(); };
 
     current_editor().set_focus(true);
 }
@@ -484,6 +489,7 @@ void HackStudioWidget::add_new_editor(GUI::Widget& parent)
     m_current_editor_wrapper = wrapper;
     m_all_editor_wrappers.append(wrapper);
     wrapper->editor().set_focus(true);
+    wrapper->editor().on_cursor_change = [this] { update_statusbar(); };
 }
 
 NonnullRefPtr<GUI::Action> HackStudioWidget::create_switch_to_next_editor_action()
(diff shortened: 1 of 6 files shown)


Userland/DevTools/HackStudio/HackStudioWidget.cpp
```diff
@@ -1083,6 +1089,46 @@ void HackStudioWidget::initialize_menubar(GUI::Menubar& menubar)
...
+        }
+        if (first_line == last_line || selection.start().column() == 0 || selection.start().column() != 0)
+            last_line += 1;
+
+        builder.appendff("        Selected: {} {}", last_line - first_line, (last_line - first_line) != 1 ? "Lines" : "Line");
+
+        int word_count = 0;
+        bool in_word = false;
+        String selected_text = current_editor().selected_text();
+        for (char c : selected_text) {
```
ADKaster, line 1116: Would it be better to iterate over this with a UTF-8 view and count code points instead?

FalseHonesty: CLion seems to consider all emojis a single character in their selection length info, which seems like reasonable behavior imo

sin-ack: The question is whether you count individual codepoints or grapheme clusters.

FalseHonesty: Right, fair point, even CLion doesn't properly handle emojis like 👨‍👩‍👦 it seems. I think just counting codepoints is reasonable behavior for now?

linusg (commented):
This looks nice! I think it would be very useful to put the selection information "calculation" into `GUI::TextEditor` though so it can also be used in the TextEditor application and other programs, and to ensure they are all consistent with how they count things - what do you think?

ry-sev:
@linusgSure, I'll see what I can do

Userland/DevTools/HackStudio/HackStudioWidget.cpp
ry-sev: I'll be honest, I never heard of code points or grapheme clusters

Userland/DevTools/HackStudio/HackStudioWidget.cpp
```diff
@@ -1083,6 +1089,46 @@ void HackStudioWidget::initialize_menubar(GUI::Menubar& menubar)
...
+            first_line = current_editor().cursor().line();
+            last_line = current_editor().cursor().line();
+        } else {
+            first_line = selection.start().line();
+            last_line = selection.end().line();
+        }
+        if (first_line == last_line || selection.start().column() == 0 || selection.start().column() != 0)
+            last_line += 1;
+
+        builder.appendff("        Selected: {} {}", last_line - first_line, (last_line - first_line) != 1 ? "Lines" : "Line");
```
awesomekling, line 1111: Please use multiple statusbar segments instead of adding a bunch of whitespace to the first segment :)

ry-sev:
Updated screenshot of the statusbar with segments. I ended up putting 3 segments because two segments like kinda awkward because of the space between the texts. But I can change this to two segments if needed.

I also added functions to TextEditor which calculates and returns the number of selected lines and words. This is so it can be used by other editors, e.g TextEditor.

Userland/DevTools/HackStudio/HackStudioWidget.cpp
```diff
@@ -1083,6 +1089,22 @@ void HackStudioWidget::initialize_menubar(GUI::Menubar& menubar)
...
 
+void HackStudioWidget::update_statusbar()
+{
+
+    m_statusbar->set_text(0, String::formatted("Line: {}, Column: {}", current_editor().cursor().line() + 1, current_editor().cursor().column()));
+
+    StringBuilder builder;
+    if (current_editor().has_selection()) {
+        auto line_count = current_editor().selected_line_count();
+        builder.appendff("Selected: {} {}", line_count, line_count != 1 ? "Lines" : "Line");
```
linusg, line 1100:
```suggestion
        builder.appendff("Selected: {} {}", line_count, line_count != 1 ? "lines" : "line");
```

Userland/Libraries/LibGUI/TextEditor.h
```diff
@@ -135,6 +135,8 @@ class TextEditor
     bool write_to_file(const String& path);
     bool has_selection() const { return m_selection.is_valid(); }
     String selected_text() const;
+    int selected_word_count() const;
+    size_t selected_line_count() const;
```
linusg, line 139: This API feels kinda weird - why are these different types, and why signed?

PR #6365, opened 2021-04-15
Ports: Enable building shared libraries for ports

gunnarbeutner:
This adds support to most of the ports to build shared libraries. For libtool-based projects I had to resort to manually linking the libraries because libtool is currently broken. I've also added a platform definition for CMake that enables shared libraries.

I have also updated all ports to install into `/usr/local`. Previously some ports were installed into `/usr` while others used `/usr/local`.

For this to work I added support to `LibELF` so that it can load shared libraries from both `/usr/lib` and `/usr/local/lib`. At some later point we'll probably want `RPATH` support.

(This PR depends on #6359.)


Userland/Libraries/LibELF/DynamicLinker.cpp
```diff
@@ -99,13 +99,18 @@ static void map_library(const String& name, int fd)
 static void map_library(const String& name)
 {
     // TODO: Do we want to also look for libs in other paths too?
-    String path = String::formatted("/usr/lib/{}", name);
-    int fd = open(path.characters(), O_RDONLY);
-    if (fd < 0) {
-        fprintf(stderr, "Could not find required shared library: %s\n", path.characters());
-        VERIFY_NOT_REACHED();
+    const char* searchPaths[] = { "/usr/lib/{}", "/usr/local/lib/{}" };
+    for (auto& searchPath : searchPaths) {
```
awesomekling, line 103: Please use snake_case:
`searchPaths` -> `search_paths`
`searchPath` -> `search_path`

awesomekling (changes requested):

Userland/Libraries/LibELF/DynamicLinker.cpp
gunnarbeutner: Oops, fixed.

awesomekling:
This is super duper cool.

PR #6371, opened 2021-04-16
Meta+Docs+CI: Require clang-format >= 12

ADKaster:
llvm 12 was released within the last 24 hours, so that must mean it's good to use right away, right? :^)

I ran clang-format-12 over all the .h and .cpp in AK, Kernel and Userland, and it didn't come up with any differences, suprisingly.

Scanning through the closed PRs, it seems the only things that are likely to be more of an option now are Concepts? Assuming CLion and QTCreator parse them properly that is. Didn't find any obvious clang-format off comments that should be addressed either.

cc @alimpfard @linusg

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -38,9 +38,9 @@ jobs:
       # These aren't:
       run: |
         wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key | sudo apt-key add -
-        sudo add-apt-repository "deb http://apt.llvm.org/focal/ llvm-toolchain-focal-11 main"
+        sudo add-apt-repository "deb http://apt.llvm.org/focal/ llvm-toolchain-focal-12 main"
         sudo apt-get update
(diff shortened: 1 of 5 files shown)


alimpfard:
Even Arch hasn't packaged clang-12 yet... (though, yay for that, it supports concepts nicely, so we're just gonna sprinkle them around now~)

Switching the CI to 12, LGTM.

linusg (approved):
While I'm generally in favour of quickly adopting new releases, 24h seems a bit too quick? :sweat_smile:

OTOH since this doesn't change any existing formatting it's probably fine to use 11 locally for a while, CI will complain about any differences. */me waits for Fedora 34 stable...*

sin-ack:
Debian has it in `sid` right now, but I'd at least wait until that trickles down to `testing`.

ADKaster:
I'll re-open this (or make a diff PR) when clang-12 is stable in more places.

PR #6375, opened 2021-04-16
LibGL: Implement a basic MiniGL 2.x compliant Library.

Quaker762:
This currently (obviously) doesn't support any actual 3D hardware,
hence all calls are done via software rendering.

Note that any modern constructs such as shaders are unsupported,
as this driver only implements Fixed Function Pipeline functionality.

====================================================

As per discussion on Discord (#opengl channel), this is very much a team effort WIP. There are a few people that expressed interest on working on this with me, with this version as a base. Currently, we build `libGL.so` (image attached) which can be linked with any OpenGL 1.x application that has the limited subset of calls we implement (I suppose this is really a `miniGL` driver at the moment as we're nowhere near full compliance). This driver in it's current state actually works, however we currently have no way to open a context in a Serenity Window. SDL is a nightmare to work with, so it would be best if we baked some sort of functionality into `LibGUI` to start with, via a Widget etcetera. 

The consensus, for now, seems to be that we'll piggyback off of the existing Software Render features used in the cube demo, the current files in the `math` folder are from the original version of this driver that I worked on and will probably be ripped out in favour of the ones we already have. 

image


Userland/Libraries/LibGL/math/mat4.h
```diff
@@ -0,0 +1,244 @@
...
+
+        m_data[2][0] -= rhs.m_data[2][0];
+        m_data[2][1] -= rhs.m_data[2][1];
+        m_data[2][2] -= rhs.m_data[2][2];
+        m_data[2][2] -= rhs.m_data[2][2];
+
+        m_data[3][0] -= rhs.m_data[3][0];
+        m_data[3][1] -= rhs.m_data[3][1];
+        m_data[3][2] -= rhs.m_data[3][2];
+        m_data[3][2] -= rhs.m_data[3][2];
```
ccapitalK, line 144: Can you please change this to a loop? If you really want to force it to be unrolled you can use a pragma to force the compiler to do so, but I think that having the code expanded like this makes it harder to read and unnecessarily harder to modify.

Userland/Libraries/LibGL/GLMat.cpp
```diff
@@ -0,0 +1,141 @@
...
+    frustum(3, 2, d);
+    frustum(3, 3, 0);
+    // Phew
+
+    if (g_gl_state->curr_matrix_mode == GL_PROJECTION) {
+        g_gl_state->projection_matrix = g_gl_state->projection_matrix * frustum;
+    } else if (g_gl_state->curr_matrix_mode == GL_MODELVIEW) {
+#ifdef GL_DEBUG
+        dbgln("glFrustum(): frustum created with curr_matrix_mode == GL_MODELVIEW!!!");
+#endif
```
ccapitalK, line 138: Nit: This should be `dbgln_if(GL_DEBUG, "glFrustum(): frustum created with curr_matrix_mode == GL_MODELVIEW!!!");`

Userland/Libraries/LibGL/GL/gl.h
```diff
@@ -0,0 +1,102 @@
...
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+#pragma once
+
+#define GL_DEBUG
```
ccapitalK, line 28: If you are defining a new *_DEBUG parameter, I believe you should also add it to `AK/Debug.h.in` and `Meta/CMake/all_the_debug_macros.cmake`.

Userland/Libraries/LibGL/math/mat4.h
GMTA: Not only that... but the code you referenced contains a copy-paste bug (`[2]` is used twice)

Userland/Libraries/LibGL/GLMat.cpp
Hendiadyoin1: Advice: if your IDE supports regex replace use
```regex
^#if (.*)\n( *)dbgln\((.*)\);\n#endif$
```
to search and
```regex
$2dbgln_if($1, $3);
```
to replace

this only works when GL_DEBUG is standard defined as 0 as all other DEBUG flags, see @ccapitalK 's next review

Quaker762:
@awesomekling Let's see how we go with this :P

Tag @sunverwerth @alimpfard

Userland/Libraries/LibGL/GLColor.cpp
```diff
@@ -12,3 +12,23 @@ void glColor3f(GLfloat r, GLfloat g, GLfloat b)
...
+}
+
+void glColor4ub(GLubyte r, GLubyte g, GLubyte b, GLubyte a)
+{
+    g_gl_context->gl_color(r, g, b, a);
+}
+
+void glColor4ubv(const GLubyte* v)
+{
+    g_gl_context->gl_color(v[0], v[1], v[2], v[3]);
```
sunverwerth, line 35: Shouldn't this divide the byte by 255.f

Userland/Libraries/LibGL/GLColor.cpp
```diff
@@ -12,3 +12,23 @@ void glColor3f(GLfloat r, GLfloat g, GLfloat b)
...
+}
+
+void glColor4fv(const GLfloat* v)
+{
+    g_gl_context->gl_color(v[0], v[1], v[2], v[3]);
+}
+
+void glColor4ub(GLubyte r, GLubyte g, GLubyte b, GLubyte a)
+{
+    g_gl_context->gl_color(r, g, b, a);
```
sunverwerth, line 30: Shouldn't this divide by 255.f first?

sunverwerth (changes requested):

Userland/Libraries/LibGL/GLColor.cpp
alimpfard: I didn't notice this in the docs, lemme lookie again

alimpfard: Ah yeah, it's written like "blah blah are mapped to..." in the middle of other stuff. oops.

Userland/Libraries/LibGL/GLContext.cpp
```diff
@@ -0,0 +1,32 @@
...
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+#include "SoftwareGLContext.h"
+#include <LibGfx/Bitmap.h>
+
+__attribute__((visibility("hidden"))) GLContext* g_gl_context;
+
+GLContext::~GLContext()
```
sunverwerth, line 13: This goes for most files: we should probably put everything that is not an official OpenGL api into a `GL` namespace like the other system libraries do.

Userland/Libraries/LibGL/GLStruct.h
```diff
@@ -0,0 +1,54 @@
...
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#pragma once
+
+#include "GL/gl.h"
+
+typedef struct
```
sunverwerth, line 31: Is tyepef struct a c-ism that we should remove?

sunverwerth (changes requested):

IdanHo:
48d351141054ab022e4e00586a0303f4731d4db1 should probably be squashed

sunverwerth (changes requested):

Userland/Libraries/LibGL/GLStruct.h
Quaker762: Yeah. C++ will allow you to use the type of anything marked `struct` as it's considered a class with members public by default.

sunverwerth (approved):

Userland/Libraries/LibGL/GLContext.cpp
```diff
@@ -0,0 +1,36 @@
...
+ * Copyright (c) 2021, Jesse Buhagiar <jooster669@gmail.com>
+ * Copyright (c) 2021, Stephan Unverwerth <s.unverwerth@gmx.de>
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+#include "SoftwareGLContext.h"
+#include <LibGfx/Bitmap.h>
+
+__attribute__((visibility("hidden"))) GL::GLContext* g_gl_context;
```
ccapitalK, line 11: Is it possible to declare this static or use an anonymous namespace instead of using a gcc attribute?

sunverwerth: Static no, it's used in multiple files. But we could probably remove this attribute completely. I was just trying to prevent users of the library from directly mucking with the context variable, hence the `extern gl_context` in every file. But if anyone is extra willing to shoot themselves in the foot... meh.

ccapitalK: Oh, I didn't notice it was being used in multiple translation units. In that case, using the attribute is probably the best way.

sunverwerth: @ccapitalK Yeah, I wasn't shure if this is attribute is ok to use. It's used in two places, the kernel and LibELF. Kling or Linus should decide 😅

Quaker762:
@awesomekling @linusg This is ready for review now :^)

Userland/Libraries/LibGL/SoftwareRasterizer.cpp
```diff
@@ -0,0 +1,257 @@
...
+    if (v2.y > max.y)
+        max.y = v2.y;
+    if (v1.x < min.x)
+        min.x = v1.x;
+    if (v2.x < min.x)
+        min.x = v2.x;
+    if (v1.y < min.y)
+        min.y = v1.y;
+    if (v2.y < min.y)
+        min.y = v2.y;
```
ccapitalK, line 105: Could you please change this to be a chain of something like `max.x = AK::max(max.x, AK::max(v1.x, v2.x));` ? GCC compiles this much better than the chain of conditional assignments (https://godbolt.org/z/xqWYdd3xs shows that the std::max version generates cmovs while the if statement chain generates lots of jumps). I think gcc can't figure out that these can be done branchlessly because you don't have else clauses for these if statements.

ccapitalK: I've found that gcc has issues converting conditional assignments into branchless assignments unless you have both if and else clauses assigning to the same variable.

Userland/Libraries/LibGL/SoftwareGLContext.cpp
```diff
@@ -0,0 +1,766 @@
...
+};
+
+static FloatVector4 clip_plane_normals[] = {
+    { 1, 0, 0, 1 },  // Left Plane
+    { -1, 0, 0, 1 }, // Right Plane
+    { 0, -1, 0, 1 }, // Top Plane
+    { 0, 1, 0, 1 },  // Bottom plane
+    { 0, 0, -1, 1 }, // Near Plane
+    { 0, 0, 1, 1 }   // Far Plane
+};
```
ccapitalK, line 43: Please mark these const.

Also, I'm not sure if this would be a micro-opt, but you could remove the clip_plane_normals array, since `clip_plane_normals[i] == clip_planes[i^1]`.

Userland/Libraries/LibGL/Clipper.cpp
```diff
@@ -0,0 +1,84 @@
...
+    for (u8 plane = 0; plane < NUMBER_OF_CLIPPING_PLANES; plane++) {
+        swap(current, output_list);
+        clipped_polygon.clear();
+
+        if (current->size() == 0) {
+            return;
+        }
+
+        // Save me, C++23
+        for (auto i = 0; i < static_cast<i32>(current->size()); i++) {
```
IdanHo, line 67:
```suggestion
        for (size_t i = 0; i < current->size(); i++) {
```

Quaker762:
Current implementation state :^)

image

Userland/Libraries/LibGL/Clipper.cpp
Quaker762: What's the effect on the modulus below this? If `i` is `0`, then we'll underflow to `0xFFFFFFFF`.

IdanHo: that wouldnt work with signed integers anyways, you would be trying to access a negative index of the vector, which would ASSERT

alimpfard: Curious, it does appear that `(-1)%x` == `-1` in C++
I'm intrigued by why this doesn't crash

Quaker762: > that wouldnt work with signed integers anyways, you would be trying to access a negative index of the vector, which would ASSERT

Interesting. This was a mistake with my and @alimpfard 's understanding with this. I'll get this fixed up.

ccapitalK:
@Quaker762 I would strongly recommend trying to get the current version merged in ASAP, development complexity of ongoing forks scales quadratically with time. :^)

Could you maybe fix up the license headers so that the CI passes, and then we'll try to get people to review it?

Quaker762:
> @Quaker762 I would strongly recommend trying to get the current version merged in ASAP, development complexity of ongoing forks scales quadratically with time. :^)
> 
> Could you maybe fix up the license headers so that the CI passes, and then we'll try to get people to review it? 
> 
> 

Agreed. I'm going to squash the license header fix into the last commit.

Userland/Libraries/LibGL/Clipper.cpp
```diff
@@ -0,0 +1,111 @@
...
+                }
+
+                clipped_polygon.append(curr_vec);
+            } else if (point_within_clip_plane(prev_vec, static_cast<ClipPlane>(plane))) {
+                FloatVector4 intersect = clip_intersection_point(prev_vec, curr_vec, static_cast<ClipPlane>(plane));
+                clipped_polygon.append(intersect);
+            }
+        }
+    }
+}
```
Quaker762, line 111: These changes seem to have reduced CPU utilisation of the library by around 30%!!!!!

ccapitalK: What exactly do you mean by this, are you stating that the CPU time spent on each frame has decreased by 30%, or that the CPU utilization percentage has decreased by 30%?

Quaker762: CPU usage as measured in `System Monitor` has reduced on average 30%, from 80 to about 50.

With regards to actually profiling, the amount of calls to `malloc`/`free` inside the clipping functions has gone from ~40% of execution time to 16%.

sunverwerth: > With regards to actually profiling, the amount of calls to malloc/free inside the clipping functions has gone from ~40% of execution time to 16%.

Move the Vector out of the function and into the class and clear them instead, then the number will drop to 0 :^)

Quaker762:
@awesomekling @linusg This is now ready for merge review. Let me know if you'd like the commit history to be anymore than it currently is and I'll fix it up. 

Tag @sunverwerth

awesomekling:
This is looking strong. Great work! Let's merge it and continue iterating in tree :)

Issue #6369, opened 2021-04-16
Kernel: There is a fundamental lack of ability to handle OOM


bgianfo:
This issue tracks the basic building blocks that are required to allow even the most basic handle of OOM in the Kernel.
At the moment the state of kmalloc + AK data structure are the biggest impediments, as even the most basic operations
blindly allocate without any way for the kernel to observe failure. 

Known Issues: 
- [ ] - `kmalloc()` intentionally PANIC's if the allocation fails at this point.
- [ ] - AK::Vector currently has no API for operations which can fail due to OOM.
- [ ] - AK::String currently has no API for operations which can fail due to OOM.
- [ ] - AK::HashTable currently has no API for operations which can fail due to OOM.

ADKaster:
Raises the question, what does normal `new` do when it fails to allocate in `-fno-exceptions` mode? Should the AK containers be passing `std::nothrow_t` for those try_XXX APIs?

bgianfo:
> Raises the question, what does normal `new` do when it fails to allocate in `-fno-exceptions` mode? Should the AK containers be passing `std::nothrow_t` for those try_XXX APIs?

Yea that's a good point, I think in a normal environment it calls `std::terminate()` on failure. Anything that wants to observe failure should use `std::nothrow` or alloc + placement `new`.

IdanHo:
Another known issue (introduced by myself :sweat_smile:) is the use of RedBlackTrees in VM Spaces (which has unfailable OOM) instead of IntrusiveRedBlackTrees

bgianfo:
> Another known issue (introduced by myself sweat_smile) is the use of RedBlackTrees in VM Spaces (which has unfailable OOM) instead of IntrusiveRedBlackTrees

Sure, added to the list. Thanks!

alimpfard:
This is a non-exhaustive list of things that allocate memory and assume that it works out.
Note that these are "root" cases, and there's a good chance that they're used in many places.

<details>
<summary>Expand for 470 items</summary>

- `AK/Bitmap.h:49:58`
- `AK/Bitmap.h:71:25`
- `AK/BitmapView.h:327:55`
- `AK/ByteBuffer.h:277:47`
- `AK/ByteBuffer.h:51:21`
- `AK/Demangle.h:47:29`
- `AK/FlyString.cpp:69:63`
- `AK/Format.h:242:8`
- `AK/HashMap.h:43:7`
- `AK/HashMap.h:56:5`
[...]

PR #6376, opened 2021-04-16
AK: Make IntrusiveList capable of holding any kind of pointer, not just a raw pointer

alimpfard:
This should allow creating intrusive lists that have smart pointers,
while remaining free (compared to the impl before this commit) when
holding raw pointers :^)
As a sidenote, this also adds a `RawPtr<T>` type, which is just
equivalent to `T*`.
Note that this does not actually use such functionality, but is only
expected to pave the way for #6369, to replace NonnullRefPtrVector<T>
with intrusive lists.

As it is with zero-cost things, this makes the interface a bit less nice
by requiring the type name of what an `IntrusiveListNode` holds (and
optionally its container, if not RawPtr), and also requiring the type of
the container (normally `RawPtr`) on the `IntrusiveList` instance.

cc @bgianfo - you asked for this :)

### DIFF
--- a/AK/IntrusiveList.h
+++ b/AK/IntrusiveList.h
@@ -28,20 +28,25 @@
 
 #include <AK/Assertions.h>
 #include <AK/BitCast.h>
+#include <AK/Forward.h>
+#include <AK/StdLibExtras.h>
 
 namespace AK {
 
+template<typename T, typename Container = RawPtr<T>>
 class IntrusiveListNode;
+
+template<typename T, typename Container>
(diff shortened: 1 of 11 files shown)


bgianfo (approved):

bgianfo:
Looks good to me! Thanks for picking this up.

alimpfard:
Ohhhh
Dammit, this needs some more work to make the template-templates work okay
Interesting that GCC 10.3 is actually okay with this 🤔

PR #6379, opened 2021-04-16
LibWeb: Fix height calculation for absolutely positioned boxes

Cleverking2003:


awesomekling (changes requested):
Please explain a little bit more in the commit message what you are doing. :)

PR #6380, opened 2021-04-16
Ports: Add auth_type verification to all package.sh files

bcoles:
Also adds a lint script to ensure all ports have `'port', 'version', 'files', 'auth_type'` properties.

This will eventually break any ports which use `master` branch from a git repository. Fortunately there's only a few of these. Also, these packages would have broken eventually anyway as the patches drift out of sync. They shouldn't have been using master in the first place, and instead used a known good and known safe version.

### DIFF
--- a/Meta/lint-ports.py
+++ b/Meta/lint-ports.py
@@ -36,7 +36,7 @@ def read_port_dirs():
     """Check Ports directory for unexpected files and check each port has a package.sh file.
 
     Returns:
-        list: all ports (set), errors encountered (bool)
+        list: all ports (set), no errors encountered (bool)
     """
 
     ports = set()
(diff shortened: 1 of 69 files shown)


Meta/lint-ports.py
```diff
@@ -45,35 +45,68 @@ def read_port_dirs():
...
+        if not os.path.exists(package_file):
+            continue
+        packages.add(package_file)
+
+    properties = ['port', 'version', 'files', 'auth_type']
+    for package in packages:
+        with open(package, 'r') as fp:
+            data = fp.read()
+            for p in properties:
+                if not re.findall(f"^{p}=", data, re.M):
```
linusg, line 83: oh that's a neat solution!

linusg (approved):
> They shouldn't have been using master in the first place, and instead used a known good and known safe version.

Totally agree with this, event if there hasn't been any change in recent years and no release at all they should still be pinned to a certain commit. Great work collecting all these hashes!

PR #6381, opened 2021-04-16
CI: Use clang-12 instead of clang-10 when building fuzzers

alimpfard:
oss-fuzz uses clang-12 anyway, so this patch shouldn't be breaking
anything, just letting us use more modern C++ without the CI being
sad.


linusg:
CI is happy, let's do it!

PR #6393, opened 2021-04-17
LibGfx: Make top highlight of classic buttons fit to the top left corner

Lubrsi:
It didn't quite fit before:
Screenshot from 2021-04-17 02-25-55

After:
Screenshot from 2021-04-17 02-26-20


awesomekling:
The previous look was an intentional clone of the standard button control look from Windows 2000.

It had two exceptions: the window buttons (the ones in your comparison screenshot) and the start menu button. Those had a thinner highlight in Win2K just like we do now as well.

This particular highlight has always bothered me in Serenity because I didn't like how the start menu button and the window buttons looked with the extra padding around the highlight, but I didn't want to make this change since it affects *all* buttons in the system.

Now that you've gone ahead and made the change anyway, I think all buttons in the system look slightly nicer, so thanks for making this happen :^)

PR #6394, opened 2021-04-17
Toolchain: Implement support for exception handling in usermode programs

gunnarbeutner:
This enables building usermode programs with exception handling. It also builds a `libstdc++` library without exception support for the kernel.

This is necessary because the `libstdc++` that gets built is different when exceptions are enabled. Using the same library binary would require extensive stubs for exception-related functionality in the kernel.

This PR also updates the kernel to allow system calls from the dynamic loader. Previously the dynamic loader would become unused after it had invoked the program's entry function. However, in order to support exceptions and - at a later point - dlfcn functionality we need to call back into the dynamic loader at runtime.

Because the dynamic loader has a static copy of LibC it'll attempt to invoke syscalls directly from its text segment. For this to work the executable region for the dynamic loader needs to have syscalls enabled.


awesomekling:
Damn, really nice work! :star:

Issue #6391, opened 2021-04-17
SerenityOS freezes when tries to open .so files.


xeriab:
I was jumping around in `FileManager` and when I accidentally hit `enter` on an `.so` file the whole system froze and I had to kill it.

I guess there is no such way to handle such a thing so the system spat the following error and froze:

```
[FileManager(28:28)]: ASSERTION FAILED: m_total_range.contains(allocated_range)
[FileManager(28:28)]: ../../Kernel/VM/RangeAllocator.cpp:165 in AK::Optional<Kernel::Range> Kernel::RangeAllocator::allocate_specific(VirtualAddress, size_t)
[#0 FileManager(28:28)]: 0xc053a514  .L1166 +0x91c
[#0 FileManager(28:28)]: 0xc050054a  Kernel::VolatilePageRanges::remove(Kernel::VolatilePageRange const&, bool&) +0x3ea
[#0 FileManager(28:28)]: 0xc05168c5  Kernel::Region::handle_inode_fault(unsigned long, Kernel::ScopedSpinLock<Kernel::RecursiveSpinLock>&) +0x1ead
[#0 FileManager(28:28)]: 0xc03a402e  Kernel::make_userspace_stack_for_main_thread(Kernel::Region&, AK::Vector<AK::String, 0ul>, AK::Vector<AK::String, 0ul>, AK::Vector<ELF::AuxiliaryValue, 0ul>)::{lambda(AK::String const&)#3}::operator()(AK::String const&) const [clone .isra.0] +0x248
[#0 FileManager(28:28)]: 0xc03a5cf2  Kernel::get_required_load_range(Kernel::FileDescription&) +0x745
[#0 FileManager(28:28)]: 0xc03a81ad  Kernel::load_elf_object(AK::NonnullOwnPtr<Kernel::Space>, Kernel::FileDescription&, unsigned int, Kernel::ShouldAllocateTls) +0xf0
[#0 FileManager(28:28)]: 0xc03b167f  Kernel::Process::do_exec(AK::NonnullRefPtr<Kernel::FileDescription>, AK::Vector<AK::String, 0ul>, AK::Vector<AK::String, 0ul>, AK::RefPtr<Kernel::FileDescription, AK::RefPtrTraits<Kernel::FileDescription> >, Kernel::Thread*&, unsigned int&, elfhdr const&) +0x4e39
[...]

PR #6396, opened 2021-04-17
Kernel: execve: Reject execution of /usr/lib/Loader.so dynamic loader

xeriab:
The PR is intended to fix the problem mentioned in #6391 issue.

I'm not totally sure this is the proper way, but I hope it is.

### DIFF
--- a/Kernel/Syscalls/execve.cpp
+++ b/Kernel/Syscalls/execve.cpp
@@ -793,7 +793,7 @@ KResultOr<RefPtr<FileDescription>> Process::find_elf_interpreter_for_executable(
         // If it's ET_DYN with no PT_INTERP, then it's a dynamic executable responsible
         // for its own relocation (i.e. it's /usr/lib/Loader.so)
         if (path != "/usr/lib/Loader.so") {
-            dbgln("exec(={}): WARNING - Dynamic ELF executable without a PT_INTERP header, and isn't /usr/lib/Loader.so", path);
+            dmesgln("exec(={}): WARNING - Dynamic ELF executable without a PT_INTERP header, and isn't /usr/lib/Loader.so", path);
         } else {
             dbgln("exec(={}): WARNING - Hold on, you cannot execute the loader!", path);
         }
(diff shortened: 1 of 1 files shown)


Kernel/Syscalls/execve.cpp
```diff
@@ -792,9 +792,13 @@ KResultOr<RefPtr<FileDescription>> Process::find_elf_interpreter_for_executable(
     if (main_program_header.e_type == ET_DYN) {
         // If it's ET_DYN with no PT_INTERP, then it's a dynamic executable responsible
         // for its own relocation (i.e. it's /usr/lib/Loader.so)
-        if (path != "/usr/lib/Loader.so")
-            dbgln("exec({}): WARNING - Dynamic ELF executable without a PT_INTERP header, and isn't /usr/lib/Loader.so", path);
-        return nullptr;
+        if (path != "/usr/lib/Loader.so") {
+            dbgln("exec(={}): WARNING - Dynamic ELF executable without a PT_INTERP header, and isn't /usr/lib/Loader.so", path);
```
bgianfo, line 796: I think it would be good to make these `dmesgln(..)` instead of `dbgln(..)` so they are visible to the user inside the OS, if they run `dmesg`. 
```suggestion
            dmesgln("exec(={}): WARNING - Dynamic ELF executable without a PT_INTERP header, and isn't /usr/lib/Loader.so", path);
```

bgianfo:
I think this looks reasonable, I tried it and confirmed the fix locally as well.

bgianfo (approved):

xeriab:
Thank you so much for reviewing it.

bcoles:
As per the [contributing guidelines](https://github.com/SerenityOS/serenity/blob/master/CONTRIBUTING.md) please squash commits and wrap commit messages at 72 characters.

Suggest something like:

```
Kernel: execve: Reject execution of /usr/lib/Loader.so dynamic loader
```

xeriab:
@bcoles Thank you for the review; I will take care from now and so on + I will squash the commit message to match the guidelines :)

PR #6397, opened 2021-04-17
Kernel+LibCore+FileOperation: Copy files without kernel/userland copies

lpereira:
Introduce a new system call, `fdcopy(int srcfd, int dstfd, ssize_t count)`, to copy between file descriptors without passing through user mode.  With it, implement `IODevice::copy_from()` and use it in FileOperation to copy files.

Creating this as a draft as I'm not sure yet about the details of this interface, but this seems to be working just fine.

### DIFF
--- a/Userland/Libraries/LibC/unistd.cpp
+++ b/Userland/Libraries/LibC/unistd.cpp
@@ -295,6 +295,12 @@ ssize_t write(int fd, const void* buf, size_t count)
     __RETURN_WITH_ERRNO(rc, rc, -1);
 }
 
+ssize_t fdcopy(int srcfd, int dstfd, size_t count)
+{
+    int rc = syscall(SC_fdcopy, srcfd, dstfd, count);
+    __RETURN_WITH_ERRNO(rc, rc, -1);
+}
+
 int ttyname_r(int fd, char* buffer, size_t size)
 {
     int rc = syscall(SC_ttyname, fd, buffer, size);
--- a/Userland/Libraries/LibCore/IODevice.cpp
+++ b/Userland/Libraries/LibCore/IODevice.cpp
@@ -61,6 +61,32 @@ int IODevice::read(u8* buffer, int length)
     return read_buffer.size();
 }
 
+size_t IODevice::copy_from(const IODevice& source, size_t buffer_size)
+{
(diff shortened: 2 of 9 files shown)


Userland/Libraries/LibC/unistd.cpp
```diff
@@ -295,6 +295,12 @@ ssize_t write(int fd, const void* buf, size_t count)
     __RETURN_WITH_ERRNO(rc, rc, -1);
 }
 
+ssize_t fdcopy(int srcfd, int dstfd, size_t count)
```
bgianfo, line 298: Your new syscall defines count as `ssize_t`, typo here? 

```suggestion
ssize_t fdcopy(int srcfd, int dstfd, ssize_t count)
```

Userland/Libraries/LibCore/IODevice.cpp
```diff
@@ -61,6 +61,32 @@ int IODevice::read(u8* buffer, int length)
...
 }
 
+size_t IODevice::copy_from(const IODevice& source, size_t buffer_size)
+{
+#ifndef __serenity__
+    (void)source;
+    (void)buffer_size;
+    return 0;
+#else
+    if (!source.is_readable())
```
bgianfo, line 71: Is there a reason for doing the validation here instead of letting the kernel reject it? Is there some common error path that you are trying to optimize for?

Userland/Libraries/LibCore/IODevice.h
```diff
@@ -84,6 +86,8 @@ class IODevice : public Object {
     ByteBuffer read_all();
     String read_line(size_t max_size = 16384);
 
+    size_t copy_from(const IODevice& source, size_t buffer_size = 65536);
```
bgianfo, line 89: This buffer size constant seems to be shared between the kernel and usermode, maybe it should go in a shared header?

Userland/Libraries/LibC/unistd.cpp
ccapitalK: I think it would make more sense to make it a size_t in the syscall definition as well, since the syscall returns EINVAL if it's negative (and there's no point being able to represent negative values if they are all rejected).

PR #6398, opened 2021-04-17
LibTemplate: Add Jinja-like Template engine

ccapitalK:
This pr adds a very simple jinja-like template engine to serenity. Template parameters are passed through a json object.

The primary motivation of this pr is to remove a fairly unsafe usage of printf in the browser. Currently, the browser generates its error page by loading a printf string **from disk**, then passing that into printf as an ad-hoc form of templating. Now, we use an actual (if simple) template processor for it, which actually validates the template instead of just allowing arbitrary printf shenanigans.


awesomekling:
I believe this functionality is already provided by `AK::SourceGenerator`, which should probably have a more generic name. :)

ccapitalK:
> I believe this functionality is already provided by `AK::SourceGenerator`, which should probably have a more generic name. :)

Oh, I wasn't aware of SourceGenerator. Looking at it, it seems that SourceGenerator does 90% of what I want, the only difference is that it seems more suited for situations where the template patterns are in source code, not in an external file. I think I'll have to rethink this PR a bit.

PR #6399, opened 2021-04-17
Ports: c-ray: Pin version to latest commit rather than use master branch

bcoles:
Prior to this PR, the port is broken due to code rot. The `replace-micro-symbol.patch` patch no longer applies.


linusg (commented):
Let's also add the first 7 characters of the commit hash as version to `AvailablePorts.md`, like `openssl` (`8.3-9ca7e9c`) and `mrsh` (`d9763a3`).

PR #6403, opened 2021-04-17
Ports: Enable building cmake on the host

gunnarbeutner:
This depends on #6394.


alimpfard:
Does this actually bootstrap now? last I tried it depended partially on the host, and partially on the target

gunnarbeutner:
It worked for me. Give it a try if you want - you'll also need the changes from that other PR though (which requires a complete rebuild).

alimpfard:
I tested this, and it also works for me :tada:

PR #6409, opened 2021-04-17
LibJS: Convert matched regex result to string in Symbol.replace

IdanHo:
This would crash on an undefined match, since the matched result was assumed to be a string (such as on discord.com). The spec suggests converting it to a string as well: https://tc39.es/ecma262/#sec-regexp.prototype-@@replace (14#c)

(I tried writing a test for it, but i wasnt able to create a regex that caused this? and the discord js bundle is much too complicated for me to find the regex they used)

### DIFF
--- a/Userland/Libraries/LibJS/Runtime/RegExpPrototype.cpp
+++ b/Userland/Libraries/LibJS/Runtime/RegExpPrototype.cpp
@@ -346,7 +346,11 @@ JS_DEFINE_NATIVE_FUNCTION(RegExpPrototype::symbol_replace)
         size_t result_length = length_of_array_like(global_object, result);
         size_t n_captures = result_length == 0 ? 0 : result_length - 1;
 
-        auto matched = result.get(0).value_or(js_undefined());
+        auto matched_value = result.get(0).value_or(js_undefined());
+        if (vm.exception())
+            return {};
+
+        auto matched = matched_value.to_string(global_object);
         if (vm.exception())
             return {};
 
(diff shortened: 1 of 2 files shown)


Userland/Libraries/LibJS/Runtime/RegExpPrototype.cpp
```diff
@@ -346,7 +346,7 @@ JS_DEFINE_NATIVE_FUNCTION(RegExpPrototype::symbol_replace)
         size_t result_length = length_of_array_like(global_object, result);
         size_t n_captures = result_length == 0 ? 0 : result_length - 1;
 
-        auto matched = result.get(0).value_or(js_undefined());
+        auto matched = result.get(0).value_or(js_undefined()).to_string(global_object);
```
linusg, line 349: This cannot be in the same expression but needs to be split into multiple lines unfortunately - get can throw, and `to_string()` can call a JS function which will go terribly wrong (:boom:) if we already have an exception stored.

alimpfard:
Group 1 in `/x(a)??/` should be undefined.
In general, `??` will definitely match nothing if it's at the end of the pattern - useful to know for cases like this :P

linusg (commented):
Nice fix! Regarding a test, here's a snippet that crashes without this patch but works correctly without, maybe you can use that?

```js
const r = /foo/;
r.exec = () => ({});
r[Symbol.replace]();  // "undefined"
```

Userland/Libraries/LibJS/Runtime/RegExpPrototype.cpp
alimpfard: Maybe `get()` should return an `ExceptionOr`?
Would then force people to check for exceptions

IdanHo: fixed! thanks for the test as well :)

Issue #6410, opened 2021-04-17
LibC: Wine checklist


sin-ack:
This is a list of things that need to be implemented in order get Wine running. Note that we still need to implement a display backend for `WindowServer` as Wine uses X11 directly, and also an `AudioServer` connection for audio playback. Most of this functionality is required by the `dlls/ntdll/unix/` directory since that seems to be where all the platform-related functionality is residing.

This list also isn't complete because it's just what I gathered from running `make --keep-going` so we might need other things.

- [x] `n_long`/`n_short` (Linux puts these in `<netinet/in_systm.h>`) - `dlls/iphlpapi/icmp.c`
- [ ] `IPPROTO_ICMPV6` and `IP_TTL` (`<netinet/in.h>`) - `dlls/iphlpapi/icmp.c`
- [x] `<langinfo.h>` - `dlls/ntdll/unix/env.c`
- [x] `MSG_PEEK` (`<socket.h>`) - `dlls/ntdll/unix/file.c`
- [ ] `statvfs` and associated data structures (tracked in #6068) - `dlls/ntdll/unix/file.c`
- [ ] `rlimit` related functionality (`<unistd.h>`) - `dlls/ntdll/unix/loader.c`
- [x] `Dl_info`, `dladdr` (`<dlfcn.h>`) - `dlls/ntdll/unix/loader.c`
  (Most of the dynamic linker stuff seems to be implemented already, and just needs to be exposed)
- [ ] ELF related things (`AT_PHDR`, `AT_PHENT`, `AT_PHNUM`, `Elf32_Phdr`) - `dlls/ntdll/unix/loader.c`
- [ ] Signal context functions (`ucontext_t` and friends) - `dlls/ntdll/unix/signal_i386.c`
- [x] `pthread_addr_setscope` and `PTHREAD_SCOPE_SYSTEM` - `dlls/ntdll/unix/thread.c`
- [ ] Exposing `struct iovec` - `dlls/ntdll/unix/virtual.c`
- [ ] `mlock`/`munlock` and `madvise` - `dlls/ntdll/unix/virtual.c`

Hendiadyoin1:
n_long and n_short seem to be just typedefines for u32 and u16 respectivly, with the caviats that they are supposed to be guaranteed to be in big endian ("high-ender" order) if I understand it correctly

Hendiadyoin1:
`<netinet/in_systm.h>`seems to declare another funciton called iptime(), but I only found that in BSD, is that required too

[...]

PR #6415, opened 2021-04-17
LibC: Add netinet/in_systm.h

Hendiadyoin1:
Work towards #6410

Adds:
* n_short (u16)
* n_long (u32)
* n_time (u32 - time since GMT 00:00)

BSD seems to declare another function in this header, but it does not seem to be used by any other Unix system

### DIFF
--- a/Userland/Libraries/LibC/netinet/in_systm.h
+++ b/Userland/Libraries/LibC/netinet/in_systm.h
@@ -1,0 +1,39 @@
+/*
+ * Copyright (c) 2021, Leon Albrecht <leon2002.la@gmail.com>
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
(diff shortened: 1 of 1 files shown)


Userland/Libraries/LibC/netinet/in_systm.h
```diff
@@ -0,0 +1,42 @@
...
+
+__BEGIN_DECLS
+
+typedef uint16_t n_short;
+typedef uint32_t n_long;
+
+typedef uint32_t n_time;
+
+// BSD implements a `n_time iptime()` function for the kernel 
+// to generate a n_time with the ms since GMT 00:00
```
sin-ack, line 40:
```suggestion
// BSD implements a `n_time iptime()` function for the kernel
// to generate a n_time with the ms since GMT 00:00
```

Hendiadyoin1: Looked at this change on my phone for a minute wondering what changed....

Userland/Libraries/LibC/netinet/in_systm.h
```diff
@@ -0,0 +1,42 @@
...
+
+__BEGIN_DECLS
+
+typedef uint16_t n_short;
+typedef uint32_t n_long;
+
+typedef uint32_t n_time;
+
+// BSD implements a `n_time iptime()` function for the kernel
+// to generate a n_time with the ms since GMT 00:00
```
awesomekling, line 40: What is the purpose of this comment? Is it meant to be a FIXME?
Also `a n_time` -> `an n_time`.

Hendiadyoin1: General information about the header and compatablity for future development, I know that BSD uses and implemetns it in the icmp_input function

OK thats an edgcase of a/an that I have never encountered before; will fix that

Hendiadyoin1: with the typo you mean the second line?

awesomekling: Both lines are typos. It's "an n_time", not "a n_time" :)

awesomekling: This information doesn't seem relevant since it refers to an API we don't implement.
If this is something we should implement, we could have a comment saying e.g "FIXME: Implement iptime() like BSD"

Hendiadyoin1: Then I'll remove it for now

PR #6419, opened 2021-04-17
LibLine: ^X^E - edit in external editor

alimpfard:
Adds support for ^X^E, not sure how we've been living without this so far!

Also fixes a dumb case where LibLine would miss a WINCH signal while another program is in the foreground.
It can be reproduced like so (prior to the changes in this PR)
- `$ te`
- resize the terminal while `te` is running
- close `te`
- now the shell doesn't know that the terminal size has changed

### DIFF
--- a/Userland/Libraries/LibLine/Editor.cpp
+++ b/Userland/Libraries/LibLine/Editor.cpp
@@ -62,6 +62,7 @@ Configuration Configuration::from_config(const StringView& libname)
     // Read behaviour options.
     auto refresh = config_file->read_entry("behaviour", "refresh", "lazy");
     auto operation = config_file->read_entry("behaviour", "operation_mode");
+    auto default_text_editor = config_file->read_entry("behaviour", "default_text_editor");
 
     if (refresh.equals_ignoring_case("lazy"))
         configuration.set(Configuration::Lazy);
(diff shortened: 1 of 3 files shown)


Userland/Libraries/LibLine/Editor.cpp
```diff
@@ -168,6 +174,9 @@ void Editor::set_default_keybinds()
     register_key_input_callback(ctrl('T'), EDITOR_INTERNAL_FUNCTION(transpose_characters));
     register_key_input_callback('\n', EDITOR_INTERNAL_FUNCTION(finish));
 
+    // ^X^E: Edit in external editor
+    register_key_input_callback(Vector<Key> { ctrl('X'), ctrl('E') }, EDITOR_INTERNAL_FUNCTION(edit_in_external_editor));
```
linusg, line 178: I've never looked at LibLine in detail, that's a nice API you got there! :D

Userland/Libraries/LibLine/InternalFunctions.cpp
```diff
@@ -522,4 +526,91 @@ void Editor::uppercase_word()
...
+        perror("mktemp");
+        return;
+    }
+
+    {
+        auto* fp = fdopen(fd, "rw");
+        if (!fp) {
+            perror("fdopen");
+            return;
+        }
```
awesomekling, line 548: This gunk makes me feel like we should have something in LibCore for creating a temp file.

IdanHo: we have similar (but slightly different) gunk in https://github.com/SerenityOS/serenity/blob/master/Userland/Utilities/mktemp.cpp

alimpfard: Definitely should, +1.

PR #6420, opened 2021-04-17
LibWebSocket: Added basic support for WebSockets

Dexesttp:
# Overview

This Pull Requests adds :
- The `LibWebSocket` library, which allows a client to create a connection to a WebSocket server.
- The `tlsws` command-line utility, which allows communicating with a WebSocket via a ~~very barebones~~ slightly less barebones REPL.

The current architecture supports both the WebSocket (`ws://`) and Secure WebSocket (`wss://`) protocols.

## Limitations

The following actions are not handled according to the WebSocket specification:
- There is currently no support for fragmented WebSocket frames.
- The client-server HTTP handshake doesn't currently follow redirects, but fails the connection instead.
- The client-server HTTP handshake doesn't currently follow additional authentification requests, but fails the connection instead.
- TLS certificate requests are not currently part of the API and cannot be handled (since the underlying protocol of Secure WebSockets is TLSv1.2). Since they are not implemented anyways yet, this didn't seem like a high priority.

## Screnshot

screenshot


Userland/Utilities/telws.cpp
```diff
@@ -0,0 +1,105 @@
...
+#include <LibCore/Notifier.h>
+#include <LibWebSocket/ConnectionInfo.h>
+#include <LibWebSocket/Message.h>
+#include <LibWebSocket/WebSocket.h>
+
+int main(int argc, char** argv)
+{
+    Core::ArgsParser args_parser;
+
+    const char* origin = nullptr;
```
linusg, line 41: I think you can use a String for this as well, then there's no need to convert to construct one later (`connection_info.set_origin(String(origin));`).

Userland/Utilities/telws.cpp
```diff
@@ -0,0 +1,105 @@
...
+    args_parser.add_option(origin, "URL to use as origin", "origin", 'o', "origin");
+
+    args_parser.parse(argc, argv);
+
+    URL url(url_string);
+
+    if (!url.is_valid()) {
+        warnln("The given URL is not valid");
+        return 1;
+    }
```
linusg, line 54: Should we also check the protocol (ws/wss)?

Userland/Utilities/telws.cpp
```diff
@@ -0,0 +1,105 @@
...
+#include <LibCore/ArgsParser.h>
+#include <LibCore/EventLoop.h>
+#include <LibCore/File.h>
+#include <LibCore/Notifier.h>
+#include <LibWebSocket/ConnectionInfo.h>
+#include <LibWebSocket/Message.h>
+#include <LibWebSocket/WebSocket.h>
+
+int main(int argc, char** argv)
+{
```
linusg, line 38: appropriate pledge/unveil calls would be nice :)

Userland/Utilities/telws.cpp
```diff
@@ -0,0 +1,105 @@
...
+            code);
+        loop.quit(0);
+    };
+    socket->start();
+
+    auto stdin = Core::File::standard_input();
+    auto stdin_watcher = Core::Notifier::construct(stdin->fd(), Core::Notifier::Read);
+    stdin_watcher->on_ready_to_read = [&] {
+        if (stdin->can_read_line()) {
+            auto line = stdin->read_line();
```
alimpfard, line 88: Since this is line-based, LibLine can probably give you a better interface, though it might mess with the event loop too much.
(this is coming from "a barebones REPL", trying to get rid of the "barebones" lol)

Userland/Utilities/telws.cpp
Dexesttp: That's actually a good question. RFC 6455 doesn't say anything about whether other protocols are valid to open a WebSocket from, just that several components of the URL must be valid ([Page 15, point 1](https://tools.ietf.org/html/rfc6455#page-15)) - but the protocol is not a part of that check.

Userland/Utilities/telws.cpp
Dexesttp: Fun fact, while there is a for `ArgsParser::add_positional_argument(String&)`, there isn't an `ArgsParser::add_option(String&)` API ^^' I tried that at first, and spent a minutes wondering why it wouldn't compile.
Anyways, I'll add the API to ArgsParser and switch to that ;^)

Userland/Utilities/telws.cpp
Dexesttp: I looked into it, and I think `Line::Editor::get_line` is blocking, right ? That might not mesh well with the event loop, since the server will regularily ping the client to maintain the connection.
In any wase, integrating telws with LibLine seems like a big task, and a bit out of scope. The `telws` program is mostly temporary anyways, since most of the work on WebSockets should happen on the JS side. So it's hopefully not too bad if it's barebones !

alimpfard: it will keep the event loop going, LibLine itself is event-based : )
The integration should just be 
```c++
auto editor = Line::Editor::construct();
...
auto line_or_error = editor->get_line();
```
if that doesn't work, then nvm.

Dexesttp: Ok, it "works", but libline reallly doesn't like when stuff outputs to the terminal while it's waiting for a prompt. Added it anyways :)

Userland/Utilities/telws.cpp
Dexesttp: Done !

Userland/Utilities/telws.cpp
Dexesttp: Done :) The new argsparser method would be useful in a bunch of other places, but we'll let things improve bit by bit on that end.

Dexesttp:
I fixed an encoding bug when sending payloads that were more than 126 bytes long, and tested some more use cases to make sure everything worked as expected !
And since the other feedbacks were added, the PR should now be ready :)

linusg (commented):
Great work!

Userland/Libraries/LibWebSocket/ConnectionInfo.h
```diff
@@ -0,0 +1,74 @@
...
+namespace WebSocket {
+
+class ConnectionInfo final {
+public:
+    ConnectionInfo(URL);
+
+    URL const& url() const { return m_url; }
+
+    String const& origin() const { return m_origin; }
+    void set_origin(String origin) { m_origin = origin; }
```
linusg, line 45: Since we're already passing the string by value, would this (and the setters below) benefit from `move()`?

```suggestion
    void set_origin(String origin) { m_origin = move(origin); }
```

Userland/Libraries/LibWebSocket/WebSocket.cpp
```diff
@@ -0,0 +1,597 @@
...
+    default:
+        VERIFY_NOT_REACHED();
+        return WebSocket::ReadyState::Closed;
+    }
+}
+
+void WebSocket::send(Message message)
+{
+    if (m_state != WebSocket::InternalState::Open)
+        return;
```
linusg, line 107: Maybe this should `VERIFY()` instead? Calling `send()` on a non-open websocket seems like something that shouldn't happen, and just returning silently seems kinda defensive..

linusg (commented):
and apparently GitHub lost some of my comments, so here are they again - apologies if they suddenly appear twice somehow..

Userland/Libraries/LibWebSocket/ConnectionInfo.h
Dexesttp: Good point ! I'm still not that comfortable with move, but that's a place where it definitely should be :)
I also added them to the constructor (to move the URL), and passed the `ConnectionInfo` as a `const&` instead of a raw value in the `AbstractWebSocketImpl::connect()` and subclasses to prevent some additional copies.

Userland/Libraries/LibWebSocket/WebSocket.cpp
Dexesttp: That makes sense. Since it might be surprising to clients who are using the `send()` and `close()` methods when these only work on opened sockets, I added some comments explaining that in the `WebSocket.h` header. And I updated the `telws` program to not crash when trying that :)

Dexesttp:
I also added full support for sending masked frames, as the spec requires for clients to mask their frames. The changes can be found [here](https://github.com/SerenityOS/serenity/pull/6420/files#diff-3c4b40709ac29fd08ac2145067b01eadf5144f5aa16500e0d210fce1a83dfb4bR553).
I left in the support for sending non-masked frames, since eventually that code (and the send_frame code) could be factored out when dealing with a WebSocket server :)
And with that, it should be good again !

linusg (approved):
:+1: from me, but I'll leave the final review to @awesomekling :)

PR #6421, opened 2021-04-17
Ports: AvailablePorts.md: Fix column alignment

bcoles:


linusg:
since we're already reformatting the whole table, maybe it would make sense to add some additional whitespace for future long URLs / descriptions so this is less likely to become a problem again?

bcoles:
> 
> 
> since we're already reformatting the whole table, maybe it would make sense to add some additional whitespace for future long URLs / descriptions so this is less likely to become a problem again?

yes. suggestions?

current size including the one space padding on either side:

```
|--------------------------------|-------------------------------------------------|-------------------|---------------------------------------------------------|
| 33                             | 50                                              | 23                | 61                                                      |
```

linusg:
Bumping to something like 40, 60, 30, 80 is probably fine..

bcoles:
> 
> 
> Bumping to something like 40, 60, 30, 80 is probably fine..

Done. I shaved 3 bytes off the version column length suggestion. This prevents line wrapping/overflow on my system.

PR #6423, opened 2021-04-17
Kernel: Add boot argument to disable the UHCI Controller

Lubrsi:
Helps with bare metal debugging, as we can't be sure our implementation
will work with a given machine.

As reported by someone on Discord, their machine hangs when we attempt
the dummy transfer.


Lubrsi:
Just realised I should add it to the documentation, marking it draft as I do that.

PR #6424, opened 2021-04-17
Toolchain: Don't produce debug symbols

xackus:
Previously debug symbols were produced and then stripped.


linusg:
This has merge conflicts, please rebase.

PR #6426, opened 2021-04-17
Browser: Display full UserAgent string in status bar on menu hover

IdanHo:
This uses the new on_action_enter & on_action_leave APIs to display the full useragent string when hovering over one of the useragent spoof menu options.

image


linusg:
That's super cool! :D

PR #6427, opened 2021-04-17
Shell: Add a posix sh compatablilty mode.

Petelliott:
Shell will run in `--posix-mode` by default when invoked as `sh` or `/bin/sh`

features added in this PR:
- [x] if
- [ ] case
- [ ] for


alimpfard:
It probably would've been better to write a new parser tbh...
would've at least let you split it into multiple PRs if you wanted, and honestly...passing the shell options to the parser is a little hacky :P

Petelliott:
your right, it just gets uglier with some of the more complicated statements.

PR #6430, opened 2021-04-17
ls: Remove extra spaces after filenames when piping

### LINKED ISSUE #5671: ls: Prints extra spaces after filename when piping output
To see the spaces, try `ls | xargs printf "'%s'\n"`:

```sh
@courage:~ $ ls | xargs printf "'%s'\n"
'Desktop       '
'Documents     '
...
'tests         '
'web-tests'    
```

(For some reason `web-tests` doesn't have extra spaces?)

jagger27:
Fixes #5671

### DIFF
--- a/Userland/Utilities/ls.cpp
+++ b/Userland/Utilities/ls.cpp
@@ -75,6 +75,8 @@ static bool output_is_terminal = false;
 static HashMap<uid_t, String> users;
 static HashMap<gid_t, String> groups;
 
+static bool is_a_tty = false;
+
 int main(int argc, char** argv)
 {
     if (pledge("stdio rpath tty", nullptr) < 0) {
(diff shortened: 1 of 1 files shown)


Userland/Utilities/ls.cpp
```diff
@@ -484,7 +484,8 @@ static bool print_names(const char* path, size_t longest_name, const Vector<Stri
         printed_on_row += column_width;
 
         for (size_t j = nprinted; i != (names.size() - 1) && j < column_width; ++j)
-            printf(" ");
+            if (isatty(STDOUT_FILENO))
```
awesomekling, line 487: Please cache the result of `isatty(STDOUT_FILENO)` somewhere instead of calling it over and over.
Also please wrap multi-line `for` blocks in `{` and `}`.

awesomekling (changes requested):

PR #6431, opened 2021-04-17
LibWeb: Support ordered lists as well as different list-style-types for unordered lists

### LINKED ISSUE #2059: LibWeb: Ordered list is rendered as unordered list
HTML ordered list `<ol>` is rendered incorrectly as unordered list `<ul>`.
This naturally affects the Markdown ordered list, which wants to be rendered as `<ol>`.
```html
<!DOCTYPE html>
<html>
    <head>
    </head>
    <body>
        <ol>
            <li>This is first list item</li>
            <li>This is second list item</li>
            <li>This is third list item</li>
        </ol>
    </body>
</html>
```
Expected behavior:
image

Actual behavior:
image

TobyAsE:
This PR implements the display of ordered lists in the web browser (just decimal counting for now).
There is a pretty old issue #2059 discussing the matter but it seems it never got anywhere.

While I was at it, it was easy to also support the different list-style-types for unordered lists:
image

There are a lot more list-style-types out there, but they aren't supported by the parser yet.

Fixes #2059

### DIFF
--- a/Userland/Libraries/LibWeb/Layout/ListItemMarkerBox.cpp
+++ b/Userland/Libraries/LibWeb/Layout/ListItemMarkerBox.cpp
@@ -45,11 +45,39 @@ void ListItemMarkerBox::paint(PaintContext& context, PaintPhase phase)
 {
     if (phase != PaintPhase::Foreground)
         return;
-    Gfx::IntRect bullet_rect { 0, 0, 4, 4 };
-    bullet_rect.center_within(enclosing_int_rect(absolute_rect()));
+
     // FIXME: It would be nicer to not have to go via the parent here to get our inherited style.
     auto color = parent()->computed_values().color();
-    context.painter().fill_rect(bullet_rect, color);
+
+    auto enclosing = enclosing_int_rect(absolute_rect());
+    int marker_width = (int)enclosing.height() / 2;
+    Gfx::IntRect marker_rect { 0, 0, marker_width, marker_width };
+    marker_rect.center_within(enclosing);
+
+    switch (m_list_style_type) {
+    case CSS::ListStyleType::Square:
+        context.painter().fill_rect(marker_rect, color);
+        break;
+    case CSS::ListStyleType::Circle:
+        // For some reason for draw_ellipse() the ellipse is outside of the rect while for fill_ellipse() the ellipse is inside.
+        // Scale the marker_rect with sqrt(2) to get an ellipse arc (circle) that appears as if it was inside of the marker_rect.
+        marker_rect.set_height(marker_rect.height() / 1.41);
+        marker_rect.set_width(marker_rect.width() / 1.41);
+        marker_rect.center_within(enclosing);
+        context.painter().draw_ellipse_intersecting(marker_rect, color);
+        break;
+    case CSS::ListStyleType::Decimal:
+        context.painter().draw_text(enclosing, String::formatted("{}.", m_index), Gfx::TextAlignment::Center);
+        break;
+    case CSS::ListStyleType::Disc:
+        context.painter().fill_ellipse(marker_rect, color);
+        break;
+    case CSS::ListStyleType::None:
+        return;
+
+    default:
+        VERIFY_NOT_REACHED();
+    }
 }
 
 }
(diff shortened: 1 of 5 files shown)


Userland/Libraries/LibWeb/Layout/ListItemMarkerBox.cpp
```diff
@@ -44,11 +44,40 @@ void ListItemMarkerBox::paint(PaintContext& context, PaintPhase phase)
...
+        break;
+    case CSS::ListStyleType::Decimal:
+        context.painter().draw_text(enclosing, String::formatted("{}.", m_index), Gfx::TextAlignment::Center);
+        break;
+    case CSS::ListStyleType::Disc:
+        context.painter().fill_ellipse(marker_rect, color);
+        break;
+    case CSS::ListStyleType::None:
+        return;
+        break;
```
awesomekling, line 76: No need for both `return` and `break` :)

IdanHo:
You should add "Fixes #2059" to your PR message, so once this is merged that issue will be automatically closed :)

linusg:
@IdanHo it's in the third commit :)

Userland/Libraries/LibWeb/TreeNode.h
```diff
@@ -97,6 +97,33 @@ class TreeNode : public Weakable<T> {
         return const_cast<TreeNode*>(this)->child_at_index(index);
     }
 
+    int child_index(const T& search_child)
```
awesomekling, line 100: Please return `Optional<size_t>` instead of `int` with a magic `-1` value.
And let's call it `index_of_child()` to emphasize "index" a little extra.
And also, maybe we can `VERIFY(search_child.parent() == this);` while we're here :)

IdanHo:
> @IdanHo it's in the third commit :)

ah i see! the "no linked issues" on the sidebar tricked me :^)

awesomekling (changes requested):

Userland/Libraries/LibWeb/Layout/ListItemMarkerBox.cpp
TobyAsE: Good point ;)

Userland/Libraries/LibWeb/TreeNode.h
```diff
@@ -97,6 +97,33 @@ class TreeNode : public Weakable<T> {
...
+    {
+        int index = 0;
+        auto* child = first_child();
```
Lubrsi, line 103: There may not be any children, which will cause first_child to return nullptr and cause a null deref when trying to get the next sibling in the while loop.

Userland/Libraries/LibWeb/TreeNode.h
```diff
@@ -97,6 +97,33 @@ class TreeNode : public Weakable<T> {
...
+            index++;
+        } while ((child = child->next_sibling()));
+        return -1;
+    }
+
+    template<typename ChildType>
+    int child_index(const T& search_child)
+    {
+        int index = 0;
+        auto* child = first_child();
```
Lubrsi, line 116: Ditto

Lubrsi (changes requested):

Userland/Libraries/LibWeb/TreeNode.h
TobyAsE: Using Kling's `VERIFY(search_child.parent() == this)` from above would mitigate this as we check whether we are the parent of the given node and therefore we have at least one child.
But in case something fishy is going on, we can still check for child being null in the while ;)

Lubrsi: If we're asserting that, we could also assert that child isn't null, since like you said that assertion means we have at least one child.

TobyAsE: Thats probably a good idea too!

Userland/Libraries/LibWeb/TreeNode.h
```diff
@@ -97,6 +97,39 @@ class TreeNode : public Weakable<T> {
         return const_cast<TreeNode*>(this)->child_at_index(index);
     }
 
+    Optional<size_t> index_of_child(const T& search_child)
+    {
+        VERIFY(search_child.parent() == this);
+        int index = 0;
```
awesomekling, line 103: s/int/size_t/

Userland/Libraries/LibWeb/TreeNode.h
```diff
@@ -97,6 +97,39 @@ class TreeNode : public Weakable<T> {
...
+            index++;
+        } while (child && (child = child->next_sibling()));
+        return {};
+    }
+
+    template<typename ChildType>
+    Optional<size_t> index_of_child(const T& search_child)
+    {
+        VERIFY(search_child.parent() == this);
+        int index = 0;
```
awesomekling, line 119: s/int/size_t/

TobyAsE:
@awesomekling While we are at it, let me also bump this :^)

PR #6412, opened 2021-04-17
Ports: Add imgcat port

panos:
- [imgcat's GitHub repository](https://github.com/eddieantonio/imgcat)

PR #6435, opened 2021-04-18
Kernel: Added sysexits.h

panos:
Created for compatibility purposes with programs primarily designed for BSD-like systems.

For example, **imgcat** (see: #6412), which currently doesn't work on the system, uses these exit codes.

### DIFF
--- a/Userland/Libraries/LibC/sysexits.h
+++ b/Userland/Libraries/LibC/sysexits.h
@@ -1,0 +1,49 @@
+/*
+ * Copyright (c) 2021, Panagiotis Vasilopoulos <hello@alwayslivid.com>
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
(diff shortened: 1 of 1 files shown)


panos:
Gah, my text editor keeps fooling me.

panos:
`clang-format` seems to think it's okay. I also added my own copyright notice, as the CONTRIBUTING.md file recommended.

Based on my experience with porting software for the Haiku operating system, porting software to Serenity will eventually require adding some BSD compatibility stuff, bigger or smaller, sooner or later, alongside with potentially pushing patches upstream. I'll reiterate that the header file was modeled after documentation in the FreeBSD man pages.

awesomekling (changes requested):
If you want this to be accessible to userspace, it needs to be in `Userland/Libraries/LibC/`, not `Kernel`.

Also headers should have `#pragma once`

panos:
> If you want this to be accessible to userspace, it needs to be in `Userland/Libraries/LibC/`, not `Kernel`.
> 
> Also headers should have `#pragma once`

Understood, expected that to be a bit incorrect, thanks for the guidance!

panos:
> Also headers should have `#pragma once`

I'm a bit way too accustomed to C and I just assumed that all I had to do was do away with the `#ifndef` conditions, still learning!

Userland/Libraries/LibC/sysexits.h
```diff
@@ -0,0 +1,50 @@
...
+#define EX_OSERR 71
+#define EX_OSFILE 72
+#define EX_CANTCREAT 73
+#define EX_IOER 74
+#define EX_TEMPFAIL 75
+#define EX_PROTOCOL 76
+#define EX_NOPERM 77
+#define EX_CONFIG 78
+#define EX__MAX 78
+
```
emanuele6, line 50:
```
Running Meta/check-newlines-at-eof.py... 
Files that have blank lines at the end: Userland/Libraries/LibC/sysexits.h
[FAIL]: Meta/check-newlines-at-eof.py
```

emanuele6 (changes requested):

emanuele6:
This is a LibC change; why is your commit prefixed with `Kernel: `?

panos:
> oh, you must have forgot to change it.

Yeah... I'm juggling around with a lot of pull requests, I'll get right on it.

PR #6436, opened 2021-04-18
Revert "Ports: Add imgcat port"

panos:
There's tons of issues with the `imgcat` port of mine.

From a mismatching md5 checksum, all the way up to source code-related incompatibilities with the system and a custom, forsaken `configure` file. It'd be best to remove it for the time being and work on it again in a pull request with a procedure, rather than sending tons of pull requests meant to patch things up.

- Original PR: #6436

### DIFF
--- a/Ports/AvailablePorts.md
+++ b/Ports/AvailablePorts.md
@@ -2,6 +2,7 @@
 
 Please make sure to keep this list up to date when adding and updating ports. :^)
 
+<<<<<<< HEAD
 | Port                                   | Name                                                       | Version                  | Website                                                                        |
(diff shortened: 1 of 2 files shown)


gunnarbeutner:
Personally I'd keep the port and just have someone (not necessarily you) fix it at a later point.

panos:
> Personally I'd keep the port and just have someone (not necessarily you) fix it at a later point.

I've already issued a couple of pull requests revolving around that port and I do have a new, improved version that should work *better* (although not work). Not sure.

linusg:
yeah, instead of removing the port and then adding it *again* later, let's just add required missing functionality and fix it up incrementally.

PR #6437, opened 2021-04-18
LibGUI: Implement Vim motion system

sin-ack:
This patch implements Vim motions.  The `VimMotion` class will accept
keycodes from the editing engine to build up a motion, and will
signal when a motion is complete via `VimMotion::is_complete()`.  The
editing engine can then call `VimMotion::get_range()` to obtain a
`TextRange` object which can be used to perform operations on the text,
or `VimMotion::get_position()` to obtain a `TextPosition` which is the
new position of the cursor after the motion.

Currently, the following motions are supported:

- h/j/k/l, regular Vim line and character movements
- 0/^/$, start/end of line and start of non-blank
- w/e/b/ge, word-related movements
- W/E/B/gE, `WORD` (anything non-blank) versions of the above motions
- gg/G, document related movements
- t/f, to/find character

All motions except gg/G accept a number prefix to repeat the motion that
many times.

This patch updates insert, normal and visual modes to use this motion
system for movement.

Current status:
- [X] motions with repetition
- [X] h/l motion (next/previous character)
- [X] j/k motion (next/previous line)
- [x] Arrow key variants of hjkl
- [X] w/W/b/B motion (next/previous word/WORD)
- [x] e/E motion (end of next word/WORD)
- [x] ge/gE motion (end of previous word/WORD) (does anyone use this?)
- [x] gg/G motion (start/end of document)
- [X] 0 motion (start of line)
- [X] $ motion (end of line)
- [x] ^ motion (first non-whitespace character)
- [ ] ~~[]/{}/() motions (probably won't, because I don't use/understand them)~~
- [x] t prefix (to character)
- [x] f prefix (find character)
- [X] Deletion uses motions
- [X] Yank uses motions
- [x] Change uses motions
- [x] Visual selection uses motions
- [x] Movement uses motions :tada: 

Things to fix before un-drafting:
- [x] Remove debug prints
- [x] Fix `j`/`k` motions not correctly operating with commands
- [x] Fix crash with editor when Vim Emulation mode is exited.

### DIFF
--- a/Userland/Libraries/LibGUI/EditingEngine.h
+++ b/Userland/Libraries/LibGUI/EditingEngine.h
@@ -29,6 +29,12 @@ public:
     void attach(TextEditor& editor);
     void detach();
 
+    TextEditor& editor()
+    {
+        VERIFY(!m_editor.is_null());
+        return *m_editor.unsafe_ptr();
+    }
+
     virtual bool on_key(const KeyEvent& event);
 
 protected:
--- a/Userland/Libraries/LibGUI/VimEditingEngine.cpp
+++ b/Userland/Libraries/LibGUI/VimEditingEngine.cpp
@@ -1,14 +1,730 @@
 /*
- * Copyright (c) 2020, the SerenityOS developers.
+ * Copyright (c) 2020-2021, the SerenityOS developers.
  *
  * SPDX-License-Identifier: BSD-2-Clause
  */
 
+#include <AK/Assertions.h>
+#include <AK/String.h>
 #include <LibGUI/Event.h>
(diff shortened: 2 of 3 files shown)


Userland/Libraries/LibGUI/EditingEngine.h
```diff
@@ -29,6 +29,8 @@ class EditingEngine {
     void attach(TextEditor& editor);
     void detach();
 
+    TextEditor* editor() { return m_editor.unsafe_ptr(); }
```
IdanHo, line 32: you VERIFY this is non-null in all the places it is used, so making it return a reference will probably be cleaner.

PR #6441, opened 2021-04-18
LibPthread: Add stub for pthread_cancel

gunnarbeutner:


linusg:
Would be nice to know what you're adding this for (a port, presumably?) so we know upfront that it'll crash when calling the function.

gunnarbeutner:
You're right, this doesn't really make much sense on its own. It's part of my effort to implement C++ thread support. I'll just put that in there. :+1:

PR #6442, opened 2021-04-18
Kernel: Make VirtIOConsole block when VirtIOQueue is full

ccapitalK:
Currently, filling the backing VirtIOQueue will cause an assertion failure. This makes it block instead.


IdanHo (approved):
LGTM

PR #6443, opened 2021-04-18
LibTLS+LibCrypto: Use the new ASN.1 parser to parse certificates / LibCore: Tiny Core::DateTime API modification

alimpfard:
Did someone say "double the code and multiply the memory bugs by \<n\> where n < 0.5"?
the "Quick and dirty" certificate parser is no more!

### DIFF
--- a/Userland/Libraries/LibCrypto/ASN1/ASN1.cpp
+++ b/Userland/Libraries/LibCrypto/ASN1/ASN1.cpp
@@ -92,4 +92,118 @@ String type_name(Type type)
     return "InvalidType";
 }
 
+Optional<Core::DateTime> parse_utc_time(const StringView& time)
+{
+    // YYMMDDhhmm[ss]Z or YYMMDDhhmm[ss](+|-)hhmm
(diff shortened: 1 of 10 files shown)


Userland/Libraries/LibCrypto/ASN1/ASN1.cpp
```diff
@@ -92,4 +92,118 @@ String type_name(Type type)
     return "InvalidType";
 }
 
+Optional<Core::DateTime> parse_utc_time(const StringView& time)
```
awesomekling, line 95: This doesn't look very fault tolerant.

awesomekling: Or wait, maybe it does. Never mind :D

alimpfard: (it was originally returning a `Result<DateTime, UTCTimeParseFailure>` but honestly...who wants to if the minute is missing, or the hours are missing from the site that gave them a broken certificate?)

PR #6444, opened 2021-04-18
Ports: Fix up imgcat

panos:
This PR is a continuation of: #6412

This fixes the md5 checksum and introduces a few blindly applied changes from other ports using the `ncurses` package. Some further changes in the source code of Serenity or imgcat itself will be required later on, particularly due to missing headers and other technicalities.

Important thing to note: imgcat's custom configure file is not detecting Serenity's `ncurses` and cannot link the library.

Relevant PRs:
- #6435
- #6436


Ports/imgcat/package.sh
```diff
@@ -1,8 +1,14 @@
...
-files="https://github.com/eddieantonio/imgcat/releases/download/v${version}/imgcat-${version}.tar.gz imgcat-v${version}.tar.gz 1b5d45ccafc6fbb7a7ee685d4a5110101418d48b"
+depends="ncurses"
+files="https://github.com/eddieantonio/imgcat/releases/download/v${version}/imgcat-${version}.tar.gz imgcat-v${version}.tar.gz 16e5386f5ce2237643785e7c8bb2cfce"
 auth_type=md5
 
-depends="ncurses"
+build() {
+    run make \
+        production=true \
+        CURSES_CFLAGS="-I${SERENITY_ROOT}/Build/i686/Root/usr/local/include/ncurses"
```
linusg, line 11: Please don't hardcode `Build/i686`, use `SERENITY_BUILD_DIR` instead. Also, why is both `CURSES_CFLAGS` and `CPPFLAGS` needed?

linusg (changes requested):

Ports/imgcat/package.sh
panos: I tried to splice parts from different ports as I was trying to debug the problem properly. I'll remove the former and remove the hardcoded thing, and edit it out of other places in the ports as well.

Issue #6188, opened 2021-04-08
Remove need for IterationDecision::Continue


keith-rollin:
Andreas, I was watching an old video of yours, where you were adding for_each_widget, for_each_widget_of_type<>, etc. In the video, you bemoaned having to turn your nice one-line native loops into for_each calls with two-lined bodies, since you needed to now return IterationDecision::Continue. But since you use a modern compiler, you can use "if constexpr" to get rid of the need for returning an IterationDecision if it can be assumed to always return IterationDecision::Continue (which is likely the common case, so we optimize for that).

Here's an example patch for for_each_in_pgrp() in Kernel/Process.h. There are four places where for_each_in_pgrp() is called: two returning Continue and two returning Break. Those first two can be changed to return nothing, as shown below.
```
diff --git a/Kernel/Process.h b/Kernel/Process.h
index 7848baa31..4edaf32e7 100644
--- a/Kernel/Process.h
+++ b/Kernel/Process.h
@@ -51,6 +51,7 @@
 #include <Kernel/VM/Space.h>
 #include <LibC/signal_numbers.h>
 #include <LibELF/exec_elf.h>
+#include <type_traits>
 
 namespace Kernel {
 
@@ -693,8 +694,12 @@ inline void Process::for_each_in_pgrp(ProcessGroupID pgid, Callback callback)
     for (auto* process = g_processes->head(); process;) {
         auto* next_process = process->next();
         if (!process->is_dead() && process->pgid() == pgid) {
-            if (callback(*process) == IterationDecision::Break)
-                break;
+            if constexpr (std::is_void_v<std::invoke_result_t<Callback, decltype(*process)>>) {
+                callback(*process);
+            } else {
+                if (callback(*process) == IterationDecision::Break)
+                    break;
+            }
         }
         process = next_process;
     }
diff --git a/Kernel/Syscalls/kill.cpp b/Kernel/Syscalls/kill.cpp
index 48eb07fd2..1a9c86fc8 100644
--- a/Kernel/Syscalls/kill.cpp
+++ b/Kernel/Syscalls/kill.cpp
@@ -67,8 +67,6 @@ KResult Process::do_killpg(ProcessGroupID pgrp, int signal)
             any_succeeded = true;
         else
             error = res;
-
-        return IterationDecision::Continue;
     });
 
     if (group_was_empty)
diff --git a/Kernel/TTY/TTY.cpp b/Kernel/TTY/TTY.cpp
[...]

PR #6445, opened 2021-04-18
AK+Kernel+LibELF: Remove the need for `IteratorDecision::Continue`

Nicholas-Baron:
Work towards #6188 

The discussion has produced a compiling and running solution.
The only remaining issue is to generalize it for the whole codebase.

Following in the footsteps of `AK_MAKE_NON{COPY,MOVE}ABLE`, the
`DO_CALLBACK` macro will generalize the `void`/`IterationDecision`
compile-time switch.

I am still not certain about the following:
1. Where should this macro be defined?
2. Is there the possibility or use case for a callback with 2 or more arguments? (There is at least one)
3. What about if the `for_each` function needs to return its own `IterationDecision`?
4. Could this somehow be a template, instead of a macro?

What is shown is a small sample of changing one function and one call site. 
The final commit will `hopefully` catch most of the use cases for this PR.

### DIFF
--- a/AK/InlineLinkedList.h
+++ b/AK/InlineLinkedList.h
@@ -7,6 +7,7 @@
 #pragma once
 
 #include <AK/Assertions.h>
+#include <AK/Concepts.h>
 #include <AK/Types.h>
 
 namespace AK {
@@ -112,7 +113,7 @@ public:
         return false;
     }
 
-    template<typename F>
+    template<IteratorFunction<T&> F>
     IterationDecision for_each(F func) const
     {
         for (T* node = m_head; node; node = node->next()) {
@@ -121,6 +122,13 @@ public:
                 return decision;
         }
         return IterationDecision::Continue;
+    }
+
+    template<VoidFunction<T&> F>
+    void for_each(F func) const
+    {
+        for (T* node = m_head; node; node = node->next())
+            func(*node);
     }
 
     using Iterator = InlineLinkedListIterator<T>;
--- a/Kernel/Arch/x86/CPU.h
+++ b/Kernel/Arch/x86/CPU.h
@@ -8,6 +8,7 @@
 
 #include <AK/Atomic.h>
 #include <AK/Badge.h>
+#include <AK/Concepts.h>
 #include <AK/Noncopyable.h>
 #include <AK/Vector.h>
 
@@ -733,7 +734,7 @@ public:
 
     static size_t processor_count() { return processors().size(); }
 
-    template<typename Callback>
+    template<IteratorFunction<Processor&> Callback>
     static inline IterationDecision for_each(Callback callback)
     {
         auto& procs = processors();
@@ -742,6 +743,16 @@ public:
             if (callback(*procs[i]) == IterationDecision::Break)
                 return IterationDecision::Break;
         }
+        return IterationDecision::Continue;
+    }
+
+    template<VoidFunction<Processor&> Callback>
+    static inline IterationDecision for_each(Callback callback)
+    {
+        auto& procs = processors();
+        size_t count = procs.size();
+        for (size_t i = 0; i < count; i++)
+            callback(*procs[i]);
         return IterationDecision::Continue;
     }
 
--- a/Kernel/Process.h
+++ b/Kernel/Process.h
@@ -7,6 +7,7 @@
 #pragma once
 
 #include <AK/Checked.h>
+#include <AK/Concepts.h>
 #include <AK/HashMap.h>
 #include <AK/InlineLinkedList.h>
 #include <AK/NonnullOwnPtrVector.h>
@@ -224,16 +225,30 @@ public:
     RefPtr<FileDescription> file_description(int fd) const;
     int fd_flags(int fd) const;
 
-    template<typename Callback>
+    // Breakable iteration functions
+    template<IteratorFunction<Process&> Callback>
     static void for_each(Callback);
-    template<typename Callback>
+    template<IteratorFunction<Process&> Callback>
     static void for_each_in_pgrp(ProcessGroupID, Callback);
-    template<typename Callback>
+    template<IteratorFunction<Process&> Callback>
     void for_each_child(Callback);
 
-    template<typename Callback>
+    template<IteratorFunction<Thread&> Callback>
     IterationDecision for_each_thread(Callback);
-    template<typename Callback>
+    template<IteratorFunction<Thread&> Callback>
+    IterationDecision for_each_thread(Callback callback) const;
+
+    // Non-breakable iteration functions
+    template<VoidFunction<Process&> Callback>
+    static void for_each(Callback);
+    template<VoidFunction<Process&> Callback>
+    static void for_each_in_pgrp(ProcessGroupID, Callback);
+    template<VoidFunction<Process&> Callback>
+    void for_each_child(Callback);
+
+    template<VoidFunction<Thread&> Callback>
+    IterationDecision for_each_thread(Callback);
+    template<VoidFunction<Thread&> Callback>
     IterationDecision for_each_thread(Callback callback) const;
 
     void die();
@@ -628,7 +643,7 @@ private:
 extern InlineLinkedList<Process>* g_processes;
 extern RecursiveSpinLock g_processes_lock;
 
-template<typename Callback>
+template<IteratorFunction<Process&> Callback>
 inline void Process::for_each(Callback callback)
 {
     VERIFY_INTERRUPTS_DISABLED();
@@ -641,7 +656,7 @@ inline void Process::for_each(Callback callback)
     }
 }
 
-template<typename Callback>
+template<IteratorFunction<Process&> Callback>
 inline void Process::for_each_child(Callback callback)
 {
     VERIFY_INTERRUPTS_DISABLED();
@@ -657,7 +672,7 @@ inline void Process::for_each_child(Callback callback)
     }
 }
 
-template<typename Callback>
+template<IteratorFunction<Thread&> Callback>
 inline IterationDecision Process::for_each_thread(Callback callback) const
 {
     ScopedSpinLock thread_list_lock(m_thread_list_lock);
@@ -669,7 +684,7 @@ inline IterationDecision Process::for_each_thread(Callback callback) const
     return IterationDecision::Continue;
 }
 
-template<typename Callback>
+template<IteratorFunction<Thread&> Callback>
 inline IterationDecision Process::for_each_thread(Callback callback)
 {
     ScopedSpinLock thread_list_lock(m_thread_list_lock);
@@ -681,7 +696,7 @@ inline IterationDecision Process::for_each_thread(Callback callback)
     return IterationDecision::Continue;
 }
 
-template<typename Callback>
+template<IteratorFunction<Process&> Callback>
 inline void Process::for_each_in_pgrp(ProcessGroupID pgid, Callback callback)
 {
     VERIFY_INTERRUPTS_DISABLED();
@@ -694,6 +709,51 @@ inline void Process::for_each_in_pgrp(ProcessGroupID pgid, Callback callback)
         }
         process = next_process;
     }
+}
+
+template<VoidFunction<Process&> Callback>
+inline void Process::for_each(Callback callback)
+{
+    return for_each([&](auto& item) {
+        callback(item);
+        return IterationDecision::Continue;
+    });
+}
+
+template<VoidFunction<Process&> Callback>
+inline void Process::for_each_child(Callback callback)
+{
+    return for_each_child([&](auto& item) {
+        callback(item);
+        return IterationDecision::Continue;
+    });
+}
+
+template<VoidFunction<Thread&> Callback>
+inline IterationDecision Process::for_each_thread(Callback callback) const
+{
+    ScopedSpinLock thread_list_lock(m_thread_list_lock);
+    for (auto& thread : m_thread_list)
+        callback(thread);
+    return IterationDecision::Continue;
+}
+
+template<VoidFunction<Thread&> Callback>
+inline IterationDecision Process::for_each_thread(Callback callback)
+{
+    ScopedSpinLock thread_list_lock(m_thread_list_lock);
+    for (auto& thread : m_thread_list)
+        callback(thread);
+    return IterationDecision::Continue;
+}
+
+template<VoidFunction<Process&> Callback>
+inline void Process::for_each_in_pgrp(ProcessGroupID pgid, Callback callback)
+{
+    return for_each_in_pgrp(pgid, [&](auto& item) {
+        callback(item);
+        return IterationDecision::Continue;
+    });
 }
 
 inline bool InodeMetadata::may_read(const Process& process) const
--- a/Kernel/Thread.cpp
+++ b/Kernel/Thread.cpp
@@ -404,7 +404,6 @@ void Thread::finalize_dying_threads()
         for_each_in_state(Thread::State::Dying, [&](Thread& thread) {
             if (thread.is_finalizable())
                 dying_threads.append(&thread);
-            return IterationDecision::Continue;
         });
     }
     for (auto* thread : dying_threads) {
@@ -740,7 +739,6 @@ DispatchSignalResult Thread::dispatch_signal(u8 signal)
             process.set_dump_core(true);
             process.for_each_thread([](auto& thread) {
                 thread.set_dump_backtrace_on_finalization();
-                return IterationDecision::Continue;
             });
             [[fallthrough]];
         case DefaultSignalAction::Terminate:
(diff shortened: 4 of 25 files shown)


Nicholas-Baron:
For item 2, there is a case in the current codebase of a 2 parameter callback: [`for_each_frame` in Profiler/Profile.cpp].(https://github.com/SerenityOS/serenity/blob/8d4780aff4d9d159c48d2c414337f3996b141dd1/Userland/DevTools/Profiler/Profile.cpp#L126-L138)

AK/StdLibExtras.h
```diff
@@ -134,6 +134,18 @@ using RawPtr = typename Detail::_RawPtr<T>::Type;
 
 }
 
+// TODO: Make this work for more than 1 parameter.
+// TODO: Make this not a macro?
```
alimpfard, line 138: Could try
```c++
template<typename Func, typename... Args>
static inline bool do_callback(Func&& func, Args&&... args) {
    if constexpr (IsVoid<decltype(func (args...))>) {
        func (forward<Args>(args)...);
        return false;
    } else {
        return func(forward<Args>(args)...) == IterationDecision::Break;
    }
}
```

and using it like `if (do_callback(func, args...)) return;`

(Note: on phone, expect typos and syntax errors :P)

Userland/Services/WindowServer/Menubar.h
```diff
@@ -49,8 +50,7 @@ class Menubar
     void for_each_menu(Callback callback)
     {
         for (auto& menu : m_menus) {
-            if (callback(*menu) == IterationDecision::Break)
-                return;
+            DO_CALLBACK(callback, *menu);
```
Hendiadyoin1, line 53: the way it is now you don't break out; try
```cpp
if (do_callback(callback, *menu))
    break;
```
instead (assuming the implementation from @alimpfard)

Hendiadyoin1:
Other iterators return an IterationDecission, you may want to specialize these for void functions then.
For that just add a conctept
```cpp
template<typename func>
concept VoidFunction = IsFunction<func> && IsVoid<func>;
```
if I'm not mistaken
and the reimplement these iterators without the check, but use the concept as the descriptor of the first type in the template like:
```cpp
teplate<VoidFunction Callback, typename... Args>
void for_each(){
   ...
}
``

File: AK/StdLibExtras.h (hunk 1 above)
Comment (Nicholas-Baron) [reply AK/StdLibExtras.h]: Thank you, I changed the implementation and documented the meaning of the return value.

File: Userland/Services/WindowServer/Menubar.h (hunk 2 above)
Comment (Nicholas-Baron) [reply Userland/Services/WindowServer/Menubar.h]: Thank you, that is what the implementation looks like now.

**Comment by Nicholas-Baron** [comment]
@Hendiadyoin1 Concepts seem to be locked [behind some macros](https://github.com/SerenityOS/serenity/blob/4f9ab0b9820c3276d85e66720eaeccb675989269/AK/Concepts.h), so I do not feel comfortable spreading concepts everywhere in the codebase.

I think the implementation would look something more like
```cpp
template<typename Func, typename... Args>
concept IsVoidFunction = IsFunction<Func> && IsVoid<decltype(Func()(Args...))>;
```
Still not so sure about it; will become certain if the concepts route opens up.

Nicholas-Baron:
Of the four original questions, here are the ones that have been answered:

Question 1 has not been discussed and so I will not write an answer for it.

2. Is there the possibility or use case for a callback with 2 or more arguments? 
There is [at least one case](https://github.com/SerenityOS/serenity/blob/8d4780aff4d9d159c48d2c414337f3996b141dd1/Userland/DevTools/Profiler/Profile.cpp#L126-L138) of such a callback. However, this is solved by the current implementation.

3. What about if the for_each function needs to return its own IterationDecision?
Currently, the `do_callback` function returns `true` if it needs to break. 
This could be changed to a `IterationDecision`, but that change would add more boilerplate.

4. Could this somehow be a template, instead of a macro?
Yes, the current implementation switched to a template version.

Hendiadyoin1:
> @Hendiadyoin1 Concepts seem to be locked [behind some macros](https://github.com/SerenityOS/serenity/blob/4f9ab0b9820c3276d85e66720eaeccb675989269/AK/Concepts.h),

If you mean the `#if defined(__cpp_concepts) && !defined(__COVERITY__)` this is to prevent any weird errors when trying to build without concepts enabled, Other parts of AK already use Concepts, and this might affect Question 3 as described before

Nicholas-Baron:
> if I'm not mistaken
> and then reimplement these iterators without the check, but use the concept as the descriptor of the first type in the template like:
> 
> ```c++
> template<VoidFunction Callback, typename... Args>
> void for_each(Callback callback){
>    ...
> }
> ```

I do not know how much better this specializing approach would be, compared to having just one implementation.

I will probably add a `CallbackFunction` concept to help with error messages, but duplicating the implementation seems to be more work for the same result. 
The only advantage I can see is a simpler implementation, one with a `NonBreakingCallback` concept and the other with a `BreakableCallback` concept.

Hendiadyoin1:
>The only advantage I can see is a simpler implementation, one with a NonBreakingCallback concept and the other with a BreakableCallback concept.
That's the point, the NonBreakable version would fall through all iterators and has no checks, while the breakable one has a check at each stage

Nicholas-Baron:
> > The only advantage I can see is a simpler implementation, one with a NonBreakingCallback concept and the other with a BreakableCallback concept.
> 
> That's the point, the NonBreakable version would fall through all iterators and has no checks, while the breakable one has a check at each stage

Time to rewrite this entire patch. 
This would eliminate the need for a `do_callback` function, as the `VoidFunction` could never signal a break.

Nicholas-Baron:
@Hendiadyoin1 Took your suggestion and implemented it for `InlineLinkedList`. 
Now, comes time to redo all of the old work. I threw it out as I mixed changing the `for_each` functions and removing `IterationDecision`.

Nicholas-Baron:
I am thinking about splitting this branch in two. One would do AK and kernel and the other would do userland.

keith-rollin:
Sorry for the late reply. I didn't see this work until @Nicholas-Baron added his comment to #6188.

It looks like this approach requires duplicating all of the various for_each functions so that there's an IteratorFunction version and a VoidFunction version. I like the "if constexpr" approach in #6188 so that code duplication is minimal. Is there a reason why we can't or shouldn't take that approach? If the approach in this patch becomes the template for all for_each's in the rest of Serenity, it would be nice if it didn't end up duplicating all those for_each's.

Nicholas-Baron:
@keith-rollin The issue with the `if constexpr` approach was the `do_callback` function, which has the interesting semantics of returning `true` when the callback _should_ break. To me, this was a strange semantic property when shown only the definition and no examples.

The other way I considered doing the `for_each` functions was to have the void version just call the non-void one, something like:
```c++
template<VoidFunction<T&> Func>
auto for_each(Func callback)
{
  return for_each([&](auto & item){
    callback(item);
  });
}
``` 
This could be better middle ground with less duplication and still avoiding the unintuitive `do_callback` function.

Note: I agree that some uses may do better with the `if constexpr` approach. However, generalizing the pattern means a one-size-fits-all solution, which can be trickier for some of the more complex iterator functions (see [`AnonymousVMObject::for_each_volatile_range`](https://github.com/SerenityOS/serenity/blob/6dafe5468e1ab45ebe4e161c2456f5911822fe41/Kernel/VM/AnonymousVMObject.h#L64-L99)).

AK/InlineLinkedList.h
```diff
@@ -143,6 +144,14 @@ class InlineLinkedList {
         return IterationDecision::Continue;
     }
 
+    template<VoidFunction<T&> F>
+    void for_each(F func) const
+    {
+        for (T* node = m_head; node; node = node->next()) {
+            func(*node);
+        }
```
linusg, line 152:
```suggestion
        for (T* node = m_head; node; node = node->next())
            func(*node);
```

Kernel/Arch/x86/CPU.h
```diff
@@ -765,6 +766,17 @@ class Processor {
...
     }
 
+    template<VoidFunction<Processor&> Callback>
+    static inline IterationDecision for_each(Callback callback)
+    {
+        auto& procs = processors();
+        size_t count = procs.size();
+        for (size_t i = 0; i < count; i++) {
+            callback(*procs[i]);
+        }
```
linusg, line 776:
```suggestion
        for (size_t i = 0; i < count; i++)
            callback(*procs[i]);
```

Kernel/Process.h
```diff
@@ -701,6 +716,67 @@ inline void Process::for_each_in_pgrp(ProcessGroupID pgid, Callback callback)
...
+inline void Process::for_each_child(Callback callback)
+{
+    VERIFY_INTERRUPTS_DISABLED();
+    ProcessID my_pid = pid();
+    ScopedSpinLock lock(g_processes_lock);
+    for (auto* process = g_processes->head(); process;) {
+        auto* next_process = process->next();
+        if (process->ppid() == my_pid || process->has_tracee_thread(pid())) {
+            callback(*process);
+        }
```
linusg, line 741:
```suggestion
        if (process->ppid() == my_pid || process->has_tracee_thread(pid()))
            callback(*process);
```

Kernel/Process.h
```diff
@@ -701,6 +716,67 @@ inline void Process::for_each_in_pgrp(ProcessGroupID pgid, Callback callback)
...
+    }
+}
+
+template<VoidFunction<Thread&> Callback>
+inline IterationDecision Process::for_each_thread(Callback callback) const
+{
+    ScopedSpinLock thread_list_lock(m_thread_list_lock);
+    for (auto& thread : m_thread_list) {
+        callback(thread);
+    }
```
linusg, line 752:
```suggestion
    for (auto& thread : m_thread_list)
        callback(thread);
```

Kernel/Process.h
```diff
@@ -701,6 +716,67 @@ inline void Process::for_each_in_pgrp(ProcessGroupID pgid, Callback callback)
...
+    return IterationDecision::Continue;
+}
+
+template<VoidFunction<Thread&> Callback>
+inline IterationDecision Process::for_each_thread(Callback callback)
+{
+    ScopedSpinLock thread_list_lock(m_thread_list_lock);
+    for (auto& thread : m_thread_list) {
+        callback(thread);
+    }
```
linusg, line 762:
```suggestion
    for (auto& thread : m_thread_list)
        callback(thread);
```

Kernel/Process.h
```diff
@@ -701,6 +716,67 @@ inline void Process::for_each_in_pgrp(ProcessGroupID pgid, Callback callback)
...
+template<VoidFunction<Process&> Callback>
+inline void Process::for_each_in_pgrp(ProcessGroupID pgid, Callback callback)
+{
+    VERIFY_INTERRUPTS_DISABLED();
+    ScopedSpinLock lock(g_processes_lock);
+    for (auto* process = g_processes->head(); process;) {
+        auto* next_process = process->next();
+        if (!process->is_dead() && process->pgid() == pgid) {
+            callback(*process);
+        }
```
linusg, line 775:
```suggestion
        if (!process->is_dead() && process->pgid() == pgid)
            callback(*process);
```

Kernel/Thread.cpp
```diff
@@ -927,11 +926,12 @@ void Thread::set_state(State new_state, u8 stop_signal)
         auto& process = this->process();
         if (process.set_stopped(true) == false) {
             process.for_each_thread([&](auto& thread) {
-                if (&thread == this || thread.is_stopped())
-                    return IterationDecision::Continue;
+                if (&thread == this)
+                    return;
+                if (!thread.is_stopped())
+                    return;
```
linusg, line 932: This inverts the `thread.is_stopped()` check!

Kernel/Thread.h
```diff
@@ -1305,6 +1311,29 @@ inline IterationDecision Thread::for_each_in_state(State state, Callback callbac
     return IterationDecision::Continue;
 }
 
+template<VoidFunction<Thread&> Callback>
+inline IterationDecision Thread::for_each(Callback callback)
+{
+    ScopedSpinLock lock(g_tid_map_lock);
+    for (auto& it : *g_tid_map) {
+        callback(*it.value);
+    }
```
linusg, line 1320:
```suggestion
    for (auto& it : *g_tid_map)
        callback(*it.value);
```

Kernel/VM/MemoryManager.h
```diff
@@ -187,6 +188,14 @@ class MemoryManager {
         }
     }
 
+    template<VoidFunction<VMObject&> Callback>
+    static void for_each_vmobject(Callback callback)
+    {
+        for (auto& vmobject : MM.m_vmobjects) {
+            callback(vmobject);
+        }
```
linusg, line 196:
```suggestion
        for (auto& vmobject : MM.m_vmobjects)
            callback(vmobject);
```

Userland/Libraries/LibELF/Image.h
```diff
@@ -242,15 +255,24 @@ class Image {
...
 };
 
-template<typename F>
+template<IteratorFunction<Image::Section&> F>
+inline void Image::for_each_section(F func) const
+{
+    auto section_count = this->section_count();
+    for (unsigned i = 0; i < section_count; ++i)
+        if (func(section(i)) == IterationDecision::Break)
+            break;
```
linusg, line 264:
```suggestion
    for (unsigned i = 0; i < section_count; ++i) {
        if (func(section(i)) == IterationDecision::Break)
            break;
    }
```

Userland/Libraries/LibELF/Image.h
```diff
@@ -263,17 +285,28 @@ inline void Image::for_each_section_of_type(unsigned type, F func) const
...
-template<typename F>
+template<VoidFunction<Image::Section&> F>
+inline void Image::for_each_section_of_type(unsigned type, F func) const
+{
+    auto section_count = this->section_count();
+    for (unsigned i = 0; i < section_count; ++i) {
+        auto section = this->section(i);
+        if (section.type() == type) {
+            func(section);
+        }
```
linusg, line 296:
```suggestion
        if (section.type() == type)
            func(section);
```

Userland/Libraries/LibELF/Image.h
```diff
@@ -283,13 +316,31 @@ inline void Image::for_each_symbol(F func) const
...
 }
 
-template<typename F>
+template<VoidFunction<Image::Symbol&> F>
+inline void Image::for_each_symbol(F func) const
+{
+    auto symbol_count = this->symbol_count();
+    for (unsigned i = 0; i < symbol_count; ++i) {
+        func(symbol(i));
+    }
```
linusg, line 325:
```suggestion
    for (unsigned i = 0; i < symbol_count; ++i)
        func(symbol(i));
```

Userland/Libraries/LibELF/Image.h
```diff
@@ -283,13 +316,31 @@ inline void Image::for_each_symbol(F func) const
...
+}
+
+template<VoidFunction<Image::ProgramHeader&> F>
+inline void Image::for_each_program_header(F func) const
+{
+    auto program_header_count = this->program_header_count();
+    for (unsigned i = 0; i < program_header_count; ++i) {
+        func(program_header(i));
     }
```
linusg, line 344:
```suggestion
    for (unsigned i = 0; i < program_header_count; ++i)
        func(program_header(i));
```

linusg (changes requested):

Kernel/Thread.cpp
Nicholas-Baron: Thank you for pointing this out. Fixed in the latest commit.

Nicholas-Baron:
@linusg Thank you for the review. All suggestions applied.

linusg (changes requested):
Please squash the changes from code review into the appropriate previous commits, instead of just adding another fixup commit - see the [contribution guidelines](https://github.com/SerenityOS/serenity/blob/master/CONTRIBUTING.md#code-submission-policy).

Nicholas-Baron:
@linusg It lapsed my mind. All squashed into one commit.

linusg:
please reword the commit message so all lines are at most 72 characters long - it won't pass CI otherwise.

This line is currently 3 chars too long:

```
passing the types as part of the constraint (eg: `IterationFunction<int>`).
```

Nicholas-Baron:
@linusg Commit message cleaned up.

linusg:
Sorry for the delay, this has a conflict now..

Nicholas-Baron:
@linusg Merge conflict resolved.

Nicholas-Baron:
@linusg Are there any issues that are preventing this pull request from being merged?

Userland/Libraries/LibELF/Image.h
```diff
@@ -222,15 +235,25 @@ class Image {
...
 
-template<typename F>
+template<IteratorFunction<Image::Section&> F>
+inline void Image::for_each_section(F func) const
+{
+    auto section_count = this->section_count();
+    for (unsigned i = 0; i < section_count; ++i) {
+        if (func(section(i)) == IterationDecision::Break)
+            break;
+    }
```
linusg, line 245: Same as above, same for the below

linusg (commented):
Sorry for the delay!

This looks nice, however there's too much code duplication. In some cases you use a nice pattern where the void function variant simply calls the "real" function with a lambda that calls the callback and returns `IterationDecision::Continue`, which looks like a good solution - any reason why we can't use that in more / all places?

Nicholas-Baron:
@linusg I do not remember the initial reason I duplicated whole functions. Must have just been laziness at the time.

PR #6447, opened 2021-04-18
LibC: Initial work for interface header definitions

panos:
Here are some incomplete header definitions that do not interface with the drivers as of yet. This is meant to enhance Serenity's POSIX compatibility and it's just some initial work being done. I began working on this as a result of reviewing whether Tor, a daemon that interfaces with the network interface on a low level, could work under Serenity.

I stopped because the definitions started differentiating between different projects and I haven't managed to find any documents/resources to use as proper pointers (other than "what did other people do?"), and decided to upstream that tiny bit of my work until I look into this a bit further, because following other peoples' work blindly when doing so could result in frustrating compatibility-related issues later on is not a good idea.

## Resources used

- [Haiku's `if.h` implementation](https://github.com/haiku/haiku/blob/d0e006f5569f24e73a4d83f5c9766f1132160d7a/headers/posix/net/if.h#L79)
- [Linux's `if.h` implementation](https://code.woboq.org/linux/linux/include/uapi/linux/if.h.html)
- [FreeBSD's `if.h` implementation](https://github.com/freebsd/freebsd-src/blob/373ffc62c158e52cde86a5b934ab4a51307f9f2e/sys/net/if.h#L142)


panos:
To note, making those definitions exist will not cause any issues.

Some programs that could eventually get ported over may be aware of the definitions existing, but manually assigning of values in the ports themselves in order to make things work will still be necessary, as it is right now.

sin-ack:
The Wine port needed this as well, thanks!

panos:
> The Wine port needed this as well, thanks!

I'm reminding that this most likely won't work from scratch and as-is, from what I'm gathering, or will it?

sin-ack:
> > The Wine port needed this as well, thanks!
> 
> I'm reminding that this most likely won't work from scratch and as-is, from what I'm gathering, or will it?

Yeah, it probably won't work, but after this it's just a matter of actually implementing `SIOCGIFFLAGS` properly. I've stubbed it out in 091d35 and we can implement it later on.

panos:
> > > The Wine port needed this as well, thanks!
> > 
> > 
> Yeah, it probably won't work, but after this it's just a matter of actually implementing `SIOCGIFFLAGS`. I've stubbed it out in [091d352](https://github.com/SerenityOS/serenity/commit/091d3525268a1884ccb14ede0b18f242109b50a4) and we can implement it later on. Wine needed the symbols themselves to compile though, so it helps.

You can still manually override the values, but that's as good as not having them in the way at all. Is there any resource you can point me to, so that I can make the definitions a bit more complete?

panos:
Slightly demotivating.

PR #6448, opened 2021-04-18
Toolchain: Build fix for macOS

gunnarbeutner:
Unfortunately their sed behaves a tiny bit different to how GNU sed does. The changes from #6394 broke the `BuildIt.sh` script.

### DIFF
--- a/Toolchain/BuildIt.sh
+++ b/Toolchain/BuildIt.sh
@@ -78,7 +78,11 @@ GCC_BASE_URL="http://ftp.gnu.org/gnu/gcc"
 buildstep() {
     NAME=$1
     shift
-    "$@" 2>&1 | sed 's|^|\x1b[34m['"${NAME}"']\x1b[39m |'
+    if [ "$(uname -s)" = "Darwin" ]; then
+        "$@" 2>&1 | sed 's|^|['"${NAME}"'] |'
+    else
+        "$@" 2>&1 | sed 's|^|\x1b[34m['"${NAME}"']\x1b[39m |'
+    fi
 }
 
 # === CHECK CACHE AND REUSE ===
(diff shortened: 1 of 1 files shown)


gunnarbeutner:
I've verified that this works on macOS.

Toolchain/BuildIt.sh
```diff
@@ -75,10 +75,16 @@ GCC_NAME="gcc-$GCC_VERSION"
 GCC_PKG="${GCC_NAME}.tar.gz"
 GCC_BASE_URL="http://ftp.gnu.org/gnu/gcc"
 
+SYSTEM_NAME="$(uname -s)"
```
linusg, line 78: this is not used anywhere? Caching the result here is nice as it's not going to change, can you replace the below uname calls with this variable?

gunnarbeutner: Ah yes, I was planning to do that but then I forgot. Thanks, will do.

gunnarbeutner: There we go. Also, I've tested it again on Linux and macOS and it seems to work without problems.

Toolchain/BuildIt.sh
```diff
@@ -78,7 +78,11 @@ GCC_BASE_URL="http://ftp.gnu.org/gnu/gcc"
 buildstep() {
     NAME=$1
     shift
-    "$@" 2>&1 | sed 's|^|\x1b[34m['"${NAME}"']\x1b[39m |'
+    if [ "$(uname -s)" = "Darwin" ]; then
+        "$@" 2>&1 | sed 's|^|['"${NAME}"'] |'
+    else
+        "$@" 2>&1 | sed 's|^|\x1b[34m['"${NAME}"']\x1b[39m |'
+    fi
```
emanuele6:
```suggestion
    "$@" 2>&1 | sed $'s|^|\x1b[34m['"${NAME}"$']\x1b[39m |'
```

@gunnarbeutner: I don't have a mac to test this, but wouldn't this work on both mac and linux?

PR #6453, opened 2021-04-18
LibGfx+FontEditor: Add Hebrew font support + Hebrew character glyphs for Katica Regular 10

IdanHo:
This PR adds support for fonts that include characters up to the Hebrew Unicode Block, as well as adds character glyphs for the Hebrew alphabet that I asked my more artistically inclined brother to draw :)

NOTE: While this makes hebrew characters render fine (for example on google.co.il), this does not add support for RTL text detection and rendering in LibGfx so currently all hebrew text is rendered backwards. This will be addressed in a future PR.


linusg:
It's great how browsing the non-english web within serenity continuously leads to the addition of new font glyphs!

PR #6454, opened 2021-04-18
LibJS: Implement console.assert()

linusg:


Userland/Applications/Browser/BrowserConsoleClient.cpp
```diff
@@ -124,4 +124,25 @@ JS::Value BrowserConsoleClient::count_reset()
...
+{
+    auto& vm = this->vm();
+    if (!vm.argument(0).to_boolean()) {
+        StringBuilder html;
+        if (vm.argument_count() > 1) {
+            html.append("<span class=\"error\">");
+            html.append("Assertion failed:");
+            html.append("</span>");
+            html.append(" ");
+            html.append(vm.join_arguments(1));
```
alimpfard, line 137: Less HTML injecty? unless `join_arguments()` escapes html stuff automatically?

linusg: dammit, I just took that from the various other console functions hoping our past selves have been careful :smile:

fixed!

PR #6455, opened 2021-04-18
LibWeb: Load resources containing spaces in url, choose better fallback fonts

danners:


sin-ack:
Probably better to split this out into two PRs?

Userland/Libraries/LibWeb/CSS/StyleProperties.cpp
```diff
@@ -197,13 +204,26 @@ void StyleProperties::load_font() const
...
 
     m_font = found_font;
     FontCache::the().set(font_selector, *m_font);
 }
 
+RefPtr<Gfx::Font> StyleProperties::font_fallback(bool monospace, bool bold) const
+{
+    if (monospace && bold) {
+        return Gfx::FontDatabase::default_bold_fixed_width_font();
+    } else if (monospace) {
```
awesomekling, line 218: No need for `else` after `return`

PR #6456, opened 2021-04-18
Toolchain/BuildIt.sh: let's have colours on macOS too :^)

emanuele6:
MacOS's sed doesn't support "\x1b" expansion unlike GNU sed, but we can still use bash's `$'\x1b'` to expand it.


gunnarbeutner:
Thanks a lot. Works just fine on my Mac.

PR #6459, opened 2021-04-18
Base/CI: Run tests from /usr/Tests in GitHub Actions

### LINKED ISSUE #5541: Kernel: Running on-target tests without KVM is flaky
Using latest from #4921, the kernel (or the shell?) sporadically times out after 10 minutes while trying to execute the counter loop for what I can judge are two different reasons.

Note that the CI job is running with KVM off.

When calculating the loop, it sometimes gets an assertion failed on RefPtr. This is suspiciously the same assertion that triggers in the macOS tests...
https://github.com/ADKaster/serenity/runs/1992109722?check_suite_focus=true#step:19:325
```
Shell(80): Job entry 'expr 63 + 1' deleted in 323 ms
Shell(10): Job entry 'expr 63 + 1' deleted in 327 ms
[Shell(10:10)]: ASSERTION FAILED: old_ref_count > 0
[Shell(10:10)]: .././AK/RefCounted.h:73 in void AK::RefCountedBase::ref() const
[#0 Shell(10:10)]: 0xc01237cf  __assertion_failed(char const*, char const*, unsigned int, char const*) +0x145
[#0 Shell(10:10)]: 0xc044cd43  Kernel::Thread::from_tid(AK::DistinctNumeric<int, __ThreadID_tag, false, true, true, false, false, false>) +0x29b
[#0 Shell(10:10)]: 0xc03417a1  Kernel::Process::send_signal(unsigned char, Kernel::Process*) +0xdd
[#0 Shell(10:10)]: 0xc03c0549  Kernel::Process::do_kill(Kernel::Process&, int) +0x5bb
[#0 Shell(10:10)]: 0xc03c1543  Kernel::Process::sys$kill(int, int) +0x6db
[#0 Shell(10:10)]: 0xc036e6d9  syscall_handler +0x1f08
[#0 Shell(10:10)]: 0xc036c799  syscall_asm_entry +0x31
```

Separately (maybe separately..) the test runner script sometimes hangs in the shell while trying to calculate the same expr N + 1 subshells. One shell is blocked on read, the parent on select. In the images below, I started SystemMonitor as root from a shell, so shells 25 and 29 should be ignored :).

```
Shell(38): Job entry 'expr 0 + 1' deleted in 1337 ms
Shell(31): Job entry 'expr 0 + 1' deleted in 1353 ms
Shell(39): Job entry 'expr 1 + 1' deleted in 1828 ms
Shell(31): Job entry 'expr 1 + 1' deleted in 1844 ms
Shell(31): JShell(40): Job entry 'expr 2 + 1' deleted in 1152 ms
ob entry 'expr 2 + 1' deleted in 1125 ms
Shell(41): Job entry 'expr 3 + 1' deleted in 1656 ms
Shell(31): Job entry 'expr 3 + 1' deleted in 1739 ms
Shell(42): Job entry 'expr 4 + 1' deleted in 1572 ms
Shell(31): Job entry 'expr 4 + 1' deleted in 1681 ms
Shell(43): Job entry 'expr 5 + 1' deleted in 930 ms
Shell(31): Job entry 'expr 5 + 1' deleted in 939 ms
Shell(31): Job entry 'expr 6 + 1' deleted in 1376 ms
Shell(44): Job entry 'expr 6 + 1' deleted in 1444 ms
Shell(45): Job entry 'expr 7 + 1' deleted in 719 ms
Shell(31): Job entry 'expr 7 + 1' deleted in 729 ms
Shell(31): Job entry 'expr 8 + 1' deleted in 1590 ms
```

PID 46 is reading (he'll be job 'expr 9 + 1'):
reader

PID 31 is selecting in the event loop waiting for SIGCHLD:
selector

cc @alimpfard @tomuta (?)

ADKaster:
Uncomment running the tests that were disabled due to frequent freezes when running without KVM.

Fixes #5541 

This still skips many tests held in the `exclude_patterns` shell variable in the test script.

Also, hide the dbg spam from shell jobs dying behind a debug flag. Without it, the console gets absolutely clobbered when running shell tests.


.github/workflows/cmake.yml
```diff
@@ -157,8 +157,12 @@ jobs:
         echo "::endgroup::"
         [ "y" = ${run_worked} ]
       timeout-minutes: 10
```
IdanHo, line 159: Increasing the timeout will probably help :)

ADKaster: Yeah.... building on my local machine again so I can `time ./run_tests_and_shutdown.sh` to see how long this is *actually* going to take without KVM... I suspect there's a few benchmarks that are ... aggressive.

bgianfo: It looks like most of the time is spent running benchmarks in LibGfx?

bgianfo: Filter out this guy? 

```
Sun, 18 Apr 2021 20:08:54 GMT ==== Running test 11 out of 88 -- /usr/Tests/LibGfx/painter ====
Sun, 18 Apr 2021 20:08:54 GMT /usr/Tests/LibGfx/painter
Sun, 18 Apr 2021 20:08:55 GMT  Running benchmark 'diagonal_lines'.
Sun, 18 Apr 2021 20:10:46 GMT  Running benchmark 'fill'.
Sun, 18 Apr 2021 20:11:04 GMT  Running benchmark 'fill_with_gradient'.
Sun, 18 Apr 2021 20:17:23 GMT  Running 3 cases out of 3.
```

Base/home/anon/tests/run-tests-and-shutdown.sh
```diff
@@ -6,26 +6,46 @@ echo "==== Running Tests on SerenityOS ===="
...
     shift
     test_cmd=($*)
-    echo "Running test $index out of $count_of_all_tests -- $test_cmd"
+    echo "==== Running test $index out of $count_of_all_tests -- $test_cmd ===="
+    echo "::group::$test_cmd"
     if $test_cmd {
         echo "::debug file=$test_cmd:: $test_cmd passed!"
     } else {
+        fail_count=$(expr $fail_count + 1)
+        failed_tests=($failed_tests $test_cmd)
```
alimpfard, line 15: The shell now has the means to let you avoid calling out to `expr` here:
```sh
failed_tests=${concat_lists $failed_test ($test_cmd)}

# later, to get the count
failed_count=${length $failed_tests}
```

.github/workflows/cmake.yml
ADKaster: Jeez. Yeah. I updated to 20 mins (hopefully not necessary) and filtered out that painter benchmark.

Base/home/anon/tests/run-tests-and-shutdown.sh
ADKaster: Vereh nice. Updated to use as suggested.

Base/home/anon/tests/run-tests-and-shutdown.sh
```diff
@@ -6,26 +6,49 @@ echo "==== Running Tests on SerenityOS ===="
...
-# TODO: test-web requires the window server
-system_tests=((test-js --show-progress=false) test-pthread /usr/Tests/LibM/test-math (test-crypto -t bigint))
-# FIXME: Running too much at once is likely to run into #5541. Remove commented out find below when stable
-all_tests=${concat_lists $system_tests} #$(find /usr/Tests -type f | grep -v Kernel | grep -v .inc | shuf))
+# Files in /usr/Tests/* that we don't want to execute match these patterns:
+#     Kernel: Kernel tests that are either old patched exploits, or hang the system/shell
+#     .inc: Shell test helper file that's not a test
+#     UserEmulator: Tests designed to run inside the Userspace Emulator
+#     stack-smash: Intentionally crashes by smashing the stack
+#     TestJSON: AK/TestJSON makes assumptions about $PWD to load its input files
```
bgianfo, line 27: I think TestJSON has been fixed in the time since this was originally sent out.

Base/home/anon/tests/run-tests-and-shutdown.sh
```diff
@@ -6,26 +6,49 @@ echo "==== Running Tests on SerenityOS ===="
...
         echo "::error file=$test_cmd:: $test_cmd returned non-zero exit code, check logs!"
     }
 }
 
-# TODO: test-web requires the window server
-system_tests=((test-js --show-progress=false) test-pthread /usr/Tests/LibM/test-math (test-crypto -t bigint))
-# FIXME: Running too much at once is likely to run into #5541. Remove commented out find below when stable
-all_tests=${concat_lists $system_tests} #$(find /usr/Tests -type f | grep -v Kernel | grep -v .inc | shuf))
+# Files in /usr/Tests/* that we don't want to execute match these patterns:
+#     Kernel: Kernel tests that are either old patched exploits, or hang the system/shell
```
bgianfo, line 23: I don't think this is true anymore... there are tests under `/usr/Tests/Kernel/` that start with Test* which use the LibTest framework and should be passing.

Base/home/anon/tests/run-tests-and-shutdown.sh
bgianfo, line 27: I think TestJSON has been fixed in the time since this was originally sent out.

Base/home/anon/tests/run-tests-and-shutdown.sh
```diff
@@ -6,26 +6,49 @@ echo "==== Running Tests on SerenityOS ===="
...
+#     .inc: Shell test helper file that's not a test
+#     UserEmulator: Tests designed to run inside the Userspace Emulator
+#     stack-smash: Intentionally crashes by smashing the stack
+#     TestJSON: AK/TestJSON makes assumptions about $PWD to load its input files
+#     .frm: Test inputs that are not tests
+#     test-web: Requires the window server in order to work
+#     test-js: We start this one manually with the show progress flag set to false
+exclude_patterns='Kernel|.inc|UserEmulator|stack-smash|TestJSON|.frm|test-web|test-js'
+
+system_tests=((test-js --show-progress=false) (test-crypto -c -t test))
```
bgianfo, line 33: Why don't the `test-js` and `test-math` failures in the run log cause the CI to fail?

Base/home/anon/tests/run-tests-and-shutdown.sh
bgianfo, line 23: I don't think this is true anymore... there are tests under `/usr/Tests/Kernel/` that start with `Test*` which use the LibTest framework and should be passing.

.github/workflows/cmake.yml
```diff
@@ -148,17 +148,14 @@ jobs:
...
         ninja run || run_worked=n
         echo "::endgroup::"
-        echo "::group::cat debug.log # Serenity output"
-        if [ "y" = "${run_worked}" ] ; then
-          cat debug.log
-        else
-          echo "skipped (qemu had non-zero exit-code)"
-        fi
-        echo "::endgroup::"
         [ "y" = ${run_worked} ]
```
bgianfo, line 151: The run_worked thing doesn't seem to be used anymore... now that dumping the log is a second step?

bgianfo (approved):

.github/workflows/cmake.yml
ADKaster: Ah true. I can clean this up more. The previous implementation BenW put here was clever, but if the ninja run step timed out it wouldn't dump the kernel log... which is kinda the opposite of what we want. That's when the debug log is most useful.

Base/home/anon/tests/run-tests-and-shutdown.sh
ADKaster: True. I'll move the older tests that need some help to /usr/Tests/Kernel/Legacy and just skip those.

Base/home/anon/tests/run-tests-and-shutdown.sh
ADKaster: Hmm, I just ran it from /home/anon and it failed the same assertion as always. It works if you `cd` to /usr/Tests/AK` though. So it still makes assumptions about $PWD. :/ Just, assumptions that are a bit more valid than they were a month ago. Feels like something to cleanup in the next round that adds more "test configuration" type features.

Base/home/anon/tests/run-tests-and-shutdown.sh
ADKaster: Only returning a non-zero exit code from the script would fail the CI job.

https://docs.github.com/en/actions/creating-actions/setting-exit-codes-for-actions
https://github.com/actions/runner/issues/662

Was talking with CxByte about this on discord earlier, might be worth to write out the "test result" to a file and mount the disk image after ninja run to check out what's up with it? But curiously, when I run locally in Qemu with CPU=max,vmx=off, those particular tests don't fail...

Base/home/anon/tests/run-tests-and-shutdown.sh
bgianfo: I see.. I thought I remembered someone deleting the test that depends on the file as input, but I guess I was wrong.

ADKaster: The dependency on the ancient 4chan.json dump file was removed, but that TestJSON test still uses `test.frm` in order to work.

```c++

TEST_CASE(load_form)
{
    FILE* fp = fopen("test.frm", "r");
    VERIFY(fp);

```

Userland/Shell/Job.h
```diff
@@ -31,7 +27,7 @@ class Job : public RefCounted<Job> {
 
     ~Job()
     {
-#ifdef JOB_TIME_INFO
+#if SHELL_JOB_DEBUG
```
alimpfard, line 30: n i t:
```suggestion
if constexpr (SHELL_JOB_DEBUG) {
```

alimpfard: (and replace the `#endif` below with the good old `}`)

ADKaster: le fixed, and moved the macro in AK/Debug.in.h to its proper sorted location at the same time

PR #6461, opened 2021-04-18
AK/RedBlackTree: Refactor redundant code

ldm5180:
Problem:
- `virtual` destructors in otherwise non-`virtual` classes.
- Non-explicit single argument constructor to `AK::RedBlackTree::Node`
- Redundant code with just side of tree differences.

Solution:
- Remove virtual destructors.
- Make `AK::RedBlackTree::Node` constructor explicit because it only has a single argument.
- Commonize code using template arguments to determine the side.

### DIFF
--- a/AK/RedBlackTree.h
+++ b/AK/RedBlackTree.h
@@ -26,6 +26,7 @@
 
 #pragma once
 
+#include <AK/Array.h>
 #include <AK/Concepts.h>
 
 namespace AK {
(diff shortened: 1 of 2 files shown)


AK/RedBlackTree.h
```diff
@@ -162,54 +140,54 @@ class BaseRedBlackTree {
...
+            parent->left_child() = node;
         } else { // we are the right child
-            parent->right_child = node;
+            parent->right_child() = node;
         }
-        node->parent = parent;
+        node->parent() = parent;
 
-        if (node->parent->parent) // no fixups to be done for a height <= 2 tree
+        if (node->parent()->parent()) // no fixups to be done for a height <= 2 tree
```
emanuele6, line 149: Shouldn't this be `node->parent().parent()` since `Node::parent()` returns a reference?

AK/RedBlackTree.h
```diff
@@ -301,91 +279,86 @@ class BaseRedBlackTree {
...
+                        sibling = parent->left_child();
                     }
                     sibling->color = parent->color;
                     parent->color = Color::Black;
-                    sibling->left_child->color = Color::Black; // null check?
+                    sibling->left_child()->color = Color::Black; // null check?
                     rotate_right(parent);
                     node = m_root; // fixed
                 }
             }
```
IdanHo, line 330: If were going for code deduplication, these 2 branches are symmteric, the only differences being left and right rotation being flipped, as well as left and right children are flipped, eliminating half the branches will result in performance uplift

AK/RedBlackTree.h
ldm5180: @emanuele6 It is a reference to a pointer. The ugliness of pointers. Is there a good way to make this more readable?

AK/RedBlackTree.h
ldm5180: @IdanHo Agreed. I just didn't make it that far yet. Thanks for pointing out this additional spot.

IdanHo: No problem, note that simply extracting and templating the branch will only deduplicate code, not make it more performant, in order to reduce executed branches the code has to be written once, but act upon the nodes array based on the direction that is computed at the start

AK/RedBlackTree.h
emanuele6: > It is a reference to a pointer

Ah, I didn't notice that was a `AK::Array` of `Node*` (makes sense).

I think it's readable enough. I was just confused by the `auto&` return type.

PR #6462, opened 2021-04-18
CMake: Quiet warnings about literal suffix

ldm5180:
Problem:
- Newer versions of clang (ToT) have a similar `-Wliteral-suffix`
  warning as GCC. A previous commit enabled it for all compilers. This
  needs to be silenced for the entire build, but it currently only is
  silenced for some directories.

Solution:
- Move the `-Wno-literal-suffix` option up in the CMakeLists.txt so
  that it gets applied everywhere.


ldm5180:
@alimpfard This should fix the noisy build issue I created.

PR #6463, opened 2021-04-18
LibSQL: Introducing a SQL library for SerenityOS

trflynn89:
The motivation here is that applications like Browser are going to want to store state to a local database (history, cookies, cache metadata, etc.). That gets tricky when e.g. multiple browsers are running and race to update state.

LibSQL's goal is to more or less be a SQLite clone. SQLite's architecture looks to be super well documented, so that seems to be the path of least resistance.

This is very barebones right now - it basically just tells you whether you entered a `CREATE TABLE` or `DROP TABLE` statement correctly :) but gotta start somewhere!

sql


Userland/Libraries/LibSQL/Lexer.h
```diff
@@ -0,0 +1,71 @@
...
+#pragma once
+
+#include "Token.h"
+#include <AK/HashMap.h>
+#include <AK/String.h>
+#include <AK/StringView.h>
+
+namespace SQL {
+
+class Lexer {
```
ADKaster, line 36: Would this Lexer benefit from inheriting from `AK::GenericLexer` ?

trflynn89: Hmm, this lexer and LibJS's lexer need to track the current column and line number in the source view, which `GenericLexer` doesn't do. Maybe it makes sense to have `GenericLexer` do that (which feels like it should be separate PR)? But as it is now, this lexer would have to encapsulate all of the `GenericLexer::consume*` methods to do that extra tracking.

bgianfo:
Consider establishing a basic test harness now, so if other people come along there's an established way to add new functionality / make sure the don't break your existing basic scenarios.

Userland/Libraries/LibSQL/Lexer.h
ADKaster: The lexer in LibCpp does similar tracking of line and column, so it's not clear to me if that would belong to generic lexer, or some kind of SourceCodeLexer abstraction. Definitely feels separate though

trflynn89:
> Small bit of feedback: Consider establishing a basic test harness now / soon, so if other people come along there's an established way to add new functionality / make sure the don't break your existing basic scenarios.

@bgianfo good call, I added tests for the parser. I also added the test and `sql` to Lagom - even though Serenity boots up super quick, I found it real convenient to just run the tests from my terminal :) Plus, I imagine we'll want to fuzz this stuff eventually.

Userland/Libraries/LibSQL/Lexer.h
benit8: Oh my, I didn't think people were still considering the `GenericLexer`. I originally wanted to extend most of Serenity's lexers with it (so it would make sense for it to track lines and columns), but life happened and I've only been lurking around for the last year... But you are right about it's limitations; I'll try to make some time and implement what you're looking for :slightly_smiling_face:

trflynn89:
Added parsing for `DROP TABLE` as well since it's such a simple expression

Userland/Libraries/LibSQL/AST.h
```diff
@@ -125,4 +125,23 @@ class CreateTable : public Statement {
     bool m_is_error_if_table_exists;
 };
 
+class DropTable : public Statement {
+public:
+    DropTable(String schema_name, String table_name, bool is_error_if_table_does_not_exist)
+        : m_schema_name(move(schema_name))
+        , m_table_name(move(table_name))
+        , m_is_error_if_table_does_not_exist(move(is_error_if_table_does_not_exist))
```
awesomekling, line 133: No need to `move()` a bool. :^)

PR #6464, opened 2021-04-18
LibWeb: Support thead & tfooter HTML Table tags

ant1441:
Correctly handle the `<thead>` and `<tfooter>` HTML tags within the `<table>` tag.

These tags are treated the same as `<tbody>` in most cases.

This also adds a HTML test page including a `<thead>` to verify the change.

### DIFF
--- a/Base/res/html/misc/cursor.html
+++ b/Base/res/html/misc/cursor.html
@@ -1,0 +1,278 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <title>Cursor Test Page</title>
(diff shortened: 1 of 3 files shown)


awesomekling:
Nice patches :)

If you want to go deep on table support, look into the [table processing model](https://html.spec.whatwg.org/multipage/tables.html#table-processing-model) from the HTML spec. It should be fairly straightforward to implement the "forming a table" algorithm in our HTML DOM, and then implementing CSS table layout on top of that. :)

PR #6475, opened 2021-04-19
LibWeb: Add DOM spec compliant Attributes (and related APIs)

FalseHonesty:
When trying to run jQuery in the browser the other day, it errored on [`Element.attributes`](https://developer.mozilla.org/en-US/docs/Web/API/Element/attributes) not existing, so this PR brings that API into existence, along with all of the other required Attribute changes.

At a high level this PR adds

- IDL getter/setter support
- IDL sequence support
- A DOM spec compliant `Attr`
- Support for all `Attr` related methods in Element and Document

and finally...
- The DOM spec's NamedNodeMap structure (solving the initial problem)

Thanks to @mattco98 for his help with the IDL generator :^)


awesomekling (changes requested):
The `Attr` API is not actually used by the vast majority of web content, so it's a common engine optimization to only create synthetic `NamedNodeMap` and `Attr` objects on the fly when requested to avoid wasting large amounts of memory.

We should implement this optimization as well, instead of bloating our runtime with NNM/Attr. :)

FalseHonesty:
> The `Attr` API is not actually used by the vast majority of web content, so it's a common engine optimization to only create synthetic `NamedNodeMap` and `Attr` objects on the fly when requested to avoid wasting large amounts of memory.
> 
> 
> 
> We should implement this optimization as well, instead of bloating our runtime with NNM/Attr. :)

I did originally implement it with that optimization, however I ran into issues with variable lifetimes. Makes sense though, I'll give it another go.

PR #6477, opened 2021-04-19
Kernel + LibC: + profile: Expose ability to free kernel profiling buffer, and the ability to disable profiling on user input

bgianfo:
A few changes here:

-  Add a syscall to clear the profiling buffer, expose it via LibC/profile frontend.

    While profiling all processes the profile buffer lives forever.
    Once you have copied the profile to disk, there's no need to keep it
    in memory. This syscall surfaces the ability to clear that buffer.

-  Add an option to wait for user input to disable profiling.

    Often you want to enable and then disable profiling after a period of time.
    To make this slightly more ergonomic, add an option to wait for user
    input after enabling profiling, this will disable profiling on exit
    after the key press.

### DIFF
--- a/Kernel/API/Syscall.h
+++ b/Kernel/API/Syscall.h
@@ -171,6 +171,7 @@ namespace Kernel {
     S(purge)                  \
     S(profiling_enable)       \
     S(profiling_disable)      \
+    S(profiling_free_buffer)  \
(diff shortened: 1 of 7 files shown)


awesomekling:
Good idea :)
The user interface is a little awkward but it'll do!

PR #6479, opened 2021-04-19
LibC: Implemented getpass()

jubnzv:
`getpass()` has now been implemented as part of the standard C library.

### DIFF
--- a/Userland/Libraries/LibC/unistd.cpp
+++ b/Userland/Libraries/LibC/unistd.cpp
@@ -781,8 +781,50 @@ ssize_t pread(int fd, void* buf, size_t count, off_t offset)
 
 char* getpass(const char* prompt)
 {
-    dbgln("FIXME: getpass('{}')", prompt);
-    TODO();
+    FILE *in, *out;
+    bool tty_changed = false;
+    struct termios s, t;
+    static char* buf;
+    static size_t bufsize;
+    ssize_t nread;
+
+    if ((in = fopen("/dev/tty", "w+e")) == NULL) {
+        in = stdin;
+        out = stderr;
+    } else {
+        out = in;
+    }
+
+    if (tcgetattr(fileno(in), &t) != -1) {
+        s = t;
+        t.c_lflag &= ~(ECHO | ISIG);
+        if (tcsetattr(fileno(in), TCSAFLUSH, &t) == 0)
+            tty_changed = true;
+    }
+
+    fprintf(out, "%s", prompt);
+    fflush(out);
+
+    nread = getline(&buf, &bufsize, in);
+    if (buf != NULL) {
+        if (nread < 0) {
+            buf[0] = '\0';
+        } else if (buf[nread - 1] == '\n') {
+            buf[nread - 1] = '\0';
+            if (tty_changed) {
+                // Write the newline that was not echoed.
+                fprintf(out, "\n");
+            }
+        }
+    }
+
+    if (tty_changed)
+        (void)tcsetattr(fileno(in), TCSAFLUSH, &s);
+
+    if (in != stdin)
+        fclose(in);
+
+    return buf;
 }
 
 long sysconf(int name)


jubnzv:
CI fail is not related to my changes:

> E: Failed to fetch http://apt.llvm.org/focal/dists/llvm-toolchain-focal-12/main/binary-amd64/Packages  File has unexpected size (8993 != 8785). Mirror sync in progress? [IP: 151.101.210.49 443]
> E: Some index files failed to download. They have been ignored, or old ones used instead.
> Error: Process completed with exit code 100.

Userland/Libraries/LibC/unistd.cpp
```diff
@@ -781,8 +781,50 @@ ssize_t pread(int fd, void* buf, size_t count, off_t offset)
...
-    dbgln("FIXME: getpass('{}')", prompt);
-    TODO();
+    FILE *in, *out;
+    bool tty_changed = false;
+    struct termios s, t;
+    static char* buf;
+    static size_t bufsize;
+    ssize_t nread;
+
+    if ((in = fopen("/dev/tty", "w+e")) == NULL) {
```
awesomekling, line 791: There is no `/dev/tty` on SerenityOS yet, so this cannot possibly work.

Please don't submit code that you have not tested.

Also, a quick Google search shows me that this is LGPL licensed code that you have stolen from **glibc**. This is not okay.

awesomekling (changes requested):

Userland/Libraries/LibC/unistd.cpp
jubnzv: > Please don't submit code that you have not tested.

I tested this code before PR and it works:

image

> There is no /dev/tty on SerenityOS yet

Thanks, I'll take a look. Currently `fopen()` on `/dev/tty` fails, so I set `in == stdin` and it works.

> Also, a quick Google search shows me that this is LGPL licensed code that you have stolen from glibc

No, it's not stolen from *glibc*. This code does not contain glibc-specific functions that are not present in Serenety LibC.
While working on this I also take a look at `getpass()` implementations in *musl* and *uclibc*. But I found that the implementation in *glibc* is more suitable for my task, since it contains an implementation of error handling. 

So I would like the logic of this function the same as in *glibc*. After removing /dev/tty, it will not look like a glibc implementation. Is this acceptable? I don't think there is any other way to implement the `getpass()`.

awesomekling: It's completely obvious that you copied this from glibc. The variable naming and declaration order is identical, and you even forgot to remove one of their comments.

And now instead of admitting it, you insist that it's not copied. Give me a break.

It would have taken you 5 minutes to look at the glibc implementation, understand what it does, close the window, and write your own implementation from scratch. Nobody would have known.

You could also have copied the code and changed all the variable names and re-styled it as C++ (this isn't an old C project, we don't put all the variable declarations at the top of the function.) Nobody would have known.

"Your" implementation:

```cpp
char* getpass(const char* prompt)
{
    FILE *in, *out;
    bool tty_changed = false;
    struct termios s, t;
    static char* buf;
    static size_t bufsize;
    ssize_t nread;

    if ((in = fopen("/dev/tty", "w+e")) == NULL) {
        in = stdin;
        out = stderr;
    } else {
        out = in;
    }

    if (tcgetattr(fileno(in), &t) != -1) {
        s = t;
        t.c_lflag &= ~(ECHO | ISIG);
        if (tcsetattr(fileno(in), TCSAFLUSH, &t) == 0)
            tty_changed = true;
    }

    fprintf(out, "%s", prompt);
    fflush(out);

    nread = getline(&buf, &bufsize, in);
    if (buf != NULL) {
        if (nread < 0) {
            buf[0] = '\0';
        } else if (buf[nread - 1] == '\n') {
            buf[nread - 1] = '\0';
            if (tty_changed) {
                // Write the newline that was not echoed.
                fprintf(out, "\n");
            }
        }
    }

    if (tty_changed)
        (void)tcsetattr(fileno(in), TCSAFLUSH, &s);

    if (in != stdin)
        fclose(in);

    return buf;
}
```

glibc's implementation:

```cpp
char *
getpass (const char *prompt)
{
  FILE *in, *out;
  struct termios s, t;
  int tty_changed;
  static char *buf;
  static size_t bufsize;
  ssize_t nread;

  /* Try to write to and read from the terminal if we can.
     If we can't open the terminal, use stderr and stdin.  */

  in = fopen ("/dev/tty", "w+ce");
  if (in == NULL)
    {
      in = stdin;
      out = stderr;
    }
  else
    {
      /* We do the locking ourselves.  */
      __fsetlocking (in, FSETLOCKING_BYCALLER);

      out = in;
    }

  /* Make sure the stream we opened is closed even if the thread is
     canceled.  */
  __libc_cleanup_push (call_fclose, in == out ? in : NULL);

  flockfile (out);

  /* Turn echoing off if it is on now.  */

  if (__tcgetattr (fileno (in), &t) == 0)
    {
      /* Save the old one. */
      s = t;
      /* Tricky, tricky. */
      t.c_lflag &= ~(ECHO|ISIG);
      tty_changed = (tcsetattr (fileno (in), TCSAFLUSH|TCSASOFT, &t) == 0);
    }
  else
    tty_changed = 0;

  /* Write the prompt.  */
  __fxprintf (out, "%s", prompt);
  __fflush_unlocked (out);

  /* Read the password.  */
  nread = __getline (&buf, &bufsize, in);
  if (buf != NULL)
    {
      if (nread < 0)
	buf[0] = '\0';
      else if (buf[nread - 1] == '\n')
	{
	  /* Remove the newline.  */
	  buf[nread - 1] = '\0';
	  if (tty_changed)
	    /* Write the newline that was not echoed.  */
	    __fxprintf (out, "\n");
	}
    }

  /* Restore the original setting.  */
  if (tty_changed)
    (void) tcsetattr (fileno (in), TCSAFLUSH|TCSASOFT, &s);

  funlockfile (out);

  __libc_cleanup_pop (0);

  if (in != stdin)
    /* We opened the terminal; now close it.  */
    fclose (in);

  return buf;
}
```

PR #6480, opened 2021-04-19
LibELF: Fix calculation for TLS relocations

gunnarbeutner:
The calculation for TLS relocations was incorrect which would result in overlapping TLS variables when more than one shared
object used TLS variables.

This bug can be reproduced with a shared library and a program like this:

```
    $ cat tlstest.c
    #include <string.h>
    __thread char tls_val[1024];
    void set_val() { memset(tls_val, 0, sizeof(tls_val)); }

    $ gcc -g -shared -o usr/lib/libtlstest.so tlstest.c

    $ cat test.c
    void set_val();
    int main() { set_val(); }
    $ gcc -g -o tls test.c -ltlstest
```

Due to the way the TLS relocations are done this program would clobber libc's TLS variables (e.g. `errno`).


awesomekling:
Great catch :)

gunnarbeutner:
Thanks! libgcc/libstdc++ just so happen to exercise a lot of ELF's functionality when building stuff with threads. :)

PR #6484, opened 2021-04-19
Ports: Add port for xz

gunnarbeutner:
The `libpthread` changes are necessary to make the `xz` port properly detect `pthread` support. The two functions are only used in the configure script and nowhere else.

This port depends on #6343, i.e. won't build properly without it because `-pthread` isn't supported.

### DIFF
--- a/Userland/Libraries/LibPthread/pthread.cpp
+++ b/Userland/Libraries/LibPthread/pthread.cpp
@@ -143,6 +143,16 @@ void pthread_exit(void* value_ptr)
     exit_thread(value_ptr);
 }
 
+void pthread_cleanup_push([[maybe_unused]] void (*routine)(void*), [[maybe_unused]] void* arg)
+{
+    TODO();
+}
+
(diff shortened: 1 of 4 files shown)


Userland/Libraries/LibPthread/pthread.cpp
```diff
@@ -143,6 +143,19 @@ void pthread_exit(void* value_ptr)
     exit_thread(value_ptr);
 }
 
+void pthread_cleanup_push(void (*routine)(void *), void *arg)
+{
+    (void)routine;
```
ADKaster, line 148: The Serenity style for this would be to leave the parameters unnamed in the function signature, I think. Also, should this go in a separate commit, like `LibPthread: add stubs for pthread_cleanup_push/pop`?

gunnarbeutner: Aye, you're right. Will get this fixed in a bit. :+1:

gunnarbeutner: There we go. :)

PR #6485, opened 2021-04-19
LibWeb: only allow nested iframes up to a certain depth

### LINKED ISSUE #6449: Browser: Nested iframes crash the browser
When there is a page that includes itself in an iframe, the browser crashes with OOM.
I tried looking at https://cbracco.github.io/html5-test-page/ when this occurred and later checked this behaviour with the included file.
[iframe.zip](https://github.com/SerenityOS/serenity/files/6331116/iframe.zip)

In firefox iframes are rendered a few levels deep and subsequent iframes aren't rendered at all. The exact number of the allowed nesting level is somewhat inconsistent, the above link renders three levels deep while the test file renders only two levels.

Whatever the wanted behaviour is, infinitely recursing is definitely not the way to go.

TobyAsE:
Previously if a document included itself in an <iframe> we would try to infinitely display that until we ran out of memory.
Now the Frame is made aware of its recursion level and stuff handling Frames (i. e. the FrameHostElement) can orient their behavior on that information.

This fixes #6449

### DIFF
--- a/Userland/Libraries/LibWeb/HTML/FrameHostElement.cpp
+++ b/Userland/Libraries/LibWeb/HTML/FrameHostElement.cpp
@@ -26,8 +26,11 @@ void FrameHostElement::inserted()
     HTMLElement::inserted();
     if (!is_connected())
         return;
-    if (auto* frame = document().frame())
+    if (auto* frame = document().frame()) {
         m_content_frame = Frame::create_subframe(*this, frame->main_frame());
+        m_content_frame->set_frame_nesting_levels(frame->frame_nesting_levels());
+        m_content_frame->register_frame_nesting(document().url());
+    }
 }
 
 Origin FrameHostElement::content_origin() const
--- a/Userland/Libraries/LibWeb/Loader/FrameLoader.cpp
+++ b/Userland/Libraries/LibWeb/Loader/FrameLoader.cpp
@@ -136,6 +136,11 @@ bool FrameLoader::load(const LoadRequest& request, Type type)
         return false;
     }
 
+    if (!m_frame.is_frame_nesting_allowed(request.url())) {
+        dbgln("No further recursion is allowed for the frame, abort load!");
+        return false;
+    }
+
     auto& url = request.url();
 
     set_resource(ResourceLoader::the().load_resource(Resource::Type::Generic, request));
--- a/Userland/Libraries/LibWeb/Page/Frame.cpp
+++ b/Userland/Libraries/LibWeb/Page/Frame.cpp
@@ -286,4 +286,14 @@ void Frame::unregister_viewport_client(ViewportClient& client)
     VERIFY(was_removed);
 }
 
+void Frame::register_frame_nesting(URL const& url)
+{
+    m_frame_nesting_levels.ensure(url)++;
+}
+
+bool Frame::is_frame_nesting_allowed(URL const& url) const
+{
+    return m_frame_nesting_levels.get(url).value_or(0) < 3;
+}
+
 }
--- a/Userland/Libraries/LibWeb/Page/Frame.h
+++ b/Userland/Libraries/LibWeb/Page/Frame.h
@@ -81,6 +81,12 @@ public:
 
     void did_edit(Badge<EditEventHandler>);
 
+    void register_frame_nesting(URL const&);
+    bool is_frame_nesting_allowed(URL const&) const;
+
+    void set_frame_nesting_levels(const HashMap<URL, size_t> frame_nesting_levels) { m_frame_nesting_levels = move(frame_nesting_levels); };
+    HashMap<URL, size_t> const& frame_nesting_levels() const { return m_frame_nesting_levels; }
+
 private:
     explicit Frame(DOM::Element& host_element, Frame& main_frame);
     explicit Frame(Page&);
@@ -105,6 +111,8 @@ private:
     bool m_cursor_blink_state { false };
 
     HashTable<ViewportClient*> m_viewport_clients;
+
+    HashMap<URL, size_t> m_frame_nesting_levels;
 };
 
 }


Userland/Libraries/LibWeb/HTML/FrameHostElement.cpp
```diff
@@ -30,6 +30,8 @@
 #include <LibWeb/Origin.h>
 #include <LibWeb/Page/Frame.h>
 
+#define MAX_RECURSION_DEPTH 3
```
awesomekling, line 33: Please use a `static constexpr` variable instead of a macro :)

Userland/Libraries/LibWeb/HTML/FrameHostElement.cpp
```diff
@@ -46,8 +48,17 @@ void FrameHostElement::inserted()
...
     if (!is_connected())
         return;
-    if (auto* frame = document().frame())
+    if (auto* frame = document().frame()) {
+        auto recursion_level = frame->get_recursion_level() + 1;
+
+        // This seems like a total hack, but it looks as if firefox is doing the same:
+        // Once a certain recursion-threshold is reached it just renders an empty document.
+        if (recursion_level > MAX_RECURSION_DEPTH)
+            document().remove_all_children();
```
awesomekling, line 57: This is still a total hack though; I think it would be cleaner to simply refuse to load anything into the subframe at the max level, instead of nuking the entire DOM which may have already run arbitrary JavaScript, etc.

awesomekling (changes requested):

Userland/Libraries/LibWeb/HTML/FrameHostElement.cpp
TobyAsE: I put the responsibility into the FrameLoader, which can decide to not even load the content if the recursion threshold is exceeded for the frame.

awesomekling (changes requested):

Userland/Libraries/LibWeb/Page/Frame.h
```diff
@@ -101,6 +103,10 @@ class Frame : public TreeNode<Frame> {
 
     void did_edit(Badge<EditEventHandler>);
 
+    size_t get_recursion_level() { return m_recursion_level; }
```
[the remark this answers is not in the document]
TobyAsE: In Firefox its not really clear to me how many iframes are allowed to render. In my simple test file which just includes itself I get a maximum of 3, opening https://cbracco.github.io/html5-test-page/ I get 4 as the deepest level. It doesn't matter that much however, as long as it's a finite number. What would you prefer?

`frame_nesting_level`sounds good, way more descriptive of what it's actually used for, I'll change it to that!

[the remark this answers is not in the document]
awesomekling: So including yourself is one thing, but what if you have a deeply nested set of frames that are actually different files?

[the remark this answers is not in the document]
TobyAsE: It seems you can nest as many iframes as you want as long as they all have different `src`. The recursion protection only hits if you try to include an iframe you have already included twice in the stack.
So we have to track every nested iframe and its src for the page and only bail out when appropriate.

I'll see what I can come up with.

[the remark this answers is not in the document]
TobyAsE: Okay, now every Frame keeps track of the nesting in a HashMap, where for each URL the nesting level is recorded. The Map gets passed from the parent to the children when a new iframe is created. 

I don't think this can be done in a more central way as multiple iframes with the same src on the same page should be no problem (and this way they aren't).

TobyAsE:
@awesomekling may I bump this pr?

Userland/Libraries/LibWeb/Page/Frame.h
```diff
@@ -101,6 +103,12 @@ class Frame : public TreeNode<Frame> {
 
     void did_edit(Badge<EditEventHandler>);
 
+    void register_frame_nesting(URL url);
+    bool is_frame_nesting_allowed(URL url);
+
+    void set_frame_nesting_levels(const HashMap<URL, size_t> frame_nesting_levels) { m_frame_nesting_levels = frame_nesting_levels; };
```
awesomekling, line 109: Please use `move()` to avoid the `HashMap` copy here:
```cpp
void set_frame_nesting_levels(HashMap<URL, size_t> frame_nesting_levels) { m_frame_nesting_levels = move(frame_nesting_levels); }
```

Userland/Libraries/LibWeb/Page/Frame.h
```diff
@@ -101,6 +103,12 @@ class Frame : public TreeNode<Frame> {
...
+    HashMap<URL, size_t> const frame_nesting_levels() { return m_frame_nesting_levels; }
```
awesomekling, line 110: Let's return this by const-reference to avoid a copy on each call.
Also let's make it const-qualified:
```cpp
HashMap<URL, size_t> const& frame_nesting_levels() const { return m_frame_nesting_levels; }
```

Userland/Libraries/LibWeb/Page/Frame.h
```diff
@@ -101,6 +103,12 @@ class Frame : public TreeNode<Frame> {
 
     void did_edit(Badge<EditEventHandler>);
 
+    void register_frame_nesting(URL url);
+    bool is_frame_nesting_allowed(URL url);
```
awesomekling, line 107: Since this does not modify the `Frame` object, let's make it const-qualified.
Also since we are not passing ownership of the `URL`, let's pass it by const-reference:
```cpp
bool is_frame_nesting_allowed(URL const&) const;
```

Userland/Libraries/LibWeb/Loader/FrameLoader.cpp
```diff
@@ -157,6 +157,11 @@ bool FrameLoader::load(const LoadRequest& request, Type type)
         return false;
     }
 
+    if (!m_frame.is_frame_nesting_allowed(request.url().to_string())) {
```
awesomekling, line 160: `is_frame_nesting_allowed()` already takes a `URL`, so I don't think we need to `to_string()` here. That will only cause an unnecessary URL->String->URL round-trip.

Userland/Libraries/LibWeb/Page/Frame.cpp
```diff
@@ -305,4 +305,18 @@ void Frame::unregister_viewport_client(ViewportClient& client)
     VERIFY(was_removed);
 }
 
+void Frame::register_frame_nesting(URL url)
+{
+    if (!m_frame_nesting_levels.contains(url))
+        m_frame_nesting_levels.set(url, 0);
+    (m_frame_nesting_levels.find(url)->value)++;
```
awesomekling, line 312: We can use `HashMap::ensure()` to simplify this function:

```cpp
m_frame_nesting_levels.ensure(url)++;
```

Userland/Libraries/LibWeb/Page/Frame.cpp
```diff
@@ -305,4 +305,18 @@ void Frame::unregister_viewport_client(ViewportClient& client)
...
+}
+
+bool Frame::is_frame_nesting_allowed(URL url)
+{
+    if (m_frame_nesting_levels.contains(url) && m_frame_nesting_levels.find(url)->value >= max_recursion_depth)
+        return false;
+    return true;
```
awesomekling, line 319: We can use `HashMap::get()` here to simplify this function:
```cpp
return m_frame_nesting_levels.get(url).value_or(0) >= max_recursion_depth;
```

Userland/Libraries/LibWeb/Page/Frame.h
```diff
@@ -41,6 +41,8 @@
 
 namespace Web {
 
+static constexpr size_t max_recursion_depth = 3;
```
awesomekling, line 44: This name is way too generic to put in the `Web` namespace.
Surely there are a bunch of other things that may also want to have a recursion limit. :)
Since it's not actually used outside of `is_frame_nesting_allowed()`, let's just put it in there.

awesomekling (changes requested):

Userland/Libraries/LibWeb/Page/Frame.cpp
TobyAsE: This PR turns into a lesson of HashMap methods :) Thanks for pointing them out!

PR #6487, opened 2021-04-19
AK::Tests: Fix file installs for TestJSON

Hendiadyoin1:
Actually installs the files needed for TestJSON

the test.frm still seems broken, it is supposed to be a symbolic link, but maybe windows messes with that...
Can somebody test this on native Linux?

### DIFF
--- a/AK/Tests/CMakeLists.txt
+++ b/AK/Tests/CMakeLists.txt
@@ -63,3 +63,7 @@ foreach(source ${AK_TEST_SOURCES})
     target_link_libraries(${name} LibCore)
     install(TARGETS ${name} RUNTIME DESTINATION usr/Tests/AK)
 endforeach()
+
+get_filename_component(TEST_FRM_RESOLVED ./test.frm REALPATH)
+install(FILES ./4chan_catalog.json DESTINATION usr/Tests/AK)
+install(FILES ${TEST_FRM_RESOLVED} DESTINATION usr/Tests/AK)


linusg:
> the test.frm still seems broken, it is supposed to be a symbolic link, but maybe windows messes with that...
> Can somebody test this on native Linux?

Doesn't work on Linux either, the symbolic link just gets copied and can't be resolved within serenity (points to `../../Base/home/anon/Source/little/test.frm`). Use this instead:

```cmake
get_filename_component(TEST_FRM_RESOLVED ./test.frm REALPATH)
install(FILES ./4chan_catalog.json DESTINATION usr/Tests/AK)
install(FILES ${TEST_FRM_RESOLVED} DESTINATION usr/Tests/AK)
```

Please also make the commit message more descriptive, something like `AK/Tests: Install test fixtures required for TestJSON`

linusg (changes requested):
> the test.frm still seems broken, it is supposed to be a symbolic link, but maybe windows messes with that...
> Can somebody test this on native Linux?

Doesn't work on Linux either, the symbolic link just gets copied and can't be resolved within serenity (points to `../../Base/home/anon/Source/little/test.frm`). Use this instead:

```cmake
get_filename_component(TEST_FRM_RESOLVED ./test.frm REALPATH)
install(FILES ./4chan_catalog.json DESTINATION usr/Tests/AK)
install(FILES ${TEST_FRM_RESOLVED} DESTINATION usr/Tests/AK)
```

Please also make the commit message more descriptive, something like `AK/Tests: Install test fixtures required for TestJSON`

Hendiadyoin1:
better now?

linusg:
yes!

PR #6482, opened 2021-04-19
Ports: Improve consistency and quality of ports

panos:
See: https://github.com/SerenityOS/serenity/pull/6444#discussion_r615617959

PR #6490, opened 2021-04-19
Ports: Changed variable names in .hosted_defs.sh

panos:
- DESTDIR and SERENITY_ROOT have not been removed for the time being, in order to ensure a steady transition and to not break all the ports at once.

... Lots of stuff being unpacked here. This was implemented based on this suggestion: https://github.com/SerenityOS/serenity/pull/6482#issuecomment-822356488

In other news, I figured out where the variables are located and what each variable does, so that's good.

### DIFF
--- a/Ports/.hosted_defs.sh
+++ b/Ports/.hosted_defs.sh
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
-export SERENITY_ROOT="$(realpath "${SCRIPT}/../")"
+export SERENITY_SOURCE_DIR="$(realpath "${SCRIPT}/../")"
 export SERENITY_BUILD_DIR="${SERENITY_ROOT}/Build/${SERENITY_ARCH}"
 export CC="${SERENITY_ARCH}-pc-serenity-gcc"
 export CXX="${SERENITY_ARCH}-pc-serenity-g++"
@@ -9,6 +9,11 @@ export PATH="${SERENITY_ROOT}/Toolchain/Local/${SERENITY_ARCH}/bin:${PATH}"
 export PKG_CONFIG_DIR=""
 export PKG_CONFIG_SYSROOT_DIR="${SERENITY_BUILD_DIR}/Root"
 export PKG_CONFIG_LIBDIR="${PKG_CONFIG_SYSROOT_DIR}/usr/lib/pkgconfig/:${PKG_CONFIG_SYSROOT_DIR}/usr/local/lib/pkgconfig"
+
+# To be removed.
+export SERENITY_ROOT="$(realpath "${SCRIPT}/../")"
+
 enable_ccache
 
 DESTDIR="${SERENITY_BUILD_DIR}/Root"
+export SERENITY_INSTALL_ROOT="$DESTDIR"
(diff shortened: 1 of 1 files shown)


panos:
The linting's a bit weird, but I didn't want to change the order of the variables or not leave a newline before a comment. Either way, that's supposed to be temporary until all ports successfully transition to the new format and until the documentation's updated.

linusg:
As mentioned on Discord, I think we can choose even more descriptive names for these while we're here. Something like `SERENITY_SOURCE_ROOT` and `SERENITY_INSTALL_ROOT`, but suggestions welcome. I also think `SERENITY_ROOT` can go away in this PR, it should just be global search and replace, nothing difficult. :)

panos:
> As mentioned on Discord, I think we can choose even more descriptive names for these while we're here. Something like `SERENITY_SOURCE_ROOT` and `SERENITY_INSTALL_ROOT`, but suggestions welcome. I also think `SERENITY_ROOT` can go away in this PR, it should just be global search and replace, nothing difficult. :)

Sorry, missed the e-mail. I feel like the latter should happen in another PR, though, as this is about to get messy.

panos:
Proposal:

- `DESTDIR` -> `SERENITY_INSTALL_ROOT`
- `SERENITY_ROOT` -> `SERENITY_PORTS_ROOT`, because the Ports folder is neither the entire source code directory or the `Root` folder as in `SERENITY_INSTALL_ROOT`.

linusg:
The current `SERENITY_ROOT` is not the ports directory though, but the the repository root (and it's used for various other things that are not ports!)

panos:
> The current `SERENITY_ROOT` is not the ports directory though, but the the repository root (and it's used for various other things that are not ports!)

Oh, then it's correct. WAIT, I completely forgot that this shell file was in the `Ports` folder and not inside of another port folder, I just got a bit mixed up. Yeah, I concur, `SERENITY_ROOT` seems just about fine.

panos:
So, let's sum this up:

- `DESTDIR` -> `SERENITY_INSTALL_ROOT`. `DESTDIR` will be kept around.
- `SERENITY_SOURCE` will point to the root folder of the repository. Let's keep the "root" terminology in the directory structure sort of sense out of here.
- `SERENITY_BUILD_DIR` will remain unchanged.
- `SERENITY_INSTALL_ROOT` will point to `$SERENITY_BUILD_DIR}/Root`.

linusg:
`SERENITY_SOURCE` -> `SERENITY_SOURCE_DIR`, but yes, this sounds good!

panos:
> `SERENITY_SOURCE` -> `SERENITY_SOURCE_DIR`, but yes, this sounds good!

True, it's important not to break the consistency.

linusg (commented):
cool - if you please could replace the two uses of `SERENITY_ROOT` in `SERENITY_BUILD_DIR` and `PATH` with `SERENITY_SOURCE_DIR` as well, I think we're good to go!

panos:
> cool - if you please could replace the two uses of `SERENITY_ROOT` in `SERENITY_BUILD_DIR` and `PATH` with `SERENITY_SOURCE_DIR` as well, I think we're good to go!

Thanks for the heads-up!

PR #6493, opened 2021-04-19
Userland: Add UDP send support to `nc` (netcat)

SpencerCDixon:


Userland/Utilities/nc.cpp
```diff
@@ -42,18 +45,48 @@ int main(int argc, char** argv)
...
     args_parser.add_option(verbose, "Log everything that's happening", "verbose", 'v');
+    args_parser.add_option(udp_mode, "UDP mode", "udp", 'u');
     args_parser.add_option(should_close, "Close connection after reading stdin to the end", nullptr, 'N');
     args_parser.add_positional_argument(addr, "Address to connect to or listen on", "address");
     args_parser.add_positional_argument(port, "Port to connect to or listen on", "port");
     args_parser.parse(argc, argv);
 
+    if (udp_mode) {
+        if (should_listen) {
+            fprintf(stderr, "listening on UDP not yet supported\n");
```
linusg, line 64: Please use `warnln()` instead, which also prints to stderr, automatically appends a newline and additionally supports `{}` formatting.

```suggestion
            warnln("Listening on UDP not yet supported");
```

Userland/Utilities/nc.cpp
```diff
@@ -42,18 +45,48 @@ int main(int argc, char** argv)
...
+            exit(1);
```
linusg, line 65:
```suggestion
            return 1;
```

(just for consistency, as the same is done below)

Userland/Utilities/nc.cpp
```diff
@@ -42,18 +45,48 @@ int main(int argc, char** argv)
...
+        }
+
+        Core::EventLoop loop;
+        auto socket = Core::UDPSocket::construct();
+
+        socket->on_connected = [&]() {
+            if (verbose)
+                fprintf(stderr, "connected to %s:%d\n", addr, port);
```
linusg, line 73:
```suggestion
                warnln("connected to {}:{}", addr, port);
```

you did the right thing since we already log to stderr below.... but any idea why? Can't this just go to stdout?

if you'd agree with that and maybe want to fix up the other calls as well:

```suggestion
                outln("connected to {}:{}", addr, port);
```

IdanHo: I've personally used nc in the form of `nc -l someport > out.file` to transfer files over the network, so that would interfere with that, but im not sure if thats intended functionality

linusg: hmm, that's a good point - let's keep it this way then.

SpencerCDixon: I'll add a comment to clarify since it is a bit confusing

SpencerCDixon:
@linusg I updated the branch with all your comments. Think it may be ready now? Going to be traveling for few days so sorry if it's not good to merge I can update any last fixes next Sunday.

PR #6494, opened 2021-04-19
LibC: Add aligned memory allocations APIs

RealKC:
I'm not sure if they're the best, but it's probably ok for now? I'd like to test them but I'm not entirely sure how I'd go about doing that.

### DIFF
--- a/Userland/Libraries/LibC/malloc.cpp
+++ b/Userland/Libraries/LibC/malloc.cpp
@@ -1,16 +1,19 @@
 /*
  * Copyright (c) 2018-2021, Andreas Kling <kling@serenityos.org>
+ * Copyright (c) 2021, Mițca Dumitru <dumitru0mitca@gmail.com>
  *
  * SPDX-License-Identifier: BSD-2-Clause
  */
 
 #include <AK/Debug.h>
 #include <AK/InlineLinkedList.h>
+#include <AK/Result.h>
 #include <AK/ScopedValueRollback.h>
 #include <AK/Vector.h>
 #include <LibELF/AuxiliaryVector.h>
 #include <LibThread/Lock.h>
 #include <assert.h>
+#include <malloc.h>
 #include <mallocdefs.h>
 #include <serenity.h>
 #include <stdio.h>
@@ -61,6 +64,7 @@ ALWAYS_INLINE static void ue_notify_realloc(const void* ptr, size_t size)
 
 struct MallocStats {
     size_t number_of_malloc_calls;
+    size_t number_of_aligned_alloc_calls;
 
     size_t number_of_big_allocator_hits;
     size_t number_of_big_allocator_purge_hits;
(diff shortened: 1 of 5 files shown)


awesomekling:
Can we find a way to implement this without duplicating so much code?
The classical way of doing this is making a wrapper that does a regular `malloc()` and then adds a little metadata header up front.
See how `kmalloc_aligned()` is implemented in the kernel, for example.

RealKC:
> The classical way of doing this is making a wrapper that does a regular malloc() and then adds a little metadata header up front.
> See how kmalloc_aligned() is implemented in the kernel, for example.

A kmalloc_aligned-like implementation doesn't really work as free() needs to be able to free pointers from both alloc_aligned() and posix_memalign(). Would an approach that makes malloc_impl() take into account alignment be a better idea then?

awesomekling:
> > The classical way of doing this is making a wrapper that does a regular malloc() and then adds a little metadata header up front.
> > See how kmalloc_aligned() is implemented in the kernel, for example.
> 
> A kmalloc_aligned-like implementation doesn't really work as free() needs to be able to free pointers from both alloc_aligned() and posix_memalign(). Would an approach that makes malloc_impl() take into account alignment be a better idea then?

Oh wow, I didn't realize that. What an unfortunate API constraint!

Then I think the best way forward would be sharing `malloc_impl()` somehow and giving it an alignment parameter yes. :)

Userland/Libraries/LibC/malloc.cpp
```diff
@@ -345,12 +380,45 @@ static void free_impl(void* ptr)
 
 [[gnu::flatten]] void* malloc(size_t size)
 {
-    void* ptr = malloc_impl(size, CallerWillInitializeMemory::No);
+    auto result = malloc_impl(size, CallerWillInitializeMemory::No);
+    if (s_profiling)
+        perf_event(PERF_EVENT_MALLOC, size, reinterpret_cast<FlatPtr>(result.value_or(nullptr)));
```
RealKC, line 385: Not sure I'm a fan of giving a bunch of PERF_EVENT_MALLOC for nullptr when we have errors, but I'm not a fan of ignoring those cases either, since allocations still take time, even if they fail. Not sure what to do tbh

RealKC:
@awesomekling wanna take another look :eyes:

Userland/Libraries/LibC/malloc.cpp
awesomekling: Let's not report a `PERF_EVENT_MALLOC` if the malloc fails. The purpose of logging the event is to record that memory allocation happened, after all. :)

Same comment for all other instances of the same thing.

awesomekling (changes requested):
The commit message says "quash me" :^)

RealKC:
> The commit message says "quash me" :^)

Fixed :P

Userland/Libraries/LibC/malloc.cpp
```diff
@@ -243,21 +277,27 @@ static void* malloc_impl(size_t size, CallerWillInitializeMemory caller_will_ini
...
     } else {
-        ptr = block->m_slot + block->m_next_lazy_freelist_index * block->m_size;
-        block->m_next_lazy_freelist_index++;
+        void* candidate_ptr = block->m_slot + block->m_next_lazy_freelist_index * block->m_size;
+        if ((reinterpret_cast<uintptr_t>(candidate_ptr) & (alignment - 1)) == 0) {
+            ptr = candidate_ptr;
+            block->m_next_lazy_freelist_index++;
+        }
     }
+
```
RealKC, line 292: @gunnarbeutner do the changes here seem reasonable to you when considering 343882943166abf4542a0e6643baf9d80806e944?

RealKC:
Changes since the last push:

- rebased on top of master
- edited a commit message, as I wasn't particularly pleased with the old one

RealKC:
So build should be fine now, but when booting I'm getting

```
[#0 SystemServer(6:6)]: fork: Cannot clone region, insufficient memory
```

not entirely sure if that's related to my patch, it could be I think, but I'm not seeing an obvious link

gunnarbeutner:
Try adding a `dbgln()` call in `os_alloc` and you'll see you're doing _way_ more allocations than the `master` branch does.

RealKC:
I'm closing this as I've gotten distracted by other projects and been a bit busy with school, and thus haven't had time to work on this

PR #6495, opened 2021-04-19
Piano: Optimize repaints

kleinesfilmroellchen:
The Piano application used to perform very poorly due to unnecessary
draw calls. This is solved with two optimziations:

1. Don't draw the widgets as often as possible. The widgets are instead
   at least updated every 150ms, except for other events.
2. Don't re-draw the entire piano roll sheet. The piano roll background,
   excluding in-motion objects (notes, the play cursor), is only re-drawn
   when its "viewport" changes.

A minor drawback of this change is that notes will appear on top of the
pitch labels if placed at the left edge of the roll. This is IMO
acceptable or may be changed by moving the text to the "foreground".

### DIFF
--- a/Userland/Applications/Piano/RollWidget.cpp
+++ b/Userland/Applications/Piano/RollWidget.cpp
@@ -1,6 +1,7 @@
 /*
  * Copyright (c) 2018-2020, Andreas Kling <kling@serenityos.org>
  * Copyright (c) 2019-2020, William McPherson <willmcpherson2@gmail.com>
+ * Copyright (c) 2021, kleines Filmröllchen <malu.bertsch@gmail.com>
  *
  * SPDX-License-Identifier: BSD-2-Clause
  */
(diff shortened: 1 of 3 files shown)


kleinesfilmroellchen:
I have to add that this was run on my medium-spec PC before I got virtualization (about 10x faster) to run. Current average CPU percent loads under high stress (complex ADSR, many notes and tracks at once) are about 6%-12%.

awesomekling (changes requested):
Please squash your commits per the [contribution guidelines](https://github.com/SerenityOS/serenity/blob/51fe25fdbbb7659e94557b0014b20242f2db408c/CONTRIBUTING.md)

kleinesfilmroellchen:
@awesomekling Commits are squashed, ready to merge now.

Dexesttp:
There's a couple of fixups needed on your commits :
- The actual syntax for piano commits seems to not include the `Userland:` prefix and instead would be `Piano: Optimize repaints` and `Piano: Add track volume` (according to the other commits in the log history)
- You should remove the "merge" commit from master, and only keep your two commits with actual changes. Merging the PR will handle this part, and if you leave it in then it'll be in the final history (which doesn't accept merges)

PR #6499, opened 2021-04-19
Ports: Remove auth_type for our own packages

gunnarbeutner:
Otherwise we'd have to update them in this repository every time something changes.

### DIFF
--- a/Meta/lint-ports.py
+++ b/Meta/lint-ports.py
@@ -81,6 +81,8 @@ def check_package_files(ports):
             data = fp.read()
             for p in properties:
                 if not re.findall(f"^{p}=", data, re.M):
+                    if p == 'auth_type' and re.findall('^files="?https://github.com/SerenityOS/', data, re.M):
+                        continue
                     print(f"Ports/{package} is missing '{p}'")
                     all_good = False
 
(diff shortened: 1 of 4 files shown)


gunnarbeutner:
Oh right, I need to update the lint checks.

gunnarbeutner:
> It does seem unlikely that the ports would be compromised without the main source tree also being compromised.

Aye, if we can't trust GitHub, then who can we trust?

bcoles:
> 
> 
> > It does seem unlikely that the ports would be compromised without the main source tree also being compromised.
> 
> Aye, if we can't trust GitHub, then who can we trust?

I know what you mean, but to clarify for historical purposes:

It's not so much as about trusting GitHub, it's about trusting a repository and everyone who has access to a repository. In this instance, using integrity hashes doesn't offer much value, as anyone who has write access to the ports likely also has write access to the SerenityOS repository.

If this assumption is wrong (does anyone have write access to the ports but not to the main source tree?) then integrity hashes *do* offer value and *should be used*.

panos:
Hashes are not only good for security, but they also help with corrupted downloads. However, they also tend to waste a lot of time.

linusg:
@gunnarbeutner please resolve conflicts :)

gunnarbeutner:
Done. :)

PR #6500, opened 2021-04-19
LibC: Avoid division by zero in fread()/fwrite()

gunnarbeutner:


dec05eba:
isn't this supposed to be if(!size) ?

gunnarbeutner:
You're absolutely right, of course. Want to submit a PR for that?

dec05eba:
I just realized that nwritten/nread will be 0 because read/write will return 0 if size is 0, i guess? so it works out ok anyways

PR #6503, opened 2021-04-20
Everywhere: Replace SERENITY_ROOT with SERENITY_SOURCE_DIR

panos:
I initially left SERENITY_ROOT in, but then I discovered the variable's presence in other places.

This *will deliberately not* break ports that are currently [in open pull requests](https://github.com/SerenityOS/serenity/pulls?q=is%3Apr+is%3Aopen+Ports%3A+Add), in order to give them a chance to be merged without friction. However, `SERENITY_ROOT` should be avoided in the future and subsequently replaced.


Ports/.hosted_defs.sh
```diff
@@ -10,8 +10,8 @@ export PKG_CONFIG_DIR=""
 export PKG_CONFIG_SYSROOT_DIR="${SERENITY_BUILD_DIR}/Root"
 export PKG_CONFIG_LIBDIR="${PKG_CONFIG_SYSROOT_DIR}/usr/lib/pkgconfig/:${PKG_CONFIG_SYSROOT_DIR}/usr/local/lib/pkgconfig"
 
-# To be removed.
-export SERENITY_ROOT="$(realpath "${SCRIPT}/../")"
+# To be deprecated soon.
+export SERENITY_SOURCE_DIR="$(realpath "${SCRIPT}/../")"
```
linusg, line 14: is this intentional? we already define that above and don't want to deprecate it :D

Userland/Utilities/test-js.cpp
```diff
@@ -744,9 +744,9 @@ int main(int argc, char** argv)
 #ifdef __serenity__
         test_root = "/home/anon/js-tests";
 #else
-        char* serenity_root = getenv("SERENITY_ROOT");
+        char* serenity_root = getenv("SERENITY_SOURCE_DIR");
```
linusg, line 747: Please also rename the variable to reflect this change

Userland/Utilities/test-web.cpp
```diff
@@ -676,9 +676,9 @@ int main(int argc, char** argv)
 #ifdef __serenity__
     TestRunner("/home/anon/web-tests", "/home/anon/js-tests", view, print_times).run();
 #else
-    char* serenity_root = getenv("SERENITY_ROOT");
+    char* serenity_root = getenv("SERENITY_SOURCE_DIR");
```
linusg, line 679: ditto

PR #6505, opened 2021-04-20
LibC: Add ctime_r() / asctime_r() / INET6_ADDRSTRLEN implementations + start LibC unit tests.

bgianfo:
These are some of the things I need for porting [fio](https://fio.readthedocs.io)

-   __LibC: Add ctime_r() and asctime_r() implementations__

-   __LibC: Setup a unit test harness for LibC, add tests.__

    LibC is no different than any other code, it should be unit tested where
    appropriate / possible. Add some ctime_r / asctime_r tests. 

-   __LibC: Add a definition for `INET6_ADDRSTRLEN`__

    Although serenity doesn't implement IPv6 yet, applications
    will often declare buffers to be of size `INET6_ADDRSTRLEN`
    so it's guaranteed to work in both IPv4 / IPv6 environments.

    This is standardized here:
    https://pubs.opengroup.org/onlinepubs/009695399/basedefs/netinet/in.h.html

### DIFF
--- a/Userland/DynamicLoader/CMakeLists.txt
+++ b/Userland/DynamicLoader/CMakeLists.txt
@@ -8,6 +8,8 @@ file(GLOB AK_SOURCES "../../AK/*.cpp")
 file(GLOB ELF_SOURCES "../Libraries/LibELF/*.cpp")
 file(GLOB LIBC_SOURCES1 "../Libraries/LibC/*.cpp")
 file(GLOB LIBC_SOURCES2 "../Libraries/LibC/*/*.cpp")
+file(GLOB LIBC_TEST_SOURCES "../Libraries/LibC/Tests/*.cpp")
+list(REMOVE_ITEM LIBC_SOURCES2 ${LIBC_TEST_SOURCES})
 
 if ("${SERENITY_ARCH}" STREQUAL "i686")
     file(GLOB LIBC_SOURCES3 "../Libraries/LibC/arch/i386/*.S")
--- a/Userland/Libraries/LibC/time.cpp
+++ b/Userland/Libraries/LibC/time.cpp
@@ -84,6 +84,12 @@ char* ctime(const time_t* t)
     return asctime(localtime(t));
 }
 
+char* ctime_r(const time_t* t, char* buf)
+{
+    struct tm tm_buf;
+    return asctime_r(localtime_r(t, &tm_buf), buf);
+}
+
 static const int __seconds_per_day = 60 * 60 * 24;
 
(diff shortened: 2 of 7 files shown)


Userland/DynamicLoader/CMakeLists.txt
```diff
@@ -8,6 +8,8 @@ file(GLOB AK_SOURCES "../../AK/*.cpp")
 file(GLOB ELF_SOURCES "../Libraries/LibELF/*.cpp")
 file(GLOB LIBC_SOURCES1 "../Libraries/LibC/*.cpp")
 file(GLOB LIBC_SOURCES2 "../Libraries/LibC/*/*.cpp")
+file(GLOB LIBC_TEST_SOURCES "../Libraries/LibC/Tests/*.cpp")
+list (REMOVE_ITEM LIBC_SOURCES2 ${LIBC_TEST_SOURCES})
```
linusg, line 12:
```suggestion
list(REMOVE_ITEM LIBC_SOURCES2 ${LIBC_TEST_SOURCES})
```

bgianfo: Thanks!

Userland/Libraries/LibC/time.cpp
```diff
@@ -180,7 +186,14 @@ struct tm* gmtime_r(const time_t* t, struct tm* tm)
...
     static char buffer[69];
-    strftime(buffer, sizeof buffer, "%a %b %e %T %Y", tm);
+    return asctime_r(tm, buffer);
+}
+
+char* asctime_r(const struct tm* tm, char* buffer)
+{
+    // Spec states buffer must be at least 26 bytes.
+    constexpr size_t assumed_len = 26;
+    strftime(buffer, assumed_len, "%a %b %e %T %Y", tm);
```
linusg, line 196: This looked suspicious, but it's indeed just a crappy API. I checked glibc and musl, and they have these comments which I enjoyed :smile:

https://github.com/bminor/glibc/blob/81f6dd2135ea761832965bc7518e2ddf949480af/time/asctime.c#L42-L48
https://github.com/ifduyue/musl/blob/cfdfd5ea3ce14c6abf7fb22a531f3d99518b5a1b/src/time/asctime_r.c#L16-L22

It seems this is missing a trailing newline though?

bgianfo: Nice spot! Yea looks like that was missing from the original implementation.

I added a commit which fixes this (crediting you), and updated the tests to verify it's presence.

PR #6508, opened 2021-04-20
LibGUI: Make statusbar label flat when displaying override_text

lpereira:
Changing the statusbar appearance when overriding text makes it less
confusing as it's supposed to be something temporary, e.g.  only when
hovering over a toolbar or menu item.

This behavior is present on old Windows systems, although things work
slightly differently there (where only the overridden text is displayed
rather than all the segments).

How it looks when a toolbar button is being hovered:
Screenshot from 2021-04-19 23-57-15


awesomekling:
Great idea! :smile:

PR #6511, opened 2021-04-20
LibJS: Automatically create numeric PropertyName from strings

linusg:
Relatively simple but possibly controversial (?) change, so leaving this up for review.


awesomekling:
This intuitively feels like it would have a strong negative effect on performance, as you now perform string-to-number conversion on every `PropertyName` construction.

linusg:
:snail:

PR #6513, opened 2021-04-20
LibCore+Literally Everywhere: Untangle the mess that is IODevice

alimpfard:
# Let's give `IODevice` the better semantics it deserves

Currently, `IODevice` is an absolute mess, it needs an fd, it has a buffer, it's a `Core::Object`...
You might ask, "is there anything this beast _cannot do_?" 
why yes, of course! it cannot have its `read()`/`write()` overridden, so there's no way to have "just" an IODevice!

## The new interface
- `Core::IODevice`'s behaviour is split into two abstract classes `Core::BufferingIODevice<T>` and `Core::FileLikeIODevice` which provide buffering capabilities for another IODevice, and keep track of (and define read/write) for IO devices using the normal `read()` and `write()` syscalls, respectively.
- `Core::IODevice` becomes an abstract interface, capable of only being a `DuplexStream` and providing a `Core::AbstractNotifier`
- `Core::Socket` is also split in half, with `Core::SocketLikeIODevice` being the abstract interface (providing `receive()`, `send()` and `connect()`) and `Core::Socket` itself defining that behaviour using the normal syscalls.
- `Core::Notifier` is also split in half, with `Core::AbstractNotifier` only providing the data members. This can be used to implement a notifier without an fd, for instance

Unfortunately, due to the virtual inheritance, a class inheriting from any of these has to manually construct all the underlying classes - I have not yet found a way to make a better way to do this.

## Current state

This is working, and ready for review :^)


Userland/Libraries/LibCore/FileLikeIODevice.cpp
```diff
@@ -0,0 +1,210 @@
...
+    int rc = fstat(fd(), &st);
+    if (rc == 0)
+        file_size = st.st_size;
+
+    Vector<u8> data;
+    data.ensure_capacity(file_size);
+
+    while (true) {
+        char read_buffer[4096];
+        int nread = ::read(m_fd, read_buffer, sizeof(read_buffer));
```
awesomekling, line 118: `auto` would be more appropriate here, as `::read` returns `ssize_t`, not `int`.

Userland/Libraries/LibCore/FileLikeIODevice.h
```diff
@@ -0,0 +1,95 @@
...
+
+class FileLikeIODevice : public virtual IODevice {
+    C_OBJECT_ABSTRACT(FileLikeIODevice)
+public:
+    // ^IODevice
+    size_t read(Bytes) override;
+    bool discard_or_error(size_t count) override;
+    bool unreliable_eof() const override { return m_eof; }
+    size_t write(ReadonlyBytes) override;
+    RefPtr<AbstractNotifier> make_notifier(unsigned event_mask) override;
```
awesomekling, line 45: Missing `virtual` ([coding style](https://github.com/SerenityOS/serenity/blob/master/Documentation/CodingStyle.md#overriding-virtual-methods))

Userland/Libraries/LibCore/Notifier.h
```diff
@@ -11,8 +11,8 @@
 
 namespace Core {
 
-class Notifier : public Object {
-    C_OBJECT(Notifier)
+class AbstractNotifier : public Object {
+    C_OBJECT_ABSTRACT(AbstractNotifier)
```
awesomekling, line 15: Let's put an `;` at the end of the line for consistency with the rest of the codebase.

Userland/Libraries/LibCore/SocketLikeIODevice.h
```diff
@@ -0,0 +1,31 @@
...
+ */
+
+#pragma once
+
+#include <LibCore/IODevice.h>
+
+namespace Core {
+
+class SocketLikeIODevice : public virtual IODevice {
+    C_OBJECT_ABSTRACT(SocketLikeIODevice)
```
awesomekling, line 14: Let's put an `;` at the end of the line for consistency with the rest of the codebase.

Userland/Libraries/LibCore/FileLikeIODevice.h
```diff
@@ -0,0 +1,95 @@
...
+enum class SeekMode {
+    SetPosition,
+    FromCurrentPosition,
+    FromEndPosition,
+};
+
+AK_ENUM_BITWISE_OPERATORS(OpenMode)
+
+class FileLikeIODevice : public virtual IODevice {
+    C_OBJECT_ABSTRACT(FileLikeIODevice)
```
awesomekling, line 38: Let's put an `;` at the end of the line for consistency with the rest of the codebase.

Userland/Libraries/LibCore/BufferingIODevice.h
```diff
@@ -0,0 +1,196 @@
...
+
+private:
+    NonnullRefPtr<BufferingIODevice<UnderlyingDeviceT>> m_device;
+    bool m_is_end { false };
+    String m_buffer;
+};
+
+template<typename UnderlyingDeviceT>
+class BufferingIODevice : public UnderlyingDeviceT {
+    C_OBJECT(BufferingIODevice)
```
awesomekling, line 43: Let's put an `;` at the end of the line for consistency with the rest of the codebase.

awesomekling (changes requested):

awesomekling:
Reading this makes me think none of these should be `Core::Object`.

Userland/Libraries/LibCore/FileLikeIODevice.cpp
alimpfard: I have a feeling that this came from existing code, but yeah, `auto` makes more sense.

Userland/Libraries/LibCore/FileLikeIODevice.h
alimpfard: Blame clion, it annoyed me with "hurr durr `virtual` is redundant"

alimpfard:
> Reading this makes me think none of these should be `Core::Object`

Well at least `Core::Socket` and `Core::File` will have to be, but I think not making IODevice itself a `Core::Object` will also resolve the multiple base ctors issue, will try.

alimpfard:
Okay, this turned out to be a big yak shaving thing, so marking as draft until that's resolved.

alimpfard:
Regarding not making `Core::IODevice` a `Core::Object`, some users use it to do `deferred_invoke`, decoupling `deferred_invoke` from `Core::Object` changes lifetime semantics (`Core::EventLoop` held a reference to the receiver while calling the invokee before).
I'm not gonna fix _that_, so `IODevice` is gonna remain broken for now.

PR #6515, opened 2021-04-20
LibGUI: Add selection_start_index to AbstractView

iCristalrope:
Modifying the selection while holding the shift button and selecting
with the mouse or the arrow keys no longer result in broken selections.

Closes SerenityOS#6279.


awesomekling (changes requested):
Since you are caching a `ModelIndex`, you'll have to take care to clear/forget it in `AbstractView::model_did_update()` as well.

iCristalrope:
As I discovered this code yesterday and this is my first contribution, I'm not very familiar with it yet. 

I'm not sure what I changed is what you asked for. I simply default initialized `m_selection_first_index` again like the other `ModelIndex` objects next to it.

PR #6519, opened 2021-04-20
LibAudio: Add MP3 decoder for Layer III

g0mb4:
Always wanted to learn about MP3 decodig, so this was a perfect opportunity.
It's currently crashes at the end of the song. The fix would be to detect the end of the file/stream, but I was unable to do it, because of the lack of my knowledge of the API of Serenity :(
The decoder is integrated into LibAudio, so a possible usage is:
```aplay <your-favourite-mp3>```

### CODE REVIEW

**Comment by IdanHo** [comment]
Please follow the [code submission policy](https://github.com/SerenityOS/serenity/blob/master/CONTRIBUTING.md#code-submission-policy), for example:

- The first line of the commit message should have the format "Category: Brief description of what's being changed"
- Squash your commits when making revisions

File: Base/home/anon/.shellrc [hunk 1]
```diff
@@ -1,7 +0,0 @@
-#!/bin/sh
-
-# IFS controls what $(...) (inline evaluate) would split its captured
-# string with. the default is \x0a (i.e. newline).
-IFS="\x0a"
-
-export HISTCONTROL="ignoreboth"
```
IdanHo, line 7: Why are you removing the shellrc file?

g0mb4: I created the fork about three weeks ago, the .shellrc file was not part of my fork :/

IdanHo: you should rebase your changes on latest master then :)

g0mb4: Ohh, ok, thanks for the tip, I'll try to do that.

ngc6302h:
We must be careful when basing implementations on existing work. The reference implementation for this pull request seems to be https://github.com/FlorisCreyf/mp3-decoder which uses the GPL2 license, which as far as I know, it's incompatile with SerenityOS's BSD license. I'm by no means a legal expert but I guess we could have trouble if the code is too similar. 

I'm also working on an MP1/2/3 decoder (you win though!), although it's not functional yet, and am basing it on https://github.com/lieff/minimp3, which is released under the Creative Commons public-domain-like license. It might be useful to you to refer to this implementation since it's fully compatible with the BSD license.

g0mb4:
You are probably right, it's better be on the safe side. I used the ISO standard and multiple sources, but the the data decoding part is very similar as the code it's based on. I just wanted a working version.
I think it's not a competition, it's just fun, we live a rather competitive world anyways, so I would be really happy, if you could help, because I'm running out of free time :)
I will fix my silly bugs in order to compile, but I'm not sure if I'll have time to rewrite the code :/

PR #6521, opened 2021-04-20
Shell: Auto-completion shouldn't suggest non-executable files for the program name

gunnarbeutner:
Right now the `Shell` suggests `README.md` when trying to auto-complete `./` in the home directory. This file isn't executable though.

### DIFF
--- a/Userland/Shell/Shell.cpp
+++ b/Userland/Shell/Shell.cpp
@@ -1304,7 +1304,8 @@ Vector<Line::CompletionSuggestion> Shell::complete()
     return ast->complete_for_editor(*this, line.length());
 }
 
-Vector<Line::CompletionSuggestion> Shell::complete_path(const String& base, const String& part, size_t offset)
+Vector<Line::CompletionSuggestion> Shell::complete_path(const String& base,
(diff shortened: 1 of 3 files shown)


Userland/Shell/Shell.cpp
```diff
@@ -1358,7 +1359,7 @@ Vector<Line::CompletionSuggestion> Shell::complete_path(const String& base, cons
             struct stat program_status;
             String file_path = String::format("%s/%s", path.characters(), file.characters());
             int stat_error = stat(file_path.characters(), &program_status);
-            if (!stat_error) {
+            if (!stat_error && (executable_only == ExecutableOnly::No || access(file_path.characters(), X_OK) == 0)) {
                 if (S_ISDIR(program_status.st_mode)) {
```
alimpfard, line 1363: Won't this stop suggesting directories?
It's often expected to have the shell complete a directory name no matter what.

gunnarbeutner: Nah, that'll still work because directories are "executable".

alimpfard: Alright then : )

PR #6522, opened 2021-04-20
Inspector: Select entire property row and add Copy Property Name/Value

DragonAlex98:
With some themes (like the default one), it was possible to select a
property, making the text of its value not visible. I solved this by
setting set_should_fill_selected_rows to true.

I also added a context menu for the property tree view to copy the
name/value of a property.

Before
before

After
after


linusg:
Please split this into two commits, this should be a change on its own:

> With some themes (like the default one), it was possible to select a property, making the text of its value not visible. I solved this by setting set_should_fill_selected_rows to true.

PR #6524, opened 2021-04-21
LibGfx: Implement bi-directional, floating-point scaling support

mattco98:
As part of my ongoing SVG quest, this PR will allow me to implement support for the SVG viewBox, which is an extremely prevalent property on most SVGs. 

This PR makes a bunch of changes in support of the Painter's new `scale(float, float)` method, which allows arbitrary scaling (both up and down) in both directions separately. Unfortunately, I'm not quite sure how to handle Bitmaps. I did add a similar scale method to `Gfx::Bitmap` (which uses bi-linear interpolation), however it is unused at the moment. 

The main issue is that if you pass in a normal bitmap to the painter every frame, and the painter has scaling, the bitmap can't be scaled every frame as it is an expensive operation (especially if the scaling in non-integer). My idea was to have a cache of (bitmap & scale) -> (scaled bitmap), but I'd like to hear some more opinions.

Also, the [HighDPI](https://github.com/mattco98/serenity/blob/painter-scale/Documentation/HighDPI.md) design document mentions possibly introducing a container class for Bitmap which would store multiple lazily-initialized bitmaps at different scales. It is only concerned with integer scaling, but such a container could be used for floating point scaling as well. 

~TODO:~
- [ ] ~Handle scaling everywhere (pretty much just where bitmaps are handled)~
- [ ] ~Remove useless methods (like making `scale()` do what `float_scale()` does, remove `has_integer_scaling()`, etc)~

### DIFF
--- a/Userland/Applications/Piano/KeysWidget.cpp
+++ b/Userland/Applications/Piano/KeysWidget.cpp
@@ -242,7 +242,7 @@ int KeysWidget::note_for_event_position(const Gfx::IntPoint& a_point) const
         return -1;
 
     auto point = a_point;
-    point.move_by(-frame_thickness(), -frame_thickness());
+    point.translate_by(-frame_thickness(), -frame_thickness());
 
     int white_keys = point.x() / white_key_width;
     int note = note_from_white_keys(white_keys);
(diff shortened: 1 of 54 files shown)


mattco98:
I've tried to brainstorm some ideas for Bitmap scaling support from the painter, but I don't really have a good solution. This PR is a bit of a blocker for some other things I want to do, so I'd really like to get it merged. Bitmap scaling support can come later... I shouldn't try to do too much in this PR anyways

PR #6525, opened 2021-04-21
Userland: Create separate directory for BSD reimplementations

panos:
- Keeping things separate will allow supporting a wide range of software, without effectively turning Serenity kernel into a BSD kernel.
- sysexits.h is a reimplementation of FreeBSD's functionality (also found in other BSD-like systems for compatibility reasons) and is not a reimplementation of LibC.

## Additional notes

- [Haiku](https://haiku-os.org) puts BSD-related headers and code in a different directory, as demonstrated [here](https://github.com/haiku/haiku/tree/d0e006f5569f24e73a4d83f5c9766f1132160d7a/src/libs/bsd).
- This change was primarily caused by my confusion as to where I should write an `err.c`/`err.h` reimplementation.
- I believe that dealing with this now will divert more focused efforts on reimplementing certain parts of the BSD kernel for software compatibility reasons much more easily.

I'm very skeptical as to whether this will work, and mostly made the pull request in order to ensure that I'm doing it right and initiate a conversation. I am also not sure whether `LibBSD` is a correct name, considering that there isn't such a library, so I decided to follow Haiku's approach and name the directory `BSD`. I additionally thought of `LibCompat/BSD` or `Compatibility/BSD`, but I'm afraid that this could very easily complicate things.

**EDIT:** Settled for `LibBSD`.


panos:
Random fun fact: Berkeley's Razor, which is the theory that science only gives explanations for certain phenomena and cannot give us true insight in how nature truly works as a whole, was devised by Karl Popper and not by George Berkeley.

panos:
Jinxed it, let's try again...

panos:
After studying the source code a bit further and stumbling upon `LibX86`, which isn't bound to a specific application in comparison to `LibHTTP`, I take it that target_link_libraries just combines different libraries together in a seamless manner, so that I don't have to use `#include <LibBSD/sysexits.h>` in source files (which would require a lot of manual patching later down the line), right?

That's what I'm trying to do, at least.

awesomekling:
I don't see why this needs to be a separate library. Sure, we implement some BSD and Linux extensions in our LibC, but it doesn't make sense to me to separate them at this level.

Instead, let's just be a bit more organized in LibC, perhaps something like `Userland/Libraries/LibC/Compat/BSD/` and `Userland/Libraries/LibC/Compat/Linux/`

panos:
> I don't see why this needs to be a separate library. Sure, we implement some BSD and Linux extensions in our LibC, but it doesn't make sense to me to separate them at this level.
>
> Instead, let's just be a bit more organized in LibC, perhaps something like `Userland/Libraries/LibC/Compat/BSD/` and `Userland/Libraries/LibC/Compat/Linux/`

I believe that they need to be separated one way or another, because if everything is implemented on the kernel level and you don't have a separate compatibility layer, the kernel is going to be turned either into a) a huge BSD reimplementation, b) a convoluted mess that cannot decide where it's Serenity or BSD. By all means, the source files have to be separate, and if a program wants to check `IF_RUNNING` or some other FreeBSD-specific interface status, the compatibility layer should directly ask the kernel about it and get a response, then translate the response in the FreeBSD format that applications understand, rather than just, *be* the kernel.

So... are you proposing that I do what I did but inside of the `LibC` directory? Functionality-wise, it should be fine and not change much, but I'm not sure how well that'll work symbolically. I envision something like a "beast of its own" rather than a header file, like, something that greatly improves the functionality of Serenity, but stands out as a mere extension that could be taken in or out of the system at any time, and the base would still work. Not just that counterpart of LibC or copying and pasting source files directly from a BSD kernel. (I'd understand that in the case of [GNU](https://github.com/haiku/haiku/tree/d0e006f5569f24e73a4d83f5c9766f1132160d7a/headers/compatibility/gnu), however.)

Design-wise, in order to get applications to run easily, the header files need to be present under `${SERENITY_INSTALL_DIR}/usr/include`, and not under a subfolder. It's going to be exceptionally hard to maintain ports or convince maintainers to upstream Serenity-specific patches if an included header, say `err.h`, looks like `#include <Compat/BSD/err.h>`. So that's probably out of the question. You'd have to follow a similar approach as the one applied in my own change and to the rest of the `LibC` directory, where CMake is instructed to put the headers in the root folder (`${SERENITY_INSTALL_DIR}/usr/include`). It sure as heck is possible, but there's the possibility of having to add an additional directory, e.g. `Example` with the headerfile `foobar.h`, under `Userland/Libraries/LibC` other than `Compat`, possibly both inside `LibC` (`#include <Example/foobar.h>`) and in other library implementations (`#include <LibJS/Example/foobar.h`). I think that this sort of inconsistency could ruin organization in the long run.

As far as having to be close-to-compatible with a system, I think that a choice will have to be made as to whether the system will be compatible with Linux **or** with BSD, or make that decision and then pull a FreeBSD and support Linux binaries too, although I think that's way out of reach. I don't think that compatibility with Linux will be sustainable in the long run.

awesomekling:
> > I don't see why this needs to be a separate library. Sure, we implement some BSD and Linux extensions in our LibC, but it doesn't make sense to me to separate them at this level.
> > Instead, let's just be a bit more organized in LibC, perhaps something like `Userland/Libraries/LibC/Compat/BSD/` and `Userland/Libraries/LibC/Compat/Linux/`
> 
> I believe that they need to be separated one way or another, because if everything is implemented on the kernel level and you don't have a separate compatibility layer, the kernel is going to be turned either into a) a huge BSD reimplementation, b) a convoluted mess that cannot decide where it's Serenity or BSD.

No need to exaggerate. The kernel is not going to become "a huge BSD reimplementation" because we have a handful of functions and macros in LibC.

> By all means, the source files have to be separate, and if a program wants to check `IF_RUNNING` or some other FreeBSD-specific interface status, the compatibility layer should directly ask the kernel about it and get a response, then translate the response in the FreeBSD format that applications understand, rather than just, _be_ the kernel.

What specific BSD extension are you proposing to implement that would require such a translation layer? This looks like a classic "solution in search of a problem" to me.
 
> Design-wise, in order to get applications to run easily, the header files need to be present under `${SERENITY_INSTALL_DIR}/usr/include`, and not under a subfolder.

Yes, just like we already do for `<pthread.h>` and `<math.h>`.

> As far as having to be close-to-compatible with a system, I think that a choice will have to be made as to whether the system will be compatible with Linux **or** with BSD, or make that decision and then pull a FreeBSD and support Linux binaries too, although I think that's way out of reach. I don't think that compatibility with Linux will be sustainable in the long run.

It is not a goal to be compatible with Linux or FreeBSD.

panos:
> No need to exaggerate. The kernel is not going to become "a huge BSD reimplementation" because we have a handful of functions and macros in LibC.

> What specific BSD extension are you proposing to implement that would require such a translation layer? This looks like a classic "solution in search of a problem" to me.

Yeah, but like, is every single definition that reports the status of the system, such as `IF_RUNNING` (excuse me for being so stuck on it), going to be directly announced by the kernel itself as if it were a BSD system, is the kernel going to do its own thing that's going to be processed by all ports afterwards, etc. There might be something that I'm missing!

I just believe that this could be a slippery slope, excuse me if I overdid it by trying to present my thoughts in such an amplified manner. I made a change that implemented a specific header file, without the source code or anything, #6447, and every approach involved sledgehammering the code in the kernel itself, so that BSD programs, such as the Tor daemon, could work better. I'm really skeptical as to whether the approach that I planned to take in order to increase the capabilities of the system were correct and whether it wouldn't conflict with the goals of the project later on. Sure, that may not apply to headers such as `err.h`, but it could apply to [`sys/statvfs.h`](https://man7.org/linux/man-pages/man3/statvfs.3.html), which I also stumbled upon and tried to figure out as to how it could fit in the system.
 
> > By all means, the source files have to be separate, and if a program wants to check `IF_RUNNING` or some other FreeBSD-specific interface status, the compatibility layer should directly ask the kernel about it and get a response, then translate the response in the FreeBSD format that applications understand, rather than just, _be_ the kernel.

> > Design-wise, in order to get applications to run easily, the header files need to be present under `${SERENITY_INSTALL_DIR}/usr/include`, and not under a subfolder.
> 
> Yes, just like we already do for `<pthread.h>` and `<math.h>`.

Yeah, that's why I tried to follow a similar approach with the `LibBerkeley` directory. But if I were to bring up `sys/stavfs.h` again, by the same logic, it'd be under `Userland/Libraries/LibC/Compat/BSD/sys/statvfs`, alongside the contents of the directory `Userland/Libraries/LibC`, `/Libraries/LibPthread/pthread.h` and `/Libraries/LibM/math.h`. It makes sense to me to make an exception here and put them under the root of `/usr/include` so that apps work properly while overriding the general pattern that the system seems to follow. So there are two exceptions so far, where the headers (down a single level) are directly copied onto the root folder for software compatibility reasons. But then, you have the subfolder of a subfolder of a subfolder (`LibC/Compat/BSD/sys/stavfs.h` -> `sys/stavfs.h`). And it all gets copied back onto the root, just like the contents of the subfolder where the rest is also present, whereas the POSIX Threads header file is not present either. It's confusing. (I spent 5 minutes trying to explain this scenario and unironically got a headache.)
 
> It is not a goal to be compatible with Linux or FreeBSD.

I just commented on the "perhaps something like `Userland/Libraries/LibC/Compat/BSD/` and `Userland/Libraries/LibC/Compat/Linux/`", but I'm not sure if I caused a misunderstanding here; I'm talking about the system being able to compile modular applications designed primarily for those platforms mostly out of the box and not binary compatibility as in the thing that FreeBSD does with Linux.

panos:
Huge apologies for taking up way too much time of your with this, it's just that I feel like that this could prove to be a problem at some point later on as a result of what ***I*** came up with in order to deal with a series of very specific problems that could end up costing either a lot of time to revert or result in high maintenance costs, then generalized the said problems that I faced, found a few more examples after reviewing some source code and trying to see where it would fit that I faced and started drawing flowcharts where I tried to make BSD definitions stick in parts of Serenity's source code (as if I were the Zodiac killer or something) and kept seeing that sort of pattern that told me that something was off that I tried to explain as well as I could.

panos:
Having thought through it a bit more, if I go by the logic of "other people do it, so why not us", it wouldn't make sense to go through with this and keeping header files that serve as extensions of the C programming language in the systems I'm seeking to support by default, rather than be separate (such as in FreeBSD's case).

Thanks for your time and attention.

PR #6527, opened 2021-04-21
LibC + LibPthread: Add pthread_attr_[set|get]scope, pwrite(..) and LOG_NOWAIT implementations. 

bgianfo:
More changes needed for porting: https://fio.readthedocs.io

-  LibPthread: Add non functional pthread_attr_[set|get]scope stubs

    Standard: https://pubs.opengroup.org/onlinepubs/009695399/functions/pthread_attr_getscope.html

-  LibC: Add pwrite(..) implementation to match the existing pread(..)

    Add a basic non-thread safe implementation of pwrite, following the
    existing pread(..) design.

-  LibC: Add LOG_NOWAIT stub to syslog.h

### DIFF
--- a/Userland/Libraries/LibC/unistd.cpp
+++ b/Userland/Libraries/LibC/unistd.cpp
@@ -289,10 +289,30 @@ ssize_t read(int fd, void* buf, size_t count)
     __RETURN_WITH_ERRNO(rc, rc, -1);
 }
 
+ssize_t pread(int fd, void* buf, size_t count, off_t offset)
+{
+    // FIXME: This is not thread safe and should be implemented in the kernel instead.
+    off_t old_offset = lseek(fd, 0, SEEK_CUR);
+    lseek(fd, offset, SEEK_SET);
+    ssize_t nread = read(fd, buf, count);
+    lseek(fd, old_offset, SEEK_SET);
+    return nread;
+}
+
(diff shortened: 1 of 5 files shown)


Userland/Libraries/LibC/unistd.cpp
```diff
@@ -289,12 +289,32 @@ ssize_t read(int fd, void* buf, size_t count)
...
 }
 
+ssize_t pwrite(int fd, const void* buf, size_t count, off_t offset)
+{
+    // FIXME: This is not thread safe and should be implemented in the kernel instead.
+    off_t old_offset = lseek(fd, 0, SEEK_CUR);
+    lseek(fd, offset, SEEK_SET);
+    ssize_t nread = write(fd, buf, count);
+    lseek(fd, old_offset, SEEK_SET);
+    return nread;
```
linusg, line 315:
```suggestion
    ssize_t nwrite = write(fd, buf, count);
    lseek(fd, old_offset, SEEK_SET);
    return nwrite;
```

gunnarbeutner: Shouldn't that be `nwritten` though?

linusg: that works too :)

bgianfo: Thanks!

PR #6529, opened 2021-04-21
Meta: Precompile some very common headers 

alimpfard:
You guys just had to go and make the build times slower (it has regressed by about 30 seconds since I last measured it in the compiletime format string check PR).

Times measured from cold ccache, fresh rebuild from idle (same as last time)
|                                           | Time | Difference from baseline |
| :---                                     | :---:   | :---: |
| Baseline (https://github.com/SerenityOS/serenity/pull/5468#issuecomment-783982451) | 452s | 0s |
| Current master (2ebb3f3) | 483s | +31s |
| This PR                            | 441s | -11s |

Is this a substantial amount? Not really.
Downsides:
- if the guess as to what header is "common" enough is wrong, a change in said header will force (basically) the entire codebase to be rebuilt.
    To avoid this, really be careful with including things in `Precompile/AK.h`!
- Nothing else, it's basically free time savings

---

Note that the "common" headers are just found through a simple `for file in $(find Kernel/ AK/ Userland/ -name '*.h' -o -name '*.cpp') { grep '^#include <AK' $file >> all-includes }`.
I picked the most common files that made sense to precompile (e.g. no TestSuite.h and no Debug.h)

---

Until we get the goodness that C++ modules are supposed to be, let's try to shave off some parse time using precompiled headers.
This commit only adds some very common AK headers, only to binaries, libraries and the kernel (tests are not covered due to incompatibility with AK/TestSuite.h).
This option is on by default, but can be disabled by passing `-DPRECOMPILE_COMMON_HEADERS=OFF` to cmake, which will disable all header precompilations.
This makes the build about 40 seconds faster on my machine (about 9%)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -26,6 +26,7 @@ option(ENABLE_ALL_THE_DEBUG_MACROS "Enable all debug macros to validate they sti
 option(ENABLE_COMPILETIME_FORMAT_CHECK "Enable compiletime format string checks" ON)
 option(ENABLE_PCI_IDS_DOWNLOAD "Enable download of the pci.ids database at build time" ON)
 option(BUILD_LAGOM "Build parts of the system targeting the host OS for fuzzing/testing" OFF)
+option(PRECOMPILE_COMMON_HEADERS "Precompile some common headers to speedup compilation" ON)
 
 add_custom_target(run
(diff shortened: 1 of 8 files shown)


linusg:
Neat! :rocket:

PR #6530, opened 2021-04-21
Meta: Add GDB pretty printers

gunnarbeutner:
This adds support for GDB pretty printers and uses them by default for `Meta/serenity.sh gdb` and `Meta/debug-kernel.sh`.

TODO:

- [x] Remove `-ggdb3` or make it configurable.

Screenshot from 2021-04-21 15-10-47

### DIFF
--- a/Meta/serenity_gdb.py
+++ b/Meta/serenity_gdb.py
@@ -1,0 +1,274 @@
+# Copyright (c) 2021, Gunnar Beutner <gunnar@beutner.name>
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are met:
+#
+# 1. Redistributions of source code must retain the above copyright notice, this
+#    list of conditions and the following disclaimer.
+#
+# 2. Redistributions in binary form must reproduce the above copyright notice,
+#    this list of conditions and the following disclaimer in the documentation
+#    and/or other materials provided with the distribution.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+# AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+# IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+# DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+# FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+# DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+# SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+# CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+# OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+import gdb
+import gdb.types
+import re
+
+
+class AKAtomic:
+    def __init__(self, val):
+        self.val = val
+
+    def to_string(self):
+        return self.val["m_value"]
+
+
+class AKDistinctNumeric:
+    def __init__(self, val):
+        self.val = val
+
+    def to_string(self):
+        return self.val["m_value"]
+
+
+class AKRefCounted:
+    def __init__(self, val):
+        self.val = val
+
+    def to_string(self):
+        return self.val["m_ref_count"]
+
+
+class AKString:
+    def __init__(self, val):
+        self.val = val
+
+    def to_string(self):
+        if int(self.val["m_impl"]["m_bits"]["m_value"]) == 0:
+            return '""'
+        else:
+            impl = AKRefPtr(self.val["m_impl"]).get_pointee().dereference()
+            return AKStringImpl(impl).to_string()
+
+
+class AKStringView:
+    def __init__(self, val):
+        self.val = val
+
+    def to_string(self):
+        if int(self.val["m_length"]) == 0:
+            return '""'
+        else:
+            characters = self.val["m_characters"]
+            str_type = characters.type.target().array(self.val["m_length"]).pointer()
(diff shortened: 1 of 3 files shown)


Meta/serenity_gdb.py
```diff
@@ -0,0 +1,263 @@
...
+# CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+# OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+import gdb
+import gdb.types
+import re
+
+
+class AKAtomic(object):
```
linusg, line 30: new style classes have been the default in Python for a long time, none of these need to explicitly inherit from object.

Meta/serenity_gdb.py
```diff
@@ -0,0 +1,263 @@
...
+        return self.val["m_ref_count"]
+
+
+class AKString(object):
+    def __init__(self, val):
+        self.val = val
+
+    def to_string(self):
+        if int(self.val["m_impl"]["m_bits"]["m_value"]) == 0:
+            return "\"\""
```
linusg, line 60:
```suggestion
            return '""'
```

Same below.

Meta/serenity_gdb.py
```diff
@@ -0,0 +1,263 @@
...
+            characters = self.val["m_characters"]
+            str_type = characters.type.target().array(self.val["m_length"]).pointer()
+            return str(characters.cast(str_type).dereference())
+
+
+def get_field_unalloced(val, member, type):
+    # Trying to access a variable-length field seems to fail with
+    # Python Exception <class 'gdb.error'> value requires 4294967296 bytes, which is more than max-value-size
+    # This works around that issue.
+    return gdb.parse_and_eval("*(%s*)((%s*)%s)->%s" % (type, val.type.name, int(val.address), member))
```
linusg, line 83: We can assume modern Python >= 3.6 (3.5 and below had EOL already), please use f-strings. Same below.

linusg (changes requested):

Meta/serenity_gdb.py
```diff
@@ -0,0 +1,263 @@
...
+        elif klass == 'AK::Vector':
+            return AKVector(val)
+        elif klass == 'VirtualAddress':
+            return VirtualAddress(val)
+
+
+gdb.printing.register_pretty_printer(None, SerenityPrettyPrinterLocator(), replace=True)
+
+
+class FindThreadCmd(gdb.Command):
```
bgianfo, line 235: Can you write python help text, so it will show up when someone asks for help of this command in gdb? 

Reference: https://sourceware.org/gdb/onlinedocs/gdb/Commands-In-Python.html

> The help text for the new command is taken from the Python documentation string for the command’s class, if there is one. If no documentation string is provided, the default value “This command is not documented.” is used.

bgianfo (changes requested):
Looks good in general, just a few small nits from me.

gunnarbeutner:
Docs are still missing. Will add those in a bit. Thanks for the reviews!

Kernel/CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-add_compile_options(-Os)
+add_compile_options(-Os -ggdb3)
```
bgianfo, line 1: Just a reminder, still  probably need to make this conditional based on; 

```
commit a731ccd4a0bd55dd1a7a320c4c6e9bc383cb2618
Author: Andreas Kling <kling@serenityos.org>
Date:   Sat Feb 22 18:39:22 2020 +0100

    Kernel: Build without debugging symbols by default

    Compiling with -g adds roughly 30% to kernel build times. Anyone who
    wants this can turn it on locally instead.

commit 46c60fd451e478ee0bc834de7ff129d85d437bb3
Author: Brian Gianforcaro <b.gianfo@gmail.com>
Date:   Sun Jan 12 22:50:47 2020 -0800

    Debugging: Add kernel debugging support

    Introduce the 'debug-kernel' script to allow developers to
    quickly attach a debugger to the QEMU debug remote. The
    setting (-s) is already enabled by ./run today when using
    QEMU for virtualisation.

    If the system is running under QEMU, the debugger
    will break in when the script is run. If you add
    the -S option to QEMU it will wait for the debugger
    to connect before booting the kernel. This allows
    you to debug the init/boot process.

    Personally I use cgdb instead of gdb, so I opted
    to make the debugger used by the script customizable
    via an environment variable.

    This change also adds -g3 to the kernel build so that
    rich debug symbols are available in the kernel binary.
```

gunnarbeutner: Ah yes, I thought I had removed that for now. Turns out I re-added it to test my other changes. :+1:

gunnarbeutner:
* Removed `-ggdb3`.
* Added docs:

```
(gdb) help find_thread

    Find SerenityOS thread for the specified TID.
    find_thread TID
    
(gdb) 
```

Kernel/CMakeLists.txt
bgianfo: It would be interesting to know how much of a regression this really is these days, considering when @awesomekling  reverted the change we weren't even compiling with optimizations enabled and were probably using a 9 series gcc.

bgianfo (approved):

Meta/serenity_gdb.py
```diff
@@ -0,0 +1,274 @@
...
+        outline_buf = self.val["m_outline_buffer"]
+
+        inner_type_ptr = self.val.type.template_argument(0).pointer()
+
+        if int(outline_buf) != 0:
+            elements = outline_buf.cast(inner_type_ptr)
+        else:
+            elements = get_field_unalloced(self.val, "m_inline_buffer_storage", inner_type_ptr)
+
+        return [("[%d]" % (i,), elements[i]) for i in range(vec_len)]
```
linusg, line 147:
```suggestion
        return [(f"[{i}]", elements[i]) for i in range(vec_len)]
```

Kernel/CMakeLists.txt
bgianfo: Here's the current numbers from my machine.

Without;  
```
Without: 
$ time ninja Kernel
[259/259] Linking CXX executable Kernel/Kernel

real	0m47.773s
user	5m57.104s
sys	0m11.670s
```
With `-gddb3`:
```
$ time ninja Kernel
[259/259] Linking CXX executable Kernel/Kernel

real	1m2.625s
user	7m44.211s
sys	0m16.173s
```

linusg (commented):
Very nice how gdb integrates with Python for pretty-printing - didn't know this was possible :)

Kernel/CMakeLists.txt
gunnarbeutner: Sounds like we'd probably want a cmake option for it. :)

Meta/serenity_gdb.py
gunnarbeutner: Oops, fixed.

PR #6534, opened 2021-04-21
LibWeb: Implement scientific notation parsing in PathDataParser

IdanHo:
This was required by webkit.org and is based on the following spec: https://svgwg.org/svg2-draft/paths.html#PathDataBNF

webkit.org still renders as mostly a white page with a bit of text, since almost none of the css gets parsed correctly, but at least it doesnt crash the browser now :)

### DIFF
--- a/Userland/Libraries/LibWeb/SVG/SVGPathElement.cpp
+++ b/Userland/Libraries/LibWeb/SVG/SVGPathElement.cpp
@@ -390,8 +390,34 @@ float PathDataParser::parse_fractional_constant()
 float PathDataParser::parse_number()
 {
     auto number = parse_fractional_constant();
-    if (match('e') || match('E'))
-        TODO();
+
+    if (!match('e') && !match('E'))
+        return number;
+    consume();
+
+    auto exponent_sign = parse_sign();
+
+    StringBuilder exponent_builder;
+    while (!done() && isdigit(ch()))
+        exponent_builder.append(consume());
+    VERIFY(exponent_builder.length() > 0);
+
+    auto exponent = exponent_builder.to_string().to_int().value();
+
+    // Fast path: If the number is 0, there's no point in computing the exponentiation.
+    if (number == 0)
+        return number;
+
+    if (exponent_sign < 0) {
+        for (int i = 0; i < exponent; ++i) {
+            number /= 10;
+        }
+    } else if (exponent_sign > 0) {
+        for (int i = 0; i < exponent; ++i) {
+            number *= 10;
+        }
+    }
+
     return number;
 }
 


Userland/Libraries/LibWeb/SVG/SVGPathElement.cpp
```diff
@@ -390,8 +390,33 @@ float PathDataParser::parse_fractional_constant()
...
+    auto exponent_sign = parse_sign();
+
+    StringBuilder exponent_builder;
+    while (!done() && isdigit(ch()))
+        exponent_builder.append(consume());
+    VERIFY(exponent_builder.length() > 0);
+
+    auto exponent = exponent_builder.to_string().to_int().value();
+
+    if (number == 0) // fast path: if the number is 0 theres no point to computing the exponentiation
```
awesomekling, line 407: theres -> there's

awesomekling: In fact, please write comments as proper sentences, e.g

```cpp
// Fast path: If the number is 0, there's no point in computing the exponentiation.
```

awesomekling (changes requested):

Userland/Libraries/LibWeb/SVG/SVGPathElement.cpp
```diff
@@ -390,8 +390,33 @@ float PathDataParser::parse_fractional_constant()
...
+    if (!match('e') && !match('E'))
+        return number;
+    consume();
+
+    auto exponent_sign = parse_sign();
+
+    StringBuilder exponent_builder;
+    while (!done() && isdigit(ch()))
+        exponent_builder.append(consume());
+    VERIFY(exponent_builder.length() > 0);
```
linusg, line 403: What happens when there's nothing after the `e` - we crash? What *should* happen?

Userland/Libraries/LibWeb/SVG/SVGPathElement.cpp
```diff
@@ -390,8 +390,33 @@ float PathDataParser::parse_fractional_constant()
...
+    if (number == 0) // fast path: if the number is 0 theres no point to computing the exponentiation
+        return number;
+
+    if (exponent_sign < 0) {
+        for (int i = 0; i < exponent; ++i) {
+            number /= 10;
+        }
+    } else if (exponent_sign > 0) {
+        for (int i = 0; i < exponent; ++i) {
+            number *= 10;
```
linusg, line 416: This can overflow, what should happen in that case?

Userland/Libraries/LibWeb/SVG/SVGPathElement.cpp
IdanHo: unfortunately this whole parser is built without any success/error reporting capabilities, so currently it asserts on unexpected input (see parse_number, parse_flag, parse_whitespace, etc), but what is likely supposed to happen is just some kind of a html parsing error

Userland/Libraries/LibWeb/SVG/SVGPathElement.cpp
IdanHo: these are floating point values, so "overflow" just means being +INFINITY, which would likely result in some very bad rendering, but im not sure if thats something the parser should handle

Userland/Libraries/LibWeb/SVG/SVGPathElement.cpp
IdanHo: sorry, fixed!

PR #6535, opened 2021-04-21
Piano: Add track Volume and improve QOL

kleinesfilmroellchen:
This patch implements a couple of enhancements to the synthesizer engine:
- Each track has a volume control.
- The input and tooltips for all controls are improved.
- The noise channel is pitched, which allows for basic drum synthesis.

### DIFF
--- a/Userland/Applications/Piano/KnobsWidget.cpp
+++ b/Userland/Applications/Piano/KnobsWidget.cpp
@@ -12,12 +12,6 @@
 #include <LibGUI/Label.h>
 #include <LibGUI/Slider.h>
 
-constexpr int max_attack = 1000;
-constexpr int max_decay = 1000;
-constexpr int max_sustain = 1000;
-constexpr int max_release = 1000;
-constexpr int max_delay = 8;
-
 KnobsWidget::KnobsWidget(TrackManager& track_manager, MainWidget& main_widget)
     : m_track_manager(track_manager)
     , m_main_widget(main_widget)
@@ -29,6 +23,7 @@ KnobsWidget::KnobsWidget(TrackManager& track_manager, MainWidget& main_widget)
     m_labels_container->set_layout<GUI::HorizontalBoxLayout>();
     m_labels_container->set_fixed_height(20);
 
+    m_volume_label = m_labels_container->add<GUI::Label>("Volume");
     m_octave_label = m_labels_container->add<GUI::Label>("Octave");
     m_wave_label = m_labels_container->add<GUI::Label>("Wave");
     m_attack_label = m_labels_container->add<GUI::Label>("Attack");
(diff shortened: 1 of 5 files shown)


Userland/Applications/Piano/main.cpp
```diff
@@ -104,6 +104,11 @@ int main(int argc, char** argv)
     });
     audio_thread->start();
 
+    auto main_widget_updator = Core::Timer::construct(150, [&] {
```
willmcpherson2, line 107: *updater
:P

Userland/Applications/Piano/RollWidget.cpp
```diff
@@ -88,6 +89,63 @@ void RollWidget::paint_event(GUI::PaintEvent& event)
     int horizontal_notes_to_paint = horizontal_paint_area / m_note_width;
 
     GUI::Painter painter(*this);
+
+    // Draw the background, if necessary.
+    if (viewport_changed() || paint_area != m_background->height()) {
+        dbgln("background redraw");
```
willmcpherson2: This `dbgln` gets added and then removed in the next commit

Userland/Applications/Piano/RollWidget.h
```diff
@@ -61,4 +62,10 @@ class RollWidget final : public GUI::ScrollableWidget {
     Optional<Gfx::IntPoint> m_note_drag_start;
     Optional<RollNote> m_note_drag_location;
     int m_drag_note;
+
+    RefPtr<Gfx::Bitmap> m_background;
+    int m_prev_zoom_level { m_zoom_level };
+    int m_prev_scroll_x { horizontal_scrollbar().value() };
+    int m_prev_scroll_y { vertical_scrollbar().value() };
+    bool viewport_changed();
```
willmcpherson2, line 70: `viewport_changed` can be `const`. It should also reside with the other methods.

willmcpherson2:
1.

We write commit message titles in the imperative mood ("do this", "add that"), so the commit messages should be more like:

```
Userland: Piano: Optimize repaints
```

```
Userland: Piano: Add volume and QOL improvements
```

2.

Preferably, the Volume/QOL commit should be split up into 3 commits: the volume knob, the new knob scaling improvements and the rewrite of `Track::noise`. Or at the very least, the commit message should mention the `Track::noise` rewrite. Also, it's really not necessary to rename `max_attack` to `attack_max` etc.

Other than those two points, these changes seem reasonable. And while I don't really understand how the optimisation works, it seems sound (just repaint the necessary stuff).

Userland/Applications/Piano/RollWidget.cpp
kleinesfilmroellchen: Yes, this is an issue with the previous commit that awaits merge. I think it's OK to leave it in if both PRs are merged successively.

Userland/Applications/Piano/main.cpp
kleinesfilmroellchen: this belongs to the previous commit

kleinesfilmroellchen:
2. I'm putting these into a single commit because they seem minor to me and are closely related. The Track::noise rewrite is part of the previous PR.

willmcpherson2 (approved):
This all looks good to me. Although try to keep your commits atomic. For example, renaming `max_attack` while also changing code that relies on `max_attack` makes the diff unnecessarily complicated. If your commit message has 3 dot points, meaning it does 3 things, it is usually reasonable to use 3 commits.

PR #6536, opened 2021-04-21
Everywhere: Remove redundant inline keyword with constexpr

ldm5180:
Problem:
- `constexpr` functions are additionally decorated with `inline`
  keyword. This is redundant since `constexpr` implies `inline`.

Solution:
- Remove redundancies.

### DIFF
--- a/AK/Format.h
+++ b/AK/Format.h
@@ -432,10 +432,10 @@ void dmesgln(CheckedFormatString<Parameters...>&& fmt, const Parameters&... para
 #endif
 
 template<typename T, typename = void>
-inline constexpr bool HasFormatter = true;
+constexpr bool HasFormatter = true;
 
 template<typename T>
-inline constexpr bool HasFormatter<T, typename Formatter<T>::__no_formatter_defined> = false;
+constexpr bool HasFormatter<T, typename Formatter<T>::__no_formatter_defined> = false;
 
 template<typename T>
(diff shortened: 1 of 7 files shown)


AK/Format.h
```diff
@@ -432,10 +432,10 @@ void dmesgln(CheckedFormatString<Parameters...>&& fmt, const Parameters&... para
 #endif
 
 template<typename T, typename = void>
-inline constexpr bool HasFormatter = true;
+constexpr bool HasFormatter = true;
```
alimpfard, line 435: This is not correct, constexpr global variables are not inline - only those with static duration are.

alimpfard: I'm surprised this doesn't make the linker cry about multiple definitions

trflynn89:
I *think* template constexpr declarations still need to be inline, i.e. these two declarations below are not equivalent:
```c++
template<typename T>
constexpr bool HasFormatter<T, typename Formatter<T>::__no_formatter_defined> = false;

template<typename T>
inline constexpr bool HasFormatter<T, typename Formatter<T>::__no_formatter_defined> = false;
```

linusg:
Ah dammit, saw these comments too late :(

ldm5180:
Yes. You are right. I am also surprised the compiler didn't catch those. Will fix.

PR #6537, opened 2021-04-21
LibGUI: Prevent selecting empty line in TextEditor on double click

Sauler:


Userland/Libraries/LibGUI/TextDocument.cpp
```diff
@@ -141,6 +141,18 @@ bool TextDocumentLine::ends_in_whitespace() const
     return isspace(code_points()[length() - 1]);
 }
 
+bool TextDocumentLine::can_select() const
+{
+    if (is_empty())
+        return false;
+    for (size_t i = 0; i < length(); ++i) {
+        auto code_point = code_points()[i];
+        if (isblank(code_point) || !isspace(code_point))
```
IdanHo, line 150: Missing negation on isblank?

Sauler: isblank checks for a blank character; that is, a space or a tab. 

I did not want to block the possibility of selecting a line containing only spaces or tabs. Just other white-space characters like newline `\n`or cariage return `\r`

IdanHo: i see, thats not the most clear way to express that (wouldnt (code_point != '\n' && code_point != '\r') be easier?), but thats fine then :)

Sauler: I can change it to this:
`code_point != '\n' && code_point != '\r' && code_point != '\f' && code_point != '\v'`
but i'm not sure if it's readable enough.

Another option is something like this:
`isprint(code_point) || code_point == '\t'`

IdanHo: `isprint(code_point) || code_point == '\t'` is worse since that breaks unicode support :)

Sauler: Ok, so i have to do this?
`code_point != '\n' && code_point != '\r' && code_point != '\f' && code_point != '\v'`

IdanHo: i think thats clearer than the current check, but its up to you, if you think its fine like that, go for it :)

Sauler: I updated it and it looks better :)

PR #6538, opened 2021-04-21
LibSQL: Parse expressions and the DELETE statement

trflynn89:
Started adding support for parsing `DELETE` statements, but these statements depend on "expressions", so the bulk of the work is parsing [that beast](https://sqlite.org/lang_expr.html). This doesn't parse all expressions; most notably, not yet parsing `function` expressions and not (fully) parsing expressions that depend on `SELECT`.

sql


Userland/Libraries/LibSQL/AST.h
```diff
@@ -48,11 +51,9 @@ class ASTNode : public RefCounted<ASTNode> {
     ASTNode() = default;
 };
 
-class Statement : public ASTNode {
-};
-
-class ErrorStatement final : public Statement {
-};
+//==================================================================================================
+// Languange types
```
awesomekling, line 55: Language

Issue #6068, opened 2021-04-02
LibC: statfs.h and statvfs.h missing


CompaqDisc:
I've recently been attempting to port over [pacman](https://archlinux.org/pacman/) from Arch Linux, and the most recent roadblock I'm encountering looks to be tied to a missing `stat(v)fs` struct and an undefined `ST_RDONLY`.

linusg:
> https://man7.org/linux/man-pages/man2/statfs.2.html
>
> CONFORMING TO
>
>        Linux-specific.  The Linux statfs() was inspired by the 4.4BSD
>        one (but they do not use the same structure).

CompaqDisc:
So I've been doing some digging on this, and it looks like either the Linux-specific `statfs()` or (what seems to be POSIX) [`statvfs()`](https://man7.org/linux/man-pages/man2/statvfs.2.html) can be used with appropriate configuration.

Based on that I'd say `statvfs()` is fair game, although it looks like it depends on adding a syscall.

sin-ack:
This is also required by Wine, so I'm looking into how hard implementing this is now.

panos:
Also required by at least two ports I've tried to work on.

Hendiadyoin1:
This seems to be fixed?

PR #6540, opened 2021-04-21
LibC: Add additional data types

panos:
- Reorganized some variables (alphabetically and based on their function) so that the new ones don't stick out.
- See: https://github.com/SerenityOS/serenity/issues/6068

Small, hesitant steps!

## Additional notes

- The following variables are important for many programs that calculate the available space on the disk. They still need to be implemented properly.

```cpp
typedef int64_t fsblkcnt_t;
typedef int64_t fsfilcnt_t;
```

- I added `typedef uint16_t u_mode_t;` while I was at it because I was familiar with the variable being used with umasks, not 100% sure if I should keep it or not.

### DIFF
--- a/Userland/Libraries/LibC/sys/types.h
+++ b/Userland/Libraries/LibC/sys/types.h
@@ -57,15 +57,19 @@ typedef __WINT_TYPE__ wint_t;
 typedef uint32_t ino_t;
 typedef int64_t off_t;
 
+typedef uint32_t blkcnt_t;
+typedef uint32_t blksize_t;
 typedef uint32_t dev_t;
 typedef uint16_t mode_t;
 typedef uint32_t nlink_t;
-typedef uint32_t blksize_t;
-typedef uint32_t blkcnt_t;
+
 typedef int64_t time_t;
 typedef uint32_t useconds_t;
 typedef int32_t suseconds_t;
 typedef uint32_t clock_t;
+
+typedef uint64_t fsblkcnt_t;
+typedef uint64_t fsfilcnt_t;
 
 #define __socklen_t_defined
 #define __socklen_t uint32_t


Userland/Libraries/LibC/sys/types.h
```diff
@@ -57,16 +57,21 @@ typedef __WINT_TYPE__ wint_t;
 typedef uint32_t ino_t;
 typedef int64_t off_t;
 
+typedef uint32_t blksize_t;
+typedef uint32_t blkcnt_t;
```
linusg, line 61: > alphabetically

```suggestion
typedef uint32_t blkcnt_t;
typedef uint32_t blksize_t;
```

Userland/Libraries/LibC/sys/types.h
```diff
@@ -57,16 +57,21 @@ typedef __WINT_TYPE__ wint_t;
...
-typedef uint32_t blkcnt_t;
+
 typedef int64_t time_t;
 typedef uint32_t useconds_t;
+typedef uint16_t u_mode_t;
 typedef int32_t suseconds_t;
 typedef uint32_t clock_t;
 
+typedef int64_t fsblkcnt_t;
+typedef int64_t fsfilcnt_t;
```
linusg, line 73: These are `uint64_t` on other systems

Userland/Libraries/LibC/sys/types.h
```diff
@@ -57,16 +57,21 @@ typedef __WINT_TYPE__ wint_t;
...
+typedef uint32_t blkcnt_t;
 typedef uint32_t dev_t;
 typedef uint16_t mode_t;
 typedef uint32_t nlink_t;
-typedef uint32_t blksize_t;
-typedef uint32_t blkcnt_t;
+
 typedef int64_t time_t;
 typedef uint32_t useconds_t;
+typedef uint16_t u_mode_t;
```
linusg, line 68: what's that? It's not present in any `types.h` I've checked and I can find literally 0 results on google..

image

linusg (changes requested):
please also squash the second commit into the first.

Userland/Libraries/LibC/sys/types.h
linusg: just saw that `umode_t` is a thing, so probably typo. then why is it next to the time related typedefs though? Also seems to be Linux-specific?

panos: > just saw that `umode_t` is a thing, so probably typo. then why is it next to the time related typedefs though? Also seems to be Linux-specific?

I also looked for it under the FreeBSD sources, and found it under the OpenZFS source code. Turns out that part of code uses a Linux compatibility layer, so I'll promptly remove it

Userland/Libraries/LibC/sys/types.h
panos: Wait, really? I was studying a few man pages which seemed a bit unclear about this and I wondered what happened in very specific exceptional cases, having remembered a few misfortunes that I dealt with when I struggled with using `dd` right under Linux... where the result to how much free space there could've been negative.

I'm assuming what I saw was the result of subtraction, since the available blocks back then  was `0`.

Userland/Libraries/LibC/sys/types.h
panos: I got it definitely wrong. I also looked at Haiku because I wanted to check whether I was missing something since Haiku has a BSD compatibility layer and not a Linux one, and since this entire change concerned filesystems in particular, I figured that I must've missed something and added it too with extreme cautiousness as I couldn't find any search results either.

I'll ask their mailing list, just in case, and promptly remove it since every other source seems to say that this is useless.

Userland/Libraries/LibC/sys/types.h
linusg: https://github.com/openbsd/src/blob/2207c4325726fdc5c4bcd0011af0fdf7d3dab137/sys/sys/types.h#L151-L152
https://github.com/openbsd/src/blob/2207c4325726fdc5c4bcd0011af0fdf7d3dab137/sys/sys/_types.h#L46-L47

panos: OHHHHHH, Yeah, I had no idea where the hell `__fsfilcnt_t` was and was thrown off by the second definition in the file.

TIL: [FreeBSD does ` sys/sys/_types.h `](https://github.com/openbsd/src/blob/2207c4325726fdc5c4bcd0011af0fdf7d3dab137/sys/sys/_types.h).

PR #6541, opened 2021-04-21
AK/Format: Compute TypeErasedParameter type and size at compile-time

ldm5180:
Problem:
- Type and size information is known at compile-time, but computations
  are being performed using run-time parameters.

Solution:
- Move function arguments to be template arguments.
- Convert to `consteval` where possible.
- Decorate functions with `constexpr` which are used in both run-time
  and compile-time contexts.

### DIFF
--- a/AK/Format.h
+++ b/AK/Format.h
@@ -65,25 +65,26 @@ struct TypeErasedParameter {
         Custom
     };
 
-    static Type get_type_from_size(size_t size, bool is_unsigned)
+    template<size_t size, bool is_unsigned>
+    static consteval Type get_type_from_size()
     {
-        if (is_unsigned) {
-            if (size == 1)
+        if constexpr (is_unsigned) {
+            if constexpr (size == 1)
                 return Type::UInt8;
-            if (size == 2)
+            if constexpr (size == 2)
                 return Type::UInt16;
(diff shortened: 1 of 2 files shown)


AK/Format.h
```diff
@@ -65,41 +65,68 @@ struct TypeErasedParameter {
...
     {
         if constexpr (IsIntegral<T>)
-            return get_type_from_size(sizeof(T), IsUnsigned<T>);
+            return get_type_from_size<sizeof(T), IsUnsigned<T>>();
         else
             return Type::Custom;
     }
 
-    size_t to_size() const;
+    constexpr size_t to_size() const
```
alimpfard, line 103: can this actually be `constexpr`? `VERIFY` calls a non-constexpr function I think (or is it inline asm?)

alimpfard: Seems to be a normal non-constexpr function (its definition isn't even available)
I'm not sure if this is allowed like it would be in `consteval` functions

alimpfard: Looks like the spec doesn't explicitly ban this in C++20, so this is fine...most likely (dcl.constexpr section 3)

alimpfard (approved):

PR #6544, opened 2021-04-21
Ports: Add launchers for some of the ports

gunnarbeutner:
There's still one thing I'm not quite happy yet: The wrapper scripts are kinda ugly.

My other idea was to change the launcher by adding two new attributes for app files: `arguments` (and _maybe_ `workdir` - though that's really a problem that's quite specific to the Super Mario port - it crashes without the right working directory and could probably be patched in the port instead).

Or maybe have the launcher detect whether an app is a GUI application (i.e., whether it links against `LibGUI`) maybe? Windows has this thing where it does an implicit `AllocConsole` for console applications. This could also be useful so you can launch console apps from the file manager. Ideas welcome.

However, that's probably a somewhat larger change for another PR.

### DIFF
--- a/Ports/.port_include.sh
+++ b/Ports/.port_include.sh
@@ -68,6 +68,9 @@ shift
 : "${auth_type:=md5}"
 : "${auth_import_key:=}"
 : "${auth_opts:=}"
+: "${launcher_name:=}"
+: "${launcher_category:=}"
+: "${launcher_command:=}"
 
 run_nocd() {
     echo "+ $@ (nocd)"
@@ -77,8 +80,33 @@ run() {
     echo "+ $@"
     (cd "$workdir" && "$@")
 }
-run_replace_in_file(){
+run_replace_in_file() {
     run perl -p -i -e "$1" $2
+}
+install_launcher() {
+    if [ -z "$launcher_name" ] || [ -z "${launcher_category}" ] || [ -z "${launcher_command}" ]; then
+        return
+    fi
+    script_name="${launcher_name,,}"
+    script_name="${script_name// /}"
+    mkdir -p $DESTDIR/usr/local/libexec
+    cat >$DESTDIR/usr/local/libexec/$script_name <<SCRIPT
+#!/bin/sh
+set -e
+cd -- "\$(dirname -- "\$(which -- $(printf %q "${launcher_command%% *}"))")"
+exec $(printf '%q ' $launcher_command)
+SCRIPT
+    chmod +x $DESTDIR/usr/local/libexec/$script_name
+
+    chmod +x $DESTDIR/usr/local/libexec
+    mkdir -p $DESTDIR/res/apps
+    cat >$DESTDIR/res/apps/$script_name.af <<CONFIG
+[App]
+Name=$launcher_name
+Executable=/usr/local/libexec/$script_name
+Category=$launcher_category
+CONFIG
+    unset script_name
 }
 # Checks if a function is defined. In this case, if the function is not defined in the port's script, then we will use our defaults. This way, ports don't need to include these functions every time, but they can override our defaults if needed.
 func_defined() {
@@ -219,6 +247,7 @@ func_defined build || build() {
 }
 func_defined install || install() {
     run make DESTDIR=$DESTDIR $installopts install
+    install_launcher
 }
 func_defined post_install || post_install() {
     echo
--- a/Ports/SDLPoP/package.sh
+++ b/Ports/SDLPoP/package.sh
@@ -7,6 +7,9 @@ workdir=SDLPoP-86988c668eeaa10f218e1d4938fc5b4e42314d68
 configopts="-DCMAKE_TOOLCHAIN_FILE=$SERENITY_SOURCE_DIR/Toolchain/CMake/CMakeToolchain.txt"
 files="https://github.com/NagyD/SDLPoP/archive/86988c668eeaa10f218e1d4938fc5b4e42314d68.zip PoP.zip d18cae8541fb8cbcc374fd998316993d561429a83f92061bc0754337ada774c5"
 auth_type=sha256
+launcher_name="Prince of Persia"
+launcher_category=Games
+launcher_command=/opt/PrinceOfPersia/prince
 
 configure() {
     run cmake $configopts ./src
@@ -15,4 +18,5 @@ configure() {
 install() {
     mkdir -p "${SERENITY_INSTALL_ROOT}/opt/PrinceOfPersia"
     run cp -r prince data SDLPoP.ini "${SERENITY_INSTALL_ROOT}/opt/PrinceOfPersia" 
+    install_launcher
 }
--- a/Ports/Super-Mario/package.sh
+++ b/Ports/Super-Mario/package.sh
@@ -7,6 +7,9 @@ workdir=Super-Mario-Clone-Cpp-master
 configopts="-DCMAKE_TOOLCHAIN_FILE=$SERENITY_SOURCE_DIR/Toolchain/CMake/CMakeToolchain.txt"
 files="https://github.com/Bennyhwanggggg/Super-Mario-Clone-Cpp/archive/refs/heads/master.zip master.zip 11f622721d1ba504acf75c024aa0dbe3"
 auth_type=md5
+launcher_name="Super Mario"
+launcher_category=Games
+launcher_command=/opt/Super_Mario/uMario
 
 configure() {
     run cmake $configopts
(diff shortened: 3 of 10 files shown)


panos:
> This could also be useful so you can launch console apps from the file manager. Ideas welcome.

Some systems, e.g. penetration testing-related Linux distributions, work around with this by installing `.desktop` entries that open a Terminal and pass a specific parameter so that the Terminal invokes a specific command as soon as it's started.

Ports/.port_include.sh
```diff
@@ -77,9 +80,33 @@ run() {
...
+        return
+    fi
+    script_name="${launcher_name,,}"
+    script_name="${script_name// /}"
+    mkdir -p $DESTDIR/usr/local/libexec
+    cat >$DESTDIR/usr/local/libexec/$script_name <<SCRIPT
+#!/bin/sh
+cd \$(dirname -- $launcher_command)
+exec $launcher_command
+SCRIPT
```
emanuele6, line 97:
```suggestion
    cat >$DESTDIR/usr/local/libexec/$script_name <<SCRIPT
#!/bin/sh
set -e
cd -- "\$(dirname -- $(printf %q "$launcher_command"))"
exec $(printf %q "$launcher_command")
SCRIPT
```

emanuele6: If you actually want to expand lancher command into `first_word other_words_as_args` (which wouldn't have worked anyway in your version), you can use this:

```suggestion
    cat >$DESTDIR/usr/local/libexec/$script_name <<SCRIPT
#!/bin/sh
set -e
cd -- "\$(dirname -- $(printf %q "${launcher_command%% *}"))"
exec $(for w in $lancher_command; do
    printf '%q\n' $launcher_command
done)
SCRIPT
```
EXTRA NOTE: This code has a small issue if `$lancher_command` is empty; It can easility be fixed \([see here](https://gist.github.com/emanuele6/5f8cf70fb3f5b1ee6d9053c13ec850b5)\), but, since there is a `-z $lancher_command` check above, we can avoid dealing with it.

linusg:
Cool stuff!

> My other idea was to change the launcher by adding two new attributes for app files: arguments (and maybe workdir - though that's really a problem that's quite specific to the Super Mario port - it crashes without the right working directory and could probably be patched in the port instead).

That sounds good, in case you want to go that route.

> Some systems, e.g. penetration testing-related Linux distributions, work around with this by installing .desktop entries that open a Terminal and pass a specific parameter so that the Terminal invokes a specific command as soon as it's started.

That's indeed a workaround at best, let's avoid it if we can :)

emanuele6 (changes requested):
In your version, if `$lanucher_command` contains spaces:

`$(dirname -- $launcher_command)` will return multiple values.

The `cd` will then fail since you didn't put `"`s around the `$()` expansion (`cd` fails if you give it more than one non-option argument).

The script continues and runs `$launcher_command` expanded into `first_word other_words_as_args...` (since it is not quoted) in the "$PWD" (because the `cd` failed).

Also, since you were not using `bash`'s `%q`, some values of `$launcher_command` can potentially generate syntax errors.

I clean it up a bit. ;^)

Ports/.port_include.sh
gunnarbeutner: Hm, how about this?

```
    cat >$DESTDIR/usr/local/libexec/$script_name <<SCRIPT
#!/bin/sh
set -e
cd -- "\$(dirname -- \$(which $(printf %q "${launcher_command%% *}")))"
exec $(for w in $launcher_command; do
    printf '%q ' $w
done)
SCRIPT
```

I added `$(which ...)` so I don't have to specify the full path in the `launcher_command`, e.g. `Terminal -e nyancat`. Plus some minor fixes for variable names from your version.

The script that gets created looks like this:

```
#!/bin/sh
set -e
cd -- "$(dirname -- $(which Terminal))"
exec Terminal -e nyancat 
```

gunnarbeutner: Looks like `/bin/dirname` and `cd` don't currently understand `--`. So while obviously more correct that way I'm removing that for now.

It also doesn't support `set`, but well, without `set -e` that's not an issue. Heh.

emanuele6: That's fine, but it should be:
```bash
    cat >$DESTDIR/usr/local/libexec/$script_name <<SCRIPT
#!/bin/sh
set -e
cd -- "\$(dirname -- "\$(which -- $(printf %q "${launcher_command%% *}"))")"
exec $(for w in $launcher_command; do
    printf '%q ' $w
done)
SCRIPT
```
```
#!/bin/sh
set -e
cd -- "$(dirname -- "$(which -- Terminal)")"
exec Terminal -e nyancat
```

gunnarbeutner: I've re-added the `--`s for `dirname` and `which` because apparently only `cd` doesn't support that. I've updated the PR again with the changes. :)

emanuele6: @gunnarbeutner: I just noticed that: `printf '%q ' $launcher_command` (with a space after `%q`) can be used instead of the lopp:
```bash
    cat >$DESTDIR/usr/local/libexec/$script_name <<SCRIPT
#!/bin/sh
set -e
cd -- "\$(dirname -- "\$(which -- $(printf %q "${launcher_command%% *}"))")"
exec $(printf '%q ' $launcher_command)
SCRIPT
```

And also solves that issue with `lancher_command` being empty.

gunnarbeutner: Updated. :+1:

Ports/.port_include.sh
```diff
@@ -77,9 +80,34 @@ run() {
...
+    if [ -z "$launcher_name" ] || [ -z "${launcher_category}" ] || [ -z "${launcher_command}" ]; then
+        return
+    fi
+    script_name="${launcher_name,,}"
+    script_name="${script_name// /}"
+    mkdir -p $DESTDIR/usr/local/libexec
+    cat >$DESTDIR/usr/local/libexec/$script_name <<SCRIPT
+#!/bin/sh
+set -e
+cd "\$(dirname -- \$(which -- $(printf %q "${launcher_command%% *}")))"
```
emanuele6, line 96: you forgot to add quotes around the which expansion:
```suggestion
cd "\$(dirname -- "\$(which -- $(printf %q "${launcher_command%% *}"))")"
```

gunnarbeutner: Not sure what I'd do without such a thorough review. Thanks! Updated again.

emanuele6: also, why did you remove the `--` after `cd`? POSIX requires that `cd` supports

emanuele6: `exec` doesn't have/need it because it doesn't have options, but `cd` needs it.

emanuele6: ping @gunnarbeutner

panos:
> That's indeed a workaround at best, let's avoid it if we can :)

Agreed. A buggy one at that too, particularly because many programs just take parameters directly and don't have a command line user interface.

Hm... What if you just displayed the help command and allowed the user to make decisions in a more interactive manner, thus giving these sorts of applications a more graphical spin?

gunnarbeutner:
Aye, that's gonna be another PR then. :)

Ports/.port_include.sh
```diff
@@ -77,9 +80,34 @@ run() {
...
+    if [ -z "$launcher_name" ] || [ -z "${launcher_category}" ] || [ -z "${launcher_command}" ]; then
+        return
+    fi
+    script_name="${launcher_name,,}"
+    script_name="${script_name// /}"
+    mkdir -p $DESTDIR/usr/local/libexec
+    cat >$DESTDIR/usr/local/libexec/$script_name <<SCRIPT
+#!/bin/sh
+set -e
+cd "\$(dirname -- "\$(which -- $(printf %q "${launcher_command%% *}"))")"
```
emanuele6, line 96: https://github.com/SerenityOS/serenity/pull/6544#discussion_r618168564

You missed this messages, you need a `--` after `cd`.

emanuele6: Why did you remove it though? did it give you errors when you tried adding it?

gunnarbeutner: Aye, `cd` doesn't seem to support `--` yet.

emanuele6: weird, I remember testing it on serenity yesterday.

I also tested it again right now and our `sh` seems to implement it.
flie
file2

----------

If you are just saying serenity doesn't seem to support it because it is not listed in the help menu, `--` is automatically supported by anything that uses an `ArgsParser` even that `ArgsParser` doesn't have any options.

gunnarbeutner: Ok, I guess it's a bit more complicated than just "doesn't seem to be supported":

Screenshot from 2021-04-23 10-14-05

(`x` doesn't  exist, so that error is OK.)

gunnarbeutner: Alright then, that's fixed. :partying_face:

emanuele6:
:shell: :bug: :^)

emanuele6:
Looks good to me.

PR #6545, opened 2021-04-22
Meta: Add basic Zsh completions for serenity.sh

linusg:
This patch adds a basic Zsh completion script for the commands and targets provided by `Meta/serenity.sh`. There's some room for improvement here, e.g. we could provide completion for available CMake targets - currently completion stops after `serenity.sh <command> <target>`.

You can enable it by adding this to your `.zshrc` before completions are loaded:

```sh
fpath=($SERENITY_SOURCE_DIR/Meta/ShellCompletions/zsh $fpath)
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@
 !Makefile
 !LICENSE
 !Base/**
+!Meta/ShellCompletions/**
 
 *.o
 *.ao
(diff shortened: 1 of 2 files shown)


linusg:
cc @alimpfard, this was inspired by citron's zsh completions - maybe you know better than me how this stuff works and wanna take a look?

alimpfard:
I don't actually remember how these work anymore :sweat_smile:

PR #6546, opened 2021-04-22
LibELF: Avoid calculating symbol hashes when we don't need them

gunnarbeutner:


Userland/Libraries/LibELF/DynamicObject.h
```diff
@@ -162,6 +162,25 @@ class DynamicObject : public RefCounted<DynamicObject> {
...
+        StringView name() const { return m_name; }
+        u32 gnu_hash() const;
+        u32 sysv_hash() const;
+
+    private:
+        StringView m_name;
+        mutable u32 m_gnu_hash;
+        mutable bool m_gnu_hash_used { false };
+        mutable u32 m_sysv_hash;
+        mutable bool m_sysv_hash_used { false };
```
awesomekling, line 181: Seems like `mutable Optional<u32>` would be a slightly more idiomatic fit here.

gunnarbeutner: Aye, good point. Fixed. :)

PR #6547, opened 2021-04-22
Everything: Move to SPDX license identifiers in all files.

bgianfo:
SPDX License Identifiers are a more compact / standardized
way of representing file license information.

See: https://spdx.dev/resources/use/#identifiers

This was done with the `ambr` search and replace tool.

 ambr --no-parent-ignore --key-from-file --rep-from-file key.txt rep.txt *


AK/Checked.h
```diff
@@ -1,28 +1,8 @@
 /*
  * Copyright (C) 2011-2019 Apple Inc. All rights reserved.
```
emanuele6, line 2: Can this really be replaced?

bgianfo: No I'll undo that thanks for checking... I found a few of these manually.. but hard to automate.

bgianfo: Ok fixed, I manually poked through the diff in vim ... I think this is the only case of this.

emanuele6: yep, I just pulled your branch and did a quick `git show | grep -F 'Copyright (c)' | sed -E 's@[0-9-]+,?@@' | sort -u | less` (72 results). everything looks good.

bgianfo: Thank you for double checking!

PR #6548, opened 2021-04-22
Toolchain/Dockerfile: Update to Ubuntu 21.04, add git

GMTA:
Also made it easier to manage the dependencies to install. Running `Meta/serenity.sh` was impossible with the resulting image since it requires `git`.

### DIFF
--- a/Toolchain/Dockerfile
+++ b/Toolchain/Dockerfile
@@ -1,12 +1,24 @@
-FROM ubuntu:20.10
+FROM ubuntu:21.04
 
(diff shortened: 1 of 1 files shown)


linusg:
20.04 is LTS, 21.04 isn't - will you update this again in ~8 months?

GMTA:
> 20.04 is LTS, 21.04 isn't - will you update this again in ~8 months?

Sure - if there are any benefits to using a LTS version for a Docker image (that gets outdated immediately) in a fast-moving environment like Serenity's, I'll gladly switch back. But I figure that we might as well use the latest greatest :-)

PR #6550, opened 2021-04-22
LibWeb: More rendering improvements

### LINKED ISSUE #6408: LibWeb: float-1.html test page renders incorrectly
The position of the elements on the float-1.html test page seems to have broken with commit 1d343116a9c814869bc708dffb61d9b100be0eda. Looking at the spec, I think we're not taking the bold text here into consideration:

> In addition, if the element has any floating descendants whose bottom margin edge is below the element's bottom content edge, then the height is increased to include those edges. **Only floats that participate in this block formatting context are taken into account**, e.g., floats inside absolutely positioned descendants or other floats are not.

float

How it should look:
float_before

Cleverking2003:
Fixes #6408


awesomekling (changes requested):
Please wrap your commit messages at 72 characters.

PR #6552, opened 2021-04-22
Toolchain: Update GCC to version 11.1.0

gunnarbeutner:
* Updates GCC to version 11.1.0.
* Suppressions for warnings in `Userland/Utilities/crash.cpp`. Well, yes, that would certainly read from/write to uninitialized memory. As it should.
* `Userland/Utilities/gzip.cpp`, `Userland/Utilities/tar.cpp`, `Userland/Utilities/zip.cpp`: GCC is complaining about holding a const-ref to a temporary that's the result of an implicit conversion (`const char*` -> `String`). While binding a const-ref to a temporary is fine (i.e. the object exists until the end of the local scope) the implicit conversion is what GCC doesn't seem to like.
* `Userland/Libraries/LibGUI/Label.cpp`: I don't see how `start.value()` could reference uninitialized memory here. In fact, hoisting `m_text.length()` out of the for loop suppresses this warning. This feels like a bug in GCC.


gunnarbeutner:
Looks like building the gcc port fails.

gunnarbeutner:
I've run a bunch of tests to make sure this works properly:

* Successfully built the GCC port.
* Successfully built some test program in SerenityOS with the GCC port.
* Checked that the test program's stacktrace shows up properly when it crashes. Works.
* Checked the profile to make sure stacktraces work there as well. Works as well.
* Tried a few more programs to see if crash dumps work. `UserspaceEmulator` has a DIE with `AttributeDataForm::ImplicitConst` which we need to add support for.

gunnarbeutner:
#6719 fixes backtraces for UE and other programs.

Super Mario crashes in its static initializers, though this is likely a bug in that - rather than gcc:
```
CrashDaemon(15): New coredump file: /tmp/coredump/uMario_27_1619674230
CrashDaemon(15): --- Backtrace for thread #0 (TID 27) ---
CrashDaemon(15): 0x6b2360a8: [/opt/Super_Mario/uMario] Music::Music() +0x50
CrashDaemon(15): 0x6b1bd4e6: [/opt/Super_Mario/uMario] __static_initialization_and_destruction_0(int, int) +0xb6
CrashDaemon(15): 0x6b1bd594: [/opt/Super_Mario/uMario] _GLOBAL__sub_I__ZN4CCFGC2Ev +0x1f```

PR #6553, opened 2021-04-22
AK/GenericLexer: constexpr where possible

ldm5180:
Problem:
- Much of the `GenericLexer` can be `constexpr`, but is not.

Solution:
- Make it `constexpr` and de-duplicate code.
- Extend some of `StringView` with `constexpr` to support.
- Add tests to ensure `constexpr` behavior.

Note:
- Construction of `StringView` from pointer and length is not
  `constexpr`-compatible at the moment because the VERIFY cannot be,
  yet.


ADKaster:
> Construction of StringView from pointer and length is not constexpr-compatible at the moment because the VERIFY cannot be, yet.

Is this a problem that could be solved with a version of `std::is_constant_evaluated()` ?

AK/Tests/TestGenericLexer.cpp
```diff
@@ -0,0 +1,180 @@
...
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
```
linusg, line 25: Copyright headers don't include the full text anymore since #6547, please update.

linusg (changes requested):

AK/Tests/TestGenericLexer.cpp
ldm5180: Oh! I like this much better.

ldm5180:
> > Construction of StringView from pointer and length is not constexpr-compatible at the moment because the VERIFY cannot be, yet.
> 
> Is this a problem that could be solved with a version of `std::is_constant_evaluated()` ?

Yes, but `AK::IsConstantEvaluated` is not yet implemented.

PR #6559, opened 2021-04-22
LibGUI: Select last word when double clicking at the end of a line

Sauler:
`first_word_break_before` method was accessing values outside of `code_points` array bounds when user double-clicked at end of the line.

Also fixes: https://github.com/SerenityOS/serenity/issues/6565

### DIFF
--- a/Userland/Libraries/LibGUI/TextDocument.cpp
+++ b/Userland/Libraries/LibGUI/TextDocument.cpp
@@ -628,7 +628,11 @@ TextPosition TextDocument::first_word_break_before(const TextPosition& position,
 
     auto target = position;
     auto line = this->line(target.line());
-    auto is_start_alphanumeric = isalnum(line.code_points()[target.column() - (start_at_column_before ? 1 : 0)]);
+    auto modifier = start_at_column_before ? 1 : 0;
+    if (target.column() == line.length())
+        modifier = 1;
+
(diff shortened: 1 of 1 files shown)


linusg:
Please explain in the commit message what the issue was :)

Sauler:
Do you mean copying PR description into commit message?
Or i have to write what issue was in code?

linusg:
> Word selection when double clicking at end of line was unpredictable. Sometimes it works and sometimes it doesn't.

...on its own this isn't helpful, please explain why it was unpredictable and what the fix is to make it not unpredictable.

Sauler:
What about something like this?

> This commit fixes double click word selection behaviour. Previously word was not always selected after double clicking on end of the line due to isalnum() check outside of line code_points array. Now there is check if cursor position is at the end of line to prevent this to happen.

English is not my native language and sometimes i have trouble with explaining things :(

alimpfard:
Perhaps it could help to first make an issue describing what exactly is going wrong, then just referencing the issue in the fix commit, and maybe explaining what was going wrong.

Sauler:
https://github.com/SerenityOS/serenity/issues/6565 issue for this PR :)

Sauler:
What's now?

alimpfard:
Reword the commit, and wait for review I guess :P

Sauler:
Ok, i think that i fixed my PR :)

Issue #6567, opened 2021-04-22
Browser: ListItemMarkers collide with the ListItem in <ol>'s


TobyAsE:
When the ListItemMarkers become too wide (because there are multiple characters in them) they collide with the ListItems.
They should be lay out taking the actual width into account, currently they have a fixed offset.

PR #6568, opened 2021-04-22
Support more <ol> list styles

TobyAsE:
I had already done a few `list-style-type`s so I thought why not add a few more?
This adds support for  `list-style-type:` `upper-latin`, `upper-alpha`, `lower-latin`, `lower-alpha` and `decimal-leading-zero`.

The conversion number->alphabet is a bit tricky since the alphabet doesn't have a zero. In case anybody else finds this, the term you want to search for is "bijective numeration" and "bijective base".
Also I made it very alphabet-agnostic to support other easy counting-methods using different charsets. There are a lot out there.
 
I noticed that the markers can collide with the list item itself when they become wider. This needs to be tackled in another PR (issue: #6567) as it seems to not be that trivial.

### DIFF
--- a/Userland/Libraries/LibWeb/CSS/Identifiers.json
+++ b/Userland/Libraries/LibWeb/CSS/Identifiers.json
@@ -160,6 +160,8 @@
   "text",
   "underline",
   "uppercase",
+  "upper-alpha",
+  "upper-latin",
   "visible",
   "vertical-text",
   "wait",
--- a/Userland/Libraries/LibWeb/CSS/Identifiers.json
+++ b/Userland/Libraries/LibWeb/CSS/Identifiers.json
@@ -104,6 +104,8 @@
   "line-through",
   "list-item",
   "lowercase",
+  "lower-alpha",
+  "lower-latin",
   "medium",
   "move",
   "ne-resize",
--- a/Userland/Libraries/LibWeb/CSS/Identifiers.json
+++ b/Userland/Libraries/LibWeb/CSS/Identifiers.json
@@ -77,6 +77,7 @@
   "crosshair",
   "dashed",
   "decimal",
+  "decimal-leading-zero",
   "default",
   "disc",
   "dotted",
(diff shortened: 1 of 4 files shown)


Userland/Libraries/LibWeb/Layout/ListItemMarkerBox.cpp
```diff
@@ -21,6 +21,21 @@ ListItemMarkerBox::~ListItemMarkerBox()
...
+        number--;
+        auto remainder = number % base;
+        number -= remainder;
+        output_string.append(alphabet[remainder]);
+        number /= base;
+    };
+    output_string.append(alphabet[number - 1]);
+    return output_string.to_string().reverse();
+}
+
```
alimpfard, line 42: This also exists in Spreadsheet (`String Spreadsheet::convert_to_string(size_t value, unsigned base = 26, StringView map = {})`) along with its inverse, maybe extract that and use it here?

TobyAsE: Ah shoot! Let's take this then. Should we move the functionality to String and friends somewhere?

alimpfard: Yeah, probably `String::bijective_base_from(number, <base>, map)` or something?

TobyAsE: Sounds good.
I will do that ;)

awesomekling: I'll merge this as-is, and we can follow up with changes to add a string helper for bijective basin'.

Userland/Libraries/LibWeb/CSS/Identifiers.json
```diff
@@ -171,4 +173,4 @@
   "xxx-large",
   "zoom-in",
   "zoom-out"
-]
\ No newline at end of file
+]
```
awesomekling: Please make sure all commit messages have a `Category:` prefix.

Also, please configure your editor to add a newline at the end of files.

awesomekling (changes requested):

Userland/Libraries/LibWeb/CSS/Identifiers.json
TobyAsE: Somehow clang-format removed the trailing newline in my setup. It's fixed now.

PR #6577, opened 2021-04-23
Revert "AK: Remove virtual destructors from non-virtual classes"

IdanHo:
Reverts SerenityOS/serenity#6573, as it causes page faults in userland. (BaseRedBlackTree::Node is a virtual class, and as such its destructor should be virtual as well)


bgianfo:
CC: @ldm5180

PR #6580, opened 2021-04-23
Ports: Use a specific version for frotz

gunnarbeutner:
The hash for the master zip file changed again. Probably because GitLab only caches those zip files for a bit and re-generates them with slightly different zip headers after some time even though the repository didn't change.

This also changes the build message for skipped ports. The message for skipping builds makes it hard to distinguish failed builds from builds that were just skipped so change it back to the old one - but don't actually build packages twice again.


linusg:
https://github.com/SerenityOS/serenity/blob/master/Ports/AvailablePorts.md ;)

PR #6581, opened 2021-04-23
LibTLS: Call the read hooks after processing messages too

alimpfard:
Otherwise the notification would be deferred until the next read event,
which means the client will not get any events if the server initiates
the appdata transfers.


linusg:
oh hey that looks like it might fix an issue with TLS websockets I was having :)

PR #6585, opened 2021-04-23
Browser: Add basic support for search engines

sppmacd:
If you enter a likely invalid URL (URL with no dots or with spaces),
the browser will use user-selected search engine (none as default).
    
For now, there are several engines hardcoded and there is no support
for custom search engines.

### DIFF
--- a/Userland/Applications/Browser/Tab.cpp
+++ b/Userland/Applications/Browser/Tab.cpp
@@ -1,5 +1,6 @@
 /*
  * Copyright (c) 2020-2021, Andreas Kling <kling@serenityos.org>
+ * Copyright (c) 2021, Maciej Zygmanowski <sppmacd@pm.me>
  *
  * SPDX-License-Identifier: BSD-2-Clause
  */
@@ -38,8 +39,14 @@
 
 namespace Browser {
 
+static String s_search_engine_format = {};
+
 URL url_from_user_input(const String& input)
 {
+    if (input.starts_with("?") && !s_search_engine_format.is_null()) {
+        return URL(String::formatted(s_search_engine_format, urlencode(input.substring(1))));
+    }
+
     auto url = URL(input);
     if (url.is_valid())
         return url;
@@ -396,6 +403,37 @@ Tab::Tab(Type type)
         this);
     js_console_action->set_status_tip("Open JavaScript console for this page");
     inspect_menu.add_action(js_console_action);
+
+    auto& settings_menu = m_menubar->add_menu("&Settings");
+
+    m_search_engine_actions.set_exclusive(true);
+    auto& search_engine_menu = settings_menu.add_submenu("&Search Engine");
+
+    auto add_search_engine = [&](auto& name, auto& url_format) {
+        auto action = GUI::Action::create_checkable(
+            name, [&](auto&) {
+                dbgln("Setting search engine to {}", url_format);
+                s_search_engine_format = url_format;
+            },
+            this);
+        search_engine_menu.add_action(action);
+        m_search_engine_actions.add_action(action);
+    };
+
+    auto disable_search_engine_action = GUI::Action::create_checkable(
+        "Disable", [this](auto&) {
+            dbgln("Disabling search engine");
+            s_search_engine_format = {};
+        },
+        this);
+    search_engine_menu.add_action(disable_search_engine_action);
+    m_search_engine_actions.add_action(disable_search_engine_action);
+    disable_search_engine_action->set_checked(true);
+
+    // FIXME: Support adding custom search engines
+    add_search_engine("Bing", "https://www.bing.com/search?q={}");
+    add_search_engine("DuckDuckGo", "https://duckduckgo.com/?q={}");
+    add_search_engine("Google", "https://google.com/search?q={}");
 
     auto& debug_menu = m_menubar->add_menu("&Debug");
     debug_menu.add_action(GUI::Action::create(
--- a/Userland/Applications/Browser/Tab.h
+++ b/Userland/Applications/Browser/Tab.h
@@ -100,6 +100,7 @@ private:
     URL m_image_context_menu_url;
 
     GUI::ActionGroup m_user_agent_spoof_actions;
+    GUI::ActionGroup m_search_engine_actions;
     RefPtr<GUI::Action> m_disable_user_agent_spoofing;
 
     RefPtr<GUI::Menu> m_tab_context_menu;


Userland/Applications/Browser/Tab.cpp
```diff
@@ -38,6 +39,13 @@
 
 namespace Browser {
 
+static String s_search_engine_format = "{}";
+
+static bool may_url_exist(URL& url)
+{
+    return url.is_valid() && !url.host().find(' ').has_value() && (url.host().find('.').has_value() || url.host() == "localhost");
```
linusg, line 46: Please don't do this - if I put `http://foo` in the addressbar, I want the browser to load `http://foo`. You can instead do this *before* constructing a URL below.

linusg (changes requested):

Userland/Applications/Browser/Tab.cpp
```diff
@@ -397,6 +414,38 @@ Tab::Tab(Type type)
...
+        this);
+    search_engine_menu.add_action(disable_search_engine_action);
+    m_search_engine_actions.add_action(disable_search_engine_action);
+    disable_search_engine_action->set_checked(true);
+
+    // FIXME: Support adding custom search engines
+    add_search_engine("Bing", "https://www.bing.com/search?q={}");
+    add_search_engine("DuckDuckGo", "https://duckduckgo.com/?q={}");
+    add_search_engine("Google", "https://google.com/search?q={}");
+    add_search_engine("Wikipedia", "https://en.wikipedia.org/w/index.php?search={}");
```
IdanHo, line 447: We cant actually load wikipedia at the moment (As we dont support any modern TLS1.2/1.3 cipher suites), so having it as an option right now is not very helpful

Userland/Applications/Browser/Tab.cpp
sppmacd: Fixed.

Userland/Applications/Browser/Tab.cpp
```diff
@@ -47,6 +55,15 @@ URL url_from_user_input(const String& input)
...
     builder.append(input);
+    url = URL(builder.build());
+
+    if (url.is_valid() && (may_url_exist(input)) || s_search_engine_format == "{}")
+        return url;
+
+    // Try using search engine
+    dbgln("Using search engine for {}", input);
+    builder.clear();
+    AK::vformat(builder, s_search_engine_format, AK::VariadicFormatParams { input });
```
alimpfard, line 66: ???
`builder.appendff(...)`?

alimpfard: `s_search_engine_format` is `"{}"`, so why not just `builder.append(input)`?

sppmacd: Oh I didn't know about this :)

Userland/Applications/Browser/Tab.cpp
```diff
@@ -47,6 +55,15 @@ URL url_from_user_input(const String& input)
     StringBuilder builder;
     builder.append("http://");
     builder.append(input);
+    url = URL(builder.build());
+
+    if ((url.is_valid() && (may_url_exist(input))) || s_search_engine_format == "{}")
```
alimpfard, line 60:
```suggestion
    if ((url.is_valid() && (may_url_exist(input))) || s_search_engine_format == "{}"sv)
```

Not really important, but let's not `strlen()` all the time : )

alimpfard (commented):
Looks okay to me as far as code goes.
The name `may_url_exist` sounds a bit weird, but I'm not entirely sure what it's supposed to do.

Userland/Applications/Browser/Tab.cpp
sppmacd: > The name `may_url_exist` sounds a bit weird, but I'm not entirely sure what it's supposed to do.

It checks if URL is likely to exist - there are no URLs with spaces, or without dots (except `localhost`). Invalid URLs are also treated as potential search queries. I didn't have a better idea for the name.
The next step would be checking if domain is valid (e.g. it has valid suffix).

alimpfard: I think it's fine, it just sounds odd is all.

awesomekling:
Instead of guessing whether the user wants to search and having to maintain a heuristic for this, let's just make it explicit.
Could be something as simple as "line starts with '?'"

Userland/Applications/Browser/Tab.cpp
```diff
@@ -38,8 +39,15 @@
 
 namespace Browser {
 
+static String s_search_engine_format = "{}";
```
linusg, line 42: I don't think we need an empty format string here, as there won't be any formatting when it's in this state. Let's just use a null string instead.

```suggestion
static String s_search_engine_format;
```

Userland/Applications/Browser/Tab.cpp
```diff
@@ -38,8 +39,15 @@
...
+
 URL url_from_user_input(const String& input)
 {
+    if (!input.is_empty() && input[0] == '?' && s_search_engine_format != "{}"sv) {
```
linusg, line 46:
```suggestion
    if (input.starts_with("?") && !s_search_engine_format.is_null()) {
```

Userland/Applications/Browser/Tab.cpp
```diff
@@ -38,8 +39,15 @@
...
+        dbgln("Using search engine for {}", input);
+        return URL(String::formatted(s_search_engine_format, input));
```
linusg, line 48: You still need to remove the leading `?`, no?

Userland/Applications/Browser/Tab.cpp
```diff
@@ -397,6 +405,37 @@ Tab::Tab(Type type)
...
+            },
+            this);
+        search_engine_menu.add_action(action);
+        m_search_engine_actions.add_action(action);
+    };
+
+    auto disable_search_engine_action = GUI::Action::create_checkable(
+        "Disable", [this](auto&) {
+            dbgln("Disabling search engine");
+            s_search_engine_format = "{}";
```
linusg, line 427:
```suggestion
            s_search_engine_format = {};
```

linusg (changes requested):

Userland/Applications/Browser/Tab.cpp
sppmacd: Fixed.

Userland/Applications/Browser/Tab.cpp
sppmacd: Fixed.

Userland/Applications/Browser/Tab.cpp
sppmacd: Fixed.

Userland/Applications/Browser/Tab.cpp
sppmacd: Fixed.

Userland/Applications/Browser/Tab.cpp
```diff
@@ -38,8 +39,14 @@
 
 namespace Browser {
 
+static String s_search_engine_format = {};
+
 URL url_from_user_input(const String& input)
 {
+    if (!input.starts_with("?") && !s_search_engine_format.is_null()) {
```
linusg, line 46: Logic error:

```suggestion
    if (input.starts_with("?") && !s_search_engine_format.is_null()) {
```

(please make sure to test your PR again when making changes!)

linusg (changes requested):

Userland/Applications/Browser/Tab.cpp
sppmacd: Fixed.

Userland/Applications/Browser/Tab.cpp
```diff
@@ -38,8 +39,14 @@
 
 namespace Browser {
 
+static String s_search_engine_format = {};
+
 URL url_from_user_input(const String& input)
 {
+    if (input.starts_with("?") && !s_search_engine_format.is_null()) {
+        return URL(String::formatted(s_search_engine_format, input.substring(1)));
```
linusg, line 47: I think it might be necessary to URL-encode the input as well, so it doesn't allow injection of additional query parameters etc.

```suggestion
        return URL(String::formatted(s_search_engine_format, urlencode(input.substring(1))));
```

sppmacd: Good idea :^)

PR #6586, opened 2021-04-23
Ports: Detect more types of errors in the AvailablePorts.md file

gunnarbeutner:
This adds support for detecting incorrect version numbers and links in the ports list.

Also, unlike before it doesn't parse the `package.sh` script but executes it instead which allows us to detect syntax errors.


Ports/nesalizer/package.sh
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env -S bash ../.port_include.sh
 port=nesalizer
-version=master
+version=git
```
linusg, line 3: needs a `workdir` entry now

gunnarbeutner: Added. :)

linusg:
❤️

PR #6588, opened 2021-04-23
Shell: jobspecs and builtin 'kill'

### LINKED ISSUE #6578: Shell: Job specs can not be passed to kill
For example:

```
$ js
< ^Z >
$ js
< ^Z >
$ jobs
[1] - stopped js
[2] - stopped js
$ fg 1
Resuming job 1 - js
< ctrl + d>
<EOF>

$ kill 2
kill: no such process
$ kill %2
'%2' is not a valid pid
```
https://www.gnu.org/software/bash/manual/bash.html#Job-Control-Builtins

Should be able to pass a `job-spec` to kill, wait, and disown.

cc @alimpfard

alimpfard:
Switches `fg`, `bg`, `disown`, `wait` and `kill` to use jobspecs.
Fixes #6578.

Note: since `%?foo` is a glob, you have to quote it, so `kill '%?sleep'`. I don't love that, but it's not too bad

### DIFF
--- a/Userland/Shell/Shell.cpp
+++ b/Userland/Shell/Shell.cpp
@@ -1806,11 +1806,16 @@ u64 Shell::find_last_job_id() const
     return job_id;
 }
 
-const Job* Shell::find_job(u64 id)
+const Job* Shell::find_job(u64 id, bool is_pid)
 {
     for (auto& entry : jobs) {
-        if (entry.value->job_id() == id)
-            return entry.value;
+        if (is_pid) {
+            if (entry.value->pid() == static_cast<int>(id))
+                return entry.value;
+        } else {
+            if (entry.value->job_id() == id)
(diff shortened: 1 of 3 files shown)


Userland/Shell/Shell.cpp
```diff
@@ -1194,6 +1194,27 @@ String Shell::unescape_token(const String& token)
     return builder.build();
 }
 
+String Shell::find_in_path(String program_name)
```
linusg, line 1197: `String const&`?

alimpfard: Actually, I think I'll make it take a `const StringView&` (and not a `StringView const&` - boo)

alimpfard: Done : )

PR #6590, opened 2021-04-23
HackStudio: Fixed a bug related to action tab widget

toottoot1:
When launching HackStudio, the action tab widget used to have a height
of 0. Also, when you change the height of a certain tab and change to 
another tab, the tab you changed to used to keep its previous scale.

Both of these issues got fixed.

### DIFF
--- a/Userland/DevTools/HackStudio/HackStudioWidget.cpp
+++ b/Userland/DevTools/HackStudio/HackStudioWidget.cpp
@@ -908,7 +908,7 @@ void HackStudioWidget::create_action_tab(GUI::Widget& parent)
         on_action_tab_change();
 
         static bool first_time = true;
-        if (!first_time)
+        if (first_time)
             m_action_tab_widget->set_fixed_height(200);
         first_time = false;
     };


linusg:
I feel like it's on purpose to have the tab content hidden initially, since we explicitly set `s_action_tab_widget->set_preferred_size(0, 24);` beforehand...?

toottoot1:
If you want to make the tab content hidden initially then you just have to remove the entirety of the if statement.

linusg:
@toottoot1 sure, if you think that's the best solution please go for it :)

Make sure to limit commit messages to 72 characters per line though, the CI build fails otherwise. Please also use a better description than "fixing a bug" - see the [contribution guidelines](https://github.com/SerenityOS/serenity/blob/master/CONTRIBUTING.md#code-submission-policy).

toottoot1:
Oh sorry, this is my first time contributing to someone's work, I'll be careful next time.
Do I have to close the request now in order for you to be able to merge it?

linusg:
no, just change the commit and force push to the branch :)

toottoot1:
I've just realized today that you're right, he wants it to be hidden at first but when you click on another tab it resizes the tab widget.
But as I've mentioned, if you have that line and you resize a certain tab (the Find in files tab for example) and you change to another tab (the Build tab for example) then the terminal inside the Build tab keeps it previous size.

PR #6591, opened 2021-04-23
 Meta: Re-use existing disk image where possible

gunnarbeutner:
This adds support for re-using and re-sizing existing disk images. Disk images are checked with `e2fsck` prior to re-use and a new disk image is automatically created when that check fails.

The PR changes the scripts to support using `rsync` to install the root filesystem. If `rsync` isn't available we fall back to using `cp`.

Using `rsync` is significantly faster when we're using an existing image because we only have to copy files which are new or have
been updated. This cuts down the image creation time to about a second or two assuming an old image exists.

### DIFF
--- a/Meta/build-image-qemu.sh
+++ b/Meta/build-image-qemu.sh
@@ -24,38 +24,77 @@ disk_usage() {
 }
 
 DISK_SIZE=$(($(disk_usage "$SERENITY_SOURCE_DIR/Base") + $(disk_usage Root) + 100))
+DISK_SIZE=${DISK_SIZE:-600}
+DISK_SIZE_BYTES=$((DISK_SIZE * 1024 * 1024))
+unset DISK_SIZE
 
-echo "setting up disk image..."
-qemu-img create _disk_image "${DISK_SIZE:-600}"m || die "could not create disk image"
-chown "$SUDO_UID":"$SUDO_GID" _disk_image || die "could not adjust permissions on disk image"
-echo "done"
+USE_EXISTING=0
 
-printf "creating new filesystem... "
-if [ "$(uname -s)" = "OpenBSD" ]; then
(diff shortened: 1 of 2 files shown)


gunnarbeutner:
Here's what the script does for a corrupted disk image:

```
[gunnar@nyx i686 (rsync-image)]$ dd if=/dev/zero of=_disk_image bs=1M seek=300 count=5
5+0 records in
5+0 records out
5242880 bytes (5.2 MB, 5.0 MiB) copied, 0.00307958 s, 1.7 GB/s
[gunnar@nyx i686 (rsync-image)]$ ninja image
[0/1] cd /home/gunnar/serenity/Build/i686 && /var/lib/snapd/snap/clion/149/bin/cmake/linux/bin/cmake -E env SERENITY_SOURCE_DIR=/home/gunnar/serenity SERENITY_ARCH=i686 /home/gunnar/serenity/Meta/build-image-qemu.sh
[fsck] checking existing image
[fsck] e2fsck 1.46.2 (28-Feb-2021)
[fsck] The filesystem size (according to the superblock) is 485632 blocks
[fsck] The physical size of the device is 78080 blocks
[fsck] Either the superblock or the partition table is likely to be corrupt!
[fsck] Abort? yes
[fsck] 
[fsck] failed, not using existing image
[image] setting up disk image... done
[mkfs] creating new filesystem... done
[mount] mounting filesystem... done
[install] installing base system... done
[install] creating initial filesystem structure... done
[install] creating utmp file... done
[install] setting up device nodes folder... done
[install] writing version file... done
[install] installing users... done
[install] adding some desktop icons...done
[install] installing shortcuts... done
[install] installing 'checksum' variants... done
[umount] unmounting filesystem... done
[gunnar@nyx i686 (rsync-image)]$ ```

PR #6593, opened 2021-04-23
UserspaceEmulator: Implement a dedicated FPU

Hendiadyoin1:
this is to separate it form the CPU, because it uses it own state machine
also this is more faithful to the original hardware implementation

this adds more support for more FloatingPoint control and comparisons and nearly all of the MMX extension

the current current roadblocks are: 
* ambiguity of some instructions with maybe implied arguments, that we do not handle separately  and I therefore don't know how to handle them
* it's very difficult to find out things like "rounding has happened" without loading the actual state word and operating in assembly
* there is no good way to differentiate QNaN from SNaN as far as I know

Still missing instructions:
* FBLDB
* FBSSTP
* FXTRACT
* FCOM[P]
* FUCOM[P[P]]
* FNENI
* FNDISI
* FNSETPM
* FPUNPCK*B // gcc wants weird types, and I was fed up with the 3 lines of casting
* MOVQ (long-mode)
Unimplemented Features:
* Rounding from fp to fp
* Full ShadowValue support
* SNan/QNaN support
* FPU-precision control
* Exception handling

Known Issues:
* LibM test:
  * hyperbolics are broken (similar in upstream)
  * gamma is really broken -> crash (it can go way further than upstream)
* 3DFileViewer looks.... empty (same on master)
* Cube-trinagles clip weird (same on master)
* Not full test coverage (please don't force me to do that)


awesomekling (changes requested):
Please spell check your code, comments and commit messages.
Also, please wrap your commit messages at 72 characters.

Hendiadyoin1:
Just fixed the few spelling mistakes I could find and fixed them, tell me if I missed any

also the maximum line length used in the commit messages is 66, and this is checked by the ci checks

Userland/DevTools/UserspaceEmulator/SoftFPU.cpp
```diff
@@ -0,0 +1,993 @@
+/*
+ * Copyright (c) 2020, Andreas Kling <kling@serenityos.org>
+ * Copyright (c) 2020, Leon Albrecht <loen2002.la@gmail.com>
```
awesomekling, line 3: It's 2021 now.

Userland/DevTools/UserspaceEmulator/SoftFPU.cpp
```diff
@@ -0,0 +1,993 @@
...
+void SoftFPU::FLDL2T(const X86::Instruction&)
+{
+    fpu_push(log2f(10.0f));
+}
+void SoftFPU::FLDLG2(const X86::Instruction&)
+{
+    fpu_push(log10f(2.0f));
+}
+
+// CONTROLL
```
awesomekling, line 1036: CONTROL

Userland/DevTools/UserspaceEmulator/SoftFPU.cpp
```diff
@@ -0,0 +1,993 @@
...
+// void PSLLW
+// void PSLLD
+// void PSRLW
+// void PSRLD
+// void PSRAW
+// void PSRAD
+// void PSLLQ
+// void PSRRQ
+
+// DATA TRANSFERE
```
awesomekling, line 1176: TRANSFER

Userland/DevTools/UserspaceEmulator/SoftFPU.h
```diff
@@ -0,0 +1,603 @@
+/*
+ * Copyright (c) 2020, Andreas Kling <kling@serenityos.org>
+ * Copyright (c) 2020, Leon Albrecht <loen2002.la@gmail.com>
```
awesomekling, line 3: It's 2021 now.

Userland/DevTools/UserspaceEmulator/SoftFPU.h
```diff
@@ -0,0 +1,603 @@
...
+    };
+
+    enum class RoundingMode : u8 {
+        NEAREST = 0x00,
+        DOWN = 0x01,
+        UP = 0x10,
+        TRUNK = 0x11
+    };
+
+    // FIXME: can we make this more a constexpr?
```
awesomekling, line 86: This function doesn't look like a good `constexpr` candidate. What am I missing?

Userland/DevTools/UserspaceEmulator/SoftFPU.h
```diff
@@ -0,0 +1,603 @@
...
+        set_c3(1);
+    }
+
+    void mmx_common()
+    {
+        // every mmx instrution (excluding EMMS) sets the fpu to `all valid`
+        m_fpu_tw = 0;
+    }
+
+    // FIXME: better MMx shadow checks
```
awesomekling, line 274: MMX

Userland/DevTools/UserspaceEmulator/SoftFPU.h
```diff
@@ -0,0 +1,603 @@
...
+    bool m_reg_is_mmx[8] { false };
+
+    // FIXME: Maybe change the structure of this
+    // do not pack, we need pointers from this
+    union {
+        long double m_st[8] { 0 };
+        struct
+        {
+            // this should represents *packed* data not an int
```
awesomekling, line 283: represents => represent

Userland/DevTools/UserspaceEmulator/SoftFPU.cpp
```diff
@@ -26,6 +26,15 @@
 
 namespace UserspaceEmulator {
 
+template<typename T>
```
awesomekling, line 29: (In commit message. complience => compliance)

Userland/DevTools/UserspaceEmulator/SoftFPU.cpp
```diff
@@ -744,20 +744,76 @@ void SoftFPU::FSQRT(const X86::Instruction&)
 void SoftFPU::FXTRACT(const X86::Instruction&) { TODO_INSN(); }
 
 // COMPARISON
+
```
awesomekling, line 747: (In commit message: suport => support)

Userland/DevTools/UserspaceEmulator/SoftFPU.cpp
```diff
@@ -810,8 +866,66 @@ void SoftFPU::FUCOMIP(const X86::Instruction& insn)
...
+        fpu_set_unordered();
+    else {
+        set_c0(fpu_get(0) < 0.);
+        set_c2(0);
+        set_c3(fpu_get(0) == 0.);
+    }
+}
+void SoftFPU::FXAM(const X86::Instruction&)
+{
+    // noexcept!
```
awesomekling, line 884: What does this mean?

Userland/DevTools/UserspaceEmulator/SoftFPU.cpp
```diff
@@ -951,7 +957,13 @@ void SoftFPU::FPTAN(const X86::Instruction&)
     fpu_set(0, tanl(fpu_get(0)));
     fpu_push(1.0f);
 }
-void SoftFPU::FPATAN(const X86::Instruction&) { TODO_INSN(); }
+void SoftFPU::FPATAN(const X86::Instruction&)
+{
+    // FIXME: set C1  on stackunderflow, or on rounding
```
awesomekling, line 962: stackunderflow => stack underflow

awesomekling (changes requested):
>also the maximum line length used in the commit messages is 66, and this is checked by the ci checks
or do you mean, I should wrap later in some cases?

Yes, please don't insert random line breaks in your commit messages and comments.
It makes your commits look different from everyone else's commits and it's much harder to read.

Also, please try to write both commit messages and comments with proper punctuation.
Your code changes are nice, but please make an effort with your human language as well. :)

Userland/DevTools/UserspaceEmulator/SoftFPU.h
Hendiadyoin1: its only called with constexpr Arguments, so its technically possible to optimize the if statements away, would ALWAYS_INLINE solve that?

Userland/DevTools/UserspaceEmulator/SoftFPU.cpp
Hendiadyoin1: this is a notice, that that function cannot raise a FPU Exception, shall I reword it or delete it?

Userland/DevTools/UserspaceEmulator/SoftFPU.cpp
Hendiadyoin1: fixed

Userland/DevTools/UserspaceEmulator/SoftFPU.h
Hendiadyoin1: fixed

Userland/DevTools/UserspaceEmulator/SoftFPU.cpp
Hendiadyoin1: fixed

Userland/DevTools/UserspaceEmulator/SoftFPU.cpp
Hendiadyoin1: oh.. And apparently I can't spell my mail correctly either

Userland/DevTools/UserspaceEmulator/SoftFPU.cpp
Hendiadyoin1: fixed

Userland/DevTools/UserspaceEmulator/SoftFPU.cpp
Hendiadyoin1: fixed

Userland/DevTools/UserspaceEmulator/SoftFPU.h
Hendiadyoin1: see above

Userland/DevTools/UserspaceEmulator/SoftFPU.h
Hendiadyoin1: fixed

Hendiadyoin1:
Ok after the other commit got merged I have to rebase, with that I'll sit down and implement the  new MMX instructions

awesomekling:
> Human languange is difficult... (english is really inconsistent with its writing system)

Please use a spell checker.

> In the first commit I chose to put the line break early, because otherwise it would be in the middle of the second part of the sentence and therefore unpleassant to read

Please format your commit messages in accordance with project style.

Hendiadyoin1:
@awesomekling I added MMX support now, I hope all commits are up to the standard and there aren't to many spelling mistakes;
If I need to squash things down, tell me, I know its a lot of small-ish committs.

Hendiadyoin1:
Sadly these still aren't tested, due to the same bug as before

Hendiadyoin1:
Now being able to test, I will need some time to verify all this...

awesomekling:
> @awesomekling I added MMX support now, I hope all commits are up to the standard and there aren't to many spelling mistakes;

Per the [contribution guidelines](https://github.com/SerenityOS/serenity/blob/773c17b6a0cb6cac1a40fccb6f285fe027497ff7/CONTRIBUTING.md):

>Write your commit messages in proper English, with care and punctuation.

The first commit message in this PR looks like this:

```
UserspaceEmulator: Start implementation of an FPU

this is to separate it form the cpu, because it uses its
own state machine

also this is more faithful to the original hardware implementation
and may make the code better structured for future iterations

also this adds the basics for a future mmx implementation
```

Here's what it would look like with spelling and punctuation fixed up:

```
UserspaceEmulator: Start implementation of an FPU

This is to separate it from the CPU, because it uses its own state
machine.

Also this is more faithful to the original hardware implementation
and may make the code better structured for future iterations.

Also this adds the basics for a future MMX implementation.
```

Please be respectful of other people's time. We put effort into our commit messages because thousands of people read them. :)

Dexesttp (commented):
I haven't dove into the code at all, but this looks fairly readable if tedious. It's a bit of a shame that the "Start implementation of a FPU" commit include both code move to a new file and new code added instead of being split in two, and that some commits go back-and-forth on the function declarations ("Inline all FPU helpers") or the mod() / m_rm bytes ("Add MMX Instructions " and "Cleanup use of ModRM byte"), but at this point it feels like it would be a hassle to rework the history and include them directly.

Here's some little things that I saw.

Userland/Libraries/LibC/stdio.cpp
```diff
@@ -657,6 +657,11 @@ char* fgets(char* buffer, int size, FILE* stream)
     return ok ? buffer : nullptr;
 }
 
+char* gets(char* buffer)
+{
+    return fgets(buffer, -1, stdout);
+}
+
```
sin-ack, line 664: Why is this even needed?

AK/Format.cpp
```diff
@@ -374,6 +374,53 @@ void FormatBuilder::put_f64(
...
+
+void FormatBuilder::put_f80(
+    long double value,
+    u8 base,
+    bool upper_case,
+    Align align,
+    size_t min_width,
+    size_t precision,
+    char fill,
+    SignMode sign_mode)
```
sin-ack, line 386: This is just `put_f64` with some types changed to `long double`. Is there a better way to do this that doesn't require copying code?

AK/Format.cpp
```diff
@@ -568,6 +615,29 @@ void Formatter<bool>::format(FormatBuilder& builder, bool value)
     }
 }
 #ifndef KERNEL
+void Formatter<long double>::format(FormatBuilder& builder, long double value)
```
sin-ack, line 618: Same as `put_f80`.

Userland/Libraries/LibX86/Types/u256.h
```diff
@@ -147,6 +147,11 @@ class u256 {
     }
 
     // bitwise
+    ALWAYS_INLINE constexpr u128 operator~() const
```
sin-ack: This should probably return a `u256` :^)

Userland/Libraries/LibX86/Types/Formatter.cpp
```diff
@@ -49,12 +50,13 @@ void AK::Formatter<u128>::format(AK::FormatBuilder& builder, u128 value)
...
 
-    u16 lower_length = sizeof(u64) * 0xFF / base;
-    if (m_width.value() > lower_length) {
-        builder.put_u64(value.high(), base, m_alternative_form, upper_case, m_zero_pad, m_align, m_width.value() - lower_length, m_fill, m_sign_mode);
-        builder.put_u64(value.low(), base, false, upper_case, m_zero_pad, m_align, m_width.value(), m_fill, m_sign_mode);
+    u16 lower_length = sizeof(u64) * 8 / base;
+    auto width = m_width.value_or(0);
+    if (value.high()) {
+        builder.put_u64(value.high(), base, m_alternative_form, upper_case, m_zero_pad, m_align, max(width - lower_length, 0ul), m_fill, m_sign_mode);
+        builder.put_u64(value.low(), base, false, upper_case, m_zero_pad, m_align, width, m_fill, m_sign_mode);
```
sin-ack, line 57: Please explain what exactly this fixes in the commit message. "Fix" doesn't explain what was fixed and why.

Userland/DevTools/UserspaceEmulator/SoftCPU.h
```diff
@@ -1153,7 +1163,10 @@ class SoftCPU final
     void update_code_cache();
 
 private:
+    friend class SoftFPU;
```
sin-ack, line 1170: `class` isn't needed here because you include `"SoftFPU.h"`.

Userland/DevTools/UserspaceEmulator/SoftFPU.cpp
```diff
@@ -0,0 +1,994 @@
...
+        reportln("\n=={}== Unimplemented instruction: {}\n", getpid(), __FUNCTION__); \
+        m_emulator.dump_backtrace();                                                  \
+        _exit(0);                                                                     \
+    } while (0)
+
+template<typename T>
+ALWAYS_INLINE void warn_if_uninitialized(T value_with_shadow, const char* message)
+{
+    if (value_with_shadow.is_uninitialized()) [[unlikely]] {
+        reportln("\033[31;1mWarning! Use of uninitialized value: {}\033[0m\n", message);
```
sin-ack, line 34: Usage of ANSI escapes directly in `reportln` feels a little bit iffy.

Userland/DevTools/UserspaceEmulator/SoftFPU.cpp
```diff
@@ -0,0 +1,994 @@
...
+        UserspaceEmulator::Emulator::the().dump_backtrace();
+    }
+}
+
+namespace UserspaceEmulator {
+
+long double SoftFPU::fpu_get(u8 index) const
+{
+    VERIFY(index < 8);
```
sin-ack: Can the 8 be put into a `constexpr` variable instead of being a magic number?

Userland/DevTools/UserspaceEmulator/SoftFPU.h
```diff
@@ -0,0 +1,665 @@
...
+    {
+        set_tag_from_value_absolute((m_fpu_stack_top + index) % 8, val);
+    }
+
+    bool fpu_isnan(u8 index) const
+    {
+        return (fpu_get_tag(index) == FPU_Tag::Special) && isnan(fpu_get(index));
+    }
+
+    bool fpu_is_set(u8 index) const
```
sin-ack: `fpu_is_set` sounds a bit awkward. How about `has_value_at`?

Userland/DevTools/UserspaceEmulator/SoftFPU.h
```diff
@@ -0,0 +1,665 @@
...
+
+#include <math.h>
+#include <string.h>
+
+namespace UserspaceEmulator {
+
+class Emulator;
+class SoftCPU;
+
+class SoftFPU final {
```
sin-ack, line 35: Almost everything in this class is `fpu_` prefixed. You are already defining the methods in the `SoftFPU` class, you do not need to prefix them.

Userland/DevTools/UserspaceEmulator/SoftFPU.h
```diff
@@ -0,0 +1,665 @@
...
+    u16 m_fpu_cs { 0 };
+
+    u32 m_fpu_dp { 0 };
+    u16 m_fpu_ds { 0 };
+
+    u16 m_fpu_iop { 0 };
+
+    // Instructions
+
+    // DATA TRANSFER
```
sin-ack, line 379: The comments are screaming at me. Please use normal casing.

Userland/DevTools/UserspaceEmulator/SoftFPU.h
```diff
@@ -0,0 +1,665 @@
...
+    void FLDL2E(const X86::Instruction&);
+    void FLDLN2(const X86::Instruction&);
+    void FLDL2T(const X86::Instruction&);
+    void FLDLG2(const X86::Instruction&);
+
+    // CONTROL
+    void FINCSTP(const X86::Instruction&);
+    void FDECSTP(const X86::Instruction&);
+    void FFREE(const X86::Instruction&);
+    void FFREEP(const X86::Instruction&); // undocumented
```
sin-ack, line 517: This comment seems unnecessary.

Userland/Libraries/LibX86/Instruction.cpp
```diff
@@ -592,7 +592,7 @@ static void build_slash_reg(u8 op, u8 slash, const char* mnemonic, InstructionFo
     // FIXME: DE/7 F9 (...but isn't this what DE/7 does naturally, with F9 just being normal R/M?)
 
     build_slash(0xDF, 0, "FILD", OP_FPU_RM32, &Interpreter::FILD_RM16);
-    build_slash_reg(0xDF, 0, "FFREEP", OP_FPU_reg, &Interpreter::FFREEP);
+    build_slash_reg(0xDF, 0, "FFREEP", OP_FPU_reg, &Interpreter::FFREEP); // undocumented
```
sin-ack, line 597: This comment seems unnecessary.

Userland/DevTools/UserspaceEmulator/SoftFPU.h
```diff
@@ -0,0 +1,665 @@
...
+            // long doubles are weird with their supposed sizes, we assume, that
+            // gcc chooses either 12 (x32) or 16 (x64) byte as its size,
+            // if neither is the case we assume for now, that they really are
+            // 10 bytes big, as they are supposed to be
+            Conditional<sizeof(long double) == 16,
+                u64,
+                Conditional<sizeof(long double) == 12,
+                    u32,
+                    u16>>
+                __high;
```
sin-ack, line 313: Uh... is this clang-format's doing? :sweat_smile: This looks a bit weird, can we unroll this with a `using`?

Userland/DevTools/UserspaceEmulator/SoftFPU.cpp
```diff
@@ -0,0 +1,994 @@
...
+
+void SoftFPU::mmx_set(u8 index, u64 value)
+{
+    m_reg_is_mmx[index] = true;
+    m_mmx[index].value = value;
+}
+
+void SoftFPU::fpu_push(long double value)
+{
+    if (fpu_is_set(7))
```
sin-ack, line 108: More magic numbers. Something like `SoftFPU::STACK_TOP - 1` would be nice.

Userland/DevTools/UserspaceEmulator/SoftFPU.cpp
```diff
@@ -107,6 +109,59 @@ long double SoftFPU::fpu_pop()
     return ret;
 }
 
+void SoftFPU::fpu_set_exception(FPU_Exception ex)
```
sin-ack: Please squash these changes. There is no reason for this change to be separate from the initial FPU patch.

Userland/DevTools/UserspaceEmulator/SoftFPU.cpp
```diff
@@ -1319,12 +1319,13 @@ void SoftFPU::PSRLQ_mm1_mm2m64(const X86::Instruction&) { TODO_INSN(); }
...
-// void MOVD // (ld from mem)
-// void MOVD // (st to mem)
+void SoftFPU::MOVD_mm1_rm32(const X86::Instruction&) { TODO_INSN(); };
+void SoftFPU::MOVD_rm32_mm2(const X86::Instruction&) { TODO_INSN(); };
 
 void SoftFPU::MOVQ_mm1_mm2m64(const X86::Instruction&) { TODO_INSN(); }
-void SoftFPU::MOVQ_mm1_m64_mm2(const X86::Instruction&) { TODO_INSN(); }
+void SoftFPU::MOVQ_mm1m64_mm2(const X86::Instruction&) { TODO_INSN(); }
+void SoftFPU::MOVQ_mm1_rm64(const X86::Instruction&) { TODO_INSN(); }; // long mode
+void SoftFPU::MOVQ_rm64_mm2(const X86::Instruction&) { TODO_INSN(); }; // long mode
```
sin-ack: This patch should also be a part of the initial FPU patch, since it doesn't do anything beyond stubbing out the MMX instructions.

sin-ack (changes requested):
Great work for such a big change, but it needn't be split across so many commits. Many of the commits just build on the initial FPU commit and should be squashed into a single commit that splits the CPU from the FPU.

Userland/Libraries/LibM/math.h
```diff
@@ -29,6 +29,21 @@ __BEGIN_DECLS
 #define NAN __builtin_nan("")
 #define MAXFLOAT FLT_MAX
 
+// ruthlessly copied these from glibc
+#define M_El 2.718281828459045235360287471352662498L        /* e */
```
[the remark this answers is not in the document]
Hendiadyoin1: See comment...

I didn't really notice them, I can remove them whe I get to working on it again.

Userland/Libraries/LibC/stdio.cpp
Hendiadyoin1: I needed this for a programm that I used for testing, that had questionable Copyrights attached though and failed the ci, so I removed the test, but left the patches

shall I make this another pull request?

AK/Format.cpp
Hendiadyoin1: I can make it a
```cpp
template<FloatingPoint T>
void FormatBuilder::put_float(
    T value,
    ...
)
{
    ...
}
```
instead, that might be a better idea even

Userland/Libraries/LibX86/Types/u256.h
Hendiadyoin1: That is indeed the case, will be fixed when I get to coding next

Userland/Libraries/LibX86/Types/Formatter.cpp
Hendiadyoin1: So elaborate in the comit message?
Ok will do.

Will be something like:
> In the past we didn't calculate the part_length correctly which led to us only printing the lower half of the Value,
also we took the length argument as granted, which is more than often not the case,
in addition to that the lenght argument isn't treated as a upperbound anymore

Userland/DevTools/UserspaceEmulator/SoftCPU.h
Hendiadyoin1: Good to know

Userland/DevTools/UserspaceEmulator/SoftFPU.cpp
Hendiadyoin1: It the same in the SoftCPU, this is a carbon copy of that...

Userland/DevTools/UserspaceEmulator/SoftFPU.cpp
Hendiadyoin1: I have conflicting feelings about that, this limit is the size of the stack and should always be 8 and making it a variable, even a constexpr one will make it seem like it can be changed, which is not possible on a hardware level - there are only 3 bits reserved to index into this- so maybe not?

Userland/DevTools/UserspaceEmulator/SoftFPU.h
Hendiadyoin1: It can have a value even if it isn't set though...

Userland/DevTools/UserspaceEmulator/SoftFPU.h
Hendiadyoin1: this is to differentiate the x87 part and the mmx part of this class, so I see your point, but there is a reason for it though

Userland/DevTools/UserspaceEmulator/SoftFPU.h
Hendiadyoin1: I formatted it this way...
and the reason for it looking weird is because it is weird.... (thank gcc for that)
> long doubles are weird with their supposed sizes

AK/Format.cpp
Hendiadyoin1: GCC does not seem very cooperative today,
If you get it working, just make a pull request to my repo and I will merge it then

Userland/DevTools/UserspaceEmulator/SoftFPU.h
Hendiadyoin1: >  can we unroll this with a using?

idk, how would you do it?

Userland/DevTools/UserspaceEmulator/SoftFPU.cpp
Hendiadyoin1: I think I should separate changes to different parts of the code as separated as possible, and this is a change Mainly for LibX86 and does not fully concern the Emulator

Hendiadyoin1:
After the rebase one of the tests seems uncooperative, and the test seems to fail on native mashines too (Works for me with `-O0 -m32` and fails with `-O1 -m32` and higher, failed on @gunnarbeutner's machine on default settings (`-m32`))
the test in question is the environment store and load test

nico:
FWIW I looked this over and it looks roughly ok to me (I didn't look in detail).

I guess this would be easier to merge if it was several pull requests (eg one that just moves stuff to a new file without any other changes, and then most of the rest in a separate one).

Hendiadyoin1:
This will be closed in favor of #8610

PR #6598, opened 2021-04-24
ProtocolServer+WebSockets: Rename ProtocolServer to DownloadServer, add an independent WebSocket service

Dexesttp:
# Overview

This PR:
- Renames `ProtocolServer` to `DownloadServer`
- Adds a new `WebSocket` service
- Updates LibProtocol so it can handle connecting to both `DownloadServer` (current behavior) and `WebSocket` (new behavior).

Note : While it seems to make sense to rename DownloadServer, given that we now have several kinds of isolated instances behind LibProtocol (and that ProtocolServer only handles downloads), that commit can easily be removed if not wanted.


Dexesttp:
Superseded by #6610

PR #6600, opened 2021-04-24
Kernel: Consolidate Scatter gather list implementations

ccapitalK:
Currently, we have two scatter gather lists in the kernel, one in the AHCI subsystem and one in the VirtIO subsystem. This commit consolidates them into `VM/ScatterGatherList.h`.

### DIFF
--- a/Kernel/VM/ScatterGatherList.h
+++ b/Kernel/VM/ScatterGatherList.h
@@ -7,21 +7,40 @@
 #pragma once
 
 #include <AK/Vector.h>
+#include <Kernel/Devices/BlockDevice.h>
 #include <Kernel/PhysicalAddress.h>
+#include <Kernel/VM/AnonymousVMObject.h>
 #include <Kernel/VM/MemoryManager.h>
 
 namespace Kernel {
 
-class ScatterGatherList {
-    struct ScatterGatherEntry {
+/// A Scatter-Gather List type that owns its buffers
+
+class ScatterGatherList : public RefCounted<ScatterGatherList> {
+public:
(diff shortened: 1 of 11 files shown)


ccapitalK:
By request of @bgianfo , I have made ScatterGatherList's API not assume that allocations will succeed.

IdanHo:
cc @supercomputer7 for approval

Kernel/VM/ScatterGatherList.h
```diff
@@ -7,29 +7,48 @@
...
+    ScatterGatherList(AsyncBlockDeviceRequest&, NonnullRefPtrVector<PhysicalPage> allocated_pages, size_t device_block_size);
+    NonnullRefPtr<AnonymousVMObject> m_vm_object;
+    OwnPtr<Region> m_dma_region;
+};
+
+/// A Scatter-Gather List type that doesn't own its buffers
+
+class ScatterGatherRefList {
+    struct ScatterGatherRef {
         FlatPtr page_base;
```
supercomputer7: If we are using a buffer of someone else, we also own it. Why don't we use `ScatterGatherList` all the time?
In general, I'd prefer that we create a `ScatterGatherList` only when we need it like how it's done within the `AHCIPort`.
Is there a way to tell the VirtIO device to fetch data from a non-fixed place each time we need to get/send data from/to the device?

ccapitalK:
After some more work on this, I've realised that the current approach to virtqueues is a bit brittle and won't really work well for more complex devices (especially ones which don't always consume virtqueue buffers in FIFO order). I'm going to close this PR, I will soon create another PR using a new approach that I'm trying to do that allows allocation-less insertion into the virtqueue, a strategy for concurrent access and doesn't require the buffers to be arranged in a ring at all times (which gives us back one of the degrees of freedom the virtio spec provided).

Issue #6474, opened 2021-04-19
Kernel: Running test-pthread repeatedly leaks processes


ADKaster:
This can cause the system to hang. Hitting Andreas's new Alt+Shift+F12 hotkey shows that there are two Threads that are `Dead`, but the Finalizer is not being scheduled.

See below in the comments for how it was narrowed down to test-pthread

Tests were run using branch https://github.com/SerenityOS/serenity/pull/6459

```
$ loop { ./tests/run-tests-and-shutdown.sh }
```

![hung](https://user-images.githubusercontent.com/8388494/115169851-6e309a80-a07c-11eb-94bd-dfe4723dc476.png)

Debug console log below vvvvv

<details>

<summary>Debug Console Log</summary>

```
Shell(921): Job entry 'echo ==== Running test 47 out of 88 -- /usr/Tests/AK/TestHex ====' deleted in 440 ms
Shell(921): Job entry 'echo ::group::/usr/Tests/AK/TestHex' deleted in 218 ms
TestHex(1409): Completed test 'should_decode_hex_digit' in 0ms
TestHex(1409): Completed test 'should_constexpr_decode_hex_digit' in 0ms
TestHex(1409): Finished 2 tests and 0 benchmarks in 27ms (0ms tests, 0ms benchmarks, 27ms other).
TestHex(1409): Out of 2 tests, 2 passed and 0 failed.
Shell(921): Job entry '/usr/Tests/AK/TestHex' deleted in 432 ms
[#0 colonel(0:0)]:   Selecting      Pr:10 DHCPClient(13:13)              @ 001b:294a71bb (nsched: 142)
[#0 colonel(0:0)]:   Queue          Pr:30 IO WorkQueue(1:1)              @ 0008:c0551ac5 (nsched: 6130)
[#0 colonel(0:0)]:   Running        Pr: 1 colonel(0:0)                   @ 0008:c0390567 (nsched: 140656)
[#0 colonel(0:0)]:   Futex          Pr:30 AudioServer(10:21)             @ 001b:575a21bb (nsched: 4)
[#0 colonel(0:0)]:   Selecting      Pr:50 WindowServer(12:12)            @ 001b:9f0011bb (nsched: 22776)
[#0 colonel(0:0)]:   Selecting      Pr:30 Shell(22:22)                   @ 001b:7e2351bb (nsched: 1037)
[...]

PR #6601, opened 2021-04-24
Kernel: Fix LOCK_DEBUG tracking to work again, move it to use AK::SourceLocation

bgianfo:
There are a hand full of changes here,  all around improving the `LOCK_DEBUG` facilities / API.

- __Kernel: Fix LOCK_DEBUG feature to work again__

    - UBSAN detected cases where we were calling current_thread->holding_lock
      but current_thread was nullptr.

    - Fix Lock::force_unlock_if_locked to not pass the correct ref delta to
      holding_lock(..).

-  __Kernel: Add lock_count to procfs$all when LOCK_DEBUG is enabled.__

- __AK: Add default constructor to SourceLocation__

- __Kernel: Utilize AK::SourceLocation for LOCK_DEBUG instrumentation.__

    The previous `LOCKER(..)` instrumentation only covered some of the
    cases where a lock is actually acquired. By utilizing the new
    `AK::SourceLocation` functionality we can now reliably instrument
    all calls to lock automatically.

    Other changes:
    - Tweak the message in `Thread::finalize()` which dumps leaked lock
      so it's more readable and includes the function information that is
      now available.

    - Make the `LOCKER(..)` define a no-op, it will be cleaned up in a
      follow up change.

- __Kernel: Remove the now defunct `RESTORE_LOCK(..)` macro.__

- __Kernel: Remove the now defunct `LOCKER(..)` macro.__

### DIFF
--- a/AK/SourceLocation.h
+++ b/AK/SourceLocation.h
@@ -23,6 +23,8 @@ public:
         return SourceLocation(file, line, function);
     }
 
+    constexpr SourceLocation() = default;
+
 private:
     constexpr SourceLocation(const char* const file, u32 line, const char* const function)
         : m_function(function)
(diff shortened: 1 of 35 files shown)


bgianfo:
This is WIP until we have enough data to diagnose https://github.com/SerenityOS/serenity/issues/6474

PR #6602, opened 2021-04-24
LibSQL: Parse INSERT, UPDATE, and ALTER TABLE statements

trflynn89:
Phew! That takes care of parsing the big 7: CREATE, ALTER, DROP, INSERT, UPDATE, DELETE, and SELECT. This seems like a good point to stop parsing and move onto the execution steps. Most notable SQL features that are not yet parsed are JOINs, TRANSACTIONs, TRIGGERs, and functions, but I think we can make do with the basic operations for now.


Userland/Libraries/LibSQL/AST.h
```diff
@@ -739,6 +815,110 @@ class DropTable : public Statement {
...
+    ConflictResolution conflict_resolution() const { return m_conflict_resolution; }
+    const String& schema_name() const { return m_schema_name; }
+    const String& table_name() const { return m_table_name; }
+    const String& alias() const { return m_alias; }
+    const Vector<String>& column_names() const { return m_column_names; }
+
+    bool default_values() const { return !has_expressions() && !has_selection(); };
+
+    bool has_expressions() const { return !m_chained_expressions.is_empty(); }
+    const NonnullRefPtrVector<ChainedExpression> chained_expressions() const { return m_chained_expressions; }
```
linusg, line 870:
```suggestion
    const NonnullRefPtrVector<ChainedExpression>& chained_expressions() const { return m_chained_expressions; }
```

trflynn89: :facepalm: good catch. I found 2 other instances of a missing `&` so I patched those up as well.

PR #6603, opened 2021-04-24
Utilities: Add cksum

bcoles:
`cksum` uses checksum algorithms and is compliant with [POSIX cksum](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/cksum.html) with some additional support for choosing a checksum algorithm. At least it would be compliant with the POSIX implementation; however, it uses CRC32 by default, not the CRC error checking in the ISO/IEC 8802-3:1996 standard (Ethernet) which LibCrypto does not yet support.


Userland/Utilities/CMakeLists.txt
```diff
@@ -19,6 +19,7 @@ target_link_libraries(aplay LibAudio)
 target_link_libraries(avol LibAudio)
 target_link_libraries(bt LibSymbolClient)
 target_link_libraries(checksum LibCrypto)
+target_link_libraries(cksum LibCrypto)
 target_link_libraries(chres LibGUI)
```
linusg, line 23:
```suggestion
target_link_libraries(chres LibGUI)
target_link_libraries(cksum LibCrypto)
```

Userland/Utilities/cksum.cpp
```diff
@@ -0,0 +1,63 @@
...
+#include <LibCrypto/Checksum/Adler32.h>
+#include <LibCrypto/Checksum/CRC32.h>
+
+int main(int argc, char** argv)
+{
+    Vector<const char*> paths;
+    const char* opt_algorithm = nullptr;
+
+    Core::ArgsParser args_parser;
+    args_parser.add_option(opt_algorithm, "Checksum algorithm (default 'crc32') ('list' to list available algorithms)", "algorithm", '\0', nullptr);
```
linusg, line 18:
```suggestion
    args_parser.add_option(opt_algorithm, "Checksum algorithm (default 'crc32', use 'list' to list available algorithms)", "algorithm", '\0', nullptr);
```

Userland/Utilities/cksum.cpp
```diff
@@ -0,0 +1,63 @@
...
+    args_parser.add_positional_argument(paths, "File", "file", Core::ArgsParser::Required::No);
+    args_parser.parse(argc, argv);
+
+    auto algorithm = (opt_algorithm == nullptr) ? "crc32" : StringView(opt_algorithm).to_string().to_lowercase();
```
linusg, line 22:
```suggestion
    auto algorithm = (opt_algorithm == nullptr) ? "crc32" : String(opt_algorithm).to_lowercase();
```

(`StringView::to_string()` is just a wrapper around the `String` constructor anyway)

linusg (changes requested):

PR #6604, opened 2021-04-24
AK: Add SourceLocation support

bgianfo:
C++20 added std::source_location, which lets you capture the
callers __FILE__ / __LINE__ / __FUNCTION__ etc as a default
argument to functions.
See: https://en.cppreference.com/w/cpp/utility/source_location

During a bug investigation @ADKaster suggested we could use this
to make the LOCK_DEBUG feature of the kernel more user friendly
and allow it to automatically instrument all call sites.

We then implemented / tested it over discord. :^)

Co-Authored-by: Andrew Kaster <andrewdkaster@gmail.com>

### DIFF
--- a/AK/SourceLocation.h
+++ b/AK/SourceLocation.h
@@ -1,0 +1,41 @@
+/*
+ * Copyright (c) 2021, Andrew Kaster <andrewdkaster@gmail.com>
+ * Copyright (c) 2021, Brian Gianforcaro <bgianf@serenityos.org>
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+#pragma once
+
+#include <AK/StringView.h>
+#include <AK/Types.h>
+
+namespace AK {
+
+class SourceLocation {
+public:
+    [[nodiscard]] constexpr StringView function_name() const { return StringView(m_function); }
+    [[nodiscard]] constexpr StringView file_name() const { return StringView(m_file); }
+    [[nodiscard]] constexpr u32 line_number() const { return m_line; }
+
+    [[nodiscard]] static constexpr SourceLocation current(const char* const file = __builtin_FILE(), u32 line = __builtin_LINE(), const char* const function = __builtin_FUNCTION())
+    {
+        return SourceLocation(file, line, function);
+    }
+
+private:
+    constexpr SourceLocation(const char* const file, u32 line, const char* const function)
+        : m_function(function)
+        , m_file(file)
+        , m_line(line)
+    {
+    }
+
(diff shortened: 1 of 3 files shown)


AK/SourceLocation.h
```diff
@@ -0,0 +1,41 @@
...
+class SourceLocation {
+public:
+    [[nodiscard]] constexpr StringView function_name() const { return StringView(m_function); }
+    [[nodiscard]] constexpr StringView file_name() const { return StringView(m_file); }
+    [[nodiscard]] constexpr u32 line_number() const { return m_line; }
+
+    [[nodiscard]] static constexpr SourceLocation current(const char* const file = __builtin_FILE(), u32 line = __builtin_LINE(), const char* const function = __builtin_FUNCTION())
+    {
+        return SourceLocation(file, line, function);
+    }
```
linusg, line 24: I don't think having default arguments makes much sense here, otherwise you could make `current()` return something completely *not current*?

Suggestion:

```suggestion
    [[nodiscard]] static constexpr SourceLocation current()
    {
        return SourceLocation(__builtin_FILE(), __builtin_LINE(), __builtin_FUNCTION());
    }
```

for anything that should return a mix of current and custom values, you could still use the regular constructor with `__builtin_*()`.

AK/Tests/TestSourceLocation.cpp
```diff
@@ -0,0 +1,32 @@
...
+#include <AK/SourceLocation.h>
+#include <AK/StringView.h>
+#include <AK/TestSuite.h>
+
+TEST_CASE(basic_scenario)
+{
+    auto location = SourceLocation::current();
+    EXPECT_EQ(StringView(__FILE__), location.file_name());
+    EXPECT_EQ(StringView(__FUNCTION__), location.function_name());
+    EXPECT(location.line_number() != 0);
```
linusg, line 17: can we be more specific here and use the exact number?

linusg (commented):
> We the then implemented / tested it over discord. :^)

Very cool!

AK/SourceLocation.h
asynts: That won't work because file will always be `SourceLocation.h` and line will alwasy be `23`. `__builtin_FILE` / `__builtin_LINE` gives you the current source location except if you use it as default argument.

linusg: oh duh, you're totally right :D

AK/Tests/TestSourceLocation.cpp
bgianfo: sure

PR #6606, opened 2021-04-24
AK: Make AK::Vector expose allocation failures in API

ccapitalK:
We re-write (almost) every operation in AK::Vector that performs allocations to return a boolean stating whether it succeeded it or not, so that allocation failures may be handled. All of these new operations are named `try_<operation_name>`. The old operation names are now thin wrappers that `VERIFY()` the result of the fallible operation.

The only issues are things like copy constructors, where we can't easily expose whether or not the allocation failed (without using something like exceptions, yuck!).

This should help us accomplish #6369 , with this implemented we should now be able to start fixing kernel components that use vectors to make sure they check their allocations. I think in the long term, we should eventually only make the wrappers and copy constructors exist in userspace AK, they should be disabled in kernel space AK so that components have to check their allocations. However, that is probably a ways away.

I would like some people who are good with c++ and templates to look over this, I'm not sure If I did everything right (especially with regards to function overloads, variadic templates and argument forwarding).

### DIFF
--- a/AK/Vector.h
+++ b/AK/Vector.h
@@ -259,12 +259,14 @@ public:
     }
 
     template<typename U = T>
-    void insert(size_t index, U&& value)
+    [[nodiscard]] bool try_insert(size_t index, U&& value)
     {
-        VERIFY(index <= size());
+        if (index > size())
+            return false;
         if (index == size())
-            return append(forward<U>(value));
-        grow_capacity(size() + 1);
+            return try_append(forward<U>(value));
+        if (!try_grow_capacity(size() + 1))
+            return false;
         ++m_size;
         if constexpr (Traits<T>::is_trivial()) {
             TypedTransfer<T>::move(slot(index + 1), slot(index), m_size - index - 1);
@@ -275,22 +277,40 @@ public:
             }
         }
         new (slot(index)) T(forward<U>(value));
+        return true;
+    }
+
+    template<typename U = T>
+    void insert(size_t index, U&& value)
+    {
+        auto did_allocate = try_insert<U>(index, forward<U>(value));
+        VERIFY(did_allocate);
+    }
+
+    template<typename C, typename U = T>
+    [[nodiscard]] bool try_insert_before_matching(U&& value, C callback, size_t first_index = 0, size_t* inserted_index = nullptr)
+    {
+        for (size_t i = first_index; i < size(); ++i) {
+            if (callback(at(i))) {
+                if (!try_insert(i, forward<U>(value)))
+                    return false;
+                if (inserted_index)
+                    *inserted_index = i;
+                return true;
+            }
+        }
+        if (!try_append(forward<U>(value)))
+            return false;
+        if (inserted_index)
+            *inserted_index = size() - 1;
+        return true;
     }
 
     template<typename C, typename U = T>
     void insert_before_matching(U&& value, C callback, size_t first_index = 0, size_t* inserted_index = nullptr)
     {
-        for (size_t i = first_index; i < size(); ++i) {
-            if (callback(at(i))) {
-                insert(i, forward<U>(value));
-                if (inserted_index)
-                    *inserted_index = i;
-                return;
-            }
-        }
-        append(forward<U>(value));
-        if (inserted_index)
-            *inserted_index = size() - 1;
+        auto did_allocate = try_insert_before_matching(forward<U>(value), callback, first_index, inserted_index);
+        VERIFY(did_allocate);
     }
 
     Vector& operator=(const Vector& other)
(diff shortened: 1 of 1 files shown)


AK/Vector.h
```diff
@@ -275,22 +277,38 @@ class Vector {
...
+            return false;
         if (inserted_index)
             *inserted_index = size() - 1;
+        return true;
+    }
+
+    template<typename C, typename U = T>
+    void insert_before_matching(U&& value, C callback, size_t first_index = 0, size_t* inserted_index = nullptr)
+    {
+        VERIFY(try_insert_before_matching(forward<U&&>(value), callback, first_index, inserted_index));
```
bgianfo, line 311: Putting code inside `VERIFY(..)` makes me nervous, as someone will inevitably try to disable the VERIFY macro to do nothing for "performance" in the future.
We might save some headaches if we just don't put code we need to run in there?

bgianfo (approved):

AK/Vector.h
awesomekling: `VERIFY` will never be disabled. Instead we should add a new `ASSERT` for things that can be disabled. :)

That said, I agree that we should not do effectful things in `VERIFY` as it's much harder to read.

ccapitalK: What would be the preferred way to do it? I could redo them like:

```
auto did_allocate = try_insert_before_matching(forward<U&&>(value), callback, first_index, inserted_index);
VERIFY(did_allocate);
```

awesomekling: Yes, that would be better.

ccapitalK: Alright, I've changed it to do that throughout.

AK/Vector.h
```diff
@@ -389,43 +425,73 @@ class Vector {
...
+
+    [[nodiscard]] ALWAYS_INLINE bool try_append(const T& value)
+    {
+        return try_append(T(value));
     }
 
     ALWAYS_INLINE void append(const T& value)
     {
         append(T(value));
     }
```
Dexesttp, line 467: For consistency, I think this could be changed to :
```
        auto did_allocate = try_append(value);
        VERIFY(did_allocate);
```

AK/Vector.h
```diff
@@ -503,20 +607,37 @@ class Vector {
...
+
+    [[nodiscard]] bool try_resize_and_keep_capacity(size_t new_size)
+    {
+        return try_resize(new_size, true);
     }
 
     void resize_and_keep_capacity(size_t new_size)
     {
-        return resize(new_size, true);
+        auto did_allocate = try_resize(new_size, true);
```
Dexesttp, line 639: Idem here :
```suggestion
        auto did_allocate = try_resize_and_keep_capacity(new_size);
```

Dexesttp (commented):
That all looks good - I didn't see any glaring issue or unmoved stuff that needed to be moved !

Just maybe a couple changes that could be interesting : for consistency, making it so the `VERIIFY()` versions of the methods actually redirect to the proper `try_` version ?
There's no behavior changes for the ones that don't (since they redirect to the smae code via a different path), but that should help prevent bugs if one were to actually change the `try_` implementation while keeping a minimal (and/or optimized away) overhead.

AK/Vector.h
```diff
@@ -275,22 +277,40 @@ class Vector {
...
+            return false;
         if (inserted_index)
             *inserted_index = size() - 1;
+        return true;
+    }
+
+    template<typename C, typename U = T>
+    void insert_before_matching(U&& value, C callback, size_t first_index = 0, size_t* inserted_index = nullptr)
+    {
+        auto did_allocate = try_insert_before_matching(forward<U&&>(value), callback, first_index, inserted_index);
```
Dexesttp, line 312:
```suggestion
        auto did_allocate = try_insert_before_matching(forward<U>(value), callback, first_index, inserted_index);
```

AK/Vector.h
```diff
@@ -275,22 +277,40 @@ class Vector {
             }
         }
         new (slot(index)) T(forward<U>(value));
+        return true;
+    }
+
+    template<typename U = T>
+    void insert(size_t index, U&& value)
+    {
+        auto did_allocate = try_insert<U>(index, forward<U&&>(value));
```
Dexesttp, line 286:
```suggestion
        auto did_allocate = try_insert<U>(index, forward<U>(value));
```

Dexesttp (commented):
Oh! Sorry, I missed these. These are not correct.
`forward<U>(value)` actually wants to take an `U&&` as input, so `forward<U&&>` would take an `(U&&)&&` or something. I'm actually _not_ sure whether these two are equivalent so `forward<U&&>` magically works, or would fail on some cases for some weird reasons, I'm not good enough at understanding rvalues and forwarding to know that kind of subtelty ^^'
Anyways, what you want there is `forward<U>(value)`.

AK/Vector.h
ccapitalK: Fixed

AK/Vector.h
ccapitalK: Fixed

AK/Vector.h
ccapitalK: I tried this originally, but encountered some strange template errors. If you give me a few moments, I'll try to get the error messages.

Dexesttp: Weird. I try compiling before making the suggestion (because I was trying to understand why you did it this way), and it seemed to work.
Please post the error messages if you get any :)

ccapitalK: Ok, it appears to compile fine now. I originally wrote it as `forward<U>(value)` the first time I tried to compile it, but then I encountered some compiler errors, some of which I managed to fix by changing to `forward<U&&>(value)`. I then fixed some other mistakes in the code and got it to build, so I assumed all of the changes I made were needed. I guess `forward<U>(value)` was correct all along, and all I managed to do was surpress some of the later errors by changing it.

I've changed it to be `forward<U>(value)` again, gonna post a new revision.

AK/Vector.h
ccapitalK: Fixed in new revision.

bgianfo:
This looks like it's good to go?

ccapitalK:
Yeah, I don't think there's anything else to change here, can we get it merged?

@linusg

PR #6607, opened 2021-04-24
LibELF+Kernel: Support initializing values of TLS data

itamar8910:
We now support initializing TLS data with the values that exist in `PT_TLS` segments, i.e doing `__thread int myvar = 0xdeadbeef`.
Previously, TLS data was always zero-initialized.

Test program:
```
// tls-test.cpp:
#include <AK/Format.h>
#include <LibThread/Thread.h>
#include <MyLib/lib.h>

static int foo()
{
    outln("3: {:08x}", tls_var);
    return 0;
}

int main()
{
    outln("1: {:08x}", tls_var);
    tls_var = 0xbaddcafe;
    outln("2: {:08x}", tls_var);
    auto thread = LibThread::Thread::construct(foo);
    thread->start();
    return 0;
}

// lib.h:
#include <stdint.h>
extern __thread uint32_t tls_var;

// lib.cpp:
#include "lib.h"
__thread uint32_t tls_var = 0xdeadbeef;
```
// result:
```
1: deadbeef
2: baddcafe
3: deadbeef
```

cc @gunnarbeutner @ADKaster

### DIFF
--- a/Kernel/Syscalls/mmap.cpp
+++ b/Kernel/Syscalls/mmap.cpp
@@ -6,6 +6,7 @@
  */
 
 #include <AK/WeakPtr.h>
+#include <Kernel/Arch/x86/SmapDisabler.h>
 #include <Kernel/FileSystem/FileDescription.h>
 #include <Kernel/PerformanceEventBuffer.h>
 #include <Kernel/Process.h>
@@ -595,11 +596,11 @@ KResultOr<FlatPtr> Process::sys$mremap(Userspace<const Syscall::SC_mremap_params
     return ENOTIMPL;
 }
 
-KResultOr<FlatPtr> Process::sys$allocate_tls(size_t size)
+KResultOr<FlatPtr> Process::sys$allocate_tls(Userspace<const char*> initial_data, size_t size)
 {
     REQUIRE_PROMISE(stdio);
 
-    if (!size)
+    if (!size || size % PAGE_SIZE != 0)
         return EINVAL;
 
     if (!m_master_tls_region.is_null())
@@ -626,6 +627,13 @@ KResultOr<FlatPtr> Process::sys$allocate_tls(size_t size)
     m_master_tls_region = region_or_error.value()->make_weak_ptr();
     m_master_tls_size = size;
     m_master_tls_alignment = PAGE_SIZE;
+
+    {
+        Kernel::SmapDisabler disabler;
+        void* fault_at;
+        if (!Kernel::safe_memcpy((char*)m_master_tls_region.unsafe_ptr()->vaddr().as_ptr(), (char*)initial_data.ptr(), size, fault_at))
+            return EFAULT;
+    }
 
     auto tsr_result = main_thread->make_thread_specific_region({});
     if (tsr_result.is_error())
--- a/Kernel/Syscalls/mmap.cpp
+++ b/Kernel/Syscalls/mmap.cpp
@@ -619,7 +619,7 @@ KResultOr<FlatPtr> Process::sys$allocate_tls(size_t size)
     if (!range.has_value())
         return ENOMEM;
 
-    auto region_or_error = space().allocate_region(range.value(), String(), PROT_READ | PROT_WRITE);
+    auto region_or_error = space().allocate_region(range.value(), String("Master TLS"), PROT_READ | PROT_WRITE);
     if (region_or_error.is_error())
         return region_or_error.error().error();
 
--- a/Userland/Libraries/LibELF/DynamicLoader.h
+++ b/Userland/Libraries/LibELF/DynamicLoader.h
@@ -78,6 +78,7 @@ public:
     bool is_dynamic() const { return m_elf_image.is_dynamic(); }
 
     static Optional<DynamicObject::SymbolLookupResult> lookup_symbol(const ELF::DynamicObject::Symbol&);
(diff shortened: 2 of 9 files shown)


gunnarbeutner:
That's one more building block for complete TLS support done. Very nice. :partying_face:

Userland/Libraries/LibELF/DynamicLoader.h
```diff
@@ -152,7 +153,7 @@ class DynamicLoader : public RefCounted<DynamicLoader> {
     VirtualAddress m_dynamic_section_address;
 
     size_t m_tls_offset { 0 };
-    size_t m_tls_size { 0 };
+    size_t m_tls_size { 0 }; // TLS size of the current object
```
awesomekling, line 157: Instead of putting a comment here, where nobody will see it, let's instead call the variable `m_tls_size_of_current_object` :)

Kernel/Syscalls/mmap.cpp
```diff
@@ -595,14 +596,21 @@ KResultOr<FlatPtr> Process::sys$allocate_tls(size_t size)
...
         return region_or_error.error().error();
 
     m_master_tls_region = region_or_error.value()->make_weak_ptr();
     m_master_tls_size = size;
     m_master_tls_alignment = PAGE_SIZE;
 
+    {
+        Kernel::SmapDisabler disabler;
+        void* fault_at;
+        if (!Kernel::safe_memcpy((char*)m_master_tls_region.unsafe_ptr()->vaddr().as_ptr(), (char*)initial_data.ptr(), size, fault_at))
```
tomuta, line 610: I think `copy_from_user` would be safer here (which would also validate the user mode pointer). That implicitly combines `SmapDisabler` with `safe_memcpy`.

tomuta (changes requested):

Kernel/Syscalls/mmap.cpp
itamar8910: Please correct me I'm confusing this, but I don't think `copy_from_user` would work here because `m_master_tls_region` is a user address, and `copy_from_user` [doesn't allow that](https://github.com/SerenityOS/serenity/blob/b0faf2287a016a1d1568313ab28a494c4b20398e/Kernel/StdLib.cpp#L195).

itamar8910: Thinking about it though there's not reason for `m_master_tls_region` to be user-accessible, so we could change that and then use `copy_from_user`

tomuta: @itamar8910 I think you're right, I didn't realize the destination was also a user accessible pointer...

awesomekling:
This has a conflict now, sorry about that.

Kernel/Syscalls/mmap.cpp
itamar8910: I think making `m_master_tls_region`  non user-accessible can be done in a different PR, so i'm marking this discussion as reolved :)

itamar8910:
> This has a conflict now, sorry about that.

No worries, should be good now.

@gunnarbeutner can you review the fix in bf7570c?

gunnarbeutner:
With the PR applied I can't get to `dlopen()` any libraries:

Screenshot from 2021-04-26 20-39-09

itamar8910:
> With the PR applied I can't get to `dlopen()` any libraries:

Yeah this reproduces for me as well. Thanks for checking this, i'll look into why this is.
@awesomekling please do not merge yet :)

itamar8910:
> With the PR applied I can't get to `dlopen()` any libraries:

So that particular bug was a silly thing in f58cfce were i removed the `bool main_program` argument from `map_library`, but forgot to change one of its callers, which then called the `map_library(const String& filename, int fd)` overload with `false` for the fd :)

To support TLS data in `dlopen`ed libraries, I changed the TLS offset calculation in 80b9e09. After this change we don't need to change the TLS offsets of already loaded libraries if we `dlopen` a library that has TLS.

Also please note that we currently don't support initializing the TLS-data of `dlopen`ed libraries (So doing `__thread int i = 7` like you did in your example won't work).

Test program for `dlopen`ing a library that has TLS:
```
// lib.cpp
__thread int i = 0;

extern "C"
{
int get_i();
void set_i(int);

int get_i(){
    return i;
}

void set_i(int val){
    i = val;
}

}

// main.cpp
#include <stdio.h>
#include <LibDl/dlfcn.h>

int main()
{
    void* lib  = dlopen("lib.so", RTLD_NOW | RTLD_GLOBAL);
    printf("lib: %p\n", lib);
    if(!lib){
        printf("%s\n", dlerror());
        return 1;
    }
    using GetIFunc = int(*)();
    using SetIFunc = void(*)(int);
    auto get_i = (GetIFunc)dlsym(lib, "get_i");
    auto set_i = (SetIFunc)dlsym(lib, "set_i");
    printf("get_i: %p\n", get_i);
    printf("get_i(): %d\n", get_i());
    set_i(1);
    printf("get_i(): %d\n", get_i());
    return 0;
}

// output:
lib: 0x00a90218
get_i: 0x70ff14c0
get_i(): 0
get_i(): 1
```

gunnarbeutner:
Appears to be working for me. :+1: 

* Test program which creates a thread and reads/writes a `__thread` value in a `dlopen()`d shared library via some `dlsym()`d getter/setter functions.
* Python 3: `import json` and `json.dumps` works.

PR #6610, opened 2021-04-24
ProtocolServer+WebSockets+LibWeb: Added Javascript bindings for WebSockets, added an isolated service for WebSockets, renamed ProtocolServer to RequestServer

Dexesttp:
# Overview

This PR:
- Adds a new `WebSocket` service, accessible via `LibProtocol`
- Renames `ProtocolServer` to `RequestServer`
- Adds Javascript bindings from `LibWeb` to the new `WebSocket` service

Note : It seemed to make sense to rename RequestServer now, given that this PR adds another kind of isolated instance accessible via LibProtocol (and that ProtocolServer only handles HTTP-like requests anyways). That commit can easily be removed if not wanted.

## Screenshot
image


linusg:
I don't have a better name, but "download server" implies, well, downloads, IMO. Sending POST requests for *file upload* via a *download server* seems wrong..

Dexesttp:
That's a fair point. I guess either `RequestServer` or `FetchServer` might fit a bit better than `DownloadServer`, since you would `XMLHttpRequest` or `fetch()` both `GET` and `POST` requests :^)

What do you think ? I'm open to any other suggestion for the rename, in any case !

linusg:
I like `RequestServer` a lot!

Dexesttp:
Ok, give me a bit and I'll change it to that :^)

Dexesttp:
The name change to `RequestServer` has been done and the build is green !

Note that, since we renamed from `download` to `request`, I made a few other renames in addition to just renaming the server and imports :
- The `Protocol::Download` class was renamed to `Protocol::Request`, which makes a bit more sense since it could already handle both GET and POST requests.
- The `Protocol::Client` class was renamed to `Protocol::RequestClient`, because it is now only used to connect to the `RequestServer` and returns a `Request` object.
- The various `m_download` members and `download` variables holding instances of the `Protocol::Request` class (previously `Protocol::Download`) have been renamed to `m_request`, `request`, and more rarely `protocol_request` (when there was conflicts). This is just for code consistency.
- The `RequestServer` IPC API calls were renamed, with `StartDownload`, `StopDownload`, `DownloadProgress` and `DownloadFinished` changed to `FooRequest` and `RequestFoo` appropriately.
- Similarily, `download_id` from the `RequestServer` IPC API has been renamed to `request_id`

So, as a result of this renaming, the only `download`-like things that are left related to the old ProtocolServer are:
- The IPC parameter `downloaded_size` of `RequestProgress`, since it is the actual size of the downloaded data
- The `m_download` member in `DownloadWidget`, since it actually only does downloads via GET requests instead of generic requests.
- And the various variables of `pro` that relates to download size.

Userland/Libraries/LibWeb/CMakeLists.txt
```diff
@@ -296,6 +298,9 @@ function(libweb_js_wrapper class)
     add_custom_target(generate_${basename}Prototype.cpp DEPENDS Bindings/${class}Prototype.cpp)
 endfunction()
 
+libweb_js_wrapper(Communication/CloseEvent)
+libweb_js_wrapper(Communication/MessageEvent)
+libweb_js_wrapper(Communication/WebSocket)
```
linusg, line 303: Looks like only `MessageEvent` is used for multiple things and should be under `Communication` (https://html.spec.whatwg.org/multipage/comms.html), can we move the other two to `WebSocket(s)`?

Userland/Libraries/LibWeb/Communication/CloseEvent.h
```diff
@@ -0,0 +1,42 @@
...
+class CloseEvent : public DOM::Event {
+public:
+    using WrapperType = Bindings::CloseEventWrapper;
+
+    static NonnullRefPtr<CloseEvent> create(const FlyString& event_name, bool was_clean, u16 code, const String& reason)
+    {
+        return adopt_ref(*new CloseEvent(event_name, was_clean, code, reason));
+    }
+
+    virtual ~CloseEvent() override { }
```
linusg, line 22: I know historically serenity has been using empty destructors like this but recently I see (and use) more and more `= default` - presumably because clang-tidy in CLion complains about it :D

```suggestion
    virtual ~CloseEvent() override = default;
```

Userland/Libraries/LibWeb/Communication/EventNames.cpp
```diff
@@ -0,0 +1,29 @@
...
+ * Copyright (c) 2021, the SerenityOS developers.
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+#include <LibWeb/Communication/EventNames.h>
+
+namespace Web::Communication::EventNames {
+
+#define __ENUMERATE_COMMUNICATION_EVENT(name) FlyString name;
```
linusg, line 11:
```suggestion
#define __ENUMERATE_COMMUNICATION_EVENT(name) \
    FlyString name;
```

Userland/Libraries/LibWeb/Communication/EventNames.h
```diff
@@ -0,0 +1,25 @@
...
+
+// FIXME: This is not all of the events
+
+#define ENUMERATE_COMMUNICATION_EVENTS       \
+    __ENUMERATE_COMMUNICATION_EVENT(close)   \
+    __ENUMERATE_COMMUNICATION_EVENT(error)   \
+    __ENUMERATE_COMMUNICATION_EVENT(message) \
+    __ENUMERATE_COMMUNICATION_EVENT(open)
+
+#define __ENUMERATE_COMMUNICATION_EVENT(name) extern FlyString name;
```
linusg, line 21:
```suggestion
#define __ENUMERATE_COMMUNICATION_EVENT(name) \
    extern FlyString name;
```

Userland/Libraries/LibWeb/Communication/MessageEvent.h
```diff
@@ -0,0 +1,39 @@
...
+class MessageEvent : public DOM::Event {
+public:
+    using WrapperType = Bindings::MessageEventWrapper;
+
+    static NonnullRefPtr<MessageEvent> create(const FlyString& event_name, const String& data, const String& origin)
+    {
+        return adopt_ref(*new MessageEvent(event_name, data, origin));
+    }
+
+    virtual ~MessageEvent() override { }
```
linusg, line 22:
```suggestion
    virtual ~MessageEvent() override = default;
```

Userland/Libraries/LibWeb/Communication/MessageEvent.h
```diff
@@ -0,0 +1,39 @@
...
+
+    String data() const { return m_data; }
+    String origin() const { return m_origin; }
```
linusg, line 25:
```suggestion
    const String& data() const { return m_data; }
    const String& origin() const { return m_origin; }
```

Userland/Libraries/LibWeb/Communication/WebSocket.cpp
```diff
@@ -0,0 +1,275 @@
...
+
+RefPtr<Protocol::WebSocket> WebSocketHandler::connect(const URL& url)
+{
+    return m_websocket_client->connect(url);
+}
+
+// https://html.spec.whatwg.org/multipage/web-sockets.html#the-websocket-interface
+DOM::ExceptionOr<NonnullRefPtr<WebSocket>> WebSocket::create_with_global_object(Bindings::WindowObject& window, const String& url)
+{
+    URL urlRecord(url);
```
linusg, line 53:
```suggestion
    URL url_record(url);
```

Userland/Libraries/LibWeb/Communication/WebSocket.cpp
```diff
@@ -0,0 +1,275 @@
...
+    if (!urlRecord.is_valid())
+        return DOM::SyntaxError::create("Invalid URL");
+    if (urlRecord.protocol() != "ws" && urlRecord.protocol() != "wss")
```
linusg, line 56:
```suggestion
    if (!url_record.protocol().is_one_of("ws", "wss"))
```

Userland/Libraries/LibWeb/Communication/WebSocket.cpp
```diff
@@ -0,0 +1,275 @@
...
+    return WebSocket::create(window.impl(), urlRecord);
+}
+
+WebSocket::WebSocket(DOM::Window& window, URL& url)
+    : EventTarget(static_cast<Bindings::ScriptExecutionContext&>(window.document()))
+    , m_window(window)
+{
+    // FIXME: Integrate properly with FETCH as per https://fetch.spec.whatwg.org/#websocket-opening-handshake
+    m_websocket = WebSocketHandler::the().connect(url);
+    m_websocket->on_open = [weak_this = make_weak_ptr()]() {
```
linusg, line 71:
```suggestion
    m_websocket->on_open = [weak_this = make_weak_ptr()] {
```

Userland/Libraries/LibWeb/Communication/WebSocket.cpp
```diff
@@ -0,0 +1,275 @@
...
+    }
+    return WebSocket::ReadyState::Closed;
+}
+
+// https://html.spec.whatwg.org/multipage/web-sockets.html#the-websocket-interface
+String WebSocket::extensions() const
+{
+    if (!m_websocket)
+        return String::empty();
+    return String::empty();
```
linusg, line 125: Missing a `// FIXME`?

Userland/Libraries/LibWeb/Communication/WebSocket.cpp
```diff
@@ -0,0 +1,275 @@
...
+}
+
+// https://html.spec.whatwg.org/multipage/web-sockets.html#the-websocket-interface
+String WebSocket::protocol() const
+{
+    if (!m_websocket)
+        return String::empty();
+    return String::empty();
```
linusg, line 133: Missing a `// FIXME`?

Userland/Libraries/LibWeb/Communication/WebSocket.cpp
```diff
@@ -0,0 +1,275 @@
...
+{
+    // HACK : we should have an Optional<u16>
+    if (code == 0)
+        code = 1000;
+    if (code != 1000 && (code < 3000 || code > 4099))
+        return DOM::InvalidAccessError::create("The close error code is invalid");
+    if (!reason.is_empty()) {
+        if (reason.bytes().size() > 123) {
+            return DOM::SyntaxError::create("The close reason is longer than 123 bytes");
+        }
```
linusg, line 147:
```suggestion
        if (reason.bytes().size() > 123)
            return DOM::SyntaxError::create("The close reason is longer than 123 bytes");
```

Userland/Libraries/LibWeb/Communication/WebSocket.cpp
```diff
@@ -0,0 +1,275 @@
...
+    }
+    auto state = ready_state();
+    if (state == WebSocket::ReadyState::Closing || state == WebSocket::ReadyState::Closed) {
+        return {};
+    }
```
linusg, line 152:
```suggestion
    if (state == WebSocket::ReadyState::Closing || state == WebSocket::ReadyState::Closed)
        return {};
```

Userland/Libraries/LibWeb/Communication/WebSocket.cpp
```diff
@@ -0,0 +1,275 @@
...
+        auto program = parser.parse_function_node<JS::FunctionExpression>();
+        if (parser.has_errors()) {
+            dbgln("Failed to parse script in event handler attribute '{}'", name);
+            return;
+        }
+        auto* function = JS::ScriptFunction::create(script_execution_context()->interpreter().global_object(), name, program->body(), program->parameters(), program->function_length(), nullptr, false, false);
+        VERIFY(function);
+        listener = adopt_ref(*new DOM::EventListener(JS::make_handle(static_cast<JS::Function*>(function))));
+    }
+    if (listener) {
```
linusg, line 253: hmm, under what circumstances do we reach this with the RefPrt still being null?

Userland/Libraries/LibWeb/Communication/WebSocket.cpp
```diff
@@ -0,0 +1,275 @@
...
+        add_event_listener(name, listener.release_nonnull());
+    }
+}
+
+HTML::EventHandler WebSocket::get_event_handler_attribute(const FlyString& name)
+{
+    for (auto& listener : listeners()) {
+        if (listener.event_name == name && listener.listener->is_attribute()) {
+            return HTML::EventHandler { JS::make_handle(&listener.listener->function()) };
+        }
```
linusg, line 269:
```suggestion
        if (listener.event_name == name && listener.listener->is_attribute())
            return HTML::EventHandler { JS::make_handle(&listener.listener->function()) };
```

Userland/Libraries/LibWeb/Communication/WebSocket.h
```diff
@@ -0,0 +1,114 @@
...
+    using WrapperType = Bindings::WebSocketWrapper;
+
+    static NonnullRefPtr<WebSocket> create(DOM::Window& window, URL& url)
+    {
+        return adopt_ref(*new WebSocket(window, url));
+    }
+
+    static DOM::ExceptionOr<NonnullRefPtr<WebSocket>> create_with_global_object(Bindings::WindowObject& window, const String& url);
+
+    virtual ~WebSocket() override;
```
linusg, line 67:
```suggestion
    virtual ~WebSocket() override = default;
```

Userland/Libraries/LibWeb/Communication/WebSocket.h
```diff
@@ -0,0 +1,114 @@
...
+#undef __ENUMERATE
+
+    void set_event_handler_attribute(const FlyString& name, HTML::EventHandler);
+    HTML::EventHandler get_event_handler_attribute(const FlyString& name);
+
+    ReadyState ready_state() const;
+    String extensions() const;
+    String protocol() const;
+
+    String binary_type() { return m_binary_type; };
```
linusg, line 88: hoping that the wrapper generator can deal with this:

```suggestion
    const String& binary_type() { return m_binary_type; };
```

Userland/Services/RequestServer/HttpCommon.h
```diff
@@ -35,8 +35,8 @@ void init(TSelf* self, TJob job)
             self->set_downloaded_size(self->output_stream().size());
         }
 
-        // if we didn't know the total size, pretend that the download finished successfully
-        // and set the total size to the downloaded size
+        // if we didn't know the total size, pretend that the request finished successfully
+        // and set the total size to the requested size
```
linusg, line 39:
```suggestion
        // and set the total size to the downloaded size
```

Userland/Libraries/LibProtocol/WebSocketClient.cpp
```diff
@@ -0,0 +1,105 @@
...
+        return nullptr;
+    auto connection = WebSocket::create_from_id({}, *this, connection_id);
+    m_connections.set(connection_id, connection);
+    return connection;
+}
+
+u32 WebSocketClient::ready_state(Badge<WebSocket>, WebSocket& connection)
+{
+    if (!m_connections.contains(connection.id()))
+        return 4;
```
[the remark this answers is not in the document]
Dexesttp: Considering it's worng, sure ^^' It should be 3, but I think we can use the `ReadyState::Closed` value directly.

Userland/Libraries/LibWeb/Communication/WebSocket.cpp
Dexesttp: Right, and good catch!

Userland/Libraries/LibWeb/Communication/WebSocket.cpp
Dexesttp: This method is actually a copy-paste from `GlobalEventHandlers.cpp` (I should mark it as such). I assume that it's not null biven the current code, but could become so if needed?

linusg: oh ok, feel free to leave it as-is then.

Userland/Libraries/LibWeb/Communication/WebSocket.h
Dexesttp: This actually doesn't work, as it needs to `unref` the `RefPtr<Protocol::WebSocket>`, which is incomplete in the header :) Also, now I understand better why virtual destructors are sometimes out of line!
Can I live it like this ?

linusg: interesting, alright then!

Dexesttp:
All the requested changes should be applied :)

Userland/Libraries/LibProtocol/WebSocket.cpp
```diff
@@ -0,0 +1,70 @@
...
+void WebSocket::did_error(Badge<WebSocketClient>, i32 error_code)
+{
+    if (on_error)
+        on_error((WebSocket::Error)error_code);
+}
+
+void WebSocket::did_close(Badge<WebSocketClient>, u16 code, String reason, bool was_clean)
+{
+    if (!on_close)
+        on_close(code, reason, was_clean);
```
awesomekling, line 59: This doesn't look right. Should be `if (on_close)`

Also, since we've taken the `String` by value, we can `move()` it when calling `on_close`

Userland/Libraries/LibProtocol/WebSocket.h
```diff
@@ -0,0 +1,78 @@
...
+    {
+        return adopt_ref(*new WebSocket(client, connection_id));
+    }
+
+    int id() const { return m_connection_id; }
+
+    ReadyState ready_state();
+
+    void send(ByteBuffer, bool is_text);
+    void send(StringView);
```
awesomekling, line 57: Would `ReadonlyBytes` make more sense (semantically) than `StringView` here?
It's arbitrary data right?

Userland/Libraries/LibProtocol/WebSocket.cpp
```diff
@@ -0,0 +1,70 @@
+/*
+ * Copyright (c) 2021, The SerenityOS developers.
```
awesomekling, line 2: Please consider adding your personal copyright instead of this generic one. (I won't ask twice, but IMO it's very nice when people put their name on things. :)

Userland/Libraries/LibProtocol/WebSocket.h
```diff
@@ -0,0 +1,78 @@
...
+class WebSocket : public RefCounted<WebSocket> {
+public:
+    struct CertificateAndKey {
+        String certificate;
+        String key;
+    };
+
+    struct Message {
+        ByteBuffer data;
+        bool is_text;
```
awesomekling, line 31: Please always initialize primitive variables explicitly.

Userland/Libraries/LibProtocol/WebSocketClient.h
```diff
@@ -0,0 +1,45 @@
...
+
+class WebSocketClient
+    : public IPC::ServerConnection<WebSocketClientEndpoint, WebSocketServerEndpoint>
+    , public WebSocketClientEndpoint {
+    C_OBJECT(WebSocketClient);
+
+public:
+    virtual void handshake() override;
+
+    RefPtr<WebSocket> connect(const URL& url, const String& origin = {}, const Vector<String>& protocols = {}, const Vector<String>& extensions = {}, const HashMap<String, String>& request_headers = {});
```
awesomekling, line 26: The parameter name `url` is implied by the type (`URL`) here and should be omitted per the project coding style.

Userland/Services/WebSocket/ClientConnection.cpp
```diff
@@ -0,0 +1,144 @@
...
+        did_connect(id);
+    };
+    connection->on_message = [this, id](auto message) {
+        did_receive_message(id, message);
+    };
+    connection->on_error = [this, id](auto message) {
+        did_error(id, (i32)message);
+    };
+    connection->on_close = [this, id](u16 code, String reason, bool was_clean) {
+        did_close(id, code, reason, was_clean);
```
awesomekling, line 71: You can `move(reason)` here.

Userland/Services/WebSocket/ClientConnection.cpp
```diff
@@ -0,0 +1,144 @@
...
+    return make<Messages::WebSocketServer::ConnectResponse>(id);
+}
+
+OwnPtr<Messages::WebSocketServer::ReadyStateResponse> ClientConnection::handle(const Messages::WebSocketServer::ReadyState& message)
+{
+    RefPtr<WebSocket> connection = m_connections.get(message.connection_id()).value_or({});
+    if (connection) {
+        return make<Messages::WebSocketServer::ReadyStateResponse>((u32)connection->ready_state());
+    }
+    return make<Messages::WebSocketServer::ReadyStateResponse>(4);
```
awesomekling, line 85: `4` eh? I feel like this should be a constant somewhere.

Userland/Libraries/LibProtocol/WebSocketClient.h
```diff
@@ -0,0 +1,45 @@
...
+private:
+    WebSocketClient();
+
+    virtual void handle(const Messages::WebSocketClient::Connected&) override;
+    virtual void handle(const Messages::WebSocketClient::Received&) override;
+    virtual void handle(const Messages::WebSocketClient::Errored&) override;
+    virtual void handle(const Messages::WebSocketClient::Closed&) override;
+    virtual void handle(const Messages::WebSocketClient::CertificateRequested&) override;
+
+    HashMap<i32, RefPtr<WebSocket>> m_connections;
```
awesomekling, line 42: Does this need to store null values? Otherwise, can we make it a `HashMap<i32, NonnullRefPtr<WebSocket>>`?

Userland/Libraries/LibWeb/CodeGenerators/WrapperGenerator.cpp
```diff
@@ -407,7 +407,7 @@ int main(int argc, char** argv)
         return 1;
     }
 
-    if (namespace_.is_one_of("CSS", "DOM", "HTML", "UIEvents", "HighResolutionTime", "NavigationTiming", "SVG", "XHR")) {
+    if (namespace_.is_one_of("Communication", "CSS", "DOM", "HTML", "UIEvents", "HighResolutionTime", "NavigationTiming", "SVG", "WebSocket", "XHR")) {
```
awesomekling, line 410: There are no `Communication` or `WebSocket` spec that I'm aware of(?)
WebSocket is part of the HTML spec, isn't it?

awesomekling (changes requested):

Userland/Libraries/LibProtocol/WebSocket.h
Dexesttp: The idea is to have two options (text or binary data). The WebSocket protocol has a flag in the frame that indicates whether the data is UTF-8 or not, which is what the first method allows to set explicitely and the second method always set implicitely to "true".

Userland/Libraries/LibWeb/CodeGenerators/WrapperGenerator.cpp
Dexesttp: `WebSocket` is indeed defined as part of the HTML living standard, under `section 9: Communication` (https://html.spec.whatwg.org/multipage/#toc-comms). I'll move it into the HTML folder directly then, that makes more sense :)

Userland/Libraries/LibProtocol/WebSocket.cpp
Dexesttp: Good point, I also moved `data` and `reason` in the methods `WebSocket::send` and `WebSocket::close` above in the file.

Userland/Services/WebSocket/ClientConnection.cpp
Dexesttp: Yep, thait should be `WebSocket::ReadyState::Closed`, and it's in fact 3 and not 4 x) Fixing that.

Userland/Libraries/LibProtocol/WebSocketClient.h
Dexesttp: Right now, using a `RefPtr<WebSocket>` allows to do things like `if ((connection = m_connections.get(message.connection_id()).value_or(nullptr)))` without a bunch of extra casting. Otherwise, it errors out (rightfully) because `value_or(nullptr)` doesn't work on a `Optional<NonnullRefPtr<T>>`. This is a simplfied version of the code in RequestClient :)

I'll try to rework that code to use `has_value()` and `value()`, though, because storing it as a `NonnullRefPtr` and checking the optional makes more sense :)

Dexesttp: Ok, the proposed code for getting the values is now :
```C++
    auto maybe_connection = m_connections.get(message.connection_id());
    if (maybe_connection.has_value())
        maybe_connection.value()->did_open({});
 ```
 Which is way clearer :)

File: Userland/Libraries/LibProtocol/WebSocket.h (hunk 21 above)
Comment (awesomekling) [reply Userland/Libraries/LibProtocol/WebSocket.h:57]: Gotcha :)

Comment (Dexesttp) [reply Userland/Libraries/LibProtocol/WebSocket.h:57]: I added parameter names to make that clearer:
```C++
void send(ByteBuffer binary_or_text_message, bool is_text);
void send(StringView text_message);
```

Userland/Libraries/LibProtocol/WebSocketClient.h
awesomekling: Cool beans! Note that `Optional` actually has `operator->` as well :)

Userland/Libraries/LibProtocol/WebSocket.cpp
Dexesttp: Ok, sure! I took the liberty of changing the copyright headers for `LibWebSockets` and `telws` in an extra commit, since I guess I'll be using my own name from now on. I hope that's fine!

Userland/Libraries/LibProtocol/WebSocketClient.h
Dexesttp: Since it's an `Optional<NonnullRefPtr<T>>`, it seems like the arrow operator returns the `NonnullRefPtr<T>` directly which doesn't redirect to the underlying call. The same code with `maybe_connection->did_open()` (skipping the intermediate `.value()`) fails to compile.

Dexesttp:
All done, and CI is green :)

linusg:
typo in 4th commit, `matches` -> `match`
typo in 5th commit, `exising` -> `existing`

Dexesttp:
Fixed!

linusg:
Ok let's do it - great work! :heart:

PR #6611, opened 2021-04-24
LibELF+LibDl: Implement proper support for dlopen and friends

gunnarbeutner:
This implements more of the `dlfcn` functionality. Most notably:

* It's now possible to `dlopen()` libraries which were already loaded at program startup time. This does not cause those libraries to be loaded twice.
* Errors are reported via `dlerror()` rather than by crashing the program.
* Calls to the `dl*()` functions are thread-safe.


itamar8910:
Awesome that you picked this up! :tada: 
Looking forward, I wonder how we can support `dlopen` for libraries that have a `PT_TLS` segment. May be tricky because we could need to re-allocate the global TLS block, and thread safety would be an issue.
I think that your work for supporting `-ftls-mode=global-dynamic ` would be useful for this, no?

gunnarbeutner:
From the bits I gathered for `global-dynamic` so far we just need to make sure the TLS segment is big enough to start with. If you exceed the size of the initial segment you're just out look and you end up not being able to load libraries.

gunnarbeutner:
Note that this breaks some ports because `dlopen` is now handling paths according to the spec. For some of the ports library paths are incorrectly hardcoded in the binaries:

Screenshot from 2021-04-24 22-28-19

PR #6613, opened 2021-04-24
LibGfx: Add basic support for bidirectional text rendering

IdanHo:
This adds a *very* simplified version of the UNICODE BIDIRECTIONAL ALGORITHM (https://www.unicode.org/reports/tr9/), that can render most bidirectional text but also produces awkward results in a large amount of edge cases, and as such, this should probably be replaced with a fully spec compliant implementation at some point. (but since it at least renders simple bidirectional text like google.co.il just fine, i think its good enough to be merged and iterated upon.)

note for reviewers: Due to having to deal with the generic text type only supporting iteration (since we support both Utf32View which supports random access, and Utf8View which does not), i feel like this came out very allocy, so suggestions about that would be appreciated :)

### DIFF
--- a/Userland/Libraries/LibGfx/TextDirection.cpp
+++ b/Userland/Libraries/LibGfx/TextDirection.cpp
@@ -1,0 +1,34 @@
+/*
+ * Copyright (c) 2021, Idan Horowitz <idan.horowitz@serenityos.org>
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+#include <AK/Array.h>
+#include <LibGfx/TextDirection.h>
+
+namespace Gfx {
+
+// FIXME: These should be parsed from the official UnicodeData.txt that specifies the class for each character (this function doesnt take into account a large amount of characters)
+static consteval Array<BidirectionalClass, 0x1F000> generate_char_bidi_class_lookup_table()
+{
+    Array<BidirectionalClass, 0x1F000> lookup_table {};
+    for (u32 ch = 0; ch < 0x1F000; ch++) {
+        auto char_class = BidirectionalClass::STRONG_LTR;
+        if ((ch >= 0x600 && ch <= 0x7BF) || (ch >= 0x8A0 && ch <= 0x8FF) || (ch >= 0xFB50 && ch <= 0xFDCF) || (ch >= 0xFDF0 && ch <= 0xFDFF) || (ch >= 0xFE70 && ch <= 0xFEFF) || (ch >= 0x1EE00 && ch <= 0x1EEFF))
+            char_class = BidirectionalClass::STRONG_RTL; // Arabic RTL
+        if ((ch >= 0x590 && ch <= 0x5FF) || (ch >= 0x7C0 && ch <= 0x89F) || (ch == 0x200F) || (ch >= 0xFB1D && ch <= 0xFB4F) || (ch >= 0x10800 && ch <= 0x10FFF) || (ch >= 0x1E800 && ch <= 0x1EDFF) || (ch >= 0x1EF00 && ch <= 0x1EFFF))
+            char_class = BidirectionalClass::STRONG_RTL; // Hebrew RTL
+        if ((ch >= 0x30 && ch <= 0x39) || (ch >= 0x660 && ch <= 0x669) || (ch >= 0x10D30 && ch <= 0x10E7E))
+            char_class = BidirectionalClass::WEAK_NUMBERS; // Numerals
+        if ((ch >= 0x23 && ch <= 0x25) || (ch >= 0x2B && ch <= 0x2F) || (ch == 0x3A))
+            char_class = BidirectionalClass::WEAK_SEPARATORS; // Seperators
+        if ((ch >= 0x9 && ch <= 0xD) || (ch >= 0x1C && ch <= 0x22) || (ch >= 0x26 && ch <= 0x2A) || (ch >= 0x3B && ch <= 0x40) || (ch >= 0x5B && ch <= 0x60) || (ch >= 0x7B && ch <= 0x7E))
+            char_class = BidirectionalClass::NEUTRAL;
+        lookup_table[ch] = char_class;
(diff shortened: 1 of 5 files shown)


Userland/Libraries/LibGfx/TextDirection.h
```diff
@@ -0,0 +1,123 @@
...
+            char_class = BidirectionalClass::WEAK_NUMBERS; // Numerals
+        if ((ch >= 0x23 && ch <= 0x25) || (ch >= 0x2B && ch <= 0x2F) || (ch == 0x3A))
+            char_class = BidirectionalClass::WEAK_SEPERATORS; // Seperators
+        if ((ch >= 0x9 && ch <= 0xD) || (ch >= 0x1C && ch <= 0x22) || (ch >= 0x26 && ch <= 0x2A) || (ch >= 0x3B && ch <= 0x40) || (ch >= 0x5B && ch <= 0x60) || (ch >= 0x7B && ch <= 0x7E))
+            char_class = BidirectionalClass::NEUTRAL;
+        lookup_table[ch] = char_class;
+    }
+    return lookup_table;
+}
+// FIXME: This lookup table is very large, can we do anything about that?
```
alimpfard, line 41: It's best made at compiletime, though perhaps it should be moved into another library so we don't copy it into a whole bunch of binaries

IdanHo: do we statically link against LibGfx? (and if we do, wouldnt we be statically linking against the new library as well?)

alimpfard: Nah, as it is, this will be copying the lookup table into whatever object file that includes `TextDirection.h`

IdanHo: hmm, since only functions in this header file access the lookup table, i can probably just split this into a header+cpp file and make the lookup table local to the cpp file?

IdanHo: hmm, since only functions in this header file access the lookup table, i can probably just split this into a header+cpp file and make the lookup table local to the cpp file?

alimpfard: Yeah, that'll work :^)

IdanHo: will go with that then :)

IdanHo: hmm, now that im thinking about it, that means its not inlineable anymore :/

alimpfard: is there anything actually known at compiletime?
this thing will always be a runtime array lookup anyway, no?

IdanHo: yeah, but we do this lookup for **every** character rendered in serenity, so adding a function call for that is sad

alimpfard: you don't need to make it a function call, just `extern that_array`?
that array isn't that big anyway :shrug:

IdanHo: oh yeah right, nvm im dumb :facepalm:

Userland/Libraries/LibGfx/TextDirection.cpp
```diff
@@ -0,0 +1,34 @@
...
+            char_class = BidirectionalClass::WEAK_NUMBERS; // Numerals
+        if ((ch >= 0x23 && ch <= 0x25) || (ch >= 0x2B && ch <= 0x2F) || (ch == 0x3A))
+            char_class = BidirectionalClass::WEAK_SEPERATORS; // Seperators
+        if ((ch >= 0x9 && ch <= 0xD) || (ch >= 0x1C && ch <= 0x22) || (ch >= 0x26 && ch <= 0x2A) || (ch >= 0x3B && ch <= 0x40) || (ch >= 0x5B && ch <= 0x60) || (ch >= 0x7B && ch <= 0x7E))
+            char_class = BidirectionalClass::NEUTRAL;
+        lookup_table[ch] = char_class;
+    }
+    return lookup_table;
+}
+constexpr Array<BidirectionalClass, 0x1F000> char_bidi_class_lookup_table = generate_char_bidi_class_lookup_table();
```
alimpfard, line 32: Now you can mark this (and its extern friend) `constinit` and appease the C++2a gods

IdanHo: i think constexpr implies constinit, no?

alimpfard: Yeah, it does but `constinit` is _fancy_ (this is more of a joke)

IdanHo:
bump :sweat_smile:

Userland/Libraries/LibGfx/TextDirection.h
```diff
@@ -0,0 +1,103 @@
...
+#include <AK/Utf32View.h>
+#include <AK/Vector.h>
+
+namespace Gfx {
+
+enum class BidirectionalClass {
+    STRONG_LTR,
+    STRONG_RTL,
+    WEAK_NUMBERS,
+    WEAK_SEPERATORS,
```
awesomekling, line 18: SEPERATORS => SEPARATORS

IdanHo: oops, fixed!

IdanHo:
Dont merge this yet, i just rebased on latest master, and looks like its crashing on google due to some font changes, looking into it

IdanHo:
So looks like google is crashing with the bidirectional patch now due to the new google doodle which has alt text, and we render alt text with text elision enabled (which ive TODOed for bidirectional text), but ive tested it with the existing text elision, and besides the text elision being on the wrong side of the text, it looks fine, sor for now i jsut replaced the TODO() with an appropriate fixme to make text elision bidirectional text-aware

IdanHo:
This should be ready to merge :)

PR #6614, opened 2021-04-24
HackStudio: Implement ToDoEntriesWidget

guerinoni:
Now it's possible to see all FIXME and TODO of the project.

Signed-off-by: Federico Guerinoni <guerinoni.federico@gmail.com>

### DIFF
--- a/Userland/DevTools/HackStudio/AutoCompleteResponse.h
+++ b/Userland/DevTools/HackStudio/AutoCompleteResponse.h
@@ -8,6 +8,7 @@
 
 #include <AK/String.h>
 #include <AK/Types.h>
+#include <LibCpp/Parser.h>
 #include <LibGUI/AutocompleteProvider.h>
 #include <LibIPC/Decoder.h>
 #include <LibIPC/Encoder.h>
@@ -99,6 +100,35 @@ inline bool decode(Decoder& decoder, GUI::AutocompleteProvider::Declaration& dec
         return false;
 
     declaration.type = static_cast<GUI::AutocompleteProvider::DeclarationType>(type);
+    return true;
+}
+
+template<>
+inline bool encode(Encoder& encoder, Cpp::Parser::TodoEntry const& entry)
+{
+    encoder << entry.content;
+    encoder << entry.filename;
+    encoder << (u64)entry.line;
+    encoder << (u64)entry.column;
+    return true;
+}
+
+template<>
+inline bool decode(Decoder& decoder, Cpp::Parser::TodoEntry& entry)
+{
+    u64 line = 0;
+    u64 column = 0;
+    if (!decoder.decode(entry.content))
+        return false;
+    if (!decoder.decode(entry.filename))
+        return false;
+    if (!decoder.decode(line))
+        return false;
+    if (!decoder.decode(column))
+        return false;
+
+    entry.line = line;
+    entry.column = column;
     return true;
 }
 
--- a/Userland/DevTools/HackStudio/HackStudioWidget.cpp
+++ b/Userland/DevTools/HackStudio/HackStudioWidget.cpp
@@ -22,6 +22,7 @@
 #include "Project.h"
 #include "ProjectDeclarations.h"
 #include "TerminalWrapper.h"
+#include "ToDoEntries.h"
 #include <AK/LexicalPath.h>
 #include <AK/StringBuilder.h>
 #include <Kernel/API/InodeWatcherEvent.h>
@@ -867,6 +868,7 @@ void HackStudioWidget::create_action_tab(GUI::Widget& parent)
     };
 
     m_find_in_files_widget = m_action_tab_widget->add_tab<FindInFilesWidget>("Find in files");
+    m_todo_entries_widget = m_action_tab_widget->add_tab<ToDoEntriesWidget>("TODO");
     m_terminal_wrapper = m_action_tab_widget->add_tab<TerminalWrapper>("Build", false);
     m_debug_info_widget = m_action_tab_widget->add_tab<DebugInfoWidget>("Debug");
     m_disassembly_widget = m_action_tab_widget->add_tab<DisassemblyWidget>("Disassembly");
@@ -875,6 +877,10 @@ void HackStudioWidget::create_action_tab(GUI::Widget& parent)
         m_diff_viewer->set_content(original_content, diff);
         set_edit_mode(EditMode::Diff);
     });
+
+    ToDoEntries::the().on_update = [this]() {
+        m_todo_entries_widget->refresh();
+    };
 }
 
 void HackStudioWidget::create_project_tab(GUI::Widget& parent)
--- a/Userland/DevTools/HackStudio/LanguageServers/Cpp/ClientConnection.h
+++ b/Userland/DevTools/HackStudio/LanguageServers/Cpp/ClientConnection.h
@@ -22,7 +22,7 @@ public:
         m_autocomplete_engine->set_declarations_of_document_callback = [this](const String& filename, Vector<GUI::AutocompleteProvider::Declaration>&& declarations) {
             async_declarations_in_document(filename, move(declarations));
         };
-        m_autocomplete_engine->set_todo_entries_of_document_callback = [this](const String& filename, Vector<String>&& todo_entries) {
+        m_autocomplete_engine->set_todo_entries_of_document_callback = [this](String const& filename, Vector<Cpp::Parser::TodoEntry>&& todo_entries) {
             async_todo_entries_in_document(filename, move(todo_entries));
         };
     }
--- a/Userland/DevTools/HackStudio/LanguageServers/Cpp/ClientConnection.h
+++ b/Userland/DevTools/HackStudio/LanguageServers/Cpp/ClientConnection.h
@@ -22,6 +22,9 @@ public:
         m_autocomplete_engine->set_declarations_of_document_callback = [this](const String& filename, Vector<GUI::AutocompleteProvider::Declaration>&& declarations) {
             async_declarations_in_document(filename, move(declarations));
         };
+        m_autocomplete_engine->set_todo_entries_of_document_callback = [this](const String& filename, Vector<String>&& todo_entries) {
+            async_todo_entries_in_document(filename, move(todo_entries));
+        };
     }
 
     virtual ~ClientConnection() override = default;
--- a/Userland/DevTools/HackStudio/LanguageServers/Cpp/CppComprehensionEngine.cpp
+++ b/Userland/DevTools/HackStudio/LanguageServers/Cpp/CppComprehensionEngine.cpp
@@ -520,6 +520,11 @@ void CppComprehensionEngine::update_declared_symbols(DocumentData& document)
     set_declarations_of_document(document.filename(), move(declarations));
 }
 
+void CppComprehensionEngine::update_todo_entries(DocumentData& document)
+{
+    set_todo_entries_of_document(document.filename(), document.parser().get_todo_entries());
+}
+
 GUI::AutocompleteProvider::DeclarationType CppComprehensionEngine::type_of_declaration(const Declaration& decl)
 {
     if (decl.is_struct())
@@ -574,6 +579,7 @@ OwnPtr<CppComprehensionEngine::DocumentData> CppComprehensionEngine::create_docu
         root->dump();
 
     update_declared_symbols(*document_data);
+    update_todo_entries(*document_data);
 
     return document_data;
 }
--- a/Userland/DevTools/HackStudio/LanguageServers/Shell/ClientConnection.h
+++ b/Userland/DevTools/HackStudio/LanguageServers/Shell/ClientConnection.h
@@ -8,6 +8,7 @@
 
 #include "ShellComprehensionEngine.h"
 #include <DevTools/HackStudio/LanguageServers/ClientConnection.h>
+#include <LibCpp/Parser.h>
 
 namespace LanguageServers::Shell {
 
@@ -21,7 +22,7 @@ class ClientConnection final : public LanguageServers::ClientConnection {
         m_autocomplete_engine->set_declarations_of_document_callback = [this](const String& filename, Vector<GUI::AutocompleteProvider::Declaration>&& declarations) {
             async_declarations_in_document(filename, move(declarations));
         };
-        m_autocomplete_engine->set_todo_entries_of_document_callback = [this](const String& filename, Vector<String>&& todo_entries) {
+        m_autocomplete_engine->set_todo_entries_of_document_callback = [this](String const& filename, Vector<Cpp::Parser::TodoEntry>&& todo_entries) {
             async_todo_entries_in_document(filename, move(todo_entries));
         };
     }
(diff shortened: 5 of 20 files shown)


linusg:
Please remove the `Signed-off-by` line from the commit message.

FalseHonesty:
Hmm, it feels to me like doing a simple text search for "TODO"/"FIXME" isn't the best way to approach this, because they should only be treated as actual todos when they are in a comment. Maybe a file's SyntaxHighlighter can communicate a list of TODOs it finds during lexing?

guerinoni:
If reading these keywords from lexing is cheaper in terms of resources and time I can re-work in this way. If the effort is the same maybe I can just improve this search to be sure to find them only within a comment. :)

FalseHonesty:
> If reading these keywords from lexing is cheaper in terms of resources and time I can re-work in this way. If the effort is the same maybe I can just improve this search to be sure to find them only within a comment. :)

Well, trying to find them in a comment is sort of just re-doing some of the work of the lexer, so might as well bundle the two together? Additionally, comment syntax is different for different languages, so it probably makes sense for the work to be done by something language aware, like the syntax highlighter, which has a different implementation depending on language.

itamar8910:
Yeah, I agree with @FalseHonesty in that we should use a lexer's output to find TODOs, and i think it would make a lot of sense to use the language server for this.
It would also be less resource consuming because the language server lexes the code anyways.

We can have the language server look for TODOs within comment tokens and notify the client via an IPC message whenever it finds a TODO entry. This would also make the `ToDo Entries` view update live as you type :)
Essentially this can be quite similar to how the `ProjectDeclarations` gets updated from the language server and displayed in the `Locator`.

guerinoni:
@FalseHonesty @itamar8910 
Thanks! Then I can add "TODO" in lexer cpp for now and then communicate all todo entries over IPC adding a special `handle`...
I'll try to work on it :)

guerinoni:
I'm working on it :)

itamar8910:
Very cool, I'll review this tomorrow :)

Userland/DevTools/HackStudio/ToDoEntries.h
```diff
@@ -0,0 +1,40 @@
...
+
+    void set_entries(const String& filename, const Vector<String>&& entries);
+
+    Vector<ToDoEntryPair> get_entries();
+
+    Function<void()> on_update = nullptr;
+
+private:
+    ToDoEntries() = default;
+    HashMap<String, Vector<String>> m_document_to_entris;
```
itamar8910, line 37: typo: `entries`

Userland/DevTools/HackStudio/ToDoEntries.h
```diff
@@ -0,0 +1,40 @@
...
+ */
+
+#pragma once
+
+#include <AK/Function.h>
+#include <AK/HashMap.h>
+#include <AK/Noncopyable.h>
+#include <AK/String.h>
+#include <LibGUI/AutocompleteProvider.h>
+#include <LibGUI/Icon.h>
```
itamar8910, line 14: Are these two imports needed?

Userland/DevTools/HackStudio/LanguageServers/Shell/ClientConnection.h
```diff
@@ -21,6 +21,9 @@ class ClientConnection final : public LanguageServers::ClientConnection {
         m_autocomplete_engine->set_declarations_of_document_callback = [this](const String& filename, Vector<GUI::AutocompleteProvider::Declaration>&& declarations) {
             async_declarations_in_document(filename, move(declarations));
         };
+        m_autocomplete_engine->set_todo_entries_of_document_callback = [this](const String& filename, Vector<String>&& todo_entris) {
```
itamar8910, line 24: typo, `todo_entries` :)

Userland/DevTools/HackStudio/LanguageServers/Cpp/CppComprehensionEngine.cpp
```diff
@@ -516,6 +516,12 @@ void CppComprehensionEngine::update_declared_symbols(DocumentData& document)
     set_declarations_of_document(document.filename(), move(declarations));
 }
 
+void CppComprehensionEngine::update_todo_entries(DocumentData& document)
+{
+    auto todo_entries = document.parser().get_todo_entries();
```
itamar8910, line 521: nitpick: no need for the `todo_entries` variable here

Userland/DevTools/HackStudio/LanguageServers/Cpp/ClientConnection.h
```diff
@@ -22,6 +22,9 @@ class ClientConnection final : public LanguageServers::ClientConnection {
         m_autocomplete_engine->set_declarations_of_document_callback = [this](const String& filename, Vector<GUI::AutocompleteProvider::Declaration>&& declarations) {
             async_declarations_in_document(filename, move(declarations));
         };
+        m_autocomplete_engine->set_todo_entries_of_document_callback = [this](const String& filename, Vector<String>&& todo_entris) {
```
itamar8910, line 25: `todo_entries`

Userland/DevTools/HackStudio/LanguageServers/ClientConnection.h
```diff
@@ -35,6 +35,9 @@ class ClientConnection
     virtual void auto_complete_suggestions(GUI::AutocompleteProvider::ProjectLocation const&) override;
     virtual void find_declaration(GUI::AutocompleteProvider::ProjectLocation const&) override;
 
+    static void set_declarations_of_document_callback(ClientConnection&, const String&, Vector<GUI::AutocompleteProvider::Declaration>&&);
+    static void set_todo_entries_of_document_callback(ClientConnection&, const String&, Vector<String>&&);
+
```
itamar8910, line 40: Are these functions needed here in `ClientConnection.h`? I don't think they're ever used :thinking:

itamar8910 (changes requested):
Please see my comments :)
Also, some nitpicking re. commit messages:
1. Please try to avoid unnecessary abbreviations such as `func`. `function` would have fit the commit title just fine :)
2. For `b35a27e`, the convention seems to be to just use `LanguageServers:` as the 'topic'.

Userland/DevTools/HackStudio/HackStudioWidget.cpp
```diff
@@ -848,6 +849,7 @@ void HackStudioWidget::create_action_tab(GUI::Widget& parent)
     };
 
     m_find_in_files_widget = m_action_tab_widget->add_tab<FindInFilesWidget>("Find in files");
+    m_todo_entries_widget = m_action_tab_widget->add_tab<ToDoEntriesWidget>("ToDo Entries");
```
itamar8910, line 852: Also, let's name this tab just "TODO"

Userland/DevTools/HackStudio/LanguageServers/Cpp/ClientConnection.h
guerinoni: Done

Userland/DevTools/HackStudio/HackStudioWidget.cpp
guerinoni: Done

Userland/DevTools/HackStudio/LanguageServers/Cpp/CppComprehensionEngine.cpp
guerinoni: Done :D

Userland/DevTools/HackStudio/ToDoEntries.h
guerinoni: Good catch!

Userland/Libraries/LibCpp/Parser.h
```diff
@@ -35,6 +35,7 @@ class Parser final {
     void print_tokens() const;
     const Vector<String>& errors() const { return m_state.errors; }
     const Preprocessor::Definitions& preprocessor_definitions() const { return m_preprocessor_definitions; }
+    Vector<String> get_todo_entries();
```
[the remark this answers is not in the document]
guerinoni: Done!

Userland/DevTools/HackStudio/LanguageServers/ClientConnection.h
guerinoni: Solved :)

itamar8910 (approved):
LGTM, thanks for working on this! :tada:

linusg (changes requested):
You have made various fixes suggested for either the first or second commit in the third commit. Please don't do that, squash them into the appropriate preceding commit instead.

The comments regarding commit messages have been ignored.

guerinoni:
@linusg thank you for remind me :) now should be consistent

itamar8910:
Please see my previous comments regarding commit messages:
> 1. Please try to avoid unnecessary abbreviations such as `func`. `function` would have fit the commit title just fine :)
> 2. For `b35a27e`, the convention seems to be to just use `LanguageServers:` as the 'topic'.

linusg (commented):
Still:

> Please try to avoid unnecessary abbreviations such as func. function would have fit the commit title just fine :)

Specific suggestions to make this easier:

- `LibCpp: Add function for retrieving TODO comments from the parser`
- `LanguageServers: Add function to collect TODO entries in a document`

Userland/DevTools/HackStudio/ToDoEntriesWidget.cpp
```diff
@@ -0,0 +1,96 @@
...
+                return Gfx::FontDatabase::default_fixed_width_font();
+            return {};
+        }
+        if (role == GUI::ModelRole::Display) {
+            auto& match = m_matches.at(index.row());
+            switch (index.column()) {
+            case Column::Filename:
+                return match.filename;
+            case Column::Text:
+                return match.comment;
```
alimpfard, line 63: Not knowing the source position of a Todo/FIXME kinda just defeats the whole purpose of a widget like this
There would be lots of context-dependent to-dos without their context, and no way to get to it...

alimpfard:
Also, please write the commit messages in proper English, the ones @linusg suggested could serve as examples (or you could just use them as-is)

Userland/DevTools/HackStudio/ToDoEntriesWidget.cpp
guerinoni: Maybe with line can be useful? Or click to bring the cursor in that position...

alimpfard: Yep.
I think LibCpp already knows about token positions, so you can probably just use those to calculate the positions for the TODOs/FIXMEs.

guerinoni: Now I pass a `Vector<String>` through the functions, maybe I need some struct of TODO entry to pass

guerinoni: Add position and click to bring the cursor on that TODO

Userland/Libraries/LibCpp/Preprocessor.h
```diff
@@ -14,6 +14,14 @@
 #include <AK/Vector.h>
 
 namespace Cpp {
+
+struct TodoEntry {
```
itamar8910, line 18: Does the `TodoEntry` struct belong in `Preprocessor.h`?
I think a better place to define it in would be in `Parser.h`, just before the `get_todo_entries()` function.

Userland/DevTools/HackStudio/ToDoEntriesWidget.cpp
```diff
@@ -0,0 +1,114 @@
...
+ToDoEntriesWidget::ToDoEntriesWidget()
+{
+    set_layout<GUI::VerticalBoxLayout>();
+    m_result_view = add<GUI::TableView>();
+
+    m_result_view->on_activation = [](auto& index) {
+        auto& match = *(const Cpp::TodoEntry*)index.internal_data();
+        open_file(match.filename);
+        current_editor().set_selection({ { match.line, match.column }, { match.line, match.column } });
+        current_editor().set_focus(true);
```
itamar8910, line 110: I think using the version of `open_file` that takes a line&column would be better here.
(Also, I don't think we really need to have the TODO line selected, just jumping to it in the source is enough).
```suggestion
        open_file(match.filename, match.line, match.column);
```

Userland/DevTools/HackStudio/AutoCompleteResponse.h
```diff
@@ -102,4 +103,33 @@ inline bool decode(Decoder& decoder, GUI::AutocompleteProvider::Declaration& dec
     return true;
 }
 
+template<>
+inline bool encode(Encoder& encoder, const Cpp::TodoEntry& entry)
```
itamar8910, line 107: According to the new conventions, new code should use east consts (Please also correct this in other places in the code).

itamar8910 (changes requested):

Userland/Libraries/LibCpp/Preprocessor.h
guerinoni: Done :)

Userland/DevTools/HackStudio/ToDoEntriesWidget.cpp
guerinoni: Done :)

Userland/DevTools/HackStudio/AutoCompleteResponse.h
guerinoni: Done :)

itamar8910 (approved):
Okay, looks good to me.

guerinoni:
@linusg @alimpfard do you think this can be ok? :^)

itamar8910:
Bumping this, @linusg @alimpfard :smile:

Userland/DevTools/HackStudio/ToDoEntries.cpp
```diff
@@ -0,0 +1,34 @@
...
+    if (on_update)
+        on_update();
+}
+
+Vector<Cpp::Parser::TodoEntry> ToDoEntries::get_entries()
+{
+    Vector<Cpp::Parser::TodoEntry> ret;
+    for (auto& it : m_document_to_entries)
+        for (auto& entry : it.value)
+            ret.append({ entry.content, it.key, entry.line, entry.column });
```
linusg, line 29: Please add curly braces to any if/else/for/etc body with more than one line:

```suggestion
    for (auto& it : m_document_to_entries) {
        for (auto& entry : it.value)
            ret.append({ entry.content, it.key, entry.line, entry.column });
    }
```

linusg (commented):
thanks @itamar8910 :smile:

one nitpick, but the 4th commit is a bit of a mess and does much more than it claims to - namely all the east-const conversions for code added in the previous commits. - those should be squashed instead.

Userland/DevTools/HackStudio/ToDoEntries.cpp
guerinoni: I can't squash them because I also changed some struct in function parameter

PR #6618, opened 2021-04-24
AK/GenericLexer: Add a Position struct and keep track of the line&column numbers

benit8:
Hi all,

I was recently reading the [first PR for the new LibSQL](https://github.com/SerenityOS/serenity/pull/6463) and @trflynn89 mentioned how the `GenericLexer` does not keep track of the line&column numbers of the string it _lexes_, so I thought I would implement it.

All reference to the member `m_index` have been replaced with calls to `tell()` and the new private method `advance()`. It manages how the `GenericLexer` keeps track of the line&column numbers, thus the `m_index` should only be increased/decreased from there.

I am a bit unsure about how to compute the column number when retreating characters though. For now I reverse-iterate until the previous `\n`, but I wonder if it would be ok on very large strings.

Finally, I am thinking of adding something like a `GenericLexer::Token`, which would look something like this:
```c++
struct Token {
    StringView trivia;
    Position start, end;
};
```
With the same goal in mind as its parent (to bootstrap lexers' tokens), and then adapt the `GenericLexer`'s method to return these tokens. But maybe this is too bloat for what the `GenericLexer` is for, please let me know what you think about it :slightly_smiling_face:

### DIFF
--- a/Userland/Libraries/LibC/scanf.cpp
+++ b/Userland/Libraries/LibC/scanf.cpp
@@ -413,7 +413,7 @@ extern "C" int vsscanf(const char* input, const char* format, va_list ap)
         // Parse width specification
         [[maybe_unused]] int width_specifier = 0;
         if (format_lexer.next_is(isdigit)) {
-            auto width_digits = format_lexer.consume_while([](char c) { return isdigit(c); });
+            auto width_digits = format_lexer.consume_while(isdigit);
             width_specifier = width_digits.to_int().value();
             // FIXME: Actually use width specifier
         }
--- a/AK/GenericLexer.cpp
+++ b/AK/GenericLexer.cpp
@@ -17,9 +17,9 @@ StringView GenericLexer::consume(size_t count)
     if (count == 0)
         return {};
 
-    size_t start = m_index;
-    size_t length = min(count, m_input.length() - m_index);
-    m_index += length;
+    size_t start = tell();
+    size_t length = min(count, tell_remaining());
+    advance(length);
 
     return m_input.substring_view(start, length);
 }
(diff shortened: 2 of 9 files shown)


alimpfard:
Could you make this an optional behaviour? e.g. via a template parameter.
there are a whole bunch of `GenericLexer `users that don't need position tracking (or do it themselves).

Userland/Libraries/LibC/scanf.cpp
```diff
@@ -413,7 +413,7 @@ extern "C" int vsscanf(const char* input, const char* format, va_list ap)
         // Parse width specification
         [[maybe_unused]] int width_specifier = 0;
         if (format_lexer.next_is(isdigit)) {
-            auto width_digits = format_lexer.consume_while([](char c) { return isdigit(c); });
+            auto width_digits = format_lexer.consume_while(isdigit);
```
alimpfard, line 416: `isdigit` returns an `int` which means it won't be accepted as a `Function<bool(char)>` - or did this interface change?

benit8: It did change: instead of taking a `Function<bool(char)>` we have a templated version of `consume_while()`-like methods. See it [there](https://github.com/SerenityOS/serenity/blob/d65fa91bc49d2db5c2ada06584f240f2b09ba7c3/AK/GenericLexer.h#L180)

alimpfard: Ah nice : )

benit8:
@alimpfard
> Could you make this an optional behaviour? e.g. via a template parameter.

I'm not sure how templates would help to make this optional, it would mean templating the whole class (if I get you correctly?). Do you know of an example in the codebase that does something similar?

Also the way you use the `GenericLexer` doesn't change at all; it's up to you to get the position if you want it, so, to me, it already feels optional in that sense :slightly_smiling_face:

IdanHo:
> Also the way you use the `GenericLexer` doesn't change at all; it's up to you to get the position if you want it, so, to me, it already feels optional in that sense

I think the problem is the added (and unneeded in those cases) complexity in the advance method.

alimpfard:
> Do you know of an example in the codebase that does something similar?

For an example of a conditionally added fields and logic, you could look at `IntrusiveList.h:154` (follow where `m_self` is referenced, and where `SelfReferenceIfNeeded<T, bool>` is defined.

> it's up to you to get the position if you want it

Lots of users simply want to consume characters in a nice way without dealing with positions at all, it would be nice to not track them to begin with :^)

benit8:
Thank you for your explanations, I think I get your point. Perhaps the position tracking should reside in another class made just for lexing source code (with more advanced features), and the lexing of simple strings left for the `GenericLexer`.

PR #6621, opened 2021-04-24
Ports: Fix python3 package so linting script doesn't error out.

bgianfo:
I have my environment configured to use https://pre-commit.com/.
I guess the scripts were changed recently to lint all ports, and
the python port was barfing on my system because of this bug.


bgianfo:
cc: @gunnarbeutner who added the new ports linting functionality in b3db01e20eeae632cc75df9af8666772bda67091

linusg:
causing a bug like this is literally the only downside of having an up-to-date Python installed :wink:

Thanks!

PR #6627, opened 2021-04-25
IPCCompiler: Use string hashes for IPC endpoint magic

sin-ack:
This patch removes the IPC endpoint numbers that needed to be specified
in the IPC files.  Since the string hash is a (hopefully) collision free
number that depends on the name of the endpoint, we now use that
instead. :^)

<details>
<summary> ... </summary>

Endpoint numbers are GONE :crab: 

</details>

### DIFF
--- a/Userland/DevTools/IPCCompiler/main.cpp
+++ b/Userland/DevTools/IPCCompiler/main.cpp
@@ -164,11 +164,7 @@ int main(int argc, char** argv)
         lexer.consume_specific("endpoint");
         consume_whitespace();
         endpoints.last().name = lexer.consume_while([](char ch) { return !isspace(ch); });
-        consume_whitespace();
-        assert_specific('=');
-        consume_whitespace();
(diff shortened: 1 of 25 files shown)


Userland/DevTools/IPCCompiler/main.cpp
```diff
@@ -164,11 +164,7 @@ int main(int argc, char** argv)
         lexer.consume_specific("endpoint");
         consume_whitespace();
         endpoints.last().name = lexer.consume_while([](char ch) { return !isspace(ch); });
-        consume_whitespace();
-        assert_specific('=');
-        consume_whitespace();
-        auto magic_string = lexer.consume_while([](char ch) { return !isspace(ch) && ch != '{'; });
-        endpoints.last().magic = magic_string.to_int().value();
+        endpoints.last().magic = Traits<String>::hash(endpoints.last().name);
```
awesomekling, line 167: While there could totally be a collision, we'll discover pretty quickly if this ever happens :)

PR #6628, opened 2021-04-25
Tests: Move TestSuite.h into LibTest, Re-swizzle CMake for tests

ADKaster:
See commit for details but summary is essentially:

- Slip in env var to disable benchmarks in CI
- Move AK/TestSuite.h into LibTest
   - Break up into multiple headers
   - Move some details into .cpp files
   - Make `main` function for tests a boilerplate .cpp file to add to executable instead of macros
- Add serenity_test cmake function to streamline adding tests that will be installed into the target
- Fix Lagom/CMakeLists.txt for the fallout 😢  That file is 🍝 

Work towards #4078


Userland/Libraries/LibTest/TestMain.cpp
```diff
@@ -0,0 +1,29 @@
+/*
+ * Copyright (c) 2021, Andrew Kaster <akaster@serenityos.org>
```
ADKaster, line 2: One thing I'm unsure about is copyright headers for these files in LibTest where I moved TestSuite.h into. Seems I was inconsistent... TestSuite.h only had andreas on it. Should all these files have Andreas 2018-2020, or wha?

PR #6630, opened 2021-04-25
Port: Add port for tcl

gunnarbeutner:
Because, well, who doesn't love Tcl. Remember eggdrop?

### DIFF
--- a/Userland/Libraries/LibC/dirent.cpp
+++ b/Userland/Libraries/LibC/dirent.cpp
@@ -35,13 +35,21 @@ int closedir(DIR* dirp)
 {
     if (!dirp || dirp->fd == -1)
         return -EBADF;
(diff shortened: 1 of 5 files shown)


Userland/Libraries/LibC/dirent.cpp
```diff
@@ -35,15 +35,23 @@ int closedir(DIR* dirp)
 {
     if (!dirp || dirp->fd == -1)
         return -EBADF;
-    if (dirp->buffer)
-        free(dirp->buffer);
+    free(dirp->buffer);
```
emanuele6, line 38: ?

awesomekling: `free(nullptr)` is a no-op

PR #6633, opened 2021-04-25
Meta+Ports: Only allow sha256 and sig for auth_type in ports 

gunnarbeutner:
This PR changes the port scripts so that:

* Updates the `ports-lint.py` script to only allow SHA256 and GPG.
* Updates all ports which still used other signature algorithms to use SHA256.
* Updates the `.port_include.sh` so that it attempts to download files again one more time when verification fails. This was previously an issue when cancelling `./package.sh fetch` during the download phase. That way you'd end up with partially downloaded files which would then fail verification when the `package.sh` script is run again.
* Makes sure we only extract files after verification has succeeded.

### DIFF
--- a/Meta/lint-ports.py
+++ b/Meta/lint-ports.py
@@ -117,6 +117,11 @@ def check_package_files(ports):
             continue
 
         props = get_port_properties(port)
+
+        if not props['auth_type'] in ('sha256', 'sig', ''):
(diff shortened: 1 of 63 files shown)


linusg:
LGTM, but I think swapping the first and third commit would make more sense in terms of git history.

gunnarbeutner:
Aye, you're right. There we go. :)

Issue #4228, opened 2020-11-29
du: doesn't work for directories


danboid:
du fails to give the disk space used by a directory, it only seems to work for files but that makes it pointless as you could use ls for that. Every dir is 4K to it, no matter how many files t contains 

du would also benefit from a -h option to show sizes in MB and GB instead of KB.

neeraj92:
Is anyone working on this ? Can I try implementing this feature ?

tomuta:
> Is anyone working on this ? Can I try implementing this feature ?

first come first served...

PR #6635, opened 2021-04-25
Fix du to display folder space and add -h option

neeraj92:
Some fixes for #4228


linusg (changes requested):
Hello! Please make sure to follow the [contribution guidelines](https://github.com/SerenityOS/serenity/blob/master/CONTRIBUTING.md). Specifically split up your work into multiple, atomic commits and prefix commit messages with the area of the system they're affecting, here `du: ...`.

PR #6636, opened 2021-04-25
FileManager+unzip: Unzip from context menu

TheGeopard:
My first OSS contribution, here goes nothing :^)

This PR adds a new context menu action to FileManager that allows for extracting zip archives without having to open the terminal. I thought this would make a good first contribution after playing around a bit with SerenityOS and finding this quality of life feature that can be found in other file managers like those of Gnome and Windows to be lacking.

"Extract Here" in Gnome
Screenshot from 2021-04-24 13-40-33

My implementation in SerenityOS
Screenshot from 2021-04-25 14-30-27

Future improvements may include support for more archive formats like .tar and .tar.gz as well as integration with the new file operation progress tracking.

### DIFF
--- a/Userland/Applications/FileManager/main.cpp
+++ b/Userland/Applications/FileManager/main.cpp
@@ -49,6 +49,7 @@
 #include <spawn.h>
 #include <stdio.h>
 #include <string.h>
+#include <sys/wait.h>
 #include <unistd.h>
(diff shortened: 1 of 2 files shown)


TheGeopard:
Whoops, forgot to run clang format...

PR #6638, opened 2021-04-25
FileManager+unzip: Unzip from context menu

TheGeopard:
Can't really call this my first PR anymore after forgetting to format the old one properly :facepalm:, sorry for the trouble, still new to this.

This PR adds a new context menu action to FileManager that allows for extracting zip archives without having to open the terminal. I thought this would make a good first contribution after playing around a bit with SerenityOS and finding this quality of life feature that can be found in other file managers like those of Gnome and Windows to be lacking.

"Extract Here" in Gnome
Screenshot from 2021-04-24 13-40-33

My implementation in SerenityOS
Screenshot from 2021-04-25 14-30-27

Future improvements may include support for more archive formats like .tar and .tar.gz as well as integration with the new file operation progress tracking.

### DIFF
--- a/Userland/Applications/FileManager/main.cpp
+++ b/Userland/Applications/FileManager/main.cpp
@@ -49,6 +49,7 @@
 #include <spawn.h>
 #include <stdio.h>
 #include <string.h>
+#include <sys/wait.h>
 #include <unistd.h>
 
 using namespace FileManager;
@@ -58,6 +59,7 @@ static int run_in_windowed_mode(RefPtr<Core::ConfigFile>, String initial_locatio
 static void do_copy(const Vector<String>& selected_file_paths, FileUtils::FileOperation file_operation);
 static void do_paste(const String& target_directory, GUI::Window* window);
 static void do_create_link(const Vector<String>& selected_file_paths, GUI::Window* window);
+static void do_unzip_archive(const Vector<String>& selected_file_paths, GUI::Window* window);
 static void show_properties(const String& container_dir_path, const String& path, const Vector<String>& selected, GUI::Window* window);
 static bool add_launch_handler_actions_to_menu(RefPtr<GUI::Menu>& menu, const DirectoryView& directory_view, const String& full_path, RefPtr<GUI::Action>& default_action, NonnullRefPtrVector<LauncherHandler>& current_file_launch_handlers);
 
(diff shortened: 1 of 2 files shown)


IdanHo:
nice work, but for next time: instead of opening a new PR, you can force push to the branch your PR tracks to replace the commits in the PR :)

Userland/Applications/FileManager/main.cpp
```diff
@@ -192,6 +194,32 @@ void do_create_link(const Vector<String>& selected_file_paths, GUI::Window* wind
...
+        if (rc < 0) {
+            perror("execlp");
+            _exit(1);
+        }
+    } else {
+        // FIXME: this could probably be tied in with the new file operation progress tracking
+        int status;
+        waitpid(unzip_pid, &status, 0);
+        if (status != 0)
+            GUI::MessageBox::show(window, "Could not extract archive", "Extract Archive Error", GUI::MessageBox::Type::Error);
```
linusg, line 219: `waitpid()` itself needs an error check as well, also I think we should use `WIFEXITED` and `WEXITSTATUS` here:

```suggestion
        int status;
        int rc = waitpid(unzip_pid, &status, 0);
        if (rc < 0 || !WIFEXITED(status) || WEXITSTATUS(status) != 0)
            GUI::MessageBox::show(window, "Could not extract archive", "Extract Archive Error", GUI::MessageBox::Type::Error);
```

Userland/Applications/FileManager/main.cpp
```diff
@@ -354,6 +394,11 @@ int run_in_desktop_mode([[maybe_unused]] RefPtr<Core::ConfigFile> config)
                 file_context_menu->add_action(directory_view.delete_action());
                 file_context_menu->add_separator();
 
+                if (node.full_path().ends_with(".zip")) {
```
linusg, line 397: Let's ensure this also works with `ARCHIVE.ZIP`, i.e. let's do a case-insensitive compare.

```suggestion
                if (node.full_path().ends_with(".zip", CaseSensitivity::CaseInsensitive)) {
```

Userland/Applications/FileManager/main.cpp
```diff
@@ -965,6 +1023,11 @@ int run_in_windowed_mode(RefPtr<Core::ConfigFile> config, String initial_locatio
                 file_context_menu->add_action(shortcut_action);
                 file_context_menu->add_separator();
 
+                if (node.full_path().ends_with(".zip")) {
```
linusg, line 1026: Same as above

Userland/Utilities/unzip.cpp
```diff
@@ -111,6 +114,20 @@ int main(int argc, char** argv)
         return 1;
     }
 
+    if (!output_directory_path.is_empty()) {
```
linusg, line 117: Let's ensure we don't allow `-o ""`, just use the string's null state instead meaning the argument wasn't passed:

```suggestion
    if (!output_directory_path.is_null()) {
```

linusg (changes requested):

TheGeopard:
I have implemented all the suggested changes 👍

PR #6640, opened 2021-04-25
Kernel: Some PATA fixes & IDEChannel design robustness improvements

supercomputer7:


supercomputer7:
I tested this on my ICH7 test machine and it works OK. The goal is to fix the code to support IDE controllers in PCI native mode. When I enable "Enhanced Mode" in the BIOS, it creates (or actually fakes, which is what hardware and firmware do here) one PCI IDE controller in ISA compatibility mode, and another PCI IDE controller in PCI native mode. When that happens, things go haywire, so I'm working on a fix for it (hopefully soon, in a separate PR).

Kernel/Storage/IDEChannel.cpp
```diff
@@ -269,28 +302,37 @@ UNMAP_AFTER_INIT void IDEChannel::detect_disks()
...
         m_io_group.io_base().offset(ATA_REG_LBA2).out<u8>(0);
         m_io_group.io_base().offset(ATA_REG_COMMAND).out<u8>(ATA_CMD_IDENTIFY); // Send the ATA_IDENTIFY command
 
-        // Wait for the BSY flag to be reset
-        while (m_io_group.control_base().in<u8>() & ATA_SR_BSY)
-            ;
-
-        if (m_io_group.control_base().in<u8>() == 0x00) {
-            dbgln_if(PATA_DEBUG, "IDEChannel: No {} {} disk detected!", channel_type_string().to_lowercase(), channel_string(i));
+        // Wait 10 second for the BSY flag to clear
```
awesomekling, line 321: 10 seconds seems like a hilariously long time to wait for a response to `ATA_IDENTIFY`

supercomputer7: Let's make it 2 seconds then, I think it's enough time for the drive to give a correct response.

PR #6648, opened 2021-04-25
Meta: Lint commits for unix-style linebreaks

IdanHo:
This ensures serenity commits only contain unix-style LF linebreaks


linusg:
Awesome, thanks!

PR #6652, opened 2021-04-26
Userland: Make rmdir take multiple paths

ccapitalK:


Userland/Utilities/rmdir.cpp
```diff
@@ -16,16 +17,19 @@ int main(int argc, char** argv)
...
-    int rc = rmdir(path);
-    if (rc < 0) {
-        perror("rmdir");
-        return 1;
+    int status = 0;
+    for (auto path : paths) {
+        int rc = rmdir(path);
+        if (rc < 0) {
+            perror("rmdir");
+            status = errno;
```
linusg, line 31: Since we only store `errno` of the last failed `rmdir()`, this doesn't seem particularly useful. Let's just go back to a generic "1" instead.

```suggestion
            status = 1;
```

linusg (changes requested):

Userland/Utilities/rmdir.cpp
ccapitalK: Fixed.

Issue #104, opened 2019-05-26
Kernel: Tickless


awesomekling:
Currently the kernel wakes up 1000 times/sec to increment the system timer.
It would be cool if we could instead compute when the next necessary wakeup would be, and then sleep for that long instead.

PR #6653, opened 2021-04-26
WIP: Kernel: Add tickless mode

tomuta:
_This is very much work in progress and does not yet work!_


ccapitalK:
This should fix #104.

Issue #4809, opened 2021-01-05
LibGUI: window/widget cursor changes do not take effect until a mouse event occurs


bcoles:
Window cursors set with `set_cursor()` and Widget cursors set with `set_override_cursor()` do not take effect immediately. The cursor is not updated until a mouse event occurs (move, click, etc).

Using `window->update()` does not force update the cursor either.

Reproduce:

```cpp
    window->show();
    window->set_cursor(Gfx::StandardCursor::Wait);
    window->update();
```

Note the cursor does not change until it is moved.

This can also be demonstrated with `Breakout` and `Pong`.

PR #6655, opened 2021-04-26
Demos: Add Starfield screensaver demo

jagger27:
I made a little Starfield screensaver. :^)

It appears that the cursor doesn't update to its override bitmap until it moves, which is a bit inconvenient for a screensaver. The other Screensaver demo has the same issue.


bcoles:
> It appears that the cursor doesn't update to its override bitmap until it moves

#4809

gunnarbeutner:
This is cool. Maybe add a launcher for it (see `Base/res/apps/Screensaver.af` for a reference)?

Issue #4392, opened 2020-12-12
Kernel modules are broken


bugaevc:
This is not a new bug, it hasn't been working *at least* since the CMake switch.

```
SHENANIGANS! program header offset (0) or section header offset (24748) overlap with ELF header!
Image::parse(): ELF Header not valid
SHENANIGANS! program header offset (0) or section header offset (24748) overlap with ELF header!
Image::parse(): ELF Header not valid
modload(26): perror(): module_load: Exec format error
```

I am now reminded of this because I am doing [a thing](https://github.com/bugaevc/dev-yes/tree/serenity/serenity) and I cannot even test if it works or not :cry:

ADKaster:
Hm, those specific checks in LibElf/Validation.cpp I added very recently though to fix a fuzz error. Perhaps there's other things wrong too?

bugaevc:
Yeah I was getting a different error (a crash even, I think?) the last time.

ADKaster:
It doesn't look like it crashes with that check fixed (it shouldn't care about that condition when elf_header.e_phnum is 0), but, the logic in sys$module_load definitely needs some help.

readelf -r on TestModule.o shows there's a few types of relocations that get generated for the text section, namely
```
 R_386_PC32
 R_386_GOTPC
 R_386_PLT32
 R_386_GOTOFF
 ```
 Also shows that all of the R_386_32 relocations are in `.rel.debug_*` sections. Seems the module load code was written before the kernel was compiled with -fPIE -pie? The current load code only handles R_386_PC32 and R_386_32 relocations. It also seems to think that the relocations or the debug_info and friends sections are 100% required, and fails to load because of that.

ADKaster:
Eh, okay so. If you strip the kernel module with `strip --strip-uneeded`, all the R_386_32 relocations go away. So that's good confirmation that they aren't really needed. For the PLT entries, we can cheat, given the way the procedure linkage table is supposed to work. Can just insert the external symbol's address directly in there and it should "just work" ™️ . 

For GOTPC and GOTOFF... I think the module load method has to actually create a global offset table for the module and fill it with the global offset table entries referenced by R_386_GOTOFF entries. Otherwise, the patching process becomes a bit more complex? The GOTOFF entries are the actual read-only data strings that TestModule is trying to print so, can't get out of relocating references to them properly.

ElectrodeYT:
I pulled main and tried a modload on the TestModule.o file, and this is printed to the Serial port:
[...]

PR #6659, opened 2021-04-26
Kernel: Fix Modules (WIP)

ElectrodeYT:
Fixes the KUBSAN crash in #4392


IdanHo:
I dont think just returning on misaligned symbols is a fix, since those actually show up in compiled binaries (as the issue shows), so we need to actually handle them

ElectrodeYT:
Oops, didnt see that! Sorry.
Will keep this open (under a new new) because i do plan on trying to fix kernel modules (if i can, that is)

ElectrodeYT:
This really isnt something for me, ill be honest. I havent read the ELF spec and i have other stuff i would rather do on serenity, and that i can actually do and properly test.

PR #6663, opened 2021-04-26
LibAudio: Support 32 and 64-bit float WAV files

kleinesfilmroellchen:
LibAudio's WavLoader plugin for loading WAV files now supports loading audio files with 32-bit float or 64-bit float samples.

By supporting these new non-int sample formats, Audio::Buffer now stores the sample format (out of a list of supported formats) instead of the raw bit depth. (The bit depth is easily calculated with pcm_bits_per_sample)

### DIFF
--- a/Userland/Libraries/LibAudio/WavLoader.cpp
+++ b/Userland/Libraries/LibAudio/WavLoader.cpp
@@ -1,5 +1,6 @@
 /*
  * Copyright (c) 2018-2020, Andreas Kling <kling@serenityos.org>
+ * Copyright (c) 2021, kleines Filmröllchen <malu.bertsch@gmail.com>
  *
  * SPDX-License-Identifier: BSD-2-Clause
  */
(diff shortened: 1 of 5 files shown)


Userland/Libraries/LibAudio/WavLoader.cpp
```diff
@@ -185,16 +187,44 @@ bool WavLoaderPlugin::parse_header()
...
+        m_sample_format = PcmSampleFormat::Int16;
+    } else if (audio_format == WAVE_FORMAT_PCM && bits_per_sample == 24) {
+        m_sample_format = PcmSampleFormat::Int24;
+    } else if (audio_format == WAVE_FORMAT_IEEE_FLOAT && bits_per_sample == 32) {
+        m_sample_format = PcmSampleFormat::Float32;
+    } else if (audio_format == WAVE_FORMAT_IEEE_FLOAT && bits_per_sample == 64) {
+        m_sample_format = PcmSampleFormat::Float64;
+    } else {
+        VERIFY_NOT_REACHED();
+    }
```
Hendiadyoin1, line 223: You can make this in less comparissons, you are checking the same thing multiple times

Hendiadyoin1: Still verifying twice, that the bits_per_sample is valid

kleinesfilmroellchen: This follows the style of the other code in this function, and I think it's outside of the scope of this PR.

PR #6667, opened 2021-04-26
LibGUI: Make common locations configurable

Dexesttp:
# Overview

This commit:
- Reworks the common locations in `GUI::FilePicker` to be less hardcoded
- Adds a new `GUI::CommonLocationsProvider` class to handle common locations used in FilePicker
- Adds a new (optional) `~/.config/CommonLocations.json` file, that can be used to configure the common locations for each user
- Adds two new common locations to the JSON file of `anon` : `/home/anon/Documents` and `/home/anon/Downloads`


awesomekling:
Well that just makes perfect sense. Nice work :star:

PR #6669, opened 2021-04-26
Kernel: Make sure the E1000 network adapter keeps receiving packets

gunnarbeutner:
Previously the E1000 network adapter would stop receiving further packets when an RX buffer overrun occurred. This was the case when connecting the adapter to a real network where enough broadcast traffic caused the buffer to be full before the kernel had a chance to clear the RX buffer.

This also updates the E1000 module to use macros instead of hard-coded magic values.

Further it adds logging for E1000 RX buffer overruns and avoids resetting the IRQ mask on each interrupt.

### DIFF
--- a/Kernel/Net/E1000NetworkAdapter.cpp
+++ b/Kernel/Net/E1000NetworkAdapter.cpp
@@ -425,10 +425,8 @@ void E1000NetworkAdapter::receive()
 {
     auto* rx_descriptors = (e1000_tx_desc*)m_rx_descriptors_region->vaddr().as_ptr();
     u32 rx_current;
-    for (;;) {
+    for (u32 i = 0; i < number_of_rx_descriptors; i++) {
         rx_current = in32(REG_RXDESCTAIL) % number_of_rx_descriptors;
-        if (rx_current == (in32(REG_RXDESCHEAD) % number_of_rx_descriptors))
-            return;
         rx_current = (rx_current + 1) % number_of_rx_descriptors;
         if (!(rx_descriptors[rx_current].status & 1))
             break;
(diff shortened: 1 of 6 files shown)


gunnarbeutner:
@supercomputer7 I was told you might be able to test this if you have some spare time. While this works just fine I wonder if it also works with real hardware.

IdanHo (approved):
LGTM

supercomputer7:
> @supercomputer7 I was told you might be able to test this if you have some spare time. While this works just fine I wonder if it also works with real hardware.

Me? Well, getting an Intel Ethernet card is still a goal I ought to achieve, so the answer unfortunately is that I can't test it right now. Nevertheless, the patch looks good. I might have a few points to indicate to fix, so everything will be marvelous :)

Kernel/Net/E1000NetworkAdapter.cpp
```diff
@@ -117,6 +117,8 @@ namespace Kernel {
 #define INTERRUPT_TXD_LOW (1 << 15)
 #define INTERRUPT_SRPD (1 << 16)
 
+#define PCI_VENDOR_INTEL 0x8086
```
supercomputer7, line 120: Since this macro is suitable for other purposes (also in the near future, a hint for my PR on Intel graphics), let's start with a file for such macros. This is also doesn't fill quite right to put this macro here, as E1000-compatible NICs (and there are lots of these) are almost guaranteed to be manufactured by Intel.

gunnarbeutner: I've created a new file for those (`Kernel/PCI/IDs.h`) and moved all existing vendor and device IDs there.

PR #6671, opened 2021-04-26
Uncrosscompiling: Various fixes, step 1

boricj:
I'm experimenting with building a far larger amount of Serenity's userland on Linux than Lagom currently does (as in, what happens if we don't cross-compile the source tree?). There are various practical reasons for this :
- Expose more of Serenity's code to 64-bit environments in anticipation for a x86_64 port ;
- Expose more of Serenity's code to foreign environments for uncovering broken and non-standard usages ;
- Make more code potentially eligible for CI testing than Lagom currently allows ;
- Use Serenity's userland elsewhere than inside SerenityOS.

These are various trivial fixes for compatibility/correctness issues uncovered during this process. The most controversial one here is probably moving non-standard constants from `math.h` into a dedicated header `LibCore/ExtraMathConstants.h` ; I'm not sure what would be the best option here actually, but the current one won't work outside of SerenityOS.


Userland/Libraries/LibJS/Runtime/Value.cpp
```diff
@@ -81,7 +81,7 @@ static String double_to_string(double d)
         builder.append(double_to_string(-d));
         return builder.to_string();
     }
-    if (d == INFINITY)
+    if (d == static_cast<double>(INFINITY))
```
linusg, line 84: as `INFINITY` is supposed to be a float, please update our `math.h` as well where it's currently defined as `__builtin_huge_val()`, which is a double according to the GCC docs.

Userland/Libraries/LibM/math.h
```diff
@@ -57,10 +57,6 @@ __BEGIN_DECLS
 #define M_SQRT2f32 1.41421356237309504880f
 #define M_SQRT1_2f32 0.70710678118654752440f
 
-#define M_TAU 6.28318530717958647692
-#define M_DEG2RAD 0.01745329251994329576
-#define M_RAD2DEG 57.2957795130823208767
```
linusg, line 62: `math.cpp` is using `M_TAU`, where is it now getting that from?

Userland/Libraries/LibThread/Thread.h
```diff
@@ -33,7 +33,7 @@ class Thread final : public Core::Object {
 
 private:
     explicit Thread(Function<int()> action, StringView thread_name = nullptr);
-    Function<int()> m_action;
+    Function<intptr_t()> m_action;
```
linusg, line 36: It's not obvious to me why this is needed - in fact, the constructor just a line above this still takes a `Function<int()>` which is used as the value for `m_action` - so this doesn't look like "Fix int to pointer conversion"...? Please explain in the commit message what's going on here.

Userland/Utilities/hostname.cpp
```diff
@@ -6,6 +6,7 @@
 
 #include <LibCore/ArgsParser.h>
 #include <errno.h>
+#include <limits.h>
```
linusg, line 9: Our `HOST_NAME_MAX` is not defined in limits.h, is it supposed to be?

linusg (changes requested):

Userland/Libraries/LibM/math.h
boricj: It's getting them from `LibCore/ExtraMathConstants.h`, as mentioned in the PR description. I'm not sure what's the right thing to do here, but on foreign environments LibC, LibCrypt, LibM and LibPthread are unavailable because the host already provides these libraries, so putting these constants inside LibM makes them out of reach.

linusg: oh, I've missed the include that was added. :+1:

Userland/Utilities/hostname.cpp
boricj: SerenityOS currently defines it inside `unistd.h`, POSIX mandates `limits.h`. I'll amend the PR to move it to its proper place.

Userland/Libraries/LibThread/Thread.h
boricj: Silly me forgot to update the constructor (and possibly all the callers). That return value is effectively passed over to `void pthread_exit(void *value_ptr)`, so that `int` needs to be an `intptr_t` to not risk losing bits for `pthread_join()`.

linusg:
Looks good, but please move themath constants to AK - it was decided that LibCore already contains things beyond its envisioned scope ("event loop & I/OO") and is not the right place for them.

PR #6680, opened 2021-04-26
New System Wallpaper

zadockmaloba:
Created a new simple system wallpaper.


ccapitalK:
Looks good, though IDK why but the kerning between the 't' and the 'y' seems a bit off to me.

Also, please remember to squash your commits.

ccapitalK:
Oh, also, please remove `(resized)` from the name of the file, the steps made in creating the file should not be visible in the final result. :^)

PR #6684, opened 2021-04-27
VimEditingEngine: allow selection of the endline character

### LINKED ISSUE #6014: TextEditor: Newline in visual selection sometimes counts as three characters
The number of selected characters displayed in the toolbar is not accurate as seen in the images below.

image
image
image

Can also be reproduced by entering visual mode on the newline character:

image

Pangoraw:
This patch fixes the visual selection of endline characters in the
VimEditingEngine. When the visual mode is disabled and the cursor
is located on the endline character, it is shifted back to the last
character of the line.

Fixes #6014

### DIFF
--- a/Userland/Libraries/LibGUI/VimEditingEngine.cpp
+++ b/Userland/Libraries/LibGUI/VimEditingEngine.cpp
@@ -657,7 +657,7 @@ void VimMotion::calculate_find_range(VimCursor& cursor, int amount)
     m_find_mode = FindMode::None;
 }
 
-Optional<TextPosition> VimMotion::get_position(VimEditingEngine& engine)
+Optional<TextPosition> VimMotion::get_position(VimEditingEngine& engine, bool in_visual_mode)
 {
     auto range_optional = get_range(engine, true);
     if (!range_optional.has_value())
@@ -708,11 +708,13 @@ Optional<TextPosition> VimMotion::get_position(VimEditingEngine& engine)
             // above in get_range normalizes some values which shouldn't be
             // end-exclusive during normal operations.
             bool is_at_start = range.end().column() == 0;
-            size_t column = is_at_start ? 0 : range.end().column() - 1;
-            // Need to not go beyond the last character, as standard in vim.
             auto& line = editor.line(range.end().line());
 
-            return { TextPosition { range.end().line(), min(column, line.length() - 1) } };
+            size_t column = is_at_start ? 0 : range.end().column() - 1;
+            column = min(column, line.length() - (in_visual_mode ? 0 : 1));
+            // Need to not go beyond the last character, as standard in vim.
+
+            return { TextPosition { range.end().line(), column } };
         }
     }
     }
@@ -998,7 +1000,7 @@ bool VimEditingEngine::on_key_in_visual_mode(const KeyEvent& event)
         m_motion.add_key_code(event.key(), event.ctrl(), event.shift(), event.alt());
         if (m_motion.is_complete()) {
             if (!m_motion.is_cancelled()) {
-                auto maybe_new_position = m_motion.get_position(*this);
+                auto maybe_new_position = m_motion.get_position(*this, true);
                 if (maybe_new_position.has_value()) {
                     auto new_position = maybe_new_position.value();
                     m_editor->set_cursor(new_position);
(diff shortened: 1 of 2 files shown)


Userland/Libraries/LibGUI/VimEditingEngine.cpp
```diff
@@ -657,7 +657,7 @@ void VimMotion::calculate_find_range(VimCursor& cursor, int amount)
     m_find_mode = FindMode::None;
 }
 
-Optional<TextPosition> VimMotion::get_position(VimEditingEngine& engine)
+Optional<TextPosition> VimMotion::get_position(VimEditingEngine& engine, bool can_go_over_endline)
```
sin-ack, line 660: How about something like `bool in_visual_mode` instead? That makes the intentions more clear.

Userland/Libraries/LibGUI/VimEditingEngine.cpp
```diff
@@ -708,11 +708,13 @@ Optional<TextPosition> VimMotion::get_position(VimEditingEngine& engine)
             // above in get_range normalizes some values which shouldn't be
             // end-exclusive during normal operations.
             bool is_at_start = range.end().column() == 0;
+            auto& line = editor.line(range.end().line());
+
             size_t column = is_at_start ? 0 : range.end().column() - 1;
+            column = min(column, line.length() - (1 - can_go_over_endline));
```
sin-ack, line 714:
```suggestion
            column = min(column, line.length() - !can_go_over_endline);
```

Since `!` maps any integer to `{0, 1}`, that will make it a bit nicer, i think.

Userland/Libraries/LibGUI/VimEditingEngine.cpp
```diff
@@ -1137,9 +1140,17 @@ void VimEditingEngine::switch_to_visual_mode()
...
+    auto end = m_selection_start_position < cursor ? cursor : m_selection_start_position;
+
+    if (end.column() >= m_editor->current_line().length()) {
+        if (end.line() != m_editor->line_count() - 1)
+            end = { end.line() + 1, 0 };
+    } else {
+        end.set_column(end.column() + 1);
+    }
+
+    m_editor->selection()->set(start, end);
```
sin-ack, line 1153: Was this the original definition before I got rid of `m_selection_start_position`? That still looks a tad bit too complicated. The if statements there might also break `dj`/`dk`, do they still work?

Userland/Libraries/LibGUI/VimEditingEngine.cpp
```diff
@@ -1148,7 +1159,14 @@ void VimEditingEngine::clear_visual_mode_data()
     if (m_editor->has_selection()) {
         m_editor->selection()->clear();
         m_editor->did_update_selection();
+
+        auto cursor = m_editor->cursor();
+        if (cursor.column() >= m_editor->current_line().length()) {
+            cursor.set_column(m_editor->current_line().length() - 1);
+            m_editor->set_cursor(cursor);
+        }
```
sin-ack, line 1167: Entering Vim emulation mode for the first time needs this too, so might make sense to put this in a separate function so both constructor and `clear_visual_mode_data` can use it.

sin-ack (changes requested):
Looks pretty good overall, just a couple minor things. :^)

Userland/Libraries/LibGUI/VimEditingEngine.cpp
Pangoraw: It wasn't, but the previous implementation had issue #6014 where the selection range at the end of the line would be invalid.
I counted three cases:
 - Normal case: column + 1
 - End of the line but not last line: wrap to the start of the next line
 - End of the line and last line: do not update the selection since there are no character at the cursor location but the cursor can move there anyway because `in_visual_mode = true`.

Userland/Libraries/LibGUI/VimEditingEngine.cpp
Pangoraw: See https://github.com/SerenityOS/serenity/pull/6684/files#diff-c13c2f0b49898f9e679fc5f13025ba8d27d9607fb40760810b906da16cbc3785R1157-R1165

sin-ack:
LGTM, `$` doesn't go to the end of line and `j`/`k` will "correct" the position to before the newline, but that can be fixed in a separate PR. :^)

Userland/Libraries/LibGUI/VimEditingEngine.cpp
```diff
@@ -708,11 +708,13 @@ Optional<TextPosition> VimMotion::get_position(VimEditingEngine& engine)
             // above in get_range normalizes some values which shouldn't be
             // end-exclusive during normal operations.
             bool is_at_start = range.end().column() == 0;
+            auto& line = editor.line(range.end().line());
+
             size_t column = is_at_start ? 0 : range.end().column() - 1;
+            column = min(column, line.length() - !in_visual_mode);
```
linusg, line 714: I think this is easier to read and understand if made explicit:

```suggestion
            column = min(column, line.length() - (in_visual_mode ? 0 : 1));
```

linusg (changes requested):

PR #6685, opened 2021-04-27
Ports: Add MilkyTracker port

jakewestrip:
Ported MilkyTracker 1.03, a music tracker which attempts to recreate FastTracker II, to SerenityOS.

The interface update issues encountered during previous port attempts were solved by replacing SDL Timer with LibCore::Timer.


jakewestrip:
Note: I've tried to name the patch files in such a way that explains what it is that they actually do - let me know if some other naming convention is preferred.

PR #6686, opened 2021-04-27
Ports: MilkyTracker - Fix install directory, remove SDL Timer changes

jakewestrip:
After my previous PR adding MilkyTracker was merged, I synchronized my local Serenity repo and rebuilt from scratch, and it seems that I missed handling this specific issue. Letting the install stage pass DESTDIR to make will result in an incorrect install path, so we specify that install() should not pass this.


jakewestrip:
SDL Timers seem to have been fixed (by https://github.com/SerenityOS/serenity/pull/6664 I presume) so I've gone ahead and removed the patches which replace SDL Timer with LibCore Timer

linusg (changes requested):
Please split this into two separate, atomic commits :)

PR #6687, opened 2021-04-27
LibTest + Tests: Convert a few Kernel tests to be LibTest based. 

bgianfo:
A few more small steps towards #4078:

 -   LibTest: Add FAIL() macro allow tests to force failure with message.

  -  Tests: Convert unveil-test-failures.cpp to be LibTest based.

 -   Tests: Convert pledge-test-failures.cpp to be LibTest based.

  -  Tests: Convert file system suid-sgid stripping test to be LibTest based.


Userland/Libraries/LibTest/Macros.h
```diff
@@ -85,3 +85,9 @@ void current_test_case_did_fail();
             ::Test::current_test_case_did_fail();                                                               \
         }                                                                                                       \
     } while (false)
+
+#define FAIL(msg)                                                                  \
+    do {                                                                           \
+        ::AK::warnln("\033[31;1mFAIL\033[0m: {}:{}: {}", __FILE__, __LINE__, msg); \
+        ::Test::current_test_case_did_fail();                                      \
+    } while (false)
```
linusg, line 93:
```suggestion
#define FAIL(message)                                                                  \
    do {                                                                               \
        ::AK::warnln("\033[31;1mFAIL\033[0m: {}:{}: {}", __FILE__, __LINE__, message); \
        ::Test::current_test_case_did_fail();                                          \
    } while (false)
```

no need to abbreviate here - https://github.com/SerenityOS/serenity/blob/master/Documentation/CodingStyle.md#names

>  Use full words, except in the rare case where an abbreviation would be more canonical and easier to understand.

Userland/Tests/Kernel/TestKernelPledge.cpp
```diff
@@ -0,0 +1,34 @@
...
+
+#include <LibTest/TestCase.h>
+#include <unistd.h>
+
+TEST_CASE(test_nonexistent_pledge)
+{
+    auto res = pledge("testing123", "notthere");
+    if (res >= 0) {
+        FAIL("Pledging on existent promises should fail.");
+    }
```
linusg, line 15: since we have a single-line body in all of these `if`s now, let's drop the curlies :)

```suggestion
    if (res >= 0)
        FAIL("Pledging on existent promises should fail.");
```

linusg (commented):
neat!

PR #6689, opened 2021-04-27
Ports: Fix path for the tcl port's patch file

gunnarbeutner:
This just didn't work at all before. Oops.


alimpfard:
:thinking: 
when did we get a tcl port again?

linusg:
> when did we get a tcl port again?

https://github.com/SerenityOS/serenity/pull/6630

PR #6695, opened 2021-04-27
WindowManager: Remove MouseMove workaround for hovered windows

### LINKED ISSUE #6692: Screensaver: Exits immediately after launch

GMTA:
In `WindowManager::reevaluate_hovered_window()` a workaround existed that triggered the hovered window by sending a fake `MouseMove` event. After 2e1320f264, the Screensaver application stopped working since it received a `MouseMove` immediately.

This commit removes the `MouseMove` workaround (introduced in 2d29bfc89ef):

* The fake `MouseMove` event is removed from `::reevaluate_hovered_window()`;
* The existing `WindowEntered` and `WindowLeft` events are converted to mouse events and made to include the mouse position;
* `GUI::Window::handle_mouse_event()` now handles the `WindowEntered` and `WindowLeft` events, performing a mouse hit test for the former.

Fixes #6692.

Tested as part of this PR: the Breakout game, the Cube demo (frameless), Screensaver.

CC @tomuta

### DIFF
--- a/Userland/Libraries/LibGUI/Window.cpp
+++ b/Userland/Libraries/LibGUI/Window.cpp
@@ -335,6 +335,12 @@ void Window::handle_drop_event(DropEvent& event)
 
 void Window::handle_mouse_event(MouseEvent& event)
 {
+    if (event.type() == Event::WindowLeft) {
+        set_hovered_widget(nullptr);
+        Application::the()->set_drag_hovered_widget({}, nullptr);
+        return;
+    }
+
     if (m_global_cursor_tracking_widget) {
         auto window_relative_rect = m_global_cursor_tracking_widget->window_relative_rect();
         Gfx::IntPoint local_point { event.x() - window_relative_rect.x(), event.y() - window_relative_rect.y() };
@@ -529,12 +535,6 @@ void Window::handle_drag_move_event(DragEvent& event)
     }
 }
 
-void Window::handle_left_event()
-{
-    set_hovered_widget(nullptr);
-    Application::the()->set_drag_hovered_widget({}, nullptr);
-}
-
 void Window::event(Core::Event& event)
 {
     ScopeGuard guard([&] {
@@ -544,7 +544,7 @@ void Window::event(Core::Event& event)
     if (event.type() == Event::Drop)
         return handle_drop_event(static_cast<DropEvent&>(event));
 
-    if (event.type() == Event::MouseUp || event.type() == Event::MouseDown || event.type() == Event::MouseDoubleClick || event.type() == Event::MouseMove || event.type() == Event::MouseWheel)
+    if (event.type() == Event::MouseUp || event.type() == Event::MouseDown || event.type() == Event::MouseDoubleClick || event.type() == Event::MouseMove || event.type() == Event::MouseWheel || event.type() == Event::WindowEntered || event.type() == Event::WindowLeft)
         return handle_mouse_event(static_cast<MouseEvent&>(event));
 
     if (event.type() == Event::MultiPaint)
@@ -561,9 +561,6 @@ void Window::event(Core::Event& event)
 
     if (event.type() == Event::WindowCloseRequest)
         return handle_close_request();
-
-    if (event.type() == Event::WindowLeft)
-        return handle_left_event();
 
     if (event.type() == Event::Resize)
         return handle_resize_event(static_cast<ResizeEvent&>(event));
--- a/Userland/Libraries/LibGUI/Window.h
+++ b/Userland/Libraries/LibGUI/Window.h
@@ -215,7 +215,6 @@ private:
     void handle_theme_change_event(ThemeChangeEvent&);
     void handle_screen_rect_change_event(ScreenRectChangeEvent&);
     void handle_drag_move_event(DragEvent&);
-    void handle_left_event();
 
     void server_did_destroy();
 
--- a/Userland/Libraries/LibGUI/WindowServerConnection.cpp
+++ b/Userland/Libraries/LibGUI/WindowServerConnection.cpp
@@ -102,13 +102,13 @@ void WindowServerConnection::handle(const Messages::WindowClient::WindowCloseReq
 void WindowServerConnection::handle(const Messages::WindowClient::WindowEntered& message)
 {
     if (auto* window = Window::from_window_id(message.window_id()))
-        Core::EventLoop::current().post_event(*window, make<Event>(Event::WindowEntered));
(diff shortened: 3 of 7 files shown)


tomuta:
This doesn't seem right. The purpose of the fake messages is to react to alpha channel changes. I don't see how it can still work with these changes.

It feels like https://github.com/SerenityOS/serenity/commit/2e1320f2645d6082db57e4d430e908e9792dc7cc is maybe not the proper use of that API.

GMTA:
@tomuta 
> This doesn't seem right. The purpose of the fake messages is to react to alpha channel changes. I don't see how it can still work with these changes.

That API sent out two events when a window became hovered: `WindowEntered` and `MouseMove` (faked).

Since the `WindowEntered` is still being sent as part of a window becoming hovered, we can now use that for the alpha channel hit testing.

tomuta:
> @tomuta
> 
> > This doesn't seem right. The purpose of the fake messages is to react to alpha channel changes. I don't see how it can still work with these changes.
> 
> That API sent out two events when a window became hovered: `WindowEntered` and `MouseMove` (faked). In the new situation, only `WindowEntered` is sent.
> 
> Since `WindowEntered` is still being sent as part of a window becoming hovered, it is now used for the alpha channel hit testing.
> 
> But let me know what exactly isn't working since my testing revealed no regressions :-)

The `MouseMove` is needed, otherwise you wouldn't notice e.g. the pointer hover a button and e.g. trigger a tooltip...

GMTA:
@tomuta 
> The `MouseMove` is needed, otherwise you wouldn't notice e.g. the pointer hover a button and e.g. trigger a tooltip...

The thing is, the mouse did not move. So why send out a mouse move event? I think it would make more sense to perform hit testing on a `WindowEntered` (since clearly the mouse has entered the window) and forward this to the widget.

tomuta:
> @tomuta
> 
> > The `MouseMove` is needed, otherwise you wouldn't notice e.g. the pointer hover a button and e.g. trigger a tooltip...
> 
> The thing is, the mouse did not move. So why send out a mouse move event? I think it would make more sense to perform hit testing on a `WindowEntered` (since clearly the mouse has entered the window) and forward this to the widget, which is exactly what this PR implements.

Well, you didn't move the mouse but anything that should be hovered at that mouse location should be hovered. Normally, if you physically move the mouse into a widget, you'll get `WindowEntered` and at least one `MouseMove`. If e.g. alpha hit testing changes to become transparent below the mouse pointer, we should emulate "moving" the mouse pointer to that spot, triggering all the usual events.

tomuta:
So, if the alpha value changes and underneath is a button with a tooltip, when the alpha value changes to transparent, the button should get highlighted and trigger the tooltip.

GMTA:
> Well, you didn't move the mouse but anything that should be hovered at that mouse location should be hovered. Normally, if you physically move the mouse into a widget, you'll get `WindowEntered` and at least one `MouseMove`. If e.g. alpha hit testing changes to become transparent below the mouse pointer, we should emulate "moving" the mouse pointer to that spot, triggering all the usual events.

If you were to continue this line of thinking: should programmatically moving a window behind the cursor also trigger a fake mouse move event?

tomuta:
> > Well, you didn't move the mouse but anything that should be hovered at that mouse location should be hovered. Normally, if you physically move the mouse into a widget, you'll get `WindowEntered` and at least one `MouseMove`. If e.g. alpha hit testing changes to become transparent below the mouse pointer, we should emulate "moving" the mouse pointer to that spot, triggering all the usual events.
> 
> If you were to continue this line of thinking: should programmatically moving a window behind the cursor also trigger a fake mouse move event?

Yes, as well as resizing (which I think https://github.com/SerenityOS/serenity/commit/2e1320f2645d6082db57e4d430e908e9792dc7cc is meant to address). Example would be the CatDog demo.

tomuta:
Example that is now broken: Take Cube, remove the window frame move it over e.g. the Terminal's menu bar. When you hover the mouse pointer over a menu item it no longer changes state properly as the cube moves in/out

GMTA:
> Yes, as well as resizing (which I think [2e1320f](https://github.com/SerenityOS/serenity/commit/2e1320f2645d6082db57e4d430e908e9792dc7cc) is meant to address). Example would be the CatDog demo.

So my line of thinking is this:
* Only send out a mouse move event if the mouse _actually moves_
* Send out window entered / left events if the mouse enters or leaves the window, regardless of how this came to be (mouse move, alpha hit testing, window move, window resize)
* Widgets should respond to the right mouse events: both on mouse move and on window enter

At any rate, I don't think that _if_ you have a workaround, it should exist in the WindowServer. We could also implement the workaround in the client, creating a fake MouseMove event on receiving a WindowEntered event?

tomuta:
I think we just need to special case the full screen window somehow

GMTA:
> Example that is now broken: Take Cube, remove the window frame move it over e.g. the Terminal's menu bar. When you hover the mouse pointer over a menu item it no longer changes state properly as the cube moves in/out

This could also be fixed by listening to `WindowEntered` and creating a local `MouseMove` for the widget to digest.

GMTA:
> I think we just need to special case the full screen window somehow

What do you mean?

tomuta:
Well I do see your point of adding the mouse coordinates to the `WindowEntered` event and only send `MouseMove` when the mouse is actually moved. But that requires every client code to be changed to handle `WindowEntered` mouse coordinates the same as `MouseMove`, except screen savers. Seems quite error prone and won't be noticed until you do the Cube test.

My thinking was merely that maybe we should just treat fullscreen windows slightly different and not send `MouseMove` unless the mouse actually moved. And maybe not `WindowEntered` at all as that's implied anyway.

GMTA:
@tomuta Please check 85eb3bb385c30bbd1c3d785a797ef119fbe2dc03 which reintroduces the `MouseMove` workaround in `GUI::Window`. What do you think?

tomuta:
Well that doesn't help with window frames, e.g. highlighted buttons in the window frame, which is handled in WindowServer directly.

GMTA:
@tomuta event / widget handling in `WindowServer` itself makes this PR somewhat more irrelevant. I think the fix for #6692 should be that the Screensaver app implements a hysteresis for mouse movement instead of triggering on the first mouse event it receives. Closing.

tomuta:
This is what I was eluding to to, which also fixes the ScreenSaver: https://github.com/SerenityOS/serenity/commit/9084bdc985eee481e25ae09189998c3f61a31cd3

PR #6696, opened 2021-04-27
WindowServer: Fix ScreenSaver not showing

### LINKED ISSUE #6692: Screensaver: Exits immediately after launch

tomuta:
We shouldn't send fake MouseMove events to fullscreen windows. There 
is no way another window could trigger moving into a fullscreen window,
so when re-evaluating hit testing, which is now also done when changing
a window's geometry, we only send a WindowEntered event, and then only
MouseMove events if the mouse was actually moved.

Fixes #6692

Successor of #6695 by @GMTA

### DIFF
--- a/Userland/Services/WindowServer/WindowManager.cpp
+++ b/Userland/Services/WindowServer/WindowManager.cpp
@@ -1116,7 +1116,7 @@ void WindowManager::reevaluate_hovered_window(Window* updated_window)
         if (currently_hovered && m_resize_candidate == currently_hovered)
             clear_resize_candidate();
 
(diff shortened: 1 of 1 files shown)


tomuta:
Although this approach has its merits, I think the better solution is to essentially ignore the first `MouseMove` in the screensaver app, as mentioned by @GMTA in https://github.com/SerenityOS/serenity/issues/6692#issuecomment-827976065

This way we consistently do a `WindowEntered` + `MouseMoved`, which may be useful even in a fullscreen (non-screensaver) case.

PR #6698, opened 2021-04-27
Screensaver: Implement mouse hysteresis

### LINKED ISSUE #6692: Screensaver: Exits immediately after launch

GMTA:
Allow the mouse to move a bit before actually closing the app.

Fixes #6692

Fun fact: this is the third attempted fix for this demo app :^)

### DIFF
--- a/Userland/Demos/Screensaver/Screensaver.cpp
+++ b/Userland/Demos/Screensaver/Screensaver.cpp
@@ -23,6 +23,7 @@ public:
 private:
     Screensaver(int width = 64, int height = 48, int interval = 10000);
     RefPtr<Gfx::Bitmap> m_bitmap;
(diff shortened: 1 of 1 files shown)


jagger27:
This change could also be ported to Starfield too. I only left out the mouse move trigger because of the cursor update bug. Perhaps these (and future) screensavers should share a common codebase.

tomuta:
> This change could also be ported to Starfield too. I only left out the mouse move trigger because of the cursor update bug. Perhaps these (and future) screensavers should share a common codebase.

I didn't even realized we had a new screen saver. Looks pretty cool!

linusg:
> Fun fact: this is the third attempted fix for this demo app :^)

And it's been nice to see you guys discussing the various possible solutions and drawbacks :)

PR #6704, opened 2021-04-28
Kernel: Avoid overrunning the user-specified buffers in select()

gunnarbeutner:
The kernel incorrectly assumes that the user-specified buffers for `select()` have all 1024 bits available. When the user specifies a smaller value for `nfds` this assumption is not always true.

### DIFF
--- a/Kernel/Syscalls/select.cpp
+++ b/Kernel/Syscalls/select.cpp
@@ -48,11 +48,16 @@ KResultOr<int> Process::sys$select(Userspace<const Syscall::SC_select_params*> u
     });
 
     fd_set fds_read, fds_write, fds_except;
-    if (params.readfds && !copy_from_user(&fds_read, params.readfds))
+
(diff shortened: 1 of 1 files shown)


Kernel/Syscalls/select.cpp
```diff
@@ -48,11 +48,16 @@ KResultOr<int> Process::sys$select(Userspace<const Syscall::SC_select_params*> u
     });
 
     fd_set fds_read, fds_write, fds_except;
-    if (params.readfds && !copy_from_user(&fds_read, params.readfds))
+
+    size_t bytes_used = (params.nfds / 8) + ((params.nfds % 8) == 0 ? 0 : 1);
```
awesomekling, line 52: You can use `ceil_div()` (from `AK/StdLibExtras.h`) to make this sort of thing more readable. :)

gunnarbeutner: Cool, fixed. :+1:

PR #6708, opened 2021-04-28
HackStudio: Detection of externally deleted files

iCristalrope:
Each time the window of Hackstudio becomes active, Hackstudio will check
if the files in `m_open_files_vector` or in the editors have been deleted.
If so, the deleted files will be removed from the open files list and
editors that showed those files will instead show another open file or
show a default editor.

Closes SerenityOS#6632

### DIFF
--- a/Userland/DevTools/HackStudio/HackStudioWidget.cpp
+++ b/Userland/DevTools/HackStudio/HackStudioWidget.cpp
@@ -33,6 +33,7 @@
 #include <LibCore/Event.h>
 #include <LibCore/EventLoop.h>
 #include <LibCore/File.h>
+#include <LibCore/FileWatcher.h>
 #include <LibDebug/DebugSession.h>
 #include <LibGUI/Action.h>
 #include <LibGUI/ActionGroup.h>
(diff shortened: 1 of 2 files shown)


awesomekling (changes requested):
This sort of thing should be done lazily using file watchers, not forcibly every time we activate the HackStudio window.

itamar8910 (changes requested):
Thanks for working on this. Please a comment i left.

Userland/DevTools/HackStudio/HackStudioWidget.cpp
```diff
@@ -1074,6 +1083,41 @@ void HackStudioWidget::initialize_menubar(GUI::Menubar& menubar)
     create_help_menubar(menubar);
 }
 
+void HackStudioWidget::handle_external_file_deletion(const String& filepath)
+{
+    struct stat st;
+    if (lstat(filepath.characters(), &st) == 0) {
+        return;
+    }
```
itamar8910, line 1091: Let's do the `lstat` check before calling `handle_external_file_deletion`.
Currently `handle_external_file_deletion` gets called whenever a file changes.

itamar8910: Also, let's explicitly check if `errno` is `ENOENT` to determine that the file was deleted.

itamar8910 (approved):
Okay, looking good.

PR #6710, opened 2021-04-28
Wallpaper: New system wallpaper

zadockmaloba:
Just a simple PNG wallpaper.


linusg:
I'm afraid this doesn't fit the system's aesthetic at all.

linusg:
Also note that the name should be stylized as "SerenityOS", not Serenity OS" :)

PR #6714, opened 2021-04-28
LibDebug: Implement support for DWARF 5

gunnarbeutner:
I have a hunch we might need support for DWARF 5 soon. So here it is:

Screenshot from 2021-04-28 22-38-57


awesomekling:
Oh nice, that's a bit less code than I expected :)

PR #6715, opened 2021-04-28
Kernel: Add MSG_PEEK support for the IPv4Socket

sw1tchbl4d3r:
MSG_PEEK allows a packet to be read from the kernel's buffer without removing it from said buffer.

This PR will add 2 things:
- A `peek()` function for the `DoubleBuffer`, to let `DoubleBuffer` data be seen without incrementing the read-index, and
- The actual MSG_PEEK, which needs the `peek()` Function mentioned beforehand.

### DIFF
--- a/Kernel/DoubleBuffer.cpp
+++ b/Kernel/DoubleBuffer.cpp
@@ -44,7 +44,6 @@ ssize_t DoubleBuffer::write(const UserOrKernelBuffer& data, size_t size)
 {
     if (!size || m_storage.is_null())
         return 0;
-    VERIFY(size > 0);
     Locker locker(m_lock);
     size_t bytes_to_write = min(size, m_space_for_writing);
     u8* write_ptr = m_write_buffer->data + m_write_buffer->size;
@@ -61,7 +60,6 @@ ssize_t DoubleBuffer::read(UserOrKernelBuffer& data, size_t size)
 {
     if (!size || m_storage.is_null())
         return 0;
-    VERIFY(size > 0);
     Locker locker(m_lock);
     if (m_read_buffer_index >= m_read_buffer->size && m_write_buffer->size != 0)
         flip();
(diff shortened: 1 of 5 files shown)


Kernel/DoubleBuffer.cpp
```diff
@@ -77,4 +77,24 @@ ssize_t DoubleBuffer::read(UserOrKernelBuffer& data, size_t size)
...
+    size_t nread = min(m_read_buffer->size - m_read_buffer_index, size);
+    if (!data.write(m_read_buffer->data + m_read_buffer_index, nread))
+        return -EFAULT;
+    compute_lockfree_metadata();
+    if (m_unblock_callback && m_space_for_writing > 0)
+        m_unblock_callback();
+    return (ssize_t)nread;
 }
+
+}
```
sin-ack, line 100: This won't pass lint, please make sure to add newlines on the last line of files.

Kernel/DoubleBuffer.cpp
```diff
@@ -77,4 +77,24 @@ ssize_t DoubleBuffer::read(UserOrKernelBuffer& data, size_t size)
     return (ssize_t)nread;
 }
 
+ssize_t DoubleBuffer::peek(UserOrKernelBuffer& data, size_t size)
+{
+    if (!size || m_storage.is_null())
+        return 0;
+    VERIFY(size > 0);
```
sin-ack, line 84: `VERIFY` doesn't do anything here, and `size > 0` wouldn't make sense when `m_storage.is_null()`.
```suggestion
    if (!size || m_storage.is_null())
        return 0;
```

IdanHo, line 84: This VERIFY is essentially useless, as the if directly before it ensures that it will never be hit (because size_t is unsigned)

sin-ack (changes requested):

Kernel/DoubleBuffer.cpp
sw1tchbl4d3r: Yeah, the original write() above does the same thing, I'll fix it for both.

sin-ack:
The `receive_packet_buffered` changes suggest that one can only peek one packet ahead. Is this the desired behavior?

sw1tchbl4d3r:
> The `receive_packet_buffered` changes suggest that one can only peek one packet ahead. Is this the desired behavior?

Testing on my computer suggests that, yes.

sin-ack:
Please squash your fixups into the parent commit :^)

sw1tchbl4d3r:
oops, forgot the VERIFY in peek() itself, let me just remove that as well really quick!

awesomekling:
Nice and clean :)

PR #6717, opened 2021-04-28
Ports: ffmpeg 4.4 (+ relevant LibC additions)

holdengreen:
Hi I've finally got something!

I've got ffmpeg 4.4 to run in SerenityOS.

It is now using a sha256 hash that I derived because I was having issues with gpg.

I have not extensively tested it in serenity yet, but it does seem to be missing media/library codecs that would make it more practically useful.

I'm still able to continue improving this port.

### DIFF
--- a/Ports/ffmpeg/package.sh
+++ b/Ports/ffmpeg/package.sh
@@ -1,0 +1,17 @@
+#!/usr/bin/env -S bash ../.port_include.sh
+port=ffmpeg
+version=4.4
+useconfigure=true
+files="https://ffmpeg.org/releases/ffmpeg-${version}.tar.gz ffmpeg-${version}.tar.gz a4abede145de22eaf233baa1726e38e137f5698d9edd61b5763cd02b883f3c7c"
+auth_type="sha256"
+installopts="INSTALL_TOP=${SERENITY_INSTALL_ROOT}/usr/local"
+configopts="SRC_PATH=."
+
+configure() {
+    run ./configure \
+        --arch="x86_32" \
+        --cc="${CC} -std=gnu99" \
+        --cxx="${CXX} -std=gnu99" \
+        --disable-network \
+        --enable-cross-compile
+}
--- a/Userland/Libraries/LibC/stdlib.h
+++ b/Userland/Libraries/LibC/stdlib.h
@@ -49,6 +49,7 @@ char* ptsname(int fd);
 int ptsname_r(int fd, char* buffer, size_t);
 int abs(int);
 long labs(long);
+long long int llabs(long long int);
 double atof(const char*);
 int system(const char* command);
 char* mktemp(char*);
--- a/Userland/Libraries/LibC/inttypes.h
+++ b/Userland/Libraries/LibC/inttypes.h
@@ -28,7 +28,9 @@ __BEGIN_DECLS
 #define PRIu32 "u"
 #define PRIu64 "llu"
 #define PRIx8 "b"
+#define PRIX8 "hhX"
 #define PRIx16 "w"
+#define PRIX16 "hX"
 #define PRIx32 "x"
 #define PRIX32 "X"
 #define PRIx64 "llx"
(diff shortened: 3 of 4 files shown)


Ports/ffmpeg/package.sh
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env -S bash ../.port_include.sh
+port=ffmpeg
+version=4.4
+useconfigure=true
+files="https://ffmpeg.org/releases/ffmpeg-4.4.tar.gz ffmpeg-4.4.tar.gz a08992a9f62cf9877582565544e8e799"
```
linusg, line 5:
```suggestion
files="https://ffmpeg.org/releases/ffmpeg-${version}.tar.gz ffmpeg-${version}.tar.gz a08992a9f62cf9877582565544e8e799"
```

Ports/ffmpeg/package.sh
```diff
@@ -0,0 +1,18 @@
...
+auth_type="md5"
```
linusg, line 6: `auth_type=md5` is not supported anymore, please use `sha256`

Ports/ffmpeg/package.sh
GMTA, line 5: You can interpolate with `${version}` here twice in the filenames

Ports/ffmpeg/package.sh
GMTA, line 6: Please use the `sha256` auth type

Ports/ffmpeg/package.sh
```diff
@@ -0,0 +1,18 @@
...
+installopts="INSTALL_TOP=${SERENITY_INSTALL_ROOT}/usr/local"
+configopts="SRC_PATH=."
+
+configure() {
+    run ./configure \
+		--arch="x86_32" \
+		--cc="i686-pc-serenity-gcc -std=gnu99" \
+		--cxx="i686-pc-serenity-g++ -std=gnu99" \
+		--disable-network \
+		--enable-cross-compile
```
linusg, line 17: - Please don't mix tabs and spaces (spaces only)
- Please don't hardcode `i686` (use `${CC}` and `${CXX}` instead of `i686-pc-serenity-gcc` and `i686-pc-serenity-g++`

Userland/Libraries/LibC/stdlib.h
```diff
@@ -49,6 +49,7 @@ char* ptsname(int fd);
 int ptsname_r(int fd, char* buffer, size_t);
 int abs(int);
 long labs(long);
+long long int llabs(long long int i);
```
linusg, line 52:
```suggestion
long long int llabs(long long int);
```

No need for the parameter name here

linusg (changes requested):
please split the first commit into two separate, atomic commits :)

linusg:
Also, the commit message should say "Ports", not "Port".

holdengreen:
Thanks guys! I think I've covered it...

Ports/ffmpeg/package.sh
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env -S bash ../.port_include.sh
+port=ffmpeg
+version=4.4
+useconfigure=true
+files="https://ffmpeg.org/releases/ffmpeg-${version}.tar.gz ffmpeg-${version}.tar.gz a4abede145de22eaf233baa1726e38e137f5698d9edd61b5763cd02b883f3c7c"
+auth_type="sha256"
+makeopts="-j$(nproc)"
```
awesomekling, line 7: This seems gratuitous, I don't think a random port should force parallelism on the build system.

awesomekling (changes requested):

Ports/ffmpeg/package.sh
holdengreen: Good call. I got that line from the lua package script. (which is what I originally copied and modified because I'm familiar with lua)

During testing I compiled ffmpeg without parallelism (to make debugging easier) and I will warn that it did take a non-negligible amount of time to compile.

I'll take that line out.

holdengreen: Ok. I took care of it!

PR #6726, opened 2021-04-29
TextEditor: Clear leftover whitespace when inserting newlines

euclidianAce:
Previously when entering a newline, previous indentation would be left,
leaving a line consisting only of whitespace. This fixes that

When watching Andreas build something inside of Serenity itself, it always irked me that there was a not insignificant amount of time correcting whitespace leftover by the autoindent. Hopefully this fixes that :^)

I haven't delved deep into the code for the TextEditor widget, so if there's a better way of doing this let me know!

### DIFF
--- a/Userland/Libraries/LibGUI/TextEditor.cpp
+++ b/Userland/Libraries/LibGUI/TextEditor.cpp
@@ -1244,7 +1244,23 @@ void TextEditor::insert_at_cursor_or_replace_selection(const StringView& text)
     VERIFY(is_editable());
     if (has_selection())
         delete_selection();
+
+    // Check if adding a newline leaves the previous line as just whitespace.
+    auto const clear_length = m_cursor.column();
+    auto const should_clear_last_line = text == "\n"
+        && clear_length > 0
+        && current_line().leading_spaces() == clear_length;
+
     execute<InsertTextCommand>(text, m_cursor);
+
+    if (should_clear_last_line) { // If it does leave just whitespace, clear it.
+        auto const original_cursor_position = cursor();
+        TextPosition start(original_cursor_position.line() - 1, 0);
+        TextPosition end(original_cursor_position.line() - 1, clear_length);
+        TextRange erased_range(start, end);
+        execute<RemoveTextCommand>(document().text_in_range(erased_range), erased_range);
+        set_cursor(original_cursor_position);
+    }
 }
 
 void TextEditor::cut()


Userland/Libraries/LibGUI/TextEditor.cpp
```diff
@@ -1244,7 +1244,23 @@ void TextEditor::insert_at_cursor_or_replace_selection(const StringView& text)
     VERIFY(is_editable());
     if (has_selection())
         delete_selection();
+
+    // check if adding a newline leaves the previous line as just whitespace
+    auto const clr_len = m_cursor.column();
+    auto const should_clear_last_ln = text == "\n"
```
awesomekling, line 1250: Please write out full words in variable names instead of using random abbreviations, unless the abbreviation makes it significantly clearer.

Userland/Libraries/LibGUI/TextEditor.cpp
```diff
@@ -1244,7 +1244,23 @@ void TextEditor::insert_at_cursor_or_replace_selection(const StringView& text)
...
+        && clr_len > 0
+        && current_line().leading_spaces() == clr_len;
+
     execute<InsertTextCommand>(text, m_cursor);
+
+    if (should_clear_last_ln) { // if it does leave just whitespace, clear it
+        auto const orig_cursor_pos = cursor();
```
awesomekling, line 1257: Same here.

awesomekling (changes requested):

Userland/Libraries/LibGUI/TextEditor.cpp
```diff
@@ -1244,7 +1244,23 @@ void TextEditor::insert_at_cursor_or_replace_selection(const StringView& text)
     VERIFY(is_editable());
     if (has_selection())
         delete_selection();
+
+    // check if adding a newline leaves the previous line as just whitespace
```
linusg, line 1248: Comments should look like sentences, with proper punctuation (https://github.com/SerenityOS/serenity/blob/master/Documentation/CodingStyle.md#comments)

```suggestion
    // Check if adding a newline leaves the previous line as just whitespace.
```

Userland/Libraries/LibGUI/TextEditor.cpp
```diff
@@ -1244,7 +1244,23 @@ void TextEditor::insert_at_cursor_or_replace_selection(const StringView& text)
...
+    auto const clear_length = m_cursor.column();
+    auto const should_clear_last_line = text == "\n"
+        && clear_length > 0
+        && current_line().leading_spaces() == clear_length;
+
     execute<InsertTextCommand>(text, m_cursor);
+
+    if (should_clear_last_line) { // if it does leave just whitespace, clear it
```
linusg, line 1256: same here

linusg (changes requested):

Userland/Libraries/LibGUI/TextEditor.cpp
```diff
@@ -803,7 +803,7 @@ void TextEditor::keydown_event(KeyEvent& event)
         return;
     }
 
-    if (event.key() == KeyCode::Key_Backspace) {
+    if (event.key() == KeyCode::Key_Backspace || event.code_point() == '\b') {
```
linusg, line 806: Why is this needed?

euclidianAce: Ah, that was my local fix for #6748. Didn't mean to commit it with this patch!

PR #6734, opened 2021-04-29
add groups program

setepenre:
add groups program listing group memberships for current user or users passed via command line ; modeled on groups from shadow-utils

I promoted get_gids function in LibCore/Account.cpp to be a static member function of Core::Account ; I don't know if it is the best move but it seemed to be the most "chirurgical" way to access the data I needed seeing as Core::Account::from_name/from_uid used /etc/shadows and thus needs superuser privilege.

### DIFF
--- a/Userland/Libraries/LibCore/Account.cpp
+++ b/Userland/Libraries/LibCore/Account.cpp
@@ -43,6 +43,7 @@ static Vector<gid_t> get_extra_gids(const passwd& pwd)
 {
     StringView username { pwd.pw_name };
     Vector<gid_t> extra_gids;
+    setgrent();
     for (auto* group = getgrent(); group; group = getgrent()) {
         if (group->gr_gid == pwd.pw_gid)
             continue;
--- a/Userland/Utilities/groups.cpp
+++ b/Userland/Utilities/groups.cpp
@@ -1,0 +1,82 @@
+/*
+ * Copyright (c) 2021, the SerenityOS developers.
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+#include <AK/Vector.h>
+#include <LibCore/Account.h>
+#include <LibCore/ArgsParser.h>
+#include <grp.h>
+#include <unistd.h>
+
+static void print_account_gids(const Core::Account& account)
+{
+    auto* gr = getgrgid(account.gid());
+    if (!gr) {
+        outln();
+        return;
+    }
+
+    out("{}", gr->gr_name);
+    for (auto& gid : account.extra_gids()) {
+        gr = getgrgid(gid);
+        out(" {}", gr->gr_name);
+    }
+    outln();
+}
+
+int main(int argc, char** argv)
+{
+    if (unveil("/etc/passwd", "r") < 0) {
+        perror("unveil");
+        return 1;
+    }
+
+    if (unveil("/etc/shadow", "r") < 0) {
+        perror("unveil");
+        return 1;
+    }
+
+    if (unveil("/etc/group", "r") < 0) {
+        perror("unveil");
+        return 1;
+    }
+
+    if (unveil(nullptr, nullptr) < 0) {
+        perror("unveil");
+        return 1;
+    }
+
+    if (pledge("stdio rpath", nullptr) < 0) {
+        perror("pledge");
+        return 1;
+    }
+
+    Vector<const char*> usernames;
+
+    Core::ArgsParser args_parser;
+    args_parser.set_general_help("Print group memberships for each username or, if no username is specified, for the current process.");
+    args_parser.add_positional_argument(usernames, "Usernames to list group memberships for", "usernames", Core::ArgsParser::Required::No);
+    args_parser.parse(argc, argv);
+
+    if (usernames.is_empty()) {
+        auto result = Core::Account::from_uid(geteuid());
+        if (result.is_error()) {
+            warnln("{}", result.error());
+            return 1;
+        }
+        print_account_gids(result.value());
+    }
+
+    for (auto username : usernames) {
+        auto result = Core::Account::from_name(username);
+        if (result.is_error()) {
+            warnln("{} '{}'", result.error(), username);
+            continue;
+        }
+        out("{} : ", username);
+        print_account_gids(result.value());
+    }
+    return 0;
+}
--- a/Base/usr/share/man/man1/groups.md
+++ b/Base/usr/share/man/man1/groups.md
@@ -1,0 +1,31 @@
+## Name
+
+groups - list group memberships
+
+## Synopsis
+
+```**sh
+$ groups [username...]
+```
+
+## Description
+
+`groups` lists group memberships.
+
+If no username is provided group memberships are listed for current user.
+
+## Arguments
+
+* username: username to list group memberships for
+
+## Examples
+
+```sh
+# List group memberships for current user
+$ groups
+# List group memberships for one user
+$ groups nona
+# List group memberships for multiple users
+$ groups nona anon root
+```
+


setepenre:
Ah yes, forgot to put a copyright header in Userland/Utilities/groups.cpp ; would 
```cpp
/*
 * Copyright (c) 2020, the SerenityOS developers.
 *
 * SPDX-License-Identifier: BSD-2-Clause
 */
```
be fine?

Userland/Libraries/LibCore/Account.cpp
```diff
@@ -31,7 +31,7 @@ static String get_salt()
     return builder.build();
 }
 
-static Vector<gid_t> get_gids(const StringView& username)
+Vector<gid_t> Account::get_gids(const StringView& username)
```
GMTA, line 34: I think the API here should have been `Account::from_name(...)->extra_gids()` instead of exposing this internal method?

Userland/Utilities/groups.cpp
```diff
@@ -0,0 +1,64 @@
...
+    args_parser.parse(argc, argv);
+
+    if (usernames.is_empty()) {
+        auto uid = geteuid();
+        auto* passwd = getpwuid(uid);
+        if (!passwd) {
+            fprintf(stderr, "Unknown user id '%d'\n", uid);
+            return 1;
+        }
+        auto gids = Core::Account::get_gids(passwd->pw_name);
```
GMTA, line 49: I think the API here should have been `Account::from_name(...)->extra_gids()` instead of exposing this internal method?

setepenre (commented):
"

Rada8732 (approved):

Userland/Utilities/groups.cpp
```diff
@@ -0,0 +1,70 @@
...
+    if (unveil("/etc/passwd", "r") < 0) {
+        perror("unveil");
+        return 1;
+    }
+
+    if (unveil("/etc/group", "r") < 0) {
+        perror("unveil");
+        return 1;
+    }
+
```
bcoles, line 35: `unveil` state should be locked once all necessary files have been unveiled.

```cpp
    unveil(nullptr, nullptr);
```

setepenre: Thanks! Change made and squashed ; should have read `man unveil` to its end :sweat_smile:

Userland/Utilities/groups.cpp
```diff
@@ -0,0 +1,75 @@
...
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+#include <grp.h>
+#include <pwd.h>
+#include <unistd.h>
+
+#include <AK/Vector.h>
+#include <LibCore/Account.h>
+#include <LibCore/ArgsParser.h>
```
linusg, line 13: no need to separate these.

```suggestion
#include <AK/Vector.h>
#include <LibCore/Account.h>
#include <LibCore/ArgsParser.h>
#include <grp.h>
#include <pwd.h>
#include <unistd.h>
```

Userland/Utilities/groups.cpp
```diff
@@ -0,0 +1,75 @@
...
+
+static void print_gid_list(const Vector<gid_t>& gids)
+{
+    for (auto gid : gids) {
+        auto* gr = getgrgid(gid);
+        printf("%s ", gr->gr_name);
```
linusg, line 19: would be nice to not print a trailing space here for the last GID :)

linusg (changes requested):
Please use `out`/`outln`/`warnln` instead of `printf`/`fprintf` in new code. Please also capitalize the first letter after `Foo: ` in commit messages.

setepenre:
Changes made and squashed into commit 9631f7b ; added a manpage also.
Thanks to all the reviewers for your time and suggestions :partying_face:

Userland/Utilities/groups.cpp
```diff
@@ -0,0 +1,77 @@
+/*
+ * Copyright (c) 2021, the SerenityOS developers.
```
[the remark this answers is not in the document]
setepenre: I chose this copyright because I don't want to put my own name / email :sweat_smile:

linusg:
With some recent changes to `Core::Account`, we can simplify this a bit - the first commit is no longer needed, we can simply construct an account from uid or username and access `gid()` and `extra_gids()`. Since the latter is now guaranteed to not include the former, we can unconditionally prepend a space to each extra GID, which enables use to use a range-base for loop :)

```diff
--- a/Userland/Utilities/groups.cpp
+++ b/Userland/Utilities/groups.cpp
@@ -8,17 +8,15 @@
 #include <LibCore/Account.h>
 #include <LibCore/ArgsParser.h>
 #include <grp.h>
-#include <pwd.h>
 #include <unistd.h>
 
-static void print_gid_list(const Vector<gid_t>& gids)
+static void print_account_gids(const Core::Account& account)
 {
-    for (size_t i = 0; i < gids.size(); ++i) {
-        auto* gr = getgrgid(gids[i]);
-        out("{}", gr->gr_name);
-        if (i != gids.size() - 1) {
-            out(" ");
-        }
+    auto* gr = getgrgid(account.gid());
+    out("{}", gr->gr_name);
+    for (auto& gid : account.extra_gids()) {
+        gr = getgrgid(gid);
+        out(" {}", gr->gr_name);
     }
     outln();
 }
@@ -53,25 +51,22 @@ int main(int argc, char** argv)
     args_parser.parse(argc, argv);
 
     if (usernames.is_empty()) {
-        auto uid = geteuid();
-        auto* passwd = getpwuid(uid);
-        if (!passwd) {
-            warnln("Unknown user id '{}'", uid);
+        auto result = Core::Account::from_uid(geteuid());
+        if (result.is_error()) {
+            warnln("{}", result.error());
             return 1;
         }
-        auto gids = Core::Account::get_gids(passwd->pw_name);
-        print_gid_list(gids);
+        print_account_gids(result.value());
     }
 
     for (auto username : usernames) {
-        auto* passwd = getpwnam(username);
-        if (!passwd) {
-            warnln("Unknown user '{}'", username);
+        auto result = Core::Account::from_name(username);
+        if (result.is_error()) {
+            warnln("{}", result.error());
             return 1;
         }
-        auto gids = Core::Account::get_gids(username);
         out("{} : ", username);
-        print_gid_list(gids);
+        print_account_gids(result.value());
     }
     return 0;
 }
```

setepenre:
Changes applied, history re-written! :partying_face:

Userland/Utilities/groups.cpp
```diff
@@ -0,0 +1,82 @@
...
+    auto* gr = getgrgid(account.gid());
+    if (!gr) {
+        outln();
+        return;
+    }
+
+    out("{}", gr->gr_name);
+    for (auto& gid : account.extra_gids()) {
+        gr = getgrgid(gid);
+        out(" {}", gr->gr_name);
```
linusg, line 24: If we need to do a nullptr check above, do we also need one here?

setepenre: The nullptr check at line 7 is due to user like `nona` not having a primary group ; `account.extra_gids()` seems to correctly return an empty array in this particular case. I don't think it is needed, but we can always add one for symetry? :sweat_smile:

linusg: That's not true, every user has a primary group (GID). For `nona` that's 200 as you can see in `/etc/passwd`. It might just not be present in `/etc/group`.
However, now that I think about it I believe this is is fine, as the extra GIDs should always match the entries in `/etc/group`, which `getgrgid()` (indirectly) reads from.

linusg:
Nice work!

Userland/Utilities/groups.cpp
setepenre: sorry I meant `nona` not having its group in `/etc/group`: 
image

PR #6735, opened 2021-04-29
Kernel: Enable building the kernel with -flto

gunnarbeutner:
GCC with `-flto` is more aggressive when it comes to inlining and discarding functions which is why we must mark some of the functions as `NEVER_INLINE` (because they contain asm labels which would be duplicated in the object files if the compiler decides to inline the function elsewhere) and `__attribute__((used))` for others so that GCC doesn't discard them.

This also adds a new CMake flag `ENABLE_KERNEL_LTO` that enables LTO when building the kernel.


Hendiadyoin1:
Wow I tried to do this today and was wondering what all these errors meant, god job!

gunnarbeutner:
`INTERPROCEDURAL_OPTIMIZATION` (https://cmake.org/cmake/help/latest/prop_tgt/INTERPROCEDURAL_OPTIMIZATION.html) is supposed to enable `-flto` (or other flags depending on your compiler, but then again we really only support GCC anyway).

As for `gcc-ar` vs. `ar` - you're right, we should be using `gcc-ar` and I've updated the PR to do that. For me it worked with regular `ar` as well though.

PR #6736, opened 2021-04-29
LibWeb: more ACID2 improvements

Cleverking2003:


TheDumpap:
The commit titles could be more descriptive. You seem to have included the actual change in the commit message body, but perhaps you could be more descriptive in the title and avoid that altogether? :)

Ex. "LibWeb: Fix floating box position calculation" -> "LibWeb: Include margins, borders and padding in floating box position calculation."

Cleverking2003:
> The commit titles could be more descriptive. You seem to have included the actual change in the commit message body, but perhaps you could be more descriptive in the title and avoid that altogether? :)
> 
> Ex. "LibWeb: Fix floating box position calculation" -> "LibWeb: Include margins, borders and padding in floating box position calculation."

I can't think of a message that would be 72 chars long :^)

awesomekling (changes requested):
These changes look reasonable, but please write a bit more in your commit messages so we know what you're doing and why.

Also please wrap commit messages at 72 characters, this text:
```
Now we include margins, borders and paddings
into calculating floating box positions
```

Should be wrapped like this:
```
Now we include margins, borders and paddings into calculating floating
box positions
```

PR #6738, opened 2021-04-29
Libcore: IODevice - fixed line buffering

### LINKED ISSUE #5907: grep: no matches if file contains no CR/LF newline
lies

r-paiva:
Fixes #5907.

Two issues are fixed with this request.

1. When grep reads from stdin, it discarded the last character of a line. b3f6313 

- lines with \n, this worked fine.
- lines without \n, it discarded the last character of that line, which caused grep to fail to match.

2. When grep searches recursevely, it uses IODevice to buffer the data in. f2bd874

- IODevice::can_read_line only buffers 1024 bytes at a time 
    - If a line is bigger than 1024 (with or without a \n) can_read_line would return false.
- IODevice::can_read_line At eof it only returns true if data has a \n
    - if a line at the end of a file does  not have a \n, can_read_line would return false, which I don't think was the intended behaviour.

  Changed IODevice::can_read_line to buffer data 1024 bytes at time until it either finds a \n in the buffered data, or,
  EOF is reached.

before and after:

bad png
good png

I've never written anything in C++ before, and it's my first time contributing, so I might be missing something.
Thanks!

### DIFF
--- a/Userland/Libraries/LibCore/IODevice.cpp
+++ b/Userland/Libraries/LibCore/IODevice.cpp
@@ -105,14 +105,31 @@ bool IODevice::can_read_line() const
 {
     if (m_eof && !m_buffered_data.is_empty())
         return true;
+
     if (m_buffered_data.contains_slow('\n'))
         return true;
+
     if (!can_read_from_fd())
         return false;
-    populate_read_buffer();
-    if (m_eof && !m_buffered_data.is_empty())
-        return true;
-    return m_buffered_data.contains_slow('\n');
+
+    while (true) {
+        // Populate buffer until a newline is found or we reach EOF.
+
+        auto previous_buffer_size = m_buffered_data.size();
+        populate_read_buffer();
+        auto new_buffer_size = m_buffered_data.size();
+
+        if (m_error)
+            return false;
+
+        if (m_eof)
+            return !m_buffered_data.is_empty();
+
+        if (m_buffered_data.contains_in_range('\n', previous_buffer_size, new_buffer_size - 1))
+            return true;
+    }
+
+    return true;
 }
 
 bool IODevice::can_read() const
--- a/AK/Vector.h
+++ b/AK/Vector.h
@@ -150,6 +150,19 @@ public:
         return false;
     }
 
+    bool contains_in_range(const T& value, const size_t start, const size_t end) const
+    {
+        VERIFY(start <= end);
+        VERIFY(end < size());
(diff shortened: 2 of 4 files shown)


Userland/Libraries/LibCore/IODevice.cpp
```diff
@@ -108,14 +108,32 @@ bool IODevice::can_read_line() const
...
+
     if (!can_read_from_fd())
         return false;
-    populate_read_buffer();
-    if (m_eof && !m_buffered_data.is_empty())
-        return true;
-    return m_buffered_data.contains_slow('\n');
+
+    while (true) {
+        // populate buffer until a \n is found or we reach eof.
```
linusg, line 119: Comments should be complete sentences:

```suggestion
        // Populate buffer until a newline is found or we reach EOF.
```

Userland/Libraries/LibCore/IODevice.cpp
```diff
@@ -108,14 +108,32 @@ bool IODevice::can_read_line() const
...
+        populate_read_buffer();
+        auto new_buffer_size = m_buffered_data.size();
+
+        if (m_error)
+            return false;
+
+        if (m_eof)
+            return !m_buffered_data.is_empty();
+
+        // search new data for a \n
```
linusg, line 131: This done doesn't add any value.

```suggestion
```

linusg (changes requested):
- The addition of `contains_in_range` should be in its own commit. Please also add a test case in `TestVector.cpp`.
- Please treat spelling and punctuation more care - see https://github.com/SerenityOS/serenity/blob/master/CONTRIBUTING.md#human-language-policy. Specifically:
  - `if a line` -> `If a line`
  - `ended wihtout` -> `ended without`
  - `eof` -> `EOF`
  - `fixes` -> `Fixes`
- Please follow the existing pattern for commit message pefixes - instead of `Libraries-IODevice`, do something like `LibCore: Let IODevice::can_read_line() buffer until \n or EOF`

Userland/Libraries/LibCore/IODevice.cpp
r-paiva: done

alimpfard (approved):
This seems okay.
I'd personally have switched the backing buffer mechanism to `AK::DuplexMemoryStream` to avoid allocating possibly a really big contiguous buffer, but perhaps that can be another commit/PR since it doesn't really have anything to do with fixing line buffering.

PR #6740, opened 2021-04-29
LibWeb: Add selector support to the "new" CSSParser

stelar7:
Adds a "naive" implementation of selector parsing.
This is stolen from the old parser, but it seems to parse fine :^)

### DIFF
--- a/Userland/Libraries/LibWeb/CSS/Parser/Parser.cpp
+++ b/Userland/Libraries/LibWeb/CSS/Parser/Parser.cpp
@@ -13,6 +13,7 @@
 #include <LibWeb/CSS/Parser/StyleComponentValueRule.h>
 #include <LibWeb/CSS/Parser/StyleFunctionRule.h>
 #include <LibWeb/CSS/Selector.h>
+#include <LibWeb/Dump.h>
 
 #define CSS_PARSER_TRACE 1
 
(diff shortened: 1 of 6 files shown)


stelar7:
Seems like its missing a clang-format.. 👀

Userland/Libraries/LibWeb/CSS/Parser/Parser.cpp
```diff
@@ -76,15 +77,232 @@ Vector<QualifiedStyleRule> Parser::parse_as_stylesheet()
...
     return rules;
 }
 
 Vector<CSS::Selector::ComplexSelector> Parser::parse_selectors(Vector<String> parts)
 {
-    (void)parts;
+    // TODO:
+    // This is a mess because the prelude is parsed as a string.
+    // It should really be parsed as its class, but the cpp gods have forsaken me
+    // and i cant make it work due to cyclic includes.
```
alimpfard, line 92: Cyclic includes?
missing `#pragma once`?

stelar7: Its classes that depends on each other via. a vector
```
// Class B
Vector<ClassA>

// Class A
Vector<ClassB>
```

I tried to forward declare them way back when i first started the parser, but just gave up because it wouldnt work

Issue #6670, opened 2021-04-26
Black screen after boot


diversys:
This is ba9a285d56ec364e3e5b63e2f3c1e7677bfaa738 but started a few days ago.
Host is macOS 10.14.6

```
0/2] Re-checking globbed directories...
[1/2] Re-running CMake...
-- Configuring done
-- Generating done
-- Build files have been written to: /Users/diver/serenity/Build
[0/2] Re-checking globbed directories...
[22/2278] Linking CXX static library Meta/Lagom/libLagomCore.a
/Library/Developer/CommandLineTools/usr/bin/ranlib: file: Meta/Lagom/libLagomCore.a(Command.cpp.o) has no symbols
/Library/Developer/CommandLineTools/usr/bin/ranlib: file: Meta/Lagom/libLagomCore.a(FileWatcher.cpp.o) has no symbols
/Library/Developer/CommandLineTools/usr/bin/ranlib: file: Meta/Lagom/libLagomCore.a(Command.cpp.o) has no symbols
/Library/Developer/CommandLineTools/usr/bin/ranlib: file: Meta/Lagom/libLagomCore.a(FileWatcher.cpp.o) has no symbols
[1715/1716] Install the project...
-- Install configuration: ""
[0/2] Re-checking globbed directories...
[0/1] cd /Users/diver/serenity/Build && /usr/local/Cellar/cmake/3.20.1/...enity SERENITY_ARCH=i686 /Users/diver/serenity/Meta/build-image-qemu.sh
checking existing image
e2fsck 1.46.2 (28-Feb-2021)
Pass 1: Checking inodes, blocks, and sizes
Pass 2: Checking directory structure
Pass 3: Checking directory connectivity
Pass 3A: Optimizing directories
Pass 4: Checking reference counts
Pass 5: Checking group summary information

_disk_image: ***** FILE SYSTEM WAS MODIFIED *****
_disk_image: 5631/46736 files (1.0% non-contiguous), 84582/186368 blocks
done
mounting filesystem... done
installing base system... done
creating initial filesystem structure... done
creating utmp file... done
setting up device nodes folder... done
writing version file... done
installing users... done
adding some desktop icons...done
installing shortcuts... done
installing 'checksum' variants... done
unmounting filesystem... done
[0/2] Re-checking globbed directories...
[0/1] cd /Users/diver/serenity/Build && /Users/diver/serenity/Meta/run.sh
qemu-system-x86_64: multiboot knows VBE. we don't
[...]

PR #6745, opened 2021-04-29
CMake: Fix building libraries on macOS

gunnarbeutner:
When building libraries on macOS they'd be missing the `SONAME` attribute which causes the linker to embed relative paths into
other libraries and executables:

```
Dynamic section at offset 0x52794 contains 28 entries:
 Type     Name/Value
(NEEDED) Shared library: [libgcc_s.so]
(NEEDED) Shared library: [Userland/Libraries/LibCrypt/libcrypt.so]
(NEEDED) Shared library: [Userland/Libraries/LibCrypto/libcrypto.so]
(NEEDED) Shared library: [Userland/Libraries/LibC/libc.so]
(NEEDED) Shared library: [libsystem.so]
(NEEDED) Shared library: [libm.so]
(NEEDED) Shared library: [libc.so]
```

The dynamic linker then fails to load those libraries which makes the system unbootable.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -139,9 +139,10 @@ foreach(lang ASM C CXX OBJC OBJCXX)
     unset(CMAKE_${lang}_OSX_DEPLOYMENT_TARGET_FLAG)
     unset(CMAKE_${lang}_SYSROOT_FLAG)
     if (CMAKE_SYSTEM_NAME MATCHES Darwin)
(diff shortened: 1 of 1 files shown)


linusg:
Is this related to / fixing #6670?

nooga:
@linusg yes, I'm testing this rn.

diversys:
I confirm, this fixes #6670. Thanks!

nooga:
Works 🎉

linusg:
Great!

Issue #5312, opened 2021-02-12
Feature Request: VDSO


ccapitalK:
Currently, calling gettimeofday() has syscall overhead, which makes it less accurate as a timing mechanism. If we introduce something like Linux's vDSO and map the pages containing the hardware clock information as read-only into userspace, we could potentially reduce the overhead to be that of a mov instruction (+ processing costs).

tomuta:
I had done some implementation, but not the exact same: #2697 It didn't map the hardware clock registers, but it was a globally accessible page.

ccapitalK:
I've been looking into ways of doing this, I don't believe there is really any way to do this since we don't allow RDTSC in userspace.

I've looked into the way that Linux does it in it's VDSO, it basically reads whatever high precision timer is available (usually using RDTSC on x86, unless it is configured to use a paravirtualized time device) then combines it with some contextual information in the kernel's vvar page containing the current timezone/value of the TSC at the start of the slice to convert it into an actual timestamp. A while ago (before 2018), Linux used to do the thing where it maps the actual timer registers for the HPET into userspace, however apparently they removed that feature entirely because:
    a) it was apparently unreliable on some DELL machines
    b) reading from the HPET registers takes "thousands of cycles", which means that we might as well use a syscall at that point.

I think if we were to implement this, maybe we could do an approximate solution (to emulate the lack of precision we try to enforce by banning RDTSC). Maybe something like recording the current time at the start of the scheduler tick/last syscall, then adding some bogus number on top of it each time we call gettimeofday()?

ccapitalK:
https://github.com/SerenityOS/serenity/commit/fdfc66db618a59242db34fd6b59e5ada095190ab seems to have implemented most of this. The only outstanding issue is that we may want to make it more accurate (current implementation only updates when  `TimeManagement::increment_time_since_boot_hpet()` is called in the kernel, so we can't use this for fine grained timing for e.g. microbenchmarking).

PR #6746, opened 2021-04-29
Ports: Fix networking for OpenTTD

gunnarbeutner:
OpenTTD incorrectly only checks for `EWOULDBLOCK` while it _should_ also check for `EAGAIN`. With this patch applied joining network games works.

The game calls `gettimeofday()` so many times per second that the game becomes unusable after joining a reasonably active network game. So I've patched that out for now.


ccapitalK:
Since this is a general posix compatibility issue on OpenTTD's side, I believe we should submit a patch upstream to fix this. Do you want to create the upstream patch, or should I?

awesomekling:
How about an alternate solution:
```cpp
#define EWOULDBLOCK EAGAIN
```

That `gettimeofday()` hack is unfortunate. The best fix would be #5312 but we can do this in the meantime.

gunnarbeutner:
Defining `EWOULDBLOCK` like that makes perfect sense. Apparently that's also how eglibc and newlib do it. I'll update the PR in a bit. :+1:

gunnarbeutner:
> Since this is a general posix compatibility issue on OpenTTD's side, I believe we should submit a patch upstream to fix this. Do you want to create the upstream patch, or should I?
> 
> Note: I believe we'd still need to merge this commit if we do decide to upstream it, since we'd need to wait until OpenTTD's next release anyway.

Yeah, technically this is a bug. Feel free to upstream the patch.

ccapitalK:
> > Since this is a general posix compatibility issue on OpenTTD's side, I believe we should submit a patch upstream to fix this. Do you want to create the upstream patch, or should I?
> > Note: I believe we'd still need to merge this commit if we do decide to upstream it, since we'd need to wait until OpenTTD's next release anyway.
> 
> Yeah, technically this is a bug. Feel free to upstream the patch.

Reported upstream https://github.com/OpenTTD/OpenTTD/pull/9142

gunnarbeutner:
Awesome, thank you!

TrueBrain:
Tnx for the upstream @ccapitalK , always appreciated!

Something I am curious about, and I have no clue if this is the place to ask: OpenTTD compiles and runs on many OSes and in 17 years that line of code exists, this has never been mentioned as a problem. So that makes me very curious to what is going on :D To be clear, POSIX clearly defines both should be expected, so a fix for that surely is the least we can do :) But it does make me curious why serenity doesn't map `EWOULDBLOCK` to `EAGAIN`. I would also expect this to be an issue for many more applications, as I would expect many would silently assume all OSes are just mapping those two to the same value :D
Very specifically, I wonder if serenity just always fires `EGAIN` and never `EWOULDBLOCK` (that it just picked either of the two values), or that it can fire both depending on some internal situation? As I can imagine a simple fix is an `#ifdef` that makes sure those two values are the same in one of our headers, similar to what @awesomekling suggested.
Sorry, long story, but these problems are fun, so it sparkles my curiosity a lot :)

Regarding the `gettimeofday()`, now that is interesting :) We have had reports that the framerate measuring tool can make the game slow in certain situations, but we never been able to pin it down. You might be on to something here.
Makes me wonder .. we also use a lot of `steady_clock`; this isn't an issue? It is only `high_resolution_clock` that is triggering `gettimeofday`? (`std::chrono` is a bit of a shit-show between OSes, because the specs leave a lot of room for the implementation :D)

Either way, if you find any other issues / weirdness / etc, feel free to poke me or create a ticket on our GitHub project; we are happy to help :)

gunnarbeutner:
I guess no particular thought was spent on whether `EAGAIN` and `EWOULDBLOCK` should be the same value internally. Someone probably just added them to `<errno.h>` and was happy things worked at all.

The `eglibc` documentation at https://www.gnu.org/software/libc/manual/html_node/Error-Codes.html says this about `EAGAIN`:

```
Portability Note: In many older Unix systems, this condition was indicated by EWOULDBLOCK,
which was a distinct error code different from EAGAIN. To make your program portable, you
should check for both codes and treat them the same.
```

As far as our kernel is concerned we never return `EWOULDBLOCK` and always use `EAGAIN`.

As for `gettimeofday`, well, `steady_clock` is calling that too, internally. I just commented out the most prominent call that came up in our profiler, because I read in a comment for `ottd_rdtsc` that those performance timers aren't that important. Our kernel doesn't allow `rdtsc` for user-mode programs, so I had to patch that out:

```patch
 /* rdtsc for all other *nix-en (hopefully). Use GCC syntax */
-#if (defined(__i386__) || defined(__x86_64__)) && !defined(RDTSC_AVAILABLE)
+#if (defined(__i386__) || defined(__x86_64__)) && !defined(__serenity__) && !defined(RDTSC_AVAILABLE)
 uint64 ottd_rdtsc()
```

Oh, btw, I think we ran into each other on IRC a few days ago when I tried to debug that SDL issue. Hello again!

TrueBrain:
> 
> 
> I guess no particular thought was spent on whether `EAGAIN` and `EWOULDBLOCK` should be the same value internally. Someone probably just added them to `<errno.h>` and was happy things worked at all.
> 

That is fair :) I do wonder how many more applications no longer follow the portability note .. as said, it is really rare these days :) Doesn't make it less wrong on our side, ofc :)

> 
> As for `gettimeofday`, well, `steady_clock` is calling that too, internally. I just commented out the most prominent call that came up in our profiler, because I read in a comment for `ottd_rdtsc` that those performance timers aren't that important. Our kernel doesn't allow `rdtsc` for user-mode programs, so I had to patch that out:
> 

In general, the performance timers you mention are to show to users if their game is slow, whereabouts the problem might be. So we constantly measure the whole code to have that data ready when a user opens the window. So yeah, they are not important as such. Having a RDTSC reduces the pain somewhat, as those should be a lot less expensive to call. Either way, not really ideal, and one can wonder if we shouldn't start measuring only when the framerate window is open instead of always. Something to think about on our end :)

> 
> Oh, btw, I think we ran into each other on IRC a few days ago when I tried to debug that SDL issue. Hello again!

Hi :D

gunnarbeutner:
There, removed the patch for the port and defined `EWOULDBLOCK` as an alias for `EAGAIN`.

PR #6747, opened 2021-04-29
LibC: Add missing initialization for the FILE mutex

gunnarbeutner:
Oops. I'm surprised it worked as well as it did.


gunnarbeutner:
Well, it looks like it did get zero-initialized most of the time, but not always. OpenTTD froze sometimes while doing stdio.

PR #6753, opened 2021-04-30
Ports: Quake II Port :^)

Quaker762:
Two is better than one!

### DIFF
--- a/Ports/quake2/package.sh
+++ b/Ports/quake2/package.sh
@@ -1,0 +1,20 @@
+#!/usr/bin/env -S bash ../.port_include.sh
+port=quake2
+version=0.1
+workdir=SerenityQuakeII-master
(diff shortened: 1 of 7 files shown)


Ports/quake2/package.sh
```diff
@@ -0,0 +1,23 @@
...
+configure() {
+    run cmake $configopts
+}
+
+install() {
+    run make install
+}
+
+# FIXME: Uhh, the things in this directory are not supposed to run on the host, why add this to $PATH?!
+export PATH="${SERENITY_BUILD_DIR}/Root/usr/bin:${PATH}"
```
awesomekling, line 23: What's going on here?

Quaker762: Whoops!

PR #6755, opened 2021-04-30
Meta: pre-commit should run lint-ports.py only when Ports change

bgianfo:
Most of the existing lint-ing shell scripts have the ability
to only run on the files which have actually changed.
The new port lint-ing script doesn't have this functionality
unfortunately. This forces us to lint ALL the ports on every
single change to any other file in the system if you have
the pre-commit hook setup for your git clone locally.

Instead we can use pre-commit's feature to only run a hook
if certain files have changed to reduce the situations in
which we would run the Meta/lint-ports.py script.

This shaves off a few seconds for me locally when committing.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -4,4 +4,12 @@ repos:
       - id: meta-lint-ci
         name: Running Meta/lint-ci.sh to ensure changes will pass linting on CI
         entry: bash Meta/lint-ci.sh
+        args: [ --no-ports ]
+        language: system
+
+      - id: meta-lint-ports
+        name: Running Meta/lint-ports.py to ensure changes will pass linting on CI
+        entry: Meta/lint-ports.py
+        pass_filenames: false
+        files: ^Ports/
         language: system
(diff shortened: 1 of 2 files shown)


bgianfo:
CC: @gunnarbeutner

.pre-commit-config.yaml
```diff
@@ -4,4 +4,12 @@ repos:
       - id: meta-lint-ci
         name: Running Meta/lint-ci.sh to ensure changes will pass linting on CI
         entry: bash Meta/lint-ci.sh
+        args: [ --no-ports ]
+        language: system
+
+      - id: ports-lint-ci
```
linusg, line 10: Let's follow the previous naming pattern and call this `meta-lint-ports`

Meta/lint-ci.sh
```diff
@@ -42,6 +47,16 @@ else
     ((FAILURES+=1))
 fi
 
+
+if [ "$ports" = true ]; then
```
linusg, line 51: Short comment for our future selves about why this was split from the other lint scripts would be great :)

PR #6756, opened 2021-04-30
Kernel: Harden more Vector usage against OOM

bgianfo:
Another series of commits working towards: #6369
There's some more details in the individual commit messages.

-   __Kernel: Harden LocalSocket Vector usage against OOM.__

 -   __Kernel: Harden Socket Vector usage against OOM__

  -  __Kernel: Harden sys$setgroups Vector usage against OOM__

  -  __Kernel: Harden DevFS Vector usage against OOM.__

     The dance here is not complicated, but it is something that should
     be taken note of. Since we append to both lists, we don't want to
     orphan the new Inode in the m_links/m_subfolders Vector in the event
     that the append to m_parent_fs.m_nodes fails.

  -  __Kernel: Harden Process Vector usage against OOM.__

 -   __Kernel: Harden Ext2FileSystem Vector usage against OOM.__

### DIFF
--- a/Kernel/FileSystem/DevFS.cpp
+++ b/Kernel/FileSystem/DevFS.cpp
@@ -275,6 +275,10 @@ KResultOr<NonnullRefPtr<Inode>> DevFSRootDirectoryInode::create_child(const Stri
         if (name != "pts")
             return EROFS;
         auto new_directory_inode = adopt_ref(*new DevFSPtsDirectoryInode(m_parent_fs));
+        if (!m_subfolders.try_ensure_capacity(m_subfolders.size() + 1))
+            return ENOMEM;
+        if (!m_parent_fs.m_nodes.try_ensure_capacity(m_parent_fs.m_nodes.size() + 1))
+            return ENOMEM;
         m_subfolders.append(new_directory_inode);
         m_parent_fs.m_nodes.append(new_directory_inode);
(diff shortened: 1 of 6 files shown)


Kernel/FileSystem/DevFS.cpp
```diff
@@ -275,8 +275,13 @@ KResultOr<NonnullRefPtr<Inode>> DevFSRootDirectoryInode::create_child(const Stri
         if (name != "pts")
             return EROFS;
         auto new_directory_inode = adopt_ref(*new DevFSPtsDirectoryInode(m_parent_fs));
-        m_subfolders.append(new_directory_inode);
-        m_parent_fs.m_nodes.append(new_directory_inode);
+        if (!m_subfolders.try_append(new_directory_inode))
+            return ENOMEM;
+        if (!m_parent_fs.m_nodes.try_append(new_directory_inode)) {
+            // Undo append above.
+            m_subfolders.take_last();
```
awesomekling, line 282: Instead of this, can we instead use `try_ensure_capacity` first, which would then guarantee that the append succeeds?

bgianfo: Yea, makes sense. I was hoping to eventually ifdef out `AK::Vector<T>::append(..)` and friends in kernel mode... but maybe that's a pipe dream.

PR #6757, opened 2021-04-30
Browser: Add FrogFind search engine

sppmacd:
Added FrogFind search engine. See relevant comment on Discord:
https://discord.com/channels/830522505605283862/830525031720943627/837641474865889300

### DIFF
--- a/Userland/Applications/Browser/Tab.cpp
+++ b/Userland/Applications/Browser/Tab.cpp
@@ -441,6 +441,7 @@ Tab::Tab(Type type)
     // FIXME: Support adding custom search engines
     add_search_engine("Bing", "https://www.bing.com/search?q={}");
     add_search_engine("DuckDuckGo", "https://duckduckgo.com/?q={}");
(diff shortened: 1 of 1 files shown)


Userland/Applications/Browser/Tab.cpp
```diff
@@ -45,7 +45,9 @@ String g_search_engine = {};
 URL url_from_user_input(const String& input)
 {
     if (input.starts_with("?") && !g_search_engine.is_null()) {
-        return URL(String::formatted(g_search_engine, urlencode(input.substring(1))));
+        String query = input.substring(1);
+        query.replace(" ", "+", true);
+        return URL(String::formatted(g_search_engine, urlencode(query)));
```
linusg, line 50: This seems odd - so when I search for `1 + 1`, it'll replace that with `1   1`? Please at least explain the reasoning for this change in the commit message :)

linusg, line 50: why is this necessary - shouldn't urlencode'ing the input on its own be fine?

Issue #6758, opened 2021-04-30
Kernel: Deadlock in NetworkTask


gunnarbeutner:
I managed to get the `NetworkTask` blocked in a `Kernel::ARPTableBlocker`:

```
(gdb) p ((Kernel::Thread*)0xc0b0b404)->m_name
$21 = "NetworkTask"
(gdb) p ((Kernel::Thread*)0xc0b0b404)->m_in_block
$22 = true
(gdb) p *((Kernel::Thread*)0xc0b0b404)->m_blocker
$23 = {
  _vptr.Blocker = 0xc084f64c <vtable for Kernel::ARPTableBlocker+8>,
  m_lock = {
    m_lock = 0,
    m_recursions = 0
  },
  m_block_condition = 0xc0b754a4 <kmalloc_pool_heap+1053860>,
  m_block_data = 0x0,
  m_blocked_thread = 0xc0b0b404 <kmalloc_pool_heap+619524>,
  m_was_interrupted_by_signal = 0 '\000',
  m_is_blocking = true,
  m_was_interrupted_by_death = false,
  m_did_timeout = false
}
```

I _think_ that can happen for a TCP socket which is in the closing state and we try to do `TCPSocket::send_tcp_packet` from the `NetworkTask` while there's no ARP entry for the IP address that belongs to the peer of that socket.

And then we deadlock onto ourselves because it'd be the `NetwerkTask` that receives the ARP reply (if any).

This is likely the case for a VM that is bridged to a LAN, i.e. where packets are directly sent - rather than via the default gateway whose ARP entry is usually in the ARP table.

PR #6762, opened 2021-04-30
Kernel: Various network fixes

gunnarbeutner:
This fixes a number of issues with the network stack:

* Avoid deadlock when trying to send packets from the NetworkTask - see #6758 for details

* Don't put closed/listener sockets into the `closing_sockets` list. Because once they're in the list they become immortal (as seen with `SystemMonitor`: Try starting `nc -l 0 7000` and then hit Ctrl-C. Observe how the listener socket doesn't get destroyed.

* Remove socket from the listener's accept list when it is closed.

Without this patch we end up with sockets in the listener's accept queue with state 'closed' when doing stealth SYN scans:

Client -> Server: SYN for port 22
Server -> Client: SYN/ACK
Client -> Server: RST (i.e. don't complete the TCP handshake, socket stays in the list and shows up as `Closed` in `SystemMonitor`)

* Record MAC addresses for incoming IPv4 packets.

This way we don't have to do ARP just to send packets back to an address which just sent us a packet.

* Tear down connections when we receive an RST packet.

Start `nc -l 0 7000` and connect to it from another host. Hit Ctrl-C on that other host and observe how `nc` in SerenityOS doesn't realize that the connection was closed.

### DIFF
--- a/Kernel/Net/NetworkTask.cpp
+++ b/Kernel/Net/NetworkTask.cpp
@@ -528,6 +528,11 @@ void handle_tcp(const IPv4Packet& ipv4_packet, const Time& packet_timestamp)
             return;
         }
     case TCPSocket::State::Established:
+        if (tcp_packet.has_rst()) {
+            socket->set_state(TCPSocket::State::Closed);
+            return;
+        }
+
         if (tcp_packet.has_fin()) {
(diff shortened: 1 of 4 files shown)


awesomekling:
Great fixes all around!

PR #6768, opened 2021-05-01
Userland: Remove an unused function declaration from tac.

Mango0x45:
The function declaration seems to have come from an earlier commit. The function is no longer here though, so there's no reason to keep this around.


linusg:
Please wrap commit messages so they don't exceed 72 characters per line.

PR #6769, opened 2021-05-01
Userland: Add the rev(1) utility to reverse lines and it's manpage.

Mango0x45:
My first time writing C++ :)
I made sure to test it, and it worked perfectly on multiple files, piped input, and redirected files.

Also I accidentally made a typo in one of the commit messages, saying I added a manpage for `man` instead of `rev`, whoops.

### DIFF
--- a/Base/usr/share/man/man1/rev.md
+++ b/Base/usr/share/man/man1/rev.md
@@ -1,0 +1,41 @@
+## Name
+
+rev - reverse lines
+
+## Synopsis
+
+```*sh
+$ rev [files...]
+```
+
+## Description
+
+`rev` reads the specified files line by line, and prints them to standard
+output with each line being reversed characterwise. If no files are specifed,
+then `rev` will read from standard input.
+
+## Examples
+
+To print two files 'foo' and 'bar' in reverse:
+```sh
+$ cat foo bar
+foo 1
+foo 2
+bar 1
+bar 2
+$ rev foo bar
+1 oof
+2 oof
+1 rab
+2 rab
+```
+
+To list files with their names in reverse:
+```sh
+$ ls
+foo
+bar
+$ ls | rev
+oof
+rab
+```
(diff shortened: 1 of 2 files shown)


Base/usr/share/man/man1/rev.md
```diff
@@ -0,0 +1,27 @@
...
+
+## Synopsis
+
+```*sh
+$ rev [files...]
+```
+
+## Description
+
+`rev` reads the specified files line by line, and prints them to standard
```
dj2, line 13: Maybe something like `rev reads the specified files line by line and prints each line reversed to standard output`? When I read this the first time I thought it was going to print the lines in the right character order, but the file would be printed bottom to top.

Base/usr/share/man/man1/rev.md
```diff
@@ -0,0 +1,27 @@
...
+output in reverse. If no files are specifed, then `rev` will read from standard
+input.
+
+## Examples
+
```
dj2, line 18: An example like the following would make it a bit more explicit?

```sh
$ cat foo
line1
line2
$ rev foo
1enil
2enil
```

Base/usr/share/man/man1/rev.md
Mango0x45: I clarified this in my latest commit.

Base/usr/share/man/man1/rev.md
Mango0x45: Good idea. I did this for both examples.

Base/usr/share/man/man1/rev.md
dj2: Awesome, thank you.

linusg:
Please observe the [contribution guidelines](https://github.com/SerenityOS/serenity/blob/master/CONTRIBUTING.md), specifically please remove the merge commit and squash any fixup commits into the appropriate previous commits.

sin-ack:
I'm pretty sure we already have `tac`, no? Does this do something else?

Mango0x45:
After some time googling around and figuring out how git works I managed to squash the commits. Good learning experience :P

PR #6771, opened 2021-05-01
Base + Userland: Enhance bt command to dump all threads in a process, add a man page

bgianfo:
-    __bt: Fix generation of click-able links for source files.__

     When the default build location was moved from /Build to the new
     architecture specific directory, /Build/i686, this code broke.
     All file names are now path relative one additional level up.

     So continue the hack, and introduce another dummy directory to
     make the relative paths resolve correctly.

-   __bt: Enumerate all threads when symbolizing stacks in the bt utility__

    Enumerate tid's from /proc/{pid}/stacks/ and use it to print the
    backtrace for all active threads in the specified process.

-    __Base: Add a man page for the 'bt' command.__

### DIFF
--- a/Base/usr/share/man/man1/bt.md
+++ b/Base/usr/share/man/man1/bt.md
@@ -1,0 +1,44 @@
+## Name
+
+bt - view the backtrace of the specified process
+
+## Synopsis
+
+```**sh
+$ bt <pid>
+```
+
+## Description
+
+This program is used to inspect the current executable state of a process.
+It will read the stack of each thread in the process, and symbolize the
+addresses for each frame in the stack producing a backtrace.
+
+**NOTE**:
+
+* Kernel addresses will not be available unless you are super user.
+
+* If Kernel addresses are available, they will not be symbolized unless the
+  `SymbolServer` service has access to the `/boot/Kernel` file.
+
+## Examples
+
+View all stacks of pid number 10:
+
+```sh
+$ bt 10
+```
+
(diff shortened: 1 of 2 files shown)


Userland/Utilities/bt.cpp
```diff
@@ -54,31 +55,42 @@ int main(int argc, char** argv)
...
+    Core::DirIterator iterator(String::formatted("/proc/{}/stacks", pid), Core::DirIterator::SkipDots);
+    if (iterator.has_error()) {
+        warnln("Error: pid '{}' doesn't appear to exist.", pid);
+        return 1;
+    }
+
+    while (iterator.has_next()) {
+        pid_t tid = iterator.next_path().to_int().value();
+        outln("tid: {}", tid);
+        // FIXME: Support multiple threads in the same process!
```
bgianfo, line 67: whoops

Base/usr/share/man/man1/bt.md
```diff
@@ -0,0 +1,44 @@
+## Name
+
+bt - View the backtrace of the specified process id
```
linusg, line 3:
```suggestion
bt - view the backtrace of the specified process
```

(lowercase just to match existing pattern)

linusg (commented):
tiny nitpick for the manpage commit :)

Base/usr/share/man/man1/bt.md
bgianfo: Fixed, thanks!

PR #6773, opened 2021-05-01
Shell: Fix some bugs found by oss-fuzz

alimpfard:
~~Also work around what appears to be a weird compiler bug:~~

Turns out it's just good old CxByte-being-dumb, the compiler is perfectly fine :P

### DIFF
--- a/Userland/Shell/Parser.cpp
+++ b/Userland/Shell/Parser.cpp
@@ -1943,10 +1943,14 @@ RefPtr<AST::Node> Parser::parse_heredoc_initiation_record()
 
     // StringLiteral | bareword
     if (auto bareword = parse_bareword()) {
-        if (bareword->is_syntax_error())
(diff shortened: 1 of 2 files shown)


Userland/Shell/Parser.cpp
```diff
@@ -1982,8 +1982,10 @@ RefPtr<AST::Node> Parser::parse_heredoc_initiation_record()
 
 bool Parser::parse_heredoc_entries()
 {
+    auto heredocs = move(m_heredoc_initiations);
+    m_heredoc_initiations.clear();
```
awesomekling, line 1986: Move-constructing the `heredocs` vector from another vector will clear out the source vector. No need to call `clear()` again :)

alimpfard: tbh that's for the reader (i.e. future me) - "no, this function won't leave the vector in a weird state, don't go looking"

PR #6777, opened 2021-05-01
LibC: Implement support for getspnam() and friends

gunnarbeutner:
This implements the functions from `<shadow.h>`. The PR also moves the declarations for `crypt()` and `crypt_r()` to a new header file `<crypt.h>` and adds `LibCrypt` to the include directories (as well as `LibPthread` which I forgot to add earlier).

With this functionality added `sshd` can successfully authenticate users:

Screenshot from 2021-05-01 12-10-58


Userland/Libraries/LibC/shadow.cpp
```diff
@@ -0,0 +1,247 @@
+/*
+ * Copyright (c) 2018-2020, Andreas Kling <kling@serenityos.org>
```
awesomekling, line 2: I don't remember writing this code. :)

gunnarbeutner: Well, technically you did, sort of. It's mostly a copy of `pwd.cpp`. I guess I'll add myself there, then. :)

awesomekling:
Can OpenSSH use `sendfd()` and `recvfd()` instead of `SCM_RIGHTS`?

gunnarbeutner:
Actually, yes, and it was surprisingly easy. OpenSSH already had:
```C
int mm_send_fd(int sock, int fd);
int mm_receive_fd(int sock);
```

awesomekling:
> Actually, yes, and it was surprisingly easy. OpenSSH already had:
> 
> ```c
> int mm_send_fd(int sock, int fd);
> int mm_receive_fd(int sock);
> ```

That's kickass 😄

PR #6779, opened 2021-05-01
HackStudio: Add "document dirty" indicator & warn when exiting/building with unsaved changes

itamar8910:
& Also some small refactoring of how we get the "currently active file".

### DIFF
--- a/Userland/DevTools/HackStudio/HackStudioWidget.cpp
+++ b/Userland/DevTools/HackStudio/HackStudioWidget.cpp
@@ -552,7 +552,7 @@ NonnullRefPtr<GUI::Action> HackStudioWidget::create_save_action()
         if (active_file().is_empty())
             return;
 
-        current_editor().write_to_file(active_file());
+        current_editor_wrapper().save();
 
         if (m_git_widget->initialized())
             m_git_widget->refresh();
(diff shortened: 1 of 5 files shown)


Userland/DevTools/HackStudio/HackStudioWidget.cpp
```diff
@@ -1085,4 +1090,27 @@ HackStudioWidget::~HackStudioWidget()
...
+{
+    if (!any_document_is_dirty())
+        return true;
+
+    auto result = GUI::MessageBox::show(window(), "There are unsaved changes, do you really want to exit?", "Unsaved changes", GUI::MessageBox::Type::Warning, GUI::MessageBox::InputType::YesNoCancel);
+
+    if (result == GUI::MessageBox::ExecYes)
+        return true;
+
+    return false;
```
awesomekling, line 1103: If "Yes" or "Not Yes" are the choices that matter, the message box should be a Yes/No box, instead of a Yes/No/Cancel box.

awesomekling (changes requested):

Userland/DevTools/HackStudio/HackStudioWidget.cpp
IdanHo: "There are unsaved changes, do you want to save before exiting?" with a "Yes" (save and exit), "No" (dont save and exit) and "Cancel" (dont save and dont exit) is how most programs i've used in the past have done it, so replacing this with that might be nice :)

itamar8910: Yes, I totally agree @IdanHo. Please see my changes.

PR #6781, opened 2021-05-01
Uncrosscompiling: Basic build system, step 2

boricj:
_What happens when we don't cross-compile SerenityOS?_

- To build Serenity's userland for the host, use `cmake -B Build/$(cc -dumpmachine) -G Ninja` ;
- To cross-build SerenityOS as usual, use `cmake -B Build/i686 -DCMAKE_SYSTEM_NAME=SerenityOS -G Ninja`

This is a follow-up of #6671, where the idea is to build Serenity for other systems than SerenityOS. Instead of trying to catch up to the upstream while porting the userland, let's upstream the build system and then port the userland.

Why would we want to not cross-build Serenity?
- This makes most of Serenity's userland potentially available for other operating systems than Serenity.
- This exposes more of Serenity's code to foreign systems in order to uncover portability and correctness issues.
- This allows building most of Serenity's userland on 64 bit hosts, fixing 64 bits issues ahead of a x86_64 port.
- This enables more code to be exposed to the CI for fuzzing, testing and so on.
- This effectively eliminates the need for Lagom, which exists in parallel of SerenityOS's build system. There is still a need for building host tools when cross-compiling that could be addressed with a preliminary host build before the cross-compilation build, but this PR doesn't touch Lagom to keep the diff at a reasonable size.

There are several limitations for now:
- ~~RPATH generation is broken for cross-building and is worked around in SerenityOS's dynamic loader. I can't figure out how to fix this properly and I'm starting to think the current build system simply can't deal with this.~~
- Most of the userland doesn't actually build right now for the host system because of various incompatibilities. Use `cmake --build Build/$(cc -dumpmachine) -- -k 0` to force building as much of it.
- Running executables on the host requires setting `LD_LIBRARY_PATH` for the dynamic loader to find the libraries.
- This PR does the bare minimum porting effort to get some utilities built in order to prove this works. An actual porting effort of Serenity's userland is required to really make use of this.

~~This is a draft because I'm not sure this is in an acceptable state for merging (mostly because of the RPATH issue)~~.


gunnarbeutner:
Are you testing this on macOS? I believe that #6745 fixed the rpath problem by properly setting `SONAME`s for the libraries. Can you show me `readelf -a Build/i686/Root/...` for one of the broken libraries?

boricj:
I'm testing this on Linux, I do not have access to a macOS system.

Output of `readelf -d ./Build/i686/Userland/Libraries/LibCore/libcore.so`:
```
Dynamic section at offset 0x52794 contains 28 entries:
  Tag        Type                         Name/Value
 0x00000001 (NEEDED)                     Shared library: [libgcc_s.so]
 0x00000001 (NEEDED)                     Shared library: [Userland/Libraries/LibCrypt/libcrypt.so]
 0x00000001 (NEEDED)                     Shared library: [Userland/Libraries/LibCrypto/libcrypto.so]
 0x00000001 (NEEDED)                     Shared library: [Userland/Libraries/LibC/libc.so]
 0x00000001 (NEEDED)                     Shared library: [libsystem.so]
 0x00000001 (NEEDED)                     Shared library: [libm.so]
 0x00000001 (NEEDED)                     Shared library: [libc.so]
 0x0000000c (INIT)                       0xe9fc
 0x0000000d (FINI)                       0x4bf81
 0x00000019 (INIT_ARRAY)                 0x52d08
 0x0000001b (INIT_ARRAYSZ)               24 (bytes)
 0x6ffffef5 (GNU_HASH)                   0xf4
 0x00000005 (STRTAB)                     0x4b4c
 0x00000006 (SYMTAB)                     0x154c
 0x0000000a (STRSZ)                      31958 (bytes)
 0x0000000b (SYMENT)                     16 (bytes)
 0x00000003 (PLTGOT)                     0x53894
 0x00000002 (PLTRELSZ)                   3144 (bytes)
 0x00000014 (PLTREL)                     REL
 0x00000017 (JMPREL)                     0xddb4
 0x00000011 (REL)                        0xc824
 0x00000012 (RELSZ)                      5520 (bytes)
 0x00000013 (RELENT)                     8 (bytes)
 0x00000016 (TEXTREL)                    0x0
 0x0000001e (FLAGS)                      TEXTREL BIND_NOW STATIC_TLS
 0x6ffffffb (FLAGS_1)                    Flags: NOW
 0x6ffffffa (RELCOUNT)                   182
 0x00000000 (NULL)                       0x0
```

I've rebased the branch on top of master. I still have the RPATH issue, but let's see if the macOS build is fixed.

gunnarbeutner:
Yeah, that's definitely because you're setting `CMAKE_SYSTEM_NAME` and cmake doesn't know how to set `SONAME`s for that target. Try changing line 153 in the main `CMakeLists.txt` to:

```CMake
if (CMAKE_SYSTEM_NAME MATCHES Darwin OR CMAKE_SYSTEM_NAME MATCHES SerenityOS)
```

boricj:
That fixed the RPATH issue, thanks. I've dropped the dynamic loader hack as a result. It seems that the macOS build has also been fixed on master in the meantime.

I guess the one major remaining issue with this PR is the fact that it doesn't touch Lagom nor the tests/fuzzers. The host build by itself should be able to replace everything but the host tools used during the build (and potentially also the host builds too if we separate the current single build into two). However, making a host build that doesn't fail and hooking everything up to it will balloon up the diff of this PR... It's probably best to keep changes incremental over several separate PRs rather than try and do everything at once.

linusg (changes requested):
TBH I'm not a huge fan of making `-DCMAKE_SYSTEM_NAME=SerenityOS` a mandatory option - can we make that a default with the option to override if needed? The primary goal here is to build a new operating system after all and not make as much of it as possible run on existing systems, so building "SerenityOS for SerenityOS" (as opposed to "SerenityOS for a Linux host system" - which makes little sense IMO, maybe you can reword it? it's "parts of SerenityOS", at best) should be the implied default.

Please also make sure to properly indent CMake `if ()` blocks.

CMakeLists.txt
```diff
@@ -112,22 +112,33 @@ set(CMAKE_STAGING_PREFIX ${CMAKE_BINARY_DIR}/Root)
...
-    message("Good job on building cmake!")
+if (${CMAKE_CROSSCOMPILING})
+    set(GCC_VERSION 11.1.0)
+    set(LIBRARY_LIBC LibC)
+    set(LIBRARY_LIBCRYPT LibCrypt)
+    set(LIBRARY_LIBM LibM)
+    set(LIBRARY_LIBPTHREAD LibPthread)
+    set(LIBRARY_LIBSYSTEM LibSystem)
+
+    if (${CMAKE_HOST_SYSTEM_NAME} MATCHES SerenityOS)
```
ADKaster, line 123: When building serenity in Serenity, we won't be cross compiling, right? So this check is in the wrong place.

CMakeLists.txt
```diff
@@ -139,8 +150,8 @@ foreach(lang ASM C CXX OBJC OBJCXX)
     unset(CMAKE_SHARED_MODULE_LOADER_${lang}_FLAG )
     unset(CMAKE_${lang}_OSX_DEPLOYMENT_TARGET_FLAG)
     unset(CMAKE_${lang}_SYSROOT_FLAG)
-    if (CMAKE_SYSTEM_NAME MATCHES Darwin)
-        ## macOS workaround. Use GNU ld flags for SONAMEs.
+    if (CMAKE_SYSTEM_NAME MATCHES Darwin OR CMAKE_SYSTEM_NAME MATCHES SerenityOS)
```
ADKaster, line 153: I think having CMake Toolchain files that we use for the normal build would allow us to remove this hack, and avoid setting CMake system name at the command line.

ADKaster (commented):
I agree with Linus that setting CMake system name at the command line is awkward... I like the goals of this, it would be great to be able to run host dynamic analysis tools on these applications and Libraries while their counterparts in serenity are being developed. (And to fix 32-bit only assumptions where they appear)

boricj:
Well, there are good reasons _not_ to do that:
- Putting `-DCMAKE_SYSTEM_NAME=target` on the command line is the official CMake-supported way to declare a cross-compilation build. Deviating from this makes the build system more atypical and harder to understand.
- Not putting that option on the command line will not set a bunch of extra variables (including but not limited to `CMAKE_CROSSCOMPILING`). The current build system hacks around these issues, but it's more complexity that could be avoided in the first place.
- The early host tools inclusion+toolchain hotswap hack currently used in the build system is rather awkward. A two-step build system separating host tools building and cross-compilation would eliminate this and simplify the build system.

That wasn't the point of this PR initially, but maybe I should try and reorganize the build system to work like this:
```
cmake -B Build/host (-DHOST_TOOLS_ONLY=True)
cmake -B Build/i686 -DCMAKE_SYSTEM_NAME=SerenityOS -DCMAKE_SYSTEM_PROCESSOR=i686
cmake --build Build/host
cmake --build Build/i686
```

If this results in a significant simplification of the build system and makes it less atypical when compared to other projects using CMake, would you be in favor of it?

boricj:
Switching PR back to draft because these changes are going to take a while to perform.

Userland/CMakeLists.txt
```diff
@@ -1,13 +1,16 @@
 add_compile_options(-O2)
 
 add_subdirectory(Applications)
-add_subdirectory(Demos)
```
Hendiadyoin1, line 4: why did this move?

boricj:
So I've tried to make this work, but after several attempts it's just too much work to do in one go.

One of the biggest issues is that not doing `-DCMAKE_SYSTEM_NAME=SerenityOS` creates a ton of problems, as CMake thinks it's building for the host and therefore sets up a configuration suitable for the host. This is currently papered over with a _lot_ of overrides inside SerenityOS's build system to make it fit, some of them specific to a given system (looking at you, macOS). This point is driven further home by the fact that the ports system does set up a proper CMake cross-compilation configuration and does not suffer from these issues. Unfortunately, switching to a two-stage build with a proper cross-compilation requires a lot of changes that I just can't quite get right (and I can't keep these changes up with upstream either).

Therefore, I'm closing this PR because this approach simply won't work with the current state of SerenityOS's build system. I need to come up with another multi-step plan to make this idea feasible.

PR #6782, opened 2021-05-01
Kernel: Prevent execution of /usr/lib/Loader.so dynamic loader

### LINKED ISSUE #5583: System crash/freeze when trying to execute shared object (.so) file
When trying to execute a .so file the system seems to be freezing/crashing. After investigating I realized that .o files are handled accordingly and therefore do not cause a crash.
Example:
`/usr/lib/Loader.so`

**These debug messages appear directly after execution:** 
```
[Shell(30:30)]: ASSERTION FAILED: m_total_range.contains(allocated_range)
[Shell(30:30)]: ../Kernel/VM/RangeAllocator.cpp:165 in AK::Optional<Kernel::Range> Kernel::RangeAllocator::allocate_specific(VirtualAddress, size_t)
[#0 Shell(30:30)]: 0xc0123723  __assertion_failed(char const*, char const*, unsigned int, char const*) +0x145
[#0 Shell(30:30)]: 0xc04c3358  Kernel::RangeAllocator::allocate_specific(VirtualAddress, unsigned long) +0xf6
[#0 Shell(30:30)]: 0xc04daa6b  Kernel::Space::allocate_range(VirtualAddress, unsigned long, unsigned long) +0x1e7
[#0 Shell(30:30)]: 0xc0384075  Kernel::load_elf_object(AK::NonnullOwnPtr<Kernel::Space>, Kernel::FileDescription&, unsigned int, Kernel::ShouldAllocateTls) +0x17e6
[#0 Shell(30:30)]: 0xc0385dd8  Kernel::Process::load(AK::NonnullRefPtr<Kernel::FileDescription>, AK::RefPtr<Kernel::FileDescription, AK::RefPtrTraits<Kernel::FileDescription> >, elfhdr const&) +0x2a6
[#0 Shell(30:30)]: 0xc0387c30  Kernel::Process::do_exec(AK::NonnullRefPtr<Kernel::FileDescription>, AK::Vector<AK::String, 0ul>, AK::Vector<AK::String, 0ul>, AK::RefPtr<Kernel::FileDescription, AK::RefPtrTraits<Kernel::FileDescription> >, Kernel::Thread*&, unsigned int&, elfhdr const&) +0x134c
[#0 Shell(30:30)]: 0xc0390cc6  Kernel::Process::exec(AK::String, AK::Vector<AK::String, 0ul>, AK::Vector<AK::String, 0ul>, int) [clone .part.0] +0x1c8e
[#0 Shell(30:30)]: 0xc0392aa4  Kernel::Process::sys$execve(AK::Userspace<Kernel::Syscall::SC_execve_params const*>) +0xa6e
[#0 Shell(30:30)]: 0xc036de9b  syscall_handler +0x1f08
[#0 Shell(30:30)]: 0xc036bf5b  syscall_asm_entry +0x31
```
That's all i got. I hope this helps :D

xeriab:
This little PR intends to fix system hang when try to execute `/usr/lib/Loader.so` and other `.so` files from ports.
Fixes #5583.
I made it here again after I fixes my mess in my forks of `SerenityOS`.
I'm sorry for the annoyance I made in the old PRs.

### DIFF
--- a/Kernel/Syscalls/execve.cpp
+++ b/Kernel/Syscalls/execve.cpp
@@ -781,9 +781,12 @@ KResultOr<RefPtr<FileDescription>> Process::find_elf_interpreter_for_executable(
     if (main_program_header.e_type == ET_DYN) {
         // If it's ET_DYN with no PT_INTERP, then it's a dynamic executable responsible
         // for its own relocation (i.e. it's /usr/lib/Loader.so)
-        if (path != "/usr/lib/Loader.so")
-            dbgln("exec({}): WARNING - Dynamic ELF executable without a PT_INTERP header, and isn't /usr/lib/Loader.so", path);
-        return nullptr;
+        if (path != "/usr/lib/Loader.so") {
+            dmesgln("exec({}): WARNING - Dynamic ELF executable without a PT_INTERP header, and isn't /usr/lib/Loader.so", path);
+        } else {
+            dbgln("exec(={}): WARNING - Hold on, you cannot execute the loader!", path);
+        }
+        return ENOEXEC;
     }
 
     // No interpreter, but, path refers to a valid elf image


awesomekling (changes requested):
This change is a hack and is not sufficient -- it will still be possible to crash the system by doing something like:
```console
$ cp /usr/lib/Loader.so foo
$ ./foo
```

We need to look at what's causing the kernel problem and fix the real issue :)

xeriab:
> This change is a hack and is not sufficient -- it will still be possible to crash the system by doing something like:
> 
> ```
> $ cp /usr/lib/Loader.so foo
> $ ./foo
> ```
> 
> We need to look at what's causing the kernel problem and fix the real issue :)

Oops, sorry; never thought about the other way to cause the hang.
Thanks for clarification =)

PR #6783, opened 2021-05-01
Base: Add bold Hebrew character glyphs to the Katica Bold 10 font

GalHorowitz:
The glyph for the letter `Shin` (U+05E9) looks a bit wonky because the width of the font is too small. Unfortunately it doesn't look like the FontEditor is capable of changing the width of an existing font, so if that option will be added in the future this glyph can be fixed.
The google.co.il webpage looks even saner, because bold Hebrew text is no longer missing.


IdanHo (approved):
LGTM
Before:
image
After:
image

PR #6788, opened 2021-05-01
Kernel: Handle both shift keys being pressed and then released

IdanHo:
Our current implementation does not work in the special case in which both shift keys are pressed, and then only one of the keys is released, as this would result in writing lower case letters, instead of the expected upper case letters.

This commit fixes that by keeping track of the amount of shift keys that are pressed (instead of if any are at all), and only switching to the unshifted keymap once all of them are released.

### DIFF
--- a/Kernel/Devices/HID/KeyboardDevice.h
+++ b/Kernel/Devices/HID/KeyboardDevice.h
@@ -58,6 +58,7 @@ protected:
     bool m_caps_lock_on { false };
     bool m_num_lock_on { false };
     bool m_has_e0_prefix { false };
+    bool m_both_shift_keys_pressed { false };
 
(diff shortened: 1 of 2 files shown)


awesomekling (changes requested):
I like the way you described it in the commit message better than the actual implementation.

If we're not gonna keep an actual count of shift keys pressed, let's at least improve the variable name a bit: how about `m_both_shift_keys_pressed`? :)

IdanHo:
> I like the way you described it in the commit message better than the actual implementation.

That was the way i originally intended to implement it, but it felt like a count variable that can only have 3 states (0/1/2) is a bit wasteful, fixed the naming though :)

PR #6790, opened 2021-05-01
Userland: Make rev(1) read from stdin if no file supplied and add manpage clarification.

Mango0x45:
This should bring the behaviour of rev inline with the BSD versions and other tools such as cat and grep. Also I clarified trailing newlines in the rev manpage.

### DIFF
--- a/Base/usr/share/man/man1/rev.md
+++ b/Base/usr/share/man/man1/rev.md
@@ -11,8 +11,9 @@ $ rev [files...]
 ## Description
 
 `rev` reads the specified files line by line, and prints them to standard
-output with each line being reversed characterwise. If no files are specifed,
-then `rev` will read from standard input.
+output with each line being reversed characterwise with the exception of
+trailing newline characters which will always print at the end of the line. If
+no files are specifed, then `rev` will read from standard input.
 
 ## Examples
 
--- a/Userland/Utilities/rev.cpp
+++ b/Userland/Utilities/rev.cpp
@@ -4,10 +4,12 @@
  * SPDX-License-Identifier: BSD-2-Clause
  */
 
+#include <AK/ScopeGuard.h>
 #include <AK/String.h>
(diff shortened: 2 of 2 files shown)


Userland/Utilities/rev.cpp
```diff
@@ -25,29 +27,43 @@ int main(int argc, char** argv)
     args_parser.parse(argc, argv);
 
     Vector<RefPtr<Core::File>> files;
-    if (paths.is_empty()) {
-        files.append(Core::File::standard_input());
```
awesomekling, line 29: Your commit message says
>With this commit, if no files are supplied to rev, it will by default read user input from stdin.

But that was already accomplished by the code you are removing here, no?

awesomekling: Or oh, I see. The old code doesn't work correctly.
Well, we should make it work correctly instead of making separate logic for stdin vs any other file. :)

Mango0x45: Yeah, it seems that appending `Core::File::standard_input()` causes it to terminate immediately if you don't supply any files. It also causes it to *sometimes* work when piping in input (when I originally tested it I got lucky it seems), however further testing it, it seems that more often then not it won't work and will terminate instantly.

Base/usr/share/man/man1/rev.md
```diff
@@ -11,8 +11,9 @@ $ rev [files...]
 ## Description
 
 `rev` reads the specified files line by line, and prints them to standard
-output with each line being reversed characterwise. If no files are specifed,
-then `rev` will read from standard input.
+output with each line being reversed characterwise with the exception of
+trailing newline characters which will always print at the end of the line. If
+no files are specifed, then `rev` will read from standard input.
```
bcoles, line 16:
```suggestion
no files are specified, then `rev` will read from standard input.
```

PR #6793, opened 2021-05-01
Kernel: Make sure we read all packets

gunnarbeutner:
The previous patch (#6669) already helped with this, however my idea of only reading a few packets didn't work and we'd still sometimes end up not receiving any more packets from the E1000 interface.

With this patch applied my NIC seems to receive packets just fine, at least for now.


awesomekling:
Next patch: "make doubly sure that we read all packets" :smile:

gunnarbeutner:
I wouldn't be surprised!

PR #6794, opened 2021-05-01
Userland: Improve how traceroute reports timing data for packets

gunnarbeutner:
Previously we'd include the time required to do DNS resolution in the time we'd report for traceroute packets. This would skew the times reported.


IdanHo (approved):
LGTM

PR #6801, opened 2021-05-01
HackStudio: Use common copy functions in ProjectTemplate.cpp

vkoskiv:
Yet another FIXME patch. I verified it works by creating a few template projects, seems to work just as it did before.

### DIFF
--- a/Userland/DevTools/HackStudio/ProjectTemplate.cpp
+++ b/Userland/DevTools/HackStudio/ProjectTemplate.cpp
@@ -19,12 +19,6 @@
 #include <sys/wait.h>
 #include <unistd.h>
 
-// FIXME: shameless copy+paste from Userland/cp. We should have system-wide file management functions.
-// Issue #5209
-bool copy_file_or_directory(String, String, bool, bool);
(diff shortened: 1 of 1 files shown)


Userland/DevTools/HackStudio/ProjectTemplate.cpp
```diff
@@ -76,8 +70,10 @@ Result<void, String> ProjectTemplate::create_project(const String& name, const S
     // Verify that the template content directory exists. If it does, copy it's contents.
     // Otherwise, create an empty directory at the project path.
     if (Core::File::is_directory(content_path())) {
-        if (!copy_directory(content_path(), path, false))
-            return String("Failed to copy template contents.");
+        auto result = Core::File::copy_file_or_directory(path, content_path());
```
linusg, line 73: not sure I agree with the choice to swap src and dest here, but that's not your fault of course :smile:

vkoskiv: Yeah, threw me off for a bit as well.

linusg:
✨

PR #6802, opened 2021-05-01
Kernel: Change Inode::{read/write}_bytes interface to return KResulOr<ssize_t>

bgianfo:
The error handling in all these cases was still using the old style
negative values to indicate errors. We have a nicer solution for this
now with KResultOr<T>. This change switches the interface and then all
implementers to use the new style.

### DIFF
--- a/Kernel/FileSystem/VirtualFileSystem.cpp
+++ b/Kernel/FileSystem/VirtualFileSystem.cpp
@@ -710,9 +710,9 @@ KResult VFS::symlink(StringView target, StringView linkpath, Custody& base)
         return inode_or_error.error();
     auto& inode = inode_or_error.value();
     auto target_buffer = UserOrKernelBuffer::for_kernel_buffer(const_cast<u8*>((const u8*)target.characters_without_null_termination()));
-    ssize_t nwritten = inode->write_bytes(0, target.length(), target_buffer, nullptr);
-    if (nwritten < 0)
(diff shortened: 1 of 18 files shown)


Kernel/FileSystem/VirtualFileSystem.cpp
```diff
@@ -710,10 +710,8 @@ KResult VFS::symlink(StringView target, StringView linkpath, Custody& base)
         return inode_or_error.error();
     auto& inode = inode_or_error.value();
     auto target_buffer = UserOrKernelBuffer::for_kernel_buffer(const_cast<u8*>((const u8*)target.characters_without_null_termination()));
-    ssize_t nwritten = inode->write_bytes(0, target.length(), target_buffer, nullptr);
-    if (nwritten < 0)
-        return KResult((ErrnoCode)-nwritten);
-    return KSuccess;
+    auto result = inode->write_bytes(0, target.length(), target_buffer, nullptr);
+    return result.result();
```
awesomekling, line 714: This looks a little iffy. If `write_bytes` succeeds, the returned `KResultOr<ssize_t>` from there would be in the "Or" state, rather than the "KResult" state. So calling `result()` feels wrong..

bgianfo: Yea, that's fair. Fixed.

PR #6803, opened 2021-05-02
Kernel+Userland: Fix 64-bit portability issues

gunnarbeutner:
This lets us build most of the userland for x86_64. There are still some build problems (around 100 files fail to compile) but those are now mostly related to 32-bit registers (e.g. in `UserspaceEmulator`, `LibDebug`, etc.).

This PR also adds _some_ support for doing x86_64 ELF relocations. This is untested and there are definitely some relocation types we're missing.

I've added `-fPIC` to the GCC specfile so that all code is built with it, including static libraries. Ordinarily this would force the compiler to not inline certain symbols and call them via the PLT instead. To counteract this I've also added `-fno-semantic-interposition` which disables ELF symbol interposition. Our dynamic loader doesn't support this anyway and we might even consider not implementing this at all.

This does not require a toolchain rebuild unless you're actively working on getting x86_64 up and running.

Adding `-fno-semantic-interposition` to the GCC command line caused a new warning in `AK/Base64.cpp`. I don't see how output.data() could be uninitialized here. Also, commenting out the `ensure_capacity()` call for the Vector also gets rid of this warning.

### DIFF
--- a/Userland/Games/Chess/ChessWidget.cpp
+++ b/Userland/Games/Chess/ChessWidget.cpp
@@ -360,13 +360,13 @@ void ChessWidget::set_piece_set(const StringView& set)
 
 Chess::Square ChessWidget::mouse_to_square(GUI::MouseEvent& event) const
 {
-    size_t tile_width = width() / 8;
-    size_t tile_height = height() / 8;
+    unsigned tile_width = width() / 8;
+    unsigned tile_height = height() / 8;
 
     if (side() == Chess::Color::White) {
-        return { 7 - (event.y() / tile_height), event.x() / tile_width };
+        return { (unsigned)(7 - (event.y() / tile_height)), (unsigned)(event.x() / tile_width) };
     } else {
-        return { event.y() / tile_height, 7 - (event.x() / tile_width) };
+        return { (unsigned)(event.y() / tile_height), (unsigned)(7 - (event.x() / tile_width)) };
     }
 }
 
--- a/Userland/Libraries/LibELF/Arch/x86_64/plt_trampoline.S
+++ b/Userland/Libraries/LibELF/Arch/x86_64/plt_trampoline.S
@@ -1,0 +1,12 @@
+/*
+ * Copyright (c) 2021, Gunnar Beutner <gbeutner@serenityos.org>
+ *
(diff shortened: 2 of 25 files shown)


Userland/Libraries/LibELF/Arch/x86_64/plt_trampoline.S
```diff
@@ -0,0 +1,80 @@
...
+ * 2. Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in the
+ *    documentation and/or other materials provided with the distribution.
+ * 3. All advertising materials mentioning features or use of this software
+ *    must display the following acknowledgement:
+ *      This product includes software developed for the NetBSD Project by
+ *      Wasabi Systems, Inc.
+ * 4. The name of Wasabi Systems, Inc. may not be used to endorse
+ *    or promote products derived from this software without specific prior
+ *    written permission.
```
awesomekling, line 21: Let's not bring in 4-clause BSD code, the 3rd clause is so annoying :^(

Userland/Games/Chess/ChessWidget.cpp
```diff
@@ -360,13 +360,13 @@ void ChessWidget::set_piece_set(const StringView& set)
 
 Chess::Square ChessWidget::mouse_to_square(GUI::MouseEvent& event) const
 {
-    size_t tile_width = width() / 8;
-    size_t tile_height = height() / 8;
+    int tile_width = width() / 8;
+    int tile_height = height() / 8;
```
Hendiadyoin1, line 364: can this be a `u32` instead? I don't see why this needs to be signed we cast it to `unsigned` later anyway.

gunnarbeutner: Aye, makes sense. I've changed it to an `unsigned`. :+1:

Userland/Libraries/LibELF/Arch/x86_64/plt_trampoline.S
gunnarbeutner: Alrighty, replaced with an `int3` until I can either come up with something on my own or find something more a more reasonable license.

PR #6805, opened 2021-05-02
Userland: Clean up return types for IPC handlers

gunnarbeutner:
This removes the `OwnPtr<...>` for IPC handlers. It also turns the return type into `void` if the IPC method doesn't have any return arguments. The methods are still synchronous in that case.

I wonder if we should turn this:

`virtual Messages::AudioServer::GetRemainingSamplesResponse ClientConnection::handle(const Messages::AudioServer::GetRemainingSamples&);`

into this:

`virtual int GetRemainingSamples();`

The return type would be:
* `void` for IPC methods without any outputs.
* Single type for IPC methods with one output.
* The `XXXResponse` class for IPC methods with more than one output.


awesomekling:
I think always returning the message feels a bit more consistent. This is very cool btw, thanks for doing it! :)

Userland/Libraries/LibIPC/Connection.h
```diff
@@ -48,13 +48,18 @@ class Connection : public Core::Object {
     }
 
     void post_message(const Message& message)
+    {
+        post_message(message.encode());
+    }
+
+    // FIXME: unnecessary copy
+    void post_message(MessageBuffer buffer)
```
alimpfard, line 56: Why not take a reference then?
what's being copied here exactly anyway (and why is this public to begin with?)

PR #6806, opened 2021-05-02
Shell: Update shebang handling logic

### LINKED ISSUE #6774: Shell: tries to execute programs in current working directory if they do not exist in `$PATH`
image

```
Shell(31:31): execvpe() leaving :(
Shell(31:31): execvp(asdf, ...) about to return -1 with errno=2
Shell(31:31): execvp(#!/bin/sh, ...) about to return -1 with errno=2
Shell(22:22): Job entry 'asdf' deleted in 10 ms
```

alimpfard:
This bit of code was kept unmodified since it was first implemented,
and I'm not entirely convinced that it ever actually worked :P
This commit updates the code to use "modern" classes and constructs,
and fixes an issue where the shebang would still contain the '#!'
when it was passed to execvp().
Fixes #6774.

### DIFF
--- a/Userland/Shell/Shell.cpp
+++ b/Userland/Shell/Shell.cpp
@@ -511,12 +511,14 @@ String Shell::resolve_alias(const String& name) const
 
 bool Shell::is_runnable(const StringView& name)
 {
-    if (access(name.to_string().characters(), X_OK) == 0)
+    auto parts = name.split_view('/');
+    auto path = name.to_string();
+    if (parts.size() > 1 && access(path.characters(), X_OK) == 0)
         return true;
 
     return binary_search(
         cached_path.span(),
-        name.to_string(),
+        path,
         nullptr,
         [](auto& name, auto& program) { return strcmp(name.characters(), program.characters()); });
 }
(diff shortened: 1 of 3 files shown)


bcoles:
Something is not quite right after this patch.

image

It is extremely rare for an interpreter path to have no slashes after the initial `/`, but I think this is still a valid use case )?) At the very least it shouldn't chop off half of the interpreter path.

alimpfard:
Well that definitely doesn't look right :P

bcoles:
That appears to have resolved the issue.

image

Userland/Shell/Shell.cpp
```diff
@@ -856,39 +858,50 @@ void Shell::execute_process(Vector<const char*>&& argv)
...
                 int rc = execvp(argv[0], const_cast<char* const*>(argv.data()));
-                if (rc < 0)
-                    fprintf(stderr, "%s: Invalid interpreter \"%s\": %s\n", argv[0], &shebang[2], strerror(errno));
-            } else
-                fprintf(stderr, "%s: Command not found.\n", argv[0]);
+                if (rc < 0) {
+                    warnln("{}: Invalid interpreter \"{}\": {}", argv[0], shebang.characters(), strerror(errno));
+                    _exit(126);
+                }
+            } while (false);
```
bcoles, line 897: Not sure I understand the purpose of this?

bcoles (approved):
These changes seem reasonable to me.

The changes in `Shell.cpp` seem a little flimsy, but parsing the shebang line is not as simple as it appears at first glance and I have nothing constructive to add.

Confirmed this PR patches the original issue #6774.

I wasn't sure how I feel about executing a program from within a sub-directory of the working directory without prefacing the command with `./`.

image

But this is consistent with other systems (Linux).

```
user@ubuntu:/tmp$ mkdir asdf
user@ubuntu:/tmp$ cp /bin/id /tmp/asdf
user@ubuntu:/tmp$ asdf/id
uid=1000(user) gid=1000(user) groups=1000(user),4(adm),24(cdrom),27(sudo),30(dip),46(plugdev),120(lpadmin),131(lxd),132(sambashare),1002(wireshark)
user@ubuntu:/tmp$ 
```

There's no immediately obvious security concerns. This is the best I could come up with:

image

PR #6809, opened 2021-05-02
LibCore: Implement a way to parse a DateTime from a string

X-yl:
I implemented the same format specifiers as `strptime`, since `to_string()` was already using the same formatters as `strftime`. The formatters I haven't implemented include the ones that use locale specific formats, the alternate number system ones and the glibc extensions — with the exception of `%z`, which parses a ISO 8601 time zone offset.

Let me know if I should make any changes :^)


awesomekling:
While this is neat, it's not used anywhere so it's just dead code.
Please include this with a feature or bugfix that depends on it if you have one. :)

PR #6815, opened 2021-05-02
Meta: Add a post-commit commit message linter hook

IdanHo:
This should help with getting commit messages tidy before they pass through CI's commit linter :^)

For this hook to work pre-commit has to be explicitly installed via:
`pre-commit install --hook-type commit-msg`

cc @bgianfo


bgianfo:
Thank You! :pray:

PR #6818, opened 2021-05-02
TextEditor: Clear the selection before deleting it

Pangoraw:
This patches fixes a crash of the Userland/TextEditor where it would
crash when deleting a range spanning two lines. This was because the
TextEditor would delete the range and modify the cursor position
before clearing the selection. This would trigger a status bar update
with the invalid selection.


awesomekling:
Oh interesting. Okay.

PR #6819, opened 2021-05-02
Fix broken ProcFS (and sysctl as a by-product)

SpencerCDixon:
In `ProcFS`, not all sub dirs are backed by an actual process. We were incorrectly returning a Process which was returning early and making it so other utilities like `sysctl` broke. A high level integration test to see the effects of this PR are:

**Boot serenity master**

```
$ su
$ sysctl -a
```

This will return a permission error:

Screenshot from 2021-05-02 11-15-09

**Boot serenity with this PR**

```
$ su
$ sysctl -a
```

The correct values:

Screenshot from 2021-05-02 13-02-23


awesomekling:
Great! I didn't even realize this was broken :)

PR #6822, opened 2021-05-02
FileManager: Added basic Trash support

xeriab:
Hello friends!

I've added basic Trash support to SerenityOS based on [Free Desktop](https://specifications.freedesktop.org/trash-spec/trashspec-1.0.html).

It adds a new context menu item to move the selected file to Trash bin.

It lacks some features like Restore or Browsing the the Trash bin for the moment but if this addition is welcome I will complete the whole thing and continue improving.

### DIFF
--- a/Userland/Applications/FileManager/FileUtils.cpp
+++ b/Userland/Applications/FileManager/FileUtils.cpp
@@ -6,8 +6,11 @@
 
 #include "FileUtils.h"
 #include <AK/LexicalPath.h>
+#include <LibCore/ConfigFile.h>
+#include <LibCore/DateTime.h>
 #include <LibCore/DirIterator.h>
 #include <LibCore/File.h>
+#include <LibCore/StandardPaths.h>
 #include <LibGUI/MessageBox.h>
 #include <sys/stat.h>
 #include <unistd.h>
@@ -65,6 +68,105 @@ void delete_paths(const Vector<String>& paths, bool should_confirm, GUI::Window*
     for (auto& path : paths) {
         delete_path(path, parent_window);
     }
+}
+
+void create_trash_bin()
+{
+    auto trash_bin_path = Core::StandardPaths::trash_directory();
+    auto trash_bin_files = String::formatted("{}/files", trash_bin_path);
+    auto trash_bin_info = String::formatted("{}/info", trash_bin_path);
+
+    if (!Core::File::is_directory(trash_bin_path)) {
+        if (mkdir(trash_bin_path.characters(), 0755) < 0) {
+            auto saved_errno = errno;
+            warnln("Failed to create directory {}: {}", trash_bin_path, strerror(saved_errno));
+        }
+    }
+
+    if (!Core::File::is_directory(trash_bin_files)) {
+        if (mkdir(trash_bin_files.characters(), 0755) < 0) {
+            auto saved_errno = errno;
+            warnln("Failed to create directory {}: {}", trash_bin_files, strerror(saved_errno));
+        }
+    }
+
+    if (!Core::File::is_directory(trash_bin_info)) {
+        if (mkdir(trash_bin_info.characters(), 0755) < 0) {
+            auto saved_errno = errno;
+            warnln("Failed to create directory {}: {}", trash_bin_info, strerror(saved_errno));
+        }
+    }
+}
+
+[[maybe_unused]] bool move_to_trash(String const& path)
+{
+    auto file_name = LexicalPath(path).basename();
+    auto file_info_name = String::formatted("{}.trashinfo", file_name);
+    auto deletion_date = Core::DateTime::now().to_string("%Y-%m-%dT%H:%M:%S");
+    auto trash_bin_files = String::formatted("{}/files", Core::StandardPaths::trash_directory());
+    auto trash_bin_info = String::formatted("{}/info", Core::StandardPaths::trash_directory());
+
+    create_trash_bin();
+
+    auto trash_info_path = String::formatted("{}/{}", trash_bin_info, file_info_name);
+
+    auto file_or_error = Core::File::open(trash_info_path, Core::IODevice::WriteOnly);
+
+    if (file_or_error.is_error()) {
+        warnln("Cannot write trash info for file '{}'", path);
+        return false;
+    }
+
+    auto file = file_or_error.release_value();
+    file->close();
+
+    auto trash_info = Core::ConfigFile::open(trash_info_path);
+
+    trash_info->write_entry("Trash Info", "Path", path);
+    trash_info->write_entry("Trash Info", "DeletionDate", deletion_date);
+    trash_info->sync();
+
+    auto trash_file_path = String::formatted("{}/{}", trash_bin_files, file_name);
+
+    auto rc = rename(path.characters(), trash_file_path.characters());
+
+    if (rc < 0) {
+        if (errno == EXDEV) {
+            auto result = Core::File::copy_file_or_directory(
+                trash_file_path,
+                path,
+                Core::File::RecursionMode::Allowed,
+                Core::File::LinkMode::Disallowed,
+                Core::File::AddDuplicateFileMarker::No);
+
+            if (result.is_error()) {
+                dbgln("Could not move '{}' to trash: {}", path, result.error().error_code);
+                return false;
+            }
+
+            rc = unlink(path.characters());
+
+            if (rc < 0)
+                dbgln("Could not unlink '{}': {}", path, strerror(errno));
+        }
+    } else {
+        dbgln("Could not rename '{}': {}", path, strerror(errno));
+        return false;
+    }
+
+    return true;
+}
+
+// FIXME: I'm not sure this is a good practice
+[[maybe_unused]] bool move_to_trash(Vector<String> const& paths)
+{
+    auto rv = false;
+
+    for (auto& path : paths) {
+        rv = move_to_trash(path);
+    }
+
+    return rv;
 }
 
 }
--- a/Userland/Applications/FileManager/FileUtils.h
+++ b/Userland/Applications/FileManager/FileUtils.h
@@ -20,4 +20,8 @@ enum class FileOperation {
 
 void delete_path(const String&, GUI::Window*);
 void delete_paths(const Vector<String>&, bool should_confirm, GUI::Window*);
+void create_trash_bin();
+[[maybe_unused]] bool move_to_trash(String const&);
+[[maybe_unused]] bool move_to_trash(Vector<String> const&);
+
 }
--- a/Userland/Applications/FileManager/main.cpp
+++ b/Userland/Applications/FileManager/main.cpp
@@ -374,6 +374,7 @@ int run_in_desktop_mode([[maybe_unused]] RefPtr<Core::ConfigFile> config)
     desktop_context_menu->add_action(cut_action);
     desktop_context_menu->add_action(paste_action);
     desktop_context_menu->add_action(directory_view.delete_action());
+    desktop_context_menu->add_action(directory_view.trash_action());
     desktop_context_menu->add_separator();
     desktop_context_menu->add_action(properties_action);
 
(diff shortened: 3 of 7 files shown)


linusg:
There are various linting issues, please fix them.

xeriab:
> There are various linting issues, please fix them.

I made the needed changes; The linter somehow stopped working in my machine and now I fixed the problem and squashed the commits :)

linusg:
> squashed the commits

forgot to push? it's still two on GitHub.

xeriab:
> > squashed the commits
> 
> forgot to push? it's still two on GitHub.

I guess it's now fixed :)

xeriab:
@linusg Everything is fixed now and ready for review!

Userland/Applications/FileManager/FileUtils.cpp
```diff
@@ -67,4 +68,91 @@ void delete_paths(const Vector<String>& paths, bool should_confirm, GUI::Window*
     }
 }
 
+void create_trash_bin()
+{
+    auto trash_bin_path = String::formatted("{}/.trash", getenv("HOME"));
```
bcoles, line 73: This will attempt to save in `/.trash` if the user's home directory is not set in the `$HOME` environment variable.

Please use `StandardPaths::home_directory()` instead:

https://github.com/SerenityOS/serenity/blob/b25eba10d2f60c54bcd3516183050b36bfa9f390/Userland/Libraries/LibCore/StandardPaths.cpp#L17-L26

Alternatively, it may be useful to add a `trash_directory` function to `LibCore::StandardPaths`.

Userland/Applications/FileManager/FileUtils.cpp
```diff
@@ -67,4 +68,91 @@ void delete_paths(const Vector<String>& paths, bool should_confirm, GUI::Window*
...
+    auto trash_bin_files = String::formatted("{}/files", trash_bin_path);
+    auto trash_bin_info = String::formatted("{}/info", trash_bin_path);
+
+    if (Core::File::is_directory(trash_bin_path)) {
+        dbgln("Trash directory exists.");
```
bcoles, line 78: Is this debug output useful?

Userland/Applications/FileManager/FileUtils.cpp
```diff
@@ -67,4 +68,91 @@ void delete_paths(const Vector<String>& paths, bool should_confirm, GUI::Window*
...
+
+    if (!Core::File::is_directory(trash_bin_info)) {
+        if (mkdir(trash_bin_info.characters(), 0755) < 0) {
+            auto saved_errno = errno;
+            warnln("Failed to create directory {}: {}", trash_bin_info, strerror(saved_errno));
+        }
+    }
+}
+
+[[maybe_unused]] void move_to_trash(String const& path)
```
bcoles, line 101: This function has multiple instances where failure is possible.

Would a return `Boolean` be useful to know whether this function failed?

Userland/Applications/FileManager/FileUtils.cpp
```diff
@@ -67,4 +68,91 @@ void delete_paths(const Vector<String>& paths, bool should_confirm, GUI::Window*
...
+{
+    [[maybe_unused]] auto file_info = "[Trash Info]\nPath={}\nDeletionDate={}\n";
+
+    auto file_name = LexicalPath(path).basename();
+    auto file_info_name = String::formatted("{}.trashinfo", file_name);
+    auto deletion_date = Core::DateTime::now().to_string("%Y-%m-%dT%H:%M:%S");
+
+    auto trash_bin_path = String::formatted("{}/.trash", getenv("HOME"));
```
bcoles, line 109: This will attempt to save in `/.trash` if the user's home directory is not set in the `$HOME` environment variable.

Please use `StandardPaths::home_directory()` instead:

https://github.com/SerenityOS/serenity/blob/b25eba10d2f60c54bcd3516183050b36bfa9f390/Userland/Libraries/LibCore/StandardPaths.cpp#L17-L26

Alternatively, it may be useful to add a `trash_directory` function to `LibCore::StandardPaths`.

Userland/Applications/FileManager/FileUtils.cpp
```diff
@@ -67,4 +68,91 @@ void delete_paths(const Vector<String>& paths, bool should_confirm, GUI::Window*
...
+
+    create_trash_bin();
+
+    auto trash_into_path = String::formatted("{}/{}", trash_bin_info, file_info_name);
+
+    auto trashed_file_info = String::formatted(file_info, path, deletion_date);
+
+    auto file_or_error = Core::File::open(trash_into_path, Core::IODevice::WriteOnly);
+    if (file_or_error.is_error())
+        return;
```
bcoles, line 121: This function prints debug output upon failure in most (all?) other instances. Why not here too?

Userland/Applications/FileManager/FileUtils.cpp
```diff
@@ -67,4 +68,91 @@ void delete_paths(const Vector<String>& paths, bool should_confirm, GUI::Window*
...
+            if (result.is_error()) {
+                dbgln("Could not move '{}' to trash: {}", path, result.error().error_code);
+                return;
+            } else {
+                rc = unlink(path.characters());
+                if (rc < 0)
+                    dbgln("Could not unlink '{}': {}", path, strerror(errno));
+            }
+        }
+    }
```
bcoles, line 148: These nested conditionals contain `else` after `return`.

Also, what happens if `rc` *is* set, but `errono` is *not* `EXDEV` ? Is that possible?

Also, the outer block has no `else`, so you could return early before performing a bunch of unnecessary nested error checking.

```suggestion
    if (rc)
        return;

    if (errno == EXDEV) {
        auto result = Core::File::copy_file_or_directory(
            trash_file_path,
            path,
            Core::File::RecursionMode::Allowed,
            Core::File::LinkMode::Allowed,
            Core::File::AddDuplicateFileMarker::No);

        if (result.is_error()) {
            dbgln("Could not move '{}' to trash: {}", path, result.error().error_code);
            return;
        }

        rc = unlink(path.characters());
        if (rc < 0)
            dbgln("Could not unlink '{}': {}", path, strerror(errno));
    } else {
        // rename failed, and errno is not EXDEV
        // ...
    }
```

Userland/Applications/FileManager/FileUtils.cpp
xeriab: I agree with you, I will do as suggested :)

Userland/Applications/FileManager/FileUtils.cpp
xeriab: I just made the debug to inform myself while working with it :) it's now that useful.

Userland/Applications/FileManager/FileUtils.cpp
xeriab: Silly me I never touched the `StandardPaths` file ever!
That's a good point, Thanks for the review.

xeriab:
@bcoles I made some changes and I think it's good for now :)

Userland/Applications/FileManager/FileUtils.cpp
```diff
@@ -67,4 +69,104 @@ void delete_paths(const Vector<String>& paths, bool should_confirm, GUI::Window*
...
+        if (mkdir(trash_bin_info.characters(), 0755) < 0) {
+            auto saved_errno = errno;
+            warnln("Failed to create directory {}: {}", trash_bin_info, strerror(saved_errno));
+        }
+    }
+}
+
+[[maybe_unused]] bool move_to_trash(String const& path)
+{
+    [[maybe_unused]] auto file_info = "[Trash Info]\nPath={}\nDeletionDate={}\n";
```
bcoles, line 102: This format string is declared here with `[[maybe_unused]]`. It is only ever used once. There's no reason to define it as a separate variable.

More importantly, the INI library should be used for writing INI data, instead of using format strings.

xeriab: Hmm do you mean something like this:

```C++
auto file = file_or_error.release_value();
    file->write("");
    file->close();

    auto trash_info = Core::ConfigFile::open(trash_info_path);

    trash_info->write_entry("Trash Info", "Path", path);
    trash_info->write_entry("Trash Info", "DeletionDate", deletion_date);
    trash_info->sync();
```

bcoles: > Hmm do you mean something like this:

Something like that, yes. Although I'm not sure why the `file->write()` would be required.

xeriab:
@linusg I made all necessary changes for my code with the kind help of @bcoles, If it's ready I hope you review it and merge it if the PR is needed.

Thank you so much guys :)

Userland/Applications/FileManager/FileUtils.cpp
```diff
@@ -67,4 +70,103 @@ void delete_paths(const Vector<String>& paths, bool should_confirm, GUI::Window*
...
+
+    if (!Core::File::is_directory(trash_bin_info)) {
+        if (mkdir(trash_bin_info.characters(), 0755) < 0) {
+            auto saved_errno = errno;
+            warnln("Failed to create directory {}: {}", trash_bin_info, strerror(saved_errno));
+        }
+    }
+}
+
+[[maybe_unused]] bool move_to_trash(String const& path)
```
linusg, line 101: Why is this marked as `[[maybe_unused]]`?

Userland/Applications/FileManager/FileUtils.cpp
```diff
@@ -67,4 +70,103 @@ void delete_paths(const Vector<String>& paths, bool should_confirm, GUI::Window*
...
+    } else {
+        dbgln("Could not rename '{}': {}", path, strerror(errno));
+        return false;
+    }
+
+    return true;
+}
+
+// FIXME: I'm not sure this is a good practice
+[[maybe_unused]] bool move_to_trash(Vector<String> const& paths)
```
linusg, line 161: Why is this marked as `[[maybe_unused]]`?

Userland/Applications/FileManager/FileUtils.cpp
```diff
@@ -67,4 +70,103 @@ void delete_paths(const Vector<String>& paths, bool should_confirm, GUI::Window*
...
+{
+    auto rv = false;
```
linusg, line 163: let's find a better name for this, something like `success` (initialized as `true`)

Userland/Applications/FileManager/FileUtils.cpp
```diff
@@ -67,4 +70,103 @@ void delete_paths(const Vector<String>& paths, bool should_confirm, GUI::Window*
...
+
+    for (auto& path : paths) {
+        rv = move_to_trash(path);
+    }
```
linusg, line 167: This will return a wrong result if moving fails for any but the last file - you don't want to overwrite the variable.

```suggestion
    for (auto& path : paths)
        success &= move_to_trash(path);
```

Userland/Applications/FileManager/FileUtils.h
```diff
@@ -20,4 +20,8 @@ enum class FileOperation {
 
 void delete_path(const String&, GUI::Window*);
 void delete_paths(const Vector<String>&, bool should_confirm, GUI::Window*);
+void create_trash_bin();
+[[maybe_unused]] bool move_to_trash(String const&);
+[[maybe_unused]] bool move_to_trash(Vector<String> const&);
```
linusg, line 25: Same as above, why `[[maybe_unused]]`?

Userland/Applications/FileManager/main.cpp
```diff
@@ -779,13 +781,23 @@ int run_in_windowed_mode(RefPtr<Core::ConfigFile> config, String initial_locatio
         },
         &tree_view);
 
+    auto tree_view_trash_action = GUI::Action::create(
+        "Move to &Trash", Gfx::Bitmap::load_from_file("/res/icons/16x16/delete.png"), [&](auto&) {
```
linusg, line 785: Please at least put a FIXME here and note that we should add and use a trash can icon :)

linusg (changes requested):
Looks good overall :)

PR #6825, opened 2021-05-02
Userland: Change IPC funcs to use plain arguments instead of a struct

gunnarbeutner:
This changes all IPC methods to use plain arguments instead a `Message` struct which removes a lot of repetitive boilerplate code. Also, instead of having a single overloaded `handle` method each method gets its own unique method name now.

Also, we auto-generate client methods now. For example:

`decode_image(Core::AnonymousBuffer data) => (bool is_animated, u32 loop_count, Vector<Gfx::ShareableBitmap> bitmaps, Vector<u32> durations);`

This will generate the following client methods:

```C++
    Messages::ImageDecoderServer::DecodeImageResponse decode_image(Core::AnonymousBuffer data);
    void async_decode_image(Core::AnonymousBuffer data);
    Result<Messages::ImageDecoderServer::DecodeImageResponse, IPC::ErrorCode> try_decode_image(Core::AnonymousBuffer data);
```

`decode_image` is synchronous and returns the peer's result - or crashes if the peer dies unexpectedly.
`async_decode_image` sends the decode request but doesn't wait for the result.
`try_decode_image` is synchronous but will return an `IPC::ErrorCode` if the peer dies.

Having these client methods also allows for better code completion support in IDEs because we know which messages are valid for each client connection object:

Screenshot from 2021-05-03 18-12-25

IPC methods cannot currently forward error codes to the client but this could be extended at a later point.

For IPC methods which only return a single return value those are returned directly. For example `get_main_mix_volume() => (i32 volume)` gets client methods with the signatures

```C++
i32 get_main_mix_volume();
void async_get_main_mix_volume();
Result<i32, IPC::ErrorCode> try_get_main_mix_volume();
```

This cuts down IPC boilerplate code to a bare minimum. Old example:

```C++
String Desktop::wallpaper() const
{
    return WindowServerConnection::the().send_sync<Messages::WindowServer::GetWallpaper>()->path();
}
```

Equivalent new code:

```C++
String Desktop::wallpaper() const
{
    return WindowServerConnection::the().get_wallpaper();
}
```


alimpfard:
Not gonna lie, I'm not a big fan of this change:
- Not sure what boilerplate you're talking about, the only changed line in your example could be reduced previously as well (unless you're talking about the generated code (I'm not sure if the generated code is actually reduced), in which case, who cares)
- Looking through the changes, a point could be made about `message.foo()` being more indicative of where the value comes from, but that's just a stylistic preference by me, so let's ignore it
- More importantly, searches for the message names no longer bring up the appropriate handlers, it's a layer of "magic" that not everyone can see through
- Also particularly important is the fact that the IPC code generator now needs to [know about any changes (particularly additions) to the types](https://github.com/SerenityOS/serenity/pull/6825/commits/c14b229dfd9382924050b6bbfda582cca450c837#diff-ce24c4805c668159f6dd1e89bbd7d20deb3729ef518f57087c7d49f775f97da5R637-R651), though I think that's just an optimisation to copy trivially copyable types?
- The handlers are particularly painful to read when they have a long list of arguments (e.g. [this one](https://github.com/SerenityOS/serenity/pull/6825/commits/c14b229dfd9382924050b6bbfda582cca450c837#diff-47512e377fb6d4f9c3832ce2bb47cac6c318c134ce4c7fefeafa0392d63c5ea7R31)), though that's just my opinion

Overall, this doesn't feel like it's actually making things better, but rather adding another layer of abstraction and/or magic (i.e. what is this improving, exactly?)

gunnarbeutner:
You make some very valid points there. I guess I'll have to think about whether this is worth salvaging.

The basic idea was to have method signatures for IPC handlers that felt more natural, but I guess that's a matter of personal preference.

I can definitely see the problem with not being able to grep for handler names anymore - and also the problem with handlers that have lots of arguments. We'll see if I can figure out something for that or maybe scrap this entirely. :)

gunnarbeutner:
I've updated the description to reflect the latest changes. For now, I don't have a solution for functions with lots of arguments.

alimpfard:
Then why not let them have the structs as arguments for now?
I'm particularly concerned with this because changing existing endpoints is going to be an absolute pain, and I've had to modify endpoints enough times to be worried about this.

+1 for making post/send_sync more ergonomic.

awesomekling:
What is the problem with many parameters? You still have to change the code *inside* the function to use the parameters you added anyway.

What's the pain point? The fact that you have to edit the function signature as well, and not just the function body?

alimpfard:
Yes, exactly.
Modifying function parameters is _terrible_, especially when they're of nondescript types like `u32` and `String`.
Note that I wouldn't _have_ to change the function body for the code to compile and work if the change was just an addition.

I'll concede this one thing if you and @gunnarbeutner are so interested in it, though I personally don't see any pros, there won't be any performance wins, and there won't be any readability wins either.
My main problem with this was not being able to search for the messages easily, and that has been resolved.

awesomekling:
Let's roll with this :)

alimpfard:
🤷‍♂️

awesomekling:
Fixed conflicts and merged @ 8a6db55e79ee484bc071fcb52ca30da069231f77

PR #6826, opened 2021-05-02
LibGUI: Debounce TextDocument undo stack

pakerwreah:
Replace the repeating 2-sec timer with a debounced single-shot timer on user input.


awesomekling:
Awesome, thank you for fixing this up :)
Also TIL the word "debounce"!

PR #6829, opened 2021-05-03
Documentation: Update stated minimum GCC version to 11

blam23:
PR #6552 updated the tool-chain to use gcc11, and it is now required to build the latest source.
This commit updates the documentation to reflect the new minimum required GCC version.

### DIFF
--- a/Documentation/BuildInstructions.md
+++ b/Documentation/BuildInstructions.md
@@ -6,7 +6,7 @@
 
 Ensure your CMake version is >= 3.16 with `cmake --version`. If your system doesn't provide a suitable version of CMake, you can download a binary release from the [CMake website](https://cmake.org/download).
 
-Ensure your gcc version is >= 10 with `gcc --version`. Otherwise, install it.
+Ensure your gcc version is >= 11 with `gcc --version`. Otherwise, install it.
 
 On Ubuntu it's in the repositories of 20.04 (Focal) and later - add the `ubuntu-toolchain-r/test` PPA if you're running an older version:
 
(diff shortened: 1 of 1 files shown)


bgianfo:
Their is nothing requiring the host toolchain to be gcc-11 though AFAIK? The requirement of gcc 10.2 for the host toolchain was for some advanced concepts usage in AK if I recall correctly? It would be a shame to tell people they need a new host gcc and have them go through the trouble of acquiring it, when it's not needed.

Documentation/BuildInstructions.md
```diff
@@ -21,11 +21,11 @@ sudo echo "deb http://http.us.debian.org/debian/ testing non-free contrib main"
 sudo apt update
 ```
 
-Now on Ubuntu or Debian you can install gcc-10 with apt like this:
+Now on Ubuntu or Debian you can install gcc-11 with apt like this:
 
 ```console
-sudo apt install gcc-10 g++-10
-sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-10 900 --slave /usr/bin/g++ g++ /usr/bin/g++-10
+sudo apt install gcc-11 g++-11
```
gunnarbeutner, line 27: Debian doesn't have `gcc-11` in any released distribution.

Documentation/BuildInstructions.md
```diff
@@ -112,7 +112,7 @@ Make sure you have all the dependencies installed:
 
 ```console
 # core
-brew install coreutils qemu bash gcc@10 ninja cmake
+brew install coreutils qemu bash gcc@11 ninja cmake
```
gunnarbeutner, line 115: This doesn't exist either.

drunderscore:
Looks like you're confused between the toolchain compiler and the host compiler -- only the toolchain one was updated.

The toolchain one is used for cross-compiling to Serenity from your host. Serenity doesn't _(yet)_ make use of any gcc 11 features, and isn't even available in most package managers yet. No reason to force people to needlessly upgrade.

PR #6832, opened 2021-05-03
DynamicLoader: Remove math functionality in favor of -lgcc

gunnarbeutner:
This links the dynamic linker against `libgcc.a` instead of having our own copy of the math functions.

Also, this fixes the `-nodefaultlibs` flag for gcc which previously linked against `libgcc_s.so` anyway. Once we have everyone on the new toolchain (which is not a required rebuild at this time) we can also remove the -fbuilding-libgcc flag in `Userland/DynamicLoader/CMakeLists.txt`.


gunnarbeutner:
Well, this latest patch now actually works. The reason why `-Wl,--whole-archive` is necessary is that `-shared` implies `-lgcc_s`, i.e. we're linking `Loader.so` against `libgcc_s.so` (the shared version of `libgcc`) and the linker prefers the symbols from the shared library over the static library:

```C
#undef LINK_SPEC
#define LINK_SPEC "%{shared:-shared} %{static:-static} %{!static: %{rdynamic:-export-dynamic} %{!fbuilding-libgcc:-lgcc_s -dynamic-linker /usr/lib/Loader.so}}"
```

I tried various tricks to get gcc to use the static library instead (argument order for the linker, `-Wl,--undefined`)

Using `-Wl,--whole-archive` makes the resulting binary slightly larger than what we have in the master branch (about 200kB).

PR #6833, opened 2021-05-03
LibC: Fix invalid 1-byte read I introduced in dirent.

bgianfo:
When attempting to fix the dirent code I also changed
this to use strlcpy instead of the custom string copy
loop that was there before. Looking over strlcpy it
looked like it should work when using a non null terminated
string, I obviously misinterpreted the implementation
as it will read till it finds a null terminator.

Manually null terminate the string to address this.

Gunnar found this after he fixed UserspaceEmulator.
I reproduced it locally using his branch, and also
found the memory leak I had in the unit test for the
scandir that I added, so lets fix that as well.

Reported-by: Gunnar Beutner <gbeutner@serenityos.org>

### DIFF
--- a/Userland/Libraries/LibC/dirent.cpp
+++ b/Userland/Libraries/LibC/dirent.cpp
@@ -72,8 +72,13 @@ static void create_struct_dirent(sys_dirent* sys_ent, struct dirent* str_ent)
     str_ent->d_off = 0;
     str_ent->d_reclen = sizeof(struct dirent);
 
-    int size = min((sys_ent->namelen + 1) * sizeof(char), sizeof(str_ent->d_name));
-    [[maybe_unused]] auto n = strlcpy(str_ent->d_name, sys_ent->name, size);
+    VERIFY(sizeof(str_ent->d_name) > sys_ent->namelen);
+
+    // Note: We can't use any normal string function as sys_ent->name is
+    // not null terminated. All string copy functions will attempt to read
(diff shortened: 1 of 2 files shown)


bgianfo:
cc: @gunnarbeutner

Userland/Libraries/LibC/dirent.cpp
```diff
@@ -72,8 +72,13 @@ static void create_struct_dirent(sys_dirent* sys_ent, struct dirent* str_ent)
     str_ent->d_off = 0;
     str_ent->d_reclen = sizeof(struct dirent);
 
-    int size = min((sys_ent->namelen + 1) * sizeof(char), sizeof(str_ent->d_name));
-    [[maybe_unused]] auto n = strlcpy(str_ent->d_name, sys_ent->name, size);
+    VERIFY(sizeof(str_ent->d_name) > sys_ent->namelen * sizeof(char));
```
gunnarbeutner, line 75: ˋsizeof(char)ˋ is always 1, so that’s probably redundant.

bgianfo: Everyone loves to tell me this 😄, it's just a habit for clarity... but i'm not attached to it.

bgianfo: fixed

gunnarbeutner:
LGTM. :+1:

gunnarbeutner (approved):

PR #6834, opened 2021-05-03
Kernel: Remove unused header includes + mark some classes as final.

bgianfo:
-    Kernel: Remove unused header includes from various files.

     Found while browsing code with CLion.

-  Kernel: Mark AsyncBlockDeviceRequest + AnonymousVMObject as final

    Mark final to aid in de-virtualization since they are not currently
    derived from.


awesomekling:
Typo, aide -> aid

bgianfo:
sorry about that, fixed.

PR #6836, opened 2021-05-03
Userland: Implement a file utility

vkoskiv:
My first larger Serenity contribution in a while, so I'm expecting to change many aspects of this based on feedback.

Fire away! :^)


vkoskiv:
Few remarks so far:
- Strictly using mime-types for file detection doesn't work, as many formats we might like to detect are not common where mime types are used, and thus lack a standard notation. I propose we either rethink using `Core::MimeData` for binary file detection and instead move that logic to `file`, or we come up with a custom mime type that relays the information that "this thing doesn't have a mime type, but here's what it is anyway" and then parse that in `file`. I've elected to try `extra/<custom type>` for now, but we could also go with `application/x-<custom type>`
- Writing optional description generators for new file formats should be as easy as possible. A description generator reads a given file, and returns a string with useful information. I've implemented this for image files supported by `LibGfx::ImageDecoder` to start with.

vkoskiv:
Feel free to review what's in there now, I'm adding new types and tweaks as I go here.

Userland/Libraries/LibCore/MimeData.cpp
```diff
@@ -79,4 +79,31 @@ String guess_mime_type_based_on_filename(const StringView& path)
...
+    V(pgm, "image/x-portable-graymap", 3, 0x50, 0x32, 0x0A)                                            \
+    V(png, "image/png", 8, 0x89, 'P', 'N', 'G', 0x0D, 0x0A, 0x1A, 0x0A)                                \
+    V(ppm, "image/x-portable-pixmap", 3, 0x50, 0x33, 0x0A)                                             \
+    V(gif_87, "image/gif", 6, 'G', 'I', 'F', '8', '7', 'a')                                            \
+    V(gif_89, "image/gif", 6, 'G', 'I', 'F', '8', '9', 'a')                                            \
+    V(bmp, "image/bmp", 2, 'B', 'M')                                                                   \
+    V(jpeg, "image/jpeg", 4, 0xFF, 0xD8, 0xFF, 0xDB)                                                   \
+    V(jpeg_jfif, "image/jpeg", 12, 0xFF, 0xD8, 0xFF, 0xE0, 0x00, 0x10, 'J', 'F', 'I', 'F', 0x00, 0x01) \
+    V(jpeg_huh, "image/jpeg", 4, 0xFF, 0xD8, 0xFF, 0xEE)                                               \
+    V(shell, "text/x-shellscript", 10, '#', '!', '/', 'b', 'i', 'n', '/', 's', 'h', '\n')
```
linusg: I wonder if it would be useful to alphabetically sort this list? Could get messy once more are added :)

Userland/Libraries/LibCore/MimeData.h
```diff
@@ -49,4 +49,5 @@ class MimeData : public Object {
 
 String guess_mime_type_based_on_filename(const StringView&);
 
+Optional<String> guess_mime_type_based_on_sniffed_bytes(const ReadonlyBytes&);
```
linusg, line 52:
```suggestion
Optional<String> guess_mime_type_based_on_sniffed_bytes(const ReadonlyBytes&);

```

Userland/Utilities/file.cpp
```diff
@@ -0,0 +1,103 @@
...
+#include <LibCore/MimeData.h>
+#include <LibGfx/ImageDecoder.h>
+#include <stdio.h>
+#include <unistd.h>
+
+static Optional<String> placeholder(String description, String& path)
+{
+    (void)path;
+    return description;
+}
```
linusg, line 20: We usually use `[[maybe_unused]]` instead of `(void)`'ing unused arguments. Also I think something like `description_only` is a slightly better name for this.

```suggestion
static Optional<String> description_only(String description, [[maybe_unused]] const String& path)
{
    return description;
}
```

Userland/Utilities/file.cpp
```diff
@@ -0,0 +1,103 @@
...
+
+//FIXME: Ideally Gfx::ImageDecoder could tell us the image type directly.
+static Optional<String> image_details(String description, String& path)
```
linusg, line 23:
```suggestion
// FIXME: Ideally Gfx::ImageDecoder could tell us the image type directly.
static Optional<String> image_details(const String& description, const String& path)
```

Userland/Utilities/file.cpp
```diff
@@ -0,0 +1,103 @@
...
+{
+    auto file_or_error = MappedFile::map(path);
+    if (file_or_error.is_error()) {
+        return {};
+    }
```
linusg, line 28:
```suggestion
    if (file_or_error.is_error())
        return {};
```

Userland/Utilities/file.cpp
```diff
@@ -0,0 +1,103 @@
...
+    auto& mapped_file = *file_or_error.value();
+    auto image_decoder = Gfx::ImageDecoder::create((const u8*)mapped_file.data(), mapped_file.size());
+
+    if (!image_decoder->is_valid())
+        return {};
+
+    StringBuilder builder;
+    builder.append(description);
+    builder.append(String::formatted(", {} x {}", image_decoder->width(), image_decoder->height()));
+    return builder.to_string();
```
linusg, line 39: Don't think we need a string builder for this:

```suggestion
    return String::formatted("{}, {} x {}", description, image_decoder->width(), image_decoder->height());
```

Userland/Utilities/file.cpp
```diff
@@ -0,0 +1,103 @@
...
+    args_parser.set_general_help("Determine type of files");
+    args_parser.add_option(flag_mime_only, "Only print mime type", "mime-type", 'I');
+    args_parser.add_positional_argument(paths, "Files to identify", "files", Core::ArgsParser::Required::Yes);
+    args_parser.parse(argc, argv);
+
+    for (auto path : paths) {
+        auto file = Core::File::construct(path);
+        if (!file->open(Core::File::ReadOnly)) {
+            perror(path);
+            return -1;
```
linusg, line 93:
```suggestion
            return 1;
```

Userland/Utilities/file.cpp
```diff
@@ -0,0 +1,103 @@
...
+        }
+        auto bytes = file->read(25);
+        auto file_name_guess = Core::guess_mime_type_based_on_filename(path);
+        auto mime_type = Core::guess_mime_type_based_on_sniffed_bytes(bytes.bytes()).value_or(file_name_guess);
+        auto human_readable_description = get_description_from_mime_type(mime_type, String(path)).value_or(mime_type);
+        printf("%s: %s\n", path, flag_mime_only ? mime_type.characters() : human_readable_description.characters());
```
linusg, line 99:
```suggestion
        outln("{}: {}", path, flag_mime_only ? mime_type : human_readable_description);
```

linusg (commented):
Please also update the last two commits so the `ENUMERATE_HEADER_CONTENTS` macro isn't broken in the first and fixed in the second - squash the fix into the commit where it's added instead :)

vkoskiv:
All done. Thanks for the great code review, Linus!

vkoskiv:
Squashed one more linter fix in there, *now* it's ready! @linusg

Userland/Libraries/LibCore/MimeData.cpp
```diff
@@ -72,11 +72,49 @@ String guess_mime_type_based_on_filename(const StringView& path)
...
+    if (path.ends_with(".md", CaseSensitivity::CaseInsensitive))
+        return "text/markdown";
     if (path.ends_with("/", CaseSensitivity::CaseInsensitive))
         return "text/html";
     if (path.ends_with(".csv", CaseSensitivity::CaseInsensitive))
         return "text/csv";
     return "text/plain";
 }
 
+#define ENUMERATE_HEADER_CONTENTS(V)                                                                   \
```
linusg, line 88: These usually follow the pattern (with some exceptions)

```
#define ENUMERATE_FOOBARS   \
    __ENUMERATE_FOOBAR(...) \
    __ENUMERATE_FOOBAR(...) \
    ...
```

Can we do the same here? `V` is not a great name :) Maybe something like `ENUMERATE_MIME_TYPE_HEADERS` / `__ENUMERATE_MIME_TYPE_HEADER`.

Userland/Libraries/LibCore/MimeData.cpp
```diff
@@ -72,11 +72,49 @@ String guess_mime_type_based_on_filename(const StringView& path)
...
+    V(pbm, "image/x-portable-bitmap", 3, 0x50, 0x31, 0x0A)                                             \
+    V(pgm, "image/x-portable-graymap", 3, 0x50, 0x32, 0x0A)                                            \
+    V(png, "image/png", 8, 0x89, 'P', 'N', 'G', 0x0D, 0x0A, 0x1A, 0x0A)                                \
+    V(ppm, "image/x-portable-pixmap", 3, 0x50, 0x33, 0x0A)                                             \
+    V(shell, "text/x-shellscript", 10, '#', '!', '/', 'b', 'i', 'n', '/', 's', 'h', '\n')              \
+    V(tiff, "image/tiff", 4, 'I', 'I', '*', 0x00)                                                      \
+    V(tiff_bigendian, "image/tiff", 4, 'M', 'M', 0x00, '*')
+
+#define V(var_name, mime_type, pattern_size, ...) \
+    static const u8 var_name[pattern_size] = { __VA_ARGS__ };
```
linusg, line 107: Any reason to not just construct the `ReadonlyBytes` here (rather than in each `guess_mime_type_based_on_sniffed_bytes` call)? Even has `constexpr` support :)

Userland/Utilities/file.cpp
```diff
@@ -0,0 +1,98 @@
...
+#include <AK/MappedFile.h>
+#include <AK/Vector.h>
+#include <LibCore/ArgsParser.h>
+#include <LibCore/FileStream.h>
+#include <LibCore/MimeData.h>
+#include <LibGfx/ImageDecoder.h>
+#include <stdio.h>
+#include <unistd.h>
+
+static Optional<String> description_only(String description, [[maybe_unused]] String& path)
```
linusg, line 16:
```suggestion
static Optional<String> description_only(String description, [[maybe_unused]] const String& path)
```

(not that it really matters here :smile:)

linusg (commented):
sorry for missing this stuff earlier, I'm not super good at review-all-in-one-go sometimes :sweat_smile:

Userland/Libraries/LibCore/MimeData.cpp
vkoskiv: It sure isn't! I just wanted to keep the lines shorter. I'll change it.

Userland/Libraries/LibCore/MimeData.cpp
vkoskiv: I had a lot of trouble trying to understand the code in `Span.h`, so I couldn't figure out a way to construct it there without `no matching constructor` errors or weird hacky solutions.

Userland/Utilities/file.cpp
vkoskiv: You know what they say - `const`sistency is key!

(sorry)

vkoskiv:
@linusg All done, thanks for the thorough review!

vkoskiv:
@linusg Rebased on upstream, should be good to merge now.

linusg:
Cool! Wanna write a man page for it next? ;)

PR #6837, opened 2021-05-03
Storage: Rearrange IDEChannel detection if's

ElectrodeYT:
In the current form, the if statement detecting device presence will never fire if status == 0xFF

### DIFF
--- a/Kernel/Storage/IDEChannel.cpp
+++ b/Kernel/Storage/IDEChannel.cpp
@@ -334,6 +334,13 @@ UNMAP_AFTER_INIT void IDEChannel::detect_disks()
             if (milliseconds_elapsed > 2000)
                 break;
             u8 status = m_io_group.control_base().in<u8>();
+
+            if (status == 0 || status == 0xFF) {
+                dbgln_if(PATA_DEBUG, "IDEChannel: {} {} device presence - none.", channel_type_string(), channel_string(i));
+                device_presence = false;
+                break;
+            }
+
             if (status & ATA_SR_ERR) {
                 dbgln_if(PATA_DEBUG, "IDEChannel: {} {} device is not ATA. Will check for ATAPI.", channel_type_string(), channel_string(i));
                 check_for_atapi = true;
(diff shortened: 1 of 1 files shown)


ElectrodeYT:
I looked around the internet a bit and now im not even sure if 0xFF is even a "no device" status; as far as i can see if the error bit is set then the error must be read from a seperate register, meaning no other bits should be valid. I also noted how on bare metal hardware it stops detecting the dvd drive as a ATAPI device.

supercomputer7:
> I looked around the internet a bit and now im not even sure if 0xFF is even a "no device" status; as far as i can see if the error bit is set then the error must be read from a seperate register, meaning no other bits should be valid. I also noted how on bare metal hardware it stops detecting the dvd drive as a ATAPI device.

According to https://wiki.osdev.org/ATA_PIO_Mode#Standard_and_Non-standard_Detection, you're indeed right that checking for floating bus won't work with ATAPI devices. 0xFF is not a valid value, as it represents both ready and error state at the same time, which is an "incorrect" (special) behavior as far as I remember from the ATA specification.
That being said, I'm not sure the right way to handle this special behavior is to move the if statement inside the loop - I don't think there's anything limiting the hardware from setting the status register to zero before setting the RDY or ERR bit... so by doing this change, we could create a regression on real hardware, which is not a good thing.

ElectrodeYT:
> That being said, I'm not sure the right way to handle this special behavior is to move the if statement inside the loop - I don't think there's anything limiting the hardware from setting the status register to zero before setting the RDY or ERR bit... so by doing this change, we could create a regression on real hardware, which is not a good thing.

I take this to mean setting ERR before clearing RDY / Setting RDY before clearing ERR?
What could be done in an "error" state is waiting for a short time to see if it resolves itself, as it may still be changing flags when one is reading said flags. This likely won't fix some buggy or non-standard hardware, which is where I found this. (a Nvidia chipset board on a Core2Quad, that uses IDE for its sata ports)

PR #6841, opened 2021-05-03
Lagom/CMake: Fix invalid constexpr error using AppleClang

pakerwreah:
We are getting an error in Color.h `interpolate` because `roundf` is not constexpr in AppleClang.

### DIFF
--- a/Meta/Lagom/CMakeLists.txt
+++ b/Meta/Lagom/CMakeLists.txt
@@ -11,6 +11,10 @@ if ("${CMAKE_CXX_COMPILER_ID}" MATCHES "Clang$")
     # Clang's default constexpr-steps limit is 1048576(2^20), GCC doesn't have one
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-overloaded-virtual -Wno-user-defined-literals -fconstexpr-steps=16777216")
 
+    if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "AppleClang")
+        set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-invalid-constexpr")
+    endif()
+
     if (ENABLE_ADDRESS_SANITIZER)
         add_definitions(-fsanitize=address -fno-omit-frame-pointer)
         set(LINKER_FLAGS "${LINKER_FLAGS} -fsanitize=address")


linusg:
This doesn't fix the error though, but instead just hides it?!

IdanHo:
note also that our build_and_test_lagom workflow tests build serenity using Clang and do not experience this error, so your local toolchain might be in need of some updates :)

pakerwreah:
@IdanHo

This is what I have locally:
```
Apple clang version 12.0.5 (clang-1205.0.22.9)
Target: x86_64-apple-darwin20.4.0
```
I get the same error in Compiler Explorer.

Did you mean I should rebuild Serenity's toolchain?

IdanHo:
our CI has `Apple clang version 12.0.0` installed, so yours should work fine as well

pakerwreah:
I think the CI is only checking Clang's version but then it builds with GCC 👀 
```
cmake -GNinja -DBUILD_LAGOM=ON -DENABLE_PCI_IDS_DOWNLOAD=OFF -DCMAKE_C_COMPILER=gcc-10 -DCMAKE_CXX_COMPILER=g++-10 ..
  shell: /bin/bash -e {0}
  env:
    SERENITY_SOURCE_DIR: /Users/runner/work/serenity/serenity
```

IdanHo:
> I think the CI is only checking Clang's version but then it builds with GCC 
> 
> ```
> cmake -GNinja -DBUILD_LAGOM=ON -DENABLE_PCI_IDS_DOWNLOAD=OFF -DCMAKE_C_COMPILER=gcc-10 -DCMAKE_CXX_COMPILER=g++-10 ..
>   shell: /bin/bash -e {0}
>   env:
>     SERENITY_SOURCE_DIR: /Users/runner/work/serenity/serenity
> -- The C compiler identification is GNU 10.2.0
> -- The CXX compiler identification is GNU 10.2.0
> ```

the fuzz builds at least are built with clang

pakerwreah:
I've created a branch on my repository to test the CI for macOS with Clang.
It won't build because it's using Xcode 12.4 with `Apple clang version 12.0.0` which doesn't understand the `requires` keyword.

With `Apple clang version 12.0.5` it builds fine, but we do need to add `-Wno-invalid-constexpr`
If we can't do that, I guess I'll have no support for AppleClang anymore. 😢

pakerwreah:
What if I change CMake to match exactly "AppleClang" and add that flag? 🤔 💭 
It wouldn't have any impact on "pure" Clang and it works for everyone.

alimpfard:
Please note that "Apple clang 12" is not "clang 12", IIRC their latest equivalent version is clang 11.1.0.

ADKaster:
I'm not sure if it's ever been a project goal to support Apple Clang for Lagom builds.

As Ali noted:

> Please note that "Apple clang 12" is not "clang 12", IIRC their latest equivalent version is clang 11.1.0.

Apple Clang provided with XCode lags behind upstream llvm, and the version numbers don't necessarily correlate (and seen to have almost never correlated).

We've been continually updating the used C++20 features as allowed by upstream llvm and upstream gcc. The fuzzer build, as noted, has been using clang++-12 for quite some time (the oss Fuzz build was using it long before we changed it our CI to build the reference fuzzer build against clang-12).

With alllllllll of that being said. roundf is not constexpr in clang-12 or clang trunk either, so clearly we're not fuzzing the affected function at all :)

https://godbolt.org/z/escMdbjj6

So a real fix would either be, proscribe use of gcc which does implement p0533, or make the offending functions not constexpr.

alimpfard:
`roundf` being constexpr depends on the libc implementation (or rather, declaration)
I think the compiler accepts it because our libc marks it `noexcept` when included in C++ code, but I doubt that's something most other libc's do...
I don't think p0533 will do much, since...you know, we're including the C header, not the C++ wrapper header :P 

I think it's okay to just mark the function `noexcept(noexcept(roundf(declval<float>())))`...

pakerwreah:
I guess @ADKaster has a point. I'll always be lagging behind trying to fix these issues if I keep using AppleClang.
I just installed llvm via homebrew and everything is fine now. 🙃
Thanks everyone! I'll close this PR for now.

PR #6849, opened 2021-05-04
LibCore: Let Account::from_* succeed if /etc/shadow is unreadable 

gunnarbeutner:
This previously worked and was broken by 302f979.


linusg:
:tada:

PR #6851, opened 2021-05-04
Add LibPDF and a PDFViewer

mattco98:
This PR introduces a PDF parser, renderer, and a very bare-bones viewer application. 

### Parser

The parser is capable of the following:
- Parsing all PDF objects
- Parsing of all graphics commands
- Navigating the page tree structure
- Lazy parsing of pages (we only parse the pages as the viewer needs them)

The parser has the following limitations:
- It cannot handle linearized PDF files
- It cannot handle incrementally updated PDF files
- It is not in any capacity prepared for the smallest deviation from the spec
  - There are `VERIFY` statements all over the parser
  - I have found multiple PDF files which, as far as I can tell are ill-formed according to the spec, but which Firefox still manages to render perfectly.

### Renderer

The renderer is capable of the following:
- Displaying basic text
- Displaying basic shapes
- Rendering text in the LiberationSerif TTF font, which this PR adds
  - The actual font used in a PDF file is ignored beyond whether the name contains `bold`/`italic`/`regular`

The renderer has the following limitations:
- Any advanced PDF rendering is unsupported, such as color spaces, images, or complex clipping
- As mentioned above, it only uses the one font. The PDF specification names 14 intrinsic fonts which should be supported by all PDF renderers. I would love to support them, but I am not sure how to as some of them are not open fonts (like Times New Roman)

### PDFViewer

The viewer is capable of the following:
- Displaying a single page at a time
- Zooming in and out of the viewed page
- Caching bitmaps of pages (including zoom) to avoid re-rendering the same thing twice

The viewer has the following limitations:
- It is VERY empty. The is one menubar entry and a single page in the middle of the application, which doesn't look very good.
- No continuous scroll (i.e. having multiple pages on the screen at the same time), which in my opinion is a much more natural way to view a PDF

### Demo

<details>
<summary>Here is a PDF file displayed in Firefox</summary>
<br />

</details>

<details>
<summary>And the same PDF displayed in Serenity</summary>
<br />

</details>


mattco98:
Closing to split into two much cleaner, easier-to-review PRs

PR #6673, opened 2021-04-26
AK: Variant<Ts...> / LibWasm: Start implementing a basic WebAssembly binary format parser

alimpfard:
Current state:
- Can parse anything I threw at it (as generated by clang, AssemblyScript, and Rust, also whatever `jq.wasm.wasm` was made with :P)
- Implements most of module instantiation, is missing some data initialisations (passive data will always be zero), tables will not be initialised
- Does _not_ implement validation (this can be done in a future PR)
- Implements most of the instructions, except the following 
<details> <summary>Expand for unimplemented instructions</summary>

- br.table
- table.init
- table.get
- table.set
- table.copy
- table.size
- table.grow
- table.fill
- ref.null
- ref.func
- ref.is_null
- i32/i64.clz
- i32/i64.ctz
- i32/i64.popcnt
- i32/i64.rotl
- i32/i64.rotr
- X.trunc.Y
- X.trunc_sat.Y
- memory.init
- memory.copy
- memory.fill
- elem.drop
- data.drop

</details>

- No JS integration (the WebIDL definitions seemed a bit impossible for our idl wrapper generator)
- `wasm` utility can parse, print, instantiate and execute a function from a wasm module (with user-supplied args)

On the non-Wasm changes (`AK::Variant<Ts...>`)
- Can't be valueless (that's a really bad API)
- Nice `.visit()` method with builtin overload resolution (can pass multiple visitors)
- `.get<T>()` method, will VERIFY that it contains the type requested, use `.get_pointer<T>()` if you're not sure that the type is there, will either return a pointer to the value, or a nullptr

<details>
<summary> 
(Original PR message)
</summary>
This can currently parse and dump a simple module.
This also adds a 'wasm' test utility that tries to parse a given webassembly binary file, and can print out its contents (mostly, instruction arguments are not printed)

image_2021-04-27_22-06-23
</details>

PR #6855, opened 2021-05-04
AK: Add a Variant<Ts...> implementation

alimpfard:
Part 1/n of #6673
About `AK::Variant<Ts...>`:

- Can't be valueless (that's a really bad API)
- Nice `.visit()` method with builtin overload resolution (can pass multiple visitors)
- `.get<T>()` method, will VERIFY that it contains the type requested, use `.get_pointer<T>()` if you're not sure that the type is there, will either return a pointer to the value, or a nullptr

Currently not used anywhere.

### DIFF
--- a/AK/Array.h
+++ b/AK/Array.h
@@ -66,6 +66,26 @@ struct Array {
         return Size;
     }
 
+    constexpr T max() requires(requires(T x, T y) { x < y; })
+    {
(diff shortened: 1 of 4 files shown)


alimpfard:
(Please hold off on merge while I write some more tests)

awesomekling:
Very tasteful! :star:

PR #6868, opened 2021-05-04
LibGUI: Remember modified state on undo/redo actions

pakerwreah:
This change allows any application that uses the undo stack to keep track of the modified state for each undo/redo action.

### DIFF
--- a/Userland/Libraries/LibGUI/UndoStack.cpp
+++ b/Userland/Libraries/LibGUI/UndoStack.cpp
@@ -80,4 +80,10 @@ void UndoStack::finalize_current_combo()
     m_stack.prepend(move(undo_commands_container));
 }
 
+void UndoStack::clear()
+{
+    m_stack.clear();
+    m_stack_index = 0;
+}
+
 }
--- a/Userland/Libraries/LibGUI/UndoStack.cpp
+++ b/Userland/Libraries/LibGUI/UndoStack.cpp
@@ -61,6 +61,14 @@ void UndoStack::push(NonnullOwnPtr<Command>&& command)
     if (m_stack_index > 0) {
         for (size_t i = 0; i < m_stack_index; i++)
             m_stack.remove(0);
+
+        if (m_clean_index.has_value()) {
+            if (m_clean_index.value() < m_stack_index)
+                m_clean_index.clear();
+            else
+                m_clean_index = m_clean_index.value() - m_stack_index;
+        }
+
         m_stack_index = 0;
         finalize_current_combo();
     }
@@ -78,12 +86,30 @@ void UndoStack::finalize_current_combo()
 
     auto undo_commands_container = make<UndoCommandsContainer>();
     m_stack.prepend(move(undo_commands_container));
+
+    if (m_clean_index.has_value())
+        m_clean_index = m_clean_index.value() + 1;
+}
+
+void UndoStack::set_current_unmodified()
+{
+    // Skip empty container
+    if (can_undo() && m_stack[m_stack_index].m_undo_vector.is_empty())
+        m_clean_index = m_stack_index + 1;
+    else
+        m_clean_index = m_stack_index;
+}
+
+bool UndoStack::is_current_modified() const
+{
+    return !m_clean_index.has_value() || m_stack_index != m_clean_index.value();
 }
 
 void UndoStack::clear()
 {
     m_stack.clear();
     m_stack_index = 0;
+    m_clean_index.clear();
 }
 
 }
--- a/Userland/Libraries/LibGUI/UndoStack.h
+++ b/Userland/Libraries/LibGUI/UndoStack.h
@@ -26,6 +26,8 @@ public:
 
     void finalize_current_combo();
 
+    void clear();
+
 private:
     struct UndoCommandsContainer {
         NonnullOwnPtrVector<Command> m_undo_vector;
--- a/Userland/Libraries/LibGUI/UndoStack.h
+++ b/Userland/Libraries/LibGUI/UndoStack.h
@@ -19,12 +19,15 @@ public:
     void push(NonnullOwnPtr<Command>&&);
 
     bool can_undo() const { return m_stack_index < m_stack.size() && !m_stack.is_empty(); }
-    bool can_redo() const { return m_stack_index > 0 && !m_stack.is_empty(); }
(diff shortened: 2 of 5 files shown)


awesomekling (changes requested):
Thank you for working on this!

I think this design is a bit confusing, since only one entry in the stack should be considered the unmodified one.

So instead of remembering this per entry in the undo stack, we should keep track of the undo stack's "clean index", the index at which the stack could be considered unmodified. Basically the same as Qt's [QUndoStack::cleanIndex()](https://doc.qt.io/qt-5/qundostack.html#cleanIndex)

Once this works, `GUI::TextDocument::is_modified()` can simply return whether the document's undo stack is dirty. That removes the need for tracking this state separately. :)

pakerwreah:
This undo stack hurt my brain a little, because it counts backwards and it keeps ~randomly~ prepending an empty container and messing up with my logic 😅

Userland/Libraries/LibGUI/UndoStack.cpp
```diff
@@ -59,7 +59,7 @@ void UndoStack::push(NonnullOwnPtr<Command>&& command)
         finalize_current_combo();
 
     if (m_stack_index > 0) {
-        for (size_t i = 0; i < m_stack_index; i++)
+        for (size_t i = 0; i < m_stack_index; i++, m_clean_index--)
```
awesomekling, line 62: This doesn't seem right. Why does the clean index change when pushing something on the undo stack?
The clean index should stay the same until we save the document (and mark the undo stack clean at its current location.)

Userland/Libraries/LibGUI/UndoStack.h
```diff
@@ -35,6 +38,7 @@ class UndoStack {
 
     NonnullOwnPtrVector<UndoCommandsContainer> m_stack;
     size_t m_stack_index { 0 };
+    size_t m_clean_index { 0 };
```
awesomekling, line 41: I think this needs to be `Optional<size_t>` because there will not always be a valid clean index.

Imagine the following sequence:
* Start (stack_index = 0, clean_index = {})
* Push command (stack_index = 1, clean_index = {})
* Push command (stack_index = 2, clean_index = {})
* Push command (stack_index = 3, clean_index = {})
* Mark stack clean (stack_index = 3, clean_index = 3)
* Undo one command (stack_index = 2, clean_index = 3)
* Undo one command (stack_index = 1, clean_index = 3)
* Push command (stack_index = 1, clean_index = ???)

At that point, I think the clean index should simply be invalid.

Also this is making me wonder if `m_stack_index` should also be `Optional<size_t>`. It's quite weird that it starts out with `0`, even though there are no entries in the stack!

awesomekling (changes requested):

Userland/Libraries/LibGUI/UndoStack.h
pakerwreah: I guess we'll have to rewrite the entire undo stack then... I was avoiding doing that.
I'll test your scenario to see if it breaks the way it is.

pakerwreah: Yeah, we have a bug there 👍🏻

Userland/Libraries/LibGUI/UndoStack.cpp
pakerwreah: The clean index changes because the m_stack_index is counted backwards.
Every time we push a command it is added to the beginning of the vector and I have to correct the clean index to stay where it was.
For example, the scenario in your comment below is not correct because as you add commands to the stack, the stack_index is always zero. When we undo the stack_index increments.

Userland/Libraries/LibGUI/UndoStack.h
pakerwreah: I fixed the bug using `Optional` as you suggested but left the `m_stack_index` as it is for now, since it's not causing any problems and is out of the scope of this PR anyway.

(you taught me not to make drive-by fixes 😄)

Userland/Libraries/LibGUI/UndoStack.cpp
```diff
@@ -61,6 +61,14 @@ void UndoStack::push(NonnullOwnPtr<Command>&& command)
     if (m_stack_index > 0) {
         for (size_t i = 0; i < m_stack_index; i++)
             m_stack.remove(0);
+
+        if (m_clean_index.has_value()) {
+            if (m_clean_index.value() < m_stack_index)
+                m_clean_index.clear();
+            else
+                m_clean_index = m_clean_index.value() - m_stack_index;
```
awesomekling, line 69: I don't understand this line, why does the clean index move when pushing something on the undo stack?

Userland/Libraries/LibGUI/UndoStack.cpp
```diff
@@ -78,12 +86,30 @@ void UndoStack::finalize_current_combo()
 
     auto undo_commands_container = make<UndoCommandsContainer>();
     m_stack.prepend(move(undo_commands_container));
+
+    if (m_clean_index.has_value())
+        m_clean_index = m_clean_index.value() + 1;
```
awesomekling, line 91: I don't understand this, why does the clean index change? It should only change when marking the undo stack clean.

Userland/Libraries/LibGUI/UndoStack.cpp
pakerwreah: That's because we don't append, we prepend commands. 🙈

awesomekling: Oh my goodness.. that's so weird! Why are we doing this 😳

pakerwreah: That's why I said my brain is hurting from working on this 🤣 

- Start (stack_index = 0, clean_index = {})
- Push command (stack_index = 0, clean_index = {})
- Push command (stack_index = 0, clean_index = {})
- Push command (stack_index = 0, clean_index = {})
- Mark stack clean (stack_index = 0, clean_index = 0)
- Undo one command (stack_index = 1, clean_index = 0)
- Undo one command (stack_index = 2, clean_index = 0)
- Push command (stack_index = 0, clean_index = {})

awesomekling: Ok yeah then this patch makes sense. We should really fix the stack to grow in the other direction, too! But we can do that separately :)

PR #6869, opened 2021-05-05
Kernel: Modify TimeManagement::current_time(..) API so it can't fail.

bgianfo:
The fact that current_time can "fail" makes it's use a bit awkward.
All callers in the Kernel are trusted besides syscalls, so assert
that they never get their, and make sure all current callers perform
validation of the clock_id with TimeManagement::is_valid_clock_id().

I have fuzzed this change locally for a bit to make sure I didn't
miss any obvious regression.


linusg:
`it's use` -> `its use`
`that they never get their,` - missing word?

PR #6870, opened 2021-05-05
Kernel: Donate time slices for socket interactions

gunnarbeutner:
This PR implements support for intelligently scheduling threads for IPC interactions:

1. It adds support to the kernel that lets it remember which user threads access a file via read()/write()/select()/poll().
2. Fixes a scheduler hang with non-runnable threads:

When a thread that is non-runnable tried to donate their remaining time slice to another thread that was also non-runnable we'd remove the thread from the ready queue but didn't add it back.

This would cause the thread to never get scheduled again. Previously this could not happen because non-runnable threads had no way of donating their time slice to anyone else.

3. Donates time slices to peer when waiting on a socket:

This causes the user-mode thread to donate their remaining time slice to the peer of a socket when it does a blocking `read()`/`write()` or waits for data with `select()`/`poll()`.

4. Adds the `donate_peer` system call:

This system call donates the calling thread's remaining time slice to the peer of the specified file descriptor (only sockets are
supported for now).

`LibIPC` has been updated to use this after writing a complete IPC message to a socket.

5. Adds statistics for time slices donated/received. These show up in `SystemMonitor`. Note, that `SystemMonitor` does not currently support threads so you might have to look at `/proc/all` instead for some processes.

This might be biased and quite subjective but Super Mario's keyboard input seems significantly more response than it was before. I'd suggest trying this in comparison with the `master` branch.

Screenshot from 2021-05-05 05-10-50

### DIFF
--- a/Kernel/Syscalls/sched.cpp
+++ b/Kernel/Syscalls/sched.cpp
@@ -4,6 +4,7 @@
  * SPDX-License-Identifier: BSD-2-Clause
  */
 
+#include <Kernel/FileSystem/FileDescription.h>
 #include <Kernel/Process.h>
 
 namespace Kernel {
@@ -26,6 +27,21 @@ KResultOr<int> Process::sys$donate(pid_t tid)
     if (!thread || thread->pid() != pid())
         return ESRCH;
     Thread::current()->donate_without_holding_big_lock(thread, "sys$donate");
+    return 0;
+}
+
+KResultOr<int> Process::sys$donate_peer(int fd)
+{
+    REQUIRE_PROMISE(stdio);
+    auto description = file_description(fd);
+    if (!description)
+        return EBADF;
+
+    ScopedCritical critical;
+    auto thread = description->file().likely_peer_thread();
+    if (!thread)
+        return ESRCH;
+    Thread::current()->donate_without_holding_big_lock(thread, "sys$donate_peer");
     return 0;
 }
 
--- a/Kernel/ThreadBlockers.cpp
+++ b/Kernel/ThreadBlockers.cpp
@@ -244,6 +244,11 @@ const FileDescription& Thread::FileDescriptionBlocker::blocked_description() con
     return m_blocked_description;
 }
 
+RefPtr<Thread> Thread::FileDescriptionBlocker::pick_donate_target() const
+{
+    return m_blocked_description->file().likely_peer_thread();
+}
+
 Thread::AcceptBlocker::AcceptBlocker(FileDescription& description, BlockFlags& unblocked_flags)
     : FileDescriptionBlocker(description, BlockFlags::Accept | BlockFlags::Exception, unblocked_flags)
 {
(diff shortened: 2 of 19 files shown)


Kernel/ThreadBlockers.cpp
```diff
@@ -436,6 +441,13 @@ void Thread::SelectBlocker::was_unblocked(bool did_timeout)
     }
 }
 
+RefPtr<Thread> Thread::SelectBlocker::pick_donate_target() const
+{
+    if (m_fds.is_empty())
+        return nullptr;
+    return m_fds[0].description->file().likely_peer_thread();
```
alimpfard, line 448: Why the first one?
Why not the second one, or the most starved one, or the least scheduled one, or why not just distribute equally?

alimpfard: (that is to say, please at least describe your reasoning for picking that specific fd as a comment above that, so people won't be stuck thinking "is there a specific reason" and will just see the specific reason)

Userland/Libraries/LibIPC/Connection.h
```diff
@@ -101,6 +101,10 @@ class Connection : public Core::Object {
             total_nwritten += nwritten;
         }
 
+#ifdef __serenity__
+        (void)donate_peer(m_socket->fd());
```
bgianfo, line 105: nitpick: should this be a separate commit, it's not mentioned in the title / description.

Userland/Libraries/LibC/unistd.cpp
```diff
@@ -681,6 +681,12 @@ int donate(int tid)
     __RETURN_WITH_ERRNO(rc, rc, -1);
 }
 
+int donate_peer(int fd)
+{
+    int rc = syscall(SC_donate_peer, fd);
```
bgianfo, line 686: nitpick: Add `+ LibC` to the commit title?

Kernel/Syscalls/sched.cpp
```diff
@@ -29,6 +30,21 @@ KResultOr<int> Process::sys$donate(pid_t tid)
     return 0;
 }
 
+KResultOr<int> Process::sys$donate_peer(int fd)
+{
+    REQUIRE_PROMISE(stdio);
```
bgianfo, line 35: The description claims only sockets are supported, should this be `REQUIRE_PROMISE_FOR_SOCKET_DOMAIN(...)` or even directly `REQUIRE_PROMISE(unix)`?

Also we don't seem to validate that the description is a socket anywhere?

awesomekling:
>LibIPC has been updated to use this after writing a complete IPC message to a socket.

This feels like it should only be done for synchronous messages, no?

gunnarbeutner:
> > LibIPC has been updated to use this after writing a complete IPC message to a socket.
> 
> This feels like it should only be done for synchronous messages, no?

That makes a lot of sense. In which case the `donate_peer` system call kind of loses its purpose because the client would be blocking on `select()` again. I think I might just remove it.

PR #6871, opened 2021-05-05
Kernel: Add Processor::is_bootstrap_processor() function, and use it.

bgianfo:
The variety of checks for `Processor::id() == 0` could use some assistance
int he readability department. This change adds a new function to
represent this check, and replaces the comparison everywhere it's used.


linusg:
`int he` -> `in the`

Issue #6772, opened 2021-05-01
BuildIt.sh - Spaces in media mount name causes script to fail


sneakernets:
Using Ubuntu 20.10 64-bit version.

```
100  132M  100  132M    0     0  13.6M      0  0:00:09  0:00:09 --:--:-- 13.8M
Extracting binutils...
/media/dell/Memorex USB/SERENITY/serenity/Toolchain/Tarballs/binutils-2.36.1 /media/dell/Memorex USB/SERENITY/serenity/Toolchain/Tarballs /media/dell/Memorex USB/SERENITY/serenity/Toolchain
/media/dell/Memorex USB/SERENITY/serenity/Toolchain/Tarballs /media/dell/Memorex USB/SERENITY/serenity/Toolchain
Extracting gcc...
/media/dell/Memorex USB/SERENITY/serenity/Toolchain/Tarballs/gcc-11.1.0 /media/dell/Memorex USB/SERENITY/serenity/Toolchain/Tarballs /media/dell/Memorex USB/SERENITY/serenity/Toolchain
/media/dell/Memorex USB/SERENITY/serenity/Toolchain/Tarballs /media/dell/Memorex USB/SERENITY/serenity/Toolchain
/media/dell/Memorex USB/SERENITY/serenity/Toolchain
/media/dell/Memorex USB/SERENITY/serenity/Toolchain/Build/i686 /media/dell/Memorex USB/SERENITY/serenity/Toolchain
/media/dell/Memorex USB/SERENITY/serenity/Toolchain/Build/i686/binutils /media/dell/Memorex USB/SERENITY/serenity/Toolchain/Build/i686 /media/dell/Memorex USB/SERENITY/serenity/Toolchain
XXX configure binutils
[binutils/configure] configure: error: path to source, /media/dell/Memorex USB/SERENITY/serenity/Toolchain/Tarballs/binutils-2.36.1, contains spaces
XXX build binutils
[binutils/build] make: *** No targets specified and no makefile found.  Stop.
[binutils/install] make: *** No rule to make target 'install'.  Stop.
/media/dell/Memorex USB/SERENITY/serenity/Toolchain/Build/i686 /media/dell/Memorex USB/SERENITY/serenity/Toolchain
XXX serenity libc, libm and libpthread headers
[...]

PR #6872, opened 2021-05-05
Meta+Toolchain: Enable linting of shell scripts under Toolchain

bgianfo:
-  __Toolchain: Make BuildPython.sh shellcheck compliant.__

    Shell checks unable to source non-literal includes,
    so inform shellcheck to just ignore this include.

 -  __Toolchain: Make BuildFuseExt2.sh shellcheck compliant.__

    BuildFuseExt2.sh was saying it should be run under /bin/sh but it is
    using bash extensions like pushd/popd, ${BASH_SOURCE[0]}, etc. So just
    run it under bash to avoid any potential issues.

 -  __Toolchain: Fix expansion bugs and make BuildIt.sh shellcheck compliant.__

    BuildIt.sh had a bunch of SC2086 errors, where we were not quoting
    variables in variable expansions. The logic being:

        Quoting variables prevents word splitting and glob expansion,
        and prevents the script from breaking when input contains spaces,
        line feeds, glob characters and such.

        Reference: https://github.com/koalaman/shellcheck/wiki/SC2086

    As bcoles noticed in #6772, shell check actually found a real bug here,
    where the users path did have a space in it.

    Closes: #6772

 - __Meta: Enable linting of shell scripts under Toolchain__

    Now that everything under Toolchain is shellcheck clean,
    remove it from the exception list.

### DIFF
--- a/Meta/lint-shell-scripts.sh
+++ b/Meta/lint-shell-scripts.sh
@@ -9,7 +9,6 @@ if [ "$#" -eq "0" ]; then
     mapfile -t files < <(
         git ls-files -- \
             '*.sh' \
-            ':!:Toolchain' \
             ':!:Ports' \
             ':!:Userland/Shell/Tests' \
             ':!:Base/home/anon/tests'
--- a/Toolchain/BuildFuseExt2.sh
+++ b/Toolchain/BuildFuseExt2.sh
@@ -1,4 +1,7 @@
-#!/bin/sh
+#!/usr/bin/env bash
+set -e
+# This file will need to be run in bash.
+
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 export PATH="/usr/local/opt/m4/bin:$PATH"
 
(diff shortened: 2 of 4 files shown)


bgianfo:
cc: @bcoles

Rada8732 (approved):

bcoles (approved):
I don't care about shells other than Bash so this looks fine to me.

I'm surprised `shellcheck` doesn't care about `$DIR` versus `${DIR}` but I don't remember why I thought the latter was proffered and can't be bothered to find out why.

`shellcheck` and `Meta/lint-shell-scripts.sh` both pass.

linusg:
Nice! Possible typo in first commit message: `Shell checks unable` -> `Shellcheck is unable`?

bcoles:
> 
> 
> Nice! Possible typo in first commit message: `Shell checks unable` -> `Shellcheck is unable`?

So long as we're nitpicking commit messages...

Inconsistent use of `.` commit subject terminator :

image

Grammar: missing possessive`'` for `user's` :

```
where the users path did have a space in it.
```

bgianfo:
I cleaned up the commit messages as requested, thanks!

PR #6879, opened 2021-05-05
Kernel: Fix `write`s to `ProcFS`

SpencerCDixon:
When using `sysctl` you can enable/disable values by writing to the
ProcFS. Some drift must of occured where writing was failing due to a
missing `set_mtime` call. Whenever one `write`'s a file the modified time
(mtime) will be updated so we need to implement this interface in ProcFS.


awesomekling (changes requested):
Nice catch! The fix looks good, but the commit message is slightly inaccurate. `set_mtime` is called on `write`, but not on `open` :)

SpencerCDixon:
Whoops good to know! Some reason thought that the `open` also modified it as if it was being `touch`ed. Updated comment to be explicit about `write` only

PR #6880, opened 2021-05-05
Solitaire: Improvements to the Solitaire GUI

trflynn89:
This started out as "hey let's add a statusbar for the score", and became a bit of a journey to make that possible. Overall this:
* Converts Solitaire to use GML
* Changes the widget holding the Solitaire game to a frame for a nice bezel
* Updates the painting logic to only paint areas that actually need updating
* Adds a status bar to display the score and the elapsed time

New look:
solitaire1
solitaire2


awesomekling:
Strong QoL improvements through and through. :)

